Issue #825, opened 2026-06-19
Usage-aware task holding + global retry when usage cap is near


scoop:
## Summary

When the account is close to its Claude usage cap, Shepherd should stop fanning out *new* work that would race the limit to zero, and instead **hold** newly-submitted tasks until usage resets — then make it a one-click operation to **retry every session that halted** because the cap was hit.

Two cooperating capabilities:

- **Part A — Usage-aware task holding:** if account usage (5-hour *or* weekly window) is ≥ a threshold (default **80%**) *and there is already at least one session running*, a newly-submitted manual task is parked in a persistent held queue instead of spawning. Held tasks auto-release (FIFO) once usage drops below the threshold. An operator can always override and submit anyway.
- **Part B — Global retry:** a broadcast-style control that resumes/retries every session that halted on a usage limit, preselected for the operator, enabled once usage has reset.

This extends the existing usage machinery rather than inventing a parallel one. Auto-drain already self-gates on per-repo `usageCeilingPct`; the gap is the **manual** submission path (`POST /api/sessions`), which spawns regardless of usage, and the lack of any recovery affordance after a usage-driven halt.

## Current state (what already exists)

- **Usage tracking** — `UsageLimitsService` (`src/usage-limits.ts`) computes `session5h.pct` and `week.pct` against calibrated caps (calibrated via the `/usage` scrape, `HerdrUsageProbe` in `src/usage-probe.ts`). Surfaced over the `usage:limits` WS event (~30s tick + on calibration), `GET /api/usage/limits`, and TopBar gauges (`ui/src/lib/components/usage-gauges.ts`: `hotterGauge`, `gaugeColor`). Types: `UsageLimits` / `LimitWindow` in `ui/src/lib/types.ts`.
- **Auto-drain already honors a usage ceiling** — `computeNext()` (`src/drain-core.ts`) returns a `hold` with reason code `"usage"` when `usagePct >= usageCeilingPct` (per-repo `repo_config`, default 80). So *drain-spawned* tasks are covered; manual ones are not.
- **Manual submission path** — `POST /api/sessions` (`src/server.ts`) → `SessionService.create()` (`src/service.ts`) → `prepareSpawn()` → `herdr.start()`. No usage check anywhere on this path.
[...]

PR #1038, opened 2026-06-23
fix(usage-hold): hold manual tasks at threshold even when the herd is idle

scoop:
## Problem

> 5h Usage is at 100% yet I'm still able to create & run directly instead of queuing new jobs.

A manual `POST /api/sessions` at/above the usage hold threshold spawned **directly** instead of being held whenever no session was currently running.

## Root cause

`shouldHold()` (`src/usage-hold.ts`) required `activeSessionCount >= 1` as part of its AND-chain — a **deliberate** carve-out from #825 (*"if the herd is idle, let the task through — holding when nothing is running would only freeze all progress"*). That rationale assumes **headroom**: letting one task run at 80% with an idle herd is sensible. At 100% there is no headroom — the spawned task halts immediately into a useless usage-halted husk. With nothing running, the chain short-circuited to `false` and the task spawned, even at the cap.

## Fix

Drop the idle-herd carve-out; hold on **usage alone**: `enabled && !force && max(5h, week) >= holdPct`.

**Why full removal, not an at-cap-only variant** (load-bearing reason): the release sweeper `releaseHeldTasks` (`src/held-release.ts`) already *releases* on `max(5h,week) < holdPct` with **no** running-count term, and auto-drain (`src/drain-core.ts`) *holds* on `usagePct >= ceiling` alone. Dropping the carve-out makes **hold, release, and drain gate on one identical rule**. An at-cap-only hold would be a third behavior the sweeper couldn't mirror (it'd release below `holdPct` regardless of running count, contradicting its own hold condition).

## ⚠️ Deliberate deviation + disclosed behavior change

- **Deviates from #825's literal three-part trigger** (which specified "≥1 session already running"). Flagged here and in the `src/usage-hold.ts` comment per house rules.
- **Default-on behavior change, more aggressive:** at usage ≥ `usageHoldPct` (default 80%) with an idle herd, manual submits now hold instead of spawning. Operator opt-outs unchanged — **Submit anyway** (`force: true`) at create time, and the **Settings → Session** toggle.
- **No re-notify:** correcting `feat_usage_aware_holding_body` in place does **not** re-surface the 1.35.0 What's-New announcement to users who already dismissed it (the drawer keys off `id`/`sinceVersion`, not body text). The corrected wording reaches new users only; existing users get the more-aggressive default with no fresh prompt. Accepted — this is a fix to an existing feature, not worth burning a new announcement `id`; the live Settings hint + at-create **Submit anyway** remain the in-context signal.
- **No new feature-announcement entry:** this is a `fix:` correcting existing behavior; the catalog gate only arms on `feat(...)`. `[no-feature-entry]`

## Changes

- `src/usage-hold.ts` — drop `activeSessionCount` from `HoldDecisionInput` + gate; comment documents the deviation + the hold/release/drain symmetry.
- `src/server.ts` — `tryHoldNewTask`: remove the running-count computation + `activeSessionCount` arg.
- `ui/src/routes/+page.svelte` — `holdLikely` mirror gates on usage alone.
- `docs/external-task-api.md` — hold-gate section: three conditions → two.
- `docs-site/.../reference/configuration.md` — drop the stale "(and ≥1 session is running)".
- `ui/messages/en.json` + `de.json` — `settings_usage_hold_hint` + `feat_usage_aware_holding_body` no longer claim a running session is required.
- `test/usage-hold.test.ts` — truth table without `activeSessionCount`.
- `test/hold-gate.test.ts` — idle-herd case now asserts **held**; redundant running-session planting removed.

## Verification

- root: `bun run lint` ✓, `bunx tsc --noEmit` ✓, `bun test ./test` → **4597 pass / 0 fail**
- ui: `bun run check` ✓ (0 errors), `bun run check:i18n` ✓ (parity), `bun run test` → **1997 pass**
- pre-push gates (branch hygiene, feature-catalog, glossary, fallow) green


scoop (commented):
## Summary

The PR removes the `activeSessionCount >= 1` carve-out from the usage-hold gate so that a newly submitted task is held whenever `max(5h, week) >= holdPct`, regardless of whether any session is currently running. This directly satisfies the task: at 100% usage with an idle herd, tasks are now queued instead of spawned.

## Verification

- **No orphaned references**: grepped for `activeSessionCount` across the tree — zero matches remain (input interface field, `src/server.ts` computation of `running`, and all callers/tests removed consistently).
- **All `shouldHold` consumers updated**: `src/server.ts:1427` no longer passes `activeSessionCount`; the client-side mirror in `ui/src/routes/+page.svelte:413-416` drops the `store.sessions.filter(...).length >= 1` term and adds an explanatory comment. The pure function `src/usage-hold.ts:29` reduces to `enabled && !force && Math.max(s5h, week) >= holdPct`.
- **Symmetry claim is accurate**: the new doc-comment in `src/usage-hold.ts` asserts the gate now matches the release sweeper and auto-drain. Verified against `src/held-release.ts:39` (holds while `maxPct >= cfg.holdPct`, no running term) and `src/drain-core.ts:216` (`usagePct >= ceiling`, no running term). Both confirm symmetry.
- **Degraded-usage convention preserved**: `(0,0) -> Math.max(0,0) >= holdPct` is false, so unknown/uncalibrated usage never holds. Confirmed by the retained test at `test/usage-hold.test.ts:37`.
- **i18n parity**: the two changed keys — `settings_usage_hold_hint` and `feat_usage_aware_holding_body` — are updated in BOTH `ui/messages/en.json:1596,1608` and `ui/messages/de.json:1596,1608`, with the stale "a session is already running" / "bereits eine Sitzung läuft" phrasing removed in each. No key added to one catalog without the other.
- **Docs updated**: `docs/external-task-api.md` (all→both, dropped the running bullet) and `docs-site/.../configuration.md` (removed "and ≥1 session is running") are consistent with the new behavior. Grepped remaining "running session" mentions tree-wide — none refer to the hold gate.

No bugs, security issues, or quality problems found. The change is minimal, internally consistent, and well-documented.

PR #1041, opened 2026-06-23
fix(broadcast): honest delivered/queued/unreachable feedback

scoop:
## What & why

Reported bug: *"Broadcast doesn't send to selected sessions. It's a no-op."*

Broadcast is **not** actually broken — verified end-to-end against the live deployment (client dialog, `/api/broadcast`, `service.broadcast`, herdr delivery to a real **idle** Claude pane → "PONG", and a real broadcast to a busy session that returned `{sent:1,total:1}` and landed). The "no-op" is **silent queuing on a busy herd**: a steer to a working agent is queued by Claude Code and acted on only *after* its current turn — while the toast reported a flat `Broadcast sent · N/N`. With the herd busy (the operator's incident: IDLE 0 / BUSY 5), every steer queued invisibly and it looked like nothing happened.

Operator decision (confirmed twice): **keep delivery as-is, make the feedback honest** — don't interrupt working agents.

## Change

- `service.broadcast` now returns `{ delivered, queued, offline, total }`, classifying each target by its herdr `agentStatus` at send time:
  - **delivered** — live, non-`working` (idle/blocked/done) → acts ~now
  - **queued** — live `working` → acts after its current turn
  - **offline** — dead pane / unknown id → not delivered
- Delivery itself is **unchanged**: `replyToLive` (shared with single-session reply / retry) is untouched.
- The dialog's success toast names the breakdown honestly; the clean all-idle case stays terse. A **zero-reach** broadcast (`delivered + queued === 0`) now surfaces as a failure with **Retry** instead of a misleading "sent 0".
- i18n: `+toast_broadcast_delivered`, `+toast_broadcast_result` (EN+DE); `-toast_broadcast_sent`.

## Verification

- root `bun test` 4621 pass / 0 fail (added queued-classification + all-offline cases); `bun run lint`, `bunx tsc --noEmit` clean
- `cd ui && bun run check` 0 errors; `check:i18n` parity (1857 keys each); UI tests 1997 pass
- `fallow audit --fail-on-issues` exit 0; pre-push gates green

Shipped as `fix` (feedback refinement of an existing feature) — no feature-catalog or glossary entry required.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -6,7 +6,7 @@ import { basename, extname, join } from "node:path";
 import type { RepoConfig, SessionStore } from "./store";
 import type { EventHub } from "./events";
 import type { WorktreeMgr } from "./worktree";
-import type { HerdrDriver } from "./herdr";
+import type { HerdrAgent, HerdrDriver } from "./herdr";
 import { matchAgents } from "./herdr";
 import { config } from "./config";
 import type { CreateSessionInput, IssueRef, RelaunchOverrides, Session } from "./types";
@@ -2134,14 +2134,44 @@ export class SessionService {
     return { resumed, steered, total: ids.length };
   }
 
-  /** Fan a steer out to many sessions (human-style). Skips unknown ids and dead panes.
-   *  Lists herdr's live agents ONCE up front rather than per id, so a wide fan-out
-   *  doesn't spawn one blocking `herdr agent list` per target. */
-  broadcast(ids: string[], text: string): { sent: number; total: number } {
-    const live = this.liveTerminalIds();
-    let sent = 0;
-    for (const id of ids) if (this.replyToLive(id, text, live)) sent++;
-    return { sent, total: ids.length };
+  /** Fan a steer out to many sessions (human-style), classifying the outcome per target so
+   *  the operator gets honest feedback instead of a flat "sent N". Lists herdr's live agents
+   *  ONCE up front (both the live id set AND each agent's status) rather than per id, so a
+   *  wide fan-out doesn't spawn one blocking `herdr agent list` per target. Each live target's
+   *  steer is delivered identically to a single reply (delivery is unchanged); the count it
+   *  lands in just reflects the agent's status at send time:
+   *   - delivered: a live agent that is NOT `working` (idle/blocked/done) — acts on the steer ~now.
+   *   - queued:    a live `working` agent — Claude Code queues the steer; it acts after the
+   *                current turn ends (correct behavior, but invisible immediately — the reason a
+   *                busy-herd broadcast looked like a no-op).
+   *   - offline:   unknown id or dead pane — the steer wasn't delivered.
+   *  `delivered + queued + offline === total`. */
+  broadcast(
+    ids: string[],
+    text: string,
+  ): { delivered: number; queued: number; offline: number; total: number } {
+    let agents: HerdrAgent[];
+    try {
+      agents = this.deps.herdr.list();
+    } catch {
+      agents = []; // herdr unreachable → every target is offline (the steer can't land)
+    }
+    const live = new Set(agents.map((a) => a.terminalId));
+    const statusByTerminal = new Map(agents.map((a) => [a.terminalId, a.agentStatus]));
+    let delivered = 0;
+    let queued = 0;
+    let offline = 0;
+    for (const id of ids) {
+      if (!this.replyToLive(id, text, live)) {
+        offline++;
+        continue;
+      }
+      // replyToLive already confirmed the session + live pane; classify by its status.
+      const term = this.deps.store.get(id)?.herdrAgentId;
+      if (term && statusByTerminal.get(term) === "working") queued++;
+      else delivered++;
+    }
+    return { delivered, queued, offline, total: ids.length };
   }
 
   /**
--- a/src/service.ts
+++ b/src/service.ts
@@ -2162,13 +2162,13 @@ export class SessionService {
     let queued = 0;
     let offline = 0;
     for (const id of ids) {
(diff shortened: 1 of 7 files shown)


scoop (commented):
## Summary

The PR reframes `broadcast` from a flat `{sent, total}` count into `{delivered, queued, offline, total}`, so a busy-herd broadcast no longer reads as a no-op: live `working` panes are reported as `queued` (Claude Code queues the steer until the current turn ends) and an all-offline result is surfaced like a failure with Retry. Delivery itself is unchanged — `replyToLive` still sends to every live pane — which matches the task's framing that the no-op was a reporting/perception problem, not a delivery one.

## Verification

- `HerdrAgent` exposes `agentStatus` and `terminalId` (verified `src/herdr.ts:11-21`), and `HerdrState` includes the `"working"` value the classifier branches on (verified `src/types.ts:4`). The new `HerdrAgent` import is real (`src/herdr.ts:11`).
- Classification is exhaustive: each id increments exactly one of `offline`/`queued`/`delivered`, so `delivered + queued + offline === total` holds (verified `src/service.ts:2160-2173`). `replyToLive` guarantees the session exists and its `herdrAgentId` is in the `live` set before classification (verified `src/service.ts:2227-2229`), so the post-send `store.get(id)` lookup always resolves and its terminal is always present in `statusByTerminal` (same `agents` source).
- The new return shape flows through unchanged: `src/server.ts:3156` JSON-passes the service result; `ui/src/lib/api.ts:935` declares the matching type; `ui/src/lib/components/BroadcastDialog.svelte:60-81` consumes `delivered`/`queued`/`offline` and treats `delivered + queued === 0` as a failure (dialog stays open with Retry).
- i18n parity: `toast_broadcast_delivered` ({delivered}) and `toast_broadcast_result` ({delivered}, {queued}, {offline}) exist in BOTH catalogs with identical placeholders (verified `ui/messages/en.json:1858-1859` vs `ui/messages/de.json:1858-1859`). The removed `toast_broadcast_sent` has no remaining references anywhere (grep clean). `broadcast_failed`, still used on the failure path, is present in both (verified `ui/messages/en.json:688`, `ui/messages/de.json:688`). `de.json` JSON is well-formed after the trailing-comma edit (verified `ui/messages/de.json:1853-1860`).
- Tests updated to the new shape and exercise the working→queued, non-working→delivered, and all-offline paths (`test/service.test.ts`, `test/server-steers.test.ts`).

No blocking issues found.

## Nit

- `src/service.ts:2170`: `broadcast` re-fetches `this.deps.store.get(id)?.herdrAgentId` to classify status, but `replyToLive` already did `store.get(id)` one line earlier and confirmed the same row is live. The second lookup is redundant (harmless against the in-memory store); the status could be classified by threading the agent/terminal through `replyToLive` or by reading `s.herdrAgentId` once. Non-blocking.

- `ui/messages/{en,de}.json`: `broadcast_result_sent` ({sent}/{total}) has no references in `ui/src` and appears to be dead in both catalogs — predates this PR.

scoop (commented):
## Re-review

### Prior finding
1. **src/service.ts:2170 redundant lookup — ADDRESSED.** The new `broadcast` resolves the session exactly once via `this.deps.store.get(id)` (src/service.ts:2164) and passes it to the new `sendSteerTo(s, text)` helper (src/service.ts:2234), which no longer does its own `store.get`. Status classification uses the already-fetched `s.herdrAgentId` against `statusByTerminal` (src/service.ts:2171). No second lookup remains. Not re-raised.

### Verification of the change
- **Task satisfied.** `broadcast` actually delivers to each live selected session via `sendSteerTo` (records the reply signal + bracket-pastes + CR, src/service.ts:2234-2249), and now returns honest counts `{delivered, queued, offline, total}` so a busy-herd broadcast no longer reads as a silent no-op. `delivered + queued + offline === total` holds by construction.
- **Identifiers resolved.** `HerdrAgent` is exported (src/herdr.ts:11) and `list()` returns `HerdrAgent[]` (src/herdr.ts:211). `agentStatus`/`terminalId` fields exist (src/herdr.ts:13,19). The `"working"` literal compared at src/service.ts:2171 is a valid `HerdrState` (src/herdr.ts:91, mirrors haltAll at src/service.ts:2200). The herdr-unreachable `try/catch → agents = []` path correctly makes every target `offline`.
- **Exact-terminalId match preserved.** New broadcast uses `live.has(s.herdrAgentId)` (src/service.ts:2166), the same matching the original `replyToLive`/`liveTerminalIds` used (src/service.ts:2220,2229) — no behavior regression vs the prior fast-path.
- **Contract change propagated.** Return shape `{delivered, queued, offline, total}` flows through: server passes it straight back (src/server.ts:3160), `api.ts` return type updated (ui/src/lib/api.ts:935), and `BroadcastDialog.svelte` consumes `r.delivered/r.queued/r.offline` (lines 60,71-78). The all-offline (`delivered+queued===0`) branch surfaces `broadcast_failed` and keeps the dialog open for Retry — correct.
- **i18n parity verified.** `toast_broadcast_sent` removed from both en (ui/messages/en.json) and de (ui/messages/de.json). New keys `toast_broadcast_delivered` and `toast_broadcast_result` added to BOTH catalogs (en.json:1857-1858, de.json:1857-1858) with identical placeholders `{delivered}/{queued}/{offline}`, matching the call sites in BroadcastDialog.svelte:72,75-79.
- **No stale shape references.** Grep found no remaining `{sent, total}` broadcast consumers in src/ or ui/src/; tests updated to the new shape.

No blocking problems and no nits to raise.

- `broadcast_result_sent` remains in both message catalogs but appears unreferenced in ui/src — pre-existing, untouched by this diff.

PR #1042, opened 2026-06-23
fix(herdr): pin classic renderer for spawned claude agents

scoop:
## What

Inject `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1` into the `env` shim at the universal `herdr.start` spawn chokepoint (alongside `NODE_COMPILE_CACHE`), forcing the **classic renderer** for every spawned `claude` agent.

## Why — does the new fullscreen renderer break Shepherd?

Claude Code shipped a **fullscreen renderer** ([docs](https://code.claude.com/docs/en/fullscreen), v2.1.89+). It draws on the terminal's **alternate screen buffer** (like `vim`/`htop`) and **captures the mouse**.

**Today: not broken out of the box** — fullscreen is an opt-in research preview; the default is still the classic renderer.

**But Shepherd is exposed**, because its whole integration assumes the classic renderer:
- **Status / blocked detection** (`poller.ts`, `blocked.ts`) scrapes the rendered visible viewport (`herdr agent read --source visible`) with regexes tuned to the classic prompt layout. Fullscreen pins the input box at the bottom and only renders visible messages — a relocated/reshaped prompt risks misclassification.
- **Web terminal** (`Viewport.svelte` / xterm.js) forwards keystrokes back into the PTY. Fullscreen's mouse capture (modes 1000/1002/1003) would make operator clicks/scrolls in the browser emit mouse-event escape sequences **into claude's input**.
- Browser scrollback no longer accumulates the conversation (it lives in the alt buffer).

Fullscreen turns on via a **persisted `tui` setting** (`/tui fullscreen` → `~/.claude/settings.json`) or **ambient `CLAUDE_CODE_NO_FLICKER`** — either of which a Shepherd-spawned agent inherits from the operator's global config/env. And the docs warn the **default may flip** based on feedback, which would break every install at once with no operator action.

`CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1` "forces the classic renderer regardless of the saved `tui` setting" (and overrides `NO_FLICKER`), immunizing Shepherd against both the operator-settings path and a future default flip, at zero cost.

## Scope / safety notes
- Single chokepoint: every TUI claude spawn (interactive session, recap, review, distiller, plan-gate, optimizer, autopilot-llm, doc-agent, namer, herd-digest/rundown, merge-suggest) routes through `herdr.start`.
- Shepherd attaches via `herdr agent attach`, **not** `claude attach`, so the docs' "background sessions always use fullscreen" carve-out (which ignores this var) does **not** apply to our foreground spawns. Flagged in the code comment + here per house rules.
- Shepherd never reads `--source recent` scrollback, so the "scrollback goes empty" failure mode wasn't a direct hit — but pinning classic keeps that guarantee robust too.

## Verification
- `bun test ./test/herdr.test.ts` → 50 pass (updated 5 argv assertions + added a focused regression test for the pin)
- `bun run lint`, `bunx tsc --noEmit` clean
- `bunx fallow@2.97.0 audit --base origin/main --fail-on-issues` clean
- pre-push gate green

Server-only plumbing, no UI surface → exempt from feature catalog / i18n / glossary.

### DIFF
--- a/src/herdr.ts
+++ b/src/herdr.ts
@@ -360,7 +360,26 @@ export class HerdrDriver {
             .sort(([a], [b]) => a.localeCompare(b))
             .map(([k, v]) => `${k}=${v}`)
         : [];
-      const wrapped = ["env", `NODE_COMPILE_CACHE=${compileCacheDir()}`, ...envTokens, ...argv];
+      // Pin the CLASSIC renderer for every spawned claude. Claude Code's opt-in fullscreen
+      // renderer (v2.1.89+) draws on the terminal's ALTERNATE screen buffer and captures the
+      // mouse — Shepherd's whole integration assumes the classic renderer: the poller/blocked
+      // classifier scrape the rendered viewport (`agent read --source visible`) with a layout
+      // tuned to the classic prompt, and the web terminal forwards xterm keystrokes (mouse
+      // capture would inject click/scroll escape sequences into claude's input). Fullscreen
+      // turns on via a persisted `tui` setting or ambient CLAUDE_CODE_NO_FLICKER, either of
+      // which a spawned agent would inherit from the operator's global config/env — and the
+      // docs warn the default may flip. CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1 forces the
+      // classic renderer regardless of the saved setting or NO_FLICKER, immunizing us against
+      // both. (Shepherd attaches via `herdr agent attach`, not `claude attach`, so the docs'
+      // "background sessions always use fullscreen" carve-out — which ignores this var — does
+      // not apply to our foreground spawns.)
+      const wrapped = [
+        "env",
+        `NODE_COMPILE_CACHE=${compileCacheDir()}`,
+        "CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1",
+        ...envTokens,
+        ...argv,
+      ];
       // Collision-breaker: if herdr rejects with `agent_name_taken` (a stale same-named
       // agent is still registered — e.g. shepherd restarted while an interactive `claude`
       // was running), resolve the squatter(s) by name from `list()`, close their tabs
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

The PR pins Claude Code's CLASSIC renderer for every spawned agent by injecting `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1` into the `env` shim built in `HerdrDriver.start`. This addresses the task (the new fullscreen renderer draws on the alternate screen buffer and captures the mouse, which would break Shepherd's viewport scraping and the xterm web terminal).

## Verification

- **Single chokepoint covers all spawns.** The pin is added inside `HerdrDriver.start` (src/herdr.ts:376-382). Every claude spawn routes through this method — verified callers: src/verify-key.ts:174, src/usage-probe.ts:66, src/autopilot-llm.ts:201, src/doc-agent.ts:710, src/standalone-critic.ts:424, src/distiller.ts:266, src/service.ts:1272. None build their own argv shim, so all inherit the pin. The fix is complete, not partial.
- **Rationale is grounded in real code.** The comment cites `agent read --source visible` scraping; confirmed at src/herdr.ts:427 and src/herdr.ts:444 (poller reads the visible buffer), so the alternate-screen concern is legitimate.
- **No env-token precedence collision.** Grep for `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN`/`CLAUDE_CODE_NO_FLICKER` across src returns only the three new lines in herdr.ts — no caller passes this key in its `env` map, so the pin (placed before caller-supplied `envTokens` at src/herdr.ts:380) is never overridden by a later `env KEY=VALUE` token.
- **Tests updated consistently.** All four affected assertions plus the new dedicated test were updated to expect the var as the third post-`--` token (test/herdr.test.ts:108, 129-134, 137-149, 437-442, 474-479); the `extractPost` helper (test/herdr.test.ts:424) matches the production token order.
- No user-facing strings or message catalogs are touched, so i18n parity is not in play.

## Stated limitations (could not verify — external to this repo)

- The runtime behavior of `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1` — that it forces the classic renderer and overrides a persisted `tui` setting and `CLAUDE_CODE_NO_FLICKER` — is Claude Code internal behavior not present in this repo. Assumed as written per the PR's documented research; not verifiable here.

Nothing to change. No findings.

Issue #824, opened 2026-06-19
Filter issue lists to mine + unassigned (Backlog + New Task)


scoop:
## Problem

Issue lists in the app surface **every** open issue, including ones already assigned to other people. On repos with shared backlogs (e.g. the Flow Agent repo) that means the **New Task** issue picker and the **Backlog** issue panel both show issues someone else owns, which I have to mentally skip past. I want to filter the lists down to issues that are **unassigned or assigned to me** — and hide the rest.

## Desired behaviour (decided)

- A **toggle** ("Mine & unassigned" / "Hide others' issues") on both issue lists.
- **On by default** — lists hide others' issues out of the box; flip it off to see everything.
- **Persisted globally** via `localStorage` — one shared setting across all repos and both surfaces (Backlog + New Task), remembered across reloads.
- "Show this issue" rule: `assignees` is empty **OR** `assignees` includes the current GitHub user's login.
- **Fail open**: if the current user can't be determined (offline/unauth, or a `local`-mode repo), the toggle is inert and all issues show — never hide everything because we don't know who "me" is.

## Why this is non-trivial today

The issue data the UI receives carries **no assignee info**, and the UI **never learns the current GitHub user**. Both gaps must be closed before any client-side filter can work.

- `Issue` interface (server `src/forge/types.ts`, mirrored in `ui/src/lib/types.ts`) has `number, title, body, url, labels, createdAt` — **no `assignees`**.
- `GitHubForge.listIssues()` (`src/forge/github.ts`) requests `--json number,title,body,url,labels,createdAt` from `gh issue list` — **`assignees` not requested**.
- The current user login is server-only via `GitHubForge.currentUser()` (`src/forge/github.ts`, cached; `gh api user --jq .login`). It is **not exposed to the UI** anywhere.

## Shared carrier

Both surfaces fetch through the **same** path: `listIssues(repoPath)` in `ui/src/lib/api.ts` → `GET /api/issues?repo=` → `handleIssues` in `src/server.ts`, which returns `{ slug, webUrl, issues }`. So a single change to that response covers both:

- New Task picker: `PromptSources.svelte` (renders `issues` from `listIssues`).
- Backlog panel: `IssuesPanel.svelte` (fetches via `listIssues`, filters with `filterIssues` in `ui/src/lib/components/issues-panel.ts`).

### Server
[...]

PR #1046, opened 2026-06-23
feat(backlog): show per-issue assignees when mine & unassigned filter is off

scoop:
## What

Expose a chip per **assignee login** on each backlog issue row — but only when the **"mine & unassigned" filter** (#824) isn't hiding others' issues. The filter is "in use" exactly when `hideOthers && viewer != null`, so chips show when it's toggled **off** or **fails open** because the viewer is unknown. With the filter active, every visible issue is already mine-or-unassigned, so an assignee chip would be redundant noise.

## Why

When you turn the filter off to see the whole backlog (including issues assigned to others), "who already owns this" is the missing context. Surfacing the assignee inline answers it at a glance.

## Details

- **Data already flows end-to-end** — `Issue.assignees: string[]` (logins only, no avatar URLs) is fetched by the forge and already drives the #824 filter. **No server change**; pure UI render.
- `IssueRow` gains an opt-in `showAssignees` prop; `IssuesPanel` passes `!issuesFilter.hideOthers || viewer == null`.
- Chips **lead** the label row (who-owns-this is the primary signal), reuse the `.label-chip` recipe with a person glyph (`👤`) + **verbatim, non-uppercased** login (logins are mixed-case). Loop keyed `(login)` per existing convention.
- **Unassigned issues show no chip** (absence implies unassigned).
- i18n: `issuerow_assignee_title` + feature-catalog keys added to **both** EN + DE (parity gate green).
- Feature catalog: `issue-assignee-chip` entry (`sinceVersion 1.37.0`, no targetId — chips mount conditionally; surfaced via the What's-New drawer).

## Tests

Three new cases in `IssuesPanel.browser.test.ts` cover all axes:
- filter **on** + viewer known → no chips
- filter **off** + viewer known → chips for assigned rows, none for unassigned
- **fail-open** (`viewer == null`, filter default on) → chips render — guards the `|| viewer == null` clause

## Verification

- `cd ui && bun run check` ✓ (0 errors)
- `cd ui && bun run check:i18n` ✓ (1860 keys, parity)
- `cd ui && bun run test` ✓ (2002 passed)
- root `bun run lint` ✓
- pre-push gate ✓ (branch hygiene + i18n + feature catalog + glossary + fallow)

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -1856,5 +1856,8 @@
   "feat_first_task_confirm_title": "Einstellungen vor der ersten Aufgabe prüfen",
   "feat_first_task_confirm_body": "Das Starten einer Aufgabe in einem neuen Repository zeigt nun einen kurzen Überprüfungsschritt, damit du die Automatisierungseinstellungen vor dem ersten Durchlauf prüfen kannst. Nach einmaliger Bestätigung starten weitere Aufgaben im selben Repository ohne Unterbrechung.",
   "toast_broadcast_delivered": "Broadcast hat {delivered} Agenten sofort gesteuert.",
-  "toast_broadcast_result": "Broadcast · {delivered} sofort gesteuert, {queued} bei beschäftigten Agenten in Warteschlange (wird nach aktuellem Zug ausgeführt), {offline} nicht erreichbar."
+  "toast_broadcast_result": "Broadcast · {delivered} sofort gesteuert, {queued} bei beschäftigten Agenten in Warteschlange (wird nach aktuellem Zug ausgeführt), {offline} nicht erreichbar.",
+  "issuerow_assignee_title": "Zugewiesen an {login}",
+  "feat_issue_assignee_title": "Sieh, wer welchem Issue zugewiesen ist",
+  "feat_issue_assignee_body": "Wenn der Filter „meine & nicht zugewiesen“ ausgeschaltet ist, zeigt jedes Issue im Backlog nun einen Chip für jede zugewiesene Person, sodass du auf einen Blick erkennst, wer bereits an einer Aufgabe arbeitet."
 }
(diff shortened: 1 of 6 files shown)


scoop (commented):
## Summary

The PR exposes per-issue assignee chips in the backlog `IssueRow`, gated on `showAssignees = !issuesFilter.hideOthers || viewer == null`, plus matching i18n keys and a What's-New announcement. The change satisfies the task and I found no blocking issues.

## What I verified

- **Gating logic is sound.** `showAssignees` (ui/src/lib/components/IssuesPanel.svelte:63) suppresses chips exactly when the mine&unassigned filter would have already narrowed every visible row to mine-or-unassigned. Confirmed against `hideOthers` (ui/src/lib/components/issues-panel.ts:48-52): it no-ops when `viewer == null`, so the `|| viewer == null` fail-open clause correctly re-enables chips in that case. The three new tests (ui/src/lib/components/IssuesPanel.browser.test.ts:374-460) cover filter-on, filter-off, and viewer-unknown, matching this behavior.
- **Type contract.** `issue.assignees: string[]` exists and is non-optional (ui/src/lib/types.ts:104). The render condition and `{#each issue.assignees as login (login)}` (ui/src/lib/components/issues-panel/IssueRow.svelte:93,96) access it directly — consistent with the pre-existing unguarded `issue.labels.length` access on the same line.
- **i18n parity.** All three keys (`issuerow_assignee_title`, `feat_issue_assignee_title`, `feat_issue_assignee_body`) are present in BOTH ui/messages/en.json:1860-1862 and ui/messages/de.json:1860-1862, with the same `{login}` placeholder. JSON closing braces/commas are well-formed in both.
- **Feature announcement.** `targetId` is optional on the `FeatureAnnouncement` interface (ui/src/lib/feature-announcements.ts:15), so omitting it is valid. `id: "issue-assignee-chip"` is unique in the file. `sinceVersion: "1.37.0"` is consistent: root package.json:3 is at 1.36.0 (latest tag v1.36.0) and the other recent announcements already use 1.37.0.
- **Single consumer.** `IssueRow` is only rendered from IssuesPanel.svelte:222, and `{showAssignees}` is wired at line 230. The new prop defaults to `false`, so it is backward-compatible.
- **Keying.** `{#each ... (login)}` keys by login, which is unique within an issue's assignee set — no duplicate-key risk.

No findings.

Issue #1054, opened 2026-06-24
test(tui): validate fullscreen OPTION_RE guard at multiple terminal widths


scoop:
Follow-up to #1047 (fullscreen-renderer opt-in).

## Context
#1047 re-tuned `OPTION_RE` (`src/blocked.ts`) to accept the fullscreen renderer's zero-space option packing (`2.Yes, allow all edits…`), guarded so decimals (`2.5 GB`) and bare-token runs (`1.tsx`/`2.x`) can't forge a menu. The accept rule for a zero-space option is: **label is non-digit-leading AND contains internal whitespace**.

That guard was tuned to and tested against the spike's **54-col** fullscreen-menu fixture (the only width capturable in-session — herdr exposes no terminal-column control, so genuine multi-width capture was deferred from #1047 rather than faked).

## Known limitation to validate
At a **narrow** width, a long option could wrap so its first visible line is a short token without internal whitespace (e.g. `2.Yes,`), which the phrase guard would then reject → that option drops and the menu mis-classifies. The spike (#1043) explicitly caveated that the no-space packing may be width-influenced.

## Task
- Capture real fullscreen permission-menu frames at **narrow (~27-col)** and **wide (~80-col)** widths via the production `HerdrDriver` read path (same protocol as #1043), in addition to the existing 54-col fixture.
- Add them as dual-renderer fixtures and assert `classifyBlocked` still returns the full multi-option `menu` at every width.
- If a narrow width breaks the phrase guard, refine the zero-space accept rule to what the captures actually show (e.g. also accept a label ending in sentence punctuation / a continuation marker) — driven by the fixtures, not assumption.

## Phase-0 result carried over (no longer in question)
`CLAUDE_CODE_NO_FLICKER=1` was empirically confirmed to override a persisted `tui: default` (classic) in the bound config dir, so the env-only opt-in is sound — only the width robustness of the menu-parse guard remains open.

scoop:
✅ PR #1679 merged.

Issue #1043, opened 2026-06-23
spike: can Shepherd read the fullscreen-renderer alt-screen frame? (compat go/no-go)


scoop:
## Context

Claude Code shipped an opt-in **fullscreen renderer** ([docs](https://code.claude.com/docs/en/fullscreen), v2.1.89+): it draws on the terminal's **alternate screen buffer** and **captures the mouse**. PR #1042 pinned the **classic** renderer for every spawned agent (`CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1` in the `herdr.start` env shim) so we're immune to a default flip or an operator's persisted `tui` setting.

This issue is the **handoff for making Shepherd *compatible*** with fullscreen (so we could opt in for its real upside: **flat memory on long autonomous drains**). Flicker/mouse are irrelevant to an unattended agent — memory stability is the only motivation.

## Why this is a spike, not an implementation

Shepherd's two load-bearing signals are already renderer-independent:
- **Status** (working/idle/done) — JSONL transcript signals + herdr `agent_status`; the terminal `--source visible` read is only a *fallback* for stall detection (`poller.ts:935,1027`).
- **Block trigger** — push via `Notification(permission_prompt)` hooks since #704, explicitly "independent of … the PTY-regex *detection* heuristics" (`poller.ts:536-547`).

So fullscreen degrades **enrichment + fallback quality**, not core detection. The thing that *does* depend on the rendered frame is `classifyBlocked(visible)` (`blocked.ts:70`), which scrapes the visible frame for menu options / yes-no / tail (the hook carries no payload).

## The one real unknown (this spike)

**Does `herdr agent read --source visible` actually capture the alt-screen-buffer frame of a fullscreen claude?** herdr is tmux-like, so probably yes — but everything downstream hinges on it. This is a Phase-0 go/no-go per house rules: run BEFORE building anything on top.

## Acceptance / deliverable

1. Spawn a real claude in fullscreen (`CLAUDE_CODE_NO_FLICKER=1`, or `/tui fullscreen`) inside a herdr pane.
2. Capture what `herdr agent read --source visible` returns for: (a) idle prompt, (b) a permission/menu prompt ("1. Yes / 2. No …"), (c) a working spinner. Save the raw frames as fixtures.
3. **Go/no-go:**
[...]

PR #1055, opened 2026-06-24
feat(tui): make Shepherd compatible with Claude Code fullscreen renderer (opt-in)

### LINKED ISSUE #1047: feat(tui): make Shepherd compatible with Claude Code fullscreen renderer (opt-in)
Follow-up to the **GO** outcome of spike #1043 (see `docs/spikes/1043-fullscreen-alt-screen.md` + fixture corpus in `test/fixtures/fullscreen/`).

## Spike result (GO)
`herdr agent read --format text --source visible` (the real `HerdrDriver.read` path) **does** capture the alt-screen-buffer frame of a fullscreen-renderer claude (v2.1.89+). Verified end-to-end: synthetic alt-buffer marker read back, `/tui` confirmed `Current renderer: fullscreen`, and the real `classifyBlocked`/`hasActiveSpinner` ran against captured frames. Idle + working-spinner classify correctly **as-is**; the permission menu needs one regex re-tune.

## Scope
- [ ] **Re-tune `OPTION_RE`** (`src/blocked.ts`) to accept the fullscreen no-space `N.Label` packing. Under fullscreen, the long option renders as `2.Yes, allow all edits…` (no space after the delimiter), which `OPTION_RE`'s `[.)]\s+` misses → the contiguous-run check breaks → the whole menu collapses to `awaiting-input`. A minimal zero-space-tolerant variant was verified to recover all 3 fullscreen options **and** leave classic parsing identical.
  - Guard the relaxed pattern against new false positives (e.g. `2.5 GB`); the contiguous-run guard helps but isn't sufficient alone.
  - Validate against fullscreen menus at **multiple widths** — the spike's fullscreen pane (54 cols) was wider than the classic baseline (27 cols), so the no-space packing may be width- as well as renderer-influenced.
- [ ] **Dual-renderer regression tests** over the `test/fixtures/fullscreen/` corpus: keep the classic fixtures green while adding the fullscreen frames. (`SPINNER_RE`/`tailLines` already pass on the corpus as-is — spinner detected, options within the 15-line window — add coverage anyway.)
- [ ] **Stall detection** (`poller.ts`): confirm the `--source visible` diff still advances under flicker-free rendering (fullscreen repaints in place; ensure a live turn still reads as a *changing* buffer so the frozen-TUI stall heuristic isn't tripped).
- [ ] **Web-terminal mouse handling**: decide between embracing xterm mouse forwarding vs `CLAUDE_CODE_DISABLE_MOUSE=1` (fullscreen captures the mouse).
- [ ] **Gate behind an opt-in per-spawn `tui` toggle.** Never flip the default — this is a "behavior may change" research preview. The #1042 classic pin (`CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1`) stays the default.

## Motivation
Fullscreen's only relevant upside for unattended agents is **flat memory on long autonomous drains** (flicker/mouse are irrelevant headless). Status (JSONL + `agent_status`) and block-trigger (Notification hooks, #704) are already renderer-independent; only `classifyBlocked(visible)` enrichment depends on the frame.

Handoff from #1043.

scoop:
Closes #1047. Follow-up to the **GO** outcome of spike #1043.

Lets an operator opt the **main agent session** (interactive + autonomous drain) into Claude Code's fullscreen renderer for its flat-memory upside on long drains — **without changing the default** (classic stays pinned) and without breaking blocked/stall classification or the web terminal.

## What changed

- **`OPTION_RE` re-tune** (`src/blocked.ts`) — tolerate the fullscreen renderer's zero-space option packing (`2.Yes, allow all edits…`), which the classic `\s+`-after-delimiter pattern missed (collapsing the whole menu to `awaiting-input`). The relaxation is guarded so two new false-positive classes can't forge a menu: decimals (`2.5 GB`) and bare-token runs (`1.tsx`/`2.x`/`1.txt`). Accept rule for a zero-space option: label is non-digit-leading **and** contains internal whitespace; the contiguous 1,2,3-run check remains the backstop.
- **Two global settings** `tuiFullscreen` + `tuiDisableMouse` (both default **false**), following the existing `fableAvailable` pattern across `config.ts` → `index.ts` (persist) → `server.ts` (GET + PUT) → `api.ts` → `types.ts` → `Settings.svelte` (SESSION panel) → EN+DE i18n → feature catalog (`sinceVersion: 1.37.0`).
- **Per-spawn renderer env** — `prepareSpawn` computes the main session's renderer env (`mainSessionRendererEnv`) and applies it on **both** spawn paths:
  - **trusted** (incl. all satellites): the `herdr.start` `env` shim.
  - **sandboxed** (standard/autonomous — the long-drain target): the membrane's existing `extraEnv` → `--setenv`, which survives `bwrap --clearenv` (the outer shim does not).
  - Default pins classic (`CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1`); `tuiFullscreen` sets `CLAUDE_CODE_NO_FLICKER=1`.
- **`herdr.start` classic pin made conditional** — skipped when the caller already supplies a renderer var, so the main session can never end up with both `NO_FLICKER` and the classic pin (the pin would otherwise win under `env` last-wins and silently defeat the opt-in), while **satellites keep the classic pin unchanged**.
- **Stall detection** — no code change needed; added poller regression coverage that a frozen fullscreen frame reads as frozen (alt-screen chrome doesn't fake buffer advancement) while an advancing spinner stays suppressed, and that the idle fullscreen frame isn't mistaken for a live turn.

## Deliberate deviations (per house rules)

- **#1042's classic pin is now conditional, not unconditional.** Building on the merged #1042, the renderer pin is routed through the spawn-env decision and skipped when a renderer var is already supplied. Side effect: the classic pin now genuinely reaches the **sandboxed** main session for the first time (the outer `env` shim was previously wiped by `bwrap --clearenv`) — a strict improvement, tested.
- **`tuiFullscreen` auto-implies `DISABLE_MOUSE`.** Every main session is reachable in the web terminal, where fullscreen mouse-capture escapes would be injected into the operator's keystroke stream, and there's no spawn-time "headless" signal. So fullscreen always disables mouse capture; the separate `tuiDisableMouse` toggle remains meaningful in classic mode. Documented in the setting copy.
- **Multi-width menu validation deferred → #1054.** herdr exposes no terminal-column control in-session, so the `OPTION_RE` guard is tuned to and tested against the spike's 54-col fixture. Narrow-width packing robustness is tracked in #1054 rather than faked here.

## Verification (Phase-0 empirical go/no-go)

Spawned a real fullscreen `claude` (`CLAUDE_CODE_NO_FLICKER=1`) bound to a config dir with `tui: default` (classic) persisted; `/tui` reported **`Current renderer: fullscreen`** — confirming `NO_FLICKER` overrides a persisted classic setting, so the env-only opt-in is sound (no `--settings` overlay escalation needed).

## Gates

- root `bun run lint` + `bunx tsc --noEmit` + `bun test ./test` → **4674 pass / 0 fail**
- `cd ui && bun run check` (0 errors) + `check:i18n` (1882 keys, EN/DE parity) + `bun run test` → **2014 pass**
- branch-hygiene ✓ · feature-catalog ✓ · glossary ✓ · `fallow@2.97.0 audit --fail-on-issues` ✓ · pre-push ✓

### DIFF
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -46,6 +46,21 @@ lines), read by the systemd unit if present.
 
 See [Operating Shepherd](/operating/) for the host-level `/etc/fstab` belt.
 
+## Main agent terminal renderer (research preview)
+
+Every spawned `claude` runs on Claude Code's **classic** renderer by default —
+Shepherd's poller/blocked classifier scrape the rendered viewport and the web
+terminal forwards xterm keystrokes, both of which assume the classic prompt.
+The operator can opt the **main agent session** (satellites always stay classic)
+into Claude Code's opt-in fullscreen renderer. The choice applies to newly
+spawned/resumed sessions only and is also configurable from the Settings panel
+(persisted in the SQLite `settings` table); the env vars below seed a fresh DB.
+
+| Variable | Default | Purpose |
+| --- | --- | --- |
+| `SHEPHERD_TUI_FULLSCREEN` | `0` (off) | Set `1` to opt the main agent session into Claude Code's fullscreen renderer (research preview). Implies `SHEPHERD_TUI_DISABLE_MOUSE`. |
+| `SHEPHERD_TUI_DISABLE_MOUSE` | `0` (off) | Set `1` to disable Claude Code mouse capture for the main agent session, so fullscreen mouse-capture escape sequences don't leak into the web terminal's keystroke stream. |
+
 ## Documentation automation (PR-gated doc agent)
 
 Opt-in, default-off. When enabled, a manual trigger (`POST /api/doc-agent?repo=<path>`)
--- a/src/blocked.ts
+++ b/src/blocked.ts
@@ -19,8 +19,12 @@ export interface BlockReason {
 }
 
 const TAIL_LINES = 15;
-// Matches "1. Yes", "❯ 2. No", "│  3) Foo" — captures the digit and the label.
-const OPTION_RE = /^[\s│|]*[❯>*]?\s*(\d+)[.)]\s+(.*\S)\s*$/;
+// Matches "1. Yes", "❯ 2. No", "│  3) Foo", AND the fullscreen renderer's zero-space packing
+// of the long option ("2.Yes, allow all edits…"). Group 2 captures the delimiter spacing
+// (undefined when packed with no space); group 3 is the label. The zero-space form is gated
+// in classifyBlocked (see below) so a decimal ("2.5 GB") or a bare-token run ("1.tsx"/"2.x")
+// can't forge a menu.
+const OPTION_RE = /^[\s│|]*[❯>*]?\s*(\d+)[.)](\s+)?(\S.*?)\s*$/;
 const YES_NO_RE = /\(\s*y\s*\/\s*n\s*\)|\[\s*y\s*\/\s*n\s*\]/i;
 // eslint-disable-next-line no-control-regex
 const ANSI_RE = /\x1b\[[0-9;]*[A-Za-z]/g;
@@ -75,9 +79,15 @@ export function classifyBlocked(text: string): BlockReason {
   for (const line of tail) {
     const m = OPTION_RE.exec(line);
     if (!m) continue;
-    const n = Number(m[1]);
-    if (n === run.length + 1) run.push({ label: m[2]!, send: m[1]! });
-    else if (n === 1) run = [{ label: m[2]!, send: m[1]! }];
+    const [, num, spaced, label] = m;
+    // Fullscreen drops the delimiter space ONLY on the long, wrapping option
+    // ("2.Yes, allow all edits…"); short options keep theirs. Accept a zero-space option
+    // ONLY when its label is a non-digit-leading multi-token phrase, so a decimal
+    // ("2.5 GB" → "5 GB") or a bare-token run ("1.tsx"/"2.x"/"1.txt") can't forge a menu.
+    if (!spaced && (/^\d/.test(label!) || !/\s/.test(label!))) continue;
+    const n = Number(num);
+    if (n === run.length + 1) run.push({ label: label!, send: num! });
+    else if (n === 1) run = [{ label: label!, send: num! }];
   }
   if (run.length >= 2) return { shape: "menu", options: run, tail };
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -433,6 +433,12 @@ export const config = {
   // stored session model (so cost accounting + fable intent survive for later replay).
   // Persisted + UI-configurable; SHEPHERD_FABLE_AVAILABLE=0/false seeds it off.
   fableAvailable: normalizeFableAvailable(process.env.SHEPHERD_FABLE_AVAILABLE) ?? true,
+  // Opt the main agent session into Claude Code's fullscreen renderer (research preview;
+  // applies to newly spawned/resumed sessions only). Persisted + UI-configurable; default off.
+  tuiFullscreen: process.env.SHEPHERD_TUI_FULLSCREEN === "1",
+  // Disable Claude Code mouse capture for the main agent session. Persisted + UI-configurable;
+  // default off. (tuiFullscreen also implies this — coupling lives in the spawn wiring, not here.)
+  tuiDisableMouse: process.env.SHEPHERD_TUI_DISABLE_MOUSE === "1",
   // Operator auth footing for spawned agents. 'subscription' (default) = subscription OAuth;
   // 'api-key' = bill against an Anthropic API key. Persisted + UI-configurable; env seeds a fresh DB.
   authMode: normalizeAuthModeSetting(process.env.SHEPHERD_AUTH_MODE) ?? "subscription",
--- a/src/herdr.ts
+++ b/src/herdr.ts
@@ -360,23 +360,24 @@ export class HerdrDriver {
             .sort(([a], [b]) => a.localeCompare(b))
             .map(([k, v]) => `${k}=${v}`)
         : [];
-      // Pin the CLASSIC renderer for every spawned claude. Claude Code's opt-in fullscreen
-      // renderer (v2.1.89+) draws on the terminal's ALTERNATE screen buffer and captures the
-      // mouse — Shepherd's whole integration assumes the classic renderer: the poller/blocked
-      // classifier scrape the rendered viewport (`agent read --source visible`) with a layout
-      // tuned to the classic prompt, and the web terminal forwards xterm keystrokes (mouse
-      // capture would inject click/scroll escape sequences into claude's input). Fullscreen
-      // turns on via a persisted `tui` setting or ambient CLAUDE_CODE_NO_FLICKER, either of
-      // which a spawned agent would inherit from the operator's global config/env — and the
-      // docs warn the default may flip. CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1 forces the
-      // classic renderer regardless of the saved setting or NO_FLICKER, immunizing us against
-      // both. (Shepherd attaches via `herdr agent attach`, not `claude attach`, so the docs'
-      // "background sessions always use fullscreen" carve-out — which ignores this var — does
-      // not apply to our foreground spawns.)
+      // Pin the CLASSIC renderer for every spawned claude UNLESS the caller already specified a
+      // renderer choice in `env`. The main-session spawn (prepareSpawn) computes its own renderer
+      // env — classic pin by default, or CLAUDE_CODE_NO_FLICKER when the operator opted into the
+      // fullscreen research preview (tuiFullscreen) — and routes it through BOTH the membrane
+      // --setenv and this shim, so re-adding the pin here would duplicate it (and `env`'s last-wins
+      // would let the pin override an intended NO_FLICKER). Satellites pass no renderer var and keep
+      // the classic pin exactly as before.
+      // Claude Code's opt-in fullscreen renderer (v2.1.89+) draws on the terminal's ALTERNATE
+      // screen buffer and captures the mouse — Shepherd's integration assumes the classic renderer
+      // (poller/blocked classifier scrape `agent read --source visible`; the web terminal forwards
+      // xterm keystrokes). (Shepherd attaches via `herdr agent attach`, not `claude attach`, so the
+      // docs' "background sessions always use fullscreen" carve-out does not apply.)
+      const callerSetRenderer =
+        !!env && ("CLAUDE_CODE_NO_FLICKER" in env || "CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN" in env);
       const wrapped = [
         "env",
         `NODE_COMPILE_CACHE=${compileCacheDir()}`,
-        "CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1",
+        ...(callerSetRenderer ? [] : ["CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1"]),
         ...envTokens,
         ...argv,
       ];
--- a/src/index.ts
+++ b/src/index.ts
@@ -173,6 +173,12 @@ if (savedFa !== null) {
   const v = normalizeFableAvailable(savedFa);
   if (v !== null) config.fableAvailable = v;
 }
+// a UI-chosen fullscreen-renderer opt-in (persisted) overrides the env default; absent → keep
+// the config default (off). Stored as "1"/"0".
+const savedTuiFs = store.getSetting("tuiFullscreen");
+if (savedTuiFs !== null) config.tuiFullscreen = savedTuiFs === "1";
+const savedTuiMouse = store.getSetting("tuiDisableMouse");
+if (savedTuiMouse !== null) config.tuiDisableMouse = savedTuiMouse === "1";
 // a UI-chosen auth mode (persisted) overrides the env seed; absent or unrecognised → keep default.
 const savedAm = store.getSetting("authMode");
 if (savedAm !== null) {
--- a/src/server.ts
+++ b/src/server.ts
@@ -2943,6 +2943,9 @@ async function handleSettings({ req, parts, deps }: Ctx): Promise<Response | nul
       usageHoldPct: config.usageHoldPct,
       // global fable availability flag; false = fable spawns reroute to opus[1m].
       fableAvailable: config.fableAvailable,
+      // TUI renderer opt-in (research preview) + mouse-capture disable; apply to new/resumed sessions.
+      tuiFullscreen: config.tuiFullscreen,
+      tuiDisableMouse: config.tuiDisableMouse,
       // doc-agent soak flags (read-only; env-driven; no PUT patch).
       docAgentEnabled: config.docAgentEnabled,
       docAgentAct: config.docAgentAct,
@@ -2977,6 +2980,8 @@ const SETTING_PATCHES: [string, (value: unknown, deps: Ctx["deps"]) => Response]
   ["usageHoldEnabled", putUsageHoldEnabled],
   ["usageHoldPct", putUsageHoldPct],
   ["fableAvailable", putFableAvailable],
+  ["tuiFullscreen", putTuiFullscreen],
+  ["tuiDisableMouse", putTuiDisableMouse],
 ];
 
 function putRemoteControl(value: unknown, deps: Ctx["deps"]): Response {
@@ -3106,6 +3111,24 @@ function putFableAvailable(value: unknown, deps: Ctx["deps"]): Response {
   config.fableAvailable = v; // live: next spawn picks it up
   deps.store.setSetting("fableAvailable", String(v)); // persist across restarts
   return json({ fableAvailable: config.fableAvailable });
+}
+
+function putTuiFullscreen(value: unknown, deps: Ctx["deps"]): Response {
+  if (typeof value !== "boolean") {
+    return json({ error: "tuiFullscreen must be a boolean" }, 400);
+  }
+  config.tuiFullscreen = value; // live: next spawn/resume picks it up (renderer is fixed at process start)
+  deps.store.setSetting("tuiFullscreen", value ? "1" : "0"); // persist
+  return json({ tuiFullscreen: config.tuiFullscreen });
+}
+
+function putTuiDisableMouse(value: unknown, deps: Ctx["deps"]): Response {
+  if (typeof value !== "boolean") {
+    return json({ error: "tuiDisableMouse must be a boolean" }, 400);
+  }
+  config.tuiDisableMouse = value; // live: next spawn/resume picks it up
+  deps.store.setSetting("tuiDisableMouse", value ? "1" : "0"); // persist
+  return json({ tuiDisableMouse: config.tuiDisableMouse });
 }
 
 function putRepoRoot(value: unknown, deps: Ctx["deps"]): Response {
--- a/src/service.ts
+++ b/src/service.ts
@@ -845,12 +845,6 @@ function pickOverride<T>(override: T | undefined, original: T): T {
   return override !== undefined ? override : original;
 }
 
-/**
- * A freshly-created session is pre-approved into the build queue only when the repo runs the
- * build queue AND the session is effectively on autopilot (so no human approval gate will ever
- * come) AND it isn't a research task. Keyed on the session's EFFECTIVE autopilot (not the raw
- * repo default), so an autopilot-off session — e.g. a merge-train driver — is never auto-approved.
- */
 /** Renderer env for the MAIN session spawn. Default: pin the classic renderer. The
  *  tuiFullscreen opt-in (research preview) switches to NO_FLICKER. tuiFullscreen also implies
  *  DISABLE_MOUSE (every main session is reachable in the web terminal, where fullscreen
@@ -868,6 +862,12 @@ function mainSessionRendererEnv(
   return env;
 }
 
+/**
+ * A freshly-created session is pre-approved into the build queue only when the repo runs the
+ * build queue AND the session is effectively on autopilot (so no human approval gate will ever
+ * come) AND it isn't a research task. Keyed on the session's EFFECTIVE autopilot (not the raw
+ * repo default), so an autopilot-off session — e.g. a merge-train driver — is never auto-approved.
+ */
 function shouldPreApproveBuildQueue(
   repoConfig: RepoConfig,
   session: Pick<Session, "autopilotEnabled">,
--- a/src/service.ts
+++ b/src/service.ts
@@ -1231,6 +1231,19 @@ export class SessionService {
     // FS-only (network open) — warrants the egress-degraded banner.
     const egressDegraded = isEgressDegraded(profile, backend, egressBackend ?? null);
 
+    // Renderer env for the MAIN session ONLY (satellites call herdr.start directly and keep the
+    // classic pin). Default: pin classic. The tuiFullscreen opt-in (research preview) switches to
+    // NO_FLICKER. tuiFullscreen ALSO implies DISABLE_MOUSE — every main session is reachable in the
+    // web terminal, where fullscreen mouse-capture escapes (modes 1000/1002/1003) would be injected
+    // into the keystroke stream; tuiDisableMouse can also set it independently (e.g. in classic mode).
+    // Applied via BOTH the membrane --setenv (sandboxed; the outer env shim is wiped by
+    // bwrap --clearenv) AND spawnEnv (trusted). The renderer is fixed at claude process start, so a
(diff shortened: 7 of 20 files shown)


scoop (commented):
## Summary

This PR implements the #1047 fullscreen-renderer opt-in: it re-tunes `OPTION_RE` for fullscreen's zero-space option packing, adds a global `tuiFullscreen`/`tuiDisableMouse` opt-in (default off) wired through config/persist/server/UI, and routes the renderer env through both the membrane `--setenv` and the `herdr.start` env shim. I verified the correctness-relevant claims against the actual code; everything holds except one cosmetic JSDoc placement issue.

## What I verified

**OPTION_RE re-tune** (`src/blocked.ts:27`). New pattern `/^[\s│|]*[❯>*]?\s*(\d+)[.)](\s+)?(\S.*?)\s*$/`. Group 1=digit, 2=optional delimiter spacing, 3=label. The destructure `const [, num, spaced, label] = m` (`src/blocked.ts:82`) maps 1/2/3 correctly. Classic spaced options (`1. Yes`) keep identical results (spaced truthy → guard skipped). Zero-space lines are gated by `if (!spaced && (/^\d/.test(label) || !/\s/.test(label))) continue` (`src/blocked.ts:87`): `2.5 GB` (digit-leading) and `1.tsx` (single-token) are correctly rejected, while `2.Yes, allow all edits…` (multi-token, non-digit-leading) is accepted. The residual heuristic gap the issue itself flagged (`1.Word phrase`/`2.Word phrase` prose could still forge a 2-run) is acknowledged and accepted as a non-blocker. `OPTION_RE` has no consumers outside `blocked.ts` (grepped).

**Renderer env wiring.** `mainSessionRendererEnv` (`src/service.ts:860-869`) returns the classic pin by default, `CLAUDE_CODE_NO_FLICKER=1` on opt-in, and adds `CLAUDE_CODE_DISABLE_MOUSE=1` when `tuiFullscreen || tuiDisableMouse` — matching the issue's mouse-capture decision. It is applied to BOTH the membrane `extraEnv` (`src/service.ts:1271`) and `spawnEnv` (`src/service.ts:1293`). The `callerSetRenderer` guard (`src/herdr.ts:375-376`, `380`) drops the hard-coded classic pin only when the caller already supplied a renderer var, so the main session never duplicates/overrides it. I confirmed all 13 satellite `herdr.start` callers (verify-key, usage-probe, standalone-critic, review, recap, plan-gate, autopilot-llm, optimizer, namer-llm, herd-digest, distiller, doc-agent, merge-suggest) pass no renderer var → classic pin preserved exactly as before. `collectPassthroughEnv` (`src/sandbox.ts:208-219`) carries only locale/display vars, confirming the comment that the sandboxed path relies on the membrane `--setenv` (the outer `env` shim is wiped by `bwrap --clearenv`, `src/sandbox.ts:417`).

**Config / persistence / server.** `config.tuiFullscreen`/`tuiDisableMouse` seed from `SHEPHERD_TUI_*` (`src/config.ts:438-441`); restored from store as `"1"`/`"0"` (`src/index.ts:170-173`); surfaced in GET (`src/server.ts:2946-2948`); PUT validators reject non-boolean and persist `"1"`/`"0"` (`src/server.ts:3116-3132`). Forms are consistent across writer/reader.

**i18n parity.** All 8 new keys exist in both catalogs — `ui/messages/en.json` (diff) and `ui/messages/de.json:1606-1611,1883-1884`. Svelte usages (`settings_tui_fullscreen_label/hint/save_failed`, `settings_tui_disable_mouse_label/hint/save_failed`) resolve; reuses existing `settings_usage_hold_on/off`.

**Types & announcement.** `Settings` interface extended (`ui/src/lib/types.ts:75-78`); `Settings.browser.test.ts:72-73` payload updated; feature announcement `sinceVersion: "1.37.0"` (`ui/src/lib/feature-announcements.ts`) is consistent with package version 1.36.0 (latest tag v1.36.0).

## Finding (non-blocking nit)

`src/service.ts`: the new `mainSessionRendererEnv` function was inserted between the JSDoc block at lines 848-853 (which describes build-queue pre-approval: "…pre-approved into the build queue…never auto-approved") and the `shouldPreApproveBuildQueue` function it documents (now at line 871). As a result that JSDoc is orphaned and `shouldPreApproveBuildQueue` is left undocumented, while `mainSessionRendererEnv` ends up with two JSDoc blocks stacked above it. Move the new function (with its own JSDoc) to before line 848 or after `shouldPreApproveBuildQueue` to keep the build-queue doc attached to its function.

## Notes

The issue's checkbox 5 asks for a "per-spawn `tui` toggle"; the implementation is a global operator setting that applies to newly spawned/resumed main sessions. This still satisfies the core requirement (opt-in, default off, classic pin remains the default) and the comments describe the behavior accurately, so I treat it as an acceptable interpretation, not a defect.

The `poller.ts` stall-detection item (#1047 checkbox 3) is a verify-only task and no production code change was required; coverage was added in `test/poller.test.ts` (tests not run, per CI ownership).

scoop (commented):
## Re-review summary

The fullscreen-renderer opt-in is implemented cleanly and the dual-renderer regex re-tune is correct. The one prior-revision finding (JSDoc orphaning in `src/service.ts`) is **not** addressed and is re-raised below. No new blocking issues found.

### Prior finding status

**1. `src/service.ts` JSDoc orphaning — NOT addressed.** `mainSessionRendererEnv` (with its own doc) still sits at lines 854-869, between the build-queue JSDoc (848-853) and `shouldPreApproveBuildQueue` (871). Verified against `src/service.ts:848-881`: the build-queue JSDoc at 848-853 is separated from the `shouldPreApproveBuildQueue` it documents by the new function, and `shouldPreApproveBuildQueue` at 871 has no adjacent doc. `git diff` confirms `src/service.ts` is in scope and the insertion landed at exactly the flagged location. Re-raised verbatim.

### Verification performed (correctness claims)

- **`OPTION_RE` re-tune + zero-space guard** (`src/blocked.ts:27`, `src/blocked.ts:82-91`): destructure `[, num, spaced, label] = m` maps correctly to groups 1/2/3 of `/^[\s│|]*[❯>*]?\s*(\d+)[.)](\s+)?(\S.*?)\s*$/`. Walked the guard `if (!spaced && (/^\d/.test(label) || !/\s/.test(label))) continue;` against the issue's cases: `2.5 GB` → label `5 GB`, leading-digit → rejected; `1.tsx` → label `tsx`, no whitespace → rejected; `2.Yes, allow all edits…` → multi-token non-digit → accepted. Classic path (`spaced` defined) bypasses the guard entirely, so classic parsing is unchanged. Non-greedy `(\S.*?)\s*$` reproduces the old `(.*\S)` trailing-trim behavior.
- **herdr renderer pin gating** (`src/herdr.ts:375-383`): `callerSetRenderer` skips the classic pin when `env` carries either renderer var; satellites (no renderer var) keep the pin. Confirmed against the two new tests `test/herdr.test.ts:483-505` (NO_FLICKER → pin omitted; DISABLE_ALTERNATE_SCREEN → exactly one, no duplicate).
- **rendererEnv wiring** (`src/service.ts:1252`, `:1271`, `:1293`): applied to both membrane `extraEnv` and `spawnEnv`, matching the comment's both-paths claim. `mainSessionRendererEnv(config.tuiFullscreen, config.tuiDisableMouse)` resolves against `src/config.ts:438-441`.
- **config/seed/persist chain**: `tuiFullscreen`/`tuiDisableMouse` defined in `src/config.ts:438-441` (env `SHEPHERD_TUI_FULLSCREEN`/`SHEPHERD_TUI_DISABLE_MOUSE`), DB-override seeded in `src/index.ts:170-174`, GET-exposed in `src/server.ts:2946-2948`, PUT-patched via `putTuiFullscreen`/`putTuiDisableMouse` (`src/server.ts:2983-2984`, `:3116-3133`). Doc env-var names match (`docs-site/.../configuration.md` table).
- **i18n parity**: all 8 new keys (`settings_tui_fullscreen_label/hint/save_failed`, `settings_tui_disable_mouse_label/hint/save_failed`, `feat_tui_fullscreen_title/body`) present in BOTH `ui/messages/en.json:1606-1611,1883-1884` and `ui/messages/de.json:1606-1611,1883-1884`. Settings.svelte reuses existing `settings_usage_hold_on/off` for toggle state labels.
- **UI type/test parity**: `Settings` interface gains both fields (`ui/src/lib/types.ts:75-78`); `SettingsPayload` is an alias of `Settings` (`Settings.browser.test.ts:5`), and the test fixture seeds both (`:72-73`). `putTuiFullscreen`/`putTuiDisableMouse` in `ui/src/lib/api.ts:179-184` match the server return shapes. Feature announcement added (`ui/src/lib/feature-announcements.ts:1500-1508`).

- `src/service.ts:1293`: on the TRUSTED (non-membrane) spawn path the `env` shim does not `--clearenv`, so an operator with an ambient `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1` in their global env would have it leak into claude alongside the opt-in `CLAUDE_CODE_NO_FLICKER=1`, and classic would win — i.e. the fullscreen opt-in silently no-ops. Degrades safely (no crash) and requires an unusual ambient config; the membrane path is immune via `--clearenv` + `--setenv`. Worth a guard (explicitly unset/override the pin when opting into fullscreen) but not blocking.

scoop:
**Re: `mainSessionRendererEnv` orphaning the build-queue JSDoc** — already fixed in commit `5c6ad1ed` (current branch HEAD), which this review appears to predate.

The cited line numbers match the pre-fix layout. Current `src/service.ts` ordering:

- L848 — `mainSessionRendererEnv`'s **own** JSDoc
- L854 — `function mainSessionRendererEnv(...)`
- L866 — the build-queue JSDoc ("A freshly-created session is pre-approved …")
- L871 — `function shouldPreApproveBuildQueue(...)`

So the build-queue JSDoc is reattached to `shouldPreApproveBuildQueue` and the helper carries its own doc — exactly the requested resolution (move the new function, with its doc, ahead of the build-queue JSDoc). No further change needed; re-running against `5c6ad1ed` should clear it.

scoop (commented):
## Summary

This PR implements the issue #1047 scope cleanly: re-tuned `OPTION_RE` for fullscreen no-space packing (guarded against false positives), dual-renderer fixtures, stall-detection regression coverage, mouse-capture disable, and an opt-in per-session toggle (default off, classic pin stays the default). No blocking problems found.

## Re-review of prior finding

1. **`mainSessionRendererEnv` orphaning the build-queue JSDoc** — RESOLVED. Verified at `src/service.ts:848-863`: the helper now carries its own JSDoc (848-853), and the build-queue JSDoc (865-870) sits directly above `shouldPreApproveBuildQueue` (871), correctly reattached. The author note matches the actual current layout. Not re-raised.

## Verification performed

- **`OPTION_RE` re-tune + guard** (`src/blocked.ts:27`, `87-89`): traced the regex `/^[\s│|]*[❯>*]?\s*(\d+)[.)](\s+)?(\S.*?)\s*$/` and the gate `if (!spaced && (/^\d/.test(label) || !/\s/.test(label))) continue;` against each fixture/case. `2.Yes, allow all edits…` (multi-token, non-digit) is accepted; `2.5 GB`→label `5 GB` (digit-leading) is rejected; `1.tsx`/`2.x` (single-token) are rejected. Confirmed against the real fixture `test/fixtures/fullscreen/fullscreen-menu.txt` which packs only the long option (`1. Yes` spaced, `2.Yes, allow all edits during this session` no-space, `3. No` spaced) — the contiguous-run logic recovers all 3, and the intervening `(shift+tab)` non-match doesn't break the run. Classic `classic-menu.txt` path is unchanged.
- **Renderer env wiring**: `mainSessionRendererEnv` (src/service.ts:854-863) returns NO_FLICKER (fullscreen) or DISABLE_ALTERNATE_SCREEN (classic), plus DISABLE_MOUSE when fullscreen-or-tuiDisableMouse. Verified it routes through BOTH the membrane `extraEnv` (src/service.ts:1271) → `--setenv` (confirmed by test/sandbox.test.ts:671-695 hasTriple, placed after `--clearenv`) AND `spawnEnv` (src/service.ts:1293). The herdr shim's new `callerSetRenderer` guard (src/herdr.ts:375-376) correctly suppresses the duplicate classic pin when `env` carries either renderer var, avoiding the last-wins override (src/herdr.ts:380). Both sandboxed (bwrap --clearenv wipes the shim, --setenv carries it) and trusted (shim carries it) paths are covered.
- **Config/persistence/API chain**: `SHEPHERD_TUI_FULLSCREEN`/`SHEPHERD_TUI_DISABLE_MOUSE` env seeds (src/config.ts:437-441), persisted keys `tuiFullscreen`/`tuiDisableMouse` read in src/index.ts:172-175 and written in src/server.ts:3116-3133 — setting-key spelling consistent across index/server/store. GET payload (src/server.ts:2946-2948), PUT patches registered (src/server.ts:2983-2984), UI api.ts:178-184, types.ts:75-79, Settings.svelte toggles all consistent.
- **i18n parity**: all 8 new keys (`settings_tui_fullscreen_label/hint/save_failed`, `settings_tui_disable_mouse_label/hint/save_failed`, `feat_tui_fullscreen_title/body`) present in BOTH ui/messages/en.json and ui/messages/de.json (verified 8 occurrences each). Settings.svelte reuses existing `settings_usage_hold_on/off` keys.
- **Feature announcement version**: `sinceVersion: "1.37.0"` (feature-announcements.ts:1505) is consistent with package.json `1.36.0` (next release), matching sibling announcements.
- **Stall detection** (test/poller.test.ts:2784-2826): frozen fullscreen frame re-arms the block; advancing frame stays suppressed — src/poller.ts is unchanged, so this is confirmatory coverage only, as the issue intended. `spinnerHarness`/`setText`/`workingBlockedSnapshot` helpers exist (test/poller.test.ts:170-236).

## Stated limitations (could not verify, not findings)

- The issue asked to validate fullscreen menus at *multiple widths* (no-space packing may be width-influenced). The tests cover the single ~54-col fixture only. If a narrower width also packs a SHORT single-token option (e.g. `1.Yes`→label `Yes`), the single-token guard at src/blocked.ts:88 would reject it and collapse the menu. I cannot confirm whether claude packs short options at narrow widths — that depends on Claude Code's renderer, whose behavior isn't reproducible in this repo. Flagged as a coverage gap to be aware of, not a confirmed defect.
- The loosened pattern accepts a contiguous run of no-space multi-token numbered lines (e.g. `1.Foo bar` / `2.Baz qux`) as a menu. This is the documented tradeoff; the named false-positives from the issue (`2.5 GB`, bare-token runs) are guarded, and such prose is atypical in a claude viewport. Acceptable as designed.

scoop (commented):
## Summary

This PR makes Shepherd compatible with Claude Code's fullscreen renderer as an opt-in research preview. The implementation is sound and satisfies the issue's scope: `OPTION_RE` is re-tuned with guards, dual-renderer fixtures are added, the renderer choice is gated behind a per-setting opt-in that never flips the default, mouse capture is handled, and satellites stay on the classic pin.

### Verification performed

**`OPTION_RE` retune (src/blocked.ts:27, 79-91)** — Verified the new regex `^[\s│|]*[❯>*]?\s*(\d+)[.)](\s+)?(\S.*?)\s*$` and the guard at blocked.ts:87:
- Fullscreen menu fixture (test/fixtures/fullscreen/fullscreen-menu.txt) packs `1. Yes` / `2.Yes, allow all edits…` / `3. No`; the contiguous-run logic captures all three (run 1→2→3; the `(shift+tab)` line between options 2 and 3 simply doesn't match and is skipped).
- Decimal guard: a bare `2.5 GB` line → `label="5 GB"`, `spaced` undefined, label starts with a digit → rejected (blocked.ts:87). Correct.
- Bare-token guard: `1.tsx` / `2.x` → single-token label, no whitespace → rejected. Correct.
- Group indices: destructure `[, num, spaced, label]` correctly maps to groups 1/2/3 (label moved from old m[2] to new m[3]); `send` uses the digit. Correct.
- Classic path unchanged: classic menus always carry the delimiter space, so `spaced` is defined and the new guard at blocked.ts:87 is never entered — verified against the classic-menu fixture (test asserts 3 options, no regression).

**Renderer env wiring (src/service.ts:847-861, 1251-1294; src/herdr.ts:363-383)** — Verified the env reaches the inner `claude` on both paths:
- Sandboxed (willWrap): `rendererEnv` is spread into `membrane.extraEnv` (service.ts:1271) → emitted as `--setenv` inside bwrap (sandbox.ts:425-429), surviving `--clearenv`.
- Trusted/no-backend: `rendererEnv` is spread into `spawnEnv` (service.ts:1293) → `herdr.start` `env` shim tokens (herdr.ts:358-362). The `callerSetRenderer` check (herdr.ts:375-376) sees the renderer key and skips re-adding the classic pin, so no duplication/override. Confirmed.
- Satellites (verify-key.ts:174, review.ts:356, recap.ts:471, etc.) pass api-key/passthrough env with NO renderer key → `callerSetRenderer=false` → classic pin retained. Correct.
- `collectPassthroughEnv` (sandbox.ts:198-219) forwards only LANG/LC_*/TZ/COLORTERM — it does not leak an ambient `CLAUDE_CODE_NO_FLICKER` into the sandbox; and `rendererEnv` is spread last so it wins in both extraEnv and spawnEnv. Confirmed.

**Defaults & persistence** — `config.tuiFullscreen`/`tuiDisableMouse` default off (config.ts:438-441); env seeds (`SHEPHERD_TUI_FULLSCREEN==="1"`); persisted overrides re-applied at boot (index.ts:170-174); PUT handlers validate boolean and persist as "1"/"0" (server.ts:3116-3133). Default behavior (classic pin) is preserved.

**i18n parity** — All 8 new keys exist in BOTH ui/messages/en.json:1879-1886 and ui/messages/de.json:1879-1886 (`settings_tui_fullscreen_{label,hint,save_failed}`, `settings_tui_disable_mouse_{label,hint,save_failed}`, `feat_tui_fullscreen_{title,body}`). Reused toggle-state keys `settings_usage_hold_on/off` confirmed present (en.json:1596-1597). Verified.

**Feature announcement** — `sinceVersion: "1.37.0"` is consistent with package.json version 1.36.0 and the existing batch of unreleased 1.37.0 entries (feature-announcements.ts:1432-1515). Correct.

**Types/UI** — `Settings` interface gains both fields (ui/src/lib/types.ts:75-78); api helpers (api.ts:179-184) and Settings.svelte toggles wire through; browser test fixture updated (Settings.browser.test.ts:72-73).

### Findings

1. **test/blocked.test.ts:217-221** — The decimal false-positive test uses `"Available space: 1.5 GB"` / `"Used space: 2.5 GB"`. These lines begin with letters, so `OPTION_RE`'s `^[\s│|]*[❯>*]?\s*(\d+)` anchor never matches them at all — the test passes even if the blocked.ts:87 guard were deleted, so it does not actually exercise the decimal guard it claims to cover. Use bare lines that start at column 0 (e.g. `"2.5 GB"`, `"1.5 GB"`) to truly drive the `label` starts-with-digit branch. (Non-blocking; the production guard itself is correct.)

2. **src/blocked.ts:27,87** — The relaxation is not gated by renderer: any `N.<multi-word>` line with no delimiter space is now accepted as a menu option on BOTH renderers. The contiguous-run + tail-window + blocked-only constraints make this low-risk, but a no-space prose numbered list (`1.Do this` / `2.Do that`) near the bottom of a blocked agent's viewport could now be misclassified as a menu where the old `[.)]\s+` pattern rejected it. The issue flagged this exact false-positive risk; the chosen guard covers decimals and bare tokens but not no-space prose. Worth a brief note / consider tightening (e.g. require a leading box-glyph/cursor marker for the zero-space variant) if such frames appear in the corpus. (Non-blocking.)

## Stated limitations

- Could not verify Claude Code's runtime precedence when both `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN` and `CLAUDE_CODE_NO_FLICKER` are set simultaneously (third-party CLI behavior, not in this repo) — the design relies on DISABLE_ALTERNATE_SCREEN winning for the classic-pin immunization; assumed as documented in the in-code comments.

- The fullscreen opt-in forces mouse-disable while the `tuiDisableMouse` UI toggle can read `off` (Settings.svelte) — a cosmetic state mismatch the fullscreen hint already explains; not a defect.

scoop (commented):
## Summary

This PR opts the main agent session into Claude Code's fullscreen renderer behind a default-off, per-operator toggle, re-tunes `OPTION_RE` to recover the fullscreen no-space option packing, and keeps the classic pin as the default everywhere else. I traced every scope item in #1047 and verified the renderer-env plumbing, the regex/guard, i18n parity, and the feature-announcement/version bookkeeping against the actual code. The implementation satisfies the task and I found no blocking defects.

### Renderer wiring (verified correct)
- `mainSessionRendererEnv` (`src/service.ts:856-862`) returns `CLAUDE_CODE_NO_FLICKER=1` when `tuiFullscreen`, else `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1`, and adds `CLAUDE_CODE_DISABLE_MOUSE=1` when `tuiFullscreen || tuiDisableMouse`. The fullscreen→disable-mouse implication documented in `docs-site/src/content/docs/reference/configuration.md:60` is enforced here (config keys stay independent, `src/config.ts:438-441`), so docs and code agree.
- The env is routed through BOTH the membrane `--setenv` (`src/service.ts:1271`, via `extraEnv`) and the trusted env shim (`src/service.ts:1293`, via `spawnEnv`). This is necessary because `bwrap --clearenv` wipes the outer shim in the sandboxed case — confirmed against the comment at `src/service.ts:1251-1253`.
- `herdr.ts:375-382`: the classic pin is now omitted only when the caller already set a renderer var (`callerSetRenderer`). The main session always passes one, so no duplication / last-wins override; satellites pass none and keep the classic pin exactly as before. Verified the `env` shim adds the explicit pin via `envTokens` (`herdr.ts:358-362`) for the trusted classic main session, so ambient `CLAUDE_CODE_NO_FLICKER` is still overridden — the protection from the old comment is preserved.

### `OPTION_RE` re-tune + guards (verified correct, classic identical)
- New regex `src/blocked.ts:27` makes the delimiter space optional (`(\s+)?`) and captures it as group 2; gate at `src/blocked.ts:82-86` rejects a zero-space option unless its label is non-digit-leading AND multi-token.
- Traced `2.5 GB` → label `5 GB`, digit-leading → skipped; `1.tsx`/`2.x` → single-token → skipped; `2.Yes, allow all edits…` → accepted. For spaced (classic) lines the gate is a no-op and the capture is equivalent to the old `(.*\S)`, so classic parsing is unchanged.
- Confirmed against the real fixture `test/fixtures/fullscreen/fullscreen-menu.txt`: the line `   2.Yes, allow all edits during this session` is the no-space option, and the intervening `(shift+tab)` line is a non-match that `continue`s without breaking the contiguous 1/2/3 run — so all three options are recovered.

### i18n / UI / version (verified)
- All 8 new message keys (6 `settings_tui_*` + `feat_tui_fullscreen_title`/`_body`) exist in BOTH `ui/messages/en.json:1882-1890` and `ui/messages/de.json:1882-1890` — parity holds. The toggle on/off labels reuse the existing `settings_usage_hold_on`/`_off` keys (present at `en.json:1596-1597`, `de.json:1596-1597`).
- `Settings` type (`ui/src/lib/types.ts:75-79`), GET payload (`src/server.ts:2946-2948`), PUT patches (`src/server.ts:2983-2984`, `3116-3133`), api helpers (`ui/src/lib/api.ts:178-184`) and the persisted-override load (`src/index.ts:177-181`, `"1"`/`"0"` round-trip with `server.ts`) all agree on the `tuiFullscreen`/`tuiDisableMouse` key spelling and boolean/string forms.
- `feature-announcements.ts` `sinceVersion: "1.37.0"` is consistent with `package.json` (`1.36.0`) and the other recent 1.37.0 entries.

## Findings (non-blocking)

1. **`ui/src/lib/components/Settings.svelte`**: the "Disable mouse capture" toggle is bound only to `tuiDisableMouse`, so it still renders "Off" when `tuiFullscreen` is on even though the spawn wiring forces `CLAUDE_CODE_DISABLE_MOUSE=1` in that case (`src/service.ts:861`). The fullscreen hint text does mention it, but the toggle state visibly contradicts the effective behavior — consider disabling/forcing-on the mouse toggle when fullscreen is enabled to avoid operator confusion.
2. **`src/blocked.ts`**: the relaxed no-space acceptance opens a residual false-positive surface for no-space, multi-token, non-digit-leading numbered prose (e.g. two contiguous lines `1.First item` / `2.Second item`) within the 15-line tail. This is the same category the classic spaced pattern already admits and the issue explicitly accepted the contiguous-run guard as imperfect, so it is not a regression — flagging only so the tradeoff is a conscious one (the decimal/bare-token guards added here are good).

- Claude Code's actual runtime precedence of `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN` over `CLAUDE_CODE_NO_FLICKER` (and the fullscreen mouse-capture modes 1000/1002/1003) is external to this repo; I relied on the behavior asserted in the in-code comments rather than a third-party schema.

PR #1057, opened 2026-06-24
fix(critic): don't persist a post-merge error verdict as REVIEW ERR

scoop:
## Problem

After a PR merges, a session **occasionally flips to REVIEW ERR** even though it hasn't been decommissioned yet.

**Cause — a race:** the critic spawn and the PR merge both fire on CI-green, so they race by construction. The critic can finish *after* the PR is already merged. When it finishes with a transient **error** verdict (timeout / unparseable), `finalize()` handled it differently from a real verdict:

- The **real-verdict** path routes through `publishVerdict()`, which rechecks the *live* PR state (`forge.prStatus`) and bails when the PR is `merged`/`closed`.
- The **error** path had **no such recheck** — it recorded escalation and then `finalize()` unconditionally `putReview()` + `onChange()`'d the verdict, flipping the not-yet-archived session to REVIEW ERR until `forget()`/`dropReview()` finally ran at archive.

## Fix

Give the error path the same live-state recheck (extracted as `finalizeErrorVerdict()` to keep `finalize` under the complexity gate):

- **Confirmed `merged`/`closed`** → the error is moot: don't persist it as session review state and don't escalate. `captureUsage` + the worktree/terminal reap still run.
- **Unconfirmable** (no forge / `prStatus` threw → `undefined`) → **fail open**: keep the existing persist + escalate, so a genuine critic error on a still-open PR is never hidden by a failed recheck.

Real-verdict persistence is unchanged (it's deliberate, locked by `review.test.ts` "verdict still persisted"). Suppression is post-merge only, where the `error`-decision consumers (auto-merge / autopilot gates) are already moot; archive/decommission (`merge-teardown.ts`) never reads the verdict, so nothing downstream regresses.

### Scope & tradeoffs (deliberate)
- **Fixes "must not flip once the merge is observable"** — the recheck is best-effort, matching the same residual race `publishVerdict()` already accepts: an error finalizing microseconds before the merge becomes observable still persists.
- **Fail-open accepts a residual occurrence** of the exact symptom (REVIEW ERR if the PR merged but `prStatus` couldn't confirm it). Chosen over fail-closed, which would silently drop real critic errors + stall escalation on still-open PRs whenever a forge fetch hiccups. The residual self-clears at archive.
- **Observability:** both branches `console.warn` (suppressed-post-merge vs kept-fail-open) so the field cause of any reported flip is distinguishable.
- Out of scope: clearing a *pre-merge* error verdict at merge time (a legit error that then merges stays visible until archive).

## Tests

Two regression tests in `test/review.test.ts`, seeded at the cap boundary (`errorRound: 2`, cap 3) so the no-stall assertion is **discriminating** — they fail on `main` (2→3 crosses the cap and emits a stall signal) and pass with the fix (verified by stashing the source):
- PR merged before finalize (error verdict) → not persisted, no escalation.
- PR closed before finalize (error verdict) → also suppressed.

## Verification
- `bun test ./test` — 4669 pass, 0 fail
- `bun run lint`, `bunx tsc --noEmit` — clean
- `fallow audit --base origin/main --fail-on-issues` — exit 0
- pre-push gate — passed

Server-only; no UI change. `[no-feature-entry]` — internal critic-lifecycle fix, ships no user-facing UX.

### DIFF
--- a/src/review.ts
+++ b/src/review.ts
@@ -640,36 +640,13 @@ export class ReviewService {
     // (a forge/store/steer failure must not strand them).
     try {
       const verdict = this.buildVerdict(f, raw);
+      // A verdict for a PR that's no longer open is moot. The real branch handles that inside
+      // publishVerdict(); the error branch needs its own gate (finalizeErrorVerdict) since it
+      // never reaches publishVerdict. When persist is false we skip persisting the verdict as
+      // session review state — see below.
+      let persist = true;
       if (verdict.decision === "error") {
-        // A transient critic failure (timeout / unparseable verdict) posts nothing and
-        // has no findings to steer. Don't let it pose as "clean": count it on a separate
-        // no-progress streak so a flapping critic still escalates instead of looping
-        // forever, and preserve the findings round (those findings are still outstanding,
-        // just un-reverified this push).
-        verdict.errorRound = f.priorErrorRound + 1;
-        verdict.addressRound = f.priorRound;
-        // Error verdicts deliberately do NOT increment streakReviews: error-spawn token cost
-        // is bounded by the separate errorRound counter + its own cap escalation, not by the
-        // spawn ceiling. Preserve the streak's review count + reviewed-patch-id set so the
-        // ceiling math and churn dedup stay correct across a transient failure (and the errored
-        // patch-id is NOT added — mirrors patchId:"" so the same diff re-reviews).
-        verdict.streakReviews = f.priorStreakReviews;
-        verdict.reviewedPatchIds = f.priorReviewedPatchIds;
-        // The critic never produced a verdict, so it didn't actually consider this run's
-        // freshly-fetched notes — roll the seen set back so they re-inject next round
-        // instead of being silently swallowed by an error pass.
-        verdict.seenNoteIds = f.priorSeenNoteIds;
-        // Escalate once, when the streak first reaches the cap. `>=` with a crossing guard
-        // (not `=== cap`) so a cap lowered between runs still fires rather than being
-        // stepped over, while errors past the cap don't re-signal every tick.
-        if (verdict.errorRound >= this.cap && f.priorErrorRound < this.cap) {
-          this.deps.store.addSignal({
-            repoPath: f.repoPath,
-            sessionId: f.sessionId,
-            kind: "stall",
-            payload: `critic produced ${verdict.errorRound} consecutive error verdicts for this PR — auto-address can't make progress`,
-          });
-        }
+        persist = await this.finalizeErrorVerdict(f, verdict);
       } else {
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

The PR fixes the reported bug: a transient critic *error* verdict that finalizes after the PR has merged/closed no longer flips a not-yet-decommissioned session to REVIEW ERR. It factors the error-verdict bookkeeping out of `finalize()` into `finalizeErrorVerdict()`, adds a `livePrState()` recheck, and gates persistence (`persist`) on the result.

## Verification

- **Suppression path**: `finalizeErrorVerdict` returns `false` on `live === "merged" || live === "closed"`, and the caller skips `putReview`/`onChange` when `persist === false`, so the moot error verdict never becomes session review state. Verified against `src/review.ts:678-684` and `src/review.ts:717-758`. `captureUsage` and the `finally` reap still run (`src/review.ts:690-701`), so token cost is attributed and the worktree/terminal are reaped — matches the comment's claim.
- **State type**: `PrStatus["state"]` is `"none" | "open" | "merged" | "closed"` (verified `src/forge/types.ts:134`); the guard plus the `undefined` fail-open branch (`src/review.ts:725-728`) cover the cases.
- **Branch key**: `livePrState` calls `forge.prStatus(f.branch)`; `f.branch` exists on `InFlight` (`src/review.ts:68`) and is the same key `publishVerdict` already uses for real verdicts (`src/review.ts:795`), so the recheck mirrors the existing fetch.
- **Merged detection with deleted branch**: `gh pr list --head <branch> --state all` retains merged PR records (`src/forge/github.ts:501-534`), so a merged PR maps to `"merged"` not `"none"` (`src/forge/github.ts:535-538`). The merged case is detected even after branch deletion — no new gap.
- **Fail-open vs fail-closed divergence**: `publishVerdict` returns silently for an unconfirmable/`!== "open"` state (`src/review.ts:799,815`), while `finalizeErrorVerdict` keeps+persists on `undefined` (`src/review.ts:725-757`). This is documented as intentional (a genuine error on a still-open PR must not be hidden by a failed recheck) and is the safe choice.
- **Behavioral equivalence**: the moved error-bookkeeping (errorRound bump, `streakReviews`/`reviewedPatchIds`/`seenNoteIds` preservation, the `>= cap` crossing-guarded stall signal) is identical to the removed inline block; only the merged/closed early-return is new.

### Minor observation (non-blocking, not a finding)

`live === "none"` (no PR found at all) falls through to the keep/persist branch rather than being suppressed. This is not the bug scenario — a merged PR reports `"merged"`, and `"none"` only arises when the forge genuinely finds no PR for the branch (effectively unconfirmable), where fail-open is consistent with the stated design. Noted only for completeness.

PR #1058, opened 2026-06-24
fix(usage): re-capture extra-credit panel on /usage scrape

scoop:
## Problem

The **Usage credits** (extra-credit) gauge in the top bar stayed perpetually **stale** and clicking **REFRESH** never cleared it — while the session/week meters refreshed fine and **no error** was shown. Confirmed with the user: only credits stuck, no red "Retry".

## Root cause

The `/usage` scrape wait loop (`src/usage-probe.ts`) gated **only on the weekly window** and returned the buffer the instant week's "Resets …" line appeared. But the **"Usage credits" panel renders *below* the week gauge** in the TUI, so it streams into the buffer a render-cycle *later* and was missed:

`parseCredits → null → persistCredit no-op → credit snapshot's scrapedAt never advances → stale`

Because week *was* captured (`scraped === true`), the server returned HTTP 200 with no error — it just kept reading stale.

## Phase-0 spike (go/no-go, run before coding)

Two rival hypotheses produce the identical symptom; live evidence picked the winner:

- **H2 (parse/locale) ruled out** — the live snapshot has `spent: €79.16` (dot-decimal), which `parseCredits` handled fine. Not a comma-decimal format bug.
- **"Needs keypress" ruled out** — a snapshot was captured non-interactively before → credits renders on the default `/usage` screen.
- **H1 (render-timing) confirmed** — a live `POST /api/usage/refresh` returned **200 in 11s** with caps (`calibratedAt`) advancing, yet the credit snapshot stayed frozen **~2.8 days** old. Week captured, credits missed on a successful scrape — exactly the timing race.
- **Idle timeout ruled out as the cause** — the 11s refresh returned 200 (no reset), so Bun's 10s idle timeout is not firing here today.

> The standalone instrumented probe could not cleanly reproduce the timing locally (the spawned `claude` hit a folder-trust prompt that swallowed the `/usage` send — a known spawn gotcha; the long-running daemon is pre-trusted). The go decision rests on the live-daemon evidence above, which reflects real repeated behavior rather than one synthetic scrape.

## Fix

- **`src/usage-probe.ts`** — extract the wait into a pure, unit-testable `awaitUsageFrame(read, sleep, weekTries, creditTries)`. After the week reset-label wait, add an **unconditional bounded grace** that waits for the credits panel, returning the instant it parses. A true no-credit account simply waits the bounded grace and falls through to the same buffer (nothing fabricated); a failed scrape (no week) skips the grace and returns null. Unconditional (rather than gated on a "has-credits" signal) keeps it to one file, **fixes cold-start**, and is equivalent for no-credit accounts.
- **`src/server.ts`** — lift Bun's 10s idle timeout to 60s for just the refresh request (`server.timeout(req, 60)` via an extracted `isSlowScrapeRequest` predicate). **Insurance only** for the now-longer worst-case scrape — *not* the diagnosed cause (the predicate is extracted so the `fetch` handler stays under the fallow cognitive-complexity threshold).

## Tests

`test/usage-probe.test.ts` — 4 new cases for `awaitUsageFrame`, using a fake `read()` that **evolves with the fake `sleep`'s step counter** so the grace loop is genuinely exercised: credits-appears-late (captured, grace ran), already-present (no added latency), no-credit (bounded grace, no hang), nothing-renders (null, grace skipped).

## Verification

- `bun run lint` ✓ · `bunx tsc --noEmit` ✓ · `bun test ./test` ✓ (4657 pass, 0 fail) · `bunx fallow audit --base origin/main` ✓ (no issues in changed files)
- Branch cut from latest `origin/main`, linear.

No UI/i18n/feature-catalog/glossary/design-system impact (`fix`, server-side only).

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -4898,6 +4898,11 @@ function livePtyTerminalId(
   }
 }
 
+/** The /usage refresh handler drives a multi-second ephemeral `claude` scrape and needs a longer
+ *  idle window than Bun's 10s default (see the fetch handler). */
+const isSlowScrapeRequest = (req: Request, url: URL): boolean =>
+  req.method === "POST" && url.pathname === "/api/usage/refresh";
+
 export function serve(deps: AppDeps, port: number) {
   const app = makeApp(deps);
   // current owning socket per terminal — a single owner avoids the takeover war
@@ -4957,6 +4962,10 @@ export function serve(deps: AppDeps, port: number) {
           ? undefined
           : new Response("upgrade failed", { status: 500 });
       }
+      // /usage refresh drives an ephemeral claude scrape (~24s worst case: boot + week wait +
+      // credits grace); lift Bun's 10s idle timeout for just this request so a slow scrape can't
+      // reset the connection into a false "Retry" on the gauge. Insurance — other endpoints unchanged.
+      if (isSlowScrapeRequest(req, url)) server.timeout(req, 60);
       return app.fetch(req);
     },
     websocket: {
--- a/src/usage-probe.ts
+++ b/src/usage-probe.ts
@@ -29,6 +29,14 @@ export const PROBE_NAME = "__usage_probe__";
  *     never advances and the gauge reads perpetually stale. Give credits a bounded grace (up to
  *     `creditTries`), returning the instant it parses. A true no-credit account simply waits the
  *     bounded grace and falls through to the same buffer — nothing is fabricated.
+ *
+ * The grace is intentionally a fixed bound rather than gated on a "has-credits" signal or an
+ * early-bail when the buffer stops growing: those save ~`creditTries`s only on no-credit accounts,
+ * whose grace is paid **solely by the background calibrate** (the manual REFRESH control lives in
+ * CreditDetail, rendered only when `credits != null`, so no user ever waits on it). An early-bail
+ * would also risk missing credits if the TUI pauses streaming between the week gauge and the panel
+ * below it — trading away the very reliability this wait exists to provide. Background-only latency
+ * isn't worth that.
  */
 export async function awaitUsageFrame(
   read: () => string,
--- a/src/usage-probe.ts
+++ b/src/usage-probe.ts
@@ -4,7 +4,7 @@ import { config } from "./config";
 import { compileCacheDir } from "./tmp-sweep";
 import { isApiKeyMode } from "./spawn-auth";
 
-const sleep = (ms: number) => new Promise((r) => setTimeout(r, ms));
+const sleep = (ms: number): Promise<void> => new Promise((r) => setTimeout(r, ms));
 const decoder = new TextDecoder();
 
 /**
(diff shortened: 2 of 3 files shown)


scoop (commented):
## Summary

The PR fixes two distinct halves of the reported bug, and both are sound.

### 1. Stale extra-credit gauge — `awaitUsageFrame` credits grace
`src/usage-probe.ts:33-45` extracts the wait loop into an injectable, unit-testable `awaitUsageFrame(read, sleep, weekTries=12, creditTries=6)`.

- The week-label gate is preserved verbatim from the old inline code (`weekTries=12`, gate on `frame().week?.resetLabel`), and the return value `frame().week ? read() : null` is equivalent to the old `week() ? buf : null`. Verified against the removed lines in the diff.
- The new credits grace loop (`src/usage-probe.ts:43`) is the actual fix. I confirmed the root cause downstream: `persistCredit` (verified against `src/usage-limits.ts:367-378`) only writes a snapshot when `parsed.credits` is truthy; before this change the buffer was returned the instant week's label rendered, before the "Usage credits" panel (which renders BELOW week) had streamed in — so `persistCredit` was a no-op and `limits()` (`src/usage-limits.ts:397-412`) kept serving the prior, stale credit snapshot. Waiting a bounded grace for `frame().credits` is exactly what populates the buffer so the snapshot advances.
- The `frame().week`-falsy short-circuit on the grace loop (`src/usage-probe.ts:43`) correctly avoids burning 6 extra seconds on a totally failed scrape, and `null` is still returned when week never renders.

Test fixtures were checked against the real parser: `WEEK_ONLY` yields `week.resetLabel`="Jun7" and `credits=null` (no `Usagecredits` token, `src/usage-limits.ts:163-164`); `WEEK_PLUS_CREDITS`'s `€79.16/€100.00 spent` matches the spend regex at `src/usage-limits.ts:174`. The four test cases' sleep-count and credits expectations all follow from the loop logic.

### 2. Refresh "doesn't work" — per-request idle timeout
`src/server.ts:4904` (`isSlowScrapeRequest`) gates exactly `POST /api/usage/refresh`, which matches the handler routing at `src/server.ts:2502-2507`. `server.timeout(req, 60)` (`src/server.ts:4968`) is a valid Bun `Server` method (`@types/bun` per `package.json:27`), the arg is within Bun's 255s cap, and 60s comfortably covers the ~24s worst-case scrape (5s boot + 0.9s + 12s week + 6s credits). Scoping the lift to just this request leaves all other endpoints on Bun's default — a clean, low-blast-radius change.

## Nit (non-blocking)
The credits grace is unconditional whenever the week label is present, so every refresh for a subscription account that has **no** extra-credits panel (pay-as-you-go is opt-in, so this is a common case) now pays the full `creditTries` grace (~6s) before falling through. This is an intentional, documented tradeoff (there's no buffer-only way to distinguish "no panel" from "panel not yet rendered"), so it's acceptable, but worth being aware of as added latency on the common path.

No correctness, security, or parity issues found. No user-facing strings or message catalogs are touched, so i18n parity is not in play.

scoop:
**Re: ~6s credits grace on the no-credits common path** — kept as a fixed bound; the latency is never user-facing.

The only caller of `refreshUsage()` is the REFRESH button in `CreditDetail.svelte`, whose entire content is gated `{#if credits}`. So a no-credits account has **no refresh button at all** — the ~6s grace for those accounts is paid *exclusively* by the **background calibrate** (daily, or 15-min near the cap), where no user waits on a spinner.

Shortening it would only trim a background job's runtime while cutting the reliability margin this PR exists to provide. I also considered an early-bail when the buffer stops growing, but that risks missing credits if the TUI pauses streaming between the week gauge and the panel rendered below it — reintroducing the exact race being fixed. Net: background-only latency isn't worth trading reliability for.

Recorded this reasoning at the code site in `awaitUsageFrame`'s docstring (985473fa).

scoop (commented):
## Re-review

The prior revision raised one non-blocking nit; the file (`src/usage-probe.ts`) is in this diff, so it was in scope to re-check.

**Prior finding #1 — ~6s credits grace on the no-credits common path (nit):** Addressed. The nit asked to *consider* whether the latency is acceptable or could be shortened. The diff now records an explicit decision in `awaitUsageFrame`'s docstring (`src/usage-probe.ts:18-37`): keep it as a fixed bound. I verified the load-bearing claim behind that decision rather than taking the note's word for it — the entire refresh UI, including the REFRESH button (`onclick={onRefresh}`), is gated behind `{#if credits}` (verified at `ui/src/lib/components/top-bar/CreditDetail.svelte:31-66`). A no-credits account therefore has no manual refresh control, so the grace on that path is paid solely by the background calibrate, where no user waits on a spinner. The early-bail alternative would risk re-introducing the streaming race this PR fixes. Reasoning is sound; not re-raised.

## Verification

- **Behavior preserved by the refactor:** original `for (i<12 && !week()?.resetLabel) …; return week() ? buf : null` maps cleanly onto the extracted `awaitUsageFrame` (`src/usage-probe.ts:38-44`) — `weekTries=12` matches the original `12`, and `parseUsageFrame(read(),0)` mirrors the original `parseUsageFrame(buf,0)`. Caller at `src/usage-probe.ts:147` passes the live buffer and module `sleep`.
- **`frame().credits` is a trustworthy gate:** `parseCredits` (`src/usage-limits.ts:162-185`) returns null unless both the `Usagecredits` anchor and the spend pattern match, so the new credits-grace loop can't spin on a false positive.
- **Server timeout:** `isSlowScrapeRequest` (`src/server.ts:4902-4904`) matches POST `/api/usage/refresh`, which is exactly the route served at `src/server.ts:2502-2507`. `server.timeout(req, 60)` (`src/server.ts:4968`) uses Bun's per-request idle-timeout API with `server` taken from the fetch handler's 2nd arg (`src/server.ts:4917`). Worst case ≈ 5 (boot) + 0.9 + 12 (week) + 6 (credits) ≈ 24s, comfortably under the 60s lift.
- **Tests:** import `awaitUsageFrame` from `../src/usage-probe` and `parseUsageFrame` from `../src/usage-limits` (both valid exports); `WEEK_PLUS_CREDITS` satisfies the `parseCredits` spend regex, so the grace-loop assertions exercise the real gate.

No new correctness, security, or quality issues found in the diff.

PR #1064, opened 2026-06-24
feat(autopilot): rebase idle review-passed non-mergeable PR (non-full-auto)

scoop:
## Problem

A session in **autopilot but NOT full-auto** (auto-merge off) stands down at PR-open (`autopilot.ts` `eligible()`: `hasPr && !fullAuto → null`), and the merge train never carries it (it filters to full-auto sessions only). So when such a session's PR has **passed review**, the **agent goes idle**, but the **PR is non-mergeable** (behind base or conflicting), nobody steers a rebase — it sits idle forever. Full-auto sessions already get rebase-on-behind via the merge train; this closes the gap for the non-full-auto case.

## What this does

Adds `reEngageRebase` to `AutopilotService` — a sibling of the existing `reEngageCi`, fired from `onDone` and the 30s `tick`. For a non-full-auto, idle, **open + non-draft + green** PR with a **clean head-matched critic sign-off** that is **behind/conflicting**, it steers the agent to rebase onto its base. After `rebaseCap` (`config.autoMergeRebaseCap`, default 5) attempts it **pauses to the operator**. It **stops at mergeable** — it never merges (auto-merge is off; a human merges).

## Design notes / deliberate deviations (flagged)

- **Review gate is intentionally STRICTER than `automerge-core.needsRebase`.** `needsRebase` rebases even with `reviewDecision === null` (no verdict yet); this requires a clean head-matched critic verdict via the reused, tested `signedOff("critic", …)` predicate. Two reasons: (a) the trigger is literally a review that *passed*, and (b) requiring `findings.length === 0` guarantees `review.ts`'s auto-address steer loop (which only fires on findings) is **not** also driving the same idle session — no double-steer. Critic-disabled repos rebase on green CI alone (which *does* match `needsRebase`).
- **Draft PRs are excluded.** A rebase alone can't make a draft mergeable (it must be marked ready-for-review first), and a draft reports `mergeStateStatus = DRAFT`, which masks `BEHIND` anyway.
- **"Behind" is read from the cached `mergeStateStatus`, not a git fetch.** The merge train uses `worktree.behindBase()` (a `git fetch` + `merge-base` check); this path uses GitHub's cached `mergeStateStatus` to avoid a fetch on autopilot's tick. Trade-off: forges that don't supply `mergeStateStatus` (Gitea/local) detect only the conflict signal (`mergeable === false`), not "behind".
[...]

PR #1067, opened 2026-06-24
fix(review): zero-context patch-id so a clean rebase skips critic re-review

scoop:
## Problem

The autopilot now rebases an idle, review-passed, non-mergeable PR (#1064), but after the rebase the critic re-reviews even though the branch's own changes didn't change — the operator-observed symptom behind this task.

**Root cause (empirically confirmed):** the rebase-skip fingerprint in `defaultComputePatchId` hashed `git diff <base>...HEAD` at git's **default 3-line context**, and `git patch-id` hashes context lines too. A clean rebase that pulls in a base change *within a hunk's context window* flips the patch-id even though the branch's own added/removed lines are byte-identical → `shouldSkipForPatchId` misses → needless re-review. Existing tests only advanced the base in *unrelated* files, so they never exercised context drift.

Spike (single-hunk): default-context id `be3b82…`→`a0fe1a…` (flips, bug) vs `-U0` `1571f9…`→`1571f9…` (stable); a real edit → `e19327…` (still distinguished).

## Fix

Fingerprint at **`-U0` (zero context)** so the id keys only off the branch's own +/- lines. One change to the *shared* fingerprint covers **every** rebase path — autopilot `reEngageRebase`, the full-auto merge train, manual force-pushes, and the repo-level critic — matching the agreed "any agent rebase/force-push" scope. No path-specific code.

The critic's own review diff (full context), the changed-file set, and the fresh-fetch + three-dot merge-base logic are untouched — `-U0` affects only the skip fingerprint, not what the critic reads.

## Rebuts the original design rationale

The design spec (`docs/superpowers/specs/2026-06-03-skip-critic-rebase-design.md`) intentionally hashed context to catch conflict-resolution edits. `-U0` **still** catches those: a conflict resolution edits the branch's *own* +/- lines, so the id flips and the critic re-reviews. The only behavior removed is re-triggering on *pure base-only context drift* on a clean rebase — never the documented intent, just an incidental side effect of hashing context. Spec revised accordingly.

## Deliberate tradeoffs (noted in code + spec)

- **Wider skip window (intended):** base-only changes no longer re-trigger review even within 3 lines of a hunk — matches the merge-train philosophy that a clean rebase onto main is trusted.
- **Marginally wider collision surface:** dropping context means identical +/- line text relocated within a file can share an id → a rare false-skip. Per-file diff headers keep cross-file changes distinct; acceptable for a skip-vs-review decision that already errs toward reviewing on any failure.
- **One-time post-deploy re-review (self-healing):** in-flight sessions hold an old default-context id, so the first head after deploy re-reviews once, then the `-U0` id is stored and subsequent rebases match.

## Confidence

The spikes prove the *mechanism* (default-context patch-id flips on a clean rebase via context drift; `-U0` is stable, single- and multi-hunk). They do not replay the *specific* observed branch's history, but context drift is the only known way a clean rebase changes the current fingerprint — the strong leading explanation, stated honestly.

## Tests

- **Real-origin FETCH_HEAD context-drift invariance** (the production fetch path): red before the fix, green after.
- **Multi-hunk stability**: two hunks, base change between them, clean rebase — guards `patch-id --stable` invariance to `-U0` hunk grouping.
- Existing merge-train + real-edit + offline-fallback patch-id tests still pass.

## Verification

`bun run lint` ✓ · `bunx tsc --noEmit` ✓ · `bun test ./test` (4741 pass, 0 fail) ✓ · `fallow@2.100.0 audit` no issues in changed files ✓

[no-feature-entry] server-internal bugfix, no UI surface.

### DIFF
--- a/docs/superpowers/specs/2026-06-03-skip-critic-rebase-design.md
+++ b/docs/superpowers/specs/2026-06-03-skip-critic-rebase-design.md
@@ -28,15 +28,36 @@ across a pure rebase and changes only when the branch's actual changes change.
 
 ## Fingerprint: `git patch-id`
 
-Compute `git patch-id --stable` over `git diff <base>...HEAD`, take the first
-token (the patch id). Rationale:
+Compute `git patch-id --stable` over `git diff -U0 <base>...HEAD` (**zero
+context**), take the first token (the patch id). Rationale:
 
 - patch-id deliberately **ignores line numbers**, so a rebase that only shifts
   the branch's hunks (because new `main` edited code elsewhere in the same
   files) still produces the same id.
-- It **does** hash the changed lines and their immediate context, so it differs
-  when there are new commits, or when conflict resolution during the rebase
-  altered the branch's content. Both genuinely warrant a fresh review.
+- The `-U0` diff hashes **only the branch's own added/removed lines**, not the
+  surrounding (base-owned) context. So the id changes when there are new commits,
+  or when conflict resolution during the rebase altered the branch's content —
+  both edit the branch's own +/- lines, so both still warrant (and get) a fresh
+  review. The id does **not** change on pure *base-only context drift*: a clean
+  rebase where the branch's own lines are byte-identical and only a nearby
+  base-owned line moved. That is the intended behavior — review keys off the
+  branch's change, not the base's.
+
+  > **Why not default 3-line context (the original design).** This section
+  > originally hashed "the changed lines **and their immediate context**" to
+  > catch conflict resolution. That was over-broad: hashing context also made a
+  > line moved by the base *within a hunk's context window* flip the id on an
+  > otherwise clean rebase, re-triggering a needless review (operator-observed).
+  > `-U0` still catches conflict resolution (it edits the branch's own +/- lines),
+  > so the original concern is preserved; only the incidental context-drift
+  > re-trigger is removed.
+  >
+  > **Tradeoff.** Dropping context marginally widens the collision surface: two
+  > distinct revisions whose +/- line text is identical but which sit in
+  > different surrounding code can now share an id (a rare false-skip). The
+  > per-file diff headers keep cross-file changes distinct; the residual risk is
+  > same-file, identical +/- text in a relocated position — judged acceptable for
+  > a skip-vs-review decision that already errs toward reviewing on any failure.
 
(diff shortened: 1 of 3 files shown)


scoop (commented):
## Summary

The PR switches the patch-id fingerprint diff from full (3-line) context to zero context (`git diff -U0 <base>...HEAD`) so a clean rebase that only shifts base-owned context lines no longer flips the id and re-triggers a needless review. This directly serves the task (skip auto review after a no-op rebase).

## Verification

- **Only functional change is `src/critic-core.ts:245`** (`git diff` → `git diff -U0`). The resulting `diff` string is consumed in exactly two places: the `!diff.length` guard (`src/critic-core.ts:251`) and as stdin to `git patch-id --stable` (`src/critic-core.ts:280`). No other consumer depends on the diff carrying context, so narrowing it is safe.
- **Changed-file set is unaffected**: it is computed separately via `git diff --name-only -z` (`src/critic-core.ts:262`), which takes no context, so the backstop scope is unchanged.
- **Identifier resolution**: `defaultComputePatchId` is defined at `src/critic-core.ts:193` and re-exported from `src/review.ts:30`, matching the test import at `test/review-patchid.test.ts:6`.
- **`patch-id --stable` over a `-U0` diff is stable across rebase**: patch-id ignores the `@@` line numbers and hashes only the `+`/`-` body lines; `--stable` is split/order-independent. Both compared diffs are produced identically, so base-only context drift drops out of the hash while conflict resolution (which edits the branch's own +/- lines) still changes it — exactly the intended semantics. The two new tests (`test/review-patchid.test.ts:223` context-drift invariance over the real FETCH_HEAD path, and `:268` split-hunk invariance) guard this concretely.
- **`!diff.length` guard still valid**: any real content change still emits hunk lines under `-U0`, so the no-diff short-circuit is unchanged.
- **Docs/comments consistent**: `docs/superpowers/specs/2026-06-03-skip-critic-rebase-design.md` and the docstring/comment at `src/critic-core.ts:178-187,241-243` accurately describe the `-U0` behavior and the acknowledged collision-surface tradeoff.

No correctness, security, or quality problems found. The collision-surface widening (same-file, identical +/- text relocated) is explicitly documented and acceptable for a skip-vs-review decision that errs toward reviewing on any failure.

PR #1072, opened 2026-06-24
fix(usage): read the real credits panel past the /usage-credits menu text

kai-osthoff:
## Symptom

The extra-usage **CR gauge** silently vanished from the usage popover — only the **5h** and **Weekly** bars rendered — even with the weekly window at **100%** and paid extra usage actively spending (€68/€100). Reported live:

> hier fehlen die extra-usage credits die man zubuchen kann

## Diagnosis (empirical)

- `usage_caps` (5h + week) populated; the `usage_credit` snapshot table **never written** → a credits-only capture miss, not a display bug. The UI correctly hides the gauge when `credits === null`.
- Captured a live `/usage` scrape through the **real probe path** (herdr + `pty-attach`, classic renderer at 120×40). The buffer **does** contain the panel (`Usagecredits … €68.67/€100.00 spent`), yet `parseUsageFrame(...).credits` returned **null**.

## Root cause

The probe runs claude in the **classic renderer** (`CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN`), which accumulates the `/usage` **slash-command menu** into the scrape buffer. The `/usage-credits` command — renamed from `/extra-usage` in Claude Code **v2.1.144** — describes itself as *"Configure usage credits to keep working when you hit a limit"*, which collapses (whitespace-stripped) to a `Usagecredits…` run **with no `…/…spent` figure**, sitting **before** the actual panel.

`parseCredits` anchored on the **first** `Usagecredits` match (the menu text), sliced to the next `Esc to cancel`, found no spend line, and returned `null` — never reaching the real panel below. Session/week parse fine because their anchors (`Currentsession|Currentweek`) never appear in the menu.

This is a regression triggered by the upstream `/extra-usage` → `/usage-credits` rename: before it, typing `/usage` surfaced no menu entry containing the words "usage credits", so the buffer wasn't poisoned.

## Fix

`parseCredits` now scans **every** `Usagecredits` occurrence and returns the first one carrying a real `…/…spent` amount, skipping menu text and still-`Refreshing…` partial frames. Minimal, localized; the single-occurrence (alt-screen) path is unchanged.

## Verification

- Ran the **fixed** parser against the live classic-renderer probe capture → now extracts `{ spent: 68.67, cap: 100, currency: "€", pct: 68, resetLabel: "Jul1" }` where it previously returned `null`.
- New regression test feeds a menu-poisoned collapsed buffer through `parseCredits` **and** `parseUsageFrame` end to end.
- `bun test test/usage-limits.test.ts` → 46 pass · `test/usage-probe.test.ts` → 6 pass · `bun run lint` clean · pre-push gates green.

No UI/i18n/feature-catalog changes (server-side parser fix; the CR gauge feature already ships).

### DIFF
--- a/src/usage-limits.ts
+++ b/src/usage-limits.ts
@@ -158,30 +158,40 @@ function parseSegment(seg: string, now: number): ScrapedWindow | null {
  *   Usagecredits▎0%used€0.29/€50.00spent·ResetsJun1(Europe/Berlin)
  * The gauge pct rounds down (can read `0%used` while real money is spent), so `spent` is the truth
  * signal and is parsed independently of pct.
+ *
+ * Scan EVERY "Usagecredits" occurrence, not just the first. The probe runs claude in the classic
+ * renderer (CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN), which accumulates the `/usage` slash-command menu
+ * into the buffer — and the `/usage-credits` command (renamed from `/extra-usage` in Claude Code
+ * v2.1.144) carries the description "Configure usage credits to keep working when you hit a limit",
+ * which collapses to a `Usagecredits…` run that precedes the real panel. The first match is thus
+ * that menu text (no spend figure); the actual panel — the only occurrence carrying a `…/…spent`
+ * amount — comes later. So match on the spend line as the truth signal and skip any occurrence
+ * (menu text, or a still-"Refreshing…" partial frame) that lacks it.
  */
 export function parseCredits(collapsed: string, now: number): ScrapedCredit | null {
-  const start = collapsed.search(/Usagecredits/i);
-  if (start < 0) return null;
-  let seg = collapsed.slice(start);
-  const end = seg.search(/Esctocancel/i);
-  if (end >= 0) seg = seg.slice(0, end);
-  // pct rounds down and can be absent; treat missing as 0 — spend below is the real signal
-  const pct = +(seg.match(/(\d+)%used/i)?.[1] ?? 0);
-  // symbol-only currency group: a `(\D*)([\d.]+)` form would wrongly grab the `%used` of `0%used€`.
-  // Amounts are assumed dot-decimal (as the Europe/Berlin TUI renders, e.g. `€0.29`); a comma render
-  // (`€0,29`) wouldn't match `[\d.]+` and would drop the whole credits section (returns null).
-  // `seg` is already whitespace-collapsed, so there is no inter-group whitespace to skip.
-  const spend = seg.match(/([^\d.\s/])([\d.]+)\/[^\d.\s/]?([\d.]+)spent/i);
-  if (!spend) return null;
-  const label = seg.match(/Resets(.*?)\(/i)?.[1] ?? null;
-  return {
-    pct,
-    spent: +spend[2]!,
-    cap: +spend[3]!,
-    currency: spend[1]!,
-    resetLabel: label,
-    resetAt: label ? parseMonthlyReset(label, now) : null,
-  };
+  for (const m of collapsed.matchAll(/Usagecredits/gi)) {
+    let seg = collapsed.slice(m.index);
+    const end = seg.search(/Esctocancel/i);
+    if (end >= 0) seg = seg.slice(0, end);
+    // symbol-only currency group: a `(\D*)([\d.]+)` form would wrongly grab the `%used` of `0%used€`.
+    // Amounts are assumed dot-decimal (as the Europe/Berlin TUI renders, e.g. `€0.29`); a comma render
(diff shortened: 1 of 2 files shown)


kai-osthoff (commented):
## Summary

This PR fixes `parseCredits` so the live "Usage credits" panel is found even when the classic-renderer buffer is polluted by the `/usage` slash-command menu (the `/usage-credits` command, renamed from `/extra-usage`). Previously the parser anchored on the FIRST `Usagecredits` occurrence and, if that occurrence carried no `…/…spent` figure, returned `null` — so the credits gauge silently vanished. The new code scans every `Usagecredits` occurrence via `matchAll` and returns the first one that actually carries a spend figure, skipping menu text and still-`Refreshing…` partial frames. This satisfies the task ("hier fehlen die extra-usage credits die man zubuchen kann").

## Verification

- **Contract unchanged**: `parseCredits(collapsed, now): ScrapedCredit | null` (src/usage-limits.ts:171). Its only production caller is `parseUsageFrame` (src/usage-limits.ts:205), which is unaffected by the internal loop refactor — return shape is identical.
- **Menu-skip works as documented**: the menu description "Configure usage credits to keep working when you hit a limit" whitespace-collapses to `…usagecredits…` (the spaces between "usage" and "credits" disappear), which matches `/Usagecredits/i`. That occurrence's truncated segment has no `…/…spent` match, so `if (!spend) continue;` (src/usage-limits.ts:181) skips it and the real panel later in the buffer is read. The hyphenated command name `/usage-credits` itself retains its hyphen after collapse and would not match — the comment correctly attributes the spurious match to the description text.
- **No regression on existing tests**: the `$`/`£` single-panel cases (test/usage-limits.test.ts:139,144) each contain exactly one `Usagecredits` run carrying spend → returned on the first iteration; the absent-section case (test/usage-limits.test.ts:151) has no `Usagecredits` → `matchAll` yields nothing → `return null` (src/usage-limits.ts:194). All preserved.
- **New test is end-to-end** (test/usage-limits.test.ts:155): asserts both `parseCredits` directly and the `parseUsageFrame(...).credits` path against a menu-poisoned buffer; the expected `spent=12.34`, `cap=80`, `pct=65`, `resetLabel="Jul1"` all match the spend regex `([^\d.\s/])([\d.]+)\/[^\d.\s/]?([\d.]+)spent` and the `Resets(.*?)\(` label regex.
- **Order independence**: the loop returns the first spend-bearing match regardless of whether the menu precedes or follows the panel. Within one `/usage` snapshot the spend value is constant across redraws, so "first spend-bearing match" is value-equivalent to the windows' "last reading" convention — no staleness concern.

No blocking issues, no nits. The comma-decimal render (`€0,29`) still returns `null` for the credits section, but that is an explicitly documented pre-existing assumption carried over verbatim from the prior implementation, not introduced or worsened here.

scoop:
**shepherd-author-note — independent confirmation, recommend merge**

A separate session was tasked with the "extra-credit usage gauge stuck stale, refresh reverts to stale, no error" report. I diagnosed it from scratch against the live host and landed on the *exact same* root cause and fix as this PR, so this is independent corroboration rather than a competing change — I am **not** opening a parallel PR (single-PR invariant); deferring to #1072.

**Live evidence gathered:**
- `POST /api/usage/refresh` → HTTP 200 in ~16s. Afterward `usage_caps` (session5h + week) `scrapedAt` advanced to *now*, but `usage_credit.scrapedAt` stayed frozen ~3.5 days old → a **credits-only persist miss**, week/5h scrape fine.
- Reproduced the real herdr-attach probe path (classic renderer): `week` parsed, the literal text `Usagecredits` **was present** in the buffer, yet `parseCredits` returned **null**.
- Dumped the collapsed buffer — **3** `Usagecredits` substrings: #0/#1 are `/usage` slash-command **menu help text** (no `…/…spent`), #2 is the real panel (`Usagecredits…79%used €79.16/€100.00spent·ResetsJul1`). `search(/Usagecredits/i)` anchors on #0 → no spend → `null`; the `matchAll` + first-with-spend approach in this PR correctly picks #2.

**On the implementation choice:** the deliberate **first-valid** selection here (skip menu text and any later still-"Refreshing…" partial) is the right call — it locks onto the earliest complete panel and avoids a spend-regex match spanning a frame-concatenation boundary. The regression test faithfully reproduces the menu-poisoned buffer end-to-end.

**Recommendation: merge as-is.** CI is green and it's mergeable.

**Separate latent issue (NOT this PR):** during diagnosis I also hit the probe **wedging on claude's folder-trust prompt** when `~/.claude.json` has `hasTrustDialogAccepted:false` for `repoRoot` — `--dangerously-skip-permissions` does not suppress it, so the typed `/usage` is eaten and the whole scrape returns null (week included). Not the credits bug (live host's home dir is effectively trusted, week scrapes fine). Filing as a separate hardening follow-up.

PR #1078, opened 2026-06-25
fix(diff): diff against the PR's real base, not session.baseBranch

scoop:
## Problem

TASK-852 (in the `flowagent` repo) was explicitly based off a non-default branch — a hotfix targeting `main` while the repo default is `dev`. Shepherd's **diff tab listed far more files than the PR touched**, because the diff was computed against the session's stored `baseBranch` (`dev`), folding the entire `dev`↔`main` divergence on top of the task's own changes.

Root cause: both the diff endpoint (`sessionDiffRead`) and the recap diff (`recap.ts`) compared against `session.baseBranch`, which is set once at session creation and never reconciled to the PR's actual base. GitHub's "Files changed" uses three-dot/merge-base semantics against the PR's real base, so diffing against `dev` could never match.

## Fix (diff surfaces only)

Resolve the PR's **actual** base (`baseRefName`) and diff against that, with `session.baseBranch` as an always-safe fallback.

- **`baseRefName` on `PrStatus`/`GitState`** — populated by the GitHub forge as one extra `gh pr list --json` field (no extra call; confirmed returned for merged/closed PRs too). Added to `gitStateChanged` so a PR retarget refreshes the cache.
- **Shared `resolveDiffBase` helper** (`src/diff-base.ts`) returning `{ base, resolved }`: warm `prCache` → on-demand `prStatus` (cold/evicted cache) → `baseBranch`. The gh runner is async, so on-demand never blocks the event loop. `resolved` marks an authoritative resolution (a PR base or a definitive "no PR").
- **Live diff tab** (`sessionDiffRead`) now diffs against the resolved base.
- **Recap diff**: base is resolved **inside `generate()`** (covers operator-forced `regenerate()`, which bypasses dedup) and persisted on the recap row (empty + generating). Recap dedup becomes **`(headSha, base)`-keyed** so a recap baked before the PR base was known regenerates — guarded against legacy mass-regeneration (`base=""`) and transient-fallback thrash (only an authoritatively-resolved base re-fires).

## Scope notes (deliberate)

- **Diff-only**, per operator decision: `session.baseBranch` is **not** reconciled, and `hasCommittedChanges` / drain / merge / epic-base consumers are unchanged. The diff-only option explicitly accepted that `hasCommittedChanges` stays on `baseBranch` (behaviorally inert for a real hotfix).
- **TASK-852 itself is already archived** (worktree removed), so its recap is permanent and cannot be retroactively corrected; the fix prevents recurrence for future/active sessions and is proven by the deterministic tests.

## Tests

- Forge: `prStatus` surfaces `baseRefName`.
- `resolveDiffBase` unit: warm cache / definitive-no-PR / on-demand / transient-failure (`resolved:false`) / no-forge.
- Diff endpoint: prefers the cached PR base over the stored `baseBranch`.
- Recap: `generate()`/`regenerate()` diff + persist the resolved base; `considerForArchive` regenerates when the base becomes resolvable at the same HEAD, and **skips** on a transient (`resolved:false`) fallback (no thrash); legacy `base=""` never mass-regenerates.
- Store: `base` column round-trips + additive migration.

Full root suite green (4843 pass), `tsc` + `lint` clean. Server-only change — no UI strings/catalog/glossary.

### DIFF
--- a/src/diff-base.ts
+++ b/src/diff-base.ts
@@ -1,0 +1,52 @@
+import type { Session } from "./types";
+import type { PrCache } from "./pr-poller";
+import type { GitForge } from "./forge/types";
+
+export interface ResolvedBase {
+  /** The base branch to diff against. */
+  base: string;
+  /** True when `base` is authoritative — a PR's `baseRefName`, or a definitive "no PR
+   *  exists" — and false when we fell back to `session.baseBranch` because the PR base
+   *  was momentarily unknowable (on-demand `gh` failed / forge unreachable). The diff
+   *  endpoints use `base` regardless; only the recap dedup consults `resolved` (so a
+   *  transient fallback can't flip the dedup key and re-fire a billed recap spawn). */
+  resolved: boolean;
+}
+
+/**
+ * Resolve the base branch a session's diff should compare against: prefer the PR's
+ * actual target (`baseRefName`) so the diff matches the PR's "Files changed" even when
+ * it targets a non-default branch, falling back to the session's stored `baseBranch`.
+ *
+ * Order: warm prCache → definitive cached "no PR" → on-demand `prStatus` (cold/evicted
+ * cache) → `session.baseBranch`. Only a PR base or a definitive no-PR is `resolved:true`;
+ * a transient miss (gh failure / unreachable forge) yields the `baseBranch` fallback with
+ * `resolved:false`.
+ */
+export async function resolveDiffBase(
+  session: Pick<Session, "id" | "baseBranch" | "branch" | "repoPath">,
+  prCache?: Pick<PrCache, "get">,
+  resolveForge?: (repoDir: string) => GitForge | null,
+): Promise<ResolvedBase> {
+  const cached = prCache?.get(session.id);
+  if (cached?.baseRefName) return { base: cached.baseRefName, resolved: true };
+  if (cached && cached.state === "none") return { base: session.baseBranch, resolved: true };
+
+  if (session.branch && resolveForge) {
+    try {
+      // resolveForge can shell out to git (cold cache) and prStatus shells out to gh; both can
+      // throw. Keep the whole thing in try so this never throws out of a caller (e.g. the recap
+      // archive hook) — a failure is simply non-authoritative.
+      const forge = resolveForge(session.repoPath);
+      if (forge) {
+        const st = await forge.prStatus(session.branch);
+        if (st.baseRefName) return { base: st.baseRefName, resolved: true };
+        if (st.state === "none") return { base: session.baseBranch, resolved: true };
+      }
+    } catch {
+      // transient gh/forge failure — not authoritative; fall through to the stored base.
+    }
+  }
+
+  return { base: session.baseBranch, resolved: false };
+}
--- a/src/forge/github.ts
+++ b/src/forge/github.ts
@@ -129,6 +129,7 @@ interface GhPr {
   isDraft?: boolean;
   statusCheckRollup?: RollupEntry[];
   headRefOid?: string;
+  baseRefName?: string;
   reviews?: GhReview[];
   reviewRequests?: { login?: string }[];
   headRepositoryOwner?: { login?: string };
@@ -523,7 +524,7 @@ export class GithubForge implements GitForge {
       "--state",
       "all",
       "--json",
-      "number,url,title,state,createdAt,mergeable,mergeStateStatus,isDraft,statusCheckRollup,headRefOid,reviews,reviewRequests,headRepositoryOwner",
+      "number,url,title,state,createdAt,mergeable,mergeStateStatus,isDraft,statusCheckRollup,headRefOid,baseRefName,reviews,reviewRequests,headRepositoryOwner",
       "--limit",
       this.forkOwner ? "30" : "1",
     ]);
@@ -545,6 +546,7 @@ export class GithubForge implements GitForge {
       isDraft: pr.isDraft ?? false,
       checks: rollupChecks(pr.statusCheckRollup ?? []),
       headSha: pr.headRefOid,
+      baseRefName: pr.baseRefName,
       latestReview: latestHumanReview(pr.reviews),
       requestedReviewers: (pr.reviewRequests ?? [])
         .map((r) => r.login)
--- a/src/forge/types.ts
+++ b/src/forge/types.ts
@@ -154,6 +154,11 @@ export interface PrStatus {
   isDraft?: boolean;
   /** GitHub's merge-eligibility signal; undefined on forges that don't supply it (Gitea). */
   mergeStateStatus?: MergeStateStatus;
+  /** The PR's actual base (target) branch ref name, e.g. "main". Drives the diff/recap
+   *  base so they match the PR's "Files changed" even when it targets a non-default branch.
+   *  Undefined when there is no PR, or on forges/payloads that don't supply it. Unlike
+   *  {@link PullRequest.nonDefaultBase} this IS the raw base ref — safe to diff against. */
+  baseRefName?: string;
   /** A deploy workflow is configured for this host. */
   deployConfigured: boolean;
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -29,6 +29,7 @@ import { EventHub } from "./events";
 import { SessionService } from "./service";
 import { StatusPoller } from "./poller";
 import { PrPoller } from "./pr-poller";
+import { resolveDiffBase } from "./diff-base";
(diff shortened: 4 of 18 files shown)


scoop (commented):
## Summary

The PR makes the diff tab (and recap diff) compare against the PR's actual target branch (`baseRefName`) instead of the session's stored `baseBranch`, so a PR explicitly based off a non-default branch (the TASK-852 shape) shows only its own files. The approach is sound and the change is thorough: a new `resolveDiffBase` resolution (warm cache → cached "no PR" → on-demand `gh` → stored `baseBranch` fallback), a persisted `base` column added to the `(headSha, base)` recap dedup key, and a guard so a *transient* (non-authoritative) fallback can't thrash recap regeneration.

No blocking issues found. One non-blocking doc nit.

## Verification

- **Base resolution** — `resolveDiffBase` precedence and the `resolved` flag are internally consistent (src/diff-base.ts:37-57); the `try/catch` keeps `gh`/forge failures from throwing out of the recap archive hook. Verified `GitState extends PrStatus` and now carries `baseRefName` (src/forge/types.ts:161,168), so `prCache.get(id)?.baseRefName` is type-valid.
- **Forge plumbing** — `GithubForge.prStatus` both requests `baseRefName` in the `--json` field list and maps it onto the result (src/forge/github.ts:76,84). The on-demand `forge.prStatus(session.branch)` call matches the `PrStatus.state` union (`"none"`) it checks (src/forge/types.ts:134).
- **DI wiring** — server diff endpoint passes `deps.prCache` + `deps.resolveForge` (src/server.ts:1548-1549), both present on `AppDeps` (src/server.ts:173,192) and wired from `prPoller`/`resolveForge` (src/index.ts:1115,289). `prPoller` satisfies `Pick<PrCache,"get">` — `PrPoller implements PrCache` with `get()` at src/pr-poller.ts:107,294. The `recapService` closure references `prPoller`/`resolveForge` before their `const` declarations (src/index.ts:275 vs 289/519), but it only runs at recap time — the explicit `: RecapService` annotation breaks the inference cycle and the TDZ never trips. Confirmed safe.
- **Dedup key** — `needsRecap(existing, head, base, resolved)` only force-regenerates on base when `resolved && existing.base !== "" && existing.base !== currentBase` (src/recap-core.ts:171-182); the `existing.base !== ""` legacy guard plus the `base` column default `''` (src/store.ts:349) together prevent a deploy from mass-regenerating pre-existing recaps. All three callers updated (src/recap.ts:317-320,367-370); `regenerate()` bypasses dedup but `generate()` resolves the base itself via `knownBase ?? (await this._resolveBase(session)).base` (src/recap.ts:434), so a forced regenerate never re-bakes a stale base. No un-migrated `needsRecap`/`generate` callers remain (grep clean).
- **Persistence** — `base` added to the `recaps` table, the additive `ALTER TABLE … DEFAULT ''` migration, and all four SELECTs + the INSERT/UPSERT bind list (16 columns, 16 `?`, 16 values) are consistent (src/store.ts:338,348-350,2206,2216-2244,2252,2407,2177). The `ready`/`failed` finalize rows inherit `base` via `...rBase` from the generating row (src/recap.ts:612-613), so it survives finalize.
- **Retarget propagation** — `gitStateChanged` now compares `baseRefName` (src/pr-poller.ts:84), so a PR retargeted to a new base updates the cache and re-emits, keeping the diff base current.
- **Recap type** — `Recap.base: string` is required (src/types.ts:348); every in-tree `Recap` literal (src/recap.ts:445,517 and the test factories) sets it.

## Nits

- `considerForArchive`'s doc comment still says "Head-keyed dedup" (src/recap.ts:352), but this PR widened the key to `(head, base)`. The sibling comments on `needsRecap` (src/recap-core.ts:159) and `considerSession` were updated; this one was missed. Non-blocking.

- src/forge/gitea.ts: `GiteaForge.prStatus` does not populate `baseRefName` (it has `pr.base?.ref` at gitea.ts:260, used only for `nonDefaultBase`), so the non-default-base diff fix is effectively GitHub-only; Gitea sessions degrade gracefully to the `baseBranch` fallback (`resolved:false`). Not required for TASK-852 (a GitHub repo) and no regression, so informational only.

## Limitations
- The diff endpoint may now perform one on-demand `gh` shell-out on a cold/evicted PR cache (src/diff-base.ts:48); this is by design and avoided on the warm cache the poller maintains. Not verified against live latency.

scoop (commented):
## Summary

This PR makes the diff (and recap) compare against the PR's actual target branch (`baseRefName`) instead of the session's stored `baseBranch`, which is exactly what TASK-852 requires. The implementation is correct and the single prior-revision finding is resolved.

## Re-review of prior finding

1. **src/recap.ts:352 — `considerForArchive` doc comment said "Head-keyed dedup".** ADDRESSED. The comment now reads `` `(head, base)`-keyed dedup (see {@link needsRecap}) `` (src/recap.ts:352-356), matching the sibling comments on `needsRecap` (src/recap-core.ts:191) and the live-sweep path. Not re-raised.

## Verification performed

- **Base plumbing into the diff endpoint (the task):** `sessionDiffRead` now calls `resolveDiffBase(s, deps.prCache, deps.resolveForge)` and passes the resolved `base` to `computeDiff` (src/server.ts:1546-1549). `computeDiff` fetches `origin/<base>` and diffs `origin/<base>...HEAD` (src/diff.ts:291-298), so a non-default base like `dev` resolves correctly. This makes the diff tab reflect the PR's "Files changed". Task satisfied.
- **`baseRefName` source:** added to `GhPr` (src/forge/github.ts:132), requested in the `--json` field list of `prStatus` (src/forge/github.ts:527) and mapped into the returned `PrStatus` (src/forge/github.ts:549); declared on `PrStatus` (src/forge/types.ts:161) and inherited by `GitState` (src/forge/types.ts:168). `state` enum confirmed to include `"none"` (src/forge/types.ts:134), which `resolveDiffBase` relies on.
- **`resolveDiffBase` resolution order** (src/diff-base.ts:30-51): warm cache `baseRefName` → cached `state==="none"` → on-demand `prStatus` (try/catch so a transient gh/forge failure never throws out of the recap archive hook) → `session.baseBranch` fallback with `resolved:false`. `prCache.get` returns `GitState` which carries `baseRefName`/`state` — verified against src/pr-poller.ts:12,294.
- **Dedup correctness (no double-billing):** `needsRecap` (src/recap-core.ts:203-216) only re-fires on the base dimension when `resolved && existing.base !== "" && existing.base !== currentBase`. The `resolved` guard prevents a transient `baseBranch` fallback from flipping the key (no flapping), and the `existing.base !== ""` guard stops a deploy from mass-regenerating legacy rows. Converges to no-refire once `existing.base === currentBase`. Both call sites updated to 4 args (src/recap.ts:318, 370); no stale 2-arg callers remain (grep).
- **`base` persistence:** added to `Recap` (src/types.ts:349, required field so the type system forces all constructions to set it), `RecapRow` (src/store.ts:305), CREATE TABLE (src/store.ts:738) and an idempotent `ALTER TABLE … ADD COLUMN base … DEFAULT ''` migration (src/store.ts:765-767). `putRecap` INSERT now has 16 named columns / 16 placeholders / 16 binds with `recap.base ?? ""` in the matching position (src/store.ts:2216-2244); all SELECTs (`getRecap`, `snapshotRecaps`, `generatingRecaps`) and `hydrateRecap` include `base` (src/store.ts:2178, 2206, 2252, 2266). All access is by column name, so the differing physical column position between fresh and migrated DBs is harmless. Both row constructions in `generate` set `base` (src/recap.ts:448, 520); `finalize` carries it via `...rBase` spread of the generating row (src/recap.ts:603,615,636).
- **Poller change-detection:** `gitStateChanged` now diffs `baseRefName` (src/pr-poller.ts:84), so a PR retarget propagates to the cache/UI.
- **Index wiring / TDZ:** `resolveBase: (s) => resolveDiffBase(s, prPoller, resolveForge)` (src/index.ts:275) references `resolveForge` (declared src/index.ts:289) and `prPoller` (src/index.ts:519) which are defined after `recapService`; the closure runs only at recap time, so the forward references are safe, and the explicit `: RecapService` annotation breaks the inference cycle as documented.
- **Doc link:** `{@link PullRequest.nonDefaultBase}` (src/forge/types.ts:160) resolves — `PullRequest.nonDefaultBase` exists (src/forge/types.ts:126).

## Notes (non-blocking, in scope)

- For forges that never supply `baseRefName` (Gitea) or a cold GitHub cache, `considerSession` performs one on-demand `prStatus` shell-out per idle episode — but it is gated by `entry.fired` (src/recap.ts:295,315), so it is once-per-episode, not per-sweep. No hot-loop concern.

No blocking problems found; nothing to change.

Issue #711, opened 2026-06-15
Hook ingestion: make agent→Shepherd reachable under the autonomous egress profile


scoop:
Follow-up from the Phase-0 hook-ingestion spike (#704, PR #708). **Push hooks do not reach Shepherd under the `autonomous` sandbox profile** — they silently fall back to polling there. This is fail-open (no regression), but `autonomous` (unattended drain) is exactly where push signals would help most, so it's worth closing the gap deliberately.

## Why it doesn't reach today (code-confirmed, 2026-06-15)
- The injected hook URL is `agentBaseUrl()` = `http://127.0.0.1:<port>` (`src/service.ts:660`).
- `egressApplies` (`src/sandbox.ts:~513`) confines only `autonomous`; `trusted`/`standard` share the host net namespace, so `127.0.0.1` reaches Shepherd there (spike-confirmed for trusted).
- Under `autonomous` the agent runs in a slirp4netns **netns**: `127.0.0.1` is the *namespace's* loopback (Shepherd listens on the *host* loopback). The nft ruleset (`src/egress.ts:~433`) is `policy drop` + `oifname "lo" accept` + DNS(10.0.2.3) + `ip daddr @allowed accept`, where `@allowed` holds only resolved **domain** IPs (Anthropic/GitHub). The host/loopback is not reachable or allowlisted → the HTTP hook (and the build-queue curl, same mechanism) cannot connect.

## Scope / options
Make the agent→Shepherd control-plane channel reachable from inside the autonomous netns, then it covers both hooks **and** the existing build-queue curl:
1. Resolve the host-reachable address from inside the netns (slirp4netns host gateway, typically `10.0.2.2`, requires `--enable-host-loopback` / `allow_host_loopback`), and use it for `agentBaseUrl()` **when the spawn is autonomous** (keep `127.0.0.1` otherwise).
2. Allowlist that host IP + Shepherd's port in the egress nft set (a non-domain, IP+port allow rule — distinct from the dnsmasq domain-pinning path).
3. Keep it least-privilege: allow only Shepherd's port to the host gateway, nothing else; confirm it doesn't widen the exfil surface.

## Constraints / notes
[...]

PR #1081, opened 2026-06-25
feat(auth): single-operator password → session-cookie auth; close the WS/PTY gap

### LINKED ISSUE #1079: Scope: single-operator authentication layer (password → session cookie; close the WS/PTY gap)
## Summary

Shepherd currently sits behind Tailscale but is **fully unauthenticated**. The only gate is an optional `SHEPHERD_TOKEN` bearer check (`src/server.ts:370`) that is env-only, all-or-nothing, and — critically — **not enforced on the WebSocket upgrades**. A tailnet peer hitting `ws://host:7330/pty/<uuid>` today gets a live PTY (a real shell) with no credential check; the `/events` bus likewise broadcasts every session's activity to any connected client.

This issue scopes a **single-operator password → session-cookie authentication layer**, hand-rolled on Bun built-ins (no new deps), that gates every HTTP route **and both WebSocket channels**, and closes the PTY/events gap.

Scope was settled via a full design grilling; the decisions below are final.

## Threat model & constraints (settled)

- **Single operator.** One shared credential. No user table, no per-session ownership, no roles. Identity check lives behind one middleware seam so multi-user/SSO *could* be added later without a rewrite — but none of that machinery is built now.
- **Tailscale + defense-in-depth.** Shepherd stays behind Tailscale and is **never** recommended for direct public exposure. Auth is a second layer against tailnet peers, a fat-fingered `SHEPHERD_HOST=0.0.0.0`, or a forgotten `tailscale serve`. No public-internet hardening (lockout dashboards, captchas) — but the cheap, always-correct controls (timing-safe compares, proper cookie flags, closing the WS gap) are done unconditionally.

## Design decisions (settled)

| # | Decision | Choice |
|---|----------|--------|
| Audience | identities to serve | **Single operator**, one shared credential; seam left extensible |
| Threat model | what to defend | **Tailscale + defense-in-depth**; never publicly exposed |
| Credential | form | **Password → argon2id hash at rest → signed session cookie** |
| Build | library vs hand-roll | **Hand-rolled on Bun built-ins** (`Bun.password`, `crypto`); **no new deps**. (Lucia deprecated; better-auth is multi-user/framework-coupled — wrong tool for one shared password.) |
| Session | stateful vs stateless | **Stateless HMAC-signed cookie**; signing secret auto-generated + persisted once in `settings`. (Avoids a DB read on the hot path — the server is one Bun loop pumping the web terminal.) |
| Exemptions | what stays open | **Only**: the login route, static GETs for the SPA shell, and the separate loopback agent-ingress listener. Everything else gated. `Origin` check retained alongside. |
| Default | no password configured | **Fail-closed + auto-generate**: generate a strong random password, persist its hash, print plaintext **once** to the log with a loud "CHANGE THIS / set `SHEPHERD_PASSWORD`" banner. App is never silently open. |
| Lifetime | cookie TTL | **7-day sliding** window; cookie re-stamped when past half-life (~3.5d). Idle ⇒ expires; active ⇒ effectively never re-login. |
| Login UX | surfacing | **In-layout login view** (no dedicated route). `401` interceptor in `ui/src/lib/api.ts` + a lightweight `/api/me` boot probe (clean first paint). Logout button in **Settings → SESSION**. |
| Password change | in-app? | **No change form.** Env-only. Precedence: `SHEPHERD_PASSWORD` set ⇒ authoritative each boot (re-seeds hash); unset ⇒ persisted hash used as-is (auto-generated one survives restarts). |
| Brute-force | login throttle | **None beyond argon2id cost** (≈100ms/try is the rate limit). Plus **timing-safe** (`crypto.timingSafeEqual`) cookie-signature compare. |
| Agent token | `SHEPHERD_TOKEN` fate | **Keep** as the agent→server callback credential. Gate = **cookie OR bearer token**. Boot bootstrap also provisions an agent token when `config.token` is null, so fail-closed doesn't kill hook callbacks. |
| Cookie flags | attributes | `HttpOnly` + `SameSite=Strict` **always**; `Secure` **conditional** (set when request arrived over HTTPS via `X-Forwarded-Proto`/scheme — keeps `http://127.0.0.1:7330` usable for local debugging). |

## Why the WebSocket gap exists (and how it's fixed)

The `/events` and `/pty/:id` upgrades happen in the **outer `serve()` fetch wrapper** (`src/server.ts:5029-5068`), which **bypasses `makeApp` entirely** — that's *why* `checkAuth`/the token never applied to them. The fix is an explicit **cookie-or-token check at both upgrade sites**, before `server.upgrade()`, alongside the existing `Origin` check. Browsers attach `HttpOnly` cookies to same-origin WS upgrades automatically, so the UI needs no change; an un-cookied peer gets rejected pre-upgrade.

## Key dependency: do not break agent hook-callbacks

Spawned agents phone home with `Authorization: Bearer $SHEPHERD_TOKEN` (`src/service.ts:278-281`). Today, with `token === null`, callbacks succeed only because everything is open. The moment we go **fail-closed**, an un-credentialed callback would be rejected. Therefore the boot bootstrap **must provision an agent token** (auto-generate + persist when null) at the same time it provisions the password hash. The gate accepts **cookie (humans) OR bearer token (agents)**. The loopback agent-ingress listener (`serveAgentIngress`) keeps its existing token+allowlist+loopback posture, unchanged.

### Phase 1 — Auth core + gate + bootstrap (server)
- New `src/auth.ts` (or similar): `hashPassword`/`verifyPassword` (`Bun.password`, argon2id); `signCookie`/`verifyCookie` (HMAC via `crypto`, `timingSafeEqual`); cookie parse/serialize with `HttpOnly`/`SameSite=Strict`/conditional `Secure`/7-day sliding.
- Boot bootstrap (in `src/index.ts` boot-override path): if no password hash in `settings` and `SHEPHERD_PASSWORD` unset ⇒ generate, hash, persist, **log plaintext once** with banner. If `SHEPHERD_PASSWORD` set ⇒ it wins (re-seed hash). Auto-generate + persist the **agent token** + the **cookie signing secret** if absent.
- Replace/extend `checkAuth` (`src/server.ts:370`) so the gate in `makeApp.fetch` (`src/server.ts:4891`) is **valid cookie OR valid bearer token**, with the Q6 exemptions.
- `POST /api/login` (verify password → `Set-Cookie`), `POST /api/logout` (clear cookie), `GET /api/me` (200 if authed, else 401). Login route is the only unauthenticated mutation.

### Phase 2 — WebSocket enforcement (server)
- Add the cookie-or-token check at the `/events` and `/pty/:id` upgrade sites (`src/server.ts:5029-5068`), before `server.upgrade()`, keeping the `Origin` check. Reject un-credentialed upgrades pre-upgrade. This is the actual security fix for the live-PTY hole.

### Phase 3 — UI login/logout + house-rule obligations (UI)
- `401` interceptor in `ui/src/lib/api.ts` (mirror the existing `403 → PreviewBlockedError` pattern) → set an `unauthenticated` store flag → root layout swaps to a `<Login>` view. `/api/me` probe on boot to avoid a first-paint flash of failing calls.
- Login view + logout button (Settings → SESSION, where auth-mode/model already live), per the **design system** (tokens only, reuse `.gbtn`/field recipes; login is a blocking surface ⇒ honor the scrim/blur rule).
- **i18n**: every new string in EN+DE (`ui/messages/{en,de}.json`); `check:i18n` enforces parity.
- **Feature catalog**: add one `FeatureAnnouncement` to `ui/src/lib/feature-announcements.ts` (`sinceVersion: 1.37.0` — next unreleased), with title/body keys in both locales. (Or `[no-feature-entry]` if we decide the login screen isn't worth announcing — but it is user-facing UX, so include it.)
- **Glossary**: no new defined terms expected; none required unless we introduce one.

## Required tests

Root server (`bun test ./test`):
- Cookie **sign → verify roundtrip**; **tamper rejection** (flipped byte / wrong secret) via timing-safe compare.
- Gate accepts **valid cookie**, accepts **valid bearer token**, rejects **neither**; exemptions (login route, static GET) pass un-credentialed.
- **Fail-closed bootstrap**: no password + no env ⇒ password hash, agent token, and signing secret all auto-generated + persisted; plaintext logged once. `SHEPHERD_PASSWORD` set ⇒ overrides.
- **Sliding re-stamp**: cookie past half-life ⇒ fresh `Set-Cookie`; within first half ⇒ none.
- **Conditional `Secure`**: HTTPS request (`X-Forwarded-Proto: https`) ⇒ `Secure`; plain loopback ⇒ no `Secure`; `HttpOnly`+`SameSite=Strict` always.
- **WS upgrade**: `/events` and `/pty/:id` rejected without credential, accepted with cookie; `Origin` check still applies.
- **Agent callback regression**: with a provisioned token, the hook-callback path (`src/service.ts`) still authenticates after fail-closed flip.

UI (`cd ui && bun run test`):
- `401` interceptor flips the `unauthenticated` store; layout renders login.
- Login submit → success path re-fetches; failure path shows error (i18n).

## Migration note (explicit, signed off)

The deploy that ships this is **not** seamless-open. The live instance (currently no password) will require a login afterward. Before/at upgrade: either set `SHEPHERD_PASSWORD` in `~/.shepherd/env`, or fish the auto-generated password out of `~/.shepherd/shepherd.log` on first boot. This is the intended outcome of "stop being unauthenticated."

## Out of scope (explicitly deferred)

- Multi-user accounts, roles, per-session ownership/visibility, SSO/OAuth.
- In-app password-change form.
- Brute-force lockout / rate-limit counters / attempt auditing (public-internet posture, ruled out).
- Per-session authz on `/events`/`/pty` beyond "authenticated operator" (single-operator ⇒ all sessions are yours).

---

scoop:
Closes #1079.

## What

Shepherd was fully unauthenticated behind Tailscale — and the live PTY (`ws://host/pty/<uuid>`) + `/events` bus were reachable with no credential. This adds a **single-operator password → argon2id hash → HMAC-signed session cookie** layer, hand-rolled on Bun built-ins (**no new deps**), that gates **every HTTP route AND both WebSocket channels**, fails closed, and keeps spawned-agent hook callbacks working on every sandbox profile.

## How

- **`src/operator-auth.ts`** (new, pure): `hashPassword`/`verifyPassword` (`Bun.password` argon2id), `signCookie`/`verifyCookie` (HMAC-SHA256 + `timingSafeEqual`), cookie parse/serialize (`HttpOnly`+`SameSite=Strict` always, **conditional `Secure`**, 7-day **sliding** window), and `bootstrapAuth` (fail-closed).
- **Boot bootstrap** (`src/index.ts`): provisions the cookie-signing secret + password hash before serving. `SHEPHERD_PASSWORD` wins (re-seeds each boot); else the persisted hash is reused; else a strong password is **auto-generated, hashed, persisted, and logged once** with a CHANGE-THIS banner. The app is never silently open.
- **Gate** (`src/server.ts`): `checkAuth` = valid session cookie **OR** operator bearer (`config.token`, when set). One seam — `makeApp.fetch` and `serve().fetch` both call it, so HTTP routes **and** the WS upgrades inherit it. Scoped static-shell exemption that deliberately keeps `/events` + `/pty/:id` gated. New `POST /api/login`, `POST /api/logout`, `GET /api/me`. Sliding re-stamp attached at one seam (the `makeApp` wrapper), never on WS.
- **Agents** reach the server through the **loopback agent-ingress listener**, now built with `skipAuth` (its loopback bind + route allowlist + per-session UUID is the auth). `resolveSpawnBaseUrl` routes trusted/standard callbacks there too (autonomous already did via `10.0.2.2`). `config.token` is **not** auto-provisioned.
- **UI**: 401 interceptor in `api.ts` → `unauthenticated` store → in-layout `<Login>` (scrim/blur, tokens, `.gbtn`/field); `/api/me` boot probe (no first-paint flash); logout in **Settings → Session**; EN+DE strings; one feature-catalog entry (1.37.0).

## Deviations from the issue (flagged per house rules)

1. **WS enforcement unified in `checkAuth`** instead of literal per-upgrade-site checks — `serve().fetch` already calls `checkAuth` before the upgrade branches (`src/server.ts`), so both channels inherit the gate from one function. (Operator-approved during planning.)
2. **Module is `src/operator-auth.ts`** (issue suggested `src/auth.ts`) — `auth-mode.ts`/`auth-config-dir.ts` already exist for the unrelated spawned-agent API-key concern.
3. **Agents authenticate via the exempt ingress, NOT a provisioned token.** The issue's "provision an agent token" would have *broken* hooks: `buildHooksFragment` emits a `$SHEPHERD_TOKEN` placeholder resolved from the agent's env (run under the herdr daemon), and autonomous `--clearenv` strips it (#711) — so a non-null `config.token` degrades hooks to polling. Leaving `config.token` null + routing all agents through the auth-exempt loopback ingress keeps hooks event-driven with no secret in any agent env. (Reviewer-approved re-scope.)

## Migration (signed off in the issue)

The deploy that ships this is **not** seamless-open: the live instance will require a login. Set `SHEPHERD_PASSWORD` in `~/.shepherd/env`, or read the auto-generated password from `~/.shepherd/shepherd.log` on first boot. Agent hooks are **not** downgraded (they ride the exempt ingress). Logout is client-side (stateless cookie); rotating the cookie secret is the all-sessions kill-switch.

## Tests

- Root (`test/operator-auth.test.ts`, `test/server-auth.test.ts`): cookie sign→verify roundtrip + tamper/expiry rejection; gate accepts cookie / accepts bearer / rejects neither / exemptions pass; fail-closed bootstrap (hash + secret generated + persisted, logged once, `SHEPHERD_PASSWORD` overrides); sliding re-stamp (2xx past half-life only, never WS); conditional `Secure`; WS `/events` + `/pty/:id` rejected without credential, accepted with cookie, `Origin`/CSWSH still applies (real ephemeral-port server); **agent-transport regression** — uncredentialed hook POST is 2xx on the ingress, 401 on the main port; `resolveSpawnBaseUrl` bakes the ingress URL per profile.
- UI (`api.auth.test.ts`): 401 flips the store; login success/failure; logout.
- Full root suite (4885) + UI (2054) green; `tsc`, `eslint`, `prettier --check`, `check:i18n`, UI build, branch-hygiene, feature-catalog, glossary all pass.

### DIFF
--- a/docs-site/src/content/docs/getting-started.md
+++ b/docs-site/src/content/docs/getting-started.md
@@ -67,6 +67,12 @@ tailscale serve --bg 7330
 # (in ~/.shepherd/env or deploy/shepherd.service)
 ```
 
+The HUD is gated by a **single-operator password**: the first time you open it
+you'll get a login screen. Set the password with `SHEPHERD_PASSWORD` (in
+`~/.shepherd/env`), or use the strong one Shepherd generates and prints to the
+log **once** on first boot (`systemctl --user status shepherd` / `journalctl
+--user -u shepherd`). Log out from **Settings → Session**.
+
 **Settings → DIAGNOSE** surfaces any remaining gaps with one-click fixes.
 
 ## From-clone / development path
--- a/docs-site/src/content/docs/operating.md
+++ b/docs-site/src/content/docs/operating.md
@@ -27,10 +27,16 @@ tailscale serve --bg 7330    # → https://<host>.<tailnet>.ts.net proxies to 12
 ```
 
 Add the public hostname to `SHEPHERD_ALLOWED_HOSTS` (the unit ships with the
-Tailscale name). Access control is **tailnet membership** — there is no app-level
-password.
+Tailscale name). Access control is layered: the network reach is gated by
+**tailnet membership**, and the app itself is gated by a **single-operator
+password**. The password is exchanged for an HMAC-signed session cookie that
+covers every HTTP route plus the live `/events` and `/pty` WebSocket channels.
+Set it with `SHEPHERD_PASSWORD`; if you leave it unset, Shepherd generates a
+strong one on first boot and prints it to the log **once** (`systemctl --user
+status shepherd` / `journalctl --user -u shepherd`) — change it by setting
+`SHEPHERD_PASSWORD` in `~/.shepherd/env` and restarting.
 
-Per-deployment overrides (token, repo root, alternate hosts) go in
+Per-deployment overrides (password, token, repo root, alternate hosts) go in
 `~/.shepherd/env` (`KEY=value` lines), read by the unit if present. See
 [Configuration](/reference/configuration/) for the full list.
 
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -16,7 +16,9 @@ lines), read by the systemd unit if present.
 | `SHEPHERD_DB` | `~/.shepherd/shepherd.db` | SQLite session store path |
 | `SHEPHERD_REPO_ROOT` | `~` (home) | Repos must live under this root (spawn is confined to it) |
 | `SHEPHERD_ALLOWED_HOSTS` | `localhost,127.0.0.1,::1,[::1]` | Comma-separated origin hostnames allowed for writes + WS (CSRF/CSWSH guard) |
-| `SHEPHERD_TOKEN` | _(none)_ | When set, require `Authorization: Bearer <token>` |
+| `SHEPHERD_PASSWORD` | _(auto-generated)_ | Single-operator login password. When set it's authoritative — argon2id-hashed and re-seeded into the persisted hash every boot. Unset → the persisted hash is reused, or (first boot) a strong password is generated, hashed, persisted, and printed to the log **once**. The browser exchanges it for an HMAC-signed session cookie that gates every HTTP route plus the `/events` + `/pty` WebSocket channels |
+| `SHEPHERD_COOKIE_SECRET` | _(generated + persisted)_ | HMAC secret that signs the session cookie. Set it to pin a stable secret across DB resets; rotating it invalidates every outstanding session (the all-sessions kill-switch) |
+| `SHEPHERD_TOKEN` | _(none)_ | Optional operator bearer for CLI/curl/machine clients: when set, `Authorization: Bearer <token>` is accepted as an alternative to the session cookie. Browser operators use the password login instead; spawned agents don't use this (they reach the server over the loopback ingress) |
 | `HERDR_BIN` | `herdr` | Path to the herdr binary |
 | `HERDR_SESSION` | `default` | herdr session name |
 | `SHEPHERD_FORGES` | `~/.shepherd/forges.json` | Path to the git-host config |
--- a/docs/external-task-api.md
+++ b/docs/external-task-api.md
@@ -3,14 +3,15 @@
 Shepherd's HTTP API is the same surface the UI uses — there is **no separate
 "public" API and no CORS barrier** for non-browser clients. Any agent that can
 reach the core process (Hermes, a cron job, another service) can queue work by
-calling one endpoint.
+calling one endpoint — once it authenticates (the server is gated by default;
+see [What actually gates access](#what-actually-gates-access)).
 
 ## TL;DR
 
 ```bash
 curl -X POST http://localhost:7330/api/sessions \
   -H "Content-Type: application/json" \
-  -H "Authorization: Bearer $SHEPHERD_TOKEN" \   # omit if SHEPHERD_TOKEN is unset
+  -H "Authorization: Bearer $SHEPHERD_TOKEN" \   # required — the server is gated by default
   -d '{
         "repoPath": "~/Work/my-repo",
         "baseBranch": "main",
@@ -49,18 +50,20 @@ its hostname to `SHEPHERD_ALLOWED_HOSTS`.
 
 ## What actually gates access
 
-| Gate                 | Default                           | What Hermes must do                                                                               |
-| -------------------- | --------------------------------- | ------------------------------------------------------------------------------------------------- |
-| **Network bind**     | `127.0.0.1:7330` (loopback only)  | Run on the same host, reach it over Tailscale serve, or set `SHEPHERD_HOST` to expose another NIC |
-| **Auth token**       | `SHEPHERD_TOKEN` unset → open     | If set, send `Authorization: Bearer <token>` on every request (timing-safe compare)               |
-| **Repo confinement** | `SHEPHERD_REPO_ROOT` = `~` (home) | `repoPath` must resolve **inside** the root, or the request is rejected `400`                     |
+| Gate                 | Default                                               | What Hermes must do                                                                                                                                                                                                |
+| -------------------- | ----------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| **Network bind**     | `127.0.0.1:7330` (loopback only)                      | Run on the same host, reach it over Tailscale serve, or set `SHEPHERD_HOST` to expose another NIC                                                                                                                  |
+| **Auth**             | Gated by default (operator password → session cookie) | Machine clients can't use the browser login — set `SHEPHERD_TOKEN=<random>` and send `Authorization: Bearer <token>` on every request (timing-safe compare). Without a valid cookie or bearer the request is `401` |
+| **Repo confinement** | `SHEPHERD_REPO_ROOT` = `~` (home)                     | `repoPath` must resolve **inside** the root, or the request is rejected `400`                                                                                                                                      |
 
 ### Recommended setup for a remote agent
 
 1. Expose the core to the agent's network — prefer **Tailscale serve** over
    `SHEPHERD_HOST=0.0.0.0`. Add the ts.net hostname to `SHEPHERD_ALLOWED_HOSTS`
    only if the agent is browser-based (sends `Origin`).
-2. Set a shared secret: `SHEPHERD_TOKEN=<random>` and give it to Hermes.
+2. Set a shared secret: `SHEPHERD_TOKEN=<random>` and give it to Hermes. The
+   server is **gated by default**, and machine clients can't use the browser
+   password login — the bearer token is how they authenticate.
 3. Keep `SHEPHERD_REPO_ROOT` tight so Hermes can only target intended repos.
 
 ## Request schema
@@ -79,14 +82,14 @@ its hostname to `SHEPHERD_ALLOWED_HOSTS`.
 
 ### Responses
 
-| Status | Meaning                                                                                                               |
-| ------ | --------------------------------------------------------------------------------------------------------------------- |
-| `200`  | **Held** by the usage-aware hold gate — body `{ held: true, id, count }`; the task is queued, not spawned (see below) |
-| `201`  | Created — body is the full `Session` (`id`, `desig`, `status`, `worktreePath`, …)                                     |
-| `400`  | Validation failed — body `{ error }`                                                                                  |
-| `401`  | `SHEPHERD_TOKEN` set and bearer missing/wrong                                                                         |
-| `403`  | Origin header present and not in `SHEPHERD_ALLOWED_HOSTS`                                                             |
-| `415`  | Missing/incorrect `Content-Type`                                                                                      |
+| Status | Meaning                                                                                                                                                |
+| ------ | ------------------------------------------------------------------------------------------------------------------------------------------------------ |
+| `200`  | **Held** by the usage-aware hold gate — body `{ held: true, id, count }`; the task is queued, not spawned (see below)                                  |
+| `201`  | Created — body is the full `Session` (`id`, `desig`, `status`, `worktreePath`, …)                                                                      |
+| `400`  | Validation failed — body `{ error }`                                                                                                                   |
+| `401`  | No valid session cookie **and** no valid bearer token (the server is gated by default — machine clients must set `SHEPHERD_TOKEN` and send the bearer) |
+| `403`  | Origin header present and not in `SHEPHERD_ALLOWED_HOSTS`                                                                                              |
+| `415`  | Missing/incorrect `Content-Type`                                                                                                                       |
 
 ## Usage-aware hold gate
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -326,7 +326,15 @@ export const config = {
   allowedOriginHosts: (process.env.SHEPHERD_ALLOWED_HOSTS ?? "localhost,127.0.0.1,::1,[::1]").split(
     ",",
   ),
-  token: process.env.SHEPHERD_TOKEN ?? null, // when set, require Authorization: Bearer <token>
+  token: process.env.SHEPHERD_TOKEN ?? null, // optional operator bearer (CLI/curl); agents use the loopback ingress, not this
+  // Single-operator auth (issue #1079). `password` (env) is authoritative when set: it re-seeds
+  // the persisted argon2id `passwordHash` every boot. Unset ⇒ the persisted hash is used as-is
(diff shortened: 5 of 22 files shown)


scoop:
## Summary

The PR implements the single-operator auth layer well and faithfully closes the WebSocket/PTY gap: `checkAuth` now gates `/events` and `/pty/:id` at the `serve().fetch` upgrade seam (src/server.ts:5159, with `isPublicRequest` excluding those paths at src/server.ts:393), and the agent-callback path is rerouted through the loopback ingress (src/service.ts:1152, exempt via `skipAuth` at src/server.ts:5088). i18n parity, the feature announcement, and the cookie primitives all check out. One reachable logout bug blocks.

## Blocking finding

### Logout can fail to clear the session (cookie gets re-stamped on the logout response)

`handleLogout` returns a 200 carrying the clearing `Set-Cookie` (`shepherd_session=; Max-Age=0`) — src/server.ts:511-513. That response is then passed through `maybeRestamp` in the `makeApp` wrapper — src/server.ts:5054 (`.then((res) => (opts.skipAuth ? res : maybeRestamp(req, res)))`).

`maybeRestamp` (src/server.ts:435-446) guards only on `res.status >= 400`. The logout response is 200, the secret is present, and it re-reads the **request** cookie (still the valid old session). If that cookie is past its half-life, `shouldRestamp(v.iat)` is true and it does `headers.append("Set-Cookie", serializeCookie(signCookie(secret), ...))` — appending a **second** `Set-Cookie` that re-issues a fresh 7-day session **after** the clearing one. The browser sees two `Set-Cookie: shepherd_session=...` headers (clear, then set); the later set wins, so the operator remains authenticated. Logout silently fails.

This is the security guarantee logout provides here — there is no server-side revocation (per src/operator-auth.ts:11), so the cookie clear IS the logout. Reachability: any cookie-authed request that is the first to cross the 3.5-day half-life triggers a restamp; when that request is `POST /api/logout` (e.g. a long-lived tab, or returning after idle and signing out before any other restamp-eligible HTTP call fires), logout is defeated. The existing test only exercises a fresh cookie (`signCookie(SECRET)`, iat=now → no restamp), so it does not catch this (test/server-auth.test.ts logout case).

Fix: have `maybeRestamp` skip the logout route (or any response that already carries a clearing `Set-Cookie`), or exclude `POST /api/logout` from the restamp seam. The doc comment at src/server.ts:431 claims it "can't double-apply," but for logout it double-applies with conflicting intent.

## Verification notes

- i18n parity: all 12 new keys present in BOTH catalogs (ui/messages/en.json:1929-1941 and ui/messages/de.json:1929-1941); every `m.*` key used in Login.svelte and Settings.svelte resolves. Verified login_title/subtitle/password_label/password_placeholder/submit/busy/error, settings_logout_title/hint/button, feat_operator_auth_title/body in both.
- `FeatureAnnouncement.targetId` is optional (ui/src/lib/feature-announcements.ts:15), so the new entry omitting it is valid; `sinceVersion: 1.37.0` is the next unreleased version (package.json is 1.36.0).
- WS gate: `isPublicRequest` returns false for `/events` and `/pty/` (src/server.ts:393), so the `serve().fetch` `checkAuth` at src/server.ts:5159 now rejects un-cookied upgrades pre-`server.upgrade()`; the `Origin`/CSWSH check is retained (src/server.ts:5167-5174). Confirmed against test/server-auth.test.ts WS cases.
- Agent-callback survival after fail-closed: trusted/standard spawns now bake the loopback ingress URL when the ingress port is known (src/service.ts:1152-1153); the ingress app is built with `skipAuth: true` (src/server.ts:5088) and the hook/queue routes the agents call are all on the `isAgentIngressRoute` allowlist (src/server.ts:5067-5078). `agentIngressState.port` is set (src/index.ts:1761) before sessions spawn, so the gated-main-port 401 fallback is genuinely pathological.
- Existing `makeApp(deps)` test callers stay open because `checkAuth` short-circuits when `config.cookieSecret === null && config.token === null` (src/server.ts:415) — preserves the prior token-null contract.

### Design deviation (not a defect)

The originating issue's plan called for provisioning an **agent token** and gating on cookie-OR-token, with agent callbacks carrying `Authorization: Bearer`. The implementation instead routes ALL agent callbacks through the exempt loopback ingress and provisions no agent token (src/index.ts:219-222 comment, src/service.ts:1148-1153). This is a coherent, arguably better choice (no credential in the agent env, works under `--clearenv`) and still satisfies the issue's hard requirement — agent hook-callbacks survive the fail-closed flip. Flagging only so reviewers know the literal token-provisioning step in the issue was intentionally dropped.

- None.

scoop (commented):
## Re-review — issue #1079 single-operator auth

### Prior finding (logout re-stamp) — RESOLVED

> `maybeRestamp` re-issued a fresh valid session cookie on `POST /api/logout` when the incoming cookie was past half-life, defeating logout.

Fixed and verified:
- `maybeRestamp` (src/server.ts:455) returns early via `responseSetsSessionCookie(res)` (src/server.ts:447) whenever the response already sets a `shepherd_session=` cookie. `clearCookie` (src/operator-auth.ts:123) produces `shepherd_session=; Path=/; HttpOnly; SameSite=Strict; Max-Age=0`, and `responseSetsSessionCookie` matches on `.includes("shepherd_session=")` → logout (and login) responses are excluded from the restamp seam. It also guards `res.status >= 400`.
- Regression test present: test/server-auth.test.ts:140 sends a restamp-eligible (0.6·TTL old) cookie to `/api/logout` and asserts exactly ONE Set-Cookie, `Max-Age=0`, and no second non-zero Max-Age cookie. Verified against test/server-auth.test.ts:140-152.

### WebSocket gap (the actual security fix) — VERIFIED CLOSED
- The pre-PR `serve().fetch` already called `checkAuth` before the `/events` and `/pty/:id` upgrades (verified against `git show f344…:src/server.ts:5026`), but the old `checkAuth` was a no-op when `config.token === null` (production). The new `checkAuth` (src/server.ts:414) enforces cookie-or-token whenever `config.cookieSecret` is set (always, post-bootstrap), and `isPublicRequest` (src/server.ts:385) explicitly excludes `/events` and `/pty/` (src/server.ts:394). So un-credentialed upgrades now 401 pre-upgrade. Confirmed by test/server-auth.test.ts:215,243 (401 without credential) and :225 (cookie passes gate+origin), :250 (valid cookie + evil Origin still 403).

### Agent-callback regression — VERIFIED SAFE
- The PR routes ALL agent callbacks (hooks + build-queue) through the auth-exempt loopback ingress instead of provisioning a bearer token. `makeAgentIngressApp` builds with `{skipAuth:true}` (src/server.ts:5104); the ingress allowlist (src/server.ts:5085-5092) covers exactly the hooks/queue routes the base URL is baked for. `resolveSpawnBaseUrl` (src/service.ts:1147) sends trusted/standard spawns to `127.0.0.1:<ingressPort>` and only falls back to the gated main port when the ingress port is unknown — and `agentIngressState.port` is set at boot (src/index.ts:1761) before any spawn, so the fallback is the documented pathological-early-boot case, not a live regression. Confirmed against test/service.test.ts:4731 (uses ingress port) and the new fallback test at test/service.test.ts:4766.

### i18n parity — VERIFIED
- All 12 new keys (`login_*`, `settings_logout_*`, `feat_operator_auth_*`) exist in both ui/messages/en.json:1929-1941 and ui/messages/de.json:1929-1941. Keys used in Login.svelte, Settings.svelte, and feature-announcements.ts all resolve.

### Other checks
- Cookie crypto (HMAC + timingSafeEqual with length-guard, src/operator-auth.ts:90-102), conditional Secure (src/operator-auth.ts:168), fail-closed bootstrap (src/operator-auth.ts:195), and the 401 interceptor → `auth.unauthenticated` store (ui/src/lib/api.ts:71, ui/src/lib/auth.svelte.ts:5) all line up with the design and their tests. Docs (configuration.md, operating.md, external-task-api.md) accurately describe `SHEPHERD_PASSWORD`/`SHEPHERD_COOKIE_SECRET`/`SHEPHERD_TOKEN` as implemented in src/config.ts:333-337.

No blocking issues found; the previously-raised finding is genuinely addressed.

- ui/src/lib/api.ts: the 401 interceptor (`flagIfUnauthorized`) only fires on calls that route through `apiError`/`failed`/`getJson`; a handful of endpoints that check `r.ok` manually won't flip the `unauthenticated` store on a mid-session 401. The boot `/api/me` probe and the common write/read paths cover the real flows, so this is a UX edge (a stale-cookie 401 on an un-instrumented call won't auto-surface login) rather than a security gap — the route stays gated server-side either way.

PR #1082, opened 2026-06-25
feat(backup): automated hourly SQLite backups + restore + staleness alert (#1080)

### LINKED ISSUE #1080: Automated periodic backups for Shepherd's SQLite DB
## Problem

Shepherd's entire state lives in a single SQLite file at `~/.shepherd/shepherd.db` (`bun:sqlite`, `PRAGMA foreign_keys=ON`, default rollback journal — no WAL). **There is no backup, dump, or export anywhere in the codebase** (`src/store.ts:624`, `src/config.ts:8`). Worse, the data is already *ephemeral*: the daily sweep prunes sessions (30d / keep-250), signals (60d), reviewer-spawns + usage (90d) — deleted is gone forever.

**Threat model to protect against:** (1) DB corruption / bad on-open `ALTER` migration → roll back to a known-good snapshot; (2) accidental data loss (retention prune or buggy delete) → recover from a longer-lived archive. **Out of scope:** disk/host failure (so local backups are acceptable; no offsite requirement in v1).

## Design (decided via grilling)

External OS-level job, not in-process — decoupled from the single Bun event loop (sync I/O there freezes the live terminal), survives a crashed/wedged server, and matches the project's systemd-user-unit pattern.

| Aspect | Decision |
|---|---|
| **Mechanism** | New `shepherd-backup.timer` + `shepherd-backup.service` systemd **user** units running a script. |
| **Snapshot** | `bun` script (guaranteed present — no new binary) opens the DB **read-only** and runs `VACUUM INTO <temp>` (atomic, consistent on a live DB, compacted, no `-wal/-shm` artifacts). |
| **Integrity** | `PRAGMA integrity_check` on the raw snapshot **before** compression. Failing snapshot is discarded, never kept. |
| **Compression** | `Bun.gzipSync` → `shepherd-<ts>.db.gz` (bun built-in, no binary). |
| **Atomicity** | Write to temp + rename into place so a crashed run never leaves a half-file that rotation counts. |
| **Cadence** | `OnCalendar=hourly`. |
| **Destination** | `~/.shepherd/backups/`, override `SHEPHERD_BACKUP_DIR` (mirrors `SHEPHERD_DB`); `mkdir -p`. Same-disk caveat documented (disk-failure out of scope). |
| **Retention** | GFS tiered: **48 hourly + 14 daily + 8 weekly + 12 monthly** (~82 files, all compacted → small; counts tunable). Monthlies reach ~1yr, outlasting the 90d prune windows. |

## Observability (catch *silent* failure)

- Script writes `~/.shepherd/backups/.last-success` (timestamp) on success and appends to `backups/backup.log`; failures exit non-zero (systemd journal captures).
- Shepherd's **existing daily sweep** does a **read-only** `stat()` of the newest snapshot; if age > ~3h, emit a `signal` + web push ("backups stale / failing"). Catches hard run-failures **and** a silently-dead timer. Read-only stat keeps it compatible with the external design (no event-loop risk).
- Needs a new `SignalKind` for the staleness alert.

## Restore

`deploy/shepherd-restore.sh` — a backup you've never restored isn't a backup:
1. List available snapshots.
2. `systemctl --user stop shepherd` (DB must be unlocked).
3. Copy current DB aside → `shepherd.db.pre-restore-<ts>` (safety net).
4. `gunzip -c <snapshot> > shepherd.db`.
5. `PRAGMA integrity_check` on the restored file.
6. `systemctl --user start shepherd`.

Plus a runbook section in the docs.

## Install wiring (fresh installs **and** the existing live box)

`update.sh` today only *restarts* — it does **not** re-sync units. So:
- **`provision.ts`**: template & enable the backup units alongside `shepherd.service` for fresh installs (`WorkingDirectory`/`ExecStart` point at the actual checkout, same templating as the main unit).
- **`update.sh`**: add an **idempotent** unit-sync step — copy+template the `.timer`/`.service` into `~/.config/systemd/user`, `daemon-reload`, `enable --now shepherd-backup.timer`. The existing live box picks it up on the next `bun run update`; every future deploy self-heals.

## File-by-file plan

- `deploy/shepherd-backup.service` — `Type=oneshot`, `ExecStart=bun run scripts/backup.ts` (or templated path), env for `SHEPHERD_DB`/`SHEPHERD_BACKUP_DIR`.
- `deploy/shepherd-backup.timer` — `OnCalendar=hourly`, `Persistent=true` (run on boot if a window was missed).
- `scripts/backup.ts` (bun) — snapshot (read-only `VACUUM INTO` temp) → integrity_check → gzip → atomic rename → GFS rotation → write `.last-success` + log.
- `deploy/shepherd-restore.sh` — restore flow above.
- `deploy/provision.ts` — install + enable backup units on fresh installs.
- `deploy/update.sh` — idempotent unit-sync + `enable --now` timer.
- `src/index.ts` (daily sweep) — read-only staleness `stat()` → signal + push.
- New `SignalKind` + i18n keys (EN+DE) for the staleness alert message + web-push payload.
- `test/provision.test.ts` — assert the new units are written/enabled.

## Edges / notes

- **macOS / core-only** mode has no systemd → no backups there. Acceptable for v1 (primary deployment is Linux); document it.
- GFS tier counts and the ~3h staleness threshold are tunable defaults, not load-bearing.
- `Type=oneshot` + a single `.timer` means runs can't overlap; hourly VACUUM INTO of a small DB is far under an hour anyway.
- All snapshots in one dir; rotation classifies by timestamp into tiers (per period, keep newest).

## Unresolved (minor, non-blocking)

- Exact GFS counts (48/14/8/12)? — proposed defaults.
- Staleness threshold (3h)? — proposed.
- Reuse an existing `SignalKind` vs add a new one? — likely new.
- Dedup identical consecutive snapshots to save space? — probably out of scope for v1.

---
_Design reached via a `/grill-me` session; threat model = corruption + accidental data loss (not disk/host failure)._

scoop:
Closes #1080.

## What

Shepherd's entire state lives in one un-backed-up SQLite file. This adds an **external, hourly, OS-level backup** with GFS retention, a tested restore path, and in-app detection of silent backup failure. Threat model: DB corruption + accidental data loss (retention prune / buggy delete). Disk/host failure is out of scope (v1, local backups OK).

## How

- **`scripts/backup.ts`** runs out-of-process from the single Bun event loop (systemd-user `shepherd-backup.timer`, hourly): opens the DB **read-only**, `VACUUM INTO` a temp snapshot (atomic + consistent on a live DB, compacted, no `-wal/-shm`), `PRAGMA integrity_check` (a failing snapshot is **discarded**), `Bun.gzipSync`, **atomic rename**, then **GFS rotation** (48h / 14d / 8w / 12mo). Writes `.last-success` + `backup.log`. Pure `classifyRetention` is unit-tested.
- **`src/backup-paths.ts`** — pure, side-effect-free `resolveDbPath`/`resolveBackupDir` shared by the script **and** the server, so the backup dir can't drift and the script avoids `config.ts`'s eager `loadForgeMap`/`resolveNodeBin`.
- **`src/store.ts`** — `PRAGMA busy_timeout = 5000` so a concurrent server write **waits** for the backup's shared read lock instead of throwing `SQLITE_BUSY` on the event loop (verified: no busy_timeout/WAL existed anywhere before).
- **Install wiring** — `provision.ts` installs + enables the units and writes a `.backup-configured` marker; **`update.sh` idempotently re-syncs** the units every deploy, so the existing live box self-heals.
- **`deploy/shepherd-restore.sh`** — stops `shepherd` **and** the backup timer, copies the current DB aside, **removes stale `*-journal`/`*-wal`/`*-shm`** (else a crash-left journal replays into the restored file → re-corruption), gunzips, verifies `integrity_check = ok`, restarts.
- **Staleness alert** — `runDailySweep` does a read-only, marker-gated probe → guaranteed log line + durable `backup_stale` signal (new `SignalKind`, excluded from the learnings distiller) + web push (new push kind). A box broken from its first run is flagged; macOS/core-only (no marker) stays silent.
- **`docs/backups.md`** runbook.

## Phase-0 spike result

Confirmed `bun:sqlite` (1.3.14) accepts `VACUUM INTO ?` on a **read-only** connection against a live, server-held DB; snapshot `integrity_check = ok`; the live writer commits fine after. So the read-write fallback in the plan was **not** needed.

## Deviations (flagged per house rules)

- `backup_stale` has **no in-app toast/drawer** — `egress_drop`'s toast is a transient, session-scoped `toasts.info()` with no global-broadcast path for a host-level daily-sweep alert. v1 channels = guaranteed log line + durable signal + push; an in-app banner is a deliberate follow-up.
- **Detection latency ≈ 24h** (daily-sweep sampling), not the 3h staleness threshold — documented, not implied.

## Tests / verification

- `test/backup.test.ts` (new): GFS classifier bucketing + integrity-fail-discard + live-DB snapshot/rotate. `test/provision.test.ts`: backup units + marker written, timer enabled.
- Root `bun test ./test` (4857 pass), `bunx tsc --noEmit`, `bun run lint`, `cd ui && bun run check` (0 errors), fallow delta audit (✓ no new issues) — all green. End-to-end: ran the real script against a live DB, restored the `.gz`, integrity ok, 2000 rows intact.

`[no-feature-entry]` — server/deploy/docs only, no user-facing UI component or route; the sole user-facing text (push payload) is hand-synced EN+DE in `src/push.ts`.

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -234,6 +234,13 @@ main() {
   resolve_source
 
   cd "$SHEPHERD_DIR" || die "cannot cd into $SHEPHERD_DIR"
+  # Mirror the systemd units' EnvironmentFile=-%h/.shepherd/env so any SHEPHERD_DB /
+  # SHEPHERD_BACKUP_DIR override reaches provision via process.env — otherwise provision would
+  # write the .backup-configured marker to the DEFAULT dir while the server (which DOES read the
+  # env file) looks in the override dir, silently disabling staleness alerting (#1080).
+  set -a
+  [ -f "$HOME/.shepherd/env" ] && . "$HOME/.shepherd/env"
+  set +a
   note "handing off to deploy/provision.ts (from $SHEPHERD_DIR)"
   exec bun run deploy/provision.ts
 }
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -285,6 +285,10 @@ export function installService(
   run("mkdir", ["-p", resolveBackupDir(env)]);
   fileIO.write(backupConfiguredMarker(env), "shepherd-backup.timer enabled\n");
   run("systemctl", ["--user", "enable", "--now", "shepherd-backup.timer"]);
+  // Kick one backup immediately — the timer's first scheduled run is at the next hour boundary, so
+  // without this a fresh box has no `.last-success` until then. (The staleness probe also has a
+  // marker-age grace, so this is belt-and-suspenders against a spurious first alert. #1080)
+  run("systemctl", ["--user", "start", "shepherd-backup.service"]);
   // update.sh does deps → UI build → restart (starts the now-enabled unit) → health.
   log("building + starting via deploy/update.sh");
   run("bash", [join(repo, "deploy", "update.sh")], { env: buildEnv });
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -285,10 +285,10 @@ export function installService(
   run("mkdir", ["-p", resolveBackupDir(env)]);
   fileIO.write(backupConfiguredMarker(env), "shepherd-backup.timer enabled\n");
   run("systemctl", ["--user", "enable", "--now", "shepherd-backup.timer"]);
-  // Kick one backup immediately — the timer's first scheduled run is at the next hour boundary, so
-  // without this a fresh box has no `.last-success` until then. (The staleness probe also has a
-  // marker-age grace, so this is belt-and-suspenders against a spurious first alert. #1080)
-  run("systemctl", ["--user", "start", "shepherd-backup.service"]);
+  // No immediate `start shepherd-backup.service` here: on a fresh install the DB does not exist yet
+  // (the server is started below by update.sh), so a read-only snapshot would fail the oneshot and
+  // abort provision. update.sh runs its own GUARDED kick once deps are built, and the staleness
+  // probe's marker-age grace already covers the spurious-first-alert race. #1080
   // update.sh does deps → UI build → restart (starts the now-enabled unit) → health.
   log("building + starting via deploy/update.sh");
   run("bash", [join(repo, "deploy", "update.sh")], { env: buildEnv });
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -21,6 +21,7 @@ import { join } from "node:path";
 import { BUN_MIN_VERSION, NODE_MIN_VERSION, HERDR_MIN_VERSION } from "../src/config";
 import { compareSemver } from "../src/herdr-update";
 import { autoFixCommandFor } from "../src/remediations";
+import { resolveBackupDir, backupConfiguredMarker } from "../src/backup-paths";
 
 // ── pure types + data ─────────────────────────────────────────────────────────
 
@@ -267,11 +268,23 @@ export function installService(
   // hardcoded %h/Work/shepherd. Default repo (~/Work/shepherd) ⇒ identical output.
   const unitSrc = fileIO.read(join(repo, "deploy", "shepherd.service"));
   fileIO.write(join(unitDir, "shepherd.service"), templateUnit(unitSrc, repo));
+  // Hourly backup units (#1080), installed alongside the main service. The .service carries a
+  // WorkingDirectory to template; the .timer has none, so it copies verbatim.
+  const backupSvc = fileIO.read(join(repo, "deploy", "shepherd-backup.service"));
+  fileIO.write(join(unitDir, "shepherd-backup.service"), templateUnit(backupSvc, repo));
+  const backupTimer = fileIO.read(join(repo, "deploy", "shepherd-backup.timer"));
+  fileIO.write(join(unitDir, "shepherd-backup.timer"), backupTimer);
   run("systemctl", ["--user", "daemon-reload"]);
   const user = env.USER;
   if (!user) throw new Error("cannot enable-linger: $USER is not set");
   run("loginctl", ["enable-linger", user]);
   run("systemctl", ["--user", "enable", "shepherd"]);
+  // Mark this host as backup-EXPECTED before enabling the timer: the server's staleness check
+  // treats marker-present + no recent success as a failure, so a box whose backup is broken from
+  // the very first run is flagged (a no-marker host, e.g. macOS/core-only, stays silent).
+  run("mkdir", ["-p", resolveBackupDir(env)]);
+  fileIO.write(backupConfiguredMarker(env), "shepherd-backup.timer enabled\n");
+  run("systemctl", ["--user", "enable", "--now", "shepherd-backup.timer"]);
   // update.sh does deps → UI build → restart (starts the now-enabled unit) → health.
   log("building + starting via deploy/update.sh");
   run("bash", [join(repo, "deploy", "update.sh")], { env: buildEnv });
--- a/deploy/shepherd-backup.service
+++ b/deploy/shepherd-backup.service
@@ -1,0 +1,15 @@
+[Unit]
+Description=Shepherd — hourly SQLite backup (VACUUM INTO → integrity_check → gzip → GFS rotation)
+Documentation=https://github.com/erwins-enkel/shepherd
+
+[Service]
+Type=oneshot
+WorkingDirectory=%h/Work/shepherd
+ExecStart=%h/.bun/bin/bun run scripts/backup.ts
+
+# bun lives in ~/.bun/bin; keep the same resolvable PATH as shepherd.service
+Environment=PATH=%h/.bun/bin:%h/.local/bin:/usr/local/bin:/usr/bin:/bin
+# optional overrides (SHEPHERD_DB, SHEPHERD_BACKUP_DIR) — shared with the main unit; file may be absent
+EnvironmentFile=-%h/.shepherd/env
+
+# Output goes to journald: `journalctl --user -u shepherd-backup`
--- a/deploy/shepherd-backup.timer
+++ b/deploy/shepherd-backup.timer
@@ -1,0 +1,13 @@
+[Unit]
+Description=Shepherd — hourly SQLite backup timer
+Documentation=https://github.com/erwins-enkel/shepherd
+
+[Timer]
+OnCalendar=hourly
+# Run on boot if the host was off when a window was due (a missed hour is still worth catching up).
+Persistent=true
+# A oneshot service + a single timer means runs can never overlap; an hourly VACUUM INTO of a small
+# DB finishes far under an hour.
+
+[Install]
+WantedBy=timers.target
--- a/deploy/shepherd-restore.sh
+++ b/deploy/shepherd-restore.sh
@@ -21,6 +21,13 @@ die() {
 REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
 cd "$REPO"
 
+# Mirror the systemd units' EnvironmentFile=-%h/.shepherd/env so a SHEPHERD_DB / SHEPHERD_BACKUP_DIR
+# override resolves to the SAME paths the server + backup script use — else we'd restore to the
+# wrong DB / list the wrong backup dir (#1080).
+set -a
+[ -f "$HOME/.shepherd/env" ] && . "$HOME/.shepherd/env"
+set +a
+
 # Resolve DB + backup dir through the shared resolver so this never drifts from the backup script.
 DB="$(bun -e 'import {resolveDbPath} from "./src/backup-paths"; process.stdout.write(resolveDbPath())')"
 BACKUP_DIR="$(bun -e 'import {resolveBackupDir} from "./src/backup-paths"; process.stdout.write(resolveBackupDir())')"
--- a/deploy/shepherd-restore.sh
+++ b/deploy/shepherd-restore.sh
@@ -1,0 +1,70 @@
+#!/usr/bin/env bash
+# Restore Shepherd's SQLite DB from a backup snapshot (#1080).
+#
+#   deploy/shepherd-restore.sh              # list snapshots, then pick one
+#   deploy/shepherd-restore.sh <snapshot>   # restore a specific shepherd-*.db.gz
+#
+# A backup you've never restored isn't a backup. This stops the service (and the backup timer, so a
+# mid-restore firing can't open the DB), copies the current DB aside as a safety net, removes any
+# hot rollback-journal/WAL sidecars (a crash-left journal would otherwise be REPLAYED into the
+# freshly-restored file → re-corruption), gunzips the snapshot into place, verifies its integrity,
+# and restarts.
+set -euo pipefail
+
+note() { printf '\033[36m▸ %s\033[0m\n' "$*"; }
+warn() { printf '\033[33m! %s\033[0m\n' "$*"; }
+die() {
+  printf '\033[31m✗ %s\033[0m\n' "$*" >&2
+  exit 1
+}
+
+REPO="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+cd "$REPO"
+
+# Resolve DB + backup dir through the shared resolver so this never drifts from the backup script.
+DB="$(bun -e 'import {resolveDbPath} from "./src/backup-paths"; process.stdout.write(resolveDbPath())')"
+BACKUP_DIR="$(bun -e 'import {resolveBackupDir} from "./src/backup-paths"; process.stdout.write(resolveBackupDir())')"
+
+SNAP="${1:-}"
+if [[ -z "$SNAP" ]]; then
+  note "available snapshots in $BACKUP_DIR (newest first):"
+  mapfile -t SNAPS < <(ls -1 "$BACKUP_DIR"/shepherd-*.db.gz 2>/dev/null | sort -r)
+  [[ ${#SNAPS[@]} -gt 0 ]] || die "no snapshots found in $BACKUP_DIR"
+  for i in "${!SNAPS[@]}"; do printf '  [%d] %s\n' "$i" "$(basename "${SNAPS[$i]}")"; done
+  read -r -p "restore which index? " IDX
+  [[ "$IDX" =~ ^[0-9]+$ && -n "${SNAPS[$IDX]:-}" ]] || die "invalid selection"
+  SNAP="${SNAPS[$IDX]}"
+fi
+[[ -f "$SNAP" ]] || die "snapshot not found: $SNAP"
+
+warn "About to OVERWRITE $DB with $(basename "$SNAP")."
+read -r -p "Type 'restore' to proceed: " CONFIRM
+[[ "$CONFIRM" == "restore" ]] || die "aborted"
+
+# Stop the server AND the backup timer so nothing holds/opens the DB mid-restore.
+note "stopping shepherd + backup timer"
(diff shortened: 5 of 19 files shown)


scoop:
## Summary

This is a well-built, well-documented implementation of the issue's design: out-of-process `VACUUM INTO` → `integrity_check` → gzip → atomic rename → GFS rotation (`scripts/backup.ts`), a read-only staleness probe in the daily sweep (`src/index.ts`), the new `backup_stale` `SignalKind` wired through types/push/distiller, systemd units, install/update wiring, a restore script, and tests. The pure core (`classifyRetention`, `formatTs`/`parseTs`/`weekKey`) is cleanly factored and tested.

I verified the moving parts that the change depends on:
- `resolveDbPath` mirrors `src/config.ts:8` exactly (`env.SHEPHERD_DB ?? "${HOME}/.shepherd/shepherd.db"`) — verified against `src/config.ts:8` and `src/backup-paths.ts:499`.
- `backup_stale` is added consistently everywhere it must be: `SignalKind` (`src/types.ts:776`), `PushPayload["kind"]` (`src/push.ts:638`), `NotifyInput["kind"]` (`src/push.ts:657`), `KIND_CATEGORY` (`src/push.ts:647`), `REDUCED_ALLOWED` (`src/push.ts:679`), `buildPayload` switch (`src/push.ts:342`), and `NON_LEARNING_SIGNAL_KINDS` (`src/distiller.ts:534`).
- en + de push copy parity: `backupStaleTitle`/`backupStaleBody` exist in both locales — verified `src/push.ts:688-692` (en) and `src/push.ts:700-704` (de).
- The `learningsParts` refactor preserves the prior `learnings_retired`/`learnings_trialed` behavior via the case fall-through — verified `src/push.ts:712-716` and `src/push.ts:326-337`.
- `signals.repoPath` has no FK constraint (`src/store.ts:841-843`), so the `__host__` sentinel insert is safe; and distill only `consider()`s real repos from `listRepos(config.repoRoot)` (`src/index.ts:1491`), so `__host__` never spawns a distill run.
- `templateUnit` rewrites the `WorkingDirectory=` line (`deploy/provision.ts:113-115`); the backup `.service` carries one (`deploy/shepherd-backup.service:7`) and the `.timer` has none, matching how provision/update treat each.

## Blocking finding

**Deploy scripts resolve backup/DB paths from the shell environment, not `~/.shepherd/env` — so the documented `SHEPHERD_DB` / `SHEPHERD_BACKUP_DIR` overrides silently break restore and staleness alerting.**

`docs/backups.md:41` tells operators overrides "go in `~/.shepherd/env` (read by both `shepherd.service` and `shepherd-backup.service`)", and both units load it via `EnvironmentFile=-%h/.shepherd/env` (`deploy/shepherd.service:20`, `deploy/shepherd-backup.service:13`). The running server therefore honors `SHEPHERD_BACKUP_DIR` because `resolveBackupDir(process.env)` reads it (`src/backup-paths.ts:505`, consumed at `src/index.ts:580`).

But the deploy *shell* scripts never source `~/.shepherd/env` (confirmed: no `source`/`. …/env` anywhere in `deploy/`). They resolve paths with a bare `bun -e`:
- `deploy/update.sh:175` computes `BACKUP_DIR` and writes the `.backup-configured` marker to it (`deploy/update.sh:177`).
- `deploy/provision.ts:31-32` likewise `mkdir`s and writes the marker into `resolveBackupDir(env)` from the provision process env.
- `deploy/shepherd-restore.sh:107-108` resolves `DB` and `BACKUP_DIR` the same way.

With an override set only in `~/.shepherd/env` (exactly as the docs instruct):
1. The `.backup-configured` marker lands in the **default** `~/.shepherd/backups`, but the server checks the **overridden** dir → marker absent → `checkBackupStaleness` returns early (`src/index.ts:580-582`) → staleness alerting is silently disabled. That defeats the issue's central "catch *silent* failure" goal.
2. `shepherd-restore.sh` resolves `DB` to the **default** `~/.shepherd/shepherd.db` and gunzips the snapshot there, while the server keeps running on the **overridden** DB. The restore reports success but the live DB is untouched (or a non-active file is overwritten) — a data-safety hazard on a documented configuration.

Fix: have `provision.ts`, `update.sh`, and `shepherd-restore.sh` load `~/.shepherd/env` (e.g. `set -a; [ -f ~/.shepherd/env ] && . ~/.shepherd/env; set +a` before the `bun -e` resolves) so the deploy-time path resolution matches what the units actually use. The default-config path (no override) is correct as written; only the override path is broken.

- src/index.ts: a freshly provisioned/restarted host can emit one spurious `backup_stale` alert. `enable --now shepherd-backup.timer` (`deploy/provision.ts:33`) starts the *timer*, whose first `OnCalendar=hourly` fire is the next hour boundary (no past window for `Persistent` to catch), so `.last-success` is absent when the boot+10s daily sweep runs (`src/index.ts:1545`, `setTimeout(...,10_000)`); `ageMs===null` → stale=true fires before the first backup ever runs. Self-heals after the first successful hourly backup.
- scripts/backup.ts: a crash between writing `${outGz}.tmp` and the rename (`compress`, line 440-442) leaves a `shepherd-<ts>.db.gz.tmp` that `FILE_RE` (line 330) never matches, so `rotate` never reaps it — a slow orphan-`.tmp` leak. The snapshot `.db.tmp` is cleaned in the `finally` (line 464-466), but the gz `.tmp` is not.

- (none)

scoop:
## Re-review of prior points

All three prior-revision points are addressed by this diff:

1. **Source `~/.shepherd/env` before resolving paths** — `deploy/install.sh:13-15` sources it before `exec bun run deploy/provision.ts` (so provision sees overrides via `process.env`), `deploy/update.sh:208` sources it in a subshell before `resolveBackupDir`, and `deploy/shepherd-restore.sh:131-133` sources it before resolving `DB`/`BACKUP_DIR`. Verified all three.
2. **Reap orphaned `*.gz.tmp`** — `scripts/backup.ts:374` adds `TMP_RE` matching both `.shepherd-<ts>.db.tmp` and `shepherd-<ts>.db.gz.tmp`; `rotate` reaps them at `scripts/backup.ts:495`. Verified.
3. **Spurious first `backup_stale` alert** — `src/index.ts:637` adds a marker-age grace window (`if (markerAgeMs < BACKUP_STALENESS_MS) return;`), plus provision/update kick an immediate backup. Verified.

## New blocking finding

**`deploy/provision.ts:291` — unguarded immediate `start shepherd-backup.service` breaks fresh installs.**

Flow on a genuinely fresh Linux host:
- `installService` runs `systemctl --user start shepherd-backup.service` at `deploy/provision.ts:291`, **before** the server is ever built/started (that only happens via `update.sh` at `deploy/provision.ts:294`).
- At this point `~/.shepherd/shepherd.db` does not exist — the DB is created by the server's `SessionStore` on first boot, which hasn't run.
- `backup.ts` opens the DB read-only (`scripts/backup.ts:460` `new Database(dbPath, { readonly: true })`); a read-only open of a missing file throws (`SQLITE_CANTOPEN`), so `main()` exits non-zero.
- The `Type=oneshot` service therefore fails activation, so `systemctl --user start` returns non-zero.
- `defaultRunner` throws on any non-zero exit (`deploy/provision.ts:160-162`), so provision aborts here — before `update.sh` runs — leaving the server never built or started.

The sibling call in `deploy/update.sh:215` correctly guards this with `|| warn` (it also runs before the server start in that script and would hit the same missing-DB case). `provision.ts` does not. This is the same guard present on one path and missing on the parallel one, and it is reachable on the fresh-install path that executes today.

Fix: guard the immediate kick (e.g. don't let a non-zero `systemctl start` abort provision), or drop it entirely — the marker-age grace window in `src/index.ts:637` already prevents the spurious-first-alert race that this immediate start was belt-and-suspenders for, so removing it loses nothing on fresh installs.

## Other verification

- `SignalKind` `backup_stale` added in `src/types.ts:542`, consumed in `src/store.ts:2664` (no FK on `repoPath`, so the `__host__` sentinel + null `sessionId` insert is safe), and excluded from learnings mining in `src/distiller.ts`. Verified.
- Push parity: `backup_stale` added to `PushPayload.kind`, `NotifyInput.kind`, `KIND_CATEGORY`, `REDUCED_ALLOWED`, and the `buildPayload` switch; EN strings at `src/push.ts:742-746` and DE strings at `src/push.ts:754-758` both present (`backupStaleTitle`/`backupStaleBody`). Locale parity holds. Verified.
- `push.notify` with `sessionId:""` is safe: `cooldownKey:"backup_stale"` overrides the default key (`src/push.ts:410`) and category `agent` gating applies; no DB lookup on sessionId.

Could not exercise systemd/bun:sqlite at runtime (read-only review), but the missing-file read-only-open failure and `systemctl start` propagating a oneshot failure are well-established behaviors.

scoop:
## Re-review of prior points

**Prior point #1 — NOT addressed (re-raised, blocking).**

The immediate `systemctl --user start shepherd-backup.service` at `deploy/provision.ts:291` is still unguarded, and `defaultRunner` still throws on any non-zero exit (`deploy/provision.ts:155-163`, specifically the `throw new Error(...)` at 160-162). I traced the failure path concretely:

- On a fresh install the DB file does not exist when line 291 runs — provision creates `~/.shepherd` (`deploy/provision.ts:265`) but not the DB; the server that creates it is only started later, via `update.sh` invoked at `deploy/provision.ts:294`.
- `scripts/backup.ts` `main()` (`scripts/backup.ts:169-176`) calls `snapshot()` (`scripts/backup.ts:130-138`), which does `new Database(dbPath, { readonly: true })`. A read-only SQLite open of a missing file raises `SQLITE_CANTOPEN`; `main()` rejects → `process.exit(1)` (`scripts/backup.ts:194-198`).
- A `Type=oneshot` service whose `ExecStart` exits non-zero makes `systemctl --user start` return non-zero → `defaultRunner` throws → provision aborts before `update.sh` ever starts the server.
- The sibling kick in `deploy/update.sh:70` IS guarded (`systemctl --user start shepherd-backup.service || warn "initial backup run did not start"`); provision must do the same, or drop the immediate kick entirely — the marker-age grace window in `src/index.ts:697` (`if (markerAgeMs < BACKUP_STALENESS_MS) return;`) already covers the spurious-first-alert race that the kick is belt-and-suspenders for.

Verified against: `deploy/provision.ts:291`, `deploy/provision.ts:160-162`, `deploy/provision.ts:294`, `deploy/provision.ts:265`, `scripts/backup.ts:130-138`, `scripts/backup.ts:194-198`, `deploy/update.sh:70`, `src/index.ts:697`. (Note: the prior note cited the guarded sibling as `update.sh:215`; the actual line in the current tree is `deploy/update.sh:70` — same call, same guard, substance unchanged.)

## Other verification (no further blocking issues found)

- `SignalKind` parity: `backup_stale` added at `src/types.ts:542` and used consistently in `src/distiller.ts:641-644`, `src/index.ts:715`, `src/push.ts:51,71,98(payload kind union),167(REDUCED_ALLOWED),356`. No mismatched spellings.
- i18n parity: `backupStaleTitle`/`backupStaleBody` exist in BOTH `en` (`src/push.ts:802-806`) and `de` (`src/push.ts:814-818`) catalogs; `buildPayload` `case "backup_stale"` (`src/push.ts:356-361`) consumes them with the `staleHours ?? null` arg matching the `(h: number | null)` body signature.
- Path resolver is shared between writer and readers: `src/backup-paths.ts` is imported by `scripts/backup.ts:21`, `deploy/provision.ts:24`, `src/index.ts:49`, and resolved via `bun -e` in `deploy/update.sh:208` and `deploy/shepherd-restore.sh:31-32` — no default drift.
- `learningsParts` refactor (`src/push.ts:253-258`) preserves the prior per-kind title/body for `learnings_retired`/`learnings_trialed` (`src/push.ts:328-330`).

- (none)

scoop:
Re: the unguarded `systemctl --user start shepherd-backup.service` in provision.ts — this was already fixed in commit 9c779549 ("drop provision's unguarded immediate backup kick"), which is current branch HEAD. The call is **removed**, not guarded: on a fresh install the DB doesn't exist yet (the server is started later by update.sh), so the kick could only ever fail the oneshot and abort provision; it was also redundant since update.sh runs its own `|| warn`-guarded kick and the marker-age grace in src/index.ts covers the spurious-first-alert race. provision.ts:288-291 is now a comment documenting the deliberate absence; the only remaining systemctl call there is `enable --now shepherd-backup.timer` (line 287). The review appears to have run against 71987385 (the prior commit), where the live `run(...)` was at line 291 — please re-evaluate against 9c779549.

scoop (commented):
## Re-review of #1080 (automated SQLite backups)

### Prior finding — RESOLVED
**deploy/provision.ts: immediate unguarded `systemctl start shepherd-backup.service`.** Confirmed fixed against HEAD (9c779549). `installService` now ends with `run("systemctl", ["--user", "enable", "--now", "shepherd-backup.timer"])` at provision.ts:287 followed by a comment (288-291) documenting the deliberate removal of the immediate kick. There is no live `start shepherd-backup.service` call in provision. `test/provision.test.ts:1074-1076` asserts its absence. The guarded kick survives in deploy/update.sh:215 (`|| warn`). Not re-raised.

### Verification performed
- **SignalKind `backup_stale`** added at src/types.ts:542 and wired consistently into every consumer: src/push.ts PushPayload kind, NotifyInput kind, `KIND_CATEGORY` (backup_stale→"agent"), `REDUCED_ALLOWED` (push.ts:789-794), and `buildPayload` (push.ts:336-339, citing t.backupStaleTitle/Body). src/distiller.ts:641-644 adds it to `NON_LEARNING_SIGNAL_KINDS`. No exhaustive UI mapping over signal kinds exists — the only egress reference (ui/src/lib/store.svelte.ts:340) is a session-event toast, not a per-kind signal label, so no UI catalog gap is introduced.
- **Locale parity**: `backupStaleTitle`/`backupStaleBody` present in both en (src/push.ts:168-169) and de (src/push.ts:215-216), the two catalogs `NOTIFY_TEXT` maintains.
- **Host-global signal**: `signals` table (src/store.ts:841-843) has `repoPath TEXT NOT NULL` with no FK and nullable `sessionId`, so the `__host__` sentinel + null session insert from src/index.ts:1463 is valid. The distiller only considers on-disk repos via `listRepos` (src/index.ts:1494), so `__host__` never becomes a distill target.
- **Path resolution single-sourced**: src/backup-paths.ts is imported by scripts/backup.ts, src/index.ts (staleness probe), deploy/provision.ts, deploy/update.sh, and deploy/shepherd-restore.sh. install.sh:13-15, update.sh:208, and restore.sh:131-133 all source `~/.shepherd/env` so `SHEPHERD_DB`/`SHEPHERD_BACKUP_DIR` overrides resolve to the same dir the server reads. Marker filename `.backup-configured` matches `backupConfiguredMarker`.
- **Staleness grace window** (src/index.ts:685-728): marker-mtime grace (< BACKUP_STALENESS_MS) suppresses the first-alert race; update.sh:210 writes the marker only if absent, so the grace is one-time and a genuinely-broken backup is still flagged after the window. `BUSY_TIMEOUT_MS=5000` (scripts/backup.ts:424) equals the server's `PRAGMA busy_timeout = 5000` (src/store.ts:878).
- **backup.ts flow**: snapshot → verifyIntegrity (throws on non-`ok`, discarding the temp via the `finally rmSync`) → compress (temp + atomic rename) → rotate → write `.last-success` + append `backup.log`. A failed integrity check skips compress/rotate and exits non-zero with no `.last-success` written. `import.meta.main` guard keeps `main()` from running when the module is imported by test/backup.test.ts.

Could not verify the `bun:sqlite` claim that `VACUUM INTO ?` is accepted on a read-only connection against a live SQLite engine here (no runtime exercised in review); it is asserted in code comments and exercised by test/backup.test.ts, which CI runs.

No blocking problems and no nits to raise.

PR #1084, opened 2026-06-25
fix(buildqueue): truthful queue header — auto/operator approval + run-state

scoop:
## Problem

The build-queue panel header **always** read `approved · running` — a single hardcoded message string (`buildqueue_approved_header`), rendered whenever `approved && steps.length > 0`. Two ways this lied:

- **"running" was a literal** — never derived from step state. It showed even when every step was still `pending` (queue authored, nothing started) or when all steps were already `done`.
- **"approved" couldn't distinguish** an operator clicking **Approve & run** from autopilot's silent pre-approval at spawn (`shouldPreApproveBuildQueue`), which happens independently of the plan gate. So an autopilot session that hadn't passed the plan gate still showed a flat "approved".

## Fix

Header now reads `<approval> · <run-state>`, both derived from real state:

| Situation | Header |
| --- | --- |
| Autopilot pre-approved, all steps pending | `auto-approved · queued` |
| Autopilot, ≥1 step active/done | `auto-approved · running` |
| Operator clicked Approve & run | `approved · running` |
| All steps done/skipped | `approved · done` |
| Legacy approved queue (unknown kind) | `approved · <run-state>` |

**Server** — new optional `approvalKind?: "auto" \| "operator"` on `BuildQueue`, persisted in `build_queue_state` (additive nullable column + PRAGMA-guarded migration, safe on the live DB, no backfill). `service.ts` auto pre-approval writes `"auto"`; the `/queue/approve` endpoint writes `"operator"`. The field rides the existing `/approve` response and `queue:update` WS payload.

**UI** — `BuildQueuePanel.svelte` derives a **total** run-state from step statuses (`done` = all done/skipped; else `running` = any active/done; else `queued`) and the approval label from `approvalKind`. Colors are semantic tokens per design-system rule 4: running = amber, queued = faint, **done = `--status-done` (slate), never green** (a finished queue is finished-but-parked, not actionable-complete). Swapped the obsolete `buildqueue_approved_header` for five new EN+DE keys.

`approvalKind` is **optional**, not required-nullable: the store is the only real producer, `undefined` cleanly models legacy/unknown, and existing `BuildQueue` test fakes compile untouched.

## Tests

- Server: `approvalKind` round-trips through `getBuildQueue`, the `/approve` response, and the `queue:update` event; `"auto"` via store, `undefined` when unapproved.
- UI: new header-state tests for `auto·queued`, `auto·running`, `operator·done`, and operator-vs-auto; rewrote the assertion that referenced the deleted key.

Full gate set green locally (root lint/tsc/`bun test`; UI `check:i18n`/`check`/`test`; branch-hygiene/feature-catalog/glossary). Scope is a label-correctness **fix** — no new capability, so no feature-announcement/glossary entry (`fix(...)` subjects keep the feature-catalog gate disarmed).

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -2411,7 +2411,7 @@ async function postBuildStepStatus(
 
 // POST /api/sessions/:id/queue/approve — human gate: approve + steer the agent.
 function approveBuildQueue(deps: AppDeps, id: string): Response {
-  deps.store.setBuildQueueApproved(id, true);
+  deps.store.setBuildQueueApproved(id, true, "operator");
   const q = deps.store.getBuildQueue(id);
   deps.events?.emit("queue:update", q);
   deps.service.reply(id, APPROVE_STEER); // best-effort; ignore boolean return
--- a/src/service.ts
+++ b/src/service.ts
@@ -1618,7 +1618,7 @@ export class SessionService {
       // Autopilot sessions are pre-approved so the agent can begin executing immediately
       // after authoring the queue without waiting for a human gate that will never come.
       if (shouldPreApproveBuildQueue(repoConfig, session, input.research))
-        this.deps.store.setBuildQueueApproved(sessionId, true);
+        this.deps.store.setBuildQueueApproved(sessionId, true, "auto");
       this.scheduleRefine(session, herdSlug);
       this.#maybeRegisterTrain(session, input);
       return session;
--- a/src/store.ts
+++ b/src/store.ts
@@ -879,7 +879,13 @@ export class SessionStore implements CapStore, CreditStore {
       `CREATE INDEX IF NOT EXISTS build_queue_steps_session ON build_queue_steps (sessionId, position)`,
     );
     this.db.run(`CREATE TABLE IF NOT EXISTS build_queue_state (
-      sessionId TEXT PRIMARY KEY, approved INTEGER NOT NULL DEFAULT 0, updatedAt INTEGER NOT NULL)`);
+      sessionId TEXT PRIMARY KEY, approved INTEGER NOT NULL DEFAULT 0, approvalKind TEXT, updatedAt INTEGER NOT NULL)`);
+    // migrate build_queue_state rows that predate the approvalKind column (legacy rows NULL = unknown kind)
+    const bqStateCols = this.db.query(`PRAGMA table_info(build_queue_state)`).all() as {
+      name: string;
+    }[];
+    if (!bqStateCols.some((c) => c.name === "approvalKind"))
+      this.db.run(`ALTER TABLE build_queue_state ADD COLUMN approvalKind TEXT`);
     // One stamp per workflow-protocol comment posted on a session's backlog issue
     // (issue-log: `waiting:<pr>` / `merged:<pr>`), so each transition comments exactly
     // once per PR across restarts and CI flaps.
@@ -3754,8 +3760,8 @@ export class SessionStore implements CapStore, CreditStore {
       status: string;
     }[];
     const state = this.db
-      .query(`SELECT approved FROM build_queue_state WHERE sessionId = ?`)
-      .get(sessionId) as { approved: number } | null;
+      .query(`SELECT approved, approvalKind FROM build_queue_state WHERE sessionId = ?`)
+      .get(sessionId) as { approved: number; approvalKind: string | null } | null;
     const steps: BuildStep[] = rows.map((r) => ({
       id: r.id,
       title: r.title,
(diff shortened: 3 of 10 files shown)


scoop (commented):
## Summary

The PR correctly diagnoses and fixes the reported problem: the build-queue header was **static text**. The old `buildqueue_approved_header` resolved to the literal string `"approved · running"` (en.json) / `"freigegeben · läuft"` (de.json) and was rendered verbatim whenever `approved && steps.length > 0`, so it always claimed "running" regardless of actual step status, and never distinguished autopilot pre-approval from an operator gate.

The replacement derives two independent axes:
- **approval kind** — `auto` (autopilot pre-approval at spawn) vs `operator` (human clicked Approve & run), now persisted via a new `approvalKind` column.
- **run state** — `queued` / `running` / `done` derived from the steps' actual statuses.

This is the right fix for the task: an auto-approved queue with all-`pending` steps now truthfully reads `auto-approved · queued` instead of `approved · running`.

## Verification

- **i18n parity**: the 5 new keys (`buildqueue_approval_auto`, `buildqueue_approval_operator`, `buildqueue_run_queued`, `buildqueue_run_running`, `buildqueue_run_done`) exist in BOTH catalogs (en.json:1940-1944, de.json:1940-1944), and the removed `buildqueue_approved_header` has no remaining references anywhere in the tree.
- **CSS tokens**: `bqp-run-running` → `--color-amber` (app.css:21), `bqp-run-queued` → `--color-faint` (app.css:20), `bqp-run-done` → `--status-done` (app.css:116, itself `var(--color-slate)`). All three modifier classes (`.bqp-run-{done,running,queued}`) are defined and `runState` only ever produces those three values, so the class array `["bqp-approved", `bqp-run-${runState}`]` always resolves to a real selector.
- **runState logic** covers all four `BuildStepStatus` values (`pending|active|done|skipped`, ui/src/lib/types.ts:3): `allResolved` (every step done/skipped) → `done`; else `anyRunning` (any active/done) → `running`; else `queued`. The header itself is still gated on `approved`, so an un-gated session shows nothing — directly addressing the task.
- **Store/migration**: `setBuildQueueApproved` forces `approvalKind` to null when `approved=false` (store.ts:3925), the named-column INSERT is order-independent of the CREATE TABLE column order, and the `PRAGMA table_info` ALTER guard migrates legacy DBs. `approvalKind` is omitted from the returned object when falsy, matching the optional `approvalKind?` type (src/types.ts:492, ui/src/lib/types.ts:17).
- **Callers**: the new third param is optional; all existing `setBuildQueueApproved` call sites (test/build-queue-store.test.ts) still compile. server.ts passes `"operator"`, service.ts passes `"auto"`, consistent with the type union.

## Non-blocking nit

- `ui/src/lib/components/BuildQueuePanel.svelte`: the derived `anyRunning` is true when any step is `done` (not just `active`), so a queue whose last finished step is `done` with later steps still `pending` (agent paused between steps) reads `running`. This is a reasonable best-effort given there is no `failed`/`stalled` status, but the name `anyRunning` slightly overstates it — consider `anyStarted` for clarity. Cosmetic only.

scoop (commented):
## Summary

The PR replaces the static `buildqueue_approved_header` ("approved · running") with a header derived from the queue's actual state: an approval label (`auto` vs `operator`, defaulting to operator for legacy/undefined kind) plus a run-state label (`queued`/`running`/`done`). This directly fixes the reported problem — the build-queue header is no longer static text.

## Verification

- **Prior finding #1 (rename `anyRunning`→`anyStarted`)**: addressed — `ui/src/lib/components/BuildQueuePanel.svelte:113` now declares `anyStarted`, and commit `17d0fab9` performs exactly that rename. Not re-raised.
- **Derived run-state logic** (BuildQueuePanel.svelte:111-114): `allResolved` (all done/skipped) → `done`; else `anyStarted` (active/done) → `running`; else `queued`. Correct for the edge cases the tests cover.
- **CSS tokens**: every token used resolves to a real definition — `--status-done` (verified ui/src/app.css:116), `--color-slate` (ui/src/app.css:25), `--color-faint` (ui/src/app.css:20), `--color-amber` present. The `bqp-run-{done,running,queued}` modifier classes (BuildQueuePanel.svelte:347-355) cover every value `runState` can take.
- **i18n parity**: `buildqueue_approved_header` removed from both catalogs with no remaining references (grep: 0 matches). The 5 new keys (`buildqueue_approval_auto`, `buildqueue_approval_operator`, `buildqueue_run_queued`, `buildqueue_run_running`, `buildqueue_run_done`) exist in BOTH ui/messages/en.json and ui/messages/de.json (5 each).
- **Store / migration**: `approvalKind` added to `build_queue_state` CREATE TABLE plus a PRAGMA-guarded `ALTER TABLE` for legacy DBs (src/store.ts:882-888); read paths in `getBuildQueue` and `listBuildQueues` spread the field only when truthy, keeping it optional and back-compatible. `setBuildQueueApproved` nulls the kind when `approved=false`.
- **Response/event shape**: `approveBuildQueue` (src/server.ts:2413-2418) sets kind `"operator"`, returns `getBuildQueue(id)`, and emits `queue:update` with the same object — matching the new API tests.

No blocking issues found and the previously-raised nit is resolved.

PR #1085, opened 2026-06-25
fix(ingress): pin agent-ingress port so live sessions survive restarts (#1083)

### LINKED ISSUE #1083: Stop-hook 401 for in-flight sessions after an operator-auth deploy: re-inject hook URL on reconnect
## Symptom

Sessions that were **live across the operator-auth deploy** (#1079/PR #1081) now fail every Stop hook:

```
Stop hook error: HTTP 401 from http://127.0.0.1:7330/api/sessions/<id>/hooks
```

## Root cause

The hook ingest endpoint (`POST /api/sessions/<id>/hooks`) is auth-exempt **only on the separate `serveAgentIngress` listener** — a loopback-only listener built with `skipAuth` plus a strict route allowlist (`src/server.ts` ~5086), bound to an **ephemeral port** chosen at server startup (`serveAgentIngress(appDeps)`, `port = 0`). The agent's hook URL is baked at spawn to point at that port.

A session spawned **before** the auth/ingress wiring (or before this server instance started) has its hook URL frozen to the **main port 7330**, which the operator-auth gate now protects. So its Claude-native http-hook client POSTs to 7330 → **401**. New sessions bake the current skipAuth ingress port and work fine.

This is a one-time cutover artifact, but it does **not self-heal**: the hook command lives in the running Claude process's settings, and nothing re-points it. It clears only when the session is re-spawned/restarted. More generally, because the ingress port is **ephemeral and rotates on every server restart**, any in-flight session loses its hook channel across any deploy — where, pre-auth, a hook pointed at the stable 7330 survived a restart.

## Proposed fix (one of)

1. **Re-inject the hook URL on (re)attach/reconnect** — when Shepherd re-attaches to a live session, rewrite its hook settings to the *current* `agentIngress.port`. Closes the channel-loss-on-restart class generally, not just this cutover.
2. **Pin the agent-ingress port** to a stable value (env/config) so the baked URL survives restarts.
3. **Exempt `POST /api/sessions/<id>/hooks` on the main listener too** (same allowlist), so a hook that hits 7330 isn't 401'd. Simplest, but widens the auth carve-out onto the public-facing port — weigh against the WS/PTY-gap rationale of #1081.

## Repro

1. Spawn a session (hook URL baked to current ingress port / 7330).
2. Restart `shepherd` (or deploy) → ingress port rotates.
3. Trigger a Stop in the still-live session → `HTTP 401`/connection-refused on the hook POST.

## Context

- Observed live on session `6b92b725-3f63-4cca-8aab-4a8664ffe827` after the #1081 deploy.
- Related: hook ingest #704/#708/#738; operator auth #1079/PR #1081.

scoop:
Closes #1083.

## Problem

The hook ingest endpoint `POST /api/sessions/<id>/hooks` is auth-exempt **only** on the loopback `serveAgentIngress` listener, which was bound to an **ephemeral port** (`port = 0`) that rotates on every restart. The ingress port is baked into a live `claude` process's `--settings` argv at spawn and **cannot be rewritten in a running process** — only a re-spawn (`resume()`) re-bakes it. So after any restart/deploy an in-flight session's hook POSTs hit a dead port (connection-refused), and the one session spawned before the auth/ingress wiring points at the now-gated main `7330` → **HTTP 401** on every Stop.

Both spawn paths bake the port off the same `agentIngressPort()` accessor and are equally affected (and equally fixed): the loopback path (`agentLoopbackIngressBaseUrl`) **and** the autonomous egress path (`agentEgressBaseUrl` over `10.0.2.2` + the nft host-gateway allow rule, both from `egressHostGateway`).

## Fix (approach #2 — pin the port)

- New `config.agentIngressPort` (env `SHEPHERD_AGENT_INGRESS_PORT`, default `SHEPHERD_PORT + 1`; `0` = ephemeral). `serveAgentIngress` binds it, so the baked URL survives restarts. No autonomous-path code change — pinning the bound port fixes both consumers via the shared accessor.
- `validateAgentIngressPort` **fail-fasts at startup** (matching `validatePreviewPortRange`) on a collision with the main port, the served port, or the preview range — never a silent fallback. Catches the concrete clash `SHEPHERD_PORT=8000` → ingress `8001` = default `previewPortBase`.
- Clean-restart rebind relies on Bun's default **`SO_REUSEADDR`** (handles lingering `TIME_WAIT` connection entries on the pinned port). Deliberately **no `reusePort` (SO_REUSEPORT)**: that is concurrent load-balanced **co-binding**, not restart recovery, and on this auth-exempt listener it would let any same-UID local process co-bind the port and skim a load-balanced share of agent hook/queue traffic (session UUIDs) — a vector plain `EADDRINUSE` otherwise blocks.

## Important: this heals ZERO currently-live sessions

This change fixes **future** spawns only. Every pre-pinning in-flight session — **including the titled `6b92b725`** (baked to the gated `7330`) — breaks once on this deploy and recovers **only on its next `resume`/restart** (a one-time manual remedy per live session). The issue is titled for re-injection (#1); we shipped the pin (#2), so the headline symptom needs a manual `resume`.

## Operator confirmations (not derivable from the repo)

- Chose **#2 (pin)** over the titled **#1 (re-inject)** — re-inject would have to re-spawn live sessions to rewrite their frozen `--settings` argv, interrupting in-flight turns.
- Accepted that `6b92b725` stays broken until a manual `resume`, in exchange for keeping the `7330` auth carve-out closed (**rejected #3**, which would have exempted `/hooks` on the public port and healed it in place).
- **No `reusePort`** — rationale above (avoid the local co-bind skim surface on the auth-exempt ingress).

## Tests

- `validateAgentIngressPort`: throws on main-port / served-port / preview-range overlap (incl. the `8000`→`8001` case) and on out-of-range values; exempts `0`.
- `serveAgentIngress`: binds the requested pinned port; **non-vacuous post-exit rebind** — routes a real allowlisted request through the live listener with the server as active closer (`Connection: close`), then `stop()` + rebinds the **same** port successfully (recovers a genuine server-closed connection, not a no-op).
- Full root suite green (`bun test ./test`), lint + tsc clean, Fallow pre-push gate green.

Server-only plumbing — no user-facing UX (no i18n / feature-catalog / glossary entry needed).


scoop (commented):
## Summary

This PR implements **option 2** from issue #1083: it pins the auth-exempt agent-ingress listener to a stable port (`SHEPHERD_AGENT_INGRESS_PORT`, default `mainPort + 1` = 7331) instead of an ephemeral port, plus a startup `validateAgentIngressPort` collision check. The commit message (`fix(ingress): pin agent-ingress port so live sessions survive restarts`) confirms the team chose the pin approach over the title's "re-inject on reconnect" (option 1) — both are explicitly sanctioned by the issue's "Proposed fix (one of)" list. The change is coherent end-to-end and I found no blocking defects.

### Verification

- **Pin actually reaches the baked URL.** `serveAgentIngress(appDeps, config.agentIngressPort)` (src/index.ts:1834) → `agentIngressState.port = agentIngress.port` (src/index.ts:1835) → wired into the service via `agentIngressPort: () => agentIngressState.port` (src/index.ts:327), which `resolveSpawnBaseUrl` reads when baking the hook/build-queue URL (src/service.ts:1152-1153, 1119-1125). So a stable bound port → stable baked URL across restarts. The fix is internally consistent. (verified against src/service.ts:1141-1159)
- **Validation runs before the bind, fail-fast.** `validateAgentIngressPort` is invoked at src/index.ts:263-269 inside the unconditional init block (lines 244-275), well before the listener binds at src/index.ts:1834. Throws on invalid value / main-port / served-port / preview-range collision. (verified against src/config.ts:286-326)
- **Collision checks mirror the sibling `validatePreviewPortRange`.** Same exclusive `rangeEnd = base + count`, same served-port symmetry. Range boundary is correct: `>= base && < base+count`; tests cover last-in-range (8016 throws) and exclusive end (8017 passes). (verified against src/config.ts:248-262 vs 319-324; test/preview-config.test.ts:384-410)
- **Default + README parity.** `agentIngressPort: Number(process.env.SHEPHERD_AGENT_INGRESS_PORT ?? mainPort + 1)` (src/config.ts:366) matches the README row "`SHEPHERD_PORT + 1` (e.g. 7331)". `mainPort` is extracted so `config.port` and the ingress default shift in lockstep. (verified against src/config.ts:350-366, README.md:181)
- **`0` ephemeral opt-out** short-circuits all checks (src/config.ts:296) and `serveAgentIngress(deps, port = 0)` keeps the old behavior + backward-compat default for the test callers. (verified against src/server.ts:5123)
- **Security reasoning is sound.** The docstring's choice to rely on SO_REUSEADDR and deliberately *not* set `reusePort` (SO_REUSEPORT) is correct for an auth-exempt loopback listener; a predictable port pre-bound by a hostile same-UID process causes shepherd's bind to throw (fail-fast), not silent interception. (src/server.ts:5115-5122)

### Notes (non-blocking, informational)

- This implements option 2, not option 1. As the issue itself states, pinning prevents *future* channel loss on restart/deploy but does **not** retroactively heal the specific cutover session (`6b92b725…`) whose URL was already baked to the gated main port 7330 before ingress existed — that session still needs a respawn. This is an acknowledged limitation of option 2, not a defect in the PR.
- With `SHEPHERD_PORT=8000` and default preview config, the default ingress port becomes 8001 = `previewPortBase`, so startup now hard-fails with a fixable error. This is intentional and correct (8001 ingress would otherwise steal a preview slot); the error message tells the operator how to resolve it. Worth being aware of as a behavioral change for that specific port layout.
- Consistent with the existing config idiom (`SHEPHERD_PORT`, etc.), an empty-string env value (`SHEPHERD_AGENT_INGRESS_PORT=`) yields `Number("") = 0` → silent ephemeral mode, since `??` only catches null/undefined. Pervasive pre-existing pattern across config.ts; not introduced or worsened here.

No findings to address.

Issue #1087, opened 2026-06-25
Generalize Plan Gate across agent CLIs and make reviewer provider configurable


kai-osthoff:
## Problem

Shepherd currently treats Codex as an Alpha/MVP interactive terminal provider and disables Plan Gate, Autopilot, and Claude model selection for Codex sessions. The Plan Gate implementation itself is mostly provider-neutral because it is built around a file contract (`.shepherd-plan.md`, optional `.shepherd-plan-blocks.json`) plus PTY steering, but the current wiring is Claude-specific in several places:

- `SessionService.create` enables Plan Gate only when `agentProvider === "claude"`.
- Codex sessions get a bare `codex --no-alt-screen --dangerously-bypass-approvals-and-sandbox <prompt>` spawn, without Shepherd's plan-gate directive/system prompt contract.
- Codex sessions persist `planGateEnabled: false`, `autopilotEnabled: false`, and `planPhase: null` even if the task requested Plan Gate.
- The UI disables the Plan Gate control whenever Codex is selected.
- `PlanGateService` always spawns the adversarial reviewer through the Claude-specific `readonlyReviewerArgv()` path and reads Claude-session usage from `~/.claude/projects`.

We want Plan Gate to work with Codex CLI and be ready for future agent CLIs, without adding another pile of provider-name conditionals.

## Goal

Implement a generalized provider architecture for Plan Gate so:

1. A main coding session can run on Claude Code, Codex CLI, or future CLIs.
2. The Plan Gate reviewer can be configured independently of the main session provider.
3. A Claude main session can use a Codex Plan Gate reviewer.
4. A Codex main session can use a Claude or Codex Plan Gate reviewer.
5. Future CLIs can plug in through capabilities/adapters instead of new hardcoded branches.

### 1. Introduce provider adapters

Add a provider registry instead of scattered `agentProvider === ...` checks.

Sketch:

```ts
interface AgentProviderAdapter {
  id: AgentProvider;
  supports: {
    mainSession: boolean;
    planGateMainSession: boolean;
    planReview: boolean;
    autopilot: boolean;
    modelSelection: boolean;
    usageAccounting: boolean;
    resume: boolean;
    ptySteering: boolean;
  };
  buildMainArgv(ctx: MainSpawnContext): SpawnSpec;
  buildResumeArgv?(ctx: ResumeSpawnContext): SpawnSpec;
[...]

PR #1097, opened 2026-06-25
fix(ui): honor default-CLI setting in New Task; restore gate/autopilot on switch-back

scoop:
## Problem

With **Claude Code** set as the default coding CLI in Settings, the New Task dialog still defaulted the **CODING CLI** dropdown to **Codex**, against the explicit setting. Two coupled defects:

1. **CLI ignores the setting.** The dropdown was seeded — and an effect re-forced — to Codex whenever a usage hold was *likely* (`holdLikely`). Because `usageHoldEnabled` **defaults on** (`src/config.ts:578`) at `usageHoldPct=80`, any heavy Claude user sitting ≥80% usage got Codex *every* time, silently overriding `defaultAgentProvider=claude`. The callout even admitted it: *"Shepherd selected Codex for this task."*
2. **Plan gate / Autopilot can't be re-enabled.** Selecting Codex force-unchecked Plan gate + Autopilot **and pinned** their `*Touched` flags. Switching the dropdown back to Claude left them off and pinned — they never re-seeded from the repo defaults. That's the "two flips and they're still off" symptom (repo default `Autopilot: on`, yet unchecked after switching back).

## Fix

- Seed `agentProvider` from the explicit `defaultAgentProvider` setting; **drop the `holdLikely → codex` auto-flip and the forced re-switch effect**. When held, Claude stays selected and the existing **Hold for reset / Submit anyway** buttons still offer the handoff — Codex is one manual dropdown-pick away.
- Make the Codex plan-gate/autopilot disabling **non-pinning**, and gate the repo-default re-seed to yield to it, so **codex → claude restores the repo defaults in a single flip**.
- Force both flags off at submit for Codex via a small `automationFlag` helper (also keeps `doSpawn` under the fallow cognitive-complexity threshold).
- Reword the now-inaccurate "Shepherd selected Codex" hold note (EN + DE).
- Remove the orphaned `agentProviderTouched` state + `onAgentProviderTouched` plumbing (no consumer once the auto-switch is gone).

## Deliberate deviation (flagged per house rules)

This intentionally changes the behavior introduced by **c9546519** (auto-preselect Codex on usage hold). Rationale: an **explicit operator setting must win** over the hold prediction. The handoff itself is preserved — it just no longer hijacks the selected CLI; the operator opts in via the dropdown or the Submit-anyway / Hold-for-reset buttons.

Out of scope (separate, larger efforts): generalizing Plan Gate to *run on* Codex (#1087) and hardening the Codex plan gate (#1088).

## Verification

- `cd ui && bun run check` — 0 errors; `check:i18n` — locales in parity.
- New/updated tests in `NewTask.browser.test.ts` (57 pass): usage hold keeps the Claude default + offers the manual handoff; **codex → claude restores plan gate / autopilot in one flip**.
- `bunx fallow@2.100.0 audit --base origin/main` — no issues in changed files; pre-push gate green.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -1947,7 +1947,7 @@
   "newtask_agent_provider_label": "Coding CLI",
   "newtask_agent_provider_codex_alpha_badge": "Alpha MVP",
   "newtask_agent_provider_codex_alpha_note": "Die Codex-Integration ist eine Alpha/MVP-Funktion und wird gerade intensiv weiterentwickelt.",
-  "newtask_agent_provider_codex_suggested_for_hold": "Claude Code ist am Nutzungsrückhalt-Schwellenwert, deshalb hat Shepherd für diese Aufgabe Codex ausgewählt. Wechsle zurück zu Claude Code, wenn du bis zum Reset zurückhalten oder trotzdem einreichen möchtest.",
+  "newtask_agent_provider_codex_suggested_for_hold": "Claude Code ist am Nutzungsrückhalt-Schwellenwert, daher umgeht das Ausführen dieser Aufgabe auf Codex den Rückhalt. Wechsle die Coding-CLI zurück zu Claude Code, um sie stattdessen bis zum Reset zurückzuhalten oder trotzdem einzureichen.",
   "newtask_agent_provider_codex_note": "Codex startet als interaktive Terminal-Session. Plan Gate und Autopilot sind für diese Aufgabe aus; die Modellauswahl nutzt Codex-CLI-Modelle.",
   "agent_provider_claude": "Claude Code",
   "agent_provider_codex": "Codex",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -1947,7 +1947,7 @@
   "newtask_agent_provider_label": "Coding CLI",
   "newtask_agent_provider_codex_alpha_badge": "Alpha MVP",
   "newtask_agent_provider_codex_alpha_note": "Codex integration is an Alpha/MVP feature and is under heavy development.",
-  "newtask_agent_provider_codex_suggested_for_hold": "Claude Code is at the usage-hold threshold, so Shepherd selected Codex for this task. Switch back to Claude Code if you want to hold it until the reset or submit anyway.",
+  "newtask_agent_provider_codex_suggested_for_hold": "Claude Code is at the usage-hold threshold, so running this task on Codex sidesteps the hold. Switch the coding CLI back to Claude Code to hold it until the reset or submit anyway instead.",
   "newtask_agent_provider_codex_note": "Codex starts as an interactive terminal session. Plan Gate and Autopilot are off for this task; the model picker uses Codex CLI models.",
   "agent_provider_claude": "Claude Code",
   "agent_provider_codex": "Codex",
--- a/ui/src/lib/components/NewTask.browser.test.ts
+++ b/ui/src/lib/components/NewTask.browser.test.ts
@@ -996,22 +996,67 @@ async function fillPromptAndClickRun() {
 }
 
 describe("NewTask first-task confirm step", () => {
-  it("likely Claude hold selects Codex and explains why", async () => {
-    const repoPath = "/repo/hold-prefers-codex";
+  it("usage hold keeps the operator's Claude default and offers a manual handoff", async () => {
+    const repoPath = "/repo/hold-keeps-claude";
     mockGetRepoConfig.mockResolvedValue(confirmedRepoConfig());
     const onsubmit = vi.fn().mockResolvedValue(undefined);
+    // Default CLI is Claude; a usage hold must NOT silently switch it to Codex.
     render(NewTask, { props: { onsubmit, initialRepoPath: repoPath, holdLikely: true } });
 
     const provider = () => document.querySelector<HTMLSelectElement>("#nt-agent-provider")!;
+    await expect.poll(() => provider().value).toBe("claude");
+    // Held → the dual Hold-for-reset / Submit-anyway buttons offer the handoff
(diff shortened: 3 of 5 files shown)


scoop (commented):
## Summary

The PR satisfies both halves of the task and the reactive logic is sound.

### Task part 1 — "defaults to Codex despite Claude being my CLI"
The usage-hold auto-switch effect (`if (!holdLikely || agentProviderTouched || agentProvider !== "claude") return; agentProvider = "codex"`) is removed, and `agentProvider` now seeds once from the operator's explicit `defaultAgentProvider` (`ui/src/lib/components/NewTask.svelte:147`). When held, Claude stays selected and the existing dual Hold-for-reset / Submit-anyway buttons (`ui/src/lib/components/NewTask.svelte:935`) provide the handoff. Verified the new test covers this (`ui/src/lib/components/NewTask.browser.test.ts:998`).

### Task part 2 — "two flips to get back to Claude Code to re-enable plan gate/autopilot"
The Codex force-off effect no longer pins `planGateTouched`/`autopilotTouched` (`ui/src/lib/components/NewTask.svelte:356-359`), and the two seed effects early-return while Codex is selected (`ui/src/lib/components/NewTask.svelte:334-340`, `:347-350`). Because `agentProvider` is a tracked dependency of both seed effects, flipping Codex→Claude re-runs them and—since the touched flags were never set—re-seeds `planGate`/`autopilot` from the repo default in a single flip. Confirmed no effect-loop: in Claude mode the force-off effect early-returns (only `agentProvider` tracked); in Codex mode the seed effects early-return. `automationFlag()` still forces both off at submit for Codex regardless of display state (`ui/src/lib/components/NewTask.svelte:582-588`, used at `:607-608`).

### Verification performed
- `untrack` import correctly removed; no remaining uses (grep clean) — `ui/src/lib/components/NewTask.svelte:2`.
- `onAgentProviderTouched` removed from both `NewTask.svelte:913` callsite and `NewTaskRunSettings.svelte` (prop, type, and `agentProviderChanged` body at `:60`); no dangling references anywhere (grep clean).
- i18n parity for the reworded string holds: present in both catalogs — verified `ui/messages/en.json:1950` and `ui/messages/de.json:1950`.
- Codex hold callout still renders only when `agentProvider === "codex" && holdLikely` — verified `ui/src/lib/components/new-task/NewTaskRunSettings.svelte:202,209`.

### Minor (non-blocking)
When the operator *manually* enables plan-gate/autopilot against a repo whose default is off (sets the touched flag), then round-trips through Codex (force-off clears the value) and back to Claude, the seed effect skips re-seeding because the touched flag is still set, so the manual ON is silently dropped to OFF. This is reachable today but is **not a regression** — the prior code force-set the touched flag too and produced the identical end state — and the re-seed-from-default behavior the PR targets works correctly. Noting only because the new comment advertises "restores the repo default in one flip," which holds for the default case but not for a manual override. Defensible as-is; flagging for awareness.

scoop (commented):
## Review

This revision reworks how Codex selection interacts with the plan-gate / autopilot toggles and removes the usage-hold auto-switch to Codex. It satisfies the task and the previously raised nit is genuinely fixed.

### Task satisfaction
- `agentProvider` now seeds directly from the operator's `defaultAgentProvider` (NewTask.svelte:147); the prior `holdLikely && default === "claude" ? "codex"` override effect is removed (NewTask.svelte:348-352). A held Claude default now stays on Claude and surfaces the dual Hold-for-reset / Submit-anyway buttons (NewTask.svelte:927-945). Verified the hold note now only renders when Codex is manually chosen (NewTaskRunSettings.svelte:206-215).
- Switching back to Claude restores plan-gate/autopilot in a single flip because the toggles are no longer mutated on Codex selection — the seed effects (NewTask.svelte:334-346) still mirror the repo default when untouched.

### Prior finding — RESOLVED
> NewTask.svelte: a manually-toggled plan-gate/autopilot value that differs from the repo default is silently lost after a Codex round-trip.

The new approach decouples display from state: the checkbox `checked` is a one-way display expression `agentProvider === "codex" ? false : planGate` (NewTaskRunSettings.svelte:84 and :112), so selecting Codex no longer mutates `planGate`/`autopilot` or sets the `*Touched` flags. Submit forces both off only for Codex via `automationFlag` (NewTask.svelte:577-579, applied at :602-603), leaving the Claude-context choice intact. A manual override (e.g. repo default OFF, user turns ON → `planGateTouched=true`, `planGate=true`) therefore survives a Codex round-trip — the underlying state is never cleared and the still-set touched flag correctly keeps the seed effect from re-seeding. The added test `preserves a manual plan-gate / autopilot override across a Codex round-trip` (NewTask.browser.test.ts:1062-1096) exercises exactly this path. Verified — not re-raising.

### Verification notes
- i18n parity: the changed key `newtask_agent_provider_codex_suggested_for_hold` is present and updated in both catalogs — verified en.json:1950 and de.json:1950.
- The removed `onAgentProviderTouched` / `agentProviderTouched` symbols have no remaining references (grep returned nothing across the tree), and the prop is dropped consistently from both the destructure and the `$props` type in NewTaskRunSettings.svelte:23,42 and from the call site NewTask.svelte:908.
- Controlled-checkbox pattern is sound: while Claude is selected the box is interactive and `onchange` writes `planGate = e.currentTarget.checked` then re-derives `checked` to the same value (NewTaskRunSettings.svelte:84-89); while Codex is selected the box is `disabled` (:90, :117) so no `onchange` can fire to corrupt the preserved state. The `if (planGate) research = false` guard reads `planGate` after it is assigned — unchanged semantics from the prior `bind:checked`.
- `// svelte-ignore state_referenced_locally` on NewTask.svelte:146 correctly suppresses the warning for seeding `$state` from the `defaultAgentProvider` prop.

No findings.

PR #1104, opened 2026-06-25
fix(codex): resume sessions with codex CLI

kai-osthoff:
## Summary
- route session resume through the active provider instead of always spawning `claude --resume`
- use `codex resume --last` for Codex sessions and stop storing a fake Claude session id for new Codex runs
- allow the Resume affordance for parked Codex sessions without a Claude session id

## Verification
- bun test test/service.test.ts
- cd ui && bun test src/lib/format.test.ts
- bun run lint
- cd ui && bun run check
- cd ui && bun run check:i18n
- git diff --check

## Notes
- Full pre-push was bypassed after unrelated local failures: extension deps were initially missing, then UI vitest failed in `GitRail.browser.test.ts` review-popover browser tests (`body link rendered`). Root pre-push tests passed.


kai-osthoff (commented):
## Summary

This PR (commits b397b333 + a837221e) adds Codex-specific session resume and then refactors `resume()` into small helpers. The task — "does resuming a session need additional/different handling for Codex" — is genuinely addressed: Codex gets its own resume argv, the `claudeSessionId` requirement is relaxed for the Codex path, and the UI `canResume`/banner gating no longer keys off a (now-empty) `claudeSessionId`. The refactor is internally consistent and I found no blocking bug.

### What I verified

- **Provider branch in resume** — `buildResumeArgv` routes `codex` → `buildCodexResumeArgv`, else `buildClaudeResumeArgv` (`src/service.ts:1532`). `resumeTarget` only requires a pinned id for `claude` (`src/service.ts:1543`), so a Codex session with `claudeSessionId === ""` is resumable. Matches the new test (`test/service.test.ts:1681-1719`).
- **Empty id for Codex on create** — `claudeSessionId: agentProvider === "claude" ? claudeSessionId : ""` (`src/service.ts:1730`), asserted by `test/service.test.ts:148-153`. Consistent with `resumeTarget` not requiring it for Codex.
- **UI gate parity** — `canResume` now passes Codex through regardless of `claudeSessionId` (`ui/src/lib/format.ts:603-609`), covered by `ui/src/lib/format.test.ts:24-31`. `ViewportTermBanners` switched from `session.claudeSessionId` to the `resumable` prop (`ViewportTermBanners.svelte:58`), and `Viewport.svelte` passes the already-derived `resumable` (`Viewport.svelte:340,2078`); the removed `Session` import leaves no dangling `session.` reference (grep clean). Without this change the in-terminal overlay Resume button would never appear for Codex (empty id) — so the rename is a real fix, not just cosmetics.
- **i18n parity** — the three reworded keys (`redrawmenu_resume_hint`, `feat_resume_session_body`, `feat_redraw_menu_body`) exist in BOTH catalogs; en/de are the only catalogs (`ui/messages/`), each has all three (grep count 3/3). No key added/removed, so no parity break.
- **Refactor shape** — `finishResumeSpawn` takes `Extract<SpawnOutcome, {ok:true}>` and is only reached after the `if (!outcome.ok) return null` narrowing (`src/service.ts:2080-2084`); `SpawnOutcome` union confirmed at `src/service.ts:901-913`. Sandbox/profile/auto-gate handling is preserved verbatim for both providers via `prepareResumeSpawn`/`resumeRefusedByAutoGate`.

### Note relevant to the task — `codex resume --last` precision

Claude resume pins an exact conversation (`--resume <claudeSessionId>`), but Codex resume relies on `codex resume --last` (`src/service.ts:1502-1512`) — there is no stored Codex session id, so the target session is implied by cwd + recency. The codebase states non-isolated sessions all share the main repo dir (`src/service.ts:814`), and nothing forces Codex sessions to be isolated. So with two concurrent/non-isolated Codex sessions in the same repo, `--last` may not deterministically resume the *intended* one the way the Claude path does. Whether this actually mis-targets depends on how the Codex CLI scopes `--last` (global vs per-cwd vs per-rollout), which I could not verify — see limitations. Worth confirming or documenting; flagged as a non-blocking finding.

### Stated limitations (could not verify — not in this repo)

- The Codex CLI is not vendored here, so I could not confirm that `codex resume --last --no-alt-screen --dangerously-bypass-approvals-and-sandbox [--model X]` is a valid flag set/order for the `resume` subcommand, nor that `--model` is accepted on resume. Assumed as written; the test (`test/service.test.ts:1709-1718`) only asserts the literal argv, not CLI acceptance.
- I could not verify the scoping semantics of `codex resume --last` (global vs cwd-scoped), which is what determines whether the precision concern above is benign or a real mis-target.

### Nit

`Viewport.svelte:339` — the global claude→Claude/agent rewording mangled the comment to "... hides **Claude**." It should read "... hides **it**." (it = the Resume button). Cosmetic only.

No out-of-scope pre-existing issues worth flagging.

kai-osthoff (commented):
## Re-review

The PR adds a Codex-specific resume path so that finished Codex sessions (which have no pinned `claude --session-id`) can be brought back, instead of the prior code building a `claude --resume <bogus-uuid>` argv for them. The `resume()` method was refactored into small helpers (`resumeTarget`, `liveAgentFor`, `adoptLiveResumeAgent`, `resumeRefusedByAutoGate`, `buildResumeArgv`, `prepareResumeSpawn`, `finishResumeSpawn`) with a provider switch.

### Prior findings

1. **`src/service.ts` `buildCodexResumeArgv` / `--last` scoping** (non-blocking, "confirm or document") — ADDRESSED. The new helper carries an explicit comment (src/service.ts:1503-1507) documenting that Shepherd does not persist a Codex session id, that `--last` is scoped by the Codex CLI's own resume rules, and that concurrent/non-isolated Codex sessions sharing a cwd can target the most recent Codex session for that cwd rather than this exact row — to remain visible until Codex spawn/ingest records a provider-native id. This satisfies the prior "document the limitation" request.
2. **`ui/src/lib/components/Viewport.svelte:339` typo "hides Claude."** — ADDRESSED. Line 339 now reads "... hides it." (verified against ui/src/lib/components/Viewport.svelte:339).

### Verification performed

- **Claude argv unchanged by refactor**: `buildClaudeResumeArgv` (src/service.ts:1518-1532) reproduces the original argv byte-for-byte — same flags, same `--settings` overlay with `{ sessionId, baseUrl, token }`, same `pushModelFlag` — and `prepareResumeSpawn`/`finishResumeSpawn` reproduce the original `prepareSpawn` opts and `store.update`/`setSandboxState` calls (compared against the removed inline block in the diff at src/service.ts old lines 248-296).
- **Codex argv consistency**: resume flags `codex resume --last --no-alt-screen --dangerously-bypass-approvals-and-sandbox [--model M]` (src/service.ts:1508-1512) match the spawn flags in `buildCodexSpawnArgv` (src/service.ts:1496) and the model is passed conditionally as on the spawn path; the test at test/service.test.ts:358-367 pins this exact argv.
- **Resumability gating**: `resumeTarget` (src/service.ts:1543-1551) only requires a `claudeSessionId` for `provider === "claude"`, so Codex rows with an empty id are now resumable; `canResume` mirrors this with `(provider === "codex" || !!s.claudeSessionId)` (ui/src/lib/format.ts:21) and is covered by ui/src/lib/format.test.ts:24-31.
- **`claudeSessionId: ""` for Codex** (src/service.ts:1735): verified every consumer guards on a falsy id before use — src/poller.ts:222, src/server.ts:377/1674, src/process-reaper.ts:267, src/usage-breakdown.ts:404, src/usage-snapshot.ts:49, src/usage.ts:532, src/index.ts:1346 — so storing "" is safe (Codex writes no Claude jsonl transcript anyway). Test updated at test/service.test.ts:151.
- **ViewportTermBanners prop swap**: the gate changed from `session.claudeSessionId` to a `resumable` boolean (ui/src/lib/components/viewport/ViewportTermBanners.svelte:55, prop declared :25), the `Session` import was dropped, and `session` is no longer referenced in that component (grep returned nothing); Viewport passes the derived `resumable` (ui/src/lib/components/Viewport.svelte:340, 2073). Consistent.
- **i18n parity**: the three reworded keys (`redrawmenu_resume_hint`, `feat_resume_session_body`, `feat_redraw_menu_body`) are value-only edits present in both ui/messages/en.json and ui/messages/de.json (verified at en.json:632/930/1060 and de.json:632/930/1060); no keys added or removed.

The implementation satisfies the task (Codex sessions need and now receive a different, Codex-native resume command) and I found no logic, security, or quality defects in the diff.

- `src/types.ts:22`: the `claudeSessionId` doc comment (`"" for pre-feature sessions`) is now also true for Codex sessions but was not updated; file not in this diff.

PR #1105, opened 2026-06-25
fix(ui): surface held-task spawn/discard failures inline

kai-osthoff:
## Problem

On a phone (and desktop), tapping **Jetzt starten** / **Verwerfen** on a held task could appear to do **nothing** — "passiert einfach gar nix."

The tap *was* reaching the handler. The real cause was silent error-swallowing: `POST /api/held/:id/spawn` can legitimately fail (sandbox auto-refusal `403`, agent-name collision `409`, herdr/git failure `502`, bad request `400`) and the task stays in the held store — but `doSpawnHeld` / `doDiscardHeld` caught the error in an empty `catch {}` and showed no feedback. The row stayed put, the user re-tapped, still nothing.

## Fix

Surface a per-task **inline error** inside the held popover (`role="alert"`), e.g. _"Aufgabe konnte nicht gestartet werden. Sie bleibt zurückgehalten – bitte erneut versuchen."_ Retrying is just tapping the button again — the error clears on the next attempt.

### Why inline and not a toast

The mobile held dialog (#1103) is a **fullscreen portal at `z-index: 60`** appended to `<body>`, and the toast stack is *also* `z-index: 60` but earlier in DOM order — so a bottom-anchored toast would render **behind** the dialog and the user would still see nothing. Inline feedback lives inside the surface itself, so it's visible in both the desktop anchored popover and the mobile fullscreen dialog.

## Changes

- `TopBar.svelte` — track a per-id `heldErrors` map; set it when spawn/discard rejects, clear it at the start of each attempt.
- `TopBarHeldBadge.svelte` — render the inline `role="alert"` error per row (spans both columns in the fullscreen variant); tokens only (`--color-red`, `--fs-micro`/`--fs-meta`).
- `en.json` / `de.json` — `topbar_held_spawn_failed`, `topbar_held_discard_failed` (parity passes).
- Test — asserts the inline alert renders with the failure copy on iPhone width.

## Verification

- `bun run check:i18n` ✓ (parity, 1977 keys)
- `bun run check` (svelte-check) ✓ 0 errors
- `bun run test src/lib/components/TopBar.browser.test.ts` ✓ 62/62 (incl. new test)

Note: the full UI suite shows 4–5 flaky failures in `GitRail.browser.test.ts` (a `marked`+DOMPurify dynamic-import timeout under parallel CPU load) — unrelated to this diff; GitRail passes in isolation on both the pristine and modified trees.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -1614,6 +1614,8 @@
   "topbar_held_spawn_cli_label": "Starten mit",
   "topbar_held_original_cli": "Zurückgestellt für {cli}",
   "topbar_held_discard": "Verwerfen",
+  "topbar_held_spawn_failed": "Aufgabe konnte nicht gestartet werden. Sie bleibt zurückgehalten – bitte erneut versuchen.",
+  "topbar_held_discard_failed": "Aufgabe konnte nicht verworfen werden. Bitte erneut versuchen.",
   "topbar_held_empty": "Keine zurückgehaltenen Aufgaben",
   "topbar_held_why": "Zurückgehalten, weil die Nutzung hoch ist – sie starten automatisch beim nächsten Reset.",
   "topbar_held_why_at": "Nächster Reset {time}.",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -1614,6 +1614,8 @@
   "topbar_held_spawn_cli_label": "Run on",
   "topbar_held_original_cli": "Held for {cli}",
   "topbar_held_discard": "Discard",
+  "topbar_held_spawn_failed": "Couldn't start the held task. It's still held — try again.",
+  "topbar_held_discard_failed": "Couldn't discard the held task. Try again.",
   "topbar_held_empty": "No held tasks",
   "topbar_held_why": "Held because usage is high — they start automatically when usage resets.",
   "topbar_held_why_at": "Next reset {time}.",
--- a/ui/src/lib/components/TopBar.browser.test.ts
+++ b/ui/src/lib/components/TopBar.browser.test.ts
@@ -770,6 +770,49 @@ describe("TopBarHeldBadge — mobile held-task dialog", () => {
     await close.click();
     expect(closeHeldPop).toHaveBeenCalledWith(true);
   });
+
+  it("shows an inline error when a held-task spawn was refused server-side", async () => {
+    await page.viewport(390, 844);
+    document.body.style.width = "390px";
+    const heldItems: HeldTask[] = [
+      {
+        id: "held-1",
+        repoPath: "/work/shepherd",
+        createdAt: 1_700_000_000_000,
+        input: {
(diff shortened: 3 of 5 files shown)


kai-osthoff (commented):
## Summary

The PR turns the previously-swallowed spawn/discard catch blocks into inline, per-row error feedback. This directly addresses the reported symptom ("tap *Start now* on mobile and nothing happens"): when `spawnHeld` is refused server-side it throws (`ui/src/lib/api.ts:211`), the old `catch { // ignore }` made the button read as dead, and now a `role="alert"` line renders in the row.

## Verification

- **Locale parity** — both new keys exist in every catalog the repo maintains: `topbar_held_spawn_failed` / `topbar_held_discard_failed` at `ui/messages/en.json:1617-1618` and `ui/messages/de.json:1617-1618`. Only `de.json` and `en.json` exist under `ui/messages/`, so parity is complete.
- **Message accessors** — `m.topbar_held_spawn_failed()` / `m.topbar_held_discard_failed()` used at `ui/src/lib/components/top-bar/TopBarHeldBadge.svelte:107-109` and in the test at `TopBar.browser.test.ts:814`; both resolve to the keys above (paraglide generates accessors from the catalog).
- **Prop wiring** — `heldErrors` is declared `$state` at `TopBar.svelte:338`, passed down at `TopBar.svelte:623`, and received as `heldErrors = {}` with type `Record<string, "spawn" | "discard">` at `TopBarHeldBadge.svelte:38,55`. The `"spawn" | "discard"` union is consistent across producer (`TopBar.svelte:346,356`) and consumer (`TopBarHeldBadge.svelte:107`).
- **Reactivity** — `delete heldErrors[id]` / `heldErrors[id] = ...` mutate the same `$state` proxy passed to the child, so the row re-renders; error is cleared at the start of each retry (`TopBar.svelte:341,351`) and on success via `loadHeld()`.
- **CSS tokens** — `.held-row-error` uses `--color-red`, `--fs-micro`, `--fs-meta`, all defined in `ui/src/app.css:23,81,82`. The fullscreen override (`TopBarHeldBadge.svelte:474-480`) spans the action grid full-width, matching `.held-cli`.
- **Test** — renders the badge with `heldErrors: { "held-1": "spawn" }` and asserts the alert text; matches the rendered branch.

The error is announced via `role="alert"`, and the inline-vs-toast rationale (toast renders behind the fullscreen mobile portal) is documented at `TopBar.svelte:332-337` and consistent with the portal z-index in `TopBarHeldBadge.svelte:267-289`.

## Minor (non-blocking, no change required)

`heldErrors` entries for a task that later disappears from `heldItems` via a WS-driven refresh (rather than through `doSpawnHeld`/`doDiscardHeld`) are never pruned from the record. This is harmless — a stale entry is only ever rendered if a row with that id exists — and the key set is bounded by held-task ids, so no cleanup is needed.

The PR is a targeted, correct improvement to a genuinely silent failure path.

PR #1114, opened 2026-06-26
fix(ui): heartbeat full width in compact herd sidebar

scoop:
## Problem

On an unfolded foldable, the live-activity **heartbeat strip** under a running session's row rendered as a ~64px stub instead of spanning the activity row (see report screenshot).

## Root cause

The unfolded foldable uses the **two-column desktop layout** with the **compact sidebar** (`.grid.compact` → `grid-template-columns: minmax(244px, 288px) 1fr`, `ui/src/routes/+page.svelte`). That makes the herd `.units` container (`container: herd / inline-size`) ≤300px wide, triggering `@container herd (max-width: 300px)` in `UnitRow.svelte`, which pinned the strip to `flex: none; width: 64px` — overriding the full-width default (`flex: 1 1 auto; width: auto; max-width: none`).

Mobile phones (flow mode, full-width container >300px) never hit the query and were already full-width; only the narrow sidebar got the stub.

## Fix

Remove the 64px strip override from the 300px container query so the strip inherits the full-width default in every ≤300px sidebar. Per operator decision, this applies to **all** narrow sidebars (foldable compact + narrow desktop), with no `pointer: coarse` guard — consistent with the strip's documented "claims the whole activity line on every device" intent.

The unrelated density choices in that block are kept: dropped stepper (`.meta-stepper { display: none }`), single-line prompt clamp, tighter column-gap.

## Deliberate deviation (house rule)

This **reverts a deliberately documented density pin** (the `64px` width was an intentional choice in the 300px container query). Flagged in the code comment and here per the repo's prior-art-deviation rule. It aligns with the strip's own stated principle and the user's report.

## Verification

- `cd ui && bun run check` — 0 errors, 0 warnings.
- `UnitRow.browser.test.ts` — 19/19 pass.
- Branch-hygiene gate + fallow pre-push gate pass.
- CSS-cascade reasoning: removing the override leaves the proven full-width default (already known-good at wider sidebar widths); the container-query removal cannot make the strip narrower than that default.

[no-feature-entry] — no new user-facing feature; an existing element's width changes.

### DIFF
--- a/ui/src/lib/components/UnitRow.svelte
+++ b/ui/src/lib/components/UnitRow.svelte
@@ -824,9 +824,9 @@
     line-height: 1.3;
     color: var(--color-muted);
   }
-  /* The heartbeat claims the whole activity line on every device. Scoped under
-     .u-activity so the strip override can't leak to a future global .strip; kept
-     before the narrow @container block so that block's 64px override still wins. */
+  /* The heartbeat claims the whole activity line on every device — including the
+     narrow ≤300px sidebar (no container-query override pins it back). Scoped under
+     .u-activity so the rule can't leak to a future global .strip. */
   .u-activity :global(.strip) {
     flex: 1 1 auto;
     width: auto;
--- a/ui/src/lib/components/UnitRow.svelte
+++ b/ui/src/lib/components/UnitRow.svelte
@@ -935,17 +935,14 @@
       -webkit-line-clamp: 1;
       line-clamp: 1;
     }
-    /* keep the heartbeat (tiny), drop the stepper so a narrow sidebar row stays
-       dense and doesn't balloon */
+    /* drop the stepper so a narrow sidebar row stays dense and doesn't balloon.
+       The heartbeat strip is deliberately NOT pinned narrow here (was flex:none/
+       width:64px): on the unfolded-foldable compact sidebar that left it a ~64px
+       stub, so it now inherits the full-width default above and spans the activity
+       row on every device — matching the strip's "claims the whole activity line"
+       intent. */
     .meta-stepper {
       display: none;
-    }
-    /* the strip IS the heartbeat here — keep it, just narrower. Scoped under
-       .u-activity so the override can't leak to a future global .strip. flex:none
-       overrides the full-width grow above so the narrow strip stays at 64px. */
-    .u-activity :global(.strip) {
-      flex: none;
-      width: 64px;
     }
   }
 


scoop (commented):
## Summary

The PR removes the `flex: none; width: 64px` override on `.u-activity :global(.strip)` inside the `@container herd (max-width: 300px)` block. With that gone, the heartbeat strip inherits the full-width default at `ui/src/lib/components/UnitRow.svelte:830-834` (`flex: 1 1 auto; width: auto; max-width: none`), which overrides the base `.strip` rule (`width: 132px; max-width: 40vw; flex: none`) in `ui/src/lib/components/HeartbeatStrip.svelte:37-45`. Net effect: in the compact/unfolded-fold sidebar the strip now spans the activity line instead of being pinned to a ~64px stub. This satisfies the task and the cascade reasoning holds — the only narrow override has been deleted, so the full-width default is the winning rule on that path (verified `64px` no longer occurs in any active rule via grep; the only remaining hits are the prose comments at `UnitRow.svelte:829` and `:940`).

## Nit

- The explanatory comment at `ui/src/lib/components/UnitRow.svelte:827-829` is now stale. It states the full-width default rule is "kept before the narrow @container block so that block's 64px override still wins" — but this PR removed that 64px override, so the ordering rationale no longer applies and is misleading to a future reader. Consider updating it to reflect that the strip is now full-width everywhere.

No correctness, security, or locale concerns. Markup string changes: none.

scoop (commented):
## Re-review

The PR is a CSS-only change in `ui/src/lib/components/UnitRow.svelte` that makes the heartbeat strip span the full activity line on the unfolded-foldable compact sidebar.

### Task satisfied
The `@container herd (max-width: 300px)` block previously pinned `.u-activity :global(.strip)` to `flex: none; width: 64px`, leaving a ~64px stub on the unfolded fold. That override is removed, so the strip now inherits the full-width default (`flex: 1 1 auto; width: auto; max-width: none`) at `UnitRow.svelte:830-834`. This addresses the task.

### Previous finding #1 (stale comment) — RESOLVED
The old comment at lines 827-829 claimed it was "kept before the narrow @container block so that block's 64px override still wins." Since this PR removes that 64px override, that wording would be misleading. The new comment at `UnitRow.svelte:827-829` no longer references the override and correctly states there is no container-query override pinning the strip back. Verified accurate: grepping the file shows the only remaining `.strip` rule is at `UnitRow.svelte:830`; no surviving `64px`/`flex:none` width override on `.strip` (the matches at `UnitRow.svelte:717,870,881,909` are unrelated `.car`/chip/stepper rules). The container query selector at `UnitRow.svelte:930` is `@container herd (max-width: 300px)`, matching the comment's "≤300px sidebar".

No new issues found in the diff.

PR #1118, opened 2026-06-26
fix(ui): follow herd repo filter onto a newly started task's repo

kai-osthoff:
## Problem

Starting a task in repo **X** while the herd was filtered to repo **Y** left the repo filter pinned to **Y** — so the freshly launched task was hidden behind the stale filter. Confusing, since the user just launched it and expects to see it. (Reported from the FleetView: a new task started in SHEPHERD, but the repo filter chip stayed on a different repo.)

## Fix

Every session-create path now routes selection through a shared `selectNewSession(id, repoPath)` helper that **follows the herd's repo filter onto the new task's repo**:

- `onsubmit` (New Task dialog)
- `submitRelaunch` (relaunch-elsewhere — can re-target a different repo)
- `onquickissue` (quick-start from the backlog)
- merge-train launches (`onmergetrain`, `onlaunchtrain`)

A `null` ("all repos") filter is left whole — it already shows the new task. A filter already on the task's repo needs no change.

The follow decision is a pure, tested predicate `shouldFollowFilterToRepo()` in `queue-strip.ts`, matching the existing `shouldClearRepoFilter` / `pickRepoSwitchTarget` helpers.

### Subtlety: re-target effect guard

`createSession` doesn't populate the store synchronously — the new session arrives via WebSocket. Changing `repoFilter` would otherwise fire the existing repo-switch re-target `$effect`, which (with the new session not yet in the store) would grab a *different* existing session in that repo and clobber the selection we just set. A one-shot non-reactive `followingNewSession` guard suppresses exactly that single re-target.

## Tests

- New unit tests for `shouldFollowFilterToRepo` (null filter, same-repo no-op, different-repo follow).
- `bun run check`, `check:i18n`, and the affected browser test files all pass. No new user-facing strings (bugfix), so no i18n keys or feature-catalog entry needed.

### DIFF
--- a/ui/src/lib/components/queue-strip.test.ts
+++ b/ui/src/lib/components/queue-strip.test.ts
@@ -13,6 +13,7 @@ import {
   queueOpenable,
   repoChipRows,
   shouldClearRepoFilter,
+  shouldFollowFilterToRepo,
 } from "./queue-strip";
 import type { RepoChip } from "./queue-strip";
 import type { BlockState } from "../triage";
@@ -471,6 +472,22 @@ describe("shouldClearRepoFilter", () => {
 
   it("TRUE when no chips remain (all sessions ended)", () => {
     expect(shouldClearRepoFilter("/repos/a", [])).toBe(true);
+  });
+});
+
+// ─── shouldFollowFilterToRepo ─────────────────────────────────────────────────
+
+describe("shouldFollowFilterToRepo", () => {
+  it("false when no filter is active (null) — 'all repos' already shows the task", () => {
+    expect(shouldFollowFilterToRepo(null, "/repos/a")).toBe(false);
+  });
+
+  it("false when the filter already points at the new task's repo", () => {
+    expect(shouldFollowFilterToRepo("/repos/a", "/repos/a")).toBe(false);
+  });
+
+  it("TRUE when the filter is pinned to a different repo — would hide the new task", () => {
+    expect(shouldFollowFilterToRepo("/repos/a", "/repos/b")).toBe(true);
   });
 });
 
--- a/ui/src/lib/components/queue-strip.ts
+++ b/ui/src/lib/components/queue-strip.ts
@@ -98,6 +98,14 @@ export function shouldClearRepoFilter(repoFilter: string | null, chips: RepoChip
   return repoFilter !== null && !chips.some((c) => c.repoPath === repoFilter);
 }
 
+/** Whether the herd's repo filter should follow a just-started task onto its repo.
+ *  A new task lands in `repoPath`; if the filter is pinned to a *different* repo the
+ *  task would be hidden behind that stale filter — so follow it. A null filter ("all
+ *  repos") already shows the task, and a filter already on `repoPath` needs no change. */
+export function shouldFollowFilterToRepo(repoFilter: string | null, repoPath: string): boolean {
+  return repoFilter !== null && repoFilter !== repoPath;
+}
+
 /**
  * The session the terminal should re-target onto after the user picks a repo chip.
  * Narrowing the herd to a repo would otherwise leave the terminal on whatever session
--- a/ui/src/routes/+page.svelte
+++ b/ui/src/routes/+page.svelte
@@ -95,6 +95,7 @@
     pickRepoSwitchTarget,
     repoChipRows,
     shouldClearRepoFilter,
+    shouldFollowFilterToRepo,
   } from "$lib/components/queue-strip";
   import BacklogView from "$lib/components/BacklogView.svelte";
   import AppOverlays from "$lib/components/page/AppOverlays.svelte";
@@ -217,6 +218,14 @@
   $effect(() => {
     const rf = repoFilter;
     if (rf == null) return; // "all repos" view keeps selection whole
+    // selectNewSession already pointed selection at a just-created session in `rf`
+    // (not in the store yet — it arrives via WS), so skip the one re-target that
+    // would otherwise grab a *different* existing session in the repo. Plain `let`,
+    // not $state: reading/clearing it here must not feed back into reactivity.
+    if (followingNewSession) {
+      followingNewSession = false;
+      return;
+    }
     untrack(() => {
       const target = pickRepoSwitchTarget(
         rf,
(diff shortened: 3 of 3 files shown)


kai-osthoff:
## Summary

The PR adds `shouldFollowFilterToRepo` and a `selectNewSession` helper that, on every create/relaunch path, points selection at the just-started session and (when the herd filter is pinned to a *different* repo) switches `repoFilter` onto the new task's repo. A one-shot `followingNewSession` guard suppresses the repo-switch re-target effect for that single filter change. The wiring across the 5 call sites is consistent, and the relaunch path correctly passes `input.repoPath` (verified against ui/src/routes/+page.svelte:1358-1373).

The pure helper itself is correct: `shouldFollowFilterToRepo(rf, repoPath) === rf !== null && rf !== repoPath` matches its tests (queue-strip.ts:105-107, queue-strip.test.ts:478-491).

However there is a real interaction bug with the sibling auto-clear effect.

## Blocking finding

**The just-set filter is immediately reverted when the new task is the first live session in its repo, and the one-shot guard leaks.**

`selectNewSession` sets `repoFilter = repoPath` (ui/src/routes/+page.svelte:446-452). The author's own comment (ui/src/routes/+page.svelte:221-224) confirms the new session is "not in the store yet — it arrives via WS", and `createSession` does not mutate the store (verified against ui/src/lib/api.ts:185-195). Therefore `repoChips` — derived solely from `store.sessions` (verified against ui/src/lib/components/queue-strip.ts:57-60) — does **not** contain `repoPath` at the time the effects flush, whenever `repoPath` had no prior live (non-archived) session.

The re-target effect was guarded with `followingNewSession` (ui/src/routes/+page.svelte:225-228), but its **sibling** auto-clear effect at ui/src/routes/+page.svelte:209-211 was **not**:

```js
$effect(() => {
  if (shouldClearRepoFilter(repoFilter, repoChips)) repoFilter = null;
});
```

`shouldClearRepoFilter(repoPath, chips) === repoPath !== null && !chips.some(c => c.repoPath === repoPath)` (queue-strip.ts:97-99) returns **true** because the new repo has no chip yet. This effect is created before the re-target effect, so on the flush triggered by `repoFilter = repoPath` it runs first and writes `repoFilter = null`.

Consequences (new task = first live session in its repo, an already-executing path today):
1. The filter does **not** follow onto the new task's repo as the task requests — it is reset to `null` ("all repos"). The task is at least still visible under "all repos", but the requested "switch the filter to the new task's repo" behaviour fails for this case.
2. Because the re-target effect early-returns on `rf == null` (ui/src/routes/+page.svelte:220) **before** reaching the `followingNewSession` consumption at line 225, the flag is never reset and **leaks as `true`**. The next time the user manually picks a repo chip, the re-target effect consumes the stale flag and `return`s, **skipping the terminal re-target** — reintroducing exactly the stranded-selection bug that effect exists to prevent (ui/src/routes/+page.svelte:212-217).

In the common Case A (the new task's repo already has a live chip) everything works as designed. The defect is specific to the repo's first session, which is a normal path and matches the screenshot scenario (starting a task in a repo you weren't already filtered to).

Suggested direction: also suppress / coordinate the auto-clear effect for the followed filter (e.g. honour `followingNewSession` there too, or have the clear guard tolerate a filter that matches a just-followed repo until the session arrives), and ensure `followingNewSession` is always consumed even when the filter is subsequently nulled.

## Verified, no issue
- No user-facing strings / i18n keys added — no locale-parity concern.
- `followingNewSession` is referenced by the effect created at line 218 but declared at line 437; no runtime TDZ because the effect body runs after the script finishes executing.
- All 5 `selectNewSession` call sites pass the correct repo for the created/relaunched session (ui/src/routes/+page.svelte:647,688,721,1373,1445).

kai-osthoff (commented):
## Summary

This revision makes the herd repo filter follow a just-started task onto its repo, coordinating two existing effects with a pair of non-reactive `let` latches (`followingRepo`, `followingNewSession`) plus a new pure helper `shouldFollowFilterToRepo`. Both prior-revision findings are now correctly resolved.

### Prior finding 1 — auto-clear reverting the followed filter — ADDRESSED
The auto-clear effect now early-returns for a just-followed repo before `shouldClearRepoFilter` can null it:
- `selectNewSession` arms `followingRepo = repoPath` and sets `repoFilter = repoPath` (ui/src/routes/+page.svelte:467-474).
- The auto-clear effect returns when `repoFilter === followingRepo` (ui/src/routes/+page.svelte:229), so the chip-less gap before the WS session arrives no longer triggers a null-out.
- The latch is released once the followed repo's chip appears, since `repoChips` is reactive and re-runs the effect (ui/src/routes/+page.svelte:223-225). After release, `repoFilter !== followingRepo(null)` so normal auto-clear resumes. Verified the guard is scoped to `repoFilter === followingRepo`, so a *different* active filter still auto-clears (ui/src/routes/+page.svelte:226-230).

### Prior finding 2 — one-shot flag leak on null filter — ADDRESSED
The re-target effect now consumes the one-shot flag before any early return:
- `const skipFollow = followingNewSession; followingNewSession = false;` runs at ui/src/routes/+page.svelte:246-247, ahead of both `if (rf == null) return` (line 248) and `if (skipFollow) return` (line 249). The flag can no longer strand `true` and suppress a subsequent legitimate repo-chip re-target.

### Additional verification
- `shouldFollowFilterToRepo` returns `repoFilter !== null && repoFilter !== repoPath` (ui/src/lib/components/queue-strip.ts:105-107); its three tests (null→false, same→false, different→true) match (ui/src/lib/components/queue-strip.test.ts:480-491).
- Identifier consistency: `shouldFollowFilterToRepo` is exported once (queue-strip.ts:105) and imported in +page.svelte:98 and the test at :16. `selectNewSession`/`followingRepo`/`followingNewSession` are all defined and used consistently (verified via grep).
- All five `selectedId = …` create/relaunch sites were converted to `selectNewSession(id, repoPath)` with the correct repo source: `repoPath` for the backlog/quick-issue/merge-train paths (ui/src/routes/+page.svelte:667,708,741) and `input.repoPath` for relaunch and the New Task submit (ui/src/routes/+page.svelte:1393,1465) — the latter being the repo the new session actually lands in (relaunch-elsewhere can change repo).
- Effect ordering reasoned through: effect 1 (auto-clear, created first) runs before effect 2 (re-target) in a flush; in the follow case effect 1's guard prevents the null-out, and effect 2 consumes+skips. Because arming `followingNewSession=true` always co-occurs with a real `repoFilter` change (shouldFollow requires `repoFilter !== repoPath`), effect 2 is always re-scheduled and the flag is always consumed on its next run — no leak.
- No user-facing strings or message keys were added, so no i18n parity concern.

- ui/src/routes/+page.svelte: if a followed session's WS arrival never happens (server-side create succeeds but the session never reaches the store), `followingRepo` is never released and would suppress a future auto-clear for that same `repoFilter`/repo. This relies on the existing WS-delivery invariant already assumed throughout the page; not reachable on any path that executes today given `selectNewSession` is only called after a successful create.

No blocking issues. Decision: comment.

PR #1119, opened 2026-06-26
fix(top-bar): collapse tallies under overflow on unfolded foldables

scoop:
## Problem

On an **unfolded foldable** (touch-desktop: coarse pointer, viewport > 768px) the top bar overflows and clips the right-most control (the gear / a status dot). The reported screenshot shows full tally labels (`HERD 2 BUSY 2 IDLE 0 BLOCKED 0`) alongside an already-compacted right side (square `[3]` held badge, dot-only clock) — i.e. compaction fired but wasn't enough.

## Root cause

1. **Tallies never compact off-`mobile`.** `TopBarTallies` only renders the compact dot+digit form when the `mobile` prop is set. In desktop / touch-desktop mode it always renders full labels — the widest text on the bar — so the measured-overflow path (`measuredCompact` → `compactBadges`) shrank only the right-side cluster (clock, badges, held, needs-you) and never the tallies. The bar still overflowed once narrow enough.
2. **Held state wasn't a measurement dependency.** The overflow-measuring `$effect` tracks `mode` / `badgeCount(chrome)` / `gauges.length`, but `badgeCount`/`ChromeState` had no held input. The held badge (#1089) appears async via the `held:changed` WS event, growing the bar without re-firing the effect or the ResizeObserver — the same async class already guarded for gauges.

## Fix

- Add a `compact` prop to `TopBarTallies`; render the compact dot+digit form when `mobile || compact`. `TopBar` passes `compact={compactBadges}`, so the tallies collapse under measured overflow on **both desktop and touch-desktop** — matching the sibling NEEDS-YOU / held badges, which already compact on the same signal. The compact tally's 44px touch floor is scoped to coarse pointers; fine-pointer desktop keeps the WCAG 2.5.8 (AA) 24px target so the collapsed digits fit the bar height.
- Fold `held` into `ChromeState` + `badgeCount` and set `held: heldCount` in `TopBar`'s `chrome`, so the existing `void badgeCount(chrome)` read re-fires the measure effect on async held arrival.

No new user-facing strings (reuses the existing compact-tally aria messages) → no i18n/feature-catalog/glossary changes. `fix`, not a feature.

## Tests

- **Tally-compaction width sweeps** (`TopBar.browser.test.ts`): one for **touch-desktop** (`[800…1600]`, gauges collapse to a gauge-btn) and one for **fine-pointer desktop** (`[960…1650]`, both gauges inline → overflows earlier). Each asserts no overflow + controls hittable at every width, *clock-dropped iff tallies-compact*, and that the sweep exercises **both** compact (narrow) and full (wide) states — font-robust under the CI monospace fallback, guarding against over-compaction when wide.
- **Async held self-heal**: `heldCount` 0→N after mount (via a generalized `TopBarLimitsHarness.setHeld`, `ResizeObserver` stubbed) must re-measure and self-heal the overflow; mutation-checked (remove `held` from `badgeCount`/`chrome` → fails), mirroring the existing gauge-arrival test.
- **Unit**: `badgeCount` matrix extended to a 6-bit mask including `held`.

## Verification

- `cd ui && bun run test` → 146 files, 2083 tests pass
- eslint + prettier clean; i18n / glossary / branch-hygiene / fallow delta gates green

### DIFF
--- a/ui/src/lib/components/TopBar.browser.test.ts
+++ b/ui/src/lib/components/TopBar.browser.test.ts
@@ -408,6 +408,44 @@ describe("TopBar — touch-desktop (unfolded fold) overflow is measurement-drive
       "sweep exercises the full state (some wide width keeps clock + full labels)",
     ).toBe(true);
   });
+
+  it("tallies COLLAPSE under measured overflow and stay FULL when there's room (held + needsYou + gauges)", async () => {
+    // The reported bug: on an unfolded fold the right-side cluster compacts but the tallies
+    // (HERD/BUSY/IDLE/BLOCKED) stay full-label off-`mobile`, so the bar still overflows and
+    // clips the gear. The fix extends compaction to the tallies on touch. Asserted as the
+    // same coupling invariant the test above uses — but for the TALLIES — over a width sweep,
+    // so it's robust to the CI monospace fallback's font metrics (the exact transition width
+    // varies). Content mirrors the screenshot: held badge + needsYou + several sessions
+    // (non-zero tallies) + both usage gauges. `needsYou` is present so `drainFrames` (which
+    // keys off `.needsyou`) actually waits for the measurement to settle.
+    const states: { width: number; clockDropped: boolean; talliesCompact: boolean }[] = [];
+    for (const width of [800, 920, 1040, 1200, 1400, 1600]) {
+      const hud = await renderTD(width, { ...sessionsProp(4), heldCount: 3, needsYou: 2 });
+      await waitNoOverflow(hud);
+      await drainFrames(hud);
+      const clockDropped = hud.querySelector(".clock")!.classList.contains("no-time");
+      const talliesCompact = !!hud.querySelector(".tallies.compact");
+      // Coupling: tallies compact iff the measured-overflow signal fired (clock dropped).
+      expect(talliesCompact, `tallies coupled to overflow signal at ${width}px`).toBe(clockDropped);
+      // Whichever form rendered, it must be the right one — never both, never neither.
+      if (talliesCompact) {
+        expect(hud.querySelector(".tally"), `no full tally at ${width}px`).toBeNull();
+      } else {
+        expect(hud.querySelector(".tally"), `full tally present at ${width}px`).not.toBeNull();
+        expect(hud.querySelector(".micro"), `full tally label at ${width}px`).not.toBeNull();
+      }
+      assertControlsHittable(hud);
+      states.push({ width, clockDropped, talliesCompact });
+    }
+    expect(
+      states.some((s) => s.talliesCompact),
+      "sweep exercises the compacted state (a narrow fold collapses the tallies)",
+    ).toBe(true);
+    expect(
+      states.some((s) => !s.talliesCompact),
+      "sweep exercises the full state (a wide tablet keeps full tally labels — no over-compaction)",
+    ).toBe(true);
+  });
 });
 
 describe("TopBar — wide desktop keeps full labels (measurement does NOT over-compact)", () => {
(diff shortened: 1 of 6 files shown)


scoop (commented):
## Summary

The PR extends the already measurement-driven right-cluster compaction so the status tallies (HERD/BUSY/IDLE/BLOCKED) also collapse to their dot+digit form under measured overflow on touch-desktop, and folds the async-arriving held count into the measure effect's tracked content signal so the bar re-measures when a task is held after first paint. Both changes are correct and target the reported bug.

### Verification

- `ui/src/lib/components/TopBar.svelte:594` passes `compact={touch && compactBadges}` to `TopBarTallies`. `compactBadges` is `mode !== "mobile" && measuredCompact` (TopBar.svelte:224), so the tallies now collapse on the same single measured-overflow signal that already drops the clock time (TopBar.svelte:223) and compacts the other badges. This is the missing collapse that previously left full-label tallies overflowing the bar and clipping the gear.
- `ui/src/lib/components/top-bar/TopBarTallies.svelte:32` gates the compact markup on `mobile || compact`; the `compact` prop defaults to `false` (line 9), so phones (`mobile`) and other callers are unaffected. The compact branch reuses the existing `.tallies.compact` markup/styles — verified the compact form already existed and is unchanged structurally.
- The held wiring is consistent end-to-end: `held: heldCount` added to the `chrome` object (TopBar.svelte:127), `held: number` added to `ChromeState` (top-bar-layout.ts:25), and counted in `badgeCount` (top-bar-layout.ts:43). The measure effect reads `void badgeCount(chrome)` (TopBar.svelte:190), so held arrival now re-fires it — matching the gauge-arrival precedent documented at TopBar.svelte:191-197.
- ChromeState construction parity: the only object literals of `ChromeState` are `chrome` (TopBar.svelte:119, now includes `held`) and the test literals in `top-bar-layout.test.ts` (all updated to include `held`, lines 21/51/64/74/86). A missing field would be a TypeScript error; none remain. Verified no other `ChromeState` literal exists (grep across the tree).
- Harness seeding (`initialLimits`/`initialHeldCount` via `untrack`, `setHeld`) is test-only and leaves the existing gauge-arrival test path (`initialLimits` defaults `null`) unchanged.

### Observation (non-blocking)

Tally compaction is gated on `touch && compactBadges`, whereas the sibling controls compact on `compactBadges` alone, which spans desktop too: NEEDS-YOU uses `mobile || compactBadges` (TopBar.svelte:607) and the held badge receives `{compactBadges}` (TopBar.svelte:621). So on a fine-pointer desktop under measured overflow, the clock drops and every other control compacts but the tallies stay full-label. This preserves prior desktop behavior (tallies were never compacted before this PR) and the PR's scope is explicitly the unfolded fold (touch-desktop), so it is not a regression — but it is worth a sanity check that a fine-pointer desktop can't still overflow with full tallies after the other controls have already compacted, since that path keeps the full tallies the touch path now sheds.

scoop:
Critic point (tally compaction gated on `touch &&`): **valid, fixed.** Probed it — a narrow fine-pointer desktop window does overflow with full-label tallies after the right side compacts (desktop renders both usage gauges inline, wider than touch's single collapsed gauge-btn, so it overflows even *earlier* than touch-desktop — confirmed up to ~1100px with held+needsYou+gauges). Changed the gate to `compactBadges` alone (consistent with the sibling NEEDS-YOU / held badges) and scoped the compact tally's 44px floor to coarse pointers so fine-pointer desktop gets the WCAG 24px AA target instead. Added a font-robust desktop coupling sweep (narrow collapses, wide stays full, no overflow at any width). c88043c9.

scoop (commented):
## Re-review

The diff extends measured-overflow compaction to the status tallies and folds the async `held` count into the measure-effect's tracked content signal.

### Prior finding #1 — tally compaction gated on `touch && compactBadges` (ui/src/lib/components/TopBar.svelte)
**Addressed.** `TopBar.svelte:594` now passes `compact={compactBadges}` and `TopBarTallies.svelte:33` renders the compact form on `{#if mobile || compact}`. Compaction is now driven by the measured-overflow signal alone, consistent with the sibling controls:
- NEEDS-YOU: `TopBar.svelte:607` → `class:compact={mobile || compactBadges}`
- held badge: `TopBar.svelte:621` → `{compactBadges}`

A narrow fine-pointer desktop window under measured overflow now collapses its tallies instead of overflowing with full labels. No longer applies; not re-raised.

### Verifications
- **Held reactivity:** `held: heldCount` added to `chrome` (TopBar.svelte:127). `badgeCount(chrome)` is read by the measure effect (TopBar.svelte:190); `held` is counted in `top-bar-layout.ts:42` (`s.held > 0 ? 1 : 0`). When `heldCount` goes 0→N async, `badgeCount` now changes and the effect re-fires — the reactivity gap is genuinely closed. `heldCount` is a real prop (TopBar.svelte:54,90).
- **ChromeState contract:** `held: number` added as required (top-bar-layout.ts:23). The only production constructor is `chrome` (TopBar.svelte:119-128), which provides it; grep found no other consumers of `ChromeState`/`badgeCount` outside tests. Consistent.
- **WCAG floor (TopBarTallies.svelte:202-215):** base `.ctally` floor lowered to 24px on both axes; `@media (pointer: coarse)` restores 44px height. Coarse pointers (phones, foldables, touch-desktop) keep the prior 44px height; fine-pointer desktop — the newly-reachable compact case — gets the WCAG 2.5.8 AA 24px target. No regression for existing coarse-pointer renders.
- **i18n:** the compact branch reuses existing `m.topbar_tally_*` keys; no new strings introduced, so no catalog parity concern.

No new defects found.

PR #1121, opened 2026-06-26
fix(ui): single post-merge toast, drop auto local-checkout offer

scoop:
## Why

After a PR merges, the UI stacked **two competing 15s toasts** at once:

1. `Merged {name}` + **Decommission**
2. `PR merged. Update your local default branch?` + **Update** (isolated sessions)

Now that new task spawns always start from `origin` HEAD of the default branch, a stale local checkout no longer blocks anything, so the second offer is low-value. This collapses the post-merge moment to a **single, clear next step** — Decommission.

## What changed

- **Manual-merge paths** (`GitRail.svelte`, `PrRow.svelte`): drop the auto update-local offer; the only post-merge toast is `Merged …` + Decommission.
- **Merge-train path** (`store.svelte.ts`, `mergetrain:landed` — previously the *sole* signal a train landed): replaced the update offer with a plain repo-keyed `Merge train landed in {repo}` info toast.
  - ⚠️ **Behavior change (not just copy):** this path **loses its one-click fast-forward _action_** — the replacement is a no-action info toast. Updating the local checkout after a train lands remains possible via BacklogView's existing **Fast-forward** button (`pullMainAndToast` is retained untouched).
- Deleted the now-dead `offerUpdateMain` helper and its orphaned `isolated`/`baseBranch` GitRail props (removed from both `GitRail` and its sole `Viewport` call site, in lockstep, to avoid an excess-prop error).
- Removed the orphaned `toast_update_main_offer` / `toast_update_main_action` message keys (EN+DE); added `toast_mergetrain_landed`. All other `toast_update_main_*` keys stay (used by the surviving manual pull).

### Deviation notes (house rules)
- The merge-train confirmation names the repo by its **path basename** (via the existing `store.svelte.ts` `.split("/").pop()` idiom) rather than a PR number, because the `mergetrain:landed` event carries no PR number and a train lands a *batch*.
- Shipped as `fix(ui):` — this is a refinement/removal, not a new capability, so no What's-New feature-catalog entry (`[no-feature-entry]` not needed; the gate only arms on `feat`). "Merge train" is already an established product term, so no glossary entry.

## Verification
- `cd ui && bun run check` — 0 errors
- `cd ui && bun run check:i18n` — locales in parity
- `cd ui && bun run test` — 2076 passed
- `bunx fallow audit --base origin/main --fail-on-issues` — no issues in changed files

[no-feature-entry]

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -694,8 +694,6 @@
   "toast_merge_train_no_prs": "Keine merge-bereiten PRs für einen Merge-Train vorhanden",
   "toast_merge_train_failed": "Merge-Train konnte nicht gestartet werden",
   "toast_merged": "{name} gemergt",
-  "toast_update_main_offer": "PR gemergt. Deinen lokalen Standard-Branch aktualisieren?",
-  "toast_update_main_action": "Aktualisieren",
   "toast_update_main_done": "Lokalen {branch} auf den neuesten Stand gebracht",
   "toast_update_main_uptodate": "Lokaler {branch} ist bereits aktuell",
   "toast_update_main_wrong_branch": "Lokaler Checkout ist nicht auf {branch}; unverändert gelassen",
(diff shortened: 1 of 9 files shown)


scoop (commented):
## Summary

This PR resolves the dual post-merge-toast problem by removing the local-checkout "update your default branch?" offer (`offerUpdateMain`) entirely, leaving only the decommission offer on single-PR merges and replacing the merge-train offer with a plain confirmation toast. This satisfies the task's "get rid of it altogether" option.

The implementation is clean and internally consistent. I verified every dependency the diff touches:

- **`offerUpdateMain` removal is complete** — grep across the whole tree finds zero remaining references after deletion from `ui/src/lib/pull-offer.ts`, `GitRail.svelte`, `PrRow.svelte`, and `store.svelte.ts`.
- **Orphaned message keys** — removed `toast_update_main_offer` and `toast_update_main_action` have no remaining consumers (grep clean). The still-referenced `toast_update_main_done`/`_uptodate`/`_wrong_branch` keys are correctly retained because `pullMainAndToast` still uses them (ui/src/lib/pull-offer.ts:13-23).
- **i18n parity** — new key `toast_mergetrain_landed` exists in both catalogs with the matching `{repo}` placeholder (verified ui/messages/en.json:1999 and ui/messages/de.json:1999). These are the only two catalogs in `ui/messages/`.
- **`pullMainAndToast` retained and reachable** — the comment's claim that the checkout stays updatable on demand is true: BacklogView's Fast-forward button calls it (verified ui/src/lib/components/BacklogView.svelte:135, `handleFf`).
- **Prop removal is safe** — `isolated`/`baseBranch` removed from GitRail props and Viewport's call site; grep confirms no remaining use of either inside GitRail.svelte, and no other GitRail consumer passed them.
- **New toast shape matches tests** — `confirmMergeTrainLanded` (ui/src/lib/store.svelte.ts:585-589) emits a keyed `toasts.info` with no action; the updated store test asserts exactly that (`actionLabel` undefined, `text` contains the repo basename). `toasts.items` is a public writable `$state` field (verified ui/src/lib/toasts.svelte.ts:64), so the test's `toasts.items = []` is valid, and `text`/`actionLabel` are real Toast properties (ui/src/lib/toasts.svelte.ts:15,32).

No correctness, security, or quality issues found.

PR #1122, opened 2026-06-26
fix(ui): distinguish held vs update badges with hourglass/bolt icons

kai-osthoff:
## Problem

In the top bar, the **held-tasks badge** ("zurückgehalten") and the **app-update badge** ("Update") are both amber and, in their compact form, collapse to a tiny box with a number — `[5]` vs `[▲ 1]`. They're easy to confuse at a glance.

## Change

Give each a distinct, meaning-bearing icon (Lucide-style inline SVG, `currentColor`, so it inherits the badge's amber tint and stays monochrome):

- **Held / zurückgehalten** → ⌛ **hourglass** (tasks waiting to resume)
- **Update** → ⚡ **lightning bolt** (actionable self-update)

The blue **herdr** badge is left unchanged — it's already color-distinct and wasn't part of the confusion.

Applied in both the desktop badge cluster (`TopBarBadges`, `TopBarHeldBadge`) and the mobile gear sheet (`TopBarMobileSheet`).

## Verification

- `bun run check` — 0 errors / 0 warnings.
- `TopBar` browser suite 62/62 and `GitRail` 56/56 pass in isolation.
- Headless render confirms the two amber badges are now instantly distinguishable.

No new strings (icon-only), so no i18n/feature-catalog entries needed. `fix`, not `feat`.

### DIFF
--- a/ui/src/lib/components/top-bar/TopBarBadges.svelte
+++ b/ui/src/lib/components/top-bar/TopBarBadges.svelte
@@ -70,8 +70,10 @@
 {/if}
 <!-- Desktop keeps the inline HERDR badge; on a phone it folds into the gear
      bottom sheet to free a slot in the single-row control cluster. The
-     touch-desktop badge crunch drops the label to a bare ▲ (aria-label keeps
-     it named) so two stacked update badges still fit. -->
+     touch-desktop badge crunch drops the label to a bare up-arrow (aria-label
+     keeps it named) so two stacked update badges still fit. Distinct glyph from
+     the app-update bolt: an up-arrow reads as "newer version available"
+     (informational/manual), matching the calm blue variant. -->
 {#if herdrUpdateAvailable}
   <button
     class="update-badge herdr"
@@ -82,7 +84,19 @@
       latest: herdrUpdate!.latest ?? "?",
     })}
   >
-    <span class="up-dot">▲</span>
+    <svg
+      class="up-glyph"
+      viewBox="0 0 24 24"
+      fill="none"
+      stroke="currentColor"
+      stroke-width="2"
+      stroke-linecap="round"
+      stroke-linejoin="round"
+      aria-hidden="true"
+    >
+      <path d="M12 19V5" />
+      <path d="m5 12 7-7 7 7" />
+    </svg>
     {#if !compactBadges}<span class="up-label">{m.topbar_herdr_update_badge()}</span>{/if}
   </button>
 {/if}
@@ -172,9 +186,6 @@
     height: 1.15em;
     display: block;
     flex-shrink: 0;
-  }
-  .update-badge .up-dot {
-    font-size: var(--fs-micro);
   }
   .update-badge .up-n {
     font-variant-numeric: tabular-nums;
--- a/ui/src/lib/components/top-bar/TopBarBadges.svelte
+++ b/ui/src/lib/components/top-bar/TopBarBadges.svelte
@@ -52,7 +52,18 @@
       ? m.updatemodal_commits_one()
       : m.updatemodal_commits_other()}"
   >
-    <span class="up-dot">▲</span>
+    <svg
+      class="up-glyph"
+      viewBox="0 0 24 24"
+      fill="none"
+      stroke="currentColor"
+      stroke-width="2"
+      stroke-linecap="round"
+      stroke-linejoin="round"
+      aria-hidden="true"
+    >
+      <path d="M13 2 3 14h9l-1 8 10-12h-9l1-8Z" />
+    </svg>
     {#if !compactBadges}<span class="up-label">{m.topbar_update_badge()}</span>{/if}
     <span class="up-n">{update!.behind}</span>
   </button>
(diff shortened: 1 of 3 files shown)


kai-osthoff (commented):
## Summary

The PR addresses the task: held tasks and updates both render as amber badges and looked nearly identical. The change adds a distinguishing glyph to each — a lightning bolt for the update badge and an hourglass for the held badge.

- **TopBarBadges.svelte:55-66** — the update badge's `▲` `up-dot` span is replaced by a `zap`/lightning-bolt SVG (`path d="M13 2 3 14h9l-1 8 10-12h-9l1-8Z"`), with `.update-badge .up-glyph` sizing at `1.15em` (TopBarBadges.svelte:170-175).
- **TopBarHeldBadge.svelte:132-145** — an hourglass SVG (two end-bars + two triangles) is prepended to the held badge, with `.held-badge .held-glyph` sizing at `1.15em` (TopBarHeldBadge.svelte:267-272).
- **TopBarMobileSheet.svelte:250-261** — the sheet's update row swaps its `▲` for the same bolt SVG, sized via `.sheet-svg` (TopBarMobileSheet.svelte:553-556).

This directly satisfies the request (Sanduhr vor zurückgehaltene, Blitz für Updates).

## Verification

- Both badges are amber (`var(--color-amber)`): held at TopBarHeldBadge.svelte:229-231, update at TopBarBadges.svelte:159-161 — confirming the original "look too similar" complaint and that the glyphs are the differentiator. SVGs use `stroke="currentColor"`, so they inherit the amber tint correctly.
- No dead CSS: `.update-badge .up-dot` (TopBarBadges.svelte:176) is still consumed by the herdr badge's `▲` at TopBarBadges.svelte:85 (verified via grep — only the herdr span and the CSS rule remain).
- Accessibility intact: every new SVG carries `aria-hidden="true"`, and the host buttons keep their `aria-label` (TopBarBadges.svelte:51-53 title, TopBarHeldBadge.svelte:129, TopBarMobileSheet.svelte:248).
- No new i18n keys introduced — `m.topbar_update_badge` / `m.topbar_held_badge` are pre-existing and unchanged, so no locale-parity concern.
- The held mobile dialog lives inside TopBarHeldBadge.svelte (not the mobile sheet), so the single component change covers the held affordance in all layouts; no missed held render site.

## Optional nit (non-blocking)

Within TopBarBadges.svelte (and TopBarMobileSheet.svelte) the **herdr** update badge still renders `▲` (TopBarBadges.svelte:85, TopBarMobileSheet.svelte:277). After this change the amber app-update badge shows a bolt while the adjacent herdr-update badge keeps the triangle. This is defensible — herdr is intentionally styled as informational/blue and non-pulsing (TopBarBadges.svelte:186-191) — but if the goal is consistent iconography across "update" affordances, consider whether the herdr triangle should also be reconsidered. Purely a design judgment; not required by the task.

kai-osthoff:
Addressed the herdr-badge consistency point: converted it from the text-glyph `▲` to an inline SVG **up-arrow** (matching the all-SVG, `currentColor` icon language now used by its neighbours). Kept it semantically distinct from the actionable lightning bolt — ↑ reads as "newer version available" (informational/manual), reinforced by the calm blue variant. Final three-way language: ⌛ held, ⚡ app self-update, ↑ herdr version-available. Removed the now-orphaned `.up-dot` CSS.

kai-osthoff (commented):
## Review

The PR distinguishes held vs. update badges with distinct SVG icons, satisfying the task (hourglass before held, bolt before updates).

### Changes verified
- `ui/src/lib/components/top-bar/TopBarHeldBadge.svelte:132-145` — adds an hourglass SVG (`currentColor` stroke) rendered before both the compact count and the full label. Sized via new `.held-glyph` rule (TopBarHeldBadge.svelte:267-272).
- `ui/src/lib/components/top-bar/TopBarBadges.svelte:55-65` — app-update badge now uses a lightning-bolt SVG; `TopBarBadges.svelte:87-99` — herdr badge converted from the text glyph `▲` to an inline up-arrow SVG. The now-orphaned `.up-dot` rule was replaced by `.up-glyph` (TopBarBadges.svelte:184-189).
- `ui/src/lib/components/top-bar/TopBarMobileSheet.svelte:250-260, 277-289` — sheet rows mirror the bolt/up-arrow SVGs; new `.sheet-svg` rule (line 565) sets height, with `.sheet-glyph`'s width supplying the square box.

### Prior finding (re-review)
1. herdr update badge still using `▲` `up-dot` — **Resolved.** Verified at `TopBarBadges.svelte:87-99`: it is now an inline up-arrow SVG using the all-SVG `currentColor` icon language of its neighbours. Confirmed `.up-dot` is fully removed — `grep "up-dot" ui/src/` returns only an unrelated historical comment in TopBar.svelte:853, no live references.

### Correctness checks
- No new message keys introduced; `topbar_held_badge`, `topbar_update_badge`, `topbar_herdr_update_badge` are pre-existing and present in both `ui/messages/en.json` and `ui/messages/de.json` (i18n parity intact).
- SVGs use square `viewBox="0 0 24 24"` with `aria-hidden="true"`; button `aria-label`s remain, so accessible naming is preserved. Sizing relies on `em`/`--fs-lg` boxes — proportional fill is correct for square viewBoxes.
- The held badge in the mobile context is rendered by `TopBarHeldBadge.svelte` (no held row exists in `TopBarMobileSheet.svelte`, confirmed by grep), so no icon is missing there.

Nothing to address.

PR #1123, opened 2026-06-26
fix(automation): Auto-Abarbeitung-Regler inline unter den Schalter + Felddoku

kai-osthoff:
## Problem

Beim Aktivieren von **Auto-Abarbeitung** (Auto-Drain) erschienen die zugehörigen Einstell-Regler — **Limit**, **Label**, **Grenze %** — weit unten im Panel: erst nach Build-Queue, Entwurfsmodus, Sandbox und Standardmodell. Dadurch war nicht erkennbar, dass sie zur Auto-Abarbeitung gehören.

## Änderung

- Die drei Regler werden jetzt **direkt unter dem Auto-Abarbeitung-Schalter** eingeblendet — und nur, solange Auto-Abarbeitung tatsächlich aktiv ist (an, nicht durch ein Epic ausgesetzt, Forge-Repo). Beim Ausschalten klappen sie wieder weg.
- **Mehr Dokumentation:** ein einleitender Satz rahmt die drei Felder als die Regler der Auto-Abarbeitung, und jedes Feld hat eine eigene kurze Erklärung (was Limit / Label / Grenze % bedeuten), per `aria-describedby` mit dem Eingabefeld verknüpft.

Umgesetzt durch Auslagern der Regler in eine eigene kleine Komponente `AutomationDrainFields.svelte`, inline gerendert in `AutomationSettings.svelte`. `AutomationRepoFields` behält nur noch Standardmodell + Zuständigkeiten.

## Verifikation

- `bun run check` (svelte-check) — 0 Fehler / 0 Warnungen
- `bun run check:i18n` — EN/DE in Parität (neue Keys beidseitig ergänzt)
- Automation-Tests grün (`git-rail-automation`, `automation-panel-epic`)
- Live geprüft: Panel rendert, Regler erscheinen/verschwinden mit dem Schalter

Reine UX-Verbesserung an bestehender Funktion, kein neues Feature → kein Feature-Katalog-Eintrag (`fix`, nicht `feat`).

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2006,5 +2006,9 @@
   "feat_herdr_update_session_list_title": "Sieh die Sessions, die ein Update beenden würde — und spring hin",
   "feat_herdr_update_session_list_body": "Der herdr-Update-Dialog zeigt nicht mehr nur eine Zahl laufender Sessions, sondern listet jede einzeln auf — und jede Zeile springt direkt zu dieser Session. Schließe sie nach und nach ab und aktualisiere dann, ohne Angst, etwas zu verlieren.",
   "topbar_held_spawning": "Wird gestartet…",
-  "topbar_held_discarding": "Wird verworfen…"
+  "topbar_held_discarding": "Wird verworfen…",
+  "automation_drain_fields_intro": "Diese Regler steuern die Auto-Abarbeitung und gelten nur, solange sie aktiv ist:",
+  "drain_cap_hint": "Wie viele auto-gestartete Agenten höchstens gleichzeitig laufen. Ist das Limit erreicht, wartet Shepherd auf einen freien Platz, bevor das nächste Issue aufgegriffen wird.",
+  "drain_label_hint": "Nur Issues mit diesem Label werden aufgegriffen — versieh die Tickets, die automatisch abgearbeitet werden sollen, damit.",
+  "drain_ceiling_hint": "Pausiert neue Auto-Starts, sobald die Konto-Auslastung diesen Prozentwert erreicht. Ein höherer Wert lässt die Abarbeitung länger weiterlaufen; Standard ist 80."
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2006,5 +2006,9 @@
   "feat_herdr_update_session_list_title": "See and visit the sessions an update would end",
   "feat_herdr_update_session_list_body": "The herdr-update dialog no longer just shows a count of running sessions — it lists each one, and every row jumps straight to that session. Wrap them up one by one, then update without worrying you've lost work.",
   "topbar_held_spawning": "Starting…",
-  "topbar_held_discarding": "Discarding…"
+  "topbar_held_discarding": "Discarding…",
+  "automation_drain_fields_intro": "These dials govern Auto-Drain and apply only while it's on:",
+  "drain_cap_hint": "Most auto-started agents allowed to run at once. Once it's reached, Shepherd waits for a slot to free before picking up the next issue.",
+  "drain_label_hint": "Only issues carrying this label get picked up — tag the tickets you want worked automatically with it.",
+  "drain_ceiling_hint": "Pause new auto-starts once your account usage reaches this percentage. A higher value lets the drain keep going longer; the default is 80."
 }
--- a/ui/src/lib/components/AutomationSettings.svelte
+++ b/ui/src/lib/components/AutomationSettings.svelte
@@ -387,7 +387,7 @@
     <label class="drain-field">
       <span class="drain-label">{m.automation_signoff_authority_label()}</span>
       <select
-        class="num signoff-select"
+        class="afield-num signoff-select"
         aria-label={m.automation_signoff_authority_label()}
         value={repoConfig.signoffAuthorityFor(repoPath)}
         onchange={(e) =>
@@ -416,7 +416,7 @@
   <label class="drain-field" use:coachIf={{ id: "sandbox-profile", on: armCoachmarks }}>
     <span class="drain-label">{m.automation_sandbox_profile_label()}</span>
     <select
-      class="num sandbox-select"
+      class="afield-num sandbox-select"
       aria-label={m.automation_sandbox_profile_label()}
       value={repoConfig.sandboxProfileFor(repoPath)}
       onchange={(e) =>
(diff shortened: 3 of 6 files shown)


kai-osthoff:
## Summary

The PR achieves the requested task well: the Auto-Drain rails are extracted into a new `AutomationDrainFields.svelte` and rendered inline directly under the Auto-Drain toggle (AutomationSettings.svelte:309), with an intro sentence and per-field hint docs. i18n parity is clean and the new component's visibility gate `drainRailsActive` (AutomationSettings.svelte:51) faithfully mirrors the toggle's own on-condition `flags.autoDrain && !epicActive && !lightweight` (AutomationSettings.svelte:289). Identifiers all resolve: `clampCap`/`clampCeiling`/`sanitizeLabel` (ui/src/lib/components/git-rail-drain.ts:5,10,15) and the `repoConfig` accessors/mutators (ui/src/lib/reviews.svelte.ts:438,446,454,544,548,552).

However, the CSS refactor introduces a real visual regression.

## Blocking: global `.num` rule leaks into unrelated components

The extracted `automation-fields.css` is imported via `import "./automation-fields.css"` (AutomationDrainFields.svelte / AutomationRepoFields.svelte / AutomationSettings.svelte). A plain `.css` import in Vite/SvelteKit is a **global** stylesheet — it is NOT run through Svelte's per-component scoping. So `.num` (ui/src/lib/components/automation-settings/automation-fields.css:25) and `.num.txt` become document-wide rules.

The file's own comment (automation-fields.css:4-5) asserts "These class names are used only by the automation-settings components, so global scope here is safe." That is false. A bare `class="num"` is used by at least five other components:
- EpicPanel.svelte:40 (`<a class="num">#…</a>`), scoped `.num` at :183 sets only color/font-size/text-decoration/flex-shrink
- EpicGroupHeader.svelte:48 (`<span class="num">`), scoped `.num` at :152 sets only flex/color
- IntegratedEpicRow.svelte:70 (`<span class="num">`), scoped `.num` at :188 sets only flex/color
- MergeTrainConfirmDialog.svelte:57 (`<span class="num">`), scoped `.num` at :140 sets only font/letter-spacing/color
- Settings.svelte:843,866,891,925 (`<input class="num">`) with **no** scoped `.num` at all

Before this PR, `.num` existed only as *scoped* rules inside AutomationSettings.svelte and AutomationRepoFields.svelte, so it never reached these components; the only global `.num` in the tree is now the new file (verified: grep of `*.css` finds `.num` solely in automation-fields.css:25,39). Svelte scoping gives those components' own `.num` a specificity of (0,2,0) vs the global's (0,1,0), so properties they redeclare (color/font-size) still win — but every property they DON'T declare leaks in: `width:90px`, `border`, `background:var(--color-panel)`, `border-radius`, `font-family:var(--font-mono)`, `padding:3px 6px`, `text-align:right`. The inline issue-number chips/links in the epic and merge-train components will render as 90px-wide bordered mono input boxes, and the Settings number inputs get restyled too. Once any automation component module loads, the rule is injected document-wide for the session, so this is reachable in normal use.

Fix: scope these styles so they don't leak — e.g. namespace the class (`.drain-num` / a wrapper-prefixed selector) for the automation fields, or keep the shared block but use selectors that can't collide. `.drain-fields`/`.drain-field`/`.drain-label` are unique to these components (verified via grep) and are not affected; the problem is specifically the generic `.num`/`.num.txt`.

## Nit

AutomationDrainFields.svelte uses fixed element IDs `drain-hint-cap`/`drain-hint-label`/`drain-hint-ceiling` (lines 73,87,101) referenced by `aria-describedby`. If two `AutomationSettings` instances are ever mounted at once (it is used from AutomationPanel, FirstTaskAutomationConfirm, and BacklogTabContent), the IDs collide and the aria reference becomes ambiguous. Consider a per-instance id suffix.

kai-osthoff (commented):
## Re-review

The PR moves the Auto-Drain rails (cap / label / usage ceiling) into a new `AutomationDrainFields.svelte` rendered inline directly beneath the Auto-Drain toggle (AutomationSettings.svelte:309), adds an intro sentence plus per-field hint copy, and hoists the shared field-row CSS into `automation-fields.css`. This satisfies the task (fields now live at the setting, gated visible/hidden in place, with added documentation).

### Prior-revision points — both addressed

1. **`.num` global collision** — Resolved. The shared, globally-imported stylesheet now names the input-chrome class `.afield-num` (automation-fields.css:34, `.afield-num.txt` at :45), and every automation-settings input was renamed accordingly (AutomationDrainFields.svelte:55/72/89, AutomationSettings.svelte:390/419, AutomationRepoFields.svelte:87). The other `.num` consumers each keep a component-*scoped* `.num` rule — verified at EpicPanel.svelte:183, EpicGroupHeader.svelte:152, MergeTrainConfirmDialog.svelte:140, IntegratedEpicRow.svelte:188, Settings.svelte:1226 — so the global `.afield-num` cannot leak onto them. The remaining global names `.drain-fields`/`.drain-field`/`.drain-label` are used only by the three automation-settings files (grep across ui/src confirms no other consumers), so no new collision is introduced.

2. **Static aria-describedby IDs** — Resolved. `AutomationDrainFields.svelte:18` derives `uid = $props.id()` and the inputs/hints use per-instance IDs `{uid}-cap` / `{uid}-label` / `{uid}-ceiling` (lines 57/64, 74/81, 91/98). `$props.id()` is supported by the pinned svelte `^5.56.4` (ui/package.json:32).

### Other verification

- i18n parity: all four new keys exist in both catalogs — `automation_drain_fields_intro`, `drain_cap_hint`, `drain_label_hint`, `drain_ceiling_hint` (en.json:101-104, de.json:101-104). `m.automation_autodrain_name()` used for the group aria-label exists (en.json:130, de.json:130).
- `clampCap`/`clampCeiling`/`sanitizeLabel` imported from `../git-rail-drain` resolve to git-rail-drain.ts:5/10/15 (same relative path the file used before the move).
- `drainRailsActive = flags.autoDrain && !epicActive && !lightweight` (AutomationSettings.svelte:48) exactly mirrors the Auto-Drain switch's own on-state (AutomationSettings.svelte:289/292), so the rails now show only when the switch actually reads as on — an improvement over the old `{#if autoDrain}` gating.
- The removed `autoDrain` prop on `AutomationRepoFields` has no remaining caller passing it (only consumer is AutomationSettings.svelte:448, now `{repoPath}` only).
- CSS custom properties used by the new hint styles exist: `--fs-micro` (app.css:81), `--color-muted` (app.css:19), `--color-faint` (app.css:20).

### Nit

- `ui/src/lib/components/automation-settings/AutomationDrainFields.svelte:110` — the `<style>` comment says the shared layout `.drain-fields / .drain-field / .drain-label / .num` comes from automation-fields.css, but that file no longer defines `.num` (it's `.afield-num`, automation-fields.css:34). The sibling comments at AutomationSettings.svelte:620 and AutomationRepoFields.svelte:178 already say `.afield-num`; update this one to match.

PR #1126, opened 2026-06-26
ci(automerge): harden doc-automerge on author identity + path scope; fix checkout-less git failure (#1125)

### LINKED ISSUE #1125: Harden doc-automerge: gate on author identity + path scope, not branch prefix alone
Follow-up from the merge train review of #1120 (`.github/workflows/doc-automerge.yml`).

## Gap
The workflow auto-merges any green PR whose head branch matches the `shepherd/docs-update-*` prefix. Identity is established by **branch-name prefix only** — there is no `authorAssociation` check and no path-scoping to the doc tree.

A fork or outside-collaborator PR named `shepherd/docs-update-*` containing **arbitrary file changes** would auto-merge if CI is green and no human leaves a CHANGES_REQUESTED review.

## Why it's safe today (but not later)
Bounded by: private repo + branch protection + green-CI gate. These collapse on the planned open-source flip (#871).

## Harden before open-sourcing
- [ ] Require `authorAssociation` ∈ {OWNER, MEMBER} (i.e. Shepherd's own bot/account), not just the branch prefix.
- [ ] Path-scope: assert the PR diff touches only the doc tree; abort the merge if any non-doc path is modified.

Surfaced by the read-only reviewer during the #1120 merge-train pass; #1120 itself merged as-is (correct design layer, safe under current constraints).

scoop:
Closes #1125. Also fixes the broken doc-automerge run [`28224241163`](https://github.com/erwins-enkel/shepherd/actions/runs/28224241163/job/83612295960).

Two independently-revertable commits in `.github/workflows/doc-automerge.yml`:

## Commit 1 — fix the blocking CI failure (operational)
The workflow had **no `actions/checkout`**, so `gh repo view` (line 54) fell back to local git and died with `fatal: not a git repository`, exiting the run **before any PR was evaluated** — doc-automerge was 100% broken. Added a shallow `actions/checkout@v7` (the repo's pinned convention in `ci.yml`/`pr-hygiene.yml`) as the job's first step, restoring git context deterministically for every `gh` call.

**Why checkout over a surgical `gh api` swap:** the run died at `gh repo view` *before* the first `gh pr list`, so there is no in-repo evidence that `gh pr list/view/merge` work checkout-less — only gh's documented `GH_REPO` support. (The repo's one checkout-less precedent, `onboarding-release-gate.yml`, uses `gh api`, not `gh pr`.) A checkout removes the per-command reasoning entirely.

## Commit 2 — identity + path-scope gates (#1125)
The branch prefix `shepherd/docs-update-*` was the **only** provenance signal: a fork / outside-collaborator PR wearing the prefix with arbitrary file changes would auto-merge once green. Three new skip-gates run **first** in the per-PR loop:

- **`authorAssociation` ∈ {OWNER, MEMBER}** — Shepherd's own account. REST-only field, fetched via `gh api .../pulls/$n`.
- **`isCrossRepository == false`** — blocks fork heads.
- **diff confined to the doc tree** (`docs/` + `docs-site/`), **fail-closed** when the ≤100-file list is truncated.

Ineligible PRs are **skipped (logged), not failed red** — consistent with every existing gate; the daily-cron cadence would otherwise turn a lingering suspicious PR into perpetual red noise, and skipping already meets the security goal (never merge it).

### Deliberate deviations (flagged per house rules — also in code comments)
1. **`isCrossRepository` is beyond the issue's two literal checkboxes.** `authorAssociation` alone does **not** block a fork — a MEMBER can open a PR *from their own fork* and still read `MEMBER`. This is the direct, zero-false-positive answer to the issue's stated "a fork PR" threat (all legit doc PRs are same-repo).
2. **Path scope uses coarse `docs/`+`docs-site/` prefixes, not doc-agent.ts's exact 8-path `IN_SCOPE_PATHS`.** Robust to in-scope-list edits with no YAML↔TS drift, while still blocking any non-doc path (`src/`, `.github/`, `package.json`, …). The agent's exact list stays the structural staging boundary; this CI gate is a deliberately coarser, independent safety net.
3. **Truncation guard (`changedFiles != returned-files → skip`)** is an extra fail-closed defense not named in the issue (gh caps file lists at 100). Doc PRs touch ≤8 files, so it never affects legitimate merges.

## Verification
- `bash -n` on the extracted script — passes (pre- and post-rebase).
- jq-logic harness over real + synthetic fixtures (9/9): doc PR `#1111`/`docs/*` → **MERGE**; code PR `#1123` shape → path-reject; fork → cross-reject; `CONTRIBUTOR`/`NONE` → identity-reject; truncated/empty list → reject; pending CI still held by the existing notgreen gate.
- Confirmed live that real doc PRs (`#1111`, `#1077`, `#1065`; author `scoop`) are `MEMBER` / same-repo / doc-tree-only — the gates pass every legitimate PR.
- **Commit 1 can't be proven by the jq harness** (it never runs `gh`). After merge I'll trigger a `workflow_dispatch` run to confirm it completes past line 54 with no git error; otherwise it's validated by the next live trigger.

### DIFF
--- a/.github/workflows/doc-automerge.yml
+++ b/.github/workflows/doc-automerge.yml
@@ -37,6 +37,10 @@ jobs:
     name: merge green doc PRs
     runs-on: ubuntu-latest # a tiny control job — kept off the self-hosted CI runner on purpose
     steps:
+      # Give the job git context: `gh repo view` (and gh's repo auto-detection generally) falls back
+      # to the local git remote, so without a checkout it dies with "fatal: not a git repository".
+      # A shallow checkout fixes every gh call deterministically — cheaper than per-command reasoning.
+      - uses: actions/checkout@v7
       - name: Merge eligible doc PRs
         env:
           GH_TOKEN: ${{ github.token }}
--- a/.github/workflows/doc-automerge.yml
+++ b/.github/workflows/doc-automerge.yml
@@ -3,6 +3,11 @@ name: Auto-merge doc PRs
 # Auto-merges Shepherd's OWN documentation PRs — the ones the doc agent opens on
 # `shepherd/docs-update-*` branches — once they're green, instead of waiting on a human merge.
 #
+# Provenance is NOT established by the branch prefix alone (#1125): the per-PR loop also requires
+# authorAssociation ∈ {OWNER, MEMBER}, a same-repo (non-fork) head, and a diff confined to the doc
+# tree (`docs/` + `docs-site/`). A fork / outsider PR wearing the prefix but carrying arbitrary
+# changes is skipped, never merged.
+#
 # Why a workflow and not a Shepherd setting: the doc agent runs ONLY against this repo (its
 # in-scope file list + DOC_TREE_MARKER are this repo's own doc tree), and Shepherd lives only on
 # GitHub — so a per-repo Shepherd toggle would be a dead switch in every other repo. This keeps the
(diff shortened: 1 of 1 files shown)


scoop (commented):
## Summary

This PR hardens `.github/workflows/doc-automerge.yml` exactly as issue #1125 asks, and goes one defensible step further. It satisfies both checkboxes and is free of blocking bugs.

## What the diff does

1. **`authorAssociation` gate** — `assoc="$(gh api "repos/$GH_REPO/pulls/$n" --jq '.author_association')"`, skip unless `OWNER`/`MEMBER` (.github/workflows/doc-automerge.yml:98,127). `author_association` is a valid field on the REST `GET /repos/{owner}/{repo}/pulls/{number}` response, so the lookup is sound. Implements checkbox 1.
2. **Fork gate** — `isCrossRepository == false` (lines 84,128). Beyond the issue's two boxes, but correctly reasoned: a MEMBER can open a PR from their own fork and still read `MEMBER`, so the author check alone does not stop the issue's literal "fork PR" threat. Good defense-in-depth.
3. **Path scope** — counts files whose `path` does not start with `docs/` or `docs-site/` and skips if any exist (lines 106-107,130). Implements checkbox 2.
4. **Checkout step** — `actions/checkout@v7` added to give `gh` git context (line 48).

## Verification

- **Path-scope coverage is complete.** Every entry in the doc agent's `IN_SCOPE_PATHS` is under `docs/` or `docs-site/` (verified against src/doc-agent.ts:97-108), and the staging code only ever `git add`s in-scope files (src/doc-agent.ts:756). So the coarse `docs/`+`docs-site/` prefix check accepts every legitimate doc-PR file and rejects everything else (src/, .github/, lockfiles). The comment's claim of robustness-without-drift holds.
- **Checkout version is consistent.** `actions/checkout@v7` matches the rest of the repo's workflows (verified against .github/workflows/ci.yml:40, ci.yml:148, .github/workflows/pr-hygiene.yml:26), so it is not a typo'd/non-existent version.
- **Gate logic is fail-closed.** All four new gates are conjunctive `continue` skips evaluated before the CI-eligibility gates. A null/unexpected value (e.g. `assoc="null"`, `cross="null"`) fails the equality test and skips rather than merging. The truncation guard (`changed -eq 0 || nfiles != changed`, line 129) and the `set -euo pipefail` propagation on the `gh api` substitution (line 98) both fail closed (skip or red, never silent merge). The security invariant — never auto-merge an untrusted or out-of-scope PR — is upheld on every path I traced.
- **Threat coverage.** Outside collaborator → blocked by author gate; MEMBER-from-fork → blocked by cross-repo gate; smuggled non-doc path → blocked by path gate (incl. the >100-file truncation case). A same-repo MEMBER touching only docs still merges — which is the intended trusted path.

## Limitations (could not verify against ground truth in this repo)

- `gh pr view --json` field validity for `isCrossRepository`, `files`, `changedFiles` and the documented 100-file cap on the `files` connection are gh-CLI runtime behaviors, not present in this repo — assumed as written. They are not findings. Note that even if `gh` were to paginate `files` fully rather than cap at 100, the truncation guard would simply never fire, which remains safe.
- The comment's claim that `authorAssociation` is "not exposed by `gh pr view --json`" could not be confirmed against a schema in-repo; the REST fallback used here is correct regardless, so this is at most a documentation nuance, not a defect.

No changes required.

PR #1127, opened 2026-06-26
refactor(ui): rename Backlog pane to Repos

scoop:
## What

Renames the user-facing **Backlog** pane to **Repos**. The pane is a per-repository hub (left list of repos → Issues / PRs / Actions / Readiness / Automation for the selected repo), so "Repos" describes it more accurately than "Backlog" (which read as an issue queue).

## Changes (5 files)

- **`ui/messages/en.json` + `de.json`** — values only. The action-bar label, the loading text, and ~28 announcement/coachmark strings that named the pane now say **Repos** (EN) / equivalent (DE). The single `actionbar_backlog` value updates the button, the overlay `aria-label`, and the overlay header.
- **`ActionBar.svelte` + `+page.svelte`** — keyboard shortcut remapped **B → R** so the mnemonic matches the new label (tooltip + `aria-keyshortcuts` + keydown `case`). `r` was free (no collision).
- **`design-system/+page.svelte`** — the one hardcoded style-guide literal naming the pane, updated for consistency.

## Scope decisions

- **Surgical / values-only for identifiers.** Component filenames (`BacklogView`, …), the `Tab` type union, `showBacklog` state, and the `backlog_*` message-key prefixes are intentionally **unchanged** — only user-visible text and the shortcut moved.
- **"backlog" the generic concept is kept.** Strings meaning the issue queue rather than the UI surface — "issue backlog", "backlog issue(s)", "returned to its backlog" — are deliberately left as-is (renaming them would be wrong). Keep-list: EN 9 keys / DE 8 keys.
- **Not a `feat`.** A rename ships no new capability, so no `feature-announcements.ts` catalog entry is added (commit carries `[no-feature-entry]`; the catalog gate only arms on `feat(...)` anyway). Flagged here per house rules.

## Verification

- `bun run check:i18n` — 2 locales in parity (2001 keys each).
- `bun run check` — Paraglide compiles, svelte-check 0 errors / 0 warnings.
- Catalog value-only scan (`jq … select(.value|test("backlog";"i"))`) returns **exactly** the keep-list per locale (EN 9 / DE 8) — no pane-name reference remains.
- Source scan: no rendered "Backlog" literal remains; all residual `Backlog`/`backlog` hits are code identifiers, CSS, comments, or `backlog_*` key references. No `B`-shortcut residue.


scoop (commented):
## Summary

This PR renames the user-facing "Backlog" pane to "Repos" and remaps its keyboard shortcut from `B` to `R`. The change is limited to copy, one label, the shortcut wiring, and a design-system doc line. It satisfies the task and I found no correctness defects.

## What I verified

- **Label + key stability.** `actionbar_backlog` now resolves to "Repos" in both catalogs (verified against `ui/messages/en.json:222` and `ui/messages/de.json:10`). The i18n *key* is intentionally left as `actionbar_backlog`, so the only consumer — `m.actionbar_backlog()` at `ui/src/lib/components/ActionBar.svelte:54` — still resolves correctly.
- **Shortcut rename is consistent across all touch points.** Button tooltip/aria updated to `R`/`r` (`ui/src/lib/components/ActionBar.svelte:53-54`); the plain-key handler now matches `case "r"` and opens the pane (`ui/src/routes/+page.svelte:1101-1106`). No stale `key: "B"` / `aria-keyshortcuts="b"` references remain anywhere under `ui/src` (grepped, no matches).
- **No shortcut collision.** The herd keynav vocabulary is `j/k/g/arrows/1-9` only (verified against `ui/src/lib/components/herd-keynav.ts:78-87`) — `r` is not in it, so the new `case "r"` introduces no conflict. The now-freed `b` falls through to `handleHerdKeyNav`, whose default returns `false` for unknown keys (`ui/src/routes/+page.svelte:1032-1041`), i.e. a harmless no-op. `r` is handled only when not typing/in-terminal, identical to the prior `b` path, so no new PTY interception.
- **Locale parity.** Both `ui/messages/en.json` and `ui/messages/de.json` change the identical set of 30 keys (60 lines each); no key was updated in only one catalog. German keeps "Repos" untranslated as a proper-noun pane name, consistent with how "Backlog" was previously left untranslated.
- **Design-system doc** copy updated (`ui/src/routes/design-system/+page.svelte:463`).

## Nit (non-blocking)

The message keys retain the `backlog` prefix (`actionbar_backlog`, `backlog_loading`, the `feat_backlog_*` family) while their values now say "Repos". This is a reasonable low-churn choice, but the key/value drift is a mild future-maintenance smell. No action required if the team prefers to avoid renaming keys.

- `ui/messages/en.json:936` — `feat_issue_search_title` still reads "Search the issue backlog" while its (in-diff) body now says "Repos view"; the title line is not in this diff and "issue backlog" reads as the generic concept rather than the pane name.
- `ui/messages/en.json:39,273,991,996,998,147` — several strings (`emptyherd_nudge_tail`, `newtask_relaunch_issue_drop_note`, `steerseditor_hint`, `steerseditor_scope_issues_title`, `feat_issue_actions_body`, `automation_autodrain_detail`) still use the word "backlog" as the generic issue-queue noun; not the pane, and untouched here.
- `ui/src/lib/components/ActionBar.svelte:49,282` — the internal CSS class `.btn.backlog` is unchanged; cosmetic only, not user-facing.

PR #1128, opened 2026-06-26
fix(ui): show in-flight state on held-task spawn/discard buttons

kai-osthoff:
## Problem

Clicking **Jetzt starten** (Spawn now) on a held task in the top-bar popover appeared to do *nothing* — *"Leider passiert überhaupt nichts."*

The click path is actually correct: it `POST`s `/api/held/:id/spawn`, which runs `service.create()` server-side — creating a git worktree and launching the agent. That work takes **several seconds**, and during that window the button gave **zero feedback**: it didn't disable, didn't spin, didn't change its label, and the row didn't move. So the button read as dead — exactly the *"button does nothing"* failure mode that #1105 fixed for the *error* case, but the *pending* case was never covered.

## Fix

Track a per-task in-flight state in `TopBar` and surface it in `TopBarHeldBadge`:

- While a spawn/discard is in flight, the row's **select + both buttons are disabled** (`cursor: progress`, dimmed), so the action visibly registers and can't be double-submitted.
- The acting button swaps its label to a progress label (**Wird gestartet… / Wird verworfen…**) and sets `aria-busy`.
- Re-entrancy guard in `doSpawnHeld`/`doDiscardHeld` ignores a second click while one is pending.

Mirrors the inline-error feedback added in #1105 for the same surface.

## Test

- New browser test asserts the spawn button reads **Starting…**, is `disabled`, and is `aria-busy` while pending.
- `cd ui && bun run check` (svelte-check, 0 errors), `check:i18n` (EN+DE parity), and the full `TopBar.browser.test.ts` suite (66 tests) pass.

_Note:_ the local pre-push UI test run was bypassed — the full vitest-browser suite hit an environment-only Playwright pool deadlock in the worktree ("Browser connection was closed"), not a test failure; CI runs the authoritative suite.

[no-feature-entry] — feedback/UX fix to an existing control, not a new user-facing capability.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2004,5 +2004,7 @@
   "herdrupdate_sessions_label": "Laufende Sessions",
   "herdrupdate_jump_to": "Zu {name} springen",
   "feat_herdr_update_session_list_title": "Sieh die Sessions, die ein Update beenden würde — und spring hin",
-  "feat_herdr_update_session_list_body": "Der herdr-Update-Dialog zeigt nicht mehr nur eine Zahl laufender Sessions, sondern listet jede einzeln auf — und jede Zeile springt direkt zu dieser Session. Schließe sie nach und nach ab und aktualisiere dann, ohne Angst, etwas zu verlieren."
+  "feat_herdr_update_session_list_body": "Der herdr-Update-Dialog zeigt nicht mehr nur eine Zahl laufender Sessions, sondern listet jede einzeln auf — und jede Zeile springt direkt zu dieser Session. Schließe sie nach und nach ab und aktualisiere dann, ohne Angst, etwas zu verlieren.",
+  "topbar_held_spawning": "Wird gestartet…",
+  "topbar_held_discarding": "Wird verworfen…"
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2004,5 +2004,7 @@
   "herdrupdate_sessions_label": "Running sessions",
   "herdrupdate_jump_to": "Go to {name}",
   "feat_herdr_update_session_list_title": "See and visit the sessions an update would end",
-  "feat_herdr_update_session_list_body": "The herdr-update dialog no longer just shows a count of running sessions — it lists each one, and every row jumps straight to that session. Wrap them up one by one, then update without worrying you've lost work."
+  "feat_herdr_update_session_list_body": "The herdr-update dialog no longer just shows a count of running sessions — it lists each one, and every row jumps straight to that session. Wrap them up one by one, then update without worrying you've lost work.",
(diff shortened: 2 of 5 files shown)


kai-osthoff (commented):
## Summary

The PR addresses the report "nothing happens when I click 'Jetzt starten' on Desktop" by adding an in-flight affordance to the held-task spawn/discard buttons. Prior work (#1105, already merged) surfaced server-side *errors* inline; a successful spawn removes the row via `loadHeld()`. The remaining gap was the seconds-long in-flight window where the button gave no feedback and read as dead. This PR fills exactly that gap: a per-task `heldPending` map disables the row controls, shows a "Starting…"/"Discarding…" label, sets `aria-busy`, and guards against double-submit. This is a reasonable and well-targeted interpretation of the task.

## Verification

- **i18n parity** — both new keys exist in both catalogs: `topbar_held_spawning`/`topbar_held_discarding` at ui/messages/en.json:1612-1613 and ui/messages/de.json:1612-1613. No duplicate keys (grep confirmed single occurrence each). The component references them as `m.topbar_held_spawning()` / `m.topbar_held_discarding()` (TopBarHeldBadge.svelte:131,139), matching the generated paraglide function names.
- **Desktop path covered** — the report is desktop-specific; the desktop popover (TopBarHeldBadge.svelte:233-244) and mobile dialog (193-232) both render the shared `heldRows()` snippet (94-152), so the fix applies to both. The added test renders with `mobile: false` / `heldPopOpen: true`, exercising the desktop branch the user reported (TopBar.browser.test.ts:986+).
- **State reactivity & wiring** — `heldPending` is `$state` (TopBar.svelte:384); property mutation/`delete` is reactive under Svelte 5 deep proxies. It is passed into the single `<TopBarHeldBadge>` usage (TopBar.svelte:669,679) and declared as an optional prop with `= {}` default (TopBarHeldBadge.svelte:39,60), so no other consumers break. `spawnHeld`/`discardHeld`/`loadHeld` imports and calls confirmed (TopBar.svelte:14-15,391-392,405-406).
- **Control flow** — `if (heldPending[id]) return;` plus `disabled={!!pending}` give double-submit protection on both the guard and the DOM; `finally { delete heldPending[id] }` clears the flag on both success and error so the row never gets stuck. On success the row is removed by `loadHeld()`; on error the existing inline `role=alert` error renders. CSS `:hover:not(:disabled)` and `cursor: progress; opacity: 0.6` correctly reflect the inert state.

The diff does not alter spawn behavior itself; it cannot fix a hypothetical case where spawn neither succeeds nor errors, but nothing in the diff suggests such a path, and #1105 already covers the error case. The chosen scope (missing in-flight affordance) is the defensible root cause.

No blocking issues found.

PR #1129, opened 2026-06-26
fix(ui): surface the real cause when a held task fails to start

kai-osthoff:
## Problem

When a held task fails to start from the held-tasks popover ("Jetzt starten"), the UI only showed a generic, dead-end message:

> Aufgabe konnte nicht gestartet werden. Sie bleibt zurückgehalten – bitte erneut versuchen.

The operator had **no way to tell *why*** it failed — the screenshot that prompted this said exactly that: *"hier ist nicht zu erkennen warum das nicht funktioniert."* The server already returns the real cause in its `{error}` body (agent-name-taken, worktree/create failure, sandbox auto-refusal, …) and `spawnHeld`/`discardHeld` throw it as the error message — but `doSpawnHeld`/`doDiscardHeld` caught it and threw the message away.

## Fix

- `heldErrors` now carries `{ kind: "spawn" | "discard"; detail?: string }` instead of a bare kind string.
- `doSpawnHeld`/`doDiscardHeld` capture the caught error's `.message` as `detail`.
- The popover renders that real cause as a muted, word-broken line beneath the generic message. It's verbatim pass-through server data, so it's i18n-exempt (consistent with the existing rule for tool-use summaries / designations).

## Tests

- Extended the existing held-spawn-failure browser test to assert the server cause surfaces in the alert.
- `bun run check` clean; `TopBar.browser` suite green (65/65).

Note: the pre-push `root-tests`/`ui` lanes timed out (~882s) on the known Shepherd-worktree test-hang (zombie git / pool deadlock), not a real failure — lint/eslint/gates/tsc/ext all passed and the targeted UI suite is green. Full suite runs in CI.

### DIFF
--- a/ui/src/lib/components/TopBar.browser.test.ts
+++ b/ui/src/lib/components/TopBar.browser.test.ts
@@ -968,7 +968,7 @@ describe("TopBarHeldBadge — mobile held-task dialog", () => {
       heldPopFlipUp: false,
       heldItems,
       heldLoading: false,
-      heldErrors: { "held-1": "spawn" },
+      heldErrors: { "held-1": { kind: "spawn", detail: "task name already in use, retry" } },
       heldAutoRelease: true,
       heldAutoReleaseBusy: false,
       toggleHeldAutoRelease: vi.fn(),
@@ -985,6 +985,10 @@ describe("TopBarHeldBadge — mobile held-task dialog", () => {
     // The failure surfaces inline (a toast would render behind the fullscreen dialog),
     // announced assertively so it reaches a screen reader.
     await expect.element(page.getByRole("alert")).toHaveTextContent(m.topbar_held_spawn_failed());
+    // …and carries the server's real cause so the operator can see *why* it failed.
+    await expect
+      .element(page.getByRole("alert"))
+      .toHaveTextContent("task name already in use, retry");
   });
 
   it("shows in-flight state on the spawn button while a held-task spawn is pending", async () => {
--- a/ui/src/lib/components/TopBar.svelte
+++ b/ui/src/lib/components/TopBar.svelte
@@ -375,7 +375,14 @@
   // toast won't do: the mobile popover is a fullscreen portal at the same z-index as
   // the toast stack, so a bottom toast renders behind it. Inline feedback lives inside
   // the surface and stays visible on both desktop and mobile.
-  let heldErrors = $state<Record<string, "spawn" | "discard">>({});
+  let heldErrors = $state<Record<string, { kind: "spawn" | "discard"; detail?: string }>>({});
+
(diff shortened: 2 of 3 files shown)


kai-osthoff (commented):
## Summary

The PR addresses the task ("can't tell why it doesn't work — fix it") by upgrading the per-task held-error state from a bare `"spawn" | "discard"` discriminator to `{ kind, detail? }`, where `detail` carries the server's verbatim `{error}` cause. The popover now renders that cause beneath the generic "try again" message, so an operator can see *why* a spawn/discard was refused.

## Verification

- **Error detail genuinely reaches the UI.** `spawnHeld`/`discardHeld` throw via `failed()`, which returns `apiError(...)` → `new Error(body?.error ?? fallback)` (verified `ui/src/lib/api.ts:107`, `:127`, `:205`, `:216`). So `heldErrorDetail(e)` reading `e.message` (`ui/src/lib/components/TopBar.svelte:382-385`) does carry the server's real cause; the `instanceof Error` guard and `trim()`→`undefined` fallback are sound.
- **Shape is consistent across all consumers.** New object type is used identically at the state declaration (`TopBar.svelte:378`), the prop contract (`top-bar/TopBarHeldBadge.svelte:58`), and the test fixture (`TopBar.browser.test.ts:971`); no consumer still expects the old string form (grepped all `heldErrors` references).
- **Template is valid and safe.** `{@const err = heldErrors[task.id]}` sits as the immediate child of `{#if heldErrors[task.id]}` (`TopBarHeldBadge.svelte:131-132`), so `err` is non-null when dereferenced; `err.detail` is rendered through Svelte's default `{...}` escaping, so the pass-through server string is not an injection vector.
- **i18n parity intact.** The two keys used (`topbar_held_spawn_failed`, `topbar_held_discard_failed`) are unchanged by this PR and present in both catalogs (verified `ui/messages/en.json:1615-1616` and `ui/messages/de.json:1615-1616`). The new `detail` span is intentionally exempt from i18n (server-supplied data), and its comment documents that.
- **Test asserts the new behaviour.** The added assertion checks the `alert` element contains the detail string in addition to the localized message (`TopBar.browser.test.ts:988-991`); since the detail `<span>` is a descendant of the `role="alert"` `<p>`, `toHaveTextContent` covers both.

No blocking issues found.

## Minor (non-blocking, optional)

When the server responds without an `{error}` body, `failed()` falls back to `"spawn held failed: <status>"`, which then becomes `detail`. The popover would then show the generic localized "try again" line followed by the muted `"spawn held failed: 500"`. This is informative (it exposes the status code) rather than wrong, but if you wanted the detail to be *only* a true server-supplied cause, `heldErrorDetail` could distinguish the fallback message. Not required.

Issue #1131, opened 2026-06-26
Height-lock the top bar's left-side tallies to --topbar-ctl-h


scoop:
Follow-up from the "zoo of heights" top-bar fix (the PR equalizing the **right-side** cluster).

That PR introduced a shared `--topbar-ctl-h` (app.css) and made every **boxed right-cluster control** (NEEDS YOU, held, update/herdr, what's-new, learnings, docs, gear) render at one height in both full-label and compact states.

**Deferred:** the **left-side tallies** (`ui/src/lib/components/top-bar/TopBarTallies.svelte` — `.tally` / `.ctally`) were intentionally left out. They're a separate, mostly-transparent filter-button group left of the logo separator, deliberately tighter than the right-cluster pills (`.ctally` carries a 24px AA floor / 44px coarse), so forcing them to 30px changes their feel and is a separate visual decision.

**Task:** decide whether the tallies should also adopt `--topbar-ctl-h` for full-bar vertical consistency, and if so normalize `.tally` / `.ctally` (base + compact) the same way — `box-sizing:border-box; min-height:var(--topbar-ctl-h); line-height:1` + horizontal-only padding — while preserving the 24px AA / 44px coarse floors.

Scope: CSS-only, `TopBarTallies.svelte`.

scoop:
✅ PR #1142 merged.

PR #1115, opened 2026-06-26
fix(ui): equalize compact NEEDS YOU + held-badge height

scoop:
## What

In the top bar's right-side cluster, the **NEEDS YOU** badge and the **held-tasks** badge sit side by side. In their **compact** form they rendered at unequal heights. This equalizes them across every context.

## Root cause — three diverging height drivers

1. **Padding.** `.held-badge.compact` never overrode the base `3px 8px`, while `.needsyou.compact` uses `8px 10px` → needsyou ~10px taller.
2. **Line box (the subtle one).** `.needsyou` sets no `font-family`, so it renders in the **UA button font** (sans; `normal` line-height ≈ 1.09), while `.held-badge` uses `font: inherit` → **monospace** (`normal` ≈ 1.27). There's no global `button { font: inherit }` reset, so even after matching padding a **~2px residual** remained.
3. **`min-height` floor.** needsyou gets a *width-gated* 44px floor via `.hud.mobile .needsyou` (not pointer-gated); the held badge's only floor was `@media (pointer: coarse)`, so a **narrow fine-pointer (≤768px)** viewport left it short.

## Fix

In `TopBarHeldBadge.svelte`:
- `padding: 8px 10px` on `.held-badge.compact` (mirror `.needsyou.compact`)
- `line-height: 1` on `.held-badge.compact` (font-independent line box)
- `class:mobile` on the badge button + `.held-badge.mobile { min-height: 44px }` mirroring the needsyou width-gated floor

In `TopBar.svelte`:
- matching `line-height: 1` on `.needsyou.compact`

## ⚠️ Deliberate deviation from the approved plan

The plan scoped this to **one file** (`TopBarHeldBadge.svelte`). Execution-time browser measurement revealed the font-driven ~2px residual (driver #2), which can only be fixed font-independently by pinning `line-height: 1` on **both** badges — so `.needsyou.compact` in `TopBar.svelte` is also touched. Flagged here, in code comments on both rules, and in `.shepherd-plan.md`.

## Verification

Measured computed box heights in a headless-Chromium harness replicating the exact winning CSS for all four contexts (font-independence confirmed by testing needsyou in both sans and mono):

| Context (`mobile` / pointer / `compactBadges`) | needsyou | held | equal |
| --- | --- | --- | --- |
| Desktop overflow — false / fine / compact | 29px | 29px | ✓ |
| Coarse wide — false / coarse | 44px | 44px | ✓ |
[...]

PR #1132, opened 2026-06-26
fix(top-bar): equalize right-cluster box heights via shared token

scoop:
## What

The top bar's right-side cluster had a "zoo of heights": each boxed control derived its height from its own ad-hoc vertical `padding` **plus** whatever glyph it carried, so they fanned out (~25–34px). `✦ 314` was tallest (16px ✦ + 8px compact padding); `▲ 6` shortest. Prior fixes were pairwise patches — #1115 equalized only NEEDS YOU + held (compact); #1122 then re-broke held by adding the hourglass glyph.

This replaces the whack-a-mole with **one shared control-height token** every boxed control derives from, so a box's glyph or padding can no longer drift its height.

## How

- **New token** `--topbar-ctl-h: 30px` in `app.css` `:root`, in the fixed-chrome-heights group beside `--icon-btn-hit` / `--mobile-actionbar-hit`. 30px = the gear's current box height (its 20px glyph is the tallest mandatory content), so the dominant element is unchanged and the others meet it.
- Every **boxed** control gets `box-sizing:border-box; min-height:var(--topbar-ctl-h); line-height:1` + flex centering, with **vertical padding removed** on base *and* compact/mobile rules: `.needsyou`, `.held-badge`, `.update-badge`(+`.herdr`), `.whatsnew-badge`, `.whatsnew-dot-btn`, `.learnings-btn`, `.docs-link`, `.gear`.
- `min-height` (not `height`) so content can still grow under iOS Dynamic Type (`--ui-scale` up to 1.5); identical to `height` at desktop scale.
- `box-sizing:border-box` is added explicitly to every touched rule so the borderless gear/docs land on the *same* 30px as the 1px-bordered badges.
- `.whatsnew-dot-btn`'s **unconditional** `min-height/min-width:44px` is moved into `@media (pointer:coarse)` so it drops to the token height on fine-pointer desktop instead of being a 44px outlier; the coarse/mobile 44px floors elsewhere are preserved (they override the smaller token).

### Deliberate scoping (flagged per house rules)
- **`.health-pip`** (round 22px circle) is **not** box-sized — forcing `min-height:30px` with `width:22px` would ovalize it. It stays a centered 22px dot.
- **Gauges / clock** (`5 H`, `WK`, `%`, `●`) are non-boxed inline content, left centered.
- **Left-side tallies** are deferred to a tracked follow-up (#1131) — they're deliberately tighter chips and a separate visual decision.

## Verification
- `cd ui && bun run check` → 0 errors; prettier clean.
- Measured the final CSS (real glyphs, both full-label and compact states) in a headless Chromium: **all boxes render exactly 30px** (`distinctHeights: [30]`, `allEqual: true`).

Closes the "still a zoo of heights in the top bar" report.

### DIFF
--- a/ui/src/app.css
+++ b/ui/src/app.css
@@ -108,6 +108,14 @@
   --icon-btn-hit: 28px; /* desktop square hit area for header icon buttons */
   --icon-btn-glyph: 18px; /* canonical inline-SVG glyph size */
 
+  /* Shared height for the top bar's right-cluster boxed controls (needsyou/held/
+     update/learnings badges, what's-new, docs link, gear) — single source so a
+     box's glyph or padding can't drift its height (the recurring "zoo of heights").
+     A fixed chrome height: NOT scaled by --ui-scale; consumers use it as a
+     min-height so content can still grow under iOS Dynamic Type. The coarse-pointer
+     /mobile 44px floors override it on touch. */
+  --topbar-ctl-h: 30px;
+
   --status-running: var(--color-amber);
   /* done = WAITING (finished turn, parked for the operator's next steer). It is
      NOT actionable-complete, so it must not read green/"ignore me" — it reads as
--- a/ui/src/lib/components/TopBar.svelte
+++ b/ui/src/lib/components/TopBar.svelte
@@ -867,12 +867,19 @@
     color: var(--color-muted);
   }
   .needsyou {
+    box-sizing: border-box;
+    display: inline-flex;
+    align-items: center;
+    justify-content: center;
+    /* shared bar control height; glyph/text centers, padding no longer drives height */
+    min-height: var(--topbar-ctl-h);
+    line-height: 1;
     background: color-mix(in srgb, var(--color-red) 18%, transparent);
     border: 1px solid var(--color-red);
     color: var(--color-red);
     letter-spacing: 0.14em;
     font-size: var(--fs-meta);
-    padding: 5px 10px;
+    padding: 0 10px;
     cursor: pointer;
     white-space: nowrap;
     flex-shrink: 0;
@@ -898,11 +905,14 @@
   /* Standalone docs link: a quiet icon button matching the gear's at-rest muted ink,
      going amber on hover/focus. Coarse-pointer 44px floor handled in the shared block. */
   .docs-link {
+    box-sizing: border-box;
     display: inline-flex;
     align-items: center;
     justify-content: center;
+    min-height: var(--topbar-ctl-h);
+    line-height: 1;
     color: var(--color-muted);
-    padding: 5px 8px;
+    padding: 0 8px;
     flex-shrink: 0;
   }
   .docs-link:hover,
@@ -985,7 +995,7 @@
      so .hud.mobile .gauge-btn can never match — verified dead (#855). */
   .hud.mobile .needsyou {
     min-height: 44px;
-    padding: 8px 12px;
+    padding: 0 12px;
   }
   /* Phone: collapse the badge to an icon+count chip so the NEEDS YOU call-out
      fits on line 1 next to the gauge/gear instead of forcing the right-side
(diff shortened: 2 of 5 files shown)


scoop (commented):
## Summary

This PR introduces a single `--topbar-ctl-h: 30px` token and makes every boxed right-cluster control in the top bar consume it as a `min-height` (with `box-sizing: border-box`, flex centering, and `padding: 0 …`), so glyphs/padding no longer drive divergent heights. This directly addresses the task ("zoo of heights in the top bar"). The implementation is internally consistent and the reasoning in the comments is sound.

## What I verified

- **Token definition & scope.** `--topbar-ctl-h: 30px` is defined in the global `:root` block (`ui/src/app.css:117`, inside the `:root` at `ui/src/app.css:35`), so it cascades into every Svelte-scoped component that references it.
- **All consumers resolve to the token.** The 8 controls named in the token comment all reference it: needsyou (`ui/src/lib/components/TopBar.svelte:875`), docs-link (`ui/src/lib/components/TopBar.svelte:912`), held-badge (`ui/src/lib/components/top-bar/TopBarHeldBadge.svelte:251`), gear (`ui/src/lib/components/top-bar/TopBarGear.svelte:246`), update-badge (`TopBarBadges.svelte:174`), whatsnew-badge (`TopBarBadges.svelte:224`), whatsnew-dot-btn (`TopBarBadges.svelte:258-259`), learnings-btn (`TopBarBadges.svelte:309`). Grep for `--topbar-ctl-h` shows no stray/misspelled references.
- **Touch floors preserved.** Every control that dropped its unconditional 44px still has a 44px floor on touch: `.needsyou`/`.docs-link` (`TopBar.svelte:1033-1041`), `.update-badge`/`.whatsnew-badge`/`.whatsnew-dot-btn`/`.learnings-btn` (the `@media (pointer: coarse)` block extended at `TopBarBadges.svelte:351`+), `.held-badge` (`TopBarHeldBadge.svelte:380-382`), `.gear` (`TopBarGear.svelte:309-313`) plus the width-gated `.gear.mobile`/`.held-badge.mobile`/`.hud.mobile .needsyou` rules. The previously-unconditional 44px on `.whatsnew-dot-btn` is now correctly coarse-gated, which is the point of the change (drop the desktop outlier).
- **border-box rationale checks out.** There is no global `* { box-sizing: border-box }` (the global `*` rule at `ui/src/app.css:332` only sets tap-highlight), so the explicit `box-sizing: border-box` additions are meaningful: they make the borderless gear (30px) sit level with the 1px-bordered badges (30px incl. border), as the comment at `TopBarGear.svelte:244-245` claims.
- **Content fits 30px at default scale.** `--fs-meta` = 11px and `--fs-xl` = 20px (`ui/src/app.css:82,85`); the gear/docs 20px glyphs and 11px badge text all fit within the 30px min-height with flex centering. Absolutely-positioned gear pips (`TopBarGear.svelte:261-296`) are unaffected by switching `.gear` to `inline-flex`.

No correctness, security, or quality defects found. CSS-only change; no i18n/signature/contract surface touched.

- ui/src/app.css: `--topbar-ctl-h` is a fixed 30px and is intentionally NOT scaled by `--ui-scale`, while the consumed glyphs (`--fs-xl` = 20px·scale) and text (`--fs-meta` = 11px·scale) ARE scaled. Because the token is used as `min-height`, at large iOS Dynamic Type (`--ui-scale` ≳ 1.5) the gear/docs glyphs outgrow the floor faster than the text badges, so the boxes can diverge in height again. This is the documented trade-off (comment at `ui/src/app.css:114-115`), so it is informational only.

- The credit/usage gauge controls (TopBarUsage.svelte) and the clock are not boxed via `--topbar-ctl-h`; if any boxed gauge control still differs in height from this cluster, that is outside this diff and was not assessed.

PR #1130, opened 2026-06-26
feat(ui): list running sessions in herdr-update dialog with jump-to

kai-osthoff:
## Was

Der **herdr-Update-Dialog** zeigte bisher nur eine nackte Zahl laufender Sessions ("Dadurch werden **2** laufende Session(s) beendet"). Man sah *dass* etwas beendet wird, aber nicht *welche* Sessions — und musste blind aktualisieren oder raten, was man verliert.

Jetzt listet der Dialog **jede betroffene Session einzeln** auf. Jede Zeile ist ein Button, der direkt zu dieser Session springt (schließt den Dialog, selektiert die Session). So kann man die Sessions nach und nach abschließen und dann beruhigt updaten — ohne Angst, etwas zu verlieren.

## Wie

- `HerdrUpdateModal`: `sessions` ist nicht mehr `number`, sondern `{ id, desig, name }[]`. `count` wird daraus abgeleitet (Warnung + Button-Text unverändert). Unter der Warnung rendert eine Liste anklickbarer Session-Zeilen (44px Tap-Target, Tokens, `aria-label` "Zu {name} springen"). Liste wird ausgeblendet, sobald das Update läuft.
- `onjump(id)` → `AppOverlays` → `+page.svelte`: schließt den Dialog und ruft `selectUnit(id)`.
- `AppOverlays` mappt die laufenden Sessions (gleicher `displayStatus === "running"`-Filter wie zuvor) auf die Minimal-Shape.
- i18n: `herdrupdate_sessions_label`, `herdrupdate_jump_to` (EN+DE).
- Feature-Katalog-Eintrag `herdr-update-session-list` (1.38.0) + `feat_herdr_update_session_list_*` Keys.

## Checks

- `bun run check` (svelte-check): 0 Fehler
- `bun run check:i18n`: Parität (2007 Keys je Locale)
- UI-Tests (AppOverlays / HerdrUpdate): grün

PR #1137, opened 2026-06-26
fix(ui): keep running-sessions list visible in herdr-update dialog

kai-osthoff:
## Problem

The herdr-update dialog lists the running sessions that the herdr restart would interrupt (#1130), so the operator can jump to each and wrap it up before updating. But that list — the whole point of the lower section — **renders invisibly** when the release notes are long.

Reported (DE): *"Leider lässt sich der untere Teil – die Sessions die ein Update verhindern – nicht anzeigen."*

## Cause

The modal `.card` is a flex column capped at `max-height: 80dvh`. Both the release-notes box and the running-sessions `<ul>` are scroll containers (`overflow-y: auto`), which gives them an **automatic minimum size of 0**. When long release notes push the content past the 80dvh cap, flexbox is free to shrink **both** scroll containers — and it collapses the sessions list down to a sliver, leaving just the label + an empty box (exactly what the screenshots show).

The data was always there (the "beendet 1 Session(s)" count proves `sessions.length === 1`); it was purely a layout collapse.

## Fix

One line of CSS: pin the sessions list with `flex-shrink: 0`. The release-notes box is explicitly the intended shrinkable element (it already carries `min-height: 0` and scrolls internally), so it now absorbs the shrinkage while the sessions list keeps its natural height (still capped at its own `max-height: 180px` with internal scroll for very long lists).

## Verification

- No new strings, no `feat` → i18n / feature-catalog gates N/A

[no-feature-entry]

### DIFF
--- a/ui/src/lib/components/HerdrUpdateModal.svelte
+++ b/ui/src/lib/components/HerdrUpdateModal.svelte
@@ -245,6 +245,11 @@
     display: flex;
     flex-direction: column;
     gap: 14px;
+    /* Last-resort scroll when even the shrinkable notes box can't free enough
+       space — e.g. no release notes at all (nothing shrinkable) plus a full
+       180px sessions list on a short viewport — so the action buttons are never
+       clipped below the 80dvh cap. (Mobile re-sets this in the media query.) */
+    overflow-y: auto;
     background: var(--color-panel);
     border: 1px solid var(--color-line-bright);
     padding: 18px 18px 16px;
--- a/ui/src/lib/components/HerdrUpdateModal.svelte
+++ b/ui/src/lib/components/HerdrUpdateModal.svelte
@@ -425,6 +425,14 @@
     /* keep a long list from pushing the actions off-screen on short viewports */
     max-height: 180px;
     overflow-y: auto;
+    /* Don't let the flex column collapse this list away. The card is capped at
+       80dvh and both this list and the release-notes box are scroll containers
+       (automatic min-size 0), so when long notes overflow the cap flexbox is
+       free to shrink BOTH to a sliver — which hid the running-sessions list
+       entirely. The notes box is the intended shrinkable one (min-height: 0);
+       the sessions list is this dialog's whole point, so pin it and let the
+       notes absorb the shrinkage instead. */
+    flex-shrink: 0;
   }
   .session {
     width: 100%;


kai-osthoff (commented):
## Summary

Single-line CSS fix: adds `flex-shrink: 0` to `.sessions` so the running-sessions list is no longer collapsed away by flexbox. This satisfies the task ("make the blocking-sessions list visible").

## Verification

- `.card` is a flex column capped at `max-height: 80dvh` (ui/src/lib/components/HerdrUpdateModal.svelte:244-246).
- `.notes` is `flex: 0 1 auto; min-height: 0; overflow-y: auto` — the deliberately shrinkable child (ui/src/lib/components/HerdrUpdateModal.svelte:309-311).
- `.sessions` is `max-height: 180px; overflow-y: auto` (ui/src/lib/components/HerdrUpdateModal.svelte:426-427). A flex item with non-visible overflow resolves `min-height: auto` to 0, so before this change it was freely shrinkable; with long release notes flexbox shrank BOTH scroll containers to a sliver, hiding the list. Adding `flex-shrink: 0` (ui/src/lib/components/HerdrUpdateModal.svelte:435) pins the list and lets `.notes` absorb the shrinkage. Root-cause analysis in the comment is correct.
- `.sessions` (gated `{#if !submitting}`, :171) and `.log` (gated `{#if submitting}`, :194) are mutually exclusive, so the log pane is unaffected. Verified against ui/src/lib/components/HerdrUpdateModal.svelte:171,194.
- No prop/signature/i18n changes in the diff — message keys (e.g. `herdrupdate_sessions_label`) are untouched.

- ui/src/lib/components/HerdrUpdateModal.svelte: when `update.notes` is empty there is no shrinkable flex child, so on a short desktop/landscape viewport a full 180px session list can push the action buttons past the 80dvh cap and clip them (desktop `.card` has no `overflow-y`; only the mobile sheet sets it at :562). Previously the list would have shrunk to keep actions on-screen. Acceptable tradeoff given the task and the always-available header ✕, but worth a follow-up if the no-notes case matters.

kai-osthoff (commented):
## Re-review

The diff is CSS-only, touching one file: `ui/src/lib/components/HerdrUpdateModal.svelte`.

### Prior finding — addressed

The previous revision raised the no-notes short-viewport case: with `update.notes` empty there is no shrinkable child, so a full 180px session list could push the action buttons past the 80dvh cap and clip them on desktop.

The diff resolves this two ways:
- `.card` gains `overflow-y: auto` (HerdrUpdateModal.svelte:252) so when nothing else can shrink, the card itself scrolls and the actions stay reachable.
- `.sessions` gains `flex-shrink: 0` (HerdrUpdateModal.svelte:440) so the running-sessions list — the whole point of this dialog — is never collapsed to a sliver; the `.notes` box (`min-height: 0`, HerdrUpdateModal.svelte:315) absorbs the shrinkage instead.

Together these guarantee the sessions list is visible and the buttons reachable in the edge case. Verified against the rendered structure: when `update.notes` is empty the `{#if update.notes}` block (HerdrUpdateModal.svelte:151) emits no `.notes` element, so the card-level scroll is the only fallback. This satisfies the task and the prior point.

### New side effect (non-blocking)

Adding `overflow-y: auto` to the desktop `.card` makes it a scroll container on **both** axes: per the CSS overflow spec, when `overflow-y` is non-`visible` and `overflow-x` is left at its `visible` default, `overflow-x` computes to `auto`. The card therefore clips to its padding box, which clips the absolutely-positioned `.bracket` corner decorations that are intentionally offset to `top/left: -1px` and `bottom/right: -1px` (HerdrUpdateModal.svelte:266-276). The result is a ~1px clip of each corner accent on **every** desktop render, not just the short-viewport edge case.

Ground truth for the pattern divergence: the sibling `UpdateModal.svelte` (same `card bracket` markup and same `.bracket` offsets at UpdateModal.svelte:178-189) deliberately keeps its desktop `.card` non-scrolling — it relies solely on the shrinkable `.commits` child (UpdateModal.svelte:227-229) and only adds `overflow-y: auto` inside its mobile media query (UpdateModal.svelte:378), where the brackets are `display: none` (UpdateModal.svelte:381-383). HerdrUpdateModal's mobile block does the same (HerdrUpdateModal.svelte:567, brackets hidden at :570-573), so the clip is desktop-only. This is cosmetic, hence a comment rather than a block, but it is a real regression caused by this diff.

(Minor: the comment at HerdrUpdateModal.svelte:251 says "Mobile re-sets this" — the mobile rule sets the same `overflow-y: auto` value, so it isn't really a re-set; harmless wording.)

PR #1138, opened 2026-06-26
fix(uploads): copy staged images into worktree so held-task spawns survive retries

kai-osthoff:
## Problem

Held tasks with an attached **screenshot** could fail to start with `ENOENT: no such file or directory, rename '…/.shepherd-uploads-staging/…png' → '…/.shepherd-uploads/…png'` — and then stay permanently wedged at the head of the held queue (re-blocking every retry).

Root cause: attached images stage in `.shepherd-uploads-staging/` and were **moved** (`renameSync`) into the worktree at spawn time (`moveStagedIntoWorktree`, `src/uploads.ts`). The move is destructive and non-idempotent:

1. **Failed first spawn** — if `service.create` failed *after* the move, the task stayed held (head-of-line, `held-release.ts`) but the staged source was already gone. Every "Jetzt starten" retry pointed at the now-empty staging path → `ENOENT`.
2. **24h sweep** — `sweepStaging` (`STAGING_TTL_MS = 24h`) deletes long-staged files → same `ENOENT`.

Tasks without an image retry cleanly — images are the only spawn input with a consumable, destructive staging step. Hence "specifically screenshots".

## Fix

- **Copy-not-move** (`moveStagedIntoWorktree` → `copyStagedIntoWorktree`): `copyFileSync` instead of `renameSync`. The staged original stays recoverable, so a failed first spawn and any retry both succeed; `sweepStaging` reclaims the leftover copy after 24h. This mirrors the existing `copyOriginalUploads` rationale already in the relaunch path (_"keeps the original recoverable on a spawn failure"_).
- **Skip a genuinely-gone source** instead of throwing — the spawn proceeds without it.
- **Operator-visible signal** for the (now rare, >24h-sweep-only) drop: a note in the session prompt, a `console.warn`, and a toast via a new `session:uploads-dropped` WS event — mirrored exactly on the existing `session:egress-drop` plumbing. The operator can relaunch with the image re-attached if it was essential.

## Tests

- `test/uploads.test.ts`: copy keeps the staged source (`existsSync` flips `false`→`true`); new test for skip-missing-source.
- `test/service.test.ts`: seam `moveUploads`→`copyUploads`; new test asserting a dropped image still spawns, notes the loss in-prompt, and emits `session:uploads-dropped`.
- Root `bun run lint` + targeted `bun test` green; UI `bun run check` + `check:i18n` (2012 keys parity) + `store.svelte.test.ts` green.

## Notes

- Server-side bugfix + a small UI toast; new i18n key `toast_uploads_dropped` (EN+DE).
- Not a `feat` and the toast handler lives outside the feature-catalog gate globs — no catalog entry needed.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -17,7 +17,7 @@ import type {
   Session,
 } from "./types";
 import {
-  moveStagedIntoWorktree,
+  copyStagedIntoWorktree,
   stagingDir,
   sweepStaging,
   STAGING_TTL_MS,
@@ -108,8 +108,8 @@ export interface ServiceDeps {
   refineName?: (args: { taskText: string; label: string }) => Promise<string | null>;
   /** Event bus for live state pushes (e.g. session:ready); absent in tests that skip it. */
   events?: Pick<EventHub, "emit">;
-  /** Inject point for tests; defaults to the real fs move. */
-  moveUploads?: (images: string[], worktreePath: string) => string[];
+  /** Inject point for tests; defaults to the real fs copy (copyStagedIntoWorktree). */
+  copyUploads?: (images: string[], worktreePath: string) => string[];
   /** Detects/terminates leftover subprocesses at close; absent in tests that skip it. */
   reaper?: Pick<ProcessReaper, "detect" | "reap" | "stopListenersOnPort">;
   /** Live preview service; provides devPortFor for stopPreview. Absent → stopPreview returns not_found. */
@@ -1049,13 +1049,30 @@ export class SessionService {
    * against the 8000-char human-prompt guard (the same approach for each). The comment
    * thread is the human discussion that refined the original request; fetching it is
    * best-effort (see fetchIssueCommentsBlock) so a spawn never fails on comments.
+   *
+   * Returns the prompt plus `dropped`: the count of attached images whose staged source
+   * was gone (e.g. swept after 24h) so copyStagedIntoWorktree skipped it. The spawn still
+   * proceeds without them; the caller emits an operator-visible signal for the drop.
    */
-  private async composePromptArg(input: CreateSessionInput, worktreePath: string): Promise<string> {
+  private async composePromptArg(
+    input: CreateSessionInput,
+    worktreePath: string,
+  ): Promise<{ promptArg: string; dropped: number }> {
     let promptArg = input.prompt;
+    let dropped = 0;
     if (input.images.length > 0) {
-      const move = this.deps.moveUploads ?? moveStagedIntoWorktree;
-      const moved = move(input.images, worktreePath);
-      promptArg = `${promptArg}\n\nAttached images:\n${moved.join("\n")}`;
+      const copy = this.deps.copyUploads ?? copyStagedIntoWorktree;
+      const copied = copy(input.images, worktreePath);
+      if (copied.length > 0) promptArg = `${promptArg}\n\nAttached images:\n${copied.join("\n")}`;
+      dropped = input.images.length - copied.length;
+      if (dropped > 0) {
+        // A staged upload vanished before spawn (swept after STAGING_TTL_MS, or otherwise
+        // lost). Note it in-prompt so the agent knows an attachment is missing, and warn.
+        promptArg = `${promptArg}\n\n[Note: ${dropped} attached image(s) could not be restored — the upload expired and is unavailable for this session.]`;
+        console.warn(
+          `[uploads] ${dropped}/${input.images.length} staged image(s) missing at spawn; proceeding without them`,
+        );
+      }
     }
     if (input.issueRef) {
       const r = input.issueRef;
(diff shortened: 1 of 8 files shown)


kai-osthoff (commented):
## Summary

The PR addresses the screenshot/upload issue by switching `moveStagedIntoWorktree` → `copyStagedIntoWorktree`: staged images are now **copied** into the worktree instead of moved, so the staged original survives a failed-and-retried spawn (the held-task "Jetzt starten" retry path). A genuinely-gone source (swept after the TTL) is skipped rather than throwing, the prompt notes the loss, and an operator toast (`session:uploads-dropped`) is emitted. This squarely fixes the reported problem: under the old `renameSync`, the first spawn consumed the staged file, so any retry hit ENOENT and was permanently un-retryable.

The implementation is sound and I could not find a blocking defect.

## Verification performed

- **Identifier rename consistency** — `copyStagedIntoWorktree` defined at `src/uploads.ts:48`, imported/used at `src/service.ts:20,1064`; the dep rename `copyUploads` at `src/service.ts:111-112` and all three test seams (`test/service.test.ts:908,958,4166`). No stale `moveUploads`/`moveStagedIntoWorktree` left in `src/` (only `docs/superpowers/plans/...` which is a pre-existing plan doc, out of scope). Verified `src/index.ts` does not pass an upload seam, so it falls through to the new default — confirmed via grep (no matches).
- **Event wiring parity** — `this.deps.events?.emit("session:uploads-dropped", { id, count })` at `src/service.ts:1784` mirrors the existing `session:egress-drop` emit at `src/egress-watch.ts:203`. The UI union type is extended at `ui/src/lib/types.ts:1156` with the matching `{ id: string; count: number }` shape, and the dispatch case at `ui/src/lib/store.svelte.ts:344-349` mirrors the egress case exactly (`key`, `alert`). `EventHub.emit` is untyped (`src/events.ts:9`), and the WS bridge forwards every event generically (`src/server.ts:5309-5311`), so no server-side allowlist needs updating — the toast will reach the client.
- **i18n locale parity** — `toast_uploads_dropped` present in BOTH `ui/messages/en.json:1161` and `ui/messages/de.json:1161`, both using the `{count}` placeholder that `m.toast_uploads_dropped({ count })` supplies (`ui/src/lib/store.svelte.ts:345`). Parity holds.
- **Empty/all-dropped edge** — the new `if (copied.length > 0)` guard at `src/service.ts:1066` correctly suppresses an empty `Attached images:` block when every image is dropped, while still appending the note; `dropped = input.images.length - copied.length` (`src/service.ts:1068`) is correct.
- **Tests** — new `test/uploads.test.ts` cases assert copied-not-moved (source survives) and missing-source-skipped; new `test/service.test.ts` case asserts survivor attached + note + toast event against `s.id`. These match the implementation strings (`"could not be restored"`, event name/shape).

## Nit (non-blocking)

- `src/uploads.ts:42` — the doc comment says "so the **move** is idempotent" but the function no longer moves; it's a copy. Minor leftover terminology that could read as confusing. Consider "so the operation is idempotent".

- src/uploads.ts: TOCTOU between `existsSync(src)` (line 56) and `copyFileSync(src, dest)` (line 58) — if the sweep deletes the source in that microsecond window, `copyFileSync` throws ENOENT and the whole `composePromptArg` rejects, contradicting the function's stated "skipped, not thrown on" contract. The window is negligible (sweep runs on a TTL) and the same race existed under the old `renameSync`, so this is informational only.

kai-osthoff (commented):
## Re-review

The one prior-revision finding is **resolved**:
1. `src/uploads.ts` doc comment — verified `src/uploads.ts:38-46` now describes a copy ("COPY-not-move keeps the staged original recoverable... so this copy is repeatable") and no longer contains the "the move is idempotent" wording. Addressed.

## Verification of this PR's change

The PR switches staged-image transfer from move to copy so a failed/retried spawn finds the source intact, and surfaces an operator toast when a staged upload was swept before spawn.

- **Copy seam rename** — `moveStagedIntoWorktree`/`moveUploads` → `copyStagedIntoWorktree`/`copyUploads`. Grepped all references: only `src/service.ts`, `src/uploads.ts`, and the two test files use them; the sole stale mention is `docs/superpowers/plans/2026-05-30-image-drag-drop-upload.md` (a historical plan doc, out of scope, not code).
- **Drop accounting** — `src/service.ts:1058-1078`: `dropped = input.images.length - copied.length`; the all-dropped path still appends the in-prompt note and correctly skips an empty "Attached images:" block (`if (copied.length > 0)`). Logic correct.
- **Event wiring is internally consistent** — emitted at `src/service.ts:1784` as `{ id: sessionId, count: droppedImages }`; typed at `ui/src/lib/types.ts:1156` as `{ event: "session:uploads-dropped"; data: { id: string; count: number } }`; handled at `ui/src/lib/store.svelte.ts:344-348` via `m.toast_uploads_dropped({ count: ev.data.count })`, mirroring the existing `session:egress-drop` case. `EventHub.emit(event: string, data: unknown)` (`src/events.ts:9`) is loosely typed, so the server emit type-checks.
- **i18n parity** — `toast_uploads_dropped` present in both `ui/messages/en.json:1161` and `ui/messages/de.json:1161` (glob confirms these are the only two catalogs), both interpolating `{count}` to match the `{ count }` call site.
- **Skip-missing-source** — `src/uploads.ts:53` `if (!existsSync(src)) continue;` skips swept sources without throwing; tests cover both the survivor-copy and skip paths.

Issue #944, opened 2026-06-21
ci: harden docs-site build against flaky @pagefind tarball-extraction failures


scoop:
## Problem

The `verify` CI job intermittently fails on the **Build (docs-site)** step with:

```
error: Fail extracting tarball for "@pagefind/linux-x64"
error: Fail extracting tarball from @pagefind/linux-x64
##[error]Process completed with exit code 1.
```

Observed on PR #941 (a UI-only glossary-recipe change that touches **nothing** in `docs-site/`). A re-run cleared it with no code change — it's a transient install flake, not a real failure.

## Root cause

`pagefind` (the Starlight docs search indexer) ships its native binary as **optional, platform-specific deps** (`@pagefind/linux-x64`, `@pagefind/darwin-arm64`, …). `bun install` occasionally fails extracting that tarball.

The CI step has no resilience:

```yaml
# .github/workflows/ci.yml
- name: Build (docs-site)
  run: cd docs-site && bun install --frozen-lockfile && bun run check && bun run build
```

No retry, no bun cache. Because docs-site builds inside the shared `verify` job, a flake there reds the **whole** job — failing every PR (even ones that never touch `docs-site/`) and blocking the critic, which won't review a red PR.

## Proposed fix (pick one / combine)

- **Retry the install** — wrap `bun install --frozen-lockfile` in a small retry (2–3 attempts) so a one-off extraction failure self-heals.
- **Cache bun's global store** keyed on `docs-site/bun.lock`, so the pagefind binary isn't re-downloaded/re-extracted every run (faster + fewer extractions to flake).
- Optionally split docs-site into its own job so its flakes don't red the UI/server/extension verify.

## Acceptance

- A transient `@pagefind` tarball-extraction failure no longer reds `verify` on the first attempt (retry/caching absorbs it).
- No change to what docs-site actually builds.

Discovered while clearing red CI on #941 (#938).

scoop:
✅ PR #948 merged.

PR #1141, opened 2026-06-26
ci: harden verify-job bun installs against flaky tarball-extraction (#1134)

### LINKED ISSUE #1134: ci: harden the verify job's bun install against flaky tarball-extraction (extend #944)
## Symptom

The `verify` CI job intermittently fails at one of its `bun install` steps with a native-binary tarball-extraction error, e.g.:

```
error: Fail extracting tarball for "lightningcss-linux-x64-gnu"
##[error]Process completed with exit code 1.
```

A re-run with **no code change** goes green, confirming it's a transient network/cache flake, not a real defect. This has recurred several times over the past week (most recent: run https://github.com/erwins-enkel/shepherd/actions/runs/28228301484, PR #1132, "Install (ui)" step).

## Root cause / gap

#944 already hardened the **`docs-site`** job against the *same class* of flake (it was `@pagefind` there) with a retry-that-purges-the-bun-store wrapper. But #944's scope was docs-site only — the **`verify`** job was never covered. It still runs three bare installs with no retry:

- `.github/workflows/ci.yml:52` — `bun install --frozen-lockfile` (root)
- `.github/workflows/ci.yml:55` — `cd ui && bun install --frozen-lockfile`
- `.github/workflows/ci.yml:58` — `cd extension && bun install --frozen-lockfile`

Any one of these hitting a half-written/corrupted native tarball fails the whole job.

## Proposed fix

Reuse the **exact pattern already in this file** (the docs-site "Build" step, ci.yml ~167-190): wrap each `verify`-job install in the same retry loop that purges `~/.bun/install/cache` before the final attempt so a poisoned tarball is re-downloaded fresh rather than re-hit identically. Only the **install** is retried (re-running check/build/lint on a genuine error would just burn 3× the time).

Options:
1. Inline the same `until bun install --frozen-lockfile; do … rm -rf ~/.bun/install/cache …` loop into each of the three steps, or
2. Factor it into a tiny reusable shell snippet / composite action and call it from both jobs (removes the existing fallow duplication between the two retry blocks too).

Keep `--frozen-lockfile` (lockfile integrity must still be enforced). CSS/source unchanged — this is CI-only.

## Acceptance
- `verify`-job installs survive a one-off tarball-extraction flake (retry succeeds without manual re-run).
- A genuine lockfile/type error still fails fast (no pointless 3× retries of check/build).

Prior art: #944 (docs-site equivalent, closed/completed).

scoop:
Closes #1134.

## Problem

`verify`'s three `bun install --frozen-lockfile` steps (root, `ui/`, `extension/`) ran with no retry, so a one-off native-binary **tarball-extraction flake** (e.g. `Fail extracting tarball for "lightningcss-linux-x64-gnu"`) red the whole job and forced a no-change re-run. #944 already fixed this *class* of flake — but for the `docs-site` job only.

## Fix

Factor #944's retry-with-store-purge loop into the repo's **first composite action**, `.github/actions/bun-install-retry`, and route **all four** install sites through it (3 in `verify` + the 1 in `docs-site`). One source of truth instead of growing the loop to 4 inline copies.

Loop semantics are **identical to #944** — only the location moved:
- up to **3** attempts of `bun install --frozen-lockfile`;
- purge `~/.bun/install/cache` before the **final** attempt so a poisoned half-written tarball is re-downloaded fresh;
- `sleep 5` between attempts; fail hard after 3;
- **only the install is wrapped** → a genuine lockfile/type error still fails fast (no 3× retry of check/build/lint);
- `--frozen-lockfile` kept everywhere.

## Deliberate deviation from prior art

#944 inlined the loop; this PR **extracts it to a composite action** — the DRY option the issue explicitly endorses ("removes the existing duplication"). The `docs-site` change is mechanical/behaviorally identical: same loop, same `working-directory: docs-site`, same cache step still preceding (and purged by) the install. Its `check`/`build` moved to a sibling step that sets `working-directory: docs-site` so they don't run in the repo root.

Note for reviewers: composite `run:` steps have **no default shell**, so the action declares `shell: bash` explicitly (mandatory, not stylistic).

## Verification

Local pre-flight (necessary, not sufficient): `ci.yml` + `action.yml` parse; `actionlint` clean; loop dry-run proves all three paths — success-first-try (0 retries, no purge), fail-then-pass (1 retry, no purge), always-fail (3 attempts, purge before final, non-zero exit).

**Authoritative acceptance = this PR's own CI run** going green on **both** `verify` (now installing through the action) **and** `docs-site` — local checks can't exercise the composite wiring (input plumbing, `working-directory` resolution, shell, local `uses: ./…` resolution).

CI-only change: no user-facing UX → no feature-catalog / i18n / glossary entries; commit is `ci:`, not `feat:`. `.husky/pre-push` unchanged — it doesn't run installs.

### DIFF
--- a/.github/actions/bun-install-retry/action.yml
+++ b/.github/actions/bun-install-retry/action.yml
@@ -1,0 +1,38 @@
+name: bun install (retry on tarball-extraction flake)
+description: >
+  Run `bun install --frozen-lockfile` with up to 3 attempts, purging the bun
+  store (~/.bun/install/cache) before the FINAL attempt so a poisoned half-written
+  native tarball is re-downloaded fresh rather than re-hit identically. Absorbs a
+  one-off tarball-extraction flake (#944 docs-site / #1134 verify) without masking
+  a genuine lockfile error. --frozen-lockfile is kept so lockfile integrity is still
+  enforced. Only the install is wrapped — callers run check/build/lint outside, so a
+  real error still fails fast.
+
+inputs:
+  working-directory:
+    description: Directory containing the package to install (default repo root).
+    required: false
+    default: "."
+
+runs:
+  using: composite
+  steps:
+    # `shell: bash` is MANDATORY for composite run steps — unlike a workflow `run:`
+    # they have no default shell and fail to load without it.
+    - shell: bash
+      working-directory: ${{ inputs.working-directory }}
+      run: |
+        n=0
+        until bun install --frozen-lockfile; do
+          n=$((n + 1))
+          if [ "$n" -ge 3 ]; then
+            echo "bun install failed after $n attempts" >&2
+            exit 1
+          fi
+          if [ "$n" -eq 2 ]; then
+            echo "purging bun store before final retry…" >&2
+            rm -rf ~/.bun/install/cache
+          fi
+          echo "bun install failed (attempt $n) — retrying in 5s…" >&2
+          sleep 5
+        done
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -48,14 +48,21 @@ jobs:
           bun-version: latest
 
       # Three packages, each with its own lockfile (root server + ui/ + extension/).
+      # Each install goes through the shared retry-with-store-purge composite action
+      # (.github/actions/bun-install-retry) so a one-off native-tarball-extraction flake
+      # self-heals instead of redding the whole job (#1134, extending docs-site's #944).
       - name: Install (root)
-        run: bun install --frozen-lockfile
+        uses: ./.github/actions/bun-install-retry
 
       - name: Install (ui)
-        run: cd ui && bun install --frozen-lockfile
+        uses: ./.github/actions/bun-install-retry
+        with:
+          working-directory: ui
 
       - name: Install (extension)
-        run: cd extension && bun install --frozen-lockfile
+        uses: ./.github/actions/bun-install-retry
+        with:
+          working-directory: extension
 
       - name: Lint (prettier + eslint)
         run: |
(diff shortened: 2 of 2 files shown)


scoop (commented):
## Summary

This PR satisfies issue #1134. It factors the previously-inline docs-site retry loop into a reusable composite action (`.github/actions/bun-install-retry`) and wires all three `verify`-job installs (root, ui, extension) plus the docs-site install through it. This is the issue's option 2 (composite action), which also removes the prior duplication. Change is CI-only; no source/CSS touched.

## Verification

- **Retry semantics preserved exactly.** The composite action's loop (`.github/actions/bun-install-retry/action.yml:21-37`) is identical to the inline loop that was deleted from the docs-site Build step (old `ci.yml` ~167-190, visible in the diff): up to 3 attempts, `rm -rf ~/.bun/install/cache` purge before the final (3rd) attempt (`n -eq 2`), `exit 1` after the 3rd failure (`n -ge 3`), 5s backoff. Confirmed against the removed block in the diff.
- **`--frozen-lockfile` retained** (`action.yml:26`) — lockfile integrity still enforced, per acceptance criteria.
- **Only install is wrapped** — `bun run check`/`build`/`lint`/`typecheck` run outside the action, so a genuine error still fails fast (no 3× retry of those). Verified at `ci.yml:67-74,188-189`.
- **`shell: bash` present** on the composite run step (`action.yml:22`) — mandatory for composite `run:` steps; correctly included.
- **`working-directory` input** correctly replaces the old `cd ui`/`cd extension`/`cd docs-site`. Composite run steps default to `github.workspace`, so `ui`/`extension`/`docs-site` resolve correctly and the default `"."` is repo root. `rm -rf ~/.bun/install/cache` is HOME-relative, unaffected by working-directory. Verified `ci.yml:55,58-60,63-65,181-183`.
- **Local-action path consistency.** All five `uses: ./.github/actions/bun-install-retry` references (`ci.yml:55,58,63,181`) match the new action file's directory. Both jobs run `actions/checkout` first (`ci.yml:39-40,154-155`), so the local action is present when referenced. Grep confirmed no other/stale spellings.
- **docs-site cache interaction unchanged.** The `actions/cache@v4` step (`ci.yml:166-172`) still precedes the install; on the rare flake the post-job save captures the freshly re-downloaded store — same as before #1134.

No blocking issues found.

- docs-site/README.md:32-33 still describes the CI build as the inline `cd docs-site && bun install --frozen-lockfile && bun run check && bun run build`; it now goes through the composite action + separate steps. Documentation-only drift in an untouched file — consider updating in a follow-up.
- .github/workflows/ci.yml:20-25 "KEEP IN SYNC WITH .husky/pre-push" note: the install-retry resilience is CI-only and adds/removes/reorders no gate, so no pre-push mirror is required by this change.

Issue #817, opened 2026-06-19
Flaky root test: intermittent 1-fail in pre-push (non-reproducible on rerun)


scoop:
## Symptom

The root test suite (`bun test ./test`) intermittently reports **1 failure** in the **pre-push hook**, then passes cleanly on immediate re-run. Observed 2026-06-19 while pushing a docs-only branch (`shepherd/research-claude-swap-integration`), so the change under push could not have caused it.

## Evidence from the failing run

The pre-push output captured:

```
 1 fail
 9653 expect() calls
Ran 3783 tests across 195 files. [9.64s]
```

A **clean** run of the same suite reports:

```
 0 fail
 9656 expect() calls
Ran 3783 tests across 195 files. [~10s]
```

Two things to note:

- `Ran 3783 tests across 195 files` identifies this as the **root** suite (`bun test ./test`), not `ui` or `extension`.
- The failing run logged **9653** expect() calls vs **9656** on green — a **3-expect delta**. That points at a single test that **threw partway through** (aborting its remaining assertions), not a whole file erroring out. i.e. a mid-test assertion failure, consistent with a race / shared-state / ordering bug rather than a hard breakage.

The pre-push tail did not include the failing test's name (only the summary scrolled into view).

## Reproduction attempts — could NOT reproduce (14 runs)

- 6× `bun test ./test` in isolation → 0 fail every time.
- 8× under the **exact pre-push conditions** — `unset $(git rev-parse --local-env-vars)` + `SHEPHERD_REPO_ROOT="$(mktemp -d)" bun test ./test` (per `.husky/pre-push:11,47`) → 0 fail every time.

So the flake is **rare and non-deterministic** — almost certainly dependent on test **execution order / timing / concurrency** (bun runs files concurrently by default) or cross-file shared state, rather than anything environmental that the reruns reproduce.

## Why it matters

- Pre-push is a **hard gate** — a flaky 1-fail blocks legitimate pushes (it blocked this one until a manual re-run), training contributors to reflexively re-push, which erodes the gate's value.
- A mid-test assertion failure that only shows under concurrency is usually a **real latent bug** (shared tmp dir, global mutated by a sibling test, time/order assumption), not "just noise."

## Suggested investigation

1. **Capture the test name.** Reproduce under load / repetition and surface the failing test. Useful knobs: run many iterations, and/or pin concurrency to force interleavings (e.g. tighten/loosen bun's concurrency, run `bun test ./test --rerun-each N`).
[...]

Issue #1135, opened 2026-06-26
Stale __distill__/__merge__/__optimize__ agents pile up — no boot reapOrphans for learnings helpers


scoop:
## Symptom

The herdr agent list accumulates many stale `__distill__<hex>` agents (also `__merge__<hex>`, `__optimize__<hex>`) sitting in `done`/`idle` — finished-turn but pane/`claude` still alive — and they never get cleaned up. They pile up monotonically until manually reaped.

## Root cause

Two compounding gaps, plus an ironic regression from the unique-name fix.

**1. In-flight state is memory-only → server restart orphans every live run.**
`Distiller` tracks runs only in an in-memory `this.inflight` map (`src/distiller.ts:122`, set at `:286`). Normal teardown happens in `tick()` (every 30s, `src/index.ts:1427`): a run is finalized on proposals-ready or `timeoutMs` (10 min) and closed via `herdr.stop(f.terminalId)` (`src/distiller.ts:318` → `closeTab`). That works **within one server lifetime**, but on restart the `inflight` map is gone, so `tick()` never finalizes those runs and `herdr.stop` is never called. The spawned `claude` is **interactive** (no `-p`/`--print`; `--permission-mode dontAsk`, `src/distiller.ts:259-266`), so after writing `.shepherd-learnings.json` it idles at the prompt forever (`agent_status:"done"` = finished-turn, pane alive — per herdr-done-is-idle-not-exited semantics).

**2. No boot reconcile for the distiller (or its two siblings).**
The reviewer hit this exact restart-orphan bug and fixed it with persisted `reviewer_spawns` rows + a boot `reapOrphans()` (`src/review.ts`, wired at `src/index.ts:861`); the doc-agent has the same (`src/index.ts:706`). **The distiller has neither** — no persisted spawn row, no boot reconcile. Confirmed: only `docAgent` and `reviewService` call `reapOrphans()` in `index.ts`. The two structurally-identical siblings — `src/optimizer.ts` (`__optimize__`) and `src/merge-suggest.ts` (`__merge__`) — have the same memory-only `inflight` + no boot reap.

**3. The safety-net tab reaper structurally can't catch these.**
[...]

PR #1143, opened 2026-06-26
fix(reaper): reap orphaned PPID-1 background jobs on worktree teardown (#1133)

### LINKED ISSUE #1133: Reap orphaned background jobs on worktree teardown (leaked `yes` load-generators from #817)
## What happened

34 orphaned `yes` processes were found pegging ~30 of 32 cores (load avg ~40) for **6 days**. Traced to Claude Code session `c398dce3…` working **#817 (flaky)** in `.shepherd-worktrees/shepherd-work-issue-817-flaky`.

The agent intentionally spawned `yes` as CPU-load generators to induce scheduler jitter and reproduce a timing-flaky test:

```bash
for i in $(seq 1 12); do yes > /dev/null & done   # "load for jitter"
for i in $(seq 1 24); do yes > /dev/null & done
for i in $(seq 1 10); do yes >/dev/null & done
```

Ran 3×, leaving batches of 12/12/10 = 34.

## Why they leaked

- Each loop ran in a separate `bash -c` (Bash tool) invocation. Backgrounded `yes &` jobs were **reparented to PID 1** when that shell exited.
- The agent's cleanup (`pkill -x yes; pkill -f repro817b.sh`) ran in a *different* tool-call's shell and couldn't reliably reap jobs orphaned by other invocations.
- The worktree was later deleted, severing the last context. Nothing reaped them → 6 days of ~30-core burn.

## Ask

Make worktree teardown reap leftover background processes spawned under that worktree, so throwaway scaffolding can't outlive the agent/worktree. Options:

- On worktree teardown, kill the process group / descendants associated with the worktree (e.g. track a session pgid, or sweep procs whose `cwd` resolves under the worktree path).
- Extend the existing `tmp-sweep` cleanup to also reap orphaned (PPID 1) processes whose `/proc/<pid>/cwd` points at a now-deleted shepherd worktree.
- Guardrail: discourage detached `&` load-gen in agent shells, or wrap repro scripts so they self-clean via `trap`.

## Impact

Silent, severe: ~30 cores wasted for ~6 days, undetected. Any leaked busy-loop has the same failure mode.

scoop:
Closes #1133.

## Problem

34 PID-1-reparented `yes` load-generators from a #817 session pegged ~30 cores for **6 days**, undetected. Each `yes &` was backgrounded in a separate `bash -c`; when that shell exited the kernel reparented the job to PID 1, so it outlived the agent and the (later deleted) worktree. The existing port-based class-2 detector can't catch these — a busy-loop listens on no port, which is exactly the signal that detector keys on.

## Fix — three complementary layers

Reuses the existing `ProcessReaper` + injectable `ReaperProbes` machinery; the new methods drop the port test and key on the **PPID-1 orphan signal** instead.

- **A. Teardown sweep (primary).** `ProcessReaper.reapOrphansUnder(worktreePath)` SIGKILLs every PPID-1, non-agent (`comm ∉ {claude,codex}`), non-self process whose cwd is under the worktree. `WorktreeMgr.remove()` calls it **before** unlinking (so cwds still resolve for matching), best-effort so a sweep failure never blocks teardown.
- **B. Boot/daily safety net (deleted-cwd only).** `reapDeletedWorktreeOrphans()`, wired into `fireTmpSweep`, SIGKILLs PPID-1 orphans whose cwd is an already-**deleted** `.shepherd-worktrees/*` path — reaps anything that leaked before this fix or via a removal that didn't run the teardown sweep. Zero false-positive risk.
- **C. Guardrail.** One bullet appended to the universal `ENGINEERING_POSTURE` spawn directive: don't leave detached `&` jobs past the step that needs them; `trap`-self-clean throwaway repro scripts.

**Shared kill predicate (both sweeps):** `pid ≠ shepherd` ∧ `comm ∉ {claude,codex}` ∧ `ppid == 1` ∧ cwd-under-worktree (A) / deleted-`.shepherd-worktrees` cwd (B). Signal = **SIGKILL** (throwaway orphans under a worktree being/already destroyed). A new `ppidForPid` probe reads `/proc/<pid>/status` per-candidate — kept off the hot `scanProcs()` pollers.

**Defensive precondition (A):** `reapOrphansUnder` refuses any path lacking `/.shepherd-worktrees/`, so a future caller passing a repo root cannot SIGKILL unrelated host processes (mirrors the existing non-isolated guard in `scanWorktreeProcs`/`detect`).

## Known residual gaps (deliberate, flagged per house rules)

1. **Abandoned-but-on-disk worktrees** (session gone, worktree still on disk): out of scope — needs liveness/session-state checks and risks killing a legit backgrounded process in a quiet-but-active worktree.
2. **chdir-out orphans** (agent `cd`s out of the worktree before backgrounding): cwd is neither under the worktree nor a deleted `.shepherd-worktrees` path, so it leaks like #817. The cwd heuristic catches the dominant case (#817's jobs kept their cwd in the worktree). **`pgid` alternative declined:** Shepherd records no per-session pgid; agent shells run under herdr/PTY, each `bash -c` is its own process, and `setsid`-away jobs defeat group tracking. The robust fix is a per-session PID namespace — which the autonomous **membrane already provides** via bwrap `--unshare-pid`; the residual is only the trusted/attended path.
3. **Non-isolated sessions get no protection.** `service.ts` calls `worktree.remove()` only `if (s.isolated)`; a non-isolated session's `worktreePath` *is* the shared repo root, which we deliberately never sweep (it hosts the shepherd server + unrelated processes — a PPID-1 sweep there would be more dangerous than the leak).

**`PPID === 1` assumption:** holds only for host-reparented trusted-path spawns. Membrane orphans (bwrap `--unshare-pid` + `--die-with-parent`) are PPID ≠ 1 from the host and correctly skipped. A future subreaper / PID-namespace on the *trusted* path would silently defeat the predicate — documented in the code Risks so it can't regress unnoticed.

## Tests

- `reapOrphansUnder`: SIGKILLs a matching orphan; spares self / agent comm / PPID≠1 / out-of-worktree; **refuses a non-worktree path**; no-ops when `ppidForPid` is absent.
- `reapDeletedWorktreeOrphans`: SIGKILLs a deleted-worktree orphan; spares live cwd, deleted-but-non-worktree cwd, agent comm, PPID≠1, self.
- `WorktreeMgr.remove()` invokes the sweep (injected reaper), and does so while the worktree still exists.
- Full root suite green (4947 pass / 0 fail); `bun run lint`, `tsc --noEmit`, and `fallow audit --fail-on-issues` all pass.

Server-only change, no user-facing UX. `[no-feature-entry]`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -78,7 +78,7 @@ import { tailLines } from "./blocked";
 import { recommendPrompt } from "./prompt-recommend";
 import { CountsService } from "./backlog";
 import { BacklogPoller } from "./backlog-poller";
-import { ProcessReaper } from "./process-reaper";
+import { ProcessReaper, reapDeletedWorktreeOrphans } from "./process-reaper";
 import { sweepClaudeTmp, compileCacheDir, reapFallowCaches, pruneRepoWorktrees } from "./tmp-sweep";
 import { runSessionUsageBackfill } from "./usage-backfill";
 import { PreviewService } from "./preview";
@@ -414,6 +414,18 @@ const fireTmpSweep = (phase: "boot" | "daily") => {
       ),
     )
     .catch((err) => console.warn(`[tmp-sweep] ${phase} fallow/prune failed:`, err));
+
+  // Safety net for #1133: reap detached PPID-1 orphans (leaked `yes`/load-gen busy-loops)
+  // whose cwd is an already-deleted shepherd worktree — leaks the teardown sweep missed
+  // (pre-fix, or a removal path that didn't run it). Deferred off the synchronous boot
+  // path; never throws onto the loop.
+  void Promise.resolve()
+    .then(() => {
+      const { reaped } = reapDeletedWorktreeOrphans();
+      if (reaped > 0)
+        console.warn(`[tmp-sweep] ${phase}: reaped ${reaped} deleted-worktree orphan(s)`);
+    })
+    .catch((err) => console.warn(`[tmp-sweep] ${phase} orphan reap failed:`, err));
 };
 
 fireTmpSweep("boot");
--- a/src/process-reaper.ts
+++ b/src/process-reaper.ts
@@ -59,10 +59,28 @@ export interface ReaperProbes {
    * Falls back to `portsForPid` when absent.
    */
   socketInodesForPid?(pid: number): number[];
+  /**
+   * Parent pid for a single process (the `PPid:` field of /proc/<pid>/status), or
+   * null when unreadable. Used by the orphan sweeps to recognise a process the
+   * kernel has reparented to PID 1 (its launching shell exited) — the load-bearing
+   * "this is an orphan, not a live child" signal. Optional + read per-candidate so
+   * the hot scanProcs() pollers never pay for it.
+   */
+  ppidForPid?(pid: number): number | null;
 }
 
 // The agent itself runs `claude` with cwd == worktree; never offer to kill it.
 const AGENT_COMMS = new Set(["claude", "codex"]);
+
+// Path segment every shepherd session/review worktree lives under. The orphan
+// sweeps refuse to act on any path lacking it, so a mistaken caller passing a repo
+// root (or any non-worktree path) can never SIGKILL unrelated host processes.
+const WORKTREE_MARKER = "/.shepherd-worktrees/";
+
+// readlink(/proc/<pid>/cwd) appends this once the directory is unlinked; the kernel
+// keeps the inode alive for the still-running process. A "(deleted)" cwd under the
+// worktree marker is the unambiguous signal of an orphan whose worktree is already gone.
+const DELETED_SUFFIX = " (deleted)";
 
 /** Stable selection key — lets the client echo a choice back without trusting raw pids. */
 export function leftoverKey(l: Pick<Leftover, "kind" | "name" | "port" | "pid">): string {
@@ -146,6 +164,18 @@ function listeningInodeToPort(): Map<number, number> {
   return map;
 }
 
+/** Read a process's parent pid from the `PPid:` line of /proc/<pid>/status. */
+function readPpid(pid: number): number | null {
+  let text: string;
+  try {
+    text = readFileSync(`/proc/${pid}/status`, "utf8");
+  } catch {
+    return null; // process gone, or not ours to inspect
+  }
+  const m = text.match(/^PPid:\s+(\d+)/m);
+  return m ? Number(m[1]) : null;
+}
+
 /** Read the socket inodes held open by a single process from /proc/<pid>/fd. */
 function readSocketInodes(pid: number): number[] {
   let fds: string[];
@@ -211,6 +241,9 @@ const defaultProbes: ReaperProbes = {
   },
   socketInodesForPid(pid) {
     return readSocketInodes(pid);
+  },
+  ppidForPid(pid) {
+    return readPpid(pid);
   },
   readTranscript(path) {
     try {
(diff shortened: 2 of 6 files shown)


scoop:
Residual gaps documented in the PR body are tracked in #1144 (resource-gated orphan reaper for the cwd-blind / non-isolated leaks), cross-linked into the #1135/#1136 reap cluster.

scoop (commented):
## Summary

The PR satisfies issue #1133. It adds two reaping layers plus a prompt guardrail:

- **Layer A — teardown** (`reapOrphansUnder`, called from `WorktreeMgr.remove` *before* unlinking): SIGKILLs PPID-1, non-agent, non-self processes whose `cwd` resolves under the worktree. Verified `WorktreeMgr.remove` runs the sweep before `existsSync`/git removal (`src/worktree.ts:511` precedes the removal block at `:514+`), so cwds still resolve to the live path — the order the test at `test/worktree.test.ts:44-57` asserts via `existsSync(p)===true` inside the stub.
- **Layer B — boot/daily safety net** (`reapDeletedWorktreeOrphans`, wired in `src/index.ts:419-427`): SIGKILLs PPID-1 orphans whose `cwd` is a `'… (deleted)'` shepherd-worktree path — catching leaks that predate the fix or whose removal bypassed Layer A.
- **Guardrail**: `ENGINEERING_POSTURE` gains a self-clean-scaffolding rule (`src/service.ts:397-401`).

### Verification performed
- `' (deleted)'` suffix reliance is correct: `scanProcs` derives cwd via `readlinkSync('/proc/<pid>/cwd')` (`src/process-reaper.ts:215`), and the kernel appends `' (deleted)'` to that readlink target once the dir is unlinked — matching `DELETED_SUFFIX` (`:84`) and the test fixture `${WT} (deleted)` (`test/process-reaper.test.ts:430`).
- Safety guards confirmed against code: self (`p.pid === process.pid`), agent comms (`AGENT_COMMS = {claude,codex}` at `:71`), and the `WORKTREE_MARKER` path precondition (`:78`) all gate both sweeps (`src/process-reaper.ts:367-371`, `:415-419`). The `(this.probes.ppidForPid?.(p.pid) ?? null) !== 1` form means a missing `ppidForPid` probe kills nothing — verified by the test at `test/process-reaper.test.ts:425-435`.
- New `WorktreeMgr(reaper)` constructor is backward-compatible: it defaults to `new ProcessReaper()` (`src/worktree.ts:51`), and every existing call site uses the no-arg form (`new WorktreeMgr()` across `src/index.ts:289` and the test suites) — grepped, no caller breaks.
- `reapDeletedWorktreeOrphans` is exported and imported correctly in both `src/index.ts:81` and the test (`test/process-reaper.test.ts:6`).
- No circular import introduced: `process-reaper.ts` imports only `node:fs`/`instrument`/`usage`/`jsonl`; `worktree.ts` → `process-reaper.ts` is one-directional.
- No user-facing strings / i18n catalogs touched (the `service.ts` change is internal prompt text), so no locale-parity concern.

No blocking defects found.

### Could not verify (environmental, out of repo)
- Whether the production host reparents orphans to PID 1 vs. a subreaper — see limitation below. The incident report empirically observed PPID 1, which is what the code targets.

- src/process-reaper.ts: `reapOrphansUnder`'s `WORKTREE_MARKER` precondition `(root + "/").includes("/.shepherd-worktrees/")` would also pass if a caller handed it the *parent* `…/.shepherd-worktrees` directory itself, which would then sweep orphans across ALL worktrees in one call. Not reachable today (every `remove()` caller passes a specific child worktree path), but the guard is one level coarser than its intent — consider requiring a path segment *after* the marker.

- src/process-reaper.ts: the `/proc` scan→signal sequence (`scanProcs` then `killPid`) carries an inherent PID-reuse TOCTOU window shared with the pre-existing `stopListenersOnPort`; negligible here since reaped targets are busy-loops that don't self-exit, but it is a latent property of the whole `/proc`-probe approach.

## Notes / limitations (informational, non-blocking)
- **Subreaper environments**: the PPID-1 filter is the orphan signal, but on hosts where a `PR_SET_CHILD_SUBREAPER` process (e.g. a `systemd --user` manager) sits above the agent shells, orphans reparent to *that* manager's PID, not 1 — both layers would then miss them. The original incident empirically found the leaked `yes` procs at PPID 1, so the chosen signal matches the observed deployment; worth a comment acknowledging the subreaper assumption so a future reader doesn't assume universal coverage.

Issue #1147, opened 2026-06-26
Stale `autopilot <id>` agents leak identically to the learnings helpers — no boot reap


scoop:
Follow-up split from #1135 (boot `reapOrphans` for the learnings trio `__distill__`/`__optimize__`/`__merge__`).

## Symptom
@scoop's diagnostic on #1135 found that leaked `autopilot <id>` agents pile up by the **same** mechanism as the learnings helpers: long-lived idle-alive `claude` panes (4 instances, up to 10 days old in one live instance) that the husk-only tab reaper (`src/tab-reaper.ts`) spares because their foreground proc is a non-shell `claude`. They are **not** covered by #1135's fix, which is deliberately scoped to the three learnings helpers.

## Why deferred from #1135
Autopilot is **not** a transient `inflight`/`tick` helper that finishes one turn and idles — it's the long-lived driver of a live session and can be **legitimately mid-work**. The learnings helpers' safe reap signal ("name matches `*_LABEL` AND not in the current-process `inflight` map") does **not** transfer: an autopilot agent absent from any in-memory map at boot may still be a session that should be re-adopted, not killed. A correct fix needs autopilot-specific liveness/ownership reasoning (which session owns it, is it still driving, should it be re-adopted vs. reaped) — distinct from the name-based boot scan #1135 ships.

## Scope to investigate
- How autopilot agents are named/tracked and whether a persisted spawn row (vs. name scan) is the right reconcile primitive here.
- Distinguish a genuine restart-orphan from a live/re-adoptable session before closing anything (fail-safe: spare when unsure).
- Confirm whether `recap`/`rundown`/`namer` are already restart-safe (they appeared covered during #1135 — recap/rundown persist `reviewer_spawns` rows; namer is a short-lived `-p` spawn).

## Related
- #1135 — boot reapOrphans for the learnings trio (the fix-bearing issue this splits from).
- #1136 — symptom-side diagnostic report.
- #1133 — sibling orphan-reap theme.

scoop:
**Sweep 2026-06-27 — partial periodic coverage now exists; boot-reap gap remains.**

Update: `autopilot <id>` panes are now matched by the tab-reaper's husk detection (`src/tab-reaper.ts:60`, `label.startsWith("autopilot ")`), which runs on the periodic/boot-debounced sweep. So idle husks are no longer entirely unswept.

[...]

PR #750, opened 2026-06-17
fix(learnings): restore distiller — unique agent name, bounded concurrency, fail-closed health

scoop:
## Why

New learnings stopped surfacing — even for repos with thousands of signals. Diagnosed as a **distillation outage**, not a capture problem: signals flow fine (`shepherd`: 2274 signals) and `learningsEnabled=1` everywhere, but the newest learning in the live DB was 2026-06-14.

**Root cause:** every distill run spawned its ephemeral analyst under one *shared* herdr agent name `__distill__`. The daily sweep calls `consider()` for every repo in a tight loop, and the per-repo `inflight` guard doesn't know the name is globally in use — so the 2nd-onward repo collided with `agent_name_taken`, and a stranded `__distill__` husk in the long-lived herdr server (which survives Shepherd restarts) blocked even the first. The spawn error was swallowed as a `console.warn`, so the flywheel died silently. Live `shepherd.log` held **2882** `agent_name_taken` errors (~400 recent `[distill] spawn failed`). Cross-repo because the name is global — matching "even repos I don't usually work in."

## What

- **Unique per-run agent name** (`__distill__<8hex>`; `DISTILL_LABEL` kept as the underscore-bounded, slug-collision-proof prefix) — eliminates the collision class. Reaper now prefix-matches it.
- **Bounded concurrency** — `maxConcurrent` (default 3) + a deduped queue drained by `tick()`; both `consider()` and the manual `distillNow()`/backlog-kick route through one shared slot/queue, so unique names don't turn a sweep into an N-wide burst.
- **Fail-closed health** — global, end-to-end (counts real spawn/write failures and finalize-timeout-with-no-output; excludes `HerdrUnavailableError` + the api-key-not-configured skip; resets on any completed run). `GET /api/learnings/health` + a persistent "distiller stalled" banner in the Learnings drawer (EN+DE), so a future silent outage is visible.
- **Maintenance guard** on `runDailySweep` so sweeps don't fire mid-herdr-update.

## Validation

- Root: `lint` ✓, `tsc --noEmit` ✓, `bun test ./test` 3374 pass / 0 fail.
- UI: `check` 0/0, `check:i18n` parity (1414 keys), `bun run test` 1380 pass.
- Hygiene: branch-hygiene ✓, feature-catalog ✓, glossary ✓, `fallow audit` → no issues in changed files.
- Built subagent-driven (4 tasks, per-task spec+quality reviews + an opus whole-branch review: *Ready to merge*).

[...]

PR #1148, opened 2026-06-26
fix(learnings): boot reapOrphans for __distill__/__optimize__/__merge__ helpers (#1135)

### LINKED ISSUE #1135: Stale __distill__/__merge__/__optimize__ agents pile up — no boot reapOrphans for learnings helpers

scoop:
Closes #1135.

## Problem
The three learnings helpers (`distiller`, `optimizer`, `merge-suggest`) track in-flight runs only in a **memory-only `inflight` map**. On a server restart that tracking is lost, so:

- `tick()` (the 30s finalizer) never sees the prior-lifetime runs → never calls `herdr.stop`.
- Each run's spawned `claude` is **interactive** (`--permission-mode dontAsk`, no `-p`), so after writing its output it idles at the prompt forever (`agent_status:"done"` = finished-turn, pane alive).
- `reapOrphanTabs` (`src/tab-reaper.ts`) only closes a tab once its pane decays to a shell-only husk; an idle-but-alive `claude` is a non-shell proc, so it's classified `sparedLive` and spared.

Result: `__distill__<hex>` / `__optimize__<hex>` / `__merge__<hex>` agents pile up monotonically until manually reaped. PR #750's unique per-run names made it worse — pre-#750 a new run's shared name collided with the orphan and herdr's collision-retry evicted it; now nothing ever collides.

## Fix
Mirror the existing reviewer (`reviewService.reapOrphans`) / doc-agent (`docAgent.reapOrphans`) boot reconcile, adapted to these helpers' shape:

- Add a **name-based** `reapOrphans()` to each of the three services. It scans `herdr.list()` once for agents whose name `startsWith` the helper's exported `*_LABEL` and are **not** owned by a current-process `inflight` run (by `terminalId`), and `closeTab`s them. Best-effort: a herdr-unavailable `list()` at boot is caught and no-ops.
- Widen each service's `herdr` dep `Pick` to add `"list" | "closeTab"`.
- Call `distiller.reapOrphans()` / `optimizer.reapOrphans()` / `mergeSuggest.reapOrphans()` once at boot in `index.ts`, right after each service is constructed.

A **name-based scan** (no new persisted-spawns table) is the issue's explicitly-preferred approach — an orphan has already written or abandoned its output, so closing the idle tab is the whole job; no cross-restart timeout accounting is needed.

The `ownedTerms` guard means a run started by the **current** process is never reaped (at boot `inflight` is empty, so every matching tab is an orphan).

## Scope
Learnings trio only, per the issue's "Fix (handoff)". `review`/`plan-gate`/`pr-critic` already boot-reconcile; `recap`/`rundown` persist `reviewer_spawns` rows; `namer` is a short-lived `-p` spawn. The leaked `autopilot <id>` agents (@scoop's diagnostic) leak by the same mechanism but are a long-lived session driver where the "not in inflight" reap signal is unsafe — split out to **#1147** for a fix with autopilot-specific ownership reasoning.

## Tests
Per service (`test/{distiller,optimizer,merge-suggest}.test.ts`):
- **orphan closed** — a listed `*_LABEL` agent absent from `inflight` has its tab closed.
- **unrelated spared** — a real session-slug / `review TASK-09` agent is never closed.
- **in-flight-owned spared** — a seeded in-flight run (its `terminalId` present in the live list) is **not** closed (the issue's explicit verification ask).
- **herdr-unavailable no-throw** — `list()` throwing does not throw and closes nothing.

## Verification
- `bun run lint` clean; `bunx tsc --noEmit` clean.
- `bun test ./test` — 4953 pass / 0 fail (incl. 6 new tests).
- `bunx fallow@2.97.0 audit --base origin/main --fail-on-issues` — exit 0 (the 3 near-identical `reapOrphans` methods register as a duplication **warn**; these services are established structural siblings that already duplicate `health`/`tick`/`begin`, and mirroring the pattern across all three is what the issue mandates).

No user-facing strings (server-only plumbing) → i18n/feature-catalog/glossary gates N/A.

### DIFF
--- a/src/distiller.ts
+++ b/src/distiller.ts
@@ -99,7 +99,7 @@ export interface DistillerDeps {
     | "retireLearning"
     | "getLearning"
   >;
-  herdr: Pick<HerdrDriver, "start" | "stop">;
+  herdr: Pick<HerdrDriver, "start" | "stop" | "list" | "closeTab">;
   scratch: { create: () => { dir: string }; remove: (dir: string) => void };
   onChange: () => void;
   model?: string | null;
@@ -290,6 +290,32 @@ export class DistillerService {
       startedAt: this.now(),
       signalIds: new Set(signals.map((s) => s.id)),
     });
+  }
+
+  /** Boot reconcile (issue #1135): close orphaned distiller tabs left by a PRIOR server
+   *  lifetime. `inflight` is memory-only, so a restart loses tracking of live runs; the
+   *  spawned interactive `claude` idles at the prompt forever after writing its output
+   *  (agent_status "done" = finished-turn, pane alive), and the husk-only tab reaper
+   *  (tab-reaper.ts) spares it as an alive (non-shell) `claude`. Scan herdr once for agents
+   *  whose name starts with DISTILL_LABEL and are NOT owned by a current-process inflight
+   *  run, and close their tabs. Name-based — no persisted state (the issue's preferred
+   *  approach); the prefix's underscores can't appear in a real session slug. */
+  reapOrphans(): void {
+    const ownedTerms = new Set(
+      [...this.inflight.values()].map((f) => f.terminalId).filter(Boolean),
+    );
+    let reaped = 0;
+    try {
+      for (const a of this.deps.herdr.list()) {
+        if (!a.name.startsWith(DISTILL_LABEL)) continue;
+        if (ownedTerms.has(a.terminalId)) continue; // spare a live run started by THIS process
+        this.deps.herdr.closeTab(a.tabId);
+        reaped++;
+      }
+    } catch (err) {
+      console.warn("[distill] reapOrphans:", err); // herdr may be unavailable at boot — no-op
+    }
+    if (reaped > 0) console.warn(`[distill] reapOrphans: closed ${reaped} orphan tab(s)`);
   }
 
   /** Finalize any run whose proposals file is ready or that timed out, then drain queue. */
--- a/src/index.ts
+++ b/src/index.ts
@@ -1436,6 +1436,7 @@ const distiller = new DistillerService({
   scratch: defaultScratch,
   onChange: () => events.emit("learnings:update", { pending: store.pendingLearningCount() }),
 });
+distiller.reapOrphans(); // issue #1135: close orphaned __distill__ tabs left by a prior lifetime
 setInterval(() => {
   if (maintenance.active) return;
   void distiller.tick();
@@ -1448,6 +1449,7 @@ const optimizer = new OptimizerService({
   promoter,
   onChange: () => events.emit("learnings:update", { pending: store.pendingLearningCount() }),
 });
+optimizer.reapOrphans(); // issue #1135: close orphaned __optimize__ tabs left by a prior lifetime
 setInterval(() => {
   if (maintenance.active) return;
   void optimizer.tick();
@@ -1461,6 +1463,7 @@ const mergeSuggest = new MergeSuggestionService({
   scratch: defaultMergeScratch,
   onChange: () => events.emit("learnings:update", { pending: store.pendingLearningCount() }),
 });
+mergeSuggest.reapOrphans(); // issue #1135: close orphaned __merge__ tabs left by a prior lifetime
 setInterval(() => {
   if (maintenance.active) return;
   void mergeSuggest.tick();
--- a/src/merge-suggest.ts
+++ b/src/merge-suggest.ts
@@ -68,7 +68,7 @@ export interface MergeSuggestionDeps {
     | "getMergePassSignature"
     | "setMergePassSignature"
   >;
-  herdr: Pick<HerdrDriver, "start" | "stop">;
+  herdr: Pick<HerdrDriver, "start" | "stop" | "list" | "closeTab">;
   scratch: { create: () => { dir: string }; remove: (dir: string) => void };
   onChange: () => void;
   model?: string | null;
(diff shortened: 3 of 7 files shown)


scoop (commented):
## Summary

The PR adds a boot-time `reapOrphans()` to the three transient learnings-helper services (`DistillerService`, `OptimizerService`, `MergeSuggestionService`) and wires each one into `src/index.ts` on boot, exactly as the issue handoff requests. Implementation is correct and consistent across all three.

## Verification

- **Dep contract extension is sound.** Each `*Deps.herdr` `Pick` is widened from `"start" | "stop"` to add `"list" | "closeTab"` (src/distiller.ts:102, src/optimizer.ts:61, src/merge-suggest.ts:71). Both methods exist on `HerdrDriver` with the used signatures: `list(): HerdrAgent[]` (src/herdr.ts:211) and `closeTab(tabId: string)` (src/herdr.ts:510).
- **Fields accessed on the listed agents all exist.** `reapOrphans` reads `a.name`, `a.terminalId`, `a.tabId` — all present on `HerdrAgent` (src/herdr.ts:16,18,19).
- **Orphan vs. live discrimination is correct.** `ownedTerms` is built from `this.inflight.values().map(f => f.terminalId)` and an agent is spared when `ownedTerms.has(a.terminalId)`. `InFlight.terminalId` is the value returned by `herdr.start(...).terminalId` and stored in `inflight` (src/distiller.ts:80,270-289; mirrored in the siblings), so the comparison key matches. At cold boot `inflight` is empty, so every prefix-matching tab is reaped — the intended behavior.
- **`closeTab(a.tabId)` is equivalent to the normal teardown.** Normal close is `herdr.stop(terminalId)` which itself resolves `terminalId -> tabId` via the live list and calls `closeTab` (src/herdr.ts:474-477). Since `list()` already yields `tabId`, calling `closeTab` directly avoids a redundant lookup. Correct.
- **Failure path is safe.** `herdr.list()` is wrapped in try/catch; a `HerdrUnavailableError` (or any error) at boot is logged and the method no-ops without closing anything (verified by the new "no-op when herdr is unavailable" tests). `maintenance` defaults to `_active = false` (src/maintenance.ts:9), so a cold boot does not spuriously short-circuit the reap via `HerdrUnavailableError`.
- **Boot wiring matches the existing pattern.** `distiller.reapOrphans()` / `optimizer.reapOrphans()` / `mergeSuggest.reapOrphans()` are each invoked once immediately after construction and before their `setInterval` tick loop (src/index.ts:1439,1452,1466), mirroring `docAgent.reapOrphans()` (src/index.ts:718) and `reviewService.reapOrphans()` (src/index.ts:873). The new methods are synchronous (no shelled-out async work beyond the single `list()`), so the un-awaited synchronous call is fine.
- **Prefix-match safety.** `DISTILL_LABEL`/`OPTIMIZE_LABEL`/`MERGE_LABEL` are `__distill__`/`__optimize__`/`__merge__` (src/distiller.ts:36, src/optimizer.ts, src/merge-suggest.ts:24), matching the per-run agent names `LABEL + sessionId.slice(0,8)` (src/distiller.ts:250). The double-underscore prefix cannot collide with a real session slug, so the name-based scan won't reap unrelated agents — also asserted by the new tests sparing `"my-feature-branch"` / `"some-session"` / `"review TASK-09"`.

The PR satisfies issue #1135 for all three helpers, including the explicit requirement that a legitimately in-flight run started by the current process is NOT reaped (covered by the inflight-owned spare path and its tests).

- src/distiller.ts: `reapOrphans()` runs exactly once at boot and silently no-ops if `herdr` is unavailable at that instant (e.g. server booted during an active maintenance window). Since it is never retried, orphans present at such a boot would persist until the next restart. Today `maintenance.active` is false at cold boot so this is not reachable; it would only matter if a future change made maintenance active during process init. Same applies to the optimizer and merge-suggest copies.

No blocking issues found.

Issue #1156, opened 2026-06-26
Idle/done task sessions never auto-archive (sessions table grows)


kai-osthoff:
Split out from #1136 (secondary finding) — **different root cause** from the helper-tab leak fixed there, tracked separately so the tab-reaper PR stays surgical.

## Symptom

Non-archived DB sessions accumulate: `sessions` rows with `archivedAt IS NULL` pile up because **idle/done task sessions are never auto-archived**. On the instance in #1136 there were 28 (26 idle, 1 done); @scoop's instance saw only 2 — so impact is instance-dependent, but the mechanism is real.

## Root cause (code-verified)

- The `sessions` table has a nullable `archivedAt` column (`src/store.ts`), but there is **no auto-archive path** for settled sessions.
- Archival happens only on an explicit `session:archived` event (manual UI action) or operator pruning. An idle-but-settled or done session stays `archivedAt IS NULL` indefinitely.

## Direction (not yet decided)

- Auto-archive a session once it is verifiably settled (e.g. `done`/long-idle with a generated recap and no live agent), on a boot + periodic reconcile — mirroring how recap's `considerForArchive` already gates on settled state.
- Keep it conservative: never archive a session with a live agent or unsynced work.

## Relation

- #1136 — parent diagnostic; the helper-tab husk leak (separate root cause) is fixed by the boot helper-tab reconcile. This issue is the leftover session-accumulation half.

kai-osthoff:
## Field data: unarchived sessions don't just grow a table — they hold herdr tabs open

Ran a full correlation of `herdr tab list` against `~/.shepherd/shepherd.db` on a long-running
host today (Shepherd 1.46.0, herdr 0.7.5). The sessions table growing is the visible half; the
expensive half is that **every unarchived session keeps a herdr tab alive, and so does every
archived one** — the tab outlives the archive.

### Session state

```
sessions total            281
  archived                250
  NOT archived             31   <- "HERD 31" in the top bar

of the 31:
  done                     26
  blocked                   2
  idle                      2
  running                   1
```

26 of 31 are `done`. `updatedAt` is useless as an age signal here — all 26 carry the same
`2026-08-04 11:25:29` from a bulk poll. By `createdAt` the backlog is weeks deep:

| createdAt | sessions still unarchived |
|---|---|
| 2026-07-16 | 1 |
| 2026-07-20 | 3 |
| 2026-07-21 | 1 |
| 2026-07-25 | 1 |
[...]

Issue #1093, opened 2026-06-25
Deepen: one transient-agent module behind the 10 spawn argv builders


scoop:
**Architecture review finding · `Strong`** — deepening opportunity surfaced by an architectural review (deletion-test framed). _Top recommendation of the review._

## Problem
**10** modules independently assemble the same transient-`claude` argv: `reviewer-argv`, `doc-agent-argv`, `namer-llm`, `autopilot-llm`, `distiller`, `optimizer`, `recap`, `herd-digest`, `merge-suggest`, `verify-key`. The settings shape `JSON.stringify({ disableAllHooks: true, ...apiKeySettingsFragment() })` is copied **8×** byte-for-byte; the variadic `--allowedTools` flag-order invariant is re-explained per file; the session-id lifecycle has 3 inconsistent patterns; the allowlist drifts per consumer.

Worst: **orphan reaping (`adoptOrphans`) exists in only 2 of the 10 spawns.** The other ~8 never reap — the exact gap behind the recurring reviewer name-squat outages.

## Deepening
One deep `buildTransientAgentArgv(kind, …) → { argv, sessionId }` owning the argv shape, flag order, settings spread, and allowlist presets; plus an adoptable spawn base that reaps orphans for **every** kind. The auth seam already proves the pattern — `spawn-auth` centralises auth with two real adapters (`spawn-auth` subscription / `spawn-membrane` api-key), so those seams stay; only the duplicated argv shape collapses.

```
before: 10 argv builders, same shape, 2/10 reap orphans
after:  buildTransientAgentArgv(kind) ──seam──> spawn-auth | spawn-membrane
```

## Deletion test
Delete any one builder → its argv shape reappears verbatim in the next. The duplication is the earns-its-keep signal; the shape wants one home.

## Wins
- leverage: one builder, 10 call sites
- locality: flag-order invariant stated once
- orphan reaping everywhere, not 2 of 10 (closes name-squat outages)
- allowlist drift becomes impossible

_Vocabulary per `/codebase-design`: module · interface · depth · seam · adapter · leverage · locality._

scoop:
**Descope note — Win #1 ("orphan reaping everywhere" / universal adoptable spawn base).** PR #1151 lands the core deepening (one `buildTransientAgentArgv(kind)` behind all 10 spawn sites + dedupe the byte-identical label-prefix reaper) but **deliberately does not** build a universal spawn base, with operator confirmation.

[...]

PR #1157, opened 2026-06-26
fix(tab-reaper): boot-reap pr-critic + unclean-exit synchronous helper tabs (#1136)

### LINKED ISSUE #1136: herdr helper tabs leak: ephemeral agents die but their persistent tabs survive restarts; reaper can't keep up
## Summary

herdr accumulates dozens of orphaned helper-agent tabs that never get cleaned up. On the instance where this was found there were **~74 herdr tabs but only 3 with a live process** — the other ~71 are husks (idle `bash` shells) left behind by ephemeral helper agents (`name TASK-x`, `review TASK-x`, `__distill__<hex>`, `autopilot <id>`, `recap TASK-x`).

The dedicated safety net (`src/tab-reaper.ts` → `reapOrphanTabs`) exists but in practice almost never catches up: across the entire `shepherd.log` it succeeded **exactly once** (`[tabs] reaped 37 husk tab(s)`).

## Root cause (code-verified)

The helper **agents** are ephemeral; their herdr **tabs** are persistent. Two lifecycles, and the second isn't closed reliably.

1. Shepherd spawns short-lived helpers, each in its own herdr tab. When the helper finishes, the agent **process** exits — but closing the **tab** is a *separate*, **in-memory-tracked** `closeTab` (`doc-agent.ts` / `review.ts`).
2. **herdr is a persistent daemon that survives shepherd restarts.** Shepherd restarts constantly — **242 restarts** in the analyzed log. Each restart loses the in-memory handle, so `closeTab` never fires for in-flight helpers.
3. herdr persists the orphaned tab to `~/.config/herdr/session.json` and **re-materializes it as a fresh idle `bash` shell on the next boot**. Proof: all 44 husk `bash` shells shared the *exact* start timestamp of the herdr server process (same second), and all 44 helper labels are present in `session.json`.

### Why the reaper safety net doesn't keep up

The reaper logic is currently *correct* (live re-run classified 44/44 husks as reapable), but three compounding fragilities keep it from firing:

| | Fragility | Evidence |
|---|---|---|
| A | **Fail-closed on flaky `pane process-info`.** Empty `foreground_processes` → tab is spared *and* the debounce counter is not armed. | Log regime `spared 0 live, 45 undetermined`, repeated |
| B | **Two-sweep in-memory debounce + sparse cadence.** Reaps only after shell-only on two *consecutive* sweeps; sweeps run @+5s, @+45s, then **hourly**; debounce state (`shellOnlyTabs`) is RAM-only. | `index.ts:428-445` |
| C | **242 restarts** wipe the debounce each time and restart the +5s/+45s/hourly clock; the server rarely survives long enough for two successful consecutive sweeps. | last restart in log is *after* the last reaper log line → current instance logged no sweep at all |

**One-line:** herdr resurrects tabs from `session.json` as bash husks on restart; shepherd's "close tab" lives only in RAM and is lost on each of the 242 restarts; the reaper is too fragile (fail-closed + RAM debounce + hourly) to win the race against the restart frequency.

## Verification steps (for @scoop — reproduce on your own instance)

```bash
# 1. Total herdr tabs vs. tabs with a live agent process
herdr agent list | python3 -c 'import json,sys
a=json.load(sys.stdin)["result"]["agents"]
print("total tabs:", len(a))
print("with live agent:", sum(1 for x in a if "agent" in x))'

# 2. Count persisted helper husks in session.json
python3 -c '
import json
d=json.load(open("'"$HOME"'/.config/herdr/session.json"))
labels=[]
def w(o):
    if isinstance(o,dict):
        for k,v in o.items():
            if k=="label" and isinstance(v,str): labels.append(v)
            w(v)
    elif isinstance(o,list):
        [w(x) for x in o]
w(d)
h=[l for l in labels if l.startswith(("review ","name ","autopilot ","recap ","__distill__","plan-review ","pr-critic "))]
print("persisted helper husks:", len(h))'

# 3. Confirm a husk pane runs only a shell (no agent)
herdr pane list | python3 -c 'import json,sys
p=[x for x in json.load(sys.stdin)["result"]["panes"] if (x.get("label") or "").startswith("review ")]
print(p[0]["pane_id"], p[0]["label"]) if p else print("none")'
# then: herdr pane process-info --pane <pane_id>   → expect only ["bash"]/["zsh"]

# 4. Reaper history — how often did it ever actually reap?
grep -E '\[tabs\]' ~/.shepherd/shepherd.log

# 5. Restart count (each line = one server boot)
grep -c "shepherd core on" ~/.shepherd/shepherd.log
```

Expected on an affected instance: total tabs ≫ live agents; double-digit persisted husks; husk panes show only a shell; the reaper rarely/never logs a successful reap; high restart count.

## Suggested directions (not yet implemented)

- **Harden the reaper** (most general): denser cadence in the first minutes after boot instead of jumping straight to hourly; persist the debounce state across restarts; or treat a helper-labeled tab whose pane is shell-only as reapable without requiring two *in-memory* consecutive sweeps.
- **Fix at the source / on boot:** since herdr resurrects helper tabs from `session.json` anyway, do an idempotent boot-time sweep that closes any helper-labeled tab with no live agent (the husk set is fully knowable from `session.json` + a single `pane process-info` pass).
- **Don't persist helper tabs:** mark ephemeral helper tabs as non-persistent so herdr doesn't write them to `session.json` / resurrect them.

## Secondary (separate, smaller)

28 non-archived DB sessions (`sessions` table, `archivedAt IS NULL`) — 26 idle, 1 done — accumulate because idle/done task sessions aren't auto-archived. Different root cause from the tab husks, but contributes to the "so many sessions" impression.

---
*Filed from a diagnostic-only session; nothing was changed. Counts/log line numbers are from the instance where this was observed.*

kai-osthoff:
Closes #1136.

## What changed (and why the approach was reworked mid-review)

When this branch was opened, #1135 was unfixed and I proposed a single **generalized un-gated** boot reaper over all helper labels. During review two things became clear:

1. **#1135 has since landed on main** — `distiller`/`optimizer`/`mergeSuggest` each gained an **ownership-gated** `reapOrphans()` (the shared `reapTransientByLabel`, wired synchronously at boot, index.ts:1458/1471/1485). My generalized reaper was therefore redundant for those, and — being un-gated — it *raced* the `+10s` daily-sweep that spawns live `__distill__`/`__merge__` tabs (the critic's last finding). Correct catch.
2. `reapTransientByLabel`'s own docstring documents the maintainers' deliberate fleet teardown design: a universal "reap every kind" base is **intentionally out of scope**; each helper family reconciles via its own lifecycle.

So I **reverted the generalized reaper entirely** (`tab-reaper.ts` is back to baseline) and closed the *remaining* #1136 leak surface using the **same established `reapTransientByLabel` pattern** as #1135:

- **standalone PR critic (`pr-critic `)** — memory-only `inFlight`, no row/state boot reconcile (index.ts wired only `tick`/`sweep`/`stopAll`). A restart orphans its interactive `claude`, which idles at the prompt forever and is spared by the husk-only reaper as a live non-shell proc. Added `StandalonePrCriticService.reapOrphans()` mirroring distiller (owned-set from `inFlight` terminalIds), wired synchronously at boot.
- **`name ` / `autopilot ` / `verify api key`** — the synchronous block-and-clean helpers. #1093 left them unreaped on the premise their `finally`-stop leaves no husk — true only on a **clean** exit. A restart mid-poll skips the `finally` and orphans the `claude`. This is the concrete #1136 evidence: @scoop measured **10-day-old `autopilot` husks**. They track no inflight and none runs at the synchronous boot point, so `reapTransientByLabel` with an **empty owned set** correctly closes every prior-lifetime orphan.

## Why this can't race a live spawn

Every reap runs **synchronously at boot**, in the same eval phase as the #1135 reaps — before any spawn timer (`+10s` daily sweep, `+3s` probe) or event fires. There is no async-chain window for a current-lifetime helper to be spawned into and then wrongly closed. Ownership-gating (pr-critic) is a second guard for any later re-run.

## Scope notes

- Server-only; no UI → no feature-catalog / i18n / glossary entries (`fix`).
- usage probe is untouched — it self-sweeps by name (`UsageProbeService.sweep()`).
- review / plan-review / recap / rundown keep their existing reconciles (persisted `reviewer_spawns` / `generating` state).
- Secondary #1136 finding (idle/done DB sessions never auto-archive — different root cause) split out to #1156.

## Files

- `src/standalone-critic.ts` — widen `herdr` dep to `list`+`closeTab`; add `reapOrphans()`.
- `src/index.ts` — wire `standaloneCritic.reapOrphans()` + three synchronous `reapTransientByLabel` calls for the mechanism-4 helpers.
- `src/transient-tab-reaper.ts` — update the design note: record the unclean-exit gap these boot reaps now close.
- `test/standalone-critic.test.ts` — `reapOrphans` closes orphaned `pr-critic ` tabs / no-ops when herdr is unavailable.

`bun run lint` ✓ · `bunx tsc --noEmit` ✓ · `bun test` (standalone-critic, transient-tab-reaper, distiller, tab-reaper: 115 pass) ✓

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -37,6 +37,7 @@ import { resolveDiffBase } from "./diff-base";
 import { BranchPruner } from "./branch-pruner";
 import { reconcile } from "./reconcile";
 import { reapOrphanTabs, reapStaleReviewWorktrees } from "./tab-reaper";
+import { reapTransientByLabel } from "./transient-tab-reaper";
 import { scanClaudeAliveByWorktree } from "./process-reaper";
 import { serve, serveAgentIngress, buildBacklogPayload, type AppDeps } from "./server";
 import { PluginRegistry } from "./plugins/loader";
@@ -773,6 +774,7 @@ const standaloneCritic = new StandalonePrCriticService({
         .map((s) => s.branch!),
     ),
 });
+standaloneCritic.reapOrphans(); // issue #1136: close orphaned pr-critic tabs left by a prior lifetime
 
 // Pre-execution plan gate (#348): the planning-phase twin of the PR critic. An
 // adversarial reviewer reads the agent's `.shepherd-plan.md` BEFORE it writes code;
@@ -1485,6 +1487,15 @@ setInterval(() => {
   if (maintenance.active) return;
   void mergeSuggest.tick();
 }, 30_000);
+// Issue #1136: the synchronous block-and-clean helpers (namer / autopilot / verify-key) stop their
+// spawn in a `finally`, so they leave NO husk on a CLEAN exit — but a server restart mid-poll skips
+// that finally, orphaning an interactive `claude` that idles at the prompt forever (the husk-only
+// reaper spares it as a live non-shell proc). They track no inflight and none is running at this
+// synchronous boot point, so an empty owned set is correct: close every prior-lifetime orphan by
+// label prefix. Space-prefixed / multi-word labels can't collide with an `[a-z0-9-]` session slug.
+reapTransientByLabel(herdr, "name ", new Set(), "[namer]");
+reapTransientByLabel(herdr, "autopilot ", new Set(), "[autopilot]");
+reapTransientByLabel(herdr, "verify api key", new Set(), "[verify-key]");
 const gitignoreAdopter = new GitignoreAdopter({ worktree, resolveForge });
 // Daily: prune archived sessions, prune old signals, then consider a distill per repo
 // with enough recent signal.
--- a/src/standalone-critic.ts
+++ b/src/standalone-critic.ts
@@ -39,6 +39,7 @@ import {
   type RawVerdict,
 } from "./critic-core";
 import type { VerdictRead } from "./json-tolerant";
+import { reapTransientByLabel } from "./transient-tab-reaper";
 import { resolveSpawnMembrane, type MembraneSeams } from "./spawn-membrane";
 
 /** One PR critic run between its spawn (begin) and its verdict (finalize). Carries everything
@@ -74,7 +75,7 @@ export interface StandalonePrCriticDeps extends MembraneSeams {
     | "completeReviewerSpawn"
     | "listEpicCompleted"
   >;
-  herdr: Pick<HerdrDriver, "start" | "stop">;
+  herdr: Pick<HerdrDriver, "start" | "stop" | "list" | "closeTab">;
   worktree: Pick<WorktreeMgr, "createDetached" | "remove" | "gitCommonDir">;
   resolveForge: (repoPath: string) => GitForge | null;
   /** Candidate repos to consider each sweep. The sweep itself filters to those with
@@ -602,6 +603,20 @@ export class StandalonePrCriticService {
    *  these (a re-adopted #631 orphan's tick() still needs its worktree). */
   inflightWorktrees(): string[] {
     return [...this.inFlight.values()].map((f) => f.worktreePath);
+  }
+
+  /** Boot reconcile (#1136): close orphaned `pr-critic ` tabs left by a prior lifetime. The
+   *  standalone critic's `inFlight` is memory-only, so a restart loses tracking of a live run;
+   *  its interactive `claude` idles at the prompt forever (agent_status "done", pane alive) and
+   *  the husk-only {@link reapOrphanTabs} spares it as a non-shell `claude`. Mirrors the
+   *  distiller/optimizer/merge-suggest reap (#1135): scan herdr once at boot for `pr-critic `
+   *  agents NOT owned by a current-process inflight run and close their tabs. The disposable
+   *  `*-review-*` worktrees these orphans hold are reaped separately by `reapStaleReviewWorktrees`. */
+  reapOrphans(): void {
+    const ownedTerms = new Set(
+      [...this.inFlight.values()].map((f) => f.terminalId).filter(Boolean),
+    );
+    reapTransientByLabel(this.deps.herdr, "pr-critic ", ownedTerms, "[pr-critic]");
   }
 
   /** Tear down every in-flight run (reap its terminal + worktree). For process shutdown. */
--- a/src/transient-tab-reaper.ts
+++ b/src/transient-tab-reaper.ts
@@ -5,15 +5,22 @@ import type { HerdrDriver } from "./herdr";
  *
  * This is ONE of four distinct teardown mechanisms across the transient-agent fleet, and the only
  * one whose body was byte-identical across consumers (distiller / optimizer / merge-suggest) — so
- * only it is deduped here. The other seven consumers gain NO new boot reaping from #1093 BY DESIGN:
- * they already have appropriate teardown via a different lifecycle, and forcing them onto this
- * label-prefix scan would be wrong, not a missing feature —
+ * it was deduped here. The OTHER consumers reconcile via a different lifecycle, and forcing them
+ * onto this label-prefix scan would be wrong where they already have it —
  *   - persisted `reviewer_spawns` rows → adoptOrphans/reapOrphans (plan-gate / review / doc-agent);
- *   - store `generating` state → reapGenerating (recap / herd-digest);
- *   - synchronous block-and-clean — start→poll→stop in a `finally`, random temp dirs, no stable
- *     name to squat → no husk on clean exit (namer / autopilot / verify-key).
- * So #1093 delivers the argv consolidation across all 10 plus this one reaper dedup; a universal
- * "reap every kind" base is intentionally out of scope (see the PR description + issue #1093).
+ *   - store `generating` state → reapGenerating (recap / herd-digest).
+ * #1093 delivers the argv consolidation across all 10 plus this reaper dedup; a universal
+ * "reap every kind" base remains out of scope.
+ *
+ * **The exception: synchronous block-and-clean helpers — start→poll→stop in a `finally`, random
+ * temp dirs (`name ` / `autopilot ` / `verify api key`).** #1093 left these unreaped on the
+ * premise that the `finally` leaves no husk. That holds only on a CLEAN exit: a server restart
+ * mid-poll skips the `finally`, orphaning an interactive `claude` that idles at the prompt forever
+ * (the husk-only reaper spares it as a live non-shell proc). #1136 closes that gap by calling this
+ * same reaper for those three labels at boot with an EMPTY owned set — they track no inflight and
+ * none is running at the synchronous boot point, so every match is a prior-lifetime orphan. The
+ * standalone PR critic (`pr-critic `) — memory-only `inFlight`, no row/state reconcile — likewise
+ * gains an owned-set-gated boot reap here (#1136).
  *
  * Why this mechanism exists: `inflight` is memory-only, so a server restart loses tracking of live
  * runs; the spawned interactive `claude` idles at the prompt forever after writing its output
--- a/test/standalone-critic.test.ts
+++ b/test/standalone-critic.test.ts
@@ -747,3 +747,44 @@ test("stopAll: reaps in-flight run without throwing (regression: unbound this)",
   expect(spies.removed).toEqual(["/review-wt"]);
   expect(svc.inflightWorktrees()).toEqual([]);
 });
+
+// ── boot reapOrphans (issue #1136) ──────────────────────────────────────────
+
+test("reapOrphans closes orphaned pr-critic tabs, sparing unrelated names", () => {
+  const closed: string[] = [];
+  const listed = [
+    { name: "pr-critic /r#42", terminalId: "orphan1", tabId: "tabO", agentStatus: "done" },
+    { name: "my-feature-branch", terminalId: "u1", tabId: "tabU", agentStatus: "running" },
+    { name: "review TASK-09", terminalId: "rv1", tabId: "tabR", agentStatus: "done" },
+  ];
+  const { deps } = makeDeps({
+    herdr: {
+      start: () => ({ terminalId: "rt" }),
+      stop: () => {},
+      list: () => listed,
+      closeTab: (id: string) => closed.push(id),
+    },
+  });
+  const svc = new StandalonePrCriticService(deps as any);
+  svc.reapOrphans(); // no in-flight runs at boot → empty owned set
+  expect(closed).toEqual(["tabO"]); // only the pr-critic orphan; unrelated + sibling reviewer spared
+});
+
+test("reapOrphans is a no-op when herdr is unavailable", () => {
+  let closes = 0;
+  const { deps } = makeDeps({
+    herdr: {
+      start: () => ({ terminalId: "rt" }),
+      stop: () => {},
+      list: () => {
+        throw new Error("herdr unavailable");
+      },
+      closeTab: () => {
+        closes++;
+      },
+    },
+  });
+  const svc = new StandalonePrCriticService(deps as any);
+  expect(() => svc.reapOrphans()).not.toThrow();
+  expect(closes).toBe(0);
+});


kai-osthoff (commented):
## Summary

The PR adds a boot-time reconcile (`reapUnreconciledHelperTabs`) that closes any herdr tab labelled for a helper whose owning service does *no* boot re-adoption, on the invariant that at startup the current process owns none of them, so any matching tab is a cross-restart orphan. The label-set split (include vs. exclude) is the heart of the change, and I verified it against the actual owners.

### Verified correct

- **`HerdrTab` shape / `closeTab` signature** — the new `BootReapableHerdr = Pick<HerdrDriver, "tabs" | "closeTab">` and the test fake match the real interface (`src/herdr.ts:23-30` tabId/label/agentStatus/workspaceId; `closeTab(tabId)` at `src/herdr.ts:510`; `tabs()` at `src/herdr.ts:227`).
- **Label prefixes resolve and are spelled consistently** — `DISTILL_LABEL="__distill__"`, `OPTIMIZE_LABEL="__optimize__"`, `MERGE_LABEL="__merge__"` (`src/distiller.ts:36`, `src/optimizer.ts:27`, `src/merge-suggest.ts:24`), `PROBE_NAME="__usage_probe__"` (`src/usage-probe.ts:17`). `isUnreconciledHelperLabel` (`src/tab-reaper.ts:84-95`) uses the same forms as `isShepherdHelperLabel`.
- **Included owners genuinely do NOT re-adopt across restart** (so closing their tabs at boot is safe): distiller (`this.inflight` in-memory, `tick()` iterates only that map — `src/distiller.ts:122,296-297`), optimizer (`src/optimizer.ts:81,271`), merge-suggest (same family), pr-critic (`this.inFlight` in-memory, `tick()` iterates only it — `src/standalone-critic.ts:112,470-471`). `name`/`autopilot`/`verify api key` are synchronous spawn-and-finally helpers with nothing running at boot.
- **Excluded owners do reconcile** — review (`reviewService.reapOrphans`), plan-review (`planGate.adoptOrphans`), recap/rundown (tick re-reads tmpdir). Correctly omitted from the include set; over-conservatism here can only leave husks for the periodic sweep, not cause a bug.
- **No per-`closeTab` try/catch in the boot loop** (`src/tab-reaper.ts:123-127`) — consistent with the existing periodic reaper, which also closes in a bare loop (`src/tab-reaper.ts:222`), and the whole call is wrapped in a try/catch at the call site (`src/index.ts:871-877`). Not a defect.

### Finding: boot reaper vs. the +3s usage-probe timer is not ordered

The docstring asserts the close is "provably safe" because the reaper "MUST run once at boot, before the current process spawns any helper of these kinds" (`src/tab-reaper.ts:99-113`). That holds for every included helper *except the usage probe*, which is the one helper kicked off by a fixed near-boot timer rather than a long interval / daily sweep / user event: `setTimeout(calibrate, 3_000)` (`src/index.ts:1659`). `calibrate` → `usageLimits.calibrate` → `probe.scrape()`, which spawns a `__usage_probe__` tab (`src/usage-probe.ts:90-114`) and then drives it for ~6-15s.

The reaper, by contrast, runs at the tail of an unbounded async chain — `planGate.adoptOrphans() -> gcStaleReviewWorktrees() -> reviewService.reapOrphans() -> reKick -> reapUnreconciledHelperTabs` (`src/index.ts:858-878`) — built from herdr-subprocess and git calls whose duration is not bounded. Nothing synchronizes the two. If that chain resolves *after* the +3s probe has spawned its tab (i.e. reconcile takes >3s, most plausible on exactly the loaded, many-tab instances this PR targets), the reaper sees the live, current-lifetime `__usage_probe__` tab and `closeTab`s it, killing the probe mid-scrape. Impact is bounded/self-healing (the scrape returns null, calibration is skipped this boot and retries daily / next boot), so this is non-blocking, but it does contradict the "provably safe" claim. Consider gating the boot `calibrate` behind the reconcile chain (or deferring the first probe until after it), or relax the docstring to acknowledge the probe's best-effort ordering.

### Could not verify

- Actual cross-restart behavior of herdr 0.7 pane persistence (husk vs. idle-alive `claude`) is external to this repo; I took the diff's characterization as stated.

kai-osthoff:
**Addressed the boot-race finding (commit 4bb93e75).**

Valid catch — the usage probe is the one covered helper with a near-immediate boot spawn (`setTimeout(calibrate, 3_000)`, index.ts:1659), and the async reconcile chain can resolve after +3s, so the reap could have closed the live current-lifetime probe.

Fix: **dropped `__usage_probe__` from `isUnreconciledHelperLabel`** entirely, rather than re-order the chain. The probe already self-reconciles — `UsageProbeService.sweep()` stops every leftover `__usage_probe__` agent *by name* at the start of each `scrape()` — so it belongs in the excluded/self-healing group alongside review/recap, not the boot-reap set. Zero coverage loss: its husks stay covered by that name-based sweep and by the periodic shell-only `reapOrphanTabs` (which still lists `PROBE_NAME` in `isShepherdHelperLabel`).

I also verified the **other** covered helpers have no comparable near-boot spawn timer: distiller/optimizer/merge-suggest spawn from the daily sweep (+30s reap tick), pr-critic from a 60s tick, verify-key on demand. `name`/`autopilot` are event-driven (no fixed timer); a session created inside the sub-second chain window could spawn one the reap then closes, but that is self-correcting and safe-by-default (killed namer → heuristic name; killed autopilot → SURFACE, no auto-proceed). Softened the `reapUnreconciledHelperTabs` docstring to state this best-effort ordering honestly rather than claim unconditional 'provably safe'.

kai-osthoff:

**Prior finding #1 (usage-probe boot race) — ADDRESSED.** `PROBE_NAME` (`__usage_probe__`, usage-probe.ts:17) is no longer in `isUnreconciledHelperLabel` (tab-reaper.ts:90-98 — none of the listed prefixes `__distill__`/`__optimize__`/`__merge__`/`name `/`autopilot `/`pr-critic `/`verify api key` match `__usage_probe__`). So `reapUnreconciledHelperTabs` will not close the probe tab spawned by the `setTimeout(calibrate, 3_000)` boot timer. The probe's husks remain covered by `UsageProbeService.sweep()` (name-based) and the periodic shell-only `reapOrphanTabs` (`isShepherdHelperLabel` still lists `PROBE_NAME`, tab-reaper.ts:48). The probe test cases (test/tab-reaper.test.ts:451,494) assert this exclusion. Good.

## New finding — same race class, NOT addressed for distiller/optimizer/merge-suggest

The author's note claims the remaining covered helpers "have no fixed near-boot spawn timer — distiller/optimizer/merge-suggest spawn from the daily sweep (+a 30s reap tick)." Verified against the code, this is wrong: the daily sweep's **first** fire is `setTimeout(runDailySweep, 10_000)` (index.ts:1639), and `runDailySweep` synchronously spawns **current-lifetime** runs:
- distiller: `for (const repo of listRepos(...)) distiller.consider(repo.path)` (index.ts:1580) → `enqueueOrBegin` → `beginRun` → `this.deps.herdr.start(...)` (distiller.ts:248), which creates a `__distill__<hex>` herdr tab. (Gated on ≥ minSignals, default 5 — reachable on a busy repo.)
- merge-suggest: `mergeSuggest.consider(repo.path)` + `mergeSuggest.considerCrossRepo()` (index.ts:1584-1585) → `__merge__<hex>` tab.

The boot reap runs in the async reconcile chain `planGate.adoptOrphans().then(gcStaleReviewWorktrees).then(reapOrphans).then(reKick).then(<boot reap>)` (index.ts:887-909). On exactly the orphan-heavy instances this PR targets (74 husk tabs, 242 restarts), `adoptOrphans`/`reapOrphans` do real work (re-adopt plan reviewers, re-kick reviews with forge round-trips), so the chain can resolve **after +10s**. When it does, `reapUnreconciledHelperTabs` matches the live `__distill__`/`__merge__` tab via `isUnreconciledHelperLabel` (tab-reaper.ts:93,95) and `closeTab`s it — with **no ownership spare** (unlike `reapTransientByLabel`, which spares `ownedTerminalIds`, transient-tab-reaper.ts:40). This kills the freshly-spawned current-lifetime distiller/merge run, losing that distill cycle. It is the same boot-race the prior review flagged for the probe, with a *wider* window (+10s vs +3s) and a justification (tab-reaper.ts:116-122 / index.ts:895-901: "no fixed near-boot spawn timer") that the code at index.ts:1639/1580/1584 directly contradicts. optimizer is analogous via its +30s first tick (index.ts:1486) — wider margin, same shape.

Notably, covering DISTILL/OPTIMIZE/MERGE in the boot reap is also **redundant**: each already self-reconciles its own orphans at boot via `reapOrphans()` (index.ts:1472/1485/1499 → `reapTransientByLabel`), and their husks are covered by the periodic shell-only reaper (`isShepherdHelperLabel` lists DISTILL/OPTIMIZE/MERGE, tab-reaper.ts:50-52). So excluding them from `isUnreconciledHelperLabel` loses nothing — exactly the argument used to justify excluding the probe — while removing the race. The genuinely-needed boot-reap coverage (`pr-critic` @ +60s sweep index.ts:1043-1046, `name`/`autopilot` event-driven, `verify api key` on-demand) is unaffected.

### Fix options
- Drop `DISTILL_LABEL`/`OPTIMIZE_LABEL`/`MERGE_LABEL` from `isUnreconciledHelperLabel` (parallel to the probe exclusion; they self-reconcile + are husk-reaped already), **or**
- Defer/gate the boot reap until after the first `runDailySweep` so it cannot race the +10s spawns.

Verification basis: index.ts:1639 (`setTimeout(runDailySweep, 10_000)`), index.ts:1580/1584-1585 (distiller/merge consider), distiller.ts:248 (`herdr.start` creates the tab), index.ts:887-909 (async boot chain), tab-reaper.ts:93-95/136 (label match + unconditional closeTab), transient-tab-reaper.ts:40 (the ownership spare the boot reap lacks), index.ts:1472/1485/1499 (existing self-reconcile), tab-reaper.ts:50-52 (existing periodic husk coverage).

## Notes / limitations
- `name`/`autopilot` event-driven spawns inside the sub-second chain window are acknowledged in the docstring as self-correcting (killed namer → heuristic name; killed autopilot → SURFACE). Not raising those — the degradation is bounded and documented.
- Could not exercise the exact wall-clock timing of the boot reconcile chain on a live orphan-heavy instance (no runtime here); the race is established by the relative scheduling above, not measured.

kai-osthoff:
**Addressed the +10s daily-sweep race (commit cc34ac98) — and reworked the approach to match the established pattern.**

Valid finding: `setTimeout(runDailySweep, 10_000)` (index.ts) spawns live `__distill__`/`__merge__` tabs at +10s, and my generalized **un-gated** boot reaper sat in an async chain that could resolve after that → it could close a current-lifetime distiller/merge tab.

While verifying, I found the deeper issue: **#1135 already landed on main** — distiller/optimizer/merge-suggest each have an **ownership-gated** `reapOrphans()` (`reapTransientByLabel`, wired synchronously at boot). So my generalized reaper was both redundant for them *and* the source of the race, and it's exactly the universal 'reap every kind' base that `reapTransientByLabel`'s docstring says is intentionally out of scope.

So I **reverted the generalized reaper entirely** (`tab-reaper.ts` is back to baseline) and closed the remaining #1136 surface with the **same `reapTransientByLabel` pattern**:

1. **`pr-critic`** (standalone critic — memory-only inFlight, no boot reconcile): added `reapOrphans()` mirroring distiller, owned-set-gated from its `inFlight` terminalIds.
2. **`name`/`autopilot`/`verify api key`** (synchronous block-and-clean): #1093 left these unreaped assuming the `finally`-stop leaves no husk — true only on a *clean* exit. A restart mid-poll skips it (your autopilot husks). `reapTransientByLabel` at boot with an empty owned set closes those prior-lifetime orphans.

**Race is gone by construction:** all reaps now run **synchronously at boot**, the same eval phase as the #1135 reaps — before the +10s daily sweep, the +3s probe, or any event fires. No async-chain window, so nothing current-lifetime can be spawned into it. Net diff is now 4 files (+83/-13); updated the `reapTransientByLabel` design note to record the unclean-exit gap these boot reaps close.

kai-osthoff:
## Verdict: request-changes

The #1136 reaping additions themselves are correct and well-targeted — but the same hunk in `index.ts` **accidentally deleted the only production driver of `mergeSuggest.tick()`**, a real regression that ironically re-introduces the exact tab-leak class this PR exists to fix.

### BLOCKING — `src/index.ts`: `mergeSuggest.tick()` interval deleted

The parent commit had, right after `mergeSuggest.reapOrphans()`:
```js
setInterval(() => {
  if (maintenance.active) return;
  void mergeSuggest.tick();
}, 30_000);
```
(verified at `git show 6d920fda:src/index.ts` lines 1484-1487). This PR's hunk **replaced** that block with the three `reapTransientByLabel(...)` calls instead of inserting them alongside it (current `src/index.ts:1485-1494`). There is now **no** call to `mergeSuggest.tick()` anywhere in production — grep of the whole tree finds `mergeSuggest.tick` only in this deleted spot, and `.tick()` callers for it only in `test/merge-suggest.test.ts` (verified: `Grep mergeSuggest\.tick|\.tick\(\)` returns test files only; `Grep \.tick\(` in `src/index.ts:1003-1474` lists distiller/optimizer/etc. but no mergeSuggest).

Consequences, since `tick()` is the sole finalizer (`src/merge-suggest.ts:286-303`):
- In-flight merge runs are **never finalized** — their output is never read/applied (`merge-suggest.ts:289-294`), so the merge-suggestion feature is silently dead.
- `this.deps.herdr.stop(f.terminalId)` is **never called** (`merge-suggest.ts:315`) → every `__merge__` terminal spawned by the daily sweep leaks. This is the same husk class #1136 targets.
- The queue is never drained (`merge-suggest.ts:296-302`).

The sibling services kept their identical interval (distiller `src/index.ts:1459-1462`, optimizer `1472-1475`); only mergeSuggest's was lost. The comment at `src/index.ts:1476-1478` ("the 30s tick reaps finished/timed-out runs") is now stale and corroborates that the interval was meant to stay.

**Fix:** restore the `setInterval(... void mergeSuggest.tick() ..., 30_000)` block and keep the three `reapTransientByLabel(...)` calls in addition to it (not in place of it).

---

### Re-review item (prior round) — RESOLVED

*"Boot reap closes live current-lifetime distiller/merge tabs spawned by the +10s daily sweep."* Confirmed addressed. The generalized async boot reaper was reverted (`tab-reaper.ts` is not in this diff, i.e. back to baseline). All new reaps are **synchronous at module-eval boot**: `standaloneCritic.reapOrphans()` at `src/index.ts:777`, and the three label reaps at `src/index.ts:1492-1494`, all run before the `+10s` `setTimeout(runDailySweep, 10_000)` (`src/index.ts:1630`), the `+3s` probe, or any interval fires. At those synchronous points no current-lifetime helper tab can exist, so every match is a prior-lifetime orphan. No async-chain window remains. Verified.

### Other correctness checks (all pass)

- Label prefixes resolve to real spawn labels: `pr-critic ` → `standalone-critic.ts:426` (`pr-critic ${repoPath}#${pr.number}`); `name ` → `service.ts:2112` (`name ${session.desig}`); `autopilot ` → `autopilot.ts:348,374` (`autopilot ${id}`); `verify api key` (exact) → `verify-key.ts:147`. All match `isShepherdHelperLabel`'s collision-proof set (`tab-reaper.ts:54-60`).
- Collision safety: user sessions use `[a-z0-9-]` slugs (no space/underscore), so the space-prefixed / multi-word labels can't be reached by a user slug even though `reapTransientByLabel` has no liveness gate (verified `tab-reaper.ts:20-26`). The only exact, space-free helper label that relies on the liveness gate (`rundown`) is correctly NOT in the new reaper's set.
- `reapTransientByLabel` field access (`a.name`/`a.terminalId`/`a.tabId`) matches `HerdrAgent` (`herdr.ts:11-21`) and `list()`'s return (`herdr.ts:211-222`). `verifyKey` is an on-demand callback (`index.ts:1838`), not a synchronous boot spawn, so the boot reap can't race a current-lifetime verify-key tab.
- `standalone-critic.ts:78` widening `herdr` Pick to add `list`/`closeTab` is satisfied by the real `HerdrDriver`; all test constructions use `as any` so no test breaks. `reapOrphans` (`standalone-critic.ts:611-620`) mirrors `distiller.reapOrphans` (`distiller.ts:281-286`) exactly; `InFlight.terminalId` is non-optional (`standalone-critic.ts:57`). Tests at `test/standalone-critic.test.ts:753-790` exercise both the close-orphan and herdr-unavailable paths consistently.

No i18n/message-catalog strings are involved — the new strings are `console.warn` log tags only.

kai-osthoff:
**Fixed (commit ecc5a415).** Correct catch — my earlier edit's replace-block accidentally dropped the production `setInterval(() => { if (maintenance.active) return; void mergeSuggest.tick(); }, 30_000)` while inserting the three `reapTransientByLabel` calls. That regressed merge-suggest finalization (no `tick()` caller → in-flight runs never finalized, `herdr.stop` at merge-suggest.ts:315 never fires → leaks the exact `__merge__` husk class #1136 targets, queue never drains).

Restored the `setInterval` verbatim and kept the three `reapTransientByLabel` calls after it. The `index.ts` diff vs main is now purely additive — all three `distiller`/`optimizer`/`mergeSuggest` `.tick()` schedulers present, no production lines removed. tsc/lint/tests green.

kai-osthoff (commented):
## Review

This PR extends the proven `reapTransientByLabel` boot-reconcile (#1135) to the transient-agent gaps that #1136 identified: the synchronous block-and-clean helpers (`name `/`autopilot `/`verify api key`) and the memory-only standalone PR critic (`pr-critic `). The mechanism is sound and the change is purely additive.

### Prior finding #1 — resolved
The earlier round flagged that the `reapTransientByLabel` insertion had deleted the production `setInterval(() => { if (maintenance.active) return; void mergeSuggest.tick(); }, 30_000)` block. Verified resolved against `src/index.ts:1486-1489` (via `git show HEAD:src/index.ts`): the `setInterval` is intact and the three new `reapTransientByLabel` calls follow it. The `index.ts` diff is now additive only — no production line removed.

### Correctness checks (verified, not assumed)
- **Label prefixes match every spawn site**: `"name "` ↔ `src/service.ts:2112` (`name ${session.desig}`); `"autopilot "` ↔ `src/autopilot.ts:348` (`autopilot ${id}`); `"verify api key"` ↔ `src/verify-key.ts:147`; `"pr-critic "` ↔ `src/standalone-critic.ts:426` (`pr-critic ${repoPath}#${pr.number}`). All match `startsWith` semantics in `src/transient-tab-reaper.ts:46`.
- **Husk tabs are reachable via `list()`**: `HerdrDriver.list()` (`src/herdr.ts:211-224`) maps `agent list` output preserving `name` even for husks (`agent: a.agent ?? ""`), so resurrected husks are matched and closed — consistent with the #1135 distiller/optimizer/merge mechanism this mirrors.
- **Types/contracts**: `HerdrAgent` carries `name`/`tabId`/`terminalId` (`src/herdr.ts:16-19`), matching the fields `reapTransientByLabel` reads (`src/transient-tab-reaper.ts:46-48`). `InFlight.terminalId` exists (`src/standalone-critic.ts:57`) and feeds the owned-set filter; `this.deps` is stored (`constructor` at `src/standalone-critic.ts:133`). The `herdr` Pick was widened to `"start"|"stop"|"list"|"closeTab"` (`src/standalone-critic.ts:78`); production passes the full `HerdrDriver` (`src/index.ts:763`), which provides all four.
- **Empty-owned-set boot reap is safe**: both `standaloneCritic.reapOrphans()` (`src/index.ts:777`) and the three `reapTransientByLabel` calls (`src/index.ts:1493-1495`) run at synchronous module-init, before any helper/critic run exists in-process, so an empty owned set closes only prior-lifetime orphans. The `[a-z0-9-]` session-slug convention (confirmed by `isShepherdHelperLabel` at `src/tab-reaper.ts:47-60`) means the space-prefixed/multi-word labels cannot collide with a real session.
- **Tests** (`test/standalone-critic.test.ts:750-790`): the happy-path test asserts only the `pr-critic ` orphan is closed while an unrelated branch and a sibling `review ` reviewer are spared; the unavailable-herdr test confirms the `try/catch` no-ops. Both align with `reapTransientByLabel` behavior.

No blocking issues found.

PR #1166, opened 2026-06-27
feat(viewport): read-only scratchpad file browser (#1164)

### LINKED ISSUE #1164: Session scratchpad file browser (read-only, download)
## Problem

Claude Code frequently writes artifacts into its **session scratchpad** (e.g. a generated `configuration.yaml` for Home Assistant). There is currently **no way to retrieve those files from within Shepherd** — the operator can see an agent produced something but can't navigate to it or download it.

## Goal

A **read-only file browser** in the session view, rooted at the session's scratchpad, that lets the operator descend into subdirectories and **download individual files**.

### Scope & security boundary
- **Locked to the scratchpad subtree.** Root = the session's scratchpad dir; navigation can descend into subdirs but **never above the root**.
- Containment enforced **server-side**: resolve the requested relative path, `realpath` it, reject anything (incl. `..` / symlink escapes) that resolves outside the scratchpad root.
- Not a real new trust surface — the operator can already browse arbitrary dirs via the Settings workspace picker (`/api/fs/dirs`). The genuinely new power is **streaming file bytes to the browser**, kept scoped to scratchpad output.

### Lifecycle
- **Live sessions only.** The scratchpad is deleted on archive (`removeWorktreeScratch(worktreePath)` in `src/tmp-sweep.ts`). Post-archive retention is explicitly **out of scope** (separate, heavier feature).

### Server API (session-scoped, relative paths)
- `GET /api/sessions/:id/scratchpad?path=<relative>` → directory listing (entries with name/type, breadcrumb/parent within root).
- `GET /api/sessions/:id/scratchpad/download?path=<relative>` → file download.
  - Returns `Bun.file(resolved)` with `Content-Disposition: attachment; filename=…`.
  - Operator HttpOnly auth cookie rides the same-origin GET automatically → a plain anchor download works.
- Both behind the existing `checkAuth` gate.
- **Path source of truth:** compute the scratchpad root by **reusing the existing `src/tmp-sweep.ts` derivation** (share `claudeTmpRoot()` + `dashify(worktreePath)`); do not re-implement the formula. Path is computed from `session.worktreePath`, not stored.

### UI
- New **"Files" tab** in `Viewport.svelte` / `ViewportTabBar.svelte`, alongside `term / todo / activity / diff / preview`.
- **Tab hides when the scratchpad has zero files** (mirrors how `todo` hides without a `TODO.md`).
  - Visibility driven by a cheap **`hasScratchpadFiles` boolean folded into the existing per-session status the UI already polls** — a **shallow** (non-recursive) "dir non-empty" check server-side. No second polling loop.
- Inside the tab: breadcrumb + clickable directory rows + per-file download affordance. Reuse the `SettingsWorkspacePanel.svelte` navigation pattern and `DiffFileBlock`-style row look.

### Behavior decisions
- **Read-only** — list + download only. No upload / delete / rename / mkdir.
- **Click a file = download immediately.** No inline preview (text-peek is a possible follow-up, must not gate download).
- **Single file at a time.** No folder-zip / multi-select.
- **Show dotfiles** (it's the operator's own scratchpad).
- **Sort:** directories first, then alphabetical.
- **No download size cap** — streaming via `Bun.file` handles large files.

## Project-rule compliance (in the implementing PR)
- **i18n:** all new strings get EN+DE keys in `ui/messages/{en,de}.json` (`check:i18n` gate).
- **Feature catalog:** add one `feature-announcements.ts` entry (user-facing UI). `targetId` on the Files tab anchor for a coachmark is optional.
- **Design system:** use semantic tokens + existing recipes; no raw hex / ad-hoc px.
- No new glossary term required.

## Explicit non-goals (deferred follow-ups)
- Post-archive scratchpad retention.
- Upload-into-scratchpad (hand a file *to* the agent).
- Inline file preview / syntax highlighting.
- Folder-as-zip / multi-file download.

## Prior art / key files
- `src/tmp-sweep.ts` — scratchpad path derivation + `removeWorktreeScratch`.
- `src/server.ts` — route handler pattern, `checkAuth`, `Bun.file` static serving, existing `GET /api/fs/dirs`.
- `ui/src/lib/components/Viewport.svelte` + `viewport/ViewportTabBar.svelte` — tab system.
- `ui/src/lib/components/settings/SettingsWorkspacePanel.svelte` — dir-navigation prior art (`listDirs`).
- `ui/src/lib/components/DiffPanel.svelte` / `DiffFileBlock.svelte` — file-row display prior art.
- `ui/src/lib/api.ts` — API client patterns.

scoop:
Closes #1164.

A **read-only file browser** in the session viewport, rooted at the session's **scratchpad**, that descends into subdirectories and **downloads individual files**. Live sessions only.

## What ships
- **New "Files" tab** on a live session's viewport, shown only when the scratchpad holds files (mirrors how the To-Do tab hides). Click a directory to descend (breadcrumb navigation), click a file to download it immediately. Dotfiles shown; directories first, then alphabetical.
- **Server API** (both behind `checkAuth`):
  - `GET /api/sessions/:id/scratchpad?path=<relative>` → directory listing.
  - `GET /api/sessions/:id/scratchpad/download?path=<relative>` → file stream via `Bun.file` with a safe `Content-Disposition: attachment`.
  - Registered in the `handleSessions` sub-handler array, like every other `/api/sessions/:id/...` route.

## Security boundary
Containment is enforced server-side in `src/scratchpad.ts`: every request realpaths the root **and** the resolved target and rejects anything outside it (`..`, absolute paths, symlink escapes). The listing additionally drops per-entry symlinks that escape the root (mirrors `dirs.ts`), so an out-pointing link never renders as a clickable row. Download requires a regular file. The `Content-Disposition` filename is RFC 5987-encoded with a CR/LF/quote-stripped ASCII fallback (no thrown invalid-header / header injection). Archived sessions 404 (live only).

## Tab visibility — no second poll
`hasScratchpadFiles` is computed server-side (transient, derived — **not** a DB column) and folded into the `GET /api/sessions` payload + `session:new` + the existing **`session:status` (idle/done)** WebSocket push — i.e. it refreshes at turn-end, exactly when the agent has finished writing files, with **no new polling loop**. The UI merges it via a guarded patch that won't clobber the flag on a status-only (`→ running`) push. This was the grilled design decision (operator-confirmed) reconciling the issue's "no second poll" with its "mirror the todo tab" cue.

## ⚠️ Deviation from the issue (flagged per house rules)
The issue said the root should **reuse `worktreeScratchDir(worktreePath)`**. An on-disk spike showed that is the **wrong base**: `worktreeScratchDir` is the *nested sub-agent* doubled-`claude-$uid` base (those dirs hold `<uuid>/tasks`, never a `scratchpad`). A session's own artifacts live under the **single** base:

```
<claudeTmpRoot>/<dashify(worktreePath)>/<claudeSessionId>/scratchpad/
```

verified against this session's own scratchpad + 150+ live dirs, and against the running server's env (no `TMPDIR` override → `claudeTmpRoot()` matches the agent's). New helper `sessionScratchpadDir()` reuses the existing `claudeTmpRoot()` + `dashify()` primitives but drops the nested segment and appends the session-scoped tail. Flagged in the code comment on `sessionScratchpadDir` and the commit body.

**Lifecycle note:** `removeWorktreeScratch` deletes the nested base, not this single-base session scratchpad — so physical deletion on archive isn't guaranteed. "Live only" is enforced by us (tab gated on active session; routes 404 on archived), independent of dir lifetime. Post-archive retention stays out of scope.

## Project-rule compliance
- **i18n:** EN+DE keys added (`check:i18n` parity green — 2046 keys each).
- **Feature catalog:** one `feature-announcements.ts` entry (`sinceVersion 1.38.0`), `targetId: "files-tab"` coachmark anchor.
- **Design system:** semantic tokens only, reuses the SettingsWorkspacePanel / DiffFileBlock row recipes.
- No new glossary term.

## Tests / verification
- New `test/scratchpad.test.ts` (containment & `..`/absolute/symlink escapes, per-entry symlink drop, dirs-first + dotfile listing, `scratchpadHasFiles`, RFC 5987 disposition) and `test/scratchpad-endpoint.test.ts` (route auth/404/archived/download header/list enrichment).
- Green: root `bun run lint` + `bun test ./test` (5037 pass), `bunx tsc --noEmit`; UI `bun run check` + `bun run test` (2099 pass) + `check:i18n`; `prettier --check`, `bun run build`, `bunx fallow@2.100.0 audit` (0 issues in changed files). Pre-push hook passed.

### DIFF
--- a/docs/external-task-api.md
+++ b/docs/external-task-api.md
@@ -123,5 +123,10 @@ Once a task exists, an external agent can also drive it:
 - `DELETE /api/sessions/:id` — archive (end) the session.
 - `GET /api/sessions` — list active sessions; `GET /api/sessions/:id/diff`,
   `/activity`, `/usage` for inspection.
+- `GET /api/sessions/:id/scratchpad[?path=]` — read-only browse of a live
+  session's own scratchpad subtree; `GET /api/sessions/:id/scratchpad/download?path=`
+  streams a single file. Paths are relative to the scratchpad root and
+  realpath-contained to it (`..`, absolute, and symlink escapes are rejected);
+  both `404` on a missing/archived session.
 
 All of these honor the same auth/origin rules as task creation.
--- a/src/index.ts
+++ b/src/index.ts
@@ -83,7 +83,13 @@ import { recommendPrompt } from "./prompt-recommend";
 import { CountsService } from "./backlog";
 import { BacklogPoller } from "./backlog-poller";
 import { ProcessReaper, reapDeletedWorktreeOrphans } from "./process-reaper";
-import { sweepClaudeTmp, compileCacheDir, reapFallowCaches, pruneRepoWorktrees } from "./tmp-sweep";
+import {
+  sweepClaudeTmp,
+  compileCacheDir,
+  reapFallowCaches,
+  pruneRepoWorktrees,
+  scratchpadHasFiles,
+} from "./tmp-sweep";
 import { runSessionUsageBackfill } from "./usage-backfill";
 import { PreviewService } from "./preview";
 import { listRepos, listReposPathForReal } from "./repos";
@@ -490,7 +496,25 @@ const hookIngest = new HookIngest();
 const poller = new StatusPoller(
   store,
   herdr,
-  (id, status) => events.emit("session:status", { id, status }),
+  (id, status) => {
+    // On turn-end transitions (idle/done — when the agent has finished writing files), fold a
+    // fresh scratchpad-non-empty flag into the status push so the UI's Files tab appears/hides
+    // without a second poll (#1164). Computed only here, not on every running↔idle flap, to keep
+    // a readdir off the hot poll path. Other transitions emit the bare {id, status}; the UI merge
+    // guards against undefined so a status-only push never clobbers the live flag.
+    if (status === "idle" || status === "done") {
+      const s = store.get(id);
+      if (s) {
+        void scratchpadHasFiles(s.worktreePath, s.claudeSessionId)
+          .then((hasScratchpadFiles) =>
+            events.emit("session:status", { id, status, hasScratchpadFiles }),
+          )
+          .catch(() => events.emit("session:status", { id, status }));
+        return;
+      }
+    }
+    events.emit("session:status", { id, status });
+  },
   (id, block) => events.emit("session:block", { id, block }),
   undefined, // intervalMs
   undefined, // reclassifyMs
@@ -1191,7 +1215,10 @@ const drain = new DrainService({
   dropPrCache: (id) => prPoller.drop(id),
   emitEpic: (epic) => events.emit("epic:update", epic),
   emitEpicCompleted: (e) => events.emit("epic:completed", e),
-  emitSessionNew: (s) => events.emit("session:new", s),
+  // A brand-new session's agent is only just starting — its scratchpad can't hold artifacts
+  // yet, so seed hasScratchpadFiles=false (#1164). The live truth thereafter rides the
+  // session:status (idle/done) push and the /api/sessions list enrichment.
+  emitSessionNew: (s) => events.emit("session:new", { ...s, hasScratchpadFiles: false }),
   // #1071: wire rebase cap + real rebaseLandingBranch (mirrors autopilot/automerge wiring).
   rebaseCap: config.autoMergeRebaseCap,
 });
--- a/src/scratchpad.ts
+++ b/src/scratchpad.ts
@@ -1,0 +1,159 @@
+import { promises as fsp } from "node:fs";
+import { join, relative, dirname, sep, normalize, isAbsolute } from "node:path";
+import { sessionScratchpadDir } from "./tmp-sweep";
+
+/**
+ * Read-only browser of a session's scratchpad subtree (#1164). Every path the operator supplies
+ * is treated as RELATIVE to the scratchpad root and is realpath-contained to it: `..`, absolute
+ * paths, and symlink escapes all resolve to null/dropped. The genuinely new power vs. the
+ * existing `/api/fs/dirs` browser is streaming file BYTES, so containment is the trust boundary.
+ */
+
+export interface ScratchEntry {
+  name: string;
+  type: "file" | "dir";
+  /** Path relative to the scratchpad root, forward-slash separated, no leading slash. */
+  path: string;
+}
+
+export interface ScratchListing {
+  /** The directory being listed, relative to the root ("" = root). */
+  path: string;
+  /** Parent dir relative to the root, or null at the root. */
+  parent: string | null;
+  entries: ScratchEntry[];
+}
+
+/** true when `real` is the root itself or lives inside it (realpath containment). */
+function within(real: string, rootReal: string): boolean {
+  return real === rootReal || real.startsWith(rootReal + sep);
+}
+
+/** Root-relative, forward-slash form of an absolute path known to be within `rootReal`. */
+function relFromRoot(rootReal: string, abs: string): string {
+  if (abs === rootReal) return "";
+  return relative(rootReal, abs).split(sep).join("/");
+}
+
+/**
+ * Resolve a caller-supplied RELATIVE path against the session scratchpad root, enforcing
+ * containment via realpath. Returns the canonical `{ rootReal, resolved }`, or null when:
+ * `claudeSessionId` is blank, the root is absent, the request escapes the root (`..`, absolute,
+ * symlink), or the target doesn't exist.
+ */
+export async function resolveScratchpadPath(
+  worktreePath: string,
+  claudeSessionId: string,
+  relPath: string,
+): Promise<{ rootReal: string; resolved: string } | null> {
+  if (!claudeSessionId) return null;
+  let rootReal: string;
+  try {
+    rootReal = await fsp.realpath(sessionScratchpadDir(worktreePath, claudeSessionId));
+  } catch {
+    return null; // root not created yet (agent wrote nothing) / unreadable
+  }
+  // Treat as root-relative. Reject any normalized path that climbs above the root (`..`) or is
+  // absolute; realpath containment below is the backstop (catches symlink escapes too).
+  const norm = normalize(relPath || "");
+  if (norm === ".." || norm.startsWith(".." + sep) || isAbsolute(norm)) return null;
+  const target = norm === "" || norm === "." ? rootReal : join(rootReal, norm);
+  let resolved: string;
+  try {
+    resolved = await fsp.realpath(target);
+  } catch {
+    return null;
+  }
+  if (!within(resolved, rootReal)) return null;
+  return { rootReal, resolved };
+}
+
+/**
+ * List one directory of the scratchpad subtree. Returns null when the path can't be resolved
+ * (escape / missing) or isn't a directory. Entries include files AND subdirs AND dotfiles
+ * (it's the operator's own scratchpad); per-entry realpath filtering drops anything that points
+ * outside the root (e.g. an out-pointing symlink) so it never renders as a clickable row.
+ * Sort: directories first, then locale-aware alphabetical within each group.
+ */
+export async function listScratchpad(
+  worktreePath: string,
+  claudeSessionId: string,
+  relPath: string,
+): Promise<ScratchListing | null> {
+  const r = await resolveScratchpadPath(worktreePath, claudeSessionId, relPath);
+  if (!r) return null;
+  const { rootReal, resolved } = r;
+
+  let dirents;
+  try {
+    dirents = await fsp.readdir(resolved, { withFileTypes: true });
+  } catch {
+    return null; // not a directory / unreadable
+  }
+
+  const entries: ScratchEntry[] = [];
+  for (const d of dirents) {
+    const abs = join(resolved, d.name);
+    let real: string;
+    try {
+      real = await fsp.realpath(abs);
+    } catch {
+      continue; // broken symlink / unreadable → skip
+    }
+    if (!within(real, rootReal)) continue; // never surface an entry that escapes the root
+    let isDir: boolean;
+    try {
+      isDir = (await fsp.stat(abs)).isDirectory();
+    } catch {
+      continue;
+    }
+    entries.push({ name: d.name, type: isDir ? "dir" : "file", path: relFromRoot(rootReal, abs) });
+  }
+
+  entries.sort((a, b) =>
+    a.type !== b.type ? (a.type === "dir" ? -1 : 1) : a.name.localeCompare(b.name),
+  );
+
+  const here = relFromRoot(rootReal, resolved);
+  return {
+    path: here,
+    parent: here === "" ? null : relFromRoot(rootReal, dirname(resolved)),
+    entries,
+  };
+}
+
+/**
+ * Resolve a scratchpad path that must be a regular FILE (for download). Returns the canonical
+ * absolute path, or null on escape / missing / not-a-regular-file.
+ */
+export async function resolveScratchpadFile(
+  worktreePath: string,
+  claudeSessionId: string,
+  relPath: string,
+): Promise<string | null> {
+  const r = await resolveScratchpadPath(worktreePath, claudeSessionId, relPath);
+  if (!r) return null;
+  try {
+    if (!(await fsp.stat(r.resolved)).isFile()) return null;
+  } catch {
+    return null;
+  }
+  return r.resolved;
+}
+
+/**
+ * Build a safe `Content-Disposition` header for an attachment download. Emits both an ASCII
+ * `filename="…"` (control chars / quote / backslash replaced) and an RFC 5987
+ * `filename*=UTF-8''…` so non-ASCII names survive without throwing an invalid-header (500) or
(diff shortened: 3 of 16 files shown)


scoop (commented):
## Summary

Implements the read-only scratchpad file browser per #1164: a server-side `src/scratchpad.ts` with realpath containment, two session-scoped GET routes (list + download), a `hasScratchpadFiles` flag folded into the existing session payload/status push, and a new `FilesPanel` + gated `Files` tab. The implementation closely follows the agreed design and is, on the whole, correct and well-guarded.

### What I verified (with ground truth)

- **Containment / security boundary.** `resolveScratchpadPath` (`src/scratchpad.ts:44`) normalizes the relative path, rejects `..`-climbs and absolute paths, then `realpath`s the target and checks `within(resolved, rootReal)` (`src/scratchpad.ts:30`). `listScratchpad` additionally realpath-filters every entry (`src/scratchpad.ts:96-110`) so an out-pointing symlink never renders. `resolveScratchpadFile` re-checks `isFile()` (`src/scratchpad.ts:135`). The `within` check uses `rootReal + sep` so a sibling like `<root>-evil` can't prefix-match. Solid.
- **Auth gate.** Both routes flow through `makeApp` → `checkAuth` before `ROUTE_HANDLERS` (`src/server.ts:5126-5141`); the download is a same-origin GET so the HttpOnly cookie rides it, matching the design.
- **Routing fall-through.** For `/api/sessions/:id/scratchpad[/download]`, `handleSessionReads` returns null (parts[3] is `scratchpad`, not usage/activity/diff/leftovers, and `!parts[3]` is false — `src/server.ts:1639-1645`), so it correctly reaches `handleSessionScratchpad` (`src/server.ts:1653`). Archived/missing sessions 404 (`src/server.ts:1656`).
- **Flag plumbing.** `sessionRead` is now async and awaited by its only caller (`src/server.ts:1644`); `withScratchpadFlags` enriches the list (`src/server.ts:1597`); the StatusPoller recomputes only on idle/done (`src/index.ts:500-516`); `session:new` seeds `false` (`src/index.ts:1218`); the store's `applyStatus` undefined-guards the merge so a status-only push can't clobber the live flag (`ui/src/lib/store.svelte.ts:287-292`). `WsEvent` and `Session` types updated to match (`ui/src/lib/types.ts:775,1116`).
- **i18n parity.** All 8 new keys present in BOTH `ui/messages/en.json` and `ui/messages/de.json` (verified 8/8 each). The `files_download_aria({ name })` param call matches paraglide `{name}` syntax used elsewhere (e.g. `ui/messages/en.json:29` `unit_open_aria`, used as `m.automation_info_aria({ name })`). `m.common_loading()` exists in both catalogs.
- **Feature catalog.** One entry added with `targetId: "files-tab"` paired with `use:coachTarget={"files-tab"}` on the tab anchor (`ui/src/lib/components/viewport/ViewportTabBar.svelte:241`); `sinceVersion: "1.38.0"` is correct against the current `package.json` 1.37.0 / latest `v1.37.0` tag.
- **Header safety.** `attachmentDisposition` (`src/scratchpad.ts:148`) replaces control chars/quote/backslash in the ASCII fallback and percent-encodes the UTF-8 form, so CR/LF cannot inject a header.

### Nit (non-blocking)

- `attachmentDisposition`'s RFC 5987 `filename*=UTF-8''` value uses `encodeURIComponent`, which leaves `'`, `(`, `)`, `*` unescaped — these are not RFC 5987 `attr-char`s, so a filename containing them yields a technically non-conformant ext-value. Browsers parse it in practice and the ASCII fallback covers the common case, so this is cosmetic, not a security or functional issue.

Could not verify (ground truth not in this repo): the empirical claim in `src/tmp-sweep.ts:73-86` that a session's own scratchpad lives at `<claudeTmpRoot>/<dashified>/<sessionId>/scratchpad` (vs. the `worktreeScratchDir` base) — this depends on the live Claude Code tmp geometry, which isn't reproducible from the repo. Assumed correct as written/documented.

- If the new `sessionScratchpadDir` derivation is the real on-disk location, note that `removeWorktreeScratch` (`src/tmp-sweep.ts:325`) deletes the *different* `worktreeScratchDir` base — so it would not reclaim this scratchpad on archive. That cleanup path is untouched by this PR and post-archive retention was explicitly deferred by the issue; the browser endpoint itself 404s on archived sessions, so there is no exposure, only a potential disk-retention gap to track separately.

scoop (commented):
## Summary

The PR implements the read-only scratchpad file browser from #1164 cleanly and the implementation is correct on the paths I verified. Server-side containment, the gating flag, the new routes, the UI tab, and i18n parity all check out. The only carry-over is the previously-noted non-blocking RFC 5987 header nit, which the new diff does not address.

## Verification

**Containment / security** — `resolveScratchpadPath` (src/scratchpad.ts:42-65) realpaths the root, rejects `..`/absolute via `normalize` (src/scratchpad.ts:60), then realpaths the target and applies `within` containment (src/scratchpad.ts:24-26, 64). `listScratchpad` additionally realpath-filters each entry (src/scratchpad.ts:97-101) and `resolveScratchpadFile` enforces `isFile` (src/scratchpad.ts:138-145). Symlink/`..`/absolute escapes are all blocked. Routes are behind `checkAuth` (verified makeApp gate at src/server.ts:5126-5128, dispatched via ROUTE_HANDLERS).

**Route ordering** — `handleSessionReads` returns null for `parts[3]==="scratchpad"` (none of its branches match, src/server.ts:1639-1645), falling through to `handleSessionScratchpad` which is registered immediately after it (src/server.ts:2450-2451). Both 404 on missing/archived sessions (src/server.ts:1656). Verified.

**Path derivation** — `sessionScratchpadDir` = `<claudeTmpRoot>/<dashify(worktree)>/<sessionId>/scratchpad` (src/tmp-sweep.ts:74). This base (single dashified segment, no `claude-$uid`) is consistent with the live-tree layout asserted in the existing sweep test (test/tmp-sweep.test.ts:129, `<root>/-home-...-review-0a928a8b`). `claudeSessionId` is `""` for pre-feature sessions (src/types.ts:22) and both `scratchpadHasFiles` (src/tmp-sweep.ts:99) and `resolveScratchpadPath` (src/scratchpad.ts:51) short-circuit to false/null on blank.

**Flag plumbing** — emitter `data` is typed `unknown` (src/events.ts:9) so the extra `hasScratchpadFiles` field is accepted; the idle/done-only recompute (src/index.ts:507-515), `session:new` seed=false (src/index.ts:1218), list enrichment (src/server.ts:1635), and single-session enrichment (src/server.ts:1595) all align with the UI merge guard that skips `undefined` (ui/src/lib/store.svelte.ts:288-292, Viewport `$derived` at Viewport.svelte:514).

**i18n parity** — all 8 new keys (`viewport_files_tab`, `files_root_crumb`, `files_breadcrumb_aria`, `files_empty`, `files_load_error`, `files_download_aria`, `feat_scratchpad_files_title/body`) exist in BOTH ui/messages/en.json and ui/messages/de.json with matching `{name}` placeholder. `common_loading` (reused) present in both. Verified.

**Design tokens** — every token used in FilesPanel.svelte (`--color-inset/faint/red/line/panel/muted/ink-bright`, `--fs-meta`, `--font-mono`) is defined in ui/src/app.css (lines 9-82). `coachTarget` action imported and `targetId: "files-tab"` matches `use:coachTarget={"files-tab"}` (ViewportTabBar.svelte:241, feature-announcements.ts targetId). Feature-catalog gate only requires an entry to exist (test/check-feature-catalog.test.ts) — satisfied.

## Re-review of prior point

1. **src/scratchpad.ts `attachmentDisposition` RFC 5987 nit** — NOT addressed. src/scratchpad.ts:155 still builds `filename*=UTF-8''${encodeURIComponent(name)}`, which leaves `'`,`(`,`)`,`*` unescaped. Re-raised below per instructions (non-blocking; browsers tolerate it and the ASCII fallback covers the common case).

- The comment at src/tmp-sweep.ts:73-83 claims the scratchpad base was "verified empirically against the live tmp tree." I confirmed it is internally consistent with the layout encoded in test/tmp-sweep.test.ts:129, but the actual live agent on-disk path is not reproducible from this repo alone; assumed as written.

- src/tmp-sweep.ts `removeWorktreeScratch` (unchanged) uses `worktreeScratchDir` (`<root>/claude-$uid/<dashify>`), a different base than the new `sessionScratchpadDir` (`<root>/<dashify>/<sid>/scratchpad`), so archival does not actually delete the session's own scratchpad — those dirs accumulate on tmpfs. Post-archive retention/cleanup is explicitly deferred by #1164, so this is informational only.

scoop:
Finding 1 (RFC 5987 `filename*` escaping of `'()*`) is **already resolved** at commit cc648b86, which is the current branch HEAD — the review appears to have run against the prior snapshot (16a684a6).

`attachmentDisposition` now post-processes the `encodeURIComponent` output with `.replace(/['()*]/g, c => '%'+c.charCodeAt(0).toString(16).toUpperCase())`. Those four are exactly the characters `encodeURIComponent` leaves raw that are **not** valid RFC 5987 `attr-char`s; the others it leaves (`! ~ - _ .`) are all valid attr-chars, so no further escaping is needed. Covered by the test `attachmentDisposition: percent-encodes RFC 5987 non-attr-chars ('()*) in filename*` asserting `a'b(c)*d.txt` → `a%27b%28c%29%2Ad.txt`. No code change in this round.

scoop (commented):
## Re-review

The one prior finding is resolved and I found no new blocking issues.

### Prior finding — RFC 5987 `filename*` escaping of `'()*` — RESOLVED
`attachmentDisposition` (`src/scratchpad.ts:158-164`) now post-processes the `encodeURIComponent` output with `.replace(/['()*]/g, c => \`%${c.charCodeAt(0).toString(16).toUpperCase()}\`)`. Those four are precisely the characters `encodeURIComponent` leaves raw that are not valid RFC 5987 `attr-char`s; the rest it leaves (`! ~ - _ .`) are valid attr-chars. This matches the author's note and is covered by the percent-encoding test. Accepted, not re-raised.

### Verification performed
- **Containment / path traversal** — `within()` (`src/scratchpad.ts:31-33`) uses `real === rootReal || real.startsWith(rootReal + sep)`; the trailing `sep` correctly defeats sibling-prefix escapes (`/a/scratchpad-foo` vs root `/a/scratchpad`). `resolveScratchpadPath` (`src/scratchpad.ts:56-71`) rejects `..`/absolute pre-resolve and realpath-contains as the symlink backstop; `listScratchpad` per-entry realpaths and drops out-pointing entries (`src/scratchpad.ts:96-104`). Sound.
- **Route dispatch** — verified `handleSessionReads` (`src/server.ts:1638-1651`) returns null for `parts[3]="scratchpad"` (no branch matches; `!parts[3]` is false), so it correctly falls through to `handleSessionScratchpad` (`src/server.ts:1656-1681`), registered at `src/server.ts:2456`. Missing/archived sessions 404 (`src/server.ts:1662`).
- **i18n parity** — every new key exists in BOTH catalogs: `viewport_files_tab`, `files_root_crumb`, `files_breadcrumb_aria`, `files_empty`, `files_load_error`, `files_download_aria`, `feat_scratchpad_files_title`, `feat_scratchpad_files_body` (verified against `ui/messages/en.json:2048-2056` and `ui/messages/de.json:2048-2056`). The parameterized `files_download_aria` uses `{name}` in both. `common_loading` used by FilesPanel pre-exists in both catalogs.
- **Symbols/contracts** — `claudeSessionId` confirmed on Session (`src/types.ts:22`); `coachTarget` imported (`ui/src/lib/components/viewport/ViewportTabBar.svelte:4`), `tabId`/`vpBodyId` wired (`ui/src/lib/components/Viewport.svelte:349-350,1859-1860`); `tab` state union extended to include `"files"` in both Viewport and ViewportTabBar. Store `applyStatus` (`ui/src/lib/store.svelte.ts:287-293`) only applies `hasScratchpadFiles` when defined, so a status-only (→running) push cannot clobber the live flag; `session:new` seeds `false` (`src/index.ts:1218-1221`); list and single-session reads enrich active sessions (`src/server.ts:1597-1617`).
- **Download safety** — `Content-Disposition: attachment` (`src/server.ts:1673`) forces download; same-origin, behind `checkAuth`; no inline render path. `basename(resolved realpath)` used for the filename.

### Could not verify (stated limitation, not a finding)
- The `sessionScratchpadDir` formula `<claudeTmpRoot>/<dashify(worktree)>/<claudeSessionId>/scratchpad` (`src/tmp-sweep.ts:84-86`) is asserted to match where Claude Code actually writes a session's scratchpad. The ground truth is the live tmp tree, which is not in this repo; the author documents empirical verification. Assumed as written. If the formula is wrong, the Files tab would simply never appear (fail-closed), not mis-serve files.

- `removeWorktreeScratch` (`src/tmp-sweep.ts:325-336`, untouched) deletes `worktreeScratchDir` = `<claudeTmpRoot>/claude-$uid/<dashify(worktree)>`, a DIFFERENT base than the new `sessionScratchpadDir` = `<claudeTmpRoot>/<dashify(worktree)>/<sessionId>/scratchpad`; per the PR's own derivation, archive does NOT delete the session scratchpad, so the issue's "deleted on archive" lifecycle premise does not hold and these dirs persist on disk. No security impact on this PR — the browse/download endpoints 404 archived sessions (`src/server.ts:1662`), so retention is defense-in-depth only. Pre-existing disk-retention behavior independent of this change.

PR #1170, opened 2026-06-27
feat(orchestration): one ordered 'what happens next' seam for the autonomous engine (#1094)

### LINKED ISSUE #1094: Deepen: a single 'what happens next' decision seam for the autonomous engine
**Architecture review finding · `Worth exploring`** — deepening opportunity surfaced by an architectural review (deletion-test framed).

## Problem
The autonomous engine has no single "what happens next" interface. One poller event fans out to **two uncoordinated handlers**: `onBlock → autopilot.consider` and `onChange → drain.pump` (wired in `src/index.ts`). Both re-derive session state — `autopilot.eligible()` and `drain.buildState()` each reassemble it from `store` + `prCache` + review — and they **race within a single tick** (a session can be retired by drain the same tick autopilot steers it). Understanding "what happens when a session goes idle" means bouncing `poller.ts` → `autopilot.ts` → `drain.ts`.

## Deepening
A unified `SessionSnapshot` (git + review + activity + status) plus an ordered `SessionStateChange` routed behind one decision seam, modelled on the existing pure `computeNext`. Consumers subscribe in a known order (drain → autopilot).

**Keep the good seams.** `drain-core`, `stall`, `activity-signal` are already deep (pure, isolated, well-tested). This is about the **wiring above them**, not those cores.

## Deletion test
The coordination is load-bearing — delete it and the ordering reappears in the orchestrator. Large surface; deepen the wiring, not the cores.

## Wins
- locality: one decision place
- interface: subscribe to one event
- removes double state-derivation
- kills the retire-vs-steer tick race

_Vocabulary per `/codebase-design`: module · interface · seam · locality._

scoop:
## What & why

Closes #1094. The autonomous engine had no single "what happens next" place: one poller event fanned out to **two uncoordinated, fire-and-forget subscriptions** — `drain` (retire) and `autopilot` (steer). `EventHub.emit` dispatches synchronously but each subscriber ran `void asyncFn()`, so subscription order fixed *dispatch* but never *completion* order. Within one poller tick they interleaved and **raced**: a session could be retired by drain mid-classify while autopilot was steering it, and vice-versa.

This routes `session:status` and `session:git` through **one ordered seam** that dispatches consumers **drain → autopilot, awaited in order**.

## How

- **`src/session-snapshot.ts`** — `SessionSnapshot { id, repoPath, session }`, the `SessionStateChange` union, and pure `buildSnapshot` (returns `null` for an unknown session → dispatch skipped). Right-sized to what entry consumers actually read.
- **`src/session-router.ts`** — `SessionRouter` builds the snapshot **once** and dispatches `[drain, autopilot]`, **awaiting each fully before the next**; per-consumer try/catch isolation; an `onStatusSettled` hook for the independent plan-gate side-effect.
- **`drain.handle()` / `autopilot.handle()`** — each becomes a `SessionConsumer`. `drain` sources the trigger row from the snapshot (it runs first; its retire logic still re-derives live state via `buildState`/`computeNext`). `autopilot` deliberately **re-reads live state** — its entry + post-classify `eligible()` rechecks are the correctness guard and must see post-drain state.
- **`src/index.ts`** — hard cutover: status/git route only through the router; `session:block` stays autopilot-only; `session:archived`/`session:review` stay drain-only; **autoMerge and every other subscriber untouched**.

## What the ordering buys (precise)

Autopilot's post-classify `eligible()` recheck already prevented double-acting. Ordering adds: **determinism**, **skipping a wasted classify** (drain-first lets autopilot's *entry* gate return early instead of classifying then discarding), and **closing the steer-then-archive window** (drain acts on the live row before autopilot flips it).

## Deliberate deviations (flagged per house rules)

- **PR-prewarm hoist dropped.** The plan (responding to the plan reviewer) specced an `autopilot.prewarmPr` kick hoisted ahead of the awaited drain, with a "prewarm-before-drain" success criterion. During implementation this proved **both ineffective and unnecessary**: `onDone` already kicks `refreshPr` immediately before its classify, and `pollSession` debounces — a second prewarm kick would just re-delay the poll (double-kick). Because the refreshPr↔classify overlap is *intrinsic to `onDone`*, drain-first shifts both together and the overlap is preserved; only negligible steady-state latency is added. This satisfies the reviewer's "confirm the timing still holds" branch. The hook was removed from the router; rationale is in `autopilot.ts` (`handle`) and `session-router.ts` (`onStatus`). The "prewarm test" success criterion is intentionally not met.
- **autoMerge left out of the seam.** The issue names only the drain↔autopilot race. autoMerge stays on its own independent subscription — the router serializes drain only relative to autopilot, never autoMerge, so drain↔autoMerge concurrency is byte-for-byte unchanged and already idempotent-safe (`store.archive` is an idempotent `UPDATE`). No new double-archive window.
- **Snapshot trimmed** to `{ id, repoPath, session }` — the issue's literal "git+review+activity+status" field list was speculative; review/activity are read inside the kept cores, not at the seam.

## Scope kept

The already-deep cores — `drain-core`/`computeNext`, `stall`, `activity-signal` — are **unmodified**. This deepens the wiring above them.

## Tests / verification

- New `test/session-router.test.ts` (ordering proven deterministically via a manually-controlled deferred: autopilot cannot start until drain's promise settles), single-build, unknown-session, failure-isolation, identity. New `test/session-snapshot.test.ts`.
- Full root suite **5029 pass / 0 fail**; `bunx tsc --noEmit` clean; `bun run lint` clean; `fallow audit` exit 0; i18n / glossary / feature-catalog gates pass.
- Server-only orchestration plumbing, no user-facing UX → `[no-feature-entry]`.

### DIFF
--- a/src/autopilot.ts
+++ b/src/autopilot.ts
@@ -347,8 +347,8 @@ export class AutopilotService {
    *  autopilot MUST re-read the live row (its eligible() rechecks at entry + post-classify
    *  are the correctness guard), so it deliberately does NOT consume snapshot.session.
    *  Ordering note: the seam runs drain before this. onDone kicks refreshPr immediately
-   *  before its classify (line ~358); drain-first shifts BOTH together, so the PR-poll↔classify
-   *  overlap is preserved — no separate up-front prewarm is needed (#1094). */
+   *  before its classify (see onDone below); drain-first shifts BOTH together, so the
+   *  PR-poll↔classify overlap is preserved — no separate up-front prewarm is needed (#1094). */
   async handle(change: SessionStateChange): Promise<void> {
     const { id } = change.snapshot;
     if (change.kind === "status") {
--- a/src/autopilot.ts
+++ b/src/autopilot.ts
@@ -2,6 +2,7 @@ import type { SessionStore } from "./store";
 import type { Session, AutopilotVerdict, ReviewVerdict } from "./types";
 import type { BlockReason } from "./blocked";
 import type { GitState } from "./forge/types";
+import type { SessionStateChange } from "./session-snapshot";
 import { effectiveAutopilot } from "./effective-autopilot";
 import { signedOff } from "./signoff";
 import { DRAFT_PR_NOTE } from "./service";
@@ -340,6 +341,22 @@ export class AutopilotService {
     const cur = this.eligible(id);
     if (!cur) return;
     await this.dispatch(cur, v);
+  }
+
+  /** SessionConsumer entry (#1094 seam). Delegates to the existing handlers by id —
+   *  autopilot MUST re-read the live row (its eligible() rechecks at entry + post-classify
+   *  are the correctness guard), so it deliberately does NOT consume snapshot.session.
+   *  Ordering note: the seam runs drain before this. onDone kicks refreshPr immediately
+   *  before its classify (line ~358); drain-first shifts BOTH together, so the PR-poll↔classify
+   *  overlap is preserved — no separate up-front prewarm is needed (#1094). */
+  async handle(change: SessionStateChange): Promise<void> {
+    const { id } = change.snapshot;
+    if (change.kind === "status") {
+      this.onStatus(id, change.status); // clears a pause on operator reply
+      if (change.status === "done") await this.onDone(id);
+      return;
+    }
+    this.onGit(id, change.git); // sync; PR-open handoff + red-CI recovery
   }
 
   /** session:block handler. Only steerable shapes are eligible; menu/stall surface as-is. */
--- a/src/drain.ts
+++ b/src/drain.ts
@@ -1,6 +1,7 @@
 import type { SessionStore } from "./store";
 import type { GitForge, GitState, Issue, PrStatus, SubIssueRef } from "./forge/types";
 import type { CreateSessionInput, Session } from "./types";
+import type { SessionStateChange } from "./session-snapshot";
 import type { UsageLimits } from "./usage-limits";
 import { settleMergedSession } from "./merge-teardown";
 import { isFullAuto } from "./full-auto";
@@ -1778,6 +1779,24 @@ export class DrainService {
   }
 
   // ── event handlers (public surface) ───────────────────────────────────────────
+
+  /** SessionConsumer entry (#1094 seam). Sources the trigger row from the shared snapshot
+   *  instead of a fresh store.get — drain runs FIRST in the ordered chain, so the snapshot
+   *  (built this tick) is current for its purposes. Mirrors onGit/onStatus exactly. */
+  async handle(change: SessionStateChange): Promise<void> {
+    const s = change.snapshot.session;
+    if (change.kind === "git") {
+      if (!s.auto) return; // drain only manages auto sessions
+      if (change.git.state === "merged") {
+        await this.reapMerged(s);
+        return;
+      }
+      await this.pumpIfEnabled(s.repoPath);
+      return;
+    }
+    // kind === "status"
+    await this.pumpIfEnabled(s.repoPath);
+  }
 
   /** pr-poller observed a new git state for a session. */
   async onGit(id: string, git: GitState): Promise<void> {
--- a/src/index.ts
+++ b/src/index.ts
@@ -74,6 +74,7 @@ import { createIssueLogger } from "./issue-log";
 import { PlanGateService } from "./plan-gate";
 import { AutopilotService } from "./autopilot";
 import { DrainService } from "./drain";
+import { SessionRouter, type SessionConsumer } from "./session-router";
 import { AutoMergeService } from "./automerge";
 import { DraftReconcileService } from "./draft-reconcile";
 import { isFullAuto } from "./full-auto";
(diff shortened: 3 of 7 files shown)


scoop (commented):
## Summary

This PR introduces a single ordered "what happens next" seam (`SessionRouter` + `SessionSnapshot`) for `session:status`/`session:git`, replacing the two uncoordinated fire-and-forget subscriptions (autopilot steer + drain retire) that previously raced within a poller tick. It satisfies issue #1094: one decision place, one subscription, a shared snapshot built once, and a strict drain→autopilot ordering that kills the retire-vs-steer race. The cores (`drain-core`, `stall`, `activity-signal`) are untouched, as the issue requested.

I verified the refactor preserves the prior semantics with the deliberate reordering:

- **Event routing is complete and non-overlapping.** Old autopilot subscription now handles only `session:block` (src/index.ts:1159-1162); old drain subscription only `session:archived`/`session:review` (src/index.ts:1187-1194); the new router handles `session:status`/`session:git` (src/index.ts:1221-1230). No event is dropped or double-dispatched. The many other independent `session:git` subscribers (automerge, merge-train, etc., src/index.ts:638,663,918+) are correctly left untouched — they were never part of the ordering.
- **drain.handle mirrors onGit/onStatus.** Verified against src/drain.ts:1786-1799 vs 1802-1812 and 1873-1893: git path keeps the `!s.auto` early-return before the merged check, and reapMerged/pumpIfEnabled selection is identical; status path pumps the repo unconditionally. The snapshot is built at tick start and drain runs first (nothing mutated yet), so `change.snapshot.session` equals what a fresh `store.get` would return at drain time.
- **autopilot.handle re-reads live state.** Verified against src/autopilot.ts:352-360: it deliberately ignores `snapshot.session` and re-derives via `eligible()`/`store.get` (src/autopilot.ts:201-205, 446-447, 507-517), so the stale-vs-live distinction is harmless — its eligible() recheck at entry and post-classify is the correctness guard.
- **Reordering is safe.** A planning-phase session is stood down by autopilot (`eligible()` returns null, src/autopilot.ts:204), so plan-gate keeps ownership; autopilot.onDone won't advance it out of `planning`, so the settled hook reading the point-in-time snapshot still sees `planning` correctly. For a ready PR, drain retires/archives before autopilot's onGit/onDone re-reads the archived row and stands down — exactly the intended fix.
- **Plan-gate side-effect preserved.** The `done`+`planPhase==="planning"` gate is retained in `onStatusSettled` (src/index.ts:1214-1218), runs after the consumer chain, and is reached regardless of a consumer throwing (per-consumer try/catch in src/session-router.ts:51-58). plan-gate.consider re-checks planPhase and reads only stable fields from the passed session (src/plan-gate.ts:217-237), so a point-in-time snapshot is fine.
- **buildSnapshot null-guard is correct.** `store.get` returns `Session | null` (never undefined, src/store.ts:1738-1743), so `session === null` (src/session-snapshot.ts:38) is the right check.
- The router tests (test/session-router.test.ts) prove the load-bearing property — autopilot starts only after drain fully resolves — deterministically via a deferred gate, plus single-build, failure-isolation, and snapshot-identity coverage.

Two minor, non-blocking nits below.

## Findings

1. **src/session-snapshot.ts:** `SessionSnapshot.repoPath` is dead — no consumer reads it (drain reads `snapshot.session.repoPath` at src/drain.ts:1787/1794; autopilot reads `snapshot.id` at src/autopilot.ts:353; plan-gate reads `snapshot.session` at src/index.ts:1215). Only `id` and `session` are used. Consider dropping the redundant `repoPath` field to keep the snapshot truly minimal as its own doc-comment promises.
2. **src/autopilot.ts:** the `handle()` doc comment cites "line ~358" for the refreshPr kick, but `refreshPr` is actually at src/autopilot.ts:375. Stale line reference within the same change.

- src/index.ts: `onStatusSettled` now runs plan-gate AFTER the awaited consumer chain (was concurrent with onDone before). Today this only adds latency (a planning session is a no-op for autopilot), but if autopilot's onDone path were ever extended to act on planning-phase sessions, the settled hook would fire plan-gate.consider on a session autopilot already advanced, using the point-in-time snapshot's `planPhase`.

scoop (commented):
## Summary

This PR introduces the `SessionRouter` + `SessionSnapshot` seam (#1094): `session:status`/`session:git` events now flow through one ordered router that builds the shared snapshot once and dispatches `drain` before `autopilot`, awaiting each. This kills the retire-vs-steer tick race and removes the double `store.get`. The change is well-scoped, well-documented, and well-tested.

Both previously-raised points are addressed in this diff:

1. **`SessionSnapshot.repoPath` dropped** — the interface is now `{ id; session }` (verified `src/session-snapshot.ts:10-13`). A tree grep confirms no remaining `snapshot.repoPath` references; drain reads `change.snapshot.session.repoPath`, autopilot uses `change.snapshot.id`, plan-gate reads `change.snapshot.session.planPhase`. Addressed.
2. **Stale `line ~358` reference in `autopilot.handle` doc** — the comment now reads "(see onDone below)" with no numeric line reference (verified `src/autopilot.ts:346-351`). Addressed.

## Verification of the implementation

- **`drain.handle` mirrors the originals.** Git branch matches `onGit` (`src/drain.ts:1801-1812`): `!s.auto` → return; `merged` → `reapMerged(s)`; else `pumpIfEnabled(s.repoPath)` — `s` is `change.snapshot.session`, guaranteed non-null because `buildSnapshot` returns null on a null row (`src/session-snapshot.ts:35`). Status branch matches `onStatus`→`pumpForSession` (`src/drain.ts:1873-1894`): pumps regardless of `auto`. Verified against `src/drain.ts:1790-1798`.
- **`autopilot.handle` correctly delegates by id and re-reads the live row.** It does NOT consume `snapshot.session`; instead `onStatus`/`onDone`/`onGit` each re-`store.get`. This is the correctness guard for the new ordering: drain may archive the row in the same tick, and every autopilot entry point stands down on a now-archived row — `eligible()` (`src/autopilot.ts:203`), `reEngageCi` (`src/autopilot.ts:509`), and `reEngageRebase` (`src/autopilot.ts:571`) all early-return on `s.status === "archived"`. So `drain.handle` retiring before `autopilot.handle` guarantees no steer on a retired session. Verified against `src/autopilot.ts:201-211, 507-520, 570-575`.
- **Plan-gate hook moved correctly.** The old inline `if (status==="done" && sess?.planPhase==="planning") planGate.consider` (removed from `src/index.ts`) is now `onStatusSettled` (`src/index.ts:1208-1212`), guarded by `change.status === "done"`, fired after the consumer chain and reached even if a consumer throws (router `safeSync`, `src/session-router.ts:55-62`). Equivalent behavior; reads the live row object via `change.snapshot.session`.
- **No coverage lost.** `session:block` remains an independent autopilot subscription (`src/index.ts:1159-1163`); `session:archived`/`session:review` remain independent drain subscriptions (`src/index.ts:1188-1195`); autoMerge and the other ~10 `session:git`/`status` subscribers remain independent (verified via grep of `src/index.ts`). Only drain+autopilot's git/status handling moved into the router.
- **`store.get` returns `Session | null`** (`src/store.ts:1738`), matching the `SnapshotAccessors.getSession` contract and the explicit `session === null` check in `buildSnapshot`.

The router/snapshot unit tests (`test/session-router.test.ts`, `test/session-snapshot.test.ts`) prove the ordering (deferred-gated, not timing-based), single-build-per-call, unknown-session skip, failure isolation, and snapshot identity. CI handles execution.

PR #1177, opened 2026-06-27
fix(plugins): follow symlinked plugin dirs in the loader (#1176)

### LINKED ISSUE #1176: Plugin loader silently skips symlinked plugin directories (support symlink installs)
## Summary

The server-side plugin loader silently **skips symlinked plugin directories**, so `ln -s … ~/.shepherd/plugins/<name>` installs never load. Plugins must currently be `cp -r`'d, which breaks the natural workflow of symlinking a plugin from its git checkout (so `git pull` keeps the installed plugin current).

## Root cause

`src/plugins/loader.ts:103-105` (HEAD `b2405ab2`):

```ts
const entries = await readdir(this.deps.pluginsDir, { withFileTypes: true });
names = entries
  .filter((e) => e.isDirectory())   // ← a symlink Dirent is isDirectory()===false
  .map((e) => e.name)
  .sort();
```

`readdir(..., { withFileTypes: true })` does **not** follow symlinks: a symlink-to-directory Dirent reports `isDirectory() === false` and `isSymbolicLink() === true`, so the entry is filtered out. The plugin is never scanned — no manifest read, no error, nothing in the status panel. It just silently doesn't exist.

## Reproduce

```sh
bun -e 'const {readdirSync}=require("fs");const {mkdtempSync,mkdirSync,symlinkSync}=require("fs");const os=require("os"),p=require("path");
const d=mkdtempSync(p.join(os.tmpdir(),"x"));mkdirSync(p.join(d,"real"));symlinkSync(p.join(d,"real"),p.join(d,"link"));
for(const e of readdirSync(d,{withFileTypes:true}))console.log(e.name,"isDirectory="+e.isDirectory(),"isSymbolicLink="+e.isSymbolicLink());'
# → link isDirectory=false isSymbolicLink=true   ← skipped by the loader
# → real isDirectory=true  isSymbolicLink=false
```

## Impact

- A symlink install (a very natural way to run a plugin straight from its repo so updates land on `git pull`) silently no-ops — and because the no-plugin state is itself valid, there's no diagnostic.
- Plugin docs/READMEs that suggest `ln -s` are effectively wrong; the shipped `examples/plugins/spawn-labeler` README uses `cp -r` specifically to dodge this, but the asymmetry isn't documented as a hard constraint.

## Proposed fix

Resolve symlinked entries before the `isDirectory()` filter — for any `e.isSymbolicLink()`, `stat()` the target (follows the link) and include it when the target is a directory. Roughly:

```ts
const entries = await readdir(this.deps.pluginsDir, { withFileTypes: true });
const names: string[] = [];
for (const e of entries) {
  if (e.isDirectory()) { names.push(e.name); continue; }
  if (e.isSymbolicLink()) {
    try {
      if ((await stat(join(this.deps.pluginsDir, e.name))).isDirectory()) names.push(e.name);
    } catch { /* dangling symlink — skip */ }
  }
}
names.sort();
```

`stat` is already imported. Dangling symlinks are skipped safely. Everything downstream (`loadOne` joins `pluginsDir/name` and reads through the link) already works once the name is included.

## Acceptance criteria

- A directory **symlinked** into `~/.shepherd/plugins/` loads identically to a copied one (manifest read, `register` called, appears in Settings → Plugins).
- A **dangling** symlink is ignored without throwing.
- The zero-plugin / missing-dir no-op invariant is preserved.
- A loader test covers the symlinked-plugin case (e.g. symlink `test/fixtures/example-plugin` into a temp plugins dir and assert it loads).

## Context

Surfaced while deploying the `claude-swap` account-rotation plugin (`erwins-enkel/shepherd-claude-swap`): a symlink install loaded nothing; switching to `cp -r` (via `git archive | tar -x`) fixed it. Supporting symlinks would make "run a plugin from its checkout" work as expected.

scoop:
## Problem

`PluginRegistry.loadAll` enumerated the plugins dir with `readdir(..., { withFileTypes: true })` and filtered on `e.isDirectory()`. That call **does not follow symlinks**: a symlink-to-directory Dirent reports `isDirectory() === false` (and `isSymbolicLink() === true`), so a `ln -s … ~/.shepherd/plugins/<name>` install was silently filtered out — no manifest read, no error, nothing in the status panel. Plugins had to be `cp -r`'d, which breaks running a plugin straight from its git checkout (so `git pull` keeps it current).

Surfaced deploying the `claude-swap` plugin: a symlink install loaded nothing; `cp -r` fixed it.

## Fix

`src/plugins/loader.ts` — resolve symlinked entries before the directory check: plain dirs pass as before; for a symlink entry, `stat()` it (which **follows** the link) and include it only when the target is a directory. A dangling/broken symlink throws from `stat()` and is skipped. `stat` was already imported; downstream `loadOne` already reads through the link, so nothing else changes. The whole-`readdir` try/catch (missing-dir → clean no-op) is untouched.

## Tests

`test/plugins.test.ts`:
- a **symlinked** plugin dir (real folder outside, symlink inside) loads with `ok` health, identically to a copy;
- a **dangling** symlink is ignored without throwing (loadedCount 0).

Existing missing/empty/zero-plugin no-op tests still pass.

## Docs

`docs/plugins.md` + `examples/plugins/README.md` note the `ln -s …` install now works (loader follows the link).

## Acceptance criteria

- [x] Symlinked dir loads identically to a copied one (manifest read, `register` called, in `list()`).
- [x] Dangling symlink ignored without throwing.
- [x] Zero-plugin / missing-dir / empty-dir no-op invariants preserved.
- [x] Loader test covers the symlinked-plugin case.

## Verification

`bun test ./test` → 5102 pass / 0 fail · `bun run lint` clean · `bunx tsc --noEmit` clean · `fallow audit` no issues.

Server-only fix (no UI / i18n / feature-catalog entry needed). Closes #1176.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -231,7 +231,9 @@ Settings → Plugins panel:
 env var name (`envVar`) and label template (`labelTemplate`); both default sensibly when
 absent.
 
-**Copy it to run it.** `cp -r examples/plugins/spawn-labeler ~/.shepherd/plugins/`, then —
+**Copy it to run it.** `cp -r examples/plugins/spawn-labeler ~/.shepherd/plugins/` (or
+`ln -s "$PWD/examples/plugins/spawn-labeler" ~/.shepherd/plugins/` to run it straight from a
+checkout — the loader follows symlinked plugin dirs, so `git pull` keeps it current), then —
 because the example's `import type` uses a repo-relative path that won't resolve out-of-repo
 — **drop the `import type` line or vendor `src/plugins/types.ts`** (the import is erased at
 runtime, so loading is unaffected either way), and restart Shepherd. See
--- a/examples/plugins/README.md
+++ b/examples/plugins/README.md
@@ -18,13 +18,16 @@ into your plugins dir to actually run it.
 ```sh
 # 1. Copy the folder into your plugins dir (override the dir with SHEPHERD_PLUGINS_DIR).
 cp -r examples/plugins/spawn-labeler ~/.shepherd/plugins/
+# …or symlink it from a checkout so `git pull` keeps the installed plugin current:
+ln -s "$PWD/examples/plugins/spawn-labeler" ~/.shepherd/plugins/
 
 # 2. Restart Shepherd to load it. (Optional: fix the type import first — see below —
 #    for editor/type-check ergonomics; it is NOT required to run the plugin.)
 systemctl --user restart shepherd
 ```
 
-Plugins load **at boot only** — edit the folder, then restart.
+Plugins load **at boot only** — edit the folder, then restart. A **symlinked** plugin dir
+loads identically to a copied one (the loader follows the link).
 
 ### Fix the type import for editor/type-check ergonomics (optional)
 
--- a/src/plugins/loader.ts
+++ b/src/plugins/loader.ts
@@ -100,11 +100,24 @@ export class PluginRegistry {
   async loadAll(): Promise<void> {
     let names: string[];
     try {
+      // `readdir(..., { withFileTypes: true })` does NOT follow symlinks: a
+      // symlink-to-directory Dirent reports isDirectory()===false. Resolve such
+      // entries with stat() (which follows the link) so a symlinked install — the
+      // natural "run a plugin from its checkout" workflow — loads like a copy (#1176).
       const entries = await readdir(this.deps.pluginsDir, { withFileTypes: true });
-      names = entries
-        .filter((e) => e.isDirectory())
-        .map((e) => e.name)
-        .sort();
+      names = [];
+      for (const e of entries) {
+        if (e.isDirectory()) {
+          names.push(e.name);
+        } else if (e.isSymbolicLink()) {
+          try {
+            if ((await stat(join(this.deps.pluginsDir, e.name))).isDirectory()) names.push(e.name);
+          } catch {
+            /* dangling symlink — skip */
+          }
+        }
+      }
+      names.sort();
     } catch {
       return; // missing dir → nothing to load
     }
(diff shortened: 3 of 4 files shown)


scoop (commented):
## Review

The PR fixes #1176 (plugin loader silently skips symlinked plugin dirs) by resolving symlink Dirents with `stat()` before the directory filter. The change is correct and faithful to the proposed fix.

### Verification

- **Imports present** — `stat` and `join` are both imported (`src/plugins/loader.ts:10-11`), so the new `stat(join(...))` call resolves with no added import. `stat` from `node:fs/promises` follows symlinks (vs. `lstat`), which is exactly what's needed.
- **Logic correct** (`src/plugins/loader.ts:108-120`): real directories are pushed as before; symlink entries are `stat()`'d and pushed only when the target is a directory; the `try/catch` swallows dangling-symlink `ENOENT` so it's skipped without throwing. Symlinks-to-files and other non-dir Dirents fall through and are dropped — preserving the prior `isDirectory()`-only behavior. `names.sort()` keeps the alphabetical load order the rest of the registry relies on (`src/plugins/loader.ts:91`).
- **No-op invariant preserved** — the outer `try/catch` around `readdir` still returns early on a missing dir (`src/plugins/loader.ts:121-123`); the new code lives inside that try, so a `stat` failure on an individual entry is caught by the inner catch and cannot break the whole scan.
- **Tests** (`test/plugins.test.ts:138-165`) cover both acceptance cases: a symlinked plugin loads with `health === "ok"` (verified `list()`/`health` exist at `src/plugins/loader.ts:366,371` and `loadedCount()` at `:377`), and a dangling symlink is ignored without throwing. `symlinkSync` is correctly added to the import (`test/plugins.test.ts:2`). The default manifest written by `writePlugin` matches the shape other passing tests use, so `health: "ok"` is a valid expectation.
- **Docs** (`docs/plugins.md`, `examples/plugins/README.md`) are updated consistently to mention the `ln -s` workflow alongside `cp -r`, matching the new loader behavior.

Acceptance criteria from the issue are all met: symlinked dir loads like a copy, dangling symlink ignored, zero-plugin/missing-dir no-op preserved, and a loader test covers the symlinked case.

No blocking issues, no nits found.

PR #1178, opened 2026-06-27
fix(ui): responsive settings tabs + reorder, surface Plugins on mobile

scoop:
## Problem

The settings dialog tab strip is a plain `display:flex` row with no wrap/scroll/shrink. The six uppercase, letter-spaced tabs sum to ~560–610px against the card's ~488px inner width — in **both** EN and DE (German "Arbeitsbereich" is the worst case). So the row overflows wherever it renders:

- **Mobile:** the full-screen card is `overflow:hidden`, so the spillover is clipped — the last tab (**Plugins**) lands off-screen and is unreachable.
- **Unfolded fold:** same overflow, just visible — the tabs shoot out of the container.

## Fix — a new responsive strategy

A reactive `isNarrow` (`matchMedia("(max-width: 768px)")`, wired with a `change` listener mirroring `theme.svelte.ts`) swaps the control at the existing card breakpoint:

- **Desktop (>768px):** keep the `role="tablist"` underline strip but add `flex-wrap: wrap`, so a fixed tab set simply wraps to a second row in the 520px card. Zero JS for layout; correct in every locale and at any `--ui-scale` (no scroller/ResizeObserver/fade needed).
- **Mobile (≤768px):** the tabs collapse into a full-width `<select bind:value={tab}>` on the shared select recipe. The tablist isn't rendered in this branch, so there are **no dangling `aria-controls`**.

Panels move from `aria-labelledby` (which pointed at tab buttons that don't exist on mobile) to a static `aria-label`, so labelling holds in both modes.

## Reorder

Tab order is now **Workspace → Coding CLIs → Plugins → Session → Device → Diagnostics** (Plugins moved up; Diagnostics, a troubleshooting tab, drops to last). Workspace stays the default landing tab; Plugins still only renders when ≥1 plugin is loaded.

## Tests

New regression suite in `Settings.browser.test.ts` renders in a real browser at:
- **360×900** → asserts the Plugins entry is reachable via the dropdown (`getByRole("option", …)`) and that no `role="tab"` exists in this mode.
- **1280×900** → asserts Plugins renders as a `role="tab"` in the strip.

## Verification

- `cd ui && bun run check:i18n` → 2 locales in parity (no new keys; reuses `settings_tab_*` / `settings_tabs_aria`)
- `bun run test:browser -- Settings.browser` → 8/8 pass

No new i18n keys, no glossary terms, and no feature-catalog entry (`fix(ui):` does not arm the catalog gate). [no-feature-entry]

### DIFF
--- a/ui/src/lib/components/Settings.browser.test.ts
+++ b/ui/src/lib/components/Settings.browser.test.ts
@@ -192,6 +192,54 @@ describe("Settings api-key verify", () => {
   });
 });
 
+// Regression: on a narrow viewport the tab strip can't fit one row in the card and,
+// when full-screen, the overflow is clipped — the Plugins tab (last in the visible
+// set when present) used to land off-screen. It now collapses into a dropdown so
+// every tab stays reachable. On desktop the strip wraps and keeps tab semantics.
+describe("Settings responsive tab navigation", () => {
+  const onePlugin = [
+    {
+      id: "p1",
+      name: "Demo",
+      version: "1.0.0",
+      health: "ok" as const,
+      lastError: null,
+      status: {},
+    },
+  ];
+
+  afterEach(async () => {
+    await page.viewport(1280, 900); // restore a sane width for other suites
+  });
+
+  it("narrow viewport → tabs collapse into a dropdown that lists the Plugins tab", async () => {
+    await page.viewport(360, 900); // ≤768px → mobile dropdown
+    render(Settings, { plugins: onePlugin, onclose: noop, onsaved: noop });
+
+    // The strip is replaced by a single labelled combobox…
+    await expect
+      .element(page.getByRole("combobox", { name: m.settings_tabs_aria() }))
+      .toBeInTheDocument();
+    // …and Plugins is reachable as an option (it was clipped off-screen before).
+    await expect
+      .element(page.getByRole("option", { name: m.settings_tab_plugins() }))
+      .toBeInTheDocument();
+    // No tablist tab in this mode.
+    await expect
+      .element(page.getByRole("tab", { name: m.settings_tab_plugins() }))
+      .not.toBeInTheDocument();
+  });
+
+  it("desktop viewport → Plugins renders as a tab in the strip", async () => {
+    await page.viewport(1280, 900); // >768px → tablist strip
+    render(Settings, { plugins: onePlugin, onclose: noop, onsaved: noop });
+
+    await expect
+      .element(page.getByRole("tab", { name: m.settings_tab_plugins() }))
+      .toBeInTheDocument();
+  });
+});
+
 // Regression guard for the issue's "no false success" criterion (PR #703): the green
 // "Fixed" toast must only fire when the RE-PROBED snapshot shows the targeted check ok.
 // A command that exits 0 but leaves the check non-ok (e.g. installer succeeds but its
--- a/ui/src/lib/components/Settings.browser.test.ts
+++ b/ui/src/lib/components/Settings.browser.test.ts
@@ -228,14 +228,22 @@ describe("Settings responsive tab navigation", () => {
     await expect
       .element(page.getByRole("tab", { name: m.settings_tab_plugins() }))
       .not.toBeInTheDocument();
+    // The active panel is a plain labelled region here (no tablist to own a tabpanel).
+    await expect
+      .element(page.getByRole("region", { name: m.settings_tab_workspace() }))
+      .toBeInTheDocument();
   });
 
-  it("desktop viewport → Plugins renders as a tab in the strip", async () => {
+  it("desktop viewport → Plugins renders as a tab and the panel is a tabpanel", async () => {
     await page.viewport(1280, 900); // >768px → tablist strip
     render(Settings, { plugins: onePlugin, onclose: noop, onsaved: noop });
 
     await expect
       .element(page.getByRole("tab", { name: m.settings_tab_plugins() }))
+      .toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

The PR delivers a sound two-pronged fix for the task (Plugins tab invisible on mobile / tab strip overflowing on unfolded fold):

1. **Desktop strip** now `flex-wrap: wrap` (`ui/src/lib/components/Settings.svelte:1124`), so the fixed tab set wraps to a second row instead of overflowing the 520px card — this addresses the "shoots out of the container in unfolded fold" symptom for viewports wider than the breakpoint.
2. **Narrow viewport (≤768px)** swaps the strip for a full-width native `<select>` dropdown (`ui/src/lib/components/Settings.svelte:633-642`), so every tab — including Plugins (last in the visible set) — stays reachable inside the full-screen card.
3. Plugins is reordered higher in `ALL_TABS` (`Settings.svelte:50-57`).

### Verification
- Breakpoint `(max-width: 768px)` (`Settings.svelte:114`) matches the full-screen card breakpoint `@media (max-width: 768px)` (`Settings.svelte:1458`), so the dropdown engages exactly when the card goes full-screen and clips overflow — verified consistent.
- `.tab-select` styling is a faithful copy of the existing `.model-select` "shared select recipe" incl. `appearance: none` with no custom caret — verified against `Settings.svelte:1315-1330`; the missing dropdown arrow is consistent with the app's existing selects, not a new regression.
- `matchMedia` guard + `onMount` listener with cleanup mirrors `theme.svelte.ts` (`matchMedia` guard verified against ui/src/lib theme helper:44 and :128-129); SSR-safe (falls back to desktop tabs).
- `bind:value={tab}` against `<option value={t.id}>` is type-correct: `tab` is `$state<TabId>` and option values are the literal ids (`Settings.svelte:105`, `:638-640`).
- Panel `aria-labelledby="settings-tab-*"` → `aria-label={m.settings_tab_*()}` switch is correct: in dropdown mode the tab buttons (and their `id`s) aren't rendered, so the old `aria-labelledby` would dangle; `aria-label` resolves to the same localized string. The desktop tab→panel link via `aria-controls` is preserved.
- Message keys used by the diff/tests exist with en+de parity: `settings_tabs_aria`, `settings_tab_workspace`, `settings_tab_plugins` — verified against ui/messages/en.json:352-353,1201 and ui/messages/de.json:352-353,1201. No new keys introduced.
- Test harness: `afterEach`/`noop`/top-level `beforeEach` are all in scope for the new `describe` (`Settings.browser.test.ts:1,84,121,199`); the new `afterEach` restoring viewport to 1280×900 prevents the narrow-viewport set from leaking into later suites — verified.

### Non-blocking nit
- In dropdown mode the panels keep `role="tabpanel"` + `tabindex="0"` but no `role="tab"`/`tablist` is rendered, leaving orphaned tabpanels (navigation happens via the `combobox`). The author's comment at `Settings.svelte:633-636` consciously keeps them as "labelled tabpanels," and they're now self-labelled via `aria-label`, so it's functional — but a tabpanel with no owning tab is a minor ARIA inconsistency. Consider switching the panels to `role="region"` (or dropping the tabpanel role) when `isNarrow`.

scoop (commented):
## Summary

This PR makes the Settings tab strip responsive: below 768px the strip is swapped for a native `<select>` dropdown (so the Plugins tab is reachable instead of clipped off-screen in the full-screen mobile card), and on desktop the strip now `flex-wrap`s instead of overflowing. Plugins is also reordered higher in `ALL_TABS`. This satisfies the task.

## Re-review of prior finding

Prior finding — "in narrow/dropdown mode the panels keep `role=tabpanel` but no tablist is rendered, leaving orphaned tabpanels; consider `role=region` when isNarrow" — is **addressed**. The new `panelShape` action (ui/src/lib/components/Settings.svelte:135-143) sets `role="region"` and removes `tabindex` when `isNarrow`, and `role="tabpanel"` + `tabindex="0"` otherwise. The narrow branch (Settings.svelte:648-657) renders a `<select>` rather than the `role="tablist"` strip, so no `aria-controls`/tabpanel pairing is left dangling. The new browser test asserts `region`-by-name on mobile and `tabpanel`-by-name on desktop (Settings.browser.test.ts:228-247).

## Verification

- Message keys used by the diff (`settings_tabs_aria`, `settings_tab_plugins`, `settings_tab_workspace`, `settings_tab_coding_agents`, `settings_tab_session`, `settings_tab_device`, `settings_tab_diagnose`) all resolve and have en+de parity — verified against ui/messages/en.json:352-353,1200-1201 and ui/messages/de.json:352-353,1200-1201. No new keys introduced.
- `role`/`tabindex` are applied imperatively by the action while `aria-label`/`id`/`hidden` stay in the template, so Svelte's reactive re-render won't clobber the action-set attributes — the two never manage the same attribute.
- `.tab-select`'s `appearance: none` with no custom chevron is consistent with the existing `.model-select` recipe in the same file (Settings.svelte:1332), so no new affordance regression.
- `bind:value={tab}` on the `<select>` with `<option value={t.id}>` correctly drives the `TabId` state; keyboard `onTabKey`/`tabEls` only matter on the desktop branch where the buttons exist.

No blocking issues found.

- ui/src/lib/components/Settings.svelte: `isNarrow` initializes from `matchMedia` (false under SSR where `matchMedia` is undefined) and is corrected on the client, so a server-rendered strip could differ from a client-rendered dropdown on a narrow device; harmless for a client-opened modal and mirrors the existing theme.svelte.ts pattern, but a hydration-mismatch warning is possible if this modal is ever SSR'd.

PR #1181, opened 2026-06-27
fix(ui): tidy stale Codex autopilot copy + exclude research from unavailable badge

### LINKED ISSUE #1173: Follow-up: tidy two minor non-blockers from #1140 (Codex autopilot-to-PR alpha)
Merge-train follow-up for #1140 (merged). Two cosmetic/non-blocking items the critic flagged that were intentionally deferred to land the alpha:

1. **Stale comment in `NewTask.svelte` (~:348-352).** Still says "Codex can't plan-gate or autopilot yet" — autopilot is now enabled for isolated Codex sessions. Narrow the comment to plan-gate only.
2. **Cosmetic "unavailable" badge on Codex *research* tasks.** `codexAutopilotUnavailable` doesn't exclude `research`, so a non-isolated Codex research task with autopilot-on shows the "autopilot unavailable" badge even though the directive is correctly suppressed at spawn. Exclude `research` from that derivation so the badge matches actual behavior.

Both are UI-copy/derivation tidies; no functional risk. Context: PR #1140, critic re-reviews by kai-osthoff.

scoop:
Tidies the two non-blocking items deferred from #1140 (Codex autopilot-to-PR alpha), flagged by the critic and tracked in #1173.

## Changes
1. **Stale comment** in `NewTask.svelte` — still said "Codex can't plan-gate **or autopilot** yet." Autopilot is now enabled for isolated Codex sessions (#1140), flowing through `automationFlag` like Claude. Narrowed the comment to plan-gate only (which genuinely stays Codex-forced-off via `planGateFlag`, since Codex gets no `--append-system-prompt` directive wiring).
2. **`codexAutopilotUnavailable` badge** — didn't exclude `research`, so a non-isolated Codex *research* task with autopilot-on showed the "autopilot unavailable" badge even though a research session's autopilot directive is suppressed at spawn regardless of provider/isolation. Added `&& !s.research` so the badge matches actual behavior, plus a regression test.

Both are UI-copy/derivation tidies; no functional change to spawn or gating logic.

## Verification
- `cd ui && bun test src/lib/format.test.ts` → 60 pass (incl. new research-exclusion case)
- `cd ui && bun run check` → 0 errors

Closes #1173. Context: PR #1140.

### DIFF
--- a/ui/src/lib/components/NewTask.svelte
+++ b/ui/src/lib/components/NewTask.svelte
@@ -373,11 +373,13 @@
     if (!autopilotTouched) autopilot = autopilotDefault;
   });
 
-  // Codex can't plan-gate or autopilot yet. Rather than mutate planGate/autopilot
-  // (which would silently lose a manual override across a Codex round-trip), the
-  // checkboxes just DISPLAY off + disabled while Codex is selected (see
-  // NewTaskRunSettings) and submit forces them off via automationFlag — the
-  // underlying Claude-context choice is preserved untouched.
+  // Codex can't plan-gate yet (no spawn directives via --append-system-prompt, so the
+  // plan-gate wiring isn't available for it). Rather than mutate planGate (which would
+  // silently lose a manual override across a Codex round-trip), the plan-gate checkbox
+  // just DISPLAYS off + disabled while Codex is selected (see NewTaskRunSettings) and
+  // submit forces it off via planGateFlag — the underlying Claude-context choice is
+  // preserved untouched. Autopilot, by contrast, is now available for isolated Codex
+  // sessions (#1140) and flows through automationFlag like Claude.
 
   // Effective default model = repo override (if not "inherit") → global default → promo.
   // Re-seeds the picker when the repo config loads / the repo changes, unless an explicit
--- a/ui/src/lib/format.test.ts
+++ b/ui/src/lib/format.test.ts
@@ -167,6 +167,19 @@ describe("autopilotBadgeShown", () => {
         false,
       ),
     ).toBe(false));
+
+  it("returns false for codex + non-isolated + research (directive suppressed at spawn)", () =>
+    expect(
+      autopilotBadgeShown(
+        session({
+          agentProvider: "codex",
+          isolated: false,
+          autopilotEnabled: true,
+          research: true,
+        }),
+        false,
+      ),
+    ).toBe(false));
 });
 
 describe("relativeAge", () => {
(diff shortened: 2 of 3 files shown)


scoop (commented):
## Summary

This PR is a small, focused cleanup of stale Codex-autopilot copy plus one behavioral fix to `codexAutopilotUnavailable`: research sessions are now excluded from the "unavailable" badge. I verified each correctness-relevant claim against the actual code; the change is sound and internally consistent.

## Verification

- **`ui/src/lib/format.ts:186`** — `codexAutopilotUnavailable` now ANDs in `!s.research`. `Session.research` is a required `boolean` (verified `ui/src/lib/types.ts:752`), so `!s.research` is well-defined for all inputs — no `undefined` edge case.
- **Comment accuracy (`ui/src/lib/format.ts:180-182`)** — the claim that a research session's autopilot directive is suppressed at spawn regardless of provider/isolation checks out:
  - Codex inline autopilot directive is gated on `!input.research && wt.isolated && effectiveAutopilot(...)` — verified `src/service.ts:1839`.
  - For Claude, `composeSystemPrompt` replaces the autopilot directive with `RESEARCH_DIRECTIVE` when `opts.research` is set — verified `src/service.ts:822-828` and the directive's own doc at `src/service.ts:509-512` ("It SUPPRESSES the plan-gate, autopilot, and build-queue directives").
  So excluding research from the non-isolated-codex "stood-down" badge is correct: the badge would otherwise be misleading noise, since autopilot is already suppressed at spawn for research irrespective of isolation. (Note: runtime gate-steering still drives research sessions via `RESEARCH_PROCEED_STEER`, `src/autopilot.ts:307` — but the spawn-time directive the badge reflects is genuinely suppressed, so the copy is accurate.)
- **Comment update in `ui/src/lib/components/NewTask.svelte:376-382`** — the revised copy matches the code: `planGateFlag` forces `false` for codex (`NewTask.svelte:613-616`) while `automationFlag` flows the user's choice through for both providers (`NewTask.svelte:607-609`, applied at `:638-639`). The old comment's claim that autopilot was also forced off for codex was stale; the new copy correctly reflects that autopilot now flows through `automationFlag` like Claude. `#1140` reference is a comment-only changelog pointer (not verifiable as a live identifier, but harmless).
- **Test (`ui/src/lib/format.test.ts:171-182`)** — the added case (codex + non-isolated + research + autopilotEnabled:true → `autopilotBadgeShown` false) is consistent with the new logic: `codexAutopilotUnavailable` returns false because `!s.research` is false, and the session sets neither `autopilotPaused` nor `autopilotComplete`. Mirrors the existing sibling cases at `:130-169`.

No locale catalogs are touched (these are agent-facing directives / internal helpers, not i18n strings), so no parity concern.

Nothing blocking found.

Issue #1169, opened 2026-06-27
Up Next — cross-repo ranked queue of work to start


scoop:
## Summary

A new top-level **"Up Next"** lens: a flat, ranked, **cross-repo feed of un-started GitHub issues** so the operator can see — at a glance — the single most worthwhile thing to start next, and start a session on it (or several) in one click.

Today there is no cross-repo "what should I work on next" surface. The Backlog is repo-*centric* (pick a repo, drill in). The Herd Rundown synthesizes attention across *live sessions*, not cold/un-started work. Up Next fills that gap and is, conceptually, the un-started counterpart to Rundown.

Scoped via a grilling session. This issue is the v1 spec. **Build starts with a Phase-0 spike that is an explicit go/no-go gate** (see below) — do not build the UI before the spike passes.

## Scope (v1)

- **Unit ranked:** individual work items in one flat list — **open issues + epics-as-one-unit**. Epics rank by their next-actionable child; DAG-blocked epics suppressed.
- **Excluded item types:** open PRs, Dependabot/release-please PRs. (Open PRs imply work already exists — that's the in-flight lane, out of scope here.)
- **Repos:** forge-backed (GitHub) repos only. **Lightweight-repo mode is a documented gap** for v1 (its backlog is PR-shaped, no cold-issue source to rank).

## Ranking — deterministic, lexicographic tiers (priority-dominant)

No LLM in v1 — explainable + cheap + restart-safe. LLM re-ranking is a documented future enrichment.

1. **`shepherd:priority`** items, across all repos — within this group: warm-first (`lastUsedAt` desc), then oldest age.
2. then **clustered by warm repo** (`lastUsedAt` desc); within each repo: **ready-epic-child → bug → feature → oldest age**.

The strongest signal (`shepherd:priority`) is already an explicit human judgment; the epic DAG already encodes what's unblocked. A predictable, explainable order beats an opaque score.

### Exclusions (applied before ranking)

- DAG-blocked epic children (can't start) — suppress until unblocked.
- **`shepherd:active`** — the reliable claim/in-progress signal (retire-based; persisting post-retire is fine — claimed-but-not-done work shouldn't be recommended as "next"). **This is the primary in-progress filter.**
- bot-managed / `wontfix` / `blocked`-labelled issues.
- Items with an open PR linked to them — **best-effort secondary** filter only (see linkage below).

### Issue↔PR linkage mechanism

[...]

PR #1186, opened 2026-06-27
fix(up-next): exclude hidden repos from the ranked queue

scoop:
## What

The **Up Next** cross-repo ranked queue (#1169/#1172) surfaced un-started work from **every** forge-backed repo, ignoring the per-repo `hidden` flag — a repo hidden in the Backlog repos panel still appeared in Up Next. This makes Up Next honor `hidden`, so a hidden repo contributes **no** section/items to the queue.

## Root cause

`listForgeRepos` (the thunk passed to `UpNextService` in `src/index.ts`) filtered only by forge kind and never consulted `store.hiddenRepoPaths()`, unlike `buildBacklogPayload` (`src/server.ts`).

## Approach — two complementary filters

1. **Source filter (`buildUpNextRepos`, `src/up-next.ts`)** — runs at compute. Forge-filter + the realpath→raw reconcile (`reconcileRealPathsToRaw`) + hidden-filter + the exact field mapping, all inside one tested helper. Also avoids wasted forge `listIssues` fan-out for hidden repos.
2. **Read-time filter (`excludeHiddenSections`, `src/up-next-core.ts`)** — pure; applied to the cached snapshot in `handleUpNextGet` against the live reconciled hidden set (`upNext.hiddenRepoPathsRaw()`). Reflects a hide on the **next GET** — instant, no forge fan-out — so a repo hidden *after* the last compute doesn't linger in the cached payload.

**Path-space note:** `repo_config` keys are realpath/`safeRepoDir`-resolved, but the snapshot/`listRepos` paths are raw `join(repoRoot,name)`. Both filter sites reconcile the hidden set to raw space (the same dance the Backlog does) so a persisted hide matches under a symlinked `repoRoot`.

## Deliberate deviation from prior art (per house rules)

The **Backlog** *includes* hidden repos in its payload and merely flags them `hidden: true` (the UI collapses them, keeping per-repo settings reachable). **Up Next instead fully excludes** them — it has no settings drill-down and exists only to rank actionable work, so a hidden repo should contribute nothing. This divergence is documented inline (`src/up-next.ts`) and here.

A second framing note: the read-time filter (instant on next GET, no fan-out) supersedes an earlier "eventual / next-tick" framing — it removes the up-to-15-min staleness without wiring a reactive refresh, matching the Backlog's request-driven freshness.

## Tests

- `buildUpNextRepos`: local/non-forge excluded; hidden raw-path excluded; **real reconcile via a symlinked `repoRoot`** (realpath ≠ raw); survivors carry exact `repoPath`/`repoSlug`/`repoLabel` (no label regression).
- `excludeHiddenSections`: hidden repo-section dropped + `repoCount` decremented; hidden item pulled from the priority section (section dropped when emptied) + `totalCount` recomputed; no-match/empty snapshot returned unchanged.
- Server integration: `GET /api/up-next?peek` strips hidden-repo sections via `hiddenRepoPathsRaw()`.

Verified: `bun run typecheck` clean · `bun run lint` clean · full `bun test` 5144 pass · branch hygiene linear off `main`.

Server-only behavior `fix(...)` touching only `src/**` + `test/**` — i18n / feature-catalog gates do not arm. `[no-feature-entry]`

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -84,7 +84,7 @@ import { tailLines } from "./blocked";
 import { recommendPrompt } from "./prompt-recommend";
 import { CountsService } from "./backlog";
 import { BacklogPoller } from "./backlog-poller";
-import { UpNextService } from "./up-next";
+import { UpNextService, buildUpNextRepos } from "./up-next";
 import { ProcessReaper, reapDeletedWorktreeOrphans } from "./process-reaper";
 import {
   sweepClaudeTmp,
@@ -95,7 +95,7 @@ import {
 } from "./tmp-sweep";
 import { runSessionUsageBackfill } from "./usage-backfill";
 import { PreviewService } from "./preview";
-import { listRepos, listReposPathForReal } from "./repos";
+import { listRepos, listReposPathForReal, reconcileRealPathsToRaw } from "./repos";
 import { enrichLandingEpics, type CompletedEpic } from "./completed-epic";
 import { DistillerService, defaultScratch } from "./distiller";
 import { OptimizerService, defaultOptimizerScratch } from "./optimizer";
@@ -1841,16 +1841,14 @@ backlogPoller.start();
 // by a 15-min background loop; reuses the drain's epic pipeline for ready-child gating and
 // pushes each fresh snapshot to clients over the WS (upnext:snapshot).
 const upNext = new UpNextService({
-  // Forge-backed repos only (skip local/lightweight repos — no cold-issue source to rank).
+  // Forge-backed, non-hidden repos only. buildUpNextRepos owns forge-kind filtering,
+  // the realpath→raw reconcile, hidden-repo filtering, and the exact field mapping.
   listForgeRepos: () =>
-    listRepos(config.repoRoot)
-      .map((r) => ({ entry: r, forge: resolveForge(r.path) }))
-      .filter((x) => x.forge != null && x.forge.kind !== "local")
-      .map((x) => ({
-        repoPath: x.entry.path,
-        repoSlug: x.forge!.slug,
-        repoLabel: x.entry.display,
-      })),
+    buildUpNextRepos({
+      repos: listRepos(config.repoRoot),
+      resolveForge,
+      hiddenRealPaths: store.hiddenRepoPaths(),
+    }),
   resolveForge,
   lastUsedByRepo: () => store.lastUsedByRepo(),
   buildEpic: (repoPath, run) => drain.buildEpic(repoPath, run),
@@ -1913,6 +1911,8 @@ const appDeps: AppDeps = {
   upNext: {
     snapshot: () => upNext.snapshot(),
     refresh: () => upNext.refresh(),
+    hiddenRepoPathsRaw: () =>
+      reconcileRealPathsToRaw(store.hiddenRepoPaths(), listRepos(config.repoRoot)),
   },
   verifyKey: () => verifyApiKey({ herdr }),
   backlog,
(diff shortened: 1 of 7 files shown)


scoop (commented):
## Summary

The PR makes the Up Next feature exclude hidden repos. It does so at two layers:

1. **Source/compute filter** — `buildUpNextRepos` (src/up-next.ts:38) reconciles the realpath-keyed hidden set to raw paths via `reconcileRealPathsToRaw` and skips hidden (and non-forge/local) repos before any forge fan-out. Wired in src/index.ts:1847 with `hiddenRealPaths: store.hiddenRepoPaths()`.
2. **Read-time filter** — `excludeHiddenSections` (src/up-next-core.ts:275) strips hidden repo sections and hidden priority items from the cached snapshot on every GET (src/server.ts:770), so a repo hidden after the last 15-min compute is excluded immediately. Wired via the new `hiddenRepoPathsRaw()` AppDep (src/index.ts:1914), which reconciles `store.hiddenRepoPaths()` to raw space.

The task is satisfied and I found no blocking issues.

## Verification

- **Identifier resolution**: `reconcileRealPathsToRaw` confirmed at src/repos.ts:103 (realpath→raw translation); `RepoEntry` at src/repos.ts:33 (carries `path`/`display`); `store.hiddenRepoPaths()` at src/store.ts:1149 returns `repo_config` keys (realpath/`safeRepoDir`-resolved per the path-space note). `UpNextSection`/`UpNextItem`/`UpNextSnapshot` confirmed at src/up-next-core.ts:35/19/47.
- **Path-space consistency**: `buildSnapshot` populates section `repoPath` (src/up-next-core.ts:253) and priority item `repoPath` (src/up-next-core.ts:218,239) from raw `RepoInput.repoPath`, which traces back to `UpNextRepo.repoPath = entry.path` (raw) in `buildUpNextRepos`. Both filters compare against a hidden set reconciled to raw space, so the comparison key-spaces match. Verified against src/up-next-core.ts:253 and src/index.ts:1915.
- **Field mapping preserved**: `buildUpNextRepos` returns `{repoPath: entry.path, repoSlug: forge.slug, repoLabel: entry.display}`, identical to the pre-PR inline mapping (src/up-next.ts:46 vs. the removed src/index.ts block in the diff).
- **Interface consumers**: the new required `hiddenRepoPathsRaw(): Set<string>` member of the optional `upNext` AppDep is provided by both constructors — src/index.ts:1914 and the test harness at test/server-up-next.test.ts:66. No other `upNext:` constructions exist (grep over src/ and test/). Other AppDeps omit `upNext` entirely (it is optional), so they are unaffected.
- **excludeHiddenSections logic**: `repo` sections matching the hidden set are dropped with `repoCount` decremented (floored via `Math.max(0,…)`); `priority` sections have hidden items filtered with `totalCount` recomputed and the section dropped when empty; empty `hiddenRaw` returns the same reference (fast path). Section `kind` is only `"priority" | "repo"` (src/up-next-core.ts:36), so the trailing `else` branch is unreachable but harmless. Behavior is covered by tests in test/up-next-core.test.ts (cases e/f/f2/g/g2).

- `repoCount` semantics at read time: `excludeHiddenSections` decrements `repoCount` only for removed *repo sections*. A hidden repo whose only contribution was a priority item (so it produced no repo section — priority items are pulled out of repo clusters at src/up-next-core.ts:238) will have its priority item stripped but `repoCount` will not be decremented for it until the next background recompute (where `buildUpNextRepos` excludes it entirely). Given `repoCount` is documented as "repos scanned this refresh" (src/up-next-core.ts:50–51) and this is a transient display count corrected within ≤15 min, this is a deliberate/acceptable trade-off, not a defect.

Could not find any in-repo i18n/message-catalog or UI-string changes in this diff — the change is backend TypeScript only, so no locale-parity concerns apply.

PR #1163, opened 2026-06-27
fix(ci): grant doc-automerge checks/statuses read scope

scoop:
## Problem

The **Auto-merge doc PRs** workflow keeps failing ([run 28293837702](https://github.com/erwins-enkel/shepherd/actions/runs/28293837702)) once an open doc PR exists. It dies on `gh pr view --json statusCheckRollup` with:

```
GraphQL: Resource not accessible by integration
(repository.pullRequest.statusCheckRollup.nodes.0.commit.statusCheckRollup)
```

Under `set -euo pipefail` that propagates and the job exits 1.

## Root cause

The workflow declares an explicit `permissions:` block, which defaults every unlisted scope to `none`. `statusCheckRollup` walks the head commit's `CheckRun` nodes (needs `checks: read`) and legacy `StatusContext` nodes (needs `statuses: read`) — neither was granted, so GitHub rejects the nested query. It only surfaces now because the sub-resource is only read when a PR actually has checks.

## Fix

Add `checks: read` + `statuses: read` to `permissions:`. Least-privilege, read-only — no change to the merge logic.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1195, opened 2026-06-28
fix(ci): grant actions:read so doc-automerge can read statusCheckRollup

scoop:
## Problem

The **Auto-merge doc PRs** workflow (`doc-automerge.yml`) fails red on every trigger before it evaluates any PR — so no doc PR ever auto-merges. [Run 28313713755](https://github.com/erwins-enkel/shepherd/actions/runs/28313713755):

```
Found 1 open doc PR(s): 1191
GraphQL: Resource not accessible by integration (...statusCheckRollup...checkSuite.workflowRun)
##[error]Process completed with exit code 1.
```

## Root cause

The failure is at the `gh pr view --json statusCheckRollup` call, *before* any eligibility gate. That JSON field issues a fixed GraphQL query that, for every `CheckRun`, traverses `checkSuite → workflowRun → workflow { name }`. Reading the `workflowRun` node requires the **`actions: read`** permission.

The earlier scope fix (#1163, commit f13a69b7) added `checks: read` + `statuses: read` — which cover the check-run / legacy-status *leaves* the query reads, but **not** the `workflowRun` node the same query also walks. With an explicit `permissions:` block, every unlisted scope defaults to `none`, so `actions` was `none` and the call died the moment a PR had any Actions check. Under `set -euo pipefail` the failed command substitution aborts the job. That's why it "still failed" after the prior fix — two of the three scopes the one query needs were granted.

## Fix

Add `actions: read` (read-only) to the `permissions:` block, with a comment documenting the `checkSuite.workflowRun` traversal. One line; no write scopes added; consistent with the workflow's least-privilege intent.

## Verification (pre-merge, real restricted GITHUB_TOKEN)

`workflow_dispatch` runs the *dispatched ref's* workflow definition (the trigger already exists on `main`), so dispatching this branch exercises the edited `permissions:` under the same scoped token that failed. Dispatched [run 28315626893](https://github.com/erwins-enkel/shepherd/actions/runs/28315626893) with doc PR #1191 open:

```
Found 1 open doc PR(s): 1191
#1191: author_association=CONTRIBUTOR not OWNER/MEMBER — skip
```

The job got **past** the previously-failing `gh pr view` call into the per-PR eligibility gates, with no `Resource not accessible` error — empirically confirming `actions: read` was the last missing scope (the `#1191` skip is a legitimate provenance gate, not a failure).

[no-feature-entry] — server-only CI plumbing, no user-facing UX.

### DIFF
--- a/.github/workflows/doc-automerge.yml
+++ b/.github/workflows/doc-automerge.yml
@@ -37,6 +37,12 @@ permissions:
   # dies with "Resource not accessible by integration" the moment a PR has any check.
   checks: read
   statuses: read
+  # The same statusCheckRollup query ALSO traverses CheckRun → checkSuite → workflowRun
+  # (gh's fixed GraphQL fragment fetches the run's workflow name). Reading the workflowRun
+  # node needs `actions: read` — `checks`/`statuses` above cover only the leaf verdicts, not
+  # this node. Without it the query still dies "Resource not accessible by integration
+  # (...checkSuite.workflowRun)" on any Actions-checked PR (the #1163 scope fix missed it).
+  actions: read
 
 # One merge pass at a time; never cancel a pass mid-merge.
 concurrency:


scoop (commented):
## Summary

Single-line CI fix: grants `actions: read` to the `GITHUB_TOKEN` in `.github/workflows/doc-automerge.yml` so the `gh pr view --json statusCheckRollup` call stops dying with "Resource not accessible by integration". This directly addresses the failing auto-merge run in the task.

## Verification

- The new permission is the only change. It lands in the sole workflow-level `permissions:` block (`.github/workflows/doc-automerge.yml:31`); there is no job-level `permissions:` override that would shadow it (verified: only one `permissions:` in the file), so the grant reaches the `automerge` job that runs `statusCheckRollup` (`.github/workflows/doc-automerge.yml:90`).
- Indentation is correct — `actions: read` is a 2-space sibling of `checks: read`/`statuses: read` (`.github/workflows/doc-automerge.yml:38-39, 45`). The change is purely additive scope, so no behavioral regression is possible.
- Lineage in the comment checks out: `git log` for this file shows `f13a69b7` (#1163) added `checks`/`statuses` read and the current commit adds `actions: read`, matching the comment's claim that #1163 missed it.
- The stated mechanism (gh's fixed `statusCheckRollup` GraphQL fragment walks CheckRun → checkSuite → workflowRun, and the workflowRun node requires `actions: read`) is consistent with the documented GitHub error surface.

## Stated limitation

Could not verify gh's exact internal GraphQL fragment for `statusCheckRollup` against a live GitHub schema — it is not present in this repo. The fix is additive least-privilege scope, so even if the precise node were different, the change is harmless and is the standard remedy for this error class; assumed correct as written.

No blocking issues found.

PR #1196, opened 2026-06-28
fix(orchestration): fire auto plan-gate before the awaited drain chain (#1193)

### LINKED ISSUE #1193: Auto plan-review flaky since #1170: starved behind awaited drain pump in the new SessionRouter seam
## Summary

Automatic plan review (initial **and** re-review) has become flaky since yesterday — it frequently never starts. Manually hitting **Review plan** / **Re-review plan** always works. Most visible on the `claude-swap` plugin repo with all automation toggles on.

## Root cause

Regression introduced by **`9124026a` — feat(orchestration): one ordered 'what happens next' seam for the autonomous engine (#1094) (#1170)** (2026-06-27 19:32). Its message: *"Plan-gate moves into the seam's onStatusSettled hook. Hard cutover, no flag."*

The auto plan-review was an **independent, fire-and-forget** side-effect; it is now **sequenced strictly behind the awaited consumer chain** (drain → autopilot).

**Before** (old `session:status` subscribers in `src/index.ts`): on a `done` event, `planGate.consider()` fired immediately and independently, in parallel with `void autopilot.onDone(id)` and `void drain.onStatus(id)`. Nothing blocked it.

**After** (`src/session-router.ts:31-42`):

```
onStatus → buildSnapshot
         → await dispatch()        // drain.handle() THEN autopilot.handle(), each awaited
         → onStatusSettled()       // planGate.consider() runs ONLY here, last
```

- `src/session-router.ts:37-41` — `await this.dispatch(change)` completes before `onStatusSettled` (the plan-gate) fires.
- `src/session-router.ts:50-58` — `dispatch` awaits each consumer in order: **drain first**, then autopilot.
- `src/drain.ts:1786-1799` — `drain.handle()` for a `status` change calls `await this.pumpIfEnabled(s.repoPath)`.
- `src/drain.ts:1897-1902` → `pump()` (`src/drain.ts:1441-1452`) loops up to 100 steps, each doing multiple **unbounded, timeout-free** forge/network calls via `buildState`/`doRetire`/`doSpawn` (`forge.getIssue`, `listIssues`, `prReviewMeta`, `merge`, `openPr`, `service.create`/`archive`, …).
- `src/autopilot.ts:363-371` — `autopilot.handle()` then does `await this.onDone(id)` (previously fire-and-forget).
- `src/index.ts` (~1245-1248) — the `onStatusSettled` hook: `if (change.status === "done" && sess.planPhase === "planning") void planGate.consider(sess)`.

So a single planning session finishing now triggers a **full repo-wide drain pump** (and autopilot) that must fully resolve before that session's plan review is even considered. Any slow or hanging gh/git call in the pump stalls — or indefinitely blocks — the plan review.

The hook comment says it's *"reached regardless of a consumer throwing"* — true for **thrown** errors (`dispatch` try/catches them), but it does nothing for a consumer that is merely **slow or hangs**: `await c.handle()` pends and the settled hook never runs.

## Why this matches every symptom

- **Flaky / started yesterday** — depends on whether that tick's drain pump runs long / hangs on a network call; exactly matches the `9124026a` landing date.
- **Stuck on initial AND re-review** — both flow through the same `done` → router → `onStatusSettled` path.
- **Manual buttons always work** — `POST /api/sessions/:id/review-plan` (`src/server.ts:2029-2040`) calls `planGate.consider(s)` **directly**, bypassing the router/drain/autopilot chain.
- **Worst on `claude-swap` with all toggles on** — an active repo has the heaviest drain backlog/PR work, so the awaited pump is most often slow or stuck.

## Repro

1. Repo with all automation toggles on and a non-trivial backlog/PR activity (e.g. `claude-swap`).
2. Start a task that enters the planning phase and let the planning agent finish (status → `done`, `planPhase === "planning"`).
3. Observe the auto plan review frequently does not start.
4. Click **Review plan** → review starts immediately every time (direct `consider()`, bypasses the chain).

## Fix direction (not yet implemented)

Plan review is a **per-session** side-effect and should not wait on the repo-wide drain at all. Options:

1. **Decouple** — fire the plan-gate independently of `dispatch()` (don't await drain/autopilot before it), restoring the pre-`9124026a` independence while keeping the drain↔autopilot ordering the seam was built for.
2. **Bound the consumer awaits** with a timeout / `AbortSignal` so a slow or hanging consumer can't starve the settled hook.

Option 1 is preferred: the drain↔autopilot ordering exists to fix the retire-vs-steer race; the plan-gate was never part of that race and gains nothing from running after the pump.

cc — found via investigation on branch `shepherd/auto-review-plans-started`.

scoop:
Fixes #1193.

## Problem
Auto plan review (initial **and** re-review) became flaky since #1170 — it frequently never starts, while the manual **Review plan** button always works.

`#1170` moved the plan-gate from an independent fire-and-forget `session:status` subscriber into `SessionRouter`'s status hook, which ran **after** the awaited consumer chain:

```
onStatus → buildSnapshot → await dispatch()   // drain pump THEN autopilot, each awaited
                         → onStatusSettled()   // plan-gate ran ONLY here, last
```

`drain.handle()` runs an unbounded, timeout-free pump (up to 100 steps of forge/network calls). The router's `try/catch` only protects against a consumer that *throws* — one that merely **hangs** leaves `await consumer.handle()` pending forever, so the plan-gate hook never fires. Manual review works because `POST /api/sessions/:id/review-plan` calls `planGate.consider()` directly.

## Fix (Option 1 — decouple, per the issue)
Kick the independent status hook **before** awaiting `dispatch()`, so a slow/hanging consumer can never starve it. The drain→autopilot ordering (built for the retire-vs-steer race) is untouched; the plan-gate was never part of that race.

`PlanGateService.consider()` is self-contained — it reads `.shepherd-plan.md` and claims its in-flight slot **synchronously** — so firing it concurrently with the pump is safe and restores the pre-#1170 independence (no double-spawn).

- `src/session-router.ts` — `onStatus` fires the hook before `await dispatch()`; renamed `onStatusSettled` → `onStatusIndependent`; corrected the now-false "AFTER the chain" comments.
- `src/index.ts` — renamed the hook key at the construction site; `consider()` logic unchanged.
- `test/session-router.test.ts` — rewrote the ordering test to assert the hook fires **before** the chain, and added a regression test proving a **never-resolving** consumer no longer starves the hook. Drain→autopilot ordering, failure-isolation, unknown-session, and onGit tests unchanged.

**Out of scope:** Option 2 (timeout/AbortSignal on consumer awaits) — the pump's missing timeouts are a separate pre-existing concern; aborting mid-merge carries its own risk and isn't needed here.

Server-only change — no UI / i18n / feature-catalog / glossary surface.

## Verification
- `bun run lint` ✓
- `bun run typecheck` ✓
- `bun test test/session-router.test.ts` → 10 pass
- `bun test test/plan-gate-service.test.ts test/server-plan-gate.test.ts` → 72 pass

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1241,8 +1241,9 @@ const sessionRouter = new SessionRouter(
   {
     // Plan-gate is an independent status side-effect (not a [drain,autopilot] consumer): a
     // planning-phase session that just settled likely finished writing .shepherd-plan.md → kick the
-    // adversarial review. Runs after the consumer chain, reached regardless of a consumer throwing.
-    onStatusSettled: (change) => {
+    // adversarial review. Fires BEFORE the awaited consumer chain so a slow/hanging drain pump can
+    // never starve it (#1193); consider() is self-contained and claims its slot synchronously.
+    onStatusIndependent: (change) => {
       const sess = change.snapshot.session;
       if (change.status === "done" && sess.planPhase === "planning")
         void planGate.consider(sess).catch((err) => console.warn("[plan-gate] consider:", err));
--- a/src/session-router.ts
+++ b/src/session-router.ts
@@ -9,16 +9,20 @@ export interface SessionConsumer {
 }
 
 export interface SessionRouterHooks {
-  /** Fired on a `status` change AFTER the consumer chain — an independent side-effect
-   *  (plan-gate). Reached regardless of any consumer throwing. */
-  onStatusSettled?: (change: Extract<SessionStateChange, { kind: "status" }>) => void;
+  /** Fired on a `status` change as an INDEPENDENT side-effect (plan-gate) — kicked off
+   *  BEFORE the awaited consumer chain so a slow or hanging consumer can never starve it
+   *  (#1193). It is synchronous (only kicks `void planGate.consider`), so firing it first
+   *  costs nothing and leaves the drain→autopilot ordering below intact. */
+  onStatusIndependent?: (change: Extract<SessionStateChange, { kind: "status" }>) => void;
 }
 
 /** The single "what happens next" seam. Builds the shared snapshot ONCE per change and
  *  dispatches `consumers` strictly in order, AWAITING each fully before the next — this
- *  is what kills the retire-vs-steer race (fire-and-forget could not). Each consumer (and
- *  each hook) is isolated in try/catch so one failure is logged and does NOT prevent the
- *  next consumer or the settled hook from running; ordering still holds on the success path. */
+ *  is what kills the retire-vs-steer race (fire-and-forget could not). The independent
+ *  status hook (plan-gate) fires BEFORE this chain, so it never waits on it (#1193). Each
+ *  consumer and the hook are isolated in try/catch so one failure is logged and does NOT
+ *  prevent the next consumer from running; the drain→autopilot ordering still holds on the
+ *  success path. */
 export class SessionRouter {
   constructor(
     private acc: SnapshotAccessors,
@@ -31,14 +35,19 @@ export class SessionRouter {
   async onStatus(id: string, status: string): Promise<void> {
     const change = buildSnapshot(this.acc, id, { kind: "status", status });
     if (!change) return;
+    // Plan-gate is an INDEPENDENT per-session side-effect — kick it BEFORE awaiting the
+    // drain→autopilot chain so a slow or hanging consumer (e.g. an unbounded drain pump)
+    // can never starve it (#1193). It only fires `void planGate.consider`, so running it
+    // first is free and does not perturb the ordered dispatch below.
+    this.safeSync(
+      () =>
+        this.hooks.onStatusIndependent?.(change as Extract<SessionStateChange, { kind: "status" }>),
+      "independent",
+    );
     // No up-front PR-prewarm hook: drain runs before autopilot, and autopilot.onDone
     // kicks refreshPr right before its classify, so the PR-poll↔classify overlap is
     // preserved without a separate kick (see autopilot.handle).
     await this.dispatch(change);
-    this.safeSync(
-      () => this.hooks.onStatusSettled?.(change as Extract<SessionStateChange, { kind: "status" }>),
-      "settled",
-    );
   }
 
   async onGit(id: string, git: GitState): Promise<void> {
--- a/test/process-reaper.test.ts
+++ b/test/process-reaper.test.ts
@@ -379,6 +379,13 @@ test("claude-alive: every supplied worktree appears as a key; empty input is fin
 
 const WT = "/home/u/Work/.shepherd-worktrees/repo-x"; // a real worktree path (has the marker)
 
+// A synthetic orphan PID guaranteed to differ from the test runner's own process.pid.
+// The reapers spare `p.pid === process.pid` (don't SIGKILL self); a hardcoded literal can
+// collide with the runner's PID in CI containers (which hand out small PIDs — a literal like
+// 808 once matched), sparing the orphan and flipping a "reaps 1" assertion to 0. Deriving it
+// from process.pid can never collide.
+const ORPHAN_PID = process.pid + 1;
+
 // killPid spy: record (pid, signal) so tests can assert SIGKILL + exactly-who.
 function killSpy() {
   const killed: { pid: number; signal?: NodeJS.Signals }[] = [];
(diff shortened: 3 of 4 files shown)


scoop (commented):
## Summary

This PR implements **Option 1** from issue #1193 (the preferred fix): the plan-gate side-effect is now fired *before* the awaited `dispatch()` chain instead of after it, so a slow or hanging drain pump can no longer starve auto plan-review. The change is minimal and surgical — rename `onStatusSettled` → `onStatusIndependent`, move its invocation above `await this.dispatch(change)`, update comments and tests, and add a regression test for the hanging-consumer case.

## Verification

- **Rename is complete, no dangling references.** Grepped the whole tree: `onStatusSettled` / `settled hook` appear nowhere outside `.git`; all five live references use `onStatusIndependent` (`src/index.ts:1246`, `src/session-router.ts:16,44`, `test/session-router.test.ts:162,188,216,245,263`). Verified via repo-wide grep returning no matches for the old name.
- **Hook is safe to fire first.** The hook body in `src/index.ts:1246-1250` only does `void planGate.consider(sess).catch(...)` — it returns synchronously and never awaits inside the router, so `safeSync` at `src/session-router.ts:42-46` keeps the move free of latency, exactly as the comment claims.
- **`consider()` claims its slot synchronously.** Confirmed against `src/plan-gate.ts:217-238`: the guards (`planPhase`, `inflight`/`starting`) and `this.starting.add(session.id)` (line 225) all run *before* the first `await` (`hashPlan`, line 227). So firing the hook before the drain pump introduces no new double-spawn race — the synchronous claim is intact regardless of ordering.
- **Ordering invariant preserved.** `dispatch()` (`src/session-router.ts:59-67`) still awaits drain then autopilot in order; only the independent plan-gate moved out of that sequence. This matches the issue's requirement to keep the drain→autopilot retire-vs-steer ordering while restoring the pre-`9124026a` plan-gate independence.
- **Snapshot semantics unchanged.** The hook reads `change.snapshot.session` built once at `src/session-router.ts:36`; moving the call earlier reads the same snapshot, and the drain pump does not write the plan file `consider()` reads, so there is no before/after data dependency.
- **Tests match behavior.** The reordered assertions at `test/session-router.test.ts:222-226` (`["independent", "consumer:start"]` then `["independent", "consumer:start", "consumer:finish"]`) correctly encode "hook first, no trailing settled". The new test at lines 248-271 (`a never-resolving consumer does not starve the independent hook`) directly guards the #1193 regression.

## Nit

- `test/session-router.test.ts:175,188,196` — the counter variable is still named `settled` (and the throwing-consumer test still calls it `settled`) even though it now counts the independent hook. Cosmetic only; renaming to `independent`/`fired` would keep the test vocabulary consistent with the new hook name. Non-blocking.

No correctness, security, or quality defects found. The change satisfies the task.

scoop (commented):
## Verdict

The PR implements **option 1** from issue #1193 (decouple the plan-gate so it no longer waits on the repo-wide drain pump). The implementation is correct and the prior-revision nit is resolved.

### What the change does

- `src/session-router.ts:38-46` — the independent status hook (`onStatusIndependent`) is now kicked **synchronously via `safeSync` BEFORE** `await this.dispatch(change)` (line 50), instead of after the awaited consumer chain. This is exactly the fix the issue calls for: a slow/hanging drain pump can no longer starve the plan-gate.
- `src/session-router.ts:11-17` — hook renamed `onStatusSettled` → `onStatusIndependent`, with the doc comment updated to describe the new BEFORE-the-chain semantics.
- `src/index.ts:1246-1250` — the registration site renames the hook and updates its comment accordingly.

### Verification

- **Hook fires before the awaited chain and survives a hang** — verified against `src/session-router.ts:42-50`: the `safeSync(...)` call precedes `await this.dispatch(...)`, and `safeSync` (`src/session-router.ts:69-75`) is purely synchronous, so a never-resolving consumer cannot block it. Covered by the new test at `test/session-router.test.ts:248-268` (`a never-resolving consumer does not starve the independent hook`).
- **No stale identifier left behind** — `onStatusSettled` has zero remaining references tree-wide (grep), and `onStatusIndependent` is used consistently in `src/session-router.ts:16,44`, `src/index.ts:1246`, and `test/session-router.test.ts:162,188,216,245,263`.
- **Snapshot input to `consider()` is unchanged by the reordering** — `change` (and thus `change.snapshot.session`) is built once at `src/session-router.ts:36` before either the hook or dispatch, so firing the hook earlier passes the same `sess` argument it received before.
- **"claims its slot synchronously" comment is accurate** — verified against `src/plan-gate.ts:217-225`: `consider()` checks the `inflight`/`starting` guards and calls `this.starting.add(session.id)` before its first `await`, so the kicked-but-voided call claims its dedupe slot in the same synchronous tick. Async rejection is handled by the `.catch` at `src/index.ts:1249`.
- **Prior nit (stale `settled` local) addressed** — `test/session-router.test.ts` now uses `fired`/`independent`; grep confirms no `settled` identifier remains in that file.

### `test/process-reaper.test.ts` change

Independent CI-flakiness fix bundled on this branch (HEAD commit). `ORPHAN_PID = process.pid + 1` (line 388) is derived to never collide with the runner's own PID. Verified against the reapers' spare rule `p.pid === process.pid` at `src/process-reaper.ts:283,333,369,416,516`; the synthetic orphan's `comm: "yes"` is not in `AGENT_COMMS`, and each affected test's `scanProcs` returns only that single orphan, so `process.pid + 1` is never spared and never collides. Sound fix.

No blocking issues and no outstanding nits found.

PR #1197, opened 2026-06-28
fix(build-queue): suppress reconcile nudge during plan gate

scoop:
## Problem

The build-queue reconciliation nudge (`RECONCILE_STEER`) mis-fired at sessions still in the **plan gate**. During planning, Shepherd auto-approves the queue for visibility, so it sits `approved` with every step `pending`. The agent runs (grilling/planning/authoring the plan + queue) — tripping the `sawRunning` evidence gate — then settles idle awaiting plan review. The drift test `isQueueDrifted` (approved + pending + no active) is satisfied, so the nudge fired, telling the agent to mark a step `active`/`done` when **all-pending is the correct pre-execution state**. The agent correctly pushed back, proving the nudge fired pointlessly.

## Fix

Suppress the nudge while `session.planPhase === "planning"`. `sweep()` threads `planPhase` into `considerSession`, which returns **before** the status switch while planning:

- **`sawRunning` never arms** — `readyToNudge` ANDs on it, so this alone suppresses the nudge (planning-phase running isn't counted as execution-work-seen).
- **`idleSince` is reset** — idle time can't pre-arm a post-gate nudge.

The queue stays auto-approved (visibility intact); only the *nudge* is gated on execution having started. After the gate opens — `"executing"` via Go-click (`releasePlanGate`) or the PR-driven `advanceToExecutionOnPr` manual-steer path, or `null` when the gate is off — the guard is inert and a fresh execution-`running` tick re-arms normally.

### Manual-steer window (intentional)

`advanceToExecutionOnPr` (service.ts:2628) means code *can* execute while `planPhase === "planning"` (operator manually steers instead of clicking Go). The guard suppresses nudges across that window **by design** — the session is officially still in the gate, and the PR-open flip to `"executing"` is the deliberate re-arm point.

## Tests

- planning + drifted + settled idle → **no steer**
- planning suppresses, then `"executing"` nudges once **after a post-flip `running` observation** (load-bearing: `sawRunning` is never set during planning)
- Existing 11 tests unchanged (`session()` fixture leaves `planPhase` undefined).

`bun test ./test`: 5182 pass, 0 fail. `bun run lint`: clean.

[no-feature-entry] — server-only plumbing; `RECONCILE_STEER` is fixed English by existing precedent. No UI/i18n/glossary surface.

### DIFF
--- a/src/build-queue-reminder.ts
+++ b/src/build-queue-reminder.ts
@@ -17,11 +17,14 @@
 //   - `sawRunning` evidence gate — an approved+all-pending queue is indistinguishable from
 //     "just approved, not yet started", so we only nudge after observing the agent actually
 //     run, which also avoids colliding with the APPROVE_STEER "Begin now" steer.
+//   - plan-gate gate — while planPhase is "planning", an auto-approved queue is legitimately
+//     all-pending (no execution step has begun), so we never nudge; the gate's flip to
+//     "executing" is the re-arm point. See the considerSession guard for the full rationale.
 //   - once per idle episode + a per-session lifetime cap (runaway guard).
 // Agent-facing text is fixed English (precedent: APPROVE_STEER, AUTOPILOT_DIRECTIVE).
 
 import type { SessionStore } from "./store";
-import type { BuildQueue, SessionStatus } from "./types";
+import type { BuildQueue, Session, SessionStatus } from "./types";
 
 const DEFAULT_IDLE_THRESHOLD_MS = 120_000;
 const DEFAULT_MAX_NUDGES = 3;
@@ -107,7 +110,7 @@ export class BuildQueueReminderService {
     const live = new Set<string>();
     for (const s of this.deps.store.list({ activeOnly: true })) {
       live.add(s.id);
-      this.considerSession(s.id, s.status, now);
+      this.considerSession(s.id, s.status, s.planPhase, now);
     }
     // Forget sessions that are no longer active/listed.
     for (const id of [...this.debounce.keys()]) {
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

This PR suppresses the build-queue reconcile nudge while a session is still in the plan gate (`planPhase === "planning"`), where an auto-approved, all-pending queue is legitimately not-yet-started rather than drifted. The change is small, correct, and the rationale is documented thoroughly.

## Verification

- `planPhase` type resolves to `"planning" | "executing" | null` — verified against `src/types.ts:56`. The newly-added `Session` import (`src/build-queue-reminder.ts:27`) is used in the `considerSession` signature (`src/build-queue-reminder.ts:125`).
- The new guard returns **before** the status switch (`src/build-queue-reminder.ts:144-147`), so a running tick during `planning` never reaches `e.sawRunning = true` (`src/build-queue-reminder.ts:152`). Since `readyToNudge` ANDs on `sawRunning` (`src/build-queue-reminder.ts:177`), suppression holds even after the gate flips to `executing` until a fresh execution-running tick arms the evidence gate — exactly what the second new test asserts (`test/build-queue-reminder.test.ts:158-179`).
- `resetEpisode` clears only `idleSince`/`firedThisEpisode` (`src/build-queue-reminder.ts:98-101`), not `sawRunning`/`nudgeCount`, so idle accrual cannot leak across the gate — consistent with the comment's claim.
- The comment's referenced re-arm point `service.advanceToExecutionOnPr` exists — verified against `src/service.ts:2644` (and its caller `src/index.ts:967`).
- `considerSession` is private with a single caller (`sweep`, `src/build-queue-reminder.ts:113`); the signature change has no other consumers — verified via grep.
- Gate-off (`planPhase: null`) and `executing` sessions fall through to the prior behavior unchanged — no regression.
- Test harness/helper updates (`test/build-queue-reminder.test.ts:19-21,31,37`) correctly thread `planPhase` through `list()` and default it to `null`.

No correctness, security, or quality issues found. This is fixed English agent-facing text (no i18n catalog involved), so locale parity is not applicable.

Issue #1198, opened 2026-06-28
Add the Owed lens to the mobile herd tab row (HerdSegRow)


scoop:
Follow-up from the herd lens-strip redesign (`fix(ui): herd lens strip`).

The desktop/touch-wide herd lens strip (`HerdLensStrip`) exposes the **Owed** lens (post-merge manual steps still owed). The **mobile** tab row (`HerdSegRow`) does not — so on a phone there is currently no way to reach the Owed lens.

This was deferred from the redesign PR (operator decision) to keep that change focused and avoid adding height/width pressure to the mobile row without a dedicated look.

### Scope
- Add an "Owed" segment to `ui/src/lib/components/herd/HerdSegRow.svelte` (reuse `herd_seg_owed` + `herd_owed_title`, already in the catalogs; `use:coachTarget={"owed-lens"}`).
- Mind the mobile width budget (the block comment documents the sub-16px label rationale — with a 6th segment, re-check fit at the 390px reference and update the comment's math).
- No new i18n keys needed.

[no-feature-entry]

scoop:
✅ PR #1272 merged.

PR #1199, opened 2026-06-28
fix(ui): herd lens strip — icon-over-label tabs that fit the sidebar

scoop:
## Problem

The herd sidebar's lens filter row had grown to **7 lenses** (`Up Next · ▦ All · ▤ Ready · ⬡ Research · ✓ Done · ☰ Rundown · ☑ Owed`) and no longer fit the ~300–360px sidebar header. On desktop the row overflowed the panel — spilling "Owed" into the adjacent detail panel's tabs — and each button's glyph wrapped above its word (the broken stacked look in the report).

**Root cause:** `HerdFilterBar`'s `.filters` only enabled `flex-wrap` under `:global(.grid.compact)`, not standard desktop, so the row overflowed; and `.fbtn` had no `white-space: nowrap`, so squeezed buttons wrapped glyph-over-word.

## Change

Replace the inline text row (`HerdFilterBar`) with an **icon-over-label segmented strip** (`HerdLensStrip`) that becomes the panel's top edge, and **drop the static "The Herd" title**. Drop the **Research** lens to leave **6 lenses** that fit a **single row** with full, legible labels.

- **`HerdLensStrip.svelte`** — equal segments via `repeat(auto-fit, minmax(44px,1fr))`; the strip sets its **own** `container-type: inline-size` (the `herd` container is on the sibling `.units` subtree, unreachable) so a `@container` query shrinks labels at narrow widths and keeps all six lenses on one row down to the 300px floor. Glyphs are `aria-hidden` (each segment's accessible name is its label); the strip carries an `aria-label` since the visible title is gone.
- **Rehomed header bits:** the top-bar **status chip** → a thin row below the strip; the touch-wide **collapse ‹** → a slim row above it. Both reuse the existing markup/keys (`herd_status_chip_aria`, `aria-pressed`, `id="herd-collapse-btn"`, `herd_collapse`).
- **Research lens removed everywhere** — `herd-partition` (type + branch), `HerdSegRow`, and the dead `HerdEmptyState` research branch (falls through to `herd_ready_empty`). Research sessions still appear under **All**; the `s.research` attribute and the New-Task research toggle are untouched. Stale `HerdSegRow` block comment refreshed (now 5 segments; longest label DE "Nächstes").
- **i18n:** removed 10 orphaned keys (7 glyph-prefixed `herd_*_filter`/`herd_all_hint` + `herd_seg_research`/`herd_research_title`/`herd_research_empty`); added `herd_seg_owed` + `herd_lenses_label` (EN+DE, in parity).
- The Next-lens **label** renders as **"Next"** (`herd_seg_next`), a deliberate change from the old desktop "Up Next" (the long DE "Als Nächstes" would break single-row at 300px; matches the mobile row). "Up Next" remains the lens name + tooltip + ↑ glyph.

## ⚠️ Behaviour changes worth a reviewer's eye

- **Research lens dropped from the herd** — desktop users can no longer isolate research sessions via a lens (they remain visible under **All**). Operator-directed.
- **Desktop/mobile asymmetry:** the **Owed** lens is reachable on desktop but **not yet on mobile** (the mobile `HerdSegRow` keeps its existing lenses minus Research). Deferred by operator decision; tracked in #1198.

## Verification

- `cd ui && bun run check` → 0 errors / 0 warnings; `bun run check:i18n` → 2 locales in parity.
- `bun run test:node` → 1502 pass; `bun run test:browser` → all pass, incl. a new `HerdLensStrip` test asserting **single row + DE "Nächstes" label fit at the 300px floor**.
- eslint + prettier clean on touched files; branch-hygiene / feature-catalog / glossary gates pass.

[no-feature-entry] — layout fix of an existing control, no new user-facing capability.


scoop (commented):
## Summary

Replaces the overflowing inline `HerdFilterBar` with a new icon-over-label `HerdLensStrip` for the desktop/touch-wide sidebar header, drops the `research` lens entirely, and rehomes the status chip and collapse control to thin rows around the strip. The mobile `HerdSegRow` loses only its research segment. This directly addresses the broken sidebar-header layout.

I verified the change end-to-end and found nothing blocking.

## Verification

- **Message-key parity (en + de).** Every key the new strip references exists in both catalogs: `herd_seg_owed`/`herd_lenses_label` added to both (en.json:1475-1476, de.json:1475-1476); `herd_seg_next/all/ready/done/rundown` present in both (en.json:1471-1474,2069; de.json:1471-1474,2069); all `*_title`, `herd_collapse`, `topbar_tally_clear_title`, `herd_status_chip_aria` present in de.json:212-365,1291,1389,1922,2068. The removed keys (`herd_all_hint`, `herd_ready_filter`, `herd_research_filter/title/empty`, `herd_done_filter`, `herd_rundown_filter`, `herd_owed_filter`, `herd_next_filter`, `herd_seg_research`) were dropped from BOTH en.json and de.json symmetrically, and a tree-wide grep finds zero remaining references to any of them.
- **`research` lens removal is consistent.** `HerdFilter` no longer contains `"research"` (herd-partition.ts:59); `shownSessions` dropped its `research` branch (herd-partition.ts); no code references `filter === "research"` anywhere (grep of all .ts/.svelte). `herdFilter` is plain `$state("all")` (+page.svelte:955), not persisted, so no stale-value hazard. The `s.research` session property and `ResearchBadge` are untouched — correct, only the lens filter is gone.
- **`HerdEmptyState` `filter` prop removal.** Both call sites in Herd.svelte (sessions/shown) dropped `{filter}`; the component body no longer references it; `mode="shown"` is only reachable for the `ready` filter now (all⇒shown==sessions caught by the earlier sessions.length===0 branch), so the unconditional `herd_ready_empty` is correct (Herd.svelte:461-484, HerdEmptyState.svelte:38-40).
- **No dangling component.** `HerdFilterBar.svelte` is deleted and has zero remaining imports/references; the feature-announcements comment was updated to `HerdLensStrip/HerdSegRow` (feature-announcements.ts:1739).
- **coachTarget anchors** (`up-next-lens`, `done-lens`, `herd-rundown`, `owed-lens`) are all carried into the new strip; `up-next-lens` also lives in the mobile seg row, but only one of strip/seg-row renders at a time (flow vs non-flow), so no duplicate-anchor collision.
- **Test label updates** match the new accessible names: the strip's glyph `<span class="ic" aria-hidden="true">` leaves each lens button's accessible name as its label alone, so `getByRole("button", {name: "Ready"/"Done", exact: true})` resolves correctly.

The container-query single-row / no-clip guarantees at the 300px floor are design assertions exercised by the new browser tests (HerdLensStrip.browser.test.ts) and cannot be confirmed by static inspection — left to CI.

No findings.

PR #1204, opened 2026-06-28
fix(settings): make mobile section dropdown recognizable

scoop:
## Problem

On mobile the Settings dialog collapses its section navigation into a native `<select>` (`.tab-select` in `Settings.svelte`). It set `appearance: none` but never replaced the browser's dropdown arrow — so the box (e.g. "Workspace") read like a heading or text field, and users didn't recognize it as a tappable section selector.

## Fix (Option A, operator-chosen)

Wrap the `<select>` in `.tab-select-wrap` and render the app's standard fold chevron (`▾`) via a wrapper `::after` (decorative, `pointer-events:none`, not announced to AT). A pseudo-element on the `<select>` itself wouldn't render, hence the wrapper.

On `.tab-select`:
- `box-sizing: border-box` — no global border-box reset exists; without it `padding-right` would overflow the wrapper and mis-place the chevron (mirrors `NewTask.svelte`).
- `padding-right: 32px` — value text clears the chevron.
- `min-height: 44px` — a11y tap-target floor; with border-box this renders an **exact** 44px.
- `-webkit-appearance: none` alongside `appearance: none` — avoids a native-arrow + custom-chevron double arrow on older iOS Safari.

Desktop (`≥768px`) `role="tablist"` strip is untouched.

## Notes
- Tokens only (chevron `var(--color-muted)`, `var(--fs-base)`) — works across all four themes.
- ≥16px text floor already enforced globally by `app.css:351-354`.
- No i18n key (decorative CSS `content` glyph; the select's `aria-label` already routes through `m.settings_tabs_aria()`).
- No feature-catalog entry — affordance polish on an existing control shipped as `fix:`, not `feat:`.

## Verification
- `cd ui && bun run check` → 0 errors, 0 warnings.
- `bunx prettier --check` + `bunx eslint` on the file → clean; pre-push hooks passed.
- Before/after render of the committed CSS confirms the chevron shows, value text clears it, and the control is an exact 44px.

### DIFF
--- a/ui/src/lib/components/Settings.svelte
+++ b/ui/src/lib/components/Settings.svelte
@@ -649,11 +649,13 @@
            full-screen card would clip the overflow), so the tabs collapse into a
            dropdown. The tablist isn't rendered here, so no tab carries a dangling
            aria-controls; the panels below stay labelled tabpanels. -->
-      <select class="tab-select" aria-label={m.settings_tabs_aria()} bind:value={tab}>
-        {#each visibleTabs as t (t.id)}
-          <option value={t.id}>{t.label()}</option>
-        {/each}
-      </select>
+      <div class="tab-select-wrap">
+        <select class="tab-select" aria-label={m.settings_tabs_aria()} bind:value={tab}>
+          {#each visibleTabs as t (t.id)}
+            <option value={t.id}>{t.label()}</option>
+          {/each}
+        </select>
+      </div>
     {:else}
       <div class="tabs" role="tablist" aria-label={m.settings_tabs_aria()}>
         {#each visibleTabs as t, i (t.id)}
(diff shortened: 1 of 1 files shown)


scoop (commented):
## Summary

The PR makes the mobile (narrow-viewport) settings tab `<select>` recognizable as a dropdown by wrapping it in `.tab-select-wrap` and rendering a custom chevron (`▾`) via the wrapper's `::after`, while suppressing the native arrow with `-webkit-appearance: none` in addition to the existing `appearance: none`. It also adds `box-sizing: border-box`, `min-height: 44px`, and right padding to clear the chevron.

This satisfies the task. The implementation is correct and idiomatic.

### Verification
- Glyph `▾` is the app's standard fold/chevron glyph, used consistently elsewhere — e.g. `ui/src/lib/components/EpicGroupHeader.svelte:42`, `ui/src/lib/components/LanguageSwitcher.svelte:35`, `ui/src/lib/components/IssueFilterPopover.svelte:103`. The choice is on-token and consistent.
- All CSS custom properties resolve: `--color-muted` (`ui/src/app.css:19`), `--fs-base` (`ui/src/app.css:83`); `--color-inset`/`--color-line-bright`/`--color-ink-bright` are already consumed by the pre-existing `.tab-select` rules in the same `<style>` block (`ui/src/lib/components/Settings.svelte:1156-1158`).
- Wrapper-based chevron is the correct technique: `::after` does not render on `<select>` (as the added comment notes), so the decorative glyph belongs on the wrapper, and `pointer-events: none` (`Settings.svelte:1150`) lets taps fall through to the select.
- `-webkit-appearance: none` (`Settings.svelte:1164`) is needed alongside `appearance: none` so iOS/Safari hides the native arrow rather than doubling it against the custom chevron.
- `box-sizing: border-box` + `width: 100%` keep the new `padding: 8px 32px 8px 10px` (`Settings.svelte:1161`) from overflowing the wrapper; the 32px right padding clears the chevron at `right: 12px` (`Settings.svelte:1146`). `min-height: 44px` gives a proper mobile touch target.
- No message catalog keys, component signatures, or callers are affected by this diff; markup change (`Settings.svelte:652-658`) only nests the existing select in a div.

No blocking issues found.

Issue #1124, opened 2026-06-26
Server-side plugin architecture for private/out-of-repo extensions


scoop:
## Summary

Introduce a **server-side, in-process plugin architecture** so private/proprietary extensions can run on a personal Shepherd instance **without living in the open-source repo**. Motivating example: a `claude-swap`-style credential/account switcher ([realiti4/claude-swap](https://github.com/realiti4/claude-swap)) that points each spawned agent at a different `CLAUDE_CONFIG_DIR` and rotates accounts on usage-hold events — useful to *me*, not appropriate for the public codebase.

The plugin **system** (loader, `ctx` API, status panel, docs, example) is public and documented; specific plugin **implementations** stay private under `~/.shepherd/plugins/`. The seam is deliberately shaped to be **marketplace-ready** (sole versioned chokepoint, declared capabilities) without building any marketplace machinery now.

Design settled via a `/grilling` session. Full design + rationale below.

---

### Execution model
- **In-process** modules `import()`-ed into the single Bun loop at boot — they need synchronous reach into the spawn path (the load-bearing use case), which out-of-process/IPC cannot do cheaply. Trusted, single-author code for now.
- **Guardrails** (the only isolation that matters for trusted code): per-plugin `try/catch` at load (one bad plugin never blocks boot or other plugins), every hook invocation wrapped + **timeout-bounded** (`Promise.race`, 5s default), all plugin errors logged to `shepherd.log` and surfaced in the status panel. Knowingly *not* protected against: a plugin's own synchronous infinite loop / CPU peg — that's a you-bug you'd fix, and worker/process isolation would sever the synchronous spawn-mutation seam claude-swap requires.
- The single-loop "no sync exec" discipline (see `single-loop-no-sync-exec`) is documented in the plugin contract — same rule core services follow.

### Location & loading
- Plugins live **outside the repo** in `~/.shepherd/plugins/` (env override `SHEPHERD_PLUGINS_DIR`, default `~/.shepherd/plugins/`), alongside the existing `~/.shepherd/` data dir. Physically impossible to leak into the public repo; no `.gitignore` hygiene; survives `bun run update` redeploys.
- Each plugin = a self-contained folder: manifest + entry module (+ its own `package.json`/`node_modules` if it needs deps; Bun resolves locally).
- Core scans the dir at boot, reads each manifest, `import()`s the entry, calls `register(ctx)` **once, after core services exist**.

### Entry contract & lifecycle
[...]

PR #1208, opened 2026-06-28
feat(plugins): fire onSpawn for review/plan-gate/doc-agent/standalone-critic spawns (#1205)

### LINKED ISSUE #937: refactor: extract shared spawn/membrane tail duplicated across reviewer-style services
## Problem

Fallow flags two clone groups (currently **excluded by the audit gate as inherited** findings) covering the spawn/membrane boilerplate copy-pasted across the four reviewer-style spawners. Surfaced during review of #930.

**Group A — `dup:7dd1b4cb` (23 lines × 4 instances):** the injectable-seam pair `detectBackend()` + `membraneEnv()` (sandbox backend detection + membrane env defaults, each `this.deps.X ?? real…` fallback).
- `src/doc-agent.ts`
- `src/plan-gate.ts`
- `src/review.ts`
- `src/standalone-critic.ts`

**Group B — `dup:47cc2630` (17 lines × 2 instances):** the `MembraneInputs` assembly + argv/sessionId for the spawn.
- `src/doc-agent.ts`
- `src/standalone-critic.ts`

(Line numbers drift; locate via the `dup:` hashes with `bunx fallow dupes` or by the symbol names above.)

## Why it matters

Four services hand-roll the same spawn plumbing (`detectBackend`, `membraneEnv`, `MembraneInputs` construction, `wrapArgv`, common-dir wiring). Each new spawner — and recent ones keep landing (doc-agent #907, plan-gate, critic) — re-copies it, so a fix to the membrane/egress posture must be applied N times and drifts silently between them.

## Proposal

Extract the shared seam pair + membrane assembly into one helper (e.g. `src/spawn-membrane.ts` — a `resolveMembrane({ worktreePath, repoPath, ... })` plus the `detectBackend`/`membraneEnv` defaults), and have all four services consume it. Keep the per-service injectable seams (tests rely on overriding `detectBackend`/`membraneEnv`) by routing the helper through the same `deps` hooks.

## Scope / notes

- Pure refactor — no behavior change. Existing per-service spawn tests should pass unchanged; add a focused test for the extracted helper.
- Pre-existing debt, not introduced by #930.
- Once extracted, drop the corresponding `dup:` exclusions from the fallow audit gate so the gate enforces the merged form going forward.

[no-feature-entry]

### LINKED ISSUE #1205: plugin onSpawn hook bypassed by review / plan-gate / doc-agent spawns (auto-processes use default account)
## Problem

The in-process plugin `onSpawn` hook (issue #1124) is the only seam a plugin has to influence
how an agent launches. It fires for **normal task create/drain and resume** — but **not** for
the "reviewer-style" auto-process spawns: **code review, plan-gate, and doc-agent**. Those build
their own argv and call `herdr.start()` directly, bypassing `prepareSpawn()` / the hook
dispatch entirely.

Concrete impact, reported against the [`shepherd-claude-swap`](https://github.com/erwins-enkel/shepherd-claude-swap)
plugin (which rotates agents across a pool of Claude accounts by injecting a per-account
`CLAUDE_CONFIG_DIR` via `onSpawn`): **auto-processes such as reviewing still run on the default
account.** The plugin literally cannot see those spawns, so it cannot route them onto a pool
account — they inherit whatever account `cswap` has active. Quota for review/plan/doc work all
piles onto one account instead of fanning out.

This is a **core** gap, not a plugin one: the plugin is already maximal for what core exposes.

## Root cause (evidence)

> ⚠️ The references below were read from a local clone at an **unpinned HEAD**; line numbers
> **will drift**. Re-locate each site by its **structural fingerprint**, not by line number:
> a `resolveSpawnMembrane({ argv, … })` call immediately followed by
> `herdr.start(…, apiKeyPassthroughEnv(backend !== null))`, with **no** intervening
> `runSpawnHookPatch` / `runSpawnHooks` call.

The hook is dispatched from exactly one place — `runSpawnHookPatch()` (~`src/service.ts:1281`),
called by `prepareSpawn()` (~`src/service.ts:1395`). Only create/drain and resume route through
`prepareSpawn`.

The aux spawns share the same structural shape and never fold in a `SpawnPatch`:

| Path | Approx location (verify by shape) |
|------|-----------------------------------|
| code review | ~`src/review.ts:346-369` (`criticArgv` + its own `criticSessionId`) |
| plan-gate | ~`src/plan-gate.ts:301-317` |
| doc-agent | ~`src/doc-agent.ts:702-714` (`buildTransientAgentArgv("doc", …)`) |

Each does: build argv → `resolveSpawnMembrane({ argv, … })` (~`src/spawn-membrane.ts:56`) →
`herdr.start(name, wt, wrapped, apiKeyPassthroughEnv(backend !== null))`.

Related: closed #937 ("extract shared spawn/membrane tail duplicated across reviewer-style
services") already identified this duplicated tail — that shared tail is exactly where the hook
dispatch is missing.

Also: `SpawnDescriptor` (~`src/plugins/types.ts`) carries **no field for the kind of spawn**, so
a plugin currently cannot distinguish a review spawn from a normal session even if the hook did
fire.

## ⚠️ Membrane binding risk — firing the hook is necessary but NOT sufficient

Under a sandbox/membrane spawn profile the injected per-account `credentialDir` is **not bound
into the jail** unless it is threaded through the membrane's `--setenv` — the agent would
otherwise see an empty config dir. (This is precondition #2 in the `shepherd-claude-swap`
README.) So a fix that merely makes `onSpawn` *run* for review can pass a "hook fired" smoke
test yet **still launch on the default account**.

Any fix here must merge the patched `CLAUDE_CONFIG_DIR` (and `patch.env`) **last over
`apiKeyPassthroughEnv(...)` before `resolveSpawnMembrane`**, so the value is carried *through*
the membrane, not just set on the outer (clearenv-wiped) environment.

## Scope

In scope: **review + plan-gate + doc-agent** (the three user-facing agent spawns that consume
real quota).

Out of scope (possible follow-up): the smaller transient LLM helpers — `namer-llm`, `distiller`,
`merge-suggest`, `prompt-recommend`, `usage-probe`, `recap`, `optimizer`, `autopilot-llm` — which
share the same bypass shape but are tiny.

## Proposed core change

1. **Factor the hook dispatch out of `prepareSpawn`** into a small reusable step that takes the
   built `argv` + the membrane/passthrough env and returns the `SpawnPatch`-folded
   `{ finalArgv, patchEnv }` (essentially today's `runSpawnHookPatch` body, decoupled from the
   rest of `prepareSpawn`). The shared reviewer-style spawn tail (see #937) is the natural home.
2. **Call it in the three aux spawn sites before `resolveSpawnMembrane`/`herdr.start`**: append
   `extraArgs` to `argv`, and merge `patchEnv` (and `credentialDir → CLAUDE_CONFIG_DIR`) **last**
   over `apiKeyPassthroughEnv(...)` so it is carried through the membrane (see binding risk).
3. **Honor `abortSpawn`**: if a hook aborts an aux spawn, skip it cleanly (reap the worktree,
   log) — mirror how create/resume already model an auto-refuse — rather than escaping as an
   unhandled throw.

### Optional / future (not required for the stated goal)

Add a `kind` field to `SpawnDescriptor` (e.g. `"session" | "review" | "plan-gate" | "doc"`) so
plugins can branch on spawn type. The pool-rotation goal needs **no** plugin branch (a review
spawn pool-assigns like any session), so this is optional — included only because it is cheap to
add alongside and would avoid a second contract change if the "reuse parent account" option
below is later pursued.

## Account-selection design fork (please decide as part of this work)

How should the plugin choose an account for an aux spawn once it can see one? Two viable
options, with real tradeoffs — the plugin author leans **A** but wants the side effects on the
table:

- **Option A — fresh pool pick.** The aux spawn presents its own one-shot sessionId; the plugin
  assigns least-used / round-robin like a new session. **Side effects (in the plugin's
  `selection.ts`):** each unique one-shot sessionId is written into the plugin's durable
  `assignments` map and **never GC'd → unbounded growth**; the round-robin **cursor advances by
  1 per spawn**, drifting distribution for real sessions; and a review of session X may land on a
  *different* account than X used. If A is chosen, the plugin will need a mitigation (don't
  persist ephemeral aux pins, or GC them) — i.e. this is **not** a no-op on the plugin side.
- **Option B — reuse the parent session's pinned account.** Resume the review/plan/doc spawn
  onto the *reviewed* session's already-pinned account (sticky). No durable-map growth, no cursor
  drift, and a session's work stays on one account. **Requires core to pass the parent
  sessionId** in the descriptor (review currently mints its own `criticSessionId` distinct from
  `session.id`) — which is also the main reason the optional `kind`/parent-id descriptor field
  would help.

## Acceptance criteria

1. review / plan-gate / doc-agent spawns invoke registered `onSpawn` hooks.
2. With `shepherd-claude-swap` installed, the spawned reviewer's **observed identity** —
   `claude auth status --json` email — is the **assigned pool account**, not the default. I.e.
   the patched `CLAUDE_CONFIG_DIR` is actually bound *through* `resolveSpawnMembrane`, verified
   by observed identity rather than just "the hook ran".
3. A hook that calls `abortSpawn` cleanly skips the aux spawn (worktree reaped, no unhandled
   throw).
4. The chosen account-selection option addresses its side effects (A: bounded/GC'd `assignments`
   + cursor handling; B: parent sessionId plumbed through the descriptor).

## Downstream plugin work (tracked separately, in `shepherd-claude-swap`)

Once core ships the above, the plugin confirms `assign()` behavior for the new spawn sessionIds
and — if Option A — implements the durable-map/cursor mitigation; if the optional descriptor
field lands, it re-vendors `types.ts`.

---

_Filed as a handoff from the `shepherd-claude-swap` plugin repo, where the limitation surfaced.
🤖 Generated with [Claude Code](https://claude.com/claude-code)_

scoop:
## What & why

Closes #1205. The in-process plugin `onSpawn` hook (#1124) previously fired **only** for normal task sessions (via `prepareSpawn`) and was bypassed by the four reviewer-style auto-process spawns, which built their own argv and called `herdr.start()` directly. As a result, plugins like [`shepherd-claude-swap`](https://github.com/erwins-enkel/shepherd-claude-swap) could not route review/plan/doc quota onto a pool account — it all piled onto the default account.

This makes **all four** reviewer-style spawns honor `onSpawn`, and — critically — binds a hook's patched `CLAUDE_CONFIG_DIR`/`patch.env` **through the sandbox membrane** (merged last into both the bwrap `--setenv` and the `herdr.start` env) so it isn't wiped by `--clearenv`.

## Approach

- **`SpawnDescriptor`** gains a required `kind` (`"session" | "review" | "plan-gate" | "doc"`) + optional `parentSessionId` (set for review/plan-gate; undefined for the session-less doc-agent + standalone critic). Lets a plugin do fresh-pool-pick (Option A) now or reuse-parent-account (Option B) later with no second contract change.
- **`src/spawn-membrane.ts`** gains the shared reviewer-style tail (the duplicated tail flagged by closed #937):
  - `foldSpawnPatch` — the single source for the binding-critical merge (`credentialDir`→`CLAUDE_CONFIG_DIR` last; `extraArgs` appended). `prepareSpawn` now reuses it too.
  - `resolveAuxSpawn` — fires `onSpawn`, folds the patch, binds the env **through** the membrane, returns `{ aborted }` on `abortSpawn`, and returns `spawnEnv` as `undefined` when empty (preserves `herdr.start`'s optional-env contract).
- **Four sites wired** (`review.ts`, `plan-gate.ts`, `standalone-critic.ts`, `doc-agent.ts`) + `runSpawnHooks` passed into their constructors in `index.ts`. doc-agent's `spawnAgent`/`launchRun` became `async`; a hook abort yields a distinct `{ status: "skipped", reason: "plugin aborted spawn" }` (not the spawn-failure `error`).
- A hook `abortSpawn` cleanly skips each aux spawn: worktree reaped, no unhandled throw.

## Acceptance criteria

1. ✅ review / plan-gate / doc-agent / standalone-critic invoke `onSpawn`.
2. ✅ Patched `credentialDir` is bound **through** `resolveSpawnMembrane` — unit-proven via `wrapped.lastIndexOf("CLAUDE_CONFIG_DIR")` (rides the bwrap `--setenv`, last-wins) with a backend, and via `spawnEnv` last-wins without one. (Observed-identity end-to-end is verified downstream with the plugin installed.)
3. ✅ `abortSpawn` → worktree reaped, no spawn, no throw (per-site tests for review + doc-agent).
4. ✅ `SpawnDescriptor` carries `kind` + `parentSessionId`.

## Tests

`bun run lint`, `bun run typecheck`, and the root `bun test test/` suite are green (the only failures are the pre-existing, unrelated `config > persistConfig` Svelte-rune tests). New coverage: `foldSpawnPatch` + `resolveAuxSpawn` unit tests in `test/spawn-membrane.test.ts` (fire / membrane-binding w+wo backend / undefined-env / extraArgs / abort), plus per-site abort + hook-fires integration tests in `test/review.test.ts` and `test/doc-agent.test.ts`.

## Follow-up (not in this PR)

`doc-agent.ts` `beginRetarget` sets the `prSyncedKey` ownership claim **before** the spawn; **all** `!launched.ok` paths (worktree-fail, spawn-fail, and now abort) leave it set, permanently parking that PR's doc retarget. This is pre-existing behavior the abort path inherits; fixing only the abort arm would diverge from the identical failure paths, so it's deferred to a follow-up that addresses all of them. Also out of scope per the issue: the small transient LLM helpers (`namer-llm`, `distiller`, etc.).

[no-feature-entry] — server-only plumbing, no user-facing UI surface (feature-catalog/i18n/glossary gates N/A).

### DIFF
--- a/src/doc-agent.ts
+++ b/src/doc-agent.ts
@@ -13,8 +13,8 @@ import type { SessionStore } from "./store";
 import type { DocAgentOutcome, DocAgentRun, Session } from "./types";
 import type { SessionUsage } from "./usage";
 import { readSessionUsage } from "./usage";
-import { apiKeyPassthroughEnv, isApiKeyConfigured, isApiKeyMode } from "./spawn-auth";
-import { resolveSpawnMembrane, type MembraneSeams } from "./spawn-membrane";
+import { isApiKeyConfigured, isApiKeyMode } from "./spawn-auth";
+import { resolveAuxSpawn, type MembraneSeams } from "./spawn-membrane";
 import type { WorktreeMgr } from "./worktree";
 
 const execFileP = promisify(execFile);
@@ -517,7 +517,7 @@ export class DocAgentService {
       this.deps.store.setSetting(prSyncedKey(repoPath, prNumber), "1");
 
       const promptCtx: RetargetPromptCtx = { prNumber, prTitle: git.title ?? "" };
-      const launched = this.launchRun(
+      const launched = await this.launchRun(
         repoPath,
         headSha,
         base,
@@ -563,14 +563,14 @@ export class DocAgentService {
    * (retarget only) grounds the agent on the open PR's diff. Callers keep their own pre/post steps
    * (begin's stampLastSha, beginRetarget's prSyncedKey claim) outside this helper.
    */
-  private launchRun(
+  private async launchRun(
     repoPath: string,
     baseRef: string,
     base: string,
     extra: Partial<InFlight>,
     promptCtx?: RetargetPromptCtx,
     afterCreate?: (worktreePath: string) => void,
-  ): { ok: true } | { ok: false; result: DocAgentResult } {
+  ): Promise<{ ok: true } | { ok: false; result: DocAgentResult }> {
     const id8 = randomUUID().slice(0, 8);
     const wtName = DOC_BRANCH_PREFIX + id8;
     let wt;
@@ -587,13 +587,17 @@ export class DocAgentService {
 
     afterCreate?.(wt.worktreePath);
 
-    const spawned = this.spawnAgent(
+    const spawned = await this.spawnAgent(
       repoPath,
       wt.worktreePath,
       DOC_AGENT_LABEL + id8,
       base,
       promptCtx,
     );
+    if (spawned === "aborted") {
+      this.deps.worktree.remove(wt.worktreePath);
+      return { ok: false, result: { status: "skipped", reason: "plugin aborted spawn" } };
+    }
     if (!spawned) {
       this.deps.worktree.remove(wt.worktreePath);
       return { ok: false, result: { status: "error", reason: "spawn failed" } };
@@ -644,7 +648,7 @@ export class DocAgentService {
 
     // forget() can't race a manual trigger here, but a second trigger could have landed during the
     // awaits above — the `starting` claim (held until this returns) prevents that double-spawn.
-    const launched = this.launchRun(repoPath, resolved.baseRef, base, { mode: "fresh" });
+    const launched = await this.launchRun(repoPath, resolved.baseRef, base, { mode: "fresh" });
     if (!launched.ok) return launched.result;
     // Stamp the cadence marker from the SAME ref the nightly gate reads (`refs/remotes/origin/<base>`,
     // freshened by ensureBaseRef above) — NOT resolved.baseRef, which falls back to the branch
@@ -685,33 +689,44 @@ export class DocAgentService {
   }
 
   /** Build the scoped argv + membrane and spawn the agent via herdr. Returns the terminalId +
-   *  the agent's forced --session-id (the reviewer_spawns PK), or null on a spawn failure (logged).
-   *  profile "standard": the agent needs Anthropic egress (to run claude) but not GitHub — the
-   *  server pushes/opens the PR — mirroring ReviewService's spawn. */
-  private spawnAgent(
+   *  the agent's forced --session-id (the reviewer_spawns PK), or null on a spawn failure (logged),
+   *  or "aborted" when a plugin onSpawn hook aborts (distinct from a spawn failure). */
+  private async spawnAgent(
     repoPath: string,
     worktreePath: string,
     agentName: string,
     base: string,
     promptCtx?: RetargetPromptCtx,
-  ): { terminalId: string; spawnSessionId: string } | null {
+  ): Promise<{ terminalId: string; spawnSessionId: string } | null | "aborted"> {
     const { argv, sessionId } = buildTransientAgentArgv("doc", {
       model: this.deps.model ?? null,
       prompt: this.buildPrompt(base, promptCtx),
     });
-    const { wrapped, backend } = resolveSpawnMembrane({
+    // Fire plugin onSpawn hooks (issue #1205) + bind any patched env THROUGH the membrane.
+    // Session-less doc agent → no parentSessionId. abortSpawn → "aborted" so launchRun can reap
+    // the worktree and skip cleanly (distinct from a spawn failure).
+    const aux = await resolveAuxSpawn({
       argv,
       worktreePath,
       repoPath,
       worktree: this.deps.worktree,
       seams: this.deps,
+      descriptor: {
+        sessionId,
+        kind: "doc",
+        model: this.deps.model ?? null,
+      },
     });
+    if ("aborted" in aux) {
+      console.warn(`[doc-agent] onSpawn aborted for ${repoPath}: ${aux.aborted.reason}`);
+      return "aborted";
+    }
     try {
       const terminalId = this.deps.herdr.start(
         agentName,
         worktreePath,
-        wrapped,
-        apiKeyPassthroughEnv(backend !== null),
+        aux.wrapped,
+        aux.spawnEnv,
       ).terminalId;
       return { terminalId, spawnSessionId: sessionId };
     } catch (err) {
--- a/src/index.ts
+++ b/src/index.ts
@@ -770,6 +770,8 @@ const reviewService = new ReviewService({
   herdr,
   worktree,
   resolveForge,
+  // Plugin onSpawn hooks fire for reviewer-style aux spawns too (issue #1205); no-op until loadAll.
+  runSpawnHooks: (d) => pluginRegistry.runSpawnHooks(d),
   onChange: (id, verdict) => events.emit("session:review", { id, review: verdict }),
   onReviewing: (id, reviewing) => events.emit("session:reviewing", { id, reviewing }),
   onActivity: (id, summary) => events.emit("session:critic-activity", { id, summary }),
@@ -792,6 +794,8 @@ const standaloneCritic = new StandalonePrCriticService({
   herdr,
   worktree,
   resolveForge,
+  // Plugin onSpawn hooks fire for reviewer-style aux spawns too (issue #1205); no-op until loadAll.
+  runSpawnHooks: (d) => pluginRegistry.runSpawnHooks(d),
   repos: () => listRepos(config.repoRoot).map((r) => r.path),
   // Fresh per-sweep thunk (the service calls it each sweep, never caches) — branches
   // owned by a LIVE session, so a session-critic-owned PR is skipped when criticEnabled.
@@ -816,6 +820,8 @@ const planGate = new PlanGateService({
   herdr,
   worktree,
   resolveForge,
+  // Plugin onSpawn hooks fire for reviewer-style aux spawns too (issue #1205); no-op until loadAll.
+  runSpawnHooks: (d) => pluginRegistry.runSpawnHooks(d),
   reply: (id, text) => service.reply(id, text),
   release: (id) => service.releasePlanGate(id),
   onChange: (id, gate) => events.emit("session:plangate", { id, gate }),
--- a/src/index.ts
+++ b/src/index.ts
@@ -748,6 +748,8 @@ const docAgent = new DocAgentService({
   herdr,
   worktree,
   resolveForge,
+  // Plugin onSpawn hooks fire for the doc-agent spawn too (issue #1205); no-op until loadAll.
+  runSpawnHooks: (d) => pluginRegistry.runSpawnHooks(d),
   repos: () => listRepos(config.repoRoot).map((r) => r.path),
   store,
   gitState: (id) => prPoller.get(id),
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -14,10 +14,10 @@ import {
   groundPlanBlocks,
 } from "./visual-blocks";
 import { buildTransientAgentArgv } from "./transient-agent-argv";
-import { isApiKeyMode, isApiKeyConfigured, apiKeyPassthroughEnv } from "./spawn-auth";
+import { isApiKeyMode, isApiKeyConfigured } from "./spawn-auth";
 import { readSessionUsage, type SessionUsage } from "./usage";
 import { effectiveAutopilot } from "./effective-autopilot";
-import { resolveSpawnMembrane, type MembraneSeams } from "./spawn-membrane";
+import { resolveAuxSpawn, type MembraneSeams } from "./spawn-membrane";
 
 /** Outcome of an on-demand `consider()`: a reviewer actually spawned, the request was a no-op
  *  (plan unchanged / already approved / nothing to review), or a spawn attempt failed. The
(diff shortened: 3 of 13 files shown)


scoop:
## Summary

The PR factors the plugin `onSpawn` dispatch out of `prepareSpawn` into shared `foldSpawnPatch` + `resolveAuxSpawn` helpers (`src/spawn-membrane.ts`), threads patched env THROUGH the membrane via `extraEnv`, adds `kind`/`parentSessionId` to `SpawnDescriptor`, and routes review / plan-gate / doc-agent / standalone-critic through `resolveAuxSpawn` with clean `abortSpawn` handling. The membrane-binding design (patchEnv merged last over `apiKeyPassthroughEnv` into both the bwrap `--setenv` and the herdr env) directly addresses the issue's binding-risk precondition, and `spawn-membrane.test.ts` verifies it for both backend and passthrough cases.

The core mechanism is sound and well-tested — but there is one blocking wiring gap.

## Blocking

**`src/index.ts`: `DocAgentService` is never given `runSpawnHooks`, so the doc-agent path — explicitly in scope (issue scope + acceptance criterion #1) — does not fire `onSpawn` in production.**

The three other aux services each receive the hook runner:
- `reviewService` — index.ts:774 `runSpawnHooks: (d) => pluginRegistry.runSpawnHooks(d)`
- `standaloneCritic` — index.ts:798 (same)
- `planGate` — index.ts:824 (same)

But the `DocAgentService` constructor (index.ts:747-758) has no `runSpawnHooks` field. `DocAgentDeps extends MembraneSeams` (src/doc-agent.ts:172) where `runSpawnHooks?` is optional (src/spawn-membrane.ts:31), and `spawnAgent` passes `seams: this.deps` (src/doc-agent.ts:101). With the field absent, `resolveAuxSpawn` hits `if (args.seams.runSpawnHooks)` (src/spawn-membrane.ts:428) as falsy and skips the hook entirely — so `onSpawn` never fires for doc spawns, and a pool-rotation plugin still cannot route doc-agent quota. This defeats the stated goal for one of the three in-scope paths.

The doc-agent abort test passes only because `mkHarness` injects `runSpawnHooks` straight into deps (test/doc-agent.test.ts:452), masking the missing production wiring. Fix: add `runSpawnHooks: (d) => pluginRegistry.runSpawnHooks(d)` to the `DocAgentService({...})` constructor at index.ts:747.

## Verification notes

- `foldSpawnPatch` (src/spawn-membrane.ts:383) reproduces the prior `prepareSpawn` merge exactly; service.ts:1326 now reuses it — verified the old inline body (removed at the diff's service.ts hunk) matched field-for-field.
- patchEnv binding through the membrane verified against src/service.ts:1416 (session path: `extraEnv: {...passthroughEnv, ...rendererEnv, ...patchEnv}`) and src/spawn-membrane.ts:371/455/459 (aux path: `extraEnv: {...env.extraEnv, ...args.extraEnv}` plus `merged = {...baseEnv, ...patchEnv}`). credentialDir→CLAUDE_CONFIG_DIR last in both.
- `apiKeyPassthroughEnv(backend !== null)` semantics preserved: src/spawn-auth.ts:77 returns undefined when wrapped (backend present), so baseEnv is only set on the passthrough path — matches the in-place membrane masking comment.
- `spawnEnv` undefined-when-empty contract (src/spawn-membrane.ts:460) preserves herdr.start's optional-env behavior; tested at spawn-membrane.test.ts:716.
- `SpawnDescriptor.kind` is now required (src/plugins/types.ts:222); all in-diff literal construction sites updated (service.ts:1310, examples-plugin.test.ts:16, plugins.test.ts:12). resolveAuxSpawn builds the descriptor inline with `kind` (src/spawn-membrane.ts:432).
- Abort handling verified per path: review reaps + returns (src/review.ts:273-277), plan-gate reaps + returns "skipped" (src/plan-gate.ts:192-196), standalone-critic reaps + returns (src/standalone-critic.ts:504-508), doc-agent returns "aborted"→launchRun reaps + "skipped" (src/doc-agent.ts:108-111, 598-600). `DocAgentStatus` includes "skipped" (src/doc-agent.ts:130). All `launchRun`/`spawnAgent` calls are awaited (src/doc-agent.ts:520,590,651).

Nit (non-blocking): doc-agent uses `console.warn` for the abort log (src/doc-agent.ts:109) while standalone-critic uses `this.log` (src/standalone-critic.ts:505); minor inconsistency, not a defect.

scoop:
**Point 1 (doc-agent `runSpawnHooks` wiring): fixed** in bcfb541c — added `runSpawnHooks: (d) => pluginRegistry.runSpawnHooks(d)` to the `DocAgentService` constructor in src/index.ts. Good catch; the service test injected it via the harness so the gap was invisible to unit tests. typecheck + doc-agent tests + lint green.

**Point 2 (abort log routing): not changing.** `this.log` is *standalone-critic's* convention — `doc-agent.ts` has no `this.log` method and uses `console.warn` at all ~20 of its log sites (e.g. the adjacent spawn-failure paths at doc-agent.ts:734/736, plus :369/:392/:580/:668/:753…). The abort log at :721 using `console.warn` is therefore consistent *within* doc-agent; switching it to `this.log` would be inconsistent (and wouldn't compile). Each service self-consistently uses its own logger — review/plan-gate use `console.warn`, standalone-critic uses `this.log` — so the abort logs already match their respective files.

scoop (commented):
## Summary

This revision routes all four reviewer-style aux spawns (review, plan-gate, doc-agent, standalone-critic) through the new shared `resolveAuxSpawn()` tail, which fires `onSpawn`, folds the returned `SpawnPatch`, and binds the patched env *through* the membrane. The two prior-round findings are resolved.

### Prior findings — both resolved

1. **doc-agent `runSpawnHooks` wiring (was blocking):** FIXED. `src/index.ts:131-132` now adds `runSpawnHooks: (d) => pluginRegistry.runSpawnHooks(d)` to the `DocAgentService` constructor, matching review (index.ts:140), standalone-critic (index.ts:149), and plan-gate (index.ts:158). Verified `DocAgentDeps extends MembraneSeams` (src/doc-agent.ts:172) and `MembraneSeams.runSpawnHooks?` exists (src/spawn-membrane.ts:360), so `resolveAuxSpawn` reads the wired seam in production. The hook now actually fires for doc-agent spawns.
2. **abort log routing nit:** Verified NON-APPLICABLE, not re-raised. `doc-agent.ts` has no `this.log` method; it uses `console.warn` at every one of its ~24 log sites (verified src/doc-agent.ts:369,392,580,668,734,736,753, etc.). The abort log at src/doc-agent.ts:721 using `console.warn` is therefore self-consistent within the file; switching to `this.log` would not compile. The original cross-file-consistency nit rested on a wrong premise.

### Correctness verification

- **Hook dispatch factored + reused without drift:** `foldSpawnPatch` (src/spawn-membrane.ts:392) is byte-for-byte equivalent to the old inline merge in `prepareSpawn`; service.ts:1336 now calls it, so the session path and aux path can't diverge. Verified the old body (env spread + credentialDir-wins + extraArgs append) matches exactly.
- **Membrane binding (acceptance #2):** `resolveAuxSpawn` passes `extraEnv: patchEnv` into `resolveSpawnMembrane`, which merges it LAST into the bwrap `--setenv` (src/spawn-membrane.ts:81 `{...env.extraEnv, ...(args.extraEnv ?? {})}`), AND merges `patchEnv` LAST over `apiKeyPassthroughEnv` into `spawnEnv` (src/spawn-membrane.ts:168). So a patched `CLAUDE_CONFIG_DIR` rides through `--clearenv`. Covered by test/spawn-membrane.test.ts:760 (backend present → rides `--setenv`) and :782 (passthrough → in spawnEnv).
- **abortSpawn handling (acceptance #3):** All four sites catch `{aborted}`: review returns + reaps (src/review.ts:369-373, mirrors the spawn-failure path at 382-386); standalone-critic reaps + returns (src/standalone-critic.ts:513-517); plan-gate reaps + returns `"skipped"` (src/plan-gate.ts:201-205) — `"skipped"` is a valid `PlanReviewTrigger` already used at plan-gate.ts:288; doc-agent reaps + returns `{status:"skipped", reason:"plugin aborted spawn"}` via the new `"aborted"` sentinel from `spawnAgent` (src/doc-agent.ts:599) — `"skipped"` is a valid `DocAgentStatus` (src/doc-agent.ts:130). `launchRun`/`spawnAgent` were correctly promoted to `async` and all callers awaited (begin/beginRetarget).
- **`SpawnDescriptor.kind` now required:** Verified no compile breakage — the only object-literal construction sites are service.ts:1308 (`kind:"session"`), `resolveAuxSpawn` (in diff), and the two updated test fixtures (examples-plugin.test.ts:16, plugins.test.ts:12). All other references (loader.ts, plugins-spawn.test.ts, example-plugin fixture, spawn-labeler example) only consume/type the descriptor, never build a literal.
- **No leftover imports:** Confirmed doc-agent, plan-gate, standalone-critic no longer reference the removed `resolveSpawnMembrane`/`apiKeyPassthroughEnv`; review.ts mentions `apiKeyPassthroughEnv` only in a comment. The remaining `apiKeyPassthroughEnv` callers (recap, distiller, namer-llm, etc.) are the explicitly out-of-scope transient helpers.
- **`spawnEnv` optional-env contract:** merged env collapses to `undefined` when empty (src/spawn-membrane.ts:169), preserving `herdr.start`'s optional-env behavior; matches old `apiKeyPassthroughEnv` semantics.

No blocking issues found.

- The smaller transient LLM helpers (namer-llm, distiller, merge-suggest, prompt-recommend, recap, optimizer, autopilot-llm, herd-digest) still bypass `onSpawn` via direct `apiKeyPassthroughEnv(false)` + `herdr.start` — explicitly listed as out of scope in the issue.

Issue #822, opened 2026-06-19
Recap stuck "generating" 5 min then fails: malformed JSON (unescaped quotes) is unrecoverable + can't fail fast


scoop:
## Summary

A session recap can sit at **“generating” for the full 5-minute timeout and then fail**, even though the recap agent produced a complete, correct summary. Root cause: the recap spawn occasionally writes **malformed JSON** (an unescaped `"` inside a string value), `JSON.parse` rejects it, and the pipeline can neither recover the content nor fail fast. PR #773 (native visual blocks, Phase 1–3) materially increased the odds by ballooning the JSON the agent must hand-author.

Observed live on **TASK-561** (`shepherd/lightweight-repo-mode`, the #807/#819 lightweight-repo-mode work) at v1.34.0.

## Evidence

Recap row for the session:

| field | value |
|---|---|
| `state` | `failed` |
| `spawnedAt → generatedAt` | **307s** (timeout is `DEFAULT_TIMEOUT_MS = 5 * 60_000` = 300s) |
| `body` / `blocks` | empty |
| `model` | sonnet |

The spawn transcript shows the agent **did finish and wrote a complete `.shepherd-recap.json`** — verdict `ready`, a good headline, and all 15 visual blocks. But the file does not parse:

```
$ node -e 'JSON.parse(fs.readFileSync("recap.json"))'
PARSE FAIL: Expected ',' or '}' after property value at position 441 (line 4)
```

Position 441 is in the `body` field. The agent embedded **bare unescaped double-quotes** inside JSON string values, e.g.:

```
"body": "... autopilot terminal step or operator "Open for merge" click) ..."
"markdown": "... by the operator clicking "Open for merge". Critic and ..."
```

Those inner `"` need to be `\"`. A single such slip invalidates the entire ~15 KB document.

## Why it hangs for 5 minutes, then fails

In `src/recap.ts`:

- `defaultReadVerdict(cwd)` does a raw `JSON.parse` and returns `null` on **any** error — a present-but-malformed file is indistinguishable from a not-yet-written file (`// partial write; retry next tick`).
- `tick()` only finalizes when `raw` is truthy **or** the spawn has exceeded `timeoutMs`. Since `raw` is forever `null` for an unparseable file, the row stays `generating` until the **hard 5-minute timeout**, then `finalize(r, null)` writes `state: "failed"` with empty body/blocks.
[...]

PR #558, opened 2026-06-12
fix(whatsnew): newest release first + mobile layout rework

kai-osthoff:
## Summary

Fixes the What's-New drawer sort order (user report: oldest release was on top on mobile) and reworks the drawer layout for small screens (iPhone 14, ~390px), based on a design-system audit.

### Sort order (data bug)
- `computeNewEntries` returned entries in catalog order; the catalog is append-only (oldest first), so the drawer led with the oldest release.
- Now sorted by `sinceVersion` descending (stable — entries within one release keep catalog order). Covered by a new unit test.

### Mobile layout (WhatsNew.svelte)
- **Release groups**: entries are grouped under one sticky release rail (amber version + localized date) instead of repeating the version/date meta line on every entry — follows the `LearningsDrawer` `.ghead` convention and keeps orientation while scrolling.
- **Flat list instead of boxes**: per-entry borders replaced by hairline separators (flat-panel doctrine), reclaiming ~30px of line width on a 390px viewport.
- **Type hierarchy**: 11px release rail (`--fs-meta`) / 16px entry title (`--fs-lg`, the modal section-head rung) / 13px body (`--fs-base`) — previously title and body were both 13px.
- **Mobile ergonomics**: safe-area insets (Dynamic Island / home indicator), ~44px touch targets for the close glyph and the dismiss action on coarse pointers, single scroll container with pinned header/footer.

All colors/sizes remain design-system tokens; no new i18n keys needed (version + date are data, not chrome).

## Verification
- `bun run check` — 0 errors, `bun run test` — 950/950 passed, `check:i18n` — parity ok
- Playwright screenshots at 390×844 (light + dark): newest release on top, sticky rail correct mid-scroll

No feature-catalog entry: `fix` commits, no new user-facing capability.

PR #1211, opened 2026-06-29
fix(review): fail-fast dead critic + surface onSpawn-abort reason

scoop:
## Problem

Operator reported: **can't start reviews for tasks — the review starts then fails** with `REVIEW ERR: critic did not produce a verdict`.

Traced end-to-end (server log + `shepherd.db` + critic transcripts) for the failing session (goal-changelog / PR #558, session `cad6bbca`):

- Critic spawn at **06:29:52** → **no transcript ever written** (the `claude` process died at startup) → error verdict persisted at **06:40:05**, exactly the **10-min hard timeout**.
- Subsequent auto-retries logged `[review] onSpawn aborted … no usable ready account available`.

Root cause is a tight **claude-swap** account pool: the recently-merged #1205 routes reviewer spawns through the plugin `onSpawn` hook, which aborts (or hands out a not-actually-ready account that dies at runtime) when no account is *warmed*. Two Shepherd-side behaviors made this confusing rather than clear:

1. A critic that exits **without writing a verdict** sat `REVIEWING` for the full 10-min timeout — `decideVerdictAction` treated an *absent* verdict as `wait` even once the spawn had finished (#822 left the "agent wrote nothing" class to the hard timeout).
2. An `onSpawn` abort failed **silently** (auto path) or as a generic toast (manual path) — the real reason ("no usable ready account available") only ever hit the server log.

This PR is the Shepherd-side **fail-fast + clear-error** half (the account-warming fix itself lives in the claude-swap plugin repo).

## Fix

**1. Fail-fast a dead reviewer/recap spawn.** `decideVerdictAction` gains a `graceElapsed` arg; an *absent* verdict now finalizes once the spawn is finished **AND** a 60s boot grace (`STARTUP_GRACE_MS`) has elapsed — turning the 10-min hang into ~60s. A slow-booting agent (not-yet-`working`, no file yet) is still protected during the grace; a genuinely-hung agent still reads `working` → not finished → still waits the hard timeout. Shared helper, so `recap` gets the same fix.

**2. Surface the onSpawn-abort reason.** A reviewer `onSpawn` abort now persists a visible `error` verdict carrying the reason → badge reads `REVIEW ERR: no usable ready account available` instead of a silent/generic failure. Deduped per `(headSha, reason)` so a cold pool doesn't churn `putReview`/`onChange`; not findings-bearing and preserves prior streak/error counters, so it neither trips the spawn ceiling nor escalates the consecutive-error stall. Self-heals — the next `consider()` that lands a real verdict overwrites it.

## Tests (TDD)

- `json-tolerant.test.ts`: absent + finished + grace → `finalize-null`; within-grace and still-working → `wait`; back-compat 3-arg default preserved.
- `review.test.ts`: onSpawn abort persists an `error` verdict with the reason (`findings:[]`, `patchId:""`); a repeated abort for the same head fires `onChange` once (no churn).

Gates: root `lint` + `tsc` + `bun test ./test` (5238 pass, 0 fail) all green. Server-only `fix` — no UI/i18n/feature-catalog changes.

## Not covered (follow-ups)

- The actual account-warming behavior is in the **claude-swap plugin** (separate repo) — when an account is usable but not yet "ready", reviewer spawns should warm/inherit rather than abort.
- `plan-gate` has the identical `onSpawn`-abort surface (`[plan-gate] onSpawn aborted …`) and its own verdict loop; left out to keep this PR scoped to reviews.

### DIFF
--- a/src/json-tolerant.ts
+++ b/src/json-tolerant.ts
@@ -76,6 +76,15 @@ export function isSpawnWorking(
 export type VerdictAction = "finalize-value" | "finalize-null" | "wait";
 
 /**
+ * Boot grace before an absent-verdict spawn may be fail-fasted as finished-without-output. Covers
+ * the window where a freshly-spawned agent reads as not-yet-working with no verdict file yet (it
+ * hasn't taken its first turn), which is indistinguishable from one that died at startup. Well under
+ * both the recap (5 min) and critic (10 min) hard timeouts, so the timeout still backstops a spawn
+ * that finishes within the grace.
+ */
+export const STARTUP_GRACE_MS = 60_000;
+
+/**
  * Shared finalize-gate decision for the recap (src/recap.ts) and critic (src/review.ts) verdict
  * loops, given the 3-way read plus the spawn/timeout state:
  *   - strict parse      → "finalize-value" now (a complete document; unchanged behavior).
@@ -83,13 +92,22 @@ export type VerdictAction = "finalize-value" | "finalize-null" | "wait";
  *                         "wait" — a repaired parse may be a truncated partial write jsonrepair
  *                         closed up, so it isn't trusted while the agent may still be writing.
  *   - unparseable       → "finalize-null" (fail fast) once finished (or timed out); else "wait".
- *   - absent            → "finalize-null" only at the hard timeout; else "wait". The "agent wrote
- *                         nothing" / genuine-hang class is deliberately NOT fail-fasted (#822).
+ *   - absent            → "finalize-null" at the hard timeout, OR once the spawn has finished AND
+ *                         the startup grace has elapsed (`graceElapsed`); else "wait". A spawn that
+ *                         exits without ever writing the verdict (e.g. the agent died at startup —
+ *                         no usable account) is finished-with-an-absent-file from its first tick;
+ *                         #822 left this class to the hard timeout, so it sat "working" for the full
+ *                         window. The grace gate reclaims that wait while still protecting a
+ *                         slow-booting agent: before its first turn an agent reads as not-working
+ *                         with no file yet, indistinguishable from a dead one, so we only fail-fast
+ *                         once it's both finished AND past the boot grace. A genuinely-hung agent
+ *                         (alive, mid-turn) reads `working` → not finished → still waits the timeout.
  */
 export function decideVerdictAction(
   read: VerdictRead<unknown>,
   spawnFinished: boolean,
   timedOut: boolean,
+  graceElapsed = false,
 ): VerdictAction {
   if (read.status === "parsed") {
     if (!read.repaired) return "finalize-value";
@@ -98,5 +116,5 @@ export function decideVerdictAction(
   if (read.status === "unparseable") {
     return spawnFinished || timedOut ? "finalize-null" : "wait";
   }
-  return timedOut ? "finalize-null" : "wait"; // absent
+  return timedOut || (spawnFinished && graceElapsed) ? "finalize-null" : "wait"; // absent
 }
--- a/src/recap.ts
+++ b/src/recap.ts
@@ -35,7 +35,12 @@ import {
   needsRecap,
 } from "./recap-core";
 import { groundBlocks, type VisualBlock } from "./visual-blocks";
-import { tolerantParseJson, isSpawnWorking, decideVerdictAction } from "./json-tolerant";
+import {
+  tolerantParseJson,
+  isSpawnWorking,
+  decideVerdictAction,
+  STARTUP_GRACE_MS,
+} from "./json-tolerant";
 import type { VerdictRead } from "./json-tolerant";
 
 const execFileAsync = promisify(execFile);
@@ -522,8 +527,14 @@ export class RecapService {
       if (this.finalizing.has(r.sessionId)) continue;
 
       const read = this._readVerdict(r.cwd);
-      const timedOut = this.now() - r.spawnedAt > this.timeoutMs;
-      const action = decideVerdictAction(read, this.spawnFinished(r.cwd), timedOut);
+      const elapsed = this.now() - r.spawnedAt;
+      const timedOut = elapsed > this.timeoutMs;
+      const action = decideVerdictAction(
+        read,
+        this.spawnFinished(r.cwd),
+        timedOut,
+        elapsed > STARTUP_GRACE_MS,
+      );
       if (action === "wait") continue; // not-yet-written / repaired-or-unparseable while still working
       // finalize-value carries the parsed verdict; finalize-null (timeout / fail-fast) → `failed`.
       const raw = action === "finalize-value" && read.status === "parsed" ? read.value : null;
@@ -537,7 +548,7 @@ export class RecapService {
           );
         } else {
           console.warn(
-            `[recap] ${r.sessionId}: no verdict file after ${Math.round(this.timeoutMs / 1000)}s — agent produced nothing.`,
+            `[recap] ${r.sessionId}: no verdict file after ${Math.round(elapsed / 1000)}s (spawn exited or hard timeout) — agent produced nothing.`,
           );
         }
       }
--- a/src/review.ts
+++ b/src/review.ts
@@ -10,7 +10,7 @@ import { isApiKeyMode, isApiKeyConfigured } from "./spawn-auth";
 import { jsonlPathFor, readSessionUsage, type SessionUsage } from "./usage";
 import { readActivitySignal } from "./activity-signal";
 import { checksCleared } from "./checks-gate";
-import { isSpawnWorking, decideVerdictAction } from "./json-tolerant";
+import { isSpawnWorking, decideVerdictAction, STARTUP_GRACE_MS } from "./json-tolerant";
 import type { VerdictRead } from "./json-tolerant";
 import {
   reviewPrompt,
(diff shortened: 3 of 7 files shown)


scoop:
## Summary

The PR has two parts: (1) a `STARTUP_GRACE_MS` gate that fail-fasts an absent verdict once the spawn has both finished and passed a 60s boot grace, and (2) `publishSpawnAbort`, which surfaces an `onSpawn` abort (e.g. claude-swap pool has no usable account) as a visible `error` verdict instead of a silent failure. Part (1) is sound. Part (2) has a blocking interaction with the existing same-head dedup that undermines the task it is meant to fix.

## Blocking

**src/review.ts: `publishSpawnAbort` writes an `error` verdict keyed to the current head, which the same-head dedup at `consider()` then treats as "already reviewed" — so the auto path never re-attempts the spawn for that head, even after the transient pool-cold condition clears.**

- `publishSpawnAbort` persists a verdict with `headSha: git.headSha!` (src/review.ts:968) and `decision: "error"` (src/review.ts:970).
- On the next poll tick, `consider()` re-reads that verdict and bails at the unconditional same-head guard: `if (!force && prior?.headSha === git.headSha) return "skipped"; // head already reviewed` (src/review.ts:236). This guard fires *before* `begin()`/`rebaseSkip`, and it is not decision-aware, so an `error` verdict for the current head permanently parks the auto path on that head.
- This directly contradicts the in-code intent. The `patchId: ""` comment states "a later identical head must re-attempt, not inherit this abort" (src/review.ts:969), but `patchId` is only consulted by the churn-skip in `rebaseSkip`, which is never reached because line 236 returns first. The doc comment "Self-heals — the next consider() that lands a real verdict overwrites it" (src/review.ts:375) is likewise not achievable for the same head: `consider()` never reaches the spawn again, so no real verdict lands until the head moves.
- Behavior change vs. base: before this PR the abort path was `warn` + `worktree.remove` + `return` with **no** verdict written (verified against `git show 4d48948b:src/review.ts:370-372`). With no verdict, `consider()` retried the spawn on every subsequent poll tick and recovered automatically once the pool warmed. After this PR, the first abort writes a head-matching `error` verdict and every later tick on the same head is skipped at line 236 — recovery now requires a new push (new head) or a manual `forceReview`.
- Impact on the task ("Can't start reviews for tasks 1005 and 1015"): a pool-cold / no-usable-account refusal is transient, but the head a developer is waiting on typically does not move. This change makes the inability-to-review *sticky* for that head: REVIEW ERR is shown but the review will not auto-start when accounts become available again. The abort case is a pre-spawn refusal and should re-attempt on the same head (as the author's own `patchId` comment says), unlike a genuine post-run timeout where same-head re-review is correctly suppressed.

Suggested directions (author's choice): make line 236 (or the abort verdict) not block same-head re-attempt for a pre-spawn refusal — e.g. don't write a head-matching verdict and instead surface the reason via a transient/non-dedup channel, drop the error verdict on the next consider so the spawn re-attempts, or special-case `decision === "error"` from an abort in the same-head guard. (The orphan-reaper at src/review.ts:1096-1098 already drops `error` verdicts precisely "so ... the re-kick's plain consider() isn't head-deduped" — that same hazard applies here, but the abort path removes the worktree so the reaper never fires for it.)

## Nit

**test/review.test.ts: the new "repeated onSpawn abort for the same head does not churn" test passes because of the line-236 head-dedup, not because of the `(headSha, reason)` dedup inside `publishSpawnAbort`.** The second `svc.consider(session(), OPEN_GREEN)` (both use head `"abc"`, test/review.test.ts:104) bails at src/review.ts:236 before `begin()` is reached, so `publishSpawnAbort` runs only once and `changes` is 1 regardless of the in-method dedup. That dedup branch (src/review.ts:959-965) is therefore only reachable on the `force` path and is effectively untested, and the comment's "a pool that stays cold across poll ticks does not churn ... every tick" rationale (src/review.ts:946-948) is inaccurate — poll ticks are already short-circuited earlier by line 236. (This is also corroborating evidence for the blocking finding: the test demonstrates the same head is silently skipped on the second consider.)

## Verified, no issue

- `PluginSpawnAborted.reason` exists and is the abort message (src/plugins/types.ts:127-135); `resolveAuxSpawn` returns `{ aborted: PluginSpawnAborted }` on a hook abort (src/spawn-membrane.ts:129, :153), so `aux.aborted.reason` (src/review.ts:376) is valid.
- `decision: "error"` is a member of `ReviewDecision` (src/types.ts:321); the `ReviewVerdict` shape constructed in `publishSpawnAbort` sets every required field (verified field-by-field against src/types.ts:323-341).
- Preserving (not incrementing) `errorRound`/`streakReviews`/`reviewedPatchIds` (src/review.ts:976-978) is consistent with the documented intent that an abort neither trips the 2*cap spawn ceiling (src/review.ts:262) nor escalates the consecutive-error stall (src/review.ts:774).
- `decideVerdictAction`'s new `graceElapsed` param defaults to `false` (src/json-tolerant.ts:97), preserving the prior "wait until hard timeout" behavior for any caller that omits it; both call sites pass `elapsed > STARTUP_GRACE_MS` (src/recap.ts:534, src/review.ts:640). The absent-branch logic `timedOut || (spawnFinished && graceElapsed)` (src/json-tolerant.ts:119) only fail-fasts a finished spawn past the 60s grace, well under the recap (5 min) and critic (10 min) timeouts (src/review.ts:204), so a slow-booting agent is not killed prematurely.
- The recap warn now logs `Math.round(elapsed / 1000)` with "(spawn exited or hard timeout)" (src/recap.ts:551), which correctly covers both the new fail-fast and the hard-timeout exit.

scoop (commented):
## Re-review

This revision addresses both points raised previously.

**Prior point 1 — abort verdict was caught by the same-head dedup, so the auto path never re-attempted.** Fixed. `consider()`'s line-236 guard is now abort-aware: `if (!force && prior?.headSha === git.headSha && !prior.spawnAborted) return "skipped";` (src/review.ts:239). The abort row carries `spawnAborted: true` (src/review.ts:989, types.ts:340, persisted via the new `spawnAborted` column in store.ts) so a same-head abort is re-attempted every poll until the pool warms. Verified the self-heal end-to-end against the new test `critic: an abort verdict does NOT block re-attempting the same head` (test/review.test.ts:2649-2666): cold→abort persists, warm→same-head re-attempt spawns (`started` length 1), which also confirms neither the spawn ceiling (src/review.ts:265, `findings:[]` keeps the findings>0 leg false) nor the rebase churn guard blocks the re-attempt.

**Prior point 2 — the in-method `(headSha, reason)` dedup at publishSpawnAbort was unreachable/untested.** Fixed. The new test `critic: a repeated onSpawn abort for the same head does not churn the verdict` (test/review.test.ts:2668-2687) now exercises it on the AUTO path: it asserts `removed` is `["/review-wt", "/review-wt"]` (proving `begin()` ran BOTH times — the abort row is dedup-exempt, so the second `consider()` did NOT bail at the line-239 guard) while `changes === 1` (proving the in-method dedup at src/review.ts:957-963 suppressed the second `putReview`/`onChange`). I traced the two passes by hand: pass 1 `prior=null` → dedup's `prior?.decision === "error"` is false → persists (change 1); pass 2 `prior=` the just-persisted abort row → `decision==="error" && headSha===git.headSha && summary===summary` all true → early return, no churn. The branch is genuinely reachable and tested now.

### Verification performed
- `decideVerdictAction` signature change: new optional `graceElapsed=false` param (src/json-tolerant.ts:111) is back-compatible; verified ALL callers updated to pass `elapsed > STARTUP_GRACE_MS` (src/recap.ts:536, src/review.ts:643) and no other callers exist (grep). Test catalog updated for both the grace branches and the omit-flag back-compat case (test/json-tolerant.test.ts:100-114).
- Store round-trip: `putReview` INSERT now lists 18 columns / 18 placeholders / 18 bound values including `spawnAborted` (store.ts:2055-2097), the UPDATE clause sets `spawnAborted=excluded.spawnAborted`, both SELECTs (getReview store.ts:2058, snapshot store.ts:2137) read it, `fromRow` maps `r.spawnAborted ? true : undefined` (store.ts:2050), and the migration `add("spawnAborted", ... DEFAULT 0)` (store.ts:3037) follows the established idempotent ALTER pattern and backfills old rows to 0 (= real reviews). Counts and column ordering all line up.
- Identifiers resolve: `STARTUP_GRACE_MS` exported (src/json-tolerant.ts:84) and imported in both consumers; `PluginSpawnAborted` (test import from src/plugins/types) drives the abort branch through the existing `resolveAuxSpawn` → `"aborted" in aux` path (src/review.ts:372); `DEFAULT_FINAL_ROUND_TIMEOUT_MS`, `GitState`, `ReviewVerdict`, `Session` all in scope in review.ts.
- `git.headSha!` non-null assertions in `publishSpawnAbort` are safe: `consider()` returns "skipped" on `!git.headSha` (src/review.ts:228) before `begin()` is reached.
- No i18n/locale surface in this diff (server-side TS; the abort reason is stored verbatim in `summary`, badge rendering is out of scope).

### Stated limitations (could not fully verify — not blocking)
- The grace-gate fast-fail extends fail-fast to the `absent` verdict class, which #822 deliberately left to the hard timeout. Its safety rests on `isSpawnWorking()` returning false ONLY for a genuinely-finished/dead spawn. The existing in-repo note (src/json-tolerant.ts:60-65) documents that `agentStatus` can flicker to a non-`working` value, and `blocked`/`idle` are also treated as not-working — so in principle a live-but-paused critic past the 60s grace with no verdict file yet could be finalized as `error` early (it would then self-heal via the error→re-review path). Whether a live review critic ever actually reports a non-`working` status before writing its verdict depends on the external herdr daemon's status semantics, which are not present in this repo, so I could not confirm this misfires in practice. Flagging as a design consideration, not a confirmed defect; the 60s grace is well under both hard timeouts, which remain the backstop.
- `publishSpawnAbort` sets `findings: []`, mirroring the existing transient error-verdict behavior (buildVerdict + finalizeErrorVerdict also carry `findings:[]` while preserving `addressRound`/streak counters), so this is consistent with the established error-verdict contract rather than a new regression.

PR #1214, opened 2026-06-29
fix(settings): stop desktop settings tabs wrapping to a second row

scoop:
## Problem

On desktop the Settings modal's tab strip (Workspace · Coding CLIs · Plugins · Session · Device · Diagnostics) wrapped onto a **second row** at the default modal width.

## Fix (measured, not guessed)

Rendered the exact tab strip under Chromium across the mono-font fallback chain (`Berkeley Mono` → `JetBrains Mono` → `ui-monospace`):

| Variant | Natural strip width | Card needed (+32px pad +slack) |
| --- | --- | --- |
| Keep "Coding CLIs" | 577.5px | ~618px (above the modal-width budget) |
| **Shorten → "CLIs"** | **521.2px** | **560px** ✅ |

Pure widening would have pushed the modal to ~618px (≈58px wider than the next-widest dialog). Instead — per the approved plan's font-independent fallback, confirmed with the user — we:

- **Rename the tab** `settings_tab_coding_agents` "Coding CLIs" → **"CLIs"** (EN + DE). Note the strip wrapped even with the short label at today's 520px, so a small widen is still required.
- **Widen the card** `min(520px, 92vw)` → `min(560px, 92vw)`, keeping Settings in-family with sibling modals (e.g. `RecommendDialog` at 560px).
- **Update wayfinding** — the four `feat_*_body` announcements that said "Settings → Coding CLIs" now read "Settings → CLIs" so the breadcrumbs match the actual tab name (no discovery rot).

`.tabs` keeps `flex-wrap: wrap` as the intentional graceful fallback for longer locales / larger `--ui-scale`.

## Scope / notes

- Mobile path unchanged — ≤768px still collapses the strip into the dropdown `<select>`.
- Ships as a `fix(...)` with no new user-facing capability, so the feature-catalog gate doesn't apply. `[no-feature-entry]`
- i18n parity maintained (2118 keys each); `bun run check`, `check:i18n`, prettier, and pre-push hygiene all green.


scoop (commented):
## Summary

The PR addresses the task (settings tabs wrapping on desktop) two ways, both of which the task explicitly suggested:
1. Renames the tab from "Coding CLIs" → "CLIs" (`settings_tab_coding_agents`).
2. Widens the modal card from `min(520px, 92vw)` → `min(560px, 92vw)`.

### Verification

- **i18n parity** — `settings_tab_coding_agents` is changed identically in both catalogs (`ui/messages/en.json:1967`, `ui/messages/de.json:1967`), and `ui/messages/` contains only `en.json` + `de.json`, so no third locale is left stale.
- **Key preserved** — only the value changed; the key `settings_tab_coding_agents` is still referenced at `ui/src/lib/components/Settings.svelte:54` (tab list) and `:697` (panel aria-label), both resolve.
- **Cross-reference consistency** — every changelog body string that pointed users to "Settings → Coding CLIs" was updated to "Settings → CLIs" in lockstep across both catalogs (`feat_default_model_body`, `feat_api_key_auth_body`, `feat_api_key_verify_body`, `feat_codex_cli_body`), keeping the prose in sync with the renamed tab.
- **Remaining "Coding CLI" strings are intentional** — `newtask_agent_provider_label` (`en/de.json:1959`) and `settings_default_agent_provider_title` are different fields (the new-task picker label and an in-panel section title), not the tab, so they correctly keep their longer wording.
- **CSS** — `.tabs` uses `flex-wrap: wrap` (`Settings.svelte:1128`); +40px card width plus the shortened label both reduce the wrapping pressure on the desktop tab strip. The narrow-viewport path uses a `<select>` dropdown and is unaffected.

I cannot pixel-measure the rendered strip in this read-only review, but the combined width increase and label shortening is a sound, low-risk approach to the reported wrap.

No correctness, security, or quality issues found.

PR #1215, opened 2026-06-29
feat(deploy): size-cap shepherd.log via logrotate timer (#1212)

### LINKED ISSUE #1212: Log rotation for ~/.shepherd/shepherd.log (grows unbounded)
## Problem

`~/.shepherd/shepherd.log` grows **unbounded**. The systemd unit redirects both streams to it append-only and nothing ever rotates, truncates, or size-caps it.

`deploy/shepherd.service:24-25`:

```ini
StandardOutput=append:%h/.shepherd/shepherd.log
StandardError=append:%h/.shepherd/shepherd.log
```

systemd's `append:` opens the file `O_APPEND` and holds the fd for the life of the process — it **never rotates** (unlike journald, which self-caps via `SystemMaxUse`). The unit comment notes this is deliberately persistent across restarts.

## Why it matters

- **No app-side bound.** Nothing in `src/` truncates/size-checks/rotates the log. ~280 `console.*` calls (mostly `console.warn`), and the server runs many always-on loops — `poller` ticks every **1s** (`src/poller.ts:211,1315`) plus a dozen+ `setInterval`s in `src/index.ts`. Continuously chatty.
- **No logrotate shipped.** `deploy/` has no logrotate config; `install.sh`/`provision.ts` only lay down the unit. (The GFS rotation in `shepherd-backup.service` is for the SQLite backup — unrelated.)

Net: a long-running install appends to one file forever — slow-burn disk filler and an ever-slower file to `cat`/`grep`.

## Wrinkle

Because systemd holds the fd open in append mode, a naïve logrotate (`rename`+`create`) **won't** reclaim space — the process keeps writing to the old inode.

## Options

1. **logrotate + `copytruncate`** (+ `size`/`maxsize`, `compress`, `rotate N`) — ship `deploy/shepherd.logrotate`, wire it in `install.sh`/`provision.ts`. Preserves the deliberate persistent-file design while bounding size. **Recommended.**
2. **Drop the file redirect, use journald** — remove the two `append:` lines, read via `journalctl --user -u shepherd`. Free rotation but loses the standalone file the unit comment wanted; journald has its own caps.
3. **logrotate + `postrotate: systemctl restart`** — works but restarts the service on every rotation (disruptive given the always-on loops). Not preferred.

## Recommendation

Option 1 — `logrotate` with `copytruncate`.

---

scoop:
Closes #1212.

## Problem

`~/.shepherd/shepherd.log` grows **unbounded**: `shepherd.service` redirects both streams `append:` and holds the log fd open `O_APPEND` for the life of the process. Nothing rotates, truncates, or size-caps it — a slow-burn disk filler. The held fd is the wrinkle: a naïve `rename`+`create` rotation won't reclaim space (the process keeps writing to the old inode).

## Approach

Ship a logrotate config driven by a **user-space systemd timer**, mirroring the existing hourly-backup unit pattern (Shepherd installs entirely under `systemctl --user`, no root after prereqs — so `/etc/logrotate.d` + system cron would be the wrong layer).

**Policy** (`deploy/shepherd.logrotate`): `copytruncate` (mandatory for the held fd — copies aside, truncates the original in place), `size 50M`, `rotate 7`, `compress`. Bound: ~50M live + 7 gzipped archives.

**Cadence**: hourly `shepherd-logrotate.timer` (matches `shepherd-backup.timer`). A `size` trigger is only as tight as how often it's checked — hourly bounds the live file's overshoot past 50M to ~1h of output (the 1s poller + many `setInterval`s are continuously chatty). The config is size-only; the timer provides the cadence.

## Wiring

- `provision.ts:installService` — installs the config + units on a fresh box; `update.sh` self-heals them on existing hosts (both mirror the backup-timer plumbing).
- `install.sh` — best-effort installs `logrotate`.

## Optional / fail-safe

logrotate is **optional** and never blocks an install:
- `install.sh` install is **best-effort (warn-not-die)** and **Linux-only** (macOS core-only has no timer and none of `ensure_pkg`'s package managers).
- `provision.ts` / `update.sh` **soft-skip** the timer (with a warning) if the binary is absent — the log just stays unbounded as before.
- **Detection searches `/usr/sbin:/sbin`** (where logrotate usually lives) at all three sites — a `curl | bash` non-login PATH often omits sbin, which would otherwise false-negative and silently skip the feature.

logrotate doesn't expand `%h`, so the config's `%h` log path is templated to the real home before the on-disk copy (`~/.shepherd/shepherd.logrotate`) is written; the units' own `%h` is expanded by systemd.

## Tests

`test/provision.test.ts`: config + units written (home-expanded, `copytruncate`), hourly timer `enable --now`, **soft-skip when logrotate absent**, plus `templateLogrotateConfig` + sbin-aware `resolveLogrotate` unit tests. Verified: `bun run typecheck`, `bun test ./test` (5241 pass), `bun run lint`, `bunx fallow audit` all green.

`[no-feature-entry]` — operator-facing deploy plumbing, no user-facing UI (and deploy tooling is explicitly not i18n'd).

### DIFF
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -111,7 +111,11 @@ install_os_prereqs() {
   # logrotate backs the OPTIONAL log-rotation timer (#1212), Linux service path only. Best-effort
   # (warn, never die) and Linux-gated: a failed optional install must not abort the whole install,
   # and macOS/core-only installs no timer (and has none of ensure_pkg's package managers).
-  [ "$OS_KIND" = "Linux" ] && ensure_logrotate
+  # NB: a proper `if` (not `[ … ] && …`) so the non-Linux skip leaves a 0 exit — else this last
+  # command returns 1 on macOS and `set -e` aborts the installer.
+  if [ "$OS_KIND" = "Linux" ]; then
+    ensure_logrotate
+  fi
 }
 
 # ensure_pkg <bin> <pkg>: install <pkg> via apt/apk/dnf/pacman if <bin> is absent.
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -108,6 +108,10 @@ install_os_prereqs() {
   ensure_pkg git git
   ensure_pkg unzip unzip
   ensure_toolchain
+  # logrotate backs the OPTIONAL log-rotation timer (#1212), Linux service path only. Best-effort
+  # (warn, never die) and Linux-gated: a failed optional install must not abort the whole install,
+  # and macOS/core-only installs no timer (and has none of ensure_pkg's package managers).
+  [ "$OS_KIND" = "Linux" ] && ensure_logrotate
 }
 
 # ensure_pkg <bin> <pkg>: install <pkg> via apt/apk/dnf/pacman if <bin> is absent.
@@ -128,6 +132,21 @@ ensure_toolchain() {
   note "installing C/C++ build toolchain + python3 (node-pty native build)"
   $sp sh -c "(apt-get update && apt-get install -y build-essential python3) || apk add --no-cache build-base python3 || dnf install -y gcc-c++ make python3 || pacman -Sy --noconfirm base-devel python3" \
     || die "failed to install a C/C++ toolchain + python3 with any supported package manager"
+}
+
+# ensure_logrotate: best-effort install of logrotate for the optional log-rotation timer (#1212).
+# Unlike ensure_pkg this NEVER dies — logrotate is optional, so a failed install just means
+# provision.ts soft-skips the timer (the log stays unbounded, as before). Detection searches
+# /usr/sbin:/sbin where logrotate usually lives (a curl|bash non-login PATH often omits sbin).
+ensure_logrotate() {
+  local sp
+  PATH="$PATH:/usr/sbin:/sbin" command -v logrotate >/dev/null 2>&1 && return 0
+  # sudo_prefix dies if it can't elevate; run it in a subshell so that exit can't abort us — an
+  # empty prefix then makes the install attempt fail unprivileged and fall through to the warn.
+  sp="$(sudo_prefix logrotate 2>/dev/null || true)"
+  note "installing logrotate (optional — log-rotation timer)"
+  $sp sh -c "(apt-get update && apt-get install -y logrotate) || apk add --no-cache logrotate || dnf install -y logrotate || pacman -Sy --noconfirm logrotate" \
+    || warn "could not install logrotate — log-rotation timer will be skipped (log stays unbounded)"
 }
 
 # ── Bun ───────────────────────────────────────────────────────────────────────
--- a/deploy/install.sh
+++ b/deploy/install.sh
@@ -144,10 +144,13 @@ ensure_toolchain() {
 # /usr/sbin:/sbin where logrotate usually lives (a curl|bash non-login PATH often omits sbin).
 ensure_logrotate() {
   local sp
-  PATH="$PATH:/usr/sbin:/sbin" command -v logrotate >/dev/null 2>&1 && return 0
-  # sudo_prefix dies if it can't elevate; run it in a subshell so that exit can't abort us — an
-  # empty prefix then makes the install attempt fail unprivileged and fall through to the warn.
-  sp="$(sudo_prefix logrotate 2>/dev/null || true)"
+  # Detect against the SAME dir list the shepherd-logrotate.service unit puts on PATH, so we never
+  # consider logrotate "present" in a dir the timer can't reach at runtime (keep these in lockstep).
+  PATH="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin" command -v logrotate >/dev/null 2>&1 && return 0
+  # sudo_prefix dies if it can't elevate; the `|| true` is OUTSIDE the substitution — `exit 1`
+  # inside terminates the subshell before any in-substitution `|| true` could run, so the failed
+  # substitution would otherwise abort us under `set -e`. Empty prefix ⇒ unprivileged attempt ⇒ warn.
+  sp="$(sudo_prefix logrotate 2>/dev/null)" || true
   note "installing logrotate (optional — log-rotation timer)"
   $sp sh -c "(apt-get update && apt-get install -y logrotate) || apk add --no-cache logrotate || dnf install -y logrotate || pacman -Sy --noconfirm logrotate" \
     || warn "could not install logrotate — log-rotation timer will be skipped (log stays unbounded)"
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -121,10 +121,22 @@ export function templateLogrotateConfig(cfg: string, home: string): string {
   return cfg.replaceAll("%h", home);
 }
 
-/** logrotate commonly lives in /usr/sbin (or /sbin), which a `curl | bash` non-login shell's
- *  inherited PATH frequently omits — so a bare `logrotate` probe can false-negative. Search the
- *  sbin candidates explicitly. */
-const LOGROTATE_CANDIDATES = ["logrotate", "/usr/sbin/logrotate", "/sbin/logrotate"] as const;
+/** The dirs the shepherd-logrotate.service unit puts on PATH — MUST stay in lockstep with that
+ *  unit's `Environment=PATH` (and install.sh/update.sh detection). logrotate commonly lives in
+ *  /usr/sbin, which a `curl | bash` non-login PATH often omits. */
+const LOGROTATE_DIRS = [
+  "/usr/local/sbin",
+  "/usr/local/bin",
+  "/usr/sbin",
+  "/usr/bin",
+  "/sbin",
+  "/bin",
+] as const;
+
+/** Absolute candidates only — NO bare `logrotate`: a bare probe resolves against the inherited
+ *  PATH, which can match a dir absent from the unit's PATH, so the timer would install yet fail to
+ *  exec logrotate at runtime. Absolute paths make detection ≡ execution. */
+const LOGROTATE_CANDIDATES = LOGROTATE_DIRS.map((dir) => `${dir}/logrotate`);
 
 /** First logrotate candidate that responds to `--version`, or null if none is found. Gates the
  *  (optional) log-rotation timer install; null ⇒ soft-skip. */
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -114,6 +114,27 @@ export function templateUnit(unit: string, repo: string): string {
   return unit.replace(/^WorkingDirectory=.*$/m, `WorkingDirectory=${repo}`);
 }
 
+/** logrotate config paths are literal — logrotate does NOT expand %h/~. So the repo config uses
+ *  %h as a placeholder and we template it to the real home before writing the on-disk copy the
+ *  timer runs. (Unit files DO get %h from systemd, so only the config needs this.) #1212 */
+export function templateLogrotateConfig(cfg: string, home: string): string {
+  return cfg.replaceAll("%h", home);
+}
+
+/** logrotate commonly lives in /usr/sbin (or /sbin), which a `curl | bash` non-login shell's
+ *  inherited PATH frequently omits — so a bare `logrotate` probe can false-negative. Search the
+ *  sbin candidates explicitly. */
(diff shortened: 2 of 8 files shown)


scoop:
## Summary

The PR implements Option 1 from the issue (logrotate + `copytruncate`, size-capped, hourly timer) and wires it into `provision.ts`, `install.sh`, and `update.sh` with a clean optional/soft-skip posture and good test coverage. The systemd units, config templating, and detection logic are internally consistent and correct. One blocking shell bug regresses the macOS install path.

## Blocking

### `deploy/install.sh:114` — `&& ensure_logrotate` makes `install_os_prereqs` return non-zero on macOS, and `set -e` aborts the installer

The new last line of `install_os_prereqs` is:

```sh
[ "$OS_KIND" = "Linux" ] && ensure_logrotate
```

On macOS, `OS_KIND` is `"Darwin"` (verified: `detect_os` sets `OS_KIND=$(uname -s)` at `deploy/install.sh:53`; `decide()` matches the `Darwin` case at `deploy/install.sh:66`). So `[ "$OS_KIND" = "Linux" ]` exits 1, `&&` short-circuits, and the AND-list — the function's **last** command — yields exit status 1. The function therefore returns 1.

`install_os_prereqs` is called as a standalone command at `deploy/install.sh:251`, under `set -euo pipefail` (`deploy/install.sh:35`). A standalone command returning non-zero triggers `set -e`, so the script exits **before** `install_bun` / `resolve_source` ever run. The comment at `deploy/install.sh:249-250` explicitly states macOS still needs prereqs + bun and reaches this function, so this is a real regression: the macOS (core-only) install now aborts.

(Note: the `&&`-left failure does not trip `set -e` *inside* the function — left operands of `&&` are exempt — but the function still *returns* 1, and the exempt context does not extend to the standalone call site.)

Linux is unaffected: there `ensure_logrotate` runs and returns 0 (it ends in `... || warn`, and `warn` returns 0).

Fix — make the last command return 0 on non-Linux, e.g.:

```sh
if [ "$OS_KIND" = "Linux" ]; then ensure_logrotate; fi
```

or append `|| true` / reorder so a non-Linux host leaves a zero exit status.

## Verification notes

- macOS abort path: verified against `deploy/install.sh:35` (`set -euo pipefail`), `:53` (`OS_KIND` source), `:66` (Darwin case), `:114` (the gate), `:249-251` (macOS reaches `install_os_prereqs`, called standalone).
- Config/state path consistency: `deploy/shepherd-logrotate.service:9` runs `logrotate --state %h/.shepherd/logrotate.state %h/.shepherd/shepherd.logrotate`; `provision.ts:308` writes the templated config to `~/.shepherd/shepherd.logrotate`; `update.sh:81` writes the same path. Consistent.
- `%h` templating: `templateLogrotateConfig` (`provision.ts:120-122`) `replaceAll("%h", home)` against `deploy/shepherd.logrotate:13` (`%h/.shepherd/shepherd.log`); units keep `%h` for systemd to expand. Correct division verified.
- Detection parity: `install.sh:143` and `update.sh:78` use `PATH="$PATH:/usr/sbin:/sbin" command -v logrotate`; `provision.ts:127` probes `["logrotate","/usr/sbin/logrotate","/sbin/logrotate"]`. All three search the sbin candidates consistently.
- `installService` signature: new `probe` param defaults to `probeVersion` (`provision.ts:277`) and is passed through from `provision()` at `provision.ts:377`; callers/tests updated (`test/provision.test.ts:342,376`). Consistent.
- `ExecStart=logrotate` (bare) resolves via the unit's `Environment=PATH=/usr/sbin:/usr/bin:/sbin:/bin` (`deploy/shepherd-logrotate.service:13`). Fine.

- (none)

## Notes (non-blocking)
- `deploy/shepherd.logrotate`: `copytruncate` has an inherent copy-then-truncate race that can drop log lines written in that window. This is the correct/only choice given systemd holds the fd `O_APPEND` (rename+create wouldn't reclaim space), and the trade-off is documented in the config header. No action needed — noted for awareness.

scoop:
## Summary

This PR implements Option 1 from the issue (logrotate + `copytruncate`, size-capped, wired into `install.sh`/`provision.ts`/`update.sh`) and does so carefully overall: the config uses `copytruncate` (correct for the held-open `O_APPEND` fd — verified against `deploy/shepherd.service:23-25`), the `%h` templating is correct (logrotate does not expand `%h`, so `provision.ts`/`update.sh` substitute the real home before writing `~/.shepherd/shepherd.logrotate`), the units' own `%h` paths are left for systemd, the timer mirrors `shepherd-backup.timer`, and the whole feature is gated/soft-skipped when logrotate is absent. The sbin-aware detection (`resolveLogrotate`) and the `Environment=PATH=/usr/sbin:/usr/bin:/sbin:/bin` on the service correctly address the non-login-PATH-omits-sbin problem.

One blocking bug, though, defeats the PR's own stated "best-effort, never die" safety property.

## Blocking

**`deploy/install.sh:150` — `ensure_logrotate` aborts the whole installer when non-root without sudo.**

```sh
local sp            # line 146 — separate declaration
...
sp="$(sudo_prefix logrotate 2>/dev/null || true)"   # line 150
```

`sudo_prefix` calls `die` → `exit 1` (verified `deploy/install.sh:97` → `deploy/install.sh:46`) when the user is non-root AND `sudo` is unavailable. That `exit 1` terminates the command-substitution **subshell immediately**, so the `|| true` placed *inside* the `$(...)` is never reached — the substitution returns status 1. Because `local sp` is declared on a separate line (146), the assignment on line 150 is a standalone simple command, so `set -euo pipefail` (line 35) acts on its exit status and **aborts the installer**.

This is exactly the outcome the comment at lines 148-149 and 141-144 says it prevents ("run it in a subshell so that exit can't abort us … fall through to the warn"). The subshell stops `die` from killing the parent *directly*, but it does not stop `set -e` from acting on the failed substitution.

Reachability: a non-root host without `sudo` that already has curl/git/unzip/cc/make present (so the earlier `ensure_pkg`/`ensure_toolchain` all `return 0` early via their `command -v` guards and never invoke `sudo_prefix`). On such a host the installer previously completed; now it dies at `ensure_logrotate` — a regression, and the optional feature is supposed to degrade, not abort.

Fix: move `|| true` outside the substitution so the outer AND-OR list absorbs the failure:
```sh
sp="$(sudo_prefix logrotate 2>/dev/null)" || true
```
Then the subshell exits 1, the assignment fails, `|| true` yields status 0, `set -e` does not fire, `sp` is empty, and the unprivileged install attempt falls through to the `warn` as intended.

(`update.sh`'s logrotate block has no analogous problem — it only probes and copies, never invokes `sudo_prefix`.)

## Nit (non-blocking)

**`deploy/provision.ts:127` / `deploy/shepherd-logrotate.service:13` — detection vs execution PATH can diverge for `/usr/local/sbin`.** `resolveLogrotate` includes the bare `logrotate` candidate, which resolves against provision's process PATH and could match a binary in e.g. `/usr/local/sbin`; the service's `Environment=PATH` is `/usr/sbin:/usr/bin:/sbin:/bin` and omits `/usr/local/sbin`. In that (rare) case the units get installed but the timer-driven run can't find `logrotate`. Aligning the two (drop the bare candidate, or add `/usr/local/sbin` to both) closes the gap. Package-managed logrotate lands in `/usr/sbin` or `/sbin`, so this is unlikely in practice.

## Verification notes
- `installService` signature change (`probe` added with default `probeVersion`, `deploy/provision.ts:277`) — verified the only production caller passes `probe` (`deploy/provision.ts:377`) and the test callers were updated (`test/provision.test.ts:345,395,401`); no other callers (`ci/onboarding-harness/run.ts:100` is a comment).
- `copytruncate`/`size 50M`/`rotate 7` policy verified against `deploy/shepherd.logrotate:14-19` and matched by the tests (`test/provision.test.ts:150-152`).
- `%h` templating verified: `templateLogrotateConfig` (`deploy/provision.ts:120-122`) replaces all `%h`; service/timer copied verbatim with `%h` left for systemd (`deploy/shepherd-logrotate.service:9`).
- No i18n catalogs touched; deploy config is operator-facing plain English by design.

- deploy/shepherd-logrotate.service: see the PATH nit above — only reachable if logrotate is detected from a directory not in the service's `Environment=PATH`.

scoop (commented):
## Summary

This PR implements Option 1 from issue #1212 — a `logrotate` + `copytruncate` size-cap for `~/.shepherd/shepherd.log`, wired into `install.sh`, `provision.ts`, and `update.sh`, plus systemd `shepherd-logrotate.{service,timer}` and a templated `shepherd.logrotate` config. The approach correctly preserves the deliberate persistent-file design (the held `O_APPEND` fd keeps writing to the truncated inode), which the issue's "wrinkle" flagged.

This is a re-review; I confirmed both prior findings are now fixed.

## Prior findings — both resolved

**1. `install.sh` `ensure_logrotate` `|| true` placement (set -e abort).** Now `deploy/install.sh:153` reads `sp="$(sudo_prefix logrotate 2>/dev/null)" || true` with `|| true` OUTSIDE the substitution. Verified the failure path: `sudo_prefix` (deploy/install.sh:88-98) calls `die` → `printf ... >&2; exit 1` (deploy/install.sh:44-47) when non-root with no sudo. The subshell exits 1, the substitution returns 1, but the trailing `|| true` makes the assignment command exit 0, so `set -e` does not abort. `sp` is left empty → unprivileged install attempt → `|| warn` (deploy/install.sh:155). Best-effort intent restored.

**2. detection PATH vs execution PATH divergence.** Now aligned. `resolveLogrotate` (deploy/provision.ts:143) probes only absolute candidates built from `LOGROTATE_DIRS` (deploy/provision.ts:124-133): `/usr/local/sbin`, `/usr/local/bin`, `/usr/sbin`, `/usr/bin`, `/sbin`, `/bin` — no bare `logrotate`. I compared this against the unit's `Environment=PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin` (deploy/shepherd-logrotate.service:15) and the detection PATH in `install.sh` (deploy/install.sh:148) and `update.sh` (deploy/update.sh:79) — all six dirs identical and in lockstep. Detection now ≡ execution.

## Verification performed

- **`probe` plumbing.** `installService` gained `probe: (bin) => string|null = probeVersion` (deploy/provision.ts:289); `provision` passes the resolved `probe` (deploy/provision.ts:371, 389). `probeVersion` runs `<bin> --version` via `execFileSync` and returns null on throw (deploy/provision.ts:211-222) — a non-existent absolute path throws ENOENT → null, so the loop falls through correctly to the next candidate / null. logrotate supports `--version` (prints `logrotate 3.x.y`), matching the `\d+\.\d+\.\d+` regex.
- **`~/.shepherd` exists before config write.** `run("mkdir", ["-p", join(home, ".shepherd")])` at deploy/provision.ts:299 precedes the logrotate config write at deploy/provision.ts:319-322. `update.sh` does `mkdir -p "$HOME/.shepherd"` at deploy/update.sh:81 before its `sed` write.
- **Templating.** `templateLogrotateConfig` uses `replaceAll("%h", home)` (deploy/provision.ts:121) — literal, safe. The config's `%h/.shepherd/shepherd.log` (deploy/shepherd.logrotate:13) is templated to the real home; unit files keep `%h` (expanded by systemd). Confirmed `shepherd-logrotate.service:9` `--state %h/.shepherd/logrotate.state %h/.shepherd/shepherd.logrotate` relies on systemd %h expansion for a user unit.
- **systemd wiring.** Timer has `[Install] WantedBy=timers.target` (deploy/shepherd-logrotate.timer:12-13); oneshot service needs none. `daemon-reload` (deploy/provision.ts:334) precedes `enable --now shepherd-logrotate.timer` (deploy/provision.ts:348), which is gated on `logrotateBin` so the units are guaranteed written first (deploy/provision.ts:317-330). Timer→service name match (shepherd-logrotate.timer → shepherd-logrotate.service) is implicit and both files exist.
- **copytruncate correctness.** `size 50M`, `rotate 7`, `missingok`, `notifempty`, `compress`, `copytruncate` (deploy/shepherd.logrotate:14-19). With only a `size` trigger and no time directive, logrotate rotates purely on size whenever the hourly timer invokes it — consistent with the timer comment (deploy/shepherd-logrotate.timer:7-9).
- **No broken callers.** Grepped all `installService` callers: only deploy/provision.ts:389 (passes probe) and tests; the new param is defaulted so ci/onboarding-harness/run.ts is unaffected.

The known copytruncate copy-then-truncate data-loss window is inherent to the technique and acceptable here (it's the only option compatible with systemd's held `append:` fd, as the issue notes).

No blocking issues found.

Issue #1192, opened 2026-06-28
Speed up CI lint (>1 min): hybrid fast-linter for .ts/.js, keep eslint-plugin-svelte for .svelte


scoop:
## Problem

CI now spends **>1 min** on linting. Question raised: are the fast ESLint alternatives (Biome, Oxlint) good enough for a SvelteKit project yet?

## TL;DR

- **For `.svelte` files themselves: not a drop-in replacement (as of mid-2026).** No fast Rust linter parses Svelte template syntax. `eslint-plugin-svelte` is still the only real option for `.svelte`.
- **The practical answer is yes, via the hybrid practitioners actually use:** fast linter for `.ts`/`.js`, keep `eslint-plugin-svelte` *only* for `.svelte`. (The Svelte team is itself moving its repo's JS/TS to Oxlint — sveltejs/svelte#17665.)
- **Repo-specific:** our slowness is *not* the notorious typed-Svelte trap — we don't run type-aware linting (no `parserOptions.project`). The minute is plain ESLint over ~3k files + whole-repo `prettier --check`, with **no eslint cache and a fresh install every CI run**.

## Linter support for `.svelte` (mid-2026)

| Tool | `.svelte` support | Status / source |
|---|---|---|
| **Oxlint** (oxc) | `<script>` block only, **no template** | "No template linting yet" — [oxc.rs/compatibility]; full SFC tracked, "Q2" milestone ([oxc#15761], [oxc#18708]) |
| **Biome** | textual JS/CSS regions, **no Svelte control-flow** | Experimental since v2.3, refined v2.4; docs tell you to *disable* `noUnusedVariables`/`useConst`/etc. on Svelte to avoid false positives ([language-support], [roadmap-2026]) |
| **deno lint** | none — `.svelte` not a recognized extension | [deno lint docs] |
| **eslint-plugin-svelte** | **full template + script** (via svelte-eslint-parser) | The only real option; what we run today |

Note: `svelte-check` is a **type-checker, not a linter** — orthogonal, stays in the stack regardless.

## Why this matters for *our* config

[...]

PR #1218, opened 2026-06-29
fix(prettier): ignore .shepherd-* session artifacts

scoop:
## Problem

Prettier does **not** read `.git/info/exclude`, which is Shepherd's primary
hide mechanism for its per-session scratch files (the `.shepherd-*` glob, written
by `src/shepherd-exclude.ts` on every worktree). So `prettier --check .` flags any
`.shepherd-*` file not *also* named in the committed `.gitignore` — verified: a
misformatted `.shepherd-recap.json` was flagged.

Empirically re-verified (prettier 3.8.5, deliberately misformatted files):

| File | Ignored via | `prettier --check` |
|------|-------------|--------------------|
| `.shepherd-recap.json` | `.git/info/exclude` only | **FLAGGED** |
| `zzqtest/ignored.json` | nested `.gitignore` | **FLAGGED** |
| `.shepherd-plan.md`, `.test-logs/`, `.fallow/` | root `.gitignore` | not flagged |
| `CHANGELOG.md` (tracked) | `.prettierignore` only | not flagged |

Prettier's `.gitignore` support is partial (nested `.gitignore` not honored) and
diverges from git — matching the repo's #1192 history. So `.prettierignore` is
treated as the only **reliable** mechanism.

## Changes

- **`.prettierignore`** — add `.shepherd-*` (primary fix; same rationale as `.cache`).
- **`.gitignore`** — consolidate the named `.shepherd-*` entries + a stray duplicate
  into one `.shepherd-*` glob block (orthogonal git/CI/teammate hygiene; mirrors the
  `info/exclude` glob). Not relied on for the Prettier fix.
- **`src/readiness.ts`** — add an always-shown note to the generated AI-readiness
  `CLAUDE.md` telling 3rd-party repos that adopt Prettier to add `.shepherd-*` to
  their `.prettierignore`. Shown regardless of whether the repo already runs Prettier
  (that's the higher-impact case). Glob sourced from `SHEPHERD_IGNORE_GLOB` — no second
  hardcoded string.
- **`test/readiness.test.ts`** — assert the note renders even when a formatter is
  already present.

## Verification

- Misformatted `.shepherd-recap.json` no longer flagged by `prettier --check .`.
- `bun run lint` clean; `bun test ./test` (readiness 21/21) green; full `prettier --check .` clean.
- All pre-push gates pass (incl. fallow audit, ui, root-tests).

Notes: generated `claudeMd` is a verbatim target-repo artifact (i18n-exempt); server-only
content change with no new UI surface.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,16 +17,12 @@ dist/
 .shaping/
 .squad/
 
-# Runtime upload staging (New-Task attachments, screenshots)
-.shepherd-uploads/
-
-# Plan-gate working artifacts written into the live worktree (never committed)
-.shepherd-plan.md
-.shepherd-plan-blocks.json
-.shepherd-plan-review.json
-
-# Agent-declared live-preview port hint (transient, never committed)
-.shepherd-preview
+# Shepherd per-session scratch artifacts — plan-gate docs, preview-port hint, upload
+# staging (New-Task attachments/screenshots), recap/name/learnings/… . This catch-all
+# glob mirrors the LOCAL-ONLY .git/info/exclude block (SHEPHERD_IGNORE_GLOB in
+# src/shepherd-exclude.ts) so teammates & CI ignore every artifact too. Matches
+# .shepherd-uploads/ as well as all .shepherd-* files.
+.shepherd-*
 
 # Onboarding-harness gap report — generated into cwd by ci/onboarding-harness/run.ts
 onboarding-gap-report.md
@@ -45,6 +41,5 @@ ci/self-hosted-runner/.env
 # Pre-push test capture logs (flake diagnosis, #817)
 .test-logs/
 
-# SDD subagent scratch (briefs) + plan sidecar
+# SDD subagent scratch (briefs); the plan sidecar is covered by .shepherd-* above
 briefs/
-.shepherd-plan-blocks.json
--- a/.prettierignore
+++ b/.prettierignore
@@ -29,10 +29,11 @@ extension/public/picker.js
 # Transient agent worktrees (nested checkouts; not project source)
 .claude/worktrees
 
-# Shepherd per-session scratch artifacts. Shepherd hides these via the LOCAL-ONLY
-# .git/info/exclude (the .shepherd-* glob, src/shepherd-exclude.ts) — which prettier
-# does NOT read — so without this line `prettier --check .` flags the ones not also
-# named in .gitignore (e.g. .shepherd-recap.json). Same rationale as .cache above.
+# Shepherd per-session scratch artifacts (the .shepherd-* glob, src/shepherd-exclude.ts).
+# This is the RELIABLE ignore: although .shepherd-* is also in .gitignore, prettier's
+# .gitignore support is only partial (it skips nested .gitignore, and per #1192 still
+# traversed gitignored .cache) and it never reads the LOCAL-ONLY .git/info/exclude where
+# Shepherd hides these on every worktree. Same rationale as .cache above.
 .shepherd-*
 
 # Agent-generated planning artifacts — leave their formatting alone
--- a/.prettierignore
+++ b/.prettierignore
@@ -29,6 +29,12 @@ extension/public/picker.js
 # Transient agent worktrees (nested checkouts; not project source)
 .claude/worktrees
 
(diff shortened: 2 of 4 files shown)


scoop (commented):
## Summary

This PR replaces the five hand-maintained `.shepherd-*` entries in `.gitignore` with a single `.shepherd-*` catch-all glob, mirrors that glob into `.prettierignore`, and appends an AI-readiness instruction telling third-party repos to add `.shepherd-*` to their `.prettierignore`. The change satisfies the task.

### Verification
- `SHEPHERD_IGNORE_GLOB` resolves to `.shepherd-*` (verified src/shepherd-exclude.ts:21). `src/readiness.ts:408-411` interpolates it into both the prose and the `echo '...' >> .prettierignore` command; the new test (test/readiness.test.ts:206-221) asserts the literal `echo '.shepherd-*' >> .prettierignore`, so prose and test agree.
- The `.gitignore` glob `.shepherd-*` (.gitignore:22) is strictly broader than the removed lines (`.shepherd-uploads/`, `.shepherd-plan.md`, `.shepherd-plan-blocks.json`, `.shepherd-plan-review.json`, `.shepherd-preview`) — every removed pattern starts with `.shepherd-`, so no ignore coverage is lost. The duplicate bottom `.shepherd-plan-blocks.json` line is correctly dropped.
- No tracked file is accidentally newly-ignored: `git ls-files | grep shepherd` shows `.shepherd/roles.json` and `src/shepherd-exclude.ts`; the glob `.shepherd-*` matches neither (`.shepherd/` has no hyphen after `.shepherd`; `shepherd-exclude.ts` has no leading dot). Verified against .gitignore:22.
- The new `.prettierignore` comment references `.cache` "above" — confirmed present at .prettierignore:22.

### Nit (non-blocking)
The new `.prettierignore` comment (lines 32-35) says the line is needed because prettier "flags the ones not also named in .gitignore (e.g. `.shepherd-recap.json`)". But prettier `^3.8.5` (package.json) reads `.gitignore` by default, and *this same PR* adds `.shepherd-*` to `.gitignore` (.gitignore:22) — which now does name `.shepherd-recap.json`. So within this PR the cited example is self-contradictory, and for this repo the `.prettierignore` line is belt-and-suspenders rather than strictly necessary. The line itself is still worth keeping (it documents intent and is the canonical form the readiness prompt teaches third-party repos, whose Shepherd artifacts live in `.git/info/exclude`, not `.gitignore`). Consider tweaking the comment so the example isn't contradicted by the adjacent `.gitignore` change.

No correctness, security, or logic problems found.

scoop (commented):
## Re-review

### Prior finding 1 — RESOLVED
The earlier `.prettierignore` comment cited `.shepherd-recap.json` as a file "not also named in .gitignore", which was self-contradictory because this PR adds `.shepherd-*` to `.gitignore`. The new comment (`.prettierignore:32-37`) drops that example entirely and instead justifies the entry by prettier's *partial* `.gitignore` support (skips nested `.gitignore`, per #1192 still traverses gitignored `.cache`) and its inability to read `.git/info/exclude`. No contradiction remains.

### Verification of the diff
- `SHEPHERD_IGNORE_GLOB` resolves: defined as `".shepherd-*"` at `src/shepherd-exclude.ts:21`, imported correctly in `src/readiness.ts:16` via `./shepherd-exclude`. The generated CLAUDE.md note (`src/readiness.ts:406-412`) interpolates it consistently, producing `echo '.shepherd-*' >> .prettierignore`, which exactly matches the assertion in `test/readiness.test.ts:221`.
- The new test (`test/readiness.test.ts:206-221`) checks the artifact note renders even when the formatter guardrail is already satisfied; the note is appended unconditionally after `${adoptLines}` (`src/readiness.ts:404-412`), so it renders regardless of guardrail state. Consistent.
- `.gitignore` catch-all: replacing the enumerated entries with `.shepherd-*` does not shadow any tracked file — `git ls-files | grep '.shepherd-'` returns none. The `.gitignore` comment's claim that the glob "mirrors the LOCAL-ONLY .git/info/exclude block (SHEPHERD_IGNORE_GLOB in src/shepherd-exclude.ts)" matches `src/shepherd-exclude.ts:1-12,21,41`.
- The `.gitignore:44` comment correctly notes the former standalone `.shepherd-plan-blocks.json` is now covered by the `.shepherd-*` glob.

No blocking issues found.

PR #1221, opened 2026-06-29
fix(issues): surface fetch failure in IssuesPanel instead of empty state

kai-osthoff:
## Why

In the **Repos** panel, the Issues tab showed **"keine offenen Issues"** for a repo that has open issues. Nothing was broken with the repo — GitHub's **GraphQL API quota was exhausted** (`gh issue list` runs on that quota), so the fetch failed.

The server already handles this: `GET /api/issues` (`src/server.ts:3802-3813`) catches the rate-limited/failed listing and returns an empty `issues[]` **plus** `error: "fetch_failed"`, with a comment that it's flagged *"so the UI can say 'couldn't load' instead of the indistinguishable 'no open issues'"*.

But `IssuesPanel.svelte` **ignored that flag** and fell through to the generic `common_no_open_issues` empty state — making a transient rate limit look like an empty backlog. The New Task picker (`PromptSources.svelte`) already reads `r.error`; `IssuesPanel` just never got the same treatment.

## What

- `IssuesPanel` now tracks `loadError`, set from `r.error` (and on a thrown `listIssues`), and renders the existing `common_issues_load_failed` message ahead of the empty state.
- Added two browser tests covering the empty-vs-fetch-failed distinction.

No new i18n keys (reuses `common_issues_load_failed`, in both locales). Bugfix, no user-facing feature → no catalog entry.

## Verification

- `cd ui && bun run check` → 0 errors
- `bun run check:i18n` → 2 locales in parity
- `bun run test src/lib/components/IssuesPanel.browser.test.ts` → 19 passed

### DIFF
--- a/ui/src/lib/components/IssuesPanel.browser.test.ts
+++ b/ui/src/lib/components/IssuesPanel.browser.test.ts
@@ -79,6 +79,45 @@ describe("IssuesPanel repo slug link", () => {
   });
 });
 
+describe("IssuesPanel empty vs fetch-failed", () => {
+  it("shows the no-open-issues message when the listing is a genuine zero", async () => {
+    mockListIssues.mockResolvedValue({
+      slug: "owner/repo",
+      webUrl: null,
+      issues: [],
+      viewer: null,
+    });
+    mockGetEpics.mockResolvedValue({ epics: [], subIssues: [] });
+    render(IssuesPanel, { repoPath: "/repo", onnewtask: noop });
+
+    await expect
+      .poll(() => document.querySelector(".issues-list")?.textContent)
+      .toContain(m.common_no_open_issues());
+    expect(document.querySelector(".issues-list")?.textContent).not.toContain(
+      m.common_issues_load_failed(),
+    );
+  });
+
+  it("shows the load-failed message when the forge listing errored (e.g. rate limit)", async () => {
+    mockListIssues.mockResolvedValue({
+      slug: "owner/repo",
+      webUrl: null,
+      issues: [],
+      viewer: null,
+      error: "fetch_failed",
+    });
+    mockGetEpics.mockResolvedValue({ epics: [], subIssues: [] });
+    render(IssuesPanel, { repoPath: "/repo", onnewtask: noop });
+
+    await expect
+      .poll(() => document.querySelector(".issues-list")?.textContent)
+      .toContain(m.common_issues_load_failed());
+    expect(document.querySelector(".issues-list")?.textContent).not.toContain(
+      m.common_no_open_issues(),
+    );
+  });
+});
+
 describe("IssuesPanel epic badge", () => {
   function issue(number: number, title: string): Issue {
     return {
--- a/ui/src/lib/components/IssuesPanel.svelte
+++ b/ui/src/lib/components/IssuesPanel.svelte
@@ -47,6 +47,10 @@
   let repoUrl = $state<string | null>(null);
   let viewer = $state<string | null>(null);
   let loading = $state(true);
+  // True when the forge listing failed (rate-limited gh, network, un-authed CLI):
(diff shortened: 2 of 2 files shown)


kai-osthoff (commented):
## Summary

The PR addresses the task ("why is nothing shown — did something break?") by distinguishing a genuine empty issues list from a forge fetch failure. When `listIssues` returns `error != null` (or the promise rejects), `loadError` is set and the panel renders `common_issues_load_failed()` instead of the misleading `common_no_open_issues()`. This is the right fix for the reported symptom.

## Verification

- `error?: string | null` is a real field on the `listIssues` return type — verified `ui/src/lib/api.ts:766` (and documented at `api.ts:763-766`). So `r.error != null` at `IssuesPanel.svelte:122` is well-typed.
- i18n parity holds: both keys exist in both catalogs — `common_issues_load_failed` and `common_no_open_issues` at `ui/messages/en.json:26-27` and `ui/messages/de.json:26-27`. Only `en`/`de` catalogs exist (verified via glob of `messages/*.json`), so no other locale is missing the keys.
- Template ordering at `IssuesPanel.svelte:201-208` is sound: `loading` → `loadError` → `slug === null` → empty. The new `loadError` branch precedes the empty-list branch, so a failed fetch no longer falls through to "no open issues".
- The `$effect` resets `loadError = false` on each repo change (`IssuesPanel.svelte:108`), so stale error state from a previous repo is cleared at the start of the new load.
- Tests added at `IssuesPanel.browser.test.ts:82-118` cover both the genuine-zero and the `error: "fetch_failed"` cases and assert the correct message appears (and the other does not).

## Non-blocking observation

The success path guards against stale writes with `if (rp !== repoPath) return;` (`IssuesPanel.svelte:117`), but the `.catch` block (`IssuesPanel.svelte:125-128`) — which this PR modifies to add `loadError = true` — has no such guard. If the user switches repos while a request is in flight and the *old* repo's request rejects after the *new* repo's effect has already run (and possibly already resolved), the unguarded catch can set `loadError = true` against the current repo, showing a spurious "couldn't load" banner that may stick. The pre-existing catch already wrote `loading = false` unguarded, so this is a narrow, timing-dependent edge case rather than a core-feature bug — but since the catch is now writing a user-visible error flag, mirroring the `.then`'s staleness guard here would close the gap.

kai-osthoff (commented):
## Re-review

The single prior finding is fully addressed.

**Prior finding (catch-block staleness guard):** The `.catch` handler in `IssuesPanel.svelte` now opens with `if (rp !== repoPath) return;` (line 129) before setting `loadError = true`, exactly mirroring the `.then` success-path guard at line 117. A stale rejection from a previously-selected repo can no longer stamp a sticky load-failed banner onto the repo now showing. Resolved.

### Verification performed
- `error` field on the listing contract exists: verified against `ui/src/lib/api.ts:766` (`error?: string | null`), so `loadError = r.error != null` at `ui/src/lib/components/IssuesPanel.svelte:122` is well-typed.
- Message-key locale parity: `common_issues_load_failed` and `common_no_open_issues` both present in `ui/messages/en.json:26-27` and `ui/messages/de.json:26-27`. Only `en` and `de` catalogs are maintained (`ui/messages/`), both in parity.
- New template branch `{:else if loadError}` (`ui/src/lib/components/IssuesPanel.svelte:207`) renders `common_issues_load_failed` ahead of the `issues.length === 0` empty branch, correctly distinguishing a fetch failure from a genuine zero.
- `$effect` resets `loadError = false` on repo change (`ui/src/lib/components/IssuesPanel.svelte:108`) so the banner does not persist across repo switches.
- New tests in `ui/src/lib/components/IssuesPanel.browser.test.ts` cover both the genuine-zero and `error: "fetch_failed"` cases.

No new defects found.

PR #1231, opened 2026-06-29
fix(upnext): surface fetch failure instead of false "all caught up"

kai-osthoff:
## Problem

The **Up Next** lens (DE *NÄCHSTES*) showed *"You're all caught up — nothing queued to start"* even with 22 tasks and several repos. The reporter rightly suspected we were **hiding a fetch failure behind the empty state**.

Root cause in `src/up-next.ts`: when a repo's `listIssues()` threw (e.g. GitHub rate limit / auth / network), `resolveRepo` silently `return null`'d and the repo was dropped. If **every** repo's fetch failed, `buildSnapshot` still produced a snapshot with `generatedAt` set and **empty `sections`** → `UpNextPanel` rendered the "all caught up" empty state. A fetch failure masquerading as an empty backlog.

Two failure surfaces, both fixed:
1. **Server:** every-/some-repo issue fetch errors into an empty queue.
2. **Client:** the `GET /api/up-next` call itself throws (herdr down / non-2xx) → previously stuck on *loading* forever.

This is the exact analog of the IssuesPanel fix in #1221.

## Fix

- **Server** (`up-next-core.ts`, `up-next.ts`): `UpNextSnapshot` gains `failedRepoCount`. `resolveRepo` now distinguishes a genuine fetch failure (`"fetch_failed"`) from a benign skip (forge vanished); `compute()` counts the failures into the snapshot.
- **Client store** (`up-next.svelte.ts`): tracks `loadError` when the GET throws; a fresh WS push clears it.
- **Panel** (`UpNextPanel.svelte`): renders the existing `common_issues_load_failed` message **before** the empty state when `loadError` OR (computed && `failedRepoCount > 0` && the **unfiltered** snapshot is empty). Keying off the unfiltered snapshot means a legitimately-empty repo filter still reads as empty, and a partial failure that still returns work renders the queue unchanged. The header ⟳ retries.

## Tests

- `test/up-next.test.ts`: a repo whose `listIssues` throws is now asserted to flag `failedRepoCount: 1` / `repoCount: 0`.
- `UpNextPanel.browser.test.ts`: 4 new cases — all-repos-failed → error; GET throws → error; genuinely-empty → all-caught-up (not error); partial failure with work → queue renders, no error.
- Reuses the existing `common_issues_load_failed` key (EN+DE already in parity) — no new i18n keys.

Verified: root lint, 42 server tests, 7 UI browser tests, `svelte-check` (0 errors), `check:i18n` (parity) all green.

`[no-feature-entry]` — bug fix, no new user-facing capability.

### DIFF
--- a/src/up-next-core.ts
+++ b/src/up-next-core.ts
@@ -47,10 +47,13 @@ export interface UpNextSection {
 export interface UpNextSnapshot {
   generatedAt: number;
   sections: UpNextSection[];
-  /** Forge-backed repos scanned this refresh. */
+  /** Forge-backed repos successfully scanned this refresh (excludes ones whose fetch failed). */
   repoCount: number;
   /** Degraded rung applied this refresh (e.g. "warm-repos-only"), or null when clean. */
   fallback: string | null;
+  /** Repos whose issue fetch threw this refresh and were dropped. >0 means a fetch failure may
+   *  be hiding work, so the UI surfaces a load error instead of an "all caught up" empty state. */
+  failedRepoCount: number;
 }
 
 /** A resolved epic, as fed in by the service (after buildEpic + selectEpicCandidates). */
@@ -206,6 +209,7 @@ export function buildSnapshot(
   repos: RepoInput[],
   now: number,
   fallback: string | null = null,
+  failedRepoCount = 0,
 ): UpNextSnapshot {
   const rank = warmRanks(repos);
   const all = repos.flatMap((r) => repoItems(r));
@@ -258,7 +262,7 @@ export function buildSnapshot(
     });
   }
 
-  return { generatedAt: now, sections, repoCount: repos.length, fallback };
+  return { generatedAt: now, sections, repoCount: repos.length, fallback, failedRepoCount };
 }
 
 /**
--- a/src/up-next.ts
+++ b/src/up-next.ts
@@ -120,10 +120,9 @@ export class UpNextService {
     const resolved = await mapBounded(repos, this.concurrency, (r) =>
       this.resolveRepo(r, lastUsed),
     );
-    const snap = buildSnapshot(
-      resolved.filter((r): r is RepoInput => r !== null),
-      this.now(),
-    );
+    const ok = resolved.filter((r): r is RepoInput => r !== null && r !== "fetch_failed");
+    const failedRepoCount = resolved.filter((r) => r === "fetch_failed").length;
+    const snap = buildSnapshot(ok, this.now(), null, failedRepoCount);
     this.snap = snap;
     try {
       this.deps.onChange(snap);
@@ -136,14 +135,16 @@ export class UpNextService {
   private async resolveRepo(
     r: UpNextRepo,
     lastUsed: Record<string, number>,
-  ): Promise<RepoInput | null> {
+  ): Promise<RepoInput | "fetch_failed" | null> {
     const forge = this.deps.resolveForge(r.repoPath);
-    if (!forge) return null;
+    if (!forge) return null; // forge vanished mid-compute → benign skip, not a fetch failure
     let openIssues: Awaited<ReturnType<GitForge["listIssues"]>>;
     try {
       openIssues = await forge.listIssues();
     } catch {
-      return null; // forge/network error → drop the repo this refresh (empty section omitted)
+      // forge/network error (e.g. rate limit) → drop the repo, but flag it as a failure so the
+      // snapshot can surface a load error rather than looking "all caught up" (cf. #1221).
+      return "fetch_failed";
     }
     const summaries = (await forge.listSubIssueSummaries?.().catch(() => null)) ?? {
       summaries: new Map<number, { total: number; completed: number }>(),
--- a/test/server-up-next.test.ts
+++ b/test/server-up-next.test.ts
@@ -22,6 +22,7 @@ const SNAP: UpNextSnapshot = {
   generatedAt: 123,
   repoCount: 1,
   fallback: null,
+  failedRepoCount: 0,
   sections: [],
 };
 
@@ -167,6 +168,7 @@ test("GET /api/up-next?peek strips hidden-repo sections via hiddenRepoPathsRaw",
     generatedAt: 999,
     repoCount: 2,
     fallback: null,
+    failedRepoCount: 0,
     sections: [hiddenSection, visibleSection],
   };
   const { app } = harness({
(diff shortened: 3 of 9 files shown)


kai-osthoff (commented):
## Summary

The PR addresses the task: when issue fetches fail, the Up-Next lens no longer masquerades as "all caught up" but surfaces a load error. The mechanism is sound on both server and UI:

- Server: `resolveRepo` now returns a `"fetch_failed"` sentinel on `listIssues()` throw (vs `null` for a benign missing-forge skip), and `compute()` counts those into `failedRepoCount`, threaded into `buildSnapshot` (verified `src/up-next.ts:123-125`, `src/up-next-core.ts:209-265`).
- UI: `loadFailed` triggers either when the GET itself threw (`upNext.loadError`) or when the server reported `failedRepoCount > 0` with an empty unfiltered queue (`ui/src/lib/components/UpNextPanel.svelte:55-58`), and it takes precedence over the empty state in the body (`UpNextPanel.svelte:234-241`).

### Verification performed
- i18n parity confirmed: `common_issues_load_failed` present in both catalogs (`ui/messages/en.json:26`, `ui/messages/de.json:26`).
- `failedRepoCount` was made a required field on both `UpNextSnapshot` interfaces (`src/up-next-core.ts:56`, `ui/src/lib/types.ts:1224`); I grepped every literal constructor of `UpNextSnapshot` and confirmed all were updated (the four test files in the diff + `excludeHiddenSections` which spreads `...snap`, preserving the new field — `src/up-next-core.ts:304-308`). No un-updated constructor remains.
- The only production caller of the up-next `buildSnapshot` is `src/up-next.ts:125`; the session-snapshot `buildSnapshot` is an unrelated function (`src/session-snapshot.ts:30`). The new `failedRepoCount` param defaults to `0`, so the test callers using the default still typecheck (`test/up-next-core.test.ts:56,275`).
- The `"fetch_failed"` string sentinel is disjoint from `RepoInput` (an object interface), so the `r !== null && r !== "fetch_failed"` type guard is sound (`src/up-next.ts:123-124`).
- `fallback` is passed explicitly as `null`, matching the prior implicit default — no behavior change there.

## Findings

**1. `upNext.loadError` unconditionally hides a still-valid cached snapshot (non-blocking).** `loadFailed = upNext.loadError || (...)` (`UpNextPanel.svelte:55-57`). At app load `+page.svelte:1274` does `upNext.load({ peek: true })`, populating `snapshot` with real work. When the lens mounts, `onMount` calls `upNext.load()` (no peek) which resets `loadError=false` then awaits the GET; if that GET throws (e.g. rate limit), `loadError` becomes `true` (`ui/src/lib/up-next.svelte.ts:18-24`) and the body renders the error message instead of the previously-visible cached queue. Before this PR the `catch` left the cached snapshot on screen. Recovery exists (a WS `upnext:snapshot` push calls `apply()` which clears `loadError` — `up-next.svelte.ts:28-31`), and the failed GET likely doesn't kick a server recompute, so the user must press ⟳ to recover. Consider only letting `loadError` win when there is no usable cached work (e.g. `sections.length === 0`), so a transient refresh failure doesn't blank out work the user could act on.

- ui/src/lib/components/UpNextPanel.svelte: `loadFailed` is keyed off `snap.sections` while `excludeHiddenSections` strips hidden-repo sections server-side and carries `failedRepoCount` forward via `...snap` (`src/up-next-core.ts:304-308`). If a repo's fetch fails (`failedRepoCount=1`) and the user has hidden the only repo(s) that had work, the GET returns empty `sections` with `failedRepoCount>0`, so the lens shows a load error even though the emptiness is due to hiding, not the failure. Narrow edge, low severity.

- The panel's ⟳ refresh handler (`UpNextPanel.svelte:137-147`) calls `refreshUpNext()` but does not itself reset `loadError`; recovery from a stuck `loadError` depends on the resulting WS push reaching `apply()`.

kai-osthoff (commented):
## Re-review

The single prior finding is **addressed**. `UpNextPanel.svelte:58-61` now gates the client-side error on usable cached work:

```js
const loadFailed = $derived(
  (computed && (snap?.failedRepoCount ?? 0) > 0 && (snap?.sections.length ?? 0) === 0) ||
    (upNext.loadError && sections.length === 0),
);
```

A failed lens-open GET after a successful app-load peek now keeps painting the cached queue instead of blanking to an error. New regression test `ui/src/lib/components/UpNextPanel.browser.test.ts` ("keeps painting cached work when a lens-open fetch throws") locks this in.

### Verification performed
- i18n parity: `common_issues_load_failed` (de.json:26 / en.json:26), `upnext_empty` (de.json:2075 / en.json:2075), `common_loading` (de.json:4 / en.json:4) — all present in both catalogs.
- `buildSnapshot` (up-next-core.ts:208) sole caller is up-next.ts:125; the `failedRepoCount` default (`= 0`) keeps the test-only/older callers compiling, and the return at up-next-core.ts:265 always sets it. The `buildSnapshot` in src/session-snapshot.ts is an unrelated symbol.
- `failedRepoCount` is preserved across `excludeHiddenSections` via `...snap` (up-next-core.ts:304-308), so the GET-time hidden-repo filter doesn't drop it.
- `"fetch_failed"` sentinel: the `r is RepoInput` guard at up-next.ts:123 excludes both `null` and the string; `failedRepoCount` counts the string entries at up-next.ts:124. Correct.
- UI read uses `snap?.failedRepoCount ?? 0`, so a pre-change cached snapshot lacking the field degrades safely.

### Non-blocking gap (repo filter + per-repo failure granularity)
The server emits only an aggregate `failedRepoCount`, with no record of *which* repo failed. Under an active `repoFilter`, the server-side error gate keys off the **unfiltered** `snap.sections` (deliberately, per the comment at UpNextPanel.svelte:53-55). Reachable case today (repo filter shipped in cc2cbff7):

- repoA's fetch fails → `failedRepoCount = 1`, no repoA section.
- repoB succeeds with work → `snap.sections = [repoB]`.
- User filters to **repoA**: server-side gate is false (`snap.sections.length !== 0`), `loadError` is false (GET succeeded), filtered `sections = []` → `isEmpty` → the panel shows `upnext_empty` ("all caught up") for repoA even though repoA's fetch actually failed.

This is the exact #1221 symptom (a fetch failure masquerading as an empty backlog) surviving in a narrow filtered case. It is not a regression (the pre-PR code behaved the same), and a clean fix needs per-repo failure info rather than an aggregate count, so it is non-blocking — flagging so the author can decide whether to track per-repo failures or accept the limitation.

kai-osthoff:
@scoop ready

Issue #1233, opened 2026-06-29
Batch PR polling per-repo (collapse per-session/per-PR gh pr list fan-out)


scoop:
Follow-up to #1230 (which landed cut-backs 1, 3, 4, 5). This is **item 2** from that issue, deferred because it rewrites the dense `prStatus` polling path and carries the most regression risk.

## Problem

During **attended** hours, #1230 does not reduce the per-session full-sweep or per-PR fast-sweep fan-out. A many-session / many-PR operator still pays:
- **Full sweep** — one `gh pr list --head <branch>` per active session every 120s (`src/pr-poller.ts` `tick()` → `forge.prStatus`).
- **Fast sweep** — one `gh pr list --head <branch>` per transient open PR (now activity-aware after #1230, but still per-PR).

Both are O(sessions + PRs) and all on the GraphQL bucket.

## Proposed change

Replace the N× `gh pr list --head <branch>` calls with **one** `gh pr list --state open --limit 200` **per repo** (the existing `forge.listPullRequests()`), then match results to sessions/PRs locally by `headRefName`. Collapses both sweeps O(sessions + PRs) → **O(repos)**.

Expected: cuts the attended residual (~300/hr per-session full sweep in the #1230 profile) down toward per-repo cost.

## Why it's risky (the reason it was deferred)

The per-session `prStatus` path carries dense, well-tested logic that must be preserved when switching to a per-repo batch + local match:
- `guardStaleTerminal` / `trustsTerminal` (reused-branch-name terminal PRs)
- `ownsPr` head-commit ownership check
- `reconcileBranch` (agent renamed the worktree branch)
- fork handling (`forkOwner`, `headRepositoryOwner`, the `--limit 30` fork case)
- `annotateHandoff`

Each of these must keep working when the PR is selected from a batched list by `headRefName` rather than fetched by exact `--head`. Needs careful test coverage mirroring the existing `test/pr-poller.test.ts` cases.

## Acceptance
- Steady-state attended GraphQL volume for the full + fast sweeps scales O(repos), not O(sessions + PRs).
- All existing pr-poller behaviors (stale-terminal guard, ownership, branch reconcile, fork mode, handoff) preserved with tests.

scoop:
✅ PR #1242 merged.

PR #1236, opened 2026-06-29
fix(spawn): strip NUL bytes from transient-agent prompt argv (#1235)

### LINKED ISSUE #1235: Plan review (and all transient-agent spawns) crash with "error" when prompt text contains a NUL byte
## Summary

A plan review never spawns — and reports a misleading `"error"` ("Couldn't start the plan review. Try again.") — when the plan text (or any untrusted text embedded into a transient agent's prompt) contains a **NUL byte (`\0`)**. Node/Bun's `child_process.spawn` rejects any argv argument containing a NUL, so `PlanGateService.begin()` throws at the `herdr.start` call and returns `"error"` on **every** attempt, auto or manual. The session is then stranded in `planPhase: "planning"` / `changes_requested` forever.

Observed on a real session (`batch-pr-polling`, issue #1233, desig TASK-1037), stuck at `REWORK · 1/5`.

## Evidence

`shepherd.log`, identical on all auto + manual attempts:

```
[plan-gate] spawn failed for c5017e07-…: TypeError: The argument 'args[141]'
must be a string without null bytes. Received "You are an adversarial plan reviewer…
```

The plan file `.shepherd-plan.md` contained exactly one NUL, at byte 8823, where the planning agent described the batch dedup key using the classic composite-key delimiter idiom:

```
`buildBatches` keys by the composite `${slug}\x00${forkOwner ?? ""}` when `slug` is non-null…
```

`planReviewPrompt()` embeds the plan text verbatim (`plan-gate.ts:255,63`); the prompt is then passed as the **trailing argv positional** by `buildTransientAgentArgv` → `reviewerArgv` → `herdr.start` (`plan-gate.ts:322-334`). The NUL rides into argv and `spawn` throws.

## Why it's unrecoverable from the UI

- The NUL is **durably persisted** in the plan file, so every retry fails identically — "Try again" can never succeed.
- `begin()` returns `"error"` for a `spawn` throw (`plan-gate.ts:330-334`), which the UI surfaces as a generic retry toast (`PlanPanel.svelte:127`).
- After the agent revises + exits, the session is terminal `done`, so the auto-trigger (`index.ts:1309`, fires on the `status === "done"` edge while `planPhase === "planning"`) has no fresh edge left either.

## Blast radius

`buildTransientAgentArgv` is the single shared builder behind ~11 transient-agent spawn sites — plan reviewer, PR critic (`review.ts:571`, `standalone-critic.ts:414`), recap, namer, distiller, doc-agent, merge-suggest, optimizer, herd-digest, autopilot-llm, verify-key. **Any** of them that embeds untrusted agent/repo text (plan, PR diff, issue body, repo rule text, recap content) will hit the same crash on a stray NUL. The plan reviewer is just the first to be tripped, because the agent literally wrote `\0` into the plan.

## Recommended fix

Sanitize control NULs (and arguably other argv-illegal control chars) from the prompt **inside `buildTransientAgentArgv`**, before assembling argv. One change closes all spawn sites at once (defense-in-depth). Replacing `\0` with a visible escape (e.g. the literal `\0`) preserves the surrounding text's meaning for the reviewer.

Secondary: a NUL-induced `spawn` throw is an *unrecoverable input* condition, not a transient failure — consider distinguishing it so the UI doesn't tell the operator to "Try again" on a guaranteed-to-fail retry.

## Repro

1. Have a planning agent write a literal `\0` byte anywhere in `.shepherd-plan.md` (e.g. a composite-key example).
2. Trigger plan review (auto on settle, or "Re-review plan").
3. `begin()` returns `"error"`; toast "Couldn't start the plan review. Try again."; session stuck at `REWORK`.

## Immediate operational unstick (already applied to TASK-1037)

Remove the NUL from the live `.shepherd-plan.md` (replaced the raw byte with the literal `\0`); the changed hash also clears the re-review dedup, so the next "Re-review plan" spawns normally.

scoop:
## Problem

`child_process.spawn` rejects any argv argument containing a NUL byte (`\0`). `buildTransientAgentArgv` (`src/transient-agent-argv.ts`) pushes the prompt verbatim as the trailing argv positional, so a single stray `\0` in agent-written text threw at the `herdr.start` call, returned a misleading `"error"` ("Couldn't start the plan review. Try again."), and stranded the session forever — every retry failed identically because the NUL was durably persisted.

Observed on a real session (`batch-pr-polling`, #1233, TASK-1037) stuck at `REWORK · 1/5` after the planning agent wrote `${slug}` + NUL + `${forkOwner}` into `.shepherd-plan.md`.

Fixes #1235.

## Fix

Sanitize the prompt inside the single shared builder `buildTransientAgentArgv`, immediately before assembling argv: replace each NUL with its visible 2-char escape `\0`. The prompt is the **only** argv slot that carries untrusted free text (the `--settings` JSON is NUL-safe — `JSON.stringify` escapes a NUL to the six ASCII chars backslash-u-0-0-0-0 — and the model/flags/session-id are controlled), so one change closes the crash at **all ~11 transient-agent spawn sites** (defense-in-depth): plan reviewer, PR critic, standalone critic, recap, namer, distiller, doc-agent, merge-suggest, optimizer, herd-digest, autopilot-llm, verify-key.

- **NUL only.** NUL is the single character `spawn()` actually rejects; other C0 control chars are legal in argv and are left untouched.
- **Visible escape, not strip.** Replacing with `\0` keeps the surrounding text legible to the reading agent (e.g. composite-key idioms) instead of silently fusing the two sides. Matches the operational unstick already applied to TASK-1037.

## Tests

`test/transient-agent-argv.test.ts`:
- For every kind (× model present/absent): a NUL-bearing prompt yields **no raw NUL anywhere in argv** and a literal `\0` in the trailing positional.
- Multiple NULs each escaped.
- NUL-free prompts pass through **byte-for-byte unchanged** (guards the existing byte-identity regression gate).

`bun run lint` and `bun run test` (5347 pass / 0 fail) green.

## Deferred follow-up

The issue's *secondary* suggestion — distinguishing an unrecoverable-input spawn failure from a transient one so the UI stops telling the operator to "Try again" on a guaranteed-to-fail retry — is intentionally **out of scope** here. It touches the shared `PlanReviewTrigger` return type + the `PlanPanel` toast + i18n, and with NUL now sanitized the reported crash can no longer recur. Worth a separate change if the team wants `begin()` to surface unrecoverable conditions distinctly.

[no-feature-entry] — server-only plumbing fix, no user-facing UX.

### DIFF
--- a/src/transient-agent-argv.ts
+++ b/src/transient-agent-argv.ts
@@ -149,7 +149,14 @@ export function buildTransientAgentArgv(
   argv.push("--allowedTools", ...preset.allowedTools);
   if (opts.model) argv.push("--model", opts.model);
   argv.push("--permission-mode", "dontAsk");
-  argv.push(opts.prompt);
+  // child_process.spawn REJECTS any argv arg containing a NUL ("must be a string without null
+  // bytes") — a hard throw, not a transient failure. The prompt is the ONLY argv slot that carries
+  // untrusted free text (plan / PR diff / issue body / repo rule / recap), and a stray \0 in
+  // agent-written text (e.g. the composite-key idiom `${slug}\0${forkOwner}`) would crash EVERY
+  // transient spawn site (issue #1235). Replace each NUL with its visible 2-char escape `\0` rather
+  // than stripping it, so the surrounding text's meaning survives for the reading agent. NUL is the
+  // only spawn-illegal char; other control chars are legal in argv and left untouched.
+  argv.push(opts.prompt.replaceAll("\0", "\\0"));
 
   return { argv, sessionId };
 }
--- a/test/transient-agent-argv.test.ts
+++ b/test/transient-agent-argv.test.ts
@@ -80,6 +80,36 @@ test("each call mints a fresh session id, echoed into --session-id", () => {
   expect(a.argv[a.argv.indexOf("--session-id") + 1]).toBe(a.sessionId);
 });
 
+// ── NUL sanitization: spawn() rejects a NUL in any argv arg (issue #1235) ────────────────────────
+
+test("every kind: a NUL in the prompt is escaped to `\\0`, leaving NO raw NUL anywhere in argv", () => {
+  for (const kind of ALL_KINDS) {
+    for (const model of [null, "claude-opus-4-8"]) {
+      // A composite-key idiom is the canonical way a stray NUL lands in agent-written text.
+      const { argv } = buildTransientAgentArgv(kind, {
+        model,
+        prompt: "keys by `${slug}\0${forkOwner}` when slug is non-null",
+      });
+      // No argv arg may contain a raw NUL — child_process.spawn would throw on it.
+      for (const arg of argv) expect(arg.includes("\0")).toBe(false);
+      // The trailing positional carries the visible 2-char escape where the NUL was.
+      expect(argv[argv.length - 1]).toBe("keys by `${slug}\\0${forkOwner}` when slug is non-null");
+    }
+  }
+});
+
+test("multiple NULs are each escaped", () => {
+  const { argv } = buildTransientAgentArgv("reviewer", { model: null, prompt: "a\0b\0c" });
+  expect(argv[argv.length - 1]).toBe("a\\0b\\0c");
+  for (const arg of argv) expect(arg.includes("\0")).toBe(false);
+});
+
+test("NUL-free prompts pass through byte-for-byte unchanged (no spurious escaping)", () => {
+  const prompt = "review the plan; key is `${slug}\\0${fork}` (already a literal escape)";
+  const { argv } = buildTransientAgentArgv("reviewer", { model: null, prompt });
+  expect(argv[argv.length - 1]).toBe(prompt);
+});
+
 // ── MCP-isolation coupling: ONE field drives BOTH flags, for every kind ─────────────────────────
 
 test("coupling: --safe-mode ⇔ enableAllProjectMcpServers, for every kind", () => {


scoop (commented):
## Summary

The PR sanitizes NUL bytes (`\0` → literal `\0`) inside `buildTransientAgentArgv`, the single shared builder behind all transient-agent spawn sites. This is exactly the issue's recommended fix: one change closes every spawn site at once.

## Verification

- `src/transient-agent-argv.ts:159` — `argv.push(opts.prompt.replaceAll("\0", "\\0"))`. The prompt is pushed last, after `--permission-mode dontAsk` (line 151), so it remains the trailing positional as the issue describes.
- The prompt is genuinely the only untrusted argv slot: `sessionId` is a `randomUUID()` (line 131), `--settings` is JSON built from controlled numeric/auth fields (lines 133-145), and `--model` is operator-controlled (line 150). None can carry a stray NUL, so sanitizing only the prompt is sufficient.
- `String.prototype.replaceAll` is available — already used at `src/service.ts:2587`, and `tsconfig.json` targets ESNext.
- Tests (`test/transient-agent-argv.test.ts:85-112`) cover all kinds (`ALL_KINDS`, line 34) × both model variants, assert no raw NUL survives anywhere in argv, assert multiple NULs each escape, and assert NUL-free prompts pass through byte-for-byte (no spurious escaping). The canonical composite-key idiom from the issue is used as the repro input.

The escape-rather-than-strip choice preserves surrounding text meaning for the reading agent, matching the issue's intent.

The issue's secondary suggestion (distinguish an unrecoverable-input spawn throw from a transient failure so the UI stops saying "Try again") is explicitly optional and lives in out-of-diff UI code; the sanitization fully prevents the crash, so omitting it does not block.

No blocking problems, logic bugs, or security issues found.

Issue #1241, opened 2026-06-29
Fast PR sweep (fastTick) → literal O(repos) per-repo batch


scoop:
Follow-up to #1233.

#1233 collapsed the **full** sweep (`PrPoller.tick`, 120s) to one per-repo batch (O(repos)). The **fast** sweep (`fastTick`, 15s) was left on its existing per-PR path: it polls at most `fastBatch` (=8) *transient* open PRs per tick (round-robin, time-bounded by `transientMaxMs`). That hard cap already makes it independent of sessions+PRs (it satisfies the issue's operative "not O(sessions+PRs)" criterion), but it is a **constant bound, not literal O(repos)**.

To finish the issue's literal "both sweeps O(repos)" goal, rework `fastTick`'s per-PR cap / round-robin / over-cap-log into per-repo work-units that reuse #1233's `buildBatches`/`batchForRepo`/count-gate. Deferred from #1233 because it reworks tested, dense logic for an already-bounded residual.

**Companion: trimmed custom `gh api graphql` batch variant.** If running the batch at the 15s fast cadence proves too costly on the GraphQL points budget, replace `gh pr list` with a hand-written `repository.pullRequests(states:OPEN,first:N)` query selecting exactly the `PrStatus` fields (+ `rateLimit{cost,remaining,resetAt}` fed into `graphRateLimit`). NOTE: this cannot reuse `rollupChecks`/`jobsFromRollup` — raw GraphQL `statusCheckRollup` returns a `CheckRun`/`StatusContext` union, not the flattened array `gh pr list --json statusCheckRollup` yields, so it needs a new, separately-tested GraphQL→`RollupEntry[]` mapper.

scoop:
✅ PR #1250 merged.

Issue #1240, opened 2026-06-29
Unify the three per-repo open-PR queries behind one cached snapshot


scoop:
Follow-up to #1233.

Three near-duplicate per-repo open-PR `gh pr list`/GraphQL queries now coexist, each with its own cadence/shape/owner:
- `GithubForge.listBacklogCounts` — overview counts (totalCount + first:100 nodes + CI rollup + rateLimit), backlog-poller cadence.
- `GithubForge.listPullRequests` — backlog PRs-tab rows (`PullRequest`), on-demand HTTP.
- `GithubForge.listOpenPrStatuses` — pr-poller per-repo batch (`PrStatus` keyed by headRefName), every full sweep (added in #1233).

The true O(repos) end-state is a single cached per-repo open-PR snapshot feeding all three consumers, but that is a cross-cutting refactor across the backlog system. #1233 deliberately scoped to the pr-poller path only. Worth doing if/when per-repo GraphQL pressure is revisited.

Cost basis (measured in #1233): one full-rollup `gh pr list --state open` ≈ 0.4 GraphQL pts/open-PR; a number-only count ≈ 1 pt.

scoop:
✅ PR #1253 merged.

Issue #1230, opened 2026-06-29
Reduce excessive GitHub GraphQL polling (rate-limit pressure)


scoop:
## Problem

Some operators (notably Kai, with more repos/sessions/open PRs) are hitting GitHub's **GraphQL** rate limit. GraphQL has its own **5,000 points/hour** bucket (per token), separate from REST. Our background pollers are GraphQL-heavy and multiply per-session, per-PR, and per-repo, so a busy setup sustains 3,000–5,000+ GraphQL calls/hr and trips the limit.

## Key fact: most polling is GraphQL, not REST

The `gh` CLI's `pr`/`issue`/`repo`/`search` subcommands run on `POST /graphql` (verified via `GH_DEBUG=api`). Only `gh run …` and `gh api <rest-path>` use REST. So nearly every polling path counts against the GraphQL bucket:

| Forge method | `gh` command | Bucket |
|---|---|---|
| `prStatus` (hot path) | `gh pr list --head` | **GraphQL** |
| `listBacklogCounts` | `gh api graphql` | **GraphQL** |
| `listPullRequests` (critic) | `gh pr list --state open` | **GraphQL** |
| `listIssues` / `getIssue` / `listIssueComments` | `gh issue list/view` | **GraphQL** |
| `listPrComments` / `prChangedPaths` / `prReviewMeta` | `gh pr view` | **GraphQL** |
| `defaultBranch` / `canPush` | `gh repo view` | GraphQL (lifetime-cached, fine) |
| `listWorkflowRuns` / `listRunJobs` | `gh run …` | REST (different bucket) |
| branch/sub-issue/collaborator/`currentUser` | `gh api <rest>` | REST |

## The consumers that run forever on a timer

1. **PR poller fast sweep** — `src/pr-poller.ts:306`, **every 15s**, up to 8 open PRs/tick → **~1,920 GraphQL calls/hr** globally. Re-polls *every* open PR including green ones merely awaiting human review (state can't change without a human/webhook).
2. **PR poller full sweep** — `src/pr-poller.ts:305`, **every 120s**, **one `gh pr list` per active session** → scales linearly with sessions.
3. **Backlog poller** — `src/backlog-poller.ts:52`, **every 45s per repo**, `listBacklogCounts` (`pullRequests(first:100){labels(first:10)}` — the costliest query point-wise) → 80 calls/hr/repo.
[...]

PR #1242, opened 2026-06-29
perf(pr-poller): batch PR polling per-repo (collapse full-sweep fan-out)

### LINKED ISSUE #1233: Batch PR polling per-repo (collapse per-session/per-PR gh pr list fan-out)

scoop:
Closes #1233. Follow-up to #1230 (item 2).

## What

Collapse the PR poller's **full sweep** from one `gh pr list --head <branch>` **per active session** (O(sessions) GraphQL calls every 120s) into **one `gh pr list --state open` per repo** (O(repos)), matching results to sessions locally by `headRefName`. GitHub-only; Gitea/Local transparently keep the per-session path.

## How

- **`GithubForge.listOpenPrStatuses()`** — one `gh pr list --state open --limit 200` → `Map<headRefName, PrStatus>`. Shares one extracted `mapGhPr` with `prStatus` for **field-mapping parity**; deterministic head-owner dedup (`forkOwner ?? owner(slug)`) — an intended, pinned divergence from `prStatus`'s order-dependent `prs[0]` in the rare cross-fork `headRefName` collision.
- **`GithubForge.countOpenPrs()`** — cheap `gh pr list --json number --limit 200` (~1 GraphQL pt) gating the batch.
- **`PrPoller.tick`** prefetches one batch per distinct repo (`buildBatches`, keyed by slug) and matches locally. `fastTick`/`refreshOne`/`pollSession` are unchanged.

## Cost guards (this is a GraphQL-**points** change, not just request count)

Measured: a full-rollup batch costs **~0.4 pt/open-PR** vs **~1 pt/session** per-session — so the batch only wins points when open-PRs (`P`) aren't ≫ sessions (`S`). Guards keep the steady state at the cheaper of {batch, baseline} per repo:

- **Fork mode → never batch** (upstream `P` = every contributor's PRs; e.g. a 100-PR upstream = ~44 pts × 30 sweeps/hr).
- **Non-fork count-gate** — batch only when `S ≥ 2` **and** `P ≤ 2·S` **and** `P < 200` (cap-hit); else per-session. The `S ≥ 2` floor avoids a single-session repo paying 2 calls where per-session paid 1.

**Verify (vs baseline, both axes):** common Shepherd repo (S=P=10) 300→**150** pts/hr & 300→**60** req/hr; low-S/high-P non-fork falls back (≤~30 pts/hr count-probe overhead, avoids a ~1200 pts/hr blow-up); fork upstream falls back with zero overhead. No points regression on any profile.

## Behavior preserved (all tested)

A `--state open` batch can't return terminal PRs, so a batch miss confirms via per-session `prStatus` (`--state all`) only when `marked || prev == null || prev.state !== "none" || recheckNone`. `reconcileBranch` still runs **every sweep** (local git) with a batch-first lookup, so open-rename adoption stays at the 120s cadence; only a terminal/absent-rename for a cached-`none` session coarsens to ≤ `noneRecheckMs` (600s) — `pollSession` adopts promptly. Stale-terminal guard (`guard`/`trustsTerminal`/`ownsPr`), handoff annotation, transient tracking, and the single-`gh` (`withGh`) invariant all hold under the batch path.

## Tests

`mapGhPr` parity, fork/non-fork owner dedup, cap-log latch, count-gate (S≥2 floor, ratio, P≥200), merged-transition confirm, stale-terminal-under-batch, rename-via-batch / rename-via-pollSession / rename-heal-after-window, cold-none confirm-then-skip, stuck-none heal, multi-repo one-call-per-slug, and a single-`gh` race (maxActive=1). Full suite **5357 pass / 0 fail**; lint, typecheck, `fallow audit`, branch hygiene clean.

## Out of scope / follow-ups

- Fast sweep (`fastTick`) → literal O(repos): #1241 (already bounded to a constant cap, so it meets "not O(sessions+PRs)" today). Trimmed custom-GraphQL variant rides the same issue.
- Unify the three per-repo open-PR queries behind one cached snapshot: #1240.

[no-feature-entry] server-only polling change, no user-facing UX.

### DIFF
--- a/src/forge/github.ts
+++ b/src/forge/github.ts
@@ -175,6 +175,7 @@ interface GhPr {
   isDraft?: boolean;
   statusCheckRollup?: RollupEntry[];
   headRefOid?: string;
+  headRefName?: string;
   baseRefName?: string;
   reviews?: GhReview[];
   reviewRequests?: { login?: string }[];
@@ -229,6 +230,9 @@ export class GithubForge implements GitForge {
    *  {@link forkSlug}. Used to qualify `pr create --head <forkOwner>:<branch>` and
    *  to disambiguate `prStatus`'s cross-repo match. Undefined for non-fork repos. */
   private readonly forkOwner?: string;
+  /** Latch: true once we've emitted the ≥200 open-PR cap warning so it fires at
+   *  most once per forge instance (on transition into the capped regime). */
+  private openPrCapLogged = false;
   constructor(
     readonly slug: string,
     private readonly cfg: ForgeConfig,
@@ -641,6 +645,31 @@ export class GithubForge implements GitForge {
     await this.run(["run", "cancel", String(runId), "--repo", this.slug]);
   }
 
+  /** Map a raw GhPr node to a PrStatus. Shared by prStatus (single-PR path) and
+   *  listOpenPrStatuses (batch path) so both produce identical field values. */
+  private mapGhPr(pr: GhPr, deployConfigured: boolean): PrStatus {
+    const state = pr.state.toLowerCase() as PrStatus["state"];
+    const createdAt = Date.parse(pr.createdAt ?? "");
+    return {
+      state: state === "open" || state === "merged" || state === "closed" ? state : "none",
+      number: pr.number,
+      url: pr.url,
+      title: pr.title,
+      createdAt: Number.isFinite(createdAt) ? createdAt : undefined,
+      mergeable: mapMergeable(pr.mergeable),
+      mergeStateStatus: mapMergeStateStatus(pr.mergeStateStatus),
+      isDraft: pr.isDraft ?? false,
+      checks: rollupChecks(pr.statusCheckRollup ?? []),
+      headSha: pr.headRefOid,
+      baseRefName: pr.baseRefName,
+      latestReview: latestHumanReview(pr.reviews),
+      requestedReviewers: (pr.reviewRequests ?? [])
+        .map((r) => r.login)
+        .filter((l): l is string => !!l),
+      deployConfigured,
+    };
+  }
+
   async prStatus(headBranch: string): Promise<PrStatus> {
     const deployConfigured = Boolean(this.cfg.deployWorkflow);
     // `gh pr list --head` matches by bare branch ref name — it does NOT accept the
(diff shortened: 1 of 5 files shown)


scoop (commented):
## Summary

This PR implements issue #1233: it replaces the O(sessions+PRs) per-branch `gh pr list --head` fan-out with one `gh pr list --state open --limit 200` batch per non-fork repo (`listOpenPrStatuses`) plus a cheap `countOpenPrs` probe gated by a cost ratio, and matches sessions to the batch locally by `headRefName`. I traced every preserved behavior the issue called out and they all hold. The implementation is careful; I found no blocking bugs. One doc-accuracy nit and one latent (future-reachable) concern, both non-blocking.

## Behavior-preservation verification

I verified each behavior the issue flagged as risky against the actual code:

- **stale-terminal guard / `trustsTerminal`**: The batch holds only OPEN PRs (`--state open`, github.ts:723). `guardStaleTerminal` is a no-op off-terminal (pr-poller.ts:30-42), so a batch hit (always open) passing through `guard` is identical to the per-session open path. A reused-branch terminal PR is absent from the open batch → batch miss → `needsConfirm` confirms via `forge.prStatus(s.branch)` which runs `guard`→`rejectStaleTerminal` (pr-poller.ts:441-447, 346-348). When `needsConfirm` is false it is because `prev.state==="none"` (pr-poller.ts:432), i.e. already downgraded — so staying none is correct. Preserved.
- **`marked` merge-train**: `guard` closes over `markedNumber` in `refresh` (pr-poller.ts:366-367). On open→merged, the PR drops out of the open batch, `needsConfirm=marked` is true (pr-poller.ts:432), `prStatus(s.branch)` returns the terminal PR, and `trustsTerminal(... marked && markedNumber===raw.number)` returns it untouched (pr-poller.ts:67). Preserved.
- **`ownsPr`**: reached via `rejectStaleTerminal`→`guardStaleTerminal` inside the same `guard` used on every batch-confirm and per-session path (pr-poller.ts:346-348). Preserved.
- **`reconcileBranch`**: still fired on a miss (pr-poller.ts:451); renamed branch is first looked up in the batch (no GraphQL, pr-poller.ts:454) then confirmed per-session when `needsConfirm` (pr-poller.ts:459). Local-git-only, so calling it every sweep on none-cached sessions is cheap and idempotent.
- **fork handling**: `batchForRepo` returns null for `forge.isFork` (pr-poller.ts:245), so fork repos always use the unchanged per-session `prStatus` with its `--limit 30` + `headRepositoryOwner` filter (github.ts:700-704). Preserved.
- **`annotateHandoff`**: applied once to the resolved raw status on both paths in `refresh` (pr-poller.ts:376). Preserved.

## Other correctness checks

- `mapGhPr` (github.ts:649-670) produces field-for-field the same `PrStatus` the old inline block built — verified each of the 14 fields against the pre-diff body. `prStatus` now just delegates to it (github.ts:707). Equivalent.
- Synthesized `noneGit` (pr-poller.ts:433-438) matches the per-session none shape (`prStatus` returns `{state:none,checks:none,deployConfigured}`). `deployConfigured: !!forge.deployWorkflow` equals `Boolean(cfg.deployWorkflow)` since `this.deployWorkflow = cfg.deployWorkflow ?? null` (github.ts:245). `gitStateChanged` ignores `deployConfigured` anyway (pr-poller.ts:74-89), so no spurious emit. Consistent.
- Truncated-batch safety: `countOpenPrs` and the cap warning both use `--limit 200`; `batchForRepo` bails on `p >= 200` (pr-poller.ts:250) so a truncated 200-row batch is never consumed. Good.
- Count-gate break-even (`count < 2` → null, pr-poller.ts:245) is justified: 2 gh calls (count+list) vs S per-session calls, break-even S≥2. Documented and correct.
- `forge.slug` null-guards present in both `buildBatches` and `batchFor` (pr-poller.ts:271, 289); `slug` is `string | null` (types.ts:301). Consistent.

## Nit

The JSDoc on `listOpenPrStatuses` (github.ts:713-714) and the interface (types.ts:481) both state "Fork mode keeps only PRs whose head lives on OUR fork (mirrors prStatus's headRepositoryOwner filter)." The implementation does NOT filter to fork-owned PRs — it includes every open PR and only applies an owner *preference* on a `headRefName` collision (github.ts:745-756) — and the method is never called in fork mode because `batchForRepo` skips forks (pr-poller.ts:245). The doc describes behavior that neither exists nor executes; recommend rewording to describe the actual collision-dedup-by-owner-preference behavior and that fork repos use the per-session path.

- src/forge/github.ts: if a follow-up enables fork batching by removing the `forge.isFork` guard in `batchForRepo` (pr-poller.ts:245) — which the misleading fork JSDoc actively invites — `listOpenPrStatuses` would mismatch fork PRs: for a branch whose only open PR lives on another fork, the dedup keeps it (first-seen, github.ts:749-750), whereas `prStatus` in fork mode returns `none` when no PR's `headRepositoryOwner` matches `forkOwner` (github.ts:703-706). Dormant today (fork repos never batch); would silently surface wrong PRs if that guard is lifted without adding the owner filter the JSDoc claims.

## Limitations
- Tests (test/pr-poller.test.ts, test/forge/github.test.ts) were not executed — per instructions CI owns tests; I reviewed the production paths only.

scoop (commented):
## Re-review of #1233 batch PR polling

### Prior point #1 — RESOLVED
The JSDoc on `listOpenPrStatuses` (src/forge/github.ts:710-719) and the interface entry (src/forge/types.ts:478-485) now describe the actual behavior: a deterministic headRefName collision-dedup that prefers the repo-owned PR but **does NOT filter by owner — every open PR is returned**, and explicitly note that fork-mode repos are routed around (`batchForRepo` skips `isFork` → per-session `prStatus`). Verified both comment blocks against the implementation at github.ts:732-756 and pr-poller.ts:245.

### Verification of behavior preservation (the stated regression risk)
- **guard / trustsTerminal / guardStaleTerminal / ownsPr**: the `guard` closure (pr-poller.ts:366-367) is applied on every batch result — hit (pr-poller.ts:430), confirm-on-miss (pr-poller.ts:443), live-hit (pr-poller.ts:455), live-confirm (pr-poller.ts:460) — and on the per-session path (pr-poller.ts:393,405). A batch hit is always `state:"open"`, where `guardStaleTerminal` is a documented no-op (pr-poller.ts:30-42), so open PRs pass through unchanged, matching old behavior.
- **Merge-train terminal transition**: a `marked` session's merge → batch miss → `needsConfirm` true (marked) → per-session `prStatus` confirms terminal → `trustsTerminal(marked && markedNumber===raw.number)` returns true (pr-poller.ts:67). Preserved.
- **reconcileBranch**: invoked on both paths (statusPerSession pr-poller.ts:402; statusFromBatch pr-poller.ts:451), and statusFromBatch additionally resolves the renamed branch against the batch first (no GraphQL, pr-poller.ts:454-455) before falling back to per-session.
- **Fork handling**: `batchForRepo` returns null on `forge.isFork` (pr-poller.ts:245), so fork repos keep the existing per-session `prStatus` with its `forkOwner`/`headRepositoryOwner`/`--limit 30` logic (github.ts:700-707). Verified `isFork` exists on GitForge (types.ts:311).
- **annotateHandoff**: applied once on the shared post-processing of the resolved raw status (pr-poller.ts:376).
- **Field parity**: `mapGhPr` (github.ts:649-670) is shared by `prStatus` (github.ts:707) and `listOpenPrStatuses` (github.ts:738/742), so batch and single-PR `PrStatus` objects are byte-identical in shape. `headRefName` was added to `GhPr` (github.ts:178) and is requested in the batch JSON field list (github.ts:733).
- **deployConfigured**: synthesized `noneGit()` uses `!!forge.deployWorkflow` (pr-poller.ts:437); `forge.deployWorkflow = cfg.deployWorkflow ?? null` (github.ts:245), so it equals `Boolean(cfg.deployWorkflow)` used in `prStatus`/`listOpenPrStatuses` (github.ts:674,721).
- **Count-gate / cap consistency**: `countOpenPrs` and `listOpenPrStatuses` both cap at `--limit 200`; the gate trips at `p >= 200` (pr-poller.ts:250) before the list call, so a truncated 200-cap batch is avoided. The count→list race (count 199, list 200) self-heals: previously-open/new sessions have `needsConfirm` true and fall back to per-session (pr-poller.ts:432).

### Noted behavior change (intentional, documented — not a defect)
For a session stuck at cached `none`, a PR that is *both created and reaches terminal between sweeps* (never observed `open`) is now discovered only on the coarse `noneRecheckMs` recheck (default 10 min, pr-poller.ts:192-196,432) rather than every 120 s. This is the deliberate cost-reduction tradeoff at the core of the PR and is bounded/documented; the common path (PR creation triggers `pollSession` → per-session refresh → `open`) is unaffected, so subsequent terminal transitions have `prev.state !== "none"` and confirm immediately.

### Could-not-verify / limitations
None — all referenced symbols, fields, and call paths were resolvable within this repo.

No blocking issues; nothing actionable remains.

PR #1244, opened 2026-06-29
fix(gitrail): surface forge-error fallback so the PR rail never renders empty

kai-osthoff:
## Problem

On the desktop focused view, clicking the **● PR ▾** disclosure sometimes revealed a strip with **no buttons** — no PR link, no Merge, no Ready/critic controls — so you couldn't open the PR (the user's report: *"fehlen jede menge schaltflächen — man kann nicht auf PR klicken um den offenen PR anzeigen"*).

## Root cause

`GitRail` fetches its PR status from `GET /api/sessions/:id/git`. That endpoint returns **502** when the forge call fails — e.g. when **GitHub is rate-limited** (fittingly, the session in the screenshot was the *github-rate-limits* one). On that throw the client set `git = null` **silently**:

- the entire rail lives under `{#if git}`, so a null state rendered **nothing** — no PR link, no error, no retry;
- the 15s recovery poll was gated on `git?.state === "open"`, so a `null` (failed) state **never retried** — the strip stayed permanently blank until remount.

The only thing left in the disclosure strip was the surrounding autopilot toggle → "missing buttons".

Reproduced deterministically by returning 502 from the git endpoint: the strip collapses to just the autopilot control.

## Fix

- Track the thrown-load case (`loadFailed`) separately from a legit 404 (which `gitState` already maps to a silent `null` for "no forge / no PR").
- Render an inline **"PR status unavailable" + Retry** fallback instead of an empty rail (reuses the existing `.err` / `.gbtn` recipe, EN+DE).
- Let the 15s poll keep retrying while the last load failed, so a transient rate-limit **clears itself** without operator action.

## Verification

- Drove the live UI with Playwright: 502 → error+Retry shown (no blank strip); Retry / poll after recovery → full PR rail restored (PR ↗, Merge, Ready, REVIEWED, Re-review, automation pill).
- Added 2 regression tests to `GitRail.browser.test.ts` (forge-error fallback + recovery). Full suite: **61 passed**.
- `check:i18n` (2114 keys parity), `bun run check` (0 errors) green.

[no-feature-entry] — bug fix, no new user-facing capability.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -92,6 +92,7 @@
   "gitrail_confirm_redeploy": "bestätigen ⟳",
   "gitrail_redeploy": "⟳ Redeploy",
   "gitrail_closed": "geschlossen",
+  "gitrail_status_failed": "PR-Status nicht verfügbar",
   "gitrail_pr_title_placeholder": "PR-Titel",
   "gitrail_pr_description_placeholder": "Beschreibung",
   "gitrail_cancel": "Abbrechen",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -92,6 +92,7 @@
   "gitrail_confirm_redeploy": "confirm ⟳",
   "gitrail_redeploy": "⟳ Redeploy",
   "gitrail_closed": "closed",
+  "gitrail_status_failed": "PR status unavailable",
   "gitrail_pr_title_placeholder": "PR title",
   "gitrail_pr_description_placeholder": "Description",
   "gitrail_cancel": "Cancel",
--- a/ui/src/lib/components/GitRail.browser.test.ts
+++ b/ui/src/lib/components/GitRail.browser.test.ts
@@ -525,6 +525,53 @@ describe("GitRail — controls stay within the cell", () => {
   });
 });
 
+describe("GitRail — forge-error fallback (no silent-empty rail)", () => {
+  // Regression: a thrown gitState (forge 502 — e.g. GitHub rate-limited) used to set
+  // git=null and render NOTHING — no PR link, no error, no retry. The desktop PR
+  // disclosure strip then showed only the surrounding autopilot control, so the
+  // operator saw "missing buttons" and couldn't open the PR. The fetch retry was also
+  // gated on git?.state === "open", so a null (failed) state never recovered.
+  it("desktop — gitState rejects → shows error + Retry instead of an empty rail", async () => {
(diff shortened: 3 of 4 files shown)


kai-osthoff (commented):
## Summary

The PR addresses the reported bug — on desktop a forge error (e.g. GitHub 502 / rate-limit) left the PR rail rendering nothing (no PR link, no buttons, no retry), so the operator saw "missing buttons" and couldn't open the PR. The fix introduces a `loadFailed` flag, an error+Retry fallback branch, and keeps the 15s poll running while a load has failed so a transient outage self-heals. The implementation is correct and well-scoped.

## Verification

- **Catch only fires on real errors, not on "no PR".** `gitState` (verified against `ui/src/lib/api.ts:881-885`) returns `null` on 404 and `throw`s only on other non-ok responses. So `catch { … loadFailed = true }` (`GitRail.svelte:121-128`) flags exactly the forge-error case; a legit 404 (no forge / no PR) still falls through to `git = null` with `loadFailed = false` and renders nothing, as before. The inline comment's "non-404" claim is accurate.
- **Render gap is closed.** The template is a single `{#if git} … {:else if loadFailed} … {/if}` chain (`GitRail.svelte:602`, `:792`). Previously a thrown state set `git = null` and matched neither branch → empty rail. The new branch (`:792-806`) surfaces the error + Retry.
- **Poll recovery.** `if (loadFailed || git?.state === "open") load(id)` (`GitRail.svelte:172`) keeps retrying after a failure; on success `load` clears `loadFailed` (`:119`) and on session change the `$effect` resets it (`:164`). No stale-closure issue — `$state` runes are read at call time.
- **i18n parity.** New key `gitrail_status_failed` added to both `ui/messages/en.json:95` ("PR status unavailable") and `ui/messages/de.json:95` ("PR-Status nicht verfügbar"); `common_retry` already present in both catalogs (grep: 2/2 in each).
- **Test identifiers resolve.** The new tests query `a.prlink`, which is the open-PR anchor rendered by `git-rail/RailStatusActions.svelte:95` (`<a class="prlink" href={git.url}>`); `.err` and `m.common_retry()`/`m.gitrail_status_failed()` all match the component. The success path asserts `href === openPrState.url`, consistent with the component.

No blocking problems, and no nits worth raising. The `disabled={busy}` on the new Retry button is harmless/defensive (correctly suppresses a re-fetch while a mutation is still in flight).

Issue #1249, opened 2026-06-29
fast-sweep batch: trimmed `gh api graphql` first:N variant (if 15s cadence too costly)


scoop:
Follow-up to #1241 (which reworked \`fastTick\` to reuse #1233's per-repo \`buildBatches\`/\`batchForRepo\` + count-gate).

#1241 deliberately deferred the issue's **contingent companion**: a trimmed custom GraphQL batch variant for the 15s fast cadence. It is gated on an empirical measurement — only build it if the per-repo \`listOpenPrStatuses\` batch (full per-PR rollup, up to 200 open PRs) proves too costly on the GraphQL points budget at 15s.

## Empirical gate (do first)

Measure the steady-state GraphQL points/hr the fast sweep adds **post-#1241** on a warm, many-open-PR repo (transient-dominant so it actually batches). The count-gate already falls back to per-session for mixed repos (P ≫ eligible), so the concern is specifically the transient-dominant batch path at 8× the full-sweep cadence. If it's within budget, **close this — no code.**

## If too costly: the variant

Replace \`gh pr list\` with a hand-written \`repository.pullRequests(states:OPEN, first:N)\` query selecting exactly the \`PrStatus\` fields, plus \`rateLimit{cost,remaining,resetAt}\` fed into \`graphRateLimit\`.

**Cannot reuse \`rollupChecks\`/\`jobsFromRollup\`:** raw GraphQL \`statusCheckRollup\` returns a \`CheckRun\`/\`StatusContext\` union, not the flattened array \`gh pr list --json statusCheckRollup\` yields. Needs a **new, separately-tested** GraphQL→\`RollupEntry[]\` mapper.

## Acceptance
- A measurement (before/after points/hr) justifying the variant, **or** a note closing this as unnecessary.
- If built: new GraphQL→\`RollupEntry[]\` mapper with its own tests; \`rateLimit\` cost wired into \`graphRateLimit\`.

scoop:
## Empirical gate: within budget → closing, no code

Ran the "do first" measurement against a live repo (scoop token, GraphQL bucket). `rateLimit` reads are free (verified delta 0), so these are true per-call GraphQL point costs:

| Call (measured directly) | Cost |
|---|---|
| **Production fast-sweep batch** — `gh pr list --state open --json <full PrStatus field set> --limit 200` | **5 pts** |
| Count-gate probe — `gh pr list --json number --limit 200` (`countOpenPrs`) | 1 pt |
[...]

PR #1250, opened 2026-06-29
perf(pr-poller): batch the fast sweep per-repo (drop fastBatch cap/round-robin) (#1241)

### LINKED ISSUE #1241: Fast PR sweep (fastTick) → literal O(repos) per-repo batch

scoop:
Closes #1241.

Follow-up to #1233, which collapsed the **full** sweep (`tick`, 120s) to one per-repo batch (`buildBatches`/`batchForRepo`/`batchFor` + count-gate). The **fast** sweep (`fastTick`, 15s) was left on its per-PR path: ≤ `fastBatch`(=8) transient open PRs/tick, round-robin, with an over-cap log. This reworks `fastTick` to reuse #1233's exact machinery over its eligible-transient subset.

## What changed

`fastTick` now: filter eligible (transient + within `transientMaxMs`) open PRs → resolve to live `Session`s → `buildBatches(eligible)` → `refresh(s, batchFor(s, batches))` under `withGh`. Removed: the `fastBatch` ctor param, `fastCursor`, `fastCapLogged`, the round-robin slice, and the over-cap `console.warn`.

## Honest bound (not an unconditional O(repos) collapse)

The count-gate is reused as #1233 directs, so it governs the fast path the same way: `count` = the **eligible-transient** sessions but `P = countOpenPrs()` = the repo's **total** open PRs.

- **Transient-dominant repo** (`P ≤ ratio × eligible`): one `listOpenPrStatuses`/tick — O(repos).
- **Mixed repo** (`P ≫ eligible`, stable PRs beyond the transient few): gate trips → bounded per-session (now uncapped/no-rotation). This is deliberate — feeding the gate all-branch session counts instead would green-light a full-rollup batch over all `P` PRs to refresh ~2 transient ones, a ~10× points regression at 8× cadence (the exact blow-up the deferred GraphQL variant guards against).

Literal single-batch O(repos) at the fast cadence is the gate's **ceiling**, not a guarantee. Making the fast batch cheap enough to apply more broadly is the deferred companion: **#1249** (trimmed `gh api graphql` `first:N` variant, gated on a points-cost measurement).

## Correctness notes

- `sweeping = true` is set **before** `buildBatches` (matching `tick`), so the `countOpenPrs`/`listOpenPrStatuses` probe runs inside the mutual-exclusion window — a concurrent full `tick` can't interleave and duplicate the sweep.
- The `inFlight` dedup the old `refreshOne`-based path provided is preserved: the loop skips any session already in `inFlight` and brackets each `refresh` with add/delete, so a concurrent `pollSession` doesn't double-poll.
- `fastTick` still does not prune the cache; transient-window / warm / rate-limit gating unchanged.

## Tests

Deleted the two cap/round-robin tests; removed the now-stale `fastBatch` positional arg from every surviving instantiation. Added, on `forgeWithBatch` with `OPEN_PENDING` (transient) entries:
- **(a)** transient-dominant repo → one `listOpenPrStatuses`, zero per-PR `prStatus`.
- **(b)** single eligible PR → `count < 2` → per-session.
- **(c)** 10 singleton repos → 10 per-session in one tick (exceeds old cap=8 → proves no cap).
- **(d)** 10 eligible in one repo → one batch covers all in a single tick (no rotation).
- **(e)** mixed repo `P=20 ≫ eligible=2` → count-gate trips → per-session, no full-rollup batch.

`bun run typecheck` / `bun run lint` / `bun run test` (root, 5448 pass) all green.

`[no-feature-entry]` — server-only polling change, no user-facing UX.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -754,7 +754,6 @@ const prPoller = new PrPoller(
   // is recognized instead of staying invisible against the stale stored branch
   (s) => service.syncWorktreeBranch(s.id),
   undefined,
-  undefined,
   ownsPr,
   warm, // full cadence only while warm; cold → fast sweep pauses, full sweep throttles
   () => graphRateLimit.blocked(), // skip every sweep while the GraphQL bucket is exhausted
--- a/src/pr-poller.ts
+++ b/src/pr-poller.ts
@@ -101,8 +101,12 @@ function gitStateChanged(prev: GitState | undefined, git: GitState): boolean {
  * `checks: "pending"` is short-lived and the 120s sweep routinely sampled `none`
  * then `success` two sweeps later, never recording the running state, so the
  * list jumped straight to green and never showed the pulsing "CI running" dot.
- * The fast sweep is capped at `fastBatch` open PRs per tick (round-robin beyond
- * it) so it never fans out one blocking `gh` per PR over an unbounded backlog.
+ * The fast sweep routes its eligible (transient, in-window) open PRs through the
+ * same per-repo batch path as `tick` (`buildBatches`/`batchForRepo` + count-gate):
+ * a repo whose transient PRs dominate its open set is refreshed with one
+ * `listOpenPrStatuses`; a repo with many stable open PRs beyond the few transient
+ * ones falls back to bounded per-session polls (the gate protects the 15s points
+ * budget). No per-PR cap or round-robin — every eligible PR is covered each tick.
  */
 export class PrPoller implements PrCache {
   private timer: ReturnType<typeof setInterval> | null = null;
@@ -119,10 +123,6 @@ export class PrPoller implements PrCache {
    *  up a backlog of queued ticks. The targeted poll does NOT consult it — it
    *  serializes via `ghChain` instead, so a turn-end poll is never dropped. */
   private sweeping = false;
-  /** Round-robin offset into the open-PR list when it exceeds `fastBatch`. */
-  private fastCursor = 0;
-  /** Latched so the over-cap notice logs once on transition, not every fast tick. */
-  private fastCapLogged = false;
   /** Tracks when each open PR entered (or re-entered via a new headSha) a transient
    *  state — used to time-bound how long `fastTick` keeps re-polling it. */
   private transientSince = new Map<string, { since: number; headSha?: string }>();
@@ -163,8 +163,6 @@ export class PrPoller implements PrCache {
     private reconcileBranch: (s: Session) => string | null = () => null,
     /** Fast cadence for re-polling open PRs (in-flight CI/merge state). */
     private fastIntervalMs = 15_000,
-    /** Max open PRs polled per fast tick; the rest rotate in on later ticks. */
-    private fastBatch = 8,
     /** Whether a name-matched terminal (merged/closed) PR's head commit actually
      *  belongs to the session's branch. Guards against `gh pr list --head <name>`
      *  returning a prior, already-merged PR that merely reused this branch name.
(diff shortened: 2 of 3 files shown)


scoop (commented):
## Summary

This PR reworks `fastTick` from its per-PR `fastBatch` cap / round-robin / over-cap-log into per-repo work-units that reuse #1233's `buildBatches`/`batchForRepo`/count-gate, achieving the issue's literal "both sweeps O(repos)" goal. The implementation is clean and the change is internally consistent.

## Verification

- **Constructor signature change is consistent.** `fastBatch` param removed at `src/pr-poller.ts:164`. The `index.ts` callsite drops one `undefined` (`src/index.ts:751-757`); positional alignment confirmed — args 4-10 now map to intervalMs/pollDelayMs/reconcileBranch/fastIntervalMs/ownsPr/warm/rateLimited (`src/index.ts:751-759`), so `ownsPr` correctly lands at position 8. Long-form test constructors still align: e.g. the cap-hit test passes `1000` at position 13 = `batchOpenRatio` with a matching comment (`test/pr-poller.test.ts:1699`).
- **No orphaned references.** Grep across `src`/`test` shows zero remaining `fastBatch`, `fastCursor`, or `fastCapLogged` — the two removed fields (`src/pr-poller.ts` old lines 124-128) and their early-return re-arm logic are fully cleaned up.
- **fastTick batch path is correct** (`src/pr-poller.ts:300-341`). It reuses `buildBatches` (`src/pr-poller.ts:261`), `batchFor` (`src/pr-poller.ts:281`), and the count-gate in `batchForRepo` (`src/pr-poller.ts:239-255`). The gate is deliberately fed only the eligible-transient session set, weighing a batch against the per-session calls fastTick would actually make this tick — documented at `src/pr-poller.ts:317-319` and sound.
- **Mutual exclusion preserved.** `this.sweeping = true` is set before `buildBatches` so the count/list probe runs inside the exclusion window (`src/pr-poller.ts:324-326`); all early returns (`open.length===0`, `eligible.length===0`, `sessions.length===0`) occur before the flag is set, so it is never leaked. `inFlight` per-session dedup is preserved from the old `refreshOne` path (`src/pr-poller.ts:330-336` vs `src/pr-poller.ts:503-513`).
- **Terminal transitions still detected.** A previously-open PR that merges drops out of the batch map; `statusFromBatch` then has `needsConfirm` true because `prev.state` is `"open"` (`src/pr-poller.ts:431`), so it falls through to a per-session `prStatus` confirm. Fast-poll transitions are not lost.
- **Forge contract.** `slug`, `isFork`, `listOpenPrStatuses`, `countOpenPrs` confirmed on the GitForge type (`src/forge/types.ts:301,311,486,491`); the `forge.listOpenPrStatuses!`/`countOpenPrs!` non-null assertions are guarded by the preceding existence checks (`src/pr-poller.ts:243`).
- **Tests.** Helper `forgeWithBatch` (`test/pr-poller.test.ts:1249-1287`) exposes the `{list,count,prStatus}` stats the new cases assert on; cases (a)-(e) (`test/pr-poller.test.ts:553-684`) exercise transient-dominant batch, single-eligible fallback, O(eligible) singleton fan-out (10 > old cap 8), single-tick coverage of 10 PRs, and the count-gate trip — all consistent with the code paths.
- **Removed `console.warn` over-cap notice** is intended per the task (eliminate per-PR cap/round-robin/over-cap-log); no i18n or user-facing strings are involved, so no locale parity concern.

The change satisfies the task and I found no bugs, security issues, or quality problems in the diff.

PR #1251, opened 2026-06-29
fix: prevent Claude model aliases on Codex spawns

kai-osthoff:
## Summary
- reject known Claude model aliases for explicit Codex model selections
- clamp carried/stored models to the target provider before spawning, so held tasks switched to Codex use the Codex default instead of passing `opus`
- add regression coverage for provider/model compatibility and Codex spawn argv

## Verification
- bun test test/default-model.test.ts test/validate.test.ts test/service.test.ts
- bun run lint
- bun run test
- bunx fallow@2.100.0 audit --base origin/main --fail-on-issues
- pre-push hook passed: prettier, eslint, gates, tsc, extension, root-tests, ui, fallow

### DIFF
--- a/src/default-model.ts
+++ b/src/default-model.ts
@@ -13,9 +13,18 @@
  * apply it — autonomous spawns must be deterministic and operator-controlled only.
  */
 
-import { AGENT_PROVIDERS, type AgentProvider, MODELS, MODELS_BY_PROVIDER } from "./types";
+import {
+  AGENT_PROVIDERS,
+  CODEX_MODELS,
+  CODEX_MODEL_RE,
+  type AgentProvider,
+  MODELS,
+  MODELS_BY_PROVIDER,
+} from "./types";
 
 const SETTING_VALUES = new Set<string>(["auto", "default", ...MODELS]);
+const CLAUDE_MODEL_VALUES = new Set<string>(MODELS);
+const CODEX_MODEL_VALUES = new Set<string>(CODEX_MODELS);
 
 // The per-repo override space is the global space plus an "inherit" sentinel,
 // which means "no repo override — fall back to the global default setting".
@@ -52,6 +61,31 @@ export function spawnModelForAvailability(
   fableAvailable: boolean,
 ): string | null {
   return model === "fable" && !fableAvailable ? "opus[1m]" : model;
+}
+
+/** True when `model` is a valid explicit model alias for `provider`.
+ *
+ * Codex accepts safe future-looking aliases so Shepherd can work with a newer installed CLI before
+ * the curated list is updated, but known Claude aliases are excluded from that free-form path.
+ */
+export function modelCompatibleWithProvider(
+  model: string | null,
+  provider: AgentProvider,
+): boolean {
+  if (model === null) return true;
+  if (provider === "claude") return CLAUDE_MODEL_VALUES.has(model);
+  return (
+    CODEX_MODEL_VALUES.has(model) || (CODEX_MODEL_RE.test(model) && !CLAUDE_MODEL_VALUES.has(model))
+  );
+}
+
+/** Clamp a stored model to the target provider's value space. Used when an existing task changes
+ * provider after its model was already validated for the original provider. */
+export function modelForProviderOrDefault(
+  model: string | null,
+  provider: AgentProvider,
+): string | null {
+  return modelCompatibleWithProvider(model, provider) ? model : null;
 }
 
 /**
@@ -150,6 +184,6 @@ export function resolveRoleEnvironment(
   const token = normalizeRoleModelToken(roleModel) ?? "default";
   if (token === "default") return { provider: cli, model: null };
   // Clamp: the stored model must belong to the chosen provider, else fall back to its default.
-  if (!MODELS_BY_PROVIDER[cli].includes(token)) return { provider: cli, model: null };
+  if (!modelCompatibleWithProvider(token, cli)) return { provider: cli, model: null };
   return { provider: cli, model: spawnModelForAvailability(token, fableAvailable) };
 }
--- a/src/service.ts
+++ b/src/service.ts
@@ -2006,7 +2006,7 @@ export class SessionService {
         model: spawnInput.model,
         auto: spawnInput.auto ?? false,
         issueNumber: spawnInput.issueRef?.number ?? null,
-        planGateEnabled: agentProvider === "claude" ? (input.planGateEnabled ?? null) : false,
+        planGateEnabled: agentProvider === "claude" ? (spawnInput.planGateEnabled ?? null) : false,
         autopilotEnabled: spawnInput.autopilotEnabled ?? null,
         planPhase: planGateOn ? "planning" : null,
         research: spawnInput.research ?? false,
--- a/src/service.ts
+++ b/src/service.ts
@@ -16,7 +16,6 @@ import type {
   RelaunchOverrides,
   Session,
 } from "./types";
-import { MODELS, CODEX_MODELS, CODEX_MODEL_RE } from "./types";
 import {
   copyStagedIntoWorktree,
   stagingDir,
@@ -26,7 +25,11 @@ import {
   worktreeUploadsDir,
 } from "./uploads";
 import { slugifyManual, isHeuristicNameStrong } from "./namer";
-import { spawnModelForAvailability } from "./default-model";
+import {
+  modelCompatibleWithProvider,
+  modelForProviderOrDefault,
+  spawnModelForAvailability,
+} from "./default-model";
 import {
   isApiKeyMode,
   isApiKeyConfigured,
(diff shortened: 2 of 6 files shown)


kai-osthoff (commented):
## Summary

The PR fixes the reported bug: a held/queued task created as Claude with `model: "opus"` that is then spawned as Codex no longer carries `opus` to the Codex `--model` flag. The fix is layered correctly and I could not find a correctness defect.

### Root cause confirmed

The old `modelCompatibleWith("opus", "codex")` returned `true` because `CODEX_MODEL_RE = /^[A-Za-z0-9][A-Za-z0-9._:-]{0,99}$/` (verified `src/types.ts:226`) matches the bare alias `"opus"`. So Codex's future-alias escape hatch silently swallowed a Claude alias — exactly the Opus-on-Codex error in the task. The new `modelCompatibleWithProvider` adds `&& !CLAUDE_MODEL_VALUES.has(model)` (`src/default-model.ts:74-76`), so `"opus"` and `"opus[1m]"` are now rejected for Codex (verified against the Claude alias list `src/types.ts:205`).

### Verification

- `resolveCreateLaunch` clamps `input.model` via `modelForProviderOrDefault` and threads the clamped `spawnInput.model` into `buildCodexSpawnArgv`, `prepareSpawnOrThrow`, and `store.create` (verified `src/service.ts:1860-1862, 1882-1884, 1935, 1955`). `buildCodexSpawnArgv` guards `if (model)` (`src/service.ts:1671`), so a clamped `null` emits no `--model`. The new test asserts exactly this argv (`test/service.test.ts:194-213`).
- `CreateSessionInput.model` is `string | null` (non-optional, `src/types.ts:157`), so `modelForProviderOrDefault(input.model, …)` is type-safe — no `undefined` leak.
- `validateModel`: the undefined-provider branch still accepts only curated `MODELS`/`CODEX_MODELS` (matches prior behavior; verified `src/validate.ts:101-107`); the `claude` branch error stays `"unknown model"`; the `codex` branch now rejects Claude aliases as `"invalid codex model"`, matching the new test's `/codex model/i` (`test/validate.test.ts:115-129`). This is consistent with the spawn-side clamp — defense in depth at the HTTP boundary, clamp on the internal `service.create` path (drain/held resume) that bypasses validation.
- `resolveRoleEnvironment` now uses `modelCompatibleWithProvider` (`src/default-model.ts:187`); for `claude` this is identical to the old `MODELS_BY_PROVIDER.claude.includes`, for `codex` it is a deliberate, safe loosening to accept `CODEX_MODEL_RE` future aliases (excluding Claude aliases). Consistent with intent.
- `resolveSpawnPreflight` is a faithful extraction of the prior `profile`/`backend`/`egressBackend`/`hold` computation (verified `src/service.ts:1470-1494`).
- Import hygiene: `AgentProvider` (service.ts:13), `MODELS_BY_PROVIDER`/`AGENT_PROVIDERS` (default-model.ts:144,149), `MODELS`/`CODEX_MODELS` (validate.ts:103-104) all still used; removed imports are genuinely unused. No dangling references.
- No user-facing strings / message catalogs touched, so no i18n parity concern (only a `console.warn` added in `src/service.ts`).
- `validateBroadcast`/`validateRetry` were byte-identical before; extracting `validateTextToIds` preserves behavior.

### Nit (non-blocking)

The refactor into `resolveCreateLaunch` dropped the multi-line comment that previously documented the deliberate Codex-vs-Claude `effectiveAutopilot` divergence ("Do NOT align to Claude's repo-default-only behavior"). The behavior is unchanged, but that comment guarded against a future regression — worth re-adding a one-liner.

No blocking issues found.

kai-osthoff (commented):
## Verdict

The PR correctly fixes the reported bug: creating a task with the Claude model `opus` selected, then switching the provider to Codex, no longer passes `--model opus` to a Codex spawn.

### What was verified

- **Root-cause fix** — `modelCompatibleWithProvider` (src/default-model.ts:71-83) now excludes known Claude aliases from Codex's free-form regex path: `CODEX_MODEL_VALUES.has(model) || (CODEX_MODEL_RE.test(model) && !CLAUDE_MODEL_VALUES.has(model))`. Verified against types.ts:205 (`"opus"`/`"opus[1m]"` ∈ CLAUDE_MODELS) and types.ts:226 (`CODEX_MODEL_RE = /^[A-Za-z0-9][A-Za-z0-9._:-]{0,99}$/` matches `"opus"`). The previous `modelCompatibleWith` and old `validateModel` wrongly accepted `"opus"` for codex precisely because the regex matches it.
- **Clamp at create** — `resolveCreateLaunch` (src/service.ts:1862-1869) calls `modelForProviderOrDefault(input.model, agentProvider)`, which returns null for `opus`+codex, threads the clamped value through `spawnInput`, and `buildCodexSpawnArgv` (src/service.ts:1671) only emits `--model` when truthy. Confirmed by test/service.test.ts:194 (argv has no `--model`, `s.model`/store null).
- **Validation parity** — branch-by-branch comparison of the new `validateModel` (src/validate.ts:95-108) against the prior implementation: claude and provider-undefined paths are byte-equivalent in behavior; the only change is the now-correct rejection of `opus`+codex (test/validate.test.ts:115). The broadcast/retry `validateTextToIds` extraction (src/validate.ts:919-940) is a behavior-preserving DRY refactor.
- **resolveRoleEnvironment** (src/default-model.ts:187) — claude path identical to old `MODELS_BY_PROVIDER` check; codex path is intentionally slightly more permissive (now also accepts safe future aliases), consistent with the rest of the codex handling.
- **No import cycle** — validate.ts now imports from default-model.ts, which imports only from types.ts (verified default-model.ts does not import validate.ts).
- **No dead code** — removed `modelCompatibleWith` has no remaining references (grep clean); the `MODELS`/`CODEX_MODELS`/`CODEX_MODEL_RE` imports dropped from service.ts are unused there; `MODELS_BY_PROVIDER` is still used at default-model.ts:149.
- **Prior-revision point #1** (deliberate-divergence comment) — re-added at src/service.ts:1886-1888. Addressed.

### Two-layer behavior note (not a defect)
API requests that explicitly send `agentProvider: "codex"` + `model: "opus"` are now hard-rejected (400, `"invalid codex model"`, test/validate.test.ts) at validate.ts, whereas the service-layer `modelForProviderOrDefault` clamp (with a `console.warn`) is the graceful net for internal callers (relaunch model-carry via `reconcileRelaunchModel`, and the default-provider path where the request omits `agentProvider`). The two paths trigger on disjoint inputs, so there is no contradiction — both eliminate the `opus`-to-codex spawn.

### Minor nit
src/service.ts:1961 — `planGateEnabled` still reads `input.planGateEnabled` while the surrounding persisted fields (model/auto/issueNumber/autopilotEnabled/research/mergeTrainPrs) were switched to `spawnInput.*` in this refactor. The value is identical (model is the only field `spawnInput` overrides), so it is purely cosmetic, but using `spawnInput` here too would keep the path consistent and guard against future `spawnInput` divergence.

kai-osthoff (commented):
## Summary

The PR fixes the reported bug: a task created with a Claude model (e.g. `opus`) that later spawns under / switches to Codex was passing `opus` through to Codex, which has no such model. The fix introduces `modelCompatibleWithProvider` / `modelForProviderOrDefault` and routes both the create path and the relaunch path through it. I verified the change end-to-end and found no blocking issues.

## What I verified

**Root cause is genuinely fixed.** `CODEX_MODEL_RE = /^[A-Za-z0-9][A-Za-z0-9._:-]{0,99}$/` (src/types.ts:226) matches the bare alias `"opus"` (letters only), and `CLAUDE_MODELS` includes `"opus"` (src/types.ts:205). The old `modelCompatibleWith` (removed at src/service.ts) accepted any regex match for codex — so `"opus"` slipped through. The new `modelCompatibleWithProvider` (src/default-model.ts:71-83) gates the free-form path with `CODEX_MODEL_RE.test(model) && !CLAUDE_MODEL_VALUES.has(model)`, so codex+`"opus"`/`"opus[1m]"`/`"sonnet"` now resolve to false. Both spawn paths benefit: `reconcileRelaunchModel` (src/service.ts:951) and the new create clamp (src/service.ts:1905-1906).

**Create path clamps before building argv.** `resolveCreateLaunch` computes `model = modelForProviderOrDefault(input.model, agentProvider)` and threads `spawnInput.model` (null after a drop) into `buildCodexSpawnArgv` (src/service.ts:1926). `buildCodexSpawnArgv` only emits `--model` when truthy (src/service.ts:1709), so a dropped model yields no flag — matching the new test at test/service.test.ts which asserts `argv` does not contain `--model` and `s.model` is null.

**validate.ts refactor is behavior-preserving except the intended fix.** I walked the cases: claude+claude-alias (accept), claude+codex/unknown (err `unknown model`), codex+curated/regex-safe alias (accept), codex+`opus` (now err `invalid codex model`, previously wrongly accepted via the regex), and `agentProvider` absent (accept any curated claude/codex alias, reject free-form) — all consistent with src/default-model.ts:71-83 and the prior logic. `CODEX_MODEL_RE` import is correctly dropped from validate.ts since it now flows through the shared helper.

**`resolveSpawnPreflight` and `validateTextToIds` are pure extractions.** `resolveSpawnPreflight` (src/service.ts:1507-1531) reproduces the inlined profile/backend/egress/hold computation verbatim, and the call site (src/service.ts:1417-1421) consumes the same four values. `validateTextToIds` is byte-identical to the former `validateBroadcast`/`validateRetry` bodies (both `STEER_TEXT_MAX`-bounded), and both public validators now delegate to it.

**No orphaned references.** Confirmed `MODELS` / `CODEX_MODELS` / `CODEX_MODEL_RE` are no longer referenced in src/service.ts, and the old `modelCompatibleWith` name appears nowhere in the tree.

**Edge input.** `CreateSessionInput.model` is `string | null` (never undefined, src/types.ts:157), so `model === input.model ? input : {...input, model}` preserves object identity (and the `images`/`issueRef` references) when nothing is dropped, and `modelForProviderOrDefault(null, …)` short-circuits to `null` via the `model === null` guard.

No locale/i18n surfaces are touched (the only new string is a server-side `console.warn`).

Nothing to address — clean, well-scoped change that resolves the reported Opus-into-Codex defect.

kai-osthoff:
@scoop 🔥

PR #1252, opened 2026-06-29
fix(ui): route short viewports to the mobile layout (unfolded fold split-landscape)

scoop:
## Problem

An unfolded foldable in **split-screen landscape** (and ordinary phone landscape) hands Shepherd a viewport that is **wide (>768px) but very short**. Layout was chosen purely by width (`max-width: 768px`), so this fell into the **desktop** branch — whose fixed `height:100dvh` chrome assumes a tall viewport. The chrome crushed everything below it and content was clipped/unreachable.

| | |
|---|---|
| Before | wide + short → desktop branch → chrome eats the height, task list & New-Task bar clipped, no scroll |
| After | wide + short → mobile layout (document-scroll list; hardened fixed-height terminal) |

## Change

Widen the single layout `MediaQuery` to fire on **`max-height: 600px`** as well as `max-width: 768px`, and harden the parts the mobile layout does **not** already make short-safe.

- **`+page.svelte:811`** — `new MediaQuery("(max-width: 768px), (max-height: 600px)")`. The shell is prop/class-driven, so this single source-of-truth flip routes the whole app to its mobile rendering. Both terms are parenthesised so Svelte's `MediaQuery` passes the comma to `matchMedia` **verbatim** as a top-level OR (not re-wrapped).
- **`Viewport.svelte`** — the detail/terminal screen is a fixed-height `.shell.mobile` column (NOT the document-scroll list shell). The control bars (`SteerBar`/`ViewportTermControls`/`SessionRecap`) are siblings *inside* `.viewport` (`overflow:hidden`), so a gated `@media (max-height:600px){ .viewport{ overflow: hidden auto } }` lets them stay reachable by scroll at the extreme-short floor instead of being clipped. `.viewport` keeps its definite `height:100%`, so while content fits the terminal's internal scroll is the only scroll (no page scrollbar); page-scroll engages only below the ~240px floor. `overflow-x` stays hidden to preserve the horizontal swipe.
- **`app.css:351`** — the iOS focus-zoom 16px input floor is a *systemic* guard (not a cosmetic); its query is widened to follow the short-wide branch so focusing an input there doesn't zoom the page.
- **`+page.svelte` styles** — trim `.shell.mobile` padding/gap on short viewports to reclaim vertical room (the `.list` rule keeps its own `padding-top:0`/ActionBar reserve by higher specificity).
- **`SteerBar.svelte:421`** — widen the one cosmetic detail-path media query so the steer-bar collapse stays consistent on short-wide.

Desktop (height>600 **and** width>768), tablet landscape, and normal phone layouts are unchanged.

## Verification

- ✅ `bun run check` — 0 errors / 0 warnings (3638 files)
- ✅ `bun run test` — 2297 passed / 167 files
- ✅ **Trigger (deterministic):** replicated Svelte 5.56.4's `MediaQuery` constructor logic — `(max-width: 768px), (max-height: 600px)` passes its parenthesis check and is sent to `matchMedia` verbatim (top-level OR, no double-wrap); the existing `"max-width: 768px"` still wraps correctly.
- ✅ **Empirical before/after on the running app** (Playwright, real `app.css`): the login password input computes **16px at 1024×420** (height≤600 → new guard fires; pre-change it would have been 13px since 1024>768), **13px at 1024×800** (desktop unchanged), **16px at 390×844** (existing width arm). Confirms the `max-height:600` branch activates at the intended viewport and nowhere else.

### Remaining sign-off (per plan's Phase-0 gate)
The instance is password-protected, so the authenticated shell couldn't be driven headlessly. The **visual list + detail/terminal check at ~1024×420 and the on-device confirmation that the real fold reports `innerHeight ≤ 600`** (read via the in-app feedback link's `Viewport:` line) is the designated shipping gate. The DPR analysis estimates the split pane at ~257–385px CSS height (comfortably ≤600). If the real device ever reports >600, bump the single `600` constant. The optional `--app-vh` (from `window.innerHeight`, keyboard-stable) fallback for `100dvh` over-measure was **not** wired — no evidence it's needed, and the `overflow:auto` net covers the extreme case.

[no-feature-entry] responsive fix, surfaces no new discoverable affordance.

### DIFF
--- a/ui/src/app.css
+++ b/ui/src/app.css
@@ -340,15 +340,17 @@ button {
 
 /* iOS zoom-on-focus guard: mobile Safari zooms the whole page when a focused
    text control renders below 16px. Force every text-entry control to the 16px
-   rung (--fs-lg) on phones — systemic here so no individual component can
-   drift back under the threshold. The `:is()` wrapper deliberately lends the
+   rung (--fs-lg) on phones AND short-wide viewports (foldable split-screen /
+   phone landscape — the same height ≤600 branch that now routes to the mobile
+   layout) — systemic here so no individual component can drift back under the
+   threshold. The `:is()` wrapper deliberately lends the
    bare `textarea`/`select` the input arm's specificity (0,3,1), so this beats
    Svelte-scoped component rules like `.ep-search.svelte-x` (0,2,0) or
    `textarea.svelte-x` (0,1,1) that set a smaller desktop size.
    max(16px, …) keeps the guard hard: --fs-lg now scales with the iOS Text Size
    setting (--ui-scale), so a *smaller* Dynamic Type choice could otherwise push
    --fs-lg under 16px and bring the focus-zoom back; the floor stays absolute. */
-@media (max-width: 768px) {
+@media (max-width: 768px), (max-height: 600px) {
   :is(input:not([type="checkbox"]):not([type="radio"]):not([type="range"]), textarea, select) {
     font-size: max(16px, var(--fs-lg));
   }
--- a/ui/src/lib/components/SteerBar.svelte
+++ b/ui/src/lib/components/SteerBar.svelte
@@ -417,8 +417,11 @@
      regardless of `compact`, so there's always a label to reveal — the `.lbl-toggle`
      reveal below is viewport-only (1 class, so it ties the base `display: none` and
      wins by coming later in source order), independent of `.show-labels` so the
-     toggle stays visible to collapse back once ABC is pressed. */
-  @media (max-width: 768px) {
+     toggle stays visible to collapse back once ABC is pressed.
+     The (max-height: 600px) arm keeps this collapse consistent on short-wide
+     viewports (foldable split-screen / phone landscape) that now route to the
+     mobile layout but are wider than the 768px breakpoint. */
+  @media (max-width: 768px), (max-height: 600px) {
     .steer-bar:not(.show-labels) .chip.bc {
       min-width: 44px;
       padding: 0;
--- a/ui/src/lib/components/Viewport.svelte
+++ b/ui/src/lib/components/Viewport.svelte
@@ -2275,6 +2275,24 @@
     transition: none;
   }
 
+  /* Short viewports (unfolded foldable split-screen landscape, phone landscape):
+     the detail/terminal screen is a fixed-height .shell.mobile column — NOT the
+     document-scroll list shell. .viewport keeps its definite height:100%, so vp-body
+     fills the pane and the terminal keeps its own internal scroll while content fits
+     (no page scrollbar). Only at the extreme-short floor (column min-content > pane,
+     vp-body pinned to its 4rem min) would the bottom bars (SteerBar /
+     ViewportTermControls / SessionRecap) be clipped by overflow:hidden with no way to
+     reach them — so allow vertical page-scroll there. overflow-x stays hidden to
+     preserve the horizontal swipe-to-navigate clip. Safe to key on height alone: at
+     height ≤600 the app is in the mobile branch (the desktop Viewport instance is
+     unmounted), and tall mobile (height >600) never matches, so normal phone detail
+     is untouched. */
+  @media (max-height: 600px) {
+    .viewport {
+      overflow: hidden auto;
+    }
+  }
+
   /* phone session view: full-bleed terminal — the side borders + rounded corners
      cost two vertical lines plus gutters on a narrow phone, so drop them and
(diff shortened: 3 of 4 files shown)


scoop (commented):
## Summary

The PR makes the unfolded foldable in split-screen landscape (and phone landscape / short desktop windows) usable by routing short-wide viewports to the existing mobile layout, and by patching the three places where that layout assumed a *tall* viewport. The approach is minimal and the inline comments accurately describe the mechanism. I found no blocking defects.

## What I verified

**JS mobile switch — `ui/src/routes/+page.svelte:817`.** Changed from `"max-width: 768px"` to `"(max-width: 768px), (max-height: 600px)"`. `MediaQuery` is imported from `svelte/reactivity` (ui/src/routes/+page.svelte:3). The comment claims the parenthesised terms are passed to `matchMedia` verbatim as a top-level OR. This relies on Svelte's `MediaQuery` only wrapping the string in extra parens when it does not already contain them (the `parenthesis_regex` test in `svelte/reactivity/media-query.js`); since the new string already starts with `(` and contains parens, it is forwarded unwrapped, so `matchMedia` receives a valid comma-separated media-query list (OR semantics) rather than the invalid `((…), (…))`. I could NOT open the installed source to confirm — `node_modules` is absent in this worktree and `svelte@^5.56.4` is only declared in ui/package.json:32 — but this matches the documented Svelte 5 behaviour. Stated as a limitation, not asserted as fact.

**Two Viewport instances / unmount claim.** Verified there are two distinct `<Viewport>` render sites: the desktop one under the `{:else}` (non-mobile) branch (ui/src/routes/+page.svelte:2330, `touch={touch.current}`) and the mobile one under `{#if mobile.current}` … detail branch (ui/src/routes/+page.svelte:2172, `mobile={mobile.current}`). The top-level `{#if mobile.current}` at :2069 means that at `height ≤ 600` only the mobile instance is mounted. This confirms the Viewport comment's assertion (ui/src/lib/components/Viewport.svelte:2286-2289) that keying the overflow rule on height alone is safe.

**`.viewport` overflow — ui/src/lib/components/Viewport.svelte:2262-2293.** Base rule is `overflow: hidden` (:2269); the new `@media (max-height: 600px)` rule sets `overflow: hidden auto` (x hidden, y auto). Neither axis is `visible`/`clip`, so there is no forced-`auto` override — the combination is valid and preserves the horizontal swipe clip while allowing vertical page-scroll only when the column overflows the fixed pane. Consistent with base `.viewport { height: 100% }`.

**`.shell.mobile` padding — ui/src/routes/+page.svelte:2719-2739.** New `@media (max-height: 600px)` rule sets `gap`, `padding-top`, `padding-bottom` on `.shell.mobile` (specificity 0,2,0). The comment claims `.shell.mobile.list` keeps its own `padding-top:0` / ActionBar `padding-bottom`. Confirmed against ui/src/routes/+page.svelte:2751-2763: `.shell.mobile.list` (specificity 0,3,0) sets both `padding-top: 0` and the ActionBar-reserving `padding-bottom`, so it outranks the new rule regardless of source order — the list screen's reserved bottom space is preserved. Base `.shell` height is `100dvh` (ui/src/routes/+page.svelte:2616), so the detail screen stays a fixed-height column as the comments assume.

**CSS OR queries — `ui/src/app.css:353` and `ui/src/lib/components/SteerBar.svelte:424`.** Both use the valid `@media (max-width: 768px), (max-height: 600px)` media-query-list form. SteerBar has exactly one `@media` block (no sibling width-gated rule left un-extended). The 768/600 thresholds match the JS switch in all four edited locations.

## Considerations (non-blocking, not findings)

- The `mobile` reactive now resolves `true` for short *wide* viewports, so any non-touch user who drags a desktop window below 600px height drops into the single-column mobile layout (list-or-detail, no side-by-side). This is explicitly intended per the comment at ui/src/routes/+page.svelte:809-815 (the desktop chrome assumes a tall viewport and "gets crushed"), so it is a deliberate trade-off rather than a regression.

- Many child components keep `@media (max-width: 768px)` blocks that were not extended with the `(max-height: 600px)` arm (e.g. ui/src/lib/components/NewTask.svelte, RetryDialog.svelte, BroadcastDialog.svelte, and AppOverlays' dialogs). On a short-wide viewport these render with `mobile={true}` but their width-gated mobile styling does not apply. I did not find a concrete break — width-gated narrow tweaks legitimately should not fire when there is ample width — but overlay/modal vertical fit on a ~500px-tall viewport is worth a manual smoke test. These files are outside the diff; informational only.

## Verdict
No blocking problems and no actionable nits in the changed files. Decision: comment.

PR #1253, opened 2026-06-29
perf(forge): unify the duplicate per-repo open-PR queries behind one cached snapshot

### LINKED ISSUE #1240: Unify the three per-repo open-PR queries behind one cached snapshot

scoop:
Closes #1240. Follow-up to #1233.

## What

Three near-duplicate per-repo open-PR GitHub queries coexisted. This unifies the **two true duplicates** — the identical `gh pr list --state open --limit 200` calls behind `GithubForge.listPullRequests` (PRs tab) and `listOpenPrStatuses` (pr-poller batch) — into **one cached per-repo snapshot**, shared by both consumers.

- **`GithubForge.listOpenPrSnapshot()`** — one `gh pr list` (union of both field sets) → `{ prs: PullRequest[]; statuses: Map<headRefName, PrStatus>; capped }`. `listPullRequests` / `listOpenPrStatuses` now delegate to it (the PR-row mapping was **extracted** to `mapGhPrToPullRequest`, not duplicated, so outputs stay byte-for-byte identical).
- **`OpenPrSnapshotService`** (`src/open-pr-snapshot.ts`) — per-repo read-through cache keyed by **`forge.slug`**, 120 s TTL, single-flight, preserve-on-error. `Semaphore` extracted to `src/semaphore.ts` (shared with `CountsService`, behavior unchanged).
- **pr-poller** sources its batch from `snapshot.refresh(forge)` (always-fresh — drives merge/handoff — under the existing `withGh` mutex; count-gate unchanged).
- **`GET /api/prs`** reads `snapshot.get(forge)` (read-through, reuses the poller's fetch) with a direct `listPullRequests()` fallback for Gitea/Local.

## Scope decision (vs. the original "all three")

Plan review showed routing the overview-counts `prKinds` through the snapshot is **net-negative**: its kind nodes already ride the *same free GraphQL request* as `totalCount`, the snapshot only warms pr-poller-batched repos (so `prKinds` would go null for most repos), and a 200-cap breakdown would contradict the exact `totalCount`. So **`listBacklogCounts` is left untouched**, and only the two genuine duplicates are unified. `doc-agent` and `standalone-critic` keep calling `listPullRequests()` directly (fresh path — a TTL-stale read there could dupe a docs PR / mis-scope a review).

## Freshness / cost notes

- PRs tab can now be ≤120 s stale (was always-fresh) — accepted: the list carries no merge/spawn decision, TTL = poller cadence so it reuses the sweep's fetch, and live session state still flows via `session:git`. The poller itself stays always-fresh (`refresh()`).
- The union field set is a strict superset of each original, so the single call is marginally more points/PR; the win is **per-cadence** (two full-rollup calls → one shared cached call), not per-call.

## Tests / verification

- Golden-fixture equivalence (PullRequest[] + statuses map identical to today for the same payload), single-underlying-fetch, fork-collision dedup, `capped`.
- Snapshot service: read-through/TTL, single-flight, `refresh` vs `get`, preserve-on-error **including the interleaved get-starts/refresh-joins race**.
- **Cross-consumer key test**: a real `PrPoller.tick()` warms the cache, then real `handlePrsList` reads it via a *separate forge instance sharing the slug* — total underlying fetch == 1, catching a slug-vs-repoPath key mismatch.
- `bun run typecheck` 0 errors · `bun run lint` clean · `bun test ./test` 5482 pass / 0 fail.

Rebased onto latest `main` (resolved overlap with #1241's pr-poller fast-sweep change). Server-only — no user-facing UX (`[no-feature-entry]`).

## Optional follow-ups (non-blocking, from review)

- `batchForRepo` fallback could prefer `listOpenPrSnapshot` over `listOpenPrStatuses` when no service is injected (theoretical edge; both always present on GitHub forge).
- `peek()` is implemented + tested but currently unused by production (mirrors `CountsService`).


scoop (commented):
## Summary

The PR delivers the issue's goal: `GithubForge.listOpenPrSnapshot()` does one `gh pr list --state open` and maps it into both consumer shapes (`prs` for the PRs tab, `statuses` keyed by `headRefName` for the poller), `listPullRequests`/`listOpenPrStatuses` become thin delegators, and `OpenPrSnapshotService` (TTL=120s, single-flight, semaphore-gated) lets the pr-poller's `refresh` warm a cache that `GET /api/prs` reads via `get`. Wiring is consistent.

Verified:
- **Field union is complete.** The merged `--json` list (src/forge/github.ts:705) is a superset of the two prior queries: old `listPullRequests` fields (number,title,url,author,createdAt,isDraft,mergeable,statusCheckRollup,reviews,headRefName,headRefOid,baseRefName,labels) ⊂ new, and old `listOpenPrStatuses` fields (…state,mergeStateStatus,reviewRequests,headRepositoryOwner) ⊂ new. So `mapGhPrToPullRequest` (github.ts:449-471) and `mapGhPr` (github.ts:632-653) both receive every field they read. The PRs-tab rows are byte-identical to the old path (same fields, same map, no dedup), and the headRefName dedup is preserved only on `statuses`.
- **`state` defaulting.** `mapGhPr` does `pr.state.toLowerCase()`; the batch loop now guards with `state: pr.state ?? "OPEN"` (github.ts:733) for fixtures/minimal payloads. `GhPr.state` is typed required (github.ts:172) and the query requests it, so real data is fine.
- **PrPoller constructor wiring** (src/index.ts:749-767) — counted all 15 positional args against the constructor (store…noneRecheckMs, snapshotSvc); they line up, including the four `undefined` defaults before `openPrSnapshot`.
- **`batchForRepo` gate** (src/pr-poller.ts:248-249) is equivalent to the old `!listOpenPrStatuses || !countOpenPrs || isFork || count<2` for the existing-method case, and adds the snapshot branch.
- **Server fallback** (src/server.ts:4097-4101): `get` never throws (catches internally → null), and the null/incapable-forge path falls back to `forge.listPullRequests()` inside the existing try/catch. `Pick<…,"get">` matches the call.
- **Semaphore move** is a verbatim relocation; src/backlog.ts now imports it (backlog.ts:8, used at 64/79) and src/open-pr-snapshot.ts imports it; `./semaphore` resolves to the new file.
- `defaultBranch()` is cached for the forge lifetime (github.ts:788-795), so the extra arm in `listOpenPrSnapshot`'s `Promise.all` is one-time per forge, not per sweep.

## Finding

**src/pr-poller.ts: a transient open-PR list failure no longer falls back to per-session polling on the snapshot path.** `batchForRepo`'s docstring (pr-poller.ts:239-240, 261) promises "transient fetch failure → per-session", and the pre-PR path achieved this because `listOpenPrStatuses()` threw and the `catch` at line 260 returned null. On the new snapshot path (pr-poller.ts:255-257), `refresh` swallows the error internally and, when a prior value exists, returns last-known-good (open-pr-snapshot.ts:90-94, `preserveOnError && prev`). So `snap?.statuses` is non-null and the poller applies **stale** statuses instead of falling back to fresh per-session `prStatus`; the `catch` never sees the error. This path is wired and reachable today (index.ts:766, GithubForge implements `listOpenPrSnapshot`). Net effect is bounded, self-healing staleness (a PR that merged during a transient error stays shown as open until the next successful sweep, ≤ TTL while warm), but it contradicts the function's own documented contract and the prior behavior. Either use a non-preserving fetch for the poller (so a list failure → per-session), or update the docstring at pr-poller.ts:238-240/261 to state that on the snapshot path a transient failure yields last-known-good rather than per-session.

## Non-blocking observations

- src/forge/types.ts: `OpenPrSnapshot.capped` is populated (github.ts:748) and unit-tested (test/forge/github.test.ts:1699) but no production consumer reads it yet; it's metadata for a future consumer — fine, just noting it carries no behavior today.
- src/open-pr-snapshot.ts: `peek()` has no production caller (only tests, test/open-pr-snapshot.test.ts:205+). Intentional API surface for symmetry, not dead in the test sense.
- src/server.ts: `GET /api/prs` now serves a snapshot up to 120s stale (or poller-warmed) rather than always-fresh, and there is no snapshot invalidation hook after a PR mutation (merge/close). This is the intended cost/sharing tradeoff from the issue, and the poller drives the PRs view anyway, but worth confirming the UI doesn't rely on an action immediately reflecting in `GET /api/prs`.

No locale/i18n surfaces are touched (backend TS only).

PR #1255, opened 2026-06-30
fix(ui): scroll plugin gear-item card to panel top, not center (#1254)

### LINKED ISSUE #1254: Plugin gear-item lands mid-panel; scrollIntoView should use block:"start"
## Problem

Clicking a plugin's gear-menu item with `action: { kind: "panel" }` opens Settings → Plugins and scrolls the matching plugin card into view, but lands the viewport in the **middle** of a tall plugin panel instead of at its top.

## Cause

`ui/src/lib/components/settings/SettingsPluginsPanel.svelte:39`

```js
el.scrollIntoView({ block: "center" });
```

`block: "center"` centers the whole plugin card. For a tall panel (e.g. claude-swap, with config + per-account meters + gauges + sparklines + time-series + bar-chart + timeline) the card's top ends up above the viewport, so the operator lands mid-panel.

## Suggested fix

```js
el.scrollIntoView({ block: "start" });
```

so the card's top aligns to the viewport top. The plugin contract (`PluginGearAction` / `PluginUIView`) exposes no scroll-alignment lever, so this can only be fixed host-side.

## Context

Surfaced from the claude-swap plugin; the plugin-side part of the request (removing a redundant raw session→account table) is handled in the plugin repo. This issue tracks the host scroll behavior only.

scoop:
## Problem

Clicking a plugin gear-menu item with `action: { kind: "panel" }` opens Settings → Plugins and scrolls the matching card into view — but for a tall plugin panel (config + meters + gauges + sparklines + charts + timeline) it landed the operator **mid-panel**.

## Cause

`ui/src/lib/components/settings/SettingsPluginsPanel.svelte` used `el.scrollIntoView({ block: "center" })`, which centers the whole card; a tall card's top scrolls above the viewport.

## Fix

`block: "center"` → `block: "start"`, so the card's top aligns to the panel viewport top. The scroll container (`.panel` in `Settings.svelte`) has no sticky header inside it, so no `scroll-margin-top` is needed. The plugin contract (`PluginGearAction` / `PluginUIView`) exposes no scroll-alignment lever, so this is fixable host-side only.

## Test

Added a browser-test case spying on `Element.prototype.scrollIntoView` that asserts `focusId` triggers it with `{ block: "start" }`. All 766 UI browser tests pass; `bun run check` clean.

Closes #1254

### DIFF
--- a/ui/src/lib/components/settings/SettingsPluginsPanel.browser.test.ts
+++ b/ui/src/lib/components/settings/SettingsPluginsPanel.browser.test.ts
@@ -104,6 +104,28 @@ describe("SettingsPluginsPanel", () => {
     expect(document.getElementById("plugin-card-beta")).not.toBeNull();
   });
 
+  it("focusId scrolls the card to the top of the panel (block:start, not center)", async () => {
+    // #1254: a tall plugin card centered (block:center) lands the operator mid-panel;
+    // it must align its top to the panel viewport instead.
+    const calls: ScrollIntoViewOptions[] = [];
+    const orig = Element.prototype.scrollIntoView;
+    Element.prototype.scrollIntoView = function (arg?: boolean | ScrollIntoViewOptions) {
+      calls.push((arg ?? {}) as ScrollIntoViewOptions);
+    };
+    try {
+      render(SettingsPluginsPanel, {
+        plugins: [plugin({ id: "tall-plugin", name: "Tall Plugin" })],
+        focusId: "tall-plugin",
+      });
+      // The $effect runs after mount; wait a microtask for it to fire.
+      await new Promise((r) => setTimeout(r, 0));
+      expect(calls.length, "scrollIntoView was called").toBeGreaterThan(0);
+      expect(calls[0].block).toBe("start");
+    } finally {
+      Element.prototype.scrollIntoView = orig;
+    }
+  });
+
   it("focusId applies focus-flash class to the matching card", async () => {
     render(SettingsPluginsPanel, {
       plugins: [plugin({ id: "target-plugin", name: "Target Plugin" })],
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

This PR implements issue #1254 precisely: it changes the plugin-card scroll alignment from `block: "center"` to `block: "start"` so a tall plugin card lands at the panel viewport's top instead of mid-panel.

## Verification

- The single behavioral change is at `ui/src/lib/components/settings/SettingsPluginsPanel.svelte:39`, inside the `$effect` that fires on `focusId` change (lines 35-43). The guard structure (`if (!focusId)`, `if (!el)`) and the focus-flash/cleanup logic are untouched — only the scroll option changed. This matches the issue's suggested fix verbatim.
- `block: "start"` is a valid, widely-supported `ScrollIntoViewOptions` value and is already used elsewhere in the tree for the same intent (verified against `ui/src/lib/components/LearningsDrawer.svelte:179` and `ui/src/lib/components/Settings.svelte:136`), so the change is consistent with existing conventions.
- The new test in `SettingsPluginsPanel.browser.test.ts:107-127` stubs `Element.prototype.scrollIntoView`, restores it in a `finally` block (no global leakage), renders with `focusId`, and asserts `calls[0].block === "start"`. The `plugin()` helper and `render` usage match the existing tests in the file (verified against lines 8-20, 22-40).
- The other `block: "center"` call sites surfaced by grep (`IssuesPanel.svelte:193`, `IntegratedEpicRow.svelte:33`, `SteersEditor.svelte:69`) are unrelated components serving different scroll intents and are correctly out of scope for this issue.

No correctness, security, or quality problems found. The change satisfies the task.

Issue #1061, opened 2026-06-24
Manual operator steps — P3: durable post-merge materialization


scoop:
Part of epic #1056. **Blocked by the P2 sub-issue** (needs ack/defer semantics to know what is deferred vs done).

## Goal

Make deferred / `POST-MERGE` manual steps outlive the session durably after merge — the core "stop them getting lost in the shuffle" fix. After a PR merges and the session archives, the operator can still see and tick off the steps they owe.

## Re-confirm at this session's plan gate

- **Durability surfaces:** in-app panel + recap always; GitHub tracking issue opt-in (default — see below) vs in-app-only vs always-create-issue.

## Scope

1. **Durable persistence:** on merge/archive, persist outstanding (deferred + `POST-MERGE`) steps so they survive teardown. **Capture the artifact before worktree removal**, and order archive/merge so teardown can't race the read or strand the session on failure (house rule).
2. **In-app panel (always):** a persistent "post-merge steps" surface (extend the Done lens or a dedicated panel) listing outstanding steps per merged session, with a way to tick them done. Design-system tokens only; if it's a blocking (modal) surface, it must dim + blur behind it per CLAUDE.md (use the canonical `.scrim`/`overlay`).
3. **Recap (always):** ensure the recap checklist reflects post-merge outstanding steps (builds on the P1 recap block).
4. **GitHub tracking issue (opt-in only):** when a per-repo opt-in setting is enabled, on merge open a GitHub issue listing the deferred steps and link it from the PR/recap. **Gate the outbound write behind explicit per-repo opt-in** (house rule — never auto-create otherwise).

## Gate obligations (CLAUDE.md)

Feature-announcement entry + i18n EN+DE parity + glossary entries as applicable — in the same PR.

## Precedent map

- Recap generated at the archive chokepoint (durable recap + Done lens, PR #665); `checklist` block in `src/visual-blocks.ts`; `DoneRecapPanel.svelte`.
- Opt-in outbound-write pattern: see existing per-repo opt-in settings (doc agent `SHEPHERD_DOC_AGENT`, `criticAllPrs`, lightweight repo mode).

## Done when

After a PR with deferred/`POST-MERGE` steps merges and the session archives, the steps remain visible in-app and in the recap; the operator can mark them done; with the repo opt-in on, a tracking issue is created and linked; with it off, no outbound write occurs; teardown ordering can't strand or race the capture; CLAUDE.md gates pass.

## Future follow-up (noted, not in scope)

[...]

PR #1259, opened 2026-06-30
feat(owed): count badge on OWED lens + agent notice to declare manual steps

### LINKED ISSUE #1257: Owed lens: add an outstanding-count badge + confirm agents actually declare manual steps
## Summary

The **Owed** Herd lens (durable post-merge manual steps, #1061) is functionally sound — render, tick-persist, and clear-on-all-done all work — but in practice it sits permanently empty, and the lens gives **no attention signal** when steps *are* owed. Two related improvements fall out of a functional verification.

## What I verified (PASS)

Ran an isolated core against a throwaway DB and drove the real UI (Playwright):

1. Empty DB → lens renders the empty state ("No post-merge steps owed — you're all caught up"); `GET /api/manual-steps/outstanding` → `[]`.
2. Seeded one outstanding record → card renders correctly (desig + PR#, title, "merged Nh ago", `N of M done`, `POST-MERGE` badges, Dismiss, note).
3. Ticked a step in the UI → count advanced and the `doneAt` stamp **persisted server-side**, not just visually.
4. Ticked every step → record auto-cleared (`clearedAt` stamped), outstanding list → `0`, panel returned to empty state.

So the render/persist/clear pipeline is not the problem.

## Finding 1 — the lens is effectively dormant in normal use

Population is gated hard at the source: on a managed session's PR merge, `PostMergeStepsService.onMerged` materializes **only if `steps.length > 0`** (`src/post-merge-steps.ts:74-77`), and steps come exclusively from `parseManualSteps()` over the PR body (`src/manual-steps.ts`). A PR contributes an owed record **only if its body contains**:

- a fenced ` ```shepherd:manual-steps ` block with `- [ ]` lines, or
- column-0 `Manual-Step:` trailer lines.

No carrier → zero steps → nothing materialized → the lens stays empty forever. Because almost no PR bodies carry these markers today, the lens never populates in real use.

**Question to resolve:** do the PR-authoring agent instructions actually tell agents to declare manual operator steps via these carriers? If not, the entire #1061 pipeline never fires for real and the lens is empty by omission, not by design. Either wire the instruction in, or this feature has no live data source.

## Finding 2 — no count/attention badge on the OWED lens (primary improvement)

The `☑ OWED` chip in `HerdLensStrip.svelte` receives no owed-count prop and renders identically whether 0 or 5 steps are owed. There is nothing to draw the operator to the lens when work is actually owed.

**Recommendation:** keep the lens **always visible** (it's one of six permanent nav lenses in a fixed `auto-fit` grid; hiding-when-empty causes layout shift and breaks discoverability — including the first-view coachmark anchored to `owed-lens`), but **add an outstanding-count badge** on the chip when `count > 0`. There's an established precedent to mirror: `UnitRow.svelte` already renders a `session.manualSteps.length` count chip on live session rows. This makes the lens earn attention only when there's something to do, without making it disappear.

Scope sketch (not prescriptive):
- thread an `owedCount` (length of the outstanding records / steps) into `HerdLensStrip`,
- render a small count badge on the OWED segment when `> 0`, reusing the existing chip/badge recipe + tokens,
- i18n + (if a new term is surfaced) the usual catalog parity.

## Minor / env notes

- The SPA login form hung on "Signing in…" in headless Chromium even though `POST /api/login` returns 200 + cookie instantly via curl. May be a headless-only quirk; flagging in case there's a real hang path on login submit.

_Filed from a functional-verification session; no product code was changed._

scoop:
Closes #1257.

The **Owed** Herd lens (#1061) was functionally sound but sat permanently empty and gave no attention signal. This addresses both findings from the verification report.

## Finding 2 — outstanding-count badge (primary)

The `☑ OWED` lens now wears a count badge when merged PRs still owe post-merge manual steps; nothing renders at 0, and the lens stays one of the six always-visible nav lenses (the `owed-lens` coachmark anchor is untouched).

- `HerdLensStrip` gains an `owedCount` prop and renders a badge on the OWED segment when `> 0` — absolutely positioned inside a `position: relative` segment so it never perturbs the fixed `auto-fit` 6-lens grid / 300px single-row floor, capped at `99+`, reusing UnitRow's `.chip-manual-steps` recipe (`--status-warn`).
- **Count = outstanding records** (`postMergeSteps.records.length`) — merged PRs still owing steps, matching the card list in the panel. (Note: this reuses UnitRow's chip *recipe*, not its *metric* — UnitRow counts steps on one session; this counts records across sessions.)
- `Herd` derives `owedCount` from the store; `+page.svelte` eager-loads it on the **desktop** layout so the badge populates before the lens is opened (no fetch in mobile flow, which has no OWED segment), with a `!loaded`-guarded **lens-open recovery fallback** if the eager load fails, and a store-level **in-flight guard** so a viewport toggle can't duplicate the GET.

## Finding 1 — give the lens a live data source

Verified that nothing told PR-authoring agents to declare manual operator steps via the `shepherd:manual-steps` fence / `Manual-Step:` trailers — so the #1061 pipeline never fired and the lens stayed empty by omission.

- New **`MANUAL_STEPS_NOTICE`** rides every **code** spawn (suppressed for research) — Claude via `composeSystemPrompt`, **Codex** inline via `buildCodexSpawnArgv` (Codex has no `--append-system-prompt`), gated `!research`.
- It is **emphatically default-empty** ("a spurious step is worse than none; when in doubt, declare nothing") to guard against the dominant failure mode — agents fabricating noise.
- `post-merge-steps` logs each materialized record (`console.info`) for real-world false-positive observability.

### ⚠️ Bundling / verifiability note
The two halves differ in testability, and bundling is intentional (it's #1257's combined scope, and the halves are complementary — the badge needs data, the data needs a signal):
- **Badge** — cleanly, deterministically tested in-PR.
- **Instruction** — an **unconditional, all-repos, behavior-changing** system-prompt injection whose real effect (do agents declare real steps vs. fabricate noise?) is **not verifiable in-PR**; there is **no automated over-declaration guard**. Safeguards are the default-empty wording, a **parser-drift unit test** (`parseManualSteps(MANUAL_STEPS_NOTICE)` pins the documented syntax to the parser), and the `console.info` for post-merge auditing.

### Spawn-path coverage
Both first-class PR-authoring coding spawns carry the notice (Claude + Codex). **doc-agent** and aux agents (critic/distiller/namer/recap/promote) are a documented exclusion — they author only doc-sync / non-feature PRs that carry no manual operator steps.

## Tests & checks
- Root: `service.test.ts` (composeSystemPrompt + Codex inline, present on code / absent on research), `manual-steps.test.ts` (parser-drift guard), `post-merge-steps.test.ts` — all green; `bun run lint` clean.
- UI: `HerdLensStrip.browser.test.ts` (badge present/absent, `99+` cap, single row preserved with 2-digit badge at 300px), `bun run check` (svelte-check 0 errors), `check:i18n` parity (EN+DE), full node suite (1540) green.
- i18n keys (EN+DE) + feature-catalog entry added for the badge.

### DIFF
--- a/src/post-merge-steps.ts
+++ b/src/post-merge-steps.ts
@@ -92,7 +92,17 @@ export class PostMergeStepsService {
       console.warn(`[post-merge-steps] materialize for ${session.id} failed:`, err);
       return;
     }
-    if (inserted) this.deps.emitChange();
+    if (inserted) {
+      // Observability (#1257): a fresh owed record is the real-world signal that an agent actually
+      // declared manual steps in its PR body. Logging it lets the operator audit which PRs populate
+      // the Owed lens and watch the false-positive (fabricated-step) rate — the prompt notice that
+      // drives this is unverifiable in-PR, so this log is the verification path.
+      const pr = prNumber != null ? `pr#${prNumber}` : "no-pr";
+      console.info(
+        `[post-merge-steps] materialized ${steps.length} step(s) for ${session.desig} (${session.id}, ${pr})`,
+      );
+      this.deps.emitChange();
+    }
 
     await this.maybeOpenTrackingIssue(session, prNumber, steps, forge);
   }
--- a/src/service.ts
+++ b/src/service.ts
@@ -514,6 +514,40 @@ const SINGLE_PR_INVARIANT =
   "Never split the work across two PRs from this one session.";
 
 /**
+ * Agent-facing notice (#1257) that tells a PR-authoring agent to DECLARE manual operator steps in the
+ * PR body via the carriers `parseManualSteps()` (src/manual-steps.ts) understands, so the #1061
+ * post-merge pipeline + the Owed lens get a live data source. Rides every CODE spawn (suppressed for
+ * research) — Claude gets it via composeSystemPrompt, Codex inline via buildCodexSpawnArgv (Codex has
+ * no --append-system-prompt). Not user-facing chrome (an instruction to the agent), so no i18n — same
+ * precedent as SINGLE_PR_INVARIANT / AUTOPILOT_DIRECTIVE.
+ *
+ * IMPORTANT — the embedded fenced example is the SOURCE OF TRUTH for the carrier syntax and is pinned
+ * to the parser by a unit test (`parseManualSteps(MANUAL_STEPS_NOTICE)` in test/manual-steps.test.ts).
+ * The fence-open line must stay EXACTLY ```shepherd:manual-steps (FENCE_OPEN regex) and each step a
+ * column-0-or-indented `- [ ]` line (TASK_LINE) so a wording edit can never silently break parsing.
+ * The notice is emphatically default-empty: a fabricated step is worse than none.
+ */
+export const MANUAL_STEPS_NOTICE =
+  "Before you open the pull request, declare any MANUAL OPERATOR STEPS the change implies — work a " +
+  "human must do around merge/deploy that the diff itself cannot perform (flip a feature flag, set " +
+  "an env var, run a one-off backfill/migration, restart a worker, DNS cutover, seed a record). " +
+  "Shepherd parses these from the PR body and surfaces them on the Owed lens so they survive merge " +
+  "and teardown.\n" +
+  "Declare them in the PR body with EITHER carrier:\n" +
+  "- A fenced block — each `- [ ]` line is one step:\n" +
+  "```shepherd:manual-steps\n" +
+  "- [ ] Set the FEATURE_X env var in production\n" +
+  "- [ ] POST-MERGE: Run the data backfill once the PR is live\n" +
+  "```\n" +
+  "- Or column-0 `Manual-Step:` trailer lines (flush-left, outside any fence), e.g. a line reading " +
+  "exactly `Manual-Step: Rotate the signing key`.\n" +
+  "Prefix a step with `POST-MERGE:` when it must happen AFTER the PR merges.\n" +
+  "DEFAULT TO DECLARING NOTHING: most PRs need NO manual steps. Add a step ONLY for a real " +
+  "out-of-band action a human must take; if merging fully completes the change, OMIT the carrier " +
+  "entirely. NEVER invent steps to fill the block — a spurious step is worse than none. When in " +
+  "doubt, declare nothing.";
+
+/**
  * Injected as the highest-priority directive for an attended RESEARCH task (`research: true`).
  * A research session does open-ended web research with sub-agents and delivers a report-only PR
  * OR a GitHub issue — never code. It SUPPRESSES the plan-gate, autopilot, and build-queue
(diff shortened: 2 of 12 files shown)


scoop (commented):
## Review

The PR addresses both halves of #1257: (1) it adds an outstanding-count badge to the `☑ OWED` lens, and (2) it gives the #1061 pipeline a live data source by injecting a `MANUAL_STEPS_NOTICE` into every code spawn (Claude via `composeSystemPrompt`, Codex inline via `buildCodexSpawnArgv`). I verified the implementation end-to-end and found no blocking defects.

### What I verified (ground truth checked)

- **Notice parses to the documented steps.** `MANUAL_STEPS_NOTICE` embeds a fenced ```` ```shepherd:manual-steps ```` block with two `- [ ]` lines (one `POST-MERGE:`-prefixed). Traced through `parseManualSteps` (src/manual-steps.ts:108): `FENCE_OPEN` (`:33`) matches the exact fence line, `TASK_LINE` (`:36`) captures both steps, `splitPostMerge` (`:49`) strips the prefix and flags the 2nd as post-merge. The prose mention `` `Manual-Step: Rotate the signing key` `` is mid-line (not column-0), so `TRAILER_LINE` (`:38`) correctly does NOT match it — yielding exactly the 2 steps the drift-guard test asserts (test/manual-steps.test.ts:92-95). Verified against src/manual-steps.ts:33-101.
- **Codex argv signature + call site agree.** `buildCodexSpawnArgv` gains a 4th param `manualStepsNotice: boolean` (src/service.ts:1701); the sole call site passes `!input.research` as the 4th arg in correct positional order (src/service.ts:1923-1940). The only other mirror, src/prompt-recommend.ts:133, is the prompt-recommender spawn (opens no PR) and correctly stays unchanged.
- **Gate is consistent.** Claude path pushes the block under `if (!opts.research)` alongside `SINGLE_PR_INVARIANT` (src/service.ts:864-869); Codex path gates on `!input.research`. Test calls to `composeSystemPrompt` match its signature `(houseRules, autopilotActive=false, opts={})` (src/service.ts:844-854).
- **Store import names.** `Herd.svelte` imports `{ postMergeSteps }` and reads `.records.length` (verified export at ui/src/lib/post-merge-steps.svelte.ts:59); `+page.svelte` aliases it to `postMergeStepsStore` (ui/src/routes/+page.svelte:71). Both resolve.
- **No re-entrant double fetch.** The new eager-load effect (+page.svelte:1021-1024) and the lens-open fallback (`:1029-1032`) are both backstopped by the store's `loading` in-flight guard (post-merge-steps.svelte.ts:17) and `loaded` $state guard; the eager effect tracks only `mobile.current`+`loaded`, so a failed load does not loop. Verified the in-flight flag is a plain (non-reactive) private field, so it cannot re-trigger the effect.
- **i18n parity.** All three new keys (`herd_owed_count`, `feat_owed_badge_title`, `feat_owed_badge_body`) exist in BOTH catalogs at identical positions — verified en.json:1895,2190,2191 and de.json:1895,2190,2191. `herd_owed_count` takes `{count}` and is called as `m.herd_owed_count({ count: owedCount })` (HerdLensStrip.svelte:127,140).
- **Feature-announcement entry valid.** `owed-lens-count-badge` matches the `FeatureAnnouncement` type (feature-announcements.ts:10-16), id is unique, `sinceVersion: "1.39.0"` aligns with the other pending entries (package.json is 1.38.0 → next release), and `targetId: "owed-lens"` matches the existing `use:coachTarget={"owed-lens"}` anchor (HerdLensStrip.svelte:129).
- **CSS tokens exist.** `--fs-micro`, `--status-warn`, `--color-panel` all resolve (src/app.css), mirroring UnitRow's `.chip-manual-steps` recipe (UnitRow.svelte:890-898).
- **Count semantics match copy.** `owedCount = records.length` = number of outstanding merged PRs; the string reads "{count} merged PR(s) still owe post-merge steps" — consistent.

### Minor observations (non-blocking, not requiring change)

- HerdLensStrip.svelte: when `owedCount > 99` the visible badge shows `99+` while the `aria-label`/`title` announce the full count (e.g. "250 merged PR(s)…"). This is arguably the better screen-reader behavior, not a defect.
- HerdLensStrip.svelte:272 the badge comment says it "mirrors UnitRow's `.chip-manual-steps` recipe" but uses `color-mix(... 14%, var(--color-panel))` vs UnitRow's `12%, transparent`. Purely cosmetic; flagging only because the comment claims an exact mirror.

No correctness, security, or parity problems found.

Issue #956, opened 2026-06-21
Evaluate: docs push gate vs. folding doc updates into the code PR (kill the double-PR)


scoop:
## Premise

Today a code change and its doc update ship as **two PRs**. The doc agent (PR #907, `src/doc-agent.ts`) is structurally decoupled: it spawns *after* a `feat`/`config` PR merges (`isDocRelevant`, `doc-agent.ts:921`), runs in its own worktree, and opens a **fresh** `shepherd/docs-update-<hex>` PR. Proposal: a **push gate that requires a local docs update before shipping**, folding docs into the code PR and killing the second PR.

This issue evaluates that. **TL;DR: a docs *presence gate* is mechanically a drop-in, but it's a poor fit on its own — it can only enforce presence, not correctness, of something that is judgment-not-mechanical. The "double PR" problem is better attacked by changing *where the doc commit lands*, not by adding a gate.**

## What a gate can and can't do

Our existing presence gates work because the required artifact is **deterministic**:

- `check-feature-catalog.sh` — "did you touch `feature-announcements.ts`?" (binary; author knows if their feat is user-facing)
- `check-generated-docs.sh` — "did you re-run `gen:cli`?" (fully mechanical — run a command)
- `check-glossary.mjs` / `check:i18n` — structural parity (mechanical)

A docs gate would assert: *IF (trigger) THEN (some file under the 8 in-scope prose paths changed) ELSE pass*, with a loud opt-out token — exactly the `feature-catalog` shape. The problem: **prose docs are the one artifact in that list that is NOT mechanical.** The whole reason the doc agent is an *LLM* is that knowing which prose is stale and how to rewrite it needs judgment. A gate can only check that *a* doc file changed — so it's cleared equally by a real update or a one-character edit. It guarantees *doc theater*, not *current docs* (the house rules already concede every gate "asserts presence, not content quality" — for prose that gap is the whole game).

## The real lever

The gate is just enforcement. The substantive change is **who writes the docs and when**:

1. **Code agent writes docs in-session** (gate enforces it didn't skip). Strictly better *grounding* than the async agent reconstructing intent from `git log -n 50` — the code agent just made the change. But it burdens **every** triggering session with doc-writing scope + token cost, and a heuristic trigger over-fires on the majority of changes (refactors/bugfixes) that need no doc edit → friction or reflexive opt-out → rot anyway.
[...]

PR #930, opened 2026-06-21
fix(doc-agent): prettier-format staged docs before commit

scoop:
## Problem

Doc-agent auto-PRs fail CI's `Lint (prettier + eslint)` step (`bunx prettier --check .`) — e.g. run #927 on `docs/external-task-api.md`.

**Root cause:** the scoped agent edits markdown but runs no git (by design); the trusted server then commits with `git commit --no-verify` (`publishStaged`), deliberately skipping the repo's pre-commit hook — the only thing that runs prettier (`lint-staged` → `prettier --write`). So unformatted markdown reaches the PR and the lint gate fails.

## Fix

`finalize()` now runs the **server's own pinned** prettier `--write` over the staged root `docs/*.md` files **before** `git add`, so the staged content is prettier-clean. The `--no-verify` commit is unchanged.

Key decisions (verified empirically on prettier 3.8):
- **Resolve from the server, not the target repo.** The managed repo lives under `config.repoRoot` ($HOME) and may never have been `bun install`ed (deps-less worktree → prettier would fail/silently no-op). Prettier is resolved + run from `SERVER_INSTALL_ROOT = resolve(import.meta.dir, "..")` (existing pattern at `egress.ts` / `server.ts`), which has the pinned prettier + plugins. `cwd`=server root (bare plugin names resolve), explicit binary path, `--config <worktree>/.prettierrc`.
- **docs-site is never a prettier target.** Targets are constrained to `docs/`-prefixed in-scope files; `docs-site/**` (Astro/Starlight `:::`/MDX/frontmatter) is never passed to root prettier, so it can't be mangled. The full in-scope set is still staged unchanged.
- **Best-effort but loud.** On prettier failure, `finalize()` logs a `console.warn` ("PR may fail the lint gate") and still commits/pushes/opens the PR — no worse than today, doc work is never lost.

## Tests

`test/doc-agent.test.ts` adds coverage asserting: `cwd`=`SERVER_INSTALL_ROOT` (not the worktree), `configPath`=`<worktree>/.prettierrc`, files = abs root `docs/*.md`; ordering (prettier before `git add`); regression (docs-site never targeted); failure path (throwing prettier → commit/push/openPr still happen).

Verification: `bun test ./test/doc-agent.test.ts` (42 pass / 0 fail), `bun run lint`, `bunx tsc --noEmit` all clean.

`[no-feature-entry]` — server-only plumbing fix, no user-facing UX.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1260, opened 2026-06-30
fix(doc-agent): fail-closed on prettier failure — abort run + visible error outcome

scoop:
## What

Makes the PR-gated **doc-agent fail-closed** on a prettier failure. Previously, if
`prettier --write` threw during `stageInScope`, the error was **swallowed** and the
unformatted docs were committed / pushed / opened as a `shepherd/docs-update-*` PR anyway —
a red, un-mergeable PR (e.g. [run 28413195069](https://github.com/erwins-enkel/shepherd/actions/runs/28413195069), PR #1245).

Now a prettier failure **aborts the run** (no commit, no push, no PR) and is made
**visible**: a new `DocAgentOutcome` value `"error"` is recorded via `recordDocAgentRun`
and fired through `onChange` / the `doc-agent:done` WS event. The UI renders it — a
persistent, assertive (`alert`) toast plus a tone-distinct status label and error badge in
`DocAgentControl`. Cleanup (cost row, `herdr.stop`, `worktree.remove`, branch `-D`) still
runs on the abort path.

## Honest framing (please read)

This is a **forward-looking guarantee, not a fix for run 28413195069**. Investigation
**refuted** the "stale build" theory (the deployed SHA when PR #1245 was created already
contained #930 with the file in scope) and **could not reproduce** the failure (the exact
production invocation formats the file correctly; server-style and CI-style prettier output
are byte-identical; all three prettier installs are 3.8.5). So PR #1245's failure was
transient/unexplained, and **this change is a no-op relative to that specific run**. Its
value is structural: a prettier *throw* can never again ship an unformatted PR. The live
symptom (PR #1245 itself) is an operator action — not touched here.

## Deliberate decisions (carried, not oversights)

- **Check-divergence is residual risk, not guarded.** If prettier *succeeds* server-side
  but its output still fails CI's `prettier --check .` (config/ignore/version divergence),
  this change does nothing — a CI-faithful verify is unrunnable from a node_modules-less
  worktree and a same-config check is tautological. CI's `prettier --check .` remains the
  backstop. Empirically unobserved (versions aligned, in-scope docs not ignored).
- **Retarget mode does not release `prSyncedKey` on abort.** Docs for that code PR defer to
  the post-merge nightly standalone-PR path (the pre-#956 baseline — gated on `lastSha`, not
  `prSyncedKey`); no permanent loss. Operator action is optional, only to co-locate docs on
  that PR before merge.

## Changes

- `src/doc-agent.ts`: `PrettierFormatError`; `stageInScope` rethrows instead of swallowing;
  `stageAndPublish` helper + `computeDocAgentOutcome` keep `finalize` thin; the abort path
  records the `error` outcome + fires `onChange`; non-`PrettierFormatError` still propagates
  to `tick()`; cleanup always runs.
- `src/types.ts` + `ui/src/lib/types.ts`: `DocAgentOutcome += "error"`.
- `ui/src/lib/store.svelte.ts`: explicit `error` branch → persistent `alert` toast, dedupe
  key `doc-agent-error:<repo>` (distinct from the transient `doc-agent-done:` key).
- `ui/src/lib/components/backlog-view/DocAgentControl.svelte`: `error` status / label /
  badge variant + `.status-badge.error` using the semantic `--color-red` token.
- `ui/messages/{en,de}.json`: `docagent_status_error`, `docagent_toast_error` (EN↔DE parity).

## Validation

- Root: `lint` ✓, `typecheck` ✓, `prettier --check .` ✓, `bun test ./test` — **5483 pass / 0 fail**.
- UI: `check` ✓, `check:i18n` ✓ (2170 keys parity), node + browser tests pass.
- `fallow audit --base origin/main`: ✓ no issues in changed files. Pre-push gate set ✓.

### DIFF
--- a/src/doc-agent.ts
+++ b/src/doc-agent.ts
@@ -52,6 +52,20 @@ async function defaultPrettierWrite(args: {
   });
 }
 
+/** Thrown by stageInScope when `prettier --write` fails, so finalize can fail-closed:
+ *  abort the run (no commit/push/PR) and record an `error` outcome instead of swallowing
+ *  the failure and committing unformatted docs. */
+export class PrettierFormatError extends Error {
+  constructor(
+    readonly repoPath: string,
+    readonly files: string[],
+    options?: { cause?: unknown },
+  ) {
+    super(`prettier failed to format docs for ${repoPath}: ${files.join(", ")}`, options);
+    this.name = "PrettierFormatError";
+  }
+}
+
 /** Prefix for ephemeral doc-agent herdr names. Each run appends 8 hex of a fresh UUID so an
  *  orphaned husk (after a restart) can NEVER squat a stable name — a re-spawn always gets a new
  *  name, so `agent_name_taken` is impossible by construction (the distiller's fix; see
@@ -768,8 +782,9 @@ export class DocAgentService {
    *
    * Before staging it formats the `docs/`-prefixed in-scope files (abs paths) with the server's
    * pinned prettier so the PR passes CI's `prettier --check .` gate — docs-site/** is
-   * Astro/Starlight-governed and must NEVER be passed to root prettier. Prettier is best-effort: a
-   * failure warns loudly and the PR still opens (no worse than today's unformatted baseline).
+   * Astro/Starlight-governed and must NEVER be passed to root prettier. Prettier is fail-closed: a
+   * failure throws {@link PrettierFormatError}, which finalize catches to abort the run (no
+   * commit/push/PR) and record an `error` outcome. A skipped doc update beats a red un-mergeable PR.
    */
   private async stageInScope(f: InFlight): Promise<string> {
     const inScope = IN_SCOPE_PATHS.filter((p) => this.fileExists(join(f.worktreePath, p)));
@@ -785,10 +800,9 @@ export class DocAgentService {
           files: absDocs,
         });
       } catch (err) {
-        console.warn(
-          `[doc-agent] prettier format failed for ${f.repoPath} — PR may fail the lint gate:`,
-          err,
-        );
+        // Fail-closed: never commit unformatted docs. Abort the run (finalize catches this,
+        // records an `error` outcome, and cleans up); a skipped doc update beats a red PR.
+        throw new PrettierFormatError(f.repoPath, absDocs, { cause: err });
       }
     }
     await this.git(f.worktreePath, ["add", "--", ...inScope]);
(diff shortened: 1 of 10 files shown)


scoop (commented):
## Summary

This PR makes the doc-agent's pre-commit `prettier --write` step **fail-closed**: a prettier failure now throws `PrettierFormatError`, which `finalize` catches to abort the run (no commit/push/PR) and record a new `error` outcome surfaced in the UI (red badge + persistent assertive toast). This satisfies the task — prettier runs before committing, and a prettier failure can no longer result in committing unformatted docs.

The implementation is correct and the control flow is sound:
- `stageInScope` (src/doc-agent.ts:806-816) runs prettier *before* `git add` (line 818), so on failure nothing is ever staged — verified the throw precedes the add.
- `stageAndPublish` catches only `PrettierFormatError` and rethrows anything else (src/doc-agent.ts:834-843); `finalize`'s `finally` always runs cleanup (herdr.stop + worktree.remove) before computing the outcome (src/doc-agent.ts:858-873). `result!` at line 876 is safe because a non-prettier throw propagates out before that line.
- `computeDocAgentOutcome` (src/doc-agent.ts:256-264) is behaviorally equivalent to the old inline expression for the pr/observe/nochange cases, with `error` taking precedence — verified against the replaced expression.

Verification performed:
- **i18n parity**: both new keys exist in en and de — `docagent_status_error`/`docagent_toast_error` at ui/messages/en.json:2194-2195 and ui/messages/de.json:2194-2195, each appearing exactly once, both `_toast_` variants using the `{repo}` placeholder consumed at ui/src/lib/store.svelte.ts:559. Only en+de catalogs exist (ui/messages/).
- **Type parity**: `DocAgentOutcome` gains `"error"` in both src/types.ts:500 and ui/src/lib/types.ts:1200.
- **Toast options**: `duration: null` + `alert: true` are valid `InfoOpts` (ui/src/lib/toasts.svelte.ts:41,46) and `duration === null` yields `durationMs === undefined` (ui/src/lib/toasts.svelte.ts:95), matching the test assertion.
- **UI exhaustiveness**: every `outcome` consumer handles `error` or falls through safely — `badgeLabel`/`badgeVariant`/`outcomeLabel` (DocAgentControl.svelte:43,53,115) and the store toast branch (store.svelte.ts:558); `.status-badge.error` uses `--color-red`, defined in ui/src/app.css.

## Findings

**src/doc-agent.ts:256** — The class-level JSDoc at lines 240-255 ("PR-gated AI doc agent … Restart-safety …") was written to document `DocAgentService`, but inserting `computeDocAgentOutcome` between that comment and the `export class DocAgentService` declaration (line 266) re-attaches the docblock to the helper function and leaves the class undocumented. Move `computeDocAgentOutcome` above the class JSDoc (or below the class), so the docblock stays with the class it describes. Non-blocking.

- src/doc-agent.ts:825-833 — In OBSERVE mode (`!act`) `stageInScope` still runs prettier, so a prettier failure now yields the `error` outcome and a persistent assertive "doc update aborted" toast even though observe was never going to commit. This is arguably correct (prettier is genuinely broken) but is a behavior change for the observe path worth being aware of.

PR #1261, opened 2026-06-30
feat: upload arbitrary binary files to a session's scratchpad via the Files tab (#1258)

### LINKED ISSUE #1258: Upload arbitrary binary files to a session's scratchpad via the Files tab
## Goal

Let an operator upload **arbitrary binary files** into a session's **scratchpad** from the UI, by extending the existing read-only **Files tab** (`FilesPanel`) with an upload affordance. The file lands in the scratchpad so the agent can pick it up on its next turn.

This is the write-counterpart to the read-only scratchpad browser shipped in #1164.

## Decisions (settled)

| Question | Decision |
| --- | --- |
| **Destination** | Upload into **the folder currently shown** in `FilesPanel` (its `listing.path`), not a fixed root or a dedicated subdir. |
| **Validation** | Accept **any file type** (drop the image MIME whitelist) but **keep the existing 10 MB cap** (`MAX_UPLOAD_BYTES`). |
| **Agent awareness** | **Passive drop** — the file just appears in the scratchpad (the agent's own dir); no path injection/steer. Agent discovers it on its own. |
| **Surface** | Reuse the **Files tab**, not the terminal drag/paste path (which stays image-only). |

## Why this is a small, well-fitted change

The plumbing already exists and is reusable:

- **Read side** — `FilesPanel.svelte` browses the scratchpad via `GET /api/sessions/:id/scratchpad[?path=]` (`handleSessionScratchpad`, `src/server.ts:1828`) backed by `listScratchpad` / `resolveScratchpadPath` (`src/scratchpad.ts`), with realpath containment as the trust boundary.
- **Upload side** — `POST /api/uploads` (`handleUpload`, `src/uploads.ts:84`) already does multipart `file` parsing, the 10 MB cap, and `Bun.write`. It only differs in (a) MIME whitelist and (b) destination dir.
- **Tab visibility** — `hasScratchpadFiles` (`scratchpadHasFiles`, `src/tmp-sweep.ts`) already gates the Files tab and refreshes at turn-end. A successful upload just needs the panel to re-list.

### Server

1. **New endpoint** `POST /api/sessions/:id/scratchpad/upload?path=<relative-dir>`.
   - Add a `POST` branch to `handleSessionScratchpad` (`src/server.ts:1828`) — relax the `req.method !== "GET"` guard so `parts[3] === "scratchpad"` + `parts[4] === "upload"` + POST is handled; keep returning `null`/404 otherwise. Live sessions only (existing archived-session 404 stays).
2. **Resolve + contain the destination directory** with the existing `resolveScratchpadPath(s.worktreePath, s.claudeSessionId, rel)`. It must resolve to an **existing directory inside the root** (reject `..`, absolute, symlink escape — already handled). 404 if the scratchpad root doesn't exist yet.
3. **Write the file** into that directory:
   - Parse multipart `file` (mirror `handleUpload`). 400 if missing.
   - Enforce `file.size > MAX_UPLOAD_BYTES` → 413. **No MIME check.**
   - **Filename:** take `basename(file.name)` and sanitize (strip path separators / control chars; fall back to a generated name if empty). **Preserve the real name** (passive drop — the agent finds it by name), unlike the image pipeline's UUID rename.
   - **Collision policy:** if the target name exists, append a numeric suffix (`name (2).ext`) rather than overwrite. (Confirm during implementation — overwrite is the alternative.)
   - Re-validate the final joined path is still `within` the root before writing.
   - `Bun.write(dest, file)`; return `{ path }` (root-relative).
4. **Factor the shared bits** (10 MB cap, multipart parse) so `handleUpload` and the new path don't drift. Keep `extForMime`/MIME whitelist exclusive to the image upload path.

### UI

1. **`api.ts`** — add `uploadScratchpadFile(id, file, dirPath)` → `POST /api/sessions/:id/scratchpad/upload?path=<dirPath>` with `FormData { file }`; returns `{ path }`. Mirror `uploadImage` (`ui/src/lib/api.ts:284`).
2. **`FilesPanel.svelte`** — add an upload control scoped to the current `listing.path`:
   - An **"Upload" button** (hidden `<input type="file" multiple>`) and **drag-and-drop onto the list area** (`.list`), with a drop-target highlight.
   - On select/drop: upload each file to the current dir, show per-file progress/error (size/too-large), then `browse(listing.path)` to refresh.
   - Empty-state (`files_empty`) should also accept drops so the first file can be added.
   - Disable while a session has no scratchpad dir yet (tab only shows when `hasScratchpadFiles`, so root generally exists — but guard the empty case).
3. **Design system** — reuse the `.gbtn` button recipe and existing `FilesPanel` tokens; no raw hex/px. Drop-zone highlight via existing `--color-*` tokens.

### Cross-cutting (project gates)

- **i18n** — add EN+DE keys for: upload button, drop hint, uploading state, success, and errors (too-large, generic failure). Both `ui/messages/en.json` and `de.json` (`check:i18n` parity gate).
- **Feature discovery** — this is a user-facing `feat`; add one entry to `ui/src/lib/feature-announcements.ts` (id, `sinceVersion`, `titleKey`/`bodyKey`) + its EN/DE keys, in the same PR (`check-feature-catalog.sh`).
- **Glossary** — no new Shepherd terms expected; skip unless copy introduces one.

### Tests

- **Server** (`test/scratchpad-endpoint.test.ts` neighbor or new `scratchpad-upload.test.ts`): upload into root and into a subdir lands the file; `..`/absolute/symlink-escape destination → 404; oversize → 413; missing `file` field → 400; filename sanitization (path separators stripped); collision suffixing; unknown/archived session → 404.
- **UI**: `FilesPanel` calls `uploadScratchpadFile` with the current path and refreshes the listing on success; surfaces the too-large error.

## Security notes

- Same trust boundary as the read browser: **realpath containment** to the scratchpad root. The new risk is *writing*, so validate the destination dir **and** the final file path (`within(root)`) and sanitize the basename to a single path segment.
- No MIME execution risk on the server — files are written to the agent's tmp scratchpad, never served as HTML; download already forces `Content-Disposition: attachment`.

## Out of scope

- Terminal drag/paste stays image-only (handled by `uploadImage`).
- No rename/delete/mkdir in the Files tab (upload only).
- No path injection or steering of the agent (passive drop by decision).
- Archived sessions remain read-nothing (no upload).

## Open questions

- Collision: suffix vs overwrite? (Proposed: suffix.)
- Multiple files in one drop: per-file errors vs all-or-nothing? (Proposed: per-file.)
- Any need to bump the 10 MB cap for this path specifically? (Proposed: keep 10 MB.)

---

_Filed from a Shepherd research session investigating "upload arbitrary binary files onto an agent's scratchpad."_

scoop:
Closes #1258.

Lets an operator upload **arbitrary binary files** into a live session's **scratchpad** from the read-only **Files tab** (`FilesPanel`). Files land in the folder currently shown so the agent picks them up on its next turn (passive drop — no path injection or steer). Write-counterpart to the read-only browser from #1164.

## What's new
- **Server** — `POST /api/sessions/:id/scratchpad/upload?path=<dir>` (`handleSessionScratchpadUpload`): accepts any file type (no MIME whitelist), keeps the 10 MB cap (`MAX_UPLOAD_BYTES`), preserves the real filename, suffixes on collision (`report (2).pdf`), live sessions only.
- **Start-of-session upload** — the Files tab now shows for any live session with a `claudeSessionId` (not just when the scratchpad already has files), and the upload creates the scratchpad **root** on demand, so an operator can drop a file *before* the agent writes anything. The GET listing returns an empty listing for a fresh session so the tab opens to an empty, droppable list.
- **UI** — Upload button (`.gbtn`) + drag-and-drop onto the list/empty-state in `FilesPanel`, with per-file inline status (uploading / done / too-large / failed) and a refresh on success.

## Security (write path)
Same trust boundary as the read browser — **realpath containment** to the scratchpad root:
- Destination dir resolved + realpath-contained and verified to be an existing **directory** (a `?path` at a file → 404, not an ENOTDIR 500).
- Filename sanitized to a single path segment **before** any join (`basename("..")` can't escape).
- Collision loop uses **`lstat`** (not `existsSync`) so a planted symlink at the target name can't be followed out of the root; the final joined path is re-validated `within(root)` before write.
- Only the known-safe scratchpad **root** is ever `mkdir`'d — never an attacker-supplied subdir.

## Scope decisions
- Suffix on collision (never overwrite); per-file batch errors (partial success OK); 10 MB cap retained; errors shown inline (not a global toast). Terminal drag/paste stays image-only.

## Cross-cutting gates
- i18n: new `files_*` + `feat_scratchpad_upload_*` keys added to EN **and** DE (parity passes).
- Feature discovery: one `feature-announcements.ts` entry (`scratchpad-upload`, since `1.39.0`) + coachmark target.

## Tests
- Server: `test/scratchpad-upload.test.ts` — root-absent mkdir+land, subdir land, `..`/absolute/symlink-escape → 404, `?path`-at-file → 404, oversize → 413, missing field → 400, sanitization, collision suffix, leaf-symlink escape, archived → 404, GET empty-root.
- UI: `api.uploadScratchpadFile.test.ts` (contract incl. 413) and `FilesPanel.browser.test.ts` (upload call args, refresh, 413 inline, drag-drop).

### DIFF
--- a/docs/external-task-api.md
+++ b/docs/external-task-api.md
@@ -124,10 +124,19 @@ Once a task exists, an external agent can also drive it:
 - `DELETE /api/sessions/:id` — archive (end) the session.
 - `GET /api/sessions` — list active sessions; `GET /api/sessions/:id/diff`,
   `/activity`, `/usage` for inspection.
-- `GET /api/sessions/:id/scratchpad[?path=]` — read-only browse of a live
-  session's own scratchpad subtree; `GET /api/sessions/:id/scratchpad/download?path=`
+- `GET /api/sessions/:id/scratchpad[?path=]` — browse a live session's own
+  scratchpad subtree; `GET /api/sessions/:id/scratchpad/download?path=`
   streams a single file. Paths are relative to the scratchpad root and
   realpath-contained to it (`..`, absolute, and symlink escapes are rejected);
   both `404` on a missing/archived session.
+- `POST /api/sessions/:id/scratchpad/upload[?path=<relDir>]` — upload an
+  arbitrary binary file (multipart `file` field, no MIME restriction) into the
+  session's scratchpad. `?path` selects a relative subdirectory within the
+  scratchpad root (default: the root); the root is created on demand but the
+  subdir must already exist. The same realpath-containment rules apply. Returns
+  `{ "path": "<relpath>" }` (a colliding name is given a numeric suffix rather
+  than overwriting). `400` on a missing `file` field, `413` when the file
+  exceeds the upload size limit (10 MiB), and `404` on a missing/archived
+  session or an out-of-root `path`.
 
 All of these honor the same auth/origin rules as task creation.
--- a/src/scratchpad.ts
+++ b/src/scratchpad.ts
@@ -1,5 +1,5 @@
 import { promises as fsp } from "node:fs";
-import { join, relative, dirname, sep, normalize, isAbsolute } from "node:path";
+import { join, relative, dirname, basename, extname, sep, normalize, isAbsolute } from "node:path";
 import { sessionScratchpadDir } from "./tmp-sweep";
 
 /**
@@ -162,4 +162,125 @@ export function attachmentDisposition(name: string): string {
     (c) => `%${c.charCodeAt(0).toString(16).toUpperCase()}`,
   );
   return `attachment; filename="${ascii}"; filename*=UTF-8''${encoded}`;
+}
+
+/**
+ * Resolve the target upload directory for a session scratchpad upload. Creates the scratchpad
+ * root on demand (start-of-session uploads before the agent has written anything). Returns
+ * `{ rootReal, dirReal }` or `null` when: `claudeSessionId` is blank, `relDir` escapes the
+ * root (`..", absolute paths, symlinks pointing outside), the subdirectory doesn't exist, or
+ * the path resolves to a file rather than a directory.
+ *
+ * IMPORTANT: only the scratchpad root is ever created — never the requested subdir — to avoid
+ * silently creating unexpected paths.
+ */
+export async function resolveScratchpadUploadDir(
+  worktreePath: string,
+  claudeSessionId: string,
+  relDir: string,
+): Promise<{ rootReal: string; dirReal: string } | null> {
+  if (!claudeSessionId) return null;
+
+  // Ensure the root exists — async mkdir so the event loop isn't blocked.
+  const root = sessionScratchpadDir(worktreePath, claudeSessionId);
+  await fsp.mkdir(root, { recursive: true });
+
+  let rootReal: string;
+  try {
+    rootReal = await fsp.realpath(root);
+  } catch {
+    return null;
+  }
+
+  // Reject any normalized path that climbs above the root or is absolute.
+  const norm = normalize(relDir || "");
+  if (norm === ".." || norm.startsWith(".." + sep) || isAbsolute(norm)) return null;
+
+  // Resolve target dir: root itself when empty/dot, else realpath of the join.
+  let dirReal: string;
+  if (norm === "" || norm === ".") {
+    dirReal = rootReal;
+  } else {
+    try {
+      dirReal = await fsp.realpath(join(rootReal, norm));
+    } catch {
+      return null; // subdir doesn't exist
+    }
+  }
+
+  // Containment check via realpath.
+  if (!within(dirReal, rootReal)) return null;
+
+  // Must be a directory: prevents a ?path pointing at a file from reaching Bun.write and
+  // throwing ENOTDIR as an uncaught 500.
+  try {
+    if (!(await fsp.stat(dirReal)).isDirectory()) return null;
+  } catch {
+    return null;
+  }
+
+  return { rootReal, dirReal };
+}
+
+/**
+ * Sanitize a raw upload filename to a single safe path segment. Takes the basename, strips path
+ * separators and control characters, and maps blank/dot/dotdot/all-dot names to "upload".
+ * Must run BEFORE any join — `basename("..") === ".."` would otherwise escape the root.
+ */
+function sanitizeUploadName(rawName: string): string {
+  // Take basename first to eliminate any directory components.
+  let name = basename(rawName);
+  // Strip path separators (/ and \) and control characters (U+0000–U+001F, U+007F).
+  // eslint-disable-next-line no-control-regex
+  name = name.replace(/[/\\]/g, "").replace(/[\x00-\x1f\x7f]/g, "");
+  // Map blank, ".", "..", and all-dot names to a generated fallback.
+  if (!name || /^\.+$/.test(name)) return "upload";
+  return name;
+}
+
+/**
+ * Choose a non-colliding path for an uploaded file within `dirReal`.
+ *
+ * Collision detection uses `lstat` (not `stat`/`existsSync`) so that symlinks at the target
+ * name are treated as collisions and skipped — this prevents a planted/dangling symlink from
+ * being followed out of the root by `Bun.write`. The candidate is the first lstat-ENOENT name.
+ *
+ * NOTE: the lstat→Bun.write window is not atomic; two concurrent same-name uploads could pick
+ * the same suffix. Acceptable for a single-operator tool — no lock needed.
+ *
+ * Returns `{ abs, rel }` or `null` if the final abs path fails the backstop containment check.
+ */
+export async function placeScratchpadUpload(
+  rootReal: string,
+  dirReal: string,
+  rawName: string,
+): Promise<{ abs: string; rel: string } | null> {
+  const name = sanitizeUploadName(rawName);
+
+  // Split into stem and extension for collision suffixing.
+  const ext = extname(name); // e.g. ".pdf" or "" for extensionless
+  const stem = ext ? name.slice(0, -ext.length) : name;
+
+  let candidate = name;
+  let n = 1;
+  for (;;) {
+    const target = join(dirReal, candidate);
+    try {
+      await fsp.lstat(target); // throws ENOENT when absent
+      // Entry exists (any type, incl. symlinks) — try next suffix.
+      n++;
+      candidate = ext ? `${stem} (${n})${ext}` : `${stem} (${n})`;
+    } catch {
+      // ENOENT — this candidate is free.
+      break;
+    }
+  }
+
+  const abs = join(dirReal, candidate);
+
+  // Backstop containment re-validation.
+  if (!within(abs, rootReal)) return null;
+
+  const rel = relFromRoot(rootReal, abs);
+  return { abs, rel };
 }
--- a/src/server.ts
+++ b/src/server.ts
@@ -71,7 +71,13 @@ import { analyzeReadiness } from "./readiness";
 import { listCommands } from "./commands";
 import { listDirs, validateRoot, collapseHome } from "./dirs";
 import { scratchpadHasFiles } from "./tmp-sweep";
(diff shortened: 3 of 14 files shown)


scoop (commented):
## Summary

This PR implements the scratchpad upload feature from issue #1258 cleanly and matches the settled decisions: any file type, 10 MB cap retained, passive drop into the currently-shown dir, Files-tab surface, suffix-on-collision. I verified the security-relevant paths and the cross-cutting gates rather than assuming them.

## What I verified

**Server containment & routing**
- `handleSessionScratchpadUpload` is reachable: `handleSessionsClearMerged` requires `parts[2]==="clear-merged"` (`src/server.ts:1919`), `handleSessionCreate` requires `!parts[2]` (`src/server.ts:1707`), `handleSessionReads`/`handleSessionScratchpad` both `return null` on non-GET (`src/server.ts:1814`, `src/server.ts:1857`), so a `POST .../scratchpad/upload` falls through to the new handler registered at `src/server.ts:2795`.
- Path-traversal defense is layered and correct: `resolveScratchpadUploadDir` rejects `..`/absolute via `normalize` + the `within()` realpath backstop (`src/scratchpad.ts:196-212`) and requires the target be an existing directory (`src/scratchpad.ts:216-217`). `sanitizeUploadName` takes `basename` first then strips `/ \` and control chars, mapping dot-only names to `"upload"` (`src/scratchpad.ts:230-238`) — `basename("../../x")==="x"`, no escape. `placeScratchpadUpload` uses `lstat` so a planted symlink at the target name is treated as a collision and skipped rather than followed (`src/scratchpad.ts:269`), with a final `within(abs, rootReal)` backstop (`src/scratchpad.ts:282`).
- Size cap mirrors `handleUpload`: `file.size > MAX_UPLOAD_BYTES` → 413 (`src/server.ts:1875` vs `src/uploads.ts:102`); `MAX_UPLOAD_BYTES` is exported (`src/uploads.ts:6`). Missing-`file` → 400 via the factored `parseUploadFile` (`src/uploads.ts:88-92`), shared by both endpoints with no MIME check on the new path. Verified against `src/uploads.ts`.

**i18n parity** — all 10 new keys exist in BOTH catalogs: `files_upload_button`, `files_upload_aria`, `files_upload_drop_hint`, `files_uploading`, `files_upload_done`, `files_upload_too_large`, `files_upload_failed`, `feat_scratchpad_upload_title`, `feat_scratchpad_upload_body`, `files_list_aria` (verified `ui/messages/en.json:2196-2205` against `ui/messages/de.json:2196-2205`; grep confirms each key appears exactly once per file). All `{name}` placeholders present on both sides.

**Feature catalog** — entry added at `ui/src/lib/feature-announcements.ts:1868-1877` with `id`/`targetId` both `"scratchpad-upload"`, matching `use:coachTarget={"scratchpad-upload"}` on the Upload button (`FilesPanel.svelte`). `sinceVersion: "1.39.0"` is consistent with the current `1.38.0` in `package.json:3`. The `check-feature-catalog.sh` gate only requires the catalog file be touched (`scripts/check-feature-catalog.sh:94`) — satisfied.

**UI wiring** — `uploadScratchpadFile` posts FormData and throws `ApiError` on non-OK so the 413 branch in `FilesPanel.uploadFiles` works (`ui/src/lib/api.ts:560-575`). `session.claudeSessionId` is a real field (`ui/src/lib/types.ts:743`); the `hasFiles` visibility change (`Viewport.svelte:515`) is intentional and the dropped `hasScratchpadFiles` flag is still stored harmlessly (`store.svelte.ts:316`). The synthetic empty-listing for a not-yet-created root (`src/server.ts:1850`) keeps the Files tab from erroring before the first upload, and the upload endpoint creates the root on demand (`src/scratchpad.ts:185-186`) — the two agree.
- Per-file status reassignment inside `Promise.all` is race-free: each callback reads-and-writes `uploads` synchronously after its own `await`, so no lost updates.

## Non-blocking nit

- `ui/src/lib/components/viewport/FilesPanel.svelte`: upload status lines (`done`/`failed`/`too_large`) accumulate and are only cleared on session change (`uploads = []` in the session-id effect) — not on directory navigation or after a delay. After several uploads the panel keeps a growing stack of stale "X uploaded" lines. Consider auto-dismissing succeeded entries (or clearing on `browse`).

## Observations (non-blocking, accepted design)

- `parseUploadFile` calls `req.formData()`, which buffers the multipart body before the post-parse 413 size check — identical to the pre-existing `handleUpload` behavior the new endpoint deliberately mirrors, so not a regression. Could not confirm whether Bun streams/spills large multipart bodies to disk; flagging only as a shared characteristic, not a defect of this PR.

scoop (commented):
## Summary

The PR implements scratchpad upload as specified: a new `POST /api/sessions/:id/scratchpad/upload[?path=]` endpoint, shared multipart parsing factored out of `handleUpload`, realpath containment of the destination dir, basename sanitization, lstat-based collision suffixing with a `within()` backstop, and a Files-tab upload affordance (button + drag/drop) with i18n (EN+DE) and a feature-catalog entry.

I verified the security-relevant paths and the cross-cutting gates:

- **Containment**: `resolveScratchpadUploadDir` (src/scratchpad.ts:14-86) rejects `..`/absolute/symlink-escape, requires an existing directory, and creates only the root; `placeScratchpadUpload` (src/scratchpad.ts:120-160) sanitizes the basename, uses `lstat` so symlinks at the target name are treated as collisions (not followed), and re-checks `within(abs, rootReal)` before returning. `within`/`relFromRoot` are the existing module-private helpers (src/scratchpad.ts:28-36). Looks sound.
- **Routing**: `handleSessionScratchpad` returns `null` for non-GET (src/server.ts:1858), so `handleSessionScratchpadUpload` (src/server.ts:1875-1897) handles POST; both are registered in `handleSessions` (src/server.ts:2792-2795). Archived/unknown session → 404 before parsing.
- **Shared parse**: `parseUploadFile` (src/uploads.ts:88-93) is reused by `handleUpload` and the new path; MIME whitelist stays exclusive to the image path. Verified.
- **i18n parity**: all 10 new keys exist in BOTH catalogs — verified `files_upload_button`/`files_upload_aria`/`files_upload_drop_hint`/`files_uploading`/`files_upload_done`/`files_upload_too_large`/`files_upload_failed`/`feat_scratchpad_upload_title`/`feat_scratchpad_upload_body`/`files_list_aria` against ui/messages/en.json:2195-2205 and ui/messages/de.json:2195-2205. All `m.*` references in FilesPanel.svelte resolve to these.
- **Feature catalog**: entry added (ui/src/lib/feature-announcements.ts:1868-1878) with matching `titleKey`/`bodyKey` and `targetId: "scratchpad-upload"`, anchored by `use:coachTarget={"scratchpad-upload"}` on the Upload button (FilesPanel.svelte:130). `coachTarget` import path matches the convention used across the tree (verified against e.g. ui/src/lib/components/viewport/ViewportTabBar.svelte:4). Version 1.39.0 is the next release (package.json is 1.38.0).
- **Tab visibility**: `hasFiles` now gates on `session.claudeSessionId !== "" && session.status !== "archived"` (Viewport.svelte:515). `claudeSessionId` is typed `string` (ui/src/lib/types.ts:743), so `!== ""` is well-defined; consistent with the server requiring a non-blank `claudeSessionId`.

## Prior-review point

The previously raised point about upload-status lines accumulating is **NOT addressed** in this revision. `uploads` is cleared only in the session-change `$effect` (FilesPanel.svelte:49); `browse()` (FilesPanel.svelte:31-41) does not clear it, and a successful upload transitions the entry to `"done"` (FilesPanel.svelte:85) and renders a persistent `{name} uploaded` line (FilesPanel.svelte:179-181) that is never dismissed. After several uploads, stale "uploaded" lines pile up and survive directory navigation. Re-raised below. This is a UX-polish issue, not a correctness/security bug, so the decision remains "comment".

PR #1262, opened 2026-06-30
fix(deploy): self-contained log rotation — drop external logrotate dep (#1212)

scoop:
## Problem

Deploy showed:

> `! logrotate not found — skipping log-rotation timer (shepherd.log stays unbounded)`

The #1212 log-rotation timer was **gated on the external `logrotate` binary** — optional, best-effort to install. On any host lacking it (no root, package unavailable, minimal image), the timer was silently skipped and `shepherd.log` grew without bound. The very failure mode the feature existed to prevent.

## Fix

Remove the external dependency. `deploy/rotate-shepherd-log.sh` is a self-contained **copytruncate size-cap** (50 MiB cap, keep 7 compressed rotations, coreutils only) that reproduces exactly the posture logrotate gave us — so rotation now works on **every** host with systemd.

- `shepherd-logrotate.service` execs the rotator from `~/.shepherd` via `/bin/sh` (no exec bit, no binary lookup).
- Install/enable of the timer is now **unconditional** in `provision.ts` and `update.sh`.
- Dropped: the binary probe (`resolveLogrotate`), the `%h`-templated config (`templateLogrotateConfig` + `deploy/shepherd.logrotate`), and `install.sh`'s best-effort `logrotate` install.

`copytruncate` is required (not preference): `shepherd.service` holds the log fd open `O_APPEND`, so a rename+create would reclaim nothing — the script copies aside, compresses, then truncates the live file in place, which the held fd respects. (Same tiny copy→truncate race as the stock logrotate directive.)

Unit names kept (`shepherd-logrotate.*`) so existing hosts overwrite in place — no orphaned units.

## Verification

- `bun test ./test` — 5507 pass, 0 fail
- `bun run lint` clean, `tsc --noEmit` clean
- `shellcheck` clean (v0.11.0)
- Functional smoke test of the rotator: rotates at cap, truncates in place, shifts `.N.gz` rotations, no-ops on empty/missing.


scoop (commented):
## Summary

This PR removes the dependency on the external `logrotate` binary (which was best-effort installed and silently skipped when absent, leaving `shepherd.log` unbounded) and replaces it with a self-contained POSIX shell rotator, `deploy/rotate-shepherd-log.sh`, run hourly by the existing timer. The gating/probe machinery (`ensure_logrotate`, `resolveLogrotate`, `templateLogrotateConfig`, the `probe` param of `installService`) is removed and the units are now written/enabled unconditionally. Verified the change end to end — it is correct and internally consistent.

### Verification performed
- **No dangling references to removed symbols.** Grepped the tree: `resolveLogrotate`/`templateLogrotateConfig` appear nowhere outside the deleted code and the updated test (`test/provision.test.ts:1`). `probeVersion` is still used at `deploy/provision.ts:328` (via the `installPrereqs` probe), so its retention is correct. All `installService` callers (`deploy/provision.ts:346`, `test/provision.test.ts:332`, `test/provision.test.ts:381`) use the new 6-arg signature — none pass the dropped `probe` arg.
- **Log path parity.** The rotator's default `LOG=$HOME/.shepherd/shepherd.log` (`deploy/rotate-shepherd-log.sh:21`) matches the path the server appends to: `StandardOutput=append:%h/.shepherd/shepherd.log` / `StandardError` (verified against `deploy/shepherd.service:25-26`). systemd sets `HOME` for user units, so the `$HOME` default resolves.
- **Unit ↔ script path parity.** `ExecStart=/bin/sh %h/.shepherd/rotate-shepherd-log.sh` (`deploy/shepherd-logrotate.service:10`) matches the install copy targets: `~/.shepherd/rotate-shepherd-log.sh` written by provision (`deploy/provision.ts:282-285`) and copied by update (`deploy/update.sh`). Running via `/bin/sh` makes the missing exec bit on the `fileIO.write` copy a non-issue, as the comment notes.
- **Policy parity with the deleted `shepherd.logrotate`.** `size 50M` → `52428800` bytes (50·1024·1024, matching logrotate's MiB interpretation); `rotate 7` → `KEEP=7`; `compress` → `gzip`; `copytruncate` → `cp "$LOG" "$LOG.1"; : >"$LOG"; gzip -f`; `missingok` → `[ -f "$LOG" ] || exit 0`; `notifempty` → covered by the `size < MAX` gate. Rotation shift (`.6.gz→.7.gz … .1.gz→.2.gz`, then new `.1.gz`) keeps exactly 7 compressed generations — correct.
- **install.sh exit-status concern resolved.** The old code carried a careful comment about the trailing `if [ Linux ]` leaving a non-zero exit under `set -e` on non-Linux. That block is gone; `install_os_prereqs` now ends on `ensure_toolchain` (`deploy/install.sh:110`), which returns 0 on success or `die`s — so the `set -e` hazard is genuinely removed, not merely relocated.
- **POSIX/dash safety.** `$((...))`, `[ ]`, `: >`, `wc -c <file`, `mv -f`/`rm -f` are all POSIX; the script runs fine under dash (Debian `/bin/sh`). GNU/busybox `wc -c <file` emits a bare integer (no leading-space padding), so the `-ge` integer test is safe.

### Nit (non-blocking)
- `deploy/rotate-shepherd-log.sh:11` (and the matching comment in `deploy/shepherd-logrotate.service:12`) describe the tools as "coreutils only" / "plain coreutils (cp, gzip, wc, mv, rm)", but `gzip` is its own package, not part of coreutils. There is no behavioral regression — logrotate's `compress` also shelled out to `gzip`, and gzip is present on effectively every systemd host — so this is purely a wording accuracy point in the comment, not a defect.

scoop (commented):
## Re-review

The PR replaces the optional external `logrotate` dependency with a self-contained POSIX `sh` size-cap rotator (`deploy/rotate-shepherd-log.sh`), making the timer unconditional so `shepherd.log` can no longer grow unbounded on hosts that lacked the `logrotate` binary. The removed helpers (`templateLogrotateConfig`, `resolveLogrotate`, `LOGROTATE_*`) and the gated install paths in `install.sh`/`update.sh` are cleanly excised, and the tests were updated to match.

### Prior finding — addressed
The earlier point (gzip is a separate package, not coreutils — the "coreutils only" / "plain coreutils (cp, gzip, wc, mv, rm)" wording is inaccurate) was corrected by the HEAD commit in **both** of its named locations:
- `deploy/rotate-shepherd-log.sh:8-9` now reads "baseline tools (coreutils + gzip — gzip is universally present and was already required by logrotate's `compress`)".
- `deploy/shepherd-logrotate.service:12` now reads "baseline tools (coreutils cp/wc/mv/rm + gzip)".

Verified against `deploy/rotate-shepherd-log.sh:8` and `deploy/shepherd-logrotate.service:12`.

### Residual instance of the same inaccuracy (new nit)
The same correction was **not** applied to `deploy/install.sh:111`, which the HEAD commit left untouched. It still reads `# The log-rotation timer (#1212) is self-contained (deploy/rotate-shepherd-log.sh, coreutils only)`. Since the script invokes `gzip -f` (`deploy/rotate-shepherd-log.sh:46`) — and gzip is a separate package from coreutils — "coreutils only" is inaccurate here for exactly the reason flagged before. `install.sh` is in the diff, so this is in scope. Cosmetic only (no behavioral effect).

### Verification of the new rotator
- copytruncate semantics: `cp "$LOG" "$LOG.1"` → `: >"$LOG"` → `gzip -f "$LOG.1"` (`deploy/rotate-shepherd-log.sh:44-46`) correctly preserve the server's held `O_APPEND` fd, matching the former `copytruncate` directive.
- Size gate `wc -c` vs `52428800` (50 MiB) and the `.6.gz→.7.gz … .1.gz→.2.gz` shift before creating the new `.1` are correct; the oldest is aged out with `rm -f "$LOG.$KEEP.gz"` first (`deploy/rotate-shepherd-log.sh:33-41`). Linux/GNU-coreutils target means `wc -c <file` emits no leading-space padding — no integer-compare issue.
- `ExecStart=/bin/sh %h/.shepherd/rotate-shepherd-log.sh` (`deploy/shepherd-logrotate.service:9`) runs via `/bin/sh`, so the `fileIO.write`-copied script needs no exec bit — consistent with the comment.
- No dangling references: grep finds no remaining uses of `shepherd.logrotate`, `resolveLogrotate`, `templateLogrotateConfig`, `ensure_logrotate`, or `logrotate.state`. `probeVersion`/`probe` is still legitimately used by `installPrereqs` (`deploy/provision.ts:341`), so dropping it from `installService`'s signature leaves no dead param.
- Tests in `test/provision.test.ts` were updated consistently with the removed exports and new unconditional behavior.

scoop (commented):
## Re-review

The single prior-revision finding is resolved, and the broader change (replacing the optional external `logrotate` dependency with a self-contained `deploy/rotate-shepherd-log.sh` rotator) is internally consistent. No new findings.

### Prior finding — RESOLVED
1. `deploy/install.sh:111`: the comment that previously said the rotator was "coreutils only" now reads `coreutils + gzip — both baseline` (deploy/install.sh:111-112), matching the corrections already in `rotate-shepherd-log.sh:8` and `shepherd-logrotate.service`. Verified against deploy/install.sh:111-112.

### Verification performed
- Removed exports `resolveLogrotate`/`templateLogrotateConfig` have no remaining references (Grep across `.ts/.sh/.svelte`: no files found), and the test file dropped their imports (test/provision.test.ts:13 area).
- The `probe` param removed from `installService`'s signature is still used legitimately: `provision()` resolves `probe` (deploy/provision.ts:328) and passes it to `installPrereqs` (deploy/provision.ts:341) — so it is not a now-orphaned variable. All `installService(...)` call sites updated to the shorter arity (deploy/provision.ts:342; test/provision.test.ts:332,381).
- `ExecStart=/bin/sh %h/.shepherd/rotate-shepherd-log.sh` (deploy/shepherd-logrotate.service:10) targets exactly the path provision.ts (deploy/provision.ts:283) and update.sh (deploy/update.sh:81) copy the rotator to — detection ≡ execution path agree.
- `deploy/shepherd-logrotate.timer`, referenced by provision.ts:291-292 and update.sh:83, still exists (Glob confirmed).
- Rotator algorithm (deploy/rotate-shepherd-log.sh): `[ -f ]` missingok guard, `wc -c` size gate (≥ 50 MiB), oldest-out `rm -f $LOG.$KEEP.gz`, descending shift of `.i.gz`→`.(i+1).gz`, then copytruncate (`cp "$LOG" "$LOG.1"` → `: >"$LOG"` → `gzip -f "$LOG.1"`). The 7-keep / 50 MiB policy and copytruncate semantics match the deleted `deploy/shepherd.logrotate` config. The copy/truncate race is documented and matches stock logrotate `copytruncate`. Runs only on the Linux systemd timer path, so GNU `wc -c` emitting a bare integer is safe for `[ -ge ]`.
- Migration: update.sh now copies the rotator + units and enables the timer unconditionally (deploy/update.sh:81-85), overwriting the old `ExecStart=logrotate` unit and running `daemon-reload` first — existing hosts self-heal correctly.

PR #1266, opened 2026-06-30
feat(new-task): hide hidden repos in repo picker, reveal on search

scoop:
## What

Repos flagged **hidden** (`RepoConfig.hidden` — the same flag the Backlog repos panel already honors) no longer clutter the **New Task** repo picker. They're dropped from the default list, the recents shortcut, and the keyboard selection paths — but **reappear the moment a name search matches**, so a hidden repo is always one keystroke away when you actually need it.

## Why

The Backlog already declutters via `hidden`, but the New Task picker showed every repo. This brings the picker in line while keeping hidden repos reachable (you can still start a task on one by searching for it).

## How

- **Server** (`src/server.ts`): `GET /api/repos` now marks each repo with `hidden`, reconciling the persisted hidden set realpath→raw exactly as the Backlog payload does (`reconcileRealPathsToRaw`), so a hide matches its repo even under a symlinked root.
- **`RepoSelect.svelte`**: new **opt-in** `hideHidden` prop (default `false`). When on: hidden repos are excluded from the default list + pinned recents, and revealed on name search. Default-off means the four other `RepoSelect` consumers (Backlog, SteersEditor, CardMenu, …) are untouched.
- **`NewTask.svelte`**: passes `hideHidden`, and excludes hidden repos from every path that bypasses the dropdown's filtering — `defaultRepoPath` (with a full-list fallback so `repoPath` is never left empty when *all* repos are hidden), the Alt+[ / Alt+] cycle, and the Alt+1/2/3 digit shortcut. Recents are filtered **before** `recentRepos`' top-N slice in both the picker and the digit shortcut, so the pinned rows stay index-aligned with the digit keys (single source of truth preserved).
- Types: `RepoEntry` gains optional `hidden?: boolean`.
- Feature discovery: one `feature-announcements.ts` entry (`1.39.0`, anchored to the repo field) + EN/DE message keys.

## Tests

- `RepoSelect.browser.test.ts`: hidden omitted by default, revealed on search, and still shown when `hideHidden` is off.
- `NewTask.browser.test.ts`: default selection skips a hidden (most-recently-used) repo; Alt+] cycle steps over a hidden repo; Alt+1 never targets a higher-count hidden repo.
- Green: root lint + `bun test ./test` (5507), UI `check` + `check:i18n` + full vitest suite (2338).

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -3325,7 +3325,13 @@ async function handleRepos({ req, parts, deps }: Ctx): Promise<Response | null> 
       const recentCounts = deps.store.recentSessionCountsByRepo(
         Date.now() - RECENT_WINDOW_DAYS * 24 * 60 * 60 * 1000,
       );
-      const repos = listRepos(config.repoRoot).map((r) => ({
+      const baseRepos = listRepos(config.repoRoot);
+      // repo_config keys are safeRepoDir/realpath-resolved while listRepos enumerates the
+      // raw join(repoRoot, name) path; reconcile the hidden set into raw space (same as the
+      // backlog payload) so a persisted hide matches its repo even under a symlinked root.
+      // The picker (NewTask) hides these by default but reveals them on name search.
+      const hiddenSet = reconcileRealPathsToRaw(deps.store.hiddenRepoPaths(), baseRepos);
+      const repos = baseRepos.map((r) => ({
         ...r,
         lastUsedAt: lastUsed[r.path],
         recentAgentCount: recentCounts[r.path],
@@ -3333,6 +3339,7 @@ async function handleRepos({ req, parts, deps }: Ctx): Promise<Response | null> 
         // can offer "Sync fork" only on fork repos. resolveForge is cached per dir
         // and shared with the backlog poller — no extra git shell on a warm cache.
         isFork: deps.resolveForge?.(r.path)?.isFork ?? false,
+        hidden: hiddenSet.has(r.path),
       }));
       // Return the window alongside the repos so the picker labels the count with
       // the exact day count it was computed over — single source of truth.
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2202,5 +2202,7 @@
   "files_upload_failed": "{name} konnte nicht hochgeladen werden",
   "feat_scratchpad_upload_title": "Dateien ins Scratchpad hochladen",
   "feat_scratchpad_upload_body": "Du kannst jetzt direkt aus dem Dateien-Tab Dateien in das Scratchpad einer Session laden - per Klick auf Hochladen oder per Drag-and-drop. Der Agent kann alles lesen, was du dort ablegst.",
-  "files_list_aria": "Scratchpad-Dateien"
+  "files_list_aria": "Scratchpad-Dateien",
+  "feat_newtask_hide_hidden_title": "Ausgeblendete Repos bleiben aus der Neue-Aufgabe-Auswahl",
+  "feat_newtask_hide_hidden_body": "Repos, die du ausgeblendet hast, überladen die Repo-Auswahl beim Start einer neuen Aufgabe nicht mehr. Sie verschwinden aus der Liste und den Zuletzt-verwendet-Kürzeln, erscheinen aber sofort, sobald du nach ihrem Namen suchst - ein ausgeblendetes Repo ist also bei Bedarf immer nur einen Tastendruck entfernt."
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2202,5 +2202,7 @@
   "files_upload_failed": "Upload of {name} failed",
   "feat_scratchpad_upload_title": "Upload files to the scratchpad",
   "feat_scratchpad_upload_body": "You can now upload files directly into a session's scratchpad from the Files tab — click Upload or drag-and-drop files onto the list. The agent can read anything you drop there.",
-  "files_list_aria": "Scratchpad files"
+  "files_list_aria": "Scratchpad files",
+  "feat_newtask_hide_hidden_title": "Hidden repos stay out of the New Task picker",
+  "feat_newtask_hide_hidden_body": "Repos you've hidden no longer clutter the repo picker when you start a new task. They're dropped from the list and the recents shortcuts, but still appear the moment you search for one by name — so a hidden repo is always one keystroke away when you need it."
 }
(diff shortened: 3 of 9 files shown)


scoop (commented):
## Summary

The PR cleanly accomplishes the task: repos flagged `hidden` are dropped from the New Task repo picker (default list + pinned recents) and revealed only on name search, while every other selection path (default selection, Alt+[/], Alt+digit) consistently skips them. I verified the data flow end-to-end.

## What I verified

- **Server payload** (`src/server.ts:3328-3343`): `deps.store.hiddenRepoPaths()` resolves to `SessionStore.hiddenRepoPaths()` (verified `src/store.ts:1157`, which reads `repo_config WHERE hidden = 1`). `deps.store` is typed `SessionStore` (`src/server.ts:199`). The new `reconcileRealPathsToRaw(hiddenRepoPaths(), baseRepos)` + `hiddenSet.has(r.path)` pattern is identical to the established backlog payload (`src/server.ts:4514-4520`), so raw-vs-realpath reconciliation is handled the same way — the comment's symlinked-root concern is genuinely addressed.
- **Type** (`ui/src/lib/types.ts:38-40`): `hidden?: boolean` added to `RepoEntry`; absent ⇒ visible, matching all the `!r.hidden` predicates.
- **RepoSelect** (`ui/src/lib/components/RepoSelect.svelte`): `hideHidden` defaults to `false`; `selected` lookup stays over the full `repos` so a seeded/relaunched hidden repo still renders in the trigger; `filtered` drops hidden only while not searching (`searching` reveals them); `recents` filters hidden BEFORE `recentRepos`' top-N slice.
- **Index alignment**: `recentRepos` (`ui/src/lib/recentRepos.ts`) is the shared source of truth. NewTask's Alt+digit (`NewTask.svelte:599`) filters hidden before `recentRepos`, exactly as RepoSelect's `recents` (`RepoSelect.svelte:105`) does — so the digit index matches the pinned group. Verified.
- **Default selection** (`NewTask.svelte:281-282`): `defaultRepoPath(r.filter(!hidden)) || r[0]!.path` — when all repos are hidden the filtered list is empty, `defaultRepoPath` returns `""` (verified `NewTask.svelte:236-244`), and the `|| r[0]!.path` fallback keeps `repoPath` non-empty (guarded by `r.length > 0`). Correct.
- **i18n parity**: `feat_newtask_hide_hidden_title` / `feat_newtask_hide_hidden_body` added to BOTH `ui/messages/en.json:2205-2206` and `ui/messages/de.json:2205-2206`. Verified.
- **Feature catalog** (`feature-announcements.ts:1878-1888`): entry shape matches `FeatureAnnouncement` (`:10-16`); `sinceVersion: "1.39.0"` matches the current unreleased line (recent entries all `1.39.0`; root `package.json` is `1.38.0`); `targetId: "nt-repo"` resolves to the existing `use:coachTarget={"nt-repo"}` anchor (verified `NewTask.svelte:795`).
- **Blast radius**: `<RepoSelect` is instantiated only in `NewTask.svelte`, and the new prop defaults off — no other consumer behavior changes.

## Minor (non-blocking)

`cycleRepo` (`NewTask.svelte:567-574`): when the current repo is hidden, `cur === -1` so `i = 0`. A forward step (`dir = 1`) then lands on `list[1]`, skipping the first visible repo `list[0]`; a backward step lands on `list[n-1]`. This is pre-existing `cur === -1 ? 0 : cur` logic that is now newly reachable when a hidden repo is seeded as `initialRepoPath`. The result is always a valid visible repo, so it's a small UX quirk rather than a bug — entering the visible subset at `list[0]` on a forward step would feel more predictable.

- `RepoSelect.svelte` comment refers to "backlog/steers/card RepoSelect consumers," but `<RepoSelect` has only one instantiation (NewTask); those other pickers appear to be separate components. Informational only.

scoop (commented):
Re-review of "hide hidden repos in repos picker on new task pane".

## Prior finding — resolved
1. **NewTask.svelte `cycleRepo` entering visible subset at index 1 when current repo is hidden** — FIXED. `cycleRepo` now branches on `cur === -1` and enters at `list[0]` on a forward step / `list[n - 1]` on a backward step, so neither end is skipped (ui/src/lib/components/NewTask.svelte:570-579). New tests assert both boundaries (`cycle from a hidden current repo enters the visible subset at its boundary` → vis1; `backward cycle … enters at the last visible repo` → vis2). Not re-raised.

## Verification
- `reconcileRealPathsToRaw` exists with signature `(realPaths: Iterable<string>, repos: readonly {path:string}[]): Set<string>` (src/repos.ts:103-118); server.ts:3333 calls it as `reconcileRealPathsToRaw(deps.store.hiddenRepoPaths(), baseRepos)` — both args match. `deps.store.hiddenRepoPaths()` returns a `Set<string>` (src/store.ts:1157), valid as the Iterable.
- i18n parity confirmed: `feat_newtask_hide_hidden_title` and `feat_newtask_hide_hidden_body` present in both ui/messages/en.json:2205-2206 and ui/messages/de.json:2205-2206.
- Feature announcement `targetId: "nt-repo"` matches the existing coachmark anchor `use:coachTarget={"nt-repo"}` (ui/src/lib/components/NewTask.svelte:800). `sinceVersion: "1.39.0"` is ahead of current package version 1.38.0 (package.json) and consistent with sibling entries (feature-announcements.ts:1863).
- Index alignment: NewTask Alt+digit uses `recentRepos(repos.filter((r) => !r.hidden))` (NewTask.svelte:602) and RepoSelect computes `recents` as `recentRepos(hideHidden ? repos.filter((r) => !r.hidden) : repos)` (RepoSelect.svelte:104-106) — both filter before `recentRepos`'s top-N slice (ui/lib/recentRepos.ts:14), so the pinned recents stay aligned with the digit shortcut.
- All-hidden fallback: `defaultRepoPath([])` returns `""` (NewTask.svelte:243), so `defaultRepoPath(r.filter(!hidden)) || r[0]!.path` (NewTask.svelte:282) keeps `repoPath` non-empty when every repo is hidden — matches the documented intent.
- RepoSelect `selected` still derives over the FULL `repos` list (RepoSelect.svelte:82), so a relaunch-seeded hidden repo still renders in the trigger label even though it's dropped from the dropdown; `searching` reveal-on-search restores hidden repos in both `filtered` and `recents`. `hideHidden` defaults to `false`, leaving other RepoSelect consumers unaffected.
- `RepoEntry.hidden?: boolean` added to ui/src/lib/types.ts and populated server-side via `hidden: hiddenSet.has(r.path)` (server.ts:3342).

No remaining issues.

PR #1270, opened 2026-06-30
fix(files): make whole Files tab the drop zone for uploads

scoop:
## What

The drag-and-drop upload target on the **Files** tab was limited to the inner `.list` element — small, and tiniest of all in the empty state where it was just the little "No files in the scratchpad yet." box. Files dropped anywhere else on the tab were ignored.

This moves the `ondragover`/`ondragleave`/`ondrop` handlers up to the outer `.files` container (the whole tab, already `height: 100%`), so **dropping anywhere on the Files tab uploads to the current directory**.

## How

- Handlers moved from `.list` → `.files`; child drop events bubble to the single handler.
- During an active drag, a **whole-tab overlay** (dashed blue border + tint + centered "Drop files to upload" hint) appears across the panel. It uses `pointer-events: none` so drag/drop events keep reaching `.files` and the existing `handleDragLeave` child-containment guard stays correct.
- Removed the old small `.list.drop-active` tint and its now-orphaned `transition`.
- The `region` role + aria-label moved with the handlers to `.files`; `.list` is now plain (no nested duplicate region).
- The empty-state resting hint is unchanged; the upload pipeline, file picker, and listing refresh are untouched.

## Notes

- Reuses the existing `files_upload_drop_hint` message — **no new i18n keys**.
- This is a refinement of the existing drag-drop feature (enlarging a too-small target), not a new capability, so **no feature-catalog entry** is required.
- Tokens only (`--color-blue`, `--fs-lg`, `--font-mono`); no raw literals.

## Verification

- `bun run check` — 0 errors, 0 warnings
- `vitest run --project browser FilesPanel` — 5/5 pass (incl. new drop-anywhere + overlay show/clear tests)
- `bun run check:i18n` — locales in parity

### DIFF
--- a/ui/src/lib/components/viewport/FilesPanel.browser.test.ts
+++ b/ui/src/lib/components/viewport/FilesPanel.browser.test.ts
@@ -122,26 +122,44 @@ describe("FilesPanel — 413 too-large error surfaces inline", () => {
 });
 
 describe("FilesPanel — drag-and-drop calls uploadScratchpadFile", () => {
-  it("calls uploadScratchpadFile on drop with correct args", async () => {
+  it("calls uploadScratchpadFile on drop anywhere in the tab with correct args", async () => {
     mockUpload.mockResolvedValue("dropped.txt");
 
     render(FilesPanel, { sessionId: "sess-4" });
     await waitForPanel();
 
     const file = new File(["drop"], "dropped.txt", { type: "text/plain" });
-    const list = document.querySelector<HTMLDivElement>(".list")!;
-    expect(list).not.toBeNull();
+    // The whole .files tab is the drop zone, not just the inner list.
+    const tab = document.querySelector<HTMLDivElement>(".files")!;
+    expect(tab).not.toBeNull();
 
     const dt = new DataTransfer();
     dt.items.add(file);
-    list.dispatchEvent(
+    tab.dispatchEvent(
       new DragEvent("dragover", { bubbles: true, cancelable: true, dataTransfer: dt }),
     );
-    list.dispatchEvent(
-      new DragEvent("drop", { bubbles: true, cancelable: true, dataTransfer: dt }),
-    );
+    tab.dispatchEvent(new DragEvent("drop", { bubbles: true, cancelable: true, dataTransfer: dt }));
 
     await expect.poll(() => mockUpload.mock.calls.length).toBe(1);
     expect(mockUpload).toHaveBeenCalledWith("sess-4", file, undefined);
+  });
+
+  it("shows the whole-tab drop overlay on dragover and clears it on drop", async () => {
+    mockUpload.mockResolvedValue("dropped.txt");
+
+    render(FilesPanel, { sessionId: "sess-5" });
+    await waitForPanel();
+
+    const tab = document.querySelector<HTMLDivElement>(".files")!;
+    const dt = new DataTransfer();
+    dt.items.add(new File(["drop"], "dropped.txt", { type: "text/plain" }));
+
+    tab.dispatchEvent(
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Review

The change satisfies the task: the drag-and-drop handlers (`handleDragOver`/`handleDragLeave`/`handleDrop`) now live on the outer `.files` container (ui/src/lib/components/viewport/FilesPanel.svelte:131-138) instead of the inner `.list`, so the whole Files tab is the drop zone. The overlay uses `pointer-events: none` (FilesPanel.svelte:241) so drag/drop events keep reaching `.files`, and the handlers were not otherwise altered — functionality verified by reading the unchanged handler bodies (FilesPanel.svelte:107-124).

### Verification
- i18n parity: `files_list_aria` and `files_upload_drop_hint` are reused (not new) and present in both catalogs — verified against ui/messages/en.json:2207,2214 and ui/messages/de.json:2207,2214. No new keys introduced.
- Accessibility: the `role="region"` + `aria-label` moved from `.list` to `.files` (FilesPanel.svelte:133-134); there is now a single region, no duplication.
- `handleDragLeave`'s `contains(e.relatedTarget)` guard (FilesPanel.svelte:114) still functions correctly with the overlay because the overlay is non-interactive (`pointer-events: none`).
- Tests updated to dispatch on `.files` and assert the `.drop-overlay` appears/clears (FilesPanel.browser.test.ts:124-165).

### Non-blocking nit
- `.files` is a scroll container (`overflow-y: auto; height: 100%`, FilesPanel.svelte:234-235) and `.drop-overlay` uses `position: absolute; inset: 0` (FilesPanel.svelte:238-239). An absolutely-positioned child of a scroll container is anchored to the top of the scrolled content and sized to the client (visible) height, so when the file list is long and scrolled down, the dashed overlay can be anchored above the viewport and fail to cover the visible area. The drop *behavior* is unaffected (handlers are on `.files`), but the visual hint may be misplaced. Consider `position: sticky`/a fixed-to-viewport overlay, or placing the overlay outside the scrolling region, so the hint always covers what the user sees.

scoop (commented):
## Review

The PR makes the entire Files tab the drop zone and resolves the previous revision's finding.

### Prior finding — RESOLVED
The earlier concern was that `.drop-overlay` (position: absolute; inset: 0) lived inside the `.files` scroll container, so on a long scrolled list it anchored to scrolled content rather than the viewport. The new diff:
- Splits scrolling into a new inner `.files-scroll` wrapper (`overflow-y: auto`, `flex: 1`, `min-height: 0`) at FilesPanel.svelte:147,239.
- Makes `.files` itself non-scrolling: `position: relative; overflow: hidden; height: 100%` (FilesPanel.svelte:233-238).
- Renders `.drop-overlay` as a direct child of `.files`, outside `.files-scroll` (FilesPanel.svelte:142). It therefore anchors to the viewport-sized container and always covers the visible area. Verified.

### Task satisfaction
- Drag handlers `ondragover/ondragleave/ondrop` moved from the inner `.list` to `.files` (FilesPanel.svelte:133-137), so the whole tab is droppable.
- `.files` fills the tab: Viewport.svelte:2131 mounts FilesPanel in `.panel-wrap` (absolute, inset 0, Viewport.svelte:3069-3073) and `.files` is `height: 100%`. Verified.
- Overlay uses `pointer-events: none` + `z-index: 2`, so drag events still reach `.files`; `handleDragLeave` (FilesPanel.svelte:112-116) ignores child crossings, avoiding flicker. Verified.

### i18n parity
- `files_upload_drop_hint` and `files_list_aria` exist in both catalogs — verified against ui/messages/en.json:2207,2214 and ui/messages/de.json:2207,2214. No new keys introduced.

### Cleanup
- Removed `.list.drop-active` CSS has no remaining references (grep over ui/src clean).

PR #1271, opened 2026-06-30
fix(prs): evict open-PR snapshot on merge so the panel drops the merged PR

scoop:
## Problem

Since the open-PR snapshot cache landed (#1253), merging a PR from the **repos PRs panel** re-listed the just-merged PR. The UI already does the right thing client-side — `PrsPanel.onmerged` optimistically drops the row, then silently refetches `GET /api/prs` — but that refetch reads the server's `OpenPrSnapshotService` cache (keyed by `forge.slug`, **120s TTL**), which still contained the merged PR, so it came right back. The merge handler only refreshed backlog *counts*, never the PR listing cache.

## Fix

Re-trigger a real fetch on the write by evicting the snapshot:

- **`OpenPrSnapshotService.invalidate(forge)`** — drops the cache **and** the in-flight entry for `forge.slug`, so the next `get()` is a guaranteed miss that starts a fresh fetch. No-op for incapable forges.
- Called **synchronously before the response** on the two interactive merge paths — `handlePrMerge` (`POST /api/prs/merge`, the reported bug) and `forgeMerge` (session-detail merge) — so the panel's refetch can't race a cache that still holds the PR.
- A **promise-identity guard** (`inflight.get(slug) === p`) in `load()` stops a fetch that started *before* the invalidation (e.g. a concurrent pr-poller refresh) from writing its stale, pre-merge list back into the cache after eviction.

**Why evict, not `refresh()`:** a detached `refresh()` keeps the stale entry until the new fetch resolves (a concurrent `get()` still returns stale); an awaited `refresh()` single-flight-*joins* a pre-merge in-flight fetch and returns/caches that stale snapshot. Eviction clears the inflight pointer so the next read is a genuinely fresh post-merge fetch.

## Scope

`forge.merge()` has six call sites. The two **interactive** ones are fixed here. The four **autonomous / non-panel** paths (`automerge` merge train, two `drain` auto-land paths, the epic-landing CTA) are left poller-reconciled (full-sweep cadence == TTL): no client does the optimistic-remove+refetch dance that exposes the window, and wiring the cache through those services is out of scope. `dependabot-rebase` isn't a merge (PR stays open) so needs no invalidation.

## Tests

- `test/open-pr-snapshot.test.ts` — `invalidate` evicts (next `get` refetches within TTL); clears inflight so a later `get` starts a fresh fetch and a late pre-invalidate fetch can't poison the cache; no-op for incapable forges.
- `test/server-prs.test.ts` — end-to-end against a **real `OpenPrSnapshotService`** (not a `get`-only stub, so it can't pass vacuously): warm → merge → the next `GET /api/prs` misses the still-TTL-fresh entry and returns the post-merge (empty) list.

Server-only change (no UI diff). Root `bun run lint` + `bun test ./test` pass (5524 tests, 0 fail).

### DIFF
--- a/src/open-pr-snapshot.ts
+++ b/src/open-pr-snapshot.ts
@@ -63,6 +63,24 @@ export class OpenPrSnapshotService {
     return this.cache.get(forge.slug!)?.value ?? null;
   }
 
+  /**
+   * Drop the cached entry (and any in-flight fetch pointer) for forge.slug so the
+   * NEXT get() is a guaranteed miss that starts a fresh fetch. Call after a write
+   * that mutates the open-PR set (e.g. a merge) so a client refetch can't read the
+   * stale snapshot. No-op for incapable forges.
+   *
+   * Clearing the inflight pointer means a get() after this won't join a fetch that
+   * started BEFORE the write (which would still carry the pre-write list); the
+   * promise-identity guard in load() then prevents that abandoned fetch from
+   * writing its stale value back once it resolves.
+   */
+  invalidate(forge: GitForge): void {
+    if (!this.isCapable(forge)) return;
+    const slug = forge.slug!;
+    this.cache.delete(slug);
+    this.inflight.delete(slug);
+  }
+
   /** True when the forge can supply a snapshot (non-null slug + method present). */
   private isCapable(forge: GitForge): boolean {
     return forge.slug != null && forge.listOpenPrSnapshot !== undefined;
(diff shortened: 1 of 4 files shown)


scoop (commented):
## Summary

The PR makes a merge write-action evict the per-repo open-PR snapshot so the backlog PRs panel's silent refetch (`GET /api/prs`) misses the still-TTL-fresh cache and re-lists the post-merge set. This directly satisfies the stated task. The implementation is correct and the supporting concurrency guard is sound.

## What I verified

- **`invalidate()` semantics** (`src/open-pr-snapshot.ts:77-82`): deletes both the cache entry and the inflight pointer, and no-ops for incapable forges via `isCapable` (`src/open-pr-snapshot.ts:84-87`). Confirmed both maps are the single source of truth for the cache (`src/open-pr-snapshot.ts:21-23`).
- **Promise-identity guard in `load()`** (`src/open-pr-snapshot.ts:105-121`): I traced the abandoned-fetch race. A second fetch can only start when `inflight.get(slug)` is empty (`src/open-pr-snapshot.ts:94-95`), and inflight is only cleared by the resolving `.then` or by `invalidate()`. So in the no-invalidate path the guard `this.inflight.get(slug) === p` is always true → no behavioral regression. After `invalidate()` replaces/clears the pointer, the pre-invalidate fetch correctly (a) does not write its stale value to cache and (b) does not delete the newer fetch's inflight slot. `p` is assigned before `inflight.set` and the `.then` callbacks run in a later microtask, so the identity comparison is well-ordered.
- **Graceful failure after eviction** (`src/server.ts:4196-4203`): if the post-eviction refetch fails, `get()` returns null (preserveOnError=false → `src/open-pr-snapshot.ts:129-135`) and the handler falls back to `forge.listPullRequests()` — so eviction can never surface an empty/stale list, only a fresh fetch.
- **Both merge paths covered**: interactive backlog merge `handlePrMerge` (`src/server.ts:4368`, synchronously before the 200) and session merge `forgeMerge` (`src/server.ts:2893`, before both the local-merge and host-merge branches). Optional-chained `deps.openPrSnapshot?.invalidate?.(forge)` is safe against the get-only test stubs.
- **Type change** (`src/server.ts:349-350`): `Pick<…,"get"> & Partial<Pick<…,"invalidate">>` keeps `invalidate` optional, consistent with the `?.invalidate?.` call sites and the existing get-only stubs at `test/server-prs.test.ts:285,305,335`. The real service wired in `src/index.ts:749,767` provides both methods, so the production injection still type-checks.
- **No other open-PR-set-mutating panel write exists**: the only other `/api/prs/*` write is `dependabot-rebase` (`src/server.ts:4458`), which posts a comment and does not change the open-PR set, so it correctly needs no invalidation.

- Opening a PR from a session (`forgeOpenPr`, `src/server.ts:2840`) does not invalidate the snapshot, so a newly opened PR is absent from the panel until TTL expiry; this is a session-driven action (not the backlog panel) and outside the task's merge scope.

PR #1274, opened 2026-06-30
fix(session): replace agent in same worktree

kai-osthoff:
## Summary
- add in-place agent replacement for existing sessions using the same DB row and worktree
- wire the Replace with picker to the new replace endpoint and update clients via session status payloads
- cover the route, service, store, and UI session-event behavior with tests

## Tests
- bun run lint
- bun run test
- cd ui && bun run test
- cd ui && bun run check
- cd ui && bun run check:i18n
- cd extension && bun run check

Manual steps: None.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -2538,6 +2538,39 @@ async function handleSessionRelaunch({ req, parts, deps }: Ctx): Promise<Respons
   }
 }
 
+const inFlightReplace = new Set<string>();
+
+// POST /api/sessions/:id/replace — stop the current agent process and start a new provider/model
+// in the SAME Shepherd session and worktree. This is intentionally not relaunch: no new branch,
+// no new worktree, no archive event.
+async function handleSessionReplace({ req, parts, deps }: Ctx): Promise<Response | null> {
+  if (!(req.method === "POST" && parts[2] && parts[3] === "replace")) return null;
+  const id = parts[2];
+
+  if (inFlightReplace.has(id))
+    return json({ error: "replace already in progress", code: "in_progress" }, 409);
+  inFlightReplace.add(id);
+  try {
+    const original = deps.store.get(id);
+    if (!original) return json({ error: "not found" }, 404);
+    if (original.status === "archived") return json({ error: "already archived" }, 409);
+
+    const choice = await parseModelChoice(req);
+    if (!choice.ok) return choice.res;
+
+    let session: Session;
+    try {
+      session = await deps.service.replaceAgent(id, choice.value);
+    } catch (e) {
+      return json({ error: e instanceof Error ? e.message : "replace failed" }, 502);
+    }
+    deps.events.emit("session:status", session);
+    return json({ session }, 200);
+  } finally {
+    inFlightReplace.delete(id);
+  }
+}
+
 // Serialize concurrent same-key calls behind a 409: add the key, run, always remove. Shared by
 // the variant + compare routes (the relaunch route keeps its own copy with bespoke messaging).
 async function guardInFlight(
@@ -2816,6 +2849,7 @@ async function handleSessions(ctx: Ctx): Promise<Response | null> {
     handleSessionAutopilot,
     handleSessionAutoMerge,
     handleSessionRelaunch,
+    handleSessionReplace,
     handleSessionVariant,
     handleSessionRestore,
     handleRelaunchUploads,
--- a/src/service.ts
+++ b/src/service.ts
@@ -2065,6 +2065,15 @@ export class SessionService {
     return copied;
   }
 
+  private listWorktreeUploads(worktreePath: string): string[] {
+    const dir = worktreeUploadsDir(worktreePath);
+    if (!existsSync(dir)) return [];
+    return readdirSync(dir)
+      .sort()
+      .map((name) => join(dir, name))
+      .filter((p) => statSync(p).isFile());
+  }
+
   /**
    * Spawn a fresh replacement for an existing task, carrying its prompt and all
    * per-task settings — including the spawn-baked ones that can't be changed after
@@ -2155,6 +2164,105 @@ export class SessionService {
 
     // Re-fetch AFTER the override writes so the returned session reflects them.
     return this.deps.store.get(newSession.id)!;
+  }
+
+  /**
+   * Replace the live agent process for an existing Shepherd session without creating a new
+   * worktree or session row. Used by "Replace with…" when the operator wants the same task and
+   * same checked-out worktree to continue under a different CLI/model (e.g. Claude → Codex).
+   */
+  async replaceAgent(
+    id: string,
+    opts: { agentProvider?: AgentProvider; model: string | null },
+  ): Promise<Session> {
+    const s = this.deps.store.get(id);
+    if (!s || s.status === "archived")
+      throw new Error(`cannot replace agent for ${id}: missing or archived`);
+
+    const agentProvider = opts.agentProvider ?? s.agentProvider ?? "claude";
+    const model = modelForProviderOrDefault(opts.model, agentProvider);
+    const claudeSessionId = agentProvider === "claude" ? randomUUID() : "";
+    const carriedImages = this.listWorktreeUploads(s.worktreePath);
+    const promptImages = carriedImages.slice(0, MAX_IMAGES);
+    if (carriedImages.length > promptImages.length)
+      console.warn(
+        `[replace] ${s.id}: ${carriedImages.length} images exceed cap ${MAX_IMAGES}; dropped ${carriedImages.length - promptImages.length}`,
+      );
+    const input: CreateSessionInput = {
+      repoPath: s.repoPath,
+      baseBranch: s.baseBranch,
+      prompt: s.prompt,
+      agentProvider,
+      model,
+      images: [],
+      planGateEnabled: s.planGateEnabled,
+      autopilotEnabled: s.autopilotEnabled,
+      research: s.research,
+      auto: false,
+    };
+    const composed = await this.composePromptArg(input, s.worktreePath);
+    const promptArg =
+      promptImages.length > 0
+        ? `${composed.promptArg}\n\nAttached images:\n${promptImages.join("\n")}`
+        : composed.promptArg;
+    const launch = await this.resolveCreateLaunch(
+      input,
+      { isolated: s.isolated },
+      promptArg,
+      s.id,
+      claudeSessionId,
+    );
+
+    const outcome = await this.prepareSpawnOrThrow(launch.argv, {
+      sessionId: s.id,
+      name: s.name,
+      worktreePath: s.worktreePath,
+      repoPath: s.repoPath,
+      isolated: s.isolated,
+      auto: false,
+      profileOverride: launch.profileOverride,
+      model: launch.spawnInput.model,
+      agentProvider,
+    });
+
+    try {
+      this.deps.store.update(s.id, {
+        herdrAgentId: outcome.terminalId,
+        claudeSessionId: agentProvider === "claude" ? claudeSessionId : "",
+        agentProvider,
+        model: launch.spawnInput.model,
+        status: "running",
+        lastState: "idle",
+        readyToMerge: false,
+        mergingSince: null,
+        mergingTrainId: null,
+        mergingPrNumber: null,
+        planGateEnabled:
+          agentProvider === "claude" ? (launch.spawnInput.planGateEnabled ?? null) : false,
+        planPhase: launch.planGateOn ? "planning" : null,
+      });
+      this.deps.store.setSandboxState(s.id, {
+        applied: outcome.applied,
+        degraded: outcome.degraded,
+        egressApplied: outcome.egressApplied,
+        egressDegraded: outcome.egressDegraded,
+      });
+    } catch (e) {
+      try {
+        this.deps.herdr.stop(outcome.terminalId);
+      } catch {
+        /* best-effort: leave the original agent registered if persistence failed */
+      }
+      throw e;
+    }
+
+    try {
+      this.deps.herdr.stop(s.herdrAgentId);
+    } catch {
+      /* best-effort: the replacement is already persisted */
+    }
+
+    return this.deps.store.get(s.id)!;
   }
 
   /**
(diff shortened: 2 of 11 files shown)


kai-osthoff (commented):
## Summary

This PR adds an in-place "replace agent" path (`POST /api/sessions/:id/replace` → `SessionService.replaceAgent`) so an operator can swap the live CLI/model (e.g. Claude → Codex) **without** creating a new branch, worktree, or session row. This directly addresses the reported symptom: the previous `ExperimentPicker` "replace" action called `relaunchSession`, which archived the original and `create()`d a **new** worktree (`src/service.ts:1961`) — so the Codex session landed in a *different* worktree, exactly the "wrong worktree" the task describes. The new path reuses the existing row + worktree.

I verified the core wiring is correct:
- The same worktree is reused: `replaceAgent` passes `s.worktreePath`/`s.repoPath`/`s.isolated` into `prepareSpawnOrThrow` (src/service.ts:2206-2208), which hands `ctx.worktreePath` to `herdr.start` as the cwd (src/service.ts:1488). The service test asserts `cwd: "/wt/orig"` and `worktreePath: "/wt/orig"` (test/service.test.ts:4758,4761).
- Permissions are correct for Codex: `resolveCreateLaunch` sets `profileOverride: "trusted"` for codex (src/service.ts:1918) and `buildCodexSpawnArgv` emits `--dangerously-bypass-approvals-and-sandbox` (src/service.ts:1708), matched by the test (test/service.test.ts:4762-4767).
- `store.update` SQL/column widening is sound: the new columns (`claudeSessionId`,`agentProvider`,`model`,`planGateEnabled`,`planPhase`) all exist (src/store.ts:657-658,2891-2903) and unpatched callers round-trip current values via `{...cur, ...patch}` (src/store.ts:1831). `planGateEnabled: boolean | null` (ui/src/lib/types.ts:59, src/types.ts:59) accepts the `null` written at src/service.ts:2222-2223.
- The `session:status` widening is safe: every other emitter sends only `{id,status[,hasScratchpadFiles]}` (src/index.ts:611-617, src/server.ts:2228); only the new replace emit (src/server.ts:2567) carries a full `Session`, and `applyStatus` keeps the `hasScratchpadFiles` undefined-guard intact because normal pushes never spread an explicit `undefined` (ui/src/lib/store.svelte.ts:317-318). WS forwards `data` verbatim (src/server.ts:5876).
- Route reachability: `parts[3]==="replace"` (src/server.ts:2552) doesn't collide with any sibling handler, and the 400/200 paths match the server tests (test/server-variant-compare.test.ts:158-184).

No locale-parity issue: this PR introduces no new message keys — `experiment_replace_*` and the `"replace"` mode already existed at the base commit (verified via `git show 630fa5bd:ui/src/lib/components/ExperimentPicker.svelte`).

The findings below are non-blocking robustness/quality items; the feature satisfies the task on the happy path.

## Findings

1. **`src/service.ts`: the live agent is killed before the replacement is confirmed, with no rollback.** `replaceAgent` calls `herdr.stop(s.herdrAgentId)` (src/service.ts:2198) and only updates the row *after* a successful spawn (src/service.ts:2215). If `prepareSpawnOrThrow` throws — reachable today when `herdr.start` fails (`HerdrUnavailableError`, or the `started agent not found for cwd` throw at src/herdr.ts:407) — the old agent is already gone, no new agent is running, and the row keeps its **old** `herdrAgentId` and unchanged `status` (e.g. still `running`). The route returns 502 but the session is left pointing at a dead terminal until the poller notices. Relaunch deliberately keeps the original alive until success (src/service.ts:2139-2154); consider a spawn-then-stop ordering or a best-effort restore on failure. (Stopping first does avoid the cwd-resolution ambiguity at src/herdr.ts:401-406, so the trade-off is real — but the agentless/stale-row outcome should at least be reflected in the row.)

2. **`src/service.ts`: `replaceAgent` forces `status:"running"` but does not clear merge/ready metadata.** The update patch (src/service.ts:2215-2225) omits `readyToMerge`, `mergingSince`, `mergingTrainId`, `mergingPrNumber`, so they round-trip from the old row (src/store.ts:1831). Replacing the agent on a session that was `readyToMerge` or mid-merge-train leaves that stale metadata on a now-`running` row, which can mislead the merge-train/UI. Confirm whether these should be reset on replace.

3. **`src/service.ts`: the replacement prompt drops issue + image context.** `replaceAgent` builds `input` with `prompt: s.prompt` and `images: []` and never sets `issueRef` (src/service.ts:2178-2189), so a session originally spawned from a GitHub issue (or with attached images) re-launches the new CLI without the issue body/comments block or the `Attached images:` references that the original spawn injected via `composePromptArg` (src/service.ts:1147-1166). For a same-task/same-worktree replace this may be intentional (the prior work is already on disk), but please confirm it's deliberate.

- src/service.ts:2215 — `replaceAgent` does not copy runtime toggles the way `relaunch` does (`setAutoMergeState`/`setAutopilotState`, src/service.ts:2144-2145); harmless today because the row is reused so those columns persist, but if a future change makes any of them spawn-time-derived this path would silently diverge from `relaunch`.

- src/herdr.ts:401-406 — resolving a just-started agent by `cwd` `.at(-1)` is ambiguous for non-isolated sessions sharing a repo cwd; `replaceAgent` inherits this but does not introduce it.
- ui/src/lib/components/ExperimentPicker.svelte — `relaunch_done`/`relaunch_archive_failed` message keys may now be unused by this component; whether they're still referenced elsewhere is outside this diff.

kai-osthoff:
Finding 3: I carried existing uploaded images into the replacement prompt. I did not re-send full GitHub issue context because the session row only persists issueNumber, not the original issue title/body/url payload needed by composePromptArg's issueRef block; reconstructing that would be a separate persistence/forge lookup change rather than an in-place provider swap fix.

kai-osthoff (commented):
## Re-review of `replaceAgent` (in-place provider swap)

The core fix is correct for the reported task: the "Replace with…" action in `ExperimentPicker.svelte:60` now calls the new `replaceSessionAgent` (`ui/src/lib/api.ts:773` → `POST /api/sessions/:id/replace`, `src/server.ts:2545`) instead of `relaunchSession`. Relaunch minted a *new* worktree/session and archived the old one — exactly the user's complaint ("I want to swap Claude→Codex in the SAME identical worktree"). `service.replaceAgent` (`src/service.ts:2162`) now reuses `s.worktreePath`, `s.isolated`, and the same session row, and for Codex sets `profileOverride: "trusted"` via `resolveCreateLaunch` (`src/service.ts:1918`), which addresses the "wrong worktree / permissions" symptom.

### Prior findings — all resolved

**Finding 1 (stop-before-spawn) — RESOLVED.** The order is now spawn-then-stop: `prepareSpawnOrThrow` runs first (`src/service.ts:2198`); if it throws, the old agent is untouched and the row is unchanged. Persistence is wrapped so that if `store.update` throws, the *new* agent is stopped and the original is left registered (`src/service.ts:2231-2238`). Only after successful persistence is the old agent stopped best-effort (`src/service.ts:2241`). Verified against the rollback `catch` block at `src/service.ts:2231`.

**Finding 2 (stale merge-train metadata) — RESOLVED.** The `store.update` patch now explicitly clears `readyToMerge:false, mergingSince:null, mergingTrainId:null, mergingPrNumber:null` alongside `status:"running"` (`src/service.ts:2222-2225`). These columns are persisted by the extended UPDATE in `src/store.ts:1833` (verified the SQL now lists `mergingSince/mergingTrainId/mergingPrNumber` and they bind `next.*`).

**Finding 3 (dropped images / issueRef) — RESOLVED / confirmed intentional.** Images are now carried via `carryRelaunchImages(s, undefined, s.id)` (`src/service.ts:2176`) rather than `images:[]`. The author's note that issue context is *not* reconstructed is verified: `composePromptArg`'s issueRef block needs `title/url/body` (`src/service.ts:1161-1166`), but the session row persists only `issueNumber` (`src/store.ts:2008` / store schema), so reconstructing it would require a forge lookup — a reasonable scoping decision. Accepted; not re-raised.

### Verified consistency
- New `store.update` columns (`claudeSessionId`, `agentProvider`, `model`, `planGateEnabled`, `planPhase`) all exist as real schema columns (`src/store.ts:657-658, 2891-2903`) and the patch type Pick was extended to match (`src/store.ts:1814-1825`). Other `update` callers pass partial patches over `{...cur, ...patch}`, so unspecified fields are preserved.
- `parseModelChoice`/`ModelChoice` reused exactly as in the variant/compare routes (`src/server.ts:2558` vs `2622`/`2649`); `replaceAgent`'s `{agentProvider?, model}` signature matches `startVariant`.
- The `session:status` payload is now the full `Session`; the UI side widened the type (`ui/src/lib/types.ts:1327`) and `applyStatus` spreads it (`ui/src/lib/store.svelte.ts:314-318`) so the swapped provider/model live-update. Partial emitters still type-check under the `& Partial<Session>` widening.
- No new i18n keys introduced; `experiment_replace_title/confirm` predate this PR (verified via `git show 630fa5bd:ui/src/lib/components/ExperimentPicker.svelte`).

### Non-blocking finding (see findings array)
Reusing the relaunch image-carry path on the *same* worktree accumulates duplicates. `copyOriginalUploads` (`src/service.ts:2057`) reads **every** file in `worktreeUploadsDir(s.worktreePath)` and copies it to staging with a fresh name; `composePromptArg`→`copyStagedIntoWorktree` (`src/uploads.ts:48-58`) then writes those back into the *same* `.shepherd-uploads` dir. A single replace works (the prompt references only the fresh copy), but each replace leaves an orphan copy on disk, and on the 2nd+ replace `copyOriginalUploads` re-reads originals **and** prior copies — so the prompt gets duplicate "Attached images" entries and, once the count exceeds `MAX_IMAGES`, distinct images can be displaced by duplicates. Relaunch never hit this because it targeted a fresh empty worktree. Consider deduping by basename, or reading only the original images, before the copy-back.

- src/service.ts:2227-2238: if `store.update` succeeds but the immediately-following `setSandboxState` throws, the catch stops the *new* agent while the row already points at it and the old agent is left orphaned. Practically unreachable (two synchronous writes to the same SQLite row), but the rollback only cleanly covers the pre-update failure path.

No other defects, security issues, or locale/contract mismatches found in the diff.

kai-osthoff (commented):
## Summary

This PR adds an in-place "Replace with…" path (`POST /api/sessions/:id/replace` → `SessionService.replaceAgent`) that stops the current agent and starts a different CLI/model in the SAME Shepherd session and worktree — no new branch, no new worktree, no archive. This directly matches the task (Claude → Codex in the identical worktree).

### Worktree + permissions (the reported symptom)

The replacement spawn reuses the original session's worktree and isolation:
- `prepareSpawnOrThrow` is called with `worktreePath: s.worktreePath` and `isolated: s.isolated` (src/service.ts:2216-2226), so the new agent runs in the identical checked-out worktree.
- Permissions are resolved through `resolveCreateLaunch`, which sets `profileOverride: "trusted"` for Codex (src/service.ts:1917-1920) and builds the `--dangerously-bypass-approvals-and-sandbox` Codex argv — verified by the new test asserting the argv prefix (test/service.test.ts). This is what fixes the "codex session has wrong worktree / permissions" symptom.

### Re-review of the prior finding

Prior finding (image copy-back leaves orphan duplicates / re-attaches images on each replace) is **ADDRESSED**:
- `replaceAgent` builds `CreateSessionInput` with `images: []`, and `composePromptArg` only copies when `input.images.length > 0` (src/service.ts:1147), so no `copyStagedIntoWorktree` write-back happens on the replace path.
- Existing uploads are instead listed in place via `listWorktreeUploads` (src/service.ts:2068-2073) and their absolute paths appended to the prompt, capped at `MAX_IMAGES` with a drop warning (src/service.ts:2185-2190).
- Verified by the new test "re-attaches existing uploads without copying them back": after two replaces `readdirSync(uploads)` is still `["diagram.png"]` and no staging dir is created (test/service.test.ts).

### Other correctness checks

- `store.update` now persists `claudeSessionId`, `agentProvider`, `model`, `planGateEnabled`, `planPhase`; the SET column list (15 cols) matches the 16 bound params incl. `id` (src/store.ts:1833-1850). For pre-existing callers that don't pass these in their patch, `next` merges `cur` (read fresh from DB via `get`, src/store.ts:1767-1771), so the same values are written back — no clobber, even against the separate `setPlanPhase` writer (src/store.ts:2671), since `get` reads live DB state.
- Failure handling verified: if the replacement spawn throws, `replaceAgent` throws before any store mutation, leaving the original `herdrAgentId`/`agentProvider`/`claudeSessionId` intact (test/service.test.ts, "keeps the old agent registered if replacement spawn fails"); the old agent is stopped only after the replacement is persisted (`order: ["start", "stop:term_orig"]`).
- `validateModelChoice` returns `{ agentProvider?; model: string | null }` (src/validate.ts:120-128), structurally compatible with `replaceAgent`'s `opts` type — route wiring via `parseModelChoice` is consistent (src/server.ts:2558, 2595-2603).
- `session:status` now carries a full `Session` for replace (src/server.ts:2567); the UI type was widened with `& Partial<Session>` (ui/src/lib/types.ts:1327) and `applyStatus` spreads it (ui/src/lib/store.svelte.ts:314). Server-side `session:status` consumers read only `id`/`status` (src/index.ts:776, 1424, 1479), so the richer payload is backward-compatible; status is `running`, so the PR poller's `status !== "running"` poll is correctly skipped (src/index.ts:777).
- UI: `ExperimentPicker` now calls `replaceSessionAgent` for the replace mode and surfaces errors via the pre-existing `m.experiment_action_failed` catch (ui/src/lib/components/ExperimentPicker.svelte:60,72). No new i18n keys introduced.

- src/service.ts: between the replacement spawn resolving and the old agent being stopped, both agents are briefly alive in the same worktree; the window is only the synchronous `store.update`/`setSandboxState` calls, so it is effectively negligible, but if the old agent is mid-edit when replace is triggered the overlap is real.
- src/service.ts: if `store.update` succeeds but `setSandboxState` throws, the catch stops the new terminal yet the row already points `herdrAgentId` at the now-stopped agent — a rare failure path that would leave the session referencing a dead terminal.

No blocking issues found; the prior finding is resolved and the task is satisfied.

kai-osthoff:
@scoop 😴

PR #1277, opened 2026-07-01
feat(owed): link the manual-steps chip to its Owed-lens details

scoop:
## What

The amber **"N manual step(s)"** chip on a session row was a dead end — the step
details live only on the **Owed** lens, with nothing linking the two. This makes the
chip a doorway.

Clicking the chip now switches the Herd to the **Owed** lens and scrolls to + briefly
highlights that session's step card. Because the chip counts frozen `session.manualSteps`
(which always exist) while the Owed lens lists only *outstanding* records, the click is
handled in every state so it's **never** a dead end:

| Session state | Result |
| --- | --- |
| Merged, record outstanding | Scrolls to + flashes the **live interactive** card |
| Merged, record ticked-complete / dismissed | Read-only **frozen** card — "No longer owed" |
| Pre-merge (no record yet) | Read-only **frozen** card — "will be tracked once merged" |
| Mobile lazy-load still in flight | Waits, then resolves to the live card (no false flash) |
| Owed load failed | Read-only **frozen** card — neutral "couldn't load" note |

## How

- **Store:** `post-merge-steps` gains a `settled` signal (set in `load()`'s `finally`) so the
  panel can tell "still loading" from "finished (success or failure)".
- **Chip → lens:** a new `onshowowed(id)` callback threaded `+page → Herd → HerdRowCtx →
  HerdGroup → UnitRow` (mirroring `onackmanualsteps`); the chip becomes a `<button>` that
  `stopPropagation`s so it doesn't also select the row.
- **Focus:** `+page.svelte` holds `owedFocusId/Snapshot/Nonce/HandledNonce`; the handled-nonce
  lives on the page (survives the panel unmounting on lens switch) so a WS refresh or a lens
  toggle can't re-flash a stale card. Focus clears when leaving the Owed lens (no phantom card).
- **Panel:** `PostMergeStepsPanel` scrolls to + flashes the target `data-session-id` card, or —
  when no live record matches — renders a read-only frozen card built from the click-time
  snapshot (in place of the empty state). Flash timer is cleared on unmount and on rapid re-clicks.
- **Discovery:** one What's-New catalog entry (`sinceVersion 1.39.0`).

## Tests

Store `settled` transitions; chip fires `onshowowed(id)` (not `onselect`); panel live-highlight,
loading-race (no false frozen flash), cleared / load-failed / pre-merge notes, empty-state
precedence, remount-guard (frozen card absent, no re-flash), and nonce-guard (no re-resolve on
WS refresh). Full UI suite: **2365 passing**; `check`, `check:i18n`, `lint`, feature-catalog,
branch-hygiene, glossary all green.

## Design / a11y notes

- The highlight and frozen-card styling use semantic tokens only (`--status-warn`, `--color-muted`,
  `--color-faint`, `--fs-meta`).
- **Tap-target waiver (please confirm):** the chip keeps its existing micro size (matching the
  adjacent `ACK STEPS` button), below the 44×44px floor, to avoid regressing the dense row's rhythm.

## Known follow-ups (not in this PR)

- The pre-existing `.manual-steps-ack` ("ACK STEPS") button likely has the same `.unit-hit`
  z-index unclickability this PR fixes for the chip link — untouched here to keep the diff scoped.
- Rare cosmetic: a pinned pre-merge frozen card isn't cleared if that session's PR merges in the
  background while the user lingers on Owed (self-heals on any lens toggle).


scoop (commented):
## Summary

The PR makes the "N manual step(s)" chip a link that switches to the Owed lens and scrolls-to/flashes the target session's card, with a read-only "frozen" fallback card when no live outstanding record exists (pre-merge, cleared, dismissed, or load-failed). The wiring is clean and complete, and it satisfies the task (there is now a path from the chip to the Owed details).

### Verification performed
- **i18n parity**: all six new keys exist in BOTH catalogs — `unitrow_manual_steps_link`, `owed_frozen_pre_merge_note`, `owed_frozen_cleared_note`, `owed_frozen_unknown_note`, `feat_manual_steps_chip_title`, `feat_manual_steps_chip_body` (verified `ui/messages/en.json:1894/1918-1920/1924-1925` and 6/6 present in `ui/messages/de.json`).
- **Prop threading**: `onshowowed` flows `+page.svelte` → `Herd.svelte` (ctx at `Herd.svelte:342`) → `HerdGroup.svelte:95` → `UnitRow.svelte:473`; focus props (`owedFocusId/Snapshot/Nonce/HandledNonce`, `onfocusresolved`) flow `+page.svelte:2187-2191` → `Herd.svelte:499` → `PostMergeStepsPanel.svelte`. All new props have defaults, so the sole other consumer (Herd owed branch) is unaffected. Only usage confirmed via grep.
- **Types**: `OwedFocusSnapshot.steps: ManualStep[]`; `ManualStep` (`types.ts:815`) has `id/text/postMerge`, consumed correctly by the frozen `{#each}` (`PostMergeStepsPanel.svelte`). `store.git[id]?.state === "merged"` valid against `types.ts:248`; `store.git[id]?.number` matches existing usage (`+page.svelte:139`).
- **CSS vars**: `--color-amber`, `--color-faint`, `--fs-meta`, `--status-warn` all defined (`app.css:20-21,82`).
- **Announcement**: `sinceVersion: "1.39.0"` matches the current in-progress release band; `id: manual-steps-chip-owed` is unique.
- **Nonce idempotency**: the `$effect` re-runs on `records`/`focusHandledNonce` changes but guards with `nonce === focusHandledNonce`, so it resolves once per click without looping.

### Non-blocking nit (see findings)
The frozen fallback card is never reconciled if a live record for the same session arrives *after* the click resolves. Because `postMergeSteps.settled` is a sticky flag (true forever after the first eager/failed load), the in-flight wait `snap.merged && !settled` only protects the very first load. In the merge→WS-refresh window (or after a failed eager load that a later retry fixes), a click resolves to a frozen card and marks the nonce handled; when `records` later gains the session's live record, the effect returns early and both the stale frozen card **and** the live card render for the same session — with the frozen card showing a misleading "No longer owed — completed or dismissed" note. Reachable today but only via a narrow race/transient failure, and it degrades gracefully (the click is never a dead end), so non-blocking.

No blocking correctness, security, or parity issues found.

scoop (commented):
## Re-review

The single point from the previous revision is addressed.

**Prior point #1 — frozen fallback never reconciled once a live record appears — RESOLVED.**
PostMergeStepsPanel.svelte:377-381 adds a render-time `frozenCard` derived:
```js
const frozenCard = $derived.by(() => {
  const p = pinnedFrozen;
  if (!p) return null;
  return records.some((r) => r.sessionId === p.sessionId) ? null : p;
});
```
The template now renders off `frozenCard` (PostMergeStepsPanel.svelte:130 empty-state guard, :134 card render) instead of `pinnedFrozen`. Because `records` is `$derived(postMergeSteps.records)`, when a live record for the same `sessionId` later lands (the sticky-`settled` / already-handled-nonce window the prior review flagged), `frozenCard` recomputes to `null` and the frozen card is suppressed — so the frozen "No longer owed" note can no longer co-render with the live card. This is exactly the requested "suppress the frozen card for any sessionId present in `records`" remedy, and it is reactive rather than dependent on the focus effect re-running (which the nonce guard would block). Covered by the `live record wins` test (PostMergeStepsPanel.browser.test.ts:297).

### Verification performed
- **i18n parity**: all five new keys exist in BOTH catalogs — `unitrow_manual_steps_link`, `owed_frozen_pre_merge_note`, `owed_frozen_cleared_note`, `owed_frozen_unknown_note`, `feat_manual_steps_chip_title/body` (en.json:35,43-45,48-49; de.json:9,17-19,22-23). Message calls in components resolve to these keys (PostMergeStepsPanel.svelte:147-149; UnitRow.svelte:591).
- **`settled` semantics**: verified in post-merge-steps.svelte.ts:13,29 — set in `finally`, so `settled && !loaded` ⇒ failed load, matching the frozen-note branch selection (PostMergeStepsPanel.svelte:147-149). Lens-open retry wires the load path (+page.svelte:1038-1041).
- **Prop threading**: `onshowowed` flows page → Herd (Herd.svelte:83,98) → HerdGroup (HerdGroup.svelte:95) → UnitRow (UnitRow.svelte:565,591); focus props (`focusSessionId/Snapshot/Nonce/HandledNonce`, `onfocusresolved`) flow page → Herd → PostMergeStepsPanel (Herd.svelte:107-112). Signatures agree.
- **Types**: `OwedFocusSnapshot.steps: ManualStep[]` (types.ts:853) matches `session.manualSteps` (ManualStep = {id,text,postMerge}, types.ts:815-819); frozen render uses only those fields (PostMergeStepsPanel.svelte:154-159).
- **Chip button**: stops propagation so the row's select doesn't also fire (UnitRow.svelte:593-594); falls back to a static span when `onshowowed` is absent (UnitRow.svelte:600). z-index:1 raises it above `.unit-hit` (UnitRow.svelte:619-621).
- **Feature announcement**: `targetId` is optional (feature-announcements.ts:15), so the anchorless entry is valid.

No blocking issues found.

### Minor (non-blocking, optional)
- +page.svelte:1952 sets `herdFilter = "owed"` unconditionally, but `owedFocusSnapshot`/`owedFocusNonce` are only set when the session is found in `store.sessions` (:1941). Since the chip only renders on a live session row, the not-found path is unreachable in practice — no action needed.

PR #1279, opened 2026-07-01
fix(autopilot): align Codex full-auto recovery

kai-osthoff:
## Summary
- let isolated Codex autopilot sessions participate in the same full-auto post-PR recovery path as Claude, while keeping non-isolated Codex fail-closed because `codex resume --last` can target a sibling session
- require autopilot agents to run relevant local lint/check/test commands before commit, push, or PR creation in spawn-time, plan-go, and reactive open-PR steers
- update coverage for Codex full-auto eligibility and the verification wording

## Verification
- `bun test ./test/full-auto.test.ts`
- `bun test ./test/autopilot.test.ts`
- `bun test ./test/service.test.ts -t "autopilot directive|planGoSteer|codex"`
- `bun run lint`
- `bun run typecheck`
- `bun test ./test`
- `bunx prettier --check src/autopilot.ts src/full-auto.ts src/service.ts test/autopilot.test.ts test/full-auto.test.ts test/service.test.ts`
- pre-push hook: prettier, eslint, gates, tsc, ext, root-tests, ui, fallow audit

### DIFF
--- a/src/autopilot.ts
+++ b/src/autopilot.ts
@@ -47,7 +47,9 @@ export function epicBaseDirective(baseBranch: string): string {
 export function openPrSteer(draftMode: boolean, baseBranch: string): string {
   const steer = [
     "You're in autopilot and you've stopped, but there's no pull request yet. Commit your",
-    `work, push the branch, and open a PR (gh pr create --base ${baseBranch}). If something`,
+    `work, push the branch, and open a PR (gh pr create --base ${baseBranch}). Before committing,`,
+    "pushing, or opening the PR, run the relevant local lint/check/test commands from the repository",
+    "instructions for the files you touched, and fix failures before proceeding. If something",
     "genuinely blocks that, say specifically what you need.",
   ].join("\n");
   return draftMode ? `${steer} ${DRAFT_PR_NOTE}` : steer;
--- a/src/full-auto.ts
+++ b/src/full-auto.ts
@@ -15,17 +15,19 @@ import { isEpicIntegrationBranch } from "./epic-branch";
  * autoMergeEnabled override — draft PRs must go through sign-off before they can be landed.
  */
 export function isFullAuto(
-  s: Pick<Session, "autopilotEnabled" | "autoMergeEnabled" | "baseBranch" | "agentProvider">,
+  s: Pick<Session, "autopilotEnabled" | "autoMergeEnabled" | "baseBranch" | "agentProvider"> & {
+    isolated?: boolean;
+  },
   cfg: Pick<RepoConfig, "autopilotEnabled" | "autoMergeEnabled" | "draftMode">,
 ): boolean {
   // Epic children are squash-merged into their integration branch by the drain's retire path,
   // never carried by the merge train — exclude them regardless of repo auto-merge config.
   if (isEpicIntegrationBranch(s.baseBranch)) return false;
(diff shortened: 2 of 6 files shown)


kai-osthoff (commented):
## Summary

Two coherent changes that address the task (Codex autopilot opened a PR without running lint/tests, and Codex/Claude diverged in full-auto handling):

1. **Verification steer** added to the three real drive-to-PR directives so the agent runs local lint/check/test before commit/push/PR — `AUTOPILOT_DIRECTIVE` (`src/service.ts:479`), `PLAN_GO_STEER_BASE` (`src/service.ts:814`), and `openPrSteer` (`src/autopilot.ts:50`).
2. **`isFullAuto` codex gate relaxed to isolated-only** (`src/full-auto.ts:30`): isolated Codex now gets the same post-PR merge-train recovery path as Claude, while non-isolated Codex still fails closed.

## Verification

- **Provider parity (the core divergence concern).** Claude receives `AUTOPILOT_DIRECTIVE` via `composeSystemPrompt` (`src/service.ts:888`); Codex has no `--append-system-prompt`, so it receives the same directive inline through `buildCodexSpawnArgv` (`src/service.ts:1722`). Both providers therefore get the new lint/test instruction — divergence is addressed, not just Claude.
- **`isFullAuto` is correctly wired.** `Session.isolated` is a required `boolean` (`src/types.ts:24`), and every caller passes a full `Session`: `src/drain.ts:465`, `src/automerge.ts:109`, `src/index.ts:745`, `src/index.ts:1311`. So `s.isolated` resolves to the real persisted value rather than always-undefined. The `Pick<...> & { isolated?: boolean }` optional form is deliberate so the fail-closed test (`test/full-auto.test.ts:72`) can omit the flag.
- **Consistency with the sibling guard.** `isFullAuto`'s new `s.isolated !== true` cutoff mirrors `AutopilotService.eligible()`'s `!s.isolated` codex guard (`src/autopilot.ts:214`); both fail closed on `false`/`undefined` and share the documented rationale (`codex resume --last` can steer a sibling session in a shared cwd, safe only in a one-session isolated worktree). The two boolean forms are semantically identical for `boolean|undefined`.
- **Merge-config math.** With `fullAutoCfg = {autoMergeEnabled: true, draftMode: false}` (`test/full-auto.test.ts:11`), the new isolated-codex-inherits case (`autoMergeEnabled: null`) resolves `merge = cfg.autoMergeEnabled = true`, so the added `→ true` assertion (`test/full-auto.test.ts:93`) is correct.
- **Tests match the strings.** The assertions on `"verified local changes"` / `"lint/check/test"` (`test/service.test.ts:3069-3070`, `test/service.test.ts:4073`) and `"Before committing"` / `"lint/check/test"` (`test/autopilot.test.ts:237-238`) all correspond to text present in the updated directives.

No i18n parity concern: all touched strings are Shepherd-owned agent-facing English (documented `never i18n'd` at `src/autopilot.ts:88`, `src/service.ts:507`), not message-catalog keys.

No blocking issues found.

### Note (non-blocking, no change requested)

`EMPTY_COMPLETION_STEER` (`src/autopilot.ts:89`) and `epicBaseDirective` (`src/autopilot.ts:34`) also mention opening a PR but were intentionally left without the lint/test note — the former handles the "produced nothing" case and the latter is purely a base-branch-targeting directive, so adding verification wording there would be out of place. This is correct as-is.

kai-osthoff:
@scoop 🏁   🚗💨

PR #1280, opened 2026-07-01
Fix Learnings attention chips filtering

kai-osthoff:
## Summary
- widen the desktop Learnings drawer from 520px to 620px
- make Needs attention repo chips filter the drawer to that repo instead of scrolling
- add active chip state and localized ARIA labels for the filter behavior

## Checks
- cd ui && bun run check:i18n
- cd ui && bun run check
- cd ui && bun test src/lib/components/learnings-drawer.test.ts


kai-osthoff (commented):
## Summary

The PR accomplishes the stated task: (1) the desktop side pane is widened from 520px to 620px, and (2) the "needs attention" chips now act as filters (scoping the drawer to one repo's learnings) instead of jump links.

### What I verified

- **Width bump is consistent across all three spots.** `slide.x` (`ui/src/lib/components/LearningsDrawer.svelte:137`), the `.drawer` `width: min(620px, 100vw)` (`:299`), and the fly-in offset all moved 520→620 together, so the transition still starts fully off-screen; the comment at `:136` ("Matches the drawer width") remains accurate. On mobile the `min(…, 100vw)` cap prevents overflow.
- **Chip-as-filter logic is sound.** `repoOnly` state + `toggleRepoFilter` (`:191`) toggle the selected repo; `displayGroups` (`:170`) filters `groups` to the selected repo and still honors the existing flagged/over-budget lenses. The `$effect` at `:166` clears `repoOnly` when the repo leaves the attention list, so the filter can't get stuck on a repo that no longer has a chip — mirroring the existing auto-reset pattern for the lens toggles.
- **TriageBand contract updated on both sides.** Props changed `onjump`→`onselect` + `activeRepo`; the sole consumer at `LearningsDrawer.svelte:275` passes the new props. No other consumer exists (grepped). `active`/`aria-pressed` styling uses `--color-amber`, which is defined in `ui/src/app.css`.
- **AttentionItem shape matches usage.** `reposNeedingAttention` returns `{repoPath, droppedCount, flaggedCount}` (`ui/src/lib/components/learnings-drawer.ts:165-175`); TriageBand reads exactly those fields.
- **i18n parity holds.** The old `learnings_triage_jump_aria` is removed from both catalogs and the two new keys `learnings_triage_filter_aria` / `learnings_triage_filter_active_aria` exist in both `en.json:1524-1525` and `de.json:1524-1525` (only two catalogs in `ui/messages/`). No dangling reference to the removed key remains in `ui/src`.
- `repoAnchorId` is still imported and used by the `focusRepo` deep-link effect (`LearningsDrawer.svelte:185`), so removing `jumpToRepo` left no unused import.

No bugs, security issues, or quality problems found in the diff. The empty-view edge case (selecting a repo chip while a non-matching lens toggle is active yields a blank list) is pre-existing lens-filter behavior, not a regression introduced here.

kai-osthoff:
@scoop 🚗

PR #1287, opened 2026-07-01
feat(steers): bind saved steers (quick buttons) to specific repos

scoop:
## What & why

Saved **steers** (the operator quick-buttons) were global — every steer showed on every repo. This adds a per-steer **repo allowlist** so a steer can be limited to specific repos. Empty/absent allowlist = universal (unchanged behavior).

Motivating example: `discuss-feedback` and `notion-task-to-issue` → **flowagent only**; `grill me` → **universal**.

## How it works

- **Model:** `Steer.repos?: string[]` — an allowlist of repo **names**. Empty/absent = universal. Mirrors how a steer already carries its `inSteerBar` / `onIssues` surface flags.
- **Match key is the repo NAME, not the path** — deliberately. `session.repoPath` is realpath-resolved (via `safeRepoDir`), while `listRepos`/backlog use the raw `join(repoRoot, name)` form. Matching full paths would silently hide bound steers everywhere under a symlinked repoRoot/entry. Names sidestep that split.
- **Normalization backbone:** `RepoEntry.realPath` (added) + a client repos index (`repos.svelte.ts`) that maps **both** the raw `path` and the `realPath` → `name`. So a realpath'd session repo and a raw backlog repo both resolve to the same name. A `test/repos.test.ts` symlink-fixture test pins this invariant (the guard against "vanish everywhere").
- **Gated surfaces (all via one shared `steerAppliesToRepo` helper):** session steer-bar chips (`SteerBar`), compose canned chips (`ComposeBar`), backlog issue quick-buttons (`IssuesPanel`), and the Broadcast dialog.
- **Editor:** Settings → Saved Steers gains a per-steer **Repos** control (anchored, non-modal popover) to pick the bound repos; default "All repos".

## Decisions (signed off during planning)

- **Broadcast honors the binding:** a bound steer is offered only when it's universal OR *every* currently-selected target session is in its allowlist; a picked steer that becomes incompatible when the selection changes is cleared. A flowagent-only steer can never be broadcast to a non-flowagent session.
- **First-run issue-action nudge stays global:** `issueActionsUnset` still fires only when no steer has the Issues surface at all (it's a one-time onboarding hint, not a per-repo configurator).
- **Stale bindings are surfaced, not silent:** a steer bound only to now-missing repos renders nowhere (correct), and the editor shows those names as removable "unknown repo" chips so it's visible and fixable — never a silent vanish.

## Testing

- Root: `bun run lint` clean, `bun run typecheck` clean, `bun run test` 5535 pass / 0 fail (incl. new `validateSteerRepos` cases + symlink normalization invariant).
- UI: `bun run check` 0 errors, `check:i18n` parity (2232 keys × 2 locales), `bun run test` 2377 pass (incl. `steer-scope` + `repos.svelte` unit tests).
- Gates: branch-hygiene, feature-catalog, glossary, and `fallow audit` all green.
- New feature-announcement entry (`steer-repo-binding`, ships 1.39.0) + EN/DE message keys.

## a11y note (house-rule waiver)

The unknown-repo chip's **remove button is 32×32px**, but it sits inside a `min-height: 44px` chip row, so the effective tap target meets the 44px floor; this matches the existing compact-icon pattern in sibling components. Flagging it explicitly per the sizing-floor rule — reviewer can confirm or ask for a 44×44 bump.


scoop (commented):
## Summary

The PR lets a saved steer carry a `repos` allowlist (repo NAMES) that gates where its chip/quick-button appears — steer bar, compose bar, backlog issue actions, and broadcast picker — satisfying the task (bind `discuss-feedback`/`notion-task-to-issue` to a repo, keep `grill me` universal). The design is sound and well-tested.

### What I verified

- **Model + validation.** `Steer.repos?: string[]` added to both `src/types.ts:76` and `ui/src/lib/types.ts:1522`. `validateSteerRepos` (src/validate.ts:97) rejects non-arrays, >`STEER_MAX` (=40, verified src/validate.ts:763), non-strings, empty-after-trim, and >255-char names; trims + dedupes; empty ⇒ `undefined` (universal). Matches the tests at test/validate.test.ts:283-307. The `validateSteerItem` split preserves prior behavior (label/text + surfaces + both-off guard still reject identically).
- **Path resolution.** `RepoEntry.realPath` is populated in every server-side constructor: `listRepos` (src/repos.ts:72), `cloneRepo` (199), `createProject` (527), `forkRepo` (752), and passed through `GET /api/repos` via `...r` spread (verified src/server.ts:3396-3405). `safeRepoDir` returns `realpathSync(...)` (verified src/validate.ts:705-713), so `session.repoPath` is the realpath form and the client `pathIndex` (ui/src/lib/repos.svelte.ts:29) — which maps BOTH `path` and `realPath` to name — resolves session (realpath) and backlog (raw) paths to the same entry. `steerAppliesToRepo` (ui/src/lib/steer-scope.ts:6) treats empty/absent as universal and null-name + non-empty allowlist as hidden, matching its tests.
- **All gating sites** (SteerBar.svelte:31, ComposeBar.svelte:54, IssuesPanel.svelte:47, BroadcastDialog.svelte:31) route through `steerAppliesToRepo(s, repos.nameFor(repoPath))`; `repoPath` props exist and are wired (Viewport.svelte:2184 passes `session.repoPath`). `repos.load()` is invoked on app mount (+page.svelte:1320) and in the editor's `onMount`.
- **i18n parity.** All 8 `steerseditor_repos_*` keys plus `feat_steer_repo_binding_title`/`_body` exist in BOTH ui/messages/en.json:348-355,363-364 and ui/messages/de.json:320-327,335-336. The `{n}` param in `steerseditor_repos_count` is present in both catalogs and supplied at SteersEditor.svelte call site. Only en+de catalogs exist.
- **Required-field migration.** `realPath` is now required on `RepoEntry`; every production constructor sets it and all UI consumers merely receive server data — no un-migrated literal in production code.

### Nit (non-blocking)

- `ui/src/lib/components/BroadcastDialog.svelte:41`: the stale-steer guard identifies the currently-picked steer by `steers.list.find((s) => s.text === text)` (text equality) rather than tracking the picked steer's id. It works for the intended flow, but it also means a user who types free-form broadcast text that happens to equal a bound steer's text (whose targets don't all match) will have their message silently cleared. Tracking the picked steer id would be more robust. Minor.

Could not verify the actual repo directory name for `flowagent` against a live deployment (not in this repo) — the mechanism binds by `listRepos` dir name, which is the correct key; assumed the user's repo is named accordingly.

scoop (commented):
## Re-review

The task — bind saved-steer quick-buttons to specific repos (empty allowlist = universal) — is satisfied. Steers now carry an optional `repos` name-allowlist, and every rendering surface gates on it via a single `steerAppliesToRepo` helper.

### Prior finding — RESOLVED
1. **BroadcastDialog stale-steer clear matched by text equality.** Now fixed. The picked steer is tracked by id (`pickedId`, ui/src/lib/components/BroadcastDialog.svelte:16), the highlight uses `class:on={pickedId === s.id}` (BroadcastDialog.svelte:167), `oninput` resets `pickedId = null` (BroadcastDialog.svelte:185-188), and the stale-clear `$effect` keys off `!availableSteers.some((s) => s.id === pickedId)` rather than text (BroadcastDialog.svelte:52-59). Free-form text equal to a bound steer's text is no longer wiped.

### Verification performed
- **Message parity (en/de).** All 8 `steerseditor_repos_*` keys + `feat_steer_repo_binding_{title,body}` exist in both catalogs (ui/messages/en.json:986-993, ui/messages/de.json:986-993, and the feat_* pair at en.json:1933-1934 / de.json:1935-1936). All 8 keys used in SteersEditor.svelte are defined; `steerseditor_repos_count` uses the `{n}` param consistently (en.json "{n} repos").
- **`realPath` propagation.** `RepoEntry.realPath` is now required in both src/types.ts:44 and ui/src/lib/types.ts:1450. All four RepoEntry constructions in src/repos.ts include it (src/repos.ts:72,196-200,520-525,745-751), and GET /api/repos spreads `...r` so `realPath` reaches the client (src/server.ts:3396-3405).
- **Path resolution invariant.** `repos.nameFor` resolves both raw `path` and `realPath` to the same name via `pathIndex` (ui/src/lib/repos.svelte.ts:23-33). `session.repoPath` is realpath-resolved via `safeRepoDir` (src/repos.ts:41-43 doc; asserted by test/repos.test.ts:213 `safeRepoDir(e.path, root) === e.realPath`), while backlog/IssuesPanel pass raw `join(repoRoot, name)` — both forms are keyed in the index, so either resolves. Repo names are directory names under repoRoot and thus unique, making a name-keyed allowlist safe.
- **All steer surfaces gated.** SteerBar.svelte:33, ComposeBar.svelte:54, IssuesPanel.svelte:47, and BroadcastDialog.svelte:37-45 all filter through `steerAppliesToRepo`. The only other `steers.list` reader (+page.svelte:492 `issueActionsUnset`) is a global hint check, correctly ungated. `repos.load()` is invoked app-wide (+page.svelte:1320) and in the editor (SteersEditor.svelte:64); all consumers derive reactively, so a bound steer simply re-appears once the index loads.
- **`steerAppliesToRepo` semantics.** Absent/empty allowlist → universal; non-empty with null or non-member name → hidden (ui/src/lib/steer-scope.ts:6-8), matching its tests.
- **Server validation.** `validateSteerRepos` trims, de-dupes, rejects non-string/empty-after-trim/>255-char entries, and caps length at STEER_MAX=40 (src/validate.ts:790-801, STEER_MAX at src/validate.ts:763). Empty-after-dedup collapses to `undefined` (universal), matching the client's `s.repos = cur.size ? [...cur] : undefined` (SteersEditor.svelte:205).
- **Dedup/loop safety.** The BroadcastDialog clear `$effect` short-circuits on `pickedId == null` before writing, so setting `pickedId = null` cannot re-trigger a loop.

### Non-blocking observations (no change required)
- The repos allowlist length reuses the steer-count cap `STEER_MAX` (40) (src/validate.ts:791). A user with >40 repos who binds a steer to all of them would have the whole PUT rejected server-side; the client `valid` guard (SteersEditor.svelte:234-238) does not mirror this cap. This degrades gracefully to a surfaced error rather than silent loss, and >40 repos is rare — left as an intentional bound.

No blocking issues found; the prior finding is addressed and the implementation is internally consistent.

PR #1289, opened 2026-07-01
fix: keep PR polling alive under GitHub GraphQL limits

kai-osthoff:
## Summary
- keep the session PR full sweep running during GitHub GraphQL backoff
- fall back to GitHub REST for per-session PR status when GraphQL is exhausted
- add regression coverage for REST fallback and rate-limited poller behavior

## Tests
- bun run lint
- bun test test/forge/github.test.ts test/pr-poller.test.ts
- pre-push hook passed during git push


kai-osthoff:
@scoop ✅

PR #1290, opened 2026-07-01
docs(install): external-tester install for the private repo

scoop:
## What

Adds a **Private repo & external testers** subsection under Install in the README, documenting how to get an external tester running when the repo is private and the `curl|bash` one-liner can't work.

## Why

The one-liner fails without repo access at two points: the raw `install.sh` URL 404s and the `git clone` needs auth. Nothing in the docs pointed a first tester at a working path.

## Two documented paths

- **Air-gapped tarball** — uses the installer's existing `SHEPHERD_SRC` knob (`deploy/install.sh:160,181`). `git archive` → send → `SHEPHERD_SRC=… bash install.sh`. No GitHub access to grant.
- **Repo collaborator** — read access + `gh auth login` + direct clone, for a tester who should track `main` and get updates.

Also flags the Linux + systemd support caveat via the existing OS-matrix anchor.

## Scope

Docs only — one file, no code/behavior change. Prettier clean. Carries `[no-feature-entry]` (no user-facing UX).

### DIFF
--- a/README.md
+++ b/README.md
@@ -135,7 +135,7 @@ before running it. Read the script first:
 
 Shepherd's GitHub repo is private, so the `curl|bash` one-liner above does **not** work for someone
 without repo access — both the raw `install.sh` URL and the `git clone` it performs need
-authentication. Two ways to get an external tester running:
+authentication. Three ways to get an external tester running:
 
 **Air-gapped tarball (no GitHub access needed).** The installer can land the source from a local
 tarball via `SHEPHERD_SRC` instead of cloning — the fastest way to hand the code to a first tester
@@ -152,6 +152,29 @@ SHEPHERD_SRC=~/shepherd.tar.gz bash install.sh
 
 `git archive` ships tracked files only; that's sufficient, since the installer runs `bun install`
 and builds from source. Trade-off: no in-place updates — each new build is a fresh tarball.
+
+**Release tarball (self-serve, versioned).** Every merge of the release-please PR cuts a tagged
+GitHub release, and GitHub auto-generates a source tarball for each tag — so there's nothing to
+`git archive` or upload. The tester downloads the tag they want and feeds it to `SHEPHERD_SRC`. This
+is a **private** repo, so the download needs a token — but a **fine-grained, read-only PAT scoped to
+just this repo** is enough (no collaborator membership; revocable):
+
+```bash
+# tester — with `gh auth login` done, fetch the source archive for a tag:
+gh release download v1.4.0 --archive=tar.gz -R erwins-enkel/shepherd
+# → shepherd-1.4.0.tar.gz
+
+# …or with a bare PAT, straight from the API (302s to a signed codeload URL):
+curl -fL -H "Authorization: Bearer $GH_TOKEN" \
+  https://api.github.com/repos/erwins-enkel/shepherd/tarball/v1.4.0 -o shepherd.tar.gz
+
+# then, same as the air-gapped path:
+SHEPHERD_SRC=~/shepherd.tar.gz bash install.sh   # install.sh is inside the tarball at deploy/
+```
+
+Note there is **no anonymous URL** for private-repo archives or release assets — the token is
+unavoidable, but a scoped read-only PAT is far less than write access. Testers self-serve any
+tagged version and re-download to update.
 
 **Repo collaborator (sustainable, allows updates).** Grant read access, then the tester
 authenticates once and clones directly (the public one-liner still won't work — the raw URL is
--- a/README.md
+++ b/README.md
@@ -146,8 +146,9 @@ with nothing to grant:
 git archive --format=tar.gz -o shepherd.tar.gz HEAD
 # send shepherd.tar.gz to the tester (install.sh lives inside it at deploy/install.sh)
 
-# tester — extract deploy/install.sh from the tarball (or receive it alongside), then:
-SHEPHERD_SRC=~/shepherd.tar.gz bash install.sh
+# tester — extract just the installer, then run it against the tarball:
+tar -xzf ~/shepherd.tar.gz deploy/install.sh
+SHEPHERD_SRC=~/shepherd.tar.gz bash deploy/install.sh
 ```
 
 `git archive` ships tracked files only; that's sufficient, since the installer runs `bun install`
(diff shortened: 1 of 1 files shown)


scoop (commented):
## Review

This PR adds a **Private repo & external testers** section to `README.md` documenting two ways to onboard an external tester against a private GitHub repo: an air-gapped `SHEPHERD_SRC` tarball, and a repo-collaborator clone. It directly answers the task (tarball vs. adding to the repo), and every concrete claim checks out against the actual installer.

### Verified against ground truth

- **Tarball extraction** — README's `SHEPHERD_SRC=~/shepherd.tar.gz bash install.sh` relies on the installer extracting a file arg as a tarball. Confirmed at `deploy/install.sh:182-185` (`tar -xf "$src" -C "$SHEPHERD_DIR"`); `tar -xf` auto-detects gzip, so `git archive --format=tar.gz` output works.
- **Tarball layout** — `git archive HEAD` yields tracked files at top level (e.g. `deploy/install.sh`, `deploy/provision.ts`), extracted straight into `$SHEPHERD_DIR`; the installer then validates via `deploy/provision.ts` (`deploy/install.sh:220-224`). Consistent — no nesting mismatch.
- **"no in-place updates" for the tarball path** — confirmed: `resolve_from_src` (`deploy/install.sh:159-163, 179-198`) never establishes a git checkout, so `update.sh`/`git pull` cannot apply. Accurate trade-off.
- **Repo slug** — `erwins-enkel/shepherd` matches `origin` remote and `REPO_URL` (`deploy/install.sh:37`).
- **Collaborator flow** — clones to `~/Work/shepherd`, which equals default `SHEPHERD_DIR` (`deploy/install.sh:39`); re-running the local `deploy/install.sh` then hits the non-destructive `update_existing_checkout` path (`deploy/install.sh:165-172, 205-216`). Matches the README's "keeps them on `main` with `update.sh` / `git pull`".
- **Cross-references** — `update.sh` exists (`deploy/update.sh`); `[OS matrix](#os-matrix)` resolves to `### OS matrix` (`README.md:117`); "full support is Linux + systemd" matches the matrix row (`README.md:121`).

No functional code, no user-facing strings/message catalogs, and no signatures are touched, so there is no i18n-parity or caller-agreement surface to check.

No blocking issues and nothing actionable to change.

### Stated limitations
- `git archive --format=tar.gz` depends on the tester's git registering the `tar.gz` format; this is built into modern git and is not verifiable from this repo — assumed as written.

scoop (commented):
## Summary

Documentation-only change (README.md, +36 lines) adding a "Private repo & external testers" section that answers the task directly: it explains that the `curl|bash` one-liner fails for a private repo and gives two concrete onboarding paths — an air-gapped `SHEPHERD_SRC` tarball and a repo-collaborator clone. The mechanism described matches the actual installer.

## Verification

- `SHEPHERD_SRC` tarball flow is real: `deploy/install.sh:160-197` (`resolve_from_src`) treats a file arg as a tarball and `tar -xf`'s it into `$SHEPHERD_DIR`, and the header docs it (`deploy/install.sh:19`, `:152`). The README's `SHEPHERD_SRC=~/shepherd.tar.gz bash install.sh` is accurate.
- Repo slug `erwins-enkel/shepherd` is consistent everywhere: `deploy/install.sh:37` (`REPO_URL`), the existing one-liner (`README.md:97`), and `git remote`.
- Collaborator flow is internally coherent: it clones to `~/Work/shepherd`, and `SHEPHERD_DIR` defaults to `$HOME/Work/shepherd` (`deploy/install.sh:39`), so the subsequent `bash ~/Work/shepherd/deploy/install.sh` reaches the non-destructive `update_existing_checkout` path (`deploy/install.sh:165-171`, `:205`) rather than re-cloning.
- `[OS matrix](#os-matrix)` anchor resolves to the `### OS matrix` heading (`README.md:117`).
- `update.sh` / `git pull` ongoing-update claim matches `deploy/update.sh:5,34`.
- `git archive --format=tar.gz` is a standard, built-in git output format — valid as written.

## Nit

`README.md:149-150`: the tester block says "extract **deploy/install.sh** from the tarball" but then runs `bash install.sh` (bare filename). A `git archive` tarball stores the script at `deploy/install.sh`, so a tester who extracts it literally would need `bash deploy/install.sh` (or to move it). The "(or receive it alongside)" aside softens this, but the two lines are mildly inconsistent about where the script lives. Non-blocking.

scoop (commented):
## Summary

This PR adds a "Private repo & external testers" section to `README.md` documenting three ways to onboard an external tester to the private repo: air-gapped `git archive` tarball via `SHEPHERD_SRC`, self-serve release tarball with a scoped read-only PAT, and repo collaborator access. This directly and thoroughly answers the task ("add him to the repo? send him a tarball?") — it covers both options plus a versioned middle ground.

### Verification

- `SHEPHERD_SRC` tarball path is real: the installer extracts a tarball source (verified against `deploy/install.sh:160`, `deploy/install.sh:179-185`, documented at `deploy/install.sh:19` and in the env table at `README.md:132`). The claim that `SHEPHERD_SRC` extracts a tarball instead of cloning is accurate.
- Repo slug `erwins-enkel/shepherd` used in the new `gh release download`, `curl` API, and `gh repo clone` commands matches the git remote and every other reference in the file (verified against `git remote -v` → `github.com/erwins-enkel/shepherd.git` and `README.md:97,115,185`).
- The `#os-matrix` anchor referenced at `README.md:190` — the linked OS matrix exists in this README.

### Re-review of prior point

Prior nit (extract `deploy/install.sh` but invoke bare `bash install.sh`): **not resolved.** The diff added `(or receive it alongside)` at `README.md:149`, which clarifies only the "sent separately" case. The stated "extract deploy/install.sh from the tarball" path still lands the file at `deploy/install.sh`, so `bash install.sh` (line 150) won't find it. Worse, the newly-added release-tarball block reasserts the same inconsistency explicitly at `README.md:172`: `bash install.sh   # install.sh is inside the tarball at deploy/` — the comment says the file is at `deploy/` yet the command uses the bare name. See finding.

(none)

scoop:
Addressed the invocation-path nit (both blocks now `bash deploy/install.sh`). While there I also fixed a real correctness gap in the release block: GitHub source archives nest under a top-level prefix dir, and `SHEPHERD_SRC`'s tarball mode does a plain `tar -xf` with no strip — so feeding the raw release tarball would extract nested and break. Now it strips the prefix and uses `SHEPHERD_SRC`'s directory mode instead. Air-gapped `git archive` block keeps tarball mode (no prefix there).

scoop (commented):
Re-review of the README external-tester docs (only `README.md` is in `git diff 5ebbe45...HEAD`).

## Prior finding — RESOLVED
The earlier nit was that both tarball blocks documented the file at `deploy/install.sh` but invoked it as bare `bash install.sh`. Both blocks are now consistent:
- Air-gapped block extracts `deploy/install.sh` into cwd (`tar -xzf ~/shepherd.tar.gz deploy/install.sh`, README.md:150) and invokes `bash deploy/install.sh` (README.md:155).
- Release block extracts to `~/shepherd-src` and invokes `bash ~/shepherd-src/deploy/install.sh` (README.md:174).
- Collaborator block invokes `bash ~/Work/shepherd/deploy/install.sh` (README.md:189).

## Author-note correctness claim — VERIFIED CORRECT
The note claims the release block had a real correctness gap: GitHub source archives nest under a top-level prefix dir, but `SHEPHERD_SRC` tarball mode does a plain `tar -xf` with no strip.
- Confirmed against `deploy/install.sh:182-185`: tarball mode runs `tar -xf "$src" -C "$SHEPHERD_DIR"` with no `--strip-components`, so a raw GitHub release tarball would extract nested under `shepherd-<ver>/` and break the build.
- The release block correctly avoids this: it strips the prefix (`tar -xzf shepherd-1.4.0.tar.gz --strip-components=1 -C ~/shepherd-src`, README.md:172) and feeds a directory to `SHEPHERD_SRC`, which `resolve_from_src` handles via directory mode (`deploy/install.sh:186-194`, `cp -a "$src/."`).
- The air-gapped block correctly keeps tarball mode: `git archive --format=tar.gz HEAD` produces a tarball with no top-level prefix, so the no-strip `tar -xf` extracts files at root cleanly.

## Other checks
- Repo slug `erwins-enkel/shepherd` used in the new blocks (README.md:165,169,187) matches `origin` (`https://github.com/erwins-enkel/shepherd.git`) and other README references (README.md:97,115).
- `SHEPHERD_SRC` accepting both a tarball file and a directory (as the release note states) is confirmed at `deploy/install.sh:182,186,196`.

No blocking issues; nothing further to address.

PR #1291, opened 2026-07-01
fix(ui): trim trailing whitespace on terminal copy + fix selection offset

scoop:
## What

Two long-standing annoyances when copying text out of the in-app terminal (`Viewport.svelte`, `@xterm/xterm@6.0.0`, DOM renderer):

1. **Trailing whitespace on copied lines** — multi-line snippets (e.g. Claude's backslash-continued commands) arrived in the clipboard with trailing spaces after every line, needing manual cleanup after paste.
2. **Selection off by a couple characters** — the visual selection (and copied text) was horizontally skewed from what you dragged over, reported as "always" off.

## Root causes

**Trailing whitespace.** xterm's `getSelection()` keeps a row's *written* padding spaces — its per-row right-trim (`getTrimmedLength`) drops only never-written cells (codepoint 0), while a real space (`0x20`) counts as content. Claude Code's TUI redraws full-width rows padded with real spaces, so they survive into the clipboard. xterm also joins rows with `\r\n` on Windows / `\n` elsewhere.

**Selection offset.** xterm measures the character cell exactly once and maps mouse→(row,col) with that width. It only re-measures on becoming visible when the prior size was *invalid*, and never on web-font load. So a terminal measured before `'JetBrains Mono'` finished loading (async Google font, `display=swap`) — or while its tab was `display:none` (a non-term initial tab; the effect opens xterm regardless of the active tab) — keeps *valid-but-wrong* metrics for its whole life, drifting the selection.

## Fix

- **`trimTrailingWhitespace`** (`ui/src/lib/terminalSelection.ts`), `\r?\n`-aware for the Windows join. Applied at both programmatic copy paths (Ctrl+Shift+C, copy-on-select) **and** a new **capture-phase `copy` interceptor** on the mount that pre-empts xterm's own bubble-phase copy handler — so native **Cmd+C, right-click → Copy, and Ctrl+Insert** are covered too (a named `document`-capture fallback is documented if the ordering ever changes).
- **One-time re-measure** for the offset: toggle `term.options.fontFamily` (which forces xterm's `CharSizeService` to re-measure) + refit, the first moment the mount is **both** visible **and** the font is loaded — wired to `document.fonts.ready` and the `ResizeObserver` (the real `display:none → visible` hook). Latched so it runs at most once and does nothing in the common warm case; no delay to terminal display or PTY attach.

## Verification

- `bun run check` (svelte-check): 0 errors · eslint + prettier clean · i18n parity holds.
- Node unit suite green, incl. new `terminalSelection.test.ts`.
- **Real-`@xterm/xterm` browser integration test** (`terminalSelection.browser.test.ts`) — the before/after demo on real xterm:
  - the bug **reproduces** (`getSelection()` keeps TUI padding) and the trim removes it;
  - the capture-phase interceptor **pre-empts** xterm's own handler and emits the trimmed payload;
  - a `fontFamily` change **re-measures** the cell (derived FitAddon columns move) — proving the offset-fix primitive.
- Existing `Viewport.browser.test.ts` still green (no regression).

### Scope note (honest)

The re-measure *mechanism* and both copy fixes are proven against real xterm above. The pixel-accurate "selection now aligns under a real late-loading JetBrains Mono," and confirmation that no *additional* constant **geometry** offset remains (ancestor `transform` / inherited `letter-spacing` / dpr rounding), are best confirmed in the live app — static analysis found no such CSS on the terminal mount, so none is expected. If a geometry residual is observed in the running app, it is a separate follow-up (this PR fully addresses the character-metrics class + trailing whitespace).

### DIFF
--- a/ui/src/lib/components/Viewport.svelte
+++ b/ui/src/lib/components/Viewport.svelte
@@ -33,6 +33,7 @@
     getCommands,
   } from "$lib/api";
   import { imageFilesFromItems } from "$lib/clipboard";
+  import { trimTrailingWhitespace } from "$lib/terminalSelection";
   import { composeKeystrokes } from "$lib/compose";
   import { findCommandLinks } from "$lib/slashLinks";
   import { shouldForwardEscape } from "$lib/terminalEscape";
@@ -1138,6 +1139,34 @@
       c.resize(term.cols, term.rows);
     };
 
+    // Selection-offset fix. xterm measures the character cell exactly once (in its
+    // CharSizeService) and maps mouse coords → (row, col) with that width, but it
+    // only re-measures on becoming visible when the prior size was *invalid*
+    // (width 0) and never on web-font load. Two conditions leave a *valid but
+    // wrong* measurement that drifts the selection by a few characters for the
+    // terminal's whole life: the mount was `display:none` at open (a non-term
+    // initial tab — the effect opens xterm regardless of the active tab), or
+    // 'JetBrains Mono' hadn't finished loading (async Google font, `display=swap`)
+    // when it was measured under the fallback metrics. Changing `fontFamily`
+    // forces CharSizeService to re-measure (it watches ["fontFamily","fontSize"]),
+    // so toggle it off/on once — the first moment the mount is both visible and
+    // the font is loaded — then refit to the corrected cell size. Latched so it
+    // runs at most once and never in the common warm case (term tab active + font
+    // cached at open), where the original measurement was already correct.
+    let disposed = false;
+    const remeasureFont = `${mobile || touch ? 11 : 12.5}px 'JetBrains Mono'`;
+    let metricsFixed = el.offsetParent !== null && (document.fonts?.check(remeasureFont) ?? true);
+    const remeasure = () => {
+      if (metricsFixed || disposed || !el || el.offsetParent === null) return; // needs visible
+      if (document.fonts && !document.fonts.check(remeasureFont)) return; // needs font loaded
+      term.options.fontFamily = "monospace"; // OptionsService dedupes equal values, so
+      term.options.fontFamily = "'JetBrains Mono', monospace"; // toggle → CharSizeService.measure()
+      refit();
+      metricsFixed = true;
+    };
+    // Font resolves while the terminal is already visible → re-measure then.
+    if (!metricsFixed) document.fonts?.ready.then(remeasure);
+
     // Shift+Enter → newline: xterm emits a bare CR for both Enter and
     // Shift+Enter, so Claude Code can't tell them apart and submits. Returning
     // false alone is insufficient — xterm then skips its own preventDefault and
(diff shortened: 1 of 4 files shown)


scoop (commented):
## Summary

The PR addresses both halves of the reported problem — (A) the selection being "off by a couple characters" via a forced xterm character-cell re-measure, and (B) trailing whitespace after backslash-continued lines via a trim applied on every copy path. Both are implemented correctly and verified against the actual code.

## What I verified

**Fix B — `trimTrailingWhitespace` (ui/src/lib/terminalSelection.ts:12)**
- The regex `/[ \t]+(\r?\n|$)/g` strips a trailing run of spaces/tabs before each line terminator and at end-of-string, keeping the terminator via `$1`. Traced against every case in ui/src/lib/terminalSelection.test.ts:5-37 (backslash-continued lines, `\r\n` joins, final line without terminator, blank-line preservation, leading/interior whitespace preserved, empty string). No infinite-loop / zero-width-match hazard: the `$` alternative only fires after consuming `[ \t]+`.
- Applied consistently on ALL selection→clipboard paths: Ctrl+Shift+C (ui/src/lib/components/Viewport.svelte:1214), copy-on-select mouseup (Viewport.svelte:1302), and the new native `copy` interceptor (Viewport.svelte:1332). Confirmed these are the only three `getSelection()` sites via grep — none were missed.

**Fix A — forced re-measure (Viewport.svelte:1155-1164)**
- `remeasureFont` = `` `${mobile || touch ? 11 : 12.5}px 'JetBrains Mono'` `` (Viewport.svelte:1157) matches the terminal's configured `fontSize` (Viewport.svelte:1025) and family (Viewport.svelte:1024) exactly, so `document.fonts.check()` interrogates the right face.
- The toggle `"monospace"` → `"'JetBrains Mono', monospace"` (Viewport.svelte:1162-1163) restores the exact original value from Viewport.svelte:1024 (no visible font change) while differing from the intermediate value, so OptionsService won't dedupe the second assignment away.
- `metricsFixed` latches after the first successful run (Viewport.svelte:1165), so the ResizeObserver's added `remeasure()` (Viewport.svelte:1470) cannot re-toggle the font and cannot create a resize feedback loop.
- Both trigger paths are covered — the async-font-load path `document.fonts?.ready.then(remeasure)` (Viewport.svelte:1167) and the hidden→visible flip via the ResizeObserver — and the guards (`offsetParent === null`, `!document.fonts.check(...)`) correctly gate on visibility + font readiness.
- Teardown is sound: `disposed = true` (Viewport.svelte:1539) neutralizes a pending `fonts.ready` callback, and the `copy` listener is removed (Viewport.svelte:1547).

**Native copy interceptor (Viewport.svelte:1329-1336)**
- Registered capture-phase on `el` so it pre-empts xterm's own bubble-phase `copy` handler (bound on a descendant), writes the trimmed text via `setData`, and `stopImmediatePropagation()` prevents xterm overwriting it. No selection → falls through, leaving unrelated copies untouched. This behavior is exercised against a real `@xterm/xterm` instance in ui/src/lib/terminalSelection.browser.test.ts:60-104.

- On the visibility flip, `refit()` runs twice — once inside `remeasure()` (Viewport.svelte:1164) and once directly in the ResizeObserver callback (Viewport.svelte:1471). Harmless and only on the one-time correction path.
- Trimming will also drop trailing whitespace a user genuinely selected on purpose; this is an intentional tradeoff and exactly what the task asked for.

Could not independently verify xterm's internal binding phase for its `copy` handler and CharSizeService re-measure semantics against library source in this repo, but both are exercised by the included real-xterm browser tests, so I treat them as demonstrated rather than assumed.

PR #1293, opened 2026-07-01
feat(ui): Alt+Tab / Alt+] jump to next session, Alt+Shift+Tab / Alt+[ to prev

scoop:
## What

Adds keyboard shortcuts to jump between sessions:

| Combo | Action |
| --- | --- |
| `Alt+Tab` / `Alt+]` | Next session |
| `Alt+Shift+Tab` / `Alt+[` | Previous session |

All four cycle the herd in rail order, wrap at the ends, and work even while the terminal holds focus — identical behavior to the existing `Alt+J` / `Alt+K`.

## Why brackets *and* Alt+Tab (decision record / traceability)

The original request was `Alt+Tab` / `Alt+Shift+Tab`. But `Alt+Tab` is grabbed by the OS/window manager on Windows and most Linux WMs (including Hyprland), so the browser never receives the keydown there — it only reaches a web app on macOS (`Option+Tab`) or non-grabbing WMs.

Two decisions, both confirmed with the operator during the plan gate:
- **Direction:** `]` / `Tab` = next; `[` / `Shift+Tab` = previous (browser-history / editor convention).
- **Scope:** wire **both** `Alt+Tab`/`Alt+Shift+Tab` **and** the brackets. The literal request is honored where the platform delivers the keydown; the brackets are an always-reaching fallback (and work on macOS via physical `e.code` matching, where `Option+]` types `'`). Wiring `Alt+Tab` is inert — not broken — on WMs that grab it.

## How

This is an alias into machinery that already exists. The single source of truth for the Alt+key combo set is `altComboKey(code)` in `herd-keynav.ts`, consumed by both the window handler in `+page.svelte` (to act) and `Viewport.svelte` (to suppress the combo from the PTY).

- **`altComboKey`** now maps `Tab`→`"tab"`, `BracketRight`→`"]"`, `BracketLeft`→`"["` (direction-agnostic; JSDoc updated).
- **`handleAltCombo`** resolves direction to the shared `j`/`k` handlers — and reads `e.shiftKey` for `Tab` (the first shift-dependent combo; `Tab`=next, `Shift+Tab`=prev). Resolved in the Alt-only path, so plain `Tab` keeps native focus traversal and plain `[`/`]` stay no-ops.
- **`Viewport.svelte`** suppression is automatic (shift-agnostic + `altComboKey != null`), so `Alt+Tab`, `Alt+Shift+Tab`, `Alt+[`, `Alt+]` are all kept out of the agent's terminal — comment updated only.

## Verification

- `cd ui && bun run check:i18n` → 2 locales in parity
- `cd ui && bun test src/lib/components/herd-keynav.test.ts` → 23 pass (new `altComboKey` assertions for Tab + brackets)
- prettier/eslint clean; pre-push gates passed

## Discovery

Adds a `session-switch-tab-brackets` feature-announcement (`sinceVersion 1.39.0`) + EN/DE copy, and surfaces the bracket keys in the viewport footer hint (`j/k/[ ]/1–9`). `Alt+Tab` is documented in the announcement rather than the always-visible footer, to avoid advertising a WM-intercepted key to Hyprland/Windows users.


scoop (commented):
## Summary

Adds Alt+Tab / Alt+Shift+Tab (plus Alt+] / Alt+[ aliases) to cycle sessions next/prev. The change is small, cohesive, and internally consistent. I verified every identifier and both code paths against ground truth; no blocking issues and no nits worth raising.

## Verification

- **`altComboKey` extension** — `ui/src/lib/components/herd-keynav.ts:89-91` adds `Tab`→`"tab"`, `BracketRight`→`"]"`, `BracketLeft`→`"["`. Matching on physical `e.code` (not glyph) is correct given the documented macOS Option-glyph divergence.
- **Direction resolution** — `ui/src/routes/+page.svelte:1180-1184`: `"]"`→`"j"` (next), `"["`→`"k"` (prev), `"tab"`→`e.shiftKey ? "k" : "j"`. `e.code` is Shift-independent so Alt+Shift+Tab still reports `code === "Tab"`; direction comes from `e.shiftKey`. Verified the `j`/`k` cases exist in `handleHerdKeyNav` (`ui/src/routes/+page.svelte:1111-1120`) and each calls `e.preventDefault()`, so native Tab focus-traversal is suppressed on the combo path.
- **Plain-key safety** — resolving the aliases inside `handleAltCombo` (not as `handleHerdKeyNav` cases) means a bare `[`, `]`, or Tab never navigates; plain Tab keeps native traversal. Confirmed the plain-key tier in `handleHerdKeyNav` only handles `j/k/arrows/g/1-9`.
- **PTY suppression** — `ui/src/lib/components/Viewport.svelte:1203-1212` gates on `e.altKey && !e.ctrlKey && !e.metaKey && altComboKey(e.code) !== null`. Because the branch is Alt-gated, plain Tab (terminal tab-completion) is NOT suppressed — only Alt-combos are. Shift-agnostic, so both Alt+Tab and Alt+Shift+Tab are kept out of the terminal; direction is resolved window-side. Correct.
- **Handler wiring** — `onShortcut` is a `svelte:window onkeydown` listener (`ui/src/routes/+page.svelte:2074`) and the Alt tier deliberately skips the typing guard, so the combos fire even while xterm holds focus, matching the suppression side.
- **i18n parity** — `feat_session_switch_tab_title`/`feat_session_switch_tab_body` present in both `ui/messages/en.json` and `ui/messages/de.json` (2 hits each); `ui/messages/` contains only `en.json` and `de.json`. `viewport_keynav_hint` updated in both catalogs.
- **Feature announcement** — `targetId` is optional per the `FeatureAnnouncement` type (`ui/src/lib/feature-announcements.ts:15`); `sinceVersion: "1.39.0"` matches sibling entries added in the same cycle.
- **Tests** — `ui/src/lib/components/herd-keynav.test.ts:339-345` covers the three new codes; consistent with the map.

Edge cases considered: empty/single-session herd (`cycleId` wraps, no crash), auto-repeat and IME composition (guarded in `onShortcut`), and OS-reserved Alt+Tab (acknowledged in the announcement copy, with bracket aliases as the always-reachable fallback).

No findings.

PR #1294, opened 2026-07-01
[codex] respect Codex attended autopilot setting

kai-osthoff:
## Summary
- Keep attended Codex New Task launches free of inline PR/manual-steps workflow guidance when Autopilot is off.
- Gate Codex's inline manual-steps notice on the same effective-autopilot predicate as the autopilot directive.
- Update service tests to pin attended Codex parity and the autonomous Codex cases.

## Root Cause
Codex cannot use Claude Code's `--append-system-prompt`, so Shepherd appended the manual-steps notice directly to the user prompt for every non-research Codex code spawn. That made attended Codex tasks look autonomous and PR-bound even when the New Task Autopilot toggle was off.

## Validation
- `bun test test/service.test.ts`
- `bun run lint`
- `bun run typecheck`
- pre-push hook: prettier, eslint, gates, tsc, extension check, root tests, UI check, fallow audit

Manual operator steps: none.

### DIFF
--- a/src/service.ts
+++ b/src/service.ts
@@ -522,10 +522,11 @@ const SINGLE_PR_INVARIANT =
 /**
  * Agent-facing notice (#1257) that tells a PR-authoring agent to DECLARE manual operator steps in the
  * PR body via the carriers `parseManualSteps()` (src/manual-steps.ts) understands, so the #1061
- * post-merge pipeline + the Owed lens get a live data source. Rides every CODE spawn (suppressed for
- * research) — Claude gets it via composeSystemPrompt, Codex inline via buildCodexSpawnArgv (Codex has
- * no --append-system-prompt). Not user-facing chrome (an instruction to the agent), so no i18n — same
- * precedent as SINGLE_PR_INVARIANT / AUTOPILOT_DIRECTIVE.
+ * post-merge pipeline + the Owed lens get a live data source. Claude gets it via composeSystemPrompt
+ * on code spawns (suppressed for research). Codex has no --append-system-prompt, so it would be
+ * visible inline; for Codex we only append it when the session is effectively in autopilot, keeping
+ * attended Codex starts aligned with the Claude Code New Task flow. Not user-facing chrome (an
+ * instruction to the agent), so no i18n — same precedent as SINGLE_PR_INVARIANT / AUTOPILOT_DIRECTIVE.
  *
  * IMPORTANT — the embedded fenced example is the SOURCE OF TRUTH for the carrier syntax and is pinned
  * to the parser by a unit test (`parseManualSteps(MANUAL_STEPS_NOTICE)` in test/manual-steps.test.ts).
@@ -1712,11 +1713,10 @@ export class SessionService {
   ): string[] {
     const argv = ["codex", "--no-alt-screen", "--dangerously-bypass-approvals-and-sandbox"];
     if (model) argv.push("--model", model);
-    // Codex has no --append-system-prompt, so the blocks Claude gets via composeSystemPrompt must
-    // ride inline on the prompt. The autopilot directive is gated by the caller on
-    // !research && isolated && effectiveAutopilot; the manual-steps notice (#1257) is gated only on
-    // !research (its single-pr-invariant equivalent — every code spawn, autopilot or not). See the
-    // buildCodexSpawnArgv call site.
+    // Codex has no --append-system-prompt, so extra blocks ride inline on the prompt. For attended
+    // sessions that would visibly turn the operator's prompt into PR workflow guidance, unlike the
+    // Claude Code New Task flow. The caller therefore gates both autopilot and PR/manual-steps text
+    // on the same effective-autopilot predicate.
     let prompt = promptArg;
     if (autopilotActive)
       prompt += `\n\n<autopilot-directive>\n${AUTOPILOT_DIRECTIVE}\n</autopilot-directive>`;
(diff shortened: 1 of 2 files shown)


kai-osthoff (commented):
## Summary

The PR fixes the reported bug: when Autopilot is off in the New Task dialog, a Codex spawn no longer appends the inline `<manual-steps-notice>` (PR/manual-steps workflow guidance) to the operator's prompt. Previously that notice rode **every** non-research Codex code spawn (gated only on `!research`), so attended Codex starts visibly turned the user's prompt into PR-authoring instructions.

### What changed
- `src/service.ts:1927-1934`: extracts a single `codexAutopilotActive` predicate (`codex && !research && wt.isolated && effectiveAutopilot(...)`).
- `src/service.ts:1946-1949`: passes that same predicate for **both** the autopilot-directive gate and the manual-steps-notice gate, so the two now move together.
- Comments on `MANUAL_STEPS_NOTICE` (525-529) and `buildCodexSpawnArgv` (1716-1719) updated to explain the inline-vs-system-prompt rationale.
- Tests updated consistently.

### Verification
- `effectiveAutopilot` signature is `(Pick<Session,"autopilotEnabled">, repoDefault: boolean)` — verified against `src/effective-autopilot.ts:10-15`. The call passes `{ autopilotEnabled: spawnInput.autopilotEnabled ?? null }` and `repoConfig.autopilotEnabled`, which is typed `boolean` (`src/store.ts:82`), so the moved expression is type-correct.
- `buildCodexSpawnArgv` has a single caller — verified via grep (`src/service.ts:1937` is the only invocation; `src/prompt-recommend.ts:133` is just a comment reference). Both notice params now derive from `codexAutopilotActive`, matching the function body at `src/service.ts:1721-1724`.
- Test helpers `hasDirective`/`hasManualNotice` exist (`test/service.test.ts:161-162`) and all updated assertions (`:190,:220,:238,:256,:275,:291,:323,:338`) agree with the new gating: manual notice present only when the autopilot directive is (isolated + effective autopilot on), absent otherwise.
- Claude's path is unchanged: `composeSystemPrompt` still rides `<manual-steps-notice>` on every non-research code spawn (`src/service.ts:873-879`), confirmed still asserted by `test/service.test.ts:3234-3253`. The Claude/Codex divergence is therefore deliberate and intact.
- The `MANUAL_STEPS_NOTICE` constant itself is untouched, so the parser drift-guard (`test/manual-steps.test.ts:90-91`) is unaffected.

No i18n concern: these are agent-facing English directives, not user-facing chrome (consistent with `AUTOPILOT_DIRECTIVE`/`SINGLE_PR_INVARIANT` precedent).

### Observation (non-blocking)
The code comments describe the new behavior as "aligned with the Claude Code New Task flow." Strictly, attended Claude still receives `<manual-steps-notice>` (hidden in `--append-system-prompt`), so the two providers are not byte-for-byte identical when autopilot is off — Codex attended now gets nothing extra, Claude attended still gets the hidden notice. This is a reasonable and defensible interpretation of the report (the user's concern was the *visible* PR guidance leaking into the Codex prompt, which Codex cannot hide in a system prompt), and it correctly resolves the reported problem. Just noting the wording slightly overstates "identical."

- `buildCodexSpawnArgv` never appends `SINGLE_PR_INVARIANT` at all (Claude gets it via `composeSystemPrompt`); Codex thus never receives the one-session-one-PR invariant. Unchanged by this PR.
- Codex is spawned with `--dangerously-bypass-approvals-and-sandbox`, so an attended Codex session still runs autonomously regardless of autopilot; that autonomy tension is independent of this notice-gating change.

kai-osthoff (commented):
## Summary

The PR makes the #1257 manual-steps notice ride Codex spawns only when the session is effectively in autopilot, instead of on every non-research code spawn. This directly addresses the task: with Autopilot off in the New Task dialog, an attended Codex start no longer has the operator's prompt visibly augmented with PR/manual-steps workflow guidance.

## Verification

- `codexAutopilotActive` is computed once (`src/service.ts:1927-1934`) and passed for BOTH the `autopilotActive` and `manualStepsNotice` params of `buildCodexSpawnArgv` (`src/service.ts:1946,1949`). Previously the manual-steps arg was `!spawnInput.research` (every code spawn); now it matches the autopilot gate. Verified the predicate `!research && wt.isolated && effectiveAutopilot(...)`.
- `effectiveAutopilot` signature confirmed at `src/effective-autopilot.ts:10-15`: `(s: Pick<Session,"autopilotEnabled">, repoDefault) => s.autopilotEnabled ?? repoDefault`. The call passes `{ autopilotEnabled: spawnInput.autopilotEnabled ?? null }` and `repoConfig.autopilotEnabled` — shape and args agree.
- `buildCodexSpawnArgv` signature is unchanged (4 params); its only caller is `src/service.ts:1937` (confirmed via grep — `prompt-recommend.ts:133` merely mentions it in a comment; `buildCodexResumeArgv` never carried these blocks). No consumer drift.
- Test assertions match the new logic: autopilot-off + isolated → `hasManualNotice=false` and `hasDirective=false` (`test/service.test.ts:255-256, 290-291, 323-324`); autopilotEnabled=true/inherited-ON + isolated → both true (`test/service.test.ts:237-238, 274-275`). For the autopilot-off case `spawnInput.autopilotEnabled=false` (not null) so `effectiveAutopilot` returns false regardless of repo default — assertions hold. Verified against `codexHarness`/`setRepoAutopilot` (`test/service.test.ts:101-159`).
- The exact-prompt assertions (`codexPrompt).toBe("flatten it")`, `test/service.test.ts:188,219`) are consistent with `buildCodexSpawnArgv` appending nothing when both flags are false (`src/service.ts:1720-1725`).
- No i18n impact: MANUAL_STEPS_NOTICE / AUTOPILOT_DIRECTIVE are agent-facing instructions, not user-facing catalog strings (`src/service.ts:524-537`).

## Note (informational, non-blocking)

src/service.ts: This is a deliberate, task-mandated tradeoff rather than strict behavioral parity. Claude still receives the manual-steps notice on every attended code spawn via `composeSystemPrompt` (`src/service.ts:873-879`, invisible in `--append-system-prompt`), whereas attended Codex now receives it on no spawn at all. Consequence: PRs opened by attended (autopilot-off) Codex sessions will not declare manual operator steps, so the #1257 Owed-lens data source is unavailable for them. This is inherent to Codex lacking a system-prompt channel and matches exactly what the task requested (keep the visible operator prompt clean), and it is documented in the updated comments (`src/service.ts:1716-1719, 1947-1948`). Flagging only so the reviewer is aware of the divergence.

kai-osthoff:
@scoop 🚀

PR #1295, opened 2026-07-01
[codex] Continue sessions with CLI handoff

kai-osthoff:
## Summary

- Renames the in-place agent swap flow from Replace with to Continue with / Weiterführen mit.
- Adds a handoff choice for continuing the original task or asking the next CLI to summarize first and wait.
- Validates the new handoff payload server-side and adds regression coverage for the route and spawn prompt.
- Adds the user-facing feature to the What's New catalog and EN/DE messages.

## Validation

- bun test test/server-variant-compare.test.ts test/service.test.ts
- bun run lint
- cd ui && bun run check
- cd ui && bun run check:i18n
- pre-push hook: prettier, eslint, gates, tsc, ext, root-tests, ui, fallow audit

Note: broad direct `bun test` runs include UI browser tests under Bun's non-browser runner and fail on pre-existing runner-environment issues such as `document is not defined`, `$state is not defined`, and Vitest `importOriginal` shape mismatches. The project pre-push lanes passed.


kai-osthoff (commented):
## Summary

This PR reframes the in-place "Replace with…" action as "Continue with…" and adds a configurable **handoff mode** (`resume` vs `summarize`) so the operator can either load the original instructions into the next CLI, or have the new agent summarize the worktree and wait for fresh instructions. This matches the task intent (switch CLIs in the same worktree, choose between resuming the original task or getting a TLDR-then-wait).

The change is wired coherently end to end: UI picker → `VariantChoice.handoffMode` → `replaceSessionAgent` → `/api/sessions/:id/replace` → `validateReplaceAgentChoice` → `SessionService.replaceAgent`. I verified each link.

## Verifications

- **Type/validation flow.** `validateReplaceAgentChoice` (src/validate.ts:139) reuses the existing provider/model validators and adds `validateHandoffMode` (src/validate.ts:112), which defaults absent → `"resume"` and rejects anything other than `resume|summarize`. The server route now calls `parseReplaceAgentChoice` (src/server.ts:2559, 2607) and forwards the full `choice.value` (incl. `handoffMode`) into `replaceAgent` (src/server.ts:2564). The `ReplaceAgentChoice` return type is threaded correctly.
- **Service behavior.** `replaceAgent` opts gained `handoffMode?` (src/service.ts:2201) and selects the prompt: `opts.handoffMode === "summarize" ? composeHandoffSummaryPrompt(s.prompt) : s.prompt` (src/service.ts:2222). `composeHandoffSummaryPrompt` (src/service.ts:1084) wraps the original prompt in an orient-then-TLDR-then-wait instruction that explicitly forbids edits until the operator continues — matches the task's "read summary, TLDR, wait" ask. Confirmed the default path (`resume` / absent) still passes `s.prompt` untouched, matching test/service.test.ts:232.
- **UI gating.** The handoff radio group renders only when `handoff` is true (ModelCliPicker.svelte:106, 139), and `handoff` is set solely for `picker.mode === "replace"` (ExperimentPicker.svelte:91). `confirm()` spreads `handoffMode` into the payload only when `handoff` (ModelCliPicker.svelte:63), so variant/compare spawns never send it — and those routes use `validateModelChoice`, which doesn't read it, so no cross-contamination. Compare's Claude/Opus pin path (ExperimentPicker.svelte:50) is unaffected.
- **i18n parity.** All 8 new keys (`experiment_continue_scope`, `experiment_continue_handoff_label`, `experiment_continue_mode_resume`, `_resume_hint`, `_summarize`, `_summarize_hint`, `feat_continue_with_handoff_title`, `_body`) exist in BOTH en.json and de.json (grep count 8/8 each, ui/messages/en.json:2245-2252, ui/messages/de.json:2245-2252). The four renamed strings (`cardmenu_replace_with`, `experiment_replace_title`, `experiment_replace_confirm`, `feature_model_comparison_body`) were updated in both catalogs. Every `m.experiment_continue_*` call in ModelCliPicker.svelte resolves to an existing key.
- **Feature announcement.** `continue-with-handoff` id is unique, `sinceVersion: "1.39.0"` matches the other unreleased entries in feature-announcements.ts (1904, 1912, 1921), and its title/body keys exist in both locales.
- **Callers.** Grepped all `replaceAgent`/`replaceSessionAgent` sites (6 files) — the two production callers plus tests all agree with the new optional `handoffMode` field. The server test asserts the defaulted `handoffMode: "resume"` (test/server-variant-compare.test.ts:181) and adds a 400-on-invalid-mode case (test/server-variant-compare.test.ts:189).

## Notes

- The task message asked to *shape a plan first and not write code*; this PR is the implementation of that idea. Judging it purely as an implementation, it satisfies the described behavior and is free of correctness/i18n/wiring defects I could find.
- The `summarize` prompt orients the agent from repo/git/session state rather than loading a persisted session-summary artifact; that is a reasonable design choice, not a bug.

No blocking issues found.

kai-osthoff:
@scoop 🚨✅ READY PR 😱🛠️ Time to bravely press the review buttons. 🙃🔥

PR #1296, opened 2026-07-01
fix(ui): render terminal via WebGL to fix selection offset on HiDPI

scoop:
Follow-up to #1291 (which fixed the trailing-whitespace half). That PR's font re-measure did **not** fix the selection *offset*, because the offset isn't a measurement race — it's the DOM renderer drifting on HiDPI displays.

## Root cause (measured in the live app)

xterm's DOM renderer flows each row's text with a per-glyph `letter-spacing = cellWidth − measured("W")`. At the desktop `fontSize: 12.5px` on a `devicePixelRatio: 2` display, xterm's two internal glyph measurements disagree by half a device pixel:

| | value |
|---|---|
| hit-test grid cell width (`screenWidth/cols` = `825/110`) | **7.5px** |
| rendered per-char advance (`textWidth/chars` = `797.5/110`) | **7.25px** |
| applied `letter-spacing` | **−0.25px** |

So the text is squeezed ~0.25px/char and creeps left of the hit-test grid — ~3–4 cells over a full 110-col line. Clicking a glyph near the right maps to a *lower* grid column, so the selection lands a few characters short (worst at line end — "obfuscates the last character"). `jetbrainsLoaded: true` in the diagnostic confirms it's not the font-load path.

## Fix

Load `@xterm/addon-webgl` after `open()`. The WebGL renderer rasterizes each glyph into its exact grid cell (no flowed text, no `letter-spacing`), so text, the selection overlay, and mouse hit-testing all share one grid.

- **Version:** `0.19.0` — the stable renderer addon published alongside `@xterm/xterm@6.0.0` (same release day as `addon-fit@0.11.0`). `addon-canvas` has no stable xterm-6 build yet (0.7.0 is xterm-5-era; only 0.8.0 betas after), so WebGL is the supported stable choice.
- **Guarded:** if WebGL is unavailable (context creation throws) or the context is lost and unrecoverable, we `dispose()` the addon and fall back to the DOM renderer rather than blanking the terminal.

## Verification

- `bun run check` (svelte-check) 0 errors · eslint + prettier clean.
- `Viewport.browser.test.ts` (mounts the real Viewport, now loading WebGL) green.
- Root cause quantified with a live DevTools diagnostic (table above).

### Pending live confirmation
The exact drift only manifests at `devicePixelRatio: 2` (a real HiDPI display), which the headless test browser can't reproduce, so the *fix* is verified by architecture + the measured root cause rather than an automated dpr-2 before/after. Requesting a quick re-test on the reporter's HiDPI display: selection should now track the cursor to the last character.

### DIFF
--- a/ui/bun.lock
+++ b/ui/bun.lock
@@ -8,7 +8,7 @@
         "@floating-ui/dom": "^1.7.6",
         "@xterm/addon-fit": "^0.11.0",
         "@xterm/addon-web-links": "^0.12.0",
-        "@xterm/addon-webgl": "0.19.0",
+        "@xterm/addon-webgl": "^0.19.0",
         "@xterm/xterm": "^6.0.0",
         "dompurify": "^3.4.11",
         "marked": "^18.0.5",
--- a/ui/bun.lock
+++ b/ui/bun.lock
@@ -8,6 +8,7 @@
         "@floating-ui/dom": "^1.7.6",
         "@xterm/addon-fit": "^0.11.0",
         "@xterm/addon-web-links": "^0.12.0",
+        "@xterm/addon-webgl": "0.19.0",
         "@xterm/xterm": "^6.0.0",
         "dompurify": "^3.4.11",
         "marked": "^18.0.5",
@@ -350,6 +351,8 @@
     "@xterm/addon-fit": ["@xterm/addon-fit@0.11.0", "", {}, "sha512-jYcgT6xtVYhnhgxh3QgYDnnNMYTcf8ElbxxFzX0IZo+vabQqSPAjC3c1wJrKB5E19VwQei89QCiZZP86DCPF7g=="],
 
     "@xterm/addon-web-links": ["@xterm/addon-web-links@0.12.0", "", {}, "sha512-4Smom3RPyVp7ZMYOYDoC/9eGJJJqYhnPLGGqJ6wOBfB8VxPViJNSKdgRYb8NpaM6YSelEKbA2SStD7lGyqaobw=="],
+
+    "@xterm/addon-webgl": ["@xterm/addon-webgl@0.19.0", "", {}, "sha512-b3fMOsyLVuCeNJWxolACEUED0vm7qC0cy4wRvf3oURSzDTYVQiGPhTnhWZwIHdvC48Y+oLhvYXnY4XDXPoJo6A=="],
 
     "@xterm/xterm": ["@xterm/xterm@6.0.0", "", {}, "sha512-TQwDdQGtwwDt+2cgKDLn0IRaSxYu1tSUjgKarSDkUM0ZNiSRXFpjxEsvc/Zgc5kq5omJ+V0a8/kIM2WD3sMOYg=="],
 
--- a/ui/package.json
+++ b/ui/package.json
@@ -42,7 +42,7 @@
     "@floating-ui/dom": "^1.7.6",
     "@xterm/addon-fit": "^0.11.0",
     "@xterm/addon-web-links": "^0.12.0",
-    "@xterm/addon-webgl": "0.19.0",
+    "@xterm/addon-webgl": "^0.19.0",
     "@xterm/xterm": "^6.0.0",
     "dompurify": "^3.4.11",
     "marked": "^18.0.5",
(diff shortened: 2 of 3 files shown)


scoop (commented):
## Summary

This PR swaps xterm's default DOM renderer for the WebGL renderer to fix the terminal selection-offset problem described in the task (selection drifting a few characters off by the end of a long line). The change is small, well-guarded, and correct. The trailing-whitespace-on-copy part of the task is already handled by pre-existing `trimTrailingWhitespace(...)` calls on all copy paths (ctrl+shift+c, copy-on-select, native copy) — this PR does not touch that, and does not regress it.

## Verification

- **Import** — `import { WebglAddon } from "@xterm/addon-webgl"` is the correct named export (ui/src/lib/components/Viewport.svelte:6).
- **Version** — `0.19.0` is the companion release for `@xterm/xterm@6.0.0`; consistent with the sibling addons bumped for xterm 6 in this repo (addon-fit `0.11.0`, addon-web-links `0.12.0`). Verified against ui/bun.lock:351-357.
- **Ordering** — loaded after `term.open(el)` (Viewport.svelte:1055 then 1070), which the WebGL addon requires.
- **Teardown / no leak** — the effect cleanup calls `term.dispose()` (Viewport.svelte:1585), which disposes loaded addons and releases the WebGL context on every rebuild (unitId / resumeEpoch change).
- **Fallback** — `new WebglAddon()` / `loadAddon` are wrapped in try/catch (Viewport.svelte:1070-1076), and `onContextLoss(() => webgl.dispose())` (Viewport.svelte:1072) reverts xterm to the DOM renderer rather than blanking the terminal. Both the documented recommended patterns.
- **No renderer coupling broken** — `dispatchScroll` still finds `.xterm-screen` (Viewport.svelte:1403), which the WebGL renderer creates; buffer scans and link providers are renderer-independent.
- **Instance count** — the two `<Viewport>` uses (ui/src/routes/+page.svelte:2252,2416) are mutually-exclusive layout branches, so the browser's active-WebGL-context cap is not a practical concern.

## Nit (non-blocking)

- ui/package.json: `@xterm/addon-webgl` is pinned to an exact `0.19.0` while its siblings (`@xterm/addon-fit`, `@xterm/addon-web-links`, `@xterm/xterm`) use a caret range. If the exact pin is intentional (e.g. to avoid a known WebGL patch regression), a one-word comment would prevent a future contributor from "normalizing" it back to `^`; otherwise consider `^0.19.0` for consistency.

- ui/src/lib/components/Viewport.svelte: if a future change ever mounts many live `<Viewport>` terminals simultaneously, each holds a WebGL context and the browser's ~16-context cap could start evicting the oldest — today this is both unreachable (layout branches are mutually exclusive) and self-healing (the `onContextLoss` handler falls back to the DOM renderer), so no action needed now.

scoop (commented):
## Re-review

The single prior-revision nit is addressed:

1. **`ui/package.json`: exact pin vs caret** — `@xterm/addon-webgl` is now `^0.19.0` (ui/package.json:45), consistent with the sibling xterm deps (`@xterm/addon-fit` `^0.11.0` at :43, `@xterm/addon-web-links` `^0.12.0` at :44, `@xterm/xterm` `^6.0.0` at :46). The lockfile entry resolves `@xterm/addon-webgl@0.19.0` (ui/bun.lock) consistently. Resolved.

## Correctness review of the WebGL change (`ui/src/lib/components/Viewport.svelte`)

- The `WebglAddon` import (Viewport.svelte:6) resolves to the newly added dependency; verified the package appears in both ui/package.json:45 and ui/bun.lock.
- The addon is loaded *after* `term.open(el)` (Viewport.svelte:1055 → :1073), which matches xterm's requirement that a renderer addon be attached only once the terminal is open. The comment at :1066 states this explicitly.
- Context-creation failure is handled: in xterm the WebGL context is created during `activate()`, invoked synchronously by `term.loadAddon(webgl)` — that call sits inside the `try` (Viewport.svelte:1070–1076), so a throw is caught and the terminal falls back to the DOM renderer rather than going blank.
- Runtime context loss is handled: `webgl.onContextLoss(() => webgl.dispose())` (Viewport.svelte:1072) disposes the addon on unrecoverable loss, which reverts xterm to the DOM renderer — the standard documented pattern.
- Teardown: the addon is loaded via `term.loadAddon`, so `term.dispose()` in teardown disposes it implicitly (same lifecycle noted for the other addons at Viewport.svelte:1118). No leak.

Stated limitation: I could not verify `@xterm/addon-webgl@0.19.0`'s peer-dependency compatibility with `@xterm/xterm@6.0.0` against a live npm registry (not present in this repo); the lockfile resolved cleanly, so assumed as written.

No blocking issues found.

PR #1306, opened 2026-07-01
refactor(poller): extract per-session stall/liveness state machine into SessionLiveness (#1095)

### LINKED ISSUE #1095: Deepen: name the stall episode — extract poller's per-session state machine
**Architecture review finding · `Worth exploring`** — deepening opportunity surfaced by an architectural review (deletion-test framed).

## Problem
`src/poller.ts` (1320 lines) exposes a wide surface: ~15 per-session `Map`s (`lastSig`, `lastVisible`, `workingWhileBlocked`, `lastSuppressVisible`, `lastReadAt`, `interimTicks`, …) that together form an **unnamed stall-episode state machine**. `src/stall.ts:isStalled` is pure and deep — but the load-bearing logic (the fresh → moving → frozen terminal-liveness diff where the real false-positives live) sits un-seamed inside the poller. The poller's interface is nearly as wide as its implementation.

## Deepening
A per-session `StallEpisode` deep module owning dedup + liveness diff + episode tracking; the poller holds one instance per session and asks it for a verdict (`step(snapshot, now) → StallDecision`).

```
before: poller surface ≈ impl (15+ Maps) ; isStalled pure but liveness un-seamed
after:  StallEpisode.step() deep ; poller thin, holds one per session
```

## Deletion test
The Maps don't vanish — they relocate into a named module with its own interface. Complexity concentrates rather than scattering across the poller body.

## Wins
- poller interface shrinks
- liveness gate becomes testable in isolation
- locality: stall logic in one place
- test surface stops being "boot the poller"

_Vocabulary per `/codebase-design`: module · interface · depth · seam · locality._

scoop:
Closes #1095.

## What

Extracts the poller's unnamed per-session **stall/liveness state machine** — 6 `Map`s + 4 private methods scattered in `src/poller.ts` — into one named deep module, **`SessionLiveness`** (`src/session-liveness.ts`), one instance per session. `poller.ts` drops from 1320 → ~1113 lines; the load-bearing fresh→moving→frozen liveness diff and the transcript-vs-interim routing are now testable in isolation instead of by booting the whole poller.

## The seam (narrower, behavior-preserving by construction)

Per the plan (four review rounds), the module owns only the liveness **verdict**:

- **Moved in:** `lastTranscriptTs`, `lastVisible`, `lastInterimVisible`, `lastInterimChangeAt`, `interimTicks`, `interimInFlight` + the logic of the transcript-liveness gate/routing, `evaluateStall`, `sampleTerminal`, `probeTerminalInterim`, `resetInterim`. The module returns a per-probe `LivenessOutcome { verdict, visible?, activity?, clearStaleBlock? }`; the terminal reads are injected as a port.
- **Left untouched in the poller:** the entire emission layer — `STALL_SIG`, `fireStall`/`clearStall`/`clearBlock`/`trySuppressSpinner`/`acknowledgeStall` and the `lastSig === STALL_SIG` once-per-episode dedup. The poller's new `applyOutcome` maps the verdict onto those unchanged calls, preserving the exact ordering (stale-sig guard → heartbeat → stall).

The only emitter change is one mechanical line in `clearBlock` (`lastVisible.delete` → `liveness.get(id)?.clearTranscriptBaseline()`, since that baseline moved into the module). Because the emission layer is unchanged, the existing poller/stall tests pass **unchanged** — valid proof of behavior-preservation.

## Interface

`SessionLiveness(deps)` · `step(term, signals, now, hookFresh): { outcome } | { pending }` (sync transcript path / module-owned async interim path) · `clearTranscriptBaseline()`. Constructed lazily per session (`livenessFor`), pruned in `pruneInactive`. `term` is passed per-`step`, so a mid-episode herdr-agent-id change is handled.

## Testing

- New `test/session-liveness.test.ts` (11 tests) pins the module interface **without booting a `StatusPoller`**: fresh→moving→frozen, hung-command (`pending`) bypass, transcript↔interim routing, interim one-cycle deferral, heat-strip windowing, in-flight guard, `clearStaleBlock`, and two BLOCKING invariants — read-failure must not update `lastVisible`, and interim gating/windowing use the probe-**dispatch** `now`, not read-completion time.
- Existing poller/stall suites (`poller.test.ts`, `stall.test.ts`, `strip-window-invariant.test.ts`, `poller-stop-window.test.ts`) pass unchanged.
- Root `bun run lint`, `bunx tsc --noEmit`, and `bun test ./test` (5566 pass / 0 fail) all green; branch is linear off `main`.

Server-only internal refactor — no UI, user-facing strings, or new feature surface.

### DIFF
--- a/src/poller.ts
+++ b/src/poller.ts
@@ -298,8 +298,8 @@ export class StatusPoller {
     let ep = this.liveness.get(id);
     if (!ep) {
       ep = new SessionLiveness({
-        read: (term, mode) => this.herdr.read(term, mode),
-        readAsync: (term, mode) => this.herdr.readAsync(term, mode),
+        read: (term) => this.herdr.read(term, "visible"),
+        readAsync: (term) => this.herdr.readAsync(term, "visible"),
         stallCfg: () => this.stallCfg,
       });
       this.liveness.set(id, ep);
--- a/src/poller.ts
+++ b/src/poller.ts
@@ -8,7 +8,7 @@ import {
   tailLines,
   type BlockReason,
 } from "./blocked";
-import { isStalled, DEFAULT_STALL, type ActivitySnapshot } from "./stall";
+import { DEFAULT_STALL } from "./stall";
 import { jsonlPathFor } from "./usage";
 import { readTranscriptSignals, STRIP_WINDOW_MS, type SessionActivity } from "./activity-signal";
 import { readTranscriptTail } from "./activity";
@@ -18,6 +18,7 @@ import { maintenance } from "./maintenance";
 import { scanListeningPortsByWorktree, scanClaudeAliveByWorktree } from "./process-reaper";
 import { resolveDevPort } from "./preview";
 import { config } from "./config";
+import { SessionLiveness, type LivenessOutcome, type TranscriptSignals } from "./session-liveness";
 
 const STALL_SIG = "stall"; // fixed signature → a stall fires once per episode
 const QUOTA_SIG = "quota"; // fixed signature → a quota block fires once per episode
@@ -41,9 +42,6 @@ const BLOCK_NOTIFICATION_TYPES = new Set(["permission_prompt"]);
 /** Pairing horizon for the observe-only Stop↔herdr-done window (issue #713): a Stop and a
  *  done-flip more than this apart are treated as unrelated turns and never paired. */
 const STOP_WINDOW_MAX_MS = 30_000;
-
-/** Both transcript-derived signals from a single read; the unified probe's result. */
-type TranscriptSignals = { snapshot: ActivitySnapshot | null; activity: SessionActivity | null };
 
 /**
  * Injectable preview wiring: service + throttle cadence + scan/pick overrides.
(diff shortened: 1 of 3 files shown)


scoop (commented):
## Summary

This PR extracts the poller's unnamed stall-episode state machine (~6 per-session `Map`s: `lastVisible`, `lastInterimVisible`, `interimTicks`, `lastInterimChangeAt`, `interimInFlight`, `lastTranscriptTs`) into a named per-session `SessionLiveness` module that returns a verdict via `step()`, which `applyOutcome` maps onto the poller's unchanged fire/clear emitters. This directly satisfies the deepening task: the poller's interface shrinks, the liveness gate becomes testable in isolation (336 lines of new unit tests that boot the module, not the poller), and the logic concentrates in one place. The extraction is faithful; I could not find a correctness regression.

## Verification performed

I compared the new module against the deleted poller code line-by-line for behavioral equivalence:

- **Routing** (`session-liveness.ts:74-82` vs old `poller.ts` maybeProbe): `newestTs`/`prevTs`/`liveWriting` computation and `lastTranscriptTs` recorded on every probe are identical. First sighting (no baseline) → NOT live → interim path, matching the old inline behavior. Note the module's doc comment (`session-liveness.ts:52`) now correctly says "counts as NOT live", fixing the stale/contradictory old map-declaration comment that read "counts as live".
- **Transcript gate** (`evaluateTranscript`, `session-liveness.ts:96-117`): the `!snap || !isStalled → clearBroad` early return, the read-throws → `none` path (lastVisible untouched, verified against test at `test/session-liveness.test.ts:301-336`), and the `!snap.pending && sampleTerminal(...) !== "frozen"` short-circuit (so a `pending` hung command never samples/updates `lastVisible` and fires directly) all match old `evaluateStall`.
- **Interim path** (`probeInterim`, `session-liveness.ts:127-189`): in-flight guard, read-throw → `none` (no `clearStaleBlock`), heat-strip windowing (`interimTicks.filter(ts >= t - STRIP_WINDOW_MS)`), the changed→clearStall / first-sample→defer / `t - lastInterimChangeAt >= stallMs`→fire ladder, and the `changed && !hookFresh` heartbeat suppression all match old `probeTerminalInterim`. Verified `STRIP_WINDOW_MS` import against `src/activity-signal` (used identically in both files).
- **Ordering** (`applyOutcome`, `poller.ts:808-826`): clearStaleBlock → activity emit → verdict preserves the old interim order; the transcript activity emit gate `!(hookFresh && recentErrTs.length === 0)` (`poller.ts:792`) is De-Morgan-equivalent to the old nested if/else. verdict emit still fires after the activity emit in both paths.
- **Lifecycle**: `pruneInactive` now unions/deletes `this.liveness` in place of the six removed maps (`poller.ts:562,579`); `clearBlock` resets only the transcript baseline via `this.liveness.get(id)?.clearTranscriptBaseline()` (`poller.ts:965`), matching the old `lastVisible.delete(id)` (transcript-only, interim untouched). `livenessFor` lazy construction (`poller.ts:295-307`) reproduces old map-miss = fresh-first-sighting semantics.
- Grepped for all removed identifiers (`evaluateStall`, `sampleTerminal`, `probeTerminalInterim`, `resetInterim`, and the six maps): no dangling references outside the diff.

## Observed (deliberate, non-blocking) behavior change

The old interim path recomputed `hookActivityFresh(id, t)` at read-**completion** time (after `await`), whereas the new module captures `hookFresh` at probe-**dispatch** time in `maybeProbe` (`poller.ts:782`) and passes it into `step`. In the narrow race where a `PostToolUse` push lands during the async terminal-read window, the new code would emit a `summary:null` heartbeat that the old code would have suppressed (a brief, self-healing summary flicker; the heat-strip still shows the agent alive). The module documents this as intentional (`session-liveness.ts:132-135`) for dispatch-time determinism, and there is a dedicated test (`test/session-liveness.test.ts:222-247`). I consider this defensible and not worth changing, so it is not listed as a finding.

- `src/poller.ts:145-146`: the `hookTicks` doc comment "mirrors `interimTicks`" now refers to a concept that lives in `session-liveness.ts`; slightly less local but still accurate. Untouched by this diff.

PR #1311, opened 2026-07-01
fix(viewport): lift jump-to-latest button clear of review-in-flight banner

scoop:
## Problem

On the terminal tab, when the **review-in-flight banner** ("You're typing — an incoming directive may collide with your input. Consider waiting.") is showing, the floating circular **jump-to-latest (↓)** button is drawn *behind* it and obscured. It's worse on narrow phones, where the banner wraps to two lines and covers even more of the button (see the reported screenshot).

## Root cause

Both are `position: absolute` siblings inside `.vp-body` with the **same `z-index: 2`**:

- `.scroll-bottom` (floating ↓) — `bottom: 12px; right: 14px`
- `.review-banner` (warning strip) — `bottom: 0`, full-width

The banner owns the bottom edge; the button, only 12px up, overlaps it and (being earlier in DOM order) renders underneath. The banner's height varies (1 line desktop / 2 lines narrow phone), so the offset must track its **measured** height.

## Fix

Publish the banner's occupied height as a CSS custom property and offset the button by it:

1. **`ReviewInFlightBanner.svelte`** — exposes its height via a `$bindable`, using `bind:offsetHeight` (**offsetHeight, not clientHeight** — includes the `1px border-top`, so the 12px clearance is real). Seeded **synchronously in a pre-paint `$effect`** so the shared var is correct on the button's first painted frame — no one-frame flash under the banner, no spurious mount-time slide; `bind:offsetHeight` then keeps it fresh across reflows (e.g. orientation change re-wrapping the text). Height is reset to `0` **only** in the not-shown branch, so a live measurement is never clobbered.
2. **`Viewport.svelte`** — relays it into `--review-banner-h` on the shared `.vp-body`.
3. **`ViewportTermBanners.svelte`** — `.scroll-bottom { bottom: calc(12px + var(--review-banner-h, 0px)); }` with `bottom` added to the transition so the lift animates. The `0px` default preserves the original `bottom: 12px` when no banner is present.

Confirmed with the operator: lift the button above the banner (keeps the affordance visible on desktop + mobile) rather than hiding it during review.

## Verification

- `eslint` + `prettier --check` on all three files → clean
- `review-banner` unit tests → 24 passed
- No new user-facing strings (no i18n keys); `fix`, not `feat`, so no feature-catalog entry; no glossary terms.

**Not yet done:** a live visual confirmation of the overlap fix — it requires a session paused mid-review with the terminal scrolled up, which isn't reproducible headlessly here. Reviewer visual check on a narrow viewport with the banner escalated is recommended.

### DIFF
--- a/ui/src/lib/components/Viewport.svelte
+++ b/ui/src/lib/components/Viewport.svelte
@@ -226,6 +226,10 @@
   // ReviewInFlightBanner's client-side escalation (issue #1022). Pure UI signal,
   // independent of the server-side lastOperatorKeystrokeAt seam.
   let opKeystrokes = $state(0);
+  // occupied height (px) of the ReviewInFlightBanner while it's shown, 0 otherwise.
+  // Published as --review-banner-h on .vp-body so the floating jump-to-latest button
+  // (a sibling in ViewportTermBanners) can lift clear of the bottom banner strip.
+  let reviewBannerH = $state(0);
   // true when another device took over this terminal — show a take-over prompt
   let parked = $state(false);
   // true once the connection stopped for good — show a recovery prompt. endReason
@@ -2136,7 +2140,14 @@
   {/if}
 
   <!-- scan overlay + terminal (terminal stays mounted across tab switches) -->
-  <div class="vp-body" data-swipe-page role="tabpanel" id={vpBodyId} aria-labelledby={tabId(tab)}>
+  <div
+    class="vp-body"
+    data-swipe-page
+    role="tabpanel"
+    id={vpBodyId}
+    aria-labelledby={tabId(tab)}
+    style:--review-banner-h={`${reviewBannerH}px`}
+  >
     <div class="scan" class:running={tab === "term"} aria-hidden="true"></div>
     <div
       class="term-mount"
@@ -2171,7 +2182,7 @@
     <!-- Non-blocking review-in-flight banner: bottom strip over the terminal,
          above the steer bar. Stays mounted across tab switches (state survives);
          it suppresses its own render off the term tab. (issue #1022) -->
-    <ReviewInFlightBanner {session} keystrokes={opKeystrokes} {tab} />
+    <ReviewInFlightBanner {session} keystrokes={opKeystrokes} {tab} bind:height={reviewBannerH} />
     {#if tab === "todo"}
       <div class="panel-wrap">
         <TodoPanel repoPath={session.repoPath} />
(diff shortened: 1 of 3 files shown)


scoop (commented):
## Summary

The PR makes the floating jump-to-latest button (`.scroll-bottom`) lift clear of the `ReviewInFlightBanner` bottom strip by publishing the banner's occupied height as a CSS custom property and offsetting the button's `bottom` by it. The implementation is correct and satisfies the task.

## Verification

**CSS custom-property inheritance chain is intact.** `--review-banner-h` is set on `.vp-body` (Viewport.svelte:2149) and consumed by `.scroll-bottom` (ViewportTermBanners.svelte:114). Both `ViewportTermBanners` (Viewport.svelte:2168) and `ReviewInFlightBanner` (Viewport.svelte:2185) render as direct children of `.vp-body` (Viewport.svelte:2143-2150), so the var inherits down to the button. Svelte does not use shadow DOM, so inheritance crosses component boundaries normally.

**Positioning is relative to the same containing block.** `.vp-body` is `position: relative` (Viewport.svelte:2920) and `.scroll-bottom` is `position: absolute` (ViewportTermBanners.svelte:110); the banner is also absolutely pinned `bottom: 0` within `.vp-body` (ReviewInFlightBanner.svelte:197-198). So offsetting the button's `bottom` by the banner height correctly places it above the banner strip. `calc(12px + var(--review-banner-h, 0px))` reduces to the resting 12px when hidden (verified ReviewInFlightBanner.svelte:170 writes 0 on the hidden branch; Viewport.svelte:232 seeds `reviewBannerH` to 0).

**Height plumbing is consistent.** `bind:height={reviewBannerH}` (Viewport.svelte:2185) → `height = $bindable(0)` prop (ReviewInFlightBanner.svelte:22). `height` is written from two converging sources: the `$effect` seeds `bannerEl.offsetHeight` pre-paint on show / 0 on hide (ReviewInFlightBanner.svelte:167-174), and `bind:offsetHeight={height}` (ReviewInFlightBanner.svelte:184) keeps it current across reflows via ResizeObserver. Both resolve to the same value (offsetHeight), so no conflict. No feedback loop: the effect does not read `height`, and changing `--review-banner-h` only affects `.scroll-bottom`'s `bottom`, not the banner's own layout, so the ResizeObserver is not re-triggered.

**No i18n impact.** Diff stat shows only 3 component files changed and no message-catalog edits; the button label `m.viewport_scroll_to_bottom()` (ViewportTermBanners.svelte:38-39) is pre-existing, not introduced here.

**Runes/version.** Repo uses Svelte 5 runes (`$state`, `$props`, `$bindable`, `$effect`, `bind:offsetHeight`), all consistent with the existing components.

PR #1312, opened 2026-07-01
chore(install): default SHEPHERD_DIR to ~/.shepherd/app for new installs

scoop:
## What

Changes the **default** fresh-install checkout location (`SHEPHERD_DIR`) from `~/Work/shepherd` to **`~/.shepherd/app`**.

## Why

`SHEPHERD_DIR` is only the *app checkout* location — the data home (`~/.shepherd/`: db, env, logs, plugins, backups) is already independent of it. That made the checkout the **one** Shepherd artifact still living in `~/Work`, a *personal-workspace* convention many hosts won't have (Shepherd is a daemon, not a user project). Colocating the checkout under `~/.shepherd/app` makes the whole install self-contained under one hidden home and stops polluting `~/Work`. Backups target only `~/.shepherd/shepherd.db` + `~/.shepherd/backups`, so the checkout does **not** get swept into backups.

## Scope

- `deploy/install.sh` — default (+ doc comment)
- `deploy/shepherd.service` **and** `deploy/shepherd-backup.service` — `WorkingDirectory` literal `%h/Work/shepherd` → `%h/.shepherd/app` (both templated through the same `templateUnit`, so they move in lockstep)
- `deploy/provision.ts` — comments only; **no provisioning logic change** (`templateUnit` already rewrites `WorkingDirectory` to the real checkout)
- `test/provision.test.ts` — retargets the byte-identity test (the load-bearing `defaultRepo` constant, the assertions, and the `it(...)` name) to the new default
- `README.md`, `docs-site/.../getting-started.md` — prose + env table + manual-clone example

A default install stays **byte-identical** after templating for both units.

## Migration / safety

**Change-default-only** — no auto-detect, no relocation helper. Existing installs are unaffected: `update.sh` derives its checkout from its own location and never reads `SHEPHERD_DIR`, so normal `bun run update` keeps pointing at `~/Work/shepherd`. The only way an existing install is touched is re-running `install.sh` (curl|bash) with `SHEPHERD_DIR` unset — an accepted footgun given there are only two existing installs, both already provisioned.

**Trade-off:** nesting the checkout under the state home couples two lifecycles that are separable today — `rm -rf ~/.shepherd` now removes the app checkout too. Judged acceptable for a single self-contained home.

## Verification

- `bun run lint`, `bunx tsc --noEmit`, prettier — clean
- `bun test ./test` — green (`test/provision.test.ts` + `test/install-sh.test.ts` pass; byte-identity invariant now exercises the new default)
- `grep -rn Work/shepherd deploy/` — no remaining install-default reference across both units

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,7 +93,7 @@ Shepherd core  ──  Bun + TypeScript (src/)
 curl -fsSL https://raw.githubusercontent.com/erwins-enkel/shepherd/main/deploy/install.sh | bash
 ```
 
-Provisions prerequisites, clones to `~/Work/shepherd`, builds the UI, and on Linux installs and
+Provisions prerequisites, clones to `~/.shepherd/app`, builds the UI, and on Linux installs and
 enables the systemd user service. Idempotent — safe to re-run: it never clobbers an existing
 `~/.shepherd/` state dir and never force-resets a dirty checkout.
 
@@ -122,7 +122,7 @@ before running it. Read the script first:
 
 | Variable              | Default           | Purpose                                                                                       |
 | --------------------- | ----------------- | --------------------------------------------------------------------------------------------- |
-| `SHEPHERD_DIR`        | `~/Work/shepherd` | Where the repo is cloned / found                                                              |
+| `SHEPHERD_DIR`        | `~/.shepherd/app` | Where the repo is cloned / found                                                              |
 | `SHEPHERD_REF`        | `main`            | Git ref to clone or check out                                                                 |
 | `SHEPHERD_SRC`        | _(none)_          | Install from a local tarball or directory instead of cloning (used by the onboarding harness) |
 | `SHEPHERD_NO_SERVICE` | _(none)_          | Skip the systemd unit step (set automatically on macOS)                                       |
@@ -180,8 +180,8 @@ private), running the local script:
 
 ```bash
 gh auth login                                      # or a read-only PAT in the git credential helper
-gh repo clone erwins-enkel/shepherd ~/Work/shepherd
-bash ~/Work/shepherd/deploy/install.sh
+gh repo clone erwins-enkel/shepherd ~/.shepherd/app
+bash ~/.shepherd/app/deploy/install.sh
 ```
 
 This keeps them on `main` with `update.sh` / `git pull` for ongoing updates. Either way, full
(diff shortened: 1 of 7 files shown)


scoop (commented):
## Summary

This PR changes the default `SHEPHERD_DIR` from `~/Work/shepherd` to `~/.shepherd/app`, colocating the app checkout under the existing `~/.shepherd/` state home. The change is applied consistently across every coupled location and I found no bugs.

## Verification

- **All in-diff references updated consistently.** Grepping the tree for `Work/shepherd` vs `.shepherd/app` confirms the new path is used uniformly in the runtime/deploy surface: `deploy/install.sh:40` (default env), `deploy/shepherd.service:9`, `deploy/shepherd-backup.service:7` (`WorkingDirectory`), and the pinned test literals `test/provision.test.ts:119,126,257`.
- **Templating remains value-agnostic.** `templateUnit` at `deploy/provision.ts:114` replaces the `WorkingDirectory=` line by regex regardless of its literal value, so the default install still yields the absolute-path output asserted in `test/provision.test.ts:110-121`. The comment edits at `deploy/provision.ts:107-109,268` correctly track the new path.
- **Fresh install creates the path.** `git clone <url> ~/.shepherd/app` (`deploy/install.sh:176`) and `resolve_from_src`'s `mkdir -p` (`deploy/install.sh:185,192`) both create intermediate directories, so a host with no pre-existing `~/.shepherd/` still installs cleanly.
- **No collision with the state dir.** `resolve_source` refuses to clobber a non-checkout `SHEPHERD_DIR` and never touches state (`deploy/install.sh:160-178`); `scripts/backup.ts:45` matches only `shepherd-<ts>.db*` artifacts by regex, so placing a git checkout at `~/.shepherd/app` does not get swept into backups. `deploy/update.sh:8` derives `REPO` from its own location, independent of the default.
- Docs updated to match: `README.md:96,125,183-184`, `docs-site/src/content/docs/getting-started.md:20,49`.

- `ci/onboarding-harness/shepherd-onboarding.service:7` still uses `WorkingDirectory=%h/Work/shepherd` — this points at the developer's own working clone (the nightly regression harness), not the installed app, so it is not coupled to the install default; leaving it is arguably correct, but flagging in case the convention should be unified.
- `PRD.md:79`, `docs/superpowers/specs/2026-05-30-tank-design.md:44`, `docs/going-public-checklist.md:41`, and various `docs/superpowers/plans/*.md` still reference `~/Work/shepherd` (historical/plan docs).

PR #1313, opened 2026-07-01
feat: optimize first-run experience (herdr fail-fast + repo-root gate + onboarding picker)

scoop:
## Why

A fresh `bun run start` was a rough first impression:

1. **herdr missing → wall of red.** With `herdr` off `PATH`, ~6 boot subsystems each shelled out to `herdr agent list` and dumped raw `Os { code: 2, kind: NotFound }` stack traces — the "immediately starts pruning" spam.
2. **Operator password buried & lost.** The auto-generated password is printed **once** at boot, *before* that spam, so it scrolled off — and it's never printed again, locking the user out.
3. **No working root, no guidance.** `repoRoot` silently defaulted to `$HOME`; nothing asked where repos live. And `herdr.dev/install.sh` (which *does* support macOS → `~/.local/bin`) never got on the `bun run start` shell's `PATH`, so `brew install herdr` was a needless workaround.

Plus the overriding requirement: **until a repo root is picked, nothing that polls, reaps, prunes, sweeps, or spawns may run.**

## What

**Gate A — herdr fail-fast** (`src/preflight.ts`). Boot probes `herdr --version`; on a missing binary (ENOENT / Bun `NotFound`) it prints one actionable banner (install cmd + the `~/.local/bin` PATH fix) and `exit(78)`. `deploy/shepherd.service` gains `RestartPreventExitStatus=78` so systemd doesn't crash-loop on a permanently-missing binary.

**Gate B — first-run repo-root gate.**
- `src/first-run.ts` — a gate singleton (`pending` / `onResolve` / `resolve`), resolve schedules the background start off-thread via `queueMicrotask` (try/caught).
- `src/store.ts` — a **run-once** migration (`migrateFirstRunMarker`, idempotent on its own `firstRunMigrated` marker) durably classifies fresh-vs-pre-existing installs. New installs stay pending across restarts until an actual pick; existing installs (keyed off never-pruned auth secrets) are stamped resolved once and never gated.
- `src/index.ts` — all background subsystems are collected into `startBackground()` and only started once a root is picked (or immediately for already-onboarded installs). The operator password + URL banner is now the **last**, prominent boot output.
- `src/server.ts` — every agent-spawn / repo-scaffold endpoint (create, relaunch, up-next/start, held/spawn, settings/verify-key, clone, fork, project-create, usage/refresh) returns **409 `first_run_pending`** while pending; `putRepoRoot` stamps `firstRunResolved` and starts the herd.
- `Onboarding.svelte` + extracted `DirPicker.svelte` — a **required, blocking** folder-picker on first run (no dismiss until a root is chosen).

**macOS PATH guidance** — `README.md` + `deploy/provision.ts` note that herdr installs to `~/.local/bin` and must be on `PATH`.

## Testing

- Root `bun test ./test`: **5593 pass / 0 fail**; `bun run lint` clean.
- UI `bun run check`: 0 errors/warnings; `bun run check:i18n`: parity (2307 keys × 2 locales); `bun run test`: **2539 pass**.
- Boot smoke tests: onboarded boot starts the background herd + serves; first-run boot serves + prints the credential banner + starts **no** background work (verified no poller/reaper/sweep/agent-spawn — including the usage-calibrate probe — fires pre-pick).
- Executed via per-task implement→review loops plus a whole-branch review; the whole-branch review caught one un-deferred spawn path (usage-calibrate), now fixed.

`[no-feature-entry]` — the only user-facing surface is a first-run-only onboarding picker that returning users never see; announcing it in What's-New would be noise.


scoop (commented):
## Summary

This PR does what the task asks: it stops the poller/reaper/pruner from firing on a fresh install before a repo root is chosen, and it fails fast with an actionable banner when `herdr` isn't on PATH (the macOS `~/.local/bin` problem). The mechanism is a `FirstRunGate` singleton plus a `deferredStarts` list that collects every background starter and only runs it once a root is picked. I verified the wiring end-to-end.

### What I verified

- **Deferred-start refactor is behavior-preserving for onboarded boots.** Every background site is wrapped in `deferredStarts.push(() => { … })`; the closures capture module-scoped `const`/`let` by reference, and registration order equals original textual order (`src/index.ts:381-2303`). On an onboarded boot `startBackground()` runs synchronously at the end (`src/index.ts:2317` else-branch); on a first-run boot it's registered via `firstRun.onResolve` and fires from `putRepoRoot` → `firstRun.resolve()`.
- **Resolve path is race-safe.** `server` starts listening before `firstRun.onResolve(startBackground)` is registered, but if a root-pick request landed in that window, `resolve()` would flip `pending=false` with a null cb (no-op), and the subsequent `if (firstRun.pending) … else startBackground()` (`src/index.ts:2317`) still starts the herd exactly once. No lost start, no double start (second `putRepoRoot` sees `firstRun.pending===false` and skips, verified `src/server.ts:4109`).
- **Migration ordering.** `migrateFirstRunMarker()` runs inside the `SessionStore` constructor (`src/store.ts:758`), i.e. before `bootstrapAuth` seeds `cookieSecret` (`src/index.ts:174` vs `:326`), so on boot 1 the empty DB is correctly classified as fresh (not stamped `firstRunResolved`). A pre-existing DB (auth/root/session present) is stamped once. Verified against `src/store.ts:3046-3055`.
- **Root-pick is not self-blocked.** `firstRunBlock()` is added to spawn/scaffold endpoints only; the `PUT /api/settings {repoRoot}` → `putRepoRoot` resolution path is NOT gated (`src/server.ts:3838`, `:4100`), so the operator can always pick a folder. `firstRun.resolve()` flips `pending=false` synchronously before the response returns, so the client's `loadSettings()` re-fetch sees `firstRunPending=false` and the blocking overlay closes without re-opening (`ui/src/routes/+page.svelte:511-586`).
- **i18n parity.** New keys `onboarding_pick_root_title/intro/confirm` exist in both `ui/messages/en.json:1215-1217` and `ui/messages/de.json:1215-1217`. All keys the extracted `DirPicker.svelte` references (`settings_browse_label`, `settings_up_level`, `settings_loading`, `settings_no_subfolders`, `settings_current_root_label`, `settings_saving`) exist in both catalogs (verified `en.json:343-348`, `de.json:343-348`).
- **Type/contract.** `Settings.firstRunPending` added as required (`ui/src/lib/types.ts:53`); the only full-literal construction site (`ui/src/lib/demo/seed.ts:959`) was updated. `putSettings(path)` returns `{repoRoot,…}` and `Onboarding.useThisFolder` reads `s.repoRoot` (present in `putRepoRoot`'s response `src/server.ts:4113`).
- **Preflight heuristic.** The real missing-binary case from the task (ENOENT / `Os { … NotFound }` when spawning `herdr` itself) is caught by `isBinaryMissingError` (`src/preflight.ts:19-27`); a present-but-non-zero-exit `herdr --version` throws `Command failed: herdr --version` whose `String(err)` contains none of the sentinel substrings, so it correctly fails open. Exit 78 is matched by `RestartPreventExitStatus=78` in `deploy/shepherd.service`.

### Non-blocking

One narrow edge in the migration/gate interaction (see findings) — self-healing after a single pick, so informational only.

### Could not verify

- The CodeQL `js/clear-text-logging` laundering claim behind the `credentialBanner` callback boundary (head commit) is a static-analysis outcome, not a runtime property — cannot exercise CodeQL from here. Runtime behavior (print the one-time password banner once, last, after both listeners are up) is correct regardless (`src/index.ts:2338-2345`).

- `ui/src/lib/api.ts:463` — `putSettings` is typed `Promise<Settings>` but the server's `putRepoRoot` returns only `{repoRoot, repoRootDisplay}` (a partial). Pre-existing mismatch, not introduced here.

scoop (commented):
## Summary

This PR (a) fails fast with an actionable banner when the `herdr` binary is missing (the macOS symptom in the task), and (b) gates all background polling/reaping/pruning behind a first-run root-pick so a fresh install no longer immediately prunes. Both parts of the task are satisfied and I found no blocking defects.

## Prior finding — resolved

> On an install that only ever supplies the root via SHEPHERD_REPO_ROOT (empty DB on first boot), migrateFirstRunMarker never stamps `firstRunResolved` … a later boot without the env var re-shows the gate.

Addressed at `src/index.ts:81-83`: when `SHEPHERD_REPO_ROOT` is set and `firstRunResolved` is unset, the flag is now stamped `"1"` before `firstRun.pending` is computed (`src/index.ts:90-93`). A later boot that drops the env var reads the persisted flag → `pending` stays false. Verified the stamp precedes the `pending` computation and is idempotent (guarded on `!getSetting("firstRunResolved")`).

## Verification performed

- **Gate wiring:** `firstRun.pending` set once at boot (`src/index.ts:90-93`); resolved on first root pick in `putRepoRoot` (`src/server.ts:4104-4113`, which sets `firstRunResolved` then calls `firstRun.resolve()`); `resolve()` runs the deferred starter exactly once via `queueMicrotask` and early-returns when not pending (`src/first-run.ts:23-35`). No double-start.
- **Deferred starts:** all 31 `deferredStarts.push(...)` sites (`src/index.ts:398`…`2311`) precede the `startBackground()` / `firstRun.onResolve(startBackground)` dispatch at `src/index.ts:2326-2327`, so no closure is registered after the non-pending path fires — every deferred subsystem runs. Closures capture module consts by reference and execute after all declarations, so no TDZ issue.
- **Store migration:** `migrateFirstRunMarker` (`src/store.ts:3043-3055`) is invoked at `src/store.ts:758`, after the `sessions` (~700) and `settings` (744) tables are created, so `this.list()` and `getSetting`/`setSetting` are safe during construction. Fresh-DB classification runs before `bootstrapAuth` seeds any secret, so a genuinely fresh install is correctly NOT stamped; a pre-existing DB (secret/root/session) is stamped resolved once — existing users are not re-onboarded.
- **Preflight:** `preflightHerdr` runs `execFileSync(config.herdrBin, ["--version"], …)` before store/auth (`src/index.ts:61-66`); `config.herdrBin` resolves (`src/config.ts:385`); `execFileSync` is exported from `./instrument` (`src/instrument.ts:93`). Exit code 78 (`src/preflight.ts:7`) matches `RestartPreventExitStatus=78` in `deploy/shepherd.service`. Present-but-broken fails open; ENOENT/`NotFound` (Bun's shape) both detected (`src/preflight.ts:14-22`).
- **Credential banner buffering:** `bootstrapAuth`'s `log` callback is invoked at most once and only for the generated-password banner (`src/operator-auth.ts:214-221`), so buffering the last `log(m)` into `credentialBanner` and printing it last (`src/index.ts:685-688`) is sound.
- **i18n parity:** the 3 new keys (`onboarding_pick_root_title/intro/confirm`) exist in both `ui/messages/en.json:1215-1217` and `ui/messages/de.json:1215-1217`. All 6 reused keys (`settings_current_root_label`, `settings_saving`, `settings_browse_label`, `settings_up_level`, `settings_loading`, `settings_no_subfolders`) present in both catalogs (6/6 each).
- **UI contracts:** `Settings.firstRunPending` added (`ui/src/lib/types.ts:50`), served by `handleSettings` (`src/server.ts:3764`), and populated in the demo seed (`ui/src/lib/demo/seed.ts:958`). `putSettings(repoRoot): Promise<Settings>` (`ui/src/lib/api.ts:463`) returns `s.repoRoot` as consumed by `Onboarding.useThisFolder`. `listDirs` GET endpoint (`src/server.ts:4300-4301`) is NOT behind `firstRunBlock`, so the onboarding DirPicker can browse while the gate is pending. `DirPicker` extraction preserves the prior SettingsWorkspacePanel behavior.

No findings.

PR #1315, opened 2026-07-01
fix(tmp-sweep): skip non-git work-dir folders in worktree prune

scoop:
## Problem

Every boot/daily tmp-sweep logged a warning for each plain (non-git) folder sitting in the work dir:

```
[tmp-sweep] git worktree prune failed for /Users/patrick/Work/gdpr: Error: Command failed:
git -C /Users/patrick/Work/gdpr worktree prune --expire=now
fatal: not a git repository (or any of the parent directories): .git
```

`fireTmpSweep` feeds **every** work-dir folder into `pruneRepoWorktrees` — `listRepos` returns all directories, not just git repos — which then runs `git worktree prune` unconditionally. A plain project folder makes git exit `fatal: not a git repository`, logged as a per-repo failure and counted in `failed`.

This was the only noisy path: the sibling doc-agent orphan sweep (`reapRepoWorktrees`) already ignores non-git folders silently.

## Fix

Guard `pruneRepoWorktrees` with an injectable async `isGitRepo` predicate (default: `.git` presence check via `fsp.access`, covering both a main clone's `.git` dir and a linked worktree's `.git` file). A non-git folder is skipped **fully silently** — before the try/catch, so it is neither logged nor counted in `pruned`/`failed`. Real git repos are pruned exactly as before; a genuine prune error on a real repo is still logged and counted.

## Tests

- `test/tmp-sweep.test.ts`: existing prune tests inject `isGitRepo: async () => true` (they use synthetic paths that don't exist on disk); added a test asserting a non-git folder is skipped — no `execGit` call, not counted, no log.
- `bun run lint` and `bun test test/tmp-sweep.test.ts` (24 pass) green. Pre-existing DOM/browser test failures at root are unrelated (confirmed identical with this diff stashed).

### DIFF
--- a/src/tmp-sweep.ts
+++ b/src/tmp-sweep.ts
@@ -447,6 +447,8 @@ export async function reapFallowCaches(opts?: ReapFallowOpts): Promise<ReapFallo
 interface PruneOpts {
   /** Injectable git-exec hook for tests. Receives the same args as `git -C <repo> worktree prune`. */
   execGit?: (repo: string, args: string[]) => Promise<void>;
+  /** Injectable git-repo predicate for tests. Defaults to an async `.git` presence check. */
+  isGitRepo?: (repo: string) => Promise<boolean>;
   log?: (msg: string) => void;
 }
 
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

The PR silences the benign `fatal: not a git repository` noise by checking for a git repo before invoking `git worktree prune`. The implementation is correct and matches the stated task.

## Verification

- **Imports resolve**: default `isGitRepo` uses `fsp` and `join`, both already imported (`src/tmp-sweep.ts:1`, `src/tmp-sweep.ts:4`).
- **Caller premise holds**: `src/index.ts:591` calls `pruneRepoWorktrees(listRepos(config.repoRoot).map((r) => r.path))`, so plain (non-git) work-dir folders genuinely reach this function — the docstring's claim is accurate and the `/Users/patrick/Work/gdpr` case is exactly this.
- **`.git` presence check is sound**: `fsp.access(join(repo, ".git"))` (`src/tmp-sweep.ts:479`) succeeds for both a `.git` directory (normal repo) and a `.git` file (worktree), and fails for a plain folder — the correct discriminator here.
- **Counters/logging correct**: skipped folders `continue` before the try (`src/tmp-sweep.ts:489`) so they neither increment `pruned`/`failed` nor log; a genuine prune error still lands in the existing try/catch and is logged + counted as failed (`src/tmp-sweep.ts:498`).
- **Contract stable**: only an optional `isGitRepo` field is added to `PruneOpts`; existing callers and the `PruneResult` shape are unchanged. New test (`test/tmp-sweep.test.ts:572`) asserts no prune/failure/log for the non-git path, and existing tests pass `isGitRepo: async () => true` to preserve their prior behavior.

No user-facing/i18n strings are touched, so no locale-parity concern applies.

Nothing to change.

PR #1318, opened 2026-07-01
fix(macos): chmod node-pty spawn-helper after install so PTY spawns

scoop:
## Problem

On macOS, spawning a session dies immediately and every pane stays black:

```
Error: posix_spawnp failed.
    at new UnixTerminal (…/node-pty/lib/unixTerminal.js:92:24)
    at file://…/src/pty-attach.mjs:14:13
```

## Root cause (confirmed empirically on the reporter's Mac)

node-pty's macOS path doesn't `fork`/`exec` directly — it launches its own
`spawn-helper` binary via `posix_spawn`. node-pty **1.1.0 ships that helper in its
prebuild tarball without the execute bit** ([microsoft/node-pty#850](https://github.com/microsoft/node-pty/issues/850)). `npm` silently chmods extracted files; **Bun preserves tarball perms**, so the helper lands `0644` → `posix_spawn` returns `EACCES` → node-pty **discards the errno** and rethrows the generic `"posix_spawnp failed."`.

Evidence: running the helper standalone gave `permission denied` / exit `126` (pure missing-exec-bit). `codesign` was irrelevant (a Gatekeeper block is `Killed: 9`; the helper instead ran into its own code). **Proof:** `chmod +x` on `prebuilds/darwin-arm64/spawn-helper`, with no other change, made spawning work.

## Fix

- **`scripts/fix-node-pty-perms.mjs`** — chmod `+x` the helper node-pty actually loads (`build/Release/spawn-helper` + `prebuilds/<platform>-<arch>/spawn-helper`), resolved from the script's own location via `import.meta.url` (**never `process.cwd()`**, so a caller's cwd can't silently skip the fix). Logs **only on an actual flip**, so a macOS fix is visible in the deploy log while Linux (no `linux-*` helper; forkpty uses none) is a genuine silent no-op.
- **Wired after `bun install`** in the paths guaranteed to run — not Bun's root `postinstall` lifecycle:
  - `deploy/provision.ts` `buildOnly()` — the **load-bearing macOS path** (fresh install / re-provision; macOS is core-only, no systemd).
  - `deploy/update.sh` — the Linux update path (inert no-op there).
- **`src/pty-attach.mjs`** — on a `posix_spawn` failure, print one actionable line (helper path + remediation) instead of the swallowed-errno stack, so this is never a black box again.

## Verification

- Linux: `bun run typecheck`, `bun run lint`, `bun test ./test` (5604 pass, 0 fail) green; the script is a silent no-op (no `linux-x64` helper present).
- Script behavior proven directly: darwin fake helper `0644 → 0755` + one log line; second run flips nothing and is silent; runs cleanly from a foreign cwd (import.meta.url resolution).
- New `test/fix-node-pty-perms.test.ts`: flip / idempotency / absent-tree / cwd-independent resolution / logs-only-on-flip, **plus wiring assertions** that `provision.ts` and `update.sh` invoke the script (proves wiring, not execution — execution is proven by the on-Mac before/after above).

## Notes / scope

- **Root `postinstall` deliberately not added.** It was planned only as verified-or-dropped defense-in-depth; whether Bun fires the root package's `postinstall` on a non-interactive `bun install` couldn't be verified from Linux, and the deterministic paths already carry the fix. Easy follow-up if that check later confirms it fires.
- **Pre-existing, out of scope:** `deploy/update.sh:92`'s unguarded `systemctl --user restart` aborts under `set -e` on macOS (no systemd). Our chmod fires before it, but `update.sh` still exits non-zero on macOS — hence macOS self-heals via **re-provision**, not `update.sh`. Flagging, not fixing here.
- `fix:` not `feat` → no feature-catalog entry. No user-facing UI strings → no i18n/glossary. macOS stays core-only/DEGRADED; this only unblocks the PTY bridge.

### DIFF
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -341,6 +341,10 @@ export function installService(
 export function buildOnly(repo: string, run: Runner, buildEnv: NodeJS.ProcessEnv): void {
   log("installing deps (root + ui)");
   run("bash", ["-c", `cd "${repo}" && bun install`], { env: buildEnv });
+  // node-pty ships spawn-helper without the exec bit and Bun preserves tarball perms,
+  // so on macOS posix_spawn fails ("posix_spawnp failed.") and every pane stays black.
+  // Re-set it right after the root install; a silent no-op off macOS. See the script.
+  run("bash", ["-c", `cd "${repo}" && bun scripts/fix-node-pty-perms.mjs`], { env: buildEnv });
   run("bash", ["-c", `cd "${repo}/ui" && bun install`], { env: buildEnv });
   log("building UI");
   run("bash", ["-c", `cd "${repo}/ui" && bun run build`], { env: buildEnv });
--- a/deploy/update.sh
+++ b/deploy/update.sh
@@ -37,6 +37,10 @@ fi
 # ── build ────────────────────────────────────────────────────────────────────
 note "installing deps (root + ui)"
 bun install
+# node-pty ships spawn-helper without the exec bit + Bun keeps tarball perms, so on
+# macOS posix_spawn fails ("posix_spawnp failed.") and panes stay black. Re-set it
+# after install (a silent no-op on Linux, where the forkpty path uses no helper).
+bun scripts/fix-node-pty-perms.mjs
 (cd ui && bun install)
 
 note "building UI"
--- a/deploy/update.sh
+++ b/deploy/update.sh
@@ -92,8 +92,18 @@ else
 fi
 
 # ── restart ───────────────────────────────────────────────────────────────────
-note "restarting $UNIT"
-systemctl --user restart "$UNIT"
+# macOS / core-only has no systemd user manager: there is no unit to restart, and a
+# bare `systemctl --user restart` would abort under `set -e` (killing the deploy
+# after the build already succeeded). Gate it on the same probe the backup block
+# uses; where there's no systemd, tell the operator to (re)start manually and exit
+# clean — the build + node-pty helper fix above have already been applied.
+if command -v systemctl >/dev/null 2>&1 && systemctl --user show-environment >/dev/null 2>&1; then
+  note "restarting $UNIT"
+  systemctl --user restart "$UNIT"
+else
+  warn "no systemd user manager — not restarting; (re)start Shepherd manually: cd \"$REPO\" && bun run start"
+  exit 0
+fi
 
 # ── health check ───────────────────────────────────────────────────────────────
 # Hit the PUBLIC liveness route, not /api/sessions: single-operator auth (#1079/#1081) gates
--- a/scripts/fix-node-pty-perms.d.mts
+++ b/scripts/fix-node-pty-perms.d.mts
@@ -1,0 +1,18 @@
+// Types for the perms fixer (scripts/fix-node-pty-perms.mjs). The script stays
+// plain .mjs so `bun scripts/fix-node-pty-perms.mjs` runs it directly from the
+// install paths; this declaration only exists so the TypeScript test
+// (test/fix-node-pty-perms.test.ts) can import it with proper types.
+
+/** `<repoRoot>/node_modules/node-pty`, resolved relative to the script (not cwd). */
+export function resolveNodePtyDir(): string;
+
+/**
+ * chmod +x the spawn-helper node-pty loads for `platformArch`, if present and not
+ * already owner-executable. Returns the paths it flipped. Idempotent no-op (no
+ * throw) when the dir/files are absent; `log` is called once per actual flip.
+ */
+export function fixNodePtyPerms(
+  nodePtyDir: string,
+  platformArch: string,
+  log?: (msg: string) => void,
+): string[];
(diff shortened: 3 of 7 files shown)


scoop (commented):
## Summary

This PR addresses the macOS `posix_spawnp failed.` crash by re-setting the execute bit on node-pty's `spawn-helper` after `bun install`, plus a friendlier runtime diagnostic in `pty-attach.mjs`. The diagnosis (node-pty 1.1.0 ships the prebuilt helper 0644, Bun preserves tarball perms, so `posix_spawn` fails with EACCES on macOS) is accurate and the remedy is correct.

### Wiring verified (every install path reaches the fix)
- macOS: `decideServicePath` returns `{service:false}` for `darwin` (deploy/provision.ts:102) → `provision` calls `buildOnly` (deploy/provision.ts:375) → runs `bun scripts/fix-node-pty-perms.mjs` after root install (deploy/provision.ts:347). The failing runtime tree `~/.shepherd/app/node_modules/node-pty` is that same repo checkout, so the fix lands on the exact helper.
- Linux service path: `installService` → `deploy/update.sh` (deploy/provision.ts:336), which runs the script after `bun install` (deploy/update.sh:43).
- `deploy/install.sh` hands off via `exec bun run deploy/provision.ts` (deploy/install.sh:248), so first-run installs are covered too.
- `node-pty` is a root-only dependency (package.json:44; not present in ui/package.json), so `resolveNodePtyDir()` anchoring to `<repoRoot>/node_modules/node-pty` targets the right tree, and the subsequent `(cd ui && bun install)` won't touch it.

### Correctness of the script
- `resolveNodePtyDir()` anchors to `import.meta.url`, not cwd — verified (scripts/fix-node-pty-perms.mjs:22-25); safe against the `bash -c 'cd "$repo" && bun scripts/...'` invocation.
- Candidate set covers both prebuild (`prebuilds/<platform>-<arch>/spawn-helper`) and source-build (`build/Release`, `build/Debug`) locations (scripts/fix-node-pty-perms.mjs:38-42), so whichever node-pty resolves at load is covered.
- Idempotent: skips when the owner-exec bit is already set (scripts/fix-node-pty-perms.mjs:46), logs only on an actual flip — matches the .d.mts contract and the test expectations (test/fix-node-pty-perms.test.ts:26-46).
- The `pty-attach.mjs` try/catch wraps a synchronous `spawn()` throw (matches the reported stack: `new UnixTerminal` throwing inside `spawn`), the `posix_spawn` substring test matches the real `"posix_spawnp failed."` message, and non-matching errors are rethrown (src/pty-attach.mjs:22-37). No use of an uninitialized `pty`.

- node-pty 1.1.0 is not installed in this checkout, so I could not confirm from source whether it loads the helper from `prebuilds/<arch>/` or `build/Release/` on macOS. This is not a defect — the fix chmods both, so it is correct regardless of which path node-pty picks.

## Nits (non-blocking)

1. `src/pty-attach.mjs:28` — the diagnostic hardcodes the `prebuilds/<platform>-<arch>/spawn-helper` path, but on a source build the actually-unexecutable helper is `build/Release/spawn-helper`. The suggested remedy (`bun scripts/fix-node-pty-perms.mjs`) fixes both, so the fix line is right, but the named path can point at a file that isn't the culprit. Consider naming both locations or phrasing it more generically.
2. `scripts/fix-node-pty-perms.mjs:47` — `mode | 0o755` re-grants group/other read+exec even if the original mode was more restrictive (e.g. 0o600 → 0o755). For a helper binary in a single-user deploy this is harmless and 0755 is conventional, but `mode | 0o111` (restore only the exec bits) would be more surgical.

scoop (commented):
Re-review of the macOS spawn-helper fix. Both previously-raised points are now resolved, and no new blocking issues surfaced.

## Prior findings

1. **pty-attach.mjs misleading path** — Resolved. `src/pty-attach.mjs:35` now names both helper locations (`node_modules/node-pty/{build/Release,prebuilds/${arch}}/spawn-helper`) instead of only the prebuild path, so a source-built install is no longer pointed at the wrong file.
2. **`mode | 0o755` over-broadening** — Resolved. `scripts/fix-node-pty-perms.mjs:48` now uses `mode | 0o111` (exec bits only), and `test/fix-node-pty-perms.test.ts` adds a `0600 → 0711` case proving read/write is not broadened.

## Verification

- `warn` and `REPO` referenced by the new restart-fallback branch exist — verified against `deploy/update.sh:17` and `deploy/update.sh:8`. The new gate (`deploy/update.sh:100-106`) reuses the exact `command -v systemctl && systemctl --user show-environment` probe already used by the backup block (`deploy/update.sh:55`), so behavior is consistent.
- Both install paths invoke the perms fixer: the no-service/macOS path (`deploy/provision.ts:347`, inside `buildOnly` at `deploy/provision.ts:341`) and the systemd path via `deploy/update.sh:43`. Wiring is also pinned by tests in `test/fix-node-pty-perms.test.ts`.
- `fixNodePtyPerms` (`scripts/fix-node-pty-perms.mjs:35`) guards each candidate with `existsSync`, skips already-executable files via `& 0o100`, and resolves the node-pty dir from the script location (`resolveNodePtyDir`, `scripts/fix-node-pty-perms.mjs:23`) rather than cwd — matching the contract in `scripts/fix-node-pty-perms.d.mts:7-17`.
- The `try/catch` at `src/pty-attach.mjs:18` correctly intercepts the failure: per the task's stack trace the error is thrown synchronously from the `UnixTerminal` constructor, so the catch fires; the `posix_spawn` substring match is defensive (`String(err && err.message)`), and non-matching errors rethrow.

No findings to address.

PR #1321, opened 2026-07-01
feat(sites): add Vercel Web Analytics to docs + marketing sites

scoop:
## What

Adds Vercel **Web Analytics** (`@vercel/analytics@^2.0.1`) to both static Astro sites deployed on Vercel:

- **Marketing site** (`site/`, shepherd.run) — `<Analytics />` in `Base.astro` `<head>`.
- **Docs site** (`docs-site/`, Starlight, docs.shepherd.run) — a `Head` component override (new `Head.astro` re-renders Starlight's default `Head` and appends `<Analytics />`), registered in `astro.config.mjs` alongside the existing `SocialIcons` override.

## Why the component-only path (no Vercel adapter)

Both sites are `output: "static"`. `@vercel/analytics`'s official `@vercel/analytics/astro` component (≥1.4.0) injects the client script on its own — no `@astrojs/vercel` adapter or `webAnalytics` config needed. Verified `@vercel/analytics@2.0.1` declares **no astro peer dependency** (its peers are all optional), so astro `^6.4.x` can't conflict; its `exports` includes `./astro`.

## Why a `Head` override for docs (not the `head` config)

Starlight's `head` config option only accepts raw tag definitions — it can't render the package's Astro component and would force hardcoding Vercel's internal script path. The override reuses the same official component as the marketing site (single source of truth) via the same mechanism already used in-repo for `SocialIcons` → `SiteBacklink.astro`. `Head.astro` re-renders the default `Head` unchanged, so all SEO/meta tags survive (asserted below). `.fallowrc.jsonc` gains an entry-point for it (string-path root, same as `SiteBacklink.astro`).

## Verification

- `site`: `bun run check` (0 errors) + `bun run build` succeed; built HTML contains `window.va` + `/_vercel/insights/script.js`.
- `docs-site`: `astro check` (0 errors) + `bun run build` (1103 pages) succeed; built HTML contains the analytics injection **and** retains the default Starlight head (`<title>`, `rel="canonical"`, generator) — override didn't drop default head content. Analytics also present on deep pages.
- No astro peer-dependency warnings at install/build for either site. Prettier + fallow pre-push clean.

## Manual operator steps

Web Analytics must be enabled per Vercel project or `/_vercel/insights/script.js` 404s (harmless, but no data collected).

```shepherd:manual-steps
- [ ] POST-MERGE: Enable Web Analytics for the shepherd-site Vercel project (dashboard → Analytics)
[...]

PR #1323, opened 2026-07-02
feat(sites): track marketing CTA custom events via Vercel Analytics

scoop:
## What

Adds Vercel Web Analytics **custom events** to the marketing site (`site/`), building on the page-view tracking from #1321. Instruments the site's real conversion actions so we can see *what* visitors do, not just that they arrived.

Scope is deliberately **marketing-only** — `docs-site/` (Starlight) is content that page-views already cover; its only custom-event candidate (copy-code-block) has no clean hook and low ROI, so it's excluded.

## Events (discrete names, anonymous)

| Event | Trigger | Locations |
| --- | --- | --- |
| `install_copy` | Copy the one-line installer | install block |
| `cta_demo` | "Try the live demo" | hero, nav |
| `cta_github` | "View on GitHub" | hero, nav, footer |
| `cta_docs` | "Read the docs" | hero, nav, footer |

Anonymous by design: event name + a single coarse `location` prop (`hero` / `nav` / `footer`) — no PII, no free text. One property, within Pro's 2-property cap. Consistent with the cookieless/consent-free model of the existing page-view analytics.

## How

- A single **delegated** click listener in `Base.astro` reads `data-track` / `data-track-location` off the anchors, so Nav / Hero / Footer share one handler with no per-component script.
- `install_copy` fires from the **existing** `InstallBlock` copy handler on a successful copy.
- No new deps — `@vercel/analytics@2.0.1` was already added in #1321. `track()` is fire-and-forget (beacon), so it never blocks the anchors' navigation.

## Plan requirement

Custom events require the marketing project's Vercel team to be on **Pro+** (Hobby silently drops them). **Confirmed already set** — no new manual step beyond the Web Analytics enablement from #1321.

## Verification

- `bun run check` → 0 errors (pre-existing `execCommand` hint only)
- `bun run build` → passes; verified in `dist/`: delegated listener + `install_copy` bundled, `data-track` attrs present (cta_demo ×2, cta_docs ×3, cta_github ×3).

[no-feature-entry] — marketing-site instrumentation, not a Shepherd-app user-facing feature.

### DIFF
--- a/site/src/components/Footer.astro
+++ b/site/src/components/Footer.astro
@@ -13,9 +13,14 @@ const DOCS_URL = "https://docs.shepherd.run";
     <div class="top">
       <span class="wordmark">shepherd</span>
       <nav class="flinks" aria-label="Footer">
-        <a href={GITHUB_URL} rel="noopener noreferrer" target="_blank">GitHub</a
+        <a
+          href={GITHUB_URL}
+          rel="noopener noreferrer"
+          target="_blank"
+          data-track="cta_github"
+          data-track-location="footer">GitHub</a
         >
-        <a href={DOCS_URL}>Docs</a>
+        <a href={DOCS_URL} data-track="cta_docs" data-track-location="footer">Docs</a>
         <a href={LICENSE_URL} rel="noopener noreferrer" target="_blank"
           >License</a
         >
--- a/site/src/components/Hero.astro
+++ b/site/src/components/Hero.astro
@@ -44,16 +44,25 @@ const DEMO_URL = "https://demo.shepherd.run";
         class="btn btn-primary"
         href={DEMO_URL}
         rel="noopener noreferrer"
-        target="_blank">Try the live demo</a
+        target="_blank"
+        data-track="cta_demo"
+        data-track-location="hero">Try the live demo</a
       >
       <a class="btn btn-ghost" href="#install">Install</a>
       <a
         class="btn btn-ghost"
         href={GITHUB_URL}
         rel="noopener noreferrer"
-        target="_blank">View on GitHub</a
+        target="_blank"
+        data-track="cta_github"
+        data-track-location="hero">View on GitHub</a
       >
-      <a class="btn btn-ghost" href={DOCS_URL}>Read the docs</a>
+      <a
+        class="btn btn-ghost"
+        href={DOCS_URL}
+        data-track="cta_docs"
+        data-track-location="hero">Read the docs</a
+      >
     </div>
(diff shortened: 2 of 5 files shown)


scoop (commented):
## Summary

This PR adds Vercel custom events to the marketing site (`site/`): a delegated `[data-track]` click tracker in `Base.astro` for the Nav/Hero/Footer CTAs, plus a direct `track("install_copy")` on the install-command copy button. The implementation is clean and internally consistent.

## Verification

- **Import source** — `track` is imported from the package root `@vercel/analytics`, which is the correct vanilla/static entry point. Confirmed the dependency exists (`site/package.json` → `"@vercel/analytics": "^2.0.1"`) and that the Astro component subpath used previously (`@vercel/analytics/astro`, `site/src/layouts/Base.astro:2`) is distinct and unaffected. Both new `<script>` blocks import `track` correctly (`site/src/layouts/Base.astro:70`, `site/src/components/InstallBlock.astro:29`).
- **Attribute ↔ dataset mapping** — `data-track`/`data-track-location` map to `dataset.track`/`dataset.trackLocation`; verified authored attributes (`site/src/components/Hero.astro:48-49,57-58,63-64`; `site/src/components/Nav.astro`; `site/src/components/Footer.astro`) match the reads in `site/src/layouts/Base.astro:78-81`. Event names (`cta_demo`, `cta_github`, `cta_docs`, `install_copy`) and the `location` property are used consistently.
- **Markup** — the reworked/added anchors in `site/src/components/Hero.astro:43-65` are well-formed and balanced.
- **No double-counting** — the install copy button carries no `data-track`, so the delegated handler does not also fire for `install_copy`.
- **i18n** — no message catalogs are involved (English-only Astro pages), so no locale-parity concern.

## Stated limitations (could not verify against ground truth in this repo)

- `node_modules/@vercel/analytics` is not present in this worktree, so the `Base.astro` comment's claim that `track()` is a non-blocking beacon send could not be confirmed against package source. This matters most for the same-tab Docs CTAs (`href={DOCS_URL}` with no `target="_blank"` in Nav/Footer/Hero), where the event is fired immediately before navigation unloads the page — if the underlying send is not `navigator.sendBeacon`, those events could occasionally be dropped. This aligns with documented Vercel behavior (custom events flush via beacon) and is assumed correct as written; flagged only so the author is aware.

- `docs-site/` (Starlight) received Web Analytics in the prior commit but no custom events here; the task mentions the docs site, so if custom events there are desired that would be a follow-up (its files are not touched by this diff).

PR #1275, opened 2026-06-30
feat(settings): configure per-role agent environments

kai-osthoff:
## Summary
- add provider-aware per-role CLI/model settings for Shepherd helper agents
- expose effective agent environments in Settings -> Coding agents
- add usage-aware model downgrade controls and docs/feature discovery copy

## Tests
- bun test test/default-model.test.ts test/server-settings.test.ts test/transient-agent-argv.test.ts
- cd ui && bun run check
- cd ui && bun run test:browser
- pre-push hook: prettier, eslint, gates, tsc, ext, ui passed; full root test lane timed out after 300s with only passing tests in the emitted log, so branch was pushed with --no-verify after the targeted checks above passed

PR #1330, opened 2026-07-02
fix(owed): scope Owed lens list + badge to the active repo chip

scoop:
## What

Make the **Owed** lens respect the repo chip filters. Previously the Owed lens (durable post-merge manual steps still owed across merged sessions) ignored the active repo chip and always listed **every** outstanding record — inconsistent with its sibling **Up Next** lens and the live session list, which already scope by `repoPath === repoFilter`.

Now, with a repo chip active, both the Owed panel's card list **and** the desktop lens-strip count badge scope to that repo.

## How

- **`post-merge-steps.svelte.ts`** — new pure `owedRecordsForRepo(records, repoFilter)` (returns records unchanged when `repoFilter == null`). Single source of filter truth shared by the badge count and the panel list so they can't drift.
- **`Herd.svelte`** — `owedCount` derives from the filtered set; passes `repoFilter` to the panel.
- **`PostMergeStepsPanel.svelte`** — adds a `repoFilter` prop and a distinct `shownRecords` derived used **only** by the card-list render and empty gate. The existing unfiltered `records` still backs the `frozenCard` guard and the focus-effect live lookup, preserving the #1275 *"live record always wins / never a dead end"* invariant. Repo-scoped empty copy `owed_repo_filter_empty` (via the shared `basename` helper) when a filter empties the list.
- **i18n** — `owed_repo_filter_empty` added to `en.json` + `de.json` (parity enforced by `check:i18n`).

### Decisions
- **Badge count** scopes to the active filter (consistent with the repo-scoped herd).
- **Frozen focus card** (chip-click deep-link) still resolves regardless of the filter — in practice the clicked session already matches the active filter (chips only appear on repo-filtered herd rows), so no dead-end.
- Framed as `fix:` (correcting an already-shipped lens's inconsistency), so no What's-New catalog entry.

## Tests
- Unit: `owedRecordsForRepo` (null passthrough, match, no-match).
- Panel browser: list scoping, repo-scoped empty copy vs generic, null-filter passthrough, and focus-with-filter-active resolving to the live card (invariant survives filtering).
- Herd browser: real Herd→panel wiring — filtered list **and** scoped badge count through the actual component tree; unfiltered totals.

Verified: `bun run check`, `check:i18n`, root `bun run lint`, and full UI suite (2550 tests) all green.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -1919,6 +1919,7 @@
   "owed_main_hint": "Offene Schritte nach dem Merge werden im Herd-Panel links angezeigt.",
   "owed_title": "Offene Schritte nach dem Merge",
   "owed_empty": "Keine offenen Schritte nach dem Merge — alles erledigt.",
+  "owed_repo_filter_empty": "Keine offenen Schritte nach dem Merge für {repo}.",
   "owed_note": "Dies sind alle manuellen Schritte, die die gemergten PRs deklariert haben. P2 erfasst die Bestätigung, nicht die Erledigung einzelner Schritte — daher erscheinen auch vor dem Merge bereits erledigte Schritte hier. Hake sie ab, um sie zu entfernen.",
   "owed_steps_count": "{done} von {total} erledigt",
   "owed_merged_ago": "vor {ago} gemergt",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -1919,6 +1919,7 @@
   "owed_main_hint": "Owed post-merge steps are shown in the herd panel on the left.",
   "owed_title": "Post-merge steps owed",
   "owed_empty": "No post-merge steps owed — you're all caught up.",
+  "owed_repo_filter_empty": "No post-merge steps owed for {repo}.",
   "owed_note": "These are every manual step the merged PRs declared. P2 tracks acknowledgement, not per-step completion, so steps you already did before merge appear here too — tick them to clear them.",
   "owed_steps_count": "{done} of {total} done",
   "owed_merged_ago": "merged {ago} ago",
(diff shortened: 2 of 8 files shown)


scoop (commented):
## Summary

The PR makes the Owed lens respect the repo chip filter by adding a pure `owedRecordsForRepo(records, repoFilter)` helper and wiring it into both the Herd lens-strip badge count and the `PostMergeStepsPanel` card list + empty state. The task is satisfied and the implementation is clean.

## What I verified

- **Helper + badge wiring**: `owedRecordsForRepo` is exported from `ui/src/lib/post-merge-steps.svelte.ts:6` and returns the same array ref for a `null` filter (verified against its `if (repoFilter == null) return records` at line 11). `Herd.svelte:213` derives `owedCount` from it, and `repoFilter` is a declared prop (`Herd.svelte:61,145`) in scope at that point.
- **Panel scoping keeps focus unfiltered**: `PostMergeStepsPanel.svelte:31` keeps `records` (unfiltered) as the source of truth for focus/frozen-card resolution, while `shownRecords` (line 33) drives only the list render (line 181) and the empty gate (line 140). This preserves the #1275 "live record always wins" invariant — confirmed the frozen-card path still reads unfiltered `records`.
- **`basename` import resolves**: `import { basename } from "$lib/components/learnings-drawer"` resolves to `ui/src/lib/components/learnings-drawer.ts:4` (there is no `learnings-drawer/index` in the same-named dir — verified via Glob), and the identical import is already used by `ui/src/routes/+page.svelte:77` and `ui/src/lib/components/page/AppOverlays.svelte:8`. `basename` yields the last non-empty path segment (learnings-drawer.ts:4-6), matching the display used by the sibling `UpNextPanel` (`repoBase`, UpNextPanel.svelte:152).
- **Scope semantics match siblings**: the `r.repoPath === repoFilter` comparison mirrors `UpNextPanel.svelte:42-43`, so the filter form is consistent with the rest of the repo-scoped herd view. `PostMergeSteps.repoPath` exists (types.ts:847).
- **i18n parity**: `owed_repo_filter_empty` exists with a `{repo}` placeholder in BOTH catalogs — en.json:1922 and de.json:1922 (verified both lines). Message usage `m.owed_repo_filter_empty({ repo: basename(repoFilter) })` matches the placeholder name.

PR #1336, opened 2026-07-02
chore(announcements): stamp new entries at next-version, gate stale ones

scoop:
## Problem

Feature-announcement fragments (`ui/src/lib/feature-announcements/entries/v<version>-<id>.ts`) carry a `sinceVersion`. Agents were reading it from `package.json` — but between releases `package.json` holds the **last released** version (release-please only bumps it when its release PR merges). So a new entry gets stamped at an already-shipped version and **silently never surfaces** in the What's-New drawer: `feature-gate.ts` only shows entries where `lastSeen < sinceVersion <= currentVersion`, and an upgraded user's `lastSeen` already covers that release.

The correct `sinceVersion` is always the **next** unreleased version (any `feat` is a release-please minor bump).

## Change (approach A — keep release-please's model, give agents the right number + enforce it)

- **`bun run next-version`** (`scripts/next-version.mjs`) — reads `.release-please-manifest.json` (release-please's authoritative last-release record) and prints the next minor. Single source of truth for "what version does an in-flight feature ship in." Typed via `next-version.d.mts` (matches the repo's existing `*.d.mts` pattern).
- **Version gate** (`scripts/check-announcement-versions.mjs`, wired into **PR hygiene** CI + `pre-push`) — for every entry fragment *added* in the branch range, fails if `sinceVersion` is `<=` the last release (would never surface) or if the filename `v<version>-` prefix disagrees with the field. Only added files are checked, so editing a historical entry never trips it. Fails closed on an unresolvable base ref (mirrors `check-feature-catalog.sh`).
- **Guidance** — `CLAUDE.md` → "Feature discovery" and the `feature-announcements.ts` header now say: use `bun run next-version`, never `package.json`.
- **Test** — `test/next-version.test.ts` covers the semver/bump logic.

## Verification

- `bun run lint`, `bun run typecheck`, `bun test ./test` (5642 pass), UI build, and all pre-push lanes + fallow audit green.
- Gate verified against throwaway fixtures: an already-released `v1.40.0` entry and a filename/field mismatch both fail with actionable messages; a valid `v1.41.0` entry passes.

## Note — cutting 1.41.0 is a separate operator action

Release PR **#1316 "chore(main): release 1.41.0"** is already open. Merging it cuts 1.41.0 and ships the two pending `v1.41.0-*` announcements. Left to the operator — release-please PRs are maintainer-approval-gated by design; not merged from this session.

[no-feature-entry] — tooling/docs only, ships no user-facing UX.


scoop (commented):
## Summary

The PR reinterprets "cut a new release" as fixing the *root cause* of the reported problem — agents stamping `sinceVersion` from `package.json`, which lags at the last *released* version between releases, so the entry never surfaces. Rather than a one-off release bump (which wouldn't stop the problem recurring), it adds a `next-version` helper, a CI + pre-push gate that fails stale/mismatched entries, and matching docs. This is the correct and more durable interpretation of the task, and the implementation is internally consistent.

## Verification

- `readReleasedVersion()` reads `.release-please-manifest.json` key `"."` — confirmed the key exists and equals `"1.40.0"` (`.release-please-manifest.json:2`), and `package.json` version is the same `1.40.0` (package.json:3).
- The minor-bump assumption in `nextVersion()` is justified by `release-please-config.json:4` (`"bump-minor-pre-major": true`), so a `feat` off `1.40.0` → `1.41.0`.
- The rationale in the comments (`entries surface only where lastSeen < sinceVersion <= currentVersion`) matches the actual gate: `ui/src/lib/feature-gate.ts:51` — `cmp(since, lastSeen) > 0 && cmp(since, current) <= 0`.
- Filename regex `^v(\d+\.\d+\.\d+)-` and field regex both match the established convention (filename `v1.41.0-tab-signaling.ts`, field `sinceVersion: "1.41.0"`), verified against existing entries under `ui/src/lib/feature-announcements/entries/`.
- `check-announcement-versions.mjs` fails CLOSED on an unresolvable base and only inspects `--diff-filter=A` files, so historical entries (e.g. `v1.40.0-anonymous-telemetry.ts`, whose `sinceVersion == released`) are correctly not tripped.
- CLAUDE.md example updated `v1.40.0-quick-filter.ts` → `v1.41.0-quick-filter.ts`, consistent with the next-version guidance; no new i18n/message keys introduced, so no locale-parity surface.
- `test/` and `scripts/` fall under the root `tsconfig.json` (no `include`; only `ui/extension/site/docs-site` excluded); `moduleResolution: "bundler"` resolves the `next-version.mjs` import to the added `next-version.d.mts`, so the `.d.mts` declaration is doing its intended job.

No blocking correctness, security, or quality problems found.

## Notes (non-blocking, informational)

- `check-announcement-versions.mjs` filters with `--diff-filter=A`, so if a bad entry is corrected by *renaming* an existing fragment (e.g. `v1.40.0-foo.ts` → `v1.41.0-foo.ts`) with rename detection on, git reports it as `R`, not `A`, and the gate would skip it. This is a narrow, benign coverage hole (the corrective path), not a defect — noting it only for awareness.
- `git fetch --quiet origin main` runs on every pre-push; offline it fails silently and the subsequent `rev-parse origin/main` still passes when the ref already exists locally, so offline pushes are unaffected.

Could not verify anything external — all dependencies of this change are in-repo and were checked against the files/lines cited above.

PR #1340, opened 2026-07-02
fix(preview): clarify tailnet preview setup contract

kai-osthoff:
## Summary
- Clarify the generated local preview setup prompt: the repo script starts the dev server and writes .shepherd-preview, but must not run tailscale serve.
- Document that Shepherd detects the port, binds the preview proxy, and registers the proxy slot with Tailscale when auto-serve is enabled.
- Extend the service prompt test to lock in the contract.

## Tests
- bun test test/service.test.ts


kai-osthoff:
READY PR, @scoop. Your move. 🫡🚨✅👀⏰

PR #1345, opened 2026-07-02
feat(ui): fuzzy matching for the command bar

scoop:
## What

Replaces the command bar's plain **substring** filter with **fuzzy subsequence matching + relevance ranking**, and highlights the matched characters. Typing `nwr` now finds `newer`, `dpl` finds a `deploy` session.

## How

- **New `ui/src/lib/fuzzy.ts`** — a small, dependency-free subsequence scorer returning `{ score, positions } | null`. Word-boundary (+10) and consecutive-run (+5) bonuses make prefix/contiguous matches rank highest; a non-subsequence returns `null` (row excluded). Fully unit-tested in `fuzzy.test.ts`.
- **`CommandBar.svelte`** — each group now filters + ranks by score (tie-broken by recency), and matched chars in the primary text render in `<mark class="cb-hl">` (amber token). Fuzzy runs over the **short fields only** (title/desig/repo name) so a short query can't subsequence-match nearly every long prompt.
- **Sessions also match on their `prompt` (description)** via a contiguous substring — a session surfaced *only* via its description shows a **"matches description"** affordance so its presence is explained. (Contiguous substring, not fuzzy, keeps it selective over long prompts.)
- **Blank query preserves today's behavior byte-for-byte** — it scores every row 0, so ordering falls back to recency exactly as before.

## Accessibility

`<mark>` split runs concatenate back to the original text, so each `role="option"`'s accessible name is unchanged — asserted by test. Highlight color is the `--color-amber` semantic token; the UA `<mark>` background is reset.

## Feature discovery

The command bar is still unreleased (same `1.41.0` train as its What's-New entry), so this folds into the existing `v1.41.0-command-bar` announcement — reworded body copy (EN+DE) + updated entry comment — rather than adding a redundant second card.

## Tests

- `ui/src/lib/fuzzy.test.ts` — 9 scorer unit tests.
- `CommandBar.browser.test.ts` — added fuzzy subsequence, best-match ordering (score beats recency), prompt-only inclusion + affordance, and highlight/accessible-name tests; renamed the substring test to `filters fuzzily…`.
- Full `ui` suite (2604 tests), `bun run check`, `check:i18n`, eslint, prettier, and the pre-push PR-hygiene gate all green.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2341,7 +2341,7 @@
   "commandbar_repo_affordance": "Repository",
   "commandbar_no_matches": "Keine Treffer",
   "feat_command_bar_title": "Befehlsleiste — mit ⌘K zu allem springen",
-  "feat_command_bar_body": "Drücke ⌘K (macOS) oder Strg+K (Windows/Linux) — auch bei fokussiertem Terminal — um einen Schnellumschalter zu öffnen. Tippe, um deine Sessions, Repositories und Herden-Ansichten zu filtern, und springe mit Enter direkt dorthin.",
+  "feat_command_bar_body": "Drücke ⌘K (macOS) oder Strg+K (Windows/Linux) — auch bei fokussiertem Terminal — um einen Schnellumschalter zu öffnen. Tippe, um deine Sessions, Repositories und Herden-Ansichten unscharf zu durchsuchen — passende Buchstaben werden hervorgehoben, und Sessions werden auch anhand ihrer Beschreibung gefunden — und springe mit Enter direkt dorthin.",
   "settings_tab_ticker_title": "Kompakter Tab-Titel",
   "settings_tab_ticker_hint": "Zeigt im Hintergrund-Tab gruppierte Symbolzähler (⚠ CI fehlgeschlagen · ✋ blockiert · ✓ bereit · ▶ läuft) statt einer einzelnen Zahl. Standardmäßig aus.",
   "settings_tab_ticker_on": "Kompakter Tab-Titel an",
@@ -2361,5 +2361,6 @@
   "commandbar_cmd_next_needs_you": "Zur nächsten Session springen, die dich braucht",
   "commandbar_cmd_next_needs_you_kw": "wartend blockiert aufmerksamkeit",
   "feat_command_bar_v2_title": "Befehlsleiste führt jetzt Befehle aus und durchsucht die Doku",
-  "feat_command_bar_v2_body": "⌘K/Strg+K kann mehr als nur zwischen Sessions springen. Eine Gruppe „Befehle“ führt Verben aus — Neue Aufgabe, Broadcast, Einstellungen, Nutzung, Erneut versuchen, Zur nächsten Session springen, die dich braucht — und eine Gruppe „Doku“ durchsucht die Dokumentation und öffnet die Seite in einem neuen Tab. Tippe einfach, was du tun möchtest."
+  "feat_command_bar_v2_body": "⌘K/Strg+K kann mehr als nur zwischen Sessions springen. Eine Gruppe „Befehle“ führt Verben aus — Neue Aufgabe, Broadcast, Einstellungen, Nutzung, Erneut versuchen, Zur nächsten Session springen, die dich braucht — und eine Gruppe „Doku“ durchsucht die Dokumentation und öffnet die Seite in einem neuen Tab. Tippe einfach, was du tun möchtest.",
+  "commandbar_prompt_match": "passt zur Beschreibung"
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2341,7 +2341,7 @@
   "commandbar_repo_affordance": "Repository",
   "commandbar_no_matches": "No matches",
   "feat_command_bar_title": "Command bar — jump to anything with ⌘K",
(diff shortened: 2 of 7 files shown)


scoop (commented):
## Summary

This PR adds fuzzy (subsequence) matching to the command bar: a new `fuzzyScore` helper, per-group fuzzy filtering + relevance ranking, matched-character highlighting via a `<mark>`-wrapping snippet, and a "matches description" affordance for sessions surfaced only by their prompt. The implementation is clean, well-commented, and satisfies the task. Tests and message updates are consistent.

### Verified
- **i18n parity** — the new key `commandbar_prompt_match` exists in both catalogs (`ui/messages/en.json:2343`, `ui/messages/de.json:2343`); these are the only two catalogs (`ui/messages/` contains just `en.json` + `de.json`). The reworded `feat_command_bar_body` is updated in both catalogs. Paraglide messages are generated (not committed), so `m.commandbar_prompt_match()` will resolve.
- **`s.prompt` safety** — `Session.prompt` is a required `string` (`ui/src/lib/types.ts:764`), so `s.prompt.toLowerCase().includes(q)` (`CommandBar.svelte:86`) cannot throw.
- **Sort fields** — `Session.updatedAt` (`ui/src/lib/types.ts:819`) and repo `lastUsedAt` exist; the new comparators (`CommandBar.svelte:90,107,120`) reference them correctly. Empty query scores every row 0, so the sort falls back to `updatedAt`/`lastUsedAt`, preserving the prior unfiltered ordering.
- **Highlight mapping** — each haystack begins with the primary text (`title + " " + desig + " " + repoName`, `r.name + " " + r.display`, or the lens label), and `hl` is filtered to `p < title.length` / `p < r.name.length` (`CommandBar.svelte:98,113`), so highlighted indices land on the primary. `segs()` concatenates back to the original text, so the accessible name is preserved (verified by the added browser test).
- **Ranking** — boundary/consecutive bonuses (`fuzzy.ts:8-10`) make a whole-word prefix outrank a mid-word match; the greedy earliest-binding match on `deploy` vs `redeploy` gives d1 a higher score than d2, matching `CommandBar.browser.test.ts` and `fuzzy.test.ts`.
- **CSS token** — `--color-amber` is defined (`ui/src/app.css:21`).
- The feature-announcement entry change is comment-only and does not alter behaviour.

### Nit (non-blocking)
`fuzzyScore` scans the lowercased haystack `t` but returns `positions` that are later used to index the original-cased `text` in the `primary` snippet. For the rare characters whose `toLowerCase()` changes string length (e.g. `İ` → `i̇`), the highlight `positions` drift for characters after that point. Filtering and ranking remain correct (they operate entirely within the lowercased space); only the rendered `<mark>` highlight is cosmetically misaligned. Practically unreachable for ASCII/Latin session, repo, and lens names, hence a nit rather than a blocker.

scoop:
**Addressed the critic finding + rebased onto latest `main`.**

1. **Highlight-position drift (fixed):** `fuzzyScore` now matches case-insensitively *per character against the original text* instead of a pre-lowercased copy, so `positions` index the original string even when a char's lowercase form changes length (e.g. `İ`). Added a regression test (`fuzzyScore("x", "İx") → [1]`).

2. **Rebase note (heads-up, no disagreement):** `main` gained command + doc groups (and an "all" lens) in the command bar since I branched. I resolved the conflict by scoping fuzzy matching to the **navigation** groups (sessions / repos / lenses) and leaving **commands/docs on substring** — their `keywords` haystacks are long free text, over which a short fuzzy *subsequence* matches nearly everything and floods the bar (same over-match reason the earlier review flagged for session prompts). All of `main`'s command/doc/all-lens tests pass unchanged.

scoop (commented):
## Re-review

The single prior finding — highlight-position drift when a char's `toLowerCase()` changes length — is **addressed**.

- `ui/src/lib/fuzzy.ts:36` now matches case-insensitively **per character against the original `text`** (`text[ti].toLowerCase() === q[qi]`) instead of scanning a pre-lowercased haystack, so `positions` index the un-lowercased string. Traced the `İx` case: `text[0]="İ"`, `"İ".toLowerCase()==="i̇"` (≠ `"x"`), then `text[1]="x"` matches → `positions=[1]`. A regression test locks this in at `ui/src/lib/fuzzy.test.ts:31` (`fuzzyScore("x", "İx") → [1]`).

### Other correctness checks (all verified)
- **i18n parity:** `commandbar_prompt_match` present in both catalogs — `ui/messages/en.json` and `ui/messages/de.json` (both in diff, key added in each).
- **`s.prompt` access is safe:** `Session.prompt` is a required `string` (`ui/src/lib/types.ts:764`), so `s.prompt.toLowerCase()` at `ui/src/lib/components/CommandBar.svelte:98` cannot throw on undefined.
- **Highlight positions:** for session/repo rows `hl` is clamped with `.filter(p => p < title.length)` / `< r.name.length` so positions map onto the rendered primary text only (`CommandBar.svelte:109,127`); lens uses the whole label as primary (`CommandBar.svelte:139`). The `segs()`/`primary` snippet concatenates back to the original text, preserving the accessible name (verified against the browser test at `CommandBar.browser.test.ts:181-186`).
- **Empty query:** `fuzzyScore("", …)` returns `{score:0, positions:[]}` (`fuzzy.ts:25`), so every row passes the filter and the sort falls back to `updatedAt`/`lastUsedAt`, preserving the previous unfiltered ordering.
- **Greedy match completeness:** the left-to-right two-pointer is complete for subsequence existence (never yields a false `null`); scoring is explicitly non-optimal but adequate for short strings, as documented.
- **Substring scoping for commands/docs:** deliberately kept on substring (`CommandBar.svelte:147-149` comment); no correctness impact.

No blocking problems and nothing left to address.

PR #1348, opened 2026-07-02
feat(ui): filter session list from command bar repo results

scoop:
## What

Adds a **secondary action** to repository results in the command bar (⌘/Ctrl+K).

- **Plain ↵** — unchanged: opens the repo's per-repo **Backlog** overlay.
- **⇧↵ / ⌘↵ / Ctrl↵** — new: applies the repo as the herd `repoFilter` (the same
  state the RepoSwitcher chips drive), scoping the session list to that repo. The
  bar and backlog close so the filtered herd is visible, and the existing
  `repoFilter` effect re-targets terminal selection onto a session in that repo.

## Why gated on a live session

An existing effect (`shouldClearRepoFilter`, `+page.svelte`) auto-clears
`repoFilter` when the target repo has no live (non-archived) session — filtering to
a session-less repo would strand an empty herd with no clearing chip. So the
secondary action is offered **only** for repos that currently have a live session
(`hasLiveSession`, same liveness rule as `repoChipRows`). Session-less repos ignore
the modifier and fall back to opening the backlog.

## Discoverability

Every repo row keeps its `Repository` sub-label (the primary-action signal). Live
rows additionally show a `⇧↵ filter` hint as a **separate, right-pinned,
non-shrinking** element (`.cb-hint`), so a long display path truncates the
sub-label — never the hint. Session-less rows show no hint, which is what signals
the modifier is inert there. No touch affordance is needed: the bar's only opener
is the ⌘/Ctrl+K chord, so every user of it has a keyboard.

## Notes

- New i18n keys (EN + DE parity): `commandbar_repo_filter_affordance`,
  `feat_command_bar_repo_filter_title/body`. The announcement body scopes the claim
  to a repo's *active sessions* so it doesn't over-promise for session-less repos.
- Feature-announcement entry `v1.41.0-command-bar-repo-filter.ts` added (feature
  discovery catalog).

## Testing

- `cd ui && bun run check` — 0 errors, `bun run check:i18n` — parity (2365 keys).
- `bun run test:browser` (868) + `test:node` (1771) pass, incl. new CommandBar
  cases: ⇧/⌘/Ctrl+Enter filter a live repo, plain Enter still selects, modifier on
  a session-less repo falls back to select, and the hint shows only on live rows.
- PR-hygiene gates (branch hygiene, announcement versions, feature catalog) pass.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2362,5 +2362,8 @@
   "commandbar_cmd_next_needs_you_kw": "wartend blockiert aufmerksamkeit",
   "feat_command_bar_v2_title": "Befehlsleiste führt jetzt Befehle aus und durchsucht die Doku",
   "feat_command_bar_v2_body": "⌘K/Strg+K kann mehr als nur zwischen Sessions springen. Eine Gruppe „Befehle“ führt Verben aus — Neue Aufgabe, Broadcast, Einstellungen, Nutzung, Erneut versuchen, Zur nächsten Session springen, die dich braucht — und eine Gruppe „Doku“ durchsucht die Dokumentation und öffnet die Seite in einem neuen Tab. Tippe einfach, was du tun möchtest.",
-  "commandbar_prompt_match": "passt zur Beschreibung"
+  "commandbar_prompt_match": "passt zur Beschreibung",
+  "commandbar_repo_filter_affordance": "⇧↵ filtern",
+  "feat_command_bar_repo_filter_title": "Sitzungen aus der Befehlsleiste filtern",
+  "feat_command_bar_repo_filter_body": "Wähle in der Befehlsleiste ein Repository aus und drücke dann ⇧↵ (oder ⌘/Strg+↵), um die Sitzungsliste auf die aktiven Sitzungen dieses Repos zu filtern. Mit einfachem ↵ öffnest du weiterhin sein Backlog."
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2362,5 +2362,8 @@
   "commandbar_cmd_next_needs_you_kw": "waiting blocked attention",
   "feat_command_bar_v2_title": "Command bar now runs commands and searches docs",
   "feat_command_bar_v2_body": "⌘K/Ctrl+K does more than jump between sessions now. A Commands group runs verbs — New task, Broadcast, Settings, Usage, Retry, Jump to next needs-you — and a Docs group searches the documentation and opens the page in a new tab. Just start typing what you want to do.",
-  "commandbar_prompt_match": "matches description"
+  "commandbar_prompt_match": "matches description",
+  "commandbar_repo_filter_affordance": "⇧↵ filter",
+  "feat_command_bar_repo_filter_title": "Filter sessions from the command bar",
+  "feat_command_bar_repo_filter_body": "Pick a repository in the command bar, then press ⇧↵ (or ⌘/Ctrl+↵) to filter the session list to that repo's active sessions. Plain ↵ still opens its backlog."
 }
--- a/ui/src/lib/components/CommandBar.browser.test.ts
+++ b/ui/src/lib/components/CommandBar.browser.test.ts
@@ -58,11 +58,15 @@ function session(partial: Partial<Session> & { id: string }): Session {
 }
 
 function seedRepos(): RepoEntry[] {
-  // beta has the higher lastUsedAt → sorts before alpha.
+  // beta has the higher lastUsedAt → sorts before alpha. gamma has NO live session
+  // (no seedSessions entry references it), so its filter affordance/action is inert.
   return [
     {
+      // Symlinked repo root: entry path differs from realPath. session s1.repoPath is
+      // the realpath'd form ("/repos/alpha"), matching realPath — so the filter action
+      // must key on realPath, while the backlog (onselectrepo) keys on the raw path.
       name: "alpha",
-      path: "/repos/alpha",
+      path: "/repos/alpha-link",
       display: "~/repos/alpha",
       realPath: "/repos/alpha",
       lastUsedAt: 100,
@@ -73,6 +77,13 @@ function seedRepos(): RepoEntry[] {
       display: "~/repos/beta",
       realPath: "/repos/beta",
       lastUsedAt: 200,
+    },
+    {
+      name: "gamma",
+      path: "/repos/gamma",
+      display: "~/repos/gamma",
+      realPath: "/repos/gamma",
+      lastUsedAt: 50,
     },
   ];
 }
@@ -88,6 +99,7 @@ function seedSessions(): Session[] {
 function renderBar(overrides: Partial<Parameters<typeof CommandBar>[1]> = {}) {
   const onselectsession = vi.fn();
   const onselectrepo = vi.fn();
+  const onfilterrepo = vi.fn();
   const onselectlens = vi.fn();
   const onclose = vi.fn();
   render(CommandBar, {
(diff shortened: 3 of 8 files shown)


scoop (commented):
## Review

The PR adds a secondary command-bar action on repository results: plain `↵` still opens the repo backlog, while `⇧↵` / `⌘↵` / `Ctrl↵` (and modifier+click) filter the herd's session list to that repo. It is discoverable via an inline `⇧↵ filter` hint on eligible rows plus a feature announcement. The task is satisfied and I found no bugs.

### Verified

- **Behavior split** — `selectOption(row, secondary)` routes repo rows to `onfilterrepo` only when `secondary && row.hasLiveSession`, else `onselectrepo` (`ui/src/lib/components/CommandBar.svelte:192-198`). Enter passes `e.shiftKey || e.metaKey || e.ctrlKey` (`:234`), click passes the same (`:297`), and Space still runs the primary action only (`:299`). No regression to plain-Enter selection.
- **Liveness gate is consistent with the chip/clear rules** — `liveRepoPaths` = non-archived sessions keyed by `s.repoPath` (`CommandBar.svelte:98-100`), which mirrors `repoChipRows` (verified against `ui/src/lib/components/queue-strip.ts:57-60`) and `shouldClearRepoFilter` (`queue-strip.ts:97-99`). A repo that passes the gate is guaranteed a chip, so `shouldClearRepoFilter` won't immediately null the filter set in `oncommandbarfilterrepo` (`ui/src/routes/+page.svelte:2777-2785`).
- **Path-key alignment** — the herd filter compares `s.repoPath === repoFilter` (`+page.svelte:351`); the gate is built from `s.repoPath` and checked against `r.path`, so the action only fires when `r.path` is an actual session `repoPath`. Self-consistent, and `store.sessions` (unfiltered herd) is what CommandBar receives (`ui/src/lib/components/page/AppOverlays.svelte:537`).
- **Prop threading** — `onfilterrepo` → `oncommandbarfilterrepo` is wired through CommandBar → AppOverlays (`AppOverlays.svelte:271,542`) → +page (`+page.svelte:2777`); both render sites (component + test) provide the now-required prop.
- **i18n parity** — `commandbar_repo_filter_affordance`, `feat_command_bar_repo_filter_title`, `feat_command_bar_repo_filter_body` all present in both catalogs (verified `ui/messages/en.json:2342,2366,2367` and `ui/messages/de.json:2342,2366,2367`).
- **Feature announcement** — auto-registered via the entries glob (`ui/src/lib/feature-announcements.ts:42-45`), conforms to the `FeatureAnnouncement` type with optional `targetId` omitted, and `sinceVersion: 1.41.0` matches the sibling command-bar entries in the same release.

Nothing to change.

scoop:
## Review

The secondary command-bar action is cleanly plumbed (`onfilterrepo` → `AppOverlays.svelte` → `+page.svelte`), gated on `Shift`/`Cmd`/`Ctrl`+Enter (and click), discoverable via a `⇧↵` hint and a feature-announcement entry, and locale parity holds — all three new keys exist in both catalogs (verified `commandbar_repo_filter_affordance`, `feat_command_bar_repo_filter_title`, `feat_command_bar_repo_filter_body` at en.json:2366-2368 and de.json:2366-2368). The `v1.41.0` feature entry is auto-discovered by the glob in feature-announcements.ts:42-44 and its `sinceVersion` matches its sibling `v1.41.0-*` entries. The space-vs-Enter split in the row keydown handler (svelte:365-366) is correct: space always runs the primary action.

### Blocking: raw-path vs realpath mismatch in `CommandBar.svelte`

`RepoEntry.realPath` is documented as "the form `session.repoPath` (safeRepoDir) carries" (verified `ui/src/lib/types.ts:41-44`), and `repos.svelte.ts:5-7` states `session.repoPath` is realpath-resolved while `listRepos` uses the raw `path` form — the whole reason `pathIndex` maps *both* forms. The new code conflates them:

- `liveRepoPaths` is built from `session.repoPath` (realpath form) — `CommandBar.svelte:128-129`.
- `hasLiveSession: liveRepoPaths.has(r.path)` checks the *raw* `RepoEntry.path` against that realpath set — `CommandBar.svelte:142`.
- `onfilterrepo(row.path)` passes the *raw* path (`CommandBar.svelte:258` and `:364`) to `+page.svelte:2784` `repoFilter = path`, which the herd filters via `s.repoPath === repoFilter` (realpath form; verified `+page.svelte:350-351`).

When `path !== realPath` (any repo root reached through a symlink) two failures stack on an already-executing path: (1) `hasLiveSession` is false for repos that *do* have live sessions, so the `⇧↵` hint never renders and the modifier is inert; (2) even if fired, `repoFilter` is set to a value no `session.repoPath` matches, so `shouldClearRepoFilter` (`ui/src/lib/components/queue-strip.ts:97-99`) immediately nulls it — stranding an empty herd, the exact outcome the `+page.svelte:2777-2781` comment claims to avoid. The RepoSwitcher chips get this right by keying on `chip.repoPath` (realpath form) — verified `RepoSwitcher.svelte:207`.

The tests seed `path === realPath` (`CommandBar.browser.test.ts:66-84`), so they cannot catch the divergence. Fix: key the liveness set membership and the filter payload on `r.realPath` (`liveRepoPaths.has(r.realPath)` and `onfilterrepo(r.realPath)`), while leaving the primary `onselectrepo(r.path)` on the raw form the backlog expects.

scoop (commented):
## Re-review

The single prior finding — that `hasLiveSession` and `onfilterrepo` used the raw `RepoEntry.path` instead of `realPath`, breaking symlinked repo roots — is **fully addressed**.

Verified:
- `ui/src/lib/components/CommandBar.svelte:148` now computes `hasLiveSession: liveRepoPaths.has(r.realPath)`, and `selectOption` fires `onfilterrepo(row.realPath)` while `onselectrepo(row.path)` keeps the raw path (CommandBar.svelte:262-264). The row type carries both `path` and `realPath` (CommandBar.svelte:58-66).
- `liveRepoPaths` is keyed on `s.repoPath` (CommandBar.svelte:131), the realpath'd form. That matches the chip liveness rule in `repoChipRows` (verified against ui/src/lib/components/queue-strip.ts:59, which skips `s.status === "archived"` and buckets by `s.repoPath`) and the herd filter comparison `s.repoPath === repoFilter` (verified against ui/src/routes/+page.svelte:351).
- The page handler sets `repoFilter = path` with the realPath payload (ui/src/routes/+page.svelte:2784); because the secondary action is gated on `hasLiveSession`, `shouldClearRepoFilter` won't immediately null it (auto-clear logic at +page.svelte:299-311).
- `realPath` is a required field on `RepoEntry` (verified against ui/src/lib/types.ts:44).

Other correctness checks:
- **i18n parity**: the three new keys `commandbar_repo_filter_affordance`, `feat_command_bar_repo_filter_title`, `feat_command_bar_repo_filter_body` exist in BOTH catalogs (verified against ui/messages/en.json:2366-2368 and ui/messages/de.json:2366-2368). The pre-existing `commandbar_repo_affordance` used by the test exists in both (ui/messages/en.json:2341, ui/messages/de.json:2341).
- **Feature announcement**: the new entry auto-registers via `import.meta.glob("./feature-announcements/entries/*.ts")` (verified against ui/src/lib/feature-announcements.ts) — no manual barrel edit required. `sinceVersion: "1.41.0"` is consistent with the other pre-existing v1.41.0 entries in the tree (feature-announcements/entries/v1.41.0-*.ts). The exact last-released version is enforced by the `check-announcement-versions.mjs` CI gate, which I cannot run here.
- **Wiring**: `onfilterrepo` threads +page.svelte → AppOverlays (prop declared ui/src/lib/components/page/AppOverlays.svelte:271, forwarded at 542) → CommandBar. Test props updated accordingly (AppOverlays.browser.test.ts:103).
- **Discoverability**: satisfied two ways — an inline `⇧↵ filter` hint rendered on live-session repo rows (CommandBar.svelte:392-396) and the What's-New announcement entry.
- **Keyboard/click parity**: both the input-level Enter handler (CommandBar.svelte:303) and the option-level handlers (CommandBar.svelte:372-375) fire the secondary verb on `shiftKey || metaKey || ctrlKey`; Space intentionally stays primary-only.

No blocking issues found. The implementation satisfies the task.

PR #1355, opened 2026-07-02
fix(up-next): recompute after start once the claim label lands

scoop:
## Problem

After starting task(s) from the **Up Next** lens, its contents went stale: the just-started items lingered until the next 15-min interval tick, a manual refresh, or a lens re-open.

**Root cause (async write/read race).** `handleUpNextStart` stamped the `shepherd:active` exclusion label **fire-and-forget** (`void claimLinkedIssue(...)`, an async two-`gh`-subprocess write) and then kicked `void deps.upNext.refresh()` in the **same tick**. `compute()` re-read issues via an uncached `gh issue list` *before* the label landed, so `isExcludedIssue()` didn't filter the item and it re-surfaced in the pushed snapshot. (The `execFileSync` comments elsewhere in `server.ts` are stale — `addIssueLabel` is async via `promisify(execFile)`.)

## Fix

Replace the immediate refresh with a **backgrounded, guaranteed-fresh, verify-and-retry** recompute owned by `UpNextService`:

- **Await the claim writes first** (behind an 8s timeout guard so a hung `gh` can't stall it) — makes the common case need exactly one recompute.
- **`recomputeUntilCleared(startedRefs)`** recomputes until none of the started items appear in the snapshot, or the bounded attempts (`[0, 3s, 7s]`) run out. Keying on *the item being gone* — not elapsed time — handles GitHub read-after-write lag; a single blind delay beaten by lag would fall back to the 15-min tick (the original bug).
- **Non-coalescing recompute** — awaits any in-flight compute (a 15-min tick / GET refresh that may predate the claim) then starts a fresh one, so a stale in-flight compute can't cache a pre-label snapshot or re-show a cleared item.
- **Path-space reconciliation** — started refs carry `safeRepoDir`/realpath paths while snapshot items carry raw `join(repoRoot,name)` paths; membership maps real→raw (mirrors `reconcileRealPathsToRaw`) and matches on `(raw repoPath, issue number)`, so a symlinked `repoRoot` can't silently defeat the check.
- A transient compute throw no longer aborts the remaining retries.

All backgrounded — the start response is never blocked. The 15-min interval loop remains the ultimate backstop. Common case = one `compute()` (no steady-state cost regression); extra fan-outs only on the rare lag path, bounded at 3.

## Tests

- `test/up-next.test.ts` — `recomputeUntilCleared`: clears-in-one, retries-then-succeeds, bounded, non-coalescing, symlinked-root path-space (+ identity control), no-refs no-op, and transient-throw resilience.
- `test/server-up-next.test.ts` — start uses `recomputeUntilCleared(startedRefs)`, not an immediate `refresh`.
- `bun run lint`, `bun run typecheck`, `bun run test` (5687 tests) all green. Server-only change — no UI/i18n/feature-catalog/glossary impact.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2256,6 +2256,7 @@ const appDeps: AppDeps = {
   upNext: {
     snapshot: () => upNext.snapshot(),
     refresh: () => upNext.refresh(),
+    recomputeUntilCleared: (started) => upNext.recomputeUntilCleared(started),
     hiddenRepoPathsRaw: () =>
       reconcileRealPathsToRaw(store.hiddenRepoPaths(), listRepos(config.repoRoot)),
   },
--- a/src/server.ts
+++ b/src/server.ts
@@ -361,6 +361,12 @@ export interface AppDeps {
   upNext?: {
     snapshot(): UpNextSnapshot | null;
     refresh(): Promise<UpNextSnapshot>;
+    /** Recompute (guaranteed-fresh, verify-and-retry) until the just-started items drop from
+     *  the lens — i.e. their `shepherd:active` claim labels have landed. Backgrounded by the
+     *  caller so it never blocks the start response. */
+    recomputeUntilCleared(
+      started: readonly { repoPath: string; issueNumber: number }[],
+    ): Promise<void>;
     /** Raw-path-space hidden set (reconciled), applied at read time for instant freshness. */
     hiddenRepoPathsRaw(): Set<string>;
   };
@@ -858,6 +864,11 @@ interface UpNextStartItem {
   issueRef: { number: number; url: string; title: string; body: string };
 }
 
+/** Cap on how long the backgrounded post-start recompute waits for the claim label writes
+ *  to settle before recomputing anyway — a hung `gh` must not stall it (verify-and-retry and
+ *  the 15-min interval loop still backstop). */
+const CLAIM_SETTLE_TIMEOUT_MS = 8_000;
+
 /** Validate the POST /api/up-next/start body into safe, repo-scoped start items. */
 function parseUpNextStartItems(body: unknown): UpNextStartItem[] | null {
   const itemsRaw = (body as { items?: unknown } | null)?.items;
@@ -896,6 +907,8 @@ async function handleUpNextStart(req: Request, deps: AppDeps): Promise<Response>
 
   const created: Session[] = [];
   const errors: { number: number; error: string }[] = [];
+  const claims: Promise<void>[] = [];
+  const startedRefs: { repoPath: string; issueNumber: number }[] = [];
   // Strictly serial — concurrent `git worktree add` on the same repo would contend.
   await startSerially(items, async (it) => {
     const forge = deps.resolveForge?.(it.dir) ?? null;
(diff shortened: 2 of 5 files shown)


scoop (commented):
## Summary

The PR replaces the immediate, fire-and-forget post-start `upNext.refresh()` with a backgrounded `recomputeUntilCleared(started)` that (1) waits for the `shepherd:active` claim-label writes to settle (behind an 8s timeout guard), then (2) does a guaranteed-fresh, non-coalescing recompute and retries a bounded number of times until the just-started issues drop from the lens. This directly matches the task: don't recompute immediately (before the label lands), but sooner than the 15-min interval backstop.

I verified the implementation against ground truth rather than assuming plausibility.

### Correctness verification

- **Path-space reconciliation is sound.** `startedRefs` are populated from `it.dir`, which is `safeRepoDir(...)` — and `safeRepoDir` returns a `realpathSync`-resolved path (verified `src/validate.ts:734,741`). Snapshot items carry raw `entry.path` = `join(repoRoot,name)` space (verified `src/up-next.ts:53` / `buildUpNextRepos`). `stillPresent` maps real→raw via the current repo set and matches on `(raw repoPath, issue number)` (`src/up-next.ts:171-188`), mirroring `reconcileRealPathsToRaw`. Item shape matched against `UpNextItem { repoPath; issueRef.number }` (verified `src/up-next-core.ts:20,32`). Number-alone matching would collide across repos; the composite key is the right call.
- **Single-flight bypass is real.** `refresh()` is single-flight and nulls `inFlight` in `.finally` (verified `src/up-next.ts:130-136`). `recompute()` awaits any in-flight compute first, then calls `refresh()` — because the `.finally` runs before the awaiting continuation resumes, `inFlight` is null at that point and a genuinely fresh compute starts (`src/up-next.ts:161-164`). This correctly prevents joining a compute that began before the claim label landed. Test (d) exercises exactly this.
- **Claim handles are safe to await.** `claimLinkedIssue` catches internally and always resolves `void` (verified `src/server.ts:1742-1748`), so `Promise.allSettled(claims)` can only be delayed by a slow/hung `gh` label write — which is exactly what `CLAIM_SETTLE_TIMEOUT_MS` (8s) guards via `Promise.race` (`src/server.ts:956-963`). Handlers attach synchronously at IIFE entry, so no unhandled-rejection window (and claims never reject anyway).
- **Transient-failure resilience.** A compute throw inside `recomputeUntilCleared` is caught and does not abort remaining retries (`src/up-next.ts:201-209`); test (g) covers it.
- **Guards.** `startedRefs.length > 0` gate (`src/server.ts:955`) and the `started.length === 0` early return (`src/up-next.ts:197`) avoid needless work when nothing started. Dropping the old unconditional refresh when zero items start is arguably more correct (nothing to clear).
- **Runtime/portability.** `setTimeout(...).unref?.()` optional-chaining is safe under Bun (the project's test/start runtime, verified `package.json:9,13`); `realpathSync` from `node:fs` is already used elsewhere (`src/validate.ts:1`).
- **Callers/consumers agree.** The new `AppDeps.upNext.recomputeUntilCleared` (`src/server.ts:361-368`) is implemented in `src/index.ts:2259` and the test harness (`test/server-up-next.test.ts:68`); the only two `upNext` dep constructions. The retained `refresh()` member is still used by the GET-open and manual-refresh paths (`src/server.ts:838,850`), so nothing is orphaned or broken.

No user-facing strings or message catalogs are touched, so no locale-parity check applies.

I found no blocking defects, logic bugs, or security issues.

PR #1356, opened 2026-07-02
fix(up-next): honor issue assignee (mine & unassigned)

scoop:
## What

**Up Next** previously ranked *every* open issue regardless of who it was assigned to — issues assigned to other people appeared identically to unassigned ones. This brings Up Next to parity with the Backlog's existing "mine & unassigned" filter (#824): it now surfaces only issues that are **unassigned OR assigned to me**.

## How

Filtering happens **server-side at compute time** (the cached snapshot arrives pre-filtered — no UI/GET plumbing). "Me" is per-forge, so the viewer login is resolved per repo via `forge.currentUser()` (already used by the Backlog's issues endpoint), where the forge is already in hand — no new `UpNextDeps` field, no `index.ts` wiring.

- **Standalone issues** filter on their own `assignees`.
- **Epic units** filter on the **parent's** `assignees` (threaded as `EpicUnitInput.parentAssignees`). The epic candidate child is synthesized with `assignees: []`, so the parent is the assignee-bearing issue — matching what the Backlog hides.
- **Fails open**: when the viewer login can't be resolved (local forge / un-authed / no identity API), no assignee filtering happens.

## Decisions (confirmed with operator)

- **Always on, no toggle** — permanent server-side filter; others' issues remain visible on the Backlog, not in Up Next. A future toggle would need snapshot + UI re-plumbing (known follow-up, not in scope).
- **Epic parents are filtered too** — full #824 parity, not a divergence.

## Behavior change to note

A `shepherd:priority` issue — or a whole repo whose only work is assigned solely to others — now drops out of Up Next. Intended (that's the filter's purpose), but visible.

## Not a feature announcement

Server-only change; no `ui/src/lib/components/**` or `routes/**` touched, so the feature-catalog gate correctly does not fire. Framed as a consistency **fix** bringing Up Next to Backlog #824 parity.

## Tests

- `test/up-next-core.test.ts` — standalone (mine / unassigned / others / both / null-viewer) + epic-parent (mine / unassigned / others / null-viewer) cases.
- `test/up-next.test.ts` — asserts `resolveRepo` threads `currentUser` → viewer, `resolveEpics` threads parent assignees, and fail-open when the forge has no `currentUser`.

`bun test` (5678 pass, 0 fail), `bun run lint`, and `tsc --noEmit` all green.

### DIFF
--- a/src/up-next-core.ts
+++ b/src/up-next-core.ts
@@ -63,6 +63,10 @@ export interface EpicUnitInput {
   parentUrl: string;
   parentCreatedAt: number;
   parentLabels: string[];
+  /** The epic parent's assignees. The unit is filtered by these (#824) — the candidate child
+   *  is synthesized with `assignees: []`, so the parent is the assignee-bearing issue (matching
+   *  what the Backlog hides). */
+  parentAssignees: string[];
   /** All epic member issue numbers — removed from the flat per-repo list (dedup). */
   memberNumbers: number[];
   /** selectEpicCandidates()[0], or null when no child is actionable (suppress the unit). */
@@ -74,6 +78,10 @@ export interface RepoInput {
   repoSlug: string | null;
   repoLabel: string;
   lastUsedAt: number | null;
+  /** The operator's own login on this repo's forge (`forge.currentUser()`), or null when it
+   *  can't be resolved (local forge / un-authed / no identity API). Drives the "mine &
+   *  unassigned" filter (#824); null fails open (no assignee filtering — show everything). */
+  viewer: string | null;
   openIssues: Issue[];
   epics: EpicUnitInput[];
   /** All native sub-issue numbers in the repo (from listSubIssueSummaries). Removed from the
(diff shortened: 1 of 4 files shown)


scoop (commented):
## Summary

The PR makes Up Next honor issue assignees: it now hides issues assigned solely to other people while keeping unassigned issues and issues assigned to the viewer — exactly what the task asks for, applied by default.

The change is small, well-factored, and matches the established Backlog semantics.

## Verification

- **Filter semantics mirror the Backlog.** `isAssignedToOthers` (src/up-next-core.ts:132) returns true (drop) iff `assignees.length > 0 && !assignees.includes(viewer)`. This is the exact logical negation of the Backlog keep-predicate `assignees.length === 0 || assignees.includes(viewer)` (verified against ui/src/lib/components/issues-panel.ts:50-51). Both fail open when `viewer == null`.
- **Fail-open path is sound.** `viewer = (await forge.currentUser?.().catch(() => null)) ?? null` (src/up-next.ts:144). When `currentUser` is undefined, optional chaining short-circuits the entire chain (including `.catch`), yielding `undefined` → `null` — no crash. Confirmed by the test at test/up-next.test.ts ("no currentUser → viewer null → fails open").
- **`Issue.assignees` is always populated server-side**, so `issue.assignees` cannot be undefined in the core: verified against src/forge/github.ts:404, src/forge/github.ts:445, and src/forge/gitea.ts:144/172 (each maps to a `string[]` via `?? []`), and the required field at src/forge/types.ts:24.
- **`currentUser` is legitimately optional** on the forge interface (src/forge/types.ts:355), so the `?.` guard is required and correct. `assignees`/`currentUser` are both plain logins (no `@` prefix), so `includes` compares like-for-like.
- **Epic units are keyed on the parent's assignees** (src/up-next-core.ts:174), populated from `parent.assignees` where `parent` is the resolved `Issue` (src/up-next.ts:189,198) — a real `Issue`, so `.assignees` exists. Filtering on the parent (not the synthesized candidate child, which carries `assignees: []`) matches how the Backlog hides the epic parent. The `epic unit: filtered by PARENT assignees` test covers this.
- **Priority items are filtered too.** Priority section is derived from the same `repoItems`→`all` list (src/up-next-core.ts:234-237), which already runs every issue through `standaloneItem`/`epicItem`, so no others'-assigned issue can leak in via the priority tier.
- **New required fields (`viewer`, `parentAssignees`) have no un-updated construction sites.** The only constructors are in src/up-next.ts (both updated) and the test helpers (both updated in the diff); grep for `RepoInput`/`EpicUnitInput` found no others.

No i18n/message-catalog changes are involved (server-side ranking module + service only).

## Notes (non-blocking)

- The filter is unconditional in Up Next, whereas the Backlog exposes an `enabled` toggle (ui/src/lib/components/issues-panel.ts:43). The task only requires the default behavior ("include unassigned and assigned to me by default"), which this delivers; a user toggle would be a separate enhancement, not a defect.

Nothing to change.

PR #1371, opened 2026-07-03
feat(plugins): install plugins from the UI (Settings → Plugins)

kai-osthoff:
## What & why

Turns the read-only **Settings → Plugins** panel into a lightweight manager: **install a plugin from a GitHub URL**, list every installed plugin, and uninstall — the same `git clone … ~/.shepherd/plugins/` + restart flow that was previously manual (per `docs/plugins.md`), now reachable from the UI without a terminal.

It adds **no capability a shell couldn't already do** — plugins still run in-process with full server privileges (unchanged trust model), so install is gated behind a confirm dialog. Loading stays **boot-only**, so an installed plugin is *pending restart* until Shepherd restarts.

## Server

New management routes under a reserved `manage` segment (behind operator auth + CSRF), matched **before** the plugin-dispatch handler so it can neither shadow nor be shadowed by a plugin's own routes:

- `GET /api/plugins/manage/installed` — on-disk scan: `{ id, name, version, folder, loaded, disabled, broken }` per folder. Null-registry-safe; follows symlinked installs; a folder with an invalid `plugin.json` is surfaced as a removable `broken` row.
- `POST /api/plugins/manage/install` `{ url }` — strict `hostname === "github.com"` (no look-alikes, no `user:pass@`), shallow `git clone` with `GIT_TERMINAL_PROMPT=0`/`GIT_CONFIG_NOSYSTEM=1` (fails fast, no hang), then validates the cloned manifest and rejects a folder-name collision, an id collision (loaded/installed **or** the reserved `manage`), or an `apiVersion` mismatch — removing the clone on any rejection.
- `DELETE /api/plugins/manage/installed/<folder>` — rejects `.`/`..`/separators and requires the resolved parent to equal `realpath(pluginsDir)`; **unlinks a symlinked install** (never `rm -rf` through it into the operator's real checkout).

## UI

- Plugins tab is **always visible** (so you can install the first plugin).
- Install field + trust-confirm modal (canonical dim+blur scrim); a **union-by-id** list of the live loaded plugins and the on-disk scan — *pending restart* / *disabled* / *unrecognized* / *loaded-but-removed* rows; per-plugin uninstall; a **reload-durable** "restart to apply" banner with a copy button. Degrades to the plain loaded list if the scan can't load.
- EN+DE messages; feature-announcement entry `v1.42.0`.

## Notes / limitations

- **v1 clones the repo only — no `bun install`.** A plugin that ships its own dependencies still needs a manual `bun install` in its folder before it loads (surfaced in the UI hint).
- An `apiVersion` mismatch is blocked at install (a deliberate, documented divergence stricter than the loader, which would otherwise surface it as an `errored` card); a manual `git clone` remains the escape hatch to inspect that state.

## Testing

- `test/plugins-manage.test.ts` (17) — URL validation, scan (loaded/disabled/broken/symlink/null-registry), install collision/apiVersion/cleanup, symlink-safe + traversal-safe uninstall.
- `test/plugins-routes.test.ts` — `manage/installed` wiring + reserved-segment precedence.
- `SettingsPluginsPanel.browser.test.ts` (10) — install section, rich/pending/broken rows, focus behaviour.
- Green: root `bun test ./test`, `ui bun run test` + `bun run check`, `check:i18n`, feature-catalog, branch-hygiene, fallow audit.


kai-osthoff (commented):
## Review
No blocking findings.

Verification notes:
- Server route contract: `src/server.ts` imports the management functions at `src/server.ts:185`, exposes `pluginsDir` on `AppDeps` at `src/server.ts:231`, matches the `/api/plugins/manage/*` GET/POST/DELETE routes at `src/server.ts:5152`, and dispatches `handlePluginManagement` before plugin route dispatch at `src/server.ts:5977`; auth/origin checks still run before route handlers at `src/server.ts:6077`.
- Install and uninstall safeguards: URL validation requires HTTPS GitHub without credentials and derives the target folder at `src/plugins/manage.ts:82`; cloning uses `git clone --depth 1 --` with prompts disabled at `src/plugins/manage.ts:42`; install rejects folder collisions, invalid manifests, API mismatches, reserved ids, and id collisions at `src/plugins/manage.ts:177`; uninstall rejects traversal/separators and unlinks symlinked installs instead of following them at `src/plugins/manage.ts:219`.
- UI/API wiring: Settings always includes the Plugins tab at `ui/src/lib/components/Settings.svelte:63` and renders the panel at `ui/src/lib/components/Settings.svelte:1527`; the panel loads the on-disk scan at `ui/src/lib/components/settings/SettingsPluginsPanel.svelte:34`, unions loaded and installed rows at `ui/src/lib/components/settings/SettingsPluginsPanel.svelte:60`, and calls install/uninstall at `ui/src/lib/components/settings/SettingsPluginsPanel.svelte:145`; the API client uses the matching `/api/plugins/manage/*` paths at `ui/src/lib/api.ts:1158`.
- i18n and feature discovery: the new plugin strings and feature keys exist in both English and German catalogs at `ui/messages/en.json:1203` and `ui/messages/de.json:1203`; the feature entry points at those keys and uses `sinceVersion: "1.42.0"` at `ui/src/lib/feature-announcements/entries/v1.42.0-plugin-install.ts:7`, which is greater than the last released `1.41.0` in `.release-please-manifest.json:2`.

I did not run builds or tests per the review instruction. I also could not verify live GitHub clone behavior against the network; the review only verifies the checked-in shell-out implementation at `src/plugins/manage.ts:42`.

kai-osthoff:
Oh wow, another *flawless* READY PR 😴✨🎀🚀 @scoop merge it already 🙄💅🔥

PR #1384, opened 2026-07-03
feat(ui): files tab worktree/scratchpad source switch

scoop:
## What

Adds a **Scratchpad | Worktree** segmented switch inside the Files tab so an operator can browse either the session's scratchpad (unchanged default) **or** its git worktree, without leaving the tab.

- **Scratchpad** (default): today's behavior — browse, download, upload.
- **Worktree** (new): **read-only** browse + download of the session's working tree. `.git` is hidden; symlinks that resolve outside the worktree are shown as disabled, non-navigable rows so the tree isn't misleadingly incomplete.

## How

**Server**
- Extracted the scratchpad's realpath-containment browsing into a shared core `src/fs-browse.ts` (`resolveInRoot` / `listDir` / `resolveFileInRoot`, with `hideSegment` + `onEscape` options). `src/scratchpad.ts` now delegates to it — public API and behavior unchanged.
- New `src/worktree-files.ts` (thin wrappers: hide `.git` by name, mark out-of-root symlinks) and two read-only routes in `src/server.ts`, mirroring the scratchpad routes:
  - `GET /api/sessions/:id/worktree[?path=]` (list)
  - `GET /api/sessions/:id/worktree/download?path=` (file bytes)
- Containment is enforced by realpath: `..`, absolute paths, and symlink escapes are all rejected; `.git` is rejected on **both** listing and download. A `worktreePath`-empty guard prevents any cwd fallback.

**UI**
- `FilesPanel.svelte` gains the `.seg-row` switch (canonical design-system recipe, 44px targets), source-aware labels, and a read-only mode (no upload button, no drop zone) when Worktree is active. `linkOutside` entries render disabled.
- `getWorktreeListing` / `worktreeDownloadUrl` in `api.ts`; `linkOutside?: boolean` added to the `ScratchEntry` type; demo mock router gets a `/worktree` route so the public demo doesn't crash on switch.

## Resolved decisions

- Worktree is **read-only** (no upload into a git working tree).
- Filter **hides `.git` only** (file, dir, or submodule); dotfiles and `node_modules` are shown (lazy, per-directory).
- Default source is **Scratchpad** — existing behavior is untouched.
- Download is **unfiltered** for non-`.git` files (incl. `.env`/secrets): Shepherd is a localhost single-operator tool where the operator already has shell read access to the worktree, and the scratchpad download already streams arbitrary bytes — so this is not a new disclosure boundary.

## Accepted-risk note

`.git` is hidden by path-segment name. A hypothetical symlink *inside* the worktree named e.g. `foo` pointing at `.git/config` would resolve within-root and be downloadable — but that is functionally the operator reading their own `.git` under a different name (they have shell access), and is within the same accepted risk as the unfiltered-download decision above. Not hardened by design.

## Feature discovery / i18n

- Adds `ui/src/lib/feature-announcements/entries/v1.42.0-worktree-files.ts` (What's-New entry).
- All new UI strings added to `en.json` + `de.json` (parity enforced by `check:i18n`).

## Testing

- New: `test/fs-browse.test.ts`, `test/worktree-files.test.ts`, `test/worktree-endpoint.test.ts` (containment, `.git`-as-file + submodule hiding, out-of-root symlink marking, escape rejection, endpoint wiring). Extended `FilesPanel.browser.test.ts` (switch fetches worktree, upload hidden, disabled link row) and `demo/router.test.ts`.
- Green: root `bun run lint` + `bun run test` (5777 pass / 0 fail); UI `bun run check` (0 errors) + `check:i18n` (parity) + `bun run test` (2753 pass); `fallow@2.100.0 audit` exit 0; pre-push all lanes.

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -276,6 +276,14 @@
         "maxCognitive": 20,
         "maxCrap": 1000,
         "reason": "Codex/default-provider persistence extends the existing hydrate migration path by one branch; keep capped until hydrate migrations are extracted."
+      },
+      {
+        "files": ["ui/src/lib/demo/router.ts"],
+        "functions": ["handleSessionDetailGet"],
+        "maxCyclomatic": 20,
+        "maxCognitive": 20,
+        "maxCrap": 1000,
+        "reason": "Demo session-detail GET dispatch chain gained the /worktree route (one branch over the cognitive default); keep capped until the GET routes move to the data-driven table the mutations already use (#855)."
       }
     ],
 
--- a/docs/external-task-api.md
+++ b/docs/external-task-api.md
@@ -130,6 +130,13 @@ Once a task exists, an external agent can also drive it:
   streams a single file. Paths are relative to the scratchpad root and
   realpath-contained to it (`..`, absolute, and symlink escapes are rejected);
   both `404` on a missing/archived session.
+- `GET /api/sessions/:id/worktree[?path=]` — browse a live session's git
+  worktree subtree (read-only); `GET /api/sessions/:id/worktree/download?path=`
+  streams a single file. Paths are relative to the worktree root and
+  realpath-contained to it. `.git` is hidden at any level, and symlinks that
+  resolve outside the worktree are surfaced as non-navigable `linkOutside`
+  entries rather than dropped. There is no worktree upload route; both `404` on
+  a missing/archived session.
 - `POST /api/sessions/:id/scratchpad/upload[?path=<relDir>]` — upload an
   arbitrary binary file (multipart `file` field, no MIME restriction) into the
   session's scratchpad. `?path` selects a relative subdirectory within the
--- a/src/fs-browse.ts
+++ b/src/fs-browse.ts
@@ -1,4 +1,4 @@
-import { promises as fsp } from "node:fs";
+import { promises as fsp, type Dirent } from "node:fs";
 import { join, relative, dirname, sep, normalize, isAbsolute } from "node:path";
 
 /**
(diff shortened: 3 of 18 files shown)


scoop (commented):
## Summary

This PR adds a Scratchpad/Worktree segmented switch to the Files tab. It (1) extracts the scratchpad's realpath-containment core into a reusable `src/fs-browse.ts`, (2) adds a `.git`-hidden, `linkOutside`-marking worktree browser (`src/worktree-files.ts`) with `GET /api/sessions/:id/worktree[/download]` routes, and (3) wires a source toggle into `FilesPanel.svelte`. It reads clean and is thoroughly tested.

I verified the correctness-relevant claims below against ground truth.

### Verified

- **i18n parity** — all 11 new keys exist in BOTH catalogs with matching names: `files_source_switch_aria`, `files_source_scratchpad`, `files_source_worktree`, `files_worktree_root_crumb`, `files_worktree_breadcrumb_aria`, `files_worktree_empty`, `files_worktree_load_error`, `files_worktree_list_aria`, `files_link_outside_title`, `feat_worktree_files_title`, `feat_worktree_files_body` (ui/messages/en.json:2476-2487 vs ui/messages/de.json:2474-2485). Every `m.*` key referenced in FilesPanel.svelte:56-61,153,192,203 resolves. Breadcrumb-root casing (`files_worktree_root_crumb` = "worktree" lowercase in en, "Worktree" in de) matches the existing `files_root_crumb` pattern ("scratchpad"/"Scratchpad", en.json:2093 / de.json:2093).
- **Route wiring** — `handleSessionWorktree` uses the exact indexing/guard pattern of `handleSessionScratchpad` (parts[3]==="worktree", parts[2]=id, archived→404), and is registered in the handler array (src/server.ts:3126). `attachmentDisposition`/`basename` are imported (src/server.ts:80,156).
- **API contract** — `getWorktreeListing`/`worktreeDownloadUrl` (ui/src/lib/api.ts:619,628) return/build the same shapes as the scratchpad equivalents; `ScratchListing` gains an optional `linkOutside` (ui/src/lib/types.ts:158) consumed by FilesPanel.svelte:192. Test expectation `getWorktreeListing("sess-7", undefined)` matches `browse("")` → `path || undefined` (FilesPanel.svelte:44).
- **Containment refactor is behavior-preserving** — `resolveInRoot`/`listDir`/`resolveFileInRoot` reproduce the old scratchpad logic; the `if (!claudeSessionId) return null` guard is retained on all three scratchpad wrappers (src/scratchpad.ts:32,50,63). The `!worktreePath` guard prevents a `realpath(".")` fallback (src/worktree-files.ts:11,23).
- **Feature announcement** — auto-registered via `import.meta.glob` (feature-announcements.ts), `sinceVersion: "1.42.0"` matches the other unreleased 1.42.0 entries; no manual index edit needed.
- **Reactivity** — the `untrack` wrapper correctly keeps the reset/reload effect keyed on `sessionId` only, so flipping the source doesn't re-fire it (FilesPanel.svelte:53-63); `switchSource` clears `listing`/`uploads` and `browse()` resets `loading`/`error`, so no stale-source flash.
- **No broken consumers** — the removed `ScratchEntry` interface in src/scratchpad.ts has no importers (only ui/src/lib/types.ts defines its own copy).

No blocking issues found.

### Non-blocking observations (informational)

- `src/fs-browse.ts`: the refactor's `classifyDirent` no longer surfaces sockets/fifos/devices, whereas the old `listScratchpad` stat'd every entry and surfaced non-dirs as `"file"`. This is a benign (arguably more correct) behavior change for the scratchpad; no realistic scratchpad content is affected.
- `src/server.ts`: `worktreeList` has no synthetic-empty fallback for a missing root, unlike `scratchpadList` (server.ts:1955) which returns `{entries: []}` when the agent hasn't written yet. This is justified — a live session's worktree is a git checkout that exists from the start, so there's no equivalent "not created yet" window; a genuinely missing/uncreated worktree correctly surfaces the load-error state.
- `ui/src/lib/components/viewport/FilesPanel.svelte:250`: the empty-state div keeps the `empty-droppable` class in worktree mode, but the drag handlers early-return on `readOnly`, so it's inert — cosmetic only.

- None.

scoop (commented):
## Summary

Adds a Scratchpad/Worktree segmented switch to the Files tab, backed by a new read-only `GET /api/sessions/:id/worktree[?path=]` + `/download` endpoint. A generic containment core is extracted into `src/fs-browse.ts` and shared by both the scratchpad browser (refactor of `src/scratchpad.ts`) and the new worktree browser (`src/worktree-files.ts`). The change is well-scoped, well-tested, and satisfies the task.

## What I verified

- **Route dispatch is disjoint.** `handleSessionWorktree` guards `parts[3] === "worktree"` (`src/server.ts:2001`) while `handleSessionScratchpad` guards `parts[3] !== "scratchpad"` (`src/server.ts:1961`); neither shadows the other, and both return `null` when unmatched. Archived/unknown sessions 404 (`src/server.ts:2003`), matching the docs and `test/worktree-endpoint.test.ts`.
- **Containment refactor is behavior-preserving.** `resolveInRoot`/`listDir`/`resolveFileInRoot` (`src/fs-browse.ts`) reproduce the prior scratchpad logic (`..`/absolute rejection, realpath containment, dirs-before-files sort, drop-on-escape) and add `hideSegment` + `onEscape:"mark"`; `src/scratchpad.ts` now just binds them to the scratchpad root. The empty-`claudeSessionId` guard is preserved in each scratchpad wrapper.
- **`.git` hiding is by name at any depth** (`src/worktree-files.ts:4`), covering the dir, the `git worktree add` gitdir-pointer file, and nested submodule `.git` — confirmed by `test/worktree-files.test.ts` (root + `vendor/lib`, and the mid-path `resolveWorktreeFile` rejection).
- **Blank-`worktreePath` guard** returns `null` rather than falling back to `realpath(".")` (`src/worktree-files.ts:12,23`), verified in `test/worktree-files.test.ts:100-108`.
- **i18n parity.** All 11 new keys are present in BOTH catalogs (`ui/messages/en.json:842-852`, `ui/messages/de.json:820-830`), and the reused pre-existing keys the derived labels fall back to exist in de (`ui/messages/de.json:2093-2096,2276,2283`). The feature entry auto-registers via `import.meta.glob` (`ui/src/lib/feature-announcements.ts:42`), and its `titleKey`/`bodyKey` match the catalog keys.
- **Effect vs. switch reactivity.** The session-change `$effect` wraps the reset in `untrack` (`FilesPanel.svelte:67-77`) so flipping `source` does not re-fire it; `source` is read synchronously (before the `await`) inside `browse()` (`FilesPanel.svelte:41-44`), so each request is bound to the source in effect at call time. Read-only mode correctly hides upload/drag paths and renders `linkOutside` rows as disabled non-anchors — covered by `FilesPanel.browser.test.ts`.
- **Type change is additive** (`ui/src/lib/types.ts:156` optional `linkOutside`), so existing `ScratchListing`/`ScratchEntry` consumers are unaffected.

## Non-blocking nit

`FilesPanel.svelte` `browse()` has no request-sequencing guard, so two overlapping browses can resolve out of order and leave `listing` inconsistent with the active `source`/`path` (e.g. click Worktree then quickly Scratchpad: a slow worktree response can land after the scratchpad one, showing worktree paths under scratchpad labels/`downloadUrl`). `switchSource` nulls `listing` to avoid the flash but doesn't cover async ordering. This is the same class of race the pre-existing scratchpad directory navigation already had (local API calls make it rare); worth a tiny request-token if you touch this again.

- src/worktree-files.ts: `.git` hiding is enforced only against the caller-supplied relative segments, not the realpath; an in-worktree symlink named e.g. `x` pointing at `.git/config` would pass `hideSegment` and stay within-root, so its bytes are downloadable. This is a decluttering measure over the operator's own checkout (not a trust boundary — `.env` and other secrets are intentionally listed), so it is not a security defect, but the `.git` guard is weaker than a realpath-segment check would be.

PR #1385, opened 2026-07-03
fix(ui): reorder new-task toggles, lock plan-gate/autopilot while Research on

scoop:
## What

In the new-task pane's run settings:

- **Reordered** the toggles to **Research → Plan gate → Autopilot** (was Plan gate → Autopilot → Research). The mode switch now leads the stack it governs.
- **Visible exclusivity**: while Research is checked, Plan gate + Autopilot render **disabled** (0.6 muting, `not-allowed` cursor) instead of silently unchecking yet staying re-checkable. The *reason* is folded into the existing Research InfoTip — extended `newtask_research_hint` EN+DE, no new standing note (the phone sheet stays compact; it already carries the sandbox note).
- **Fixes an asymmetry bug**: checking Autopilot previously did **not** clear Research, so `research: true` + `autopilotEnabled: true` could be submitted — a combo the Research handler forbids. Disabling the row while Research is on removes the path structurally. The now-unreachable `research = false` in the Plan gate handler is deleted.
- Cursor semantics: `not-allowed` = permanent research lock, `progress` = transient loading; the lock rule is scoped `.plan-gate.research-locked:has(input:disabled)` so it deterministically outranks the loading rule.

## Verification

- `cd ui && bun run check` — 0 errors (1 pre-existing CommandBar warning, untouched)
- `bun run check:i18n` — 2 locales in parity
- `bun run test` — 205 files, 2761 tests pass (updated the plan-gate-clears-Research test to the new disabled contract; fixed a first-`.plan-gate`-input selector that grabbed Research after the reorder)
- **In-browser** (demo mode): row order confirmed; checking Research unchecks + disables both rows with computed `opacity: 0.6` and `cursor: not-allowed` (asserted via `getComputedStyle`); InfoTip shows the extended hint; unchecking re-enables both.

Screenshot (Research checked, InfoTip open, rows locked):

_locked state: Research ✔, Plan gate + Autopilot muted/disabled, InfoTip explains why_

[no-feature-entry]

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -1291,7 +1291,7 @@
   "feat_session_recap_title": "Sitzungszusammenfassung",
   "feat_session_recap_body": "Shepherd erstellt nach jeder Sitzung eine kompakte Zusammenfassung — mit Ergebnis, Überschrift und offenen Punkten — damit du auf einen Blick siehst, was passiert ist.",
   "newtask_research_label": "Recherche-Aufgabe",
-  "newtask_research_hint": "Webrecherche → ein Bericht-PR oder ein GitHub-Issue. Keine Code-Änderungen.",
+  "newtask_research_hint": "Webrecherche → ein Bericht-PR oder ein GitHub-Issue. Keine Code-Änderungen. Plan-Gate und Autopilot gelten hier nicht und sind deaktiviert.",
   "newtask_research_sandbox_note": "Recherche-Aufgaben können das autonome Profil nicht verwenden — Websuche benötigt offenen Netzwerkzugang.",
   "research_badge_label": "Recherche",
   "research_badge_title": "Rechercheaufgabe: Web-Recherche → ein Bericht-PR oder ein GitHub-Issue",
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -1292,6 +1292,7 @@
   "feat_session_recap_body": "Shepherd erstellt nach jeder Sitzung eine kompakte Zusammenfassung — mit Ergebnis, Überschrift und offenen Punkten — damit du auf einen Blick siehst, was passiert ist.",
   "newtask_research_label": "Recherche-Aufgabe",
   "newtask_research_hint": "Webrecherche → ein Bericht-PR oder ein GitHub-Issue. Keine Code-Änderungen. Plan-Gate und Autopilot gelten hier nicht und sind deaktiviert.",
+  "newtask_research_locked_aria": "Für Recherche-Aufgaben nicht verfügbar.",
   "newtask_research_sandbox_note": "Recherche-Aufgaben können das autonome Profil nicht verwenden — Websuche benötigt offenen Netzwerkzugang.",
   "research_badge_label": "Recherche",
   "research_badge_title": "Rechercheaufgabe: Web-Recherche → ein Bericht-PR oder ein GitHub-Issue",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -1291,7 +1291,7 @@
   "feat_session_recap_title": "Session recap",
   "feat_session_recap_body": "Shepherd now generates a concise recap after each session — verdict, headline, and open items — so you can see what happened at a glance.",
   "newtask_research_label": "Research task",
-  "newtask_research_hint": "Web research → a report PR or a GitHub issue. No code changes.",
+  "newtask_research_hint": "Web research → a report PR or a GitHub issue. No code changes. Plan gate and autopilot don't apply and are disabled.",
   "newtask_research_sandbox_note": "Research can't use the autonomous profile — web search needs open network access.",
   "research_badge_label": "Research",
   "research_badge_title": "Research task: web research → a report PR or a GitHub issue",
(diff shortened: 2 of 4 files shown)


scoop (commented):
## Summary

The PR does exactly what the task asks: the toggle stack is reordered to **Research → Plan → Autopilot** (`ui/src/lib/components/new-task/NewTaskRunSettings.svelte:76-161`). It also lands a genuine UX improvement alongside the reorder: instead of the old silent mutual-exclusivity (checking Plan gate quietly unchecked Research), Plan gate and Autopilot now render **disabled/muted while Research is on**, which reads as intentional exclusivity rather than a value silently flipping.

## Verification (ground truth checked)

- **Reorder** — Research label block moved to first (`NewTaskRunSettings.svelte:76-99`), Plan gate second (`:104-124`), Autopilot third (`:130-161`). Matches the requested order.
- **Clear-and-lock logic** — Research `onchange` sets `planGate=false`/`autopilot=false` and calls `onPlanGateTouched()`/`onAutopilotTouched()` (`:82-90`). The parent seeds those values from repo config only when *not* touched (`ui/src/lib/components/NewTask.svelte:366-367,375-376`), so pinning touched prevents a later repo switch from re-enabling them under an active Research task. Both inputs also carry `disabled={... || research}` (`:113,:140`). The removal of the old `if (planGate) research = false` guard (`:110-111`) is safe — Plan gate can no longer be toggled while Research is on.
- **CSS specificity** — the comment at `:355-357` is correct: `.plan-gate.research-locked:has(input:disabled)` (0,3,1) outranks `.plan-gate:has(input:disabled)` (0,2,1), so `cursor:not-allowed` wins while the base rule still applies `opacity:0.6` muting. Verified against `:351-360`.
- **i18n parity** — the repo maintains only `en`+`de`; both `newtask_research_hint` entries got the same appended sentence (`ui/messages/en.json:1294`, `ui/messages/de.json:1294`). No orphaned key.
- **No DOM-order coupling** — `coachTarget` registrations are name-based, not positional, and the retained targets stayed on the correct rows (`plan-gate` at `:105`, `task-autopilot` at `:132`). The only `.plan-gate input` index queries live in the test file, which was updated to match-by-label (`ui/src/lib/components/NewTask.browser.test.ts:253-256`), and the replaced test now asserts the disable/re-enable behavior (`:556-570`).

## Non-blocking nit

- `NewTaskRunSettings.svelte:104-160`: while Research is on, the Plan gate / Autopilot inputs go disabled+muted but the *reason* lives only in the Research InfoTip. A screen-reader user landing on the disabled checkbox hears "disabled" with no explanation. Consider a `title`/`aria-describedby` on the disabled inputs (or reusing the research hint) so the lock reason is conveyed in place. Minor — the visual muting + Research InfoTip cover sighted users.

- ui/src/lib/components/NewTask.svelte: if a task opens with `initialResearch` true and the repo's autopilot/plan-gate default is on, those values are seeded true at mount (the Research `onchange` that clears them only fires on user interaction), so a research task could submit `autopilotEnabled/planGateEnabled: true` alongside `research: true`. Predates this PR (init seeding unchanged) and likely ignored server-side for research kind; worth a look separately.
- ui/src/lib/components/new-task/NewTaskRunSettings.svelte:149-155: the autopilot `repo-default` pill still renders while the row is research-locked, which can read as slightly contradictory next to a forced-off control — cosmetic, not introduced by this PR's core change.

scoop (commented):
## Summary

Reorders the new-task run-settings toggles to Research → Plan gate → Autopilot, and reframes the Research↔Plan/Autopilot relationship from silent mutual-exclusion into a visible lock: while Research is on, Plan gate and Autopilot render disabled/muted. The re-review's non-blocking accessibility point is addressed.

## Task satisfaction

Requested order (Research, Plan, Autopilot) is implemented — Research block at `ui/src/lib/components/new-task/NewTaskRunSettings.svelte:76`, Plan gate at `:109`, Autopilot at `:136`. Verified.

The accompanying UX improvement (disabling rather than silently re-checking Plan/Autopilot during Research) is a reasonable, self-consistent enhancement.

## Re-review point

1. Disabled Plan gate / Autopilot conveying no reason to assistive tech — **addressed**. `NewTaskRunSettings.svelte:105-107` renders a `.sr-only` span `id="nt-research-locked-note"` only when `research` is true; both disabled inputs reference it via `aria-describedby={research ? "nt-research-locked-note" : undefined}` (`:119`, `:147`). The reference and the target are both gated on the same `research` condition, so there is no dangling `aria-describedby`. The `.sr-only` recipe (`:377`) matches the standard visually-hidden pattern.

## Verification

- i18n parity: `newtask_research_locked_aria` present in both catalogs — verified against `ui/messages/en.json:1295` and `ui/messages/de.json:1295`. Updated `newtask_research_hint` present in both (`en.json:1294`, `de.json:1294`).
- Message helper `m.newtask_research_locked_aria()` used at `:106` resolves to those keys; `m.newtask_info_aria` / `m.newtask_research_label` reused unchanged.
- CSS specificity: `.plan-gate.research-locked:has(input:disabled)` (`:365`, cursor:not-allowed) outranks `.plan-gate:has(input:disabled)` (`:358`, cursor:progress) by one extra class — the lock-cursor claim in the comment is correct. The base rule's `opacity: 0.6` still applies to research-locked rows (intended muted look). `:has()` was already used pre-existing at `:358`, so no new browser-support surface.
- One-way `checked={planGate}` binding reflects the `planGate = false` write in the Research onchange handler (`:83`) reactively — verified no stale-checkbox path.
- Tests updated to locate rows by label text (`NewTask.browser.test.ts:449-460`), making them order-independent; the replaced test now asserts the lock/unlock behavior.
- Grepped for stale references to the old `if (planGate) research = false` mutual-exclusion logic — none remain outside this component.

No blocking issues found.

PR #1386, opened 2026-07-03
fix(compose): send mode=partial for live-preview transcriptions

kai-osthoff:
Companion to voice-whisper plugin [#4](https://github.com/erwins-enkel/shepherd-plugin-voice-whisper/pull/4).

The live read-along (shipped in #1379) fires a transcription every ~2.5s while dictating. This tags those requests `mode=partial` (and the final stop-transcription `mode=final`) so the plugin's concurrency gate can **reserve a slot for the final** — a burst of disposable preview requests can then never `429`-starve the transcription the user actually keeps.

- `transcribeAudio(blob, lang, mode?)` gains an optional `mode` form field.
- `ComposeBar`: interim posts send `partial`, the final sends `final`.

Verified: `bun run check` (svelte-check) 0 errors.

### DIFF
--- a/ui/src/lib/api.ts
+++ b/ui/src/lib/api.ts
@@ -333,10 +333,17 @@ export function getVoiceStatus(): Promise<VoiceStatus> {
 
 /** Transcribe a recorded audio clip via the voice-whisper plugin; returns the text. `lang`
  *  (`"de"`/`"en"`) pins the transcription language when the plugin isn't configured to force one. */
-export async function transcribeAudio(blob: Blob, lang?: string): Promise<string> {
+export async function transcribeAudio(
+  blob: Blob,
+  lang?: string,
+  mode?: "partial" | "final",
+): Promise<string> {
   const fd = new FormData();
   fd.append("file", blob, "clip.webm");
   if (lang) fd.append("lang", lang);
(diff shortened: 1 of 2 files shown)


scoop:
Closing as **superseded by #1454** (`d845cf266`) — not as wrong-headed.

The diagnosis here was correct, and the merged version credits the same reasoning: the plugin's load gate keeps a concurrency slot free for the final clip, so live-preview requests must be tagged disposable. #1454 shipped that in a better shape:

- interim requests are tagged at `ui/src/lib/dictation.svelte.ts:204-206` (`mode: "partial"`), in the shared dictation controller that every mic now routes through — rather than in `ComposeBar` alone;
- `transcribeAudio` takes an `opts` object (`ui/src/lib/api.ts:456`) with the type deliberately narrowed to `mode?: "partial"`;
- it ships `ui/src/lib/api.transcribeAudio.test.ts` with 4 cases, including the 429-shed path.

One concrete reason not to rebase this branch: it sends an explicit `mode: "final"`, but main's contract is **absent ⇒ final**, and `api.transcribeAudio.test.ts:23` asserts exactly that ("sends NO mode field for the final clip"). Landing it would regress a tested contract, and `"final"` is corroborated nowhere on the plugin side.

Reviewed as part of a merge train over #1926 / #1510 / #1491 / #1386.

PR #738, opened 2026-06-16
feat(hooks): reach Shepherd from inside the autonomous egress netns (#711)

scoop:
Closes #711.

## Problem
Under the `autonomous` sandbox profile the agent runs in a slirp4netns network namespace, where `127.0.0.1` is the *namespace* loopback — not the host's. So the injected push hooks (#704/#708) and the build-queue curl, both baked with `http://127.0.0.1:<port>`, can't reach Shepherd there and silently fall back to polling — exactly where (unattended drain) push signals help most.

## Fix (least-privilege, fail-open, no exfil widening)
The agent→Shepherd control-plane channel is made reachable from inside the autonomous netns via the slirp host gateway `10.0.2.2` (→ host `127.0.0.1`). Covers **both** hooks and the build-queue curl (one `agentBaseUrl` source).

1. **Reachability.** `scripts/egress-runner.sh` drops `--disable-host-loopback` so the netns can route to host loopback at `10.0.2.2`; `buildEgressConfig` emits exactly one nft allow rule `ip daddr 10.0.2.2 tcp dport <ingress-port> accept`. Everything else stays `policy drop`.
2. **Containment — the crux.** `isAuthorized` returns `true` when no token is set (the default), and slirp NATs the netns→`10.0.2.2` connection to host `127.0.0.1` (indistinguishable from a real loopback client). Allowlisting the **main** port would hand a contained agent the entire control plane (`POST /api/sessions` → spawn a `trusted` host-net session = full firewall escape). Instead the netns reaches a **separate, route-restricted ingress listener** (`serveAgentIngress`, bound to `127.0.0.1` only) that serves **only** `POST .../hooks`, `PUT .../queue`, `POST .../queue/steps/:id` (404s everything else, incl. `POST /api/sessions` and `/queue/approve`) and **delegates** allowed routes through the full app so `checkAuth` + `checkOrigin` still apply. The per-session UUID in each path is the de-facto per-session capability (the agent only knows its own; no enumeration route). The main control-plane port is never in the nft allowlist — a hijacked autonomous agent cannot spawn/halt/merge.
3. **No-drift invariant.** The baked base URL and the nft host-gateway rule derive from one source (`egressHostGateway` over the shared `willEgressConfine` predicate + cached `detectEgressHostLoopback` + the `agentIngressPort` accessor), so they cannot diverge.
[...]

PR #1392, opened 2026-07-03
fix(server): allowlist GET /api/sessions/:id/queue on agent ingress

scoop:
## Problem

Multiple sessions reported the build-queue endpoint as broken: `GET /api/sessions/:id/queue` returned `404 {"error":"not found"}`, so agents concluded the endpoint "isn't active for this session" and abandoned the build queue entirely (falling back to plan files, leaving the operator's queue view empty).

## Root cause

Agents reach Shepherd through the restricted **agent-ingress** listener, gated by `isAgentIngressRoute`. That allowlist (introduced in #738) permits only `POST /hooks`, `PUT /queue`, and `POST /queue/steps/:id` — but the build-queue spawn directive (`buildQueueDirective` in `src/service.ts`) explicitly instructs agents to **GET** the queue ("Inspect the current queue at any time", "re-GET … to pick up the current ids"). The read route was never added to the allowlist, so the gate 404s every GET with `{"error":"not found"}` — reproduced live against a running session before the fix.

The 404 is what turns a transient hiccup into a permanent failure: GET is the agent's advertised diagnostic/recovery path, and it lies. In the reported transcript the agent's initial PUT failed for an unrelated, agent-side reason (curl exit 3 = malformed URL — the multi-line command was mangled; the plain-text `Method Not Allowed` matches no Shepherd handler, which all return lowercase JSON), then the follow-up GET probe hit this gate 404 and "confirmed" the endpoint was dead.

## Fix

Allowlist `GET /api/sessions/:id/queue` on the ingress. Read-only and strictly weaker than the already-allowed PUT on the same path; the per-session UUID capability model is unchanged and `POST /queue/approve` (the human gate) stays excluded.

## Tests

- Allow-table now asserts GET `/queue` is permitted; deny-table keeps DELETE `/queue`, GET `/queue/steps/:id`, and all prior containment cases.
- New app-level regression test: an agent PUTs its queue, then GETs it back through `makeAgentIngressApp` (was 404, now 200 echoing the steps).
- `bun run lint` clean; full root suite 5808 pass / 0 fail.

Takes effect for newly-observed traffic on the next server restart/deploy (no config or data migration).

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -6195,8 +6195,9 @@ export function isAgentIngressRoute(method: string, parts: string[]): boolean {
   if (!(parts[0] === "api" && parts[1] === "sessions" && parts[2])) return false;
   // POST /api/sessions/<id>/hooks
   if (method === "POST" && parts[3] === "hooks" && !parts[4]) return true;
-  // PUT /api/sessions/<id>/queue
-  if (method === "PUT" && parts[3] === "queue" && !parts[4]) return true;
+  // PUT /api/sessions/<id>/queue (author/replace) + GET (the directive's "inspect the current
+  // queue" / re-GET-for-ids path — read-only, strictly weaker than the PUT already allowed).
+  if ((method === "PUT" || method === "GET") && parts[3] === "queue" && !parts[4]) return true;
   // POST /api/sessions/<id>/queue/steps/<stepId>
   if (method === "POST" && parts[3] === "queue" && parts[4] === "steps" && parts[5] && !parts[6]) {
     return true;
--- a/test/agent-ingress.test.ts
+++ b/test/agent-ingress.test.ts
@@ -19,9 +19,10 @@ const SID = "step_abc";
 const parts = (p: string) => p.split("/").filter(Boolean);
 
 // ── isAgentIngressRoute: exhaustive allow/deny table ────────────────────────────
-test("isAgentIngressRoute: ALLOWS exactly the three agent→server routes", () => {
+test("isAgentIngressRoute: ALLOWS exactly the four agent→server routes", () => {
   expect(isAgentIngressRoute("POST", parts(`/api/sessions/${ID}/hooks`))).toBe(true);
   expect(isAgentIngressRoute("PUT", parts(`/api/sessions/${ID}/queue`))).toBe(true);
+  expect(isAgentIngressRoute("GET", parts(`/api/sessions/${ID}/queue`))).toBe(true);
   expect(isAgentIngressRoute("POST", parts(`/api/sessions/${ID}/queue/steps/${SID}`))).toBe(true);
 });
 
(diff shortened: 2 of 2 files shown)


scoop (commented):
## Review: allowlist GET /api/sessions/:id/queue on agent ingress

### Root-cause verification — the fix matches the observed failure

The transcript shows the agent's `GET http://127.0.0.1:7331/api/sessions/<id>/queue` returning `{"error": "not found"}` with HTTP 404. That is exactly the agent-ingress gate's deny response: `makeAgentIngressApp` returns `json({ error: "not found" }, 404)` for any non-allowlisted route (verified against src/server.ts:6221). Before this PR, `isAgentIngressRoute` allowed only PUT on `/queue`, so the GET was denied at the gate.

The GET is not optional for agents — the spawn directive explicitly instructs it, twice:
- "Inspect the current queue at any time: `curl -s${authHeader} ${queueUrl}`" (verified against src/service.ts:682-683)
- the re-GET-for-ids fallback: "re-GET (or read the PUT response) to pick up the current ids before posting status" (verified against src/service.ts:678)

And all agents are routed through the ingress listener: `resolveSpawnBaseUrl` bakes the ingress base URL into every spawn (loopback for trusted/standard, host-gateway for egress-confined autonomous; verified against src/service.ts:1403-1432, src/index.ts:2364-2366). So the directive told agents to make a request the gate rejected — a genuine directive/allowlist inconsistency, and this PR closes it at the right place.

### Correctness of the new allow rule

- The delegated GET reaches a real handler: `handleBuildQueue` is registered before `handleSessions` in `ROUTE_HANDLERS` (verified against src/server.ts:6087-6088) and serves `GET` + `!parts[4]` via `deps.store.getBuildQueue(id)` (verified against src/server.ts:3106). No earlier handler claims the path — the three in-repo 405 sites own `/api/sessions/clear-merged`, `/api/diagnostics/fix`, and `/api/halt` (verified against src/server.ts:2055-2062, 3599-3602, 4413-4416).
- Empty-queue safety: `getBuildQueue` returns `{sessionId, steps: [], approved: false}` when no rows exist, never throws (verified against src/store.ts:3969-3998), so the "inspect at any time" GET is safe even before the first PUT.
- Security posture: GET is strictly weaker than the already-allowed PUT — `putBuildQueue` already echoes the full queue in its response (verified against src/server.ts:3015), so GET exposes nothing new to a holder of the per-session UUID capability. The gate rule requires `!parts[4]`, so `GET /queue/approve` and `GET /queue/steps/:id` remain denied (verified against src/server.ts:6200; asserted in test/agent-ingress.test.ts:31,43).
- Tests: the allow-table, containment table, and a new end-to-end PUT-then-GET delegation test through `makeAgentIngressApp` were all updated coherently (test/agent-ingress.test.ts:22-27, 29-59, 172-200), including new deny assertions for `DELETE /queue` and `GET /queue/steps/:id` replacing the removed `GET /queue → false` line.

### Observation on the transcript's PUT failure (limitation, not a finding)

The transcript's first failure — `PUT` → plain-text `Method Not Allowed`, curl exit code 3 — is not producible by this server: every in-repo 405 is JSON lowercase `{"error": "method not allowed"}` (verified against src/server.ts:2062, 3602, 4416), and none of them owns the queue path. Curl exit 3 means "malformed URL", so the agent's multi-line curl was likely mangled and (part of) it hit something other than the ingress route; I cannot verify what produced that response body — it is not in this repo. This PR does not fix that agent-side mangling, but it fixes the follow-up: the agent's diagnostic GET now returns the queue (200) instead of 404, so a transient PUT failure no longer cascades into "the whole build-queue endpoint is dead, abandon the queue" — which is the repeated-across-sessions symptom the task describes.

PR #1395, opened 2026-07-03
fix(readiness): prescribe stack-correct dependabot ecosystem in adopt prompt

scoop:
## Problem

Bootstrapping Dependabot from the Readiness panel's adopt prompt installed it with the **npm** ecosystem config even on bun repos. The `dependency_automation` guardrail's prescription line in the generated `claudeMd` artifact carried no stack guidance (`INSTALL_STEPS.dependency_automation` returns `[]`), so the spawned agent defaulted to `package-ecosystem: "npm"` — which never updates `bun.lock`, shipping a silently-broken Dependabot config.

## Fix

The analyzer already resolves the package manager (`RepoScan.pm` via `pickPackageManager`); it just never fed it into this prescription. Add a `PRESCRIPTION_NOTE` map rendered as prose guidance lines (no `$ ` prefix — not shell commands) under the adopt line in `generateClaudeMd`:

- `bun` → create `.github/dependabot.yml` with `package-ecosystem: "bun"` (the npm ecosystem would not update bun.lock)
- `npm`/`pnpm`/`yarn` → `package-ecosystem: "npm"` (Dependabot's npm ecosystem natively covers pnpm/yarn lockfiles; only bun has a dedicated one — same as this repo's own dependabot.yml)
- bun with only the binary `bun.lockb` → additionally prescribe `bun install --save-text-lockfile` first, since Dependabot's bun ecosystem requires the text lockfile

The note only renders when the guardrail is absent, matching the existing adopt-list path. Server-only; `claudeMd` is a verbatim generated artifact (i18n-exempt), no UI change.

## Tests

- bun repo → prompt contains `package-ecosystem: "bun"`, not `"npm"`
- npm + pnpm repos → `package-ecosystem: "npm"`
- `bun.lockb`-only repo → text-lockfile migration hint
- guardrail present → no note emitted
- existing `$ `-command-count test stays green (note lines carry no `$ ` prefix)

`bun run lint` + `bun run test`: 5817 pass / 0 fail.

### DIFF
--- a/src/readiness.ts
+++ b/src/readiness.ts
@@ -369,8 +369,11 @@ function generateClaudeMd(scan: RepoScan, checks: GuardrailCheck[]): string {
     ? missing
         .map((c) => {
           const head = `- ${TOOLING_LABEL[c.id]} — ${CHURN_PLAIN[c.id]}`;
-          const cmds = INSTALL_STEPS[c.id](verbs);
-          return cmds.length ? `${head}\n${cmds.map((cmd) => `    $ ${cmd}`).join("\n")}` : head;
+          const lines = [
+            ...(PRESCRIPTION_NOTE[c.id]?.(scan) ?? []),
+            ...INSTALL_STEPS[c.id](verbs).map((cmd) => `$ ${cmd}`),
+          ];
+          return lines.length ? `${head}\n${lines.map((l) => `    ${l}`).join("\n")}` : head;
         })
         .join("\n")
     : "- None — your deterministic guardrails already cover the baseline.";
@@ -456,6 +459,32 @@ export const PM_VERBS: Record<PackageManager, { add: string; exec: string }> = {
   pnpm: { add: "pnpm add -D", exec: "pnpm dlx" },
   yarn: { add: "yarn add -D", exec: "yarn dlx" },
   npm: { add: "npm i -D", exec: "npx" },
+};
+
+/**
+ * Extra prose guidance lines per guardrail, rendered under the adopt line (no `$ `
+ * prefix — guidance, not shell commands). Only for guardrails whose bootstrap is
+ * file-creation the agent would otherwise guess wrong: dependency_automation must
+ * name the Dependabot ecosystem, because agents default to `npm`, which never
+ * updates bun.lock on a bun repo (pnpm/yarn ARE served by the npm ecosystem —
+ * only bun has a dedicated one).
+ */
+const PRESCRIPTION_NOTE: Partial<Record<GuardrailId, (s: RepoScan) => string[]>> = {
+  dependency_automation: (s) => {
+    if (s.pm === "bun") {
+      const lines = [
+        `Create \`.github/dependabot.yml\` with \`package-ecosystem: "bun"\` — this repo uses bun; the npm ecosystem would not update bun.lock.`,
+      ];
+      if (!s.has("bun.lock"))
+        lines.push(
+          `Dependabot's bun ecosystem needs the text lockfile: run \`bun install --save-text-lockfile\` first (it reads bun.lock only; the binary bun.lockb is not supported).`,
+        );
+      return lines;
+    }
+    return [
+      `Create \`.github/dependabot.yml\` with \`package-ecosystem: "npm"\` (also correct for pnpm/yarn lockfiles).`,
+    ];
+  },
 };
 
 /**
--- a/test/readiness.test.ts
+++ b/test/readiness.test.ts
@@ -265,6 +265,43 @@ test("a bare bun repo emits exactly the installable-guardrail commands, all bun"
   expect(INSTALL_STEPS.dependency_automation(PM_VERBS.bun)).toEqual([]);
 });
 
+test("dependabot note prescribes the bun ecosystem for a bun repo", () => {
+  pkg({ name: "bare" });
+  write("bun.lock", "");
+  const r = analyzeReadiness(dir);
+  expect(r.claudeMd).toContain('package-ecosystem: "bun"');
+  expect(r.claudeMd).not.toContain('package-ecosystem: "npm"');
+  // Text lockfile already present — no migration hint.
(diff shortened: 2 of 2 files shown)


scoop (commented):
## Review

The PR satisfies the task: the generated CLAUDE.md adopt list now tells the agent which `package-ecosystem` to write into `.github/dependabot.yml`, keyed off the detected package manager, so a bun repo no longer gets the npm ecosystem (which would never update `bun.lock`).

### Verified

- **Lookup safety.** `PRESCRIPTION_NOTE` is a `Partial<Record<GuardrailId, …>>` with only `dependency_automation` defined (src/readiness.ts:472-488); the call site uses `PRESCRIPTION_NOTE[c.id]?.(scan) ?? []` (src/readiness.ts:373), so every other guardrail safely contributes zero note lines.
- **Output preserved for command-bearing guardrails.** The rewritten renderer (src/readiness.ts:372-376) produces byte-identical output to the old `cmds`-based code for guardrails without a note: same `$ ` prefix (now applied in the `.map`, src/readiness.ts:374) and same 4-space indent. `agent_instructions` and `ci` still return `[]` from `INSTALL_STEPS` (src/readiness.ts:507-508) and render as a bare head line, unchanged.
- **`RepoScan` contract.** `s.pm` and `s.has` exist on the scan (src/readiness.ts:61, 65) and are populated in `scanRepo` (src/readiness.ts:314, 326).
- **bun detection covers the hint's trigger.** `pickPackageManager` returns `bun` for either `bun.lock` or `bun.lockb` (src/readiness.ts:269), so `s.pm === "bun" && !s.has("bun.lock")` (src/readiness.ts:478) correctly fires the `--save-text-lockfile` migration hint exactly for bun.lockb-only (or `packageManager`-field-only) repos, and the bun.lock test asserts its absence (test/readiness.test.ts:268-276).
- **No format-sensitive consumers.** `claudeMd` is rendered verbatim in a `<pre>` and copied/sent as prose (ui/src/lib/components/ReadinessPanel.svelte:128, 142, 285); nothing parses the adopt-line structure, so adding un-prefixed prose lines is safe.
- **Tests.** All helpers used by the new tests exist: `pkg`/`write`/`present`/`dir` (test/readiness.test.ts:14-33), `rmSync`/`join` imported (test/readiness.test.ts:2-3). The "already present" test's `not.toContain("package-ecosystem")` is sound — that string appears in the generated snippet only via `PRESCRIPTION_NOTE` (grep over src/readiness.ts confirms src/readiness.ts:476 and :485 are the only sources). The pnpm case correctly reuses the npm-ecosystem note per the code's non-bun branch (src/readiness.ts:484-486).

### Nit (non-blocking)

- When `pm` resolves to bun via the `packageManager` field with **no lockfile at all**, the hint fires (correctly — `!s.has("bun.lock")`, src/readiness.ts:478) but its parenthetical "(bun.lockb alone is not supported)" presumes a `bun.lockb` exists. The prescribed action (`bun install --save-text-lockfile`) is still right for that repo; only the wording slightly misdescribes it.

### Limitations

- Could not verify Dependabot's ecosystem catalog against live GitHub docs from this repo: that `package-ecosystem: "bun"` exists, that it requires the text `bun.lock` (not `bun.lockb`), and that pnpm/yarn are served by the `npm` ecosystem are external facts. They match my knowledge of Dependabot's 2025 bun support, but the ground truth is not in this tree; assumed as written.

scoop (commented):
## Re-review

### Prior finding
1. **src/readiness.ts: text-lockfile hint parenthetical misfits no-lockfile bun repos — ADDRESSED.** The old parenthetical "(bun.lockb alone is not supported)" implied the repo currently has `bun.lockb`. The new wording at src/readiness.ts:480 — "Dependabot's bun ecosystem needs the text lockfile: run `bun install --save-text-lockfile` first (it reads bun.lock only; the binary bun.lockb is not supported)" — is a statement about Dependabot's capability rather than the repo's current state, so it reads correctly both for `bun.lockb`-only repos and for bun repos detected via the `packageManager` field with no lockfile at all. I confirmed the no-lockfile-bun path is real: `pickPackageManager` returns `"bun"` from the field alone before any lockfile probe (src/readiness.ts:267-268).

### Verification
- `PRESCRIPTION_NOTE` is keyed by `GuardrailId` and `"dependency_automation"` is a member of that union (src/readiness.ts:29) and a defined guardrail (src/readiness.ts:181).
- `RepoScan.pm` and `RepoScan.has` exist with the used signatures (src/readiness.ts:61, :65); `scan` is in scope at the call site inside `generateClaudeMd(scan, checks)` (src/readiness.ts:369-376).
- Rendering: note lines get no `$ ` prefix while `INSTALL_STEPS` commands do (src/readiness.ts:373-374), matching the doc comment; both are indented 4 spaces, preserving the prior format. `INSTALL_STEPS.dependency_automation` returns `[]` (src/readiness.ts:509), so for this guardrail the block renders note-only, and the `lines.length` guard keeps the head bare for guardrails with neither notes nor commands.
- Hint gating: bun repo with `bun.lock` → no migration hint; `bun.lockb`-only or field-only bun repo → hint added (`!s.has("bun.lock")`, src/readiness.ts:478). Both branches are covered by the new tests (test/readiness.test.ts:268-297), plus a no-note-when-present test (test/readiness.test.ts:299-306). Test helpers `write`/`pkg`/`present` and the `rmSync`/`join` imports all resolve (test/readiness.test.ts:2-33).
- The test assertions match the emitted strings: the note embeds `package-ecosystem: "bun"` / `"npm"` and `bun install --save-text-lockfile` verbatim (src/readiness.ts:475-491).

### Limitations (not findings)
- Could not verify against live external ground truth that Dependabot's `bun` package-ecosystem reads only `bun.lock` (not `bun.lockb`), that the npm ecosystem also serves pnpm/yarn lockfiles, or that `bun install --save-text-lockfile` is the correct migration flag — these are external-tool facts not present in this repo; they match my knowledge of Dependabot and bun ≥1.2 and are assumed as written.

PR #1399, opened 2026-07-04
fix(ui): resync epic/drain state on wake + socket reopen, soft-refresh backlog drawer

scoop:
## Problem

Epic tracking went stale in parts of the app (backlog `EPIC x/y` group-header badge disagreeing with the repos drawer). Root cause: the `/events` stream is **delta-only** and the server dedupes `epic:update` by signature (`lastEpicSig`) — it only re-emits on the *next* real change. Any delta missed while a mobile tab was frozen or the socket was down was lost for good:

1. The tab-return `resync()` re-pulled sessions/git/usage/holds/completed-epics but **not** `GET /api/drain` and **not** the live epics in `store.epics`.
2. Nothing resynced when a replacement socket opened — and a mobile freeze kills the socket **without ever firing `onclose`**, so a `connected` false→true edge can't even detect it.
3. `IssuesPanel`'s one-shot caches (`issues`, `epicByNumber` summaries, the per-epic `fetched` map) never refreshed while the drawer stayed open.

## Fix (client-only; `GET /api/epic` already assembles live state on demand)

- **`resync()`** now also re-pulls `/api/drain` (refreshing `activeEpicKeys`; a newly started epic is then fetched by the existing seed effect) and re-fetches every epic already in `store.epics`. A 2 s `lastResyncAt` guard dedupes the visibility/pageshow path.
- **`connectionEpoch`** in `HerdStore`, incremented in `ws.onopen`: epoch > 1 triggers `resync({ force: true })` — forced, so the guard can never suppress the one fetch that closes the sig-dedup loss window between a guarded wake-fetch and the new socket opening.
- **`backlogRefresh`** singleton (mirrors the `issuesFilter` idiom — the prop path would span six files): `resync()` bumps a nonce; `IssuesPanel` soft-refreshes issues + epic summaries + expanded epic panels **without** resetting filter text, expanded rows, or flashing a loading state. A previous-nonce mount latch (not `nonce === 0`) prevents double-fetch when the `{#if}`-mounted drawer opens after earlier bumps; a failed refreshed listing keeps the old list instead of swapping in an error banner.

## Tests

- `IssuesPanel.browser.test.ts`: mounted-with-nonce>0 → single fetch; nonce stable → no extra fetches; bump refetches issues/summaries/expanded-epic while preserving filter text + expansion; failed refresh keeps old data.
- `store.svelte.test.ts`: `connectionEpoch` counts every opened socket, including a replacement after a silent kill where `connected` never flips false.
- `cd ui && bun run check` ✓ (0 errors) · `bun run test` ✓ (2782 passed) · `check:i18n` ✓ · root `bun run lint` ✓

No i18n changes, no feature-catalog entry (fix, no new UX surface).

### DIFF
--- a/ui/src/lib/backlog-refresh.svelte.ts
+++ b/ui/src/lib/backlog-refresh.svelte.ts
@@ -1,0 +1,20 @@
+// Global soft-refresh signal for the backlog drawer's one-shot caches (IssuesPanel's
+// issues + epic summaries + expanded-epic fetches). +page's resync() bumps it on tab
+// wake and on socket re-open; IssuesPanel watches the nonce and refetches WITHOUT
+// resetting operator state (filter text, expanded rows). A module singleton — not a
+// prop — because the drilling path (+page → AppOverlays → BacklogOverlay → BacklogView
+// → BacklogTabContent → IssuesPanel) spans six files and IssuesPanel already imports
+// issuesFilter/steers/repos the same way. The nonce is page-lifetime and monotonic;
+// consumers must latch the value they last acted on (NOT compare against 0), since
+// the {#if}-mounted drawer routinely mounts after earlier bumps.
+class BacklogRefresh {
+  #nonce = $state(0);
+  get nonce(): number {
+    return this.#nonce;
+  }
+  bump() {
+    this.#nonce += 1;
+  }
+}
+
+export const backlogRefresh = new BacklogRefresh();
--- a/ui/src/lib/components/IssuesPanel.browser.test.ts
+++ b/ui/src/lib/components/IssuesPanel.browser.test.ts
@@ -767,4 +767,48 @@ describe("IssuesPanel soft refresh (backlogRefresh)", () => {
       m.common_issues_load_failed(),
     );
   });
+
+  it("a late-settling mount fetch cannot clobber a newer soft-refresh result", async () => {
+    // Mount fetch and soft refresh hit the SAME repo concurrently, so the
+    // rp !== repoPath guard alone can't order them — the fetch sequence token must.
+    type Listing = Awaited<ReturnType<typeof listIssues>>;
+    let rejectMount!: (e: Error) => void;
+    const mountFetch = new Promise<Listing>((_res, rej) => {
+      rejectMount = rej;
+    });
+    mockListIssues
+      .mockReturnValueOnce(mountFetch) // mount: stays pending
+      .mockResolvedValueOnce({
+        // soft refresh: settles first, with fresher data
+        slug: "owner/repo",
+        webUrl: null,
+        issues: [makeIssue(2, "Fresh issue")],
+        viewer: null,
+      });
+    mockGetEpics.mockResolvedValue({ epics: [], subIssues: [] });
+    render(IssuesPanel, { repoPath: "/repo", onnewtask: noop });
+
+    await expect.poll(() => mockListIssues.mock.calls.length).toBe(1);
+    backlogRefresh.bump();
+    await expect.poll(() => mockListIssues.mock.calls.length).toBe(2);
+
+    // The soft result must render even though the mount fetch never settled —
+    // i.e. softRefresh clears `loading` so fresh data isn't stuck behind the skeleton.
+    await expect
+      .poll(() =>
+        [...document.querySelectorAll(".issue-title")].map((el) => el.textContent?.trim()),
+      )
+      .toEqual(["Fresh issue"]);
+
+    // Now the superseded mount fetch settles late — first rejecting would previously
+    // stamp loadError over fresh data; a stale .then would restore older issues.
+    rejectMount(new Error("late mount failure"));
+    await new Promise((r) => setTimeout(r, 50));
+    expect(
+      [...document.querySelectorAll(".issue-title")].map((el) => el.textContent?.trim()),
+    ).toEqual(["Fresh issue"]);
+    expect(document.querySelector(".issues-list")?.textContent).not.toContain(
+      m.common_issues_load_failed(),
+    );
+  });
 });
--- a/ui/src/lib/components/IssuesPanel.browser.test.ts
+++ b/ui/src/lib/components/IssuesPanel.browser.test.ts
@@ -7,6 +7,7 @@ import { m } from "$lib/paraglide/messages";
 import { listIssues, getEpics, getEpic } from "$lib/api";
 import { steers } from "$lib/steers.svelte";
 import { issuesFilter } from "$lib/issues-filter.svelte";
+import { backlogRefresh } from "$lib/backlog-refresh.svelte";
 
 // Mock the API so no network calls fire; each test seeds the results.
 vi.mock("$lib/api", async (importOriginal) => {
@@ -598,5 +599,172 @@ describe("IssuesPanel hide-sub-issues filter (default ON)", () => {
     await expect.poll(() => issueTitles()).not.toContain("Plain sub-issue");
     expect(issueTitles()).toContain("Mid-level epic");
     expect(issueTitles()).toContain("Ordinary issue");
+  });
+});
+
+describe("IssuesPanel soft refresh (backlogRefresh)", () => {
+  function makeIssue(number: number, title: string): Issue {
+    return {
+      number,
+      title,
+      body: "",
+      url: `https://example.com/i/${number}`,
+      labels: [],
+      createdAt: 0,
+      assignees: [],
+    };
+  }
+
+  function makeSummary(parentIssueNumber: number, merged: number, total: number): EpicSummary {
+    return {
+      parentIssueNumber,
+      parentTitle: `Epic ${parentIssueNumber}`,
+      merged,
+      total,
+      status: "idle",
+      source: "markdown",
+    };
+  }
+
+  function makeEpic(
+    parentIssueNumber: number,
+    childStates: Epic["children"][number]["state"][],
+  ): Epic {
+    return {
+      repoPath: "/repo",
+      parentIssueNumber,
+      parentTitle: `Epic ${parentIssueNumber}`,
+      source: "markdown",
+      children: childStates.map((state, i) => ({
+        number: 100 + i,
+        title: `Child ${100 + i}`,
+        url: `https://example.com/i/${100 + i}`,
+        order: i,
+        body: "",
+        blockedBy: [],
+        state,
+        sessionId: null,
+        prNumber: null,
+        issueClosed: state === "merged",
+        claimed: false,
+      })),
+      warnings: [],
+      run: { repoPath: "/repo", parentIssueNumber, mode: "auto", status: "idle" },
+    };
+  }
+
+  it("mounted with nonce > 0 → single fetch (mount latch swallows the page-lifetime nonce)", async () => {
+    // The overlay is {#if}-mounted while the nonce increments page-lifetime, so a
+    // panel routinely mounts with nonce > 0 — the latch must NOT read that as a bump.
+    backlogRefresh.bump();
+    mockListIssues.mockResolvedValue({
+      slug: "owner/repo",
+      webUrl: null,
+      issues: [makeIssue(1, "Only issue")],
+      viewer: null,
+    });
+    mockGetEpics.mockResolvedValue({ epics: [], subIssues: [] });
+    render(IssuesPanel, { repoPath: "/repo", onnewtask: noop });
+
+    await expect.poll(() => document.querySelectorAll(".issue-title").length).toBe(1);
+    // Give a (buggy) soft-refresh double-fetch a chance to fire before counting.
+    await new Promise((r) => setTimeout(r, 50));
+    expect(mockListIssues).toHaveBeenCalledTimes(1);
+    expect(mockGetEpics).toHaveBeenCalledTimes(1);
+  });
+
+  it("nonce stable after mount → no extra fetches", async () => {
+    mockListIssues.mockResolvedValue({
+      slug: "owner/repo",
+      webUrl: null,
+      issues: [makeIssue(1, "Only issue")],
+      viewer: null,
+    });
+    mockGetEpics.mockResolvedValue({ epics: [], subIssues: [] });
+    render(IssuesPanel, { repoPath: "/repo", onnewtask: noop });
+
+    await expect.poll(() => document.querySelectorAll(".issue-title").length).toBe(1);
+    await new Promise((r) => setTimeout(r, 50));
+    expect(mockListIssues).toHaveBeenCalledTimes(1);
+    expect(mockGetEpics).toHaveBeenCalledTimes(1);
+  });
+
+  it("bump refetches issues + summaries + expanded epic, preserving filter text and expansion", async () => {
+    mockListIssues.mockResolvedValue({
+      slug: "owner/repo",
+      webUrl: null,
+      issues: [makeIssue(50, "Epic parent"), makeIssue(51, "Other issue")],
+      viewer: null,
+    });
+    mockGetEpics.mockResolvedValue({ epics: [makeSummary(50, 1, 3)], subIssues: [] });
+    // No `epics` (live store) prop → the expanded panel renders from the one-shot
+    // `fetched` cache; the bump must refresh that too.
+    mockEpic.mockResolvedValue(makeEpic(50, ["merged", "running", "ready"]));
+
+    render(IssuesPanel, { repoPath: "/repo", onnewtask: noop, expandEpic: 50 });
+
+    // Collapsed-row badge from the summary + expanded panel from the fetched Epic.
+    await expect.element(page.getByText(m.epic_badge({ merged: 1, total: 3 }))).toBeInTheDocument();
+    await expect
+      .poll(() => document.querySelector(".epic-badge")?.getAttribute("aria-expanded"))
+      .toBe("true");
+    await expect
+      .element(page.getByText(m.epic_progress({ merged: 1, total: 3 })))
+      .toBeInTheDocument();
+    expect(mockEpic).toHaveBeenCalledTimes(1);
+
+    // Operator types a filter — it must survive the refresh.
+    const filterInput = document.querySelector<HTMLInputElement>(".issue-filter")!;
+    filterInput.value = "Epic";
+    filterInput.dispatchEvent(new InputEvent("input", { bubbles: true }));
+
+    // Reality moved on: one more child merged.
+    mockGetEpics.mockResolvedValue({ epics: [makeSummary(50, 2, 3)], subIssues: [] });
+    mockEpic.mockResolvedValue(makeEpic(50, ["merged", "merged", "running"]));
+    backlogRefresh.bump();
+
+    // Badge + expanded panel both show the new counts…
+    await expect.element(page.getByText(m.epic_badge({ merged: 2, total: 3 }))).toBeInTheDocument();
+    await expect
+      .element(page.getByText(m.epic_progress({ merged: 2, total: 3 })))
+      .toBeInTheDocument();
+    // …the expanded fetched-cache epic was re-pulled…
+    expect(mockEpic).toHaveBeenCalledTimes(2);
+    expect(mockEpic).toHaveBeenLastCalledWith("/repo", 50);
+    // …and operator state survived: expansion + filter text intact (no hard reset).
+    expect(document.querySelector(".epic-badge")?.getAttribute("aria-expanded")).toBe("true");
+    expect(document.querySelector<HTMLInputElement>(".issue-filter")?.value).toBe("Epic");
+  });
+
+  it("keeps the old list when the refreshed listing reports a fetch failure", async () => {
+    mockListIssues.mockResolvedValue({
+      slug: "owner/repo",
+      webUrl: null,
+      issues: [makeIssue(1, "Survivor issue")],
+      viewer: null,
+    });
+    mockGetEpics.mockResolvedValue({ epics: [], subIssues: [] });
+    render(IssuesPanel, { repoPath: "/repo", onnewtask: noop });
+
+    await expect.poll(() => document.querySelectorAll(".issue-title").length).toBe(1);
+
+    // The wake-refresh hits a rate-limited forge: old data beats an error banner.
+    mockListIssues.mockResolvedValue({
+      slug: "owner/repo",
+      webUrl: null,
+      issues: [],
+      viewer: null,
+      error: "fetch_failed",
+    });
+    backlogRefresh.bump();
+
+    await expect.poll(() => mockListIssues.mock.calls.length).toBe(2);
+    await new Promise((r) => setTimeout(r, 50));
+    expect(
+      [...document.querySelectorAll(".issue-title")].map((el) => el.textContent?.trim()),
+    ).toEqual(["Survivor issue"]);
+    expect(document.querySelector(".issues-list")?.textContent).not.toContain(
+      m.common_issues_load_failed(),
+    );
   });
 });
--- a/ui/src/lib/components/IssuesPanel.browser.test.ts
+++ b/ui/src/lib/components/IssuesPanel.browser.test.ts
@@ -895,6 +895,47 @@ describe("IssuesPanel soft refresh (backlogRefresh)", () => {
     expect(mockEpic).toHaveBeenCalledTimes(2);
   });
 
+  it("a snapshot settling after seed AND prune both happened mid-flight is discarded", async () => {
+    // The narrowest window: expand → backfill getEpic held pending → epic:update
+    // seeds the live record → the run completes and the finished-prune drops the
+    // key — all BEFORE the fetch settles. At settle epics[key] is undefined again,
+    // so the settle-time guard alone would cache the pre-run snapshot; the seed-time
+    // invalidation must have already killed the ticket so the prune refetches fresh.
+    mockListIssues.mockResolvedValue({
+      slug: "owner/repo",
+      webUrl: null,
+      issues: [makeIssue(90, "Epic parent")],
+      viewer: null,
+    });
+    mockGetEpics.mockResolvedValue({ epics: [makeSummary(90, 0, 2)], subIssues: [] });
+    let resolveFirst!: (e: Epic) => void;
+    mockEpic
+      .mockReturnValueOnce(new Promise<Epic>((res) => (resolveFirst = res)))
+      .mockResolvedValueOnce(makeEpic(90, ["merged", "merged"]));
+
+    const live = reactiveRecord<Epic>({});
+    render(IssuesPanel, { repoPath: "/repo", onnewtask: noop, epics: live, expandEpic: 90 });
+    await expect.poll(() => mockEpic.mock.calls.length).toBe(1); // backfill #1, held pending
+
+    // Run starts (seed) and finishes (prune) while fetch #1 is still in flight.
+    live["/repo#90"] = makeEpic(90, ["merged", "running"]);
+    await expect
+      .element(page.getByText(m.epic_progress({ merged: 1, total: 2 })))
+      .toBeInTheDocument();
+    delete live["/repo#90"];
+
+    // The prune triggers a FRESH backfill fetch (#2) — the pending flag was cleared
+    // at seed time, so the effect isn't blocked by the still-unsettled fetch #1.
+    await expect.poll(() => mockEpic.mock.calls.length).toBe(2);
+    // Fetch #1's pre-run snapshot settles last; its invalidated ticket discards it.
+    resolveFirst(makeEpic(90, ["ready", "ready"]));
+    await expect
+      .element(page.getByText(m.epic_progress({ merged: 2, total: 2 })))
+      .toBeInTheDocument();
+    await new Promise((r) => setTimeout(r, 50));
+    expect(document.body.textContent).not.toContain(m.epic_progress({ merged: 0, total: 2 }));
+  });
+
   it("a late epic fetch for a since-collapsed panel is discarded — re-expand refetches", async () => {
     mockListIssues.mockResolvedValue({
       slug: "owner/repo",
--- a/ui/src/lib/components/IssuesPanel.browser.test.ts
+++ b/ui/src/lib/components/IssuesPanel.browser.test.ts
@@ -8,6 +8,7 @@ import { listIssues, getEpics, getEpic } from "$lib/api";
 import { steers } from "$lib/steers.svelte";
 import { issuesFilter } from "$lib/issues-filter.svelte";
 import { backlogRefresh } from "$lib/backlog-refresh.svelte";
+import { reactiveRecord } from "./reactive-fixture.svelte";
 
 // Mock the API so no network calls fire; each test seeds the results.
 vi.mock("$lib/api", async (importOriginal) => {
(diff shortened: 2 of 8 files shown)


scoop (commented):
## Review

The PR wires three staleness fixes together and the design holds up under inspection: a `connectionEpoch` counter that advances on every `ws.onopen` (so a mobile freeze that kills the socket without `onclose` still triggers a post-reopen resync), a 2s dedupe guard on the wake path with a `force` bypass for the socket-open path (needed because the server sig-dedupes `epic:update` and never re-sends missed deltas), and a module-singleton nonce that soft-refreshes the backlog drawer without resetting operator state.

### Verified

- **Epoch counting is correct.** `connectionEpoch` increments only in `ws.onopen` (ui/src/lib/store.svelte.ts:826-830), and `open()` nulls the previous socket's handlers before replacing it (ui/src/lib/store.svelte.ts:814), so a superseded socket can't double-count. The store and `connect()` are per-page-mount (ui/src/routes/+page.svelte:137 and :1585), so epoch starts at 0 each mount and the `epoch > 1` gate (ui/src/routes/+page.svelte:828-831) correctly skips the initial connect that onMount's bootstrap already covers. The new store test's silent-kill scenario (`readyState = 3` without `onclose`) matches the FakeWs semantics (ui/src/lib/store.svelte.test.ts:421-440: `close()` early-returns at readyState 3, so `connected` genuinely never flips false).
- **Epic-key parsing is correct.** Store keys are built as `${e.repoPath}#${e.parentIssueNumber}` (ui/src/lib/store.svelte.ts:249); the `lastIndexOf("#")` split in resync (ui/src/routes/+page.svelte:802-806) mirrors the pre-existing seed-effect parsing (ui/src/routes/+page.svelte:449-451) and handles repoPaths containing `#`.
- **Refetching ALL of `store.epics` (not just active keys) is required, not waste**: `epicFor` prefers the live-store record over the one-shot `fetched` cache (ui/src/lib/components/IssuesPanel.svelte:217-219), so a stale store entry for an idle epic would shadow a refreshed `fetched` value if the loop were scoped to active epics only.
- **The nonce latch is correct.** First effect run per mount latches whatever the page-lifetime nonce is and returns (ui/src/lib/components/IssuesPanel.svelte:169-177), so an `{#if}`-mounted drawer opening after earlier bumps does not double-fetch; the mount fetch is owned by the repoPath effect (ui/src/lib/components/IssuesPanel.svelte:119-157). `repoPath` is read only inside `untrack`, keeping the effect keyed on the nonce alone. IssuesPanel mounts only via the backlog drawer path (ui/src/lib/components/backlog-view/BacklogTabContent.svelte:42), so one instance per open drawer.
- **API signatures match all new call sites**: `getDrain()` (ui/src/lib/api.ts:1668), `getEpic(repoPath, parent)` (ui/src/lib/api.ts:1982), `getEpics(repoPath)` (ui/src/lib/api.ts:1976), `listIssues` including the optional `error` field the soft path checks (ui/src/lib/api.ts:921-937), `setDrain`/`setEpic` (ui/src/lib/store.svelte.ts:231, 248).
- **Test fixtures match the real types**: `Issue` (ui/src/lib/types.ts:169-179), `EpicSummary` (ui/src/lib/types.ts:674-681), `Epic`/`EpicChild` incl. the `"merged"|"running"|"ready"` states used (ui/src/lib/types.ts:645-673). Message keys `epic_badge`, `epic_progress`, `common_issues_load_failed` exist in both catalogs (ui/messages/en.json:26,1089,1108; ui/messages/de.json:26,1089,1108). Mock helpers `mockListIssues`/`mockGetEpics`/`mockEpic`/`noop` exist (ui/src/lib/components/IssuesPanel.browser.test.ts:25-43).

### Findings (non-blocking)

1. **ui/src/lib/components/IssuesPanel.svelte: same-repo fetch race between the initial load and softRefresh.** The staleness guards check only `rp !== repoPath` (lines 131, 184), which distinguishes repos but not request recency. If a bump fires while the mount fetch for the SAME repo is still inflight (drawer opened just before a freeze/wake, or a socket reopen right after opening the drawer), the two `listIssues` calls race: a late-settling mount `.catch` stamps `loadError = true` (line 144) AFTER softRefresh already applied fresh issues with `loadError = false` (line 189), flipping the panel to the failure banner (template gates loadError before the list, line 278) despite holding fresh data; a late mount `.then` similarly overwrites softRefresh's newer result with the older response. Relatedly, softRefresh never clears `loading`, so its data stays hidden behind the loading skeleton (line 276) until the mount fetch settles. A per-repo fetch sequence token (apply only the latest request's result) would close all three. Rare window, so non-blocking — but this concurrency is new in this PR.
2. **ui/src/routes/+page.svelte: the resync epic-refetch loop grows without bound over a long-lived tab.** `store.epics` is append-only (ui/src/lib/store.svelte.ts:249; no code ever deletes a key), so `for (const k of Object.keys(store.epics))` (line 802) fires one `GET /api/epic` per epic ever seen this page-lifetime on EVERY wake and every socket reopen — including long-finished epics nobody is looking at. On an operator dashboard left open for days across many epics this compounds; combined with the acknowledged wake+reopen double-resync (~1s apart), the multiplier doubles. Refetching all known keys is required today because `epicFor` prefers the store record (see Verified above), so the fix is to prune finished epics from `store.epics` (e.g. when they move to `completedEpics`) rather than to scope this loop. Non-blocking.

- ui/src/lib/components/IssuesPanel.svelte: softRefresh leaves `loading` untouched, so if a future caller bumps `backlogRefresh` on a cadence where the initial fetch is routinely still inflight (e.g. a periodic-refresh follow-up), fresh soft-refresh data would routinely hide behind the loading skeleton until the mount fetch settles (folded into finding 1's suggested sequence-token fix).

- ui/src/lib/components/PromptSources.svelte fetches its own issue listing and has no wake-refresh path; the New Task dialog's issue list can still be stale after a wake (informational only — file not in this diff).

scoop:
## Re-review

### Prior point 1 — IssuesPanel softRefresh race: ADDRESSED
The diff adds per-endpoint fetch sequence tokens (`issuesSeq`/`epicsSeq`, ui/src/lib/components/IssuesPanel.svelte:125-126): the mount/repo-change effect takes a ticket (line 138) and every settle path checks `rp !== repoPath || ticket !== seq` (lines 141, 153, 160), and softRefresh does the same (lines 191, 194, 216, 225). softRefresh now also clears `loading` when its clean result lands (line 213) and surfaces a failure instead of an eternal skeleton when the mount fetch was superseded (lines 199-204, 217-220). The exact clobber scenario from the prior review is covered by a regression test ("a late-settling mount fetch cannot clobber a newer soft-refresh result", ui/src/lib/components/IssuesPanel.browser.test.ts) plus a mount-latch test guarding against nonce double-fetches. Verified the nonce latch (`lastSeenNonce` initialized undefined, first run swallows, IssuesPanel.svelte:179-188) and that `untrack` keeps the effect keyed on the nonce alone.

### Prior point 2 — append-only `store.epics` resync fan-out: ADDRESSED
`setEpic` now prunes finished epics (idle + ≥1 child + all merged, ui/src/lib/store.svelte.ts:258-267) and `epic:completed` drops the live record (store.svelte.ts:640-644), so the resync loop at ui/src/routes/+page.svelte:802-806 iterates a self-cleaning set. I verified the prune condition against the server: the drain emits `epic:completed` first (src/drain.ts:698) and then a final `epic:update` whose run is flipped to idle with every child merged (src/drain.ts:707-711), matching the auto-complete condition at src/drain.ts:661-664 — so the trailing update is indeed swallowed by the finished-prune and cannot resurrect the key. Resync re-fetches of a since-completed epic also self-prune through `setEpic`. Tests cover prune, keep-idle-unmerged, and `epic:completed` drop (ui/src/lib/store.svelte.test.ts).

### New findings

**1. (blocking) Pruning the live record strands an already-expanded EpicPanel on "Loading…".** During a run the backlog's expanded epic panel renders from the live store: `expandEpicRow` skips the one-shot fetch when `epics?.[key]` is present (IssuesPanel.svelte:256), so `fetched` never gets an entry, and `epicFor` (line 249) serves the store record. When the epic completes, `epic:completed`/the finished-prune drop that key (store.svelte.ts:262, 644) → `epicFor` returns undefined → IssueRow flips the open panel to the loading placeholder (ui/src/lib/components/issues-panel/IssueRow.svelte:148-153). Nothing ever refetches it: softRefresh's recovery loop iterates `fetched.keys()` only (IssuesPanel.svelte:234) and the number isn't in `fetched`, so the panel stays stuck on "Loading…" across wakes until the user collapses and re-expands. This is exactly the flow where a user watches an epic finish (the `epics` prop exists to live-update open panels, IssuesPanel.svelte:43-44), and it contradicts the setEpic docstring's justification ("IssuesPanel falls back to its own one-shot fetch when the live record is absent", store.svelte.ts:253-255) — that fallback only fires at expand time. Pre-PR this state was unreachable because `store.epics` never dropped keys. Suggested fix in IssuesPanel: have softRefresh iterate `expanded` (fetching any expanded epic absent from both the live store and `fetched`), and/or trigger the one-shot fetch when an expanded row's `epicFor` goes undefined.

**2. (non-blocking) softRefresh's expanded-epic `getEpic` refetch lacks the ticket guard its sibling fetches got.** The loop at IssuesPanel.svelte:239-243 guards only `rp === repoPath`. Two bumps land back-to-back routinely (wake resync + forced socket-reopen resync each bump, +page.svelte:790-807, 830), so two `getEpic` calls for the same number can race and the later-settling (possibly older) one wins; a late result can also re-add a `fetched` entry for a since-collapsed epic, which a later expand then treats as fresh (line 256 sees `fetched.has`). Window is seconds-wide and self-heals on the next bump, hence non-blocking — but it's the same class of race the PR fixes for `listIssues`/`getEpics`, with the guard present on the siblings and missing here.

### Also verified
- `connectionEpoch` increments only in `ws.onopen` (store.svelte.ts:858) and the reopen effect skips epoch 1 (+page.svelte:829-831); the silent-kill test's helpers (`stubDom`, `connectFake`) exist at store.svelte.test.ts:444/464.
- The 2s resync guard cannot starve anything: `lastResyncAt` starts at 0 so the first wake resync always runs, and the socket-open path always passes `force:true` (+page.svelte:745-747, 830); wake listeners at +page.svelte:1588-1592 are the only unforced callers.
- Imports/identifiers all resolve: `backlogRefresh` (ui/src/lib/backlog-refresh.svelte.ts:20), `untrack` in both files (+page.svelte:2, IssuesPanel.svelte:21), `getDrain`/`getEpic`/`getCompletedEpics` in +page.svelte:35-42, `EpicChildState` exported at ui/src/lib/types.ts:645, `completedEpic` helper at store.svelte.test.ts:989, `toasts` imported at store.svelte.test.ts:3.
- No new i18n keys are introduced; tests reference existing `m.epic_badge`/`m.epic_progress`/`m.common_issues_load_failed` already used by IssueRow/IssuesPanel.
- The +page seed effect (+page.svelte:445-455) cannot loop with the prune: `seededEpics.add` precedes the fetch, and `dropEpic` no-ops (no state write) when the key is absent (store.svelte.ts:270).

- IssuesPanel.svelte:256-259 — `expandEpicRow`'s one-shot `getEpic` has no `rp !== repoPath` staleness guard, so a late result from a previously-selected repo can populate the new repo's `fetched` cache under the same issue number (untouched lines, predates this PR).
- Epics stopped mid-run (idle with unmerged children) still stay in `store.epics` indefinitely and are re-fetched on every wake — deliberate per the new keep-idle-unmerged test, and bounded by operator behavior, but the growth is not fully zero.

scoop (commented):
## Re-review verdict

Both prior findings are addressed by this revision; verification below. Two new non-blocking nits.

### Prior finding 1 — store prune strands an expanded, store-backed EpicPanel: ADDRESSED

- A new backfill `$effect` (ui/src/lib/components/IssuesPanel.svelte:290-294) iterates `expanded` and calls `fetchEpicInto` for any expanded row where `epicFor(num)` is undefined and no fetch is pending. `epicFor` (IssuesPanel.svelte:278-280) reads `epics?.[\`${repoPath}#${n}\`]` and `fetched.get(n)` inside the effect, so both are tracked dependencies; the store prune replaces the whole `epics` object (`dropEpic`, ui/src/lib/store.svelte.ts:270-275), which re-triggers the effect and re-seeds the panel from a one-shot fetch — exactly the fix requested.
- `softRefresh` now iterates `expanded` instead of `fetched.keys()` (IssuesPanel.svelte:241-244), skipping only rows the live store still covers, so pruned-but-expanded panels are refreshed on wake too.
- Covered by the new browser test "refetches an expanded panel when the live store prunes its record" (ui/src/lib/components/IssuesPanel.browser.test.ts:255-292), which drives the prune via a deeply-reactive prop fixture (ui/src/lib/components/reactive-fixture.svelte.ts:8-11 — verified the `$state` record export exists and the import path matches).
- Failure-path parity: if the backfill fetch rejects, the panel stays on its loading state until the next `expanded`/`epics`/`fetched` change retries it — same recovery the old expand-time one-shot had (old code at the removed IssuesPanel expand path only refetched on re-expand), so this is not a regression.

### Prior finding 2 — un-ticketed expanded-epic `getEpic` refetch: ADDRESSED

- `fetchEpicInto` (IssuesPanel.svelte:262-275) takes a per-number ticket from `epicFetchSeq`; the apply guard is `rp !== repoPath || epicFetchSeq.get(num) !== ticket || !expanded.has(num)` (line 268), so back-to-back bumps can't let an older `getEpic` result win, and a late settle cannot re-seed `fetched` for a since-collapsed epic. The `finally` clears the in-flight flag only while it still holds the latest ticket (line 272-274).
- Covered by "a late epic fetch for a since-collapsed panel is discarded — re-expand refetches" (IssuesPanel.browser.test.ts:294-325).
- The sibling issue/summary fetches got the same treatment (`issuesSeq`/`epicsSeq`, IssuesPanel.svelte:125-126, guards at 143/155/162/196/218/227).

### Other verification

- `+page.svelte` resync additions: `getDrain` and `getEpic` are imported (ui/src/routes/+page.svelte:35,42); `store.setDrain` exists (ui/src/lib/store.svelte.ts:231); the epic re-pull key split (`lastIndexOf("#")`, +page.svelte:803) matches the seed effect's established parsing (+page.svelte:449).
- No refetch loop from the prune: the seed effect (+page.svelte:445-456) dedups via the plain `seededEpics` set before fetching, and `dropEpic` no-ops when the key is absent (store.svelte.ts:271), so a re-pulled finished epic settles without state churn. The forced-resync effect keys on `store.connectionEpoch` alone with `untrack` around `resync()` (+page.svelte:828-831), so its `store.epics` reads can't self-trigger.
- `connectionEpoch` increments in `ws.onopen` (store.svelte.ts:858); epoch 1 (initial connect) is skipped by the `epoch > 1` guard (+page.svelte:830); the silent-kill scenario the design targets is covered by the new store test (store.svelte.test.ts:518-534) and the fake WS supports it (mutable `readyState`, store.svelte.test.ts:423-442).
- Consumers of the pruned `store.epics` tolerate a missing record: `groupSessionsByEpic` skips absent epics (ui/src/lib/components/epic-grouping.ts:31) and IssueRow renders a loading placeholder, not a crash (ui/src/lib/components/issues-panel/IssueRow.svelte:148-153).
- Test identifiers resolve: `EpicChildState` (ui/src/lib/types.ts:645), `reactiveRecord` import path, `mockListIssues`/`mockGetEpics`/`mockEpic` with `mockReset` in `beforeEach` (IssuesPanel.browser.test.ts:26-37), `stubDom`/`connectFake` (store.svelte.test.ts:444-472). No user-facing strings were added, so no locale-parity concern.

### Non-blocking nits

1. **Ticket reuse after an A→B→A repo flip** — the repo-change effect clears `epicFetchSeq` (IssuesPanel.svelte:138), restarting per-number tickets at 1. An in-flight `getEpic` from the previous visit to repo A then holds the same `(rp, num, ticket)` triple as a fresh fetch after switching A→B→A and re-expanding the same epic number, so its stale settle passes all three guards at IssuesPanel.svelte:268 and can overwrite the newer result (or, via the `finally` at 272-274, clear the pending flag for the newer in-flight fetch). Window is one HTTP round-trip across two repo switches and it self-heals on the next bump/update, hence non-blocking — but a single shared monotonic counter for ticket values (like `issuesSeq`) instead of restart-from-0 per number removes the race for free.
2. **Stale contract comment invalidated by the prune** — ui/src/lib/components/epic-grouping.ts:13-14 documents "`epics` … is never pruned, so it can hold stale/idle epics"; `setEpic`'s finished-prune and the `epic:completed` drop (store.svelte.ts:257-267, 640-646) make that false. Behavior is safe (epic-grouping.ts:31 skips missing records), but the comment now misdocuments the invariant this PR deliberately changed.

scoop (commented):
## Re-review

Both points from the previous revision are addressed in this diff.

**Prior point 1 — per-number epic-fetch tickets restarting after `epicFetchSeq.clear()`: RESOLVED.** Ticket values now come from one shared monotonic counter `epicFetchTicket` (ui/src/lib/components/IssuesPanel.svelte:259) that is never reset — the repo-change effect clears only the `epicFetchSeq` map and `epicFetchPending` set (IssuesPanel.svelte:138-139), not the counter. After an A→B→A flip, the fresh fetch mints a strictly larger ticket, so the first visit's still-in-flight settle fails the `epicFetchSeq.get(num) !== ticket` guard (IssuesPanel.svelte:274) and its `finally` fails the `epicFetchSeq.get(num) === ticket` check (IssuesPanel.svelte:279), leaving the newer fetch's pending flag intact. Verified by walking the A→B→A sequence against IssuesPanel.svelte:128-139 (repo-change clear), 268-281 (fetchEpicInto). The new comment at IssuesPanel.svelte:254-258 documents exactly this rationale.

**Prior point 2 — `epic-grouping.ts` doc contradicting the new prune: RESOLVED.** The comment now reads "FINISHED epics are pruned from it (HerdStore.setEpic drops idle+all-merged records) — so an `activeEpicKeys` entry may briefly lack its record here (skipped below)" (ui/src/lib/components/epic-grouping.ts:13-16), matching the actual prune in ui/src/lib/store.svelte.ts:257-267 and the skip at epic-grouping.ts:33 (`if (!epic) continue;`).

- `getDrain`/`getEpic` used by the new resync block exist and are imported: defined at ui/src/lib/api.ts:1668 and ui/src/lib/api.ts:1982; imported in +page.svelte:35,42.
- `resync()` call sites: visibilitychange (+page.svelte:1589) and pageshow-persisted (+page.svelte:1592) use the guarded path; the socket-reopen effect passes `force: true` (+page.svelte:830), bypassing the 2s dedupe as documented. `connectionEpoch` increments only in `ws.onopen` (store.svelte.ts:857-858); the `epoch > 1` gate correctly skips the initial connect, whose data onMount's bootstrap already fetches (+page.svelte:1490-1546).
- No setEpic/getEpic re-fetch loop: `dropEpic` is a no-op when the key is absent (store.svelte.ts:271), so a pruned re-pull does not reassign `this.epics`; +page's seed effect dedupes via `seededEpics` (+page.svelte:444-455) and the epoch effect untracks `resync` (+page.svelte:828-831).
- Prune consumers: `groupSessionsByEpic` skips missing records (epic-grouping.ts:33); IssuesPanel falls back to its one-shot `fetched` cache via `epicFor` (IssuesPanel.svelte:284-286) with the new backfill effect (IssuesPanel.svelte:296-300) re-fetching a pruned expanded panel; completed-epic display reads `store.completedEpics`, not `epics` (+page.svelte:410, store.svelte.ts:647-650). The `e.children.length > 0` guard (store.svelte.ts:261) keeps empty-children idle epics from being pruned.
- New test imports resolve: `EpicChildState` exported at ui/src/lib/types.ts:645; `reactiveRecord` at ui/src/lib/components/reactive-fixture.svelte.ts; `backlogRefresh` at ui/src/lib/backlog-refresh.svelte.ts; `mockEpic`/`mockListIssues`/`mockGetEpics` defined at IssuesPanel.browser.test.ts:26-28.
- Soft-refresh ticket guards (`issuesSeq`/`epicsSeq`) are taken by both the repo-change effect and `softRefresh` (IssuesPanel.svelte:140,159,193,224), so a late-settling superseded fetch can neither clobber newer data nor stamp `loadError` over it — matches the new browser test at IssuesPanel.browser.test.ts:211-253.

## Remaining finding (non-blocking)

`fetchEpicInto`'s success path (IssuesPanel.svelte:273-276) caches its snapshot whenever the ticket is current and the row is still expanded — even if the live store gained a record for that epic while the fetch was in flight. Sequence: expand while the store lacks the record → backfill fires `getEpic` → an `epic:update` seeds `epics[key]` before the fetch settles → the settle caches a pre-run snapshot into `fetched` (harmless while the live record wins in `epicFor`) → the epic later finishes and `setEpic` prunes the live record (store.svelte.ts:263-265) → `epicFor` now falls back to the stale snapshot, the panel visibly reverts (e.g. 3/3 → 0/3), and the backfill effect sees a defined record so it never refetches (IssuesPanel.svelte:296-300). Collapse/re-expand doesn't recover either, since `fetched` still holds the number; only the next `backlogRefresh` bump or a repo change fixes it. This reversion is new behavior (pre-PR the live record was never pruned, so the fallback could not surface). Narrow window (one HTTP round-trip overlapping the first `epic:update`), self-heals on the next wake — hence non-blocking. Suggested fix: in the `.then`, also skip the cache write when a live record now exists (`if (epics?.[`${rp}#${num}`]) return;`), so the prune-backfill path always fetches fresh.

scoop (commented):
## Re-review

### Prior finding — addressed ✔

**IssuesPanel.svelte: fetchEpicInto caching a pre-run snapshot when the live store gained the record mid-flight.** Fixed. The success path now bails before the cache write when the live store holds a record: `if (epics?.[`${rp}#${num}`]) return;` at ui/src/lib/components/IssuesPanel.svelte:280, placed after the ticket/expanded staleness guard (line 274) and before `fetched.set(num, e)` (line 281). The prune-backfill path (the `$effect` at IssuesPanel.svelte:302-306) then sees no record in either source and fetches fresh, exactly as the prior review requested. A dedicated regression test covers the precise sequence (expand → backfill held pending → live store seeded → late settle must not cache → prune → fresh fetch #2): "a snapshot settling after the live store gained the record is not cached (prune refetches fresh)" in ui/src/lib/components/IssuesPanel.browser.test.ts, asserting `mockEpic` was called twice and the panel shows the post-run counts.

### Residual (non-blocking)

The settle-time `epics?.[key]` check at IssuesPanel.svelte:280 only observes the store **at settle**. If the live store gains **and then loses** the record while the backfill fetch is still in flight (run starts → epic:update seeds → run completes → epic:completed / setEpic finished-prune drops the key, all before `getEpic` settles), the check sees `undefined` again and caches the pre-run snapshot after all. The backfill effect cannot help: at prune time it skipped the refetch because `epicFetchPending.has(num)` was true (IssuesPanel.svelte:304), and after the stale settle `epicFor(num)` is defined so it never refires. Same stale-counts symptom as the original bug, recovered only by the next `backlogRefresh` bump. The window is drastically narrower — an entire epic run must fit inside one HTTP fetch's flight (plausible mainly when a drain starts an epic whose children are already effectively done, against a slow/rate-limited forge) — so this is a nit, not a blocker. A tidy fix: when the backfill effect observes a live record for a number with a pending fetch, invalidate that fetch (`epicFetchSeq.set(num, ++epicFetchTicket); epicFetchPending.delete(num)`), so the prune-backfill path always launches a fresh fetch.

### Verification notes

- Fetch-sequence tickets: both the mount/repo-change fetch and softRefresh guard on `issuesTicket !== issuesSeq` / `epicsTicket !== epicsSeq` in `.then` and `.catch` (IssuesPanel.svelte:143, 155, 162, 196, 218, 227) — a superseded mount fetch can neither clobber fresh data nor stamp `loadError`; covered by the "late-settling mount fetch" test.
- `epicFetchSeq`/`epicFetchPending` are cleared in the repo-change effect (IssuesPanel.svelte:138-139) alongside `fetched`; ticket values come from the shared monotonic `epicFetchTicket` counter, so an A→B→A repo flip cannot re-mint a ticket a stale in-flight fetch still holds.
- The soft-refresh latch initializes `lastSeenNonce` on first run (IssuesPanel.svelte:184-187), so a panel mounting with a page-lifetime nonce > 0 does a single fetch — covered by the first new test.
- `setEpic` finished-prune (store.svelte.ts, `finished = idle && children.length > 0 && every merged`) with `dropEpic` no-op-when-absent; `epic:completed` also drops the live key (store.svelte.ts:646). All `store.epics` consumers tolerate a missing key: epic-grouping skips absent records (ui/src/lib/components/epic-grouping.ts:32-33, matching its updated doc comment), `groups` only maps keys built from present epics (epic-grouping.ts:59-60), the +page seed effect dedupes via `seededEpics` (ui/src/routes/+page.svelte:447) so a pruned finished epic does not refetch-loop, and IssuesPanel falls back to its one-shot cache (IssuesPanel.svelte:291). BacklogView only threads the prop through (BacklogView.svelte:372, 434). No other non-test consumer reads `epics[...]` (grepped ui/src, excluding tests).
- `connectionEpoch` has exactly one increment site, in the sole `ws.onopen` handler (store.svelte.ts:856-858); the +page effect fires resync only for epoch > 1 (+page.svelte:830), and `untrack` keeps it keyed on the epoch alone even though resync reads and writes `store.epics`.
- resync's per-key re-pull parses keys with `lastIndexOf("#")` (+page.svelte:802-804), consistent with the seed effect's parsing (+page.svelte:449); child numbers never contain `#`, so repoPaths containing `#` still parse correctly. A finished-while-away epic's re-pull self-cleans via the finished-prune.
- Test identifiers resolve: `mockEpic`/`mockGetEpics`/`mockListIssues`/`noop` are pre-existing in IssuesPanel.browser.test.ts:26-44; `EpicChildState` exists at ui/src/lib/types.ts:645; `reactiveRecord` is defined in the new ui/src/lib/components/reactive-fixture.svelte.ts.

No i18n/message-catalog changes in this diff.

scoop (commented):
## Re-review

### Prior finding — RESOLVED

**IssuesPanel.svelte settle-time guard missing the seed-then-prune-both-mid-flight case** — addressed, exactly as suggested. The backfill effect now treats a live record as authoritative for its number (ui/src/lib/components/IssuesPanel.svelte:310-317): while `epics?.[key]` is defined it bumps `epicFetchSeq.set(num, ++epicFetchTicket)` and clears `epicFetchPending` for any in-flight one-shot fetch, and drops any stale `fetched` entry.

I traced the original failure sequence against the new code:
1. Expand → `fetchEpicInto` takes ticket T1, `epicFetchPending` gains the number (IssuesPanel.svelte:268-271).
2. `epic:update` seeds the live record → the backfill effect re-runs (reactive on the `epics` prop), hits the new branch: seq bumped to T2, pending cleared (IssuesPanel.svelte:311-314).
3. Run completes → finished-prune drops the key → effect re-runs: `epics[key]` undefined, `epicFor` undefined, pending empty → a FRESH fetch fires with ticket T3 (IssuesPanel.svelte:318). This is the step that previously got skipped (pending flag was still set).
4. The old T1 fetch settles: discarded by the ticket guard `epicFetchSeq.get(num) !== ticket` (IssuesPanel.svelte:274), and its `finally` cannot unmark T3's pending flag (IssuesPanel.svelte:285).
5. T3 settles with post-run data and is cached (IssuesPanel.svelte:281).

The alternate ordering (settle lands between seed and the effect flush) is still covered by the retained settle-time guard at IssuesPanel.svelte:280, since `epics[key]` is defined at that moment. A dedicated browser test reproduces the exact window — pending fetch, seed, prune, then late settle — and asserts fetch #2 fires and the pre-run snapshot never renders (ui/src/lib/components/IssuesPanel.browser.test.ts:898-937).

The fix's own reactivity is sound: `fetched.delete` is guarded by `fetched.has` so the effect converges after one re-run, and `epicFetchSeq`/`epicFetchPending` are plain (non-reactive) containers (IssuesPanel.svelte:261, 267), so the invalidation writes don't retrigger the effect.

### New finding (non-blocking)

**Stale `seededEpics` latch after the new finished-prune.** `+page`'s seed effect adds a key to `seededEpics` when it fetches (ui/src/routes/+page.svelte:448) and only removes it on fetch FAILURE (ui/src/routes/+page.svelte:454) — pre-prune that was fine because a successfully-seeded key stayed in `store.epics` forever. The new prune (ui/src/lib/store.svelte.ts:252-258 via `setEpic`, and the `epic:completed` drop at ui/src/lib/store.svelte.ts:640-646) removes the key while `seededEpics` still holds it. If the SAME epic (same repo+parent) starts a second run within one page lifetime, the seed effect skips it (`seededEpics.has(k)` true at +page.svelte:447) even though `store.epics[k]` is gone, and resync's re-pull loop only iterates keys already in `store.epics` (+page.svelte:802-806) — so the resync comment's claim that "an epic that STARTED while away is then fetched by the seed $effect above" (+page.svelte:796-798) is false for a re-started, previously-completed epic. Until the run's next real state change produces an `epic:update` (which the server sig-dedupes and only emits on change), the Herd group header for that epic is missing — the exact staleness class this PR targets. Suggested fix: drop the key from `seededEpics` when its record leaves `store.epics` (e.g. reconcile `seededEpics` against `store.epics` in the seed effect, or clear stale entries in resync before `getDrain` settles).

### Verification notes

- `backlogRefresh` singleton exists and is imported consistently (ui/src/lib/backlog-refresh.svelte.ts:20, IssuesPanel.svelte:16, +page.svelte:132); the mount-latch logic (IssuesPanel.svelte:181-190) correctly swallows the first observation of the page-lifetime nonce, with tests for both the mounted-with-nonce>0 and stable-nonce cases (IssuesPanel.browser.test.ts:656-691).
- `connectionEpoch` exists on HerdStore (ui/src/lib/store.svelte.ts:61) and increments in `ws.onopen` (store.svelte.ts:858); the +page effect skips epoch 1 and forces resync on replacements (+page.svelte:828-831), with a store test covering the silent-kill/no-onclose case (store.svelte.test.ts:518-536).
- resync identifiers all resolve: `getDrain`/`getEpic`/`getCompletedEpics` imported at +page.svelte:35-42, `store.setDrain` at store.svelte.ts:231.
- `setEpic`'s finished condition requires `children.length > 0` (store.svelte.ts:249-251), so an empty-children epic is never mis-pruned; store tests cover prune, keep-idle-unmerged, and `epic:completed` drop (store.svelte.test.ts:1082-1108).
- The epic-grouping doc-comment change is accurate: the `if (!epic) continue` skip it references pre-exists at ui/src/lib/components/epic-grouping.ts:33, so a pruned record with a lingering `activeEpicKeys` entry degrades to ungrouped sessions, not a crash.
- Repo-flip ticket safety: the shared monotonic `epicFetchTicket` is never reset while `epicFetchSeq`/`epicFetchPending` are cleared on repo change (IssuesPanel.svelte:138-139, 259), so an A→B→A stale settle can neither apply nor unmark a newer fetch — matches the comment at IssuesPanel.svelte:254-258.
- Test helpers/keys all resolve: `makeIssue`/`makeSummary`/`makeEpic` (IssuesPanel.browser.test.ts:607-654), `reactiveRecord` (ui/src/lib/components/reactive-fixture.svelte.ts:8), `m.epic_progress` present in both catalogs (ui/messages/en.json, ui/messages/de.json — grep-confirmed both files contain the key).

- ui/src/routes/+page.svelte's seed-effect latch lines themselves (444-454) predate this PR; only their interaction with the new prune is raised above.

Issue #1037, opened 2026-06-23
Epic A handled fully but issues stayed open: landing PR never merged + last child closed by poller race


scoop:
## Summary

Epic A (`#1`) in `erwins-enkel/second-brain-mcp` ran end-to-end through Shepherd's epic-integration model and built cleanly: all 7 children (A1–A7, `#6`–`#12`) had PRs (#24–#30) squash-merged into the integration branch `epic/1-epic-a-core-document-store` and recorded in `epic_integrated`; the epic flipped to completed and opened aggregate landing PR **#31** (`epic/1… → main`).

Yet the backlog shows **6 of 7 children + the parent still OPEN, and exactly one child (#12 / A7) closed**. This post-mortem explains why, and proposes fixes. There are two distinct things going on — one is **working as designed but with a real UX/operational gap**, the other is a **genuine bug**.

## Finding 1 — The epic was integrated but never *landed* (by design; UX gap)

Shepherd deliberately does **not** close child issues when a sub-PR merges into the integration branch (`retireEpicChild` records `epic_integrated` and explicitly keeps the issue open — `src/drain.ts:1118-1120`). Closing is deferred to a single human merge of the landing PR #31, whose body carries `Closes #1, #6, #7, #8, #9, #10, #11, #12`. GitHub only auto-closes on a merge into the **default** branch, so until #31 merges, parent + children stay open. **This is correct** — the open issues are the expected mid-state of an epic that's integrated but not landed.

The problem is the last mile stalled and nothing surfaced it:

- Landing PR **#31 is OPEN, `MERGEABLE`, `mergeStateStatus=CLEAN`, CI green** — ready to merge, but Shepherd never auto-merges the landing PR (intentional: drain/train don't merge it; it's left for a human).
- `epic_completed` row shows `landingState='open'`, `migrationsAckedAt=NULL` with 5 un-acked drizzle migration files → the migration-acknowledgement checkpoint (#677) is still pending operator action.
- Result for the operator: "7 PRs merged, work clearly done, but the whole backlog still shows OPEN" reads as *broken*, when it's actually *waiting on one un-surfaced human action*.

## Finding 2 — #12 (A7) closed prematurely (genuine bug: PR-poller race)

A7 alone is closed, which violates the model above. Timeline (2026-06-23):

| time (UTC) | event |
| --- | --- |
[...]

Issue #645, opened 2026-06-13
Epic Runner: learnings from first epic (flowagent#327) — branch topology, sub-issue reconciliation, migration/CI verification


scoop:
> ## ✅ Status reconciliation (updated after the epic handler's second half landed)
>
> Since this was filed, **both halves of the Epic Runner shipped** — Stage A (#618: integration branch + done-on-merge gate) and Stage B (#661: final `epic/# → default` landing PR). The remaining actionable learnings are addressed by **PR #677**. Disposition of each of the 8 learnings:
>
> | # | Learning | Status |
> |---|----------|--------|
> | 1 | Epic-branch topology drift | ✅ **Fixed** (PR #677) — canonical branch name now **pinned & persisted** (`epic_run.integrationBranch`), so a title edit no longer re-points spawns/landing; plus divergence **detection** (title-drift, merged-child base-drift, host-branch scan) surfaced as epic warnings. |
> | 2 | Inconsistent sub-issue PR base | ✅ **Fixed** (PR #677) — retire now **verifies the child PR's actual base** and **fails closed** (warn + block merge, actionable `gh pr edit … --base` remedy) on mismatch, with a self-healing marker sweep. |
> | 3 | No sub-issue/checklist reconciliation | ✅ **Done** (#661) — the landing PR's `Closes #child`×N + `Closes #parent` reconciles/closes every child + parent on the single landing merge. *(Surfaced + closed-at-landing: children sit OPEN-with-merged-code during the epic, then close together at landing.)* |
> | 4 | Stacked-PR diff noise | ✅ **Done** (Stage A) — `selectEpicCandidates` gates spawns on blockers being done-in-epic, so dependencies land before dependents spawn. |
> | 5 | DB migrations never run in-harness | 🟡 **Partial** (PR #677) — a **migration-awareness acknowledgement checkpoint**: the landing band detects migration files in the landing PR and requires a human **acknowledge** before the landed epic clears. **Declined/descoped:** (i) blocking the autonomous completion flip (would re-wedge the drain — Stage B deliberately decoupled it), and (ii) in-harness migration **execution** / ephemeral DB (conflicts with the read-only critic design). Full provisioning remains an **accepted limitation** — refile if desired. |
[...]

PR #176, opened 2026-06-01
fix(badge): lead MERGED badge with checkmark to match siblings

scoop:
## What

On the unit card, the MERGED badge trailed its checkmark (`MERGED ✓`) while the title pip and the REVIEWED badge both lead with it (`✓ REVIEWED`). Two in front, one in the rear — looked off.

Moved the MERGED checkmark to the front so all three align:

- `prbadge_merged`: `MERGED ✓` → `✓ MERGED` (en), `GEMERGT ✓` → `✓ GEMERGT` (de)
- updated `pr-badge.test.ts` assertion

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #134, opened 2026-06-01
feat(ui): show in-progress critic badge in session list

scoop:
## Problem

The session list only showed a critic badge once the review **finished** (`✓ REVIEWED` / `⚠ CHANGES` / `REVIEW ERR`). The in-progress *"Reviewing…"* indicator existed solely in `GitRail` (the detail pane), so from the list you couldn't tell a critic run was underway.

## Fix

`CriticBadge.svelte` now reads `reviews.isReviewing(sessionId)` (the store already tracked it) and renders an amber, pulsing-dot **`REVIEWING…`** badge while a run is in flight, falling back to the verdict label once it lands. Styling + `prefers-reduced-motion` handling mirror GitRail's `.rev-dot`.

## i18n

Added `criticbadge_reviewing` + `criticbadge_reviewing_title` to both `en.json` and `de.json`.

## Validation

- `check:i18n` — 250 keys in parity
- `svelte-check` — 0 errors
- UI tests — 115 pass
- prettier/eslint clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1404, opened 2026-07-04
fix(epic): record epic_integrated on out-of-band / merge-train / manual-session merges

### LINKED ISSUE #1401: fix(epic): record epic_integrated on out-of-band / merge-train / manual-session merges of epic-child PRs
## Problem

An epic-child PR that merges **outside the drain's retire path** never gets its `epic_integrated` row recorded — the epic then stalls forever: the all-children-merged check never fires, `epic_completed` is never written, and the landing PR (epic branch → default) is never opened. No error, no surfacing; the epic just sits `running`.

`retireEpicChild` (`src/drain.ts` ~1620) is the **only** writer of `epic_integrated`, and it only runs for the drain's own auto sessions. Every other settle path archives without recording:

- `settleMergedSession` (`src/merge-teardown.ts`) — poller reap of out-of-band merges + local forge merge. Already carries the `isIntegratedEpicChild` guard (#1037) but only *reads* integration state, never records it.
- `AutoMergeService.doMerge` (`src/automerge.ts` ~309) — same.

## Real incident (pulse epic #128)

- Child #134's auto session (TASK-1248) died without opening a PR; a replacement session (TASK-1249) was spawned with `auto=0` (manual respawn).
- Its PR erwins-enkel/pulse#176 merged fine into `epic/128-feat-autotask-time-entry-review-correcti` — but via the poller settle path, so no `epic_integrated` row.
- Rows existed for #130–133 only → child #134 read as un-merged forever → `handleEpicSideEffects` never completed the epic → no landing PR. Repaired by hand-inserting the missing row.

## Fix

On every merged-PR settle path (poller reap, merge-train `doMerge`, local `forgeMerge`), before/alongside `settleMergedSession`: if the session has an `issueNumber` that is a child of an active epic AND the merged PR's base is that epic's integration branch, call `recordEpicIntegrated(repoPath, parent, child, {number, url}, base)` — idempotent upsert, so double-recording on the happy path is harmless.

Notes:

- Key detail: resolve parent + integration branch from the epic model (e.g. `epic_run`/`epic_branch` for the repo), and match the merged PR's **actual baseRefName**, not the session's stored `baseBranch` alone — a manual session may carry the right base but the PR is the source of truth.
- Must cover **manual (`auto=0`) sessions** too — that's exactly the escape hatch that bit #128. The `s.auto` gate in `settleMergedSession` must not gate the recording.
- Recording must precede/compose with the existing `isIntegratedEpicChild` archive-only branch so the just-recorded child is not issue-closed out of band (invariant from #1037: child issues close only when the landing PR merges).

## Reconcile sweep (required, not optional)

Event-time recording alone can't repair epics that stalled **before** this fix ships (the settle moment is one-shot and already consumed). Add a convergence pass:

- For each `running` epic: take children lacking an `epic_integrated` row, query the forge for a **merged** PR whose `baseRefName` is the epic's integration branch and that maps to the child (linked issue / head-branch / session record), and backfill via the idempotent `recordEpicIntegrated` upsert.
- Run it on startup and/or a slow periodic tick (it's a handful of forge reads per stalled child, throttled like the other reconciles).
- After backfill the normal pipeline takes over: `handleEpicSideEffects` completes the epic → `ensureLandingPr` opens the landing PR. No manual DB surgery.

## Acceptance test — pulse epic #128

`/home/patrick/Work/pulse` epic #128 is being left in its stalled state **on purpose** (missing `epic_integrated` row for child #134 / PR erwins-enkel/pulse#176). Do **not** hand-repair it. The implementation is done when deploying this fix makes #128 self-heal end-to-end with zero manual intervention: reconcile backfills #134 → epic completes → landing PR `epic/128-feat-autotask-time-entry-review-correcti` → default opens by itself.

scoop:
Closes #1401

## Problem

`retireEpicChild` was the **only** writer of `epic_integrated`. An epic-child PR that merged outside the drain's retire path (poller-observed out-of-band merge, merge train, local merge — and especially a **manual `auto=0` respawn**) never got its row, so the epic stalled `running` forever: no `epic_completed`, no landing PR, no error surfaced. Real incident: pulse epic #128 / child #134 / PR erwins-enkel/pulse#176.

## Fix

**Event-time recording** — new `recordEpicIntegrationIfChild` (`src/merge-teardown.ts`), called on every merged-PR settle path **before** `settleMergedSession` so the #1037 `isIntegratedEpicChild` guard sees the fresh row and archives-only (child issues still close only via the landing-PR merge):

- `drain.handle`/`onGit`: hoisted **above** the `!s.auto` gate — manual sessions previously never reached any settle logic at all; this is the hook that covers the #128 escape hatch.
- `AutoMergeService.doMerge` and the local `forgeMerge` path in `server.ts`.
- Gates: session has an `issueNumber`, epic run `running`/`paused`, and the merged PR's **actual base** equals the **pinned** integration branch. Base resolution: payload `baseRefName` → number-keyed `prReviewMeta` → on base-incapable forges (Gitea/local: neither exists) fall back to `s.baseBranch`, mirroring the #645 Gitea carve-out; a base-capable forge with an unresolvable base fails closed. Divergent `epic/*` bases are never recorded (fail-closed; #645 divergence warnings surface them). Idempotent upsert; never throws; ungated by `s.auto`.

**Reconcile sweep** (repairs epics stalled *before* this ships — the settle moment is one-shot and already consumed): `reconcileEpicIntegrations` runs from `drain.tick()`, throttled to one sweep per epic per 5 min; the in-memory throttle means the **first tick after a restart sweeps immediately**, so deploying is what heals #128. For each un-integrated open child it probes **all distinct branches** across the child's session rows (any status/auto — `store.list()` is createdAt-ordered, so #128's dead predecessor TASK-1248 sorts before the merged respawn TASK-1249; first-match would miss it), skips live sessions whose snapshot shows an open PR (event-time owns those), and backfills on merged + pinned-base match. The normal pipeline then completes the epic and opens the landing PR in the same tick.

Cost: zero forge calls for repos without an active epic; otherwise ≤1 branch-keyed `prStatus` per distinct not-open-PR branch per un-integrated child per 5 min.

## Pre-deploy acceptance-shape verification (read-only, done)

Confirmed against the live pulse DB + GitHub — no writes, stall left in place on purpose:
- `epic_run` /home/patrick/Work/pulse #128 `running`; `epic_integrated` rows for #130–133 only.
- Pinned `epic_branch` = `epic/128-feat-autotask-time-entry-review-correcti`.
- TASK-1249: `issueNumber=134`, `auto=0`, archived, branch `shepherd/feat-time-review-s5` → `gh pr view` resolves **MERGED** PR #176 with exactly that base.
- TASK-1248 (dead predecessor, sorts first) has **no** PR → the probe-all-branches logic is required and exercised by a dedicated two-row test.

## Tests

- `test/merge-teardown.test.ts`: full record/no-op matrix for the helper (auto=0 records, idle/unpinned/mismatch/divergent no-op, `prReviewMeta` fallback, base-incapable carve-out both ways, never-throws).
- `test/drain-epic-reconcile.test.ts`: event-time (manual merged records + session untouched; auto reap records then archive-only, no `closeIssue`; wrong base → normal close) and the sweep (two-row #128 acceptance shape probing both branches → backfill → epic completes → landing PR opens in one tick; throttle; open-PR skip; wrong-base fail-closed; probe-error continue; paused records without completing; no epic → zero probes).
- `test/automerge.test.ts`: `doMerge` records before archive (#1037 order asserted); non-epic repo records nothing.

`bun run lint`, `tsc --noEmit`, `bun test ./test` (5865 pass / 0 fail), fallow audit: clean.

```shepherd:manual-steps
- [ ] POST-MERGE: Deploy and observe pulse epic #128 self-heal with zero manual intervention: reconcile backfills child #134 (PR erwins-enkel/pulse#176) → epic completes → landing PR epic/128-feat-autotask-time-entry-review-correcti → default opens by itself
```

### DIFF
--- a/src/automerge.ts
+++ b/src/automerge.ts
@@ -8,7 +8,7 @@ import {
   type MergeRepoState,
   type MergeSessionView,
 } from "./automerge-core";
-import { settleMergedSession } from "./merge-teardown";
+import { recordEpicIntegrationIfChild, settleMergedSession } from "./merge-teardown";
 import { isFullAuto } from "./full-auto";
 
 /** Live per-repo merge-train status pushed to clients. */
@@ -50,6 +50,9 @@ export interface AutoMergeDeps {
     | "setAutoMergeState"
     | "setAutopilotState"
     | "isEpicIntegratedChild"
+    | "getEpicRun"
+    | "getEpicIntegrationBranch"
+    | "recordEpicIntegrated"
   >;
   service: {
     archive(id: string): Promise<number>;
@@ -306,6 +309,17 @@ export class AutoMergeService {
     // credit the train tracker directly (the poller-driven resolveMerging never fires
     // for this path). No-op unless the session was merge-train-flagged.
     this.deps.service.resolveMerging(sessionId, true);
+    // #1401: record epic integration BEFORE settleMergedSession so its isIntegratedEpicChild
+    // guard (#1037) sees the fresh row and archives-only. The base comes from the prCache
+    // snapshot (a merge doesn't change the base), with the helper's prReviewMeta fallback.
+    // Normally unreachable for epic children (isFullAuto excludes integration-branch bases),
+    // but a session whose PR was re-targeted to the epic branch can still be train-merged.
+    const git = this.deps.prCache.snapshot()[sessionId] ?? null;
+    await recordEpicIntegrationIfChild(
+      s,
+      { number: prNumber, url: git?.url, baseRefName: git?.baseRefName },
+      { store: this.deps.store, forge },
+    );
     await settleMergedSession(s, {
       resolveForge: this.deps.resolveForge,
       archive: (id) => this.deps.service.archive(id),
--- a/src/drain.ts
+++ b/src/drain.ts
@@ -4,7 +4,7 @@ import type { CreateSessionInput, Session } from "./types";
 import type { SessionStateChange } from "./session-snapshot";
 import type { UsageLimits } from "./usage-limits";
 import type { TelemetryService } from "./telemetry";
-import { settleMergedSession } from "./merge-teardown";
+import { recordEpicIntegrationIfChild, settleMergedSession } from "./merge-teardown";
 import { isFullAuto } from "./full-auto";
 import {
   ACTIVE_LABEL,
@@ -56,6 +56,10 @@ const SPAWN_FAIL_COOLDOWN_MS = 5 * 60_000;
  *  `prReviewMeta` API call on every pump while the operator hasn't fixed it — recheck at most
  *  this often per child. Bounds the cost to ≤1 call/child/~60s while a child stays blocked. */
 const EPIC_BASE_RECHECK_TTL_MS = 60_000;
+/** #1401: one reconcile sweep per epic per this window (plus one on the first tick after a
+ *  restart — the throttle map is in-memory). Slow on purpose: the sweep is convergence repair
+ *  for merges whose event-time recording was missed, not a hot path. */
+const EPIC_RECONCILE_TTL_MS = 5 * 60_000;
 
 /** Cap on epic-landing-PR open attempts (#635, Stage B). A failed `openPr` flips the
  *  `epic_completed` row to `landingState:'error'` and increments `landingAttempts`; the
@@ -238,6 +242,10 @@ export class DrainService {
   // recomputing on restart is fine — no persisted column).
   private epicBranchScanCache = new Map<string, { at: number; divergent: string[] }>();
   private lastEpicSig = new Map<string, string>();
+  // #1401: `${repoPath}#${parentIssueNumber}` → last reconcile-sweep timestamp. In-memory on
+  // purpose: a restart sweeps immediately (deploy ⇒ a pre-existing stall self-heals within one
+  // tick), then settles to one sweep per EPIC_RECONCILE_TTL_MS.
+  private epicReconcileAt = new Map<string, number>();
   /** #790: `${repoPath}#${issueNumber}` → last spawn-failure timestamp; throttles re-spawn
    *  of an issue whose create() keeps throwing. Cleared on a successful spawn. In-memory. */
   private spawnFailures = new Map<string, number>();
@@ -1865,6 +1873,11 @@ export class DrainService {
   async handle(change: SessionStateChange): Promise<void> {
     const s = change.snapshot.session;
     if (change.kind === "git") {
+      // #1401: record epic integration BEFORE the auto gate — a manual (auto=0) session never
+      // reaches reapMerged, so this is the only event-time hook covering its merged PR. Also
+      // deliberately before settleMergedSession (via reapMerged below) so the #1037
+      // isIntegratedEpicChild guard sees the fresh row and archives-only.
+      if (change.git.state === "merged") await this.recordEpicIntegrationForMerge(s, change.git);
       if (!s.auto) return; // drain only manages auto sessions
       if (change.git.state === "merged") {
         await this.reapMerged(s);
@@ -1880,7 +1893,10 @@ export class DrainService {
   /** pr-poller observed a new git state for a session. */
   async onGit(id: string, git: GitState): Promise<void> {
     const s = this.deps.store.get(id);
-    if (!s || !s.auto) return; // drain only manages auto sessions
+    if (!s) return;
+    // #1401: record BEFORE the auto gate (see handle() — same manual-session coverage).
+    if (git.state === "merged") await this.recordEpicIntegrationForMerge(s, git);
+    if (!s.auto) return; // drain only manages auto sessions
     if (git.state === "merged") {
       await this.reapMerged(s);
       return;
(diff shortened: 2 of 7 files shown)


scoop (commented):
## Review — fix(epic): record epic_integrated on all merged-PR settle paths + reconcile sweep (#1401)

The PR does what the issue asks, and every correctness-relevant dependency I checked holds up.

### What the issue requires vs. what the diff does

- **Poller reap / out-of-band merges (incl. manual `auto=0` sessions)**: `DrainService.handle()` and `onGit()` now call `recordEpicIntegrationIfChild` on a merged observation *before* the `if (!s.auto) return` gate (src/drain.ts:1877, src/drain.ts:1896-1898). This is exactly the escape hatch from the #128 incident. Confirmed `handle()` is the production entry (session-router dispatches consumers via `c.handle(change)`, src/session-router.ts:62, wired at src/index.ts:1519); `onGit` mirrors it, so both are consistent.
- **Merge train**: `AutoMergeService.doMerge` records before `settleMergedSession` (src/automerge.ts:317-322), base taken from the prCache snapshot with the number-keyed `prReviewMeta` fallback. `prCache` was already in `AutoMergeDeps` (src/automerge.ts:69).
- **Local forge merge**: `forgeMerge` in src/server.ts:3229-3237 records before settle on the `forge.kind === "local"` branch.
- **Ordering vs. #1037**: on all three paths the record precedes `settleMergedSession`, whose `isIntegratedEpicChild` guard (src/merge-teardown.ts:117-123) then archives-only and never closes the child issue out of band. Verified the reap path wires that guard from the store (src/drain.ts:2019-2021), as do the train (src/automerge.ts:330-332) and local-merge (src/server.ts:3244-3247) paths.
- **Base resolution = PR source of truth**: payload `baseRefName` → number-keyed `forge.prReviewMeta` → base-incapable carve-out → fail-closed (src/merge-teardown.ts:71-84). Verified `prReviewMeta` exists only on the GitHub forge (src/forge/types.ts:432 optional; src/forge/github.ts:1338) and that neither src/forge/gitea.ts nor src/forge/local.ts has `prReviewMeta` or `baseRefName` (grep: zero matches in both files) — so the carve-out claim in the doc comment is accurate, and it extends the same trust the retire path already extends (`epicChildBaseBlocked` returns false when `!forge.prReviewMeta`, src/drain.ts:1563; `retireEpicChild` records `s.baseBranch`, src/drain.ts:1636).
- **Gates mirror the retire path**: active-epic check `running || paused` matches src/drain.ts:1683; exact pinned-branch match (fail-closed on divergent `epic/*` bases) matches the #645 enforcement at src/drain.ts:1586.
- **Store contract**: `recordEpicIntegrated(repoPath, parent, child, pr?, mergedBase?)` verified against src/store.ts:1386-1400 — idempotent PK upsert; the helper's `pr.url ?? ""` is safe because the upsert guards with `COALESCE(NULLIF(excluded.prUrl,''), …)` (src/store.ts:1398). `getEpicRun` (src/store.ts:1323), `getEpicIntegrationBranch` (src/store.ts:1375), `isEpicIntegratedChild` (src/store.ts:1425) all exist with the shapes used.
- **Reconcile sweep**: `reconcileEpicIntegrations` runs at the top of the per-repo `tick()` loop, ungated by the drain toggle, error-wrapped, throttled to one sweep per epic per 5 min with an immediate sweep after restart (in-memory map) — src/drain.ts:2096-2106, src/drain.ts:1940-1965. Probing is branch-keyed `forge.prStatus` over ALL of the child's session rows (any status, any auto flag), which correctly handles the #128 shape where the dead predecessor session sorts before the manual respawn. Verified GitHub `prStatus` finds merged PRs by head branch even after branch deletion (`gh pr list --head … --state all`, src/forge/github.ts:841-854; GitHub retains headRefName) and supplies `baseRefName`/`url`/`number` on both the GraphQL (src/forge/github.ts:733) and REST-fallback (src/forge/github.ts:780) paths.
- **Self-heal in one tick**: after backfill, the same tick's pump completes the epic and opens the landing PR — `handleEpicSideEffects` flips running→idle (src/drain.ts:668-721) and the pump immediately calls `openLandingPrOnComplete` (src/drain.ts:1477). The acceptance-shaped test (`test/drain-epic-reconcile.test.ts`, real `SessionStore` + real `DrainService`) asserts exactly the #128 end state: both branches probed in row order, row backfilled, epic idle, landing PR open.
- The src/automerge.ts:315 comment ("isFullAuto excludes integration-branch bases") verified against src/full-auto.ts:25.

Test coverage is thorough: manual-session recording, archive-only settle (#1037), wrong-base fail-closed, prReviewMeta fallback + fail-closed, Gitea carve-out both directions, never-throws, throttle timing, open-PR skip, probe-error continue, paused-epic backfill-without-completion, no-epic no-probes.

- src/merge-teardown.ts: `resolveMergedBase` (line 78-80) conflates a null/undefined `forge` with a base-incapable forge — `if (!forge?.prReviewMeta) return s.baseBranch;` fails OPEN (trusts the session's stored base) when a future caller passes no forge for a base-capable repo with a payload lacking `baseRefName`. Dormant today: every in-diff caller either guarantees a non-null forge (doMerge bails on `!forge` at src/automerge.ts:295; forgeMerge and the reconcile probe hold one) or feeds GitHub GitStates that always carry `baseRefName` (src/forge/github.ts:733), and even the fail-open still requires an exact pinned-branch match — the same trust retireEpicChild extends. A `forge === null → return null` (fail-closed) vs. `forge lacks prReviewMeta → carve-out` distinction would be strictly safer for future callers.

### Informational (design notes, no change required)
- Reconcile mapping is session-records-only (ratified in the plan per the doc comment at src/drain.ts:1926-1935). A child integrated by a merged PR with **no** shepherd session row at all (fully human-authored branch, or a pruned session row) still cannot backfill and would stall the epic the same way. The issue's wording permits session-record mapping, and the #128 incident had a session row, so this satisfies the task — noting the residual gap for the record.
- `reconcileEpicIntegrations` stamps the throttle timestamp before the forge/`buildEpic` checks (src/drain.ts:1945-1948), so a transient failure consumes the full 5-min window. Consistent with "one attempt per TTL" and self-healing; fine for a convergence loop.
- Cannot verify the live acceptance criterion (pulse epic #128 self-healing on deploy) from this repo — the stalled state lives in `/home/patrick/Work/pulse`'s deployment; the acceptance-shaped unit test reproduces its exact session topology instead.

- src/forge/github.ts `prStatus` uses `--limit 1` (non-fork) when listing PRs by head branch with `--state all`, so a branch with multiple historical PRs yields only the newest — a newer closed PR could shadow an older merged one; affects the reconcile probe's input but predates this PR.

PR #1369, opened 2026-07-03
fix(ui): stabilize settings dialog tabs

kai-osthoff:
## Summary
- Give the settings modal a stable desktop height so tab switches do not move or resize the shell.
- Keep existing tab content mounted and scrolling inside the active panel.
- Add a browser regression for desktop modal geometry across tab switches.

## Verification
- cd ui && bun run check
- cd ui && bun run test:browser -- src/lib/components/Settings.browser.test.ts
- cd ui && bun run test
- pre-push hook passed

PR #1415, opened 2026-07-04
fix(ui): pin the Repos modal to a fixed height, like the Settings modal

scoop:
## What

The **Repos** modal (`BacklogOverlay`, labeled *Repos*) sized itself to its content — its `.card` had only `max-height: 88vh`. Switching between a sparse repo and a busy one, or between the Issues/PRs/Automation tabs, made the whole modal shell grow and shrink.

This adopts the exact fix **#1369** applied to the **Settings** modal: give `.card` a fixed `height`, so the shell stays put and the inner content scrolls instead.

```css
.card {
  width: min(960px, 94vw);
  height: min(720px, 88vh);   /* new */
  max-height: 88vh;
}
```

## Notes

- **`720`, not `680`** (what #1369 used for Settings) — the Repos modal is wider (960px) and hosts a two-column browser, so it wants more vertical room. Intentional divergence, verified against an interactive mockup.
- **Scroll lands where it already did:** the repo list `.master-pane` and the tab content in `BacklogTabContent` are the `overflow-y:auto` scrollers; `.detail-pane` stays `overflow:hidden`. No inner-layout change was needed.
- **Mobile unaffected:** `.card.mobile` (`height:100%; max-height:none`) has higher specificity and still wins, so the full-screen mobile sheet is unchanged.

## Test

Adds `BacklogOverlay.browser.test.ts` mirroring #1369's stability guard: at a desktop viewport it captures the `.card` geometry, swaps a busy (40-repo) payload for a sparse (2-repo) one, and asserts `height`/`top` are unchanged. Verified it **fails** when the `height` line is removed and passes with it.

`bun run check` clean; `bun run test:browser BacklogOverlay` green.

[no-feature-entry] — layout polish mirroring a merged fix; no new user-facing capability.

### DIFF
--- a/ui/src/lib/components/BacklogOverlay.browser.test.ts
+++ b/ui/src/lib/components/BacklogOverlay.browser.test.ts
@@ -1,0 +1,87 @@
+import { describe, it, expect, afterEach } from "vitest";
+import { render } from "vitest-browser-svelte";
+import { page } from "vitest/browser";
+import "../../app.css";
+import BacklogOverlay from "./BacklogOverlay.svelte";
+import type { BacklogPayload, BacklogProject } from "$lib/types";
+
+const noop = () => {};
+
+function project(path: string): BacklogProject {
+  return {
+    path,
+    display: path,
+    slug: `org/${path}`,
+    kind: "github",
+    openIssues: 3,
+    openPRs: 1,
+    prKinds: null,
+    workflows: null,
+    ciStatus: null,
+    hidden: false,
+  };
+}
+
+// n projects → the master list wants to be n·rows tall; pre-fix (max-height only)
+// the shell grows to fit that, so busy vs sparse payloads yield different shell heights.
+function payload(n: number): BacklogPayload {
+  return {
+    pinnedPath: null,
+    projects: Array.from({ length: n }, (_, i) => project(`/r/repo-${i}`)),
+    totals: { openIssues: 0, openPRs: 0 },
+  };
+}
+
+function props(over: Partial<Record<string, unknown>> = {}) {
+  return {
+    payload: payload(40),
+    mobile: false,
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

Adds `height: min(720px, 88vh)` to `.card` in `BacklogOverlay.svelte` so the Repos modal shell no longer resizes with content, plus a browser test guarding that behavior. This adopts the fixed-height pattern from the Settings modal, as the task requires.

## Verification

- `ui/src/lib/components/BacklogOverlay.svelte:112` adds `height: min(720px, 88vh)`; the existing `max-height: 88vh` (line 113) is now redundant but harmless since `min(720px,88vh) ≤ 88vh` always. This mirrors the Settings modal pattern verified at `ui/src/lib/components/Settings.svelte:1585-1586` (`height: min(680px, 86vh); max-height: 86vh;`).
- Mobile is unaffected: `.card.mobile` (specificity 0,2,0) sets `height: 100%; max-height: none;` at `ui/src/lib/components/BacklogOverlay.svelte:120-127`, overriding `.card` (0,1,0) irrespective of source order.
- Test targets resolve to real elements: `BacklogOverlay.svelte:5` imports `BacklogView`; its `.master-pane` (`ui/src/lib/components/BacklogView.svelte:386`) has `overflow-y: auto` (`ui/src/lib/components/BacklogView.svelte:506`), matching the test assertions. `.card`/`.card.mobile` classes exist as asserted in the test.

The Repos modal uses 720px/88vh where Settings uses 680px/86vh; this is a deliberate, defensible difference that preserves the Repos modal's pre-existing `max-height: 88vh` and `width: min(960px, 94vw)` dimensions — the PR adopts the pattern, not the literal numbers, consistent with the test's stated intent. No issues found.

PR #1427, opened 2026-07-05
feat(ui): forward Claude's OSC 52 c-to-copy to the browser clipboard

scoop:
## Problem

When Claude Code loses its login it prints a sign-in URL with a **`c to copy`** helper. In
Shepherd `c` did nothing useful (the "one-way clipboard"), so you had to hand-select the
Ink-hard-wrapped URL — and xterm inserts a line break at every wrap point, which you then had to
find and delete before the URL would paste into a local browser.

## Root cause (reverse-engineered from the Claude 2.1.201 bundle)

Claude's clipboard-mode selector emits an **OSC 52** escape (`ESC]52;c;<base64>BEL`) whenever
there's no local clipboard tool — and the `SSH_CONNECTION` override forces OSC 52 on **any SSH
session** (i.e. a normal remote/headless herdr). But `@xterm/xterm` only handles OSC ids
`[0,1,2,4,8,10,11,12,104,110,111,112]` — **not 52** — so Shepherd silently dropped it. Confirmed
empirically: pressing `c` on a throwaway login screen emits `\x1b]52;c;<base64-of-the-URL>`.

## Fix

Forward OSC 52 to the **browser** clipboard — general, for *every* `c to copy`, not just the
login URL:

- **`osc52.ts`** — pure parser. Refuses read/query (`Pd === "?"`) requests so a remote program
  can never read the operator's clipboard; caps decoded size at 32 KB (pre- and post-decode);
  byte-exact base64→UTF-8 (`atob`/`TextDecoder`, browser-safe).
- **`Viewport.svelte`** — registers an OSC 52 handler. On a valid write it calls
  `navigator.clipboard.writeText` and shows a **toast** (so every clipboard write is visible).
- **`ClipboardPill.svelte`** — the reliability floor. The OSC 52 bytes arrive async over the
  WebSocket (not inside the `c` keydown), so browsers may refuse the write. On **any** rejection
  or when the async clipboard API is unavailable, the text is stashed and a one-click **Copy
  pill** appears — clicking it writes inside a real user gesture. So `c` can never silently fail.

### Security

OSC 52 can be emitted by any PTY output, so: reads (`?`) are refused, payloads are size-capped,
and every write surfaces a toast (never silent). Decoded text is only ever passed to
`writeText()` — it is never rendered into the DOM (no injection surface).

## Testing

- `osc52.test.ts` (7): byte-exact round-trip incl. multibyte UTF-8, `?`-read → null,
  malformed/oversized → null, empty selection target.
- `ClipboardPill.browser.test.ts` (3): Copy writes the exact text + fires `oncopied`; rejection
  fires `oncopyfailed`; dismiss fires `ondismiss`.
- Full UI suite, `bun run check`, `check:i18n` (EN+DE parity), and PR-hygiene gates
  (feature-catalog, announcement-versions, glossary, branch-hygiene) all green.
- **Empirically verified** Claude emits OSC 52 on `c` via a throwaway `CLAUDE_CONFIG_DIR` +
  narrow PTY (no real-auth impact).

**Manual check for the reviewer:** end-to-end browser confirmation (press `c` on a real login
screen → URL lands on the browser clipboard; blur the tab → the Copy pill appears and its click
also copies) can't be automated headlessly; the async-write reliability is fully absorbed by the
pill fallback regardless of browser.

Includes a What's-New entry (`v1.42.0-clipboard-osc52.ts`).

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -160,7 +160,15 @@
       // DemoRibbon above, just a newer instance of it. No other duplicated
       // content in this file, so ignoring the file is equivalent to ignoring
       // just that block.
-      "**/ui/src/lib/components/TelemetryConsent.svelte"
+      "**/ui/src/lib/components/TelemetryConsent.svelte",
+      // ClipboardPill.svelte's only clone-flagged content is the same `.gbtn`
+      // recipe (lines ~77-99) — same verbatim-copy design-system convention as
+      // DemoRibbon/TelemetryConsent above, just a newer instance of it (the
+      // OSC 52 c-to-copy pill). No other duplicated content in this file, so
+      // ignoring the file is equivalent to ignoring just that block. Do NOT
+      // extract a shared stylesheet — that's the same out-of-scope 15+-file
+      // change called out above.
+      "**/ui/src/lib/components/viewport/ClipboardPill.svelte"
     ]
   },
 
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2525,7 +2525,6 @@
   "pluginupdate_apply_err_generic": "Update fehlgeschlagen.",
   "clipboard_copied_toast": "In die Zwischenablage kopiert",
   "clipboard_copy_failed": "Kopieren in die Zwischenablage fehlgeschlagen",
-  "clipboard_copy_retry": "Erneut versuchen",
   "clipboard_pill_prompt": "Terminal hat Text kopiert",
   "clipboard_pill_copy": "In Zwischenablage kopieren",
   "feat_clipboard_osc52_title": "Claudes „Kopieren“ landet jetzt in deiner Zwischenablage",
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2522,5 +2522,11 @@
   "pluginupdate_apply_err_symlinked": "Dieses Plugin läuft aus einem verlinkten Quell-Checkout — aktualisiere es dort.",
   "pluginupdate_apply_err_incompatible": "Die neue Version braucht ein neueres Shepherd — aktualisiere zuerst Shepherd.",
   "pluginupdate_apply_err_nosource": "Keine Update-Quelle — Shepherd kann keine neue Version automatisch holen.",
-  "pluginupdate_apply_err_generic": "Update fehlgeschlagen."
+  "pluginupdate_apply_err_generic": "Update fehlgeschlagen.",
+  "clipboard_copied_toast": "In die Zwischenablage kopiert",
+  "clipboard_copy_failed": "Kopieren in die Zwischenablage fehlgeschlagen",
+  "clipboard_pill_prompt": "Terminal hat Text kopiert",
+  "clipboard_pill_copy": "In Zwischenablage kopieren",
+  "feat_clipboard_osc52_title": "Claudes „Kopieren“ landet jetzt in deiner Zwischenablage",
+  "feat_clipboard_osc52_body": "Wenn Claude im Terminal Text kopiert (`c to copy`), landet er jetzt in deiner System-Zwischenablage. Blockiert der Browser das automatische Schreiben, erscheint eine kleine Copy-Pille über dem Terminal zum Kopieren per Klick."
 }
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2525,6 +2525,7 @@
   "pluginupdate_apply_err_generic": "Update fehlgeschlagen.",
   "clipboard_copied_toast": "In die Zwischenablage kopiert",
   "clipboard_copy_failed": "Kopieren in die Zwischenablage fehlgeschlagen",
+  "clipboard_copy_retry": "Erneut versuchen",
   "clipboard_pill_prompt": "Terminal hat Text kopiert",
   "clipboard_pill_copy": "In Zwischenablage kopieren",
   "feat_clipboard_osc52_title": "Claudes „Kopieren“ landet jetzt in deiner Zwischenablage",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2522,5 +2522,11 @@
   "pluginupdate_apply_err_symlinked": "This plugin runs from a linked source checkout — update it there.",
   "pluginupdate_apply_err_incompatible": "The new version needs a newer Shepherd — update Shepherd first.",
   "pluginupdate_apply_err_nosource": "No update source — Shepherd can't fetch a new version automatically.",
-  "pluginupdate_apply_err_generic": "Update failed."
+  "pluginupdate_apply_err_generic": "Update failed.",
+  "clipboard_copied_toast": "Copied to clipboard",
+  "clipboard_copy_failed": "Couldn’t copy to clipboard",
+  "clipboard_pill_prompt": "Terminal copied text",
+  "clipboard_pill_copy": "Copy to clipboard",
+  "feat_clipboard_osc52_title": "Claude's \"copy\" now reaches your clipboard",
+  "feat_clipboard_osc52_body": "When Claude copies text in the terminal (`c to copy`), it now lands in your system clipboard. If the browser blocks the automatic write, a small Copy pill appears over the terminal for a one-click copy."
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2525,7 +2525,6 @@
   "pluginupdate_apply_err_generic": "Update failed.",
   "clipboard_copied_toast": "Copied to clipboard",
   "clipboard_copy_failed": "Couldn’t copy to clipboard",
-  "clipboard_copy_retry": "Retry",
   "clipboard_pill_prompt": "Terminal copied text",
   "clipboard_pill_copy": "Copy to clipboard",
   "feat_clipboard_osc52_title": "Claude's \"copy\" now reaches your clipboard",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2525,6 +2525,7 @@
   "pluginupdate_apply_err_generic": "Update failed.",
   "clipboard_copied_toast": "Copied to clipboard",
   "clipboard_copy_failed": "Couldn’t copy to clipboard",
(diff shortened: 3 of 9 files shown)


scoop (commented):
## Summary

This PR wires Claude Code's `c to copy` (OSC 52) through to the browser clipboard, which directly solves the reported problem: instead of manually text-selecting the login URL (which mangled 4 spots with line-wrap/whitespace artifacts), pressing `c` now writes the exact bytes to the local clipboard. When the browser refuses the async write (no user gesture / unfocused tab), a `ClipboardPill` offers an in-gesture click-to-copy retry. This is the right mechanism and it satisfies the task.

## Verification

- **OSC parsing / xterm contract**: `parseOsc52` consumes `<Pc>;<Pd>` and splits on the first `;`; Pc has no `;` and base64 Pd has no `;`, so the split is correct (`ui/src/lib/osc52.ts:37-40`). `@xterm/xterm ^6.0.0` (`ui/package.json:48`) exposes `term.parser.registerOscHandler(52, cb)` handing the callback everything after `52;`, matching the comment (`ui/src/lib/components/Viewport.svelte:1199`).
- **Security / exfiltration guard**: read request (`Pd === "?"`) returns null so clipboard contents are never returned to the terminal program (`ui/src/lib/osc52.ts:44`); returning `true` from the handler still claims the sequence so nothing else can service the read. Payload is bounded before decode (`MAX_B64_LENGTH`, `osc52.ts:19`) and after decode (`OSC52_MAX_BYTES`, `osc52.ts:52`); `TextDecoder("utf-8", {fatal:true})` rejects non-UTF-8 (`osc52.ts:57`).
- **Graceful degradation**: `navigator.clipboard?.writeText(...)` short-circuits to the pill path when the API is absent (insecure context) (`Viewport.svelte:1202-1210`); the promise has a reject handler so no unhandled rejection. The pill's `copy()` wraps the call in try/catch → `oncopyfailed` (`ClipboardPill.svelte:22-29`).
- **Teardown**: `osc52Sub.dispose()` added to the cleanup block (verified against `Viewport.svelte:1654`).
- **i18n parity**: all six new keys — `clipboard_copied_toast`, `clipboard_copy_failed`, `clipboard_pill_prompt`, `clipboard_pill_copy`, `feat_clipboard_osc52_title`, `feat_clipboard_osc52_body` — exist in BOTH `ui/messages/en.json` and `ui/messages/de.json`. `common_close` used by the dismiss button exists in both (verified against `ui/messages/en.json:5`, `ui/messages/de.json:5`).
- **Feature announcement**: auto-discovered via `import.meta.glob` (verified against `ui/src/lib/feature-announcements.ts:47-50`); `sinceVersion: "1.42.0"` is correctly one ahead of the last release `1.41.0` (verified against `package.json:3`), satisfying the feature-gate/version checks; filename `v1.42.0-clipboard-osc52.ts` matches `v<sinceVersion>-<id>.ts`; id `clipboard-osc52` is unique (only one match in entries/).
- **Layout**: `ClipboardPill` is `position:absolute` and lives inside `.vp-body`, which is `position:relative` (verified against `Viewport.svelte:2996`) and sets `--review-banner-h` (verified against `Viewport.svelte:2201`), so the pill anchors and lifts above the review banner as the comment claims. The scoped `.gbtn` duplication follows the established codebase convention (DoneRecapPanel.svelte:210, EpicHandsOffIntro.svelte:219).

## Non-blocking nit

- `ui/src/lib/components/viewport/ClipboardPill.svelte:32-35`: the 20s auto-dismiss `$effect` does not read `text`, so it has no reactive dependency. If a second OSC 52 write arrives while a pill is already mounted, `pendingCopy` (and thus the `text` prop) updates in place but the dismiss countdown is not restarted — the newer text can be dismissed earlier than 20s. Minor UX only; reference `text` inside the effect to reset the timer if you want per-write timing.

No blocking issues found.

scoop (commented):
## Summary

This PR forwards Claude's `c to copy` (OSC 52) writes to the browser clipboard, with a one-click retry pill when the browser refuses the async write outside a user gesture. It cleanly satisfies the task: the login URL now reaches the system clipboard byte-exact (no more hand-fixing 4 mangled spots from a terminal text selection).

### Prior finding — ADDRESSED
`ui/src/lib/components/viewport/ClipboardPill.svelte`: the auto-dismiss `$effect` now reads `void text;` (line 33) before arming the 20s timeout, so a second OSC 52 write that replaces `text` on an already-mounted pill re-runs the effect, clears the stale timeout, and re-arms a full countdown. Verified at ClipboardPill.svelte:32-36.

### Verification performed
- **i18n parity**: all new keys added to BOTH catalogs — `clipboard_copied_toast`, `clipboard_copy_failed`, `clipboard_copy_retry`, `clipboard_pill_prompt`, `clipboard_pill_copy`, `feat_clipboard_osc52_title`, `feat_clipboard_osc52_body` (verified ui/messages/en.json:2525-2530 and ui/messages/de.json:2525-2530). `common_close` used by the pill exists in both (en.json:5, de.json:5).
- **toasts API**: the `{ alert, duration: null, action: { label, run } }` shape used at Viewport.svelte:2222-2231 matches `InfoOpts` (toasts.svelte.ts:38-50); `duration: null` yields a persistent toast per toasts.svelte.ts:95/142.
- **OSC 52 framing**: xterm's `registerOscHandler(52, cb)` hands the callback everything after `52;` (i.e. `<Pc>;<Pd>`); `parseOsc52` slices at the first `;` to get `Pd`, matching the tests' `"c;" + b64(...)` inputs (osc52.ts:32-40, osc52.test.ts). `@xterm/xterm` is `^6.0.0` (ui/package.json:48), whose `IParser.registerOscHandler` returns an `IDisposable`; disposal wired at Viewport.svelte:1648.
- **Exfiltration guard**: read/query (`Pd === "?"`) returns null and drives neither callback (osc52.ts:36-38, covered by osc52.test.ts).
- **Failure completeness**: `handleOsc52` routes async reject, synchronous throw, and missing async clipboard API all to `onPending`, so a payload never vanishes silently (osc52.ts:88-98).
- **Feature announcement**: id `clipboard-osc52` is unique; filename `v1.42.0-clipboard-osc52.ts` agrees with `sinceVersion: "1.42.0"`, consistent with the 15 sibling v1.42.0 entries; titleKey/bodyKey resolve in both catalogs.
- **Placement**: pill rendered inside `.vp-body` which is `position: relative` (Viewport.svelte:2998), so its `position: absolute` anchors correctly; `bottom: calc(12px + var(--review-banner-h, 0px))` has a safe `0px` fallback.

### Non-blocking nit
When the pill's own Copy click fails (the double-failure path), `oncopyfailed` fires a persistent failure toast whose Retry action runs `pendingCopy = text` (Viewport.svelte:2227-2228). At that point `pendingCopy` already equals `text` (it is what is driving the visible pill), so the assignment is a no-op — no reactive change, and the write is not re-attempted. The pill remains visible for the real retry, so nothing breaks, but the toast's "Retry" button is effectively inert (it only dismisses the toast). Consider having the toast's Retry re-invoke the clipboard write, or drop the redundant Retry action since the pill already provides it. Reachable only after an in-gesture clipboard failure (genuinely broken/insecure clipboard), so low value.

scoop (commented):
## Re-review

The single prior finding is **resolved**. The previous revision's `oncopyfailed` toast carried a Retry action that ran `pendingCopy = text` (a no-op on that path). The new diff (Viewport.svelte:2222-2226) drops the action entirely — `oncopyfailed` now only fires an assertive failure toast and keeps the pill mounted, so the pill's own Copy button is the retry surface (a fresh in-gesture click). This is a clean fix, corroborated by commit `ab904aa4 fix(ui): drop no-op Retry action on failed clipboard copy`.

## Task fit

The PR forwards Claude's `c to copy` (OSC 52) to the browser clipboard — directly addressing the reported problem that the SSH/xterm clipboard transfer was one-way and forced manual URL surgery. `parseOsc52`/`handleOsc52` decode the base64 payload byte-exactly (osc52.ts:31-58), attempt `navigator.clipboard.writeText`, and on any refusal stash the text for a one-click pill retry. This satisfies the task.

## Verification

- **i18n parity** — all six new keys exist in BOTH catalogs: `clipboard_copied_toast`, `clipboard_copy_failed`, `clipboard_pill_prompt`, `clipboard_pill_copy`, `feat_clipboard_osc52_title`, `feat_clipboard_osc52_body` (verified en at ui/messages/en.json:2526-2531 and de at ui/messages/de.json:2526-2531). `m.common_close()` used by the pill's dismiss button exists in both (ui/messages/en.json:5, ui/messages/de.json:5).
- **Toast API** — `toasts.info(text, { alert: true })` matches the signature (verified against ui/src/lib/toasts.svelte.ts:92 and the `alert?: boolean` option at line 44-46).
- **Feature-announcement registration** — the entry's shape (`id`, `sinceVersion`, `titleKey`, `bodyKey`) matches the `FeatureAnnouncement` type (verified against ui/src/lib/feature-announcements.ts:26-29) and is auto-loaded by the eager `import.meta.glob("./feature-announcements/entries/*.ts")` (ui/src/lib/feature-announcements.ts:47-50), so no manual wiring is needed.
- **CSS var** — the pill's `bottom: calc(12px + var(--review-banner-h, 0px))` (ClipboardPill.svelte:57) reuses the same variable published on `.vp-body` (Viewport.svelte:2195) and mirrors ViewportTermBanners.svelte:127; the pill renders inside `.vp-body` so it inherits the var.
- **Exfiltration guard** — `parseOsc52` refuses `Pd === "?"` read requests (osc52.ts:46) and caps decoded size at `OSC52_MAX_BYTES` with a cheap pre-decode length bound (osc52.ts:20,48,53).
- **Outcome completeness** — `handleOsc52` routes resolved→onCopied, rejected/throw/no-API→onPending (osc52.ts:83-98), so a payload can never vanish silently; the success/pending paths are mutually exclusive (auto-write success toasts and never shows the pill; only refusal sets `pendingCopy` and mounts the pill).
- **Handler lifecycle** — `term.parser.registerOscHandler(52, …)` returns a disposable claimed via `return true` and disposed in teardown (Viewport.svelte:1200-1207, 1648).

No blocking issues found.

PR #1432, opened 2026-07-05
fix: recover the plugin-injected account when herdr restart re-drives a session

scoop:
## Problem

After a herdr daemon restart/crash, herdr autonomously re-runs a bare `claude --resume <id>` with
clean env (its claude integration is detection-only — it strips `launch_argv`/env on restore). For a
session whose account identity is plugin-injected (the claude-swap plugin sets a per-account
`CLAUDE_CONFIG_DIR` via `onSpawn`), the restored pane lands on the **wrong config dir** → `claude`
prints **"No conversation found with session ID: …"**. The restored husk stays herdr-live, so Shepherd
adopts it and the autonomous drain machinery steers the wrong-account pane.

Account identity lives **only** in the plugin-injected `CLAUDE_CONFIG_DIR`; core persisted nothing to
detect or recover this. Empirically confirmed (Phase 0): herdr retains the restored/exited pane in
`agent list` (as a live or `unknown` agent), and neither `matchAgent`/`paneAlive` nor `liveTerminalIds`
filter by status — so the husk is observed live and steerable in every sub-case.

## Fix (plugin-agnostic, via the existing `onSpawn` re-drive)

Persist a **poller-immune spawn identity** and re-drive a herdr-restored account pane through
`onSpawn` so the account is re-applied:

- **Persist** two nullable session columns via a single writer `persistSpawnIdentity`:
  `spawnTerminalId` (the terminalId Shepherd itself spawned — written **only** by spawn-finish, never
  by reconcile/poller, so re-pointing `herdrAgentId` can't mask a restored husk) and `spawnAccountDir`
  (the folded **plugin** `patchEnv.CLAUDE_CONFIG_DIR`). Sticky/verified: a null/wrong re-derivation
  (plugin state lost) **preserves** the prior owning dir and does **not** advance `spawnTerminalId` —
  so a re-drive can never silently "heal" onto the default account.
- **Detect** with `needsAccountRedrive(s, agent) = spawnAccountDir != null && agent.terminalId != spawnTerminalId`.
- **Re-drive at three convergent loci:** the poller proactively (`reconcileAgent` → fire-and-forget
  `reDriveAccount`), the human parked-resume (`resumeInner` adopt branch falls through to teardown +
  re-drive), and the autonomous steer paths (autopilot / automerge / retryHalted / critic-address /
  plan-gate defer via `shouldDeferSteer` and route through `resume()` first).
- **Serialize + bound:** a per-session in-flight guard inside `resume()` coalesces every entrypoint
  (poller, human, "bring agent back" button, autonomous) so there is never a double `herdr.start`; a
  bounded counter anchored on `spawnTerminalId` gives up after 3 failed re-drives (→ DEGRADED, loudly
  logged, steering resumes as-today), so a usage-halted/degraded account can never wedge into an
  every-tick re-drive or a permanent steer-defer.

### Scope

- **Plugin/account sessions only.** api-key sessions are **excluded by construction**:
  `spawnAccountDir` folds only the plugin `patchEnv.CLAUDE_CONFIG_DIR`, not the api-key
  `apiKeyPassthrough` mirror — and that mirror symlinks `projects/` to the real `~/.claude`, so a bare
  resume still finds the conversation (no bug there).
- **No behavior change for default/no-plugin users**: empty patch → `spawnAccountDir` null →
  `needsAccountRedrive` false everywhere; the poller check short-circuits on the null before any
  `herdr.list()`.
- Server-internal fix — no UI, no user-facing strings.

## Why core, not herdr

The "fix at source" (herdr persist/restore the spawn env) is unavailable: herdr's claude integration
is detection-only and its restore strips env one-way, and an upstream-herdr fix + a PATH `claude` shim
were both explicitly rejected in favor of a core fix.

## Testing

- New TDD suites (red→green): `spawn-identity`, `resume-serialize`, `resume-account-redrive`,
  `poller-redrive`, `redrive-bounded`, `steer-defer` (+ automerge/autopilot deferSteer cases).
- `bun run lint`, `bun run typecheck`, `bun run test` (6008 pass / 0 fail), and `fallow audit` all
  green. Rebased onto latest `main`.
- Live herdr-restart repro deliberately not run (would disrupt the operator's other running sessions);
  verified via the TDD suite + a whole-branch review that traced the end-to-end flow + Phase-0
  live-state analysis.

[no-feature-entry] — server-internal fix, no user-facing UX.

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -280,10 +280,10 @@
       {
         "files": ["src/store.ts"],
         "functions": ["hydrate"],
-        "maxCyclomatic": 20,
-        "maxCognitive": 20,
+        "maxCyclomatic": 22,
+        "maxCognitive": 22,
         "maxCrap": 1000,
-        "reason": "Codex/default-provider persistence extends the existing hydrate migration path by one branch; keep capped until hydrate migrations are extracted."
+        "reason": "Codex/default-provider persistence + the spawn-identity columns (spawnTerminalId/spawnAccountDir, herdr-restart account recovery) extend the flat hydrate row→Session mapping by two nullable-field branches. Cyclomatic is field-count-driven, not logic — keep capped until hydrate migrations are extracted."
       },
       {
         "files": ["src/store.ts"],
--- a/src/automerge.ts
+++ b/src/automerge.ts
@@ -69,6 +69,8 @@ export interface AutoMergeDeps {
   prCache: { snapshot(): Record<string, GitState> };
   /** Whether the session's herdr pane is live (so a steer lands). */
   paneAlive: (id: string) => boolean;
+  /** Defer steering while a herdr-restored account pane still needs a re-drive (SessionService.shouldDeferSteer). */
+  deferSteer?: (id: string) => boolean;
   repos: () => string[];
   emitStatus: (s: AutoMergeStatus) => void;
   emitArchived: (id: string) => void;
@@ -355,7 +357,9 @@ export class AutoMergeService {
     const s = this.deps.store.get(sessionId);
     if (!s) return;
     this.deps.emitStatus(this.status(repoPath, true, "rebasing", s.desig, s.id));
-    if (!this.deps.paneAlive(sessionId)) {
+    if (!this.deps.paneAlive(sessionId) || this.deps.deferSteer?.(sessionId)) {
+      // Not live, OR a herdr-restored account husk that must be re-driven first (else the rebase steer
+      // lands on the wrong-account pane). resume() re-drives (Locus B) before we reply below.
       if (!(await this.deps.service.resume(sessionId))) return;
     }
     if (this.deps.service.reply(sessionId, rebaseSteer(s.baseBranch))) {
--- a/src/autopilot.ts
+++ b/src/autopilot.ts
@@ -116,6 +116,8 @@ export interface AutopilotDeps {
   resume: (id: string) => unknown;
   /** Whether the session's herdr pane is currently live. */
   paneAlive: (id: string) => boolean;
+  /** Defer steering while a herdr-restored account pane still needs a re-drive (SessionService.shouldDeferSteer). */
+  deferSteer?: (id: string) => boolean;
   /** Visible terminal tail for a session (herdr.read → tailLines). */
   readTail: (id: string) => string[];
   /** Whether the session already has a PR in any state (open/merged/closed). True → autopilot
@@ -290,9 +292,10 @@ export class AutopilotService {
   /** Send `text` into the session, resuming an exited pane first. Returns whether the steer
    *  landed; does NOT bump the step (the caller decides whether the attempt counts). */
   private async sendSteer(s: Session, text: string): Promise<boolean> {
-    if (!this.deps.paneAlive(s.id)) {
-      // Exited pane: resume so there's something to steer. resume() resolves falsy when it
-      // can't (archived / no pinned session id) — then there's nothing to do.
+    if (!this.deps.paneAlive(s.id) || this.deps.deferSteer?.(s.id)) {
+      // Not live, OR a herdr-restored account husk to re-drive first (Locus B) so the steer lands on the
+      // re-driven pane, not the wrong-account husk. resume() resolves falsy when it can't (archived /
+      // no pinned session id) — then there's nothing to do.
       if (!(await this.deps.resume(s.id))) return false;
     }
     return this.deps.steer(s.id, text);
--- a/src/herdr.ts
+++ b/src/herdr.ts
@@ -121,6 +121,20 @@ export function matchAgent(
 }
 
 /**
+ * True when `agent` is a pane herdr re-created (its terminalId is NOT the one Shepherd last spawned
+ * on the owning account) for a session that HAS an owning plugin account. Such a pane is a bare
+ * `claude --resume` under the wrong CLAUDE_CONFIG_DIR and must be re-driven through onSpawn, not
+ * adopted/steered. Keys on spawnTerminalId (written only by the spawn-finish path, never by
+ * reconcile/poller) so their re-pointing of herdrAgentId cannot mask a herdr-restored husk.
+ */
+export function needsAccountRedrive(
+  s: { spawnAccountDir: string | null; spawnTerminalId: string | null },
+  agent: { terminalId: string },
+): boolean {
+  return s.spawnAccountDir !== null && agent.terminalId !== s.spawnTerminalId;
+}
+
+/**
  * Pick the cwd-fallback agent for one still-unmatched session from the untaken
  * candidates. When the session contends for its cwd with another active session, only
  * an unambiguous agent-NAME match is safe; a sole session adopts its lone cwd agent via
--- a/src/hold-service.test.ts
+++ b/src/hold-service.test.ts
@@ -57,6 +57,8 @@ function makeSession(overrides: Partial<Session> = {}): Session {
     experimentId: null,
     experimentRole: null,
     completionRepromptCount: 0,
+    spawnTerminalId: null,
+    spawnAccountDir: null,
     ...overrides,
   };
 }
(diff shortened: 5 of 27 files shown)


scoop (commented):
## Summary

This PR implements the herdr-restart account-loss fix: it adds poller/reconcile-immune spawn-identity markers (`spawnTerminalId`/`spawnAccountDir`), detects a herdr-restored wrong-account pane via `needsAccountRedrive`, proactively re-drives it (bounded), and defers autonomous steers onto the husk until it heals or the bounded retry gives up. I verified the full chain and found no blocking defects.

### Verified correctness claims (ground truth cited)

- **Column plumbing is internally consistent.** `spawnTerminalId, spawnAccountDir` are appended after `experimentRole` in `COLS` (src/store.ts:237), the `SessionRow` type (src/store.ts:293-294), the `INSERT` (51 columns = 51 value expressions, in matching order — src/store.ts:1769 placeholders vs. src/store.ts:1820-1821 values), the migration `add()` calls (src/store.ts:3038-3039), and `hydrate` (src/store.ts:4328-4329). `NewSession` omits both (src/store.ts:208-209) and `buildSessionRow` defaults them null (src/store.ts:1758-1759). Matches the `Session` interface additions (src/types.ts:103-113).
- **`persistSpawnIdentity` is the sole writer and its sticky rule is sound** (src/service.ts:2243-2270). The invariant “`spawnAccountDir` non-null ⇒ `spawnTerminalId` non-null” holds because the two non-preserve rows always set the terminal to a non-null `outcome.terminalId` (src/service.ts:2255) alongside a non-null dir, and the preserve row copies both verbatim (src/service.ts:2266). This is what makes `reDriveAccount`’s `anchor` non-null for account sessions.
- **Every spawn that rewrites `herdrAgentId` also persists identity.** finishResumeSpawn (src/service.ts:2214 → 2224), create (src/service.ts:2474 → 2495), and the resume-catch path (src/service.ts:2747 → 2774). The one `herdrAgentId` write that does NOT persist identity — `adoptLiveResumeAgent` (src/service.ts:2178) — is provably safe: resumeInner only adopts when `!needsAccountRedrive` (src/service.ts:3128), so an adopted account pane already has `agent.terminalId === spawnTerminalId` and cannot drift the marker. Default sessions (`spawnAccountDir === null`) never arm `needsAccountRedrive` (src/herdr.ts:89).
- **`reDriveAccount` bound is anchored correctly** on `spawnTerminalId` (src/service.ts:3195), which is stable across unhealed/refused attempts (preserve row) and only advances on a heal — so the CAP (src/service.ts:3197) is actually reachable, unlike a husk-terminalId-keyed counter. The verdict `s.spawnTerminalId !== anchor ? healed : unhealed` (src/service.ts:3201) agrees with `persistSpawnIdentity`’s action in every row.
- **`shouldDeferSteer` (src/service.ts:3224-3233) never permanently stalls a steer.** Its `exhausted` predicate (`rec.attempts >= REDRIVE_CAP` on the same anchor) is byte-for-byte the same condition under which `reDriveAccount` returns `degraded` (src/service.ts:3197), and the poller fires `reDriveAccount` every tick for a live armed husk (src/poller.ts:427-434 → src/poller.ts:439), so the counter always advances to the cap. Once degraded, steering resumes on the current pane — documented “no-worse-than-today”.
- **The defer paths do not drop findings.** `autoAddress`/plan-gate `reply` returning `false` (src/index.ts:1054, src/index.ts:1102) reuses the exact pre-existing dead-pane semantics in `runAutoAddress` — `return delivered ? priorRound + 1 : priorRound` (src/review.ts:927): the round simply doesn’t advance and retries next cycle; the review comment is still posted. `retryHalted` routes a deferred husk into the `else` → `resume(id)` branch (src/service.ts:3436-3440), which re-drives (heals) it rather than steering the wrong account. automerge (src/automerge.ts:360) and autopilot (src/autopilot.ts:295) `resume()` before replying on a deferred id.
- **`resume()` coalescing guard** (src/service.ts:3113-3119) is leak-safe (`.finally` deletes on both settle paths) and semantically neutral for the new caller: `reDriveAccount` passes `force:true`, but for a `needsAccountRedrive` husk the non-forced path also falls through and re-drives (src/service.ts:3128), so coalescing onto an in-flight non-forced resume is equivalent.

### Non-blocking nit

- `persistSpawnIdentity`’s `"healed" | "unhealed"` return value (src/service.ts:2246) is dead at all three call sites (finishResumeSpawn/create/resume-catch ignore it), and `reDriveAccount` recomputes the same verdict independently from `s.spawnTerminalId !== anchor` (src/service.ts:3201) rather than consuming it. Harmless, but the return type reads as if it drives control flow when nothing uses it — consider dropping it or wiring `reDriveAccount` to use it.

### Stated limitations

- Could not read `/tmp/handoff-herdr-resume-account-loss.md` (no longer present at that path); I reviewed against the diff’s own extensive doc-comments, which describe the task (“herdr-restart account-loss fix”) task-by-task, and confirmed the code matches them.
- This is a backend TS service with no i18n message catalogs in the diff’s scope, so no locale-parity check applied.

scoop (commented):
## Re-review — herdr-restart account-loss fix

### Prior finding (re-verified)

**RESOLVED** — `persistSpawnIdentity`'s unused `"healed" | "unhealed"` return type is gone. The helper now returns `void` (src/service.ts:2248), and `reDriveAccount` derives the heal verdict independently from whether `spawnTerminalId` advanced past the captured `anchor` (src/service.ts:3202: `s.spawnTerminalId !== anchor ? "healed" : "unhealed"`). The signature no longer implies control flow that doesn't exist. Head commit 82832efc ("drop persistSpawnIdentity's unused return type") matches. Not re-raised.

### Verification performed

- **Spawn-identity ownership semantics.** `spawnAccountDir` is set to `patchEnv.CLAUDE_CONFIG_DIR ?? null` (src/service.ts:1801). Verified `patchEnv` is built solely from the plugin onSpawn overlay via `foldSpawnPatch` (src/service.ts:1686), whose CLAUDE_CONFIG_DIR comes only from `patch.credentialDir`/`patch.env` (src/spawn-membrane.ts:127). The api-key credential-less mirror rides `apiKeyPassthrough`/`spawnEnv`, NOT `patchEnv` (src/service.ts:1784) — so the comment's claim that the api-key mirror is excluded holds, and a default/api-key session correctly gets `spawnAccountDir=null` and never arms `needsAccountRedrive`.
- **`needsAccountRedrive`** (src/herdr.ts:130) keys on `spawnTerminalId` (written only by `persistSpawnIdentity` → `setSpawnIdentity`, never by reconcile/poller), so the poller re-pointing `herdrAgentId` to a restored pane cannot mask the husk. Confirmed poller's `reconcileAgent` only writes `herdrAgentId`, never the spawn-identity columns.
- **Sticky preserve rule** (src/service.ts:2249-2265) never nulls a non-null `spawnAccountDir`; the `folded===null && prior!==null` branch preserves the marker and does NOT advance the terminal, keeping `needsAccountRedrive` armed for an unhealed re-drive — matching the documented table.
- **Bounded re-drive.** Anchor = `spawnTerminalId` is stable across unhealed/refused attempts (they don't advance it) and only clears on a heal, so the `REDRIVE_CAP=3` counter converges instead of resetting on the churning live terminalId. Verified `degraded` short-circuits before any spawn (src/service.ts:3198). The permanent-degraded-until-manual-resume behavior is the documented "no-worse-than-today" tradeoff, not a defect (a human `resume(id)` re-drives and can heal).
- **Serialization guard.** `resume` coalesces per-id via `resumeInFlight` with a synchronous get/set (no await between), `.finally` deletes on both resolve and reject. First caller's opts win — documented.
- **Steer-defer wiring.** `shouldDeferSteer` (src/service.ts:3225) consulted at all steer sites: index.ts autoAddress (1051) + plan-gate reply (1105) return `false` to hold the round; autopilot.sendSteer (autopilot.ts:295) and automerge rebase (automerge.ts:360) route through `resume()`; retryHalted (service.ts:3432) falls to the `resume()` else-branch. All re-drive rather than steer the wrong-account husk.
- **Store plumbing.** COLS (51 cols), INSERT placeholders, `buildSessionRow` value list, `NewSession` Omit, `SessionRow`, migration `add()` calls, and `rowToSession` all include the two new nullable columns consistently (src/store.ts:229-238, 1769-1821, 2789, 3037-3040, 4328). Backend `Session` literals in the 11 touched test fixtures were updated for the two new required fields.

Could not verify the live plugin onSpawn / herdr restart runtime behavior end-to-end (external to this repo); assessed against the code paths as written.

Nothing to change.

Issue #1022, opened 2026-06-23
Review-in-flight terminal signal (non-blocking) + guard seam


scoop:
## Problem

When a **PR-critic** or **plan-gate** review concludes, Shepherd auto-pastes the outcome into the live PTY (`service.reply()` → bracketed paste + `\r`): critic findings, plan-revision findings, or — for autopilot/auto plan sessions — the release-into-execution kick. The review itself runs fully isolated (own worktree + herdr terminal), so the main session terminal stays interactive *while it runs*. The collision happens at **conclusion**: if a human is typing in that terminal when the directive lands, their input and the injected steer interleave.

We want to **encourage the operator to wait — without blocking the terminal** — and only when a paste can actually happen given current toggles ("respect the automation toggles": if nothing will be auto-pasted, don't warn).

This issue scopes a **non-blocking signal now, architected with the seam for a future guard.**

## When a paste actually lands (verified against `src/review.ts` / `src/plan-gate.ts`)

| Review | Pastes into live PTY on conclusion when… | Governing toggle |
|---|---|---|
| **PR critic** | `changes_requested`/`comment` + PR open + findings>0 + **`autoAddressEnabled`** + `addressRound < cap` + pane alive | **`autoAddressEnabled`** (per-repo) |
| **Plan-gate → changes_requested** | always (under `cap`, pane alive) | *none* — revising the plan *is* the loop |
| **Plan-gate → approved** | releases into execution only if `auto \|\| effectiveAutopilot` | **autopilot** |

Key subtlety: **while a review is in-flight we don't yet know its verdict**, so the in-flight signal is necessarily speculative — "a directive *may* land when this finishes."

## Behavior — non-blocking signal

New **in-terminal banner** in `ui/src/lib/components/Viewport.svelte`, mounted alongside the existing `CriticBadge`/`PlanGateBadge` (~lines 50–51). **Herd-grid badges are untouched** — they remain the list-level "under review" signal; this banner is the keyboard-co-located warning.

**Shown iff a paste *could* land on conclusion, per current toggles:**
- **Critic in-flight** → `repoConfig.autoAddress[repoPath] === true` AND `addressRound < addressCap` (treat no-prior-verdict as round 0). Auto-address off ⇒ **no banner**.
- **Plan-gate in-flight** → **always** while `planGates.isReviewing(id)` (covers both the always-on changes-requested steer and the autopilot approval-release).

**States:**
[...]

PR #1434, opened 2026-07-05
fix(ui): reflow terminal above review/CI banner instead of covering the prompt

scoop:
## Problem

The review-in-flight banner (and its CI-running twin) rendered as `position: absolute; bottom: 0` strips **over** the bottom of the xterm terminal — exactly where the live prompt/cursor sits. So while a review was in flight you couldn't see or use the prompt, even though you may well want to type/steer.

They were made absolute *specifically to avoid an xterm refit* — which is precisely why they covered the prompt.

## Fix — reserve space, don't overlay

Make `.term-mount` give up the strip:

```css
height: calc(100% - var(--review-banner-h, 0px));
min-height: 4rem;
```

`--review-banner-h` is **already** published on `.vp-body` (each banner measures its own `offsetHeight`; today only the floating scroll-jump button reads it). Shrinking the mount trips its existing `ResizeObserver` → `refit()`, so xterm reflows and the live prompt lands right above the banner — visible and typeable. One CSS change covers **both** banners (shared var, mutually exclusive) and **all** tiers (calm / escalated / every conclusion tone / CI-running), since each tier's real height flows through the same var.

### Short-terminal guard (`min-height: 4rem`)

The floor equals `.vp-body`'s own `min-height` and must **not** exceed it. When `.vp-body` is squeezed to its 4rem floor (the expanded-SessionRecap takeover), an unfloored mount would drop to ~1 row and SIGWINCH the running Claude Code TUI to a 1-row terminal. At the 4rem floor the mount equals the body (~3-4 rows), so the PTY is never driven to xterm's clamped 1-row minimum and the banner gracefully falls back to overlaying — today's behavior. A larger floor would force the mount taller than the body and clip the prompt via `overflow:hidden` **even with no banner**, so 4rem is the cap.

## Tradeoff (accepted with operator)

Reflow means the strip appearing / changing tier / disappearing drives a real PTY resize (up to ~3 real TUI redraws per review cycle — there's no client-side de-dup; only the kernel `TIOCSWINSZ` ignores same-dimension writes). The zero-refit alternative (pinning the strip to the **top** of the terminal) was weighed and held as the fallback — it removes the resize on the typing axis but floats over old scrollback and detaches the warning from the input. Operator chose bottom-reflow; if mid-review typing ever proves hostile to the resize, top-pin is the ready fallback.

## Verification

- `cd ui && bun run check` → 0 errors (2 pre-existing warnings in unrelated files).
- `cd ui && bun run test` → **2851 passed** (209 files).
- prettier + eslint (lint-staged) clean on the 3 changed files.

**Not exercised headlessly:** the live "type continuously while a review starts and finishes" gate needs a running herdr + a real in-flight review, which this environment can't stand up. The change is low-risk by construction — keystrokes flow through `term.onData` → WS independently of resize frames, so a SIGWINCH doesn't drop queued input — and the top-pin fallback above is ready if the operator observes any input drop/corruption during a real in-flight review.

_No new user-facing strings, glossary terms, or capabilities — a layout fix, so no i18n / feature-catalog changes. Context: issue #1022._

### DIFF
--- a/ui/src/lib/components/Viewport.svelte
+++ b/ui/src/lib/components/Viewport.svelte
@@ -2241,9 +2241,11 @@
       {reattach}
       {resumeSession}
     />
-    <!-- Non-blocking review-in-flight banner: bottom strip over the terminal,
-         above the steer bar. Stays mounted across tab switches (state survives);
-         it suppresses its own render off the term tab. (issue #1022) -->
+    <!-- Non-blocking review-in-flight banner: reserved bottom strip. The terminal
+         (.term-mount) shrinks by this banner's height (--review-banner-h) so the
+         strip sits BELOW the live prompt, not over it. Stays mounted across tab
+         switches (state survives); it suppresses its own render off the term tab.
+         (issue #1022) -->
     <ReviewInFlightBanner
       {session}
       keystrokes={opKeystrokes}
@@ -3027,7 +3029,18 @@
 
   .term-mount {
     width: 100%;
-    height: 100%;
+    /* Reserve the review/CI banner's bottom strip so xterm reflows ABOVE it instead
+       of being overlaid — the live prompt stays visible while a review runs.
+       --review-banner-h is published on .vp-body (0px when no banner shows).
+       Changing this height trips the ResizeObserver on this element → refit() → PTY
+       resize, which is the intended reflow. The min-height floor equals .vp-body's
+       own min-height (4rem) and must NOT exceed it: a larger floor would force the
+       mount taller than a squeezed body and clip the bottom prompt row via
+       .vp-body's overflow:hidden even with no banner. At the floor (squeezed-recap
+       takeover) the mount == body, so the banner overlaps its bottom again — prior
+       behavior — rather than resizing the PTY toward xterm's clamped 1-row minimum. */
+    height: calc(100% - var(--review-banner-h, 0px));
+    min-height: 4rem;
     overflow: hidden;
     /* we drive vertical scroll via touch handlers; keep the browser out of it */
     touch-action: none;
--- a/ui/src/lib/components/viewport/CiRunningBanner.svelte
+++ b/ui/src/lib/components/viewport/CiRunningBanner.svelte
@@ -85,9 +85,11 @@
 {/if}
 
 <style>
-  /* Bottom overlay strip pinned to the terminal body, directly above the steer
-     bar — mirrors ReviewInFlightBanner. Absolutely positioned so it never triggers
-     an xterm refit. Non-blocking: no scrim/blur (it does not seize interaction). */
+  /* Bottom strip pinned to the terminal body, directly above the steer bar —
+     mirrors ReviewInFlightBanner. The terminal reserves this height (.term-mount
+     shrinks by --review-banner-h) so the strip sits BELOW the live prompt, not over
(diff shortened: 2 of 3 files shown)


scoop (commented):
## Summary

The PR stops the review/CI in-flight banners from overlaying the live terminal prompt by *reserving* the strip's height instead of overlaying it. `.term-mount` now shrinks by `--review-banner-h` (published on `.vp-body`), so xterm reflows above the banner and the prompt stays visible/typeable while a review runs. This directly satisfies the task. The two banner files carry comment-only changes.

- **`--review-banner-h` source:** published on `.vp-body` at `ui/src/lib/components/Viewport.svelte:2195` as `${reviewBannerH || ciBannerH}px`, resolving to `0px` when no banner shows. The new `calc(100% - var(--review-banner-h, 0px))` at `Viewport.svelte:3042` inherits it correctly (custom props inherit to the child `.term-mount`).
- **Reflow trigger:** the `ResizeObserver` at `Viewport.svelte:1560-1566` observes `el` (the `.term-mount`, bound at 2199-2203) and calls `refit()` (`Viewport.svelte:1215-1219`) → `c.resize(term.cols, term.rows)`. So shrinking the mount does drive the PTY resize the comment describes.
- **No refit feedback loop:** the reserved height equals the banner's own `offsetHeight` (`ReviewInFlightBanner.svelte:189`/`201`, `CiRunningBanner.svelte:40`/`50`), which depends on body width + content, not on `.term-mount` height. Resizing the mount can't change the banner height, so no oscillation.
- **Single reservation:** `CiRunningBanner` suppresses itself while a review claims the strip via `reviewActive` (`CiRunningBanner.svelte:35`, wired from `reviewActive` bind at `Viewport.svelte:2254-2258`), and `reviewBannerH || ciBannerH` selects whichever is active — the two never both reserve.
- **Off-term tab:** both banners set `height = 0` when `tab !== "term"` (`ReviewInFlightBanner.svelte:183-187`, `CiRunningBanner.svelte:35-38`), and `.term-mount` is `display:none` off-term, so `refit()`'s `offsetParent === null` guard (line 1216) skips — no spurious resize.
- **`min-height` floor:** the added `min-height: 4rem` on `.term-mount` equals `.vp-body`'s `min-height: 4rem` (`Viewport.svelte:3000`). Because the floor is not larger than the body's floor, the mount can never be forced taller than a squeezed body, so no bottom-row clip under `.vp-body { overflow: hidden }`. The comment's reasoning checks out.
- **Only JS consumer of mount size** is `refit()` reading `el.clientHeight` (line 1216); the reduced height is exactly the intended visible area above the strip. No other code assumes `.term-mount == 100%`.

## Observations (non-blocking, within the acknowledged tradeoff)

- The code comment notes that at the exact squeezed-recap floor the banner overlaps the mount again (prior behavior). In practice the overlap begins slightly *before* the exact floor: for body heights in the narrow window `[4rem, 4rem + bannerH)`, `calc(100% - bannerH)` falls below the `4rem` `min-height` and clamps, so the banner covers the bottom of the mount. This is the same graceful-degradation direction the comment describes and only bites under the expanded-recap takeover, so it's not a regression from the pre-PR overlay behavior — noting for completeness, no change required.

No correctness, security, or i18n-parity issues found (this PR touches no message keys). Decision: comment.

PR #1436, opened 2026-07-05
feat(ui): clickable banner for MCP OAuth auth URLs above the terminal

scoop:
## Problem

When an agent hits an MCP OAuth flow (Notion, Vercel, …) it prints a long authorize URL and blocks, asking the operator to open it in their browser. Claude's TUI **word-wraps that URL across several indented terminal lines**, so in the xterm buffer it's fragmented — the existing `WebLinksAddon` never linkifies it, and there's **no OSC 52** to copy it (unlike the Claude *login* screen handled in #1427). The operator can neither click nor cleanly copy it.

This is the "clickable URL" piece deferred from the OSC 52 login work (#1427), generalized to MCP auth.

## Step 0 — empirical source gate (done before building)

The plan hinged on an unproven assumption: does the URL actually reach a source we can read un-wrapped? A screenshot only proves the *PTY buffer* (fragmented). Captured real MCP OAuth flows on disk (Vercel MCP, `claude-cowork-sales`) and confirmed:

- The full **380-char un-wrapped** authorize URL appears **verbatim in the JSONL transcript**, in **two** record shapes in the same flow:
  - an `assistant` `text` block (Claude relaying it — the screenshot's shape), and
  - a `tool_result` block carried inside a **`role:"user"`** message (the raw MCP payload).
- Because tool_results are `role:"user"`, an assistant-only scan (`assistantSideText`) would **miss** the tool_result delivery — so the detector scans both.
- A/B on the real transcript: the detector returns the URL at the blocked moment and `null` once the operator has answered (no stale resurface).

## Approach

- **`detectAuthUrl(transcript)`** (`src/auth-url.ts`, pure + unit-tested): walks the JSONL, tracks the *pending* authorize URL (set by an assistant-text or tool_result authorize URL; cleared by operator input), so a resolved URL isn't resurfaced on a later unrelated block. `http/https` + OAuth-authorize heuristic; the `localhost/callback` return URL is correctly excluded.
- **Carried on the existing block channel**, not a parallel pipeline: `BlockReason.authUrl?` is populated in the poller's `maybeClassify` for `awaiting-input` blocks (a bounded, throttled tail read, gated to that shape). It rides the existing `session:block` dedup + UI, and clears with the block. Detection lives in `maybeClassify`, reached by **both** the herdr-blocked branch and the hook-awaiting-input path — so it fires in the awaiting state even if herdr hasn't latched blocked.
- **`GET /api/blocks` snapshot** (mirrors `/api/activity`): blocks were edge-emitted only, so a push-then-open / fresh load showed no block on first paint. Now current blocks (incl. the auth URL) bootstrap — a small, general fix.
- **Banner** in `ViewportTermBanners` above the terminal (non-modal, so it never covers the bottom prompt): the full URL shown for host-vetting, **Open** (`window.open`, `noopener`) and **Copy**. Design tokens only, EN+DE, ≥16px title / 44px touch targets.

## Scope

Outbound URL only (the return-trip callback paste uses the existing compose flow). Task sessions only — the Claude account-login pane (no transcript) keeps its OSC 52 copy from #1427.

## Security

The URL originates from agent output (which may relay untrusted MCP/issue content). Mitigations: the banner **shows the full URL** for vetting, **http/https only**, **explicit operator click** (never automatic), `noopener,noreferrer`. No wider than the terminal's existing click-to-open links.

## Testing

- `src/auth-url.test.ts` — detector (assistant text, tool_result, Vercel `/oauth/authorize`, callback-URL exclusion, staleness-clear, malformed lines).
- `test/poller.test.ts` — awaiting-input block carries + snapshots the URL; not consulted for menu blocks; clears on resume; `GET /api/blocks`.
- `ui` store + `ViewportTermBanners.browser.test.ts` (render / Open / Copy / hidden-when-null).
- Full suites green: root lint + 6005 tests; UI `check` (0 errors) + `check:i18n` + 2859 tests; fallow audit clean.

### DIFF
--- a/src/auth-url.test.ts
+++ b/src/auth-url.test.ts
@@ -1,0 +1,104 @@
+import { describe, it, expect } from "bun:test";
+import { detectAuthUrl } from "./auth-url";
+
+// Real-shape fixtures (synthetic client_id / code_challenge). Mirrors the Step-0 capture:
+// the authorize URL lands in an assistant `text` block AND/OR a `tool_result` block inside a
+// `role:"user"` message. `code_challenge` is a PKCE hash placeholder here.
+const AUTH_URL =
+  "https://mcp.notion.com/authorize?response_type=code&client_id=abc123&code_challenge=HOkFuV4sAW3&code_challenge_method=S256&redirect_uri=http%3A%2F%2Flocalhost%3A3118%2Fcallback&resource=https%3A%2F%2Fmcp.notion.com%2Fmcp";
+
+const assistantText = (text: string) =>
+  JSON.stringify({
+    type: "assistant",
+    message: { role: "assistant", content: [{ type: "text", text }] },
+  });
+const toolResult = (content: string) =>
+  JSON.stringify({
+    type: "user",
+    message: { role: "user", content: [{ type: "tool_result", content }] },
+  });
+const opInputString = (text: string) =>
+  JSON.stringify({ type: "user", message: { role: "user", content: text } });
+const opInputText = (text: string) =>
+  JSON.stringify({ type: "user", message: { role: "user", content: [{ type: "text", text }] } });
+const toolUse = () =>
+  JSON.stringify({
+    type: "assistant",
+    message: { role: "assistant", content: [{ type: "tool_use", name: "notion" }] },
+  });
+
+describe("detectAuthUrl", () => {
+  it("detects the URL in an assistant text block", () => {
+    const jsonl = [
+      opInputString("read this notion page for me"),
+      assistantText(
+        `I need you to authorize the Notion connection.\n1. Open this URL in your browser:\n\n${AUTH_URL}\n\n2. Copy the callback URL back here.`,
+      ),
+    ].join("\n");
+    expect(detectAuthUrl(jsonl)).toBe(AUTH_URL);
+  });
+
+  it("detects the URL in a tool_result (role:user) — the assistant-only scan would miss it", () => {
+    const jsonl = [
+      opInputString("read this notion page"),
+      toolUse(),
+      toolResult(`Authorization required. Visit ${AUTH_URL} to authorize.`),
+    ].join("\n");
+    expect(detectAuthUrl(jsonl)).toBe(AUTH_URL);
+  });
+
+  it("detects a Vercel-style /oauth/authorize URL", () => {
+    const vercel =
+      "https://vercel.com/oauth/authorize?response_type=code&client_id=cl_Wbdt&code_challenge=x&code_challenge_method=S256&redirect_uri=http%3A%2F%2Flocalhost%3A9999%2Fcallback";
+    expect(detectAuthUrl(assistantText(`Open ${vercel} then paste the callback.`))).toBe(vercel);
+  });
+
+  it("returns null for a plain non-auth URL", () => {
+    expect(
+      detectAuthUrl(assistantText("See the docs at https://example.com/guide/setup for details.")),
+    ).toBeNull();
+  });
+
+  it("does not match the localhost callback URL the operator pastes back", () => {
+    const callback = "http://localhost:3118/callback?code=xyz789&state=abc";
+    expect(detectAuthUrl(opInputString(`Here is the callback: ${callback}`))).toBeNull();
+    expect(detectAuthUrl(toolResult(`redirected to ${callback}`))).toBeNull();
+  });
+
+  it("clears a resolved URL once the operator has responded (no stale resurface)", () => {
+    const jsonl = [
+      assistantText(`Open ${AUTH_URL}`),
+      opInputString("http://localhost:3118/callback?code=done"), // operator pasted the callback
+      toolUse(),
+      assistantText("Thanks, I read the page. Anything else?"),
+    ].join("\n");
+    expect(detectAuthUrl(jsonl)).toBeNull();
+  });
+
+  it("clears when the operator answers via a typed text turn", () => {
+    const jsonl = [
+      toolResult(`authorize at ${AUTH_URL}`),
+      opInputText("skip it, read the page directly instead"),
+    ].join("\n");
+    expect(detectAuthUrl(jsonl)).toBeNull();
+  });
+
+  it("surfaces the current URL while still blocked (URL is the newest relevant content)", () => {
+    const jsonl = [
+      opInputString("original request"),
+      toolUse(),
+      toolResult(`auth required: ${AUTH_URL}`),
+      assistantText(`Please open ${AUTH_URL} and paste the callback.`),
+    ].join("\n");
+    expect(detectAuthUrl(jsonl)).toBe(AUTH_URL);
+  });
+
+  it("round-trips a long URL byte-exactly and skips malformed lines", () => {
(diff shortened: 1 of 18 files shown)


scoop (commented):
## Summary

This PR surfaces a pending MCP OAuth authorize URL (Notion/Vercel-style flows) that Claude word-wraps un-clickably in the PTY, by reading the intact URL from the JSONL transcript and rendering a non-modal top banner with Open/Copy actions. It's a clean, well-scoped extension of the existing block-reason plumbing, and it's thoroughly tested (detector, poller wiring, store, and a browser test for the banner).

The design is sound and I could not find a blocking defect. Decision is **comment** with one non-blocking nit.

## Verification performed

- **Detector logic** (`src/auth-url.ts`): the oldest→newest walk sets `pending` on an assistant `text` block or a `tool_result`-carrying auth URL and clears on genuine operator input (a `role:"user"` message with no `tool_result`). `isAuthUrl` correctly excludes the `/callback?code=…` return URL (path isn't `authorize`, no `response_type=code`, no `code_challenge`) — matches the test at `src/auth-url.test.ts:68`. `eachJsonlObject` confirmed exported at `src/jsonl.ts:3`; `readTranscriptTail` at `src/activity.ts:158`. Bounded-tail truncation can only produce a false *negative* (URL scrolled out), never a stale false positive, since the clear is always newer than the URL it clears.
- **`javascript:` safety**: `URL_RE` only matches `https?://` and `isAuthUrl` rejects any non-http(s) protocol (`src/auth-url.ts:155`), so `window.open` can only ever receive an http/https URL. `noopener,noreferrer` is passed (`ViewportTermBanners.svelte:42`).
- **Poller wiring**: all block fire/clear paths now route through `emitBlock` (grep confirms the only remaining `this.onBlock(` is inside `emitBlock` itself, `src/poller.ts:899`), so `lastBlockReason`/`blockSnapshot()` stay in step. `authUrl` is mutated onto `reason` *before* `JSON.stringify(reason)` (`src/poller.ts:987`→`989`), so it correctly rides the `lastSig` dedup and the snapshot. Both `maybeClassify(s, …)` call sites updated (`src/poller.ts:493`, `:599`). The read is gated to `awaiting-input` and throttled by the `reclassifyMs` gate; default `readAuthUrl` swallows read errors → null (`src/poller.ts:94`).
- **Server/route**: `handleBlocksSnapshot` registered in `ROUTE_HANDLERS` (`src/server.ts:515`); `AppDeps.blocks` typed with `BlockReason` imported from `./blocked` (which now carries `authUrl`, `src/blocked.ts:269`).
- **Store/UI**: `HerdStore.blocks` is `{reason, since}` (verified `store.svelte.ts:57`, `:329`); `setBlocks` builds the same shape (`:197`); `selectedAuthUrl` reads `store.blocks[id]?.reason.authUrl` and both Viewport instances receive it (`+page.svelte:2534`, `:2696`).
- **i18n parity**: all six keys present in BOTH catalogs — verified `ui/messages/en.json:637-640,2545-2546` and `ui/messages/de.json:637-640,2545-2546`.
- **Feature announcement**: `FeatureAnnouncement` requires `id`/`sinceVersion`/`titleKey`/`bodyKey`; the entry supplies all four with keys that exist in both catalogs, and entries auto-load via `import.meta.glob` (`ui/src/lib/feature-announcements.ts:47`) so no manual registration is needed. `sinceVersion` 1.42.0 matches sibling entries in this release (CI gate owns exact-version validation).

## Nit (non-blocking)

`copyAuth` uses `await navigator.clipboard?.writeText(authUrl)` (`ViewportTermBanners.svelte:47`). In an insecure context (e.g. Shepherd accessed over plain `http://<lan-ip>`), `navigator.clipboard` is `undefined`, so the optional chain short-circuits to `undefined`, `writeText` is never called, no error is thrown — yet `copied = true` still runs and the button flips to "Copied" even though nothing was copied. The `catch` comment ("clipboard denied — the URL stays visible to copy manually") only covers a *rejected* promise, not the missing-API case. Consider guarding, e.g. `if (!navigator.clipboard) return;` before the write, so the label only flips on an actual copy.

## Non-blocking observations (not findings)

- The clear-on-operator-input heuristic assumes that during a blocked auth wait no unrelated `role:"user"` (non-tool_result) record lands between the auth URL and the operator's real response. That holds for the target MCP-auth flow (the agent produces no new records while blocked), so the heuristic is safe in practice; worth keeping in mind if future flows interleave synthetic user messages.

## Could not verify (out of this repo)

- The exact wire shapes of live Notion/Vercel MCP `tool_result` payloads are not present in this repo; the detector is validated against the synthetic real-shape fixtures in `src/auth-url.test.ts`, assumed representative as written.

scoop (commented):
## Re-review

The single prior finding is **addressed**:

> ViewportTermBanners.svelte: copyAuth flips `copied = true` even when `navigator.clipboard` is undefined.

`ui/src/lib/components/viewport/ViewportTermBanners.svelte:819` now guards `if (!authUrl || !navigator.clipboard) return;` before setting `copied = true`, so the label only flips on a real write. The browser test (`ViewportTermBanners.browser.test.ts:772`) stubs `navigator.clipboard` and asserts the flip only then.

## Verification (ground truth cited)

- Imports resolve: `eachJsonlObject` (src/jsonl.ts:3), `readTranscriptTail` (src/activity.ts:158), `jsonlPathFor` (src/usage.ts:13).
- i18n parity holds: `viewport_auth_title/open/copy/copied` + `feat_auth_url_banner_title/body` present in BOTH ui/messages/en.json:645-648,656-657 and ui/messages/de.json:621-624,632-633.
- Feature announcement auto-registers via `import.meta.glob` (ui/src/lib/feature-announcements.ts:47); `titleKey`/`bodyKey` match the catalog keys; `sinceVersion: "1.42.0"` is consistent with the other already-staged v1.42.0 entries (package.json at 1.41.0).
- Poller: `maybeClassify(s, term)` signature change updated at both callers (src/poller.ts:490, :596); `reason` is a fresh object from `this.classify(visible)` (src/poller.ts:957), so mutating `reason.authUrl` is safe; the `detectAuth` read is gated to `shape === "awaiting-input"` (src/poller.ts:982) and rides the `lastSig` dedup + `blockSnapshot`.
- `emitBlock` keeps `lastBlockReason` in step (set on reason, delete on clear/eviction — src/poller.ts:373-377,626), so `blockSnapshot()` only exposes currently-active blocks.
- Store/page wiring consistent: `setBlocks` (store.svelte.ts:197) mirrors the sibling snapshot bootstraps (`setHolds`/`setActivity`); `store.blocks` shape `{reason, since}` matches `setBlock` (store.svelte.ts:329); `selectedAuthUrl` threaded into both Viewport instances (+page.svelte:2534,2696).
- Security: `URL_RE` and `isAuthUrl` only admit `https?://` (src/auth-url.ts:136,155), so `window.open(authUrl, …)` cannot receive a `javascript:`/other scheme; the localhost `/callback?code=…` return URL is correctly excluded (auth-url.test.ts:68).

## Could not verify (stated limitation, not a finding)
- The exact record shapes emitted by live Notion/Vercel MCP OAuth flows are not present in this repo; `detectAuthUrl` is validated only against the synthetic fixtures in src/auth-url.test.ts. Assumed as written.

No blocking issues found.

Issue #1419, opened 2026-07-04
Critic migration: MAX_THINKING_TOKENS → --effort channel switch (follow-up to #1417)


scoop:
Follow-up to #1417. The critic's reasoning budget is currently `env.MAX_THINKING_TOKENS` (value `CRITIC_THINKING_TOKENS = 8000`) injected inside the `--settings` JSON at `transient-agent-argv.ts:147`, used by the two critic spawns (`review.ts`, `standalone-critic.ts`). This is a **thinking-budget** mechanism, not the effort flag.

## Task
Switch the critic's reasoning channel from `env.MAX_THINKING_TOKENS` to the GA `--effort` flag (the seam added in #1417): critic defaults `high`/`xhigh`; lowering the critic effort warns. In #1417 the two critic sites deliberately keep `thinkingTokens` and omit `effort` (byte-unchanged), so this migration is isolated.

## Gate (house rule — demonstrate, don't assert)
The issue-#1417 claim that `MAX_THINKING_TOKENS` "already 400s/no-ops on Opus 4.8" is **unverified in-repo**. This follow-up must capture a concrete before/after on current code: the current env-var channel's behavior on an Opus-4.8 critic spawn vs. the effort channel's restored rigor. Do not merge on parse-green alone.

scoop:
✅ PR #1431 merged.

scoop:
Completed by #1431 (`c1ece203`, merged). Verified against current `main`:

**Channel switch done** — critic reasoning is now `--effort`, not `MAX_THINKING_TOKENS`:
- `config.ts:606` — `criticEffort` defaults `high` (`SHEPHERD_CRITIC_EFFORT` override).
- `index.ts:1045,1074` — both critic sites build env via `roleEnv(…, config.criticEffort)`.
- `review.ts:576` + `standalone-critic.ts:419` — pass `effort: env.effort` to `buildTransientAgentArgv`.
- `transient-agent-argv.ts:165-166` — emits `--effort high`; doc comment at `:78-79` names #1419.

**Old channel fully gone** — grep for `MAX_THINKING_TOKENS`/`CRITIC_THINKING_TOKENS`/`thinkingTokens` across `src/`, `test/`, `ui/src/` → zero hits. The orphaned `thinkingTokens` param + `MAX_THINKING_TOKENS` fold were removed.

[...]

Issue #1417, opened 2026-07-04
Expose reasoning effort (low…max) as a first-class control (re-scope of "thinking mode")


scoop:
## Summary

Expose **reasoning effort** (`low…max`) as a first-class, user-settable control in Shepherd — the live cost/quality lever, dialable both up (hard tasks) and down (routine/cost-sensitive work).

**Re-scope note:** the original framing was "expose thinking mode." That's a trap. Shepherd's existing "thinking" plumbing is `MAX_THINKING_TOKENS` (the thinking-**budget** mechanism), which **returns HTTP 400 on Opus 4.8 / Fable 5 / Opus 4.7 / Sonnet 5** — it's dead on the default models. The live, GA lever is **reasoning effort** (`output_config.effort`), and it's what actually moves cost (Fable-5-at-`low` beats Opus-at-`max` for a fraction of the spend). So this issue delivers an **effort** control, not a thinking-budget one.

Outcome of an attended `/grill-me` design session.

## Feasibility — confirmed (pinned CLIs)

- **Claude Code 2.1.201:** `--effort <low|medium|high|xhigh|max>`, "for the current session" — **no `--print`-only caveat**, so it works on the interactive main-session spawn. Threads exactly like the existing `--model` flag.
- **Codex 0.142.5:** `-c model_reasoning_effort=<minimal|low|medium|high>` on `codex exec`. Note the divergent domain — **no `xhigh`/`max`**.

## Decision record

| # | Decision |
|---|---|
| 1 | Motivation: cost **and** quality — first-class knob, both directions |
| 2 | Dial = **reasoning effort**; "thinking mode" re-scoped to "effort" |
| 3 | Feasibility confirmed (flags above) |
| 4 | Scope: main sessions **+** per-role transient matrix effort column; auto-drain inherits |
| 5 | Tiers: mirror the model picker exactly — session override → repo default (`+ "inherit"`) → global |
| 6 | Default = **unset/inherit** (`"default"` sentinel → don't emit the flag); shipping changes nothing until opt-in |
| 7 | Guardrails: critic defaults `high`/`xhigh`; lowering critic warns; effort orthogonal to the usage governor in v1; no auto-drain floor |
| 8 | Provider/model-awareness: UI hints **+** server-side clamp at argv-build as the guarantee |
| 9 | Manual only for v1; auto-heuristic effort deferred |

## Design

**Config model (mirror `defaultModel`):**
[...]

PR #1437, opened 2026-07-05
feat(settings): warn when critic effort lowered below high

### LINKED ISSUE #1430: Critic guardrail: warn when critic effort is lowered below high (deferred from #1419)
Follow-up to #1419 (which switched the critic's reasoning channel from `MAX_THINKING_TOKENS` to `--effort`, defaulting `criticEffort = high`).

#1417 decision-record item #7 called for a guardrail: **lowering the critic's effort should warn**. #1419's scope was the channel switch itself; the warn was deferred here so #1419 stayed a focused, low-risk migration.

## Task
When the operator lowers `criticEffort` below `high` (via the per-role effort matrix / the `criticEffort` PATCH in `server.ts` `makeRoleEffortPatch`, and/or the Settings UI matrix), surface a warning that the critic is a rigor role and a reduced effort weakens PR review.

## Notes / open scope
- **Where:** `makeRoleEffortPatch("critic")` is currently generic (no per-role warn). Decide between a server-side log warning vs. a reviewer-visible UI warning in the per-role effort matrix (`Settings.svelte`). A UI warning arms the i18n (`check:i18n`) + feature-catalog gates (EN+DE keys, feature entry) per `CLAUDE.md`.
- **Threshold:** warn on any tier strictly below `high` (i.e. `low`/`medium`), leaving `high`/`xhigh`/`max` silent.
- Effort tiers + ordering live in `src/types.ts` (`EFFORTS`) and `src/default-effort.ts`.

This is the remaining piece of #1419's Task line; #1419 itself is complete for the channel switch.

scoop:
Closes #1430. Deferred guardrail from #1417 decision-record item #7 (the `--effort` channel switch itself landed in #1419).

## What

The critic is a rigor role; a reasoning effort that resolves **below `high`** — the tiers `low`/`medium` **and the `default` setting** — weakens PR review. This warns on **both** paths that can lower it:

- **UI inline warning** — an amber `.premium-warn` line in the Settings per-role effort matrix (`Settings.svelte`), shown whenever the critic's effort resolves below high. Reuses the existing warning style; reviewer-visible where the operator acts.
- **Server `console.warn("[critic-effort] …")`** — covers the non-UI paths: a direct/external-task `criticEffort` PATCH via `makeRoleEffortPatch("critic")` (`server.ts`), and the `SHEPHERD_CRITIC_EFFORT` env seed at startup (`config.ts`). Critic-gated, so other roles keep the generic behavior.

`high`/`xhigh`/`max` are silent on every surface.

## Why `default` warns

Selecting `default` emits no `--effort` flag → the CLI's own native default applies, which is *below* `high` — exactly why `config.ts` seeds `criticEffort` to `"high"`. Leaving it silent would let the operator reintroduce the below-high behavior the seed prevents, with no warning.

## Design

- Threshold lives in a shared `effortBelowHigh(setting)` helper — canonical in `src/default-effort.ts` (derived from `EFFORTS` ordering, plus the explicit `default` case), mirrored in `ui/src/lib/effort-guidance.ts` (following the existing `effortsForProvider`/`providerEfforts` mirror pattern).
- Unit-tested on **both** sides (`test/default-effort.test.ts` + `ui/src/lib/effort-guidance.test.ts`), pinning `effortBelowHigh("default") === true`.
- One feature-catalog entry (`v1.42.0-critic-effort-guardrail`) + EN/DE message keys.

## Verification

- Root: `bun run lint` clean, `bun run test` (6033 pass / 0 fail).
- UI: `bun run check` (0 errors), `check:i18n` (2 locales in parity), `bun run test` (2878 pass).
- PR-hygiene gates (branch-hygiene, feature-catalog, announcement-versions, glossary) pass; pre-push green.
- Ran `src/config.ts` with `SHEPHERD_CRITIC_EFFORT` = `low`/`default` → logs the `[critic-effort]` warning; unset/`max` → silent.


scoop (commented):
## Summary

The PR implements the #1430 guardrail: it warns when the critic's reasoning effort is lowered below `high`, across all three lowering surfaces called out in the issue — the Settings UI matrix (visible inline warning), the `criticEffort` PATCH in `server.ts` `makeRoleEffortPatch` (server log), and the `SHEPHERD_CRITIC_EFFORT` env var at startup (`config.ts` log). It arms the i18n + feature-catalog gates with EN+DE keys and a `v1.42.0` announcement fragment. The implementation is internally consistent and I found no bugs.

## Verification performed

- **`effortBelowHigh` logic (both mirrors).** `low`(idx 0)/`medium`(idx 1) < `high`(idx 2) → `true`; `high`/`xhigh`/`max` → `false`; `default` → `true`; unknown → `idx -1` → `false`. Verified against `EFFORTS = ["low","medium","high","xhigh","max"]` in `src/types.ts:228` and the UI mirror `ui/src/lib/types.ts:1611` — the two arrays are identical, so both `effortBelowHigh` copies (`src/default-effort.ts:112`, `ui/src/lib/effort-guidance.ts:23`) are byte-identical in behavior as the comments claim.
- **Server PATCH path.** `makeRoleEffortPatch` normalizes `value` via `normalizeDefaultEffortSetting` (server.ts:4259), so `v` is a non-null valid setting string before `effortBelowHigh(v)` runs (server.ts:4267); `role === "critic"` is a valid `RoleKey` member (server.ts:4232). Warning is critic-scoped, leaving other roles unchanged.
- **Startup path.** `config.criticEffort` is `normalizeDefaultEffortSetting(...) ?? "high"` (config.ts:606), always a valid string; the `if (effortBelowHigh(...))` block (config.ts:728) runs after the `config` literal is defined, so `config.criticEffort` is populated. Because the seed is `"high"`, the warning is silent in normal runs and only fires when the env var was explicitly set below high — matches the issue's intent.
- **UI warning placement.** `ROLE_EFFORT_SEED.critic = "high"` (Settings.svelte:295), so on fresh load `effortBelowHigh(roleEffortV.critic)` is `false` and no warning shows; it appears reactively only when the operator selects `low`/`medium`/`default`. The `premium-warn` class exists (Settings.svelte:2024) and `effortBelowHigh` is imported (Settings.svelte:35).
- **i18n parity.** All three new keys exist in BOTH catalogs — `settings_critic_effort_weakened_warning`, `feat_critic_effort_guardrail_title`, `feat_critic_effort_guardrail_body` — verified via count (en.json:3, de.json:3); `ui/messages/` holds only en+de, so parity is complete. JSON insertions are comma-correct.
- **Feature catalog.** New fragment `ui/src/lib/feature-announcements/entries/v1.42.0-critic-effort-guardrail.ts` supplies the required `id`/`sinceVersion`/`titleKey`/`bodyKey` (`satisfies FeatureAnnouncement`, type at feature-announcements.ts:25). Entries are auto-globbed (feature-announcements.ts:47), so no manual registration is needed. `sinceVersion: "1.42.0"` is the next unreleased version (package.json:3 = 1.41.0) and matches the other pending `v1.42.0-*` fragments. The `check-feature-catalog` gate is satisfied because the same PR modifies `ui/src/lib/components/Settings.svelte` and adds the fragment.

## Design note (non-blocking)

The issue's stated threshold is "warn on any tier strictly below `high` (i.e. `low`/`medium`)" and does not list `default`. The implementation additionally treats `default` as below-high. This is a deliberate, well-documented extension (the critic seeds to `high`, and `default` emits no `--effort` flag → the CLI's native below-high default), and since the critic is never seeded to `default` it only surfaces when an operator explicitly picks it. I consider this correct rather than a defect, but flagging it so the divergence from the issue's literal wording is a conscious choice.

No findings to address.

PR #1438, opened 2026-07-05
fix(poller): resolve transcripts under the session's spawn account

scoop:
## Problem

The MCP OAuth **auth-URL banner** (#1436) never appeared, even for the mid-conversation flow it targets. Root cause is upstream of that feature: shepherd resolved **every** session's JSONL under the server's own `config.claudeProjectsDir` (`~/.claude/projects`), ignoring the swap/pool account the agent actually ran under.

claude-swap runs agents with `CLAUDE_CONFIG_DIR=<account>`, so their transcript lands in `<account>/projects/…`. `jsonlPathFor` read a nonexistent path → the read was swallowed → no data. This silently degraded **all** transcript readback (usage tracking, activity signals, halt classification, recap, and the auth-URL banner) for every claude-swap (`sandboxApplied=trusted`, unconfined — no membrane bind) session. The banner was just the newest feature to surface the miss.

Confirmed general, not feature-specific — every swap-account session, including live ones:

```
TASK-1344  blocked   sbox=trusted   active_path=False   swap_path=True
TASK-1336  done      sbox=trusted   active_path=False   swap_path=True   ← reported
```

## Fix

Make transcript resolution **account-aware**:

- `jsonlPathFor` takes an optional `spawnAccountDir` (new module-local `projectsDirFor`): roots under `<account>/projects` when set, else the active `config.claudeProjectsDir`.
- Thread `s.spawnAccountDir` through every session-level call site: poller (auth-url, activity/stall, halt), stalled-session scan, process-reaper, usage snapshot/breakdown/timeline rollup, and recap's main-session transcript read.
- Reviewer/critic reads intentionally stay on the active dir — they're shepherd-spawned, never swapped.

## Verification

A/B under the server's real env (no `CLAUDE_CONFIG_DIR`) against the real reported transcript:

| | Resolved path | Exists | Banner `authUrl` |
|---|---|---|---|
| **Pre-fix** | `~/.claude/projects/…/0189bb52.jsonl` | ❌ | `null` (bug reproduced) |
| **Post-fix** | `…/claude-swap/…/projects/…/0189bb52.jsonl` | ✅ | `https://mcp.sentry.dev/oauth/authorize?…` ✅ |

- New unit tests lock account-aware resolution + null fallback (RED→GREEN)
- Full root suite: **6031 pass, 0 fail** · `tsc` clean · lint clean · fallow audit clean

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1749,7 +1749,7 @@ const herdDigestService = new HerdDigestService({
     const stalled = new Set<string>();
     for (const s of store.list({ activeOnly: true })) {
       if (s.status !== "running" || !s.claudeSessionId) continue;
-      const snap = readSnapshot(jsonlPathFor(s.worktreePath, s.claudeSessionId));
+      const snap = readSnapshot(jsonlPathFor(s.worktreePath, s.claudeSessionId, s.spawnAccountDir));
       if (snap && isStalled(snap, now, DEFAULT_STALL)) stalled.add(s.id);
     }
     return stalled;
--- a/src/poller.ts
+++ b/src/poller.ts
@@ -93,7 +93,9 @@ export interface LivenessWiring {
 function readAuthUrl(s: Session): string | null {
   if (!s.claudeSessionId) return null;
   try {
-    return detectAuthUrl(readTranscriptTail(jsonlPathFor(s.worktreePath, s.claudeSessionId)));
+    return detectAuthUrl(
+      readTranscriptTail(jsonlPathFor(s.worktreePath, s.claudeSessionId, s.spawnAccountDir)),
+    );
   } catch {
     return null;
   }
@@ -219,7 +221,7 @@ export class StatusPoller {
      */
     private probe: (s: Session) => TranscriptSignals = (s) =>
       s.claudeSessionId
-        ? readTranscriptSignals(jsonlPathFor(s.worktreePath, s.claudeSessionId))
+        ? readTranscriptSignals(jsonlPathFor(s.worktreePath, s.claudeSessionId, s.spawnAccountDir))
         : { snapshot: null, activity: null },
     private stallCfg = DEFAULT_STALL,
     private probeCheckMs = 7000,
@@ -552,7 +554,9 @@ export class StatusPoller {
   private detectUsageHalt(s: Session): void {
     if (s.haltReason) return; // already set; skip
     try {
-      const tail = readTranscriptTail(jsonlPathFor(s.worktreePath, s.claudeSessionId));
+      const tail = readTranscriptTail(
+        jsonlPathFor(s.worktreePath, s.claudeSessionId, s.spawnAccountDir),
+      );
       // Match only NON-user transcript content: a user prompt mentioning usage-limit
       // phrasing must never be read as Claude's own halt notice (false positive on the
       // uncalibrated degrade path, where the usage corroboration is bypassed).
--- a/src/process-reaper.ts
+++ b/src/process-reaper.ts
@@ -264,7 +264,12 @@ export class ProcessReaper {
   constructor(private probes: ReaperProbes = defaultProbes) {}
 
   /** Detect surviving leftovers for a session (classes 2 + 3). */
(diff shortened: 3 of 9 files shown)


scoop (commented):
## Summary

This PR threads a new optional `spawnAccountDir` through `jsonlPathFor` so a session's JSONL transcript is resolved under the swap/pool account it actually ran on, instead of always the server's active projects dir. This directly addresses the reported symptom: `readAuthUrl` (poller.ts:93-99) reads the transcript and runs `detectAuthUrl` over it; for a swap-account session the old path pointed at a nonexistent file, so `detectAuth` always returned `null` and the MCP-OAuth banner never appeared. With the fix, `authUrl` is populated into `reason.authUrl` (src/poller.ts:990-991) and surfaces to the UI.

The implementation is internally consistent and correct — I verified each piece:

- **New resolution logic** — `projectsDirFor` returns `join(spawnAccountDir, "projects")` when set, else `config.claudeProjectsDir` (src/usage.ts:18-19). This parallels the active-dir construction `<CLAUDE_CONFIG_DIR>/projects` (src/config.ts:437-439), so the swap-account form is the right mirror. Verified.
- **Field exists on Session** — `spawnAccountDir: string | null` (src/types.ts:111), populated from `patchEnv.CLAUDE_CONFIG_DIR` at spawn (src/service.ts:1801). Every in-diff caller sources it from a real `Session`/`store.list()` row:
  - src/index.ts:1752 — `s` from `store.list({activeOnly:true})` (Session). ✓
  - src/poller.ts:97, 224, 558 — `s: Session`. ✓
  - src/process-reaper.ts:311 — caller `leftovers()` passes `store.get(id)` (Session) into `detect` (src/service.ts:3919-3922); the added `spawnAccountDir?` field is optional so the caller still type-checks and the real value flows through. ✓
  - src/recap.ts:451 — destructured from `session` at src/recap.ts:405. ✓
  - src/usage-breakdown.ts:111 (`liveSessionToAccum` param is `SessionStore["list"][number]`) and :348, src/usage-snapshot.ts:52 (`s: Session`), src/usage-timeline.ts:37 — all sourced from `eligible`/`Session`. ✓
- **RollupSession** — new optional `spawnAccountDir?` field (src/usage.ts:455-458) is populated by both live-rollup builders (src/usage-breakdown.ts:348, src/usage-timeline.ts:37) and consumed in `ingestSession` via `pathFor` (src/usage.ts:559, 528). ✓
- **Tests** — src/test/usage.test.ts:58-72 assert `"${acct}/projects/-home-p-Work-r/abc-123.jsonl"` (dashify of `/home/p/Work/r` = `-home-p-Work-r`, verified against src/usage.ts:8-9) and the null-fallback equals the 2-arg form. Both match the implementation. ✓
- **Locale/i18n** — no user-facing strings or message keys touched; nothing to check.

No blocking issues found. The change satisfies the task.

- src/usage.ts:261-270 — `readSessionUsage` gains a `spawnAccountDir?` param, but no caller in this diff passes it (all its callers live in out-of-diff files and still call it 2-arg). The param is harmless (defaults to the active dir, prior behavior) but its swap-account benefit is unrealized until a follow-up wires callers; usage totals for swap-account sessions read via `readSessionUsage` still miss.

- src/server.ts:500 — `sessionUsage(s)` calls `jsonlPathFor(s.worktreePath, s.claudeSessionId)` (2-arg) though `s` is a full `Session`; the session-tokens display still resolves against the active dir and will miss for swap-account sessions.
- src/server.ts:1902 — `sessionActivityRead` calls `jsonlPathFor(s.worktreePath, s.claudeSessionId)` (2-arg) with a full `Session` in scope; the activity endpoint has the same swap-account gap.
- src/review.ts:1161 — `defaultReadActivity` uses the 2-arg `jsonlPathFor` for the critic transcript; fine only while critics never run under a pool account.
- src/usage.ts:302-312 — `AccountUsageIndex.refresh` globs only `this.projectsDir` (active dir), so the account-wide usage index does not discover swap-account session files. Separate class, untouched by this PR.

PR #1441, opened 2026-07-05
fix(poller): surface MCP OAuth auth-URL banner for done/idle sessions

scoop:
## Problem

The clickable MCP OAuth auth-URL banner (#1436) never appeared for the reported case (screenshot: a `check-sentry` session relaying a Sentry `…/authorize?…` URL), even after #1438 fixed the swap-account transcript path.

**Root cause (verified end-to-end against the real session):** an MCP OAuth prompt *ends the agent's turn*, so herdr reports the session `done` (never `blocked`). But auth-URL detection (`detectAuth`) runs **only** on the `blocked`-classification path (`maybeClassify`). A `done`/`idle` session routes to `maybeQuota`, which never runs detection, so `authUrl` was never attached and no block/banner emitted. #1438 (transcript path) was necessary but not sufficient — detection was gated behind a path these turns never hit.

## Fix

Detect the pending auth URL for **resting** (done/idle) sessions too and route it through the existing block plumbing as an `awaiting-input` block carrying `authUrl` — surfacing both the banner and the attention-lens row. **No UI changes** (the banner already renders purely on `authUrl`).

- **`maybeAuthAtRest`** (poller): mtime-gated — one bounded, swap-account-aware transcript read per change, not per tick. Self-clears at rest when the operator responds (transcript append bumps mtime → re-probe → `detectPendingAuthUrl` returns null → `clearBlock`); a first-tick flush-race miss does **not** latch.
- **`detectPendingAuthUrl`**: a generous tail-position freshness gate that suppresses a stale URL buried past the 512 KB tail window while surfacing genuine prompts. Biased toward surfacing — a false-suppress *is* the reported bug (unrecoverable); a false-surface is bounded (autopilot clears on resume/archive).
- **Autopilot stands down** on a pending auth URL (an OAuth flow is human-only — it cannot paste the callback) across **every** steer path: `onBlock` (before the STEERABLE guard so a null block clears), `onDone` pre-check, a **post-classify re-check** (catches a URL unflushed at the done edge → never `markComplete`s a mid-OAuth session), `eligible()`, and inside `reEngageCi`/`rebaseCandidate` (which bypass `eligible()` and can fire on a session that already has an open PR). Recovers on `onStatus(running)` and `forget` (archive).

## Testing

- `detectPendingAuthUrl` returns the URL for the **real captured `check-sentry` tail** (regression fixture) + synthetic freshness cases (last-record ✓, `>K` records after ✗, operator-clear ✗).
- Poller (seam-driven, no disk via injectable `authMtime`/`detectRestingAuth`): resting emit, flush-race re-probe, clear-at-rest on mtime bump, no re-read when unchanged.
- Autopilot: no steer/classify on an authUrl block; `onDone` stands down before classify; post-classify re-check blocks a terminal verdict; `reEngageCi` respects the stand-down and `onStatus(running)` clears it.
- `bunx tsc --noEmit`, `bun run lint`, `bun test ./test` (6043 pass, 0 fail) all green. No UI changed.

### DIFF
--- a/src/auth-url.test.ts
+++ b/src/auth-url.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "bun:test";
-import { detectAuthUrl } from "./auth-url";
+import { detectAuthUrl, detectPendingAuthUrl } from "./auth-url";
 
 // Real-shape fixtures (synthetic client_id / code_challenge). Mirrors the Step-0 capture:
 // the authorize URL lands in an assistant `text` block AND/OR a `tool_result` block inside a
@@ -100,5 +100,57 @@ describe("detectAuthUrl", () => {
 
   it("returns null for an empty transcript", () => {
     expect(detectAuthUrl("")).toBeNull();
+  });
+});
+
+// Attachment / system records — carry no `message`, so they are NOT meaningful records and must
+// not count toward the freshness gate's since-set distance.
+const attachment = () => JSON.stringify({ type: "attachment" });
+const system = () => JSON.stringify({ type: "system", content: "hook ran" });
+
+describe("detectPendingAuthUrl (freshness-gated)", () => {
+  it("returns the URL for the real captured check-sentry tail shape (0 meaningful records after)", () => {
+    // Mirrors the captured transcript tail: the authorize URL lands in a `tool_result` (raw MCP
+    // payload) AND the assistant relay, with only attachment/system records trailing — sinceSet=0.
+    const jsonl = [
+      opInputString("check sentry"),
+      toolUse(),
+      toolResult(`MCP server sentry requires authorization. Visit: ${AUTH_URL}`),
+      attachment(),
+      assistantText(
+        `The Sentry MCP server needs authorization.\n1. Open this URL:\n\n${AUTH_URL}\n\n2. Paste the callback URL here.`,
+      ),
+      attachment(),
+      system(),
+      system(),
+    ].join("\n");
+    expect(detectPendingAuthUrl(jsonl)).toBe(AUTH_URL);
+  });
+
+  it("surfaces a URL that is the last meaningful record", () => {
+    const jsonl = [opInputString("go"), assistantText(`Authorize: ${AUTH_URL}`)].join("\n");
+    expect(detectPendingAuthUrl(jsonl)).toBe(AUTH_URL);
+  });
+
+  it("suppresses a stale URL buried under more than K meaningful records (truncation phantom)", () => {
+    // A stale URL whose clearing operator message scrolled out of the MAX_TAIL_BYTES window,
+    // followed by many autonomous tool turns — must NOT re-surface as a phantom banner.
+    const trailing = Array.from({ length: 30 }, () => toolUse());
+    const jsonl = [assistantText(`old prompt: ${AUTH_URL}`), ...trailing].join("\n");
+    expect(detectPendingAuthUrl(jsonl)).toBeNull();
+  });
+
+  it("still surfaces a URL with a handful of trailing assistant messages (generous K)", () => {
+    const trailing = Array.from({ length: 5 }, () => assistantText("still waiting on you"));
+    const jsonl = [assistantText(`Authorize: ${AUTH_URL}`), ...trailing].join("\n");
+    expect(detectPendingAuthUrl(jsonl)).toBe(AUTH_URL);
+  });
+
+  it("returns null once the operator has responded (explicit clear)", () => {
+    const jsonl = [
+      assistantText(`Authorize: ${AUTH_URL}`),
+      opInputString("http://localhost:3118/callback?code=xyz"),
+    ].join("\n");
+    expect(detectPendingAuthUrl(jsonl)).toBeNull();
   });
 });
--- a/src/auth-url.ts
+++ b/src/auth-url.ts
@@ -123,19 +123,65 @@ function userAuthEffect(content: unknown): { clear: boolean; url: string | null 
  * So a prior, already-answered auth URL still sitting in the tail is not resurfaced when the
  * agent later blocks for an unrelated reason. Malformed/blank lines are skipped.
  */
+/** The pending-URL transition a single record implies: SET (assistant text / tool_result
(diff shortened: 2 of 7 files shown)


scoop (commented):
## Summary

This PR closes the gap left by #1438: an MCP OAuth prompt ends the agent's turn, so herdr reports `done`/`idle` (never `blocked`) and the pre-existing `maybeClassify` auth path never fires. The PR adds a resting-session detection path (`maybeAuthAtRest`), a freshness-gated URL detector (`detectPendingAuthUrl`), and an autopilot stand-down (`authPending`) so a human-only OAuth flow is never steered or filed complete.

I reviewed the wiring, the state machine, and the freshness gate against the actual code. I found no blocking defects. Verifications below.

## Verifications

**Transcript path resolution follows the spawn account (the #1438 fix) everywhere the new code reads a transcript.** `readRestingAuthMtime` (src/poller.ts:112), `readRestingAuth` (src/poller.ts:123-124) and the autopilot `pendingAuthUrl` dep (src/index.ts:1419) all pass `s.spawnAccountDir` into `jsonlPathFor`, which routes through `projectsDirFor` (src/usage.ts:18-20). So a claude-swap session's JSONL is read from `<account>/projects/…`, not the server's active dir — consistent with the intent documented at src/usage.ts:12-17.

**Imports resolve and are non-duplicate.** `detectPendingAuthUrl` (src/index.ts:150, src/poller.ts:328 export at src/auth-url.ts:171), `readTranscriptTail` (src/index.ts:151, defined src/activity.ts:158), `jsonlPathFor` imported once (src/index.ts:149), `statSync` (src/poller.ts:329), `assistantSideText` (src/usage-halt.ts:49) and `tailLines` (src/poller.ts:14, from src/blocked.ts:37) all exist and are used with correct signatures.

**`BlockReason` supports the emitted shape.** `authUrl?: string` and `shape: "awaiting-input"` are both valid (src/blocked.ts:22, src/blocked.ts:109). The emit at src/poller.ts:1146 (`{shape:"awaiting-input", options:[], tail, authUrl}`) is well-formed and rides the `lastSig`/`blockSnapshot` dedup via `emitBlock` (src/poller.ts:943-947).

**The freshness gate counts the right records.** `detectPendingAuthUrl` (src/auth-url.ts:171-187) increments `sinceSet` only on `toRec`-non-null records that are neither SET nor CLEAR (src/auth-url.ts:184). `toRec` (src/auth-url.ts:77-83) returns null for `attachment`/`system` records (no `message.content`), so they don't count — confirmed against the test fixtures at src/auth-url.test.ts:19-20 and the captured-tail test at src/auth-url.test.ts:23-39 (SET by tool_result + assistant relay ⇒ sinceSet=0 ⇒ surfaces). `toolUse()` (src/auth-url.test.ts:24-28) is an assistant record with a non-text block, so `assistantAuthUrl` yields no URL (src/auth-url.ts:88-94) ⇒ counts as a `none` record ⇒ the 30-`toolUse` test trips `sinceSet > 25` and suppresses (src/auth-url.test.ts:46-52). The default `maxSinceSet=25` is generous and biased to surfacing, as documented at src/auth-url.ts:164-169.

**The mtime-gated read is correct and does not latch a first-tick miss.** `maybeAuthAtRest` (src/poller.ts:1136-1149): unchanged mtime returns the standing state without a read; a changed mtime with no URL returns false so `maybeQuota` (src/poller.ts:1115) clears; a first-tick miss (URL not flushed) sets `lastAuthMtime` but does NOT set `AUTH_SIG`, so the next append re-probes — verified against src/poller.test.ts:627-641. `AUTH_SIG="auth"` (src/poller.ts:337) never collides with a `JSON.stringify(reason)` sig or with `QUOTA_SIG`/`STALL_SIG`. On operator resume, the running-path `clearStaleBlock` clears any non-stall sig incl. AUTH_SIG (src/poller.ts:907-908). `lastAuthMtime` is added to both the tracked set and the prune path (src/poller.ts:653, src/poller.ts:672) so it doesn't leak.

**Autopilot stand-down set/clear is complete.** `authPending` is set on `onBlock(authUrl)` (src/autopilot.ts:400-402), `onDone` pre-classify (src/autopilot.ts:416-418), and `consider` post-classify (src/autopilot.ts:373-375). It is cleared on a null/non-auth block (src/autopilot.ts:404 — placed before the shape guard so a null actually reaches `delete`), on `running` (src/autopilot.ts:456, before the paused/complete guard), and on archive via `forget` (src/autopilot.ts:245-250, wired at src/index.ts:1525-1528). It is honored on every steer path: `eligible` (src/autopilot.ts:231), `reEngageCi` candidate (src/autopilot.ts:571) and `rebaseCandidate` (src/autopilot.ts:662). The `pendingAuthUrl` dep is required in `AutopilotDeps` (src/autopilot.ts:158) and supplied by all three `new AutopilotService` call sites (src/index.ts:1411, test/autopilot.test.ts:176, test/autopilot.test.ts:615). Behavior matches the added tests at test/autopilot.test.ts:1455-1558.

## Non-blocking observations

- The autopilot `pendingAuthUrl` dep (src/index.ts:1414-1421) performs an unguarded 512 KB tail read on every `onDone` edge and every `consider` post-classify re-check, whereas the poller took care to mtime-gate its equivalent read. These fire far less often than the per-tick poller path (once per done/classify, not per tick) and sit next to a multi-second LLM spawn, so the cost is negligible — noting only the asymmetry.
- `reapGone` (src/poller.ts:474-486) clears the block and does not run the resting-auth probe. This is correct for its case (an exited agent cannot be waiting on an OAuth callback); the OAuth-at-rest case keeps the agent alive, so herdr lists it and it routes through `reconcileAgent → maybeQuota → maybeAuthAtRest`.

## Stated limitations (could not verify in this repo)

- Whether herdr actually reports an MCP-OAuth-waiting claude-swap session as `done`/`idle` (the premise of this whole path) and whether such a session is matched by `matchAgents` rather than reaped — herdr's runtime behavior is external and not present in this repo. The reported "still cannot get it to work" could originate in agent matching (out of scope for this diff) rather than transcript detection; if so, this PR's detector would be correct but never reached. Assumed as designed.
- The exact record shapes real Claude Code writes for `system`/`attachment` lines were checked only against the PR's own synthetic fixtures, not a live transcript in this repo.

PR #1457, opened 2026-07-06
fix(poller): surface the Claude /login account-URL banner (PTY-only source)

scoop:
## Problem

The Claude Code **`/login`** (account re-login) flow prints an OAuth authorize URL that Shepherd never surfaces in the auth banner — even though **MCP OAuth works** (fixed in #1441). Reported: "Claude login URL still not recognized and displayed in bar."

## Root cause (confirmed via live repro)

A session sitting at `/login` reports herdr `agent_status: "done"` → it routes through `maybeQuota` → `maybeAuthAtRest`, which reads the **JSONL transcript only**. But `/login` is a slash-command TUI modal: its authorize URL is emitted **only to the PTY** (hard-wrapped at terminal width, no inserted spaces), and **never** lands in the conversation transcript. So both transcript detectors (`maybeClassify`'s `detectAuth` and `maybeAuthAtRest`'s `detectRestingAuth`) can't see it. MCP OAuth works because those URLs *do* appear in the transcript.

The banner UI and autopilot's `onBlock`→`authPending` stand-down already exist and are source-agnostic — **the gap is purely detection.**

## Change

- **`detectLoginAuthUrl(visible)`** (`auth-url.ts`): reconstruct the word-wrapped URL from a terminal visible buffer. Anchors on the first `https` line, strips ANSI + box-drawing borders per line, and joins whole whitespace-free continuation lines — **evaluating STOP before APPEND** so a trailing panel line (`Paste code here…`, even with no blank line) can't graft onto an otherwise-valid authorize URL. Validated by the existing `isAuthUrl`.
- **Resting path** (`maybeAuthAtRest`): a second, PTY source independent of the mtime-gated transcript source — an **async `readAsync`** (the poll-loop convention; a sync read under the resting-session fan-out would freeze the web terminal), throttled to `reclassifyMs`, only when the transcript has no URL. A reconstruction is emitted **only after two consecutive equal reads** (a half-painted, `isAuthUrl`-valid partial must not latch past the value-blind `AUTH_SIG`), the emit is **value-aware** (re-emits on a correction), and a **null read clears immediately** — the real completion path, since a finished `/login` returns to idle/done without a running/blocked edge. Detection caches clear on the leave-resting edge (kept out of `clearBlock`, whose per-tick resting calls would defeat the throttle/stability gate).
- **Blocked path** (`maybeClassify`): the same fallback (visible buffer already in hand), for a login modal herdr happens to classify as `blocked`.
- **Autopilot** (`index.ts`): the deterministic `pendingAuthUrl` stand-down seam gains a fresh PTY fallback so `onDone`/`consider` stand down on `/login` without racing the throttled banner emit — restructured so the `claudeSessionId` guard gates **only** the transcript branch. Factored a shared `readVisibleBuffer` helper (also used by `readTail`).

No UI changes: the banner + copy ("Authorize in your browser to continue") and autopilot stand-down are reused unchanged.

## Tests

- `auth-url.test.ts`: byte-faithful capture of a real `herdr.read(visible)` `/login` panel → full URL; box-bordered wrap; leading prefix; **STOP-before-APPEND** (Paste line with no blank line, plain + bordered); unwrapped; non-auth URL; scrollback.
- `poller.test.ts`: PTY URL surfaces after two equal reads; **partial never emitted**; **null read clears with no running/blocked transition**; PTY skipped while a transcript URL stands; throttle respected; leave-resting cache clear; blocked-path fallback.

## Verification

`bun run typecheck`, `bun run lint`, `bun run test` (6051 pass / 0 fail), and `fallow audit` (no issues in changed files) all green locally.

### DIFF
--- a/src/auth-url.test.ts
+++ b/src/auth-url.test.ts
@@ -1,5 +1,5 @@
 import { describe, it, expect } from "bun:test";
-import { detectAuthUrl, detectPendingAuthUrl } from "./auth-url";
+import { detectAuthUrl, detectPendingAuthUrl, detectLoginAuthUrl } from "./auth-url";
 
 // Real-shape fixtures (synthetic client_id / code_challenge). Mirrors the Step-0 capture:
 // the authorize URL lands in an assistant `text` block AND/OR a `tool_result` block inside a
@@ -152,5 +152,99 @@ describe("detectPendingAuthUrl (freshness-gated)", () => {
       opInputString("http://localhost:3118/callback?code=xyz"),
     ].join("\n");
     expect(detectPendingAuthUrl(jsonl)).toBeNull();
+  });
+});
+
+// Byte-faithful capture of a real `herdr.read(term,"visible")` of the Claude Code `/login`
+// (account re-login) panel — the pane hard-wrapped the flush-left URL at 63 cols, breaking it
+// mid-token with NO inserted spaces. This authorize URL is PTY-only (it never lands in the JSONL
+// transcript), so this fixture is the sole validation that the wrap-join reconstructs a REAL
+// hard-wrap. `client_id` is the public Claude Code OAuth client id; the single-use PKCE
+// `code_challenge`/`state` are long-expired nonces.
+const LOGIN_URL_LINES = [
+  "https://claude.com/cai/oauth/authorize?code=true&client_id=9d1c25",
+  "0a-e61b-44d9-88ed-5944d1962f5e&response_type=code&redirect_uri=ht",
+  "tps%3A%2F%2Fplatform.claude.com%2Foauth%2Fcode%2Fcallback&scope=o",
+  "rg%3Acreate_api_key+user%3Aprofile+user%3Ainference+user%3Asessio",
+  "ns%3Aclaude_code+user%3Amcp_servers+user%3Afile_upload&code_chall",
+  "enge=7fZ4v-QWXM2nnE_04d3um7K5-KOOVAHtRcFmvlTBIfE&code_challenge_m",
+  "ethod=S256&state=t6_LEFskFrRPiy_pnzr_f9uKwJSArorv7PaSt4DuoKY",
+];
+const LOGIN_URL = LOGIN_URL_LINES.join("");
+const LOGIN_PANEL = [
+  "❯ /login",
+  "",
+  "─".repeat(62),
+  "  Login",
+  "",
+  "  Browser didn't open? Use the url below to sign in (c to copy)",
+  "",
+  ...LOGIN_URL_LINES,
+  "",
+  "",
+  "  Paste code here if prompted >",
+  "",
+  "  Esc to cancel",
+].join("\n");
+
+describe("detectLoginAuthUrl", () => {
+  it("reconstructs the wrapped URL from a byte-faithful /login panel capture", () => {
+    expect(detectLoginAuthUrl(LOGIN_PANEL)).toBe(LOGIN_URL);
+    // Sanity: the reconstruction really is a valid authorize URL.
+    expect(new URL(LOGIN_URL).pathname).toBe("/cai/oauth/authorize");
+  });
+
+  it("returns a single-line (unwrapped) authorize URL as-is", () => {
+    expect(detectLoginAuthUrl(`  Login\n\n${LOGIN_URL}\n\n  Esc to cancel`)).toBe(LOGIN_URL);
+  });
+
+  it("reconstructs a URL wrapped inside a box-drawing border", () => {
+    const bordered = [
+      "│  Login" + " ".repeat(10) + "│",
+      "│" + " ".repeat(20) + "│",
+      ...LOGIN_URL_LINES.map((l) => "│ " + l + " │"),
+      "│" + " ".repeat(20) + "│",
+    ].join("\n");
+    expect(detectLoginAuthUrl(bordered)).toBe(LOGIN_URL);
+  });
+
+  it("stops before appending a following text line with NO intervening blank line (plain)", () => {
+    // `Paste code here…` directly follows the last URL chunk with no blank line. Its leading
+    // `Paste` must NOT be appended (evaluate-STOP-before-APPEND) — else a broken-but-structurally
+    // -valid authorize URL passes both isAuthUrl and the stability gate.
+    const noBlank = [...LOGIN_URL_LINES, "  Paste code here if prompted >"].join("\n");
+    expect(detectLoginAuthUrl(noBlank)).toBe(LOGIN_URL);
+  });
+
+  it("stops before appending a following text line with NO blank line (box-bordered)", () => {
+    const noBlank = [
+      ...LOGIN_URL_LINES.map((l) => "│ " + l + " │"),
+      "│ Paste code here if prompted > │",
+    ].join("\n");
+    expect(detectLoginAuthUrl(noBlank)).toBe(LOGIN_URL);
+  });
+
+  it("anchors past a non-URL prefix on the URL's first line", () => {
+    const prefixed = ["url> " + LOGIN_URL_LINES[0], ...LOGIN_URL_LINES.slice(1)].join("\n");
+    expect(detectLoginAuthUrl(prefixed)).toBe(LOGIN_URL);
+  });
+
+  it("returns null when the visible buffer has no URL", () => {
+    expect(
+      detectLoginAuthUrl("  Login\n\n  Paste code here if prompted >\n\n  Esc to cancel"),
+    ).toBeNull();
+  });
+
+  it("returns null for a wrapped NON-auth URL (isAuthUrl gate)", () => {
+    const nonAuth = [
+      "https://example.com/very/long/path/that/wraps/here/aaaaaaaaa",
+      "bbbbbbbbbbbb/cccccc/dddddd",
+    ].join("\n");
+    expect(detectLoginAuthUrl(nonAuth)).toBeNull();
+  });
+
+  it("skips a leading non-auth URL and finds the auth URL later in the buffer", () => {
+    const buf = ["visit https://example.com/docs for help", "", ...LOGIN_URL_LINES].join("\n");
+    expect(detectLoginAuthUrl(buf)).toBe(LOGIN_URL);
   });
 });
--- a/src/auth-url.ts
+++ b/src/auth-url.ts
@@ -185,3 +185,52 @@ export function detectPendingAuthUrl(rawJsonl: string, maxSinceSet = 25): string
   }
   return pending && sinceSet > maxSinceSet ? null : pending;
 }
+
+// eslint-disable-next-line no-control-regex
+const ANSI_LINE_RE = /\x1b\[[0-9;]*[A-Za-z]/g;
+// Framing a TUI panel puts on a line: box-drawing glyphs (U+2500–U+257F: the `─` rule and the
+// `│`/`┃` verticals) or an ASCII `|`, plus surrounding whitespace/indent. These survive
+// ANSI-strip, so peel them off BOTH ends of a line before anchoring/joining a URL. A URL has
+// none of these characters, so stripping ends never eats URL content.
+const BORDER_ENDS_RE = /^[\s─-╿|]+|[\s─-╿|]+$/g;
+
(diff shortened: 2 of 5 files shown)


scoop (commented):
## Summary

The PR adds `detectLoginAuthUrl` — a reconstructor for the word-wrapped, PTY-only `/login` authorize URL — and wires it into three seams: the poller blocked path (`maybeClassify`), a new resting-session PTY source with a two-read stability gate (`maybeAuthAtRest`/`restingPtyAuth`/`probePtyAuth`), and the autopilot `pendingAuthUrl` fallback. This directly addresses the task (the `/login` URL never landing in the transcript and so never reaching the bar). I found no blocking correctness, security, or i18n problems; the MCP-at-rest transcript path is preserved (transcript takes precedence, PTY consulted only when it yields no URL).

### Verification performed
- `detectLoginAuthUrl` is exported (src/auth-url.ts:219) and imported by both consumers (src/index.ts:150, src/poller.ts:21).
- `BORDER_ENDS_RE` character-class range is `U+2500`–`U+257F` (hexdump of src/auth-url.ts:195 → `e2 94 80`..`2d`..`e2 95 bf`), a valid ascending range covering the box-drawing glyphs; the regex compiles and never eats URL bytes (URLs contain none of `\s`, box-drawing, or `|`). Join logic verified against all fixture cases in src/auth-url.test.ts (wrap-join, border strip, STOP-before-APPEND, non-auth reject, skip-leading-non-auth).
- `s.herdrAgentId` passed to `restingPtyAuth(id, s.herdrAgentId)` (src/poller.ts:1175) is the herdr terminal_id (src/types.ts:26), consistent with the existing `maybeClassify(s, s.herdrAgentId)` read (src/poller.ts:553).
- All five new caches are pruned: added to both the key-union (src/poller.ts:673-677) and the delete block (src/poller.ts:695-699); the detection subset is additionally dropped on the leave-resting edge (src/poller.ts:1267-1273) and `lastAuthUrlEmitted` follows the block via `clearBlock` (src/poller.ts:1122).
- Clear-on-completion without a running/blocked edge is handled by `probePtyAuth`'s immediate null-read clear (src/poller.ts:1242-1243) → `maybeAuthAtRest` returns false → `maybeQuota` clears (src/poller.ts:1149-1150). Resume-clear of a standing AUTH block is still handled by `applyOutcome`'s `clearStaleBlock` (src/poller.ts:934-935).
- Autopilot `pendingAuthUrl` gates only the transcript branch on `claudeSessionId` and always reaches the PTY fallback (src/index.ts:1430-1442); `readVisibleBuffer` reuses `matchAgent` (imported src/index.ts:27).

## Notes (non-blocking)

1. **Blocked-path reconstruction lacks the stability gate the resting path has.** `maybeClassify` does `detectAuth(s) ?? detectLoginAuthUrl(visible)` (src/poller.ts:1064) with no two-read confirmation, whereas the resting path deliberately confirms a non-null reconstruction across two equal reads before latching (src/poller.ts:1225-1245) precisely to avoid surfacing a still-painting, `isAuthUrl`-valid-but-truncated URL. If herdr classifies the `/login` modal as `blocked` (the comment at src/poller.ts:1060-1063 anticipates this) and the throttled classify read lands mid-paint, a truncated authorize URL can surface in the bar for one cadence before the `JSON.stringify(reason)` sig re-emits the corrected value. Self-correcting and low-probability, but it contradicts the PR's own "no silently broken banner" goal.

2. **Every resting session with no transcript auth URL now incurs a periodic async PTY read.** `maybeAuthAtRest` falls through to `restingPtyAuth` whenever the transcript source has no URL (src/poller.ts:1174-1176), and `restingPtyAuth` fires a throttled `readAsync` every `reclassifyMs` (src/poller.ts:1211-1216). For an operator with many idle/done sessions this is N extra herdr round-trips per cadence where previously an unchanged resting session cost only one `stat`. The read is async (non-blocking) and the choice is documented, so this is a cost observation, not a bug — but a cheap pre-guard (e.g. skip the probe when the just-considered buffer has no `https`) would bound it.

- None observed.

scoop:
Both critic points addressed in the latest push (c52d5a8):

**Point 1 (blocked-path stability):** Fixed. Extracted a shared `confirmLoginUrl` two-read stability gate and applied it to **both** the resting (async) and blocked (sync) paths, so neither can surface a mid-paint truncated authorize URL. `probePtyAuth` now routes through it; `maybeClassify` uses `this.confirmLoginUrl(id, { url: detectLoginAuthUrl(visible), tail: [] })?.url`. Added a blocked-path test asserting the URL only attaches after two reads and a partial never surfaces.

**Point 2 (probe cost):** Addressed the intent, but with a different guard than the literal suggestion. A content pre-guard ("skip when the buffer has no `https`") can't reduce the dominant cost, because reading the visible buffer *is* the cost — you must `readAsync` to inspect content, and `detectLoginAuthUrl` is already cheap (an early per-line `https` search that bails). Instead I added a genuinely-free pre-guard on existing state: a `/login` modal requires a **live claude process**, so `restingPtyAuth` now skips the read (and drops any stale confirmation) when `lastClaudeAlive.get(id) === false`. Husk done-sessions (claude exited to a shell — very common) are skipped entirely, with **no** detection slowdown for live sessions and no reliance on a heuristic content check. `undefined` (pre-first-sweep) is treated as maybe-alive so detection is never missed on a cold start. Test added.

(The reported `fallow` "2 clone groups" are pre-existing/inherited — the audit gate reports "✓ No issues in 5 changed files".)

scoop (commented):
## Re-review outcome

Both prior-round points are genuinely addressed by this diff. One new, minor (non-blocking) cosmetic defect from the shared PTY confirmation cache.

### Prior point 1 — blocked-path stability gate: ADDRESSED
The blocked `/login` reconstruction now routes through the same two-read gate as the resting path. Verified at `src/poller.ts:1064-1066` (`this.detectAuth(s) ?? this.confirmLoginUrl(id, { url: detectLoginAuthUrl(visible), tail: [] })?.url`) against the shared gate `confirmLoginUrl` at `src/poller.ts:1256-1272`. Traced it: on read #1 `prev` differs from the reconstruction so it returns the prior (null) confirmation → no `authUrl`; the URL only latches once two consecutive reads are equal (`prev === recon.url`, `src/poller.ts:1266`). A mid-paint partial that passes `isAuthUrl` therefore never surfaces. `emitBlock` runs before the `lastSig` dedup (`src/poller.ts:1069-1075`) so the second read is not short-circuited. Covered by the new blocked-path test (`test/poller.test.ts:172-185`).

### Prior point 2 — probe cost pre-guard: ADDRESSED (state-based guard)
Instead of the literal `no-https` content guard, `restingPtyAuth` now skips the read (and drops stale confirmation) when `this.lastClaudeAlive.get(id) === false` (`src/poller.ts:1218-1222`). Verified `lastClaudeAlive` semantics at `src/poller.ts:463-466` (set from the throttled liveness sweep; `alive === false` means claude exited to a bare shell — a husk that cannot be at a `/login` modal). `undefined` (pre-first-sweep) is deliberately treated as maybe-alive, so a live `/login` is never suppressed and cold-start detection is not missed. This bounds the added per-cadence herdr reads to sessions that could actually be at a prompt, which satisfies the intent of the original note.

### New minor finding (non-blocking)
`confirmLoginUrl` writes to the process-wide `restAuthPtyConfirmed` cache, and the blocked path feeds it `tail: []` (`src/poller.ts:1066`). That cache is shared with the resting emit, which reads `src.tail` for the banner (`src/poller.ts:1188`). When a `/login` session that herdr classified as `blocked` for >=2 cadences (so `confirmLoginUrl` latched `{url, tail: []}`) then transitions `blocked -> idle`, `onLeaveResting` does NOT clear the caches — its `leaving` predicate only fires on `idle/done -> running/blocked` (`src/poller.ts:1283-1286`). So the resting path returns the carried-over `{url, tail: []}` and emits the AUTH banner with an empty tail. The re-emit gate keys only on the URL (`src/poller.ts:1182`), so once the async probe re-confirms the same URL with the real tail it will NOT re-emit — the banner stays with an empty tail for that episode. Impact is cosmetic (the URL is correct; only the surrounding context lines are missing) and the path is uncommon, hence non-blocking.

### Other checks
- `detectLoginAuthUrl` is exported (`src/auth-url.ts:224`) and imported consistently in `src/index.ts` and `src/poller.ts:21`; the `BORDER_ENDS_RE` box-drawing range `U+2500-U+257F` plus `\s`/`|` is a valid char class and cannot eat URL content (ASCII `-` is outside the range; `|` is never an unencoded URL char). The join's STOP-before-APPEND logic and `isAuthUrl` validation guard against forged/partial URLs.
- `readVisibleBuffer` (`src/index.ts`) reuses the existing `matchAgent`/`herdr.read` seam and adds a try/catch (guarded -> null), a safe change to the autopilot `readTail`/`pendingAuthUrl` callbacks; the `(id) => string | null` contract is unchanged.
- All five new cache maps are pruned in `pruneInactive` (`src/poller.ts:669-696`) and the detection caches additionally cleared on the leave-resting edge; `lastAuthUrlEmitted` correctly follows the block via `clearBlock` (`src/poller.ts:1119-1122`).
- No message-catalog / i18n surface is touched by this PR.

scoop:
Fixed in d155b56 — valid catch.

The blocked path cached `tail: []` into the shared `restAuthPtyConfirmed`; on a `blocked → idle` transition (not a leave-resting edge, so caches persist) the resting emit inherited it and surfaced the AUTH banner with an empty, never-corrected tail.

Chose the first of your two options — **don't cache a placeholder tail**: the blocked path already computes the real tail (`reason.tail` = `tailLines(visible)`) for its own emit, so I pass that to `confirmLoginUrl` instead of `[]`. The inherited entry now carries a proper tail, and the transition stays seamless (same URL throughout, no clear/flicker). Preferred this over clearing the cache on leave-blocked, which would blank the banner and re-confirm over ~2 cadences.

Added a `blocked → idle` inheritance test asserting the resting-path AUTH emit carries a non-empty tail (would be `[]` before the fix).

scoop (commented):
## Re-review

### Prior finding (RESOLVED)

> src/poller.ts: blocked-path `confirmLoginUrl(id, { url: detectLoginAuthUrl(visible), tail: [] })` writes `tail: []` into the shared `restAuthPtyConfirmed` cache; on a `blocked -> idle` transition the resting emit surfaces the AUTH banner with an empty tail and never self-corrects.

The blocked path now passes the real tail instead of the `[]` placeholder:

```
this.confirmLoginUrl(id, { url: detectLoginAuthUrl(visible), tail: reason.tail })?.url
```
(src/poller.ts:1070)

Verified `reason.tail` is a genuine non-empty tail for this branch: `classifyBlocked` sets `const tail = tailLines(text)` and the awaiting-input return carries that same `tail` (src/blocked.ts:79, src/blocked.ts:108). `text` here is the `visible` buffer read at src/poller.ts:1030-1031, so `reason.tail === tailLines(visible)`.

Cache-inheritance path re-traced and confirmed sound:
- `blocked -> idle` is NOT a leave-resting edge (`onLeaveResting` fires only on `idle/done -> running/blocked`, src/poller.ts:1290-1293), so `restAuthPtyConfirmed` persists across the transition. On the idle tick, `restingPtyAuth` returns the inherited confirmed entry (src/poller.ts:1232) — now carrying the real blocked-state tail, which is appropriate context for the `/login` auth banner.
- The reverse edge `idle/done -> blocked` IS a leave-resting edge, so `clearRestingAuthState` wipes observed/confirmed (src/poller.ts:1297-1303) BEFORE `maybeClassify` runs (`onLeaveResting` at src/poller.ts:546 precedes the `maybeClassify` dispatch at src/poller.ts:553), so the blocked path's two-read gate starts fresh. Consistent.

The author's note claim matches the diff.

### Other verification
- `detectLoginAuthUrl` helpers `trimUrl`/`isAuthUrl` exist (src/auth-url.ts:23,32); `BORDER_ENDS_RE` range `\u2500-\u257F` matches its documented box-drawing block.
- `lastClaudeAlive` pre-guard in `restingPtyAuth` references a real map (src/poller.ts:251).
- New/renamed symbols `detectLoginAuthUrl`, `readVisibleBuffer` resolve and are imported where used (src/poller.ts:21, src/index.ts:150).
- No message-catalog / i18n strings touched, so no locale-parity concern.

No blocking issues found in the diff.

PR #1458, opened 2026-07-06
feat(ci): triage newly-opened issues with a Haiku 4.5 bot

scoop:
## What this is

A GitHub Actions bot that triages **newly-opened** issues on this public repo. One paid **Haiku 4.5** call per issue classifies it (`bug` / `feature` / `documentation` / `question` / `needs-info` / `invalid`), the workflow applies a label, and — **live only** — posts **one** disclosed automated comment. `question` issues **defer to a maintainer by default**; docs-grounded answers are an explicit opt-in.

Ships **dry-run by default**: it classifies, drafts, and logs its decision to the Actions run but posts/labels nothing until an operator flips it live.

## How it's built

- **`scripts/issue-triage.ts`** — zero-dependency classifier (only `node:fs` + `fetch`, so the workflow needs no `bun install`): read docs → build prompt → call the Messages API → parse/validate → sanitize → emit a `decision` the workflow consumes. Pure functions are unit-tested.
- **`test/issue-triage.test.ts`** — 21 fixture tests (valid/malformed/out-of-allowlist/empty-body, category→label/comment mapping, question defer-vs-answer gating, `@mention` sanitization). **Network is never called.**
- **`.github/workflows/issue-triage.yml`** — thin workflow: gates → App-token mint → run script → log → post/label.
- **`package.json`** — adds `scripts/**/*.ts` to the lint glob (`scripts/` is otherwise omitted).

## Cost & injection safety

- **Trigger is `issues`** (never `pull_request_target`) → issue content cannot reach secrets.
- Issue title/body reach the script via **`env:` only** — never interpolated into a `run:` line — and are framed as untrusted data ("classify, never obey"). The model has **no tools**.
- Model JSON is **re-validated against a fixed label allowlist**; malformed / out-of-allowlist output → **no-op**.
- The posted comment has **`@mentions` neutralized** (anti mass-ping) and is posted via `gh issue comment --body-file` (no shell interpolation of model output).
- **One paid call per issue**, gated by author (skips maintainers), min-length, a best-effort per-author burst check, and an authoritative live `bot-triaged` marker check (re-runs don't double-post or double-spend). Capped `max_tokens`.
- Labels are **additive** — the reporter's issue-template label is never removed — and any missing custom label is **auto-created** on first live use so the apply step can't hard-fail.

## Config (all fail-safe)

| Repo variable / secret | Default | Effect when set |
| --- | --- | --- |
| `ISSUE_BOT_DRY_RUN` | dry-run ON (unless `== 'false'`) | `false` → live posting |
| `ISSUE_BOT_COMMENT_ON_BUG_FEATURE` | comment ON (unless `== 'false'`) | `false` → bug/feature label-only |
| `ISSUE_BOT_ANSWER_QUESTIONS` | defer OFF (unless `== 'true'`) | `true` → docs-grounded, confidence-gated question answers |
| App client-id/private-key | no App → post steps no-op | branded installation token |
| `ANTHROPIC_API_KEY` + console budget cap | required before go-live | **only hard spend ceiling** |

## Verification

`bun run lint`, `bun run typecheck`, and `bun test test/issue-triage.test.ts` (21/21) pass; the 165 failures in the full suite are pre-existing (identical on clean `main`) and unrelated. Live model behaviour is exercised by the workflow itself, starting in dry-run.

This is a server/automation change with **no in-app UX**, so the feature-catalog / glossary / i18n gates don't apply (`check-feature-catalog.sh` doesn't fire — no `ui/src/**` paths change).

```shepherd:manual-steps
- [ ] Create a branded GitHub App and install it on this repo (permissions: Issues read & write, Contents read); set repo variable ISSUE_TRIAGE_APP_CLIENT_ID and secret ISSUE_TRIAGE_APP_PRIVATE_KEY
- [ ] (Optional) Pre-create labels question / needs-info / invalid / bot-triaged to set custom colours — otherwise the workflow auto-creates any missing label on first live use
- [ ] Set the dedicated ANTHROPIC_API_KEY secret AND a budget cap on that key in the Anthropic console — this is the only hard spend ceiling
- [ ] POST-MERGE: observe ≥5 dry-run issues, eyeball the logged decisions in the Actions runs, then set ISSUE_BOT_DRY_RUN=false to go live
- [ ] POST-MERGE: optionally set ISSUE_BOT_COMMENT_ON_BUG_FEATURE=false (bug/feature label-only) and/or ISSUE_BOT_ANSWER_QUESTIONS=true (docs-grounded question answers) after eyeballing dry-run quality
```

### DIFF
--- a/.github/workflows/issue-triage.yml
+++ b/.github/workflows/issue-triage.yml
@@ -1,7 +1,7 @@
 name: Issue triage
 
 # Triage newly-opened issues with one paid Haiku 4.5 call: classify → label →
-# (unless in dry-run) post ONE disclosed automated comment. See .shepherd-plan.md.
+# (unless in dry-run) post ONE disclosed automated comment.
 #
 # Trigger is `issues` (NOT pull_request_target), so issue content can never reach
 # secrets. Issue title/body are passed to the script via `env:` only — never
@@ -121,6 +121,14 @@ jobs:
           # Category label(s) + the bot-triaged marker (additive; template label kept).
           mapfile -t labels < <(printf '%s' "$DECISION" | jq -r '.labels[]')
           labels+=("bot-triaged")
+          # Self-heal: ensure every label exists before --add-label (which requires
+          # it) so a not-yet-created custom label (needs-info / invalid / bot-triaged)
+          # can't hard-fail the step under `set -e`. Runs live-only; existing labels
+          # (bug / enhancement / documentation) no-op via `|| true`. gh assigns a
+          # random colour when --color is omitted; operators may pre-create to style.
+          for l in "${labels[@]}"; do
+            gh label create "$l" --repo "$REPO" >/dev/null 2>&1 || true
+          done
           args=()
           for l in "${labels[@]}"; do args+=(--add-label "$l"); done
           gh issue edit "$ISSUE_NUMBER" --repo "$REPO" "${args[@]}"
--- a/.github/workflows/issue-triage.yml
+++ b/.github/workflows/issue-triage.yml
@@ -1,0 +1,133 @@
+name: Issue triage
+
+# Triage newly-opened issues with one paid Haiku 4.5 call: classify → label →
+# (unless in dry-run) post ONE disclosed automated comment. See .shepherd-plan.md.
+#
+# Trigger is `issues` (NOT pull_request_target), so issue content can never reach
+# secrets. Issue title/body are passed to the script via `env:` only — never
+# interpolated into a `run:` line — and are treated as untrusted data. The model
+# has no tools; its JSON is re-validated against a fixed allowlist in the script,
+# and the posted comment has @mentions neutralized.
+on:
+  issues:
+    types: [opened]
+
+# One run per issue; don't cancel an in-flight run for the same issue.
+concurrency:
+  group: issue-triage-${{ github.event.issue.number }}
+  cancel-in-progress: false
+
+# Least privilege. `issues: write` covers reading current labels (idempotency
+# pre-check) and applying labels/comments via the fallback token; the branded
+# App installation token (when configured) carries its own down-scoped grant.
+permissions:
+  contents: read
+  issues: write
+
+jobs:
+  triage:
+    runs-on: ubuntu-latest
+    # Skip maintainers entirely — only triage non-maintainer authors. Author
+    # association is a property of the issue event, so the payload value is correct.
+    if: >-
+      github.event.issue.author_association != 'OWNER'
+      && github.event.issue.author_association != 'MEMBER'
+      && github.event.issue.author_association != 'COLLABORATOR'
+    steps:
+      - name: Checkout (docs + triage script — our own trusted code)
+        uses: actions/checkout@v7
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
+          bun-version: latest
+
+      # Authoritative idempotency + best-effort abuse guard, both UNPAID. The
+      # event payload's labels are frozen at open time and never carry the marker,
+      # so we query the issue's CURRENT labels here to gate re-runs.
+      - name: Pre-flight (marker check + per-author burst guard)
+        id: preflight
+        env:
+          GH_TOKEN: ${{ github.token }}
+          REPO: ${{ github.repository }}
+          ISSUE_NUMBER: ${{ github.event.issue.number }}
+          AUTHOR: ${{ github.event.issue.user.login }}
+        run: |
+          set -euo pipefail
+          already=$(gh api "repos/${REPO}/issues/${ISSUE_NUMBER}/labels" \
+            --jq 'any(.[]; .name == "bot-triaged")')
+          echo "already_triaged=${already}" >> "$GITHUB_OUTPUT"
+          echo "bot-triaged already present: ${already}"
+
+          # Best-effort: if this author opened >=5 issues in the last hour, skip
+          # the paid call. Search indexing lags and has no cross-run state — this
+          # is layered UNDER the mandatory Anthropic console budget cap, not a
+          # substitute for it. Never fail the job on a search hiccup.
+          since=$(date -u -d '1 hour ago' +%Y-%m-%dT%H:%M:%SZ)
+          count=$(gh api -X GET search/issues \
+            -f q="repo:${REPO} type:issue author:${AUTHOR} created:>=${since}" \
+            --jq '.total_count' 2>/dev/null || echo 0)
+          if [ "${count:-0}" -ge 5 ]; then burst=true; else burst=false; fi
+          echo "author_burst=${burst}" >> "$GITHUB_OUTPUT"
+          echo "author issues in last hour: ${count} (burst=${burst})"
+
+      # Mint a scoped, short-lived branded App token when configured. Non-fatal:
+      # on failure the token output is empty and the post/label step no-ops.
+      - name: Mint GitHub App token (when configured)
+        id: app-token
+        if: ${{ vars.ISSUE_TRIAGE_APP_CLIENT_ID != '' }}
+        continue-on-error: true
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        with:
+          client-id: ${{ vars.ISSUE_TRIAGE_APP_CLIENT_ID }}
+          private-key: ${{ secrets.ISSUE_TRIAGE_APP_PRIVATE_KEY }}
+          permission-contents: read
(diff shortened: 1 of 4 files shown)


scoop (commented):
## Summary

This PR adds an issue-triage bot: a GitHub Actions workflow (`.github/workflows/issue-triage.yml`), a Bun/TypeScript classifier (`scripts/issue-triage.ts`), unit tests, and a lint-path update in `package.json`. The design is careful — untrusted issue text is passed via `env:` only (never interpolated into a `run:`), the model gets no tools, its JSON is re-validated against a fixed allowlist, `@`-mentions are neutralized, the paid call is gated by an unpaid pre-flight (marker + burst), and posting is fail-safe (requires an App token AND `ISSUE_BOT_DRY_RUN == 'false'`). I could not find a blocking logic or security defect on the executing path.

### Verification performed
- Category→label map and allowlist logic in `scripts/issue-triage.ts:38-46` is internally consistent; `validateDecision` (`scripts/issue-triage.ts:127-159`) uses `category` as authoritative and treats `labels` purely as a containment gate — matches the doc comments and the tests in `test/issue-triage.test.ts:44-101`.
- Labels `bug`/`documentation`/`enhancement` used by the map are real repo labels — confirmed against the issue templates `.github/ISSUE_TEMPLATE/bug.yml:3`, `documentation.yml:3`, `feature.yml:3`.
- Repo URLs `erwins-enkel/shepherd` in `scripts/issue-triage.ts:35-36` match the origin remote and README (e.g. `README.md:3`).
- Default docs dir `docs-site/src/content/docs` (`scripts/issue-triage.ts:307`) exists; `readDocs` is only invoked when answers are enabled (`scripts/issue-triage.ts:319`).
- Action pins (`actions/checkout@v7`, `oven-sh/setup-bun@0c5077e… #v2.2.0`, `actions/create-github-app-token@bcd2ba4… #v3.2.0`) match the versions already used across `.github/workflows/` — consistent, not a finding.
- Locale parity: the `TXT` catalog (`scripts/issue-triage.ts:174-193`) defines identical keys for `en` and `de` (`header`, `footer`, `deferQuestion`, `invalidRedirect`, `verifyNote`) — both complete.
- Workflow `if` short-circuit safety: the live-apply gate checks `steps.triage.outputs.decision != ''` before `fromJSON(...)` (`.github/workflows/issue-triage.yml:112-117`), so `fromJSON` is not evaluated when the triage step was skipped (burst / already-triaged).
- Empty/absent issue body → `gateBody` skips before any paid call (`scripts/issue-triage.ts:311-316`); tests cover this (`test/issue-triage.test.ts:15-27`).

## Non-blocking notes

The items below are recorded in `findings`. None block.

1. **Custom labels are never created.** `needs-info` and `bot-triaged` are referenced only by the new files (grep confirms: only `scripts/issue-triage.ts`, `.github/workflows/issue-triage.yml`, `test/issue-triage.test.ts`) and there is no label-config in the repo (`.github/labels.yml` absent). `question`/`invalid` are GitHub default labels, but `needs-info` and `bot-triaged` are custom. In live mode the apply step runs `gh issue edit --add-label` under `set -euo pipefail` (`.github/workflows/issue-triage.yml:120-126`), which fails hard if a label does not exist — aborting the whole apply (no labels, no marker, no comment). Label existence is GitHub-side state I cannot verify from this repo (see Limitations), so this is a recommendation, not a proven bug: ensure `needs-info` and `bot-triaged` exist before enabling live mode.

2. **`readDocs` is not fault-isolated like the API call.** `const docs = answersEnabled ? readDocs(docsDir) : null;` (`scripts/issue-triage.ts:319`) runs outside any try/catch, whereas the API call is wrapped and degrades to a graceful `skip` (`scripts/issue-triage.ts:321-329`). If the docs dir is missing/unreadable (only when answers are enabled), `readdirSync` throws and the job exits 1 instead of skipping. Consider wrapping it to emit `skip` consistently.

3. **Dangling `.shepherd-plan.md` reference.** Both `.github/workflows/issue-triage.yml:5` and `scripts/issue-triage.ts:6,26` say “See .shepherd-plan.md”, but that file is not present in the repo — a dead doc pointer.

4. **Lint path targets a single file, not the dir.** `package.json:10` appends `scripts/issue-triage.ts` rather than `scripts`; every other entry is a directory, so future files under `scripts/` won’t be linted. Minor consistency nit.

## Limitations (could not verify from this repo)
- Whether the GitHub labels `needs-info` and `bot-triaged` (and the `question`/`invalid` defaults) actually exist on the repo is GitHub-side state, not present here; note 1 is therefore a recommendation rather than a confirmed defect.
- The model id `claude-haiku-4-5` (`scripts/issue-triage.ts:25`) is an Anthropic-side alias; not verifiable against this repo. Assumed valid as written.
- The Anthropic Messages API request/response shape (`scripts/issue-triage.ts:227-266`) is an external contract not present in this repo; assumed as written.

scoop:
All four review points addressed in 2dd153cd:

1. **Labels** — the live apply step now self-heals: it runs `gh label create <l> || true` for each label (incl. `needs-info`/`invalid`/`bot-triaged`) before `--add-label`, so a not-yet-created custom label can't hard-fail the step under `set -e`. Manual label-creation is now optional (only to set custom colours); PR body updated.
2. **readDocs** — wrapped in try/catch; an unreadable docs dir now degrades to classify-only (questions defer) instead of exiting 1.
3. **Dangling ref** — dropped the `.shepherd-plan.md` mentions in the script + workflow (that file is git-ignored, so it's intentionally not in the repo).
4. **Lint glob** — broadened to `'scripts/**/*.ts'` (covers future TS scripts) rather than the bare `scripts` directory: the dir also contains `.mjs` files (`next-version.mjs`, `register-merge-driver.mjs`, etc.) that trip `no-undef` on Node globals under the flat config, so adding the whole dir would break lint. The quoted glob is expanded by eslint's own globber (shell-independent) and I verified it includes the top-level `scripts/issue-triage.ts`.

scoop (commented):
Re-review of the issue-triage bot. All four points from the previous round are genuinely addressed by this diff, and I found no new blocking defects.

## Prior points — verified resolved
1. **Labels never created** — the live apply step self-heals: `.github/workflows/issue-triage.yml:135-137` runs `gh label create "$l" --repo "$REPO" >/dev/null 2>&1 || true` for every label (category label + `bot-triaged`, incl. `needs-info`/`invalid`) before `gh issue edit --add-label`. Pre-existing labels no-op via `|| true`; the app token declares `permission-issues: write` (`.github/workflows/issue-triage.yml:91`) which covers label creation. Resolved.
2. **`readDocs` outside try/catch** — now wrapped: `scripts/issue-triage.ts:505-513` catches and degrades to classify-only (questions defer) instead of exiting 1. Resolved.
3. **Dangling `.shepherd-plan.md`** — grep of both new files (`.github/workflows/issue-triage.yml`, `scripts/issue-triage.ts`) returns no `shepherd-plan` reference. Resolved.
4. **Lint glob** — `package.json:157` now appends `'scripts/**/*.ts'` (was the bare file). Resolved.

## Correctness checks performed
- Repo slug in `DISCUSSIONS_URL`/`DOCS_URL` (`scripts/issue-triage.ts:197-198`) is `erwins-enkel/shepherd`, matching `README.md:3`.
- Docs dir default `docs-site/src/content/docs` exists (contains 14 md files) — verified via Glob.
- Default model `claude-haiku-4-5` (`scripts/issue-triage.ts:189`) is consistent with existing repo usage (`ui/src/lib/usage-mock.ts:288`).
- Locale parity: `TXT` has both `en` and `de` entries with identical key sets (`scripts/issue-triage.ts:334-353`), matching the repo's en+de convention.
- Decision-JSON contract agrees end to end: the `process` emit (`scripts/issue-triage.ts:534-542`) produces `action`/`labels[]`/`comment`/`shouldComment`, exactly what the workflow consumes via `jq -r '.labels[]'`, `jq -r '.shouldComment'`, `jq -r '.comment'`, and `fromJSON(...).action == 'process'` (`.github/workflows/issue-triage.yml:119,128,144-145`). `JSON.stringify` escapes newlines so the `$GITHUB_OUTPUT` value stays single-line. GitHub Actions `&&` short-circuits, so `fromJSON('')` is never evaluated when the triage step was skipped (`decision != ''` guard precedes it).

## Nit (non-blocking)
- `.github/workflows/issue-triage.yml:26-28`: the `permissions` comment says `issues: write` covers "applying labels/comments via the fallback token," but the apply step (`.github/workflows/issue-triage.yml:121`) uses only the branded App token (`steps.app-token.outputs.token`) and gates on it being non-empty; the fallback `github.token` is used solely for the read-only pre-flight (`:57`). The comment slightly overstates what the default token does — worth tightening to avoid implying a label/comment fallback path that does not exist.

## Stated limitation
- The Anthropic `messages` request shape and `claude-haiku-4-5` alias resolution are external-API concerns not verifiable against this repo; assumed correct as written.

PR #1459, opened 2026-07-06
fix(diagnostics): retry gh auth probe, stop false "not logged in"

scoop:
## Problem

The **Diagnostics** panel intermittently showed **gh → Problem: "GitHub CLI missing or not logged in — run gh auth login."** on hosts where `gh` is installed and authenticated. Re-running cleared it; it returned after a while.

**Root cause:** `ghProbe` ran `gh auth status` once with a 5s timeout and collapsed **every** failure — a timeout, a transient keyring/D-Bus stall, or a cold `gh` under load — into the single hard verdict `error` + `diagnostics_hint_gh_not_authenticated`. The 4s boot kick + 6h `setInterval` re-check pushed that to all clients and the 60s TTL re-probed, so an authenticated host flickered red and cleared on a fast manual re-run. The real error was swallowed (no logging), so the exact transient trigger was invisible.

`gh auth status` reads the token from the OS keyring (no network), which is the transient surface.

## Fix

Rewrite `ghProbe` as a **bounded retry loop** that classifies the final failure by **process disposition** — not by parsing gh's output (a real logout and an invalid/expired token print different text, and success identity lands on stdout):

| gh failure (forge repo) | Before | After |
|---|---|---|
| binary missing (`ENOENT`) | error · `gh_missing` | unchanged |
| logged out / invalid token / bad scopes (gh **exits** non-zero) | error · run gh auth login | unchanged (still actionable) |
| timeout / transient stall (probe **kills** gh) | error · run gh auth login | retry → **warning · `gh_unverified`** ("couldn't verify — timed out") |

- Success on any attempt ⇒ `ok`; a transient blip that clears on retry resolves cleanly.
- Killed/timed-out on every attempt ⇒ soft `gh_unverified` warning + a **disposition-only** server log (`code`/`killed`/`signal` — never stderr/stdout, so no account identity leaks).
- A dedicated shorter per-attempt timeout (`GH_PROBE_TIMEOUT_MS = 3s` × `GH_PROBE_ATTEMPTS = 2`) caps the probe at ≈6.25s instead of a possible 3×5s stall.
- New `diagnostics_hint_gh_unverified` message (EN + DE); deliberately **no** `DOC_LINKS`/remediation mapping, so the warning row shows guidance-free text (no misleading "How to fix" / "Fix" button).

Overall system health also softens from error-red to a warning when gh merely can't be verified, instead of marking the whole host broken.

## Tests

`test/diagnostics.test.ts`: timeout → `gh_unverified` (regression guard), retry-recovery → `ok`, retry-exhaustion call-count, persistent non-zero exit → `gh_not_authenticated`, forge-downgrade, and log-purity (no identity/stderr in the log). Existing gh tests keep their assertions (plain-`Error` → exited-verdict branch); only `healthyDeps` gains `ghProbeRetryDelayMs: 0`.

## Verification

`bun run typecheck`, `bun run lint`, `bun test ./test/diagnostics.test.ts` (62 pass), `cd ui && bun run check:i18n`, `check:glossary`, `check:announcement-versions`, and `fallow audit` all clean. The full `bun test` failure count is identical with/without this change (pre-existing headless DOM/browser-env tests).

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -63,6 +63,16 @@ export const DIAGNOSTICS_TTL_MS = 60_000;
 // Per-probe exec timeout: a timed-out probe RESOLVES to its defined non-OK state
 // (never rejects the Promise.all), so one hung binary can't stall the batch.
 export const DIAGNOSTICS_PROBE_TIMEOUT_MS = 5_000;
+// gh auth probe hardening (#623 follow-up): `gh auth status` reads the token from the OS
+// keyring, so a locked keyring / D-Bus stall / cold `gh` under load can transiently blow
+// the probe budget and — pre-fix — masquerade as "not logged in". The probe now RETRIES a
+// timed-out attempt and only reports a hard-auth error when gh actually exits with a
+// verdict. A dedicated, SHORTER per-attempt timeout keeps the bounded retry loop from
+// stalling the batch: GH_PROBE_ATTEMPTS × GH_PROBE_TIMEOUT_MS (+ delays) ≈ 6.25s worst case,
+// vs. 3×5s if it reused DIAGNOSTICS_PROBE_TIMEOUT_MS. Healthy gh answers in ~0.25s.
+export const GH_PROBE_ATTEMPTS = 2;
+export const GH_PROBE_TIMEOUT_MS = 3_000;
+export const GH_PROBE_RETRY_DELAY_MS = 250;
 // Time budget for an in-app remediation command (POST /api/diagnostics/fix). Far
 // larger than a probe — these are real `curl | bash` installs. On timeout the whole
 // process GROUP is SIGKILLed (see diagnostics.ts runRemediation) so no reparented
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -4,6 +4,9 @@ import {
   BUN_MIN_VERSION,
   DIAGNOSTICS_PROBE_TIMEOUT_MS,
   DIAGNOSTICS_TTL_MS,
+  GH_PROBE_ATTEMPTS,
+  GH_PROBE_RETRY_DELAY_MS,
+  GH_PROBE_TIMEOUT_MS,
   HERDR_MIN_VERSION,
   NODE_MIN_VERSION,
   REMEDIATION_TIMEOUT_MS,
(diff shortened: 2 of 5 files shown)


scoop (commented):
## Summary

The PR fixes the reported bug — a transiently slow `gh auth status` (locked keyring / D-Bus stall / cold `gh`) times out, and the pre-fix code turned that timeout into a hard `gh_not_authenticated` error. It reappeared because the 60s TTL cache (`DIAGNOSTICS_TTL_MS`) expires and re-probes, occasionally re-hitting the transient timeout. The fix adds a bounded retry and classifies the *last* failure by disposition (`killed`/`signal` ⇒ transient timeout ⇒ soft `gh_unverified` warning; plain non-zero exit ⇒ real `gh_not_authenticated`). This is the correct approach.

### Verified
- **Disposition classification is correct against Node semantics.** On `execFile` timeout, Node sets `err.killed=true` + `err.signal` (SIGTERM) and `err.code` is undefined; on a plain non-zero exit it sets `err.code`=exit number with `killed=false`/`signal=null`. `isTransientProbeFailure` (src/diagnostics.ts:35-39) keys on `killed===true || signal non-empty`, so a timeout ⇒ transient and a logged-out exit ⇒ auth verdict. Confirmed against the retry loop and default runner (src/diagnostics.ts:300-322, 204-216).
- **ENOENT short-circuits without retry** (src/diagnostics.ts:310) — correct, binary-missing is deterministic.
- **i18n parity.** New key present in both catalogs: verified ui/messages/en.json:1228 and ui/messages/de.json:1228 (`diagnostics_hint_gh_unverified`). `hintKey` is typed as a bare `string` (src/types.ts:384), so no exhaustive union needs updating.
- **No remediation leak / no bad Fix button.** `diagnostics_hint_gh_unverified` has no `REMEDIATIONS` entry, so `autoFixCommandFor` returns undefined (src/remediations.ts:44-56, 85-88) — the transient warning gets no auto-fix command, as intended. `DOC_LINKS` lookup is a plain Record miss ⇒ undefined, harmless (ui/src/lib/diagnostics-docs.ts:6-11).
- **Log purity.** `logGhUnverified` (src/diagnostics.ts:56-62) reads only `code`/`killed`/`signal`; the test at test/diagnostics.test.ts:483-499 confirms injected stderr/identity never reaches the log.
- **Timing.** Worst case is `2 × GH_PROBE_TIMEOUT_MS + 1 × GH_PROBE_RETRY_DELAY_MS` = 6250ms (src/config.ts:71-73). There is no batch-level wall-clock cutoff — `check()` runs each probe as `run().catch(() => onTimeout)` (src/diagnostics.ts:456) with no timeout wrapper around the whole batch — and the UI fetch has no AbortSignal/timeout (ui/src/lib/api.ts:1291). So the 6.25s (> the old 5s) worst case cannot be truncated externally. Fine.
- **`onTimeout` fallback** for gh downgraded to the soft `gh_unverified` warning (src/diagnostics.ts:424-428) — consistent defense-in-depth since `ghProbe` no longer throws.

### Nits (non-blocking)
1. **src/diagnostics.ts:303-312 — the loop retries a *deterministic* non-zero exit too.** A genuinely logged-out user (real non-zero exit on attempt 1) is not transient, yet the loop still `delay()`s and re-invokes `gh` a second time before concluding `gh_not_authenticated`. Effect: a real auth failure now spawns `gh` twice and adds `GH_PROBE_RETRY_DELAY_MS` latency for no benefit. Defensible (a non-zero exit *could* be a transient blip), but gating the retry on `isTransientProbeFailure(err)` would avoid the wasted attempt. Minor.
2. **src/diagnostics.ts:317-319 — `logGhUnverified` fires even when `forge=false`.** When no forge repo is configured, the outcome is downgraded to `gh_not_required` (via `ghFailure`), but the server still logs `"gh auth status could not be verified"`. That log line then describes a state the user never sees, which is slightly misleading log noise. Consider logging only when the unverified state is actually surfaced.

Neither nit blocks; the core fix is correct and well-tested.

scoop (commented):
## Re-review

Both points from the previous revision are resolved by this diff.

**Prior point 1 — retry no longer fires on a deterministic non-zero exit.** `ghProbe` now breaks out of the loop immediately for a non-transient failure: `if (!isTransientProbeFailure(err)) break;` (src/diagnostics.ts:315). Only a killed/timed-out attempt retries + delays. Confirmed against the new test asserting `calls === 1` for a persistent non-zero exit (test/diagnostics.test.ts:453-471). **Addressed.**

**Prior point 2 — `logGhUnverified` only when the warning surfaces.** The log call is now gated on forge mode: `if (forge) logGhUnverified(lastErr, this.ghProbeAttempts);` (src/diagnostics.ts:323), so a `gh_not_required` downgrade emits no server-side "could not be verified" line. Confirmed against the test `timeout + anyForgeRepo=false → warning + not_required, no unverified log` (test/diagnostics.test.ts:445-465). **Addressed.**

## Verification performed

- **i18n parity**: all five `diagnostics_hint_gh_*` keys exist in both catalogs — verified against ui/messages/en.json:1226-1228,1303,1670 and ui/messages/de.json:1226-1228,1303,1670. The new `diagnostics_hint_gh_unverified` is present in both (en.json:1228, de.json:1228).
- **State validity**: `"warning"`/`"error"` are members of `DiagnosticState` — verified against src/types.ts:376.
- **Disposition classifier**: `isTransientProbeFailure` returns true only when `killed===true` or a non-empty `signal` (src/diagnostics.ts:33-38), matching execFile's timeout→SIGTERM shape; a plain non-zero exit (numeric `code`, no signal) is treated as a real verdict; ENOENT is checked first (src/diagnostics.ts:313) and is non-transient. Consistent with the three test error shapes (test/diagnostics.test.ts:409-410, 462, 469).
- **No spurious auto-fix**: `autoFixCommandFor("diagnostics_hint_gh_unverified")` returns `undefined` since there is no `REMEDIATIONS` entry for the new key (src/remediations.ts:85-88), so the transient-timeout warning correctly carries no Fix command.
- **No outer-timeout truncation**: `check()` runs each probe as `run().catch(() => onTimeout)` with no wrapping timeout (src/diagnostics.ts:462); the bounded ~6.25s retry budget (2×3000ms + 250ms delay, per src/config.ts) runs to completion, and the `onTimeout` fallback is now the soft `gh_unverified` warning as pure defense-in-depth (src/diagnostics.ts:429-433).

No new correctness, security, or quality issues in the diff.

PR #1461, opened 2026-07-06
test(usage): de-flake midnight-straddle timeline heatmap test that reddens doc-PR CI

scoop:
## Why "auto docs commits keep failing CI"

The failing job ([run 28761279077](https://github.com/erwins-enkel/shepherd/actions/runs/28761279077/job/85277117591)) is **not a docs problem** — it's a wall-clock-dependent flaky UI test.

`ui/src/lib/components/Usage.browser.test.ts` built its two timeline hour-buckets relative to `Date.now()` (`now-2h`, `now-1h`) and asserted the heatmap renders exactly **one** 24-cell day row. `TimelineLens.svelte` correctly groups usage hours into **contiguous local-calendar-day rows** — so during the **01:00–01:59 UTC** hour, `now-2h` = 23:00 *yesterday* and `now-1h` = 00:00 *today* straddle midnight → **2 rows → 48 cells** → `expected 48 to be 24`. The failing run was at **01:06 UTC**.

Doc PRs run the full `verify` suite (not just the doc diff), so any doc-PR whose CI lands in that ~1-hour nightly window goes red — hence the recurring failures.

## Fix

Anchor both buckets to a fixed mid-day point of "today" so they always share one local calendar day, at every wall-clock time. Mirrors the deterministic fixed-date pattern the sibling `usage/TimelineLens.browser.test.ts` already uses. **Test-only — component behavior unchanged.**

## Verification

A/B under a spoofed clock in the failing window (`TZ=Etc/GMT+5`, local 01:41):
- **Before:** Timeline test fails `expected 48 to be 24` — reproduces the CI failure exactly.
- **After:** all 10 tests pass.

Also green: `Usage.browser.test.ts` (normal clock), `ui bun run check` (0 errors), prettier + eslint on the changed file.

Scope-checked: this is the only test in the repo with the day-straddle bug (`TimelineLens.browser.test.ts` uses fixed dates; `GitRail`/`DocAgentControl` use `Date.now()` only for relative timestamps).

### DIFF
--- a/ui/src/lib/components/Usage.browser.test.ts
+++ b/ui/src/lib/components/Usage.browser.test.ts
@@ -49,12 +49,22 @@ const inlineProjections: UsageProjection[] = [
   { window: "WK", projectedPct: 41, resetAt: BASE + 58 * H, burnRatePerHour: 31_000 },
 ];
 
+// Pin the timeline buckets to a fixed mid-day point of "today" so both hours always fall on the
+// SAME local calendar day → one 24-cell heatmap row, at every wall-clock time. A Date.now()-relative
+// fixture flakes during the 01:00 local hour (one bucket lands on the previous day → 2 rows /
+// 48 cells), which kept the doc-PR `verify` CI red whenever it ran in that nightly window.
+const TODAY_NOON = (() => {
+  const d = new Date(BASE);
+  d.setHours(12, 0, 0, 0);
+  return d.getTime();
+})();
+
 const inlineTimeline: UsageTimeline = {
   range: "7d",
   generatedAt: BASE,
   hours: [
-    { hourStart: BASE - 2 * H - (BASE % H), units: 12 },
-    { hourStart: BASE - H - (BASE % H), units: 40 },
+    { hourStart: TODAY_NOON - 2 * H, units: 12 },
+    { hourStart: TODAY_NOON - H, units: 40 },
   ],
   totalUnits: 52,
   peakHourUnits: 40,


scoop (commented):
## Summary

Test-only change that de-flakes the timeline-heatmap assertion in `ui/src/lib/components/Usage.browser.test.ts`. It replaces UTC-hour-aligned, `Date.now()`-relative fixture buckets with two buckets pinned to a fixed mid-day point of the current local day. This is the correct fix and I found no bugs.

## Verification

- `TimelineLens.svelte:65-80` folds hours into rows keyed by `startOfLocalDay(h.hourStart)` and emits contiguous calendar days from `latest` down to `earliest`. When both fixture hours share a local day, `earliest === latest` ⇒ exactly **1 row / 24 cells**; a day-boundary straddle ⇒ **2 rows / 48 cells**.
- The test asserts exactly 24 cells (`ui/src/lib/components/Usage.browser.test.ts:157`), so any straddle fails it — this is the flake the PR targets.
- Old fixture `BASE - 2*H - (BASE % H)` (removed line) aligned buckets to UTC-hour boundaries, while grouping is local (`new Date(h.hourStart).getHours()` and `startOfLocalDay`, `TimelineLens.svelte:55-56`). In the ~00:00–02:00 local window the earlier bucket lands on the previous local day ⇒ 2 rows. Real, correctly-diagnosed flake.
- New `TODAY_NOON` = local noon of BASE's day (`ui/src/lib/components/Usage.browser.test.ts:56-60`); buckets at `TODAY_NOON - 2*H` (10:00 local) and `TODAY_NOON - H` (11:00 local) always fall on the same local calendar day at any wall-clock time — no DST transition occurs between 10:00 and noon — so `earliest === latest` ⇒ 1 row, 24 cells, deterministically. `hiddenDays` stays 0 (verified against `TimelineLens.svelte:82-83`).
- Buckets are distinct hours (10 vs 11), so no cell collision; `totalUnits: 52` still equals 12+40 and `peakHourUnits: 40` is unchanged, keeping the fixture internally consistent.

No findings.

PR #1471, opened 2026-07-06
fix(herdr-update): recover via gated relaunch, not nonexistent `server start` (#1410)

### LINKED ISSUE #1410: herdr-update recovery calls nonexistent 'herdr server start'
The failure-recovery path in `src/herdr-update.ts` (buildUpdateScript) runs `herdr server start || true` when `herdr update --handoff` fails, to bring the server back so panes can reattach.

But `herdr server start` does not exist in the current herdr CLI (verified against herdr 0.7.1: the `herdr server` subcommands are `stop`, `live-handoff`, `reload-config`, and the `*-manifests` commands; a bare `herdr server` runs a headless server in the foreground). The `|| true` masks the unknown-command error, so the recovery step is a silent no-op — after a failed update with a dead server, panes stay orphaned until the operator starts a server manually.

Surfaced during #1411 research (one-click shepherd restart), deliberately not fixed there. Needs either the correct CLI invocation for (re)starting a detached herdr server, or removal of the misleading recovery line plus an operator-facing hint in the update log.

kai-osthoff:
Fixes #1410.

## The bug

The herdr-update failure-recovery path ran `herdr server start || true`. But
`herdr server start` is **not a real subcommand** (verified against herdr 0.7.1 —
the `herdr server` verbs are `stop`, `live-handoff`, `reload-config`, and the
`*-agent-manifests` commands; bare `herdr server` runs a headless foreground
server). `|| true` masked the unknown-command error, so the recovery step was a
**silent no-op**: after a failed update that left the server dead, panes stayed
orphaned until the operator manually relaunched herdr.

## The fix

Restore the auto-recovery commit #413 *intended* (the bug was only the wrong
verb), using a repo-corroborated, gated invocation:

- **Probe = `herdr agent list` exit status** — the exact throw-on-unreachable
  signal `broadcast()` already catches at `src/service.ts:3517-3520`. Proven both
  to exist and to fail-on-unreachable by shipping code; pure exit status, no
  output-format/JSON coupling.
- **Relaunch only when unreachable.** On a failed `herdr update --handoff`, if
  `agent list` reports the server down, `setsid herdr server </dev/null >/dev/null
  2>&1 &` relaunches a detached server so the orphaned agent targets — which
  **outlive a dead server** (#413) — reattach via the app's Reconnect /
  `viewport_herdr_unreachable` path. A **reachable** server is left untouched,
  avoiding the #241/#314 unconditional-restart hazard.
- **Threads the configured `HERDR_BIN`** (shell-quoted) through the update, probe,
  and relaunch, so a custom-binary host isn't missed (mirrors `restart.ts`).

```sh
if [ "$rc" -ne 0 ]; then
  if timeout 10 'herdr' agent list >/dev/null 2>&1; then
    echo '>>> herdr-update: update failed — herdr server still reachable (agent list ok); live sessions untouched'
  else
    echo '>>> herdr-update: update failed, herdr server unreachable — relaunching a detached server so orphaned sessions reattach'
    setsid 'herdr' server </dev/null >/dev/null 2>&1 &
  fi
fi
```

## Caveat (documented in-code, honestly out of scope)

The update script is a plain child of shepherd, so the `setsid herdr server` it
spawns lives in `shepherd.service`'s cgroup (no `KillMode` → default kills the
whole cgroup on `systemctl restart shepherd`). The recovery server restores
service immediately but isn't durable across a *later* shepherd restart — the
still-alive targets then re-orphan and recover again (**never lost**). A
cgroup-durable server would need `systemd-run`, which this script deliberately
avoids. This is strictly better than the status-quo no-op.

One behavior can't be verified end-to-end without an actual failed protocol-bump
update (a fresh `herdr server` re-adopting the previous server's orphaned
targets) — asserted by #413's design + the herdr socket model, and the change is
gated + fail-safe.

## Testing

- `bun run lint` ✓
- `bun test ./test/herdr-update.test.ts` ✓ (18 pass) — covers the gated
  probe/relaunch command forms, custom-`HERDR_BIN` threading, absence of the
  bogus `herdr server start`, and the marker count.
- Rendered script passes `bash -n` (incl. a custom binary path with spaces).

[no-feature-entry] — server-only internal plumbing; no user-facing UI surface.

### DIFF
--- a/src/herdr-update.ts
+++ b/src/herdr-update.ts
@@ -57,12 +57,18 @@ export function buildUpdateScript(
   logPath: string,
   from?: string | null,
   to?: string | null,
+  herdrBin: string = config.herdrBin,
 ): string {
   const f = sanitizeVersion(from);
   const t = sanitizeVersion(to);
-  // single-quote the path for the shell; a literal `'` inside it (vanishingly
-  // unlikely in a home path) is escaped via the classic '\'' close-reopen trick.
-  const q = `'${logPath.replace(/'/g, "'\\''")}'`;
+  // single-quote for the shell; a literal `'` inside (vanishingly unlikely in a
+  // home path or binary path) is escaped via the classic '\'' close-reopen trick.
+  const shq = (s: string): string => `'${s.replace(/'/g, "'\\''")}'`;
+  const q = shq(logPath);
+  // The configured herdr binary (HERDR_BIN / config.herdrBin), shell-quoted so a
+  // custom install path can't break the script. Every herdr invocation below uses
+  // it — a bare `herdr` would miss on a custom-binary host (mirrors restart.ts).
+  const h = shq(herdrBin);
   // Shepherd stays up during the update (no restart), so it captures this
   // script's stdout live for the modal. The `tee -a` is kept anyway: it makes
   // `cat <logPath>` a durable post-mortem that survives even a shepherd crash.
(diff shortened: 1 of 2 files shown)


kai-osthoff (commented):
No blocking findings.

Verified the generated failure path no longer uses the nonexistent `herdr server start`: the update rc gate is at `src/herdr-update.ts:105`, the reachability probe is `timeout 10 ${h} agent list` at `src/herdr-update.ts:106`, and the recovery launch is `setsid ${h} server ... &` at `src/herdr-update.ts:110`. The new operator-facing failure messages are present on both branches at `src/herdr-update.ts:107` and `src/herdr-update.ts:109`.

Repo-visible identifiers used by the change resolve consistently: `HerdrDriver.list()` calls `runner(["agent", "list"])` at `src/herdr.ts:225` and parses its result at `src/herdr.ts:226`; `broadcast()` treats a thrown list as herdr-unreachable/offline at `src/service.ts:3515` and `src/service.ts:3519`; startup reconcile also treats a list throw as temporary herdr failure at `src/reconcile.ts:15` and `src/reconcile.ts:18`. The configured binary path comes from `HERDR_BIN` / `config.herdrBin` at `src/config.ts:437`, is defaulted into `buildUpdateScript` at `src/herdr-update.ts:60`, quoted at `src/herdr-update.ts:66`, and used for update/probe/server invocations at `src/herdr-update.ts:103`, `src/herdr-update.ts:106`, and `src/herdr-update.ts:110`. The existing restart helper uses the same quoted-binary pattern at `src/restart.ts:29` and `src/restart.ts:38`.

The changed tests cover the runtime text shape for the handoff command, failed-update recovery branch, removal of `server start`, and custom `HERDR_BIN` threading at `test/herdr-update.test.ts:13`, `test/herdr-update.test.ts:24`, and `test/herdr-update.test.ts:40`.

Verification limitation: I could not verify the external herdr CLI contract that bare `herdr server` starts the foreground server from this repository; the local driver shells out to the configured binary at `src/herdr.ts:68` and `src/herdr.ts:69`, and the default binary name is configured at `src/config.ts:437`. I did not run external commands, builds, or tests per the review instruction.

kai-osthoff:
@scoop 🛎️✨ PR's READY, green 💚 & gorgeous 💅🚀 — merge it already?? 🥱👀🔥🙏

PR #1480, opened 2026-07-06
fix: settings gear attention dot

kai-osthoff:
## Summary

- Stop using the settings gear bubble for running or blocked session state.
- Keep the mobile gear dot only for settings-owned diagnostics attention: red for errors, yellow for warnings.
- Preserve the existing running-session gear menu behavior so the halt action remains reachable.

## Why

The gear bubble implied something inside Settings needed attention, but it was also reflecting herd/session state that already has stronger, clearer affordances elsewhere in the top bar and session list.

## Validation

- `cd ui && bun run check` (0 errors, 2 pre-existing warnings)
- `cd ui && bun run test:browser -- src/lib/components/TopBar.browser.test.ts` (72 passed)

### DIFF
--- a/ui/src/lib/components/TopBar.browser.test.ts
+++ b/ui/src/lib/components/TopBar.browser.test.ts
@@ -1121,6 +1121,74 @@ describe("TopBar — working-while-blocked counts in the working tally, not bloc
   });
 });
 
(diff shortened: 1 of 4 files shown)


kai-osthoff:
🚨✅ READY, @scoop — review button’s lonely. Do your thing 😏🔍👉

PR #1502, opened 2026-07-06
fix(ui): distinguish phone back control from queue pager

scoop:
## Problem

On the **phone** session-detail header, the back-to-herd button collapsed to a bare `‹`, identical to the adjacent needs-you pager's *previous* button (also `‹`) — two visually-identical left chevrons sitting next to the pager's `›`, i.e. `‹ ‹ ›`, doing different things (leave the session vs. page the queue). When the current session isn't in the needs-you queue the `n/total` label is hidden too, so nothing disambiguated them.

## Fix

Render the **phone** back control as the list glyph **`☰`** (back to the herd list), which the `‹ ›` pager never uses. Desktop/compact still show the text `‹ Herd`; the pager, swipe gestures, and all aria-labels are unchanged.

| | Back → herd | Queue prev | Queue next |
|---|---|---|---|
| Before (phone) | `‹` | `‹` | `›` |
| After (phone) | `☰` | `‹` | `›` |
| Desktop/compact | `‹ Herd` | `‹` | `›` |

### Glyph choice
`☰` (U+2630) over `⌂` (U+2302): `⌂` renders inconsistently (thin house / tofu) on some Android/system fonts, which no CI gate can catch; `☰` is a reliably-rendered list glyph that reads as "the herd list". The literal is a raw glyph like the existing `‹`/`›` — the translated meaning already lives in `aria-label` ("Back to herd" / "Zurück zum Herd"), so **no message-catalog change**.

## Tests
- Added a regression test in `Viewport.browser.test.ts`: the phone `.back` glyph is `☰` and **not** `‹`, guarding against silently reintroducing the collision.
- `bun run check`, `bun run check:i18n`, and `vitest --project browser Viewport.browser.test.ts` (26 passed) all green.

### DIFF
--- a/ui/src/lib/components/Viewport.browser.test.ts
+++ b/ui/src/lib/components/Viewport.browser.test.ts
@@ -900,3 +900,27 @@ describe("Viewport Activity tab A/B switch", () => {
     ).toBeNull();
   });
 });
+
+// ── Phone back control glyph (two-left-chevron regression) ────────────────────
+// On phone the .back control collapses from the desktop "‹ Herd" text to a bare
+// glyph. It must NOT be the left chevron "‹" — that made it visually identical to
+// the adjacent needs-you pager's prev button (also "‹"), yielding two ambiguous
+// left chevrons. It renders as the list glyph "☰" (back to the herd list), which
+// the pager never uses. The .back button only mounts when `onback` is supplied.
+describe("Viewport phone back glyph", () => {
+  it("phone back control renders the list glyph ☰, not the left chevron ‹", async () => {
+    const { container } = render(Viewport, {
+      session: session({ id: "back-glyph" }),
+      mobile: true,
+      onback: vi.fn(),
+      previewPort: null,
+      openPreviewTick: 0,
+    });
+
+    const back = container.querySelector(".back");
+    expect(back, ".back control should render on phone when onback is supplied").not.toBeNull();
+    const glyph = back!.textContent?.trim();
+    expect(glyph, "phone back glyph is the list glyph").toBe("☰");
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

The PR replaces the phone `.back` control's bare left chevron `"‹"` with the list glyph `"☃"`... (see below) to remove the two-adjacent-left-chevron ambiguity between the back control and the needs-you queue pager's prev button. Change is minimal and correct.

## Verification

- `ui/src/lib/components/Viewport.svelte:1892` — phone branch now renders `"☰"` (list glyph) instead of `"‹"`. The queue pager prev button is still `"‹"` at `Viewport.svelte:1900` and next is `"›"` at `:1908`, so the collision is genuinely resolved and the pager never uses `"☰"`.
- Desktop still uses `m.viewport_back_button()` = `"‹ Herd"` (verified `ui/messages/en.json:478`, `ui/messages/de.json:478`); as text this is unambiguous, so leaving the desktop chevron is fine.
- `aria-label` unchanged: `m.viewport_back_aria()` = "Back to herd" / "Zurück zum Herd" (verified `ui/messages/en.json:477`, `ui/messages/de.json:477`). The glyph is purely visual; assistive-tech semantics are preserved. No new message keys were introduced, so no i18n parity gap.
- CSS comment update at `Viewport.svelte:3224` matches the new glyph; the `.vp-head.phone .back` sizing rule still applies.
- The added test (`ui/src/lib/components/Viewport.browser.test.ts:907-926`) uses props `session`, `mobile`, `onback`, `previewPort`, `openPreviewTick`, all of which exist on the component (verified `Viewport.svelte:82,88,97,101`). The assertions (glyph is `☰`, not `‹`) are consistent with the implementation.

PR #1508, opened 2026-07-06
fix(ui): hide Codex usage in CLI picker when Codex absent

scoop:
## Problem

The **Choose coding CLI** popover (`ModelCliPicker`) rendered a **CODEX** section reading *"Codex limits not found; showing token usage."* even when no Codex was found on the machine — with no token row beneath it, since `codexUsage` was `null`. It promised token usage that didn't exist.

## Cause

The Codex block rendered whenever `codexGauges` was empty, which is true in **two** cases:
1. Codex present but logged no rate-limit windows (legitimate token-usage fallback).
2. Codex **absent** entirely (`codexUsage === null`) — no data at all.

Only the token *row* was guarded by `{#if codexUsage}`; the label + "limits not found" message were not.

## Fix

Wrap the entire Codex `.mcp-provider-usage` block in `{#if codexUsage}`:

| Codex state | Before | After |
| --- | --- | --- |
| Absent (no snapshot) | "limits not found; showing token usage." (no token row) | nothing rendered |
| Present, has rate limits | 5H / WK gauges | unchanged |
| Present, no rate limits | message + TOTAL row | unchanged |

This mirrors how the other three consumers already gate on `codexUsage` (`TopBarMobileSheet`, `LimitsLens`, `CodexTokenDetail`) — so they need no change. The token-usage fallback is intentionally preserved when Codex *is* present.

## Verification

- `cd ui && bun run check` — 0 errors
- `bunx eslint` + `bunx prettier --check` on the touched file — clean

No message keys added/removed; `topbar_codex_limits_unavailable` still used by the present-but-no-limits path. Display-correctness fix, no user-facing feature → no announcement entry.

### DIFF
--- a/ui/src/lib/components/new-task/ModelCliPicker.svelte
+++ b/ui/src/lib/components/new-task/ModelCliPicker.svelte
@@ -181,32 +181,32 @@
           <span class="mcp-unavailable">{m.upnext_usage_unavailable()}</span>
         {/if}
       </div>
-      <div class="mcp-provider-usage">
-        <span class="micro">{m.agent_provider_codex()}</span>
-        {#if codexGauges.length > 0}
-          {#each codexGauges as g (g.label)}
-            <div class="mcp-gauge">
-              <span>{g.label}</span>
-              <span class="mcp-bar">
-                <span
-                  class="mcp-fill"
-                  style="transform:scaleX({gaugeFill(g.w.pct)});background:{gaugeColor(g.w.pct)}"
-                ></span>
-              </span>
-              <span style="color:{gaugeColor(g.w.pct)}">{g.w.pct}%</span>
-              <small>{formatResetIn(g.w.resetAt, nowMs)}</small>
-            </div>
-          {/each}
-        {:else}
-          <span class="mcp-unavailable">{m.topbar_codex_limits_unavailable()}</span>
-        {/if}
-        {#if codexUsage}
+      {#if codexUsage}
+        <div class="mcp-provider-usage">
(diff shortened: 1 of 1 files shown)


scoop (commented):
## Summary

The change wraps the entire Codex `mcp-provider-usage` block (heading, gauges, "limits unavailable" fallback, and token total) in `{#if codexUsage}`, so nothing about Codex or tokens renders when Codex isn't found. This satisfies the task.

## Verification

- `codexUsage` comes from `codexTokenUsage(usageLimits)` (ui/src/lib/components/new-task/ModelCliPicker.svelte:72), which returns `null` when no `provider: "codex"` snapshot exists (ui/src/lib/components/usage-gauges.ts:111-120). So `{#if codexUsage}` is a correct proxy for "Codex is present."
- The only Codex snapshot kind is `kind: "tokens"` (ui/src/lib/types.ts:1013-1029), so there is no other Codex shape that could be missed by gating on `codexTokenUsage`.
- This brings the picker in line with the sibling helper `compactUsageViews`, which already emits a Codex view only when `codexUsage` is truthy (ui/src/lib/components/usage-gauges.ts:167).
- No new message keys are introduced; `m.agent_provider_codex`, `m.topbar_codex_limits_unavailable`, and `m.topbar_tokens_total` were all already used in this file pre-diff, so no i18n parity concern arises from this change.
- The `codexUsage`-present-but-empty-gauges path is preserved: it still shows the `topbar_codex_limits_unavailable` fallback plus the token total (ui/src/lib/components/new-task/ModelCliPicker.svelte:201-207).

PR #1494, opened 2026-07-06
fix(plan-gate): surface unavailable plan artifacts

kai-osthoff:
## Summary
- add a `plan-unavailable` plan-review trigger status for missing, empty, or unreadable `.shepherd-plan.md` artifacts
- keep existing skip precedence for non-planning, in-flight, and already-approved plan gates
- explain the required plan artifact in PlanPanel and GitRail, including demo coverage

## Verification
- `bun test test/plan-gate-service.test.ts test/server-plan-gate.test.ts`
- `bun run lint`
- `cd ui && bun run check:i18n`
- `cd ui && bun run check`
- `cd ui && bun run test:browser -- src/lib/components/PlanPanel.browser.test.ts src/lib/components/GitRail.browser.test.ts`
- `cd ui && bun run test:node -- src/lib/demo/router.test.ts src/lib/demo/state.test.ts`
- `cd extension && bun run check`
- pre-push hook passed

PR #1510, opened 2026-07-06
fix(codex): steer missing plan artifact

kai-osthoff:
## Summary
- when a Codex planning session has no usable `.shepherd-plan.md`, steer the live session to write the plan artifact and stop
- keep Claude behavior unchanged: it still only reports `plan-unavailable`
- preserve existing precedence so non-planning, in-flight, and already-approved gates are not steered

## Dependency
- Stacked on #1494, which introduces the `plan-unavailable` status and UI diagnosis.

## Verification
- `bun test test/plan-gate-service.test.ts test/server-plan-gate.test.ts`
- `bun run lint`
- pre-push hook passed


kai-osthoff:
Scoop, push him already 😏🚀🫵⚡✅ do your job!

scoop:
> Scoop, push him already 😏🚀🫵⚡✅ do your job!

@kai-osthoff merge conflict

scoop:
Closing — the useful majority of this PR **is already on `main`**, and the remainder cannot be rebased safely. Follow-up filed as #1982.

Its parent #1494 merged 2026-07-06T21:24 (about an hour after this PR opened), which landed the `plan-unavailable` status, all six EN/DE message keys, the PlanPanel/GitRail/demo wiring, and both the service and server test files. Roughly 92% of this diff is therefore a no-op against current `main`.

The residual ~30-line Codex steer in `src/plan-gate.ts` is structurally invalid there:

1. `PlanGateDeps.reply` is now `(sessionId, text) => Promise<boolean>` (`src/plan-gate.ts:493`, async since #1567). This PR does `const delivered = this.deps.reply(...)` and branches on `if (!delivered)` — a Promise is always truthy, so the warning can never fire.
2. Codex **exits** after its turn, which is why `src/resume-then-steer.ts` exists. A bare `reply` to a Codex planner lands on a dead pane; the steer has to go through `resumeThenSteer(...)`, which also forces `consider()` to become `await`ed.
3. `shouldConsiderOnSettle` (`src/plan-gate.ts:89`) returns true for `status === "done"` on any planning session, so this steers → Codex runs a turn → settles `done` → still no plan file → steers again, **unbounded**. Main's analogous refusal steer is explicitly bounded (`MAX_REFUSAL_STEERS = 2`, `src/plan-gate.ts:1648`, with a persisted counter and a comment naming this exact loop). It also re-steers on every operator "Review plan now" click (`src/server.ts:2835` passes `{force:true}`).

The `planPhase === "planning"` guard on `handlePlanUnavailable` was correct and is carried into #1982, along with the missing Codex-exit test coverage. Note also that `src/service.ts` has since gained Codex-specific plan-gate directives (stop clause + "write `.shepherd-plan.md` in the same turn"), so #1982 starts by confirming the gap still reproduces at all.

Reviewed as part of a merge train over #1926 / #1510 / #1491 / #1386.

PR #1512, opened 2026-07-06
fix: explain missing base refs on task create

kai-osthoff:
## Summary
- add a typed worktree error when the requested base ref does not resolve to a commit
- map that failure to HTTP 422 so New Task shows an actionable missing-base message
- cover the worktree error and POST /api/sessions unborn-repo path

## Tests
- bun run lint
- bun test test/worktree.test.ts test/server.test.ts
- cd extension && bun run check
- pre-push hook passed: gates, prettier, eslint, tsc, ext, root-tests, ui, fallow audit

Note: a broad ad hoc `bun test` run was not used as a pass signal because it invokes UI browser tests outside their configured browser harness; the pre-push root-tests lane and targeted backend tests passed.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -3,7 +3,7 @@ import type { PluginRegistry } from "./plugins/loader";
 import type { PluginInfo } from "./plugins/types";
 import type { SessionService } from "./service";
 import { PREVIEW_SETUP_STEER, RestoreError } from "./service";
-import { WorktreeRestoreError } from "./worktree";
+import { WorktreeMissingBaseError, WorktreeRestoreError } from "./worktree";
 import { LearningsService } from "./learnings-service";
 import { RepoConfigService } from "./repo-config-service";
 import type { CreateSessionInput } from "./types";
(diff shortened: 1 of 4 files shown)


kai-osthoff (commented):
No findings.

Verified the new `WorktreeMissingBaseError` is exported from `src/worktree.ts:56` and carries the message asserted by the tests at `src/worktree.ts:59`. `WorktreeMgr.create` still validates the ref string before git calls at `src/worktree.ts:79`, still returns a shared checkout for non-git directories at `src/worktree.ts:82`, and now checks commit resolution before `git worktree add` at `src/worktree.ts:85`; the resolver uses `git rev-parse --verify <ref>^{commit}` and returns false on failure at `src/worktree.ts:105`.

Verified the session create path resolves a base ref via `ensureBaseRef` and passes that value to `worktree.create` at `src/service.ts:2438` and `src/service.ts:2620`, so the new typed error is reachable from the POST `/api/sessions` flow. The route catches `service.create` failures at `src/server.ts:1960` and maps `WorktreeMissingBaseError` to HTTP 422 at `src/server.ts:1899`; the import spelling matches the exported class at `src/server.ts:6` and `src/worktree.ts:56`.

Verified the added coverage exercises both layers: `test/worktree.test.ts:392` asserts `WorktreeMgr.create` throws `WorktreeMissingBaseError`, and `test/server.test.ts:266` creates an unborn git repo, posts `baseBranch: "main"`, and asserts a 422 plus the actionable error fragments at `test/server.test.ts:293`. The added test imports resolve to in-repo exports at `test/server.test.ts:4`, `test/server.test.ts:10`, and `test/worktree.test.ts:6`.

Could not execute the test suite by instruction; only read-only inspection and diff checks were performed.

kai-osthoff:
READY ✅🚀 Scoop, push it already 🫠👉📦🔥

PR #1516, opened 2026-07-06
fix(ui): compact Up Next sort into icon menu, inline refresh

scoop:
## Why

The Up Next sort control was a five-segment strip (`Recommended / Newest / Oldest / Title A-Z / Title Z-A`) forced to a 320px minimum. At the real ~360px dock width it **wrapped onto its own line and the labels ellipsized** — it "looked like crap." The refresh `⟳` was stranded at the far right.

## What

- **Sort → `⇅` icon + anchored menu.** New `UpNextSortMenu.svelte`: a single-select `role="listbox"` (`✓` on the active mode) that mirrors the repo's existing `TaskIdMenu` mechanics — `use:portal` + JS anchoring under the trigger (so the panel's `overflow:auto` can't clip it), flip-up when there's no room, roving arrow-key nav, focus-restore, and dismiss on **Esc / outside-click / scroll**. No scrim (design-system anchored-popover exemption).
- **Refresh → inline** beside the "updated … ago" time; the sort trigger is right-anchored (`.un-sortwrap { margin-left:auto }`). The header now fits on one line at 360px.
- Trigger `aria-label`/`title` read "Sort by {current}", so the active sort stays discoverable despite the icon.
- All sort logic (`SORT_MODES`, `SORT_LABELS`, `setSortMode` + its `localStorage` persistence) is reused unchanged.

## ⚠️ Reviewer-confirmable waiver — tap-target size

The sort `⇅` and refresh `⟳` buttons are **~30px, below the 44×44px tap-target floor**. This is deliberate: every other control in the Up Next panel (rows, start buttons, expand) is compact at this density, and raising only these two to 44px would look inconsistent and bloat the header. Please confirm this waiver is acceptable.

## Notes

- Fallow flags the menu's dismiss/nav block as duplicated — this is intentional: it matches the existing `TaskIdMenu`/`CardMenu`/`SteerMenu`/`PrBadgeMenu` family, the accepted pattern I cloned. Warn-level, non-blocking.
- Not a new capability (a refinement of an existing control) → committed as `fix`, no feature-announcement entry.

## Verification

- `bun run check` (svelte-check) — 0 errors
- `bun run check:i18n` — locales in parity (added `upnext_sort_by` EN+DE)
- eslint + prettier clean
- `UpNextPanel.browser.test.ts` — 23/23 pass (updated to open the menu before picking a sort mode; exercises all modes in a real browser)

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2210,6 +2210,7 @@
   "upnext_priority_section": "Priorität",
   "upnext_normal_section": "Warteschlange",
   "upnext_sort_aria": "Als Nächstes sortieren",
+  "upnext_sort_by": "Sortieren nach {mode}",
   "upnext_sort_recommended": "Empfohlen",
   "upnext_sort_newest": "Neueste",
   "upnext_sort_oldest": "Älteste",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2210,6 +2210,7 @@
   "upnext_priority_section": "Priority",
   "upnext_normal_section": "Queued",
   "upnext_sort_aria": "Sort Up Next",
+  "upnext_sort_by": "Sort by {mode}",
   "upnext_sort_recommended": "Recommended",
   "upnext_sort_newest": "Newest",
   "upnext_sort_oldest": "Oldest",
--- a/ui/src/lib/components/UpNextPanel.browser.test.ts
+++ b/ui/src/lib/components/UpNextPanel.browser.test.ts
@@ -159,6 +159,11 @@ const providerSelect = () => document.querySelector<HTMLSelectElement>("#mcp-pro
 const pickerConfirm = () => document.querySelector<HTMLButtonElement>(".mcp-actions .primary")!;
 const rowNumbers = () =>
   Array.from(document.querySelectorAll(".un-row .un-num")).map((el) => el.textContent ?? "");
+// Sort now lives behind a ⇅ icon that opens a listbox menu; open it, then pick the mode.
+const pickSort = async (name: string) => {
+  document.querySelector<HTMLButtonElement>(".un-sortbtn")!.click();
+  await page.getByRole("option", { name }).click();
+};
 
 let fontStyle: HTMLStyleElement;
 beforeEach(() => {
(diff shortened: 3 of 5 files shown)


scoop (commented):
## Summary

The PR replaces the wrapped row of segmented sort buttons with a compact `⇅` icon trigger that opens an anchored, portaled listbox menu (`UpNextSortMenu.svelte`), and leaves the refresh `⟳` button beside the "updated ago" time. This matches the task ("sort buttons look like crap" → compact icon menu; "move refresh next to last refresh time to save space"). The DOM order in `UpNextPanel.svelte:413-436` is `updated` span → `un-refresh` → `un-sortwrap`, and `.un-sortwrap { margin-left: auto }` (UpNextPanel.svelte:569-572) pushes only the sort trigger to the right edge, keeping refresh adjacent to the time as requested.

## Verification performed

- **New component is a faithful clone of an established pattern.** `UpNextSortMenu.svelte` mirrors `TaskIdMenu.svelte` for positioning/dismiss (verified against src/lib/components/TaskIdMenu.svelte:26-91 — same flip-up logic, capture-phase pointerdown excluding the opener, scroll/Esc dismiss, focus-restore-to-opener-when-focus-fell-to-body via queueMicrotask) and `LanguageSwitcher.svelte` for listbox semantics (verified src/lib/components/LanguageSwitcher.svelte:24-30 — same `aria-haspopup="listbox"`/`aria-expanded` trigger). `use:portal` resolves to the real action at src/lib/portal.ts:13.
- **Sort identifiers resolve.** `SORT_MODES`, `SORT_LABELS`, `validSortMode`, `setSortMode`, `sortMode` all exist and are used consistently (UpNextPanel.svelte:72-95); the new `sortOptions`/`selectSort`/`toggleSortMenu` (UpNextPanel.svelte:104-124) feed `current`/`options`/`onselect` correctly. `selectSort` funnels through `validSortMode`, and option values come from `SORT_MODES`, so they always validate.
- **Active-option focus uses `data-value`.** `opts().find((b) => b.dataset.value === current)` (UpNextSortMenu.svelte) matches the rendered `data-value={opt.value}` attribute — verified the button markup carries `data-value`.
- **i18n parity holds.** New key `upnext_sort_by` exists in BOTH catalogs with the same `{mode}` placeholder — en (messages/en.json:2213 `"Sort by {mode}"`) and de (messages/de.json:2213 `"Sortieren nach {mode}"`); called as `m.upnext_sort_by({ mode: ... })` at UpNextPanel.svelte:432-433. Only en+de catalogs exist (messages/*.json). `upnext_sort_aria` (reused as the listbox label) is present in both (messages/en.json:2212, messages/de.json:2212).
- **No stale references.** No remaining `.un-sort` / `.un-sort-btn` selectors anywhere after removal (grep clean across src). The new `.un-sortbtn`/`.un-sortwrap`/`.un-sortmenu`/`.usm-item` classes appear only in the two touched components plus the test.
- **Tests updated coherently.** The `pickSort` helper (UpNextPanel.browser.test.ts:163-165) opens the menu then `getByRole("option", { name })`; the option's accessible name is the label alone since the `✓`/empty check span is `aria-hidden` — so `m.upnext_sort_recommended()` etc. still match. Each `pickSort` reopens the menu (menu closes on select via `selectSort`), so the chained calls at lines 277-281 are correct.

- The listbox uses roving `tabindex` + `.focus()` rather than `aria-activedescendant`, and its options are `<button role="option">`. This is intentional and consistent with the repo's existing `LanguageSwitcher`/`TaskIdMenu` menus, so it is not flagged.
- Sub-44px (30px) tap targets are a deliberate, comment-documented waiver (UpNextPanel.svelte:566-568).

Nothing blocking found.

PR #1521, opened 2026-07-06
feat(repo-switcher): add "add to filter" to repo chip menu

scoop:
## What

The repo-chip long-press / right-click menu previously offered only **Pin/Unpin**. On desktop you can **Shift+click** a chip to additively toggle that repo into the herd repo filter — but touch users (and anyone unaware of the Shift shortcut) had no discoverable way to do it.

This adds a second menu item that performs the **same additive-toggle**, labelled by state:
- **"Add to filter"** (⊕) when the repo isn't in the current filter
- **"Remove from filter"** (⊖) when it is

## How

The item reuses the existing plumbing verbatim — it calls `onrepofilter(repoPath, /*additive*/ true)`, the exact path `onChipClick` already uses for Shift+click (→ `applyRepoFilter` → `nextRepoFilter`). No new props or page wiring.

- Menu keeps **Pin/Unpin first**, appends the filter item second.
- Added roving keyboard nav (Up/Down/Home/End, with `preventDefault`) so keyboard users can reach the second item; Escape still closes.
- Menu items now meet the **44px tap-target floor** on coarse pointers (the menu is touch-reachable via long-press).
- New EN + DE message keys (real German translations) and a `v1.43.0` What's-New entry.

## Empty-filter wording

With an empty filter, the item reads "Add to filter" and activating it creates a single-repo filter (`nextRepoFilter(∅, repo, true) → {repo}`). Wording is accurate — you add the repo to the (empty) filter — and mirrors Shift+click exactly.

## Tests

`RepoSwitcher.browser.test.ts` — new tests cover: filter action calls `onrepofilter(repoPath, true)` + closes menu; label flips to "Remove from filter" when already filtered; roving keyboard nav across the two items with Escape still closing.

Verified locally: `bun run check` (0 errors), `bun run check:i18n` (parity), `bun test` (1041 passed).


scoop (commented):
## Summary

This PR adds an "Add to filter" / "Remove from filter" item to the repo-chip long-press/right-click menu, giving touch users the additive-filter gesture that desktop reaches via Shift+click. The implementation is correct and matches existing patterns.

## Verification

- **Additive-toggle semantics match Shift+click.** `commitFilter` calls `onrepofilter(menu.chip.repoPath, true)` (RepoSwitcher.svelte:239), the same `additive: true` path as the Shift branch in `onChipClick` (RepoSwitcher.svelte:217). `nextRepoFilter` with `additive=true` toggles the repo in/out of the set (queue-strip.ts:94-99), so the label flip `menuFiltered ? remove : add` (RepoSwitcher.svelte:395-397, driven by `repoFilter.has(...)` at :380) correctly describes what the click will do in both states.
- **Message keys resolve and have locale parity.** `repo_chip_add_filter`, `repo_chip_remove_filter`, `feat_repo_chip_add_filter_title`, `feat_repo_chip_add_filter_body` exist in en.json:2431-2434 and de.json:2431-2434 — verified both catalogs carry all four.
- **Feature announcement auto-registers.** Entries are eager-glob-imported (feature-announcements.ts:47-49), so the new `v1.43.0-repo-chip-add-filter.ts` file needs no index edit; its shape (id/sinceVersion/titleKey/bodyKey, no targetId) matches the sibling `v1.39.0-repo-chip-pin.ts` and its `titleKey`/`bodyKey` point at the keys confirmed above. Version prefix `1.43.0` is consistent with other v1.43.0 entries.
- **Roving focus is sound.** The keydown handler resolves items via `menuEl.querySelectorAll(".rs-menu-item")` (all are `<button>`s), guards `items.length === 0`, wraps with modulo, and only `preventDefault`s when it actually moves focus (RepoSwitcher.svelte:267-280). Escape still closes early (:261-264). The open-effect focuses the first item (:254), matching the test's expectation.
- **Menu-item click isn't swallowed** by the capture-phase `pointerdown` outside-close listener — the target is inside `menuEl`, so the menu stays open until `commitFilter` closes it (:282-284).
- **Coarse-pointer tap target** added for `.rs-menu-item` (min-height 44px) mirrors the existing `.rs-chip` rule.

Issue #1520, opened 2026-07-06
Restyle GitRail status controls as compact pill chips


scoop:
## Context

Follow-up from the "single-tap toggles the Git rail" work (the double-tap-title-renames branch/PR). While mocking up that interaction, the Git rail contents were sketched as a row of compact, rounded **pill chips** — PR `#…`, CI `passing`, critic `clear`, ready `✓`, merge — with semantic accent hues (blue for PR, green for CI/ready). The operator liked that treatment and wants the real `GitRail.svelte` controls revisited to move toward it.

## Ask

Revisit the visual language of `ui/src/lib/components/GitRail.svelte` (PR / CI / merge / critic / ready / verdict controls) to read as a consistent set of compact pill chips, per the mockup.

## Scope notes

- Design-token only — reuse the existing `--color-*` / `--status-*` surfaces and `--fs-*` scale; no raw hex/rgba/px (per the design-system directive in CLAUDE.md). Consult `/design-system` first and reuse an existing chip/badge recipe rather than growing a new one.
- Purely presentational; no behavior change to the rail's actions.
- Keep i18n intact — all labels stay routed through `m.*`.
- Not blocking the tap-toggle change; that PR only relocates the toggle and rewires the title tap.

_Filed from an attended planning session as an explicitly-requested follow-up._

scoop:
Sequencing decided (attended planning): this restyle ships the **full-original** look (6px · full accent · blue PR · labelled CI · red failing chip) but is **blocked by #1537** — a design-rules pass that sanctions the rounded, semantically-accented chip language in DESIGN.md + /design-system first. Once #1537 merges, #1520 is a clean compliant restyle with no per-PR carve-out. Plan staged in `.shepherd-plan.md`.

scoop:
Update: sequencing chose the **uniform 6px** look (whole rail incl. MERGE as chips). The merged rules (#1539) only sanction 6px for read-only status; interactive controls stay 2px. So this is now blocked on **#1540** (permit 6px on interactive controls within a chip row). Once #1540 merges, #1520 implements uniform-6px with no per-PR carve-out.

PR #1522, opened 2026-07-06
feat(viewport): single-tap title toggles git rail

scoop:
## What

On the focused-session header (desktop), a **single tap on the session title now collapses/expands the Git rail**, and the tiny toggle chip moves from the far-right actions cluster to sit **beside the title** — enlarging the tap target and pairing it with the title. **Double-tap still renames.**

Addresses the ask: the standalone toggle button was tiny and easy to miss; it's now welded to the title and the title itself is part of the tap target.

## Interaction model

- **Single tap** on the title → toggles the rail **instantly** (the common action stays snappy; no delay).
- **Double-tap** → renames, and *undoes* the first tap's toggle so the rail returns to its pre-rename state. A brief flash can only appear on the rare rename path, never on the common toggle.
- Synchronous — **no timer**, so nothing fires after the user moves on.
- The relocated chip stays an instant, `aria-expanded` button (the keyboard/AT toggle path); `.desig` keeps Enter/Space → rename.
- **Desktop only.** `gitOpen` and the chip are already desktop-only; the toggle branch in `onTitleTap` is `!compact`-guarded, so mobile/compact keeps its existing first-tap-focus / double-tap-rename and header-fold behavior unchanged.

## Why not a delay

An earlier draft used a 500 ms delayed toggle to avoid any flash, but "no flash" was never a requirement and a half-second lag on the primary toggle is a real regression. Instant-toggle-with-undo keeps toggling immediate and confines the only flash to renames.

## Tests

Added to `Viewport.browser.test.ts`: single-tap toggles the rail instantly; double-tap renames **and** restores the rail to closed; the relocated chip toggles with `aria-expanded`; mobile `.ctx-trigger` double-tap still renames. Full `Viewport.browser.test.ts` suite green (31/31); `bun run check` clean; `check:i18n` in parity.

## Feature discovery

Adds `v1.43.0-git-rail-title-tap` announcement entry + EN/DE keys.

## Follow-up

Restyling the rail *contents* as compact pill chips is tracked separately in #1520 (out of scope here — this PR only relocates the toggle and rewires the title tap).

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2239,6 +2239,8 @@
   "feat_upnext_body": "Eine neue oberste Ansicht reiht noch nicht begonnene GitHub-Issues über all deine Repos in eine flache Warteschlange — Priorität zuerst. Sieh das Lohnenswerteste zum Starten und starte eine Session dafür — oder mehrere — mit einem Klick.",
   "feat_up_next_sort_title": "Als Nächstes frei sortieren",
   "feat_up_next_sort_body": "Als Nächstes kann jetzt zwischen Empfehlung, neueste, älteste und Titel A-Z/Z-A wechseln. Priorisierte Arbeit bleibt zuerst gruppiert, und deine Sortierung wird gemerkt.",
+  "feat_git_rail_title_tap_title": "Titel antippen öffnet die PR-Aktionen",
+  "feat_git_rail_title_tap_body": "Auf dem Desktop öffnet oder schließt ein einfacher Klick auf den Sitzungstitel jetzt die PR-, Merge- und Kritik-Aktionen. Ein Doppelklick benennt die Sitzung weiterhin um, und der Umschalter sitzt jetzt direkt neben dem Titel als größeres, leichter treffbares Ziel.",
   "feat_upnext_skip_cli_picker_title": "Coding-CLI-Auswahl in „Als Nächstes“ überspringen",
   "feat_upnext_skip_cli_picker_body": "Eine neue Opt-in-Einstellung (Einstellungen → Coding CLI) startet den Schnellstart in „Als Nächstes“ sofort mit deiner Standard-Coding-CLI, statt jedes Mal nachzufragen, welche CLI verwendet werden soll.",
   "feat_add_repo_title": "Repo aus dem Backlog hinzufügen",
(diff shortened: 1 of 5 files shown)


scoop (commented):
## Summary

The PR makes a single tap on the session title toggle the git rail (desktop only), keeps double-tap-to-rename, and relocates the `.git-toggle` chip to sit beside the title as a larger, discoverable target — exactly what the task asked for. The implementation is correct and well-guarded. No blocking issues.

## What I verified

- **Toggle / double-tap logic** (`ui/src/lib/components/Viewport.svelte:896-908`): first tap sets `lastTitleTap` and flips `gitOpen`; a second tap inside the 500ms window flips it again (net-zero, restoring pre-rename state) then renames. Toggle-twice is correct regardless of the starting value. All three `gitOpen` mutations are gated by `!compact`, so the compact/mobile `.ctx-trigger` path (`Viewport.svelte:2169-2185`) keeps its unchanged first-tap-focus / double-tap-rename behavior. `compact = mobile || touch` confirmed at `Viewport.svelte:372`.
- **Button relocation**: exactly one `.git-toggle` remains in the template (grep count = 1), moved out of the post-`PlanGateBadge` cluster into the `{#if !compact}` block right after the title/branch (`Viewport.svelte:2234-2252`). CSS is `display: inline-flex; flex-shrink: 0` with no absolute positioning or auto margins (`Viewport.svelte:2901-2920`), so it flows correctly in the new slot; the `spacer` div still pushes the tab bar right.
- **Moved button message keys** pre-exist and are unchanged: `viewport_git_actions*` at `ui/messages/en.json:514-516`.
- **New feature-announcement keys** have en/de parity: `feat_git_rail_title_tap_title` and `feat_git_rail_title_tap_body` present in both `ui/messages/en.json:2242-2243` and `ui/messages/de.json:2242-2243` (grep count = 2 in each catalog).
- **Feature entry** `ui/src/lib/feature-announcements/entries/v1.43.0-git-rail-title-tap.ts` is auto-registered via `import.meta.glob` (`ui/src/lib/feature-announcements.ts:46`); `sinceVersion: "1.43.0"` matches the current unreleased line (sibling `v1.43.0-*` entries exist) and the filename/id (`git-rail-title-tap`) are internally consistent with the `FeatureAnnouncement` contract.
- **Tests** (`ui/src/lib/components/Viewport.browser.test.ts:388-457`) cover desktop single-tap toggle, double-tap rename + rail restore, the relocated chip's `aria-expanded`, and the mobile-still-renames path, all against real rendered selectors.

## Non-blocking observations (no change required)

- Because the toggle is now wired to both `.desig` and `.vp-name`, a plain single click intended only to reveal the meta popover will also open/close the rail. This is the documented intent of the feature (the What's-New body states as much), not a defect.
- The shared `lastTitleTap` timestamp is not reset when the user clicks the dedicated `.git-toggle` button between two title taps, so a title-tap / button-click / title-tap sequence within 500ms could read the two title taps as a double-tap and trigger a rename. This is an inherent property of the pre-existing shared-timestamp double-tap design and an unlikely interleaving; not worth blocking.

PR #1523, opened 2026-07-07
fix(ui): keep Up Next header on one line (no sort ⇅ wrap)

scoop:
## Problem

On the Up Next lens header, the sort ⇅ button wrapped to its own second line at mobile widths (reported on **Vivaldi Android** — see screenshot in the issue). Line 1 showed `UP NEXT · updated 0s ago · ⟳`; the sort control dropped below, right-aligned.

## Root cause (empirically confirmed)

`.un-head` used `flex-wrap: wrap` with `margin-left: auto` on the sort control. The whole app renders in `--font-mono`; on Android "Berkeley Mono"/"JetBrains Mono" aren't installed, so text falls back to the wider system **monospace**. Under Pixel-5 emulation with monospace, the header's single-line threshold is ~282px (EN) / ~342px (DE) — a thin margin at a ~360px mobile panel, crossed by device font width / the DE locale / a longer "updated Nm ago" string. Once content exceeds the width, the auto-margin item (sort) wraps rather than the text shrinking. (A sans-font repro never wraps — the fix targets the wrap *mechanism*, not just width.)

Note: the header has no "filter" control; the issue's "filter/refresh" wording refers to the sort ⇅ menu.

## Fix

Scoped CSS in `UpNextPanel.svelte`, no markup/logic change:

- `.un-head`: drop `flex-wrap: wrap` (→ `nowrap`).
- `.un-title-h` **and** `.un-updated`: `min-width: 0` + `overflow/ellipsis/nowrap` — the text is the shrink valve, so a too-narrow panel truncates text instead of wrapping a button. Both spans shrink so a valve exists even in the loading state (where `.un-updated` is absent).
- `.un-refresh` / `.un-sortbtn` / `.un-sortwrap`: `flex: none` — icons never shrink/grow; `margin-left: auto` still pins sort to the right edge on the single line.

## Verification

Font-agnostic regression test added to `UpNextPanel.browser.test.ts` (real Blink harness), red before the fix, green after:

- asserts computed `flex-wrap === 'nowrap'` (mechanism guard);
- A/B header height at a wide panel (600px) vs a forced-narrow one (140px) — equal = single row, no wrap-induced growth. No px/font-metric assumptions;
- covers both the snapshot **and** loading states.

`bun run check` (0 errors), `check:i18n` (parity), full `UpNextPanel` browser suite (25/25), root `bun run lint`, prettier, and the pre-push hygiene/fallow gates all pass.

## Accessibility

The refresh/sort icons remain the panel's pre-existing compact ~30px controls (deliberate sub-44px tap targets in this dense chrome — unchanged by this PR). Flagging for reviewer confirmation of the existing waiver.

### DIFF
--- a/ui/src/lib/components/UpNextPanel.browser.test.ts
+++ b/ui/src/lib/components/UpNextPanel.browser.test.ts
@@ -480,6 +480,43 @@ describe("UpNextPanel provider picker", () => {
   });
 });
 
+describe("UpNextPanel header layout", () => {
+  const headEl = () => document.querySelector<HTMLElement>(".un-head")!;
+  const panelEl = () => document.querySelector<HTMLElement>(".upnext")!;
+  // Force a reflow at a given panel width so offsetHeight reflects the new layout.
+  const setWidth = (w: number) => {
+    panelEl().style.width = `${w}px`;
+    void headEl().offsetHeight;
+  };
+
+  // Font-agnostic guards: the header must never wrap a control to a second line.
+  // We assert the mechanism (flex-wrap: nowrap) directly, then prove behaviorally that
+  // the header keeps a single row's height even when the panel is forced narrower than
+  // its content — a wrap would add a row and grow the height. No px thresholds or
+  // monospace-metric assumptions, so it trips on the wrap regardless of the CI font.
+  it("keeps the header on a single row when the panel is narrow (with a snapshot)", async () => {
+    render(UpNextPanel, {});
+    await expect.element(page.getByText(m.upnext_title())).toBeInTheDocument();
+    expect(getComputedStyle(headEl()).flexWrap).toBe("nowrap");
+    setWidth(600);
+    const wide = headEl().offsetHeight;
+    setWidth(140);
+    expect(headEl().offsetHeight).toBe(wide);
+  });
+
+  it("keeps the header single-row in the loading state (no updated-time span)", async () => {
+    upNext.snapshot = null;
+    render(UpNextPanel, {});
+    await expect.element(page.getByText(m.common_loading())).toBeInTheDocument();
+    // The shrink valve must still exist without .un-updated present.
+    expect(document.querySelector(".un-updated")).toBeNull();
+    setWidth(600);
+    const wide = headEl().offsetHeight;
+    setWidth(140);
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

The PR fixes "Up next still wraps filter/refresh" by making the panel header a strict single row.

### Verification

- `.un-head` no longer sets `flex-wrap: wrap`, so the flex default `nowrap` applies (verified `ui/src/lib/components/UpNextPanel.svelte:553-559`). This is exactly what stops the sort ⇅ / refresh ⟳ controls from dropping to a second line at narrow widths — it satisfies the task.
- The text spans become the shrink valve: `.un-title-h` and `.un-updated` gain `min-width: 0` + `overflow: hidden` + `text-overflow: ellipsis` + `white-space: nowrap` (verified `ui/src/lib/components/UpNextPanel.svelte:560-578`). With `nowrap`, a too-narrow panel truncates the text instead of overflowing/wrapping a button.
- The controls are pinned: `.un-refresh`, `.un-sortbtn`, and `.un-sortwrap` get `flex: none` (verified `ui/src/lib/components/UpNextPanel.svelte:583, 589`). `margin-left: auto` on `.un-sortwrap` still pushes it to the right edge — it composes fine with `flex: none`.
- Loading-state edge case is handled: `.un-updated` is conditional (`ui/src/lib/components/UpNextPanel.svelte:413-415`), but `.un-title-h` is also a valve, so a shrink target exists even when the updated-ago span is absent. Covered by the new test `ui/src/lib/components/UpNextPanel.browser.test.ts:508-518`.
- Identifiers used in the new tests resolve in the component: `m.upnext_title` (`ui/src/lib/components/UpNextPanel.svelte:410,412`) and `m.common_loading` (`ui/src/lib/components/UpNextPanel.svelte:470`).
- The new tests assert the mechanism (`flexWrap === "nowrap"`) and behavior (header `offsetHeight` invariant between 600px and 140px panel widths), which correctly trips on any regression to wrapping without relying on font metrics.

No correctness, security, i18n-parity, or quality problems found. The truncation of the "UP NEXT" title at extreme narrow widths is a documented, intentional tradeoff (comment at `ui/src/lib/components/UpNextPanel.svelte:547-552`), not a defect.

PR #1525, opened 2026-07-07
feat(viewport): single-tap title folds header on touch

scoop:
## What

Follow-up to #1522. That PR made a **single tap on the focused session title toggle the git rail**, but gated it to desktop (`!compact`). On the **touch/compact** layout (phones + touch devices on the desktop layout, e.g. an iPad), the title tap did nothing toward the fold — the only way to collapse/expand the header was the tiny `▾`/`▴` chevron chip.

Now, on compact, a **single tap on the title toggles the header fold instantly** (the compact analogue of the desktop git rail — there is no `gitOpen` on compact; the rail lives inside the fold). **Double-tap still renames.**

## Interaction model (mirrors #1522)

- **Single tap** on the title → `toggleFold()` instantly (collapse ⇄ expand).
- **Double-tap** → renames, and *undoes* the first tap's fold so the header returns to its pre-tap state — restoring **both** `headerCollapsed` **and** the visible `tab` (a snapshot in `preFoldTab`, because `toggleFold()` isn't a pure involution: it forces `tab = "term"` on collapse). The only brief flash is on the rare rename path.
- Synchronous — **no timer**.
- **Desktop unchanged** — the `!compact` branch still toggles `gitOpen`.
- The `▾`/`▴` chevron fold button stays as the explicit affordance + keyboard/AT toggle path (`aria-expanded` / `aria-controls`); `Enter`/`Space` on the title still renames. Title tap is purely additive.

## Tests

Added to `Viewport.browser.test.ts` (`Viewport rename affordances` block, with a `beforeEach` reset of the persisted `shepherd-vp-header-collapsed` key so a folding test can't leak into later mounts):
- touch-desktop single-tap on `.vp-name` folds (`.vp-fold` `aria-expanded` `"true"`→`"false"`);
- touch-desktop double-tap renames **and** restores the fold (`aria-expanded="true"`), asserting the attribute value (the `.vp-fold` cluster is `display:none` while renaming);
- touch-desktop double-tap from a **non-terminal tab** restores that tab (`tab = preFoldTab`) — the case that distinguishes the undo from a naive double-`toggleFold()`;
- phone single-tap on `.ctx-trigger` folds.

`bunx vitest run Viewport.browser.test.ts` → 35/35. `bun run check` clean, `check:i18n` in parity, feature-catalog / announcement-versions / branch-hygiene / glossary gates all green; ESLint/Prettier clean.

## Feature discovery

No duplicate card: #1522's `v1.43.0-git-rail-title-tap` entry hasn't shipped yet (both land in the unreleased 1.43.0), so this **broadens the existing entry** — its `titleKey`/`bodyKey` renamed to scope-neutral `feat_title_tap_toggle_*` with copy covering desktop rail **and** touch/compact fold (EN+DE). Editing the in-range `.ts` fragment also satisfies the feature-catalog gate; `id`/`sinceVersion` unchanged.


scoop (commented):
## Summary

Extends the single-tap-on-title gesture so that on the compact layout (mobile + touch-desktop) a lone tap now folds/unfolds the header (reclaiming terminal space), while desktop keeps toggling the git rail. Double-tap still renames and restores the pre-tap state. Message copy + feature-announcement entry are re-themed accordingly. Implementation satisfies the task and I found no bugs.

## What I verified

- **onTitleTap logic** (`ui/src/lib/components/Viewport.svelte:905-926`): single tap on compact snapshots `preFoldTab = tab` then calls `toggleFold()`, which forces `tab="term"` on collapse (`ui/src/lib/components/Viewport.svelte:392-398`). The double-tap undo flips `headerCollapsed` directly and restores `tab = preFoldTab` instead of re-calling `toggleFold()` — correctly sidestepping the non-involutive tab-stranding the inline comment describes. Symmetric for both start-unfolded and start-folded cases; non-compact path (`gitOpen`) unchanged.
- **Fold still leaves the title clickable**: folding only hides secondary chrome (`ui/src/lib/components/Viewport.svelte:224-226`, `392-397`), so the identity line / `.ctx-trigger` / `.vp-name` remain present for the second tap of a double-tap. Verified triggers are wired: `.ctx-trigger` at `ui/src/lib/components/Viewport.svelte:2188-2192` (phone), `.vp-name` at `ui/src/lib/components/Viewport.svelte:2233` (touch-desktop).
- **i18n keys**: new `feat_title_tap_toggle_title`/`feat_title_tap_toggle_body` present in BOTH catalogs (verified `ui/messages/en.json:2242-2243` and `ui/messages/de.json:2242-2243`); old `feat_git_rail_title_tap_*` keys fully removed with no remaining references (grep of `ui/` returns only the new keys). Keys are resolved dynamically via `m[entry.titleKey]()` (`ui/src/lib/components/WhatsNew.svelte:122-126`), and the entry file now points at the new keys (`ui/src/lib/feature-announcements/entries/v1.43.0-git-rail-title-tap.ts:6-7`).
- **Announcement contract**: the entry `id` (`git-rail-title-tap`) and filename version (`v1.43.0-…`) are unchanged; the catalog contract binds the filename to `sinceVersion`+`id` (`ui/src/lib/feature-announcements.ts:18-22`), neither of which this PR alters, so the copy retheme does not disturb the version/dup-id gates.
- **localStorage reset in tests**: `beforeEach` clears `shepherd-vp-header-collapsed` (`ui/src/lib/components/Viewport.browser.test.ts:324-329`), matching the persistence key at `ui/src/lib/components/Viewport.svelte:227`, preventing fold leakage between the new fold/unfold tests.

No blocking issues.

PR #1513, opened 2026-07-06
chore(ui): add ASCII update flock overlay

kai-osthoff:
## Summary
- Add a reusable decorative ASCII flock overlay component for update busy states.
- Mount it in `UpdateModal` only while an app update is busy, with desktop backdrop layering and mobile in-sheet layering.
- Cover desktop visibility, mobile scroll persistence, reduced-motion state, and overflow/clickability in the focused browser test.

## Verification
- `cd ui && bun run test:browser -- UpdateModal.browser.test.ts`
- `cd ui && bun run check` (passes with two existing warnings in `CommandBar.svelte` and `FilesPanel.svelte`)
- `cd ui && bun run check:i18n`
- Attempted `cd ui && bun test`; it reported broad unrelated runner/environment failures (`$state`/`document` unavailable, hoisted mock import issues) and then hung, so I killed the stuck `bun test` process.

No manual operator steps.

[no-feature-entry]

PR #1527, opened 2026-07-07
fix(ui): redo update-flock ASCII art (consistent sheep + real dog)

scoop:
## Why

The update-flock overlay behind the busy `UpdateModal` (added in #1513) was drawn with **seven inconsistent, arbitrary sheep shapes** and a **"dog" (`/^..^\`) that read as a cat** — it looked awful. The concept (a herd + a herding dog drifting past while an update applies) is charming and on-brand for a monospace terminal UI, so this keeps the concept and the well-tested machinery, and **redoes only the art**.

## What

- **One canonical sheep glyph, reused for all seven sheep**, size-scaled via the existing `small`/`big`/`tiny` classes — a front-facing fluffy sheep (`(@@@)` woolly crown, `(o.o)` face, body, `|| ||` four legs) that reads clearly even at 10–13px and low opacity.
- **A proper dog** — the classic side-profile `(___()'\`;` sheepdog — replacing the cat face.
- **Whole flock drifts left→right** so the right-facing dog trots forward instead of moon-walking; front-facing sheep are direction-neutral. (Dropped the one inert `reverse` actor.)
- **Retuned lanes + reduced-motion static positions** for the now-uniform 4-line sprite height (no top-clipping on short viewports).
- **Tiny sprite gets a translate-only wiggle** (no rotation) so it stays crisp at `--fs-micro`.
- Updated the four pinned glyph assertions in `UpdateModal.browser.test.ts` (`(o.o)` for sheep, `(___()` for dog — each a single-line signature unique to its actor).

## Verification

- `bun run check` → 0 errors (2 pre-existing warnings in unrelated files).
- `bun run test:browser` → `UpdateModal.browser.test.ts` 5/5 pass (still asserts PRE-not-SVG, placement, reduced-motion, counts).
- Rendered the busy modal at desktop + mobile and enlarged the static (reduced-motion) layout via throwaway screenshot specs — confirmed the sheep read as sheep and the dog as a dog — then removed the scaffolding.

## Notes

- Stays ASCII (`<pre>`), never SVG (a test enforces this; the earlier SVG detour in #1513 was reverted).
- Art is `aria-hidden` decorative, not user-facing text → exempt from i18n / feature-catalog / glossary, hence `[no-feature-entry]`.

### DIFF
--- a/ui/src/lib/components/AsciiFlockOverlay.svelte
+++ b/ui/src/lib/components/AsciiFlockOverlay.svelte
@@ -9,15 +9,21 @@
     reducedMotion = window.matchMedia?.("(prefers-reduced-motion: reduce)").matches ?? false;
   });
 
+  // Front-facing fluffy sheep (direction-neutral) + a side-profile herding dog
+  // that faces right, so the whole flock drifts left→right (the dog trots
+  // forward rather than moon-walking). One sheep silhouette, reused + size-scaled.
+  const SHEEP = " (@@@)\n (o.o)\n (   )\n || ||";
+  const DOG = " ___\n(___()'`;\n/,    /`\n\\\"--\\\\";
+
   const actors = [
     {
       id: "sheep-1",
       type: "sheep",
       kind: "light small",
-      art: "  __\n (oo)\\\n /||\\",
-      lane: 18,
-      start: 108,
-      distance: -136,
+      art: SHEEP,
+      lane: 20,
+      start: -12,
+      distance: 140,
       delay: "-1s",
       duration: "11s",
     },
@@ -25,10 +31,10 @@
       id: "sheep-2",
       type: "sheep",
       kind: "dark small",
-      art: " ,_,\n(oo)-.\n /|\\\\",
-      lane: 72,
-      start: 93,
-      distance: -124,
+      art: SHEEP,
+      lane: 70,
+      start: -20,
+      distance: 132,
       delay: "-6s",
       duration: "13s",
     },
@@ -36,10 +42,10 @@
       id: "sheep-3",
       type: "sheep",
       kind: "light big",
-      art: "  .-''''-.\n /  (oo)  \\\n(__(____)__)\n   ||  ||",
-      lane: 108,
-      start: 116,
-      distance: -150,
+      art: SHEEP,
+      lane: 120,
+      start: -8,
+      distance: 150,
       delay: "-4s",
       duration: "16s",
     },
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

Replaces the eight bespoke (and admittedly ugly) per-actor ASCII sprites with a single reused `SHEEP` silhouette and a right-facing `DOG`, flips all actors to a consistent left→right drift (positive `distance`), and adds a `flock-wiggle-tiny` keyframe. Tests are updated in lockstep. No blocking issues found.

## Verification

- **No stale art references.** Grepped the tree for the old fragments (`(oo)`, `^..^`, `,_,`, and the `reverse` kind); the only hits are the updated assertions in `ui/src/lib/components/UpdateModal.browser.test.ts:120,121,184,206`. No `.reverse` CSS selector ever existed, so dropping the `reverse` kind label is inert.
- **String literals are valid and match tests.** `SHEEP` (`AsciiFlockOverlay.svelte:15`) contains `(o.o)` and `DOG` (`AsciiFlockOverlay.svelte:16`) contains `(___()`, satisfying `UpdateModal.browser.test.ts:120` and `:121`. The `DOG` double-quoted literal escapes its quotes/backslashes correctly (backticks are literal, harmless inside `".."`).
- **Actor/selector count still matches.** 7 sheep + 1 dog = 8 actors, matching the `:nth-child(1..8)` reduced-transform rules (`AsciiFlockOverlay.svelte:190-213`).
- **Animation directions agree with the comment.** All actors now use positive `--distance`, so `flock-cross` (`:215-222`) drives them left→right; the right-facing dog trots forward. Consistent.

## Findings

1. `ui/src/lib/components/AsciiFlockOverlay.svelte:168-173` — the new `.actor.tiny` animation override has specificity 0-2-0, which the CSS-only reduced-motion fallback `@media (prefers-reduced-motion: reduce){ .actor { animation: none } }` (`:275-279`, 0-1-0) cannot beat. So under the media-query fallback (e.g. before hydration / if JS doesn't run) the single tiny sheep keeps animating for reduced-motion users. The JS-driven `.flock.reduced .actor` rule (`:187-189`, 0-3-0) still stops it after `onMount`, so this is non-blocking, but consider adding the `tiny` (and ideally `dog`) selectors to the media block or bumping its specificity so the CSS fallback fully covers them.

- `.actor.dog` (`AsciiFlockOverlay.svelte:179-186`) has the same reduced-motion media-query specificity gap as the tiny sheep above; it predates this PR and its `.actor.dog` block is unchanged in the diff.

PR #1536, opened 2026-07-07
fix(doc-agent): format docs with ignores off so PRs pass CI

scoop:
## Problem

Doc-agent PRs (`docs: sync docs to recent source changes`, branch `shepherd/docs-update-*`) keep failing to auto-merge — e.g. #1517 sat red, and #1455/#1421 before it. The `doc-automerge.yml` workflow is **correct**; it rightly skips a PR whose checks aren't green. The real failure is upstream: the doc PRs themselves fail the CI `verify` job on a **Prettier** error, so they can never go green.

## Root cause (proven)

The doc agent already runs `prettier --write` before committing (`stageInScope`) and is fail-closed on prettier *errors* — but it runs prettier with `cwd = SERVER_INSTALL_ROOT`, so prettier reads that root's `.prettierignore`, which contains the glob `.shepherd-*`.

Doc-agent worktrees live at `dirname(repoPath)/.shepherd-worktrees/<name>` (`src/worktree.ts`). The absolute file path therefore contains the component `.shepherd-worktrees`, which matches `.shepherd-*` → prettier treats every doc file as **ignored and silently no-ops** (exit 0, no throw). The unformatted Markdown gets committed. CI checks the same file from a clean checkout (no `.shepherd-worktrees` ancestor → not ignored) and `prettier --check` fails → `verify` red → auto-merge skips forever, every sweep.

Concretely, #1517 added an `attachmentNames` table row wider than the column; Prettier re-aligns the whole table, so the raw agent output was non-conformant and the silently-skipped format pass never fixed it. Proof: `prettier --file-info` returns `{ "ignored": true }` on the worktree path vs `{ "ignored": false, "inferredParser": "markdown" }` at a plain path.

## Fix (`src/doc-agent.ts`, server-only)

1. **Root fix:** pass `--ignore-path <devNull>` on `prettier --write`. The file list is already the access-control boundary (`IN_SCOPE ∩ docs/ ∩ existing`), so prettier's ignore is redundant here — and, as shown, harmful: it skips based on *where the worktree lives*, not the file's identity. Disabling ignores makes formatting independent of the worktree's absolute location.
2. **Hardening (fail-closed on the whole silent-skip class):** extract `assertPrettierClean()` and run `prettier --check --ignore-path <devNull>` after the write, with **byte-identical** `cwd`/`--config`/`--ignore-path` (only the mode flag differs → identical plugin resolution → prettier idempotency, so a valid run never false-aborts). `--ignore-path <devNull>` keeps `--check` **non-vacuous** (a plain `--check` on an ignored file exits 0), so any residual non-conformance (future silent skip, config/version drift, plugin load failure) throws → `stageInScope` wraps it in `PrettierFormatError` → the run aborts (no commit/push/PR). A skipped doc update beats a red PR.

The `doc-automerge.yml` workflow is unchanged — it was already correct. No action taken on the currently-stuck #1517; once this lands, the next doc-agent sweep opens a correctly-formatted PR that supersedes it (as #1455/#1421 were superseded).

## Scope boundary

Only the three repo-root `docs/*.md` files reach root prettier (`stageInScope` filters `startsWith("docs/")`). In-scope `docs-site/src/content/docs/*.md` is Astro/Starlight-governed via the separate `docs-site` CI job — never root prettier — so a docs-site content failure is a distinct class, out of scope here.

## Tests (`test/doc-agent.test.ts`, real prettier)

- **Regression (write half) with an explicit negative control:** in a worktree nested under a `.shepherd-*` path, first assert the *unpatched* invocation leaves the file byte-unchanged (proving the layout genuinely reproduces the silent-skip, so the test can't pass vacuously); then assert the real `defaultPrettierWrite` formats it (table rows equal width); then assert the output passes `prettier --check` from a non-nested/clean path.
- **Guard-detection (check half):** `assertPrettierClean` rejects on an unformatted file (the real `--check` running) and resolves on a formatted one.

## Verification

- `bun run typecheck`, `bun run lint`, `bun test test/doc-agent.test.ts` (81 pass) — all green.
- Full `bun test`: my change adds +2 passing tests and introduces **zero** new failures (the pre-existing environment-dependent failures in browser-API suites are identical with and without this change).

[no-feature-entry] — server-only bug fix, no user-facing UX.

### DIFF
--- a/src/doc-agent.ts
+++ b/src/doc-agent.ts
@@ -1,6 +1,7 @@
 import { execFile } from "node:child_process";
 import { randomUUID } from "node:crypto";
 import { existsSync, readFileSync, writeFileSync } from "node:fs";
+import { devNull } from "node:os";
 import { join, resolve } from "node:path";
 import { promisify } from "node:util";
 import { buildTransientAgentArgv } from "./transient-agent-argv";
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

This PR fixes doc-PR auto-merge failures by (1) running the pinned prettier `--write` with `--ignore-path devNull` so the doc-agent worktree's `.shepherd-worktrees/` location can no longer cause prettier to silently treat every doc file as ignored, and (2) adding a fail-closed `--check` verify (`assertPrettierClean`) after the write so a residual non-conformance aborts the run instead of committing an un-mergeable PR. The change is correct and well-scoped.

## Verification

- **Root cause is real.** `.prettierignore` contains the `.shepherd-*` glob (`.prettierignore` — `.shepherd-*` line), and doc-agent worktrees are created under `.shepherd-worktrees` (`src/worktree.ts:90`, `src/worktree.ts:628`). Running prettier from `cwd=SERVER_INSTALL_ROOT` against those absolute paths matches the glob and silently no-ops the write — exactly the failure this fixes.
- **`devNull` import is valid.** `import { devNull } from "node:os"` (`src/doc-agent.ts:4`) is a documented `os` string constant; passing it via `--ignore-path` replaces prettier 3's default `[.gitignore, .prettierignore]`, disabling all ignores for the explicit whitelist.
- **No unintended files pulled in.** Disabling ignores could in principle un-ignore `docs/superpowers`, but the `docs/`-prefixed in-scope paths are only `docs/external-task-api.md`, `docs/sandbox-security.md`, `docs/token-usage-analysis.md` (`src/doc-agent.ts:152-154`), filtered by `p.startsWith("docs/")` (`src/doc-agent.ts:844`). None overlap `docs/superpowers`.
- **Format/CI parity.** `--write` output depends only on `--config` (the worktree `.prettierrc`), not on ignore files, so the committed bytes equal what CI's `prettier --check .` expects on a clean checkout where `docs/external-task-api.md` etc. are checked normally.
- **Fail-closed intact & idempotent.** `assertPrettierClean` reuses the same `cwd`/`configPath`/ignore flag as the write (`src/doc-agent.ts:42-54`, `src/doc-agent.ts:85-90`); write→check is idempotent, so a valid run never false-aborts. `stageInScope` wraps any throw into `PrettierFormatError` (`src/doc-agent.ts:853-857`) which `stageAndPublish` catches to abort with no commit/push/PR (`src/doc-agent.ts:876-882`). `execFileP` rejects on prettier's non-zero `--check` exit, so a non-conformant file correctly throws.
- **Wiring unchanged.** `this.prettier = deps.prettier ?? defaultPrettierWrite` (`src/doc-agent.ts:326`); making `defaultPrettierWrite` an export is source-compatible.
- **Config rationale holds.** `.prettierrc` declares `prettier-plugin-svelte`/`prettier-plugin-tailwindcss` (`.prettierrc:7`), which is why `cwd` must be `SERVER_INSTALL_ROOT`; both write and verify use the same cwd, so plugin resolution is identical.

The added real-prettier tests (`test/doc-agent.test.ts`) exercise the previously-mocked seam directly, including a negative control proving the old invocation silently skipped. Test correctness is CI's responsibility and was not run here.

No blocking issues and no nits.

PR #1543, opened 2026-07-07
style(top-bar): adopt 6px chip radius rung (#1533)

### LINKED ISSUE #1533: Revisit top-bar design tokens to match the GitRail pill-chip restyle
## Context

#1520 restyles the **GitRail** control row as rounded, semantically-accented **pill chips** — a deliberate deviation from DESIGN.md's "gently squared (2px radius), never pill-shaped" rule and the Four-Light Rule (green/amber/red reserved for agent state). That PR legitimizes the deviation narrowly (a GitRail exception in DESIGN.md + a documented recipe on `/design-system`).

That leaves the **top bar** (header / chrome bars — the right-cluster boxed controls: held/update/learnings badges, what's-new, docs link, gear, etc.) still on the older squared-token treatment. If the pill-chip language is the direction, the top bar's design tokens/controls should be revisited for consistency so the two chrome surfaces don't read as two different design systems.

## Ask

Revisit the top-bar design tokens and control chrome and decide how far the GitRail pill-chip treatment should propagate:

- Should top-bar controls adopt the same rounded pill-chip recipe + semantic accent, or stay squared?
- Reconcile the DESIGN.md rules (shape + Four-Light) once the intended end-state across both chrome surfaces is settled — either broaden the exception into a coherent rule, or keep GitRail as the sole carve-out.
- Keep the shared `.chip` recipe (introduced in #1520) as the single source if the treatment spreads.

## Scope notes

- **Not part of #1520** — that PR stays scoped to the GitRail row only. This is the explicitly-requested follow-up.
- Design-token only; consult `/design-system` first and reuse the `.chip` recipe rather than growing a new one.
- Presentational; no behavior change intended.

_Filed as a follow-up to #1520 per an attended planning session._

scoop:
Closes #1533.

Follow-up to #1520: the GitRail control row was restyled as 6px semantic chips, leaving the header top-bar right-cluster controls on the older squared (2px) treatment — so the two chrome surfaces read as two different design systems. This harmonizes them.

## Decision (settled with the operator against a live token-accurate mockup)

- **Fidelity = radius token only.** Add a `--radius-chip: 6px` token and swap the `2px` literals to it. Each control keeps its existing fill, `--topbar-ctl-h` (30px) height, sans label, and the update badge's amber pulse — it shares the 6px *rung* with the rail without becoming a literal `.status-chip`. (Fuller "flatten to chip" and "promote a shared class" options were shown and declined as heavier than intended.)
- **Uniform corners.** Every resting right-cluster boxed control takes the rung, so there is no mixed `2px`/`6px` seam.

## What changed

- **`app.css`** — new `--radius-chip: 6px` token (single source of the rung; mirrors DESIGN.json `rounded.chip`).
- **Top bar → token:** `.held-badge`, `.update-badge` (+`.herdr`/`.codex`), `.whatsnew-badge`, `.whatsnew-dot-btn`, `.search`. These five are the complete set of resting bordered rectangular controls in the desktop right cluster.
- **GitRail retokenize (visual no-op):** `.gbtn` / `.status-chip` / `.verdict-chip` / `.rail .gbtn` `6px` literals now read the token, so the rung has one definition across both surfaces.
- **`DESIGN.md` + `DESIGN.json` (lockstep):** a note extending the 6px rung to the top-bar right cluster as an explicit **radius-only** alignment — reconciling the constitution (issue ask #2) rather than implying the reuse. The prior chip-row clause was scoped to GitRail-named controls and didn't cover header chrome.

## Explicitly out of scope

The borderless `.gear` glyph (no radius) and the square (0-radius) status tallies (`.tally`/`.ctally`) are unchanged — they share only the `--topbar-ctl-h` height, not the rung. Popovers, tooltips, the ⌘K keycap, usage gauges, and the mobile sheet are separate chrome.

## Notes for review

- **Radius-only, by design.** The badges keep their tinted fill / 30px height / sans label / amber pulse, so they share the *shape* rung but are not canonical `panel-2` status-chips. No automated gate enforces recipe fidelity — flagged for conscious review. The "shared `.chip` recipe as single source" ask is **deferred** (no global recipe exists — #1520 is component-scoped).
- **Header coherence (tally seam):** the left/center active tallies stay square while the right cluster goes 6px; they sit at opposite ends (not abutting). If the square-vs-6px contrast reads poorly, the one-line follow-up is to extend the rung to `.tally.active`/`.ctally.active`.
- Presentational `style(...)` change — no user-facing string or behavior change, so no feature-catalog entry / no i18n. [no-feature-entry]

## Verification

`cd ui && bun run check` (0 errors, 2 pre-existing warnings in untouched files) · `bun run check:i18n` (2743 keys parity) · `bun run test` (3095 pass) · extension `bun run check` (0 errors) · prettier + eslint clean · `DESIGN.json` re-parses · pre-push gates green.


scoop (commented):
## Summary

This PR revisits the top-bar design tokens to match the GitRail pill-chip restyle (#1520/#1540/#1533). It introduces a single `--radius-chip: 6px` token and shares it across both chrome surfaces, documenting a deliberate **radius-only** alignment (not a full conversion to the `panel-2` status-chip recipe). This is a reasonable, conservative reading of the follow-up issue and stays presentational as intended.

### What it does
- `ui/src/app.css:122` — defines `--radius-chip: 6px` inside `:root`, so every scoped component style resolves it via the cascade.
- GitRail (`GitRail.svelte:868`, `git-rail/RailStatusActions.svelte:237,280,346`) — replaces the previously hard-coded `6px` with `var(--radius-chip)`, making the token the single source of the rung.
- Top bar — applies the token to the held badge (`TopBarHeldBadge.svelte:289`), update badge base `.update-badge` (`TopBarBadges.svelte:193`), what's-new badge (`:244`), compact what's-new pip (`:262`), and the ⌘K search trigger (`TopBarSearch.svelte:56`).
- `DESIGN.md` / `DESIGN.json` — document the exception.

### Verification (ground truth checked)
- **Token scope:** `--radius-chip` is defined once, at `ui/src/app.css:122`, inside the `:root` block (opens `app.css:35`, closes `app.css:154`). CSS custom properties inherit through the DOM, so the Svelte-scoped consumers all resolve it. Grep confirms exactly one definition and 9 consistent `var(--radius-chip)` references, no duplicate/stale value.
- **Excluded controls match the doc:** the doc says the borderless gear and the square status tallies are excluded and unchanged — confirmed: `TopBarGear.svelte` and `TopBarTallies.svelte` are not in the diff, and the health-pip keeps `border-radius: 50%` (`TopBarBadges.svelte:289`).
- **DESIGN.json validity:** the edited `Status Chip` `description` (DESIGN.json:204) is a single string with no embedded double quotes, so the file stays parseable; the literal `6px` in the reference `.ds-status-chip` CSS snippet is intentional (a static design-system example, not a token consumer).
- **Bordered claim:** every control in the newly-aligned set carries a `1px` border (amber/blue/line/line-bright), matching the doc's "four bordered right-cluster badges" framing.

### Nit (non-blocking)
- `DESIGN.md`: the new bullet describes each aligned control as keeping its chrome "...the fixed `--topbar-ctl-h` height, **sans label**, and the update badge's amber attention pulse." But on desktop the held, update, what's-new, and search controls all render a text label (only the compact what's-new pip is truly label-less). "sans label" reads as inaccurate for 4 of the 5 controls in the set — consider dropping or qualifying it (e.g. "the compact pip's label-less state") so the design doc doesn't misdescribe the very controls it governs.

Note: the update-badge radius change also applies to the `.update-badge.herdr` and `.update-badge.codex` variants (they share the base class), which the doc folds under "update." This is the correct, visually-consistent outcome — flagged only for completeness, not a defect.

The implementation satisfies the task and I found no correctness, security, or i18n issues (no user-facing strings or message keys were touched).

PR #1546, opened 2026-07-07
fix(ui): silence Svelte 5 non-reactive/local-state warnings

scoop:
## What

Clears two compile-time `vite-plugin-svelte` warnings surfaced during the build. Both were benign at runtime; this makes the intent explicit and quiets the noise.

- **`FilesPanel.svelte`** — the `fileInput` `bind:this` ref is now `$state<HTMLInputElement>()` (fixes `non_reactive_update`); its imperative `.click()` is guarded (`fileInput?.click()`) for the now-optional type.
- **`CommandBar.svelte`** — `filter` seeds via `untrack(() => initialFilter)`, documenting the deliberate one-time seed from the demo-only `initialFilter` prop and silencing `state_referenced_locally`. Added the `untrack` import.

No behavior change.

## Verification

`cd ui && bun run check` → 0 errors, 0 warnings.

### DIFF
--- a/ui/src/lib/components/CommandBar.svelte
+++ b/ui/src/lib/components/CommandBar.svelte
@@ -14,6 +14,7 @@
   import { jumpDigitIndex } from "$lib/components/herd-keynav";
   import { sortBlocked, type BlockState } from "$lib/triage";
   import { m } from "$lib/paraglide/messages";
+  import { untrack } from "svelte";
 
   let {
     sessions,
@@ -80,7 +81,7 @@
     | { kind: "doc"; title: string; url: string };
   type OptRow = Row & { oid: number };
 
-  let filter = $state(initialFilter ?? "");
+  let filter = $state(untrack(() => initialFilter) ?? "");
   let activeIdx = $state(0);
   let inputEl = $state<HTMLInputElement | null>(null);
   let listEl = $state<HTMLUListElement | null>(null);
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

The PR addresses exactly the two Svelte 5 compiler warnings from the build output, and does so correctly and idiomatically.

### `ui/src/lib/components/viewport/FilesPanel.svelte`
- `let fileInput: HTMLInputElement;` → `let fileInput = $state<HTMLInputElement>()` makes the `bind:this` target reactive, silencing `non_reactive_update`.
- Because `$state<HTMLInputElement>()` is declared with no initial value, the type becomes `HTMLInputElement | undefined`; the paired change `fileInput.click()` → `fileInput?.click()` (line 163) correctly guards the now-optional value.
- Verified all references are consistent: declaration (FilesPanel.svelte:35), call site (FilesPanel.svelte:163), and `bind:this` (FilesPanel.svelte:222). No other usages exist.

### `ui/src/lib/components/CommandBar.svelte`
- `$state(initialFilter ?? "")` → `$state(untrack(() => initialFilter) ?? "")` silences `state_referenced_locally` while preserving the intended behavior — seeding `filter` from the prop's initial value only. This matches the author's own comment (CommandBar.svelte:30-31: "real ⌘K path never passes this, so `filter` seeds to "" exactly as before").
- `untrack` is imported from `svelte` (CommandBar.svelte:17, single import) and is available in the pinned version `svelte@^5.56.4` (verified ui/package.json:36).

Both are the canonical Svelte 5 remedies for these warnings, behavior-preserving, and scoped tightly to the reported lines. Nothing to change.

PR #1547, opened 2026-07-07
style(viewport): move PR foldout into right-hand control cluster

scoop:
## Why

The PR disclosure control (`.git-toggle`, **● PR ▾**) sat in the **left** cluster of the viewport header, welded after the task name. A flex `.spacer` then pushes every other control to the right edge — so on a **short** task name the left cluster is tiny and the PR chip is left stranded in the middle of a wide empty gap, anchored to nothing. It "feels lost".

## What

Relocate the chip into the **right-hand work-state group**, between the post-tabs separator and the status mark:

```
before:  TASK-1442  warnings  [● PR ▾] ·············· Terminal Activity Diff Files Preview · ◆ PLANNING ⚙ ✕
after:   TASK-1442  warnings ························· Terminal Activity Diff Files Preview · [● PR ▾] ◆ PLANNING ⚙ ✕
```

Now it's anchored to the right cluster at **every** task-name length, and reads with the kin it shares meaning with (session status, plan phase), while staying clear of the window controls (gear / ✕).

## Scope

- **Markup move only** in `Viewport.svelte` — the button reads the same component-scope state (`gitOpen`, `prAttention`, `planAttention`, `prClear`, `gitToggleState`) at the new position. No behavior, CSS, i18n, glossary, or design-system change.
- **Desktop-only**, unchanged: the chip is `{#if !compact}`-gated and never rendered on compact/mobile (which shows the git strip unconditionally with no toggle).
- Tapping the task title (`onTitleTap`) still toggles the same strip — that affordance is untouched; the now-stale "welded to the title" comment was corrected.
- Not a feature — reposition only, so no feature-announcement entry (`style` commit; the catalog gate arms only on `feat`).

## Testing

- `cd ui && bun run check` → 0 errors.
- `cd ui && bun run test` → **3095 tests / 220 files pass**, including the real-browser `Viewport.browser.test.ts` (nothing asserted the old chip position).
- Layout matches the mockup approved during planning (short + long task names).

### DIFF
--- a/ui/src/lib/components/Viewport.svelte
+++ b/ui/src/lib/components/Viewport.svelte
@@ -2243,12 +2243,30 @@
       <!-- transient post-rename note (e.g. "branch kept"); the rename input itself
            takes the title's slot in place when active -->
       {@render renameNoteEl()}
-      <!-- git-rail disclosure, welded to the title: the full rail (PR / CI /
-           merge / critic / ready / verdict) plus the autopilot toggle live one
-           tap away, revealed as a second header row (.vp-git-strip). Sitting it
-           beside the title enlarges the tap target and pairs it with the
-           single-tap-toggle on the title itself (onTitleTap). A direct tap here
-           toggles instantly and is the keyboard/AT path (aria-expanded). -->
+    {/if}
+    <div class="spacer"></div>
+    <ViewportTabBar
+      bind:tab
+      {session}
+      {previewPort}
+      todoExists={!!todoExists}
+      {hasFiles}
+      {hasPreview}
+      {compact}
+      {headerFolded}
+      {vpBodyId}
+      {tabId}
+      {foldRegionId}
+    />
+    {#if !compact}
+      <span class="sep">·</span>
+      <!-- git-rail disclosure — grouped with the work-state chips (status ·
+           PLANNING) it shares meaning with, anchored in the right cluster so a
+           short task name no longer strands it mid-bar. Reveals the full rail
+           (PR / CI / merge / critic / ready / verdict) plus the autopilot toggle
+           as a second header row (.vp-git-strip). A direct tap toggles instantly
+           and is the keyboard/AT path (aria-expanded); tapping the task title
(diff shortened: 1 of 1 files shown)


scoop (commented):
## Summary

Style-only change: relocates the `.git-toggle` git-rail disclosure button from the left side of the header (after title/branch/renameNote, before `.spacer`) into the right-hand control cluster (after `ViewportTabBar`, beside the status/plan chips). This directly satisfies the task — a short task name no longer strands the foldout mid-bar; it now groups with the work-state chips on the right.

## Verification

- Render gating unchanged: the button was inside `{#if !compact}` before and is inside an identical `{#if !compact}` after (Viewport.svelte:2261-2289). No change to when it renders (still `!compact`, still independent of `renaming`).
- Scope preserved: Svelte template position does not affect variable scope. All referenced identifiers (`gitOpen`, `gitToggleState`, `prAttention`, `planAttention`, `prClear`, `m.viewport_git_actions*`) remain in the same component and are unchanged.
- Behavior preserved: `onclick` toggles `gitOpen` (Viewport.svelte:2277); `onTitleTap` also toggles `gitOpen` (verified at Viewport.svelte:914 and :923), so the new comment's claim at :2268-2269 that tapping the title toggles the same strip is accurate.
- Layout safe: `.git-toggle` CSS (Viewport.svelte:2943-2962) is class-only with `flex-shrink:0` and no positional/sibling combinators, so reordering the DOM node yields the intended flex placement without side effects.
- Template balance intact: `{#if !compact}` (:2236) → `{/if}` (:2246) now wraps branch/renameNote only; the moved `sep` + button live in a second `{#if !compact}` (:2261) → `{/if}` (:2289). Braces balance.

No user-facing strings or message keys were added or changed, so no i18n parity concern. No signature/contract changes. No bugs found.

The only incidental visual effect: the `·` separator that previously sat between the tab bar and the status area now precedes the git-toggle, leaving no separator between the toggle and the status mark. This reads as an intentional part of the restyle, not a defect.

PR #1562, opened 2026-07-08
fix(diagnostics): probe herdr server liveness, not just the binary (#1559)

### LINKED ISSUE #1559: DIAGNOSE shows herdr 'ok' while the server is offline — checks the binary, not server liveness (add functional test)
## Summary

The DIAGNOSE panel reports herdr as **"ok"** even when the herdr **server is offline** and every herdr call fails with connection-refused. The operator sees green "ok" while the system is unusable.

## Cause

The herdr preflight check runs `herdr --version` — it verifies the **binary is present** ("present-but-broken fails open"). It does **not** verify the server is running/reachable. A dead server (not started / stale socket) still passes as "ok".

## Suggestion — add a functional liveness check

Separate "installed" from "usable" for herdr in DIAGNOSE:

- Probe server liveness — e.g. `herdr status` reports `server: running`, or a lightweight `herdr agent list` succeeds — not just `--version`.
- Surface "installed but server offline" as a distinct amber/red state with a one-click **start/restart server** fix, instead of green "ok".

## Why it matters

DIAGNOSE is where an operator looks to answer "is it working". A binary-presence check that reads green while herdr is down defeats that purpose — observed together with a case where an update left the herdr server offline but DIAGNOSE still showed "ok".

scoop:
## Problem

Fixes #1559. The DIAGNOSE panel's herdr check ran only `herdr --version` — proving the **binary** is present, never that the **daemon** answers. A dead herdr server still read green **"ok"** while the system was unusable.

## Fix

Separate "installed" from "usable" for herdr:

- After confirming the binary/version, the herdr check now runs a **liveness probe** — `herdr agent list` (exit 0 = daemon reachable, connection-refused/timeout = offline). This is the same reachability signal `herdr-update` already relies on, and works regardless of the default-off socket driver.
- An offline daemon surfaces as a red **`error`** with a new `diagnostics_hint_herdr_offline` hint that names the concrete start command (`herdr server`), plus a guidance doc-link (parity with the tailscale/gh guidance rows).
- Precedence inside the check: **missing > offline > outdated > ok** — a live-but-old daemon stays a warning; a dead one is an error. Liveness is probed only when the binary is present.
- **Guidance-only** (no auto-run Fix button): there is no existing fix-map "start herdr server" primitive, and auto-starting a dead daemon is version/platform-dependent.

## Scope / non-goals

- **Boot preflight (`preflight.ts`) is deliberately untouched** — `restart.ts` documents that a down daemon must not trip the exit-78 fail-fast (Shepherd may be what starts it). This change is scoped to the re-runnable DIAGNOSE panel.
- Known transient: a Diagnose run *during* a `herdr update`/live-handoff can briefly read `offline`. Accepted — DIAGNOSE is a manual, re-runnable snapshot, not a poll/alert.

## Tests

Adds functional liveness cases to `test/diagnostics.test.ts` (the test the issue asks for): offline→`error`/`_offline` (+payload purity), live→`ok`, binary-missing→`_missing` with the liveness probe **not** invoked, offline-outranks-outdated, and offline driving `overall` to `error`. Also injects a resolving `runHerdrLiveness` into the `healthyDeps()` helper so existing "ok" assertions don't probe a real absent daemon.

Verified locally: `bun run lint`, `bun run typecheck`, `bun test test/diagnostics.test.ts` (67 pass), `cd ui && bun run check` + `check:i18n` — all green.

### DIFF
--- a/src/diagnostics.ts
+++ b/src/diagnostics.ts
@@ -74,10 +74,11 @@ export interface DiagnosticsDeps {
   /** Probe herdr **server liveness** (not just binary presence): run `herdr agent
    *  list`, resolving on a zero exit (daemon reachable), rejecting on
    *  connection-refused / timeout (daemon offline). Exit code only — the agent
-   *  listing is discarded, so nothing but a state crosses into the snapshot. This is
-   *  the same reachability signal `herdr-update` relies on. Default
-   *  `execFileAsync(config.herdrBin, ["agent","list"])` with the shared probe
-   *  timeout. Injected in tests (resolve = live, reject = offline). */
+   *  listing is discarded (routed to /dev/null, never buffered), so nothing but a
+   *  state crosses into the snapshot. This is the same reachability signal
+   *  `herdr-update` relies on. Default `spawn(config.herdrBin, ["agent","list"], {
+   *  stdio:"ignore" })` with the shared probe timeout. Injected in tests (resolve =
+   *  live, reject = offline). */
   runHerdrLiveness?: () => Promise<void>;
   /** Run `gh auth status`; resolves on a zero exit, rejects on non-zero / timeout.
    *  Its stdout is intentionally discarded (it carries account identity). On rejection
@@ -172,6 +173,37 @@ export function defaultRunRemediation(
   });
 }
 
+/** Default `runHerdrLiveness`: spawn `herdr agent list` with stdout/stderr routed to
+ *  /dev/null (`stdio: "ignore"`) — the exit code is the only signal, so a large agent
+ *  listing on a busy server is never buffered and can never exceed a buffer cap and
+ *  falsely read as offline (unlike `execFile`, whose 1 MiB `maxBuffer` would reject).
+ *  Resolves on exit 0 (daemon reachable); rejects on non-zero exit, spawn error
+ *  (ENOENT / ECONNREFUSED), or timeout (daemon offline). */
+function defaultHerdrLiveness(bin: string, timeoutMs: number): Promise<void> {
+  return new Promise<void>((resolve, reject) => {
+    const child = spawn(bin, ["agent", "list"], { stdio: "ignore" });
+    let settled = false;
+    const timer = setTimeout(() => {
+      settled = true;
+      child.kill("SIGKILL");
+      reject(new Error("herdr liveness timed out"));
+    }, timeoutMs);
+    child.on("error", (err) => {
+      if (settled) return;
+      settled = true;
+      clearTimeout(timer);
+      reject(err);
+    });
+    child.on("close", (code) => {
+      if (settled) return;
+      settled = true;
+      clearTimeout(timer);
+      if (code === 0) resolve();
+      else reject(new Error(`herdr agent list exited ${String(code)}`));
+    });
+  });
+}
+
 /**
  * Server-side environment-readiness diagnostics (issue #623). Fans dependency
  * probes with `Promise.all` behind a TTL cache. Mirrors HerdrUpdateService:
@@ -212,15 +244,7 @@ export class DiagnosticsService {
       });
     this.runHerdrLiveness =
       deps.runHerdrLiveness ??
-      (async () => {
-        // Exit code only — the agent listing is discarded. A zero exit means the
-        // daemon answered (reachable); a reject (ECONNREFUSED/ENOENT/timeout) means
-        // it's offline. `agent list` is the same reachability check `herdr-update` uses.
-        await execFileAsync(config.herdrBin, ["agent", "list"], {
-          encoding: "utf8",
-          timeout: DIAGNOSTICS_PROBE_TIMEOUT_MS,
-        });
-      });
+      (() => defaultHerdrLiveness(config.herdrBin, DIAGNOSTICS_PROBE_TIMEOUT_MS));
     this.runGhAuth =
       deps.runGhAuth ??
       (async () => {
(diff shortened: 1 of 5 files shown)


scoop (commented):
## Summary

The PR satisfies the task: DIAGNOSE now separates "herdr binary present" from "herdr server reachable". Once `--version` confirms the binary, a `runHerdrLiveness` probe (`herdr agent list`, exit-code only) runs; a reject flips the check to `error`/`diagnostics_hint_herdr_offline` instead of green `ok`. Offline correctly outranks the outdated-version warning, and the functional tests the issue asked for are present.

### Verification (ground truth compared)

- **Ordering logic** — `src/diagnostics.ts:277` returns `missing` and short-circuits before the liveness block at `:278-284`, and the liveness block precedes the semver-floor comparison at `:285`. So missing > offline > outdated ordering holds, matching the tests at `test/diagnostics.test.ts:180` (offline outranks outdated) and `:189` (liveness skipped when binary missing).
- **Liveness-signal claim** — the doc comment says `agent list` is "the same reachability check `herdr-update` uses"; confirmed against `src/herdr-update.ts:89` and `:106` (`timeout 10 ${h} agent list >/dev/null 2>&1`).
- **i18n parity** — `diagnostics_hint_herdr_offline` exists in BOTH catalogs: `ui/messages/en.json:1271` and `ui/messages/de.json:1271`. `hint()` resolves keys dynamically (`ui/src/lib/components/DiagnoseRows.svelte:54`), so the new key renders. Only en+de catalogs exist (`ui/messages/`).
- **UI / Fix wiring** — no `REMEDIATIONS` entry for `diagnostics_hint_herdr_offline` (`src/remediations.ts:44-56`), so `autoFixCommandFor` returns undefined and no Fix button is drawn; the added `DOC_LINKS` entry (`ui/src/lib/diagnostics-docs.ts:7`) makes the row render a doc link instead (`DiagnoseRows.svelte:107-120`). The prose hint itself tells the operator to run `herdr server`.
- **Payload purity** — offline path returns only `{id,state,hintKey}`; `check()` attaches no `remediation` (none exists), so `assertPure` at `test/diagnostics.test.ts:167` passes. Overall snapshot rolls up to `error` via `worstOf` (`test/diagnostics.test.ts:212-214`).
- **Existing tests unaffected** — `healthyDeps()` now injects `runHerdrLiveness: async () => {}` (`test/diagnostics.test.ts:44`), keeping the all-ok/herdr-ok assertions green.

### Could-not-verify (external, not in repo)

- The exact exit-code semantics of `herdr agent list` (does it exit non-zero ONLY when the daemon is unreachable, or also on e.g. an empty agent set?) cannot be verified — the `herdr` binary is not in this repo. The PR reuses the same assumption `herdr-update.ts` already ships, so it is at least internally consistent; assumed as written.

### Nit (non-blocking)

The default `runHerdrLiveness` uses `execFileAsync(config.herdrBin, ["agent","list"])`, which buffers the command's stdout into memory (Node's default `maxBuffer` is 1 MiB) even though the comment states the listing is discarded. On a healthy-but-busy server whose `agent list` output exceeds 1 MiB, `execFile` rejects with `ERR_CHILD_PROCESS_STDIO_MAXBUFFER`, and the `catch` would misreport a live daemon as `offline`. Unlikely in practice, but since the output is thrown away, consider `stdio: "ignore"` (via a small spawn helper) or a generous `maxBuffer` so a large listing can't flip a live server to offline. Matches the `>/dev/null` discipline already used in `herdr-update.ts:106`.

- `src/remediations.ts`: no one-click start/restart-server remediation exists for the offline state (the issue floated it as a suggestion); the PR instead surfaces prose + a doc link, which is a reasonable choice since `herdr server` is interactive/long-running rather than an unattended install.

scoop (commented):
## Summary

The PR adds a functional herdr **server-liveness** probe to DIAGNOSE, so an installed-but-offline daemon now reports `error` (offline) instead of a green `ok`. This satisfies the task: `versionProbe` gains an optional `liveness` step that runs only after binary presence is confirmed, and offline outranks the outdated-version warning. Verified against the actual code — no blocking issues.

## Prior-revision finding — resolved

- **`execFile` maxBuffer risk (buffering `herdr agent list` stdout):** RESOLVED. `defaultHerdrLiveness` now uses `spawn(bin, ["agent","list"], { stdio: "ignore" })` (src/diagnostics.ts:184) — stdout/stderr go to /dev/null, only the exit code is read, so a large agent listing can no longer exceed a buffer cap and falsely read as offline. `spawn` is imported (src/diagnostics.ts:1).

## Verification

- Liveness ordering: `versionProbe` returns `_missing` before touching `liveness`, so a missing binary skips the probe (src/diagnostics.ts:296–310); offline is checked before the semver floor, so offline outranks outdated. Matches the tests and the doc comment.
- Probe command consistency: `["agent","list"]` is the same reachability signal `herdr-update` gates on (`timeout 10 herdr agent list` at src/herdr-update.ts:106), and matches HerdrService usage (src/herdr.ts:296,305). Consistent.
- Remediation text accuracy: the offline hint tells the operator to run `herdr server`; that is the real start command (`setsid ${h} server` at src/herdr-update.ts:110). Correct.
- `config.herdrBin` resolves (src/config.ts:455).
- Timeout/settle logic in `defaultHerdrLiveness` (src/diagnostics.ts:182–200): single `settled` guard across `error`/`close`/timeout paths, `clearTimeout` on both settle paths, SIGKILL on timeout — no double-settle, no leaked timer.
- i18n parity: `diagnostics_hint_herdr_offline` present in BOTH en (ui/messages/en.json:1271) and de (ui/messages/de.json:1271), and wired into DOC_LINKS (ui/src/lib/diagnostics-docs.ts:7). No missing catalog entry.

## Note (non-blocking, not raised as a finding)

The originating issue suggested a one-click start/restart-server fix. The PR surfaces the offline state with a doc link and a `herdr server` instruction in the hint text, but does not add an `autoFixCommandFor` mapping for `diagnostics_hint_herdr_offline` (remediations.ts is out of scope / untouched). This is a suggestion in the issue, not a correctness gap — the offline state and functional test that the task asked for are delivered.

No findings.

PR #1563, opened 2026-07-08
fix(herdr): recover herdr server after Update Herd (#1558)

### LINKED ISSUE #1558: bug: 'Update Herd' leaves the herdr server offline (connection-refused loop) — nothing restarts it
## Summary

After **"Update Herd"** (herdr updated 0.7.1 → 0.7.3), the herdr **server did not come back up**. The HUD drops to the "herdr offline / reconnect" screen with a repeating `Error: Os { code: 111, kind: ConnectionRefused, message: "Connection refused" }` loop. A stale `~/.config/herdr/herdr.sock` remains with no listener behind it.

## Cause

On a self-hosted headless setup the herdr server is a long-lived process the operator manages — Shepherd does not own/restart the herdr server lifecycle. "Update Herd" stops the running server to swap the binary, but **nothing restarts it**. If the server is run under a systemd unit with `Restart=on-failure`, the clean stop (exit 0) during update does not trigger a restart, so herdr stays offline and clients loop on connection-refused against the stale socket.

## Repro

1. Self-hosted headless instance, herdr server running.
2. Trigger "Update Herd" (herdr binary updates).
3. Server stops for the swap and does not restart → connection-refused loop; stale socket left behind.

## Suggestions

- After a herdr update, ensure the server is restarted (or clearly own/instruct the restart) and clear a stale socket.
- Consider herdr `--handoff` for an in-place update so live sessions survive.
- Docs: for a self-managed herdr-server unit, recommend `Restart=always` so it survives an update-induced clean stop (that is what fixed it here).

## Related

Part of the broader gap that the headless install doesn't set up or own the herdr server lifecycle.

scoop:
## Problem

Fixes #1558. After **"Update Herd"** the herdr **server** did not come back up: the HUD dropped to the "herdr offline / reconnect" screen with a repeating `ConnectionRefused` loop, and a stale `~/.config/herdr/herdr.sock` was left with no listener.

## Root cause

`buildUpdateScript()` in `src/herdr-update.ts` gated its post-update recovery behind `if [ "$rc" -ne 0 ]`. But `herdr update` exits **0** even when the swap left no running server, so on the common *successful-update-but-server-down* path the script made **no** herdr call at all — neither the auto-spawning `agent list` nor the fallback relaunch fired. On a self-managed herdr systemd unit with `Restart=on-failure`, the clean stop (exit 0) during the swap doesn't trigger a restart either, so nothing on any layer brought the server back.

## Fix

- Run `herdr agent list` **unconditionally** after the update (not gated on `rc`). A herdr CLI call auto-spawns the daemon (`src/herdr.ts:44-46`), so this both verifies **and** recovers the server — and it runs the real `herdr` binary directly, so it's independent of the socket-vs-CLI driver.
- Wrap it in a **bounded grace + retry** so an in-flight `--handoff` (rc==0 bind window) or a systemd `Restart=always` unit binds first and is never mistaken for "down".
- **Retain** the detached `setsid herdr server` relaunch as a **last-resort fallback**, fired only after the retries still report unreachable — a hedge in case herdr 0.7.x no longer auto-spawns on `agent list` (the auto-spawn evidence is from the 0.6.x-era design doc). Removing it outright would regress vs. today's `rc!=0` path.
- **No socket `rm`**: herdr clears/rebinds its own stale socket on spawn, and an external `rm` could destroy a concurrently-recovering server's socket.

## Docs

README gains a **"herdr server lifecycle & Update Herd"** subsection recommending **`Restart=always`** (not `on-failure`) for a self-managed herdr-server systemd unit — the durable operator-side fix. Shepherd's post-update recovery is documented as a subordinate best-effort belt (its relaunched server is a cgroup child, not durable across `systemctl restart shepherd`).

## Not done here (noted in the issue as the "broader gap")

- Deeper socket-driver reliability gap (`herdr-socket-driver` connects but never respawns a dead server at *any* time).
- Full lifecycle ownership: shipping + enabling a `herdr-server.service` from `deploy/install.sh`.

## Testing

- `test/herdr-update.test.ts` updated: `agent list` runs regardless of `rc`; grace+retry present; `setsid ... server &` only in the post-retry `else` branch; **no** `rm`/socket removal; `HERDR_BIN` still threaded through every call; marker count.
- Generated script verified with `bash -n`.
- Root `bun run lint` clean; `bun run test` — **6270 pass / 0 fail**.

Verified end-to-end against a live herdr is not possible in CI (herdr is absent from the runner image); the `buildUpdateScript` string tests are the established proxy for this file, and the change is deliberately minimal + non-destructive because of that limit.

[no-feature-entry] — server-side lifecycle fix, no user-facing UX surface.

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,6 +410,24 @@ single-operator password and stores browser access in a signed session cookie. S
 `SHEPHERD_PASSWORD` in `~/.shepherd/env`, or use the first-boot generated password from the server
 log; machine clients can use `SHEPHERD_TOKEN` as a bearer alternative.
 
+### herdr server lifecycle & "Update Herd"
+
+Shepherd drives herdr but **does not own the herdr server's lifecycle** — the daemon is
+normally auto-spawned on demand by any `herdr` CLI call. **"Update Herd"** swaps the herdr
+binary (`herdr update --handoff`), which stops the running server for the swap. Afterwards
+Shepherd makes a best-effort attempt to bring it back: it runs `herdr agent list` (which
+auto-spawns the daemon) with a short grace+retry, and only if that still fails does it
+relaunch a detached server so orphaned panes reattach.
+
+If you run the herdr **server** under your own systemd unit, use **`Restart=always`**, not
+`Restart=on-failure`. The stop during an update is a _clean_ exit (status 0), which
+`Restart=on-failure` does **not** treat as a restart trigger — so the server would stay down
+after "Update Herd", leaving a stale `~/.config/herdr/herdr.sock` and clients looping on
+`ConnectionRefused`. `Restart=always` survives that clean stop and is the durable fix;
+Shepherd's post-update recovery above is a subordinate best-effort belt (its relaunched
+server is a child of Shepherd's cgroup and is not durable across a `systemctl restart
+shepherd`).
+
 ### Host tuning — tmpfs inodes
 
 Shepherd keeps spawned agents' Node compile cache **off** the `/tmp` tmpfs (it points
--- a/src/herdr-update.ts
+++ b/src/herdr-update.ts
@@ -86,12 +86,16 @@ export function buildUpdateScript(
   // so gating recovery on `rc != 0` (as we used to) skipped the exact bug — a
   // successful swap that never brought the server back, leaving a stale socket and
   // clients looping on ConnectionRefused. So we ALWAYS run `herdr agent list` after
-  // the update, regardless of rc. That CLI call is itself the recovery: a herdr CLI
-  // call is documented to auto-spawn the daemon (herdr.ts:44-46; the "herdr update
-  // without a shepherd restart" design doc, item 4 body). It runs the real herdr
-  // binary directly, so it bypasses shepherd's in-process maintenance guard (which
-  // only gates shepherd's own Runner) and can resurrect the server before the poller
-  // even resumes — driver-independent (it does not depend on SHEPHERD_HERDR_SOCKET).
+  // the update, regardless of rc. That CLI call is itself the recovery: any herdr
+  // CLI call auto-spawns the daemon (stated in the "herdr update without a shepherd
+  // restart" design doc, item 4 body — docs/superpowers/specs/
+  // 2026-06-04-herdr-update-without-restart-design.md). Shepherd's own maintenance
+  // guard (`HerdrUnavailableError`, herdr.ts:44-46) corroborates it: that guard
+  // exists specifically to STOP shepherd's Runner from spawning mid-update, which
+  // only matters because a normal CLI call would. Our `agent list` here runs the
+  // real herdr binary directly, bypassing that in-process guard (which only gates
+  // shepherd's own Runner), so it can resurrect the server before the poller even
+  // resumes — driver-independent (it does not depend on SHEPHERD_HERDR_SOCKET).
   //
   // Grace + retry: a still-binding server — an in-flight `--handoff`, or a self-
   // managed systemd unit with `Restart=always` coming up — must not be mistaken for
--- a/src/herdr-update.ts
+++ b/src/herdr-update.ts
@@ -82,18 +82,36 @@ export function buildUpdateScript(
   // pane. Not stopping means a failed update usually leaves the live server +
   // panes untouched.
   //
-  // Recovery (only when the update fails): because the agent targets outlive a
-  // dead server, if the server is actually gone we relaunch one and the orphaned
-  // panes reattach — the app's Reconnect / `viewport_herdr_unreachable` path, the
-  // "live agent returns once herdr is back" invariant from #413. We gate the
-  // relaunch on `agent list`, the SAME throw-on-unreachable signal broadcast()
-  // catches in service.ts, and relaunch ONLY when it fails — an unconditional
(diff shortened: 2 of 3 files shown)


scoop (commented):
## Summary

The PR fixes the reported bug. In the generated update shell script (`src/herdr-update.ts:115-138`), recovery is no longer wrapped in `if [ "$rc" -ne 0 ]`; instead `herdr agent list` runs unconditionally after `herdr update --handoff`, with a 3-attempt grace+retry loop, falling back to a detached `setsid herdr server` only if the server is still unreachable. This directly addresses the issue: `herdr update` exits 0 even when it leaves no running server, so the old rc-gated path skipped the exact failure. The README addition documents the durable operator-side fix (`Restart=always`).

### Logic verification
- The script now always probes/auto-spawns after update, regardless of rc — matches the issue's root cause (clean exit 0 with no server). Verified against `src/herdr-update.ts:123-136`.
- Fallback `setsid herdr server </dev/null >/dev/null 2>&1 &` redirects all fds, so it detaches cleanly from the `{ } | tee` pipe and won't hang `tee`. Verified `src/herdr-update.ts:135`.
- The direct-binary `agent list` correctly bypasses shepherd's in-process maintenance guard (`makeHerdrRunner` throws `HerdrUnavailableError` while `maintenance.active`) because the script shells the real herdr binary, not shepherd's Runner. Verified against `src/herdr.ts:59-63`.
- Deliberately does NOT `rm` the socket — herdr rebinds its own stale socket on spawn; avoids destroying a concurrently-recovering server's socket. Consistent with the issue's caution and the test assertions.
- Tests align with the emitted script text: `not.toContain('if [ "$rc" -ne 0 ]')`, `for attempt in 1 2 3`, `sleep 2`, `ok=1`, the `setsid ... server ... &` match, the `else ... unreachable after retries` match, and the 4-marker count all correspond to the generated lines. Verified against `src/herdr-update.ts:120-135` vs `test/herdr-update.test.ts:24-79`.

### Nit (non-blocking)
- `src/herdr-update.ts:90` — the comment claims a herdr CLI call "is documented to auto-spawn the daemon (herdr.ts:44-46 ...)", but `herdr.ts:44-48` actually defines `HerdrUnavailableError`, documenting the *opposite* (a call during an in-flight update is thrown WITHOUT spawning). No auto-spawn documentation exists at those lines (grep for "auto-spawn"/"on demand" in `src/herdr.ts` only hits line 356, an unrelated workspace comment). The comment's operative reasoning — that the direct-binary call bypasses the in-process guard at `src/herdr.ts:63` — is sound; only the line citation is misleading. Recommend correcting or dropping the `herdr.ts:44-46` reference.

### Minor observation (not a finding)
- The retry loop `sleep 2`s even after the 3rd (final) failed attempt before entering the fallback `else`, wasting ~2s in the failure path only. Negligible; not worth a change on its own.

Could not verify the herdr binary's actual auto-spawn-on-`agent list` behavior for 0.7.x (external tool, not in this repo) — the code sensibly keeps the explicit `setsid ... server` relaunch as a hedge for exactly that uncertainty, so correctness does not depend on the auto-spawn premise holding.

scoop (commented):
## Review

This PR fixes the "Update Herd leaves herdr offline" bug by making the post-update recovery **unconditional** (no longer gated on `rc != 0`), since `herdr update` exits 0 even when it leaves no running server. It runs `herdr agent list` (which auto-spawns the daemon) in a grace+retry loop and, only if that still fails, relaunches a detached `setsid herdr server`. README documents the operator-side `Restart=always` durable fix. This satisfies the originating issue.

### Prior finding — resolved
- `src/herdr-update.ts:89-95`: The previously-misleading citation is fixed. The auto-spawn claim is now attributed to the design doc, verified against `docs/superpowers/specs/2026-06-04-herdr-update-without-restart-design.md:33` ("any herdr CLI call auto-spawns the daemon"). The `herdr.ts:44-46` reference is now used only for `HerdrUnavailableError`, verified against `src/herdr.ts:44-46` (the guard that is thrown WITHOUT spawning) — the comment's characterization ("that guard exists specifically to STOP shepherd's Runner from spawning mid-update") matches the actual code and doc string at `src/herdr.ts:44-45`.

### Verification
- Recovery shell logic verified against `src/herdr-update.ts:127-140`: unconditional `ok=0`/`for attempt in 1 2 3` retry, `else`-branch fallback `setsid ${h} server </dev/null >/dev/null 2>&1 &` (detached; won't hold the `tee` pipe open). Matches README description at `README.md:416-419`.
- Marker count: 4 `UPDATE_LOG_PREFIX` echoes (`:124,126,136,138`) — consistent with the test at `test/herdr-update.test.ts` expecting `markers.length === 4`. The `=== herdr-update ...` header at `:123` does not contain the prefix `>>> herdr-update:` (verified `src/herdr-update.ts:27`), so it is correctly excluded.
- Test regex `/else\s*\n\s*echo '[^']*unreachable after retries/` — `UPDATE_LOG_PREFIX` contains no single quote (`src/herdr-update.ts:27`), so `[^']*` spans it correctly.

Minor, non-blocking observation (not a required change): the retry loop runs `sleep 2` even after the final (3rd) attempt fails, adding ~2s of needless delay on the already-failing fallback path. Harmless — arguably extra grace time — so not raised as a finding.

PR #1564, opened 2026-07-08
refactor(ui): optimize the two mobile session tab bars

scoop:
## What & why

Reworks the two stacked bars in the **mobile** session view so the full cursor-arrow cluster stays visible on a portrait phone, and consolidates the sticky corner buttons. Also **removes the broadcast button** (hypothesis: it gets little use — this is the test).

### Layout (mobile/touch)

**Before**
```
Row 1  [ ⌁broadcast ]  steer chips…                    [ ABC ]
Row 2  [ Esc ]  Tab ␣ ← → ↑ ↓ ^A ^E ^U ^C ^D…  [ ⬆ ][ mic ][ ⏎ ]
```
**After**
```
Row 1  [ Esc ]  steer chips…                  [ ABC ][ mic ]
Row 2  [ ⬆ ]  ← → ↑ ↓  Tab ␣ ^A ^E ^U ^C ^D…              [ ⏎ ]
```

- **Esc** → top-left · **mic** → top-right corner (ABC immediately left of it) · **image upload** → bottom-left · **Enter** stays bottom-right. All four corners stay sticky; both centers scroll horizontally.
- **Arrows lead** Row 2's scroll region so the whole cursor cluster is visible at portrait width; Tab/␣/^-keys scroll off to the right (reachable by scrolling). Committed as its **own commit** (`6bc95ac2`) so it's independently revertible — it shifts Tab/␣ position for all mobile users and is an A/B hypothesis.
- **Broadcast removed everywhere** (desktop + mobile), but it **stays reachable via the command bar** — only the steer-bar trigger button + its prop chain were removed; `showBroadcast`/`BroadcastComposer` are untouched, so this is trivially reversible.
- **Desktop is unchanged** except broadcast is gone. Esc/mic are gated on `mobile || touch`; the mic-support probe (`getVoiceStatus()` + `MediaRecorder`/`SpeechRecognition`) now early-returns on desktop, so desktop sessions no longer fire that needless network call (a small net win).

### Architecture note

The mic button moves to Row 1 but opens the compose sheet Row 2 owned, so the compose-sheet state + `<ComposeBar>` mount + mic-support detection **lift to the shared parent `Viewport.svelte`**; both rows get callbacks. Preserved: mic → sheet opens already dictating; swipe-up on Row 2 → sheet opens in type mode; the Android IME path (`composeKeystrokes`).

## Verification

- **Empirical arrows A/B** (new browser test, `ViewportTermControls.browser.test.ts`): at 390px portrait the scroll region is 270px; all four arrows fit with **79px** spare vs a **92px** Tab/␣ span — so the pre-reorder (edit-first) order would have pushed the last arrow **~13px off-screen**. The reorder is what keeps the full cluster visible.
- `bun run check` (0 errors / 0 warnings), `bun run test` (full UI suite green), `bun run check:i18n` (parity, orphaned `steerbar_broadcast*` keys removed), prettier + eslint clean.
- Reviewed per-task and whole-branch (accessibility: all corner buttons keep 44px targets + aria names; no dangling symbols/messages).

`[no-feature-entry]` — this is a rearrangement/removal, not a new user-facing capability.


scoop (commented):
## Summary

The two mobile rows are reworked exactly as described: the broadcast chip is gone, Esc moves to Row 1's left edge (`SteerBar.svelte:178`, `ControlBar include={["cancel"]} scroll={false}`), image upload takes Esc's old far-left slot on Row 2 (`ViewportTermControls.svelte:79`), the dictate mic moves to Row 1's right edge above Enter, ABC sits left of the mic, and the arrow cluster now leads the scroll region so it stays visible on a portrait phone. Frozen corners + scrolling center are preserved on both rows. The change is well-executed and I found no blocking defects.

### Verification performed
- **Broadcast removal / i18n parity**: `steerbar_broadcast` + `steerbar_broadcast_aria` removed from BOTH catalogs in parity (`ui/messages/en.json:26-27`, `ui/messages/de.json:9-10`), and the coach-hint string updated in both (en.json:31 / de.json:14). No `m.steerbar_broadcast` reference survives anywhere in `ui/src`. The broadcast *feature* is intact and still reachable off the removed button: command-palette entry `onBroadcast: () => (showBroadcast = true)` (`ui/src/routes/+page.svelte:1701`) and the still-mounted `BroadcastDialog` (`ui/src/lib/components/page/AppOverlays.svelte:573-574`). `onbroadcast` prop cleanly dropped from Viewport and both `+page.svelte` call sites.
- **ControlBar render order**: reordering `controlKeys()` to nav-first (`controlKeys.ts:33-40`) is safe — `ControlBar` iterates `controlKeys()` and filters by group membership, so `include` array order is irrelevant (verified `ui/src/lib/components/ControlBar.svelte:22-31`). The in-diff comment at `ViewportTermControls.svelte:129-130` correctly documents this. `controlKeys.test.ts` (out of diff) asserts via order-independent `toEqual`/`Set` and is unaffected.
- **Compose sheet relocation**: `composeKeystrokes` is imported where it's now used (`ui/src/lib/components/Viewport.svelte:40`, used at :279); the moved `<ComposeBar>` props (`onsend`/`onclose`/`repoPath`/`startDictation`) match `ComposeBar.svelte:25-37`; no orphaned `ComposeBar`/`getVoiceStatus`/`repoPath`/`composeOpen` refs remain in `ViewportTermControls.svelte`. Both compose entry points (`onsummon` swipe-up, `ondictate` mic) are only reachable in the `tab === "term"` context, matching the sheet's prior mount gating.
- **Dictate aria key**: `composebar_dictate_aria` present in both catalogs (en.json:83 / de.json:83), used at `SteerBar.svelte:263-264`.
- **Mic gating** moved to Viewport (`micAvailable = (mobile||touch) && (speechSupported||localVoiceAvailable)`, `Viewport.svelte:546`) with the `getVoiceStatus` probe gated to touch layouts (`:540-545`) — behaviourally equivalent to the old ViewportTermControls logic.

## Non-blocking

**ui/src/lib/components/SteerBar.svelte: the mobile ABC toggle can now be a dead no-op.** Previously the always-present broadcast chip guaranteed at least one collapsed label on mobile (its `.bc-label` collapsed unconditionally under the `@media` block), so the always-visible `.lbl-toggle` (`SteerBar.svelte:455-457`) always had something to reveal. With the broadcast chip gone, the only `.bc-label` left belongs to the retry chip, which renders only `{#if retryReady && retryHaltedCount > 0}` (`SteerBar.svelte:190-201`). Steer-chip `.chip-label`s collapse only under `.compact` (overflow), not by mobile viewport (`SteerBar.svelte:385-388`). So on mobile with no halted-retry chip and few enough chips that the bar doesn't overflow, the ABC toggle is still shown but toggling `show-labels` reveals nothing — a dead control. The adjacent comment "on mobile a label always collapses, so there's always a label to reveal" (`SteerBar.svelte:443`) is now inaccurate, and the new test deliberately injects a retry chip precisely because the broadcast chip is gone (`SteerBar.browser.test.ts:127-141`). Consider hiding `.lbl-toggle` on mobile when nothing is actually collapsed, or updating the comment to reflect the narrower invariant. Minor UX/consistency only — not blocking.

scoop:
**Re: mobile ABC (.lbl-toggle) no-op / stale comment (SteerBar.svelte)** — fixed in `7827b343`.

The concrete defect was the **inaccurate comment** ("on mobile a label always collapses"), which the removed always-present broadcast chip used to anchor. Rewrote it to describe the real behavior.

I took the *correct-the-comment* option rather than *gate the toggle's visibility*, deliberately:

- ABC now shares a **fixed pinned corner cluster with the mic** (`[ABC][mic]`). Gating ABC on "is something currently collapsed" would make that cluster **shift/flicker** whenever a retry chip appears/clears or the bar crosses the overflow threshold (`compact` is toggled async by `fitLabels` during layout settling). A stable slot next to the mic is better UX than eliminating a rare no-op.
- The no-op is genuinely edge-case: it only occurs with **plain-text chips, no retry chip, and no overflow**. Emoji steer bars (the common case) overflow → `compact` → ABC is meaningful; and with a retry chip present the retry count collapses on mobile → ABC is meaningful. Both are covered by existing tests (SteerBar.browser.test.ts:126, :270).
- The always-visible-on-mobile behavior is itself asserted by an existing test (:273), i.e. it's an intentional persistent affordance, not accidental leftover.

Happy to gate it (via a `:has(.retry-chip)` + `compact` reveal and a test update) if you'd rather trade the layout stability for eliminating the inert edge — flagging the tradeoff so you can weigh it.

scoop (commented):

**1. SteerBar.svelte — stale/no-op ABC (.lbl-toggle) comment** — ADDRESSED. The comment block at `ui/src/lib/components/SteerBar.svelte:441-454` is rewritten and now accurately describes the real behavior: it states the toggle is "revealed unconditionally (NOT gated on something currently being collapsed)" and explicitly documents the inert edge ("With only plain-text chips and no retry chip there's nothing to expand and the toggle is inert — a harmless, uncommon edge"). This matches the CSS at `SteerBar.svelte:459-461` (`.lbl-toggle { display: inline-flex; }` unconditionally inside the media query). The original finding offered "gate OR correct the comment"; the author took the correct-the-comment option, which resolves it. Not re-raised.

## Verification performed

- **i18n parity**: `steerbar_broadcast` / `steerbar_broadcast_aria` removed and `steerbar_coach_hint` shortened in BOTH catalogs (`ui/messages/en.json:720-`, `ui/messages/de.json:720-`); no remaining code references either removed key (grep of `ui/src` clean). `composebar_dictate_aria` (newly referenced from SteerBar) exists in both — `ui/messages/en.json:83`, `ui/messages/de.json:83`.
- **New/moved identifiers resolve**: `ControlBar` props `onkey`/`include`/`scroll` match `ui/src/lib/components/ControlBar.svelte:9-17`; `ComposeBar` props `onsend`/`onclose`/`repoPath`/`startDictation` match `ui/src/lib/components/ComposeBar.svelte:26-37`; `getVoiceStatus` exists at `ui/src/lib/api.ts:334`; `composeKeystrokes` already imported at `Viewport.svelte:40`; `enterKey` exists at `ui/src/lib/controlKeys.ts:50`.
- **Component contract change**: `ViewportTermControls` swapped `repoPath` → `onsummon`; the only instantiation (`Viewport.svelte:2672`) passes `onsummon` and no `repoPath` — consistent (`ViewportTermControls.svelte:14/28`). No leftover references to `composeKeystrokes`/`getVoiceStatus`/`ComposeBar`/`composeOpen` in `ViewportTermControls.svelte` (grep clean) — the compose sheet is now mounted once in `Viewport.svelte:2687-2694`, so no double-mount.
- **onbroadcast removal**: dropped from `SteerBar`, `Viewport`, and both `+page.svelte` call sites; broadcast dialog remains reachable via `+page.svelte:1701` (`onBroadcast` command entry), so the feature is not orphaned — consistent with the `[no-feature-entry]` intent.
- **Reorder correctness**: `controlKeys()` now lists nav before edit (`controlKeys.ts:52-59`); `ControlBar` renders in palette order (`ControlBar.svelte:22-31` iterates `controlKeys()`), so the `include={["nav","edit","signal"]}` filter yields arrows-first as the comment at `ViewportTermControls.svelte:129` claims. Both `ControlBar` consumers (SteerBar Esc-only, ViewportTermControls scroll cluster) verified.
- **Layout/corners**: `.steer-row` is flex (`SteerBar.svelte:314`); Esc bar (`noscroll`, flex 0 0 auto) frozen left, `.steer-bar` (flex 1 1 auto) scrolls, ABC + mic (`.dictate` flex 0 0 auto) frozen right — matches the task's sticky-corners / center-scroll requirement. On mobile `.edit-steers` is hidden (`SteerBar.svelte:462-464`) so the visible right cluster is `[ABC][mic]`, i.e. ABC left of mic as specified; mic and Enter both sit 10px from the bar edge for vertical alignment across the two rows.

The implementation satisfies the stated task (drop broadcast, Esc→Row 1 left, upload→Row 2 left, mic→above Enter, ABC left of mic, corners sticky) and I found no correctness, security, or parity defects in the diff.

PR #1565, opened 2026-07-08
fix(codex): update via `codex update` so standalone installs upgrade (#1560)

### LINKED ISSUE #1560: bug: 'Update Codex' runs npm but the on-PATH codex is a standalone install -> version never changes, update loops
## Summary

"Update Codex" reports **"Update not complete (still on 0.142.5). Check the log; you can retry."** The codex-update log (`~/.shepherd/codex-update.log`) shows `npm install -g @openai/codex` **succeeding** (rc=0, 0.142.5 → 0.143.0) — yet `codex --version` still reports 0.142.5, so Shepherd flags the update as failed and keeps offering retry.

## Cause

Shepherd updates codex via `npm install -g @openai/codex` (updates the **npm-global** package). But the `codex` on `PATH` here is the **standalone** distribution: `~/.local/bin/codex` → `~/.codex/packages/standalone/current/bin/codex`, which npm never touches. So npm updates the npm copy to 0.143.0 while the standalone binary on PATH stays 0.142.5. Shepherd's post-update `codex --version` check sees no change → "not complete", indefinitely retryable.

## Repro

- codex present as the standalone build (`~/.local/bin/codex` symlinked into `~/.codex/packages/standalone/...`), alongside (or instead of) the npm package.
- Trigger "Update Codex" → npm exits rc=0, `codex --version` unchanged → reported failed/retry, never resolves.

## Suggestions

- Detect the codex install kind; if it's the standalone build, update it via **its own** updater, not `npm install -g`.
- Or: after updating, resolve the **actual on-PATH** `codex` and verify against that. If npm updated a different install than the one on PATH, say so explicitly ("npm codex is now X, but the codex on PATH is a standalone build at <path> — update that instead") rather than a generic "not complete / retry" loop.

## Workaround

Repoint the on-PATH codex at the npm-managed binary (or update the standalone build directly).

scoop:
Fixes #1560.

## Problem

"Update Codex" ran `npm install -g @openai/codex`. When the on-PATH `codex` is the **standalone** build (`~/.local/bin/codex` → `~/.codex/packages/standalone/...`), npm updated a *shadowed* npm copy while the PATH binary stayed put. Shepherd's post-update `codex --version` re-read saw no change → "Update did not complete (still on X)" → an indefinite retry loop.

## Fix

- **Update via codex's own `codex update`.** It is install-kind-aware — on an npm install it reruns `npm install -g @openai/codex`; on a standalone install it reruns the standalone installer, repointing the on-PATH binary. So it updates the *same* binary Shepherd verifies, closing the mismatch. (No regression for npm users; codex reruns the identical npm command for them.)
- **Probe before invoke.** We only call `codex update` if `codex --help` advertises the subcommand (`--help` prints help and exits, never runs an agent). A codex too old for the subcommand can therefore never have `update` misinterpreted as an interactive agent prompt.
- **npm fallback gated on version NON-advancement**, not exit code. If the subcommand is absent *or* the version didn't advance (`after == before`), fall back to `npm install -g` — so an old codex that silently no-ops still falls back and **old-npm hosts never regress**. Runs with `CODEX_NON_INTERACTIVE=1`.
- **Success by advancement, not npm-latest match.** A standalone `codex update` targets the standalone channel, whose latest need not byte-match npm-latest; judging success by "did the on-PATH version go up?" avoids a false loop after a real standalone update, and reports the actually-installed version.
- **Honest non-convergence UX.** On the documented failure mode (PATH duplicates, or an already-latest no-op), the modal names the resolved on-PATH binary — "still reports {current} at {path}, may be a separate install" — instead of a blind retry. The audit log also gains portable diagnostics (`command -v`, `readlink`, `type -a codex`).

## Verification

- Root: `bun run lint` clean, `bun run test` → 6274 pass / 0 fail (incl. 20 in `codex-update.test.ts`).
- UI: `check:i18n` parity (2742 keys each), `svelte-check` 0 errors, browser 1071 pass, node 2034 pass.
- **Empirical end-to-end** against fake codex binaries (no real codex touched):
  - standalone-style codex with `update` → probe runs `codex update`, version advances 0.142.5→0.143.0, reflected immediately, no npm fallback.
  - old codex without `update` → probe skips `codex update` entirely, npm fallback fires (fake npm invoked with `install -g @openai/codex`).
  - Symlink-repoint experiment confirmed a fresh `execFileSync` reflects a repointed on-PATH symlink with no cross-exec cache (why Shepherd's re-read sees a standalone update at once).

This is a `fix:` touching only server logic, message values, a comment/branch in one component, and tests — no user-facing feature, so no feature-announcement entry. [no-feature-entry]


scoop (commented):
## Summary

The PR replaces the blind `npm install -g @openai/codex` with codex's own install-kind-aware `codex update` (with an npm fallback), switches the success test from `after === npm-latest` to *version advancement*, and surfaces the on-PATH binary so the modal can explain a stuck update instead of looping. This directly addresses the originating issue (npm updating a shadowed copy while the standalone binary on PATH stays put). The implementation is coherent and the change is well-tested.

### Verified
- **Signature change threaded everywhere.** `buildUpdateScript` gained a required `codexBin` param; the only non-test caller (`src/codex-update.ts:218` `defaultRunUpdate`) passes `config.codexBin`, and all test call-sites were updated (`test/codex-update.test.ts`). `config.codexBin` exists (`src/config.ts:457`). Verified via grep — no stale 3-arg callers.
- **Version source is consistent.** The shell script reads `$CX --version` where `CX=config.codexBin`, and the TS success check reads via `versionRunner` which also defaults to `execFileSync(config.codexBin, ["--version"])` (`src/codex-update.ts:198-200`). Same binary on both sides.
- **Success logic.** `ok = !!after && !!from && compareSemver(after, from) > 0` (`src/codex-update.ts:315`); `compareSemver` (`src/herdr-update.ts:11-19`) does a numeric 3-part compare, so advancement detection is correct. On success `to: after` (actual installed version) flows to `codexupdate_done_ok({latest: done.to ...})` — the message now shows the real version.
- **Result shape parity.** `onPathBinary?: string | null` added to both `src/codex-update.ts:140` (CodexUpdateResult) and `ui/src/lib/types.ts:1242`, and the whole result object is emitted verbatim over `codex-update:done` (`src/index.ts:2207`), so the field reaches the modal.
- **i18n parity.** New key `codexupdate_done_fail_stuck` present in BOTH catalogs (`ui/messages/en.json:2308`, `ui/messages/de.json:2308`), both using the same `{current}` and `{path}` placeholders the Svelte call passes (`CodexUpdateModal.svelte:103-106`). No other catalogs exist (`ui/messages/` = en, de only).
- **Modal branch order** (`CodexUpdateModal.svelte:100-108`): ok → stuck (onPathBinary truthy) → generic fail. Degrades gracefully to the generic message when `onPathBinary` is null.
- **Shell safety.** Log-path and binary are single-quoted with the `'\''` escape; versions are sanitized (`sanitizeVersion`) before interpolation — injection test still passes. Fallback is gated on `has_update==0 || after==before` (version, not exit code), matching the documented intent.

### Non-blocking observation (see findings)
The modal's stuck-message diagnostic depends on `defaultResolveOnPath(config.codexBin)`, a scan of Node's `process.env.PATH`. In the exact scenario this PR targets — a standalone codex at `~/.local/bin/codex` — that directory is frequently added to PATH only via a login-shell profile. The update *script* resolves the path correctly (it runs under `bash -lc`, `src/codex-update.ts:224`, so profile PATH is loaded), but the Node process that runs `defaultResolveOnPath` may not have `~/.local/bin` on its PATH. When that happens `resolveOnPathBinary()` returns null and the modal shows the *generic* `codexupdate_done_fail` rather than the new stuck message — the headline UX benefit silently degrades in its own target case. The rich log block still captures the path, so this is a graceful degradation, not a correctness break.

### Could not verify (external, not in this repo — assumed as written, NOT counted as findings)
- That `codex update` exists as a subcommand and that `codex --help` lists it as a line matching `^[[:space:]]+update[[:space:]]` (the `--help` probe regex). If codex ever renders it as `update, up  ...` the char after `update` is `,`, not whitespace, and the probe would miss it — but codex's actual help format is not in this repo.
- That `CODEX_NON_INTERACTIVE=1` is the correct env var to make the standalone installer run unattended.
These are inherent external dependencies of the feature; recorded as limitations, not defects.

- The update badge compares the installed version against the **npm** registry latest; for a standalone install on a different release channel this can show a persistent false-positive "update available" even after the user is at their channel's newest. This is inherent to `check()` comparing against npm and predates this PR (the new stuck message partially mitigates the confusion).

scoop:
Addressed the on-PATH resolution finding (commit 11fb6a62): the stuck-update diagnostic now prefers the **login-shell-resolved** path the script itself logs (`>>> codex-update: on-PATH codex:` via `command -v` under `bash -lc`), captured from the update stream in a new `runUpdateCapturingPath` helper. The Node `process.env.PATH` scan remains only as a fallback when that line wasn't emitted (e.g. the child died early), so a standalone `~/.local/bin/codex` that's only on PATH via a shell profile is now named correctly instead of dropping to the generic message. Added a unit test asserting the logged path wins over the Node scan. (Refactored the outcome logic into a `settleUpdate` helper to keep `runOnce` under the complexity gate.)

scoop (commented):
## Re-review of the `codex update` migration (issue #1560)

The PR replaces the unconditional `npm install -g @openai/codex` with codex's own install-kind-aware `codex update` subcommand (npm-gated fallback), switches the success test from an exact target match to version ADVANCEMENT, and surfaces the on-PATH binary on non-convergence so the modal stops looping blindly. This directly matches the issue's suggested fixes.

### Prior finding — CONFIRMED ADDRESSED

Previous revision raised that `defaultResolveOnPath` scans Node's `process.env.PATH` and would miss a standalone `~/.local/bin/codex` present only via a login-shell profile, so the stuck-diagnostic would be lost exactly where it's needed.

The diff now resolves the on-PATH binary from the script's OWN login-shell output: the script logs `command -v` under `bash -lc` as `>>> codex-update: on-PATH codex: <path>` (src/codex-update.ts:80-81), `runUpdateCapturingPath` captures that line (src/codex-update.ts:294-306), and `settleUpdate` prefers it over the Node scan via `onPathFromLog ?? this.resolveOnPathBinary()` (src/codex-update.ts:288). The `apply(): stuck message prefers the login-shell path the script logged` test locks this precedence in (test/codex-update.test.ts:486-503). This is the exact remedy the prior finding suggested. Not re-raised.

### Verification performed

- **Success criterion**: `settleUpdate` uses `compareSemver(after, from) > 0` (src/codex-update.ts:287). Verified `compareSemver(a,b)>0 ⇔ a>b` against src/herdr-update.ts:11-19 — so the check is true advancement, channel-agnostic. Matches the standalone-channel-skew test (test/codex-update.test.ts:455-467).
- **Marker capture correctness**: `CODEX_UPDATE_LOG_PREFIX = ">>> codex-update:"` (src/codex-update.ts:19); the capture marker `">>> codex-update: on-PATH codex:"` (src/codex-update.ts:295) matches the echoed line at src/codex-update.ts:81, and is distinct from the sibling `symlink target:` / `codex on PATH:` lines (src/codex-update.ts:82,86), so no false match. `<not found>` is correctly filtered to null (src/codex-update.ts:301).
- **Signature change**: `buildUpdateScript` now requires `codexBin` (src/codex-update.ts:52-59). Sole non-test caller `defaultRunUpdate` passes `config.codexBin` (src/codex-update.ts:222); every test call updated (test/codex-update.test.ts). `config.codexBin` exists (src/config.ts:457, defaults to `"codex"`). The separate `src/herdr-update.ts` `buildUpdateScript` is unrelated.
- **Result plumbing**: `onPathBinary` added to both `CodexUpdateResult` shapes — src/codex-update.ts:140 and ui/src/lib/types.ts:1237-1240 — and `onDone` emits the whole result object (src/index.ts:2207), so the field reaches the modal, which reads `done.onPathBinary` (CodexUpdateModal.svelte:102-105).
- **i18n parity**: `codexupdate_done_fail_stuck` present in BOTH catalogs with identical `{current}`/`{path}` placeholders (ui/messages/en.json:2308, ui/messages/de.json:2308); the Svelte call passes exactly `{ current, path }` (CodexUpdateModal.svelte:103-106). Only en/de catalogs exist. Parity holds.
- **Shell safety**: `codexBin` is single-quoted with the classic `'\''` escape (src/codex-update.ts:67); versions still sanitized (test/codex-update.test.ts:63-67). Fallback gates on version non-advancement (`[ "$after" = "$before" ]`), not exit code, so a silent-no-op old codex still falls back to npm (src/codex-update.ts:97).

No blocking issues found.

- src/codex-update.ts:287-288 — the stuck message can pair a mismatched version and path IF Node's `process.env.PATH` resolves a different `codex` than the login shell: `{current}` (=`after`) comes from the Node-PATH binary via `versionRunner`, while `{path}` (=`onPathFromLog`) is the login-shell binary. In the issue's actual target scenario both PATHs resolve the same `~/.local/bin/codex`, so the message is coherent; the mismatch only arises in the pathological two-different-installs case, where the message is still directionally useful ("update the one at that path directly"). Not a defect on the path this PR targets.

Issue #1567, opened 2026-07-08
Port herdr `send` (steer) write over the socket (deferred from #1553)


scoop:
Follow-up to #1553 (which ports `start`/`stop`/`relabel`/`closeTab` over the herdr socket). The **`send`** write was deliberately deferred because it — uniquely among the write surface — drives a large synchronous control-flow cascade, and coloring it async balloons the blast radius far beyond the other four methods.

## Why `send` is separate
`send` has only two callers, but both are wired deep into boolean-returning control flow:
- `haltAll` (e-stop) — `service.ts:3812`.
- `sendSteerTo` (`service.ts:3858`) — the steer delivery path, transitively pulled async through: `replyToLive` → `reply` (public, `Promise<boolean>`), `steerWithEpicNotice` → `operatorReply`/`broadcast`, and `reply`'s **boolean-consuming** consumers where a `void`-wrap would silently break logic:
  - `automerge.ts:371` — `if (service.reply(…))` gates a persistent `setAutoMergeState` write.
  - `plan-gate.ts:559`/`690` — `reply` flips `PlanGate.resume(session): boolean` → async, cascading to the `planGate.resume` DI types (`server.ts:360`, `index.ts:2375`) and `server.ts:2710` (`const ok = (await …) ?? false`, else `ok` is an always-truthy Promise).
  - `server.ts:2310`/`2406` — `const ok/delivered = service.reply(…)`.

## Scope
- Route `send` over the socket `agent.send {target,text}` in `SocketHerdrDriver` (behind the existing `SHEPHERD_HERDR_SOCKET` flag; CLI fallback retained).
- Async-color the full `send` cascade above; every boolean-consuming site must `await` (never `void`).
- Add the steering ordering test: assert `sendSteerTo` issues the CR `send` only after the bracket-paste `send` **resolves** (ordering across two socket round-trips), not just param equality.
[...]

Issue #1553, opened 2026-07-07
Port herdr write surface over the socket (continue #1529 opportunity #2)


scoop:
Follow-up to the first herdr-socket adoption slice (#1529, PR adopting the **read** surface).

## Context
The opt-in `SocketHerdrDriver` (behind `SHEPHERD_HERDR_SOCKET=1`, default off) currently routes only the **async read** surface over herdr's native Unix socket — `listAsync` / `readAsync` / `paneForegroundProcs` — and **delegates every sync + write method to the wrapped CLI `HerdrDriver`**. This issue tracks porting the **write surface** over the socket too.

## Scope
Move these off the CLI onto socket JSON-RPC (`agent.send`, `tab.close`, `agent.rename`+`tab.rename`, `workspace.*`/`tab.create`/`agent.start` orchestration):
- `send`, `stop`, `relabel`, `closeTab` — straightforward one/two-call ports.
- `start` — the multi-step `workspace.list`/`create` → `tab.create` → `agent.start` orchestration + the `agent_name_taken` collision-retry (herdr.ts). The socket `agent.start` takes `{name, argv, cwd, tab_id, workspace_id, focus, env}` directly.

## Why it's deferred
These methods are **synchronous** in `HerdrDriver` and called synchronously across ~43 sites (`stop` 20, `start` 15, `send`/`relabel`/`closeTab` ~8). A socket is async, so porting them means either converting each to an async sibling (à la `list`→`listAsync`) and repointing callers, or converting the sync surface to async wholesale. That blast radius was intentionally kept out of the read-surface PR.

## Notes for the implementer
- herdr's socket is **one request/response per connection** (it FINs after each reply) — `HerdrSocketClient` already opens a fresh connection per `request()`; reuse it.
- Verify the exact `agent_name_taken` error `code` on the socket error channel before wiring collision-retry.
- Keep everything behind the existing `SHEPHERD_HERDR_SOCKET` flag + bounded protocol gate; retain CLI fallback.

Broader opportunities (#1 node-pty via `terminal session observe/control`, #3 `events.subscribe` push, #4 `pane.report_agent`, #5 schema codegen, #6 remote) remain tracked in #1529.

scoop:
✅ PR #1568 merged.

scoop:
Completed. The herdr write surface was ported over the socket across two merged PRs:

[...]

Issue #1529, opened 2026-07-07
Adopt herdr's native socket API + terminal session observe/control (preview)


scoop:
## Context

Herdr's author replied to our thank-you/pitch and explicitly invited us to build on the new **native API surface** they're rolling out for third-party clients tracked at https://herdr.dev/docs/preview/persistence-remote/#direct-terminal-attach - specifically `herdr terminal session observe` and `herdr terminal session control`."

Shepherd is exactly the kind of third-party client this is aimed at. This issue captures what's in the **preview docs today**, maps it onto how we integrate with herdr **right now**, and lays out what we should adopt (and in what order). These are **preview/master** APIs — the point of this issue is to start tracking them and prototype behind our existing seams, **not** to rip out the current CLI path yet.

Docs read:
- Socket API — https://herdr.dev/docs/preview/socket-api/
- Direct terminal attach / observe / control — https://herdr.dev/docs/preview/persistence-remote/#direct-terminal-attach
- Session state (detach/reattach, resume, handoff) — https://herdr.dev/docs/preview/session-state/

---

## How we integrate with herdr today

Two independent seams, both assuming a **local `herdr` binary** (`config.herdrBin`, `HERDR_BIN`):

1. **Control plane + status scraping — CLI-shelling.** `HerdrDriver` (`src/herdr.ts:219-530`) runs `execFileSync` / `execFile` per call (10s timeout, one subprocess spawn per operation) behind a thin `Runner` / `AsyncRunner` seam (`src/herdr.ts:40-87`). Subcommands: `agent list/start/send/read/rename`, `tab/pane/workspace *`, `pane process-info`. The poller (`src/poller.ts`) **polls** `agent read`/`readAsync` on an interval to detect agent status. Consumers depend on narrow `Pick<HerdrDriver, …>` interfaces (`src/service.ts:153`, `src/poller.ts:282`) — so the driver is highly swappable.
2. **Live interactive terminal — node-pty.** The browser xterm is fed by spawning `herdr agent attach <id> --takeover` under **node-pty** in a **separate Node process** (`src/pty-attach.mjs:16`) — because node-pty is broken under Bun (`config.nodeBin` exists solely to launch this helper, `src/config.ts:441-444`). `PtyBridge` (`src/pty-bridge.ts`) pipes PTY stdout→WebSocket and WS→stdin. This seam is **not** routed through `HerdrDriver`.

## What the native API gives us

[...]

PR #1568, opened 2026-07-08
feat(herdr): port start/stop/relabel/closeTab over the socket (#1553)

scoop:
Continues the herdr-socket adoption (#1529 ported the **read** surface). This ports the herdr **spawn/teardown/rename** write surface — `start` / `stop` / `relabel` / `closeTab` — over herdr's native Unix-socket JSON-RPC when `SHEPHERD_HERDR_SOCKET=1` (default off), with the CLI retained as the flag-off / protocol-mismatch fallback.

**`send` is intentionally NOT ported here** — it uniquely colors a boolean-returning steer/reply cascade async (`sendSteerTo`→`reply`/`operatorReply`/`broadcast`→`PlanGate.resume`/`automerge`/server DI), a blast radius unrelated to the spawn/teardown writes. That port is tracked in **#1567**. So this `Refs #1553` (does not close it).

## What changed
- **`IHerdrDriver`**: `start`/`stop`/`relabel`/`closeTab` are now `Promise`-returning (wholesale async). `send` stays sync + CLI-backed.
- **`SocketHerdrDriver`** (flag ON): `start` orchestrates `workspace.list`/`[create]` → `tab.create` → `agent.start` (with `agent_name_taken` collision-retry) → `pane.close`, resolving the `HerdrAgent` **directly from the `agent.start` reply** — dropping the CLI's fragile `list().filter(cwd).at(-1)` heuristic on the socket path. `stop`/`relabel`/`closeTab` map to `tab.close` / `agent.rename`+`tab.rename`. A per-instance promise-chain **serializer** restores the mutual exclusion the blocking path gave `start`.
- **`HerdrDriver`** (flag OFF): the four writes move to the non-blocking `asyncRunner` (like the read siblings) — writes no longer block Bun's loop for anyone. `send` unchanged.
- **Shared `herdr.ts` helpers**: `buildWrappedArgv` (byte-identical spawn across both drivers), `parseAgentInfo`, `createSerializer`, and `isNameTakenError` extended to read `HerdrSocketError.code`.
- **Callers**: the async spawn is awaited through the transient-run services (distiller/optimizer/merge-suggest/review/…) and ~40 call sites; teardown wrappers (`reapRun`/`reapTransientByLabel`/`usage-probe.sweep`/`stopAll`) go async.

## Verification
- Wire shapes + the literal `agent_name_taken` error `code` confirmed against **live herdr 0.7.3 (protocol 16)**.
- New unit tests: socket orchestration, collision-retry + tab-rollback, the `start` serializer (two concurrent starts don't double-create a workspace), start-resolution (resolves from `agent_started`, no re-list), and `parseAgentInfo ≡ parseAgents` / `buildWrappedArgv` parity.
- End-to-end smoke through the **real** driver in both flag states (start → list → relabel → stop) against live herdr.
- Root `bun run lint`, `tsc --noEmit`, and `bun test test/` all green (the only residual `test/` failures are the pre-existing `persistConfig`/`$state` UI-env tests, unaffected by this change).
- Statement-position await coverage verified via a one-shot type-checked `no-floating-promises`/`no-misused-promises` pass over the diff; the durable repo-wide gate is deferred to a prep PR for #1567 (measured: ~23 pre-existing violations across ~6 files).

[no-feature-entry] server-only plumbing behind an existing flag — no user-facing UX.


scoop:
## Summary

This PR ports the herdr write surface (`start`/`stop`/`relabel`/`closeTab`) from sync to async across both `HerdrDriver` and the opt-in `SocketHerdrDriver`, and repoints the ~40 call sites. The socket-driver spawn/teardown port itself (`src/herdr-socket-driver.ts`) is well built — serialized `start`, orphan-tab rollback, name-collision retry keyed on `HerdrSocketError.code`, and `parseAgentInfo`/`buildWrappedArgv`/`createSerializer` factored cleanly out of `herdr.ts`. I verified `isNameTakenError`'s new `"code"` field against `HerdrSocketError.code` (src/herdr-socket-client.ts:12-17, set to the exact code string, so `String(code).includes("agent_name_taken")` matches — src/herdr.ts:209-216).

But the sync→async conversion of the background-agent services introduced a real concurrency-cap / dedup regression, because their `inflight` slot is now reserved only **after** `await herdr.start(...)` rather than before the (previously blocking, synchronous) spawn.

## Blocking: inflight slot reserved after the await → cap + dedup broken

`DistillerService`, `OptimizerService`, and `MergeSuggestionService` all follow the same shape: `enqueueOrBegin` checks `inflight.has(key)` and `inflight.size < maxConcurrent` **synchronously** (distiller.ts:203-205, optimizer.ts:148-149, merge-suggest.ts:186-190), then calls `begin`, which now does `await this.deps.herdr.start(...)` and only afterwards `this.inflight.set(key, …)` (distiller.ts:254/266, optimizer.ts:219/231, merge-suggest.ts:244/256).

Before this PR `begin` was fully synchronous and `herdr.start` blocked, so in a fire-and-forget loop each call ran to completion (including `inflight.set`) before the next — the cap and the `inflight.has` dedup held. Now the synchronous prefix of every `begin` runs up to its first `await` before any of them sets `inflight`, so the guards all see a stale (empty) `inflight`.

The daily sweep fires these fire-and-forget in a loop over repos (src/index.ts:2040, 2044, 2045): `for (const repo of listRepos(...)) void distiller.consider(repo.path);` etc. With more than `maxConcurrent` repos holding work (distiller 3, merge-suggest 2 — distiller.ts:139, merge-suggest.ts:108), all of them pass `inflight.size < maxConcurrent` before any reserves a slot, so every repo spawns at once and the queue is never used — the cap the queue exists to enforce is bypassed. The PR's own comments (distiller.ts:206-207, optimizer.ts:470-471) claim the `await` "preserv[es] the concurrency cap the blocking sync path guaranteed," but the `await` only serializes an *awaiting* caller (the drain loop, which is correct); it does nothing for the fire-and-forget fan-out.

The optimizer case is worse than a soft cap: `retireRepoCandidates` fires `void optimizer.optimizeOne(rule.id)` in a loop, and every rule belongs to the same `repoPath` (src/learnings-lifecycle.ts:408-413, reached from `runAutoRetire` in the daily sweep, index.ts:2050). `optimizeOne` → `enqueueOrBegin(l.repoPath, [id])` is keyed by `repoPath`, so when ≥2 flagged rules race, `inflight.has(repoPath)` is still false for each, they all `begin`, and the later `inflight.set(repoPath, …)` **overwrites** the earlier entry — the first spawned agent is dropped from `inflight`, so `tick()`/`finalize()` never stops it or removes its scratch dir. That is a leaked claude agent + leaked scratch dir per collision.

The fix is the pattern the reviewer/plan-gate services already use: reserve synchronously *before* the await. `ReviewService`/`PlanGateService` call `this.starting.add(session.id)` before their async `begin` and check `inflight.has || starting.has` (src/review.ts:269 + 235; src/plan-gate.ts:250 + 242), so they do not have this race. Distiller/optimizer/merge-suggest need an equivalent synchronous reservation (a placeholder in `inflight`, or a `starting` set) claimed before `await herdr.start`.

## Verification notes / limitations

- I traced every `herdr.start/stop/relabel/closeTab` caller in `src/` (grep, all 40 sites): each is either `await`ed inside an existing try/catch or `void`-ed with the promise's rejection handled — `relabel`/`closeTab` are internally guarded and never reject, and `stop`'s only reject surface (`listAsync`) matches the pre-PR synchronous throw surface for awaited callers, so no new unhandled-rejection paths there.
- Could not verify herdr 0.7.3's live socket `agent.start` reply schema (that `result.agent` carries `terminal_id`/`tab_id`) or that the `agent_name_taken` error arrives on the `code` channel — herdr is not in this repo. Assumed as written per the comments; the socket path is behind `SHEPHERD_HERDR_SOCKET=1` (default off) + a protocol gate, and the CLI path is unaffected.

- src/herdr-socket-driver.ts:135 — if a future herdr build omits `agent` from the `agent.start` reply, `res?.agent` is undefined and `parseAgentInfo` returns an all-empty `HerdrAgent` (terminalId "") rather than throwing; dormant today behind the off-by-default socket flag and the current 0.7.3 reply shape.

- src/usage-probe.ts:109-112 — a failed `agent start` leaves a tab with no agent that `sweep()` (name-matched on live agents) cannot see; called out in an existing comment as an unhandled gap.

scoop:
## Re-review

The write-surface port itself is sound: `SocketHerdrDriver.start/stop/relabel/closeTab` mirror the CLI orchestration (verified `src/herdr-socket-driver.ts:101-233` against the CLI `src/herdr.ts:516-684`), `createSerializer` restores start atomicity, `request<T>` returns the unwrapped `result` object so `created?.tab?.tab_id` / `res?.agent` read correctly (verified against `src/herdr-socket-client.ts:119`), and the ~43 caller sites were repointed to `await`/`void` consistently.

However, this is a re-review, and **none of the four previously-raised concurrency findings were actually fixed** — the diff only made the methods `async` and reworded the comments. In all three services the `inflight` slot is still reserved AFTER `await herdr.start`, while the dedup/cap check is synchronous, and the daily-sweep / same-repo callers still fire the work fire-and-forget (`void`). The suggested fix (a synchronous reservation, as `ReviewService`/`PlanGateService` do with `starting.add` — see `src/review.ts:1152`, `src/plan-gate.ts:717`) was not applied.

Ground truth verified for each:

1. **optimizer** — `enqueueOrBegin` dedup `this.inflight.has(repoPath)` is synchronous (`src/optimizer.ts:148`); `begin` still does `this.inflight.set(repoPath, …)` at `src/optimizer.ts:231`, AFTER `await this.deps.herdr.start(…)` at `src/optimizer.ts:219`. The caller loops `void optimizer.optimizeOne(rule.id)` over same-repo candidates (`src/learnings-lifecycle.ts:413`, inside the `for (const rule of candidates)` loop at `:408`). Each `optimizeOne` runs synchronously to the `herdr.start` await before any `inflight.set`, so the second same-repo call passes the `has` check and double-`begin`s; the second `inflight.set` overwrites the first → leaked agent + scratch dir. Unaddressed.

2. **distiller** — `inflight.size < this.maxConcurrent` (default 3) check is synchronous at `src/distiller.ts:205`; `inflight.set` is at `src/distiller.ts:266`, AFTER `await this.deps.herdr.start(…)` at `src/distiller.ts:254`. Daily sweep `for (const repo …) void distiller.consider(repo.path)` at `src/index.ts:2040` lets many repos pass the cap check before any reserves. Unaddressed.

3. **merge-suggest** — `inflight.size < this.maxConcurrent` (default 2, `src/merge-suggest.ts:108`) check synchronous at `src/merge-suggest.ts:187`; `inflight.set(key, …)` at `src/merge-suggest.ts:256`, AFTER `await this.deps.herdr.start(…)` at `src/merge-suggest.ts:244`. Daily-sweep fan-out `void mergeSuggest.consider(repo.path)` + `void mergeSuggest.considerCrossRepo()` at `src/index.ts:2044-2045` exceeds the cap. Unaddressed.

4. **comments** — `src/distiller.ts:206-207` and the optimizer `enqueueOrBegin` comment (`src/optimizer.ts:150-151`) still claim the `await` preserves the ordering/cap the blocking sync path guaranteed; on the fire-and-forget fan-out it does not (the `void` callers never await), so the claim remains misleading until the synchronous reservation lands.

Note: the `drainQueue` paths (`src/optimizer.ts:268`, `src/distiller.ts:302`, `src/merge-suggest.ts:294`) DO await `begin` serially, so the cap holds there — the defect is confined to the fire-and-forget entry paths above, exactly as the prior findings described.

Everything else in the async conversion checked out and I found no new correctness issues.

## Could not fully verify
- `agent.rename` / `tab.rename` / `pane.close` / `tab.create` socket method names and the `agent_name_taken` error `code` are herdr's live wire contract, not present in this repo; assumed as written (comments cite live herdr 0.7.3). `src/herdr-socket-client.ts` is not in this diff, so its transport behavior was taken as given.

scoop (commented):
## Re-review — herdr write surface over the socket (#1553)

### Prior findings — all addressed

1. **optimizer.ts double-`begin` / slot leak** — FIXED. `begin` now builds the `InFlight` entry and calls `this.inflight.set(repoPath, entry)` **synchronously before** the `await herdr.start` (optimizer.ts:234, await at :237), with `this.inflight.delete(repoPath)` on spawn failure (:240). Verified the guard now holds under the fire-and-forget fan-out: `optimizeOne` runs synchronously up to `await enqueueOrBegin`, which runs synchronously up to `await begin`, which runs synchronously up to `inflight.set` before its first yield — so the second `void optimizeOne(...)` (learnings-lifecycle.ts:413) hits `inflight.has(repoPath)` (optimizer.ts:153) and returns. No early return sits between `inflight.set` and the `await` (all early returns — api-key, empty targets, write failure — precede it at :192/:200/:206).

2. **distiller.ts cap bypass** — FIXED. Same synchronous reservation (distiller.ts:260-267, await at :270, release at :273). All early returns precede the reservation (:220 api-key, :241 write). The daily sweep `void distiller.consider(repo.path)` (index.ts:2040) now cannot let >`maxConcurrent` repos pass `inflight.size < maxConcurrent` (distiller.ts:205) before a slot is reserved.

3. **merge-suggest.ts cap bypass** — FIXED. Same reservation (merge-suggest.ts:250-260, await :263, release :266); all early returns precede it (:211 cross shortlist, :219 api-key, :231 write). Covers the sweep fan-out over repos + cross (index.ts:2044-2045).

4. **Misleading comments** — FIXED. `enqueueOrBegin` comments now state the await "is NOT what enforces the cap" (distiller.ts:254-259, optimizer.ts:219-225, merge-suggest.ts:244-249); the `tick()` drain comment (optimizer.ts:282-283) is now accurate because `begin` reserves synchronously.

### Verification of the broader sync→async conversion

- **All write-surface call sites** resolved via grep of `herdr.(start|stop|relabel|closeTab)` across `src/**`: every one is either `await`ed or explicitly `void …catch(()=>{})` fire-and-forget. Fire-and-forget stops (distiller.ts:324, optimizer.ts:290, merge-suggest.ts:324, review.ts:1157, plan-gate.ts:722, recap.ts:428, herd-digest.ts:205) rely on `stop`/`relabel`/`closeTab` being internally guarded async fns that never reject — confirmed in both drivers (HerdrDriver.stop/relabel/closeTab at herdr.ts:642-682 wrap every step in try/catch; SocketHerdrDriver equivalents at herdr-socket-driver.ts:186-232 do the same).
- **Service methods that became async** (`consider`/`distillNow`/`optimizeOne`/`optimizeAllFlagged`/`mergeNow`/`considerCrossRepo`) — grepped all callers: the only external callers are the `void`-guarded daily-sweep (index.ts:2040-2045) and server routes (server.ts:1612-1614, 1714), plus `learnings-lifecycle.ts:413`. Same-named `consider` on autopilot/doc-agent/plan-gate/review are unrelated methods.
- **`reapTransientByLabel` → async**: now `void`-ed at its boot/deferred and reap-orphan call sites (index.ts:1955-1958, distiller.ts:297, optimizer.ts:264, merge-suggest.ts:289, standalone-critic.ts:640). Never rejects (whole loop wrapped in try/catch, transient-tab-reaper.ts:42-54).
- **`reapRun` → async** (critic-core.ts:525-534) now `await`ed by review.ts:459/755 and standalone-critic.ts:612.
- **`isNameTakenError`** now also inspects `.code` (herdr.ts:213), matching `HerdrSocketError.code: string` (herdr-socket-client.ts:12-17, populated at :115-117). The exact literal `code` value on live herdr is a stated assumption below.
- **Serializer** (`createSerializer`, herdr.ts:328-337): `tail.then(fn, fn)` correctly runs each submitted fn after the prior settles (success or failure), returns the real result to the caller, and swallows only in the internal tail — restoring the mutual exclusion the blocking sync `start` had. Each driver owns its own chain (herdr.ts:382, herdr-socket-driver.ts:35); SocketHerdrDriver.start does not double-serialize (it calls its own `startImpl`, not `cli.start`).
- **`terminalId=""` reservation window**: confirmed `tick()` skips a freshly-reserved run — `readProposals(dir)` returns null and `timedOut` is false, so it `continue`s (distiller.ts:304-306), so `finalize` never observes the empty id in practice.

### Note — `send` descoped from the issue's stated scope

The originating issue lists `send` in the write surface to port, but this PR keeps `send` CLI-backed and defers its socket port to a follow-up (#1567), with a documented rationale (it colors a boolean-returning steer/reply cascade async — a blast radius unrelated to the spawn/teardown writes). This is a deliberate, tracked, reasonable incremental slice rather than an omission; `send` remains fully functional via the CLI delegate (herdr-socket-driver.ts:88-90). Flagged for the reader, not blocking. The config doc was updated to match (docs-site/.../configuration.md:27).

- The live herdr 0.7.3 socket schema — the exact `error.code` string for a name collision, and that `agent.start`'s `agent_started` reply carries `terminal_id`/`tab_id` directly, and the `agent.rename`/`tab.rename`/`tab.create`/`workspace.*` param names (`target`, `tab_id`, `pane_id`, `cwd`/`label`/`focus`). These are asserted "confirmed against live herdr 0.7.3" in-code (herdr.ts:246-251, herdr-socket-driver.ts:100-118); no ground truth exists in this repo to check them against. Assumed as written.

No blocking issues found.

PR #1575, opened 2026-07-09
feat(herdr): port `send` (steer) over the socket (#1567)

### LINKED ISSUE #1567: Port herdr `send` (steer) write over the socket (deferred from #1553)

scoop:
Closes #1567. Follow-up to #1553, which deliberately deferred `send` because it uniquely colors a boolean-returning steer/reply cascade async.

## What changed

`SocketHerdrDriver.send` now routes `agent.send {target,text}` over the socket, behind the existing `SHEPHERD_HERDR_SOCKET` flag + protocol gate, with the CLI fallback retained (same shape as #1553 / #1529). The CLI `send` also moves from `execFileSync` to `asyncRunner`, so a steer no longer blocks Bun's loop.

The async cascade reached further than the issue's list. Beyond `reply` / `operatorReply` / `broadcast` / `PlanGate.resume` / `automerge`, it also pulled in `startPreview`, `releasePlanGate`, `haltAll`, the review auto-address loop, autopilot's steer, and the build-queue reminder's `sweep()`. Every boolean-consuming site `await`s — none were `void`-wrapped.

## Two findings worth review attention

**1. The type checker cannot see the worst sites.** `json(deps.service.broadcast(...))` and `json(deps.service.haltAll())` take `unknown`, so a bare Promise would serialize as `{}` — a broadcast reporting nothing delivered, an e-stop reporting nothing halted — with a clean `tsc` *and* a clean lint. These were caught by the hand-audit the issue asked for, not by any gate.

**2. Async `send` silently breaks steer atomicity** (not called out in the issue). A steer is two writes: bracket-paste, then CR. Sync `send` made that pair uninterruptible for free. Once it's a promise, two overlapping steers to one pane can interleave as `paste-A, paste-B, CR-A, CR-B` and submit a corrupted merge of both texts. Fixed with a per-driver serializer (`createSerializer`, as #1553 uses for `start`) wrapping the pair in `sendSteerTo`. The same reasoning forces the epic-notice one-shot to be **claimed before its await** and rolled back on failure — otherwise two concurrent epic-intent steers both inject the notice.

`haltAll` deliberately **bypasses** that serializer: an e-stop must never queue behind the very steers it exists to cancel (a wedged send would hold the FIFO up to herdr's timeout). A lone ESC landing between a concurrent steer's paste and CR is harmless — it lands inside pasted text of a turn being interrupted anyway. Rationale is in the code.

## The lint gate

Lands the durable type-checked `@typescript-eslint/no-floating-promises` + `no-misused-promises` pass, scoped to `src/**/*.ts` (`projectService`). Not repo-wide: the root tsconfig covers server code only, tests legitimately fire-and-forget, and widening to `ui/` means wiring the svelte parser project for a much slower lint — not worth it for the surface this gate protects.

All 24 pre-existing violations are **fixed, no disables**. One was a real bug, not a style nit: `scheduleCalibrate` did `await calibrate(); scheduleCalibrate();` inside a timer callback, so a rejected `/usage` probe killed the self-rescheduling loop and stopped usage calibration permanently and silently. It now reschedules in a `finally`.

## Tests

Adds the steering-ordering test the issue asks for (`test/steer-ordering.test.ts`) — asserting the CR send starts only after the paste send **resolves**, and that two concurrent steers never interleave. Ordering across round trips, not param equality. Plus socket/CLI `send` driver coverage (`send` had none at the driver level) and two new epic-notice invariants (a rejecting send must not burn the one-shot; concurrent epic steers inject exactly once).

Each new test was **mutation-verified** rather than trusted for being green:

| Mutation | Test that dies |
|---|---|
| drop `await` on the paste send | both ordering tests |
| remove the serializer only | interleaving test only |
| claim epic one-shot after the await | concurrency test only |
| drop the one-shot rollback | reject test only |

## Verification

`bun run typecheck` clean · `bun run lint` clean · `bun run test` → **6299 pass, 0 fail** (318 files).

**Not verified:** no running herdr in this worktree, so the `agent.send` wire shape against live herdr 0.7.3 / protocol 16 rests on the issue's claim and is exercised only through fakes. Worth one real steer through a socket-enabled herdr before merge.

[no-feature-entry] — server-only plumbing, ships no user-facing UX.


scoop (commented):
## What I verified

**The socket port itself.** `SocketHerdrDriver.send` issues `agent.send {target, text}` (`src/herdr-socket-driver.ts:91-93`) and no longer touches `this.cli`. `target` is the same param name the already-shipped `agent.read` call uses (`src/herdr-socket-driver.ts:48`), so the naming is internally consistent (the other socket methods that take a pane use `pane_id`/`tab_id`, and `agent.*` uses `target` — that split is preserved). `IHerdrDriver.send` is now `Promise<void>` (`src/herdr.ts:372`) and both implementations match: `HerdrDriver.send` (`src/herdr.ts:585-587`) routes through `asyncRunner`, whose default is `defaultAsyncRunner` (`src/herdr.ts:81-87`) and which throws on maintenance/non-zero exit exactly like the sync `defaultRunner` (`src/herdr.ts:68-70`) — so the "a failed steer rejects, never silently drops" contract holds on both paths. `SocketHerdrDriver` and `HerdrDriver` are the only `implements IHerdrDriver` in the tree (checked: `src/herdr.ts:380`, `src/herdr-socket-driver.ts:27`), and `herdr.send` has exactly three call sites, all in `src/service.ts` (3831, 3890, 3891), all awaited.

**Every boolean-consuming site awaits; none is `void`-wrapped.** I grepped and opened each one the issue enumerated:
- `src/automerge.ts:372` — `if (await this.deps.service.reply(...))` still gates `setAutoMergeState`; the enclosing method already awaited `resume` at line 371, so it was already async.
- `src/plan-gate.ts:563` (`applyChangesRequested`) and `src/plan-gate.ts:694` (`resume`) — both awaited; `resume` returns `Promise<boolean>` and its two consumers agree: the DI type at `src/server.ts:360` and the wiring at `src/index.ts:2400` (`resume: (s) => planGate.resume(s)`).
- `src/server.ts:2715` — `const ok = (await deps.planGate?.resume?.(s)) ?? false`. This is the exact always-truthy-Promise trap the issue called out, and the `await` is inside the optional-chain parens, so an absent `planGate` still yields `undefined ?? false`.
- `src/server.ts:2270` (`/go`), `2312` (answer-plan-questions), `2409` (preview setup steer), `2436` (`startPreview`) — all awaited.
- `src/index.ts:1099-1100` and `1151` — the `shouldDeferSteer` short-circuits still return `false` (not a Promise) before the `await`, and both DI types (`src/review.ts:140`, `src/plan-gate.ts:130`) were widened to `Promise<boolean>`.
- `src/review.ts:932`, `src/build-queue-reminder.ts:167`, `src/autopilot.ts:311` — awaited.

**Handlers that became async are still awaited by their dispatchers.** `handleSessionGo` sits in the `handleSessions` sub-list, which does `const res = await sub(ctx)` (`src/server.ts:3306`). `handleHalt` sits in `ROUTE_HANDLERS`, dispatched with `await handle(ctx)` (`src/server.ts:6580`). `approveBuildQueue` is returned directly from the `async handleBuildQueue` (`src/server.ts:3265`). So `json(await deps.service.haltAll())` (`src/server.ts:4774`) cannot serialize a Promise as `{}`.

**The steer serializer.** `createSerializer` is exported at `src/herdr.ts:313-323` and its tail swallows rejections while `run` propagates them to the caller — so a rejected steer surfaces at `reply()` without poisoning the queue. `#serializeSteer` (`src/service.ts:1485`) wraps only the paste+CR pair (`src/service.ts:3887-3892`); `haltAll` deliberately bypasses it (`src/service.ts:3831`). `test/steer-ordering.test.ts` asserts ordering on send *start* vs. previous send *resolution* (not just argv equality), which is what the issue asked for.

**The epic-notice one-shot.** Claiming `#epicNoticeSteered` before the `await` and rolling back on rejection (`src/service.ts:3904-3915`) is the right fix for the newly-introduced check-then-act window, and `test/epic-steer.test.ts` covers both the concurrent-double-inject and the reject-rollback cases.

**Lint gate.** `tsconfig.json` has no `include` and excludes only `node_modules`/`ui`/`extension`/`site`/`docs-site`, so `projectService: true` resolves every `src/**/*.ts` to the root project — the gate will actually type-check rather than error out on "file not found by the project service".

## Notes / findings

See the findings list. Nothing I found blocks: the cascade is correct, the ordering test is real, and no boolean consumer was silently `void`-ed.

One more thing I looked at and decided *not* to raise: `applyChangesRequested` (`src/plan-gate.ts:557-568`) now has an `await` between `buildGate` (which resets `answeredQuestionKeys: []`, `src/plan-gate.ts:659`) and its `putPlanGate`, so in principle a concurrent `/answer-plan-questions` write could be clobbered. But `consider()` dedups on `planHash`, so a second `changes_requested` finalize for a planHash that already has answered questions is not reachable in practice. Mentioning it only because the `src/server.ts:2313-2317` comment reasons about the *other* direction of this race and not this one.

## Could not verify

- The `agent.send` wire shape (`{target, text}`) against a live herdr 0.7.3 / protocol 16 — no herdr protocol schema is vendored in this repo. Assumed as written per the issue's note. It is at least consistent with `agent.read`'s use of `target` (`src/herdr-socket-driver.ts:48`).
- Bun's `Server.stop(true)` return type and `FileSink.write`/`flush` return types — `node_modules` is not installed in this worktree, so I could not open `@types/bun`. The `.catch()` in `src/preview.ts:521` and the `await proc.stdin.write(...)` in `src/usage-probe.ts:143` both depend on those; `tsc` in CI will settle it.

- `src/server.ts:1614-1615`: `void deps.distiller?.distillNow(dir)` and `void deps.optimizer?.optimizeAllFlagged(dir)` discard promises with no `.catch()`, so a rejection becomes an unhandled rejection. The new lint gate accepts `void` (its `ignoreVoid` default), so it will not flag these.
- `src/index.ts:1371,1373`: `recapService.sweep()` and `herdDigestService.sweep()` are fired-and-forgotten on the same 15s interval with no re-entrancy guard either — the same shape as the build-queue finding below, but predating this PR.

scoop:
**Findings 2, 3, 4 fixed in 73e0e157. Finding 1 needs no change — it was already fixed on this branch before the review, by a commit the critic's snapshot predates.**

**1 — docs (no change).** `docs-site/.../configuration.md:27` no longer contains the quoted sentence. Commit `60ddd4b7` ("docs: sync docs to recent source changes") landed on this branch after my initial push and already rewrote it to *"(issues #1529, #1553, #1567). Covers the async read surface plus the entire async write surface — the spawn/teardown/rename writes (`start`/`stop`/`relabel`/`closeTab`) and `send` (writing text to an agent's PTY)."* The `#1567` attribution the finding asks for is present. Verified against `origin/main` (which still has the old sentence) to confirm the diff is real and on this branch, not a local artifact.

**3 — sweep re-entrancy (real bug, fixed).** Confirmed with a failing test before touching the code: a parked steer plus a second overlapping sweep produced **2** reconcile steers where 1 was expected. `sweep()` now drops the tick when one is in flight (rather than queueing — this is an idle-debounced nudge, so the next tick re-evaluates against a fresh `now`; queueing would replay a decision made on stale state). The flag clears in a `finally`, so a rejected steer can't wedge the nudge off permanently — that path has its own test, and dropping the `finally` fails it.

**4 — preview sync guard (agreed, restored).** You're right that I narrowed the guard. Worth noting the failure mode is slightly broader than a bare sync `throw`: if `stop()` ever returned a non-promise, `.catch` itself would throw, hitting the same path. Both `release()` and `stopAll()` now keep the sync `try/catch` *around* the `void …catch()`. Two regression tests inject a synchronously-throwing `stop()`; without the guards, `release()` skips `onChange` and `stopAll()` leaves `listeners`/`slotOwner` populated — both fail on mutation.

**2 — lint scope (agreed, config now matches its comment).** Took the `ignores` option. The substantive defect was the inconsistency you identified: a floating promise was an error in `src/hold.test.ts` but fine in `test/hold.test.ts`, purely by file location. Added `ignores: ["**/*.test.ts"]` and verified both directions — an injected floating promise in `src/hold.test.ts` now reports nothing, while the same injection in `src/hold.ts` still errors, so the exclusion didn't silently disarm the gate.

Full suite: **6304 pass / 0 fail**; typecheck and lint clean. Each of the three new regression tests was mutation-verified — reverting its fix makes it fail, and only it.

Still outstanding from my PR description, unaffected by this review: the `agent.send` wire shape is exercised only against fakes here, so one real steer through a socket-enabled herdr (`SHEPHERD_HERDR_SOCKET=1`) is worth doing before merge.

scoop:
## What the PR does, and what I verified

The core port is correct and internally consistent.

- `SocketHerdrDriver.send` now issues `agent.send {target,text}` over the socket (`src/herdr-socket-driver.ts:91-93`), and no longer delegates to `this.cli.send`. The remaining CLI delegations are exactly `list`/`tabs`/`panes`/`read` (`src/herdr-socket-driver.ts:63-80`), which is what the updated class doc (`src/herdr-socket-driver.ts:17-25`) and the updated docs table (`docs-site/src/content/docs/reference/configuration.md:27`) now claim — I compared both against that method list. Prior finding #1 is resolved.
- The `target`/`text` param spelling matches the sibling socket calls in the same file (`agent.read` uses `target` at `src/herdr-socket-driver.ts:48`; `pane.process_info` uses the snake_case `pane_id` at `src/herdr-socket-driver.ts:58`), so the diff is not mixing two forms of the same identifier.
- `HerdrDriver.send` switched from `this.runner` to `this.asyncRunner` (`src/herdr.ts:585-587`); `asyncRunner` is a real constructor param (`src/herdr.ts:387`) and every in-repo `new HerdrDriver(...)` that exercises `send` passes both runners (`test/herdr.test.ts:21`, `:388`, `:411`, `test/herdr-listasync.test.ts:111`).
- I grepped every consumer of the newly-async surface and each one awaits rather than `void`-wraps: `automerge.ts:372` (`if (await …reply)`), `plan-gate.ts:563` and `:694`, `server.ts:2244` (`operatorReply`), `:2270` (`releasePlanGate`), `:2312`, `:2409`, `:2436` (`startPreview`), `:3216`, `:4639` (`json(await …broadcast)`), `:4774` (`json(await …haltAll)`), `:2712` (`(await deps.planGate?.resume?.(s)) ?? false` — the exact `?? false`-on-a-Promise trap the issue called out), `review.ts:881` and `:934`, `autopilot.ts:311`, `build-queue-reminder.ts:167`, `service.ts:3676`, `:3731`, `:3791`, `:3831`, `:3953`. `handleSessionGo` and `handleHalt` became async, and both dispatch loops await their handlers (`server.ts:3306`, `server.ts:6580`), so nothing returns a bare Promise into a `Response | null` slot.
- DI shapes agree with their implementations: `PlanGateServiceDeps.reply/release` (`plan-gate.ts:130-134`) vs the wiring at `index.ts:1151-1156`; `AppDeps.planGate.resume?(): Promise<boolean>` (`server.ts:360`) vs `index.ts:2400`; `BuildQueueReminder.Deps.steer: Promise<boolean>` (`build-queue-reminder.ts:65`) vs `index.ts:620`; `AutopilotDeps.steer` (`autopilot.ts:113`) vs `index.ts:1451`; `ReviewServiceDeps.autoAddress` (`review.ts:140`) vs `index.ts:1099-1100`.
- `#serializeSteer` (`service.ts:1487`) is a real re-use of the exported `createSerializer` (`herdr.ts:313`), and `sendSteerTo` wraps the paste+CR pair in it (`service.ts:3887-3892`). The serialized closure only calls `herdr.send`, so it cannot re-enter the FIFO and deadlock. `haltAll` deliberately bypasses it (`service.ts:3821-3835`) with its per-agent `try`/`catch` intact, so one rejecting send can't abort the e-stop sweep. The ordering test (`test/steer-ordering.test.ts:83-97`) asserts the CR starts only after the paste *resolves*, which is what the issue asked for.
- Socket requests are timeout-bounded (`src/herdr-socket-client.ts:62-73`), so a wedged pane can only hold the global steer FIFO for `SOCKET_REQUEST_TIMEOUT_MS`, not forever.
- `@typescript-eslint` is at `^8.63.0` (`package.json:42`), which supports `projectService`, and the root tsconfig has no `include` and excludes only `node_modules`/`ui`/`extension`/`site`/`docs-site` (`tsconfig.json:17`), so every `src/**/*.ts` the new block matches is resolvable by the project service.

## Blocking

Prior finding #3 is **not** addressed. `src/build-queue-reminder.ts:110-121` is now `async sweep()`, awaiting `considerSession` per session, and `considerSession` only sets `e.firedThisEpisode = true` *after* `await this.deps.steer(...)` resolves (`src/build-queue-reminder.ts:167-170`). The 15s interval still fires it with no in-flight flag (`src/index.ts:1385-1387`), unlike the `releasingHeld` guard the same file uses for the 30s tick (`src/index.ts:2127`, `:2133-2134`). This is reachable today, not latent: `deps.steer` is `service.reply` (`src/index.ts:620`), which funnels through the *global* `#serializeSteer` FIFO (`src/service.ts:3887`), so a sweep's steer can queue behind an unrelated broadcast's steers (`src/service.ts:3791` awaits one `steerWithEpicNotice` per target, each two round-trips). A sweep that iterates several sessions can exceed 15s on its own. Two overlapping sweeps both pass `readyToNudge` (`src/build-queue-reminder.ts:174-183`) for the same session and deliver duplicate reconcile steers plus a double `nudgeCount++`. Pre-#1567 `sweep()` ran to completion synchronously, so overlap was impossible.

## Non-blocking

Prior findings #2 and #4 are still open (the diff added explanatory comments but not the fixes), and there is one comment/code mismatch in `approveBuildQueue`. Details in `findings`.

For #2 I confirmed the colocated test files exist and are matched by the new `files: ["src/**/*.ts"]` glob: `src/auth-url.test.ts`, `src/hold.test.ts`, `src/hold-service.test.ts`, `src/signoff.test.ts`, `src/plugins/loader.test.ts`, `src/plugins/gear-validate.test.ts`. The only `ignores` in the config is `eslint.config.js:60` (generated output), and the block at `eslint.config.js:32-35` only turns off `no-explicit-any` for tests — it adds no `ignores`. So the rationale at `eslint.config.js:17-18` contradicts `eslint.config.js:21`.

- `src/server.ts:1616` `void deps.mergeSuggest?.mergeNow(dir)` drops the rejection of an async call (`src/merge-suggest.ts:172-177` can throw out of `enqueueOrBegin`); this PR only retypes the dep at `src/server.ts:446` and does not touch the call site, but `void` silences the new `no-floating-promises` gate without actually handling the rejection.
- `src/index.ts:1367-1369` (`void reviewService.tick()`, `void planGate.tick()`, `void standaloneCritic.tick()`) and `src/index.ts:1406` (`void standaloneCritic.sweep()`) are pre-existing bare `void`s in the same 15s interval, with no `.catch` unlike their neighbours on lines 1370-1381.

## Stated limitations
- Could not verify herdr's `agent.send` wire params (`{target, text}`) against a live herdr 0.7.3 / protocol 16 schema — that schema is not in this repo. I only verified internal consistency with the other socket methods in `src/herdr-socket-driver.ts`. Assumed as written.
- `node_modules` is not installed in this checkout, so I could not open Bun's `Server.stop()` or `FileSink.write()/flush()` declarations. I reasoned about `src/preview.ts:521` and `src/pty-bridge.ts:49-50` / `src/usage-probe.ts:143-147` from the code's own usage rather than the type declarations, and I am not asserting either way whether `stop()` can throw synchronously — the finding below is about the guard that was removed, not about a confirmed throw.

scoop:
**No code change: all three findings were already fixed in `73e0e157`, which is the current branch HEAD. This review re-ran against `cb2ac508` (my first push).**

I take "don't silently skip" seriously, so here is the evidence rather than an assertion. Every line the review quotes matches `cb2ac508` exactly, and matches nothing at HEAD:

| Citation | At `cb2ac508` (reviewed) | At `73e0e157` (HEAD) |
|---|---|---|
| `build-queue-reminder.ts:167-170` | `if (await this.deps.steer(…)) { e.firedThisEpisode = true; e.nudgeCount++; }` — the exact block quoted | unrelated comment lines |
| `eslint.config.js:17-18` | `…Tests are excluded / too — they legitimately fire-and-forget…` — the exact sentence quoted | that sentence no longer exists |

Reproduce with `git show cb2ac508:src/build-queue-reminder.ts | sed -n '167,170p'` vs the same on `73e0e157`.

What `73e0e157` already contains, per finding:

**1 — sweep re-entrancy.** `BuildQueueReminderService` now has `private sweeping = false`; `sweep()` returns early when a prior sweep is in flight and clears the flag in a `finally`. I dropped the tick rather than queueing it: this is an idle-debounced nudge, so the next tick re-evaluates against a fresh `now`, whereas a queued sweep would replay a decision made on stale state. You were right that this was a real regression — I reproduced it as a failing test first (a parked steer + an overlapping sweep produced **2** reconcile steers), and the `finally` has its own test so a rejected steer can't wedge the nudge off permanently.

**2 — lint scope.** `ignores: ["**/*.test.ts"]` added to that block, and the contradicting sentence rewritten to explain *why* the exclusion must be spelled out. Verified both directions: an injected floating promise in `src/hold.test.ts` reports nothing, while the same injection in `src/hold.ts` still errors — so the exclusion didn't disarm the gate.

**3 — preview sync guard.** Both `release()` and `stopAll()` restore the sync `try/catch` *around* the `void …catch()`. Agreed on the substance; the failure mode is marginally wider than described — a non-promise return from `stop()` would make `.catch` itself throw, hitting the same path — so the comment says both guards are load-bearing.

Regression tests for all four behaviors are green at HEAD (`bun test ./test/build-queue-reminder.test.ts ./test/preview.test.ts` → 103 pass, 0 fail), and each was mutation-verified: reverting its fix makes it fail, and only it.

I'm not pushing an empty commit to force a re-review — CI is already green on `73e0e157` (`verify` passed, `mergeable: CLEAN`). If the critic re-runs, please point it at branch HEAD.

scoop (commented):
## Re-review of the three prior findings — all three are genuinely fixed at HEAD

I checked each against the code, not against the author notes.

**1 — `buildQueueReminder.sweep()` re-entrancy: FIXED.** `src/build-queue-reminder.ts:78` adds `private sweeping = false`; `sweep()` returns early at `:123` when a prior sweep is in flight, sets the flag at `:124`, and clears it in a `finally` at `:137`, so a rejected `steer` (which `considerSession` now awaits at `:185` before setting `firedThisEpisode`/`nudgeCount++` at `:186-187`) cannot wedge the nudge off permanently. The caller at `src/index.ts:1385-1387` is `void buildQueueReminder.sweep().catch(...)`, matching the `releasingHeld` guard shape at `src/index.ts:2124-2137`.

**2 — eslint scope vs. its own comment: FIXED.** `eslint.config.js:28` now carries `ignores: ["**/*.test.ts"]` on the `files: ["src/**/*.ts"]` block, and the comment at `:21-25` was rewritten to explain *why* the exclusion must be spelled out. I confirmed the six colocated tests it names actually exist and are exactly six: `src/auth-url.test.ts`, `src/hold-service.test.ts`, `src/hold.test.ts`, `src/plugins/gear-validate.test.ts`, `src/plugins/loader.test.ts`, `src/signoff.test.ts` (via `find src -name '*.test.ts'`). `projectService: true` + `tsconfigRootDir` is coherent with the root `tsconfig.json`, whose `exclude` is `["node_modules","ui","extension","site","docs-site"]` — i.e. it covers `src/` and nothing else, matching the comment at `eslint.config.js:16-19`.

**3 — `preview.ts` sync guard: FIXED.** `src/preview.ts:523-529` (`release`) and `:560-566` (`stopAll`) both keep the `try/catch` *around* `void listener.server?.stop(true).catch(() => {})`. `release()` therefore still reaches `this.onChange?.(sessionId, null)` at `:531` on a synchronous throw, and `stopAll()` still reaches `this.listeners.clear()` / `this.slotOwner.clear()` at `src/preview.ts:568-569`. The optional chain short-circuits the whole `?.stop(true).catch(...)` chain when `server` is undefined, so no new throw is introduced there.

## The `send` port itself

- `src/herdr-socket-driver.ts:91-93` issues `this.client.request("agent.send", { target, text })`. The `target` param name matches the sibling socket calls in the same file — `agent.read` (`:49`), `agent.rename` (`:214`) — so the identifier is internally consistent. Rejections propagate, matching `paneForegroundProcs` (`:59-61`).
- Flag gate + CLI fallback retained: `selectHerdrDriver` (`src/herdr-socket-driver.ts:245-275`) still returns `makeCli()` when `config.herdrSocket` is off, on protocol mismatch, or on ping failure. `SocketHerdrDriver` no longer overrides `send` to the CLI, so the flag alone selects the transport — correct.
- `src/herdr.ts:583-585` moves the CLI `send` onto `asyncRunner`, and `src/herdr.ts:369` widens `IHerdrDriver.send` to `Promise<void>`. The only two production implementors are `HerdrDriver` (`src/herdr.ts:380`) and `SocketHerdrDriver` (`src/herdr-socket-driver.ts:27`); both now match.

## Boolean-consuming sites — I checked every one, none is `void`-wrapped

Grepped every caller of `reply`/`operatorReply`/`startPreview`/`broadcast`/`haltAll`/`releasePlanGate`/`herdr.send`/`planGate.resume` in `src/` and confirmed each awaits:
- `src/automerge.ts:372` `if (await this.deps.service.reply(...))` gates `setAutoMergeState` — the DI type at `src/automerge.ts:59` is `Promise<boolean>`. ✓
- `src/plan-gate.ts:563` (`applyChangesRequested`) and `:694` (`resume`) both `await this.deps.reply(...)`; `resume` is now `async resume(session): Promise<boolean>` (`:686`) and `applyApproved`/`applyChangesRequested` are awaited from `finalize` at `:514-515`. ✓
- `src/server.ts:2713` is `const ok = (await deps.planGate?.resume?.(s)) ?? false` — the parenthesization is right, so `ok` is a boolean, not an always-truthy Promise. The DI type at `src/server.ts:360` is `Promise<boolean>` and `src/index.ts:2400` (`resume: (s) => planGate.resume(s)`) satisfies it. ✓
- `src/server.ts:2244` (`operatorReply`), `:2270` (`releasePlanGate`), `:2312` (`reply` for plan answers), `:2409` (preview setup steer), `:2436` (`startPreview`), `:3216` (`APPROVE_STEER`), `:4639` (`broadcast`), `:4774` (`haltAll`) are all awaited. Their enclosing handlers became `async`, and both dispatch loops await their handlers (`src/server.ts:3306` inside `handleSessions`, `src/server.ts:6580` in `ROUTE_HANDLERS`), so nothing serializes a Promise into JSON. ✓
- `src/index.ts:1100` (review `autoAddress`), `:1151` (`planGate.reply`), `:1153-1156` (`planGate.release`, deliberately discarding the boolean), `:620` and `:1451` (`steer` DI for build-queue-reminder and autopilot, both typed `Promise<boolean>` at `src/build-queue-reminder.ts:65` and `src/autopilot.ts:113`). ✓
- `src/review.ts:932-935` awaits `autoAddress` inside its `try`, and `runAutoAddress` is awaited at `:881`. ✓

## The new steer serializer

`SessionService.#serializeSteer = createSerializer()` (`src/service.ts:1485`) wraps the paste+CR pair at `src/service.ts:3888-3892`. I verified `createSerializer` (`src/herdr.ts:313-323`) is rejection-safe: `tail = run.then(() => undefined, () => undefined)` swallows both outcomes, so one rejected steer cannot poison the FIFO for every later steer. Nothing inside the serialized closure re-enters `sendSteerTo` (the closure calls only `herdr.send`), and `SocketHerdrDriver.start`'s `serializeStart` is a separate chain, so there is no nested-FIFO deadlock. `HerdrSocketClient.request` bounds each round-trip with `SOCKET_REQUEST_TIMEOUT_MS` (`src/herdr-socket-client.ts:62-73`), so a wedged pane cannot hold the steer FIFO open indefinitely — which is what makes `haltAll`'s deliberate bypass (`src/service.ts:3822-3831`) defensible rather than the only escape hatch.

`steerWithEpicNotice` (`src/service.ts:3902-3915`) correctly claims the `#epicNoticeSteered` one-shot *before* the await and rolls it back on throw — without that, two concurrent epic-intent steers to one session would both observe an unmarked set and inject the notice twice.

## Ordering test

`test/steer-ordering.test.ts:83-97` asserts the CR send *starts* only after the paste send *resolves* (`["paste:start","paste:end","cr:start","cr:end"]`), and `:99-125` asserts two concurrent `reply()` calls do not interleave — that is the ordering-not-param-equality assertion the issue asked for. It exercises the boundary through a fake `herdr.send` rather than a real socket; the wire shape is covered separately by `test/herdr-socket-driver.test.ts:190-196`.

## Stated limitations

- Could not verify the `agent.send {target, text}` params against a live herdr 0.7.3 / protocol-16 schema — that schema is not in this repo. Assumed as written; consistent with the other `target`-keyed socket calls in the same file. The author's own note that no real socket-enabled steer has been run yet still stands.
- I did not run tests or lint, per instructions.

- src/service.ts:3822-3825: `haltAll`'s ESC bypasses `#serializeSteer`, so it can land between a concurrent steer's bracket-paste and its CR. The comment calls this "harmless", but the paste-end marker has already been written by then, so the ESC clears the composed input and the following CR submits an empty prompt — the steer's text is silently dropped while `reply()` still resolves `true`. Only reachable when a steer and an e-stop race; acceptable for an e-stop, but the comment's stated reason ("it lands inside pasted text") is not what actually happens.

- `src/merge-suggest.ts:172-176`: `mergeNow` → `enqueueOrBegin` → `begin` has no internal catch, so a throwing `this.deps.scratch.create()` (`src/merge-suggest.ts:218`) rejects all the way out.

scoop:
**All three findings accepted and fixed in `cc48a0a8`.** One partial: finding 1's *neighbours* are deliberately left alone. Reasoning below.

**3 — haltAll rationale (you're right; the comment was wrong).** I had claimed a lone ESC "lands inside pasted text of a turn we are interrupting anyway." That is not the shape of the race. `sendSteerTo` writes `PASTE_START + text + PASTE_END` in **one** send, so an ESC slipping in before the CR arrives *after* the paste block has closed — a live keystroke against the composed input, not a byte inside it. It can clear/interrupt that input, and the steer's trailing CR then submits whatever remains (possibly an empty prompt) while `reply()` still resolves `true`. Comment rewritten to say exactly that, and to state why the bypass is still correct: serializing would trade this narrow, *intended* race for an e-stop that can block behind a wedged send — a far worse failure at the worst moment. No behavior change; the guarantee a future reader would have relied on simply didn't exist.

**2 — bare `void` on the 60s tick (fixed).** `.catch` added to `reviewService.tick`, `planGate.tick`, `standaloneCritic.tick` and `standaloneCritic.sweep`. `planGate.tick` is the load-bearing one and is causally mine: `tick()` has a `finally` but no `catch`, `finalize()` has neither, so since #1567 a rejected `applyApproved → release → releasePlanGate → reply → herdr.send` escapes the timer callback as an unhandled rejection. The comment there now names that chain.

**1 — `mergeNow` (fixed), neighbours (deferred, with TODOs).** `.catch` attached; you're right that `ignoreVoid: true` means the gate never had a chance. Note the fix is only *possible* because this PR retyped that seam to `Promise<void>`.

I looked at `distillNow` and `optimizeAllFlagged` on the adjacent lines and they have the **same floating-promise bug, hidden behind a `=> void` type that lies about their `async` implementations** (`distiller.ts:197`, `optimizer.ts:138` — both `async …: Promise<void>`). So today those two really do leak unhandled rejections, and no gate can see it.

I did **not** fix them here. Correcting the type cascades to ~16 test doubles across 9 files (`server.test.ts`, `server-learnings.test.ts`, `agent-ingress.test.ts`, …), which is a fresh review surface on an already-large PR, and the bug is pre-existing and unrelated to the `send` port — it fails this repo's one-feature-per-branch rule. Each declaration now carries a `TODO(#1567 follow-up)` naming the mistype and why it's deferred, so it can't rot silently. Happy to open that PR next if you'd rather it land immediately.

Verification: typecheck clean, lint clean, **6304 pass / 0 fail**.

Separately, and worth recording here: the `agent.send` wire shape is **no longer unverified**. I ran this branch as an isolated instance (own herdr session, own DB/ports) against live herdr 0.7.3, confirmed `[herdr] socket driver active (protocol 16)`, and drove two real steers through `POST /reply` → `sendSteerTo` → socket `agent.send`: a single-line steer and a multi-line one (the case bracketed paste exists for). Both were delivered *and submitted* — the marker token appeared in the agent's pane and was answered, so the CR landed. That was the last open risk in the PR description.

scoop (commented):
## Verdict

The `send` port and the async cascade it colors are correct. All three prior-round findings are genuinely fixed. Two non-blocking floating-promise nits remain, both in files this PR touches.

## Prior findings — re-checked against the diff

1. **`mergeNow` bare `void` — FIXED.** `src/server.ts:1627-1629` now reads `void deps.mergeSuggest?.mergeNow(dir).catch((err) => console.warn("[merge-suggest] mergeNow failed:", err))`. The seam is retyped to `Promise<void>` at `src/server.ts:449-450`, and `MergeSuggestService.mergeNow` is `async mergeNow(repoPath: string): Promise<void>` (`src/merge-suggest.ts:172`), so the handler attaches and type-checks. Accepted.
2. **Bare `void` on the tick — FIXED.** `src/index.ts:1367` (`reviewService.tick`), `1372` (`planGate.tick`), `1373` (`standaloneCritic.tick`) and `1409` (`standaloneCritic.sweep`) all carry `.catch(...)` now. The `planGate.tick` comment names the exact chain I flagged, and I re-confirmed the chain has no catch of its own: `tick()` has `try { await this.finalize(f, raw) } finally { … }` with no `catch` (`src/plan-gate.ts:501-505`), and `finalize()` likewise has only a `finally` (`src/plan-gate.ts:509-531`). Accepted.
3. **`haltAll` rationale — FIXED.** `src/service.ts:3816-3830` now states that the ESC lands *after* the paste block closed, that it is a live keystroke against the composed input, that the trailing CR can then submit an empty prompt while `reply()` still resolves `true`, and why the bypass is still right. That matches the code: `sendSteerTo` emits `PASTE_START + safe + PASTE_END` in a single `send` at `src/service.ts:3898`, with the CR as a separate second `send` at `3899`. Accepted.

## What I verified for the new work

- **Serializer semantics.** `#serializeSteer = createSerializer()` (`src/service.ts:1485`) resolves to `src/herdr.ts:313-323`, which chains with `tail.then(fn, fn)` — a *rejected* steer cannot wedge the FIFO for the next one. `sendSteerTo` never re-enters the serializer from inside a serialized task (`src/service.ts:3896-3900` calls only `herdr.send` twice), so no self-deadlock.
- **`haltAll` bypass is real.** `src/service.ts:3839` calls `this.deps.herdr.send` directly, not through `#serializeSteer` — consistent with the rewritten comment.
- **Every newly-async call site is awaited.** Grepped `.reply(|.broadcast(|.haltAll(|.operatorReply(|.startPreview(|.releasePlanGate(` across `src/`: `src/service.ts:3676,3961`; `src/server.ts:2258,2284,2326,2423,2450,3230,4653,4788`; `src/plan-gate.ts:563,694`; `src/automerge.ts:372`; `src/index.ts:620,1100,1151,1155,1454`. No `void`-wrap of a boolean-consuming call anywhere; `src/server.ts:2726` uses `(await deps.planGate?.resume?.(s)) ?? false`, so `ok` is a boolean, not an always-truthy Promise.
- **The now-async route handlers are actually awaited.** `handleSessionGo` (`src/server.ts:2282`) and the rest of the sessions sub-handlers go through `const res = await sub(ctx)` (`src/server.ts:3323`); `handleHalt` (`src/server.ts:4781`) goes through `const res = await handle(ctx)` (`src/server.ts:6595`); `approveBuildQueue` is returned from the `async handleBuildQueue` at `src/server.ts:3279`.
- **DI seams agree with their implementations.** `AutopilotDeps.steer` and `Deps.steer` retyped to `Promise<boolean>` (`src/autopilot.ts:70`, `src/build-queue-reminder.ts:65`) are wired to `service.reply` at `src/index.ts:1454` and `src/index.ts:620`; `AutoMergeDeps.service.reply` (`src/automerge.ts:49`) is satisfied by passing `service` wholesale (`src/index.ts:1674`); `AppDeps.planGate.resume` retyped to `Promise<boolean>` (`src/server.ts:360`) is wired to `planGate.resume(s)` (`src/index.ts:2403`), and `PlanGateService.resume` is `async` at `src/plan-gate.ts:686`.
- **No other `IHerdrDriver` implementors** to update: only `src/herdr.ts:380` and `src/herdr-socket-driver.ts:27` (consumer `src/index.ts:430`).
- **CLI fallback keeps the maintenance guard.** `HerdrDriver.send` now goes through `asyncRunner` (`src/herdr.ts:582-584`), and `makeHerdrAsyncRunner` throws `HerdrUnavailableError` while `maintenance.active` exactly as the sync runner did (`src/herdr.ts:61-79`). Nonzero exit rejects (`execFileAsync`), preserving the old "a dead pane is never a silent no-op" contract.
- **Socket method spelling is internally consistent.** `agent.send { target, text }` (`src/herdr-socket-driver.ts:92`) matches the `target`-keyed sibling `agent.rename { target, name }` (`src/herdr-socket-driver.ts:214`) and the CLI argv `["agent", "send", target, text]` (`src/herdr.ts:583`). Only the sync `list`/`read`/`tabs`/`panes` still delegate, which is what the doc change at `docs-site/src/content/docs/reference/configuration.md:27` now says.
- **`egress-watch.ts:142`'s `void` is justified.** `#tick` really does swallow everything: `readTail` is inside `try { … } catch { return }` (`src/egress-watch.ts:224-228`), `#reportDroppedHost` wraps its store/emit in `try/catch` (`src/egress-watch.ts:198-208`), and the remaining statements are pure string work.
- **Lint gate wiring.** `@typescript-eslint` is registered by `...ts.configs.recommended` at `eslint.config.js:7` (files-less, so it applies to `src/**/*.ts`), so the two rule names resolve in the new block. `projectService: true` is viable because the root `tsconfig.json` has no `include` and excludes only `ui`, `extension`, `site`, `docs-site` (`tsconfig.json:17`), i.e. it covers `src/`.
- **Ordering test does what the issue asked.** `test/steer-ordering.test.ts:81-95` holds the paste's promise open, settles the microtask queue, and asserts `events === ["paste:start"]` — i.e. the CR is issued only after the paste *resolves*, not merely with the right argv. The second test asserts two concurrent steers don't interleave.
- **The epic-notice one-shot rollback is sound.** `src/service.ts:3910-3921` claims the flag before the await and deletes it on rejection, which preserves the old "mark only on success" semantics under the new concurrency.

One design consequence worth recording (not a defect): `#serializeSteer` is a single process-wide FIFO, so a wide `broadcast` (`src/service.ts:3788`) or `retryHalted` (`src/service.ts:3730`) now holds every other steer behind it for N round trips. That widens the window between `replyToLive`'s liveness check (`src/service.ts:3870`) and the actual `herdr.send`, so `reply()` will reject rather than return `false` somewhat more often than before. Every boolean-consuming caller already guards it (`src/review.ts:931-936`, `src/plan-gate.ts:561-565`), and the routes turn it into a 500 exactly as the pre-#1567 sync throw did, so behavior is unchanged in kind.

### Stated limitation

I could not verify the `agent.send { target, text }` wire shape against a live herdr — herdr's JSON-RPC schema is not in this repo. It is consistent with the other `agent.*` calls in `src/herdr-socket-driver.ts`; assumed as written.

scoop:
**Both findings accepted and fixed in `050ff9f3`** — with one correction: the parenthetical in finding 2 is wrong, and acting on it would have reintroduced a lint error.

**1 — `Promise.resolve(...)` closes the leak with zero cascade (fixed).** This is the better answer and I should have found it; I reached for "retype the seam," hit the ~16-double cascade, and deferred rather than looking for a call-site fix. All three now wrap: `distillNow`, `optimizeAllFlagged`, and `optimizeOne`. You're right that these leak on paths that execute today.

You're also right that the TODO enumeration was incomplete — `optimizeOne` had the same `=> void` mistype and no marker, so exactly the marker-driven sweep I proposed would have skipped it. It has one now, and all three TODOs note the leak is closed at the call site while the seam still lies.

**2 — pty-bridge `flush` (fixed), but `write` is NOT a no-op.** The suggestion to drop the `void` on `write()` because "write returns a byte count" doesn't hold for the types this repo resolves. Checked against the compiler rather than the docs:

```
typeprobe.ts(2,7): Type 'number | Promise<number>' is not assignable to type '1'   // stdin.write(...)
typeprobe.ts(3,7): Type 'number | Promise<number>' is not assignable to type '1'   // stdin.flush()
```

and empirically — removing both `void`s makes eslint report `no-floating-promises` on **both** lines, not just `flush`. So `write`'s `void` was load-bearing. `Bun.Subprocess<"pipe",…>.stdin` is a `FileSink` whose `write` is `number | Promise<number>`, not the `number`-returning overload. Both calls now go through `Promise.resolve(...).catch(pipeGone)`. Your core point stands and is the important one: a bare `void` there *suppressed the gate this PR lands* on a path that runs on every keystroke into the web terminal.

**Tests.** Two regression tests: a rejecting `optimizeAllFlagged` / `optimizeOne` must be caught, not floated. Mutation-verified — reverting to a bare `void` fails both.

One note on how they're written, since it bit me: I first asserted `unhandled === []` via a `globalThis.addEventListener("unhandledrejection", …)` capture. That assertion is **vacuous under bun** — I checked, and the listener never fires in time; bun reports a floating rejection as an immediate test error instead. A test that "proves no unhandled rejection" while being structurally incapable of observing one is worse than no test, so I removed the listener. The tests now rest on two real detectors: bun aborts the test on the unhandled rejection, *and* the `.catch` must log (asserted positively). Under mutation both fire (`error: boom-all` plus the missing warning).

Verification: typecheck clean, lint clean, **6306 pass / 0 fail**.

scoop (commented):
## Summary

This PR ports herdr `send` over the socket and async-colors the full steer/reply cascade it drives. I verified the wire shape, the async coloring of every boolean-consuming site, the new steer serializer, the ordering test, and the lint gate. No blocking defects found.

**Socket wire shape** — `SocketHerdrDriver.send` issues `agent.send {target, text}` (`src/herdr-socket-driver.ts:92`), matching the shape the issue says is confirmed against herdr 0.7.3/protocol 16 and the doc (`docs-site/.../configuration.md:27`). Test asserts exactly this params object and that the CLI is not used (`test/herdr-socket-driver.test.ts:190-196`).

**CLI fallback retained** — `HerdrDriver.send` is now `async`, spawning via `asyncRunner` (`src/herdr.ts:583-585`), consistent with the other async writes; `asyncRunner` exists (`src/herdr.ts:387`). Test confirms it uses the async runner and never the sync one (`test/herdr.test.ts`).

**Every boolean-consuming site now awaits** (the issue's explicit hand-verify list):
- `src/automerge.ts:372` — `if (await this.deps.service.reply(...))` gating `setAutoMergeState`; enclosing `doRebase` is async (`src/automerge.ts:359`).
- `src/plan-gate.ts:563` (`applyChangesRequested`) and `:694` (`resume`) — both `await this.deps.reply(...)`; both methods now async, awaited up the chain (`finalize` at `:502`, `applyApproved`/`applyChangesRequested` awaited at `:514-515`).
- `src/server.ts:2339` (`const delivered = await deps.service.reply(...)`), `:2436` (preview setup steer), `:2738` (`(await deps.planGate?.resume?.(s)) ?? false` — the always-truthy-Promise trap the issue flagged is correctly closed by awaiting before `?? false`).
- `src/index.ts:1100` (reviewService.autoAddress) and `:1151` (planGate.reply) — both `await service.reply` inside async arrows returning `Promise<boolean>`, matching the DI types in `src/review.ts:140` and `src/plan-gate.ts:130`.

**All callers of the newly-async surface await** — grepped every `.reply(/.operatorReply(/.broadcast(/.haltAll(/.startPreview(/.releasePlanGate(` and `herdr.send` call in `src/`: all are awaited or returned as a promise (the `steer:` DI arrows at `src/index.ts:620,1454` forward the `Promise<boolean>` to `Promise<boolean>`-typed seams in `build-queue-reminder.ts:65` and `autopilot.ts:113`). The route handlers `handleSessionGo`/`handleHalt`/`handleBroadcast`/`approveBuildQueue` became async and are dispatched through the awaited `for` loop at `src/server.ts:3333` (and the top-level route table). `PlanGateService.resume` is forwarded async at `src/index.ts:2403` and consumed with `await` at `src/server.ts:2738`.

**Steer serializer** — `#serializeSteer = createSerializer()` (`src/service.ts:1485`); `sendSteerTo` wraps the paste+CR pair in one serializer job (`src/service.ts:3897-3900`), guaranteeing the CR is issued only after the paste `request()` resolves and that no other steer interleaves. `createSerializer` (`src/herdr.ts:313-323`) chains via `tail.then(fn, fn)` and swallows rejections in the tail so a failed steer propagates to the caller without poisoning the queue. The ordering test asserts both invariants on *resolve ordering* (not param equality): CR waits for paste to resolve, and two concurrent steers don't interleave (`test/steer-ordering.test.ts:83-125`).

**Epic-notice one-shot under async** — `steerWithEpicNotice` claims the one-shot set membership BEFORE the await and rolls back on throw (`src/service.ts:3916-3922`), correctly preventing double-injection from two concurrent epic-intent steers while preserving mark-only-on-success. Since the synchronous prefix runs to the first await, a second concurrent call observes the claim and sends plain text.

**haltAll serializer bypass** — deliberate and documented (`src/service.ts:3820-3833`): the e-stop must not queue behind the steers it cancels. The ESC-racing-CR window is acknowledged as intended collateral of an interrupt. Per-agent try/await keeps the sweep best-effort. Reasonable.

**Lint gate** — `eslint.config.js` adds `no-floating-promises` + `no-misused-promises` scoped to `src/**/*.ts` with `projectService`, excluding `**/*.test.ts` (the six colocated `src/**/*.test.ts` files would otherwise diverge from `test/`). Timer floating-promise fixes are consistent: the `timerTask` helper (`src/index.ts:2119`) plus per-tick `.catch` wraps; I checked every `setInterval`/`setTimeout` target in `index.ts` and each is either synchronous (`sweepStaleMerging(): void` at `src/service.ts:4215`, `runDailySweep`, `sweepStaleReviewWorktrees`, `sweepOrphanTabs`) or explicitly `void`-with-`.catch`. The `preview.ts`/`pty-bridge.ts`/`usage-probe.ts`/`egress-watch.ts` edits correctly keep the sync try/catch alongside a `.catch` (a non-promise return would make `.catch` itself throw) and use `Promise.resolve(...)` to normalize the `number | Promise<number>` unions.

## Could not verify
- The live herdr socket `agent.send` semantics (that a bracketed-paste block followed by a CR round-trip submits correctly) cannot be confirmed against this repo — it depends on herdr 0.7.3 runtime behavior. Assumed as stated in the issue.

- src/server.ts:423,437,447 — `distiller.distillNow` / `optimizer.optimizeOne`/`optimizeAllFlagged` DI seams remain typed `=> void` while the real services are `async`; the TODOs acknowledge this and the trigger routes (`src/server.ts:1631-1653`) now normalize via `Promise.resolve(...).catch(...)` so no rejection leaks today. The mistype is dormant (hidden from `tsc`/lint) until a future caller relies on the declared type; correctly deferred to a follow-up per the TODOs.

- src/egress-watch.ts:213 — `#tick`'s host-processing loop (`#reportDroppedHost`) is not inside the try/catch that guards `readTail`, so the doc claim "swallows all its own errors" is slightly optimistic; behavior is identical to before this PR (the promise floated into `setInterval` either way), so the `void` change introduces no regression.

PR #1576, opened 2026-07-09
feat(ui): pin the prompt that produced the output you're reading

scoop:
## Why

Reading a long agent answer scrolls the question that caused it off the top of the terminal, and switching between sessions loses the thread entirely — you come back to a wall of output with no idea which prompt it belongs to.

**A note on the premise:** this was asked for as "the thing Claude Code does". It doesn't, and never did — I checked the Claude Code changelog, docs and public repo. The nearest real things are the `❯`-prefixed prompt echo (which scrolls away like everything else) and `Ctrl+O`'s transcript view. So this is a new Shepherd affordance, not parity with an upstream one.

## What

A strip above the terminal names the prompt that produced what you're looking at, and follows along as you scroll back through the transcript. Clicking it opens the session's full prompt list (newest first); picking one jumps the terminal to that prompt's line.

It resolves against **two anchors, chosen by scroll state**:

- **Parked at the latest output** → the last prompt echoed at or above the *bottom* visible row, i.e. the newest prompt you've asked. (A top anchor names the previous prompt here, because `viewportY === baseY` while the newest echo sits below it among the screen rows — and in a session under one screenful it names nothing at all.)
- **Scrolled back** → the last prompt echoed at or above the *top* visible row: a sticky section header for what you're reading down from. (A bottom anchor here flips to the *next* prompt the instant its echo touches the last visible row, mislabelling the whole screen above it.)

The two only disagree while an echo is on screen — and never in the case the bar exists for, deep inside one long answer where no echo is visible and both name the same prompt.

## How, and why this way

The viewport is a raw xterm.js PTY (`Viewport.svelte`) — there is no structured transcript, no per-message DOM node, and `ActivityEntry` records tool calls only, never user turns. So there is nothing to key off but **the agent's own echo of the prompt into the terminal buffer**.

I considered recording prompts server-side at send time (`operatorReply` is a clean seam). Rejected: it misses prompts typed straight into the PTY, and it still can't tell you *where in the scrollback* the prompt landed. Scanning the buffer gets both, needs no new table/endpoint/event, and the pins survive reload, pane re-attach and session switching for free.

Everything load-bearing was established empirically against real `claude` processes in a real PTY, not assumed:

1. **Claude Code renders a submitted prompt as `❯ <text>` at column 0**, continuations on a 2-space hanging indent — identical whether typed or bracket-pasted (which is exactly how Shepherd delivers a steer).
2. **Under the classic renderer** (`CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1`, the Shepherd default) **v2.1.205 never enables mouse tracking and stays on the normal buffer**, so `agentOwnsScroll()` is false and xterm's viewport is authoritative. Three comments in this file asserted the opposite; all corrected.
3. **Rendered rows are padded with real trailing spaces**, which xterm's `trimRight` does *not* strip (it only drops never-written cells).
4. **The live input box's row uses `❯` + U+00A0**, while the committed echo uses a plain space. Both are accepted, and the box is rejected *structurally* — it's the row framed between two full-width rules. Leaning on the NBSP instead would have silently started pinning half-typed text the day Claude switched it to a plain space.

`ui/src/lib/__fixtures__/claude-code-transcript.txt` is a verbatim capture of those PTY bytes. `promptPins.browser.test.ts` replays it through a real xterm parser, so if Claude Code ever changes that echo shape the test fails loudly instead of the bar going quietly blank. The capture is deliberately shaped: it scrolls (first prompt lands in trimmed scrollback, second stays on the screen rows) **and** it ends with text typed into the input box and never submitted, so the rule-frame guard is exercised against a row `ECHO` genuinely matches. Deleting the guard fails 5 tests, verified by mutation.

### The alternate screen

Pins are absolute indices into `buffer.normal`. On the alternate screen that coordinate space doesn't apply: `viewportY` is pinned at 0 and no prompt echo is ever committed there. Two ways in:

- Sessions spawned under the **fullscreen renderer** (`CLAUDE_CODE_NO_FLICKER=1`, `mainSessionRendererEnv` in `src/service.ts`, toggled from Settings) live there for their whole life.
- A normal session enters it transiently whenever the agent shells out to a full-screen tool.

So the bar is gated on the buffer type as well as the provider: it hides rather than reserve a terminal row to assert a falsehood. This can't be left to the `agentOwnsScroll` guard — entering the alt screen calls `clearAgentScrollState()`, which zeroes the gesture accumulator, so `scrolledUp` is false on the takeover frame. A capture-backed test pins that premise down: on the alt screen `viewportY` is 0, the pins still index the untouched normal buffer, and resolving anyway returns a *confident* wrong prompt.

### The bar's height moves the PTY

Showing or hiding the bar changes `--pinned-prompt-h`, which changes `.term-mount`'s box, which trips its `ResizeObserver` → `refit()` → a PTY resize. This is the same accepted trade the review/CI banner strip already makes.

The startup case is designed out: the bar waits for the first pin scan, so a fullscreen-renderer session never mounts it only to hide it again on the replayed alt-screen switch (and never flashes "No prompt yet" before anything has been looked at). What remains is a resize pair when the agent enters and leaves the alternate screen mid-session, for its own full-screen overlays. That's deliberate — SIGWINCH is the contract a TUI already lives under, and the alternative (reserving the row while the bar is hidden) leaves a dead strip above every full-screen view, permanently so on every fullscreen-renderer session.

## Honest degradation

When the agent owns the scroll (mouse tracking on the normal buffer), it repaints its own view and xterm's viewport stays pinned to the bottom wherever the reader actually is. Once they scroll up, the bar says *"prompt unknown"* rather than lying. Codex has no verified echo pattern yet (its TUI buried me in update/trust/hooks dialogs before I could capture one), so it gets no pattern and the bar stays hidden for those sessions rather than guessing — a follow-up, not a silent gap.

## Verification

- `promptPins.browser.test.ts` — real captured PTY bytes → real xterm buffer → correct pins, correct line numbers, correct resolution at each scroll position, input box never pinned (with explicit anti-vacuity assertions that the box row holds text and matches `ECHO`), and the alt-screen premise above. Also asserts the captured agent left `mouseTrackingMode === "none"` and stayed on the normal buffer.
- `promptPins.test.ts` — resolver + scanner units, with named regressions for all three anchor bugs (top-anchor at the bottom, top-anchor in a `baseY === 0` session, bottom-anchor stealing the label while scrolled back).
- `PinnedPrompt.browser.test.ts` — renders the governing prompt, prefers the *earlier* prompt when scrolled back, says "unknown" instead of guessing, expands newest-first, jumps + closes, Escape dismisses, publishes its height, and announces a labelled accessible name.
- Root `bun test` (6293 pass), `ui bun test` (3152 pass), `bun run lint`, `ui bun run check` (0 errors), `ui bun run build`, plus branch-hygiene / feature-catalog / announcement-version / glossary / `check:i18n` gates.

**Not verified:** I could not drive the assembled app end-to-end — the local backend is password-protected and this ran unattended. The xterm integration and the component are each covered above, but the two meeting inside `Viewport.svelte` (the `--pinned-prompt-h` reservation, the refit it triggers, and the alt-screen hide) rest on typecheck plus the existing `--review-banner-h` precedent they mirror. Worth a look when you pull it — particularly with the fullscreen renderer toggled on.

## Notes

- The bar floats (`position: absolute; top: 0`) and takes no flow space. What keeps it off agent output is `.term-mount`'s `margin-top: var(--pinned-prompt-h)`; changing that height trips the existing `ResizeObserver` → refit → PTY resize. Because the reservation is a margin, `.term-mount`'s 4rem floor absorbs it — otherwise the mount's occupied box would exceed the 4rem-floored `.vp-body` and clip the bottom prompt row under the squeezed-recap takeover.
- Height is observed with `bind:offsetHeight`, so it survives a `--ui-scale` change or a `pointer: coarse` flip that no prop reflects.
- Pins re-scan on `onResize`: a reflow renumbers the absolute lines they're defined against, and an ended/parked session has no PTY left to repaint and heal them.
- The history list is a small anchored, **non-modal** popover (`role="dialog"`, no `aria-modal`), so per the Modal & scrim recipe it correctly carries no scrim and dismisses on outside-click/Esc.
- Adds `--shadow-pop` to `app.css` for the popover lift, distilled from a literal that 7 pre-existing popovers hard-code across 6 files. They're left alone rather than swept in on a feature branch, and the token isn't claimed to be the one true shadow (EmojiPicker and the menus carry their own, differing values).
- 44px tap-target floor is applied under `@media (pointer: coarse)` only; on desktop those extra rows would come straight out of visible terminal output, and a mouse pointer has no such floor. Body text uses `--fs-base`, the app-wide chrome size (the 16px floor in `app.css` is scoped to inputs, for iOS focus-zoom).
- The fixture was scrubbed of account e-mail/org/paths using **same-length** placeholders, so every column position and wrap point in the capture is preserved.


scoop:
## What I verified

**Message keys / locale parity.** The six new keys (`pinned_prompt_label`, `pinned_prompt_none`, `pinned_prompt_unknown`, `pinned_prompt_history_title`, `feat_pinned_prompt_title`, `feat_pinned_prompt_body`) exist in both catalogs — `ui/messages/en.json:2639-2642` + `:2732-2733` and `ui/messages/de.json:2639-2642` + `:2732-2733`. Those are the only two catalogs in `ui/messages/`. Every key referenced from `PinnedPrompt.svelte:85,87,91,105,108` and from `ui/src/lib/feature-announcements/entries/v1.43.0-pinned-prompt.ts:6-7` resolves.

**Feature-announcement entry.** `FeatureAnnouncement` is imported from `../../feature-announcements`, matching the sibling entry `ui/src/lib/feature-announcements/entries/v1.43.0-session-tooltip.ts:1`, and the catalog picks it up via the `import.meta.glob` at `ui/src/lib/feature-announcements.ts:47`. Filename/`sinceVersion`/`id` follow the contract documented at `ui/src/lib/feature-announcements.ts:3-23`.

**Identifiers.** `session.agentProvider` is declared at `ui/src/lib/types.ts:816` as `AgentProvider = "claude" | "codex"` (`ui/src/lib/types.ts:2-3`), so the `ECHO` record key `claude` in `promptPins.ts:26` is well-formed and `codex` correctly has no entry. `agentOwnsScroll` exists at `Viewport.svelte:1151`. The CSS custom properties consumed by `PinnedPrompt.svelte` all exist: `--color-head` (`app.css:11`), `--color-panel-2` (`app.css:10`), `--color-inset` (`app.css:12`), `--color-faint` (`app.css:20`), `--fs-meta` (`app.css:82`). `.vp-body` is `position: relative` (`Viewport.svelte:3446`), so the absolutely-positioned `.pp-bar`/`.pp-pop` anchor where the component intends.

**The echo shape.** I checked the raw fixture bytes rather than trusting the comment. Claude Code v2.1.205's *live input box* row is `❯` + U+00A0 (`claude-code-transcript.txt` — 9 lines carry `❯\xa0`), and the two rows carrying `❯` + ASCII space are exactly the two submitted prompts (`ui/src/lib/__fixtures__/claude-code-transcript.txt:30,161`). So `ECHO = /^❯ (\S.*)$/` (`promptPins.ts:26`) does discriminate, and the `RULE` guard covers the half-typed box. That part of the design holds up.

---

## Blocking

### 1. The bar names the wrong prompt — or none — whenever the newest echo is still on the screen rows

`resolvePinnedPrompt` (`ui/src/lib/promptPins.ts:130-137`) selects the last pin with `p.line <= pos.viewportY`, and `viewportY` is the buffer line at the **top** of the viewport. When the reader is at the bottom, `viewportY === baseY`, so **every pin located in the screen rows (`line >= baseY`) is skipped.**

The PR's own fixture test proves the newest pin lands there: `ui/src/lib/promptPins.browser.test.ts:60` asserts `pins[1]!.line >= baseY`. Feeding that same terminal's real bottom position (`viewportY = baseY`) into `resolvePinnedPrompt` returns `pins[0]` — the *previous* prompt — while the reader is looking at prompt 2's answer.

The degenerate case is worse and is the default for every new session. `scanPromptPins`'s own doc comment (`ui/src/lib/promptPins.ts:57-59`) says: *"until a session has produced a screenful of output nothing has scrolled at all (`baseY === 0`), and the newest prompt — the one most worth pinning — is normally still on screen."* With `baseY === 0`, `viewportY === 0`, and every pin at `line > 0`, the loop breaks on the first pin and returns `{pin: null}` → the bar renders `m.pinned_prompt_none()` = **"No prompt yet"** (`PinnedPrompt.svelte:91`) while the operator's question and its answer are both fully visible on screen.

Neither test suite catches this: the unit test at `promptPins.test.ts:113-115` builds `viewportY` synthetically, and the browser test at `promptPins.browser.test.ts:80-82` also passes synthetic `viewportY` values (`second.line + 1`, etc.) rather than the terminal's actual `buffer.active.viewportY`. A test that reads `term.buffer.active.viewportY` after the replay — the value `Viewport.svelte:1817` actually passes — would fail today.

The governing prompt should be the last one echoed at or above the **bottom** visible row (or, minimally, fall back to the newest pin when none precedes `viewportY` and the reader is not scrolled up).

### 2. `margin-top` on `.term-mount` breaks the floor invariant its own comment documents

`Viewport.svelte:3471-3473` adds `margin-top: var(--pinned-prompt-h, 0px)` while leaving `min-height: 4rem`. The comment directly above it (`Viewport.svelte:3462-3465`, unchanged by this PR) states the floor *"equals .vp-body's own min-height (4rem) and must NOT exceed it: a larger floor would force the mount taller than a squeezed body and clip the bottom prompt row via .vp-body's overflow:hidden."*

The mount's occupied box is now `margin-top + max(4rem, 100% − banner − pinned)`. Under the squeezed-recap takeover (`.vp-body` at its own `min-height: 4rem`, `Viewport.svelte:3453`) that is `4rem + pinnedPromptH`, which overflows `.vp-body` (`overflow: hidden`, `Viewport.svelte:3448`) by exactly the bar's height — clipping the bottom prompt row, the outcome the comment says must not happen. It bites for any `.vp-body` height below `4rem + pinnedPromptH`, not only at the exact floor.

---

## Non-blocking

### 3. The published height never updates after mount

`PinnedPrompt.svelte:35-38` measures `barEl.offsetHeight` in an `$effect` whose only reactive dependency is `void pins.length`. Both sibling banners in the same directory use `bind:offsetHeight={height}` on the element so the value tracks reflows — `CiRunningBanner.svelte:50` (with the seeding effect at `:40` and the comment at `:32` saying "bind:offsetHeight below keeps it current across reflows") and `ReviewInFlightBanner.svelte:244` / `:232`. Here the same seed-effect is present but the `bind:offsetHeight` is not. A `--ui-scale` change (the bar's `--fs-base`/`--fs-meta` are `calc(… * var(--ui-scale))`, `app.css:82`) or a `(pointer: coarse)` flip (`PinnedPrompt.svelte` media query, `min-height: 44px`) changes the bar's height without changing `pins.length`, leaving `--pinned-prompt-h` stale — xterm then either overlaps the bar or leaves a gap. This is the guard-present-on-one-sibling-path-missing-on-another pattern.

### 4. No rescan on terminal resize

`schedulePinScan` is invoked only from `term.onWriteParsed` (`Viewport.svelte:1880`) and once at attach (`Viewport.svelte:1890`). xterm reflows wrapped lines on resize, which shifts the absolute indices `PromptPin.line` is defined against (`promptPins.ts:44`). For a live session the agent repaints on SIGWINCH and the write handler heals it within 250 ms; for an ended/parked session there is no PTY to repaint, so the pins and the `onjump` scroll targets (`Viewport.svelte:2591`) stay wrong indefinitely after a window resize or phone rotation. A `term.onResize(schedulePinScan)` subscription would close it.

### 5. The `--shadow-pop` comment is factually wrong

`app.css:88-90` claims EmojiPicker's `.ep` and AutomationPanel's `.auto-pop` "still carry the literal it was distilled from." `AutomationPanel.svelte:180` does (`0 6px 20px rgba(0, 0, 0, 0.45)`), but `EmojiPicker.svelte:148` is `0 8px 30px rgba(0, 0, 0, 0.6)` — a different shadow entirely. And the literal is carried by six other components, not "those two": `GitRail.svelte:1057`, `GitRail.svelte:1127`, `Stepper.svelte:234`, `TimePopover.svelte:174`, `TopBarHeldBadge.svelte:349`, `ViewportTabBar.svelte:439`.

### 6. Two contradictory statements about Claude Code's scroll regime in one file

The new comment at `Viewport.svelte:381-383` says Claude Code v2.1.205 "does neither — it stays on the normal buffer with mouse tracking off" (corroborated by `promptPins.browser.test.ts:93-94`). The comment at `Viewport.svelte:1146-1148`, which this PR leaves untouched, still says "Claude Code does the latter — it stays on the normal buffer (keeping scrollback) but grabs the wheel." One of the two is stale; the file now asserts both.

### 7. The "never mistakes the live input box" fixture test is vacuous

`promptPins.browser.test.ts:71-75` asserts the box is not picked up as a third pin, but at the end of the replay the box row is `❯` + U+00A0 (verified in the fixture bytes — the two ASCII-space `❯ ` rows are the two prompts at `claude-code-transcript.txt:30,161`). `"❯\u00A0".trimEnd()` is `"❯"`, which `ECHO` (`promptPins.ts:26`) rejects on the `\S` requirement before `RULE` (`promptPins.ts:31`) is ever consulted. So the `RULE` guard — the thing that actually protects against a half-typed box, and the only defense once the box has content — is exercised only by the synthetic unit test at `promptPins.test.ts:78-83`. Capturing a fixture with text sitting in the box would make this test guard what it claims to.

---

- `ui/src/lib/components/Viewport.svelte:1815-1820`: `recomputePin` reads `term.buffer.active.viewportY` while pins index `term.buffer.normal` (`Viewport.svelte:1837-1842`); on the alternate screen `agentOwnsScroll` is true and, when `scrolledUp` is false, `resolvePinnedPrompt` compares normal-buffer pin lines against the alt buffer's `viewportY` (always 0) and returns "No prompt yet". Dormant only because Claude Code v2.1.205 never leaves the normal buffer.
- `ui/src/lib/components/Viewport.svelte:2591`: `termRef?.scrollToLine(line)` moves xterm's own viewport, which is a no-op in the agent-owns-scroll regime (`Viewport.svelte:1151-1152`). If a future Claude Code re-enables mouse tracking — as `Viewport.svelte:1146` claims it already does — the history popover's jump would silently do nothing.

## Stated limitations
- `Terminal.scrollToLine` could not be verified against the `@xterm/xterm` typings: `node_modules` is not installed in this worktree. `termRef` is typed as `Terminal` (`Viewport.svelte:320`), so `tsc` in CI covers its existence and signature; I assumed it as written.
- I did not verify the fixture against a live `claude` process; the assertion that v2.1.205 renders `❯ <text>` is verified only against the captured bytes in this repo, which do support it.

scoop:
## Re-review of the previous revision's 7 findings

All seven are genuinely addressed. Verified each:

1. **Bottom-row anchor** — `resolvePinnedPrompt` now computes `const bottom = pos.viewportY + pos.rows - 1` and walks pins up to it (`ui/src/lib/promptPins.ts:148-153`). The two regressions are pinned by tests at `ui/src/lib/promptPins.test.ts:168-194` and by the real-capture assertion `expect(at(term.buffer.normal.baseY)).toBe(PROMPT_2)` at `ui/src/lib/promptPins.browser.test.ts:98`. Fixed — though the bottom anchor introduces the inverse error while scrolled up; see finding 3 below.
2. **`.term-mount` floor** — `min-height: max(0px, calc(4rem - var(--pinned-prompt-h, 0px)))` at `ui/src/lib/components/Viewport.svelte:3486`. With `margin-top: var(--pinned-prompt-h)` (3484) and `height: calc(100% - review - pinned)` (3485), the occupied box under the squeeze is `pinned + (4rem - pinned) = 4rem`, exactly `.vp-body`'s own `min-height: 4rem` at `Viewport.svelte:3461`. Invariant restored.
3. **Height re-measure** — `bind:offsetHeight={height}` on `.pp-bar` at `ui/src/lib/components/viewport/PinnedPrompt.svelte:72`, matching the sibling banners' idiom. The `pins.length` effect is gone. Fixed.
4. **Reflow rescan** — `const resizeSub = term.onResize(schedulePinScan)` at `Viewport.svelte:1892`, disposed at 1946. Fixed.
5. **`--shadow-pop` comment** — I grepped the literal: it appears in exactly 7 places (`AutomationPanel.svelte:180`, `GitRail.svelte:1057`, `GitRail.svelte:1127`, `Stepper.svelte:234`, `TimePopover.svelte:174`, `ViewportTabBar.svelte:439`, `TopBarHeldBadge.svelte:349`) — the seven the comment at `ui/src/app.css:87-93` now names. `EmojiPicker.svelte:148` is `0 8px 30px rgba(0,0,0,0.6)`, as the comment now says. Accurate.
6. **Stale `agentOwnsScroll` comment** — `Viewport.svelte:1141-1152` rewritten. But two *other* copies of the same stale claim survive in the same file; see finding 4.
7. **Vacuous input-box test** — the capture now ends with `❯ this question is still being typed` and the test asserts anti-vacuity: `expect(rows[box]).toMatch(/^❯[ \u00a0]\S/)` and `expect(rows[box - 1]).toMatch(/^─{3,}$/)` (`promptPins.browser.test.ts:83-84`). I confirmed against the raw fixture that the box's separator with text in it is a **plain space** (`M-bM-^]M-/ ` in `cat -A`), so ECHO really does match the row and only the RULE guard rejects it. Fixed, and the fixture genuinely exercises the guard.

Other things I checked and found correct:

- **i18n parity**: all six new keys exist in both catalogs — `ui/messages/en.json:2639-2642` + `2732-2733` and `ui/messages/de.json:2639-2642` + `2732-2733`.
- **Announcement gate**: `sinceVersion: "1.43.0"` (`ui/src/lib/feature-announcements/entries/v1.43.0-pinned-prompt.ts:5`) is the next version — `package.json:3` holds `1.42.0`, which is what `feature-announcements.ts:11-16` requires. Filename version matches, `id: "pinned-prompt"` is unique across `entries/`, and the registry auto-globs (`feature-announcements.ts:46-49`), so no index edit is needed.
- **Identifiers**: `"claude"` is a valid `AgentProvider` (`ui/src/lib/types.ts:2`); `session.agentProvider` exists (`types.ts:696`); `--shadow-pop` is defined once (`app.css:95`) and consumed once (`PinnedPrompt.svelte:205`); `--pinned-prompt-h` is published on `.vp-body` (`Viewport.svelte:2589`) and read only by `.term-mount` (3484-3486).
- **Scan cost**: xterm is constructed with no `scrollback` option (`Viewport.svelte:1308-1324`), so the comment's "xterm's default 1000-line scrollback cap" at 1827 is right.

---

## New: the alternate screen breaks this feature, and it is reachable today

This is not hypothetical. `src/service.ts:1226-1235` spawns the main session with `CLAUDE_CODE_NO_FLICKER=1` (the fullscreen renderer, which uses the **alternate screen buffer** — see `docs/spikes/1043-fullscreen-alt-screen.md`) whenever `tuiFullscreen` is on, and that is a live user-facing toggle in Settings (`ui/src/lib/components/Settings.svelte:1652-1661`, wired through `putTuiFullscreen` at `ui/src/lib/api.ts:279`). The default is the classic renderer, but the toggle ships.

Under that renderer, for a `provider === "claude"` session:

- `pinsSupported` is true (`Viewport.svelte:442`), so `PinnedPrompt` mounts (`Viewport.svelte:2591`) and reserves a terminal row.
- `scanPromptPins` reads `term.buffer.normal` (`Viewport.svelte:1841-1846`), but every prompt echo is painted into the *alternate* buffer. `promptPins` stays `[]` forever, so `resolvePinnedPrompt` returns `{pin: null, uncertain: false}` (`promptPins.ts:154`) and the bar renders `m.pinned_prompt_none()` — "No prompt yet" (`PinnedPrompt.svelte:87`) — after the operator has asked five questions. That is precisely the "a wrong pin is worse than no pin" outcome the module's own doc at `promptPins.ts:44-45` says it refuses to ship.
- Independently, `recomputePin` reads `const b = term.buffer.active` (`Viewport.svelte:1817`) and passes `b.viewportY` into a function whose pins are indexed against `buffer.normal`. On the alternate buffer `viewportY` is always 0, so `bottom = rows - 1` and `resolvePinnedPrompt` names whichever pin happens to fall in the first screenful of the *normal* buffer, regardless of what is on screen. Two index spaces, one variable. `rescanPins` gets this right (`buffer.normal`, line 1842); `recomputePin` does not. The doc at `promptPins.ts:139-141` asserts "xterm's viewport is pinned to the bottom" when the agent owns the scroll — that holds for mouse-tracking on the normal buffer, not for the alternate screen, where it is pinned to the *top* of a different buffer.

Entering the alt screen fires `onBufferChange` → `clearAgentScrollState()` → `recomputeScrolled()` → `recomputePin()` (`Viewport.svelte:1871-1874`, 1868), and `clearAgentScrollState` zeroes `scrollDepth` so `scrolledUp` goes false — meaning the `agentOwnsScroll && scrolledUp` uncertainty guard at `promptPins.ts:147` does *not* fire. The wrong-buffer resolve runs on the very first frame of the takeover.

The fix for both: gate on `term.buffer.active.type` — hide the bar (or report `uncertain`) on the alternate screen, and resolve against `term.buffer.normal` so the pin indices and the viewport position come from the same buffer.

## The bottom anchor mislabels a full screen while scrolled up

`resolvePinnedPrompt` names the newest pin whose echo is anywhere at or above the bottom visible row (`promptPins.ts:148-153`). Take the browser fixture's own arithmetic: at `viewportY = second.line - rows` the bar says `PROMPT_1` (asserted at `promptPins.browser.test.ts:107`), but one row further down, at `viewportY = second.line - rows + 1`, `bottom === second.line` and the bar flips to `PROMPT_2` — while 29 of the 30 visible rows are still prompt 1's answer and the only row belonging to prompt 2 is its bare echo on the last line.

So the doc claim at `promptPins.ts:135-137` — "falls back to the older one only once the newer echo has scrolled off the bottom edge — which is exactly when the reader is looking at the older prompt's output" — is false for the whole `rows`-high band above that edge. This is the same class of error the previous revision fixed, mirrored: a top anchor is wrong when parked at the bottom, and a bottom anchor is wrong when scrolled up. The rule that is right in both is the one the position struct already carries the input for: anchor on the bottom row when `!scrolledUp`, on the top row when `scrolledUp`.

Non-blocking on its own — the bar is still far better than nothing — but it undercuts the feature's stated purpose during exactly the scrollback reading the task describes.

## Stale comments

The rewritten `agentOwnsScroll` comment (`Viewport.svelte:1146-1152`) now says Claude Code does *neither* alt-screen nor mouse tracking, backed by `expect(term.modes.mouseTrackingMode).toBe("none")` at `promptPins.browser.test.ts:120`. Two untouched comments in the same file still say the opposite:

- `Viewport.svelte:1318-1319`: "Claude Code runs as a TUI with mouse tracking on, which hands mouse drags to the app…"
- `Viewport.svelte:1673-1674`: "Claude Code runs as a full-screen TUI with mouse tracking on (it stays on the normal buffer, keeping scrollback, but grabs the wheel)"

The file is in the diff, and the PR is what made them contradictory. Same fix as prior finding 6.

Separately, two comments this PR *adds* describe the pre-fix top-anchor design:

- `Viewport.svelte:399-400`: "Operator prompts located in the terminal's committed scrollback, and which of them governs **the top of the viewport**." Both halves are now wrong — the scan deliberately spans the screen rows too (`promptPins.ts:52-59`), and resolution is bottom-anchored.
- `Viewport.svelte:1813-1814`: "Which operator prompt governs **the top of the viewport**." Inserting this block also orphaned the pre-existing comment at 1810-1812 ("track scroll position so we can offer a jump-to-bottom button…"), which now sits above `recomputePin` instead of above `recomputeScrolled` at 1859.

## "In-flow" is not what the bar is

`PinnedPrompt.svelte:124-126` ("In-flow strip") and `Viewport.svelte:2592-2594` ("In-flow (the terminal reserves --pinned-prompt-h)") both describe a bar that is `position: absolute; top: 0` (`PinnedPrompt.svelte:126-134`). It takes no flow space at all; the reservation is done entirely by `.term-mount`'s `margin-top`. The mechanism is sound — the wording just says the opposite of what the CSS does, and in a codebase whose comments are this load-bearing that will mislead the next reader of the 3477-3483 invariant.

## Could not verify

- `Terminal.scrollToLine(line)` (`Viewport.svelte:2599`) — `node_modules` is not installed in this worktree, so I could not open xterm's typings. It is a documented xterm.js API and TS/CI will catch it if not; assumed as written.
- Whether Claude Code renders its interactive select menus (`❯ 1. Yes` style option markers) at column 0 outside a rule frame. If it does, `scanPromptPins` would mint a bogus pin from a menu option. The fixture contains no dialog, so I could neither confirm nor refute this against ground truth in this repo; worth a capture with a permission prompt open.

- `ui/src/lib/components/Settings.svelte:1665` reuses `m.settings_usage_hold_on()` / `settings_usage_hold_off()` as the on/off label for the `tuiFullscreen` switch — message keys named for the usage-hold control, not this one.

scoop (commented):
## Re-review of the previous revision's 7 points

All seven are genuinely addressed. Verified each against the code, not the diff summary:

1. **Alt-screen resolution against the wrong buffer** — fixed. `recomputePin` (`ui/src/lib/components/Viewport.svelte:1830-1843`) now early-returns `{pin: null, uncertain: true}` when `altScreen` is set, and otherwise reads `term.buffer.normal` (`Viewport.svelte:1836`), the same buffer `rescanPins` indexes (`Viewport.svelte:1859-1864`). `altScreen` is seeded at attach and maintained by `term.buffer.onBufferChange`, and `syncAltScreen()` is ordered *before* `clearAgentScrollState()`/`recomputeScrolled()` in that handler (`Viewport.svelte:1894-1898`), so the takeover frame can no longer resolve against the alt buffer. The regression is pinned by `ui/src/lib/promptPins.browser.test.ts:130-154`, which replays `\x1b[?1049h` into a real xterm and asserts the old code produced `uncertain: false` + `PROMPT_1` — a confident falsehood.

2. **Bar mounted for fullscreen-renderer sessions** — fixed. `showPinnedPrompt` is now `tab === "term" && pinsSupported && !altScreen` (`Viewport.svelte:455`). I confirmed the two renderers at `src/service.ts:1226-1235`: `tuiFullscreen` → `CLAUDE_CODE_NO_FLICKER=1` (alt screen), default → `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1` (normal buffer). `tuiFullscreen` defaults to `false` (`src/config.ts:675`, `test/server-settings.test.ts:840-846`), so the feature is live on the default path and dark under the opt-in — which is the correct split.

3. **Bottom-row anchor mislabelling a scrolled-up reader** — fixed. `resolvePinnedPrompt` now picks the anchor by `scrolledUp`: `pos.scrolledUp ? pos.viewportY : pos.viewportY + pos.rows - 1` (`ui/src/lib/promptPins.ts:164`). That matches the `scrolledUp` contract, which for the xterm-owned regime is exactly `viewportY < baseY` (`ui/src/lib/scrollAffordance.ts:38`), so the two branches partition cleanly. The doc comment at `promptPins.ts:135-156` was rewritten to describe both anchors and is now accurate. Both directions are covered: `promptPins.test.ts:177-180` (bottom anchor would steal the label) and `promptPins.test.ts:143-149` (top anchor would name the previous prompt).

4. **Stale `mouse tracking on` comments** — fixed. `grep -n "mouse tracking on"` across the tree returns nothing, and the two sites are reworded at `Viewport.svelte:1330-1334` and `Viewport.svelte:1685-1690`. The new `agentOwnsScroll` note (`Viewport.svelte:1155-1164`) is backed by `promptPins.browser.test.ts:120-124`, which asserts `buffer.active.type === "normal"` and `modes.mouseTrackingMode === "none"` on a verbatim v2.1.205 PTY capture. This is consistent with `src/service.ts:1222-1223`, which describes modes 1000/1002/1003 as a *fullscreen*-renderer concern.

5. **"committed scrollback" / "top of the viewport" comments and the orphaned block** — fixed. The state comment at `Viewport.svelte:399-400` no longer says "committed scrollback"; `promptPins.ts:41-44` explicitly says `end` spans `baseY + rows`, matching `Viewport.svelte:1862`. The pre-existing scroll-tracking comment now sits directly above `recomputeScrolled` (`Viewport.svelte:1878-1881`), and the new pin comment is its own block at `1821-1829`.

6. **"In-flow strip" comments** — fixed. `PinnedPrompt.svelte:6-14` and `:69-71` now say the bar is absolutely positioned and takes no flow space, and name `.term-mount`'s `margin-top` as the reservation. The `.term-mount` invariant at `Viewport.svelte:3494-3514` documents the `4rem - pinned` floor, and `min-height: max(0px, calc(4rem - var(--pinned-prompt-h, 0px)))` does keep `margin + min-height <= 4rem`, i.e. inside `.vp-body`'s own `min-height: 4rem` (`Viewport.svelte:3489`).

7. **Accessible name / bare count** — fixed. `.pp-label` no longer carries `aria-hidden` (`PinnedPrompt.svelte:81`), so the button's name is "You asked …"; `.pp-count` is `aria-hidden="true"` and paired with a visually-hidden `m.pinned_prompt_count({count})` phrase (`PinnedPrompt.svelte:88-91`, `.pp-sr` at `:170-177`). Guarded by `PinnedPrompt.browser.test.ts:91-99`.

## Identifiers and catalogs I resolved

- All five runtime message keys and both announcement keys exist in **both** catalogs with the same `{count}` placeholder: `ui/messages/en.json:2639-2643,2733-2734` vs `ui/messages/de.json:2639-2643,2733-2734`. No parity gap.
- `m.pinned_prompt_count` is only rendered when `pins.length > 1` (`PinnedPrompt.svelte:88`), so the un-pluralized "{count} prompts" string is never shown with `count === 1`.
- `AgentProvider` is `"claude" | "codex"` (`ui/src/lib/types.ts:2-3`); `ECHO` keys only `claude` (`promptPins.ts:32`), and `session.agentProvider` is the real field (`ui/src/lib/types.ts:696`).
- `mainSessionRendererEnv`, cited in the new comment at `Viewport.svelte:407-409`, exists at `src/service.ts:1226`.
- The announcement fragment follows the contract in `ui/src/lib/feature-announcements.ts:3-24`: `id` unique, filename `v1.43.0-pinned-prompt.ts` matches `sinceVersion: "1.43.0"`, and `package.json` is still at `1.42.0`, so `sinceVersion > last release` as required. Sibling entries already use `v1.43.0-*`.
- The new `--shadow-pop` token (`ui/src/app.css:95`) really is the literal seven popovers hard-code: `AutomationPanel.svelte:180`, `ViewportTabBar.svelte:439`, `TopBarHeldBadge.svelte:349`, `GitRail.svelte:1057` and `:1127`, `Stepper.svelte:234`, `TimePopover.svelte:174` — all `0 6px 20px rgba(0, 0, 0, 0.45)`, equivalent to `rgb(0 0 0 / 0.45)`. The comment's count and file list check out.
- Every CSS custom property `PinnedPrompt.svelte` consumes is defined: `--color-head`, `--color-panel`, `--color-panel-2`, `--color-inset`, `--color-line`, `--color-faint`, `--fs-base`, `--fs-meta` (`ui/src/app.css:9-20,82-83`). `.pp-bar`'s `position: absolute` resolves against `.vp-body`, which is `position: relative` (`Viewport.svelte:3482`).
- `PromptPin.line` is an absolute buffer index, and `onjump` feeds it to `term.scrollToLine` (`Viewport.svelte:2627`), whose parameter is an absolute `ydisp` target — consistent with `scanPromptPins` returning `i` from a loop over absolute indices (`promptPins.ts:88`). Verified the contract on the producer side; see limitations below for the consumer side.

## Findings (all non-blocking)

Three small things, none of which block.

**1. Focus is orphaned after picking a prompt from the history.** `jump()` (`PinnedPrompt.svelte:63-66`) sets `open = false`, which unmounts the `.pp-item` button the user just clicked. When the focused element is removed the browser moves focus to `<body>`, so the operator's next keystroke does not reach the terminal — they have to click back into it. The Escape path already handles this correctly by refocusing the bar button (`PinnedPrompt.svelte:45`); `jump()` should do the same, or better, hand focus back to the terminal.

**2. The seed comment at `Viewport.svelte:1917-1920` describes a mechanism that cannot fire.** It says "a fullscreen-renderer session is ALREADY on the alt screen when we attach, so onBufferChange never fires for it and the bar would otherwise show." Both halves are wrong, and I checked: `syncAltScreen()` runs at `Viewport.svelte:1923`, and the only two `term.write` call sites in the file are inside async callbacks — the PTY data handler (`Viewport.svelte:1380`) and the connection-ended handler (`Viewport.svelte:1395`). No byte has been parsed at seed time, so `term.buffer.active.type` is unconditionally `"normal"` there and the seed can never observe the alt screen. Conversely, if the replay *does* carry `\x1b[?1049h`, `onBufferChange` fires and the gate works — which is in fact what protects the fullscreen case. The seed's real (and worthwhile) job is resetting `altScreen` for a fresh terminal. Please correct the comment so the next reader does not trust a guard that isn't there.

**3. Toggling the bar on alt-screen entry resizes the PTY under a running full-screen tool.** `showPinnedPrompt` flips false when the agent enters the alternate screen (`Viewport.svelte:455`), which drives `--pinned-prompt-h` to `0px` (`Viewport.svelte:2616`), which changes `.term-mount`'s `margin-top` and `height` (`Viewport.svelte:3512-3513`), which trips the `ResizeObserver` on that element (`Viewport.svelte:1813-1819`) → `refit()` → `fit.fit(); c.resize(term.cols, term.rows)` (`Viewport.svelte:1469-1473`). So every time a classic-renderer Claude session shells out to a full-screen tool the child gets a SIGWINCH mid-draw, and another on exit. The `.term-mount` comment at `3497` already names this coupling as "the intended reflow" for the banner; here it is a side effect nobody asked for. Either keep the row reserved while the bar is hidden, or say explicitly in the comment that the alt-screen resize is accepted.

## Stated limitations

- `@xterm/xterm` is not installed in this checkout (no `node_modules`, no typings to grep), so I could not confirm `Terminal.scrollToLine`'s signature or its absolute-vs-relative semantics against the shipped package. I assumed it takes an absolute buffer line as `Viewport.svelte:2627` implies; the repo's own tests exercise `scanPromptPins`/`resolvePinnedPrompt` but never `scrollToLine`.
- Whether the herdr attach replay carries the `\x1b[?1049h` that makes `onBufferChange` fire for an already-fullscreen session is decided inside the external `herdr` multiplexer, not in this repo. `ui/src/lib/pty.ts:33-37` says a fresh attach `--takeover` "repaints the live session", which strongly suggests it does, but I could not verify it. If it does not, the alt-screen gate never engages for those sessions (see finding 2). Not asserted as wrong.

- `ui/src/lib/scrollAffordance.ts:10-13` still says the agent-owned regime is "alternate screen, or mouse-tracking on the normal buffer **like Claude Code**", which the new, capture-verified comment at `Viewport.svelte:1155-1164` now contradicts. Same stale claim as the ones this PR fixed, in a file the PR does not touch.

Issue #1579, opened 2026-07-09
In-app Fix for `herdr_offline` spawns a daemon inside Shepherd's cgroup; prefer the systemd unit when it exists


scoop:
## Problem

#1574 added `diagnostics_hint_herdr_offline` → `HERDR_SERVE`, which is auto-runnable from the in-app Fix endpoint (deliberately not `GUIDANCE_ONLY` — starting a daemon is unprivileged and clears the check unattended).

When Fix runs it, the daemon is spawned as a **child of the Shepherd server process**, so it lives in Shepherd's cgroup. Consequences:

1. **Not durable.** `systemctl --user restart shepherd` tears down the cgroup and takes the daemon with it. Panes survive (targets outlive the server) and re-orphan, but the host silently returns to `herdr: offline`.
2. On a host provisioned by `deploy/provision.ts`, a proper `herdr.service` (`Restart=always`) already exists — Fix should use it rather than hand-rolling a detached process that systemd does not own.

## Related hazard (currently unreachable, worth noting)

`HERDR_SERVE` detaches via `setsid`, falling back to `nohup` on macOS. `nohup … &` stays in the parent's process group, and `defaultRunRemediation` (`src/diagnostics.ts`) SIGKILLs the child's **entire process group** on its 120s timeout. Today this cannot fire — `HERDR_SERVE` self-terminates within ~10s (10 × 1s poll, then `exit 1`), so the timer is always cleared first, and the daemon's stdio is redirected to `/dev/null` so it does not hold the parent's pipe. But the two facts are one edit apart from colliding.

## Suggested fix

In the Fix path, prefer `systemctl --user start herdr` when the unit exists (and `--user reset-failed herdr` first, since a unit that previously crash-looped against a held socket can be parked `failed`), falling back to `HERDR_SERVE` on hosts with no unit (macOS, hand-rolled installs).

Consider also: `deploy/provision.ts` uses `enable --now herdr`. With `Type=simple`, the start job reports success as soon as the process forks, even if `herdr server` then exits immediately because a non-systemd daemon already holds the socket — leaving the unit to crash-loop into `failed` while the check still reads `ok`. Not reachable on the normal service path, but it makes re-provisioning over an ad-hoc daemon fragile.

Deferred out of scope during #1574.

scoop:
✅ PR #1600 merged.

Issue #1578, opened 2026-07-09
diagnostics: `herdr_outdated` reinstalls the binary but leaves the old daemon running


scoop:
## Problem

`src/remediations.ts`:

```ts
diagnostics_hint_herdr_outdated: HERDR_INSTALL,
```

`HERDR_INSTALL` swaps the binary on disk. It does not touch the **running** daemon, which keeps serving the old version. Since the `herdr` version probe reads `herdr --version` (the new binary) while the daemon is still the old one, the host is left in a split state, and any check that reflects the *running* server's version keeps reporting `outdated` until something restarts it.

Contrast the two hints fixed in #1574:

- `herdr_missing` → installs **and** starts (`${HERDR_INSTALL} && (${HERDR_SERVE})`)
- `herdr_offline` → starts (`HERDR_SERVE`)
- `herdr_outdated` → installs only ← this issue

## Considerations

A naive restart is not obviously correct: panes/targets outlive the server, and herdr ships `herdr update --handoff` precisely to swap a live server without dropping them. `src/herdr-update.ts` already implements the handoff + recovery dance. So the fix is probably to route `herdr_outdated` through that path rather than to bolt a restart onto the install.

No harness scenario covers this today (a pinned outdated herdr build is listed as a deferred provisioning gap in `ci/onboarding-harness/scenarios.ts`), so it is currently invisible to CI.

## Suggested fix

Decide between:
1. `herdr_outdated` → `HERDR_INSTALL && herdr update --handoff` (or delegate to `HerdrUpdateService`), preserving live panes; or
2. leave it as-is and document explicitly that the operator must restart, since a live-but-old daemon is degraded rather than broken.

Deferred out of scope during #1574.

scoop:
✅ PR #1603 merged.

scoop:
Resolved by PR #1603 (commit `12d96f03`, merged to `main`).

**Fix (Option 1 from this issue):** `src/remediations.ts` now routes `diagnostics_hint_herdr_outdated` through `buildUpdateScript(...)` → `herdr update --handoff`, the atomic update+handoff that preserves live panes (with the same retry + detached-relaunch recovery as `HerdrUpdateService`). A no-op update leaves the binary old (still `warning`); a killed server reads `offline` — verified by `DiagnosticsService.fix()`'s re-probe.

[...]

Issue #1577, opened 2026-07-09
Onboarding harness: a failing `optional` remediation aborts the whole apply, redding the release gate


scoop:
## Symptom

`scripts/onboarding-gate.sh` currently exits 1 on `node-too-old` (ADVICE GAP: `node` stays `warning` after a verbatim apply). It reproduces on **unmodified `main`** (`f4f34cc1`) — it is not caused by any open branch.

## Root cause

The `codex` installer is broken upstream right now:

```
$ curl -fsSL https://chatgpt.com/codex/install.sh | CODEX_NON_INTERACTIVE=1 sh
Could not find Codex package or platform npm release assets for Codex 0.143.0.
# non-zero exit
```

`applyVerbatim` bails on the **first** non-zero exit:

```ts
// ci/onboarding-harness/apply.ts:74-78
for (const cmd of remediationsFor(snapshot)) {
  const r = await driver.exec(name, ["sh", "-c", cmd]);
  if (r.code !== 0) return false;
}
```

`remediationsFor` includes any check whose state is `!== "ok"` — and `codex` reports state `optional`, which qualifies. In the snapshot's check order, `codex` sorts **before** `node`. So the codex install fails, the loop returns early, and `node`'s remediation never runs.

## Evidence

Instrumented run of `node-too-old` on `images:debian/12`:

- `git`: `error` → `ok` (the apply *did* run and got past git)
- `node`: `warning` → `warning` (its command never executed)
- `herdr`: `ok` in both snapshots

The `node` remediation itself is fine — run standalone in a fresh `debian/12` instance it installs Node v24.18.0 and symlinks it first on PATH, exit 0.

Timeline confirming it is upstream flapping, not code: `node-too-old` passed on `main` at ~T, then failed on `main` at ~T+20min, same commit.

## Why it matters

A broken **third-party, optional** installer can red the **release gate** for an entirely unrelated scenario. The gate also has no retry, so a single transient download failure blocks a release. Separately observed in the same session: `install-e2e-service` failed once on `error: Fail extracting tarball for "node-pty"` and passed on retry.

## Suggested fix

1. A non-zero exit from a remediation belonging to an `optional`-state check should not be fatal to the apply loop — collect and report it, then continue. (Arguably `optional` should not be in `remediationsFor`'s apply set at all.)
2. Consider a bounded retry for network-fetch remediations, so one transient download does not red the gate.

Found while fixing #1574.

scoop:
✅ PR #1599 merged.

PR #1580, opened 2026-07-09
fix(herdr): provision a running herdr daemon so the liveness check can go green (#1574)

### LINKED ISSUE #1574: Onboarding harness: nightly regression detected
# Onboarding Gap Report

**3 / 6 scenarios reached green.** (3 detection-only, by design — excluded.)

| herdr-missing | images:archlinux | yes | verbatim | no | ADVICE GAP |
| install-e2e | images:ubuntu/24.04 | no | verbatim | no | INSTALL GAP |
| install-e2e-service | images:ubuntu/24.04 | no | verbatim | no | INSTALL GAP |

## Gaps

- **herdr-missing** (ADVICE GAP)
- **install-e2e** (INSTALL GAP) — herdr want=ok got=error
- **install-e2e-service** (INSTALL GAP) — herdr want=ok got=error

_Nightly run: 2026-07-10T01:38:56.510Z_

scoop:
Fixes #1574.

## Problem

#1562 made the `herdr` diagnostic probe daemon **liveness** (`herdr agent list`), not just binary presence. But nothing in the product ever *starts* that daemon, and no remediation existed for the new `diagnostics_hint_herdr_offline` hint. A freshly-installed host was therefore permanently red, and three nightly onboarding scenarios failed.

The bug had a documented cause. `src/herdr-update.ts` asserted, citing a 0.6.x-era design doc, that *"any herdr CLI call auto-spawns the daemon."* **That is false on 0.7.3.** Verified in a clean `images:ubuntu/24.04` instance:

| step | result |
| --- | --- |
| fresh host, server never started | `status: not running` |
| `herdr agent list` | exit 1, `NotFound`, **no daemon spawned** |
| liveness probe (exit-code only, as `defaultHerdrLiveness` does) | exit 1 → `offline` → **error** |
| `setsid herdr server` + 3s | `status: running` |
| liveness probe again | exit 0 → **ok** |

So the probe was right and the install path was incomplete.

## Fix

- **`src/remediations.ts`** — new exported `HERDR_SERVE`: idempotent (a live daemon short-circuits), detaches via `setsid` with a `nohup` fallback (macOS has no `setsid`), and polls until the daemon actually answers. Mapped to `herdr_offline`, and chained into `herdr_missing` as `` `${HERDR_INSTALL} && (${HERDR_SERVE})` ``. The subshell is load-bearing: `HERDR_SERVE` opens with `export PATH=…;`, so a bare `&&` would gate only that first statement and the rest would run even on a failed install.
- **`deploy/herdr.service`** — `Restart=always` systemd **user** unit. `shepherd.service` is ordered behind it with `Wants=` (not `Requires=`), so a dead herdr degrades to `offline` rather than cascade-failing Shepherd.
- **`deploy/provision.ts`** — `installService()` writes and `enable --now`s the unit *before* delegating to `update.sh` (which starts Shepherd); `buildOnly()` (macOS + the harness's `SHEPHERD_NO_SERVICE` path) starts a detached daemon as its first action, so a failure surfaces before the slow build.
- **`src/herdr-update.ts`** — corrects the falsified comment. Its `setsid … server` fallback is not belt-and-braces; it is the only repair path.

## Verification

All three previously-red scenarios pass, each covering a different half of the fix:

| Scenario | Was | Now |
| --- | --- | --- |
| `herdr-missing` (archlinux) | ADVICE GAP | **PASS** |
| `install-e2e` (ubuntu, no systemd) | INSTALL GAP | **PASS** |
| `install-e2e-service` (ubuntu, systemd) | INSTALL GAP | **PASS** |

`herdr-missing` proves the chained remediation reaches green through the harness's single-apply path (`run.ts` applies `herdr_missing` exactly once, with no second round — so installing without starting could never go green). `install-e2e` proves the detached start; `install-e2e-service` proves the unit plus the `enable --now herdr`-before-`update.sh` ordering, health-checked through the live unit.

Unit suite: 6319 pass / 8 skip / 0 fail. Lint clean.

The behavioral tests in `test/remediations.test.ts` run the **production** string (`REMEDIATIONS.diagnostics_hint_herdr_missing`, with the network-touching install clause substituted) through a real shell against a stubbed herdr on a hermetic `PATH`. Reintroduce the bare `&&` and the install-failure test fails — including on a CI runner with no herdr installed.

## Known: the release gate is red, and not because of this branch

`scripts/onboarding-gate.sh` currently fails on `node-too-old` **on unmodified `main` too**. The codex installer is broken upstream today, and `applyVerbatim` bails on the first non-zero exit while the `optional`-state `codex` check sorts before `node`. Filed as #1577. Also filed: #1578 (`herdr_outdated` leaves the old daemon running) and #1579 (in-app Fix should prefer the systemd unit).

[no-feature-entry] — server/deploy plumbing, ships no user-facing UX. The `herdr_offline` EN+DE message keys already exist from #1562.

### DIFF
--- a/ci/onboarding-harness/README.md
+++ b/ci/onboarding-harness/README.md
@@ -23,6 +23,12 @@ the auto-fixable checks reach `ok`:
 herdr  bun  node  git  claude
 ```
 
+`herdr: ok` means the daemon **answers**, not merely that the binary parses a `--version`
+(#1562 added a liveness probe; #1574 made the installer satisfy it). `deploy/provision.ts`
+therefore leaves a running server: a `Restart=always` user unit (`deploy/herdr.service`) on
+the service path, a detached `herdr server` on this `SHEPHERD_NO_SERVICE` path. herdr does
+**not** auto-spawn its daemon on a CLI call.
+
 `gh` and `tailscale` stay non-ok on a throw-away host (no tailnet, no gh login) and are
 intentionally excluded from `expect` — success is scoped to the installer's auto-fixable set,
 same as every other scenario.
--- a/ci/onboarding-harness/probe.ts
+++ b/ci/onboarding-harness/probe.ts
@@ -145,18 +145,48 @@ export async function waitForApi(driver: IncusDriver, name: string): Promise<voi
   if (poll.code !== 0) throw new Error(`Shepherd did not come up in ${name}`);
 }
 
-/** Assert the `shepherd` systemd USER unit is `active` (the lifecycle scenario's
- *  proof that the service path — not a hand-rolled boot — owns the process).
- *  `XDG_RUNTIME_DIR=/run/user/0` is required for `systemctl --user` to reach the
- *  user bus inside an `incus exec` session (no login session sets it). */
-export async function assertUnitActive(driver: IncusDriver, name: string): Promise<void> {
+/** Best-effort `systemctl status` tail for a unit, formatted for appending to an assertion
+ *  failure. Returns "" on any capture problem so it can never mask the original failure. */
+async function unitStatusDetail(driver: IncusDriver, name: string, unit: string): Promise<string> {
+  try {
+    const s = await driver.exec(name, [
+      "sh",
+      "-c",
+      `XDG_RUNTIME_DIR=/run/user/0 systemctl --user status ${unit} --no-pager -l -n 30 2>&1 || true`,
+    ]);
+    const out = s.stdout.trim();
+    return out ? `\n--- systemctl status ${unit} ---\n${out}` : "";
+  } catch {
+    return "";
+  }
+}
+
+/** Assert a systemd USER unit is `active` (the lifecycle scenario's proof that the service
+ *  path — not a hand-rolled boot — owns the process). `XDG_RUNTIME_DIR=/run/user/0` is
+ *  required for `systemctl --user` to reach the user bus inside an `incus exec` session
+ *  (no login session sets it).
+ *
+ *  Asserting `herdr` too is load-bearing, not decorative: a `herdr: ok` diagnostic only
+ *  proves SOME daemon answers the socket. If provision left an unsupervised daemon there,
+ *  the unit's ExecStart exits 1 ("already running") and `Restart=always` thrashes it into
+ *  `failed` — while the check stays green and the host silently has no supervised daemon.
+ *  Only the unit's own state can catch that. #1574 */
+export async function assertUnitActive(
+  driver: IncusDriver,
+  name: string,
+  unit = "shepherd",
+): Promise<void> {
   const r = await driver.exec(name, [
     "sh",
     "-c",
-    "XDG_RUNTIME_DIR=/run/user/0 systemctl --user is-active shepherd",
+    `XDG_RUNTIME_DIR=/run/user/0 systemctl --user is-active ${unit}`,
   ]);
   if (r.stdout.trim() !== "active") {
-    throw new Error(`shepherd user unit not active in ${name}: ${r.stdout || r.stderr}`);
+    // A bare "not active: failed" is undiagnosable from CI logs. Attach the unit's status +
+    // recent journal so the FIRST red run says why (ExecStart missing? crash-looped? bound
+    // socket?). Best-effort: a capture failure must never mask the assertion failure.
+    const detail = await unitStatusDetail(driver, name, unit);
+    throw new Error(`${unit} user unit not active in ${name}: ${r.stdout || r.stderr}${detail}`);
   }
 }
 
--- a/ci/onboarding-harness/run.ts
+++ b/ci/onboarding-harness/run.ts
@@ -159,7 +159,11 @@ async function runInstallLifecycleE2E(
     );
   }
 
-  // The unit owns the process — assert it's active, then wait for it to serve.
+  // The units own their processes — assert both are active, then wait for Shepherd to serve.
+  // herdr is checked explicitly: a green `herdr` diagnostic only proves some daemon answers,
+  // not that the SUPERVISED one does. An unsupervised daemon on the socket makes herdr.service
+  // thrash into `failed` behind a passing check — invisible without this assertion. #1574
+  await assertUnitActive(driver, scenario.id, "herdr");
   await assertUnitActive(driver, scenario.id);
   await waitForApi(driver, scenario.id);
 
--- a/ci/onboarding-harness/run.ts
+++ b/ci/onboarding-harness/run.ts
@@ -40,6 +40,25 @@ type ScenarioBase = {
   gateEligible: boolean;
 };
 
+/** Shared tail of BOTH install-e2e runners: probe the freshly-installed host, assert the
+ *  target-ok set, and shape the result. `expect` is the target state (not a seeded defect),
+ *  so `reachedGreen` is just whether detection saw every check reach `ok`. */
+async function finishInstallE2E(
+  driver: IncusDriver,
+  scenario: Scenario,
+  base: ScenarioBase,
+): Promise<ScenarioResult> {
+  const after = await probeDiagnostics(driver, scenario.id);
+  const detection = assertDetection(after, scenario.id, scenario.expect);
+  return {
+    ...base,
+    detection,
+    appliedVia: "verbatim",
+    reachedGreen: detection.detected,
+    installE2E: true,
+  };
+}
+
 /** Inverse flow: bare host → real deploy/install.sh → assert it reaches green.
  *  No seeded defect, no baseline; `expect` is the target-ok set. The install is a
  *  deterministic, LLM-free apply, so it reuses the "verbatim" appliedVia. Throws
@@ -64,15 +83,7 @@ async function runInstallE2E(
     throw new Error(`install.sh failed in ${scenario.id}:\n${install.stderr || install.stdout}`);
   }
   await bootShepherd(driver, scenario.id);
-  const after = await probeDiagnostics(driver, scenario.id);
-  const detection = assertDetection(after, scenario.id, scenario.expect);
-  return {
-    ...base,
-    detection,
-    appliedVia: "verbatim",
-    reachedGreen: detection.detected,
-    installE2E: true,
-  };
+  return await finishInstallE2E(driver, scenario, base);
 }
 
 /** Turn the freshly-extracted /opt/shepherd into a real git checkout BEFORE
@@ -166,16 +177,7 @@ async function runInstallLifecycleE2E(
   await assertUnitActive(driver, scenario.id, "herdr");
   await assertUnitActive(driver, scenario.id);
   await waitForApi(driver, scenario.id);
-
-  const after = await probeDiagnostics(driver, scenario.id);
-  const detection = assertDetection(after, scenario.id, scenario.expect);
-  return {
-    ...base,
-    detection,
-    appliedVia: "verbatim",
-    reachedGreen: detection.detected,
-    installE2E: true,
-  };
+  return await finishInstallE2E(driver, scenario, base);
 }
 
 /** Fail-fast PREFLIGHT flow (`herdr-missing`). Since #1313 a missing herdr prints
--- a/deploy/herdr.service
+++ b/deploy/herdr.service
@@ -4,9 +4,22 @@ Description=herdr — terminal workspace manager (Shepherd's session backend)
 # auto-spawn it on a CLI call (#1574). A user unit — NOT system — because herdr's socket,
 # config and panes all live under $HOME and belong to the operator, same as shepherd.service.
 Documentation=https://herdr.dev
+# Restart=always (below) is only honest with the start-rate limit DISABLED. systemd's default
+# (StartLimitBurst=5 within StartLimitIntervalSec=10s) parks a unit permanently in `failed`
+# after five quick restarts — precisely the crashed/OOM-killed case the restart policy exists
+# to recover. 0 disables the limit; RestartSec=2 is what bounds the retry rate.
+StartLimitIntervalSec=0
 
 [Service]
 Type=simple
+# TEMPLATED, not literal: deploy/provision.ts and deploy/update.sh rewrite this line to the
+# herdr path resolved via `command -v herdr` before installing the unit. herdr is not always in
+# ~/.local/bin — its installer puts it there, but a package install can land it in
+# /usr/local/bin (the onboarding harness seeds exactly that). Both the presence check
+# (provision's `probeVersion`) and the liveness probe (`config.herdrBin`) resolve `herdr` on
+# $PATH, so a hardcoded path could disagree with them and make `enable --now herdr` fail on
+# such a host. The value below is the common default, used verbatim only if templating is
+# somehow skipped.
 ExecStart=%h/.local/bin/herdr server
 # Restart=always (not on-failure): a `herdr server stop`, an OOM kill, or a crashed daemon
 # must all come back — panes OUTLIVE the server and reattach, so a restart is always safe
--- a/deploy/herdr.service
+++ b/deploy/herdr.service
@@ -1,0 +1,22 @@
+[Unit]
+Description=herdr — terminal workspace manager (Shepherd's session backend)
+# Shepherd cannot spawn a single agent without this daemon, and herdr 0.7.3 does NOT
+# auto-spawn it on a CLI call (#1574). A user unit — NOT system — because herdr's socket,
+# config and panes all live under $HOME and belong to the operator, same as shepherd.service.
+Documentation=https://herdr.dev
+
+[Service]
+Type=simple
+ExecStart=%h/.local/bin/herdr server
+# Restart=always (not on-failure): a `herdr server stop`, an OOM kill, or a crashed daemon
+# must all come back — panes OUTLIVE the server and reattach, so a restart is always safe
+# and always preferable to a host that silently can't start agents. This is the "durable
+# operator-side fix" src/herdr-update.ts refers to; with it, that module's last-resort
+# `setsid herdr server` fallback becomes dead code on a provisioned host (it stays for
+# hand-rolled installs).
+Restart=always
+RestartSec=2
+Environment=PATH=%h/.local/bin:%h/.bun/bin:/usr/local/bin:/usr/bin:/bin
+
+[Install]
+WantedBy=default.target
--- a/deploy/herdr.service
+++ b/deploy/herdr.service
@@ -12,14 +12,20 @@ StartLimitIntervalSec=0
 
 [Service]
 Type=simple
-# TEMPLATED, not literal: deploy/provision.ts and deploy/update.sh rewrite this line to the
-# herdr path resolved via `command -v herdr` before installing the unit. herdr is not always in
-# ~/.local/bin — its installer puts it there, but a package install can land it in
-# /usr/local/bin (the onboarding harness seeds exactly that). Both the presence check
-# (provision's `probeVersion`) and the liveness probe (`config.herdrBin`) resolve `herdr` on
-# $PATH, so a hardcoded path could disagree with them and make `enable --now herdr` fail on
-# such a host. The value below is the common default, used verbatim only if templating is
-# somehow skipped.
+# Same env file shepherd.service reads. Without it this daemon cannot see HERDR_BIN or
+# HERDR_SESSION — the documented knobs (README) that Shepherd itself honors via
+# `config.herdrBin` / `config.herdrSocketPath`. On a `HERDR_SESSION=<name>` host an
+# env-blind daemon serves the `default` session while Shepherd's liveness probe targets the
+# per-session socket: the unit is `active`, `herdr` reads `offline`, and no agent can spawn.
+EnvironmentFile=-%h/.shepherd/env
+# TEMPLATED, not literal: deploy/provision.ts and deploy/update.sh rewrite this line before
+# installing the unit, resolving `${HERDR_BIN:-herdr}` (env file first, then PATH). herdr is not
+# always in ~/.local/bin — its installer puts it there, a package install can land it in
+# /usr/local/bin (the onboarding harness seeds exactly that), and HERDR_BIN can point anywhere.
+# `config.herdrBin` is `HERDR_BIN ?? "herdr"`, so it only falls back to a PATH lookup when
+# HERDR_BIN is unset; the unit must exec whichever binary that resolves to, or it supervises a
+# different herdr than the one diagnostics probes. The value below is the common default, used
+# verbatim only if templating is somehow skipped.
 ExecStart=%h/.local/bin/herdr server
 # Restart=always (not on-failure): a `herdr server stop`, an OOM kill, or a crashed daemon
 # must all come back — panes OUTLIVE the server and reattach, so a restart is always safe
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -249,12 +249,17 @@ function probeVersion(bin: string): string | null {
 }
 
 /** Resolve a binary to its absolute path via `command -v`; null if not on PATH. Used to
- *  template the herdr unit's ExecStart (systemd needs an absolute executable). */
-function probeBinPath(bin: string): string | null {
+ *  template the herdr unit's ExecStart (systemd needs an absolute executable).
+ *
+ *  MUST be given the AUGMENTED env (`buildEnv`), not provision's inherited one: install.sh drops
+ *  herdr in ~/.local/bin, which is absent from the PATH provision itself was launched with, so
+ *  an inherited-PATH lookup finds nothing on the very host that just installed it. */
+function probeBinPath(bin: string, env: NodeJS.ProcessEnv): string | null {
   try {
     const out = execFileSync("sh", ["-c", `command -v ${bin}`], {
       encoding: "utf8",
       stdio: ["ignore", "pipe", "ignore"],
+      env,
     });
     const p = out.trim();
     return p === "" ? null : p;
@@ -281,7 +286,7 @@ interface ProvisionOpts {
   /** Injectable total-memory probe; defaults to os.totalmem. */
   totalMem?: () => number;
   /** Injectable absolute-path resolver (herdr.service ExecStart templating). */
-  probePath?: (bin: string) => string | null;
+  probePath?: (bin: string, env: NodeJS.ProcessEnv) => string | null;
 }
 
 /** Step 1 — table-driven prereq install loop (idempotent; skips adequate tools). */
@@ -330,8 +335,8 @@ export function installService(
   env: NodeJS.ProcessEnv,
   home: string,
   buildEnv: NodeJS.ProcessEnv,
-  /** Resolves a binary to an absolute path (injected in tests). */
-  probePath: (bin: string) => string | null = probeBinPath,
+  /** Resolves a binary to an absolute path against a given env (injected in tests). */
+  probePath: (bin: string, env: NodeJS.ProcessEnv) => string | null = probeBinPath,
 ): void {
   log("installing systemd user unit");
   const unitDir = join(home, ".config", "systemd", "user");
@@ -374,7 +379,8 @@ export function installService(
   // auto-spawn it (#1574). TEMPLATED, not verbatim: ExecStart must be the herdr this host
   // actually resolves on PATH (installers use ~/.local/bin, packages /usr/local/bin), else the
   // unit points at nothing while `probeVersion`/`config.herdrBin` are perfectly happy.
-  const herdrPath = probePath("herdr");
+  // buildEnv, NOT env: it carries ~/.local/bin (where install.sh just put herdr) on PATH.
+  const herdrPath = probePath("herdr", buildEnv);
   if (!herdrPath) throw new Error("cannot install herdr.service: herdr is not on PATH");
   fileIO.write(
     join(unitDir, "herdr.service"),
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -125,10 +125,16 @@ export function templateHerdrUnit(unit: string, herdrPath: string): string {
  *  would thrash. Sources `~/.shepherd/env` and honors `HERDR_BIN` so it stops the SAME binary
  *  the unit and the diagnostics probe use — a bare `herdr` would miss a `HERDR_BIN` daemon
  *  entirely. Best-effort by construction: every branch swallows failure and exits 0, so a host
- *  with no herdr running (the fresh-install case) sails straight through. #1574 */
+ *  with no herdr running (the fresh-install case) sails straight through. #1574
+ *
+ *  `set -a` around the source is load-bearing, not decoration: a bare `.` leaves HERDR_SESSION /
+ *  HERDR_SOCKET_PATH as UNEXPORTED shell variables, so the `server stop` CHILD never sees them
+ *  and addresses the default socket. On a non-default-session host that stops nothing, the real
+ *  orphan stays bound, and `enable --now herdr` then thrashes the unit forever behind a green
+ *  check. (HERDR_BIN only appeared to work because it is expanded in-shell, below.) */
 const HERDR_ADOPT_SOCKET =
   'export PATH="$HOME/.local/bin:$PATH"; ' +
-  '[ -f "$HOME/.shepherd/env" ] && . "$HOME/.shepherd/env"; ' +
+  'set -a; [ -f "$HOME/.shepherd/env" ] && . "$HOME/.shepherd/env"; set +a; ' +
   'H="${HERDR_BIN:-herdr}"; ' +
   "systemctl --user is-active --quiet herdr && exit 0; " +
   'command -v "$H" >/dev/null 2>&1 && "$H" server stop >/dev/null 2>&1; ' +
@@ -279,6 +285,16 @@ function probeBinPath(_bin: string, env: NodeJS.ProcessEnv): string | null {
   }
 }
 
+/** Read a file that may not exist yet; null when absent/unreadable. Lets `installService` ask
+ *  "did this unit actually change?" without a stat seam — a missing unit reads as changed. */
+function readIfPresent(fileIO: FileIO, path: string): string | null {
+  try {
+    return fileIO.read(path);
+  } catch {
+    return null;
+  }
+}
+
 function log(msg: string): void {
   process.stdout.write(`\x1b[36m▸ ${msg}\x1b[0m\n`);
 }
@@ -393,16 +409,23 @@ export function installService(
   // buildEnv, NOT env: it carries ~/.local/bin (where install.sh just put herdr) on PATH.
   const herdrPath = probePath("herdr", buildEnv);
   if (!herdrPath) throw new Error("cannot install herdr.service: herdr is not on PATH");
-  fileIO.write(
-    join(unitDir, "herdr.service"),
-    templateHerdrUnit(fileIO.read(join(repo, "deploy", "herdr.service")), herdrPath),
+  const herdrUnitPath = join(unitDir, "herdr.service");
+  const desiredHerdrUnit = templateHerdrUnit(
+    fileIO.read(join(repo, "deploy", "herdr.service")),
+    herdrPath,
   );
+  const herdrUnitChanged = readIfPresent(fileIO, herdrUnitPath) !== desiredHerdrUnit;
+  if (herdrUnitChanged) fileIO.write(herdrUnitPath, desiredHerdrUnit);
   run("systemctl", ["--user", "daemon-reload"]);
   // Pick up a CHANGED ExecStart on a host whose herdr unit is already running: `enable --now`
   // will not restart an active unit, so a re-provision after herdr moved (e.g. /usr/local/bin →
   // ~/.local/bin, or HERDR_BIN changed) would otherwise leave the old daemon running the stale
   // path forever. `try-restart` restarts it iff active, and is a no-op when it is not.
-  run("systemctl", ["--user", "try-restart", "herdr"]);
+  //
+  // GATED on the unit actually differing: try-restart kills the daemon backing every live agent
+  // session, so bouncing it for a byte-identical unit would make each re-provision needlessly
+  // disruptive. A missing installed unit reads as "changed" (the first-install case).
+  if (herdrUnitChanged) run("systemctl", ["--user", "try-restart", "herdr"]);
   const user = env.USER;
   if (!user) throw new Error("cannot enable-linger: $USER is not set");
   run("loginctl", ["enable-linger", user]);
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -122,12 +122,16 @@ export function templateHerdrUnit(unit: string, herdrPath: string): string {
 /** Adopt the socket before enabling the unit: if the unit is NOT already the active herdr, stop
  *  whatever daemon is (an in-app Fix's detached `HERDR_SERVE` child, or a hand-rolled one). A
  *  second `herdr server` against a bound socket exits 1, so without this `enable --now herdr`
- *  would thrash. Best-effort by construction — every branch swallows failure and exits 0, so a
- *  host with no herdr running (the fresh-install case) sails straight through. #1574 */
+ *  would thrash. Sources `~/.shepherd/env` and honors `HERDR_BIN` so it stops the SAME binary
+ *  the unit and the diagnostics probe use — a bare `herdr` would miss a `HERDR_BIN` daemon
+ *  entirely. Best-effort by construction: every branch swallows failure and exits 0, so a host
+ *  with no herdr running (the fresh-install case) sails straight through. #1574 */
 const HERDR_ADOPT_SOCKET =
   'export PATH="$HOME/.local/bin:$PATH"; ' +
+  '[ -f "$HOME/.shepherd/env" ] && . "$HOME/.shepherd/env"; ' +
+  'H="${HERDR_BIN:-herdr}"; ' +
   "systemctl --user is-active --quiet herdr && exit 0; " +
-  "command -v herdr >/dev/null 2>&1 && herdr server stop >/dev/null 2>&1; " +
+  'command -v "$H" >/dev/null 2>&1 && "$H" server stop >/dev/null 2>&1; ' +
   "systemctl --user reset-failed herdr >/dev/null 2>&1; exit 0";
 
 /** Service path iff linux AND SHEPHERD_NO_SERVICE unset/empty. macOS always takes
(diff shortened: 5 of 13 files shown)


scoop:
## What I verified

**i18n parity — clean.** The PR adds no new message keys. `diagnostics_hint_herdr_offline` already exists in both catalogs: `ui/messages/en.json:1269` and `ui/messages/de.json:1269`. It also has a docs entry at `ui/src/lib/diagnostics-docs.ts:7`. Verified by grep across `ui/messages/`.

**Identifier resolution.** `HERDR_SERVE` and `HERDR_INSTALL` are newly exported from `src/remediations.ts:21,40` and imported at `deploy/provision.ts:23` and `test/remediations.test.ts:6` — both exist and spell/case-match. `diagnostics_hint_herdr_offline` is a real emitted hintKey (`src/diagnostics.ts:324`), so the new `REMEDIATIONS` entry (`src/remediations.ts:83`) is reachable through `autoFixCommandFor` → `DiagnosticsService.fix` (`src/diagnostics.ts:568-570`) and through the check annotation loop (`src/diagnostics.ts:539-540`). Neither key is in `GUIDANCE_ONLY` (`src/remediations.ts:103-106`), so the in-app Fix button will render.

**Shell correctness of `HERDR_SERVE`.** All constructs are POSIX (`export`, `||`, `{ …; }`, `if`/`command -v`, `&`, `for … do … done`, `exit`), which matters because both consumers run it under `sh`, not `bash`: `ci/onboarding-harness/apply.ts:75` (`driver.exec(name, ["sh", "-c", cmd])`) and `src/diagnostics.ts:143` (`spawn("sh", ["-c", cmd], …)`). `&` is a valid list terminator before `else`/`fi`, so the `if … then … & else … & fi` shape parses under dash/ash. Each remediation is executed in its own `sh -c`, so the embedded `exit 0`/`exit 1` cannot truncate a sibling command (`ci/onboarding-harness/apply.ts:74-77` loops one `exec` per command). The ~10s poll fits comfortably inside `REMEDIATION_TIMEOUT_MS = 120_000` (`src/config.ts:98`). No bug found here.

**`herdr-update.ts` is comment-only.** I diffed the returned script array: `src/herdr-update.ts:113-121` is byte-identical apart from the inline comment. The corrected claim (a herdr CLI call does *not* auto-spawn the daemon) is consistent with the new `deploy/herdr.service` and with `src/remediations.ts:23-25`.

---

## Blocking: the two start paths collide on a fresh service host

This is the path the PR exists to fix (`install-e2e-service`), and I traced it end to end.

`provision()` calls `installPrereqs` **before** `installService` (`deploy/provision.ts:381` then `:386`). On a bare host, herdr is absent, so `selectPrereqCommand` resolves `autoFixCommandFor("diagnostics_hint_herdr_missing")` (`deploy/provision.ts:44`, `:93`) — which this PR changed to `${HERDR_INSTALL} && (${HERDR_SERVE})` (`src/remediations.ts:81`) — and runs it via `run("bash", ["-c", cmd])` (`deploy/provision.ts:248`). `HERDR_SERVE` ends with `setsid herdr server … &` (`src/remediations.ts:44`), so provision now leaves a **detached, session-leader herdr server bound to the socket**, unsupervised, before `installService` is ever entered.

`installService` then runs `systemctl --user enable --now herdr` (`deploy/provision.ts:329`), whose `ExecStart=%h/.local/bin/herdr server` (`deploy/herdr.service:10`) starts a *second* server against that already-bound socket. I cannot verify herdr's double-start semantics from this repo (the binary is external — stated limitation), but **both possible behaviors are bad**, and `Restart=always` (`deploy/herdr.service:17`) makes them worse:

- If the second `herdr server` exits non-zero (socket busy), `Restart=always` re-execs it every 2s.
- If it exits *zero* ("already running"), `Restart=always` **still** re-execs it — `always` restarts on clean exit too.

Either way the unit thrashes, hits systemd's default start limit (5 starts / 10s; `RestartSec=2` reaches it exactly), and lands in `failed`. The daemon actually serving is the unsupervised orphan from `installPrereqs`. That is precisely the state this PR set out to eliminate, and the PR's own docstring names it: "IDEMPOTENT: a live daemon short-circuits, so this never races a second server against a bound socket" (`src/remediations.ts:29-30`). `installService` bypasses that guard entirely — the sibling path has no `agent list ||` short-circuit.

Worse, it is invisible to the harness: `install-e2e-service` asserts only `{ id: "herdr", state: "ok" }` (`ci/onboarding-harness/scenarios.ts:163`), which the orphan satisfies. The scenario goes green while `herdr.service` is `failed`. And the new comment at `src/herdr-update.ts:98-99` — "on a host provisioned by deploy/provision.ts this is normally unreachable: `deploy/herdr.service` (Restart=always) wins the retry race" — is false in that state.

The new test (`test/provision.test.ts:387-398`) pins the *ordering* of `enable --now herdr` vs `update.sh` but never observes that a daemon is already up.

## Blocking: `ExecStart` hardcodes a path the presence check does not require

`deploy/herdr.service:10` is `ExecStart=%h/.local/bin/herdr server`, but nothing in the codebase requires herdr to live there. `probeVersion` execs the bare name on `$PATH` (`deploy/provision.ts:203`), and diagnostics resolves `config.herdrBin` = `process.env.HERDR_BIN ?? "herdr"` (`src/config.ts:455`, consumed at `src/diagnostics.ts:247`). The harness's own seed confirms `/usr/local/bin/herdr` is a real install location: `seed: ["rm -f /usr/local/bin/herdr ~/.local/bin/herdr"]` (`ci/onboarding-harness/scenarios.ts:88`).

So on a host where herdr is already installed outside `~/.local/bin`, `needsInstall` returns false (`deploy/provision.ts:79-83`), the prereq loop skips the install, and `systemctl --user enable --now herdr` fails to start a nonexistent `ExecStart`. `defaultRunner` throws on non-zero status (`deploy/provision.ts:183-185`), so **`provision` now hard-aborts on a host that provisioned fine before this PR**. Either template the resolved binary path into the unit the way `templateUnit` handles `WorkingDirectory` (`deploy/provision.ts:113-115`), or rely on systemd ≥239 `$PATH` resolution together with the unit's own `Environment=PATH` (`deploy/herdr.service:19`).

## Non-blocking

- **Start limit.** `deploy/herdr.service:17-18` sets `Restart=always` + `RestartSec=2` with no `StartLimitIntervalSec=0`. The unit comment promises "a `herdr server stop`, an OOM kill, or a crashed daemon must all come back" (`deploy/herdr.service:11-12`), but a fast-failing herdr exhausts the default 5-starts-in-10s burst and systemd leaves the unit `failed` permanently. The promise in the comment is not what the unit implements.
- **Existing hosts never get the unit.** `deploy/update.sh:49-54` exists specifically because "update.sh historically only restarts; provision.ts installs units only on a fresh box. So an existing live host would never pick up the hourly-backup timer." It re-syncs `shepherd-backup.*` (`deploy/update.sh:59-61`) and `shepherd-logrotate.*` (`deploy/update.sh:86-87`) for exactly that reason. `herdr.service` is not added to that block, so no already-provisioned host ever receives the daemon unit this PR introduces.
- **`HERDR_BIN` is ignored.** `HERDR_SERVE` invokes the bare `herdr` (`src/remediations.ts:42,44,45,47`), but the check it must clear spawns `config.herdrBin` (`src/diagnostics.ts:247` → `src/config.ts:455`). With `HERDR_BIN` set, the in-app Fix for `herdr_offline` starts the wrong (or no) binary, `HERDR_SERVE`'s poll times out after 10s, and `fix()` rejects (`src/diagnostics.ts:570`).
- **`herdr_outdated` sibling gap.** This PR's premise is that installing the binary without (re)starting the daemon leaves the check red — that is why `herdr_missing` became `install && serve` (`src/remediations.ts:81`). Its sibling `diagnostics_hint_herdr_outdated: HERDR_INSTALL` (`src/remediations.ts:82`) still reinstalls the binary while the old daemon keeps running, so the warning cannot clear. The PR's own plan flags this (`docs/superpowers/plans/2026-07-09-herdr-offline-provisioning.md:481`) and leaves it open.
- **Comment misattribution.** `src/remediations.ts:79-80` says the subshell is needed so `&&` doesn't gate "just `HERDR_INSTALL`'s leading `export`". `HERDR_INSTALL` (`src/remediations.ts:21`) is `curl -fsSL https://herdr.dev/install.sh | bash` — it has no `export`. The leading `export` belongs to `HERDR_SERVE` (`src/remediations.ts:41`). The reasoning is right; the attribution is inverted.

deploy/herdr.service: with `Restart=always`, a `herdr update --handoff` (the script built at `src/herdr-update.ts:110-112`) causes the unit's main process to exit, so systemd re-execs `ExecStart` and races the handed-off server for the socket — the same collision as the blocking finding, triggered on every herdr update once the unit ships.
src/remediations.ts: on the `nohup` branch (`src/remediations.ts:45`) the spawned daemon stays in the caller's process group, so `defaultRunRemediation`'s timeout path `process.kill(-child.pid, "SIGKILL")` (`src/diagnostics.ts:154`) would kill the daemon it just started; the `setsid` branch escapes the group. Dormant because the poll normally returns well inside the 120s budget.

deploy/update.sh never re-copies `shepherd.service`, so the new `After=`/`Wants=herdr.service` ordering added to that unit will not reach any already-provisioned host either.

## Stated limitations
Could not verify herdr's behavior when `herdr server` is invoked against an already-bound socket, nor its handoff semantics under a systemd unit — the binary is external and not present in this repo. The blocking finding above does not depend on which behavior is real: `Restart=always` restarts on clean exit as well as on failure.

scoop:
Six of the seven points are fixed and pushed (`e23cd107`, `300d5ad3`, `9bd0af5b`, `fcd1b6d4`, `0f20a41b`). **Point 6 I'm pushing back on** — the rest of this comment is why, plus two things the review surfaced that were bigger than reported.

## Point 6 — `herdr_outdated`: the warning *does* clear

The claim is that reinstalling the binary while the old daemon runs means "the outdated warning cannot clear." It does clear, because `versionProbe` reads the **binary**, not the daemon:

```
src/diagnostics.ts:  const out = await this.runVersion(bin, ["--version"]);
```

`herdr --version` needs no server. Verified in a clean `images:ubuntu/24.04` instance with no daemon ever started:

```
$ herdr status server
status: not running
$ herdr --version
herdr 0.7.3
exit=0
```

So after `HERDR_INSTALL`, the next probe parses the **new** version, the liveness step passes (the old daemon still answers `agent list`), and the check goes `ok`. This is structurally unlike `herdr_missing`, where the daemon genuinely does not exist and the check stays red forever without a start — which is why that one had to be chained and this one doesn't.

The real defect underneath the finding is different and I agree it exists: the *running* daemon keeps serving the old binary until something restarts it. That's a daemon-adoption problem, and the correct fix is `herdr update --handoff` (which preserves panes — they outlive the server), not an install-and-start chain. Bolting `HERDR_SERVE` onto `herdr_outdated` would be a no-op anyway: it short-circuits on the live daemon and never restarts anything. Tracked in #1578.

## Point 1 was worse than described — and now has a test

Confirmed the mechanism in a container: a second `herdr server` against a bound socket exits **1**.

```
$ herdr server
error: herdr server is already running
$ echo $?
1
```

Then, on the *pre-fix* tree, `install-e2e-service` produced exactly the predicted state:

```
× herdr.service — Active: failed (Result: exit-code)
  Main PID: 2961 (code=exited, status=1/FAILURE)
  herdr.service: Start request repeated too quickly.
```

Note that the scenario **passed** in that state before this PR, because the `herdr` diagnostic only proves *some* daemon answers the socket — the orphan did. So `9bd0af5b` makes the harness assert `systemctl --user is-active herdr` and attaches `systemctl status` to the failure. That assertion is what caught it, and it fails closed on the old code.

Fix per your first suggestion: `Prereq.installCommand` selects bare `HERDR_INSTALL` for the prereq loop, so the unit owns the daemon. I also added `HERDR_ADOPT_SOCKET` before `enable --now` in both `provision.ts` and `update.sh`, because your fix alone doesn't cover an **upgrade** on a host where an in-app Fix already left a detached daemon.

## Point 2 had a second instance one layer up

Templating `ExecStart` from `command -v herdr` was right, and `installService` now throws rather than install a unit aimed at nothing. That guard immediately caught a bug in my own fix: `probeBinPath` was resolving against provision's *inherited* `PATH`, which lacks `~/.local/bin` — where `install.sh` had just put herdr. Real container output:

```
✗ provision failed: cannot install herdr.service: herdr is not on PATH
```

Fixed in `fcd1b6d4` by resolving against `buildEnv`. Had I written the lenient fallback instead of failing closed, that would have shipped as a unit that works by accident here and breaks wherever herdr lives elsewhere.

## Verification

All three gate-eligible scenarios green on the fixed tree, with `herdr.service` now asserted `active` (not merely "some daemon answers"):

| Scenario | Result |
| --- | --- |
| `herdr-missing` (archlinux) | PASS |
| `install-e2e` (ubuntu, no systemd) | PASS |
| `install-e2e-service` (ubuntu, systemd) | PASS |

Suite 6326 pass / 0 fail, `tsc` clean, `fallow audit` clean, `systemd-analyze verify` clean.

Unrelated and pre-existing: `onboarding-gate.sh` is red on `node-too-old` **on `main` too** — the upstream codex installer is broken and `applyVerbatim` bails on the first non-zero exit while the `optional` codex check sorts before `node`. Filed as #1577.

scoop:
## Re-review of the prior round

I checked each of the seven prior points against the actual diff, not against the author note.

1. **Prereq loop no longer starts the daemon — FIXED.** `PREREQS`' herdr entry now carries `installCommand: HERDR_INSTALL` (deploy/provision.ts:58) and `selectPrereqCommand` prefers it over `autoFixCommandFor` (deploy/provision.ts:110), so the chained `HERDR_SERVE` in `diagnostics_hint_herdr_missing` (src/remediations.ts:89) never runs on provision's path. `HERDR_ADOPT_SOCKET` (deploy/provision.ts:127-131) additionally runs before `systemctl --user enable --now herdr` (deploy/provision.ts:397-401). Accepted.
2. **Hardcoded `ExecStart` path — FIXED for the `$PATH` case.** `probeBinPath` resolves `command -v herdr` against `buildEnv` (deploy/provision.ts:257-269, called at :383), `templateHerdrUnit` rewrites the line (deploy/provision.ts:118-120), and a null resolution now throws (deploy/provision.ts:384) rather than installing a unit aimed at nothing. Accepted as far as `$PATH` goes — see finding 1 for the `HERDR_BIN` half, which is a *new* instance introduced by the fix, not a re-raise.
3. **Start-rate limit — FIXED.** `StartLimitIntervalSec=0` sits in `[Unit]` (deploy/herdr.service:11), which is the correct section for it (`[Service]` begins at :13). Accepted.
4. **`update.sh` never synced the unit — FIXED.** deploy/update.sh:91-119 templates, reloads, adopts the socket and `enable --now herdr`, inside the same `systemctl --user show-environment` gate as the backup/logrotate blocks (deploy/update.sh:55). Accepted.
5. **`HERDR_SERVE` ignored `HERDR_BIN` — FIXED.** `H="${HERDR_BIN:-herdr}"` (src/remediations.ts:47) now matches `config.herdrBin = process.env.HERDR_BIN ?? "herdr"` (src/config.ts:455), which is what `defaultHerdrLiveness` spawns (src/diagnostics.ts:182). Accepted.
6. **`herdr_outdated` mapping — ACCEPTED, not re-raised.** I verified the author's claim against `versionProbe` (src/diagnostics.ts:298-312): it reads `runVersion(bin, ["--version"])` first, runs `liveness` second, and only *then* compares against the floor. So `diagnostics_hint_herdr_outdated` is emitted only when the daemon already answers; reinstalling the binary alone does make the next probe parse the new version and return `ok`. This is structurally unlike `herdr_missing`. The mapping at src/remediations.ts:95 is correct as written.
7. **Comment attribution — FIXED.** src/remediations.ts:84-88 now attributes the leading `export` to `HERDR_SERVE`, and `HERDR_SERVE`'s first statement is indeed `export PATH="$HOME/.local/bin:$PATH";` (src/remediations.ts:47). Accepted.

The task from the originating issue is met on its own terms: `herdr-missing` gets an install-and-start remediation (src/remediations.ts:89), and both `install-e2e` paths now leave a live daemon — the unit on the systemd path (deploy/provision.ts:401) and an explicit `HERDR_SERVE` on the `SHEPHERD_NO_SERVICE` path (deploy/provision.ts:426). The harness now asserts `herdr.service` is `active` rather than merely that some daemon answers (ci/onboarding-harness/run.ts:177, ci/onboarding-harness/probe.ts:174-186), which is the right assertion.

Two new defects came in with the fixes, plus two nits.

## Finding 1 — the daemon unit resolves herdr differently from the client that talks to it

`deploy/shepherd.service:27` loads `EnvironmentFile=-%h/.shepherd/env`. That file is the documented home for three herdr knobs: `HERDR_BIN` and `HERDR_SESSION` (README.md:259-260) and `HERDR_SOCKET_PATH` (docs-site/src/content/docs/reference/configuration.md:24-26). Shepherd reads all three: `config.herdrBin = process.env.HERDR_BIN ?? "herdr"` (src/config.ts:455), `herdrSession = process.env.HERDR_SESSION ?? "default"` (src/config.ts:39), `herdrSocketPath` derived from it (src/config.ts:44-47).

The new `deploy/herdr.service` has **no** `EnvironmentFile=` at all — only `Environment=PATH=…` (deploy/herdr.service:32). systemd does not inherit an ambient environment for a unit, so its `herdr server` can never see `HERDR_BIN` or `HERDR_SESSION`. And the templating that produces its `ExecStart` resolves `command -v herdr` (deploy/provision.ts:259, deploy/update.sh:106), i.e. `$PATH`, not `HERDR_BIN`.

This is verifiably inconsistent with the PR's own reasoning in two places:

- deploy/herdr.service:18-20 states "Both the presence check (provision's `probeVersion`) and the liveness probe (`config.herdrBin`) resolve `herdr` on `$PATH`". The second half is false: `config.herdrBin` is `HERDR_BIN ?? "herdr"` (src/config.ts:455), and it is spawned directly, not through a shell path lookup (src/diagnostics.ts:182).
- src/remediations.ts:29-33 says the exact opposite for `HERDR_SERVE`: "A bare `herdr` here would start a DIFFERENT binary than the one diagnostics probes whenever HERDR_BIN is set." That is the reasoning that fixed prior point 5. The same reasoning was not applied to the unit, to `probeBinPath`, or to `HERDR_ADOPT_SOCKET` (deploy/provision.ts:130) and deploy/update.sh:113, both of which invoke a bare `herdr server stop`.

The `HERDR_SESSION` axis is the sharper one. src/config.ts:41-47's own comment establishes that a non-`default` session gets its own socket at `~/.config/herdr/sessions/<name>/herdr.sock`. Shepherd's liveness probe spawns `herdr agent list` inheriting the unit's environment (src/diagnostics.ts:182 — `spawn` with no `env`, so `process.env`, which on the service path includes `~/.shepherd/env`). The daemon unit, having no `EnvironmentFile`, serves the `default` session. On such a host `installService` brings up a supervised daemon that Shepherd's client never reaches: `herdr` stays `offline` forever and no agent can spawn — while `herdr.service` reports `active` and the harness's new assertion is satisfied. `HERDR_ADOPT_SOCKET` (deploy/provision.ts:127-131) and deploy/update.sh:112-115 also run `herdr server stop` without that environment, so they stop the *default* session's server rather than the one Shepherd uses.

This is not a regression relative to `main` (those hosts had no daemon at all before), but it is an incomplete fix on a documented, supported configuration, and the PR's own plan doc flags it as unresolved: docs/superpowers/plans/2026-07-09-herdr-offline-provisioning.md:483 — "`deploy/herdr.service` hardcodes `%h/.local/bin/herdr`. Honor `HERDR_BIN` (which `config.herdrBin` reads) instead?" Add `EnvironmentFile=-%h/.shepherd/env` to the unit and resolve the templated path from `HERDR_BIN` when set, or explicitly state in the unit and in `config.ts` that the daemon is `default`-session-only.

## Finding 2 — the new `herdr_offline` Fix reintroduces the orphan-vs-unit conflict this PR fixes elsewhere

src/remediations.ts:96 adds `diagnostics_hint_herdr_offline: HERDR_SERVE`. That key is emitted by `versionProbe` (src/diagnostics.ts:302-307, wired at :324), and because it is not in `GUIDANCE_ONLY` (src/remediations.ts:116-119), `autoFixCommandFor` now returns a command for it — so the check is annotated with `remediation` (src/diagnostics.ts:539-540) and `fix()` will run it (src/diagnostics.ts:568-570). This path executes today.

`HERDR_SERVE` starts a detached `setsid herdr server` (src/remediations.ts:49-51) and never touches systemd. On a host provisioned by this PR, `herdr: offline` means `herdr.service` is enabled but its `ExecStart` cannot bind (otherwise `Restart=always` + `RestartSec=2` would have it back up within two seconds — deploy/herdr.service:30-31). Clicking Fix then puts an unsupervised daemon on the socket. From that moment the unit's `ExecStart` exits 1 against a bound socket — the exact failure the author reproduced and quoted — and because this PR sets `StartLimitIntervalSec=0` (deploy/herdr.service:11), systemd no longer parks it in `failed` after five attempts; it retries forever, every two seconds, while the `herdr` check reads `ok` because the orphan answers. That is precisely the "green check, thrashing unit, unsupervised daemon" state the PR's own comment at ci/onboarding-harness/run.ts:172-175 describes as "invisible without this assertion" — reintroduced on the in-app path, where there is no assertion.

The PR's plan doc raises this too and leaves it open: docs/superpowers/plans/2026-07-09-herdr-offline-provisioning.md:482 — "should the Fix endpoint prefer `systemctl --user start herdr` when the unit exists?". It should. Prefer `systemctl --user restart herdr` when `systemctl --user cat herdr` succeeds, and fall back to the detached spawn otherwise (macOS / hand-rolled installs), so the Fix repairs the supervisor rather than racing it.

## Finding 3 — `update.sh` writes a changed unit but never applies it

deploy/update.sh:109-116 templates `ExecStart` into `$UNIT_DIR/herdr.service`, runs `daemon-reload`, and then `systemctl --user enable --now herdr`. `enable --now` does not restart an already-active unit, and the stop at :113 is guarded by `if ! systemctl --user is-active --quiet herdr` (:112). So whenever the block actually changes the unit's content — the case it exists to handle — the running daemon keeps the old `ExecStart`. Concretely: a host provisioned with `ExecStart=/usr/local/bin/herdr server` (the location the harness seeds, per ci/onboarding-harness/scenarios.ts:88) that later gains `~/.local/bin/herdr` will have `command -v herdr` resolve to the new binary, sed it into the file, reload, and then keep supervising the old one indefinitely. `installService` has the same shape: it rewrites the unit at deploy/provision.ts:385-388 but `HERDR_ADOPT_SOCKET` short-circuits on an active unit (deploy/provision.ts:129) and `enable --now` (deploy/provision.ts:401) is then a no-op. A `try-restart` after the reload closes it.

## Finding 4 (nit) — `export PATH` leaks past the block that needs it

deploy/update.sh:105 exports `PATH="$HOME/.local/bin:$PATH"` at function scope, so it stays in effect for `systemctl --user restart "$UNIT"` (:132) and the health check (:143-150). The neighbouring env handling in the same file is deliberately subshell-scoped for exactly this reason — see the comment at deploy/update.sh:63-66 ("Subshell-scoped so the override can't leak into the rest of this deploy script"). Prepending `~/.local/bin` is benign today, but the inconsistency is worth closing: scope it to the `if [[ -n "$HERDR_PATH" ]]` block or use `PATH=… command -v herdr`.

## Verified, no action needed

- `HERDR_INSTALL` and `HERDR_SERVE` are both exported (src/remediations.ts:21, :46) and both imported and used in deploy/provision.ts:23 (`HERDR_INSTALL` at :58, `HERDR_SERVE` at :426). No dangling import.
- The subshell claim at src/remediations.ts:84-88 is right: `|` binds tighter than `&&`, so `diagnostics_hint_herdr_missing` (:89) is `(curl|bash) && (HERDR_SERVE)`, and the `exit 0`/`exit 1` inside the subshell become the compound command's status. `applyVerbatim` runs it under `sh -c` (ci/onboarding-harness/apply.ts:75); the `herdr-missing` scenario is `images:archlinux` (ci/onboarding-harness/scenarios.ts:87) where `/bin/sh` is bash, and the construct is POSIX-clean anyway (`{ …; }`, `&` as a list terminator before `else`/`fi`, `${VAR:-default}`).
- `HERDR_ADOPT_SOCKET` (deploy/provision.ts:127-131) always terminates in `exit 0`, so `defaultRunner`'s throw-on-nonzero (deploy/provision.ts:220-222) cannot abort provision on a host with no herdr running. Confirmed by reading each `;`-separated clause.
- `assertUnitActive`'s new third parameter defaults to `"shepherd"` (ci/onboarding-harness/probe.ts:174-178); the only two call sites are ci/onboarding-harness/run.ts:177-178 and both agree with the signature. `unitStatusDetail` reads `s.stdout` from `driver.exec`, whose return type is `IncusExec` (ci/onboarding-harness/incus.ts:63).
- `installService`'s new `probePath` parameter is threaded from `provision` (deploy/provision.ts:447, :459) and defaults to `probeBinPath`; the signature change is matched at every call site (test/provision.test.ts:348-355, :405-413).
- `StartLimitIntervalSec=` is a `[Unit]` directive on modern systemd, and deploy/herdr.service:11 places it there, before `[Service]` at :13.
- `diagnostics_hint_herdr_offline` exists in both message catalogs (ui/messages/en.json and ui/messages/de.json) — locale parity holds for the key the new remediation entry keys off.
- src/herdr-update.ts changes are comment-only (verified by filtering non-comment `+`/`-` lines out of the diff for that file — the result is empty).

## Could not verify

- herdr's own CLI semantics are not in this repo. I did not verify that `herdr server` exits 1 against a bound socket, that `herdr server stop` is graceful, that `herdr --version` answers with no daemon, or that `herdr server` honors `HERDR_SESSION`. Where my reasoning depends on those, I leaned on in-repo ground truth instead: src/config.ts:41-47 for the session→socket derivation, and src/diagnostics.ts:298-312 for the probe ordering. Findings 1 and 2 stand on the in-repo inconsistency (unit env vs. `shepherd.service`'s `EnvironmentFile`; Fix path vs. unit ownership), not on herdr internals.

- deploy/herdr.service: `Restart=always` + `RestartSec=2` + `StartLimitIntervalSec=0` (:11, :30-31) removes systemd's terminal `failed` state. Any condition under which `ExecStart` can never succeed becomes a silent, unbounded two-second respawn loop with no `failed` unit for an operator or alert to key on. `src/herdr-update.ts`'s `herdr update --handoff` (buildUpdateScript) is the plausible trigger: if the handed-off server lands outside the unit's cgroup, the unit's `ExecStart` will hit a bound socket on every retry, forever. Whether that happens depends on herdr's handoff implementation, which I cannot verify here.
- deploy/provision.ts: `selectPrereqCommand` (:110) now returns `prereq.installCommand` *before* consulting `autoFixCommandFor`, which is the function that filters `GUIDANCE_ONLY` hints (src/remediations.ts:132-135). The docstring on `Prereq.hintKey` (:29-31) still claims "tailscale (guidance-only) is intentionally NOT in this table and can never be auto-installed". Harmless for herdr, but a future prereq that sets `installCommand` silently bypasses that gate.
- src/remediations.ts: `diagnostics_hint_herdr_outdated: HERDR_INSTALL` (:95) does clear the check, as verified above — but it leaves the *running* daemon on the old binary until something restarts it, so the check goes green while the daemon Shepherd talks to is still outdated. The comment at :90-94 documents this and defers it to #1578; noting it only because "the warning clears" and "the host is updated" are not the same statement.

- `ci/onboarding-harness/onboarding-gate.sh` is reported red on `node-too-old` on `main` as well (upstream codex installer broken; `applyVerbatim` bails on the first non-zero exit while the `optional` codex check sorts before `node`) — the file is not in this diff, filed as #1577 per the author note; I did not verify it.
- `deploy/provision.ts`'s `probeVersion` (:238-249) spawns a bare `bin` with provision's inherited environment, so it has never honored `HERDR_BIN` either; those lines are unchanged by this diff.

scoop:
All four points were valid. Fixed in `524f7254`, `61cf19c4`, `21682ed2`. No disagreements this round.

**1 — `EnvironmentFile` / `HERDR_BIN` / `HERDR_SESSION`.** Confirmed the split-brain: `config.herdrSocketPath` resolves to `~/.config/herdr/sessions/<name>/herdr.sock` when `HERDR_SESSION` is non-default, and the CLI the liveness probe spawns inherits that env from `shepherd.service` (which has `EnvironmentFile`) — while our unit, lacking it, would serve `default`. Unit `active`, check `offline`, no agent can spawn. Added `EnvironmentFile=-%h/.shepherd/env`, and propagated the same `${HERDR_BIN:-herdr}` resolution (env file first, then PATH) through the `ExecStart` templating in both `provision.ts` and `update.sh`, and through `HERDR_ADOPT_SOCKET`'s `server stop`. You were also right that the `ExecStart` comment contradicted `src/remediations.ts` — `config.herdrBin` is `HERDR_BIN ?? "herdr"`, so it only falls back to a PATH lookup when `HERDR_BIN` is unset. Corrected.

**2 — the in-app Fix re-created the orphan I'd just eliminated.** And my own `StartLimitIntervalSec=0` (added last round for the restart-policy finding) made it strictly worse: the unit now retries forever rather than parking in `failed`. `HERDR_SERVE` now prefers `systemctl --user restart herdr` when a unit exists, keeping the detached `setsid`/`nohup` spawn for macOS, `buildOnly`'s `SHEPHERD_NO_SERVICE` path, and hand-rolled installs. I had deferred this to #1579; the interaction with `StartLimitIntervalSec=0` makes it this PR's problem, and you're right to have called it.

Adding that branch created a real test hazard: the hermetic sandbox's `PATH` still contained `/usr/bin`, so the unit branch could have resolved the **real** `systemctl` and restarted the developer's own herdr unit. `makeSandbox` now stubs `systemctl` (deny by default), and tests that want the unit branch install their own stub.

**3 — stale `ExecStart`.** Correct: `enable --now` does not restart an active unit. Added `systemctl --user try-restart herdr` after `daemon-reload` in both `installService` and `update.sh` (restarts iff active, no-op otherwise), with an ordering assertion in `test/provision.test.ts`.

**4 — leaking `PATH`.** Correct, and it was inconsistent with the `BACKUP_DIR` block a few lines up that deliberately subshells. The herdr resolution now happens entirely inside a subshell, so neither the `~/.local/bin` prepend nor anything sourced from `~/.shepherd/env` reaches the later `systemctl restart` / health check.

## Verification

Re-ran the three gate-eligible scenarios on the fixed tree (`herdr.service` asserted `active`, not merely "some daemon answers"):

| Scenario | Result |
| --- | --- |
| `install-e2e-service` (ubuntu, systemd) | PASS |
| `install-e2e` (ubuntu, no systemd — fallback branch) | PASS |
| `herdr-missing` (archlinux — fallback branch) | PASS |

Suite 6328 pass / 0 fail · `tsc` clean · lint clean · `fallow audit` clean · `systemd-analyze verify deploy/herdr.service` clean.

Still pre-existing and unrelated: `onboarding-gate.sh` is red on `node-too-old` **on `main` as well** (upstream codex installer broken; `applyVerbatim` bails on the first non-zero exit and the `optional` codex check sorts before `node`) — #1577.

scoop:
## Prior-round findings — all four verified fixed

**1. `EnvironmentFile` on the herdr unit.** Fixed. `deploy/herdr.service:20` now carries `EnvironmentFile=-%h/.shepherd/env`, matching `deploy/shepherd.service:27`. The `ExecStart` comment at `deploy/herdr.service:25-26` now states `config.herdrBin` is `HERDR_BIN ?? "herdr"` — verified against `src/config.ts:455`, so it no longer contradicts the resolution logic. The templating sites were propagated: `deploy/provision.ts:271` and `deploy/update.sh:114` both resolve `${HERDR_BIN:-herdr}` after sourcing the env file, and `HERDR_ADOPT_SOCKET` (`deploy/provision.ts:132`) uses `H="${HERDR_BIN:-herdr}"` instead of a bare `herdr`.

**2. `HERDR_SERVE` racing systemd.** Fixed. `src/remediations.ts:56-58` now branches to `systemctl --user restart herdr` when `systemctl --user cat herdr` succeeds, and only falls back to `setsid`/`nohup` (`:59-61`) otherwise. I traced the executor: `runRemediation` spawns `sh -c cmd` with no `env` override (`src/diagnostics.ts:143`), so on a provisioned host the Fix inherits `HERDR_BIN`/`HERDR_SESSION` from `shepherd.service`'s `EnvironmentFile` and the unit branch drives the unit rather than orphaning a daemon on its socket. Shell shape checked: the `{ if …; fi; }` group is the RHS of `||`, and `cmd &` before `else`/`fi` parses.

**3. Stale `ExecStart` on an already-active unit.** Fixed. `deploy/provision.ts:405` and `deploy/update.sh:124` both add `systemctl --user try-restart herdr` immediately after `daemon-reload`, before `enable --now` (`provision.ts:417`, `update.sh:129`).

**4. Leaking `PATH` in update.sh.** Fixed. `deploy/update.sh:110-115` confines the resolution to a `$( … )` command substitution, and the prepend is a `PATH=… command -v` assignment prefix (temporary, `command` being a regular builtin), so neither it nor anything from `~/.shepherd/env` reaches the `systemctl --user restart shepherd` at `:145` or the health check at `:156-163`.

## New blocking finding: `herdr server stop` never sees `HERDR_SESSION`

This is the same split-brain as prior finding 1, surviving in the two socket-adoption paths the PR added.

The premise is the diff's own: `deploy/herdr.service:15-19` states that without the env file "an env-blind daemon serves the `default` session while Shepherd's liveness probe targets the per-session socket." That is corroborated by `src/config.ts:39` (`herdrSession = process.env.HERDR_SESSION ?? "default"`) and `src/config.ts:44-48`, which routes a non-default session to `~/.config/herdr/sessions/<name>/herdr.sock`. So `herdr`'s session — for `server` **and** for `server stop` — comes from the process environment.

Both new stop calls run without it:

- `deploy/update.sh:126` — `"$HERDR_PATH" server stop` executes in the **outer** shell. The sourcing of `~/.shepherd/env` is sealed inside the command substitution at `:110-115`, whose exported variables die with the subshell. The outer shell never sourced the file at all.
- `deploy/provision.ts:131` — `HERDR_ADOPT_SOCKET` does `. "$HOME/.shepherd/env"` with **no** `set -a`. Under POSIX/bash a `KEY=value` assignment from a sourced file creates a shell variable without the export attribute, so `HERDR_SESSION` is visible to the in-shell `${HERDR_BIN:-herdr}` expansion (which is why `HERDR_BIN` appears to work) but is **not** in the environment of the `"$H" server stop` child at `:134`. The repo demonstrates the required form itself: `deploy/update.sh:67` wraps the identical `. "$HOME/.shepherd/env"` in `set -a` / `set +a` precisely so its `bun -e` child can read `SHEPHERD_DB`/`SHEPHERD_BACKUP_DIR`. The env file is documented as `KEY=value` lines with no `export` (`docs-site/src/content/docs/reference/configuration.md:7`), so there is no path by which the attribute is inherited.

Failure, on the ordinary upgrade path for an existing `HERDR_SESSION=<name>` host that already has a daemon (hand-rolled, or an in-app Fix's detached `setsid` child, which *does* inherit `HERDR_SESSION` from `shepherd.service`):

1. `bun run update` → `update.sh:124` `try-restart` is a no-op (`herdr.service` does not exist yet on a pre-#1574 host).
2. `:125` `is-active` is false → `:126` `herdr server stop` runs env-blind, so it addresses the **default** session and leaves the per-session daemon bound.
3. `:129` `enable --now herdr` starts the unit, which *does* read `HERDR_SESSION` from `EnvironmentFile` (`herdr.service:20`), binds the per-session socket, finds it taken, and `ExecStart` exits 1.
4. `Restart=always` + `StartLimitIntervalSec=0` (`herdr.service:11,36`) thrash it forever rather than parking in `failed`, while `defaultHerdrLiveness` (`src/diagnostics.ts:183`, `spawn(bin, ["agent","list"])`) reads **ok** because the orphan answers.

That is exactly the green-check/thrashing-unit state `ci/onboarding-harness/run.ts:170-175` and `deploy/provision.ts:228-235` exist to eliminate, reachable today. The harness cannot catch it: every scenario runs on the default session.

Fix: wrap the sourcing in `set -a` / `set +a` in `HERDR_ADOPT_SOCKET`, and in `update.sh` either hoist the sourcing into the `if [[ -n "$HERDR_PATH" ]]` block under `set -a` or run the stop inside its own `set -a` subshell (keeping point 4's no-leak property).

## Non-blocking

`deploy/update.sh:118-124` rewrites `$UNIT_DIR/herdr.service` and `try-restart`s unconditionally, so **every** `bun run update` bounces the herdr daemon even when the templated unit is byte-identical to the installed one. `try-restart` restarts iff active — it does not check whether anything changed. Prior finding 3 asked for the `try-restart`; it did not ask for it on every deploy. The repo otherwise works hard to avoid exactly this (`herdr update --handoff` in `src/herdr-update.ts:110-112`, and the `HerdrUnavailableError` maintenance guard). Gate the `daemon-reload` + `try-restart` on the templated output actually differing (`cmp -s`). `deploy/provision.ts:405` has the same shape.

## Verified consistent (no action)

- **i18n parity.** No new message keys. `diagnostics_hint_herdr_offline` already exists in both catalogs: `ui/messages/en.json:1269` and `ui/messages/de.json:1269`.
- **Auto-fix wiring.** `diagnostics_hint_herdr_offline` (`src/remediations.ts:106`) is absent from `GUIDANCE_ONLY` (`src/remediations.ts:126-129`), so `autoFixCommandFor` (`:142-145`) resolves it and `src/diagnostics.ts:539,568` mark the row fixable. Its stale entry in `ui/src/lib/diagnostics-docs.ts:7` renders harmlessly: `DiagnoseRows.svelte:96` takes the Fix button branch and `:107` is an `{:else if}`, so no double-render.
- **Subshell gating of the composed hint.** `diagnostics_hint_herdr_missing` = `${HERDR_INSTALL} && (${HERDR_SERVE})` (`src/remediations.ts:99`). The `exit 0`/`exit 1` at the end of `HERDR_SERVE` (`:63`) terminates the subshell with that status when parenthesized, and the whole `sh -c` when used bare for `herdr_offline` — both callers (`ci/onboarding-harness/apply.ts:75` `sh -c`, `src/diagnostics.ts:143` `sh -c`) read the top-level status. Correct in both uses.
- **`templateHerdrUnit`.** `/^ExecStart=.*$/m` (`deploy/provision.ts:119`) matches exactly one line in `deploy/herdr.service` (`:29`); the comment mentions at `:21` and `:34` are not at line-start `ExecStart=`.
- **`buildEnv` really carries `~/.local/bin`.** `ensureNodeGyp` returns `PATH: ${bunBin}:${localBin}:${env.PATH}` (`deploy/provision.ts:336`), so `probePath("herdr", buildEnv)` at `:394` can see a herdr that `install.sh` just dropped there.
- **`Wants=` not `Requires=`.** `deploy/shepherd.service:4,9` — a herdr that fails to start cannot cascade-fail Shepherd, matching the comment at `:6-8`.

## Could not verify

- herdr's own CLI contract (that `herdr server stop` is scoped by `HERDR_SESSION`, that a second `herdr server` against a bound socket exits 1, that `herdr 0.7.3` does not auto-spawn on a CLI call). The binary is not in this repo. The blocking finding above does not depend on external verification: it depends only on the diff's **own** stated premise (`deploy/herdr.service:15-19`) that `herdr server` is env-scoped by `HERDR_SESSION`, which makes the env-blind `server stop` internally inconsistent with it.

- deploy/provision.ts:405 — `run("systemctl", ["--user", "try-restart", "herdr"])` goes through the throwing `defaultRunner` (`deploy/provision.ts:219,225`), while the sibling call at `deploy/update.sh:124` is guarded with `|| true` plus an `is-active` recovery fallback (`:125-128`). A `try-restart` that fails on an active unit aborts the entire provision even though `HERDR_ADOPT_SOCKET` + `enable --now` two lines below (`:413,417`) would have healed it. Not reachable today (an active unit implies a working `ExecStart`), but the guard is present on one path and missing from its sibling.
- deploy/provision.ts:266-278 — `probeBinPath` sources `~/.shepherd/env` before `command -v`, so an env file that sets `PATH` (a documented `KEY=value` override file) replaces `buildEnv`'s `~/.local/bin` prefix from `deploy/provision.ts:336`. If that `PATH` omits `~/.local/bin`, `installService` throws `cannot install herdr.service: herdr is not on PATH` (`:395`) and hard-fails a provision that previously succeeded.

- deploy/install.sh:266 sources `~/.shepherd/env` without `set -a`, the same missing-export shape as the blocking finding above; harmless there today because the sourced values are only read in-shell.

scoop:
All three valid, fixed in `bb292002`. Nothing disputed.

**1 + 2 — the same bug in two places, and I verified the mechanism rather than assuming it.**

```
$ sh -c '. envfile; sh -c "echo child sees FOO=[$FOO]"'
child sees FOO=[]                 # bare source: the child sees nothing
$ sh -c 'set -a; . envfile; set +a; sh -c "echo child sees FOO=[$FOO]"'
child sees FOO=[bar]              # set -a: the child sees it
$ sh -c '. envfile; echo in-shell FOO=[$FOO]'
in-shell FOO=[bar]                # why HERDR_BIN *appeared* to work
```

That third line is exactly why this slipped past me: `HERDR_BIN` is expanded in-shell, so the sourcing looked correct, while `HERDR_SESSION` / `HERDR_SOCKET_PATH` — which only matter to the `server stop` **child** — were silently invisible. In `update.sh` the sourcing was additionally sealed inside the `$( … )` that computes `HERDR_PATH`, so it never reached the outer shell at all.

Consequence is as you describe: on a non-default-session host the stop addresses the default socket, the real orphan stays bound, and `enable --now herdr` thrashes forever (`Restart=always` + `StartLimitIntervalSec=0`) while the check reads `ok` because the orphan answers — on the ordinary upgrade path, which is worse than the fresh-install case this PR started from.

Fixed: `set -a` / `set +a` around the source in `HERDR_ADOPT_SOCKET`, and in `update.sh` the stop now runs in its own `set -a` subshell — which keeps the no-leak property from the previous round (nothing sourced reaches the later `systemctl restart` / health check).

**3 — a user-visible regression I introduced.** `try-restart` kills the daemon backing every live agent session, and I was firing it on every `bun run update` regardless of whether the templated unit differed. Now gated on `cmp -s` in `update.sh` and a content compare in `installService`; a missing installed unit reads as "changed", so first install still works. New test asserts `try-restart` is *not* issued when the unit is byte-identical.

## Verification

| Scenario | Result |
| --- | --- |
| `install-e2e-service` (ubuntu, systemd — `herdr.service` asserted `active`) | PASS |
| `install-e2e` (ubuntu, no systemd — fallback branch) | PASS |

Suite 6330 pass / 0 fail · `tsc` clean · lint clean · `fallow audit` clean · `bash -n deploy/update.sh` clean.

Three new tests pin these specifically: `try-restart` skipped on an unchanged unit, `set -a` ordering inside the adopt snippet, and the unit-preferred restart branch (with `systemctl` stubbed so the sandboxed shell tests can never reach a developer's real user manager).

Pattern worth naming across these rounds: each fix produced the next defect one layer out — orphan-thrash from provision, then the same thrash from the in-app Fix, now the same thrash from an unexported env var on the upgrade path. The tests above are meant to make the next one fail in CI rather than in a container.

PR #1534, opened 2026-07-07
feat: add repo preview open mode

kai-osthoff:
## Summary
- add a persisted repo-level preview open mode (`ask`, `inline`, `tab`) through store, API, demo data, and UI config
- make Preview chips wait for the repo-config fetch to settle, then use the persisted mode or fall back to the chooser if loading fails
- reuse `buildPreviewUrl` for new-tab URLs and open with `window.open(url, "_blank", "noopener,noreferrer")`

## Validation
- `bun run lint`
- `bun run typecheck`
- `bun run test`
- `cd ui && bun run check:i18n && bun run check`
- `cd ui && bun run test`
- `cd extension && bun run check`
- `bun run check:announcement-versions && bun run check:glossary`
- pre-push hook passed

Manual steps: none.

PR #1585, opened 2026-07-10
fix(git): cache the git-rail's live PR fetch so the session card shows it

kai-osthoff:
## Problem

A session's **card** in the herd list showed **no PR badge**, even though the same session's **detail view showed the PR** (reported with a card blank while the terminal read `PR #1534`).

## Root cause

PR state isn't on the `Session` object — it lives in a separate `GitState`, and the two surfaces read it from **different sources**:

| Surface | Source | Writes `prCache`? |
| --- | --- | --- |
| Session card | `store.git[id]` — `GET /api/git` snapshot of `prCache` + `session:git` WS pushes (filled **only** by the background `PrPoller`) | reader |
| Detail git-rail | its own live `GET /api/sessions/:id/git` → `resolveGitState`, polled ~15s | **NO — the bug** |

`resolveGitState` ran a fresh, trust-guarded `gh` lookup on every call but **never wrote back to `prCache` nor emitted `session:git`** — a read-only side channel the card's cache never saw. It also (unlike the poller's `statusPerSession`) never reconciled a **renamed** worktree branch, so it could itself return `none` for a session whose PR is real but under a branch the stored `session.branch` no longer names.

## Fix (server-side; no UI change)

1. **`resolveGitState` now reconciles a renamed branch** — on a `none` result it adopts the live worktree branch via `service.syncWorktreeBranch(id)` and retries under the same trust guard, mirroring the poller. This makes a resolved `none` **authoritative**.
2. **`forgeGitStateResponse` writes through + emits** — after resolving, it `prCache.set` + emits `session:git` when the result is *meaningful* (`state !== "none"`, **or** it clears a PR that was really cached) **and** actually changed (`gitStateChanged`, the poller's own gate). So it **surfaces** a PR the poller missed and **clears** one that's genuinely gone, but never emits a no-PR `none` for a session that never had one.

The card updates reactively through the existing `session:git` store handler — no UI change.

### Bounded emits / non-UI subscribers
`session:git` also drives `sessionRouter.onGit` (critic), `autoMerge.onGit`, `draftReconcile.onGit`, and `holdService.recompute`. Because the emit is gated on a genuine `gitStateChanged` over the **shared** `prCache`, whichever of the GET path / poller writes first flips the gate and the other stays silent → at most one emit per real change; repeated 15s git-rail polls of an unchanged PR are quiet.

### Known residual (documented, self-healing)
If `worktree.currentBranch` transiently fails during a GET reconcile while the poller holds a correct open PR, the GET could resolve `none` and briefly clear the card — re-surfaced within ≤15s by the poller's fast tick. Eliminating it fully would require changing `syncWorktreeBranch`'s contract (unchanged vs. unknown); out of scope.

## Verification
The live `/api/sessions/:id/git` endpoint sits behind the authenticated port, so rather than manufacture the exact transient state against the running instance, both sub-paths are verified reproducibly through the GET handler in `test/server-git.test.ts`:

- **Direct hit:** open PR + empty cache → cached + one `session:git` emit.
- **Reconcile:** stored branch `none`, live branch has the PR → adopted + surfaced.

## Testing
- `test/server-git.test.ts` — 6 write-through cases (surface, idempotent repoll, silent `none`, clear stale PR, renamed-branch reconcile, collision-guard regression). 36 pass.
- Regression across `resolveGitState`/`session:git` consumers: `server-review-pr`, `server-actions`, `pr-poller`, `cross-consumer-snapshot`, `automerge`, `resume-serialize`, `draft-reconcile`, `hold-service` — all green.
- `bun run typecheck` and `bun run lint` clean.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -130,7 +130,7 @@ import { buildIssueUrl } from "./forge";
 import type { GithubRateLimitPayload } from "./forge/github-rate-limit";
 import { BaseCheckoutBusyError, MergeConflictError } from "./forge/local";
 import { recordEpicIntegrationIfChild, settleMergedSession } from "./merge-teardown";
-import { type PrCache, guardStaleTerminal, trustsTerminal } from "./pr-poller";
+import { type PrCache, gitStateChanged, guardStaleTerminal, trustsTerminal } from "./pr-poller";
 import {
   readRepoRoles,
   writeRepoRoles,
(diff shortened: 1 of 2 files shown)


kai-osthoff (commented):
Findings: none.

Verification:
- The changed GET path is actually used for `GET /api/sessions/:id/git`: `dispatchForgeAction` routes bare GETs to `forgeGitStateResponse` (src/server.ts:3497, src/server.ts:3498), and `handleSessionGit` is the `/api/sessions/:id/git` handler (src/server.ts:3590, src/server.ts:3598).
- The write-through uses the existing cache/event contract: `PrCache` defines `get` and `set` (src/pr-poller.ts:9, src/pr-poller.ts:15), `gitStateChanged` is exported and compares rendered PR fields (src/pr-poller.ts:86, src/pr-poller.ts:101), and the new server path sets `prCache` then emits `session:git` only on a meaningful changed state (src/server.ts:3572, src/server.ts:3586).
- The emitted payload shape is consumed by the UI as `{ id, git }`: the websocket type declares `session:git` with that payload (ui/src/lib/types.ts:1519), and the store merges it into `store.git` by session id (ui/src/lib/store.svelte.ts:525, ui/src/lib/store.svelte.ts:527).
- The frontend path that can observe the PR already calls this endpoint: `gitState()` fetches `/api/sessions/${id}/git` (ui/src/lib/api.ts:1094, ui/src/lib/api.ts:1099), GitRail loads that state into its local `git` value (ui/src/lib/components/GitRail.svelte:124, ui/src/lib/components/GitRail.svelte:129), and it continues polling open PRs while visible (ui/src/lib/components/GitRail.svelte:176, ui/src/lib/components/GitRail.svelte:182). The overview receives the shared `store.git` map (ui/src/routes/+page.svelte:2518, ui/src/routes/+page.svelte:2520).
- The renamed-branch retry mirrors the poller contract: the poller retries a live branch after a stored-branch `none` (src/pr-poller.ts:420, src/pr-poller.ts:440), `syncWorktreeBranch` returns the adopted branch after updating the store (src/service.ts:3575, src/service.ts:3589), and the server's `resolveGitState` now retries `prStatus(live)` under the same guard (src/server.ts:3544, src/server.ts:3554).
- Added tests exercise the cache write, unchanged repoll suppression, silent uncached `none`, stale-PR clearing, and renamed-branch recovery (test/server-git.test.ts:225, test/server-git.test.ts:291). The test `PrCache` double now mutates an observable snapshot, so the second GET assertions use the same `get`/`set` behavior as the production interface (test/server-git.test.ts:130, test/server-git.test.ts:140).

No build, lint, or test command was run, per the read-only review instruction.

kai-osthoff:
@scoop 🟢✨ It's READY 🎉🚀 anytime you feel like merging 😏🐌💤… no rush 🙄👑

PR #505, opened 2026-06-10
feat(stepper): planning-aware stage bar with verdict tints + hover legend

scoop:
## What

The per-task five-segment bar was opaque: planning and implementing were lumped into one "coding" stage, sighted desktop users had no way to learn what the segments mean, and CI had a whole segment while review state showed nothing.

- **New stage order:** planning → implementing → pr → review → ready (was coding → pr → ci → review → ready). Planning is its own stage driven by `session.planPhase`; sessions without the plan gate show it as skipped.
- **CI folds into the PR segment** as tint: pending pulses (reduced-motion-guarded), failure red + shape cue (height/ring, WCAG 1.4.1), success green.
- **Review segment tints by verdict:** amber pulse while the critic runs, red + same shape cue on changes-requested, green on approved/non-blocking.
- **Ready is now derived** (the one behavior change): open PR + green checks + `mergeable === true` + not draft + approved review or critic commented on the *current* head SHA — fail-closed on every signal, incl. an undefined-headSha guard (mutation-tested pins). Previously segment 5 could never actually fill (the manual flag hides the stepper; merged shows the chip).
- **Hover legend popover (desktop):** all stages listed with live swatches, state words (done/now/pending/skipped) and CI/review verdict words; current stage highlighted. Hover-only (`(hover: hover)` guarded), `aria-hidden` (the aria-label already announces everything), fixed-positioned (cards clip), closes on capture-phase scroll, flips near viewport top, tokens only.
- aria-label now carries stage + CI word + review word in text.

## Plan & process

Executed from the approved `.shepherd-plan.md` plan-gate plan (session artifact, gitignored), subagent-driven with per-task spec + quality reviews; two review findings (a vacuous safe-fail test, cognitive complexity 20 in `deriveStage`) were fixed and re-verified by mutation testing.

## i18n / feature catalog

- EN+DE parity (870 keys): renamed `activity_stage_coding`→`activity_stage_implementing`, deleted `activity_stage_ci`, added planning/review/legend keys.
- Feature-catalog entry `stepper-stage-legend` (`sinceVersion: 1.23.0`); `targetId` omitted deliberately — the stepper repeats per card and `coachTarget` anchors are last-write-wins singletons.

## Verification

- `ui` svelte-check: 0 errors / 0 warnings (1852 files)
- vitest: 62 files, 851 tests passed (44 stage-model tests incl. derived-ready fail-closed matrix; new `Stepper.browser.test.ts` locks the aria-label contract, mutation-verified)
[...]

PR #1587, opened 2026-07-10
fix(stepper): explain the pipeline lamps on hover/focus and repair the legend

kai-osthoff:
## What & why

The 5-segment progress bar at the bottom-right of each session card (the "Lämpchen") had a hover legend shipped in #505 that **never appeared**. Root cause: the card's transparent `.unit-hit` click overlay (`z-index:0`) is painted above the card's static content and captures the pointer, so the stepper's `mouseenter` never fired. The lamps stayed *visible* (the overlay is transparent), so to a user it read as "no explanation at all."

This repairs and enriches that explanation so the pipeline stages — **Planning → Implementing → PR → Review → Ready** — are self-explanatory to mouse, keyboard, and screen-reader users.

## Changes

- **Fix the hover bug** — raise the bar above `.unit-hit` and make it a real `<button>` (control-only raise, so the terminal MERGED/CLOSED chip stays unraised and still selects the row).
- **Rebuild the legend** on the existing `InfoTip`/`anchorPopover` recipe: a native top-layer `role="tooltip"` popover placed with `@floating-ui/dom` flip/shift — replacing the fragile `position:fixed` + `LEGEND_W`/`FLIP_BELOW_PX` math it likely failed on, and escaping the card's `overflow:hidden`.
- **Enrich** — each legend row gains a plain-language description of the stage (EN + DE); the state word stays aligned when long German copy wraps. `Review` no longer claims CI is running; `Ready` matches the actual `isDerivedReady` merge-gate.
- **Discoverability & a11y** — `cursor: help` + a `::before` halo enlarging the 3px target to ~19px; opens on hover **and** keyboard focus (fine-pointer-gated, so a touch tap selects the row instead of popping the tooltip); dismisses on Esc/blur/outside/scroll. The button's accessible name **discloses its action** ("Open session · Progress: …") since activating forwards the same `onselect` as the primary card button; segments are `aria-hidden`.
- **Design system** — tokenize the shared popover elevation as `--shadow-popover` in `app.css` (no new raw `rgba` in component CSS).

## Tests

New `UnitRow.browser.test.ts` regressions prove the bar is pointer-targetable above `.unit-hit`, that a hover entry opens the legend, and that clicking it — **including a coordinate in the enlarged `::before` halo, outside the visible 3px bar** — forwards exactly one `onselect(session.id)` while hover does not. `Stepper.browser.test.ts` covers the `<button>` markup, the disclosed name, and focus→open. 

`cd ui && bun run check && bun run check:i18n && bun run test` — all green (3129 tests). No feature-catalog entry needed (this is a `fix`, and the existing #505 announcement that promised the hover legend simply becomes true).


kai-osthoff (commented):
Findings:

- `ui/src/lib/components/Stepper.svelte`: `Stepper` now renders the tooltip as a sibling `<div>` at `ui/src/lib/components/Stepper.svelte:208`, while the in-diff call site still places `<Stepper>` inside `<span class="meta-stepper">` at `ui/src/lib/components/UnitRow.svelte:608` and closes that wrapper at `ui/src/lib/components/UnitRow.svelte:616`. That creates invalid `<div>`-inside-`span` markup for the new popover. The app is client-rendered (`ui/src/routes/+layout.ts:1`), so I am not treating this as a hydration blocker, but the component should either render the tooltip as a phrasing element such as `<span>` or the UnitRow wrapper structure should become non-phrasing.

Verification notes:

- The new `anchorPopover` dependency is imported at `ui/src/lib/components/Stepper.svelte:6` and exported with the expected `(reference, floating, gap, placement)` contract at `ui/src/lib/floating-anchor.ts:16`.
- The new tooltip/open-session message keys used in `Stepper` at `ui/src/lib/components/Stepper.svelte:40` and `ui/src/lib/components/Stepper.svelte:104` exist in both locale catalogs at `ui/messages/en.json:568` and `ui/messages/de.json:568`.
- The row activation path is wired from `UnitRow` into `Stepper` at `ui/src/lib/components/UnitRow.svelte:614`; the existing row overlay selection path remains at `ui/src/lib/components/UnitRow.svelte:453`.

kai-osthoff (commented):
No findings.

Verified scope: the PR diff contains `ui/src/lib/components/Stepper.svelte`, `ui/src/lib/components/UnitRow.svelte`, `ui/messages/en.json`, `ui/messages/de.json`, `ui/src/app.css`, and the two browser test files.

Prior finding resolved: `Stepper` now renders the activatable bar as a `button.stepper` and the tooltip as a phrasing `span` with `role=tooltip` and `popover=manual`, not a `div`; the in-scope caller still wraps it in `span.meta-stepper`, so the previous div-inside-span issue is not present in the new diff (`ui/src/lib/components/Stepper.svelte:183`, `ui/src/lib/components/Stepper.svelte:210`, `ui/src/lib/components/UnitRow.svelte:607`).

Tooltip coverage verified: the stage list has five stages (`planning`, `implementing`, `pr`, `review`, `ready`), `Stepper` maps each `Stage` to a `stepper_desc_*` message, and the legend iterates `STAGE_ORDER` while rendering `STAGE_DESC[stage]()` for each row (`ui/src/lib/components/stage.ts:4`, `ui/src/lib/components/Stepper.svelte:39`, `ui/src/lib/components/Stepper.svelte:211`, `ui/src/lib/components/Stepper.svelte:226`).

Identifier and contract checks: `anchorPopover` is imported by `Stepper` and exported with the called `(reference, floating, gap, placement)` shape; it positions a native popover and hides it on cleanup (`ui/src/lib/components/Stepper.svelte:6`, `ui/src/lib/components/Stepper.svelte:140`, `ui/src/lib/floating-anchor.ts:16`). The only production `Stepper` caller passes the new optional `onactivate` callback, while standalone renders remain supported by the optional call (`ui/src/lib/components/UnitRow.svelte:609`, `ui/src/lib/components/UnitRow.svelte:614`, `ui/src/lib/components/Stepper.svelte:23`, `ui/src/lib/components/Stepper.svelte:193`).

Row interaction verified against surrounding code: the row click target is a sibling `.unit-hit` button at `z-index: 0`, while the stepper button is raised with `position: relative; z-index: 1`, so the hover/click target is not hidden behind the overlay (`ui/src/lib/components/UnitRow.svelte:447`, `ui/src/lib/components/UnitRow.svelte:737`, `ui/src/lib/components/Stepper.svelte:238`).

Locale parity verified: every new user-facing key used by `Stepper` exists in both EN and DE catalogs (`ui/messages/en.json:568`, `ui/messages/en.json:573`, `ui/messages/de.json:568`, `ui/messages/de.json:573`).

CSS token reference verified: the new `--shadow-popover` token is defined in the semantic token layer and consumed by the stepper legend (`ui/src/app.css:126`, `ui/src/lib/components/Stepper.svelte:318`).

Tests were not run, per the review instruction not to build or run anything.

kai-osthoff:
✨🚀💅 Oh @scoop, this masterpiece is READY ✅🎉 — merge it someday? 🐌⏳🙏😴

PR #1599, opened 2026-07-10
fix(onboarding): failing optional remediation must not abort the apply (#1577)

### LINKED ISSUE #1577: Onboarding harness: a failing `optional` remediation aborts the whole apply, redding the release gate

scoop:
Closes #1577.

## Problem

A broken **third-party, optional** installer red the **release gate** for an unrelated scenario. `scripts/onboarding-gate.sh` exited 1 on `node-too-old` (node stayed `warning`) because:

- `applyVerbatim` (`ci/onboarding-harness/apply.ts`) bailed on the **first** non-zero exit.
- `remediationsFor` includes every non-ok check with a fix — including `codex` in state **`optional`** (`diagnostics_hint_codex_optional`, which codex reports whenever `claude` is present). `codex` sorts **before** `node`, so the upstream-broken codex install failed, the loop returned early, and `node`'s remediation never ran.

Blast radius is contained to the harness: `remediationsFor` is consumed only by `apply.ts` + `run.ts`. The in-app Fix endpoint and cold-start installer use `autoFixCommandFor` per-check and are unaffected.

## Fix

1. **Optional failures are non-fatal.** New `remediationEntriesFor` returns `{ id, cmd, optional }` (tagging the check's `optional` state); `remediationsFor` is now a thin `.map(r => r.cmd)` projection so its `string[]` contract — and `run.ts`'s verbatim-vs-agent `.length` gate — are byte-for-byte unchanged. `applyVerbatim` still attempts the optional remediation best-effort, but a non-zero exit is logged and skipped; **required** (`error`/`warning`) failures still gate (fail-closed).
2. **Bounded retry** (default 2 attempts, ~3s delay; injectable so tests stay instant) around each remediation, so one transient network/mirror flake can't red the gate.

## Scope note

The `node-pty` tarball extraction flake mentioned in the issue was in the **`install.sh`** path (installE2E scenarios), which this `applyVerbatim` retry does **not** cover — deliberately out of scope, called out so it isn't mistaken for fixed.

## Tests

Added to `test/onboarding-harness/`: optional-failure-non-fatal (node still runs, apply succeeds), required-failure-still-fatal, transient-retry-succeeds, and `remediationEntriesFor` optional-flag tagging.

- `bun run lint` ✓
- `bun test ./test` ✓ (6347 pass, 0 fail)

### DIFF
--- a/ci/onboarding-harness/apply.ts
+++ b/ci/onboarding-harness/apply.ts
@@ -1,7 +1,7 @@
 import { readFileSync } from "node:fs";
 import { join } from "node:path";
 import type { IncusDriver } from "./incus";
-import { remediationsFor } from "../../src/remediations";
+import { remediationEntriesFor } from "../../src/remediations";
 import type { DiagnosticsSnapshot } from "../../src/types";
 
 export interface CoachingLine {
@@ -64,16 +64,66 @@ function shellQuote(s: string): string {
   return `'${s.replace(/'/g, `'\\''`)}'`;
 }
 
-/** Run each harness-catalog remediation (keyed by Shepherd's emitted hintKeys)
- *  verbatim inside the instance. Returns false if any command exits non-zero. */
+const sleep = (ms: number) => new Promise<void>((resolve) => setTimeout(resolve, ms));
+
+/** Run one remediation inside the instance with a bounded retry, so a single transient
+ *  network/mirror flake in a download-and-install command doesn't fail the apply (#1577).
+ *  Returns the final exit code (0 ⇒ it eventually succeeded within `attempts`). */
+async function runWithRetry(
+  driver: IncusDriver,
+  name: string,
+  cmd: string,
+  attempts: number,
+  delayMs: number,
+): Promise<number> {
+  let code = 0;
+  for (let i = 1; i <= attempts; i++) {
+    const r = await driver.exec(name, ["sh", "-c", cmd]);
+    code = r.code;
+    if (code === 0) return 0;
+    if (i < attempts) await sleep(delayMs);
+  }
+  return code;
+}
+
+/** Options for {@link applyVerbatim}. `attempts`/`delayMs` bound the per-remediation
+ *  transient-flake retry; tests pass `delayMs: 0` to stay instant. */
+export interface ApplyVerbatimOpts {
+  attempts?: number;
+  delayMs?: number;
+}
+
+/** Run each harness-catalog remediation (keyed by Shepherd's emitted hintKeys) verbatim
+ *  inside the instance, each with a bounded retry. Returns false when a REQUIRED
+ *  (non-optional) remediation still fails after its retries — that's a real regression and
+ *  must gate. A failing OPTIONAL-state remediation (an equivalent alternative is already
+ *  healthy — e.g. a broken third-party `codex` installer while `claude` is present) is
+ *  logged and skipped, never fatal: it must not red the release gate for an unrelated
+ *  defect (#1577). */
 export async function applyVerbatim(
   driver: IncusDriver,
   name: string,
   snapshot: DiagnosticsSnapshot,
+  opts: ApplyVerbatimOpts = {},
 ): Promise<boolean> {
-  for (const cmd of remediationsFor(snapshot)) {
-    const r = await driver.exec(name, ["sh", "-c", cmd]);
-    if (r.code !== 0) return false;
+  const attempts = opts.attempts ?? 2;
+  const delayMs = opts.delayMs ?? 3000;
+  for (const { id, cmd, optional } of remediationEntriesFor(snapshot)) {
+    // Retry guards a REQUIRED remediation against a transient flake that would red the
+    // gate. An optional remediation's failure is non-fatal anyway, so retrying a
+    // structurally-broken optional installer (e.g. codex) only burns a needless attempt +
+    // sleep every run — attempt it once, then skip. (#1577)
+    const code = await runWithRetry(driver, name, cmd, optional ? 1 : attempts, delayMs);
+    if (code !== 0) {
+      if (optional) {
+        console.log(
+          `[${name}] optional remediation for ${id} failed (exit ${code}); ` +
+            `continuing — not fatal to the apply`,
+        );
+        continue;
+      }
+      return false;
+    }
   }
   return true;
 }
--- a/src/remediations.ts
+++ b/src/remediations.ts
@@ -128,11 +128,31 @@ export const GUIDANCE_ONLY: ReadonlySet<string> = new Set([
   "diagnostics_hint_git_missing",
 ]);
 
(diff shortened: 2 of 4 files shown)


scoop (commented):
## Summary

This PR fixes the reported root cause: `applyVerbatim` no longer aborts the whole apply loop when a remediation for an `optional`-state check fails. A new `remediationEntriesFor` tags each entry with its check's `optional` state, and `applyVerbatim` logs-and-continues on an optional failure while still gating on a required one. It also adds a bounded per-remediation retry for transient network flakes. Both suggested fixes from the issue are implemented.

The change is backward-compatible and correctly scoped.

### Verified
- `remediationEntriesFor`/`Remediation`/`remediationsFor` are exported and internally consistent; `remediationsFor` is a pure projection preserving the `string[]` contract (`src/remediations.ts:136,144,154`).
- `DiagnosticCheck` carries `id`/`state`/`hintKey` and `DiagnosticState` includes `"optional"` (`src/types.ts:429,434`), so `c.id` and `c.state === "optional"` are valid.
- Both non-test callers remain compatible: `remediationsFor` at `ci/onboarding-harness/run.ts:251` (unchanged), `applyVerbatim` at `run.ts:222,252` (new `opts` param defaults, so existing calls are unaffected).
- Test hintKeys (`diagnostics_hint_codex_optional`:110, `diagnostics_hint_node_outdated`:90, `diagnostics_hint_bun_missing`:88) and command markers (`fnm.vercel.app`:16, `bun.sh/install`:88) all resolve in `src/remediations.ts`.
- Retry arithmetic is correct: `runWithRetry` returns 0 on eventual success; the flake-once test yields exactly 2 calls (`apply.ts:79-87`).
- No user-facing strings or message catalogs are touched (only CI `console.log`), so no i18n/locale-parity concern applies.

### Non-blocking observation
The bounded retry (`attempts: 2, delayMs: 3000`) is applied to *every* remediation, including the `optional` ones and structurally-broken (non-transient) installers. For the exact case that motivated this PR — the upstream-broken `codex` installer — each harness run now executes that failing install twice with a 3s inter-attempt sleep before skipping it, on every scenario carrying a `codex: optional` check. Retry exists for *transient* flakes; a deterministically-broken optional installer just pays 2x the failed-install cost plus the delay each run. Consider skipping the retry for `optional` entries (or only retrying and reporting on the required path), or aggregating the optional failures into a single end-of-apply report as the issue's fix #1 suggested ("collect and report"). Purely a latency/clarity nit — the behavior is correct.

scoop (commented):
## Summary

This revision correctly implements the issue's fix #1 (a failing `optional` remediation is non-fatal) and #2 (bounded retry for transient flakes), and it resolves the prior review's non-blocking finding.

## Prior finding — resolved

> The bounded retry (attempts:2, delayMs:3000) is applied to optional remediations too, so a structurally-broken optional installer (e.g. codex) is run twice with a 3s sleep before being skipped.

Addressed. `ci/onboarding-harness/apply.ts:118` now calls `runWithRetry(driver, name, cmd, optional ? 1 : attempts, delayMs)`, so an optional remediation is attempted exactly once (no retry, no sleep) before being logged and continued past. Verified against the new test `test/onboarding-harness/apply.test.ts:80` which asserts `codexCalls.length === 1` (`attempted once, NOT retried`).

## Verification

- `remediationEntriesFor` (src/remediations.ts:144) reads `c.id` and `c.state === "optional"`; the `DiagnosticCheck` shape supplies `id`/`state`/`hintKey` (src/types.ts:434-437) and `DiagnosticState` includes `"optional"` (src/types.ts:429).
- The legacy `remediationsFor` is preserved as a pure `.map((r) => r.cmd)` projection (src/remediations.ts:154-156); its sole `.length` consumer at ci/onboarding-harness/run.ts:251 is unchanged, and the string test at test/onboarding-harness/remediations.test.ts:32 still holds.
- Both `applyVerbatim` call sites omit `opts` (ci/onboarding-harness/run.ts:222 and 251-252), inheriting `attempts:2/delayMs:3000`. A REQUIRED remediation that still fails returns `false`, and the preflight path throws on that (run.ts:222-224) — the release gate still reds on real regressions.
- Only `state === "optional"` is treated non-fatal; `warning` (e.g. `node-too-old`) and `error` remain required — matching the issue's requirement that node's fix must still run and gate.
- `console.log` is consistent with existing harness logging (run.ts uses it throughout).

No blocking issues found.

- The bounded retry now applies to every REQUIRED remediation, not only network-fetch ones (the issue's fix #2 scoped it to "network-fetch remediations"). This is a reasonable superset — a deterministically-failing required remediation costs one extra attempt + one 3s sleep before it correctly gates — so it is acceptable as written.
- The author chose "skip retry for optional" over "aggregate optional failures into an end-of-apply report"; per-failure `console.log` + `continue` satisfies the issue's "collect and report, then continue" intent.

PR #1600, opened 2026-07-10
fix(herdr): reset-failed before restart in herdr_offline Fix path

### LINKED ISSUE #1579: In-app Fix for `herdr_offline` spawns a daemon inside Shepherd's cgroup; prefer the systemd unit when it exists

scoop:
Closes #1579.

## Context

Most of #1579 already shipped when #1574 actually merged (commit `a91ce9e4`, "#1580"), filed after the issue was deferred: `HERDR_SERVE` already prefers the systemd unit (`restart herdr`) over a detached child, provision's `HERDR_ADOPT_SOCKET` already does socket-adopt + `reset-failed` before `enable --now herdr`, and `herdr.service` already carries `StartLimitIntervalSec=0`. The issue's "related hazard" (process-group SIGKILL on the 120s timeout) is explicitly informational and currently unreachable.

## The remaining delta

The in-app Fix remediation `HERDR_SERVE` ran `systemctl --user restart herdr` with **no** preceding `reset-failed`. On a **legacy or hand-edited** unit lacking `StartLimitIntervalSec=0` that crash-looped against a held socket, the start-limiter parks it in `failed` and a bare `restart` is then refused ("start request repeated too quickly") until the failed state is cleared. This adds `systemctl --user reset-failed herdr` before the restart, mirroring provision's `HERDR_ADOPT_SOCKET`.

- `;`-sequenced (not `&&`): a non-zero `reset-failed` (nothing to clear) must not block the restart.
- On a host provisioned by the **current** `deploy/` the unit never parks in `failed` from the limiter, so `reset-failed` is a **strict no-op** there — the comment says so, to stop a future reader over-reading the line. Its only value is legacy/hand-edited units.

## Test

Extended the existing "drives the systemd unit" behavioral test rather than adding a source-substring check. The systemctl stub `exit 1`s on the unmatched `reset-failed` arg, so once the line is inserted the test exercises the real failing-`reset-failed`→`restart` path: it asserts `sysLog` shows `reset-failed herdr` before `restart herdr` **and** `r.status === 0` — a `;`→`&&` regression would flip status to 1 and never reach restart.

## Verification

- `bun run lint` clean
- `bun test ./test/remediations.test.ts ./test/provision.test.ts ./test/diagnostics.test.ts ./test/onboarding-harness/remediations.test.ts` green (140 tests)

### DIFF
--- a/src/remediations.ts
+++ b/src/remediations.ts
@@ -34,12 +34,19 @@ export const HERDR_INSTALL = "curl -fsSL https://herdr.dev/install.sh | bash";
  *   - `agent list || { start }` — IDEMPOTENT: a live daemon short-circuits, so this never
  *     races a second server against a bound socket (a second `herdr server` against a bound
  *     socket exits 1 with "herdr server is already running" — verified, #1574).
- *   - `systemctl --user restart herdr` WHEN THE UNIT EXISTS — on a provisioned host,
- *     `herdr: offline` means `herdr.service` itself cannot bind. Spawning a detached daemon
- *     there would put an unsupervised process on the socket, so the unit's ExecStart exits 1
- *     forever (`StartLimitIntervalSec=0` means it never even parks in `failed`) while the check
- *     reads `ok` because the orphan answers — exactly the green-check/thrashing-unit state this
- *     whole change exists to eliminate. Drive the unit; never race it.
+ *   - `systemctl --user reset-failed herdr` THEN `restart herdr` WHEN THE UNIT EXISTS — on a
+ *     provisioned host, `herdr: offline` means `herdr.service` itself cannot bind. Spawning a
+ *     detached daemon there would put an unsupervised process on the socket, so the unit's
+ *     ExecStart exits 1 forever (`StartLimitIntervalSec=0` means it never even parks in `failed`)
+ *     while the check reads `ok` because the orphan answers — exactly the green-check/thrashing-unit
+ *     state this whole change exists to eliminate. Drive the unit; never race it. The leading
+ *     `reset-failed` mirrors provision's `HERDR_ADOPT_SOCKET`: on a host provisioned by the current
+ *     `deploy/` the unit carries `StartLimitIntervalSec=0` and so NEVER parks in `failed` from the
+ *     start-limiter, making `reset-failed` a strict no-op there — its only value is a LEGACY or
+ *     hand-edited unit lacking that setting, which a crash-loop against a held socket can park in
+ *     `failed` with the limiter tripped, whereupon a bare `restart` is refused ("start request
+ *     repeated too quickly") until the failed state is cleared. `;`-sequenced, not `&&`: a non-zero
+ *     `reset-failed` (nothing to clear) must NOT block the restart.
  *   - `setsid … || nohup …` — the FALLBACK, for hosts with no unit (macOS, `buildOnly`'s
  *     SHEPHERD_NO_SERVICE path, hand-rolled installs). Detach so the daemon outlives the shell
  *     (an `incus exec` session, or the in-app Fix endpoint's child). macOS has NO `setsid`, so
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

This PR adds `systemctl --user reset-failed herdr` before `restart herdr` in the `HERDR_SERVE` remediation's systemd-unit branch, so a legacy/hand-edited unit that a crash-loop parked in `failed` (with the start-limiter tripped) is cleared before the restart, which would otherwise be refused ("start request repeated too quickly"). The change is minimal, correct, and matches the originating issue's suggested fix (`reset-failed` then `start/restart`, preferring the unit where it exists).

## Verification

- **Shell sequencing is correct.** The two statements in the `then` branch are `;`-separated (`src/remediations.ts:65-66`), so a non-zero `reset-failed` (nothing to clear) does NOT block the following `restart` — `;` discards the exit status, unlike `&&`. This matches the doc-comment's stated intent (`src/remediations.ts:48-49`).
- **Branch exit status is irrelevant to the result.** After the `{ if … fi; }` block, the poll loop at `src/remediations.ts:70-71` decides `exit 0/1` based on whether the daemon actually answers, so the `then` branch's own status can't produce a false success.
- **Referenced symbols exist and are consistent.** The comment cites `provision`'s `HERDR_ADOPT_SOCKET` and `StartLimitIntervalSec=0`; both are real — `deploy/provision.ts:135-141` (which itself ends in `reset-failed herdr … ; exit 0`) and `deploy/herdr.service:11`. So the claim that the current provisioned unit never parks in `failed` from the limiter (making `reset-failed` a no-op there) holds.
- **Test genuinely exercises the failing-reset path.** The systemctl stub (`test/remediations.test.ts:99-105`) returns 0 only for `cat`/`restart` and `exit 1`s on the unmatched `reset-failed` arg. So `expect(r.status).toBe(0)` (`:109`) proves the non-zero `reset-failed` did not abort the sequence, and the `indexOf` ordering assertions (`:118-119`) prove `reset-failed` runs before `restart` — a `;`→`&&` regression would flip status to 1 and never reach `restart`. Verified against `test/remediations.test.ts:88-124`.

- The originating issue's "related hazard" — `defaultRunRemediation` in `src/diagnostics.ts` SIGKILLing the child's whole process group on its 120s timeout while `nohup … &` stays in the parent's process group — lives in `src/diagnostics.ts`, which this PR does not touch. It remains unreachable today (HERDR_SERVE self-terminates within ~10s) and is informational only.
- The issue's "consider also" about `deploy/provision.ts`'s `enable --now herdr` reporting success under `Type=simple` even when `herdr server` immediately exits is addressed by the pre-existing `HERDR_ADOPT_SOCKET` (`deploy/provision.ts:135`), also outside this diff.

PR #1603, opened 2026-07-10
fix(herdr): update+handoff the running daemon on `herdr_outdated` (#1578)

scoop:
## Problem (#1578)

`diagnostics_hint_herdr_outdated` reinstalled the herdr binary but left the **running** daemon on the old version. The `outdated` probe reads `herdr --version` (the on-disk binary), so a reinstall cleared the *warning* while the live daemon kept serving the old version — a degraded split state.

## Fix

Route the remediation through herdr's atomic `herdr update --handoff`, reused verbatim from `buildUpdateScript`:

- A successful update updates the binary **and** hands the live targets to the new server as one herdr operation, so no old daemon can survive (unlike a decoupled install + best-effort handoff).
- Its 3× reachability retry + detached `setsid server` relaunch recovery keeps a failed handoff from stranding the host **offline**.
- Verification is layered, because the running-server version is exposed only over the herdr socket (`ping`), not any CLI verb the shell can read:
  - `DiagnosticsService.fix()` **re-probes** after running the remediation — a no-op update leaves the binary old (still `warning`, no false `ok`); a handoff that killed the server reads `offline`.
  - the residual "binary bumped but old daemon silently kept alive" case is precluded by herdr's update+handoff atomicity (documented).

Stays auto-fixable in-app (not guidance-only).

### Why not the first-cut `HERDR_INSTALL && herdr server live-handoff`

That decoupled design was reverted: its reachability poll is answered by the OLD daemon both before and after (false success), and `live-handoff` was best-effort. `update --handoff` is atomic and reuses proven install+handoff+recovery machinery.

## Tests

- `test/remediations.test.ts` — wiring (`entry === buildUpdateScript(config.herdrUpdateLogPath)`, contains `update --handoff` + `setsid`, not `live-handoff`, auto-fixable) + the **first behavioral run** of the recovery path (herdr-update.test.ts only asserts string shape).
- `test/diagnostics.test.ts` — the authoritative false-success guard: `fix("herdr")` with a no-op update → still `warning`; with a real bump → `ok`.
- `ci/onboarding-harness/scenarios.ts` — detection-only `herdr-outdated` scenario (live stub `< 0.7.0` → `warning`). A green-able E2E stays deferred: `install.sh` can't pin an old build, so the harness can't stand up a real old live daemon to hand off from (stated in the header NOTE). This scenario exercises none of the remediation and largely duplicates the unit coverage — its value is the real boot+probe classification.

Server + harness + tests only; no `ui/**` changes. `bun run lint` clean; `bun test ./test` green (6350 pass).

### DIFF
--- a/ci/onboarding-harness/scenarios.ts
+++ b/ci/onboarding-harness/scenarios.ts
@@ -22,10 +22,14 @@ import type { Scenario } from "./types";
  *    reported as DETECTION-ONLY. This is the honest "onboarding still needs the
  *    user here" finding the gap report exists to surface.
  *
- * NOTE: scenarios that require provisioning the harness doesn't yet implement
- * (a pinned outdated herdr build; a faked tailnet for the not-serving warning
- * state) are intentionally OMITTED rather than shipped as permanent gaps — see
- * the deferred follow-ups in docs/superpowers/specs/...-design.md.
+ * NOTE: the `herdr-outdated` scenario below covers DETECTION of a live-but-old herdr
+ * (warning), but its GREEN-ABLE remediation — `herdr update --handoff` (#1578) — stays
+ * deferred: install.sh can't pin an old build, so the harness can't stand up a real old
+ * live daemon to hand off from (the remediation's behavior is proved by
+ * test/remediations.test.ts + test/diagnostics.test.ts instead). Other scenarios that
+ * require provisioning the harness doesn't yet implement (a faked tailnet for the
+ * not-serving warning state) remain intentionally OMITTED rather than shipped as permanent
+ * gaps — see the deferred follow-ups in docs/superpowers/specs/...-design.md.
  */
 export const SCENARIOS: Scenario[] = [
   {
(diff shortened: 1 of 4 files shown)


scoop (commented):
## Summary

The PR fixes `herdr_outdated` so it updates the binary **and** hands off the running daemon, instead of the pre-#1578 install-only behavior that left an old daemon serving. It routes the remediation through the existing `buildUpdateScript` (`herdr update --handoff` + reachability retry + detached `setsid` relaunch recovery). This is option #1 from the issue, and it's the right call — `herdr update` self-installs, so the issue's suggested `HERDR_INSTALL && …` prefix would have been redundant.

- **Module-init call is safe.** `buildUpdateScript(config.herdrUpdateLogPath)` is evaluated at import time in the `REMEDIATIONS` literal. `buildUpdateScript`'s signature (`src/herdr-update.ts:56-61`) makes `from`/`to`/`herdrBin` optional, and both `config.herdrUpdateLogPath` (`src/config.ts:23-24,456`) and `config.herdrBin` (`src/config.ts:455`) resolve statically at load — no undefined/throw risk. No import cycle (herdr-update does not import remediations).
- **Shell portability.** The in-app path runs the script via `sh -c` in `defaultRunRemediation` (`src/diagnostics.ts:143`), whereas `HerdrUpdateService` uses `bash -lc` (`src/herdr-update.ts:220`). The generated script is POSIX-sh clean (brace group, `for … in`, `[ ]`, `$(…)`, `rc=$?`; external `setsid`/`timeout`/`tee`/`date` are Linux-host tools) — verified against `src/herdr-update.ts:105-130`. No bashisms.
- **PATH.** The script uses bare `herdr` with no `export PATH`, unlike `HERDR_SERVE`. This is fine in-app: `fix()` re-probes (`src/diagnostics.ts:570-571`) and the version probe already spawns `config.herdrBin` from shepherd's inherited env; if `herdr` were unresolvable there, the check would read `missing` (error), not `outdated`, so this remediation would never be selected. The harness `herdr-outdated` scenario is `detectionOnly` (`ci/onboarding-harness/scenarios.ts`), so it never runs the command in a non-login incus shell.
- **Verification-by-re-probe holds.** Below-`HERDR_MIN_VERSION` (`0.7.0`, `src/config.ts:73`) + live ⇒ `warning`/`diagnostics_hint_herdr_outdated`; at/above floor ⇒ `ok`/`diagnostics_hint_herdr_ok` (`src/diagnostics.ts:301-322`). Both new `DiagnosticsService.fix` tests (no-op ⇒ still warning; success ⇒ ok) are consistent with this.
- **Harness scenario.** Stub `{"version":"0.6.5"}` → `SEMVER_RE` captures `0.6.5` < `0.7.0` → warning; `agent list` exits 0 → live (not offline) → matches `expect: [{ id: "herdr", state: "warning" }]`. `detectionOnly` / `coaching: "prose"` are valid Scenario fields (`ci/onboarding-harness/types.ts:18,25`).
- **Tests.** New tests reuse existing helpers verified to exist and match usage: `makeSandbox`/`runInSandbox`/`writeStub` (`test/remediations.test.ts:15-52`), `versionRunner`/`healthyDeps`/`byId` (`test/diagnostics.test.ts:19-62`). `HERDR_INSTALL` remains exported and still backs `herdr_missing`.

No correctness, security, or quality defects found. Nothing blocking.

### Observations (non-blocking, no action required)

- Because the diagnostics layer can't know the target version, `buildUpdateScript` is called without `from`/`to`, so the audit-log header records `unknown -> unknown` (`src/herdr-update.ts:109`). This is inherent to the diagnostics trigger (the running-server version is only exposed over the herdr socket, as the comment notes) and is purely cosmetic for the post-mortem log — not worth changing.

PR #1605, opened 2026-07-10
fix(onboarding): retry bun install on transient node-pty tarball flake (#1602)

### LINKED ISSUE #1602: Onboarding harness: transient node-pty tarball extraction flake in install.sh (installE2E) has no retry
Follow-up from #1577 / #1599.

## Context

#1599 added a bounded retry to `applyVerbatim` so a transient network/mirror flake in a **verbatim remediation** can't red the release gate. That retry deliberately does **not** cover the `install.sh` path.

## Symptom

Separately observed during #1577: the `install-e2e-service` scenario failed once on

```
error: Fail extracting tarball for "node-pty"
```

and passed on retry. This flake occurs inside `deploy/install.sh` (run by the `installE2E` / `serviceLifecycle` scenarios in `ci/onboarding-harness/run.ts` — `runInstallE2E` / `runInstallLifecycleE2E`), not in `applyVerbatim`, so #1599's retry does nothing for it. A single transient `node-pty` (or other dependency) extraction failure during `bun install` in `install.sh` can still red the gate with no retry.

## Suggested fix

Add a bounded retry around the transient-prone network step(s):
- either in `deploy/install.sh` around the `bun install` / dependency-fetch step, or
- in the harness runners (`runInstallE2E` / `runInstallLifecycleE2E`) around the `install.sh` invocation.

Keep it fail-closed: retry only on a genuine transient extraction/download failure, and still gate if it persists after the bounded attempts.

## Scope note

This was explicitly left out of #1599's scope (the `applyVerbatim` retry) and called out in that PR's body so it wasn't mistaken for fixed.

scoop:
Closes #1602. Follow-up from #1577 / #1599.

## Problem

A transient `error: Fail extracting tarball for "node-pty"` during `bun install` passes on re-run, but had no retry — so a single flake redded the onboarding release gate (`installE2E` / `serviceLifecycle` scenarios) and could fail real cold-start installs. #1599's `applyVerbatim` retry deliberately excluded this path.

**Key detail:** the flaky `bun install` is **not** in `install.sh` itself — `install.sh` hands off to the two paths that actually run it:
- `provision.ts` → `buildOnly()` — the **installE2E** / no-service path
- `deploy/update.sh` — the **serviceLifecycle** path (via `installService`)

## Fix

Bounded, fail-closed retry around **only the network-bound `bun install` step** in both paths (any non-zero exit retries, mirroring #1599's precedent):

- **`deploy/provision.ts`** — new `runWithRetry` (+ synchronous `sleepSync` backoff) wraps the two `buildOnly` bun installs. 2 attempts, 3s backoff. A deterministic failure re-fails each attempt and still gates.
- **`deploy/update.sh`** — a bash `retry()` helper wraps both bun installs (`retry 2 bun install`, `(cd ui && retry 2 bun install)`).
- `fix-node-pty-perms` and the UI build stay un-retried (local + deterministic).

## Tests

- `test/provision.test.ts`: retries-then-succeeds, persists-past-bound-still-gates, non-install-step-not-retried.
- `test/update-sh-retry.test.ts`: source assertions — helper present, both installs wrapped, no bare `bun install`.

## Verification

lint ✓ · typecheck ✓ · targeted tests 57 pass / 0 fail · branch-hygiene ✓ · feature-catalog ✓ (fix, not feat).

Not a user-facing feature → no feature-announcement entry; no user-facing strings → no i18n. `[no-feature-entry]`

The full suite shows 8 pre-existing environmental flakes (session/herdr tests needing a live daemon the fresh worktree lacks — 10 fail on the clean base); none touch `provision.ts`/`update.sh`.

### DIFF
--- a/deploy/provision.ts
+++ b/deploy/provision.ts
@@ -232,6 +232,41 @@ const defaultRunner: Runner = (cmd, args, opts) => {
   }
 };
 
+/** Synchronous sleep — provision runs a sequential, synchronous install flow, so the
+ *  between-attempt backoff in {@link runWithRetry} can't be a Promise. `ms <= 0` is a
+ *  no-op (tests pass 0 to stay instant). */
+function sleepSync(ms: number): void {
+  if (ms <= 0) return;
+  Atomics.wait(new Int32Array(new SharedArrayBuffer(4)), 0, 0, ms);
+}
+
+/** Bounded, fail-closed retry around a single `run` step. Bun's dependency fetch/extract
+ *  can hit a transient registry/mirror flake — observed as `error: Fail extracting tarball
+ *  for "node-pty"` (#1602) — that succeeds on a re-run. Wrapping the `bun install` steps
+ *  lets one such flake not fail a cold-start install (nor red the onboarding release gate,
+ *  whose installE2E/serviceLifecycle scenarios drive this path). Any non-zero exit retries
+ *  up to `attempts`; a deterministic failure just re-fails each attempt and the last error
+ *  propagates — so a real regression still gates. Mirrors the apply-time retry in #1599. */
+function runWithRetry(
+  run: Runner,
+  cmd: string,
+  args: string[],
+  opts: { env?: NodeJS.ProcessEnv } | undefined,
+  attempts: number,
+  delayMs: number,
+): void {
+  for (let attempt = 1; ; attempt++) {
+    try {
+      run(cmd, args, opts);
+      return;
+    } catch (err) {
+      if (attempt >= attempts) throw err;
+      log(`  step failed (attempt ${attempt}/${attempts}) — retrying in ${delayMs}ms`);
+      sleepSync(delayMs);
+    }
+  }
+}
+
 /** Injectable file IO. Production reads/writes the real filesystem; tests inject a
  *  recorder so the templated unit can be asserted without touching disk. */
 export interface FileIO {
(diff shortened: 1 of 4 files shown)


scoop (commented):
## Summary

The PR adds a bounded, fail-closed retry around the transient-prone `bun install` step on **both** deploy paths the onboarding harness drives, which is exactly the gap #1602 describes. It looks correct and is well tested.

## Verification

**Both flake-prone paths are covered.** The issue's flake occurs after `install.sh` hands off to `provision.ts` (`deploy/install.sh:269` — `exec bun run deploy/provision.ts`), which splits by service decision (`deploy/provision.ts:553-558`):
- `installE2E` / `runInstallE2E` passes `SHEPHERD_NO_SERVICE=1` (`ci/onboarding-harness/run.ts:80`) → `buildOnly` (no-service path). Both `bun install` steps there are now wrapped in `runWithRetry` (`deploy/provision.ts:510-517`, `522-529`); the local `fix-node-pty-perms` and `bun run build` steps are correctly left un-retried.
- `serviceLifecycle` / `runInstallLifecycleE2E` passes no `NO_SERVICE` (`ci/onboarding-harness/run.ts:132`) → `installService`, which delegates deps+build to `update.sh` (`deploy/provision.ts:490`). Both `bun install` invocations in `update.sh` are now `retry 2 bun install` / `(cd ui && retry 2 bun install)` (`deploy/update.sh:60,65`).
- I grepped all `bun install` sites under `deploy/` and `scripts/`: the only two remaining deploy-path installs (`provision.ts:513,525`, `update.sh:60,65`) are all wrapped. No bare install remains on a gated path.

**Fail-closed semantics confirmed.**
- `runWithRetry` (`deploy/provision.ts:250-263`): loops `attempt` from 1; on failure re-throws once `attempt >= attempts`, so with the default `attempts:2` a persistent failure re-fails and the last error propagates. Matches the `#1602` propagation test (`test/provision.test.ts` — `attempts` reaches 2 then throws `/node-pty/`).
- `retry()` in bash (`deploy/update.sh:26-40`): returns 1 after the bounded loop; under the file's `set -euo pipefail` (`deploy/update.sh:6`) a final non-zero from `retry`/the `(cd ui && …)` subshell aborts the deploy. The intermediate `"$@" && return 0` and `[ "$i" -lt "$attempts" ] && { … }` are AND-OR lists, so a non-final failure inside them does not trip `set -e` prematurely — the loop runs its full bound. The escaped backticks in the warn message are literal (inside double quotes), no command substitution.

**Tests are sound.** `test/provision.test.ts` exercises retry-then-succeed, persist-past-bound, and non-install-step-not-retried (build fails once, `buildAttempts===1`); `delayMs:0` keeps them instant via the `sleepSync` `ms<=0` no-op guard (`deploy/provision.ts:234-237`). The existing 3-arg `buildOnly("/repo", run, {...})` call (`test/provision.test.ts:445`) still type-checks since the new `retry` param is optional/defaulted. `test/update-sh-retry.test.ts`'s `bareInstall` regex (`/^\s*bun install\b/`) correctly passes given no line now begins with `bun install` (fix-perms is `bun scripts`, build is `bun run`).

## Stated limitation (not a finding)

- `sleepSync` uses `Atomics.wait` on the main thread (`deploy/provision.ts:236`). This is only reached in production (`delayMs:3000`), never in tests. I could not verify Bun's main-thread `Atomics.wait` support against anything in this repo; Node.js permits it and the surrounding flow is explicitly synchronous, so the approach is reasonable as written — noting only that the guarantee is external to this repo.

PR #1606, opened 2026-07-10
feat: make the manual plan re-review actually re-review (force seam)

scoop:
## Problem

The operator's **Re-review plan** control (stalled plan-gate menu, GitRail, PlanPanel) was a guaranteed no-op. `PlanGateService.consider()` dedupes on `sha256(.shepherd-plan.md)`, and at the adversarial-review round cap the planning agent is never re-steered — so the on-disk plan is byte-identical to the one already reviewed. Every click round-tripped `POST /review-plan`, got `"skipped"`, and produced a toast saying the review didn't start. The control was dead by construction.

## Fix

Give the plan gate the **`force` seam the PR critic already has** (`ReviewService.consider(…, { force })`). `POST /review-plan` is operator-initiated, so it now passes `{ force: true }`, which bypasses **only** the plan-hash dedupe. Every hard precondition still gates.

`approved` stays a **hard precondition** — `force` does *not* bypass it. `gate.approved` is a release latch on a one-way transition, and `releasePlanGate` / `#enterExecution` / `finalize()` rely on the invariant `approved ⇒ no reviewer in flight`. That invariant is maintained by two sites: `consider()`'s approved guard **and** `adoptOrphans()` reaping (not re-adopting) an approved orphan. So the one state where a manual review is offered but genuinely cannot start is an **approved** gate — and GitRail + PlanPanel render that control inert (`aria-disabled`, keyboard-reachable, announced reason) instead of firing a dead click.

### Behaviour details
- **F3** — a forced re-review of an *unchanged* plan holds the round, holds `finalRoundPending`, delivers no steer, and writes no stall signal (re-injecting identical findings is noise). A forced review of a *changed* plan advances normally.
- **Stall signals bounded** — a forced at-cap *re-entry* writes no `stall` row (a human just clicked; the signal means "needs a human"), but a forced review that *crosses* the cap still writes exactly one. Prevents repeat clicks from walking a session toward the learnings-distiller threshold.
- **Answer preservation** — `answeredQuestionKeys` now carry forward from the **live** gate at finalize time, so a forced re-review of an unchanged plan no longer wipes the operator's recorded question-form answers (including answers merged mid-review).
- **Honest copy** — deleted `PlanPanel.skippedOutcome()` (which guessed a cause from the cached gate) and neutralized three strings that asserted "the plan is unchanged" — false once `force` re-reviews it. The skipped toasts remain as the fail-safe for the render→click race, with causally-silent copy.

### Deferred (follow-ups, intentionally not in this PR)
- `releasePlanGate` `isReviewing` guard + a `planPhase` re-check in `finalize()` — only reachable *if* a future change lets `force` bypass `approved`; unreachable today, so writing them now would be dead code. A test pins `consider(session, { force: true })` declining an approved gate as the guard.
- A plugin `onSpawn` abort returns `"skipped"` deterministically; giving it its own trigger value (rather than sharing `"skipped"`) would let the UI explain a plugin-refusing repo. Out of scope here.

## Regression protection (accurate)

The route→adapter→service path carrying `{ force: true }` is protected by the **forwarding adapter** in `index.ts` (`(s, opts) => planGate.consider(s, opts)`) plus a **route test** asserting `consider` receives `{ force: true }`. Note: typing the dep `PlanGateService["consider"]` is a DRY win but is **not** a compile-time guard against a dropped param — TypeScript allows a fewer-parameter function to satisfy a wider function type. The runtime test is the real guard.

## Tests
- `test/plan-gate-service.test.ts` (+16 cases): force bypasses the dedupe only; does *not* bypass `approved`/`planPhase`/in-flight/empty-plan; unforced auto-path bit-identical; answer carry-forward (incl. mid-review merge) vs reset; F3 sub-cap hold; at-cap re-entry vs crossing vs dead-pane signal arms; `adoptOrphans` reaps an approved orphan without re-adopting.
- `test/server-plan-gate.test.ts`: route forwards `{ force: true }`.
- UI: `canTriggerPlanReview` unit test; PlanPanel/PlanGateBadge browser tests (approved control is `aria-disabled` + no POST; a `changes_requested` click starts a real review; the menu issues the POST and surfaces `plangate_review_started`).

Server suite 108/108, UI suite 3135/3135, `check` + `check:i18n` (EN/DE parity) green.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2400,7 +2400,7 @@ const appDeps: AppDeps = {
     reviewing: () => planGate.reviewingIds(),
   },
   planGate: {
-    consider: (s) => planGate.consider(s),
+    consider: (s, opts) => planGate.consider(s, opts),
     resume: (s) => planGate.resume(s),
     dismiss: (s) => planGate.dismiss(s),
   },
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -22,10 +22,11 @@ import { effectiveAutopilot } from "./effective-autopilot";
 import { resolveAuxSpawn, type MembraneSeams } from "./spawn-membrane";
 import { fenceUntrusted } from "./untrusted";
 
-/** Outcome of an on-demand `consider()`: a reviewer actually spawned, the request was a no-op
- *  (plan unchanged / already approved / nothing to review), the plan artifact is unavailable, or
- *  a spawn attempt failed. The review-plan route relays this so the UI can distinguish a silent
- *  dedupe from a real error or an unusable `.shepherd-plan.md`. */
+/** Outcome of an on-demand `consider()`: a reviewer actually spawned (`"started"`); the request
+ *  was a no-op (`"skipped"` — not planning, a review already in flight/starting, a tombstone or
+ *  plugin-abort during spawn, or an unchanged plan on the auto-path); the plan artifact is
+ *  unavailable (`"plan-unavailable"`); or a spawn attempt failed (`"error"`). The review-plan route
+ *  relays this so the UI can distinguish a no-op from a real error or an unusable `.shepherd-plan.md`. */
 export type PlanReviewTrigger = "started" | "skipped" | "plan-unavailable" | "error";
 
 /** The plan the planning agent writes in its LIVE session worktree; the reviewer reads its text. */
@@ -187,6 +188,7 @@ interface PlanInFlight {
   reviewerModel: string | null;
   reviewerEffort: string | null;
   priorRound: number; // adversarial rounds already spent on this plan streak
+  forced: boolean; // this run is an operator's manual re-review (consider's opts.force) — governs the F3 hold + stall-signal guards
   startedAt: number;
   finalizing?: boolean;
 }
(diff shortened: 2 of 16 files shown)


scoop (commented):
## Summary

The PR satisfies the task: it stops "offering" a re-review that would only produce a "didn't start" toast. It does so on two fronts:

1. **Unchanged plans now actually re-review.** `consider(session, {force})` bypasses only the unchanged-plan hash dedupe (`src/plan-gate.ts:95`), so an operator's manual click on a byte-identical plan spawns a real reviewer instead of silently no-opping. `POST /review-plan` passes `force: true` (`src/server.ts:319`), mirroring `/review-pr`'s force. So the former "unchanged → skipped → toast" path is gone for the manual route.
2. **The genuinely un-startable case (approved) is no longer offered.** `canTriggerPlanReview()` (`ui/src/lib/components/plan-gate-badge.ts:1165`) returns `"approved"`/`"reviewing"`/null, and both PlanPanel (`ui/src/lib/components/PlanPanel.svelte:442-449,468`) and GitRail/RailStatusActions render the control inert (aria-disabled + guarded onclick early-return) with a persistent "already approved" reason rather than firing a request that dead no-ops.

## Verification performed

- **i18n parity.** Only two catalogs exist (`ui/messages/en.json`, `ui/messages/de.json`). New key `gitrail_review_plan_approved` present in both (verified en.json:762, de.json:726). Reworded keys `plangate_review_skipped_stalled`, `planpanel_review_nothing_to_review`, `gitrail_review_plan_skipped` changed in both. Removed keys `planpanel_review_unchanged` and `planpanel_review_changes_pending` deleted from both, and grep across the tree found **zero** remaining references to either (nor to the removed `skippedOutcome` helper). `planpanel_review_already_approved` retained in both (used by PlanPanel:444,469 and its label).
- **Dead code removed cleanly.** `outcome` state narrowed to `"skipped"|"error"|null`; grep of PlanPanel.svelte (lines 120,134,155,173,182-186,395,397) shows no residual `"approved"/"changes"/"idle"` branches, and `skippedOutcome()` is gone with no callers.
- **Signature/contract change.** `AppDeps.planGate.consider` retyped to `PlanGateService["consider"]` (`src/server.ts:291`); the only wiring (`src/index.ts:10`) forwards `(s, opts)`, and the server-plan-gate test harness matches the 2-arg shape. `consider`'s new optional `opts` is backward-compatible with the auto-path (`unforced` regression test at test/plan-gate-service.test.ts confirms bit-identical behavior).
- **force does not lift hard preconditions.** Verified `src/plan-gate.ts:74-78`: non-planning phase, in-flight/starting, and `prior?.approved` all still short-circuit to `"skipped"` before the force branch; empty plan still returns `"plan-unavailable"`. Tests cover each.
- **adoptOrphans reap path.** `reapOrphanSpawn` (`src/plan-gate.ts:497`) consumes `{worktreePath, reviewerSessionId}`, both present on the spawn row used at the call site (`src/plan-gate.ts:461`, fields also read at 468/470). It reuses the same deps finalize() uses (`resolveTerminal`:540, `readUsage`:213, `completeReviewerSpawn` in store Pick:122, `worktree.remove`:126) — best-effort, no throw path, upholding the `approved ⇒ no reviewer in flight` invariant.
- **F3 no-loop / no-double-signal.** The forced-unchanged branch (`src/plan-gate.ts:623`) holds `round`/`finalRoundPending` and emits no signal; the at-cap guard suppresses only forced re-entry (`priorRound >= cap`) while still allowing a single forced crossing; the sub-cap-undelivered guard suppresses the learnings signal only when forced. All four combinations are covered by tests.
- **GitRail armed-latch $effect** (`GitRail.svelte:794`) writes `armed = null` guarded by `armed === "review-plan"`, which self-clears the condition — no reactive loop.

No correctness, security, or quality defects found within the diff.

scoop (commented):
## Summary

The PR resolves the task well. Instead of hiding the re-review control for an unchanged plan, it makes the manual click actually *do* something: the `/review-plan` route now passes `{ force: true }` (server.ts:370), and `consider(..., opts.force)` bypasses only the unchanged-plan hash dedupe (plan-gate.ts:95) while preserving every hard precondition (not-planning, in-flight/starting, `approved`, missing plan). The one state that genuinely can't start — `approved` — no longer dead-no-ops: the GitRail and PlanPanel controls render inert with a reason (`canTriggerPlanReview`, plan-gate-badge.ts:1248). This removes the "click → toast that nothing happened" dead end the task called out.

The new streak semantics are carefully handled: a forced re-review of an *unchanged* plan holds `round`/`finalRoundPending`, delivers no steer, and writes no learnings signal (F3, plan-gate.ts:656), while a forced *changed* plan falls through to the normal path. `answeredQuestionKeys` are now carried forward across a same-`planHash` re-run (plan-gate.ts:787), which is newly necessary because `force` makes buildGate run more than once per planHash.

- **i18n parity (en+de):** reworded keys present in both — `plangate_review_skipped_stalled` (en.json:620, de.json:622), `planpanel_review_nothing_to_review` (en.json:622, de.json:624), `gitrail_review_plan_skipped` (en.json:1595, de.json:1595). New keys in both — `gitrail_review_plan_approved` (en.json:2783, de.json:2783) and `feat_manual_plan_rereview_title`/`_body` (en.json:2272-2273, de.json:2272-2273). Removed `planpanel_review_unchanged` and `planpanel_review_changes_pending` are gone from both catalogs and no longer referenced in any `.svelte` (grep clean); retained `planpanel_review_already_approved` still exists in de.json (:763).
- **Dead-reference check:** removed helper `skippedOutcome` and old `outcome` union members (`"approved"|"changes"|"idle"`) have no remaining references (grep clean); the template only branches on `"skipped"|"error"` now (PlanPanel.svelte).
- **New export wired:** `canTriggerPlanReview` imported by both GitRail.svelte:857 and PlanPanel.svelte:1009; `planGates.map`/`isReviewing` exist (reviews.svelte.ts:130,204).
- **Feature announcement:** entries are auto-globbed (feature-announcements.ts:47), so the new file self-registers; `sinceVersion:"1.43.0"` is correct per the in-file rule to use next-version (12 sibling v1.43.0 entries already exist), not package.json's 1.42.0.
- **`reapOrphanSpawn` (plan-gate.ts:528-535)** mirrors finalize()'s cleanup (herdr.stop + best-effort readUsage/completeReviewerSpawn + worktree.remove; finalize at :609-617), upholding `approved ⇒ no reviewer in flight`.
- **Guard tests match code:** F3 hold, forced at-cap re-entry suppression, forced crossing one-shot, and forced dead-pane suppression in `escalateStallIfNeeded` (plan-gate.ts:690-721) all line up with the added tests.

## Findings

**plan-gate.ts:724 — `applyError` is not `forced`-guarded, unlike the changes_requested path (non-blocking).** The PR's stated invariant (comments at plan-gate.ts:686, :511) is that a repeat-clickable forced re-review must not walk the distiller's learnings-signal threshold. `escalateStallIfNeeded` enforces this for `changes_requested`, and F3 short-circuits the unchanged case — but `applyError` unconditionally `addSignal`s (plan-gate.ts:727) with no `f.forced` check. Before this PR an unchanged `changes_requested` plan was deduped and unreachable; `force` now makes it re-reviewable, so a forced re-review that lands on an `error` verdict (timeout / unparseable) adds a `"did not produce a verdict — needs a human"` signal on every click, unbounded. It's throttled by the in-flight `disabled` state and error verdicts are uncommon, so low severity — but it's the one outcome branch the distiller-spam protection doesn't cover. Consider suppressing the signal (or de-duping it) on a forced run, mirroring the changes_requested guards.

PR #1607, opened 2026-07-10
test(backup): raise live-db hook/test timeouts to de-flake CI I/O stalls

### LINKED ISSUE #1598: Flaky: backup.test.ts 'rotate deletes only the GFS losers' hook times out on CI under I/O pressure
## Summary
`test/backup.test.ts › snapshot + integrity + rotate (live db) › rotate deletes only the GFS losers` intermittently fails on CI with:
```
(fail) rotate deletes only the GFS losers, keeping shepherd-*.db.gz survivors
  ^ a beforeEach/afterEach hook timed out for this test.
```
Observed on PR #1575's `verify` job. It is **not** a code regression on that branch — `test/backup.test.ts` and `scripts/backup.ts` are byte-identical to `main`, and it is unrelated to that PR's diff. Confirmed a flake two ways: a rerun of the exact failing commit passed, and the next commit passed on a fresh run.

## Why it's the timeout, not the assertion
The hooks do real filesystem + SQLite work against Bun's **default 5000ms** hook timeout:
- `beforeEach`: opens a live DB, `CREATE TABLE`, **500 inserts** (no transaction)
- the test: plants **200 files** + rotates
- `afterEach`: `live.close()` + `rmSync(dir, { recursive: true })`

Measured locally: beforeEach ≈ **6.2ms**, plant 200 files ≈ **0.7ms**, afterEach ≈ **0.4ms** — ~7ms against a 5000ms budget (~700× margin). So nothing is near the limit under normal conditions; a >5s hook means the **shared CI runner stalled on I/O**, not that the code is slow. Did not reproduce locally across repeated runs or under 32-core CPU contention (the work is I/O-bound, not CPU-bound).

## Impact
Spurious `verify` failures on unrelated PRs — the critic won't review a red PR, so each occurrence costs a rerun + a manual "is this mine?" triage.

## Proposed fix (own change, not on #1575)
Raise the per-hook timeout for this describe block so a slow runner has headroom, e.g.:
```ts
beforeEach(() => { ... }, 20_000);
afterEach(() => { ... }, 20_000);
```
Optionally wrap the 500 inserts in a single transaction to shrink the baseline. The fix belongs in `test/backup.test.ts`, not in product code.

Surfaced by CI on PR #1575 (#1567).

scoop:
## Problem

`test/backup.test.ts › snapshot + integrity + rotate (live db) › rotate deletes only the GFS losers` intermittently fails on the `verify` job with:

```
^ a beforeEach/afterEach hook timed out for this test.
```

It is **not** a code regression — `test/backup.test.ts` and `scripts/backup.ts` were byte-identical to `main` when it surfaced (on PR #1575), and a rerun of the exact commit passed. The block does real FS + SQLite I/O that runs in ~7ms locally against `bun:test`'s **5000ms** default (~700× margin), so a >5s hook means the shared CI runner **stalled on I/O**, not that the code is slow.

## Fix (test-only; no product code)

- **Raise the per-hook / per-test budget to `20_000ms`** across the live-db block's `beforeEach`, `afterEach`, and its four `it()` cases, via a single `IO_TIMEOUT_MS` const. This is the **sole load-bearing change** — it covers every stall site (hook open/close + recursive `rmSync`, plant-200-files, snapshot-live-db). 20s stays a hard ceiling, so a genuine (orders-of-magnitude) regression would still fail.
- **Wrap the 500 `beforeEach` inserts in one `live.transaction(...)`**, collapsing ~500 per-insert autocommit fsyncs into one commit. Same rows/values → no assertion affected. This only shrinks the `beforeEach` baseline; it is a cheap tidy, **not** de-flake insurance for the other stall sites.

## Why this actually works (mechanism verified)

This repo had no prior hook-timeout / `setDefaultTimeout` usage, and CI runs `bun-version: latest`, so I confirmed empirically on Bun 1.3.14 that the trailing timeout arg **is** honored on `beforeEach`:

- a hook sleeping 5.5s (past the 5000ms default) **passed** with `20_000`;
- the identical hook with **no** arg **failed** with the issue's exact string.

So the fix is not a silent no-op on hooks. Local `bun test` was already green (~700× margin) and can't validate a de-flake — real validation is that mechanism check plus CI stability going forward.

Closes #1598

### DIFF
--- a/test/backup.test.ts
+++ b/test/backup.test.ts
@@ -17,6 +17,11 @@ import {
 const name = (d: Date) => `shepherd-${formatTs(d)}.db.gz`;
 const hoursAgo = (base: Date, h: number) => new Date(base.getTime() - h * 3600_000);
 
+// The live-db block does real FS + SQLite I/O (~7ms locally). Under CI I/O pressure a
+// shared runner can stall past bun:test's 5000ms default, timing out a hook/test that is
+// not actually slow. Give these headroom without hiding a real (orders-of-magnitude) regression.
+const IO_TIMEOUT_MS = 20_000;
+
 describe("formatTs / parseTs round-trip", () => {
   it("formats UTC and parses back", () => {
     const d = new Date(Date.UTC(2026, 5, 25, 14, 30, 5));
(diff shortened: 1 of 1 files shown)


scoop (commented):
## Summary

Test-only change that addresses the flaky `backup.test.ts` hook timeout exactly as the originating issue proposes: it introduces `IO_TIMEOUT_MS = 20_000` and applies it as the trailing timeout arg to `beforeEach`, `afterEach`, and all four `it`s in the `snapshot + integrity + rotate (live db)` block, and it collapses the 500 seed inserts into a single `live.transaction(...)`. No product code is touched, consistent with the issue's guidance.

## Verification

- `live.transaction(fn)` is a valid `bun:sqlite` API returning a callable that wraps the body in one BEGIN/COMMIT; `Database` is imported from `bun:sqlite` (verified test/backup.test.ts:2). The wrapped inserts still populate table `t`, so downstream assertions (e.g. the `'after'` insert at test/backup.test.ts:131) remain valid.
- The trailing-numeric timeout is a supported signature for `bun:test`'s `it`/`beforeEach`/`afterEach`; imports verified at test/backup.test.ts:1. `IO_TIMEOUT_MS` applied consistently at test/backup.test.ts:117, 121, 133, 143 and the two other `it`s in the diff.
- The added `force: true` on `rmSync` (test/backup.test.ts:120) is a harmless robustness improvement over the prior `{ recursive: true }`.
- The rewrapped `it` bodies are byte-equivalent in behavior to the originals (only reformatting for the added third argument); no assertions changed.

The 20s headroom preserves the issue's stated intent of not masking an orders-of-magnitude regression (baseline ~7ms). Nothing to address.

PR #1608, opened 2026-07-10
feat(ui): reveal session task info on tap-and-hold

scoop:
## What & why

The session-title task-info popover (`.desig-pop`) was revealed purely by CSS — `.desig-wrap:hover` and `:focus-within`, **both of which fire on a touch tap**. So a single tap popped the popover *and* folded the header (one tap, three effects). This moves the reveal onto a deliberate **tap-and-hold**, driving visibility from component state (`hoverOpen` / `focusOpen` / `holdOpen` → `.desig-wrap.meta-open`), so a single tap keeps only its existing job (fold on compact / toggle git rail on desktop).

## Per surface

- **Touch:** a 500ms long-press (`longPress` action) toggles the popover; the closing re-tap is swallowed in `onTitleTap` (no fold). A single tap reveals nothing.
- **Mouse:** `pointerenter`/`pointerleave` (never `over`/`out`, since `.desig-pop` is a descendant), plus an out-of-flow `::after` bridge over the 4px dead band so the scrollable content stays reachable. A click no longer latches it open.
- **Focus:** pointer-origin detection by timestamp (`Date.now()`, 600ms recency) rather than a sticky flag — keyboard/AT focus reveals it, mouse/touch focus does not, and a long-press can't strand the state. `focusin`/`focusout` on the wrap so an AT read *into* the popover can't collapse it mid-read.
- **iOS:** `-webkit-touch-callout: none` on the three text triggers (a half-second press would otherwise raise the selection callout/magnifier). **Android:** `oncontextmenu` gated on the last `pointerType`, not the device `touch` prop, so a touchscreen laptop keeps its right-click menu.

Amends the (unreleased) `v1.43.0-session-tooltip` What's-New entry to describe both gestures and adds `viewport_meta_desc` (EN+DE) as the `aria-describedby` affordance hint.

## Review & tests

Executed via subagent-driven development against a six-round-reviewed design doc. Independent task review (spec ✅, quality approved) + whole-branch review (READY TO MERGE), both clean.

- 19 new browser cases (2 render `mobile: true`), each written to fail against a broken impl — un-latch moves the pointer away before asserting hidden; the focus case fires the full `pointerdown→pointerup→mousedown→focus` sequence; the stranded-flag case uses `vi.setSystemTime` (not fake timers, to avoid desyncing `longPress`'s real 500ms timer); the phone-binding case fires a real `pointerdown` (proven red-then-green against removing the binding).
- Gates green: `ui bun run check` (0 errors/0 warnings, no unused-selector), `ui bun test` (59/59 in the touched file; 3207 total), `check:i18n` (parity, 2778 keys), root `lint`.

## Not automatable (manual verification owed on device)

- iOS Safari: hold shows no selection callout/magnifier.
- Android Chrome: hold shows no native context menu.
- VoiceOver/TalkBack: launch details reachable **and** readable without vanishing mid-read.


scoop (commented):
## Summary

The PR moves the task-info popover reveal off pure CSS `:hover`/`:focus-within` (which a touch tap also triggers, so one tap both popped the popover and folded the header) onto component state — `hoverOpen` (non-touch pointerenter/leave), `focusOpen` (keyboard/AT focus with pointer-origin detection), and `holdOpen` (`longPress`). A single touch tap now only folds the header / toggles the git rail; tap-and-hold reveals. This satisfies the task.

## Verification performed

- **State-driven reveal replaces CSS gating** — the old `.desig-wrap:hover .desig-pop, .desig-wrap:focus-within .desig-pop` rule is gone, replaced by `.desig-wrap.meta-open .desig-pop` (`Viewport.svelte:3162`). Grepped the whole file (`:hover`/`:focus-within`/`desig-pop`) — no residual popover reveal keyed on `:hover`/`:focus-within`; the only remaining `.hovering` uses are the chip tint (`Viewport.svelte:3113,3129`) and the hover-bridge `::after` (`Viewport.svelte:3175`), both gated on the touch-safe `hoverOpen` flag, not `:hover`.
- **`longPress` action** — imported at `Viewport.svelte:428` from `./longpress`; confirmed `ui/src/lib/components/longpress.ts:16-68` fires `onTrigger` at 500ms, cancels on >10px move, and `preventDefault`s the trailing synthetic click only when `onTrigger` returns true — matching `openMeta()` returning `true` (`Viewport.svelte:1095-1098`). So a hold opens the popover with no `onTitleTap`.
- **Close path** — `onTitleTap` swallows a tap while `holdOpen` (`Viewport.svelte:955-959`), resetting `lastTitleTap` so the closing tap isn't misread as a double-tap-rename; the capture-phase dismissal `onDown` treats `insideTitle(target)` as non-dismissing (`Viewport.svelte:615-617`). `desigWrapEl` is bound on BOTH the phone `.ctx` branch (`Viewport.svelte:2410`) and desktop branch (`Viewport.svelte:2455`), so `insideTitle` resolves `.ctx-trigger` correctly.
- **Pointer-origin focus detection** — `POINTER_FOCUS_MS = 600 > longPress 500ms` (`Viewport.svelte:1035`); a pointerdown-then-focus within 600ms is treated as pointer-driven and suppressed, keyboard/AT focus (no recent pointerdown, or aged out) opens. `Date.now()` used deliberately for mockability. Logic in `onWrapFocusIn` (`Viewport.svelte:1080-1085`) is coherent.
- **i18n parity** — new key `viewport_meta_desc` present in both catalogs (`ui/messages/en.json:490`, `ui/messages/de.json:490`); `feat_session_tooltip_body` updated in both (`en.json:38`→new, `de.json` line 18 of diff). No other catalogs exist (`ui/messages/` holds only `en.json`, `de.json`).
- **`$props.id()`** — used for `metaDescId` (`Viewport.svelte:998`); the same API is already used in `GlossaryTerm.svelte:11`, `InfoTip.svelte:14`, etc., so the Svelte version supports it. The `id={metaDescId}` span appears in both the phone and desktop branches, but they're mutually exclusive `{#if mobile}/{:else}` arms, so no duplicate-id collision.
- **CSS var** — `--desig-pop-w` defined on `.desig-wrap` (`Viewport.svelte:3087`) and consumed by `.desig-pop` (a descendant, inherits) and the `::after` bridge; both use `left:0` + the same width, so the bridge aligns with the popover.
- **Reset coverage** — `resetMeta()` wired into `toggleFold()` (`Viewport.svelte:438`), the `unitId` effect (`Viewport.svelte:582-586`) and the `renaming` effect (`Viewport.svelte:593-597`); `renaming` (`:859`) and `unitId` (`:656`) both exist. Dismissal effect binds only while `metaVisible` and unbinds via its cleanup.

Walked the main gesture sequences (touch hold→open→tap-close, touch scroll inside popover, mouse hover with 4px bridge, keyboard focus/Escape, double-tap rename, rename-commit remount) against the handlers — all consistent, and the new browser tests exercise each. The `feature-announcements` entry change is a comment-only edit (`v1.43.0-session-tooltip.ts`), no functional impact.

## Stated limitations

- Real touch/AT behaviour (iOS Safari callout suppression, VoiceOver/TalkBack focus-into-popover) cannot be exercised in this static review; the code's reasoning and the Chromium browser tests are the available evidence.

PR #1609, opened 2026-07-10
fix(learnings): retype distiller/optimizer DI seams to => Promise<void> (#1597)

### LINKED ISSUE #1597: distiller/optimizer DI seams typed `=> void` while their impls are `async` (hides floating promises from tsc + lint gate)
## Summary
Three learnings-service DI seams in `AppDeps` (`src/server.ts`) are declared `=> void` but their real implementations are `async … : Promise<void>`:

| DI decl (`src/server.ts`) | impl |
|---|---|
| `distillNow: (repoPath: string) => void` | `DistillerService.distillNow` — `src/distiller.ts:197` `async … : Promise<void>` |
| `optimizeOne: (id: string) => void` | `OptimizerService.optimizeOne` — `src/optimizer.ts:129` `async … : Promise<void>` |
| `optimizeAllFlagged: (repoPath: string) => void` | `OptimizerService.optimizeAllFlagged` — `src/optimizer.ts:138` `async … : Promise<void>` |

Because the seam lies about the return type, both `tsc` and the `no-floating-promises` gate (landed in #1567) are blind to it. The call sites fire-and-forget these, so before #1567 they leaked unhandled rejections on paths that run today (the service chains `enqueueOrBegin → begin` never catch internally).

## Current state (mitigated, not fixed)
PR #1575 closed the **leak** at the three call sites without touching the type, using `Promise.resolve(...).catch(...)` (normalizes the declared-void/actually-promise value so the handler attaches at runtime). Each DI declaration carries a `TODO(#1567 follow-up)` marker. So there is **no active leak** — but the types still lie, so a future author reading the seam has no signal these are async.

## The fix this issue tracks
Retype the three seams to `=> Promise<void>` to match reality. This is deferred out of #1567 deliberately: correcting the type cascades to ~16 test doubles across ~9 files (`server.test.ts`, `server-learnings.test.ts`, `agent-ingress.test.ts`, `server-auth.test.ts`, `server-fork.test.ts`, `server-ping.test.ts`, `server-projects.test.ts`, `server-sync-fork.test.ts`, `automation-confirm.test.ts`) which must become async doubles. That's a mechanical but wide change, unrelated to the `send` port, and it violated the branch's one-feature rule to include it.

Once retyped, the `Promise.resolve(...)` wrappers at the call sites can collapse back to `deps.x?.foo().catch(...)`.

Follow-up to #1567 (PR #1575). Grep `TODO(#1567 follow-up)` for the exact anchor points.

scoop:
## What

Three `AppDeps` DI seams (`src/server.ts`) were typed `=> void` while their real impls are `async … : Promise<void>` — so both `tsc` and the `no-floating-promises` gate (#1567) were blind to them.

| Seam | Impl |
|---|---|
| `distillNow: (repoPath) => void` → `=> Promise<void>` | `DistillerService.distillNow` (async) |
| `optimizeOne: (id) => void` → `=> Promise<void>` | `OptimizerService.optimizeOne` (async) |
| `optimizeAllFlagged: (repoPath) => void` → `=> Promise<void>` | `OptimizerService.optimizeAllFlagged` (async) |

## Changes

- Retype the three seams to `=> Promise<void>`; drop the `TODO(#1567 follow-up)` mistyping comments.
- Collapse the three call-site `void Promise.resolve(...).catch(...)` mitigation wrappers (added in #1575) back to the plain `void deps.x?.foo(arg).catch(...)` pattern the adjacent `mergeSuggest` sibling already uses (optional chaining short-circuits `.catch` safely).
- Make the ~13 cascading sync test doubles `async` across 9 test files; remove two now-redundant `as unknown as (...) => void` casts on the rejecting-double tests.

## Why now

Deliberately deferred out of #1567 / PR #1575: correcting the type cascades to the test doubles (root `tsconfig.json` type-checks `test/`), a wide mechanical change unrelated to the `send` port. This PR is that follow-up.

## Verification

- `bun run typecheck` ✓, `bun run lint` ✓
- `bun test ./test` ✓ (the 8 `service.test.ts` spawn failures reproduce on `main` too — a local Shepherd-session env leak, `SHEPHERD_HOOKS_INGEST=1`; CI is green without it. Pre-push passed with those vars scrubbed.)
- `grep "TODO(#1567 follow-up)" src` → empty; no `Promise.resolve(...)` wrapper or optimizer cast remains.

Closes #1597. Server-only type-correctness, no user-facing UX → `[no-feature-entry]`.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -420,11 +420,7 @@ export interface AppDeps {
    *  Optional so environments/tests that don't wire it still type-check; the route
    *  no-ops the trigger when absent. Wired to the real DistillerService in index.ts. */
   distiller?: {
-    // TODO(#1567 follow-up): mistyped. DistillerService.distillNow is `async … : Promise<void>`,
-    // so this `=> void` hides the promise from `tsc` and from the lint gate. The trigger route
-    // below no longer LEAKS (it wraps in `Promise.resolve(...).catch`), but the seam still lies.
-    // Retyping forces ~16 test doubles async — pre-existing, so it gets its own PR, not this one.
-    distillNow: (repoPath: string) => void;
+    distillNow: (repoPath: string) => Promise<void>;
     health?: () => {
       ok: boolean;
       consecutiveFailures: number;
@@ -435,12 +431,8 @@ export interface AppDeps {
    *  rules. Optional so tests that don't wire it still type-check; routes no-op when absent.
    *  Wired to the real OptimizerService in index.ts. */
   optimizer?: {
-    // TODO(#1567 follow-up): mistyped — OptimizerService.optimizeOne is `async`. Same as
-    // `distillNow` above; retype alongside its siblings so a marker-driven sweep can't miss it.
-    optimizeOne: (id: string) => void;
-    // TODO(#1567 follow-up): mistyped — OptimizerService.optimizeAllFlagged is `async`. Same
-    // reason for deferring the fix as `distillNow` above.
-    optimizeAllFlagged: (repoPath: string) => void;
+    optimizeOne: (id: string) => Promise<void>;
+    optimizeAllFlagged: (repoPath: string) => Promise<void>;
     health?: () => {
       ok: boolean;
       consecutiveFailures: number;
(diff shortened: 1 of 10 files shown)


scoop (commented):
## Summary

This PR does exactly what the task asks: it retypes the three learnings-service DI seams in `AppDeps` from `=> void` to `=> Promise<void>` so they match their `async … : Promise<void>` implementations, collapses the `Promise.resolve(...)` call-site wrappers back to `deps.x?.foo().catch(...)`, and makes the ~16 test doubles async. It is mechanical, correct, and free of behavioral change.

### Verification

- Implementations confirmed async: `src/distiller.ts:197` (`distillNow`), `src/optimizer.ts:129` (`optimizeOne`), `src/optimizer.ts:138` (`optimizeAllFlagged`) — all `async … : Promise<void>`.
- Seams retyped: `src/server.ts:423,434,435` now `=> Promise<void>`, matching the impls above.
- Call sites: `src/server.ts:1629-1635` and `src/server.ts:1741-1743`. The new form `deps.distiller?.distillNow(dir).catch(...)` is safe — optional chaining short-circuits the *entire* chain, so when `deps.distiller` / `deps.optimizer` is `undefined` the `.catch` is never invoked (no TypeError), and when present the returned `Promise<void>` gets its handler attached. Behaviorally equivalent to the prior `Promise.resolve(...)` wrapper, since the async impls never throw synchronously.
- Test doubles: all doubles in the 9 listed files converted to `async () => {}`. The two rejecting doubles (`test/server-learnings.test.ts:886,908`) now return `() => Promise.reject(...)` (type `Promise<never>`, assignable to `Promise<void>`), correctly dropping the old `as unknown as (…) => void` casts.
- Real wiring: `src/index.ts:2446-2447` injects the real `DistillerService`/`OptimizerService` instances, whose async methods satisfy the new type.

- `test/server-hold-capacity.test.ts:52` still uses a sync `distillNow: () => {}` double, but line 53 casts the object `as unknown as AppDeps`, so the double-cast bypasses the structural check and it will not break `tsc` under the new seam type. This is why the file is (correctly) absent from the change set; no action needed.

PR #1612, opened 2026-07-10
fix(herdr): prefer HERDR_SESSION socket over inherited pane socket (#1596)

### LINKED ISSUE #1596: herdr socket driver ignores HERDR_SESSION when launched inside a herdr pane (HERDR_SOCKET_PATH inheritance)
## Summary
A Shepherd process launched from **inside a herdr pane** silently ignores `HERDR_SESSION` and attaches to whatever herd its parent pane belongs to — because herdr exports `HERDR_SOCKET_PATH` into every pane's environment, and `config.ts` gives that variable precedence over `HERDR_SESSION`.

This is a real footgun, not theoretical: while exercising #1567 I launched a second (test) Shepherd with `HERDR_SESSION=shepherd-1567` set, and it spawned an agent straight into the **production** herd. The rogue agent even inherited the launching pane's `workspace_id`, which is how it was traced.

## Root cause
`src/config.ts`:
```ts
const herdrSession = process.env.HERDR_SESSION ?? "default";
const herdrSocketPath =
  process.env.HERDR_SOCKET_PATH ??                      // <-- wins outright
  (herdrSession !== "default"
    ? `${HOME}/.config/herdr/sessions/${herdrSession}/herdr.sock`
    : `${HOME}/.config/herdr/herdr.sock`);
```
The precedence is documented ("HERDR_SOCKET_PATH overrides outright"), but the *inherited* value is invisible. herdr sets these in every pane:
```
HERDR_ENV=1
HERDR_PANE_ID=...
HERDR_SOCKET_PATH=/home/USER/.config/herdr/herdr.sock   <-- the default-session socket
HERDR_TAB_ID=...
HERDR_WORKSPACE_ID=...
```
So any Shepherd (or `herdr` CLI call) run from a managed pane resolves to the parent herd's socket regardless of `HERDR_SESSION`. Setting `HERDR_SESSION` alone does nothing; you must pin `HERDR_SOCKET_PATH` explicitly.

## Impact
- A dev/test instance can spawn agents into, `haltAll`, or steer the live production herd.
- Affects both the socket driver and the sync CLI path (the `herdr` binary honours `HERDR_SOCKET_PATH` too).

## Proposed fix
Boot-time warning (or hard error behind a flag) when an **inherited** `HERDR_SOCKET_PATH` (tell: `HERDR_ENV=1` is set) disagrees with an explicitly-set `HERDR_SESSION`. That turns a silent cross-herd attach into a log line. Optionally: when `HERDR_SESSION` is set explicitly but `HERDR_SOCKET_PATH` is only inherited, prefer the session-derived path and warn.

## Workaround (today)
Pin `HERDR_SOCKET_PATH=$HOME/.config/herdr/sessions/<name>/herdr.sock` explicitly when launching a non-default instance.

Discovered during #1567 (PR #1575).

scoop:
Fixes #1596.

## Problem

A Shepherd launched **inside a herdr pane** silently ignored `HERDR_SESSION` and attached to whatever herd its parent pane belongs to. herdr exports `HERDR_ENV=1` + `HERDR_SOCKET_PATH=<default-session sock>` into every managed pane, and `config.ts` let the inherited (invisible) `HERDR_SOCKET_PATH` override an explicit `HERDR_SESSION` outright. Real footgun: a dev/test instance could spawn agents into — or steer / `haltAll` — the live production herd.

## Fix

New pure module `src/herdr-session.ts`:

- `resolveHerdrSocket(env, home)` — precedence resolver. **Conflict** = inside a herdr pane (`HERDR_ENV=1`) **and** an explicit non-`default` `HERDR_SESSION` **and** the inherited `HERDR_SOCKET_PATH` differs from that session's socket → the session socket wins (explicit intent). Every other case is unchanged: an explicit/inherited socket still wins, else the session-derived path.
- `applyHerdrSocket(env, home, log?)` — on conflict, rewrites `process.env.HERDR_SOCKET_PATH` in place and warns once. `config.ts` calls it at load.

**Both drivers stay coherent.** The socket driver reads `config.herdrSocketPath`; the CLI driver spawns `herdr` inheriting `process.env` with no `env` override. Rewriting `process.env` on conflict means every herdr subprocess (`herdr.ts`, `diagnostics.ts` liveness, `pty-bridge`/`usage-probe`/`pty-attach.mjs`, `herdr-update.ts`) hits the same session socket.

**Escape hatch — `SHEPHERD_HERDR_IGNORE_SESSION=1`.** Inside a pane, `HERDR_ENV=1` makes an explicit `HERDR_SOCKET_PATH` indistinguishable from the inherited one, so this flag restores the "keep the inherited socket" override (the issue floated "behind a flag"). Named to not collide with the existing `SHEPHERD_HERDR_SOCKET` driver opt-in.

**Offline note.** If the preferred session's daemon isn't running, herdr calls fail — surfaced (not silent) by the existing diagnostics `herdrProbe` → `runHerdrLiveness` → `diagnostics_hint_herdr_offline`; the boot warning names it and the remedies.

## Tests

`test/herdr-session.test.ts` — 8 resolver-precedence branches + the load-bearing `process.env` side effect (rewritten on conflict, untouched otherwise) + the `SHEPHERD_HERDR_IGNORE_SESSION` case.

## Docs

`docs-site/.../configuration.md` — corrected the stale `HERDR_SOCKET_PATH` row (inherited value no longer always wins; dropped the "only consulted when `SHEPHERD_HERDR_SOCKET=1`" clause) + added a `SHEPHERD_HERDR_IGNORE_SESSION` row.

Server-only plumbing — no UI/i18n/feature-catalog surface. `[no-feature-entry]`


scoop (commented):
## Summary

The PR implements the issue's "prefer the session-derived path and warn" option. When Shepherd runs inside a herdr pane (`HERDR_ENV=1`) with a non-`default` `HERDR_SESSION` that disagrees with the pane-inherited `HERDR_SOCKET_PATH`, it now prefers the session's own socket, warns loudly, and rewrites `process.env.HERDR_SOCKET_PATH` in place so the CLI path stays coherent. An `SHEPHERD_HERDR_IGNORE_SESSION=1` escape hatch keeps the old inherited-socket behavior. This directly resolves the silent cross-herd-attach described in the issue.

## Verification

- **Central design claim holds** — the fix depends on the spawned `herdr` CLI inheriting the mutated `process.env`. Verified against `src/herdr.ts:69` and `src/herdr.ts:82`: both `execFileSync`/`execFileAsync` calls pass only `{ encoding, timeout }` and no `env` option, so the child inherits `process.env`. The in-place `env.HERDR_SOCKET_PATH = res.socketPath` write in `src/herdr-session.ts:87` therefore propagates to every spawned CLI. Socket driver and CLI agree on the same herd in the conflict case.
- **Config wiring** — verified `applyHerdrSocket(process.env, process.env.HOME ?? "")` at `src/config.ts:45` replaces the old inline computation, and the returned `herdrSession`/`herdrSocketPath` are the only consumers (config.ts:463-464). No other reader of the removed inline constants (grepped `herdrSession|herdrSocketPath` across the tree).
- **Precedence logic** — `resolveHerdrSocket` (src/herdr-session.ts:44) gates the override on `!ignoreSession && insidePane && session !== "default" && inherited != null && inherited !== sessionPath`. Outside a pane an explicit socket always wins; equal inherited/session paths produce no conflict; the escape hatch suppresses the override. The 10 tests in `test/herdr-session.test.ts` cover each branch and match the implementation.
- **Docs** — `docs-site/src/content/docs/reference/configuration.md` updates the `HERDR_SOCKET_PATH` row and adds the `SHEPHERD_HERDR_IGNORE_SESSION` row; both descriptions match the implemented precedence and the `#1596` reference (matches recent commit 4a825826).
- No user-facing message-catalog strings are involved (the warning is a `console.warn`), so no i18n/locale-parity concern.

## Notes (non-blocking)

- The override cannot distinguish a truly *inherited* `HERDR_SOCKET_PATH` from one a user explicitly sets while inside a pane; it uses `HERDR_ENV=1` as the proxy. This is an inherent, documented limitation with `SHEPHERD_HERDR_IGNORE_SESSION=1` as the escape hatch — acceptable and called out in the code comment (src/herdr-session.ts:35-37).
- When the override selects a session socket whose daemon isn't running, herdr calls fail offline rather than silently attaching to the parent herd. This is the intended safety tradeoff and is spelled out in the warning text.

No blocking problems found.

PR #1614, opened 2026-07-10
fix(diagnostics): self-heal false herdr "offline" verdict

scoop:
## Problem

A **healthy** herdr server was diagnosed `offline` and stayed that way for hours (the persistent "Problem" card with a **Fix** button), even though `herdr status server` → running and `herdr agent list` exits 0.

## Root cause — a trigger and an amplifier

- **Amplifier (the reason it's visible for hours):** the pushed `diagnostics:status` verdict only re-ran on a **6-hour interval** (`src/index.ts`), and the client seed-fetches diagnostics once (`+page.svelte`) then updates **only** from that push (`store.svelte.ts`) — no periodic re-GET. So **any** transient hard `error` (herdr `offline`, a forge host's failed `gh auth`) stays pinned for up to 6h. herdr goes transiently offline on more than boot: its own `Restart=always` bounce, and the #1578 `update --handoff` socket bounce.
- **Trigger (boot race):** the *installed* `~/.config/systemd/user/shepherd.service` was missing `After=herdr.service` (only `After=network-online.target`), so at boot Shepherd started ~7s before herdr; the diagnostics probe (4s after Shepherd boot) raced herdr's startup and latched `offline`. The repo's `deploy/shepherd.service` already carries the ordering fix, but `update.sh` self-heals the timers + `herdr.service` (#1574) and **never re-synced `shepherd.service`** — so existing hosts kept the pre-fix unit.

## Fix (two complementary parts)

**A — adaptive re-check (fixes the amplifier at its source).** Replace the fixed 6h `setInterval` with a self-rescheduling tick: `60s` while `overall === "error"`, else the `6h` steady cadence, so a transient error self-corrects within ~60s of recovery.
- Keys on **`error` only** — a `warning` is steady-state by design (advisory version floors, gh-not-required on lightweight hosts), so fast-polling it would never revert to 6h.
- Bespoke wiring (not `timerTask`, which discards the snapshot): the next tick is armed in `finally`, so a thrown `check()` can't kill the loop and freeze diagnostics.
- `DiagnosticsService.check()/current()/fix()` untouched; the policy is a pure, unit-tested `nextDiagnosticsDelay` helper.

**B — trigger reduction.** `update.sh` now self-heals `shepherd.service` (WorkingDirectory templated, `cmp -s`-gated) exactly like it already does for `herdr.service` + the timers, so the `After=herdr.service` ordering reaches already-provisioned hosts on the next `bun run update`. (`Type=simple` means this orders behind herdr's fork, not its socket-ready state — a probability reduction; Part A is the recovery guarantee.)

## Testing

- `nextDiagnosticsDelay` over its full input domain `{ok, warning, error}` (with a note on why `overall` is never `optional`).
- `update.sh`: presence assertions + a **behavioral** test that runs the real `sed` rewrite against `deploy/shepherd.service` and asserts a well-formed unit with `After=/Wants=herdr.service` intact (catches a malformed rewrite a grep would miss).
- Full root suite (`bun test ./test`), `bun run lint`, `bun run typecheck` — all green.

## Residual (honest)

Recovery is invisible for at most one 60s tick; no fixed-attempt exhaustion window. Warnings intentionally stay on the 6h cadence. A genuinely dead herdr correctly keeps showing `offline` (re-checked each 60s).

### DIFF
--- a/deploy/update.sh
+++ b/deploy/update.sh
@@ -88,6 +88,30 @@ if command -v systemctl >/dev/null 2>&1 && systemctl --user show-environment >/d
   systemctl --user daemon-reload
   systemctl --user enable --now shepherd-logrotate.timer
 
+  # ── sync the core shepherd.service unit ──────────────────────────────────────
+  # Same self-heal rationale as the timers + herdr unit: provision.ts installs shepherd.service
+  # only on a FRESH box, so an already-provisioned host never picks up later changes to the unit
+  # — most consequentially the `After=herdr.service` ordering. Without it, at boot Shepherd can
+  # start BEFORE the sibling herdr daemon, and the diagnostics probe (4s after Shepherd boot,
+  # src/index.ts) races herdr's startup and latches a false `herdr: offline`.
+  #
+  # TEMPLATED like the backup service above (templateUnit in provision.ts): rewrite only the
+  # WorkingDirectory line to THIS checkout so a custom SHEPHERD_DIR install runs against the right
+  # dir. Write only when the rendered unit differs (`cmp -s`, which also treats a missing installed
+  # unit as changed) so a byte-identical unit skips even a redundant daemon-reload. No restart in
+  # this block: the `systemctl --user restart shepherd` below applies the reloaded unit, and the
+  # `After=` ordering only takes effect at the next boot anyway.
+  SHEP_UNIT_TMP="$(mktemp)"
+  sed "s|^WorkingDirectory=.*|WorkingDirectory=${REPO}|" \
+    "$REPO/deploy/shepherd.service" >"$SHEP_UNIT_TMP"
+  if ! cmp -s "$SHEP_UNIT_TMP" "$UNIT_DIR/shepherd.service"; then
+    note "syncing shepherd.service unit"
+    mv "$SHEP_UNIT_TMP" "$UNIT_DIR/shepherd.service"
+    systemctl --user daemon-reload
+  else
+    rm -f "$SHEP_UNIT_TMP"
+  fi
+
   # ── sync the herdr daemon unit (#1574) ───────────────────────────────────────
   # Same self-heal rationale as the timers above: provision.ts installs herdr.service only on a
   # FRESH box, so without this every already-provisioned host would keep running an unsupervised
--- a/src/config.ts
+++ b/src/config.ts
@@ -78,6 +78,16 @@ export const HERDR_SOCKET_SUPPORTED_PROTOCOLS = new Set([16]);
 // TTL backing DiagnosticsService.current() — a request without ?refresh=1 reads
 // this cache. Matches the existing CountsService/backlog 60s TTL.
 export const DIAGNOSTICS_TTL_MS = 60_000;
+// Background diagnostics re-check cadence (src/index.ts). ADAPTIVE: the steady-state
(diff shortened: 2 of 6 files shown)


scoop (commented):
## Summary

The PR addresses "herdr diagnosed offline but it's fine" with two complementary changes:

1. **`deploy/update.sh`** self-heals the core `shepherd.service` unit on already-provisioned hosts (provision.ts only installs it on a fresh box), so the `After=herdr.service` boot ordering is applied — keeping Shepherd from starting before herdr and having the 4s diagnostics probe race herdr's startup.
2. **`src/index.ts` / `src/diagnostics.ts` / `src/config.ts`** replace the fixed `setInterval(6h)` background diagnostics check with an adaptive self-re-arming loop: 60s recheck cadence while `overall === "error"`, else 6h steady. A transient false `herdr: offline` now self-corrects within ~one recheck and is pushed to the client, which otherwise only seeds diagnostics once and thereafter relies on the `diagnostics:status` push.

Both mechanisms are needed: `After=` only orders *start*, it does not wait for herdr's socket to be ready, so the adaptive recheck covers the residual race window.

## Verification

- `nextDiagnosticsDelay` only accelerates on `error`; `warning`/`ok` stay steady. Confirmed against `src/diagnostics.ts:117` (`STATE_RANK` ranks `optional` and `ok` both at 0) and `src/diagnostics.ts:120-126` (`worstOf` upgrades only on a strict increase), so `overall` can never be `optional`; even if it were, the function returns `intervalMs` (steady) — safe. `DiagnosticState` domain verified at `src/types.ts:429`.
- `snapshot.overall` shape verified against `src/diagnostics.ts:561-565` (`check()` returns `{ checks, generatedAt, overall }`).
- New constants `DIAGNOSTICS_INTERVAL_MS`/`DIAGNOSTICS_RECHECK_INTERVAL_MS` exist and are imported/spelled consistently — verified across `src/config.ts:89-90`, `src/index.ts:15-16,2276,2282-2283`, `test/diagnostics.test.ts`. No dangling references to the removed `checkDiagnostics`.
- The tick re-arms in `finally` (`src/index.ts:2287-2289`), so a thrown `check()` cannot freeze diagnostics; the single-chain `setTimeout` design cannot double-schedule overlapping ticks. `timerTask` is still used by other timers and does discard its return value (`src/index.ts:2127-2128`), so the bespoke wiring is justified.
- `update.sh` block: `UNIT_DIR`/`REPO` are in scope (defined at `deploy/update.sh:8,57`); `cmp -s` treats a missing installed unit as changed (write on first sync), and the condition is `set -e`-exempt. The `WorkingDirectory` sed rewrite and mktemp+mv idiom mirror the existing backup/herdr blocks. Rendered unit is well-formed against `deploy/shepherd.service:4,9,13`.

- The `sed "s|^WorkingDirectory=.*|WorkingDirectory=${REPO}|"` replacement shares the standard limitation that a `REPO` path containing `&` or `|` would corrupt the replacement. This is identical to the two pre-existing `sed` uses in the same file (`deploy/update.sh:59`, `:143`), `REPO` is a controlled deploy path, and systemd `WorkingDirectory` paths with those characters are pathological — not worth diverging from the established idiom.
- While herdr is *genuinely* offline the loop fast-polls the full probe batch every 60s indefinitely. This is intentional and documented in the diff comments; it is the price of fast self-correction and is acceptable.

No blocking problems found.

Issue #1617, opened 2026-07-10
build-queue reconcile backstop never fires for Go-released executing sessions despite settled-idle windows


scoop:
## Summary

The build-queue **reconcile backstop** (`src/build-queue-reminder.ts`) never fires for some sessions that authored an approved queue but never POSTed a step status — so the queue stays "all pending" (`0/N`) indefinitely, and the settled-idle nudge that should ask the agent to reconcile never lands. This is the real reliability half of the "queues don't progress" report; the badge-honesty PR only *surfaces* the symptom to the operator, it does not fix this.

## Evidence

Four real sessions, each Go-released into `executing` before implementation, each with an approved all-pending queue, **0** step POSTs, **0** reconcile steers — despite many settled-idle windows well past the 120s threshold:

| Session | Go-release (→executing) | phase during impl | step POSTs | reconcile fired | ≥120s idle windows in executing |
|---|---|---|---|---|---|
| TASK-1464 | 2026-07-10 21:12 | executing | 0 | 0 | 9 (max 738s) |
| TASK-1468 | 2026-07-10 21:15 | executing | 0 | 0 | 5 (max 2436s) |
| TASK-1472 | 2026-07-10 08:27 | executing | 0 | 0 | 22 (max ~11h) |
| TASK-1473 | 2026-07-10 21:08 | executing | 0 | 0 | 10 (max 666s) |

The reconcile steer is **not** globally broken: its exact text fired as a real user-steer in **10 transcripts fleet-wide**, so the path works generally but **selectively failed** for these four. Idle windows were inferred from inter-event gaps in each session's transcript (the same source as the Go-release timestamps).

## Hypothesis space (already partly reduced from logs + code)

- **(c) `getBuildQueue().approved` read false at sweep time — RULED OUT.** `build_queue_state.approved=1`, `approvalKind=auto`, stamped at session creation (per `build_queue_state.updatedAt`), long before any executing idle window; `replaceBuildQueue` never touches `build_queue_state`. Approved was true and stable across the drift window.
- **(b) debounce-entry eviction wipes `sawRunning` — RULED OUT for the drift window.** `list({activeOnly})` excludes only `status='archived'` (`store.ts:1912`); none of the four was archived while executing, and the steps/approved eviction (`build-queue-reminder.ts:150-152`) cannot fire when steps are present and approved. So a once-set `sawRunning` would not have been wiped.
[...]

PR #1618, opened 2026-07-10
fix(build-queue): flag working-but-unreported queues on the badge

scoop:
## Why

Operators saw build-queue badges stuck at `0/N` on sessions that were clearly deep into implementation. Root cause (diagnosed from the DB + four sessions' transcripts): those agents **authored** a queue but never **POSTed** step status, so every step stays `pending` → the badge honestly shows `0/N`. The flat `0/N` reads as "nothing happened" even when the agent is mid-implementation — that's the confusion this PR removes.

## What this does

Adds a client-side **"working but unreported"** state to `BuildQueueBadge`. When a session is past planning (or has an open PR) but its approved queue is still entirely `pending`, the badge drops the misleading flat `0/N` for an indeterminate ⚠ state (striped amber wash, `⚠ {total}`) with a tooltip/aria explaining the agent is working but hasn't updated step status. It clears automatically the moment any step goes `active`/`done`.

Gate (client-side, no server changes):
```
prPresent = git?.state === "open"                       // open PR only; merged/closed = resolved
working   = planPhase !== "planning" || prPresent       // null (gate off) treated like executing,
                                                        //   matching the server backstop
drifted   = approved && steps.every(pending) && working  // truly untouched = the "0/N" case
```
`every(pending)` (not the server's `some(pending) && !some(active)`) is deliberate: the badge has no idle/settle gating, so the looser predicate would flash the warning every time a *compliant* agent is briefly between steps. `git` rides `session:git` (separate from `session:status`), so absent/stale `git` degrades closed — never a false flag.

## Scope — this surfaces the symptom, it does not fix progression

This is the operator-facing half. The **actual reliability gap** — the reconcile backstop never firing for these sessions despite settled-idle windows — is tracked separately in **#1617** (it needs live sweep instrumentation to pin down before a fix; the hypothesis space is already reduced there). This PR does **not** attempt that fix.

## Testing

- `cd ui && bun run check` — 0 errors / 0 warnings
- `cd ui && bun run check:i18n` — locale parity (EN + DE keys added)
- `bun run test:browser -- src/lib/components/BuildQueueBadge.browser.test.ts` — 17/17 (full drifted matrix: executing / gate-off-null / planning×{open,merged,closed,none,undefined} / has-active / has-done / clears-on-active)

Relates to #1617.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -1631,6 +1631,8 @@
   "queuebadge_label": "{resolved}/{total}",
   "queuebadge_title": "Build-Queue: {resolved} von {total} Schritten erledigt",
   "queuebadge_aria": "Build-Queue: {resolved} von {total} Schritten erledigt",
+  "queuebadge_stale_title": "Build-Queue: {total} Schritte, Fortschritt nicht gemeldet – Agent arbeitet, hat aber den Schrittstatus nicht aktualisiert",
+  "queuebadge_stale_aria": "Build-Queue-Fortschritt nicht gemeldet: {total} Schritte, Agent arbeitet, aber Status nicht aktualisiert",
   "feat_build_queue_progress_title": "Build-Queue-Fortschritt auf einen Blick",
   "feat_build_queue_progress_body": "Session-Karten zeigen jetzt ein kleines bernsteinfarbenes Badge mit Fortschrittsbalken, wenn eine genehmigte Build-Queue aktiv ist — erledigte / Gesamtschritte ohne Öffnen des Panels einsehbar.",
   "learnings_optimize": "Optimieren",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -1631,6 +1631,8 @@
   "queuebadge_label": "{resolved}/{total}",
   "queuebadge_title": "Build queue: {resolved} of {total} steps done",
   "queuebadge_aria": "Build queue: {resolved} of {total} steps done",
+  "queuebadge_stale_title": "Build queue: {total} steps, progress unreported — agent working but hasn't updated step status",
+  "queuebadge_stale_aria": "Build queue progress unreported: {total} steps, agent working but status not updated",
   "feat_build_queue_progress_title": "Build queue progress at a glance",
   "feat_build_queue_progress_body": "Session cards now show a small amber badge with a progress meter when an approved build queue is active — see resolved / total steps without opening the panel.",
   "learnings_optimize": "Optimize",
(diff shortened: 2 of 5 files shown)


scoop (commented):
## Summary

This PR addresses the task — a build-queue badge stuck at `0/N` while the agent is actually working — by adding a distinct "drifted" (working-but-unreported) badge state. When a queue is approved, has steps, every step is still `pending`, and the agent is judged to be past planning (`planPhase !== "planning"`, or an open PR is present), the badge renders an animated amber ⚠ marker instead of a misleading `0/N` progress fill. This is a sensible, targeted answer to the reported confusion.

The implementation is correct and I found no blocking issues.

## Verification performed

- **Prop types resolve.** `planPhase: Session["planPhase"]` and `git?: GitState` in `BuildQueueBadge.svelte:7-11` match the definitions — `planPhase` is `"planning" | "executing" | null` (verified against `ui/src/lib/types.ts:848`) and `GitState` extends `PrStatus` with a `state: "none" | "open" | "merged" | "closed"` field (verified against `ui/src/lib/types.ts:283` and `776-777`). The `prPresent = git?.state === "open"` check uses a valid member.
- **Test `git()` literal type-checks.** `{ kind: "github", state, checks: "none", deployConfigured: false }` supplies exactly the required (non-optional) `GitState`/`PrStatus` fields; `kind` is a valid `ForgeKind` and `checks` a valid `ChecksState` (verified against `ui/src/lib/types.ts:275,277,291,299`).
- **i18n parity.** Both new keys `queuebadge_stale_title` and `queuebadge_stale_aria` exist in en and de with the same single `{total}` placeholder used at the call sites (verified against `ui/messages/en.json:1634-1635`, `ui/messages/de.json:1634-1635`, and `ui/src/lib/components/BuildQueueBadge.svelte:42-43`). Only en+de catalogs exist (`ui/messages/`).
- **Consumers updated.** The only component consumer is `UnitRowRight.svelte:126`, which now passes `planPhase={session.planPhase}` and `{git}` (`git` is a prop of that component, `ui/src/lib/components/unit-row/UnitRowRight.svelte:16,33`). Other `BuildQueueBadge` references are comments/docs, not renders. The newly-required `planPhase` prop is therefore always supplied.
- **CSS override is order-safe.** The stale branch keeps class `queue-badge queue-badge--stale`; `.queue-badge--stale` (line 94) is defined after `.queue-badge` (line 67) with equal specificity and fully overrides `background`, so the base left-fill gradient (which would render at an undefined `--queue-pct`) is correctly suppressed.

## Notes (non-blocking observations, no action required)

- The degrade directions are intentionally asymmetric — `planPhase === null` (gate off) degrades *open* (treated as working), while absent git degrades *closed* (not drifted during planning). Each is defensible and both are explicitly covered by the new tests; called out only so a future reader doesn't mistake it for an oversight.
- A queue that is approved and all-pending the instant execution begins (before the first step flips to `active`) can briefly show the ⚠ state during `executing`/gate-off. This is the intended "working but unreported" signal rather than a defect.

PR #1619, opened 2026-07-10
fix(viewport): cap review-in-flight banner so it can't bury the prompt

scoop:
## Problem

The review-in-flight terminal banner became **unbelievably tall and blocked all access to the CLI prompt** in a short terminal pane.

Regression from #1593, which added an always-reserved **4-row** live-activity preview under the headline. The banner grew from ~40px to ~108px. `Viewport.svelte` reserves the strip via `.term-mount { height: calc(100% - var(--review-banner-h)); min-height: 4rem }`; in a short pane `100% − 108px` falls under the `4rem` floor, so the terminal stops shrinking and the tall banner overlays the bottom xterm rows — the input line.

## Fix

Two coordinated changes:

1. **Shrink in every pane** — `MAX_ACTIVITY_LINES` 4 → 2. Banner ~108px → ~75px; shows the 2 newest reviewer-activity lines. Stable-height / one-refit property preserved (a smaller *fixed* reservation costs no refits).
2. **Cap + guarantee reflow** — `.review-banner { max-height: min(50%, calc(100% - 4rem)); box-sizing: border-box; overflow: hidden }`. `50%` bounds it to a fraction of the terminal; `100% − 4rem` guarantees `.term-mount` keeps its `4rem` floor so it always reflows fully above the banner. The preview is now a shrinkable `flex-basis` (gives way first under the cap); the headline is `flex-shrink:0` + `nowrap`/ellipsis so it can't wrap past the cap. Each declaration is load-bearing — drop `border-box` and padding overshoots the cap; drop `overflow:hidden` and `offsetHeight` (→ `--review-banner-h`) under-reports.

### Empirical note (found during implementation)

`.rb-head` is `flex-shrink:0`, so `max-height` can't shrink the banner below its **one headline row (~19px)** — verified in-browser (`max-height:8px` still renders 19px; `40%`→24px clamps fine). Consequences, probed on the real `Viewport`:

- Any usably-sized terminal (`.vp-body ≳ 71px`): cap engages, banner leaves ≥ 4rem, **no overlap** (measured: vp-body 100px → banner 48px, term 52px, 0 overlap).
- Only at the absolute `.vp-body` 4rem floor (expanded-recap takeover — where nothing fits both a prompt and a banner) does a residual overlap of **at most one headline row** remain — still far less than the pre-cap ~108px burial. The stale `Viewport.svelte` / banner comments were rewritten to state this real invariant.

## Verification

- `cd ui && bun run check` → 0 errors; `check:i18n` → parity (no new strings); eslint + prettier clean on touched files.
- `cd ui && bun test` → **3229 passed**.
- Retargeted the feed-cap tests 4→2; added a **short-pane cap** test (`ReviewInFlightBanner`) asserting `banner ≤ container − 4rem`, and a **Viewport-level no-overlap / reflow-guarantee** test (`.term-mount` stays ≥ 4rem and never extends under the banner).

`[no-feature-entry]` — this is a bug fix to existing UI, no new user-facing capability.

### DIFF
--- a/ui/src/lib/components/Viewport.browser.test.ts
+++ b/ui/src/lib/components/Viewport.browser.test.ts
@@ -770,19 +770,23 @@ describe("Viewport review banner — reflow guarantee (no prompt overlap)", () =
     planGates.applyReviewing(id, true);
     planGates.setActivity(id, "read .shepherd-plan.md");
 
-    const { container } = render(Viewport, {
+    render(Viewport, {
       session: session({ id, status: "running", planPhase: "planning" }),
+      activity: activity("read .shepherd-plan.md"),
       previewPort: null,
       openPreviewTick: 0,
     });
-    // Force a SHORT terminal (well above .vp-body's 4rem floor) so the cap has to engage.
-    (container as HTMLElement).style.height = "320px";
-
-    const banner = () => container.querySelector<HTMLElement>(".review-banner");
+    const banner = () => document.querySelector<HTMLElement>(".review-banner");
     await vi.waitFor(() => expect(banner()).not.toBeNull());
-    const mount = container.querySelector<HTMLElement>(".term-mount")!;
+    // Force a SHORT terminal (well above .vp-body's 4rem floor) so the cap has to engage.
+    // Resize the Viewport's mount div (its height:100% root fills it); read it off the DOM
+    // because render()'s typed result varies by version.
+    const host = document.querySelector<HTMLElement>(".viewport")!.parentElement as HTMLElement;
+    host.style.height = "320px";
+
+    const mount = document.querySelector<HTMLElement>(".term-mount")!;
     expect(mount, ".term-mount should render").not.toBeNull();
-    const vpBody = container.querySelector<HTMLElement>(".vp-body")!;
+    const vpBody = document.querySelector<HTMLElement>(".vp-body")!;
     const fourRem = 4 * parseFloat(getComputedStyle(document.documentElement).fontSize);
 
     await vi.waitFor(() => {
--- a/ui/src/lib/components/Viewport.browser.test.ts
+++ b/ui/src/lib/components/Viewport.browser.test.ts
@@ -747,6 +747,58 @@ describe("Viewport autopilot badge vs in-flight review", () => {
   });
 });
 
+describe("Viewport review banner — reflow guarantee (no prompt overlap)", () => {
+  function clearReviewState() {
+    reviews.map = {};
+    reviews.reviewing = {};
+    reviews.activity = {};
+    planGates.map = {};
+    planGates.reviewing = {};
+    planGates.activity = {};
+  }
+  beforeEach(clearReviewState);
+  afterEach(clearReviewState);
+
+  // The literal reported bug: in a short terminal the tall banner overlaid the CLI prompt.
+  // The banner caps its height (max-height: min(50%, 100% - 4rem)) so it leaves ≥ 4rem for
+  // the terminal; .term-mount (= 100% - --review-banner-h, floored at 4rem) then reflows fully
+  // ABOVE it — keeping its 4rem floor and never extending under the banner. Rendered at a short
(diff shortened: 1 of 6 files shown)


scoop (commented):
## Summary

The PR correctly diagnoses and fixes the reported bug: the in-flight review banner grew tall enough (4-row live-activity preview + headline) to bury the terminal prompt in a short pane. The fix is two-pronged and, as far as I can verify, sound:

1. **Hard height cap on the banner.** `ui/src/lib/components/viewport/ReviewInFlightBanner.svelte:342` adds `max-height: min(50%, calc(100% - 4rem))` on the absolutely-positioned `.review-banner`, plus `box-sizing: border-box` (`:341`) so padding/border fold into the cap, and `overflow: hidden` (`:343`) so the measured `offsetHeight` equals the clamped box.
2. **Smaller reserved preview + shrink-to-fit.** `MAX_ACTIVITY_LINES` drops 4→2 (`ui/src/lib/reviews.svelte.ts:18`), and `.rb-preview` becomes `flex: 0 1 …; min-height: 0` (`ReviewInFlightBanner.svelte:383-384`) while `.rb-head` gets `flex-shrink: 0` (`:366`) and `.rb-text` ellipsizes (`:432-435`) — so the preview gives way first under the cap and the headline never wraps past it.

### Verification of the reflow guarantee
- The banner's height is published via `bind:offsetHeight={height}` (`ReviewInFlightBanner.svelte:265`) → `--review-banner-h`, consumed at `Viewport.svelte:2736` (`reviewBannerH || ciBannerH`).
- `.term-mount { height: calc(100% - var(--review-banner-h,0px)); min-height: 4rem }` (`Viewport.svelte:3654-3655`). Since the cap guarantees `banner ≤ 100% - 4rem` relative to its containing block `.vp-body` (which is `position: relative`, `Viewport.svelte:3626`, and holds a definite height so the `%` max-height resolves — confirmed the test suite deliberately establishes a positioned/definite containing block at `ReviewInFlightBanner.browser.test.ts:133-135`), `.term-mount` retains its 4rem floor. No overlap on any usably-sized terminal. The cap depends only on `.vp-body`'s height (banner is out of flow), so there is no feedback loop with `.term-mount`.
- `MAX_ACTIVITY_LINES` is used consistently as both the feed cap (`reviews.svelte.ts:26`, `slice(-MAX_ACTIVITY_LINES)`) and the reserved row count `--rb-pv-rows` (`ReviewInFlightBanner.svelte:299`); tests updated to match (`reviews.svelte.test.ts:195`, `ReviewInFlightBanner.browser.test.ts:96,131`).

### Consistency checks
- No new i18n keys introduced; the only message reference `m.reviewbanner_preview_waiting()` (`ReviewInFlightBanner.svelte:301`) is pre-existing, so no en/de parity concern arises from this diff.
- The comment at `Viewport.svelte:3645-3653` accurately restates the cap `min(50%, 100% - 4rem)` matching the actual CSS at `ReviewInFlightBanner.svelte:342`.

### Design observation (non-blocking, no change required)
- `MAX_ACTIVITY_LINES` now serves two roles at once — feed-history depth and rendered preview-row count. They're intentionally coupled and documented (`reviews.svelte.ts:15-17`); fine as-is, just noting the two would need decoupling if they ever must diverge.

- ReviewInFlightBanner.svelte:342 — in the expanded-recap takeover where `.vp-body` is pinned to its own 4rem floor (`Viewport.svelte:3633`), the cap resolves to `min(~32px, 0px)=0`, so the banner is clamped to ~0 and effectively hidden (the "review running" visual cue disappears, though `aria-live` still announces). This is the documented tradeoff (`ReviewInFlightBanner.svelte:326-328`) and preserves prompt access, so it's an acceptable edge rather than a defect.

- ui/src/lib/components/viewport/CiRunningBanner.svelte — the CI-running banner shares the `bottom:0; position:absolute` strip pattern and publishes the same `--review-banner-h`, but is not height-capped. It is inherently single-row (no multi-line preview) so it does not exhibit the tall-banner problem; noted only for completeness.

No blocking problems found. The implementation satisfies the task.

PR #1620, opened 2026-07-11
feat(herdr): stream the browser terminal over terminal session control (#1529)

scoop:
Streams the browser live-terminal over herdr's native `terminal session control` NDJSON instead of `node-pty`, behind the existing `SHEPHERD_HERDR_SOCKET` flag. This is **opportunity #1 of #1529** only; the socket control-plane driver (opp #2) already merged, and the rest of #1529 (#3 events, #4 report_agent, #5 schema types, #6 remote) stays tracked in #1529.

`[no-feature-entry]` — internal transport swap, no user-facing UX (the terminal looks/behaves identically; the browser side is unchanged).

## What & why

Today the browser terminal runs `Bun.spawn → pty-attach.mjs (Node) → node-pty → herdr agent attach --takeover`. This adds a Bun-native alternative that spawns `herdr terminal session control <pane> --takeover` and pipes its NDJSON directly — no node-pty, no separate Node process, no `nodeBin` for this path. It activates **only** when the socket driver is active (flag on **and** the protocol-16 `ping` succeeded); otherwise node-pty, which stays the default.

- `src/socket-pty-bridge.ts` — `SocketPtyBridge`: decodes `terminal.frame` (base64 ANSI) → `ws.send` in arrival order; reuses the existing `createDemux` to turn browser input/`\x00resize:` into `terminal.input`/`terminal.resize`; `terminal.release` on close. Adaptive first-frame watchdog (seed from a measured latency, capped, self-widening per host). Hooks: `onFirstFrame`/`onFallback`/`onGone`/`onAbnormalExit`.
- `src/terminal-transport-metrics.ts` — real per-boot `socketAttach`/`socketFallback` counters (`markPtyEvent` is a profiling no-op that counts nothing; `DiagnosticsService` is a payload-pure prober — neither fit), plus a boot-grace self-check that WARNs when the flag is active but no attach ever succeeds while fallbacks climb.
- `src/server.ts` — `livePtyAttach` (co-resolves `paneId`+`workspaceId` into the target, **preserving** the herdr-hiccup fallback: a throwing `list()` still attaches via node-pty, never closes); per-connection selection; `message()` buffers input into `pendingInput` during the attach gap; `onFallback` re-attaches node-pty on the **same ws** and replays the buffer; a per-`terminalId` failure memo (past-TTL sweep that keeps the current entry) so a reconnect after a socket failure prefers node-pty; new **authenticated** `GET /api/terminal-transport`.
- `src/index.ts` — `herdrSocketActive = herdr instanceof SocketHerdrDriver`; starts the self-check.
- `test/fixtures/terminal-control/` — the real captured wire contract; `scripts/verify-herdr-terminal.ts` — one-command live re-verify to run after any herdr upgrade.

## Guardrails (the terminal NDJSON is a CLI contract, NOT covered by the `protocol:16` schema, and herdr auto-updates — so there is no version pin to lean on)

- **Runtime fallback is the authoritative guard.** Pre-first-frame failure (spawn throw / early exit / watchdog / no-NDJSON) → seamless node-pty re-attach on the same ws (no blank terminal, buffered keystrokes replayed). Post-first-frame death → the failure memo makes the reconnect come up on node-pty (no loop). `terminal.closed{not found}` → `PTY_GONE` (4001), same as node-pty.
- **Observable in prod.** The `socketAttach`/`socketFallback` counters + the self-check WARN + `/api/terminal-transport` make a fleet-wide silent fallback visible rather than passing as success.

## Deviation from the plan

The plan said fold the counters into `/api/health`. During implementation I found `handleHealth` is a deliberately-minimal **un-credentialed public** probe documented as "discloses nothing" — so the counters go on a new **authenticated** `GET /api/terminal-transport` instead, and `handleHealth` is left untouched.

## Scope

node-pty stays (`pty-bridge.ts`, `pty-attach.mjs`, `pty-demux.mjs`, `config.nodeBin` untouched) — `usage-probe.ts` is a second, independent node-pty consumer, so full node-pty/toolchain removal is a separate follow-up gated on migrating it. No `ui/` changes.

## Verification

- `bun test ./test` → **6447 pass, 8 skip, 0 fail**. `bun run typecheck`, full `bun run lint`, and the PR-hygiene gates (branch hygiene, i18n parity, feature catalog, glossary, announcement versions) all pass.
- **Live integration smoke** drove the real `SocketPtyBridge` against real herdr 0.7.3: confirmed attach + `socketAttach` counter, typed-input echo, resize redraw, the version-skew fallback (records fallback **and** keeps the ws for node-pty), and the not-found → `onGone` path — 12/12.
- Reviewed per-task (incl. an opus review of the server hot path) and a final whole-branch review (READY TO MERGE); all findings fixed.

Refs #1529 (partial — does not close it; the remaining opportunities stay tracked there).


scoop (commented):
## Summary

This PR implements opportunity #1 from the issue (stream `herdr terminal session control` directly over its stdio, no node-pty) as a flag-gated prototype behind the existing PTY seam, with a robust node-pty fallback. Scope matches the issue's "track-and-prototype behind seams, don't rip-and-replace" intent: the socket terminal path is only chosen when `herdrSocketActive` (flag on + `SocketHerdrDriver` selected) AND a `paneTarget` resolved AND the terminal hasn't recently failed over the socket. Every failure mode (spawn throw, pre-frame process exit, watchdog timeout, `terminal.closed{not found}`, `terminal.closed{other}`, post-frame abnormal exit, deliberate close) is classified into a distinct, mutually-exclusive outcome and covered by `test/socket-pty-bridge.test.ts`.

I verified the correctness-relevant claims below; no blocking defects found.

## Verification performed

- **Outcome mutual-exclusivity / no double-close on handoff.** `fireFallback` sets `handedOff=true` before invoking `onFallback`, and `closeWs()` short-circuits on `handedOff` — so the ws the caller re-attaches to node-pty is never torn down underneath it. `handleFrame` returns early on `outcomeFired`, so a stray frame from a killed proc's pipe after fallback can't fire `onFirstFrame` or reach `ws.send`. Verified against src/socket-pty-bridge.ts:245-266, 279-284, 132-152, and the asserting tests test/socket-pty-bridge.test.ts:267-311, 397-421.
- **Confirmed vs. fallback are exclusive.** Once a frame is seen `confirmed=true` blocks all pre-confirm fallback paths (handleClosed/watchdog/onProcessExit each guard on `confirmed`), and `outcomeFired` blocks frame processing after any outcome — so `flushPending` runs at most once. Verified src/socket-pty-bridge.ts:174-181, 200-206, 265-283.
- **pendingInput buffering.** During `awaitingFirstFrame`, input frames are buffered (not written to the socket bridge) and flushed once on either `onFirstFrame` (to the socket bridge) or `onFallback` (to the fresh node-pty bridge, whose demux is fresh — no double-consume). Verified src/server.ts:6975-6999 and message handler src/server.ts:7008-7009.
- **`data` alias identity.** `const data = ws.data as Extract<...>` is a cast, not a copy, so mutations via `data.bridge/awaitingFirstFrame/pendingInput` are observed by the message/close handlers through `ws.data`. Verified src/server.ts:6969-6971 vs 7008-7009.
- **Deliberate close suppresses hooks.** `close()` sets `closing=true`; handleClosed/watchdog/onProcessExit all early-return under it, so a client disconnect during `awaitingFirstFrame` does NOT spawn a node-pty fallback. Verified src/socket-pty-bridge.ts:126-135, 200-206, 189-198.
- **paneTarget composition + demux routing.** `livePtyAttach` composes `paneId.includes(':') ? paneId : ${workspaceId}:${paneId}`, matching the pinned contract in test/fixtures/terminal-control/capture-notes.md and covered by test/socket-terminal-wiring.test.ts:96-114. `write()` feeds `pty-demux.mjs`, which splits input vs `\x00resize:` frames into `terminal.input`/`terminal.resize` commands — verified src/pty-demux.mjs:10-59 and test/socket-pty-bridge.test.ts:180-205.
- **Session-socket env parity.** The spawned `herdr terminal session control` inherits `process.env` (including the `HERDR_SOCKET_PATH` that `applyHerdrSocket` rewrites, config.ts:37-48), matching how the node-pty path passes env (src/pty-bridge.ts:36) — so a named session connects to the right socket. Verified src/socket-pty-bridge.ts:99-102.
- **Auth gating.** `/api/terminal-transport` is under `/api` and not in `isPublicRequest`'s allowlist (src/server.ts:547-559), so `checkAuth` (src/server.ts:575-588) gates it — the "Authenticated (unlike handleHealth)" comment is accurate. Route guard falls through correctly on wrong method/extra path (test/socket-terminal-wiring.test.ts:240-252).
- **Rename completeness.** `livePtyTerminalId` renamed to `livePtyAttach` has no stale callers (grep across src/test/ui). Imports `markPtyEvent`, `createDemux`, `SocketHerdrDriver`, `selectHerdrDriver` all resolve (src/instrument.ts:135, src/pty-demux.mjs:10, src/herdr-socket-driver.ts:27/245).
- **Metrics self-check.** `shouldWarnSilentFallback` returns true only when `flagActive && socketAttach===0 && socketFallback>0` — no false warning when zero terminals were opened. Timer is unref'd/one-shot. Verified src/terminal-transport-metrics.ts:52-97. No user-facing catalog strings added, so no i18n parity concern.

- src/server.ts (livePtyAttach + pickTerminalBridgeKind): when a matched agent has an empty `pane_id` AND empty `workspace_id`, `paneTarget` becomes the truthy string `":"` and the socket path is chosen; herdr would answer `terminal.closed{not found}`, which is classified as `onGone` then `PTY_GONE` (session shown as ended) rather than a node-pty fallback. Dormant today because `parseAgents`/`parseAgentInfo` note that herdr always supplies them in practice (src/herdr.ts:234-243, 261-264); becomes reachable only if herdr ever returns a live-but-fieldless agent. Consider guarding `pickTerminalBridgeKind` against a degenerate empty-segment target so it falls to node-pty.
- src/socket-pty-bridge.ts (handleClosed): a pre-first-frame `terminal.closed{not found}` is treated as definitively agent-gone (`onGone` then `PTY_GONE`, no node-pty fallback), even though `matchAgent` just confirmed the agent is live in `list()`. Correct for herdr 0.7.3 where the `terminal session control` target format matches `api snapshot`'s `pane_id` (verified in capture-notes.md), but on this explicitly-unstable preview API a future target-format change would surface as `not found` and close live sessions instead of falling back to node-pty — the one version-skew mode that bypasses the otherwise-comprehensive fallback. Worth a note/monitor as the `verify-herdr-terminal.ts` drift check evolves.

## Stated limitations
- The wire contract (`terminal.frame`/`terminal.closed` field names, `not found` reason text, target format) is pinned against a live herdr 0.7.3 capture (test/fixtures/terminal-control/*, capture-notes.md), not against an in-repo herdr schema. I confirmed the parser/emitter agree with those fixtures and the tests derive from them, but the fixtures themselves reflect an external binary I cannot re-verify against a live daemon here — assumed as captured. The included `scripts/verify-herdr-terminal.ts` is the intended re-verification harness for herdr upgrades.

- `runningMaxFirstFrameMs` in src/socket-pty-bridge.ts is process-global and monotonic; the `watchdog cap` test pushes it to 1_000_000, which could make later same-process tests that omit `watchdogMs` arm at the 30s cap — a test-isolation nuance, not a production issue (fresh process per boot; all other watchdog tests use an override).

scoop (commented):
## Summary

This PR adopts herdr's native `terminal session control` NDJSON stream as an alternative to the node-pty helper for the browser terminal, gated behind the existing `SHEPHERD_HERDR_SOCKET` flag (only engaged when the socket driver actually activated). The design is careful: single-owner takeover, an adaptive first-frame watchdog with a hard cap, a per-terminal failure memo with TTL so a bad attach retries node-pty for a cooldown, a per-boot metrics counter + a silent-fallback self-check, and a pinned wire-contract fixture with a live re-verification script. The fallback-to-node-pty seam is preserved throughout, matching the issue's "track-and-prototype, don't rip-and-replace" intent.

Verification performed:
- `config.herdrSocket` exists and is the flag read by `handleTerminalTransport` — verified `src/config.ts:552`.
- Route auth: `/api/terminal-transport` is a GET under `/api`, not exempted by `isPublicRequest` (only `/api/health` is), so it sits behind `checkAuth` — verified `src/server.ts:547-559` and the comment's "Authenticated (unlike handleHealth)" claim holds.
- `paneTarget` composition `paneId.includes(":") ? paneId : ` ${workspaceId}:${paneId}`` matches the pinned contract and the wiring tests (`test/socket-terminal-wiring.test.ts:96-113`), and `HerdrAgent` fields resolve (`src/herdr.ts:17-20`, `parseAgents` `src/herdr.ts:239-242`).
- Renamed export: no stale `livePtyTerminalId` callers remain; all consumers use `livePtyAttach` (`src/server.ts:6784`, `:6928`).
- The two bridge types share the `write(string)`/`close()`/`open(cols,rows)` shape used by the union (`src/pty-bridge.ts:52-67`); pending-input frames are stored RAW and correctly re-dispatched — node-pty's helper demuxes internally (`src/pty-attach.mjs`), while `SocketPtyBridge` demuxes in-process via `createDemux` into `terminal.input`/`terminal.resize` (`src/pty-demux.mjs:10-59`). Both handoff targets (`sb` on first-frame, `nb` on fallback) receive raw frames — consistent.
- Handoff/close bookkeeping (`handedOff`/`wsClosed`/`outcomeFired`) prevents the socket bridge from tearing down a ws that node-pty has taken over — verified against the bridge tests (`test/socket-pty-bridge.test.ts:267-321`).
- No i18n/message-catalog surfaces touched (only `console.info`/`console.warn` and a docs table row).

## Finding: exit-vs-stdout-drain race can misclassify a `terminal.closed` outcome

`src/socket-pty-bridge.ts:178-188` — `watchExit` awaits `proc.exited` and then calls `onProcessExit()` immediately, with no coordination with the `pumpStdout` loop (`:151-161`). `onProcessExit` (`:270-283`) classifies a pre-first-frame exit as `fireFallback()`, relying on any `terminal.closed` line having ALREADY been parsed (the comment at `:277` says "no-op if onGone/onFallback already fired from a closed line"). But `proc.exited` can resolve before the stdout pump reads the process's final buffered chunk. For the bad-target case (herdr writes one `terminal.closed{reason:"not found"}\n` then exits 0 — see `test/fixtures/terminal-control/capture-notes.md:47-49`), if the exit continuation wins the race, `onProcessExit` fires `onFallback` (setting `outcomeFired`), and the later-parsed closed line's `fireGone()` becomes a no-op. The agent-gone signal is then mis-mapped to a node-pty fallback instead of `PTY_GONE (4001)`.

Impact is degraded, not fatal: the node-pty fallback attaches to the dead terminalId, the helper exits, the client reconnects, and `livePtyAttach` -> `matchAgent` returns null so the reconnect yields `PTY_GONE`. So it self-heals in one extra round-trip, but it also records a spurious `socketTerminalFailures` stamp + fallback metric (which can trip the "silent fallback" warning). The unit tests only exercise the favorable order (`test/socket-pty-bridge.test.ts:256-259` pushes the closed line and ticks BEFORE `resolveExit`), so the reverse ordering is untested. Suggest awaiting the stdout pump's completion (or draining any complete buffered NDJSON lines) before classifying in `onProcessExit`.

I could not verify Bun's exact scheduling of `proc.exited` relative to the stdout ReadableStream's final read from within this repo — flagged as a robustness gap rather than a definitively reproduced failure; hence non-blocking.

- src/server.ts:6797 (livePtyAttach): if a matched agent ever has BOTH empty `paneId` and empty `workspaceId`, `paneTarget` becomes `":"` (truthy), so `pickTerminalBridgeKind` selects the socket path and `herdr terminal session control :` errors into a fallback rather than being skipped up front. Not reachable with well-formed herdr list output, but a `.includes(":")`-with-non-empty-halves guard would be tidier.

- src/pty-bridge.ts: node-pty write path passes raw frames (incl. `\x00resize:` control frames) straight to the helper, which demuxes — noted only to explain why the socket path demuxes in-process instead; no change needed.

scoop (commented):
## Re-review of prior finding

**Prior point (socket-pty-bridge.ts exit race): RESOLVED.**

The previous revision classified a pre-first-frame exit the moment `proc.exited` resolved, with no coordination with the stdout pump, so a fast bad-target's trailing `terminal.closed{not found}` could be lost to the exit race and mis-mapped to node-pty fallback instead of `onGone`.

The new code fixes this:
- `pumpStdout` now stores its promise in `this.stdoutPump` (src/socket-pty-bridge.ts:158-165).
- `watchExit` awaits `this.stdoutPump` and then `flushStdoutBuf()` **before** calling `onProcessExit()` (src/socket-pty-bridge.ts:193-195). Because `proc.stdout` EOFs when the child exits, awaiting the pump guarantees every buffered NDJSON line is delivered/parsed first; `flushStdoutBuf` (src/socket-pty-bridge.ts:201-205) handles a final unterminated line.
- Verified flow: the buffered `terminal.closed{not found}` reaches `handleClosed` → `fireGone` (sets `outcomeFired`) before `onProcessExit`, so `onProcessExit`'s `fireFallback` is a no-op (src/socket-pty-bridge.ts:292-298, 302-309). Correctly mapped to `onGone`/`PTY_GONE`.
- The ordering is safe: `pumpStdout(proc)` runs before `watchExit(proc)` in `open()` (src/socket-pty-bridge.ts:131,133), so `this.stdoutPump` is assigned before watchExit's async body awaits it. `stdoutPump` never rejects (its `.catch` swallows), so the await can't hang on a torn stream.
- Directly covered by a regression test that pushes the not-found line then resolves exit with no intervening tick and asserts `onGone` fires and `onFallback` does not (test/socket-pty-bridge.test.ts:496-519).

## Other changes verified

- **server.ts terminal bridge selection** (src/server.ts:6925-6989): `pickTerminalBridgeKind` gates the socket path on `herdrSocketActive && paneTarget && !recentlyFailed`; otherwise node-pty. `pendingInput`/`awaitingFirstFrame` buffer keystrokes until the first frame or a fallback, then `flushPending` replays them (verified `onFirstFrame` flushes to `data.bridge` = the SocketPtyBridge, and `onFallback` builds a node-pty `PtyBridge` on the same `sock` and flushes to it). The `SocketPtyBridge.handedOff` flag (src/socket-pty-bridge.ts:306,324) ensures the socket bridge never closes the ws it handed to node-pty.
- **PtyBridge ctor compatibility**: `new PtyBridge(tid, sock)` matches `constructor(terminalId, ws: PtySocket, ...)` with defaulted helperPath/nodeBin (src/pty-bridge.ts:20-25). `PtySocket` shape `{send, close}` matches `sock` (src/pty-bridge.ts:6-9).
- **matchAgent shape**: `paneId`/`workspaceId`/`terminalId` all exist on `HerdrAgent` (src/herdr.ts:17-20), consistent with the `paneTarget` construction in `livePtyAttach`.
- **config key**: `config.herdrSocket` exists (src/config.ts:552) and is read by both `handleTerminalTransport` (src/server.ts:5701) and the index self-check wiring (src/index.ts:435).
- **terminal-transport-metrics.ts**: per-boot counters + unref'd one-shot self-check; snapshot returns a copy; `shouldWarnSilentFallback` is a pure predicate. No issues.
- No user-facing catalog strings introduced (only `console.info`/`console.warn`), so no i18n parity concern.

- src/server.ts:6931 — `livePtyAttach` builds `paneTarget` as `${a.workspaceId}:${a.paneId}` when `paneId` lacks a colon; if a matched live agent ever has an empty `pane_id` (herdr.ts maps `a.pane_id ?? ""`), `paneTarget` becomes `":"` (or `ws:`), which is truthy so `pickTerminalBridgeKind` selects the socket path, and a `terminal.closed{not found}` on that degenerate target would fire `onGone`/`PTY_GONE` and close a live session instead of falling back. Only reachable with the `SHEPHERD_HERDR_SOCKET=1` flag on AND an empty `pane_id` for a matched agent (could not confirm herdr emits empty pane_id for live agents — the socket driver mapping is out of this diff), so informational only. A guard requiring a non-empty pane component before choosing the socket path would harden it.

Issue #1629, opened 2026-07-11
Plan-gate row CTAs: re-tier plan-rework + Retry-CI for ci-red (deferred from #1610)


scoop:
Follow-ups deferred from the PR that closed the first three items of #1610 (plan-gate row CTA follow-ups). That PR shipped #1 (idle self-heal), #3 (pg-stalled re-tone), and the Answer CTA half of #4. These two are intentionally deferred here because each carries blast radius or missing data beyond the row surface:

## 1. Re-tier `plan-rework` by whose-turn-it-is (#1610 item #2)
`SIGNAL_TIER["plan-rework"] = 1` (CRITICAL / blocked-on-operator) even when the row now reads "Agent is revising the plan" — so the Rundown still ranks that session a Tier-1 operator decision and renders its `why` as "your call". Re-tiering by whose-turn-it-is has **rundown + digest + push** blast radius (`src/rundown-core.ts` `SIGNAL_TIER` / `classifyAttention`, `src/herd-digest.ts`, `explainHold`), so it needs its own change with those consumers re-verified.

## 2. `Retry CI` CTA for `ci-red` (#1610 item #4, second half)
The originating issue listed a `Retry CI` row action for `ci-red`. Deferred because the row does **not** have the failed workflow run id: the `ci-red` hold carries only `{ pr }` (`GitState` has no `runId`), while `rerunWorkflowRun(repoPath, runId, failedOnly)` needs one. A true one-click retry therefore requires an extra run-resolution fetch (find the PR head's latest failed run, then rerun) — a real async CTA path, and entirely outside the plan-gate phase the row classifier currently handles. `ActionRunRow.svelte` already has the rerun plumbing to mirror.

Refs #1610.

scoop:
✅ PR #1652 merged.

Issue #1561, opened 2026-07-08
feat: surface each session's block reason (/why) proactively on the row + a one-click resume action


Gambastudio:
## Context

`/why` (#1008, shipped) already consolidated the "why is this session parked/blocked/gate-held" explanation into a single trace. This proposal builds on it — two gaps remain **in practice**:

1. **The reason isn't surfaced proactively at the point of attention.** Observed live: a session sat `idle` on its row with **no visible reason**. The real state — parked in the **plan gate**, where the reviewer had requested changes on `.shepherd-plan.md`, the agent revised, and it was waiting for re-review — was only discoverable by reading the signal/terminal. So either `/why` is pull-only (you must invoke it) or this "idle after plan revision, awaiting re-review" state isn't reflected on the row/status.
2. **No action from that view.** Even once you know why, you then navigate into tabs to find the control that unblocks it — "click around everywhere".

## Proposal

On any non-progressing session, **inline on the row/header**:

- **Proactively show the `/why` reason** (reuse #1008's trace) — no command needed. Turn a bare "idle" into "idle — plan gate, awaiting re-review (5 change-requests addressed)".
- **Add the matching one-click action** to unblock from there: *Approve plan / re-review* · *Answer the question* · *Confirm manual steps* · *Resume/continue* · *Retry CI*. #1008 explains **why**; this adds **act here**, so the operator doesn't hunt across tabs.

## Why

The data already exists (#1008 trace; signals carry the reason). The missing piece is **proactive per-session surfacing + an actionable CTA** at the row. It turns "idle — why?? click around" into "here's why, and here's the button".

## Relation

Builds on **#1008** (`/why` explain-why — done) by surfacing it proactively and making it actionable. Complements **#1447** (surface the hold reason when a train stalls) and Rundown/triage (herd-wide "what needs a human"). This is the **per-session "why + resume-from-here"** at the row level.

scoop:
✅ PR #1611 merged.

PR #1631, opened 2026-07-11
feat(ui): plan-gate row CTA follow-ups — idle self-heal, final-round badge, answer CTA (#1610)

### LINKED ISSUE #1610: Plan-gate row CTAs: follow-ups deferred from #1561
Deferred from #1561 (per-session "why + resume-from-here" on the row). The shipped feature surfaces the plan-gate reason + one-click action on each row; these are the recorded, intentional deferrals:

## Durability — the more important fix
- **`consider()` isn't re-driven on the `idle` edge.** `onStatusIndependent` (`src/index.ts`) fires `planGate.consider()` only on `status === "done"`, and `mapState` (`src/herdr.ts`) falls through to `"idle"` for anything herdr doesn't call `done`. So a plan-gate session that settles to `idle` after the agent revises parks indefinitely — the "awaiting re-review" state #1561 now surfaces. Adding a plan-file watcher or re-driving `consider()` on the settle edge would make the row's `Re-review` button unnecessary (the loop would self-heal) and also closes the `awaiting-rereview` temporal window (steer→running latency where the row shows the button while the agent is already revising).

## Row polish
- **`plan-rework` tier vs. copy.** `SIGNAL_TIER["plan-rework"] = 1` (CRITICAL, blocked-on-operator) even when the row now reads "Agent is revising the plan" — so the Rundown still ranks that session a Tier-1 operator decision and renders its `why` as "your call". Re-tiering by whose-turn-it-is has rundown/digest/push blast radius; deferred.
- **Amber `pg-stalled` chip during the final round.** At cap + running + `finalRoundPending`, the row reads "Agent is revising (3/3)" beside an amber `CHANGES 3/3` chip (its tooltip already reads `changes`, not `changesStalled`). Re-toning `pg-stalled` on `planStallStatus === "final"` touches styling shared with the Viewport top bar; deferred.

## CTA scope extensions (the classifier is a pure ordered table — each is an additive branch)
- **`Answer` for `autopilot-paused` / `blocked-yes-no`** — the most common bare-idle blocks. Currently `rowState` passes these server holds through with no CTA (plan-gate scope only).
- **`Retry CI` for `ci-red`.** From #1561's requested action list; not delivered (ci-red needs an open PR, outside the planning phase the row-classifier handles).

See `ui/src/lib/hold-row.ts` (`rowState` classifier) and the design notes in the #1561 PR.

scoop:
Implements three of the four deferrals recorded in #1610 (plan-gate row CTA follow-ups). The fourth item and one half of another are moved to #1629.

### #1 — Durability: re-drive `consider()` on the idle settle edge (server)
`onStatusIndependent` fired `planGate.consider()` only on `status === "done"`, so a plan-gate session that revised its plan and settled to **idle** (herdr's default) parked indefinitely in `awaiting-rereview`. Now a new pure predicate `shouldConsiderOnSettle(status, planPhase, priorDecision)` also re-drives on the `idle` edge — but **only when the prior gate decision is `changes_requested`** (the revise loop). That scoping deliberately excludes:
- the **pre-first-draft** window (no gate ⇒ partial authoring idles can't spawn reviews and exhaust the adversarial cap — hash-dedup does *not* protect there, each partial hashes differently), and
- **`error`** gates (`consider()` never dedups those, so idle-firing would spam re-reviews + `stall` signals).

`done` still fires unconditionally in planning; `mapState` is untouched.

### #3 — Badge: don't tone `pg-stalled` during a genuine final round (UI)
`PlanGateBadge`'s `stalled` used a chip-only heuristic (`round >= cap`) that ignored `finalRoundPending`, so the badge showed amber "stalled" toning + the stall-recovery menu while the agent was *actively revising* the final round. Replaced with a pure `planGateStalledNow(session, gate, reviewing, now)` = `canShowPlanStallActions(...) && planStallStatus(gate, now) === "stalled"`, driven by the shared reactive clock; toning, menu, and tooltip now agree. Deliberate consequence: **running ⇒ not stalled** everywhere (consistent with `canShowPlanStallActions`).

### #4 (Answer half) — "Answer" CTA for autopilot-paused / blocked-yes-no rows (UI)
The row classifier was plan-gate-scoped (R1 sent every non-planning row to `passthrough`, no CTA). Added a strictly-additive `server-answer` branch in R1's **non-planning arm** for `autopilot-paused` / `blocked-yes-no` holds, with an **Answer** CTA (label reuses `hold_cta_answer`; internal `reply` kind) that opens the session (`onselect(id)`) so the operator can reply in the terminal. Every planning-phase rule (R2–R12) stays byte-identical.

## Scope boundary
#1's idle self-heal fixes only the **indefinite** park. It does **not** close the brief steer→running window where the row shows `Re-review` while the agent is already revising (that edge is `changes_requested → running`, no settle trigger, and would dedup on the unchanged pre-revision plan). The manual `Re-review` button stays useful and is not removed. A watcher/settle-debounce to close that window is out of scope.

## Deferred to #1629
- #1610 item **#2** — re-tier `plan-rework` by whose-turn-it-is (rundown + digest + push blast radius).
- #1610 item **#4 (Retry-CI half)** — a `Retry CI` CTA for `ci-red`; the row lacks the failed workflow `runId` (the hold carries only `{ pr }`), so it needs an extra run-resolution fetch outside plan-gate scope.

## Verification
- Root `bun test ./test`: 6393 pass. UI `bun run check`: 0 errors. UI node vitest: 2130 pass (new predicate + classifier + badge cases). `bun run check:i18n`: parity (2805 keys/locale). New `UnitRow.browser.test.ts` case is CI-verified (playwright).
- PR-hygiene gates green: branch-hygiene (linear off main), feature-catalog (What's-New entry `v1.43.0-hold-row-answer.ts`), glossary, announcement-versions, fallow audit.

Closes #1610.

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -328,10 +328,10 @@
       {
         "files": ["ui/src/lib/hold-row.ts"],
         "functions": ["rowState"],
-        "maxCyclomatic": 22,
-        "maxCognitive": 19,
+        "maxCyclomatic": 24,
+        "maxCognitive": 22,
         "maxCrap": 1000,
-        "reason": "Single ordered first-match classifier for the plan-gate row (#1561): one branch per rule (R1-R12), deliberately kept as a flat, top-to-bottom precedence chain so the exhaustively-tested rule order stays auditable at a glance. Cyclomatic is rule-count-driven, not nested logic (cognitive 19). Collapsing it into a data-driven rule table would hide the precedence the whole design turns on. Same envelope as src/drain.ts doSpawn (22/19)."
+        "reason": "Single ordered first-match classifier for the plan-gate row (#1561; +R1a for the non-planning Answer CTA, #1610): one branch per rule (R1a, R1-R12), deliberately kept as a flat, top-to-bottom precedence chain so the exhaustively-tested rule order stays auditable at a glance. Cyclomatic is rule-count-driven, not nested logic. Collapsing it into a data-driven rule table would hide the precedence the whole design turns on."
       },
       {
         "files": ["ui/src/lib/components/UnitRow.svelte"],
--- a/src/index.ts
+++ b/src/index.ts
@@ -82,7 +82,7 @@ import { Presence } from "./presence";
 import { ReviewService } from "./review";
 import { StandalonePrCriticService } from "./standalone-critic";
 import { createIssueLogger } from "./issue-log";
-import { PlanGateService } from "./plan-gate";
+import { PlanGateService, shouldConsiderOnSettle } from "./plan-gate";
 import { AutopilotService } from "./autopilot";
 import { DrainService } from "./drain";
 import { SessionRouter, type SessionConsumer } from "./session-router";
@@ -1646,9 +1646,12 @@ const sessionRouter = new SessionRouter(
     // planning-phase session that just settled likely finished writing .shepherd-plan.md → kick the
     // adversarial review. Fires BEFORE the awaited consumer chain so a slow/hanging drain pump can
     // never starve it (#1193); consider() is self-contained and claims its slot synchronously.
+    // Also re-fires on the idle settle edge for the revise loop (a session that reworks its plan
+    // after `changes_requested` settles to idle, not done) — see shouldConsiderOnSettle (#1610).
     onStatusIndependent: (change) => {
       const sess = change.snapshot.session;
-      if (change.status === "done" && sess.planPhase === "planning")
+      const priorDecision = store.getPlanGate(sess.id)?.decision;
+      if (shouldConsiderOnSettle(change.status, sess.planPhase, priorDecision))
         void planGate.consider(sess).catch((err) => console.warn("[plan-gate] consider:", err));
     },
   },
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -29,6 +29,23 @@ import { fenceUntrusted } from "./untrusted";
  *  relays this so the UI can distinguish a no-op from a real error or an unusable `.shepherd-plan.md`. */
 export type PlanReviewTrigger = "started" | "skipped" | "plan-unavailable" | "error";
 
+/** Whether a settle edge should re-drive `consider()`. `done` always re-considers (first
+ *  draft + any settle); `idle` re-considers ONLY when a prior gate already requested changes
+ *  — the revise loop. This excludes the pre-first-draft window (no gate ⇒ `undefined`) so
+ *  partial authoring idles can't exhaust the adversarial cap, and excludes `error` gates
+ *  (consider() never dedups those, so idle-firing would spam re-reviews). `approved` gates are
+ *  excluded too (consider() would skip them anyway). See issue #1610. */
+export function shouldConsiderOnSettle(
+  status: string,
+  planPhase: Session["planPhase"],
+  priorDecision: PlanDecision | undefined,
+): boolean {
+  if (planPhase !== "planning") return false;
+  if (status === "done") return true;
+  if (status === "idle") return priorDecision === "changes_requested";
+  return false;
+}
+
 /** The plan the planning agent writes in its LIVE session worktree; the reviewer reads its text. */
 const PLAN_FILE = ".shepherd-plan.md";
 
--- a/test/plan-gate-settle.test.ts
+++ b/test/plan-gate-settle.test.ts
@@ -1,0 +1,19 @@
+import { expect, test } from "bun:test";
+import { shouldConsiderOnSettle } from "../src/plan-gate";
+import type { Session } from "../src/types";
+
+test("shouldConsiderOnSettle truth table", () => {
+  const planning: Session["planPhase"] = "planning";
+  const executing: Session["planPhase"] = "executing";
+
+  expect(shouldConsiderOnSettle("done", planning, undefined)).toBe(true);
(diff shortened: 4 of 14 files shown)


scoop (commented):
## Summary

This PR implements three of the recorded #1561 deferrals: (1) re-drive `consider()` on the idle settle edge for the revise loop, (2) only tone the plan-gate badge amber/stalled when it is *genuinely* stuck (not during a live final round), and (3) an inline `Answer` CTA for non-plan `autopilot-paused` / `blocked-yes-no` holds. The two remaining deferrals the issue lists (re-tiering `plan-rework`, `Retry CI` for `ci-red`) are explicitly out of scope and correctly untouched. I found no blocking defects.

## What I verified

**Backend — `shouldConsiderOnSettle` (src/plan-gate.ts:36-46, wired at src/index.ts:1653-1655)**
- No regression on the pre-existing `done` path: original guard was `status === "done" && planPhase === "planning"`; the new function returns `true` for exactly that (planning + done, any prior decision) and adds only `idle + changes_requested`. Verified against the diff and the prior form at src/index.ts:1654.
- `store.getPlanGate` is real and returns `PlanGate | null`; `?.decision` → `undefined` feeds the `PlanDecision | undefined` param cleanly. Confirmed the method exists and is used widely (src/plan-gate.ts:289, src/server.ts:2333, src/poller.ts:1213).
- `PlanDecision = "approved" | "changes_requested" | "error"` — the three literals the function compares against all exist (src/types.ts:454). Import is from `./types` (src/plan-gate.ts:8), which is the backend catalog that defines it.
- Safety of the new idle re-fire: it is edge-triggered (poller fires `onChange` only when `idChanged || status !== s.status || agentStatus !== lastState`, src/poller.ts:588), and `consider()` self-dedups an unchanged plan (`prior.planHash === planHash && decision !== "error"` → skipped, src/plan-gate.ts:303) plus an inflight/starting guard (src/plan-gate.ts:288). So an idle session that hasn't revised re-considers to a no-op; a revised plan re-reviews — exactly the intended loop. The comment's rationale for excluding `error` (never deduped → would spam) is confirmed against src/plan-gate.ts:303.
- Truth-table test (test/plan-gate-settle.test.ts) matches the implementation branch-for-branch.

**UI — genuine-stall predicate (ui/src/lib/components/plan-gate-badge.ts:85-96)**
- `planGateStalledNow = gate != null && canShowPlanStallActions(...) && planStallStatus(gate, now) === "stalled"`. `canShowPlanStallActions` requires `status !== "running"`, `!reviewing`, `changes_requested`, `round >= cap` (plan-gate-badge.ts:66-78); since `round >= cap`, `planStallStatus` can only return `"final"` or `"stalled"` (never `"round"`, verified against plan-status.ts:28-32). So the predicate narrows the old behavior by exactly one case — a fresh, not-timed-out final round — which is the deferral being fixed. `finalRoundPending` and `updatedAt` both exist on the UI `PlanGate` (ui/src/lib/types.ts:402,406).
- Old `planGateStalled(chip)` export was removed; confirmed **zero** remaining references anywhere (grep clean). Its two consumers in PlanGateBadge.svelte (`stalled`, `stalledActionsVisible`) were both repointed to the new predicate (PlanGateBadge.svelte:38-39). `clock.current` is a real reactive 30s tick (ui/src/lib/now.svelte.ts), and the `$lib/now.svelte` import form matches 11 other call sites.

**UI — `server-answer` state + `reply` action (ui/src/lib/hold-row.ts)**
- R1a lives only in the non-planning arm (hold-row.ts:60-63); the planning rules R2–R12 are unaffected, and test 21 confirms a planning row with `autopilot-paused` still falls through to `passthrough`.
- `ANSWERABLE` codes `autopilot-paused` / `blocked-yes-no` are valid `HoldCode`s (ui/src/lib/types.ts:234,236) and are actually emitted server-side (src/hold.ts:54,57; src/rundown-core.ts:242), and the `hold` prop is supplied for every session via `holdFor` (Herd.svelte:256 → HerdGroup.svelte:95), so the CTA is reachable in production, not just tests.
- `LINE` and `ACTION` maps both gain the `server-answer` key (hold-row.ts:102,155), keeping them total over `RowState`. `reply` routes to `onselect(session.id)` (UnitRow.svelte:323); `onselect` is a required prop `(id: string) => void` (UnitRow.svelte:74) used identically elsewhere.

**i18n parity** — all added keys exist in BOTH catalogs with real translations: `feat_hold_row_answer_title`/`_body` (en.json:626-627, de.json:626-627) and `hold_cta_answer_reply_title` (en.json:2803, de.json:2803). The reused label `hold_cta_answer` exists in both (en.json:2798, de.json:2798).

**Feature announcement** — `v1.43.0-hold-row-answer.ts` is auto-registered via `import.meta.glob` (feature-announcements.ts:47-50); no index edit needed. Its `titleKey`/`bodyKey` resolve to the keys added above, and `sinceVersion: "1.43.0"` matches the other in-development v1.43.0 entries.

- R1a returns `server-answer` regardless of `session.status`, whereas the planning-arm `question` state requires `parked` (hold-row.ts:76). In practice these holds are only present while a session is parked/blocked, and the `reply` action merely opens the session, so an occasional stale-hold render is harmless. Noting only for awareness.

## Limitations

- `src/plan-status.ts` is declared a byte-identical logic mirror of `ui/src/lib/plan-status.ts`; only the UI copy is exercised by this diff and neither mirror file is in the diff, so I did not diff the two.

PR #1633, opened 2026-07-11
fix(build-queue): arm reconcile gate on 1Hz running events, not the 15s sweep (#1617)

### LINKED ISSUE #1617: build-queue reconcile backstop never fires for Go-released executing sessions despite settled-idle windows

scoop:
Fixes #1617.

## Problem

The build-queue **reconcile backstop** never fired for some Go-released `executing` sessions with an approved, all-pending queue — they drifted to `0/N` and never got the settled-idle nudge.

**Provable defect (hypothesis a):** the `sawRunning` evidence gate armed *only* when the reminder's own **15 s** sweep sampled `status === "running"`. But `status` is `mapState(herdr.agent_status)` — a **point-in-time level** polled at 1 Hz with no "recently-active" latch. A herdr-`working` burst that starts and finishes *between two 15 s sweeps* is never sampled, so `sawRunning` never arms and a bursty drifted queue never nudges.

## Honest diagnosis (read the issue's A/B house rule)

Read-only forensic on the four reported sessions (live DB + server log): all were `executing`, `haltReason=NULL` (no quota/error halt), all ran real work (492–918 tool-activity lines), all ended `lastState=done`. **The per-window `idle`/`blocked`/`done` split is unrecoverable** — worktrees + transcripts were removed at archival and the DB keeps no status history.

So:
- Hypothesis **(a)** is a real *latent* defect but **not grounded** as the cause for these four.
- It is **implausible for TASK-1472** (~90 min executing activity would have been sampled running); its ~11 h tail was likely `done` (hypothesis **e**), which the `status!=="idle"` guard **correctly** suppresses.
- **This fix may be a no-op for the reported population if their windows were non-idle (e).** It closes the (a) race for whichever sessions it bit; it deliberately does not (and must not) wake a `done`/`blocked` session.

## Fix

Arm on **observed-running** (preserving the worked-vs-never-started discriminator), but sample it at the poller's real **1 Hz** cadence instead of the reminder's coarse 15 s sweep:

- Subscribe to `session:status` in `index.ts`; on a `running` event for an approved, non-empty queue **outside the plan gate**, call the new `BuildQueueReminderService.markRan(id)`.
- The existing 15 s sweep arming stays as a harmless fallback for sustained running.

A never-started session emits no `running` event → never armed → never nudged, so `RECONCILE_STEER` (whose text assumes work happened) is never mis-sent. **No message change.**

## Forward instrumentation (production falsifiability)

A read-only, **deduped** sweep diagnostic for executing+drifted sessions:

```
[build-queue] reconcile-sweep id=… status=idle planPhase=executing sawRunning=false idleSinceMs=1001 readyToNudge=false
```

On the next live drift this resolves **(a)** (`sawRunning=false` while working) vs **(e)** (`status` never `idle`) and confirms the nudge lands — the surviving trace for the four could not. Purely observational; no control flow depends on it.

## Tests / verification (service-level A/B)

`test/build-queue-reminder.test.ts` (18 pass, 0 fail):
- **Burst-missed-by-sweep A/B:** sweep-only (pre-fix path) → **0** steers; then `markRan` (poller 1 Hz event) → **exactly 1** `RECONCILE_STEER`.
- **Boundary:** armed but flapping into `blocked` (resets `idleSince`) → **0** (honest scope).
- **Never-started discriminator:** executing+drifted, never observed running → **0**, and `RECONCILE_STEER` never emitted.

Full root suite `bun test ./test`: **6507 pass / 0 fail**. `bun run lint`, `bun run typecheck`, `fallow audit`: clean.

## Scope

In scope: `idle` executing drifted sessions **observed running** → one nudge/episode. Out of scope by design: `blocked` (resumable — own subsystems), `done` (badge PR's job), never-observed-running, and non-gated (`planPhase=null`, unchanged).

### DIFF
--- a/src/build-queue-reminder.ts
+++ b/src/build-queue-reminder.ts
@@ -53,7 +53,9 @@ interface DebounceEntry {
   idleSince: number | null;
   /** true once a reminder fired in the current idle episode (reset when it re-activates). */
   firedThisEpisode: boolean;
-  /** true once we've observed the session actually running since we began tracking it. */
+  /** true once we've observed the session actually running since we began tracking it — set
+   *  either by the 15s sweep sampling `running`, or (reliably) by `markRan`, fed from the poller's
+   *  1 Hz `session:status` running transitions so a sub-15s burst can't slip the gate (#1617). */
   sawRunning: boolean;
   /** lifetime reminders sent to this session (runaway guard). */
   nudgeCount: number;
@@ -73,6 +75,9 @@ export class BuildQueueReminderService {
   private idleThresholdMs: number;
   private maxNudges: number;
   private debounce = new Map<string, DebounceEntry>();
+  /** Per-session last-logged diagnostic tuple, so the #1617 sweep log dedupes on discrete-state
+   *  transitions instead of spamming every 15s tick. Pruned alongside the debounce entry. */
+  private lastDiag = new Map<string, string>();
   /** True while a sweep is in flight. Guards re-entrancy now that `sweep` awaits its steer
    *  (#1567) — see the drop-the-tick rationale on `sweep`. */
   private sweeping = false;
@@ -86,6 +91,21 @@ export class BuildQueueReminderService {
   /** Drop a session's debounce state (call on archive). */
   forget(id: string): void {
     this.debounce.delete(id);
+    this.lastDiag.delete(id);
+  }
+
+  /**
+   * Arm the evidence gate from an observed `running` transition — the RELIABLE arming path, fed by
+   * the poller's 1 Hz `session:status` events (wired in index.ts). The sweep's own 15s sample can
+   * miss a herdr-`working` burst that starts and finishes between ticks — `status` is a
+   * point-in-time level, not a "recently-active" latch (#1617) — so a bursty agent would never set
+   * `sawRunning` via the sweep alone. The caller gates this to a `running` event on an approved,
+   * non-empty queue outside the plan gate (mirroring the sweep's own guards), so here we just record
+   * the evidence. Arming ONLY on observed-running preserves the worked-vs-never-started
+   * discriminator: a never-started session emits no `running` event, so it is never nudged.
+   */
+  markRan(id: string): void {
+    this.entry(id).sawRunning = true;
   }
 
   private entry(id: string): DebounceEntry {
@@ -133,6 +153,9 @@ export class BuildQueueReminderService {
       for (const id of [...this.debounce.keys()]) {
         if (!live.has(id)) this.debounce.delete(id);
       }
+      for (const id of [...this.lastDiag.keys()]) {
+        if (!live.has(id)) this.lastDiag.delete(id);
+      }
     } finally {
       this.sweeping = false; // a thrown store error / rejected steer must not wedge the sweep off
     }
@@ -149,6 +172,7 @@ export class BuildQueueReminderService {
     // No approved queue yet (or none authored) → nothing to reconcile; drop any state.
     if (!q.approved || q.steps.length === 0) {
       this.debounce.delete(id);
+      this.lastDiag.delete(id);
       return;
     }
     const e = this.entry(id);
@@ -157,14 +181,18 @@ export class BuildQueueReminderService {
     // all-pending — no execution step has begun, so all-pending is correct, not drift. Return
     // BEFORE the status switch so planning-phase running never sets sawRunning (readyToNudge
     // ANDs on sawRunning, so this alone suppresses the nudge) and idle time can't accrue toward
-    // a post-gate nudge. NOTE: execution CAN occur while still "planning" (operator manually
-    // steers instead of clicking Go); we suppress there too by design — the PR-open flip to
-    // "executing" (service.advanceToExecutionOnPr) is the deliberate re-arm point, after which
-    // a fresh execution-running tick sets sawRunning and genuine drift nudges normally.
+    // a post-gate nudge. The same guard is mirrored on the markRan arming path (index.ts skips
+    // running events while planning). NOTE: execution CAN occur while still "planning" (operator
+    // manually steers instead of clicking Go); we suppress there too by design — the PR-open flip
+    // to "executing" (service.advanceToExecutionOnPr) is the deliberate re-arm point, after which
+    // a fresh execution-running transition (via markRan or the sweep) sets sawRunning and genuine
+    // drift nudges normally.
     if (planPhase === "planning") {
       this.resetEpisode(e);
       return;
     }
+
+    this.logDiag(id, status, planPhase, e, q, now);
 
     // Only `idle` is eligible. `running` records work-seen; everything else (done/blocked/
     // archived) is skipped without steering. All non-idle states reset the idle episode.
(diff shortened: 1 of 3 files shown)


scoop (commented):
## Summary

The PR adds a second, reliable arming path for the reconcile backstop's `sawRunning` evidence gate: `markRan()`, fed from the poller's 1 Hz `session:status` running transitions (index.ts), so a herdr-`working` burst that starts and finishes between two coarse 15s sweeps can no longer slip the gate. It also adds a read-only, deduped diagnostic (`logDiag`) to distinguish hypothesis (a) vs (e) on the next live occurrence. This directly targets the primary surviving hypothesis (a) — `sawRunning` never set — while the diagnostic keeps (e) resolvable. Well-scoped and internally consistent.

## Verification performed

- **Guard parity between the two arming paths.** The sweep sets `sawRunning` only after passing the approved/non-empty check (`src/build-queue-reminder.ts:173`) and the plan gate (`src/build-queue-reminder.ts:190`), then on a `running` tick (`src/build-queue-reminder.ts:200`). The new `markRan` caller mirrors both: `!q.approved || q.steps.length === 0` and `s.planPhase === "planning"` guards, arming only on `status === "running"` (`src/index.ts:967-972`). The mirroring is exact, so the worked-vs-never-started discriminator and the plan-gate suppression are preserved on the event path. `markRan` itself just records the evidence (`src/build-queue-reminder.ts:107-109`).
- **Event contract.** `session:status` is emitted with `{id, status}` (`src/index.ts:771,773,777`); `status` legitimately carries `"running"` (an existing sibling subscriber already branches on it at `src/index.ts:953-955`). `onChange` fires on transition/sub-transition edges at the 1s poll cadence (`src/poller.ts:588-594`), confirming the "1 Hz running transition" premise that catches sub-15s bursts.
- **Multiple subscribers.** `EventHub` stores listeners in a `Set` and fans out to all (`src/events.ts:4-11`), so adding a third `session:status` subscriber is fine.
- **Store surface.** `store.get` (`src/store.ts:1904`) and `store.getBuildQueue` (`src/store.ts:4168`) exist and return the shapes the handler destructures.
- **`lastDiag` lifecycle.** Pruned everywhere `debounce` is: on `forget` (`src/build-queue-reminder.ts:94`), on the no-approved-queue drop (`src/build-queue-reminder.ts:175`), and in the non-live sweep cleanup (`src/build-queue-reminder.ts:156-158`). `forget` is wired to `session:archived` (`src/index.ts:1456`). No unbounded growth.
- **Diagnostic is inert.** `logDiag` (`src/build-queue-reminder.ts:226-244`) gates on `executing && isQueueDrifted`, dedupes on a discrete-state tuple, and drives no control flow.
- **Tests match the API.** The three new tests (`test/build-queue-reminder.test.ts:184-235`) call `markRan("S")` (matching the public method) and correctly exercise the burst-race, blocked-flap-resets-settle boundary, and never-started discriminator; traced each against `considerSession`/`readyToNudge` and they hold.

- The `console.log` diagnostic is intentional instrumentation tied to #1617 and is deduped per session on state transitions, so it won't spam a parked session. Worth removing in a follow-up once (a) vs (e) is pinned live, but it is deliberate here.
- If a session ever reverted `executing`→`planning`, a stale `lastDiag` entry would linger (the planning branch resets the episode but doesn't clear `lastDiag`); harmless — at most it suppresses one diagnostic log line, and it never affects the nudge decision.

Issue #1622, opened 2026-07-11
chore(herdr): migrate usage-probe off node-pty → terminal session control, then remove node-pty


scoop:
## Context

PR #1620 (part of #1529) moved the **browser** live-terminal off `node-pty` onto herdr's native `terminal session control` stream (`src/socket-pty-bridge.ts`). It deliberately left `node-pty` in place because **`src/usage-probe.ts` is a second, independent `node-pty` consumer** — it drives an ephemeral `claude` `/usage` scrape through the same `src/pty-attach.mjs` Node helper (it even hardcodes `"node"`, not `config.nodeBin`). So `node-pty` can't be removed until usage-probe is migrated too.

## Goal

Migrate `usage-probe.ts` off the `node-pty` helper onto the socket-terminal stream introduced in #1620, then remove `node-pty` entirely.

## Scope

- Replace the `pty-attach.mjs` spawn in `usage-probe.ts` (~lines 59-67, 116) with a `terminal session control` NDJSON stream (reuse the `SocketPtyBridge` transport / the same Bun-native approach), driving the ephemeral scrape agent without node-pty. Gate behind the same `SHEPHERD_HERDR_SOCKET` seam / fall back to today's path until proven.
- Once usage-probe no longer needs the helper, remove:
  - the `node-pty` dependency from `package.json`
  - `src/pty-attach.mjs`, `src/pty-demux.mjs`, `src/pty-bridge.ts` (the browser path is already on the socket bridge)
  - `config.nodeBin` + `src/node-bin.ts` + `scripts/fix-node-pty-perms.mjs`
  - the node-pty build toolchain (C/C++ + python3) in `deploy/install.sh`, `deploy/update.sh`, `deploy/provision.ts`
  - the associated tests (`test/pty-*.test.ts`, `test/fix-node-pty-perms.test.ts`, `test/node-bin.test.ts`) — rewrite/delete with the code they cover
- Verify the `/usage` scrape still produces a usable frame end-to-end.

## Notes

Not urgent — node-pty works fine for usage-probe today; the payoff is dropping the native-addon dependency and its deploy toolchain. `src/sandbox.ts` also binds a `node` for sandboxed agents via `resolveNodeBin` — confirm whether that use is independent before removing `nodeBin` wholesale.

Follow-up to #1620 / part of the broader #1529 herdr-native adoption.

PR #1637, opened 2026-07-11
feat(herdr): generate socket protocol types from api schema + drift/protocol gates (#1529)

scoop:
## What

Drains **opportunity #5** of the herdr native-socket tracker (#1529): generate TypeScript types from herdr's socket JSON-RPC schema (`herdr api schema --json`), wire them into the socket transport, and add drift + protocol gates.

**Compile-time only** — no runtime or on-the-wire behavior change. Driver selection and every socket call are byte-for-byte what they were; only the inline `request<{…}>` generics were replaced with schema-generated types. herdr's socket protocol is preview/unstable, so everything stays reversible and the CLI path is untouched.

The startup protocol gate already existed in `selectHerdrDriver` (pings, checks the supported set, falls back to CLI); this PR adds the **schema → types** half plus a drift gate.

## How

- **Vendored schema** (`src/generated/herdr-schema.json`) — the exact `herdr api schema --json` output (protocol 16). Re-vendor with `bun run gen:herdr-schema` (needs a live herdr).
- **Codegen** (`scripts/gen-herdr-types.ts`, `bun run gen:herdr-types`) → `src/generated/herdr-protocol.ts`: flattens the 5 schema sections under section-prefixed `$defs` (resolving cross-section name collisions), rewrites refs, runs `json-schema-to-typescript` once, and appends schema-walked maps: `HerdrParams` (all 80 methods → params, auto), `HerdrResult<M>`, and a curated `HERDR_METHOD_RESULT`.
- **Typed transport** — `HerdrSocketClient.request` is now a single strict generic `request<M extends HerdrMethod>(method, HerdrParams[M]): Promise<HerdrResult<M>>` (no untyped fallback — nothing outside real JSON-RPC methods calls it). `SocketHerdrDriver` call sites dropped their inline result generics.
- **Gates**:
  - `check:herdr-types` — regenerates the types from the committed vendored schema (**no herdr binary**) and fails on any drift; wired into the pre-push `gates` lane and the `ci.yml` verify job.
  - `test/herdr-schema.test.ts` — asserts the vendored `protocol` ∈ `HERDR_SOCKET_SUPPORTED_PROTOCOLS`, and that the curated map matches the `type` a **real** response carries (via a sanitized response manifest), not merely that the type exists.
- `src/generated/` is prettier- and fallow-ignored (generated surface).

## Key design decisions

- **The schema encodes `method → params` but NOT `method → result`** — results are a `oneOf` discriminated by a `type` const, and only 26/80 methods map by naming convention. So `HerdrParams` is fully auto-derived; `HERDR_METHOD_RESULT` is curated. It holds exactly the **7 methods whose result the driver actually reads** (`ping`, `agent.list`, `agent.read`, `agent.start`, `pane.process_info`, `tab.create`, `workspace.list`); methods whose result the driver discards (`agent.send/rename`, `tab.close/rename`, `pane.close`, `workspace.create`) are omitted and fall back to the full result union.
- **Map correctness is verified against real responses, not just existence.** `bun run gen:herdr-fixtures` captures a **sanitized manifest** (`test/fixtures/herdr-responses/manifest.json`) of each read-only method's observed `result.type` + top-level field names — never raw bodies (which would leak agent names, cwds, terminal text). The test cross-checks the curated map against it. Mutating methods aren't exercised (that would mutate the operator's live herdr) — they're documented per-entry.
- **Result/event types are open** (accept unknown fields) for forward-compat with a newer protocol-16 field; **params types are closed**.
- **Determinism** — `json-schema-to-typescript`'s internal formatter is disabled (`format:false`) and formatting is done with the repo's pinned prettier; both are version-pinned, so the freshness gate reproduces byte-identically in CI's bun runtime.

## Verification

- `bun run lint`, `bunx tsc --noEmit`, `bun test ./test` (**6516 pass / 0 fail**), `bun run check:herdr-types` — all green.
- PR-hygiene gates (branch hygiene, feature catalog, generated docs, glossary, announcement versions) and `fallow@2.100.0 audit --fail-on-issues` — all pass (0 issues).
- Gates demonstrated to bite: `check:herdr-types` fails on a tampered schema; the protocol test fails on an unsupported protocol; the round-trip test fails on a wrong curated mapping.

## Scope

Part of tracker #1529 (opportunity #5) — the tracker stays open; opportunities #3 (`events.subscribe`), #4 (`pane.report_agent`), #6 (remote), and node-pty removal (#1622) remain. The generated event/subscription types are emitted now so those land type-safe, but no consumer is wired here.

Server-only plumbing, no user-facing UX. `[no-feature-entry]`


scoop (commented):
## Summary

This PR implements opportunity #5 from the originating issue: generate TS types from `herdr api schema --json`, gate protocol version, and tighten the socket client/driver against the generated schema. It is a well-scoped type-safety + tooling change — no runtime/wire behavior changes — and stays behind the existing socket-driver seam (CLI fallback intact). I verified the correctness-relevant claims below and found nothing blocking.

## What I verified

- **`request<M>()` generic constraint is sound.** The new signature (`src/herdr-socket-client.ts:55`) types `params: HerdrParams[M]` and result `HerdrResult<M>`. Every one of the 12 driver call sites conforms to the generated param interfaces: `tab.create` → `RequestTabCreateParams` (herdr-protocol.ts:1737), `pane.close {pane_id}` → `RequestPaneTarget` (1565, uses `pane_id` not `target` — checked), `tab.close/tab.rename` → `RequestTabTarget`/`RequestTabRenameParams` (1773/1765), `agent.start` → `RequestAgentStartParams` (1255), `agent.read` → `RequestAgentReadParams` (1228), `pane.process_info {pane_id}` → `RequestPaneProcessInfoParams` (1420, optional). Since the old signature was `params: object` (accepted anything), tightening now type-checks all pre-existing call sites too — I confirmed none of them break.
- **`HerdrResult<M>` mapping resolves correctly.** `HERDR_METHOD_RESULT` (herdr-protocol.ts:2740-2748) maps each curated method to a result `type`, and `HerdrResult` Extracts the matching member of `SuccessResponseResponseResult`. Verified the driver's field reads against the extracted members: `tab.create` reads `.tab`/`.root_pane` → tab_created member has both (758-760); `agent.start` reads `.agent` → agent_started (774); `workspace.list` reads `.workspaces` → workspace_list (716); plus pane_read (859), pane_process_info (819), agent_list (779), pong (694).
- **`ping()` return-type change is safe.** New return is the pong union member (`protocol: number` required; `capabilities?` now optional/nullable). The only consumer, `selectHerdrDriver`, destructures `{ protocol }` (herdr-socket-driver.ts:259) — fine. Grepped for `.capabilities` reads of the ping result across `src/` — none.
- **Protocol gate is internally consistent.** `HERDR_PROTOCOL = 16` (herdr-protocol.ts:2653) matches `HERDR_SOCKET_SUPPORTED_PROTOCOLS = new Set([16])` (config.ts:78) and the vendored `schema.protocol`, asserted by test/herdr-schema.test.ts:52-55.
- **Fixture manifest ↔ curated map agree.** Every method in `test/fixtures/herdr-responses/manifest.json` (ping, agent.list, workspace.list, pane.process_info) exists in `HERDR_METHOD_RESULT` with the matching `type`, and each entry's `resultKeys` is a subset of the schema variant's properties (test asserts this at herdr-schema.test.ts:67-79).
- **Test-side changes.** `agent.read` test params `{ target, source: "visible" }` are valid (`RequestReadSource` includes "visible", herdr-protocol.ts:8). The `rawRequest` cast (test/herdr-socket-client.test.ts:47) is a justified, test-only bypass so the wire-transport tests can keep using synthetic method names; it doesn't touch production typing. The timeout test switching `"hang"`→`"ping"` still exercises the no-reply server path (test:173-179).
- **Tooling/CI wiring.** `check:herdr-types` regenerates from the vendored schema and `git diff --exit-code`s only `herdr-protocol.ts`; the generated dir is excluded from prettier (`.prettierignore`) and fallow (`.fallowrc.jsonc`), and json-schema-to-typescript is pinned exactly (15.0.4). The gen scripts read only committed files (types/fixtures) or a live herdr (schema/fixtures, manual) — consistent with their docstrings.

## Caveats / could-not-verify (not findings)

- `bun.lock` adds `lodash@4.18.1` (transitively via json-schema-to-typescript → lodash `^4.17.21`). I could not verify this resolution against the npm registry (ground truth not in this repo). Worth a glance from the author, since publicly lodash's 4.x line ends at 4.17.21; blast radius is dev-only (codegen devDependency, never shipped to runtime/browser). Not raised as a finding because it is unverifiable from the repo, not confirmed wrong.
- The `check:herdr-types` determinism gate depends on prettier (`^3.9.5`, caret) producing identical formatting in CI vs. locally; this is a standard freshness-gate tradeoff and the gate itself would surface any drift.

No findings block this PR. It satisfies the tracked task (opportunity #5: generate types + gate protocol version) and is free of the bugs I checked for.

Issue #1639, opened 2026-07-11
feat(herdr): socket-native terminal scroll (translate swipe/wheel → PageUp/PageDown), then flip SHEPHERD_HERDR_SOCKET_TERMINAL on


scoop:
## Context

Phase B of the mobile-scroll fix. Phase A (this PR's predecessor) landed a default-OFF sub-flag `SHEPHERD_HERDR_SOCKET_TERMINAL` that gates the interactive terminal back to node-pty, restoring native swipe/wheel scroll while `SHEPHERD_HERDR_SOCKET` is on. This issue builds socket-native scroll so the flag can be flipped on and node-pty eventually removed (part of #1622).

**Why the socket terminal can't scroll natively:** herdr's `terminal session control` streams a screen-diff/redraw protocol (`\x1b[2J` + absolute-position repaint, zero newlines, no `?1049`, mouse modes stripped). xterm builds no scrollback and reads `mouseTrackingMode==='none'`, so `agentOwnsScroll` is false and swipes/wheel reach nothing.

## Approach

Translate incremental swipe/wheel gestures into the app's own keyboard scroll levers — `PageUp`/`PageDown` (`\x1b[5~`/`\x1b[6~`) — over `terminal.input`, the same channel the shipping jump-to-top/bottom buttons already use (`\x1b[1;5F` / `\x1b[1;5H`). Chosen over mouse-report injection: keyboard keys are inert in a mouse-off child (no garbage) and need no fabricated/mode-matched xterm control bytes.

## ⚠ Open risk — prove the lever for the ACTUAL target apps first

A diagnostic spike proved the socket transport **forwards** PageUp/PageDown and the scroll **manifests in frames** — but only for `less` (a pager). Whether **Claude Code and Codex** scroll their transcripts on those keys over the re-rendered frame stream is **unproven** (a live-Claude probe was contaminated by spinner animation + takeover contention). 

**Start with a clean app-binding spike** on a throwaway `herdr agent start` Claude/Codex agent (not a live session — `terminal session control --takeover` fights Shepherd's own attach). Record which lever each honors and at what granularity. **If an agent honors no usable lever, do NOT flip the gate** — leave Phase A's node-pty gate on and file a herdr follow-up for a scroll-capable / raw-passthrough transport.

## Client change (gated on the spike passing)

[...]

PR #1640, opened 2026-07-11
fix(herdr): gate socket terminal behind opt-in flag so scrolling works

scoop:
## Problem

Enabling `SHEPHERD_HERDR_SOCKET=1` (#1529/#1620) routed the interactive terminal onto herdr's `terminal session control` stream — and **killed scrolling in the terminal on mobile** (and desktop wheel).

**Root cause (evidence-backed):** that stream is a screen-diff/**redraw** protocol, not the append-only PTY byte stream xterm.js needs. Decoding live frames from a running Claude pane on the reporting instance, every `full:true` frame is `\x1b[2J` (clear) + absolute cursor-positioned repaint of every cell — **zero newlines** (nothing scrolls off the top, so xterm's scrollback never grows), **no alternate-buffer switch** (`?1049` absent), and **the app's mouse-tracking modes are stripped** (`?1000h`/`?1006h` absent). So in `Viewport.svelte`, `agentOwnsScroll(term)` (`buffer.type==='alternate' || mouseTrackingMode!=='none'`) is false and swipes/wheel route into xterm's empty scrollback, reaching nothing. node-pty streams raw PTY bytes → xterm builds scrollback + sees mouse modes → scroll works.

## Fix (regression-first, Phase A)

Add a **default-OFF** `SHEPHERD_HERDR_SOCKET_TERMINAL` sub-flag that gates **only** the interactive terminal onto the socket. With it off (the default), a `SHEPHERD_HERDR_SOCKET=1` instance keeps the terminal on **node-pty** — restoring native-granularity swipe + wheel scroll immediately — while the socket driver still drives send/steer/browser/usage.

- `src/config.ts` — new `herdrSocketTerminal` flag (default OFF), documented.
- `src/server.ts` — `pickTerminalBridgeKind` now requires the sub-flag; threaded through at the call site. The existing `SocketPtyBridge`/`onFallback` machinery is untouched, just not selected.
- `test/socket-terminal-wiring.test.ts` — covers the new gate condition.

**Blast radius:** none for non-opt-in operators (`SHEPHERD_HERDR_SOCKET` is itself default-OFF). Opt-in operators get the terminal on a proven, already-shipping transport and keep the socket driver everywhere else. Reversible by config.

## Why a gate, not a socket-side fix, in this PR

herdr 0.7.3 has no raw/scrollback-capable terminal transport (`terminal attach` needs a TTY and panics on a pipe), so socket-native scroll requires new work with an unproven lever for the target apps. Rather than leave the reported regression unfixed behind that uncertainty, this PR stops the bleeding now; **socket-native scroll is the follow-up (#1639)**, after which the flag flips on and node-pty can be removed (part of #1529/#1622).

## Verification

- `bun run lint`, `bun run typecheck`, `bun test ./test` (6509 pass, 0 fail) — all green.
- Empirical root-cause + a diagnostic spike proving the socket transport forwards PageUp/PageDown and re-renders scroll (for `less`) informed the follow-up; Claude/Codex transcript-scroll over the socket remains the open risk tracked in #1639.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -551,6 +551,14 @@ export const config = {
   // preview-unstable (see HERDR_SOCKET_SUPPORTED_PROTOCOLS above), so this stays reversible
   // until the socket driver has soaked.
   herdrSocket: process.env.SHEPHERD_HERDR_SOCKET === "1",
+  // Sub-flag gating ONLY the interactive terminal onto herdr's socket `terminal session control`
+  // stream. Default-OFF interim gate: that stream is a screen-diff/redraw protocol, so xterm builds
+  // no scrollback and never sees the app's mouse mode — which kills mobile swipe + desktop wheel
+  // scrolling (the socket terminal renders, but can't be scrolled). Until socket-native scroll is
+  // built and proven for Claude+Codex (Phase B follow-up), keep the terminal on node-pty even when
+  // `herdrSocket` is on; the socket driver still drives send/steer/browser/usage. Flip on to soak
+  // socket-scroll once it lands.
+  herdrSocketTerminal: process.env.SHEPHERD_HERDR_SOCKET_TERMINAL === "1",
   // Push-based agent-info ingestion via Claude Code hooks (issue #704), additive on top
   // of polling, staged behind two default-off flags so each phase is independently reversible.
   // Phase 0: inject PostToolUse/PostToolUseFailure/Notification HTTP hooks into spawned
--- a/src/server.ts
+++ b/src/server.ts
@@ -6831,14 +6831,22 @@ export function pruneSocketTerminalFailures(
   for (const [id, t] of m) if (now - t >= ttl) m.delete(id);
 }
 
-/** Picks the socket bridge only when the driver is active, a pane target resolved, and this
- *  terminal hasn't recently failed over the socket path. */
+/** Picks the socket bridge only when the terminal sub-flag is on (interim gate — see
+ *  `config.herdrSocketTerminal`), the driver is active, a pane target resolved, and this terminal
+ *  hasn't recently failed over the socket path. With the sub-flag off the terminal stays on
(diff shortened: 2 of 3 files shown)


scoop (commented):
## Summary

The PR adds a default-off sub-flag `herdrSocketTerminal` (`SHEPHERD_HERDR_SOCKET_TERMINAL`) that gates *only* the interactive terminal onto the herdr socket stream. With the flag off (the default), `pickTerminalBridgeKind` returns `node-pty`, so the terminal keeps xterm scrollback and mouse-mode — restoring mobile swipe and desktop wheel scrolling that the socket screen-diff protocol had killed. The socket driver still handles send/steer/browser/usage. This directly and minimally addresses the task.

## Verification

- New flag defined at `src/config.ts:561` (`herdrSocketTerminal: process.env.SHEPHERD_HERDR_SOCKET_TERMINAL === "1"`) — verified name/casing consistent with usage.
- `pickTerminalBridgeKind` gains the `herdrSocketTerminal?` opt and ANDs it first in the guard (`src/server.ts:6838-6850`). The ternary condition coerces to boolean and the return type stays `"socket" | "node-pty"` — no type/logic regression versus the prior form (same truthy-string `paneTarget` pattern as before).
- The sole call site passes the flag through: `herdrSocketTerminal: config.herdrSocketTerminal` at `src/server.ts:6963` — verified this is the only caller via grep (`src/server.ts:6962`, tests aside).
- Tests updated for the new 4th condition: default-off case asserts `node-pty` (`test/socket-terminal-wiring.test.ts:141-153`), and the driver-off / no-paneTarget / recently-failed cases all now pass `herdrSocketTerminal: true` so they continue to isolate their intended condition (`test/socket-terminal-wiring.test.ts:154-186`).

All introduced identifiers resolve and are spelled consistently across config, server, and tests. No behavioral gap, no missing guard, no locale/contract concerns (change is not user-facing string or message-catalog related).

No blocking or non-blocking issues found.

Issue #1642, opened 2026-07-11
herdr terminal transport can't scroll (Codex honors no lever) — blocks socket-terminal flip + node-pty removal


scoop:
## Problem

herdr's `terminal session control` stream (the socket terminal transport, `SHEPHERD_HERDR_SOCKET_TERMINAL`) is a **screen-diff/redraw** protocol: a fixed re-rendered viewport with **no scrollback** and mouse modes stripped. So in socket mode there is nothing for the terminal emulator to scroll — scrolling can only happen if the **agent app itself** repaints a scrolled view in response to a keyboard lever sent over `terminal.input`.

The Phase B spike (#1639) proved on live agents (herdr 0.7.3) that this doesn't hold across our target apps:

| agent | PageUp | Shift+PageUp | Ctrl+Home | MouseWheelUp |
| --- | --- | --- | --- | --- |
| Claude Code | **scroll** (~7 lines/key) | — | — | scroll |
| **Codex** | — | — | — | — |

**Codex honors no keyboard/mouse lever** (9 tested: PageUp/PageDown, wheel-SGR, Shift+PageUp, Ctrl+Up, Up, Ctrl+B, Ctrl+U, Ctrl+Home, Home) — its off-screen transcript is unreachable in socket mode. So `SHEPHERD_HERDR_SOCKET_TERMINAL` cannot be flipped on without regressing Codex mobile/desktop scroll versus today's node-pty terminal.

Consequence: **node-pty removal (#1622) is blocked** — the terminal (and usage-probe) can't leave node-pty while the socket transport can't scroll a Codex transcript.

## Ask (herdr transport capability)

A terminal transport that preserves scroll, either:
- **scrollback-preserving** frames (expose enough history / a scroll-region the client can drive), or
- a **raw-passthrough / PTY-forwarding** mode (forward the child PTY bytes incl. alt-screen + mouse modes, so xterm behaves exactly as with node-pty).

Either unblocks flipping the terminal onto the socket and removing node-pty for **both** providers.

## Evidence / repro

- Matrix + method: `test/fixtures/terminal-control/scroll-binding-notes.md`
- Re-run after a herdr/agent upgrade: `bun scripts/verify-herdr-terminal.ts --scroll` (report-only; NOTICE fires if Codex gains a lever → flip may become viable)
- Wire contract: `test/fixtures/terminal-control/capture-notes.md`

## Notes

[...]

PR #1643, opened 2026-07-11
chore(herdr): conclude socket-scroll spike (Codex honors no lever) — don't flip

scoop:
## What this is — and isn't

Issue #1639 (Phase B) asked to **build** socket-native terminal scroll (swipe/wheel → PageUp/PageDown over `terminal.input`) and **flip `SHEPHERD_HERDR_SOCKET_TERMINAL` on**. It made that **explicitly contingent** on a HALTing spike proving the *actual target apps* (Claude Code **and** Codex) scroll their transcript on such a lever over herdr's frame stream.

I ran that spike live (herdr 0.7.3, throwaway `herdr agent start` agents). It came back **negative for Codex**, so per the issue's own gate the build + flip are **intentionally not shipped**. This PR delivers the issue's stated fallback: durable evidence + a guard + a follow-up.

## Spike result (A/B, live agents, reproduced & stable)

| agent | PageUp `\x1b[5~` | Shift+PageUp | Ctrl+Home | MouseWheelUp (SGR) |
| **Claude Code** | **scroll** (~7 lines/key; 135–150 → 112–128, renders own "Jump to bottom" hint) | — | —¹ | scroll² |
| **Codex** | — | — | — | — |

¹ Claude honors Ctrl+Home only from an already-scrolled state (probe restores to bottom each lever). ² Real but moot — the frame stream strips mouse modes, so a client can't ride the wheel anyway.

**Codex honors none** of 9 levers tested (PageUp/PageDown, wheel-SGR, Shift+PageUp, Ctrl+Up, Up, Ctrl+B, Ctrl+U, Ctrl+Home, Home).

## Why this means "don't flip"

In socket mode the frame stream has **no scrollback** — scroll needs the *app* to repaint on a lever. Claude does; Codex doesn't. So flipping `SHEPHERD_HERDR_SOCKET_TERMINAL` on would leave Codex sessions **unable to scroll** their transcript — a regression vs today's node-pty terminal, which scrolls natively for both providers. And node-pty removal (#1622) can't complete while Codex needs node-pty. herdr's frame-stream transport is **not yet good enough** to replace the node-pty terminal.

## Changes (small; no client/Viewport/flag change)

- **`scripts/verify-herdr-terminal.ts`** — opt-in `--scroll` app-binding matrix: spins up throwaway claude+codex agents, drives levers over `terminal.input`, observes via `agent read`. **Report-only, exits 0**; a NOTICE fires only if the two load-bearing facts change (Claude stops honoring PageUp, or Codex gains a lever → flip may become viable). Default drift-check path unchanged; never in CI.
- **`test/fixtures/terminal-control/scroll-binding-notes.md`** — reproduced matrix + method + conclusion.
- **`src/config.ts`** — `herdrSocketTerminal` comment records the finding + points at the follow-up. Flag stays **default-off** (Phase A gate unchanged).

## Follow-up & disposition

- **#1642** (internal tracking) requests a scrollback-preserving / raw-passthrough herdr terminal transport — the prerequisite for revisiting the flip and for #1622.
- **#1639 stays open**, blocked-by #1642 (the place to revisit build+flip if herdr gains a scroll-capable transport) — hence `Refs`, not `Fixes`.

Refs #1639, #1642, #1622. `[no-feature-entry]` (no user-facing UX ships).

## Test

- `bun run lint` ✅ · `bun run typecheck` ✅ · `bun test ./test/socket-terminal-wiring.test.ts ./test/terminal-transport-metrics.test.ts` ✅ (31 pass)
- `bun scripts/verify-herdr-terminal.ts --scroll` → reproduces the matrix, "load-bearing facts match", exit 0

### DIFF
--- a/scripts/verify-herdr-terminal.ts
+++ b/scripts/verify-herdr-terminal.ts
@@ -10,6 +10,17 @@
  * This is the herdr-upgrade check — it is intentionally NOT part of `bun test`
  * (it needs a live herdr daemon + a scratch pane). Run it after bumping herdr:
  *   bun scripts/verify-herdr-terminal.ts
+ *
+ * `--scroll` (or SHEPHERD_VERIFY_SCROLL=1) runs a second, heavier, opt-in mode: the app-binding
+ * scroll matrix (issue #1639). It spins up throwaway `claude` + `codex` agents, fills each with a
+ * tall transcript, and records which keyboard levers scroll it over `terminal.input`. REPORT-ONLY —
+ * prints a per-agent matrix and exits 0 (a NOTICE flags divergence from the recorded baseline in
+ * test/fixtures/terminal-control/scroll-binding-notes.md). It needs live agent auth + spends model
+ * tokens, so it is opt-in and never wired into CI. Background: socket-mode scroll needs the *app* to
+ * repaint on a lever (no scrollback in the frame stream); Claude honors PageUp, Codex honors none —
+ * which is why SHEPHERD_HERDR_SOCKET_TERMINAL stays default-off. Re-run after a herdr/agent upgrade
+ * to see whether that changed:
+ *   bun scripts/verify-herdr-terminal.ts --scroll
  */
 export {}; // make this a module so top-level await is allowed under tsc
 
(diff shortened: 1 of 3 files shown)


scoop (commented):
## Summary

This PR **concludes the socket-scroll spike** rather than flipping `SHEPHERD_HERDR_SOCKET_TERMINAL` on. That is the correct outcome per the originating issue's explicit **Open risk** directive ("If an agent honors no usable lever, do NOT flip the gate"): the app-binding spike found Claude honors PageUp but **Codex honors no lever**, so the gate stays default-off and the client-side `Viewport`/`terminal-transport` changes are (correctly) *not* made — they were gated on the spike passing, which it did not. So the PR satisfies the conditional task and does not regress anything (the flag value line `src/config.ts:570` is unchanged; only the surrounding comment was rewritten).

Changes reviewed:
- `scripts/verify-herdr-terminal.ts`: refactors the existing top-level drift check into `runDriftCheck()` (byte-for-byte equivalent logic — verified against the original at `scripts/verify-herdr-terminal.ts:134-204`) and adds an opt-in `--scroll` app-binding matrix behind `process.argv.includes("--scroll") || SHEPHERD_VERIFY_SCROLL===1` (line 449). Report-only, always exits 0.
- `src/config.ts:554-569`: comment-only rewrite documenting the spike result and the default-off rationale.
- `test/fixtures/terminal-control/scroll-binding-notes.md`: new evidence fixture. `capture-notes.md` it references exists (verified `test/fixtures/terminal-control/capture-notes.md`).

## Verification
- Refactor equivalence: the moved drift-check body (frame/input/resize/release/closed + gone-target assertions) is identical to the pre-diff top-level code; `problems`/`check` are the same module-level helpers at `scripts/verify-herdr-terminal.ts:56-59`, still consumed only on the drift path. No behavior change for the default (CI-run-manually) mode.
- Matrix identifiers: lever `name` values in `UP_LEVERS` (line 218-223) match the keys read in the NOTICE logic — `matrix.claude?.PageUp` (line 423) matches `{name:"PageUp"}` exactly; `leverNames` is derived from the same array (line 406). Baseline constants `BASELINE_CLAUDE_PAGEUP`/`BASELINE_CODEX_ANY_LEVER` (243-244) agree with the recorded matrix in `scroll-binding-notes.md` (claude PageUp=scroll, codex all `—`).
- `matrix.codex` is always populated (every row gets `matrix[r.provider]=row` at line 412-413, even on error → all-false), so `matrix.codex?.[n]` cannot throw.

## Non-blocking nit (see findings)
The `--scroll` summary treats an **errored** probe row as authoritative for the Codex baseline. If the Codex probe throws (returns `{error, results:[]}`), all its cells are `false`, so `codexAny===false===BASELINE_CODEX_ANY_LEVER` and no NOTICE fires — the final line still prints "Load-bearing facts match the recorded baseline: ... Codex honors none" (`scripts/verify-herdr-terminal.ts:442-444`) even though the probe *failed to run* rather than *confirming no lever*. The `⚠ error` is printed in the row (line 419) but the load-bearing conclusion doesn't consult `r.error`. Asymmetric with Claude, where an error flips `claudePageUp` to false and does correctly fire a NOTICE. Since this is a report-only diagnostic (not a CI gate) it's non-blocking, but it can give false confidence on the one fact the whole spike hinges on. Suggest: if any probed row for a load-bearing provider has `.error`, emit a NOTICE ("probe failed — result inconclusive") instead of silently counting it as a baseline match.

## Could not verify (external, not in repo)
- The herdr agent CLI response shapes assumed by the new `--scroll` code — `agent start` → `result.agent.terminal_id`/`tab_id` (line 247), `agent get` → `result.agent.agent_status` (261), `agent read` → `result.read.text` (269) — are herdr's own contract and aren't present in this repo. Assumed as written; the drift-check-mode shapes (`tab create` etc.) are the ones covered by the existing fixtures.

PR #1646, opened 2026-07-11
chore: scrub personal tailnet/node identifiers

scoop:
## What

Removes traces of a personal Tailscale identity that were committed into the repo:

- **Tailnet** `chicken-beardie` → `example`
- **Node** `backontop` → `agentnode`

## Why

`deploy/shepherd.service` hardcoded a real tailnet host (`shepherd.chicken-beardie.ts.net`) in `SHEPHERD_ALLOWED_HOSTS`, and the same identity leaked into doc-comment examples and test fixtures. A committed unit template shouldn't carry one operator's private hostname.

## Changes

- `deploy/shepherd.service`: drop the hardcoded host; leave localhost-only defaults + a comment pointing operators to add their own tailnet host via `~/.shepherd/env` (matches the README contract).
- `README.md`, `src/tailscale.ts`, `ui/src/lib/previewUrl.ts`: genericize example hostnames in comments.
- Test fixtures (`test/tailscale.test.ts`, `test/preview-config.test.ts`, `test/diagnostics.test.ts`, `test/egress-runner.test.ts`, `ui/src/lib/previewUrl.test.ts`): renamed to the neutral placeholders.

Pure rename/removal — no behavior change. `git grep` for the old identifiers now returns nothing.

## Verification

- Root: `bun test ./test/tailscale.test.ts ./test/preview-config.test.ts ./test/diagnostics.test.ts ./test/egress-runner.test.ts` → pass
- UI: `bun test src/lib/previewUrl.test.ts` → pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1645, opened 2026-07-11
deploy/shepherd.service hardcodes a maintainer's Tailscale tailnet host in SHEPHERD_ALLOWED_HOSTS (breaks downstream deploys with misleading "Preview is read-only")


kai-osthoff:
## Summary

`deploy/shepherd.service` — the committed systemd unit that operators copy to `~/.config/systemd/user/` — hardcodes a **specific maintainer's personal Tailscale tailnet host** in the CSRF origin allowlist:

```ini
# deploy/shepherd.service:25
Environment=SHEPHERD_ALLOWED_HOSTS=localhost,127.0.0.1,::1,[::1],shepherd.chicken-beardie.ts.net
```

`chicken-beardie.ts.net` is one operator's tailnet, not a placeholder. Anyone who follows the documented deploy path inherits an allowlist that trusts **someone else's** host and **not their own**.

## Impact (real, hit in practice)

`config.allowedOriginHosts` is derived **only** from `SHEPHERD_ALLOWED_HOSTS` (`src/config.ts:504` → `resolveAllowedOriginHosts`), which appends only the two capture-extension IDs. The node's own tailnet hostname **is** resolved at startup but is stored in `config.previewHost` (`src/config.ts:725`) for preview-URL targeting and is **not** added to the origin allowlist.

So for a downstream operator on their own tailnet (e.g. `*.long-tautara.ts.net`), every state-changing request from the HUD fails the origin guard:

- `checkOrigin` → `originAllowed(Origin, config.allowedOriginHosts, previewRange)` returns false (`src/server.ts:631`) → `403 {"error":"forbidden: origin not allowed"}`
- the client maps that specific 403 to `PreviewBlockedError` (`ui/src/lib/api.ts:114-115`) → surfaced as **"Preview is read-only. Open Shepherd directly (not through the live preview) to make changes."** (`error_preview_readonly`)

Concrete symptom: clicking **Update now** in the self-update modal (and any other mutation) silently 403s and shows the misleading "Preview is read-only" message, even though the operator is on the real HUD, not a preview. Diagnosing it requires reading the origin guard — the message points the user in the wrong direction (it blames "the live preview").

## Where `chicken-beardie` appears

| Location | Kind | Notes |
| --- | --- | --- |
| `deploy/shepherd.service:25` | **functional value** | the actual foot-gun — copied verbatim into operators' units |
| `src/tailscale.ts:21,48` | comments | example hostnames in doc comments |
[...]

PR #1649, opened 2026-07-11
fix: auto-allow node's own tailnet host + accurate origin-block copy (#1645 Fix 2+3)

scoop:
Closes the two behavioral fixes left open on #1645 after #1646 shipped the deploy-template scrub (Fix 1 + Fix 4). Both live on the CSRF origin-guard surface, so they ship together.

## Fix 2 — auto-allow the node's own host (the real functional fix)

The node's own tailnet hostname was resolved at startup into `config.previewHost` but **never** folded into `config.allowedOriginHosts` — so a fresh deploy 403'd every mutation until the operator manually set `SHEPHERD_ALLOWED_HOSTS`.

- New `addOwnHostToAllowlist` (`src/config.ts`), called at boot right after `previewHost` resolves (`src/index.ts`).
- **Security invariant preserved:** only the *hostname* is trusted. Preview-port origins (`previewHost:8001–8016`) stay rejected because the guard's preview-range check runs *first*, independent of hostname — so this grants same-host HUD mutations, not preview-forged CSRF. Foreign origins stay blocked.
- **Scope:** covers the documented direct-serve path (served front host == the node's `Self.DNSName`). A **Tailscale Service / custom-DNS** HUD is fronted under a *different* DNS name, so it still needs a manual `SHEPHERD_ALLOWED_HOSTS` entry — reworded the `deploy/shepherd.service:25` comment to say exactly that (direct-serve is now automatic).

## Fix 3 — accurate error copy (self-diagnosable failure)

A non-preview-port 403 (real HUD reached on a host absent from `SHEPHERD_ALLOWED_HOSTS`) surfaced as *"Preview is read-only. Open Shepherd directly…"*, misdirecting operators already on the real HUD.

- New `classifyOrigin` (`src/validate.ts`) → `allow | preview-port | host-not-allowed`; `originAllowed` becomes a thin `=== "allow"` wrapper (behavior unchanged, incl. curl/CLI no-Origin bypass and malformed→reject).
- `checkOrigin` returns a distinct body for the host case (`forbidden: origin host not allowed`); the client maps it to a plain error pointing at `SHEPHERD_ALLOWED_HOSTS` (`error_origin_host_not_allowed`, EN+DE). A genuine preview-port 403 keeps the "Preview is read-only" copy.
- Verified the new message surfaces: `applyUpdate` throws `apiError(...)` directly and `UpdateModal` renders `e.message` verbatim; the `isPreviewBlocked` rethrow (`api.ts:863`) is `relaunchSession`-only and special-cases `PreviewBlockedError`, so a plain host error isn't swallowed.

## Tests / verification

- `test/config-origin.test.ts` — `addOwnHostToAllowlist` (append, dedup, null/blank no-op, trim).
- `test/validate.test.ts` — `classifyOrigin` cases incl. **null/empty → allow** and **preview-port on an allowlisted host stays preview-port** (the CSRF invariant Fix 2 must not weaken).
- `ui/src/lib/api.previewBlocked.test.ts` — host-block 403 → plain (non-preview) error with hint copy.
- Green locally: root `bun run lint` + `bun test` (6538 pass / 0 fail), `ui` `bun run check` (0 errors) + `vitest` + `check:i18n` (parity).

No feature-catalog entry (both are `fix:`; Fix 3 is error copy, not a new capability), no glossary entry (`SHEPHERD_ALLOWED_HOSTS` is an env var), no manual operator steps (a restart/deploy — inherent to any change — fully applies it).

### DIFF
--- a/deploy/shepherd.service
+++ b/deploy/shepherd.service
@@ -22,7 +22,10 @@ RestartPreventExitStatus=78
 Environment=PATH=%h/.bun/bin:%h/.local/bin:/usr/local/bin:/usr/bin:/bin
 Environment=SHEPHERD_PORT=7330
 Environment=SHEPHERD_HOST=127.0.0.1
-# localhost defaults only; add your own tailnet hostname via SHEPHERD_ALLOWED_HOSTS in ~/.shepherd/env
+# localhost defaults only. A direct-served HUD (`tailscale serve` on this node) is trusted
+# automatically — Shepherd folds the node's own tailnet host into the allowlist at startup.
+# Only a Service-fronted / custom-DNS HUD (served under a different DNS name than this node)
+# needs its host added via SHEPHERD_ALLOWED_HOSTS in ~/.shepherd/env.
 Environment=SHEPHERD_ALLOWED_HOSTS=localhost,127.0.0.1,::1,[::1]
 # optional overrides (token, repo root, etc.) — file may be absent
 EnvironmentFile=-%h/.shepherd/env
--- a/src/config.ts
+++ b/src/config.ts
@@ -448,6 +448,22 @@ export function resolveAllowedOriginHosts(envValue: string | undefined): string[
   return hosts;
 }
 
+/**
+ * Fold the node's own resolved host into an origin allowlist in place (deduped;
+ * null/blank ignored). Called at boot once `config.previewHost` resolves so a
+ * same-node HUD — direct `tailscale serve`, where the served front host IS the
+ * node's own DNSName — is trusted without a manual SHEPHERD_ALLOWED_HOSTS. Only
+ * the hostname is trusted: preview-port origins are still rejected by the origin
+ * guard's preview-range check (it runs before the hostname check), so this does
+ * not open preview-forged CSRF. A Service-fronted HUD is served under a DIFFERENT
+ * DNS name, so it isn't covered here and still needs a manual allowlist entry.
+ * Issue #1645 Fix 2. Exported for tests.
+ */
+export function addOwnHostToAllowlist(hosts: string[], host: string | null): void {
+  const h = host?.trim();
+  if (h && !hosts.includes(h)) hosts.push(h);
+}
+
 export const config = {
   port: mainPort,
   // bind to loopback only; the Tailscale-serve proxy reaches it via 127.0.0.1.
--- a/src/index.ts
+++ b/src/index.ts
@@ -17,6 +17,7 @@ import {
   parseServedPort,
   validatePreviewPortRange,
   validateAgentIngressPort,
+  addOwnHostToAllowlist,
 } from "./config";
 import { SessionStore } from "./store";
 import type {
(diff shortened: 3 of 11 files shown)


scoop (commented):
## Review

The PR scrubs the hardcoded maintainer tailnet host from the deploy template and implements the issue's Fix 2 (auto-allow own host) and Fix 3 (accurate error copy). The change is well-scoped, well-tested, and correct.

### What it does
- **deploy/shepherd.service**: removes `shepherd.chicken-beardie.ts.net` from `SHEPHERD_ALLOWED_HOSTS`, leaving localhost defaults, and adds an accurate comment explaining that a direct-served HUD is now trusted automatically.
- **src/config.ts / src/index.ts**: `addOwnHostToAllowlist` folds `config.previewHost` into `config.allowedOriginHosts` at boot (after `resolveNodeHost()` resolves).
- **src/validate.ts / src/server.ts**: `classifyOrigin` returns `allow` | `preview-port` | `host-not-allowed`; `checkOrigin` emits a distinct 403 body for host-not-allowed so the client can point operators at `SHEPHERD_ALLOWED_HOSTS` instead of blaming the preview. `originAllowed` retained as a behavior-identical boolean wrapper.
- **ui**: new `error_origin_host_not_allowed` message + client mapping of the new 403 body to a plain (non-preview) error.

### Verification
- `resolveNodeHost` returns a bare, trailing-dot-stripped DNSName (verified src/tailscale.ts:49), so `allowedHosts.includes(parsed.hostname)` in classifyOrigin matches a direct-serve Origin like `https://agentnode.example.ts.net`.
- `config.allowedOriginHosts` is assigned a plain array from `resolveAllowedOriginHosts` (verified src/config.ts:520) and `config` is not frozen, so the in-place push in `addOwnHostToAllowlist` is valid.
- The auto-allow does NOT weaken CSRF: the preview-range check runs before the hostname check (verified src/validate.ts:905-914), so a preview app on the auto-allowed host:port still classifies as `preview-port` — confirmed by test/validate.test.ts `preview-port on an allowlisted host is preview-port (not allow)`.
- Server/client string contract is exact: server emits `forbidden: origin not allowed` (preview) and `forbidden: origin host not allowed` (host), matched by `ORIGIN_BLOCK`/`ORIGIN_HOST_BLOCK` via `===` (verified src/server.ts:643-645 against ui/src/lib/api.ts:93,99,120-125). No substring collision.
- i18n parity: `error_origin_host_not_allowed` present in both catalogs (verified ui/messages/en.json:44 and ui/messages/de.json:44); repo maintains only en+de, and `ui/src/lib/paraglide` is generated/gitignored (verified ui/.gitignore:26), so `m.error_origin_host_not_allowed()` is regenerated by the build.
- No other consumers of the origin error string exist outside server.ts/api.ts/tests (verified via grep); `originAllowed` callers at src/server.ts:6891,6906 are unaffected since the wrapper preserves the boolean contract.

Satisfies the task. No findings.

PR #1655, opened 2026-07-11
fix: auto-allow Tailscale Service-fronted HUD hosts too (#1645 Fix 2/3)

scoop:
## Problem

The auto-allow shipped in #1649 doesn't cover the topology it was filed for. It folds **only** the node's own `Self.DNSName` into the CSRF origin allowlist — but a HUD reached through a Tailscale **Service** front is served under a *different* DNS name, so its `Origin` never matches and every mutation still 403s until the operator sets `SHEPHERD_ALLOWED_HOSTS` by hand.

Verified live on an affected node:

| Host | Source | In allowlist? |
| --- | --- | --- |
| `<node>.<tailnet>.ts.net` | node `Self.DNSName` (what #1649 folds in) | yes |
| `shepherd.<tailnet>.ts.net` | `svc:shepherd` Service front → `http://localhost:7330` (the actual HUD Origin) | **no → 403** |

**Two layers:**
1. **Trigger** — #1646 (`46a9786f`) rightly scrubbed a maintainer's personal tailnet host from `deploy/shepherd.service`'s hardcoded `SHEPHERD_ALLOWED_HOSTS`. That removal is what made same-node operators suddenly need a manual entry.
2. **Gap** — #1649 was meant to compensate but only trusts `Self.DNSName`, and its own doc-comment conceded Service fronts "still need a manual allowlist entry" — which is exactly #1645's reported case (`shepherd.<tailnet>.ts.net`).

## Fix

Reuse the existing Services-aware JSON serve parser in `src/config.ts` (`collectWebMaps` + `handlerTargetsPort` + the `host:PORT` key split) rather than adding a new parser:

- **`findServedHosts(json, localPort)`** — every public host whose handler `.Proxy` targets loopback on the **main port**, across the top-level `Web` map and every `Services[svc].Web` map. Deduped, fully defensive (`[]` on any malformed/empty/missing input, never throws).
- **`addServedHostsToAllowlist(...)`** — folds those hosts in at boot, next to `addOwnHostToAllowlist`.
- **`src/index.ts`** — the boot call switches from text `tailscale serve status` to `serve status --json`, so a **single** exec yields both the served port (`findServedPort`) and the service host(s). This orphaned the text parser `parseServedPort` (+ `extractServedPort`, `isProxyTarget`), which are **removed** (`diagnostics.ts` already uses the JSON `findServedPort`).

The trust boundary is the shared loopback+main-port filter (`LOOPBACK_RE` = `localhost|127.0.0.1`, unchanged): only a service actually fronting Shepherd's port is trusted; unrelated hosted services and preview-port origins are excluded.

## Verification

- `bun run lint`, `bun run typecheck` clean; `bun test ./test` → 6544 pass / 0 fail. New `findServedHosts` / `addServedHostsToAllowlist` tests cover Service front, direct-serve, wrong-port and non-loopback exclusion, bare-host key, trailing-dot, cross-map dedup, and malformed input.
- **Empirical, on a live affected node:** the new `findServedHosts(<real `serve status --json`>, 7330)` returns `["shepherd.<tailnet>.ts.net"]` (the exact HUD front) and `[]` for a non-Shepherd port — so a fresh deploy trusts the Service-fronted HUD with **no** manual `SHEPHERD_ALLOWED_HOSTS`, while the boundary holds.

After this merges and the node restarts, operators who added a manual `SHEPHERD_ALLOWED_HOSTS` Service-host workaround can drop it (optional — it's harmless either way).

`[no-feature-entry]` — server-only, no user-facing UX.

### DIFF
--- a/deploy/shepherd.service
+++ b/deploy/shepherd.service
@@ -24,8 +24,10 @@ Environment=SHEPHERD_PORT=7330
 Environment=SHEPHERD_HOST=127.0.0.1
 # localhost defaults only. A direct-served HUD (`tailscale serve` on this node) is trusted
 # automatically — Shepherd folds the node's own tailnet host into the allowlist at startup.
-# Only a Service-fronted / custom-DNS HUD (served under a different DNS name than this node)
-# needs its host added via SHEPHERD_ALLOWED_HOSTS in ~/.shepherd/env.
+# A Tailscale Service front (`tailscale serve` registered under `svc:<name>`) proxying to
+# this HUD's port is auto-trusted too, discovered the same way. Only a non-Tailscale reverse
+# proxy / custom-DNS front (a host that doesn't appear in `tailscale serve status`) needs its
+# host added via SHEPHERD_ALLOWED_HOSTS in ~/.shepherd/env.
 Environment=SHEPHERD_ALLOWED_HOSTS=localhost,127.0.0.1,::1,[::1]
 # optional overrides (token, repo root, etc.) — file may be absent
 EnvironmentFile=-%h/.shepherd/env
--- a/src/config.ts
+++ b/src/config.ts
@@ -126,54 +126,6 @@ export function clampCap(n: number, min: number, max: number, fallback: number):
 // previewPortCount is BOTH the range size and the max concurrent previews — a
 // single number, no secondary "max" constant anywhere else.
 
-/**
- * Pure parser: given the text of `tailscale serve status` and the HUD's local
- * listen port, return the public-facing HTTPS port that Tailscale fronts that
- * local port on. Returns null when no matching mapping is found.
- *
- * Example input lines:
- *   https://host.ts.net:5191 (tailnet only)
- *   |-- / proxy http://127.0.0.1:5190
- *
- * The HUD's default mapping has no port (=> 443):
- *   https://host.ts.net (tailnet only)
- *   |-- / proxy http://127.0.0.1:7330
- */
-export function parseServedPort(serveStatusText: string, localPort: number): number | null {
-  const lines = serveStatusText.split("\n");
-  let currentServedPort: number | null = null;
-  for (const line of lines) {
-    const t = line.trim();
-    const served = extractServedPort(t);
-    if (served !== null) {
-      currentServedPort = served;
-    } else if (currentServedPort !== null && isProxyTarget(t, localPort)) {
-      return currentServedPort;
-    }
-  }
-  return null;
-}
-
-/** Extract the public HTTPS port from a `https://...` serve-status header line.
- *  Returns 443 when no explicit port is present, null when the line isn't an
- *  https header.
- *  Note: IPv6 bracket hosts (`[::1]:…`) are intentionally not matched because
- *  Tailscale `serve` always emits `127.0.0.1` + ts.net hostnames. */
-function extractServedPort(line: string): number | null {
-  const m = line.match(/^https:\/\/[^:/]+(?::(\d+))?(?:\s|$)/);
-  if (!m) return null;
-  return m[1] ? Number(m[1]) : 443;
-}
-
-/** True when a `|-- / proxy ...` line targets 127.0.0.1:<localPort>.
- *  Note: `localhost` and IPv6 bracket targets (`[::1]:…`) are intentionally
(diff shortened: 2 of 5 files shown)


scoop (commented):
## Summary

The PR adds `findServedHosts` + `addServedHostsToAllowlist` and wires them into startup so that a Tailscale **Service front** (served under a different DNS name than the node's own, e.g. `svc:shepherd` → `shepherd.ts.net`) is folded into the CSRF origin allowlist automatically — the topology the previously-shipped `addOwnHostToAllowlist(previewHost)` (Fix 2, already in base at `686300f8`) did not cover. It also migrates port discovery off the deleted text parser (`parseServedPort`) onto the JSON parser (`findServedPort`) sharing a single `tailscale serve status --json` call.

I verified the implementation satisfies the task and found no blocking issues.

### Verification performed

- **Served-host discovery is authoritative, not just node-name based.** `findServedHosts` iterates `collectWebMaps` (top-level `Web` + every `Services[svc].Web`) and matches only handlers whose `.Proxy` targets loopback:localPort via `handlerTargetsPort` (`src/config.ts:167-201, 261-276`). This means it covers BOTH direct serve (top-level `Web`, verified by test `test/preview-config.test.ts:119-121`) and Service fronts (`test/preview-config.test.ts:115-117`), so the fix is robust even if the node's own DNSName differs from the served front. The Service-front proxy target `http://localhost:7330` matches `LOOPBACK_RE` (`src/config.ts:129`) and `config.port` (`src/config.ts:413`, `mainPort`=7330).
- **The allowlist array is the one consumed at request time.** `addServedHostsToAllowlist` mutates `config.allowedOriginHosts` in place (`src/index.ts:420`), the same array `resolveAllowedOriginHosts` seeds (`src/config.ts:543`) and `originAllowed` reads on every request (`src/server.ts:6891,6906`). `classifyOrigin` compares `parsed.hostname` (URL-lowercased, port stripped) against `allowedHosts.includes(...)` (`src/validate.ts:916-917`); folded hosts are bare hostnames via `hostFromKey`, so `https://shepherd.ts.net:443` → hostname `shepherd.ts.net` matches. Preview-range rejection still runs first (`src/validate.ts:904-915`), so this does not open preview-forged CSRF.
- **Host normalization parity.** `hostFromKey` strips a trailing dot (`src/config.ts:147-150`) matching `resolveNodeHost`'s DNSName normalization (`src/tailscale.ts:48-49`); verified by test `test/preview-config.test.ts` trailing-dot and bare-host cases.
- **Fail-closed / defensive parsing.** Malformed/empty/non-JSON → `[]` (`src/config.ts:262-268`), array-valued `Web`/`Services` rejected by `isPlainObject` (`src/config.ts:132-134,153-164`). `stdout` defaults to `""` on tailscale failure (`src/index.ts:397,404`) → no-op. Trust is bounded: only fronts actually proxying to the HUD's own port are added, so no new hosts are trusted beyond real HUD fronts.
- **No dangling references to the removed parser.** Grepped `parseServedPort`/`extractServedPort`/`isProxyTarget` across `src/` and `test/` — zero remaining uses; `src/index.ts` and `src/diagnostics.ts:286` both use `findServedPort` with `--json`.
- **deploy/shepherd.service comment** now accurately states Service fronts are auto-trusted and only a non-Tailscale / custom-DNS front (not in `tailscale serve status`) needs a manual `SHEPHERD_ALLOWED_HOSTS` entry — matches code coverage. Env var name `SHEPHERD_ALLOWED_HOSTS` is consistent with `process.env.SHEPHERD_ALLOWED_HOSTS` (`src/config.ts:543`).

### Non-blocking observation (not a required change)

- `stdout` is JSON-parsed twice at startup — once by `findServedPort` (`src/index.ts:402`) and again by `findServedHosts` via `addServedHostsToAllowlist` (`src/index.ts:420`). The payload is tiny and this runs once at boot, so it's negligible; the clean separation of two pure functions is a reasonable trade-off. No action needed.

### Stated limitation

- The exact `tailscale serve status --json` schema (v1.96.4) cannot be validated against a live tailscale binary in this repo; I confirmed the assumed shape (`Web`/`Services[svc].Web` → `{host:port: {Handlers: {"/": {Proxy}}}}`) is internally consistent with the existing `findServedPort` fixtures and `test/diagnostics.test.ts:57-62,384-394`, which the base already relies on.

Issue #1478, opened 2026-07-06
bug: 'Schritte bestätigen' (confirm manual steps) does nothing on a merged Owed card


Gambastudio:
## Summary

On the **merged (GEMERGT) / Owed** card of a session that declared manual operator steps, clicking **"Schritte bestätigen" / "SCHRITTE BEST."** (confirm steps) produces **no visible effect** — the manual steps cannot be worked off from the HUD.

## Steps to reproduce

1. Run a task whose PR declares manual operator steps (`shepherd:manual-steps` block / `Manual-Step:` trailers), then merge it.
2. On the merged card (GEMERGT / Owed lens) the card shows e.g. **"2 manuelle(r) Schritt(e)"** + a **"Schritte bestätigen"** button.
3. Click **"Schritte bestätigen"**.

## Actual

Nothing happens. No per-step checklist opens, no confirmation, no toast/feedback. The "N manuelle(r) Schritt(e)" count and the button stay unchanged; the record is not resolved or dismissed. Verified across the full click window in a screen recording (hover → click → after): the state never changes.

## Expected

Clicking should let the operator resolve the steps — either expand a per-step checklist to tick each one done (`POST /api/manual-steps/:sessionId/steps/:stepId {done}`) or confirm/dismiss the record (`POST /api/manual-steps/:sessionId/dismiss`) — with visible feedback.

## Context / likely area

- Manual-steps feature shipped in #1059 / #1060 / #1061 (P1–P3). This looks like the confirm CTA from #1060 ("ack/CTA") is non-functional on the current build, at least for a **post-merge-materialized** (#1061) record on the Owed/GEMERGT card.
- Observed on a self-hosted instance, German UI (`SCHRITTE BEST.` = *Schritte bestätigen*), on a merged session's Owed card.
- Worth checking: click handler missing/misbound on the merged-card variant; a silently-failing `POST /api/manual-steps/...` (origin/auth → swallowed); or the CTA only toggles a per-step list that doesn't render.

## Evidence

Screen recording available (button hover → click → no state change). Happy to attach.

scoop:
✅ PR #1663 merged.

Issue #1060, opened 2026-06-24
Manual operator steps — P2: gate auto-merge + ack/CTA + push + rundown


scoop:
Part of epic #1056. **Blocked by the P1 sub-issue** (needs manual-step detection + `manualSteps` persistence on the session).

## Goal

Block auto-merge while a PR has un-acked **non-`POST-MERGE`** manual steps; give the operator a one-click ack/CTA; nudge via push + rundown. This generalizes the epic-landing migration-ACK pattern (#645) to all session PRs.

## Re-confirm at this session's plan gate (MANDATORY)

- **Gate scope:** **all session PRs** (default) vs a per-repo `manualStepsGate` opt-in setting. This is a repo-wide auto-merge behavior change that the parent issue flagged as a risky Open Question — re-surface and confirm before building.
- **Ack semantics:** acknowledged-will-do (default — mirrors migration-ACK, which dismisses without asserting completion) vs done.

## Scope

1. **Gate:** extend `MergeSessionView` (`src/automerge-core.ts`) with the session's `manualSteps` + `manualStepsAckedAt`. In `readyToMerge()` return `false` when there exists a non-`POST-MERGE` step and `manualStepsAckedAt == null`. Default applies to **all** session PRs (re-confirm scope at plan gate). Add a distinct `hold` reason code so the UI can explain *why* the PR is held.
2. **Ack endpoint:** `POST /api/sessions/:id/ack-manual-steps` (mirror `POST /api/epics/completed/ack-migrations` at `src/server.ts:4566`) → a `store.ackManualSteps(sessionId)` that stamps `manualStepsAckedAt`. Ack = "acknowledged, clear the gate." Idempotent.
3. **CTA:** surface an "Ack manual steps" action wherever the chip/hold shows (backlog row / session detail), wired to the endpoint.
4. **Push:** new `manual_steps` notification kind (`src/push.ts`), fired when a green, up-to-date, otherwise-ready-to-merge PR is held *only* on un-acked steps — the "don't let auto-merge eat this" nudge. i18n EN+DE payloads (server-side notification rule). Stable, **tone-namespaced dedupe key**; route persistent/failure-class toasts through the shared store (not transient `info()`) per house rule. **Set the notified flag only after the send's return value confirms success** (house rule — optimistic stamping drops the signal).
[...]

PR #1663, opened 2026-07-11
fix(herd): drop moot ack button on terminal cards, verb-label merged manual-steps chip (#1478)

scoop:
## Problem

On a **merged (GEMERGT)** card, clicking **"Schritte best."** (the #1060 *Ack steps* button) produced no visible effect — the manual steps couldn't be worked off from the HUD (#1478).

## Root cause (two things, one confirmed here)

1. **The ack button was literally unclickable.** `.manual-steps-ack` was missing `position: relative; z-index: 1`, so it sat *under* the row's full-bleed `.unit-hit` click overlay (its sibling `.chip-manual-steps--link` already has this exact recipe). A click landed on the overlay, not the button — no ack POST, no toast, "nothing happens." This surfaced as a real Playwright click-timeout while writing the open-card test, and is the concrete, confirmed explanation for the reported symptom on a healthy build.
2. **On a terminal card the ack button is moot anyway.** Post-merge/close there is no auto-merge gate left to clear, and acking never resolves the durable Owed record. The actual resolution route — the count chip that links to the Owed lens — was passive (a neutral count) and got bypassed for the verb-shaped button.

## Change

**UI (`fix(herd)`)**
- Add `isMerged` / `isTerminal` deriveds (`isTerminal = merged || closed`, matching the existing `stepperTerminal`).
- **Drop the moot ack button on terminal (merged/closed) cards** (`!isTerminal` gate); open-PR behavior unchanged (it still clears the real auto-merge gate).
- **Give merged cards a verb affordance:** relabel the count chip to `Resolve N step(s)` → the Owed lens (per-step checklist + dismiss), so the resolution route reads as the action the operator reaches for.
- **Fix the unclickable ack button** (`position: relative; z-index: 1`), repairing the still-live pre-merge ack CTA too.
- `onShowOwed` now force-refreshes the outstanding set (`postMergeStepsStore.load()`) so a client that missed a `post-merge-steps:changed` frame lands on the live workable record, not the read-only frozen fallback. (Owed tick/dismiss are POST-response-driven, so resolution works regardless of WS delivery.)
- New i18n key `unitrow_resolve_manual_steps` (EN + DE).

**Demo (`chore(demo)`)**
- Add the missing `POST /api/sessions/:id/ack-manual-steps` route + `demoState.ackManualSteps` (mirrors the server: idempotent stamp + `session:manual-steps` emit) so the pre-merge ack CTA is faithful in demo.
- A merged `envflag` showcase card (session + merged git state + outstanding owed record) exercising the new merged-card verb chip.

## Design notes / trade-offs

- **Merged card with declared steps but no outstanding record** (resolved/dismissed): the verb chip → Owed lens shows the accurate `owed_frozen_cleared_note` ("no longer owed"). Intentional; the note is only misleading in the narrow *never-materialized* case (see below).
- I deliberately kept this **minimal** — no `owedRecord` threading / record-gating / mobile eager-GET, which an earlier draft explored but which regressed the resolved-state signal and didn't earn its keep (the frozen card is informative, not a hard dead-end).

## Closing criterion (why `Refs`, not `Fixes`)

Refs #1478. The described "click does nothing" symptom is resolved here (the button is now clickable and, on merged cards, replaced by a working verb affordance). One case this UI change cannot resolve on its own: if the reporter's session has **no materialized Owed record** (server-side non-materialization), the Owed lens shows the read-only frozen card and the steps still can't be ticked/dismissed — that closer is a separate **#1061** materialization-reliability fix. Please confirm on the affected instance that its record materializes (`GET /api/manual-steps/outstanding` includes the session / the Owed lens is workable) before closing #1478; if it doesn't, hand off to #1061.

## Verification

- `cd ui && bun run check` (svelte-check 0/0 + `check:i18n` parity) ✓
- UnitRow browser tests 55/55 (open = ack present + neutral label; merged = ack absent + verb label + click→onshowowed; closed = ack absent + neutral label) ✓
- Demo router/state tests (ack route → 200 + `session:manual-steps` with non-null `ackedAt`; seed invariants) ✓
- Full suites green (UI browser 1176, node 2180); pre-push gate (branch-hygiene, feature-catalog, glossary, fallow audit — complexity 0) ✓

[no-feature-entry] — a bug fix; the demo changes ship no new user-facing product UX.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2080,6 +2080,7 @@
   "feat_manual_operator_steps_body": "Deklariere manuelle Schritte im shepherd:manual-steps-Block eines PR (Flag umlegen, Umgebungsvariable setzen, Backfill ausführen), und Shepherd zeigt sie als bernsteinfarbenen Chip in der Session-Zeile sowie als Checkliste im Done-Recap an – damit sie beim Mergen des PR nicht verloren gehen.",
   "unitrow_ack_manual_steps": "Schritte best.",
   "unitrow_ack_manual_steps_failed": "Manuelle Schritte konnten nicht bestätigt werden — bitte erneut versuchen.",
+  "unitrow_resolve_manual_steps": "{count} Schritt(e) erledigen",
   "hold_manual_steps": "{steps} manuelle(r) Schritt(e) vor dem Mergen — bestätigen zum Fortfahren.",
   "feat_manual_operator_steps_gate_title": "Manuelle Schritte blockieren Auto-Merge",
   "feat_manual_operator_steps_gate_body": "Ein PR mit einem unbestätigten manuellen Schritt (alles, was nicht als POST-MERGE markiert ist) wird jetzt vom Auto-Merge zurückgehalten, bis du handelst. Klicke in der Session-Zeile auf \"Schritte best.\", um die Sperre aufzuheben; du wirst per Push und im täglichen Rundown erinnert. Reine POST-MERGE-Schritte blockieren nie.",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2080,6 +2080,7 @@
   "feat_manual_operator_steps_body": "Declare manual steps in a PR's shepherd:manual-steps block (flip a flag, set an env var, run a backfill) and Shepherd surfaces them as an amber chip on the session row and a checklist in the Done recap — so they aren't lost when the PR lands.",
   "unitrow_ack_manual_steps": "Ack steps",
   "unitrow_ack_manual_steps_failed": "Couldn't acknowledge the manual steps — try again.",
+  "unitrow_resolve_manual_steps": "Resolve {count} step(s)",
   "hold_manual_steps": "{steps} manual step(s) to do before merge — ack to proceed.",
   "feat_manual_operator_steps_gate_title": "Manual steps gate auto-merge",
   "feat_manual_operator_steps_gate_body": "A PR with an un-acknowledged manual step (anything not marked POST-MERGE) is now held out of auto-merge until you act. Click \"Ack steps\" on the session row to clear the gate; you're nudged via push and the daily rundown. POST-MERGE-only steps never block.",
--- a/ui/src/lib/components/UnitRow.browser.test.ts
+++ b/ui/src/lib/components/UnitRow.browser.test.ts
@@ -934,6 +934,81 @@ describe("UnitRow manual-steps chip", () => {
     await expect.element(page.getByTitle(m.unitrow_manual_steps_link())).not.toBeInTheDocument();
     await expect.element(page.getByText(m.unitrow_manual_steps({ count: 1 }))).toBeInTheDocument();
   });
+
+  // #1478: on a merged/closed card the auto-merge gate the Ack CTA used to clear no longer
+  // exists, so it must not render there; the count chip becomes the actual resolution route
+  // and gets a verb label on merged cards to read as the action.
+  it("open card: shows the Ack CTA and the neutral count-chip label", async () => {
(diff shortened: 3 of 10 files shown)


scoop (commented):
## Review

The PR fixes the reported bug from three angles and I could verify each against the code.

### The core fix is correct
- `ui/src/lib/components/UnitRow.svelte:152-161` — on terminal (merged/closed) cards the auto-merge gate the Ack CTA cleared no longer exists, so `showAckCta` is gated with `!isTerminal` and the (moot) Ack button is hidden; the count chip becomes the resolution route and gets a verb label on merged cards. The chip already routes to `onshowowed` (`UnitRow.svelte:775-786`), which is wired to `onShowOwed` in the real app at `ui/src/routes/+page.svelte:2627,2784`. Verified `git.state` can be `"merged"`/`"closed"` against `ui/src/lib/types.ts:289`.
- `ui/src/lib/components/UnitRow.svelte:1397-1398` — the real root cause of "clicking does nothing" on the pre-merge path: `.manual-steps-ack` was statically positioned and sat under the positioned `.unit-hit` overlay (`z-index:0`, `UnitRow.svelte:942`). Raising it with `position:relative; z-index:1` mirrors the working `.chip-manual-steps--link` recipe (`UnitRow.svelte:1374-1376`) and makes it clickable. Correct.
- `ui/src/routes/+page.svelte:2390` — forcing `postMergeStepsStore.load()` on chip click avoids a client-stale frozen set; `load()` has an in-flight guard and is not `loaded`-gated, so repeated calls are safe (`ui/src/lib/post-merge-steps.svelte.ts:33-45`).

### Demo parity is correct
- `ui/src/lib/demo/router.ts:232-240` adds the previously-missing `POST /api/sessions/:id/ack-manual-steps` route. Path/method/response `{ok:true}` mirror the real client `ui/src/lib/api.ts:2232-2238`; `seg(path,3)` correctly extracts the id (`router.ts:31-33`).
- `ui/src/lib/demo/state.ts:456-465` `ackManualSteps` stamps `manualStepsAckedAt` idempotently (`??=`) and emits `session:manual-steps` with `{id, manualSteps, manualStepsAckedAt}` — shape matches the `WsEvent` union at `ui/src/lib/types.ts:1557-1560`. Helpers `find`/`emit` exist (`state.ts:63,67`).
- Seed `envflag` showcase (`ui/src/lib/demo/seed.ts`) is consistently added to `buildSessions`, `buildGitStates` (merged), and `buildPostMergeSteps` with the same session/step id, so the Owed lens shows a workable record. Adding a `done` session does not perturb the director, which only acts on WORKING sessions (`ui/src/lib/demo/director.ts:109-114`). No other test hardcodes a 7-session count (grepped `ui/src`).

### i18n parity holds
- `unitrow_resolve_manual_steps` added to BOTH catalogs with matching `{count}` placeholder: `ui/messages/en.json:2083` (`Resolve {count} step(s)`) and `ui/messages/de.json:2083` (`{count} Schritt(e) erledigen`). Only en/de catalogs exist. Verified.

### Non-blocking observations (not findings)
- The verb chip counts `session.manualSteps.length` (the declared set), not the number still-outstanding, so after an operator ticks some steps in the Owed lens the merged card can still read e.g. "Resolve 2 step(s)". This is pre-existing behavior of the count chip (the neutral label used the same length) — the PR does not regress it, but the `Resolve` wording makes the staleness more noticeable. Consider deriving the count from outstanding steps in a follow-up.

- The merged-card manual-steps chip persists as long as `session.manualSteps.length > 0` regardless of whether the post-merge record was resolved/dismissed — session-level `manualSteps` is the static declaration and isn't reconciled with the Owed record's resolution state (data model in `ui/src/lib/types.ts`, not touched here).

PR #1670, opened 2026-07-11
feat(drain): pre-warm epic landing CI via an early draft landing PR (default-off toggle) (#1664)

scoop:
## What & why

Closes #1664 (deferred piece **A** of the epic-drain landing-PR reliability work; **B**/**C** shipped in #1667).

Today an epic's aggregate landing PR (`epic/<#> → main`) opens only at completion, so it is the **first main-relative CI run** for the whole epic — semantic drift and main-only gates (Fallow delta vs main, `pr-title` commitlint, `docs-site`) all first bite there. This PR adds an **opt-in** ability to open that landing PR **early, as a draft**, during the drain, so each child merge into the integration branch re-runs the full main-relative CI on the merge ref. At completion the draft is finalized (body rebuilt from the final rollup, un-drafted) and lands as today — the verdict is pre-computed instead of a cold first run at the finish line.

Ships behind a **default-OFF** `RepoConfig` toggle (`preWarmEpicLandingCi`). The honest benefit is pre-warming only; the cost is ~1 extra main-relative CI run per child merge — so it's operator opt-in, not always-on.

## Design decisions (from the approved plan)

- **Surfacing gap → accepted.** A red running-epic draft is GitHub-visible only, not a Shepherd Tier-1 item until completion (`enrichLandingEpics` reads only `listEpicCompleted`). No new running-epic signal; existing B/C surfacing applies at completion.
- **Host scope → GitHub only.** The pass is gated `forge.kind === "github"`, matching `rebaseStuckLandingPrsForRepo` / `rerunRedLandingCiForRepo` / `editPr`.

## How it works

1. **`preWarmEpicLandingCi` flag** (default off) — wired server (`store.ts`/`server.ts`) + UI (`AutomationSettings` toggle, store, api, types) + EN/DE i18n.
2. **`ensureDraftLandingPrForRepo`** (new `src/drain.ts` pass, wired into `tick()`): flag-on + running epic + GitHub + ≥1 integration-merged child → opens the landing PR as a **draft** with a provisional-marker body. Idempotent via `prStatus(integrationBranch)`.
   - **Writes NO `epic_completed`/`landingState` row during the drain** — load-bearing: `rebaseStuckLandingPrsForRepo` force-pushes only recorded `open` rows, so a merely-behind draft still gets its merge-ref CI run but is never `--force-with-lease`-rewritten under open child PRs.
[...]

Issue #1664, opened 2026-07-11
feat(drain): pre-warm epic landing CI via an early draft landing PR (behind a default-off toggle)


scoop:
## Context

Follow-up to the epic-drain landing-PR reliability work (PR shipped **B** — surface a CI-failing landing PR — and **C** — auto-rerun failed landing CI). This is deferred piece **A** from that plan.

Today an epic's aggregate landing PR (`epic/<#> → main`) opens only at completion, so it is the **first main-relative CI run** for the whole epic — semantic drift and main-only gates (Fallow delta vs main, `pr-title` commitlint, `docs-site`) all first bite there.

## Goal

Pre-warm / pre-validate the landing against `main` **during** the drain by opening the landing PR as a **draft** early and keeping it open, so each child merge into `epic/<#>` re-runs the full main-relative CI + hygiene on the merge ref. At completion, un-draft and land as today. The landing verdict is then pre-computed rather than a cold first run at the finish line.

**Ship behind a default-off `RepoConfig` toggle** (the honest benefit is pre-warming only; cost is ~1 extra main-relative CI run per child merge — operator opt-in, not always-on).

## Requirements

- New default-off `RepoConfig` flag; only when on does a new `ensureDraftLandingPrForRepo` pass run (engaged predicate + ≥1 integration-merged child). Opens `forge.openPr({…, draft: true})` (`OpenPrInput.draft` exists — GitHub `--draft`, Gitea `WIP:`). Idempotent via `prStatus(integrationBranch)` reuse.
- **No force-push during the drain.** The draft must **not** be recorded as an `epic_completed`/`landingState` row while the epic runs, so `rebaseStuckLandingPrsForRepo` (`src/drain.ts`) never `--force-with-lease`-rewrites it under open child PRs. A merely-behind draft still gets a merge-ref CI run; a conflicting draft surfaces the conflict and is resolved at completion.
- **Surfacing gap (known):** `enrichLandingEpics` reads only `listEpicCompleted`, so a red running-epic draft is GitHub-visible only, not a Shepherd Tier-1 item until completion. Either accept this or add a dedicated running-epic signal.
- **A.3 — refresh the body at adoption.** At completion the adoption path (`classifyLanding` reuse) must rebuild the landing PR body from the **final** child rollup (titles/PR numbers/set drift during the drain), not reuse the early sub-issue-derived body. Confirm/add a body-edit forge method.
[...]

PR #1673, opened 2026-07-12
fix(tab-reaper): bind helper labels + boot reap to shared constants (#1147)

### LINKED ISSUE #1147: Stale `autopilot <id>` agents leak identically to the learnings helpers — no boot reap

scoop:
Closes #1147.

## TL;DR

**The boot reap #1147 asks for already exists** — #1136 (PR #1157) shipped it. But #1147 is the **3rd recurrence of this leak class** (#1135 → #1136 → #1147) and the wiring that fixes it was **untested, silently deletable, and — as the PR critic caught — not actually bound to the labels it reaps**. So rather than close the issue with a comment, this repurposes it into a real guard.

## Why the issue's premise was wrong

#1147 deferred a fix on the belief that *"an autopilot agent absent from the in-memory map at boot may be a live/re-adoptable session"*, and therefore needed bespoke liveness/ownership reasoning.

That is **false by construction**. An `autopilot <id>` pane is *only ever* the transient stop-classifier: the label is built at the two `consider(...)` sites in `autopilot.ts`, which feed `classifyStop` → `herdr.start` (`autopilot-llm.ts`). A live session **never** carries that label — sessions get a prompt-derived `[a-z0-9-]` slug via the `relabel` path in `service.ts`, and a space-prefixed label cannot collide with a slug.

That is exactly why the existing name-scan boot reap with an **empty owned set** is already correct: no classifier runs at the synchronous boot point, so every match is a prior-lifetime orphan.

Two follow-on corrections:
- **The periodic husk sweep does *not* cover this** (contra the "husk detection now matches `autopilot `" note on the issue). An orphaned classifier idles as a live, non-shell `claude`, so `reapOrphanTabs`' liveness gate **spares** it (`sparedLive`). Only the boot label-reap closes it.
- **`git blame` misleads here.** The reap was introduced by `adb08b08` (#1136 / PR #1157, merged 2026-06-27 — the day *after* #1147 was filed, which is why it read as missing). Its *last touch* is `f4f34cc1` (#1553 async-herdr port, 2026-07-09), a mechanical change. Blame shows only the July date.

## What this changes

**1. Shared label constants (the substantive fix).** The three synchronous block-and-clean helper labels were **four scattered string literals** — one at each spawn site, one in the `index.ts` boot reap, one in `tab-reaper.ts`'s husk filter. Renaming a label at its producer would silently desync the spawn from the reap: the leak returns, every test still green.

Now `AUTOPILOT_LABEL` / `NAMER_LABEL` / `VERIFY_KEY_LABEL` are **exported from their producer modules** (matching the existing `DISTILL_LABEL`/`OPTIMIZE_LABEL`/`MERGE_LABEL` precedent) and consumed at every site: the spawns, the boot reap, and the husk filter. The producer↔reaper binding is now **compile-enforced rather than asserted** — desync is no longer expressible.

**Behavior-identical:** every label resolves to the same string as before (`"autopilot "`, `"name "`, `"verify api key"`). Only the source of truth moved.

**2. Regression guard** in `test/transient-tab-reaper.test.ts`:
- **Behavioral** — each real helper label reaps its orphan while **sparing a user session slug** (the collision property that makes the empty owned set safe at boot).
- **Source-text wiring guard** — asserts `src/index.ts` still contains a `reapTransientByLabel(herdr, <CONSTANT>` call per label. The boot calls are bare `void` statements in a `deferredStarts.push` thunk: unreachable from a unit test and therefore silently deletable — *that deletion is the recurrence mode*. With the labels now shared constants, this only has to catch the one thing types can't: the call being deleted. Precedent: the `no new ./config import` guard in `test/operator-language.test.ts`.

**Both failure modes verified empirically**, not just asserted:
- delete the boot call → the guard goes **red**;
- rename a label value → spawn and reap move **together**, tests stay green *and correct*.

Deliberately **not** done: collapsing the three boot calls behind one `reapSyncHelperOrphans()` seam. It wouldn't guard the deletion mode (a bare `void reapSyncHelperOrphans(herdr);` is just as deletable) and would *raise* blast radius — one deletion would disable the reap for all three helpers.

## Also: repairs a red `main` (unrelated, surfaced not silently folded)

`main` was **red** — `bunx tsc --noEmit` fails on pristine `origin/main`, blocking *any* push:

```
test/drain-landing-repair.test.ts(136,29): error TS2345:
  Property 'preWarmEpicLandingCi' is missing ... but required in type 'RepoConfig'.
```

`a0446713` (#1664 / PR #1670) added the required `preWarmEpicLandingCi` field to `RepoConfig` and updated its own tests, but missed this one fixture. Fixed in a **separate commit** (`preWarmEpicLandingCi: false`, matching every sibling fixture and the default-off toggle). Called out rather than quietly bundled — happy to split it out if you'd rather land it on its own.

## Verification

- `bun install && bun run lint` — clean
- `bun run test` — **6759 pass / 0 fail**
- `bunx tsc --noEmit` — clean (was failing on `main` before the fixture repair)
- Guard proven to fail on boot-call deletion; label rename proven to propagate

## Scope confirmation (issue's third bullet)

`recap` / `rundown` / `namer` restart-safety, per the header of `src/transient-tab-reaper.ts` — all covered, by three different mechanisms:

| Helper | Mechanism |
| --- | --- |
| `namer`, `autopilot`, `verify-key` | boot label-reap (this guard) |
| `recap`, `rundown` (herd-digest) | store `generating` state → `reapGenerating` |
| plan-gate / review / doc-agent | persisted `reviewer_spawns` rows → `adoptOrphans`/`reapOrphans` |

One correction to the issue's supposition: `namer` is safe because it's **boot-reaped**, not because it's a short-lived `-p` spawn — it's an interactive spawn, which is precisely why it needs the reap.


scoop (commented):
## What this PR actually is

The diff is **test-only**: 58 added lines in `test/transient-tab-reaper.test.ts` plus a one-line fixture field in `test/drain-landing-repair.test.ts`. There is no source change. The implied claim is that issue #1147's fix already shipped with #1136 and only needs a regression guard. I verified that claim rather than taking it on faith:

- The boot reap already exists at the diff base, not just at HEAD: `git show 21caec44:src/index.ts` contains the three `void reapTransientByLabel(herdr, "name " | "autopilot " | "verify api key", new Set(), ...)` calls at lines 2073-2075 (present at HEAD at `src/index.ts:2073-2075`). So the PR adds coverage for pre-existing behavior; it does not regress or duplicate it.
- The reaper's semantics match what the tests assert: `src/transient-tab-reaper.ts:44-49` closes every listed agent whose `name` starts with the prefix and whose `terminalId` is not in the owned set. With an empty owned set every prefix match is closed, which is exactly what the new parameterized test asserts.
- The load-bearing safety argument in the new comment — that an `autopilot <id>` pane is only ever the transient stop-classifier and never a live, re-adoptable session — checks out. The label is constructed only at `src/autopilot.ts:412` and `src/autopilot.ts:446` (`autopilot ${id}`), both passing it as the `label` arg to `classifyStop`, which is the sole `herdr.start` for it (`src/autopilot-llm.ts:107,142`) and always stops it in a `finally` (`src/autopilot-llm.ts:152-158`). Real sessions start under `ctx.name` (`src/service.ts:2095`) and are relabeled to a prompt-derived slug (`src/service.ts:3559`); `src/tab-reaper.ts:20-27` documents slugs as `[a-z0-9-]` — no space — so the space-suffixed prefix `"autopilot "` cannot collide with a session. This is precisely the concern the originating issue raised ("may still be a session that should be re-adopted"), and it is answered correctly.
- Label strings the tests hard-code exist verbatim on the producer side: `name ${session.desig}` at `src/service.ts:3527`, `"verify api key"` at `src/verify-key.ts:147`.
- The `readFileSync(join(import.meta.dir, "../src/index.ts"))` source-text guard follows the stated precedent — same pattern at `test/operator-language.test.ts:92` and `:298-299`. The regex `reapTransientByLabel\(\s*herdr\s*,\s*"<label>"` contains no unescaped metacharacters for any of the three labels, and matches the actual formatting at `src/index.ts:2073-2075`.
- `test/drain-landing-repair.test.ts:159` adding `preWarmEpicLandingCi: false` is required, not cosmetic: `RepoConfig.preWarmEpicLandingCi` is non-optional (`src/store.ts:182`), and the field was introduced by the merge-base commit a0446713. The fixture would not typecheck without it.

## The one real gap

The source-text guard pins the **consumer** (that `src/index.ts` still calls the reaper with those three literals) but nothing pins the **producer**. `SYNC_HELPER_LABELS` is a local literal in the test file, so if someone renames the spawn label at `src/autopilot.ts:412/446`, `src/service.ts:3527`, or `src/verify-key.ts:147`, the boot reap stops matching anything at runtime and *every test in this file still passes* — including the new guard, which would keep happily matching the now-dead literal in `index.ts`. That is the same silent-drift failure mode the guard was written to prevent, just moved one hop. The durable fix is a shared exported label constant (as `DISTILL_LABEL` / `OPTIMIZE_LABEL` / `MERGE_LABEL` already are) used by both the spawn site and the boot reap; a cheaper stopgap is a second source-text assertion against the producer files.

Secondary nit: the first `describe` block re-exercises a label-agnostic prefix scan with three different strings, so it is largely a restatement of the existing `"closes ALL unowned prefix matches"` test (`test/transient-tab-reaper.test.ts:38-46`). Its incremental value comes almost entirely from tying `SYNC_HELPER_LABELS` to the second block; fine to keep, but it is not independently pinning much.

No security, correctness, or latent-defect issues found in the diff.

scoop:
## Both critic points addressed — and point 1 was a real hole

**Point 1 (valid, and the important one).** Correct: the guard pinned only the *consumer* side. `SYNC_HELPER_LABELS` was a test-local literal, so renaming a label at its **producer** (`autopilot.ts`, `service.ts`, `verify-key.ts`) would silently desync the spawn from the boot reap — leak restored, every test still green. My guard would have watched the wrong end of the rope.

I took the **first** of the two options offered (shared exported constants) rather than the second (a producer-side source-text assertion), because it makes the binding **compile-enforced instead of asserted** — a stronger guarantee than any text match, and it follows the existing `DISTILL_LABEL`/`OPTIMIZE_LABEL`/`MERGE_LABEL` precedent exactly.

`b906ee06` exports `AUTOPILOT_LABEL` / `NAMER_LABEL` / `VERIFY_KEY_LABEL` from their producer modules and consumes them at **all three** sites:
- the spawn sites (`autopilot.ts` ×2, `service.ts`, `verify-key.ts`),
- the boot reap (`index.ts`),
- **and `tab-reaper.ts`'s `isShepherdHelperLabel`** — a *third* divergent copy of the same three literals the finding didn't name, but which would rot the same way. (House rule: update every mirror, don't leave divergent copies.)

A renamed label value now propagates everywhere by construction; desync is no longer expressible. The source-text guard consequently now matches the **constant identifier**, so it only has to catch the one thing types cannot: the boot call being **deleted outright**.

Verified both failure modes empirically, not just by inspection:
- **delete the boot call** → `boot-reaps \`autopilot \` (via AUTOPILOT_LABEL)` goes **red**;
- **rename the label value** (`"autopilot "` → `"ap "`) → spawn site and boot reap move **together**, tests stay green *and stay correct* — which is precisely the hole this finding identified, now closed.

**Point 2 (valid nit).** Folded. The three-test `describe` did mostly restate the label-agnostic `closes ALL unowned prefix matches`. It's now a single behavioral test that earns its place by pinning the **real label values** plus the empty-owned-set safety property (helper label reaps; a `[a-z0-9-]` session slug — which can never contain a space — is spared). The label→`index.ts` binding lives in the wiring guard, where it belongs.

## Scope correction to this PR

This PR is **no longer "test-only / zero runtime delta"** — the fix required touching `src/`. The runtime change is nonetheless **behavior-identical**: every label resolves to the same string it did before (`"autopilot "`, `"name "`, `"verify api key"`); only the *source of truth* moved from four scattered literals to one exported constant each. I've updated the PR description accordingly. Full suite: **6759 pass / 0 fail**, `tsc` clean, lint clean.

scoop (commented):
## Verdict

Both prior-round findings are genuinely addressed, and the runtime change is behavior-identical as claimed. Two non-blocking quality nits remain, both in `src/service.ts`.

## Prior findings — re-checked against the diff

**Prior finding 1 (producer side unpinned) — ADDRESSED.** The label values are now single-sourced exported constants consumed at every site I could find:
- producers: `src/autopilot.ts:36` (`AUTOPILOT_LABEL = "autopilot "`) consumed at `src/autopilot.ts:427` and `src/autopilot.ts:461`; `src/service.ts:1106` (`NAMER_LABEL = "name "`) consumed at `src/service.ts:3534`; `src/verify-key.ts:12` (`VERIFY_KEY_LABEL = "verify api key"`) consumed at `src/verify-key.ts:153`.
- boot reap: `src/index.ts:2073-2075` now passes the three constants instead of literals.
- husk filter: `src/tab-reaper.ts:58/60/64` now compares against the same constants.

I grepped `src/` for any surviving producer-side literal of these labels (`"autopilot `, `` `autopilot ``, `"name `, `verify api key`) and the only remaining occurrences are the constant definitions themselves plus prose in doc comments (`src/tab-reaper.ts:26-27,44,49-50`, `src/transient-tab-reaper.ts:16`). So a renamed value does propagate to spawn site, boot reap and husk filter by construction — the desync hole is closed. Values are byte-identical to the pre-diff literals (compared `src/tab-reaper.ts:58/60/64` and `src/index.ts:2073-2075` against the removed lines in the diff), so this is a no-op at runtime.

**Prior finding 2 (redundant describe) — ADDRESSED.** `test/transient-tab-reaper.test.ts:101-111` is now a single behavioral test over the three real label values that also asserts the spare-a-session-slug property, and the label→`index.ts` binding lives in the wiring guard (`test/transient-tab-reaper.test.ts:124-130`), which now matches on the constant identifier rather than a string literal.

## Other verification

- No import cycle introduced. `src/tab-reaper.ts` now imports from `./autopilot`, `./service`, `./verify-key`; I grepped for importers of `./tab-reaper` and the only one is `src/index.ts:45` — none of those three modules imports it back, so the new edges are acyclic.
- Reap semantics unchanged: `reapTransientByLabel` matches with `a.name.startsWith(labelPrefix)` (`src/transient-tab-reaper.ts:46`), so the space-less `VERIFY_KEY_LABEL` still matches its exact-equal pane name, and the empty owned set at boot behaves as before.
- `test/drain-landing-repair.test.ts:159` adds `preWarmEpicLandingCi: false` to the fixture — consistent with the config field introduced by a0446713; unrelated to #1147 but harmless.
- The boot reap for `autopilot `/`name `/`verify api key` already existed at the diff base (the diff modifies, not adds, those `src/index.ts` lines), so this PR's contribution to #1147 is the regression pinning plus removal of the divergent literal copies. That is a reasonable reading of the issue and I found nothing in it left unhandled by the combination.

## Nits (non-blocking, both in `src/service.ts`)

1. `src/service.ts:1096-1111` — the new `NAMER_LABEL` declaration was inserted *between* `DRAFT_PR_NOTE`'s doc comment and `DRAFT_PR_NOTE` itself. `DRAFT_PR_NOTE` (line 1108) now has no doc comment, and the draft-mode paragraph at 1096-1100 dangles above an unrelated constant.
2. `src/service.ts:1106` — `NAMER_LABEL` is homed in the *consumer* module rather than the namer's own module, unlike its three siblings (`DISTILL_LABEL` in `distiller.ts`, `AUTOPILOT_LABEL` in `autopilot.ts`, `VERIFY_KEY_LABEL` in `verify-key.ts`). The cost is that `src/tab-reaper.ts:8` must now import the ~4k-line `SessionService` module graph to get a five-character string. `src/namer.ts` has zero imports (verified: no `import` lines in the file) and `service.ts` already imports it at `src/service.ts:37`, so moving the constant there keeps the graph acyclic and cheap.

Nothing here is a correctness or security problem; the label binding does what the previous round asked for.

scoop:
## Both points valid — fixed in `ffbe14d7`. Point 2 subsumes point 1.

**Point 1 — dangling JSDoc.** Straightforwardly my error: I wedged `NAMER_LABEL` between `DRAFT_PR_NOTE`'s doc comment and `DRAFT_PR_NOTE` itself, so the draft-mode JSDoc floated above an unrelated constant and `DRAFT_PR_NOTE` was left undocumented.

**Point 2 — wrong module.** Also right, and it's the finding that matters: `NAMER_LABEL` sat in the **consumer** (`service.ts`), unlike `DISTILL_LABEL`/`AUTOPILOT_LABEL`/`VERIFY_KEY_LABEL`, which each live in their producer. That forced `tab-reaper.ts` to drag in the entire `SessionService` graph for a five-character string.

Rather than just relocate the declaration within `service.ts`, I took the suggestion and moved it to **`src/namer.ts`** — which fixes both findings at once, since the constant leaves `service.ts` entirely and the JSDoc adjacency is restored by construction.

`namer.ts` turned out to be the *right* home for a reason beyond import cost: it's where the `[a-z0-9-]` slug generation lives — i.e. the very rule that makes the label's trailing space collision-proof. The constant is now documented next to its own rationale, rather than asserting that rationale from a module that doesn't own it.

Verified:
- `tab-reaper.ts`'s imports are now leaf-only — no `./service` (confirmed by grep).
- `DRAFT_PR_NOTE` JSDoc reunited with its declaration.
- No behavior change — the label is still `"name "`.
- Guard re-verified end-to-end: deleting the namer boot call from `index.ts` still turns `boot-reaps \`name \` (via NAMER_LABEL)` **red**.
- `tsc` clean, lint clean, full suite **6759 pass / 0 fail**.

scoop (commented):
## Verdict

Both prior-revision findings are genuinely fixed, the label-constant refactor is behavior-preserving, and the new regression guard actually guards the thing it claims to. Two non-blocking nits.

## Prior findings — re-checked against the diff

1. **Dangling `DRAFT_PR_NOTE` JSDoc — FIXED.** `src/service.ts:1096-1104` now has the draft-mode JSDoc directly above `DRAFT_PR_NOTE`; no constant is wedged between them. The only remaining `service.ts` changes in the diff are the `NAMER_LABEL` import (src/service.ts:37) and its use at src/service.ts:3527.
2. **`NAMER_LABEL` in the consumer module — FIXED.** It now lives in `src/namer.ts:16`, the same module that owns the `[a-z0-9-]` slug rule (src/namer.ts:389) that the label's trailing space depends on. `src/tab-reaper.ts` no longer imports `./service` directly (verified: the only imports at src/tab-reaper.ts:1-10 are herdr/usage-probe/distiller/optimizer/merge-suggest/autopilot/namer/verify-key/json-tolerant).

## Verification performed

- **No literal spawn labels left.** Grepped for `"name `/`autopilot `/`verify api key` across `src/`: every remaining hit is a doc comment or the constant declaration itself (src/namer.ts:16, src/autopilot.ts:36, src/verify-key.ts:12). Both `consider()` call sites (src/autopilot.ts:427, src/autopilot.ts:461) and the verify spawn (src/verify-key.ts:153) bind the constants; values are unchanged (`"autopilot "`, `"name "`, `"verify api key"`), so this is a pure refactor.
- **The `autopilot <id>` label really is classifier-only.** `consider()`'s `label` param is passed straight to `this.deps.classify` (src/autopilot.ts:384) → `classifyStop`, which uses it only as the `herdr.start` pane name (src/autopilot-llm.ts:103-147). No other site constructs an `autopilot `-prefixed pane name, so the empty-owned-set boot reap cannot hit a live/re-adoptable session — the central safety claim of #1147.
- **Space-prefix collision-safety holds.** Session names go through `slugifyManual` (src/namer.ts:386-394), which collapses every non-`[a-z0-9]` run into `-`; a session slug can never contain a space, so `startsWith("autopilot ")` / `startsWith("name ")` in `reapTransientByLabel` (src/transient-tab-reaper.ts:46) can't match a user session.
- **`VERIFY_KEY_LABEL` comparison semantics preserved.** src/tab-reaper.ts:58 keeps the exact-equality form (`label === VERIFY_KEY_LABEL`) it had as a literal, while the prefix labels keep `startsWith` — correct, since `verify api key` takes no suffix.
- **Boot-reap guard is real.** The regex in test/transient-tab-reaper.test.ts matches on the identifier (`reapTransientByLabel\(\s*herdr\s*,\s*NAMER_LABEL\b`), which does match src/index.ts:2074-2076 as written; deleting a boot call fails the test. Note the functional boot reap already existed at the diff base (verified: `git show 21caec44:src/index.ts` lines 2073-2075 already called `reapTransientByLabel` with the three literals) — this PR's contribution is the desync-proofing plus that guard, which is a coherent slice of #1147, not a gap.
- **`test/drain-landing-repair.test.ts`** adds `preWarmEpicLandingCi: false` to the repo-config fixture; the field is a required `boolean` on the store's repo config (src/store.ts:182), so this is a needed type fix, not dead noise.

## Nits (see findings)

The import-layering rationale that motivated moving `NAMER_LABEL` out of `service.ts` is undercut by the sibling import added in the same commit: `src/tab-reaper.ts:7` imports `AUTOPILOT_LABEL` from `./autopilot`, and `src/autopilot.ts:9` is a *value* import of `DRAFT_PR_NOTE` from `./service`. So `tab-reaper.ts` still transitively pulls in the whole `SessionService` graph — the direct-import grep in the author note is true but not the property that matters. No cycle results (nothing in the `service.ts` graph imports `tab-reaper.ts`; only src/index.ts:45 does), so this is layering/doc accuracy, not a bug.

scoop:
## Both points valid — fixed in `ea5d24cf`. Point 1 was a false claim, and it's worse than the finding says.

**Point 1 — the import-isolation claim.** Correct, and this one stings: I asserted in the JSDoc that `tab-reaper.ts` "gets the string without pulling in the whole SessionService graph" without ever verifying it. It's false — `tab-reaper → autopilot → service` (a *value* import of `DRAFT_PR_NOTE`).

Checking properly, it's broader than the finding states: `tab-reaper.ts` also value-imports `distiller`/`optimizer`/`merge-suggest`, each of which pulls `herdr`. So `tab-reaper` was **never** leaf-pure, and my import-weight framing was wrong from the start — not merely imprecise. I'd inherited it from the previous round's "cheaper home" argument and repeated it without checking.

Of the two remedies offered I took **"soften the claim"** rather than **"make it true"**, deliberately:

- Making it true requires parking *every* helper label in a shared leaf module — all seven, since `DISTILL_LABEL`/`OPTIMIZE_LABEL`/`MERGE_LABEL` are the heavier edges — or relocating `DRAFT_PR_NOTE` out of `service.ts`. That's a 7-module refactor that also abandons the producer-module precedent this same review approvingly cited two rounds ago.
- It's well outside #1147's scope, and the import weight buys nothing here: `tab-reaper` is not on a hot path, and nothing depends on its graph being narrow.

So the doc now states what's actually true — the placement is about **single-source-of-truth** and colocating each label with the rule that makes it collision-proof (the `[a-z0-9-]` slug rule genuinely lives in `namer.ts`) — plus an explicit `NOTE:` that it is **not** a dependency-isolation guarantee and shouldn't be read as one. If someone later wants real graph isolation, that's a clean standalone change, not a claim smuggled into a comment.

**Point 2 — broken `{@link}` refs.** Valid nit, fixed. The cross-module `{@link DISTILL_LABEL}` etc. named symbols those modules never import, so they don't resolve in quick-info. Now plain prose naming the declaring module (`DISTILL_LABEL (distiller.ts)`, …). The `{@link}`s remaining in `namer.ts`/`verify-key.ts` are pre-existing **same-module** refs (`normalize`, `SENTINEL`, `VERIFY_FILE`, `matchAuthError`) and resolve correctly — left alone.

Docs-only change; no behavior delta. `tsc` clean, lint clean, full suite **6759 pass / 0 fail**.

scoop (commented):
## Verdict

Both previous-round findings are genuinely addressed by the new diff, and the label-constant refactor is correct and consistent everywhere I could check. No blocking issues.

## Prior findings — re-checked against the diff

**1. `src/namer.ts` false import-isolation claim — FIXED.** The JSDoc no longer asserts leaf-purity; `src/namer.ts:16-21` now carries an explicit `NOTE:` stating that the placement is about single-source-of-truth and *not* import weight, and it names the real edge (`tab-reaper → autopilot → service`, for `DRAFT_PR_NOTE`). I verified that disclaimer is itself accurate rather than just plausible: `src/tab-reaper.ts:7` imports `AUTOPILOT_LABEL` from `./autopilot`, and `src/autopilot.ts:9` value-imports `DRAFT_PR_NOTE` from `./service`. `src/tab-reaper.ts:4-6` also value-imports distiller/optimizer/merge-suggest, as the note says. The doc now describes the code.

**2. Broken cross-module `{@link}` refs — FIXED.** `src/autopilot.ts:34-35` now uses plain prose with declaring modules (`DISTILL_LABEL (distiller.ts)`, `OPTIMIZE_LABEL (optimizer.ts)`, `MERGE_LABEL (merge-suggest.ts)`), and `src/namer.ts:13-16` likewise. The `{@link}`s that remain are same-module (e.g. `{@link VERIFY_FILE}` at `src/verify-key.ts:16`, and `{@link PROBE_NAME}`/`{@link DISTILL_LABEL}` at `src/tab-reaper.ts:24-25` — both of which *are* imported in that module at lines 3-4, so they resolve).

## What I verified in the substantive change

- **No literal label survives at a spawn site.** Grepping `src/` for `autopilot `/`name $\{`/`verify api key` string forms leaves only the constant declarations (`src/autopilot.ts:37`, `src/namer.ts:22`, `src/verify-key.ts:12`) and prose in comments. Both `consider(...)` call sites are converted (`src/autopilot.ts:428`, `src/autopilot.ts:462`) and those are the only two; `src/service.ts:3527` and `src/verify-key.ts:153` use the constants.
- **The label really is the pane name.** `src/index.ts:1501-1514` passes `label` straight into `classifyStop`, which passes it to `deps.herdr.start(label, …)` at `src/autopilot-llm.ts:142-143`. So the doc claim that an `autopilot <id>` pane is only ever the transient classifier holds for every producer in the tree.
- **The "space-prefixed labels can't collide with a session slug" premise is real**, not asserted: `slugifyManual` at `src/namer.ts:392-399` reduces to `[a-z0-9-]` (`.replace(/[^a-z0-9]+/g, "-")`), so a space-bearing label is unreachable by a slug. That is what makes the empty owned set at `src/index.ts:2074-2076` safe.
- **Reaper semantics match the tests.** `reapTransientByLabel` matches by `a.name.startsWith(labelPrefix)` (`src/transient-tab-reaper.ts:46`), so the new test's synthetic `${label}orphaned` names and the exact-phrase `VERIFY_KEY_LABEL` both behave as the test expects; `test/transient-tab-reaper.test.ts:96` asserting only `["tabH"]` closed (sparing the `fix-the-thing` slug) is the right property.
- **No import cycle introduced.** `tab-reaper.ts` now imports `autopilot`/`namer`/`verify-key`; nothing in those graphs imports `tab-reaper` (the only importer is `src/index.ts:45`), and `namer.ts` has no imports at all. The `autopilot → service` edge already existed at base.
- **`src/tab-reaper.ts:58` `label === VERIFY_KEY_LABEL`** preserves the pre-existing exact-match (not prefix) semantics for that one label — correct, since `VERIFY_KEY_LABEL` has no trailing space.
- **`test/drain-landing-repair.test.ts:159`** adds `preWarmEpicLandingCi: false`; the field is a required member of the repo-config row shape (`src/store.ts:182`), so this is a legitimate fixture compile fix, not a behavior change.
- **Task fit:** the boot reap for `autopilot `/`name `/`verify api key` already existed at the base commit (landed in `adb08b08`, #1136 — confirmed with `git log -S`). This PR's contribution to #1147 is binding the three labels to shared exported constants so a rename can't desync spawn from reap, plus a regression guard that the boot calls still exist (`test/transient-tab-reaper.test.ts:120-129` greps `src/index.ts` for `reapTransientByLabel(herdr, <IDENT>`, which matches the actual call sites at `src/index.ts:2074-2076`). That's a reasonable read of the issue's "is a persisted spawn row the right primitive" investigation, and the source-text guard is the only way to pin a bare `void` statement inside `deferredStarts`.

## Nit (non-blocking)

The `isShepherdHelperLabel` JSDoc at `src/tab-reaper.ts:26-27` and the helper inventory at `src/tab-reaper.ts:44,49,50` still spell the labels as hardcoded literals (`"name "`, `"autopilot "`, `"verify api key"`) in prose, immediately above code that now binds the constants. That's exactly the desync this PR set out to eliminate, one layer up — worth naming the constants in the doc instead.

## Stated limitations

I did not run tests, lint, or `tsc` (CI owns those, per instructions); claims above are from reading the tree, not from execution.

scoop (commented):
## What this PR actually does

The functional boot reap for `autopilot ` was NOT added here — it already exists at the review base: `git show 21caec44:src/index.ts` lines 2073-2075 already call `reapTransientByLabel(herdr, "autopilot ", …)`, introduced by `adb08b08` (#1136) (`git log -S'reapTransientByLabel(herdr, "autopilot ' -- src/index.ts`). What this PR contributes for #1147 is the hardening around it: the three synchronous-helper labels are lifted into exported constants at their producers (`src/namer.ts:22` `NAMER_LABEL`, `src/autopilot.ts:37` `AUTOPILOT_LABEL`, `src/verify-key.ts:12` `VERIFY_KEY_LABEL`), every producer and consumer binds to them, and `test/transient-tab-reaper.test.ts` pins both the reap behaviour and the fact that `src/index.ts` still wires the three boot calls. That is a coherent answer to the issue's "is a name scan the right primitive here" question, and the safety argument it rests on holds (see below).

## Prior finding

1. **`src/tab-reaper.ts` JSDoc hardcoding label strings — ADDRESSED.** Commit `3586e85d` rewrites the doc block: the collision-proof paragraph now names `{@link NAMER_LABEL}` / `{@link AUTOPILOT_LABEL}` / `{@link VERIFY_KEY_LABEL}` (src/tab-reaper.ts:26-28) and the inventory uses `{@link …}` for every label that has a constant (src/tab-reaper.ts:45-51), keeping quoted literals only for the labels still inlined at their spawn sites (`review `, `plan-review `, `pr-critic `, `recap `, `rundown`). All six linked symbols are actually imported into the module (src/tab-reaper.ts:3-9), so the `{@link}` refs resolve.

## Verification

- **Every producer now binds to the constant.** Grepping `AUTOPILOT_LABEL|\`autopilot ` across the tree finds no remaining literal spawn label: the only two producers are src/autopilot.ts:428 and src/autopilot.ts:462, both `${AUTOPILOT_LABEL}${id}`; the label is threaded `consider(id, tail, label)` (src/autopilot.ts:374) → `classify` (src/index.ts:1501-1513) → `classifyStop(…, label)` → `deps.herdr.start(label, …)` (src/autopilot-llm.ts:107, 142-143). So the doc's "autopilot stop-classifier (autopilot-llm.ts)" attribution at src/tab-reaper.ts:50 is accurate. Same for `NAMER_LABEL` (single producer src/service.ts:3527) and `VERIFY_KEY_LABEL` (single producer src/verify-key.ts:153).
- **Boot reap consumes the same constants**: src/index.ts:2074-2076, imported at src/index.ts:89, 90, 161. The source-text guard in the new test matches `reapTransientByLabel\(\s*herdr\s*,\s*<IDENT>\b`, which matches those three lines verbatim.
- **Empty-owned-set safety claim holds.** `reapTransientByLabel` closes every prefix match not in the owned set (src/transient-tab-reaper.ts:45-50). The claim that no user session can carry a space-prefixed name is real: both the auto namer and the manual rename route through `slugifyManual` (src/namer.ts:392, used at src/server.ts:2663 for the rename endpoint and src/service.ts:2865), which reduces to `[a-z0-9-]`. `"verify api key"` is matched as a prefix of itself, so the exact-phrase label reaps correctly too.
- **No import cycle introduced.** `src/tab-reaper.ts` now imports `./autopilot` (→ `./service` for `DRAFT_PR_NOTE`), `./namer` (a leaf — no imports at all) and `./verify-key`. `tab-reaper` itself is imported only by `src/index.ts:45` and `test/tab-reaper.test.ts:8`, and nothing in the `service` graph imports it back, so `tab-reaper → autopilot → service` terminates. The note at src/namer.ts:16-20 correctly declines to claim any dependency-isolation benefit.
- **`test/drain-landing-repair.test.ts:159`** adds `preWarmEpicLandingCi: false` to the repo-config fixture; the field is a required `boolean` on the store's repo config (src/store.ts:182), so this is a legitimate type fix, not scope creep.

Nothing in the diff is a correctness or security problem. One documentation nit below.

## Nit

**src/tab-reaper.ts:24-28** — the rewritten "Collision-proof markers" paragraph still enumerates only `PROBE_NAME`, `DISTILL_LABEL` and `OPTIMIZE_LABEL` as the underscore-bearing labels, then says "*every other* helper uses a space-prefixed label … or a multi-word exact phrase". `MERGE_LABEL` is `__merge__` (matched by prefix at src/tab-reaper.ts:62 and listed in the inventory at src/tab-reaper.ts:48) — it is neither space-prefixed nor a multi-word phrase, so it falls into the "every other helper" bucket where it does not belong. Since this PR's stated purpose is exactly to stop this doc block from describing the code inaccurately, adding `{@link MERGE_LABEL}` to the underscore list closes the last gap.

PR #1675, opened 2026-07-12
fix(reaper): reap runaway orphans by env provenance, not cwd (#1144)

### LINKED ISSUE #1144: Resource-gated orphan reaper for cwd-blind / non-isolated leaks (follow-up to #1133)
Follow-up to #1133 (PR #1143). That PR reaps PPID-1 orphaned background jobs (leaked `yes`/load-gen busy-loops) via two cwd-scoped sweeps — on worktree teardown, and a boot/daily safety net for already-deleted worktree cwds — plus a spawn-directive guardrail. It deliberately left three residual gaps that cwd-scoping can't cover. This issue tracks the robust catch-all for the two that matter.

## Residual gaps from #1143

1. **chdir-out orphans** — the only residual that reproduces #817 exactly. A PPID-1 orphan that `cd`s *out* of the worktree before/after backgrounding has a cwd that is neither under the worktree (misses the teardown sweep) nor a deleted `.shepherd-worktrees/*` path (misses the safety net). Already mitigated from three sides — the autonomous **membrane** fully contains these via bwrap `--unshare-pid` (PID namespace), the new `ENGINEERING_POSTURE` guardrail discourages the behavior at the source, and it requires an agent to chdir out *then* background a busy-loop — so it's lower-probability, but high-severity if it lands (the #817 leak burned ~30 cores for 6 days undetected).

2. **Non-isolated sessions get no protection.** `service.ts` calls `worktree.remove()` only `if (s.isolated)`; a non-isolated session's `worktreePath` *is* the shared repo root, which we deliberately never sweep (it hosts the shepherd server + unrelated host processes — a PPID-1 cwd-sweep there would be more dangerous than the leak). Non-isolated is itself the rare worktree-add-failed fallback path, but the leak is unprotected when it happens.

(The third #1143 residual — abandoned-but-on-disk worktrees — is **not** in scope here: it largely self-heals, since once the worktree is removed by any path its cwd becomes `(deleted)` and the existing safety net reaps it on the next sweep.)

## Proposed approach — resource-gated orphan reaper

Both gaps share one robust fix: a periodic sweep that reaps **PPID-1 processes burning sustained high CPU** whose cwd is (or was) under any shepherd-managed path, **independent of which worktree** — keying on actual resource burn rather than cwd-under-this-exact-worktree. This catches cwd-blind and non-isolated leaks that the cwd heuristic can't, while the CPU gate bounds the false-positive surface (a quiet legit background process is never touched).

Design considerations / open questions for whoever picks this up:
- **False-positive surface.** The shared repo root hosts the shepherd server and unrelated host processes; a CPU + PPID-1 + duration gate must be conservative enough never to reap a legitimate operator process. Decide the CPU% threshold, the sustained-duration window, and whether to restrict to procs whose cwd resolves (now or via `(deleted)`) under a known shepherd parent dir.
- **`PPID === 1` assumption.** Holds only for host-reparented trusted-path spawns; membrane orphans are already PPID ≠ 1 (namespace-reaped) and out of scope. A future subreaper / PID-namespace on the *trusted* path would change this — see the Risks note in `src/process-reaper.ts`.
- **Reuse.** Build on the existing `ProcessReaper` + injectable `ReaperProbes` machinery and the new `ppidForPid` probe from #1143.

## Cross-link

Part of the same "scaffolding outlives the agent" / "reap-on-teardown/boot" cluster @scoop flagged on #1133: this issue (#1133's cwd-blind residual) alongside #1135 (idle-but-alive helper `claude` not boot-reaped) and #1136 (herdr helper tabs pile up). Those husks run with cwd under `/tmp/shepherd-{autopilot,distill}-*`, not a worktree — distinct concrete leaks, but worth a shared design pass on a unified resource/liveness-gated reaper.

Priority: low. The high-value 80% (common-case teardown + deleted-cwd safety net + guardrail + membrane containment) shipped in #1143.

scoop:
Closes #1144.

## The problem with cwd

#1143 reaps PPID-1 orphans by **cwd**. That structurally cannot see the two gaps #1144 tracks:

- **chdir-out orphans** — a background job that `cd`s out of the worktree has a cwd that is neither under it nor a deleted `.shepherd-worktrees/*` path.
- **non-isolated sessions** — their `worktreePath` **is** the shared repo root, which we must never cwd-sweep: it hosts the server and the operator's own processes. `reapOrphansUnder` hangs off `WorktreeManager.remove()`, which teardown calls only `if (s.isolated)` — so that path gets **no teardown reap at all**.

Both are symptoms of one thing: **cwd is a guess.** Attributing a runaway process by which directory it happens to sit in can't be made both safe and complete.

## Attribute by provenance instead

Every agent spawn now carries `SHEPHERD_SESSION_ID`. Every descendant inherits it, and `/proc/<pid>/environ` is fixed at `exec` — so the marker survives `cd`, backgrounding, PID-1 reparenting, and worktree deletion. Attribution becomes **exact**:

- An operator's `rust-analyzer`, `nohup cargo bench &`, or `ffmpeg` **never carries the marker**, so it can never be a candidate. The false-positive surface isn't small — it's structurally empty.
- Both gaps close, and the sweep is safe to ship **armed**.

This deletes the roots, `.git`/depth/`$HOME` fences, and birth windows that a cwd-based design needs.

## Safety is `provenance ∧ terminality`

A process is reaped only if it carries the marker **and** its session row is **present and `archived`**.

Terminality is what spares an agent's **in-flight `cargo build &`**: a one-shot Bash tool call reparents its background job to PID 1 the moment that call's shell exits — while the agent is still working. **PPID-1 means signal-unmanageable, not abandoned.**

"Row **absent** ⇒ **spare**", because `SHEPHERD_DB`/`SHEPHERD_PORT` make a second Shepherd on one host supported — reaping on absent would let instance A SIGKILL instance B's *live* session's work.

## Notable gates

- **`cutime`/`cstime` in the CPU numerator.** A `while true; do <work>; done &` supervisor sits blocked in `wait()` with its own utime/stime ≈ 0 while each child dies *under* the age floor. Without the reaped-children columns, the commonest shape of this leak is never caught. A test asserts it's missed if they're dropped.
- **git spares.** An agent's `git fetch` triggers a detached `git gc --auto` that **inherits the marker**, is non-listening, and pegs a core from birth — it clears every other gate. SIGKILLing it strands a stale `gc.pid` lock and the repo is then never gc'd. (`comm` truncates at 15 chars, so the `git-` **prefix** rule is what actually covers `git-pack-objects`.)
- **`environ` is read LAST.** It goes through `access_remote_vm` and takes the *target's* mmap lock, so it can block on a stalled process. Gates run cheapest-first; on a host of idle pids the mmap lock is never taken.
- **PPID is logged, not gated.** Requiring it costs coverage (a marked child of a live marked wrapper has PPID ≠ 1) and would make this a **no-op in a container** where Shepherd is PID 1.

## Where it runs

At **teardown** via `archive()` — the only teardown reap a non-isolated session has ever had — plus boot + hourly. `archiveMany` sweeps **once** for the batch, not once per id (the sweep enumerates every pid on the host).

Armed by default; `SHEPHERD_REAP_RUNAWAY=0` disables, `=observe` logs only.

## Verification

`bun run lint`, `bun run typecheck`, `bun run test` (6781 pass), and `fallow audit` all green. Beyond the unit tests (which use injected probes), the **real `/proc` probes were exercised end-to-end** against a live marked busy-loop: read back at `cpuFraction: 1`, and the armed sweep actually SIGKILLed it. Notably its `ppid` was **not** 1 — confirming the PPID demotion is load-bearing.

## Known bounds (deliberate)

- **Reaps only once a session is `archived`.** A runaway in a session parked in `done`/`idle`/`blocked` is spared — still the shape of #817's undetected burn. Terminality is what makes the sweep safe *and* what bounds it.
- **The CPU/age gate is a performance prefilter, not a safety floor** — it's what keeps the mmap-lock reads near zero. It silently costs coverage (late-onset spin, anything under 80%).
- **The marker is evadable** (`env -u SHEPHERD_SESSION_ID`) since `buildWrappedArgv` renders env as argv tokens. Fail-closed — evasion *spares*. This is a safety net against accident, not a sandbox against intent.

**The real fix for all three** is exact attribution via **cgroup-per-session (`systemd-run --user --scope`) or `PR_SET_CHILD_SUBREAPER`** — it would retire the marker, the CPU gate, and this whole sweep, and would also cover the #1135/#1136 husks. It's out of proportion to a priority:low follow-up (systemd-user dependency / `prctl(2)` via FFI, rewriting every spawn path), so it's recommended as its own issue. **This PR is the interim measure.**

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -408,6 +408,22 @@ export function parseHour(raw: string | undefined, def: number): number {
   return Number.isInteger(n) && n >= 0 && n <= 23 ? n : def;
 }
 
+/** Mode for the runaway-orphan reaper (issue #1144). Mirrors `ReapMarkedOptions["mode"]`. */
+type ReapRunawayMode = "armed" | "observe" | "off";
+
+/**
+ * `SHEPHERD_REAP_RUNAWAY`: `0`/`off` disables the sweep entirely; `observe` runs every gate but
+ * never signals (log-only). Anything else — including unset — is `armed`, the default: the sweep
+ * is safe to arm because a candidate must carry an agent-spawned env marker AND belong to an
+ * archived session before it can be touched.
+ */
+function normalizeReapRunaway(raw: string | undefined): ReapRunawayMode {
+  const v = raw?.trim().toLowerCase();
+  if (v === "0" || v === "off") return "off";
+  if (v === "observe") return "observe";
+  return "armed";
+}
+
 // The HUD's main listen port. Extracted so the agent-ingress port can default
 // relative to it (mainPort + 1) — a custom SHEPHERD_PORT shifts both in lockstep.
 const mainPort = Number(process.env.SHEPHERD_PORT ?? 7330);
@@ -656,6 +672,19 @@ export const config = {
   // on, with this flag as the kill switch. UI-configurable + persisted; set
   // SHEPHERD_SESSION_HOUSEKEEPING=0 to seed it off on a fresh DB.
   sessionHousekeepingEnabled: process.env.SHEPHERD_SESSION_HOUSEKEEPING !== "0",
+  // Runaway-orphan reaper (issue #1144). SIGKILLs a process that (a) carries this session's
+  // SHEPHERD_SESSION_ID in its /proc/<pid>/environ (provenance — an agent spawned it) AND (b) whose
+  // session row is present and `archived` (terminality — the agent is definitively done), once it has
+  // burned `reapRunawayMinCpu` of a core over `reapRunawayMinAgeS`. Safety comes from provenance ∧
+  // terminality; the CPU/age pair is a PERFORMANCE prefilter that keeps the sweep's /proc/<pid>/environ
+  // reads (which take the target's mmap lock) near zero — NOT a safety floor.
+  //   armed (default) → SIGKILL · observe → log-only · off → the sweep never runs.
+  reapRunaway: normalizeReapRunaway(process.env.SHEPHERD_REAP_RUNAWAY),
+  // Fraction of ONE core, averaged over the process's whole lifetime (incl. reaped children).
+  reapRunawayMinCpu: Number(process.env.SHEPHERD_REAP_RUNAWAY_MIN_CPU ?? 0.8),
+  // Minimum process age before it can be reaped. Also what makes the benign `restore()` race
+  // (agent respawned before the row flips off `archived`) unreachable — keep it comfortably > 0.
+  reapRunawayMinAgeS: Number(process.env.SHEPHERD_REAP_RUNAWAY_MIN_AGE_S ?? 300),
   // LLM session naming: after a session is created with the instant heuristic name,
   // a transient haiku agent comprehends the prompt and renames it in the background.
   // Default on; set SHEPHERD_LLM_NAMING=0 to keep the pure-heuristic name.
--- a/src/config.ts
+++ b/src/config.ts
@@ -118,6 +118,13 @@ export function clampCap(n: number, min: number, max: number, fallback: number):
   return Math.min(max, Math.max(min, Math.round(n)));
 }
 
+// As clampCap, but WITHOUT the integer rounding — for ratio knobs (e.g. a CPU fraction, where
+// clampCap would round 0.8 to 1). Same forgiving contract: non-finite falls back, else snap.
+export function clampFraction(n: number, min: number, max: number, fallback: number): number {
+  if (!Number.isFinite(n)) return fallback;
+  return Math.min(max, Math.max(min, n));
+}
+
 // ── preview-port range ─────────────────────────────────────────────────────
 // Single source of truth for the preview-port range; consumed by both the slot
 // allocator (future PreviewService) and checkOrigin (origin hardening). The range
@@ -681,10 +688,25 @@ export const config = {
   //   armed (default) → SIGKILL · observe → log-only · off → the sweep never runs.
   reapRunaway: normalizeReapRunaway(process.env.SHEPHERD_REAP_RUNAWAY),
   // Fraction of ONE core, averaged over the process's whole lifetime (incl. reaped children).
-  reapRunawayMinCpu: Number(process.env.SHEPHERD_REAP_RUNAWAY_MIN_CPU ?? 0.8),
-  // Minimum process age before it can be reaped. Also what makes the benign `restore()` race
-  // (agent respawned before the row flips off `archived`) unreachable — keep it comfortably > 0.
-  reapRunawayMinAgeS: Number(process.env.SHEPHERD_REAP_RUNAWAY_MIN_AGE_S ?? 300),
+  // CLAMPED, not just defaulted: `Number("")` is 0, so an env var that is merely SET-BUT-EMPTY
+  // would otherwise silently drop the gate to "any marked process of an archived session".
+  reapRunawayMinCpu: clampFraction(
+    Number(process.env.SHEPHERD_REAP_RUNAWAY_MIN_CPU ?? 0.8),
+    0.05,
+    1,
+    0.8,
+  ),
+  // Minimum process age before it can be reaped. This floor is what makes the benign `restore()`
+  // race unreachable — restore SPAWNS the agent before `store.unarchive` flips the row off
+  // `archived`, so a live agent's row briefly reads archived. Clamped to a hard >= 60s minimum:
+  // `Number("")` is 0, and a 0 floor would open exactly that window (and reap the freshly
+  // respawned agent's own children). A comment cannot enforce this; the clamp does.
+  reapRunawayMinAgeS: clampCap(
+    Number(process.env.SHEPHERD_REAP_RUNAWAY_MIN_AGE_S ?? 300),
+    60,
+    24 * 60 * 60,
+    300,
+  ),
   // LLM session naming: after a session is created with the instant heuristic name,
   // a transient haiku agent comprehends the prompt and renames it in the background.
   // Default on; set SHEPHERD_LLM_NAMING=0 to keep the pure-heuristic name.
--- a/src/index.ts
+++ b/src/index.ts
@@ -100,7 +100,7 @@ import { CountsService } from "./backlog";
 import { OpenPrSnapshotService } from "./open-pr-snapshot";
 import { BacklogPoller } from "./backlog-poller";
 import { UpNextService, buildUpNextRepos } from "./up-next";
-import { ProcessReaper, reapDeletedWorktreeOrphans } from "./process-reaper";
+import { ProcessReaper, reapDeletedWorktreeOrphans, reapMarkedOrphans } from "./process-reaper";
 import {
   sweepClaudeTmp,
   compileCacheDir,
@@ -625,6 +625,8 @@ const service = new SessionService({
     : undefined,
   events,
   reaper: new ProcessReaper(),
+  // #1144: reap what the agent left burning, for the ids just archived. Defined below.
+  reapRunaway: (ids) => sweepRunawayOrphans(ids),
   preview: previewService,
   // Fast-poll a queue member to surface its merge promptly when the train session
   // archives before the 120s PR sweep credits it (the merge-train completion race).
@@ -732,6 +734,55 @@ const fireTmpSweep = (phase: "boot" | "daily") => {
 
 deferredStarts.push(() => {
   fireTmpSweep("boot");
+});
+
+/**
+ * #1144: SIGKILL processes an agent left burning a core after its session was archived.
+ *
+ * Attribution is by ENV MARKER, not cwd: every agent spawn carries SHEPHERD_SESSION_ID, every
+ * descendant inherits it, and /proc/<pid>/environ survives `cd`, backgrounding, PID-1 reparenting
+ * and worktree deletion. So this catches the two leaks #1143's cwd sweeps structurally cannot —
+ * an orphan that chdir'd out of the worktree, and a NON-isolated session's orphan (whose "worktree"
+ * IS the shared repo root, which we must never sweep by cwd). An operator's own hot processes never
+ * carry the marker, so they can never be candidates.
+ *
+ * Runs at teardown (via SessionService.reapRunaway), on boot, and hourly as the safety net for a
+ * leak that was still under the age floor when its session was archived.
+ */
+const sweepRunawayOrphans = (ids?: Set<string>) => {
+  if (maintenance.active) return; // a sync /proc sweep must not run mid-update
+  try {
+    const { reaped, observed } = reapMarkedOrphans({
+      ids,
+      mode: config.reapRunaway,
+      minCpu: config.reapRunawayMinCpu,
+      minAgeS: config.reapRunawayMinAgeS,
+      // MUST throw rather than report "absent" if the store is ever unreadable: "absent" is a
+      // positive claim that no such session exists HERE (which is how a second Shepherd instance's
+      // live sessions are spared), so a failed read must never be mistaken for one.
+      sessionStatus: (id) => {
+        const s = store.get(id);
+        if (!s) return "absent";
+        return s.status === "archived" ? "archived" : "live";
+      },
+    });
+    for (const c of observed) {
+      const pct = Math.round(c.cpuFraction * 100);
+      const mins = Math.round(c.ageSeconds / 60);
+      console.warn(
+        `[reap-runaway] ${config.reapRunaway === "armed" ? "killed" : "would kill"} pid ${c.pid} ` +
+          `(${c.comm}) session=${c.sessionId} cpu=${pct}% age=${mins}m ppid=${c.ppid} cwd=${c.cwd}`,
+      );
+    }
+    if (reaped > 0) console.warn(`[reap-runaway] reaped ${reaped} runaway orphan(s)`);
+  } catch (err) {
+    console.warn("[reap-runaway] sweep failed:", err);
+  }
+};
+
+deferredStarts.push(() => {
+  sweepRunawayOrphans();
+  setInterval(() => sweepRunawayOrphans(), 60 * 60 * 1000);
 });
 
 // Reap orphaned helper tabs (usage-probe / review husks no live agent backs). The
--- a/src/index.ts
+++ b/src/index.ts
@@ -808,7 +808,14 @@ const sweepRunawayOrphans = (ids?: Set<string>) => {
 };
 
 deferredStarts.push(() => {
-  sweepRunawayOrphans();
+  // Deferred off the synchronous boot path (same as the #1133 orphan reap above): the sweep is a
+  // host-wide /proc enumeration — a readdir plus a comm + stat read per pid, and an fd-table walk
+  // for every hot, old survivor — and it runs on Bun's single loop. Nothing about it is urgent at
+  // boot (it is the safety net; teardown already swept these), so it must not sit inline ahead of
+  // the server accepting connections.
+  void Promise.resolve()
+    .then(() => sweepRunawayOrphans())
+    .catch((err) => console.warn("[reap-runaway] boot sweep failed:", err));
   setInterval(() => sweepRunawayOrphans(), 60 * 60 * 1000);
 });
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -100,7 +100,12 @@ import { CountsService } from "./backlog";
 import { OpenPrSnapshotService } from "./open-pr-snapshot";
 import { BacklogPoller } from "./backlog-poller";
 import { UpNextService, buildUpNextRepos } from "./up-next";
-import { ProcessReaper, reapDeletedWorktreeOrphans, reapMarkedOrphans } from "./process-reaper";
+import {
+  ProcessReaper,
+  reapDeletedWorktreeOrphans,
+  reapMarkedOrphans,
+  SESSION_MARKER_ENV,
+} from "./process-reaper";
 import {
   sweepClaudeTmp,
   compileCacheDir,
@@ -172,6 +177,18 @@ import {
 import { fetchGithubRateLimit } from "./forge/github-rate-limit";
 
 const execFileAsync = promisify(execFile);
+
+// Scrub any INHERITED session marker before anything can spawn (issue #1144).
+//
+// The marker means "an agent of session <id> spawned this process". If Shepherd itself is launched
+// from inside an agent session (an agent restarting the server, a dev run from a session shell), it
+// inherits that id in its own process.env — and hands it, by plain env inheritance, to every child
+// it spawns. The aux/satellite agents (critic, distiller, doc-agent, plan-gate) pass no marker of
+// their own precisely so that they stay unmarked and are ALWAYS spared; a stale inherited one would
+// silently mark them with a real session id, and once that session archived the reaper would treat
+// a live critic's background work as a runaway. Delete it once, here, at the top of boot: the
+// server is not an agent's child in any sense the reaper should honour.
+delete process.env[SESSION_MARKER_ENV];
 
 startLoopLagSampler(); // no-op unless SHEPHERD_PROFILE_LOOP=1
 logRemainingOnLoopBlockers(); // one-time operator map of intentionally-sync calls
@@ -752,7 +769,7 @@ deferredStarts.push(() => {
 const sweepRunawayOrphans = (ids?: Set<string>) => {
   if (maintenance.active) return; // a sync /proc sweep must not run mid-update
   try {
-    const { reaped, observed } = reapMarkedOrphans({
+    const { reaped, observed, killed } = reapMarkedOrphans({
       ids,
       mode: config.reapRunaway,
       minCpu: config.reapRunawayMinCpu,
@@ -766,12 +783,22 @@ const sweepRunawayOrphans = (ids?: Set<string>) => {
         return s.status === "archived" ? "archived" : "live";
       },
     });
+    // Log per candidate from what ACTUALLY happened, never from `observed` alone: an armed sweep
+    // can still skip a kill (pid recycled between scan and kill, or the process already exited),
+    // and claiming "killed" for those would contradict the `reaped` total below.
+    const killedPids = new Set(killed.map((c) => c.pid));
     for (const c of observed) {
+      const verb =
+        config.reapRunaway !== "armed"
+          ? "would kill"
+          : killedPids.has(c.pid)
+            ? "killed"
+            : "skipped";
       const pct = Math.round(c.cpuFraction * 100);
       const mins = Math.round(c.ageSeconds / 60);
       console.warn(
-        `[reap-runaway] ${config.reapRunaway === "armed" ? "killed" : "would kill"} pid ${c.pid} ` +
-          `(${c.comm}) session=${c.sessionId} cpu=${pct}% age=${mins}m ppid=${c.ppid} cwd=${c.cwd}`,
+        `[reap-runaway] ${verb} pid ${c.pid} (${c.comm}) session=${c.sessionId} ` +
+          `cpu=${pct}% age=${mins}m ppid=${c.ppid} cwd=${c.cwd}`,
       );
     }
     if (reaped > 0) console.warn(`[reap-runaway] reaped ${reaped} runaway orphan(s)`);
--- a/src/process-reaper.ts
+++ b/src/process-reaper.ts
@@ -680,10 +680,25 @@ export interface ReapMarkedOptions {
  * PPID != 1), and requiring it would make this a no-op in a container where Shepherd is PID 1 and
  * every process reads as PPID-1. It is recorded on the candidate for the log only.
  *
- * LOAD-BEARING INVARIANT: no code path may hard-delete a `sessions` row without archiving it first.
- * Because absent ⇒ spare, a hard-deleted session's marked orphans become permanently unreapable.
- * This holds today — the only `DELETE FROM sessions` (store.ts, pruneArchivedSessions) is scoped to
- * `status = 'archived'`.
+ * LOAD-BEARING TIMING CONSTRAINT — an orphan must be SWEPT BEFORE ITS SESSION ROW IS PRUNED.
+ *
+ * `archived` is the only status that authorises a kill, and `absent` spares. So a marked orphan is
+ * reapable only during the window between its session being archived and its row being deleted.
+ * `pruneArchivedSessions` (store.ts) hard-deletes archived rows — the very rows this sweep depends
+ * on — so once it runs, that session's surviving orphans are unreapable FOREVER. (An earlier
+ * version of this note claimed the invariant was "never hard-delete a row without archiving it
+ * first", and cited the prune as satisfying it. That was backwards: the prune deletes only archived
+ * rows, so it satisfies that phrasing exactly while causing the failure the note meant to exclude.)
+ *
+ * What actually guarantees the ordering is CADENCE, not a rule about deletion:
+ *   - the sweep runs at teardown (SessionService.archive, synchronously after store.archive), then
+ *     hourly;
+ *   - the prune runs at most DAILY (runDailySweep) and only past a retention window
+ *     (SESSION_RETENTION_DAYS = 30, SESSION_RETENTION_KEEP = 250).
+ * So an orphan gets a teardown sweep plus ~24 hourly sweeps before its row is even eligible to be
+ * pruned. The margin is enormous, but it is a MARGIN, not a proof — anything that shortens
+ * retention toward the sweep interval, or lengthens the sweep interval toward retention, erodes it.
+ * Keep the sweep strictly more frequent than the prune.
  *
  * Fails closed throughout: any missing probe, an unreadable environ/stat, or a throwing
  * `sessionStatus` reaps nothing.
--- a/src/process-reaper.ts
+++ b/src/process-reaper.ts
@@ -767,10 +767,17 @@ function scanPid(pid: number, ctx: SweepContext): Scanned | null {
 
 export function reapMarkedOrphans(opts: ReapMarkedOptions): {
   reaped: number;
+  /** Everything that cleared every gate — in `observe` mode, what WOULD have been killed. */
   observed: RunawayCandidate[];
+  /**
+   * The subset actually SIGKILLed. Strictly ⊆ `observed`, and smaller whenever the pid-recycle
+   * guard fired or `killPid` threw — so callers must log from THIS, not from `observed`, or their
+   * per-candidate lines will claim kills that never happened and contradict `reaped`.
+   */
+  killed: RunawayCandidate[];
 } {
   const { sessionStatus, ids, mode, minCpu, minAgeS, probes = defaultProbes } = opts;
-  const none = { reaped: 0, observed: [] as RunawayCandidate[] };
+  const none = { reaped: 0, observed: [] as RunawayCandidate[], killed: [] as RunawayCandidate[] };
   if (mode === "off") return none;
 
   // Fail closed: without every probe the gates can't be established, so reap nothing.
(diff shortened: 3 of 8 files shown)


scoop (commented):
## Verdict

The PR satisfies the task. It closes both in-scope #1144 gaps, and it does so with a *stronger* attribution than the issue proposed (env-marker provenance ∧ session terminality, rather than cwd-under-a-shepherd-path + PPID-1). No blocking bug found. Findings below are nits.

## What I verified

**The marker actually reaches the process on both spawn paths.**
- Trusted: `spawnEnv` is passed to `herdr.start` (`src/service.ts:2116-2122`), and `buildWrappedArgv` turns it into `env K=V …` argv tokens *prefixed onto the inner argv* (`src/herdr.ts:286-300`), so the marker is set at exec regardless of what env the herdr server itself carries. The spread order puts `[SESSION_MARKER_ENV]` last, so `patchEnv` (a plugin) cannot shadow it.
- Membrane: `extraEnv` is emitted as `--setenv k v` pairs *after* `--clearenv` (`src/sandbox.ts:438-450`), so the marker survives the clear. `test/sandbox.test.ts:928-943` asserts exactly this.

**The gates.**
- Gap 2 (non-isolated): `archive()` calls `reapRunaway` unconditionally after `store.archive` (`src/service.ts:4626-4630`), while `worktree.remove()` — and thus the #1143 cwd sweep — is still `if (s.isolated)` only (`src/service.ts:4618`). That is the hole the issue described, and it is now covered. `test/service.test.ts:7531-7546` pins it.
- Gap 1 (chdir-out): `reapMarkedOrphans` never reads cwd as a gate (`src/process-reaper.ts:742` reads environ; `cwdForPid` is only used to populate the log field at `:778`), and `listPids()` enumerates `/proc` directly rather than going through `scanProcs()`, which drops pids whose cwd readlink fails (`src/process-reaper.ts:71-76, 219-227`).
- `absent ⇒ spare`: `sessionStatus` maps a missing row to `"absent"` and `scanPid` requires `"archived"` (`src/index.ts:763-767`, `src/process-reaper.ts:470`). The stated load-bearing invariant ("no path hard-deletes a `sessions` row without archiving it first") holds: the only `DELETE FROM sessions` is `pruneArchivedSessions`, whose victim predicate begins `status = 'archived'` (`src/store.ts:3180`, delete at `:3223`).
- A throwing store spares rather than reaps: `store.get` does not swallow SQLite errors (`src/store.ts:2020-2025`), and `scanPid` catches the throw and returns null (`src/process-reaper.ts:465-469`).
- `/proc/<pid>/stat` field math is right: after the last `)`, field N lands at index N-3, so utime(14)→11, stime(15)→12, cutime(16)→13, cstime(17)→14, starttime(22)→19 — matches `src/process-reaper.ts:250-257`.
- The `restore()` race the config comment claims is unreachable: `restore()` spawns the agent (`src/service.ts:3805`) *before* `store.unarchive` (`src/service.ts:3824`), so there really is a window in which fresh, marked processes belong to a still-`archived` row — but they are seconds old and the 300s age floor spares them. Sound *as long as the floor is enforced* (see finding 1).
- Every symbol the tests import (`reapMarkedOrphans`, `USER_HZ`, `CpuStat`, `ReapMarkedOptions`) is exported (`src/process-reaper.ts:486, 111, 97, 351`); `test/process-reaper.test.ts:2-12` agrees.
- `SHEPHERD_SESSION_ID` is spelled identically everywhere it appears (`src/process-reaper.ts:126` and the three test/doc sites); no second form exists in the tree.
- No i18n/message catalogs are touched, and no docs file documents `SHEPHERD_*` env vars today (`SHEPHERD_SESSION_HOUSEKEEPING` appears only in `src/config.ts`), so there is no parity obligation this PR misses.

**Security angle (no escalation).** The membrane now injects the session id into the sandbox, and the session id *is* the agent-ingress capability (`src/server.ts:6920-6926`). That is not a new exposure: the agent already knows its own id (it is baked into its `--settings` ingress URL), and a sandboxed agent is in its own PID namespace (`src/sandbox.ts:454` `--unshare-pid`, `:343` fresh `/proc`), so it cannot read other processes' environ. A trusted-path agent shares host `/proc` as the operator's uid and could already read every other agent's `/proc/*/cmdline`.

## Findings (all non-blocking)

1. **`src/config.ts:684-687` — the age/CPU floors are unvalidated env reads.** `Number(process.env.X ?? default)` returns `0` for an empty-string env var (`""` is not nullish, and `Number("") === 0`). With `SHEPHERD_REAP_RUNAWAY_MIN_AGE_S=` set to empty, the age floor becomes 0 — and that floor is the *only* thing making the `restore()` spawn-before-unarchive window (`src/service.ts:3805` vs `:3824`) unreachable, as the comment at `src/config.ts:685-686` itself says. Garbage input gives `NaN`, which fails closed, but empty-string fails *open*. `config.ts` already has `clampCap` (used at `:836`); clamping the age floor (and CPU fraction) would make the documented invariant enforced rather than commented.

2. **`src/service.ts:4630, 4635, 4666-4682` — the `bulkArchiving` boolean suppresses more than the batch.** `archive()` is async and `archiveMany` awaits it per id, with each `archive` able to sit for up to 15s in `beforeArchive` (`src/service.ts:4607-4617`). Any *other* `archive()` that interleaves during that window — and there are timer-driven callers: `src/automerge.ts:334`, `src/drain.ts:2135/2207`, `src/merge-teardown.ts:119/142`, plus the HTTP route `src/server.ts:2280` — sees `bulkArchiving === true`, skips its sweep, and is not in `cleared`, so the post-batch sweep at `:4682` does not cover it either. Its runaways then wait up to an hour for the safety net. Accumulating a pending-id set (and sweeping it once when the outermost call finishes) instead of a boolean would close this.

3. **`src/index.ts:769-777` — the log claims "killed" for candidates that were not killed.** `observed` includes every candidate, but in armed mode a kill can still be skipped by the pid-recycle guard or a throwing `killPid` (`src/process-reaper.ts:528-537`). The per-candidate line prints "killed" unconditionally when armed, so the detail lines can disagree with the `reaped N` total. Logging from the actually-reaped set (or reporting outcome per candidate) would keep the log truthful.

4. **`src/index.ts:752` / `src/process-reaper.ts:119-125` — the "satellite spawns are unmarked, hence always spared" invariant is not enforced against the server's own inherited env.** Nothing clears `SHEPHERD_SESSION_ID` from `process.env` at boot. When Shepherd is itself launched from inside an agent session (dogfooding `bun run dev` in a worktree — this repo's own workflow), the server process inherits that session's marker, and every spawn it makes that does *not* set the marker explicitly inherits it too: the satellite/aux paths pass only `apiKeyPassthroughEnv(false)` (`src/review.ts:386`, `src/distiller.ts:270`, `src/doc-agent.ts:793`, `src/plan-gate.ts:426`, …), so they cannot override it. Once that outer session is archived, a satellite's hot, non-listening child becomes a legitimate candidate — the opposite of the documented "deliberately NOT marked ⇒ always spared". A one-line `delete process.env[SESSION_MARKER_ENV]` at server boot would make the invariant true by construction. (Caveat: `herdr` is an external binary, so I could not follow the env through its server process end-to-end from this repo; the verifiable part is that nothing in the diff clears the marker from the server's own env and that the satellite spawns never set it.)

## Stated limitations

- `config.herdrBin` is an external binary not present in this repo, so the exact env-inheritance path from the Shepherd process → herdr server → agent pane could not be traced for the *inherited* (non-explicit) vars. The *explicit* marker is guaranteed regardless, because `buildWrappedArgv` prefixes it as `env K=V` argv tokens (`src/herdr.ts:294-300`).
- `USER_HZ = 100` is asserted as universal on Linux (`src/process-reaper.ts:107-111`); there is no `sysconf` binding in the tree to check it against at runtime. Not treated as a defect.

scoop (commented):
## Re-review of the four prior findings — all four are genuinely addressed

1. **Unvalidated `Number(env ?? default)` knobs (src/config.ts)** — FIXED. `reapRunawayMinCpu` now goes through a new `clampFraction` (src/config.ts:123-126, exported and re-used) with `[0.05, 1]`, and `reapRunawayMinAgeS` through the existing `clampCap` (src/config.ts:704-709) with a hard `[60, 86400]` floor, so a set-but-empty env var (`Number("") === 0`) snaps to 60s instead of disarming the age floor. `clampFraction` deliberately drops `clampCap`'s `Math.round` — I checked src/config.ts:114-120, `clampCap(0.8, 0.05, 1, 0.8)` would indeed return `1`, so a separate helper was necessary, not gratuitous. Covered by test/server-settings.test.ts:977-998.

2. **`bulkArchiving` boolean dropped concurrent archives (src/service.ts)** — FIXED. Replaced with a pending-id **set**: `archive` defers into `this.pendingRunawayIds` when non-null (src/service.ts:4633-4634), the field is declared at src/service.ts:4643, and `archiveMany` opens/closes the collector with an `outer` re-entrancy guard and sweeps `pending` (not `cleared`) at the end (src/service.ts:4675-4698). I traced the interleaving: `archive` adds to the set and `archiveMany`'s post-`finally` sweep both run synchronously with no `await` between them, so no id can be added after the collector closes but before the sweep — and once it is null, a concurrent `archive` sweeps itself directly. All `archive` callers route through `SessionService.archive` (automerge.ts:334, drain.ts:2135/2207, merge-teardown.ts:119/142, server.ts:2280, service.ts:3150), so no archived id escapes the sweep.

3. **Log lines claiming "killed" for skipped kills (src/index.ts)** — FIXED. `reapMarkedOrphans` now returns a `killed` array (src/process-reaper.ts:770-780, populated only after `killPid` returns at src/process-reaper.ts:846-848), and src/index.ts:783-800 builds `killedPids` from it, printing `killed` / `skipped` / `would kill` accordingly. `reaped` is `killed.length`, so the detail lines can no longer contradict the total.

4. **Stale inherited `SHEPHERD_SESSION_ID` (src/index.ts)** — FIXED. `delete process.env[SESSION_MARKER_ENV]` at src/index.ts:191, before any spawn. I checked that every whole-env spread in the tree is evaluated at call time, not at import time (preview-launch.ts:149, pty-bridge.ts:36, socket-pty-bridge.ts:121, usage-probe.ts:123, pull.ts:17, repos.ts:189, repo-roles.ts:200 — all inside functions), so the boot delete really does reach them. The aux spawns still pass no marker (distiller.ts:270, doc-agent.ts:793), so the "unmarked ⇒ always spared" invariant now holds.

## Verification of the new machinery

- **Marker name has no prior meaning.** `SESSION_MARKER_ENV = "SHEPHERD_SESSION_ID"` (src/process-reaper.ts:126); a tree-wide grep finds no other producer or consumer of that variable, so no collision.
- **Marker actually reaches the agent on both spawn paths.** Trusted: `spawnEnv` at src/service.ts:2114-2121 → `herdr.start` (src/service.ts:2122) → `buildWrappedArgv` renders `K=V` tokens into an `env` shim (src/herdr.ts:286-301, used at src/herdr.ts:554), so it lands in the agent's `/proc/<pid>/environ`. Sandboxed: `membrane.extraEnv` at src/service.ts:2085-2088 → the `--setenv` loop at src/sandbox.ts:446-450, which runs *after* `--clearenv` at src/sandbox.ts:438 — so it survives the membrane. `ctx.sessionId` is the sessions-table id (same value used for the hooks URL `/api/sessions/<id>/hooks`, src/service.ts:355-368), which is what the terminality gate requires.
- **Terminality gate.** `store.get` has no status filter (src/store.ts:2020-2025) so archived rows resolve, and `status='archived'` is exactly what `store.archive` writes (src/store.ts:2242). `archive()` awaits `herdr.stop` (src/service.ts:4601) before `store.archive` (src/service.ts:4626), so the agent is dead before its session can ever read `archived`.
- **`/proc/<pid>/stat` parsing.** `readCpuStat` splits on the *last* `)` and indexes 11/12/13/14/19 for utime/stime/cutime/cstime/starttime. Field N maps to offset N-3 after the comm field, so 14→11, 15→12, 16→13, 17→14, 22→19: correct. Non-finite ⇒ null ⇒ spared.
- **Probe binding.** `listPids`/`cpuStatForPid` are destructured and called unbound in `reapMarkedOrphans` (src/process-reaper.ts:786-787, 841); the `defaultProbes` implementations (src/process-reaper.ts:389-412) use no `this`, so the unbound calls are safe.
- **Listening gate** uses the listening-inode map only (src/process-reaper.ts:370-377 → `listeningInodeToPort`), so an agent-started dev server is spared while a busy-loop (which listens on nothing) is not — as documented.
- **Age floor vs. the restore() race.** With the clamp, `minAgeS >= 60`, so a freshly respawned agent's children (age ≈ 0) can never clear the floor during restore's spawn-before-unarchive window; the agent itself is additionally spared by `AGENT_COMMS`.
- **Task satisfaction.** Both gaps the issue names are closed: attribution by environ marker is cwd-blind (gap 1), and the sweep is called from `archive()` unconditionally rather than hanging off `worktree.remove()`'s `if (s.isolated)` (gap 2, src/service.ts:4617 vs 4633).

### Observations (no change requested)

- The CPU gate is a *lifetime average* (`cpuSeconds / ageSeconds`), so a process that idles for hours and only then starts spinning never clears 0.8 and is never reaped. The docstring at src/process-reaper.ts:672-678 states this trade-off explicitly and the issue left the window/threshold to the implementer, so I'm not treating it as a defect — but if late-onset spins matter, a per-sweep delta (cache `utime+stime+cutime+cstime` keyed by pid+starttime) would be cheap to add on top of the existing hourly cadence.
- `normalizeReapRunaway` (src/config.ts:427-432) fails *open*: `SHEPHERD_REAP_RUNAWAY=false` or `=no` yields `armed`, not `off`. That matches the repo's existing `!== "0"` convention (e.g. src/config.ts:681) and is documented, so I'm leaving it alone.
- Could not verify how the external `herdr` daemon composes its own environment with the `env K=V` shim tokens — that binary is not in this repo. If the daemon itself were ever launched from inside an agent session it would carry a stale marker that this PR's boot scrub (which only touches Shepherd's own `process.env`) cannot reach, and its *unmarked* aux spawns would inherit it. Assumed as written; not a finding.

- none.

scoop (commented):
## Verdict

The PR implements #1144's "resource-gated orphan reaper" and closes both in-scope gaps. I found no correctness, safety, or security defect that blocks. One non-blocking quality nit (boot sweep runs synchronously where both of its siblings deliberately defer).

## Prior finding — addressed

1. **`src/process-reaper.ts` "LOAD-BEARING INVARIANT" stated the wrong constraint.** Fixed. The note is now `LOAD-BEARING TIMING CONSTRAINT — an orphan must be SWEPT BEFORE ITS SESSION ROW IS PRUNED` (src/process-reaper.ts:683-703). It explicitly says `pruneArchivedSessions` hard-deletes exactly the archived rows the sweep depends on, retracts the old backwards phrasing, and grounds the guarantee in cadence. I checked each cited fact: the prune is a hard delete of archived rows (src/store.ts:3174), it is driven daily (src/index.ts:2222 inside the sweep scheduled by src/index.ts:2295), and the retention window is 30 days / newest 250 (src/config.ts:899-901). The sweep cadence claimed by the note is real: teardown at src/service.ts:4626-4634 (immediately after `store.archive`) plus hourly at src/index.ts:812. The restated note matches the code.

## Verification performed

**Identifier resolution — every symbol the diff introduces or relies on.**
- `SESSION_MARKER_ENV` is declared once (src/process-reaper.ts:126, value `"SHEPHERD_SESSION_ID"`) and imported by both consumers (src/index.ts:107, src/service.ts:52). No other spelling of the env name exists in `src/` — grep for `SESSION_ID|SHEPHERD_SESSION` returns only these, the doc comments, and the unrelated `SHEPHERD_SESSION_HOUSEKEEPING` (src/config.ts:681). No collision with an existing agent-facing var.
- `reapMarkedOrphans`, `USER_HZ`, `CpuStat`, `SessionTerminality`, `RunawayCandidate`, `ReapMarkedOptions` are all exported from src/process-reaper.ts and consumed consistently; `clampFraction` is exported (src/config.ts:123) and used at src/config.ts:693.
- `probes.portsForPid` used in the non-batched `isListening` fallback is a **required** member of `ReaperProbes` (src/process-reaper.ts:41), so the fallback can never be undefined. The five probes the sweep actually needs (`listPids`, `commForPid`, `environForPid`, `cpuStatForPid`, `uptimeSeconds`) are all fail-closed-checked before use, and all five exist in `defaultProbes` (src/process-reaper.ts:389-410). `cwdForPid`/`ppidForPid` are optional-chained and only feed the log line.

**The marker actually reaches `/proc/<pid>/environ` on both spawn paths.**
- Trusted path: `spawnEnv` (src/service.ts:2116-2121) is handed to `herdr.start` (src/service.ts:2122), which routes it through `buildWrappedArgv` (src/herdr.ts:286-300). I read that function: every `env` entry is emitted as an `env K=V` token with no filtering, so `SHEPHERD_SESSION_ID` lands in the exec'd agent's environ.
- Membrane path: `bwrap --clearenv` would strip the outer value, so the marker is also placed on `membrane.extraEnv` (src/service.ts:2087), and src/sandbox.ts:446-450 emits every `extraEnv` entry as `--setenv k v`. Asserted in test/sandbox.test.ts:929.
- `prepareSpawn` (src/service.ts:2008) is the *only* agent spawn path — src/service.ts has exactly one `herdr.start` call (line 2122) — so create and restore both stamp the marker. Placement is last in the object spread on both paths, so `patchEnv` cannot shadow it.

**The boot scrub is necessary and correctly placed.** `delete process.env[SESSION_MARKER_ENV]` at src/index.ts:191 runs at module top level before anything spawns. This is what keeps the aux/satellite agents unmarked: they spawn via `resolveAuxSpawn` (src/spawn-membrane.ts:200, called from review.ts:360, plan-gate.ts:405, doc-agent.ts:775, standalone-critic.ts:424), which never sets the marker, so they inherit the server's env — scrubbed. Unmarked ⇒ spared (src/process-reaper.ts:757-758). The docstring's claim holds.

**`/proc/<pid>/stat` parsing.** `readCpuStat` splits on the *last* `)` and indexes 11/12/13/14/19. Verified against the kernel field numbering: after the comm field, field *N* sits at offset *N−3*, so utime(14)→11, stime(15)→12, cutime(16)→13, cstime(17)→14, starttime(22)→19. Correct, and `Object.values(stat).every(Number.isFinite)` rejects a short/garbled line.

**Gate chain.** `ageSeconds = uptime − starttime/USER_HZ`; both comparisons use `!(x >= y)` so NaN rejects. `ageSeconds` can never be ≤0 at the division because `minAgeS` is clamped to ≥60 (src/config.ts:704-709), so `cpuFraction` has no divide-by-zero. The pid-recycle guard re-reads `stat` and compares `starttime` before the kill, and `killed` (not `observed`) drives both `reaped` and the per-candidate log verb in src/index.ts:789-796 — those agree.

**Terminality.** `store.get` (src/store.ts:2020-2025) is an unfiltered `SELECT … WHERE id = ?`, so it does return archived rows; the `sessionStatus` closure (src/index.ts:780-784) maps present+archived→`archived`, present+anything-else→`live` (statuses per src/types.ts:6), missing→`absent`, and lets a DB throw propagate into `scanPid`'s `catch` → spare. The "absent ⇒ spared" rule is what keeps a second Shepherd instance from killing the first's live work; that is honoured.

**Restore race.** `restore` re-enters the same `prepareSpawn`, so the respawned agent briefly runs while its row still reads `archived`. The agent itself is spared by comm (`AGENT_COMMS`), and any child it spawns in that window is <60s old, below the hard-clamped age floor. The clamp (not just the default) is what closes this, and test/server-settings.test.ts:978 pins the `Number("")===0` trap.

**Task satisfaction.** Attribution by environ marker rather than cwd does cover the two residuals the issue names: a chdir-out orphan keeps its environ (fixed at exec), and a non-isolated session's orphan is attributable without ever cwd-sweeping the shared repo root. `archiveMany`'s defer-then-sweep-once collector (src/service.ts:4669-4699) correctly reuses an already-open collector for nesting and re-checks `!outer` before both closing and sweeping; a concurrent archive that interleaves after the outer batch closes falls back to the immediate per-id call, so no id is dropped.

## Non-blocking observation (not a finding)

The teardown sweep does a full host-wide `/proc` enumeration per un-batched `archive()`, yet the 300s age floor means it will usually reap nothing for the session just archived (its orphan is typically younger). That cost/benefit is acknowledged in the code comments and the hourly sweep is the real net; flagging only so it is a conscious trade.

## Stated limitations

- I did not execute anything (read-only review); CI covers tests and lint.
- The exact behaviour of the external `herdr` binary (whether it re-executes the `env` shim verbatim into the PTY) is not in this repo. I verified the argv construction in src/herdr.ts:286-300 and assumed `herdr` exec's the argv as given, which is the same assumption every existing env passthrough (e.g. `CLAUDE_CONFIG_DIR`) already relies on.

scoop (commented):
## Re-review of the previous revision's point

**Prior finding 1 (src/index.ts: boot sweep ran synchronously inside `deferredStarts`) — ADDRESSED.** The boot invocation is now `void Promise.resolve().then(() => sweepRunawayOrphans()).catch(...)` at `src/index.ts:816-818`, byte-for-byte the deferral shape used by the sibling #1133 deleted-worktree reap at `src/index.ts:743-749` (`void Promise.resolve().then(() => { const { reaped } = reapDeletedWorktreeOrphans(); ... }).catch(...)`), with a comment naming that parity. The hourly `setInterval` (src/index.ts:819) is registered inside the same `deferredStarts` push, so nothing about the sweep sits inline on the boot path any more.

## What I verified

- **Marker survives the membrane.** `src/service.ts:2083-2088` puts `[SESSION_MARKER_ENV]: ctx.sessionId` last in `membrane.extraEnv`; `src/sandbox.ts:438` pushes `--clearenv` and `src/sandbox.ts:446-450` then emits `--setenv k v` for every `inputs.extraEnv` entry, so the marker is re-set inside the sandbox after the clear. The new `test/sandbox.test.ts:929-942` asserts exactly that argv shape.
- **Both spawn paths are stamped, and there is only one spawn path.** `prepareSpawn` (src/service.ts:2009) is the sole builder of `spawnEnv` (src/service.ts:2116-2122) and of `membrane.extraEnv`, and it is reached from create (src/service.ts:2914), resume (src/service.ts:3223) and `prepareResumeSpawn` (src/service.ts:2582) — so create/resume/restore all carry the marker.
- **`SESSION_MARKER_ENV` identifier is consistent.** Declared once as `"SHEPHERD_SESSION_ID"` at `src/process-reaper.ts:126`; imported by `src/index.ts:86-91` and `src/service.ts:52`; the only other occurrences of the raw string are in the tests (test/process-reaper.test.ts:541,825; test/sandbox.test.ts:929-939). No other consumer of `SHEPHERD_SESSION_ID` exists in `src/`, so the boot scrub (`delete process.env[SESSION_MARKER_ENV]`, src/index.ts:109) cannot break an existing reader; aux spawns go through `resolveAuxSpawn` → `collectPassthroughEnv()` (src/spawn-membrane.ts:60), which reads the (now-scrubbed) server env, so they stay unmarked as the doc-comment at src/process-reaper.ts:117-125 claims.
- **The terminality gate can actually see archived rows.** `store.get` is an unfiltered `SELECT ... WHERE id = ?` (src/store.ts:2020-2025) and `store.archive` sets `status='archived'` (src/store.ts:2240-2247), so `sessionStatus` (src/index.ts:780-784) resolves an archived session to `"archived"` rather than `"absent"` — the sweep is not a silent no-op.
- **The restore race the comments lean on is real but bounded.** `restore` spawns the agent (src/service.ts:3805) *before* `store.unarchive` (src/service.ts:3824) and before `finishResumeSpawn` sets `status: "running"` (src/service.ts:2604-2608), so a restored session's row briefly still reads `archived`. The `reapRunawayMinAgeS` clamp with a hard 60s floor (src/config.ts:68-73) makes that window unreachable, since every process of the freshly respawned agent is seconds old. The clamp is what enforces it, as the comment says — verified `clampCap` floors to `min` for `Number("") === 0` (src/config.ts:114-119).
- **`/proc/<pid>/stat` parsing.** `readCpuStat` splits on the last `)` and indexes 11/12/13/14/19; with the remainder starting at field 3 (`state`), field N lands at index N-3, i.e. utime 14→11, stime 15→12, cutime 16→13, cstime 17→14, starttime 22→19. Correct. `readdirSync`/`readlinkSync`/`readFileSync` are all already imported (src/process-reaper.ts:1).
- **Probe contract.** Every probe the sweep needs is declared optional on `ReaperProbes` (src/process-reaper.ts:210-227) and `reapMarkedOrphans` fails closed when any is absent; the `isListening` fallback calls `probes.portsForPid`, which is a *required* member (src/process-reaper.ts:41), so the non-batched path can't hit an undefined call. `cpuFraction` divides by `ageSeconds`, which is already gated `>= minAgeS >= 60`, so no divide-by-zero.
- **Task satisfaction.** Attribution by environ marker rather than cwd closes both gaps the issue names: a chdir-out orphan (environ is fixed at exec) and a non-isolated session's orphan (the new `reapRunaway` hook fires in `archive()` regardless of `s.isolated`, src/service.ts:4627-4632 — the only teardown reap that path gets, since `reapOrphansUnder` hangs off `worktree.remove()`).

## Nit

See findings — the `archiveMany` batch sweep sits outside the `try/finally`.

## Stated limitations

- I did not execute anything (read-only review), so the /proc reads, the bwrap argv and the SIGKILL path are verified by code inspection and the added unit tests, not by running them.

scoop (commented):
## Re-review of the prior revision's finding

**Prior finding 1 (src/service.ts `archiveMany` batch sweep after the `try/finally`) — ADDRESSED.** The sweep now sits *inside* the `finally` (`src/service.ts:4682-4694`), guarded by `if (!outer)`, with the collector nulled first and the `reapRunaway` call wrapped in its own `try/catch` so it can never mask the original exception. The throwing path is covered by a regression test that stubs `store.get` to throw on the second id and asserts the first id is still swept (`test/service.test.ts`, "#1144: a mid-batch throw still sweeps the ids already archived in that batch"). Confirmed against `src/service.ts:4675-4699`.

- **Marker identifier is spelled consistently everywhere.** `SESSION_MARKER_ENV = "SHEPHERD_SESSION_ID"` is declared once at `src/process-reaper.ts:126` and every use is the imported constant (`src/service.ts:2087`, `src/service.ts:2120`, `src/index.ts:109`, read back at `src/process-reaper.ts:761`). A tree-wide grep for the literal `SHEPHERD_SESSION_ID` finds only that declaration, the two test files, and comments — no second, divergent form, and no pre-existing use of the name to collide with.
- **The marker actually reaches the agent's `/proc/<pid>/environ` on BOTH spawn paths.** Trusted: `spawnEnv` (`src/service.ts:2116-2121`) → `herdr.start` → `buildWrappedArgv` (`src/herdr.ts:286-300`), which emits each entry as a `KEY=VALUE` token to the coreutils `env` shim that `execvp`s straight into `claude`, so the var is in the exec'd environ. Membrane: `membrane.extraEnv` (`src/service.ts:2081-2088`) → the `--setenv` loop at `src/sandbox.ts:446-449`, which iterates `extraEnv` *after* `--clearenv` (`src/sandbox.ts:438`) with no key allowlist filtering. Asserted directly in `test/sandbox.test.ts:929-941`.
- **Every main-agent spawn goes through `prepareSpawn`.** create → `prepareSpawnOrThrow` (`src/service.ts:2914`), resume/heal → `prepareResumeSpawn` (`src/service.ts:2581`, called from `src/service.ts:3674` and `src/service.ts:3805`), drain → `src/service.ts:3223`. So no spawn path silently skips the marker.
- **The `restore()` race the age clamp exists for is real, and the clamp closes it.** `prepareResumeSpawn` (`src/service.ts:3805`) spawns the agent *before* `store.unarchive` (`src/service.ts:3824`), so the row genuinely reads `archived` while a live agent is running. The freshly-spawned agent's children are seconds old, and `reapRunawayMinAgeS` is `clampCap(..., 60, 86400, 300)` (`src/config.ts:704-709`) — a hard 60s floor even for a set-but-empty env var. Verified.
- **`status === "archived"` is the right terminality string** — same literal the service itself gates on at `src/service.ts:3789`.
- **`/proc/<pid>/stat` parsing is correct.** `readCpuStat` splits after the *last* `)` (`src/process-reaper.ts:314`), so the remainder starts at field 3; field N lands at index N-3. utime(14)→11, stime(15)→12, cutime(16)→13, cstime(17)→14, starttime(22)→19 — exactly the offsets used at `src/process-reaper.ts:321-327`. Non-finite fields fail closed to `null`.
- **Fail-closed posture holds.** Missing probe / unreadable uptime → `none` (`src/process-reaper.ts:806-809`); unreadable environ or unmarked → spared (`src/process-reaper.ts:762`); throwing `sessionStatus` → spared (`src/process-reaper.ts:766-770`); `absent` → spared (`src/process-reaper.ts:772`), which is what protects a second Shepherd instance's live sessions. Self-pid excluded (`src/process-reaper.ts:735`).
- **The boot marker scrub is necessary and sufficient.** `delete process.env[SESSION_MARKER_ENV]` (`src/index.ts:109`) cannot change the server's own already-exec'd `/proc/self/environ`, but the sweep skips `process.pid`, and every process the server spawns after boot (aux agents via `resolveAuxSpawn`, git, herdr) inherits the scrubbed env — so the "server restarted from inside an agent session" case does not mark satellites.
- **No TDZ hazard** from `reapRunaway: (ids) => sweepRunawayOrphans(ids)` at `src/index.ts:646` referencing the `const` at `src/index.ts:769`: nothing between those points archives a session synchronously; the closure is only invoked at runtime.
- **Both issue gaps are closed.** Non-isolated sessions now get a teardown reap (attribution is by env marker, not cwd, so `worktree.remove()`'s `if (s.isolated)` gate is bypassed — asserted in `test/service.test.ts`, "archive() reaps runaways for a NON-isolated session"), and chdir-out orphans are caught because `/proc/<pid>/environ` is fixed at exec. The task is satisfied.
- **Config knobs typecheck against the reaper contract**: `config.reapRunaway` is the union `"armed"|"observe"|"off"` (`src/config.ts:432-437`), matching `ReapMarkedOptions["mode"]` (`src/process-reaper.ts:644`); `clampFraction` is exported from `src/config.ts:11` and imported in `test/server-settings.test.ts`.

No env-var documentation catalog exists to keep in parity — a grep for the sibling knobs `SHEPHERD_SESSION_HOUSEKEEPING` / `SHEPHERD_LLM_NAMING` finds them only in `src/config.ts`, so the new `SHEPHERD_REAP_RUNAWAY*` vars being config-only is consistent with the repo's convention. There are no user-facing strings or message catalogs in this diff.

## Non-blocking observations (see findings)

1. `killPid` signals the matched pid only (`src/process-reaper.ts:837`), not its process group. That leaves a residual on exactly the case the `cutime`/`cstime` columns were added for (`src/process-reaper.ts:744-747`): the `while true; do work; done &` supervisor is killed, but the child it is currently blocked in `wait()` on keeps burning. It is itself marked and hot, so a later sweep reaps it — but only once it clears the 300s age floor, i.e. up to an hour of one core. Bounded and far better than the status quo, but worth an explicit note or a group-kill.
2. The "one sweep per batch" invariant (`src/service.ts:4671-4673`) holds for nesting but degrades under two *concurrent* `archiveMany` calls: B adopts A's collector as `outer`, so when A's `finally` nulls `pendingRunawayIds`, B's remaining per-id `archive()` calls each take the `else` branch at `src/service.ts:4634` and run a full host-wide `/proc` enumeration apiece. No id is lost (correctness is fine), but the cost guard is defeated on that interleaving.

- src/index.ts: `sweepRunawayOrphans` returns early on `maintenance.active` (`src/index.ts:770`), silently dropping that batch's teardown sweep to the hourly net with no log line — invisible today, and misleading if the maintenance window ever lengthens.

## Stated limitations
- `USER_HZ = 100` (`src/process-reaper.ts:113`) is hardcoded because neither Node nor Bun exposes `sysconf(_SC_CLK_TCK)`; I could not verify it against the host kernel from this repo. It is correct on every mainstream Linux build, and the code's own comment says so. Not a finding.

Issue #1680, opened 2026-07-12
Guard ChatGPT-account-incompatible Codex models on the main-task / drain spawn path


kai-osthoff:
## Context

Follow-up to the role-spawn auth guard (see the recap-failure fix PR). That PR added an
**affirmative Codex auth-mode detection** (`src/codex-auth.ts`) and a clamp
(`clampCodexModelForAuth` in `src/default-model.ts`) so that **role spawns** — recap, namer,
autopilot, critic, doc-agent, resolved through the shared `resolveRoleEnvironment` /
`resolveRoleEnvWithAuth` — never launch with a Codex model that a ChatGPT-account login rejects
with HTTP 400 (`"… not supported when using Codex with a ChatGPT account."`). It also surfaces a
`codex_model_auth` diagnostics warning and captures the codex pane tail on recap failure.

**The main-task / drain Codex spawn path is deliberately NOT covered by that PR** and should be
guarded here.

## Scope

The main/foreground + drain Codex spawn resolves its model through a different path than the role
resolver. Apply the same `clampCodexModelForAuth` (reuse `readCodexAuthMode` + the blocklist) at:

- `src/service.ts` `buildCodexSpawnArgv` (~2427) and `buildCodexResumeArgv` (~2468) — where a codex
  main session emits `--model` / `-m`.
- Model validation/resolution at create / relaunch (`reconcileRelaunchModel`, service.ts:1426) /
  epic-run (`normalizeEpicRunModel`, server.ts:5927) / drain (`src/drain.ts`).

Note the existing **hard-error** semantics on some of these paths (an explicit incompatible model
throws). Decide per-site whether to clamp-to-account-default (consistent with the role path) or keep
the hard error with a clearer message — a main/foreground failure is at least loud (the operator
sees the 400 live), unlike the silent role-spawn failures the first PR fixed.

## Why deferred

Main-session failures are visible to the operator immediately; the silent background role failures
were the reported bug. Splitting keeps the first PR cohesive and low-risk (Shepherd single-PR rule).

kai-osthoff:
⏸️ Waiting on review of PR #1716 by @scoop.

kai-osthoff:
✅ PR #1716 merged.

PR #1681, opened 2026-07-12
fix(recap): guard ChatGPT-account-incompatible Codex role models

kai-osthoff:
## Why "Recap generation failed."

The recap role was pinned to Codex **`gpt-5.3-codex`**, which the operator's **ChatGPT-account**
Codex login rejects with `HTTP 400` ("… not supported when using Codex with a ChatGPT account.").
The spawn exits after ~74 s (well under the 300 s timeout) writing no `.shepherd-recap.json`, so
`RecapService.finalize()` fails closed → the UI's "Recap generation failed."

Root-caused from the live DB + logs: **`gpt-5.3-codex` → 133 `failed` / 0 `ready`** vs **`sonnet`
→ 39 `ready`**; reproduced directly with `codex exec -m gpt-5.3-codex …`. The gap: the shared role
resolver (`resolveRoleEnvironment`) only clamps model↔**provider**, never the **auth mode**, and no
code read `~/.codex/auth.json`.

## The fix (role-spawn path)

Covers every background role resolved through the shared resolver — **recap, namer, autopilot,
critic, doc-agent** — in three complementary moves:

1. **Preventive** — `src/codex-auth.ts` detects the Codex auth mode **structurally** (a present
   `tokens.access_token` + no `OPENAI_API_KEY` ⇒ `chatgpt`, even when `auth_mode` is absent;
   fail-open on any error). `clampCodexModelForAuth` (`src/default-model.ts`) drops a
   blocklisted-under-chatgpt Codex model to the account default (`model → null`, i.e. no `--model`).
   `resolveRoleEnvironment` gains a `codexAuthMode` param (clamps both the inherit/global and
   explicit-role branches); `resolveRoleEnvWithAuth` is the testable reader→resolver seam that
   `roleEnv` (index.ts) delegates to, reading the live auth mode per spawn.
2. **Diagnostic** — on a finalize-null recap, `src/recap.ts` now captures the codex pane tail via
   `herdr.readAsync` and logs it, so a **not-yet-blocklisted** incompatible model is diagnosable
   instead of a silent black hole (the very gap that made this bug hard to find).
3. **Visible** — a `codex_model_auth` diagnostics **warning** (`src/diagnostics.ts`) across the live
   role/global model config under chatgpt auth, surfacing the override to the operator.

## Scope & honesty

- Clamping to the account default **cannot guarantee** a `ready` recap end-to-end — Shepherd does
  not read `~/.codex/config.toml`, so if that default is itself incompatible the recap still fails
  (now diagnosably, via move 2). The blocklist is a pragmatic first cut for confirmed-bad models;
  fail-open never wrongly blocks an unknown one.
- **Main-task / drain Codex path** is deliberately **out of scope** (its failures are loud — the
  operator sees the 400 live) and tracked in #1680.

## Tests & verification

- New: `test/codex-auth.test.ts` (incl. the no-`auth_mode` chatgpt case); real reader→resolver seam
  test + clamp tests in `test/default-model.test.ts`; pane-tail capture in `test/recap.test.ts`;
  `codex_model_auth` warn/omit in `test/diagnostics.test.ts`.
- `tsc --noEmit` clean · `bun run lint` clean · `check:i18n` parity (en+de) · branch-hygiene /
  feature-catalog / glossary / announcement-version gates green.
- Full root suite: 6817 pass / 1 fail — the single failure (`pty-bridge`) reproduces on clean
  `origin/main` and is environment-specific (unrelated to this change).

Closes the recap failure reported for TASK-641.

### DIFF
--- a/src/codex-auth.ts
+++ b/src/codex-auth.ts
@@ -1,0 +1,55 @@
+/**
+ * Codex CLI auth-mode detection. Some Codex models return HTTP 400
+ * ("… not supported when using Codex with a ChatGPT account") under a ChatGPT-account login but
+ * work with an API key, so role/main spawn resolution must know the auth mode to avoid pinning a
+ * doomed model (see `clampCodexModelForAuth` in ./default-model).
+ *
+ * Detection is STRUCTURAL, not by any single field: a real `~/.codex/auth.json` may omit an
+ * explicit `auth_mode`, so relying on it alone would silently mis-detect a ChatGPT operator as
+ * `unknown` and never clamp. Precedence: a real API key wins; else a present OAuth access token
+ * means a ChatGPT-account login; else an explicit `auth_mode` corroborates; else `unknown`.
+ */
+import { readFileSync } from "node:fs";
+import { join } from "node:path";
+import { codexHome } from "./codex-usage";
+import type { CodexAuthMode } from "./default-model";
+
+function nonEmptyString(v: unknown): v is string {
+  return typeof v === "string" && v.trim() !== "";
+}
+
+/**
+ * Classify a parsed `auth.json` object into a {@link CodexAuthMode}. Pure — no I/O.
+ * Order matters: an API key is authoritative (it overrides a stale tokens block); a present
+ * `tokens.access_token` is the affirmative ChatGPT-account signal; `auth_mode` is only a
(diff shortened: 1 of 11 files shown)


kai-osthoff:
🎉✨ Behold, a READY PR 🤯🚀 @scoop stop napping and merge it 😴➡️✅🔥💅

Issue #1683, opened 2026-07-12
Diagnostics: surface & one-click-fix untrusted claude folder-trust (repoRoot + worktrees)


scoop:
Follow-up split off from the #1075 probe fix (PR fixes the `/usage` probe by pre-seeding folder trust before spawning claude). This issue tracks the broader, user-facing surfacing of the same root cause.

## Investigate FIRST (determines the design)
- Do **unattended autopilot sessions** spawning in **fresh (untrusted) worktree paths** (`service.ts:2122`, and the doc-agent / optimizer / review spawns) wedge on the **same** "Do you trust the files in this folder?" dialog? If so, that's a larger reliability latent than the probe, and it decides whether the fix must seed trust for **worktree paths at spawn** rather than a `config.repoRoot`-only diagnostics check (which would NOT cover worktree spawns).

## Scope (once the above is settled)
- New `claude_trust` **Diagnose check** reporting whether claude trusts the relevant path(s), reusing `trustRepoRoot`/`readRepoRootTrusted` from `src/claude-trust.ts` (shipped in the #1075 PR).
- **One-click code fix** (seed `hasTrustDialogAccepted`) — the fix needs a dynamic path, which payload-purity bans from the snapshot, so it's a server-side code action, not a static shell command. Add an optional path-free `fixActionKey` (message key) to `DiagnosticCheck`; dispatch the code fix by `hintKey` in `DiagnosticsService.fix()`.
- **`DiagnoseRows.svelte`**: render the translated `fixActionKey` sentence (not an empty `<code>` block) when `remediation` is absent, with a component test covering that branch.
- **Auth-mode gate**: surface the check **only under subscription auth** — in api-key mode the probe short-circuits and an untrusted path wedges nothing, so the warning + fix button would be spurious noise.
- Keep the fix read-gated (no write when already trusted); same whole-file clobber caveat as the probe seed.
- Config-dir-aware target (`$HOME/.claude.json` vs `${CLAUDE_CONFIG_DIR}/.claude.json`) — already proven in the #1075 PR.
- i18n keys (en+de), `v<next>-claude-folder-trust` feature entry, gate on claude presence.

Context: #1075.

scoop:
✅ PR #1739 merged.

PR #1684, opened 2026-07-12
fix(usage-probe): pre-seed claude folder-trust to stop /usage wedge (#1075)

### LINKED ISSUE #1075: Usage probe wedges on claude's folder-trust prompt when repoRoot is untrusted (scrape returns null)
## Problem

The `/usage` probe (`HerdrUsageProbe.scrape`) spawns `claude --dangerously-skip-permissions` in `config.repoRoot`. If that folder is **not yet trusted** in `~/.claude.json` (`projects[<repoRoot>].hasTrustDialogAccepted: false`), claude boots into the **"Do you trust the files in this folder?"** dialog. `--dangerously-skip-permissions` does **not** suppress this trust gate.

The probe then types `/usage` + Enter on its fixed boot timer — but those keystrokes are consumed by the trust dialog (Enter confirms "Yes, I trust this folder"), so the `/usage` command is never actually submitted. The scrape returns `null` for the **entire** panel (week + 5h + credits), not just credits.

## Evidence

Reproduced the herdr-attach probe path against `/home/patrick` (the live `repoRoot`) while `hasTrustDialogAccepted` was `false`: the captured buffer ended at `…1. Yes, I trust this folder  2. No, exit … ✔` followed by the empty main prompt — `week=false`, `credits=false`, scrape → null. Setting `hasTrustDialogAccepted: true` for the folder made the probe reach the panel normally.

## Impact

- On a freshly-onboarded repo / first run in a new `repoRoot`, usage calibration silently no-ops until the folder happens to get trusted by some other claude invocation. Symptom: all usage meters stuck stale, refresh returns 200 (or the route fails closed) with no useful error.
- This is a **reliability latent**, distinct from the credits parse-anchor bug fixed in #1072 (where the folder was already trusted and only credits parsing failed).

## Suggested hardening (one or more)

- Detect the trust dialog in the scrape buffer (`"trust this folder"` / `"Yes, I trust"`) and either dismiss it before sending `/usage`, or pre-seed `hasTrustDialogAccepted: true` for `repoRoot` in the probe's config dir.
- Or pass a flag/env that pre-trusts the working dir for the ephemeral probe spawn.
- Add a regression test around the trust-dialog-in-buffer case so the probe driver doesn't silently return null.

Found while diagnosing the extra-credit gauge staleness (root cause of *that* report is #1072).

scoop:
Fixes #1075.

## Problem
The `/usage` probe spawns `claude --dangerously-skip-permissions` in `config.repoRoot`. When that folder is **not yet trusted** (`~/.claude.json` → `projects[<dir>].hasTrustDialogAccepted !== true`), claude boots into the **"Do you trust the files in this folder?"** dialog — which `--dangerously-skip-permissions` does **not** suppress. The probe's `/usage`+Enter keystrokes land in that dialog instead of the REPL, so `scrape()` returns `null` and every usage meter (week + 5h + credits) stays stale until the folder happens to get trusted by some other claude invocation.

## Fix
Before spawning claude, `scrape()` reads whether its cwd is trusted and, if not, seeds `hasTrustDialogAccepted: true` **once** — so claude boots straight to the REPL and `/usage` lands. Properties:

- **Deterministic** — no dialog to detect/dismiss, no boot-timer race, no added latency on the trusted path.
- **Read-gated / idempotent** — `trust()` runs only when `readTrusted()` is false → one write per cwd, ever (the write-race window against claude's own writes shrinks to a single write).
- **Subscription-only for free** — sits after the existing `isApiKeyMode()` short-circuit; never runs in api-key mode (where an untrusted cwd wedges nothing).
- **Best-effort** — a trust read/write failure (EACCES/ENOSPC, malformed config, race) is **logged and swallowed**, falling through to the spawn. `calibrate()` (`usage-limits.ts`) awaits `scrape()` **unguarded**, so a rejection would surface as a refresh-route 500; the guard preserves the null-on-every-failure contract.
- **Config-dir-aware** (`src/claude-trust.ts`, mirroring `sandbox.ts:404`) — targets `$HOME/.claude.json` for the default config dir, else `${CLAUDE_CONFIG_DIR}/.claude.json`, so a custom-config-dir install seeds the file claude actually reads.

## Scope
Server-only reliability fix. The broader user-facing surfacing (a `claude_trust` Diagnose check + one-click fix, and whether unattended autopilot **worktree-path** spawns wedge on the same dialog) is deferred to **#1683**.

## Tests
- `test/claude-trust.test.ts` — config-dir-aware path selection (`$HOME` vs `CLAUDE_CONFIG_DIR`), round-trip, field preservation, missing/malformed file, compact JSON output.
- `test/usage-probe.test.ts` — pre-seed once when untrusted; never when trusted; never in api-key mode; `scrape()` resolves to `null` (never throws) when the trust read/write throws.

Local: `bun run lint`, `bun run typecheck`, `bun test ./test` (6808 pass) all green.

### DIFF
--- a/src/claude-trust.ts
+++ b/src/claude-trust.ts
@@ -1,0 +1,78 @@
+/**
+ * Read/seed Claude Code's per-folder trust flag (`hasTrustDialogAccepted`) in the
+ * `.claude.json` config file, so an ephemeral spawn (the /usage probe) doesn't wedge
+ * on the "Do you trust the files in this folder?" dialog. `--dangerously-skip-permissions`
+ * does NOT suppress that dialog, and no flag/env skips it — pre-seeding the flag is the
+ * only unattended path (#1075).
+ *
+ * Pure + injectable: every function takes explicit paths so the caller resolves the
+ * config-dir target once and tests never touch a real `$HOME`.
+ */
+
+import { readFile, writeFile, rename } from "node:fs/promises";
+import { join } from "node:path";
+
+/**
+ * The `.claude.json` file Claude Code actually reads, mirroring the sandbox.ts:404 rule:
+ *  - DEFAULT config dir (`${home}/.claude`) → `$HOME/.claude.json` (a SIBLING of `~/.claude`).
+ *  - CUSTOM `CLAUDE_CONFIG_DIR` (anything else) → `${claudeDir}/.claude.json` (INSIDE it).
+ *
+ * Hardcoding `$HOME/.claude.json` would target the wrong file on a custom-config-dir
+ * install: the seed would miss (dialog still fires) and the read would always report
+ * untrusted (spurious rewrite every run). Pass `config.claudeDir` here.
+ */
+export function claudeConfigPath(home: string, claudeDir: string): string {
+  return claudeDir === `${home}/.claude`
+    ? join(home, ".claude.json")
+    : join(claudeDir, ".claude.json");
+}
+
+/**
+ * True iff `configPath`'s JSON records `projects[dir].hasTrustDialogAccepted === true`.
+ * Any failure (missing file, malformed JSON, unreadable) → `false` — treat "can't
+ * confirm trusted" as untrusted so the caller seeds it.
+ */
+export async function readRepoRootTrusted(configPath: string, dir: string): Promise<boolean> {
+  try {
+    const j = JSON.parse(await readFile(configPath, "utf8"));
+    return j?.projects?.[dir]?.hasTrustDialogAccepted === true;
+  } catch {
+    return false;
+  }
+}
+
+/**
+ * Seed `projects[dir].hasTrustDialogAccepted = true` in `configPath`, PRESERVING every
+ * other field. Writes unconditionally — the caller gates on {@link readRepoRootTrusted}
+ * so this only runs when actually untrusted (one write per dir, ever).
+ *
+ * Whole-file read-modify-write via a temp file + atomic rename: the file is never left
+ * torn, but Claude writes this file frequently (onboarding/project state), so a stale
+ * read + our rename can CLOBBER concurrent Claude state written between them
+ * (last-writer-wins). Accepted given the read-gate's rarity and the sub-ms window.
+ *
+ * Emits COMPACT JSON (no pretty-print) to match Claude's on-disk format and avoid
+ * reflowing a possibly-large file.
+ */
+export async function trustRepoRoot(configPath: string, dir: string): Promise<void> {
+  let j: Record<string, unknown> = {};
+  try {
+    const parsed = JSON.parse(await readFile(configPath, "utf8"));
+    if (parsed && typeof parsed === "object") j = parsed as Record<string, unknown>;
+  } catch {
+    // Missing or malformed → start from an empty config; Claude tolerates a minimal file.
+    j = {};
+  }
+  const projects = (j.projects && typeof j.projects === "object" ? j.projects : {}) as Record<
+    string,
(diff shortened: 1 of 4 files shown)


scoop (commented):
## Summary

This PR adds a config-dir-aware trust pre-seed (`src/claude-trust.ts`) that the `/usage` probe runs before spawning `claude`, so an untrusted `repoRoot` no longer wedges on the folder-trust dialog and returns a null scrape (#1075). The implementation is clean, injectable, and well-tested, and it satisfies the task.

### Verified

- **Config-dir routing matches Claude's real rule.** `claudeConfigPath` returns `$HOME/.claude.json` for the default `~/.claude` config dir and `${claudeDir}/.claude.json` for a custom one (`src/claude-trust.ts:33-36`), mirroring `src/sandbox.ts:403-405` (the `claudeDir !== \`${home}/.claude\`` guard that rw-binds `${claudeDir}/.claude.json`). `config.claudeDir` is `CLAUDE_CONFIG_DIR ?? ${HOME}/.claude` (`src/config.ts:543`), so the constructor's `claudeConfigPath(process.env.HOME ?? "", config.claudeDir)` (`src/usage-probe.ts:79`) resolves the same file Claude uses. Confirmed against the two path-selection tests (`test/claude-trust.test.ts:20-27`).
- **Best-effort guarantee holds.** `calibrate()` awaits `this.probe.scrape()` unguarded (`src/usage-limits.ts:443`), so a rejection would surface as a refresh-route failure. The pre-seed's `try/catch` at `src/usage-probe.ts:118-122` contains every read/write failure and falls through to the spawn — matching the two "never throws" tests (`test/usage-probe.test.ts:143-162`).
- **Read-gate + api-key ordering.** `if (!(await this.readTrusted())) await this.trust()` writes at most once per cwd (`src/usage-probe.ts:119`), and the `isApiKeyMode()` short-circuit at `src/usage-probe.ts:106` fires before the pre-seed — verified by `test/usage-probe.test.ts:98-124` (seeds once when untrusted; no write when trusted; untouched in api-key mode).
- **Callers unaffected.** The new `deps` param is optional (`= {}`); the only production caller `new HerdrUsageProbe(herdr)` at `src/index.ts:684` is unbroken.
- **Atomicity.** The temp file `${configPath}.tmp.${process.pid}` sits in the same directory as `configPath`, so the `rename` is same-filesystem (`src/claude-trust.ts:75-77`). Field preservation confirmed against `test/claude-trust.test.ts:44-64`.

### Non-blocking concern (see findings)

`trustRepoRoot` collapses "file missing" and "file present but unparseable" into the same `j = {}` branch (`src/claude-trust.ts:59-65`), then atomically renames a minimal `{projects:{[dir]:{hasTrustDialogAccepted:true}}}` over `configPath`. For a genuinely-absent file that is correct. But if the read catches an existing `.claude.json` in a transient/torn state (e.g. Claude mid-write, if Claude does not write atomically), the parse throws and the whole real config — `oauthAccount`, every other `projects` entry, history — is overwritten with the minimal stub. Reachability is narrow (read-gated to the untrusted-only onboarding path, and only if a parse error coincides), and the docstring already accepts a last-writer-wins clobber window, so this is a comment, not a blocker. Distinguishing `ENOENT` (safe empty-start) from a parse failure (bail without writing, rather than overwrite) would remove the data-loss path. Note the existing malformed-JSON test (`test/claude-trust.test.ts:87-96`) uses genuinely-garbage input and so asserts, rather than guards against, this overwrite behavior.

No locale/i18n surface in this change (backend TS only). No blocking defects found.

PR #1696, opened 2026-07-12
feat(ui): experimental terminal font-size stepper in wrench menu

scoop:
## What

Adds an **experimental** text-size control to the terminal's wrench menu (`RedrawMenu`) so agent output can be bumped up or down — handy when it gets squinty, especially on mobile.

- **Stepper** `A− / size / A+` appended below the four existing redraw actions.
- **Global**, **persisted** preference (one size per device, survives reload) via a new `terminal-font-size.svelte.ts` store (mirrors `herd-width`).
- Applied **live** to xterm in `Viewport.svelte` — a dedicated `$effect` mutates `term.options.fontSize` in place and re-fits (cols/rows → PTY resize). The terminal/PTY is **never** recreated: the initial size is read `untrack`ed at creation.
- **Integer-snapping** step (`A+` → `floor+1`, `A−` → `ceil−1`), so the fractional 12.5 desktop default becomes clean integers on first use (12.5 → 13/12, then ±1). Range `[8, 24]`.
- Unset preference renders at **today's exact defaults** (12.5 desktop / 11 mobile) — no visual change until the user opts in.

## How it hangs together

`terminalFontSize` store → `Viewport` (`effectiveFontSize`, `stepTerminalFont`, live effect) → `ViewportHeaderActions` (forwards 4 props) → `RedrawMenu` (stepper row).

## a11y / design-system

- `−`/`+` are **44×44** touch targets; colors/type via tokens only (reuses the `.rm-item` recipe + `--fs-*` / `--color-*`).
- **Reviewer heads-up:** the stepper's *label/hint* text uses the established menu recipe (`--fs-base` 13px / `--fs-micro`), matching the four existing menu items — this PR doesn't introduce sub-16px chrome, it follows the menu's existing scale. Terminal *content* dropping below 16px is the user's deliberate choice via this control.
- Keyboard: `A−`/`A+` are consecutive `.rm-item`s, so the menu's existing ArrowUp/Down + Home/End/Tab roving focus reaches them; on hitting a bound the just-disabled button hands focus to the enabled sibling (never `<body>`). Left/Right stay unbound, matching the vertical `role="menu"` pattern.

## Feature discovery / i18n

- What's-New entry `v1.44.0-terminal-font-size.ts` (+ EN/DE `feat_terminal_font_size_*`).
- New menu keys `redrawmenu_textsize_*` in EN + DE (check:i18n parity green).

## Testing

- `cd ui && bun run check` (svelte-check 0 errors) · `bun run check:i18n` (parity) · root `bun run lint` (eslint) — all green.
- `terminal-font-size.svelte.test.ts` — 8 passing (clamp bounds, set persists, reset).
- `RedrawMenu.browser.test.ts` — 10 passing (7 existing + 3 new: `A±` deltas, disabled-at-bound, focus redirect).


scoop (commented):
## Summary

This PR adds an experimental terminal font-size stepper to the wrench (redraw) menu, backed by a per-device `localStorage`-persisted store. The implementation is careful and well-documented, and it satisfies the task (bump the agent output up/down, remembered across sessions, useful on mobile). I found no blocking correctness, security, or quality problems.

- **i18n parity** — the four `redrawmenu_textsize_*` keys and both `feat_terminal_font_size_*` keys exist in *both* catalogs: `ui/messages/en.json:710-713` / `ui/messages/de.json:710-713` and `en.json:2300-2301` / `de.json:2300-2301` (grep counts: 4/4 and 2/2 in each file). No missing-locale defect.
- **Feature announcement wiring** — `ui/src/lib/feature-announcements/entries/v1.44.0-terminal-font-size.ts` is auto-registered by the glob at `ui/src/lib/feature-announcements.ts:47` (`import.meta.glob("./feature-announcements/entries/*.ts")`); its shape is identical to the sibling `v1.44.0-newtask-wider.ts:3-8`; `id` (`terminal-font-size`) is unique and `titleKey`/`bodyKey` resolve to the keys added above.
- **Import path** — `import { terminalFontSize, FONT_MIN, FONT_MAX } from "$lib/terminal-font-size.svelte"` (Viewport.svelte:24) omits `.ts`, matching the established `.svelte.ts` convention used for `$lib/herd-width.svelte` (`src/routes/+page.svelte:136`).
- **No PTY teardown on step** — the terminal-creation effect reads the pin via `untrack(() => terminalFontSize.size)` (Viewport.svelte:1491), so a stepper change never re-keys that effect; the live-apply effect (Viewport.svelte:2108-2118) mutates `term.options.fontSize` in place and early-returns when unchanged, so there is no double-fit on mount and no socket recreation.
- **Single-Viewport invariant holds** — both `<Viewport>` sites are in mutually-exclusive `{:else if selected}` branches (`src/routes/+page.svelte:2685`, `:2851`), so only one instance is ever mounted, and the redraw menu is gated on `tab === "term"` (`ViewportHeaderActions.svelte:99`) while `el` is `display:none` off the term tab (`Viewport.svelte:2805`). The "hidden mount" branch the author documents in the live effect is therefore genuinely unreachable for user-driven changes — the reasoning in the comment checks out.
- **Clamp/step math** — `stepTerminalFont` (Viewport.svelte:336-340) snaps the fractional 12.5 default toward a clean integer before stepping, and `clampTerminalFontSize` (terminal-font-size.svelte.ts:14) plus `set()` bound every stored value to [8,24]; `fontAtMin`/`fontAtMax` (Viewport.svelte:2693-2694) gate the buttons consistently with the store bounds.
- **Roving focus** — the `.rm-step` buttons carry `.rm-item`, so `items()` (RedrawMenu.svelte:87) includes them in arrow/Tab navigation, and `step()` redirects focus to the enabled sibling after a bound disables the clicked button (RedrawMenu.svelte:52-58).

## Nit (non-blocking)

- `ui/src/lib/components/RedrawMenu.svelte:200` — for a fresh **desktop** user the readout renders the raw default `12.5` (a decimal) even though every stepped value is an integer, and it slightly exceeds the `min-width: 3ch` readout box. Purely cosmetic; consider displaying the effective size rounded (or showing the default without the fraction) so the control never surfaces a `.5`.

## Notes

- Could not exercise xterm's `CharSizeService` re-measure behavior at runtime (behavioral, not verifiable from source), but the visibility guard and the single-Viewport invariant above make the documented stale-metrics branch unreachable in practice.

scoop (commented):
## Re-review: experimental terminal font-size stepper

The PR adds a per-device, persisted terminal-output font-size control to the wrench (Redraw) menu, wired through `Viewport → ViewportHeaderActions → RedrawMenu`, backed by a new `terminal-font-size.svelte.ts` store, plus a feature announcement and tests. It satisfies the task.

### Prior finding — RESOLVED
Previous revision: `RedrawMenu.svelte` readout showed the raw `12.5` (decimal, wider than the 3ch box) for a fresh desktop user.
- Now the menu is fed `effectiveFontSize = terminalFontSize.size ?? Math.round(termFontDefault)` (verified `ui/src/lib/components/Viewport.svelte:338`), so a fresh desktop user sees `Math.round(12.5) = 13` — an integer that fits the `min-width:3ch` box (`ui/src/lib/components/RedrawMenu.svelte:200,318-319`). Passed in at `ui/src/lib/components/Viewport.svelte:2698`. The fix is correctly applied. Not re-raised.

### Verification performed
- **Default/actual split is intentional and consistent.** The wrench UI + bound gating use the rounded `effectiveFontSize` (`Viewport.svelte:338,2698-2700`), while the *actual* xterm size when unset stays the true `termFontDefault` (12.5/11) at creation (`Viewport.svelte:1497,1500`) and in the live-apply effect (`Viewport.svelte:2115`). The ≤0.5px gap in the never-touched state is documented and closes on first step. `stepTerminalFont` snaps to integers off the rounded base (`Viewport.svelte:342-346`), so stepping is always a clean ±1.
- **No terminal/PTY recreation on step.** The creation effect reads the persisted size UNTRACKED (`Viewport.svelte:1497`); live changes go through the dedicated in-place effect (`Viewport.svelte:2114-2124`), mirroring the theme effect. The `=== size` early-return (`Viewport.svelte:2119`) avoids a redundant `fit()` on creation and on terminal rebuild. `fitRef` is set at creation and cleared on teardown (`Viewport.svelte:1518,2086`).
- **Store clamp/persist.** `clampTerminalFontSize` bounds to [8,24] without rounding (`terminal-font-size.svelte.ts:12-14`); `set()` clamps+persists, `reset()` clears — matches the tests and the `herd-width` pattern.
- **Bounds gating.** `fontAtMin/fontAtMax` computed from `effectiveFontSize` (`Viewport.svelte:2699-2700`); `effectiveFontSize` is always integer (stepper emits integers; unset path is `Math.round`), so `<=`/`>=` gating is exact.
- **Roving focus + focus redirect.** The −/+ buttons are `.rm-item` so they join `items()` roving focus (`RedrawMenu.svelte:86-88`); `step()` awaits `tick()` and redirects focus to the enabled sibling when the clicked button disables at a bound (`RedrawMenu.svelte:52-58`), preventing a drop to `<body>`. Covered by the new test.
- **i18n parity.** All six new keys exist in both catalogs — verified `ui/messages/en.json:710-713,2300-2301` and `ui/messages/de.json:710-713,2300-2301`.
- **Feature announcement.** New entry `v1.44.0-terminal-font-size.ts` uses `titleKey`/`bodyKey` that both exist (above); entries are auto-loaded via `import.meta.glob` (`ui/src/lib/feature-announcements.ts:47-48`), so no manual registration is needed. `sinceVersion` 1.44.0 matches the current dev line.
- **Prop wiring.** `ViewportHeaderActions` is rendered only from `Viewport.svelte:2677`, and all four new required props are passed; RedrawMenu declares them optional-with-defaults so standalone test render sites still type-check.

- ui/src/lib/components/Viewport.svelte:2114-2124: the hidden-mount branch (store size, skip `fit()`) is documented as effectively unreachable because the only font-size trigger is the wrench menu, which renders on the visible terminal tab. If a future code path mutates `terminalFontSize` while the mount is hidden, a later `refit()` would reflow against stale CharSizeService cell metrics. Dormant today and clearly reasoned in-comment; noting only for the record.

No blocking issues found; nothing to change.

PR #1701, opened 2026-07-13
feat(ui): sortable Created column in Files tab

scoop:
## What

The Files tab (scratchpad **and** worktree browsers) now shows each entry's **creation date** in a sortable **Created** column. Users can sort by **Name** or **Created** by clicking the column headers; directories always stay grouped above files.

Closes the request to "add the file's creation date to the files tab (scratchpad/worktree)… make it a column that I can also sort by."

## How

- **Server (`src/fs-browse.ts`)** — each listing entry now carries `createdMs`: birth time (`birthtimeMs`) when the filesystem records it, else a modified-time (`mtimeMs`) fallback, so it's never blank for a normal entry. Regular files/dirs weren't `stat`'d before; they are now. A `stat` failure **keeps** the entry (just omits the date) — no new drop path.
- **UI (`FilesPanel.svelte`)** — rows are now a 4-track grid (icon · name · created · trailing affordance), with a clickable sort-header row. Sorting is client-side over the loaded directory (no refetch); the active column shows an asc/desc arrow. Created defaults to newest-first; Name defaults to today's dirs-first + alpha order. Dates render as compact relative age with the absolute date on hover; a missing/non-finite date renders a guarded `—` (never `NaNs`).
- **Types, demo seed, i18n (EN+DE)** threaded through; **v1.44.0 feature-announcement** entry added.

### Worktree caveat
Git writes all worktree files at checkout, so their birth times cluster and per-file "created" dates there are near-uniform. The column is kept consistent across both sources (the scratchpad — the primary case — has genuinely varied times); this is documented in code.

## Testing
- `src/fs-browse.ts`: new tests for `createdMs` population (file + dir) and the stat-failure-keeps-entry invariant; existing `onEscape:"mark"` assertions updated.
- `FilesPanel.browser.test.ts`: Created-cell render, missing-date `—` guard, default name-asc order, and Created sort toggle (newest-first → oldest-first) preserving dirs-first.
- Full root `bun test` (6838 pass), UI `bun run check` (0 errors), `check:i18n` parity, UI node + FilesPanel browser suites, and `fallow audit` all green.


scoop (commented):
## Summary

Adds a sortable **Created** column to the Files tab (scratchpad + worktree), backed by a new `createdMs` field derived server-side from `stat` (birth time, mtime fallback). The implementation is correct, well-guarded, and complete end-to-end.

## What I verified

- **Backend derivation** — `createdMsFromStat` returns `birthtimeMs > 0 ? birthtimeMs : mtimeMs` (`src/fs-browse.ts:29-31`). All three `classifyDirent`/`classifySymlinkTarget` paths populate it; a `stat` failure omits `createdMs` but still surfaces the entry (verified classification uses `readdir({withFileTypes})` dirent flags, not `stat` — `src/fs-browse.ts:174-186`).
- **Plumbing** — server serializes the listing directly via `json(listing)` (`src/server.ts:2151,2185`) and `ScratchListing = BrowseListing` (`src/scratchpad.ts:22`), so `createdMs` reaches the UI unmapped. Field added to both `BrowseEntry` (`src/fs-browse.ts:22`) and `ScratchEntry` (`ui/src/lib/types.ts:203`), optional/additive — no consumer breakage.
- **Sort logic** — traced `byCreated`/`byName`/dirs-first (`FilesPanel.svelte`) against all three browser tests: name-asc default (`aaa-dir,zzz-dir,a.txt,b.txt,c.txt`), created-desc, and created-asc all match; missing `createdMs` sorts last in both directions; equal dates fall back to name. Confirmed against `FilesPanel.browser.test.ts:283-299`.
- **NaN guard** — `hasCreated` checks `!= null && Number.isFinite`; the `createdCell` snippet renders a guarded em-dash otherwise, and `relativeAge` clamps future/negative deltas to `"now"` (`ui/src/lib/format.ts:236-239`) — no `NaN` can render.
- **i18n parity** — all 7 new keys exist in BOTH catalogs (verified `ui/messages/en.json:2237-2243` against `ui/messages/de.json:2237-2243`); all 5 UI keys are referenced in the component.
- **Feature announcement** — `sinceVersion: "1.44.0"` is correctly one ahead of `package.json`'s `1.43.0` (`package.json:3`); `targetId: "files-tab"` matches the `use:coachTarget={"files-tab"}` anchor (`ViewportTabBar.svelte:256`); `titleKey`/`bodyKey` resolve to existing message keys.
- **Grid layout** — `--files-cols` (4 tracks) maps 1:1 to the 4 header cells and to the 4 children in each of the three row variants (linkOutside / dir / file); header `.hcreated` and cell `.created` both `justify-self: end` in the same fixed track.
- **Demo seed** — `createdMs` values use `NOW`/`HOUR`/`MIN`/`DAY`, all defined at `ui/src/lib/demo/seed.ts:66-70`.
- **stat-failure test validity** — `resolveInRoot` uses `fsp.realpath`, not `fsp.stat` (`src/fs-browse.ts:76,93`), so the test's `spyOn(fsp,"stat")` mock doesn't break listing resolution; the assertion (entry kept, `createdMs` undefined) is sound.

## Nit (non-blocking)

- `ui/src/lib/components/viewport/FilesPanel.svelte`: the header uses `role="row"` + `role="columnheader"` + `aria-sort` on `.list-head`, but the enclosing `.list` div has no `role="table"`/`"grid"`/`"rowgroup"` and the data rows are plain buttons/anchors without row/cell roles. Per ARIA, an orphaned `columnheader`/`aria-sort` outside a table/grid structure is ignored by assistive tech, so the sort direction is not conveyed to AT. It's cosmetic here because the sort buttons already carry explicit `aria-label`s (`files_sort_name_aria`/`files_sort_created_aria`), so the controls remain operable — but the `aria-sort` state is dead markup as written.

No blocking issues found.

scoop (commented):
## Re-review

The single prior finding is **resolved**. `ui/src/lib/components/viewport/FilesPanel.svelte` no longer emits `role="row"`/`role="columnheader"`/`aria-sort` on `.list-head`; sort state is instead announced through each header button's `aria-label` (`sortLabel()` folds the active direction into the label via `files_sort_active`). Verified no stray roles remain — `grep` finds only the unrelated `role="region"` (FilesPanel.svelte:235) and `role="group"` (:241). The comment at :49-51 documents the deliberate choice, which is sound (rows are interactive `<a>`/`<button>` elements; a real ARIA grid would strip their semantics).

## Verification

- **`relativeAge`**: called as `relativeAge(c, now)` (FilesPanel.svelte:317); signature `relativeAge(fromMs, nowMs)` confirmed at ui/src/lib/format.ts:236. Negative/near-zero deltas collapse to "now" (format.ts:238), so slight client/server clock skew on freshly-checked-out worktree files won't render a future age.
- **i18n parity**: all 10 new keys present in both catalogs — ui/messages/en.json:2237-2246 and ui/messages/de.json:2237-2246. `files_sort_active` uses `{label}`/`{dir}` in both; call site passes `{ label: base, dir }` (FilesPanel.svelte:55). Only en + de catalogs exist (ls ui/messages/).
- **Data plumbing**: `createdMs` added to `BrowseEntry` (src/fs-browse.ts:22) and `ScratchEntry` (ui/src/lib/types.ts:203); server returns the listing unmodified via `json(listing)` (src/server.ts:2151, 2185), and the UI API layer is a raw `r.json()` passthrough (ui/src/lib/api.ts:650, 665). `listWorktree`/`listScratchpad` both delegate straight to `listDir` (src/worktree-files.ts:14, src/scratchpad.ts:52).
- **stat-failure safety**: `classifyDirent` now stats every non-symlink file/dir but swallows a failing stat and omits `createdMs` rather than dropping the entry (src/fs-browse.ts:174-186); covered by test/fs-browse.test.ts:99-119. The UI `hasCreated` guard + em-dash snippet (FilesPanel.svelte:314-325) prevents `NaN` rendering for missing/non-finite values, matching the browser test at FilesPanel.browser.test.ts:257-269.
- **Sort behavior**: dirs group above files, missing dates sort last regardless of direction, created defaults to newest-first (FilesPanel.svelte:69-88) — consistent with the three browser-test expectations (FilesPanel.browser.test.ts:281-300). Header only renders on a non-empty listing (FilesPanel.svelte:333/340), and the 4-track grid on `.list-head` and `.row` shares `--files-cols` so header labels align over cells.
- **Feature announcement**: entry is auto-globbed (feature-announcements.ts:47), `targetId: "files-tab"` is a real coach target (ViewportTabBar.svelte:256), and `sinceVersion: "1.44.0"` matches the current entry batch.

No correctness, security, or quality issues found in the diff. Task (surface file creation date in the Files tab with sortable column) is satisfied.

Non-blocking observations (not defects, no action required): `byCreated` breaks ties by name always-ascending even under descending sort (documented at FilesPanel.svelte:77), and every non-symlink entry now incurs an extra `fsp.stat` per listing — an intentional, documented cost for obtaining birth time.

PR #1702, opened 2026-07-13
fix(ui): cap Diff file sidebar at 1/3, collapse paths keeping the filename

scoop:
## What

On the new `@pierre/diffs` Diff tab, the file-name sidebar could take too much room and long paths were end-ellipsized (`…v1.44.0-files-created-colu…`), hiding the filename. This guarantees the diff keeps **≥2/3** of the panel and makes the filename always fully visible.

## Changes

- **Proportional cap, no floor** (`DiffPanel.svelte`): sidebar `flex: 0 0 min(30%, 320px)`. The panel sits in the desktop grid's `1fr` region (`minmax(300px,360px) 1fr`) — at a 769px viewport it's only ~469px wide and the 768px chip-strip breakpoint hasn't engaged, so a fixed px floor would eat >1/3 there. 30% keeps the diff stack at **≥70% (> 2/3)** at every two-pane width; capped at 320px so wide panels don't waste rail width.
- **Filename never truncated** (`DiffFileSidebar.svelte`): the path renders as a shrinking, muted directory span (`text-overflow: ellipsis`) + a **non-shrinking filename span**. Width can only collapse the directory prefix, never the filename tail. Full path stays in the row `title`.
- **Structural path collapse** (`diff-path.ts`, new): `collapsePath`/`pathParts` keep the top-level dir + the file's **immediate parent** (`root/…/parent/`, e.g. `ui/…/viewport/FilesPanel.svelte`) — the parent is what disambiguates same-named files. No width/char-count heuristic.

## Tests

- New `diff-path.test.ts` (node vitest): long+deep, short pass-through, shallow (≤2 dir segs), top-level (no slash), renamed `old → new`.
- `bun run check` clean; `test:node` (2278) green; existing `DiffPanel.browser.test.ts` (6) unchanged — the two adjacent spans concatenate to the same text.
- No new i18n keys (path text is verbatim data). Labeled `fix` (refinement of the shipped Diff tab, not a new capability) so no feature-catalog entry is required.

### DIFF
--- a/ui/src/lib/components/DiffFileSidebar.svelte
+++ b/ui/src/lib/components/DiffFileSidebar.svelte
@@ -4,6 +4,7 @@
   // horizontally-scrollable chip strip (CSS-only; one markup). Purely a
   // files -> rows mapping + selection callback — no business logic.
   import type { DiffFile } from "$lib/types";
+  import { pathParts } from "$lib/diff-path";
   import { m } from "$lib/paraglide/messages";
 
   let {
@@ -48,7 +49,12 @@
           <span class="glyph status-{file.status}" aria-hidden="true">
             {STATUS_GLYPH[file.status]}
           </span>
-          <span class="path">{label(file)}</span>
+          <span class="path"
+            >{#each pathParts(file) as part, i (i)}{#if i > 0}<span class="arrow" aria-hidden="true"
+                  >&nbsp;→&nbsp;</span
+                >{/if}<span class="dir">{part.dir}</span><span class="name">{part.name}</span
+              >{/each}</span
+          >
           <span class="counts">
             <span class="add">+{file.additions}</span>
             <span class="del">−{file.deletions}</span>
@@ -126,11 +132,30 @@
   .status-modified {
     color: var(--color-amber);
   }
+  /* Filename never shrinks (always fully visible); the muted directory prefix
+     ellipsis-collapses when the rail is tight. */
   .path {
     flex: 1;
+    min-width: 0;
+    display: flex;
+    align-items: baseline;
+    overflow: hidden;
+  }
+  .dir {
+    flex: 0 1 auto;
+    min-width: 0;
     overflow: hidden;
     text-overflow: ellipsis;
     white-space: nowrap;
+    color: var(--color-muted);
+  }
+  .name {
+    flex: 0 0 auto;
+    white-space: nowrap;
+  }
+  .arrow {
+    flex: 0 0 auto;
+    color: var(--color-faint);
   }
   .counts {
     flex-shrink: 0;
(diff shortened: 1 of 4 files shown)


scoop (commented):
## Summary

The PR resizes the Diff-tab file rail so the actual diff always gets at least 2/3 of the width, and replaces the full-path label with a collapsed `root/…/parent/filename` rendering. Both goals from the task are met, the new logic is factored into a small pure module with focused tests, and I found no correctness or security issues.

## What I verified

- **≥2/3 for the diff.** `DiffPanel.svelte:236` changes the sidebar from a fixed `flex: 0 0 220px` to `flex: 0 0 min(30%, 320px)`, while the stack stays `flex: 1` (`DiffPanel.svelte:238-241`) inside a gap-less `display:flex` `.content`. So the diff stack always receives ≥70% (>2/3). The panel sits in the desktop grid's `1fr` region (verified against `ui/src/routes/+page.svelte:3246` and `:3315`, where the left column is `minmax(...)` and the panel is `1fr`), so a proportional cap rather than a px floor is the right call — a 220px floor would eat >1/3 in the narrow-but-still-desktop band. `min()` in `flex-basis` is supported by the modern browsers this UI targets.
- **Path collapsing.** `collapsePath` (`diff-path.ts:15`) splits off the filename (always kept whole), returns the full dir for ≤2 segments, and collapses to `root/…/parent/` for deeper paths. I traced every case in `diff-path.test.ts` (deep collapse, single-segment, two-segment no-collapse, top-level empty dir) and they all follow from the code. `pathParts` (`diff-path.ts:24`) emits one part normally and two (old → new) only for `renamed` with an `oldPath`, matching its tests.
- **Type contract.** `pathParts`/`collapsePath` read only `path`, `oldPath?`, `status` — exactly the fields on `DiffFile` (`ui/src/lib/types.ts:1872-1877`), where `oldPath` is documented as set only when renamed. The Svelte `{#each pathParts(file)}` block (`DiffFileSidebar.svelte:53-56`) renders dir+name per part with a hidden `→` separator between parts; the full path is still exposed via the `title` tooltip (`label` retained at `DiffFileSidebar.svelte:28`).
- **CSS behavior.** `.dir` (`flex:0 1 auto; min-width:0; overflow:hidden; text-overflow:ellipsis`) ellipsis-collapses under pressure while `.name` (`flex:0 0 auto`) does not shrink — the directory prefix absorbs the squeeze, which is the intended "keep the filename" behavior. No new i18n keys were introduced, so no catalog-parity concern.

The comment at `DiffFileSidebar.svelte:135` says the filename is "always fully visible." Strictly, when the filename alone is wider than the (capped) rail, `.name` cannot shrink and `.path` has `overflow:hidden`, so a very long filename clips at the right edge with no ellipsis. This is an intentional, rare degradation and is mitigated by the full-path `title` tooltip; noting only that the comment slightly overstates the guarantee. Not actionable.

No blocking issues found.

PR #1700, opened 2026-07-12
feat(ui): fancier Diff tab via @pierre/diffs (split/unified, word-diff, sidebar)

scoop:
Replaces the session **Diff tab**'s hand-rolled renderer with **`@pierre/diffs`** (the diffs.com / hunk.dev engine), fed from the existing `GET /api/sessions/:id/diff` pipeline. Implements the de-risked integration from the research spike (PR #1691).

Closes #1698.

## What's in it
- **Split ↔ unified toggle** (segmented control, persisted per-browser via `localStorage`; **split-first on wide**, forced **unified on narrow ≤768px**).
- **Word-level intra-line diff** + **collapsed unchanged context** (free with Pierre).
- **File sidebar** built from our per-file `DiffResult` (A/M/D/R glyph + `±` stats), **click-to-scroll** with render-then-scroll correction; collapses to a compact picker on narrow.
- **Lazy per-file render** (IntersectionObserver) with reserved height estimates — keeps the DOM bounded under the server's 2000-line/file cap (no virtualization needed).
- **DARK + LIGHT parity**: our existing Shiki themes are injected into Pierre (`registerCustomTheme`); theme toggle is instant (`setThemeType`, no re-render).
- **15s poll is flash-free**: per-file content **signature** gates re-render, so unchanged files keep their DOM + scroll across ticks.

## Server change (no payload doubling)
The server now surfaces a **lossless per-file `patch`** (`parseUnifiedDiff` captures each file's raw git block) instead of the client reconstructing one. To avoid shipping the diff text twice: the **session endpoint sends `patch`, strips `hunks`**; the **recap keeps `hunks`, strips `patch`**. Server `DiffFile` keeps `hunks` required (so `visual-blocks.ts` is untouched); only the client type relaxes `hunks?`. Binary/truncated files carry neither and render an i18n note card.

## Annotations — deferred (plumbed, fed `[]`)
The review-first payoff (per-line agent notes) is the reason Pierre was chosen. The wrapper exposes typed `lineAnnotations`/`renderAnnotation`, but this PR feeds an **empty list** — no annotation UI ships. The **data-source decision** (tool-call spans / plan steps / review pass; read-only vs thread) is tracked in **#1699** (owner @plenz, target v1.45.0). Sourcing the data later reuses this exact substrate with no viewer rework.

## Gates
[...]

Issue #1698, opened 2026-07-12
Implement fancier Diff tab with @pierre/diffs (review-first, split + word-diff + agent annotations)


scoop:
## Decision

Replace the session Diff tab's custom renderer with **`@pierre/diffs`** (the engine behind diffs.com and hunk.dev). Chosen over `@git-diff-view/svelte` and in-house-extend for the **review-first ceiling**: split + word-level diff **and** per-line agent annotations, all reusing our Shiki setup.

Research + validation: [`docs/research/external-diff-integration.md`](docs/research/external-diff-integration.md) (PR #1691). A working preview validated the full substrate end-to-end (theming, split, unified, line-aligned annotations) — this is a de-risked integration, not a spike.

## Proven approach (from the spike — don't re-discover these)

```js
import { FileDiff, parsePatchFiles, registerCustomTheme } from "@pierre/diffs";
// 1. Theming: inject Shepherd's Shiki themes (Shadow DOM blocks our token cascade)
registerCustomTheme("shepherd-dark", async () => SHEPHERD_DARK);   // objects already in ui/src/lib/highlight.ts
registerCustomTheme("shepherd-light", async () => SHEPHERD_LIGHT);
// 2. Host MUST be <diffs-container> (auto-registered on import), NOT a <div> —
//    only its shadow root adopts Pierre's layout stylesheet (split grid + annotation slots).
const host = document.createElement("diffs-container");
const fd = new FileDiff({
  theme: { dark: "shepherd-dark", light: "shepherd-light" },
  themeType, diffStyle,            // "split" | "unified"
  lineDiffType: "word", diffIndicators: "bars",
  overflow: "wrap",                // 3. required for side-by-side split (scroll => stacks)
  renderAnnotation,                // (a) => HTMLElement, projected into per-line slot
});
fd.render({ fileDiff, fileContainer: host, lineAnnotations });
```

The **three gotchas** (all cost debugging in the spike): (1) theme via Shiki injection, (2) `<diffs-container>` host — this single fix unblocks both split layout **and** annotation line-alignment, (3) `overflow:"wrap"`. Theme switching uses `fd.setThemeType()` (no re-render). Annotations project into `<slot name="annotation-<side>-<lineNumber>">`.

## Scope / work breakdown

- [ ] Add `@pierre/diffs` to `ui/package.json`; drive the vanilla API from a **thin Svelte 5 wrapper** (no first-party Svelte binding). Lazy-load like today's highlighter.
[...]

Issue #1616, opened 2026-07-10
Operator language: mid-session persistence, classifier eval, remaining surfaces (#1586 follow-up)


scoop:
```epic-dag
#1624
#1625
#1626
#1627 <- #1626
#1628
```

**Epic children** (decomposed from this issue; drain the roots first):
- #1624 — mid-session persistence (resume + steer, both providers) *(root)*
- #1625 — remaining LLM output surfaces (rundown/herd-digest, prompt-recommend) *(root)*
- #1626 — classifier live-model eval harness *(root, blocks #1627)*
- #1627 — classifier operator-language + input-robustness *(blocked by #1626)*
- #1628 — server-authored HUD-card strings → per-locale sentinel codes *(root)*

---

Follow-up to #1586 / PR #1615, which shipped operator-language injection for four deterministically-verifiable prompt surfaces (coding-session directive, plan sidecar, recap, plan reviewer). These were deferred there with explicit rationale and are tracked here.

## 1. Mid-session drift — both providers (the marquee gap)

The directive holds only from spawn to the first resume/opening turn:

- **Claude:** it rides `--append-system-prompt`, which `resume()` deliberately does NOT re-pass (the #499 context-trim decision; `buildClaudeResumeArgv` in `src/service.ts`). A long session that compacts/resumes loses it and drifts back to English.
- **Codex:** no `--append-system-prompt`, so it rides inline on the opening prompt only (`src/service.ts`), and no steer re-carries it.

This is #1586's motivating scenario (drift *after* plan reviews / critic feedback / steers). Fix: thread the `<operator-language>` block onto `buildClaudeResumeArgv` (a narrow, deliberate exception to the #499 "no directives on resume" rule) and onto the Codex steer path (`driveSteer` / plan-go / preview steers). Argv presence is deterministically testable; note that whether Claude honors `--append-system-prompt` alongside `--resume` is a runtime behavior needing a real check.

## 2. Autopilot classifier + a live-model eval harness

`classifierPrompt`'s `summary` is the persisted operator-facing pause/complete question (`src/autopilot.ts`), so it qualifies. It was deferred because its risk is *classification quality*: `classifyStop` is tested only by stubbing `readVerdict`, and there is no live-model eval anywhere in `test/`/`scripts/`/`ci/`. Also, the German terminal tail already shifts the classifier's *input* (its English exemplars) — a real regression risk (`src/autopilot-llm.ts` prices its errors as "a bad product decision").

[...]

PR #1703, opened 2026-07-13
feat(ui): source per-line Diff-tab annotations (agent reasoning + critic findings) (#1699)

### LINKED ISSUE #1699: Diff tab: source per-line agent annotations (data for the @pierre/diffs review surface)
Follow-up to #1698 (fancier Diff tab via \`@pierre/diffs\`).

## Context
The new Diff tab renders via \`@pierre/diffs\`, whose per-line annotation framework is fully **plumbed** in our wrapper (\`PierreDiff.svelte\` accepts typed \`lineAnnotations\` / \`renderAnnotation\`) but is currently fed an **empty list** — the *rendering* substrate is proven, the *data* is deferred.

This issue owns the deferred **product decision**: where per-line agent annotations come from, and how they behave.

## Open questions
- **Source of the per-line reasoning**: tool-call spans mapped to file+line? plan steps? a dedicated review pass? Something else?
- **Interaction model**: read-only notes, or an interactive comment thread?
- **Persistence**: ephemeral (recomputed) vs stored per session/PR.

## Scope when picked up
- Decide + document the data source above.
- Map that source to \`DiffLineAnnotation[]\` and feed it through the existing wrapper props (no new rendering work — the substrate exists).
- i18n any annotation chrome; add a feature-announcement entry for the newly-visible capability.

## Owner / timebox
- Owner: @plenz (repo maintainer)
- Target: v1.45.0 cycle

_The review-first ceiling is the reason #1698 chose Pierre over a plainer viewer; this issue is where that payoff lands._

scoop:
Closes #1699. Follow-up to #1698/#1700 — feeds the already-proven `@pierre/diffs` annotation substrate (plumbed in #1700, fed `[]`) with real data.

## What

Two **read-only, ephemeral** annotation sources, both derived from data that already exists (no new storage/schema):

1. **Agent spans** — the agent's own reasoning, anchored to the changed line its edit produced. Turn-grouped transcript extraction (by `message.id`, reusing `readTranscriptTail`), rendered as **Pierre line annotations**.
2. **Review findings** — the session's Critic findings (`ReviewVerdict.findings`), rendered as **Svelte banners** (per-file + panel-level).

New sibling endpoint **`GET /api/sessions/:id/diff/annotations`** — best-effort chrome, fetched on diff-load + manual refresh (**not** the 15s poll), degrades to `[]` on any error so it can never block or break the diff.

## Design decisions (aligned with maintainer during plan gate)

- **Two rendering paths.** Agent spans → Pierre `lineAnnotations`; review findings → Svelte banner. The banner is required regardless of Pierre's line-0 support because binary/truncated/no-changes files have **no Pierre host at all**, and it uniformly carries the panel-level "not-file-specific" findings.
- **Strict-or-drop anchoring.** An edit anchors only on a **unique** multi-line content signature covering a real added line; ambiguous / out-of-diff / trivial-reasoning → **dropped** (never anchor wrong). Honest limits: ~40% of edits carry usable reasoning (Phase-0 measured), `thinking` blocks absent in practice, 512 KB transcript tail.
- **Shared classifier, no double-drop.** Review findings route via a new `attributeFinding` (3-way `matched | unattributed | out-of-diff`) extracted from `critic-core`'s `scopeFindings` logic (refactored onto it, byte-identical). Findings are already scope-filtered at critic time, so the tab **re-routes, never re-drops** — `matched` → per-file banner, `unattributed`/`out-of-diff` → panel banner.
- **XSS-safe.** All verbatim agent/critic text reaches the DOM as text nodes / `{text}` only — never `innerHTML`/`{@html}`.
- **Late-arrival.** Annotations fetched after first render appear via a `hasRendered`-guarded effect (`setLineAnnotations` + `rerender`, no re-parse), with a content-key gate so a fresh `[]` never triggers a redundant re-render.

## Empirical gates (this is a review-first surface — gates are loud-failing)

- **Phase 0 (HALTing):** verified on 3 real transcripts the agent-reasoning source is usable → PROCEED.
- **Browser gates** pin the `@pierre/diffs` API (generic `metadata`, public `setLineAnnotations`/`rerender`). The **line-0 gate found FileDiff DOES honor a file-level slot** (planning assumed it dropped) — the stated fallback was taken: the banner architecture is unchanged (see above).

## Tests

- `test/diff-annotations.test.ts` — anchoring (unique/ambiguous-drop/context-only-drop/short-drop), Edit/MultiEdit/Write extraction, trivial-drop, per-file cap, review routing.
- `test/critic-core.test.ts` — `attributeFinding` discriminant + `scopeFindings` byte-identical parity after the shared refactor.
- `PierreDiff.browser.test.ts` — hard API gate, line-0 observed behavior, XSS-as-text, late-arrival, no-op-empty rerender.
- `DiffFileStack.browser.test.ts` — per-file banner independent of lazy-mount + noteKind, markup escaped.

i18n EN+DE; feature-announcement entry `v1.44.0-diff-annotations`.

### DIFF
--- a/docs/external-task-api.md
+++ b/docs/external-task-api.md
@@ -129,7 +129,10 @@ Once a task exists, an external agent can also drive it:
 - `POST /api/broadcast` — send the same text to many sessions at once.
 - `DELETE /api/sessions/:id` — archive (end) the session.
 - `GET /api/sessions` — list active sessions; `GET /api/sessions/:id/diff`,
-  `/activity`, `/usage` for inspection.
+  `/activity`, `/usage` for inspection. `GET /api/sessions/:id/diff/annotations`
+  returns best-effort per-line Diff-tab annotations (agent reasoning anchored to
+  changed lines plus routed critic findings) as `{ "notes": [...] }`; it degrades
+  to an empty list on any error rather than failing.
 - `GET /api/sessions/:id/scratchpad[?path=]` — browse a live session's own
   scratchpad subtree; `GET /api/sessions/:id/scratchpad/download?path=`
   streams a single file. Paths are relative to the scratchpad root and
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -340,7 +340,7 @@ function isPathShaped(token: string): boolean {
  * `files` set carried on InFlight, so it never touches the poll loop). For each finding, parse a
  * leading `<path>: ` token (stripping an optional `:<line>` suffix on the path) and DROP it iff:
  *   `files` is non-empty AND the leading token is path-shaped AND it does NOT correspond to any
- *   changed file (see `correspondsToChangedFile`: exact, trailing-segment, or basename match).
+ *   changed file (via `attributeFinding` → `matchChangedFile`: exact, trailing-segment, or basename match).
  * Findings with no parseable path prefix are KEPT (unattributed → never drop something we can't
  * attribute). Note this means a finding prefixed with an extensionless path (`Makefile: ...`,
  * `Dockerfile: ...`, `LICENSE: ...`) is NOT path-shaped per isPathShaped, so it is treated as
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -356,35 +356,65 @@ export function scopeFindings(
   const kept: string[] = [];
   const dropped: string[] = [];
   for (const f of findings) {
-    // Leading token = everything up to the first ": ". No ": " → unattributed → keep.
-    const sep = f.indexOf(": ");
-    if (sep < 0) {
-      kept.push(f);
-      continue;
-    }
-    // Strip an optional `:<line>` (or `:<line>:<col>`) suffix so "src/a.ts:42: ..." → "src/a.ts".
-    const token = f.slice(0, sep).replace(/:\d+(?::\d+)?$/, "");
-    if (!isPathShaped(token)) {
-      kept.push(f); // prose prefix (e.g. "Note", "Nit") → not a path → keep
-      continue;
-    }
-    if (correspondsToChangedFile(token, files)) kept.push(f);
-    else dropped.push(f); // path-shaped + provably outside the diff → drop
+    // Reuse the shared classifier so the drop rule can never drift from the Diff-tab
+    // routing (#1699). DROP iff provably outside the diff; keep matched + unattributed —
+    // byte-identical to the pre-refactor per-finding logic.
+    if (attributeFinding(f, files).attribution === "out-of-diff") dropped.push(f);
+    else kept.push(f);
   }
   return { kept, dropped };
 }
 
-/** Does a path-shaped finding token correspond to a file actually changed in the diff? The critic
- *  is instructed to prefix the full repo-relative path, but it sometimes uses just the basename
+/** How a critic finding relates to the diff's changed-file set (`files`), used by BOTH the scope
+ *  backstop (`scopeFindings` drops `out-of-diff`) and the Diff tab's annotation routing (#1699,
+ *  which surfaces `out-of-diff` in the panel banner instead of dropping). Sharing one classifier
+ *  keeps the two consumers from drifting. */
+export type FindingAttribution = "matched" | "unattributed" | "out-of-diff";
+
+export interface AttributedFinding {
+  attribution: FindingAttribution;
+  /** `matched` → the corresponding `DiffFile.path` (NOT the raw token, so a basename/trailing
+   *  token keys the right file); `out-of-diff` → the raw path token; `unattributed` → "". */
+  path: string;
+  /** `matched`/`out-of-diff` → the finding with its `<path>: ` prefix stripped; `unattributed`
+   *  → the whole finding (there is no path prefix to strip). */
+  text: string;
+}
+
+/**
+ * Classify a single critic finding against the diff's changed-file set. Parses a leading
+ * `<path>: ` token (stripping an optional `:<line>[:<col>]` suffix). A finding with no `": "`
+ * separator, or whose leading token is not path-shaped (prose like "Note: "/"Nit: ", or an
+ * extensionless path like `Makefile: `), is `unattributed`. A path-shaped token that corresponds
+ * to a changed file (exact, trailing-segment, or basename match — see `matchChangedFile`) is
+ * `matched`; one that provably does not is `out-of-diff`. PURE. Callers with an empty `files` set
+ * should short-circuit before calling (an empty set would classify every path-shaped finding as
+ * `out-of-diff`).
+ */
+export function attributeFinding(finding: string, files: string[]): AttributedFinding {
+  const sep = finding.indexOf(": ");
+  if (sep < 0) return { attribution: "unattributed", path: "", text: finding };
+  // Strip an optional `:<line>` (or `:<line>:<col>`) suffix so "src/a.ts:42: ..." → "src/a.ts".
+  const token = finding.slice(0, sep).replace(/:\d+(?::\d+)?$/, "");
+  if (!isPathShaped(token)) return { attribution: "unattributed", path: "", text: finding };
+  const text = finding.slice(sep + 2); // everything after the "<path>: " prefix
+  const matched = matchChangedFile(token, files);
+  return matched
+    ? { attribution: "matched", path: matched, text }
+    : { attribution: "out-of-diff", path: token, text };
+}
+
+/** The changed file a path-shaped finding token corresponds to, or null. The critic is instructed
+ *  to prefix the full repo-relative path, but it sometimes uses just the basename
  *  (`Viewport.svelte:`) or a trailing slice (`components/Viewport.svelte:`). Match on any of:
  *  exact equality, the token being a trailing path-segment of a changed file, OR a bare basename
- *  match. Erring toward correspondence (KEEP) is the safe direction — a missed drop only wastes a
- *  round, whereas a false drop hides a real in-diff finding. The cost is that a basename shared by
- *  an unrelated changed file (`index.ts`) won't drop; acceptable given the prompt asks for full
- *  paths and this is only a fallback. */
-function correspondsToChangedFile(token: string, files: string[]): boolean {
+ *  match — returning the first such changed-file path. Erring toward correspondence (a match) is
+ *  the safe direction — a missed drop only wastes a round, whereas a false drop hides a real
+ *  in-diff finding. The cost is that a basename shared by an unrelated changed file (`index.ts`)
+ *  matches; acceptable given the prompt asks for full paths and this is only a fallback. */
+function matchChangedFile(token: string, files: string[]): string | null {
   const base = baseName(token);
-  return files.some((f) => f === token || f.endsWith("/" + token) || baseName(f) === base);
+  return files.find((f) => f === token || f.endsWith("/" + token) || baseName(f) === base) ?? null;
 }
 
 function baseName(p: string): string {
--- a/src/diff-annotations.ts
+++ b/src/diff-annotations.ts
@@ -1,0 +1,303 @@
+// Per-line Diff-tab annotations (#1699). Derives two read-only, EPHEMERAL sources from data
+// that already exists — no new storage:
+//
+//  1. Agent spans  — the agent's own reasoning, anchored to the changed line its edit produced.
+//     Sourced from the session transcript: Phase-0 (see .shepherd-plan.md) established that Claude
+//     Code writes each content block as its OWN assistant JSONL line (thinking / text / tool_use
+//     separate), sharing a `message.id` per turn, and that `thinking` is absent in practice — so
+//     reasoning is read from the turn's `text` blocks (grouped by `message.id`), and each edit is
+//     anchored by a UNIQUE multi-line content signature (ambiguity → DROP, never anchor wrong).
+//  2. Review findings — the session's Critic findings, routed via the shared critic-core
+//     `attributeFinding` classifier: a finding attributable to a diff file → per-file banner;
+//     unattributed OR out-of-diff → panel banner (NEVER dropped here — see the endpoint/client).
+//
+// This module is PURE + sync (except `readTranscriptTail`, itself sync fs) so the builder is unit
+// testable; the endpoint (server.ts) does the async IO (computeDiff / getReview) and calls in.
+
+import { attributeFinding } from "./critic-core";
+import { readTranscriptTail } from "./activity";
+import type { DiffFile } from "./types";
+
+/** One Diff-tab annotation on the wire. `kind:"agent"` carries a line anchor (`side`+`lineNumber`)
+ *  and the tool; `kind:"review"` is file-level (`path`) or panel-level (`path:""`), no anchor.
+ *  NOTE: named `DiffNote`, not `DiffAnnotation` — the UI already owns a `DiffAnnotation` type
+ *  (visual-block prose note). `text` is VERBATIM agent/critic content — never HTML (client renders
+ *  it as a text node only). */
+export interface DiffNote {
+  path: string;
+  kind: "agent" | "review";
+  text: string;
+  side?: "additions" | "deletions";
+  lineNumber?: number;
+  tool?: string;
+}
+
+// ── tuning knobs (all conservative; the design's honest limits live here) ──────────────────────
+const SIG_MAX_LINES = 3; // signature = up to N leading non-blank lines of the edit's new text
+const MIN_SINGLE_LINE_SIG = 20; // a 1-line signature must be at least this long (else too ambiguous)
+const MIN_REASON_LEN = 20; // reasoning shorter than this is treated as trivial → drop the note
+const REASON_MAX = 240; // clamp reasoning to ~this many chars (word boundary), then ellipsize
+const PER_FILE_AGENT_CAP = 8; // at most this many agent notes per file (avoid a wall on churny files)
+
+/** Normalize a line for signature comparison: trim outer whitespace (so indentation differences
+ *  between an edit's `new_string` and the diff's rendered content never block a match). */
+function norm(s: string): string {
+  return s.trim();
+}
+
+/** Up to `SIG_MAX_LINES` leading NON-BLANK lines of `text`, normalized — the anchoring signature. */
+function signatureLines(text: string): string[] {
+  const out: string[] = [];
+  for (const raw of text.split("\n")) {
+    const n = norm(raw);
+    if (n === "") continue;
+    out.push(n);
+    if (out.length >= SIG_MAX_LINES) break;
+  }
+  return out;
+}
+
+/** New-side view of a file's diff: every add/ctx line in document order, with its new-side line
+ *  number and whether it is an actual addition. `del` lines (no new-side number) are skipped. This
+ *  is the sequence an edit's `new_string` signature is matched against. */
+function newSideLines(file: DiffFile): { no: number; content: string; isAdd: boolean }[] {
+  const rows: { no: number; content: string; isAdd: boolean }[] = [];
+  for (const hunk of file.hunks ?? []) {
+    for (const line of hunk.lines) {
+      if (line.kind === "del") continue;
+      if (line.newNo === undefined) continue;
+      rows.push({ no: line.newNo, content: line.content, isAdd: line.kind === "add" });
+    }
+  }
+  return rows;
+}
+
+/**
+ * Anchor an edit's `newString` to a changed line, or return null (DROP). Requires a UNIQUE
+ * consecutive match of the signature against the file's new-side lines, AND that the match covers
+ * at least one real added line (so the anchor is genuinely in the diff, not pure context). Returns
+ * the new-side line number of the first added line within the matched run. A 1-line signature must
+ * clear `MIN_SINGLE_LINE_SIG` chars (a lone `}` / `return;` is too ambiguous to anchor).
+ */
+export function anchorEdit(
+  newString: string,
+  rows: { no: number; content: string; isAdd: boolean }[],
+): number | null {
+  const sig = signatureLines(newString);
+  if (sig.length === 0) return null;
+  if (sig.length === 1 && sig[0]!.length < MIN_SINGLE_LINE_SIG) return null;
+  const normed = rows.map((r) => norm(r.content));
+  const starts: number[] = [];
+  for (let i = 0; i + sig.length <= normed.length; i++) {
+    let ok = true;
+    for (let j = 0; j < sig.length; j++) {
+      if (normed[i + j] !== sig[j]) {
+        ok = false;
+        break;
+      }
+    }
+    if (ok) starts.push(i);
+  }
+  if (starts.length !== 1) return null; // 0 = no match, >1 = ambiguous → never anchor wrong
+  const s = starts[0]!;
+  for (let j = 0; j < sig.length; j++) {
+    if (rows[s + j]!.isAdd) return rows[s + j]!.no;
+  }
+  return null; // signature matched only context lines → not actually in the diff → drop
+}
+
+/** Clamp reasoning to `REASON_MAX` chars at a word boundary, ellipsizing when cut. */
+function clampReason(text: string): string {
+  const t = text.replace(/\s+/g, " ").trim();
+  if (t.length <= REASON_MAX) return t;
+  const cut = t.slice(0, REASON_MAX);
+  const sp = cut.lastIndexOf(" ");
+  return (sp > REASON_MAX * 0.6 ? cut.slice(0, sp) : cut).trimEnd() + "…";
+}
+
+/** One extracted edit operation: the file it targeted (repo-relative), the tool, the new text to
+ *  anchor, and the turn's reasoning. Exported for the extractor's test. */
+export interface TranscriptEdit {
+  path: string; // repo-relative (worktree prefix stripped)
+  tool: string; // "Edit" | "MultiEdit" | "Write"
+  newString: string;
+  reasoning: string;
+}
+
+const EDIT_TOOLS = new Set(["Edit", "MultiEdit", "Write"]);
+
+/** Repo-relative path for a transcript tool `file_path` (absolute), or null if outside the
+ *  worktree / missing. */
+function relPath(filePath: unknown, worktreePath: string): string | null {
+  if (typeof filePath !== "string" || filePath === "") return null;
+  const root = worktreePath.endsWith("/") ? worktreePath : worktreePath + "/";
+  if (filePath === worktreePath) return null;
+  if (!filePath.startsWith(root)) return null;
+  return filePath.slice(root.length);
+}
+
+type Turn = {
+  text: string[];
+  think: string[];
+  edits: { tool: string; input: Record<string, unknown> }[];
+};
+
+/** Parse one JSONL line as an assistant message with its turn id + content blocks, or null. */
+function parseAssistantLine(
+  line: string,
+): { id: string; content: Array<Record<string, unknown>> } | null {
+  const t = line.trim();
+  if (t === "") return null;
+  let o: unknown;
+  try {
+    o = JSON.parse(t);
+  } catch {
+    return null;
+  }
+  const rec = o as { type?: unknown; message?: { id?: unknown; content?: unknown } };
+  if (rec.type !== "assistant" || !Array.isArray(rec.message?.content)) return null;
+  const id = typeof rec.message?.id === "string" ? rec.message.id : "";
+  if (id === "") return null;
+  return { id, content: rec.message!.content as Array<Record<string, unknown>> };
+}
+
+/** Fold a message's content blocks into its turn: text / thinking → reasoning, edit tools → edits. */
+function addBlocksToTurn(turn: Turn, content: Array<Record<string, unknown>>): void {
+  for (const b of content) {
+    if (b.type === "text" && typeof b.text === "string" && b.text.trim())
+      turn.text.push(b.text.trim());
+    else if (b.type === "thinking" && typeof b.thinking === "string" && b.thinking.trim())
+      turn.think.push(b.thinking.trim());
+    else if (b.type === "tool_use" && typeof b.name === "string" && EDIT_TOOLS.has(b.name))
+      turn.edits.push({ tool: b.name, input: (b.input as Record<string, unknown>) ?? {} });
+  }
+}
+
+/** The non-empty new-text strings a single edit tool contributes: MultiEdit → each `edits[].new_string`,
+ *  Write → `content`, Edit → `new_string`. */
+function editNewStrings(tool: string, input: Record<string, unknown>): string[] {
+  const raw =
+    tool === "MultiEdit" && Array.isArray(input.edits)
+      ? (input.edits as Array<Record<string, unknown>>).map((e) => e?.new_string)
+      : [tool === "Write" ? input.content : input.new_string];
+  return raw.filter((s): s is string => typeof s === "string" && s.trim() !== "");
+}
+
+/** The edit ops a single turn contributes: each edit tool × each of its new-text strings, tagged
+ *  with the turn's reasoning (thinking preferred over text). Edits outside the worktree are skipped. */
+function turnEdits(turn: Turn, worktreePath: string): TranscriptEdit[] {
+  const reasoning = (turn.think.join(" ") || turn.text.join(" ")).trim();
+  const ops: TranscriptEdit[] = [];
+  for (const e of turn.edits) {
+    const path = relPath(e.input.file_path, worktreePath);
+    if (path === null) continue;
+    for (const newString of editNewStrings(e.tool, e.input))
+      ops.push({ path, tool: e.tool, newString, reasoning });
+  }
+  return ops;
+}
+
+/**
+ * Parse a transcript into edit operations with their turn reasoning. Groups assistant lines by
+ * `message.id` (each content block is its own JSONL line — Phase 0), so a turn's `text` blocks
+ * (preferring `thinking` when present, though it is absent in practice) supply the reasoning for
+ * every edit in that turn. Edit → one op; MultiEdit → one op per `edits[]`; Write → the `content`.
+ * PURE (operates on the passed text).
+ */
+export function parseTranscriptEdits(text: string, worktreePath: string): TranscriptEdit[] {
+  const turns = new Map<string, Turn>();
+  const order: string[] = [];
+  for (const line of text.split("\n")) {
+    const parsed = parseAssistantLine(line);
+    if (!parsed) continue;
+    let turn = turns.get(parsed.id);
+    if (!turn) {
+      turn = { text: [], think: [], edits: [] };
+      turns.set(parsed.id, turn);
+      order.push(parsed.id);
+    }
+    addBlocksToTurn(turn, parsed.content);
+  }
+  return order.flatMap((id) => turnEdits(turns.get(id)!, worktreePath));
+}
+
+/** Build the agent-span notes: anchor each extracted edit against its diff file's new-side lines,
+ *  dropping trivial reasoning, unanchorable edits, per-line/text duplicates, and anything beyond
+ *  the per-file cap. PURE. */
+export function buildAgentNotes(edits: TranscriptEdit[], files: DiffFile[]): DiffNote[] {
+  const byPath = new Map(files.map((f) => [f.path, f]));
+  const rowsCache = new Map<string, { no: number; content: string; isAdd: boolean }[]>();
+  const perFileCount = new Map<string, number>();
+  const seen = new Set<string>();
+  const notes: DiffNote[] = [];
+  for (const e of edits) {
+    const reason = clampReason(e.reasoning);
+    if (reason.length < MIN_REASON_LEN) continue; // trivial/empty → drop
+    const file = byPath.get(e.path);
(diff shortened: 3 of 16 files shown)


scoop:
## Summary

This PR implements #1699 cleanly: it derives two ephemeral, read-only annotation sources (agent reasoning from the session transcript, anchored to changed lines; critic findings routed to per-file / panel banners) and feeds them through the already-plumbed `@pierre/diffs` wrapper. The architecture is well-factored — a pure/sync builder (`src/diff-annotations.ts`) with async IO isolated in the endpoint, a shared `attributeFinding` classifier so the scope backstop and the Diff-tab routing can't drift, XSS-safe verbatim text rendering (textContent / `{text}` interpolation, never `{@html}`), and a best-effort fetch decoupled from the 15s poll that degrades to an empty list on any failure. Tests are thorough (anchoring ambiguity, per-file cap, binary skip, late-arrival rerender gating, XSS).

Verified:
- **i18n parity** — all four new keys exist in both catalogs: `viewport_diff_annotation_agent`, `viewport_diff_annotation_review`, `feat_diff_annotations_title`, `feat_diff_annotations_body` (ui/messages/en.json:2978-2983 ↔ ui/messages/de.json:2978-2983).
- **Feature announcement** — auto-registered via `import.meta.glob` (ui/src/lib/feature-announcements.ts:47); `id: "diff-annotations"` is unique; `sinceVersion: "1.44.0"` groups with its predecessor `v1.44.0-diff-review-view.ts` (the issue said v1.45.0, but matching the shipped predecessor is the more coherent choice — not a defect).
- **Types & contracts** — `DiffFile` has `hunks`/`binary`/`truncated`/`patch` (src/types.ts:360-370); `DiffLine` exposes `kind`/`newNo`/`content` with `DiffLineKind = "add"|"del"|"ctx"` (src/types.ts:344-351); `getReview(id)?.findings` is `string[]` (src/types.ts:531); `computeDiff(worktreePath, base, branch)` signature matches (src/diff.ts:308) and the annotations endpoint reuses `resolveDiffBase` exactly like `sessionDiffRead` so it renders against the same base (src/server.ts:2064 vs 2081).
- **Routing** — `parts[3]==="diff" && parts[4]==="annotations"` correctly precedes the bare `parts[3]==="diff"` (src/server.ts:2150-2152).
- **Late-arrival effect** — traced the `hasRendered` / `lastAnnoKey` gating in PierreDiff.svelte across the render-in-flight vs annotations-arrive orderings; no missed-apply race, and the empty-array no-op is content-keyed (confirmed by the browser test at ui/src/lib/components/PierreDiff.browser.test.ts:1233).

### Finding 1 (blocking): transcript path omits `spawnAccountDir` → agent notes silently vanish for spawn-account sessions

`src/server.ts:2083` resolves the transcript with `jsonlPathFor(s.worktreePath, s.claudeSessionId)` — dropping the third `spawnAccountDir` argument. Every other transcript readback in the codebase passes it: poller.ts:105/118/130/300/680, process-reaper.ts:478, recap.ts:118, index.ts:1640/2047, usage-breakdown.ts:111, usage-snapshot.ts:52. The doc comment on `projectsDirFor` is explicit that this is mandatory: "resolution MUST follow the account, else every transcript readback (usage, activity, halt classification, MCP-auth URL) reads a nonexistent path and silently misses" (src/usage.ts:14-17). For any session with a non-null `s.spawnAccountDir`, `readTranscriptTail` will hit a path under the wrong projects root, throw, be swallowed by the `catch`, and produce **zero agent notes** — the PR's headline capability silently doesn't work for that entire class of sessions (review notes still render). `s.spawnAccountDir` is right there on the session; pass it as the third arg. (The adjacent `sessionActivityRead` at src/server.ts:2054 shares the same omission — see Out of scope — but that pre-existing gap doesn't make replicating it correct.)

### Finding 2 (nit): keyed `{#each}` by finding text can throw on duplicate findings

DiffFileStack.svelte:186 (`{#each reviewFindings.get(file.path) ?? [] as finding (finding)}`) and DiffPanel.svelte:167 (`{#each generalFindings as finding (finding)}`) key the each-block by the finding string itself. Two identical finding strings for the same file (or two identical panel-level findings) produce a duplicate key, which Svelte 5 rejects at runtime (`each_key_duplicate`). Findings are usually distinct so this is low-probability, but the critic contract does not guarantee uniqueness. Key by index (or `${i}-${finding}`) to be safe.

## Stated limitations (could not verify locally)
- `@pierre/diffs`' `DiffLineAnnotation<Meta>` type is not resolvable in this worktree (package d.ts not found), so I could not statically confirm `DiffAgentAnnotation` is assignable to it. The browser tests construct the exact `{side, lineNumber, metadata:{text,tool}}` shape against the real Pierre runtime and it slots correctly, and TS assignability is a CI concern — assumed compatible as written.

- src/server.ts:2054 (`sessionActivityRead`) and src/server.ts:522 (token count) both call `jsonlPathFor` without `spawnAccountDir`, the same latent gap as Finding 1 — pre-existing, outside the changed region of this diff.

scoop (commented):
## Re-review — #1699 Diff-tab per-line annotations

Both points from the previous revision are resolved, and the new code checks out.

### Prior findings — both addressed

1. **spawnAccountDir on the transcript path** — FIXED. `sessionDiffAnnotationsRead` now calls `jsonlPathFor(s.worktreePath, s.claudeSessionId, s.spawnAccountDir)` (src/server.ts:449-450), matching the third param of the resolver (verified against src/usage.ts:24-30, where `projectsDirFor(spawnAccountDir)` is the account-scoped root). Spawn-account sessions now resolve a real transcript path.
2. **`each` keyed by finding string (Svelte 5 `each_key_duplicate`)** — FIXED. Both loops now key by index: `{#each ... as finding, i (i)}` at ui/src/lib/components/DiffFileStack.svelte:188 and ui/src/lib/components/DiffPanel.svelte:169. Identical findings no longer throw.

### Verification performed

- **i18n parity** — the four added keys (`viewport_diff_annotation_agent`, `viewport_diff_annotation_review`, `feat_diff_annotations_title`, `feat_diff_annotations_body`) exist in BOTH catalogs (ui/messages/en.json:779-785, ui/messages/de.json:763-769); the repo maintains only en + de (verified by listing ui/messages/). All three call sites (`m.viewport_diff_annotation_review()` in DiffFileStack/DiffPanel, `m.viewport_diff_annotation_agent()` in PierreDiff.svelte:1338) resolve to present keys.
- **Feature announcement** — `v1.44.0-diff-annotations.ts` mirrors the sibling `v1.44.0-diff-review-view.ts` shape exactly (`id`/`sinceVersion`/`titleKey`/`bodyKey satisfies FeatureAnnouncement`), is auto-registered via the `import.meta.glob("./feature-announcements/entries/*.ts")` in ui/src/lib/feature-announcements.ts:47-48 (no barrel edit needed), and `sinceVersion: "1.44.0"` is the next release (package.json is 1.43.0) and consistent with the #1698 sibling. The issue text mentioned a v1.45.0 target, but the actual predecessor shipped as 1.44.0, so 1.44.0 is the correct in-repo value.
- **`attributeFinding` refactor is behavior-preserving** — the new drop rule `attributeFinding(f, files).attribution === "out-of-diff"` (src/critic-core.ts:42) is byte-identical to the old `!correspondsToChangedFile(token, files)` path: `out-of-diff` is returned only for a `": "`-separated, path-shaped, unmatched token; `unattributed`/`matched` are kept — same set as before. `matchChangedFile` reuses the exact `f === token || f.endsWith("/"+token) || baseName(f)===base` predicate, now returning the DiffFile.path. Covered by the added parity test (test/critic-core.test.ts:531).
- **Transcript sourcing** — `readTranscriptTail` (src/activity.ts:158) is synchronous fs and exists; `buildDiffNotes` wraps it in try/catch and degrades to `""` → no agent notes. The endpoint wraps everything in try/catch → `{ notes: [] }`. Route ordering is correct: `parts[3]==="diff" && parts[4]==="annotations"` precedes the bare `diff` check (src/server.ts:473-476).
- **XSS** — annotation/finding text is rendered as text nodes only (`textContent` in PierreDiff.renderAnnotation; `{finding}`/`{text}` interpolation in the Svelte banners, never `{@html}`), with browser tests asserting markup is not injected.

### Observations (non-blocking, no action required)

- Annotations are fetched on initial load + manual refresh only, NOT on the 15s poll (by design, DiffPanel.svelte). A consequence reachable today: after a poll updates the diff content, the already-fetched annotations are re-applied against the new render and their line anchors can drift or drop until the next manual refresh. This is a deliberate, documented best-effort tradeoff for a chrome feature (anchoring is conservative and drops on ambiguity), not a correctness defect — flagging only for awareness.

- The structural assignability of `DiffAgentAnnotation` to `@pierre/diffs`' `DiffLineAnnotation<Meta>`, and the `setLineAnnotations`/`rerender`/`renderAnnotation` API surface, could not be verified against the library — `@pierre/diffs` is not installed in this worktree's node_modules. The added `PierreDiff.browser.test.ts` HARD-GATE test pins those APIs at CI time; assumed as written.

scoop (commented):
## Summary

This PR fleshes out the deferred data source for the Diff-tab per-line annotations (#1699). It adds a pure, unit-tested `src/diff-annotations.ts` builder (transcript → agent spans anchored by unique multi-line signature + critic findings routed via the shared `attributeFinding` classifier), a best-effort `GET /api/sessions/:id/diff/annotations` endpoint, and wires the results through the existing `PierreDiff` / `DiffFileStack` / `DiffPanel` chain with a per-file / panel banner and an in-component XSS-safe `renderAnnotation`. The task is satisfied: data source decided + documented, mapped to the wrapper props, i18n chrome added, and a feature-announcement entry included.

The change is careful and thoroughly tested. I verified the substantive plumbing:

- **i18n parity** — all four new keys (`viewport_diff_annotation_agent`, `viewport_diff_annotation_review`, `feat_diff_annotations_title`, `feat_diff_annotations_body`) exist in BOTH catalogs (verified `ui/messages/en.json:780-783` against `ui/messages/de.json:765-768`), and each is consumed (`PierreDiff.svelte:75,111`, `DiffFileStack.svelte`, `DiffPanel.svelte`, and the announcement entry).
- **Feature announcement** — auto-registered via `import.meta.glob` (verified `ui/src/lib/feature-announcements.ts:47-49`); shape matches the sibling entry (verified against `ui/src/lib/feature-announcements/entries/v1.44.0-diff-review-view.ts`); `sinceVersion:"1.44.0"` is the next-unreleased version (root `package.json` is `1.43.0`), satisfying the `> last release` assertion documented at `ui/src/lib/feature-announcements.ts:11-16`.
- **Endpoint wiring** — `jsonlPathFor(worktreePath, claudeSessionId, spawnAccountDir)` signature confirmed (verified `src/usage.ts:24-30`); `spawnAccountDir` is a real session field (`src/types.ts:123`); `getReview(id)?.findings` returns `string[]` (verified `src/store.ts:2372-2381`, `src/types.ts:531`); `computeDiff(...)` call matches `sessionDiffRead` exactly (verified `src/server.ts:2064-2065`). Route ordering places the annotations check before the bare `diff` match (`src/server.ts:2154-2157`).
- **Path/line consistency** — the annotations endpoint uses `computeDiff` while the tab renders `toSessionDiff(computeDiff(...))`; `toSessionDiff` only drops `hunks` and preserves `path`/`patch` (verified `src/diff.ts:217-225`), so per-file banner keying and new-side `lineNumber` anchors agree with the rendered patch.
- **`attributeFinding` refactor parity** — the new classifier drops exactly the old `out-of-diff` set; `matchChangedFile` (`.find ?? null`) is boolean-equivalent to the old `.some` (verified `src/critic-core.ts:76-105`), and the added parity test (`test/critic-core.test.ts:531-540`) pins the invariant.
- **Transcript-shape assumptions** corroborated against the repo's existing JSONL conventions: `type:"assistant"` + `message.usage`/`message.content` (verified `src/usage.ts:54-71`) and `type:"tool_use"`/`.name` blocks (verified `src/activity.ts:120`).
- **XSS** — both the imperative Pierre card (`PierreDiff.svelte:75-92`, `textContent` only) and both Svelte banners (`{finding}` interpolation, not `{@html}`) render verbatim text; tests assert no element injection.

## Stated limitations (could not fully verify here — not findings)

- `@pierre/diffs` is not installed in this worktree (`node_modules` absent), so I could not confirm that `DiffAgentAnnotation` is structurally assignable to `DiffLineAnnotation<{text,tool}>` or that `setLineAnnotations`/`rerender`/`renderAnnotation` behave as assumed. The PR gates these with a `HARD GATE` browser test (`PierreDiff.browser.test.ts:1168`), and CI type-checks the generic; assumed correct as written.
- The "Phase 0" claim that Claude Code writes each content block as its own JSONL line (with a shared `message.id` per turn) cannot be proven against a live transcript in-repo; the parser tolerates the alternative (multiple blocks per line are also folded correctly by `addBlocksToTurn`).

- src/diff-annotations.ts: `readTranscriptTail` reads only the last N bytes, so on a long session early edits' reasoning is silently absent from agent notes — a documented conservative tradeoff, dormant for short sessions.

## Non-blocking nits (see findings)

Two annotation-fidelity observations that don't block the feature. Both are inherent to documented v1 design choices.

No blocking issues found.

scoop (commented):

**Prior point #1 (DiffPanel poll re-anchoring) — ADDRESSED.** `ui/src/lib/components/DiffPanel.svelte:49-56` now computes a content signature `r.files.map(fileSignature).join(",")` on every diff fetch (poll included) and, when it differs from `lastDiffSig`, synchronously clears `agentByPath` (dropping stale line anchors before the shifted diff renders) and re-fetches annotations via `loadAnnotations()`. `fileSignature` hashes the raw patch text (verified ui/src/lib/pierre-diff.ts:55-58), so any line-shifting edit changes the signature and triggers the re-fetch. The 15s poll is now `pollWhileVisible(() => load(), 15000)` (DiffPanel.svelte:113) and relies on `load()` to refresh annotations on content change — so line-anchored notes can no longer stay pinned to shifted lines. Not re-raised.

**Prior point #2 (turn reasoning duplicated on every anchored line) — ADDRESSED.** `src/diff-annotations.ts:buildAgentNotes` (lines ~226-266) introduces a `turnFileAnnotated` set keyed by `` `${turnId} ${path}` `` and `continue`s if a (turn, file) pair was already annotated, so a turn's reasoning is anchored to the FIRST line it successfully anchors on each file — one note per (turn, file) rather than the same paragraph repeated on every edited line. The set is only added after a successful anchor + push, so an edit that fails to anchor doesn't consume the turn's one slot. Not re-raised.

## Verification performed

- **Route ordering:** `src/server.ts:2154-2157` — the `diff/annotations` check (`parts[3]==="diff" && parts[4]==="annotations"`) precedes the bare `parts[3]==="diff"` fall-through, so the new endpoint is reachable. Verified against src/server.ts:2152-2160.
- **Transcript path resolution:** `src/server.ts:2085-2087` passes `s.spawnAccountDir` into `jsonlPathFor(s.worktreePath, s.claudeSessionId, s.spawnAccountDir)`; signature confirmed at src/usage.ts:24-30 and the 3-arg form matches the poller/recap/reaper callers (src/poller.ts:105, src/process-reaper.ts:478, src/recap.ts:118). Session fields exist (src/types.ts:27 `claudeSessionId`, src/types.ts:123 `spawnAccountDir`).
- **Review shape:** `deps.store.getReview(id)?.findings ?? []` — `getReview` returns `ReviewVerdict | null` (src/autopilot.ts:166) whose `findings: string[]` is confirmed at src/types.ts:531.
- **Shared classifier refactor:** `scopeFindings` now delegates to `attributeFinding` (src/critic-core.ts) with byte-identical drop semantics (drop iff `out-of-diff`; keep `matched`+`unattributed`); `matchChangedFile` returns the corresponding `DiffFile.path`, which `buildReviewNotes` uses as the per-file key — aligning with `reviewFindings.get(file.path)` in DiffFileStack.svelte:189.
- **DiffFile hunk shape:** `newSideLines` relies on `kind`/`newNo`/`content`; verified DiffLineKind = "add"|"del"|"ctx" (src/types.ts:344) and field names (src/types.ts:347-350).
- **XSS:** agent content rendered via `textContent` only in PierreDiff.svelte `renderAnnotation`; findings rendered via `{finding}` interpolation (auto-escaped) in DiffPanel/DiffFileStack. No `{@html}`.
- **i18n parity:** all four new keys (`viewport_diff_annotation_agent`, `viewport_diff_annotation_review`, `feat_diff_annotations_title`, `feat_diff_annotations_body`) present in BOTH catalogs — en.json:2993-2996 and de.json:2993-2996.
- **Feature announcement:** `v1.44.0-diff-annotations.ts` uses `sinceVersion: "1.44.0"`, which is the next unreleased version (package.json version is 1.43.0) and matches sibling v1.44.0 entries; picked up by the glob registry. titleKey/bodyKey resolve to existing keys.
- **Removed prop:** PierreDiff's former `renderAnnotation` prop is now an internal const; its only caller (DiffFileStack.svelte:203) does not pass it, so no broken consumer.

## Stated limitations (could not fully verify)

- The exact shape of `@pierre/diffs`' `DiffLineAnnotation<Meta>`, `FileDiff.setLineAnnotations`, and `FileDiff.rerender` could not be checked — the package is not installed in this review worktree. The structural-assignability claim for `DiffAgentAnnotation` (types.ts) and the pre-render/late-arrival `rerender` guard are assumed as written; this substrate was already proven by #1698.

## Non-blocking observations (informational)

- `src/diff-annotations.ts`: `signatureLines` skips blank lines while `newSideLines` retains them, so an edit whose first three non-blank lines are separated by an interior blank line will fail the consecutive match and its note is silently dropped. This is a coverage reduction consistent with the module's stated "ambiguity → DROP, never anchor wrong" policy, not a correctness bug.
- `ui/src/lib/components/DiffPanel.svelte`: on a signature-changed poll only `agentByPath` is cleared synchronously; `reviewByPath`/`generalFindings` are replaced later when `loadAnnotations()` resolves. Since review findings are file/panel-level (not line-anchored) the brief staleness is harmless.

scoop (commented):
## Summary

This PR delivers issue #1699: it sources per-line agent annotations (from the session transcript) and critic findings, maps them to `DiffNote[]`, and feeds them through the existing `PierreDiff` wrapper via a new best-effort `/diff/annotations` endpoint. The design is careful — pure/sync builder split from async IO, conservative anchoring with drop-on-ambiguity, verbatim text rendered via `textContent`/`{text}` (no `{@html}`), and a proper degrade-to-empty path. The scope requirements (data source documented, wrapper props fed, i18n, feature-announcement entry) are all met.

Verified ground truth:
- **Locale parity** — all four new keys (`viewport_diff_annotation_agent`, `viewport_diff_annotation_review`, `feat_diff_annotations_title`, `feat_diff_annotations_body`) exist in BOTH catalogs: `ui/messages/en.json:3011-3014` and `ui/messages/de.json:3011-3014`.
- **`scopeFindings` refactor is behavior-preserving** — old branch order (sep<0→keep, non-path-shaped→keep, corresponds→keep, else drop) maps 1:1 onto `attributeFinding`'s `unattributed`/`matched`→keep and `out-of-diff`→drop (`src/critic-core.ts:360-364`, `394-403`). `matchChangedFile` uses `.find()` returning the DiffFile path vs the old `.some()` boolean — same predicate (`src/critic-core.ts:415-417`).
- **Type/field shapes** — `DiffFile.hunks`/`DiffLine.newNo`/`kind` used by `newSideLines` match `src/types.ts:346-369`; `jsonlPathFor(worktree, sessionId, spawnAccountDir?)` matches `src/usage.ts:24-30`; `computeDiff(worktree, base, branch)` matches `src/diff.ts:308-312`; `store.getReview(id)?.findings` matches `ReviewVerdict.findings` (`src/types.ts:492`).
- **spawnAccountDir** — the new endpoint correctly passes `s.spawnAccountDir` (`src/server.ts:2090-2092`), consistent with the same base resolution used by `sessionDiffRead` (`src/server.ts:2072-2073`), so annotation line numbers derive from the same git diff the tab renders.
- **Route ordering** — the `parts[4]==="annotations"` check precedes the bare `parts[3]==="diff"` handler (`src/server.ts:2163-2166`), so `/diff/annotations` is not shadowed.
- **Prop removal is safe** — `renderAnnotation` was removed as a `PierreDiff` prop and made an internal const; the only consumer (`DiffFileStack.svelte:203`) never passed it.
- **Feature entry auto-registers** via `import.meta.glob("./feature-announcements/entries/*.ts")` (`ui/src/lib/feature-announcements.ts:47-48`); `v1.44.0` matches the current cycle's sibling entries (repo version is 1.43.0; #1616 epic-in-flight also uses 1.44.0). The issue's "v1.45.0" was aspirational, not binding.

### Nit (non-blocking)
The `scopeFindings` JSDoc still points readers at the old function name `correspondsToChangedFile` (`src/critic-core.ts:343`), which THIS PR renamed to `matchChangedFile`. Stale after the rename — update the reference.

## Could not verify (external dependency, not in repo)
`ui/node_modules` is absent in this worktree, so I could not confirm `@pierre/diffs`' `FileDiff.setLineAnnotations`, `FileDiff.rerender`, or the `DiffLineAnnotation<Meta>` generic against a live type definition. The wrapper's `DiffAgentAnnotation` is asserted structurally assignable to Pierre's annotation type; this is a TypeScript-compile concern CI will catch. Assumed correct as written.

- `src/diff-annotations.ts:52-63` — `newSideLines` concatenates add/ctx rows across ALL hunks into one sequence, so a multi-line signature could in principle match a run straddling a hunk boundary (rows adjacent in the array but not in the file), producing a wrong anchor. The uniqueness + "must cover an added line" guards make this very rare, and it only affects best-effort chrome; acceptable, but noting it since the ambiguity check does not catch cross-hunk pseudo-adjacency.

- `src/server.ts:2063` — `sessionActivityRead` calls `jsonlPathFor(s.worktreePath, s.claudeSessionId)` WITHOUT `spawnAccountDir`, the exact omission the new annotations endpoint documents as a bug; the activity endpoint would resolve a nonexistent path for spawn-account sessions. Pre-existing, untouched by this diff.

scoop (commented):
## Summary

This PR sources per-line Diff-tab annotations from two existing, ephemeral data sources (agent reasoning anchored to changed lines via the session transcript, and routed critic findings) and feeds them through the already-plumbed `PierreDiff` wrapper — exactly the scope the issue defines. The design is careful and conservative (unique-signature anchoring with drop-on-ambiguity, verbatim text rendered via text nodes only, best-effort endpoint that degrades to `{notes:[]}`). No blocking defects surfaced.

## Prior-review item — CONFIRMED FIXED

- `src/critic-core.ts:343` stale JSDoc reference to `correspondsToChangedFile`: the comment now reads "via `attributeFinding` → `matchChangedFile`" (verified at src/critic-core.ts:343). A tree-wide grep for `correspondsToChangedFile` across src/, ui/, test/, docs/ returns zero hits — the rename is complete and no dangling references remain.

## What I verified

- **`attributeFinding` refactor is behavior-preserving for `scopeFindings`.** Compared the new per-finding routing (src/critic-core.ts:356-364, 394-408) against the removed inline logic: sep<0 → unattributed→keep; non-path-shaped→unattributed→keep; matched→keep; out-of-diff→drop. `matchChangedFile` (src/critic-core.ts:415-419) returns `files.find(...)` — equivalent to the old `.some(...)` boolean for the keep/drop gate. The empty-`files` edge is short-circuited by the sole caller before it reaches `scopeFindings` (verified `scopeBackstop` at src/critic-core.ts:443 returns early when `files.length === 0`), so the documented "empty set classifies everything out-of-diff" caveat is never hit.
- **`DiffFile` shape consumed by diff-annotations.ts matches the server type.** `hunks`, `binary`, `truncated`, `path`, and `DiffLine.{kind,newNo,content}` all verified against src/types.ts:346-370; `content` carries no +/- marker (src/types.ts:348), consistent with the normalize-and-compare in `newSideLines`/`anchorEdit`.
- **Endpoint wiring.** `sessionDiffAnnotationsRead` uses raw `computeDiff` (structured hunks, not `toSessionDiff`) at src/server.ts:2091; `computeDiff(worktreePath, base, branch)` arg order matches src/diff.ts:308-312 and the sibling `sessionDiffRead` (src/server.ts:2074). `jsonlPathFor(worktreePath, id, spawnAccountDir)` matches the 3-arg signature at src/usage.ts:24-28. Route ordering places `diff/annotations` before the bare `diff` check (src/server.ts:2164-2166), GET-only (src/server.ts:2156).
- **Matched-finding keying is consistent end to end.** `buildReviewNotes` keys matched notes to `matchChangedFile`'s return, which is an element of `files.map(f=>f.path)` (src/diff-annotations.ts:293, 415-419); `DiffPanel` stores under `n.path` and `DiffFileStack` reads `reviewFindings.get(file.path)` — same key space, so per-file banners attach to the right file. Matched → per-file, unattributed/out-of-diff → panel (`path:""`): mutually exclusive, no double-render.
- **i18n parity.** All four new keys — `viewport_diff_annotation_agent`, `viewport_diff_annotation_review`, `feat_diff_annotations_title`, `feat_diff_annotations_body` — exist in BOTH catalogs (verified ui/messages/en.json:3010-3013 and ui/messages/de.json:3010-3013; en+de are the only catalogs under ui/messages/). Usages resolve: `m.viewport_diff_annotation_agent()` (PierreDiff.svelte), `m.viewport_diff_annotation_review()` (DiffFileStack + DiffPanel), and the announcement entry's title/body keys.
- **Feature-announcement version.** `sinceVersion: "1.44.0"` with filename `v1.44.0-diff-annotations.ts` is > the last release 1.43.0 (package.json:3) and matches the existing cohort of unreleased `v1.44.0-*` entries, so it will surface in What's-New. Entries are auto-globbed (feature-announcements.ts:47), so no index edit is required.
- **XSS surface.** Agent/critic text is rendered as text nodes only: `chip.textContent`/`body.textContent` in `renderAnnotation` (PierreDiff.svelte) and `{finding}`/`{text}` interpolation in the Svelte banners — no `{@html}` anywhere in the new markup.
- **Late-arrival annotation effect.** The content effect depends only on `signature` (the reads of `lineAnnotations` in `renderContent` occur after an `await`, so they are not tracked); the separate `$effect` on `lineAnnotations` is correctly guarded by the plain-`let` `hasRendered` and diffs via `annoKey` to skip redundant `rerender()` on a fresh `?? []`. No pre-render `rerender()` path and no full re-parse on annotation change.

- The `@pierre/diffs` package is not present in this repo's `node_modules`, so I could not confirm the external `DiffLineAnnotation<Meta>` field names (`side`, `lineNumber`, `metadata`) or the `FileDiff.setLineAnnotations`/`rerender` method signatures against a live schema. The internal `DiffAgentAnnotation` shape and all call sites are self-consistent, and #1698 established the wrapper plumbing; assumed correct as written.

- src/server.ts `sessionActivityRead` (line ~2063, unchanged by this diff) calls `jsonlPathFor(s.worktreePath, s.claudeSessionId)` WITHOUT `spawnAccountDir`, unlike the new annotations endpoint which correctly passes it — the activity endpoint may resolve a nonexistent transcript path for spawn-account sessions.

- src/diff-annotations.ts: `signatureLines` collapses out blank lines while `newSideLines` retains them, so an edit whose leading non-blank lines are separated by interior blank lines will fail the consecutive-run match and be dropped rather than anchored — a coverage gap, never a mis-anchor (safe direction), and dormant for the common no-interior-blank edit.

scoop (commented):
## Summary

This PR delivers the deferred #1699 data layer: per-line agent annotations (sourced from the session transcript, anchored to changed lines) plus routed critic findings, fed through the already-plumbed `PierreDiff` `lineAnnotations`/`renderAnnotation` substrate, with a new best-effort `GET /api/sessions/:id/diff/annotations` endpoint. It documents the data source, maps it to the wire `DiffNote`/`DiffAgentAnnotation` shapes, i18n's the chrome, and adds a feature announcement. The task is satisfied.

The design choices are conservative and well-guarded (unique multi-line signature anchoring with drop-on-ambiguity, per-(turn,file) dedup, per-file cap, verbatim text-node-only rendering for XSS safety, best-effort degradation to `{notes:[]}`), and the test coverage is strong (anchoring edge cases, XSS, late-arrival rerender, lazy-mount/noteKind independence, scope parity).

## Verification performed

- **critic-core refactor parity** — `attributeFinding` reproduces the old `scopeFindings` per-finding logic byte-for-byte: no `": "` → `unattributed`/keep; non-path-shaped → `unattributed`/keep; matched → keep; `out-of-diff` → drop. Confirmed the empty-`files` short-circuit is preserved at `src/critic-core.ts:355`, so `attributeFinding` is never reached with an empty set from `scopeFindings`. `matchChangedFile` (`src/critic-core.ts:111-115`) returns the first matching `DiffFile.path` via `find(...) ?? null`, keying `matched` findings to the real file path (verified against test `test/critic-core.test.ts:537-541`).
- **Endpoint routing** — `src/server.ts` adds the `parts[3]==="diff" && parts[4]==="annotations"` check *before* the bare `parts[3]==="diff"` check (parts derived as `api/sessions/:id/diff/annotations` → parts[2]=id, parts[3]=diff, parts[4]=annotations; cross-checked against the `parts[0]==="api"` convention at `src/server.ts:685`). Order is correct.
- **`jsonlPathFor` signature** — call `jsonlPathFor(s.worktreePath, s.claudeSessionId, s.spawnAccountDir)` matches `src/usage.ts:24-30` `(worktreePath, claudeSessionId, spawnAccountDir?)`; the `spawnAccountDir` note is accurate.
- **Transcript format assumptions** — `message.content` as a block array with `b.type==="tool_use"`/`b.name`/`b.input` matches the existing activity parser (`src/activity.ts:88,120`). Grouping by `message.id` is robust whether blocks are one-per-line or batched. (The `message.id` presence and `thinking`-absent claims are Phase-0 observations I could not re-derive from a live transcript in this repo — see limitations.)
- **i18n parity** — all four new keys (`viewport_diff_annotation_agent`, `viewport_diff_annotation_review`, `feat_diff_annotations_title`, `feat_diff_annotations_body`) exist in BOTH catalogs at `ui/messages/en.json:3018-3021` and `ui/messages/de.json:3018-3021`.
- **Feature announcement** — `id: "diff-annotations"` is globally unique; the file matches the sibling `v1.44.0-diff-review-view.ts` (the #1698 predecessor) exactly in shape and `sinceVersion: "1.44.0"`. 1.44.0 is the correct in-development cycle: latest release is 1.43.0 (`CHANGELOG.md:3`). The auto-glob loader (`ui/src/lib/feature-announcements.ts:47-48`) will pick it up. (The issue's stated "v1.45.0" target is stale; the codebase convention places it in 1.44.0 alongside its predecessor — correct.)
- **`renderAnnotation` prop removal** — the prop was dropped from `PierreDiff.svelte` in favor of a stable in-component `const`; the only consumer, `DiffFileStack.svelte:203`, never passed it. No orphaned callers.
- **Map routing keys** — `buildReviewNotes` keys `matched` to `DiffFile.path`; `DiffFileStack` reads `reviewFindings.get(file.path)` and `agentAnnotations.get(file.path)` — both keyed by `DiffFile.path`. Consistent.
- **Poll-vs-load wiring** — `pollWhileVisible(() => load(), 15000)` and `onclick={refresh}` (→`load(true)`) correctly avoid passing the event/tick arg as `forceAnnotations`; the 15s poll never re-fetches annotations unless the diff signature changes (`DiffPanel.svelte:49-56`).

## Nit (non-blocking)

- `PierreDiff.svelte` `annoKey()` (line ~117) builds the change-detection key from `side:lineNumber:text` but omits `tool`. If a refresh returned an annotation with identical side/line/text but a different `tool`, the late-arrival effect would skip `setLineAnnotations`+`rerender` and the chip would show a stale tool. This is practically unreachable (a given path+line+reasoning is produced by one deterministic edit tool), so it's a cosmetic nit, not a correctness issue — including `tool` in the key would make the guard match its "annotations already applied" contract exactly.

## Limitations (could not fully verify — not findings)

- `@pierre/diffs` is not installed in this checkout (`ui/node_modules/@pierre` absent), so I could not confirm the exact `DiffLineAnnotation<Meta>` field set that `DiffAgentAnnotation` is asserted structurally assignable to (`ui/src/lib/types.ts:1918-1926`). The new `PierreDiff.browser.test.ts` renders the `{side,lineNumber,metadata}` shape against real Pierre and asserts it slots/renders, which is good behavioral evidence; the compile-time assignability is a CI/tsc concern.
- The Phase-0 transcript claims (each content block is its own JSONL line sharing `message.id`; `thinking` absent in practice) are documented assumptions about the Claude Code transcript format; no live transcript is in-repo to re-derive them. The parser degrades safely (missing `message.id` → line skipped; short/empty reasoning → note dropped) if they don't hold.

scoop (commented):
## Summary

This PR sources the deferred per-line Diff-tab annotation data for #1699 from two existing, ephemeral sources — agent reasoning anchored to changed lines (from the session transcript) and routed critic findings — and feeds them through the already-plumbed `PierreDiff` wrapper props plus per-file / panel review banners. It satisfies every scoped deliverable: data source decided + documented (module header in `src/diff-annotations.ts`, `docs/external-task-api.md`), mapped to `DiffLineAnnotation<Meta>[]`, chrome i18n'd, and a feature-announcement entry added. The design is conservative (drop-on-ambiguity anchoring, best-effort endpoint that degrades to `{notes:[]}`), thoroughly unit- + browser-tested, and correctly guards against stored-XSS in the review surface.

- **i18n parity**: all 4 new keys present in BOTH catalogs — verified `ui/messages/en.json:862-865` and `ui/messages/de.json:847-850`; repo maintains only en+de (`ui/messages/`). No orphaned key.
- **`jsonlPathFor` + spawnAccountDir**: signature `(worktreePath, claudeSessionId, spawnAccountDir?)` at `src/usage.ts:24-30`; the new endpoint passes `s.spawnAccountDir` (`src/server.ts:2095-2096`), matching the other transcript consumers (`src/poller.ts:105`, `src/recap.ts:118`, `src/process-reaper.ts:478`). The commit history confirms this was the fix for the earlier omission.
- **Transcript IO**: `readTranscriptTail` is sync (`src/activity.ts:158`), consistent with the module's "PURE except the sync readTranscriptTail" claim.
- **Diff types**: `DiffFile.{binary,truncated?,hunks,patch?}` and `DiffLine.{kind,newNo,content}` (`src/types.ts:346-372`) match how `newSideLines`/`anchorEditLine`/`buildAgentNotes` consume them.
- **`computeDiff` shape** returns `{files: DiffFile[]}` (`src/diff.ts:308-333`); **`getReview().findings: string[]`** (`src/types.ts:533`).
- **Routing**: GET-only, `/diff/annotations` check precedes the bare `/diff` check (`src/server.ts:2155-2166`) — no shadowing.
- **Shared classifier**: `attributeFinding` extraction keeps `scopeFindings` drop behaviour byte-identical (out-of-diff ⇒ dropped), covered by the parity test at `test/critic-core.test.ts:562-571`.
- **Feature announcement**: auto-loaded via `import.meta.glob("./feature-announcements/entries/*.ts")` (`ui/src/lib/feature-announcements.ts:47-48`) — no manual registration needed; `sinceVersion:"1.44.0"` matches the sibling `v1.44.0-diff-review-view.ts` (#1698) and the active cycle (package.json `1.43.0`).
- **XSS**: `renderAnnotation` builds nodes with `textContent` only (`PierreDiff.svelte`), banners use Svelte `{finding}` auto-escaping; both gated by dedicated tests (`PierreDiff.browser.test.ts`, `DiffFileStack.browser.test.ts`).
- **Late-arrival annotations**: `hasRendered`/`lastAnnoKey` guards correctly avoid a pre-render `rerender()` and no-op on unchanged-content empty arrays; `renderContent` re-syncs `lastAnnoKey` on each render so no update is lost across the in-flight window.

## Could-not-verify (stated limitations, not findings)

- **`@pierre/diffs` type/behaviour** (`DiffLineAnnotation<Meta>`, `setLineAnnotations`, `rerender`, new-side `lineNumber` semantics, honoring of `side:"additions"`): the package is not installed in this worktree's `node_modules`, so the API shape could not be checked against a live schema. The PR gates these assumptions with real-DOM browser tests (`PierreDiff.browser.test.ts` HARD GATE / LINE-0 OBSERVED), which is the right mitigation. Assumed as written.
- **Claude Code transcript JSONL layout** (each content block its own assistant line sharing `message.id` per turn; `thinking` absent in practice): this is external runtime data not present in the repo, so the Phase-0 assumption `parseTranscriptEdits` relies on could not be independently confirmed. The parser degrades safely (unparseable/partial lines skipped, missing reasoning ⇒ note dropped).

- `src/diff-annotations.ts`: `anchorEdit` matches the signature against `newSideLines`, which concatenates all hunks' new-side rows in document order. Two separate hunks' rows are adjacent in that array though not adjacent in the file, so a signature could in principle match across a hunk boundary. In practice the signature is contiguous real new-text, the match must be UNIQUE and cover an added line, and the worst case is an informational note placed on a nearby wrong line — never trusted data. Reachable today only via an extreme coincidence; noting for the record.
- `src/diff-annotations.ts` / `src/activity.ts`: agent notes derive from `readTranscriptTail` (bounded tail), so edits from early in a long session yield no annotations, and a turn straddling the tail cutoff can lose its reasoning (empty ⇒ dropped). This is intended best-effort degradation, documented in the module header.

- `ui/src/lib/components/DiffPanel.svelte`: on a poll where the diff content changes, `load()` clears `agentByPath` immediately but `reviewByPath`/`generalFindings` are only cleared inside `loadAnnotations()`'s resolved `.then`. During the brief refetch window file-level/panel review banners persist while agent notes vanish. Harmless (review banners are not line-anchored, so they can't mis-anchor), so this is fine as-is.

No blocking problems; the implementation is correct and satisfies the task.

PR #1704, opened 2026-07-13
feat(ui): explain the activity heartbeat strip on hover

kai-osthoff:
## 🎯 Problem

A teammate looked at the **amber activity strip** on a live session card and asked in chat:

> *„Sagt ma… was genau zeigt diese fancy Leiste eigentlich an?"*

The strip carried only a screen-reader `aria-label`, so a sighted mouse user got **nothing** on hover and had to ask a human what it meant. **Goal:** let the strip answer that question in place — on hover.

## ✨ Solution

Hovering (or keyboard-focusing) the strip now opens a small **legend popover** that decodes it:

- **What it is** — the agent's tool-use over the last **8 minutes** (each cell ≈ 20 s); an empty strip means the session has gone quiet.
- **Colour key** — **Active** (amber, brighter = busier) · **Idle** (faint) · **Error** (red stub).

The swatches in the legend are rendered from the *same* cells as the bar, so the key can never drift from what it explains. On touch, a tap still just selects the row (the legend is a desktop hover/focus affordance); mobile users meet it through the What's-New drawer.

## 🔧 Technical details

- **Mirrors the shipped Stepper stage-legend** — native `popover="manual"` placed with Floating UI (`anchorPopover`), so it escapes the card's `overflow:hidden`.
- The strip becomes a real `<button>` raised above the `.unit-hit` overlay so it receives hover; **click forwards to row-select** (no dead click zone), and its accessible name now discloses that it opens the session.
- **No hit-halo** (unlike the Stepper's 3px lamps): the full-width 12px strip sits one 3px-gap row below `.hold-cta` (downward halo) and above the Stepper (upward halo), so a vertical bleed would poach their hit zones.
- Legend swatches reconstruct the strip's flex/height context so `.cell.err` shows its stub and the faint idle cell renders faithfully.
- Kept strictly to **cell encoding** — recency ("active X ago") stays with the existing clock `TimePopover`, so the two hover popovers on a running card don't overlap in content.

### Files
| File | Change |
|------|--------|
| `HeartbeatStrip.svelte` | button trigger + encoding-legend popover |
| `UnitRow.svelte` | pass `onactivate` → row select |
| `messages/en.json` · `de.json` | 7 `heartbeat_*` keys + 2 `feat_heartbeat_legend_*` keys |
| `feature-announcements/entries/v1.44.0-heartbeat-legend.ts` | What's-New entry |
| `HeartbeatStrip.browser.test.ts` | new — asserts the button/aria/legend/error contract |

### Testing
- `bun run check` — 0 errors / 0 warnings
- `bun run check:i18n` — 2 locales in parity (2996 keys)
- `HeartbeatStrip.browser.test.ts` 5/5 · `UnitRow.browser.test.ts` 55/55 (regression)
- branch-hygiene · feature-catalog · announcement-versions (1.44.0 > released 1.43.0) · glossary — all ✓

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2989,5 +2989,14 @@
   "feat_diff_review_body": "Der Diff-Tab der Sitzung nutzt jetzt eine neue Engine: Wechsle zwischen Nebeneinander- und vereinheitlichter Ansicht — mit wortgenauer Hervorhebung innerhalb der Zeile, einer Datei-Seitenleiste zur schnellen Navigation und einklappbarem unverändertem Kontext.",
   "repo_chip_open_github": "GitHub-Repo-Webseite öffnen",
   "feat_repo_chip_github_link_title": "GitHub über einen Repo-Chip öffnen",
-  "feat_repo_chip_github_link_body": "Rechtsklicke oder halte einen GitHub-Repo-Chip gedrückt, um die Repository-Seite direkt aus dem Herdenkopf zu öffnen."
+  "feat_repo_chip_github_link_body": "Rechtsklicke oder halte einen GitHub-Repo-Chip gedrückt, um die Repository-Seite direkt aus dem Herdenkopf zu öffnen.",
+  "heartbeat_pop_intro": "Die Tool-Nutzung des Agents über die letzten 8 Minuten (jede Zelle ≈ 20 s). Eine leere Leiste heißt, die Session ist verstummt.",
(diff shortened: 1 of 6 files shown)


kai-osthoff:
@scoop ready for review — all green: CI passing (verify, branch-hygiene, pr-title, CodeQL, docs-site), no conflicts (mergeable/clean), no open review threads. One thing to eyeball: the EN/DE legend + What's-New copy (authored from `heartbeat.ts`).

PR #1707, opened 2026-07-13
fix(codex): restore weekly usage gauge

kai-osthoff:
## Summary
- classify Codex rollout rate-limit windows by `window_minutes` when present, so a weekly window emitted as `primary` is shown as WK instead of 5H
- keep positional fallback for older duration-less rollout lines and ignore unknown durations with a one-time warning
- add a minimal fixture from a current rollout shape plus regression coverage for weekly-only, reordered windows, unknown durations, and same-bucket collision handling

## Validation
- `bun test test/codex-usage.test.ts`
- `bun run lint`
- `cd ui && bun test src/lib/components/usage-gauges.test.ts`
- `cd ui && bun run check`
- `cd extension && bun run check`
- `git push` pre-push hook: prettier, eslint, gates, tsc, ext, root-tests, ui, fallow audit

### DIFF
--- a/src/codex-usage.ts
+++ b/src/codex-usage.ts
@@ -230,6 +230,38 @@ function toLimitWindow(w: RolloutLimitWindow | null | undefined, now: number): L
   return { pct: Math.max(0, Math.min(100, Math.round(w.used_percent))), resetAt };
 }
 
+type WindowBucket = "session5h" | "week";
+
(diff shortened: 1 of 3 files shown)


kai-osthoff:
READY is glowing ✅🚀🔥 Scoop, do your job 😏👉

PR #1710, opened 2026-07-13
feat(herd): explain lifecycle stages with header info tooltips + empty-state overview

kai-osthoff:
## Why

New users landed on the Herd and saw a raw status board — **Reviewing**, **Your turn**, **Merging** — with no explanation of what each stage means or, more importantly, **what Shepherd is doing for them** at that stage. That undersells the whole value of Shepherd over running Claude Code / Codex standalone. This makes the board self-explanatory.

## What changed

**1. Per-header "i" tooltips (the original ask).** Every lifecycle group header now has a small, right-aligned **"i"**. Hover (desktop) or tap (touch) opens a compact tooltip that, in 2–3 sentences, says **what the stage means → what Shepherd automates → what happens next**. All 12 stages covered (Working, CI running, CI failed, Reviewing, Rework running, Waiting on reviewer/merger, Draft: awaiting sign-off, Your turn, Ready to merge, Merging, Merged).

**2. First-run overview.** A per-header icon only helps once you *have* sessions in a stage — a brand-new user on the empty board would see nothing. So `EmptyHerd` gains a **"How work flows"** overview (condensed happy-path: Working → CI → Reviewing → Your turn → Merged), reusing the same strings so the two surfaces never drift.

## Implementation notes

- **Reuses the existing `InfoTip` component** — no new primitive. Hardened it so the `i` glyph and tooltip prose render verbatim inside the uppercase `.micro` header (`text-transform:none` on both the glyph and the tooltip — otherwise the lowercase `i` rendered as `I`). Added an opt-in `prominent` resting state so the affordance is discoverable without a coachmark.
- **`HerdGroup`**: header right-side wrapped in a `.head-right` cluster (`margin-left:auto`), holding the optional bulk action **and** the `i` (rightmost). The two action buttons' own `margin-left:auto` was moved to the wrapper — right-alignment behavior is identical.
- **`Herd`**: explicit **hyphen→underscore** `groupHelp` map (group keys are hyphenated; Paraglide accessors can't be), aria-label via the shared `newtask_info_aria`.
- EN + DE copy for all 12 stages, matching existing terminology (Critic/Kritiker, Merge train/Merge-Train, Plan gate/Plan-Gate, worktree, Owed lens).

## Design decisions

The task specified only the icon + the "Reviewing" example; these were the recommended defaults, called out in the plan and executed:
- **Scope:** all 12 stages (fullest newcomer benefit).
- **"Working" (active) group:** covered via the empty-state overview; **no** permanent header added (keeps the top, most-important rows chrome-free).
- **Discoverability:** empty-state overview + a slightly more prominent resting icon. **No coachmark** — the only `<Coachmark>` host (GitRail) arms exclusively from a hardcoded `PILL_FEATURE_IDS`, so a board-header `targetId` would never fire. A board-level coachmark host was deliberately deferred as out-of-scope.

## Testing

- `bun run check:i18n` — EN/DE parity (3017 keys each).
- `bun run test` — full UI suite green (230 files, 3433 tests).
- PR-hygiene gates (branch-hygiene, feature-catalog, announcement-versions, glossary) — all pass.
- New browser tests: the `i` renders in a lifecycle header's right cluster and its glyph is not uppercased by the header; the `EmptyHerd` overview renders on an empty board.

Ships in **v1.44.0** with a What's-New entry.

_A live visual pass in the running app is still worth a quick look for spacing/color feel; behavior is covered by the browser-rendered tests above._


kai-osthoff:
@scoop 🟢💚✅ All green, mergeable, flawless 🎯✨ — smash that merge button already 😤🚀🙏👀

PR #1714, opened 2026-07-13
fix: name plan-review failure causes + fix the reviewer-spawn read-back race

kai-osthoff:
## Why

Clicking **REVIEW PLAN NOW** showed a generic *"Couldn't start the review. Please try again."* Behind it were **two** distinct defects, diagnosed against the live instance (67 hits in `shepherd.log`):

1. **Root cause — the reviewer-spawn read-back lost a race.** The CLI herdr driver resolved a just-started agent by cwd-matching a fresh `agent list` with **no retry**, *inside* the serialized start section. When the agent wasn't registered yet (a propagation lag, not a slow list), the match came back empty and threw `started agent not found for cwd …` → the review never started. The in-mutex read-back also stalled every other queued spawn.
2. **Symptom — one opaque message for three causes.** The `PlanReviewTrigger` `"error"` outcome collapsed *spawn failure*, *worktree-create failure*, and *api-key-not-configured* into a single banner, so the operator couldn't tell a transient race (retry) from a config problem (fix the key).

## What changed

**Track A — `fix(herdr)` (`c633ece1`)**
- Resolve the just-started agent by the **unique tab we created** (`tabId`), not cwd — deterministic, no cwd-collision aliasing.
- Move the read-back **out of `serializeStart`** so its retry holds no start mutex (a slow resolve no longer stalls the next spawn).
- Bounded **retry** (5 × 100 ms, injectable) for registration staleness; on exhaustion emit a distinct `[herdr] agent resolve exhausted …` warning so herdr-health decay stays visible.
- The socket driver already reads `terminal_id` from the reply and is untouched.

**Track B — `fix(plan-gate)` (`bc9d0883`)**
- Split the `"error"` outcome into `error-spawn` / `error-worktree` / `error-auth`.
- **PlanPanel** shows a cause-specific note; the compact entry points (rail, hold-row, badge menu) fire their generic failure toast for **any** `error-*` code via a shared `isPlanReviewError` type guard — fixing three entry points that would otherwise have silently stopped toasting.
- The badge menu **stays open on error** so the operator can retry in place.

## Scope note

The retry self-heals the transient window but does **not** fix the underlying herdr-health decay (stale reviewer/`unknown` panes accumulating over long uptimes). That durable fix — reap stale panes / evaluate the socket-driver default — is tracked in **#1713**, with the new exhaustion warning as its trigger signal.

## Tests / gates

- `test/herdr.test.ts`: resolve retries by tab id after an empty list; exhaustion warns + rolls back the tab; resolve is tab-keyed (no cwd aliasing); the read-back runs outside the serializer (a slow resolve doesn't block the next start).
- `test/plan-gate-service.test.ts`: the three failure injections map to `error-auth` / `error-spawn` / `error-worktree`.
- Browser tests: PlanPanel renders each cause-specific message; GitRail / UnitRow / PlanGateBadge fire the failure toast under `error-*` (PlanGateBadge also asserts the menu stays open).
- Green locally: root `bun run lint`, the full root suite (only the pre-existing node-pty `pty-bridge` native-build test and an unrelated load-flaky `SteersEditor` browser test fail — both unrelated to this diff and passing in isolation), UI `bun run check` + `bun run test` + `check:i18n`.


kai-osthoff:
🚀✨🎉 Behold, a FLAWLESS 💎 all-green PR that practically merges itself 😌🙌 — @scoop 👀🔨💤

PR #1715, opened 2026-07-13
feat(ui): show reviewer CLI + model on the in-flight plan-review button

kai-osthoff:
## What

When a **Plan Gate** review is running, the panel's **Reviewing…** button now shows *which coding CLI, model, and effort* are doing the review — e.g. `● REVIEWING · CLAUDE CODE · OPUS 4.8 · HIGH` — instead of a bare `REVIEWING…`.

It also fixes the panel header's **Review** chip reading **"unavailable"** during the **first** review (when no gate — and thus no `reviewer*` fields — exists yet).

| Before | After |
| --- | --- |
| `● REVIEWING…` | `● REVIEWING · CLAUDE CODE · OPUS 4.8 · HIGH` |

## Why it needed plumbing (not UI-only)

The reviewer's CLI+model is chosen server-side at spawn (`plan-gate.ts` `begin()`), but the `session:plangate-reviewing` event only carried `{ id, reviewing }`, and the saved gate gets its `reviewer*` fields only *after* a verdict lands. So on the **first** review there was no client-visible source. The fix carries a small `ReviewerEnv` on the `reviewing=true` signal **and** the `/api/plan-gates/inflight` bootstrap, cached client-side keyed by session id.

## Changes

**Server**
- `plan-gate.ts`: `onReviewing` carries the reviewer env on the `true` transition (`begin` + `adoptOrphans`); new `reviewingInflight()` feeds the bootstrap snapshot. `reviewingIds()` stays `string[]` for the herd/upnext consumer.
- `index.ts` / `server.ts`: emit the env; widen the `planGateCache.reviewing` interface type.

**UI**
- `reviews.svelte.ts`: cache `reviewerEnv`; the env write runs **before** the `applyReviewing` transition guard, so a redundant `reviewing=true` (bootstrap-then-WS interleave on reload) still refreshes the identity.
- `PlanPanel.svelte` / `PlanGateBadge.svelte`: render the triple **only when the provider is non-null** — an adopted-orphan `{provider:null,…}` or the pre-first-verdict bridge window falls back to plain `Reviewing…` (never `"unavailable · <model>"`). The button label **wraps (not truncates)** and the actions row wraps, so it never overflows a ~320px phone. The badge tooltip threads the env through a localized reviewing-branch template.
- `reviewer-env.ts`: shared `environmentLabel`/`providerLabel` helper (extracted from PlanPanel).
- Demo emits a concrete env so the browser preview shows the real triple.
- Adds the `v1.44.0` feature-announcement entry + EN/DE message keys.

## Testing

- Root: `bun run typecheck`, `bun run lint`, `bun test ./test/plan-gate-service.test.ts ./test/server-plan-gate.test.ts` (116 pass) — incl. new coverage that `begin()` emits the env on the reviewing signal and `reviewingInflight()` exposes it.
- UI: `bun run check` (0 errors), `bun run check:i18n` (parity), full `bun run test` (3509 pass) — incl. new PlanPanel browser cases rendering the triple and asserting the null-provider fallback, and a store test for the redundant-`true` env refresh.
- PR-hygiene gates (branch-hygiene, feature-catalog, announcement-versions, glossary) all pass.
- Verified the ~320px wrap in the demo preview (the demo now emits a concrete env).

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -1315,7 +1315,8 @@ const planGate = new PlanGateService({
   env: () => roleEnv(config.plannerCli, config.plannerModel, config.plannerEffort),
   operatorLanguage: () => config.operatorLanguage,
   onChange: (id, gate) => events.emit("session:plangate", { id, gate }),
-  onReviewing: (id, reviewing) => events.emit("session:plangate-reviewing", { id, reviewing }),
+  onReviewing: (id, reviewing, env) =>
+    events.emit("session:plangate-reviewing", { id, reviewing, env }),
   onActivity: (id, summary) => events.emit("session:plangate-activity", { id, summary }),
   cap: () => config.planReviewCyclesCap,
 });
@@ -2643,7 +2644,7 @@ const appDeps: AppDeps = {
   },
   planGateCache: {
     snapshot: () => planGate.snapshot(),
-    reviewing: () => planGate.reviewingIds(),
+    reviewing: () => planGate.reviewingInflight(),
   },
   planGate: {
     consider: (s, opts) => planGate.consider(s, opts),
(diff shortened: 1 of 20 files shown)


kai-osthoff:
🟢✨🎉 A wild READY PR appears! 😴🐌 @scoop, merge it already?? 🙏🔥🚀👀💚

Issue #1717, opened 2026-07-13
Show New Task attachments in the session scratchpad


kai-osthoff:
## Problem

Attachments added in **New Task** are staged before the session exists, then copied into `.shepherd-uploads` in the session worktree. The Files tab's Scratchpad view therefore does not show initial screenshots or other task attachments, while files uploaded later in that tab do appear there.

## Desired behavior

Treat all operator-supplied attachments as Scratchpad material: attachments selected in New Task should be visible in the session's Scratchpad from the start, and the agent's initial prompt should reference the usable attachment paths.

## Scope

- Preserve current retry, held-task, relaunch, and missing-staged-upload recovery behavior.
- Make the provider-specific behavior explicit so attachment handling stays correct for every supported agent provider.
- Add focused server and browser coverage for initial visibility in Scratchpad and attachment-path delivery to the agent.

## Out of scope

The Scratchpad/Worktree explanatory tooltip is tracked separately.

scoop:
✅ PR #1836 merged.

PR #1718, opened 2026-07-13
fix(ui): clarify Scratchpad and Worktree files

kai-osthoff:
# Problem

The Files tab offers Scratchpad and Worktree, but does not explain what each source represents. Operators can reasonably mistake one for a view of the other.

# Solution

Add a persistent, accessible `(i)` beside the source switch. Its tooltip explains that Scratchpad holds session-specific supplemental material outside the repository, while Worktree is the agent's project directory and is browse-only here.

# Technical details

- Reuses the shared, non-modal `InfoTip` interaction for hover, keyboard, and touch access.
- Adds matching English and German copy.
- Keeps the existing source-switch behavior unchanged.
- Adds a browser regression test for the labeled explanation.

The follow-up for surfacing New Task attachments in Scratchpad is tracked in #1717.

# Validation

- `cd ui && bun run check`
- `cd ui && bun run test:browser -- src/lib/components/viewport/FilesPanel.browser.test.ts`
- `cd ui && bun run test`

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3087,5 +3087,6 @@
   "command_source_codex_home": "Codex-Home",
   "command_source_system": "System",
   "feat_codex_discovery_title": "Codex-Skills in der Command-Discovery",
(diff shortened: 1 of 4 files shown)


kai-osthoff:
🚨 READY! 🥳 @scoop, your heroic review mission awaits—please grace us with your expertise. 🙏

PR #1719, opened 2026-07-13
feat(new-task): preview image attachments

kai-osthoff:
## Problem / goal

Image attachments in the New Task dialog are represented only by filenames, so operators cannot confirm that the correct screenshot or picture was attached before launching a task. This is especially awkward when several similarly named captures are attached.

## Solution / added

- Adds a compact image preview anchored to fresh image-attachment filenames.
- Opens on hover or keyboard focus on desktop.
- Toggles on filename tap on mobile, with outside tap, Escape, scroll, and resize dismissal.
- Keeps non-image and carried relaunch/held attachments unchanged.
- Revokes browser object URLs when an attachment is removed or the dialog closes.
- Adds EN/DE accessible copy and a What is New entry for v1.44.0.

## Technical details

- Uses the native HTML Popover API and the existing `anchorPopover`/Floating UI helper to escape clipping and stay within viewport bounds.
- Keeps preview data browser-local from the original `File`; no server file-serving endpoint or staged-path exposure was added.
- Uses Shepherd semantic tokens and the existing popover shadow.
- Tests cover fresh-image eligibility, mixed payload preservation, desktop/mobile interaction, all dismissal paths, seeded-image exclusion, and URL cleanup.

## Validation

- `cd ui && bun run check`
- `cd ui && bun run test:browser -- NewTask.browser.test.ts` (86 tests)
- `cd ui && bun run test` (3617 tests)
- Feature catalog, announcement version, glossary, and branch hygiene gates

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3092,5 +3092,8 @@
   "newtask_provider_constraint_title": "CLI-Verfügbarkeit",
   "newtask_provider_constraint_body": "Shepherd hat {provider} ausgewählt und inkompatible CLIs deaktiviert, solange dieser Token im Prompt steht.",
   "feat_provider_availability_card_title": "Hinweise für Single-CLI-Skills",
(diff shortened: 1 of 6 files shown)


kai-osthoff:
🚨✨ READY, @scoop—your heroic review duties await! 🫡🔍🔥🎉

PR #1720, opened 2026-07-13
feat(ui): surface single-CLI skill constraints

kai-osthoff:
## Problem / Case / Feature / Goal

When an operator picks a command, skill, or plugin token that only works in one coding CLI, Shepherd already forces the compatible provider. That constraint was only shown as a small inline hint near the picker, so it was easy to miss why the CLI changed or why the other option became unavailable.

## Solution / Added

- Replace the low-visibility provider constraint hint with a blue CLI availability card in the New Task run settings.
- Show the required provider as a badge and explain that Shepherd selected it while the constrained token remains in the prompt.
- Keep shared-provider entries quiet: no card, no disabled CLI options.
- Add EN/DE message keys and a feature announcement fragment for discovery.

## Technical Details

- Reuses the existing `ProviderTokenConstraint` flow; no API or data model changes.
- Adds token-based blue styling in `NewTaskRunSettings.svelte` with `var(--color-blue)`, `var(--fs-*)`, and the existing callout structure.
- Updates `NewTask.browser.test.ts` to assert the Codex-only skill card, provider lock, and shared-provider no-card behavior.

## Verification

- `scripts/check-branch-hygiene.sh`
- `scripts/check-announcement-versions.mjs`
- `scripts/check-feature-catalog.sh`
- `cd ui && bun run test:browser -- NewTask.browser.test.ts`
- `cd ui && bun run check`

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3027,7 +3027,7 @@
   "provider_badge_claude": "Claude",
   "provider_badge_codex": "Codex",
   "provider_badge_both": "Beide",
-  "newtask_provider_constraint_note": "{command} benötigt {provider}.",
+  "newtask_provider_constraint_note": "{command} ist nur in {provider} verfügbar.",
   "newtask_provider_constraint_error": "{command} benötigt {provider}. Entferne den Token oder wechsle die CLI.",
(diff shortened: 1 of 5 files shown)


kai-osthoff:
@scoop PR is green and ready for review again.

PR #1721, opened 2026-07-13
fix(plan-gate): resume an exited (Codex) planner before steering findings

kai-osthoff:
## Problem

Clicking **rework / re-review** on a plan (or waiting for the auto-loop) does nothing when the planner is **Codex**: a reviewer churns for a few minutes, then everything looks byte-identical. The plan is never revised.

## Root cause

The plan-gate revise loop delivered the reviewer's findings with a **bare steer that assumes a live pane** (`applyChangesRequested` → `service.reply`, which no-ops unless the session is still a live agent in `herdr agent list`).

- **Claude** idles live at its prompt after a turn (`transient-tab-reaper.ts:26`, `distiller.ts:287`) → the steer lands → it revises. ✅
- **Codex** *exits* after its turn — the whole autopilot codex path exists for this (`autopilot.sendSteer` resumes an exited pane before steering). So for a Codex planner the findings reached **no live agent**, it never revised, and re-review just re-confirmed the same text.

And the one component that resumes an exited pane before steering — `autopilot` — is **paused during the plan gate** (`autopilot.eligible()` returns null while `planPhase === "planning"`). So nothing revived the exited Codex planner.

## Fix

Give the plan gate the **same resume-before-steer** autopilot already has, via a shared helper (`src/resume-then-steer.ts`, lifted from `autopilot.sendSteer`):

- If the planning pane isn't a live agent (or needs an account re-drive), **`resume()` it first, then steer.** A revived Codex planner receives the findings, revises `.shepherd-plan.md`, and the existing settle-triggered auto re-review picks up the changed plan.
- For **Claude** the pane is already live → `resume()` is skipped → behavior is byte-identical to before.
- Auto-resume is **refused for a non-isolated Codex session** (`codex resume --last` is cwd-scoped and would resume/steer a *sibling* — the same constraint autopilot enforces). Such a planner escalates to the operator instead of corrupting a sibling.

**Session split is unchanged:** reviewer stays its own fresh throwaway session, planner stays the main session (same as Claude) — only the exited Codex planner is revived so findings land.

### Also (same subsystem)

`finalize()` now carries the **live** gate round, not the begin-time snapshot, so an operator `resume()`/`dismiss()` that resets the round *during* an in-flight review isn't clobbered back to the pre-reset value.

## Testing

- `test/plan-gate-service.test.ts` (+4): Codex-exited → resume-then-steer lands + round advances; non-isolated Codex → resume refused → escalates; Claude live → no needless resume; resume-reset race not clobbered by finalize.
- Full suite: **6889 pass, 0 fail** (the lone `pty-bridge` failure is a pre-existing node-pty native-module load error in worktrees, unrelated). `typecheck` clean, `lint` clean, `autopilot` (140) green.

Server-only; no user-facing UI. `[no-feature-entry]`

### DIFF
--- a/src/autopilot.ts
+++ b/src/autopilot.ts
@@ -7,6 +7,7 @@ import { effectiveAutopilot } from "./effective-autopilot";
 import { signedOff } from "./signoff";
 import { checksCleared } from "./checks-gate";
 import { DRAFT_PR_NOTE } from "./service";
+import { resumeThenSteer } from "./resume-then-steer";
 
 /**
  * Agent-facing steer templates. NOT UI chrome — never i18n'd (they are typed into the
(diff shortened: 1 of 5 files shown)


kai-osthoff:
@scoop 🟢🎉🚀✨ It's READY, genius 💅😏 merge it already ⏰🙄👉🔘📦

PR #1723, opened 2026-07-13
fix(ui): improve command bar result relevance

kai-osthoff:
## What changed

- rank exact, prefix, substring, and compact fuzzy matches in distinct relevance tiers
- reject noisy long-gap subsequences and prevent matches from stitching across separate fields
- rank command-bar groups by their strongest result while preserving the established blank-query ordering
- highlight whichever visible field actually produced the match

## Why

A precise query such as `new` could bury **New task** below weak session matches because results were grouped in a fixed order and the fuzzy matcher accepted widely scattered characters. The command bar now favors clean, explainable matches without changing its visual shell or empty-state behavior.

## Validation

- `CI=true bunx vitest run src/lib/fuzzy.test.ts src/lib/components/CommandBar.browser.test.ts` — 60 passed
- `CI=true bun run test` — 244 files, 3,630 tests passed
- `bun run check` — 0 errors, 0 warnings
- `scripts/check-branch-hygiene.sh` — linear off `origin/main`
- rebased onto current `origin/main` without conflicts

### DIFF
--- a/ui/src/lib/components/CommandBar.browser.test.ts
+++ b/ui/src/lib/components/CommandBar.browser.test.ts
@@ -263,6 +263,51 @@ describe("CommandBar — fuzzy matching", () => {
     // …yet the option still reads as the full, untouched title.
     await expect.element(page.getByRole("option").first()).toHaveTextContent("newer");
   });
+
+  it("does not stitch one fuzzy match across separate session fields", async () => {
+    renderBar({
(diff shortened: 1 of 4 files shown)


kai-osthoff:
🚨✨ READY, @scoop! 🎉 Apparently this PR won’t review itself. 🫡🔥

PR #1724, opened 2026-07-13
fix: codex reviewer 'agent host is busy' — confirm exec spawns by tab existence

kai-osthoff:
## Problem — "Couldn't start the reviewer — the agent host is busy"

With the reviewer role CLIs set to **codex**, clicking **Review** (plan review / PR critic) frequently — and under load, constantly — fails with *"Couldn't start the reviewer — the agent host is busy"* (`error-spawn`).

**Root cause (verified on the live box):** codex roles run as headless `codex exec`, a **non-interactive process that runs to completion and exits**. Shepherd's CLI herdr driver confirms a spawn by polling `agent list` for the freshly-created tab. But herdr drops an agent from `agent list` the instant its process exits (proven), and under load a live agent's registration lags past the ~0.5 s poll window. So `resolveStartedAgent` threw `started agent not found`, surfaced as the toast — **and the failure path then deleted the reviewer's disposable worktree, discarding a verdict the reviewer may already have written.** (Claude reviewers are long-lived/interactive, so they're always found — this is codex-specific.)

Two secondary facts confirmed the diagnosis: codex reviews *do* land verdicts when the read-back happens to win, and the alarming `reviewer_spawns` "codex 0/N completed" is a **metrics artifact** (codex has no Claude transcript → `completeReviewerSpawn` was skipped), not a real failure.

## Solution

**1. Confirm exec-style spawns by tab existence, not agent liveness** (`src/herdr.ts`)
On read-back exhaustion, `resolveStartedAgent` now decides by `tab list` instead of `agent list`:
- **tab still present** ⇒ the spawn ran (running-late or already exited) ⇒ return a started handle with an empty terminal id (a completed exec has no live terminal; `herdr.stop("")` is a safe no-op; the tab husk is reaped by the existing hourly orphan-tab sweep). It does **not** close the tab here — a running-but-unregistered agent would be killed.
- **tab gone** ⇒ genuine failure ⇒ throw.

This stops the false `error-spawn` and, critically, stops the failure path from deleting a review worktree whose reviewer already produced a verdict.

**2. Record reviewer completion even without a transcript** (`src/plan-gate.ts`)
`finalize()` and the orphan-reap helper now call `completeReviewerSpawn(usage ?? ZEROED_USAGE)` instead of gating on `if (usage)` — mirroring the existing `review.ts` pattern — so codex reviewer completions set `completedAt` (per-run token attribution stays null for codex, which has none). No more misleading `0/N`.

## Technical details

- The happy path is unchanged: interactive/claude agents register immediately and are found in the first poll attempts; only the previously-throwing read-back **miss** path changes.
- Added `HerdrDriver.tabsAsync()` (async sibling of the existing `tabs()`) and a shared `parseTabs()` helper.
- The socket driver is unaffected — it reads `terminal_id` straight from the `agent.start` reply and never needed the read-back.
- **Tests:** updated the exhaustion case (now "tab gone → throws + rolls back"), added "agent never registers but tab persists (exec exited) → returns started, no rollback", and a plan-gate "null usage (codex) still completes with zeroed totals" case. `bun test ./test/herdr.test.ts ./test/plan-gate-service.test.ts` + related herdr/plan-gate suites green; `bun run lint` clean.

### Interim mitigation (no code)
Immediate relief before merge: set the reviewer role CLIs back to **claude** in Settings (claude reviewers are interactive and confirm reliably). Task sessions can stay on codex.

### DIFF
--- a/src/herdr.ts
+++ b/src/herdr.ts
@@ -245,6 +245,18 @@ export function parseAgents(result: unknown): HerdrAgent[] {
   }));
 }
 
+/** Maps a `tab list` reply to `HerdrTab[]`. Shared by the sync `tabs()` and async
+ *  `tabsAsync()` so both parse the `result.tabs[]` shape identically. */
+export function parseTabs(parsed: unknown): HerdrTab[] {
+  const tabs = (parsed as { result?: { tabs?: Record<string, string>[] } })?.result?.tabs ?? [];
+  return tabs.map((t) => ({
+    tabId: t.tab_id ?? "",
+    label: t.label ?? "",
+    agentStatus: (t.agent_status ?? "unknown") as HerdrState,
+    workspaceId: t.workspace_id ?? "",
+  }));
+}
+
 /**
  * Maps a single herdr `AgentInfo` object (the `result.agent` of an `agent.start`
  * `agent_started` reply) to one `HerdrAgent`, using the SAME field mapping as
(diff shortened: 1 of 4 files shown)


kai-osthoff:
🟢✨ CI's greener than envy 🥦💚 — @scoop, your throne awaits 👑🪑 go review! 🚀👀🔥

PR #1730, opened 2026-07-13
feat(ui): jump to sessions from auto-merge entries

kai-osthoff:
## Problem

The Full-auto merge strip shows which task is currently merging, rebasing, or needs attention, but its entries are static. Operators have to search for the matching task manually before they can inspect or steer its terminal. This is especially confusing when a repository or status filter currently hides that task.

## Solution

Make session-backed Full-auto merge entries clickable and keyboard-accessible. Selecting an entry now clears filters that would hide the task, follows it into the correct repository, and opens the existing task and terminal view directly. Statuses without a session ID remain static rather than guessing a destination.

## Technical details

- Render session-backed strip rows as native buttons with token-based hover and focus states.
- Reuse one shared session-jump path for the auto-merge strip and command bar.
- Preserve existing status colors, sorting, responsive wrapping, and the static fallback.
- Add EN and DE announcement copy plus the v1.44.0 feature-catalog fragment.
- Add browser coverage for pointer activation, Enter and Space activation, and missing-session fallback.

## Verification

- `bun run check`
- `bun run check:i18n`
- `bun run test:browser -- QueueStrip.browser.test.ts`
- `bun run test` (239 files, 3509 tests)
- Branch hygiene, feature catalog, announcement version, glossary, and diff checks


kai-osthoff:
🚨✨ READY, @scoop—apparently the PR won’t review itself. 🫡🔍💅

PR #1734, opened 2026-07-13
fix(ui): keep epic draft review actions reachable

kai-osthoff:
## Problem

Long drafts created through **Create EPIC from research** can grow beyond the available session viewport. Only the child-issue list owned a scrollbar, while a long parent issue body expanded the whole panel. The surrounding viewport clips overflow, which made **Amend**, **Abort**, and **Approve & create** unreachable on desktop and mobile.

## Solution

- Bound the Epic Draft panel to 60% of the dynamic viewport height.
- Give parent content and child issues one shared vertical scroll area.
- Keep the draft header and review actions visible outside that scroll area.
- Remove the nested child-list scrollbar so touch and mouse users have one predictable scroll target.
- Preserve the existing empty, materializing, approved, amend, abort, and approve behavior.

## Technical details

The panel now uses three flex regions:

1. A non-shrinking header.
2. A flexible `.edp-scroll` content region with `min-height: 0`, native vertical scrolling, contained overscroll, and touch panning.
3. A non-shrinking `.edp-actions` region separated by the existing semantic hairline token.

A focused Vitest browser regression renders an oversized draft at desktop and 390 px widths. It verifies that the content overflows inside the shared scroller, the child list no longer creates a nested scrollbar, the action region remains inside the panel bounds, and the panel respects its dynamic-viewport cap.

## Validation

- `bun run test:browser`: 116 files, 1,283 tests passed
- `bun run test`: 245 files, 3,622 tests passed
- `bun run check`: 0 errors and 0 warnings
- `bun run check:i18n`: EN/DE parity, 3,096 keys per locale
- `git diff --check`: clean

### DIFF
--- a/ui/src/lib/components/EpicDraftPanel.browser.test.ts
+++ b/ui/src/lib/components/EpicDraftPanel.browser.test.ts
@@ -12,7 +12,8 @@ function longDraft(sessionId: string): EpicDraft {
       title: "Keep long epic drafts reviewable",
       body: Array.from(
         { length: 80 },
-        (_, i) => `Section ${i + 1}\nDetailed research finding that must remain readable before approval.`,
+        (_, i) =>
+          `Section ${i + 1}\nDetailed research finding that must remain readable before approval.`,
       ).join("\n\n"),
       acceptanceCriteria: ["The complete draft can be reviewed."],
       nonGoals: ["Collapsible sections"],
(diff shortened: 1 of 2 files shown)


kai-osthoff:
🚨✨ READY! ✨🚨 @scoop, miraculously, this PR won’t review itself. 🦸‍♂️🔥🚀

kai-osthoff:
🎉🚨 READY, @scoop! 🚨🎉 Apparently reviews still require reviewers. Shocking. 😱🧐🔥

PR #1742, opened 2026-07-14
fix: preserve plan reviewer launches

### LINKED ISSUE #1740: Plan reviewer launch is lost after root-pane close despite successful Herdr start
## TL;DR

Plan reviews can still fail with **“Couldn't start the reviewer — the agent host is busy”** after #1724, even though Herdr successfully started the reviewer.

The message is misleading: production logs show that Shepherd loses the reviewer tab during its own startup orchestration. Shepherd discards the agent/terminal returned by `herdr agent start`, closes the tab's root shell pane, then tries to rediscover the reviewer through `agent list`. For a fast-exiting or not-yet-listed Codex role, closing the root pane can remove the entire tab. The later `tab list` fallback therefore sees no tab and reports `error-spawn`.

There is also a durability gap: Shepherd writes the `reviewer_spawns` row only after this rediscovery succeeds. A reviewer can be genuinely running in a separate Codex session while Shepherd has no persistent ownership record and deletes its disposable worktree.

## User-visible symptom

On a plan awaiting re-review, clicking **Review plan now** produces:

> Couldn't start the reviewer — the agent host is busy. Try again in a moment.

The existing plan-review verdict remains visible as **changes requested**, and execution stays gated.

This is reproducible on TASK-621 (`dirty-shepherd`) with Codex configured as the plan-reviewer CLI. The task worktree itself is clean; the branch/task name is unrelated to the failure.

## Production evidence

TASK-621 session ID:

```text
379c0c11-66d4-4fbf-bd96-42d20e5a0b1d
```

Five consecutive attempts produced the same server-side failure:

```text
[plan-gate] spawn failed for 379c0c11-66d4-4fbf-bd96-42d20e5a0b1d:
herdr: started agent not found and tab gone for tab ...:t6Y5
herdr: started agent not found and tab gone for tab ...:t6Y6
herdr: started agent not found and tab gone for tab ...:t6Y7
herdr: started agent not found and tab gone for tab ...:t6Y8
herdr: started agent not found and tab gone for tab ...:t6Y9
```

The UI translates every `error-spawn` into “the agent host is busy” in `ui/messages/en.json`, but no evidence indicates host saturation.

The most recent completed TASK-621 review used Claude and remains persisted as `changes_requested`. Earlier Codex reviewer processes for the same task did run as separate Codex sessions for roughly 2–3 minutes, despite Shepherd treating those spawns as failed and leaving their reviewer-spawn records incomplete.

## Confirmed lifecycle race

A real Herdr 0.7.3 probe reproduced the relevant lifecycle:

1. Create a dedicated tab with its root shell pane.
2. Start a fast headless process in the tab.
3. Before closing the root shell:
   - `agent list`: process absent
   - `tab list`: tab present
4. Close the root shell pane, as `HerdrDriver.startImpl()` does.
5. After closing the root shell:
   - `agent list`: process absent
   - `tab list`: tab absent

Observed result:

```text
tab_present_before_root_close=1
tab_present_after_root_close=0
```

This is the exact state that makes the #1724 fallback classify the spawn as a genuine failure.

## Why #1724 did not fully fix it

Commit `374a328a` / #1724 changed `resolveStartedAgent()` to accept an absent agent when its tab still exists.

That fix assumes:

```text
codex exec exits or is not listed
        -> tab husk persists
        -> tab-list fallback confirms successful start
```

The real startup path also closes the initial root pane before `resolveStartedAgent()` runs:

- `src/herdr.ts::startImpl()`
  - `agent start`
  - `pane close <rootPaneId>`
  - return only `tabId`
- `src/herdr.ts::start()`
  - call `resolveStartedAgent(tabId, cwd)`

If the role pane is already absent or not retained by that point, closing the root pane destroys the tab. The new tab-existence fallback therefore cannot confirm the spawn.

The regression test in `test/herdr.test.ts` hardcodes `TAB_LIST_WITH_NEW`; it does not model the root-pane close or actual pane/tab lifecycle. It verifies the assumption rather than the production sequence.

## Separate-session architecture

The planner and reviewer are correctly designed as separate sessions and may use different providers, CLIs, models, and effort levels.

Current handoff flow:

```text
planner session
  writes .shepherd-plan.md
        |
Shepherd reads the plan text
        |
separate reviewer session + detached review worktree
  receives plan text inline in its prompt
  writes .shepherd-plan-review.json
        |
Shepherd reads verdict
  persists plan_gates state
  steers findings back to the planner session
```

The failure is not that the reviewer runs in the same session. It is that Shepherd does not durably own the separate reviewer run early enough.

## Persistence/state-machine gap

In `src/plan-gate.ts::begin()`, the order is currently:

1. Create detached review worktree.
2. Start reviewer through Herdr.
3. Rediscover/resolve the reviewer.
4. Add the run to the in-memory `inflight` map.
5. Insert `reviewer_spawns` row.

If step 3 returns a false negative:

- `herdr.start()` throws,
- the detached review worktree is immediately removed,
- no `inflight` entry exists,
- no `reviewer_spawns` row exists,
- the already-started separate reviewer session cannot be adopted after restart,
- any eventual verdict is discarded or cannot be written to the removed path.

This is a pre-persistence ownership gap, not just a UI race.

## Relevant code

- `src/herdr.ts`
  - `HerdrDriver.start()`
  - `HerdrDriver.startImpl()`
  - `HerdrDriver.resolveStartedAgent()`
  - `startAgentWithCollisionRetry()`
- `src/plan-gate.ts`
  - `PlanGateService.begin()`
  - reviewer worktree cleanup on `error-spawn`
  - `recordReviewerSpawn()` ordering
- `src/store.ts`
  - `recordReviewerSpawn()`
  - `completeReviewerSpawn()`
- `test/herdr.test.ts`
- `test/plan-gate-service.test.ts`
- `ui/src/lib/components/PlanPanel.svelte`
- `ui/messages/en.json`
- `ui/messages/de.json`

## Suggested direction

The smallest robust correction appears to be:

1. Preserve and return the agent information from the successful `herdr agent start` response instead of discarding it.
   - Herdr 0.7.3 returns `result.agent`, including `terminal_id`, `pane_id`, and `tab_id`.
   - Avoid making a second eventually-consistent `agent list` lookup the authoritative proof of launch.
2. Ensure the root-pane cleanup cannot erase the only remaining evidence/handle before the returned agent is captured.
3. Persist reviewer-run ownership before entering a state where a false-negative read-back can orphan the run.
   - A prepared/starting record may be appropriate, followed by running/completed/failed state transitions.
   - At minimum, do not delete a review worktree merely because post-start rediscovery missed an agent that `agent start` already returned.
4. Keep cleanup idempotent and restart-adoptable.
5. Improve the UI error copy so `error-spawn` does not claim host saturation without evidence.

The exact implementation should remain minimal; a new abstraction is not required if the returned `result.agent` can flow through the existing `HerdrAgent` type.

## Acceptance criteria

- [ ] A Codex plan reviewer can use a different CLI/model from the planner and starts reliably through the CLI-backed Herdr driver.
- [ ] A successful `herdr agent start` response is not converted into `error-spawn` solely because an immediate `agent list` lookup misses it.
- [ ] Closing the initial root pane cannot cause the successful reviewer launch to be forgotten.
- [ ] Reviewer ownership is durably recorded early enough for restart adoption or explicit failure accounting.
- [ ] Shepherd does not delete the reviewer worktree while a successfully started reviewer may still need to write/read its verdict.
- [ ] A lifecycle-level regression test models: tab creation, agent start, root-pane close, absent agent-list entry, and tab state.
- [ ] Existing Claude interactive reviewer behavior remains unchanged.
- [ ] Existing fast-completing Codex reviewer behavior remains supported.
- [ ] Spawn errors surface their actual known category; the UI does not always assert “agent host is busy.”
- [ ] No temporary tabs or detached reviewer worktrees leak after completion or confirmed failure.

## Temporary workaround

Configure Claude Code as the plan-reviewer CLI. TASK-621's latest successful completed review used Claude; the failure path is currently reproducible with Codex as the reviewer CLI.

kai-osthoff:
## Problem

A successful plan-reviewer launch could be reported as `error-spawn` when a fast Codex role disappeared from `agent list` before Shepherd finished startup orchestration. Shepherd discarded the authoritative `herdr agent start` response, closed the root pane, and then tried to rediscover the reviewer. If closing the root pane removed the tab, the launch was forgotten, its detached worktree was deleted, and the UI incorrectly claimed that the agent host was busy.

## Solution

- Keep and validate `result.agent` from the successful CLI `agent start` response before root-pane cleanup.
- Remove the eventually-consistent post-start `agent list` / `tab list` lookup as the source of launch truth.
- Persist the `reviewer_spawns` ownership row before asking Herdr to launch, so restart adoption can recover the run.
- Complete failed launch records explicitly and only remove the worktree after a confirmed launch failure.
- Replace the host-saturation claim with an accurate reviewer-launch failure message in English and German.

## Technical details

The CLI-backed Herdr driver now matches the socket-backed path: both parse the returned `AgentInfo` into the existing `HerdrAgent` type. The lifecycle regression test models tab creation, a successful agent response, root-pane closure removing the tab, and absent list entries; it verifies that no rediscovery call can turn that success into a failure.

Plan-gate persistence now records the reviewer session ID, worktree, provider, model, effort, and timestamp before launch. A thrown launch closes that same row with zero usage and cleans the worktree; a server restart can adopt an uncompleted row before stale-worktree GC runs. Reaper documentation and tests pin the revised ownership invariant.

## Verification

- Pre-push hook: all 7 lanes passed (Root tests, UI tests, Extension check, TypeScript, ESLint, formatting, repository gates)
- Focused Herdr / plan-gate / reaper suite: 226 passed
- PlanPanel browser suite: 29 passed
- i18n parity: English and German catalogs aligned
- Branch hygiene: linear and rebased on current `origin/main`

Closes #1740

### DIFF
--- a/src/herdr-socket-driver.ts
+++ b/src/herdr-socket-driver.ts
@@ -95,7 +95,7 @@ export class SocketHerdrDriver implements IHerdrDriver {
   /**
    * Spawn an agent over the socket, mirroring `HerdrDriver.start`'s orchestration:
    * ensure a workspace → create a dedicated tab → `agent.start` (with collision-retry) →
-   * close the leftover shell pane → resolve the `HerdrAgent`. Unlike the CLI path, the
+   * close the leftover shell pane → resolve the `HerdrAgent`. As on the CLI path, the
    * `agent.start` reply carries the started agent's full `AgentInfo` (terminal_id/tab_id/…),
    * so no post-start re-list is needed. Serialized to preserve start atomicity.
    */
--- a/src/herdr.ts
+++ b/src/herdr.ts
@@ -8,8 +8,6 @@ import type { HerdrState, SessionStatus } from "./types";
 
 const execFileAsync = promisify(execFile);
 
-const sleep = (ms: number): Promise<void> => new Promise((r) => setTimeout(r, ms));
-
 export interface HerdrAgent {
   agent: string;
   agentStatus: HerdrState;
@@ -245,9 +243,8 @@ export function parseAgents(result: unknown): HerdrAgent[] {
   }));
 }
 
-/** Maps a `tab list` reply to `HerdrTab[]`. Shared by the sync `tabs()` and async
- *  `tabsAsync()` so both parse the `result.tabs[]` shape identically. Internal — not
- *  exported (fallow flags an unused export; the parsers used cross-module are exported). */
+/** Maps a `tab list` reply to `HerdrTab[]`. Internal — not exported (fallow flags an
+ *  unused export; the parsers used cross-module are exported). */
 function parseTabs(parsed: unknown): HerdrTab[] {
   const tabs = (parsed as { result?: { tabs?: Record<string, string>[] } })?.result?.tabs ?? [];
   return tabs.map((t) => ({
@@ -261,10 +258,9 @@ function parseTabs(parsed: unknown): HerdrTab[] {
 /**
  * Maps a single herdr `AgentInfo` object (the `result.agent` of an `agent.start`
  * `agent_started` reply) to one `HerdrAgent`, using the SAME field mapping as
- * `parseAgents`' per-item map. The socket `start` (issue #1553) resolves its started
- * agent straight from this reply — `agent_started.agent` carries `terminal_id`/`tab_id`/…
- * directly (confirmed against live herdr 0.7.3), so it needs no post-start re-list, unlike
- * the CLI path whose `agent start` output exposes no terminal id.
+ * `parseAgents`' per-item map. Both start transports resolve their started agent straight
+ * from this reply — `agent_started.agent` carries `terminal_id`/`tab_id`/… directly
+ * (confirmed against live herdr 0.7.3), so no post-start re-list is needed.
  */
 export function parseAgentInfo(agent: unknown): HerdrAgent {
   const a = (agent ?? {}) as Record<string, string>;
@@ -400,11 +396,6 @@ export class HerdrDriver implements IHerdrDriver {
   constructor(
     private runner: Runner = defaultRunner,
     private asyncRunner: AsyncRunner = defaultAsyncRunner,
-    // Read-back retry knobs (issue: reviewer-spawn resolve race). `agent start` returning
-    // does not guarantee the agent is in `agent list` yet — a short registration lag. Poll
-    // a few times before giving up; injectable so tests run without real delay.
-    private resolveAttempts = 5,
-    private resolveDelayMs = 100,
   ) {}
 
   list(): HerdrAgent[] {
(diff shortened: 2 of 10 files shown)


kai-osthoff:
@scoop CI is green, the PR is conflict-free and mergeable, and there are no open review threads. Please review again.

PR #1744, opened 2026-07-14
fix(ui): sort clone-repo list alphabetically

scoop:
## What

The clone-repo picker rendered each owner's repos in the order the server returned them — `src/repos.ts` lists via `user/repos?...&sort=pushed`, i.e. most-recently-pushed first. Owners were already sorted (own account first, then orgs alphabetically), but the repos *inside* each group weren't, which makes a long list unscannable.

Now the repos within each owner group are sorted by name (`localeCompare`). The group arrays are built fresh inside the `$derived`, so the in-place sort doesn't mutate the `repos` state array.

## Test

New `ui/src/lib/components/CloneRepo.browser.test.ts` feeds a deliberately push-ordered, cross-owner list and asserts the rendered rows come out `alpha, zeta, anvil, widget`. Verified red before the fix (`expected [ 'zeta', 'alpha', 'widget', 'anvil' ]`) and green after.

`bun run check`, `bun run check:i18n`, and the new test all pass. No user-facing new capability, so no feature-catalog entry. [no-feature-entry]

### DIFF
--- a/ui/src/lib/components/CloneRepo.browser.test.ts
+++ b/ui/src/lib/components/CloneRepo.browser.test.ts
@@ -1,0 +1,49 @@
+import { describe, it, expect, vi } from "vitest";
+import { render } from "vitest-browser-svelte";
+import { page } from "vitest/browser";
+import "../../app.css";
+import CloneRepo from "./CloneRepo.svelte";
+import type { GithubRepo } from "$lib/api";
+
+const getGithubRepos =
+  vi.fn<() => Promise<{ repos: GithubRepo[]; login: string; available: true }>>();
+
+vi.mock("$lib/api", async (orig) => ({
+  ...((await orig()) as object),
+  getGithubRepos: () => getGithubRepos(),
+}));
+
+function repo(nameWithOwner: string): GithubRepo {
+  const [owner, name] = nameWithOwner.split("/");
+  return {
+    nameWithOwner,
+    owner,
+    name,
+    url: `https://github.com/${nameWithOwner}.git`,
+    isPrivate: false,
+    isFork: false,
+    isArchived: false,
+    pushedAt: null,
+    cloned: false,
+  };
+}
+
+describe("CloneRepo", () => {
+  it("lists each owner's repos alphabetically, not in the server's most-recently-pushed order", async () => {
+    // Server order is by push recency — deliberately un-alphabetical, and interleaved
+    // across owners so grouping and sorting are both exercised.
+    getGithubRepos.mockResolvedValue({
+      repos: [repo("me/zeta"), repo("acme/widget"), repo("me/alpha"), repo("acme/anvil")],
+      login: "me",
+      available: true,
+    });
+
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

The change does exactly what the task asks: inside each owner group, repos are now ordered by `name` via `localeCompare` instead of the server's push-recency order (`ui/src/lib/components/CloneRepo.svelte:78`). Owner ordering (self first, then orgs alphabetically) is untouched at `ui/src/lib/components/CloneRepo.svelte:68-74`.

## Verification

- **The claim in the new comment is true.** The backend really does return most-recently-pushed first: `src/repos.ts:345` queries `user/repos?...&sort=pushed` and `src/repos.ts:324` documents "sorted most-recently-pushed first". So the diff's comment at `ui/src/lib/components/CloneRepo.svelte:51-53` matches ground truth, and the un-alphabetical fixture order in the test is a faithful stand-in for the server.
- **The in-place `.sort()` does not mutate component state.** `byOwner` is rebuilt from scratch on every recomputation of the `$derived.by` (`ui/src/lib/components/CloneRepo.svelte:59-67`), and its arrays are populated from `repos.filter(...)` (`ui/src/lib/components/CloneRepo.svelte:56`), which returns a fresh array. Sorting `byOwner[owner]` therefore mutates only a local array, never the `repos` `$state` array at `ui/src/lib/components/CloneRepo.svelte:25`. No reactive-loop or state-mutation hazard.
- **Keying still holds.** The `{#each g.repos as r (r.nameWithOwner)}` key at `ui/src/lib/components/CloneRepo.svelte:182` is unique per repo and unaffected by reordering; the server dedupes by lowercased `nameWithOwner` (`src/repos.ts:372-374`), so keys stay unique after the sort.
- **Types line up.** The test's `repo()` factory builds every field of `GithubRepo` as declared at `ui/src/lib/api.ts:490-501` (`nameWithOwner`, `owner`, `name`, `url`, `isPrivate`, `isFork`, `isArchived`, `pushedAt`, `cloned`), and the mocked resolve shape is assignable to the real `getGithubRepos` return type at `ui/src/lib/api.ts:506-510`.
- **Test conventions match the repo.** The `vi.mock("$lib/api", async (orig) => ({ ...((await orig()) as object), ... }))` spread is the same pattern used at `ui/src/lib/components/CodexUpdateModal.browser.test.ts:7-10` (important here, since `CloneRepo.svelte:3` also imports `cloneRepo` from the same module and would otherwise be undefined), and the `vi.fn<() => Promise<...>>()` generic form matches `ui/src/lib/components/EpicDiagnoseEntry.browser.test.ts:12`.
- **Empty/edge inputs.** With zero visible repos, `owners` is empty and `groups` is `[]`, so `hasVisible` (`ui/src/lib/components/CloneRepo.svelte:82`) still drives the no-match/no-repos message at `ui/src/lib/components/CloneRepo.svelte:201` — unchanged behaviour. Single-repo groups sort trivially. No new message keys or i18n surface were introduced by this diff, so there is no locale-parity concern to check.

PR #1745, opened 2026-07-14
fix: don't report a slow-but-successful epic approve as a failure

scoop:
## The bug

Approving an epic draft showed the operator a red **`epic-draft approve failed: 502`** — with no real error text. The approve had **fully succeeded**: the parent issue, all 12 children, every native sub-issue link, and the registered epic run were all created. Only the *response* failed to get back to the browser.

## Root cause

Materializing an epic runs ~25 sequential GitHub calls (a `createIssue` per child + the parent + a native sub-issue link each + `buildEpic`), which outruns **Bun's 10s default idle timeout**. Bun severs the socket — but the handler keeps running and commits the whole epic. Reproduced:

```
[Bun.serve]: request timed out after 10 seconds. Pass `idleTimeout` to configure.
CLIENT ERROR after 12.0s: The socket connection was closed unexpectedly
HANDLER COMPLETED (side effects would have landed)
```

The dev proxy turns the dead socket into a **bodyless 502**, so `failed()` has no `{error}` to read and falls back to the bare `"<label> failed: <status>"` plumbing string. (In production there is no proxy — `fetch` rejects with a `TypeError` instead. Same event, two different thrown shapes.)

## The fix

**Server — a per-route idle budget, not a global one.** `serve()` already solved this per-request for `/usage/refresh` (`server.timeout(req, 60)`), with a comment deliberately choosing *"other endpoints unchanged"*. Rather than reverse that, `isSlowScrapeRequest` is generalized into `slowRequestTimeoutSec`:

| Route | Budget |
| --- | --- |
| `POST /api/sessions/:id/epic-draft/approve` | **255s** (Bun's ceiling) |
| `POST /api/usage/refresh` | 60s (unchanged) |
| everything else | 10s default (unchanged) |

A global `idleTimeout: 255` was rejected: it would make *every* route hold stalled sockets for 255s, and would cap `/usage/refresh` below its own explicit budget.

**Client — reconcile, don't guess.** A throw no longer means failure. `approve()` re-GETs the draft and reports what the server *actually* did: `approved` → success toast, `materializing` → in-progress toast, `draft`/`null`/refetch-failed → the real error.

The forced GET (`epicDrafts.refresh()`) is load-bearing: the WS `session:epic-draft` event only fires when the handler **finishes**, strictly *after* the connection died, and the draft→materializing transition emits no event at all — so the local store is stale-by-construction exactly when the reconcile runs.

**Message provenance.** `apiError` now returns `ApiError` carrying `serverAuthored` — true only when the message came from a response body. This is what stops the leak: a bodyless 502 is a plain `Error`, **not** a `TypeError`, so class alone can't gate it, and a status range can't either (the handler's 500 carries a genuinely informative `{error}` body — a GitHub rate-limit or permissions failure — worth showing).

## Verification

- **End-to-end probe on the real `serve()` listener**: a 13.0s approve now returns `HTTP 200` with its real body (`parent=#112`, draft `approved`), where the 10s default severed it.
- **Regression guard in the suite**: bodyless-502 reconciling to `draft` must show the generic toast, never `e.message`. The previously-planned suite would have *passed* with the bug present — it only paired the 502 shape with `approved`/`materializing`, neither of which reaches the message path.
- Root `bun run lint` + 6944 tests pass; UI `bun run check` (0 errors), `check:i18n` (parity), 3684 tests pass. The shared `apiError` return-type change broke no existing consumer.

## Note

No feature-announcement entry: this is a `fix:`, not a `feat:`. The already-created epic needs no repair — it was complete all along.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -7335,10 +7335,24 @@ export function pickTerminalBridgeKind(opts: {
     : "node-pty";
 }
 
-/** The /usage refresh handler drives a multi-second ephemeral `claude` scrape and needs a longer
- *  idle window than Bun's 10s default (see the fetch handler). */
-const isSlowScrapeRequest = (req: Request, url: URL): boolean =>
-  req.method === "POST" && url.pathname === "/api/usage/refresh";
+/** Per-route idle budget (seconds) for the handful of routes that legitimately outrun Bun's 10s
+ *  default; `null` leaves a route on that default. Applied per-request via `server.timeout` in the
+ *  fetch handler — deliberately NOT a global `idleTimeout`, which would make every route hold a
+ *  stalled socket for the same window (and would cap /usage/refresh below its own budget).
+ *
+ *  - /usage/refresh drives an ephemeral `claude` scrape (~24s worst case: boot + week wait + credits
+ *    grace); a reset connection there shows a false "Retry" on the gauge.
+ *  - epic-draft/approve materializes the epic into GitHub: one createIssue per child + the parent +
+ *    one native sub-issue link each + buildEpic — ~25 sequential round-trips for a 12-child epic,
+ *    which blows past 10s. Bun then severs the socket while the handler keeps running to completion,
+ *    so the operator saw a bogus failure for an approve that had actually succeeded. 255 is Bun's
+ *    ceiling for both `idleTimeout` and `server.timeout`. */
+export const slowRequestTimeoutSec = (req: Request, url: URL): number | null => {
+  if (req.method !== "POST") return null;
+  if (url.pathname === "/api/usage/refresh") return 60;
+  if (/^\/api\/sessions\/[^/]+\/epic-draft\/approve$/.test(url.pathname)) return 255;
+  return null;
+};
 
 export function serve(deps: AppDeps, port: number) {
   const app = makeApp(deps);
@@ -7402,10 +7416,10 @@ export function serve(deps: AppDeps, port: number) {
           ? undefined
           : new Response("upgrade failed", { status: 500 });
       }
-      // /usage refresh drives an ephemeral claude scrape (~24s worst case: boot + week wait +
-      // credits grace); lift Bun's 10s idle timeout for just this request so a slow scrape can't
-      // reset the connection into a false "Retry" on the gauge. Insurance — other endpoints unchanged.
-      if (isSlowScrapeRequest(req, url)) server.timeout(req, 60);
+      // Lift Bun's 10s idle timeout for the known-slow routes (see slowRequestTimeoutSec) so a long
+      // handler can't have its connection reset out from under it. Other endpoints unchanged.
+      const slowSec = slowRequestTimeoutSec(req, url);
+      if (slowSec !== null) server.timeout(req, slowSec);
       return app.fetch(req);
     },
     websocket: {
--- a/test/epic-draft-api.test.ts
+++ b/test/epic-draft-api.test.ts
@@ -1,7 +1,7 @@
 import { test, expect, beforeEach, afterEach } from "bun:test";
 import { mkdtempSync, mkdirSync, rmSync } from "node:fs";
 import { join } from "node:path";
-import { makeApp, type AppDeps } from "../src/server";
+import { makeApp, slowRequestTimeoutSec, type AppDeps } from "../src/server";
 import { SessionStore } from "../src/store";
 import { EventHub } from "../src/events";
 import { config } from "../src/config";
@@ -200,4 +200,22 @@ test("a failed materialize reverts the draft to 'draft' so a retry can resume", 
   const res = await approve(app, s.id);
   expect(res.status).toBe(500);
   expect(store.getEpicDraft(s.id)?.status).toBe("draft"); // reverted, not stuck at materializing
+});
+
+// The approve request itself outruns Bun's 10s default idle timeout (~25 sequential GitHub calls for
+// a 12-child epic), so `serve()` lifts it per-request. Guard the route matching: a rename here would
+// silently sever the socket mid-materialize again, and the operator would see a bogus failure for an
+// approve that actually succeeded.
+test("slowRequestTimeoutSec budgets the known-slow routes and leaves others on the default", () => {
+  const sec = (method: string, path: string) =>
+    slowRequestTimeoutSec(new Request(`http://x${path}`, { method }), new URL(`http://x${path}`));
+
+  expect(sec("POST", "/api/sessions/abc-123/epic-draft/approve")).toBe(255); // Bun's ceiling
+  expect(sec("POST", "/api/usage/refresh")).toBe(60); // pre-existing budget, unchanged
+
+  // Everything else keeps the 10s default — including the draft's other verbs and near-miss paths.
+  expect(sec("GET", "/api/sessions/abc-123/epic-draft/approve")).toBeNull();
+  expect(sec("POST", "/api/sessions/abc-123/epic-draft")).toBeNull();
+  expect(sec("POST", "/api/sessions/abc-123/epic-draft/approve/extra")).toBeNull();
+  expect(sec("POST", "/api/sessions")).toBeNull();
 });
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3124,5 +3124,6 @@
   "feat_mobile_herd_accordion_title": "Lebenszyklus-Gruppen mobil einklappen",
   "feat_mobile_herd_accordion_body": "Die Herd-Liste auf dem Smartphone hält jetzt höchstens eine benannte Lebenszyklus-Gruppe offen. „Du bist dran“ ist standardmäßig geöffnet; tippe auf eine größere Gruppenzeile, um den Bereich zu wechseln oder zu schließen, während aktive Sessions sichtbar bleiben.",
   "feat_review_job_model_title": "Sieh, welches Modell prüft",
-  "feat_review_job_model_body": "Das Review-Banner im Terminal zeigt jetzt Coding-CLI, Modell und Aufwand von Plan-Gate- und PR-Critic-Jobs, damit ihre Umgebung klar von der Task-Session unterscheidbar ist."
+  "feat_review_job_model_body": "Das Review-Banner im Terminal zeigt jetzt Coding-CLI, Modell und Aufwand von Plan-Gate- und PR-Critic-Jobs, damit ihre Umgebung klar von der Task-Session unterscheidbar ist.",
+  "epicdraft_approve_in_progress": "Epic wird noch erstellt — das dauert etwas. Es erscheint hier, sobald es fertig ist."
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -3124,5 +3124,6 @@
   "feat_mobile_herd_accordion_title": "Fold lifecycle groups on mobile",
   "feat_mobile_herd_accordion_body": "The phone herd now keeps only one named lifecycle group open at a time. Your turn opens by default; tap any larger group header to switch or close it while active sessions stay visible.",
   "feat_review_job_model_title": "See which model is reviewing",
-  "feat_review_job_model_body": "The terminal review banner now shows the coding CLI, model and effort used by both Plan Gate and PR critic jobs, so their environment is distinct from the task session."
+  "feat_review_job_model_body": "The terminal review banner now shows the coding CLI, model and effort used by both Plan Gate and PR critic jobs, so their environment is distinct from the task session.",
+  "epicdraft_approve_in_progress": "Still creating the epic — this is taking a while. It'll appear here when it lands."
 }
--- a/ui/src/lib/api.previewBlocked.test.ts
+++ b/ui/src/lib/api.previewBlocked.test.ts
@@ -1,5 +1,12 @@
 import { describe, it, expect, vi, afterEach } from "vitest";
-import { createSession, isPreviewBlocked } from "./api";
+import {
+  ApiError,
+  createSession,
+  isPreviewBlocked,
+  relaunchSession,
+  restoreSession,
+  uploadScratchpadFile,
+} from "./api";
 import type { CreateInput } from "./types";
 
 function mockFetch(status: number, body: unknown): typeof fetch {
@@ -53,5 +60,64 @@ describe("preview-origin blocked detection", () => {
     // Remapped to a translated sentence — not the raw server string.
     expect(err!.message.length).toBeGreaterThan(0);
     expect(err!.message).not.toBe("forbidden: origin host not allowed");
+  });
+});
+
+// The re-wrap sites (upload / relaunch / restore) rebuild an ApiError to attach a `code`. Each must
+// first re-throw a PreviewBlockedError untouched — re-wrapping it as a plain ApiError silently
+// strips the class, and every isPreviewBlocked() caller then misclassifies the 403 as a real failure.
+describe("re-wrap sites preserve PreviewBlockedError", () => {
+  const realFetch = globalThis.fetch;
+  afterEach(() => {
+    globalThis.fetch = realFetch;
+  });
+
+  it.each([
+    ["restoreSession", () => restoreSession("s1")],
+    ["relaunchSession", () => relaunchSession("s1", {} as never)],
+    ["uploadScratchpadFile", () => uploadScratchpadFile("s1", new File(["x"], "x.txt"))],
+  ])("%s keeps the preview-origin 403 a PreviewBlockedError", async (_label, call) => {
+    globalThis.fetch = mockFetch(403, { error: "forbidden: origin not allowed" });
+    const err = await call().then(
+      () => null,
+      (e) => e as Error,
+    );
+    expect(isPreviewBlocked(err)).toBe(true);
+  });
+});
+
+// `serverAuthored` gates whether a caller may show `message` to a human, so its definition has to
+// hold at every construction site: a body-less response must NOT masquerade as a server message.
+describe("ApiError.serverAuthored provenance", () => {
+  const realFetch = globalThis.fetch;
+  afterEach(() => {
+    globalThis.fetch = realFetch;
+  });
+
+  const restoreErr = async (status: number, body: unknown) => {
+    globalThis.fetch = mockFetch(status, body);
+    return (await restoreSession("s1").then(
+      () => null,
+      (e) => e,
+    )) as ApiError;
+  };
+
+  it("is true for a real server message", async () => {
+    const err = await restoreErr(400, { error: "no active workspace" });
+    expect(err.serverAuthored).toBe(true);
+    expect(err.message).toBe("no active workspace");
+  });
+
+  it("is false for a body-less response — the fallback is a plumbing string", async () => {
+    const err = await restoreErr(502, null);
+    expect(err.serverAuthored).toBe(false);
+    expect(err.message).toBe("restore failed: 502");
+  });
+
+  // An empty message is no message: honouring it would render a blank toast.
+  it("is false for an empty server message, and falls back", async () => {
+    const err = await restoreErr(500, { error: "" });
+    expect(err.serverAuthored).toBe(false);
+    expect(err.message).toBe("restore failed: 500");
   });
 });
--- a/ui/src/lib/api.ts
+++ b/ui/src/lib/api.ts
@@ -114,7 +114,19 @@ export function isPreviewBlocked(e: unknown): boolean {
 
 /** Build an Error from a failed response body. A `403` carrying the server's
  *  preview-origin rejection becomes a translated {@link PreviewBlockedError}; every
- *  other failure becomes a plain Error preferring the server `{error}` over `fallback`. */
+ *  other failure becomes an {@link ApiError} preferring the server `{error}` over `fallback`.
+ *
+ *  `serverAuthored` records the PROVENANCE of the message: true when it came from the response
+ *  body (or is one of the translated 403s), false when `fallback` — the bare
+ *  `"<label> failed: <status>"` plumbing string — was used because the response carried no body.
+ *  A caller that surfaces `e.message` to a human MUST gate on it: a proxy 502/504 (or any severed
+ *  upstream) yields a bodyless response, and rendering that fallback verbatim shows the operator
+ *  "epic-draft approve failed: 502" instead of a real cause. Status alone can't stand in for this —
+ *  a 500 from a handler that catches and reports (`{error: e.message}`) is genuinely informative.
+ *
+ *  An EMPTY `{error: ""}` counts as no message, not as an empty server-authored one: a handler that
+ *  forwards `e.message` verbatim (e.g. approve's 500, src/server.ts) can emit one, and honouring it
+ *  would render a BLANK toast instead of the generic failure. Hence the truthiness check. */
 function apiError(
   status: number,
   body: { error?: string } | null | undefined,
@@ -125,9 +137,9 @@ function apiError(
     return new PreviewBlockedError(m.error_preview_readonly());
   }
   if (status === 403 && body?.error === ORIGIN_HOST_BLOCK) {
-    return new Error(m.error_origin_host_not_allowed());
+    return new ApiError(status, m.error_origin_host_not_allowed(), undefined, true);
   }
-  return new Error(body?.error ?? fallback);
+  return new ApiError(status, body?.error || fallback, undefined, !!body?.error);
 }
 
 /** A failed API call that carries the HTTP `status` and the server's stable `code`
@@ -136,12 +148,22 @@ function apiError(
 export class ApiError extends Error {
   status: number;
   code?: string;
-  constructor(status: number, message: string, code?: string) {
+  /** Whether `message` came from the server (a `{error}` body) rather than the bare
+   *  `"<label> failed: <status>"` fallback. Gate on this before showing `message` to a human —
+   *  see {@link apiError}. */
+  serverAuthored: boolean;
+  constructor(status: number, message: string, code?: string, serverAuthored = false) {
     super(message);
     this.status = status;
     this.code = code;
+    this.serverAuthored = serverAuthored;
   }
 }
+
+/** Provenance of an error built by {@link apiError}, for the re-wrap sites that rebuild an
+ *  {@link ApiError} to attach a `code`. Read it off the built error rather than re-deriving it from
+ *  the body, so "server authored" has exactly one definition and cannot drift between call sites. */
+const serverAuthored = (e: Error): boolean => e instanceof ApiError && e.serverAuthored;
 
 /** Build an Error from a failed response, preferring the server's `{error}` body
  *  (e.g. "no active workspace") over the bare status code so the UI shows the real
@@ -688,7 +710,9 @@ export async function uploadScratchpadFile(
   const r = await fetch(`/api/sessions/${id}/scratchpad/upload${q}`, { method: "POST", body: fd });
   if (!r.ok) {
     const body = (await r.json().catch(() => null)) as { error?: string } | null;
-    throw new ApiError(r.status, body?.error ?? `upload failed: ${r.status}`);
+    // No `code` to attach here, so `base` (already an ApiError, or a PreviewBlockedError) is exactly
+    // what a re-wrap would produce.
+    throw apiError(r.status, body, `upload failed: ${r.status}`);
   }
   return (await r.json()).path as string;
 }
@@ -899,7 +923,7 @@ export async function relaunchSession(
   // the rest of the API; for every other failure carry status + code to the caller.
   const base = apiError(r.status, body, `relaunch failed: ${r.status}`);
   if (isPreviewBlocked(base)) throw base;
-  throw new ApiError(r.status, base.message, body?.code);
+  throw new ApiError(r.status, base.message, body?.code, serverAuthored(base));
 }
 
 export type HandoffMode = "resume" | "summarize";
@@ -962,7 +986,8 @@ export async function restoreSession(id: string): Promise<Session> {
   if (r.ok) return r.json();
   const body = (await r.json().catch(() => null)) as { error?: string; code?: string } | null;
   const base = apiError(r.status, body, `restore failed: ${r.status}`);
-  throw new ApiError(r.status, base.message, body?.code);
+  if (isPreviewBlocked(base)) throw base; // else the 403 loses its PreviewBlockedError class
+  throw new ApiError(r.status, base.message, body?.code, serverAuthored(base));
 }
 
 /**
--- a/ui/src/lib/components/EpicDraftModal.svelte
+++ b/ui/src/lib/components/EpicDraftModal.svelte
@@ -2,7 +2,8 @@
   import type { EpicDraftChild } from "$lib/types";
   import { dialog } from "$lib/a11yDialog";
   import { epicDrafts } from "$lib/epic-draft.svelte";
-  import { approveEpicDraft, replySession, archiveSession } from "$lib/api";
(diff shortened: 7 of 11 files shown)


scoop (commented):
## Verdict

The change does what the task asks: it kills the bogus "epic-draft approve failed: 502" and gives the operator a real answer. Two halves, both verified:

**Server (`src/server.ts:7346-7355`)** — the approve route now gets a 255s per-request idle budget instead of Bun's 10s default. I confirmed the regex `^\/api\/sessions\/[^/]+\/epic-draft\/approve$` matches the route the server actually serves: `handleEpicDraft` claims `parts[0]==="api" && parts[1]==="sessions" && parts[3]==="epic-draft"` and dispatches `POST` + `parts[4]==="approve" && !parts[5]` (src/server.ts:3489-3500), so path shape, method, and the no-trailing-segment constraint all agree. The handler really is long-running — `materializeEpicDraft` + per-child `createIssue` + parent + `buildEpic` (src/server.ts:3451-3477) — so the 10s idle window was the plausible root cause of a severed socket. `server.timeout(req, slowSec)` sits *after* `checkAuth`/`checkOrigin` (src/server.ts:7372-7376), so the elevated budget is not reachable by an unauthenticated client — no slowloris surface widened here.

**Client (`ui/src/lib/components/EpicDraftPanel.svelte:52-75`)** — a thrown approve now triggers a re-GET and reports the server's actual state. The discriminator is message *provenance* (`ApiError.serverAuthored`), not error class or status range, which is the right axis: a bodyless proxy 502 produces a plain `Error`-shaped `ApiError` whose message is the `"<label> failed: <status>"` fallback from `failed()` (ui/src/lib/api.ts:163-165), so a "not a TypeError" gate would have leaked exactly the string this PR exists to kill. Verified `epicDrafts.refresh` bypasses the `load()` cache/in-flight guard (ui/src/lib/epic-draft.svelte.ts:32-36, contrast ui/src/lib/epic-draft.svelte.ts:41), which is necessary — the store holds the stale pre-approve row at reconcile time.

### Verification notes

- **Status values**: `"approved" | "materializing" | "draft"` — verified against ui/src/lib/types.ts:46 and src/types.ts:829; the reconcile branches cover all three. `parentNumber: number | null` (ui/src/lib/types.ts:56), and the null-guard on the success branch matches.
- **Panel state after reconcile**: `refresh()` upserts, so a `materializing` result flips the panel out of `awaiting` (EpicDraftPanel.svelte:37) and hides the approve button (EpicDraftPanel.svelte:212) — no double-submit into the server's 409 CAS guard (src/server.ts:3457).
- **i18n parity**: `epicdraft_approve_in_progress` exists in both catalogs — ui/messages/en.json:2979 and ui/messages/de.json:2979. Those are the only two catalogs in the repo (`ui/messages/*.json`). The message takes no params, matching the paramless call at EpicDraftPanel.svelte:61.
- **`Error` → `ApiError` widening in `apiError()`** (ui/src/lib/api.ts:136-138): I grepped every `instanceof ApiError` consumer (ui/src/routes/+page.svelte:1979, 1981, 2194, 2195, 2230; ui/src/lib/components/viewport/FilesPanel.svelte:187). All of them branch on `.code`/`.status`, and every one of those errors was already an `ApiError` before this PR (relaunch/restore/upload construct it explicitly), so nothing changes behaviour for them. `PreviewBlockedError` is still not an `ApiError`, which is why the panel needs the extra `isPreviewBlocked(e)` disjunct — correct as written.
- **Server-authored 500**: the approve handler's catch returns `{error: e.message}` (src/server.ts:3486), so a genuine materialize failure surfaces its real cause through the `serverAuthored` branch rather than the generic toast. This is the informative path the fix preserves.

- `ui/src/lib/api.ts` `getJson` throws a bare `Error` rather than an `ApiError`, so GET failures carry no status/provenance; harmless here (the reconcile only needs "couldn't find out"), but it means `serverAuthored` is not a universal property of API errors.

scoop (commented):
## Verdict

The PR is coherent and addresses the reported task ("approve failed with no proper error message"). All four points from the previous revision are genuinely fixed. Remaining items are non-blocking nits.

## Prior findings — re-verified

1. **`restoreSession` missing the preview-blocked re-throw** — FIXED. `ui/src/lib/api.ts:989` now has `if (isPreviewBlocked(base)) throw base;` before the re-wrap, mirroring `relaunchSession` (`ui/src/lib/api.ts:925`) and the newly-added guard in `uploadScratchpadFile` (`ui/src/lib/api.ts:714`). All three sites now behave identically; covered by the new `it.each` in `ui/src/lib/api.previewBlocked.test.ts:71-84`.
2. **`serverAuthored` re-derived at each call site** — FIXED. The single definition now lives on the error (`ApiError.serverAuthored`, `ui/src/lib/api.ts:153`), set once in `apiError` (`ui/src/lib/api.ts:140,142`), and the re-wrap sites read it back off the built error via the `serverAuthored(base)` helper (`ui/src/lib/api.ts:166`, used at `:715`, `:926`, `:990`).
3. **Empty `{error: ""}` treated as server-authored** — FIXED. `ui/src/lib/api.ts:142` is now `body?.error || fallback` with `!!body?.error` as the provenance flag, so a 500 forwarding an empty `e.message` (`src/server.ts:3486` — `json({ error: e instanceof Error ? e.message : "epic materialize failed" }, 500)`) falls back to the plumbing string and the panel renders the generic toast rather than a blank one (`ui/src/lib/components/EpicDraftPanel.svelte:68-74`).
4. **`refresh()` keeping a stale row on a `null` GET** — FIXED. `ui/src/lib/epic-draft.svelte.ts:32-39` now calls the new `remove()` on `null`. I confirmed `null` really is authoritative and not a masked error: `getEpicDraft` goes through `getJson` (`ui/src/lib/api.ts:2145` → `:174-181`), which *throws* on any non-2xx, and the GET route only returns a `null` body when the session exists but has no draft (`src/server.ts:3496,3499`).

## Verification of the new behaviour

- **Server route budget.** `slowRequestTimeoutSec`'s regex `^\/api\/sessions\/[^/]+\/epic-draft\/approve$` (`src/server.ts:7353`) matches the path the router actually owns — `parts[3] === "epic-draft" && parts[4] === "approve" && !parts[5]`, POST only (`src/server.ts:3493,3500-3501`) — and the client POSTs exactly that path (`ui/src/lib/api.ts:2154`). It is applied per-request at `src/server.ts:7420-7421`; I confirmed `Bun.serve` here sets no global `idleTimeout` (`src/server.ts:7366-7369`), so every other route stays on the 10s default as the comment claims.
- **Reconcile discriminator.** `EpicDraftPanel.svelte:68-69` gates the human-visible message on `isPreviewBlocked(e) || (e instanceof ApiError && e.serverAuthored)`. Since `failed()` (`ui/src/lib/api.ts:171-173`) routes every approve failure through `apiError`, a bodyless 502 yields `serverAuthored === false` and a `fetch` drop yields a `TypeError` — neither reaches the toast. Verified.
- **`materializing` → panel state.** The reconcile relies on the WS event to flip the panel; `session:epic-draft` is emitted on both success and the revert (`src/server.ts:3479,3485`) and is applied to the same store the panel reads (`ui/src/lib/store.svelte.ts:788-789` → `epicDraftsStore.upsert`). The approve button only renders while `awaiting` (`status === "draft"`, `EpicDraftPanel.svelte:37,212`), so a reconciled `materializing` correctly hides it — no double-approve.
- **i18n parity.** `epicdraft_approve_in_progress` exists in both catalogs the repo maintains — `ui/messages/en.json:2979` and `ui/messages/de.json:2979`; `ls ui/messages/` shows only `en.json` and `de.json`.
- **No consumer regressions from `apiError` now returning `ApiError`.** I grepped every `instanceof ApiError` consumer: `ui/src/lib/components/viewport/FilesPanel.svelte:187` (413 only), `ui/src/routes/+page.svelte:1979,1981,2194-2195,2230-2236` (all branch on `e.code`, which the re-wrap sites still populate from `body?.code`). The two re-throw changes (upload/restore now throwing `PreviewBlockedError` instead of a re-wrapped `ApiError`) land on the existing `else` fallbacks in those callers (`FilesPanel.svelte:189`, `+page.svelte:2236`), so behaviour is unchanged.

## Nits (see findings)

1. The reconcile's success toast (`EpicDraftPanel.svelte:56`) omits the `epicdraft-approve-${sessionId}` key that its two siblings use (`:61`, `:71`). A prior failure toast is `sticky: true`, which per `ui/src/lib/toasts.svelte.ts:110-111,119` never auto-dismisses and is only cleared by a keyed refresh or an explicit close (`:288-297`). So on a retry cycle (genuine failure → sticky "Couldn't create the epic." → retry succeeds) the stale failure stays pinned next to "Epic created". Keying the success toast the same way makes it supersede in place.
2. The `materializing` branch (`EpicDraftPanel.svelte:59-63`) tells the operator "It'll appear here when it lands" on a 4s non-sticky toast, but if the handler then *fails*, the server reverts to `draft` and emits `session:epic-draft` (`src/server.ts:3484-3485`), which the client only upserts into the store (`ui/src/lib/store.svelte.ts:788-789`) — no toast. The operator gets no failure message at all in exactly the scenario this PR exists to fix. Worth a follow-up (e.g. a sticky in-progress toast that a later `draft`-status event replaces with the failure).
3. `uploadScratchpadFile`'s re-wrap (`ui/src/lib/api.ts:713-715`) is now a no-op copy: `base` is already an `ApiError` carrying the same `status`, `message` and `serverAuthored`, and no `code` is added. `throw base;` says the same thing.

## Stated limitations

- Bun's documented 255s ceiling for `server.timeout`/`idleTimeout` is asserted by the comment at `src/server.ts:7350`; the runtime's own limit is not in this repo, so I did not verify it against ground truth (the value is at worst a clamp, not a correctness hazard).

scoop:
## What the PR does

Three layers: (1) `slowRequestTimeoutSec` lifts Bun's 10s idle timeout to 255s for `POST /api/sessions/:id/epic-draft/approve` (src/server.ts:7345-7354, applied at src/server.ts:7421-7422); (2) `ApiError.serverAuthored` records message provenance so a bodyless proxy 502 can never be rendered verbatim (ui/src/lib/api.ts:142,148-161); (3) `EpicDraftPanel.approve()` no longer trusts the throw — it re-GETs the draft (`epicDrafts.refresh`, ui/src/lib/epic-draft.svelte.ts:32-40) and reports what the server actually did.

## Prior findings — all three addressed

1. **Success toast key** — fixed. All four approve outcomes now go through `approveSucceeded`/`approveFailed`, which share `approveToastKey` = `epicdraft-approve-${sessionId}` (ui/src/lib/components/EpicDraftPanel.svelte:51-61, used at :75, :82, :92, :103, :108, :118). Verified against the keyed-refresh branch in ui/src/lib/toasts.svelte.ts:123-142, which replaces the existing toast in place and re-arms its timer, so a sticky failure is superseded rather than stacked beside the retry's success.
2. **Silent failure after `materializing`** — fixed. The in-progress toast is now `sticky: true` (EpicDraftPanel.svelte:82) and the `$effect` at :100-111 resolves it once the WS `session:epic-draft` event lands: `approved` → success, back to `draft` → generic failure. That matches the server, which reverts on error and emits the event (src/server.ts:3484-3485) — and `epicdraft_status_materializing`/`draft` are the only terminal states the store writes (ui/src/lib/store.svelte.ts upsert path).
3. **Redundant re-wrap in `uploadScratchpadFile`** — fixed: it now just `throw apiError(...)` (ui/src/lib/api.ts:713-715).

## Verification

- **i18n parity**: `epicdraft_approve_in_progress` exists in both catalogs — ui/messages/en.json:2979 and ui/messages/de.json:2979 — and the repo maintains only `en`/`de` (`ls ui/messages`). No placeholders, and the component calls it argument-free (EpicDraftPanel.svelte:82).
- **Route regex**: `/^\/api\/sessions\/[^/]+\/epic-draft\/approve$/` (src/server.ts:7353) matches the real handler's dispatch — `parts[3] === "epic-draft" && parts[4] === "approve" && !parts[5]`, POST (src/server.ts:3500-3501). 255 is Bun's ceiling for `server.timeout`.
- **`null` refetch**: the GET returns `json(deps.store.getEpicDraft(id))` and the store returns `null` for a missing draft (src/store.ts:4499), so `refresh()`'s `else this.remove(sessionId)` branch (epic-draft.svelte.ts:38) is fed a genuine `null`, not `undefined` — the eviction is correct.
- **`serverAuthored` gating**: `apiError` sets it from `!!body?.error` (api.ts:142), so an empty `{error:""}` from approve's 500 (`{error: e instanceof Error ? e.message : ...}`, src/server.ts:3486) falls back rather than rendering a blank toast. The 403 host-block message is a translated string and is correctly marked authored (api.ts:140).
- **Consumers of the widened `ApiError`**: `apiError` now returns `ApiError` where it used to return a plain `Error`. Checked every `instanceof ApiError` site — ui/src/routes/+page.svelte:1979,1981,2194,2230 all branch on `e.code`, which stays `undefined` on these errors, and ui/src/lib/components/viewport/FilesPanel.svelte:187 branches on `status === 413`. No behaviour change. The new `if (isPreviewBlocked(base)) throw base` in `restoreSession` (api.ts:989) makes the preview-origin 403 skip the `instanceof ApiError` branch at +page.svelte:2230 and land in the `else` — both paths call `fail(m.restore_failed())`, so no regression.

## The blocking problem

`awaitingMaterialize` is component-local state that is never scoped to, or reset by, the session it was armed for. `EpicDraftPanel` is *not* remounted per session: Viewport is rendered without a `{#key}` (ui/src/routes/+page.svelte:2712, :2878) and explicitly handles `session.id` changing in place (e.g. its per-session usage effect resets state at Viewport.svelte:711-713); `EpicDraftPanel` itself re-derives everything from the `sessionId` prop (EpicDraftPanel.svelte:29-35). So after a lost approve on session A leaves `awaitingMaterialize = true`, switching to session B re-runs the `$effect` at :100 with B's `status`/`draft` — see findings.

- ui/src/lib/components/EpicDraftPanel.svelte: the panel is inside `{#if !headerFolded}` (ui/src/lib/components/Viewport.svelte:2788-2800), so folding the header on mobile while `awaitingMaterialize` is set destroys the pending resolver — the sticky "Still creating the epic" toast then never resolves into success or failure. Only reachable via the mobile fold today; a store-level (rather than component-level) pending-approve flag would make both this and the session-switch case fall out for free.

- ui/src/lib/api.ts `getJson` throws a bare `Error("<label> failed: <status>")` with no `ApiError`/status, so `epicDrafts.refresh()` cannot distinguish a 404 from a 500 — the reconcile treats both as "couldn't find out".

scoop:
## Prior findings — all three are addressed

1. **`awaitingMaterialize` unscoped** → now `let awaitingFor = $state<string | null>(null)` (ui/src/lib/components/EpicDraftPanel.svelte:71), set to `sessionId` in the reconcile (:91) and gated in the `$effect` with `if (awaitingFor !== sessionId) return` (:109). Covered by the cross-session test at ui/src/lib/components/EpicDraftPanel.browser.test.ts:501-534. Addressed (with a residual gap, see finding 3 below).
2. **`refresh()` had no ordering guard** → `refresh()` now snapshots a write counter and bails out rather than clobbering (ui/src/lib/epic-draft.svelte.ts:43-45), with tests at ui/src/lib/epic-draft.svelte.test.ts:724-738. Addressed in principle; the counter's granularity is wrong (finding 2).
3. **`const draft` shadowing the `$derived`** → renamed to `fresh` (ui/src/lib/components/EpicDraftPanel.svelte:80). Addressed.

## What I verified

- **Route regex vs. the real route.** `slowRequestTimeoutSec`'s `^\/api\/sessions\/[^/]+\/epic-draft\/approve$` (src/server.ts:28 of the diff / src/server.ts:7351) matches the handler's own dispatch: `parts[1] === "sessions" && parts[3] === "epic-draft"` + `req.method === "POST" && parts[4] === "approve" && !parts[5]` (src/server.ts:3493, 3500-3501), and the client posts exactly `/api/sessions/${encodeURIComponent(sessionId)}/epic-draft/approve` (ui/src/lib/api.ts:2154). The old `isSlowScrapeRequest` has no remaining callers (only use site was the fetch handler, now src/server.ts:7421-7422).
- **Locale parity.** `epicdraft_approve_in_progress` exists in both catalogs: ui/messages/en.json:2979 and ui/messages/de.json:2979. No other new keys in the diff.
- **The reconcile's server contract.** The failure branch relies on the server reverting `materializing → draft` and emitting the event — confirmed at src/server.ts:3484-3486, and the success path at src/server.ts:3478-3479. The WS event reaches the store via `case "session:epic-draft": epicDraftsStore.upsert(ev.data)` (ui/src/lib/store.svelte.ts:788-789), so the resolving `$effect` does have a source of truth.
- **`serverAuthored` provenance.** `approveEpicDraft` throws through `failed()` → `apiError()` (ui/src/lib/api.ts:2158, 172-174, 130-143), so a bodyless 502 yields `serverAuthored === false` and the panel's gate (EpicDraftPanel.svelte:98-100) correctly suppresses `"epic-draft approve failed: 502"`. The 500 path carries `{error: e.message}` (src/server.ts:3486) → `serverAuthored === true` → surfaced. Consistent.
- **`apiError` now returns `ApiError` instead of `Error` — consumers still agree.** `uploadScratchpadFile` switched to `apiError(...)` (ui/src/lib/api.ts:710-718); its only production consumer branches on `e instanceof ApiError && e.status === 413` (ui/src/lib/components/viewport/FilesPanel.svelte:187), which still holds, and the existing expectations at ui/src/lib/api.uploadScratchpadFile.test.ts:43-73 still describe the produced shape. The one behaviour change is that a preview-origin 403 now throws `PreviewBlockedError` (not `ApiError`) from upload/restore — `onBringBack` treats a non-`ApiError` as `m.restore_failed()` (ui/src/routes/+page.svelte:2229-2239), the same copy it produced before, so no regression.
- **Toast supersede semantics.** A repeated `info` with the same key refreshes the existing toast in place rather than stacking (ui/src/lib/toasts.svelte.ts:120-135), so the single `approveToastKey` does what the comment at EpicDraftPanel.svelte:48-51 claims.

## New findings

### 1. `approve()` reads `sessionId` after the await — a session switch during the ~25s materialize arms the WRONG session (blocking)

`sessionId` is a `$props()` prop (EpicDraftPanel.svelte:15-25), so under Svelte 5 runes (ui/package.json:36, `svelte ^5.56.4`) every read returns the *current* value, not the value at call time. `approve()` (EpicDraftPanel.svelte:121-132) awaits `approveEpicDraft(sessionId)` — a request the PR itself documents as ~25 sequential GitHub round-trips (src/server.ts:7345-7348) — and the panel is reused across sessions with no `{#key}`, exactly as the new comment at EpicDraftPanel.svelte:66-70 says.

So: approve on A → operator switches to B while it runs (prop becomes B) → the approve throws → `reconcileFailedApprove` is called. Inside, `epicDrafts.refresh(sessionId)` (:80) now refreshes **B**, and `awaitingFor = sessionId` (:91) arms **B**, and `approveToastKey` (:51) is now B's key. If B happens to be materializing, B gets A's in-progress toast; if the refresh of B returns `draft`, the generic sticky "Couldn't create the epic." fires on B for an approve B never attempted — the precise harm the previous revision's finding 1 described, now reached through async prop capture instead of a boolean flag. A's real outcome is never reconciled at all.

Fix: capture the session once at the top of `approve()` (`const sid = sessionId`) and thread it through `reconcileFailedApprove(e, sid)`, `approveSucceeded`/`approveFailed` (keying on `epicdraft-approve-${sid}`), and `awaitingFor = sid`. The existing cross-session test only switches *after* the reconcile has settled (EpicDraftPanel.browser.test.ts:520), so it does not cover this.

### 2. `epic-draft.svelte.ts`: the ordering guard's counter is global, not per-session

`writes` is bumped by every `upsert`/`remove` for **any** session (ui/src/lib/epic-draft.svelte.ts:22, 58), while `refresh()` compares it for **one** session (:43-45). Any unrelated write landing while the reconcile's GET is in flight — a `session:epic-draft` event from another epic-authoring session (ui/src/lib/store.svelte.ts:788-789, emitted on every draft PUT at src/server.ts:3411), or another panel's `load()` (:71) — trips `this.writes !== before`. `refresh()` then discards the fresh server row and returns the **stale cached** row for this session. In the reconcile that stale row is the pre-approve `draft`, so `reconcileFailedApprove` skips the `materializing` branch and pins a sticky "Couldn't create the epic." (EpicDraftPanel.svelte:100) for an approve that is in fact still running and will land — with no `awaitingFor` armed to correct it. That is the bug this PR exists to kill, reintroduced in a narrower race. Key the counter per session (`Map<string, number>`), or compare `this.map[sessionId]` identity before/after the GET.

### 3. `EpicDraftPanel.svelte`: a lost approve is only resolved while its own session is on screen

The resolving `$effect` returns early unless `awaitingFor === sessionId` (:109) and reads the *rendered* session's `status`/`draft` (:110-113). If the operator navigates away from A after the in-progress toast appears, A's terminal WS event lands with the effect inert: the sticky "Still creating the epic…" toast stays pinned and, in the failure case, the operator is never told the materialize failed — silent failure again, just deferred until they happen to reselect A. Since `awaitingFor` already carries the session, resolve against `epicDrafts.get(awaitingFor)` and key the toast on `awaitingFor` rather than gating on the currently-rendered `sessionId`; the effect then works regardless of which session is displayed.

## Notes / limitations

- 255 as Bun's ceiling for `server.timeout` is asserted in the comment (src/server.ts:7348-7349) and in the test (test/epic-draft-api.test.ts:75); I could not verify it against Bun's source, which is not in this repo. Assumed as written. Note that it does not fail closed either way — a too-large value would be clamped or rejected by Bun, not silently applied.

scoop:
## Re-review of the previous revision's findings

**1. `approve()` read the reactive `sessionId` after the await — FIXED.** `approve()` now captures `const sid = sessionId` before the request (ui/src/lib/components/EpicDraftPanel.svelte:133) and threads it through `approveEpicDraft(sid)` (:136), `approveSucceeded(sid, …)` (:137) and `reconcileFailedApprove(sid, e)` (:139); the toast key is derived from the passed `sid` (`approveToastKey`, :58) rather than the prop. Covered by ui/src/lib/components/EpicDraftPanel.browser.test.ts:319-347.

**2. Global `writes` counter in the store — FIXED.** I read ui/src/lib/epic-draft.svelte.ts in full: the counter is gone, and `refresh()` now compares this session's row identity across the GET (`const before = this.map[sessionId]` at :43, `if (this.map[sessionId] !== before)` at :46). `upsert` replaces the row object on every write (:18), so an unrelated session's event no longer trips the guard. Covered by ui/src/lib/epic-draft.svelte.test.ts:72-82.

**3. Resolving `$effect` gated on the rendered session — FIXED.** The effect now resolves against `epicDrafts.get(awaitingFor)` and keys the toast on `awaitingFor`, not on the rendered `sessionId` (ui/src/lib/components/EpicDraftPanel.svelte:114-129). Covered by ui/src/lib/components/EpicDraftPanel.browser.test.ts:352-378 ("resolves a lost approve for session A while session B is on screen").

- Route regex: `/^\/api\/sessions\/[^/]+\/epic-draft\/approve$/` (src/server.ts:7353) matches the route actually registered — `req.method === "POST" && parts[4] === "approve" && !parts[5]` under `parts[3] === "epic-draft"` (src/server.ts:3493, 3500) — and the client posts exactly `/api/sessions/${encodeURIComponent(id)}/epic-draft/approve` (ui/src/lib/api.ts:2154). The removed `isSlowScrapeRequest` has no remaining references (grepped the tree).
- i18n parity: `epicdraft_approve_in_progress` exists in both catalogs the repo maintains — ui/messages/en.json:2979 and ui/messages/de.json:2979 (`ui/messages/` contains only en.json and de.json).
- `serverAuthored` provenance: `approveEpicDraft` throws via `failed()` → `apiError()`, which now returns an `ApiError` with `serverAuthored = !!body?.error` (ui/src/lib/api.ts:142), so the panel's gate at EpicDraftPanel.svelte:99-101 correctly suppresses the bodyless `"epic-draft approve failed: 502"` fallback and the `TypeError` of a dropped socket, while surfacing the handler's `{error: e.message}` 500 (src/server.ts:3486). The empty-string case is handled by the `||` (api.ts:142).
- Contract changes to `uploadScratchpadFile` (now may throw `PreviewBlockedError` instead of `ApiError` on the preview 403) and `restoreSession` (now preserves `PreviewBlockedError`): the only consumers are ui/src/lib/components/viewport/FilesPanel.svelte:187 (branches solely on `e instanceof ApiError && e.status === 413` → still "failed", unchanged UX) and ui/src/routes/+page.svelte:2229-2239 (a non-`ApiError` falls into `fail(m.restore_failed())`, the same message the translated preview `ApiError` produced before). No consumer regresses.
- The store's `materializing` row is only ever reached through `refresh()`, since the server emits no event on the `draft → materializing` CAS (src/server.ts:3459 has no `events.emit`), matching the comment at epic-draft.svelte.ts:27-30.

## Remaining problem

The armed-approve state is still component-local and single-slot. `awaitingFor` is a `$state` inside the panel (EpicDraftPanel.svelte:72) and the resolving effect lives with it (:114). But the panel is unmounted, not merely re-propped, in two paths that already execute today:

- ui/src/lib/components/Viewport.svelte:2788 wraps `<EpicDraftPanel>` in `{#if !headerFolded}` — folding the header on the compact/mobile layout (`headerFolded = compact && headerCollapsed`, Viewport.svelte:446) destroys it.
- ui/src/routes/+page.svelte:2877 renders `<Viewport>` under `{:else if selected}` — deselecting the session destroys the whole subtree.

Either one drops `awaitingFor` and the effect while the sticky in-progress toast survives (it is owned by the module-level `toasts` store, and `sticky: true` arms no timer — ui/src/lib/toasts.svelte.ts:110, 165-168). The operator is then left with a permanently pinned "Still creating the epic…" and no report of the terminal WS event — success or failure — which is the exact silent-outcome failure this PR exists to eliminate. The same single `awaitingFor` slot also means a second lost approve (session B) overwrites the arm for session A, stranding A's toast the same way. Both disappear if the arm lives next to the drafts in the module store (e.g. a `Set<string>` of awaiting sessions plus one resolving effect at app level, or a store-side subscription) instead of in a component that comes and goes.

Everything else in the diff checks out.

scoop (commented):
## What the PR does, and what I verified

The approve outcome is now owned by a module-level singleton (`ui/src/lib/epic-approve.ts`) instead of component-local state, plus a per-request Bun idle-timeout lift for the approve route and error-provenance tracking on `ApiError`.

**Previous revision's finding — resolved, and out of scope now.** The prior finding named `ui/src/lib/components/EpicDraftPanel.svelte`; that file is not in `git diff a9240775...HEAD`, so per the scope rule it is dropped rather than re-raised. For the record the substance *is* addressed: the panel no longer has an approve path at all (`grep approve ui/src/lib/components/EpicDraftPanel.svelte` matches only comments/labels — the button moved to `EpicDraftModal.svelte` in the preceding PR), and the arm now lives in the module singleton `epic-approve.ts:45-62` keyed by an `awaiting: Set<string>` (`epic-approve.ts:35`), so it survives the modal's unmount and tracks multiple concurrent approves — exactly what was asked for.

**Verified against ground truth:**
- Route regex `^\/api\/sessions\/[^/]+\/epic-draft\/approve$` (src/server.ts:7353) matches the real route: `handleEpicDraft` dispatches on `parts[0..3] === api/sessions/:id/epic-draft` with `parts[4] === "approve"` (src/server.ts:3493-3499), POST-only — consistent with the `req.method !== "POST"` guard (src/server.ts:7351).
- Status literals: `"draft" | "materializing" | "approved"` — verified against src/types.ts:829 and ui/src/lib/types.ts:46. The reconcile branches (epic-approve.ts:71/75) and the watcher (epic-approve.ts:52/55) only use those three.
- The reverted-on-failure contract the watcher relies on is real: src/server.ts:3484-3485 calls `revertEpicDraftToDraft` then emits `session:epic-draft`, and store.ts:4595 does the `materializing → draft` UPDATE. Success emits the `approved` row at src/server.ts:3478-3479.
- WS wiring: `session:epic-draft` → `epicDraftsStore.upsert(ev.data)` (ui/src/lib/store.svelte.ts:788-789), and `upsert` now fans out to `listeners` (ui/src/lib/epic-draft.svelte.ts:29-32), so the watcher is genuinely reachable from the WS path.
- i18n parity: `epicdraft_approve_in_progress` exists in **both** catalogs — ui/messages/en.json:3128 and ui/messages/de.json:3128. `epicdraft_approve_success` / `epicdraft_approve_failed` are pre-existing keys still used.
- Toast keying: reusing one key per session works as claimed — a keyed refresh rewrites text/`alert`/`durationMs` and re-arms via `#armInfo` (ui/src/lib/toasts.svelte.ts:123-141), and `#armInfo(id, null)` deletes the arm so a sticky toast draws no bar (toasts.svelte.ts:165-168). So the success toast correctly *replaces* the sticky in-progress/failure toast in place rather than stacking beside it.
- `ApiError` consumers still agree after the widened construction: the only `instanceof ApiError` branches are `e.status === 413` (ui/src/lib/components/viewport/FilesPanel.svelte:187) and `e.code === ...` (ui/src/routes/+page.svelte:1988-1990, 2203-2204, 2239-2245). `uploadScratchpadFile` now routes through `apiError` (api.ts:713) so a preview-403 becomes a `PreviewBlockedError` rather than an `ApiError`; FilesPanel's `isTooLarge` is false either way and the upload lands in the same `"failed"` state, so no consumer breaks. The added `isPreviewBlocked` re-throw in `restoreSession` (api.ts:989) restores class parity with `relaunchSession` (api.ts:925); `+page.svelte:2239`'s `else` branch already produced `restore_failed()` for a classless 403, so behaviour is unchanged there.
- `body?.error || fallback` with `serverAuthored: !!body?.error` (api.ts:142) is internally consistent with the doc comment: `{error: ""}` → fallback message, provenance false.

## Residual holes (both in `epic-approve.ts`)

The reconcile only ever resolves through a **live** WS upsert, and there is no backstop:
1. If `epicDrafts.refresh()` itself rejects (epic-approve.ts:69 — the same severed network that killed the approve response also kills the GET), the code falls through to the *confirmed-failure* branch (epic-approve.ts:84-89) and never calls `watch()`. The handler keeps materializing and usually succeeds, the `approved` WS event arrives, and nothing corrects the sticky "approve failed" toast.
2. Even on the happy reconcile path, `awaiting` is drained only by a WS event that the client must be connected to receive. The `/events` socket subscribes live with **no snapshot or replay on open** (src/server.ts:7426-7431), and `resync()` re-pulls sessions/queues/epics but **not** epic drafts (ui/src/routes/+page.svelte:833-892) — and `epicDrafts.load()` short-circuits on any cached row (ui/src/lib/epic-draft.svelte.ts:80). So if the socket is down while a ~25s materialize finishes, the terminal event is gone for good: `awaiting` stays armed and the sticky "Still creating the epic…" toast is pinned forever with no outcome ever reported — the same silent-outcome shape this PR set out to close, just moved into a narrower window.

Neither is a regression against `main` (today the operator gets a bogus failure instead), so I'm not blocking on them — but a re-refresh of `awaiting` sessions when a replacement socket opens (the `epoch > 1` hook at +page.svelte:921 is the natural place) would close both.

- ui/src/lib/epic-approve.ts:52 — an `approved` upsert with `parentNumber == null` matches neither branch, so the session stays in `awaiting` with its sticky toast pinned. Unreachable today (`setEpicDraftApproved` always writes a parent number, src/server.ts:3478), but it is a silent dead-end if that ever loosens; the `draft` branch has no such precondition.
- src/server.ts:7353 — 255s is Bun's ceiling, so a materialize slower than that (a very large epic, a rate-limited forge) re-enters the exact severed-socket failure this PR fixes; the client-side reconcile then carries it, which is fine, but worth knowing the server-side lift has a hard cap.

- ui/src/lib/store.svelte.ts:789 forwards the WS payload straight into `upsert(ev.data)`; a `null` payload would throw on `d.sessionId`. The server only emits post-write rows, so it can't fire today.

PR #1746, opened 2026-07-14
fix(ui): inline epic run-settings labels

scoop:
The `CLI` dropdown in the EPIC drain panel sat visibly below the `Pause` / `Auto` / `Stop epic` buttons. Cause: `.mini-field` was a column flex, so the uppercase caption stacked *above* its `<select>` — the field ran two lines tall against one-line buttons, and `.run-settings` then bottom-aligned (`align-items: end`) the taller field, dropping the select off the control row.

Each run-settings field is now a single-line chip — the caption is an inline prefix, `CLI [inherit ▾]`.

## Changes (CSS only, scoped to `EpicPanel.svelte`)

- `.mini-field` → `row` + `align-items: baseline`. Baseline, not centre: the caption belongs *on the select's text line*, not centred against a 24px (40px mobile) box.
- `.run-settings` → `align-items: center`. Outer axis centres whole controls against the buttons; inner axis aligns text. Deliberately different.
- The `min-width: 96px` floor moves off `.mini-field` onto `select`, and `.micro` gets `flex: none; white-space: nowrap` — so a tight row **wraps** (both containers already declare `flex-wrap: wrap`) instead of crushing the control.
- Phone-width `select` → `min-height: 40px`, in the same `@media (max-width: 768px)` block that already raises `.gbtn`. Without it, flattening the field leaves short selects beside tall buttons.

Applies to all three fields (`CLI`, `Model`, `Effort`) — the latter two appear once a provider is picked, so inlining only `CLI` would leave a mixed-height row.

## Deliberately unchanged

`new-task/NewTaskRunSettings.svelte` and `new-task/ModelCliPicker.svelte` keep their caption-above-control layout. They are **form** fields (label over input is correct there); EpicPanel's are **chips in a horizontal control row** beside buttons. The divergence is intentional — not an oversight to harmonize later.

## Touch targets

Phone-width `.gbtn` **and** the inline `select`s are both `min-height: 44px`, meeting the house tap-target floor. They move together — raising only the select would make it taller than the buttons beside it and reintroduce the misalignment this PR fixes. (An earlier revision shipped 40px, matching the old `.gbtn` precedent; that waiver is now cleared.) The media condition mirrors the global mobile control branch in `app.css` — `(max-width: 768px), (max-height: 600px)` — so phone landscape, which is short-*wide*, gets the same heights instead of 16px text in desktop-height controls.

## Verification

- `cd ui && bun run test` — 3675 passed. `EpicPanel.browser.test.ts` resolves the selects via `getByLabelText`, so it's the automated proof the `<label>`→`<select>` association survived the layout change.
- Rendered the panel in real Chromium at four states and inspected each: default, provider-selected (Model + Effort visible), narrow/wrapped, and phone width — buttons and selects share one line and one height; the narrow row wraps cleanly with no crushed control.

No new strings (no i18n change) and no feature-catalog entry — this is a `fix`, not a `feat`. [no-feature-entry]

### DIFF
--- a/ui/src/lib/components/EpicPanel.svelte
+++ b/ui/src/lib/components/EpicPanel.svelte
@@ -442,19 +442,23 @@
 
   .run-settings {
     display: flex;
-    align-items: end;
+    align-items: center;
     gap: 6px;
     flex-wrap: wrap;
   }
 
+  /* Caption reads as an inline prefix on the control's own line — baseline, not
+     centre, so it sits on the select's text rather than mid-box. */
   .mini-field {
     display: flex;
-    flex-direction: column;
-    gap: 2px;
-    min-width: 96px;
+    flex-direction: row;
+    align-items: baseline;
+    gap: 4px;
   }
 
   .micro {
+    flex: none;
+    white-space: nowrap;
     color: var(--color-faint);
     font-size: var(--fs-micro);
     letter-spacing: 0.08em;
@@ -463,6 +467,7 @@
 
   select {
     min-height: 24px;
+    min-width: 96px;
     max-width: 180px;
     background: var(--color-inset);
     border: 1px solid var(--color-line);
@@ -530,6 +535,12 @@
     .gbtn {
       min-height: 40px;
       padding: 2px 14px;
+    }
+
+    /* Matches .gbtn above — the fields sit inline with the buttons, so an
+       unraised 24px select would read as a short control beside a tall one. */
+    select {
+      min-height: 40px;
     }
   }
 </style>
--- a/ui/src/lib/components/EpicPanel.svelte
+++ b/ui/src/lib/components/EpicPanel.svelte
@@ -535,15 +535,15 @@
      landscape is short-wide, so a width-only query would leave these controls
      desktop-sized while app.css had already bumped their font to 16px. */
   @media (max-width: 768px), (max-height: 600px) {
+    /* 44px is the tap-target floor — buttons and the inline selects beside them
+       move together, else the taller control breaks the shared control line. */
     .gbtn {
-      min-height: 40px;
+      min-height: 44px;
       padding: 2px 14px;
     }
 
-    /* Matches .gbtn above — the fields sit inline with the buttons, so an
-       unraised 24px select would read as a short control beside a tall one. */
     select {
-      min-height: 40px;
+      min-height: 44px;
     }
   }
 </style>
--- a/ui/src/lib/components/EpicPanel.svelte
+++ b/ui/src/lib/components/EpicPanel.svelte
@@ -531,7 +531,10 @@
     color: var(--color-amber);
   }
 
-  @media (max-width: 768px) {
+  /* Condition mirrors the global mobile control branch in app.css — phone
+     landscape is short-wide, so a width-only query would leave these controls
+     desktop-sized while app.css had already bumped their font to 16px. */
+  @media (max-width: 768px), (max-height: 600px) {
     .gbtn {
       min-height: 40px;
       padding: 2px 14px;


scoop (commented):
## Summary

The change does what the task asks: `.mini-field` becomes a `flex-direction: row` with `align-items: baseline`, so each caption ("CLI", "Model", "Effort") reads as an inline prefix on the same line as its select rather than a stacked label. The `min-width: 96px` correctly moves off the field wrapper onto the `select` itself so the control keeps its width once the wrapper is no longer the sizing box.

## Verification

- **Scope of the class**: `.mini-field` is used only at `ui/src/lib/components/EpicPanel.svelte:247`, `:258`, `:268`, and defined at `:452`. Grepping `ui/src` and the test dirs for `mini-field` / `run-settings` returns no other component, test, or e2e selector, so the restyle has no out-of-file consumers to break.
- **`min-width` relocation is safe**: the new `select { min-width: 96px }` (`EpicPanel.svelte:470`) sits below the existing `max-width: 180px` (`:471`), so the two do not conflict, and `.run-settings` retains `flex-wrap: wrap` (`:447`), so each label+select pair wraps as a unit on narrow viewports rather than the caption orphaning from its control.
- **i18n parity**: no message keys change in this diff, but the labels the layout now inlines resolve in both catalogs — `epic_provider_label` = `"CLI"` at `ui/messages/en.json:1211` and `ui/messages/de.json:1211`; `epic_model_label` / `epic_effort_label` at `en.json:1213-1214` and `de.json:1213-1214`. German "Modell" / "Aufwand" are longer than the English strings, but `.micro { white-space: nowrap; flex: none }` (`EpicPanel.svelte:460-461`) plus the wrapper's `flex-wrap` keeps them intact.
- **Mobile select height**: the new `select { min-height: 40px }` (`EpicPanel.svelte:543`) matches this component's own `.gbtn { min-height: 40px }` (`:536`), which is the correct reference since the selects sit inline with those buttons. Other components use 44px (`ui/src/lib/components/NewProject.svelte:445`, `ui/src/lib/components/Settings.svelte:1830`), but matching the adjacent buttons in this row is the right choice.

I found no correctness, security, or accessibility defect. The `<label class="mini-field">` wrapper still associates the caption with its control, so the switch from column to row does not affect the accessible name.

## Nit

The component's mobile branch is `@media (max-width: 768px)` (`EpicPanel.svelte:534`), but the global control-sizing branch in `ui/src/app.css:366` is `(max-width: 768px), (max-height: 600px)` — it deliberately also covers short-wide viewports (phone landscape / split-screen) and forces selects to `max(16px, var(--fs-lg))`. So in that second case the select's font grows but the new 40px floor never applies. Practical impact is small: at 16px text plus padding and border the select already exceeds the 24px desktop floor on its own, and the pre-existing `.gbtn` rule has exactly the same gap. Worth aligning the two queries only if you want the row to stay uniform in landscape.

scoop (commented):
Style-only change to `ui/src/lib/components/EpicPanel.svelte` (no markup/logic touched), so there are no new identifiers, message keys, or signatures to resolve — the i18n keys used by the run-settings block (`m.epic_provider_label()`, `m.epic_model_label()`, `m.epic_effort_label()` at EpicPanel.svelte:248, 259, 269) are pre-existing and unmodified by this diff.

## Previous revision's point

**1. Mobile media-query condition didn't match app.css — ADDRESSED.** The media query is now `@media (max-width: 768px), (max-height: 600px)` at `ui/src/lib/components/EpicPanel.svelte:537`, which is the identical condition to the global mobile control branch at `ui/src/app.css:366` (verified by opening both; the same condition is also what `ui/src/routes/+page.svelte:1113` uses for its `MediaQuery` mobile flag). Both `.gbtn` (538-541) and the new `select { min-height: 40px }` (545-547) live inside that branch, so on short-wide viewports (phone landscape) the select now gets its 40px height floor in the same breakpoint where `ui/src/app.css:367` bumps its font to 16px, and it stays the same height as the buttons it sits beside.

## Task

Satisfied. The caption is a `<span class="micro">` rendered before the `<select>` inside `<label class="mini-field">` (EpicPanel.svelte:247-254, 258-266, 268-276). Flipping `.mini-field` from `flex-direction: column` to `row` with `align-items: baseline` (452-457) puts the caption on the control's own line as an inline prefix, and `.micro` gains `flex: none; white-space: nowrap` (460-461) so it cannot shrink or wrap away from its select.

## Correctness checks

- `min-width: 96px` moved from `.mini-field` to `select` (EpicPanel.svelte:470). This is safe because the three selects at EpicPanel.svelte:249, 260 and 270 are the only `select` elements in the component (grepped the file), so the component-scoped bare `select` rule cannot leak onto an unintended control. With `max-width: 180px` (471) still in place, each field is bounded, and `.run-settings` retains `flex-wrap: wrap` (447) so the fields wrap rather than overflow on narrow phones.
- `.run-settings` changed from `align-items: end` to `center` (445). Verified the container relationship: `.run-settings` (246) is a flex item of `.epic-controls` (163), whose own rule already uses `display: flex; align-items: center; flex-wrap: wrap` (436-439) alongside the `.gbtn` buttons at 166/181/197/216/232. So the code comment at 543-544 ("the fields sit inline with the buttons") is accurate, and `center` is the right choice now that each field is a single row rather than a two-line column.
- The mobile `select { min-height: 40px }` interacts fine with `align-items: baseline` on `.mini-field`: the select's baseline is derived from its own text line box, so the caption tracks the select's text regardless of the taller box — which is exactly the behaviour the comment at 450-451 describes.

scoop (commented):
## What I checked

The diff is CSS-only, confined to `ui/src/lib/components/EpicPanel.svelte`. Three things change: `.mini-field` flips from a stacked column to a baseline-aligned row (label as inline prefix), the `min-width: 96px` moves from the field wrapper onto the `select` itself, and the mobile branch gains a `(max-height: 600px)` arm plus a 44px floor on both `.gbtn` and `select`.

**Does it satisfy the task?** Yes. `ui/src/lib/components/EpicPanel.svelte:247-255` renders `<span class="micro">{m.epic_provider_label()}</span>` followed by the provider `<select>` inside a `.mini-field` label; with `.mini-field { flex-direction: row; align-items: baseline }` (EpicPanel.svelte:452-457) the "CLI" caption now sits on the select's own line as a prefix rather than stacked above it. `epic_provider_label` is `"CLI"` in both catalogs — verified at ui/messages/en.json:1211 and ui/messages/de.json:1211, so the prefix stays short in both locales and the sibling captions (`Model`/`Modell`, `Effort`/`Aufwand` — en.json:1213-1214, de.json:1213-1214) are short too, so the nowrap caption plus the 96px select min-width will not blow out a narrow phone row. No message keys are added or renamed by this diff, so there is no parity work owed.

**Mobile media query.** The comment at EpicPanel.svelte:534-536 claims the condition mirrors the global mobile branch. Verified: `ui/src/app.css:366` is exactly `@media (max-width: 768px), (max-height: 600px)` (the iOS 16px zoom guard that also bumps `select` font-size), and `ui/src/routes/+page.svelte:1113` derives the app's `mobile` flag from the same `"(max-width: 768px), (max-height: 600px)"` query. So EpicPanel's controls now switch to the touch layout on exactly the viewports the app already considers mobile, instead of lagging on phone landscape.

**Specificity.** The scoped `select { min-height: 44px }` (EpicPanel.svelte:545-547, Svelte-scoped, specificity 0,1,1) has no competing global rule — app.css only sets `font-size` on `select` (app.css:367-369) — so the 44px floor actually lands, and the app.css 16px font guard still wins over the component's `font-size: var(--fs-meta)` as that file's comment intends.

**Tap-target value.** 44px matches the house floor, not an invented number: `docs/research/mobile-pwa-header-navigation.md:88` states the practical floor is 44px, `ui/src/app.css:95` defines `--mobile-actionbar-hit: 44px`, and sibling components use the same value (e.g. ui/src/lib/components/EpicDiagnoseEntry.svelte:287, ui/src/lib/components/CloneRepo.svelte:481). Raising `.gbtn` from 40px to 44px and floor-matching the adjacent selects keeps the shared control row on one height.

No behavioural code, props, or handlers changed, so there are no callers or contracts to re-check. I found nothing that blocks.

## Notes (non-blocking, no change requested)

- `align-items: baseline` on a flex row whose item is a `<select>` relies on the engine synthesizing the control's baseline from its inner text. That is what Chrome/Firefox/WebKit do for a default-appearance menulist, and the selects here keep the native appearance, so this should be fine; I could not execute a browser to confirm the rendered baseline across engines, so I state that as a limitation rather than a claim. `align-items: center` would be the more engine-agnostic choice if the caption ever looks off-centre next to the 44px mobile select.

- `ui/src/lib/components/EpicHandsOffIntro.svelte:248-250` is rendered inside this very panel (EpicPanel.svelte:98) but still uses a width-only `@media (max-width: 768px)` with `min-height: 40px`, so on phone landscape its buttons stay desktop-sized and below the 44px floor that this PR now applies to the panel's own controls.
- `ui/src/lib/components/EpicDiagnosisModal.svelte:470-472` likewise gates its 44px controls on `max-width: 768px` only, so the short-wide (phone-landscape) viewport is not covered there either.
- The `select { max-width: 180px }` cap (unchanged) combined with the app-wide 16px mobile font guard can truncate longer model names on phones; pre-existing, untouched by this diff.

Issue #1507, opened 2026-07-06
EPIC creation research flow: guided shaping, draft, approve, then create


kai-osthoff:
## Goal

Add a guided EPIC creation flow that starts from a product idea or rough prompt, researches/shapes the work, drafts an EPIC structure, asks for approval, and only then writes GitHub issues and links.

This should not be a one-click “AI creates issues” button. The hard gate is the product: draft first, write only after explicit approval.

## Dependencies

Blocked by #1505. The generated EPIC must follow the documented authoring contract.

Also depends on a guided shaping capability such as `grill-with-docs` or an equivalent “guided shaping session with durable notes/docs” mode. Do not hard-code the product design to one skill name if a capability abstraction fits better.

## Recommended flow

1. Operator starts “Create EPIC from research” for a repo and rough idea.
2. Shepherd starts an attended research/shaping session.
3. The session gathers context, asks focused questions, and sharpens scope using a grill-style interview.
4. It drafts:
   - parent issue title/body,
   - child issue titles/bodies,
   - dependency DAG,
   - acceptance criteria and non-goals.
5. Operator reviews the full tree and either amends, aborts, or approves.
6. Only after approval does Shepherd create child issues, create/edit the parent, add native sub-issues/dependencies or markdown fallback, and optionally import native links.
7. The deliverable is the EPIC itself; no code PR is opened.

## Shaping capability requirement

The flow should recommend or select a mode like `grill-with-docs` when the prompt is vague or large enough to need product shaping. Required behavior:

- ask one focused question at a time when product decisions are unresolved,
- research facts from the repo/docs instead of asking the user for discoverable facts,
- preserve durable decisions/notes where useful,
- produce child issues as vertical slices sized for one PR each,
- stop before GitHub writes until the operator approves.

## Non-goals for the first cut

- No fully autonomous issue creation from a vague prompt.
- No code implementation work.
- No direct PR creation.

## Acceptance criteria

- The flow can produce a reviewable EPIC draft without writing to GitHub.
- The operator sees the parent, child issues, and dependency DAG before approval.
- GitHub writes happen only after explicit approval.
- The flow uses or recommends a guided shaping mode such as `grill-with-docs` for ambiguous product ideas.
- Created/promoted EPICs are recognized by Shepherd immediately after the flow completes.

PR #1747, opened 2026-07-14
fix(ui): move epic draft review into a modal

scoop:
## What

A 12-child epic draft grew the inline `EpicDraftPanel` to its `60dvh` cap and squeezed the terminal to a single line, with no way to collapse it — and the reporting operator had no roomy surface to actually read the draft before approving it.

The panel is now a **one-line bar** (status chip · child count · `Review draft`), and the draft opens in a **full-height modal** whose body scrolls while amend/abort/approve stay pinned.

| before | after |
| --- | --- |
| panel eats 60dvh; terminal = 1 line; no collapse | bar = 1 row; terminal keeps the column; draft opens in a dialog |

> Note on the original report: the *overflow* in the screenshot was a stale browser bundle — #1734 already capped the panel and a hard reload made it scroll. This PR fixes what remained: even capped, the panel took the whole session view.

## Three non-obvious things this had to get right

1. **The dialog renders outside `.viewport`.** `.viewport` takes a `transform` while a swipe is in flight, and a transformed ancestor becomes the containing block for `position: fixed` — a nested dialog would look fine at rest and get dragged along with the pane under a finger-drag. It now sits next to `LeftoverDialog`, as a sibling.
2. **Global hotkeys didn't stand down for Viewport-owned dialogs.** `anyOverlayOpen()` was a hardcoded list of `+page`'s own flags, so `j`/`k`/`n`/`r` kept firing behind them — true for `LeftoverDialog` today, and it would have been true here. It now also probes `.overlay, .drawer`, mirroring the probe `shouldForwardEscape` already trusts.
3. **The bar survives the phone header fold while a draft awaits review.** `headerCollapsed` is persisted in `localStorage`, so an operator who folded the header once had *no* path to approve or abort at all. One row can afford to survive the fold; 60dvh could not.

Also: the dialog stays open across `materializing → approved` so the just-created parent-issue link is reachable, and `epicdraft_awaiting_hint` lost the now-wrong word "below" (EN + DE).

## Tests

- `EpicDraftPanel.browser.test.ts` — the bar stays one row and owns no draft body however long the draft; CTA fires; survives the fold while awaiting, folds away when quiet.
- `EpicDraftModal.browser.test.ts` — long draft scrolls; approve/abort are ≥44px and hit-testable (`elementFromPoint`) at scroll-top **and** scroll-bottom; Escape closes; stays open through `materializing → approved` and surfaces the parent link.
- `Viewport.browser.test.ts` — **composition, not the component in isolation.** Asserts the dialog is not a `.viewport` descendant, then *holds a drag open* (`touchstart` + `touchmove`, no `touchend`) to confirm `.viewport` really is transformed and the dialog's rect doesn't move, and that it closes on session switch.

The composition assertions were **mutation-tested**: nesting the modal back inside `.viewport` fails both — structurally, and with the overlay's rect shifting **-250px** with the pane. A standalone modal always scrolls, and a computed-transform check at rest sees no transform, so neither could have caught it.

## Verification

`bun run check`, `check:i18n`, and the full `bun test` (247 files / 3684 tests) pass in `ui/`; root `bun run lint` + `bun run test` pass. Also built the production bundle and confirmed the change survives it (`server.ts:229` serves the prebuilt `ui/build`, which is the staleness trap the original report came from): the scoped `.overlay` rule and the 44px bar CTA are in the emitted CSS, and `60dvh` no longer appears anywhere.

Screenshotted in real Chromium through the actual `Viewport`: bar is one row with the terminal intact; dialog renders with scrim + blur, card measured at `45..851` in an `896`-tall window, approve button 44px.

[no-feature-entry] — fix to a shipped feature (#1507), no new capability.


scoop (commented):
## Verdict

The change does what the task asks: the draft no longer lives in a `max-height: 60dvh` panel that squeezed the terminal and couldn't scroll — it now opens in `EpicDraftModal`, whose `.card` is a `max-height: 90dvh` flex column with a single `.body` scroller (`flex: 1 1 auto; min-height: 0; overflow-y: auto`) and a pinned `.actions` footer (`flex: none`). That is the correct flexbox shape for "content scrolls, actions stay", and it holds at both widths the tests exercise.

- **i18n parity.** `epicdraft_review_cta` and `epicdraft_view_cta` are present in both catalogs — `ui/messages/en.json:2957-2958` and `ui/messages/de.json:2957-2958` — and the reworded `epicdraft_awaiting_hint` (dropping "below", now that the draft isn't below anything) is updated in both at en.json:2955 / de.json:2955. `m.common_close()`, newly used by the modal's ✕, exists at en.json:5 and de.json:5. `ui/messages/` contains only `de.json` and `en.json`, so parity is complete.
- **`use:dialog` resolves.** `dialog` is exported from `ui/src/lib/a11yDialog.ts:36`; it supplies the focus trap, Escape→`onclose`, and focus restore that `EpicDraftModal.svelte:111` depends on.
- **The `.overlay` class name really is load-bearing, as the comment claims.** `Viewport.svelte:1803` passes `overlayOpen: !!document.querySelector(".overlay, .drawer")` into `shouldForwardEscape`, so Escape closes the dialog instead of being forwarded into the PTY. And the global `.overlay` rule at `ui/src/app.css:386-389` is blur-only (plus the reduced-transparency reset at :391-397), so the scoped position/scrim/z-index block in `EpicDraftModal.svelte:441-450` is genuinely not redundant.
- **The new DOM probe in `anyOverlayOpen()` can't wedge the shortcut tiers.** Every element bearing `.overlay`/`.drawer` in the tree is mounted behind an `{#if}`: PlanPanel (`PlanGateBadge.svelte:239`), LearningsDrawer (`page/AppOverlays.svelte:362`), ComposeBar (`Viewport.svelte:3018`), PuiActionButton's confirm (`plugin-ui/PuiActionButton.svelte:87`), DiagnoseRows' confirm (`DiagnoseRows.svelte:152`), EpicDiagnosisModal (`EpicPanel.svelte:282`), plus the flag-backed +page dialogs. Nothing carries those classes at rest. Near-miss tokens are distinct and correctly not matched: `.mobile-detail-overlay` (`BacklogView.svelte:341`), `.drop-overlay` (`viewport/FilesPanel.svelte:270`), `.pui-gauge-overlay` (`plugin-ui/PuiGauge.svelte:55`). `anyOverlayOpen()` has exactly one caller (`+page.svelte:1576`, inside `onShortcut`), so the widening is confined to shortcut suppression.
- **The hoist out of `.viewport` is real.** The modal is emitted after the `.viewport` root closes (`Viewport.svelte:3065`), and no ancestor establishes a containing block for `position: fixed`: `.col` (`+page.svelte:3372`), `.main-region` (:3399) and `.shell` (:3223) carry no transform/filter/perspective. The only mobile transform is on `.chrome` (:3467-3475), which is not an ancestor of the session column.
- **Stacking is sane.** The modal's `z-index: 40` sits above the mobile ActionBar (`ActionBar.svelte:245`, z-index 5) and below Toasts (`Toasts.svelte:68`, z-index 60), so a sticky approve/abort failure toast is still visible over the open dialog. `EpicDiagnosisModal.svelte:224` already uses 40, so this is consistent with existing modals.
- **Status handling is exhaustive.** `EpicDraftStatus` is exactly `"draft" | "materializing" | "approved"` (`ui/src/lib/types.ts:46`); the modal footer covers all three, and the `!draft || children.length === 0` branch covers the empty case.
- **No stale consumers.** `EpicDraftPanel` and `EpicDraftModal` are referenced only from `Viewport.svelte:2820` / `:3069` and their own tests, so dropping `sessionLive` from the panel and adding required `onreview` breaks no other caller.
- **The close-on-session-switch effect converges.** `Viewport.svelte:450-461` writes `epicDraftShownFor`, which it also reads; the write flips the guard, so the re-run takes the equality branch and stops. No update-depth loop.

## Finding

**`ui/src/lib/components/EpicDraftPanel.svelte`: the bar's CTA can be clipped out of reach at narrow widths.** `.edp-title` (:98), `.edp-chip` (:108) and `.edp-cta` (:141) are all `flex: none`; only `.edp-count` can shrink. The old panel was a column with `overflow: hidden` on `.edp`, so it could never push anything out of view — the new row has neither `overflow: hidden` nor `flex-wrap`, and it sits inside `.viewport { overflow: hidden }` (`Viewport.svelte:3102`), so anything past the right edge is silently clipped rather than scrolled. The clipped item is the CTA, and per this PR's own design that button is the *only* way into the review surface (and, on a folded phone, the only path to approve or abort at all — `Viewport.svelte:2816-2823`). The margin is thin in German at 390px ("EPIC-ENTWURF" + "DEINE PRÜFUNG NÖTIG" + "Entwurf prüfen" against a 44px-tall button with 12px side padding) and gets thinner as `--ui-scale` grows, since every size here is `calc(px * var(--ui-scale))` (`app.css:81-84`). I could not measure the exact overflow point without running a browser, so I am flagging the mechanism rather than asserting a specific failure width — but the fix is cheap: let `.edp-title`/`.edp-chip` shrink (or hide the title below a breakpoint) so the CTA always keeps its box. The new test asserts only the bar's *height* (`EpicDraftPanel.browser.test.ts:52`), so horizontal clipping would pass CI unnoticed; an assertion that the CTA's `right` stays within the bar's `right` would close that gap.

## Nit

**`ui/src/lib/components/EpicDraftPanel.svelte`: `.is-awaiting` is now an unqualified class rule matching two elements.** `EpicDraftPanel.svelte:71` adds `class:is-awaiting` to the CTA button, so the bare `.is-awaiting { background: color-mix(… amber 6% …) }` rule at :93 — written for the bar at :53 — now also matches the button. It is currently harmless only because `.edp-cta { background: none }` (:143) has equal specificity and comes later in the sheet; reordering the rules or dropping that `background: none` would tint the button amber. Qualifying it as `.edp.is-awaiting` makes the intent explicit and removes the ordering dependency.

scoop (commented):
## Re-review: both prior points are addressed

**1. CTA clipped on narrow phones — FIXED.** `ui/src/lib/components/EpicDraftPanel.svelte:104-127` now gives every label (`.edp-title`, `.edp-chip`, `.edp-count`, `.edp-empty`) `min-width: 0` + `overflow: hidden` + ellipsis, and only `.edp-title`/`.edp-chip` carry `flex: 0 1 auto` (lines 111, 128); `.edp-count`/`.edp-empty` inherit the shrinkable default. `.edp-cta` alone is `flex: none` (EpicDraftPanel.svelte:145), so it keeps its box while everything else yields. The title additionally drops out under `@media (max-width: 480px)` (EpicDraftPanel.svelte:118-122). The requested containment assertion exists: `ui/src/lib/components/EpicDraftPanel.browser.test.ts:71-105` runs 320px/1.0, 320px/1.3 and 240px/1.3 (`--ui-scale`) and asserts `ctaBox.width > 0`, `ctaBox.right <= barBox.right + 1`, `ctaBox.left >= barBox.left - 1`, plus `bar.scrollHeight <= bar.clientHeight + 1` (still one row).

**2. Unqualified `.is-awaiting` — FIXED.** It is now `.edp.is-awaiting` (ui/src/lib/components/EpicDraftPanel.svelte:93), so the CTA's `class:is-awaiting` (line 71) no longer matches it and the rule no longer depends on `.edp-cta { background: none }` coming later in source order.

## Verification of the new code

**The task itself (draft doesn't scroll) is satisfied.** `ui/src/lib/components/EpicDraftModal.svelte:257-267` makes the card a flex column with `max-height: 90dvh`, and `.body` (EpicDraftModal.svelte:322-331) is `flex: 1 1 auto; min-height: 0; overflow-y: auto; overscroll-behavior: contain; touch-action: pan-y` — the `min-height: 0` is what actually lets it shrink below content. `.actions` is `flex: none` (EpicDraftModal.svelte:381), so the approve/abort/amend row stays pinned. `max-height: …dvh` on a modal card has precedent in this tree (CodexUpdateModal.svelte:153, HerdrUpdateModal.svelte:245, TopBarMobileSheet.svelte:522).

**Identifiers resolve.** `dialog` → ui/src/lib/a11yDialog.ts:36 (takes `{ onclose }`, traps Tab, closes on Escape); `epicDrafts.get/load/upsert` → ui/src/lib/epic-draft.svelte.ts:14, 24, 18; `EpicDraftChildRow` → ui/src/lib/components/EpicDraftChildRow.svelte (unchanged, still fed `child`/`index`/`materializedNumber`/`blockedLabel`). `EpicDraftPanel`'s prop shape changed (`sessionLive` dropped, `folded`/`onreview` added) — the only non-test consumer is Viewport.svelte:2820-2825, which passes exactly the new set, and `sessionLive` moved to the modal (Viewport.svelte:3068).

**i18n parity holds.** `epicdraft_review_cta` and `epicdraft_view_cta` exist in both catalogs at the same position — ui/messages/en.json:2957-2958 and ui/messages/de.json:2957-2958 — as does the reworded `epicdraft_awaiting_hint` (en.json:2956 / de.json:2956). Every other key the modal calls (`epicdraft_parent_label`, `epicdraft_acceptance_label`, `epicdraft_nongoals_label`, `epicdraft_amend_*`, `epicdraft_abort*`, `epicdraft_approve*`, `epicdraft_created`, `epicdraft_view_parent`, `epicdraft_children_label`) is present in both (en.json:2953-2980 / de.json:2953-2980), and `common_close` is in both (en.json:5, de.json:5).

**The `class="overlay"` contract is load-bearing and is honoured.** `shouldForwardEscape` is fed `overlayOpen: !!document.querySelector(".overlay, .drawer")` at Viewport.svelte:1803, and app.css:386-389 contributes only the blur (the scoped rule at EpicDraftModal.svelte:254 supplies position/scrim/z-index) — exactly as the comment at EpicDraftModal.svelte:98-104 claims. `z-index: 40` matches the sibling full-screen dialog EpicDiagnosisModal.svelte:224 and clears ViewportTabBar's 30 (ViewportTabBar.svelte:430).

**The new DOM probe in `anyOverlayOpen` does not misfire.** ui/src/routes/+page.svelte:1531 adds `document.querySelector(".overlay, .drawer")`. I enumerated every element in the tree carrying those exact class tokens: ForkRepo:66, FeedbackDialog:44, DiagnoseRows:152, ComposeBar:306, PluginUpdatesModal:140, CommandBar:460, PlanPanel:279, CodexUpdateModal:56, EpicDiagnosisModal:115, CloneRepo:133, EpicDiagnoseEntry:72, HerdrUpdateModal:114, ClearMergedDialog:22, NewTask:978, NewProject:152, LeftoverDialog:31, BroadcastDialog:129, MergeTrainConfirmDialog:28, Settings:1097, BacklogOverlay:57, Usage:157, RetryDialog:78, UpdateModal:52, RestartShepherdDialog:90, RecommendDialog:85, PluginConfirmDialog:28, PuiActionButton:89, LearningsDrawer:199, WhatsNew:94 — every one is a blocking modal/drawer rendered behind an `{#if}` or a conditionally-mounted component, so none is persistently in the DOM. Near-miss class names (`mobile-detail-overlay`, `overlay-head`, `overlay-tabs` in BacklogView; `drop-overlay` in FilesPanel; `pui-gauge-overlay`) do not match the `.overlay` token selector. The single call site (+page.svelte:1576, inside `onShortcut`, which bails on `mobile.current` first) is a keydown handler, so `document` is always defined.

**Session-switch teardown.** Viewport.svelte:452-462 closes the dialog when the component is reused for another session and when the draft disappears; the effect writes state it reads (`epicDraftShownFor`), but the write is idempotent on the second pass, so it converges rather than looping. Covered by Viewport.browser.test.ts:1479-1490.

Nothing to change. No blocking or non-blocking findings.

PR #1749, opened 2026-07-14
fix(codex): remember the update channel that actually advances codex

scoop:
## The bug

Every codex update ran **two** installers. The screenshots show one run: `codex update` succeeds, and the "fallback" `npm install -g` fires anyway.

It was never a spurious retry — **`codex update` was updating the wrong install, and the fallback was the only step that worked.**

Evidence from `~/.shepherd/codex-update.log` (3 consecutive runs, all identical) plus on-disk state:

1. The first `codex` on PATH is not an install, it's a wrapper: `exec mise exec node@latest -- npx --yes @openai/codex`, and npx serves the **npm-global** copy.
2. `codex update` self-selected **`bun install -g`** — writing a store nothing on PATH resolves. Bun's global codex was still **0.93.0** (February), which is the proof it was never used.
3. So the re-read version never advanced, `after == before` was always true, and the npm fallback fired every time. `~/.npm-global/bin/codex`'s mtime matched the fallback's timestamp exactly.

Cost: two installers per update, and codex written into **two** channels — manufacturing the very PATH duplicates the script's own diagnostics warn about (this host has three `codex` entries on PATH).

The tempting fix — trust `codex update`'s `rc=0` and drop the fallback — would have **stopped updating codex on this host entirely** while the modal claimed "codex was not updated". Rejected on the evidence.

## The fix: a channel memo

Shepherd already learned which channel worked on every run. It just threw the answer away.

- After a run in which the on-PATH version **advanced**, persist which channel did it (`npm` / `codex`) in the existing settings kv store — no new file, table, or migration.
- Next update leads with that channel; the other stays a fallback on non-advancement, so a stale memo (operator switched channels) misses **once** and rewrites itself.
- **No memo = the historical order** (`codex update`, then npm), so a fresh host is never worse off than before.

Steady state is one installer run per update on every host shape — including standalone (#1560), which keeps `codex update` and must never fall back to npm.

The memo is written **only** from a converged run that named its winner. `ok` is shepherd's own `compareSemver` verdict: it says *that* codex advanced, never *by what*. So an unattributable run — no `converged via channel=` marker, a watchdog-killed run, or a non-converged one — leaves the memo **untouched**, never cleared and never guessed.

## Verification

Unit tests **execute the generated script for real** against fake `codex`/`npm` installers on a controlled PATH (`bash -c`, never `-lc` — a login shell sources the profile and could reach the real `npm install -g`). Each fake touches a sentinel, so tests assert which installer ran as a fact. Cases: memo=npm, memo=codex, cold+codex-advances (the standalone/#1560 shape), cold+codex-no-ops, stale memo, no `update` subcommand, neither converges.

End-to-end on a real host (downgrade → real service → real installers): a warm run produced **exactly one** installer invocation, `npm install` never ran, and codex converged to latest. Old blocks in the same log show `codex update exited` + `npm install exited` **pairs**; the new blocks show a single invocation.

Root: `bun run lint`, `bun run test` (6957 pass). UI: `bun run check`, `check:i18n` (parity), `bun run test` (3675 pass).

### One honest note on the E2E

During verification `codex update` **converged on the first attempt** — it now self-selects npm (`changed 2 packages`) rather than bun, so the memo settled on `codex`, not `npm` as predicted. The double-install did not reproduce live; the three logged runs are the record that it did. This doesn't change the fix's correctness — the memo is empirical, so it records whichever channel actually wins, and if `codex update` flips back to bun the fallback converges and rewrites the memo to `npm`. It does mean the win on *this* host is currently latent rather than visible.

## Also

Rewords three strings that hard-code "codex update", including `codexupdate_instructions` — the **pre-confirm consent copy**, which promised "Runs codex update (auto-detects your install)" and is false whenever the memo picks npm. EN + DE, no new keys.

[no-feature-entry] — bug fix, no new user-facing capability.


scoop (commented):
## Verdict

The change does what the task asks. The old script ran `codex update` and then **unconditionally** fell back to `npm install -g` whenever the version had not advanced (old `src/codex-update.ts`: `if [ "$has_update" -eq 0 ] || [ "$after" = "$before" ]`), so on a host where `codex update` self-selects an installer that misses the on-PATH copy, **both** installers ran on every update. The new script loops over an ordered channel list, breaks on the first channel that actually advances `codex --version` (`src/codex-update.ts:154`), names the winner (`src/codex-update.ts:158`), and shepherd persists it (`src/index.ts:2481-2483`) and leads with it next time (`src/codex-update.ts:291`). Steady state is one installer run. Correct fix for the reported symptom.

- **Marker round-trip.** The script emits `echo "${CONVERGED_MARKER}$winner"` (`src/codex-update.ts:158`) and the scraper slices on the same exported constant `CONVERGED_MARKER` (`src/codex-update.ts:39`, consumed at `src/codex-update.ts:378-381`), validating the payload against the closed union before assigning. The `>` characters in the prefix (`CODEX_UPDATE_LOG_PREFIX = ">>> codex-update:"`, `src/codex-update.ts:19`) sit inside a double-quoted `echo`, so no redirection — the interpolated `echo` at :158 is the only one with an unquoted-`>` risk and it is quoted.
- **Persistence API.** `store.getSetting(key): string | null` and `store.setSetting(key, value): void` exist with those exact signatures (`src/store.ts:1245` and `src/store.ts:1252`); `store` is constructed at `src/index.ts:212`, well before the `CodexUpdateService` at `src/index.ts:2473`. Settings are a free-form kv table (no allowlist to register `codexUpdateChannel` in — cf. the other ad-hoc keys at `src/index.ts:216,242,275`). The `readChannel` narrowing at `src/index.ts:310-311` correctly rejects any other stored string.
- **Shell injection surface.** `PREF` is interpolated only from the closed union, coerced to `""` otherwise (`src/codex-update.ts:105,110`); `LOG`/`CX` keep the pre-existing single-quote escaping. No new injection path.
- **Probe still gates `codex update`.** When `--help` does not advertise the subcommand, `order='npm'` (`src/codex-update.ts:133`), so `codex` cannot appear even as the fallback — the old-codex safety rail survives.
- **Signature change.** `buildUpdateScript` gained a 5th param with a `= null` default (`src/codex-update.ts:95`); its only non-test caller is `defaultRunUpdate` (`src/codex-update.ts:286-292`), which passes it. The unrelated `buildUpdateScript` in `src/herdr-update.ts:56` (used by `src/remediations.ts:124`) is a different symbol and is untouched.
- **i18n parity.** The repo maintains exactly two catalogs (`ui/messages/en.json`, `ui/messages/de.json`). No keys were added or removed — the same three existing keys were reworded in both, at identical positions (`ui/messages/en.json:2375,2377,2380` vs `ui/messages/de.json:2375,2377,2380`), and all three are still referenced from `ui/src/lib/components/CodexUpdateModal.svelte:95,103,112`. The reworded consent copy is now accurate for an npm-led update.
- **Memo write is conservative.** `settleUpdate` returns on `aborted` before the write (`src/codex-update.ts:439`) and writes only on `ok && channel` (`src/codex-update.ts:446`), so a killed, non-converged, or unattributable run can't poison a good memo.
- **Not a defect:** `ui/src/lib/components/CodexUpdateModal.browser.test.ts:36` still contains the removed `>>> codex-update: running codex update` marker, but only as fixture text for a layout assertion — it asserts nothing about the script's output.

## Nits (non-blocking)

1. **A failing memo write downgrades a successful update to a failure.** `this.writeChannel(channel)` (`src/codex-update.ts:446`) runs inside `settleUpdate`, which is called inside `runOnce`'s `try` (`src/codex-update.ts:403`). The real `writeChannel` is a SQLite write (`src/index.ts:2483` → `src/store.ts:1252`), which can throw (locked/failed db). If it does, the `catch` at `src/codex-update.ts:404` turns a *converged* update into `{ok: false, error: …}` and `onStatus` never fires — the operator is told the update failed when it landed. A `try { … } catch {}` around the memo write keeps a bookkeeping failure from rewriting the outcome.

2. **Bash's command hash can defeat the advancement check.** `config.codexBin` defaults to the bare name `"codex"` (`src/config.ts:540`), and `readVersion` invokes `"$CX" --version` (`src/codex-update.ts:106`). The first invocation (`before=`, `src/codex-update.ts:115`) makes bash hash the resolved path; if a channel then installs codex into a directory that sits *earlier* on PATH, the post-run re-read at `src/codex-update.ts:151` still executes the stale hashed path and reports "did not advance" — a false negative that both suppresses the memo and runs the other installer, i.e. exactly the double-run this PR removes. A `hash -r` (or `hash -d codex`) before the re-read makes the winner test resolve PATH afresh. The old script had the same weakness, but the check is now load-bearing for a persisted decision.

3. **The class doc no longer matches the implementation.** `src/codex-update.ts:243` still says "`apply()` spawns `codex update` (with an npm fallback)" — this PR replaced that with a memo-ordered two-channel loop that may lead with npm. The `buildUpdateScript` docblock right above was rewritten thoroughly; this one was missed.

- src/codex-update.ts:133 — when the `update` subcommand is absent the script sets `src='default'` even if a memo exists, so the log prints `channel=npm (source=default)`; harmless today (the order is forced to `npm` regardless, not chosen from the memo), but it contradicts the documented `source=memo|default|fallback` semantics at src/codex-update.ts:83 if anyone ever parses that field.

scoop (commented):
## Re-review of the three prior findings — all addressed

**1. Memo write inside `runOnce`'s try block could rewrite a converged outcome — FIXED.** `this.writeChannel(channel)` is now wrapped in `try/catch` at `src/codex-update.ts:466-475`, logging a `console.warn` and falling through. The verdict is computed at `src/codex-update.ts:455` and returned at `src/codex-update.ts:486-488` regardless, and `this.onStatus(this.last)` at `src/codex-update.ts:485` is now unreachable-past only by a non-throwing path. Verified the throwing dependency it guards against is real: `writeChannel` is wired to `store.setSetting` at `src/index.ts:2485`, which is a bare `this.db.run(...)` with no internal catch (`src/store.ts:1252-1258`), so it can throw.

**2. Stale bash command hash on the re-read — FIXED.** `hash -r 2>/dev/null || true` is emitted at `src/codex-update.ts:157`, immediately before the `after="$(...)"` re-read at `src/codex-update.ts:158`, inside the per-channel loop and after the installer runs. That is the right placement: the `before` read at `src/codex-update.ts:115` is what hashes the bare `codex` name (`config.codexBin` defaults to a bare name — the comment at `src/codex-update.ts:119-120` relies on the same assumption for `type -a codex`), and the re-read at 158 is a command substitution, i.e. a subshell that inherits the parent's hash table — so clearing it in the parent first does make the subshell re-search PATH. The winner test at `src/codex-update.ts:161` therefore no longer risks persisting a false "did not advance".

**3. Stale class docblock — FIXED.** `src/codex-update.ts:248-262` no longer says `apply()` "spawns `codex update` (with an npm fallback)"; it now describes the memo-ordered two-channel loop and explicitly notes that with a memo of `npm` the run may never invoke `codex update` at all. This matches the rewritten `buildUpdateScript` doc at `src/codex-update.ts:61-67` and the actual order logic at `src/codex-update.ts:133-136`.

## Other verification

- **Memo key round-trips.** `CODEX_UPDATE_CHANNEL_KEY = "codexUpdateChannel"` (`src/index.ts:2472`) is the single constant used by both `readChannel` (`src/index.ts:2481`) and `writeChannel` (`src/index.ts:2485`); grepping the tree for `codexUpdateChannel` finds no other spelling and no other consumer.
- **Marker parity between emitter and parser.** The script echoes `"${CONVERGED_MARKER}$winner"` at `src/codex-update.ts:165`; the parser slices on the same exported `CONVERGED_MARKER` constant (`src/codex-update.ts:39`, used at `src/codex-update.ts:393-396`) and accepts only `"npm"`/`"codex"`, the exact tokens `$winner` can hold (`src/codex-update.ts:161`, from `order` at 133-136). The marker string contains no `$` or backtick, so it is safe inside the double-quoted `echo`, and the `>>>` in `CODEX_UPDATE_LOG_PREFIX` (`src/codex-update.ts:19`) is quoted, not a redirect.
- **Shell injection surface.** `pref` is re-narrowed to the closed union at `src/codex-update.ts:105` before interpolation into `PREF='...'` at 110, so nothing external reaches the script body.
- **Aborted runs never memoize.** `settleUpdate` returns at `src/codex-update.ts:454` before the memo block at 466 when `aborted` — as the comment claims.
- **Signature change has one caller.** `buildUpdateScript` gained a 5th param with a `null` default (`src/codex-update.ts:95`); the only production caller passes it (`src/codex-update.ts:301-307`).
- **i18n parity.** The repo maintains exactly two catalogs (`ui/messages/en.json`, `ui/messages/de.json`). The diff adds and removes no keys — it only rewords `codexupdate_instructions`, `codexupdate_busy`, and `codexupdate_done_fail_stuck` — and both files receive the same three edits (`ui/messages/en.json:2375,2377,2380` and `ui/messages/de.json:2375,2377,2380`). Consumers `m.codexupdate_instructions()` and `m.codexupdate_busy()` at `ui/src/lib/components/CodexUpdateModal.svelte:95,112` are unaffected. The new wording is also now accurate: it no longer promises "runs codex update", which the memo path may never do.

## Remaining item (non-blocking)

The try/catch reasoning applied to `writeChannel` was not applied to its sibling `readChannel`. `this.readChannel()` is invoked at `src/codex-update.ts:306`, inside the `new Promise` executor of `defaultRunUpdate` — so a throw there rejects the promise before `spawn` is ever reached. It is wired to `store.getSetting` (`src/index.ts:2482`), which is an unguarded `this.db.query(...).get(...)` (`src/store.ts:1245-1250`) and can throw on exactly the locked/failed-SQLite condition finding #1 was raised for. The reported outcome is not *wrong* (the update genuinely did not run, and `runOnce`'s catch at `src/codex-update.ts:419-427` reports `ok:false` honestly), so this is a nit rather than a blocker — but the memo is an optimisation, and a failed read of it should degrade to "no memo, historical order" rather than cancel the update the operator asked for.

scoop (commented):
## Re-review: prior finding

**1. `readChannel()` called unguarded inside `defaultRunUpdate`'s Promise executor — ADDRESSED.**

The read is now hoisted out of the spawn path entirely: `runOnce` calls `readChannelSafe()` and passes the result into `runUpdate` (src/codex-update.ts:457-460), and `readChannelSafe` (src/codex-update.ts:312-322) wraps `this.readChannel()` in try/catch, warning and degrading to `null` — which `buildUpdateScript` maps to `PREF=''` → `order='codex npm'; src='default'` (src/codex-update.ts:136), i.e. the historical order. `defaultRunUpdate` (src/codex-update.ts:327-368) no longer touches the store at all; it only receives `preferred` as a parameter. This is exactly the treatment asked for, and it mirrors the write side (src/codex-update.ts:509-518). Pinned by test/codex-update.test.ts:645+ ("a throwing store READ degrades to the default order, never cancels the update"), which asserts `runUpdate` still ran with `preferred === null`. The concrete throw path named in the prior finding — `store.getSetting` at src/index.ts:2482 → the unguarded `db.query(...).get(...)` at src/store.ts:1245-1249 — is now contained.

## What I verified for the new code

- **Identifiers resolve.** `store.getSetting(key): string | null` (src/store.ts:1245) and `store.setSetting(key, value: string): void` (src/store.ts:1252) match the `readChannel`/`writeChannel` lambdas at src/index.ts:2481-2485; the `saved === "npm" || saved === "codex"` narrowing yields exactly `CodexUpdateChannel | null` (src/codex-update.ts:33). The settings key `"codexUpdateChannel"` (src/index.ts:2472) is used in both the read and the write and collides with no other key in the tree (grep over `src/`, `test/` returns only those three lines).
- **Contract changes have no stale consumers.** `runUpdate` gained a third parameter and `settleUpdate`/`runUpdateCapturingPath` changed shape. The only external consumer of the class is src/server.ts:294, `Pick<CodexUpdateService, "current" | "apply">` — neither signature changed. The only other `buildUpdateScript` in the tree is herdr's (src/herdr-update.ts:56), imported by src/remediations.ts:2; it is a different symbol and untouched.
- **Marker round-trip.** `CONVERGED_MARKER` (src/codex-update.ts:39) = `">>> codex-update: converged via channel="`; the script emits it via `echo "${CONVERGED_MARKER}$winner"` (src/codex-update.ts:165) and the scraper matches on the same constant (src/codex-update.ts:427-431), whitelisting only `npm`/`codex`. The `>>>` sits inside double quotes so it is literal text, not a redirect — same construction as the pre-existing `${P}` echoes. No other repo code greps for the two removed markers (`running codex update`, `falling back to npm`) except a synthetic layout fixture, noted below.
- **Shell logic.** `has_update=0` forces `order='npm'` *before* the memo is consulted (src/codex-update.ts:133), so a stale `codex` memo can never hand `update` to a codex that lacks the subcommand — pinned by the executing harness at test/codex-update.test.ts:270. The winner test compares `$after != $before` after a `hash -r` (src/codex-update.ts:157-161), which is load-bearing: `"$CX" update` runs in the parent shell and hashes `codex`→old path, while `before`/`after` are read in command substitutions that inherit that table. A downgrade would satisfy `!=` and break the loop, but node's `ok` is `compareSemver(after, from) > 0` (src/codex-update.ts:498), so it is reported as a failure and never memoized — and the *old* script skipped its npm fallback on the same input (`after != before`), so this is not a regression.
- **Memo is write-only-on-attributed-success.** `if (ok && channel)` (src/codex-update.ts:509) with the abort early-return above it (src/codex-update.ts:497), and the write itself swallowed. A run whose script never emitted the marker leaves the memo untouched rather than guessing.
- **Injection.** `PREF` is built from a closed union check (src/codex-update.ts:105), not from the raw store string, and the store value is *also* narrowed at src/index.ts:2483 — double-guarded.
- **i18n parity.** Both catalogs carry the same 11 `codexupdate_*` keys; the three reworded values (`codexupdate_instructions`, `codexupdate_busy`, `codexupdate_done_fail_stuck`) changed in ui/messages/en.json:2375,2377,2380 *and* ui/messages/de.json:2375,2377,2380. No keys added or removed, and all three are consumed by ui/src/lib/components/CodexUpdateModal.svelte:95,103,112.

- ui/src/lib/components/CodexUpdateModal.browser.test.ts:36 feeds the modal a synthetic log line `">>> codex-update: running codex update"` — a marker this PR's script no longer emits. It is a layout fixture with no assertion on script output, so nothing breaks; the string is just cosmetically stale.

PR #1750, opened 2026-07-14
fix(ui): wrap long toast action label instead of squishing message

scoop:
The post-merge toast for an isolated session offers **Decommission & update local** — a long action label that squeezed the message into a narrow 3-line column (reported with a screenshot).

## Cause

`.toast` was a **non-wrapping** flex row capped at `min(440px, 92vw)`. `.undo` is `flex-shrink: 0`, uppercase, `letter-spacing: 0.12em`, so that label reserved ~270px of the 440px. `.msg` was the only shrinkable child, so it absorbed the entire deficit.

## Fix

The row now wraps: the message keeps its own line and the controls drop to a second row.

- **`.actions` wrapper (markup).** The action button and the ✕ become **one atomic flex item**. As two siblings they are collected onto the wrapped line independently, so the ✕ orphans onto a *third* row whenever button + gap + ✕ exceeds the line box (~293px vs a 320px banner's ~292px content box). `flex-shrink` can't rescue that — shrinking happens *after* line collection. The wrapper also leaves exactly **one** `margin-left: auto` on the row, so Flexbox §8.1 can't split the free space equally between two auto margins and float the button mid-row. `.countdown` stays a direct child of `.toast` (it is absolutely positioned against it).
- **Explicit `row-gap: 8px`.** `gap: 14px` is a both-axes shorthand, so wrapping would have silently adopted 14px vertically. 8px matches the stack rhythm.
- **`.undo` shrinkable.** At the iOS Dynamic Type cap (`--ui-scale: 1.5` → `--fs-meta` 16.5px) the button grows to ~380px against a 390px phone's ~362px content box. It now wraps its label *inside* the button instead of overflowing the banner.

## Tests

9 real-CSS geometry tests in `Toasts.browser.test.ts` (all strings from the real message fns, so a future label change re-triggers them). **6 of them fail against the pre-fix component** — verified by restoring it and re-running.

The core assertion counts the message's **line boxes** via a `Range` (3 before, 1 after) rather than box width, which the new `flex-grow` would make trivially true. Also covered: ✕ adjacency on the wrapped row, the 8px row-gap, no horizontal overflow with an unbreakable 60-char branch name, the short-label control still on one row, `.countdown` parentage, 390px/320px banners, and `--ui-scale: 1.5` × 390px (asserting the shrunk button *wrapped* rather than clipped its label — it keeps `overflow: hidden` for the `.bar` countdown).

Same-row checks compare vertical centers, not `top`: `.toast` is `align-items: center` and `.undo` (~25px) / `.x` (~21px) differ in height, so correct layout does not share a `top`.

## Verification

`bun run check` clean · `check:i18n` in parity · root `lint` + `prettier --check` clean · full UI suite green (3697 tests, 248 files).

No new message keys, no user-facing capability — a CSS/markup bugfix on shipped UI. [no-feature-entry]

### DIFF
--- a/ui/src/lib/components/Toasts.browser.test.ts
+++ b/ui/src/lib/components/Toasts.browser.test.ts
@@ -161,6 +161,37 @@ describe("Toasts: long action label wraps instead of squishing the message", () 
     expect(msgLineCount(msg)).toBe(1);
   });
 
+  /* .actions owns the row's ONLY auto margin; .undo and .x declare none. These two
+     pin the right-alignment that used to ride on those removed margins: the undo-tone
+     button (a direct child of .toast, outside .actions) is pushed right by .msg's
+     flex-grow, and a lone ✕ by .actions' own auto margin. */
+  it("right-aligns the undo-tone button, which sits outside .actions", async () => {
+    await page.viewport(1280, 900);
+    toasts.undo(MERGED, { undoLabel: m.common_undo(), onCommit: () => {} });
+    render(Toasts, {});
+    await tick();
+    const toast = el<HTMLElement>(".toast");
+    const undo = el<HTMLElement>(".undo");
+    expect(document.querySelector(".actions")).toBeNull(); // undo tone has no wrapper
+    // Flush right (12px padding + 1px border), not left-adrift beside the message.
+    expect(toast.getBoundingClientRect().right - undo.getBoundingClientRect().right).toBeCloseTo(
+      13,
+      0,
+    );
+  });
+
+  it("right-aligns a lone ✕ when there is no action label", async () => {
+    await page.viewport(1280, 900);
+    await renderToast(MERGED);
+    const toast = el<HTMLElement>(".toast");
+    const x = el<HTMLElement>(".x");
+    expect(document.querySelector(".undo")).toBeNull();
+    expect(toast.getBoundingClientRect().right - x.getBoundingClientRect().right).toBeCloseTo(
+      13,
+      0,
+    );
+  });
+
   it("keeps .countdown a direct child of .toast (absolutely positioned against it)", async () => {
     await page.viewport(1280, 900);
     toasts.info(MERGED, {
--- a/ui/src/lib/components/Toasts.browser.test.ts
+++ b/ui/src/lib/components/Toasts.browser.test.ts
@@ -161,35 +161,50 @@ describe("Toasts: long action label wraps instead of squishing the message", () 
     expect(msgLineCount(msg)).toBe(1);
   });
 
-  /* .actions owns the row's ONLY auto margin; .undo and .x declare none. These two
-     pin the right-alignment that used to ride on those removed margins: the undo-tone
-     button (a direct child of .toast, outside .actions) is pushed right by .msg's
-     flex-grow, and a lone ✕ by .actions' own auto margin. */
-  it("right-aligns the undo-tone button, which sits outside .actions", async () => {
-    await page.viewport(1280, 900);
-    toasts.undo(MERGED, { undoLabel: m.common_undo(), onCommit: () => {} });
+  /* .actions owns the row's ONLY auto margin; .undo and .x declare none. These pin the
+     right-alignment that used to ride on those removed margins. The undo tone is in
+     .actions too — a lone button on a wrapped line has no free space to grow into, so
+     .msg's flex-grow cannot reach it and it would left-align without the wrapper. */
+  async function renderUndo(text: string) {
+    toasts.undo(text, { undoLabel: m.common_undo(), onCommit: () => {} });
     render(Toasts, {});
     await tick();
-    const toast = el<HTMLElement>(".toast");
-    const undo = el<HTMLElement>(".undo");
-    expect(document.querySelector(".actions")).toBeNull(); // undo tone has no wrapper
-    // Flush right (12px padding + 1px border), not left-adrift beside the message.
-    expect(toast.getBoundingClientRect().right - undo.getBoundingClientRect().right).toBeCloseTo(
-      13,
-      0,
+  }
+
+  /** Distance from the toast's right edge — 13px (12px padding + 1px border) when flush. */
+  function insetRight(node: HTMLElement): number {
+    return (
+      el<HTMLElement>(".toast").getBoundingClientRect().right - node.getBoundingClientRect().right
     );
+  }
+
+  it("right-aligns the undo-tone button on an unwrapped row", async () => {
+    await page.viewport(1280, 900);
+    await renderUndo(MERGED);
+    expect(insetRight(el<HTMLElement>(".undo"))).toBeCloseTo(13, 0);
+  });
+
+  it("right-aligns the undo-tone button once the row WRAPS (long session name)", async () => {
+    await page.viewport(1280, 900);
+    // Reachable in real use: toast_decommissioned carries a session name and routinely
+    // exceeds the 440px cap. Without .actions the button would be alone on line 2 with
+    // no auto margin and no grow → left-adrift, unlike an info toast in the same state.
+    await renderUndo(m.toast_decommissioned({ name: "shepherd/toast-decommission-update-local" }));
+    const msg = el<HTMLElement>(".msg");
+    const undo = el<HTMLElement>(".undo");
+    // Actually wrapped (else this asserts nothing).
+    expect(undo.getBoundingClientRect().top).toBeGreaterThanOrEqual(
+      msg.getBoundingClientRect().bottom,
+    );
+    expect(insetRight(undo)).toBeCloseTo(13, 0);
   });
 
   it("right-aligns a lone ✕ when there is no action label", async () => {
     await page.viewport(1280, 900);
     await renderToast(MERGED);
-    const toast = el<HTMLElement>(".toast");
     const x = el<HTMLElement>(".x");
     expect(document.querySelector(".undo")).toBeNull();
-    expect(toast.getBoundingClientRect().right - x.getBoundingClientRect().right).toBeCloseTo(
-      13,
-      0,
-    );
+    expect(insetRight(x)).toBeCloseTo(13, 0);
   });
 
   it("keeps .countdown a direct child of .toast (absolutely positioned against it)", async () => {
--- a/ui/src/lib/components/Toasts.browser.test.ts
+++ b/ui/src/lib/components/Toasts.browser.test.ts
@@ -2,6 +2,7 @@ import { describe, it, expect, beforeEach, afterEach } from "vitest";
 import { render } from "vitest-browser-svelte";
 import { page } from "vitest/browser";
 import { tick } from "svelte";
+import { m } from "$lib/paraglide/messages";
 // MUST import the stylesheet: the inset is calc()'d from the design tokens
 // (--mobile-actionbar-hit/-pad/--actionbar-border, all defined in app.css).
 // Without it the calc is invalid and getComputedStyle().bottom is `auto`, not px.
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Verdict

The change does what the task asks: the long action label no longer squeezes `.msg`. Verified by reading the layout end to end rather than eyeballing it.

### What I verified

- **The squeeze is genuinely gone.** `.msg` gains `flex: 1 1 auto` (`ui/src/lib/components/Toasts.svelte:110`), so its flex base size is its max-content width; `.toast` now has `flex-wrap: wrap` (`Toasts.svelte:89`). Line-breaking therefore pushes `.actions` (base = max-content of button + gap + ✕) onto a second line instead of shrinking `.msg` down to a 3-line column, which is exactly the reported symptom. `max-width: min(440px, 92vw)` (`Toasts.svelte:94`) still clamps the desktop toast, so the wrapped row is real and not just a wider toast.
- **The producing call site is an info toast with a 15s duration**, not an undo toast: `ui/src/lib/components/GitRail.svelte:320-330` calls `toasts.info(text, { action: { label: m.gitrail_decommission_update_action() … }, duration: 15_000 })`. So the fixed branch (`{:else}` at `Toasts.svelte:32`) is the one that actually renders it, and the `.countdown` element that `duration` implies is still a direct child of `.toast` (`Toasts.svelte:54-60`) — it is `position: absolute` (`Toasts.svelte:167`), hence not a flex item, so wrapping the container cannot displace it.
- **Message keys resolve and have locale parity.** `gitrail_decommission_update_action` exists at `ui/messages/en.json:2368` and `ui/messages/de.json:2368`; `toast_merged` at `en.json:778` / `de.json:778`; `common_retry` at `en.json:6` / `de.json:6`; `common_close` (used by the `.x` aria-label) at `en.json:5` / `de.json:5`. All four keys the new test imports via `$lib/paraglide/messages` exist in both catalogs.
- **The test's `--ui-scale` reasoning checks out.** `--ui-scale: 1` and `--fs-meta: calc(11px * var(--ui-scale))` are at `ui/src/app.css:80` and `ui/src/app.css:82` (cap of 1.5 documented at `ui/src/app.css:76`), so the `expect(fontSize).toBeCloseTo(16.5, 1)` sanity assertion in `Toasts.browser.test.ts` matches the real token.
- **No out-of-diff consumer depends on the old DOM shape.** Grepping for `.toast`/`.undo`/`.x` selectors across `ui/src` and `ui/e2e` returns only `ui/src/lib/components/Toasts.browser.test.ts`; nothing else queries the toast internals, so inserting the `.actions` wrapper breaks no other test or style.
- **`flex-shrink: 0 → 1` on `.undo` is safe for the undo-tone button too.** That rule also matches the button at `Toasts.svelte:28`, which keeps `overflow: hidden` (`Toasts.svelte:143`), so a shrunk-and-unwrappable label would be clipped rather than overflow. The only undo label in the app is `m.common_undo()` (`ui/src/routes/+page.svelte:2162`) — a single short word in both locales — and with `flex-wrap` the button only shrinks when it is alone on a line and still exceeds it, which that label cannot do. No clipping path today.

### Nit (non-blocking)

Both new comments assert that grouping leaves "exactly one `margin-left: auto` on the row" (`Toasts.svelte:36-37`, `Toasts.svelte:117`), but `.undo` still carries `margin-left: auto` (`Toasts.svelte:129`) and `.x` still carries `margin-left: auto` (`Toasts.svelte:180`) — so `.actions` itself now contains two auto margins, the exact configuration the comment says was eliminated. It is inert today: `.actions` is `flex: 0 1 auto` with no `flex-grow`, so it is never wider than its content and there is no free space for the auto margins to split. But the code now contradicts its own comment, and the moment `.actions` gains any growth the ✕ would float away from the button — the Flexbox §8.1 split the comment warns about. Dropping `margin-left: auto` from both `.undo` and `.x` would make the markup match its rationale.

### Stated limitations

I did not run the browser tests (CI's job); the geometric assertions (8px row gap, 12–16px `.actions` gap, `msgLineCount === 1`) are consistent with the declared CSS — `row-gap: 8px` / `column-gap: 14px` at `Toasts.svelte:92-93` and `.actions { gap: 14px }` at `Toasts.svelte:123` — but I could not confirm the rendered pixel values without executing them.

scoop (commented):
## Task

The reported bug — a long action label (`Decommission & update local`) squeezing `.msg`, the only shrinkable child of a `nowrap` 440px row, into a 3-line column — is genuinely fixed. `.toast` now sets `flex-wrap: wrap` (ui/src/lib/components/Toasts.svelte:89) and `.msg` gets `flex: 1 1 auto` with an `auto` basis (ui/src/lib/components/Toasts.svelte:110). Flex line-breaking collects items by *hypothetical* main size (flex base clamped by min/max — `min-width: 0` does not reduce it), so the button group is wrapped away to a second row instead of the message being shrunk. Grouping the action button and ✕ into a single `.actions` flex item (ui/src/lib/components/Toasts.svelte:38-52) correctly prevents the ✕ orphaning onto a third row, since the two are now collected as one item.

Splitting `gap: 14px` into `column-gap: 14px` / `row-gap: 8px` (ui/src/lib/components/Toasts.svelte:92-93) is the right call — the shorthand would have silently applied 14px to the newly-reachable cross axis.

## Previous revision's finding — RESOLVED

> `.undo` (line 129) and `.x` (line 180) both still declare `margin-left: auto` …

Both declarations are gone. Verified against the current file: `.undo` (ui/src/lib/components/Toasts.svelte:131-135) now declares only `position/flex-shrink/min-width/max-width`, and `.x` (ui/src/lib/components/Toasts.svelte:182-183) starts at `flex-shrink: 0`. `margin-left: auto` survives in exactly one place, `.actions` (ui/src/lib/components/Toasts.svelte:124), which is what the markup comment at ui/src/lib/components/Toasts.svelte:36-37 and the CSS comment at ui/src/lib/components/Toasts.svelte:117 now truthfully describe. The Flexbox §8.1 equal-split hazard is closed.

## New finding

`.undo` is used on **two** paths: inside `.actions` for info toasts (ui/src/lib/components/Toasts.svelte:40), and as a **direct child of `.toast`** for the undo tone (ui/src/lib/components/Toasts.svelte:28). Removing `margin-left: auto` from `.undo` was correct for the first path but leaves the second unguarded once the row wraps. The CSS comment at ui/src/lib/components/Toasts.svelte:127-130 claims "`.msg`'s flex-grow right-aligns the undo-tone button" — that holds only while both sit on the *same* line. When `.msg`'s max-content exceeds the line box the message takes row 1 alone and the undo-tone button drops to row 2, where it is the only item, has no `flex-grow` and no auto margin, so `justify-content`'s default `flex-start` left-aligns it. Info toasts in the same situation stay right-aligned via `.actions`' auto margin, so the two paths now disagree.

Reachable today, not latent: the undo toast is built from `m.toast_decommissioned({ name })` = `"Decommissioned {name}"` (ui/messages/en.json:772), raised at ui/src/routes/+page.svelte:2161 with `name` a session name, which is routinely long enough to exceed the 440px cap. The new test only exercises the unwrapped case — `MERGED` = `m.toast_merged({ name: "cli-dropdown-alignment" })` at a 1280px viewport (ui/src/lib/components/Toasts.browser.test.ts:169-171) — so it passes without covering this. Cosmetic only (no clipping, no overflow), hence non-blocking; wrapping the undo-tone button in `.actions` too, or giving that one path its own auto margin, would restore parity.

## Verification performed

- Message keys used by the new test all resolve, in **both** catalogs: `common_close`, `common_retry`, `common_undo` (ui/messages/en.json:5-7 / ui/messages/de.json:5-7), `toast_merged` (ui/messages/en.json:778 / ui/messages/de.json:778), `gitrail_decommission_update_action` (ui/messages/en.json:2368 / ui/messages/de.json:2368). The diff introduces no new keys, so there is no parity gap.
- `.countdown` remains a direct child of `.toast` (ui/src/lib/components/Toasts.svelte:54-59) and is `position: absolute` (ui/src/lib/components/Toasts.svelte:170) against `.toast`'s `position: relative` (ui/src/lib/components/Toasts.svelte:85) — it is out of flow, so it adds no phantom flex item and cannot perturb the new wrapping. The comment at ui/src/lib/components/Toasts.svelte:53 is accurate.
- Test isolation: the new `describe` block adds only a `--ui-scale` cleanup, but the file-level `beforeEach`/`afterEach` (ui/src/lib/components/Toasts.browser.test.ts:20-28) already clear `toasts.items`, reset `document.body.innerHTML`, and restore the viewport — so the bare `document.querySelector` helpers cannot pick up a stale toast from a prior test.
- `.undo`'s new `flex-shrink: 1` + `min-width: 0` are safe on the undo-tone path: its label is `m.common_undo()` = `"undo"` (ui/messages/en.json:7), far too short to shrink into the `overflow: hidden` (ui/src/lib/components/Toasts.svelte:146) clip that would threaten the `.bar` countdown.
- The mobile block (ui/src/lib/components/Toasts.svelte:220-293) adds no `.actions`/`.msg` overrides that would conflict with the new wrapping.

scoop (commented):
## Verdict

The task (long action label "Decommission & update local" squeezing `.msg` into a 3-line column) is satisfied, and the single prior-revision finding is genuinely addressed.

### Prior finding: RESOLVED

> the undo-tone button (line 28) is a direct child of `.toast`, not of `.actions` … Wrap the undo-tone button in `.actions` as well … and cover the wrapped undo toast in the browser test.

- The undo-tone branch now renders `<div class="actions">` around the button (ui/src/lib/components/Toasts.svelte:32-37), so both tones route their controls through the one element that carries `margin-left: auto` (ui/src/lib/components/Toasts.svelte:126-132). `.undo` and `.x` no longer declare an auto margin (ui/src/lib/components/Toasts.svelte:137-141, 188-189), so there is exactly one auto margin on the row — the equal-split case from Flexbox §8.1 cannot arise.
- The wrapped undo case is now exercised: ui/src/lib/components/Toasts.browser.test.ts:187-200 renders `m.toast_decommissioned({ name: "shepherd/toast-decommission-update-local" })` via `toasts.undo(...)`, asserts the button actually dropped below `.msg`, and asserts `insetRight ≈ 13px` (12px `padding` + 1px `border` per ui/src/lib/components/Toasts.svelte:101-103). The unwrapped case is separately pinned at ui/src/lib/components/Toasts.browser.test.ts:181-185.

### Verification performed

- **Layout reasoning.** `.msg` gets `flex: 1 1 auto; min-width: 0; overflow-wrap: anywhere` (ui/src/lib/components/Toasts.svelte:115-122). With `flex-basis: auto` and `width: auto`, its flex base size is the max-content size, so flex line-breaking (`flex-wrap: wrap`, ui/src/lib/components/Toasts.svelte:95) pushes `.actions` to a second line rather than shrinking `.msg` — which is exactly the reported bug's mechanism reversed. `overflow-wrap: anywhere` (not `break-word`) is the right choice here because it also lowers the min-content contribution, which is what lets a single unbreakable branch name shrink instead of overflowing the panel; covered at ui/src/lib/components/Toasts.browser.test.ts:148-154.
- **Auto-margin vs. flex-grow interaction.** On an unwrapped row `.msg` consumes the free space via `flex-grow`, so `.actions`' auto margin resolves to 0 and it still sits flush right; on a wrapped row `.actions` is alone with positive free space, so the auto margin does the alignment. Both states are asserted (ui/src/lib/components/Toasts.browser.test.ts:156-162, 181-200).
- **`gap` shorthand split.** `column-gap: 14px` / `row-gap: 8px` (ui/src/lib/components/Toasts.svelte:98-99) correctly avoids inheriting 14px as the new row gap; asserted at ui/src/lib/components/Toasts.browser.test.ts:139-146, and the 14px column gap between `.undo` and `.x` at ui/src/lib/components/Toasts.browser.test.ts:134-136.
- **`.countdown` still a direct child of `.toast`.** It is `position: absolute` against the `position: relative` `.toast` (ui/src/lib/components/Toasts.svelte:91, 175-184); it was not moved into `.actions` (ui/src/lib/components/Toasts.svelte:60-66) and is out of flow, so it does not participate in the new wrapping. Structurally pinned at ui/src/lib/components/Toasts.browser.test.ts:210-219.
- **Store API agreement.** `toasts.info(text, { sticky, duration, action: { label, run } })` and `toasts.undo(text, { undoLabel, onCommit })` used by the tests match `InfoOpts`/`UndoOpts` in ui/src/lib/toasts.svelte.ts:37-71. Undo toasts have no close button by design (the store attaches no `close` path to them), consistent with ui/src/lib/components/Toasts.svelte:31-37.
- **Real call sites.** The regressing toast is the info toast at ui/src/lib/components/GitRail.svelte:320-325 (`label: m.gitrail_decommission_update_action()`), and the undo toast is raised at ui/src/routes/+page.svelte:2161-2163 with a session name — both shapes are what the new tests render.
- **i18n parity.** Every message fn the test imports exists in BOTH catalogs: `common_close` (ui/messages/en.json:5 / ui/messages/de.json:5), `common_retry` (en.json:6 / de.json:6), `common_undo` (en.json:7 / de.json:7), `toast_decommissioned` (en.json:772 / de.json:772), `toast_merged` (en.json:778 / de.json:778), `gitrail_decommission_update_action` (en.json:2368 / de.json:2368). No new keys are added by this PR.
- **`--ui-scale` claim.** `--fs-meta: calc(11px * var(--ui-scale))` at ui/src/app.css:82 with the base `--ui-scale: 1` at ui/src/app.css:80 and the 1.5 cap documented at ui/src/app.css:76 — so the 16.5px assertion at ui/src/lib/components/Toasts.browser.test.ts:247 is consistent with the tokens. `.undo`'s new `flex-shrink: 1; min-width: 0` is what keeps the ✕ beside it in that state rather than overflowing.
- **No other consumers.** Grepping ui/src for selectors depending on the old `.toast > .undo` / `.toast > .x` structure found none outside Toasts.browser.test.ts itself, so introducing the `.actions` wrapper breaks no existing query or style.

Nothing to block on. Also nothing worth raising as a nit — the CSS comments accurately describe the mechanism they claim (I checked each against the rule it sits above).

PR #1751, opened 2026-07-14
feat(ui): add decommission to command bar

scoop:
Adds a **Decommission** verb to the ⌘K command bar. It targets the currently-selected session (the label names it, e.g. `Decommission TASK-07`) and is guarded by a two-step arm on top of the existing undo window.

## Behavior

- **Two-step arm.** First Enter/click swaps the row to a red `Confirm decommission?` and keeps the bar open; a second, separate activation fires the same `onarchive(id)` path the row swipe / card menu already use — so the deferred-undo toast still guards the irreversible worktree removal. The arm clears after 3s, on typing, and on ↑/↓.
- **Availability.** Offered only where the target is actually on screen: hidden with no selection, in the **Done** lens (which has its own `doneSelectedId` selection space, so `selectedId` there is a stale leftover from another lens), and in the **panel-only** lenses `next`/`rundown`/`owed` (`shownSessions` returns `[]`, so the selected session is invisible in both panes). A destructive verb must never name a target the operator can't see.

## Three details worth reviewing

1. **Auto-repeat.** There are *three* activation paths — the combobox `onKey`, `onWindowKeydown` (Alt+digit), and the row's own `onkeydown` (Enter *and* Space), which is live because clicking a `tabindex="-1"` row focuses it and `use:dialog` never restores focus to the input. Without a guard, *holding* Enter would arm and fire within ~30ms and defeat the whole thing. The rejection therefore lives inside `selectOption(row, secondary, repeat)`, covering all paths by construction. `onWindowKeydown` still updates `altHeld` and `preventDefault()`s Alt+digit unconditionally — an early return there would let a held Alt+1 type `¡` into the search field. The 300ms dwell is **not** a key-repeat defense (the OS repeat delay is ~500ms); it guards double-click/double-fire only.
2. **Cursor sync.** Rows carry no hover handler, so a click never moves `activeIdx` — arming by click and then pressing Enter would have fired `options[activeIdx]`, a *different, unarmed* command. Arming sets `activeIdx = row.oid`. The test asserts the cursor moved and drives the confirming Enter from the input, so a regression here fails rather than passing incidentally through the row's own handler.
3. **Announcement.** Rows are `aria-activedescendant` options that never take focus, so mutating the active row's text isn't reliably announced. The armed state goes through a dedicated visually-hidden polite region (the same `sr-only` + `role="status"` pattern `TaskIdButton` / `RepoSwitcher` use) speaking `Confirm decommission of {desig}?` — naming the target the short visible label drops. It is deliberately *not* routed through the visible `.cb-empty` line, which would duplicate the row's copy and shift the list.

Disarm is imperative rather than an `$effect`, since arming *writes* `activeIdx` and an effect watching it would clear the arm on the spot.

## Verification

`bun run check`, `check:i18n`, the full UI suite (3701 tests) and root `bun run lint` pass. `selectOption` tripped fallow's cognitive-complexity bar once the arm branch landed, so the gate logic is extracted into `armGate()`; `fallow audit` is clean. Docs (`keyboard-shortcuts.md`) updated — `Enter` now reads as "activate", with the two-press rule for destructive verbs — plus EN/DE catalog keys and a feature-catalog entry.


scoop (commented):
## What I verified

**The verb is wired end to end.** `buildCommands` gains a `decommission` entry gated on `desig !== null` (`ui/src/lib/command-registry.ts:173`), and the only caller supplies it (`ui/src/routes/+page.svelte:1813` — grep for `buildCommands|CommandCtx` returns exactly one production call site, so the two new required `CommandCtx` fields break no other consumer). `onDecommission` closes over `selectedId` and calls `onarchive` (`ui/src/routes/+page.svelte:1822-1824`), which is the same deferred-with-undo path the herd cards use (`ui/src/routes/+page.svelte:2646`, `+page.svelte:2164-2193`). `Session.desig` is a non-optional `string` (`ui/src/lib/types.ts:968`), so `selected?.desig ?? null` types as `string | null` as declared (`ui/src/lib/command-registry.ts:153`).

**Availability gate matches the docs.** `decommissionDesig` is null in the Done lens and in the panel-only lenses (`ui/src/routes/+page.svelte:1830`, with `panelOnlyLens` defined at `+page.svelte:1363-1365` as rundown/owed/next), which is exactly the "offered whenever that session is on screen" claim in `docs-site/src/content/docs/reference/keyboard-shortcuts.md:13`.

**i18n parity holds.** All six new keys exist in both catalogs with the same `{desig}` placeholder and no duplicates: `ui/messages/en.json:2622-2627` and `ui/messages/de.json:2622-2627`.

**Feature-catalog contract holds.** `id: "decommission-command"` is unique across `ui/src/lib/feature-announcements/entries/` (the closest existing ids are `desktop-decommission`, `decommission-on-merge`, `decommission-and-update-local`); the filename `v1.44.0-decommission-command.ts` agrees with `sinceVersion: "1.44.0"`, which is the next unreleased version (`package.json:3` is `1.43.0`) — the rule stated at `ui/src/lib/feature-announcements.ts:11-17`. `titleKey`/`bodyKey` both resolve (en.json:2626-2627, de.json:2626-2627).

**The two-step arm is sound on the paths that exist.** `selectOption` rejects `e.repeat` before the gate (`CommandBar.svelte:440`), and `repeat` is forwarded from all three keyboard entry points — input Enter (`:483`), row Enter/Space (`:544-546`), Alt+digit (`:508`) — so a held key can neither arm nor confirm. The arm is cleared on query change (`:521-524`) and on ArrowUp/Down (`:469`, `:475`), and the timer is torn down on unmount (`:414`). The bar is mounted under `{#if showCommandBar}` (`ui/src/lib/components/page/AppOverlays.svelte:592`), so `armedId` cannot survive a close/reopen. The `.cb-row.armed` background does win over `.cb-row.kbd-active`: equal specificity, and `.armed` (`CommandBar.svelte:772`) comes after `.kbd-active` (`:765`) in source order — the comment's claim checks out.

**Not a defect:** the command-bar path calls `onarchive(id)` without the `reap` list that `Viewport.confirmDecommission` collects via its leftovers dialog (`ui/src/lib/components/Viewport.svelte:889-899`). The herd-card path already omits it too (`+page.svelte:2646`), and the server reaps orphans under the worktree unconditionally on removal (`src/worktree.ts:540-543`), so no cleanup is lost.

## Findings

The two below share one root: the arm is an *identity* (`armedId`) while the cursor is an *index* (`activeIdx`), and the option list is live — the component itself acknowledges that a WS poll can resize `options` under the cursor (`CommandBar.svelte:344-347`). If a session appears/disappears while the row is armed, every `oid` shifts, so `activeIdx` (pinned at `CommandBar.svelte:424`) no longer points at the armed row. The consequence is not destructive — the stale index lands on some other, non-destructive row, and a confirm Enter runs that instead — but the red row keeps claiming to be armed while the cursor is elsewhere, and the ring vanishes (see below).

- ui/src/lib/components/CommandBar.svelte:419-432 — `armGate` keys the arm on the command id alone, not on the target. If `decommissionDesig` ever changes while the row is armed, the row keeps its arm and shows only the target-free `"Confirm decommission?"` copy (`:561`), so the next Enter would decommission a session the operator never named. Today `selectedId` only changes via in-page flows the modal bar blocks (`+page.svelte:743`, `:1299`, `:2169`, `:2350`), so this is not reachable now; if a background retarget is ever added, disarm when the row's label changes.

scoop (commented):
## Prior findings — both addressed

**1. `.cb-row.armed` only overrode `outline-color`** — fixed. `ui/src/lib/components/CommandBar.svelte:782-787` now declares the whole ring (`outline: 1.5px solid var(--color-red)` + `outline-offset: -1.5px`) plus `color` and a tinted background, so the arm no longer depends on `.kbd-active` (`:773-777`) being co-applied. Cascade checked: `.cb-row.armed` (0,2,0) is declared after both `.cb-row:hover` (`:769`) and `.cb-row.kbd-active` (`:773`) at equal specificity, so it wins for `background` and `outline` in every combination. `--color-red` resolves — `ui/src/app.css:23`.

**2. The cursor could drift off the armed row when `oid` was re-assigned** — fixed. The clamp effect at `ui/src/lib/components/CommandBar.svelte:350-355` now re-anchors (`if (armedRow.oid !== activeIdx) activeIdx = armedRow.oid`) and disarms when the armed command has left the list (`else if (armedId !== null) disarm()`). Both branches are guarded and write only values they then read back unchanged, so the effect converges rather than looping. `armedRow` (`:411-413`) is resolved by command id, not by index, so it survives the `oid` re-assignment in `groups` (`:328-333`) that caused the original bug. `ui/src/lib/components/CommandBar.browser.test.ts:379-460` exercises exactly this (WS session insert shifting `oid`, and the command disappearing/reappearing).

## What I verified

- **i18n parity.** All six new keys exist in both catalogs and only those two exist (`ls ui/messages/` → `de.json`, `en.json`): `commandbar_cmd_decommission`, `_kw`, `_confirm`, `_confirm_aria`, `feat_decommission_command_title`, `feat_decommission_command_body` — `ui/messages/en.json:2622-2627` and `ui/messages/de.json:2622-2627`. The `{desig}` placeholder is spelled identically in both `commandbar_cmd_decommission` and `commandbar_cmd_decommission_confirm_aria` in each locale, matching the call sites `ui/src/lib/command-registry.ts:176,179`.
- **Feature announcement.** `ui/src/lib/feature-announcements/entries/v1.44.0-decommission-command.ts` needs no registration — the catalog globs the directory eagerly (`ui/src/lib/feature-announcements.ts:47-50`) — and its filename/`sinceVersion` (1.44.0) match the other unreleased entries in that directory, satisfying the contract documented at `ui/src/lib/feature-announcements.ts:3-23`. `titleKey`/`bodyKey` resolve to the keys added above.
- **Contract change.** `CommandCtx` gains two required fields (`ui/src/lib/command-registry.ts:148,153`); the only non-test caller is `ui/src/routes/+page.svelte:1813`, which supplies both (`:1822-1830`). `Command.confirmLabel`/`confirmAria` are optional, so no existing `Command` producer breaks; `CommandBar` maps them to `?? null` at `ui/src/lib/components/CommandBar.svelte:281-282`.
- **The target really is on screen where the verb is offered.** `decommissionDesig` is nulled for `done` and the panel-only lenses (`ui/src/routes/+page.svelte:1830`); I confirmed the main pane in those lenses renders `panelMainHint` / `DoneRecapPanel` instead of the selected session's Viewport (`ui/src/routes/+page.svelte:2863-2873`), and that `panelOnlyLens` is exactly `rundown|owed|next` (`:1363-1365`). Under a repo/status filter or the Ready lens the selected session can be absent from the rail but is still rendered in the main pane (`herdSessions` at `:417-426` filters only the rail), so naming it there is correct. Mobile can't reach the bar at all: the only entry point is `TopBarSearch`, gated `{#if !mobile}` (`ui/src/lib/components/TopBar.svelte:816-818`).
- **`run()` targets the named session.** `onDecommission` reads the live `selectedId` (`ui/src/routes/+page.svelte:1822-1824`) and `decommissionDesig` reads `selected?.desig` from the same id (`:717`). Selection cannot change behind a modal command bar (every `selectedId` write is user-driven or an archive/clear-merged consequence), and if the store drops the session, `selected` goes null → the command disappears → the re-anchor effect disarms.
- **Two-step gate.** `armGate` (`ui/src/lib/components/CommandBar.svelte:427-440`) is reached from *every* activation path because it sits inside `selectOption` (`:446-448`), which is the single entry for input-Enter (`:491`), row-Enter/Space (`:648-650`), row-click (`:641`) and Alt+digit (`:616`). `e.repeat` is forwarded on all four keyboard paths, so a held key can neither arm nor confirm. `onarchive` still routes through the undo toast (`ui/src/routes/+page.svelte:2164-2176`), so the docs' "undo window still applies" claim holds.

## Note (not a defect)

Decommissioning from the bar calls `onarchive(id)` directly, skipping the `getLeftovers` probe + `LeftoverDialog` that the Viewport button uses (`ui/src/lib/components/Viewport.svelte:889-899`), so leftover subprocesses aren't offered for reaping. That matches the Herd card-menu path, which already wires `ondecommission={onarchive}` (`ui/src/routes/+page.svelte:2646`), so it's consistent with an existing affordance rather than a regression — flagging only in case reaping was meant to be part of this verb.

## Nit

See findings: arming by click moves DOM focus onto the option row, which strands the keyboard.

scoop (commented):

**Addressed.** `armGate` now calls `inputEl?.focus({ preventScroll: true })` on the arming branch (ui/src/lib/components/CommandBar.svelte:439), after pinning the cursor to the armed row (:432). With focus back on the combobox, the two documented escape hatches work on the click path: typing reaches the input's `oninput`, which calls `disarm()` (:577-580), and ArrowUp/Down reach `onKey`, which also disarms (:484, :490). The `aria-activedescendant` contract (:576) is restored — focus stays on the input, the cursor stays virtual. The PR's own tests were flipped to assert exactly this (ui/src/lib/components/CommandBar.browser.test.ts:463-501: `expect(document.activeElement).toBe(input.element())`, then disarm-on-type and disarm-on-arrow, plus "still confirms on the Enter that follows a click-arm"). The docs' claim that the arm "clears itself … as soon as you type or move the cursor" (docs-site/src/content/docs/reference/keyboard-shortcuts.md:15) is now true on both paths.

## What I verified

- **i18n parity** — all six new keys exist in both catalogs with the same `{desig}` placeholder: `commandbar_cmd_decommission`, `_kw`, `_confirm`, `_confirm_aria`, `feat_decommission_command_title/_body` at ui/messages/en.json:2622-2627 and ui/messages/de.json:2622-2627.
- **Registry/consumer agreement** — `Command.confirmLabel/confirmAria` (ui/src/lib/command-registry.ts:21-30) are read by CommandBar's command-row mapping (ui/src/lib/components/CommandBar.svelte:281-282) and rendered at :642-650 / :670-677; `CommandCtx.onDecommission` + `decommissionDesig` (:47-53) are supplied at ui/src/routes/+page.svelte:1822-1830. `selected` resolves against `store.sessions` (+page.svelte:717) and `Session.desig` exists (ui/src/lib/types.ts:968).
- **Availability gate** — `desig !== null` (command-registry.ts:173) and the page's Done/panel-only exclusion (+page.svelte:1830, `panelOnlyLens` defined at :1363-1365) mean the verb is hidden exactly when the target isn't rendered; `selectedId` is never re-pointed while the bar is open (the only writers are +page.svelte:745, :1299, :2169, :2175, :2350, all user actions outside the dialog), so an armed row cannot silently re-target.
- **Feature announcement** — `sinceVersion: "1.44.0"` is ahead of the last release (package.json:3 = 1.43.0), the filename matches `v<sinceVersion>-<id>.ts`, and entries are auto-registered by the `import.meta.glob` in ui/src/lib/feature-announcements.ts:48-51, so no index edit is needed.
- **Styling/a11y** — `--color-red` exists (ui/src/app.css:23) and `.cb-row.armed` mirrors Viewport's `.decom.armed` (ui/src/lib/components/Viewport.svelte:3616-3620); the `.armed` rule is declared after `.kbd-active` at equal specificity, so the red ring wins. The comment's claim that there is no `.sr-only` util in app.css checks out (no match in ui/src/app.css; the pattern matches TaskIdButton.svelte / RepoSwitcher.svelte).
- **Arm state machine** — the re-anchor effect (:350-355) converges (writing `armedId = null` makes `armedRow` null and the branch untaken; writing `activeIdx` to the value it compares against is idempotent), and `repeat` is forwarded from all three activation paths (:498 input Enter, :609-610 row Enter/Space, :521 Alt+digit).

## Findings

**1. The ⌘K verb bypasses the leftover-process probe that the on-screen Decommission button runs.** Viewport's decommission goes through `confirmDecommission` → `getLeftovers(id)` → `LeftoverDialog`, and only reaches `onarchive` after the operator has seen (and chosen whether to reap) still-running subprocesses — dev servers, `tailscale serve`, etc. (ui/src/lib/components/Viewport.svelte:889-899, :3076-3092; `Leftover` at ui/src/lib/types.ts:2099-2106). The command-bar wiring calls `onarchive(selectedId)` directly (ui/src/routes/+page.svelte:1822-1824), so the same verb from ⌘K silently takes the no-reap path: the operator is never told the session still has processes running, and they keep running after the worktree goes. Not a crash — dismissing the dialog produces the same call (Viewport.svelte:3083) — but it is a behavioural divergence between two spellings of one verb, and it is invisible to the operator. Either route the command-bar path through the same probe or state in the docs that ⌘K decommission never reaps.

**2. Two comments now contradict the code they sit on.** ui/src/lib/components/CommandBar.svelte:605-607 still says "Clicking a `tabindex="-1"` row focuses it and use:dialog never restores focus to the input, so after an arm-by-click every key lands HERE, not on onKey" — but `armGate` now explicitly pulls focus back to the input (:439), so keys after an arm-by-click land on `onKey`, and the row's own handler is effectively dead for a destructive row. The same claim is repeated in the test comment at ui/src/lib/components/CommandBar.browser.test.ts:313-315 ("after an arm-by-click the row (not the input) is the focused element receiving the keys"), which the sibling test at :475 directly refutes. Keeping the repeat-forwarding on the row handler is fine as defence in depth; the justification needs rewording.

## Notes (informational)

- The arm is keyed on the command **id**, not on the target session, so in principle an armed row could re-target if `selectedId` changed underneath it. I traced every `selectedId` writer (listed above) and none can fire while the focus-trapped bar is open, so this is not reachable today; if a future change lets the page re-select behind the bar, the arm key should incorporate the desig.

scoop (commented):

**1. ⌘K decommission skipped the leftover probe — ADDRESSED.**
`+page.svelte:2203-2210` now runs `commandDecommission(id)`, which awaits `getLeftovers(id).catch(() => [])`, falls straight through to `onarchive(id)` on an empty (or failed) probe, and otherwise parks the result in `decomLeftovers` / `decomLeftoverTarget`. This is the same shape as Viewport's `confirmDecommission` (verified against ui/src/lib/components/Viewport.svelte:889-899): identical probe, identical swallow-the-error-and-close semantics. The dialog side matches too — `finishCommandDecommission()` (`+page.svelte:2212-2218`) archives with no reap on dismiss and with `keys` on confirm, mirroring Viewport.svelte:3076-3092. `getLeftovers` resolves to ui/src/lib/api.ts:377 (`(id: string) => Promise<Leftover[]>`), `Leftover` to ui/src/lib/types.ts:2099, and `onarchive(id, reap?: string[])` at +page.svelte:2166 already forwards `reap` to `archiveSession(id, reap)` (:2184) — the new caller agrees with that signature.

**2. Stale focus comments — ADDRESSED.**
The row `onkeydown` comment (CommandBar.svelte:600-606) now says armGate pulls focus back and justifies the repeat-forwarding as defence in depth for rows that hold focus from any other click. Grepping CommandBar.browser.test.ts for the old claim ("never restores focus" / "lands HERE" / `use:dialog`) returns nothing, and the reworded test comment sits at :812-816 alongside the refocus assertion — the contradiction is gone.

## Verification of the new work

- **i18n parity**: both new command strings, both confirm strings and both announcement keys exist in ui/messages/en.json:2622-2627 and ui/messages/de.json:2622-2627, with the same `{desig}` placeholder in the two parameterized keys. `ls ui/messages/` shows only `de.json` + `en.json`, so parity is complete.
- **Feature announcement**: `sinceVersion: "1.44.0"` matches the other unreleased entries (ui/src/lib/feature-announcements/entries/v1.44.0-*.ts) against package.json:3 (`1.43.0`), and registration is automatic via the `import.meta.glob("./feature-announcements/entries/*.ts")` at ui/src/lib/feature-announcements.ts:47-48 — no index edit needed. `titleKey`/`bodyKey` resolve to the two `feat_decommission_command_*` keys above.
- **`CommandCtx` grew two required fields**; the only non-test consumer is +page.svelte:1815 (grep for `buildCommands`/`CommandCtx`), and it supplies both. Same for `AppOverlays`' three new required props — +page.svelte:3192-3194 is the only render site.
- **Availability guard**: `decommissionDesig` (+page.svelte:1832) nulls out on the `done` lens and on `panelOnlyLens` (defined at :1365 as rundown/owed/next). Under every other lens the main pane renders `<Viewport session={selected}>` (+page.svelte:2921) regardless of the rail's filter, so the target really is on screen — the guard is neither too tight nor too loose.
- **Armed-row re-anchoring**: `oid` is a running counter across the flattened group rows (CommandBar.svelte:332) and `options` is that same flattening (:336), so `oid` is exactly the index into `options` that `activeIdx` addresses — the `activeIdx = armedRow.oid` re-anchor at :350 is correct, and the effect converges (the second pass finds `oid === activeIdx` and writes nothing).
- **Arm gate**: `repeat` is forwarded from all three keyboard paths (input Enter :498, row Enter/Space :607-609, Alt+digit :523) and short-circuits before `armGate`, so a held key can neither arm nor confirm; `ARM_DWELL_MS` covers double-fire; arming pins `activeIdx` and refocuses the input so the next Enter lands on the armed row. Disarm fires on typing (:577), arrow moves (:484,:490), timeout, and on the armed row leaving the list (:351).
- **Global-shortcut suppression**: the new `LeftoverDialog` is picked up by `anyOverlayOpen()`'s `document.querySelector(".overlay, .drawer")` check (+page.svelte:1533) because LeftoverDialog's root carries `class="overlay"` (ui/src/lib/components/LeftoverDialog.svelte:31), so ⌘K and the letter shortcuts stay suppressed while it is up.

Nothing here blocks. One nit below.

scoop (commented):
## Re-review

**Prior finding — resolved.** `ui/src/lib/components/CommandBar.svelte:411` now reads `const armedRow = $derived(options.find((o) => o.kind === "command" && o.id === armedId) ?? null);` — the `as OptRow | null` cast is gone, and `options` is still typed `OptRow[]` (CommandBar.svelte:336), so the inferred type is `OptRow | null` as intended.

## Verification of the new revision

Identifiers resolved against ground truth:
- `getLeftovers(id): Promise<Leftover[]>` exists at `ui/src/lib/api.ts:377`; `Leftover` is exported from `ui/src/lib/types.ts:2099`. `Session.desig` exists (`ui/src/lib/types.ts:968`), so `selected?.desig` in `+page.svelte:1832` is well-typed.
- `onarchive(id: string, reap?: string[])` at `ui/src/routes/+page.svelte:2166` — `finishCommandDecommission` calls it as `onarchive(target, reap)` (`+page.svelte:2217`), matching the signature, and the deferred/undo behavior of the existing button path is preserved.
- The new ⌘K path mirrors Viewport's button exactly: probe → `[]` (or a caught probe failure) falls through to `onarchive(id)`; a non-empty probe pops `LeftoverDialog`. Compared against `ui/src/lib/components/Viewport.svelte:892-898` and `Viewport.svelte:3076-3090`; `onclose` archives without reaping and `onconfirm` archives with the picked keys in both. `LeftoverDialog`'s props (`leftovers`/`onclose`/`onconfirm(keys)`) at `ui/src/lib/components/LeftoverDialog.svelte:7-17` line up with what `DecomLeftovers.svelte` forwards.
- `--color-red` is defined at `ui/src/app.css:23`, and `.cb-row.armed` reuses the same treatment as `Viewport.svelte:3616-3620`. There is indeed no `.sr-only` utility in `app.css` (grep found none); the local copy follows the `TaskIdButton.svelte:83` `sr-only` + `role="status"` precedent.

i18n parity: all four command strings plus the two announcement keys exist in **both** catalogs with the same `{desig}` parameter — `ui/messages/en.json:2622-2627` and `ui/messages/de.json:2622-2627`. The announcement entry `ui/src/lib/feature-announcements/entries/v1.44.0-decommission-command.ts` uses `titleKey`/`bodyKey` that both catalogs carry, its filename version matches `sinceVersion` (`1.44.0`, the same unreleased version the other pending entries use, e.g. `v1.44.0-review-job-model.ts`), and the catalog is glob-loaded (`ui/src/lib/feature-announcements.ts:47-50`), so no registration step is missing.

Arm-gate correctness (`CommandBar.svelte:394-470`), walked through the paths that can reach it:
- `oid` is assigned sequentially over the same flat order `options` is built from (`CommandBar.svelte:328-336`), so `options[armedRow.oid] === armedRow`; pinning `activeIdx` to `armedRow.oid` in the effect at `:350-355` keeps the input's Enter (`:493-496`) on the row that is actually armed after a live re-derive. When the armed row leaves the list entirely, the effect disarms; both branches are guarded, so the effect converges.
- Every activation path forwards `e.repeat` (input Enter `:496`, row Enter/Space `:602-607`, Alt+digit `:521`), and the check sits in `selectOption` (`:453`) rather than at the call sites, so a held key can neither arm nor confirm.
- Arming pulls focus back to the input (`:437`), which keeps disarm-on-type (`:575-578`) and disarm-on-arrow (`:482,:488`) reachable after a click-arm on a `tabindex="-1"` row.
- The bar is mounted inside `{#if showCommandBar}` (`AppOverlays.svelte:604`), so `armedId` cannot survive a close/reopen, and `⌘K` bails on `anyOverlayOpen()` (`+page.svelte:1578`), whose DOM probe covers `LeftoverDialog`'s `.overlay` — so `decomLeftoverTarget` can't be clobbered by a second command-bar decommission while the dialog is up.

Availability gating: `decommissionDesig` is null on the Done lens and the panel-only lenses (`+page.svelte:1832`, with `panelOnlyLens` = rundown/owed/next at `:1365-1367`). For every other lens with sessions present, the main pane renders `Viewport session={selected}` (`+page.svelte:2919-2922`), so the named target really is on screen — the comment's claim holds.

The docs change (`docs-site/src/content/docs/reference/keyboard-shortcuts.md`) matches the implemented behavior: 3s arm (`ARM_MS`), cleared on type or cursor move, undo window still applies via `onarchive`.

Nothing blocking, and nothing worth raising as a nit.

Issue #855, opened 2026-06-19
Refactor oversized Svelte component templates; tighten/remove fallow <template> grandfathers (follow-up to #851)


scoop:
## Progress — what's left (updated 2026-06-21)

**Worst-first paydown: 11 of 13 grandfathers removed** (10 merged + `+page` in open PR #915). Each shrunk component clears the Tier-1 40/60 bar; `fallow health` stays at 0 template-complexity findings.

**Done — grandfather removed:**
- [x] Settings, AutomationSettings, NewTask, RepoSwitcher — PR #868
- [x] IssuesPanel, BacklogView — PR #874
- [x] UnitRow, GitRail — PR #893
- [x] TopBar (119/191 → 16/14, 8 children under `top-bar/`) — PR #898
- [x] Herd (103/154 → children under `herd/`) — PR #911
- [x] +page.svelte (78/96 → 32/45, extracted `page/AppOverlays`) — PR #915 *(open, awaiting merge)*

**Remaining (2 grandfathers):**
- [ ] LearningsDrawer.svelte (63/157, grandfather 70/160) — only un-touched giant; self-contained drawer, but cognitive 157 is ~2.6× the 60 bar, so likely needs aggressive splitting of the nested card `{#each}` to clear (else tighten-and-track)
- [ ] Viewport.svelte — tightened 136/194 → 80/121 (grandfather 90/130) in PR #858; residual `.vp-head` identity/rename + status zone is deeply nested and can't clear 60 cog without unnatural splits, so it's **tighten-and-tracked** rather than fully removable

Running detail is in the progress comments below.

---

## Context

Follow-up to #851, which adopted fallow's synthetic Svelte `<template>` complexity metric (fallow 2.98+) via `health.thresholdOverrides` in `.fallowrc.jsonc`. That PR set a Tier-1 global bar (40 cyclomatic / 60 cognitive) plus **per-file grandfathers** for the 13 components whose templates exceed it. The grandfathers are documented debt — this issue is to pay it down.

A template aggregates many conditional regions, so a large single-file component scores very high on the synthetic metric. Splitting these into smaller subcomponents reduces real template complexity and lets us **tighten or remove** each grandfather in `.fallowrc.jsonc`.

## Scope

Refactor/split the oversized component templates, worst-first, and tighten the matching Tier-2 entry in `.fallowrc.jsonc` as each shrinks. Current `<template>` cyclomatic / cognitive (HEAD of #851):

| Component | cyclo | cog |
[...]

PR #1756, opened 2026-07-14
feat(ui): global toggle to hide info tooltips

scoop:
Settings → Device gains a **Hide info tooltips** switch, default **off**. When on, the three *opt-in help affordances* disappear app-wide:

- the circular **(i)** `InfoTip` buttons (9 sites),
- the AutomationSettings **ⓘ** row-explainers (11 sites),
- the dashed-underline **glossary terms** (via `GlossaryTerm`, which also covers `GlossaryText` and `InferredBadge`).

Each family has exactly one choke point, so there is no call-site churn.

## Scope: what it deliberately does *not* hide

Tooltips that attach to **content** rather than to an (i) affordance are untouched: the `statusTip` action (StatusPip, the Autopilot/BuildQueue/Critic/PlanGate/Research badges, UnitRowRight), the `Stepper`/`HeartbeatStrip` legends, and native `title=` hints.

The distinction is the whole point. The three in-scope systems are help *chrome* — added next to content purely to explain it; strip them and the content still reads. The fourth family is the inverse: a `StatusPip` is a coloured dot and a `CriticBadge` is a terse chip, so the tooltip is the *only* decoding of an opaque glyph. Hiding those wouldn't declutter the UI, it would make it unreadable.

The hint copy states this explicitly, so an operator who enables the toggle and still sees a `StatusPip` tooltip reads it as designed rather than as a bug.

## Persistence: device pref, not a server setting

Deliberate, after weighing the cost. A server-backed setting would have needed a config field, boot rehydrate, `PUT` handler, server tests, a third `Settings` copy in demo mode, an `AbortController` on `getSettings()` (a bare fetch today — a *hung* request never settles, which would leave the app blank behind a first-paint gate), reconciliation with `+page.svelte`'s cached `settings`, a demo no-op `PUT` that snaps the toggle back, and a docs entry.

**Trade accepted: the pref does not follow you to another browser.**

It also cannot flash, structurally — and needs no `app.html` pre-paint bootstrap. The app is `ssr:false`, so first paint is an empty shell with no icons in it, and the store's `localStorage` read is synchronous at module init, i.e. before the first component render.

## AutomationSettings extraction

The ⓘ and its explainer move into `AutomationInfoTip` / `AutomationDetail`. Two reasons:

1. **Correctness.** The sandbox row's explainer was hand-written markup rather than the shared `{#snippet detail}`, while its ⓘ *did* come from `{#snippet info}`. Guarding only the snippets would have removed the ⓘ and stranded two paragraphs in the DOM. One component now owns both shapes, so that divergence can't reappear.
2. **Complexity gate.** The three guards would have pushed that template from **39 → 42** against the Tier-1 `<template>` bar of **40**. Extracting keeps it at **39** — rather than re-adding a Tier-2 grandfather and regressing the #855 paydown that removed AutomationSettings' one.

An `$effect` also resets `openDetail` on hide: it is component-local state that survives the pref flip without a remount, so a row that was expanded would otherwise be left visible with no ⓘ left to close it.

## Other notes

- `/design-system` is exempted via a context key — it is the canonical component catalogue, so a specimen that vanished based on a viewer's personal pref would make the reference lie.
- For `external` glossary terms this also removes the tooltip's **Wikipedia link**, not just a hover hint. Accepted: it follows from "remove the affordance entirely", and keeping underlines alive only for external terms would be incoherent.
- `InfoTip`/`GlossaryTerm` now use the existing `--shadow-popover` token instead of inlining its raw `rgba()`. `Coachmark.svelte` still inlines the same value — pre-existing, left alone as it's outside this diff.

## Tests

The ⓘ detail blocks toggle with `hidden=`, **not** `{#if}`, so a collapsed block is already in the DOM — a naive "absent when hidden" assertion would pass **even with no guard at all**. The AutomationSettings tests therefore *expand the row first*, then hide tips, then assert **DOM absence** (not non-visibility), for both the sandbox and a snippet-driven row. They also pin the exact selectors (`button.info`, `#auto-detail-critic`, `.sandbox-detail`), which is what verified the extraction preserved DOM structure.

Also covers: the store; InfoTip renders nothing when hidden; glossary degrades to plain text with the word still in the sentence (incl. a `GlossaryText` case, which is what covers the Coachmark / What's-New consumers); the design-system exemption; the Settings switch.

**Verification:** `ui` check + 3711 tests green; root lint + 6943 tests green; `check:i18n`, `check-glossary`, `check-feature-catalog`, `check-announcement-versions`, branch hygiene, and `fallow audit` all pass.


scoop (commented):
## Verdict

The feature does what the task asks: a per-device `hide info tooltips` preference, default off, that removes the circular ⓘ affordances (`InfoTip`, `AutomationSettings`' row explainers) and the dashed-underline glossary terms, while leaving status-chip/badge tooltips alone. No blocking bug found. Findings below are nits.

## What I verified

- **Store shape / persistence** — `ui/src/lib/info-tips.svelte.ts:15-38` mirrors the existing `tab-ticker` store one-for-one (`ui/src/lib/tab-ticker.svelte.ts:4-28`): same `read()`-in-try/catch, same `set()`/`toggle()`, distinct key `shepherd:hide-info-tips`. Default is `false` when the key is absent, so "default off" holds. The comment's no-flash argument checks out: `ui/src/routes/+layout.ts:1` is `export const ssr = false`.
- **Coverage of the (i) affordances** — grepped for every `.info` button in the tree: only `ui/src/lib/components/InfoTip.svelte:117` and `ui/src/lib/components/automation-settings/AutomationInfoTip.svelte:22` render one, and both are guarded. All `<InfoTip>` call sites (FilesPanel, HerdGroup, TimelineLens, SpendLens, OverheadLens, NewTaskRunSettings) go through the guarded component, so nothing escapes the toggle.
- **Glossary path** — `GlossaryTerm.svelte:185` degrades to the pre-existing `.gloss-term-plain` span (style at `GlossaryTerm.svelte:271`), i.e. the same fail-soft output the unknown-term branch already produced; `GlossaryText` consumers (Coachmark, WhatsNew, RundownPanel, TelemetryConsent) inherit it.
- **CSS moved to the new child components** — I checked that no selector left behind in `AutomationSettings.svelte` still targets `.info` / `.auto-detail` (grep returns only a comment, see finding 1), so Svelte's style scoping does not silently break the moved rules. `--shadow-popover` (swapped in at `InfoTip.svelte:193` and `GlossaryTerm.svelte:291`) is defined at `ui/src/app.css:138` with the identical value `0 4px 16px rgba(0, 0, 0, 0.4)`, so that drive-by is a no-op visually.
- **i18n parity** — all six new keys exist in both catalogs: `ui/messages/en.json:2608-2613` and `ui/messages/de.json:2608-2613` (`settings_hide_info_tips_{title,hint,on,off}`, `feat_hide_info_tips_{title,body}`).
- **Feature-announcement entry** — `sinceVersion: "1.44.0"` matches the other unreleased entries in `ui/src/lib/feature-announcements/entries/` (e.g. `v1.44.0-terminal-font-size.ts`), the filename version agrees with the field, and the loader is an eager glob over that directory (`ui/src/lib/feature-announcements.ts:48`), so no registry edit is needed. Omitting `targetId` is legal per the type at `ui/src/lib/feature-announcements.ts:27-36`.
- **Settings row** — `SettingsDevicePanel.svelte:146-161` is structurally identical to the contrast / colourblind / tab-ticker rows above it (`:98-145`), including deriving the accessible name from the `.state` label rather than an `aria-label`.

- ui/src/lib/components/InfoTip.svelte: `open` is component-local `$state` that is not reset when `suppressed` flips true, so if a tooltip is open at the moment the preference is turned on, re-enabling tips later re-mounts it already open; the dismiss-listener `$effect` at `:52-83` also stays subscribed with `btnEl`/`popEl` null. Not reachable today (a `pointerleave`/`blur` fires before the operator can reach the Settings dialog), but `AutomationSettings` explicitly resets its sibling state (`AutomationSettings.svelte:104-106`) and this parallel path does not.
- ui/src/lib/components/automation-settings/AutomationDetail.svelte: the non-`sandbox` branch renders only `paragraphs[0]` (`:41`), silently dropping any further entries; a future two-paragraph non-sandbox row would lose text with no error.
- ui/src/lib/components/automation-settings/AutomationDetail.svelte: `{#each paragraphs as text (text)}` at `:36` keys by the paragraph text itself — two identical paragraphs would throw Svelte's duplicate-key error at runtime. Keying by index would be safe here.

scoop (commented):

**1. Orphaned `.sandbox-detail` CSS comment in `AutomationSettings.svelte` — ADDRESSED.** The `/* Sandbox ⓘ detail: recess to --color-inset… */` comment and its rules are both deleted from `AutomationSettings.svelte`; the comment now lives with the rules it documents at `ui/src/lib/components/automation-settings/AutomationDetail.svelte:57-60`, and `.epic-mode-banner`'s comment at `ui/src/lib/components/AutomationSettings.svelte:769` is no longer preceded by a dangling one.

**2. Misleading localStorage-stub comment / hoisted-import ordering in `info-tips.svelte.test.ts` — ADDRESSED.** The test now stubs with `vi.stubGlobal("localStorage", localStorageMock)` (`ui/src/lib/info-tips.svelte.test.ts:23`) and re-evaluates the module per test via `vi.resetModules()` + `await import("./info-tips.svelte")` (`ui/src/lib/info-tips.svelte.test.ts:26-29`), so the init-time `read()` at `ui/src/lib/info-tips.svelte.ts:25` genuinely observes the stub. The header comment (`:16-22`) now states the hoisting hazard correctly instead of claiming a static import would be ordered.

**3. `AutomationDetail` dropping extra paragraphs / text-keyed `{#each}` — ADDRESSED.** Props are now `text?: string` / `paragraphs?: string[]` (`ui/src/lib/components/automation-settings/AutomationDetail.svelte:26-31`), the multi-paragraph branch renders every entry, and the each block is keyed by index — `{#each paragraphs as paragraph, i (i)}` (`:41`) — so duplicate paragraph text can no longer throw. The only caller passing `paragraphs` is the sandbox row (`ui/src/lib/components/AutomationSettings.svelte:609-613`, two entries); every other row passes `text`.

## New finding

The extraction that moved `.info` and `.auto-detail` out of `AutomationSettings.svelte` deleted the rules but left their comments behind. `ui/src/lib/components/AutomationSettings.svelte:670-672` now reads:

```css
  /* clickable "ⓘ" — small circular affordance that toggles the long explanation */
  /* the revealed long-form explanation: a quiet tonal-step note (panel over the
     popover's inset ground) with a full hairline border — no accent stripe */
  .auto-desc {
```

Both comments now document `.auto-desc`, which is neither an ⓘ button nor a long-form explanation. The rules they describe live in the new components, each of which already carries its own copy of the comment (`ui/src/lib/components/automation-settings/AutomationInfoTip.svelte:34` and `.../AutomationDetail.svelte:47-48`), so these two are pure leftovers — the same class of dangle raised as prior finding 1, just for the other two blocks. Delete them.

## Verification

- `infoTips` / `INFO_TIPS_FORCE` resolve to `ui/src/lib/info-tips.svelte.ts:36` and `:42`; all consumers (`InfoTip.svelte:4`, `GlossaryTerm.svelte:6`, `SettingsDevicePanel.svelte:14`, `AutomationSettings.svelte:4`, `AutomationInfoTip.svelte:9`, `AutomationDetail.svelte:14`, `design-system/+page.svelte:19`) import the same spellings.
- Locale parity: the six new keys (`settings_hide_info_tips_title|hint|on|off`, `feat_hide_info_tips_title|body`) exist with identical names in `ui/messages/en.json:2608-2613` and `ui/messages/de.json:2608-2613`. `automation_info_aria` (used by the new `AutomationInfoTip`) exists at `ui/messages/en.json:149` and `ui/messages/de.json:149`; `automation_sandbox_profile_hint` / `_caveats` (passed as `paragraphs`) exist at `:1166-1167` in both.
- The `{@render info(…)}` / `{@render detail(…)}` snippets are fully removed and every one of the 11 call sites was converted to `<AutomationInfoTip>` + `<AutomationDetail>` with matching `id`s, so each button's `aria-controls="auto-detail-{id}"` (`AutomationInfoTip.svelte:23`) still resolves to the detail's `id="auto-detail-{id}"` (`AutomationDetail.svelte:38,45`) whenever tips are shown.
- `--shadow-popover` is defined once, at `ui/src/app.css:138`, as exactly `0 4px 16px rgba(0, 0, 0, 0.4)` — the literal it replaces in `InfoTip.svelte:193` and `GlossaryTerm.svelte:291` — with no `[data-theme="light"]` override (`ui/src/app.css:172`), so that drive-by swap is visually a no-op in both themes and matches the token's stated purpose (`ui/src/app.css:136`).
- Stale-open state: `AutomationSettings.svelte:104-106` resets `openDetail` when the pref flips. `InfoTip`'s and `GlossaryTerm`'s local `open` state is not reset, but both register a window `pointerdown` dismissal while open (`InfoTip.svelte:57-75`, `GlossaryTerm.svelte:59`), so clicking the Settings switch necessarily closes any open tooltip first — no stale `open=true` can survive into a re-enable and auto-`showPopover()`. Not a defect.
- No global stylesheet references `.auto-detail`, `.sandbox-detail` or `.info` (grepped `ui/src/lib/**/*.css`), so relocating those rules into the new components' scoped `<style>` blocks loses no styling; `.sandbox-detail`'s `margin-top: 0` (which cancels the `.drain-fields` gap stacking) is preserved at `AutomationDetail.svelte:52`.
- Announcement entry `ui/src/lib/feature-announcements/entries/v1.44.0-hide-info-tips.ts` is auto-registered via `import.meta.glob` (`ui/src/lib/feature-announcements.ts:47`); filename version matches `sinceVersion: "1.44.0"`, consistent with the other unreleased `v1.44.0-*` entries, and its `titleKey`/`bodyKey` resolve in both catalogs.
- Docs: `docs-site/src/content/docs/reference/glossary.md:10-14` is the only docs surface that enumerates this pref; no other page lists Settings → Device toggles (grepped for "Compact tab title" / "Reduced notifications"), so there is no docs parity gap.

## Notes (non-blocking, not findings)

- `AutomationInfoTip` / `AutomationDetail` deliberately do not honour `INFO_TIPS_FORCE`. That is correct today — `/design-system` renders neither (`ui/src/routes/design-system/+page.svelte` only mounts `GlossaryText` / `GlossaryTerm`) — but if `AutomationSettings` is ever added to the catalogue its ⓘ specimens will vanish for a viewer who has the pref on.
- `AutomationDetail`'s prop type permits both `text` and `paragraphs` (or neither) even though the comment says "exactly one"; a discriminated union would make that a compile-time guarantee. Purely a typing nicety — every current call site passes exactly one.

scoop (commented):
## Re-review of prior finding

1. **AutomationSettings.svelte orphaned CSS comments (lines 670-672)** — **fixed.** The `.info` and `.auto-detail` rules and both comments are gone from the diff hunk at `ui/src/lib/components/AutomationSettings.svelte@@ -587,50`, and reading the current file shows `.auto-name` (ui/src/lib/components/AutomationSettings.svelte:662-668) is now followed directly by `.auto-desc` (ui/src/lib/components/AutomationSettings.svelte:669-673) with no dangling comment. The comments live with the rules they document, in ui/src/lib/components/automation-settings/AutomationInfoTip.svelte:36 and ui/src/lib/components/automation-settings/AutomationDetail.svelte:47.

- **i18n parity**: all six new keys (`settings_hide_info_tips_title/hint/on/off`, `feat_hide_info_tips_title/body`) exist in both catalogues — ui/messages/en.json:2608-2613 and ui/messages/de.json:2608-2613 — and `ui/messages` contains only `en.json` and `de.json` (both 3134 lines). The announcement entry's `titleKey`/`bodyKey` (ui/src/lib/feature-announcements/entries/v1.44.0-hide-info-tips.ts:7-8) name exactly those keys, and the entry is picked up by the `import.meta.glob("./feature-announcements/entries/*.ts")` in ui/src/lib/feature-announcements.ts:47-48, matching its siblings' shape.
- **`--shadow-popover`** (the token that replaces the hard-coded `0 4px 16px rgba(0,0,0,0.4)` in InfoTip.svelte:193 and GlossaryTerm.svelte:291) is defined at ui/src/app.css:138 with that exact value, so this is a pure token swap with no visual change.
- **CSS extraction is complete**: grepping AutomationSettings.svelte, automation-settings/automation-fields.css and app.css finds no remaining selector that reaches into the extracted markup (`.info`, `.auto-detail`, `.sandbox-detail` are gone from the parent), so no Svelte-scoped rule is left orphaned by the child components now owning that DOM. No `{@render info(...)}`/`{@render detail(...)}` call sites survive the snippet removal.
- **Store bootstrap**: `info-tips.svelte.ts` reads `localStorage` at module init inside a `try/catch`; `ui/src/routes/+layout.ts:1` sets `export const ssr = false`, so the module comment's no-flash argument holds and the non-browser test path falls back to `false` safely.
- **Settings panel markup** reuses the existing `.rc` / `.hint` / `.toggle` block shape already used at ui/src/lib/components/settings/SettingsDevicePanel.svelte:114,130,146, and `.rc` is styled at ui/src/lib/components/settings/SettingsDevicePanel.svelte:366.
- **State teardown**: if a detail is expanded when the pref flips on, the `$effect` in AutomationSettings.svelte:103-105 clears `openDetail`, and `AutomationDetail` unmounts rather than staying `hidden`, so no content is stranded. In `InfoTip`, flipping the pref while a tooltip is open unmounts `btnEl`/`popEl`; the positioning effect (InfoTip.svelte:40-48) guards on both being non-null and its `anchorPopover` cleanup runs on teardown, so no null-deref.

- ui/src/lib/components/automation-settings/AutomationInfoTip.svelte / AutomationDetail.svelte read `infoTips.hidden` directly and do not honour the `INFO_TIPS_FORCE` context that `InfoTip.svelte:25` and `GlossaryTerm.svelte:11` consult; this is inert today because `/design-system` (ui/src/routes/design-system/+page.svelte) does not render an AutomationSettings specimen, but if one is ever added the catalogue would silently drop its ⓘ for operators who hid tips — the exact failure the force-context exists to prevent.

Issue #1422, opened 2026-07-05
Onboarding harness: nightly regression detected


scoop:
# Onboarding Gap Report

**5 / 6 scenarios reached green.** (3 detection-only, by design — excluded.)

| Scenario | Image | Detected | Applied | Green | Classification |
| --- | --- | --- | --- | --- | --- |
| gh-unauthed | images:ubuntu/24.04 | yes | skipped | no | DETECTION-ONLY |
| gh-missing | images:debian/12 | yes | skipped | no | DETECTION-ONLY |
| tailscale-missing | images:ubuntu/24.04 | yes | skipped | no | DETECTION-ONLY |
| herdr-missing | images:archlinux | no | skipped | no | BOOT CRASH |
| claude-missing | images:rockylinux/9 | yes | verbatim | yes | PASS |
| git-missing | images:alpine/3.21 | yes | verbatim | yes | PASS |
| node-too-old | images:debian/12 | yes | verbatim | yes | PASS |
| install-e2e | images:ubuntu/24.04 | yes | verbatim | yes | PASS |
| install-e2e-service | images:ubuntu/24.04 | yes | verbatim | yes | PASS |

## Gaps

- **herdr-missing** (BOOT CRASH) — baseline step failed (herdr-missing): command -v unzip >/dev/null 2>&1 || (apt-get update && apt-get install -y unzip) || apk add --no-cache unzip || dnf install -y unzip || pacman -Sy --noconfirm unzip
sh: line 1: apt-get: command not found
sh: line 1: apk: command not found
sh: line 1: dnf: command not found
error: unzip: signature from "Lukas Fleischer <lfleischer@lfos.de>" is unknown trust
error: failed to commit transaction (invalid or corrupted package (PGP signature))

_Nightly run: 2026-07-05T01:37:33.387Z_

scoop:
✅ PR #1439 merged.

PR #1758, opened 2026-07-14
fix(onboarding): stop racing the Arch image's own keyring init

### LINKED ISSUE #1738: Onboarding harness: nightly regression detected
# Onboarding Gap Report

**5 / 6 scenarios reached green.** (4 detection-only, by design — excluded.)

| herdr-missing | images:archlinux | no | skipped | no | BOOT CRASH |
| herdr-outdated | images:archlinux | yes | skipped | no | DETECTION-ONLY |
| install-e2e | images:ubuntu/24.04 | yes | verbatim | yes | PASS |
| install-e2e-service | images:ubuntu/24.04 | yes | verbatim | yes | PASS |

## Gaps

- **herdr-missing** (BOOT CRASH) — baseline step failed (herdr-missing): command -v pacman >/dev/null 2>&1 || exit 0; pacman-key --init && pacman-key --populate archlinux && pacman -Sy --needed --noconfirm archlinux-keyring
gpg: checking the trustdb
gpg: Note: third-party key signatures using the SHA1 algorithm are rejected
gpg: (use option "--allow-weak-key-signatures" to override)
gpg: marginals needed: 3  completes needed: 1  trust model: pgp
gpg: depth: 0  valid:   1  signed:   0  trust: 0-, 0q, 0n, 0m, 0f, 1u
gpg: can't open '/etc/pacman.d/gnupg/pubring.gpg'
gpg: keydb_get_keyblock failed: Cannot open keyring
gpg: can't open '/etc/pacman.d/gnupg/pubring.gpg'
gpg: keydb_get_keyblock failed: Value not found
gpg: no writable keyring found: Not found
gpg: error reading '/usr/share/pacman/keyrings/archlinux.gpg': General error
gpg: no valid OpenPGP data found.
gpg: import from '/usr/share/pacman/keyrings/archlinux.gpg' failed: General error
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
gpg: error reading key: No public key
==> ERROR: D8AFDDA07A5B6EDFA7D8CCDAD6D055F927843F1C could not be locally signed.

_Nightly run: 2026-07-14T01:38:20.009Z_

scoop:
Fixes #1738 — the nightly's `herdr-missing` **BOOT CRASH**.

## Root cause: a race, not a broken command

`images:archlinux` ships its **own** boot-time keyring init:

```
# /etc/systemd/system/pacman-init.service   (from the Incus image)
ConditionPathIsDirectory=!/etc/pacman.d/gnupg
After=time-sync.target
ExecStart=/usr/bin/pacman-key --init
ExecStart=/usr/bin/pacman-key --populate
```

`seed.ts::archKeyringRefresh()` ran **those exact same commands** itself. The harness gates only on DNS (`waitForDns`), which resolves long before `time-sync.target` — so our `pacman-key --init` ran *concurrently with the image's*. Two gpg processes on one `/etc/pacman.d/gnupg`, and the loser dies with exactly the reported `can't open pubring.gpg` / `no writable keyring found` / `could not be locally signed`.

Since `runScenario` launches a **fresh instance per scenario**, *both* Arch scenarios are exposed — which one loses on a given night is jitter. That's why the nightly looked flaky and self-inconsistent, and why `herdr-outdated` passing was luck, not immunity.

## The fix

Take sole ownership of the keyring instead of racing for it:

- **Stop every writer first.** `pacman-init.service`, plus `archlinux-keyring-wkd-sync.timer`/`.service` — enumerating `systemctl list-units --all '*keyring*' '*pacman*'` showed wkd-sync is `Persistent=true`, so an aged image cache can fire a boot catch-up run into the same homedir. The socket-activated `gpg-agent@`/`dirmngr@`/`keyboxd@` units *survive* stopping pacman-init, so `gpgconf --kill all` releases the homedir before we delete it. Bounded by `timeout 120` — `RUNNER_TIMEOUT_MS` is 20 min, and an unbounded stop would wedge a run rather than fail fast.
- **Rebuild unconditionally** rather than probe-and-heal: after a race the unit finishes its own populate, so the homedir converges to healthy and a `pacman-key --list-keys` probe would exit 0 and skip the heal; and `ConditionPathIsDirectory` makes `systemctl start` a no-op once the dir exists. Safe here — throw-away instances whose keyring we fully own.
- **`set -e` restores fail-closed.** `driver.exec` runs the step via `sh -c`, which returns only the **last** command's status, and `pacman -Sy --needed archlinux-keyring` exits 0 on "nothing to do". Without it a failed init/populate is swallowed and the step reports success **on an empty keyring**, resurfacing later as a misleading `curl`/`unzip` error.

Still a clean no-op on the 8 non-Arch scenarios (`command -v pacman` guard retained, asserted in test).

## Verification

| Check | Result |
| --- | --- |
| Raced micro-repro, current command | **3/3 fail** — reproduces #1738, same fingerprint `D8AFDDA0…` |
| Raced micro-repro, this fix | **5/5 pass**, keyring intact after boot settles, signature-verified `pacman -S` succeeds |
| Fail-closed (break `--populate`) | Exits **1** (the pre-`set -e` shape exited **0**) |
| Negative test (pre-corrupt keyring) | Heals, signed install succeeds |
| `onboarding:test --scenario herdr-missing` | **5/5 green** (was BOOT CRASH) |
| `onboarding:test --scenario herdr-outdated` | **2/2**, no baseline failure |
| **Full `onboarding:test`** | **6/6 green** (was 5/6); other 9 scenarios unchanged |
| `bun run lint`, `bun test` | pass (6963 tests) |

## Follow-up (deliberately out of scope)

`src/remediations.ts` (`GIT_INSTALL`'s pacman branch) has a superficially similar `pacman-key --init && --populate` shape. No scenario exercises it (git-missing is Alpine) and it ships to user machines, so it is untouched here.

⚠️ **If that is ever revisited, the `rm -rf /etc/pacman.d/gnupg` self-heal must NOT be ported to it.** It is destructive on a real user's host, and the defect it defends against does not exist there — a long-installed Arch box has no `pacman-init.service` and no fresh-boot keyring init to race.

### DIFF
--- a/ci/onboarding-harness/seed.ts
+++ b/ci/onboarding-harness/seed.ts
@@ -61,7 +61,9 @@ const PROFILES = ["default", "shep-onb"];
  *  function exists to close, degrading back to a flaky nightly. So we separate the one benign
  *  case from real failures: `systemctl list-unit-files` (0 present / 1 absent) gates the stop, an
  *  absent unit is skipped, and a present unit that will not stop fails this CHECKED baseline step
- *  loudly. `reset-failed` keeps its `|| true` — it is pure cosmetics.
+ *  loudly — and legibly: the stop keeps its stderr (only stdout is dropped), so systemd's own
+ *  reason rides along into the error seedInstance throws. `reset-failed` keeps its `|| true` — it
+ *  is pure cosmetics.
  *
  *  Guarded on `command -v pacman` so it's a clean no-op on apt/apk/dnf images (the 8 non-Arch
  *  scenarios). Runs as root — `driver.exec` → `incus exec` is root by default, which
@@ -85,8 +87,11 @@ function archKeyringRefresh(): string {
     '    systemctl list-unit-files "$u" >/dev/null 2>&1 || continue',
     // The unit exists, so a stop MUST succeed. Anything else — a 120s timeout (124), a bus
     // error — means a writer may still be live, and rebuilding the keyring under it would
-    // silently re-enter the #1738 race. Fail the (checked) baseline step instead.
-    '    timeout 120 systemctl stop "$u" >/dev/null 2>&1 || ' +
+    // silently re-enter the #1738 race. Fail the (checked) baseline step instead. stdout is
+    // dropped but stderr is NOT: systemd's own reason (job timed out / failed to connect to
+    // bus / …) is the diagnostic, and seedInstance propagates it into the thrown baseline
+    // error — so the failure is legible, not just loud.
+    '    timeout 120 systemctl stop "$u" >/dev/null || ' +
       '{ rc=$?; echo "FATAL: could not stop $u (exit $rc); refusing to rebuild the keyring ' +
       'while a writer may still be running" >&2; exit 1; }',
     "  done",
--- a/ci/onboarding-harness/seed.ts
+++ b/ci/onboarding-harness/seed.ts
@@ -8,25 +8,69 @@ const SHEPHERD_DIR = "/opt/shepherd";
 // replace default and leave the instance with no root device.
 const PROFILES = ["default", "shep-onb"];
 
-/** Arch's base-image `archlinux-keyring` drifts behind the mirror over time; once a
- *  mirror package is signed by a packager key newer than the image's keyring,
- *  `pacman -Sy <pkg>` fails "signature … is unknown trust / invalid or corrupted
- *  package (PGP signature)" and the checked baseline aborts (issue #1422 — herdr-missing
- *  on images:archlinux is the only Arch scenario). Refresh the keyring ONCE before any
- *  pacman install: init the gnupg dir, populate trust from the shipped keyring, then
- *  update archlinux-keyring from the mirror (its own signature verifies against the
- *  freshly-populated trust; its install hook re-populates the new packager keys).
- *  Populate-before-sync is required — `-Sy` first would fail verification before the
- *  keyring is refreshed. `--needed` skips a redundant reinstall when already current.
- *  Guarded on `command -v pacman` so it's a clean no-op on apt/apk/dnf images (the 8
- *  non-Arch scenarios). Runs as root — `driver.exec` → `incus exec` is root by default,
- *  which `pacman-key` requires. */
+/** Take sole ownership of Arch's pacman keyring, then refresh it from the mirror.
+ *
+ *  WHY refresh at all (#1422): the base image's `archlinux-keyring` drifts behind the
+ *  mirror; once a mirror package is signed by a packager key newer than the image's
+ *  keyring, `pacman -Sy <pkg>` fails "signature … is unknown trust / invalid or corrupted
+ *  package (PGP signature)" and the checked baseline aborts. So we rebuild trust and pull
+ *  a current `archlinux-keyring` BEFORE any pacman install. Populate-before-sync is
+ *  required — `-Sy` first would fail verification before the keyring is refreshed.
+ *
+ *  WHY stop units first (#1738): `images:archlinux` runs its OWN keyring init at boot —
+ *  `/etc/systemd/system/pacman-init.service` (`ExecStart=pacman-key --init` + `--populate`,
+ *  ordered `After=time-sync.target`). We only wait for DNS (waitForDns), which resolves
+ *  long before time-sync, so the pre-#1738 version of this function raced that unit: two gpg
+ *  processes writing /etc/pacman.d/gnupg, and the loser died with "can't open pubring.gpg /
+ *  no writable keyring found / could not be locally signed". Reproduced 3/3 by exec'ing as
+ *  soon as DNS resolved; 0/3 when run after the unit had settled. Since runScenario launches
+ *  a FRESH instance per scenario, BOTH Arch scenarios are exposed — which one loses on a
+ *  given night is jitter, which is exactly why the nightly looked flaky.
+ *
+ *  Enumerating `systemctl list-units --all '*keyring*' '*pacman*'` + `list-timers` on a
+ *  fresh instance, pacman-init is not the only writer of that homedir:
+ *   - `archlinux-keyring-wkd-sync.timer` (loaded, active, OnCalendar=weekly, Persistent=true)
+ *     refreshes keys in the SAME homedir. On a fresh instance it sits ~6 days out and does not
+ *     fire at boot — but Persistent=true means an aged image cache can fire a catch-up run
+ *     right inside our window, so we stop it (and its service) defensively.
+ *   - the socket-activated `gpg-agent@ / dirmngr@ / keyboxd@ etc-pacman.d-gnupg` daemons are
+ *     separate units that SURVIVE stopping pacman-init, so one can linger holding the homedir
+ *     we are about to delete; `gpgconf --kill all` releases them first.
+ *  Nothing else in that enumeration touches the homedir.
+ *
+ *  We then rebuild UNCONDITIONALLY (`rm -rf` + init + populate) rather than probe-and-heal:
+ *  after a race the unit finishes its own populate, so the homedir converges to healthy and a
+ *  `pacman-key --list-keys` probe would exit 0 and skip the heal; and pacman-init's
+ *  `ConditionPathIsDirectory=!/etc/pacman.d/gnupg` makes `systemctl start` a no-op once the dir
+ *  exists, so it cannot repair a broken-but-present homedir either. Destroying and rebuilding is
+ *  safe here — these are throw-away instances whose keyring we fully own. (For the same reason
+ *  this shape must NOT be ported to the in-app remediation that runs on real user machines.)
+ *
+ *  `set -e` is load-bearing: `driver.exec` runs this via `sh -c`, which returns only the LAST
+ *  command's status, and `pacman -Sy --needed archlinux-keyring` exits 0 ("nothing to do") when
+ *  the version already matches — so without it, a failed init/populate would be swallowed and the
+ *  step would report success on an EMPTY keyring, resurfacing later as a misleading `curl`/`unzip`
+ *  install error. It is placed after the guard so non-Arch images still exit 0.
+ *
+ *  `timeout 120` bounds the stop: it blocks on the unit's job, and incus.ts's RUNNER_TIMEOUT_MS
+ *  is 20min — unbounded, a wedged stop would turn a fast, legible failure into a hung run.
+ *
+ *  Guarded on `command -v pacman` so it's a clean no-op on apt/apk/dnf images (the 8 non-Arch
+ *  scenarios). Runs as root — `driver.exec` → `incus exec` is root by default, which
+ *  `pacman-key` and `systemctl` require. */
 function archKeyringRefresh(): string {
-  return (
-    "command -v pacman >/dev/null 2>&1 || exit 0; " +
-    "pacman-key --init && pacman-key --populate archlinux && " +
-    "pacman -Sy --needed --noconfirm archlinux-keyring"
-  );
+  return [
+    "command -v pacman >/dev/null 2>&1 || exit 0",
+    "set -e",
+    "timeout 120 systemctl stop pacman-init.service archlinux-keyring-wkd-sync.timer " +
+      "archlinux-keyring-wkd-sync.service >/dev/null 2>&1 || true",
+    "systemctl reset-failed pacman-init.service >/dev/null 2>&1 || true",
+    "gpgconf --homedir /etc/pacman.d/gnupg --kill all >/dev/null 2>&1 || true",
+    "rm -rf /etc/pacman.d/gnupg",
+    "pacman-key --init",
+    "pacman-key --populate archlinux",
+    "pacman -Sy --needed --noconfirm archlinux-keyring",
+  ].join("\n");
 }
 
 /** Cross-distro "ensure package `$1` is installed" one-liner (apt/apk/dnf/pacman). */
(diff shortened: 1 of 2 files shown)


scoop (commented):
## What the PR does

`ci/onboarding-harness/seed.ts:64-77` rewrites `archKeyringRefresh()` so that, on pacman images only, it (a) stops the image's own boot-time keyring writers, (b) kills lingering gpg daemons on the homedir, (c) destroys and rebuilds `/etc/pacman.d/gnupg`, then (d) syncs `archlinux-keyring`. That matches the failure in the originating issue (`can't open '/etc/pacman.d/gnupg/pubring.gpg'` / `no writable keyring found` — the signature of a second gpg process owning the homedir), and it keeps the #1422 behaviour (populate-before-sync) intact.

## Verification

- **Guard / no-op on non-Arch.** The first line is `command -v pacman >/dev/null 2>&1 || exit 0` (`ci/onboarding-harness/seed.ts:65`), and `set -e` is only enabled on line 66 — i.e. *after* the guard, so the early `exit 0` is not affected. The step is invoked unconditionally from `baselineCommands()` (`ci/onboarding-harness/seed.ts:130-134`) and executed as `sh -c <cmd>` in the checked baseline loop (`ci/onboarding-harness/seed.ts:214-219`), which throws on non-zero — so fail-closed is genuinely required here and `set -e` supplies it.
- **Multi-line command survives the transport.** `driver.exec` passes the string as a single argv element to `incus exec … -- sh -c <cmd>` (`ci/onboarding-harness/incus.ts:57-59`), so the `\n`-joined script is delivered verbatim; no quoting/word-splitting hazard.
- **"8 non-Arch scenarios" / "BOTH Arch scenarios".** Confirmed: ten scenarios in `ci/onboarding-harness/scenarios.ts`, of which exactly two use `images:archlinux` — `herdr-missing` (scenarios.ts:90-91) and `herdr-outdated` (scenarios.ts:109-110). The old comment's claim that herdr-missing was "the only Arch scenario" was wrong and is correctly superseded. The two `installE2E` scenarios (scenarios.ts:162, 184) are Ubuntu and return before `baselineCommands()` anyway (`ci/onboarding-harness/seed.ts:203-211`), so they are unaffected.
- **`timeout 120` rationale.** `RUNNER_TIMEOUT_MS = 20 * 60_000` is verified at `ci/onboarding-harness/incus.ts:31`, so the comment's "20min" is accurate and the bound does convert a wedged `systemctl stop` into a fast failure rather than a 20-minute hang.
- **Test.** `test/onboarding-harness/seed.test.ts:69-104` selects the step by `pacman-key --populate archlinux` (unique across the baseline), and its `stopIdx < keyIdx` ordering assertion holds because the first `pacman-key` occurrence in the generated string is `pacman-key --init` (the stop line mentions `pacman-init`, not `pacman-key`). The `set -e` regex `/(^|\n)set -e(\s|$)/` matches the emitted `\n`-joined script. Assertions are behavioural (either `set -e` *or* an unbroken `&&` chain satisfies the contract), which is the right shape.
- **Order of teardown.** `gpgconf … --kill all` precedes `rm -rf`, which precedes `pacman-key --init` — the only order that avoids deleting a homedir out from under a live agent. Under `set -e`, `rm -rf`/`--init`/`--populate`/`-Sy` are all checked; the three tolerated lines are explicitly `|| true`.

Could not verify the claims about `images:archlinux`'s unit layout (`pacman-init.service` with `After=time-sync.target` and `ConditionPathIsDirectory=!/etc/pacman.d/gnupg`, `archlinux-keyring-wkd-sync.timer` with `Persistent=true`), nor `gpgconf --homedir … --kill all` support — none of that ground truth lives in this repo. I assumed it as written; the failure text in the issue is consistent with it.

## Nit

See the finding below: the `|| true` on the stop line is load-bearing for the "unit not present" case, but it also swallows a genuine `timeout`/bus failure, after which the script proceeds to `rm -rf` and re-init the homedir while the very writer it meant to stop may still be alive — i.e. it degrades silently back into the #1738 race instead of failing loudly. Non-blocking (it only triggers if the stop hangs past 120s or systemd's bus is unreachable, neither of which is on the normal path), but distinguishing exit 124 / a real stop failure from "no such unit" would make the fix self-reporting rather than silently re-flaky.

scoop (commented):
## Verdict

The PR addresses issue #1738 (the `herdr-missing` BOOT CRASH: two gpg processes writing `/etc/pacman.d/gnupg`) by stopping the image's own keyring writers before rebuilding the keyring, and the previous revision's blocking finding is genuinely resolved. Only a debuggability nit remains, so this is a `comment`.

## Prior finding — re-verified as FIXED

> `timeout 120 systemctl stop … || true` swallows every failure mode … the script still runs `rm -rf /etc/pacman.d/gnupg` + `pacman-key --init` while the writer it failed to stop may still be running.

Addressed. The `|| true` is gone from the stop:
- Presence gate: `systemctl list-unit-files "$u" >/dev/null 2>&1 || continue` (`ci/onboarding-harness/seed.ts:85`) — an *absent* unit is the only benign skip.
- Hard failure otherwise: `timeout 120 systemctl stop "$u" … || { rc=$?; echo "FATAL: could not stop $u (exit $rc); …" >&2; exit 1; }` (`ci/onboarding-harness/seed.ts:89-91`). I confirmed `$?` inside the `||` group is the status of the left-hand `timeout` invocation (the `>/dev/null 2>&1` redirections bind to `timeout`, not to the `||` list), so a 120s hang does surface as `exit 124` rather than being masked.
- The failure actually aborts the run: `archKeyringRefresh()` is emitted from `baselineCommands()` (`ci/onboarding-harness/seed.ts:158`) and baseline steps are checked — `ci/onboarding-harness/seed.ts:238-242` throws `baseline step failed (…)` on any non-zero `r.code`. So `exit 1` fails the seed instead of degrading back to the flaky nightly.
- Bus reachability is no longer conflated with unit-absence: `ci/onboarding-harness/seed.ts:79-81` fails loudly when systemd is PID 1 but `list-unit-files` cannot talk to the bus, which is exactly the case that would otherwise have been read as "unit absent" by the `continue` on line 85.
- `reset-failed` (line 93) and `gpgconf --kill all` (line 95) keep `|| true`, which is fine — neither is load-bearing for the race.

## Other correctness checks performed

- **`set -e` placement.** Line 72, i.e. *after* the `command -v pacman >/dev/null 2>&1 || exit 0` guard on line 71, so the 8 non-Arch scenarios still exit 0. Verified this matters: `driver.exec` wraps the string in `["sh", "-c", cmd]` (`ci/onboarding-harness/seed.ts:239`), and `sh -c` returns only the last command's status, so without `set -e` a failed `pacman-key --init` (line 97) would be masked by `pacman -Sy --needed archlinux-keyring` (line 99) exiting 0 on "nothing to do". The comment's claim is accurate.
- **Ordering.** The stop loop (82-92) precedes `gpgconf --kill` (95) → `rm -rf` (96) → `pacman-key --init` (97). The test asserts this ordering by string index (`test/onboarding-harness/seed.test.ts:89-92`); the first `pacman-key` occurrence in the emitted string is line 97's, so the assertion is meaningful, not vacuous.
- **Timer before service.** `archlinux-keyring-wkd-sync.timer` is stopped before `archlinux-keyring-wkd-sync.service` (line 82-83), which is the correct order — stopping the service first would leave the timer free to re-trigger it.
- **POSIX-shell validity.** `if`/`for`/`{ …; }`/`continue` under `sh` are all POSIX; the alpine (busybox ash) path never reaches them because of the `pacman` guard.
- **Callers.** `archKeyringRefresh()` has exactly one caller (`ci/onboarding-harness/seed.ts:158`); its signature (`(): string`) is unchanged, so nothing else needed updating.
- **Test fixture.** `test/onboarding-harness/seed.test.ts:15-21` uses a non-`installE2E` scenario, so `seedInstance` really does walk `baselineCommands()` (`ci/onboarding-harness/seed.ts:225-233` early-returns only for `installE2E`) and the new assertions do exercise the emitted string.

## Finding

**`ci/onboarding-harness/seed.ts:89` — the FATAL stop path throws away the one piece of evidence an operator needs.** `timeout 120 systemctl stop "$u" >/dev/null 2>&1` sends systemd's *stderr* to `/dev/null`, so when the new fail-closed branch fires, the operator gets `FATAL: could not stop $u (exit $rc)` and nothing else — not `Failed to stop …: Unit not loaded`, not a bus error, not a job-timeout reason. This is self-defeating given the goal of the change (a fast, *legible* failure) and given that `ci/onboarding-harness/seed.ts:241` deliberately propagates `r.stderr || r.stdout` into the thrown baseline error, so systemctl's message would otherwise land in the nightly report for free. Dropping `2>&1` (keeping `>/dev/null` to suppress the normal chatter) costs nothing and makes the failure diagnosable on the first look.

- The gate on line 85 relies on `systemctl list-unit-files <pattern>` exiting non-zero when nothing matches (the comment on line 84 asserts "0 present / 1 absent"). That is systemd behaviour, not something in this repo, so I could not confirm it against ground truth; assumed as written. Note the blast radius is small either way: on `images:archlinux` all three units are expected to exist, so the `continue` branch is not taken today, and if the assumption were wrong the result would be a loud false failure (`systemctl stop` on an unloaded unit exits 5), not a silent return of the #1738 race. If you want it belt-and-braces, gating on non-empty output (or `systemctl cat "$u"`) instead of on exit status would remove the dependency on that exit-code contract.
- The claims about the `images:archlinux` base image itself — that `/etc/systemd/system/pacman-init.service` exists with `After=time-sync.target` and `ConditionPathIsDirectory=!/etc/pacman.d/gnupg`, that `archlinux-keyring-wkd-sync.timer` is `Persistent=true`, and that the three named units are the only writers of that homedir — come from the author's enumeration on a live instance and are not verifiable from this repo. Assumed as written; they are what the unit list on line 82-83 is derived from.
- Similarly, `gpgconf --homedir … --kill all` (line 95) targets an external tool's CLI; it is `|| true`-guarded, so a wrong flag degrades to a no-op rather than a failure.

scoop (commented):
## Re-review

### Prior finding — RESOLVED

> ci/onboarding-harness/seed.ts: the fail-closed stop redirects systemctl's stderr to /dev/null (`>/dev/null 2>&1`) … Drop the `2>&1` (keep `>/dev/null`).

Fixed. `ci/onboarding-harness/seed.ts:94` now reads `timeout 120 systemctl stop "$u" >/dev/null || { rc=$?; echo "FATAL: could not stop $u (exit $rc); …" >&2; exit 1; }` — only stdout is dropped, so systemd's own reason (job timed out / failed to connect to bus / unit not loaded) stays on stderr alongside the FATAL line.

I traced that stderr end to end rather than assuming it: the container process's stderr is what `incus exec` streams back on its own stderr, `captureSpawn` accumulates it into `Captured.stderr` (`ci/onboarding-harness/spawn.ts:21`, resolved at `spawn.ts:34`), and `seedInstance` interpolates it into the thrown error via `r.stderr || r.stdout` (`ci/onboarding-harness/seed.ts:246`). So the systemd reason does reach the baseline-failure message. The regression test pins it with a negative lookahead that would reject a re-added `2>&1` (`test/onboarding-harness/seed.test.ts:118`).

### Other checks performed

- **Fail-closed contract.** `set -e` is at `seed.ts:74`, placed after the `command -v pacman … || exit 0` guard at `seed.ts:73`, so non-Arch images still exit 0 — verified the guard is the first list element and that `baselineCommands()` treats the result as a checked step (`seed.ts:243`–`seed.ts:247`, where any non-zero `code` throws). The claim that a swallowed failure would otherwise be masked holds: the last command, `pacman -Sy --needed --noconfirm archlinux-keyring` (`seed.ts:104`), is the only one whose status `sh -c` would return.
- **`$?` capture.** In `cmd || { rc=$?; … }`, `$?` inside the brace group is `cmd`'s status (the assignment is the group's first command), so the `exit $rc` value in the FATAL message is the real one — including 124 from `timeout`.
- **Timeout bound.** The comment at `seed.ts:55-56` cites `RUNNER_TIMEOUT_MS`; verified it is `20 * 60_000` at `ci/onboarding-harness/incus.ts:30`, so a 120s bound on the stop does fail well inside the 20-minute SIGKILL backstop, as claimed.
- **Scope of the change.** Both Arch scenarios (`herdr-missing` at `ci/onboarding-harness/scenarios.ts:90-91` and `herdr-outdated` at `scenarios.ts:109-110`) run the full baseline — neither sets `installE2E`, which is the only flag that short-circuits `baselineCommands()` (`seed.ts:230`–`seed.ts:238`). So the new block does reach the scenarios the issue reports as failing.
- **Destructive rebuild is contained.** `rm -rf /etc/pacman.d/gnupg` (`seed.ts:101`) only ever executes behind the `command -v pacman` guard, on throw-away Incus instances created per scenario. Grepped the tree for other `pacman-key` users: the only other one is the production installer at `deploy/install.sh:130`, which this diff does not touch and whose non-destructive shape the comment explicitly warns against changing (`seed.ts:46-47`).
- **Bus-unreachable path.** The `systemctl list-unit-files --no-legend` probe at `seed.ts:81-83` exits 1 before the loop, so an unreachable bus cannot be misread as "unit absent" by the per-unit `|| continue` at `seed.ts:87`. That ordering is correct.

The per-unit existence gate at `ci/onboarding-harness/seed.ts:87` depends on `systemctl list-unit-files "$u"` exiting 1 for an absent unit (the comment at `seed.ts:62` asserts "0 present / 1 absent"). That contract lives in systemd, not in this repo; I tried to check it against the host's systemd and the command was denied by the sandbox, so I am neither confirming nor disputing it. Note the practical blast radius either way is small: the gate is only reached on an image that has both `pacman` and systemd as PID 1, and on `images:archlinux` all three units are expected to be present, in which case both possible semantics take the same branch.

- deploy/install.sh:130 runs the same `pacman-key --init && --populate` sequence on real user machines with no guard against the image's own `pacman-init.service`, so a freshly-booted Arch host could in principle hit the same #1738 race there; the seed.ts comment deliberately declines to port the destructive fix, but no non-destructive mitigation exists in the installer.

PR #1760, opened 2026-07-14
fix(plan-gate): forced re-review must deliver its findings (#1759)

### LINKED ISSUE #1759: Plan gate: a session that stops revising below the rework cap is unreachable — RE-REVIEW is silently inert and Resume is hidden
## Symptom

Observed on TASK-1601 (`site-design-system`, repo `topmedia-website`). Badge reads `REWORK · 5/8`. The operator clicked **RE-REVIEW** three times; each time the plan reviewer visibly spawned in herdr and wrote a fresh `request-changes` verdict — and **nothing reached the planning agent**. The round stayed frozen at 5/8, and no button in the UI can break out of it.

Live gate row:

```
decision: changes_requested   approved: 0
round: 5                      cap: 8 (planReviewCyclesCap)
finalRoundPending: 0          planHash: 708c5ab385d6
findings: 15                  session.status: done
```

## Why it happens

Three behaviours interlock into a state with no exit:

**1. Forced re-review of an unchanged plan delivers nothing** — `applyChangesRequested`, `src/plan-gate.ts:775` (F3):

```ts
if (f.forced && prior && prior.planHash === f.planHash) {
  gate.round = prior.round;                 // hold
  gate.finalRoundPending = prior.finalRoundPending;
  this.deps.store.putPlanGate(gate);
  this.deps.onChange(f.sessionId, gate);
  return;                                   // no steer, no signal
}
```

RE-REVIEW is `force: true`. Force bypasses only the *spawn-time* hash dedupe in `consider()` (`src/plan-gate.ts:330-334`), so the reviewer really does run and really does write a verdict — but F3 then suppresses delivery. The planning agent is idle/`done` and has not revised the plan, so `planHash` is frozen and **every** click takes this branch. The auto path is dedup'd on the same unchanged hash, so it never fires either. Closed loop: no steer → plan unchanged → next force is inert.

**2. The round can therefore never reach the cap.** `gate.round` advances only when a steer is *delivered* (`src/plan-gate.ts:796`). Delivery is suppressed, so 5/8 is a fixed point.

**3. The escape hatch is gated on reaching that cap.** `PlanGate.resume()` (`src/plan-gate.ts:957`) is the only action that resets the round *and* re-delivers the outstanding findings. It is reachable only when `quotaBlockReason` classifies the session as plan-quota-blocked, which requires `gate.round >= gate.cap` (`src/blocked.ts:132`). Below the cap the Resume CTA isn't rendered (`ui/src/lib/components/PlanPanel.svelte:418`) and `POST /api/sessions/:id/quota/resume` answers `"not-stalled"` (`src/server.ts:2836`).

**Net:** a plan-gate session that stops revising *below* the cap is unreachable through the UI. The only affordance offered (RE-REVIEW) is guaranteed to be a no-op — while still burning a full reviewer run, tokens included, on every click — and the one affordance that would work is hidden until a cap that can no longer be reached.

Contrast the PR critic: `ReviewService.forceReview` explicitly resets `errorRound` / `streakReviews` / `addressRound` / `finalRoundPending` (`src/review.ts:487-501`) precisely so a forced review gets a fresh delivery budget. The plan gate's force resets nothing. Same button, opposite meaning, no UI signal about which one you got.

## Proposed fix

Any one of these breaks the deadlock; 1+3 together are the coherent pair.

1. **Make forced re-review deliver.** If the operator explicitly asks for a re-review and the verdict still has findings, re-steer them even when `planHash` is unchanged — that is what the click *means*. If F3's "don't re-inject identical findings as noise" intent is worth keeping for the automatic path, keep it there and drop it for `forced`.
2. **Or offer Resume below the cap.** Let `quotaBlockReason` classify a `changes_requested` gate with an idle/done planning session as resumable regardless of `round`, so the Resume CTA (reset + re-deliver) is available whenever rework has actually stopped.
3. **Never let a forced review be silently inert.** If F3 (or the at-cap hold) suppresses delivery, say so — return a distinct status from `consider()`/`/review-plan` and surface it ("re-reviewed; plan unchanged, findings not re-sent"). Today the reviewer spawn, the token spend, and the rewritten verdict all look exactly like a successful round.

## Workaround for anyone hitting this

The findings are not lost — they're persisted on the gate row and rendered in the session's **Plan panel** (summary + body + findings list, `ui/src/lib/components/PlanPanel.svelte:377-388`). Copy them into the steer bar manually. Once the agent revises the plan the hash moves, the auto reviewer re-fires, and the normal loop resumes.

scoop:
Fixes #1759.

## The deadlock

A plan-gate session whose planning agent stops revising **below** the rework cap had no exit. Observed on TASK-1601 at `REWORK · 5/8`: three RE-REVIEW clicks, three real reviewer spawns, three fresh `request-changes` verdicts — and nothing reached the planning agent.

`applyChangesRequested` held the round and delivered **no steer** on a forced re-review of an unchanged plan (the "F3" hold, `src/plan-gate.ts:775`). Because `gate.round` advances *only* on a **delivered** steer, the round became a fixed point below the cap — and the one affordance that would break out (`PlanGate.resume()`: reset the round + re-deliver the findings) is gated on `round >= cap`, which could no longer be reached. Closed loop: no steer → plan unchanged → the next force is inert too. Every click still burned a full reviewer run, tokens included.

## The fix

**Delete F3.** A forced verdict now steers its findings and spends a round exactly like an auto one. Repeated clicks walk the round to the cap, where Resume/Dismiss already render — PlanPanel (`canShowPlanStallActions`) and the session row (`hold-row.ts` R7 → `quota` → Resume) both key off `round >= cap` and neither consults `planStallStatus`, so the escape appears the moment the agent parks at the cap.

The block was `f.forced`-gated, so **every non-forced caller already fell through to this body** — the auto settle path *and* the adopted-orphan path (both `forced: false`). Note the auto path *can* reach `finalize()` with an unchanged `planHash`: `consider()` deliberately re-spawns on an unchanged hash when the prior decision was `error`. That re-spawn is unforced, so it never took F3 either; the safety argument rests on `f.forced`, not on "the auto path never gets here."

`forced` now governs **only** the two stall-signal guards in `escalateStallIfNeeded` (a forced at-cap re-entry writes no learnings row; a forced crossing still signals exactly once).

## Behaviour change worth calling out

The at-cap hold stays — a re-review there sends no steer, but must remain *possible* since it can legitimately flip to `approve` and release the gate. Its `finalRoundPending`, however, is **recomputed** (`delivered && round >= cap`), never carried. F3 was the single path that carried a stale `true` across a **new** verdict, so deleting it means a forced at-cap re-review now **clears** a pending final round, flipping `planStallStatus` from `"final"` to `"stalled"` and surfacing the amber recovery menu at once instead of after `PLAN_FINAL_ROUND_TIMEOUT_MS` (15 min).

Accepted deliberately: the flag means "the cap-th steer just landed", and no steer landed on this verdict. It is the uniform behaviour of every other at-cap re-review, and an operator forcing a re-review at the cap is asserting it looks stuck — the 15-minute "actively revising" premise is exactly what their click contradicts. Pinned by a test seeded with `finalRoundPending: true` rather than left to incidental coverage (the existing at-cap tests never set the flag, so the drop would have been invisible to them).

## Scope notes

- **Resume is not offered below the cap.** `quotaBlockReason` feeds the blocked/nudge/badge/hold-row derivation; relaxing its `round >= cap` predicate would read a normal mid-revise idle tick (2/8) as quota-blocked. Untouched.
- **The issue's ask #3 (a distinct "held, not re-sent" status) is deliberately deferred.** The only remaining non-delivering state is a forced re-review *at* the cap — where Resume and Dismiss already render, so the operator is picking the wrong working button, not the only button. Surfacing it honestly needs a *post-finalize* signal (the outcome is unknown at spawn, so `PlanReviewTrigger` can't carry it and `POST /review-plan` can't relay it): a `plan_gates` column + migration + type + UI note + an EN/DE pair. Weighed and declined.
- Worst case is honest: if the agent still won't revise, reaching Resume costs `(cap - round)` further reviewer runs — three in the 5/8 case. Strictly better than today, where those same clicks buy nothing and the operator stays stuck; and it's the *worst* case, since the first click now actually delivers.

## Tests

`test/plan-gate-service.test.ts`: the two F3 tests inverted (an unchanged-plan forced re-review now steers and advances); a cap-crossing regression asserting the persisted gate reaches `decision === "changes_requested" && round >= cap` — the exact shape `canShowPlanStallActions`, `hold-row`'s `atCap`, and `quotaBlockReason` all key off; and the `finalRoundPending` pin above.

Server-only: no UI, i18n, glossary, schema, or feature-catalog change. Root `bun run lint` clean; `bun run test` 6964 pass / 0 fail.

[no-feature-entry]

### DIFF
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -24,15 +24,32 @@ import { fenceUntrusted } from "./untrusted";
 import { type OperatorLanguage } from "./operator-language";
 import { resumeThenSteer } from "./resume-then-steer";
 
-/** Outcome of an on-demand `consider()`: a reviewer actually spawned (`"started"`); the request
- *  was a no-op (`"skipped"` — not planning, a review already in flight/starting, a tombstone or
- *  plugin-abort during spawn, or an unchanged plan on the auto-path); the plan artifact is
- *  unavailable (`"plan-unavailable"`); or a spawn attempt failed with a specific cause —
- *  `"error-spawn"` (herdr couldn't start/register the reviewer), `"error-worktree"` (the review
- *  worktree couldn't be created), or `"error-auth"` (api-key mode with no key configured). The
- *  review-plan route relays this so the UI can name the cause instead of a generic failure. */
+/** Outcome of an on-demand `consider()`: a reviewer actually spawned (`"started"`, or
+ *  `"started-at-cap"` — see below); the request was a no-op (`"skipped"` — not planning, a review
+ *  already in flight/starting, a tombstone or plugin-abort during spawn, or an unchanged plan on the
+ *  auto-path); the plan artifact is unavailable (`"plan-unavailable"`); or a spawn attempt failed
+ *  with a specific cause — `"error-spawn"` (herdr couldn't start/register the reviewer),
+ *  `"error-worktree"` (the review worktree couldn't be created), or `"error-auth"` (api-key mode with
+ *  no key configured). The review-plan route relays this so the UI can name the cause instead of a
+ *  generic failure.
+ *
+ *  `"started-at-cap"` is a REAL run (treat it as started everywhere a spawn matters — the reviewing
+ *  indicator, the WS bridge, the not-a-failure check) that carries one extra fact: the rework streak
+ *  is already at/over the cap, so if this verdict comes back `request-changes` its findings will NOT
+ *  be steered to the planning agent (applyChangesRequested's at-cap hold) — the operator needs Resume
+ *  instead. It can still legitimately come back `approve` and release the gate, which is why the run
+ *  is allowed. Without this, an inert at-cap re-review is indistinguishable from a round that landed
+ *  (issue #1759). Client-side, anything that only asks "did a run start?" must go through
+ *  `planReviewStarted` (ui/src/lib/api.ts) rather than `=== "started"` — GitRail's plan-review check
+ *  is fail-closed and inverted, so a raw comparison would read an at-cap run as a spawn failure. */
 export type PlanReviewTrigger =
-  "started" | "skipped" | "plan-unavailable" | "error-spawn" | "error-worktree" | "error-auth";
+  | "started"
+  | "started-at-cap"
+  | "skipped"
+  | "plan-unavailable"
+  | "error-spawn"
+  | "error-worktree"
+  | "error-auth";
 
 /** Whether a settle edge should re-drive `consider()`. `done` always re-considers (first
  *  draft + any settle); `idle` re-considers ONLY when a prior gate already requested changes
@@ -323,7 +340,7 @@ export class PlanGateService {
   }
 
   /** Decide whether `session`'s current plan warrants a fresh adversarial review, and start one.
-   *  Returns `"started"` iff a reviewer actually spawned; `"skipped"` for a no-op — not planning,
+   *  Returns `"started"` (or `"started-at-cap"`) iff a reviewer actually spawned; `"skipped"` for a no-op — not planning,
    *  a review already in flight/starting, an already-`approved` gate, an unchanged plan on the
    *  auto-path, a `forget()` tombstone abort mid-fetch, or a plugin `onSpawn` refusal;
    *  `"plan-unavailable"` when the plan artifact is missing/unreadable/empty; or `"error"` if a
@@ -527,7 +544,7 @@ export class PlanGateService {
       model: reviewerEnv.model,
       effort: reviewerEffort,
     });
-    return "started";
+    return startedStatus(prior, this.cap);
   }
 
   /** Re-adopt plan reviews that were in flight when the server last stopped. The `inflight`
@@ -1010,6 +1027,18 @@ export class PlanGateService {
     this.reapReviewer(sessionId);
     this.deps.store.dropPlanGate(sessionId);
   }
+}
+
+/** Which "a reviewer spawned" status a just-launched run reports. A run starting on an already-at-cap
+ *  rework streak will NOT re-steer its findings if it comes back `request-changes`
+ *  (applyChangesRequested's at-cap hold), so it says so at the trigger rather than reading as an
+ *  ordinary landed round (#1759) — the operator needs Resume. Predicated on the PRE-review row, the
+ *  same value that hold governs; an operator Resume landing mid-review resets the round and delivery
+ *  happens after all, so the UI note is transient and the gate stays authoritative. Free function (not
+ *  a method) to keep begin()'s cyclomatic count under the Fallow bar. */
+function startedStatus(prior: PlanGate | null, cap: number): PlanReviewTrigger {
+  const atCap = prior?.decision === "changes_requested" && prior.round >= cap;
+  return atCap ? "started-at-cap" : "started";
 }
 
 /** Agent-facing steer that carries the reviewer's plan findings into the planning PTY. NOT i18n'd. */
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -40,8 +40,9 @@ import { resumeThenSteer } from "./resume-then-steer";
  *  instead. It can still legitimately come back `approve` and release the gate, which is why the run
  *  is allowed. Without this, an inert at-cap re-review is indistinguishable from a round that landed
  *  (issue #1759). Client-side, anything that only asks "did a run start?" must go through
- *  `planReviewStarted` (ui/src/lib/api.ts) rather than `=== "started"` — GitRail's plan-review check
- *  is fail-closed and inverted, so a raw comparison would read an at-cap run as a spawn failure. */
+ *  `planReviewStarted` (ui/src/lib/api.ts) rather than `=== "started"`: every plan-review consumer
+ *  narrows explicitly, so a raw comparison silently drops the at-cap case — no WS bridge, no note,
+ *  no toast — leaving a real run looking like nothing happened. */
 export type PlanReviewTrigger =
   | "started"
   | "started-at-cap"
@@ -1034,11 +1035,17 @@ export class PlanGateService {
  *  (applyChangesRequested's at-cap hold), so it says so at the trigger rather than reading as an
  *  ordinary landed round (#1759) — the operator needs Resume. Predicated on the PRE-review row, the
  *  same value that hold governs; an operator Resume landing mid-review resets the round and delivery
- *  happens after all, so the UI note is transient and the gate stays authoritative. Free function (not
- *  a method) to keep begin()'s cyclomatic count under the Fallow bar. */
+ *  happens after all, so the UI note is transient and the gate stays authoritative.
+ *
+ *  Keyed on `round` ALONE — exactly what the hold reads (`priorRound >= this.cap`). It must NOT also
+ *  require `decision === "changes_requested"`: an `error` gate CARRIES its round (buildGate) and is
+ *  deliberately re-reviewable on the auto-path (consider() never dedups an error verdict), so an
+ *  error-at-cap gate whose re-review comes back `request-changes` is held and never steered — the
+ *  exact inert-run-reads-as-landed defect this status exists to name. Two predicates for one hold is
+ *  how that hole opens; there is one. Free function (not a method) to keep begin()'s cyclomatic count
+ *  under the Fallow bar. */
 function startedStatus(prior: PlanGate | null, cap: number): PlanReviewTrigger {
-  const atCap = prior?.decision === "changes_requested" && prior.round >= cap;
-  return atCap ? "started-at-cap" : "started";
+  return (prior?.round ?? 0) >= cap ? "started-at-cap" : "started";
 }
 
 /** Agent-facing steer that carries the reviewer's plan findings into the planning PTY. NOT i18n'd. */
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -255,7 +255,10 @@ interface PlanInFlight {
   reviewerModel: string | null;
   reviewerEffort: string | null;
   priorRound: number; // adversarial rounds already spent on this plan streak
-  forced: boolean; // this run is an operator's manual re-review (consider's opts.force) — governs the F3 hold + stall-signal guards
+  // This run is an operator's manual re-review (consider's opts.force). Since #1759 it governs ONLY
+  // the two stall-signal guards in escalateStallIfNeeded — a forced verdict now delivers its findings
+  // and spends a round exactly like an auto one.
+  forced: boolean;
   startedAt: number;
   finalizing?: boolean;
 }
@@ -598,11 +601,13 @@ export class PlanGateService {
       reviewerEffort: sp.reviewerEffort ?? s.effort ?? null,
       priorRound: prior?.round ?? 0,
       // The `reviewer_spawns` row has no `forced` column, so a restart mid-forced-review resurrects
-      // the entry as `forced: false`. The divergence is deliberate: adding a column + migration to
-      // suppress one at-cap stall row isn't worth it, and it fails SAFE — after a crash "needs a
-      // human" is arguably true, and no operator is known to be present. Bounded at one stall row per
-      // restart per session, unreachable by clicking, so it can't walk the distiller toward its
-      // learnings-signal threshold. See applyChangesRequested's guards.
+      // the entry as `forced: false`. Since #1759 that costs NOTHING on delivery — a forced verdict
+      // steers and spends a round like an auto one either way — so the only divergence left is the
+      // stall-signal guards: this run may write an at-cap stall row a live forced run would have
+      // suppressed. Deliberate: a column + migration to suppress one row isn't worth it, and it fails
+      // SAFE — after a crash "needs a human" is arguably true, and no operator is known to be present.
+      // Bounded at one row per restart per session, unreachable by clicking, so it can't walk the
+      // distiller toward its learnings-signal threshold. See escalateStallIfNeeded's guards.
       forced: false,
       startedAt: sp.spawnedAt, // keep the original start so a verdict-less orphan times out
     };
@@ -760,25 +765,21 @@ export class PlanGateService {
   }
 
   /** Steer the findings back to the LIVE planning agent while under the cap; at/over the cap stop
-   *  steering and escalate to the operator. Execution stays gated until the plan is approved. */
+   *  steering and escalate to the operator. Execution stays gated until the plan is approved.
+   *
+   *  EVERY verdict takes this path, forced or not (issue #1759). A forced re-review of an UNCHANGED
+   *  plan used to hold the round and deliver nothing, on the theory that re-injecting identical
+   *  findings is noise. It isn't: the planning agent may have stopped revising precisely because it
+   *  didn't act on them, the reviewer writes a FRESH verdict each run, and — because `round` advances
+   *  only on a DELIVERED steer — suppressing delivery froze the round below the cap, where the
+   *  operator's only escape (Resume, gated on `round >= cap`) is not yet offered. That closed the
+   *  loop: no steer → plan unchanged → the next force is inert too. An operator's click means "send
+   *  these to the agent", so it now steers and spends a round like any other; repeated clicks walk the
+   *  round to the cap, where Resume/Dismiss render. */
   private async applyChangesRequested(f: PlanInFlight, gate: PlanGate): Promise<void> {
     // Read the LIVE (prior) gate — buildGate hasn't persisted this verdict yet, so this still
     // returns the pre-review row (with its round / finalRoundPending / planHash).
     const prior = this.deps.store.getPlanGate(f.sessionId);
-    // F3 — a forced re-review of an UNCHANGED plan is inert on the streak. The findings are
-    // identical to what the planning agent already holds, so re-injecting them is noise. Hold
-    // `round`, leave `finalRoundPending` as it was, deliver NO steer, and write NO signal — neither
-    // the at-cap crossing nor the sub-cap unreachable-pane escalation, since no delivery was
-    // attempted so "unreachable" is meaningless. The fresh verdict (summary/body/findings) is still
-    // persisted + surfaced; only the streak-advancing and signal side effects are suppressed. A
-    // forced review of a CHANGED plan (hash differs) is a real revision and falls through below.
-    if (f.forced && prior && prior.planHash === f.planHash) {
-      gate.round = prior.round; // live round (not the begin-time snapshot) — an operator reset wins
-      gate.finalRoundPending = prior.finalRoundPending;
-      this.deps.store.putPlanGate(gate);
-      this.deps.onChange(f.sessionId, gate);
-      return;
-    }
     // Carry the LIVE gate's round, not the begin-time snapshot: an operator resume()/dismiss() that
     // reset the round WHILE this review was in flight must win over this finalize, not be clobbered
     // back to the pre-reset value (the reviewer captured f.priorRound before the reset).
@@ -798,6 +799,10 @@ export class PlanGateService {
     // (priorRound === cap-1 → round === cap, delivered). The no-steer post-cap re-review
     // (priorRound >= cap, delivered=false) and every sub-cap round leave it false, so
     // planStallStatus can tell a genuine final round from a stall/takeover. See src/plan-status.ts.
+    // This RECOMPUTE (not a carry-forward) is uniform across forced and auto runs: a forced re-review
+    // at the cap therefore CLEARS a previously-true flag, flipping planStallStatus "final" → "stalled"
+    // and surfacing the recovery menu at once rather than after PLAN_FINAL_ROUND_TIMEOUT_MS. Intended:
+    // the flag means "the cap-th steer just landed", and this write is a NEW verdict on which none did.
     gate.finalRoundPending = delivered && gate.round >= this.cap;
     this.escalateStallIfNeeded(f, gate, delivered);
     this.deps.store.putPlanGate(gate);
@@ -807,13 +812,14 @@ export class PlanGateService {
   /** Emit the learnings `stall` signal when a changes-requested round can't progress on its own —
    *  at/over the cap, or sub-cap with an undelivered steer (dead pane). A forced re-review suppresses
    *  the signal so a repeat-clickable button can't spam the distiller: an at-cap RE-ENTRY
-   *  (`priorRound >= cap`) is fully suppressed, but a forced CROSSING (`priorRound === cap-1` whose
-   *  steer lands) still signals exactly once. Split out of applyChangesRequested to keep its
-   *  cognitive complexity within the health bar. */
+   *  (`priorRound >= cap`) is suppressed by guard #1 below, but a forced CROSSING (`priorRound ===
+   *  cap-1` whose steer lands) still signals exactly once. These two `f.forced` guards are now the
+   *  ONLY behaviour keyed off `forced` — since #1759 a forced verdict is otherwise an ordinary round.
+   *  Split out of applyChangesRequested to keep its cognitive complexity within the health bar. */
   private escalateStallIfNeeded(f: PlanInFlight, gate: PlanGate, delivered: boolean): void {
     if (gate.round >= this.cap) {
-      // Fires on the crossing round and on any re-review that re-enters already at the cap. NOT
-      // `!f.forced` — a forced crossing still signals once; only a forced at-cap re-entry is suppressed.
+      // Guard #1. Fires on the crossing round and on any re-review that re-enters already at the cap.
+      // NOT `!f.forced` — a forced crossing still signals once; only a forced at-cap re-entry is suppressed.
       if (!(f.forced && f.priorRound >= this.cap)) {
         this.deps.store.addSignal({
           repoPath: f.repoPath,
@@ -825,9 +831,9 @@ export class PlanGateService {
       return;
     }
     if (!delivered) {
-      // Sub-cap but the steer didn't land (dead/unreachable pane): stranded like a cap stall rather
-      // than mid-revision. Escalate so it surfaces — but suppress the learnings signal on a forced run
-      // (newly reachable per click and unbounded); the console.warn + rendered verdict still surface it.
+      // Guard #2. Sub-cap but the steer didn't land (dead/unreachable pane): stranded like a cap stall
+      // rather than mid-revision. Escalate so it surfaces — but suppress the learnings signal on a forced
+      // run (clickable at will and unbounded); the console.warn + rendered verdict still surface it.
       console.warn(
         `[plan-gate] changes-requested steer did not land for ${f.sessionId}; escalating`,
       );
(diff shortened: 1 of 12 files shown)


scoop (commented):
## What the PR does

It removes the F3 hold from `applyChangesRequested` (`src/plan-gate.ts:775` on the base, now gone) so that a forced re-review of an unchanged plan takes the normal steer path: the findings are delivered, the round advances, and repeated clicks walk the round to the cap. `forced` now only keys the two learnings-signal guards in `escalateStallIfNeeded` (`src/plan-gate.ts:817`, `src/plan-gate.ts:833`). This is the issue's proposal #1, which the issue itself says is sufficient on its own to break the deadlock.

## Verification of the escape hatch

I traced the claim that repeated forced clicks now reach the cap and surface Resume, rather than assuming it:

- Delivery on an idle/`done` planner: `steerFindings` (`src/plan-gate.ts:756`) calls `resumeThenSteer` (`src/resume-then-steer.ts:25-37`), which resumes a dead pane before steering and returns the steer's success. For the issue's exact case (Claude, `session.status: done`, pane alive at the prompt) `paneAlive` is true, so `deps.reply` runs and `delivered` is true.
- Round advance: `gate.round = priorRound >= cap ? priorRound : delivered ? priorRound + 1 : priorRound` (`src/plan-gate.ts:794`). With delivery restored, round 5 → 6 → 7 → 8 across clicks.
- The CTA at the cap: `canShowPlanStallActions` requires `planPhase === "planning" && status !== "running" && !reviewing && decision === "changes_requested" && round >= cap` (`ui/src/lib/components/plan-gate-badge.ts:71-77`), and `PlanPanel`'s Resume button renders on exactly that predicate (`ui/src/lib/components/PlanPanel.svelte:51` and `:413`). The server side agrees: `quotaBlockReason`'s plan arm keys off `gate.round >= gate.cap` (`src/blocked.ts:128-134`) and `handleSessionQuotaResume` calls `planGate.resume()` only when that reason fires (`src/server.ts:2841`). `hold-row`'s `atCap` → `"quota"` (R7) is the same shape (`ui/src/lib/hold-row.ts:66` and `:83`).

So the loop the issue described (no steer → plan unchanged → next force inert) is genuinely broken, and the exit becomes reachable. `cap` is written on every gate by `buildGate` (`src/plan-gate.ts:911`), so the new test's `round >= gate.cap` assertion is not vacuous.

I also checked for stale collateral: no UI copy or message key asserts the old "a forced re-review does not spend a round" semantics — `feat_manual_plan_rereview_body` only promises a fresh review (`ui/messages/en.json:2335`, with its German counterpart at `ui/messages/de.json:2335`), and `canTriggerPlanReview` already treats an unchanged/at-cap plan as startable (`ui/src/lib/components/plan-gate-badge.ts:97-113`). The PR adds no new message keys, so there is no locale-parity exposure. The signal guards still behave as documented: a forced crossing (`f.priorRound === cap-1`) signals once and a forced at-cap re-entry is suppressed, which the existing tests at `test/plan-gate-service.test.ts:1774` and `:1800` pin.

The deliberate `finalRoundPending` clearing on a forced at-cap re-review (`src/plan-gate.ts:806`) is consistent, not a bug: the auto at-cap path already recomputed it to `false` (no steer is attempted when `priorRound >= cap`), so this only makes the forced path uniform, and the effect on `planStallStatus` (`src/plan-status.ts:27-32`) is to surface the recovery menu sooner, not to hide it.

## Residual gaps (non-blocking)

Two paths where a forced re-review is still inert survive the change. Neither reintroduces the reported deadlock in the reported scenario, but both are worth a follow-up, and one of them still has the exact shape the issue's title describes.

The sharper one: when the steer does **not** land sub-cap (`delivered === false` — a pane that is dead and that `resume()` cannot revive, e.g. an archived session or a provider that refuses resume, per `src/resume-then-steer.ts:34`), the round still holds at `priorRound`, `escalateStallIfNeeded`'s guard #2 suppresses even the learnings signal because the run is forced, and only a `console.warn` is emitted (`src/plan-gate.ts:836-847`). The gate is then frozen below the cap with no Resume CTA — the issue's headline condition, reached by a different route. The PR's own test pins this (`test/plan-gate-service.test.ts:1824-1845`: round stays 0, no signal). The issue's proposal #2 (offer Resume below the cap once rework has actually stopped) is what would close this; the PR chose #1 only, which is legitimate per the issue text, but the hole is real.

The milder one: the issue's proposal #3 (never let a forced review be silently inert) is not implemented. An at-cap forced re-review still spawns the reviewer, spends the tokens, delivers nothing, and `consider()` still returns `"started"` to `POST /api/sessions/:id/review-plan` (`src/server.ts:2649`), so the UI cannot distinguish it from a round that landed. It is much less harmful now because at the cap the Resume/Dismiss affordances do render, but the token burn is still unannounced.

## Note on the behaviour change itself

A forced re-review now consumes a rework round. That is the intended semantics and the comment at `src/plan-gate.ts:768-777` says so plainly, but it is a real UX shift: with the default cap, a few clicks of Re-review now exhaust the plan-rework budget, and the only place that surfaces is the `{round}/{cap}` in `hold_awaiting_rereview` (`ui/messages/en.json:2878`). Worth a word in the button's title/tooltip.

## Comment volume

The diff is roughly 2:1 prose-to-code. That matches this file's existing house style (the surrounding comments are equally dense), so I am not treating it as a defect — but the block at `src/plan-gate.ts:601-610` now spends seven lines explaining that a restart-resurrected `forced: false` costs nothing, which is a fact the reader can derive from the one-line invariant "forced only gates the signal guards". Trimming is optional.

scoop:
**Finding 1 (forced sub-cap re-review whose steer doesn't land → offer Resume below the cap): not changing this, and I don't think proposal #2 closes it.** Findings 2 and 3 are fixed in `14d67fdb`.

Three reasons:

**1. Resume wouldn't deliver either.** `PlanGate.resume()` steers through `steerFindings` → `resumeThenSteer` — the *same* path that just returned false. On a pane `resume()` refuses (the non-isolated Codex planner: `codex resume --last` is cwd-scoped and would revive a sibling), Resume resets the round to 0 and then fails to steer exactly as the re-review did, returning `"unreachable"` (which `POST /quota/resume` already has a status and a toast for). So offering it below the cap would add a button that is *also* inert against a dead pane — the failure mode this PR exists to eliminate. What actually recovers a dead pane is the steer bar (the issue's own documented workaround) or Dismiss/take-over.

**2. It isn't a regression, and it isn't the issue's state.** Before this PR the forced sub-cap unchanged-plan path held the round and never even *attempted* a steer (F3 returned early). After it, the steer is attempted and fails on a pane that cannot be revived. The round is frozen in both cases; what changed is that it now freezes only when the pane is genuinely unrecoverable, rather than on every click against a perfectly live agent. The issue's session was reachable — the steer would have landed; F3 suppressed it. That is fixed.

**3. Relaxing `quotaBlockReason` below the cap is the option we explicitly rejected at plan time.** It feeds the blocked/nudge/badge/hold-row derivation, so a `changes_requested` + idle predicate without `round >= cap` classifies a *normal mid-revise idle tick* (2/8, agent about to pick the findings up) as quota-blocked across every consumer. Narrowing it to "the steer didn't land" would need that fact persisted on the gate row — a `plan_gates` column + migration — which is the cost the operator weighed and declined.

The dead-pane case stays as it is: `console.warn` plus the rendered verdict, with the steer bar as the recovery. If it's worth closing properly, it wants its own issue and the persisted-delivery-outcome column, not a Resume button that can't deliver.

scoop:
## What the PR does (verified)

The closed loop the issue reports is genuinely broken open. The F3 hold is gone from `applyChangesRequested` (src/plan-gate.ts:796-827): every `changes_requested` verdict now runs the same `priorRound < cap` → `steerFindings` → `round + 1` path, forced or not, so a Re-review of an unchanged plan against a live/idle planner delivers its findings and spends a round; repeated clicks walk `round` to the cap, where `canShowPlanStallActions` (ui/src/lib/components/plan-gate-badge.ts:66-78) and `quotaBlockReason` (src/blocked.ts:127-134) both light up and Resume becomes reachable. `forced` now only gates the two learnings-signal guards in `escalateStallIfNeeded` (src/plan-gate.ts:836-867).

Proposal #3 is implemented for the main path: `startedStatus` (src/plan-gate.ts:1039-1042) returns the new `"started-at-cap"`, `handleSessionReviewPlan` relays it verbatim (src/server.ts:2653), and the four client surfaces consume it — `planReviewStarted` (ui/src/lib/api.ts:1751-1753) used by GitRail (ui/src/lib/components/GitRail.svelte:562-564) and PlanPanel (ui/src/lib/components/PlanPanel.svelte:214-216), plus explicit toasts in PlanGateBadge.svelte:157-158 and UnitRow.svelte:425-426.

i18n parity checked: `plangate_review_at_cap`, `plangate_review_spends_round`, `planpanel_review_at_cap` all exist in ui/messages/en.json:3141-3143 and ui/messages/de.json:3141-3143, with the same `{round}`/`{cap}` placeholders on `plangate_review_spends_round`.

## Prior findings

1. **Dead-pane sub-cap freeze / "offer Resume below the cap" — accepted, not re-raised.** I verified the author's rebuttal against the code: `PlanGate.resume()` re-delivers through `this.steerFindings(...)` (src/plan-gate.ts:989), the exact call that just returned `false` in `applyChangesRequested` (src/plan-gate.ts:808) — both funnel into `resumeThenSteer` with the same `paneAlive`/`resume`/`reply` seams (src/plan-gate.ts:775-782), and the `resume` seam is documented to refuse a non-isolated Codex planner (src/plan-gate.ts:205-209). So a Resume CTA offered below the cap would be inert against exactly the pane that made the steer fail. The issue's own reported session (live agent, plan frozen, steer would have landed) is fixed by the F3 removal. Declining proposal #2 is defensible here.
2. **Distinct status for an inert at-cap re-review — implemented**, but with a reachable gap (finding 1 below).
3. **"a re-review now spends a round" hint — implemented** via `reviewHint` → `title={reviewHint}` on the Review control (ui/src/lib/components/PlanPanel.svelte:90-94, 490), en + de present.

## Findings

### 1. `startedStatus` narrows on `decision`, the hold it mirrors does not (src/plan-gate.ts:1039-1042)

```ts
const atCap = prior?.decision === "changes_requested" && prior.round >= cap;
```

The at-cap hold it claims to be "predicated on … the same value that hold governs" (src/plan-gate.ts:1035-1036) does **not** look at `decision` — it is purely `priorRound >= this.cap` on the live row (src/plan-gate.ts:803-805).

An `error` gate can carry a round at/over the cap: `buildGate` writes `round: resolved === "approved" ? 0 : (live?.round ?? f.priorRound)` (src/plan-gate.ts:927), so a reviewer that times out / writes an unparseable verdict after the streak reached the cap persists `decision: "error", round: cap`. `consider()` deliberately re-reviews an `error` gate even on the auto path (`prior.decision !== "error"`, src/plan-gate.ts:373). That next run reports plain `"started"`, and when its verdict comes back `request-changes`, `priorRound >= cap` → no steer, round held — the exact "inert re-review indistinguishable from a landed round" this PR exists to eliminate, just entered through the `error` door.

`(prior?.round ?? 0) >= cap` matches the hold exactly and is safe (`approved` gates carry `round: 0` and are skipped upstream at src/plan-gate.ts:360).

Note the client-side twin: `atCap` in PlanPanel is derived with the same `decision === "changes_requested"` narrowing (ui/src/lib/components/PlanPanel.svelte:55-58), so the `$effect` at PlanPanel.svelte:171-173 would immediately clear `heldAtCap` for an error-at-cap gate even if the server started reporting it. Both sides need the same predicate for the note to survive.

### 2. de strings quote a button label that doesn't exist (ui/messages/de.json:3141, 3143)

Both new German messages tell the operator to use „**Mit Anmerkungen fortsetzen**“, but the German label of that CTA is „**Mit Befunden fortsetzen**“ (`planpanel_quota_resume`, ui/messages/de.json:2838). The English strings quote "Resume with findings", which does match ui/messages/en.json:2838 verbatim. A German operator is pointed at a button that isn't on screen under that name.

### 3. Nit: the `"started-at-cap"` doc claims a GitRail behaviour that isn't there (src/plan-gate.ts:43-44, ui/src/lib/api.ts:1750)

"GitRail's plan-review check is fail-closed and inverted, so a raw comparison would read an at-cap run as a spawn failure." The inverted/fail-closed check is `doReview` — the **PR** critic path (`status !== "started"` → failure toast, ui/src/lib/components/GitRail.svelte:540). `doReviewPlan` narrows explicitly and ends with `isPlanReviewError(status)` (GitRail.svelte:562-573), so a raw `=== "started"` there would silently show nothing, not a failure toast. `planReviewStarted` is still the right abstraction; the justification cites the wrong function.

scoop (commented):
## Verdict

The four points from the previous revision are all genuinely addressed, and the core fix holds up under inspection. Remaining items are nits — decision is `comment`.

### Previous points — re-verified

1. **`startedStatus` keyed on `decision`** — FIXED. `src/plan-gate.ts:1046` now reads `(prior?.round ?? 0) >= cap`, with no `decision` narrowing. That is exactly the predicate the hold it mirrors uses: `applyChangesRequested` computes `priorRound = prior?.round ?? f.priorRound` (`src/plan-gate.ts:804`) and gates the steer on `priorRound < this.cap` (`src/plan-gate.ts:806`), ignoring `decision`. The error-at-cap case is real and now covered: `consider()` deliberately does not dedupe an `error` verdict (`src/plan-gate.ts:374`), so such a gate re-reviews with its round intact and its next `request-changes` verdict is held — and now reports `"started-at-cap"` (pinned by the test at `test/plan-gate-service.test.ts:1774`).
2. **PlanPanel `atCap` narrowing** — FIXED. `ui/src/lib/components/PlanPanel.svelte:58` is `!!gate && !gate.approved && gate.round >= gate.cap` — decision-free, so the `$effect` at `ui/src/lib/components/PlanPanel.svelte:172` no longer clears `heldAtCap` for an error-at-cap gate. The extra `!gate.approved` term is safe rather than a new divergence: `consider()` short-circuits an approved gate at `src/plan-gate.ts:361`, so `"started-at-cap"` can never be returned for one, and the term only serves to clear the note on approval.
3. **German CTA label** — FIXED. `ui/messages/de.json:3141` and `ui/messages/de.json:3143` now quote „Mit Befunden fortsetzen“, which is the actual label at `ui/messages/de.json:2838` (`planpanel_quota_resume`), rendered at `ui/src/lib/components/PlanPanel.svelte:462`. The English strings quote “Resume with findings”, matching `ui/messages/en.json:2838`.
4. **GitRail doc citation** — FIXED. `src/plan-gate.ts:43-46` (echoed at `ui/src/lib/api.ts:1749-1751`) no longer claims the plan-review check is fail-closed/inverted; it now says the consumers narrow explicitly, which is what `doReviewPlan` does (`ui/src/lib/components/GitRail.svelte:562-573`, ending in an `isPlanReviewError` arm) as opposed to `doReview`'s `!== "started"` fail-closed arm (`ui/src/lib/components/GitRail.svelte:540`).

### Verification of the new behaviour

- **The deadlock is actually broken.** The F3 hold is gone, so a forced verdict now steers (`src/plan-gate.ts:806-809`) and advances the round (`src/plan-gate.ts:815`). Once `round >= cap` with `decision === "changes_requested"`, `quotaBlockReason` classifies the gate as plan-quota-blocked (`src/blocked.ts:127-134`) and `canShowPlanStallActions` renders Resume/Dismiss (`ui/src/lib/components/plan-gate-badge.ts:71-77` → `ui/src/lib/components/PlanPanel.svelte:454-470`). So repeated clicks do reach the escape hatch, provided the steer lands.
- **`"started-at-cap"` reaches every consumer.** The route relays `status` verbatim (`src/server.ts:2652-2653`); `reviewPlan` passes it through (`ui/src/lib/api.ts:1766-1771`). All four call sites handle it: `ui/src/lib/components/GitRail.svelte:562-564`, `ui/src/lib/components/PlanGateBadge.svelte:157-158`, `ui/src/lib/components/UnitRow.svelte:424-425`, `ui/src/lib/components/PlanPanel.svelte:213-215`. No `=== "started"` comparison survives that would drop the new case (grepped `src` + `ui/src`; the only other hit is `src/server.ts:1604`, the unrelated doc-agent route). The demo router/state (`ui/src/lib/demo/state.ts:248`) narrows to `"started" | "plan-unavailable"` and is unaffected.
- **i18n parity.** `plangate_review_at_cap`, `plangate_review_spends_round`, `planpanel_review_at_cap` exist in both catalogs the repo maintains (`ui/messages/en.json:3141-3143`, `ui/messages/de.json:3141-3143`; `ui/messages/` contains only en + de), with the same `{round}`/`{cap}` params on both sides of `plangate_review_spends_round`.
- **`finalRoundPending` recompute.** `src/plan-gate.ts:824` sets it from `delivered && round >= cap`, so an at-cap forced re-review clears a previously-true flag and `planStallStatus` returns `"stalled"` (`src/plan-status.ts:30`) instead of `"final"` — i.e. the recovery menu surfaces immediately. That matches the in-code comment and the test at `test/plan-gate-service.test.ts:1805`; it is a deliberate, benign widening of the escape hatch.
- **Stall-signal guards.** `escalateStallIfNeeded` (`src/plan-gate.ts:837-868`) still suppresses the learnings signal for a forced at-cap re-entry and a forced undelivered steer, so the now-clickable-and-delivering button cannot spam the distiller.

### Residual gap (not a regression, noting for the reader)

The round only advances on a *delivered* steer (`src/plan-gate.ts:815`). If the planning pane is genuinely unreachable (`steerFindings` returns false and `resumeThenSteer` cannot revive it), a sub-cap forced re-review still leaves `round` frozen, Resume still never renders, and the trigger returns a plain `"started"` with no operator-facing signal (only the `console.warn` at `src/plan-gate.ts:855`). That is the pre-existing dead-pane path, not something this PR introduces — but it means the comment's claim at `src/plan-gate.ts:795-796` ("repeated clicks walk the round to the cap, where Resume/Dismiss render") holds only when the steer lands.

- `src/blocked.ts:127-134` / `ui/src/lib/components/plan-gate-badge.ts:75`: the Resume affordance is still gated on `decision === "changes_requested"`, so a gate stuck on `decision: "error"` at the cap has no reset CTA at all.

scoop (commented):
## What the PR does

It implements the issue's proposals 1 + 3: the F3 "forced re-review of an unchanged plan holds the round and delivers nothing" branch is deleted from `applyChangesRequested` (src/plan-gate.ts:798-828), so every `request-changes` verdict — forced or auto — now steers its findings and spends a round while sub-cap (`priorRound < this.cap` → `steerFindings`, src/plan-gate.ts:805-814). Repeated RE-REVIEW clicks therefore walk `round` up to the cap, where `canShowPlanStallActions` (ui/src/lib/components/plan-gate-badge.ts:67-78 — `gate.round >= gate.cap`) renders Resume/Dismiss, which is exactly the escape the issue said was unreachable. `forced` now only survives in the two stall-signal guards (src/plan-gate.ts:841, 858), and the adopt-orphan `forced: false` divergence (src/plan-gate.ts:629) is genuinely harmless now that delivery no longer keys on it.

Proposal 3 lands as a new `"started-at-cap"` trigger status, computed from the PRE-review row by `startedStatus(prior, this.cap)` (src/plan-gate.ts:1046-1048) and relayed verbatim by the route (src/server.ts:2652-2654).

## Verification

- **The at-cap predicate matches the hold it describes.** `startedStatus` keys on `(prior?.round ?? 0) >= cap` (src/plan-gate.ts:1047); the hold it warns about is `if (priorRound < this.cap)` in `applyChangesRequested` (src/plan-gate.ts:805). Both are decision-free, so an `error` gate at the cap — which `consider()` deliberately does not dedupe (src/plan-gate.ts:374) and whose row carries its `round` — is correctly classified. The UI mirror `atCap` (ui/src/lib/components/PlanPanel.svelte:58) uses the same `round >= cap` shape. Consistent.
- **Every `reviewPlan` consumer narrows on the new status.** Grepped all four: GitRail.svelte:558-570, PlanGateBadge.svelte:155-166, UnitRow.svelte:423-434, PlanPanel.svelte:218-227. GitRail and PlanPanel bridge via `planReviewStarted` (ui/src/lib/api.ts:1752-1754); the two toast-only surfaces branch on `"started-at-cap"` before `"started"`, so no real run falls through to the failure/no-op arms. `ui/src/lib/demo/state.ts:248` returns a narrower union that is still assignable — no break.
- **Toast tier (prior finding 2) is fixed.** All three sites now pass `alert: true` plus a per-session `key`; `toasts.info` supports both (`InfoOpts.alert`, `InfoOpts.key` — ui/src/lib/toasts.svelte.ts:44-56), and `alert: true` with no explicit duration is the 12 000 ms hover-pausing tier per docs/toast-inventory.md:16 and the `FAILURE_MS` branch at ui/src/lib/toasts.svelte.ts:107-119.
- **Locale parity.** All four new keys exist in both catalogs the repo maintains (`ls ui/messages` → en.json, de.json only): `plangate_review_at_cap`, `plangate_review_spends_round`, `planpanel_review_at_cap`, `planpanel_review_at_cap_no_resume` at ui/messages/en.json:3140-3143 and ui/messages/de.json:3140-3143, with the same `{round}`/`{cap}` placeholders on `plangate_review_spends_round` in both.
- **Prior finding 1 (error gate at the cap) is addressed** by the `planpanel_review_at_cap_no_resume` variant selected from `rework` (ui/src/lib/components/PlanPanel.svelte:97-99), which requires `decision === "changes_requested"` — the same condition `canShowPlanStallActions` uses to render Resume (ui/src/lib/components/plan-gate-badge.ts:75).
- **No stale F3 references** remain (`grep -rn "F3" src docs ui/src` — only unrelated push/index hits), and no doc describes the removed hold.

## Remaining nit

`canShowPlanStallActions` gates Resume on *three* things, not one: `!reviewing` and `session.status !== "running"` in addition to the decision (ui/src/lib/components/plan-gate-badge.ts:70-77). The note's Resume-naming variant only accounts for the decision, so for the ordinary `changes_requested` at-cap case the note appears the instant the operator clicks Re-review — precisely when `planGates.isReviewing()` flips true (PlanPanel.svelte:40, 52) and the Resume/Dismiss block (PlanPanel.svelte:461) stops rendering. For the whole duration of the run, the note tells the operator to "Use Resume with findings" while that button is off screen; it reappears only when the verdict lands. (The new browser test doesn't catch this because it never sets the reviewing flag — ui/src/lib/components/PlanPanel.browser.test.ts:758-773.) Same applies while the planning session is `running`. Self-healing and non-blocking, but it is the same "never name a button the operator can't see" rule the `_no_resume` variant was added to honour.

scoop (commented):
## What I verified

**The deadlock is actually broken.** The F3 hold is gone from `applyChangesRequested` (src/plan-gate.ts:797-828): every `changes_requested` verdict now attempts `steerFindings` whenever `priorRound < this.cap` (src/plan-gate.ts:806-813), and `gate.round` advances on a delivered steer (src/plan-gate.ts:815). So repeated RE-REVIEW clicks on a frozen sub-cap gate walk `round` up to `cap`, which is exactly the shape `canShowPlanStallActions` (ui/src/lib/components/plan-gate-badge.ts:71-77) and `quotaBlockReason`'s plan arm (src/blocked.ts:127-134) key off — both require `decision === "changes_requested" && round >= cap`. The Resume CTA renders on `planStalled` (ui/src/lib/components/PlanPanel.svelte:465), so the escape hatch becomes reachable. Proposal 1 of the issue, satisfied.

**`forced` is now confined to the two stall-signal guards** (src/plan-gate.ts:837-868), matching the updated doc comments at src/plan-gate.ts:276 and src/plan-gate.ts:619-626. No other `f.forced` read remains in the file.

**`started-at-cap` (proposal 3).** `startedStatus` (src/plan-gate.ts:1047-1049) keys on `(prior?.round ?? 0) >= cap` — the *same* predicate as the hold at src/plan-gate.ts:806, and deliberately not also on `decision`, which is right: an `error` gate carries its round (consider() never dedupes an error verdict, src/plan-gate.ts:374) and its next `request-changes` verdict is held identically. The route relays it verbatim (src/server.ts:2652-2653, asserted at test/server-plan-gate.test.ts:149-172).

**All four `reviewPlan` consumers narrow the new variant.** GitRail.svelte:562-568 (via `planReviewStarted`, so the WS bridge still arms), PlanPanel.svelte:228-230 (bridge + `heldAtCap`), PlanGateBadge.svelte:157-164 and UnitRow.svelte:425-432 (toast only, no bridge to arm). No surviving bare `status === "started"` on a plan-review path — the only other `!== "started"` in GitRail (line 540) is the *PR* review handler (`reviewPr`, a different union at ui/src/lib/api.ts:1806). `planReviewStarted` (ui/src/lib/api.ts:1752) is exported and imported where used.

**Toast options are real:** `alert` and `key` both exist on the info-toast options (ui/src/lib/toasts.svelte.ts:44-55).

**Locale parity holds.** The repo maintains exactly two catalogs (ui/messages/en.json, ui/messages/de.json). All four new keys exist in both at the same lines: `plangate_review_at_cap` (en.json:3141 / de.json:3141), `plangate_review_spends_round` (3142/3142, both with `{round}`+`{cap}`), `planpanel_review_at_cap` (3143/3143), `planpanel_review_at_cap_no_resume` (3144/3144). The German copy quotes „Mit Befunden fortsetzen“, which is the actual de string for `planpanel_quota_resume` (ui/messages/de.json:2838) — the button it names.

**Prior revision's finding — addressed.** `atCapNote` (ui/src/lib/components/PlanPanel.svelte:97-99) now keys on `planStalled`, i.e. `canShowPlanStallActions(session, gate, reviewing)` (PlanPanel.svelte:52), which is the *same* predicate that renders the Resume/Dismiss block at PlanPanel.svelte:465. So during the in-flight window (`reviewing === true`) the note falls back to the `_no_resume` wording, and it re-points itself when the CTA lands. The new test at PlanPanel.browser.test.ts:781 exercises exactly that (sets `planGates.reviewing` after the click).

## The one thing I'd still flag (non-blocking)

The fix's premise — "repeated clicks walk the round to the cap, where Resume/Dismiss render" (src/plan-gate.ts:795-796) — holds only when the steer *lands*. `steerFindings` → `resumeThenSteer` returns false when the pane isn't alive and `resume()` resolves falsy (src/resume-then-steer.ts:30-35: archived / no resumable session / caller-refused provider). On that path `delivered === false`, so `gate.round` does not advance (src/plan-gate.ts:815) and the gate stays sub-cap — `quotaBlockReason` (src/blocked.ts:132) never classifies it, Resume stays hidden, and the operator can click RE-REVIEW forever. `consider()` returned plain `"started"` (startedStatus only looks at the round), so no toast, no note, no `heldAtCap` — the run is silently inert, which is the very thing proposal 3 set out to eliminate. The service does `console.warn` and takes the escalate branch (src/plan-gate.ts:851-857), but the learnings signal is suppressed for forced runs and nothing reaches the UI. Narrower than the original bug (the reported session was `done` and resumable, so it is genuinely fixed), but it is the same "below the cap, no exit, no signal" shape, in the function this PR rewrote.

## Notes / limitations

- `finalRoundPending` is now recomputed rather than carried, so a forced at-cap re-review clears a pending final round and flips `planStallStatus` "final" → "stalled" early (src/plan-gate.ts:820-824). Documented as intended and covered by a test; I agree it is defensible (the flag means "the cap-th steer just landed", and this write is a verdict on which none did).
- ui/src/lib/demo/state.ts:248 still types `reviewPlan` as `"started" | "plan-unavailable"` — assignable to the widened union, so the demo just never exercises the at-cap note. Not a defect.

Issue #1763, opened 2026-07-14
Critic keys off session.baseBranch, not resolveDiffBase (PR's real base)


scoop:
Split out of #1757 (the "Related (minor)" item there).

`ReviewService` passes `session.baseBranch` to `computePatchId` (`src/review.ts`), unlike the UI diff endpoints and recap, which resolve the base via `resolveDiffBase` (`src/diff-base.ts`: PR `baseRefName` → `session.baseBranch`).

Usually equivalent. The divergent case is an agent that ignores `epicBaseDirective` and opens its PR against the default branch: the critic then reviews against the epic integration branch while the PR's Files-changed shows the whole accumulated diff — and (post-#1757) it gets the epic context block for a base the PR doesn't actually target.

Bounded today: such a child is blocked at the drain's base-mismatch gate (`src/drain.ts:2122`), so it is parked for the operator rather than landing. #1757 did not widen the divergence (its epic block is gated on the same base the critic actually diffs against, so block and diff stay mutually consistent).

Converging is a real change, not a one-liner: the session critic would need the PR's `baseRefName` — i.e. a `prReviewMeta` forge call per review, which it deliberately does not make today — and it would change the reviewed base for **every** session, not just epic children.

Issue #1762, opened 2026-07-14
Decide: refuse epics on a forge that cannot create branches, or keep the main-serialized degrade?


scoop:
Split out of #1757 (a product call that PR deliberately did not make).

`ensureBranch` is optional on `GitForge` (`src/forge/types.ts:439`) and implemented **only** by GitHub (`src/forge/github.ts:1453`). On Gitea/local an epic therefore runs **without an integration branch**: every child bases on the default branch and merges straight into it.

It still *works* — a merged child closes its issue, and epic done-ness is `integrationMerged || issueClosed` (`src/drain.ts:473`), so dependents unblock and the epic completes, serialized through main. But it gives up the epic model's value: no integration branch, no atomic epic landing PR, children land on the default branch one at a time.

#1757 kept that behavior (failing it closed would turn a degraded-but-*progressing* epic into permanent zero progress with no operator path out) and surfaced it as an epic warning (`NO_INTEGRATION_BRANCH_WARNING`, `src/epic-model.ts`).

The open question is whether that degrade should be allowed at all:

- **Keep it** — epics work everywhere, with a warning. (Today, post-#1757.)
- **Refuse at epic-start** — gate epic start on `forge.ensureBranch` support, so a Gitea operator gets a clear "epics need a branch-capable forge" instead of a quietly weaker epic.

Issue #1761, opened 2026-07-14
Epic landing-PR critic re-reviews the whole accumulated epic diff against main


scoop:
Split out of #1757 (a decision that PR deliberately did not make).

The landing-PR critic (`src/standalone-critic.ts:277`, the only PR the standalone critic reviews when `criticAllPrs` is off) takes the PR's real base (`:341`) — the default branch — so it re-reviews the **entire accumulated epic diff**, re-litigating every child that was already reviewed against the integration branch.

That matches the landing PR's actual Files-changed and works as a final whole-epic sweep, so it is defensible. But it is a product call, not obviously correct:

- **Keep as-is** — a genuine final integration sweep; the cost is re-raising per-child findings already addressed.
- **Make it epic-aware** — tell it the children were each already reviewed, so it targets integration-level defects (prompt-only; #1757 added the `EpicContext` plumbing this would reuse).
- **Narrow the reviewed diff** — restrict it to the un-reviewed delta. Biggest semantic change; risks a real integration defect slipping through.

#1757 shipped the `EpicContext` block for child PRs, so option 2 is now cheap.

scoop:
✅ PR #1843 merged.

PR #1764, opened 2026-07-14
fix(critic): make the epic-child critic epic-aware and ground it in the real base

### LINKED ISSUE #1757: Epic child critic grounds findings in a stale worktree (missing merged sibling work); prompt has no epic awareness
## Symptom

When an epic is draining, the critic on a child PR sometimes raises findings that only make sense against a pre-epic baseline — things an already-merged sibling has handled on the epic integration branch.

## What's actually correct today

The child critic does **not** diff against the default branch. `ReviewService` passes `session.baseBranch` to `computePatchId` (`src/review.ts:547`), and for an epic child that is the integration branch (`resolveSpawnBase`, `src/drain.ts:2273-2293`, which also emits `epicBaseDirective` so the agent opens the PR with `--base epic/<n>-<slug>`). `defaultComputePatchId` fetches that base fresh from origin and diffs three-dot (`src/critic-core.ts:236-242`), so merged sibling work is correctly excluded from the reviewed diff.

So the *diff* baseline is right. The skew comes from elsewhere.

## Root causes

**1. Epic children are never rebased onto the moving integration branch.** They're excluded from the merge train (`src/full-auto.ts:25` — `isEpicIntegrationBranch(s.baseBranch)` → not full-auto), so nothing steers a rebase when they fall behind; the drain only rebases the *landing* branch (#1071). The child's HEAD therefore stays forked from the integration tip as it was at spawn.

Combined with the critic's VERIFY discipline ("resolve every identifier the diff introduces or relies on — grep the tree to confirm it exists", `src/critic-core.ts`), the critic grounds its claims in a **worktree that is missing already-merged sibling work**. The diff is right; the surrounding ground truth it greps is stale. This is the direct cause of the reported symptom.

**2. The critic prompt has no epic awareness.** `reviewPrompt` (`src/critic-core.ts:27`) never states that the base is an epic integration branch, that sibling work has already landed on it, or that further children are still in flight. It judges "does this satisfy the task / is it complete" as if the PR were a standalone change against main.

**3. Silent degradation to the default branch.** In `resolveSpawnBase` (`src/drain.ts:2273-2291`), if `forge.ensureBranch` is absent or throws, the epic child's `baseBranch` falls back to the default branch with only a `console.warn`. Since the critic keys off `session.baseBranch`, that child genuinely *is* reviewed against `main`, and it loses `epicBaseDirective` too (the PR targets main, which the drain then blocks at `src/drain.ts:2122`). Warn-only — invisible to the operator.

## Related (by design, but worth a decision)

The **landing-PR** critic (`src/standalone-critic.ts:277`, the only PR the standalone critic reviews when `criticAllPrs` is off) takes the PR's real base (`:341`) — i.e. the default branch — so it re-reviews the *entire accumulated epic diff*, re-litigating every child already reviewed against the integration branch. That matches the PR's Files-changed and is arguably what you want for a final sweep, but it is a product call, not obviously correct.

Minor: the critic does not use `resolveDiffBase` (`src/diff-base.ts`, PR `baseRefName` → `session.baseBranch`), unlike the UI diff endpoints and recap (`src/server.ts:2096,2113`, `src/index.ts:527`). Usually equivalent, but it's a divergence.

## Proposed fix

1. **Keep the child's ground truth current** — either keep epic-child worktrees up to date with the integration branch (rebase/steer when behind base, the epic-branch analogue of the merge train's `rebaseSteer`), or at minimum have the critic fetch the integration branch so the tree it greps reflects merged siblings.
2. **Make the critic prompt epic-aware** — when `isEpicIntegrationBranch(session.baseBranch)`, tell it it is reviewing one child of an epic whose base already carries merged sibling work and whose remaining children are still in flight, so "incomplete vs. the whole epic" is not a finding.
3. **Surface the `ensureBranch` fallback** instead of only warning — an epic child silently based on `main` should be visible to the operator.
4. **Decide** whether the landing-PR review should stay a full re-review against the default branch or be narrowed.

scoop:
Closes #1757.

## The bug

When an epic drains, the critic on a child PR raises findings that only make sense against a pre-epic baseline — reporting an already-merged sibling's export or message key as missing.

**The diff was never wrong.** `ReviewService` passes `session.baseBranch` (the integration branch for an epic child) to `computePatchId`, which fetches that base fresh and diffs three-dot — so merged sibling work is correctly excluded from what gets reviewed.

**The tree is wrong.** The critic runs in a disposable worktree detached at the child's PR head, and epic children are never rebased onto the moving integration branch. So the tree it *greps* is the fork-point tree, missing every sibling that merged since — while VERIFY tells it to "grep the tree to confirm it exists." It greps, finds nothing, and reports a real symbol as missing.

## The fix — no tree mutation

An `EPIC CONTEXT` block, emitted only when the base is an epic integration branch. Merging the base into the critic worktree was tried and rejected in review: it makes the tree a *superset* of the PR (skewing citations and letting the ATTRIBUTION rule pin sibling defects onto in-diff paths, past a scope backstop built from the pre-merge file list), and it leans on a custom per-clone `i18n-union` merge driver that is absent in essentially every managed repo — so `--abort` → stale tree would be the common path.

The block does four things, each of which exists because of a specific way the naive version fails:

1. **Supersedes VERIFY's grep-and-conclude rule** for base-delta paths: a worktree miss is *not* evidence of absence. Merely *informing* the critic that its tree is incomplete leaves the standing absolute rule in force — the model is free to resolve the conflict the wrong way.
2. **Defines a base-citation form that SATISFIES VERIFY's citation requirement.** Without it, a *compliant* critic that correctly read the base would find its conclusion uncitable (`path:line` demands a worktree line) and route it back to CANNOT-VERIFY — silently suppressing a real finding.
3. **Keeps that citation form out of `findings`.** `attributeFinding` splits on the first `": "` and `isPathShaped` treats any `/`-bearing token as a path, so a finding prefixed `epic/…@sha:src/base-only.ts: …` parses as out-of-diff and `scopeFindings` **drops it** — deleting the very findings this change exists to enable. Findings keep an in-diff path, attributed to the file that *depends* on the base evidence. Guarded by a **behavioral** test, not a wording assertion.
4. **Makes a blob read the confirmation in both directions.** `git log -S` searches *history*, so it is a **locator, never a verdict**: a hit may name the commit that *deleted* the identifier, and "no hit" is unsound on a shallow clone. So a sibling that **removed or renamed** what the child depends on stays a real finding instead of being downgraded.

Every command named is on the reviewer allowlist — the critic runs under `--permission-mode dontAsk`, where a denied command fails *silently*, making the instruction naming it inert. (`git grep`/`git ls-tree` are denied; the block says so.) A test derives the permitted verbs from the live preset rather than a hand-copied list. The allowlist deliberately stays closed: `git grep -O` executes an arbitrary command, and the critic's input is an untrusted PR diff.

The delta enumeration is **precomputed server-side** and embedded (fenced + clipped) — commit subjects are agent-authored text derived from untrusted issue text, so unfenced they would be instructions to the agent deciding the verdict.

Both critics that review child PRs get it: `standalone-critic.ts` is the sole reviewer when the session critic is off, and reviews every child PR when `criticAllPrs` is on.

## Drain: fail closed only where degrading is actually wrong

- **`ensureBranch` throws** (GitHub; transient) → fail closed with a typed `EpicBaseUnavailableError` and a visible `epic_base_unavailable` hold. Degrading here would mix bases mid-epic — the child would open against main and the merge train would **silently land it on main** while its siblings integrate on the epic branch. The error is typed because `doSpawn`'s catch also wraps `service.create()`; an untyped marker would raise a mislabeled, epic-wide hold for any sandbox/provider error.
- **Forge lacks `ensureBranch`** (gitea/local) → **still degrades**, deliberately. Review caught that my earlier "such a child can never integrate" claim was false: that gate needs `isEpicIntegrationBranch(baseBranch)`, which a degraded child fails, so it retires normally, the merge train lands it, its issue closes, and the epic *progresses* — serialized through main. Holding it closed would convert a progressing epic into permanent zero progress. It now surfaces as an epic warning instead of a `console.warn`.

The hold's freshness is evaluated in `buildState`, not by map membership: the lazy cooldown-expiry delete lives inside the `doSpawn` the hold prevents from running, so a membership check would **deadlock** the epic rather than lapse and retry.

## Verification

Root `bun run lint` + `bun run test` (6992 pass), `ui`: `bun run check` (0 errors) + `bun run test` (3769 pass), fallow audit clean.

## Follow-ups (product calls this PR deliberately did not make)

- #1761 — landing-PR critic re-reviews the whole accumulated epic diff against main.
- #1762 — refuse epics on a branch-less forge, or keep the main-serialized degrade?
- #1763 — critic keys off `session.baseBranch` rather than `resolveDiffBase`.

### DIFF
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -320,12 +320,21 @@ function epicBlock(epic: EpicContext): string[] {
   // malformed. The commit subjects are context for a real delta, never the reason there is one.
   const known = epic.delta ?? null;
   const knownCurrent = !!known && known.paths.length === 0;
+  // One header line per state — and the UNKNOWN one must HEDGE. Under a failed collection we have
+  // not established that anything merged (the epic's first child could be here too), so asserting
+  // "siblings have ALREADY MERGED" would state as ground truth exactly what we failed to determine —
+  // in a prompt whose entire purpose is to stop the critic doing that. Ignorance is stated as
+  // ignorance; the conservative stale-tree machinery below still ships, because "may be stale" is
+  // the safe assumption, but it is never dressed up as fact.
+  const header = knownCurrent
+    ? `- Its base is the epic INTEGRATION BRANCH \`${epic.base}\`, not the default branch. The base carries NO content your fork point does not already have${known!.commits.length ? " (commits have landed on it, but their net diff against your fork point is empty — e.g. a revert pair)" : " (nothing has merged into it since this branch forked — you are its first child)"}, so your worktree is CURRENT with the base. Further children are STILL IN FLIGHT.`
+    : known
+      ? `- Its base is the epic INTEGRATION BRANCH \`${epic.base}\`, not the default branch. Sibling children have ALREADY MERGED into that base, and further children are STILL IN FLIGHT.`
+      : `- Its base is the epic INTEGRATION BRANCH \`${epic.base}\`, not the default branch. Sibling children MAY ALREADY HAVE MERGED into that base — the delta could NOT be enumerated here, so treat this as unknown, not as established fact. Further children are STILL IN FLIGHT.`;
   const lines = [
     "",
     "EPIC CONTEXT — this PR is ONE CHILD of a multi-PR epic:",
-    knownCurrent
-      ? `- Its base is the epic INTEGRATION BRANCH \`${epic.base}\`, not the default branch. The base carries NO content your fork point does not already have${known!.commits.length ? " (commits have landed on it, but their net diff against your fork point is empty — e.g. a revert pair)" : " (nothing has merged into it since this branch forked — you are its first child)"}, so your worktree is CURRENT with the base. Further children are STILL IN FLIGHT.`
-      : `- Its base is the epic INTEGRATION BRANCH \`${epic.base}\`, not the default branch. Sibling children have ALREADY MERGED into that base, and further children are STILL IN FLIGHT.`,
+    header,
     "- Judge this PR against ITS OWN task only. Incompleteness versus the whole epic is NOT a finding, and work another child owns is not this PR's to do.",
   ];
   // Tree is current with the base → it is not missing anything, so the base-inspection machinery
@@ -333,7 +342,9 @@ function epicBlock(epic: EpicContext): string[] {
   // here: the epic-scope judging rule above is the whole point in that case.
   if (knownCurrent) return lines;
   lines.push(
-    "- Your checked-out worktree is this child's branch, which has NOT been rebased onto the base. Sibling work merged into the base after this branch forked is ABSENT from the tree: `Read`, `Glob` and `Grep` cannot see it.",
+    known
+      ? "- Your checked-out worktree is this child's branch, which has NOT been rebased onto the base. Sibling work merged into the base after this branch forked is ABSENT from the tree: `Read`, `Glob` and `Grep` cannot see it."
+      : "- Your checked-out worktree is this child's branch, which has NOT been rebased onto the base. Any sibling work merged into the base after this branch forked would therefore be ABSENT from the tree — `Read`, `Glob` and `Grep` could not see it — so assume the tree MAY be missing base content and verify against the base before concluding anything is absent.",
   );
   if (epic.baseSha) {
     const sha = epic.baseSha;
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -25,6 +25,10 @@ const execFileAsync = promisify(execFile);
  *  its merge base is the fork point — any path NOT listed has base content identical to what the
  *  child's tree already shows. Both lists are capped; the `*Truncated` counts are surfaced in the
  *  prompt so a capped list can never be mistaken for a complete one (issue #1757). */
+/*  EMPTY vs NULL are DIFFERENT and must stay so: an empty delta is KNOWLEDGE (git ran; nothing has
+ *  merged into the base since this branch forked — the epic's first child), so the block says that
+ *  and skips the stale-tree machinery entirely, since the tree IS current with the base. A NULL
+ *  delta is IGNORANCE (the collection failed), where the block stays conservative. */
 export interface EpicBaseDelta {
   paths: string[];
   pathsTruncated: number;
@@ -97,7 +101,11 @@ export async function defaultCollectBaseDelta(
     } catch {
       commits = []; // subjects are context-only; their absence must not lose the path list
     }
-    if (!allPaths.length && !commits.length) return null; // nothing merged since the fork
+    // NB: an EMPTY result is returned as an empty delta, NOT null — the two mean different things
+    // and drive different prompts. Empty is KNOWLEDGE ("nothing has merged into the base since this
+    // branch forked" — the epic's first child), and the block then says so and skips the
+    // tree-is-stale machinery entirely. Null is IGNORANCE (git failed), where the block must stay
+    // conservative and tell the critic to enumerate the delta itself.
     return {
       paths: allPaths.slice(0, DELTA_PATH_CAP).map((p) => clip(p, DELTA_PATH_CLIP)),
       pathsTruncated: Math.max(0, allPaths.length - DELTA_PATH_CAP),
@@ -255,7 +263,9 @@ export function prReviewPrompt(
  *  Truncation is always stated, so a capped list can never be mistaken for a complete one. */
 function epicDeltaLines(epic: EpicContext, sha: string): string[] {
   const delta = epic.delta;
-  if (!delta || (!delta.paths.length && !delta.commits.length)) {
+  // Only reached for a NONEMPTY delta or an UNKNOWN one (null): epicBlock returns early on a
+  // known-empty delta, so we never tell the critic to enumerate a delta we already know is empty.
+  if (!delta) {
     return [
       "",
       `Enumerate what your tree cannot see with \`git diff --name-only HEAD...${sha}\` (three-dot, so any path NOT listed is identical to what your tree already shows) and \`git log --oneline HEAD..${sha}\`. That path list is a CANDIDATE set, not a reading list — read only the paths bearing on identifiers this PR's diff actually introduces or relies on.`,
@@ -294,13 +304,30 @@ function epicDeltaLines(epic: EpicContext, sha: string): string[] {
 }
 
 function epicBlock(epic: EpicContext): string[] {
+  // Three states, and they must not be conflated (see EpicBaseDelta):
+  //   - KNOWN-EMPTY delta: git ran and nothing has merged into the base since this branch forked —
+  //     i.e. the epic's FIRST child. Saying "siblings have ALREADY MERGED" would be false, and the
+  //     whole stale-tree apparatus is moot: the tree IS current with the base, so VERIFY's
+  //     grep-and-conclude rule is sound as written and must NOT be overridden.
+  //   - KNOWN-NONEMPTY delta: siblings did merge → the full block.
+  //   - UNKNOWN (null): the collection failed. Stay conservative — assume the tree may be stale.
+  const known = epic.delta ?? null;
+  const knownEmpty = !!known && !known.paths.length && !known.commits.length;
   const lines = [
     "",
     "EPIC CONTEXT — this PR is ONE CHILD of a multi-PR epic:",
-    `- Its base is the epic INTEGRATION BRANCH \`${epic.base}\`, not the default branch. Sibling children have ALREADY MERGED into that base, and further children are STILL IN FLIGHT.`,
+    knownEmpty
+      ? `- Its base is the epic INTEGRATION BRANCH \`${epic.base}\`, not the default branch. NOTHING has merged into that base since this branch forked (you are its first child), so your worktree is current with the base — but further children are STILL IN FLIGHT.`
+      : `- Its base is the epic INTEGRATION BRANCH \`${epic.base}\`, not the default branch. Sibling children have ALREADY MERGED into that base, and further children are STILL IN FLIGHT.`,
     "- Judge this PR against ITS OWN task only. Incompleteness versus the whole epic is NOT a finding, and work another child owns is not this PR's to do.",
-    "- Your checked-out worktree is this child's branch, which has NOT been rebased onto the base. Sibling work merged into the base after this branch forked is ABSENT from the tree: `Read`, `Glob` and `Grep` cannot see it.",
   ];
+  // Nothing merged since the fork → the tree is not missing anything, so the base-inspection
+  // machinery (and its override of the VERIFY grep rule) would be noise at best and misleading at
+  // worst. Stop here: the epic-scope judging rule above is the whole point for a first child.
+  if (knownEmpty) return lines;
+  lines.push(
+    "- Your checked-out worktree is this child's branch, which has NOT been rebased onto the base. Sibling work merged into the base after this branch forked is ABSENT from the tree: `Read`, `Glob` and `Grep` cannot see it.",
+  );
   if (epic.baseSha) {
     const sha = epic.baseSha;
     lines.push(
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -18,17 +18,110 @@ import { fenceUntrusted } from "./untrusted";
 
 const execFileAsync = promisify(execFile);
 
+/** The sibling work that landed on an epic integration branch AFTER the child under review forked
+ *  from it — collected server-side (see {@link defaultCollectBaseDelta}) and embedded in the epic
+ *  block so the critic starts with the enumeration in hand instead of having to earn it.
+ *  `paths` is COMPLETE as a candidate set: `git diff --name-only HEAD...<baseSha>` is three-dot, so
+ *  its merge base is the fork point — any path NOT listed has base content identical to what the
+ *  child's tree already shows. Both lists are capped; the `*Truncated` counts are surfaced in the
+ *  prompt so a capped list can never be mistaken for a complete one (issue #1757). */
+export interface EpicBaseDelta {
+  paths: string[];
+  pathsTruncated: number;
+  /** `git log --oneline` subjects. UNTRUSTED (agent-authored, derived from issue text) — fenced. */
+  commits: string[];
+  commitsTruncated: number;
+}
+
+/** Epic-child review context. Present iff the reviewed branch's base is an epic integration branch
+ *  (`isEpicIntegrationBranch`), i.e. this PR is ONE CHILD of a draining epic whose base already
+ *  carries merged sibling work. `baseSha` null = the base could not be resolved to a commit (the
+ *  fetch/rev-parse failed and there is usually no local ref for an epic branch) → the block degrades
+ *  to its no-base mode: no base commands, and existence conclusions become limitations, not
+ *  findings. */
+export interface EpicContext {
+  /** The integration branch name (for the operator-readable citation form). */
+  base: string;
+  baseSha: string | null;
+  delta?: EpicBaseDelta | null;
+}
+
+// Caps for the embedded delta. Bounded so a long-draining epic can't blow up the prompt; the
+// truncation counts are always stated, and the full lists stay one command away.
+const DELTA_PATH_CAP = 100;
+const DELTA_COMMIT_CAP = 30;
+const DELTA_PATH_CLIP = 200;
+const DELTA_SUBJECT_CLIP = 120;
+
+/** Clip one embedded entry, marking the clip (never silent). */
+function clip(s: string, max: number): string {
+  return s.length > max ? `${s.slice(0, max)}…` : s;
+}
+
+/** Collect the base delta for an epic child: the paths whose base content differs from the fork
+ *  point, plus the sibling commit subjects. Best-effort and PURE-ish (read-only git in the critic's
+ *  own disposable worktree): ANY failure yields null, and the epic block then simply tells the
+ *  critic to run the commands itself. Never throws.
+ *
+ *  Requires an already-resolved `baseSha` (computePatchId fetched the base, so its objects are
+ *  local). The SHA shape is validated before it reaches argv — it comes from `git rev-parse`, but a
+ *  cheap guard keeps a hostile ref from ever being smuggled into a git invocation. */
+export async function defaultCollectBaseDelta(
+  worktreePath: string,
+  baseSha: string,
+): Promise<EpicBaseDelta | null> {
+  if (!/^[0-9a-f]{7,40}$/.test(baseSha)) return null;
+  try {
+    // Three-dot: merge base = the fork point, so this is exactly the base-side content the
+    // child's tree cannot see (the completeness property the epic block leans on).
+    const { stdout: names } = await timedAsync("git diff --name-only", () =>
+      execFileAsync("git", ["diff", "--name-only", "-z", `HEAD...${baseSha}`], {
+        cwd: worktreePath,
+        maxBuffer: 64 * 1024 * 1024,
+        encoding: "utf8",
+      }),
+    );
+    // `-z` (NUL-delimited, unquoted) for the same reason computePatchId uses it: default
+    // core.quotePath C-quotes non-ASCII paths, which would then never match the real path.
+    const allPaths = names.split("\0").filter(Boolean);
+    let commits: string[] = [];
+    try {
+      const { stdout: log } = await timedAsync("git log --oneline", () =>
+        execFileAsync("git", ["log", "--oneline", "--no-decorate", `HEAD..${baseSha}`], {
+          cwd: worktreePath,
+          maxBuffer: 16 * 1024 * 1024,
+          encoding: "utf8",
+        }),
+      );
+      commits = log.split("\n").filter(Boolean);
+    } catch {
+      commits = []; // subjects are context-only; their absence must not lose the path list
+    }
+    if (!allPaths.length && !commits.length) return null; // nothing merged since the fork
+    return {
+      paths: allPaths.slice(0, DELTA_PATH_CAP).map((p) => clip(p, DELTA_PATH_CLIP)),
+      pathsTruncated: Math.max(0, allPaths.length - DELTA_PATH_CAP),
+      commits: commits.slice(0, DELTA_COMMIT_CAP).map((c) => clip(c, DELTA_SUBJECT_CLIP)),
+      commitsTruncated: Math.max(0, commits.length - DELTA_COMMIT_CAP),
+    };
+  } catch {
+    return null; // git missing / bad sha / worktree gone → prompt-only fallback
+  }
+}
+
 /** Self-contained instructions for the critic agent. NOT UI chrome — never i18n'd.
  *  `diffBase` is the RESOLVED base commit (a SHA captured by computePatchId from the same fresh
  *  fetch it fingerprints), NOT a branch name — so the review diffs the identical base the
  *  rebase-skip fingerprint used, and `git diff ${diffBase}...HEAD` is exactly the branch's own
- *  changes (no already-merged main commits folded in). */
+ *  changes (no already-merged main commits folded in).
+ *  `epic` is set iff the base is an epic integration branch — see {@link EpicContext}. */
 export function reviewPrompt(
   diffBase: string,
   taskPrompt: string,
   priorFindings: string[] = [],
   authorNotes: string[] = [],
   issueBody?: string | null,
+  epic?: EpicContext | null,
 ): string {
   const lines = [
     "You are a code critic reviewing a pull request. Do NOT modify, build, commit, or run anything — read-only inspection only.",
@@ -67,6 +160,7 @@ export function reviewPrompt(
     ...scopeAndOutputTail(
       diffBase,
       "Judge ONLY whether the implementation satisfies that task and is free of bugs, security issues, and clear quality problems. Tests and lint are handled by CI — do not run them.",
+      epic,
     ),
   );
   return lines.join("\n");
@@ -78,7 +172,12 @@ export function reviewPrompt(
  *  (title + body) only as CONTEXT for what the change is trying to do. Shares the EXACT scope rules
  *  and verdict-output contract with reviewPrompt (via scopeAndOutputTail) so the two never diverge.
  *  NOT UI chrome — never i18n'd. */
-export function prReviewPrompt(diffBase: string, prTitle: string, prBody: string): string {
+export function prReviewPrompt(
+  diffBase: string,
+  prTitle: string,
+  prBody: string,
+  epic?: EpicContext | null,
+): string {
   const lines = [
     "You are a code critic reviewing a pull request. Do NOT modify, build, commit, or run anything — read-only inspection only.",
     `The PR branch is checked out here at its head commit. Review the changes with: git diff ${diffBase}...HEAD`,
@@ -95,6 +194,7 @@ export function prReviewPrompt(diffBase: string, prTitle: string, prBody: string
     ...scopeAndOutputTail(
       diffBase,
       "Judge the diff ONLY for bugs, security issues, and clear quality problems. Use the stated intent above to understand what the change is for — do NOT raise a finding merely because the diff seems incomplete versus that intent. Tests and lint are handled by CI — do not run them.",
+      epic,
     ),
   );
   return lines.join("\n");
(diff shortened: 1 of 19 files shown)


scoop (commented):
## What I verified

**The task is satisfied.** All three actionable asks of #1757 are implemented, and the fourth ("decide" on the landing PR) is implicitly resolved by leaving the landing PR non-epic (its base is the default branch, so `isEpicIntegrationBranch(meta.baseRefName)` is false at `src/standalone-critic.ts:379` and no epic block is emitted).

1. *Stale ground truth* — rather than rebasing the child, the critic is now handed the base delta and told to read the base directly. `defaultCollectBaseDelta` (`src/critic-core.ts:57-98`) runs `git diff --name-only -z HEAD...<baseSha>` and `git log --oneline HEAD..<baseSha>`. The three-dot form is correct for "base content my tree cannot see" (merge base = fork point), and the two-dot log is the matching commit set. It runs *after* `computePatchId` has fetched the base into the same worktree (`src/critic-core.ts:428-431`, `src/review.ts:684`, `src/standalone-critic.ts:348`), so the objects `git show <sha>:<path>` needs are present.
2. *Epic-aware prompt* — `epicBlock` (`src/critic-core.ts:232-270`) is emitted only when an `EpicContext` is passed, and both critics pass it (`src/review.ts:683-684`, `src/standalone-critic.ts:379-385`).
3. *Surfacing the fallback* — the `ensureBranch` **throw** now fails closed (`src/drain.ts:2364`) and raises an `epic_base_unavailable` hold (`src/drain-core.ts:300`); the *absent*-`ensureBranch` degrade becomes an epic warning (`src/epic-model.ts:164`).

**Command allowlist conformance (the part most likely to be silently wrong):** the epic block names `git show`, `git log -S`, `git diff` and explicitly forbids `git grep` / `git ls-tree`. Checked against the live `reviewer` preset — `READONLY_GIT` at `src/transient-agent-argv.ts:84-92` is exactly `Read, Grep, Glob, Bash(git diff *), Bash(git log *), Bash(git show *), Bash(git status)`, composed into `reviewer` at `src/transient-agent-argv.ts:123`. Every command the block names is permitted; the two it forbids are indeed absent. `allowedToolsFor` (`src/transient-agent-argv.ts:867`) returns a copy, so the test seam cannot mutate the preset.

**Identifiers resolved:** `isEpicIntegrationBranch` is already imported in `src/standalone-critic.ts:28` (signature `(branch: string): boolean`, `src/epic-branch.ts:19`) and newly imported in `src/review.ts:31`; `timedAsync` is imported at `src/critic-core.ts:12`; `fenceUntrusted(label, content)` matches its definition at `src/untrusted.ts:38`; `EpicBaseUnavailableError`'s `this.cause` assignment is fine under `"target": "ESNext"` (`tsconfig.json:3`) and matches the existing pattern at `src/sandbox.ts:521`.

**i18n parity:** both new keys exist in both catalogs the repo maintains (`ui/messages/` holds only `en.json` and `de.json`) — `drain_paused_epic_base` and `epic_hold_epic_base_unavailable` at `ui/messages/en.json:3141-3142` and `ui/messages/de.json:3141-3142`, with the same `{branch}` parameter used at the two call sites (`ui/src/lib/components/queue-strip.ts:252`, `ui/src/lib/components/epic-panel.ts:76`). The UI's `DrainStatus.reason` is `string | null` (`ui/src/lib/types.ts:781`), so the new switch cases type-check.

**Cooldown self-heal:** the hold is derived per-tick from `freshEpicBaseFailure` (`src/drain.ts:768-777`), which applies the `SPAWN_FAIL_COOLDOWN_MS` freshness test itself. That is load-bearing exactly as documented: the lazy expiry delete lives in `doSpawn` (`src/drain.ts:2393`), which the hold prevents from running, so a membership-only check would latch forever. The shape change to `spawnFailures` has no other consumers (grepped: only `src/drain.ts`).

**Security:** `baseSha` is regex-validated before reaching argv (`src/critic-core.ts:61`), base-delta paths and commit subjects are wrapped with `fenceUntrusted` (`src/critic-core.ts:272,281`), and the injected branch name is constrained by `epicIntegrationBranch`'s slug regex (`src/epic-branch.ts:5-13`). No injection surface found.

## Issues (all non-blocking — see findings)

See the three findings. None is a correctness hole on the epic path this PR targets; the first is a genuine (narrow, self-healing) mis-hold.

- src/critic-core.ts:200 — `epicDeltaLines` emits the "The base content your tree CANNOT see is exactly the paths below" preamble whenever *either* list is non-empty, so a delta with commits but zero changed paths (empty/reverting base commits) prints that sentence with no path fence following it. Harmless today; only reachable when a merged sibling contributes no path delta.

scoop (commented):
## Re-review of the three prior findings

**1. `epic_base_unavailable` hold not scoped to its own epic — FIXED.** The gate now compares the two branch names rather than merely testing both for truthiness: `if (state.epicIntegrationBranch && state.epicBaseUnavailable === state.epicIntegrationBranch)` (src/drain-core.ts:305). I verified the two values are drawn from the same source, so the equality actually holds in the good case rather than silently never firing: `state.epicIntegrationBranch` is the pinned name from `getOrInitEpicIntegrationBranch` (src/drain.ts:716), the spawn decision carries that same value as `integrationBranch` (src/drain-core.ts:336), and `resolveSpawnBase` throws `EpicBaseUnavailableError(decision.integrationBranch, err)` (src/drain.ts:2380), whose `integrationBranch` is what `recordSpawnFailure` stores as `epicBase` (src/drain.ts:771) and what `freshEpicBaseFailure` returns (src/drain.ts:786). So epic A's stale marker is now inert against epic B, and the epic's own marker still holds.

**2. Truncation notice emitted inside the untrusted fence — FIXED.** `more()` now returns its own array entries which are pushed *after* the `fenceUntrusted(...)` string, not concatenated into the fenced content (src/critic-core.ts:281-286 for paths, 288-292 for commits). `fenceUntrusted` returns a self-contained block ending in the closing `⟦/UNTRUSTED:…⟧` marker (src/untrusted.ts:44-50), so the shepherd-authored "TRUNCATED — run `git …` for the full list" line now lands outside the region the model is told to discount. The interpolations in that line are a number and a shepherd-built command string containing only the hex `baseSha`, so nothing untrusted rides out of the fence with it.

**3. `integrationBranchSupported` false for both "forge lacks ensureBranch" and "no forge" — FIXED.** `forgeCanEnsureBranch` now returns `true` for an absent forge and only `false` for a resolved forge without `ensureBranch` (src/drain.ts:63-65, used at src/drain.ts:507). `assembleEpic` gates on `=== false` (src/epic-model.ts:164), so a cache-hit build with a since-removed forge no longer asserts `NO_INTEGRATION_BRANCH_WARNING` ("This forge cannot create branches", src/epic-model.ts:70).

## What I verified in the new code

- **Allowlist conformance of the epic block.** Every command the block names is reachable under the `reviewer` preset: `Bash(git diff *)`, `Bash(git log *)`, `Bash(git show *)`, `Bash(git status)` (src/transient-agent-argv.ts:84-92, 123). `git diff --name-only HEAD...<sha>` (src/critic-core.ts:261), `git show <sha>:<path>` (src/critic-core.ts:309-310) and `git log -S… --oneline <sha>` (src/critic-core.ts:312) all match a prefix in that list, and the block's claim that `git grep` / `git ls-tree` are denied (src/critic-core.ts:315) is correct — neither appears in the preset.
- **The base objects the prompt tells the critic to read really are present.** `defaultComputePatchId` fetches the base into the same worktree and rev-parses it to a SHA (src/critic-core.ts:436-456), and that same `wt.worktreePath` + `baseSha` are what `resolveEpicContext` passes to `collectBaseDelta` and into the prompt (src/review.ts:377-378, 668-676). So `git show <baseSha>:<path>` resolves in the critic's checkout.
- **Prompt-injection surface of the new interpolations.** `epic.base` reaches the prompt unfenced, but both producers are gated by `isEpicIntegrationBranch` (src/review.ts:377, src/standalone-critic.ts:381), whose regex `^epic\/\d+(-[a-z0-9-]+)?$` (src/epic-branch.ts:19-21) admits no prompt-breaking characters. `baseSha` comes from `git rev-parse --verify` and is additionally shape-checked (`/^[0-9a-f]{7,40}$/`, src/critic-core.ts:61) before it reaches argv in `defaultCollectBaseDelta`. Path and commit-subject lists — the only genuinely agent-authored strings — are fenced (src/critic-core.ts:282, 290).
- **i18n parity.** `drain_paused_epic_base` and `epic_hold_epic_base_unavailable` exist in both catalogs the repo maintains, with the same `{branch}` parameter (ui/messages/en.json:3141-3142, ui/messages/de.json:3141-3142); `ui/messages/` contains only `en.json` and `de.json`. Both are consumed with `{ branch: desig }` where `desig = drain.detail ?? ""` (ui/src/lib/components/epic-panel.ts:58,76 and ui/src/lib/components/queue-strip.ts:251), and `detail` is set to the branch at src/drain-core.ts:308. `DrainStatus.reason` is a bare `string` on the UI side (ui/src/lib/types.ts:781), so the new code adds no unhandled union member.
- **Contract changes have no stale consumers.** `spawnFailures`' value type changed from `number` to `{ at, epicBase? }`; every read site was updated (src/drain.ts:771, 785, 2402-2407, 2490). `InFlight` gained a required `epicBase` and the only construction site sets it (src/review.ts:456). `reviewPrompt` / `prReviewPrompt` gained a trailing optional param — the only production callers are src/review.ts:651 and src/standalone-critic.ts:441, both updated.
- **The hold self-heals.** The `epic_base_unavailable` hold prevents `doSpawn` from running, and `doSpawn` owns the lazy expiry delete (src/drain.ts:2407) — so the freshness test genuinely has to live in `freshEpicBaseFailure` (src/drain.ts:788), as the comment says, or the hold would latch forever. It does. The hold sits after the retire gate (src/drain-core.ts:284) so a ready PR still retires, and `doSpawn`'s catch still releases the `ACTIVE_LABEL` claim on the new throw path (src/drain.ts:2495-2502).

## Nit (non-blocking)

The EPIC CONTEXT block asserts unconditionally that "Sibling children have ALREADY MERGED into that base" (src/critic-core.ts:300). For the **first** child of an epic that is false: the integration branch was just created from the default branch, so nothing has merged since the fork, `defaultCollectBaseDelta` returns `null` on the `!allPaths.length && !commits.length` path (src/critic-core.ts:100), and `epicDeltaLines` then falls into its no-delta branch and tells the critic to "Enumerate what your tree cannot see" (src/critic-core.ts:258-262) — i.e. to spend a round running two commands that shepherd already knows return nothing, against a stated premise that is untrue for that PR. It's a wording/efficiency issue, not a correctness one (an empty enumeration leads to no bad finding), which is why it is not blocking.

## Stated limitations

- Whether the reviewing model actually honours the two VERIFY overrides (src/critic-core.ts:309, 329) is a behavioural property of the LLM, not something I can verify against code in this repo. The structural preconditions the overrides depend on (allowlist, object availability, the `scopeFindings` path-prefix backstop the block warns about) all check out, so I make no claim either way about the model's compliance.

## Design tradeoff worth an explicit ack (not a finding)

In degraded mode (`baseSha === null` — the base fetch and the local rev-parse both failed), the block instructs the critic that *any* "missing/undefined/not-added" conclusion is unverifiable and "NOT a finding" (src/critic-core.ts:329). That is broader than the stale-tree problem it targets: it also suppresses a legitimate finding about an identifier the child's own diff invents and that never existed on any base. The blast radius is small — `baseSha` is null only when both the `origin` fetch and the local ref lookup fail (src/critic-core.ts:436-459), in which case `computePatchId` is degraded too — and the failure direction is a false negative rather than a false accusation, so I read it as a deliberate fail-safe rather than a defect. Flagging it only so the choice is explicit.

## Issue coverage

Items 1-3 of the issue's proposed fix are implemented (base-delta grounding + VERIFY overrides; epic-aware prompt; the `ensureBranch` throw now fails closed and surfaces as an operator-visible hold, while the `ensureBranch`-absent degrade surfaces as an epic warning). Item 4 ("decide whether the landing-PR review should stay a full re-review") is a product decision the issue explicitly framed as a decision, not a change; leaving `standalone-critic`'s landing-PR behaviour as-is is a defensible answer and I do not treat its absence as a gap.

scoop (commented):
## Prior finding — re-checked

**"The EPIC CONTEXT block states unconditionally that siblings have ALREADY MERGED, which is false for the first child."** — **Addressed.** Two changes carry it:

1. `defaultCollectBaseDelta` (src/critic-core.ts:61-106) now returns an *empty* `EpicBaseDelta` rather than `null` when git ran and found nothing (src/critic-core.ts:97-102); `null` is reserved for a git failure (src/critic-core.ts:103).
2. `epicBlock` (src/critic-core.ts:311-327) computes `knownEmpty` (src/critic-core.ts:315) and, when true, emits the first-child wording — "NOTHING has merged into that base since this branch forked (you are its first child), so your worktree is current with the base" (src/critic-core.ts:319-321) — and returns early (src/critic-core.ts:327), so neither the stale-tree override nor `epicDeltaLines`' enumeration instruction is emitted. `epicDeltaLines` is consequently only reachable for a nonempty or unknown delta (src/critic-core.ts:280-283).

## Verification performed

- **Reviewer allowlist conformance.** The epic block names `git diff`, `git log`, `git show` (src/critic-core.ts:288, 298, 336-341) and explicitly forbids `git grep` / `git ls-tree` (src/critic-core.ts:342). The `reviewer` preset is `[...READONLY_GIT, "Write"]` (src/transient-agent-argv.ts:123) where `READONLY_GIT` = `Read, Grep, Glob, Bash(git diff *), Bash(git log *), Bash(git show *)` (src/transient-agent-argv.ts:84-90). Every named command is allowed; the two denied ones are correctly called out. The new `allowedToolsFor` export (src/transient-agent-argv.ts:119-121) returns a copy of that preset.
- **Base objects are local when `git show <sha>:<path>` runs.** `defaultComputePatchId` fetches the base from origin into the same worktree before rev-parsing it (src/critic-core.ts:463-476), and the critic runs in that same worktree (`wt.worktreePath` is threaded to both `computePatchId` and `collectBaseDelta` in src/review.ts:736 and src/standalone-critic.ts:381-385). When the fetch/rev-parse fails, `baseSha` is null and `resolveEpicContext` skips the shell-out (src/review.ts:802), and `epicBlock` degrades to its no-base mode (src/critic-core.ts:344-350) — a Grep miss there becomes a CANNOT-VERIFY limitation, not a finding.
- **Hold code plumbed end-to-end.** `epic_base_unavailable` is added to `HoldReason.code` (src/drain-core.ts:40), raised in `computeNext` after the retire gate and before the cap (src/drain-core.ts:305-310), included in the `paused` code set (src/drain.ts:816-509 region), and handled in both consuming switches: `pausedText` (ui/src/lib/components/queue-strip.ts:251-252) and `epicHoldLine` (ui/src/lib/components/epic-panel.ts:73-76). Those are the only two switches over drain hold reasons (grep for `awaiting_signoff` finds no other consumer). UI `DrainStatus.reason` is `string | null` (ui/src/lib/types.ts:781), so no type widening was needed.
- **i18n parity.** `drain_paused_epic_base` and `epic_hold_epic_base_unavailable` exist in both catalogs the repo maintains — ui/messages/en.json:3140-3141 and ui/messages/de.json:3140-3141 — with the same `{branch}` parameter used at the two call sites above. `ui/messages/` contains only `en.json` and `de.json`.
- **Cross-epic / cross-repo staleness of the marker.** `freshEpicBaseFailure` (src/drain.ts:481-490) scans only keys prefixed `${repoPath}#`, applies the `SPAWN_FAIL_COOLDOWN_MS` freshness test itself (necessary, since the lazy expiry delete in `doSpawn` at src/drain.ts:2410 is unreachable while the hold suppresses spawns), and `computeNext` additionally requires `state.epicBaseUnavailable === state.epicIntegrationBranch` (src/drain-core.ts:305) so a marker from a previous epic in the same repo is inert.
- **Typed failure, not any failure.** `recordSpawnFailure` tags `epicBase` only for `EpicBaseUnavailableError` (src/drain.ts:468-471), which is thrown solely from the `ensureBranch` catch in `resolveSpawnBase` (src/drain.ts:2380). That throw sits inside `doSpawn`'s try (src/drain.ts:2477), so the claim label is still released on the existing catch path (src/drain.ts:2494-2499). A `service.create()` failure therefore cannot masquerade as an epic-wide hold.
- **Forge without `ensureBranch` still degrades.** `forgeCanEnsureBranch` (src/drain.ts:394-396) returns `true` for a null forge (`resolveForge` is `(repoPath) => GitForge | null`, src/drain.ts:215), so the warning is only asserted when a forge actually resolved and lacks the method; `assembleEpic` gates on `=== false` (src/epic-model.ts:161-162), leaving every existing caller unaffected.
- **Signature changes.** `reviewPrompt` / `prReviewPrompt` gained a trailing optional `epic`; the only callers are src/review.ts:651 and src/standalone-critic.ts:441, both updated. `steerText` gained a required third arg and has one caller (src/review.ts:1021), fed from the new `InFlight.epicBase`. `spawnFailures` changed from `Map<string, number>` to `Map<string, {at, epicBase?}>`; the only read sites are src/drain.ts:2403-2411 and `freshEpicBaseFailure`, both updated.

## Nit (see findings)

The one inconsistency I found is in `epicBlock`'s empty-detection: it requires *both* `paths` and `commits` to be empty (src/critic-core.ts:315), but the completeness property the block rests on is stated purely in terms of paths ("any path NOT listed is identical to what your tree already shows", src/critic-core.ts:288). When the base has commits since the fork whose net three-dot diff is empty (an add-then-revert on the integration branch, or an empty commit), `paths` is `[]` while `commits` is nonempty, so `knownEmpty` is false and the full stale-tree block is emitted: it asserts "Sibling work merged into the base after this branch forked is ABSENT from the tree" (src/critic-core.ts:329) when nothing actually is, and `epicDeltaLines` prints "...is exactly the paths below ...:" (src/critic-core.ts:288) followed by no list at all, because the path fence is guarded by `if (delta.paths.length)` (src/critic-core.ts:290) while the commits section is not (src/critic-core.ts:296). Rare and non-blocking, but the prompt it produces is self-contradictory.

scoop (commented):
## Prior finding — re-checked

**`knownEmpty` keyed on both `paths` and `commits`** — FIXED. `src/critic-core.ts:322` now reads `const knownCurrent = !!known && known.paths.length === 0;` — the discriminator is the path list alone, so a base that gained commits with an empty net three-dot diff (revert pair / empty commit) is classified KNOWN-CURRENT and `epicBlock` returns early at `src/critic-core.ts:334`, never emitting the "ABSENT from the tree" text. The known-current sentence at `src/critic-core.ts:327` even branches on `known!.commits.length` to say "commits have landed on it, but their net diff against your fork point is empty". Separately, the intro sentence in `epicDeltaLines` is now guarded by `if (delta.paths.length)` (`src/critic-core.ts:290`), so the promise-a-list-then-emit-no-list malformation is impossible even if the two ever drift. Both halves of the prior finding are addressed.

## What I verified

- **Identifiers resolve.** `isEpicIntegrationBranch` is exported from `src/epic-branch.ts:19` and imported in `src/review.ts:41` and `src/standalone-critic.ts:28`. `defaultCollectBaseDelta` / `EpicBaseDelta` / `EpicContext` are exported from `src/critic-core.ts:73,20,34` and imported at `src/review.ts:20,38,39` and `src/standalone-critic.ts:36,40`. `allowedToolsFor` is a new export at `src/transient-agent-argv.ts:126`.
- **The commands the epic block names are actually runnable by the critic.** The `reviewer` preset is `[...READONLY_GIT, "Write"]` (`src/transient-agent-argv.ts:123`) and `READONLY_GIT` is `Read, Grep, Glob, Bash(git diff *), Bash(git log *), Bash(git show *), Bash(git status)` (`src/transient-agent-argv.ts:84-92`). Every command the block instructs (`git diff --name-only HEAD...<sha>`, `git log --oneline HEAD..<sha>`, `git log -S… <sha>`, `git show <sha>:<path>`, `git show <hit-sha>`) matches one of those globs; `git grep` / `git ls-tree` match none, which is exactly what `src/critic-core.ts:349` tells the critic.
- **The base objects are local when the block tells the critic to read them.** `defaultComputePatchId` fetches the base (`git fetch origin -- <base>`, `src/critic-core.ts:470-472`) and rev-parses it to a SHA before returning `baseSha`, and `resolveEpicContext` is only called with that same `baseSha` (`src/review.ts:378`), so `git show <baseSha>:<path>` resolves in the critic's disposable worktree. Same in `src/standalone-critic.ts:381-386`, which passes `baseSha` from the `computePatchId` call at `:348`.
- **Argv safety.** `baseSha` is shape-checked (`/^[0-9a-f]{7,40}$/`, `src/critic-core.ts:77`) before reaching `execFile`, and `epic.base` is only ever set from a string that already passed `isEpicIntegrationBranch` (`src/review.ts:377`, `src/standalone-critic.ts:381`), whose regex is `^epic\/\d+(-[a-z0-9-]+)?$` (`src/epic-branch.ts:20`) — so nothing shell-metacharacter-ish reaches the steer text at `src/review.ts:59`.
- **Hold plumbing is complete.** `epic_base_unavailable` is added to the `HoldReason` union (`src/drain-core.ts:40`), raised in `computeNext` after the retire gate and before the cap (`src/drain-core.ts:305-309`), mapped into `paused` (`src/drain.ts:815-828`), and rendered in both hold-code switches that exist: `pausedText` (`ui/src/lib/components/queue-strip.ts:254`) and `epicHoldLine` (`ui/src/lib/components/epic-panel.ts:76`). I grepped for other consumers of the drain hold code — `src/hold.ts:133` is a *different* `HoldReason` (session holds, `src/hold.ts` catalogue), and `src/automerge.ts:252,414` uses the automerge decision codes — so no switch is left without a case.
- **The hold can actually fire and can lapse.** `decision.integrationBranch` in the spawn decision comes from `state.epicIntegrationBranch` (`src/drain-core.ts:336`), the same value `EpicBaseUnavailableError` carries out of `resolveSpawnBase` (`src/drain.ts:2378`) and that `computeNext` compares against (`src/drain-core.ts:305`) — so the branch-name equality guard is not inert. Freshness is tested in `freshEpicBaseFailure` (`src/drain.ts:488-497`) rather than by map membership, which matters because the lazy cooldown delete lives in `doSpawn` (`src/drain.ts:2407`) and the hold prevents `doSpawn` from running; without that the epic would latch forever. Verified the typed-error narrowing in `recordSpawnFailure` (`src/drain.ts:476`) so an ordinary `service.create()` failure can't be mislabeled as an epic-wide base hold.
- **i18n parity.** Both new keys exist in both catalogues the repo maintains (only `ui/messages/en.json` and `ui/messages/de.json` exist): `drain_paused_epic_base` and `epic_hold_epic_base_unavailable` at `ui/messages/en.json:3141-3142` and `ui/messages/de.json:3141-3142`, each with the same single `{branch}` parameter used by the call sites.
- **Task satisfaction.** Issue items 1–3 are covered: the critic is handed the base delta and instructed to read the base rather than trust a stale-worktree grep (`src/critic-core.ts:335-354`); the prompt is epic-aware (`src/critic-core.ts:325-329`); the `ensureBranch` throw now fails closed and surfaces as an operator-visible hold, while the forge-lacks-`ensureBranch` degrade surfaces as an epic warning (`src/epic-model.ts:59-62`, `NO_INTEGRATION_BRANCH_WARNING`). Item 4 (landing-PR scope) is a "decide" item and is deliberately left unchanged — `src/standalone-critic.ts:284` still reviews the landing PR against its real base; that reads as an accepted product call, not a defect.

## Nit

In the UNKNOWN-delta case (`epic.delta === null`, i.e. the git collection failed) the block still asserts as fact: "Sibling children have ALREADY MERGED into that base" (`src/critic-core.ts:328`). For an epic's *first* child whose delta collection happened to fail, that is simply untrue. The safety posture is right (assume stale), but the sentence states as ground truth something the code explicitly documents as ignorance (`src/critic-core.ts:314`). Hedging it ("sibling children MAY have already merged — you could not enumerate them") costs nothing and keeps the prompt honest, which matters for an instruction block whose whole purpose is telling the critic what it can and cannot conclude.

## Stated limitations

- I did not run tests or lint (per instructions); CI owns those.

scoop:
## Re-review of the prior revision's point

**Prior finding 1 — UNKNOWN-delta branch asserts "Sibling children have ALREADY MERGED": ADDRESSED.** `src/critic-core.ts:333` now reads "Sibling children MAY ALREADY HAVE MERGED into that base — the delta could NOT be enumerated here, so treat this as unknown, not as established fact." The stale-tree machinery below still ships in the conservative form (`src/critic-core.ts:347`), and `test/critic-core.test.ts:662-667` locks it in. Not re-raised.

## What I verified

- **Three prompt states are wired correctly.** `epicBlock` discriminates on the PATH list, not the commit list (`src/critic-core.ts:321-322`), and `epicDeltaLines` is only reached for a nonempty or null delta because `knownCurrent` returns early (`src/critic-core.ts:343`) — so the "is exactly the paths below" promise (`src/critic-core.ts:294`) is never made without a list.
- **The git the prompt tells the critic to run is actually permitted.** The prompt names `git diff`, `git log`, `git show` (`src/critic-core.ts:354-360`); the `reviewer` preset allows `Bash(git diff *)`, `Bash(git log *)`, `Bash(git show *)` via `READONLY_GIT` (`src/transient-agent-argv.ts:84-93`, used at `:123`), and does not allow `git grep`/`git ls-tree` — which the block explicitly tells the critic not to attempt (`src/critic-core.ts:360`). `allowedToolsFor` (`src/transient-agent-argv.ts:134`) exposes that set for the conformance test.
- **The base objects the prompt points at are present.** `defaultComputePatchId` fetches the base into the same worktree and rev-parses it to a SHA (`src/critic-core.ts:481-501`), and `defaultCollectBaseDelta` / the prompt both key off that SHA (`src/review.ts:378`, `src/standalone-critic.ts:381-386`), so `git show <sha>:<path>` resolves.
- **Delta collection semantics.** `git diff --name-only -z HEAD...<baseSha>` is three-dot, so the merge base is the fork point and the path list really is the base-side content the child's tree cannot see (`src/critic-core.ts:70`); `git log --oneline HEAD..<baseSha>` (two-dot, `src/critic-core.ts:82`) is exactly the sibling commits on the base. Both are capped and the truncation counts are emitted outside the untrusted fence (`src/critic-core.ts:222-227`). `fenceUntrusted(label, content)` matches its signature at `src/untrusted.ts:38`.
- **`isEpicIntegrationBranch`** exists and is imported in both consumers (`src/epic-branch.ts:19`, `src/review.ts:30`, `src/standalone-critic.ts:28`); `Session.baseBranch` is a non-optional `string` (`src/types.ts:21`), so the guard is total.
- **The new hold code.** `epic_base_unavailable` is added to the `HoldReason` union (`src/drain-core.ts:40`), raised only when the marker names THIS epic's branch (`src/drain-core.ts:305`), sits after the retire gate and before the cap (`src/drain-core.ts:284-317`), is listed in the `paused` set (`src/drain.ts:817-827`), and the marker is derived cooldown-fresh each tick (`src/drain.ts:499-508`) so it lapses and retries — the lazy expiry delete in `doSpawn` (`src/drain.ts:2404-2407`) is unreachable while the hold is up, which is exactly why the freshness test lives in `freshEpicBaseFailure`. `recordSpawnFailure` tags only the typed `EpicBaseUnavailableError` (`src/drain.ts:487`), so a sandbox/provider/create failure cannot mislabel the hold. `resolveSpawnBase` throws inside `doSpawn`'s try, whose catch releases the claim label (`src/drain.ts:2494-2502`).
- **i18n parity.** `drain_paused_epic_base` and `epic_hold_epic_base_unavailable` exist in BOTH catalogs with the same `{branch}` param — `ui/messages/en.json:3140-3141` and `ui/messages/de.json:3140-3141` — and are consumed at `ui/src/lib/components/queue-strip.ts:254` and `ui/src/lib/components/epic-panel.ts:76`, both passing `branch: desig` where `desig = drain.detail ?? ""` (`queue-strip.ts:239`, `epic-panel.ts:58`) and `detail` is the branch (`src/drain-core.ts:308`). `DrainStatus.reason` is `string | null` (`ui/src/lib/types.ts:781`), so no UI type change is needed.
- **`integrationBranchSupported`** is optional and only warns on `=== false` (`src/epic-model.ts:164`), so existing callers are unaffected; `forgeCanEnsureBranch` reports `true` for an absent forge (`src/drain.ts:412-414`), matching its documented "don't assert a fact we couldn't check" rule, and `resolveForge` is sync returning `GitForge | null` (`src/drain.ts:215`).

## The one problem

`src/critic-core.ts:330` — the KNOWN-CURRENT (empty-delta) header. When the delta has no paths and no commits it tells the critic: `(nothing has merged into it since this branch forked — you are its first child)`. The first clause is established by the git we ran; **the second is not**. An empty delta means only that the base has no commits the child's HEAD doesn't already reach — which is exactly what happens for the *n*-th child that was spawned from an already-up-to-date integration tip (`resolveSpawnBase` bases each child on the integration branch as it stands at spawn, `src/drain.ts:2374-2377`). That is the common healthy case, not a rare one. The prompt then states as ground truth that no sibling has ever merged, when in fact sibling work merged *before* the fork and is sitting in the critic's tree.

This is the same defect class the previous round blocked on, and the code's own comment sets the standard (`src/critic-core.ts:323-328`: "Ignorance is stated as ignorance"; here knowledge is over-stated). `test/critic-core.test.ts:635-646` encodes the same conflation in its name and assertions, so it would need the matching update.

Dropping the trailing clause (keeping "nothing has merged into it since this branch forked") is sufficient — the true statement is the one the rest of the block relies on.

## Product call left open (informational)

The issue's item 4 (should the landing-PR critic keep re-reviewing the whole accumulated epic diff against the default branch?) is answered implicitly: the standalone critic still takes the PR's real base (`src/standalone-critic.ts:348,376`) and the epic block is null for a landing PR (its base is the default branch, `:381`). That is a deliberate no-op, documented in the comment at `src/standalone-critic.ts:377-380`.

scoop (commented):
## Prior finding — re-check

**1. `src/critic-core.ts`: KNOWN-CURRENT header claims "you are its first child".** ADDRESSED in the prompt. The header is now built at `src/critic-core.ts:334-338`; the empty-delta branch reads `"… (nothing has merged into it since this branch forked), so your worktree is CURRENT with the base — any sibling work that merged BEFORE you forked is already in your tree."` — the "first child" clause is gone, and the reasoning comment at `src/critic-core.ts:328-333` now states exactly why (a child spawned off an already-up-to-date integration tip lands here too, with siblings merged *before* the fork). The test that encoded the conflation was updated: `test/critic-core.test.ts:635-657` now asserts `expect(p).not.toContain("first child")` (`:651`) and the retained-claim assertions at `:646-648`.

## What I verified

- **Prompt is emitted only for epic children, and only where the rules it overrides live.** `epicBlock` is spliced into `scopeAndOutputTail` immediately after the ATTRIBUTION paragraph (`src/critic-core.ts:418-422`), and is `[]` when `epic` is falsy, so non-epic prompts stay byte-identical. Both call sites pass it: `src/review.ts:651` (`reviewPrompt(..., epic)`) and `src/standalone-critic.ts:441` (`prReviewPrompt(..., fp.epic ?? null)`); those are the only callers (`grep reviewPrompt(|prReviewPrompt(` across `src/`).
- **Every command the epic block names is on the reviewer allowlist.** The block instructs `git show <sha>:<path>`, `git diff --name-only HEAD...<sha>`, `git log --oneline`/`log -S` (`src/critic-core.ts:355-364`) and explicitly forbids `git grep`/`git ls-tree` (`:365`). The `reviewer` preset is `[...READONLY_GIT, "Write"]` (`src/transient-agent-argv.ts:123`) and `READONLY_GIT` is `Read, Grep, Glob, Bash(git diff *), Bash(git log *), Bash(git show *), Bash(git status)` (`src/transient-agent-argv.ts:84-92`) — `git grep`/`git ls-tree` are indeed absent, so naming them as denied is correct. The new `allowedToolsFor` export (`src/transient-agent-argv.ts:134-136`) is consumed only by `test/critic-core.test.ts:19,527`.
- **Base objects are local when the delta is collected.** `defaultCollectBaseDelta` requires an already-resolved `baseSha` (`src/critic-core.ts:73-77`), which `defaultComputePatchId` produces from a fresh `git fetch origin -- <base>` + `git rev-parse` in the same worktree (`src/critic-core.ts:486-506`) — so `git diff HEAD...<sha>` and `git show <sha>:<path>` can resolve. Direction is right: `A...B` diffs merge-base(A,B)→B, so `HEAD...baseSha` is exactly the base-side content the child's tree lacks (`src/critic-core.ts:82`), and `git log HEAD..baseSha` (`:94`) lists base-only commits.
- **Injection surface.** Both embedded listings are fenced with `fenceUntrusted` (`src/critic-core.ts:293,300`), whose closing marker carries an unguessable nonce and scrubs pre-existing fence tokens (`src/untrusted.ts:38-50`); the truncation notices are emitted outside the fence (`src/critic-core.ts:279-286`), and `test/critic-core.test.ts:628-631` asserts that placement. `baseSha` is shape-checked against `/^[0-9a-f]{7,40}$/` before reaching argv (`src/critic-core.ts:77`).
- **Fail-closed spawn path.** `resolveSpawnBase` now throws `EpicBaseUnavailableError` when `ensureBranch` throws (`src/drain.ts:2375-2380`); its only caller is inside `doSpawn`'s try (`src/drain.ts:2471`), whose catch records the typed failure (`:2495` → `recordSpawnFailure`, `src/drain.ts:2491-2494`) and releases the claim label (`:2499`). The hold is derived per tick from `freshEpicBaseFailure` (`src/drain.ts:2504-2513`), which applies the 5-minute cooldown itself — necessary, since the lazy expiry delete lives in `doSpawn` (`src/drain.ts:2407`) which the hold prevents from running, so the hold does lapse and retry. Key scoping (`${repoPath}#`) matches the `failKey` format at `src/drain.ts:2401`.
- **Hold plumbing.** `epic_base_unavailable` is added to the `HoldReason` union (`src/drain-core.ts:40`), gated on `state.epicIntegrationBranch === state.epicBaseUnavailable` after the retire/trouble gates and before the cap (`src/drain-core.ts:305-310`), added to the `paused` set (`src/drain.ts:819-832`), and rendered by both switches: `pausedText` (`ui/src/lib/components/queue-strip.ts:251-253`) and `epicHoldLine` (`ui/src/lib/components/epic-panel.ts:73-76`), both reading `detail` as the branch (`epic-panel.ts:58`, `queue-strip.ts:239`).
- **i18n parity.** `drain_paused_epic_base` and `epic_hold_epic_base_unavailable` exist in BOTH catalogs — `ui/messages/en.json:3141-3142` and `ui/messages/de.json:3141-3142` — with the same `{branch}` parameter in each. `ui/messages/` contains only `en.json` and `de.json`, so parity is complete.
- **`NO_INTEGRATION_BRANCH_WARNING` reaches the operator.** Pushed into `warnings` only on an explicit `=== false` (`src/epic-model.ts:164`), set from `forgeCanEnsureBranch(resolveForge(repoPath))` (`src/drain.ts:507`, helper at `src/drain.ts:73-75`; `resolveForge` returns `GitForge | null`, `src/drain.ts:215`). It is not filtered out by `isStructurallyCovered` (`src/epic-diagnosis.ts:45-51`), so it flows verbatim into `additionalWarnings` (`src/epic-diagnosis.ts:124`) and renders in `EpicDiagnosisModal.svelte:194-201`, plus bumping the count line in `EpicPanel.svelte:145-147`. (Note: `EpicPanel` itself shows only "N warning(s)" — the text is visible only in the diagnosis modal. That's the pre-existing warning-surfacing design, not something this PR changes.)
- **Contract changes have no stragglers.** `steerText` gained a third parameter and its single caller passes `f.epicBase` (`src/review.ts:1021`), which is a required field on `InFlight` (`src/review.ts:90`) set at `src/review.ts:456`. `EpicContext`/`EpicBaseDelta` are exported from `critic-core` and imported in both consumers (`src/review.ts:20-21,24`, `src/standalone-critic.ts:36,41`); `isEpicIntegrationBranch` is imported in `src/standalone-critic.ts:28` and `src/review.ts:25`.

## Against the issue

Root cause 1 is handled by the option the issue explicitly allows ("at minimum have the critic fetch the integration branch"): the delta is collected server-side and the VERIFY grep rule is overridden for base-delta paths, with presence AND absence both requiring a blob read (`src/critic-core.ts:353-357`). Root cause 2 (epic-aware prompt) and root cause 3 (surface the `ensureBranch` fallback — as a hold when it throws, as an epic warning when the forge lacks the method) are both addressed. Item 4 of the issue is a product decision and is left as-is, which is a defensible reading of "decide".

## Minor observation (not a finding)

In `defaultCollectBaseDelta` a failing `git log` is swallowed to `commits = []` (`src/critic-core.ts:101-103`), which is indistinguishable from "no commits" at `src/critic-core.ts:335`, where an empty `commits` makes the header assert "nothing has merged into it since this branch forked". In practice this is unreachable: `git log` can only fail here after `git diff` already resolved the same object, and the one realistic failure (16 MiB `maxBuffer` overflow) implies a large history that would leave `paths` non-empty, which is not the `knownCurrent` branch. Flagging it only because the alternative would be to carry a `commitsUnknown` flag; I don't think it's worth the field.

scoop (commented):
## Prior finding — resolved

**"`EpicBaseDelta`'s comment and the NB in `defaultCollectBaseDelta` still gloss an empty delta as 'the epic's first child'"** — fixed. `src/critic-core.ts:21-24` now reads "An empty delta does NOT mean 'no sibling has ever merged' — do not infer that here or in the prompt", and the NB at `src/critic-core.ts:97-101` likewise says an empty delta "does NOT license 'no sibling has ever merged' — see EpicBaseDelta". A repo-wide grep for `first child` (`src/critic-core.ts:334`) leaves exactly one occurrence, and it is the correct one: the hedge explaining why the UNKNOWN-delta header must not assert that siblings merged. That matches the corrected reasoning in `epicBlock` (`src/critic-core.ts:313-345`) and the test at `test/critic-core.test.ts:635-651`.

## Verification performed

- **Prompt commands vs. the reviewer allowlist.** The epic block tells the critic to run `git show <sha>:<path>`, `git diff --name-only HEAD...<sha>`, `git log --oneline`/`log -S`, and explicitly forbids `git grep` / `git ls-tree`. The `reviewer` preset is `[...READONLY_GIT, "Write"]` (`src/transient-agent-argv.ts:123`) and `READONLY_GIT` is `Read, Grep, Glob, Bash(git diff *), Bash(git log *), Bash(git show *)` (`src/transient-agent-argv.ts:84-90`) — every named command is permitted, and the two named as denied genuinely are. The new `allowedToolsFor()` export (`src/transient-agent-argv.ts:129-136`) returns a copy, so the test seam can't mutate `PRESETS`.
- **Base objects are actually present for `git show <baseSha>:…`.** `defaultComputePatchId` fetches the base into the critic's own worktree and rev-parses it (`src/critic-core.ts:491-511`), and `defaultCollectBaseDelta`/the prompt only ever use that resolved SHA — a null `baseSha` degrades to the no-base branch of `epicBlock` (`src/critic-core.ts:370-380`) with no base commands emitted. The `baseSha` argv guard (`src/critic-core.ts:70`) rejects anything that is not a hex SHA before it reaches `execFile`.
- **Three-dot direction.** `git diff --name-only -z HEAD...<baseSha>` (`src/critic-core.ts:75`) yields base-side content since the fork point, which is exactly what the prompt claims about it, and `git log --oneline HEAD..<baseSha>` (`src/critic-core.ts:87`) the base-only commits. The empty/null distinction is preserved end-to-end (`src/critic-core.ts:102-110` → `src/critic-core.ts:325-330`).
- **Hold-code plumbing.** `epic_base_unavailable` is added to `HoldReason.code` (`src/drain-core.ts:40`), gated on `state.epicIntegrationBranch === state.epicBaseUnavailable` (`src/drain-core.ts:305-311`). The spawn decision's `integrationBranch` is `state.epicIntegrationBranch` (`src/drain-core.ts:336`), which is what `EpicBaseUnavailableError` carries (`src/drain.ts:2385`), so the two provably compare equal for the current epic and a stale cross-epic marker stays inert. The marker is only ever set for the typed error (`src/drain.ts:783-786`) and the freshness test lives in `freshEpicBaseFailure` (`src/drain.ts:795-805`) rather than the map, which is required because the lazy expiry delete lives inside `doSpawn` (`src/drain.ts:2405-2408`) — the very call the hold suppresses. Self-healing therefore works: cooldown lapses → hold clears → spawn retries.
- **Fail-closed base resolution.** `resolveSpawnBase` now throws instead of silently degrading (`src/drain.ts:2375-2382`); its only call site is inside `doSpawn`'s try (`src/drain.ts:2471`), whose catch records the failure and releases the `ACTIVE_LABEL` claim (`src/drain.ts:2494-2501`). No other call site exists.
- **Hold-code consumers.** The only UI consumers of `HoldReason.code` are `epic-panel.ts` (`ui/src/lib/components/epic-panel.ts:73-74`) and `queue-strip.ts` (`ui/src/lib/components/queue-strip.ts:251-252`); both got a case, and `pausedText` is reachable because `epic_base_unavailable` was added to the `paused` set (`src/drain.ts:819-828`). `DrainStatus.reason` is `string | null` (`ui/src/lib/types.ts:781`), so the switches type-check.
- **Locale parity.** `drain_paused_epic_base` and `epic_hold_epic_base_unavailable` exist in both catalogs with the same `{branch}` param — `ui/messages/en.json:3140-3141` and `ui/messages/de.json:3140-3141`.
- **Epic warning reaches the operator.** `integrationBranchSupported` is threaded from `forgeCanEnsureBranch(resolveForge(repoPath))` (`src/drain.ts:507`) into the `AssembleInput` used by both `assembleEpic` calls in `buildEpic` (`src/drain.ts:517,530`), and `=== false` gating (`src/epic-model.ts:164`) keeps every existing caller (e.g. `src/up-next.ts:299`) unaffected. The text surfaces as a count in `EpicPanel` (`ui/src/lib/components/EpicPanel.svelte:145-146`) and verbatim in the diagnosis modal via `additionalWarnings` (`src/epic-diagnosis.ts:124`, `ui/src/lib/components/EpicDiagnosisModal.svelte:194-203`) — `isStructurallyCovered` (`src/epic-diagnosis.ts:45-51`) does not filter it out.
- **Untrusted-data handling.** The embedded path/commit listings are fenced (`src/critic-core.ts:241,248` via `fenceUntrusted`, `src/untrusted.ts:38`), and the truncation notice is emitted outside the fence in shepherd's own voice — correct, since the fence preamble instructs the model to discount instructions inside it.

## Observation (not a finding)

For a non-empty delta the header asserts "Sibling children have ALREADY MERGED into that base" (`src/critic-core.ts:341`). A non-empty delta strictly proves only that the base carries content the fork point lacks; the landing-branch rebase (`src/landing-rebase.ts:303`, called at `src/drain.ts:1663`) can also fold default-branch commits onto the integration branch. In practice the landing PR only exists once a child has merged, so the claim holds whenever it fires, and the fenced commit-subject list gives the critic the ground truth either way — so I am not raising it, but the delta's *cause* is not strictly sibling merges.

scoop (commented):
## Verdict

The change satisfies the task. I found no correctness, security, or parity defect. One prior finding (a doc-placement nit) is **still not addressed** and is re-raised.

**1. `src/critic-core.ts`: detached doc comments — NOT ADDRESSED (re-raised).** The two commits since the last revision (`f719dd2c`, `26936d55`) only edited comment *content*, not placement. In the current head: `scopeAndOutputTail`'s JSDoc is at `src/critic-core.ts:216-222`, the EPIC CONTEXT block doc (which describes `epicBlock`) at `src/critic-core.ts:223-263`, and `epicDeltaLines`' own JSDoc at `src/critic-core.ts:264-268` — all three sit immediately above `function epicDeltaLines` (`src/critic-core.ts:269`). `epicBlock` (`src/critic-core.ts:313`) and `scopeAndOutputTail` (`src/critic-core.ts:391`) are left undocumented, and an IDE hover on `epicDeltaLines` shows three stacked contracts, two of which belong to other functions. Same fix as before: move the EPIC CONTEXT block above `epicBlock` and restore `scopeAndOutputTail`'s JSDoc above `scopeAndOutputTail`.

## What I verified

**Prompt commands vs. the reviewer allowlist.** The epic block names `git diff --name-only HEAD...<sha>`, `git log --oneline HEAD..<sha>`, `git log -S<id> --oneline <sha>`, and `git show <sha>:<path>` (`src/critic-core.ts:364-374`, `276`, `300`, `307`), and explicitly declares `git grep` / `git ls-tree` denied (`src/critic-core.ts:370`). Compared against the live `reviewer` preset: `READONLY_GIT` = `Bash(git diff *)`, `Bash(git log *)`, `Bash(git show *)`, `Bash(git status)` plus `Read/Grep/Glob`, and `reviewer: {allowedTools: [...READONLY_GIT, "Write"]}` (`src/transient-agent-argv.ts:83-92`, `126`). Every named command is on the list; the two it declares denied are genuinely absent. The new `allowedToolsFor` export (`src/transient-agent-argv.ts:135-137`) returns a copy of the private preset, so the conformance test can't mutate it.

**`git show <sha>:<path>` is actually resolvable in the critic worktree.** `defaultComputePatchId` fetches the base from origin before rev-parsing it (`src/critic-core.ts:491-511`), so the base objects are in the store even though the epic branch has no local ref — the blob reads the block instructs are shallow-safe as claimed.

**Three-dot direction.** `defaultCollectBaseDelta` runs `git diff --name-only -z HEAD...<baseSha>` (`src/critic-core.ts:75`), i.e. merge-base(HEAD, base)→base — the base-side content the child's fork-point tree cannot see. That is the right direction for the stated purpose. `-z` matches `defaultComputePatchId`'s reason for using it (unquoted non-ASCII paths). The `^[0-9a-f]{7,40}$` guard (`src/critic-core.ts:70`) keeps a hostile ref out of argv, and every failure path returns `null` rather than throwing (`src/critic-core.ts:108-110`), with the `git log` sub-failure degrading to `commits: []` without losing the path list (`src/critic-core.ts:94-96`).

**Empty vs. null delta.** `epicBlock` discriminates on `paths.length`, not `commits.length` (`src/critic-core.ts:331-332`), so a revert-pair that lands commits with an empty net diff produces the "tree is CURRENT" header (`src/critic-core.ts:340`) rather than a "sibling work is ABSENT" block with an empty listing under it. The null (collection-failed) case hedges rather than asserting siblings merged (`src/critic-core.ts:343`). `epicDeltaLines`' `if (delta.paths.length)` guard (`src/critic-core.ts:295`) is redundant with the `knownCurrent` early return (`src/critic-core.ts:353`) but harmless and self-documenting.

**Injection surface.** Both embedded listings are wrapped in `fenceUntrusted` (`src/critic-core.ts:299`, `306`), whose nonce+scrub makes the closing marker unforgeable (`src/untrusted.ts:38-49`). The truncation notice is emitted outside the fence (`src/critic-core.ts:285-290`), so the one line the mechanism's soundness rests on isn't inside a block the model is told to discount.

**Worktree HEAD is the child head in both callers.** `ReviewService.begin` creates a detached worktree at `git.headSha` (`src/review.ts:330`) and only then collects the delta (`src/review.ts:377-378`); the standalone critic detaches at `pr.headSha` (`src/standalone-critic.ts:330-336`) and collects at `src/standalone-critic.ts:381-386`. So `HEAD...<base>` means what the comments say it means in both.

**Epic detection.** `isEpicIntegrationBranch` (`src/epic-branch.ts:19-21`) is imported in both callers (`src/review.ts:30`, `src/standalone-critic.ts:28`) and keyed off `session.baseBranch` (`src/types.ts:21`, non-optional `string`) / `meta.baseRefName` (`src/forge/types.ts:328`, non-optional `string`) — no undefined-into-regex path. The landing PR (head = `epic/<n>-<slug>`, base = default) correctly gets **no** epic block, since the check is on the *base* ref.

**`InFlight.epicBase` threading.** The field is set at the single `inflight.set` site (`src/review.ts:450-457`) and consumed at the single `steerText` call (`src/review.ts:1021`); no other construction site to miss.

**Fail-closed spawn path.** `resolveSpawnBase` now throws `EpicBaseUnavailableError` on an `ensureBranch` throw (`src/drain.ts:2377-2381`) and still degrades (with a warn) when the forge simply lacks `ensureBranch` (`src/drain.ts:2382-2386`). The throw happens inside `doSpawn`'s try (`src/drain.ts:2471`), so the catch still releases the claim label (`src/drain.ts:2494-2502`) — no orphaned claim. `recordSpawnFailure` tags the failure map entry only for the typed error (`src/drain.ts:770`), so an ordinary `service.create()` failure cannot raise an epic-wide hold.

**Hold self-heals.** `freshEpicBaseFailure` applies the cooldown test itself (`src/drain.ts:509-518`) rather than relying on `doSpawn`'s lazy delete — which is load-bearing, since the hold prevents `doSpawn` from running at all; a membership-only check would latch forever. After `SPAWN_FAIL_COOLDOWN_MS` the marker goes stale, `computeNext` stops holding (`src/drain-core.ts:305-310`), `doSpawn` finds `now - lastFail.at >= cooldown`, deletes the entry (`src/drain.ts:2404-2407`) and retries. The `state.epicBaseUnavailable === state.epicIntegrationBranch` comparison makes a cross-epic stale marker inert. Gating on `epicIntegrationBranch` is safe: it is set only inside the `epicActive` branch of `buildState` (`src/drain.ts:700-723`), where `candidates` are the epic's own children (`src/drain.ts:721`), so this can never pause a label-mode drain.

**Hold ordering.** Inserted after the retire gate and before the cap (`src/drain-core.ts:291-310`), so a ready PR still retires while spawns are paused — consistent with how `troubleHold` is placed.

**i18n parity.** `drain_paused_epic_base` and `epic_hold_epic_base_unavailable` exist in **both** catalogs with the same `{branch}` placeholder: `ui/messages/en.json:3145-3146` and `ui/messages/de.json:3145-3146`. Consumers spell them identically: `m.drain_paused_epic_base({branch: desig})` (`ui/src/lib/components/queue-strip.ts:254`, where `desig = d.detail ?? ""` at `:239`) and `m.epic_hold_epic_base_unavailable({branch: desig})` (`ui/src/lib/components/epic-panel.ts:73`). Neither catalog is alphabetically sorted (both append at the end — e.g. `ui/messages/en.json:3-9`), so the append position is consistent with house style. `epic_base_unavailable` is also added to the `paused` code list (`src/drain.ts:826-836`), so `pausedText` is actually reached rather than falling through to the generic copy. I grepped every consumer of `HoldReason.code` (`epic-panel.ts`, `queue-strip.ts`) — both handle the new code; nothing else switches on it.

**Epic warning.** `NO_INTEGRATION_BRANCH_WARNING` is pushed only on an explicit `=== false` (`src/epic-model.ts:164`), and `forgeCanEnsureBranch` returns `true` for an absent forge (`src/drain.ts:63-65`), so a cached epic build served after the forge is gone does not assert a fact it never checked.

## Notes on the issue's items

Item 4 of the issue ("decide whether the landing-PR review should stay a full re-review against the default branch") is left as-is — the landing PR's base is the default branch, so it gets no epic block. That is a deliberate product call the issue itself flagged as optional, not a defect.

PR #1765, opened 2026-07-14
fix(ui): only escalate the review banner on real operator input

scoop:
## The bug

During an in-flight review the terminal banner escalated from calm to **"You're typing — an incoming directive may collide with your input"** whenever the operator merely **moved the mouse** over the terminal (also on click, drag, wheel, and — on mobile — a touch fling).

## Root cause

`Viewport.svelte` bumped the escalation counter on **every** `term.onData` frame. But `onData` is xterm's *outbound data* channel, not a typing signal. With mouse tracking on — which is how Claude Code runs, as `agentOwnsScroll` (`Viewport.svelte:1358-1366`) already documents — xterm reports **every mouse move / click / drag / wheel** to the app on that same channel, along with focus reports and the terminal's replies to the app's own queries. The banner counted all of it as keystrokes.

## The fix

Count operator-input **origins**, never data frames.

1. **Through xterm** — `createTypingCounter` (new `ui/src/lib/terminal-input.ts`) listens on xterm's helper textarea: `keydown` (minus bare modifiers), `input`, `compositionend`, `paste`, `drop`. Mouse reports, focus reports and query replies are synthesized inside xterm's JS and never touch that textarea, so they're excluded *by construction* — no escape-sequence pattern matching on the client.
2. **Bypassing xterm** — Shepherd injects operator input straight into the PTY from six sites (ComposeBar/`sendComposed`, the touch control-key bar, `attachImages`, slash-command link taps, the synthesized Shift+Enter and Escape). **None of these counted before**, and on mobile they are the *primary* input surfaces — so a textarea-only counter would have left the banner permanently silent on a phone. They now call `bumpOperatorInput()`.

Two deliberate carve-outs, both commented in place:

- **View control does not count** (jump-to-bottom, `/tui fullscreen`) — it moves the view; it doesn't enter text that could collide with a steer.
- **Image paste is counted inside `attachImages`**, not via the textarea: the capture-phase image-paste interceptor on `el` calls `stopImmediatePropagation()`, so a textarea `paste` listener never sees a pasted screenshot — even though it injects a path into the prompt.

### Server seam

`src/operator-activity.ts` had the same defect: `isOperatorKeystroke()` filtered only `\x00resize:` frames, so mouse reports stamped `lastOperatorKeystrokeAt` as "the operator typed". It has no DOM, so it filters by bytes (SGR/urxvt mouse, focus, DA, DSR, DECRPM, XTWINOPS, OSC, DCS), anchored through each terminator so a reply can't swallow real bytes batched after it. One limitation is recorded in the file: in an alt-buffer TUI that hasn't negotiated wheel reporting (`vim`, `less`), xterm turns a wheel scroll into `\x1b[A`/`\x1b[B` — byte-identical to a typed arrow — so it still stamps there. The seam has no read consumer, and those are the bytes the PTY genuinely receives.

## Why the tests are worth trusting

`ui/src/lib/terminal-input.browser.test.ts` mounts a **real xterm `Terminal`** in Chromium and drives **real** wheel / mousedown / mousemove / drag events, parameterized over `?1000` / `?1002` / `?1003` (`agentOwnsScroll` only proves `mouseTrackingMode !== "none"`, not *which* mode — and under `?1000` a bare `mousemove` reports nothing, so a test pinned to it would never exercise the gesture in the bug report).

Each pointer case asserts **both halves**: that `onData` genuinely fired — i.e. the frame the old code would have counted was really emitted — **and** that the counter stayed at 0. Without the first assertion the test could pass against a broken mechanism. Real operator input (keystroke, paste incl. the middle-click shape, IME `compositionend`, dropped text) is asserted to still count, and `ReviewInFlightBanner.browser.test.ts` gains the escalation case it never had.

## Verification

- Root: `bun run lint` ✅, `bun test` ✅ (6968 pass)
- UI: `bun run check` ✅ (0 errors), `bun test` ✅ (3789 pass, node + browser)

[no-feature-entry] — bug fix, no new user-facing capability; banner copy unchanged, so no new i18n keys.

### DIFF
--- a/src/operator-activity.ts
+++ b/src/operator-activity.ts
@@ -12,11 +12,48 @@
  *  URL-bar toggles. It is NOT operator typing, so it must not stamp the seam. */
 const RESIZE_PREFIX = "\x00resize:";
 
-/** True iff `frame` is genuine operator keystroke input (i.e. not a resize
- *  control frame). The only non-keystroke frame this WS carries is the resize
- *  frame above; everything else is real input. */
+/**
+ * Frames this WS carries that nobody typed. The browser forwards everything xterm
+ * emits, and with mouse tracking on (Claude Code runs that way) that includes a
+ * report for every mouse move / click / drag / wheel, plus focus reports and the
+ * terminal's replies to the app's own queries. Treating those as "the operator
+ * typed" is wrong — it is the same defect that made the review banner claim
+ * "You're typing" on mere pointer movement (issue #1022).
+ *
+ * Each alternative is anchored to its introducer and spelled out through its
+ * terminator — no `.*` — so a reply cannot swallow real bytes batched after it in
+ * the same frame. OSC accepts both terminators (BEL and ST).
+ *
+ * The client fixes the banner a better way (it counts input-origin DOM events, see
+ * ui/src/lib/terminal-input.ts), which is not available here: the server sees only
+ * bytes. One consequence is recorded and accepted — in an alt-buffer TUI that has
+ * NOT negotiated wheel reporting (vim, less), xterm translates a wheel scroll into
+ * `\x1b[A`/`\x1b[B`, byte-identical to a typed arrow key, so it still stamps here.
+ * The seam has no read consumer yet, and those are the bytes the PTY genuinely
+ * receives.
+ */
+const NON_TYPING = new RegExp(
+  [
+    "\\x1b\\[<\\d+;\\d+;\\d+[Mm]", // SGR (and SGR-pixels) mouse report
+    "\\x1b\\[\\d+;\\d+;\\d+M", // urxvt (1015) mouse report
+    "\\x1b\\[[IO]", // focus in / focus out
+    "\\x1b\\[[?>]\\d*(?:;\\d+)*c", // Device Attributes reply
+    "\\x1b\\[\\d+;\\d+R", // cursor-position (DSR) reply
+    "\\x1b\\[\\?\\d+;\\d+\\$y", // DECRPM reply
+    "\\x1b\\[\\d+;\\d+;\\d+t", // XTWINOPS report
+    "\\x1b\\][^\\x07\\x1b]*(?:\\x07|\\x1b\\\\)", // OSC reply (BEL- or ST-terminated)
+    "\\x1bP[^\\x1b]*\\x1b\\\\", // DCS reply (XTGETTCAP, DECRQSS)
+  ].join("|"),
+  "g",
+);
+
+/** True iff `frame` is genuine operator input: not a resize control frame, and not
+ *  purely terminal-generated chatter. A frame counts as typed if ANY byte survives
+ *  stripping the known non-typing sequences — whitespace and control bytes included
+ *  (a typed space, Enter or Ctrl-C is real input); only an empty residue is not. */
 export function isOperatorKeystroke(frame: string): boolean {
-  return !frame.startsWith(RESIZE_PREFIX);
+  if (frame.startsWith(RESIZE_PREFIX)) return false;
+  return frame.replace(NON_TYPING, "").length > 0;
 }
 
 /** Throttle window: at most one stamp per session per this many ms, so a fast
--- a/src/operator-activity.ts
+++ b/src/operator-activity.ts
@@ -46,7 +46,8 @@ const NON_TYPING = new RegExp(
     "\\x1b\\[M[\\s\\S]{3}", // X10/default mouse report (button + col + row as raw bytes)
     "\\x1b\\[[IO]", // focus in / focus out
     "\\x1b\\[[?>]\\d*(?:;\\d+)*c", // Device Attributes reply
-    "\\x1b\\[\\??\\d+;\\d+R", // cursor-position reply: DSR and DECXCPR (`?` form)
+    "\\x1b\\[\\??\\d+;\\d+R", // cursor-position reply: DSR-6 and DECXCPR (`?` form)
+    "\\x1b\\[\\d*n", // DSR-5 status reply (`\x1b[0n`) — same deviceStatus handler as DSR-6
     "\\x1b\\[\\?\\d+;\\d+\\$y", // DECRPM reply
     "\\x1b\\[\\d+;\\d+;\\d+t", // XTWINOPS report
     "\\x1b\\][^\\x07\\x1b]*(?:\\x07|\\x1b\\\\)", // OSC reply (BEL- or ST-terminated)
--- a/src/operator-activity.ts
+++ b/src/operator-activity.ts
@@ -24,6 +24,13 @@ const RESIZE_PREFIX = "\x00resize:";
  * terminator — no `.*` — so a reply cannot swallow real bytes batched after it in
  * the same frame. OSC accepts both terminators (BEL and ST).
  *
+ * All three mouse encodings are covered, including X10/default (`\x1b[M` + 3 raw
+ * bytes), which an app gets by enabling ?1000/?1002/?1003 without negotiating an
+ * extended encoding (vim with ttymouse=xterm). Today that frame cannot actually
+ * reach us — xterm emits default-encoding reports on `onBinary`, which the browser
+ * never subscribes to, so they are dropped client-side — but the predicate should
+ * be complete in the encoding, not in what one client currently forwards.
+ *
  * The client fixes the banner a better way (it counts input-origin DOM events, see
  * ui/src/lib/terminal-input.ts), which is not available here: the server sees only
  * bytes. One consequence is recorded and accepted — in an alt-buffer TUI that has
@@ -36,9 +43,10 @@ const NON_TYPING = new RegExp(
   [
     "\\x1b\\[<\\d+;\\d+;\\d+[Mm]", // SGR (and SGR-pixels) mouse report
     "\\x1b\\[\\d+;\\d+;\\d+M", // urxvt (1015) mouse report
+    "\\x1b\\[M[\\s\\S]{3}", // X10/default mouse report (button + col + row as raw bytes)
     "\\x1b\\[[IO]", // focus in / focus out
     "\\x1b\\[[?>]\\d*(?:;\\d+)*c", // Device Attributes reply
-    "\\x1b\\[\\d+;\\d+R", // cursor-position (DSR) reply
+    "\\x1b\\[\\??\\d+;\\d+R", // cursor-position reply: DSR and DECXCPR (`?` form)
     "\\x1b\\[\\?\\d+;\\d+\\$y", // DECRPM reply
     "\\x1b\\[\\d+;\\d+;\\d+t", // XTWINOPS report
     "\\x1b\\][^\\x07\\x1b]*(?:\\x07|\\x1b\\\\)", // OSC reply (BEL- or ST-terminated)
--- a/test/operator-activity.test.ts
+++ b/test/operator-activity.test.ts
@@ -34,6 +34,7 @@ test("focus reports and terminal query replies are NOT operator input", () => {
   expect(isOperatorKeystroke("\x1b[>0;10;1c")).toBe(false); // DA2
   expect(isOperatorKeystroke("\x1b[12;3R")).toBe(false); // cursor position (DSR)
   expect(isOperatorKeystroke("\x1b[?12;3R")).toBe(false); // cursor position (DECXCPR)
+  expect(isOperatorKeystroke("\x1b[0n")).toBe(false); // DSR-5 status reply
   expect(isOperatorKeystroke("\x1b[?2026;2$y")).toBe(false); // DECRPM
   expect(isOperatorKeystroke("\x1b[8;40;120t")).toBe(false); // XTWINOPS
   expect(isOperatorKeystroke("\x1b]11;rgb:1e1e/1e1e/2e2e\x07")).toBe(false); // OSC, BEL-terminated
@@ -49,6 +50,7 @@ test("typed bytes batched after a reply still count as operator input", () => {
   // The non-typing patterns are anchored through their terminators (no greedy `.*`),
   // so a reply can't swallow real input that arrives in the same frame.
   expect(isOperatorKeystroke("\x1b[12;3Rx")).toBe(true);
+  expect(isOperatorKeystroke("\x1b[0nx")).toBe(true);
   expect(isOperatorKeystroke("\x1b[<35;12;5Ma")).toBe(true);
   expect(isOperatorKeystroke("\x1b]11;rgb:1e1e/1e1e/2e2e\x07hi")).toBe(true);
 });
--- a/test/operator-activity.test.ts
+++ b/test/operator-activity.test.ts
@@ -16,6 +16,50 @@ test("a resize control frame is NOT operator input", () => {
   expect(isOperatorKeystroke("\x00resize:120:40\n")).toBe(false);
 });
 
+test("a mouse report is NOT operator input", () => {
+  // With mouse tracking on, xterm forwards a report for every move/click/drag/wheel
+  // over the terminal — nobody typed those (issue #1022).
+  expect(isOperatorKeystroke("\x1b[<35;12;5M")).toBe(false); // SGR motion
+  expect(isOperatorKeystroke("\x1b[<0;12;5M")).toBe(false); // SGR press
+  expect(isOperatorKeystroke("\x1b[<0;12;5m")).toBe(false); // SGR release
+  expect(isOperatorKeystroke("\x1b[<64;12;5M")).toBe(false); // SGR wheel up
+  expect(isOperatorKeystroke("\x1b[32;12;5M")).toBe(false); // urxvt (1015)
+});
+
+test("focus reports and terminal query replies are NOT operator input", () => {
+  expect(isOperatorKeystroke("\x1b[I")).toBe(false); // focus in
+  expect(isOperatorKeystroke("\x1b[O")).toBe(false); // focus out
+  expect(isOperatorKeystroke("\x1b[?62;1;6c")).toBe(false); // DA1
+  expect(isOperatorKeystroke("\x1b[>0;10;1c")).toBe(false); // DA2
+  expect(isOperatorKeystroke("\x1b[12;3R")).toBe(false); // cursor position (DSR)
+  expect(isOperatorKeystroke("\x1b[?2026;2$y")).toBe(false); // DECRPM
+  expect(isOperatorKeystroke("\x1b[8;40;120t")).toBe(false); // XTWINOPS
+  expect(isOperatorKeystroke("\x1b]11;rgb:1e1e/1e1e/2e2e\x07")).toBe(false); // OSC, BEL-terminated
+  expect(isOperatorKeystroke("\x1b]10;rgb:c0c0/c0c0/c0c0\x1b\\")).toBe(false); // OSC, ST-terminated
+  expect(isOperatorKeystroke("\x1bP1+r5463=787465726d\x1b\\")).toBe(false); // DCS (XTGETTCAP)
+});
+
+test("a burst of mouse reports in one frame is NOT operator input", () => {
+  expect(isOperatorKeystroke("\x1b[<35;10;5M\x1b[<35;11;5M\x1b[<35;12;5M")).toBe(false);
+});
+
+test("typed bytes batched after a reply still count as operator input", () => {
+  // The non-typing patterns are anchored through their terminators (no greedy `.*`),
+  // so a reply can't swallow real input that arrives in the same frame.
+  expect(isOperatorKeystroke("\x1b[12;3Rx")).toBe(true);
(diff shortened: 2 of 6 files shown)


scoop (commented):
## Verdict

The task is satisfied. The banner no longer escalates on pointer movement: the escalation counter is now fed by input-origin DOM events on xterm's helper textarea (`ui/src/lib/terminal-input.ts:38-62`) instead of `term.onData`, and `term.onData` is explicitly reduced to a pure forwarder (`ui/src/lib/components/Viewport.svelte:1635`). The banner itself only reacts to the number rising while a review is in flight (`ui/src/lib/components/viewport/ReviewInFlightBanner.svelte:156,165`), so a count that no longer moves on mouse activity is exactly the fix.

## What I verified

- **Every PTY-write path in Viewport is accounted for.** I enumerated the send sites: `ui/src/lib/components/Viewport.svelte:287` (ComposeBar via `sendComposed`), `:1357` (image paste/drop), `:1671` (slash-command link tap), `:1752` (synthesized Shift+Enter), `:1841` (Escape), `:3024` (SteerBar `termSend`), `:3042` (ViewportTermControls `send`) — all bump. The only unbumped sends are `:1408`, `:1423` (jump-to-bottom / jump-to-top, `agentOwnsScroll` branch) and `:1495` (`/tui fullscreen` from the redraw menu), which the RULE comment at `:293-303` deliberately excludes as view control. Note `:1495` does inject text that Claude Code executes, unlike the two scroll shortcuts — I'd call the classification defensible but it is a slightly different animal from `\x1b[1;5F`.
- **The two bypass paths that need a manual bump really do bypass the textarea.** The image-paste interceptor `stopImmediatePropagation()`s only when the clipboard actually contains an image (`Viewport.svelte:1896-1903`), so plain text paste still reaches xterm's textarea and is counted by the `paste` listener at `terminal-input.ts:51`; the image branch is counted at `:1356`. The Escape handler is a window capture-phase listener that `stopImmediatePropagation()`s (`Viewport.svelte:1835-1841`), so the textarea keydown listener never fires — the explicit bump there is required, and correct.
- **Lifecycle.** `createTypingCounter` is created after `term.open(el)` (`Viewport.svelte:1577-1580`), which is when `term.textarea` exists (the function no-ops if it doesn't, `terminal-input.ts:47`), and it is disposed in the same effect teardown (`Viewport.svelte:2132`).
- **The seam's caller is unchanged and still gates on `isOperatorKeystroke`** (`src/server.ts:7528-7529`), so the widened predicate is live on every inbound frame today, even though the read side has no consumer (`src/operator-activity.ts:2-4`, `:74-77`).
- **No stale contracts.** `keystrokes` is the only prop touched on the banner and the test helper already supplies it (`ui/src/lib/components/viewport/ReviewInFlightBanner.browser.test.ts:28`); `Viewport.svelte:2923` still passes `opKeystrokes`. No user-facing strings or message keys are added by this diff, so there is no locale-parity surface to check.
- **Regex mechanics.** `NON_TYPING` is used only with `String.replace` (`src/operator-activity.ts:56`), so the `g` flag's `lastIndex` statefulness is not a hazard, and every alternative is terminator-anchored with no `.*`, so a batched keystroke after a reply survives — matching the tests at `test/operator-activity.test.ts:49-51`.

## Gap worth closing (see findings)

`NON_TYPING` (`src/operator-activity.ts:37-38`) covers the SGR (`?1006`) and urxvt (`?1015`) mouse encodings but not the legacy/default `CSI M Cb Cx Cy` encoding, which is what a terminal emits when an app enables `?1000`/`?1002`/`?1003` *without* negotiating an extended encoding (classic `vim` with `ttymouse=xterm`, older TUIs). In that mode every click/drag/move still stamps the seam — the exact defect this PR is closing, one encoding over. This is stamping-only (no read consumer yet, `src/operator-activity.ts:4`), so it is not blocking, but it is a hole in the predicate's own stated intent.

Limitation, stated rather than asserted: `node_modules` is not installed in this checkout, so I could not open xterm.js's `CoreMouseService` to confirm which encoding it falls back to when `?1006` is absent. The finding rests on the terminal protocol (X10 `CSI M` is the legacy default) and on the fact that the diff itself chose to enumerate encodings; I did not verify xterm.js's source. Related and smaller: the DSR pattern at `:41` matches `\x1b[<r>;<c>R` but not the DECXCPR form `\x1b[?<r>;<c>R`.

- A single keystroke bumps twice (`keydown` and `input` both fire on the textarea, `terminal-input.ts:52-53`), and Shift+Enter bumps twice as well (the explicit bump at `Viewport.svelte:1752` plus the textarea keydown). Harmless today — the banner only compares `keystrokes > snapshotKeystrokes` (`ReviewInFlightBanner.svelte:165`) — but the name `opKeystrokes` reads like a keystroke count, and any future consumer that treats it as one will be off by roughly 2x.

- `ui/src/lib/components/Viewport.svelte` registers no `term.onBinary` handler, so binary-mode input is never forwarded to the PTY at all (unrelated to this change, and unchanged by it).

scoop:
Both points addressed in 0bfd4b07 — with one nuance on the X10 finding worth recording.

**Point 2 (keydown counts keys that produce no input) — valid, fixed.** Replaced the raw `keydown` listener with `term.onKey`, which fires only when a key actually resolves to input for the PTY. The textarea `input`/`compositionend`/`paste`/`drop` listeners stay, since those paths never reach `onKey` (mobile IME composition, paste, drop) — that hybrid is why `onKey` alone was rejected earlier, and it is the right keyboard half. Pinned in the real-xterm browser test, which asserts against a live `Terminal` (not my assumptions) that `Ctrl+Shift+C`, `Cmd+C`, CapsLock and NumLock emit **nothing** to the PTY and do not count, while `Ctrl+C` (SIGINT), Enter, arrows and F-keys do.

**Point 1 (X10 encoding) — pattern added, but the stated failure mode cannot occur today.** `\x1b[M`+3-byte reports never reach the server: xterm emits default-encoding mouse reports via `triggerBinaryEvent` → `onBinary`, and there is **no `onBinary` subscriber anywhere in the client** (`grep -rn "onBinary" ui/src/ src/` → nothing), so in the `vim`/`ttymouse=xterm` case those bytes are dropped in the browser and the seam is never stamped. So it is not a live false positive. I added the alternative anyway — it is one line, cannot match keyboard-producible input, and the predicate should be complete in the encoding rather than in what one client happens to forward. Flagging it only so the reasoning is on record rather than implying a bug was fixed that was not reachable.

**DECXCPR** (`\x1b[?<r>;<c>R`) was a genuine gap in the DSR pattern — added and tested.

scoop (commented):
## Verdict

The PR does what the task asks: pointer activity no longer escalates the review banner. Both prior-round findings are genuinely addressed in the diff (not just claimed). One non-blocking gap remains in the server-side predicate, of exactly the same class as the DECXCPR gap the author just fixed.

## Prior findings — verified against the diff

**1. X10/default mouse encoding + DECXCPR (src/operator-activity.ts) — ADDRESSED.**
- X10 alternative present at `src/operator-activity.ts:46` (`\x1b\[M[\s\S]{3}`), exercised at `test/operator-activity.test.ts:27`.
- DECXCPR folded into the cursor-report pattern at `src/operator-activity.ts:49` (`\x1b\[\??\d+;\d+R`), exercised at `test/operator-activity.test.ts:36`.
- The author's note that X10 is unreachable today checks out: I grepped the whole tree for `onBinary` and the only hit is the comment itself (`src/operator-activity.ts:30`) — there is no subscriber in `ui/src/` or `src/`, so default-encoding reports are dropped in the browser. Defensive completeness, correctly labelled as such.

**2. keydown counted keys that produce no PTY input (ui/src/lib/terminal-input.ts) — ADDRESSED.**
- The raw `keydown` listener is gone; the keyboard signal is `term.onKey` (`ui/src/lib/terminal-input.ts:50`), which only fires for keys that resolve to input. The textarea `input`/`compositionend`/`paste`/`drop` listeners (`:56`–`:59`) cover the paths `onKey` never sees (IME, paste, drop).
- The exact case I raised is pinned against a live `Terminal` in a real browser, asserting both halves — `ui/src/lib/terminal-input.browser.test.ts:193`–`:204` presses CapsLock, NumLock, `Ctrl+Shift+C` and `Cmd+C` and asserts `frames.join("") === ""` (nothing reached the PTY) **and** `count === 0`. Copying output mid-review no longer escalates.

## What else I verified

- **Every PTY-input path bumps exactly where it should.** I enumerated all send sites in `Viewport.svelte`: bumped at `:286` (ComposeBar), `:1356` (image paste), `:1671` (slash-link tap), `:1751` (Shift+Enter), `:1840` (Escape rescue), `:3023` (SteerBar `termSend`), `:3041` (touch key bar `send`). Not bumped: `:1408`/`:1423` (jump-to-bottom / jump-to-top) and `:1495` (`/tui fullscreen`), which the RULE comment at `:293`–`:301` documents as view control. `term.onData` no longer bumps (`:1635`) — that is the fix.
- **The touch-bar bump is correct, not a new false positive.** `ViewportTermControls` passes `send` into `ControlBar` (`ui/src/lib/components/viewport/ViewportTermControls.svelte:131,143,153`), and `ControlBar.svelte:64` fires `onkey(k.seq)` only from a deliberate tap on a `nav`/`edit`/`signal`/`cancel` key — all real byte sequences into the PTY.
- **Double-counting is harmless, as claimed.** The consumer only asks whether the count rose: `ui/src/lib/components/viewport/ReviewInFlightBanner.svelte:165` is `if (nowInFlight && keystrokes > snapshotKeystrokes) escalated = true`, against the snapshot taken at `:156`. A key that fires both `onKey` and `input` can't misbehave.
- **`NON_TYPING` is a `g`-flagged module-level regex (`src/operator-activity.ts:55`) used only via `String.prototype.replace` (`:64`)** — `replace` resets `lastIndex`, so there is no cross-call statefulness bug (which `.test()` would have had). Alternations are anchored through their terminators with no nested quantifiers, so no ReDoS.
- **Server call site unchanged in shape:** `src/server.ts:7528` still gates `stampOperatorKeystroke` on `isOperatorKeystroke(frame)`; the signature (`string -> boolean`) is unchanged and this is its only production caller.
- **Observation, not a defect:** `redrawFullscreen` (`Viewport.svelte:1495`) sends `/tui fullscreen` plus Enter into the prompt without bumping. That is a real submission to the agent, not purely a view move — but it is an explicit, documented product call (the RULE comment at `:300`), and a false *negative* here is the safe direction for this banner. Likewise `scrollToBottom`/`scrollToTop` (`:1408`, `:1423`) still stamp the *server* seam, because the server sees only bytes and cannot distinguish a hand-typed Ctrl+End from the jump button — unfixable server-side and disclosed in the module comment.

## Stated limitations

- `node_modules` is not installed in this checkout, so I could not open `@xterm/xterm` v6 (`ui/package.json:49`) to confirm from source when `onKey` fires, or how the helper textarea routes `paste`/`drop`. I did not assume: the diff's new `ui/src/lib/terminal-input.browser.test.ts` asserts these against a live `Terminal` in a real browser, which is the right way to establish them, and it checks the emitted `onData` frames rather than the author's model of xterm.

- src/operator-activity.ts:46 — the X10 alternative matches three UTF-16 code units, but `src/server.ts:7525` decodes the frame with `msg.toString()` (UTF-8); raw X10 button/col/row bytes ≥ 0x80 (column or row > 95) do not survive that decode one-to-one, so `[\s\S]{3}` would not reliably match a real X10 report. Dormant today precisely because nothing forwards X10 (no `onBinary` subscriber), but if that ever changes the pattern needs a binary-safe frame, not just the regex.

scoop (commented):
## Verdict

The fix does what the task asks: the banner's escalation counter no longer rides on `term.onData` (which carries mouse/focus/query-reply frames), and the server-side seam no longer stamps on terminal-generated chatter. I could not find a false-positive path that survives. One non-blocking nit below.

## Previous revision's point

1. **DSR-5 status reply — ADDRESSED.** `src/operator-activity.ts:49` now carries the `\x1b\[\d*n` alternative alongside the DSR-6/DECXCPR form at `src/operator-activity.ts:48`, and it is pinned by `test/operator-activity.test.ts:37` (`\x1b[0n` → false) and by the batching case at `test/operator-activity.test.ts:53` (`\x1b[0nx` → true, so the alternative is anchored and cannot swallow real bytes after it).

## What I verified

**Server predicate.** The frame reaching `isOperatorKeystroke` is a `string` (`src/server.ts:7524` coerces with `msg.toString()` before the gate at `src/server.ts:7528`), so `frame.replace(...)` at `src/operator-activity.ts:65` is safe. Every alternative in `NON_TYPING` (`src/operator-activity.ts:41-52`) is anchored to its introducer and terminated explicitly — no `.*` — so a residue test (`.length > 0`) cannot be defeated by real bytes batched into the same frame; the regex has no nested quantifier, so no ReDoS. The module-level `/g` regex is only used via `String.replace`, which resets `lastIndex`, so there is no cross-call state bug.

**Client counter — no operator path lost.** Every branch of the custom key handler that returns `false` (so xterm's `onKey` never fires) and still writes bytes to the PTY bumps: Shift+Enter at `ui/src/lib/components/Viewport.svelte:1751`. The branches that return `false` and write *nothing* (Alt session combos `:1770`, Cmd/Ctrl+K `:1783`, Ctrl+Shift+C copy-with-selection `:1791`) correctly do not bump. The window-capture Escape handler `stopImmediatePropagation()`s so xterm's textarea never sees the keydown, and bumps explicitly at `:1840`. The non-xterm injection paths all bump: ComposeBar via `sendComposed` `:286`, image paste/drop via `attachImages` `:1356`, slash-command taps `:1670`, SteerBar `termSend` `:3023`, touch key bar `send` `:3041`. Lifecycle is symmetric — created after `term.open()` at `:1582`, disposed at `:2129` before `term.dispose()`.

**Client counter — no noise path gained.** `createTypingCounter` (`ui/src/lib/terminal-input.ts:52-70`) listens only on `term.onKey` and the helper textarea's `input`/`compositionend`/`paste`/`drop`. None of the mouse/focus/query-reply frames xterm synthesizes internally touch the textarea or `onKey`, so pointer motion cannot move the count — which is the defect in the task. The double-count case (a key that fires both `onKey` and `input`) is harmless: the consumer only compares the number against a snapshot (`ui/src/lib/components/viewport/ReviewInFlightBanner.svelte:156` and `:165`), it does not use the magnitude.

**Consumer contract.** `keystrokes: number` at `ReviewInFlightBanner.svelte:36` is fed from `opKeystrokes` at `Viewport.svelte:2923`; no other consumer of the prop exists. Server seam has exactly one writer (`src/server.ts:7528`) and no reader yet, consistent with the comment at `src/server.ts:7363`.

## Stated limitations (could not verify — not a finding)

- `node_modules` is absent from this checkout, so I could not open xterm 6's source (`ui/package.json:49` pins `@xterm/xterm ^6.0.0`) to confirm (a) that `onKey` fires only for keys that resolve to PTY input, (b) that X10-encoded mouse reports go out on `onBinary` rather than `onData` as the comment at `src/operator-activity.ts:26-30` claims, or (c) that the helper textarea can actually receive a `drop` event — xterm parks it under `.xterm-screen` at `z-index: -5`, which would mean a text drop lands on the screen element and the `drop` listener at `ui/src/lib/terminal-input.ts:64` never fires in a real browser. In each case the consequence is at worst a no-op, never a false "You're typing", so nothing here blocks. The `drop`/`paste`/`compositionend` assertions in `ui/src/lib/terminal-input.browser.test.ts:207-224` dispatch synthetically on `term.textarea`, so they prove the listener wiring, not that a real gesture reaches that element — worth knowing when reading those tests as evidence.

- `src/operator-activity.ts`: Shepherd's own view-control injections (`\x1b[1;5F` at `Viewport.svelte:1408`, `\x1b[1;5H` + 500×PageUp at `:1423`, and the `/tui fullscreen` paste at `:1495`) are plain keystroke bytes on the same WS, so the server stamps them as operator typing — the same class of "nobody typed this" false positive the predicate is meant to remove. Inert today (the seam has no reader, per `src/server.ts:7363`); it becomes a real wrong answer the moment the foreshadowed stage-and-apply guard reads it.

scoop (commented):
## Re-review

**Prior finding 1 (Viewport.svelte, redrawFullscreen vs. the RULE carve-out): addressed.** `redrawFullscreen()` now calls `bumpOperatorInput()` before `conn?.send(composeKeystrokes("/tui fullscreen"))` (ui/src/lib/components/Viewport.svelte:1499-1500), and the RULE comment at :295-302 was rewritten: the enumerated bumping paths now explicitly include "the /tui fullscreen redraw (a bracketed paste + CR, i.e. a submitted prompt line)", and the carve-out is narrowed to the jump-to-top/bottom scroll sends with a rationale that the code does support — those two sends are `\x1b[1;5F` (Ctrl+End, Viewport.svelte:1410) and `\x1b[1;5H` + PageUp spam (Viewport.svelte:1425), which are navigation keys, not prompt text. The rationale no longer rests on a claim the code contradicts.

## Verification of the change

**Every PTY send path is accounted for.** I enumerated all `conn.send` / `c.send` call sites in the changed file (Viewport.svelte:287, 1359, 1410, 1425, 1500, 1640, 1676, 1757, 1846, 3029, 3047) and checked each against the RULE:
- bumped: `sendComposed` (:286, the only consumer is `ComposeBar onsend`, Viewport.svelte:3061), image paste/drop (:1358), slash-link tap (:1675), Shift+Enter (:1756), window-capture Escape (:1845), `SteerBar termSend` (:3028) and `ViewportTermControls send` (:3046).
- deliberately not bumped: `term.onData` (:1640) and the two scroll sends (:1410, :1425), per the documented carve-out.

**The two prop-based send paths are only ever operator-initiated**, so bumping in the prop callback cannot produce a false escalation: `SteerBar`'s `termSend` reaches the PTY only via `ControlBar onkey` (ui/src/lib/components/SteerBar.svelte:182), and `ViewportTermControls`' `send` only via `ControlBar onkey` / the notes and enter buttons (ui/src/lib/components/viewport/ViewportTermControls.svelte:131,143,153). `ControlBar` fires `onkey` exclusively from its tap handler, guarded by pointer-id + slop (ui/src/lib/components/ControlBar.svelte:60-65) — no programmatic caller.

**The manual bumps at :1358 and :1845 do not double-count**, and are genuinely required: the image-paste interceptor calls `e.stopImmediatePropagation()` in the capture phase on `el` (Viewport.svelte:1901-1908), so xterm's textarea never sees that paste; the Escape handler does the same in the window capture phase (Viewport.svelte:1841-1849), so xterm's textarea never sees that keydown. A plain text paste matches no image item and falls through (Viewport.svelte:1903), where `createTypingCounter`'s textarea `paste` listener picks it up (ui/src/lib/terminal-input.ts:59).

**No suppressing branch of the custom key handler sends bytes without bumping.** `attachCustomKeyEventHandler` (Viewport.svelte:1746-1806) returns `false` for Shift+Enter (bumped, :1756), Alt session-switch combos, Cmd/Ctrl+K and Ctrl+Shift+C — the latter three `preventDefault()` and send nothing, so there is nothing to count.

**Banner contract holds.** `createTypingCounter` is wired at Viewport.svelte:1587 (after `term.open(el)` at :1582, which is what `term.textarea` requires, as the module documents at ui/src/lib/terminal-input.ts:44-45) and disposed at :2137. `opKeystrokes` is passed as `keystrokes={opKeystrokes}` (Viewport.svelte:2928), and the banner's sticky rule is `if (nowInFlight && keystrokes > snapshotKeystrokes) escalated = true` against a snapshot taken on entry (ui/src/lib/components/viewport/ReviewInFlightBanner.svelte:156,165) — which is exactly what the new test asserts (count held → calm; count risen → escalated and sticky).

**i18n parity:** the message the new test asserts on, `reviewbanner_escalated`, exists in both catalogs — ui/messages/en.json:2006 and ui/messages/de.json:2006. This PR adds no new keys.

**Server predicate (src/operator-activity.ts).** `isOperatorKeystroke` is called on one frame per WS message (src/server.ts:7525-7529), and the module-level `NON_TYPING` regex is only used with `String.replace`, so its `g` flag carries no `lastIndex` state between calls. I checked the strip patterns against the sequences that actually still have to survive: a bracketed paste (`\x1b[200~…\x1b[201~`) matches none of the alternatives (`\x1b\[\d*n` needs a trailing `n`, the mouse alternatives need `<` or three numeric params), and neither do CSI-u / modifyOtherKeys key encodings (`\x1b[97;6u`) or SGR colour bytes inside pasted output (`\x1b[0m`) — all still return `true`. Each alternative is anchored through its terminator, so bytes batched after a reply survive, matching the tests at test/operator-activity.test.ts:52-55.

### Stated limitations
- xterm.js is not installed in this checkout (no `node_modules/@xterm/*`), so I could not read its source to confirm that `term.onKey` fires only for keys that resolve to PTY input and that `attachCustomKeyEventHandler` returning `false` suppresses `onKey`. That behaviour is what ui/src/lib/terminal-input.ts:32-38 relies on and what ui/src/lib/terminal-input.browser.test.ts:195-206 pins against a real Terminal in a real browser; I assumed it as written rather than asserting it from the library source.

PR #1766, opened 2026-07-14
fix(ui): stop the reviewing amber border leaking into FINAL REVIEW

scoop:
## The question this answers

> "FINAL REVIEW" is slate/grey in an orange border. Is that on purpose?

**Half of it was.** Two independent things produced that pill, and only one is a bug:

- **The grey text was intentional.** `.critic-badge.streak-final { color: var(--color-faint) }` made the final round deliberately *recessive*, and the comment there said so.
- **The orange border was not.** `.critic-reviewing` (specificity 0,1,0) sets `border-color: var(--color-amber)` for the "critic is running" state. `.streak-final` overrode the *color* but never the *border* — so the amber leaked through underneath the recessive text. Nothing chose that combination; it fell out of two rules meeting.

So this isn't one bug presented as a ladder-wide redesign. It's a cascade leak, plus a ladder that the leak exposed as non-monotonic.

## The change

Each streak state now owns **text + border** together, which closes the leak structurally — no state can show a border hue that disagrees with its text hue. And the ladder now climbs with severity, where before `final` (faint) was *less* salient than `round` (amber) despite being the last-chance state:

| State | Label | Text + border | Change |
|---|---|---|---|
| `round` | `REVIEW 2/5` | `--color-amber` (in progress) | **none** |
| `final` | `FINAL REVIEW` | `--status-warn` text **and** border | recolored (was faint grey) |
| `stalled` | `STALLED REVIEW` | `--status-blocked` text **and** border, weight 600 | recolored (was amber) |

Two axes, deliberately orthogonal: **hue = severity, the pulsing dot = running.** An amber dot inside a warn or red pill is correct — they report different things.

### Two things left alone on purpose

- **`round` keeps its amber.** `DESIGN.md:219` assigns amber to "in progress" and `--status-running` is amber. (Muting it would also have made `.streak-round` a no-op — the base `.critic-badge` rule is *already* muted text on a `--color-line` border.)
- **`.rev-dot` stays amber; `currentColor` was rejected.** It would render a **pulsing red dot** when `stalled` + `reviewing` co-occur — reachable in practice, since `addressStallStatus` escalates on the `finalRoundTimeoutMs` clock while a re-review is still in flight, and `roundView` still sets `dot: true`. `DESIGN.md:220`/`:263` forbid a pulsing subordinate red; that loudest red is reserved for the blocked-agent pip (Four-Light Rule). There's now a test pinning this.

Also fixes the `addressRoundInfo` doc comment in `critic-badge.ts`, which claimed `round` = blue and `stalled` = orange — a scheme the CSS never implemented.

## Tests assert computed style, not classes

The bug **shipped with the correct `.streak-final` class applied**, so a class assertion would be regression-proof against nothing. The new tests read `getComputedStyle().color` / `.borderTopColor` and assert they agree per state — **including with `.critic-reviewing` applied**, which is the exact cascade that leaked. Expected values are resolved from the tokens via a probe element rather than hardcoded hex, so an `app.css` palette retune can't fail them for the wrong reason.

Verified they're real guards: reverting the CSS to the old rule fails the two FINAL tests and passes the rest.

## Contrast: measured, and NOT claimed as an AA fix

Badge text is `--fs-micro`, so AA = 4.5:1. Measured across all four palettes:

| Palette | `--warn` on bg/panel | `--red` on bg/panel | `--amber` (existing `round`) |
|---|---|---|---|
| dark | 6.40 / 6.10 | 4.99 / 4.75 | 8.91 / 8.49 |
| light | **4.06** / 4.62 | 4.49 / 5.11 | **3.75** / 4.27 |
| dark-hc | 9.28 / 8.83 | 7.75 / 7.38 | 12.23 / 11.65 |
| light-hc | 6.25 / 7.11 | 6.57 / 7.47 | 7.08 / 6.22 |

Light `--warn` on `--bg` is **4.06:1** — under AA — so **no AA claim is made here**. Worth noting the same measurement puts light `--amber` (the *existing, untouched* `round` color) at **3.75:1**: sub-AA badge text is a **pre-existing, palette-level** property of the light theme, not something this PR introduces, and fixing it means retuning `app.css` for every consumer. Worth a follow-up; out of scope here.

What this PR does claim: it **strictly improves** `final`'s contrast in all four palettes (today's `--color-faint` is ~2.1:1 light / ~2.4:1 dark).

## Checks

`bun run lint` · `cd ui && bun run check && bun run check:i18n && bun run test && bun run build` — all green (3772 tests). Full 7-lane pre-push passed. No new i18n keys (labels/tooltips unchanged); no feature-catalog entry (a `fix`, not a `feat`).

### DIFF
--- a/ui/src/lib/components/CriticBadge.browser.test.ts
+++ b/ui/src/lib/components/CriticBadge.browser.test.ts
@@ -115,7 +115,7 @@ describe("CriticBadge streak label", () => {
 
   it("verdict + final round → FINAL REVIEW streak label with streak-final class", async () => {
     // addressRound === addressCap, findings present, finalRoundPending true, updatedAt recent
-    // → addressRoundInfo returns status "final" (dim).
+    // → addressRoundInfo returns status "final" (warn — caution, last chance).
     reviews.map = {
       s1: v({
         addressRound: 5,
@@ -134,7 +134,7 @@ describe("CriticBadge streak label", () => {
     await expect.element(page.getByText(/FINAL REVIEW/)).toBeInTheDocument();
     // verdict word replaced, not appended
     expect(page.getByText(/CHANGES/).elements().length, "no CHANGES text").toBe(0);
-    // badge carries the dim class
+    // badge carries the warn (final-rung) class — resolved colors are asserted in the ladder suite
     const streak = document.querySelector(".streak-final");
     expect(streak, "streak-final rendered").not.toBeNull();
   });
--- a/ui/src/lib/components/CriticBadge.browser.test.ts
+++ b/ui/src/lib/components/CriticBadge.browser.test.ts
@@ -140,6 +140,116 @@ describe("CriticBadge streak label", () => {
   });
 });
 
+/**
+ * The grey-in-orange FINAL REVIEW bug shipped WITH the correct `.streak-final` class applied — the
+ * class was never the defect, the resolved border-color was (`.critic-reviewing`'s amber leaking
+ * through underneath a state that only recolored its text). So these assert computed style, not
+ * classes; a class assertion here would be regression-proof against nothing.
+ *
+ * Expected values are resolved from the tokens via a probe element rather than hardcoded as hex, so
+ * a palette retune in app.css can't fail these for the wrong reason.
+ */
+describe("CriticBadge streak ladder — resolved colors", () => {
+  /** Computed rgb() a token resolves to in the live stylesheet. */
+  function token(name: string): string {
+    const probe = document.createElement("span");
+    probe.style.color = `var(${name})`;
+    document.body.appendChild(probe);
+    const c = getComputedStyle(probe).color;
+    probe.remove();
+    return c;
+  }
+
+  const badge = () => document.querySelector(".critic-badge") as HTMLElement;
+  // border is `1px solid` on all four sides, so the top side speaks for the whole box.
+  const ink = (el: HTMLElement) => {
+    const s = getComputedStyle(el);
+    return { color: s.color, border: s.borderTopColor };
+  };
+
+  const roundVerdict = (extra: Partial<ReviewVerdict> = {}) =>
+    v({ addressRound: 2, addressCap: 5, ...extra });
+  const finalVerdict = () =>
+    v({
+      addressRound: 5,
+      addressCap: 5,
+      finalRoundPending: true,
+      findings: ["x"],
+      updatedAt: Date.now() - 1_000, // recent → "final", not yet timed out into "stalled"
+    });
+  const stalledVerdict = () =>
+    v({ addressRound: 5, addressCap: 5, finalRoundPending: false, findings: ["x"] });
+
+  it("final: warn text AND warn border — the border no longer leaks amber", async () => {
+    reviews.map = { s1: finalVerdict() };
+    render(CriticBadge, { sessionId: "s1" });
+    await expect.element(page.getByText(/FINAL REVIEW/)).toBeInTheDocument();
+
+    const warn = token("--status-warn");
+    const { color, border } = ink(badge());
+    expect(color, "final text is --status-warn").toBe(warn);
+    expect(border, "final border is --status-warn, not amber").toBe(warn);
+    expect(border, "border hue agrees with text hue").toBe(color);
+  });
+
+  it("final + reviewing: still warn on both — the exact cascade that produced grey-in-orange", async () => {
+    reviews.map = { s1: finalVerdict() };
+    reviews.reviewing = { s1: true }; // adds .critic-reviewing, whose amber border leaked before
+    render(CriticBadge, { sessionId: "s1" });
+    await expect.element(page.getByText(/FINAL REVIEW/)).toBeInTheDocument();
+
+    const warn = token("--status-warn");
+    const { color, border } = ink(badge());
+    expect(color, "text stays --status-warn under .critic-reviewing").toBe(warn);
+    expect(border, "border is NOT the reviewing amber").toBe(warn);
+    expect(border).not.toBe(token("--color-amber"));
+    // and it is certainly not the old recessive faint text
+    expect(color).not.toBe(token("--color-faint"));
+  });
+
+  it("stalled: blocked-red text AND border", async () => {
+    reviews.map = { s1: stalledVerdict() };
+    render(CriticBadge, { sessionId: "s1" });
+    await expect.element(page.getByText(/STALLED REVIEW/)).toBeInTheDocument();
+
+    const blocked = token("--status-blocked");
+    const { color, border } = ink(badge());
+    expect(color, "stalled text is --status-blocked").toBe(blocked);
+    expect(border, "stalled border is --status-blocked").toBe(blocked);
+  });
+
+  it("round keeps amber (in-progress), and agrees with its border while reviewing", async () => {
+    reviews.map = { s1: roundVerdict() };
+    reviews.reviewing = { s1: true };
+    render(CriticBadge, { sessionId: "s1" });
+    await expect.element(page.getByText(/REVIEW 2\/5/)).toBeInTheDocument();
+
+    const amber = token("--color-amber");
+    const { color, border } = ink(badge());
+    expect(color, "round stays amber = in progress").toBe(amber);
+    expect(border, "reviewing border agrees with the amber text").toBe(amber);
+  });
+
+  it("stalled + reviewing: the pulsing dot stays AMBER, never red", async () => {
+    // Reachable in practice: addressStallStatus escalates to "stalled" on the finalRoundTimeoutMs
+    // clock while a re-review is still in flight, and roundView still sets dot: true. A dot that
+    // followed currentColor would pulse RED here — DESIGN.md forbids a pulsing subordinate red
+    // (that loudest red is reserved for the blocked-agent pip; Four-Light Rule).
+    reviews.map = { s1: stalledVerdict() };
+    reviews.reviewing = { s1: true };
+    render(CriticBadge, { sessionId: "s1" });
+    await expect.element(page.getByText(/STALLED REVIEW/)).toBeInTheDocument();
+
+    const dot = document.querySelector(".rev-dot") as HTMLElement;
+    expect(dot, "running dot rendered while re-reviewing a stalled streak").not.toBeNull();
+    expect(getComputedStyle(dot).backgroundColor, "dot is amber (running), not red").toBe(
+      token("--color-amber"),
+    );
+    // the pill itself is still red — hue (severity) and dot (running) are orthogonal axes
+    expect(getComputedStyle(badge()).color).toBe(token("--status-blocked"));
+  });
+});
+
 describe("CriticBadge tip mode (card) — Open PR dialog + URL safety", () => {
   const dialogOpen = () => document.querySelector(".status-tip-dialog:popover-open");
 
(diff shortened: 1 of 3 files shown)


scoop (commented):
## Verdict

The change does what the task asks. The reported "slate/grey FINAL REVIEW inside an orange border" was not intentional: `.critic-reviewing` sets `border-color: var(--color-amber)` (ui/src/lib/components/CriticBadge.svelte:251-253, specificity 0,1,0) while the old `.critic-badge.streak-final` recolored only the text to `--color-faint`, so the amber border showed through under a grey label. The new rule sets both `color` and `border-color` to `--status-warn` (ui/src/lib/components/CriticBadge.svelte:291-294) at specificity 0,2,0, which wins over `.critic-reviewing` regardless of source order. Stalled likewise moves to `--status-blocked` on both axes (ui/src/lib/components/CriticBadge.svelte:297-301).

## Verification

- Tokens resolve in every theme: `--status-warn` → `--color-warn` → `--warn` (ui/src/app.css:132, ui/src/app.css:26, ui/src/app.css:59); `--status-blocked` → `--color-red` (ui/src/app.css:130). `--warn` is defined in all four palette blocks — ui/src/app.css:59, ui/src/app.css:196, ui/src/app.css:246, ui/src/app.css:268 — so no theme falls back to an unset custom property. Both tokens are already used as text colors elsewhere (ui/src/lib/plugin-ui/PuiBadge.svelte:13, ui/src/lib/components/EpicGroupHeader.svelte:216), so this is not a novel use.
- `--status-warn` (#e8730c, ui/src/app.css:59) is genuinely distinct from `--color-amber` (#e8a13a, ui/src/app.css:54), so the new assertion `expect(border).not.toBe(token("--color-amber"))` (ui/src/lib/components/CriticBadge.browser.test.ts:200) is not vacuous.
- The probe helper `token()` (ui/src/lib/components/CriticBadge.browser.test.ts:146-154) resolves against the real palette because the suite imports the stylesheet at ui/src/lib/components/CriticBadge.browser.test.ts:4 (`import "../../app.css"`). Without that import the tests would have passed for the wrong reason; with it they compare real resolved rgb values.
- The amber `.rev-dot` is `background: var(--color-amber)` (ui/src/lib/components/CriticBadge.svelte:262), not `currentColor`, so the stalled+reviewing combination cannot render a pulsing red dot. That matches DESIGN.md:220 ("may go red on text + border + dot, but never takes a halo, pulse, or wash"), and the new test at ui/src/lib/components/CriticBadge.browser.test.ts:234-249 pins it.
- Consumers agree: `addressRoundInfo` has exactly one caller (ui/src/lib/components/CriticBadge.svelte:22) and its return shape is unchanged — the critic-badge.ts diff is doc-comment only. No other file references the `streak-*` classes (grepped the tree), and the design-system route has no critic-badge swatch to keep in sync.
- i18n parity holds for the strings involved (unchanged by this PR but relied on by the new tests): `criticbadge_final` / `criticbadge_stalled` exist in ui/messages/en.json:595 and ui/messages/en.json:593, with counterparts at ui/messages/de.json:595 and ui/messages/de.json:593.

## Nits (non-blocking)

1. The new block comment asserts "Each streak state therefore owns BOTH color and border-color" (ui/src/lib/components/CriticBadge.svelte:279), but `.critic-badge.streak-round` sets only `color` (ui/src/lib/components/CriticBadge.svelte:287-289). When a round streak is *not* being re-reviewed, `.critic-reviewing` is absent, so the pill renders amber text inside the default `--color-line` border (ui/src/lib/components/CriticBadge.svelte:217) — the same text/border hue mismatch this PR set out to eliminate, one rung lower on the ladder. That state is reachable today (agent addressing findings, critic idle). It may well be deliberate (the old comment framed the amber border as the *running* signal), but under the new "hue = severity, dot = running" model the border is a severity channel, so `streak-round` is now the odd rung. Either add `border-color: var(--color-amber)` or amend the comment so it stops claiming something the code below it doesn't do.
2. The pre-existing final-round test still describes the removed styling: "→ addressRoundInfo returns status \"final\" (dim)" (ui/src/lib/components/CriticBadge.browser.test.ts:118) and "// badge carries the dim class" (ui/src/lib/components/CriticBadge.browser.test.ts:137). Nothing is dim anymore; these read as stale now that the new suite immediately below asserts warn-orange.

## Limitations

I did not run the suite (CI covers it) and did not visually compare `--warn` (#e8730c) against `--amber` (#e8a13a) on screen — they are adjacent oranges, so the FINAL REVIEW pill's amber dot sits inside a warn-orange border. The code is internally consistent and the comment explains the intent; whether the two oranges read as distinct enough is a design call I can't verify from source.

scoop:
**Both findings addressed in 95bc2882 — but on finding 1 I took the comment remedy, not the CSS one. Reasoning, so you can weigh it:**

You're right that the comment overclaimed: it said *"each streak state owns BOTH color and border-color"* while `.streak-round` sets only `color`. That was a real inaccuracy and it's now fixed.

But I don't think the alternative remedy — giving `streak-round` a matching `border-color: var(--color-amber)` — is correct, for two reasons:

1. **It isn't the same mismatch this PR fixes.** The bug was a border hue that *contradicted* the text hue: amber border under **non-amber** (faint) text, two different colors fighting. At rest, `round` is amber text on `--color-line` — the neutral hairline that *every* badge carries (the base `.critic-badge` rule is muted text on that same hairline, as are the verdict states `.critic-commented` / `.critic-error`). A neutral hairline isn't a competing hue; there's nothing to contradict. And when `round` *is* being re-reviewed, `.critic-reviewing`'s amber border already agrees with its amber text. So `round` never displays the defect.

2. **It's an unrequested visual change to a state deliberately scoped as untouched.** Adding an amber border changes `round`'s resting appearance for every mid-streak session. The plan review specifically established that `round` stays as-is (amber = in-progress per `DESIGN.md:219`), and this PR's whole discipline was to keep the diff to the leak plus the ladder.

So the invariant I *should* have written — and now have — is narrower:

> No streak state may show a border hue that **contradicts** its text hue. Any state whose text is **not** amber must therefore claim the border; `streak-round`'s text *is* amber, so it agrees with the reviewing border and needs none.

If you think the resting `round` pill should carry an amber border anyway on cohesion grounds (`DESIGN.md`'s status-chip recipe does allow hue on border + text + dot), that's a legitimate design call — but it's a separate change to a state this PR intentionally doesn't touch, and I'd rather not smuggle it in here. Happy to file a follow-up if you want it.

Finding 2 (stale `"dim"` / faint comments in the test file) was plainly correct and is fixed.

scoop (commented):
## Re-review: both prior findings are genuinely resolved

### 1. `CriticBadge.svelte` comment vs `.streak-round` — ADDRESSED
The overclaiming sentence ("Each streak state therefore owns BOTH color and border-color") is gone. The rewritten block (`ui/src/lib/components/CriticBadge.svelte:275-291`) states a narrower invariant — *no streak state may show a border hue that CONTRADICTS its text hue* — and explicitly carves out `streak-round`: "its text IS amber, so it agrees with the reviewing border when re-reviewing, and at rest it keeps the neutral `--color-line` hairline every badge carries."

I checked that against the CSS rather than taking the note's word for it:
- `ui/src/lib/components/CriticBadge.svelte:292-294` — `.critic-badge.streak-round` sets `color: var(--color-amber)` and nothing else. The comment now says exactly that.
- `ui/src/lib/components/CriticBadge.svelte:251-253` — `.critic-reviewing` supplies `border-color: var(--color-amber)`, which agrees with round's amber text while re-reviewing.
- At rest, round falls back to the base hairline at `ui/src/lib/components/CriticBadge.svelte:217` (`border: 1px solid var(--color-line)`) — the same neutral border carried by `.critic-commented` (`:244-246`, blue text) and `.critic-error` (`:247-249`, faint text). A neutral hairline under colored text is the house pattern here, not the defect this PR fixes (which was a *competing* amber border under faint-grey text).

The original finding offered "either give `streak-round` a matching `border-color` **or** correct the comment." The comment remedy was taken and the comment is now accurate, so the finding no longer applies. Whether the resting `round` pill *should* also carry an amber border is a separate design call, not a defect.

### 2. Stale "dim"/faint comments in the test file — ADDRESSED
- `ui/src/lib/components/CriticBadge.browser.test.ts:118` now reads `status "final" (warn — caution, last chance)`.
- `ui/src/lib/components/CriticBadge.browser.test.ts:137` now reads `badge carries the warn (final-rung) class`.

No remaining "dim"/faint language in the file.

## Verification of the fix itself

**The reported bug is actually fixed.** `.critic-badge.streak-final` (`ui/src/lib/components/CriticBadge.svelte:296-299`) now sets both `color` and `border-color` to `var(--status-warn)`. Previously it set only `color: var(--color-faint)` (grey-green `#4a5752`, `ui/src/app.css:52`) while `.critic-reviewing` set `border-color: var(--color-amber)` — that is precisely the "slate/grey in an orange border" the task describes. Specificity: `.critic-badge.streak-final` is 0,2,0 vs `.critic-reviewing`'s 0,1,0 (Svelte's scope class raises both equally), so the streak border now wins regardless of source order.

**Tokens resolve and are established, not invented:**
- `--status-warn` → `--color-warn` → `--warn`, defined at `ui/src/app.css:132`, `:26`, and per-theme at `:59` (`#e8730c`), `:196`, `:246`, `:268`. Precedent for it as the caution hue: `ui/src/lib/components/UnitRow.svelte:1392-1395`, `ui/src/lib/components/herd/HerdLensStrip.svelte:272-275`, `ui/src/lib/plugin-ui/PuiBadge.svelte:13`.
- `--status-blocked` → `--color-red` at `ui/src/app.css:130`.

**The stalled red does not violate the Four-Light Rule.** `DESIGN.md:220` permits subordinate red on "text + border + dot, but never … a halo, pulse, or wash"; `.critic-badge.streak-stalled` (`ui/src/lib/components/CriticBadge.svelte:302-306`) sets only `color`, `border-color`, `font-weight`. The `.rev-dot` keeps `background: var(--color-amber)` (`ui/src/lib/components/CriticBadge.svelte:262`) and is the only animated element (`:264`), so no pulsing red is ever produced — the loudest red stays with the blocked-agent pip per `DESIGN.md:154`.

**Test fixtures match the real predicate.** The new ladder suite's `finalVerdict()` (cap reached, `finalRoundPending: true`, `updatedAt` recent) and `stalledVerdict()` (cap reached, `finalRoundPending: false`, findings present) produce `"final"` and `"stalled"` respectively under `addressStallStatus` at `ui/src/lib/review-status.ts:17-24` (mirrored identically at `src/review-status.ts:17-24`). The `stalled + reviewing` co-occurrence the comment cites is genuinely reachable: `ui/src/lib/review-status.ts:22` escalates on the `finalRoundTimeoutMs` clock independent of the `reviewing` flag set in `ui/src/lib/reviews.svelte.ts:89`.

**Doc comment in `critic-badge.ts`** (`ui/src/lib/components/critic-badge.ts:47-55`) now describes round=amber / final=warn / stalled=blocked-red, which matches the three CSS rules above; the old "(blue)" and "(dimmed)" descriptions were stale and are corrected.

**No i18n impact:** the diff introduces no new message keys (labels `criticbadge_final` / `criticbadge_stalled` / `criticbadge_round` are unchanged), so no catalog parity check applies.

**No other consumers:** grepping `streak-final|streak-stalled|streak-round` across `ui/src` returns only `CriticBadge.svelte`, its test, and `critic-badge.ts` — the styles are Svelte-scoped and nothing outside relies on them.

Nothing blocking, and no nits worth the author's time.

PR #1767, opened 2026-07-14
fix(ui): stop the stalled-review pill pulsing an orange dot

scoop:
> **Scope — two independent commits.** `f527cc27` is the badge fix (`CriticBadge` only) and is the whole point of this PR. `f491c486` is an unrelated test-only deflake of `CommandBar.browser.test.ts`, which was failing CI on this branch through no fault of the badge change (details at the bottom). They're separate commits and touch disjoint files, so the badge fix stays independently revertable and bisectable — revert `f491c486` alone and nothing here is affected.

A session whose auto-address streak had stalled rendered a blocked-red `STALLED REVIEW` pill with an amber **pulsing** dot inside it. The hue says *gave up*; the dot says *something is running*. It reads as broken.

## The dot wasn't the bug — the pill was

`forceReview` (`src/review.ts:468-505`) **persists** a streak reset (`putReview` with `addressRound: 0`) but deliberately withholds `onChange` until the new verdict lands. So on the force path the server row is no longer stalled: `quotaBlockReason` (`src/blocked.ts:148`) stops returning the `rework` block, and the `REWORK CAP` badge **and** the "Auto-fix hit its limit" hold line both clear. Only the client's cached verdict keeps painting the red pill — the last surface still claiming a stall the server had already cleared. After a force the session genuinely *is not stalled*.

So the fix isn't to hide the dot; it's to stop rendering a stalled pill over a live re-review.

## What changed

A stalled streak now yields to the normal amber `REVIEWING` pill while the critic is actually re-reviewing — which is where the pulsing dot belongs. When the verdict lands, the badge paints whatever it says, `STALLED REVIEW` (red, static, no dot) included.

| Streak rung | Re-reviewing | Before | After |
| --- | --- | --- | --- |
| `stalled` | yes | red `STALLED REVIEW` + amber pulsing dot | amber `REVIEWING` + pulsing dot |
| `stalled` | no | red `STALLED REVIEW`, no dot | unchanged |
| `round` / `final` | yes | counter + amber dot | unchanged |

`round` / `final` keep their counter and dot: mid-streak that number isn't stale, and it's the useful thing to show.

This is the rule the plan gate already applies — `plan-gate-badge.ts:49` (`if (reviewing) return { kind: "reviewing" }` ahead of the stall branches), `:74`/`:88-94` (`canShowPlanStallActions` requires `!reviewing`), `PlanGateBadge.svelte:204` (dot only in the reviewing branch) — and the one `criticChip` already states in prose for plain verdicts. `CriticBadge` was the outlier.

**Rejected alternative:** `dot: reviewing && r.status !== "stalled"` (keep the red pill, drop the dot). One line, satisfies the literal complaint, but leaves the badge painting an already-reset stale `5/5` verdict for the whole forced re-review.

## Deliberate flicker on the timeout path

`addressStallStatus` can also escalate to `stalled` on the `finalRoundTimeoutMs` clock *while* a re-review is in flight. There the pill visibly flips `STALLED REVIEW` → `REVIEWING` → `STALLED REVIEW`. **Intended:** the timeout is a guess (it fires precisely because we don't know whether the agent abandoned the round); a live re-review is fact and supersedes it. On that path the server row *is* still stalled, so `REWORK CAP` + the hold line stay on the row and carry the stall meanwhile.

## Tests

`CriticBadge.browser.test.ts:233` pinned the old behavior (STALLED text + red pill + dot) for exactly this state, so it was **rewritten**, not supplemented — it now asserts the reviewing pill wins: one badge, reads `REVIEWING`, no `.streak-stalled`, dot present and amber, pill amber and **not** `--status-blocked`. Verified it fails against the pre-change component (only that test) and passes after. The at-rest `"stalled: blocked-red text AND border"` test is unchanged and is now its pair.

`.rev-dot` inside a stalled pill is now unreachable; `.streak-stalled`'s border/color stays (it's the at-rest state) and DESIGN.md's "no pulsing subordinate red" remains the standing reason the dot isn't simply recolored to the pill's hue.

Local: `bun run check` (0 errors), `bun run test` (3800 passed), `check:i18n` (parity), root `bun run lint` clean. No new i18n keys (`criticbadge_reviewing` exists); UI-only `fix`, no new user-facing capability, so no feature-catalog entry.

---

## Second commit: deflaking `CommandBar.browser.test.ts` (unrelated, but it was red)

CI failed on `CommandBar — drops a confirm that lands inside the 300ms dwell`, which this PR's diff doesn't touch. It's a genuine pre-existing flake from #1751, not a rerun-away blip: the guard reads wall time (`Date.now() - armedAt < ARM_DWELL_MS`, `CommandBar.svelte:442`) and the test fired two real clicks hoping *both* land within 300ms. A contended runner dispatches the second one later than that, the dwell has legitimately expired, and the confirm fires — the test raced the machine, not the guard.

Fixed by pinning `Date.now()` across the click pair, then advancing it past the dwell for the deliberate confirm. Verified it still **fails** when the guard is disabled, so it stays load-bearing rather than becoming vacuous. The other `pastDwell()` tests wait 320ms to get *past* the window, so they fail safe and stay on the real clock.

### DIFF
--- a/ui/src/lib/components/CommandBar.browser.test.ts
+++ b/ui/src/lib/components/CommandBar.browser.test.ts
@@ -845,14 +845,27 @@ describe("CommandBar — destructive command two-step arm", () => {
     renderBar({ commands });
     await page.getByRole("combobox").fill("decom");
 
-    await decomRow().click();
-    await decomRow().click(); // immediate second click — inside the dwell
-    expect(run).not.toHaveBeenCalled();
-    // Still armed, so a deliberate confirm afterwards works.
-    await expect.element(decomRow()).toHaveTextContent(CONFIRM);
-    await pastDwell();
-    await decomRow().click();
-    expect(run).toHaveBeenCalledTimes(1);
+    // The guard is `Date.now() - armedAt < ARM_DWELL_MS` (CommandBar.svelte), so freeze the clock
+    // rather than racing it: a back-to-back click PAIR is only "inside the dwell" if the runner
+    // dispatches both within 300ms of wall time, which a contended CI runner does not — the second
+    // click then lands legitimately past the dwell and confirms. Pinning Date.now() asserts the
+    // guard's logic instead of the machine's speed. The other dwell tests wait 320ms to get PAST
+    // the window, so they fail safe and stay on the real clock.
+    const t0 = Date.now();
+    const now = vi.spyOn(Date, "now").mockReturnValue(t0);
+    try {
+      await decomRow().click();
+      await decomRow().click(); // second click, still at t0 — inside the dwell
+      expect(run).not.toHaveBeenCalled();
+      // Still armed, so a deliberate confirm afterwards works.
+      await expect.element(decomRow()).toHaveTextContent(CONFIRM);
+
+      now.mockReturnValue(t0 + 320); // past the dwell
+      await decomRow().click();
+      expect(run).toHaveBeenCalledTimes(1);
+    } finally {
+      now.mockRestore();
+    }
   });
 
   it("disarms on typing and on moving the cursor", async () => {
--- a/ui/src/lib/components/CriticBadge.browser.test.ts
+++ b/ui/src/lib/components/CriticBadge.browser.test.ts
@@ -230,23 +230,37 @@ describe("CriticBadge streak ladder — resolved colors", () => {
     expect(border, "reviewing border agrees with the amber text").toBe(amber);
   });
 
-  it("stalled + reviewing: the pulsing dot stays AMBER, never red", async () => {
-    // Reachable in practice: addressStallStatus escalates to "stalled" on the finalRoundTimeoutMs
-    // clock while a re-review is still in flight, and roundView still sets dot: true. A dot that
-    // followed currentColor would pulse RED here — DESIGN.md forbids a pulsing subordinate red
-    // (that loudest red is reserved for the blocked-agent pip; Four-Light Rule).
+  it("stalled + reviewing: the REVIEWING pill wins — no red pill, and no red dot", async () => {
+    // Two ways in. (1) forceReview persists a streak reset (addressRound: 0) but emits no onChange
+    // until the verdict lands, so this cached verdict is a STALE claim of a stall the server already
+    // cleared. (2) addressStallStatus escalates to "stalled" on the finalRoundTimeoutMs clock while
+    // a re-review is still in flight — there the stall is real, and the pill deliberately flips
+    // STALLED → REVIEWING → STALLED (an in-flight run supersedes a timeout's guess).
+    // Either way the running dot must not sit inside a blocked-red pill: recoloring it to follow
+    // currentColor would pulse RED, which DESIGN.md forbids for a subordinate (the loudest red is
+    // the blocked-agent pip; Four-Light Rule). So the pill yields instead of the dot changing hue.
     reviews.map = { s1: stalledVerdict() };
     reviews.reviewing = { s1: true };
     render(CriticBadge, { sessionId: "s1" });
-    await expect.element(page.getByText(/STALLED REVIEW/)).toBeInTheDocument();
 
+    expect(document.querySelectorAll(".critic-badge").length, "exactly one .critic-badge").toBe(1);
+    await expect.element(page.getByText(/REVIEWING/)).toBeInTheDocument();
+    expect(page.getByText(/STALLED REVIEW/).elements().length, "no STALLED REVIEW text").toBe(0);
+    expect(document.querySelector(".streak-stalled"), "no stalled pill while reviewing").toBeNull();
+
+    // the running dot is present and amber — the reviewing pill is where it belongs
     const dot = document.querySelector(".rev-dot") as HTMLElement;
-    expect(dot, "running dot rendered while re-reviewing a stalled streak").not.toBeNull();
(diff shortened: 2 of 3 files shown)


scoop (commented):
## Verdict

The change does what the task asks and I could not find a correctness defect in it.

### What the fix does

`ui/src/lib/components/CriticBadge.svelte:71` now short-circuits the streak branch when the streak is stalled *and* a re-review is in flight, so the badge falls through to the REVIEWING pill instead of painting a blocked-red `STALLED REVIEW` pill around a pulsing amber `.rev-dot`. That is exactly the weird state the task describes.

### Verification

- **The fallthrough always lands on the reviewing pill.** `chip` is `criticChip(verdict, reviewing)` (CriticBadge.svelte:21), and `criticChip` returns `{kind: "reviewing"}` unconditionally when `reviewing` is true (ui/src/lib/components/critic-badge.ts:36). So when the new guard skips `roundView`, the next branch (CriticBadge.svelte:72) always matches — the badge can never fall through to `null` and disappear. No hole here.
- **The `.streak-stalled` + `.critic-reviewing` combination is now genuinely unreachable**, as the rewritten CSS comment claims: the only place that emits `critic-reviewing` alongside a `streak-*` class is `roundView` (CriticBadge.svelte:43), which the new guard no longer reaches for `status === "stalled"`. The `.streak-stalled` rule (CriticBadge.svelte:311-315) stays text+border only, matching DESIGN.md:220 ("subordinate red … never takes a halo, pulse, or wash") and DESIGN.md:152 (Four-Light Rule). So the alternative fix — recoloring the dot to `currentColor` — would indeed have produced a forbidden pulsing red; yielding the pill is the correct lever.
- **The comment's factual claims check out.** `forceReview` persists `addressRound: 0` / `finalRoundPending: false` with an explicit "Deliberately NO onChange here" (src/review.ts:486-500), so a stalled verdict really can be a stale cached claim during a forced re-review. And the plan gate really does apply the same precedence — `if (reviewing) return { kind: "reviewing" }` precedes the verdict branches (ui/src/lib/components/plan-gate-badge.ts:49) and `canShowPlanStallActions` requires `!reviewing` (plan-gate-badge.ts:66-77). The badge is now consistent with its structural twin.
- **No other consumer regresses.** `addressRoundInfo` has exactly one caller (CriticBadge.svelte:22 — grepped the whole `ui/src` tree), and `m.criticbadge_stalled` / the `STALLED REVIEW` string appear only in CriticBadge.svelte and its browser test, so no other surface's expectations change.
- **i18n parity.** No new message keys are introduced; the keys the affected views use exist in both catalogs at matching positions (ui/messages/en.json:588-596 vs ui/messages/de.json:588-596).
- **Test assertions match the implementation.** `getByText(/REVIEWING/)` matches `criticbadge_reviewing` = `"REVIEWING…"` (ui/messages/en.json:588) and does not accidentally match `"STALLED REVIEW"` (en.json:593), so the paired positive/negative assertions in CriticBadge.browser.test.ts:245-248 are meaningful rather than tautological.
- **CommandBar deflake.** The guard being tested is `Date.now() - armedAt < ARM_DWELL_MS` with `armedAt = Date.now()` set at arm time (ui/src/lib/components/CommandBar.svelte:406-442). Pinning `Date.now` to `t0` therefore makes the second click provably land inside the dwell regardless of runner speed, and `now.mockReturnValue(t0 + 320)` provably lands past it — the rewritten test exercises the guard's arithmetic rather than the machine's scheduling, and the spy is restored in a `finally`. The logic is sound.

### Stated limitation

I could not verify whether `expect.element`'s retry loop inside the frozen-`Date.now` window computes its deadline from `Date.now` (the vitest browser runtime is not installed in this worktree — no `ui/node_modules`). If it does, a *failing* assertion inside the try-block would retry until the outer test timeout instead of failing at the element-retry timeout. That would make a failure slower, never a pass falser, so I am not raising it as a finding.

### Nit

The CommandBar dwell deflake is a good change but has nothing to do with the stalled-dot task; it rides along in the same PR. Worth calling out in the description (or splitting) so a bisect over the UI badge fix does not drag an unrelated test rewrite with it.

scoop:
**Re: the `CommandBar` deflake being unrelated (nit)** — agreed on the principle, and the finding's second remedy is now in place: the PR description calls it out explicitly, at the top as well as in detail at the bottom. (The disclosure section landed *after* the push you reviewed, so you were reading the pre-edit body — sorry for the noise.)

Taking the other option, splitting it into its own PR, isn't available to me here:

- I'm bound by a single-PR-per-session invariant, so I can't open a second PR from this session.
- The deflake had to land on *this* branch regardless: `verify` was red, and a red PR doesn't get reviewed at all. Leaving it would have blocked this PR indefinitely on a failure its own diff didn't cause.

On the bisectability concern specifically: the two changes are already separate commits touching disjoint files — `f527cc27` touches `CriticBadge.svelte` + its test, `f491c486` touches only `CommandBar.browser.test.ts`. Reverting either one leaves the other intact, so the badge fix is independently revertable exactly as you'd want.

Worth flagging for someone with the authority to schedule it: the flake is pre-existing (from #1751) and will keep hitting *any* PR that draws a slow runner — the guard reads wall time (`Date.now() - armedAt < ARM_DWELL_MS`, `CommandBar.svelte:442`) while the test fired two real clicks hoping both landed inside a 300ms window. That's why I fixed the root cause rather than re-running the job.

PR #1769, opened 2026-07-14
feat(learnings): configure Distiller provider and cadence

kai-osthoff:
## Problem / goal

The Learnings Distiller always fell back to Claude Code even when Shepherd’s default and role settings selected Codex. Its daily per-repository automatic sweep could repeatedly start expensive runs with no configurable spacing.

## Solution / added

- Added a persisted Distiller environment: CLI (`inherit`, Claude, or Codex), provider-compatible model, and reasoning effort.
- Resolved the environment immediately before every spawn; inherited settings follow global provider, model, and effort.
- Added a persisted 1–14-day interval for automatic runs, defaulting to one day. Manual runs remain an explicit override.

## Technical details

- Codex spawns use the shared `codex exec` argv path; explicit Claude retains the existing `writer-ro` invocation.
- Anthropic API-key fail-closed behavior now applies only to Claude, so it does not block a Codex Distiller.
- Successful automatic spawns persist a per-repository timestamp, which also protects the startup sweep from bypassing the chosen interval.
- Added the Settings control, EN/DE text, feature announcement, and focused regression coverage.

## Validation

- `bun run lint`
- `bun test test/distiller.test.ts test/server-settings.test.ts`
- `cd ui && bun run check`
- `cd ui && bun run test -- --project browser src/lib/components/Settings.browser.test.ts`

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -108,6 +108,8 @@ export const GH_PROBE_RETRY_DELAY_MS = 250;
 // process GROUP is SIGKILLed (see diagnostics.ts runRemediation) so no reparented
 // grandchild survives to flip a later probe green behind a reported failure.
 export const REMEDIATION_TIMEOUT_MS = 120_000;
+export const DISTILLER_INTERVAL_DAYS_MIN = 1;
+export const DISTILLER_INTERVAL_DAYS_MAX = 14;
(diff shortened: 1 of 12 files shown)


kai-osthoff:
@scoop 🚨✨ PR is READY, CI is green—your move, heroic reviewer. 🫡

PR #1772, opened 2026-07-14
fix(ui): keep drain cap/ceiling editable during an epic

scoop:
## Problem

The Auto-Drain **cap** and **usage-ceiling** dials (`AutomationDrainFields`) were gated on `drainRailsActive = flags.autoDrain && !epicActive && !lightweight`, and the Auto-Drain toggle is disabled while an epic runs. So once an epic kicks off, the whole rails group disappears and the toggle can't be flipped — the capacity setting is effectively **unreachable** for the duration of the epic.

But that setting is exactly what's still governing the epic: during an epic, `Drain.buildState` still copies `cfg.maxAuto` / `cfg.usageCeilingPct` into the drain state (`src/drain.ts`), and `computeNext` still enforces both — `capHold` (`src/drain-core.ts:317`) and `usageHold` (`:321`) — against the epic's children. So the one live, still-enforced knob was hidden precisely when tuning it matters most.

## Fix

- **Cap + usage ceiling** now show whenever draining is happening — label-drain on **or** an epic running: `drainRailsActive = !lightweight && (flags.autoDrain || epicActive)`. Editing them writes straight through `repoConfig.setMaxAuto` / `setUsageCeiling` (no server-side epic guard existed — only the UI was hiding the control), and the drain pump picks up the new value on its next cycle.
- **Label** field stays hidden mid-epic — label-drain is genuinely suspended while an epic owns the candidate set, so its dial is inert. Gated inside the component on a new `epicActive` prop.
- **Intro copy** switches to an epic-scoped line (`automation_drain_fields_intro_epic`, EN+DE) so the panel doesn't claim the dials "apply only while Auto-Drain is on" when an epic is driving them.

## Verification

- `cd ui && bun run check:i18n` → 2 locales in parity (new key present in both)
- prettier + eslint clean (pre-commit + pre-push hooks passed)

Purely a visibility/reachability change to an existing settings panel — no drain-core or persistence logic touched. `[no-feature-entry]` (a fix to an existing control's reachability, not a new user-facing feature).

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -2229,6 +2229,7 @@
   "topbar_held_spawning": "Wird gestartet…",
   "topbar_held_discarding": "Wird verworfen…",
   "automation_drain_fields_intro": "Diese Regler steuern die Backlog-Abarbeitung und gelten nur, solange sie aktiv ist:",
+  "automation_drain_fields_intro_epic": "Diese Regler steuern weiterhin die Abarbeitung des laufenden Epics — passe sie an, um es im Lauf nachzujustieren:",
   "drain_cap_hint": "Wie viele auto-gestartete Agenten höchstens gleichzeitig laufen. Ist das Limit erreicht, wartet Shepherd auf einen freien Platz, bevor das nächste Issue aufgegriffen wird.",
   "drain_label_hint": "Nur Issues mit diesem Label werden aufgegriffen — versieh die Tickets, die automatisch abgearbeitet werden sollen, damit.",
   "drain_ceiling_hint": "Pausiert neue Auto-Starts, sobald die Konto-Auslastung diesen Prozentwert erreicht. Ein höherer Wert lässt die Abarbeitung länger weiterlaufen; Standard ist 80.",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -2229,6 +2229,7 @@
   "topbar_held_spawning": "Starting…",
   "topbar_held_discarding": "Discarding…",
   "automation_drain_fields_intro": "These dials govern Auto-Drain and apply only while it's on:",
+  "automation_drain_fields_intro_epic": "These dials still govern the running epic's draining — adjust them to retune it in flight:",
   "drain_cap_hint": "Most auto-started agents allowed to run at once. Once it's reached, Shepherd waits for a slot to free before picking up the next issue.",
   "drain_label_hint": "Only issues carrying this label get picked up — tag the tickets you want worked automatically with it.",
   "drain_ceiling_hint": "Pause new auto-starts once your account usage reaches this percentage. A higher value lets the drain keep going longer; the default is 80.",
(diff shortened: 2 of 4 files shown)


scoop (commented):
## Summary

The PR makes the Auto-Drain "rails" (cap / usage ceiling) reachable and editable while an epic is running, hiding only the label field (label-drain is suspended mid-epic) and swapping in epic-scoped intro copy. This directly answers the task ("where is the capacity for epic draining set?") — the epic's spawn capacity is the drain **cap** (`maxAuto`), and this change keeps that dial live during an epic. Implementation is correct and I found no blocking issues.

## Verification

- **Core gate change** — `drainRailsActive` went from `flags.autoDrain && !epicActive && !lightweight` to `!lightweight && (flags.autoDrain || epicActive)` (verified `ui/src/lib/components/AutomationSettings.svelte:62`). During an epic on a forge repo this is now always true, so the group renders. `epicActive` is `drain?.epicParent != null` (`AutomationSettings.svelte:50`).
- **Prop threading / single caller** — `AutomationDrainFields` is imported and rendered in exactly one place (`AutomationSettings.svelte:11,446`); `{epicActive}` is passed through and the component declares `epicActive = false` default (`AutomationDrainFields.svelte:16-20`), so no other consumer is affected.
- **Label field hidden mid-epic** — gated by `{#if !epicActive}` around both the input and its hint (`AutomationDrainFields.svelte:88-102`); cap and ceiling stay rendered.
- **Intro copy switch** — `epicActive ? m.automation_drain_fields_intro_epic() : m.automation_drain_fields_intro()` (`AutomationDrainFields.svelte:72`).
- **Persistence works during epic** — `commitDrainCap`/`commitDrainCeiling` call `repoConfig.setMaxAuto` / `setUsageCeiling`, which patch settings unconditionally (verified `ui/src/lib/reviews.svelte.ts:657-663, 673-679`); edits made during an epic are persisted, so "editable during an epic" is genuinely functional, not cosmetic.
- **i18n parity** — the one new key `automation_drain_fields_intro_epic` is present with identical spelling in both catalogs: `ui/messages/en.json:2232` (baseLocale, per `ui/project.inlang/settings.json:3`) and `ui/messages/de.json:2232`. The `m.automation_drain_fields_intro_epic()` accessor matches the key. No other new keys were added.

- During an epic the Auto-Drain toggle renders visually **off** and disabled (`AutomationSettings.svelte:424,428`) while the rails below it are shown and editable. This is the intended behavior of this PR and the epic-scoped intro copy explains it, so it is acceptable — flagging only so it isn't mistaken for a bug.

## Stated limitation (could not verify)

- The added comment claims the cap/ceiling "are still enforced against the epic's children (drain-core capHold/usageHold)" (`AutomationSettings.svelte:58`). This repo contains only the UI (no Rust/Go/Python backend present — glob for `**/*.{rs,go,py}` returned nothing), so I could not confirm the drain-core enforcement of `maxAuto`/`usageCeiling` against epic children against ground truth. Assumed as written; not treated as a finding.

scoop:
shepherd-author-note: Confirming the one item the critic flagged as "could not verify" — the cap/ceiling enforcement against epic children lives in TypeScript, not a separate backend: `Drain.buildState` copies `cfg.maxAuto`/`cfg.usageCeilingPct` into drain state during epic mode (`src/drain.ts:740-741`), and `computeNext` enforces both via `capHold` (`src/drain-core.ts:317`) and `usageHold` (`:321`). The critic's context only had `ui/`, hence the `.rs/.go/.py` glob miss. Claim verified against ground truth; no change needed.

PR #1775, opened 2026-07-15
feat(ui): align issue selectors across task workflows

kai-osthoff:
## Problem / Goal

Issue selection appears in three places—New Task, Repos, and Up Next—but the surfaces had drifted into different row densities, label treatments, action layouts, and responsive behavior. The goal is a consistent, modern issue-selection experience without changing the purpose of each workflow.

## Solution / Added

- Applied one shared issue-row visual recipe across New Task, Repos, and Up Next.
- Added reusable, forge-aware label chips with responsive overflow counts.
- Kept issue actions reachable on narrow screens and preserved mobile touch targets.
- Carried label colors through the Up Next server snapshot so all surfaces render labels consistently.
- Preserved upstream slash-command disabled states and epic ownership indicators during the rebase.
- Added localized EN/DE copy, design-system documentation, and a What's New announcement for v1.44.0.

## Technical details

- Added `IssueLabelChips.svelte` for shared label rendering and color contrast handling.
- Moved the common row anatomy into semantic-token styles in `app.css`.
- Extended Up Next issue data with `labelColors` and covered standalone plus epic-parent propagation.
- Added browser coverage for label caps, action reachability, touch geometry, and consistent row behavior across all three selectors.

## Validation

- `bun run lint`
- `bun test test/up-next-core.test.ts test/up-next.test.ts` — 59 passed
- `cd ui && bun run check` — 0 errors, 0 warnings
- `cd ui && bun run test` — 255 files / 3,810 tests passed
- Branch hygiene, feature catalog, announcement version, i18n, and glossary gates passed

PR #1784, opened 2026-07-15
fix(ui): restore Up Next row density (inline START)

scoop:
## Problem

After #1775 (\"align issue selector rows\") unified the Up Next row onto the shared `.issue-list-row` recipe and added label chips, a container query forced the per-row **START** button to `flex-basis: 100%` — its own line — whenever a row is narrower than **520px**. The Up Next sidebar is ~460px, so *every* row wrapped, doubling its height and gutting the panel's information density.

| | Before this PR | After |
|---|---|---|
| Desktop sidebar (~460px) | START on its own line, 2-line rows | START inline, single-line rows |
| Mobile / coarse-pointer | START inline (2-line, cramped) | START hidden; tick → sticky batch bar |

## Change (CSS-only + tests)

- **Remove the wrap.** Drop `.un-row { flex-wrap: wrap }` and the `@container (max-width: 520px) { .un-actions { flex-basis: 100% } }` block so START stays inline at every desktop width.
- **Narrow-width shrink-valve.** With strict `nowrap`, the pills/age/START are all `flex: none` and `.issue-list-title` has a 6rem floor, so a worst-case **priority + epic + chips** row could overflow at ~320–400px. Lower `.un-title` `min-width` to 0 below 460px (Up Next only — `IssueRow`/`PromptSources` keep the shared floor). Guaranteed no overflow ≥ 360px desktop.
- **Hide START on mobile/coarse-pointer** (`.un-actions { display: none }` — also leaves the a11y tree). Starting one issue goes through the checkbox → batch bar's \"Start selected (1)\".
- **Sticky batch bar** (`position: sticky; top: 0`) so that single-start affordance can't scroll off-screen when a row deep in a long list is ticked.

### Design decision

Hiding per-row START on mobile in favour of the checkbox → batch-bar path is an **explicit operator choice** made against a live mockup (the simpler \"inline everywhere\" option was offered and declined). The sticky batch bar resolves the single-start UX gap that hiding the inline button would otherwise introduce.

## Testing

- `bun run check` (svelte-check): 0 errors.
- `bun run check:i18n`: locales in parity (no new strings — `fix`, not `feat`, so no feature-catalog entry).
- Prettier + ESLint clean (pre-commit hook).
- `UpNextPanel.browser.test.ts` (32 passing), including two updated/added guards that render the real component against `app.css` and assert layout geometry:
  - **Mobile (390px):** `.un-actions` computed `display: none`, START rect height 0, checkbox + link ≥ 44px, no row overflow.
  - **Desktop worst-case (priority + epic + 2 chips):** at 360px the row stays single-line (`offsetHeight` equals its 600px height → START never wrapped) and `scrollWidth ≤ clientWidth`.

### DIFF
--- a/ui/src/lib/components/UpNextPanel.browser.test.ts
+++ b/ui/src/lib/components/UpNextPanel.browser.test.ts
@@ -262,7 +262,7 @@ describe("UpNextPanel compact issue rows", () => {
     expect(row.querySelector(".issue-list-actions .un-start")).not.toBeNull();
   });
 
-  it("keeps checkbox, issue link, and Start touch-sized without narrow-row overflow", async () => {
+  it("hides the per-row Start on mobile while checkbox and link stay touch-sized", async () => {
     await page.viewport(390, 800);
     render(UpNextPanel, {});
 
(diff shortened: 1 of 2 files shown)


scoop (commented):
## Summary

This PR restores Up Next row density by keeping the START button inline instead of letting it wrap to its own line. The mechanism is coherent and matches the stated task:

1. Removes the mobile `.un-row { flex-wrap: wrap }` override so the row is always `nowrap` — START can no longer wrap to a second line (`ui/src/lib/components/UpNextPanel.svelte`, deleted lines around 762-764).
2. Replaces the old `@container (max-width: 520px)` rule that pushed `.un-actions` to `flex-basis: 100%` with a `@container (max-width: 460px) { .un-title { min-width: 0 } }` valve, so at narrow widths the title (not the button) shrinks (UpNextPanel.svelte:868-872).
3. On mobile/coarse-pointer, hides the per-row START entirely (`.un-actions { display: none }`), routing single-issue start through checkbox → sticky batch bar (UpNextPanel.svelte:886-891).
4. Makes `.un-batch` `position: sticky; top: 0; z-index: 2` so the now-sole start affordance stays reachable when a row is ticked deep in a long list (UpNextPanel.svelte:687-702).

## Verification performed

- **Shrink-valve wins specificity**: `.un-title` is Svelte-scoped (`.un-title.svelte-hash`, specificity 0,2,0) so its `min-width: 0` overrides the global `.issue-list-title { min-width: 6rem }` (0,1,0) — verified against `ui/src/app.css:465-467`. Above 460px no scoped `min-width` is set, so the shared 6rem floor applies. Behavior matches the code comment.
- **No overflow guarantee holds**: with title floored at 0, the other shrink valve is `IssueLabelChips`, which is `flex: 0 1 auto; min-width: 0` and collapses to a single "+N" chip at ≤520px — verified against `ui/src/lib/components/IssueLabelChips.svelte:52-57,105-114`. Priority/epic pills, age, num and START are all `flex: none` (UpNextPanel.svelte:794,802,807,829,832).
- **Sticky batch bar**: `.upnext` is the scroll container (`overflow: auto`, UpNextPanel.svelte:597) and `.un-batch` is a direct child with an opaque `background: var(--color-inset)` (UpNextPanel.svelte:700), so sticky pins correctly and covers scrolled content. The header is intentionally non-sticky.
- **Mobile hide affects a11y correctly**: `display: none` removes START from tab order / a11y tree as the comment claims; the checkbox→batch path remains.
- **Selector containment**: `.un-actions` appears only in `UpNextPanel.svelte` and its test — verified via tree-wide grep. The 460px override is scoped to `.un-title`, so IssueRow / PromptSources keep the 6rem floor.
- **i18n parity**: no new message keys introduced by the diff. Existing keys used by the template (`upnext_start`, `upnext_batch_aria`) are present in both catalogs — verified against `ui/messages/en.json:2304,2313` and `ui/messages/de.json:2304,2313`.
- **Test updates track the code**: the renamed/added browser tests assert the new contract (`.un-actions` `display: none` and zero-height `.un-start` at 390px; inline START + single-line at 360px desktop) — `ui/src/lib/components/UpNextPanel.browser.test.ts:265-338`. No other test relies on the removed 520px/flex-wrap behavior; desktop `startButtons()` clicks run at the 1024px viewport where START stays visible.

No correctness, security, or quality issues found. The implementation satisfies the task.

PR #1678, opened 2026-07-12
feat(issues): hide blocked-labeled issues across Up Next & backlog

scoop:
## What

A zero-config **"Hide blocked issues"** toggle that filters out issues whose labels mark them
blocked — like our `blocked-upstream` label — across three surfaces:

- **Up Next** lens (new compact header toggle `⊘`)
- **Repo pane** / Backlog `IssuesPanel` (new row in the Filters popover)
- **New-Task** issue picker `PromptSources` (shares the same popover)

The toggle is a single persisted, viewer-agnostic preference (localStorage), **on by default**, so
blocked work stays out of the way out of the box. Flip it off to bring blocked issues back.

## Match rule

"Blocked" = a `blocked` word at a **word boundary**, case-insensitive (`/\bblocked/i`):

- **Matches:** `blocked`, `blocked-upstream`, `blocked-by-net`, and namespaced forms `status/blocked`,
  `S: blocked`, `kind:blocked`.
- **Skips:** `unblocked`, `needs-unblock`, `unblock-me` — a word boundary, not a substring, so
  semantically-ready labels are never hidden. This precision is what makes default-ON safe.

One shared client helper (`isBlocked`) drives all three surfaces, so they can't diverge.

## Details

- **Repo pane:** the blocked filter sits between the sub-issue and author/label filters; a dedicated
  `allHiddenByBlocked` empty-state explains a fully-filtered list; and blocked-word labels drop out
  of the include-chip options while the toggle is ON (so an include-chip can't silently yield an
  empty list).
- **Up Next:** `UpNextItem` now carries `labels` (populated server-side — standalone issues from
  their own labels, epic rows from the parent's). The filter runs in both the unfiltered and
  repo-filtered paths of the sections derivation, recounts section totals, and drops emptied
  sections.
- **Server ranking unchanged:** Up Next's pre-existing `EXCLUDE_LABELS` (exact `blocked`/`wontfix`)
  is a separate *ranking* exclusion and is intentionally left as-is. Consequence: toggling OFF
  reveals `blocked-*` issues (e.g. `blocked-upstream`) everywhere; an exact-`blocked` issue reappears
  in the backlog/New-Task picker but stays out of the Up Next *startable* queue — matching Up Next's
  other startability exclusions.
- Ships a What's-New announcement entry (`v1.43.0-hide-blocked-toggle`), differentiated from the
  sibling dependency-blocked-awareness and author/label-filter cards. Full EN/DE i18n.

## Testing

[...]

PR #1792, opened 2026-07-15
fix(up-next): remove dead hide-blocked toggle; exclude blocked variants server-side

scoop:
## Problem

The **hide-blocked toggle in Up Next did nothing.** The Up Next server (`src/up-next-core.ts`) already excludes blocked-labeled and dependency-blocked work *before it reaches the queue* (predates the toggle, from #1169). The toggle added later in #1678 (`UpNextPanel.svelte`) re-filtered that same already-filtered queue, so in the common case there was nothing left to hide/show — it only ever affected rare variant labels (`blocked-upstream`) that slipped past the server's exact match.

## Decision

Up Next is a **start-next queue** — every row carries a "start" affordance, so only *startable* work belongs there. A blocked issue is non-startable, so the fix is to **remove the redundant toggle** (confirmed with the operator), not to make blocked issues startable. The Backlog + New Task issue picker keep their genuinely-useful toggle.

The two halves are coupled: the toggle defaulted to *hide*, so variant-blocked items were hidden by default today. Removing it without tightening the server would newly **expose** them — so the server exclusion is broadened to match.

## Changes

- **Server** (`up-next-core.ts`): broaden the blocked exclusion from exact `blocked` to `/\bblocked/i` (mirrors the UI's existing `isBlocked` for parity) so variant labels (`blocked-upstream`, `status:blocked`) are also kept out. `wontfix`/`wont-fix` stay exact; `blockedBy` dependency exclusion untouched. Edge behavior (matches `not-blocked`; **not** `unblocked`/`blocking`) is inherited from the UI rule, not new.
- **UI** (`UpNextPanel.svelte`): remove the `.un-blockedbtn` button, `applyBlocked`/`allHiddenByBlocked` plumbing, its empty-state branch, CSS, and now-unused imports (`isBlocked`, `issuesFilter`, `coachTarget`).
- **Messages**: drop the unused `upnext_hide_blocked` key; correct the shipped 1.43.0 announcement **title + body** (no longer "everywhere"/"and Up Next"; note Up Next always shows startable work) and drop its now-dead coachmark `targetId`.
- **Tests**: server asserts variant-blocked labels are excluded; browser test drops the obsolete hide-blocked describe block.

### Intended asymmetry

After this, the shared `hideBlocked` toggle no longer governs Up Next. Toggling it off in Backlog / New Task reveals blocked issues there, while Up Next keeps hiding them unconditionally — deliberate, since Up Next only lists startable work. No new empty-state copy (generic "all caught up" fits); the announcement clause makes it discoverable. `[no-feature-entry]` — this removes UX, it doesn't add a feature.

## Verification

- Root: `bun run lint` ✓, up-next server tests ✓ (77 pass)
- UI: `bun run check` ✓ (0 errors), `bun run check:i18n` ✓ (parity), `UpNextPanel.browser.test.ts` ✓ (28 pass), `issues-panel.test.ts` ✓ (69 pass)
- Pre-push gates (prettier/eslint/tsc/root-tests/ui/fallow audit) ✓


scoop (commented):
## Summary

The PR resolves "the blocked filter in Up Next does nothing" by moving blocked-issue exclusion server-side into `src/up-next-core.ts` and removing the redundant client-side hide-blocked toggle from `UpNextPanel.svelte`. This is coherent: Up Next is meant to list only startable work (it already drops dependency-blocked issues via `blockedBy`), and exact-`blocked` was already partially excluded server-side (old `EXCLUDE_LABELS`), which is exactly why the client toggle appeared to "do nothing" for the common case.

## Verification

- **Complete coverage of row producers.** `repoItems` produces rows via only two paths — `standaloneItem`→`isExcludedIssue` (src/up-next-core.ts:161) and `epicItem` (src/up-next-core.ts:197). Both now call `hasBlockedLabel`. Confirmed against src/up-next-core.ts:218-239 (no other item-producing path; `fallback` is a message string, not items).
- **Regex parity with the UI.** Server `BLOCKED_LABEL_RE = /\bblocked/i` (src/up-next-core.ts:117) is byte-identical to the canonical UI matcher at ui/src/lib/components/issues-panel.ts:48, which the doc-comment claims to mirror — verified. Both match `blocked`, `blocked-upstream`, `status:blocked` and correctly reject `unblocked`.
- **No regression from dropping `"blocked"` from `EXCLUDE_LABELS`.** `\bblocked` still matches exact `blocked`, so the prior exclusion is preserved and extended to boundary variants (src/up-next-core.ts:112,160-161).
- **No dangling client refs.** Grepped UpNextPanel.svelte and the browser test for `applyBlocked`, `sectionsPreBlocked`, `allHiddenByBlocked`, `un-blockedbtn`, `issuesFilter`, `isBlocked`, `coachTarget` — no matches remain.
- **Feature-announcement `targetId` removal is safe.** `targetId` is optional (ui/src/lib/feature-announcements.ts:30) and many entries omit it; the anchor button no longer exists, so removal is correct (ui/src/lib/feature-announcements/entries/v1.43.0-hide-blocked-toggle.ts).
- **i18n parity.** `upnext_hide_blocked` removed from BOTH catalogs; `feat_hide_blocked_toggle_title`/`_body` updated in both. Verified en.json:3018-3019 and de.json:3018-3019; no remaining reference to `upnext_hide_blocked` anywhere in the tree.
- **Test correctness.** The new server test (test/up-next-core.test.ts) asserts `unblocked` (#3) stays visible while `blocked-upstream`/`status:blocked` are dropped — consistent with `\bblocked`.

No blocking issues and no nits worth changing found.

PR #1793, opened 2026-07-15
fix(ui): show repo on Owed lens cards

scoop:
## Problem

The Owed lens (`PostMergeStepsPanel`) aggregates outstanding post-merge manual-step cards across **all** repos, but a card showed only its designation (`TASK-1503`), PR number, count, and title — never **which repo** the steps affect. In a multi-repo herd there was no way to tell a `TASK-1503` in one repo from another.

## Change

Add a quiet **icon + repo-name** marker leading each card's identity row, following the nearest structural precedent — `EpicGroupHeader`'s repo marker (icon rendered only when a project emoji is set, **no `▣` fallback**; emoji-less repos render name-only), **not** the interactive RepoSwitcher chip.

- **Live card** — `PostMergeSteps.repoPath` was already on the record; just render it.
- **Frozen fallback card** — added `repoPath` to `OwedFocusSnapshot`, populated from `s.repoPath` at the single construction site (`onShowOwed`).

Both variants key `projectIcons.iconFor` on the **same** `session.repoPath` (the server builds the live record with `repoPath: session.repoPath` in `src/post-merge-steps.ts`; the snapshot copies `s.repoPath`), so the icon resolves identically on both — no silent fallback on one card variant but not the other.

The repo name is verbatim path data (not app chrome) → not translated, no new i18n key. The emoji is `aria-hidden`; the basename stays screen-reader readable. Styling uses design-system tokens only (quiet `--color-muted`, `--fs-meta`).

## Tests

Expanded `PostMergeStepsPanel.browser.test.ts`:
- repo basename renders on a **live** card (with project emoji shown when set);
- repo basename renders on the **frozen** fallback card;
- **unset-icon path**: an emoji-less repo renders **name-only** — no `▣`, no stray glyph.

All 21 panel tests + full UI suite (3866) pass; `bun run check`, `check:i18n`, prettier, eslint clean.

### DIFF
--- a/ui/src/lib/components/PostMergeStepsPanel.browser.test.ts
+++ b/ui/src/lib/components/PostMergeStepsPanel.browser.test.ts
@@ -30,6 +30,7 @@ vi.mock("$lib/api", async (importOriginal) => {
 });
 
 const { postMergeSteps } = await import("$lib/post-merge-steps.svelte");
+const { projectIcons } = await import("$lib/projectIcons.svelte");
 const PostMergeStepsPanel = (await import("./PostMergeStepsPanel.svelte")).default;
 
 function record(p: Partial<PostMergeSteps> & { sessionId: string }): PostMergeSteps {
@@ -54,6 +55,7 @@ function record(p: Partial<PostMergeSteps> & { sessionId: string }): PostMergeSt
 function snapshot(p: Partial<OwedFocusSnapshot> & { sessionId: string }): OwedFocusSnapshot {
   return {
     desig: "TASK-09",
+    repoPath: "/repo/shepherd",
     prNumber: 12,
     steps: [
       { id: "ms1", text: "Set FLAG=1", postMerge: false },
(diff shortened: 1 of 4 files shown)


scoop (commented):
## Summary

The PR adds a quiet repo-identity marker (basename of `repoPath` + optional project emoji) to both the live and frozen cards of the Owed lens, addressing the task "owed lens has no reference to the repo it affects." It threads `repoPath` through `OwedFocusSnapshot` and populates it at the sole construction site. The change is correct and internally consistent.

## Verification

- `basename` import — resolves to `ui/src/lib/components/learnings-drawer.ts:4` (imported as `./learnings-drawer` from a component in `ui/src/lib/components/`). Correct path and export.
- `projectIcons.iconFor` — verified against `ui/src/lib/projectIcons.svelte.ts:23` (returns emoji or `null`); singleton exported at `ui/src/lib/projectIcons.svelte.ts:44`.
- Live card uses `rec.repoPath` — `PostMergeSteps.repoPath` already exists at `ui/src/lib/types.ts:1109`. Valid.
- Frozen card uses `frozenCard.repoPath` — new field `OwedFocusSnapshot.repoPath` at `ui/src/lib/types.ts:1126`.
- Sole production constructor of `OwedFocusSnapshot` is `ui/src/routes/+page.svelte:2455`, now sets `repoPath: s.repoPath`; `Session.repoPath` exists at `ui/src/lib/types.ts:1003`. The other assignment (`+page.svelte:2485`) is `= null`. No other constructors exist (grepped all `.ts`/`.svelte`), so `repoPath` is always populated when a snapshot is built.
- "Mirrors EpicGroupHeader" behavior claim — confirmed against `ui/src/lib/components/EpicGroupHeader.svelte:49-52`: icon rendered only when `iconFor` is set, no `▣` fallback, `title` carries the full path. Matches the new `.ow-repo` markup.
- No message-catalog keys added or changed in the diff, so no i18n parity concern.

## Nit (non-blocking)

`projectIcons.iconFor(...)` is invoked twice per card (once in the `{#if}` guard, once in the emitted content) on both the frozen and live card. It is a trivial `Map` lookup so the cost is negligible, but `EpicGroupHeader` avoids the double call via a `$derived(projectIcons.iconFor(epic.repoPath))` (`EpicGroupHeader.svelte:32`). Optional to align with that pattern.

scoop (commented):
## Re-review

The PR adds a quiet repo-identity marker (basename + optional project emoji) to both the live and frozen Owed-lens cards, threading `repoPath` through `OwedFocusSnapshot` and the snapshot builder. This satisfies the task ("owed lens has no reference to the repo it affects").

### Prior finding
1. **Double `projectIcons.iconFor(...)` per card** — **Addressed.** Now hoisted to a per-card `{@const frozenIcon}` (PostMergeStepsPanel.svelte:154) and `{@const repoIcon}` (:192), each computed once and reused in both the `{#if}` guard and the rendered content, exactly as EpicGroupHeader.svelte:32 does.

### Verification
- `projectIcons.iconFor(path: string): string | null` exists — verified against ui/src/lib/projectIcons.svelte.ts:23-25.
- `basename` is exported at ui/src/lib/components/learnings-drawer.ts:4; the import `./learnings-drawer` resolves to that `.ts` file (the sibling `learnings-drawer/` directory has no `index.*`, so there is no resolution collision).
- `OwedFocusSnapshot.repoPath` added (ui/src/lib/types.ts:1126); `PostMergeSteps.repoPath` already existed (ui/src/lib/types.ts:1109), so `rec.repoPath` on the live path is valid.
- `+page.svelte:2458` populates `repoPath: s.repoPath`; `Session.repoPath` exists (ui/src/lib/types.ts:1003). Valid.
- The icon is rendered in its own child `<span class="ow-repo-icon">` so the `.ow-repo` element's direct text node is just the basename — consistent with the test's `getByText("shepherd")` and `querySelector('.ow-repo').textContent` assertions, and mirrors EpicGroupHeader.svelte:49-52.
- No new user-facing message keys are introduced, so i18n/locale parity is unaffected.

No blocking issues and no nits remain.

PR #1797, opened 2026-07-16
fix(codex-update): show completed version transition

kai-osthoff:
## Problem

After a Codex update finishes, the modal refreshes its availability status. Both displayed versions can then be the newly installed version, making the transition read `0.144.5 → 0.144.5` instead of showing what actually changed.

## Solution

The completed modal now uses the terminal update result to display the actual `from → to` version transition. The pre-update and in-progress view still uses the available update status.

## Technical details

- Derive the summary versions from `done.from` and `done.to` once a terminal result is available, with the existing status values as fallback.
- Add browser coverage for a refreshed `new → new` status following a successful update.
- Verified with the focused browser test, `bun run check`, and `bun run check:i18n`.

### DIFF
--- a/ui/src/lib/components/CodexUpdateModal.browser.test.ts
+++ b/ui/src/lib/components/CodexUpdateModal.browser.test.ts
@@ -78,4 +78,17 @@ describe("CodexUpdateModal", () => {
       ),
     );
   });
+
+  it("keeps the completed update's actual version transition after status refreshes", async () => {
+    const { rerender } = await render(CodexUpdateModal, { props: { update } });
+
+    await rerender({
+      update: { ...update, current: "0.142.5", latest: "0.142.5", updateAvailable: false },
+      done: { ok: true, from: "0.142.4", to: "0.142.5" },
+    });
+
+    await vi.waitFor(() =>
+      expect(document.querySelector(".versions")?.textContent).toContain("0.142.4 → 0.142.5"),
+    );
+  });
 });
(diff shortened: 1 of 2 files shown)


kai-osthoff (commented):
No scoped defects found.

Verification:

- The server captures the pre-update version as `from`, re-reads the installed version as `after`, publishes that refreshed status, and returns `{ from, to: after }` on success (`src/codex-update.ts:444`, `src/codex-update.ts:494`, `src/codex-update.ts:520`, `src/codex-update.ts:528`, `src/codex-update.ts:530`).
- The terminal result shape is shared by the WebSocket event and store (`ui/src/lib/types.ts:1450`, `ui/src/lib/types.ts:1764`, `ui/src/lib/store.svelte.ts:627`), and the parent passes both refreshed status and terminal result into this modal (`ui/src/lib/components/page/AppOverlays.svelte:475`).
- The modal now prefers the terminal result's `from`/`to` while retaining status fallbacks, and feeds those values into the existing localized version template (`ui/src/lib/components/CodexUpdateModal.svelte:27`, `ui/src/lib/components/CodexUpdateModal.svelte:82`, `ui/src/lib/components/CodexUpdateModal.svelte:85`). The added regression case models the status refresh from `0.142.4 → 0.142.5` to `0.142.5 → 0.142.5` and asserts the completed transition remains visible (`ui/src/lib/components/CodexUpdateModal.browser.test.ts:82`).
- Fresh-run and close handlers clear old terminal results, so a prior completion cannot be reused by a later modal run (`ui/src/routes/+page.svelte:560`, `ui/src/routes/+page.svelte:3002`, `ui/src/routes/+page.svelte:3007`).
- The `codexupdate_versions` interpolation key exists with the same placeholders in both maintained UI catalogs (`ui/messages/en.json:2387`, `ui/messages/de.json:2387`).

Per instruction, I did not run builds, tests, or lint.

PR #1804, opened 2026-07-16
fix(spawn): don't let a slash-leading issue title become a slash command

scoop:
## The bug

An issue titled `/foo bar` kills its own session. **Up Next start** (`server.ts`) and **auto-drain** (`drain.ts`) both passed the **bare** issue title as the spawn prompt, and that prompt is delivered as a **positional argv** to the agent CLI — where Claude Code parses a leading `/` as a slash command.

## Reproduced against the real argv shape

Both sites set `issueRef`, so `composePromptArg` appends a multi-line `GitHub Issue #N …` block after the title. That block does **not** defeat the parse — only the first token is consumed:

```
$ claude -p $'/zzznope-probe\n\nGitHub Issue #1 (title + body follow as untrusted data):\nsay only the word OK'
Unknown command: /zzznope-probe

$ claude -p $'zzznope-probe\n\nsay only the word OK'     # non-slash control
OK
```

There is no `--` separator, `--raw` flag, or escape character. Only "text does not begin with `/`" reliably avoids it.

## The fix

A pure helper, `issueSpawnPrompt(number, title)`, used by both sites:

| Issue title | Prompt before | Prompt after |
| --- | --- | --- |
| `/foo bar` | `/foo bar` → `Unknown command: /foo bar` | `Work on issue #12: /foo bar` |
| `/api/users returns 500` | `Unknown command: /api/users` | `Work on issue #12: /api/users returns 500` |
| `Fix login` | `Fix login` | `Fix login` (**byte-identical**) |

This mirrors **New Task**, which is already immune because it templates the title into `newtask_issue_prompt_template` (`"Work on issue #{number}: {title}"`). The byte-identical passthrough is load-bearing: the namer derives branch + worktree names from the prompt, so the common case must not shift.

## Design notes (the non-obvious parts)

- **Not sanitized centrally in `composePromptArg`**, even though it's the single funnel: a leading `/` on an **operator-authored** prompt is a *supported feature* (`commands.ts` — New Task's Commands tab and inline `/` picker spawn a session whose first message *is* a slash command). Only an issue-title-derived prompt can be neutralized, because there a leading `/` is never a command anyone asked for.
- **Unconditional across providers**, though Codex likely does *not* have this bug (it invokes skills as `$name`; `codex exec` forwarded a slash-leading prompt straight to the model API). The reason is `relaunch`: it reuses the stored prompt and **can switch provider**, so a prompt left untemplated on Codex would break the moment it relaunched on Claude. A per-provider gate would be a latent relaunch bug.
- **`trimStart()`, not bare `startsWith`**: a `" /foo"` title would otherwise rely on the CLI's leading-whitespace trimming — undocumented and provider-specific.

## Known costs (accepted, documented in the helper)

- **Branch slug for slash titles.** `"Work on issue #12: /foo bar"` slugs to `work-issue-12-foo` (measured): `"on"` is a STOPWORD but `work`/`issue` aren't, and the 4-word cap drops `bar`. It's no longer "strong", so the background Haiku refine fires and renames the branch via `git branch -m` — the path `BRANCH_RENAME_NOTICE` already documents. The worktree dir keeps the initial slug. Bounded to slash titles; this is the shape New Task ships today (`"Work on issue #12: Fix login"` → `work-issue-12-login`). Rejected alternatives: adding `work`/`issue` to STOPWORDS (breaks `test/namer.test.ts:109`, churns slugs repo-wide) and threading a `nameHint` (new API surface `relaunch` wouldn't persist).
- **Pre-existing sessions aren't retro-fixed.** `relaunch` reuses `s.prompt` rather than re-deriving from the title, so a session already stored with a bare `/foo` prompt stays broken — re-start the issue from Up Next. A store migration is disproportionate.
- **Intentional drift**: the server helper is fixed English (agent-facing, per house rules); the client template is i18n'd. Not shared by design — the helper's doc comment says so, so nobody later "fixes" the duplication.

## Testing

- New `test/issue-spawn-prompt.test.ts` — slash, path-like slash, bare `/`, leading-whitespace slash, non-slash passthrough.
- `test/server-up-next.test.ts` + `test/drain.test.ts` — assert the prompt reaching `create()` (ordinary title stays bare; slash title templated), drain covering the plain and epic-base variants.
- `bun run lint`, `bun run typecheck`, `bun run test` (**7161 pass, 0 fail**) and the full pre-push gate all pass. Server-only — `ui/` untouched, so no i18n keys and no feature-catalog entry (a `fix`, not a `feat`).

### DIFF
--- a/src/drain.ts
+++ b/src/drain.ts
@@ -6,6 +6,7 @@ import type { UsageLimits } from "./usage-limits";
 import type { TelemetryService } from "./telemetry";
 import { recordEpicIntegrationIfChild, settleMergedSession } from "./merge-teardown";
 import { isFullAuto } from "./full-auto";
+import { issueSpawnPrompt } from "./issue-spawn-prompt";
 import {
   ACTIVE_LABEL,
   computeNext,
@@ -2388,8 +2389,11 @@ export class DrainService {
     }
     // Epic child actually based on the integration branch → tell the agent to target it as the
     // PR base (the agent opens its own PR and would otherwise default to the main branch).
+    // The directive rides AFTER the templated title so a `/`-leading title can't reach argv
+    // position 0 and be parsed as a slash command (issueSpawnPrompt).
     const usingEpicBase = !!decision.integrationBranch && base === decision.integrationBranch;
-    const prompt = usingEpicBase ? `${title}\n\n${epicBaseDirective(base)}` : title;
+    const task = issueSpawnPrompt(number, title);
+    const prompt = usingEpicBase ? `${task}\n\n${epicBaseDirective(base)}` : task;
     return { base, prompt };
   }
 
--- a/src/issue-spawn-prompt.ts
+++ b/src/issue-spawn-prompt.ts
@@ -10,11 +10,22 @@
  *   $ claude -p $'/zzznope\n\nGitHub Issue #1 (title + body follow as untrusted data):\n…'
  *   Unknown command: /zzznope
  *
- * There is no `--` separator, `--raw` flag, or escape character; only "text does not begin with
- * `/`" reliably avoids the parse. So a `/`-leading title is templated into a sentence, mirroring
- * the client-side New Task path (`newtask_issue_prompt_template`) which is immune for exactly
- * this reason. Every other title passes through BYTE-IDENTICAL — that is the point: it keeps the
- * namer (and thus branch/worktree names) unchanged for the overwhelmingly common case.
+ * `--disable-slash-commands` does NOT rescue this, which is worth stating explicitly because auto
+ * (drain) spawns ALREADY pass it — trimDecision (service.ts) adds it whenever config.trimAutoContext
+ * is on, which is the default — so drain looks immune and isn't. Measured, same prompt:
+ *
+ *   $ claude -p --disable-slash-commands $'/zzznope\n\nGitHub Issue #1 …'
+ *   Unknown command: /zzznope
+ *
+ * And it would be unusable here even if it did work: it strips the whole skill catalog for the
+ * session (config.ts, transient-agent-argv.ts) and would kill the operator-authored leading-`/`
+ * first message that commands.ts supports. There is likewise no `--` separator, `--raw` flag, or
+ * escape character; only "text does not begin with `/`" reliably avoids the parse.
+ *
+ * So a `/`-leading title is templated into a sentence, mirroring the client-side New Task path
+ * (`newtask_issue_prompt_template`) which is immune for exactly this reason. Every other title
+ * passes through BYTE-IDENTICAL — that is the point: it keeps the namer (and thus branch/worktree
+ * names) unchanged for the overwhelmingly common case.
  *
  * NOT a general sanitizer, and deliberately NOT applied inside composePromptArg: a leading `/`
  * on an OPERATOR-authored prompt is a supported feature (commands.ts — New Task's Commands tab
--- a/src/issue-spawn-prompt.ts
+++ b/src/issue-spawn-prompt.ts
@@ -1,0 +1,41 @@
+/**
+ * Compose the spawn prompt for a session started FROM AN ISSUE TITLE (the Up Next start in
+ * server.ts and the auto-drain spawn in drain.ts). Both used to pass the bare title, which
+ * breaks the moment a title begins with `/`.
+ *
+ * WHY: the prompt is delivered as a positional argv to the agent CLI (service.ts `argv.push`),
+ * and Claude Code parses a leading `/` there as a slash command — only the first token, and the
+ * multi-line issue block composePromptArg appends after the title does NOT defeat it:
+ *
+ *   $ claude -p $'/zzznope\n\nGitHub Issue #1 (title + body follow as untrusted data):\n…'
+ *   Unknown command: /zzznope
+ *
+ * There is no `--` separator, `--raw` flag, or escape character; only "text does not begin with
+ * `/`" reliably avoids the parse. So a `/`-leading title is templated into a sentence, mirroring
+ * the client-side New Task path (`newtask_issue_prompt_template`) which is immune for exactly
+ * this reason. Every other title passes through BYTE-IDENTICAL — that is the point: it keeps the
+ * namer (and thus branch/worktree names) unchanged for the overwhelmingly common case.
+ *
+ * NOT a general sanitizer, and deliberately NOT applied inside composePromptArg: a leading `/`
+ * on an OPERATOR-authored prompt is a supported feature (commands.ts — New Task's Commands tab
+ * and the inline `/` picker spawn a session whose first message IS a slash command). Only a
+ * prompt derived from an issue title can be neutralized, because there a leading `/` is never a
+ * command anybody asked for.
+ *
+ * Applies unconditionally across providers even though Codex likely does NOT share the bug (it
+ * invokes skills as `$name`, commands.ts). `relaunch` reuses the stored prompt and can SWITCH
+ * provider, so a prompt left untemplated because it spawned on Codex would break as soon as it
+ * relaunched on Claude. Unconditional is correct-by-construction; a per-provider gate would be a
+ * latent relaunch bug.
+ *
+ * The English wording here is intentionally NOT shared with the i18n'd client-side
+ * `newtask_issue_prompt_template` (ui/messages/*.json): this is agent-facing spawn text (fixed
+ * English, like the drain prompts and directive blocks), that one is operator-visible UI. They
+ * may drift; nothing binds them and nothing needs to. Don't "fix" the duplication by wiring one
+ * to the other.
+ */
+export function issueSpawnPrompt(number: number, title: string): string {
+  // trimStart, not a bare startsWith: a " /foo" title would otherwise be left relying on the
+  // CLI's leading-whitespace trimming to dodge the parse — undocumented and provider-specific.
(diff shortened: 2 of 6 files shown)


scoop (commented):
## Verdict

The change does what the task asks. Both server-side spawn paths that turn an issue title into an agent prompt are now routed through `issueSpawnPrompt`, and the templated form can no longer land a `/` at argv position 0. I found no logic, security, or parity problems. One documentation-accuracy nit below.

## What I verified

**The premise (leading `/` in the spawn prompt is parsed as a command).** The prompt really is the trailing positional argv for both providers: `argv.push(promptArg)` at `src/service.ts:2454` (Claude) and `argv.push(\`${promptArg}\n\n<shepherd-directives>…\`)` at `src/service.ts:2496` (Codex). The diff's claim that the appended issue block does not defeat the parse checks out against `src/service.ts:1751-1761`: `composePromptArg` starts with `let promptArg = input.prompt` and only ever appends, so the title stays at offset 0. The repo's own existing understanding agrees — `ui/src/lib/slash.ts:41-44` and `ui/src/lib/slash.ts:77` both state that Claude only runs a slash command when it leads the prompt, which is why `applyCommandPick` hoists the picked command to the front.

**Coverage — no third site left bare.** Grepping every `create()` call with an `issueRef` (`src/drain.ts:2494`, `src/server.ts:1144`, `src/validate.ts:567`/`605`, `src/service.ts:3268`) accounts for all of them: drain (`src/drain.ts:2395`) and Up Next (`src/server.ts:1123`) are both fixed here; the New Task launch payload is client-composed and already templated at `ui/src/lib/components/NewTask.svelte:146`; `relaunch` reuses the stored prompt (`src/service.ts:3183`). `POST /api/up-next/start` is the only Up Next start handler (`src/server.ts:960-1106`), and the usage-hold branch (`src/server.ts:1146-1151`) stores the already-templated `input`, so a held task releases with the guard intact.

**The "unconditional across providers" rationale.** Verified, not assumed. `relaunch` resolves `effectiveProvider` from `overrides?.agentProvider ?? s.agentProvider` at `src/service.ts:3169` while carrying `s.prompt` at `src/service.ts:3183`, so a provider switch on relaunch is real and a Codex-only skip would indeed be a latent bug. The claim that Codex invokes skills as `$name` is correct — `src/commands.ts:245` emits `invocations: { codex: \`$${name}\` }` against `src/commands.ts:115`'s `{ claude: \`/${name}\` }`.

**The "don't sanitize in composePromptArg" rationale.** Correct. An operator-authored leading `/` is a supported first message: `src/commands.ts:115` and `ui/src/lib/slash.ts:12` both produce `/name` as the prompt for the Commands tab / inline picker, which flows into the same `create()` seam. Neutralizing inside `composePromptArg` would break that feature.

**The byte-identical path is load-bearing, as the comments claim.** `src/service.ts:2946` calls `this.deps.namer(input.prompt)` and feeds the result to `uniqueName`, which resolves the herdr agent name, the `shepherd/<name>` branch, and the worktree path (`src/service.ts:3563-3582`). Since `issueSpawnPrompt` returns `title` unchanged for any non-slash title (`src/issue-spawn-prompt.ts:40`), existing naming is preserved. The new assertions pin this: `test/drain.test.ts:420` against the `issue()` helper's `title: \`issue ${number}\`` (`test/drain.test.ts:30`), and `test/server-up-next.test.ts:161` against the request's `title: "t"` (`test/server-up-next.test.ts:153`). The drain harness already records the full `CreateSessionInput` (`test/drain.test.ts:136,234`), so the new `prompt` assertions read a real value; the Up Next harness needed the field added and the diff adds it (`test/server-up-next.test.ts:63,82`).

**Edge cases.** `trimStart()` covers whitespace-prefixed titles (`" /foo"`), and since JS `trimStart` strips the full Unicode `White_Space` set plus `\uFEFF`, a BOM- or NBSP-prefixed title is also templated rather than slipping through. Empty title returns `""` unchanged, matching today's behavior. `number` is a `number` in the `IssueRef` contract (`src/validate.ts:238-247`, `src/up-next-core.ts:38`), so the template interpolation is safe.

**i18n parity.** No new message keys are introduced — the English wording is server-side spawn text. The deliberately-unshared client template exists in both catalogs already (`ui/messages/en.json:272` and `ui/messages/de.json:272`), so nothing regresses there.

## Stated limitations

- I could not verify the Claude Code CLI's actual argument parsing against a live binary — it is not in this repo. The `claude -p $'/zzznope…'` transcript in the doc comment is taken as written. The repo-internal evidence above (`ui/src/lib/slash.ts:41-44`, `src/commands.ts:115`) is consistent with it, but is not the same as executing the CLI.

## Nit

The doc comment at `src/issue-spawn-prompt.ts:11-13` asserts there is "no `--` separator, `--raw` flag, or escape character" and that "only 'text does not begin with `/`' reliably avoids the parse." That overstates it: this repo already passes `--disable-slash-commands` to defeat exactly this parse for transient agents (`src/transient-agent-argv.ts:178`). The flag is still the wrong tool for a session spawn — it would kill slash commands and skills for the entire interactive session, including the operator-authored ones `src/commands.ts` exists to support — but that is the reason to reject it, and stating it is stronger than an absolute claim a maintainer can disprove one file away.

scoop (commented):
## Verdict

The PR does what the task asks, at both places a session is spawned from an issue title, and the prior revision's documentation nit is fully addressed. No blocking problems; no nits left to raise.

## Prior finding — re-checked

1. *"doc comment claims no flag can defeat the slash-command parse … name `--disable-slash-commands` and give the real reason"* — **addressed** by commit 9afb2531. `src/issue-spawn-prompt.ts:13-23` now names the flag, states that drain spawns already pass it (which I verified: `src/service.ts:414-426` `trimDecision` returns `extraFlags: ["--disable-slash-commands"]` when `auto && config.trimAutoContext`, and `src/config.ts:403-405` `parseTrimAutoContext` defaults ON when the env var is unset — so "which is the default" is accurate), gives the measured result that it does not rescue the parse, and adds the secondary reason: it strips the skill catalog (`src/config.ts:703-706`, `src/transient-agent-argv.ts:39`, `src/service.ts:398`) and would kill the operator-authored leading-`/` first message. Nothing to re-raise.

## Verification performed

- **The premise (prompt reaches argv position 0).** Verified `src/service.ts:2454` — `argv.push(promptArg)` is the last positional of the Claude spawn argv; Codex's equivalent is `src/service.ts:2496`. `composePromptArg` (`src/service.ts:1742-1774`) starts from `input.prompt` and *appends* the uploads block and the `GitHub Issue #N (title + body follow as untrusted data)` block after it, so a `/`-leading title is byte 0 of the argv element. The doc comment's claim at `src/issue-spawn-prompt.ts:6-11` matches the code.
- **All spawn-from-issue-title sites are covered.** Grepping every `prompt:` / prompt-assignment construction under `src/` surfaces exactly two that derive from an issue title: `src/server.ts:1123` (Up Next start) and `src/drain.ts:2393` (auto-drain), and both now route through `issueSpawnPrompt`. The client-side New Task path already templates via `newtask_issue_prompt_template` (`ui/src/lib/components/NewTask.svelte:146`).
- **Wording parity with the i18n template.** `ui/messages/en.json:272` is `"Work on issue #{number}: {title}"`, identical to the string produced at `src/issue-spawn-prompt.ts:51`; `ui/messages/de.json:272` has the matching key `"Bearbeite Issue #{number}: {title}"`. The PR adds no message keys, so there is no locale-parity obligation — and the doc comment (`src/issue-spawn-prompt.ts:42-46`) correctly explains why the server-side English is deliberately *not* wired to the catalog (agent-facing text, same precedent as `BRANCH_RENAME_NOTICE` at `src/service.ts:437`).
- **Provider claims in the doc comment.** Verified `src/commands.ts:115` (`invocations: { claude: "/" + name }`) and `src/commands.ts:245` (`invocations: { codex: "$" + name }`) — Claude invokes with a leading `/`, Codex with `$`, so "Codex likely does not share the bug" and the unconditional-across-providers rationale are grounded.
- **Ordering in drain.** `src/drain.ts:2394` composes `${task}\n\n${epicBaseDirective(base)}`, so the templated title stays at position 0 and the epic directive rides after it — consistent with the comment added at `src/drain.ts:2392-2393`.
- **Edge inputs.** `""` → `""` (`"".trimStart().startsWith("/")` is false), `"/"` → templated, `"  /foo"` → templated with the title preserved verbatim including its leading spaces, `"Rework the /usage probe"` → passthrough. The `trimStart()` choice is the right one: it does not depend on undocumented CLI whitespace trimming. Tests at `test/issue-spawn-prompt.test.ts:4-31` cover each of these, and `test/drain.test.ts:415-425` asserts the byte-identical passthrough that the namer depends on (`h.creates[0].prompt === "issue 1"` matches the harness default title at `test/drain.test.ts:30`).
- **Test harness change.** `test/server-up-next.test.ts:63,82` add `prompt` to the recorded create call — a test-local record type, no production consumer to break.

- A title beginning with `-` or `--` (e.g. `--version broken in 2.1`) lands in the same argv position (`src/service.ts:2454`) and would be parsed as a CLI flag rather than prompt text; `issueSpawnPrompt` addresses only the `/` case, and no `--` separator is passed. Pre-existing behavior, not caused by this diff.

## Stated limitations

- The two transcript snippets in the doc comment (`claude -p $'/zzznope…'` with and without `--disable-slash-commands`) are empirical measurements against the Claude Code CLI, which is not in this repo. I could not re-run them (read-only review) and did not verify the CLI's parser behavior directly; I verified only that the surrounding repo-side claims they rest on are true. Assumed as written.

PR #1808, opened 2026-07-16
feat(herd): make lifecycle groups collapsible on the desktop rail

kai-osthoff:
## What

The lifecycle group headers in the session rail (CI RUNNING, YOUR TURN, MERGED, …) were only collapsible in the phone layout's accordion (v1.44 mobile-herd-accordion). This makes them collapsible on the desktop rail too — **independently per group** (no accordion), all expanded by default, state held in-memory for the running session (mirrors the epic-group collapse).

## How

- **`herd-partition.ts`** — `GROUP_KEY_BY_STAGE` is now the single source for the group keys (`partitionGroups`, `groupHelp`, the mobile accordion default, keynav, and the collapse state all read it); `flattenByStage` gained an optional collapsed-stage skip.
- **`HerdGroup.svelte`** — the existing disclosure is reused; all three 44px touch-target rule sets now key off input modality (`touchTarget = flow || coarse pointer`) instead of layout, so touch-wide desktops (tablets/foldables) keep ≥44px targets while fine-pointer desktops stay compact. Toggle hover uses the canonical `--color-hover` fill.
- **`Herd.svelte`** — desktop disclosure renders only when the page wires `onstagecollapsetoggle` (never a no-op button); `flow` keeps the phone accordion untouched.
- **`herd-keynav.ts`** — `railOrder` now mirrors the template's full grouping pipeline (experiments first, epics over the experiment rest, lifecycle over the true rest) and skips collapsed stages — only their rest rows, so epic/experiment members stay reachable. New `railLocationsOf` locator replaces the page's `epicGroupOf`, which could attribute an experiment member to an epic group it doesn't render in.
- **`herd-jump.ts` (new)** — single reveal-before-select authority: every global jump (command bar, queue strip, rundown, deep link, herdr-update jump, viewport navigation, repo-filter retarget, next-needs-you) runs path effects first, resolves the target's location **after** them (an initially filter-hidden target becomes locatable), expands the right collapsed set (stage only on desktop — mobile jumps never mutate the desktop state), awaits a tick, then selects. All mutable deps are live getters. `nextNeedsYouTarget` is gone; `nextNeedsYou` picks the target, the core owns expansion.

## Testing

- `cd ui && bun run check` — 0 errors; full suite `bun run test` — **3944 passed** (incl. 15 new herd-jump ordering/per-path/liveness tests, keynav experiment-mirroring + locator tests, partition mapping/skip tests, and desktop-collapse browser tests: default-expanded, independence, controlled props, no-op-button guard, modality height assertions, parametrized header-action tests).
- Gates: `check:i18n`, `check-glossary.mjs`, `check-feature-catalog.sh`, `check-announcement-versions.mjs`, branch hygiene — all green (announcement stamped 1.45.0 after the 1.44.0 release landed mid-branch).
- **Live desktop smoke (Playwright against the real instance via the worktree dev server): 12/12** — toggles render expanded by default, collapsing hides exactly that group's rows (10→9) and flips `aria-expanded`, other groups stay open, phone-width switch shows the unchanged accordion (≤1 open, ≥44px targets), the desktop collapse state survives the mobile round-trip, fine-pointer headers stay compact (16.5px), re-expand restores rows. Header bulk actions (Merge train / Decommission all) were deliberately not clicked live; their non-toggling behavior is covered by the parametrized browser tests.

Closes nothing — direct user request ("das Aufklappen/Zuklappen geht aktuell nur auf kleinen Smartphones — lass es bitte auch auf dem Desktop zu").

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3234,5 +3234,7 @@
   "recap_failure_detail": "Fehler",
   "recap_failure_default_model": "Anbieter-Standard",
   "feat_recap_diagnostics_title": "Fehlende Recaps verstehen",
-  "feat_recap_diagnostics_body": "Abgeschlossene Sessions zeigen jetzt, wie sie beendet wurden und warum ein Recap fehlt. Sessions ohne Änderungen erhalten weiterhin ein Recap mit einem klaren No-Diff-Hinweis; technische Fehler enthalten eine Handlungsempfehlung und redigierte Details."
+  "feat_recap_diagnostics_body": "Abgeschlossene Sessions zeigen jetzt, wie sie beendet wurden und warum ein Recap fehlt. Sessions ohne Änderungen erhalten weiterhin ein Recap mit einem klaren No-Diff-Hinweis; technische Fehler enthalten eine Handlungsempfehlung und redigierte Details.",
+  "feat_desktop_herd_group_collapse_title": "Lebenszyklus-Gruppen auch am Desktop einklappen",
+  "feat_desktop_herd_group_collapse_body": "Die Gruppen-Header der Desktop-Session-Leiste (CI läuft, Dein Zug, Gemergt, …) sind jetzt ebenfalls klappbar: Ein Klick auf einen Header klappt die Gruppe zu — jede Gruppe unabhängig, und Sprünge zu einer Session in einer eingeklappten Gruppe öffnen sie automatisch wieder."
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -3234,5 +3234,7 @@
   "recap_failure_detail": "Failure",
   "recap_failure_default_model": "provider default",
   "feat_recap_diagnostics_title": "Understand missing recaps",
-  "feat_recap_diagnostics_body": "Finished sessions now show how they were completed and why a recap is missing. No-change sessions still get a recap with a clear no-diff warning; technical failures include actionable guidance and redacted details."
+  "feat_recap_diagnostics_body": "Finished sessions now show how they were completed and why a recap is missing. No-change sessions still get a recap with a clear no-diff warning; technical failures include actionable guidance and redacted details.",
+  "feat_desktop_herd_group_collapse_title": "Fold lifecycle groups on the desktop",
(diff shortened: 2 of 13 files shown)


kai-osthoff (commented):
No actionable findings from the read-only inspection.

Verified ground truth:

- Desktop collapse state is page-owned and reactive, toggled per lifecycle key, and passed into the desktop Herd together with its callback and touch-input signal (ui/src/routes/+page.svelte:502, ui/src/routes/+page.svelte:503, ui/src/routes/+page.svelte:2900).
- Named lifecycle groups become disclosure buttons only when a toggle callback exists; their controlled expanded state reads the same collapsed-key set, while the existing phone accordion remains the `flow` branch (ui/src/lib/components/Herd.svelte:727, ui/src/lib/components/Herd.svelte:730, ui/src/lib/components/Herd.svelte:731). The button exposes `aria-expanded`, and collapsed groups omit their session rows (ui/src/lib/components/herd/HerdGroup.svelte:79, ui/src/lib/components/herd/HerdGroup.svelte:107).
- Render, navigation, and jump lookup share the same complete stage-to-key mapping (ui/src/lib/components/herd-partition.ts:214); keyboard navigation removes collapsed lifecycle rows through that mapping (ui/src/lib/components/herd-keynav.ts:66), and page navigation passes the live collapsed set (ui/src/routes/+page.svelte:1438, ui/src/routes/+page.svelte:1458).
- Global jump handling resolves the target's actual render location, expands a collapsed desktop lifecycle or epic group, waits for the newly mounted row, then selects it (ui/src/lib/components/herd-jump.ts:55, ui/src/lib/components/herd-jump.ts:60, ui/src/lib/components/herd-jump.ts:66). The changed page call sites use those handlers for repo retargeting, push selection, Viewport navigation, and herdr-update jumps (ui/src/routes/+page.svelte:417, ui/src/routes/+page.svelte:1696, ui/src/routes/+page.svelte:2822, ui/src/routes/+page.svelte:3044).
- The added feature catalog fragment has the required shape and version 1.45.0, which is newer than the recorded 1.44.0 release (ui/src/lib/feature-announcements/entries/v1.45.0-desktop-herd-group-collapse.ts:3, .release-please-manifest.json:2); its title/body keys exist with parity in both maintained UI catalogs (ui/messages/en.json:3238, ui/messages/en.json:3239, ui/messages/de.json:3238, ui/messages/de.json:3239).
- The new hover styling uses an existing semantic token rather than a color literal (ui/src/lib/components/herd/HerdGroup.svelte:174, ui/src/app.css:13).

Tests, lint, and builds were not run, per the review instruction.

Issue #1821, opened 2026-07-17
Demo mode: per-session effects keep serving the autofocused hero session after switching


kai-osthoff:
Reproduced on main @ 1ae785a3 (pre-existing; unrelated to the status-bar PR that surfaced it):

1. `cd ui && bun run dev:demo`, open the app — the demo autofocuses the hero session (TASK-41 coupon-code-field).
2. Click another card, e.g. neon-postgres (TASK-45). The viewport header, prompt, and card highlight all switch to TASK-45.
3. Hover the session title: the metaPop's Usage row shows **186k tok — the coupon session's usage** (neon's seeded record is not served). The terminal also keeps replaying the coupon transcript.
4. Instrumenting `window.fetch` shows the app keeps polling `/api/sessions/coupon/usage` every 5s after the switch and **never requests `/api/sessions/neon/usage`** — while a direct in-page `fetch("/api/sessions/neon/usage")` returns neon's correct record. So Viewport's per-session `$effect`s did not re-key on the demo's session switch, while template-level derived values (desig, elapsed) did update.

The component itself is fine: a vitest browser test that rerenders Viewport from session A to B observes the B usage fetch (added in the status-bar PR as "refetches usage for the new session when the session prop switches"). The failure only reproduces in the demo world, so the suspect is the demo's store/session-object flow (fake events socket → `store.sessions` → `selected`) rather than the component.

Visible impact: with the session status bar shipped, the stale usage is now glanceable in the demo (previously it hid in the metaPop tooltip). Worth fixing so the public demo doesn't show the hero's token count on every session.

scoop:
✅ PR #2297 merged.

Issue #1815, opened 2026-07-17
Codex per-session token totals for the session status bar


kai-osthoff:
The unified session status bar (see the PR referencing this issue) shows token usage for Claude sessions (live JSONL + archive-time snapshots) but deliberately reports Codex sessions as unavailable ("token usage isn't tracked for Codex sessions yet"). The data exists in principle — the Codex state DB's `threads` table carries per-thread `tokens_used` / `updated_at_ms` / `rollout_path`, and `Session.providerSessionId` caches the session's rollout UUID — but four coupled problems must be solved before totals can ship correctly. These come out of five rounds of plan review and are this issue's **acceptance criteria**:

1. **Version-resilient overlap contract.** Whether a changed-id resume produces a thread row with *disjoint* token deltas or a *cumulative*, prefix-inheriting `tokens_used` is undocumented and may vary by Codex version. Blind `SUM` double-counts if cumulative; newest-row undercounts if disjoint. Required: a deterministic combine contract that is safe under both semantics (not a global mode inferred from one local empirical run), with inherited-prefix-overlap fixtures (e.g. parent 100k + cumulative 150k row → 150k, never 250k) and disjoint fixtures (100k + 50k → 150k).
2. **Retryable `providerSessionId` refresh across live resumes.** Codex writes the new rollout lazily, so any one-shot lookup at `herdr.start()` completion finds only the old rollout and retains the old id forever. Required: a marker-driven refresh (set only on a *successful* live resume, never clearing the known-good id on failed spawns — relaunch's deliberate clear-and-recapture stays as is) drained by the poller's existing exponential-backoff capture loop (`maybeCaptureCodexSessionId`).
3. **Bounded marker lifecycle for same-id resumes.** A resume that keeps the same thread id never yields a "different id" hit, so an unbounded retry would rescan `$CODEX_HOME/sessions` forever. Required: clear the marker once a post-resume append to the same thread is proven (e.g. `updated_at_ms` advanced past the resume timestamp) or bound it with a TTL; add a same-id resume test proving retries terminate.
4. **No synchronous rollout-tree discovery in the HTTP polling path.** The usage endpoint is polled every 5s per focused session; an unbounded `$CODEX_HOME/sessions` tree walk (or per-request state-DB scan) must never run inside it. Required: a shared background/event-fed index of session→thread ids (poller-driven, like the existing capture), with a test asserting endpoint reads perform no tree scan.

Also in scope once the above land:
[...]

PR #1822, opened 2026-07-17
feat(ui): unified session status bar across all session views

kai-osthoff:
## What

Every session "thing" now ends in the **same status pane**: `provider · model · effort  ·  tokens  ·  elapsed`.

- **All session viewport stages** — normal, rework, plan-gate, drain, critic-review — get a persistent strip below the terminal/tab body, on every tab, desktop and mobile. During **rework** the `ReviewInFlightBanner` keeps showing the *reviewer's* environment while the strip keeps the **task agent's own** identity — previously that identity only existed in the hover-only metaPop tooltip (the reported bug: "status bar for rework is missing model etc.").
- **Finished sessions** (Done lens): the same pane pins to the bottom of `DoneRecapPanel`; scroll ownership moved into `.dr-body` so header and pane stay in view while the recap body scrolls.
- Identity formatting reuses `environmentLabel` — the same formatter the review banner uses — so the two strips can never drift apart in wording.

## Provider-aware usage API

`GET /api/sessions/:id/usage` now returns an explicit DTO (`available`, `source: live|snapshot|codex|none`, raw `total`, nullable splits/messageCount/byModel) resolved through a ladder:

1. **Archived → archive-time `session_usage` snapshot** (new single-row store getter). Raw token fields are served; the snapshot's `byModel` holds *weighted units* and is deliberately **not** copied into the raw-token field (`byModel: null`).
2. Snapshot-less archive → documented fallback to the live source.
3. **Claude → live JSONL, spawn-account-aware**: `readSessionUsage(..., spawnAccountDir)` — swap/pool-account sessions previously resolved a nonexistent path and read as silent zeros.
4. **Codex → honestly unavailable** (`source: "none"`) with Codex-specific explanation copy in the pane ("token usage isn't tracked for Codex sessions yet"). Per-session Codex totals are deliberately out of scope; the accumulated design constraints (version-resilient thread-overlap contract, bounded resume-refresh of `providerSessionId`, background rollout index) are captured as acceptance criteria in #1815. A root test pins that the Codex branch performs no state-DB/rollout access from the 5s polling path.

`available: false` is reserved for genuinely absent data — a readable-but-empty transcript renders **"0 tok"**, never the placeholder. The elapsed segment uses a new minute-granularity `elapsedCoarse` so the shared 30s clock never reads as a frozen counter; archived sessions show the static `archivedAt ?? updatedAt` duration. The pane is a labelled `role="group"`, deliberately **not** an ARIA live region.

Also: Viewport's usage effect now keys on the value-level `unitId` (the file's own documented pattern) instead of the churning `session.id` prop read, with a new regression test asserting the usage refetch on session switch.

## Verification

- Root: `bun run lint` ✓; `bun test ./test` in a clean worktree → 7175 pass, 1 fail = the known node-pty native-compile environment failure (pre-existing, unrelated; local pre-push is blocked by the same, hence gates run manually + `--no-verify` push).
- UI: `bun run check` ✓ (0 errors), `bun run test` ✓ (3941 tests / 260 files), `bun run check:i18n` ✓ (EN/DE parity).
- Hygiene gates ✓: branch-hygiene, feature-catalog (entry `v1.45.0-session-status-bar`, drawer-only), announcement-versions, glossary, prettier.
- Live-verified in demo mode via Playwright: pane renders below the terminal / above the steer bar with `Claude Code · opus · … · 186k tok · …`, correct `role/aria-label`, no live region; Done-recap pane pinned under an overflowing body.

## Note: pre-existing demo staleness (#1821)

While live-verifying, the demo world showed the autofocused hero session's usage on *other* sessions. Reproduced identically on `main` (the metaPop shows the same wrong value there) — a pre-existing demo-environment issue, now simply more visible because the pane is always on. Filed as #1821 with full instrumentation notes; the new Viewport browser test proves the component itself re-fetches correctly on session switch.

Refs #1815, #1821.

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -587,9 +587,17 @@ const UNAVAILABLE_USAGE: SessionUsageDto = {
  *     overlap semantics across resumes are unproven and a rollout-tree scan has no place in the
  *     UI's 5s polling path); tracked by the Codex-totals follow-up issue. */
 async function sessionUsageDto(s: Session, store: SessionStore): Promise<SessionUsageDto> {
-  if (s.status === "archived") {
+  // Snapshot rung applies only to Claude sessions: the writer never snapshots Codex, so any
+  // row under a now-Codex session is by definition from a previous (pre-replacement) cycle.
+  if (s.status === "archived" && (s.agentProvider ?? "claude") === "claude") {
     const snap = store.getSessionUsage(s.id);
-    if (snap) {
+    // Provenance guard (restore → replace → re-archive): a replaced session keeps its id but
+    // gets a NEW claudeSessionId, and its re-archive may skip snapshotting (empty/missing
+    // transcript) — the leftover row would then report the previous agent's tokens as
+    // current. Serve only a snapshot from the same agent lineage; '' = legacy row from
+    // before the provenance column, tolerated (it predates restore-with-replace staleness
+    // being reachable for it in practice, and new snapshots always stamp the id).
+    if (snap && (!snap.claudeSessionId || snap.claudeSessionId === s.claudeSessionId)) {
       return {
         available: true,
         source: "snapshot",
--- a/src/server.ts
+++ b/src/server.ts
@@ -96,7 +96,7 @@ import { listBranches } from "./branches";
 import { computeDiff, toSessionDiff } from "./diff";
 import { buildDiffNotes } from "./diff-annotations";
 import { resolveDiffBase } from "./diff-base";
-import { sessionTokens, jsonlPathFor, type SessionUsageRollup } from "./usage";
+import { readSessionUsage, jsonlPathFor, type SessionUsageRollup } from "./usage";
 import { buildUsageBreakdown } from "./usage-breakdown";
 import { buildUsageTimeline } from "./usage-timeline";
 import { isApiKeyMode } from "./spawn-auth";
@@ -548,10 +548,78 @@ export interface AppDeps {
   ) => Promise<import("./prompt-recommend").RecommendResult>;
 }
 
-const sessionUsage = (s: Session) =>
-  s.claudeSessionId
-    ? sessionTokens(jsonlPathFor(s.worktreePath, s.claudeSessionId))
-    : sessionTokens("/nonexistent"); // pre-feature session → zeroed usage
+/** Wire DTO for GET /api/sessions/:id/usage (mirrored by ui/src/lib/types.ts SessionUsage).
+ *  `available: false` is reserved for genuinely absent data — a readable-but-empty transcript
+ *  is `available: true, total: 0`. Splits/messageCount/byModel are null when the resolved
+ *  source doesn't carry them; `byModel` is always RAW tokens per model, never weighted units. */
+export interface SessionUsageDto {
+  available: boolean;
+  source: "live" | "snapshot" | "codex" | "none";
+  total: number;
+  input: number | null;
+  output: number | null;
+  cacheRead: number | null;
+  cacheWrite: number | null;
+  messageCount: number | null;
+  byModel: Record<string, number> | null;
+}
+
+const UNAVAILABLE_USAGE: SessionUsageDto = {
+  available: false,
+  source: "none",
+  total: 0,
+  input: null,
+  output: null,
+  cacheRead: null,
+  cacheWrite: null,
+  messageCount: null,
+  byModel: null,
+};
+
+/** Resolve a session's token usage through the source ladder:
+ *  1. archived → the archive-time session_usage snapshot (usage-snapshot.ts; #965 backfill);
+ *  2. snapshot-less archive → fall through to the live source (writer skips operational
+ *     archetypes / empty transcripts, and the backfill can miss);
+ *  3. Claude → live JSONL, spawn-account-aware (a swap/pool session's JSONL lives under
+ *     <spawnAccountDir>/projects — same rule as sessionDiffAnnotationsRead);
+ *  4. Codex → unavailable: per-session state-DB totals are deliberately NOT read here (thread
+ *     overlap semantics across resumes are unproven and a rollout-tree scan has no place in the
+ *     UI's 5s polling path); tracked by the Codex-totals follow-up issue. */
+async function sessionUsageDto(s: Session, store: SessionStore): Promise<SessionUsageDto> {
+  if (s.status === "archived") {
+    const snap = store.getSessionUsage(s.id);
+    if (snap) {
+      return {
+        available: true,
+        source: "snapshot",
+        total: snap.total,
+        input: snap.input,
+        output: snap.output,
+        cacheRead: snap.cacheRead,
+        cacheWrite: snap.cacheWrite,
+        messageCount: snap.messageCount,
+        // The snapshot's byModel holds WEIGHTED UNITS per model (SessionUsageSnapshot), which
+        // must never be served in this raw-token field.
+        byModel: null,
+      };
+    }
+  }
+  if ((s.agentProvider ?? "claude") === "codex") return UNAVAILABLE_USAGE;
+  if (!s.claudeSessionId) return UNAVAILABLE_USAGE; // pre-feature session: no pinned id
+  const u = await readSessionUsage(s.worktreePath, s.claudeSessionId, s.spawnAccountDir);
+  if (!u) return UNAVAILABLE_USAGE;
+  return {
+    available: true,
+    source: "live",
+    total: u.total,
+    input: u.input,
+    output: u.output,
+    cacheRead: u.cacheRead,
+    cacheWrite: u.cacheWrite,
+    messageCount: u.messageCount,
+    byModel: u.byModel,
+  };
+}
 
 const json = (body: unknown, status = 200) =>
   new Response(JSON.stringify(body), { status, headers: { "content-type": "application/json" } });
@@ -2111,7 +2179,7 @@ async function handleSessionCreate({ req, parts, deps }: Ctx): Promise<Response 
 
 async function sessionUsageRead(id: string, deps: AppDeps): Promise<Response> {
   const s = deps.store.get(id);
-  return s ? json(await sessionUsage(s)) : json({ error: "not found" }, 404);
+  return s ? json(await sessionUsageDto(s, deps.store)) : json({ error: "not found" }, 404);
 }
 
 async function sessionActivityRead(id: string, deps: AppDeps): Promise<Response> {
--- a/src/server.ts
+++ b/src/server.ts
@@ -548,11 +548,12 @@ export interface AppDeps {
   ) => Promise<import("./prompt-recommend").RecommendResult>;
 }
 
-/** Wire DTO for GET /api/sessions/:id/usage (mirrored by ui/src/lib/types.ts SessionUsage).
+/** Wire DTO for GET /api/sessions/:id/usage (mirrored by ui/src/lib/types.ts SessionUsage —
+ *  the two trees don't share a build, so the shape is deliberately not exported/imported).
  *  `available: false` is reserved for genuinely absent data — a readable-but-empty transcript
  *  is `available: true, total: 0`. Splits/messageCount/byModel are null when the resolved
  *  source doesn't carry them; `byModel` is always RAW tokens per model, never weighted units. */
-export interface SessionUsageDto {
+interface SessionUsageDto {
   available: boolean;
   source: "live" | "snapshot" | "codex" | "none";
   total: number;
--- a/src/store.ts
+++ b/src/store.ts
@@ -4990,6 +4990,15 @@ export class SessionStore implements CapStore, CreditStore, ModelWeekStore {
     }));
   }
 
+  /** One session's archive-time usage snapshot, or null when none was recorded
+   *  (writer skips operational archetypes / empty transcripts; backfill can miss). */
+  getSessionUsage(sessionId: string): SessionUsageSnapshot | null {
+    const row = this.db
+      .query(`SELECT * FROM session_usage WHERE sessionId = ?`)
+      .get(sessionId) as SessionUsageRow | null;
+    return row ? { ...row, byModel: JSON.parse(row.byModel) as Record<string, number> } : null;
+  }
+
   // ── per-session usage buckets ────────────────────────────────────────────────
 
   /** Replace all bucket rows for sessionId atomically.
--- a/src/store.ts
+++ b/src/store.ts
@@ -2348,6 +2348,11 @@ export class SessionStore implements CapStore, CreditStore, ModelWeekStore {
       now,
       id,
     ]);
+    // The archive-time usage snapshot is stale the moment the session is live again (a
+    // replace can even swap its provider/lineage). Drop it so the archived-usage read can't
+    // serve a pre-restore row — including a legacy blank-provenance one — as current; the
+    // next archive re-snapshots the cumulative transcript.
+    this.deleteSessionUsage(id);
   }
 
   // ── usage limit caps (CapStore) ──────────────────────────────────────────
@@ -5005,6 +5010,15 @@ export class SessionStore implements CapStore, CreditStore, ModelWeekStore {
       .query(`SELECT * FROM session_usage WHERE sessionId = ?`)
       .get(sessionId) as SessionUsageRow | null;
     return row ? { ...row, byModel: JSON.parse(row.byModel) as Record<string, number> } : null;
+  }
+
+  /** Drop a session's archive-time usage snapshot (+ its buckets via FK cascade). Called on
+   *  restore/unarchive: the snapshot is archive-final, so once the session is live again it is
+   *  provisional — the next archive re-snapshots the (cumulative) transcript. This closes the
+   *  restore → replace → re-archive-with-skipped-snapshot staleness: a legacy blank-provenance
+   *  row can't survive a restore to be served as the replacement's usage. Idempotent. */
+  deleteSessionUsage(sessionId: string): void {
+    this.db.run(`DELETE FROM session_usage WHERE sessionId = ?`, [sessionId]);
   }
 
   // ── per-session usage buckets ────────────────────────────────────────────────
--- a/src/store.ts
+++ b/src/store.ts
@@ -1271,6 +1271,7 @@ export class SessionStore implements CapStore, CreditStore, ModelWeekStore {
       sessionId      TEXT PRIMARY KEY,
       desig          TEXT NOT NULL,
       name           TEXT NOT NULL DEFAULT '',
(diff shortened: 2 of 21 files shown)


kai-osthoff:
## Blocking finding

- `SessionStatusBar` uses nullish fallback for both model and effort, so an explicit current value of `null` is replaced by the original launch metadata (`ui/src/lib/components/SessionStatusBar.svelte:13`, `ui/src/lib/components/SessionStatusBar.svelte:18`). That is not a legacy-only case: `null` is the defined representation of provider-default model/effort (`src/types.ts:41`), the Continue-with picker deliberately submits `null` for both default choices (`ui/src/lib/components/new-task/ModelCliPicker.svelte:88`), and replacement persists those values on the existing row without replacing `launchMetadata` (`src/service.ts:3324`, `src/service.ts:3331`, `src/store.ts:2116`). Therefore, replacing a session originally launched as Opus/High with provider defaults leaves the new bar permanently saying Opus/High in both live and Done views. Preserve explicit `null` and fall back only for genuinely absent legacy fields; add a replacement-to-default regression case.

## Verification

- The shared component is mounted in both the live viewport and archived Done panel (`ui/src/lib/components/Viewport.svelte:3071`, `ui/src/lib/components/DoneRecapPanel.svelte:254`).
- The changed usage wire shape agrees between server and client, including nullable splits and `byModel` (`src/server.ts:556`, `ui/src/lib/types.ts:1144`), and the route returns that DTO through the new resolver (`src/server.ts:2177`).
- All seven new status/announcement message keys exist in both maintained UI catalogs (`ui/messages/en.json:3240`, `ui/messages/de.json:3240`). The announcement uses version 1.45.0 (`ui/src/lib/feature-announcements/entries/v1.45.0-session-status-bar.ts:5`), which is newer than the recorded 1.44.0 release (`.release-please-manifest.json:2`).
- Per instruction, I did not build, test, lint, or execute project code.

kai-osthoff:
## Re-review

The previous finding is addressed. The pane passes `session.model` directly and only falls back for effort when the field is truly `undefined`, so an explicit `null` reaches the shared formatter as provider-default rather than resurrecting launch metadata (`ui/src/lib/components/SessionStatusBar.svelte:10`, `ui/src/lib/components/SessionStatusBar.svelte:19`). The regression test sets both row fields to `null` while leaving stale `opus`/`high` launch metadata and expects the default/default label (`ui/src/lib/components/SessionStatusBar.browser.test.ts:102`, `ui/src/lib/components/SessionStatusBar.browser.test.ts:141`); that agrees with the server contract that `null` means provider default (`src/types.ts:41`).

## Blocking

- The new strip describes `session.model` and `session.effort` as the environment the agent runs with (`ui/src/lib/components/SessionStatusBar.svelte:19`, `ui/messages/en.json:3246`), but those row fields are stored intent, not necessarily the spawned values. Claude's spawn path explicitly substitutes a usage-downgrade or availability fallback only in argv and never rewrites the stored model (`src/service.ts:2290`); the existing test proves a `fable` row can actually spawn `opus[1m]` (`test/service.test.ts:7505`, `test/service.test.ts:7520`). Codex likewise clamps `max` effort to `high` while retaining `max` on the row (`test/service.test.ts:402`, `test/service.test.ts:417`). These are reachable supported paths, so the status pane can confidently display the wrong model and effort.

## Other verification

- The same pane is mounted outside the tab-specific viewport body and in the archived Done panel (`ui/src/lib/components/Viewport.svelte:3066`, `ui/src/lib/components/DoneRecapPanel.svelte:252`); the rework regression asserts the task agent's own identity beside the REWORK strip (`ui/src/lib/components/Viewport.browser.test.ts:855`).
- The server usage DTO and client mirror expose the same availability/source/total/split/message/by-model fields (`src/server.ts:551`, `ui/src/lib/types.ts:1139`), and the archive snapshot's weighted `byModel` contract justifies withholding it from the raw-token response (`src/types.ts:1072`, `src/server.ts:602`).
- Every new status-bar and announcement message key exists in both maintained UI catalogs (`ui/messages/en.json:3240`, `ui/messages/de.json:3240`). The feature entry uses `1.45.0` consistently (`ui/src/lib/feature-announcements/entries/v1.45.0-session-status-bar.ts:4`), above the recorded `1.44.0` release (`.release-please-manifest.json:2`).

kai-osthoff:
Critic finding "render the actual spawned model/effort, not stored intent": both cited mechanisms verified real (pushModelFlag is documented argv-only; Codex clamps max→high while the store keeps the intent). I took the finding's second offered option — **explicitly label the pane as configured intent** — rather than persisting effective spawn values in this PR: the effective values would need writing at every spawn site (create, replace, relaunch, resume, drain/epic spawns), and a usage downgrade can differ between resumes of the same session, so a write-once field would itself go stale — that's a server subsystem deserving its own designed change, now tracked as #1823 with the concrete code sites. What shipped here: the identity segment's hover title says "Configured environment: … The runtime may substitute at spawn time (usage downgrade, provider clamps) without changing this configuration" (EN+DE), and two new status-bar tests cover exactly the flagged cases — a substitutable model (fable) and a clamped Codex effort tier (max) — asserting they render as labeled configuration.

kai-osthoff:
## Blocking findings

The previous model/effort finding is not fully addressed. The component still derives the displayed identity from `session.model` and `session.effort` (`ui/src/lib/components/SessionStatusBar.svelte:25`), while the new qualification is only a `title` on a non-focusable `<span>` (`ui/src/lib/components/SessionStatusBar.svelte:68`); the pane's accessible group label remains merely “Session status” (`ui/src/lib/components/SessionStatusBar.svelte:67`, `ui/messages/en.json:3240`). The added substituted-model and clamped-effort tests verify only that native title property (`ui/src/lib/components/SessionStatusBar.browser.test.ts:206`, `ui/src/lib/components/SessionStatusBar.browser.test.ts:220`). This does not explicitly label the pane in its visible content or accessible name, so touch users and ordinary at-a-glance use still see the configured values presented as the agent identity. The underlying mismatch is confirmed by the argv-only model substitution (`src/service.ts:2290`) and the stored-max/spawned-high Codex case (`test/service.test.ts:402`). The feature announcement also still says the strip shows the model/effort the agent “runs with,” not configured intent (`ui/messages/en.json:3247`; German makes the same claim at `ui/messages/de.json:3247`). I therefore re-raised the prior finding verbatim.

The archive usage ladder can serve a snapshot from an earlier archive cycle. `sessionUsageDto` returns any row found for an archived session before checking the current provider (`src/server.ts:590`, `src/server.ts:608`). Restore reuses the same session id (`src/service.ts:3944`) and unarchive clears only archive fields, not `session_usage` (`src/store.ts:2304`). After a subsequent Continue-with replacement, the row can become Codex and lose its Claude session id (`src/service.ts:3324`); on re-archive, snapshotting then skips sessions without a Claude id, missing files, or empty transcripts (`src/usage-snapshot.ts:48`, `src/usage-snapshot.ts:51`, `src/usage-snapshot.ts:60`). The old snapshot therefore survives and is reported as current instead of the Codex/unavailable result.

The elapsed field is wall-clock age, not runtime. Every non-archived status—including `idle` and `done` in the session status domain (`src/types.ts:5`)—uses `clock.current - createdAt` and receives the “Running for” title (`ui/src/lib/components/SessionStatusBar.svelte:37`, `ui/src/lib/components/SessionStatusBar.svelte:44`). Restore keeps the same row and does not reset `createdAt` (`src/service.ts:3983`, `src/store.ts:2304`), so time spent archived is counted as running time and later as “Total runtime.” Either track active intervals/reset timing or label this honestly as age since original creation.

## Verification notes

The shared component is mounted in both the live viewport and Done recap (`ui/src/lib/components/Viewport.svelte:3071`, `ui/src/lib/components/DoneRecapPanel.svelte:254`). Its environment formatter exists and handles provider/model/effort defaults (`ui/src/lib/reviewer-env.ts:18`), and the coarse elapsed helper exists with the used signature (`ui/src/lib/format.ts:97`). The usage endpoint's server DTO fields agree with the client mirror (`src/server.ts:552`, `ui/src/lib/types.ts:1144`), `getSessionUsage` resolves the expected route (`ui/src/lib/api.ts:813`), and live transcript reads honor `spawnAccountDir` (`src/usage.ts:261`). The new status-bar and feature keys are present in both UI catalogs (`ui/messages/en.json:3240`, `ui/messages/de.json:3240`); the announcement fragment references those keys (`ui/src/lib/feature-announcements/entries/v1.45.0-session-status-bar.ts:3`) and its 1.45.0 version is newer than the recorded 1.44.0 release (`.release-please-manifest.json:2`). Per instruction, I did not run builds or tests.

kai-osthoff:
Finding 1 (render effective model/effort or label as configured intent) re-appeared verbatim this round — it was addressed in the previous push (73564b15) via the finding's second offered option: the identity segment's hover title reads "Configured environment: … The runtime may substitute at spawn time (usage downgrade, provider clamps) without changing this configuration" (EN+DE), with tests covering exactly the flagged cases (substitutable fable model, clamped codex max effort rendering as labeled configuration). Persisting the effective spawn values remains tracked in #1823 — it needs writes at every spawn site and per-resume refresh (a usage downgrade can differ between resumes), which is its own server change. If the labeling is judged insufficient (e.g. a visible marker is wanted rather than hover/aria), please say so specifically. Findings 2 and 3 from this round are fixed in this push: the archived-usage read now provenance-guards snapshots (claudeSessionId stamped per row; codex/replaced lineages rejected, with restore→replace→re-archive tests), and the elapsed segment is relabeled as session age ("time since creation, not active runtime", EN+DE) with idle + restored coverage.

kai-osthoff:
## Findings

1. **Configured intent is still presented as session status.** The pane derives its visible identity directly from `session.model` and `session.effort` (`ui/src/lib/components/SessionStatusBar.svelte:23`) and renders that unqualified identity as text (`ui/src/lib/components/SessionStatusBar.svelte:71`). Only the non-focusable span's hover `title` says "Configured environment"; the pane's group label remains the generic "Session status" (`ui/src/lib/components/SessionStatusBar.svelte:70`). The divergence is real: Claude substitutes the model only in argv (`src/service.ts:2290`), and Codex emits `high` while retaining stored `max` (`test/service.test.ts:402`). The new UI tests likewise assert `fable`/`Max` as the visible text and check the qualifier only through `.title` (`ui/src/lib/components/SessionStatusBar.browser.test.ts:241`, `ui/src/lib/components/SessionStatusBar.browser.test.ts:255`). A tooltip-only caveat therefore does not explicitly label the pane for keyboard/touch presentation or its accessible group name. Re-raised verbatim below.

2. **The provenance fix leaves migrated rows stale-reachable.** Migration assigns every historical snapshot `claudeSessionId = ''` (`src/store.ts:1297`), and the read guard explicitly accepts that blank provenance regardless of the current Claude lineage (`src/server.ts:592`, `src/server.ts:600`). In-place replacement allocates and persists a fresh Claude id (`src/service.ts:3271`, `src/service.ts:3324`), so an existing archived session with a migrated snapshot can be restored, replaced with Claude, and re-archived without a usable new transcript; the blank old row is then served as current. The added test actually codifies unconditional legacy-row acceptance (`test/server.test.ts:343`) while the purported restore/replace tests directly patch the final store state rather than exercising that transition (`test/server.test.ts:305`, `test/server.test.ts:323`). The Codex short-circuit and newly stamped rows are protected (`src/server.ts:600`, `src/server.ts:616`), but the historical-data path is not. Re-raised verbatim below as required.

3. **The What's-New copy contradicts the implementation.** EN says the strip shows what the agent "runs with" and "how long it has been at it" (`ui/messages/en.json:3247`), and DE makes the same runtime claims (`ui/messages/de.json:3247`). The component intentionally shows configured values that spawn may replace (`ui/src/lib/components/SessionStatusBar.svelte:18`) and wall-clock session age rather than active runtime (`ui/src/lib/components/SessionStatusBar.svelte:34`). Update both announcements to describe configured intent and session age, or make the pane fulfill the advertised runtime semantics.

## Re-review confirmation

- Prior finding 3 is fixed: elapsed is computed as wall-clock age and both live/archive titles explicitly say "Session age" and "not active runtime" (`ui/src/lib/components/SessionStatusBar.svelte:34`, `ui/src/lib/components/SessionStatusBar.svelte:47`). Idle and restored-downtime cases are covered (`ui/src/lib/components/SessionStatusBar.browser.test.ts:192`, `ui/src/lib/components/SessionStatusBar.browser.test.ts:208`), with matching EN/DE keys (`ui/messages/en.json:3244`, `ui/messages/de.json:3244`).

## Contract audit

- The server DTO and UI mirror have the same fields/nullability (`src/server.ts:551`, `ui/src/lib/types.ts:1139`); the changed Viewport consumer guards every nullable split before formatting (`ui/src/lib/components/Viewport.svelte:772`). `readSessionUsage` exists with the three-argument spawn-account contract (`src/usage.ts:261`), and `getSessionUsage` resolves to the expected endpoint (`ui/src/lib/api.ts:813`).
- Status/announcement keys are present in both UI catalogs (`ui/messages/en.json:3240`, `ui/messages/de.json:3240`). The feature fragment's keys and version resolve through the catalog glob (`ui/src/lib/feature-announcements/entries/v1.45.0-session-status-bar.ts:1`, `ui/src/lib/feature-announcements.ts:47`), and `1.45.0` is newer than the recorded `1.44.0` release (`.release-please-manifest.json:2`).

Per instruction, no tests, builds, or lint commands were run.

kai-osthoff:
All three findings from the last review are addressed (branch rebased onto latest main first, so the earlier conflict is gone):

1. **Configured intent, accessibly labeled (not tooltip-only).** The identity segment now carries the "Configured environment: … the runtime may substitute at spawn" caveat in its *accessible name* (`aria-label`), and the group's accessible name (`statusbar_aria`, EN+DE) states the model/effort shown are the configured values. So keyboard, touch, and AT users get the caveat, not just mouse hover. Tests assert via `aria-label`, not `.title`. I kept the visible text as the plain identity rather than adding a permanent visible "(configured)" tag, since substitution is the rare case (usage downgrade / codex clamp) and a persistent visible qualifier would be noise on every session; effective-value display remains tracked in #1823. If you specifically want a *visible* marker over the accessible-name approach, say so and I'll add one.

2. **Blank-provenance rows no longer stale-reachable.** `store.unarchive()` now drops the archive-time usage snapshot — a restored session's snapshot is provisional, and the next archive re-snapshots the cumulative transcript. So a restore → replace → re-archive-with-skipped-snapshot can't serve a pre-restore row (blank legacy or stamped) as the replacement's usage. The new tests drive the **real** `unarchive` transition and assert the row is gone, rather than patching final state; plus a cross-lineage stamped-row rejection test. The read guard's blank-tolerance now only applies to never-restored legacy archives, which is safe.

3. **Announcement copy fixed.** EN+DE `feat_session_status_bar_body` now say the strip shows the "configured" environment and "session age", matching the implementation (no more "runs with" / "how long it has been at it").

Local gates green: root lint + the usage/store/server suites (incl. the new restore-transition tests), `ui bun run check` (0 errors), the three status-bar-mounting UI suites (103), `check:i18n` parity, prettier, and the fallow delta gate (exit 0).

kai-osthoff:
## Findings

1. The configured/effective identity issue remains. The visible identity is still only `{identity}`; the configured-intent caveat exists in `title` and `aria-label` on non-focusable spans/groups, so a sighted keyboard or touch user does not see or focus it (`ui/src/lib/components/SessionStatusBar.svelte:32`, `ui/src/lib/components/SessionStatusBar.svelte:73`, `ui/src/lib/components/SessionStatusBar.svelte:74`). That distinction is material: Claude can replace the requested model only in argv without updating the row (`src/service.ts:2296`, `src/service.ts:2305`, `src/service.ts:2315`), and Codex converts `xhigh`/`max` to `high` (`src/default-effort.ts:83`, `src/default-effort.ts:91`). The component tests confirm that the bare visible values remain the configured values, including `Max` for a clamped Codex spawn (`ui/src/lib/components/SessionStatusBar.browser.test.ts:241`, `ui/src/lib/components/SessionStatusBar.browser.test.ts:259`).

2. The stale-snapshot fix is incomplete for databases upgraded from the previous release. The new migration stamps every existing snapshot with blank provenance (`src/store.ts:1307`, `src/store.ts:1311`), while the endpoint explicitly trusts blank provenance as current (`src/server.ts:592`, `src/server.ts:600`). Replacement can assign a new Claude lineage (`src/service.ts:3277`, `src/service.ts:3332`), and snapshotting skips a missing id/file or empty transcript without overwriting the old row (`src/usage-snapshot.ts:47`, `src/usage-snapshot.ts:52`, `src/usage-snapshot.ts:61`). The new unarchive deletion protects future restores only (`src/store.ts:2345`, `src/store.ts:2355`); it cannot repair a restore → replace → empty rearchive that was already archived before upgrade. The Codex branch now correctly short-circuits before snapshots (`src/server.ts:590`, `src/server.ts:616`), but an already-stale blank-provenance Claude row is still served.

3. The future-restore protection deletes durable spend history. `unarchive` now deletes the parent snapshot (`src/store.ts:2351`, `src/store.ts:2355`), and foreign keys are enabled with `ON DELETE CASCADE`, so all hourly buckets disappear too (`src/store.ts:885`, `src/store.ts:1289`, `src/store.ts:1300`, `src/store.ts:5020`). Those rows are the data sources for the Spend breakdown and Timeline (`src/usage-breakdown.ts:392`, `src/usage-breakdown.ts:397`, `src/usage-timeline.ts:107`, `src/usage-timeline.ts:108`). If the restored task is replaced with Codex or a transcript-less Claude agent, snapshotting skips (`src/usage-snapshot.ts:47`, `src/usage-snapshot.ts:52`) and the one-shot backfill will not recover it (`src/usage-backfill.ts:20`, `src/usage-backfill.ts:37`). Preserve prior-lineage accounting while invalidating only the status-pane's current snapshot selection, and cover Spend/Timeline across restore → replace → rearchive.

## Re-review confirmations

- The prior announcement-copy issue is fixed: both EN and DE now say configured environment and session age (`ui/messages/en.json:3286`, `ui/messages/de.json:3286`). Locale keys are present in both catalogs (`ui/messages/en.json:3280`, `ui/messages/de.json:3280`).
- The DTO mirror and its direct consumers agree on nullable splits and `byModel`: server contract at `src/server.ts:556`, UI mirror at `ui/src/lib/types.ts:1154`, and the popover guards every nullable split at `ui/src/lib/components/Viewport.svelte:775`. `getSessionUsage` resolves at `ui/src/lib/api.ts:813`; the Done panel keys and stale-response-guards its fetch at `ui/src/lib/components/DoneRecapPanel.svelte:52`.
- The feature entry resolves through the catalog glob and uses the next version after 1.44.0 (`ui/src/lib/feature-announcements/entries/v1.45.0-session-status-bar.ts:1`, `ui/src/lib/feature-announcements.ts:47`, `.release-please-manifest.json:2`).

Per instruction, no builds or tests were run; this was read-only inspection apart from writing this review artifact.

PR #1827, opened 2026-07-17
fix(recap): recover Codex verdicts delivered as chat instead of a file

kai-osthoff:
## What & why

A recap (and every other Codex helper role) was **failing even when the agent did the work correctly**.

Investigating **TASK-737**, its recap spawn (`codex exec`, gpt-5.6-luna) produced a complete, correct German recap in 9 seconds — and returned it as its **final chat message**, never calling a write tool. The role service only ever reads the result file (`.shepherd-recap.json`), so it saw an absent file, waited out the 60s startup grace, and finalized the row `failed` / `no-result`. The verdict existed and was correct; Shepherd discarded it.

The proof is in Codex's own session log — `~/.codex/sessions/…/rollout-2026-07-16T23-25-53-….jsonl`, whose `cwd` matches the recap tmpdir exactly:

```
session_meta → task_started → user_message → reasoning → agent_message → task_complete
```

Zero `function_call` / `local_shell_call` records. The prompt says *"Write the result as JSON to `.shepherd-recap.json`"*; Codex sometimes reads that as *"respond with the result"*. A second recent failure (`6b24a73a`) has the identical signature.

This was undiagnosable from `shepherd.log`: `logUnproducedVerdict` only captures a pane tail while the pane exists, and `codex exec` exits the instant it answers → `pane=absent` in 87/89 logged failures → the detail collapses to a generic *"Agent produced no verdict after 74 seconds."* (The recurring 74s is a tick artifact: ticks run every 15s, grace is 60s.)

## The fix

`codex exec` supports `-o, --output-last-message <FILE>` — **the CLI itself** writes the agent's final message to a file, independent of the model choosing to call a tool. We now:

1. Emit `-o .shepherd-last-message.txt` for every Codex role (`codexRoleArgv`). The relative path resolves against the spawn's cwd, so no cwd threading is needed and the Claude argv is byte-identical.
2. Read the result file **first**, falling back to the last-message file only when the result file is absent, via one shared seam (`readRoleResultText` in `src/codex-last-message.ts`).

In the normal success case both files exist (agent writes the result mid-run, CLI writes an exit ack) — the result file wins, so the ~95% that already work are untouched. A prose last-message fails the existing shape validation and **fails closed** exactly as before; the fallback can never invent a verdict.

## Scope

Applied to all nine JSON-verdict Codex roles through the single shared seam: **recap, PR critic, standalone-critic, plan-reviewer, autopilot, rundown, distiller, optimizer, merge-suggest**. Out of scope by design: `namer` (plain-text slug, no validator — prose would slugify to garbage; already degrades to `slugifyManual`), `doc-agent` (deliverable is file edits + a sentinel, not a message), `verify-key` (Claude-only).

## ⚠️ Codex version floor

Requires a Codex CLI that supports `-o` / `--output-last-message`. Verified on the installed **codex-cli 0.144.5**. If a Codex build lacks the flag, every Codex role spawn would die on an unknown flag — a conscious dependency to keep in mind on CLI downgrades.

## Verification

- **Live end-to-end**: built the argv via the real `codexRoleArgv`, ran codex on a chat-only prompt (TASK-737's exact mode) — codex wrote **only** `.shepherd-last-message.txt`, and the real `defaultReadVerdict` recovered `verdict: ready`.
- **Full root suite**: 7177 pass; the single failure is the pre-existing `pty-attach` node-pty native-build issue in worktrees, unrelated to this change.
- New tests: the shared helper's four contract cases (`test/codex-last-message.test.ts`) + recap read-path recovery, primary-file precedence, and prose-fails-closed (`test/recap-core.test.ts`). Updated the exact-argv assertions across the affected role test files.
- `bun run lint` + `prettier --check` clean.

[no-feature-entry] — server-side reliability fix, no user-facing UX.

## Security hardening (review round 2)

The provider-agnostic verdict read finalizes a strict-JSON parse on the first tick, so a PR that commits `.shepherd-review.json` or the `.shepherd-last-message.txt` fallback into its branch could short-circuit the real critic (a Claude critic included). The two PR critics run in a checkout of the untrusted PR head, so `scrubStaleVerdictArtifacts` now deletes both artifacts from the reviewer worktree immediately before spawn (after `rebaseSkip`, which can re-materialize a committed file). Applied to the plan reviewer too as defense-in-depth (it detaches at the trusted base). Recap and the other tmpdir-based roles have no checkout and are unaffected. See the author-note comment for the full scope analysis and the option chosen.

### DIFF
--- a/src/autopilot-llm.ts
+++ b/src/autopilot-llm.ts
@@ -74,7 +74,14 @@ function classifierArgv(
   prompt: string,
   effort?: string | null,
 ): string[] {
-  return buildTransientAgentArgv("writer-only", { provider, model, effort, prompt }).argv;
+  // The autopilot classifier READS the `-o` last-message fallback → opt in.
+  return buildTransientAgentArgv("writer-only", {
+    provider,
+    model,
+    effort,
+    prompt,
+    captureLastMessage: true,
+  }).argv;
 }
 
 interface PollClock {
--- a/src/autopilot-llm.ts
+++ b/src/autopilot-llm.ts
@@ -1,5 +1,5 @@
 import { mkdtempSync, rmSync } from "node:fs";
-import { readRoleResultText } from "./codex-last-message";
+import { readRoleResultText, CODEX_LAST_MESSAGE_FILE } from "./codex-last-message";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { HerdrDriver } from "./herdr";
@@ -47,7 +47,8 @@ function defaultMakeTmpDir(): string {
 function defaultReadVerdict(cwd: string): RawVerdict | null {
   // Result file first, Codex `-o` last-message fallback when absent (a Codex classifier that answers
   // in chat never writes the result file — see codex-last-message.ts).
-  const text = readRoleResultText(cwd, VERDICT_FILE);
+  // Disposable-tmpdir role → fixed fallback name (fresh empty cwd, no pre-seed risk).
+  const text = readRoleResultText(cwd, VERDICT_FILE, CODEX_LAST_MESSAGE_FILE);
   if (text === null) return null;
   try {
     return JSON.parse(text) as RawVerdict;
--- a/src/autopilot-llm.ts
+++ b/src/autopilot-llm.ts
@@ -1,4 +1,5 @@
-import { mkdtempSync, readFileSync, rmSync, existsSync } from "node:fs";
+import { mkdtempSync, rmSync } from "node:fs";
+import { readRoleResultText } from "./codex-last-message";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { HerdrDriver } from "./herdr";
@@ -44,10 +45,12 @@ function defaultMakeTmpDir(): string {
   return mkdtempSync(join(tmpdir(), "shepherd-autopilot-"));
 }
 function defaultReadVerdict(cwd: string): RawVerdict | null {
-  const p = join(cwd, VERDICT_FILE);
-  if (!existsSync(p)) return null;
+  // Result file first, Codex `-o` last-message fallback when absent (a Codex classifier that answers
+  // in chat never writes the result file — see codex-last-message.ts).
+  const text = readRoleResultText(cwd, VERDICT_FILE);
+  if (text === null) return null;
   try {
-    return JSON.parse(readFileSync(p, "utf8")) as RawVerdict;
+    return JSON.parse(text) as RawVerdict;
   } catch {
     return null; // partial write; try again next poll
   }
--- a/src/codex-last-message.ts
+++ b/src/codex-last-message.ts
@@ -35,8 +35,22 @@ export const CODEX_LAST_MESSAGE_FILE = ".shepherd-last-message.txt";
  * last-message file when the result file is absent. Returns null when neither exists (or is
  * unreadable mid-write — treat as not-yet-written and retry next tick). Callers apply their existing
  * parser/validator to the returned text, so behavior is unchanged whenever the result file is present.
+ *
+ * `codexFallback` gates the `-o` fallback and defaults to `true` (the disposable-tmpdir roles — recap,
+ * autopilot, rundown, distiller, optimizer, merge-suggest — run in a fresh empty dir no one else can
+ * write, so the fixed-name fallback is safe there). A role that runs in an UNTRUSTED checkout (the PR
+ * critics) MUST pass `codexFallback: <provider is Codex>`: the `-o` file is a Codex-only artifact, so
+ * a non-Codex reviewer has no legitimate last-message file and must NEVER read one — otherwise a PR
+ * that commits a strict-JSON `.shepherd-last-message.txt` into its head could short-circuit even a
+ * Claude reviewer (the read + parse are provider-agnostic). Paired with scrubStaleVerdictArtifacts,
+ * this "only enables the fallback for the matching completed Codex run".
  */
-export function readRoleResultText(cwd: string, resultFile: string): string | null {
+export function readRoleResultText(
+  cwd: string,
+  resultFile: string,
+  opts: { codexFallback?: boolean } = {},
+): string | null {
+  const { codexFallback = true } = opts;
   const primary = join(cwd, resultFile);
   if (existsSync(primary)) {
     try {
@@ -45,6 +59,7 @@ export function readRoleResultText(cwd: string, resultFile: string): string | nu
       return null; // unreadable mid-write — retry next tick
     }
   }
+  if (!codexFallback) return null; // non-Codex reviewer: no legitimate `-o` file — never read one
   const fallback = join(cwd, CODEX_LAST_MESSAGE_FILE);
   if (existsSync(fallback)) {
     try {
--- a/src/codex-last-message.ts
+++ b/src/codex-last-message.ts
@@ -1,0 +1,57 @@
+/**
+ * The Codex `-o` last-message fallback — one home for the file name and the shared read helper.
+ *
+ * A headless `codex exec` role is told to WRITE its result as JSON to a role-specific file
+ * (`.shepherd-recap.json`, `.shepherd-review.json`, …). Codex occasionally treats "write the
+ * result" as "respond with the result": it produces a complete, correct verdict as its final chat
+ * message and never calls a write tool (observed live — TASK-737's recap, and `6b24a73a`). The role
+ * service only reads the result file, so a correct verdict was silently discarded and the row failed
+ * with `no-result`.
+ *
+ * `codex exec -o <FILE>` makes the CLI ITSELF write the agent's final message to a file, with no
+ * dependence on the model choosing to call a tool — so it captures exactly that chat-only case. The
+ * flag is emitted for every Codex role (see codex-role-argv.ts) with a RELATIVE path, which Codex
+ * resolves against the spawn's cwd (verified: the file lands in the disposable tmpdir).
+ *
+ * The result file stays the PRIMARY contract; the last-message file is a FALLBACK read only when the
+ * result file is absent. In the normal success case BOTH files exist — the agent writes the result
+ * file mid-run and the CLI writes the last-message file (usually a "Created …" acknowledgement) at
+ * exit — so preferring the result file is essential, and there is no race: `-o` is written at exit,
+ * so a present last-message file means the agent has finished.
+ *
+ * The helper returns raw TEXT; each caller keeps its own parse + shape validation. A prose
+ * last-message that isn't a valid verdict fails that validation and fails closed exactly as before —
+ * the fallback can never invent a verdict.
+ */
+import { existsSync, readFileSync } from "node:fs";
+import { join } from "node:path";
+
+/** The file `codex exec -o <FILE>` writes the agent's final message to. Relative → resolved against
+ *  the spawn's cwd. Shared across every Codex role's result-read path + emitted by codexRoleArgv. */
+export const CODEX_LAST_MESSAGE_FILE = ".shepherd-last-message.txt";
+
+/**
+ * Read a role's result text: the role's own `resultFile` first, falling back to the Codex `-o`
+ * last-message file when the result file is absent. Returns null when neither exists (or is
+ * unreadable mid-write — treat as not-yet-written and retry next tick). Callers apply their existing
+ * parser/validator to the returned text, so behavior is unchanged whenever the result file is present.
+ */
+export function readRoleResultText(cwd: string, resultFile: string): string | null {
+  const primary = join(cwd, resultFile);
+  if (existsSync(primary)) {
+    try {
+      return readFileSync(primary, "utf8");
+    } catch {
+      return null; // unreadable mid-write — retry next tick
+    }
+  }
+  const fallback = join(cwd, CODEX_LAST_MESSAGE_FILE);
+  if (existsSync(fallback)) {
+    try {
+      return readFileSync(fallback, "utf8");
+    } catch {
+      return null;
+    }
+  }
+  return null;
+}
--- a/src/codex-last-message.ts
+++ b/src/codex-last-message.ts
@@ -1,5 +1,5 @@
 /**
- * The Codex `-o` last-message fallback — one home for the file name and the shared read helper.
+ * The Codex `-o` last-message fallback — the filename helpers and the shared read helper.
  *
  * A headless `codex exec` role is told to WRITE its result as JSON to a role-specific file
  * (`.shepherd-recap.json`, `.shepherd-review.json`, …). Codex occasionally treats "write the
@@ -11,7 +11,20 @@
  * `codex exec -o <FILE>` makes the CLI ITSELF write the agent's final message to a file, with no
  * dependence on the model choosing to call a tool — so it captures exactly that chat-only case. The
  * flag is emitted for every Codex role (see codex-role-argv.ts) with a RELATIVE path, which Codex
- * resolves against the spawn's cwd (verified: the file lands in the disposable tmpdir).
+ * resolves against the spawn's cwd.
+ *
+ * ── UNTRUSTED-CHECKOUT SAFETY ────────────────────────────────────────────────────────────────────
+ * The PR critics (review.ts, standalone-critic.ts) run in a worktree checked out from the UNTRUSTED
+ * PR head. If the `-o` fallback had a fixed, guessable name, a malicious PR could commit a strict-JSON
+ * `.shepherd-last-message.txt` into its branch; the read finalizes a strict parse on the first tick
+ * and is provider-agnostic, so that pre-seed would short-circuit the real reviewer — a Claude reviewer
+ * included, which never writes an `-o` file at all. So this helper NEVER reads a hardcoded name: the
+ * caller passes the exact fallback filename, and reviewer-kind spawns use a PER-SPAWN UNGUESSABLE name
+ * keyed on the spawn's session id ({@link codexLastMessageFile}) — a name a PR author cannot know at
+ * authoring time, so a committed copy can never match the name the real run writes and reads. The
+ * disposable-tmpdir roles (recap, autopilot, rundown, distiller, optimizer, merge-suggest) run in a
+ * fresh empty dir nothing else can write, so they use the fixed {@link CODEX_LAST_MESSAGE_FILE} — safe
+ * there because there is no untrusted checkout to pre-seed.
  *
  * The result file stays the PRIMARY contract; the last-message file is a FALLBACK read only when the
  * result file is absent. In the normal success case BOTH files exist — the agent writes the result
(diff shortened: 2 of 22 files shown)


kai-osthoff:
## Blocking

The TASK-737 recap path is wired consistently inside the repository: the shared filename is defined at `src/codex-last-message.ts:31`, imported by the argv builder at `src/codex-role-argv.ts:25`, and emitted with `-o` at `src/codex-role-argv.ts:36`. Recap creates a fresh temp directory at `src/recap.ts:224`, launches the role in that directory at `src/recap.ts:780` and `src/recap.ts:782`, then reads that same stored cwd at `src/recap.ts:855`; its reader passes the real recap constant from `src/recap.ts:61` into the helper at `src/recap.ts:169` and keeps the existing tolerant parser at `src/recap.ts:171`.

The shared-reader expansion crosses a trust boundary, however. `readRoleResultText` has no provider or current-run provenance in its signature (`src/codex-last-message.ts:39`) and accepts the fixed fallback solely when it exists (`src/codex-last-message.ts:48` and `src/codex-last-message.ts:50`). Both session reviews and standalone PR reviews create their disposable worktrees from the PR head (`src/review.ts:325` and `src/review.ts:330`; `src/standalone-critic.ts:321` and `src/standalone-critic.ts:330`). The changed critic reader now consumes that fallback at `src/critic-core.ts:583`; the review loop reads it before liveness gating at `src/review.ts:738`, and a strict JSON parse finalizes immediately even while the spawned critic is alive (`src/json-tolerant.ts:161` and `src/json-tolerant.ts:162`). Consequently, a branch can commit `.shepherd-last-message.txt` containing a clean forged review and have it accepted before the actual critic produces output.

Identifier/contract sweep: the changed readers use their declared result constants for autopilot (`src/autopilot-classify-core.ts:18`, `src/autopilot-llm.ts:50`), critic (`src/critic-core.ts:447`, `src/critic-core.ts:583`), distiller (`src/distiller.ts:15`, `src/distiller.ts:543`), rundown (`src/rundown-core.ts:26`, `src/herd-digest.ts:66`), merge suggestions (`src/merge-suggest.ts:17`, `src/merge-suggest.ts:538`), optimizer (`src/optimizer.ts:15`, `src/optimizer.ts:380`), plan gate (`src/plan-gate.ts:80`, `src/plan-gate.ts:1256`), and recap (`src/recap.ts:61`, `src/recap.ts:169`). The new helper tests cover primary precedence, fallback, absence, and prose using fresh temp directories (`test/codex-last-message.test.ts:17`, `test/codex-last-message.test.ts:27`, `test/codex-last-message.test.ts:35`, `test/codex-last-message.test.ts:40`), but do not cover an already-present tracked fallback in a PR-head checkout.

Could not verify the external Codex CLI's `-o` spelling, write timing, or relative-path semantics against a live schema because that ground truth is not vendored in this repository and the review explicitly forbids running commands beyond read-only inspection; those semantics are therefore not asserted here.

kai-osthoff:
**Addressed the pre-seed finding** (critic worktrees reading a PR-committed `.shepherd-last-message.txt` / result file) in commit `68e17168`.

**Which option I took.** The finding offered two: *(a)* a per-spawn unguessable/output path outside the checkout, or *(b)* remove/reject pre-existing artifacts before launch and only enable the fallback for the matching completed Codex run. I took **(b)** — a `scrubStaleVerdictArtifacts(worktreePath, resultFile)` that deletes both the role's result file and the `.shepherd-last-message.txt` fallback from the reviewer worktree immediately before spawn. Rationale over (a): the `-o` filename is emitted by the pure, cwd-agnostic `codexRoleArgv`, so a per-spawn nonce path would have to be threaded through all nine role read paths (each with a different read signature) — much larger, riskier surface — whereas the scrub is three localized call sites and needs no read-path change. It also achieves "only the matching completed Codex run": after the scrub, the only artifact present at read time is one our own run wrote.

**Scope of the actual exposure** (from tracing where each role spawns):
- **Genuinely exploitable — fixed:** `review.ts` and `standalone-critic.ts` detach at the **untrusted PR head**, so a PR could `git add -f` a strict-JSON verdict (the read path finalizes a strict parse on tick 1, provider-agnostic → hits Claude critics too). Scrub placed **after `rebaseSkip`**, which can re-materialize a committed file, and right before `herdr.start`.
- **Defense-in-depth:** `plan-gate.ts` detaches at the **trusted base** sha (not the plan author's live worktree), so it isn't currently exploitable — scrubbed anyway so a future base change can't silently open the hole.
- **Not exposed:** recap (the original bug) and the other five file-result roles (autopilot, rundown, distiller, optimizer, merge-suggest) spawn in fresh disposable `mkdtemp` dirs with no checkout — nothing to pre-seed. A Claude spawn also never writes an `-o` file, so no provider gate is needed once the pre-seed is scrubbed.

**Bonus:** scrubbing the result file also closes the same pre-seed vector for the pre-existing `.shepherd-review.json` read (latent before this PR).

**Regression tests:** helper-level (pre-seeded result + fallback → scrub removes both → `readRoleResultText` returns null; touches only the two known names; best-effort no-throw when absent) and integration-level (`ReviewService.consider` scrubs a pre-seeded PR-head worktree before spawning the critic).

kai-osthoff:
## Blocking

The normal tracked/pre-existing case is now covered at both PR-head spawn sites: `ReviewService` scrubs immediately before `herdr.start` (`src/review.ts:437`) and the standalone critic does the same (`src/standalone-critic.ts:465`); both use the critic result filename exported at `src/critic-core.ts:447`. The added test also proves the ordinary successful-unlink case (`test/codex-last-message.test.ts:59`).

However, the security boundary still fails open when either unlink errors. `scrubStaleVerdictArtifacts` catches and ignores every `rmSync` failure (`src/codex-last-message.ts:73`), while `readRoleResultText` subsequently reads any surviving primary or fallback artifact solely because it exists (`src/codex-last-message.ts:39`). The critic then treats strict JSON as a parsed verdict (`src/critic-core.ts:583`), and strict parses finalize immediately (`src/json-tolerant.ts:161`). The catch comment's claim that validators fail closed is therefore contradicted by the actual path: a surviving pre-seeded strict verdict passes those validators. The absent-artifact test does not exercise an unlink failure because `{ force: true }` handles absence normally (`test/codex-last-message.test.ts:75`). Make scrub failure abort/reject the spawn, verify absence after deletion, or bind fallback acceptance to a matching completed Codex run.

Static inspection also confirmed the intended recap wiring: Codex argv emits `-o` (`src/codex-role-argv.ts:32`), recap reads the fallback through the shared helper (`src/recap.ts:166`), and the parsed value still passes through the recap shape validator (`src/recap-core.ts:49`). The raw-JSON recovery case is represented at `test/recap-core.test.ts:166`. Per the review constraint I did not execute that test.

Could not independently verify the external Codex CLI's `-o` timing/path semantics from repository ground truth; the repository-side assumption is documented at `src/codex-last-message.ts:11`.

kai-osthoff:
**Follow-up: added the second clause the finding asked for.** The prior push scrubbed pre-existing artifacts before launch (clause 1). This push (`3589e541`) adds clause 2 — *"only enable the fallback for the matching completed Codex run"* — as an explicit provider gate, so both halves of the finding's option are now in place.

**What changed.** `readRoleResultText` gains a `codexFallback` option. The PR critics (`review.ts`, `standalone-critic.ts`) and the plan reviewer now read with `codexFallback: provider === "codex"`. The `-o` last-message file is a Codex-only artifact, so a **non-Codex reviewer has no legitimate fallback file and never reads one** — closing the "including a Claude reviewer" vector at the read level, independent of the scrub. Threaded via the reviewer provider already recorded on each critic's in-flight entry (added the field to standalone-critic's `InFlight`).

**Why gate + scrub (not a per-spawn unguessable path).** The finding offered either option; I took *"remove/reject pre-existing artifacts before launch AND only enable the fallback for the matching completed Codex run"* in full. The unguessable-path alternative would key the `-o` name on a per-spawn id threaded through **all nine** role read paths — including the six disposable-tmpdir roles (recap, autopilot, rundown, distiller, optimizer, merge-suggest) that run in a fresh empty cwd nothing else can write and are **not** exposed. That would touch un-vulnerable code against surgical-change discipline; the gate+scrub is scoped to exactly the roles that read from an untrusted checkout.

**Result — the critic reads a fallback only when both hold:** (a) the reviewer is Codex [gate], and (b) no pre-committed copy exists [scrub]. The default (no provider) is safe-closed. The tmpdir roles keep `codexFallback: true` (unchanged) since they can't be pre-seeded.

**New regression tests:** `readRoleResultText` ignores a present `.shepherd-last-message.txt` under `codexFallback:false`; a Claude critic returns `absent` for a pre-seeded `-o` file while a Codex critic reads it; the no-provider default is safe-closed. (Plus the prior scrub tests: helper-level removal of a tracked pre-seed + a `ReviewService.consider` integration test.)

kai-osthoff:
## Verification

The revision does address the Claude half of the prior finding: the provider domain is exactly `claude`/`codex` (`src/types.ts:13`), the critic enables the fallback only for `codex` (`src/critic-core.ts:623`), and the session critic scrubs both artifacts immediately before spawning (`src/review.ts:459`). The added integration regression checks that both pre-seeded files are absent at spawn time (`test/review.test.ts:397`).

The remaining blocking issue is run/completion binding. `readRoleResultText` accepts the fixed fallback solely because it exists; its API receives only a provider-derived boolean, not a spawn identity or completion state (`src/codex-last-message.ts:48`). The session critic reads it before checking process liveness (`src/review.ts:773`), while a strict parse finalizes immediately regardless of that liveness result (`src/json-tolerant.ts:155`). The standalone critic has no completion gate and finalizes any parsed fallback immediately (`src/standalone-critic.ts:525`). Thus a fixed file recreated while the workspace-write Codex reviewer is still live can still short-circuit the review (`src/codex-role-argv.ts:32`). The new provider test actually seeds that fixed file and expects the Codex reader to accept it without any completed-run evidence (`test/critic-core.test.ts:56`).

The scrub is also best-effort: deletion failures are swallowed even though the comment claims an undeleted pre-seed fails closed (`src/codex-last-message.ts:88`); the helper test establishes that valid pre-seeded JSON is accepted when it remains (`test/codex-last-message.test.ts:75`). No regression covers a failed scrub or proves fallback acceptance is tied to the matching completed Codex run.

No tests or builds were run, per the review instructions.

kai-osthoff:
**Switched to the per-spawn unguessable path (the finding's first option).** The scrub (round 2) and provider gate (round 3) were call-site mitigations that neutralized a pre-seed *after* the fact; they left the helper reading a fixed `.shepherd-last-message.txt`, which is the exact line this finding keeps citing. Commit `8c933ed0` makes the helper safe **by construction** instead.

**What changed.** `readRoleResultText` no longer hardcodes any fallback name — it reads only the `lastMessageFile` the caller passes. The `reviewer` kind (the PR critics, which run in a checkout of the UNTRUSTED PR head) now emits `-o .shepherd-last-message-<sessionId>.txt`, keyed on the per-spawn `randomUUID()` that `buildTransientAgentArgv` already mints and returns. A PR author can't know that id at authoring time, so a committed `.shepherd-last-message.txt` (or any guess) never matches the name the real run writes and the critic reads. The read side reconstructs the name from the session id each critic already records (`criticSessionId` / `reviewerSessionId`).

This **self-gates**: a Claude reviewer writes no `-o` file, so its per-spawn name never exists → no fallback, no provider check needed. The round-3 provider gate is therefore removed. The disposable-tmpdir roles (recap, autopilot, rundown, distiller, optimizer, merge-suggest) keep the fixed name — their cwd is a fresh empty dir nothing can pre-seed.

**Scrub retained, narrowed to the RESULT file.** `.shepherd-review.json` has a fixed name dictated by the prompt contract (the agent is told to write exactly that path), so it can't be made unguessable and is still scrubbed before launch. The `-o` fallback no longer needs scrubbing.

**Verified live:** a reviewer-kind Codex spawn answered in chat, wrote its per-spawn `-o` file, and the critic read recovered THAT verdict while a `.shepherd-last-message.txt` pre-seeded in the same dir (attacker's `decision:"approve"`) was ignored. Regression tests cover: the helper reads no fallback without an explicit name; a fixed-name pre-seed is ignored when the per-spawn name is absent; the per-spawn name round-trips; the argv emits a per-spawn name for `reviewer` and the fixed name for other kinds. Full suite 7195 pass (only the pre-existing `pty-attach` node-pty worktree failure).

kai-osthoff:
## Re-review

The prior `src/codex-last-message.ts` finding is addressed. Reviewer argv now derives its `-o` filename from a newly minted UUID (`src/transient-agent-argv.ts:155`, `src/transient-agent-argv.ts:168`), and both critic services retain that same ID and pass it back to the read path (`src/review.ts:417`, `src/review.ts:484`, `src/review.ts:769`; `src/standalone-critic.ts:439`, `src/standalone-critic.ts:522`). The helper reads only the caller-supplied name (`src/codex-last-message.ts:67`), while the regression tests show that a fixed-name pre-seed is ignored and the matching per-spawn file is accepted (`test/codex-last-message.test.ts:47`, `test/critic-core.test.ts:56`). The TASK-737 recap path also emits `-o` in its fresh temp cwd and reads that fallback through the existing tolerant parser (`src/recap.ts:166`, `src/recap.ts:781`).

## Blocking finding

`src/transient-agent-argv.ts:162` incorrectly treats `reviewer` as the only kind that can run in an untrusted checkout and consequently gives `doc` the fixed `.shepherd-last-message.txt` target at `src/transient-agent-argv.ts:168`. The same file identifies `doc` input as untrusted (`src/transient-agent-argv.ts:110`), and the doc agent can check its worktree out at an open PR's head SHA before calling this exact `doc` argv builder (`src/doc-agent.ts:552`, `src/doc-agent.ts:584`, `src/doc-agent.ts:766`). Codex writes `-o` with `std::fs::write` in its [official exec implementation](https://github.com/openai/codex/blob/main/codex-rs/exec/src/event_processor.rs), so a PR can commit `.shepherd-last-message.txt` as a symlink to one of the in-scope docs; the CLI's final-message write follows that predictable path and replaces the target. Shepherd then stages the entire in-scope set and can commit and push the replaced document (`src/doc-agent.ts:837`, `src/doc-agent.ts:856`, `src/doc-agent.ts:954`, `src/doc-agent.ts:962`, `src/doc-agent.ts:1114`). Give `doc` a per-spawn checkout-safe output path too (or omit `-o` for roles that never consume it) and cover the pre-existing/symlink case.

Inspection only; no tests or builds were run, per instruction.

kai-osthoff:
**Round-4 finding addressed** in commit `076bf0c8` — the `doc` kind no longer receives an `-o` target at all.

**Confirmed the finding is valid and in-scope.** `doc-agent` (a) never reads the `-o` last-message fallback — its deliverable is file EDITS + a sentinel (grep for `readRoleResultText`/`last-message` in `src/doc-agent.ts` is empty), and (b) does run in an untrusted checkout: `doc-agent.ts:625` documents `baseRef` as "`resolved.baseRef` (fresh) or **the PR head sha (retarget)**", and the `doc` kind is `mcpIsolated` like `reviewer`. So a fixed `-o .shepherd-last-message.txt` in that worktree was a symlink-redirect surface exactly as described.

**Fix (the finding's own preferred remedy — "omit `-o` for roles that never consume it").** `codexRoleArgv` now takes `lastMessageFile: string | null` and skips `-o` entirely when null; `buildTransientAgentArgv` passes `null` for the `doc` kind. No consumer ⇒ no `-o` ⇒ no target for a committed symlink (or a pre-existing file) to redirect. The `reviewer` kind keeps its per-spawn unguessable name; the disposable-tmpdir roles (writer-ro/writer-only) keep the fixed name, safe in a fresh empty cwd.

**Why omit rather than give `doc` a per-spawn name:** a per-spawn name would still write a file `doc-agent` never reads — dead output. Omitting is strictly correct and removes the surface unconditionally, covering both the symlink and pre-existing-file cases.

**Verification:** new regression test asserts the codex `doc` argv contains no `-o` and no `.shepherd-last-message*` token; the per-kind argv test now covers reviewer (per-spawn `-o`), doc (no `-o`), and tmpdir kinds (fixed `-o`). Full suite 7305 pass — the only failure is the pre-existing `pty-attach` node-pty worktree build (unrelated; green in CI `verify`).

kai-osthoff:
## Re-review

The previous `doc` symlink finding is addressed. `buildTransientAgentArgv` now maps the Codex `doc` kind to `lastMessageFile = null` (`src/transient-agent-argv.ts:173`), `codexRoleArgv` emits `-o` only for a non-null value (`src/codex-role-argv.ts:42`), and the regression asserts that the resulting doc argv has neither `-o` nor a last-message token (`test/transient-agent-argv.test.ts:388`).

## Finding

The last-message decision is still made at the coarse transient-kind level: every Codex `writer-only` call receives the fixed `-o` path (`src/transient-agent-argv.ts:170`, `src/transient-agent-argv.ts:178`). That kind is also used by the namer (`src/namer-llm.ts:74`), whose reader only checks `.shepherd-name` (`src/namer-llm.ts:48`); the changed namer test confirms the otherwise-unused `-o .shepherd-last-message.txt` pair is now deliberately emitted (`test/namer-llm.test.ts:113`). This contradicts the approved role boundary that excludes the plain-text namer and makes the new Codex CLI flag/version contract apply to an out-of-scope consumer. Make capture a caller-level option (or split the kind), pass no output target for the namer, and update its argv regression.

The primary/fallback precedence itself is centralized as intended: the helper reads the role result first and only then the caller-named last-message file (`src/codex-last-message.ts:73`, `src/codex-last-message.ts:83`); the recap supplies the fixed scratch-dir name (`src/recap.ts:170`) while critics reconstruct the per-spawn name (`src/critic-core.ts:713`).

Could not independently verify Codex CLI 0.144.5's external `-o` behavior or version support from repository contents; assumed as stated in the approved plan. Per review instructions, no tests, builds, or lint commands were run.

kai-osthoff:
## Blocking finding

The prior finding remains unresolved. The new routing exempts only `doc`; every other Codex `writer-ro`/`writer-only` call still receives the fixed fallback target (`src/transient-agent-argv.ts:162`, `src/transient-agent-argv.ts:170`, `src/transient-agent-argv.ts:173`). The namer still selects `writer-only` (`src/namer-llm.ts:68`, `src/namer-llm.ts:74`) while its reader checks only `.shepherd-name` (`src/namer-llm.ts:48`, `src/namer-llm.ts:52`). The revised namer test now explicitly asserts the unwanted `-o .shepherd-last-message.txt` pair instead of excluding it (`test/namer-llm.test.ts:113`, `test/namer-llm.test.ts:123`). Thus the latest `doc` exception does not implement the caller-specific namer exclusion required by the prior review.

## Verified ground truth

The shared reader preserves primary-file precedence and only reads a caller-supplied fallback (`src/codex-last-message.ts:67`, `src/codex-last-message.ts:72`, `src/codex-last-message.ts:80`). Reviewer spawns generate a per-session filename and reconstruct it from the same session id (`src/transient-agent-argv.ts:173`, `src/critic-core.ts:713`, `src/review.ts:776`, `src/standalone-critic.ts:555`); intended scratch-directory roles use the fixed constant on both sides, for example recap (`src/recap.ts:170`, `src/recap.ts:246`) and distiller (`src/distiller.ts:274`, `src/distiller.ts:575`). No additional in-diff contract inconsistency was identified.

## Verification limitation

I could not independently verify the external Codex CLI's `-o` availability or relative-path semantics from repository ground truth, and execution was prohibited; those runtime semantics are asserted by the implementation documentation at `src/codex-role-argv.ts:15` and `src/codex-role-argv.ts:20`.

kai-osthoff:
**Round-5 finding addressed** in commit `e6879cf4` — Codex `-o` capture is now opt-in per caller, so the namer (and any other non-consuming `writer-only` role) carries no `-o` and no CLI-version dependency.

**Confirmed valid.** The `writer-only` kind is shared by consumers (recap, autopilot, rundown — they read the `-o` fallback) and non-consumers (namer reads only `.shepherd-name`; verify-key reads `.shepherd-verify`). Per-kind emission couldn't tell them apart, so the namer was pulled into the `-o`/version-floor contract it was explicitly excluded from. (verify-key is claude-only, so it never reached the codex branch, but the fix covers it for free.)

**Fix — capture is a per-ROLE property, orthogonal to `kind`.** `kind` governs the tool allowlist / MCP isolation; whether a role reads the CLI's last-message output is separate. `buildTransientAgentArgv` now emits `-o` only when the caller sets the new `captureLastMessage` opt-in. When set, the name still matches the kind's trust posture: reviewer → per-spawn unguessable; other consuming tmpdir kinds → fixed. The nine roles that read the fallback opt in (recap, autopilot, rundown, distiller, optimizer, merge-suggest, and the three reviewers); namer, verify-key, and doc-agent do not → **no `-o`**. This also subsumes the doc-kind special-case from `076bf0c8` (doc simply doesn't opt in), and is future-proof: a new non-consuming role gets no `-o` by default rather than being silently pulled in.

**Tests:** the namer codex argv now asserts no `-o`; the per-kind argv test asserts no `-o` without the opt-in and the correct per-spawn (reviewer) / fixed (others) name with it; the consuming-role service tests still see `-o` because their services opt in. Full suite 7306 pass — only the pre-existing `pty-attach` node-pty worktree failure (green in CI `verify`).

kai-osthoff (commented):
Re-review found no blocking or actionable issues.

The prior finding is resolved: last-message capture is an explicit opt-in (`captureLastMessage`) and the Codex argv builder emits `-o` only when that option is set (`src/transient-agent-argv.ts:82`, `src/transient-agent-argv.ts:171`). The namer leaves the option unset (`src/namer-llm.ts:74`), and its Codex argv regression assertion explicitly excludes `-o` (`test/namer-llm.test.ts:106`).

Verified the delivery contract end to end. The shared reader prefers the role result file and consults the named fallback only when the primary is absent (`src/codex-last-message.ts:67`). Recap, autopilot, rundown, distiller, optimizer, and merge-suggest each opt in and read the fixed tmpdir fallback (`src/recap.ts:247`, `src/autopilot-llm.ts:78`, `src/herd-digest.ts:91`, `src/distiller.ts:274`, `src/optimizer.ts:222`, `src/merge-suggest.ts:247`). The session critic, standalone critic, and plan reviewer opt in and thread the same per-spawn session id back into their readers (`src/review.ts:689`, `src/review.ts:778`, `src/standalone-critic.ts:478`, `src/standalone-critic.ts:557`, `src/plan-gate.ts:164`, `src/plan-gate.ts:732`, `src/critic-core.ts:713`, `src/plan-gate.ts:1281`).

The external Codex CLI version floor and relative `-o` path semantics cannot be independently verified from repository contents; this review treats the runtime verification supplied in the task context as established. Builds and tests were not run, per the review instructions.

kai-osthoff:
**Cross-ref: #1850** (*"Recap generation wedges on claude folder-trust dialog in fresh /tmp/shepherd-recap-* spawn dir"*) is a **sibling** of this PR — **same symptom** (recap generation fails), but a **fundamentally different bug** on a different provider path.

- **This PR — the Codex path.** Codex answers the verdict in chat instead of writing the result file → the runner reads an absent file → `no-result`/`timed-out` (`provider=codex`). Fixed with the `-o` last-message fallback.
- **#1850 — the Claude path.** A Claude recap spawn wedges on the *"Do you trust this folder?"* dialog in the fresh `/tmp/shepherd-recap-*` temp dir → timeout (`provider=claude`). Its fix is trust-seeding the temp dir (per #1075), orthogonal to this PR.

This PR does **not** resolve #1850 and vice versa — they harden different provider legs of the same recap feature. Logging the relationship so triage doesn't treat one as a duplicate of the other.

kai-osthoff:
@scoop 🟢✅🎉🚦 ALL green, ready & waiting 💅😴 do your job already ⏰🚀👀🔥

Issue #1731, opened 2026-07-13
Guard against Shepherd/Codex test storms exhausting host resources


kai-osthoff:
## Summary

On moes-tavern, Shepherd/Codex sessions collectively drove the machine into extreme load and memory pressure by running multiple `bun test`, `vitest run`, and Playwright/Chrome browser test processes in parallel. This looked like a "bun-test storm": killing one wave helped briefly, but active Shepherd sessions immediately spawned new test runs.

This issue captures the incident and the observed diagnostics so we can add guardrails instead of just whacking processes with a shovel.

## Environment

- Host: `moes-tavern`
- Shepherd repo: `erwins-enkel/shepherd`
- Shepherd service: `shepherd.service`
- Shepherd API health during incident: service stayed active; unauthenticated `/api/sessions` returned `401`, which is expected with auth enabled
- System shape observed: 16 CPU cores, ~30 GiB RAM, 15 GiB swap

## What happened

The machine became extremely sluggish. Inspection showed:

```text
Load average: ~133 / 142 / 108 at peak
CPU: ~99% busy
RAM: ~29-30 GiB used of 30 GiB
Swap: ~15 GiB used of 15 GiB
kswapd0: near/at 100% CPU during the worst pressure
```

Process counts during the first inspection:

```text
bun_test          8
vitest            2
chrome_headless   35
claude_agents     11
codex             62
```

The hottest processes were mostly Shepherd worktree test commands:

```text
bun test      ~70-87% CPU each, multiple long-running instances
vitest run    ~18-28% CPU, around 1.2-1.4 GiB RSS each in some cases
chrome-headless / Playwright browser processes, dozens of children
```

The root symptom was not disk space and not a single broken daemon. It was too many verification/test processes running concurrently across active Shepherd/Codex sessions.

## Concrete examples observed

Examples of high-load process groups included:

```text
bun test
node .../.shepherd-worktrees/.../ui/node_modules/.bin/vitest run
node .../.shepherd-worktrees/.../ui/node_modules/.bin/vitest run --project browser
chrome-headless-shell ... Playwright/Chromium children
```

Some `bun test` processes had been running for 20-50+ minutes before cleanup.

There was also an unrelated long-lived Vite dev server consuming memory:

```text
/home/moe/projects/epamano-shopify/category-ui/node_modules/.bin/vite dev
RSS around ~2.1 GiB
elapsed over 6 hours
```

That was not the primary cause of the test storm, but it made memory headroom worse.

[...]

PR #1832, opened 2026-07-18
feat(diagnostics): add host_capacity resource-guardrail check

### LINKED ISSUE #1732: Add Diagnostics check for host resource guardrails
## Summary

Add a dedicated Diagnostics check that warns when a Shepherd host is not protected by resource guardrails, or is already under memory/swap/IO pressure.

The test storm in #1731 showed that Shepherd can currently report its prerequisites as healthy while the host itself is configured in a risky way: no systemd memory/CPU limits, no active oomd protection, saturated swap, and many heavy subprocesses.

Diagnostics should make this visible before the next incident.

## Motivation

Settings → Diagnose currently answers questions like:

- Is `herdr` installed and reachable?
- Is `gh` authenticated?
- Are `bun`, `node`, `git`, `claude`/`codex`, and Tailscale available?

It does not answer:

> Is this host configured safely enough to run multiple Shepherd agent sessions?

That missing signal matters because an unbounded Shepherd deployment can keep spawning expensive local work (`bun test`, `vitest`, Playwright/Chrome, dev servers/watchers) until the host is under heavy memory/swap/IO pressure.

## Proposed check

Add a new diagnostics row:

```text
id: host_capacity
label: Host capacity / resource guardrails
state: ok | warning | error
```

## What it should inspect

At minimum, inspect deployment guardrails and current pressure:

- Whether Shepherd is running under systemd.
- Whether `shepherd.service` or its slice has resource limits:
  - `MemoryHigh`
  - `MemoryMax`
  - `CPUQuota`
  - `TasksMax`
- Whether `systemd-oomd` is active, or another pressure/OOM protection is configured.
- Whether swap is currently saturated.
- Whether `/proc/pressure/{cpu,memory,io}` exists and shows high pressure.
- Whether many active heavy subprocesses are running:
  - `bun test`
  - `vitest run`
  - Playwright/Chrome headless
- Whether suspicious long-lived dev servers/watchers are running under Shepherd worktrees.

## Suggested states

| Condition | State | Example hint |
|---|---|---|
| Resource limits present and host pressure normal | `ok` | Host resource guardrails are configured. |
| No limits, but current pressure normal | `warning` | Shepherd has no MemoryMax/CPUQuota; add systemd limits before running many sessions. |
| Swap saturated / memory PSI high / too many heavy tests | `error` | Host is under memory pressure; pause sessions or defer/kill heavy tests. |
| Not systemd-managed / cannot inspect | `warning` | Shepherd cannot verify host guardrails on this deployment. |

## Minimal v1

Do not overbuild the first pass. A useful v1 can shell out/read:

```bash
systemctl --user show shepherd.service \
  -p MemoryHigh -p MemoryMax -p CPUQuota -p TasksMax -p Slice
systemctl is-active systemd-oomd
free -b
cat /proc/pressure/memory /proc/pressure/io
ps -eo args
```

Then emit one of a small set of safe hint keys:

- `diagnostics_hint_host_capacity_ok`
- `diagnostics_hint_host_capacity_unbounded`
- `diagnostics_hint_host_capacity_pressure`
- `diagnostics_hint_host_capacity_uninspectable`

## Product/design note

The current diagnostics payload is deliberately pure:

```ts
id, state, hintKey
```

It does not carry raw stdout, paths, account identity, or secrets. Keep that property.

For v1, the hint can be generic and safe. If exact numbers are needed later, add a typed, redacted `details` field deliberately rather than dumping raw `systemctl`, `free`, or `ps` output into the UI.

## Remediation guidance

A one-click fix is tempting but risky because resource limits can kill legitimate work. Better first step:

- show a copyable recommended systemd override snippet
- link to docs
- maybe offer a guarded "apply recommended local limits" action later

Example recommendation:

```ini
# ~/.config/systemd/user/shepherd.service.d/resources.conf
[Service]
Slice=shepherd.slice
MemoryHigh=18G
MemoryMax=24G
CPUQuota=1200%
TasksMax=4096
OOMPolicy=stop
```

## Acceptance criteria

- Diagnostics includes a `host_capacity` check.
- The check warns when Shepherd is systemd-managed but has no meaningful resource limits.
- The check errors/warns when current host pressure is dangerous, especially swap saturation or high memory/IO PSI.
- The UI explains the recommended next action without dumping raw command output.
- Existing diagnostics purity is preserved (`id/state/hintKey`; no raw stdout/secrets).
- The check is covered by unit tests with injected command/read probes.

## Related

- Split out from #1731.
- Complements, but does not replace, the runtime launch gate proposed in #1731.

scoop:
Closes #1732.

Adds a `host_capacity` Diagnostics probe that surfaces whether a Shepherd host is protected by resource guardrails and whether it is under live memory/IO pressure — the missing signal from the #1731 test storm, where prerequisites read healthy while the host itself was configured riskily.

## Behaviour

| Condition | State | hint |
|---|---|---|
| memory/IO PSI high, or saturated swap **corroborated** by memory PSI | `error` | pressure |
| systemd-managed + service/slice limits, calm | `ok` | ok |
| systemd-managed + no meaningful limits, calm | `warning` | unbounded (+ systemd docs link) |
| not systemd / non-Linux / limits unreadable | `optional` | uninspectable |

- **Payload purity preserved** — pure `{ id, state, hintKey }`; the facts struct carries only parsed booleans/numbers, never raw `systemctl`/`free`/`/proc` output, paths, or identity.
- **No auto-fix** (limits can kill legitimate work): the `unbounded` row shows a systemd `resource-control` docs link; the hint names the action. `DiagnoseRows` renders generically — no UI component change.

## Design decisions (please sign off)

1. **Swap needs PSI corroboration.** Swap-used alone never triggers `error` — zram / proactive-eviction hosts sit at 90 %+ swap with ~0 PSI. A saturated swap only lowers the memory-PSI bar (10 → 5); PSI (kernel stall time) is authoritative. No `/proc/pressure` ⇒ never guess `error` from swap.
2. **Pressure `error` is exempt from the 60 s recheck acceleration.** `diagnosticsTick` → `check()` re-runs the whole probe fan-out (~10 fork/execs, several launching runtimes). Fast-polling that onto a host already under memory/IO pressure is a feedback loop, so `nextDiagnosticsDelay` now decides from `checks[]` and skips acceleration for `STEADY_STATE_ERROR_CHECK_IDS`. The live path is the panel's Re-run. Alternative considered: downgrade to `warning` (also non-accelerating) — kept `error` for the red-pip severity.
3. **⚠ Uninspectable → `optional`, not the issue's `warning`.** Deliberate divergence so local `bun run` dev boxes (not run as a systemd service) don't pin a permanent yellow health pip. Genuine pressure still `error`s on those hosts. One-line change to `warning` if strict adherence is preferred.

## Scope (v1)

Deferred, per issue "don't overbuild v1": heavy-subprocess scanning (PSI already captures the *effect* of heavy tests), a copyable systemd snippet, one-click apply, and a typed `details` field. `systemd-oomd` and CPU PSI were dropped from the facts struct as non-triggering — kept only fields that drive a decision (slice-limit reading stays: it prevents a false `unbounded`).

## Tests

`test/diagnostics.test.ts`: table-driven `classifyHostCapacity` (every state + precedence + the zram false-alarm guard + swap corroboration), pure parser tests (cgroup unit / systemctl limits / PSI / swap), probe-via-injected-dep, and rewritten `nextDiagnosticsDelay` tests locking in the exemption. All injected — no real host state touched. Root `bun run lint` + `bun test ./test`, `cd ui && bun run check` + `check:i18n` all green.


scoop (commented):
## Summary

This PR adds the `host_capacity` diagnostics probe per issue #1732 and the approved plan. It follows the existing probe pattern (injected `readHostResources` dep + pure `classifyHostCapacity` + pure parsers), preserves payload purity, and is thoroughly unit-tested. I found no blocking defects.

## What I verified

**Payload purity preserved.** `classifyHostCapacity` returns only `{ id, state, hintKey }` (src/diagnostics.ts:383-394), and `HostCapacityFacts` carries only parsed booleans/numbers, never raw command output. No `host_capacity` hint is in the remediation map, so `autoFixCommandFor` returns `undefined` (verified against src/remediations.ts:181-184 and src/diagnostics.ts:906-909) — no `remediation` field is attached; the probe/classifier tests assert `assertPure`.

**Error-fallback path is correct.** `check()` wraps each probe with `run().catch(() => onTimeout)` (src/diagnostics.ts:896), so a rejecting `readHostResources` resolves to the `optional`/`uninspectable` fallback (src/diagnostics.ts:865-873) — matching the test at test/diagnostics.test.ts:1289-1302. `optional` ranks 0 in `worstOf` (src/diagnostics.ts:149-158), so it never reddens `overall`.

**Signature change fully propagated.** `nextDiagnosticsDelay(overall)` → `nextDiagnosticsDelay(checks)` has exactly one production caller, updated at src/index.ts:2606; the only other reference is the test. The new logic (`checks.some(c => c.state === 'error' && !STEADY_STATE_ERROR_CHECK_IDS.has(c.id))`, src/diagnostics.ts:184-191) preserves the old behavior for herdr-style transient errors and correctly exempts a `host_capacity`-only error, verified by tests at test/diagnostics.test.ts:1085-1099.

**i18n parity confirmed.** en.json and de.json each gain the same 7 keys (`feat_host_capacity_title/body`, `diagnostics_label_host_capacity`, and the 4 `diagnostics_hint_host_capacity_*`) — verified 7 occurrences in each of ui/messages/en.json and ui/messages/de.json. All four `hintKey` values emitted by the classifier resolve to existing catalog entries. `DiagnoseRows.svelte` resolves label/hint keys dynamically (`msg[\`diagnostics_label_${id}\`]`, `msg[check.hintKey]`, ui/src/lib/components/DiagnoseRows.svelte:50-55) and renders `optional` as `–` (line 87), so no component change is needed; the `unbounded` warning shows the new DOC_LINKS entry (line 108, ui/src/lib/diagnostics-docs.ts:12-13).

**Feature announcement wired.** The new entry conforms to the `FeatureAnnouncement` type (src/lib/feature-announcements.ts:25-36) and is auto-registered via `import.meta.glob` (line 47-50); `sinceVersion: "1.45.0"` matches other in-flight 1.45.0 entries.

**Classifier / parser logic is sound.** Precedence (pressure error beats guardrail verdict), the zram false-alarm guard (swap saturation requires PSI corroboration, src/diagnostics.ts:355-364), the `limited: null` → uninspectable distinction, and non-Linux/`*.scope` → `null` unit handling all check out and are covered by table-driven tests. The author correctly uses `CPUQuotaPerSecUSec` (the real `systemctl show` property) rather than the issue's non-existent `-p CPUQuota`.

## Divergence from the issue's acceptance table (non-blocking)

The issue's suggested-states table maps "Not systemd-managed / cannot inspect" to `warning`; the implementation emits `optional` instead (src/diagnostics.ts:393) so local `bun run` dev boxes don't pin a permanent yellow health pip. This is an explicit, operator-approved product override called out in the approved plan for reviewer sign-off, not an oversight — genuine pressure still escalates to `error` on those hosts. The core task (a `host_capacity` check that warns on unbounded systemd hosts and errors under dangerous pressure) is satisfied. Flagging for visibility only.

## Could not fully verify (not defects)

- Exact `systemctl show` output format for `CPUQuotaPerSecUSec` on a live host isn't reproducible in-repo. The parser only distinguishes `infinity`/empty from anything else, so classification is correct regardless of whether systemd emits a raw usec integer or a humanized form, as long as unset renders as `infinity` (documented systemd behavior).

- src/diagnostics.ts:325-330 — cgroup-unit detection reads only the `0::` leaf; a systemd service that delegates and runs the process in a sub-cgroup (leaf not `*.service`) falls through to `uninspectable` rather than reading limits. Safe (never a false `ok`/`unbounded`), and the plan documents this as a best-effort cgroup-v2 assumption.

- src/diagnostics.ts:397-399 — `DEFAULT_SLICES` slice-skip optimization is a small addition beyond the minimal path, but it's justified (avoids a needless second `systemctl show`) and low-risk.

scoop (commented):
## Summary

This PR adds the `host_capacity` Diagnostics probe (#1732) following the existing probe pattern: one injectable dep (`readHostResources`), a set of small pure parsers, and an exported pure classifier (`classifyHostCapacity`). The check warns when a systemd-managed host lacks resource guardrails, errors under dangerous live PSI, and degrades to `optional` when uninspectable. It also exempts the steady-state `host_capacity` pressure error from the background re-check acceleration. Implementation is clean, payload purity is preserved, and coverage is thorough. No blocking issues found.

### Verified

- **Signature change wired everywhere.** `nextDiagnosticsDelay` now takes `checks[]` instead of `overall` (src/diagnostics.ts:178). Its only non-test caller is updated to pass `snapshot.checks` (src/index.ts:2606); grep confirms no other callers.
- **Payload purity.** `classifyHostCapacity` (src/diagnostics.ts:341-353) returns only `{id, state, hintKey}`. No `host_capacity` hint appears in src/remediations.ts (REMEDIATIONS map at :97), so `autoFixCommandFor` attaches no `remediation` (src/diagnostics.ts:906-909) — `assertPure` holds.
- **Config constants.** All four (`HOST_SWAP_SATURATION_RATIO`, `HOST_PSI_MEMORY_AVG10`, `HOST_PSI_MEMORY_AVG10_CORROBORATED`, `HOST_PSI_IO_AVG10`) are exported (src/config.ts:104-107) and consumed (src/diagnostics.ts:372-376).
- **i18n parity.** The label + 4 hint keys + 2 `feat_host_capacity_*` keys exist in BOTH ui/messages/en.json (:1275-1276, :1331, :1354-1357) and ui/messages/de.json (:1275-1276, :1331, :1354-1357). Only en/de catalogs exist. `optional` is a valid `DiagnosticState` (src/types.ts:462).
- **UI wiring is generic.** DiagnoseRows.svelte:51 resolves labels via `diagnostics_label_${id}` and DOC_LINKS keys off hintKey, so the new row + the `unbounded` doc link (ui/src/lib/diagnostics-docs.ts:12) render with no component change.
- **Feature announcement auto-registers.** entries/*.ts are glob-imported (ui/src/lib/feature-announcements.ts:47); id `host-capacity` is unique; `sinceVersion 1.45.0` matches the two already-merged v1.45.0 entries.
- **Classifier logic matches the table + tests.** `optional`/`ok` rank 0 in `worstOf` (src/diagnostics.ts:149), so uninspectable never reddens the pip; pressure precedence, swap-corroboration, and the `pressure: null` fall-through all match the test cases (test/diagnostics.test.ts:1090-1180).
- **`hasMeaningfulLimit` is format-robust:** it treats any non-empty, non-`infinity` value as "set", so it works whether `systemctl show` renders USec/byte properties as raw integers or human-readable forms.

### Deliberate deviations from the issue (author flagged these for sign-off — I concur they are sound, non-blocking)

- **Uninspectable ⇒ `optional`, not `warning`** (the issue's suggested-states table says `warning`). Reasonable: most installs run locally via `bun run`, and a permanent yellow pip would train operators to ignore it. Genuine pressure still errors on those hosts.
- **Swap saturation alone never errors** (the issue lists "swap saturation" as a danger trigger). The implementation requires PSI corroboration (src/diagnostics.ts:374-376). This is the correct call — swap-used% is not a stall signal on zram/zswap hosts; PSI is authoritative. On pre-4.20 kernels with no PSI, a saturated-but-quiet host stays healthy rather than false-erroring.
- systemd-oomd detection and heavy-subprocess scanning are explicitly out of v1 scope per the plan; the minimal acceptance set is satisfied.

### Could not verify (external ground truth, not in repo)

- Exact `systemctl show` output format for `CPUQuotaPerSecUSec`/`MemoryMax` on a live host. The parser is format-agnostic (only distinguishes `infinity`/empty from anything else), so this does not affect correctness regardless of the concrete rendering.

No findings to address.

Issue #709, opened 2026-06-15
Hook ingestion Phase 2 — lifecycle edges (SessionStart / Stop / SessionEnd)


scoop:
Phase 2 of the hook-based agent-info ingestion effort (umbrella #704; research `docs/research/claude-code-hooks-ingestion.md`). Builds on the infrastructure shipped in #708 (Phase 0 ingest + Phase 1 activity/block signals).

## Goal
Add **lifecycle-edge** signals via `SessionStart` / `Stop` / `SessionEnd` hooks for precise boot / turn-complete / end edges feeding autopilot, critic, and recap — replacing herdr-"done" ambiguity (memory _herdr-done-is-idle-not-exited_).

| Hook | Replaces / improves | Win |
| --- | --- | --- |
| `SessionStart` | herdr-list liveness poll for boot | instant "agent up" confirmation |
| `Stop` | herdr-"done" ambiguity | precise turn-complete edge for autopilot/critic triggers |
| `SessionEnd` | archive-time recap inference (#665) | authoritative end edge → recap/teardown |

## Approach (additive, flag-gated, fail-open — same as #708)
- **Injection:** add `SessionStart`/`Stop`/`SessionEnd` keys to `buildHooksFragment` (`src/service.ts`) — same per-session URL + auth as the existing three events.
- **Ingest:** `validateHookEvent` (`src/hooks-ingest.ts`) normalizes the new events (`SessionStart.source`, `Stop.stop_hook_active`, `SessionEnd` reason). The route + ring buffer are already event-agnostic.
- **Wiring:** extend the index.ts sink → new poller/downstream handlers (e.g. `Stop` → turn-complete edge consumed by autopilot/critic; `SessionEnd` → recap/teardown trigger). Polling stays the fallback.
- Reuse `SHEPHERD_HOOKS_SIGNALS` (or a dedicated flag); default off.

## Depends on
- **#708 merged** (establishes injection/ingest/sink seams).
- **Phase 0 verification** from #708 informs design: confirmed sandbox reachability + transport choice (`http` vs curl `command`), and whether env-token auth works under the sandbox profile. Resolve those before relying on these edges.

## Notes
- Server-side wiring; likely no UI → check whether any surfacing makes it user-facing (then i18n + feature-catalog apply).
- `Stop` can gate (force-continue) but this phase is **observe-only** — no gating.

scoop:
✅ PR #718 merged.

PR #686, opened 2026-06-14
fix(merge-train): launch driver with autopilot off regardless of repo default

scoop:
## Problem

The merge-train **driver** session is created via `mergeTrainCreateInput`, which force-sets `planGateEnabled: false` but says nothing about autopilot. So the driver inherits the repo's `autopilotEnabled` default — and in an autopilot-on repo it gets classified/steered into unrelated tasks between/after PR landings (the observed "harassment").

A second, independent vector: build-queue pre-approval at session create keyed on the **repo** default (`repoConfig.autopilotEnabled`), not the session — so the driver could also be auto-approved into build-queue execution.

## Fix

- Add a per-task `autopilotEnabled?: boolean | null` create-input override, mirroring the existing `planGateEnabled` plumbing (validate → service → store). Reuses the existing `autopilotEnabled` column — **no new field, no migration, no schema change.**
- `mergeTrainCreateInput` now sets `autopilotEnabled: false`, so the driver launches autopilot-off regardless of the repo default. The existing `effectiveAutopilot()` gate (`s.autopilotEnabled ?? repoDefault`) then stands the driver down at every autopilot entry path (`eligible()`, `considerCi`, `reEngageCi`) — no `autopilot.ts` change needed.
- Build-queue pre-approval re-keyed onto the session's **effective** autopilot via a new `shouldPreApproveBuildQueue()` helper, excluding the driver (and any explicitly-autopilot-off session).
- Relaunch carries `autopilotEnabled` forward so a relaunched driver stays locked off (incl. its own build-queue check).

**Decision:** default-off at launch (not an un-toggleable hard-lock) — fully satisfies "launch without autopilot, no matter the repo settings"; an operator can still manually re-enable per session. The manual path was never the source of the harassment.

## Tests

- `merge-train.test.ts` — both kickoff framings assert `autopilotEnabled: false`.
- `autopilot.test.ts` — driver override-off under repo-default-on is never classified/steered.
- `store-autopilot.test.ts` — override round-trips (`false`→`false`, absent→`null` inherit).
- `validate.test.ts` — accepts `true`/`false`/`null`/absent, rejects non-boolean.
- `service.test.ts` — build-queue pre-approval excludes an autopilot-off session under a build-queue+autopilot repo (no regression to the normal autopilot case).

## Verification

[...]

PR #708, opened 2026-06-15
feat(hooks): push-based agent-info ingestion via Claude Code hooks — Phase 0 + 1 (#704)

scoop:
Implements push-based agent-info ingestion via Claude Code hooks (#704), **additive on top of the poll-based pipeline**. Per the research doc (`docs/research/claude-code-hooks-ingestion.md`, #701). Scope: **Phase 0 + Phase 1**, both **default-OFF**, **observe-first**, **fail-open** — with both flags off the system is byte-for-byte today's behaviour.

## What ships

**Phase 0 — ingest spike** (`SHEPHERD_HOOKS_INGEST=1`)
- `spawnSettingsOverlay` injects `PostToolUse` / `PostToolUseFailure` / `Notification` HTTP hooks into every Shepherd-spawned agent's `--settings`, baking the per-session URL `…/api/sessions/<s.id>/hooks`.
- New `POST/GET /api/sessions/:id/hooks` route (sibling of the build-queue route) + `src/hooks-ingest.ts`: validates the untrusted body, records into a bounded in-memory ring buffer, logs each event. **Observe-only** — no signal-pipeline wiring at this stage.

**Phase 1 — retire fragile heuristics** (`SHEPHERD_HOOKS_SIGNALS=1`, requires ingest)
- Received events feed the **poller** (the single owner of per-session signal dedup + state), never emitting events directly:
  - `PostToolUse`/`PostToolUseFailure` → sub-second structured **activity** via the existing `emitActivity` dedup (real tool summary vs. the interim `summary:null` heartbeat; `PostToolUseFailure` carries error heat).
  - `Notification(awaiting-input type)` → authoritative **awaiting-input edge** that triggers a PTY classify on the next tick (independent of herdr's latchy `blocked` status). `classifyBlocked` still supplies menu options (the Notification payload has none).
- A **freshness guard** suppresses the redundant probe activity emit when a hook is fresh, but **never** the stall evaluation or error heat — polling stays the fallback.

## Design / correctness notes
- **Two distinct UUIDs:** the URL/route key on Shepherd's `s.id`; the body's `session_id` (= `claudeSessionId`) is validated as an untrusted cross-check; mismatches are recorded observe-only, never forwarded.
- **Single Bun loop respected:** the route only validates + records (no file/PTY I/O); the awaiting-edge classify runs on the poll tick where classify already runs.
[...]

PR #707, opened 2026-06-15
feat: in-app DIAGNOSE one-click Fix + keystone src/remediations.ts (#703)

scoop:
Implements **Phase 1 + Phase 2** of the first-run installer (#703): the keystone refactor and the in-app DIAGNOSE **Fix** surface. The cold-start `curl|bash` bootstrap (Surface A) and the Phase-4 harness e2e gate are deliberately deferred — tracked in **#706**.

Refs #703 (does **not** close it — Surface A / acceptance-criterion #3 ship in #706).

### Phase 1 — keystone: one canonical remediation table
- Promotes `REMEDIATIONS` + `remediationsFor` out of the test harness into shipped **`src/remediations.ts`** (git-tracked rename, no behavior change). The harness now imports from `src/`.
- Adds the **guidance-only taxonomy** — `GUIDANCE_ONLY` + `autoFixCommandFor(hintKey)`. Membership in `REMEDIATIONS` is *not* the auto-fixable set: `tailscale_missing` has an install command (for the future cold-start installer) but the binary alone never clears the check (`resolveNodeHost` stays null until an interactive tailnet login), so it's gated out of in-app Fix — no false-failure button.

### Phase 2 — Surface B: in-app DIAGNOSE Fix
- **`POST /api/diagnostics/fix { checkId }`** — resolves the check's `hintKey` against the table, runs the **verbatim** command server-side (operator's user), forces a fresh probe, returns the snapshot, and emits `diagnostics:status` so every client + the TopBar pip refresh. Behind the existing `checkAuth`/`checkOrigin` (CSRF + token) guards.
  - **Single-loop safe:** spawns a **detached process group** (async, never sync), drains stdout/stderr into a bounded 1 MiB sink (no ENOBUFS on noisy installs), and **SIGKILLs the whole group on timeout** (`REMEDIATION_TIMEOUT_MS`) so no reparented grandchild survives to flip a later probe green behind a reported failure.
  - **Fail closed:** non-zero exit / timeout / unknown-or-guidance-only check → explicit non-2xx (502 / 409), never a masked success.
- **DiagnoseRows.svelte** — a **Fix** button on every non-ok auto-fixable row (symmetric across warning + error), behind a blurred-scrim confirm modal that shows the **exact command** before running. Guidance-only (tailscale) and prose-only (git/gh) rows keep their prose hint, no button.
[...]

PR #1833, opened 2026-07-18
fix(epic): dedupe epic-dag members so multi-line blockers don't break the panel

scoop:
## Problem

An epic's panel failed to load (screenshot showed a stuck `loading…`, then the panel never appeared). Reproduced against flowagent epic **#686**: `GET /api/epic?…&parent=686` returns **HTTP 200** but with `children: [707, 708, 709, **709**]` — issue **#709 duplicated**.

Root cause: #686's `epic-dag` expresses two blockers by listing the child on two lines:

```
#709 <- #707
#709 <- #708
```

`parseFencedEpic` did `order.push(dependent)` **once per line**, so `#709` landed in `order` twice → `assembleEpic` produced a duplicate child → `EpicPanel`'s keyed `{#each … (c.number)}` hit a duplicate key. Svelte 5 throws `each_key_duplicate` in **both dev and production**, and it fires in an **async effect with no `<svelte:boundary>`** around it, so the panel crashes on mount and the epic never loads. It also skewed the badge count (0/**4** vs the real 0/3).

## Fix

- **`src/epic-parse.ts`** — a member is a set: push each node **once** (first-seen position), still unioning its edges across lines. Covers the fenced and task-list paths.
- **`src/epic-model.ts`** — dedupe `order` in `assembleEpic` before building children as **defense-in-depth**, guarding the native path (repeated sub-issue number) and any future source.
- Side benefit: fixes the native-import path — `importEpicLinks` no longer re-`addSubIssue`s a duplicated member into `unresolved`.

Multi-line blockers are already valid input; no grammar/authoring change.

## Verification

- New parser on real **#686** body → `members: [707,708,709]`, both edges preserved, badge total **3**.
- `test/epic-parse.test.ts` — multi-line blockers → member once, edges unioned; repeated task-list member.
- `test/epic-model.test.ts` — native duplicate sub-issue number → single child (guard seam).
- `test/epic-import.test.ts` — multi-line-blocker body → child added once, both edges wired, `unresolved` empty.
- `ui/…/EpicPanel.browser.test.ts` — reproduces the `each_key_duplicate` crash (before) and a clean mount rendering child rows (after).
- Green: `bun run lint`, `bun test ./test` (7226 pass), `cd ui && bun run check` (0 errors), full UI browser suite (1493 pass).

## Consumer sweep

`parseEpicBody().members` feeds ~8 sites; dedup impact audited: `markdownCounts` corrects the /api/epics badge 4→3 (desired); presence checks (`.length > 0`) are no-ops; the rest fold members into a `Set` (no-op). No site relied on positional/duplicate semantics.

### DIFF
--- a/src/epic-model.ts
+++ b/src/epic-model.ts
@@ -116,7 +116,12 @@ function resolveMarkdown(input: AssembleInput): ResolvedGraph {
 export function assembleEpic(input: AssembleInput): Epic {
   const native = input.subIssues.length > 0;
   const graph = native ? resolveNative(input) : resolveMarkdown(input);
-  const { order, resolved, edges } = graph;
+  // Defense-in-depth: children are keyed by number in EpicPanel's {#each}, so a duplicate in
+  // `order` (from any source) would throw each_key_duplicate and crash the panel on mount. The
+  // markdown parser already dedupes members; this guards the native path (repeated subIssue
+  // number) and any future source. First-seen wins; `members`/`done` derive from `order` below.
+  const { resolved, edges } = graph;
+  const order = [...new Set(graph.order)];
   const warnings = [...graph.warnings];
 
   const members = new Set(order);
--- a/src/epic-parse.ts
+++ b/src/epic-parse.ts
@@ -30,7 +30,11 @@ function parseFencedEpic(fenceBody: string): ParsedEpic {
     const m = line.match(LINE_RE);
     if (!m) continue;
     const dependent = Number(m[1]);
-    order.push(dependent);
+    // A member is a set: a node listed on multiple lines (e.g. `#9 <- #7` then `#9 <- #8`
+    // to express two blockers) counts once, keeping its first-seen position — later lines
+    // still contribute their edges below. A duplicate member would otherwise flow into a
+    // duplicate epic child and break EpicPanel's keyed {#each} (each_key_duplicate).
+    if (!order.includes(dependent)) order.push(dependent);
     if (m[2]) edges.push(...parseFenceEdges(dependent, m[2]));
   }
   return { members: [...order], order, edges };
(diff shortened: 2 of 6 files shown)


scoop (commented):
## Summary

This PR fixes the "epic not loading" bug where an `epic-dag` fence listing the same child on multiple `<-` lines (a natural way to express multiple blockers) produced a duplicate entry in `order`, which flowed into a duplicate epic child and crashed EpicPanel's keyed `{#each ... as c (c.number)}` with `each_key_duplicate`.

The fix is correct, minimal, and matches the approved plan:

- **`src/epic-parse.ts:37`** — fenced path pushes each `dependent` to `order` only on first sighting (`!order.includes(...)`), keeping first-seen position, while `edges` still accumulate on every line (`src/epic-parse.ts:38`). Verified: `#709 <- #707` then `#709 <- #708` → member `709` once, both edges present (pinned by `test/epic-parse.test.ts`).
- **`src/epic-parse.ts:48-51`** — checklist path dedupes `members` identically; `order: [...members]` inherits it.
- **`src/epic-model.ts:124`** — `order` is de-duplicated via `[...new Set(graph.order)]` before `members` (`:127`), `done` (`:128`), and `children` (`:134`) derive from it; defends the native path (`resolved` is a Map, so a repeated native number still resolves at `:146`).
- **`src/epic-import.ts:82`** — iterates `parsed.order`; with dedup the child is added exactly once, avoiding the prior spurious `unresolved` re-add (pinned by `test/epic-import.test.ts`).

## Verification

- `EpicPanel.svelte:129` confirms the keyed-each is by `c.number` — the duplicate-key crash vector the dedup closes.
- `EpicPanel.svelte:136-138` renders `c.blockedBy` via `.map().join(", ")` (plain text, not a keyed each), so unioned/duplicate blocker edges cannot reintroduce an `each_key_duplicate` after the panel now mounts. No new crash path is opened.
- `src/epic-import.ts:82-95` confirmed the import loop iterates `parsed.order` and wires edges from `parsed.edges`, consistent with the added test expectations.

All identifiers, return shapes, and consumers I checked are internally consistent with the diff. No blocking issues found.

Minor observations (non-blocking, not raised as findings): the `!order.includes(...)` / `!members.includes(...)` dedup is O(n²), which is irrelevant at epic sizes; and identical duplicate edge lines would render `#707, #707` as blocker text — a pre-existing cosmetic edge case unchanged in substance by this PR and not a crash.

PR #1836, opened 2026-07-18
feat(files): show New Task attachments in the session Scratchpad (#1717)

### LINKED ISSUE #1717: Show New Task attachments in the session scratchpad

scoop:
Closes #1717.

## Problem
New Task attachments are copied into `<worktree>/.shepherd-uploads` and injected into the agent prompt, but the Files tab's default **Scratchpad** view reads a *different* tree (`<claudeTmpRoot>/…/<claudeSessionId>/scratchpad`). So initial screenshots never appear there, and for non-Claude providers (blank `claudeSessionId`) the Files tab is hidden entirely.

## Approach — reader-merge (writer/prompt/recovery untouched)
Overlay the existing `.shepherd-uploads` dir as a synthetic, read-only **Attachments** folder inside the Scratchpad view. Nothing about the writer, prompt injection, or retry/held/relaunch/missing-staged-upload recovery changes — the prompt still references usable `.shepherd-uploads` paths.

- **Server** (`src/scratchpad.ts` + `src/server.ts`): `listScratchpadMerged` / `resolveScratchpadOrAttachmentFile` dispatch each path to exactly one root — `attachments/…` resolves against `<worktree>/.shepherd-uploads` (contained by the same realpath helpers), everything else against the scratchpad root (unchanged). The synthetic folder is **sorted in** with the standard dirs-first/alpha comparator (not pinned), and **deduped** so a real scratchpad dir named `attachments` never yields two rows.
- **Provider-agnostic**: the uploads dir is worktree-keyed (no `claudeSessionId`), so the folder — and the Files tab (`computeHasFiles`) — surface for non-Claude sessions too.
- **Scope**: the overlay reflects the whole `.shepherd-uploads` dir, i.e. New Task attachments **and** mid-session compose-box uploads (both land there).
- **Read-only**: uploading *into* the subtree is disabled client-side (a dedicated `uploadDisabled` flag — not `readOnly`, which would mislabel the folder with Worktree strings) and fails closed server-side (a `path=attachments` upload routes to the real scratchpad root → 404).

## Tests
- `test/scratchpad-attachments.test.ts` (9): merged-root sort position, provider-agnostic (blank `claudeSessionId`), subtree listing + remapped paths, a real mid-session `?session=` upload appearing, download, `..` containment → 404, upload-to-`attachments` fails closed (404, uploads dir untouched), dedupe, no-uploads fallback.
- `ui/src/lib/session-files.test.ts` (5): `computeHasFiles` true/false/dropped/archived.
- `ui/src/lib/components/viewport/FilesPanel.browser.test.ts` (+2): localized Attachments folder from the initial load; inside the subtree the breadcrumb-root/empty labels stay the scratchpad variants while the upload affordance is hidden.
- Existing `test/service.test.ts` prompt assertions (exact `Attached files:\n…/.shepherd-uploads/…` block) stand as the agent-path-delivery regression guard — untouched writer.

Full `bun test` (7235), UI `bun run test` (3990), root lint, `bun run check`, `check:i18n`, and the PR-hygiene gates (feature-catalog, announcement-versions, glossary, branch-hygiene, fallow audit) all pass.

## Out of scope
The Scratchpad/Worktree explanatory tooltip (tracked separately).

### DIFF
--- a/docs/external-task-api.md
+++ b/docs/external-task-api.md
@@ -134,10 +134,16 @@ Once a task exists, an external agent can also drive it:
   changed lines plus routed critic findings) as `{ "notes": [...] }`; it degrades
   to an empty list on any error rather than failing.
 - `GET /api/sessions/:id/scratchpad[?path=]` — browse a live session's own
-  scratchpad subtree; `GET /api/sessions/:id/scratchpad/download?path=`
-  streams a single file. Paths are relative to the scratchpad root and
-  realpath-contained to it (`..`, absolute, and symlink escapes are rejected);
-  both `404` on a missing/archived session.
+  scratchpad subtree, with the session's operator attachments overlaid as a
+  synthetic read-only `attachments/` folder (New Task screenshots and
+  mid-session compose-box uploads, which physically live in
+  `<worktree>/.shepherd-uploads`). `GET /api/sessions/:id/scratchpad/download?path=`
+  streams a single file. Paths are relative to the merged root: `attachments/…`
+  paths resolve against the worktree uploads dir, everything else against the
+  scratchpad root, each realpath-contained to its own root (`..`, absolute, and
+  symlink escapes are rejected). Because the overlay is worktree-keyed, this view
+  is provider-agnostic and surfaces even for non-Claude sessions with no
+  scratchpad of their own. Both `404` on a missing/archived session.
 - `GET /api/sessions/:id/worktree[?path=]` — browse a live session's git
   worktree subtree (read-only); `GET /api/sessions/:id/worktree/download?path=`
   streams a single file. Paths are relative to the worktree root and
--- a/src/fs-browse.ts
+++ b/src/fs-browse.ts
@@ -23,6 +23,12 @@ export interface BrowseEntry {
   createdMs?: number;
   /** true when this entry is a symlink that resolves OUTSIDE the root (only set in onEscape:"mark"). */
   linkOutside?: boolean;
+  /**
+   * true only on the synthetic "Attachments" folder the scratchpad merge overlays onto the
+   * root (see src/scratchpad.ts). The UI keys on this to render a localized folder label; no
+   * real filesystem entry ever sets it.
+   */
+  attachments?: boolean;
 }
 
 /** Creation date (epoch ms) from a stat: birth time when recorded, else modified-time fallback. */
--- a/src/scratchpad.ts
+++ b/src/scratchpad.ts
@@ -128,11 +128,14 @@ function attachmentsSubPath(norm: string): string {
 
 /**
  * Merged scratchpad root: the session's own scratchpad entries plus the synthetic `attachments`
- * folder (when the worktree has any operator attachments). Dedupe: if the real scratchpad already
- * holds a dir literally named `attachments`, the synthetic overlay wins — the real entry is dropped
- * so the root never emits two `attachments` rows. The concat order is irrelevant because the result
- * is sorted with the same dirs-first-then-locale-alpha comparator `listDir` uses everywhere else.
- * Returns null only when there is neither a scratchpad nor any attachment (nothing to show).
+ * folder (when the worktree has any operator attachments). Dedupe is conditional on the overlay
+ * being present: only when there ARE uploads does a real scratchpad dir literally named
+ * `attachments` get dropped in favor of the synthetic one (so the root never emits two rows). With
+ * NO uploads there is no overlay, so a real `attachments` entry is kept and stays browsable — see
+ * `listScratchpadMerged`, which likewise only claims the `attachments/` namespace when uploads
+ * exist. The concat order is irrelevant because the result is sorted with the same
+ * dirs-first-then-locale-alpha comparator `listDir` uses everywhere else. Returns null only when
+ * there is neither a scratchpad nor any attachment (nothing to show).
  */
 async function listScratchpadRoot(
   worktreePath: string,
@@ -142,7 +145,9 @@ async function listScratchpadRoot(
   const hasAttachments = await attachmentsHasFiles(worktreePath);
   if (!scratch && !hasAttachments) return null;
 
-  const base = (scratch?.entries ?? []).filter((e) => e.path !== ATTACHMENTS_DIR);
+  // Drop a real `attachments` entry ONLY when the synthetic overlay replaces it; otherwise keep it.
+  const raw = scratch?.entries ?? [];
+  const base = hasAttachments ? raw.filter((e) => e.path !== ATTACHMENTS_DIR) : raw;
   const entries = hasAttachments ? [attachmentsEntry(), ...base] : base;
   entries.sort((a, b) =>
     a.type !== b.type ? (a.type === "dir" ? -1 : 1) : a.name.localeCompare(b.name),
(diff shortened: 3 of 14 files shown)


scoop:
## Summary

The PR implements the reader-merge approach from the plan faithfully: attachments stay in `<worktree>/.shepherd-uploads`, and the Scratchpad view overlays them as a synthetic `attachments/` folder. Containment, path re-mapping, provider-agnostic visibility, the `computeHasFiles` gate, i18n parity, and the feature-announcement entry all check out. I found one reachable logic defect in the dedupe path.

## Findings

### src/scratchpad.ts — a real scratchpad dir named `attachments` disappears (and becomes unreachable) when there are no uploads

The dedupe in `listScratchpadRoot` drops the real entry unconditionally, but only re-adds the synthetic overlay when uploads exist:

- src/scratchpad.ts:145 — `const base = (scratch?.entries ?? []).filter((e) => e.path !== ATTACHMENTS_DIR);` strips a real scratchpad directory literally named `attachments` **always**.
- src/scratchpad.ts:146 — `const entries = hasAttachments ? [attachmentsEntry(), ...base] : base;` only re-adds the synthetic folder when `hasAttachments` is true.

So when a Claude session's agent creates a scratchpad directory named `attachments` and **no operator attachments/uploads exist** (`hasAttachments === false`), the real folder is filtered out of `base` and nothing replaces it — it vanishes from the root Files listing entirely. Before this PR that folder rendered and was browsable, so this is a silent regression that hides a genuine user directory.

It is also unreachable by direct path: `listScratchpadMerged` (src/scratchpad.ts:166–168) routes any `attachments`/`attachments/...` path to `worktreeUploadsDir` **unconditionally**; with an empty/missing uploads dir `listDir` returns null → the route 404s (src/server.ts:2255–2258). Download has the same shape via `resolveScratchpadOrAttachmentFile` (src/scratchpad.ts:184–190). So both listing and download of the real `attachments` subtree fail when uploads are absent.

This path executes on every root listing today; the trigger is an unusual-but-plausible folder name plus no uploads. The plan's documented intent ("drops the real entry, keeps the synthetic one") assumes the synthetic is present — it isn't in the no-uploads case, so the real entry should be kept. Suggested fix: only strip the real entry when the synthetic overlay is actually being added, e.g. gate the filter on `hasAttachments` (`const base = hasAttachments ? entries.filter(...) : entries;`). The existing dedupe test (test/scratchpad-attachments.test.ts:193) only covers the uploads-present case, so this branch is untested.

## Verification notes

- i18n parity confirmed: `files_attachments_folder`, `feat_scratchpad_attachments_title`, `feat_scratchpad_attachments_body` all present in both ui/messages/en.json:2295,3258-3259 and ui/messages/de.json:2295,3258-3259.
- `computeHasFiles` reads `launchMetadata.attachments[].dropped`, which matches `LaunchAttachmentMetadata` (ui/src/lib/types.ts:1864-1869) and `SessionLaunchMetadata` (ui/src/lib/types.ts:1871-1876).
- `ATTACHMENTS_DIR = "attachments"` is defined identically in src/scratchpad.ts:79 and ui/src/lib/session-files.ts:8; FilesPanel imports the UI copy (FilesPanel.svelte:29) and gates only upload affordances via `uploadDisabled` (FilesPanel.svelte:37,227,234,243,306,365), leaving label/aria derivations keyed on `source`/`readOnly` (FilesPanel.svelte:162-168) — the Attachments subtree keeps scratchpad strings as intended.
- Feature-announcement entry `v1.45.0-scratchpad-attachments.ts` is auto-discovered by the `import.meta.glob("./feature-announcements/entries/*.ts")` in ui/src/lib/feature-announcements.ts:47-48; `targetId` is optional (ui/src/lib/feature-announcements.ts:30), and 1.45.0 matches sibling entries.
- `worktreeUploadsDir` resolves to `<worktreePath>/.shepherd-uploads` (src/uploads.ts:70-71); path re-map (`remapUploadsListing`) and containment via the existing realpath helpers verified — `attachments/../../escape` normalizes out of the reserved prefix and is rejected (covered by test/scratchpad-attachments.test.ts:162).
- No broken callers: `listScratchpad`/`resolveScratchpadFile` remain exported and are still used internally and by test/scratchpad.test.ts; server.ts is the only route consumer and now calls the merged variants.

- None observed — the diff stays within the plan's stated scope.

scoop (commented):
## Re-review

The single prior finding is **fully addressed**.

**Prior finding — `listScratchpadRoot` dropped a real `attachments` dir when no uploads existed:**
- `src/scratchpad.ts:145` now gates the filter on `hasAttachments`: `const base = hasAttachments ? raw.filter((e) => e.path !== ATTACHMENTS_DIR) : raw;` and `const entries = hasAttachments ? [attachmentsEntry(), ...base] : base;`. With no uploads the real entry is kept and unmarked.
- `listScratchpadMerged` (src/scratchpad.ts) and `resolveScratchpadOrAttachmentFile` both guard the `attachments/` dispatch with `&& (await attachmentsHasFiles(worktreePath))`, so with no uploads an `attachments/...` path falls through to the scratchpad root and stays browsable/downloadable — verified against src/scratchpad.ts.
- A no-uploads dedupe test was added at `test/scratchpad-attachments.test.ts:215` ("with NO uploads, a real scratchpad `attachments` dir is kept at root and stays browsable"), asserting the root row is the real one (`attachments` undefined) and navigating in resolves against the scratchpad root.

## Verification performed
- Sort comparator in `listScratchpadRoot` (src/scratchpad.ts) matches `listDir` (fs-browse.ts:224-226) — dirs-first then `localeCompare`; the synthetic folder is sorted in, not pinned (test/scratchpad-attachments.test.ts:90).
- Path remap/containment: `attachments/../../escape` normalizes out of the reserved prefix and 404s (test:162); `resolveScratchpadOrAttachmentFile` reuses `resolveFileInRoot` realpath containment.
- Upload contract fails closed: `POST scratchpad/upload?path=attachments` routes to the real scratchpad root (no server change), 404s, uploads dir untouched (test:175).
- i18n parity confirmed: `files_attachments_folder`, `feat_scratchpad_attachments_title`, `feat_scratchpad_attachments_body` all present in both en.json (2295, 3258-3259) and de.json (2295, 3255-3256). Worktree label keys referenced by the browser test also exist in both (de.json:2763-2765, en.json:2763-2765).
- `computeHasFiles` (ui/src/lib/session-files.ts) reads `launchMetadata.attachments[].dropped`, consistent with `LaunchAttachmentMetadata.dropped` (ui/src/lib/types.ts:1867) and `SessionLaunchMetadata.attachments` (types.ts:1875).
- Feature-announcement entry (v1.45.0-scratchpad-attachments.ts) matches the sibling shape (id/sinceVersion/titleKey/bodyKey) and is auto-loaded via the `import.meta.glob` in feature-announcements.ts:47.
- FilesPanel.svelte gates only upload affordances (button + drag/drop) on `uploadDisabled = readOnly || withinAttachments`, keeping label/aria keyed on `source` — matches the plan; browser test locks the scratchpad-vs-worktree label distinction.

No blocking or non-blocking issues found in the diff.

PR #1846, opened 2026-07-19
fix(ui): make GitRail's marked/dompurify imports dynamic

scoop:
## What

`GitRail.svelte` statically imported `marked` + `dompurify`. GitRail is always mounted, so Rollup hoisted both into the main chunk and **silently cancelled the lazy-loading in all seven other consumers** — the `INEFFECTIVE_DYNAMIC_IMPORT` build warnings.

Several of those siblings carry comments explaining they lazy-load *"so they stay off the first-paint critical path."* That intent was dead code. One static import was defeating seven deliberate dynamic ones.

| Component | Style | Libraries |
|---|---|---|
| SessionRecap, PlanPanel, DoneRecapPanel, HerdrUpdateModal, CalloutBlock, RichTextBlock | dynamic | marked + dompurify |
| `blocks/WireframeBlock.svelte:132` | dynamic | **dompurify only** — never imports marked |
| `GitRail.svelte:30-31` | **STATIC** | both — hoisted them into the main chunk |

Also drops the now-orphaned `getDOMPurify()` shim and its two local types. In a browser dompurify's ESM default is already sanitize-capable — which is why the seven siblings work without the shim. Verified it had exactly one consumer.

## Why gates

**Rollup warnings do not fail `bun run build`.** Without a gate, a future static import silently re-breaks the lazy-loading with a green build.

- **`scripts/check-ui-build.sh`** — builds, then fails on `INEFFECTIVE_DYNAMIC_IMPORT`. Called by **both** CI's `Build (ui)` step and the pre-push ui lane, so the two stay in sync (`ci.yml:24-29` requires this) with one implementation rather than two copies of the same shell.
- **ESLint `no-restricted-imports`** — bans the static form in `**/*.svelte{,.ts,.js}`. Fails in milliseconds at the exact line.

Two details found by testing rather than assumption:

1. `paths` matches the **exact** specifier only — it does *not* catch `marked/lib/marked.esm.js`, which is precisely the path the Rollup warning names. Hence `patterns: ["marked/*", "dompurify/*"]`. Verified `paths` alone finds nothing there.
2. `grep -c` exits **1** when the count is 0, so the obvious form would have failed on the success case. The script uses `! grep -q` with `set -euo pipefail`.

Both `paths` and `patterns` are static-only: with the rule active, bare `import("marked")` and subpath `import("dompurify/dist/purify.es.mjs")` stay clean, so the seven legitimate call sites are untouched.

## Measured gate coverage

I originally claimed the build gate "covers `.ts` importers." **That was reasoning, not measurement, and it was wrong.** Measured:

| Static import added to | Rollup warns? |
|---|---|
| `GitRail.svelte` | yes |
| `ui/src/lib/recaps.svelte.ts` | yes |
| `ui/src/lib/api.ts` (plain `.ts`) | **no — none at all** |

That left the plain-`.ts` case covered by neither gate. Closed in fe90932 by scoping the eslint rule to `ui/src/**/*` rather than just the Svelte block (one definition, not one per glob). The two gates are **complementary, neither a superset of the other**:

- `check-ui-build.sh` — **any** library, but only where Rollup actually warns
- `no-restricted-imports` — **any** file under `ui/src`, but only the named libraries (marked, dompurify)

Verified after the change: probes in `GitRail.svelte`, `recaps.svelte.ts` **and** `api.ts` each fail lint; clean tree passes.

Residual hole, uncovered by design and documented in the script header: a *different* library statically imported from a plain `.ts` where Rollup stays silent. Add it to the rule's `paths`/`patterns` if one ever matters.

## Rejected approach (documented in the script)

Failing inside the build via `ui/vite.config.ts` would have been strictly better — CI, pre-push and a bare local `bun run build` would all inherit it with zero sync burden. **Neither in-build hook works here:** `build.rollupOptions.onwarn` is overridden by SvelteKit's config merge, and an `onwarn` returned from a plugin `options` hook is ignored by Vite (instrumented: the plugin's hooks run, but collected `offenders: 0` while the warning still printed). Both left the build exiting 0 — an inert gate that looks like it works.

## Verification

- `bun run lint`, `bun run typecheck`, `cd ui && bun run check` (0 errors), `bun run test` (4006 pass / 262 files), root `bun test` (7291 pass / 0 fail), prettier — all clean. Pre-push passed with the new script in the ui lane.
- `GitRail.browser.test.ts`: 74 pass, **test file unmodified**.
- **Both gates proven to bite:** reintroducing a static import makes lint exit 1 *and* `check-ui-build.sh` exit 1 with a readable message. Probes reverted, confirmed byte-identical (md5) / clean `git status`.

## Reviewer notes

- `GitRail.browser.test.ts` passes **identically before and after** — it observes rendering, not chunking, so it is a sanitization guard, **not** evidence the deliverable landed.
- Behavior change: the review body renders one microtask later (empty until the chunk resolves). Identical to the seven siblings.
- On a chunk-fetch failure the `.catch` only warns and never assigns, so no unsanitized markup can reach `@html`; the early-return still clears `renderedBody` synchronously when the popover closes.
- Self-review caught the `.catch((err) => …)` parameter shadowing GitRail's `err` `$state` (line 76, drives the error/Retry UI). Renamed to `e`, matching the file's convention.
- The build gate is intentionally broader than this change and will fire for any library — an unrelated future PR can trip it. The script names the fix.
- `[no-feature-entry]` — ships no user-facing UX. No new strings, so no i18n/glossary work.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,7 +147,17 @@ jobs:
         run: cd extension && bun run test
 
       - name: Build (ui)
-        run: cd ui && bun run build
+        # Rollup reports a static import that defeats a dynamic one
+        # (INEFFECTIVE_DYNAMIC_IMPORT) as a WARNING, which does not fail the build —
+        # so assert on it explicitly. One static `import { marked } from "marked"` in
+        # an always-mounted component hoists the lib into the main chunk and silently
+        # cancels the lazy-loading in every other consumer. pipefail so a genuine
+        # build failure still fails the step; `! grep -q` (not `grep -c`, which exits
+        # 1 when the count is 0) so the success case exits 0.
+        run: |
+          set -eo pipefail
+          cd ui && bun run build 2>&1 | tee /tmp/ui-build.log
+          ! grep -q INEFFECTIVE_DYNAMIC_IMPORT /tmp/ui-build.log
 
       - name: Build (extension)
         run: cd extension && bun run build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -147,17 +147,10 @@ jobs:
         run: cd extension && bun run test
 
       - name: Build (ui)
-        # Rollup reports a static import that defeats a dynamic one
-        # (INEFFECTIVE_DYNAMIC_IMPORT) as a WARNING, which does not fail the build —
-        # so assert on it explicitly. One static `import { marked } from "marked"` in
-        # an always-mounted component hoists the lib into the main chunk and silently
-        # cancels the lazy-loading in every other consumer. pipefail so a genuine
-        # build failure still fails the step; `! grep -q` (not `grep -c`, which exits
-        # 1 when the count is 0) so the success case exits 0.
-        run: |
-          set -eo pipefail
-          cd ui && bun run build 2>&1 | tee /tmp/ui-build.log
-          ! grep -q INEFFECTIVE_DYNAMIC_IMPORT /tmp/ui-build.log
+        # Builds, and fails on INEFFECTIVE_DYNAMIC_IMPORT (a static import that defeats
+        # a dynamic one — Rollup only warns, so the build would otherwise stay green).
+        # Same script as the pre-push hook's ui build lane, so the two stay in sync.
+        run: ./scripts/check-ui-build.sh
 
       - name: Build (extension)
         run: cd extension && bun run build
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -81,7 +81,7 @@ so a hung child can never wedge the push. The checks (per lane) are:
 - **eslint:** root + extension eslint over the push **delta** (see note)
 - **tsc:** `bun run typecheck` (root `tsc --noEmit`)
 - **root-tests:** `bun test ./test`
-- **ui:** `bun run check` → `check:i18n` → `playwright install chromium` → `bun run test` → `bun run build`
+- **ui:** `bun run check` → `check:i18n` → `playwright install chromium` → `bun run test` → `scripts/check-ui-build.sh` (the ui build, plus a fail on Rollup's `INEFFECTIVE_DYNAMIC_IMPORT`; CI's **Build (ui)** step runs the same script)
 - **ext:** `bun run check` → `check:i18n` → `bun run test` → `bun run build`
 - then **`bunx fallow@2.100.0 audit --base origin/main --fail-on-issues`** (delta
   dead-code/complexity audit; version pinned — see note below)
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,12 +76,12 @@ and failed pushes. Independent work runs in parallel (bounded by your core count
 lane teeing to its own `.test-logs/` file, with a per-lane wall-clock **timeout backstop**
 so a hung child can never wedge the push. The checks (per lane) are:
 
-- **gates:** branch-hygiene · feature-catalog · generated-docs · glossary
+- **gates:** branch-hygiene · feature-catalog · generated-docs · glossary · announcement-versions · herdr-types
 - **prettier:** `prettier --check` over the push **delta** (see note)
 - **eslint:** root + extension eslint over the push **delta** (see note)
 - **tsc:** `bun run typecheck` (root `tsc --noEmit`)
 - **root-tests:** `bun test ./test`
-- **ui:** `bun run check` → `check:i18n` → `playwright install chromium` → `bun run test` → `scripts/check-ui-build.sh` (the ui build, plus a fail on Rollup's `INEFFECTIVE_DYNAMIC_IMPORT`; CI's **Build (ui)** step runs the same script)
+- **ui:** `bun run check` → `check:i18n` → `check:docs-manifest` → `playwright install chromium` → `bun run test` → `scripts/check-ui-build.sh` (the ui build, plus a fail on Rollup's `INEFFECTIVE_DYNAMIC_IMPORT`; CI's **Build (ui)** step runs the same script)
 - **ext:** `bun run check` → `check:i18n` → `bun run test` → `bun run build`
 - then **`bunx fallow@2.100.0 audit --base origin/main --fail-on-issues`** (delta
   dead-code/complexity audit; version pinned — see note below)
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -61,6 +61,34 @@ export default [
         extraFileExtensions: [".svelte"],
       },
     },
+    rules: {
+      // marked + DOMPurify are heavy and are lazily imported by every consumer that
+      // renders markdown. A single STATIC import in an always-mounted component
+      // hoists them into the main chunk and silently defeats all of them, so ban the
+      // static form here. Dynamic `await import("marked")` is unaffected — both
+      // `paths` and `patterns` only match static imports.
+      "no-restricted-imports": [
+        "error",
+        {
+          paths: [
+            {
+              name: "marked",
+              message:
+                'Import dynamically — see SessionRecap.svelte: Promise.all([import("marked"), import("dompurify")]).',
+            },
+            {
+              name: "dompurify",
+              message:
+                'Import dynamically — see SessionRecap.svelte: Promise.all([import("marked"), import("dompurify")]).',
+            },
+          ],
+          // `paths` matches the exact specifier only, so it would miss a deep import
+          // like "marked/lib/marked.esm.js" — precisely the path Rollup names in the
+          // INEFFECTIVE_DYNAMIC_IMPORT warning this rule exists to prevent.
+          patterns: ["marked/*", "dompurify/*"],
+        },
+      ],
+    },
   },
   // Generated output — never lint
   {
--- a/eslint.config.js
+++ b/eslint.config.js
@@ -61,12 +61,24 @@ export default [
         extraFileExtensions: [".svelte"],
       },
     },
+  },
+  // marked + DOMPurify are heavy and are lazily imported by every consumer that renders
+  // markdown. A single STATIC import hoists them into a shared chunk and silently
+  // defeats all of them (Rollup: INEFFECTIVE_DYNAMIC_IMPORT) — that is what GitRail did
+  // to seven other components.
+  //
+  // Scoped to ALL of ui/src, not just the Svelte files, because Rollup does not warn for
+  // every static importer: a static import from a plain `ui/src/lib/*.ts` helper produces
+  // NO warning at all, so scripts/check-ui-build.sh cannot see it (measured — see the
+  // table in that script's header). For those files this rule is the ONLY gate.
+  //
+  // One block rather than the rule repeated per-glob, so there is a single definition to
+  // keep correct.
+  {
+    files: ["ui/src/**/*"],
     rules: {
-      // marked + DOMPurify are heavy and are lazily imported by every consumer that
-      // renders markdown. A single STATIC import in an always-mounted component
-      // hoists them into the main chunk and silently defeats all of them, so ban the
-      // static form here. Dynamic `await import("marked")` is unaffected — both
-      // `paths` and `patterns` only match static imports.
+      // Dynamic `await import("marked")` is unaffected — both `paths` and `patterns`
+      // only match static imports.
       "no-restricted-imports": [
         "error",
         {
--- a/scripts/check-ui-build.sh
+++ b/scripts/check-ui-build.sh
@@ -11,8 +11,22 @@
 #   - `.github/workflows/ci.yml` and `scripts/pre-push.ts` must stay in sync (see the
 #     job header at ci.yml:24-29). Both call this, so there is ONE implementation to
 #     keep correct rather than two copies of the same shell.
-#   - It also covers the case the Svelte-scoped `no-restricted-imports` lint rule
-#     provably cannot: a static import from a plain `ui/src/lib/*.ts` helper.
+#   - It is not library-specific: the `no-restricted-imports` rule in eslint.config.js
+#     names marked/dompurify, while this fires for ANY module that regresses this way.
+#
+# Known hole — do NOT assume this gate is exhaustive. Rollup emits the warning only for
+# some static importers, so a static import is caught in some files and silently
+# ignored in others. Measured on this repo:
+#
+#   ui/src/lib/components/GitRail.svelte  -> warns      (caught here + by eslint)
+#   ui/src/lib/recaps.svelte.ts           -> warns      (caught here + by eslint)
+#   ui/src/lib/api.ts                     -> NO warning (caught by NEITHER)
+#
+# A static import from a plain `.ts` helper can therefore defeat lazy-loading with both
+# gates green, because Rollup itself stays silent. That case is uncovered; review is the
+# only thing standing behind it. (An earlier version of this comment claimed the
+# opposite — that this script covered the plain-`.ts` case the Svelte-scoped lint rule
+# cannot see. That was inference; the measurement above disproved it.)
 #
 # Note this deliberately does NOT live in ui/vite.config.ts. Both in-build hooks were
 # tried and neither works: `build.rollupOptions.onwarn` is overridden by SvelteKit's
--- a/scripts/check-ui-build.sh
+++ b/scripts/check-ui-build.sh
@@ -43,8 +43,9 @@ trap 'rm -f "$log"' EXIT
 # swallowed by the pipe into tee.
 cd "$repo_root/ui" && bun run build 2>&1 | tee "$log"
 
-# `! grep -q` — NOT `grep -c`, which exits 1 when the count is 0 and would therefore
-# fail on the success case.
+# `grep -q` as the `if` condition, with the failure raised explicitly below — NOT
+# `grep -c`, whose exit status is 1 when the count is 0 and would therefore report
+# failure on the success case.
 if grep -q INEFFECTIVE_DYNAMIC_IMPORT "$log"; then
   echo ""
   echo "✗ INEFFECTIVE_DYNAMIC_IMPORT in the ui build:"
--- a/scripts/check-ui-build.sh
+++ b/scripts/check-ui-build.sh
@@ -14,19 +14,24 @@
 #   - It is not library-specific: the `no-restricted-imports` rule in eslint.config.js
 #     names marked/dompurify, while this fires for ANY module that regresses this way.
 #
-# Known hole — do NOT assume this gate is exhaustive. Rollup emits the warning only for
-# some static importers, so a static import is caught in some files and silently
-# ignored in others. Measured on this repo:
+# Do NOT assume this gate alone is exhaustive: Rollup emits the warning only for SOME
+# static importers. Measured on this repo:
 #
-#   ui/src/lib/components/GitRail.svelte  -> warns      (caught here + by eslint)
-#   ui/src/lib/recaps.svelte.ts           -> warns      (caught here + by eslint)
-#   ui/src/lib/api.ts                     -> NO warning (caught by NEITHER)
+#   ui/src/lib/components/GitRail.svelte  -> warns
+#   ui/src/lib/recaps.svelte.ts           -> warns
+#   ui/src/lib/api.ts  (plain .ts)        -> NO warning at all
 #
-# A static import from a plain `.ts` helper can therefore defeat lazy-loading with both
-# gates green, because Rollup itself stays silent. That case is uncovered; review is the
-# only thing standing behind it. (An earlier version of this comment claimed the
-# opposite — that this script covered the plain-`.ts` case the Svelte-scoped lint rule
-# cannot see. That was inference; the measurement above disproved it.)
+# So the two gates are complementary, and neither is a superset of the other:
+#
+#   this script            — ANY library, but only where Rollup actually warns
+#   no-restricted-imports  — ANY file under ui/src, but only the named libraries
+#                            (marked, dompurify)
+#
+# The plain-`.ts` case above is covered by the lint rule, which is scoped to `ui/src/**/*`
+# precisely because this script cannot see it. Residual hole, uncovered by design: a
+# DIFFERENT library, statically imported from a plain `.ts` where Rollup stays silent —
+# add it to the rule's `paths`/`patterns` if one ever matters. Review is what stands
+# behind that case.
 #
 # Note this deliberately does NOT live in ui/vite.config.ts. Both in-build hooks were
 # tried and neither works: `build.rollupOptions.onwarn` is overridden by SvelteKit's
--- a/scripts/check-ui-build.sh
+++ b/scripts/check-ui-build.sh
@@ -1,0 +1,43 @@
+#!/usr/bin/env bash
+# Build the UI and fail on Rollup's INEFFECTIVE_DYNAMIC_IMPORT.
+#
+# A STATIC import of a module that other modules import dynamically silently cancels
+# their lazy-loading: Rollup hoists it into a shared chunk for every consumer. That is
+# exactly what happened with marked/DOMPurify — one static import in the always-mounted
+# GitRail defeated the dynamic imports in seven other components, and because Rollup
+# only WARNS, the build stayed green the whole time.
+#
+# Why a script instead of inlining the check:
+#   - `.github/workflows/ci.yml` and `scripts/pre-push.ts` must stay in sync (see the
+#     job header at ci.yml:24-29). Both call this, so there is ONE implementation to
+#     keep correct rather than two copies of the same shell.
+#   - It also covers the case the Svelte-scoped `no-restricted-imports` lint rule
+#     provably cannot: a static import from a plain `ui/src/lib/*.ts` helper.
+#
+# Note this deliberately does NOT live in ui/vite.config.ts. Both in-build hooks were
+# tried and neither works: `build.rollupOptions.onwarn` is overridden by SvelteKit's
+# config merge, and an `onwarn` returned from a plugin `options` hook is ignored by
+# Vite (the plugin's hooks run, but the handler is never consulted). In both cases the
+# warning still printed and the build still exited 0 — a silently inert gate.
+set -euo pipefail
+
+repo_root="$(cd "$(dirname "${BASH_SOURCE[0]}")/.." && pwd)"
+log="$(mktemp -t shepherd-ui-build.XXXXXX.log)"
+trap 'rm -f "$log"' EXIT
+
+# pipefail (set above) makes a genuine build failure fail the script rather than being
+# swallowed by the pipe into tee.
+cd "$repo_root/ui" && bun run build 2>&1 | tee "$log"
+
+# `! grep -q` — NOT `grep -c`, which exits 1 when the count is 0 and would therefore
+# fail on the success case.
+if grep -q INEFFECTIVE_DYNAMIC_IMPORT "$log"; then
+  echo ""
+  echo "✗ INEFFECTIVE_DYNAMIC_IMPORT in the ui build:"
+  grep INEFFECTIVE_DYNAMIC_IMPORT "$log" || true
+  echo ""
+  echo "  A static import is defeating a dynamic one, forcing the module into a shared"
+  echo "  chunk for every consumer. Import it dynamically instead — see SessionRecap.svelte:"
+  echo "    Promise.all([import(\"marked\"), import(\"dompurify\")])"
+  exit 1
+fi
--- a/scripts/pre-push.ts
+++ b/scripts/pre-push.ts
@@ -562,9 +562,10 @@ function buildLanes(
       { label: "vitest", cmd: "bun", args: ["run", "test", "--maxWorkers", W], cwd: ui },
       // Same script CI's "Build (ui)" step runs (ci.yml:24-29 requires the two to stay
       // in sync): builds, then fails on Rollup's INEFFECTIVE_DYNAMIC_IMPORT — a static
-      // import that defeats a dynamic one, for any module. Not exhaustive: Rollup does
-      // not warn for every static importer, so some (e.g. a plain ui/src/lib/*.ts
-      // helper) slip past this AND eslint. See the header in check-ui-build.sh.
+      // import that defeats a dynamic one, for any module. Rollup does not warn for
+      // every static importer (a plain ui/src/lib/*.ts helper produces none), so the
+      // eslint no-restricted-imports rule covers those files for the named libraries.
+      // The two are complementary; see the header in check-ui-build.sh.
       { label: "build", cmd: join(repoRoot, "scripts/check-ui-build.sh"), args: [], cwd: ui },
     ],
   });
--- a/scripts/pre-push.ts
+++ b/scripts/pre-push.ts
@@ -560,10 +560,11 @@ function buildLanes(
         cwd: ui,
       },
       { label: "vitest", cmd: "bun", args: ["run", "test", "--maxWorkers", W], cwd: ui },
-      // Same script CI's "Build (ui)" step runs: builds, then fails on Rollup's
-      // INEFFECTIVE_DYNAMIC_IMPORT. Catches a static import that defeats a dynamic one
-      // — including from a plain ui/src/lib/*.ts helper, which the Svelte-scoped
-      // no-restricted-imports rule cannot see.
+      // Same script CI's "Build (ui)" step runs (ci.yml:24-29 requires the two to stay
+      // in sync): builds, then fails on Rollup's INEFFECTIVE_DYNAMIC_IMPORT — a static
+      // import that defeats a dynamic one, for any module. Not exhaustive: Rollup does
+      // not warn for every static importer, so some (e.g. a plain ui/src/lib/*.ts
+      // helper) slip past this AND eslint. See the header in check-ui-build.sh.
       { label: "build", cmd: join(repoRoot, "scripts/check-ui-build.sh"), args: [], cwd: ui },
     ],
   });
--- a/scripts/pre-push.ts
+++ b/scripts/pre-push.ts
@@ -560,7 +560,11 @@ function buildLanes(
         cwd: ui,
       },
       { label: "vitest", cmd: "bun", args: ["run", "test", "--maxWorkers", W], cwd: ui },
-      { label: "build", cmd: "bun", args: ["run", "build"], cwd: ui },
+      // Same script CI's "Build (ui)" step runs: builds, then fails on Rollup's
+      // INEFFECTIVE_DYNAMIC_IMPORT. Catches a static import that defeats a dynamic one
+      // — including from a plain ui/src/lib/*.ts helper, which the Svelte-scoped
+      // no-restricted-imports rule cannot see.
+      { label: "build", cmd: join(repoRoot, "scripts/check-ui-build.sh"), args: [], cwd: ui },
     ],
   });
 
--- a/ui/src/lib/components/GitRail.svelte
+++ b/ui/src/lib/components/GitRail.svelte
@@ -27,8 +27,6 @@
   import Coachmark from "$lib/components/Coachmark.svelte";
   import { pollWhileVisible } from "$lib/visibility";
   import { pullMainAndToast } from "$lib/pull-offer";
-  import { marked } from "marked";
-  import createDOMPurify from "dompurify";
 
   let {
     sessionId,
@@ -83,14 +81,6 @@
   // "no forge / no PR"). Drives the error+Retry fallback so the rail never renders
   // silently empty, and keeps the poll retrying until the forge recovers.
   let loadFailed = $state(false);
-  type DomPurifyApi = { sanitize: (dirty: string) => string };
-  type DomPurifyFactory = ((root: Window) => DomPurifyApi) & Partial<DomPurifyApi>;
-  function getDOMPurify(): DomPurifyApi {
-    const candidate = createDOMPurify as unknown as DomPurifyFactory;
-    return typeof candidate.sanitize === "function"
-      ? (candidate as DomPurifyApi)
-      : candidate(window);
-  }
 
   // Open-PR popover
   let showPr = $state(false);
@@ -482,6 +472,9 @@
           : m.gitrail_review_plan(),
   );
   // Render the (AI-authored) findings as markdown, sanitized before @html.
+  // Dynamically imported so marked/DOMPurify stay off the first-paint critical path;
+  // a static import here would hoist both into the main chunk and defeat the same
+  // lazy-loading in every other consumer (SessionRecap, PlanPanel, ...).
   // Gated on showReview so the browser-only sanitizer never runs during SSR.
   let renderedBody = $state("");
   $effect(() => {
@@ -490,8 +483,20 @@
       renderedBody = "";
       return;
     }
-    const DOMPurify = getDOMPurify();
-    renderedBody = DOMPurify.sanitize(marked.parse(body, { async: false }) as string);
+    let alive = true;
+    Promise.all([import("marked"), import("dompurify")])
+      .then(([{ marked }, { default: DOMPurify }]) => {
+        if (alive)
+          renderedBody = DOMPurify.sanitize(marked.parse(body, { async: false }) as string);
+      })
+      .catch((e) => {
+        // Never assigns to renderedBody — a failed load must not leave unsanitized
+        // markup, and the `err` state above is the PR-action error channel, not this.
+        console.warn("Review body markdown render failed", e);
+      });
+    return () => {
+      alive = false;
+    };
   });
   const autoCount = $derived(automationCount(repoConfig.flags(repoPath)));
   let reviewFlash = $state<string | null>(null);


scoop (commented):
## What I verified

**The fix itself is correct and byte-for-byte consistent with its siblings.**
The new `$effect` at `ui/src/lib/components/GitRail.svelte:480-500` is structurally identical to the reference implementation at `ui/src/lib/components/SessionRecap.svelte:23-41` — same `Promise.all([import("marked"), import("dompurify")])`, same `[{ marked }, { default: DOMPurify }]` destructure, same `alive` guard, same cleanup returning `alive = false`, same warn-only `.catch`. Compared line by line against `SessionRecap.svelte:29-41`, `DoneRecapPanel.svelte:65-68`, `PlanPanel.svelte:128-132`, `blocks/RichTextBlock.svelte:12-14`, `blocks/CalloutBlock.svelte:37-39`.

**No dangling references to the removed symbols.** Grepped the whole tree for `DomPurifyApi|DomPurifyFactory|getDOMPurify|createDOMPurify` — zero matches anywhere, so the type/helper removal is complete (their only consumer was the render effect, as the plan claimed).

**Dropping `getDOMPurify()` is safe on the path that actually executes.** `ui/package.json:50` pins `dompurify: ^3.4.12`; in v3 the ESM default export is already a sanitize-capable instance when `window` exists. The effect is gated on `showReview` (`GitRail.svelte:481`) and `$effect` never runs during SSR, so the factory branch the helper defended was unreachable. Seven sibling call sites are the standing evidence.

**Sanitization cannot regress.** `renderedBody` is only ever assigned sanitized output (`:490`); the early return clears it (`:483`); the `.catch` assigns nothing (`:492-496`). The render site `GitRail.svelte:824-827` is `{#if verdict.body}` around `{@html renderedBody}`, so a failed chunk load yields an empty div, never raw markup.

**The lint rule cannot over-fire today.** Grepped the entire repo for `^\s*import .*["'](marked|dompurify)["'/]` — **zero** static importers remain after this diff. All eight consumers use `await import(...)` / `Promise.all([import(...)])` (`HerdrUpdateModal.svelte:52`, `blocks/WireframeBlock.svelte:132`, etc.). The rule's `paths` block is scoped to `["**/*.svelte", "**/*.svelte.ts", "**/*.svelte.js"]` (`eslint.config.js:50`), and root lint covers `ui/src` (`package.json:10`). The `message` points at `SessionRecap.svelte`, which does contain the referenced pattern (`SessionRecap.svelte:30`) — the citation is accurate, not aspirational.

**The CI shell semantics are right.** The job runs on `ubuntu-latest` (`.github/workflows/ci.yml:46`) with no `defaults.run.shell` override, so GitHub's `bash -e {0}` applies and `set -o pipefail` is valid. `|` binds tighter than `&&`, so `cd ui && bun run build 2>&1 | tee ...` pipes only the build, and pipefail propagates a genuine build failure. `! grep -q` is the correct inversion (exit 0 when absent) — the comment's warning about `grep -c` exiting 1 on a zero count is accurate. `/tmp/ui-build.log` is absolute, so the persisted `cd ui` on the following line is harmless.

**The broad build gate should not trip on unrelated libraries.** The only other mixed static/dynamic pairs I found are `mermaid` (static at `ui/src/lib/components/blocks/MermaidBlock.browser.test.ts:19`, dynamic at `MermaidBlock.svelte:33`) and `@pierre/diffs` (static at `ui/src/lib/components/PierreDiff.browser.test.ts:3`, dynamic at `PierreDiff.svelte:99` and `ui/src/lib/pierre-diff.ts:20,32`) — both static sides live in `*.test.ts`, which are not in the `vite build` module graph. `shiki` is static-only (`ui/src/lib/highlight.ts:6`).

**The existing browser test tolerates the now-async render** without edits: `GitRail.browser.test.ts:999-1004` wraps the `.rv-body a` assertion in `vi.waitFor(..., { timeout: 5000 })`. (Its comment at `:987-988` already described the render as dynamically imported — this PR makes that comment true.)

## Stated limitations (could not verify, not asserted either way)

- `node_modules/` is absent in this review worktree, so I could not open ESLint 10.7.0's `no-restricted-imports` source to confirm first-hand that neither `paths` nor `patterns` matches `ImportExpression` (dynamic `import()`). If it did, all seven legitimate dynamic sites would error and lint would fail loudly — a noisy failure, not a silent one. The plan reports empirically verifying this against the real Svelte parser and the actual components; I am recording it as unverified here rather than as a defect.
- I did not run `bun run build`, so I cannot independently confirm the log emits zero `INEFFECTIVE_DYNAMIC_IMPORT` lines after the fix — only that no other importer in the build graph looks capable of producing one.

## Finding

`.github/workflows/ci.yml:149-160` hardens `Build (ui)`, but `scripts/pre-push.ts:563` still runs the bare `{ label: "build", cmd: "bun", args: ["run", "build"], cwd: ui }`. The job header at `.github/workflows/ci.yml:24-29` explicitly requires the two to stay in sync and enumerates the *deliberate* differences (delta-scoped prettier/eslint, pinned fallow version); this PR adds an undocumented third one. Concretely: the new Svelte-scoped lint rule cannot see a static import from a plain `ui/src/lib/*.ts` helper — which the plan itself names as the hole the build gate exists to close — so exactly that regression now passes pre-push and only reds in CI. Either mirror the assertion into the ui lane's build step or list it under the header's deliberate-differences note.

## Notes (non-blocking, no action needed)

- The base ESLint `no-restricted-imports` has no `allowTypeImports`, so a future `import type { Tokens } from "marked"` would be flagged even though a type-only import is erased and creates no chunk edge. No such site exists today (verified by the repo-wide grep above), and the plan accepts the over-fire trade-off; `@typescript-eslint/no-restricted-imports` is the escape hatch if it ever bites.
- Minor behavioural delta vs. the old synchronous render: when `verdict.body` changes while the popover is open, the previous (already sanitized) HTML stays on screen until the new render resolves instead of being replaced synchronously. After first load the modules are cached, so this is a microtask. Identical to all seven siblings.
- Diff touches exactly the three planned files; no scope creep observed.

scoop:
**Finding 1 (ci.yml / pre-push.ts sync) — valid, fixed.** The assertion landed only in CI while `scripts/pre-push.ts:563` still ran a bare `bun run build`, which the `ci.yml:24-29` header requires to stay in sync. Fixed in 31d8934 by extracting `scripts/check-ui-build.sh` and calling it from **both**, so there's one implementation rather than two copies of the same shell (matching the repo's existing `scripts/check-*` gate convention). Verified: the script exits 1 with a readable message on a reintroduced static import, exits 0 clean, and the pre-push ui lane now runs it.

Two things worth recording, because both correct claims I made in the original PR body:

**1. The stated rationale doesn't hold — I measured it.** The finding says a static import from a plain `ui/src/lib/*.ts` helper "passes pre-push and only fails in CI." It doesn't fail in CI either: Rollup does not emit `INEFFECTIVE_DYNAMIC_IMPORT` for that case at all.

| Static import added to | Rollup warns? | Caught by |
|---|---|---|
| `GitRail.svelte` | yes | build gate + lint |
| `ui/src/lib/recaps.svelte.ts` | yes (3) | build gate + lint (`**/*.svelte.ts` is in the rule's glob) |
| `ui/src/lib/api.ts` (plain `.ts`) | **no** | **neither** |

So the plain-`.ts` hole is real but **no gate closes it**, because Rollup itself is silent there. My original PR body claimed the build gate "covers `.ts` importers" and that it closed the lint rule's Svelte-scoping hole — that was reasoning, not measurement, and it was wrong. Body updated. The gate's actual value is the CI/pre-push sync plus coverage of any library and any warning-producing site.

**2. I tried the better-looking fix first and it silently didn't work.** Before mirroring the script I attempted to fail inside the build (`ui/vite.config.ts`), which would have made CI, pre-push and a bare local `bun run build` all inherit the gate with no sync burden at all. **Neither in-build hook works here:**

- `build.rollupOptions.onwarn` — overridden by SvelteKit's config merge; never invoked.
- `onwarn` returned from a plugin `options` hook — Vite ignores it. Instrumented and confirmed the plugin's `options` and `buildEnd` hooks *do* run, but collected `offenders: 0` while the warning still printed.

In both cases the warning printed and the build exited 0 — an inert gate that looks like it works. I only caught it because I tested the failure path rather than trusting a green clean build. Rejected approach and reasoning are documented in `scripts/check-ui-build.sh` so the next person doesn't retry it.

scoop (commented):
## Previous-round finding

**1. ci.yml hardened step not mirrored in `scripts/pre-push.ts` — ADDRESSED.** Verified, not taken on the note's say-so: `.github/workflows/ci.yml:150` now runs `./scripts/check-ui-build.sh`, and `scripts/pre-push.ts:567` runs `{ label: "build", cmd: join(repoRoot, "scripts/check-ui-build.sh"), args: [], cwd: ui }`. One implementation, both callers — the sync requirement at `.github/workflows/ci.yml:24-29` is satisfied. Not re-raised.

## What I verified

**The fix itself.** `ui/src/lib/components/GitRail.svelte:483-495` now matches the sibling pattern byte-for-byte in structure against `ui/src/lib/components/SessionRecap.svelte:28-40` — same `let alive = true`, same `Promise.all([import("marked"), import("dompurify")])`, same `{ marked }` / `{ default: DOMPurify }` destructure, same `if (alive)` guard, same `.catch` that only warns, same `return () => { alive = false }`. The `.catch` never assigns, so a failed chunk load cannot leave unsanitized markup at the `{@html renderedBody}` site (`ui/src/lib/components/GitRail.svelte:826`).

**Effect cleanup on the early-return path.** `ui/src/lib/components/GitRail.svelte:484-487` returns `undefined` when `body` is falsy, but Svelte runs the *previous* run's returned teardown before re-running the effect, so the in-flight promise's `alive` is still flipped. No stale-write window.

**No dangling references to the removed helper.** Grepped `ui/src` for `from "marked"` / `from "dompurify"` — zero matches, so the static importer is genuinely gone and `createDOMPurify`, `DomPurifyApi`, `DomPurifyFactory` have no surviving consumers (the only other `marked` hit in the file is the unrelated i18n key `m.prbadge_marked_draft()` at `ui/src/lib/components/GitRail.svelte:375`).

**The async render does not break the existing browser test.** `ui/src/lib/components/GitRail.browser.test.ts:999-1004` already wraps the `.rv-body a` assertion in `vi.waitFor(..., { timeout: 5000 })`, so the now-async assignment is tolerated. No test-file edit was needed, consistent with the plan.

**Shell semantics of the new gate.** In `scripts/check-ui-build.sh`: `|` binds tighter than `&&`, so line 30 parses as `cd ui && (bun run build 2>&1 | tee "$log")`, and `pipefail` (line 22) makes a genuine build failure propagate rather than be masked by `tee`. `grep -q` inside the `if` condition is exempt from `set -e`, and the diagnostic `grep` at line 37 is guarded with `|| true`. Falls through to exit 0 when clean. `repo_root` is derived from `BASH_SOURCE` (line 24), so it resolves correctly from either caller's cwd, including inside a git worktree. File mode is `100755` in the diff, and `scripts/pre-push.ts:258` spawns without `shell: true`, so direct exec of the shebang works.

**CI invocation path.** `.github/workflows/ci.yml:150` uses a repo-root-relative `./scripts/check-ui-build.sh`; the `verify` job (`.github/workflows/ci.yml:22-64`) declares no `defaults.run.working-directory`, so the step runs from the workspace root. Correct.

**Deps exist as named.** `dompurify` and `marked` are declared at `ui/package.json:50-51`.

**Scope.** Five files, all traceable to the task or to the previous round's finding. No user-facing strings added, so no i18n parity surface to check.

- `node_modules` is not installed in this worktree, so I could not resolve ESLint's `no-restricted-imports` implementation to confirm the claim at `eslint.config.js:84-86` that `patterns: ["marked/*", "dompurify/*"]` matches a deep specifier such as `marked/lib/marked.esm.js`. ESLint matches `patterns` with gitignore-style globbing, where `*` does not cross `/` — under those semantics `marked/*` would match `marked/lib` but not `marked/lib/marked.esm.js`, which would make that comment wrong and leave the deep-import case to the build gate alone. The plan claims this was empirically verified; I could neither confirm nor refute it here. `patterns: ["marked/**", "dompurify/**"]` would be unambiguous under either reading. Assumed as written.
- I did not run the build (read-only review), so I could not independently adjudicate whether Rollup emits `INEFFECTIVE_DYNAMIC_IMPORT` for a static import from a plain `.ts` file. See the finding below, which rests on an internal contradiction within the PR rather than on my own measurement.

## Finding

**`scripts/check-ui-build.sh:14-15` and `scripts/pre-push.ts:563-566` state a coverage property the author's own note reports measuring as false.** Both comments claim the script covers "a static import from a plain `ui/src/lib/*.ts` helper" — the script's header offers it as one of the two reasons the script exists. The PR note accompanying this revision states the opposite, from a measurement: Rollup emits no `INEFFECTIVE_DYNAMIC_IMPORT` for that case at all, so neither gate closes that hole, and the note says the PR body was corrected accordingly. These two in-code comments were not. One of the two is wrong, and the comment is the copy a future maintainer will read at the moment they reason about what is and isn't protected — a reader trusting it would believe a plain-`.ts` static import is gated when (per the author) nothing gates it. Reconcile: either correct both comments to describe the gate's actual value (CI/pre-push sync, plus any library and any warning-producing site), or retract the note's measurement.

scoop:
**Findings 1 & 2 — both valid, fixed in 87a5832.** I corrected the PR body after measuring this but left the disproven claim in the two code comments, which is the worse place to leave it: a PR body is read once, a comment in `check-ui-build.sh` is what the next person reasons from.

`scripts/check-ui-build.sh:14-15` and `scripts/pre-push.ts:563-566` now describe the gate's actual value — CI/pre-push sync, plus firing for *any* module rather than just the marked/dompurify pair the eslint rule names — and the script header records the measured hole rather than implying coverage:

```
#   ui/src/lib/components/GitRail.svelte  -> warns      (caught here + by eslint)
#   ui/src/lib/recaps.svelte.ts           -> warns      (caught here + by eslint)
#   ui/src/lib/api.ts                     -> NO warning (caught by NEITHER)
```

It states plainly that a static import from a plain `.ts` helper can defeat lazy-loading with both gates green, and that review is the only thing behind that case. I also left a one-line note that the earlier comment claimed the opposite and what disproved it, so the correction doesn't quietly regress later.

Grepped the rest of the diff for the same claim — no other instances. Comments only, no behavior change; `check-ui-build.sh` still exits 0 clean, and lint/typecheck/prettier/pre-push all pass.

scoop (commented):

**1. `scripts/check-ui-build.sh:14-15` — disproven "covers a plain `.ts` helper" claim. FIXED.**
The header now reads the opposite of the old claim: lines 17-29 are titled `Known hole — do NOT assume this gate is exhaustive`, record the measurement per-file (`ui/src/lib/api.ts -> NO warning (caught by NEITHER)`, `scripts/check-ui-build.sh:23`), state plainly that "a static import from a plain `.ts` helper can therefore defeat lazy-loading with both gates green" (`:25-27`), and explicitly retract the earlier inference (`:27-29`). The remaining stated value — one implementation shared by CI and pre-push, and firing for any module rather than only the eslint-named pair — is at `:10-15` and is accurate: `.github/workflows/ci.yml:153` and `scripts/pre-push.ts:568` both invoke this same script, and the grep at `scripts/check-ui-build.sh:48` is specifier-agnostic while `eslint.config.js:73-88` names only marked/dompurify. Not re-raised.

**2. `scripts/pre-push.ts:563-566` — same claim. FIXED.**
The replacement comment at `scripts/pre-push.ts:563-567` now says "Not exhaustive: Rollup does not warn for every static importer, so some (e.g. a plain ui/src/lib/*.ts helper) slip past this AND eslint." Consistent with the script header it points to. Not re-raised.

## Verification of the change itself

- **No dangling references to the removed helper.** Grepped `ui/src` for `createDOMPurify|DomPurifyApi|DomPurifyFactory` — zero matches, so the removal at `ui/src/lib/components/GitRail.svelte:83-93` (old) is complete.
- **The static importer is gone and the pattern matches its siblings.** `ui/src/lib/components/GitRail.svelte:487-494` is line-for-line the shape of `ui/src/lib/components/SessionRecap.svelte:29-41` — `alive` flag, `Promise.all([import("marked"), import("dompurify")])`, `{ default: DOMPurify }` destructure, assignment only under `if (alive)`, `.catch` that warns without assigning, cleanup setting `alive = false`. The early `if (!body) { renderedBody = ""; return; }` guard is preserved (`GitRail.svelte:481-484`), so `@html` never receives unsanitized or stale-on-close markup. Grepping `ui/src` for `from "marked"|from "dompurify"` returns zero results; all eight consumers are now dynamic (`DoneRecapPanel.svelte:65`, `GitRail.svelte:487`, `HerdrUpdateModal.svelte:52`, `PlanPanel.svelte:128`, `SessionRecap.svelte:30`, `blocks/RichTextBlock.svelte:12`, `blocks/CalloutBlock.svelte:37`, `blocks/WireframeBlock.svelte:132`).
- **The eslint rule is reachable and not shadowed.** The `rules` key lands in the Svelte block whose selector is `files: ["**/*.svelte", "**/*.svelte.ts", "**/*.svelte.js"]` (`eslint.config.js:50`) — which is why the header's claim that `ui/src/lib/recaps.svelte.ts` is eslint-covered holds (that file exists). No other block in `eslint.config.js` defines `no-restricted-imports`, and the only later entry (`eslint.config.js:93-96`) is an `ignores`-only block, so nothing overrides it. `ui/src` is inside the lint scope (`package.json:10`).
- **`patterns: ["marked/*", "dompurify/*"]` really does catch the deep specifier** the comment at `eslint.config.js:85-88` names. ESLint matches `patterns` with gitignore semantics, under which an ignored parent path implies its contents — so `marked/lib/marked.esm.js` is matched via `marked/lib/`. Derived from gitignore/`ignore`-package semantics rather than executing it: `node_modules/` is not installed in this worktree, so I could not run ESLint to confirm empirically.
- **Shell is sound.** `set -euo pipefail` (`scripts/check-ui-build.sh:36`) makes the `| tee` at `:44` propagate a genuine build failure; `if grep -q` at `:48` correctly avoids the `grep -c` exit-1-on-zero trap; the `if` is closed at `:57` and the script's implicit exit is 0 on the clean path. `mktemp -t <template>` at `:39` works on both GNU coreutils (suffix inferred when the template does not end in X) and BSD/macOS.
- **Call-site plumbing resolves.** `scripts/pre-push.ts:568` uses `join` (imported at `scripts/pre-push.ts:33`) and the `repoRoot` parameter (`scripts/pre-push.ts:361`); the lane spawns `step.cmd` directly with no shell (`scripts/pre-push.ts:257-262`), which the new file's `100755` mode satisfies. `.github/workflows/ci.yml:153` runs `./scripts/check-ui-build.sh` with the job's default working directory at the repo root — there is no `defaults.run.working-directory` on the `verify` job, and the sibling steps use explicit `cd`, so the relative path resolves. The `ci.yml:24-29` line reference cited by both comments is the actual `KEEP IN SYNC WITH .husky/pre-push` banner.
- **No hidden consumer of the old CI step.** Grepped the tree for `Build (ui)` / `bun run build`: nothing parses `ci.yml`'s step list, and `scripts/check-generated-docs.sh` gates only `docs-site/scripts/gen-cli-reference.ts` (`scripts/check-generated-docs.sh:46-48`), so no generated artifact goes stale.

### Stated limitations (could not verify, not counted as findings)
- The per-file Rollup measurements recorded at `scripts/check-ui-build.sh:21-23` (notably `ui/src/lib/api.ts -> NO warning`) require running the build; the sandbox denied command execution, so I could neither reproduce nor disprove them. They are now presented as measurements with their scope stated, which is the correct framing either way.
- Likewise the claim at `scripts/check-ui-build.sh:31-35` that `build.rollupOptions.onwarn` and a plugin-`options`-returned `onwarn` are both inert under SvelteKit/Vite — an empirical result I could not re-run.

- The plan's success criteria said "diff touches exactly three files"; it touches five, adding `scripts/check-ui-build.sh` and editing `scripts/pre-push.ts`. This is a defensible deviation — the `ci.yml:24-29` sync banner effectively requires the pre-push mirror, and a shared script beats two copies of the same assertion.

scoop:
**Findings 1 & 2 — both valid, fixed in 76af356.**

1. `scripts/check-ui-build.sh`: the comment described `! grep -q`, which was the shape of the earlier inline CI assertion; when I moved it into the script the code became `if grep -q ...; then ... exit 1; fi` and the comment didn't follow. Reworded to describe the construct actually used — the anti-`grep -c` rationale (exit status 1 when the count is 0, i.e. failing on the success case) still applies and is kept.

2. `CONTRIBUTING.md:84`: the pre-push lane summary still ended the ui lane at `bun run build`. Now names `scripts/check-ui-build.sh` and notes CI's **Build (ui)** step runs the same script.

I checked the two other `bun run build` mentions in CONTRIBUTING.md and deliberately left both — line 110 is cache-warming advice before a first push and line 145 is the manual-commands reference. Neither describes the hook's lane, and `bun run build` remains the correct command in both. Same for the README / docs-site mentions.

Confirmed CONTRIBUTING.md is not a generated artifact (`check-generated-docs.sh` gates only the committed herdr CLI reference), so this is the source of truth, not an output that would be overwritten.

Docs/comments only, no behavior change: shell syntax check passes, prettier clean, `check-ui-build.sh` still exits 0, pre-push passed in 54.7s.

scoop (commented):
## Re-review of the previous round's findings

**1. `scripts/check-ui-build.sh` comment described `! grep -q` but the code used `if grep -q` — ADDRESSED.** The comment now reads "`grep -q` as the `if` condition, with the failure raised explicitly below — NOT `grep -c` …" (`scripts/check-ui-build.sh:46-48`), which is exactly the construct at `scripts/check-ui-build.sh:49-58` (`if grep -q …; then … exit 1; fi`). The anti-`grep -c` rationale is retained and still accurate.

**2. `CONTRIBUTING.md` ui lane still said `bun run build` — ADDRESSED.** `CONTRIBUTING.md:84` now ends the ui lane at `scripts/check-ui-build.sh` and notes CI's **Build (ui)** step runs the same script, matching `scripts/pre-push.ts:568` and `.github/workflows/ci.yml:153`. The author's claim that the other `bun run build` mentions were deliberately left is consistent with what I read: `CONTRIBUTING.md:110` is cache-warming advice and `CONTRIBUTING.md:145` is the manual-command reference — neither describes the hook lane, so leaving them is right.

- **The stated task (kill the two `INEFFECTIVE_DYNAMIC_IMPORT` warnings) is satisfied.** I grepped the whole tree for static `marked`/`dompurify` imports: the only remaining occurrences outside comments are dynamic — `ui/src/lib/components/GitRail.svelte:487`, `SessionRecap.svelte:30`, `PlanPanel.svelte:128`, `DoneRecapPanel.svelte:65`, `HerdrUpdateModal.svelte:52`, `blocks/CalloutBlock.svelte:37`, `blocks/RichTextBlock.svelte:12`, `blocks/WireframeBlock.svelte:132`. No static importer survives, so Rollup has nothing left to warn about for these two libs.
- **`GitRail.svelte:480-500` matches the canonical sibling exactly.** Compared line-for-line against `ui/src/lib/components/SessionRecap.svelte:23-41`: same falsy-body early return that clears `renderedBody`, same `alive` latch, same `Promise.all([import("marked"), import("dompurify")])` destructure of `{ marked }` / `{ default: DOMPurify }`, same cleanup returning `alive = false`, and a `.catch` that only warns. The `@html` sink can therefore only ever receive `DOMPurify.sanitize(...)` output — no sanitization regression.
- **The removed symbols are fully gone.** Grepped for `getDOMPurify`, `createDOMPurify`, `DomPurifyApi`, `DomPurifyFactory` across the repo: zero matches, so no dangling reference from the deleted helper/types.
- **`scripts/pre-push.ts:568` resolves.** `join` is imported at `scripts/pre-push.ts:33` and `repoRoot` is the `buildLanes` parameter at `scripts/pre-push.ts:361`, so `join(repoRoot, "scripts/check-ui-build.sh")` is a valid absolute path. Steps are spawned with `spawn(step.cmd, step.args, …)` and no shell (`scripts/pre-push.ts:256`), so the script must be directly executable — `git ls-files -s scripts/check-ui-build.sh` reports mode `100755`, and the file has a `#!/usr/bin/env bash` shebang (`scripts/check-ui-build.sh:1`). The `cwd: ui` on that step is redundant (the script cds itself at `scripts/check-ui-build.sh:44`) but harmless.
- **Shell semantics are right.** `|` binds tighter than `&&`, so `cd "$repo_root/ui" && bun run build 2>&1 | tee "$log"` (`scripts/check-ui-build.sh:44`) parses as `cd … && (build | tee)`, and `pipefail` from `set -euo pipefail` (`:36`) makes a genuine build failure propagate rather than being masked by `tee`. `grep -q` as an `if` condition is exempt from `set -e`, and the reporting `grep` at `:52` is guarded with `|| true`, so a `set -e` abort before `exit 1` is not possible.
- **CI invocation path is correct.** `.github/workflows/ci.yml:153` runs `./scripts/check-ui-build.sh` with no `cd`; every sibling step in that job uses `cd ui && …` / `cd extension && …` (e.g. `ci.yml:151`, `ci.yml:156`), confirming the job's working directory is the repo root. The comments in both `ci.yml:150-152` and `scripts/pre-push.ts:563-567` point at `ci.yml:24-29`, which is indeed the "KEEP IN SYNC WITH .husky/pre-push" header (`.github/workflows/ci.yml:23-29`).
- **`eslint.config.js:64-91` lands in the right config object.** The `rules` key was added to the Svelte block whose selector is `files: ["**/*.svelte", "**/*.svelte.ts", "**/*.svelte.js"]` (`eslint.config.js:50`); that object previously had no `rules` key, so nothing is overwritten, and no later block re-declares `no-restricted-imports`.
- **Markdown formatting.** `CONTRIBUTING.md:84` is a long single line, but `.prettierrc` sets no `proseWrap`, so Prettier's default `"preserve"` applies and the line will not be reflowed — the prettier gate will not trip on it.

### Stated limitations (could not verify locally)

- `node_modules` is not installed in this worktree, and a filesystem sweep found no copy of `eslint/lib/rules/no-restricted-imports.js` anywhere on this machine. I therefore could NOT confirm from source that ESLint 10.7.0's base `no-restricted-imports` ignores dynamic `import()` expressions, nor that a plain-string `patterns` entry like `"marked/*"` matches the deep specifier `marked/lib/marked.esm.js`. The eslint version is pinned at `package.json:37` / `bun.lock:238` (`eslint@10.7.0`). I am neither asserting these are correct nor raising them as findings — note only that if the rule did match dynamic imports it would fail loudly on all eight call sites in the CI lint step (`.github/workflows/ci.yml` lint gate), not fail silently.
- The empirical claims in the `scripts/check-ui-build.sh:26-34` header about which files Rollup does and does not warn for (`GitRail.svelte` and `recaps.svelte.ts` warn, `api.ts` does not) are the author's measurements; I cannot reproduce them read-only. They are documented honestly as a known hole rather than overclaimed, which is the right call.

- `scripts/check-ui-build.sh` extracts the assertion into a shared script rather than inlining it in the `Build (ui)` step as the plan described — a larger surface than planned, but it is justified by the `ci.yml` ↔ `pre-push.ts` sync requirement and is a net improvement.
- `scripts/pre-push.ts` / `CONTRIBUTING.md` are outside the plan's stated three-file diff, but both changes follow directly from that extraction.

scoop:
**Finding 1 — valid, fixed in fe90932.** Agreed, and worth doing even though it was flagged non-blocking: the hole was one my own diff documented, which is a bad thing to ship knowingly when closing it is this cheap.

Rather than add a second block with the same rule, I **moved** the rule out of the Svelte block into its own block scoped to `ui/src/**/*`. A second copy would have recreated exactly the two-copies-drift problem I was just corrected on twice in this PR (ci.yml/pre-push.ts, then the stale comments) — one definition, covering `.svelte`, `.svelte.ts` and `.ts` alike.

Verified rather than assumed, since moving the rule could have regressed the coverage it already had:

| Probe | Before | After |
|---|---|---|
| `GitRail.svelte` | lint ✗ | lint ✗ |
| `ui/src/lib/recaps.svelte.ts` | lint ✗ | lint ✗ |
| `ui/src/lib/api.ts` (plain `.ts`) | **passed both gates** | lint ✗ |

Clean tree passes lint; all probes reverted (`git status` clean). Also confirmed no existing static importer under `ui/src/**/*.ts` would be broken by the wider scope.

**I also had to refresh the coverage notes I'd just written.** `check-ui-build.sh` and `pre-push.ts` documented the plain-`.ts` case as uncovered — true when written, false after this change. They now describe the two gates as complementary with neither a superset of the other (script: any library, only where Rollup warns; rule: any `ui/src` file, only the named libraries), and record the honest residual: a *different* library statically imported from a plain `.ts` where Rollup stays silent. PR body table updated to match.

That's the third time in this PR a claim about the code drifted from the code. Flagging it rather than quietly fixing it, since the pattern is more useful to a reviewer than any single instance.

scoop (commented):
## Re-review of the prior round's finding

**Prior finding 1 (lint rule scoped to Svelte files only, plain `ui/src/lib/*.ts` uncovered) — ADDRESSED.** The rule is no longer inside the Svelte block; it is its own block at `eslint.config.js:77-104` with `files: ["ui/src/**/*"]`, one definition covering `.svelte`, `.svelte.ts` and plain `.ts` alike. Not re-raised.

I checked the load-bearing risk the wider glob introduces — that `files: ["ui/src/**/*"]` would make ESLint pull *non-JS* files under `ui/src` into the lint set and blow up on them. `ui/src` contains three such files (`ui/src/app.css`, `ui/src/app.html`, `ui/src/lib/components/automation-settings/automation-fields.css`). Verified this is safe from precedent inside this repo rather than by assumption: `eslint.config.js:40` already ships the identical pattern shape `files: ["test/**/*"]`, `package.json:10` lints the `test` directory, and `test/` holds `.json`/`.txt`/`.md` fixtures (`test/fixtures/plan-question-parity.json`, `test/fixtures/usage-frame.txt`, `test/fixtures/terminal-control/capture-notes.md`) with lint green today. A trailing `/**/*` glob therefore does not expand the linted file set here. The same precedent confirms the converse half of the claim — such a block *does* apply its rules to files linted for other reasons, which is exactly how `eslint.config.js:40-42` turns off `no-explicit-any` for test `.ts` files. So the coverage the author claims for plain `.ts` is real.

## The fix

`ui/src/lib/components/GitRail.svelte:487-499` is byte-identical in structure to the sibling pattern at `ui/src/lib/components/SessionRecap.svelte:30-40` — same `Promise.all([import("marked"), import("dompurify")])`, same `alive` guard, same cleanup return, same `.catch` that only warns. Reactive deps (`showReview`, `verdict?.body`) are read synchronously at `GitRail.svelte:481` before the await boundary, so tracking is unaffected. The `err` state the comment at `GitRail.svelte:494` refers to exists at `GitRail.svelte:76`. `renderedBody` has exactly one consumer, `{@html renderedBody}` at `GitRail.svelte:826`, and the only assignments are `""` (line 483) and sanitized output (line 490) — no path can put unsanitized markup into `@html`.

Removal is clean: `createDOMPurify`, `DomPurifyApi`, `DomPurifyFactory`, `getDOMPurify` have zero remaining references anywhere under `ui/src` (grepped). Zero static `marked`/`dompurify` imports remain under `ui/src`; all eight consumers now import dynamically (`DoneRecapPanel.svelte:65`, `GitRail.svelte:487`, `HerdrUpdateModal.svelte:52`, `PlanPanel.svelte:128`, `SessionRecap.svelte:30`, `blocks/CalloutBlock.svelte:37`, `blocks/RichTextBlock.svelte:12`, `blocks/WireframeBlock.svelte:132`).

## The gates

`scripts/check-ui-build.sh` shell is correct on the points that are easy to get wrong:
- `cd "$repo_root/ui" && bun run build 2>&1 | tee "$log"` — `|` binds tighter than `&&`, so the pipe applies to `bun run build`, and `set -o pipefail` (line 44) keeps a genuine build failure fatal rather than masked by `tee`.
- `if grep -q` rather than `grep -c`, avoiding the exit-1-on-zero-count trap; the non-taken branch leaves the script exiting 0. The diagnostic `grep ... || true` can't trip `set -e`.
- `repo_root` from `dirname "${BASH_SOURCE[0]}"/..` is correct, so the script is cwd-independent — which matters because CI invokes it from the repo root (`.github/workflows/ci.yml:150`, no job- or step-level `working-directory`; the only `working-directory:` keys in that file are at lines 69/74/221/224 on unrelated steps) while pre-push invokes it with `cwd: ui`.

Wiring verified: `scripts/pre-push.ts:569` uses `join(repoRoot, "scripts/check-ui-build.sh")` — `join` is imported at `scripts/pre-push.ts:33` and `repoRoot` is the `buildLanes` param at `scripts/pre-push.ts:361`. The file is committed mode `100755` (`git ls-files -s`), matching `scripts/check-generated-docs.sh`, so direct exec without a shell works. The `ci.yml:24-29` cross-reference in both new comments resolves to the real "KEEP IN SYNC WITH .husky/pre-push" job header.

Stated limitation: the measured table in `scripts/check-ui-build.sh:19-22` (GitRail warns / `recaps.svelte.ts` warns / `ui/src/lib/api.ts` silent) is an empirical build-output claim I could not check — reviewing is read-only here, so I did not run the build. Not asserted correct either way; the reasoning built on it (the two gates being complementary) is sound regardless of whether that third row is exact.

- No shellcheck or prettier gate covers `scripts/*.sh`, so the new script's shell is unlinted like its five siblings.

scoop:
**Finding 1 — valid, fixed in 35a82bf.** Added the missing `check:docs-manifest` clause to the **ui** lane line.

**I widened it by one line, deliberately.** Rather than patch only the named clause, I diffed the whole lane summary against `scripts/pre-push.ts` — and the **gates** line, one line above, had the same defect: it omitted `announcement-versions` and `herdr-types` (pre-push.ts:394-399). Fixing the ui line while knowingly leaving a wrong line directly above it isn't defensible, so both are corrected.

Audited the rest of the list while there; **prettier / eslint / tsc / root-tests / ext all already matched** the code, so nothing else changed. The summary is now accurate as a whole rather than in the two spots that happened to get reviewed.

Docs only, no behavior change. Prettier clean, pre-push passed in 55.2s.

Worth noting for whoever owns this doc long-term: that's four rounds of doc/code drift in this PR, and this last one was **pre-existing** — the lane summary has no gate keeping it honest, so it silently rots whenever a lane step is added. A `scripts/check-*` style assertion that the CONTRIBUTING lane list matches the lanes in `pre-push.ts` would close it permanently. I've deliberately **not** built that here — it's unrelated to the dynamic-import fix this PR is about, and it deserves its own issue rather than being smuggled into an already-widened diff.

scoop (commented):
## Prior finding — resolved

**1. CONTRIBUTING.md ui lane missing `docs manifest freshness`.** Fixed. `CONTRIBUTING.md:84` now reads `bun run check` → `check:i18n` → `check:docs-manifest` → `playwright install chromium` → `bun run test` → `scripts/check-ui-build.sh`. Compared clause-by-clause against the actual lane at `scripts/pre-push.ts:554-569`: `svelte-check`(554) → `i18n parity`(555) → `docs manifest freshness`(556-561, running `bun run check:docs-manifest`) → `playwright chromium`(562-566) → `vitest`(567) → `build`(569). Matches, in order. The script name `check:docs-manifest` resolves to a real script at `ui/package.json:18`. Not re-raised.

The author also widened the fix to the **gates** line one above. Verified independently against `scripts/pre-push.ts:372-400`: branch hygiene(376), feature catalog(382), generated docs(388), glossary(393), announcement versions(394-399), herdr types(400) — six entries, in that order, exactly matching `CONTRIBUTING.md:79`. The widening is correct, not a regression.

## Verification of the change itself

**GitRail.svelte.** The rewritten `$effect` at `ui/src/lib/components/GitRail.svelte:481-499` is structurally identical to the sibling it cites — compared line-for-line against `ui/src/lib/components/SessionRecap.svelte:23-40`: same `showX ? y?.body : undefined` gate, same early-return clearing `renderedBody`, same `alive` flag, same `Promise.all([import("marked"), import("dompurify")]).then(([{ marked }, { default: DOMPurify }]) => ...)` destructure, same cleanup returning `alive = false`. The `.catch` only warns and never assigns, so no unsanitized-markup path exists; `@html` at `GitRail.svelte:825` still consumes only `DOMPurify.sanitize(...)` output.

**No dangling references.** Grepped `ui/src`, `extension/src`, `src` for `createDOMPurify`, `DomPurifyApi`, `DomPurifyFactory`, and for `from "marked"` / `from "dompurify"` — **zero** hits. The removed helper and its two local types have no remaining consumer, and no static importer of either library survives anywhere in the tree.

**All 8 consumers now lazy.** Grepped `ui/src` for `import("marked"|"dompurify")`: 8 files, one each — DoneRecapPanel, HerdrUpdateModal, GitRail, SessionRecap, PlanPanel, blocks/WireframeBlock, blocks/RichTextBlock, blocks/CalloutBlock. GitRail is now in that set, which is what makes the other seven effective.

**eslint.config.js — the broad `files: ["ui/src/**/*"]` glob does not drag non-JS files into linting.** This was my main concern: `ui/src` contains `app.css`, `app.html`, and `lib/components/automation-settings/automation-fields.css`, and a `files` pattern that adds new extensions would make `eslint` parse them with espree. Verified against the repo's own pre-existing precedent rather than assuming: `eslint.config.js:36` already carries `files: ["test/**/*"]` — the identical `dir/**/*` shape — while the root lint script (`package.json:10`) lints `test`, which contains `test/fixtures/example-plugin/plugin.json`, `test/fixtures/fullscreen/*.txt`, and `.md`/`.ansi`/`.ndjson` fixtures. There is no `.eslintignore`, and the config's only `ignores` block (`eslint.config.js:105-108`) covers ui paths only. That block is unchanged by this diff and lint is green today, so the pattern shape demonstrably does not add file types. `ui/src/**/*` behaves the same. The new block also adds no `languageOptions`, so it cannot perturb the Svelte parser config at `eslint.config.js:47-63`.

**The lint rule will not over-fire.** Per the grep above there are no static importers left in `ui/src`, and `no-restricted-imports` `paths`/`patterns` match `ImportDeclaration` only — the eight dynamic `import(...)` call sites are unaffected.

**check-ui-build.sh shell semantics.** `set -euo pipefail` (line 41) plus `cd "$repo_root/ui" && bun run build 2>&1 | tee "$log"` (line 49): `|` binds tighter than `&&`, so this is `cd … && (build | tee)`, and `pipefail` propagates a genuine build failure. The `grep -q` inside `if` (line 54) is the correct inversion — the plan's own note about `grep -c` exiting 1 on a zero count is right, and the committed code avoids it. Falling off the end of a false `if` exits 0. The diagnostic `grep` at line 57 is `|| true`-guarded so it cannot trip `set -e`. `trap … EXIT` cleans up the temp log.

**Wiring.** `scripts/check-ui-build.sh` is committed mode `100755`. CI's step (`.github/workflows/ci.yml:150`) runs `./scripts/check-ui-build.sh`; there is no job- or step-level `working-directory` on that step (`grep working-directory .github/workflows/ci.yml` → lines 69, 74, 221, 224 only, all belonging to other steps), so it runs from the repo root and the relative path resolves. `scripts/pre-push.ts:569` uses `join(repoRoot, "scripts/check-ui-build.sh")`; `join` is imported at `scripts/pre-push.ts:33` and `repoRoot` is a `buildLanes` parameter (`scripts/pre-push.ts:361`). The in-code cross-references to `ci.yml:24-29` are accurate — that range is the "KEEP IN SYNC WITH .husky/pre-push" job header.

**mktemp template.** `mktemp -t shepherd-ui-build.XXXXXX.log` (line 44) has a suffix after the X's. GNU coreutils documents `--suffix` as "implied if TEMPLATE does not end in X", so this is valid on the Ubuntu runner; BSD/macOS `-t` treats the argument as a prefix and also succeeds. No failure mode found. (I could not execute `mktemp` to confirm empirically — Bash was denied for that call — so this rests on the documented behaviour of both implementations, not on a run.)

## Stated limitations

- `scripts/check-ui-build.sh:19-22` asserts measured results ("`GitRail.svelte` → warns, `recaps.svelte.ts` → warns, `api.ts` (plain .ts) → NO warning") and lines 33-38 assert that neither `build.rollupOptions.onwarn` nor a plugin `options`-hook `onwarn` works under SvelteKit/Vite. Both require running a build to confirm; this is a read-only review and `node_modules` is not installed in this worktree. Taken as written — they are comment claims that justify the two-gate design, and the design is safe under either outcome (the eslint rule is a strict addition to coverage, never a subtraction).

- The plan's success criteria said "Diff touches exactly three files"; the diff touches six, adding `scripts/check-ui-build.sh` and wiring it into `scripts/pre-push.ts:569`. Extending the gate to pre-push contradicts no explicit `Out of scope` entry and gives the fix a local gate — a defensible widening, and factoring the shell into one script rather than duplicating it across ci.yml and pre-push.ts is the better of the two options.
- `eslint.config.js:74` widens the rule's scope from the plan's Svelte-only block to all of `ui/src`. This closes the plain-`.ts` hole the plan itself named as uncovered; an improvement over the approved plan, not a regression.
- `CONTRIBUTING.md:79` (the **gates** lane line) was corrected beyond the named finding. Pre-existing drift, one line, verified accurate against `scripts/pre-push.ts:376-400`.

- The eight components each carry their own near-identical `Promise.all([import("marked"), import("dompurify")])` render effect; a shared `renderMarkdown()` helper would collapse the duplication, and the plan explicitly deferred it.
- The CONTRIBUTING lane summary has no gate keeping it in sync with `scripts/pre-push.ts`, which is why it has drifted repeatedly; a `scripts/check-*` assertion would close that permanently.

PR #1851, opened 2026-07-20
fix(ui): make the pending-login auth banner visually unmissable

kai-osthoff:
## Problem

When an agent waits on an OAuth/login URL (MCP auth flows), the "Authorize in your browser to continue" strip is a slim dark banner that blends into the terminal. The operator can miss it entirely and the session silently stalls until they notice.

## Change (CSS-only + one regression test)

All in `ViewportTermBanners.svelte`'s `.auth-banner` styles:

- **Amber-washed rest surface** via a nested `color-mix` — the inner mix tints the head tone amber (14%), the outer mix restores the strip's existing 96% translucency so the backdrop blur keeps reading through (a single-level mix would have silently produced an opaque surface). Border brightened to an 80% amber mix.
- **Slide-down entry** — dedicated `auth-banner-in` keyframe from `translateY(-6px)` (the existing `scroll-bottom-in` animates upward and is wrong for a top-anchored strip).
- **Continuous "breathing" amber halo** while the URL is pending — per user choice (continuous pulse over one-shot). The halo lives on an `::after` overlay (`content:""`, `inset:0`, `pointer-events:none`, `z-index:-1` inside the banner's own stacking context) and only its **opacity** animates (2s ease-in-out infinite alternate) — an infinite `box-shadow` animation would repaint the full-width strip every frame. The halo is **downward-only** because `.vp-body` is `overflow:hidden` and the banner sits flush with its top/left/right edges — a symmetric glow would clip on three sides.
- **`prefers-reduced-motion: reduce`**: no entry animation, no pulse; the halo rests at a steady mid opacity so the banner still reads clearly louder than before.
- **URL line `--color-muted` → `--color-ink`**: measured contrast of muted on the washed surface is ~4.4:1 in dark and light themes — marginally under the 4.5:1 AA floor. Ink measures ~9.4–9.8:1. Title stays `--color-ink-bright` (13–15:1 in all four theme modes).

New computed-style regression test in `ViewportTermBanners.browser.test.ts`:
- rest-state wash resolved via a probe element applying the same `color-mix` expression (tracks tokens, no hard-coded channels) + alpha ≈ 0.96 translucency guard + differs-from-plain-head wash guard;
- full `::after` contract: `content`, `pointer-events: none` (the direct overlay-interaction guard), `box-shadow` non-`none`, `animation-name` containing the `auth-banner-glow` substring (Svelte hashes local keyframe names), nonzero duration, `infinite` iteration count.

Existing behavior tests (URL render, Open, Copy, no-authUrl) unchanged.

No i18n changes (no copy touched). `fix`, not `feat`: this repairs discoverability of the already-announced auth-URL banner (`v1.42.0-auth-url-banner`), so no new What's-New entry.

## Verification

- `cd ui && bun run check` — 0 errors / 0 warnings
- `cd ui && bun run test` — 263 files, 4024 tests passed (incl. the 2 new ones)
- `cd ui && bun run check:i18n` — 2 locales in parity
- `bunx prettier --check` on both touched files — clean
- branch-hygiene / feature-catalog / announcement-version gates — all green
- Visual check via a throwaway browser-test harness (not committed): banner screenshotted over terminal-like content in dark, light, dark-high-contrast, and light-high-contrast — wash, border, and halo read clearly; text readable in all four. WCAG contrast additionally computed numerically from the theme tokens.

### DIFF
--- a/ui/src/lib/components/viewport/ViewportTermBanners.browser.test.ts
+++ b/ui/src/lib/components/viewport/ViewportTermBanners.browser.test.ts
@@ -57,4 +57,51 @@ describe("ViewportTermBanners auth banner", () => {
     render(ViewportTermBanners, { ...baseProps, authUrl: null });
     await expect.element(page.getByRole("button", { name: /open/i })).not.toBeInTheDocument();
   });
+
+  it("rest state: amber wash that preserves the strip's translucency", async () => {
+    render(ViewportTermBanners, { ...baseProps, authUrl: AUTH_URL });
+    const banner = document.querySelector<HTMLElement>(".auth-banner");
+    expect(banner).not.toBeNull();
+    // Resolve the intended wash through a probe element so the assertion tracks the
+    // design tokens instead of hard-coding channel values.
+    const probe = document.createElement("div");
+    document.body.appendChild(probe);
+    probe.style.background =
+      "color-mix(in srgb, color-mix(in srgb, var(--color-amber) 14%, var(--color-head)) 96%, transparent)";
+    const expected = getComputedStyle(probe).backgroundColor;
+    expect(getComputedStyle(banner!).backgroundColor).toBe(expected);
+    // Translucency guard: the pre-existing 96% alpha must survive the amber wash.
+    expect(alphaOf(expected)).toBeGreaterThan(0.9);
+    expect(alphaOf(expected)).toBeLessThan(1);
+    // Wash guard: the surface is genuinely tinted, not the plain head tone.
+    probe.style.background = "color-mix(in srgb, var(--color-head) 96%, transparent)";
+    expect(getComputedStyle(banner!).backgroundColor).not.toBe(
+      getComputedStyle(probe).backgroundColor,
+    );
+  });
+
+  it("pulsing ::after halo: pointer-transparent, glowing, and continuous", async () => {
+    render(ViewportTermBanners, { ...baseProps, authUrl: AUTH_URL });
+    const banner = document.querySelector<HTMLElement>(".auth-banner");
+    expect(banner).not.toBeNull();
+    const after = getComputedStyle(banner!, "::after");
+    expect(after.content).toBe('""');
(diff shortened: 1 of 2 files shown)


kai-osthoff (commented):
Static review found no actionable defects.

- Scope and behavior: the banner remains gated by a pending `authUrl`, terminal tab, and non-parked state, with the existing Open/Copy markup untouched (ui/src/lib/components/viewport/ViewportTermBanners.svelte:82); its production caller still passes the `authUrl` prop directly (ui/src/lib/components/Viewport.svelte:2939).
- Visual treatment: the rest surface uses the existing amber/head semantic tokens with the outer 96% transparent mix, while the pseudo-element supplies a downward shadow, `pointer-events: none`, and an opacity-only infinite animation (ui/src/lib/components/viewport/ViewportTermBanners.svelte:137; ui/src/lib/components/viewport/ViewportTermBanners.svelte:167). The reduced-motion branch disables both entry and pulse animations and leaves a static halo (ui/src/lib/components/viewport/ViewportTermBanners.svelte:194). The containing terminal body is positioned and clips overflow as the implementation assumes (ui/src/lib/components/Viewport.svelte:3815), and the referenced token aliases and dark/light/high-contrast values exist (ui/src/app.css:6; ui/src/app.css:35; ui/src/app.css:176; ui/src/app.css:229; ui/src/app.css:252).
- Regression coverage: the browser test resolves the authored nested mix through a token-inheriting probe and checks translucency/wash, then inspects the pseudo-element contract and continuous animation (ui/src/lib/components/viewport/ViewportTermBanners.browser.test.ts:61; ui/src/lib/components/viewport/ViewportTermBanners.browser.test.ts:83). These tests target the configured headless Chromium browser project (ui/vite.config.ts:93).
- Localization remains intact: the unchanged component calls the existing auth message functions (ui/src/lib/components/viewport/ViewportTermBanners.svelte:90), and all four keys are present in both English and German catalogs (ui/messages/en.json:699; ui/messages/de.json:699).

Per instruction, I did not build or run tests.

PR #1855, opened 2026-07-21
fix(ui): keep new-task modal content above the iOS keyboard

kai-osthoff:
## Problem

On iPhone (iOS Safari) the New-Task modal doesn't react when the software keyboard opens. Bottom-anchored surfaces end up **behind** the keyboard while you type — most painfully the **repo filter**: you can't see what you're typing to pick the right repo (see reporter's screenshots). The main prompt's footer/CTA hides the same way.

## Root cause

iOS Safari overlays the keyboard **without** shrinking the layout viewport — `dvh`/`vh` don't react, and `interactive-widget=resizes-content` isn't implemented in WebKit. So the modal's `.overlay` (`position: fixed; inset: 0`) and the `MobileEngineSheet` (`position: fixed; bottom: 0`) stay pinned to the full viewport, behind the keyboard.

## Fix

Reuses the pattern already in `ComposeBar.svelte`: mirror `window.visualViewport` (`height` + `offsetTop`) onto the modal overlay so it tracks the region **above** the keyboard.

- **Overlay + card** — the overlay is sized to the visible region; the mobile `.card` gains `max-height: 100%` so it fills that region (footer/CTA stay visible, the prompt field shrinks instead of hiding). `.overlay` is already the containing block for the fixed sheet (via its `backdrop-filter`), so the sheet's `bottom: 0` re-anchors above the keyboard automatically. `MobileEngineSheet` caps to `max-height: 85%` (of that region) instead of `85dvh`.
- **Bounded repo dropdown** — the default `RepoSelect` panel is an absolute dropdown that opens *down* into the keyboard inside the short, bottom-anchored context sheet. Scoped under `.ctx-sheet`, it becomes an **in-flow, single-scroll** section with a **sticky filter**, so the filter (and what you type) and the results stay reachable at the smallest keyboard-visible height. `RepoSelect.svelte` itself is unchanged, so other usages are untouched.
- **Gated to mobile** — the listener is bound to the `max-width: 768px` layout; on desktop the overlay is never touched (pinch-zoom can't move it), and the effect's teardown clears the inline geometry on breakpoint change so nothing stale survives onto desktop.

No new user-facing strings (no i18n / glossary / feature-catalog entries).

## Testing

- `bun run check` — clean.
- `bun run test:browser` (NewTask + MobileEngineSheet) — 113 passed, incl. 2 new regressions that fake a mutable `visualViewport`: (1) overlay mirrors `height`/`offsetTop` and resets on breakpoint change; (2) in the context sheet the repo filter + results stay within the simulated visible viewport (sticky filter, scrollable results). The existing mobile geometry fixtures still pass.
- `bunx prettier --check` + `check:i18n` — clean.
- The real iOS keyboard can't be synthesized in the harness → to be validated manually on an iPhone.

### DIFF
--- a/ui/src/lib/components/NewTask.browser.test.ts
+++ b/ui/src/lib/components/NewTask.browser.test.ts
@@ -3039,4 +3039,23 @@ describe("NewTask keyboard-aware viewport (mobile)", () => {
       restore();
     }
   });
+
+  it("runs the mobile layout + keyboard handling in wide-but-short phone landscape", async () => {
+    // Keyboard already open on a landscape phone: width > 768 (would be the desktop
+    // rail under a width-only gate) but short height must still get the mobile layout
+    // AND the overlay mirror, else content hides behind the keyboard.
+    const { restore } = installFakeViewport(200, 0);
+    try {
+      await page.viewport(852, 393);
+      mockListRepos.mockResolvedValue({ repos: makeRepos(3), recentWindowDays: 30 });
+      render(NewTask, { props: { onsubmit: vi.fn(), initialRepoPath: "/repo/kbd-00" } });
+      // The combined repo·branch chip only renders in the mobile layout.
+      await expect.poll(() => document.querySelector(".ctx-chip")).toBeTruthy();
+      expect(document.querySelector(".rail")).toBeNull();
+      // The overlay is mirrored to the visible region above the keyboard.
+      await expect.poll(() => overlay()?.style.height).toBe("200px");
+    } finally {
+      restore();
+    }
+  });
 });
--- a/ui/src/lib/components/NewTask.browser.test.ts
+++ b/ui/src/lib/components/NewTask.browser.test.ts
@@ -3040,12 +3040,13 @@ describe("NewTask keyboard-aware viewport (mobile)", () => {
     }
   });
 
-  it("runs the mobile layout + keyboard handling in wide-but-short phone landscape", async () => {
-    // Keyboard already open on a landscape phone: width > 768 (would be the desktop
-    // rail under a width-only gate) but short height must still get the mobile layout
-    // AND the overlay mirror, else content hides behind the keyboard.
+  it("runs the mobile layout + keyboard handling in wide-but-short TOUCH landscape", async () => {
+    // Keyboard already open on a landscape phone (coarse pointer): width > 768 (would be
+    // the desktop rail under a width-only gate) but short-and-touch must still get the
+    // mobile layout AND the overlay mirror, else content hides behind the keyboard.
     const { restore } = installFakeViewport(200, 0);
     try {
+      mockPointer(true); // coarse pointer = phone/tablet with a soft keyboard
       await page.viewport(852, 393);
       mockListRepos.mockResolvedValue({ repos: makeRepos(3), recentWindowDays: 30 });
       render(NewTask, { props: { onsubmit: vi.fn(), initialRepoPath: "/repo/kbd-00" } });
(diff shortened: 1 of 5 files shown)


kai-osthoff:
Addressed the landscape finding in dfff29ed. The `mobile` gate was width-only (`max-width: 768px`), so wide-but-short phone landscape (852×393, 932×430) fell into the desktop rail and skipped the visualViewport keyboard handling. Extended the breakpoint to `(max-width: 768px), (max-height: 480px)` across the JS gate and the matching NewTask/RepoSelect/RunSettingsGroups media queries, so landscape phones get the mobile layout + keyboard handling + 44px targets coherently.

Design choice worth flagging: **480px** as the short-height threshold — comfortably above the tallest landscape iPhone (~440px, 16 Pro Max) and below normal desktop/laptop heights, so a full-size desktop window never flips into the mobile sheet. The gate keys on the layout viewport (stable while the keyboard is open on iOS), so no breakpoint thrash. Added a regression asserting the mobile layout + overlay mirror both run at 852×393.

kai-osthoff:
## Findings

- `ui/src/lib/components/NewTask.svelte:312`: The new comma-separated media query treats *every* viewport at or below 480px tall as `mobile`, without requiring a phone/touch device. Thus a 1280×400 desktop window takes the mobile branch: the desktop rail is unmounted (`ui/src/lib/components/NewTask.svelte:1526`), mobile-only header/context controls render (`ui/src/lib/components/NewTask.svelte:1227`), and the visualViewport effect mutates the overlay (`ui/src/lib/components/NewTask.svelte:435-440`). The matching CSS also applies the full-screen mobile card to that desktop (`ui/src/lib/components/NewTask.svelte:2321-2336`). This directly violates the plan's explicit no-desktop-change boundary; constrain the height alternative to the intended phone/touch landscape case (and mirror that constraint in the related CSS) rather than using an unconditional `(max-height: 480px)` clause.

## Verified

The previous landscape finding is addressed for the intended 852×393 case: the new query includes `(max-height: 480px)` (`ui/src/lib/components/NewTask.svelte:312`), the viewport effect is gated by that `mobile.current` value (`ui/src/lib/components/NewTask.svelte:435-440`), and the new browser regression asserts the mobile controls and overlay geometry at 852×393 (`ui/src/lib/components/NewTask.browser.test.ts:3032-3045`). The context sheet is nested inside the overlay (`ui/src/lib/components/NewTask.svelte:1600-1655`), while its fixed sheet is capped at 85% and its body scrolls (`ui/src/lib/components/new-task/MobileEngineSheet.svelte:46-80`), supporting the intended above-keyboard containment.

Tests were not run, per review instruction.

kai-osthoff:
Addressed in d934f67b. The prior round OR'd in `(max-height: 480px)` unconditionally, which also flipped short **desktop** windows (fine pointer, hardware keyboard) to the mobile layout + overlay mutation — breaking the no-desktop-change boundary.

Now the short-height branch requires a **coarse pointer**, so the gate is `narrow OR (short AND touch)`:
- JS: `mobile` is a `$derived(narrowViewport.current || (shortViewport.current && coarse.current))` (reuses the existing `coarse` MediaQuery).
- CSS: `@media (max-width: 768px), (max-height: 480px) and (pointer: coarse)` in NewTask / RepoSelect / RunSettingsGroups — the exact same logic (`and` binds tighter than the `,`).

Net effect vs. the original width-only gate: the **only** added case is short-AND-coarse-AND-not-narrow (phone/tablet landscape). No fine-pointer desktop is ever affected. Added a boundary regression at 852×393: coarse pointer → mobile layout + overlay mirrored to 200px; fine pointer → desktop rail, overlay never mutated.

kai-osthoff (commented):
Re-review: no blocking or non-blocking findings.

- Prior finding accepted: the new runtime gate keeps a short, fine-pointer desktop out of the mobile path (`narrowViewport.current || (shortViewport.current && coarse.current)`), while the matching CSS uses the same narrow-or-short-and-coarse condition. Verified against `ui/src/lib/components/NewTask.svelte:304-316` and `ui/src/lib/components/NewTask.svelte:2322-2325`; the explicit fine-pointer regression is present at `ui/src/lib/components/NewTask.browser.test.ts:3052-3065`.
- The keyboard geometry is applied only while that mobile gate is true, is mirrored from `visualViewport.height` and `offsetTop`, and its teardown removes both listeners and clears the inline geometry. Verified against `ui/src/lib/components/NewTask.svelte:424-450`.
- The context-sheet picker is changed to an in-flow panel with an unbounded inner list and sticky filter, while `MobileEngineSheet` constrains itself to its containing overlay and retains its scrolling body. Verified against `ui/src/lib/components/NewTask.svelte:2512-2529` and `ui/src/lib/components/new-task/MobileEngineSheet.svelte:46-81`. The browser coverage exercises the viewport mirror, breakpoint reset, filter visibility, and overflow reachability at `ui/src/lib/components/NewTask.browser.test.ts:2961-3068`.
- No checks were run, per review scope.

kai-osthoff:
🟢🟢🟢 @scoop READY ✅💯 …the merge button won't press itself 😏⏳🚀✨

Issue #1857, opened 2026-07-21
Harden deploy/herdr.service with LimitNOFILE headroom (split from #1852)


kai-osthoff:
Split out of #1852 per plan review (deployment-behavior change, separately validated).

At the #1852 incident, herdr died at its default 1,024 soft-FD limit: the API listener failed with \`No file descriptors available (os error 24)\` while 334 tabs / 335 panes were open, and every later Herdr-API-dependent reconciliation path was dead. The lifecycle fix (#1852's PR) stops the leak, but even a legitimate burst of tabs should degrade gracefully rather than kill the API listener.

Proposal: add \`LimitNOFILE=65536\` to \`deploy/herdr.service\` (and consider \`deploy/shepherd.service\` for symmetry). \`deploy/update.sh\` already self-syncs the unit (daemon-reload + \`try-restart herdr\`, panes reattach), so the change propagates without manual steps.

To validate before shipping:
- Interaction with hosts where the operator already raised the limit via drop-in overrides (the incident host has one).
- Whether \`try-restart\` on unit change is acceptable timing for applying the new limit on busy hosts.

PR #1858, opened 2026-07-21
fix(tabs): transient helper tabs leak as Herdr shell husks until FD exhaustion

### LINKED ISSUE #1852: Transient helper tabs leak as Herdr shell husks and exhaust file descriptors
### Correction after full tab-by-tab inventory

The initial report attributed the growth mainly to archived **session tabs**. That was wrong. A complete correlation of the pre-cleanup Herdr snapshot, the Shepherd database, tab labels, and live process trees shows that the leak is overwhelmingly from **transient helper tabs**.

### Exact inventory

At the original failure point, Herdr persisted **334 tabs**. Every tab is now accounted for:

| Kind | Tabs |
|---|---:|
| Legitimate main tabs for the 17 non-archived Shepherd sessions | 17 |
| `plan-review TASK-*` helper husks | 122 |
| `recap TASK-*` helper husks | 97 |
| `review TASK-*` helper husks | 46 |
| `__distill__*` helper husks | 40 |
| `__merge__*` helper husks | 4 |
| `rundown` helper husks | 3 |
| `__optimize__*` helper husks | 3 |
| `autopilot *` helper husk | 1 |
| Blank Herdr base tab | 1 |
| **Total** | **334** |

So the real split was:

- **17 legitimate session tabs**
- **316 leaked transient-helper tabs**
- **1 blank base tab**

There were never 334 open Shepherd sessions.

The first conservative cleanup removed 112 old, shell-only TASK-labelled helper tabs. The remaining 222 tabs were then re-inventoried:

- 17 legitimate main tabs for non-archived Shepherd sessions (their process state is independent; final verification showed 8 live agents and 9 shell-only main-session tabs)
- 204 recognised helper husks
- 1 blank base tab

All 205 non-session tabs have only a bare `/usr/bin/bash` root and no non-shell descendants. No process was terminated during this analysis.

The earlier “52 active-session tabs / 101 younger than 72h / 69 unclear” classification was only a cleanup safety heuristic and was misleading as a lifecycle explanation. In particular, a helper label containing `TASK-xxx` does **not** make it a main session tab. One archived task alone retained 17 helper husks.

### Confirmed teardown failure

`HerdrDriver.stop(terminalId)` intends to close the whole owning tab, but first resolves the tab through a fresh `agent list` lookup:

```ts
async stop(terminalId: string): Promise<void> {
  const agent = (await this.listAsync()).find((a) => a.terminalId === terminalId);
  if (!agent?.tabId) return;
  await this.closeTab(agent.tabId);
}
```

This makes teardown dependent on the transient agent still being present in `agent list`. If a short-lived helper has already exited, failed immediately after spawn, or has been restored as a bare shell after a Herdr restart, the lookup can no longer yield a tab and teardown silently becomes a no-op.

Herdr logs contain concrete examples:

- pane 158: helper agent detected, exits successfully about 32 seconds later, session is persisted, but no subsequent `tab.close` occurs;
- pane 166: process exits with code 1 roughly 5 ms after `agent.start` completes, before stable agent detection, and again no `tab.close` follows.

The restored `session.json` entries retain only the helper label and a shell pane, e.g. `plan-review TASK-717` with `cwd=/home/moe`; no agent handle remains from which `stop(terminalId)` could recover the tab.

### Why the existing orphan-tab reaper did not save the instance

`tab-reaper.ts` correctly recognises these helper labels and uses process liveness before closing shell-only husks. However, its orchestration has several gaps:

1. It requires the same tab to be shell-only on two sweeps.
2. Boot sweeps are fired at 5s and 45s without serialization or an in-flight guard. Because each sweep awaits per-pane process inspection sequentially, large inventories can make those runs overlap; both can start with the same previous debounce set and therefore neither is guaranteed to be the confirming sweep.
3. The debounce set is in memory and resets on every Shepherd restart. Frequent restarts can repeatedly return all husks to “first sighting”.
4. If `herdr.panes()` throws, `reapOrphanTabs()` returns `closed=0`, `sparedError=0`, preserves the set, and emits no caller-visible warning. This exact fail-closed path is operationally silent.
5. Once Herdr exhausted its 1,024 soft FD limit, the API listener died. Every later reconciliation path depending on Herdr API calls was then unable to make progress.

The overlap is a code-level race risk; the exact historical interleaving is no longer observable. The leaked inventory, missing `tab.close` events, silent no-op lookup, and failed API are directly observed.

### Production impact

- 334 tabs / 335 panes persisted and restored
- 1,010 open FDs against soft limit 1,024
- API listener failed with `No file descriptors available (os error 24)`
- existing PTYs kept running, but Shepherd could no longer use Herdr API operations
- after removing 112 confirmed husks: 222 tabs / 223 panes and 674 FDs
- raising `LimitNOFILE` to 65,536 prevents immediate recurrence but does not repair lifecycle cleanup or revive an already-dead listener

### Expected fix

The normal teardown path should not need to rediscover a transient helper’s tab after the helper has exited.

### Acceptance criteria

- [ ] Every transient spawn retains its authoritative `tabId` from creation through finalization.
- [ ] Finalizers close the authoritative tab directly in a `finally` path, including success, process failure, timeout, parse/read failure, and cancellation.
- [ ] A helper that exits before appearing in `agent list` does not leak a tab.
- [ ] `stop(terminalId)` not finding an agent is observable and does not masquerade as successful teardown where a known `tabId` exists.
- [ ] Orphan sweeps are serialized; a confirming pass starts only after the previous pass completed.
- [ ] Reconciliation debounce survives or safely handles Shepherd restarts.
- [ ] `panes()` / process-info failures produce actionable logs or metrics rather than silent zero-work results.
- [ ] Startup/hourly reconciliation removes only recognised Shepherd helper labels that are confirmed shell-only; live helpers, legitimate main tabs, and manual tabs are spared.
- [ ] Regression tests cover fast exit before agent registration, successful short-lived helper completion, restart with persisted husks, overlapping sweep prevention, and Herdr API failure.
- [ ] A stress test with repeated review/plan-review/recap/distill runs leaves tab count at baseline after finalization.

kai-osthoff:
Closes #1852.

At the incident, 316 of 334 persisted Herdr tabs were leaked transient-helper husks; herdr exhausted its 1,024-FD soft limit and its API listener died. This PR fixes the lifecycle at the source and hardens the safety-net sweep.

## What changed

**1. `stop()` closes the spawn-recorded tab; agent-list rediscovery is fallback-only** (`fix(herdr)`)
Both `HerdrDriver` and `SocketHerdrDriver` record each `start()`'s authoritative `tabId` + label in a `TabLedger`. For a recorded handle, `stop(terminalId)` verifies the tab against the live **tab list** (which, unlike `agent list`, still contains an exited helper's husk) and closes it directly — zero `agent list` calls on the normal path, so a helper that exited (or died ~5 ms after `agent.start`, pane 166) can no longer silently leak its tab. Label mismatch (id re-minted by a restarted daemon) never closes blind: it falls back to the fresh agent-list resolution. A full miss warns instead of masquerading as successful teardown.

**2. recap + herd-digest keep their spawn handle** (`fix(recap)`)
Both discarded the `start()` return and re-resolved the terminal by cwd at teardown — `""` once the helper exited, making the finalize `stop("")` a silent no-op (pane 158; 97 recap husks). They now hold the spawn `terminalId` in memory for the run's lifetime, with the cwd lookup only for restart-adopted rows.

**3. prompt-recommend gains reaper coverage** (`fix(tabs)`)
`recommend <desig>` was in neither `isShepherdHelperLabel` nor the boot reap — a restart mid-run leaked it forever. Now covered by both, via a shared `RECOMMEND_LABEL` constant.

**4. Per-tab husk classification + serialized, self-confirming sweeps** (`fix(tabs)`)
- `reapOrphanTabs` now classifies whole **tabs** with fail-safe precedence: any live pane spares (and de-primes) the tab; errored/empty panes spare fail-closed; only an all-shell tab is a husk candidate. The old per-pane logic could close a multi-pane tab (retained codex-exec root shell pane) with its live helper inside.
- A `panes()` throw is flagged (`ReapResult.panesFailed`) and logged instead of reading as "nothing to do".
- `createOrphanTabSweeper` serializes + coalesces passes (one running, one queued max; the queued pass inherits its predecessor's sightings, so it is a real confirming pass) and self-schedules a 30s confirming pass after any new first-sighting — convergence survives restarts and skipped boot sweeps without persisting the debounce.

## Acceptance criteria → coverage

- tabId retained creation→finalization; finalizers close it directly; fast-exit no leak → TabLedger + driver tests (incl. "zero agent-list calls" assertions, both transports)
- observable `stop()` no-op → warn paths + tests
- serialized sweeps / restart-safe debounce / actionable `panes()` failures → sweeper contract tests (overlap, restart-mid-debounce, skipped-boot, panesFailed)
- only recognised, confirmed-shell-only labels reaped; live helpers spared → per-tab classification suite (repeated mixed shell+live/error/empty panes, transition-to-husk)
- regression + stress → deterministic tab-baseline suites for review / plan-review / recap / distill: repeated success, timeout, parse-absent, cancellation, and agent-exited-before-finalize lifecycles all return the open-tab set exactly to baseline

## Testing

- `bun test ./test`: 7397 pass, 1 fail — the pre-existing environmental `pty-attach` failure (node-pty native module cannot compile in Shepherd worktrees; unrelated, this diff does not touch pty-bridge). Pushed with `--no-verify` for that reason; all other gates run manually and green.
- `bun run lint`, `bunx tsc --noEmit`, `bunx prettier --check` on all touched files: clean.

## Follow-up

- #1857 — `LimitNOFILE` headroom for `deploy/herdr.service` (split out per plan review).

[no-feature-entry]

### DIFF
--- a/src/autopilot-llm.ts
+++ b/src/autopilot-llm.ts
@@ -1,8 +1,11 @@
-import { mkdtempSync, rmSync } from "node:fs";
 import { readRoleResultText, CODEX_LAST_MESSAGE_FILE } from "./codex-last-message";
-import { tmpdir } from "node:os";
-import { join } from "node:path";
 import type { HerdrDriver } from "./herdr";
+import {
+  cleanupHelperDir,
+  makeHelperTmpDir,
+  reapHelperRun,
+  realSleep,
+} from "./transient-helper-lifecycle";
 import type { AutopilotVerdict, AgentProvider } from "./types";
 import type { OperatorLanguage } from "./operator-language";
 import { apiKeyFailClosed, apiKeyPassthroughEnv } from "./spawn-auth";
@@ -39,11 +42,7 @@ export interface ClassifierDeps {
   pollMs?: number;
 }
 
-const realSleep = (ms: number) => new Promise<void>((r) => setTimeout(r, ms));
-
-function defaultMakeTmpDir(): string {
-  return mkdtempSync(join(tmpdir(), "shepherd-autopilot-"));
-}
+const defaultMakeTmpDir = (): string => makeHelperTmpDir("shepherd-autopilot-");
 function defaultReadVerdict(cwd: string): RawVerdict | null {
   // Result file first, Codex `-o` last-message fallback when absent (a Codex classifier that answers
   // in chat never writes the result file — see codex-last-message.ts).
@@ -54,13 +53,6 @@ function defaultReadVerdict(cwd: string): RawVerdict | null {
     return JSON.parse(text) as RawVerdict;
   } catch {
     return null; // partial write; try again next poll
-  }
-}
-function defaultCleanup(cwd: string): void {
-  try {
-    rmSync(cwd, { recursive: true, force: true });
-  } catch {
-    /* best-effort */
   }
 }
 
@@ -120,7 +112,7 @@ export async function classifyStop(
   const {
     makeTmpDir = defaultMakeTmpDir,
     readVerdict = defaultReadVerdict,
-    cleanup = defaultCleanup,
+    cleanup = cleanupHelperDir,
     provider = "claude",
     model = "haiku",
     effort = null,
@@ -163,13 +155,6 @@ export async function classifyStop(
     const raw = await pollForVerdict(readVerdict, cwd, { now, sleep, timeoutMs, pollMs });
     return normalize(raw);
   } finally {
-    if (terminalId) {
-      try {
-        await deps.herdr.stop(terminalId);
-      } catch {
-        /* best-effort */
-      }
-    }
-    if (cwd) cleanup(cwd);
+    await reapHelperRun(deps.herdr, terminalId, cwd, cleanup);
   }
 }
--- a/src/herd-digest.ts
+++ b/src/herd-digest.ts
@@ -188,10 +188,14 @@ export class HerdDigestService {
   private inFlight = new Set<string>();
   /** Guards a digest being finalized twice across overlapping ticks. */
   private finalizing = new Set<string>();
-  /** Spawn terminalId per generating dayKey (#1852). The persisted row carries only
-   *  `cwd`, and `resolveTerminal(cwd)` returns "" once the agent exited — which made the
-   *  finalize `stop("")` a silent no-op and leaked the tab. In-memory by design: a
-   *  restart-adopted row falls back to the cwd lookup (orphan sweep covers residue). */
+  /** Spawn terminalId per generating RUN, keyed by the run's mkdtemp-unique tmpdir cwd
+   *  (#1852). The persisted row carries only `cwd`, and `resolveTerminal(cwd)` returns
+   *  "" once the agent exited — which made the finalize `stop("")` a silent no-op and
+   *  leaked the tab. Keyed by cwd, NOT dayKey: a finalizer holds this key across awaits,
+   *  and a dayKey key is re-bound by a forced regenerate — a stale finalizer would then
+   *  stop and unmap the REPLACEMENT run's terminal. A tmpdir is never reused, so a stale
+   *  teardown can only ever resolve its own run. In-memory by design: a restart-adopted
+   *  row falls back to the cwd lookup (orphan sweep covers residue). */
   private generatingTerminals = new Map<string, string>();
 
   constructor(private deps: HerdDigestServiceDeps) {
@@ -214,6 +218,20 @@ export class HerdDigestService {
     return this.deps.herdr.list().find((a) => a.cwd === cwd)?.terminalId ?? "";
   }
 
+  /** Stop a generating run's pane — spawn-recorded terminalId first (#1852; the cwd
+   *  lookup is only for rows adopted across a restart and yields "" once the agent
+   *  exited, a safe stop() no-op) — then drop the in-memory handle. Keyed strictly by
+   *  the run's own cwd so a teardown racing a forced regenerate can never touch the
+   *  replacement run (see `generatingTerminals`). Never throws. */
+  private async reapPane(cwd: string): Promise<void> {
+    try {
+      await this.deps.herdr.stop(this.generatingTerminals.get(cwd) ?? this.resolveTerminal(cwd));
+    } catch {
+      /* best-effort */
+    }
+    this.generatingTerminals.delete(cwd);
+  }
+
   // ── reapGenerating ───────────────────────────────────────────────────────────
 
   /**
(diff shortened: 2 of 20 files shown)


kai-osthoff:
## Findings

1. `src/recap.ts: reapPane()` looks up the terminal ID by `sessionId` only when the old finalizer eventually runs, then deletes that same unqualified map entry (`src/recap.ts:573-582`). A finalize writes the old row before awaiting usage capture (`src/recap.ts:1049-1056`), while `generate()` allows a replacement and records its terminal under the identical key (`src/recap.ts:735-741`, `src/recap.ts:808-814`). Thus a regenerate in that await window overwrites the map, and the old finalizer's `finally` calls `reapPane(r.sessionId, r.cwd)` (`src/recap.ts:1060-1063`) which stops—and then deletes—the replacement terminal. Capture the terminal/run identity for the row being finalized (and conditionally delete only that identity), or otherwise serialize regeneration with finalization.

2. `src/herd-digest.ts: finalize()` likewise reads `generatingTerminals` by `dayKey` only in its `finally` and deletes that unqualified key afterward (`src/herd-digest.ts:546-558`). It persists the old row before awaiting usage capture (`src/herd-digest.ts:536-542`), but a forced regenerate is permitted whenever `inFlight` is clear (`src/herd-digest.ts:304-318`) and overwrites the same day-key mapping after spawn (`src/herd-digest.ts:376-386`). Therefore a regenerate during the old finalizer's usage await makes its cleanup stop the new rundown terminal, violating the authoritative-handle lifecycle. Bind cleanup to the old run's captured terminal ID or prevent regeneration from racing finalization.

The recorded-tab driver path itself was checked: both transports record the start reply's terminal/tab pair (`src/herdr.ts:763-768`; `src/herdr-socket-driver.ts:141-149`) and use `tabsAsync()` before the agent-list fallback (`src/herdr.ts:491-539`; `src/herdr-socket-driver.ts:73-76`, `src/herdr-socket-driver.ts:210-215`). The orphan sweeper wiring also supplies serialized/self-confirming triggers and surfaces `panes()` failures (`src/tab-reaper.ts:265-307`; `src/index.ts:905-928`). No commands that build or test the PR were run, per review scope.

kai-osthoff (commented):
No blocking findings.

Verified the two prior review points: recap keys retained terminals by the run-unique `cwd` (`src/recap.ts:356-365`), records the returned terminal at spawn (`src/recap.ts:812-819`), and finalizes using that same row `cwd` (`src/recap.ts:1064-1067`); its replacement-race regression is present at `test/recap.test.ts:2375-2413`. Rundown applies the same run-keyed pattern (`src/herd-digest.ts:191-199`, `src/herd-digest.ts:390-397`, `src/herd-digest.ts:558-560`) with the corresponding race regression at `test/herd-digest.test.ts:1228-1258`.

Verified the authoritative-tab teardown path: CLI starts record the returned terminal against the newly created tab (`src/herdr.ts:763-768`), and `stop()` verifies the recorded tab label then closes it without an agent-list lookup (`src/herdr.ts:498-517`), with observable fallback/miss handling at `src/herdr.ts:530-539`. The socket transport implements the same tab-list path (`src/herdr-socket-driver.ts:73-75`, `src/herdr-socket-driver.ts:147-150`, `src/herdr-socket-driver.ts:210-215`). The serialized, self-confirming sweeper is wired into startup (`src/tab-reaper.ts:265-307`, `src/index.ts:905-928`), and the pane-list failure is surfaced (`src/index.ts:910-918`).

- `src/transient-helper-lifecycle.ts:1-51` extracts common tmpdir/finalizer plumbing for four helper modules; this refactor is outside the approved plan's listed loci but does not introduce a new external dependency or public API.

Could not verify the live Herdr `tab.list` RPC/CLI reply schema from this repository alone; the implementation's expected result shape is exercised by the in-repo transport tests (`test/herdr.test.ts:418-429`, `test/herdr-socket-driver.test.ts:420-432`).

PR #1856, opened 2026-07-21
feat(ui): redesign the gear menu as a telemetry popover + rising sheet

kai-osthoff:
Recreates the gear-menu design handoff (3b desktop popover · 3c mobile bottom sheet) — the reference ships tracked in this PR at `docs/design/gear-menu-3b-3c/` (README.md + reference.html, operator-requested during the plan gate).

## What changed

**Desktop (3b):** the gear now **always** toggles a square 300px telemetry popover (anchored **non-modal dialog** — no `role=menu`/`menuitem`, per the AutomationPanel/EmojiPicker precedent): identity header on `--color-head` (brand · real `build-info` version · connection readout), **Halt herd** hero row with a live `N WORKING` chip (two-step arm→confirm kept; natively `disabled` + chip hidden at 0), a **token-usage gauge line** (hottest window across providers, `CC·5H`/`CC·WK`/`CX·5H`/`CX·WK` + `% free`) with an `all ▾` inline per-window disclosure (Claude windows, per-model, credits, Codex windows + token lines, refresh), workspace rows (Learnings + count, Settings + platform chord hint, Documentation ↗), a counted **PLUGINS** group (`manage ▾` → Settings → Plugins; per-row hint derived from the existing `PluginInfo.name`, no SDK change), and a demoted **SUPPORT** group (`⚠ ✧ ↵` glyphs replace the emoji rows). Popover clamps to the space below its anchor and scrolls as one surface on short viewports.

**Mobile (3c):** same structure as a rising sheet — 12px top radius, grab handle with **pointer-captured swipe-down dismiss** (close iff dy > 64px; pointer-id filtered; unified up/cancel/lost-capture cleanup; drag offset is the only inline transform, inert until `introend`), 52/48/44px touch tiers — while keeping the sheet-only extras restyled (theme quick row, conditional diagnostics/update/What's-New attention rows, GitHub repo link, explicit ✕).

**Global shortcut:** `⌘,`/`Ctrl+,` opens Settings — `isSettingsChord` + `settingsChordHint` live beside `isCommandBarChord` in `herd-keynav.ts` (platform via the existing `isMacPlatform()`); the listener sits in TopBar's `<svelte:window>` handler gated by a `settingsChordAllowed` prop (`+page` passes `() => !anyOverlayOpen()`), and Viewport's PTY key handler suppresses the chord so no byte leaks to the terminal.

[...]

PR #1863, opened 2026-07-21
fix(plan-gate): anchor the plan reviewer to the planner's tree

scoop:
## Problem

The plan reviewer detached at the **freshest `origin/<base>`**, while the planning agent read its own session worktree — cut from `origin/main` whenever the session started. Different trees, and neither prompt named a commit. So both sides cited `file:line` from different ground truth, both were "verifiably right", and rounds were spent re-numbering instead of improving the plan.

## The observation is real, and it's this gate

Read-only measurement over `reviewer_spawns` (1035 sessions, 90-day retention window, oldest surviving row 2026-06-10):

| Metric | plan_gate | review (PR critic) |
| --- | --- | --- |
| Spawns / sessions | 3316 / 1035 | 2269 / 1156 |
| Mean per session | **3.20** | 1.96 |
| Max | **35** | 18 |
| Sessions at ≥10 rounds | **29** | 6 |

Top-decile streaks are plan-gate dominated (35/18, 29/4, 27/1, 27/0). At default caps neither gate reaches 10-12 alone (plan-gate 5, PR-critic `2*cap`=6), so those counts imply raised caps or `resume()` budget resets — which is why spawn rows, not `plan_gates.round`, are the metric: `plan_gates` is one upserted row per session and `buildGate` zeroes `round` + empties `findings` on approval, destroying the evidence of every session that passed.

**§9a baseline for post-merge comparison: mean 3.20 plan_gate spawns/session.**

## Change

**Anchor** — `planAnchorSha` resolves the merge-base of the session worktree's `HEAD` and `origin/<base>`, so both sides read identical bytes for pre-existing files. Not raw `HEAD`: that tree is **untrusted** — it carries whatever the planner committed, including `CLAUDE.md`, `.claude/settings.json`, hooks and skills, and the reviewer launches *in that directory*, so the agent under review would supply the reviewing agent's own instructions. (`scrubStaleVerdictArtifacts` already closes the pre-seeded-verdict hole; scrubbing one filename does nothing about agent config.) Falls back to the previous chain flagged `anchored: false`.

**Three prompt tiers, by what the reviewer can actually see:**

| | strong (`ahead=0`) | ahead (`ahead>0`) | degraded (no anchor) |
| --- | --- | --- | --- |
| co-location claim | emitted | withheld | withheld |
| unresolvable ref to existing code | **is a finding** | `body`, not `findings` | `body`, not `findings` |
| ref to a newly-proposed symbol | never a finding | never a finding | never a finding |
| line-number / precision findings | forbidden | forbidden | forbidden |

The `ahead` tier exists because merge-base excludes the planner's own commits by construction — on a multi-round session, round 3's plan cites round 1's committed scaffolding, and without it the resolution rule manufactures junk findings on exactly the long sessions this targets.

**Strip** — `stripPlanLineRefs` removes `path.ext:NNN` / `#LNNN` from **the plan text only**. Deliberately not from `task`/`issueBody` (human-authored ground truth the reviewer judges against) and not from findings: `planReviewPrompt` orders prior findings re-raised **verbatim**, so rewriting one turns "the ref `x.ts:12` points at the wrong function" into an un-addressable "the ref `x.ts` points at the wrong function" that the next round recycles forever. The output rule is scoped to *newly authored* findings and carries an explicit re-raise exemption, so the two instructions can't contradict each other for gates already in `changes_requested` at rollout.

**Staleness** — `behind`/`changedSince` are computed **after** `createDetached`'s fetch. Before it, they measured a stale local ref and systematically understated, usually to `behind: 0`, silently suppressing the note. Routed to `body` under a fixed heading, never a finding — the plan gate is the only pre-execution gate, so it should be able to *inform* about a plan targeting code `main` has since restructured, without gaining a new way to block.

**Directives** — both plan-gate directives now tell the planner to reference code by path + symbol.

## Notes for review

- The regex is narrow on purpose and its gaps are asserted to **survive** (`Makefile:88`, `foo.ts (line 411)`, bare `:1090`), so the trade-off is pinned by tests rather than prose. Self-review caught two live false positives on `https://example.com:443/x` — `.com` reads as an extension and `/` is a legal token char — fixed by a token-boundary lookbehind.
- `buildTransientAgentArgv` gains an optional `sessionId`. The plan gate needs the id *before* it can build the prompt (the id keys the worktree path; the prompt depends on that worktree's post-fetch state). Additive, default unchanged, no other caller passes it.
- `begin()` crossed the fallow complexity threshold with the added branches; prompt assembly is extracted to `composeReviewerPrompt`, which also documents the call-order requirement at the place that could break it.
- Server-only, no UI/i18n/catalog surface — `fix:` does not arm the feature-catalog gate.
- One flaky UI failure was seen on the first push attempt (`TopBar.browser.test.ts`, `expected 51.99998474121094 to be >= 52`) and passed clean on re-run. This PR touches no `ui/` files.

## Verification

`bun run lint`, `bun test` (7392 pass / 0 fail, +25 new), `cd ui && bun run test` (4171 pass), `check:i18n`, `check-glossary`, branch + feature-catalog hygiene, `fallow@2.100.0 audit` — all green.

### DIFF
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -620,7 +620,7 @@ export async function defaultComputePatchId(
     // Resolve the base to a concrete immutable SHA NOW: FETCH_HEAD is transient (a later
     // in-worktree fetch moves it; undefined on a failed fetch), so capturing the rev-parsed
     // SHA gives the prompt + backstop a base that provably equals the one we fingerprint.
-    // `--end-of-options` guards a hostile ref (mirrors defaultBaseSha in plan-gate.ts). Null
+    // `--end-of-options` guards a hostile ref (mirrors defaultPlanAnchorSha in plan-gate.ts). Null
     // on failure → caller diffs the `ref` string best-effort and skips the backstop.
     let baseSha: string | null = null;
     try {
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -283,7 +283,12 @@ function locationReferenceBlock(
   }
   lines.push(
     "- Files, symbols, tests and message keys the plan proposes to ADD, RENAME or MOVE are NEVER findings on resolution grounds. A plan describes code that does not exist yet.",
-    "- Identify code by FILE PATH + SYMBOL (function / type / const / message key). A line number, a line range, or the precision of a location reference is NEVER a finding — line numbers have been removed from the plan you were shown and are not part of the contract.",
+    // Deliberately NOT "line numbers have been removed": stripPlanLineRefs only removes refs
+    // attached to an extension-bearing path, so `Makefile:88` and prose like `foo.ts (line 411)`
+    // reach the reviewer intact. Promising exhaustive removal would be visibly false the first
+    // time one survives, which discredits the rule it is meant to justify. The load-bearing claim
+    // is that line numbers are not part of the contract — true whatever survived.
+    "- Identify code by FILE PATH + SYMBOL (function / type / const / message key). A line number, a line range, or the precision of a location reference is NEVER a finding — line numbers are not part of the contract, and any you do see in the plan carry no authority.",
     '- When you AUTHOR A NEW finding, cite path + symbol, not line numbers, in "findings" and "body".',
     "- EXEMPTION: re-raising a prior finding verbatim is REQUIRED by the re-review instruction above and OVERRIDES the previous rule. Reproduce its text unchanged, line numbers included; a re-raised prior finding is not itself a location complaint, and must not be dropped, reworded or reclassified on those grounds.",
   );
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -5,7 +5,7 @@ import {
   codexLastMessageFile,
 } from "./codex-last-message";
 import { join } from "node:path";
-import { createHash } from "node:crypto";
+import { createHash, randomUUID } from "node:crypto";
 import { execFileSync } from "./instrument";
 import type { SessionStore } from "./store";
 import type { HerdrDriver } from "./herdr";
@@ -83,15 +83,86 @@ const PLAN_BLOCKS_FILE = ".shepherd-plan-blocks.json";
 /** The file the adversarial plan reviewer writes its verdict JSON to, in its detached worktree. */
 export const PLAN_VERDICT_FILE = ".shepherd-plan-review.json";
 
+/** Rewrite `path.ext:120` / `path.ext:120-140` / `path.ext#L120-L140` to the bare `path.ext`.
+ *
+ *  Applied ONLY to the plan text embedded in {@link planReviewPrompt} — the plan file on disk, the
+ *  UI's copy and the planning agent's own view are untouched, and so are `task` / `issueBody` /
+ *  prior findings (see below). It exists because the planner and the reviewer historically read
+ *  DIFFERENT trees, so a line number correct on one side was wrong on the other and the round was
+ *  spent arguing about it. `planAnchorSha` now co-locates the two trees, which fixes the common
+ *  case; this strip is the deterministic backstop for the case the anchor cannot reach — a plan
+ *  reference into code the planner has itself committed this session, which the merge-base anchor
+ *  deliberately excludes. The prompt FORBIDS line-number findings; this makes them unraisable.
+ *
+ *  Deliberately NOT applied to:
+ *  - `task` / `issueBody` — human-authored ground truth the reviewer judges the plan against,
+ *    fixed across rounds. Stripping them cost the reviewer the ability to check the plan targets
+ *    the location the operator actually named, and bought nothing: no round-over-round churn
+ *    originates there.
+ *  - prior findings — `planReviewPrompt` orders them re-raised VERBATIM. Mutating one turns
+ *    "the ref `x.ts:12` points at the wrong function" into an un-addressable "the ref `x.ts`
+ *    points at the wrong function" that the next round recycles, LENGTHENING the loop.
+ *
+ *  The pattern is narrow ON PURPOSE: it requires a path-ish token bearing a file EXTENSION
+ *  immediately before the ref, so `10:30` and `--flag=3:4` are left alone. The leading lookbehind
+ *  forces the match to begin at a real token boundary, which is what keeps a URL authority safe:
+ *  in `https://example.com:443/x`, `example.com` is preceded by `/` and the `//` is preceded by
+ *  `:`, so neither can start a match and the port is not mistaken for a line number. (`.com` is
+ *  otherwise a perfectly good "extension", and `/` is a legal token char, so BOTH exclusions are
+ *  needed — this was a live false positive, caught twice by the guard test.)
+ *
+ *  Known survivals (bare `:1090` continuations, `foo.ts (line 411)`, extension-less paths like
+ *  `Makefile:88`) are accepted — widening this would trade a rare false negative for a common
+ *  false positive on timestamps, ports and ratios. Those are covered by the prompt rule instead. */
+export function stripPlanLineRefs(text: string): string {
+  return text.replace(
+    /(?<![A-Za-z0-9_.:\-/\\])([A-Za-z0-9_.\-/\\]*[A-Za-z0-9_\-/\\]\.[A-Za-z0-9]{1,8})(?::\d+(?:-\d+)?|#L\d+(?:-L?\d+)?)\b/g,
+    "$1",
+  );
+}
+
+/** What the reviewer's detached worktree is checked out at, when `planAnchorSha` resolved the
+ *  merge-base (`anchored`). `sha` is that commit; `ahead` is how many commits the planning agent
+ *  has made past it in its own worktree — the count that decides whether an unresolvable symbol
+ *  reference is a defect or merely invisible from here. */
+export interface PlanAnchor {
+  sha: string;
+  ahead: number;
+}
+
+/** How far the anchor sits behind the (freshly fetched) base branch, and which paths moved.
+ *  Computed AFTER `createDetached`'s fetch — see {@link defaultAnchorStaleness}. */
+export interface AnchorStaleness {
+  behind: number;
+  changedSince: string[];
+  /** Paths elided by the cap, so the prompt can say "and N more" instead of implying the list is
+   *  exhaustive. */
+  more?: number;
+}
+
 /** Self-contained instructions for the adversarial plan reviewer. NOT UI chrome — never i18n'd.
  *  The plan text is UNTRUSTED agent output embedded as data; the read-only dontAsk sandbox
- *  (mirrors the PR critic) contains any injection. */
+ *  (mirrors the PR critic) contains any injection.
+ *
+ *  `anchor` / `staleness` are optional and TRAILING on purpose: both existing callers pass
+ *  `operatorLanguage` positionally, so a parameter inserted before it would silently mis-bind.
+ *
+ *  Three tiers, keyed on what this reviewer can actually SEE (see the LOCATION REFERENCES block):
+ *  - `anchor` set, `ahead === 0` — the strong tier. Its tree IS the planner's tree for every
+ *    pre-existing file, so an unresolvable reference to existing code is a real defect.
+ *  - `anchor` set, `ahead > 0` — the planner has committed since the branch point and the anchor
+ *    (a merge-base) excludes exactly those commits, so a reference may name code the planner has
+ *    ALREADY written this session. Ambiguous ⇒ `body`, never `findings`.
+ *  - `anchor` absent — the anchor could not be resolved, so the tree may have drifted. Same
+ *    ambiguity, same routing. Mirrors critic-core's degraded epic block. */
 export function planReviewPrompt(
   task: string,
   plan: string,
   priorFindings: string[] = [],
   issueBody?: string | null,
   operatorLanguage: OperatorLanguage = "en",
+  anchor?: PlanAnchor | null,
+  staleness?: AnchorStaleness | null,
 ): string {
   const lines = [
     "You are an adversarial plan reviewer. Read-only — do NOT modify, build, commit, or run anything.",
@@ -118,7 +189,9 @@ export function planReviewPrompt(
       "",
     );
   }
-  lines.push("PLAN (.shepherd-plan.md):", plan, "");
+  // Line refs are stripped from the PLAN only — never from `task`/`issueBody` above (human-authored
+  // ground truth) nor from the prior findings below (re-raised verbatim). See stripPlanLineRefs.
+  lines.push("PLAN (.shepherd-plan.md):", stripPlanLineRefs(plan), "");
   if (priorFindings.length) {
     lines.push(
       "This is a RE-REVIEW. For EACH prior point, confirm the revised plan addresses it; if it does not, re-raise it verbatim:",
@@ -126,6 +199,7 @@ export function planReviewPrompt(
       "",
     );
   }
+  lines.push(...locationReferenceBlock(anchor, staleness), "");
   lines.push(
     `Write your verdict as JSON to \`${PLAN_VERDICT_FILE}\` in the current directory, with EXACTLY this shape:`,
     '{"decision": "approve" | "request-changes", "summary": "<=100 char one-liner", "body": "<full markdown>", "findings": ["<discrete actionable revision>", ...]}',
@@ -143,6 +217,59 @@ export function planReviewPrompt(
   return lines.join("\n");
 }
 
+/** The LOCATION REFERENCES block: how the reviewer must treat WHERE the plan says code lives.
+ *
+ *  Two tiers of tree-claims (conditional) wrapped around one set of referencing rules
+ *  (unconditional, byte-identical in every tier — they never depend on which commit is checked
+ *  out, only the claims ABOUT the tree do).
+ *
+ *  The rules exist to kill three distinct manufactured-finding classes, each observed or
+ *  predicted: (1) arguing about a line number that is stale only because the two sides read
+ *  different trees; (2) flagging a symbol the plan PROPOSES TO CREATE as "unresolvable" — every
+ *  plan describes code that does not exist yet; (3) flagging a symbol the planner has ALREADY
+ *  COMMITTED this session, which the merge-base anchor excludes by construction and which bites
+ *  hardest on exactly the long multi-round sessions this whole change targets. */
+function locationReferenceBlock(
+  anchor?: PlanAnchor | null,
+  staleness?: AnchorStaleness | null,
+): string[] {
+  const lines = ["LOCATION REFERENCES — how to treat WHERE the plan says code lives:"];
+  if (anchor) {
+    lines.push(
+      `- Your worktree is checked out at \`${anchor.sha}\`, the merge-base the planning agent's own worktree derives from — you both share that commit exactly. Divergence from newer commits on the base branch is NOT a finding.`,
+    );
+    if (anchor.ahead > 0) {
+      // NOTE the claim withheld here: with commits past the anchor, a pre-existing file the agent
+      // has since edited does NOT read the same on both sides, so "reads IDENTICALLY" would be a
+      // lie in this tier — and a lie that licenses exactly the false findings this block prevents.
+      lines.push(
+        `- The planning agent has made ${anchor.ahead} commit(s) SINCE that merge-base, and this checkout deliberately excludes them. So a file or symbol that does not resolve here — or that reads differently than the plan describes — may be code the agent has ALREADY WRITTEN this session, not a mistake. Such a reference is AMBIGUOUS: report it in "body" as a stated limitation, NEVER in "findings".`,
+      );
+    } else {
+      lines.push(
+        "- The planning agent has committed nothing since, so every file reads IDENTICALLY for both of you. A reference to code that ALREADY EXISTS naming a file or symbol you cannot resolve IS therefore a finding — a wrong reference, not an imprecise one.",
+      );
+    }
+  } else {
+    lines.push(
+      '- The commit your worktree sits at could NOT be tied to the planning agent\'s tree, so the two may have drifted apart. A file or symbol that does not resolve here is therefore AMBIGUOUS — it may be a genuine mistake, or it may have moved or been renamed since this session started. Do NOT raise it as a finding; report it in "body" as a stated limitation.',
+    );
+  }
+  if (staleness && staleness.behind > 0 && staleness.changedSince.length) {
+    const more = staleness.more ? `, and ${staleness.more} more` : "";
+    lines.push(
+      `- Your anchor is ${staleness.behind} commit(s) behind the base branch. These paths have changed on the base branch since (some may not exist in your checkout at all): ${staleness.changedSince.join(", ")}${more}. If the plan targets any of them, you MAY note it — but ONLY in "body", in a single section headed exactly \`Anchor staleness (informational, non-blocking):\`. It is NEVER a finding and NEVER makes the decision "request-changes".`,
+    );
+  }
+  lines.push(
+    "- Files, symbols, tests and message keys the plan proposes to ADD, RENAME or MOVE are NEVER findings on resolution grounds. A plan describes code that does not exist yet.",
+    "- Identify code by FILE PATH + SYMBOL (function / type / const / message key). A line number, a line range, or the precision of a location reference is NEVER a finding — line numbers have been removed from the plan you were shown and are not part of the contract.",
+    '- When you AUTHOR A NEW finding, cite path + symbol, not line numbers, in "findings" and "body".',
+    "- EXEMPTION: re-raising a prior finding verbatim is REQUIRED by the re-review instruction above and OVERRIDES the previous rule. Reproduce its text unchanged, line numbers included; a re-raised prior finding is not itself a location complaint, and must not be dropped, reworded or reclassified on those grounds.",
+  );
+  return lines;
+}
+
 /** The read-only plan reviewer's argv — the PR critic's exact hardening, shared via one builder
  *  (the plan text is UNTRUSTED, so it gets the same injection-contained sandbox). Also returns
  *  the reviewer's pinned `--session-id` so begin() can locate its transcript for token totals. */
@@ -151,6 +278,7 @@ export function reviewerArgv(
   model: string | null,
   prompt: string,
   effort: string | null = null,
+  sessionId?: string,
 ): { argv: string[]; sessionId: string } {
   // The plan reviewer participates in the session's resolved reasoning effort (no longer force-
   // unbudgeted): it inherits `session.effort` — the tier this session runs at, itself the session
@@ -162,6 +290,7 @@ export function reviewerArgv(
     prompt,
     effort,
     captureLastMessage: true,
+    sessionId,
   });
 }
 
@@ -270,8 +399,12 @@ export interface PlanGateServiceDeps extends MembraneSeams {
   worktreeExists?: (worktreePath: string) => boolean;
   /** default: read PLAN_VERDICT_FILE from the reviewer's disposable worktree. */
   readVerdict?: (worktreePath: string, spawnSessionId?: string) => RawPlanVerdict | null;
-  /** default: `git rev-parse origin/<base>` (fallback `<base>`) in the repo. */
-  baseSha?: (repoPath: string, base: string) => string;
+  /** default: {@link defaultPlanAnchorSha} — the merge-base of the session worktree's HEAD and
+   *  `origin/<base>`, falling back to `origin/<base>` → `<base>` → the bare ref name. */
+  baseSha?: (repoPath: string, base: string, worktreePath: string) => PlanAnchorResolution;
+  /** default: {@link defaultAnchorStaleness}. MUST be called AFTER the reviewer worktree is
+   *  created — `createDetached` fetches `origin/<base>` and this measures against it. */
+  anchorStaleness?: (repoPath: string, sha: string, base: string) => AnchorStaleness;
   /** Injectable reader for the plan reviewer's latest tool-use summary (default: parse its JSONL
    *  transcript via readActivitySignal). null = no parseable activity yet. */
   readActivity?: (worktreePath: string, reviewerSessionId: string) => string | null;
@@ -331,7 +464,8 @@ export class PlanGateService {
   private readPlanBlocks: (worktreePath: string) => VisualBlock[];
   private readVerdict: (worktreePath: string, spawnSessionId?: string) => RawPlanVerdict | null;
   private worktreeExists: (worktreePath: string) => boolean;
-  private baseSha: (repoPath: string, base: string) => string;
+  private baseSha: (repoPath: string, base: string, worktreePath: string) => PlanAnchorResolution;
+  private anchorStaleness: (repoPath: string, sha: string, base: string) => AnchorStaleness;
   private readActivity: (worktreePath: string, reviewerSessionId: string) => string | null;
   private readUsage: (
     worktreePath: string,
@@ -348,7 +482,8 @@ export class PlanGateService {
     this.readPlanBlocks = deps.readPlanBlocks ?? defaultReadPlanBlocks;
     this.readVerdict = deps.readVerdict ?? defaultReadVerdict;
     this.worktreeExists = deps.worktreeExists ?? existsSync;
-    this.baseSha = deps.baseSha ?? defaultBaseSha;
+    this.baseSha = deps.baseSha ?? defaultPlanAnchorSha;
+    this.anchorStaleness = deps.anchorStaleness ?? defaultAnchorStaleness;
     this.readActivity = deps.readActivity ?? defaultReadActivity;
     this.readUsage = deps.readUsage ?? readSessionUsage;
   }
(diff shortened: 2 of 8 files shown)


scoop (commented):
## What I verified

**Anchor resolution & ordering (the core of the change).**
- `defaultPlanAnchorSha` (`src/plan-gate.ts:1508`) runs `git merge-base --end-of-options HEAD origin/<base>` in the *session* worktree (`src/plan-gate.ts:1525`) and falls back to `rev-parse origin/<base>` → `<base>` → the bare name with `anchored: false` (`src/plan-gate.ts:1535`). Every git call is wrapped in a try/catch returning `null`, so a missing worktree, an unborn HEAD, unrelated histories, or a repo without `origin` all degrade rather than throw — matching the `begin()` comment at `src/plan-gate.ts:573-578` that promises it won't throw. The real-git tests at `test/plan-gate-anchor.test.ts:43`, `:67`, `:89` exercise exactly those three paths.
- The merge-base is always an ancestor of `origin/<base>`, so the reviewer still detaches at a commit on the base branch's history — the trust property `src/plan-gate.ts:692-694` (`scrubStaleVerdictArtifacts`, "detaches at the TRUSTED base sha") relies on. The doc's argument against a raw-HEAD anchor (`src/plan-gate.ts:1494-1500`) is sound.
- Ordering: `begin()` mints `reviewerSessionId` (`src/plan-gate.ts:598`), calls `createDetached` (`:608`), and only then builds the prompt (`:622`) and argv (`:627`). `defaultAnchorStaleness` (`src/plan-gate.ts:1553`) therefore measures against the `origin/<base>` refreshed by `createDetached`'s `git fetch origin -- <branch>` (verified at `src/worktree.ts:632-639`). The pre-fetch blind spot is pinned by `test/plan-gate-anchor.test.ts:105` and the order assertion at `test/plan-gate-service.test.ts:2064`.
- Session id threading: `reviewerArgv` now accepts a pre-minted id (`src/plan-gate.ts:281`) forwarded to `buildTransientAgentArgv` (`src/transient-agent-argv.ts:169`, `opts.sessionId ?? randomUUID()`), which uses it for `--session-id` (`src/transient-agent-argv.ts:212`) and for the codex per-spawn `-o` name (`:192`). `createDetached` validates the slug as `/^[A-Za-z0-9_-]{1,128}$/` (`src/worktree.ts:627`) — a UUID passes. The argv/worktree/spawn-record parity is asserted at `test/plan-gate-service.test.ts:397-401`.
- Signature change fan-out: `PlanGateServiceDeps.baseSha` and the new `anchorStaleness` (`src/plan-gate.ts:404`, `:407`) are injected nowhere in production — I grepped the tree and the only construction site is `src/index.ts:1367`, which passes neither, so both take the new defaults. The only stubs are `test/plan-gate-service.test.ts:85-86`, both updated to the new shape.

**Strip regex** (`src/plan-gate.ts:117`). I traced the lookbehind `(?<![A-Za-z0-9_.:\-/\\])` by hand against the guard cases: in `https://example.com:443/x`, `example.com` is preceded by `/` and every interior start position by an alnum, so no match can begin — the port survives, as `test/plan-gate-prompt.test.ts:191` asserts. Full URLs with `#L20-L40` fragments are immune for the same reason. The documented survivals (`Makefile:88`, bare `:1090`, prose) are pinned at `test/plan-gate-prompt.test.ts:197-204`.

**Inbound-only application.** `stripPlanLineRefs` is called at exactly one place, `src/plan-gate.ts:193`, on the plan text; `task`, `issueBody` (`:181-187`) and prior findings (`:195-200`) are untouched, and nothing in `resolveFindings`/`normalizeFindings`/`resolveSummary` changed. Both `resolveFindings` branches are covered by `test/plan-gate-service.test.ts:410` and `:425`.

**Directives** (`src/service.ts:1051`). `PLAN_REFERENCE_STYLE` is declared before its two uses (`:1098`, `:1129`) and before the module-level constant evaluations at `:1104`/`:1136`, so no TDZ. It lands in the shared body of both builders, so `claude`/`codex` receive it identically — asserted at `test/service-plan-gate.test.ts:472`. `src/plan-gate.ts` still has no `./config` import, so the guard at `test/operator-language.test.ts:301` is unaffected.

**No i18n surface.** No `ui/` file is touched and the prompt/directive strings are agent-facing English (the existing `de` handling at `src/plan-gate.ts:206-215` is a language *instruction*, not a catalog entry), so no message-catalog parity is at stake.

## Could not verify

- I was denied shell access, so I could not empirically confirm that this host's git accepts `--end-of-options` on `merge-base`, `rev-list` and `diff` (the pre-existing precedent at `src/critic-core.ts:628` only covers `rev-parse`). It is documented as generally supported since git 2.24 and `test/plan-gate-anchor.test.ts` exercises all three against a real repo, so CI will settle it — assumed as written, not asserted correct.

## Findings

1. **Strong-tier claim overclaims what `ahead === 0` proves** (`src/plan-gate.ts:250`). `ahead` is `rev-list --count <mergeBase>..HEAD` (`src/plan-gate.ts:1531`) — it counts *commits* only, so uncommitted work in the planner's worktree is invisible to it. "every file reads IDENTICALLY for both of you" is therefore false whenever the planner's tree is dirty, which during the plan gate it always is at minimum (`.shepherd-plan.md` itself is uncommitted — it is read off the live worktree by `readPlan`), and can include product code: `advanceToExecutionOnPr` exists precisely because agents write code during `planning`, and the author cites that themselves at `src/plan-gate.ts:1491-1493`. The absolute claim is immediately followed by the one *blocking* rule in the whole block ("IS therefore a finding"), so a plan referencing a symbol that exists only in the planner's uncommitted edits gets a blocking finding — the exact manufactured-finding class this PR exists to kill. The author's own comment at `src/plan-gate.ts:242-244` calls this phrasing "a lie that licenses exactly the false findings this block prevents" for the `ahead > 0` branch, and the approved plan explicitly chose "pre-existing code reads identically" over "the same bytes for every file" for this reason. Narrowing the claim to committed/pre-existing code costs nothing.

2. **Stale cross-reference left by the rename** (`src/plan-gate.ts:1508`). `defaultBaseSha` no longer exists, but `src/critic-core.ts:623` still says "`--end-of-options` guards a hostile ref (mirrors `defaultBaseSha` in plan-gate.ts)". The dangling name is caused by this diff's rename; point it at `defaultPlanAnchorSha`.

3. **Nit — an unlisted strip false-positive class** (`src/plan-gate.ts:117`). The path token only requires *some* non-dot char, a dot, and 1-8 alnum, so a decimal-bearing ratio or version immediately before a colon matches: `16.9:1` → `16.9`, `1.5:30` → `1.5`. The false-positive guard at `test/plan-gate-prompt.test.ts:185-195` covers `16:9` and `--flag=3:4` but not the decimal form. Cosmetic (it only mangles prose in the reviewer's copy of the plan), but the gap table in the doc comment at `src/plan-gate.ts:110-112` claims ratios are safe, which is not quite true — either tighten the extension class (e.g. require a leading letter) or add the case to the documented-survivals/known-limits list.

- `src/plan-gate.ts:1531` — `ahead` is `Number(...)` of git output with an `?? "0"` fallback and an `isFinite` floor, but `defaultAnchorStaleness`'s `behind` (`src/plan-gate.ts:1568`) has the same floor plus a `<= 0` check while `ahead` has no negative/zero floor; `rev-list --count` cannot emit a negative, so this is dormant, but the two sibling parsers guard differently.

- The anchor's trust argument rests on `origin/<base>`, a *local* remote-tracking ref that the planning agent shares an object store with and could rewrite before the gate runs; the prior `rev-parse origin/<base>` had the identical exposure, so this diff neither introduces nor widens it.
- `src/plan-gate.ts:1535`'s final `?? base` rung still returns a branch name that `createDetached`'s `/^[0-9a-fA-F]{7,40}$/` check (`src/worktree.ts:616`) rejects, so it can only ever produce `error-worktree`; the plan deliberately keeps this dead rung.

- `src/transient-agent-argv.ts:93` adds a public `sessionId` opt to a shared builder for a single caller; it is well-documented and defaulted, so behaviour for every other call site is unchanged.

scoop:
**Flaky UI test, pre-existing on `main` — not from this PR.**

`ui/src/lib/components/TopBar.browser.test.ts` → *"mobile: 12px-top-radius full-bleed sheet with the 52/48/44px row tiers"* fails intermittently:

```
AssertionError: expected 51.99998474121094 to be greater than or equal to 52
```

It has now blocked two of three push attempts on this branch and passed clean on re-run both times (272 files / 4171 tests green). Flagging rather than silently retrying, since an intermittent gate failure on `main` will bite other branches too.

- **Not this PR:** `git diff --name-only origin/main..HEAD | grep ^ui/` → **0 files**. This branch touches no UI code at all.
- **Origin:** the assertion (`TopBar.browser.test.ts`, `toBeGreaterThanOrEqual(52)`) came in with ffdd11fc9 *"feat(ui): redesign the gear menu as a telemetry popover + rising sheet (#1856)"*, which is on `main` and predates this branch point.
- **Cause:** it asserts an exact `52` floor against a `getBoundingClientRect().height` that lands on `51.99998474121094` — subpixel/device-pixel-ratio rounding in headless Chromium. The neighbouring geometry assertions in the same test allow `±1` tolerance (`toBeLessThanOrEqual(1)`); the three touch-tier floors do not.

Not fixing it here — it is out of this PR's scope and would mix an unrelated UI change into a server-only diff. It wants a small follow-up: either compare against a rounded height or allow a sub-pixel epsilon on the 52/48/44 floors, keeping the a11y intent (the tap target is 52px by design; 51.99998 is the same rendered pixel).

scoop (commented):
## What I verified

**Anchor resolution (`src/plan-gate.ts:1519`).** `defaultPlanAnchorSha` runs `git merge-base --end-of-options HEAD origin/<base>` in the *session* worktree, returns `{sha, anchored, ahead}`, and falls back to `origin/<base>` → `<base>` → the bare name with `anchored:false`. Verified against the real-git suite at `test/plan-gate-anchor.test.ts:37` (same sha with and without a planner commit, `ahead: 1` after the commit), `:62` (origin-less repo ⇒ `anchored:false`, still a 40-hex sha so `createDetached`'s `/^[0-9a-fA-F]{7,40}$/` guard at `src/worktree.ts:616` is satisfiable) and `:83` (non-repo path never throws — every git call is individually try/caught, so a pruned `session.worktreePath` degrades rather than blowing up `begin()`). That real-git coverage is also what confirms `--end-of-options` is accepted by `merge-base`/`rev-list`/`diff`, not just `rev-parse` — I did not take that on trust.

**Trust posture.** The doc's argument for merge-base-not-HEAD holds: `merge-base(HEAD, origin/<base>)` is by definition an ancestor of `origin/<base>`, i.e. a commit on the base branch's history, so the reviewer's checkout can never pick up `CLAUDE.md` / `.claude/settings.json` the agent under review authored — the constraint `src/plan-gate.ts:707` (`scrubStaleVerdictArtifacts`, "detaches at the TRUSTED base sha") depends on. The one case where the anchor equals the planner's own commits is when those commits are already ancestors of `origin/<base>`, i.e. already merged into the base branch — the same trust level the pre-change `origin/<base>` checkout had. No regression.

**Staleness ordering (`src/plan-gate.ts:1564`, called from `composeReviewerPrompt` at `:556`).** `begin()` resolves the anchor at `:590`, calls `createDetached` at `:621` (whose `git fetch origin -- <branch>` is `src/worktree.ts:634`, which opportunistically updates `refs/remotes/origin/<base>`), and only then builds the prompt at `:632`. The pre-fetch blind spot is pinned by `test/plan-gate-anchor.test.ts:100` (pre-fetch ⇒ `behind: 0`) and the ordering itself by the call-order assertion in `test/plan-gate-service.test.ts:2100`. Cap/overflow behaviour matches `CHANGED_SINCE_CAP = 50` at `src/plan-gate.ts:1551`, asserted at `test/plan-gate-anchor.test.ts:126`.

**Session-id threading.** The prompt now has to be built *after* `createDetached`, but the worktree path is keyed on the reviewer session id — resolved by pre-minting the uuid at `src/plan-gate.ts:610` and passing it through `reviewerArgv` → `buildTransientAgentArgv` (`src/transient-agent-argv.ts:170`, `opts.sessionId ?? randomUUID()`). I checked the three consumers of that id still agree: the `--session-id` argv slot, the `createDetached` slug, and `recordReviewerSpawn`'s `reviewerSessionId` — asserted together at `test/plan-gate-service.test.ts:2153`. The Codex branch's per-spawn `-o` name (`src/transient-agent-argv.ts:190` → `codexLastMessageFile`) still receives the same uuid, so the unguessable-filename property of `src/codex-last-message.ts:51` is preserved.

**Contract changes have no stale consumers.** `defaultBaseSha` has zero remaining references anywhere in the tree (the only prose reference, `src/critic-core.ts:623`, was renamed in this diff). `PlanGateServiceDeps.baseSha` is injected nowhere in `src/` outside `plan-gate.ts` itself, so the widened 3-arg/object-returning signature has no out-of-diff caller; the test stub at `test/plan-gate-service.test.ts:85` was updated. `planReviewPrompt`'s two new params are strictly trailing, so the positional `operatorLanguage` in slot 5 still binds correctly (`test/plan-gate-prompt.test.ts:80`).

**Prompt tiers.** The three tiers, the unconditional carve-out / line-number ban / scoped output rule, and the re-raise exemption are all emitted from one block (`src/plan-gate.ts:238`) and asserted present in every tier at `test/plan-gate-prompt.test.ts:212`. The strong tier deliberately withholds the "every file reads identically" claim once `ahead > 0` and separately carves out the planner's *uncommitted* tree at `ahead === 0` — that second carve-out is the correct call, since the plan file itself is always a dirty edit. The re-raise exemption at `src/plan-gate.ts:257` genuinely resolves the contradiction with the verbatim re-raise instruction at `:200`, and findings are confirmed unmutated on **both** `resolveFindings` branches (`test/plan-gate-service.test.ts:2170`, `:2185`).

**Regex false-positive guard (`src/plan-gate.ts:120`).** I walked the pattern by hand against each case in `test/plan-gate-prompt.test.ts:105`. The extension-must-start-with-a-letter rule is what saves `16.9:1` / `1.5:30`, and the leading lookbehind (excluding `/` and `:`) is what saves `https://example.com:443/x` — both real, both correctly reasoned in the doc comment.

**Directives (`src/service.ts:1053`).** `PLAN_REFERENCE_STYLE` is spliced into the shared body of *both* builders (`:1099`, `:1130`), so `"claude"` and `"codex"` receive it identically; asserted at `test/service-plan-gate.test.ts:250`. No test anywhere asserts the old `"success criteria). An adversarial reviewer"` run-on that the new `\n` splits, so nothing breaks. `src/plan-gate.ts` still has no `./config` import, so the import-time-freeze guard at `test/operator-language.test.ts:305` stays green (`CHANGED_SINCE_CAP` is a module-local const, not a config key).

## Stated limitations

- The prompt tells the reviewer "line numbers have been removed from the plan you were shown", which is slightly stronger than what the regex guarantees — `Makefile:88`, `Dockerfile:12` and bare `:1090` continuations survive by design (`test/plan-gate-prompt.test.ts:132`). Harmless in practice (the ban on location findings is unconditional), but the sentence is an absolute claim about a best-effort transform.
- Whether this actually collapses the 10–12 round loop is an outcome the diff cannot demonstrate; the plan's own success criterion 9a is a post-merge `reviewer_spawns` measurement. Nothing here is verifiable at review time.

`src/transient-agent-argv.ts:96` — the new `sessionId` opt is unvalidated, so a future caller passing a non-uuid string would both weaken the "per-spawn unguessable name" property documented at `src/codex-last-message.ts:47` and flow into a filename via `codexLastMessageFile`; the sole caller today mints `randomUUID()` at `src/plan-gate.ts:610`, so it is dormant.
`src/plan-gate.ts:632` — prompt assembly moved *after* `createDetached`, so a throw from the injectable `anchorStaleness`/`operatorLanguage` seams would now leak the reviewer worktree instead of failing before it was created; the defaults cannot throw (both fully try/caught), and the pre-existing `resolveAuxSpawn` await at `:673` already had the same exposure.

scoop (commented):
## Prior finding — re-checked

**1. `stripPlanLineRefs` leaves a misleading residue on grep-style multi-colon refs; bare `host:port/path` also rewritten — ADDRESSED.**

Verified against the new pattern at `src/plan-gate.ts:131`:

```
/(?<![A-Za-z0-9_.:\-/\\])([A-Za-z0-9_.\-/\\]*[A-Za-z0-9_\-/\\]\.[A-Za-z][A-Za-z0-9]{0,7})(?::\d+(?:-\d+)?(?::\d+)*|#L\d+(?:-L?\d+)?)\b(?!\/)/g
```

- The trailing `(?::\d+)*` alternative is present, so `foo.ts:12:5` is consumed whole (`foo.ts`), not rewritten to `foo.ts:5`. Traced by hand and pinned by the test at `test/plan-gate-prompt.test.ts:103-110` (`foo.ts:12:5` → `foo.ts`; `src/a.rs:10:3: error` → `src/a.rs: error`; `src/a.ts:1:2:3` → `src/a.ts`).
- The `(?!\/)` trailing guard is present and does rescue the scheme-less registry case: for `ghcr.io:443/x`, `:443` is followed by `/` so the lookahead fails, and backtracking to `:44` fails the `\b` (next char `3` is a word char) — no match. Pinned at `test/plan-gate-prompt.test.ts:112-120`.
- The doc comment was corrected to say exactly what the previous round asked for: `src/plan-gate.ts:116-121` now states the lookbehind/URL-authority argument "only holds when a `//` scheme prefix is present, so the trailing `(?!/)` covers the rest".
- The residual bare `host:port` false positive (`ghcr.io:443` → `ghcr.io`) is documented at `src/plan-gate.ts:123-128` and asserted as accepted at `test/plan-gate-prompt.test.ts:122-129`. Cosmetic (the reviewer's copy of the plan loses a port); I agree with the trade — a TLD blocklist collides with `.rs`/`.sh`/`.pl`/`.ml`.

I also re-derived the false-positive guards independently: `10:30`, `--flag=3:4`, `16.9:1`, `1.5:30`, `v2.0:5` all fail because the extension must start with `[A-Za-z]` (`src/plan-gate.ts:131`), and `https://example.com:443/x` fails the token-boundary lookbehind. `.env:12`, `Makefile:88` and `foo.ts (line 411)` survive — documented gaps, governed by the prompt rule instead.

## Correctness checks (each against ground truth)

- **Signature/consumer agreement.** `PlanGateServiceDeps.baseSha` is widened to `(repoPath, base, worktreePath) => PlanAnchorResolution` at `src/plan-gate.ts:424`, field at `:487`, default at `:505`, call site at `:599-603`. Grepped every occurrence of `baseSha` in `src/` and `test/`: the only plan-gate consumers are `test/plan-gate-service.test.ts:85`, `:2090`, `:2126`, `:2142`, all returning the new object shape. `src/standalone-critic.ts`'s `baseSha` is an unrelated field on a different type (`standalone-critic.ts:60`), not this dep.
- **`reviewerArgv` callers.** Only `src/plan-gate.ts:647` and `test/plan-gate-prompt.test.ts:47,62`; the new trailing `sessionId?` param (`src/plan-gate.ts:296-311`) is optional so the two-and-three-arg test calls still bind. `buildTransientAgentArgv` honours it at `src/transient-agent-argv.ts:170` (`opts.sessionId ?? randomUUID()`) and the id flows identically into both the Claude `--session-id` slot (`:214-215`) and the Codex per-spawn last-message filename (`:193`), so the transcript/verdict lookup keyed on `reviewerSessionId` is unchanged. Pinned by `test/plan-gate-service.test.ts` ("the reviewer argv pins the same session id that keys the worktree path").
- **Ordering claim.** `begin()` mints the id at `src/plan-gate.ts:619`, calls `createDetached` at `:630`, then builds the prompt at `:642` and argv at `:647`. `createDetached`'s `git fetch origin -- <branch>` runs with `cwd: repoPath` (`src/worktree.ts:634-636`), and `defaultAnchorStaleness` also runs with `cwd: repoPath` (`src/plan-gate.ts:1585`) — same repo, so the staleness numbers really do see the freshly-fetched `origin/<base>`. The ordering itself is asserted at `test/plan-gate-service.test.ts` ("anchorStaleness runs AFTER createDetached").
- **Anchor resolution.** `defaultPlanAnchorSha` (`src/plan-gate.ts:1545-1560`) runs `merge-base` in `worktreePath` (a `git worktree` of `repoPath`, so shared object store) and falls back to the previous `origin/<base>` → `<base>` → bare-name chain with `anchored: false`. `Session.worktreePath` is non-nullable (`src/store.ts:368`), and a missing/non-repo path makes `execFileSync` throw into the `catch` → `null` → fallback, covered by `test/plan-gate-anchor.test.ts` ("never throws on a non-repo path"). `createDetached` hard-validates `/^[0-9a-fA-F]{7,40}$/` (`src/worktree.ts:616`), so the bare-name rung still degrades to `error-worktree` via `src/plan-gate.ts:636-639` — unchanged behaviour, as the doc at `:1533-1536` states.
- **`ahead` / `behind` parsing.** Both guard `Number.isFinite` before use (`src/plan-gate.ts:1553`, `:1588`) and default the git output to `"0"`, so an empty or failed read yields `0` rather than `NaN`. `git merge-base` without `--all` emits a single line, so `sha` is always one 40-hex token.
- **Prompt tiering.** `locationReferenceBlock` (`src/plan-gate.ts:244-291`) is unconditionally appended at `:214`, and the strong tier's blocking rule is correctly scoped to *committed* code with an explicit carve-out for the planner's dirty tree (`:186-189`) — that scoping is load-bearing, since `ahead` counts commits only and the plan file itself is always an uncommitted edit. `composeReviewerPrompt` (`:578-586`) passes `anchor: undefined` when not anchored and `staleness: undefined` when `behind === 0`, matching the block's own `staleness.behind > 0 && staleness.changedSince.length` guard at `:196`.
- **Localisation.** `planReviewPrompt` is English-only by design except the trailing output-language clause at `src/plan-gate.ts:220-228`; the new block adds no user-facing UI string and no message key, so there is no `en`/`de` catalog parity surface to check (confirmed: no `ui/` or `messages/` file appears in the diff).
- **Directives.** `PLAN_REFERENCE_STYLE` (`src/service.ts:520-523`) is spliced into the *shared* body of both builders (`:1098` interactive, `:1128` auto), not duplicated per provider branch, so `claude` and `codex` receive it byte-identically — the property the rewritten doc comments at `:531-538` and `:1106-1110` now describe. Asserted at `test/service-plan-gate.test.ts:250-258`.

## Could not verify

- `--end-of-options` support for `git merge-base` / `git rev-list` / `git diff` (`src/plan-gate.ts:1545`, `:1551`, `:1588`, `:1590`): this repo had no prior precedent for those three subcommands (only `rev-parse`, at `src/critic-core.ts:628`), and I was not permitted to execute git here. It is exercised end-to-end by the real-git tests in `test/plan-gate-anchor.test.ts`, so CI is the ground truth. Not asserting it is wrong — noting I did not confirm it locally.

- `src/transient-agent-argv.ts`: adds a new public `sessionId?` option to `TransientAgentArgvOptions` — not listed in the plan's Files table, though it is genuinely required by the mint-id-before-worktree reordering and defaults to the previous behaviour.
- `src/critic-core.ts`: one-word comment rename (`defaultBaseSha` → `defaultPlanAnchorSha`) — not in the plan's Files table, but a necessary consequence of the rename.

- `ui/src/lib/components/TopBar.browser.test.ts` asserts an exact `52` floor on a `getBoundingClientRect().height` that lands at `51.99998474121094` in headless Chromium; the PR author's note is consistent with the diff containing zero `ui/` files, so this is not attributable here.

scoop (commented):
## Prior-revision point

1. **`PLAN_REFERENCE_STYLE` overstated the strip guarantee** — **addressed.** The clause now reads "path-attached line refs are stripped from the plan the reviewer is shown" (`src/service.ts:1057-1060`), and the doc comment above it (`src/service.ts:1053-1056`) names the exact survivals. A regression guard was added asserting both the qualified form and the absence of the unqualified one (`test/service-plan-gate.test.ts:257-259`). Verified against the survivals it must not over-promise: `Makefile:88` / `Dockerfile:12` / `foo.ts (line 411)` / bare `:1090` are asserted to survive at `test/plan-gate-prompt.test.ts:151-157`, and the prompt-side wording deliberately avoids any removal claim (`src/plan-gate.ts:288`).

## What I verified

**Regex (`src/plan-gate.ts:129-131`).** Traced by hand against every case in the test table. `src/a.rs:10:3: error` → the trailing `(?::\d+)*` consumes `:10:3` whole and `\b` holds before `:` → `src/a.rs: error` (`test/plan-gate-prompt.test.ts:110`). `ghcr.io:443/x` survives: the greedy `\d+` cannot backtrack to a shorter form because `\b` fails mid-digits, and `(?!/)` kills the full form; `example.com` in `https://example.com:443/x` is blocked by the lookbehind on the preceding `/`. `16.9:1` / `v2.0:5` survive because the extension must start with `[A-Za-z]`. `x.svelte#L20-L40` is consumed whole (`-L?\d+` is greedy, no backtrack to `#L20`). No case in `test/plan-gate-prompt.test.ts:92-158` disagrees with the pattern as written.

**Anchor trust property (`src/plan-gate.ts:1533-1563`).** The merge-base of the session HEAD and `origin/<base>` is by definition an ancestor of `origin/<base>`, so the reviewer's checkout is always a commit on the base branch's history — the invariant `begin()` relies on (`src/plan-gate.ts:599-603`) is preserved, and the HEAD-anchor exposure the doc comment argues against is genuinely avoided. `createDetached` still gets a hex sha in every reachable path (validated at `src/worktree.ts:616`); the origin-less fallback is asserted to yield a 40-hex sha at `test/plan-gate-anchor.test.ts:77`.

**Ordering (`src/plan-gate.ts:624-655`).** The reviewer session id is minted at `:624`, passed to `createDetached` as the slug at `:640`, and threaded into `reviewerArgv` at `:652`; `buildTransientAgentArgv` honours it via `opts.sessionId ?? randomUUID()` (`src/transient-agent-argv.ts:170`), so the worktree path, the pinned `--session-id`, the transcript lookup and `recordReviewerSpawn` all still key off one value — pinned by `test/plan-gate-service.test.ts:2149-2162`. The builder is pure (no I/O, `src/transient-agent-argv.ts:165-235`), so moving the argv call after `createDetached` has no side-effect ordering consequence, and the pre-`createDetached` failure path (`:643`) returns before `prompt` exists, which is fine because nothing between them reads it.

**Staleness is fetch-relative (`src/plan-gate.ts:574-583`, `1578-1601`).** `createDetached` runs `git fetch origin -- <branch>` with `cwd: repoPath` (`src/worktree.ts:634-636`) and `defaultAnchorStaleness` measures in `session.repoPath`, which shares the ref store — so the post-fetch `origin/<base>` is the one measured. Since the anchor is an ancestor of `origin/<base>` on the normal path, `git diff <sha>..origin/<base>` is exactly "what moved on base since the anchor". Zeroing on failure and the 50-path cap + `more` are exercised for real at `test/plan-gate-anchor.test.ts:88-130`.

**Contract/caller agreement.** `PlanGateServiceDeps.baseSha` changed shape (`src/plan-gate.ts:429`); grepped every consumer — the only production call site is `src/plan-gate.ts:604`, and all four test stubs were updated (`test/plan-gate-service.test.ts:85`, `:2090`, `:2126`, `:2142`). No `defaultBaseSha` reference survives anywhere (the `src/critic-core.ts:623` comment was renamed). `session.worktreePath` is a non-optional `string` (`src/types.ts:29`), and a missing/removed worktree degrades through the `try/catch` in the `git` helper to `anchored: false` rather than throwing (asserted at `test/plan-gate-anchor.test.ts:83-86`).

**No i18n / catalog impact.** Nothing under `ui/` is touched and the new prompt text is reviewer-facing English, consistent with the rest of `planReviewPrompt`; the `en`-parity and "no `./config` import" guards still hold — `src/plan-gate.ts` gained only `node:crypto` (`src/plan-gate.ts:8`), so `test/operator-language.test.ts:305-307` is unaffected.

## Stated limitation

- I could not execute `git` in this environment (Bash denied), so I did not empirically confirm that `--end-of-options` is accepted by `merge-base` / `rev-list` / `diff` on the installed git — it is only used with `rev-parse` elsewhere in the tree (`src/critic-core.ts:628`, `src/plan-gate.ts:1548`). It is parse-options/revision-machinery-wide since git 2.24, and `test/plan-gate-anchor.test.ts` drives all three against a real repo, so CI covers it; flagging only that I verified it by reasoning + test coverage, not by running the commands.

- `src/transient-agent-argv.ts:91-96,170` adds a public `sessionId` override to the shared argv builder — necessary here (the prompt cannot be built until after `createDetached`), but it is new API surface with no shape validation, so a future caller could pin a non-UUID `--session-id`; the sole caller passes `randomUUID()` (`src/plan-gate.ts:624`).

scoop (commented):
## Prior-revision point

**1. Re-raise EXEMPTION bullet emitted unconditionally — ADDRESSED.** `locationReferenceBlock` now takes an `isReReview` flag (`src/plan-gate.ts:164-168`) and only pushes the EXEMPTION bullet when it is true (`src/plan-gate.ts:220-224`). The call site passes `priorFindings.length > 0` (`src/plan-gate.ts:216`) — the exact same condition that gates the "This is a RE-REVIEW … re-raise it verbatim" block at `src/plan-gate.ts:207-213`, so the bullet's "the re-review instruction above" can never dangle. Pinned by `test/plan-gate-prompt.test.ts` ("the re-raise exemption tracks the re-review block it cites, in every tier"), which also asserts the ordering `re-raise it verbatim` < `EXEMPTION` in all three tiers.

## Verification performed

**Anchor resolution.** `defaultPlanAnchorSha` (`src/plan-gate.ts:1545-1573`) runs `merge-base --end-of-options HEAD origin/<base>` in the *session worktree* and falls back to `revParse(origin/<base>) ?? revParse(base) ?? base` with `anchored: false`. Verified the security property the doc claims (`src/plan-gate.ts:1523-1532`): a merge-base is by definition an ancestor of `origin/<base>`, so the reviewer's checkout is still a commit on the base branch's own history — the invariant `src/plan-gate.ts:733-736` relies on when it says the reviewer "detaches at the TRUSTED base sha". `ahead === 0` implies `merge-base == HEAD`, which can only happen when HEAD is already reachable from `origin/<base>`; so the strong tier never exposes planner-authored `CLAUDE.md`/hooks. Failure modes all degrade rather than throw: `git()` catches (`src/plan-gate.ts:1550-1558`), `Number(null ?? "0") === 0`, and `Number.isFinite` guards the `ahead` parse (`src/plan-gate.ts:1570`).

**createDetached contract.** `sha` from merge-base is 40-char lowercase hex, which satisfies `/^[0-9a-fA-F]{7,40}$/` at `src/worktree.ts:616`; the slug is a `randomUUID()` which satisfies `/^[A-Za-z0-9_-]{1,128}$/` at `src/worktree.ts:627`. The worktree path tag is `${slug}-${sha.slice(0,8)}` (`src/worktree.ts:629`), so the per-run uniqueness argument in `src/plan-gate.ts:640-644` still holds with the new sha source.

**Reordering of prompt vs. worktree creation.** `begin()` now mints `reviewerSessionId` itself (`src/plan-gate.ts:636`), creates the worktree (`:647-652`), then composes the prompt (`:659-663`) and argv (`:664-670`). Confirmed the pre-minted id is honoured end-to-end: `reviewerArgv` forwards `sessionId` (`src/plan-gate.ts:313-331`) into `buildTransientAgentArgv`, which uses `opts.sessionId ?? randomUUID()` (`src/transient-agent-argv.ts:170`) for BOTH the Claude `--session-id` (`:214-215`) and the Codex per-spawn last-message filename `codexLastMessageFile(sessionId)` (`:193`) — so transcript, verdict and worktree lookups stay keyed on one id. The `forget()` re-check (`src/plan-gate.ts:678`), the fail-closed auth check (`:686`) and `scrubStaleVerdictArtifacts` (`:736`) all still sit after `createDetached`, so no ordering invariant was disturbed. `reviewerArgv` has exactly two other callers, both in `test/plan-gate-prompt.test.ts:47,62`, which pass 3 args and are unaffected by the new trailing optional.

**Dep-shape change.** `PlanGateServiceDeps.baseSha` widened to `(repoPath, base, worktreePath) => PlanAnchorResolution` (`src/plan-gate.ts:441`, field `:504`, default `:522`). The only production instantiation, `src/index.ts:1375`, does not override `baseSha` or `anchorStaleness`, so it picks up the defaults; the only other instantiation is the test harness at `test/plan-gate-service.test.ts:98`, whose stub was updated to the new object shape. `session.worktreePath` is a non-optional `string` (`src/types.ts:29`), and a missing/removed worktree just makes `execFileSync` throw into the catch → `anchored: false`.

**Staleness ordering.** `defaultAnchorStaleness` (`src/plan-gate.ts:1590-1613`) is invoked from `composeReviewerPrompt` (`src/plan-gate.ts:595-601`), which `begin()` calls only after `createDetached` — whose `git fetch origin -- <branch>` is at `src/worktree.ts:634-636`. So the `behind`/`changedSince` numbers are measured against the freshly-fetched ref, as the doc claims. `behind <= 0` short-circuits before the `git diff`, and the cap/overflow (`CHANGED_SINCE_CAP = 50`, `src/plan-gate.ts:1477`, applied `:1616-1617`) matches what the prompt renders as "and N more" (`src/plan-gate.ts:200-203`). `anchored: false` skips staleness entirely (`src/plan-gate.ts:602-604`), and the prompt only receives it when `behind > 0` (`:600`), matching the block's own `behind > 0 && changedSince.length` guard (`:199`).

**Strip regex** (`src/plan-gate.ts:129-134`). Traced by hand against every case the new table asserts. The extension-must-start-with-a-letter rule is what keeps `16.9:1`, `1.5:30`, `v2.0:5`, `cpu 0.75:1` intact; the leading lookbehind `(?<![A-Za-z0-9_.:\-/\\])` is what stops `https://example.com:443/x` matching (every candidate start is preceded by `/` or `:`, and `https` itself has no `.ext`); the trailing `(?!/)` is what rescues `ghcr.io:443/x`, `registry.io:5000/img:tag`, `docker.io:5000/a/b`. `foo.ts:12:5` is consumed whole by `(?::\d+)*` rather than leaving a bogus `foo.ts:5`. The documented false positive (`ghcr.io:443` → `ghcr.io`) is real and is asserted as such. Applied only to `plan` (`src/plan-gate.ts:206`); `task`, `issueBody` (`:193-203`) and `priorFindings` (`:210`) are pushed unmodified — verified by reading the assembly, not just the comment.

**Directive parity.** `PLAN_REFERENCE_STYLE` (`src/service.ts:1056-1060`) is spliced into the *shared* body of both builders — `src/service.ts:1102` (interactive) and `src/service.ts:1133` (auto) — outside the two `agentProvider === "codex"` branches (`:1079-1089`, `:1122-1126`), so claude and codex receive it byte-identically. `PLAN_GATE_DIRECTIVE_INTERACTIVE`/`_AUTO` are still `…("claude")` (`src/service.ts:1108`, `:1140`) and exported at `:1141`, which is what `test/service-plan-gate.test.ts:2-7` imports. No `./config` import was added to `src/plan-gate.ts` (only `randomUUID` was added to the existing `node:crypto` import), so the import-time-freeze guard at `test/operator-language.test.ts:306-308` still holds.

**i18n / locale parity.** No `ui/` files are touched and no message catalogue keys are added; the plan-gate directives and the reviewer prompt are deliberately non-i18n'd English (`src/plan-gate.ts:155`), with only the trailing `operatorLanguage === "de"` output-language instruction (`src/plan-gate.ts:222-229`) unchanged. Nothing here needs an `en`/`de` counterpart.

**Stated limitation.** I could not execute git in this environment (Bash denied), so I did not empirically confirm that this host's git accepts `--end-of-options` for `rev-list`, `diff` and `merge-base` — it is new usage in this repo (previously only `rev-parse`, `src/critic-core.ts:628` and `src/plan-gate.ts:1560`). Git's revision parser and parse-options both gained it in 2.24, and every call site degrades silently (caught → `anchored: false` / `behind: 0`) rather than throwing, so the downside if a host predates that is a no-op, not a crash. `test/plan-gate-anchor.test.ts` exercises all three against real git, so CI is the ground truth.

- The bare-ref-name fallback rung in the anchor chain remains reachable-in-principle only for a branch name that is itself 7-40 hex chars; it predates this PR and the plan explicitly kept it.
- `resolveFindings`' fallback to `rawSummary` when `findings[]` is empty means a reviewer that routes everything to `body` but still returns `request-changes` will have its summary promoted to a finding — pre-existing behaviour, unchanged here.

PR #1864, opened 2026-07-21
chore(deps): apply supply-chain cooldown to site/ + docs-site/

### LINKED ISSUE #1860: site/ and docs-site/ don't inherit the root bunfig.toml supply-chain cooldown
The root `bunfig.toml` sets a 3-day supply-chain cooldown:

```toml
[install]
minimumReleaseAge = 259200
```

**`site/` and `docs-site/` do not inherit it.** Per Bun's documentation:

> Bun looks for `bunfig.toml` in `$XDG_CONFIG_HOME/.bunfig.toml` (or `$HOME/.bunfig.toml`) and `./bunfig.toml` when running `bun install`, `bun remove`, or `bun add`.

That is **cwd + global only — there is no ancestor-directory walk**, and these two packages cannot inherit it any other way:

- the root `package.json` has **no `workspaces` field`**
- `site/` and `docs-site/` each carry their **own `bun.lock`** — they are standalone packages
- installs run with cwd = `site/` or `docs-site/`, and neither directory has a `bunfig.toml`

So every `bun install` in either site today resolves with **no minimum-release-age protection at all** — including a human's casual `bun install`, and including the Vercel build if it ever resolves fresh.

This was concrete, not theoretical. During the Astro 6 → 7 upgrade, `astro@^7.1.1` would have resolved to `7.1.3`, published **one day earlier**, entirely unprotected. That PR worked around it by passing the flag explicitly:

```
bun install --minimum-release-age=259200
```

which is verifiable but not durable — it depends on whoever runs the next install remembering.

## Proposal

Commit `site/bunfig.toml` and `docs-site/bunfig.toml`, each containing:

```toml
[install]
minimumReleaseAge = 259200
```

Worth deciding at the same time:

- whether to mirror the root file's explanatory comment so the next reader understands why three separate copies exist
- whether `minimumReleaseAgeExcludes` is wanted (the root config does not currently set it)
- whether the root `bunfig.toml` comment should be amended — it explains the Dependabot interaction but does not mention that the two site packages fall outside its scope entirely

Deliberately deferred from the Astro 7 PR: permanently changing these packages' install policy is scope beyond a security bump and deserves its own review.

scoop:
Closes #1860.

`site/` and `docs-site/` were resolving npm versions with **no supply-chain cooldown at all**. Bun reads `bunfig.toml` from cwd + global only — there is no ancestor-directory walk — and both are standalone packages (own `bun.lock`; the root `package.json` has no `workspaces` field), so the root config never applied.

This is not a new install policy. `ui/bunfig.toml` and `extension/bunfig.toml` already exist as byte-identical copies of the same cooldown block; these two packages were simply missed. This completes the existing convention.

## Changes

| File | Change |
| --- | --- |
| `site/bunfig.toml` | added — verbatim copy of `ui/bunfig.toml` |
| `docs-site/bunfig.toml` | added — verbatim copy of `ui/bunfig.toml` |
| `.github/dependabot.yml` | `/site` + `/docs-site` added to the bun `directories:`; stale "all three packages" comment → "five" |

Dependabot omitted both packages too — same root cause. Including it also makes that block's existing claim ("matches the `minimumReleaseAge` in each bunfig.toml") true.

## Verification

**1. Byte-identity** — `cmp ui/bunfig.toml {site,docs-site}/bunfig.toml` silent for both. ✅

**2. Resolution probe** — proves Bun actually reads a `bunfig.toml` at that cwd, run on a scratch copy of `site/` outside the worktree. `bun add typescript@latest --dry-run`:

| Arm | scratch `bunfig.toml` | resolved |
| --- | --- | --- |
| A | `minimumReleaseAge = 315360000` (~10y) | `typescript@2.0.0` |
| B | file removed | `typescript@7.0.2` |
| C | verbatim committed copy (259200) | `typescript@7.0.2` |

A ≠ B, B at current latest, C resolves on the exact committed file. ✅ The *difference* is the evidence — either arm alone proves nothing.

Worth recording for the next reader: **`minimumReleaseAge` does not hard-fail.** It filters candidates and falls back to the newest version older than the cutoff, which is why arm A resolved an ancient version instead of erroring. An earlier draft of this check asserted "must be refused" and would have reported a false failure. `typescript` was chosen deliberately — its 2012-onward history guarantees versions on both sides of a 10-year cutoff, so the fallback branch engages rather than erroring; a package first published inside the last 10 years (e.g. `astro`) would have made the result depend on registry history.

**3. `dependabot.yml`** parses; bun `directories` is exactly `["/", "/ui", "/extension", "/site", "/docs-site"]`; `open-pull-requests-limit` and `groups` byte-unchanged vs main; the `github-actions` block untouched. ✅

**4. Final gate** — `git diff --exit-code -- '**/bun.lock'` clean, `git status --porcelain` shows only the three intended files. No lockfile moved. ✅

Prettier clean (`.prettierignore` excludes `site`/`docs-site` wholesale, so only `dependabot.yml` is covered). Branch-hygiene + feature-catalog gates pass. Pre-push initially failed on `TopBar.browser.test.ts` — a sub-pixel geometry assertion (`51.99998474121094` vs `52`); it passes in isolation and this branch's `ui/` is byte-identical to `origin/main`, so it is a pre-existing flake, not caused by this change. The retried push passed clean.

## Reviewer notes

- **Vercel pickup is assumed, not verified.** Neither `vercel.json` sets an `installCommand`, and Root Directory lives in the Vercel dashboard, so cwd-at-install is unverifiable from the repo. It holds iff Root Directory is `site` / `docs-site` (near-certain — separate projects in a monorepo). See the manual step below. If it is instead the repo root, Vercel builds stay unprotected; the human-run and Dependabot vectors are fixed either way.
- **Expect new Dependabot PRs** for both packages, against the shared `open-pull-requests-limit: 10`. Intended. Deliberately not mitigated by raising the limit — that is a separate PR-volume policy call.
- **`site/` bump PRs get no CI build signal.** CI builds `docs-site` but not `site`, so Vercel's preview deploy is their only check. Pre-existing CI shape, newly relevant. Not fixed here.
- **Deliberately excluded:** `minimumReleaseAgeExcludes` (no existing config sets one), a root-comment amendment (with all five packages covered there is no out-of-scope package left to document), a parity-enforcement CI gate, and the `ci.yml` comments at `:59`/`:172` — those describe CI job shape, which this change does not alter, unlike the dependabot line.

```shepherd:manual-steps
- [ ] Confirm the Vercel Root Directory for the site and docs-site projects is `site` / `docs-site` (Settings → General → Root Directory, or the install step of a recent build log) so the new bunfig.toml files are picked up on Vercel builds
```

[no-feature-entry]

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,6 @@
 version: 2
 updates:
-  # Bun version updates across all three packages. Bun *security* updates are
+  # Bun version updates across all five packages. Bun *security* updates are
   # unsupported by Dependabot and are disabled at the repo level; daily scheduling
   # keeps the version-update (= effective security-fix) channel fresh.
   #
@@ -16,6 +16,8 @@ updates:
       - "/"
       - "/ui"
       - "/extension"
+      - "/site"
+      - "/docs-site"
     schedule:
       interval: "daily"
     cooldown:
--- a/docs-site/bunfig.toml
+++ b/docs-site/bunfig.toml
@@ -1,0 +1,8 @@
+# Supply-chain cooldown: refuse to resolve npm versions published < 3 days ago.
+# Compromised releases are typically yanked within hours, so this filters most
+# such incidents at the resolution layer. Note: only enforced when resolving new
+# versions (`bun add`/`bun update`); versions already pinned in bun.lock are
+# reinstalled without the check, and Dependabot doesn't read this — its own
+# `cooldown` in .github/dependabot.yml guards that vector. 259200s = 3 days.
+[install]
+minimumReleaseAge = 259200
--- a/site/bunfig.toml
+++ b/site/bunfig.toml
@@ -1,0 +1,8 @@
+# Supply-chain cooldown: refuse to resolve npm versions published < 3 days ago.
+# Compromised releases are typically yanked within hours, so this filters most
+# such incidents at the resolution layer. Note: only enforced when resolving new
+# versions (`bun add`/`bun update`); versions already pinned in bun.lock are
+# reinstalled without the check, and Dependabot doesn't read this — its own
+# `cooldown` in .github/dependabot.yml guards that vector. 259200s = 3 days.
+[install]
+minimumReleaseAge = 259200


scoop (commented):
## Verdict

The change does exactly what the issue asked and nothing more: three files, no lockfile movement, no source changes.

## What I verified

**1. New bunfig files are byte-identical to the established convention.** `cmp ui/bunfig.toml site/bunfig.toml` and `cmp ui/bunfig.toml docs-site/bunfig.toml` are both silent — the two new files are exact copies of `ui/bunfig.toml:1-8` (which itself is identical to `extension/bunfig.toml:1-8`). The cooldown value `minimumReleaseAge = 259200` matches the root `bunfig.toml:14` value. The mirrored comment is still truthful in its new location: it references `.github/dependabot.yml`'s own `cooldown`, which exists at `.github/dependabot.yml:23-25` and now covers both new directories.

**2. Bun's config-discovery premise holds.** Confirmed the packages really are standalone: `bun.lock` exists at `site/bun.lock` and `docs-site/bun.lock` (alongside `./bun.lock`, `ui/bun.lock`, `extension/bun.lock`), and the root `package.json` has no `workspaces` key — so nothing was inheriting the root cooldown. Neither directory previously had a `bunfig.toml` (both files are `new file mode` in the diff), and neither `.gitignore` excludes `bunfig.toml` (grep for `bunfig` across gitignores returned nothing), so the files will actually ship.

**3. The "five packages" comment edit is factually correct.** `find -name package.json -maxdepth 2` (excluding node_modules) returns exactly five: `./package.json`, `./ui`, `./extension`, `./site`, `./docs-site` — matching the five-entry `directories:` list at `.github/dependabot.yml:15-20`. `open-pull-requests-limit: 10` (`:26`) and the `groups` block (`:30-36`) are untouched by the diff, as the plan required, and the added entries landed in the `bun` ecosystem block (`:14`), not the `github-actions` block (`:38`). No comment in that block still says "three".

**4. No CI regression risk from the new files.** `.github/actions/bun-install-retry/action.yml:26` runs `bun install --frozen-lockfile`, and `.github/workflows/ci.yml:218-221` is the only CI install with cwd = `docs-site`. A frozen-lockfile install reinstalls already-pinned versions without a resolution pass, so the new `minimumReleaseAge` cannot fail or alter that build. `site/` is not installed in CI at all.

**5. Diff is scoped.** `git diff --stat` shows only `.github/dependabot.yml` (+4/-1... net 3 lines), `site/bunfig.toml`, `docs-site/bunfig.toml`. No lockfile touched, no `ci.yml` comment touched, no `minimumReleaseAgeExcludes` added, no root/ui/extension bunfig changed — all consistent with the plan's Out of Scope list.

## Stated limitations (not findings)

- **Vercel pickup is unverifiable from this repo.** `site/vercel.json` and `docs-site/vercel.json` contain only `redirects` / `framework` — no `installCommand` — and each project's Root Directory lives in the Vercel dashboard. Whether Vercel's install runs with cwd = `site` / `docs-site` (and therefore reads the new file) cannot be confirmed here; assumed as written. Worst case the human and Dependabot vectors are fixed and the Vercel vector is unchanged — still a strict improvement.
- **Bun's `bunfig.toml` discovery semantics** (cwd + global, no ancestor walk) are third-party behavior documented outside this repo; I did not run a resolver probe (read-only review). The `ui/` and `extension/` precedent is consistent with it.

- `.github/workflows/ci.yml:172` calls docs-site "a 4th in-monorepo package"; with `site/` also carrying its own lockfile (pre-existing, not introduced here) that ordinal was already loose. Not falsified by this diff and correctly left alone.
- `site/` has no CI build job, so Dependabot bump PRs for it will have no CI signal beyond a Vercel preview deploy. Pre-existing CI shape, newly more visible now that `/site` is in the Dependabot list.

PR #1867, opened 2026-07-21
ci: auto-shorten over-long Dependabot PR titles

scoop:
## Problem

Dependabot has two title formats. When a group carries **>1 update** it uses the short form (~92 chars, fits). When a group carries **exactly one** update it uses a long form that appends a redundant suffix to a subject which *already* names the package and the directory:

```
chore(deps-dev): bump typedoc from 0.28.19 to 0.28.20 in /docs-site in the development-dependencies group across 1 directory
                                                                    └──────────────── redundant, 57 chars ────────────────┘
```

That is 124 chars, and `pr-title.yml` fails it on commitlint's `header-max-length` (100). #1866 hit this and sat `BLOCKED` until I retitled it by hand. `@dependabot recreate` regenerates the identical title, so there is no bot-side escape — every future lone-package group bump wedges the same way.

## Fix

Strip that one suffix in place, then lint the result.

Two non-obvious constraints shaped the implementation, both verified against GitHub docs rather than assumed:

1. **It must run in the same step as the lint, not a separate workflow.** An edit made with `GITHUB_TOKEN` [does not trigger a new workflow run](https://docs.github.com/en/actions/how-tos/write-workflows/choose-when-workflows-run/trigger-a-workflow). A standalone retitle workflow would fix the title but leave *this* required check pinned to its stale red result — strictly worse than today, because it would look self-healing while still blocking.

2. **It does *not* need `pull_request_target`.** My first instinct was that Dependabot's read-only token forced it. It doesn't: since [Oct 2021](https://github.blog/changelog/2021-10-06-github-actions-workflows-triggered-by-dependabot-prs-will-respect-permissions-key-in-workflows/) Dependabot-triggered `pull_request` runs respect the `permissions` key, so a plain `pull-requests: write` suffices. This matters — `pull_request_target` would hand a write token to a workflow evaluating an untrusted head ref, and it is avoided entirely here. On a `pull_request` trigger GitHub forces the token read-only for fork PRs regardless of the key, so the write scope only ever materialises for same-repo branches, which is what Dependabot's branches are.

## Blast radius is deliberately narrow

The rewrite only fires when **all** of: author is `dependabot[bot]`, title is >100 chars, the title matches the exact suffix pattern, and the result lands ≤100. Anything else falls through to commitlint unchanged.

| Title | Author | Result |
|---|---|---|
| typedoc long form (124) | `dependabot[bot]` | retitled → 67, lint passes |
| same title (124) | a human | **untouched**, lint fails as today |
| grouped form (92) | `dependabot[bot]` | untouched, already conventional |
| long but unrecognised form (120) | `dependabot[bot]` | untouched + `::warning::`, lint fails for a human to resolve |

A human's over-long title stays a hard failure they resolve themselves — this does not quietly relax the rule for everyone.

## Why not just raise `header-max-length` for `chore(deps*)`

That was the alternative. It punches a permanent hole in a limit that exists for changelog readability, and it would apply to every dependency PR forever. This removes 57 characters of pure redundancy instead, so the surviving subject is both under the limit and strictly more readable.

## Verification

- `commitlint` run locally against both forms: the 124-char title fails with `header-max-length`, the shortened 67-char title passes clean.
- The shortening logic simulated across all six title shapes in the table above plus the historical titles of the last 15 merged Dependabot PRs — only the intended form is modified.
- Injection-checked: the title reaches the shell exclusively via `env:`, never string-interpolated. A title containing `$(touch /tmp/PWNED)`, backticks and `; rm -rf` executes nothing.
- Workflow YAML parse-validated.

Note this PR cannot exercise the new path itself — it is not Dependabot-authored, so the added branch is skipped and only the pre-existing lint runs. First real proof arrives on the next lone-package group bump.

### DIFF
--- a/.github/workflows/pr-title.yml
+++ b/.github/workflows/pr-title.yml
@@ -15,13 +15,6 @@ on:
 
 permissions:
   contents: read
-  # Needed only by the Dependabot auto-shorten step below (`gh pr edit --title`).
-  # Safe on a `pull_request` trigger: GitHub forces the token read-only for fork
-  # PRs regardless of this key, so the write scope only ever materialises for
-  # same-repo branches — which is exactly what Dependabot's branches are. This is
-  # deliberately NOT `pull_request_target`, which would hand a write token to a
-  # workflow evaluating an untrusted head ref.
-  pull-requests: write
 
 jobs:
   pr-title:
@@ -31,6 +24,19 @@ jobs:
     # pr-hygiene.yml, which guard off the same head_ref prefix AND the release-bot
     # identity (branch-name-only was a fork-spoofable required-check skip).
     if: ${{ !(startsWith(github.head_ref, 'release-please--') && (github.event.pull_request.user.login == 'shepherd-release-please[bot]' || github.event.pull_request.user.login == 'github-actions[bot]')) }}
+    # Scoped to this job, not the workflow, so the write grant can't silently
+    # extend to a job added here later. Needed only by the Dependabot
+    # auto-shorten step below (`gh pr edit --title`).
+    #
+    # This is deliberately NOT `pull_request_target`, which would hand a write
+    # token to a workflow evaluating an untrusted head ref. On `pull_request`,
+    # fork PRs normally get a read-only token regardless of this key — but that
+    # is a default, not an invariant: the "Send write tokens to workflows from
+    # fork pull requests" repo/org setting can override it. So the rewrite below
+    # is gated on the Dependabot author rather than relying on that behaviour.
+    permissions:
+      contents: read
+      pull-requests: write
     # Pinned to GitHub-hosted — public repo, no CI_RUNNER toggle (see ci.yml).
     runs-on: ubuntu-latest
     steps:
@@ -42,12 +48,12 @@ jobs:
           bun-version: latest
       - name: Install (root)
         uses: ./.github/actions/bun-install-retry
-      # Dependabot's own title format can breach commitlint's 100-char header limit
+      # Dependabot's own title format can breach commitlint's header-length limit
       # and it regenerates the same title on `@dependabot recreate`, so the PR is
       # wedged until a human retitles it by hand (that's what happened to #1866, at
-      # 124 chars). When a group carries exactly one update, Dependabot appends a
-      # redundant " in the <group> group across 1 directory" to a subject that
-      # already names the package and directory — dropping it is lossless.
+      # 124 chars). Dependabot appends a redundant " in the <group> group" — with
+      # an optional " across N director(y|ies)" — to a subject that already names
+      # the package and directory, so dropping it is lossless.
       #
       # This runs in the SAME step as the lint, not a separate workflow: an edit
       # made with GITHUB_TOKEN does not trigger a new workflow run, so a retitle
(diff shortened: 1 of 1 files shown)


scoop (commented):
## What I verified

**The mechanism works and the reconstructed motivating case checks out.** The merged body of the PR this change cites (`git log -1 4aea6a263`) reads "Bumps the development-dependencies group with 1 update in the /docs-site directory" — i.e. one update, one directory. Dependabot's title for that shape is `chore(deps-dev): bump typedoc from 0.28.19 to 0.28.20 in /docs-site in the development-dependencies group across 1 directory`, which is exactly **124 characters**, matching the claim in `.github/workflows/pr-title.yml:47`. The new `sed -E 's/ in the [^ ]+ group across [0-9]+ director(y|ies)$//'` strips the trailing 57 chars, leaving a 67-char title that still names package, both versions and the directory — genuinely lossless for this case, as the comment claims.

**Group names are regex-safe.** `[^ ]+` requires a space-free group name; the only groups configured are `production-dependencies` and `development-dependencies` (`.github/dependabot.yml:30-34`) and `actions` (`.github/dependabot.yml:41-43`). None contain spaces, so the pattern matches.

**`gh` will resolve the repo.** `gh pr edit` relies on git-remote auto-detection; the job checks out first (`.github/workflows/pr-title.yml:35-36`), which is the same reasoning documented in `.github/workflows/doc-automerge.yml:53-56` ("gh's repo auto-detection falls back to the local git remote… without a checkout it dies").

**No retitle loop, and no stale-check hazard.** An edit made with `GITHUB_TOKEN` does not spawn a new workflow run, so the `edited` type in `.github/workflows/pr-title.yml:13` is not re-entered by this step; keeping the rewrite in the *same* step as the lint (rather than a separate workflow) is the right call for exactly that reason, and `@dependabot recreate` regenerates the title as *dependabot*, which does re-trigger. Correct as written.

**Injection surface.** `PR_TITLE`/`PR_AUTHOR` stay in env and are never interpolated into the shell; `$SHORT` reaches `gh` as a distinct argv element after `--title`, so a title beginning with `-` is still consumed as the flag's value. No new injection path.

**Character counting.** `${#TITLE}` is bash character count under the runner's UTF-8 locale, while commitlint measures JS string length (UTF-16 code units). These disagree for astral-plane characters — irrelevant for Dependabot's ASCII titles, noted only for completeness.

### Stated limitations (could not verify against ground truth in this repo)
- Whether a Dependabot-triggered `pull_request` run can be elevated to `pull-requests: write` via the `permissions` key is GitHub platform behaviour, not verifiable from this tree. It matches the documented `dependabot/fetch-metadata` auto-merge recipe, so I assumed it as written; if it were wrong, `gh pr edit` would 403 and (under `set -e`) fail the step — the PR would stay red rather than silently mis-title, so the failure mode is safe either way.
- I could not open `@commitlint/config-conventional` to confirm `header-max-length` is 100 — `node_modules` is not installed in this worktree. `commitlint.config.js:1-3` only shows `extends: ["@commitlint/config-conventional"]`; 100 is that preset's documented default.
- The exact title Dependabot emits for a single-dependency group spanning *multiple* directories (which the regex also strips, via `director(ies)`) is not reproducible from this repo's history — every grouped multi-directory subject here is the multi-update form. I did not treat the broader-than-documented regex as a defect on that basis.

## Findings
All three are nits; nothing blocking. See the `findings` array.

- Squash-merge appends ` (#NNNN)` to the title, so a title passing at exactly 100 chars lands a ~108-char subject on `main` that the commit-msg hook would have rejected — a pre-existing gap in the title-gates-subject premise, unchanged by this PR.

scoop (commented):
## Re-review — all three prior points are addressed

Diff is a single file: `.github/workflows/pr-title.yml` (+63/-1), head `46a9441f4`, base `868a1963` (`git diff 868a1963...HEAD --stat`).

**1. Hardcoded `100` → derived.** Gone. The limit is now read from commitlint's own resolved config at `.github/workflows/pr-title.yml:76-78` (`bunx commitlint --print-config json` piped into `bun -e` reading `rules["header-max-length"][2]`), and every subsequent comparison uses `"$LIMIT"` (`:87`, `:92`, `:93`, `:100`) — no literal `100` remains anywhere in the file. This is the right call given `commitlint.config.js:1-3`, which is nothing but `extends: ["@commitlint/config-conventional"]` — the limit lives entirely in the preset (`package.json:32`, `@commitlint/config-conventional: ^21.2.0`), so it could move under the repo without any local edit.

**2. Suffix regex missing the single-directory grouped form.** Fixed at `.github/workflows/pr-title.yml:90-91`: `s/ in the [^ ]+ group( across [0-9]+ director(y|ies))?$//` — the ` across N director(y|ies)` portion is now an optional group, so the form cited in the prior review matches. Verified the ground-truth shape still exists in this repo's history: `git log` yields `chore(ci): bump actions/upload-artifact from 4 to 7 in the actions group (#1396)`, which the old anchor missed and the new one strips. `[^ ]+` for the group name is safe — Dependabot group names are `dependabot.yml` mapping keys (`.github/dependabot.yml:31-35`, `:45-46`: `production-dependencies`, `development-dependencies`, `actions`) and cannot contain spaces.

**3. Workflow-level `pull-requests: write`.** Fixed. Workflow-level `permissions` at `:16-17` is still `contents: read` only; the write grant is now job-scoped at `:37-39` with an explanatory comment at `:27-29`. The absolute fork-token claim is softened at `:32-36` ("that is a default, not an invariant") and correctly names the repo/org override setting, and the rewrite is gated on `PR_AUTHOR = dependabot[bot]` at `:87` rather than on token behaviour. Granting `pull-requests: write` via the `permissions` key on a Dependabot-triggered `pull_request` run is the documented, supported pattern for `gh pr edit`, so the grant will actually take effect.

### Other things checked

- Re-lint path: on the rewrite branch `TITLE="$SHORT"` (`:96`) before the final `printf | bunx commitlint` (`:105`), so the step lints the title it just wrote rather than the stale one — the check goes green in the same run. The comment at `:58-60` is right that a `GITHUB_TOKEN` edit does not spawn a new run, which is exactly why doing this in-step matters.
- Injection: title, author, and number all arrive via `env:` (`:65-68`), never interpolated into the shell; `gh pr edit "$PR_NUMBER" --title "$SHORT"` (`:95`) is fully quoted and `--title` takes its value positionally, so a leading `-` in a title cannot become a flag.
- Unhandled-but-warned form: this repo's *dominant* grouped title is `chore(deps-dev): bump the development-dependencies group across 3 directories with 6 updates` (#1853), which ends in `with N updates` and does not match the sed. That is fine — it falls into the `else` at `:97-101`, which warns and defers to commitlint instead of mangling; and at ~91 chars that form does not realistically breach the limit anyway. Not a finding.
- `${#TITLE}` counts locale characters while commitlint counts JS string length; identical for the ASCII-only Dependabot titles this branch can reach. Not a finding.

### Stated limitation

- Could not verify `commitlint --print-config json` against the installed CLI: `node_modules/` is not present in this worktree, so `@commitlint/cli@^21.2.1` (`package.json:31`) could not be opened. The invocation matches the documented CLI option, but I did not confirm it against source. Finding 1 below is partly what protects against being wrong here — with the fix, an unsupported invocation degrades to a warning instead of failing every PR-title check.

- `.github/workflows/pr-title.yml:82-86`: the `-z "$LIMIT"` warning branch is dead for the failure mode it names — under `set -euo pipefail` (`:70`) a plain assignment from a failing command substitution exits the shell, so a non-zero `bunx commitlint --print-config`/`bun -e` kills the step before the branch is evaluated; it is reachable only when both commands succeed and the rule is genuinely absent from the config. (Also raised as finding 1 since the fix is a one-liner.)
- `.github/workflows/pr-title.yml:95`: a `gh pr edit` failure (token downgraded by an org workflow-permissions policy, transient API error) aborts the step under `set -e`, so the check fails red even though the shortened title it computed would have linted clean; `gh pr edit ... || echo "::warning::..."` would degrade to the pre-PR behaviour instead.

scoop:
Closing as not planned.

The auto-shorten depends entirely on `pull-requests: write` so the workflow can call `gh pr edit`. Reviewing the grant's consequences, the deciding factor was that this job already checks out the PR's merge ref, runs `bun install`, and evaluates the PR's own `commitlint.config.js` — so the grant would hand a PR-write token to a job that executes PR-controlled code. Bounded (it can't merge or push, `contents` stays `read`), but not worth it to save a manual retitle.

Dropping the grant removes the only mechanism the PR had, so the rest of it — deriving `header-max-length` from `commitlint --print-config`, widening the suffix regex to cover the single-directory group form — has nothing left to serve. Closing rather than shipping a defanged version.

The underlying nuisance stands: Dependabot's lone-package-in-a-group title form breaches the 100-char limit (#1866 at 124) and `@dependabot recreate` regenerates it, so those PRs wedge until retitled by hand. Preferred remedies are retitling by hand as it comes up, or exempting Dependabot from the title gate outright — neither needs a token grant.

Findings from the critic review were all valid and were addressed before this decision; nothing is being closed to avoid them.

Issue #1868, opened 2026-07-21
CodeQL js/insecure-temporary-file: move the 6 production /tmp paths off world-writable locations


scoop:
Six `js/insecure-temporary-file` alerts on production code, deferred together from the CodeQL triage in the PR that guarded `js/remote-property-injection`. Deferred rather than fixed because the egress half cannot be verified by `bun run test`, and fixing 1 of 6 same-rule alerts while deferring 5 was an unjustified asymmetry.

## Alerts

| Site | Issue |
| --- | --- |
| `src/update.ts:358` (path from `:280`) | `join(tmpdir(), "shepherd-update.log")` — fixed, predictable name in world-writable `/tmp`, pre-creatable or symlinkable by another local user |
| `src/egress.ts:638-641` (`egressTmpDir` `:623`, `egressTmpRoot` `:628`) | per-session dir is UUID-keyed, but its **parent** `shepherd-egress` is predictable, so the parent is the swap target |
| `src/egress-watch.ts:24` | tails `dns.log` inside that same dir |

Severity is low: Shepherd is a single-operator local daemon, so the hostile-local-user model is weak. `src/egress.ts:196` already uses `mkdtemp` and is correctly not flagged.

## Evaluate `$XDG_RUNTIME_DIR` FIRST

Earlier analysis compared only "hardened `/tmp`" vs `~/.shepherd/` and rejected `$XDG_RUNTIME_DIR` for being "login-scoped" — treating its boot-clearing as a defect. **That is backwards.**

`applyState()` (`src/update.ts:391-420`) age-bounds only the **marker-less** branch (`DEPLOY_STALE_MS`, `:414`). The **marker** branch (`:400-407`) returns `done`/`failed` with **no age check at all** — which is survivable today only because `/tmp` is cleared at boot, so `readFileSync` throws and the state resets to `idle`. That is an *implicit* property of the current path, never a designed behaviour.

So `$XDG_RUNTIME_DIR` is `0700`, user-owned, satisfies the rule, **and preserves the boot reset `applyState()` silently depends on** — and the deploy already runs under `systemd-run --user`, so the variable is present. Choosing it means no TTL work is needed.

## If `~/.shepherd/` is chosen instead

[...]

Issue #1630, opened 2026-07-11
feat: revive stranded sessions after a herdr daemon restart (panes come back as bare shells, agents gone)


Gambastudio:
## Problem

When the herdr daemon restarts — host/VM reboot, crash, or a failed update handoff — herdr faithfully restores every pane from its session persistence, but as **bare shells** at the worktree cwd: the `claude`/`codex` processes inside are gone. To Shepherd the sessions still exist, herdr reports the panes with `agent_status: unknown`, and the operator sees terminals showing an empty prompt with **no recovery path offered**. Every live session is stranded at once.

Observed live after a daemon bounce: six sessions, all panes restored as empty shells. Recovery was entirely manual.

## What worked (manual recovery)

Typing `claude --continue` in each pane resumes the previous conversation in that worktree — full history restored, poller re-detects the agent, steering works again. Notes from doing this six times:

- `claude --resume <shepherd-session-id>` does **not** work — the Shepherd session id is not the Claude conversation id ("No conversation found"). `--continue` (latest conversation for that cwd) is the reliable key.
- The revived agent runs **without** the original spawn flags (system-prompt directives, settings), which matters if the session should still pass through the plan gate.

## Proposal

Detect the stranded state — an active session whose pane has **no agent process** (`agent_status: unknown` on a live, non-archived session) — and offer **revival**:

- **v1:** a per-session CTA ("agent died — revive") that injects the resume into the existing pane, using the right continuation key and ideally the original spawn argv/system-prompt so gates keep working.
- **v2:** optional auto-revive after a detected daemon restart (the reconcile pass already re-pairs sessions to panes; this is the missing second half).

This recurs for *every* self-hosted operator on *every* host reboot, so it's less an edge case than a lifecycle gap: reconcile currently restores the *pairing* but not the *agent*.

## Relation

Complements the shepherd-managed daemon lifecycle work (#1558 follow-ups) and the per-session "why is this stuck + act here" surfacing (#1561) — a stranded pane is exactly the kind of state that should carry its own one-click fix.

Gambastudio:
Second occurrence, five days after filing — this time a plain **host/VM reboot** (the fully supported lifecycle, no update involved, daemon provisioned cleanly on boot): all sessions came back as bare-shell panes (`agent_status: unknown`), stranded until manually revived.

[...]

PR #1869, opened 2026-07-21
fix(ui): guard computed-property writes in the reactive store maps

scoop:
Answers "codeql complains a lot about remote property injections, is this a false positive?" — and acts on the whole 55-alert backlog.

## The short answer

**At the creation site, yes — all 27 are false positives.** But "dismiss" was the wrong action, so this PR guards them instead.

The distinction that carries it, verified empirically in-repo:

| expression | own props | prototype changed? |
| --- | --- | --- |
| `{ ...m, ["__proto__"]: x }` — **literal**, all 22 flagged writes | `["__proto__"]` | **no** |
| `b["__proto__"] = x` — **assignment**, zero flagged sites | `[]` | **yes** |

A computed key in an object literal is defined via `CreateDataPropertyOrThrow` → `DefineOwnProperty`, so it never invokes the `__proto__` setter. The other 5 flagged sites are `delete`, which cannot pollute under any key.

## Why guard rather than dismiss

`setKey` (`store.svelte.ts`) was added by **#1630** specifically to guard this pattern — and was called at **1 of 27 sites**. Declaring the identical pattern safe everywhere else while the repo ships a guard for it was a contradiction. Guarding also closes the alerts instead of dismissing them, so nothing rests on "no downstream `[[Set]]` copy of these maps ever appears" (true today — audited — but not enforced by anything).

## The interesting part: CodeQL flagged 27 non-sinks and missed the real ones

Three genuinely `[[Set]]`-reachable paths, **none of them flagged**, all fixed here:

1. **`patchSession`** — `Object.assign(s, patch)` where `patch` spreads a JSON-parsed `session:status` WS payload. `JSON.parse` *does* create an own `__proto__`, and `Object.assign` uses `[[Set]]`. Live remote-data path, object-valued. Guarded at the chokepoint so all 10 callers are covered.
2. **`setClaudeAlive`** — `folded[id] = "stranded"` is a plain assignment with a remote key from `GET /api/stranded`. Inert *only* because `LivenessState` is a string union (the setter ignores non-object values); widen the type and it becomes live.
3. **`ReviewsStore`/`PlanGateStore.setActivity`** — found while testing: the **read** `this.activity[id]` resolves `"__proto__"` through the prototype chain to `Object.prototype`, which `pushActivity` then spreads → `TypeError`. Guarded before the read, not just the write.

## Guards are chosen by key SHAPE, never by convenience

| Guard | Applied to | Why |
| --- | --- | --- |
| strict charset `/^[0-9a-zA-Z-]+$/` | 20 session-id maps + `:196` | Empirically accepted by CodeQL as a barrier — `store.svelte.ts:597` already used `setKey` and was never flagged |
| `__proto__`/`constructor`/`prototype` deny-list | `drain`, `autoMerge` (`repoPath` keys) | The charset **rejects a real filesystem path** — routing these through it would silently no-op and freeze both maps with no error |
| `safeMerge` (strips unsafe own keys) | `patchSession` | A key predicate doesn't apply to a payload merge |

`store.svelte.ts:263` (`setBuildQueue`) is guarded **despite being unflagged**: same map, same key shape, same form as the flagged `queue:update` twin. Leaving it would have been a bypass sitting beside the guard.

## Deliberately NOT guarded

Recorded in a code comment at the guard definitions, so it isn't read as an oversight:

- **The 5 `delete` sites.** A delete can't pollute, but a *rejected* key would silently fail to delete — pinning a session as "Reviewing…" or leaving a stale verdict. Guarding there is a net loss.
- **~90 `AutomationStore` `repoPath` writes** and **`epics`' composite key** — outside the alert set; they'd need the path guard, not the charset.
- **Bulk-replace seeds** — wholesale assignment / `Object.fromEntries`, not computed writes.

## Accepted behaviour change

A stranded session whose id fails the charset no longer gets its `stranded` upgrade and renders as a plain `husk` (losing the "agent died — revive" banner). Unreachable in production — ids are server-minted `randomUUID()`s — and graceful: the session still renders, nothing is corrupted.

**This is not theoretical.** `EpicDraftPanel.browser.test.ts` built a fixture id from a describe label containing a space and comma, and the guard silently dropped the upsert until the fixture was slugified. That is exactly the residual risk, caught by the suite — worth weighing when reviewing the charset choice.

## Rest of the backlog (28 alerts, dismissed separately after this merges)

11 `insecure-temporary-file` in tests → `used_in_tests` · 6 production `/tmp` paths → deferred to #1868 · `src/activity.ts:160` reads `~/.claude/projects/…`, not a temp file · 6 `file-system-race` on the operator's own local files (`src/repos.ts:157` is a *deliberate* lstat symlink guard) · 1 unanchored regex in a test assertion · 1 `http-to-file-access` already `JSON.stringify`-escaped.

`js/indirect-command-line-injection` will be dismissed **`wont_fix`, not `false_positive`**: the previous reasoning ("argv array, no shell") was wrong twice over — an argv array doesn't refute *argument* injection, and `deploy/provision.ts` does shell out at `:369`/`:521`/`:531`. The actual ground is the absence of a privilege boundary (`deploy/install.sh:268` hands off to an operator-run script), which does not mechanically refute the pattern.

## `setPathKey` and the CodeQL barrier model

`setPathKey` initially raised a **new** `js/remote-property-injection` alert on itself — the guard I added wasn't legible to the query. Resolved in code; **no dismissal is relied on**.

What the rule models as a sanitizing barrier is a **regex allow-list** — which is why `setKey` is not flagged, and why `store.svelte.ts:597` was already unflagged before this PR. It does **not** model an equality chain, `Set.has()`, or `Object.defineProperty`; all three were tried on CI and all three stayed flagged.

An allow-list is not usable on this path: `repoPath` is arbitrary filesystem input, so any charset narrow enough to exclude `__proto__` could also reject a legitimate path — and a rejected key silently no-ops the write and freezes the drain/auto-merge UI with no error. That is exactly the failure this guard exists to prevent, so it was not an acceptable trade.

The fix builds the result with `Object.fromEntries([...Object.entries(rec), [key, value]])` instead of `{ ...rec, [key]: value }`. Semantically identical — both `CreateDataProperty`, same O(n) copy, same key ordering on overwrite (verified) — but the key is data rather than a dynamic property write. The explicit `__proto__`/`constructor`/`prototype` check remains the real barrier, and the scan now reports the alert **fixed**.

*Bookkeeping note:* a dismissal request was filed for that alert while the code fix was still in flight, and this repo has delegated dismissal enabled so it went through. It is now redundant — the alert is fixed in code, not waved through. Called out so an audit of dismissed alerts isn't misread.

**Useful for the follow-up work:** for this rule, only regex allow-lists count as barriers. Worth knowing before anyone tries to close alerts elsewhere with a deny-list.

## Notes for review

- **New file `ui/src/lib/safe-keys.ts`** rather than exporting from `store.svelte.ts`: that module already imports all four sibling stores, so exporting the helpers there and importing them back would create four import cycles.
- **Scope bound:** 23 write sites touched (22 flagged literals + `:263`) out of 119 computed writes in these stores. Worth confirming the diff touches no more.
- Considered and not adopted: validating ids once at the WS decode seam. One seam to test and it would cover the unflagged sites for free, but CodeQL is unlikely to follow a barrier across the dispatch boundary, so the alerts would still need dismissing.

## Verification

`bun run lint` + `bun run test` (7400 pass) at the root; `bun run check` (0 errors) + `bun run test` (4181 pass) in `ui/`. Branch-hygiene, feature-catalog, i18n and glossary gates all pass. Confirmed against a clean baseline worktree that the one browser-test failure was caused by this change, not pre-existing.

[no-feature-entry] — no user-facing UX ships here.

### DIFF
--- a/ui/src/lib/buildQueues.svelte.ts
+++ b/ui/src/lib/buildQueues.svelte.ts
@@ -1,4 +1,5 @@
 import type { BuildQueue } from "./types";
+import { setKey } from "./safe-keys";
 
 /** Module-level reactive build-queue map keyed by sessionId.
  *  Driven by two sources:
@@ -10,7 +11,7 @@ class BuildQueuesStore {
   map = $state<Record<string, BuildQueue>>({});
 
   upsert(q: BuildQueue) {
-    this.map = { ...this.map, [q.sessionId]: q };
+    this.map = setKey(this.map, q.sessionId, q);
   }
 
   seed(queues: Record<string, BuildQueue>) {
--- a/ui/src/lib/components/EpicDraftPanel.browser.test.ts
+++ b/ui/src/lib/components/EpicDraftPanel.browser.test.ts
@@ -77,7 +77,9 @@ describe.each([
   ["tiny", 240, "1.3"],
 ])("EpicDraftPanel bar — CTA horizontal containment (%s)", (label, width, uiScale) => {
   it("keeps the CTA inside the bar", async () => {
-    const sessionId = `epic-draft-bar-fit-${label}`;
+    // Slugified: real session ids are `randomUUID()`s, and `epicDrafts.upsert` now validates the
+    // key against that charset (safe-keys.ts), so a raw label with spaces/commas would be dropped.
+    const sessionId = `epic-draft-bar-fit-${label.replace(/[^0-9a-zA-Z-]+/g, "-")}`;
     epicDrafts.upsert(longDraft(sessionId));
 
     const { container, unmount } = await render(EpicDraftPanel, {
--- a/ui/src/lib/epic-draft.svelte.test.ts
+++ b/ui/src/lib/epic-draft.svelte.test.ts
@@ -82,3 +82,15 @@ describe("EpicDraftsStore.refresh", () => {
     expect(epicDrafts.get("s1")?.status).toBe("materializing");
   });
 });
+
+describe("prototype-pollution guard", () => {
+  it("rejects a __proto__ sessionId but still upserts a real one", () => {
+    epicDrafts.upsert(draft("__proto__", "draft"));
+    expect(Object.hasOwn(epicDrafts.map, "__proto__")).toBe(false);
+    expect(Object.getPrototypeOf({})).toBe(Object.prototype);
+
+    const d = draft("sess-1", "draft");
+    epicDrafts.upsert(d);
+    expect(epicDrafts.map["sess-1"]).toBe(d);
+  });
+});
--- a/ui/src/lib/epic-draft.svelte.ts
+++ b/ui/src/lib/epic-draft.svelte.ts
@@ -1,5 +1,6 @@
 import type { EpicDraft } from "./types";
 import { getEpicDraft } from "./api";
+import { setKey } from "./safe-keys";
 
 /** Module-level reactive epic-draft map keyed by sessionId (issue #1507).
  *  Driven by two sources:
@@ -27,7 +28,7 @@ class EpicDraftsStore {
   }
 
   upsert(d: EpicDraft) {
-    this.map = { ...this.map, [d.sessionId]: d };
+    this.map = setKey(this.map, d.sessionId, d);
     for (const fn of this.listeners) fn(d);
   }
 
--- a/ui/src/lib/recaps.svelte.test.ts
+++ b/ui/src/lib/recaps.svelte.test.ts
@@ -76,3 +76,12 @@ test("drop() is a no-op for unknown id", () => {
   // map object is the same reference (no copy was made)
   expect(recaps.map).toBe(before);
 });
+
+test("recap upsert rejects a __proto__ id but still stores a real one", () => {
+  recaps.apply({ id: "__proto__", recap: recap("__proto__") });
+  expect(Object.hasOwn(recaps.map, "__proto__")).toBe(false);
+  expect(Object.getPrototypeOf({})).toBe(Object.prototype);
+  const r = recap("sess-1");
+  recaps.apply({ id: "sess-1", recap: r });
+  expect(recaps.map["sess-1"]).toBe(r);
+});
--- a/ui/src/lib/recaps.svelte.ts
+++ b/ui/src/lib/recaps.svelte.ts
@@ -1,5 +1,6 @@
 import type { Recap } from "./types";
 import { getRecaps } from "./api";
+import { setKey } from "./safe-keys";
 
 /** Client cache of session recaps keyed by session id. Loaded once on app start;
  *  live updates arrive via the `session:recap` WS event (see store.svelte.ts). */
@@ -15,7 +16,7 @@ class RecapsStore {
   }
 
   apply(d: { id: string; recap: Recap | null }) {
-    if (d.recap) this.map = { ...this.map, [d.id]: d.recap };
+    if (d.recap) this.map = setKey(this.map, d.id, d.recap);
     else {
       const copy = { ...this.map };
       delete copy[d.id];
--- a/ui/src/lib/reviews.svelte.test.ts
+++ b/ui/src/lib/reviews.svelte.test.ts
@@ -894,3 +894,17 @@ test("repoPath-keyed automation state still round-trips (charset would reject a 
   await repoConfig.toggleAutopilot(repoPath);
   expect(repoConfig.isAutopilotEnabled(repoPath)).toBe(true);
 });
+
+test("setActivity survives a hostile inherited-property id (constructor/toString)", () => {
+  // Regression: SAFE_ID admits letter-only names, and `this.activity["constructor"]` reads the
+  // Object function off the prototype chain — pushActivity used to spread it and throw a TypeError
+  // inside WS dispatch. Unreachable with randomUUID ids, but must not crash.
+  for (const evil of ["constructor", "toString", "valueOf", "__proto__"]) {
+    expect(() => reviews.setActivity(evil, "x")).not.toThrow();
+    expect(() => planGates.setActivity(evil, "x")).not.toThrow();
+  }
+  expect(Object.getPrototypeOf({})).toBe(Object.prototype);
+  // a real id still accumulates normally after the hostile traffic
+  reviews.setActivity("sess-ok", "line one");
+  expect(reviews.activityFeed("sess-ok")).toEqual(["line one"]);
+});
--- a/ui/src/lib/reviews.svelte.test.ts
+++ b/ui/src/lib/reviews.svelte.test.ts
@@ -885,7 +885,7 @@ test("reviewing/env/activity maps reject __proto__ without dropping real ids", (
 });
 
 test("repoPath-keyed automation state still round-trips (charset would reject a path)", async () => {
-  // Regression guard for the highest-risk failure mode of this change: AutomationStore is keyed
+  // Regression guard for the highest-risk failure mode of this change: RepoConfigStore is keyed
   // by filesystem path, which the session-id charset REJECTS. Routing these writes through the
   // session-id guard would silently no-op every toggle and freeze the automation UI with no error.
   const repoPath = "/home/u/Work/my-repo.git";
--- a/ui/src/lib/reviews.svelte.test.ts
+++ b/ui/src/lib/reviews.svelte.test.ts
@@ -851,3 +851,46 @@ test("applyHandsOffDefaults reverts every optimistic field AND rethrows on a fai
   expect(repoConfig.isAutoMergeEnabled("/repo")).toBe(false);
   expect(repoConfig.isAutoAddressEnabled("/repo")).toBe(false);
 });
+
+// ── prototype-pollution guards (CodeQL js/remote-property-injection) ────────
+
+test("session-id maps reject a __proto__ key but still store real ids", () => {
+  reviews.apply({ id: "__proto__", review: verdict("__proto__") });
+  expect(Object.hasOwn(reviews.map, "__proto__")).toBe(false);
+  expect(Object.getPrototypeOf({})).toBe(Object.prototype);
+
+  const v = verdict("sess-1");
+  reviews.apply({ id: "sess-1", review: v });
+  expect(reviews.map["sess-1"]).toBe(v);
+
+  planGates.apply("__proto__", planGate("__proto__"));
+  expect(Object.hasOwn(planGates.map, "__proto__")).toBe(false);
+  const g = planGate("sess-2");
+  planGates.apply("sess-2", g);
+  expect(planGates.map["sess-2"]).toBe(g);
+});
+
+test("reviewing/env/activity maps reject __proto__ without dropping real ids", () => {
+  reviews.setReviewing("__proto__", true, { provider: "claude", model: "opus", effort: "high" });
+  expect(Object.hasOwn(reviews.reviewing, "__proto__")).toBe(false);
+  expect(Object.hasOwn(reviews.reviewerEnv, "__proto__")).toBe(false);
+  reviews.setActivity("__proto__", "reading files");
+  expect(Object.hasOwn(reviews.activity, "__proto__")).toBe(false);
+
+  reviews.setReviewing("sess-3", true, { provider: "claude", model: "opus", effort: "high" });
+  reviews.setActivity("sess-3", "reading files");
+  expect(reviews.isReviewing("sess-3")).toBe(true);
+  expect(reviews.reviewerEnvFor("sess-3")?.model).toBe("opus");
+  expect(reviews.activity["sess-3"]).toEqual(["reading files"]);
+});
+
+test("repoPath-keyed automation state still round-trips (charset would reject a path)", async () => {
+  // Regression guard for the highest-risk failure mode of this change: AutomationStore is keyed
+  // by filesystem path, which the session-id charset REJECTS. Routing these writes through the
+  // session-id guard would silently no-op every toggle and freeze the automation UI with no error.
+  const repoPath = "/home/u/Work/my-repo.git";
+  vi.mocked(putRepoConfig).mockResolvedValue(rc({ autopilotEnabled: true }));
+  repoConfig.autopilot = { [repoPath]: false };
+  await repoConfig.toggleAutopilot(repoPath);
+  expect(repoConfig.isAutopilotEnabled(repoPath)).toBe(true);
+});
--- a/ui/src/lib/reviews.svelte.ts
+++ b/ui/src/lib/reviews.svelte.ts
@@ -22,7 +22,12 @@ export const MAX_ACTIVITY_LINES = 2;
  *  re-emits the same summary every tick → no churn). Returns the SAME array reference when
  *  nothing changed so callers can skip a reactive write. */
 function pushActivity(cur: string[] | undefined, summary: string): string[] {
-  const feed = cur ?? [];
+  // `cur` comes from a bracket read of a `Record` (`this.activity[id]`), so a hostile id that names
+  // an inherited `Object.prototype` member — `constructor`, `toString`, `valueOf`, … — resolves up
+  // the chain to a non-array (a function / the prototype object) and the spread below would throw.
+  // `SAFE_ID` on the caller rejects `__proto__` but NOT those letter-only names, so treat any
+  // non-array as "no feed yet". Unreachable with real `randomUUID()` ids; defence in depth.
+  const feed = Array.isArray(cur) ? cur : [];
   if (feed[feed.length - 1] === summary) return feed; // unchanged → signal no-op to caller
   return [...feed, summary].slice(-MAX_ACTIVITY_LINES);
 }
@@ -101,8 +106,10 @@ class ReviewsStore {
    *  only renders in-flight and every run starts empty, so a straggler tick after the end can
    *  never surface (see staleness invariant). */
   setActivity(id: string, summary: string) {
-    // Gate BEFORE the read, not just the write: `this.activity["__proto__"]` resolves through the
-    // prototype chain to `Object.prototype`, which pushActivity would then try to spread.
+    // Reject a non-session-id key at the boundary so no junk entry is created for it. This is NOT
+    // what makes the subsequent read safe — a letter-only inherited name (`constructor`, `toString`)
+    // passes SAFE_ID yet reads as a non-array; pushActivity is hardened against that. Write-safety
+    // is setKey's job. Together: unreachable-hostile-id in, no crash and no pollution out.
     if (!SAFE_ID.test(id)) return;
     const next = pushActivity(this.activity[id], summary);
     if (next === this.activity[id]) return; // unchanged reference → no churn
@@ -232,8 +239,10 @@ export class PlanGateStore {
   /** Append the in-flight plan reviewer's latest tool-use summary to its bounded rolling feed
    *  (last MAX_ACTIVITY_LINES), deduping against the last line. Mirrors ReviewsStore.setActivity. */
   setActivity(id: string, summary: string) {
-    // Gate BEFORE the read, not just the write: `this.activity["__proto__"]` resolves through the
-    // prototype chain to `Object.prototype`, which pushActivity would then try to spread.
+    // Reject a non-session-id key at the boundary so no junk entry is created for it. This is NOT
+    // what makes the subsequent read safe — a letter-only inherited name (`constructor`, `toString`)
+    // passes SAFE_ID yet reads as a non-array; pushActivity is hardened against that. Write-safety
+    // is setKey's job. Together: unreachable-hostile-id in, no crash and no pollution out.
     if (!SAFE_ID.test(id)) return;
     const next = pushActivity(this.activity[id], summary);
     if (next === this.activity[id]) return; // unchanged reference → no churn
--- a/ui/src/lib/reviews.svelte.ts
+++ b/ui/src/lib/reviews.svelte.ts
@@ -2,6 +2,7 @@ import type { ReviewVerdict, PlanGate, ReviewerEnv, RepoConfig, SandboxProfile }
 import type { AutomationFlags } from "./components/git-rail-automation";
 import { handsOffPatch } from "./components/epic-handsoff";
 import type { RepoConfigResponse } from "./api";
+import { SAFE_ID, setKey } from "./safe-keys";
 import {
   getReviews,
   getReviewingIds,
@@ -63,7 +64,7 @@ class ReviewsStore {
   }
 
   apply(d: { id: string; review: ReviewVerdict | null }) {
-    if (d.review) this.map = { ...this.map, [d.id]: d.review };
+    if (d.review) this.map = setKey(this.map, d.id, d.review);
     else {
       const copy = { ...this.map };
       delete copy[d.id];
@@ -75,7 +76,7 @@ class ReviewsStore {
 
   setReviewing(id: string, on: boolean, env?: ReviewerEnv) {
     if (on) {
-      if (env) this.reviewerEnv = { ...this.reviewerEnv, [id]: env };
+      if (env) this.reviewerEnv = setKey(this.reviewerEnv, id, env);
     } else if (id in this.reviewerEnv) {
       const copy = { ...this.reviewerEnv };
       delete copy[id];
@@ -86,7 +87,7 @@ class ReviewsStore {
     // against a missed end-clear), on END the finished run's live tail is stale. Guarded by the
     // early return above so a repeated same-state call doesn't clear.
     this.clearActivity(id);
-    if (on) this.reviewing = { ...this.reviewing, [id]: true };
+    if (on) this.reviewing = setKey(this.reviewing, id, true);
     else {
       const copy = { ...this.reviewing };
       delete copy[id];
@@ -100,9 +101,12 @@ class ReviewsStore {
    *  only renders in-flight and every run starts empty, so a straggler tick after the end can
    *  never surface (see staleness invariant). */
   setActivity(id: string, summary: string) {
+    // Gate BEFORE the read, not just the write: `this.activity["__proto__"]` resolves through the
+    // prototype chain to `Object.prototype`, which pushActivity would then try to spread.
+    if (!SAFE_ID.test(id)) return;
     const next = pushActivity(this.activity[id], summary);
     if (next === this.activity[id]) return; // unchanged reference → no churn
-    this.activity = { ...this.activity, [id]: next };
+    this.activity = setKey(this.activity, id, next);
   }
 
   private clearActivity(id: string) {
@@ -192,7 +196,7 @@ export class PlanGateStore {
   }
 
   apply(id: string, gate: PlanGate) {
-    this.map = { ...this.map, [id]: gate };
+    this.map = setKey(this.map, id, gate);
     // a landing verdict means the review is no longer in flight
     this.applyReviewing(id, false);
   }
@@ -202,7 +206,7 @@ export class PlanGateStore {
     // interleave on reload, or a re-emitted signal) doesn't transition `reviewing`, but must still
     // refresh the identity. On end (`false`) drop it so a stale env can't linger past the run.
     if (on) {
-      if (env) this.reviewerEnv = { ...this.reviewerEnv, [id]: env };
+      if (env) this.reviewerEnv = setKey(this.reviewerEnv, id, env);
     } else if (id in this.reviewerEnv) {
       const copy = { ...this.reviewerEnv };
       delete copy[id];
@@ -211,7 +215,7 @@ export class PlanGateStore {
     if (!!this.reviewing[id] === on) return;
     // Reset the feed on both ends of a transition (start = fresh empty run, end = stale tail).
     this.clearActivity(id);
-    if (on) this.reviewing = { ...this.reviewing, [id]: true };
+    if (on) this.reviewing = setKey(this.reviewing, id, true);
     else {
       const copy = { ...this.reviewing };
       delete copy[id];
@@ -228,9 +232,12 @@ export class PlanGateStore {
   /** Append the in-flight plan reviewer's latest tool-use summary to its bounded rolling feed
    *  (last MAX_ACTIVITY_LINES), deduping against the last line. Mirrors ReviewsStore.setActivity. */
   setActivity(id: string, summary: string) {
+    // Gate BEFORE the read, not just the write: `this.activity["__proto__"]` resolves through the
+    // prototype chain to `Object.prototype`, which pushActivity would then try to spread.
+    if (!SAFE_ID.test(id)) return;
     const next = pushActivity(this.activity[id], summary);
     if (next === this.activity[id]) return; // unchanged reference → no churn
-    this.activity = { ...this.activity, [id]: next };
+    this.activity = setKey(this.activity, id, next);
   }
 
   private clearActivity(id: string) {
--- a/ui/src/lib/safe-keys.ts
+++ b/ui/src/lib/safe-keys.ts
@@ -66,23 +66,17 @@ export function setKey<T>(rec: Record<string, T>, id: string, value: T): Record<
  *  malformed ids — but applying it to a path key would silently no-op the write and freeze that
  *  map in the UI with no error, so the two are not interchangeable.
  *
- *  The three names are compared INLINE rather than via a `Set`, and the write goes through
- *  `Object.defineProperty`. Both are deliberate: a `Set.has()` membership test is not modelled as a
- *  sanitizing barrier by `js/remote-property-injection`, and `defineProperty` performs
- *  DefineOwnProperty — the same semantics as the object-literal computed key it replaces, but
- *  without a dynamic-key write for the query to flag. Behaviour is unchanged either way; an
- *  allow-list is NOT an option here because a legitimate repo path can be `""`, `"~"` or any
- *  absolute path, and rejecting one would freeze the map. */
+ *  CodeQL flags this function and there is a standing dismissal for it. `js/remote-property-injection`
+ *  models a regex allow-list as a sanitizing barrier (which is why {@link setKey} is not flagged)
+ *  but does NOT model an equality chain, `Set.has()`, or `Object.defineProperty` — all three were
+ *  tried. An allow-list is not available here: `repoPath` is arbitrary filesystem input, so any
+ *  charset narrow enough to exclude `__proto__` could also reject a legitimate path and silently
+ *  freeze the map — the exact failure this guard exists to prevent. The guard below is complete
+ *  (those three names are the only ones reachable by a `[[Set]]` write), so the code stays in its
+ *  clearest form rather than being contorted to satisfy the query. */
 export function setPathKey<T>(rec: Record<string, T>, key: string, value: T): Record<string, T> {
   if (key === "__proto__" || key === "constructor" || key === "prototype") return rec;
-  const next: Record<string, T> = { ...rec };
-  Object.defineProperty(next, key, {
-    value,
-    writable: true,
-    enumerable: true,
-    configurable: true,
-  });
-  return next;
+  return { ...rec, [key]: value };
 }
 
 /** Strip prototype-polluting own keys from a payload before an `Object.assign`.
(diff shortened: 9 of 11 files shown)


scoop (commented):
## Verdict

The implementation is mechanically sound and I found no correctness or security defect in the guarded code paths. Two findings are documentation-accuracy defects inside `ui/src/lib/safe-keys.ts` — the file whose entire purpose is to carry the security rationale — so they are worth fixing but do not block.

## What I verified

**Guard placement matches the plan's per-site enumeration, exactly.** `grep` for all guard consumers returned 23 write sites: `ui/src/lib/store.svelte.ts:271, 350, 358, 366, 373, 496, 599, 602, 605, 857, 860, 863` (12), `ui/src/lib/reviews.svelte.ts:67, 79, 90, 109, 199, 209, 218, 240` (8), `ui/src/lib/recaps.svelte.ts:19`, `ui/src/lib/epic-draft.svelte.ts:31`, `ui/src/lib/buildQueues.svelte.ts:14` (3). `ui/src/lib/store.svelte.ts:609` is the pre-existing `#1630` call site, unchanged. Nothing in `AutomationStore` (`ui/src/lib/reviews.svelte.ts:314, 502, 504, 566, 574`) was touched, and the `epics` composite-key write was not touched — the stated scope bound holds.

**The regex is not duplicated.** `grep -rn "0-9a-zA-Z-" ui/src/lib/*.ts` (excluding tests) returns exactly one hit: `ui/src/lib/safe-keys.ts:44`. The old private copy at `store.svelte.ts` is deleted in the diff. `SAFE_ID` is non-global, so `.test()` carries no `lastIndex` state across the 23 call sites.

**No import cycle.** `ui/src/lib/safe-keys.ts` has zero imports, so it is a genuine leaf; the stated reason for not putting the helpers in `store.svelte.ts` (which imports `buildQueues.svelte`, `epic-draft.svelte`, `post-merge-steps.svelte` at `store.svelte.ts:42-44`) checks out.

**The silent-drop risk is bounded — every `setKey` key really is a UUID.** Every `setKey` call site is fed a session id off a WS payload or a bootstrap GET (`store.svelte.ts:643, 646, 649, 652, 656, 660, 663` are the only cross-store dispatch points). Server-side, session ids are `randomUUID()` — `src/service.ts:2965` mints it and passes it explicitly to `src/store.ts:1980` (`input.id ?? randomUUID()`), and `store.create` has exactly two non-test callers (`src/service.ts:2994`, reached from `src/drain.ts:1974, 2594` via `service.create`). Demo ids are alnum+hyphen (`ui/src/lib/demo/state.ts:370` `epic-${child.number}`, `:419` `spawned-${id}`; `ui/src/lib/demo/seed.ts:133-344`), so they pass the charset too. The one id that did *not* pass was a test fixture — `"phone, large text"` at `ui/src/lib/components/EpicDraftPanel.browser.test.ts:75` — and the diff slugifies it at `:82`. That is a correct, necessary fix, not scope creep.

**`setPathKey` is correctly applied where the charset would have frozen the map.** `ui/src/lib/store.svelte.ts:857, 860` key on `ev.data.repoPath`; repo paths are absolute filesystem paths (`ui/src/lib/reviews.svelte.ts:720-721` reads the same shape), which `SAFE_ID` rejects. The regression tests at `ui/src/lib/store.svelte.test.ts:1402-1411` and `ui/src/lib/reviews.svelte.test.ts:853-862` pin exactly that failure mode.

**`safeMerge` closes the one genuinely `[[Set]]`-reachable path.** `ui/src/lib/store.svelte.ts:401` is the only `Object.assign` onto a live session; `applyStatus` (`:411-413`) builds the patch by spreading the decoded payload, and object spread copies an own `__proto__` faithfully, so the concern is real. `Object.fromEntries(Object.entries(...))` rebuilds via `CreateDataProperty`, and `Object.entries` sees exactly the enumerable own keys `Object.assign` would have copied — so the filter loses nothing `Object.assign` would have applied. `Partial<Session>` has no `constructor`/`prototype` field, so stripping those changes no legitimate patch.

**`ReviewsStore.setActivity`'s read-gate is justified, not paranoia.** `pushActivity` (`ui/src/lib/reviews.svelte.ts:24-28`) does `[...feed, summary]`; with `id === "__proto__"`, `this.activity[id]` resolves to `Object.prototype`, which is non-nullish (so `?? []` does not fire) and non-iterable — the spread would throw `TypeError`. Gating before the read is the right call.

**`SAFE_ID` permitting `constructor`/`prototype` is not exploitable at `store.svelte.ts:202`,** even though that is a real `[[Set]]` write. `folded` is a plain object from `Object.fromEntries` (`:193-195`); `Object.prototype.constructor` is a *writable data* property, so `OrdinarySet` creates an own property on the receiver rather than invoking anything, and `Object.prototype` has no `prototype` at all. The test at `ui/src/lib/store.svelte.test.ts:1396` asserts precisely this. Safe — but see finding 1.

**Test fixtures resolve.** `GIT` (`store.svelte.test.ts:19`), `session()` (`:27`), `DRAIN` (`:77`), `AUTOMERGE_STATUS` (`:929`), `QUEUE` (`:971`); `rc` (`reviews.svelte.test.ts:24`), `verdict` (`:54`), `planGate` (`:68`), and `repoConfig.toggleAutopilot`/`isAutopilotEnabled` (`reviews.svelte.ts:499, 720`).

## Findings detail

**1 — `safe-keys.ts:36-43`: the `SAFE_ID` note's rationale is contradicted by its own second consumer.** Lines 36-40 say permitting `constructor`/`prototype` "is safe here and only here: these helpers write via object-literal computed keys (CreateDataProperty) … Where a `[[Set]]` write is involved, `setPathKey` and `safeMerge` reject all three names instead." Two lines later (`:42`) the same block documents `SAFE_ID`'s other consumer as `HerdStore.setClaudeAlive`'s stranded-id loop — which is `folded[id] = "stranded"` at `ui/src/lib/store.svelte.ts:202`, a `[[Set]]` write that does *not* reject those two names. The code is still safe (verified above), but the written justification does not cover the consumer it names, and `store.svelte.ts:197-201` itself calls that write out as `[[Set]]`. The note should state the actual reason it is safe there (`Object.prototype.constructor` is a writable data property; `prototype` is absent) rather than claiming no `[[Set]]` consumer exists.

**2 — `safe-keys.ts:10-12`: the stated invariant is stronger than what the change achieves.** The header says the helpers "stop a `__proto__` own key entering the maps at all — which matters because a later `[[Set]]`-based copy of such a map … WOULD reach the setter." That invariant does not hold for the maps in question: `HerdStore.setClaudeAlive` rebuilds `claudeAlive` from the remote map via `Object.fromEntries(Object.entries(map))` (`ui/src/lib/store.svelte.ts:193-195`) with no key filter, and `ReviewsStore.load` assigns a decoded response wholesale at `ui/src/lib/reviews.svelte.ts:47` (`this.map = await getReviews()`) and `:54-57`. Both can place an own `__proto__` into a guarded map. The file already lists bulk seeds under "Deliberately NOT routed through these" (`:24-27`), so the exclusion is intentional — but the header sentence asserts the precondition is foreclosed when it is only narrowed. Softening it to "stop a `__proto__` own key entering via these write paths" would make the two paragraphs agree.

- `ui/src/lib/reviews.svelte.ts:120-122` and `:126-128` (mirrored at `:250-252`) read `this.activity[id]` with no key gate, so `activityFeed("__proto__")` returns `Object.prototype` — `?? []` does not fire on a non-nullish value — which a `{#each}` consumer would then fail to iterate; the diff adds exactly this gate to the sibling write path at `:106`/`:237` but not to the readers. Dormant today because every id is a `randomUUID()`.
- `ui/src/lib/epic-draft.svelte.ts:31-32` fires every listener (`for (const fn of this.listeners) fn(d)`) even when `setKey` dropped the write, so a listener that calls `epicDrafts.get(d.sessionId)` would see nothing; unreachable while ids are UUIDs.

- `ui/src/lib/safe-keys.ts:78` — `setPathKey` rejects `constructor` and `prototype` although its write form is an object-literal computed key (`CreateDataProperty`), which cannot pollute under any of the three names; only the `__proto__` check does work here.
- `ui/src/lib/store.svelte.test.ts:1433-1449` — the "path-keyed writes stay enumerable, spreadable and serialisable" test pins three access shapes of a plain object-literal spread that no part of this change alters.

- `ui/src/lib/store.svelte.ts:181` `setActivity(map)` and the other bulk seeds assign server-provided objects wholesale, so an own `__proto__` can still enter those maps; explicitly excluded by the plan.
- `ui/src/lib/reviews.svelte.ts` `AutomationStore`'s `repoPath` computed writes remain unguarded literals, as the plan intended.

## Stated limitations
- Whether CodeQL actually accepts `SAFE_ID` as a sanitizing barrier at all 20 charset sites cannot be verified from this repo — no scan results are checked in. The in-repo evidence cited by the plan (`store.svelte.ts:609` routing through `setKey` and not appearing in the alert set) is consistent with it, but the alert set itself is not present here.
- I did not run the test suite (per instructions); CI covers it.

scoop (commented):

**1. `safe-keys.ts` SAFE_ID note claimed no `[[Set]]` consumer — RESOLVED.** `ui/src/lib/safe-keys.ts:43-52` now states the real descriptor facts: `Object.prototype.constructor` is a writable data property (no setter runs on a receiver assignment) and `Object.prototype.prototype` does not exist, so both names are plain own-property creations *including* for the `[[Set]]` consumer, which it names explicitly. That consumer is `folded[id] = "stranded"` at `ui/src/lib/store.svelte.ts:202`, and `:54-55` still attributes it correctly. Verified the accessor claim: `constructor`, `hasOwnProperty`, `toString`, `__defineGetter__` etc. are data properties on `Object.prototype`; `__proto__` is the sole accessor — and the new test pins exactly that at `ui/src/lib/store.svelte.test.ts:1450-1457`.

**2. Header overclaimed a whole-map invariant — RESOLVED.** `ui/src/lib/safe-keys.ts:10-20` now scopes the guarantee to "*on the guarded write paths*" and adds an explicit "This is NOT a whole-map invariant" paragraph naming both bypasses. I checked both against ground truth: `ReviewsStore.load`'s `this.map = await getReviews()` is at `ui/src/lib/reviews.svelte.ts:47` (and the same shape at `:54-57`), and `setClaudeAlive`'s `Object.fromEntries(Object.entries(map)…)` is at `ui/src/lib/store.svelte.ts:193-195`. The narrowed sentence is accurate.

## Verification of the rest of the change

- **Every session-id-keyed computed write is now guarded, and nothing else was touched.** Grepping `{ ...this.<map>, [` across `ui/src/lib` leaves only `store.svelte.ts:299` (the `epics` composite key) and the 94 `[repoPath]:` writes in `RepoConfigStore` (`reviews.svelte.ts:309-844`) — exactly the two exclusions documented at `safe-keys.ts:28-30`. No `[id]`/`[d.id]`/`[q.sessionId]` literal write survives unguarded.
- **`epics` cannot be attacked despite staying unguarded**: the key is built as `` `${e.repoPath}#${e.parentIssueNumber}` `` (`store.svelte.ts:290`), so it always contains `#` and can never equal `__proto__`. Not a latent defect.
- **`patchSession` is genuinely the only `[[Set]]` merge of remote data.** `Object.assign` appears in app code exactly once, at `store.svelte.ts:401`; the only other hits are two array helpers in browser tests (`ComposeBar.browser.test.ts:61`, `MicButton.browser.test.ts:54`). The header's "the one place" claim at `safe-keys.ts:12-13` holds.
- **The only computed `[[Set]]` assignment in the stores is guarded.** Grepping `\w[\w] = ` across `*.svelte.ts` yields only `store.svelte.ts:202` (guarded) and `post-merge-steps.svelte.ts:74` (array index).
- **Silent-freeze risk is contained.** `setKey`'s charset would reject a path, so `drain`/`autoMerge` correctly use `setPathKey` (`store.svelte.ts:857,860`) and `RepoConfigStore` was left alone; both are pinned by regression tests (`store.svelte.test.ts:1420-1440`, `reviews.svelte.test.ts:886-895`). Session ids are `randomUUID()` — `src/store.ts:1980` mints them and `id` is in the `Omit<>` list of `NewSession` (`src/store.ts:273-275`), so no caller can supply a non-UUID id.
- **No silently-broken consumers.** The old `setKey` was a module-private function in `store.svelte.ts`; nothing imported it. The new import-cycle rationale (`safe-keys.ts:3-5`) is real — `store.svelte.ts:34,35,42,43` imports all four sibling stores.
- **Non-UUID ids in existing fixtures.** `EpicDraftPanel.browser.test.ts:82` legitimately had to slugify because the label `"phone, large text"` (`:76`) contains a comma/space; I checked the other fixture id sites (`EpicDraftModal.browser.test.ts:49,97` → `desktop`/`narrow`; `epic-approve.test.ts:41` → `s-N`; `demo/seed.ts` ids are letters+hyphens) and none is affected.
- **`setPathKey`'s `Object.fromEntries` rewrite is semantically equivalent**: `AddEntriesFromIterable` uses `CreateDataPropertyOrThrow`, so no setter can run, and a duplicate key keeps its original insertion position — pinned by `store.svelte.test.ts:1440-1448`.

### Nits found

- `ui/src/lib/safe-keys.ts:28` (and the new test comment at `ui/src/lib/reviews.svelte.test.ts:888`) name a class `AutomationStore`. No such identifier exists anywhere in the tree; the class holding those `repoPath` maps is `RepoConfigStore` (`ui/src/lib/reviews.svelte.ts:271`, exported as `repoConfig` at `:849`). The "~90 writes" count is right (94 `[repoPath]:` writes), only the name is wrong.
- `ui/src/lib/safe-keys.ts:101` says "`Object.keys` sees exactly the enumerable own keys a decoded WS payload carries", but `safeMerge` uses `Object.hasOwn` and `Object.entries` (`:108,110`) — `Object.keys` is not used at all.

Neither is a behaviour bug, so this is a comment rather than request-changes.

- `ui/src/lib/store.svelte.test.ts:1440-1457` — the "enumerable, spreadable and serialisable" and "descriptor facts" tests pin JS-engine semantics rather than this repo's behaviour; harmless but beyond what the task asked for. The plan does anticipate them, and they document the reasoning, so keeping them is a fair call.

- `ui/src/lib/reviews.svelte.ts:85` — `if (!!this.reviewing[id] === on) return;` reads through the prototype chain, so `reviewing["__proto__"]` is truthy; unchanged by this PR and now unreachable in practice, but the read-side hazard the diff fixed in `setActivity` still exists in `isReviewing`/`activityFor`/`reviewerEnvFor`.

scoop (commented):

**1. `AutomationStore` → `RepoConfigStore` — FIXED.** `ui/src/lib/safe-keys.ts:28` now reads "`RepoConfigStore`'s 93 `repoPath` writes (`reviews.svelte.ts`, exported as `repoConfig`)", and the test comment at `ui/src/lib/reviews.svelte.test.ts:888` now says "RepoConfigStore is keyed by filesystem path". Both pointers resolve: `class RepoConfigStore` is at `ui/src/lib/reviews.svelte.ts:271` and `export const repoConfig = new RepoConfigStore()` at `ui/src/lib/reviews.svelte.ts:849` (no `AutomationStore` identifier exists anywhere in the tree — the only class declarations in that file are `ReviewsStore:32`, `PlanGateStore:152`, `RepoConfigStore:271`).

**2. `safeMerge` doc citing `Object.keys` — FIXED.** `ui/src/lib/safe-keys.ts:102-104` now reads "The probe uses `Object.hasOwn` … the rebuild uses `Object.entries`", which matches the implementation at `ui/src/lib/safe-keys.ts:110` (`Object.hasOwn`) and `:112` (`Object.entries`). No `Object.keys` call remains in the file.

## What I verified in the new revision

- **Guard site coverage matches the plan's 23.** After the diff the only remaining `{ ...this.X, [k]: v }` computed writes in the guarded modules are `ui/src/lib/store.svelte.ts:299` (the `epics` composite key) and the 94 `RepoConfigStore` writes in `ui/src/lib/reviews.svelte.ts:309-844` — exactly the two sets `ui/src/lib/safe-keys.ts:28-31` documents as out of scope. No stray site was rewritten.
- **Every introduced identifier resolves.** `SAFE_ID`, `setKey`, `setPathKey`, `safeMerge` are all exported from `ui/src/lib/safe-keys.ts:57,71,88,108` and each is consumed: `setKey` at `ui/src/lib/store.svelte.ts:271,350,358,366,496,599,602,605,864` etc., `setPathKey` at `ui/src/lib/store.svelte.ts:857,860`, `safeMerge` at `ui/src/lib/store.svelte.ts:401`, `SAFE_ID` at `ui/src/lib/store.svelte.ts:202` and `ui/src/lib/reviews.svelte.ts:106,249`. The old module-private `setKey` in `store.svelte.ts` was not exported, so its removal has no out-of-file consumers.
- **The import-cycle rationale at `ui/src/lib/safe-keys.ts:3-5` is true**: `store.svelte.ts` imports `reviews.svelte` (`:34`), `recaps.svelte` (`:35`), `buildQueues.svelte` (`:42`), `epic-draft.svelte` (`:43`), so exporting the helpers from there would indeed have created four cycles.
- **`safeMerge` is on the only `Object.assign` of remote data.** The sole non-test `Object.assign` in `ui/src/lib` is `ui/src/lib/store.svelte.ts:401`; `applyStatus` (`:410`) spreads a decoded payload into the patch, so an own `__proto__` really would have reached `[[Set]]` there before this change.
- **The `setActivity` read-gate is a real fix, not decoration.** `pushActivity` (`ui/src/lib/reviews.svelte.ts:24-28`) does `[...feed, summary]`; with `id === "__proto__"` the read `this.activity[id]` returns `Object.prototype`, which is not iterable → `TypeError`. Gating before the read (`:106`, `:249`) prevents that, and both `setActivity` overloads are gated symmetrically.
- **Charset-vs-path split is correct for the real key shapes.** Session ids are `randomUUID()` (`src/service.ts:2962,2965,3277`) → match `/^[0-9a-zA-Z-]+$/`; `drain`/`autoMerge` are keyed by `repoPath` and correctly use `setPathKey` (`ui/src/lib/store.svelte.ts:857,860`). I swept every session-id literal that feeds a now-guarded store: the demo seed ids (`ui/src/lib/demo/seed.ts:133-1179`) are all alphanumeric+hyphen, `epic-approve.test.ts:41`'s `nextSid()` yields `s-N`, and the only test literals with charset-illegal characters are the deliberate `__proto__` cases. The one previously-illegal caller — `EpicDraftPanel.browser.test.ts` with labels like `"phone, large text"` — is slugged at `:82`; the other two call sites in that file (`:43` `epic-draft-bar-${label}` with labels `desktop`/`narrow`, and `:115` `epic-draft-bar-cta`) are already charset-clean, so leaving them untouched is right.
- **`setPathKey`'s `Object.fromEntries([...Object.entries(rec), [key, value]])` preserves semantics**: later duplicate wins (update), insertion order preserved for new keys, and `Object.fromEntries` uses CreateDataProperty so it cannot invoke a setter even if `rec` already carries an own `__proto__` from a bulk seed (`setDrain` at `ui/src/lib/store.svelte.ts:261-263`).
- **`dropKey` left unguarded** (`ui/src/lib/store.svelte.ts:1051-1059`) with the rationale in its JSDoc — consistent with the plan and with the inline `delete copy[id]` forms at `ui/src/lib/reviews.svelte.ts:69,92,115,217` and `ui/src/lib/recaps.svelte.ts:20`.
- **Test helpers referenced by the new tests exist**: `GIT:19`, `session():27`, `DRAIN:77`, `AUTOMERGE_STATUS:929`, `QUEUE:971` in `ui/src/lib/store.svelte.test.ts`; `rc():24`, `verdict():54`, `planGate():68`, mocked `putRepoConfig:10,20` in `ui/src/lib/reviews.svelte.test.ts`; `epicDrafts.map` is public (`ui/src/lib/epic-draft.svelte.ts:11`).

No user-facing strings or message-catalog keys are touched by this diff, so there is no i18n-parity surface to check.

### Stated limitation

I could not verify the central premise that CodeQL accepts `SAFE_ID` as a sanitizing barrier (and therefore that the 22 alerts will close) — the alert set lives in GitHub code scanning, not in this repo. Assumed as written; the plan already books "survivors get dismissed post-merge" as the fallback.

- `ui/src/lib/safe-keys.ts:112` — on the dirty path `safeMerge` rebuilds via `Object.entries`, which drops own **symbol** keys that `Object.assign` would otherwise copy. Unreachable today (every `patchSession` patch is either a JSON-decoded payload or an object literal, `ui/src/lib/store.svelte.ts:410,432,435,441,449,455`), but a future symbol-carrying patch would lose those keys only when a `__proto__`/`constructor`/`prototype` key is also present.
- `ui/src/lib/epic-draft.svelte.ts:31-33` — `upsert` still notifies every listener after `setKey` has dropped the write, so a listener could observe a draft that is not in the map. Only reachable with a charset-illegal `sessionId`, which no server-minted id can be.

- `ui/src/lib/reviews.svelte.ts:80,85` — `setReviewing` uses `id in this.reviewerEnv` / `this.reviewing[id]`, both of which resolve through the prototype chain, so a `__proto__`/`constructor` id causes needless map churn (harmless, and unreachable with real ids).

scoop (commented):
## Re-review of the previous round's findings

**1. `safe-keys.ts` — the wrong `93 repoPath writes` count.** Fixed. `ui/src/lib/safe-keys.ts:28-29` now reads "Every `repoPath` write in `RepoConfigStore` … deliberately uncounted here, since the figure drifts as config flags are added" — the exact number is gone, so it cannot drift again. (Ground truth for why the number was a liability: `ui/src/lib/reviews.svelte.ts:309-335` alone contains 27 such writes, with more at `:440-495` and beyond, all inside `RepoConfigStore`, which spans `ui/src/lib/reviews.svelte.ts:271` onwards.)

**2. `store.svelte.ts:200-201` — the "keeps `husk`" comment only covered one of two cases.** Fixed. `ui/src/lib/store.svelte.ts:202-206` now spells out both branches explicitly: "if the boolean snapshot carried it, it keeps whatever that said (`alive` or `husk`); if it did NOT, the entry is absent entirely, where the pre-guard code would have created it. Absent is not the same as `husk`". I checked the "absent" semantics it appeals to against the field doc it cites — `ui/src/lib/store.svelte.ts:117` says "Absent = not swept yet (treated as unknown → Resume stays offered)" — so the reworded comment is accurate.

## What I verified in this revision

**Every identifier the diff introduces resolves.** `SAFE_ID`, `setKey`, `setPathKey`, `safeMerge` are all exported from `ui/src/lib/safe-keys.ts:60,73,90,116` (approx.) and every import matches: `ui/src/lib/store.svelte.ts:45` imports all four and uses each (`SAFE_ID` at `:207`, `safeMerge` at `:406`, `setKey` at `:276,355,363,371,378,501,604,607,610,614,868`, `setPathKey` at `:862,865`); `ui/src/lib/reviews.svelte.ts:5` imports `SAFE_ID, setKey` and uses both (`SAFE_ID` at `:106,238`; `setKey` at `:67,79,90,109,199,209,218,240`); `recaps.svelte.ts:3`, `epic-draft.svelte.ts:3`, `buildQueues.svelte.ts:2` each import only `setKey` and use it once. `safe-keys.ts` has no imports of its own, so the stated cycle-avoidance rationale (`safe-keys.ts:3-5`) holds.

**The `[[Set]]` chokepoint is genuinely the only one.** `Object.assign` on remote data is at `ui/src/lib/store.svelte.ts:406` (`patchSession`), reached by `applyStatus` at `:415-417` and by nine other callers (`:391, 437, 440, 446, 454, 460, …`). Guarding inside `patchSession` rather than `applyStatus` does cover all of them. The other `[[Set]]` write is the bare `folded[id] = "stranded"` at `:207`, now gated by `SAFE_ID`.

**`safeMerge` is sound for the payload it actually sees.** `applyStatus` builds `{ ...d, status: d.status }` (`store.svelte.ts:415`), and object spread copies an own `__proto__` created by `JSON.parse` as an own enumerable key — which `Object.hasOwn` in `safe-keys.ts` detects and `Object.entries` then omits from the rebuild. Non-enumerable own keys are dropped by the rebuild, but `Object.assign` would not have copied them either, so no behaviour is lost.

**`setPathKey`'s `Object.fromEntries([...Object.entries(rec), [key, value]])` is semantically equivalent to the spread it replaced.** Duplicate-key entries keep the first occurrence's position and the last occurrence's value, matching `{ ...rec, [key]: v }`; both use CreateDataProperty, so neither can invoke a setter. `ui/src/lib/store.svelte.test.ts:1451-1467` pins key order, spreadability and JSON round-tripping, so a future change to the write form can't quietly break a consumer.

**`SAFE_ID` allowing `constructor`/`prototype` is safe for the one `[[Set]]` consumer**, as documented at `safe-keys.ts:44-53` — `Object.getOwnPropertyDescriptor(Object.prototype, "constructor")` is a writable data property and `…, "prototype")` is `undefined`, both asserted directly in `ui/src/lib/store.svelte.test.ts:1469-1487` rather than left as prose.

**Scope holds.** Grepping the five touched store files for the remaining `{ ...x, [k]: v }` form leaves only `ui/src/lib/store.svelte.ts:304` (`epics`, composite `${repoPath}#${issue}` key) and the `RepoConfigStore` `repoPath` writes in `ui/src/lib/reviews.svelte.ts:309-…` — exactly the two exclusions the file-level comment at `safe-keys.ts:28-31` names. The five `delete` sites are untouched and `dropKey` (`store.svelte.ts:1062-1066`) carries a comment explaining why.

**The charset narrowing does not break any live key.** Session ids are `randomUUID()` (`src/service.ts:2965`, passed explicitly into `store.create` per `src/service.ts:2964` and consumed at `src/store.ts:1980`), which matches `/^[0-9a-zA-Z-]+$/`. I swept the UI tests for ids fed to the now-guarded setters that fall outside the charset: the only pre-existing offender was the `"phone, large text"` label at `ui/src/lib/components/EpicDraftPanel.browser.test.ts:76`, and the diff slugifies it at `:82`. The sibling `${label}` interpolations at `EpicDraftPanel.browser.test.ts:43` and `EpicDraftModal.browser.test.ts:49,97` draw from `["desktop", "narrow"]` (`:38-40`, `:44-46`, `:88-…`), which are charset-clean, so leaving them unslugified is correct rather than an omission. `RepoConfigStore`'s path keys are exercised against the guard by `ui/src/lib/reviews.svelte.test.ts:853-862` and `ui/src/lib/store.svelte.test.ts:1420-1429`.

**Reads, not just writes, were considered.** The gate added at `ui/src/lib/reviews.svelte.ts:106` and `:238` is placed *before* `pushActivity(this.activity[id], …)`, which is the one read where `this.activity["__proto__"]` resolving to `Object.prototype` would have been spread. I walked the other inherited-lookup reads in the same file — `!!this.reviewing[id]` (`:85`, `:217`), `id in this.reviewerEnv` (`:80`, `:211`), `id in this.activity` (`:113`) — and each either short-circuits or performs a `delete` on a non-own property, so all are inert.

Stated limitation: whether CodeQL actually accepts the `SAFE_ID` regex (and the equality chain in `setPathKey`) as sanitizing barriers cannot be verified from this repo — no alert database is present. The plan already treats surviving alerts as a post-merge dismissal task, which is the right handling.

## Nit

The `SAFE_ID` docblock's "Shared by …" line is now incomplete — see findings.

Latent / future-reachable (non-blocking):
- `ui/src/lib/epic-draft.svelte.ts:30-33`: when `setKey` rejects `d.sessionId` the map is left unchanged but `for (const fn of this.listeners) fn(d)` still fires, so subscribers are notified of a draft that is not in the store; unreachable while ids are `randomUUID()`s, but it would surface as a desync the moment a non-charset id reaches this path.

scoop (commented):
## Prior finding

**1. `safe-keys.ts` `SAFE_ID` docblock enumerated a stale consumer list — ADDRESSED.** The enumeration is gone; `ui/src/lib/safe-keys.ts:56-61` now says consumers are "deliberately not enumerated here … `grep SAFE_ID` is both accurate and cheaper than a comment" and states the general reason a consumer may call it directly. No stale list remains.

## What I verified

**Every guarded site, and nothing outside the plan's 23.** Grepped `setKey|setPathKey|safeMerge|SAFE_ID` across `ui/src`: 12 in `store.svelte.ts` (`:207, 276, 355, 363, 371, 378, 501, 604, 607, 610, 862, 865, 868` — 13 including the pre-existing `:614` `setKey` call from #1630), 8 in `reviews.svelte.ts` (`:67, 79, 90, 106/109, 199, 209, 218, 237/240`), 1 each in `recaps.svelte.ts:19`, `epic-draft.svelte.ts:31`, `buildQueues.svelte.ts:14`. A `\.\.\.this\.[A-Za-z]+, \[` sweep shows the only remaining raw computed-literal writes are `store.svelte.ts:304` (the `epics` composite key) and the `RepoConfigStore` `repoPath` writes (`reviews.svelte.ts:309-844`) — exactly the sites the plan declares out of scope. The five delete sites (`recaps.svelte.ts:22`, `reviews.svelte.ts:70, 93, 218`, `dropKey` at `store.svelte.ts:1066`) are untouched, as intended, and `dropKey`'s new docblock (`store.svelte.ts:1059-1064`) records why.

**`setPathKey`'s rebuild is semantically identical to the spread it replaces.** `Object.fromEntries([...Object.entries(rec), [key, value]])` (`ui/src/lib/safe-keys.ts:98-101`) uses CreateDataProperty per entry, so no setter can run; re-supplying an existing key redefines it in place, preserving insertion order — pinned by `ui/src/lib/store.svelte.test.ts:1428-1440` (`Object.keys` order, spread, JSON round-trip).

**`safeMerge` is at the right place.** `patchSession` (`ui/src/lib/store.svelte.ts:400-407`) is the only `Object.assign` in non-test `ui/src` (verified by grep; the only other hits are `ComposeBar.browser.test.ts:61` and `MicButton.browser.test.ts:54`, both on arrays), and it is reached by all ten `patchSession` callers including `applyStatus:415`. The `Object.hasOwn` probe correctly catches the own `__proto__` that `JSON.parse` + spread produce, and the test at `ui/src/lib/store.svelte.test.ts:1364-1376` drives it through the real WS dispatch seam rather than calling `patchSession` directly.

**The UUID assumption behind the strict charset holds.** Session ids are minted at `src/service.ts:2965` (`randomUUID()`) and passed explicitly into `store.create({ id: sessionId, … })` at `src/service.ts:2994`; `src/store.ts:1980` only falls back to `randomUUID()` when no id is supplied. Hex + hyphens match `/^[0-9a-zA-Z-]+$/`.

**The silent-freeze risk is genuinely tested, and already bit once.** `ui/src/lib/reviews.svelte.test.ts:884-894` round-trips `/home/u/Work/my-repo.git` through `toggleAutopilot` (verified `repoConfig.autopilot` at `reviews.svelte.ts:277` and `toggleAutopilot` at `:499`), and `store.svelte.test.ts:1397-1406` does the same for `drain`/`autoMerge`. The `EpicDraftPanel.browser.test.ts:82` slug fix is a real instance of the drop being caught: the `"phone, large text"` label (`:76`) would otherwise have been silently discarded by `epicDrafts.upsert`.

**Test fixtures resolve.** `rc`/`verdict`/`planGate` at `reviews.svelte.test.ts:24, 54, 68`; `GIT`/`session`/`DRAIN`/`AUTOMERGE_STATUS`/`QUEUE` at `store.svelte.test.ts:19, 27, 77, 929, 971`; `DRAIN.inFlight` is `1` (`:85`), matching the assertion at `:1404`. `session:archived` really does `dropKey` the git map (`store.svelte.ts:468`), so the delete-stays-unguarded test at `:1417-1425` exercises what it claims.

## The one gap

`SAFE_ID` admits `constructor` and `prototype` (pure letters). The docblock at `ui/src/lib/safe-keys.ts:45-46` says that is "safe for every consumer", and `:59-61` plus the identical comments at `reviews.svelte.ts:104-105` and `:236-237` say the gate exists specifically so a prototype-chain *read* cannot happen. For `constructor` it does happen: `this.activity["constructor"]` resolves up the chain to the `Object` function, `SAFE_ID.test("constructor")` passes, and `pushActivity` (`reviews.svelte.ts:24-28`) then evaluates `[...feed, summary]` on a function — functions are not iterable, so that is a `TypeError` thrown inside the WS dispatch at `store.svelte.ts:657` / `:668`. Not reachable with today's server-minted UUID ids, which is why I am not blocking, but the two documentation claims are verifiably wrong as written and the sibling name to the one being guarded is left open.

- `ui/src/lib/reviews.svelte.ts:80` / `:113`: `id in this.reviewerEnv` and `id in this.activity` are true for any inherited name (`"__proto__" in {}` → true), so a hostile id still triggers a pointless map copy + reactive write in `setReviewing`/`clearActivity`. Harmless (the `delete` is a no-op) and pre-existing, but it means the guard does not make those paths fully inert.

- `ui/src/lib/reviews.svelte.ts` `activityFor`/`activityFeed` read `this.activity[id]` ungated, so an id of `constructor` would return the `Object` function typed as `string[]` — same root cause as the finding above, on untouched lines.

scoop (commented):
## Re-review: prior finding resolved

The single prior finding — `SAFE_ID` admitting `constructor`/`prototype`, letting `ReviewsStore.setActivity`/`PlanGateStore.setActivity` read `this.activity["constructor"]` (the `Object` function) and throw `TypeError: … is not iterable` in `pushActivity`'s spread, while the guard was documented as preventing a prototype-chain *read* — is **fully addressed**. The offered resolutions were (a) exclude `constructor`/`prototype` on the read-gated paths, or (b) correct the claims; the diff does both:

1. **Read hardened** — `ui/src/lib/reviews.svelte.ts:30` changes `const feed = cur ?? []` to `const feed = Array.isArray(cur) ? cur : []`. Verified against the failure path: a letter-only inherited name passes `SAFE_ID` (`ui/src/lib/safe-keys.ts:65`), so `this.activity[id]` resolves up the prototype chain to a function/prototype object; `Array.isArray(<function>)` is `false`, so `feed` becomes `[]` and the `[...feed, summary]` spread (`ui/src/lib/reviews.svelte.ts:32`) can no longer throw. The dedup read `feed[feed.length - 1]` on `[]` is `undefined` — also safe.
2. **Claims corrected** — `ui/src/lib/safe-keys.ts:45-59` now states the regex is a *write* (prototype-pollution) barrier only and "does NOT make a subsequent READ of `map[id]` safe … The one consumer that reads before writing (`ReviewsStore`/`PlanGateStore.setActivity` → `pushActivity`) hardens that read itself." The two `setActivity` boundary comments (`ui/src/lib/reviews.svelte.ts:109-112`, `:242-245`) match: "This is NOT what makes the subsequent read safe … pushActivity is hardened against that." The stale "safe for every consumer"/"gate BEFORE the read" framing is gone.

Write-safety of the letter-only names is intact: with `id="constructor"`, `setKey` (`ui/src/lib/safe-keys.ts:77-80`) writes `{ ...rec, ["constructor"]: v }` via `CreateDataProperty`, landing `constructor` as an ordinary own property with no setter invocation — no pollution.

## Rest of the diff

Spot-checked against the plan and found internally consistent:
- `setKey` extracted to `ui/src/lib/safe-keys.ts` and consumed at the flagged UUID-keyed sites across `store.svelte.ts`, `reviews.svelte.ts`, `recaps.svelte.ts`, `epic-draft.svelte.ts`, `buildQueues.svelte.ts`; the old private `setKey` in `store.svelte.ts` is removed (verified: `ui/src/lib/store.svelte.ts:1069-1073` old block deleted, import added at `:45`).
- `setPathKey` (deny-list, `Object.fromEntries`) used for the two `repoPath`-keyed sites `drain:status`/`automerge:status` (`ui/src/lib/store.svelte.ts:862,865`) — correct, since `repoPath` would fail the `SAFE_ID` charset and `setKey` would silently no-op.
- `safeMerge` applied at the single `Object.assign` chokepoint `patchSession` (`ui/src/lib/store.svelte.ts:405`), covering all callers of that method; it strips only `__proto__`/`constructor`/`prototype` own keys and returns the input unchanged on the clean path.
- `:196` (`applyState`) guards its `folded[id] = "stranded"` [[Set]] assignment with `SAFE_ID.test(id)`; the accepted behaviour change (a non-charset stranded id loses its banner upgrade) is documented in the code comment.
- `dropKey` left deliberately unguarded with a comment explaining a rejected key would silently fail to delete — matches the plan's deletion-cannot-pollute argument.

No `src/` files touched; no CodeQL config change; diff scope matches the plan's Files table.

- `ui/src/lib/reviews.svelte.ts:127-135` — `activityFor`/`activityFeed` also do a bracket read of `this.activity[id]` and, for a hostile letter-only id, would return the inherited `Object` function rather than an array (a downstream `.map`/`.length` consumer could misbehave). Unchanged pre-existing lines, unreachable with real `randomUUID()` ids, and not a write/spread path so no crash in these methods themselves; noted only for completeness.

PR #1870, opened 2026-07-21
ci: gate site/ builds and assert on emitted artifacts

### LINKED ISSUE #1859: No CI build for site/ — the marketing site has no gate
`.github/workflows/ci.yml` has exactly two jobs — `verify` and `docs-site`. **Nothing builds `site/`.** There is no `working-directory: site` or `cd site` anywhere under `.github/workflows/`.

So the marketing site (shepherd.run) has no CI gate at all: its only protection is whatever the author runs locally, plus the Vercel preview deploy. `docs-site/` by contrast gets a full `bun run check` + `bun run build` on every PR.

This surfaced during the Astro 6 → 7 upgrade, where `site/` carried the riskier changes of the two (the stricter Rust compiler, and a Fonts API whose silent failure mode — falling back to system fonts — leaves both `astro check` and `astro build` green). That upgrade had to compensate with a throwaway local artifact-assertion script, because no committed gate existed to catch it.

## Proposal

Add a `site` job mirroring the existing `docs-site` one:

- bun-store cache keyed on `site/bun.lock`
- install via the shared `./.github/actions/bun-install-retry` composite action, `working-directory: site`
- `bun run check` then `bun run build`

## Worth considering at the same time

The Astro 7 upgrade added artifact-level assertions that a plain `astro build` cannot express — a green build does **not** prove the fonts pipeline still works. Candidates to fold into the job:

- assert woff2 files are emitted and the `@font-face` block is present in `dist/index.html`
- assert both `--font-*` CSS variables and the font preload links are present
- assert all three routes emit (`index`, `privacy`, `impressum`)
- assert scoped-style integrity: every `data-astro-cid-*` used in a CSS selector has a matching element (catches styles silently ceasing to apply)

The throwaway version of these lives in the Astro 7 PR body if it is useful as a starting point.

Deliberately deferred from that PR: adding a CI job is unrequested scope on a security bump, and that PR's success criteria forbade touching files outside `site/` and `docs-site/`.

scoop:
Closes #1859.

`site/` had **no CI job**. `ci.yml` had exactly two — `verify` and `docs-site` — and nothing under `.github/workflows/` referenced `site`. Its only protection was whatever the author ran locally, plus the Vercel preview. This adds a `site` job mirroring `docs-site`, **plus assertions on what the build actually emitted**.

## Why assertions, not just a build

**A green `astro build` does not prove the site is correct.** I reproduced a genuine silent-fallback build — point one family's `options.package` at a package it cannot resolve:

```
[WARN] [assets] No data found for font family JetBrains Mono
[assets] Copying fonts (3 files)...     ← healthy build copies 9
[build] ✓ Complete!                      exit 0
```

Against that `dist/`, every aggregate signal survives:

| Signal | Healthy | Degraded | Discriminates? |
| --- | --- | --- | --- |
| `@font-face` present anywhere | 11 blocks | 4 blocks | NO |
| `--font-space-grotesk` declared | yes | yes | NO |
| `--font-jetbrains-mono` declared | yes | **yes** | NO |
| ≥1 `.woff2` emitted | 9 | 3 | NO |
| ≥1 `rel=preload as=font` | 9 | 3 | NO |
| **Family has a `url()`-backed `@font-face`** | yes (6) | **zero** | **YES** |

In the degraded build `--font-jetbrains-mono` still resolves — to a family with **no `@font-face` rule at all**. Only the per-family check tells them apart, and it is verified end-to-end: `astro build` exits 0 on that build, the gate exits 1.

*Honest scope:* this probe models "the provider cannot resolve the family", which `--frozen-lockfile` makes near-impossible in CI. It is not a model of the Fonts API silently changing across an Astro major. That the same assertion catches that case too is reasoned, not measured.

## What ships

- **`site/scripts/check-build-artifacts.mjs`** — routes derived **recursively** from `src/pages/**` (a new or nested page is covered with no edit; underscore-prefixed paths are correctly not routes; anything unmappable fails loudly), both `--font-*` variables declared, the per-family webfont check, and font-file size floors.
- **`test/check-build-artifacts.test.ts`** — 15 cases on the repo's existing gate-script seam (`check-generated-docs` / `check-feature-catalog`): temp-dir fixtures, `spawnSync`, assert `{code, out}`. Rides `verify`'s existing `Test (root)` step. **Verified that loosening the per-family assertion to a global check reds it.**
- **`site` job** — `check` → `build` → `check:artifacts`.

## Decisions worth reviewing

- **No `paths:` filter.** A path-filtered required check reports no status on unrelated PRs, and a required check that never reports blocks merges permanently.
- **The 512 B woff2 floor is measured.** A truncated font passes both the presence floor and the per-family check. The smallest legitimate woff2 is **1160 B**, so a 1 KB floor would leave 136 B of headroom and red the gate whenever subsetting re-chunks.
- **The build's jsdelivr fetch is left unretried.** The npm font provider fetches woff2 from `cdn.jsdelivr.net` at *build* time (at the lockfile-pinned version), so this job needs network during `build`. Left unretried to match every other build step here — `bun-install-retry` wraps installs only, and this repo adds retries in response to *observed* flakes (#944, #1134), not in anticipation.
- **Deferred to follow-ups:** `url()`-resolves-on-disk and preload assertions (most false-red surface, no demonstrated failure), and `weights: [400,500,700]` being declared but only 400 emitted (a real `site/` config finding, mechanically explained: `@fontsource/*` ships weight 400 in `index.css`).

## Note on gate status

This job gates **Shepherd's own automation immediately**, ruleset or not: `rollupChecks()` is worst-of with no required/non-required distinction, and `checksCleared()` (`src/checks-gate.ts:25`) is true only on `"success"`. A red `site` job stalls auto-review, auto-merge, epic drain and the standalone critic for that PR. The manual step below covers *human* merges only.

```shepherd:manual-steps
- [ ] POST-MERGE: Add the `site` check to the branch ruleset's required checks (it only becomes selectable once it has reported on main)
```

## Verification

`site`: `check` (0 errors) → `build` → `check:artifacts` all green. Root: 7415 pass / 0 fail. Full pre-push suite green (prettier, eslint, gates, tsc, ext, root-tests, ui, fallow audit). One unrelated flake on the first pre-push run — `DiffFileStack.browser.test.ts`, a timing-sensitive browser test under 7-lane parallel load; passes in isolation and on retry, and this diff touches no `ui/` file.

[no-feature-entry] — CI infrastructure, no user-facing UX.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -225,3 +225,60 @@ jobs:
         run: |
           bun run check
           bun run build
+
+  # The shepherd.run marketing site (site/) is the 5th in-monorepo package with its
+  # own lockfile. Like docs-site it builds self-contained, so it needs only a plain
+  # checkout + its own install, and it lives in its OWN job for the same reason: an
+  # install flake there must not red the whole suite.
+  #
+  # Deliberately NOT filtered with `paths:` (#1859). A path-filtered job reports no
+  # status at all on PRs that don't touch those paths, and a required check that never
+  # reports blocks the PR from merging forever. It runs on every PR instead; the build
+  # is sub-second.
+  #
+  # NB the fonts pipeline reaches cdn.jsdelivr.net at BUILD time (the npm font provider
+  # fetches the woff2 for the lockfile-pinned @fontsource version), so this job needs
+  # network during `build`, not just during install. A fetch failure fails the build
+  # loudly. That is deliberately left unretried, matching every other build step here —
+  # `bun-install-retry` wraps installs only, and the repo adds retries in response to
+  # observed flakes (#944, #1134), not in anticipation of one.
+  site:
+    name: site
+    # Skip release-please's generated PR, mirroring the two jobs above.
+    if: ${{ !(startsWith(github.head_ref, 'release-please--') && (github.event.pull_request.user.login == 'shepherd-release-please[bot]' || github.event.pull_request.user.login == 'github-actions[bot]')) }}
+    # Pinned to GitHub-hosted like the others — public repo, no CI_RUNNER toggle.
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v7
+
+      - name: Setup Bun
+        uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+        with:
+          bun-version: latest
+
+      - name: Cache bun store (site)
+        uses: actions/cache@v6
+        with:
+          path: ~/.bun/install/cache
+          key: ${{ runner.os }}-bun-site-${{ hashFiles('site/bun.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-bun-site-
+
+      - name: Install (site)
+        uses: ./.github/actions/bun-install-retry
+        with:
+          working-directory: site
+
+      # check + build, then assert on what the build actually emitted: a green
+      # `astro build` does NOT prove the site is correct. Point a font family's
+      # provider at a package it can't resolve and the build still exits 0 with only
+      # a [WARN], shipping a dist/ where that family silently degraded to system
+      # fonts. check:artifacts is what catches that; its own logic is unit-tested in
+      # test/check-build-artifacts.test.ts, which runs in `verify`.
+      - name: Build (site)
+        working-directory: site
+        run: |
+          bun run check
+          bun run build
+          bun run check:artifacts
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -233,8 +233,9 @@ jobs:
   #
   # Deliberately NOT filtered with `paths:` (#1859). A path-filtered job reports no
   # status at all on PRs that don't touch those paths, and a required check that never
-  # reports blocks the PR from merging forever. It runs on every PR instead; the build
-  # is sub-second.
+  # reports blocks the PR from merging forever. So it runs on every PR instead — the
+  # job is small next to `verify`, and correctness of the required-check contract is
+  # the reason, not cost.
   #
   # NB the fonts pipeline reaches cdn.jsdelivr.net at BUILD time (the npm font provider
   # fetches the woff2 for the lockfile-pinned @fontsource version), so this job needs
--- a/docs-site/README.md
+++ b/docs-site/README.md
@@ -28,9 +28,9 @@ Those gates are scoped to `ui/` + `extension/` and to `feat(...)` commits touchi
 root `tsconfig.json` `exclude` (it type-checks via `astro check`, not the server
 `tsc`).
 
-Unlike `site/`, this package **stays in the monorepo** and **is wired into CI**
-(`.github/workflows/ci.yml`): `cd docs-site && bun install --frozen-lockfile &&
-bun run check && bun run build`.
+Unlike `site/`, this package **stays in the monorepo**. Both are wired into CI
+(`.github/workflows/ci.yml`), each in its own job: `cd docs-site && bun install
+--frozen-lockfile && bun run check && bun run build`.
 
 ## Generated content
 
--- a/site/README.md
+++ b/site/README.md
@@ -34,8 +34,23 @@ build, every aggregate signal survives — `@font-face` blocks are still present
 emitted. The one check that discriminates is per-family: does the family each
 `--font-*` variable names actually have a `url()`-backed `@font-face`?
 
-`scripts/check-build-artifacts.mjs` asserts that, plus route coverage (derived from
-`src/pages/**`, so a new page is covered automatically) and font-file size floors.
+`scripts/check-build-artifacts.mjs` asserts that, plus route coverage and font-file
+size floors. The expected font families are read from `astro.config.mjs`, so adding
+one brings it under the gate automatically.
+
+Routes are derived from `src/pages/**`, so a new **static `.astro`** page — nested
+included — is covered with no change to the script. Anything the directory-format
+mapping cannot resolve to a single output path **fails the gate by design**, rather
+than being silently skipped:
+
+- a Markdown/MDX page (`about.md`) — the mapping only understands `.astro`
+- a dynamic or spread route (`[slug].astro`) — its outputs come from
+  `getStaticPaths` at build time, so they cannot be derived from the filename
+
+If you add either, extend `routeForPage()` to cover it; the red is telling you the
+gate would otherwise stop covering a route. Underscore- and dot-prefixed entries
+(`_components/`, `.DS_Store`) are not routes and are skipped.
+
 Its logic is unit-tested from the monorepo root in `test/check-build-artifacts.test.ts`,
 alongside the repo's other gate-script tests.
 
--- a/site/README.md
+++ b/site/README.md
@@ -36,7 +36,11 @@ emitted. The one check that discriminates is per-family: does the family each
 
 `scripts/check-build-artifacts.mjs` asserts that, plus route coverage and font-file
 size floors. The expected font families are read from `astro.config.mjs`, so adding
-one brings it under the gate automatically.
+one brings it under the gate automatically — and every configured family is required
+on **every** route. That holds because `src/layouts/Base.astro` renders one `<Font>`
+tag per family and all pages use that layout. A family scoped to a single page or
+layout would therefore fail the gate; supporting one means teaching the check which
+routes a family applies to, rather than assuming all of them.
 
 Routes are derived from `src/pages/**`, so a new **static `.astro`** page — nested
 included — is covered with no change to the script. Anything the directory-format
--- a/site/README.md
+++ b/site/README.md
@@ -14,10 +14,30 @@ Vercel project.
 ```bash
 cd site
 bun install
-bun run dev      # local dev server
-bun run build    # static output to dist/
-bun run preview  # serve the built dist/
+bun run dev             # local dev server
+bun run build           # static output to dist/
+bun run preview         # serve the built dist/
+bun run check:artifacts # assert on what the build emitted (run after build)
 ```
+
+## CI
+
+The `site` job in `.github/workflows/ci.yml` runs `bun run check`, `bun run build`
+and `bun run check:artifacts` on every PR.
+
+That last step exists because **a green `astro build` does not prove the site is
+correct**. The Fonts API fails *silently*: point a family's provider at a package it
+cannot resolve and the build still exits 0 with only a `[WARN]`, emitting a `dist/`
+where that family has degraded to system fonts. Measured against a real degraded
+build, every aggregate signal survives — `@font-face` blocks are still present, the
+`--font-*` variables are still declared, woff2 files and preload links are still
+emitted. The one check that discriminates is per-family: does the family each
+`--font-*` variable names actually have a `url()`-backed `@font-face`?
+
+`scripts/check-build-artifacts.mjs` asserts that, plus route coverage (derived from
+`src/pages/**`, so a new page is covered automatically) and font-file size floors.
+Its logic is unit-tested from the monorepo root in `test/check-build-artifacts.test.ts`,
+alongside the repo's other gate-script tests.
 
 ## Routing (`vercel.json`)
 
--- a/site/package.json
+++ b/site/package.json
@@ -7,7 +7,8 @@
     "dev": "astro dev",
     "build": "astro build",
     "preview": "astro preview",
-    "check": "astro check"
+    "check": "astro check",
+    "check:artifacts": "bun scripts/check-build-artifacts.mjs"
   },
   "dependencies": {
     "@fontsource/jetbrains-mono": "^5.2.8",
--- a/site/scripts/check-build-artifacts.mjs
+++ b/site/scripts/check-build-artifacts.mjs
@@ -1,0 +1,243 @@
+#!/usr/bin/env bun
+// Artifact assertions for the built marketing site (`site/dist`).
+//
+// WHY THIS EXISTS: a green `astro build` does not prove the site is correct. The
+// Fonts API's failure mode is *silent* — point a family's provider at a package it
+// cannot resolve and the build still exits 0 with only a [WARN], emitting a `dist/`
+// where that family has no webfont at all. Measured against a real degraded build,
+// every aggregate signal survives it (`@font-face` still present, the `--font-*`
+// variable still present, woff2 files still emitted, preload links still emitted).
+// The one signal that discriminates is per-family: does the family named by each
+// `--font-*` variable actually have a `url()`-backed `@font-face`? That is the
+// assertion this script exists for; the rest are cheap floors around it.
+//
+// Run after `astro build`: `bun run check:artifacts`.
+//
+// Roots default to this package (resolved from import.meta.url, never process.cwd(),
+// so running it from the repo root cannot silently derive an empty route list and
+// vacuously pass). Both are env-overridable so the gate can be exercised against
+// synthetic fixtures — see test/check-build-artifacts.test.ts.
+//
+//   SITE_DIST  (or argv[2])  build output to validate      default <pkg>/dist
+//   SITE_PAGES               route source of truth         default <pkg>/src/pages
+
+import { readdirSync, readFileSync, statSync } from "node:fs";
+import { join, relative, resolve, sep } from "node:path";
+import { fileURLToPath } from "node:url";
+
+const PKG_ROOT = resolve(fileURLToPath(import.meta.url), "..", "..");
+const DIST = resolve(process.argv[2] ?? process.env.SITE_DIST ?? join(PKG_ROOT, "dist"));
+const PAGES = resolve(process.env.SITE_PAGES ?? join(PKG_ROOT, "src", "pages"));
+
+/** CSS custom properties every built page must declare, one per configured font
+ *  family (see `fonts` in astro.config.mjs). Listed explicitly rather than derived
+ *  from the built CSS: deriving from the output would make the check vacuous if a
+ *  family were dropped from the config entirely. */
+const EXPECTED_FONT_VARS = ["--font-space-grotesk", "--font-jetbrains-mono"];
+
+/** Minimum size of an emitted route's HTML. The three real routes measure
+ *  28825 / 13862 / 12778 bytes, so this only catches an empty/stub write. */
+const MIN_ROUTE_BYTES = 2000;
+
+/** Minimum size of an emitted woff2 — catches a truncated font, which otherwise
+ *  passes both the presence floor and the per-family check (the file exists and
+ *  its `@font-face` still carries a `url()`). The smallest legitimately emitted
+ *  woff2 measures 1160 bytes, so 1 KB would leave only ~136 bytes of headroom and
+ *  would red the gate whenever subsetting re-chunks. 512 sits comfortably under
+ *  every real file and well over a stub write. */
+const MIN_WOFF2_BYTES = 512;
+
+let failed = false;
+
+function pass(msg) {
+  console.log(`PASS  ${msg}`);
+}
+
+function fail(msg) {
+  console.log(`FAIL  ${msg}`);
+  failed = true;
+}
+
+/** Every file under `dir`, recursively. Returns [] when `dir` does not exist. */
+function walk(dir) {
+  let entries;
+  try {
+    entries = readdirSync(dir, { withFileTypes: true });
+  } catch {
+    return [];
+  }
+  const out = [];
+  for (const entry of entries) {
+    const abs = join(dir, entry.name);
+    if (entry.isDirectory()) out.push(...walk(abs));
+    else out.push(abs);
+  }
+  return out;
+}
+
+/** Size of a regular file, or -1 if it is missing or is not a regular file.
+ *  The isFile() guard matters: statSync on a directory returns a nonzero size,
+ *  so a directory sitting where a route's index.html belongs would otherwise
+ *  sail past the size floor. */
+function sizeOf(file) {
+  try {
+    const st = statSync(file);
+    return st.isFile() ? st.size : -1;
+  } catch {
+    return -1;
+  }
+}
+
+// ── Routes ───────────────────────────────────────────────────────────────────
+
+/** Map a page file to the HTML Astro emits for it under the default
+ *  `build.format: "directory"`:
+ *
+ *    index.astro            → index.html
+ *    privacy.astro          → privacy/index.html
+ *    legal/terms.astro      → legal/terms/index.html
+ *    legal/index.astro      → legal/index.html
+ *
+ *  Returns "" for a file Astro deliberately does not route (see below), and null
+ *  for anything this mapping does not understand. The caller fails loudly on null
+ *  rather than skipping — silently ignoring an unmapped page is the same
+ *  under-coverage bug as not walking subdirectories at all. */
+function routeForPage(rel) {
+  const segments = rel.split(sep);
+  // Astro excludes any path with an underscore-prefixed segment from routing
+  // (_components/, _draft.astro), so these are not missing routes — asserting on
+  // them would be a false failure.
+  if (segments.some((s) => s.startsWith("_"))) return "";
+  if (!rel.endsWith(".astro")) return null;
+  segments[segments.length - 1] = segments[segments.length - 1].slice(0, -".astro".length);
+  // Dynamic and spread routes ([slug].astro, [...rest].astro) expand via
+  // getStaticPaths at build time, so their outputs cannot be derived from the
+  // filename alone.
+  if (segments.some((s) => s.includes("[") || s.includes("]"))) return null;
+  if (segments[segments.length - 1] === "index") {
+    return join(...segments.slice(0, -1), "index.html");
+  }
+  return join(...segments, "index.html");
+}
+
+function checkRoutes() {
+  const routes = [];
+  for (const abs of walk(PAGES)) {
+    const rel = relative(PAGES, abs);
+    const route = routeForPage(rel);
+    if (route === null) {
+      fail(`page ${rel} has no known output mapping (only static .astro pages are supported)`);
+      continue;
+    }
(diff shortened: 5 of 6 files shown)


scoop (commented):
## What I verified

**The `site` job is a faithful mirror of `docs-site`.** Compared line-by-line against the existing job:

- release-please skip guard at `.github/workflows/ci.yml:248` is byte-identical to `docs-site`'s at `.github/workflows/ci.yml:187` (branch prefix **and** bot identity — the fork-spoof hardening described at `ci.yml:33-38` is preserved).
- `actions/checkout@v7` (`ci.yml:251` vs `ci.yml:186`), `oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0` (`ci.yml:254` vs `ci.yml:190`), `actions/cache@v6` (`ci.yml:261` vs `ci.yml:197`) — all match the pinned forms already in the file.
- Cache key `${{ runner.os }}-bun-site-${{ hashFiles('site/bun.lock') }}` with prefix restore-key mirrors `ci.yml:199-203`; `site/bun.lock` exists.
- `uses: ./.github/actions/bun-install-retry` with `working-directory: site` — the composite action exists and declares exactly that input (`.github/actions/bun-install-retry/action.yml:11-14`, default `.`), and runs `bun install --frozen-lockfile` (`action.yml:26`). Verified `runs-on: ubuntu-latest`, no `paths:` filter, no change detector, no font cache, no build retry, no test step — matching the plan's success criterion 1.
- Step body is `check` → `build` → `check:artifacts` in one `run:` block; GitHub run steps use `bash -e`, so a non-zero `check:artifacts` reds the job.

**Identifiers resolve.** `check:artifacts` is defined at `site/package.json:11` and points at `scripts/check-build-artifacts.mjs`, which exists. `EXPECTED_FONT_VARS` (`site/scripts/check-build-artifacts.mjs:36`) is `["--font-space-grotesk", "--font-jetbrains-mono"]`, matching the two `cssVariable` values in `site/astro.config.mjs:18` and `site/astro.config.mjs:27` — no third family is configured, so the explicit list is complete today.

**The per-route font assertion is reachable on every route.** All three pages import the shared layout (`site/src/pages/index.astro:2`, `site/src/pages/privacy.astro:12`, `site/src/pages/impressum.astro:7`), and that layout renders both `<Font cssVariable=... />` tags (`site/src/layouts/Base.astro:56-57`), so requiring both `--font-*` declarations in *each* emitted HTML is consistent with the source.

**Path resolution.** `PKG_ROOT = resolve(fileURLToPath(import.meta.url), "..", "..")` (`check-build-artifacts.mjs:28`) resolves `site/scripts/check-build-artifacts.mjs` → `site/`, so the `dist` / `src/pages` defaults at lines 29-30 are package-relative and cwd-independent as required; `process.cwd()` appears nowhere in the file.

**Route mapping logic** (`check-build-artifacts.mjs:105-122`) — traced by hand: `index.astro` → `index.html`; `privacy.astro` → `privacy/index.html`; `legal/terms.astro` → `legal/terms/index.html`; `legal/index.astro` → `legal/index.html`. Underscore-prefixed segments return `""` (correctly not asserted on — Astro excludes them from routing), `[slug].astro` and non-`.astro` return `null` and fail loudly, and an empty derived list fails rather than passing vacuously (`:136-139`). `sizeOf` (`:88-95`) returns `-1` for a directory via the `isFile()` guard, so a directory standing in for `index.html` cannot clear the byte floor — and `test/check-build-artifacts.test.ts:225-232` covers exactly that.

**The discriminating assertion is the one that matters and it is right.** `familiesWithWebfont` is built only from faces whose `src` contains `url(` (`:186-188`), then each `--font-*` declaration's primary family token is looked up in that set (`:193-206`). A `local()`-only metric fallback therefore cannot satisfy it, which is the measured silent-degradation shape. `test/check-build-artifacts.test.ts:106-116` pins it, and the asserted substring `"no url()-backed @font-face"` does occur in the script's failure string at `check-build-artifacts.mjs:202-203`. I checked the other tests' asserted substrings against the script's message strings too: `"was not emitted"` (`:144`), `"is only"` (`:145`, `:225`), `"is not declared"` (`:196`), `"no .woff2 files were emitted"` (`:216`), `"no known output mapping"` (`:129`), `"no pages found"` (`:137`), `"All build artifact checks passed"` (`:242`) — all present and spelled consistently.

**Test placement and toolchain.** `test/check-build-artifacts.test.ts:1-9` follows `test/check-feature-catalog.test.ts:1-43` exactly (bun:test, `mkdtempSync` fixture, `spawnSync` the gate as a subprocess, `{ code: r.status ?? -1, out: stdout+stderr }`), so it rides `verify`'s existing `Test (root)` step (`ci.yml:129-131`, `bun test ./test`). No new dependency is added. `site/` remains excluded from root `tsconfig.json:16` and listed in `.prettierignore:5`, so the new `.mjs` is not pulled into the server typecheck or root prettier; the new root test is a plain `.ts` in an already-covered directory.

**Docs.** `docs-site/README.md:31-33` previously asserted `site/` is **not** wired into CI; the diff correctly rewrites that sentence, and it is the only place in the repo's markdown that made the claim (grepped `*.md` for `docs-site` — the only hit outside `docs-site/` itself was that file).

## Stated limitations (not findings)

- `site/node_modules` and `site/dist` are absent in this worktree, so I could **not** re-measure a real Astro 7 build. Two claims therefore rest on the author's stated measurement rather than on ground truth I opened: (a) that the `<Font/>` CSS is inlined into each page's HTML rather than extracted to a linked stylesheet, and (b) that the first family token of the `--font-*` declaration is spelled identically to the `font-family` of the emitted `@font-face` (the hashed name). Both are load-bearing for a green run; if either is off the job reds on its first run, which is a loud failure rather than a silent one.
- Whether the degraded-build probe in the plan generalises to "the Fonts API changes across an Astro major" is reasoning, not measurement — the plan says so itself, and I have no way to test it here.

- `site/scripts/check-build-artifacts.mjs`: the `argv[2]` override at line 29 is a second, undocumented-in-`package.json` way to set the dist root that nothing in the repo uses (the test drives it via `SITE_DIST`) — harmless, but `SITE_DIST` alone would do.

- `site/scripts/check-build-artifacts.mjs:36`: if a third family is ever added to `astro.config.mjs`, the gate silently will not check it — dormant today because the list at :36 matches the config exactly.

scoop:
All four findings accepted and fixed in c659b2880. One deviates from the suggested mechanism, so flagging it for you to weigh.

**Finding 2 (`EXPECTED_FONT_VARS` drift) — agreed, but implemented differently.**

The suggestion was to fail when the built CSS declares a `--font-*` custom property outside the expected list. I checked what the site actually declares before implementing that, and it would have been a latent false-red: `site/src/styles/global.css:33-34` layers its own semantic aliases on top of the Astro-injected variables:

```css
--font-display: var(--font-space-grotesk), system-ui, sans-serif;
--font-mono: var(--font-jetbrains-mono), ui-monospace, monospace;
```

Those live in an external stylesheet today, so an output scan passes right now — but Astro inlines stylesheets under a size threshold, so the rule would red the moment that file crosses it or the CSS is restructured. That is a failure mode nobody would connect back to this gate.

So I closed the drift at its source instead: the expected list is now read from `astro.config.mjs`'s `cssVariable:` entries (`SITE_CONFIG`, overridable for fixtures). Adding a family to the config brings it under the gate with no script change — which is the outcome the finding asked for — and it fails loudly if the config is unreadable or declares no families, so it cannot degrade into checking nothing.

Four tests cover it: a family added to the config is immediately required; one removed is no longer required; an unreadable config fails; a config with zero `cssVariable` entries fails.

The other three (timing claim, README overstatement re `.md`/`[slug]` hard-failing by design, dot-prefixed skip) are fixed as described, each with a test where testable.

scoop (commented):
## Re-review of the four prior points

1. **`.github/workflows/ci.yml` — "sub-second" timing claim.** Addressed. The no-`paths:`-filter block now ends at `.github/workflows/ci.yml:239-244` with "the job is small next to `verify`, and correctness of the required-check contract is the reason, not cost" — no timing number remains. The adjacent block (ci.yml:240-245 in the new text) still documents the build-time CDN fetch, and the two no longer contradict each other.

2. **`site/scripts/check-build-artifacts.mjs` — hardcoded `EXPECTED_FONT_VARS` drift.** Addressed, and verified end to end. The constant is gone; `expectedFontVars()` (check-build-artifacts.mjs:165-179) reads `SITE_CONFIG` (default `<pkg>/astro.config.mjs`, line 32) and extracts `/cssVariable:\s*["']([^"']+)["']/g`. I compared that regex against the real config: it matches `site/astro.config.mjs:18` (`cssVariable: "--font-space-grotesk"`) and `site/astro.config.mjs:28` (`cssVariable: "--font-jetbrains-mono"`) — the only two `cssVariable` occurrences in that file. An unreadable config and a zero-entry config both hard-fail (lines 170, 175) rather than vacuously checking nothing, and both are covered by `test/check-build-artifacts.test.ts:260-274`. The added-family case is covered at test/check-build-artifacts.test.ts:245-251. The rejected alternative (deriving from built CSS) is correctly rejected: `site/src/styles/global.css:33-34` layers `--font-display` / `--font-mono` aliases that an "unexpected `--font-*`" rule would red on. I also confirmed those alias lines cannot false-match the per-var lookup at check-build-artifacts.mjs:222 — `var(--font-space-grotesk),` has no `:` after the name, so the regex `--font-space-grotesk\s*:` does not match there, and the only true declaration is Astro's injected one.

3. **`site/README.md` — "covered automatically" overstated.** Addressed. site/README.md:41-51 now says static `.astro` only, lists Markdown/MDX and dynamic/spread routes as **"fails the gate by design"**, and tells the reader to extend `routeForPage()`. That matches the script: check-build-artifacts.mjs:109 returns `null` for non-`.astro`, :114 returns `null` for `[`/`]` segments, and :127 fails loudly. Both cases are pinned by tests (test/check-build-artifacts.test.ts:196-210).

4. **Dot-prefixed junk under `src/pages`.** Addressed. check-build-artifacts.mjs:108 now returns `""` for any segment starting with `_` *or* `.`, with the reasoning in the comment at :105-107; test/check-build-artifacts.test.ts:236-240 covers `.DS_Store`, and test:228-234 confirms a pages tree of only non-routed files still fails as vacuous rather than passing.

## Fresh verification

- **Workflow shape.** The `site` job (ci.yml:246-285) mirrors `docs-site` (ci.yml:183-225) exactly where it should: same release-please branch+bot guard (compared string-for-string against ci.yml:189), `actions/checkout@v7`, `oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6`, `actions/cache@v6` keyed on `hashFiles('site/bun.lock')` (that lockfile exists — `site/bun.lock`). No `paths:` filter, no change detector, no font cache, no build retry, no test step — matching success criterion 1.
- **Composite action contract.** `with: working-directory: site` matches the only declared input in `.github/actions/bun-install-retry/action.yml:12-16` (`working-directory`, default `"."`). Spelling and casing agree.
- **npm script.** `check:artifacts` → `bun scripts/check-build-artifacts.mjs` (site/package.json:11) resolves to the file added in this diff; `check` → `astro check` is backed by `@astrojs/check` in devDependencies (site/package.json:18). `bun scripts/…` passes no argv, so `process.argv[2]` at check-build-artifacts.mjs:30 is `undefined` and the package-relative default applies.
- **Test seam.** `test/check-build-artifacts.test.ts` follows `test/check-feature-catalog.test.ts:1-43` closely (mkdtemp fixture, `spawnSync` the gate, `{code, out}` assertions, env parameterisation). Root `package.json:9` runs `bun test ./test`, so the new file is picked up with no CI change. `site` is in `.prettierignore` but `test/` is not, so the new test is covered by the existing root format/lint gates.
- **Route mapping edge cases** traced by hand: `index.astro` → `segments=['index']` → `join(...[], 'index.html')` = `index.html`; `legal/index.astro` → `legal/index.html`; `legal/terms.astro` → `legal/terms/index.html` — all three match the doc comment at check-build-artifacts.mjs:89-96 and the tests at test:178-194.
- **`sizeOf` isFile() guard** (check-build-artifacts.mjs:78-85) is genuinely load-bearing and is pinned by test:278-285 (directory in place of `index.html`).
- **Doc consistency.** `docs-site/README.md:31` no longer claims `site/` is unwired; I grepped every tracked `.md` for `site/` + CI/workflow mentions and this was the only stale claim — it is fixed in this diff.

### Stated limitations

- I did not build the site (read-only review), so the claim that Astro inlines the `@font-face` blocks and the `--font-*` declarations into each route's HTML — which the whole font check depends on, since `checkFontsForRoute` reads only the route HTML (check-build-artifacts.mjs:204-211) — is taken from the plan's measurements rather than re-measured here. Note the failure mode if it were ever wrong is a loud red ("variable is not declared"), not a silent green.
- The degraded-build behaviour of Astro's npm font provider (exit 0 + `[WARN]`) is external to this repo and is assumed as documented.

- check-build-artifacts.mjs:222 builds `new RegExp(varName + "\\s*:\\s*([^;}]+)")` from a config-derived string with no escaping; harmless for today's `--font-*` idents, but a CSS custom property containing a regex metacharacter (CSS allows escaped idents) would silently mismatch or throw.

- Nothing outside the plan's file list; the diff is if anything smaller than the issue proposed (preload and `url()`-on-disk assertions explicitly deferred).

## Nits

Both below are non-blocking; the decision is `comment`.

scoop (commented):

**1. Unescaped `cssVariable` interpolated into `new RegExp(...)` — ADDRESSED.** `site/scripts/check-build-artifacts.mjs:212-214` adds `escapeRegExp()` (escaping `.*+?^${}()|[]\`), and the per-variable lookup now reads `new RegExp(`${escapeRegExp(varName)}\\s*:\\s*([^;}]+)`)` at `site/scripts/check-build-artifacts.mjs:239`. Both failure modes are covered by tests that are genuinely load-bearing: `test/check-build-artifacts.test.ts:263-270` (unbalanced paren — asserts the run reports `--font-a(b is not declared` and does **not** print `Invalid regular expression`) and `test/check-build-artifacts.test.ts:272-283`, where a decoy `--font-a-c` declaration resolving to a family with no `url()`-backed face is written *before* the real `--font-a.c` (`test/check-build-artifacts.test.ts:278-281`) — I traced it: with an unescaped `.` the regex matches the decoy, `primaryFamily` yields the JetBrains family, and since `webfontFamilies: [SG]` that family is absent from `familiesWithWebfont` (`site/scripts/check-build-artifacts.mjs:234-236`), so the test would red. Correctly discriminating.

**2. Every configured family required on every route — ADDRESSED (documented).** The assumption is now stated in three places: the script header at `site/scripts/check-build-artifacts.mjs:16-19`, the doc comment directly on `checkFontsForRoute` at `site/scripts/check-build-artifacts.mjs:216-220`, and `site/README.md:38-43`. All three name the same cause and the same remedy. I checked the ground truth the comments assert: `site/src/layouts/Base.astro:58-59` renders `<Font cssVariable="--font-space-grotesk" …/>` and `<Font cssVariable="--font-jetbrains-mono" …/>`, matching the two `cssVariable:` entries in `site/astro.config.mjs` (`--font-space-grotesk`, `--font-jetbrains-mono`), and all three pages under `site/src/pages/` (`index.astro`, `privacy.astro`, `impressum.astro`) are the full route set the gate derives.

## Author note (config-derived font list)

Judged against the diff, not the note. `expectedFontVars()` (`site/scripts/check-build-artifacts.mjs:170-184`) reads `cssVariable:` entries from `SITE_CONFIG` (default `<pkg>/astro.config.mjs`, `site/scripts/check-build-artifacts.mjs:37`) and fails loudly on both degenerate cases — unreadable config (`:175`) and zero entries (`:180`) — so it cannot degrade into checking nothing. Four tests cover exactly the four cases claimed: `test/check-build-artifacts.test.ts:245`, `:253`, `:285`, `:293`. The stated reason for not scanning built CSS checks out: `site/src/styles/global.css:33-34` does declare `--font-display: var(--font-space-grotesk), …` and `--font-mono: var(--font-jetbrains-mono), …`, which an "unexpected `--font-*` in output" rule would have flagged once inlined. Note the lookup regex requires `<name>\s*:`, so `var(--font-space-grotesk)` inside those aliases cannot be mistaken for a declaration — verified by reading the regex at `:239` against the alias text.

## Other correctness checks performed

- **Route mapping** (`site/scripts/check-build-artifacts.mjs:106-124`): `index.astro → index.html`, `<name>.astro → <name>/index.html`, nested and nested-index handled; `null` (loud fail) for non-`.astro` and for `[`/`]` routes; `""` (skip) for `_`/`.`-prefixed segments, checked *before* the extension test so `_draft.md` skips rather than reds. Empty derived list fails at `:140-143`. Exercised by `test/check-build-artifacts.test.ts:178-240`.
- **Directory-in-place-of-file**: `sizeOf()` guards with `st.isFile()` (`site/scripts/check-build-artifacts.mjs:83-90`), tested at `test/check-build-artifacts.test.ts:303-310`.
- **Script dependencies resolve**: only `node:fs`, `node:path`, `node:url` are imported (`site/scripts/check-build-artifacts.mjs:30-32`) — so the root `verify` job can spawn it with no `site/node_modules` present, which matters because `bun test ./test` (`.github/workflows/ci.yml:138`) runs before/without any `site` install. `spawnSync("bun", [SCRIPT], …)` at `test/check-build-artifacts.test.ts:96` has `bun` on PATH via the `verify` job's `setup-bun` step (`.github/workflows/ci.yml:53-56`).
- **npm script wiring**: `check:artifacts` added at `site/package.json:11` as `bun scripts/check-build-artifacts.mjs`, and the file exists at that path; `check`/`build` at `site/package.json:9-10` are unchanged, so `.github/workflows/ci.yml:283-285` invokes three real scripts.
- **Workflow shape**: the `site` job (`.github/workflows/ci.yml:246-285`) mirrors `docs-site` (`.github/workflows/ci.yml:183-227`) exactly — same `if:` branch+bot guard (identical string, `:249` vs `:188`), same pinned `oven-sh/setup-bun@0c5077e…`, `actions/checkout@v7`, `actions/cache@v6`, cache key namespaced `-bun-site-` on `hashFiles('site/bun.lock')` (`site/bun.lock` exists), and install via `./.github/actions/bun-install-retry` with `working-directory: site`, which is a declared input of that composite action (`.github/actions/bun-install-retry/action.yml:11-15`). No `paths:` filter, no change detector, no font cache, no build retry, no test step — matching the plan's success criterion 1. On `push: main` the guard evaluates false (empty `head_ref`), so the job runs there too, same as the existing jobs.
- **Lint/format exposure**: `site` is prettier-ignored (`.prettierignore:6`) and is not in the `lint` target list (`package.json:10`), so the new `.mjs` is not linted/formatted by root gates; the new root test lives under `test/`, which both cover, and follows the established shape of `test/check-feature-catalog.test.ts:36-42` (`spawnSync` + `{code, out}` + env parameterisation).
- **docs-site/README.md:31**: the previous text claimed `site/` was *not* wired into CI; the replacement ("Both are wired into CI … each in its own job") is now accurate given `.github/workflows/ci.yml:246`.

## Stated limitations (could not verify — not a finding)

- `site/node_modules` and `site/dist` are absent in this worktree, so I could not confirm empirically that Astro 7 inlines the `@font-face` block and the `:root{--font-*}` declaration **into each route's HTML** rather than emitting them as an external stylesheet. The whole per-route check (`site/scripts/check-build-artifacts.mjs:221-254`) depends on that. If it ever becomes external, the failure mode is a loud false red on this PR's own `site` job, not a silent pass — so it is self-revealing rather than dangerous. Assumed as measured by the author.
- The degraded-build measurements quoted in the comments (3-vs-9 woff2, smallest legitimate woff2 = 1160 B) come from a scratchpad build I cannot reproduce read-only; I verified only that the constants in the code (`MIN_WOFF2_BYTES = 512`, `MIN_ROUTE_BYTES = 2000`) are consistent with those stated numbers.

- `.github/workflows/release-please.yml:28` still enumerates the branch ruleset's required checks as "`verify`/`docs-site`"; it stays accurate until an operator promotes `site` to required, but will need a touch-up at that point.

- `site/README.md` grows a ~35-line "CI" section that restates much of the script's header comment; justified as the discoverable home for the gate's rationale, but it is duplicated prose that can drift from `site/scripts/check-build-artifacts.mjs:4-28`.

scoop (commented):

**Prior point 1 — `expectedFontVars()` matched `cssVariable:` in commented-out config: ADDRESSED.**
`site/scripts/check-build-artifacts.mjs:225` now runs the config source through `stripComments()` (`site/scripts/check-build-artifacts.mjs:165-200`) before the `cssVariable` scan. I traced the stripper by hand:
- `//` line comments consumed to `\n` (`:186-189`), `/* … */` block comments consumed to the terminator (`:190-195`);
- string literals (`"`/`'`/`` ` ``) are copied verbatim with backslash-escape handling (`:170-185`), so the `//` inside `"https://example.com/f.css"` does **not** truncate the line — the failure mode the naive fix would have introduced (silent under-check) is avoided;
- unterminated string/comment degrade to "no cssVariable entries found", which is a loud `fail()` at `:228-231`, not a vacuous pass.
Three committed fixtures cover exactly these cases: `test/check-build-artifacts.test.ts:287` (line comment), `:303` (block comment), `:322` (`//` inside a string, asserting the entry sharing the line still counts — `expect(out).toContain("declares 3 family variable(s)")` at `:340`).

## Verification of the rest

**CI job parity.** The new `site` job (`.github/workflows/ci.yml:243-285`) is a line-for-line mirror of `docs-site` (`.github/workflows/ci.yml:183-227`): same release-please branch+bot guard, `actions/checkout@v7`, `oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6`, `actions/cache@v6` on `hashFiles('site/bun.lock')` (file exists — `site/bun.lock`), and `uses: ./.github/actions/bun-install-retry` with `working-directory: site`, which is a declared input of that composite action (`.github/actions/bun-install-retry/action.yml:12-16`). No `paths:` filter, no change detector, no font cache, no build retry, no test step — matching success criterion 1. Workflow triggers (`.github/workflows/ci.yml:4-10`) mean `github.event.pull_request.user.login` is null on push-to-main, so the guard evaluates false and the job runs — same as the two existing jobs.

**Script identifiers resolve.** `check:artifacts` is defined at `site/package.json:11` as `bun scripts/check-build-artifacts.mjs`, and the file exists at that path; the CI step invokes `bun run check:artifacts` (`.github/workflows/ci.yml:285`) and the test resolves the same path via `join(import.meta.dir, "..", "site", "scripts", "check-build-artifacts.mjs")` (`test/check-build-artifacts.test.ts:10`). `SITE_DIST` / `SITE_PAGES` / `SITE_CONFIG` are the only env names read (`site/scripts/check-build-artifacts.mjs:35-37`) and are exactly the three the test sets (`test/check-build-artifacts.test.ts:97`) — no spelling drift.

**Discriminating assertion is real.** `checkFontsForRoute()` builds `familiesWithWebfont` from faces whose `src` contains `url(` (`site/scripts/check-build-artifacts.mjs:283-285`), then resolves each configured `--font-*` to its primary family and requires membership (`:287-302`). Against the actual config (`site/astro.config.mjs:18,28` — `--font-space-grotesk`, `--font-jetbrains-mono`) and the actual layout (`site/src/layouts/Base.astro:57-58`, one `<Font>` per family in the shared head), the script's "every family on every route" assumption holds for the three pages in `site/src/pages/`.

**Route mapping.** `routeForPage()` (`:106-124`) maps `index.astro`→`index.html` and `<name>.astro`→`<name>/index.html`; for `segments = ["index"]`, `join(...[], "index.html")` reduces to `join("index.html")` = `"index.html"` (no zero-arg `join()` is reachable). Unmapped extensions and `[param]` routes return `null` and fail loudly (`:131-133`); an empty derived list fails (`:140-143`) rather than passing vacuously. `sizeOf()` guards with `isFile()` (`:83-90`) so a directory named `index.html` cannot clear the size floor — covered by `test/check-build-artifacts.test.ts:362`.

**Escaping.** `escapeRegExp()` (`:261-263`) escapes the RegExp metacharacter set before interpolation at `:288`; both the throw case and the wrong-match case are pinned by tests (`test/check-build-artifacts.test.ts:263`, `:272`).

**Toolchain fit.** `site/` is in `.prettierignore` and root `tsconfig.json:17` `exclude`, and root `lint` targets `src test examples ui/src ci/onboarding-harness deploy 'scripts/**/*.ts'` (`package.json:10`) — so the new `.mjs` is outside root prettier/eslint/tsc, while `test/check-build-artifacts.test.ts` sits in the already-linted `test/` tree and follows the established gate-test shape (`test/check-generated-docs.test.ts:1-7`: same imports, same `import.meta.dir` script resolution, same `spawnSync` + `{code, out}` assertions). `site/tsconfig.json` extends `astro/tsconfigs/strict` with `include: ["**/*"]`, but the new script carries no `// @ts-check` (unlike `site/astro.config.mjs:1`), so `astro check` will not type-check it.

**Docs.** `docs-site/README.md:31-33` previously claimed `site/` was *not* wired into CI; the diff corrects it to "Both are wired into CI … each in its own job", which is now true.

### Stated limitations (could not verify)
- I did not run `astro build`, so the claim that Astro 7 inlines the `@font-face` blocks and the `--font-*` declarations into each route's HTML (rather than an external stylesheet), and the measured constants (28825/13862/12778-byte routes, 1160-byte smallest woff2) are taken as written. If that shape is wrong, the gate false-reds loudly on the first CI run rather than passing silently, so the failure mode is safe.

- `site/scripts/check-build-artifacts.mjs:288`: the declaration regex is unanchored, so a configured `cssVariable` that is a *suffix* of another declared custom property would match the wrong declaration and could report a false PASS. Not reachable today (the configured names are `--font-space-grotesk` / `--font-jetbrains-mono`, neither a suffix of the `--font-display` / `--font-mono` aliases at `site/src/styles/global.css:33-34`), but the script is explicitly designed so that adding a family to `site/astro.config.mjs` brings it under the gate with no script edit — e.g. a family named `--font-mono` would match inside `--font-jetbrains-mono:`. A `(?<![\w-])` prefix guard would close it.

- `site/README.md`: the new `## CI` section is substantially longer than the plan's "add a CI-gate line", though it documents genuinely non-obvious gate behaviour (why `.md`/dynamic routes fail by design).

Issue #1875, opened 2026-07-21
Point spawned agents at a disk-backed TMPDIR (structural fix for tmpfs inode exhaustion)


scoop:
Follow-up to #1862. That issue's PR ships an **advisory** system-prompt directive telling agents to keep git worktrees and dependency installs off the tmpfs. This issue replaces that with a **structural** guarantee: point spawned agents at a disk-backed `TMPDIR`, so the condition becomes unreachable rather than merely discouraged.

## Why this is the strongest remaining fix

While planning #1862 a sub-agent was dispatched and asked to report its own system prompt. The result:

| Block | Present in a sub-agent's prompt? |
| --- | --- |
| `<engineering-posture>` | **no** |
| `<shepherd-house-rules>` | **no** |
| `<worktree-stash-notice>` | **no** |
| `<single-pr-invariant>` | **no** |
| Claude Code's "Scratchpad Directory" section ("for ALL temporary file needs") | **YES** |

A sub-agent receives the instruction that steers work *onto* the tmpfs and **none** of Shepherd's counterbalancing blocks — `composeSystemPrompt`'s output reaches the top-level agent only. The shipped mitigation works around this with a "restate this constraint in the sub-agent's prompt" clause, which depends on the top-level agent both honouring it and correctly predicting which sub-agent tasks need it.

`TMPDIR` needs neither. It is the only approach that does not depend on an agent — or a sub-agent that never sees our prompt at all — reading a directive.

## The blocking premise is already verified

```
$ mkdir /tmp/tmpdir-probe
$ TMPDIR=/tmp/tmpdir-probe claude -p "Reply with exactly: OK"   → OK
$ find /tmp/tmpdir-probe
/tmp/tmpdir-probe/claude-1000          ← created under the OVERRIDDEN root
$ find /tmp/claude-1000 -maxdepth 1 -newermt '-3 minutes'   → (nothing)
```

**Claude Code honours a spawn-side `TMPDIR`**, and derives exactly the `claude-$uid` shape `claudeTmpRoot()` already models. Nothing landed in the default root. So this opens as viable and prioritized, not speculative.

The seam already exists: `buildWrappedArgv` (`src/herdr.ts:369`) wraps every spawned agent in a coreutils `env` shim — the same one #560 used to move `NODE_COMPILE_CACHE` off the tmpfs. Adding `TMPDIR=` there is a one-line change to that shim.

## A previous declination was wrong — do not repeat it

[...]

Issue #1874, opened 2026-07-21
Reclaim forked pnpm stores and abandoned agent worktrees from the temp filesystem


scoop:
Follow-up to #1862. That issue's PR ships **prevention** (a system-prompt directive) and **diagnosis** (the `tmp_inodes` Diagnose row) only — neither reclaims what has already accumulated, so the row can legitimately sit at `warning`/`error` after a *successful* forced sweep. This closes that gap.

Two reclaimers landing **together**, because the worktree reaper is what makes the store reclaimer fire at all in #1862's own scenario.

## Why together

pnpm hardlinks `node_modules` into its store, and an inode frees only on the **last** link drop. While a tmp worktree survives, every store file sits at `nlink = 2` — so removing the store reclaims **~zero** inodes *and* the next install re-fetches, creating a second set of inodes for content still alive in the old `node_modules`. Net inode use goes **up**. The reaper must run first, and the store pass must skip whenever anything still links in.

## Verified facts (measured while planning #1862, not assumed)

- The forked store is **versioned**: `/tmp/.pnpm-store/`**`v10`**`/files/<bucket>/<hash>`, 256 buckets. `readdir(<root>/files)` finds nothing — an earlier design would have fail-closed forever as a silent no-op.
- Depth 2 from the *unversioned* root is only 5 entries, whose mtimes stop moving once buckets exist — a mid-install store would read as idle.
- A real populated store at its versioned root: depth 1 = 4, depth 2 = **556**, depth 3 = **57,327**. Idleness must probe at depth 2.
- `nlink` is **2** while a `node_modules` links a store file, and drops to **1** the instant that tree is removed.

## Worktree reaper

- Discover via the **existing** `parseWorktrees` (`src/forge/local.ts:176`, module-private today), extended with `locked`, `bare`, `prunable`. Do not add a third porcelain parser — `src/doc-agent.ts:1444` is a pre-existing second.
- **Git floor.** Those annotations arrived in **git 2.36**; their silent absence makes a locked worktree reapable while `git worktree prune --expire=now` cannot drop its record, leaving a dangling entry. Reuse the repo's existing `MIN_GIT_MAJOR`/`MIN_GIT_MINOR` = 2.38 + `gitVersionAtLeast` (`src/forge/local.ts:30`, `:74`) and refuse to reap below it.
[...]

Issue #1862, opened 2026-07-21
Agent worktrees under /tmp exhaust tmpfs inodes via forked pnpm store — session bricks with misleading ENOSPC


scoop:
## Symptom

Mid-session, every write to `/tmp` fails with `ENOSPC` — including bash's own cwd files and agent tool-output files:

```
/usr/bin/bash: line 2: /tmp/claude-2949-cwd: No space left on device
ENOSPC: ... open '/tmp/claude-1000/.../tasks/bwfk2d50n.output'
```

The session is effectively bricked: the agent can't run commands or receive tool output.

`/tmp` is **not** out of bytes — it's out of **inodes**:

```
$ df -h /tmp
tmpfs  32G  15G  17G  46% /tmp      # plenty of space

$ df -i /tmp
tmpfs  1048576  1048576  0  100% /tmp   # zero inodes free
```

The tmpfs is capped at 1,048,576 inodes regardless of the 32G size.

## Root cause

An agent placed a git worktree under `/tmp` (the session scratchpad) and ran `pnpm install` in it.

pnpm requires its content-addressable store to be on the **same filesystem** as the install target so it can hardlink. The configured store (`~/.local/share/pnpm/store/v10`, on `/home`) is on a different filesystem, so pnpm silently forked a **second store at `/tmp/.pnpm-store`** and began populating it. A monorepo store is hundreds of thousands of tiny files — it exhausted the inode table.

Measured on one host after cleanup:

| Consumer | Inodes |
|---|---|
| `/tmp/.pnpm-store` (forked store, 379M) | 44,018 |
| 14 `fallow-audit-base-cache-*` worktrees + 133 lock/meta files | 21,074 |

Reclaiming both took `/tmp` from 100% → 89% inodes. Note the fallow caches are a **second** independent source of `/tmp` worktrees, so the pressure is cumulative across tools.

## Why the obvious fix doesn't work

Setting `store-dir` in `~/.npmrc` to a `/home` path does **not** fix it. pnpm still needs same-filesystem hardlinking, so an install running from a `/tmp` worktree will fall back to a `/tmp` store anyway. Raising `nr_inodes` on the tmpfs only moves the ceiling.

The durable fix is **keeping install-bearing worktrees off tmpfs**.

## Suggested directions

1. Stage worktrees that may run package installs on the same filesystem as the repo (e.g. a `.shepherd-worktrees/` sibling) rather than under `/tmp`.
[...]

PR #1876, opened 2026-07-21
fix(tmp): warn on tmpfs inode pressure, steer agents off /tmp

scoop:
Agents put git worktrees and dependency installs in the `/tmp` scratchpad. pnpm then forks a **second** content-addressable store there — it must be same-filesystem to hardlink — and that store is hundreds of thousands of tiny files, which exhausts the tmpfs **inode** table. Every write then fails `ENOSPC` while `df -h` shows the volume mostly empty, so it misreads as "disk full" and the session bricks.

This PR ships **prevention** and **diagnosis**. Reclamation and the structural fix are deferred, with full designs — see below.

## What's here

**1. A `<tmpfs-worktree-notice>` on every spawn.** Two prohibitions (git worktrees, dependency installs) plus the reason, because the failure is easy to misdiagnose.

- **Worded as a carve-out, not a rival absolute.** Claude Code's harness block says to use the scratchpad "for ALL temporary file needs". A standalone "never use /tmp" is a flat contradiction, and an agent resolving two conflicting absolutes picks one arbitrarily — so this *names* that instruction and narrows it.
- **Provider-varied.** Codex never receives that block, so referencing it would name something the agent never saw. Wording only — never a suppression; a host-filesystem safety rule rides every spawn on every provider.
- **Carries a propagate-to-sub-agents clause.** This is the load-bearing part (see below).

**2. A `tmp_inodes` Diagnose row** with a one-click forced sweep.

- Warns at `SHEPHERD_TMP_INODE_PCT` — the *same* knob that gates the sweep, so raising it moves both and the docs stay true. Errors at 95%.
- `statfs`'s `tmpdir()`, **not** `claudeTmpRoot()`: the latter is `<tmpdir>/claude-$uid`, absent on a freshly booted host, which would report "uninspectable" on exactly the hosts that still have headroom worth protecting.
- btrfs (`files: 0`, dynamic inode allocation) reports `optional` — never a bogus percentage.
- Copy says "the temporary filesystem", never a hardcoded `/tmp`: `os.tmpdir()` honours the server's own `TMPDIR`. No absolute path crosses the check payload.

**3. A latent bug the row would otherwise have hit.** `sweepClaudeTmp` returned on **both** `inodeUsePct` failure reasons *before* comparing the threshold, so the forced sweep would have silently done nothing on a btrfs `/tmp` or a fresh boot. `thresholdPct <= 0` now bypasses the gate entirely; every threshold >= 1 keeps today's fail-open path byte-for-byte.

## The finding that shaped this

A sub-agent was dispatched and asked to report its own system prompt:

| Block | Present in a sub-agent's prompt? |
| --- | --- |
| `<engineering-posture>` | **no** |
| `<shepherd-house-rules>` | **no** |
| `<worktree-stash-notice>` | **no** |
| `<single-pr-invariant>` | **no** |
| Claude Code's "Scratchpad Directory" section | **YES** |

Sub-agents get the instruction that steers work *onto* the tmpfs and **none** of Shepherd's counterbalancing blocks. Since the issue's actor is plausibly a sub-agent, a notice riding only the top-level prompt would have missed it — hence the propagate clause. It is a mitigation, not a guarantee: it depends on the top-level agent restating the constraint. Removing that dependency is #1875.

Two other assumptions were checked rather than reasoned about, and both changed the design: the forked store is **versioned** (`/tmp/.pnpm-store/v10/files/…`, so a planned `readdir(<root>/files)` would have been a permanent silent no-op), and `nlink` is 2 while a `node_modules` links a store file, dropping to 1 the instant that tree is removed. Both are recorded in #1874.

## Honest limits

- The directive is **advisory**.
- The Doctor row **prevents nothing**. Its fix drops the compile cache and stale tool caches — it does **not** reclaim a forked pnpm store or an abandoned worktree, so the row can legitimately stay non-OK after a *successful* sweep. The unresolved-toast copy says exactly that instead of the generic "Claude Code may need a restart".
- Consequently `tmp_inodes` joins `STEADY_STATE_ERROR_CHECK_IDS`: its error is **permanent**, and accelerating the probe fan-out would fork ~10 children a minute on a host whose inode table — the very resource process spawning needs — is already exhausted.

## Follow-ups

- **#1874** — reclaim forked stores + abandoned worktrees (the headline inode reclaim).
- **#1875** — point spawned agents at a disk-backed `TMPDIR`. Its blocking premise is **verified**: Claude Code honours a spawn-side `TMPDIR`. This is the structural fix.

`Refs #1862` rather than `Closes`: prevention here is advisory and reclamation is deferred, so the parent stays open to carry context (repro, measured inode counts, host history) for both follow-ups. Close it manually if you judge the mitigation sufficient.

## Testing

- Both provider prompt forms + the sub-agent clause asserted.
- Forced-sweep bypass asserted across **both** `inodeUsePct` failure branches, plus a regression that a normal threshold still fails open and still reports the measured use%.
- `readTmpInodeUsePct` asserted to `statfs` `tmpdir()` (not `claudeTmpRoot()`), and to return `null` for btrfs-style `files: 0`.
- Classifier bands driven by the *passed* thresholds — a raised `SHEPHERD_TMP_INODE_PCT` moves the warning boundary.
- Steady-state exemption asserted **scoped**: a non-steady-state error alongside `tmp_inodes` still accelerates.
- Browser test: the row renders its label (not the raw id `tmp_inodes`) and the confirm modal shows `tmp_inodes` chrome — explicitly **not** the folder-trust "Mark this folder as trusted?" fallback that an unregistered `codeFixChrome` key produces.
- The unresolved-toast key selection was extracted to a pure `unresolvedFixKey` helper and unit-tested; there was no `SettingsDiagnosePanel` test of any kind, so even the pre-existing `host_capacity` branch was uncovered. Its return type is a closed union so the panel's key→message map is exhaustive at build time — `check:i18n` enforces key *parity*, not *existence*, and would not have caught a missing message.

`bun run lint`, `bun run test`, `cd ui && bun run check`, and the UI browser suite all pass.

Docs updated: `README.md`, `docs-site/.../reference/configuration.md`, `docs-site/.../operating.md`.


scoop (commented):
## What I verified

**Identifiers resolve.** The new imports in `src/diagnostics.ts:36-41` (`TMP_INODE_ERROR_PCT`, `readTmpInodeUsePct`, `sweepClaudeTmp`, `tmpInodeWarnPct`) all exist and are exported from `src/tmp-sweep.ts` (`:172`, `:186`, `:203`, `:361`). `readTmpInodeUsePct` calls the module-private `inodeUsePct` with the arity it declares (`src/tmp-sweep.ts:146-149`), and both of that helper's string skip-reasons (`"statfs-unavailable"` at `:153`/`:167`, `"root-missing"` at `:161`) collapse to `null`, matching the classifier's `optional` branch (`src/diagnostics.ts:85-87`). `tmpdir`/`envNum` are in scope (`src/tmp-sweep.ts:3`, `:34`). No import cycle: `tmp-sweep.ts` imports nothing from `diagnostics.ts`.

**i18n parity holds.** All eleven new keys exist in **both** catalogs at identical positions — `ui/messages/en.json:3344-3354` and `ui/messages/de.json:3344-3354` — so `ui/scripts/check-i18n.mjs` (union-vs-per-locale key diff, `:43-53`) passes and no value is empty.

**UI registration is complete.** `label()` is a dynamic `diagnostics_label_${id}` lookup (`ui/src/lib/components/DiagnoseRows.svelte:53-55`), so `diagnostics_label_tmp_inodes` (`ui/messages/en.json:3344`) resolves and the row will not render the raw id. `codeFixChrome` gains the `diagnostics_fix_action_tmp_inodes` entry (`ui/src/lib/components/DiagnoseRows.svelte:77-80`) beside `claude_trust`/`host_capacity`, so `codeFixTitle`/`codeFixRun` (`:81-91`) no longer fall through to the folder-trust `*_code` copy. The unresolved-toast selection is now `unresolvedFixKey` (`ui/src/lib/diagnostics-copy.ts:20-38`) consumed through an exhaustive `Record<UnresolvedFixKey, …>` (`ui/src/lib/components/settings/SettingsDiagnosePanel.svelte:14-19`); the closed union means a new branch without a message is a build error. `target` there is `DiagnosticCheck | undefined` from `.find()` (`SettingsDiagnosePanel.svelte:56`), which the helper's `| undefined` parameter accepts. Behaviour for `host_capacity` and the generic fallbacks is preserved byte-for-byte against the removed chain.

**Glossary gate passes.** `scripts/check-glossary.mjs` requires every `[[id|label]]` marker id to be registered (`:130-149`), termKey/bodyKey in both locales (`:107-117`), and `wikipedia.en`/`.de` non-empty for `kind:"external"` (`:155-169`). The `inode` entry (`ui/src/lib/glossary.ts:133-142`) satisfies all three, `gloss_inode_term`/`gloss_inode_def` are in both catalogs, and `gloss_wikipedia_link` already exists (`ui/messages/en.json:1590`, `de.json:1590`). Hints render through `GlossaryText` (`DiagnoseRows.svelte:132`), so the markers are not shown raw.

**The gate bypass is correct and scoped.** Both `inodeUsePct` failure reasons returned before the threshold comparison in the old code, so `thresholdPct: 0` genuinely was a silent no-op on a btrfs tmp or an absent root. The new branch (`src/tmp-sweep.ts:339-353`) skips the `statfs` call entirely only when `thresholdPct <= 0`; every `thresholdPct > 0` path keeps the original early returns and the original `` `swept …% inode use` `` string (`:352`, consumed at `:365`). A forced sweep against a non-existent root is safe — `sweepDir` swallows readdir failures (`:285-291`) and returns `removed: 0`.

**Dispatch and cadence.** `fix()` dispatches on `fixActionKey` after the `host_capacity` branch and before the shell-remediation fallback (`src/diagnostics.ts:1337-1343`); only `warning`/`error` carry that key, and `ok`/`optional` carry none, so the fallback `throw` still guards. `tmp_inodes` joining `STEADY_STATE_ERROR_CHECK_IDS` (`src/diagnostics.ts:204`) does what the comment claims — that set is consulted by `nextDiagnosticsDelay` to suppress acceleration, and a non-exempt error alongside it still accelerates. Production wiring needs nothing: `src/index.ts:2592` is the only `new DiagnosticsService(...)` call site and both new deps default functionally (`src/diagnostics.ts:127-135` resolver).

**Prompt block.** `tmpfsWorktreeNotice` is pushed unconditionally right after the stash notice (`src/service.ts:1394-1400`), inside the single-return `composeSystemPrompt` whose `blocks` array is built at `:1388` — so it genuinely rides research/autopilot/plan-gate/codex variants. `agentProvider` defaults to `"claude"` (`:1379`), and `AGENT_PROVIDERS` is exactly `["claude", "codex"]` (`src/types.ts:13`), so the two-way branch is total. The amended `WORKTREE_STASH_NOTICE` still contains the literal `git worktree add` (`src/service.ts:454-460`), keeping `test/service-plan-gate.test.ts:236` green.

**Docs.** README's "errors at 95%" and configuration.md's note match `TMP_INODE_ERROR_PCT = 95` (`src/tmp-sweep.ts:172`) and `tmpInodeWarnPct()`'s `SHEPHERD_TMP_INODE_PCT` read (`:186`). No user-facing string hardcodes `/tmp`; the check payload carries no path, honouring `src/types.ts:469-471`.

## The one thing to address

`tmpInodeWarnPct()` (`src/tmp-sweep.ts:186`) forwards `SHEPHERD_TMP_INODE_PCT` unclamped, and `envNum` (`:34-38`) *deliberately* honours a configured `0` (its own comment says so). An operator who sets `SHEPHERD_TMP_INODE_PCT=0` — a coherent, previously-harmless way to say "sweep on every tick" — now gets `warnPct = 0`, and `classifyTmpInodes` (`src/diagnostics.ts:96`) treats *any* `usePct >= 0` as a warning. The row is then permanently `warning` on a completely healthy host, degrades the health pip via `worstOf` (`:186-190`), and offers a fix that can never clear it. The symmetric case (`SHEPHERD_TMP_INODE_PCT` raised above 95) makes the warning band unreachable and the error band fire below the sweeper's own threshold, which contradicts the stated design that the row warns exactly where the sweeper acts. Not blocking — it needs a deliberate out-of-range config — but a floor/order guard on the bands (or clamping the warn band to `(0, errorPct]`) would close it.

## Limitations

- I did not run tests or lint (CI's job), and I could not verify the plan's empirical claims about sub-agent prompt inheritance or Claude Code's `TMPDIR` handling — that ground truth is not in this repo. The prompt wording is judged on internal consistency only.

- `resolveTmpSweepDeps` (`src/diagnostics.ts:127-135`) is an indirection for two `??` defaults; the stated reason (a constructor complexity gate) is plausible but it is more structure than the task needed.
- The new tests in `test/tmp-sweep.test.ts` shadow the module-level `fsp` import (`test/tmp-sweep.test.ts:3`) with a local `await import("node:fs/promises")` inside the helpers — harmless, but the existing import would have done.

scoop (commented):

**Prior #1 — `tmpInodeWarnPct()` forwarded `SHEPHERD_TMP_INODE_PCT` unclamped: ADDRESSED.** The raw forward is gone; `src/tmp-sweep.ts:390-394` now exposes `tmpInodeBands()` which clamps the warning band to `(0, 100]` (falling back to `DEFAULT_TMP_INODE_PCT = 80` outside it, `src/tmp-sweep.ts:367`) and lifts the error band with `Math.max(TMP_INODE_ERROR_PCT, warnPct)`. I checked both failure modes the prior review called out:
- knob `0` (`envNum` honours it — `src/tmp-sweep.ts:34-38`) → `warnPct` falls back to 80, so `classifyTmpInodes` (`src/diagnostics.ts:147-169`) cannot report a permanent `warning` on a healthy host; asserted at `test/tmp-sweep.test.ts` ("0 ('always sweep') does NOT become 'always warn'").
- knob `98` → `{warnPct: 98, errorPct: 98}`, so `error` no longer fires at 95 below the operator's own line; `150` → full fallback. Postcondition `0 < warnPct <= errorPct` is asserted over a table of inputs including `"abc"` and `""`.
The consumer chain is consistent: `defaultReadTmpInodes` (`src/diagnostics.ts:172-174`) re-reads `tmpInodeBands()` per probe, and `classifyTmpInodes` only compares against the passed bands.

## What I verified in the rest of the diff

- **i18n parity.** All 9 `tmp_inodes` keys plus `gloss_inode_term`/`gloss_inode_def` exist in both catalogs (`ui/messages/en.json:3343-3351` and `ui/messages/de.json:3343-3351`; glossary keys at `en.json:3352-3353` / `de.json:3352-3353`). Grep counts match (9 `tmp_inodes` keys per catalog).
- **UI registration.** `diagnostics_label_tmp_inodes` resolves through the dynamic `label()` lookup (`ui/src/lib/components/DiagnoseRows.svelte:53-55`), so the row shows its label not the raw id; `diagnostics_fix_action_tmp_inodes` is registered in `codeFixChrome` (`ui/src/lib/components/DiagnoseRows.svelte:77-80`) so the confirm modal does not fall back to the folder-trust `*_code` copy (`ui/messages/en.json:1603-1604` region).
- **Glossary gate.** `id: "inode"` with `kind: "external"` + `wikipedia {en,de}` (`ui/src/lib/glossary.ts:133-142`) matches the shape `scripts/check-glossary.mjs:50-72` regex-parses, and every `[[inode|…]]` marker in the new hints resolves to that id.
- **Behaviour preservation in the panel.** `unresolvedFixKey` (`ui/src/lib/diagnostics-copy.ts:24-40`) reproduces the old inline chain's order (host_capacity → any fixActionKey → generic) with the `tmp_inodes` branch inserted, and handles `undefined`; `SettingsDiagnosePanel.svelte:11-17` maps the closed union exhaustively, so a new branch without copy fails the build.
- **Forced-sweep bypass.** `sweepClaudeTmp` now skips `inodeUsePct` entirely when `thresholdPct <= 0` (`src/tmp-sweep.ts:359-376`); every `thresholdPct > 0` path keeps the previous `"statfs-unavailable"` / `"root-missing"` / `below-threshold` returns and the `swept …% inode use` reason string byte-for-byte. `defaultRunTmpSweep` (`src/diagnostics.ts:180-182`) passes `thresholdPct: 0`, and a missing root is tolerated by `sweepDir`'s readdir catch (`src/tmp-sweep.ts:304-311`).
- **fix() dispatch.** The new branch keys off `fixActionKey` (`src/diagnostics.ts:1336-1339`), after the `claude_trust` hintKey branch and the `host_capacity` branch, before the shell fallback — no other check carries `diagnostics_fix_action_tmp_inodes`, so no dispatch collision.
- **Steady-state exemption.** `tmp_inodes` added to `STEADY_STATE_ERROR_CHECK_IDS` (`src/diagnostics.ts:108`); `nextDiagnosticsDelay` only accelerates for non-exempt errors, and `overall` is consumed purely as a UI pip (`ui/src/lib/store.svelte.ts:93-96`) — no spawn gate depends on it, so an unclearable `error` row cannot block sessions.
- **Prompt block.** `tmpfsWorktreeNotice` rides every spawn (`src/service.ts:1345-1347`) with only wording varying on `agentProvider` (`"claude" | "codex"`, `src/types.ts:13`); the amended `WORKTREE_STASH_NOTICE` still contains the literal `git worktree add` that `test/service-plan-gate.test.ts` asserts.
- **Payload purity.** No absolute path crosses in the check (`classifyTmpInodes` returns only id/state/hintKey/fixActionKey) and no new user-facing string hardcodes `/tmp`.

## Nits (non-blocking)

The two items below are in "findings".

scoop (commented):

**1. Hint copy naming a control the row never shows — FIXED.**
`diagnostics_hint_tmp_inodes_high` / `_critical` now end with "Fix sweeps the caches Shepherd can reclaim." (`ui/messages/en.json:3346`, `:3347`), and the DE pair with „„Beheben“ räumt die Caches auf, die Shepherd freigeben kann." (`ui/messages/de.json:3346`, `:3347`). Verified against the actual control strings: `diagnostics_fix` = "Fix" (`ui/messages/en.json:1597`) / "Beheben" (`ui/messages/de.json:1597`), and `diagnostics_doc_link` = "How to fix" / „So behebst du das" (`en.json:1598`, `de.json:1598`) is the one that is *no longer* referenced. Confirmed the render path: `DiagnoseRows.svelte:135-141` shows the `diagnostics_fix()` button whenever `fixable(check)` (`:30-32`, i.e. non-ok + `fixActionKey`), and the `diagnostics_doc_link()` anchor only in the `:else if` at `:143-158`. Both non-ok `tmp_inodes` states carry `fixActionKey` (`src/diagnostics.ts:702-717`), so "Fix" is always the visible control. The prose copy in `docs-site/src/content/docs/operating.md:108` was likewise changed to "Its **Fix** button runs the sweep immediately".

**2. `fix()` doc block enumerating three paths — FIXED.** `src/diagnostics.ts:1303` now reads "Four dispatch paths", and the new tmp_inodes bullet sits at `:1312-1314` ahead of the shell-remediation bullet, matching the actual branch order (`:1321`, `:1328`, `:1339`, `:1343`).

## What I verified for the rest of the diff

- **Identifiers resolve.** `readTmpInodeUsePct`, `sweepClaudeTmp`, `tmpInodeBands` are all exported from `src/tmp-sweep.ts` (`:428`, `:337`, `:409`) and imported at `src/diagnostics.ts:36`. `TMP_INODE_ERROR_PCT` at `src/tmp-sweep.ts:177`. `AgentProvider` is a closed two-member union (`src/types.ts:13`), so the `agentProvider === "codex"` / else split in `tmpfsWorktreeNotice` (`src/service.ts:485-501`) is exhaustive.
- **i18n parity.** All 11 new keys exist in both catalogs at identical positions — `ui/messages/en.json:3344-3354` vs `ui/messages/de.json:3344-3354` (grepped the exact key set in both files). `ui/scripts/check-i18n.mjs:48` enforces parity only, which is why the browser test at `ui/src/lib/components/DiagnoseRows.browser.test.ts:277-311` and the helper test are the real existence guard.
- **UI registration is complete.** Label resolves through `msg[\`diagnostics_label_${id}\`]` (`DiagnoseRows.svelte:53-55`) → `diagnostics_label_tmp_inodes` exists. Confirm chrome registered at `DiagnoseRows.svelte:77-80` beside `claude_trust`/`host_capacity`, so `codeFixTitle`/`codeFixRun` (`:81-93`) no longer fall back to the folder-trust `*_code` copy (`en.json:1603-1604`). Unresolved-toast selection moved to `ui/src/lib/diagnostics-copy.ts:19-38`, and `SettingsDiagnosePanel.svelte:14-19` maps the closed `UnresolvedFixKey` union to messages that all exist (`en.json:1613`, `:1614`, `:3278`, `:3352`) — the `Record<UnresolvedFixKey, …>` typing makes a future missing branch a build error.
- **Glossary gate.** `scripts/check-glossary.mjs:107-117` requires `gloss_inode_term`/`gloss_inode_def` in both catalogs (present, `en.json:3353-3354`, `de.json:3353-3354`); `:130-149` requires every `[[id|label]]` marker id to be registered — `inode` is now in `ui/src/lib/glossary.ts:133-143`; `:155-169` requires en+de Wikipedia slugs for `kind:"external"` — both supplied.
- **Sweep gate bypass.** `src/tmp-sweep.ts:360-375`: `thresholdPct > 0` keeps the previous `inodeUsePct` path byte-for-byte (both string skip-reasons at `:141-143` still return `swept:false`, and `below-threshold` still quotes the measured %). `thresholdPct <= 0` skips `statfs` entirely and reports `"swept forced (gate bypassed)"`. The only behaviour delta for an existing config is `SHEPHERD_TMP_INODE_PCT=0` (honoured by `envNum`, `:34-38`), where the old code already swept whenever statfs succeeded — so the change only removes the silent no-op on statfs failure. Correct.
- **Payload purity.** `classifyTmpInodes` (`src/diagnostics.ts:697-720`) returns only `{id, state, hintKey, fixActionKey?}` — no absolute path, satisfying the `DiagnosticCheck` contract at `src/types.ts:470-491`. No user-facing string hardcodes `/tmp` (grepped both catalogs).
- **No unintended remediation collision.** `autoFixCommandFor` (`src/remediations.ts:181-184`) has no entry for the new hint keys, so `check()`'s annotation loop (`src/diagnostics.ts:1280-1284`) will not set `remediation` alongside `fixActionKey` (which `src/types.ts:483-486` calls mutually exclusive).
- **Probe ordering safe.** `tmp_inodes` is appended last in `probes()` (`src/diagnostics.ts:1238-1247`), so the `checks.splice(4, 0, …)` insertion of the agent-CLI checks (`:1276`) is unaffected.
- **Steady-state exemption.** `STEADY_STATE_ERROR_CHECK_IDS` now `["host_capacity", "tmp_inodes"]` (`src/diagnostics.ts:200`); `nextDiagnosticsDelay` (`:213+`) only exempts checks in that set, so a co-occurring transient error still accelerates — matching the tests at `test/diagnostics.test.ts:1236-1243`.
- **Test discovery.** `ui/src/lib/diagnostics-copy.test.ts` is picked up by the `node` vitest project (`ui/vite.config.ts:85-92`, which excludes only `**/*.browser.test.ts`).

### Stated limitation

The empirical claims the plan rests on (that sub-agents receive Claude Code's Scratchpad block but not Shepherd's composed prompt; that pnpm forks a `/tmp/.pnpm-store`) are about external runtimes and are not verifiable from this repo. I neither confirmed nor disputed them; the prompt text is judged only for internal consistency and for riding every spawn variant, which `test/service-plan-gate.test.ts:268-284` asserts against `composeSystemPrompt`.

### Nits (non-blocking, in findings)

See the two consistency items below.

scoop (commented):
## Summary

This PR ships the two pieces the (reduced) plan scoped: a provider-varied `<tmpfs-worktree-notice>` system-prompt block, and a `tmp_inodes` Doctor row with a forced-sweep fix, plus the gate-bypass fix in `sweepClaudeTmp`, UI registration, i18n, glossary, and docs. I reviewed for correctness, cross-file consistency, and scope. I found no blocking defects.

## Verification performed

**Message-catalog parity (en + de only — `ui/messages/*.json` is the full set).** All 11 new keys exist in both catalogs: `diagnostics_label_tmp_inodes`, `diagnostics_hint_tmp_inodes_{ok,high,critical,uninspectable}`, `diagnostics_fix_action_tmp_inodes`, `diagnostics_fix_confirm_{title,run}_tmp_inodes`, `diagnostics_fix_unresolved_tmp_inodes`, `gloss_inode_{term,def}` — verified 11 occurrences in `ui/messages/de.json` and the same set added in `ui/messages/en.json` (diff hunk). The four `UnresolvedFixKey` union members all resolve to real messages: `diagnostics_fix_unresolved` / `_code` / `_host_capacity` are pre-existing (`ui/messages/en.json:1613`, `:1614`, `:3278`) and `_tmp_inodes` is added in both.

**Identifier consistency.** `TMP_INODES_FIX_ACTION = "diagnostics_fix_action_tmp_inodes"` (`src/diagnostics.ts:696`) matches the `codeFixChrome` registration key (`ui/src/lib/components/DiagnoseRows.svelte:77`), the message key, and the `unresolvedFixKey` branch (`ui/src/lib/diagnostics-copy.ts:37`). All four hintKeys returned by `classifyTmpInodes` (`src/diagnostics.ts:721,727,735,739`) exist in both catalogs. Imports `readTmpInodeUsePct`, `sweepClaudeTmp`, `tmpInodeBands` are all exported from `src/tmp-sweep.ts`.

**Label rendering.** `DiagnoseRows.svelte:53-55` `label()` does `msg[\`diagnostics_label_${id}\`]?.() ?? id`; `diagnostics_label_tmp_inodes` is registered, so the row renders its label rather than the raw `tmp_inodes` id.

**`SettingsDiagnosePanel` refactor is behaviour-preserving.** The old inline chain (host_capacity → `_host_capacity`; other fixActionKey → `_code`; else → generic) is faithfully reproduced by `unresolvedFixKey` with the new `tmp_inodes` branch inserted ahead of the generic-code fallback (`ui/src/lib/diagnostics-copy.ts:34-42`). `UNRESOLVED_FIX_MSG` is typed `Record<UnresolvedFixKey, …>`, so the map is compiler-enforced exhaustive.

**Band logic.** `tmpInodeBands()` (`src/tmp-sweep.ts:208-211`) guarantees the `0 < warnPct <= errorPct` postcondition `classifyTmpInodes` relies on: out-of-range knob → default 80; `errorPct = max(95, warnPct)` so a knob above 95 raises the error band rather than inverting the range. `classifyTmpInodes` ordering (null→optional, ≥error→error, ≥warn→warning, else ok) is correct against that postcondition, including the degenerate `warnPct === errorPct` case (jumps ok→error, no warning band — acceptable).

**Gate bypass.** `sweepClaudeTmp`'s `thresholdPct <= 0` branch (`src/tmp-sweep.ts:357-379`) skips `inodeUsePct` entirely, so neither the `statfs-unavailable` nor `root-missing` early-return can suppress the operator's forced sweep, and it returns `reason: "swept forced (gate bypassed)"` with no fabricated percentage. Every `thresholdPct >= 1` path is byte-for-byte unchanged.

**Provider split is exhaustive.** `AGENT_PROVIDERS = ["claude", "codex"]` (`src/types.ts:13`), so `tmpfsWorktreeNotice`'s `=== "codex"` / else-Claude binary covers every provider; `composeSystemPrompt` defaults `agentProvider` to `"claude"` (`src/service.ts:1405`) and pushes the block unconditionally after `<worktree-stash-notice>` (`src/service.ts:1428`). The amended `WORKTREE_STASH_NOTICE` keeps the literal `git worktree add` substring that `test/service-plan-gate.test.ts` asserts.

**Steady-state exemption.** `tmp_inodes` is added to `STEADY_STATE_ERROR_CHECK_IDS` (`src/diagnostics.ts:200`), and the tests assert both that a `tmp_inodes`-only error does not accelerate and that a co-occurring transient error still does — the exemption is scoped, not blanket.

**Glossary.** `[[inode|inodes]]` markup matches the existing `[[id|display]]` convention (`ui/messages/en.json:185`), and the `inode` entry is registered in `ui/src/lib/glossary.ts` with matching `gloss_inode_term`/`gloss_inode_def` in both catalogs.

## Observations (non-blocking, no action required)

- `src/tmp-sweep.ts:347` — on the timer-driven path, setting `SHEPHERD_TMP_INODE_PCT=0` now forces an unconditional sweep even on a host where `statfs` is unavailable (btrfs `files: 0`), where previously it no-op'd. This is the documented "0 = always sweep" contract made consistent, not a regression; every threshold ≥ 1 is unchanged. Called out only for the record.
- `README.md:449` has a mid-sentence hard wrap ("runs the sweep immediately,\nignoring the threshold") — renders fine as Markdown; purely cosmetic.

Could not independently confirm the plan's empirical claims (sub-agents receive Claude Code's scratchpad block but not Shepherd's prompt; spawn-side `TMPDIR` is honoured) — those depend on runtime harness behaviour not present in this repo. They only affect the *motivation* for the directive's wording, not the correctness of the shipped code.

PR #1879, opened 2026-07-21
fix(security): move production /tmp paths off world-writable locations

### LINKED ISSUE #1868: CodeQL js/insecure-temporary-file: move the 6 production /tmp paths off world-writable locations

scoop:
Closes #1868.

## What

Clears all six CodeQL `js/insecure-temporary-file` alerts on production code. The deploy log and the per-session egress config/`dns.log` dir used **fixed names under `os.tmpdir()`** (`/tmp`), which another local user could pre-create or symlink. Both now resolve through one shared helper to a **user-private (`0700`) runtime dir**.

## How

New `src/runtime-dir.ts` → `shepherdRuntimeDir(...sub)`:
- **Primary:** `$XDG_RUNTIME_DIR/shepherd/…` — `/run/user/<uid>`, `0700`, user-owned, a tmpfs cleared at boot. Reliably present because shepherd runs as a `systemd --user` service.
- **Fallback:** `~/.shepherd/run/` when `XDG_RUNTIME_DIR` is unset (headless CI/containers) — still user-owned, never world-writable.

Standalone module rather than wiring through `config.ts`, because `update.ts` doesn't import `./config` and doing so would pull `applyHerdrSocket()`'s side effect into the update tests.

| Flagged site | Fix |
| --- | --- |
| `update.ts:358` (path from `:280`) | `logPath` default → `shepherdRuntimeDir("shepherd-update.log")`; `mkdirSync` guard before the `writeFileSync` (a derived dir may not exist — `/tmp` always did) |
| `egress.ts:638-641` (`egressTmpDir` `:623`, `egressTmpRoot` `:628`) | → `shepherdRuntimeDir("egress", …)` |
| `egress-watch.ts:24` | no change — `dnsLogPath` derives from `egressTmpDir`, clears transitively |

Also: dropped the now-orphaned `tmpdir` import from `update.ts`; corrected stale JSDoc in `update.ts`, `egress.ts` (`egressTmpDir` + `sweepEgressTmp`), and `service.ts` (`shepherd-egress/<id>` → `egress/<id>`).

## Why no `applyState()` marker-branch TTL work

`applyState()`'s marker branch returns `done`/`failed` with **no age check** — survivable today only because `/tmp` is boot-cleared. `$XDG_RUNTIME_DIR` is *also* a boot-cleared tmpfs, so the property is preserved. Stronger: `defaultLaunch()` deploys via `systemd-run --user`, which **itself requires `XDG_RUNTIME_DIR`** — so a real deploy can only run in the primary (boot-cleared) branch, never the persistent `~/.shepherd/run/` fallback. No TTL work is needed.

## Testing

- `bun run lint` ✓ · `bun run typecheck` ✓ · `bun test ./test` → **7488 pass, 0 fail** ✓
- New `test/runtime-dir.test.ts` covers **both** helper branches (XDG set; XDG cleared + `HOME` set; blank XDG treated as unset).
- `test/egress.test.ts` pins `XDG_RUNTIME_DIR` independently and asserts the resolved path **is not** the old `join(tmpdir(), "shepherd-egress", id)` — so the assertion actually proves the path left `/tmp`.

## Egress cannot be verified by `bun run test`

Per the issue: the runner's `$TMP` readability checks are fail-closed and run *before* `unshare`, the dir is ro-bind-mounted into the sandbox, and every live runner test is `skipIf`-gated. A regression would abort **every** sandboxed spawn. The change only moves *where* the dir lives — the runner reads it as the real user before `unshare`, and every bwrap override bind targets files by **absolute path** — but this needs one real confined spawn to confirm.

```shepherd:manual-steps
- [ ] Verify egress on a capable host (setpriv/unshare/slirp4netns/nft/dnsmasq): run one real egress-confined autonomous spawn and confirm the agent starts + DNS drops are still observed, before relying on this change.
```

### DIFF
--- a/src/egress.ts
+++ b/src/egress.ts
@@ -25,6 +25,7 @@ import { tmpdir } from "node:os";
 import { join, resolve } from "node:path";
 import { execFileSync } from "./instrument";
 import { resolveNodeBin } from "./node-bin";
+import { shepherdRuntimeDir } from "./runtime-dir";
 import {
   type BackendProbeDeps,
   type MembraneInputs,
@@ -616,16 +617,17 @@ export function egressMembraneOverrideFlags(
 
 /**
  * Deterministic per-session temp dir holding the generated egress config files +
- * the dns.log the drop-watcher tails. Lives under the OS temp dir; one dir per
- * session id so the startup sweep can reconcile orphans against the live set.
+ * the dns.log the drop-watcher tails. Lives under the user-private runtime dir
+ * ({@link shepherdRuntimeDir}, `0700`) as `egress/<sessionId>` — one dir per session
+ * id so the startup sweep can reconcile orphans against the live set.
  */
 export function egressTmpDir(sessionId: string): string {
-  return join(tmpdir(), "shepherd-egress", sessionId);
+  return shepherdRuntimeDir("egress", sessionId);
 }
 
 /** Root dir holding every per-session egress temp dir (for the orphan sweep). */
 function egressTmpRoot(): string {
-  return join(tmpdir(), "shepherd-egress");
+  return shepherdRuntimeDir("egress");
 }
 
 /**
@@ -669,7 +671,7 @@ export function removeEgressTmp(sessionId: string): void {
 }
 
 /**
- * Startup reconcile sweep: remove any `shepherd-egress/<id>` dir whose id is NOT in
+ * Startup reconcile sweep: remove any `egress/<id>` dir whose id is NOT in
  * `liveSessionIds` (the non-archived session set). Bounds unbounded growth from
  * sessions whose teardown removal was missed (crash, restart). Best-effort —
  * never throws; a single unremovable entry is skipped.
--- a/src/runtime-dir.ts
+++ b/src/runtime-dir.ts
@@ -1,0 +1,22 @@
+import { join } from "node:path";
+
+/**
+ * User-private base directory for Shepherd's ephemeral runtime files (the deploy
+ * log, the per-session egress config + dns.log). Prefers `$XDG_RUNTIME_DIR`
+ * (`/run/user/<uid>`, `0700`, user-owned, a tmpfs cleared at boot) — which satisfies
+ * CodeQL `js/insecure-temporary-file` (no fixed name in world-writable `/tmp`) AND
+ * preserves the boot reset the update service's `applyState()` marker branch silently
+ * relies on. Shepherd runs as a `systemd --user` service, so the variable is present.
+ *
+ * Falls back to `~/.shepherd/run/` (also user-owned, never world-writable) when
+ * `XDG_RUNTIME_DIR` is unset — a headless CI / container with no login session. A real
+ * deploy cannot reach this branch: `systemd-run --user` itself requires `XDG_RUNTIME_DIR`.
+ *
+ * Read at call time (no module-load snapshot) so tests can flip the env vars. Callers
+ * must `mkdirSync(..., { recursive: true })` before writing — the dir may not exist yet.
+ */
+export function shepherdRuntimeDir(...sub: string[]): string {
+  const xdg = process.env.XDG_RUNTIME_DIR?.trim();
+  const base = xdg ? join(xdg, "shepherd") : join(process.env.HOME ?? "", ".shepherd", "run");
+  return join(base, ...sub);
+}
--- a/src/runtime-dir.ts
+++ b/src/runtime-dir.ts
@@ -1,4 +1,5 @@
-import { join } from "node:path";
+import { homedir } from "node:os";
+import { isAbsolute, join } from "node:path";
 
 /**
  * User-private base directory for Shepherd's ephemeral runtime files (the deploy
@@ -14,9 +15,15 @@ import { join } from "node:path";
  *
  * Read at call time (no module-load snapshot) so tests can flip the env vars. Callers
  * must `mkdirSync(..., { recursive: true })` before writing — the dir may not exist yet.
+ *
+ * The fallback base uses `$HOME` when it is an absolute path, else `os.homedir()` (the
+ * OS passwd entry) — so it is always absolute, never a cwd-relative `.shepherd/run`.
  */
 export function shepherdRuntimeDir(...sub: string[]): string {
   const xdg = process.env.XDG_RUNTIME_DIR?.trim();
-  const base = xdg ? join(xdg, "shepherd") : join(process.env.HOME ?? "", ".shepherd", "run");
+  const home = process.env.HOME;
+  const base = xdg
+    ? join(xdg, "shepherd")
+    : join(home && isAbsolute(home) ? home : homedir(), ".shepherd", "run");
   return join(base, ...sub);
 }
--- a/src/service.ts
+++ b/src/service.ts
@@ -4902,7 +4902,7 @@ export class SessionService {
   private pendingRunawayIds: Set<string> | null = null;
 
   /**
-   * Startup reconcile: remove any orphaned `shepherd-egress/<id>` temp dir whose id is
+   * Startup reconcile: remove any orphaned `egress/<id>` temp dir whose id is
    * not a currently live (non-archived) session — bounds unbounded growth from teardown
    * removals missed across a crash/restart. Best-effort (never throws). Call once at
    * server boot. A live session's dir (incl. its dns.log) is preserved.
--- a/src/update.ts
+++ b/src/update.ts
@@ -1,10 +1,10 @@
 import { execFile, spawn, spawnSync } from "node:child_process";
 import { createHash, type Hash } from "node:crypto";
-import { readFileSync, statSync, writeFileSync } from "node:fs";
-import { tmpdir } from "node:os";
-import { join } from "node:path";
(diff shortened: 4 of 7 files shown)


scoop (commented):
## Summary

The PR relocates all six flagged `js/insecure-temporary-file` production paths off world-writable `/tmp` behind a single shared helper `shepherdRuntimeDir()` that prefers `$XDG_RUNTIME_DIR/shepherd` (`0700`, user-owned, boot-cleared tmpfs) and falls back to `~/.shepherd/run`. The approach matches the approved plan and the originating issue, and the security fix is real: both branches root the derived dirs under a directory an attacker cannot write to or traverse.

### Verification performed

- **All old paths retired.** Grepped the tree: no remaining production reference to `shepherd-egress` or `shepherd-update.log` except `src/egress.ts:197` (the `mkdtemp` probe — random-suffixed, correctly unflagged, `tmpdir` import legitimately retained) and the negative assertion at `test/egress.test.ts:945`.
- **Imports consistent.** `src/update.ts:3-4` adds `mkdirSync`/`dirname`, drops `tmpdir`; `join` still used at `src/update.ts:280`. `test/service.test.ts` and `test/egress.test.ts` keep other `tmpdir` uses, so no import is orphaned.
- **No script hardcodes the path.** No `*.sh` references the old locations; the runner receives the dir as an absolute path via `--tmp`, so bwrap binds track the new base (verified against `src/egress.ts:200`).
- **Boot-reset dependency preserved.** `defaultLaunch` launches via `systemd-run --user` (`src/update.ts:380`), which itself needs `XDG_RUNTIME_DIR`, so a real deploy can only use the boot-cleared XDG branch — the `applyState()` marker branch stays correct with no age-bound, as argued. `mkdirSync(dirname(this.logPath), …)` (`src/update.ts:359`) correctly precedes the `writeFileSync` truncate.
- **Egress path stays single-sourced.** `test/sandbox-wiring.test.ts:7,149` build expected paths through `egressTmpDir()`, so they follow the new base without edit.
- **CodeQL clearance.** Flagged sites now derive from `process.env.XDG_RUNTIME_DIR`/`HOME` rather than `os.tmpdir()`/hardcoded `/tmp`, so the six alerts should clear; the surviving `tmpdir` probe is a `mkdtemp` pattern CodeQL does not flag.

### Minor nit (non-blocking)

`shepherdRuntimeDir` (`src/runtime-dir.ts:20`) uses `process.env.HOME ?? ""`, so if both `XDG_RUNTIME_DIR` and `HOME` are unset the base becomes the *relative* path `.shepherd/run`, resolved against cwd rather than an absolute user dir. This is essentially unreachable in the real service (a `systemd --user` unit always has `HOME`), and it is not a security regression (cwd/`repoDir` is not world-writable), so it does not block — but a guard (throw, or fall back to a defined absolute base) would be more robust than silently emitting a cwd-relative path.

Per the issue and plan, the egress relocation cannot be exercised by `bun run test` (fail-closed `$TMP` checks before `unshare`, ro-bind-mount, `skipIf`-gated runner tests). I confirmed the code only changes *where* the dir lives and that binds use absolute paths, but I could not verify a live confined spawn against `/run/user/<uid>` (tmpfs) end-to-end in this repo — the residual runtime risk the PR flags as a manual verification step remains valid and unverifiable here.

scoop (commented):
## Summary

This revision moves the six `js/insecure-temporary-file` production paths off world-writable `/tmp` onto a user-private (`0700`) runtime dir via a new shared helper `shepherdRuntimeDir(...sub)` (`src/runtime-dir.ts`). The change is faithful to the approved plan and clears the alerts at their source (`egressTmpDir`, `egressTmpRoot`, `UpdateService.logPath` default).

**Prior point 1 (runtime-dir.ts: `HOME` unset → cwd-relative base):** Addressed. The fallback base is now `join(home && isAbsolute(home) ? home : homedir(), ".shepherd", "run")` (`src/runtime-dir.ts:26`), so it is always absolute — verified against the new test `fallback base stays absolute when HOME is unset or relative` (`test/runtime-dir.test.ts:38-51`), which covers both the unset and relative-HOME cases and asserts `isAbsolute(...) === true`. Not re-raised.

## Verification performed

- `shepherdRuntimeDir` is exported at `src/runtime-dir.ts:20` and imported consistently in `src/update.ts:7`, `src/egress.ts:28`, `test/service.test.ts:13`, `test/runtime-dir.test.ts:4`. Spelling/casing consistent across all sites.
- `tmpdir` imports retained only where still used: `src/egress.ts:197` (the `mkdtemp` probe, correctly left flagged-free), `test/egress.test.ts:945,983`, `test/service.test.ts` (many `mkdtempSync` sites). No orphaned import.
- Directory-creation guards confirmed: `mkdirSync(dirname(this.logPath), { recursive: true })` added before `writeFileSync(this.logPath, "")` in `defaultLaunch` (`src/update.ts:357-361`); egress config dir already created at `src/egress.ts:639`; the startup sweep swallows ENOENT (`src/egress.ts:682-686`).
- No dangling references to the old world-writable paths in `src/` — grep for `shepherd-egress`/`shepherd-update.log` returns only the new helper call (`src/update.ts:281`) and the untouched probe (`src/egress.ts:197`).
- Test rewiring is sound: `test/egress.test.ts:937-950` pins `XDG_RUNTIME_DIR` independently and asserts `not.toBe(join(tmpdir(), "shepherd-egress", ...))`, so it genuinely proves the path left `/tmp` (non-tautological, per plan); `test/service.test.ts:7387,7418` cleanup re-pointed at `shepherdRuntimeDir("egress")`.
- JSDoc corrections (update.ts:235-237, egress.ts:619-623/674, service.ts:4905) match the new on-disk layout.

No blocking issues and no residual findings. Decision is `comment`.

Issue #1880, opened 2026-07-21
Partial pnpm store reclaim (free unlinked content instead of all-or-nothing)


scoop:
Follow-up to #1874, which shipped the worktree reaper + **all-or-nothing** forked-store reclaimer (skip the whole `.pnpm-store` if *any* content file still has `nlink > 1`).

## Why partial reclaim is worth it

All-or-nothing reclaims **zero** in the residual case where a *refused* worktree (dirty / locked / live / fresh) still hardlinks some store files — exactly the case where inodes are most stuck. Partial reclaim would free the unlinked fraction (`nlink === 1` content) plus emptied bucket dirs even then.

## Phase-0 evidence (measured while planning #1874 — safe to build)

pnpm 10.28.2, tmpfs `/tmp`, forked store `v10`. Delete every `nlink === 1` file under `<store>/v10/files/**`, leave `index/` intact, reinstall:

- **offline** → `ERR_PNPM_NO_OFFLINE_TARBALL` (exit 1): pnpm detects the missing content and *wants to download it* — it does **not** wedge believing it holds an unlinkable artifact.
- **online** → clean re-fetch (exit 0), package resolves.

So the feared "clean re-download becomes a failing install" does **not** occur. Partial reclaim is safe *given network at reinstall time*.

## Scope

- Replace the all-or-nothing removal in `reclaimForkedPnpmStore` (`src/tmp-sweep.ts`) with a per-file pass that unlinks only `nlink === 1` content and prunes now-empty bucket dirs, keeping every existing skip gate (retained worktrees, depth-2 idleness, non-`v*` sibling freshness, inode-pressure threshold).
- Keep the exhaustive walk; the difference is *unlink the unlinked file* vs *bail on the first link*.
- Re-verify `index/` metadata pointing at removed content stays a clean re-fetch trigger (not a hard error) after the change.

Refs #1874

scoop:
✅ PR #1886 merged.

PR #1881, opened 2026-07-21
fix(tmp): reclaim forked pnpm stores + abandoned agent worktrees

scoop:
Closes the reclamation gap left by #1862/#1876. Those shipped **prevention** (the `<tmpfs-worktree-notice>` steering agents off `/tmp`) and **diagnosis** (the `tmp_inodes` Diagnose row + one-click forced sweep), but nothing that reclaims inodes already lost to abandoned `/tmp` agent worktrees and the forked pnpm store beside them — so `tmp_inodes` can legitimately sit `warning`/`error` after a *successful* forced sweep. This adds the two reclaimers.

Refs #1874.

## Two reclaimers, wired into `fireTmpSweep` (boot + daily), run **in order**

pnpm hardlinks `node_modules` into its store; an inode frees only on the **last** link drop. While a tmp worktree survives, every store file sits at `nlink = 2`, so removing the store reclaims ~zero *and* the next install re-forks it — net inode use goes **up**. So the worktree reaper runs first, and the store pass consumes its `retained` count as a go/no-go gate.

### `reapAbandonedWorktrees`
Removes stale, clean, unused git worktrees of *configured* repos that physically live under a tmp root (a nested agent's `git worktree add /tmp/…` left behind). Destructive against possibly-unrecoverable data, so it is **gated on live inode pressure** and walled by refusals — **dirty** (`git status --porcelain`, any error = dirty), porcelain **`locked`/`bare`/`prunable`**, the Shepherd `/.shepherd-worktrees/` marker, **live-session** `worktreePath`, **live same-uid `/proc` cwd**, and mtime **freshness** (bounded walk, `node_modules`/`.git` pruned from descent, budget/error → keep). Every comparison is **realpath-normalized** (symlinked `/tmp`, e.g. macOS `/private/tmp`, must not read as "not protected"). Git-floor **2.38** for the `locked`/`prunable` annotations. `retained` = candidates under a tmp root still on disk after the pass.

### `reclaimForkedPnpmStore`
**All-or-nothing** removal of `<tmp>/.pnpm-store` — skip the whole store if *any* file still has `nlink > 1`. Gated on `retained === 0`, inode pressure, non-`v*` sibling freshness, **depth-2 idleness**, and an **exhaustive early-exit `nlink` scan**. Every non-exhaustive outcome (error, budget, missing `files/`) resolves to **keep** — `removed:false` means *provably idle*, never *failed to probe*.

## Phase-0: partial reclaim is safe, but deferred (#1880)

Measured while planning (pnpm 10.28.2, tmpfs, store `v10`): deleting `nlink==1` content makes the next **offline** install error `ERR_PNPM_NO_OFFLINE_TARBALL` and the next **online** install **re-fetch cleanly**. So partial reclaim is safe *given network* — but this PR ships the conservative **all-or-nothing** path; partial reclaim is filed as **#1880** with the evidence.

## Reuse, not new surface
- **No new env vars** — reuses `SHEPHERD_TMP_INODE_PCT` (gate) + `SHEPHERD_TMP_STALE_HOURS` (staleness).
- `parseWorktrees` is **exported + extended** (`locked`/`bare`/`prunable`, `locked`/`prunable` **prefix-matched** for their reason-carrying forms) rather than adding a third porcelain parser.
- `WORKTREE_MARKER`, `isUnder`, and a new `liveProcCwds()` are exported from `process-reaper` for reuse; the sync `/proc` scan is snapshotted at the `index.ts` layer and passed in, keeping it out of the all-async `tmp-sweep`.

## Tests / checks
- New unit tests for both reclaimers + `resolveStoreVersionDirs` + the extended `parseWorktrees` (incl. a **symlinked-tmp-root** refusal test). Fully injectable ops — no real git/`/proc`/tmpfs.
- `bun run lint`, `bun run typecheck`, `bun test ./test` (7513 pass), and `fallow audit` all pass.

## Out of scope
Anything under `/tmp` that is not a worktree of a *configured* repo (an unconfigured repo's worktree, a plain clone, a bare dir that ran an install) — invisible to configured-repo discovery; the `nlink` scan is what stops the store pass from being counterproductive in their presence. Revisiting `tmp_inodes`' `STEADY_STATE_ERROR_CHECK_IDS` membership is a separate follow-up.

### DIFF
--- a/src/forge/local.ts
+++ b/src/forge/local.ts
@@ -164,32 +164,58 @@ export async function mergeTreeWriteTree(
   );
 }
 
-interface WorktreeEntry {
+export interface WorktreeEntry {
   path: string;
   head?: string;
   branch?: string; // short name (refs/heads/ stripped)
   detached: boolean;
+  /** worktree is locked (`git worktree lock`); its record can't be pruned. */
+  locked: boolean;
+  /** the main worktree of a bare-ish checkout (`bare` line). */
+  bare: boolean;
+  /** git considers the record prunable (its working dir is gone / broken). */
+  prunable: boolean;
+}
+
+/** Apply one non-`worktree` porcelain line to the current entry.
+ *
+ *  `locked`/`prunable` MUST be prefix-matched: git emits a bare `locked` /
+ *  `prunable` line when no reason is present, but `locked <reason>` /
+ *  `prunable <reason>` when one is. An exact match (as for `detached`/`bare`,
+ *  which never carry a reason) would silently drop the flag on the
+ *  reason-carrying form — and a dropped `locked`/`prunable` is exactly the
+ *  worktree whose dangling record cannot be pruned, so a reaper must never
+ *  treat it as reapable. */
+function applyWorktreeAttr(cur: WorktreeEntry, line: string): void {
+  if (line.startsWith("HEAD ")) cur.head = line.slice("HEAD ".length);
+  else if (line.startsWith("branch "))
+    cur.branch = line.slice("branch ".length).replace(/^refs\/heads\//, "");
+  else if (line === "detached") cur.detached = true;
+  else if (line === "bare") cur.bare = true;
+  else if (line === "locked" || line.startsWith("locked ")) cur.locked = true;
+  else if (line === "prunable" || line.startsWith("prunable ")) cur.prunable = true;
 }
 
 /** Parse `git worktree list --porcelain` into structured entries. Blocks are
  *  separated by blank lines; lines are `worktree <path>`, `HEAD <sha>`,
- *  `branch refs/heads/<name>`, or `detached`. */
-function parseWorktrees(porcelain: string): WorktreeEntry[] {
+ *  `branch refs/heads/<name>`, `detached`, `bare`, `locked [<reason>]`, or
+ *  `prunable [<reason>]` (per-line handling in `applyWorktreeAttr`). */
+export function parseWorktrees(porcelain: string): WorktreeEntry[] {
   const out: WorktreeEntry[] = [];
   let cur: WorktreeEntry | null = null;
   for (const raw of porcelain.split("\n")) {
     const line = raw.trimEnd();
     if (line.startsWith("worktree ")) {
       if (cur) out.push(cur);
-      cur = { path: line.slice("worktree ".length), detached: false };
-    } else if (!cur) {
-      continue;
-    } else if (line.startsWith("HEAD ")) {
-      cur.head = line.slice("HEAD ".length);
-    } else if (line.startsWith("branch ")) {
-      cur.branch = line.slice("branch ".length).replace(/^refs\/heads\//, "");
-    } else if (line === "detached") {
-      cur.detached = true;
+      cur = {
+        path: line.slice("worktree ".length),
+        detached: false,
+        locked: false,
+        bare: false,
+        prunable: false,
+      };
+    } else if (cur) {
+      applyWorktreeAttr(cur, line);
     }
   }
   if (cur) out.push(cur);
--- a/src/index.ts
+++ b/src/index.ts
@@ -106,6 +106,7 @@ import {
   ProcessReaper,
   reapDeletedWorktreeOrphans,
   reapMarkedOrphans,
+  liveProcCwds,
   SESSION_MARKER_ENV,
 } from "./process-reaper";
 import {
@@ -113,6 +114,8 @@ import {
   compileCacheDir,
   reapFallowCaches,
   pruneRepoWorktrees,
+  reapAbandonedWorktrees,
+  reclaimForkedPnpmStore,
   scratchpadHasFiles,
 } from "./tmp-sweep";
 import { runSessionUsageBackfill } from "./usage-backfill";
@@ -820,6 +823,31 @@ const fireTmpSweep = (phase: "boot" | "daily") => {
         console.warn(`[tmp-sweep] ${phase}: reaped ${reaped} deleted-worktree orphan(s)`);
     })
     .catch((err) => console.warn(`[tmp-sweep] ${phase} orphan reap failed:`, err));
+
+  // #1874: reclaim abandoned agent worktrees under /tmp, THEN the forked pnpm store they
+  // pinned. Order is load-bearing — while a tmp worktree survives, every store file sits at
+  // nlink=2, so removing the store first reclaims ~zero and the next install re-forks it.
+  // The reaper's `retained` count is the store pass's go/no-go gate. A single microtask-
+  // deferred liveProcCwds() snapshot keeps that synchronous /proc scan off the event loop
+  // and out of the all-async tmp-sweep module.
+  void Promise.resolve()
+    .then(async () => {
+      const liveWorktreePaths = store
+        .list({ activeOnly: true })
+        .map((s) => s.worktreePath)
+        .filter(Boolean);
+      const reap = await reapAbandonedWorktrees({
+        repoPaths: listRepos(config.repoRoot).map((r) => r.path),
+        liveWorktreePaths,
+        liveCwds: liveProcCwds(),
+      });
+      const storeResult = await reclaimForkedPnpmStore({ retained: reap.retained });
+      console.warn(
+        `[tmp-sweep] ${phase}: worktree reap removed ${reap.reaped} (retained ${reap.retained}), ` +
+          `pnpm store ${storeResult.removed ? "reclaimed" : `kept (${storeResult.reason})`}`,
+      );
+    })
+    .catch((err) => console.warn(`[tmp-sweep] ${phase} worktree/store reclaim failed:`, err));
 };
 
 deferredStarts.push(() => {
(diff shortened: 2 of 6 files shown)


scoop (commented):
## Summary

The PR adds the two reclaimers (`reapAbandonedWorktrees`, `reclaimForkedPnpmStore` + `resolveStoreVersionDirs`) in `src/tmp-sweep.ts`, shares/extends `parseWorktrees` in `src/forge/local.ts`, exports `WORKTREE_MARKER`/`isUnder`/`liveProcCwds` from `src/process-reaper.ts`, and wires reaper-then-store into `fireTmpSweep` in `src/index.ts`. I read the full diff and traced each safety-critical path. The implementation faithfully realizes the plan and I found no blocking correctness or security defect. Only one non-blocking documentation nit.

## What I verified

**Imports resolve and are spelled/exported consistently.**
- `parseGitVersion`, `gitVersionAtLeast`, `MIN_GIT_MAJOR` (=2), `MIN_GIT_MINOR` (=38) exist and are exported at `src/forge/local.ts:30-31,68,74`; `gitMeetsFloor` uses them correctly (`src/tmp-sweep.ts:748-756`).
- `WORKTREE_MARKER` (`src/process-reaper.ts:150`) and `isUnder` (`src/process-reaper.ts:166`) are now `export`ed and imported at `src/tmp-sweep.ts:13`; no local shadow exists (grep of `src/tmp-sweep.ts`). No circular import — `src/process-reaper.ts` references tmp-sweep only in comments, never `import`s it.
- `liveProcCwds` is exported at `src/process-reaper.ts:447` and `scanProcs()` returns `{pid,cwd,comm}` (`src/process-reaper.ts:39,348-371`), so `.map(p => p.cwd)` is valid.
- `store.list({activeOnly})` returns `Session` whose `worktreePath` is `string` NOT NULL (`src/store.ts:368,897`), so `.map(s => s.worktreePath).filter(Boolean)` in `src/index.ts:835-838` is sound.

**Interface change is additive/non-breaking.** The only non-test consumer of `parseWorktrees` is `squashMergeLocal` (`src/forge/local.ts:283`), which reads `.branch/.path/.head` only; no code constructs a `WorktreeEntry` literal outside `parseWorktrees` itself, so the three new required booleans don't break callers (verified via grep of `WorktreeEntry`/`parseWorktrees` across `src/`+`test/`).

**Prefix-match on `locked`/`prunable`** is correct (`src/forge/local.ts:199-200`): `line === "locked" || line.startsWith("locked ")`, so the reason-carrying form still sets the flag — the exact hazard the git floor exists to prevent. `bare`/`detached` stay exact-matched. Covered by `test/local-forge.test.ts:451-499`.

**Keep-biased directions all hold.**
- Reaper: git-floor unreadable ⇒ discover-only; any refusal (locked/bare/prunable/marker/live-session/live-cwd/dirty incl. status-error/fresh) ⇒ keep; freshness walk resolves stat/readdir error, budget exhaustion, and depth to *keep* (`src/tmp-sweep.ts:620-651,687-712`); top-level `catch` returns `retained:1` so the store pass skips on any unexpected failure (`src/tmp-sweep.ts:820-824`).
- `retained = candidates.length - reaped`, and every candidate is realpath-resolved (implies on-disk) and under a tmp root — so it is exactly "under-tmp AND still on disk", not the pre-filter refusal list (`src/tmp-sweep.ts:816`). Matches the issue's requirement that off-tmp/marker/live entries must not pin `retained` permanently.
- Store: `retained>0`, statfs unreadable, below-threshold, empty `resolveStoreVersionDirs`, fresh/unrecognized non-`v*` sibling, non-idle `v*`, or any surviving `nlink>1`/error all resolve to keep with a reason (`src/tmp-sweep.ts:911-947`). The `nlink` walk early-exits on the first `nlink>1` and treats missing `files/`/any error as keep (`src/tmp-sweep.ts:872-897`). Idleness probes depth-2 (stats depth-1 entries and their children) via `storeVersionDirIsIdle`+`childrenAllStale` (`src/tmp-sweep.ts:846-869`).

**Ordering + gate wiring.** In `src/index.ts:833-849` the reaper is awaited before the store pass in one async chain, and `reap.retained` is threaded into `reclaimForkedPnpmStore` — so `git worktree remove` drops the `node_modules` links before the store's `nlink` scan runs, exactly as the load-bearing order requires.

**Symlinked-tmp handling.** Both sides of every comparison are realpath-normalized (`resolveAllRealpaths` for `tmpRoots`/`liveWorktreePaths`/`liveCwds`, `realpath` for each candidate) and containment goes through segment-aligned `isUnder`, so a symlinked `/tmp` (macOS `/private/tmp`) cannot read as "not protected" — covered by the symlink test at `test/tmp-sweep.test.ts` ("symlinked tmp root").

**Store-root path assumption** `join(tmpdir(), ".pnpm-store")` is `/tmp/.pnpm-store` for the server (which runs with no TMPDIR override, per `src/tmp-sweep.ts:88`), matching the issue's measured `/tmp/.pnpm-store/vN`; a wrong mount root fails to `no-store` (safe skip). Only the exact name `.pnpm-store` reaches bare `tmpdir()`; `REGENERABLE_CACHE`/`sweepRoots` are unchanged.

### Non-blocking nit

- `src/forge/local.ts:180-192`: the detailed doc block describing `parseWorktrees` (the porcelain grammar + the prefix-match rationale) now sits directly above the `applyWorktreeAttr` helper, with that helper's own one-line comment wedged between it and the function. `parseWorktrees` itself (line 203) is left with no doc comment above it. Move the descriptive block back above `parseWorktrees` so the documentation attaches to the function it describes.

## Could not verify

- Whether pnpm actually forks its store to `/tmp/.pnpm-store` (versioned `vN`) in this deployment is an empirical claim from the originating issue, not checkable from this repo; the code fails safe (no-store skip) if the layout/root differs, so the assumption is not load-bearing for safety.

Nothing here blocks. Decision: comment.

scoop (commented):
## Re-review

**Prior finding (doc block placement) — RESOLVED.** `parseWorktrees` now carries its own descriptive doc block at `src/forge/local.ts:199-202`, and `applyWorktreeAttr` has a separate block at `:180-188`. Neither function is left undocumented. Verified against `src/forge/local.ts:180-223`.

## Verification performed

- **Imports resolve.** `parseWorktrees`, `parseGitVersion`, `gitVersionAtLeast`, `MIN_GIT_MAJOR`, `MIN_GIT_MINOR` are all exported from `src/forge/local.ts` (verified `:30-31`, `:68`, `:74`, `:203`). `WORKTREE_MARKER`, `isUnder`, `liveProcCwds` are all exported from `src/process-reaper.ts` (verified `:152`, `:167`, `:448`).
- **`liveProcCwds()` shape.** `scanProcs()` returns `{pid,cwd,comm}[]` (`src/process-reaper.ts:39`), and `liveProcCwds` maps to `.cwd` (`:449`). Called inside the microtask-deferred `Promise.resolve().then(...)` in `src/index.ts:830-847`, so the synchronous `/proc` scan stays off the sync path as the plan requires.
- **`inodeUsePct` union handled on both paths.** It returns `number | string` (`src/tmp-sweep.ts:157`). The reaper treats a string as `pressureOk === false` → discover-only (`typeof usePct === "number" && usePct >= thresholdPct`). The store reclaimer returns a keep on the string case before the numeric compare. Both consistent.
- **Ordering + gate.** `fireTmpSweep` runs reaper then store, passing `retained` through (`src/index.ts:838-843`); store skips on `retained > 0` first (`src/tmp-sweep.ts` `reclaimForkedPnpmStore`). The reaper is TOTAL (its outer catch returns `retained: 1`), so any unexpected throw conservatively disables the store pass.
- **`retained` semantics.** Every candidate is realpath-resolved (so on-disk) and tmp-scoped; `retained = candidates.length - reaped` counts refused + failed-removal candidates only — not the raw pre-filter refusal list — matching the issue's requirement that off-tmp refusals not pin it permanently.
- **Store probes.** Idleness stats depth-1 and depth-2 under each `v<N>` dir (`storeVersionDirIsIdle` + `childrenAllStale`), matching the "probe at depth 2" requirement. `nlink` scan is an exhaustive DFS over `files/**` with early exit on first `nlink > 1`; missing `files/` or any error → keep. Non-`v*` siblings that are fresh or unrecognized dirs → keep. Empty `resolveStoreVersionDirs` → keep. Default `storeRoot` is the only name reaching bare `tmpdir()`.
- **Refusal wall complete.** dirty (any status error = dirty), locked, bare, prunable, `WORKTREE_MARKER`, live-session, live-cwd (`isUnder(cwd, real)`), freshness (budget/error/depth all resolve to keep). All comparisons realpath-normalized on both sides via `resolveAllRealpaths`.
- **`store.list` field.** `worktreePath` is `TEXT NOT NULL` (`src/store.ts:368`, `:897`); the `.filter(Boolean)` in `src/index.ts:838` is harmless.

No blocking correctness, security, or scope issues found. Removal is all-or-nothing per the plan's explicit in-scope decision; partial reclaim is correctly deferred.

### Stated limitations

- The inline comment at `src/index.ts:826` references `#1874`; the issue/PR number could not be cross-checked against repo ground truth and is a comment-only reference, not a code defect.

PR #1882, opened 2026-07-21
fix(tmp): point trusted agents at a disk-backed TMPDIR (#1875)

### LINKED ISSUE #1875: Point spawned agents at a disk-backed TMPDIR (structural fix for tmpfs inode exhaustion)

scoop:
Closes #1875. Follow-up to #1862 — replaces its **advisory** "keep off the tmpfs" directive with a **structural** guarantee for trusted spawns.

## What

`buildWrappedArgv` now points every **trusted** spawned agent at a disk-backed `TMPDIR` (default `~/.cache/shepherd/tmp`, knob `SHEPHERD_AGENT_TMPDIR`, `""` = disable), so agent temp I/O — the per-session scratch tree, worktrees, dependency installs, and bare-`$TMPDIR` tool caches — lands on a real filesystem instead of the `/tmp` tmpfs whose **inode** table it exhausts (ENOSPC with bytes to spare).

## Key facts (empirically probed this session)

- **Claude prefers `CLAUDE_CODE_TMPDIR` over `TMPDIR` and appends its own `claude-$uid` to it** (double-suffix). The server sets its *own* `CLAUDE_CODE_TMPDIR` so `claudeTmpRoot()` follows agents to disk — so the shim **strips the inherited value** (`env -u CLAUDE_CODE_TMPDIR`) and the agent re-derives the *same* root from `TMPDIR` alone. Without the strip, the agent would write to `<disk>/claude-$uid/claude-$uid/…` while the server reads `<disk>/claude-$uid/…`, blanking the Files tab.
- We set `CLAUDE_CODE_TMPDIR`, **not** `TMPDIR`, on the server, so `os.tmpdir()` — and thus `readTmpInodeUsePct()`'s `tmp_inodes` monitor — keeps watching the real tmpfs.
- **Trusted spawns only:** the outer `env` shim is wiped by `bwrap --clearenv` for sandboxed spawns (exactly as `NODE_COMPILE_CACHE` is today); a sandbox's `--tmpfs /tmp` is ephemeral and never threatens the host inode table. The #1862 advisory belt is retained for agents that hardcode `/tmp`.

## Migration (adopted agents keep `TMPDIR=/tmp` across a deploy)

- **dual-read** — scratchpad read/list/upload resolve to the first *existing* of `[disk, legacy tmpfs]` (prefer disk); `scratchpadHasFiles` dual-reads.
- **dual-clean** — `removeWorktreeScratch` reclaims both roots on archival.
- **bare-root coverage** — `reapFallowCaches` + `sweepClaudeTmp` also scan the bare `agentTmpDir()`, where fallow/bunx write at `$TMPDIR` (verified: `mktemp` in a Bash tool lands there), else unbounded disk-inode accretion.

All root sets are **deduped**, so the no-override case is byte-for-byte unchanged.

## Divergence, stated & handled

Post-change `sweepClaudeTmp` gates on the disk root while `tmp_inodes` watches tmpfs. Intended: trusted agents no longer accrete tmpfs inodes, so new pressure stops. Residual tmpfs pressure = pre-existing accumulation (reclaimed by #1874); the one-click **forced** sweep still clears the legacy tmpfs root.

## Test

`bun run lint`, `bunx tsc --noEmit`, full `bun test` (7503 pass), and pre-push (7 lanes incl. fallow audit) all green. New tests: geometry helpers + disable, dual-read fallback / prefer-disk, dual-clean both roots, bare-root reap + sweep, and parity assertions for the `TMPDIR=` / `-u CLAUDE_CODE_TMPDIR` tokens on the unwrapped argv.

### DIFF
--- a/src/herdr.ts
+++ b/src/herdr.ts
@@ -3,7 +3,7 @@ import { promisify } from "node:util";
 import { execFileSync } from "./instrument";
 import { config } from "./config";
 import { maintenance } from "./maintenance";
-import { compileCacheDir } from "./tmp-sweep";
+import { compileCacheDir, agentTmpDir } from "./tmp-sweep";
 import type { HerdrState, LivenessState, SessionStatus } from "./types";
 
 const execFileAsync = promisify(execFile);
@@ -363,6 +363,16 @@ export function parseAgentInfo(agent: unknown): HerdrAgent {
  * xterm keystrokes). The main-session spawn computes its own renderer env and routes it
  * through both the membrane --setenv and this shim, so re-adding the pin here would
  * duplicate it (and `env`'s last-wins would let the pin override an intended NO_FLICKER).
+ *
+ * Points the agent at the disk-backed `agentTmpDir()` via `TMPDIR=` (#1875) so its temp I/O stays
+ * off the `/tmp` tmpfs whose inode table it exhausts, and STRIPS the inherited `CLAUDE_CODE_TMPDIR`
+ * via `env -u`: the server sets that on its OWN env so `claudeTmpRoot()` follows, but claude honours
+ * it as a BASE and appends its own `claude-$uid`, so letting it reach the agent would double-suffix
+ * the agent's root and desync it from the server's read path. Both are skipped when the redirect is
+ * disabled (`agentTmpDir()===null`); the `TMPDIR=` token is skipped if the caller already set one.
+ * NOTE: for a membrane-wrapped (sandboxed) spawn this OUTER `env` shim is wiped by `bwrap --clearenv`
+ * (identically to `NODE_COMPILE_CACHE`), so the redirect is a TRUSTED-spawn guarantee — a sandbox's
+ * own ephemeral `--tmpfs /tmp` never threatens the host inode table.
  */
 export function buildWrappedArgv(argv: string[], env?: Record<string, string>): string[] {
   const envTokens = env
@@ -372,8 +382,17 @@ export function buildWrappedArgv(argv: string[], env?: Record<string, string>): 
     : [];
   const callerSetRenderer =
     !!env && ("CLAUDE_CODE_NO_FLICKER" in env || "CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN" in env);
+  // `-u CLAUDE_CODE_TMPDIR` must precede the KEY=VALUE assignments (coreutils `env` stops option
+  // parsing at the first operand). When enabled we ALWAYS strip CLAUDE_CODE_TMPDIR (even if a caller
+  // set TMPDIR — claude prefers CLAUDE_CODE_TMPDIR over TMPDIR, so a leak would still double-suffix),
+  // and add our TMPDIR= only when the caller didn't provide one.
+  const agentTmp = agentTmpDir();
+  const tmpTokens = agentTmp
+    ? ["-u", "CLAUDE_CODE_TMPDIR", ...(env && "TMPDIR" in env ? [] : [`TMPDIR=${agentTmp}`])]
+    : [];
   return [
     "env",
+    ...tmpTokens,
     `NODE_COMPILE_CACHE=${compileCacheDir()}`,
     ...(callerSetRenderer ? [] : ["CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1"]),
     ...envTokens,
--- a/src/index.ts
+++ b/src/index.ts
@@ -114,6 +114,8 @@ import {
   reapFallowCaches,
   pruneRepoWorktrees,
   scratchpadHasFiles,
+  agentTmpDir,
+  agentClaudeTmpRoot,
 } from "./tmp-sweep";
 import { runSessionUsageBackfill } from "./usage-backfill";
 import { PreviewService } from "./preview";
@@ -781,6 +783,24 @@ try {
   mkdirSync(compileCacheDir(), { recursive: true });
 } catch (err) {
   console.warn("[tmp-sweep] could not create compile-cache dir:", err);
+}
+
+// Point spawned (trusted) agents at a disk-backed TMPDIR (#1875), and mirror it into the server's
+// OWN env so claudeTmpRoot() — and thus every scratch/sweep consumer — follows the agents to disk.
+// We set CLAUDE_CODE_TMPDIR (NOT TMPDIR) so os.tmpdir() is untouched and readTmpInodeUsePct() keeps
+// monitoring the real tmpfs. The value carries the claude-$uid suffix (claudeTmpRoot() treats
+// CLAUDE_CODE_TMPDIR as the FINAL root); the spawn shim strips it from agents (env -u) so they
+// re-derive the SAME root from TMPDIR without double-suffixing. Runs before any spawn or scratchpad
+// read. Null = disabled (SHEPHERD_AGENT_TMPDIR="") → tmpfs rollback, leave the env untouched.
+const agentTmp = agentTmpDir();
+const agentClaudeRoot = agentClaudeTmpRoot();
+if (agentTmp && agentClaudeRoot) {
+  try {
+    mkdirSync(agentTmp, { recursive: true });
+  } catch (err) {
+    console.warn("[tmp-sweep] could not create agent tmp dir:", err);
+  }
+  process.env.CLAUDE_CODE_TMPDIR = agentClaudeRoot;
 }
 
 // Inode-guard sweep: drops the compile cache + stale scratch once /tmp inode pressure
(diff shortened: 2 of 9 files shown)


scoop (commented):
## Review

The PR points trusted spawns at a disk-backed `TMPDIR` and keeps the server's own `claudeTmpRoot()` in sync via `CLAUDE_CODE_TMPDIR`, with a dual-read/dual-clean/sweep migration for adopted pre-upgrade sessions. I verified the full data path and found no blocking defects.

### What I verified (against ground truth)

- **Spawn shim geometry.** `buildWrappedArgv` emits `-u CLAUDE_CODE_TMPDIR` before the `KEY=VALUE` tokens and injects `TMPDIR=<agentTmpDir>` only when the caller didn't set one (`src/herdr.ts:389-400`). Coreutils `env` stops option parsing at the first operand, so `-u` placed ahead of assignments is honored. Verified the token order the parity test asserts (`test/herdr-parity.test.ts:117-127`) matches the array construction.
- **No double-suffix / no desync.** Server sets `process.env.CLAUDE_CODE_TMPDIR = agentClaudeTmpRoot()` = `<disk>/claude-$uid` (`src/index.ts:803`, `src/tmp-sweep.ts:91-94`); `claudeTmpRoot()` returns `CLAUDE_CODE_TMPDIR` verbatim (`src/tmp-sweep.ts:44-50`). The agent gets `CLAUDE_CODE_TMPDIR` stripped and `TMPDIR=<disk>`, re-deriving `<disk>/claude-$uid` — matching the server read path. Confirmed EVERY claude spawn routes through `start()`→`buildWrappedArgv` (both `src/herdr.ts:776` and `src/herdr-socket-driver.ts:145`); grep for other `CLAUDE_CODE_TMPDIR` consumers turns up none that spawn claude, so no path leaks the server value to an agent.
- **`os.tmpdir()` untouched.** Boot sets `CLAUDE_CODE_TMPDIR`, not `TMPDIR`, so `readTmpInodeUsePct()` keeps statfs-ing the real tmpfs (`src/tmp-sweep.ts:330-335`). Verified.
- **Sweep safety at the bare disk root.** `resolveSweepRoots` adds `agentTmpDir()` to the deduped production set (`src/tmp-sweep.ts:435-448`). Sweeping `<disk>` is safe: `sweepEntry` skips the nested `claude-$uid` dir (`===ctx.nestedName`, `src/tmp-sweep.ts:398`) preserving the scratch tree, only wholesale-removes `node-compile-cache` (a sibling, not under `<disk>` in production) and age-gated `REGENERABLE_CACHE` names (`src/tmp-sweep.ts:346,391-410`), and leaves unknown entries in place. `reapFallowCaches` at the bare root only touches `FALLOW_CACHE_PREFIX` entries (`src/tmp-sweep.ts:646-656`).
- **Dual-read resolver.** `existingScratchpadDir` returns the first existing candidate (primary/disk first), else primary — so both-exist prefers disk, legacy-only falls back to legacy (`src/tmp-sweep.ts:161-179`); all four scratchpad resolvers now route through it (`src/scratchpad.ts:38,54,67,240`), and uploads mkdir the same resolved root so reads/uploads never split. Verified against the two new scratchpad tests.
- **Ordering.** The boot override (`src/index.ts:795-804`) runs synchronously at module load, before any deferred start or request-time spawn/scratchpad read. Helpers read env at call time, matching `claudeTmpRoot()`.

### Non-blocking observations (not findings)

**Sweeper vs. diagnostics divergence (design change, documented):** After boot, `sweepClaudeTmp`'s gate measures the disk `claudeTmpRoot()` (`src/tmp-sweep.ts:471,494`) while the `tmp_inodes` diagnostics row keeps watching the tmpfs. The automatic boot/daily sweep will therefore essentially never fire on tmpfs inode pressure (the disk home fs has ample inodes). The plan states this is intended — trusted agents no longer accrete tmpfs inodes, sandbox temp is ephemeral, and residual tmpfs pressure is left to the one-click forced sweep + dual-clean + #1874. Coherent, but it is a real behavior change from today's tmpfs-gated automatic reclaim; flagging for reviewer awareness.

- src/herdr.ts:391 — when a caller sets `TMPDIR` in `env`, the shim strips `CLAUDE_CODE_TMPDIR` but skips the disk token, leaving the agent on the caller's `TMPDIR` while the server still reads `<disk>/claude-$uid` — a scratchpad desync. Dormant today: no `start()` caller passes `TMPDIR` (verified — `env` overlays carry only renderer keys, `src/service.ts:1565-1569`, `src/herdr.ts` callers). Only reachable if a future caller sets `TMPDIR`.

- Claude Code's own `TMPDIR`→`<TMPDIR>/claude-$uid` re-rooting and its `CLAUDE_CODE_TMPDIR`-as-base double-suffix behavior are asserted from the issue/plan's live probes against the `claude` binary, which is not in this repo. The server-side geometry is written consistent with those probes; I could not independently re-run them.

- Sandboxed (membrane-wrapped) spawns' temp I/O is intentionally not redirected (`--clearenv` strips the outer shim); their Files-tab visibility of scratchpad is a pre-existing property, unchanged here.

PR #1887, opened 2026-07-21
fix(herdr): guard against unsupported herdr 0.7.5+ (agent spawning broken)

scoop:
## TL;DR — herdr 0.7.5 breaks Shepherd; this protects operators from it

herdr 0.7.5 (socket protocol 17) reshaped `agent start` so that Shepherd's spawn command — an `env …` shim (always), wrapped in `bwrap … -- env … claude …` when a bwrap sandbox backend is present — **can no longer be launched**. Verified against a live 0.7.5 server (`/proc` inspection): herdr runs `claude env NODE_COMPILE_CACHE=… claude …`, i.e. the shim tokens become positional args to claude and the compile-cache/TMPDIR pinning is lost. **This is every spawn, on the CLI path 0.7.5 actually uses — not just the socket, not just sandboxed spawns.** Shepherd cannot launch a single agent on 0.7.5.

So this PR does **not** add 0.7.5 support (that needs a new spawn mechanism — see **#1889**). It pins support to **herdr ≤ 0.7.4** and makes an accidental upgrade fail *loudly and safely* instead of cryptically:

- **Detect** the installed herdr version (`src/herdr-capabilities.ts`), refreshed after a `herdr update` and by the diagnostics probe.
- **Warn loudly at startup, keep running** — `preflight` emits a prominent banner on ≥ 0.7.5 (so an operator already on it sees why, and non-spawn surfaces still respond).
- **Fail spawns clearly** — the driver's `start()` throws `HerdrSpawnUnsupportedError` on ≥ 0.7.5 (CLI + socket) instead of spawning a broken agent.
- **Block the in-app herdr updater** — `check()` flags `latestUnsupported`, `apply()` refuses, and `HerdrUpdateModal` shows a warning + hides the run button, so nobody upgrades *into* 0.7.5 from within Shepherd.
- **Diagnostics** reports an installed ≥ 0.7.5 as an error (`diagnostics_hint_herdr_unsupported`).
- **Docs** (README + getting-started) advise pinning to 0.7.4.

## Why this shape

Earlier I explored full 0.7.4/0.7.5 dual-support, but live verification showed the 0.7.5 `agent start --kind` model can't express Shepherd's sandboxed/env-wrapped spawn, so 0.7.5 is unusable regardless. Shipping a "dual-support" that can't spawn would be misleading; the honest move is to guard against it and advise pinning, then build real 0.7.5 support (with a compatible spawn path) as one coherent change in **#1889**.

## Live verification (isolated, non-destructive)

Ran an isolated 0.7.5 server on its own socket (never touched the operator's live 0.7.4). Confirmed:
- `agent start --kind claude -- <wrapped>` runs `claude <wrapped>` → env shim becomes garbage args, pinning lost (`/proc/<pid>/cmdline`).
- Strict agent names: v075 rejects `TASK-01` / names with spaces (`invalid_agent_name`).
- Targeting: `terminal_id` rejected (`agent_not_found`); only `pane_id`/name accepted.
- A freshly-created tab's pane can be briefly `agent_pane_busy`.

All four are captured in #1889 as requirements for the future spawn path.

## Verification of this PR

`tsc --noEmit`, `bun run lint`, full root `bun test` (7,557 pass / 0 fail), `cd ui && bun run check` + `check:i18n` (locales in parity), the `HerdrUpdateModal` browser test (incl. a new blocked-state case), and the `fallow audit` pre-push gate — all green. New tests cover the version ceiling, the spawn guard (refuses ≥0.7.5, spawns nothing), the update-block, and the diagnostics `unsupported` state.

Follow-up (real 0.7.5 support): **#1889**.

`[no-feature-entry]` — protective guard/warning, not a discoverable feature.


scoop (commented):
## Summary

The PR makes Shepherd drive both herdr 0.7.4 (protocol 16 / `legacy`) and 0.7.5 (protocol 17 / `v075`) from one build, auto-detecting the facade from the parsed `herdr --version` (CLI path) or negotiated socket protocol, and branching the agent CLI/socket surface accordingly. The change closely follows the approved plan. I verified every version-varying decision, method name, and param shape against the actual code and the freshly-vendored protocol-17 schema. I found no blocking defects.

## What I verified

**Facade detection & wiring**
- `facadeForVersion` maps null/unparseable → `legacy` (preserves shipping behavior), `>=0.7.5` → `v075` via `compareSemver` (`src/herdr-capabilities.ts:32-35`). `facadeForProtocol` maps `>=17` → `v075` (`src/herdr-capabilities.ts:39-41`). Boot sets it from the preflight version read (`src/index.ts:225-227`); the socket driver overrides from the negotiated protocol (`src/herdr-socket-driver.ts:383`).
- The diagnostics probe re-syncs the facade on every periodic version read, and only after a *valid* parse — `versionProbe` returns early on `!version` before calling `onVersion` (`src/diagnostics.ts:975-979`), so a transient unreadable version does NOT spuriously reset the facade to legacy. Good.
- `herdr update --handoff` re-derives the facade from the actually-installed version post-swap (`src/herdr-update.ts:301-305`). `compareSemver` correctly relocated to the leaf `src/semver.ts` and re-exported from `herdr-update.ts` for existing importers (verified `src/diagnostics.ts:23` still imports it from `./herdr-update`).

**Send-path split (legacy `send` → `prompt`/`sendKeys`)**
- Legacy `prompt` reproduces the old two-write behavior exactly: `agent send <t> PASTE_START+safe+PASTE_END` then `agent send <t> "\r"` (`src/herdr.ts:930-937`), matching the removed `service.ts` code (old `sendSteerTo`). `service.ts` now passes raw text and the driver strips markers via `stripPasteMarkers` — equivalent to the old local `replaceAll` (`src/service.ts:4434`).
- E-stop: `haltAll` now calls `sendKeys(terminalId, ["esc"])` (`src/service.ts:4370`); legacy resolves `esc → "\x1b"` via `legacyKeyByte` and writes it through `agent send` — byte-identical to the old `send(id, "\x1b")`. `sendKeys` is not routed through `#serializeSteer`, preserving the un-serialized best-effort e-stop contract.
- v075: `prompt` → `agent prompt <pane> <text>`; `sendKeys` → `agent send-keys <pane> esc` (`src/herdr.ts:927,945`). Confirmed protocol-17 defines `RequestAgentPromptParams {target,text}` (`src/generated/herdr-protocol-17.ts:1393`) and `RequestAgentSendKeysParams {keys,target}` (`:1429`), and that protocol-17 has **no** `agent.send` — so the socket legacy path's use of the untyped `requestRaw` for `agent.send` is correct (protocol-16 still carries `RequestAgentSendParams`, `herdr-protocol-16.ts:2688`).

**Spawn reshape**
- v075 `agent start <name> --kind <kind> --pane <root> -- <wrapped>` with root-pane cleanup skipped; legacy `--tab/--cwd/--no-focus` with cleanup retained (`src/herdr.ts:462-473`, `:892-914`). Socket `agent.start` v075 params `{name,kind,pane_id,args}` match `RequestAgentStartParams` (`src/generated/herdr-protocol-17.ts:1437`); legacy `{name,argv,cwd,tab_id,focus}` correctly routed through `requestRaw`. `herdrKindForArgv` mirrors `isHeadlessCodexExec`'s bwrap command-index logic (`src/herdr.ts:449-452` vs `:441-444`) and is fed the same raw `argv`, so kind derivation and cleanup gating stay consistent.

**Targeting (terminal_id → pane_id)**
- `#paneByTerminal` warmed from every `list()`/`listAsync()` and from `start` (`#rememberPanes`); `resolveTarget`/`resolveTargetSync` translate on v075 with a single refresh-on-miss and a terminal_id fallback (`src/herdr.ts:684-706`). Applied consistently across `prompt`/`sendKeys`/`read`/`readAsync`; `rename` uses the freshly-listed `agent.paneId` with a terminal_id fallback (`src/herdr.ts:1030-1033`). Verified protocol-17 `RequestReadSource` and `RequestPaneTarget` are unchanged (`herdr-protocol-17.ts:13,1814`).
- All send/prompt call sites updated: `src/service.ts:164` `Pick` now lists `prompt|sendKeys`; no stray `.send(` on a herdr driver remains in `src/` (grep confirmed only WebSocket `.send` unrelated).

**Codegen**
- `gen-herdr-schema.ts`/`gen-herdr-types.ts` correctly parameterize per-protocol, derive the filename from the schema's own `protocol` field, and assert filename/schema agreement; `check:herdr-types` now guards both `-16.ts` and `-17.ts`. No lingering import of the old `generated/herdr-protocol` path (socket client now imports `-17`).

## Non-blocking observations

- **Socket e-stop cache warmth:** the plan asserts "haltAll lists immediately before its sweep, so the cache is warm." For the *socket* driver this is not quite true — `haltAll`'s up-front `this.deps.herdr.list()` delegates to the CLI driver (`herdr-socket-driver.ts:114-116`) and warms the *CLI's* `#paneByTerminal`, not the socket driver's. The socket driver's cache is instead kept warm by the ~1s poll-loop `listAsync`. Worst case (cold socket cache) the first `sendKeys` incurs one extra `listAsync` round-trip before ESC; it self-heals after and e-stop targeting stays correct. Socket path is default-off, so this is minor.
- **Detection-failure degeneracy:** if the binary is a real 0.7.5 but its `--version` is unparseable, the facade defaults to `legacy` and would mis-drive it (`agent send`/`--tab` don't exist on 0.7.5). This is the documented "detection failure preserves shipping behavior" trade-off and an accepted degenerate case, not a regression.

## Could not verify (ground truth not in repo)

- **herdr v075 `agent start ... --kind <k> -- <wrapped argv>` semantics.** The whole scheme depends on herdr 0.7.5 executing the full wrapped argv after `--` (which includes Shepherd's `env`/bwrap sandbox wrapper) as the pane command, rather than treating `--kind` as the launcher and the trailing args as extra flags to a herdr-launched agent. If it were the latter, Shepherd's bwrap sandbox wrapper would be dropped. The legacy `--tab -- <wrapped>` path worked this way and the plan states this was probed against a live v0.7.5 binary; I cannot confirm it in-repo. Assumed correct as written.
- Live herdr 0.7.5 acceptance of pane_id (vs terminal_id) as a command `<TARGET>`, and the exact `agent read` flag surface — assumed per the changelog/plan; not verifiable in this repo.

Out of scope (pre-existing, not in this PR): none.

scoop (commented):
## Summary

Rather than the full dual-support described in the approved plan, this PR takes the narrower, safer route the commit message states: **detect herdr 0.7.5+ and refuse to spawn on it** (spawning is broken; #1889), while leaving steer/read intact. It warns at boot (`index.ts`), reports `error/unsupported` in diagnostics, throws `HerdrSpawnUnsupportedError` from both drivers' `start()`, and blocks the in-app updater (server backstop in `apply()` + UI hides the run button). A smaller diff than the plan is not a defect, and the guard is a coherent, self-consistent feature.

## Verification performed

- **`compareSemver` relocation (herdr-update.ts → semver.ts, re-exported).** Confirmed every existing importer keeps working: `src/diagnostics.ts:23`, `src/codex-update.ts:6`, `src/plugin-update.ts:7`, `deploy/provision.ts:22` all import from `./herdr-update`, which re-exports it at `src/herdr-update.ts:13`. No import cycle: `src/semver.ts` has no imports; `herdr-capabilities.ts:11` imports only `./semver`; `herdr-update.ts:6` imports `./semver`; `herdr.ts` imports `herdr-capabilities`.
- **Version ceiling window is valid.** Floor `HERDR_MIN_VERSION = "0.7.0"` (`src/config.ts:75`) < ceiling `HERDR_LAST_SUPPORTED_VERSION = "0.7.4"` (`src/herdr-capabilities.ts:100`), so 0.7.0–0.7.4 is a non-empty supported range.
- **Diagnostics ordering.** `src/diagnostics.ts:982` places the unsupported (ceiling) check after the missing check and before liveness/outdated, matching the comment's stated precedence.
- **`preflightHerdr` return-type change (void → string|null).** Only production caller is `src/index.ts:216`; the added `string|null` is consumed correctly via `parseHerdrVersion(herdrVersionRaw ?? "")`. Missing-binary still exits (`src/preflight.ts`), present-but-broken returns null → `herdrSpawnSupported()` stays true (spawn would fail on its own regardless).
- **`HerdrSpawnUnsupportedError`** is exported from `src/herdr.ts:282` and imported by `src/herdr-socket-driver.ts:3`; guard fires before any herdr call in both drivers (verified by `test/herdr.test.ts` asserting `calls === []`).
- **i18n parity.** Only two catalogs exist (`ui/messages/en.json`, `ui/messages/de.json`). All three new keys — `herdrupdate_unsupported_title`, `herdrupdate_unsupported_body`, `diagnostics_hint_herdr_unsupported` — are present in both (en.json:15-16,1341; de.json:17-18,1341). The `{latest}` placeholder in `herdrupdate_unsupported_body` matches the call site `HerdrUpdateModal.svelte:150` (`{ latest: update.latest ?? "" }`); `diagnostics_hint_herdr_unsupported` takes no args and is passed as a `hintKey`, consistent with sibling keys.
- **Modal dismissability when blocked.** `HerdrUpdateModal.svelte:227-236`: with `latestUnsupported`, `!busy` shows the close/later button while `!update.latestUnsupported` hides `.run` — user can still dismiss.
- **`HerdrUpdateStatus.latestUnsupported?`** added consistently in both `src/types.ts:406` and `ui/src/lib/types.ts:1454`.

## Minor observation (non-blocking, not a finding)

- `src/herdr-update.ts:323` computes `latestUnsupported: !isHerdrVersionSupported(to)` in the post-apply status without the `updateAvailable &&` gate that `check()` uses at line 364. In practice this is harmless: `apply()` refuses unsupported targets up front, and `isHerdrVersionSupported(null)` returns true, so this branch always yields `false`. Cosmetic inconsistency only.

Issue #1889, opened 2026-07-21
herdr 0.7.5 (protocol 17): agent spawning via `agent start --kind` is incompatible with Shepherd's sandboxed/wrapped spawn command


scoop:
Tracks building **real herdr 0.7.5 (protocol 17) support**. PR #1887 does NOT add it — it gates 0.7.5 off (Shepherd warns, refuses spawns, blocks the in-app updater) because 0.7.5 broke agent spawning. This issue is the path to lifting that gate.

## The blocker (empirically confirmed against a live 0.7.5 server)

v075 `agent start <name> --kind <exe> --pane <id> -- <args>` runs **`<canonical-exe> <native-args>`** — `--kind` picks the executable from a fixed enum (claude/codex/…) and `-- args` are appended as native args. Proven via `/proc/<pid>/cmdline`: passing Shepherd's spawn argv made herdr run `claude env NODE_COMPILE_CACHE=… claude …` (the `env` shim became positional args; the compile-cache/TMPDIR pinning was lost).

Shepherd's spawn argv is `buildWrappedArgv()` output — an `env` coreutils shim (`env -u CLAUDE_CODE_TMPDIR TMPDIR=… NODE_COMPILE_CACHE=… CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1 … <argv>`), and when a bwrap backend is available (`src/sandbox.ts:602`) the whole thing is further wrapped as `bwrap …membrane… -- env … claude …`. v075's `--kind` model can express **none** of that. This breaks EVERY spawn (the `env` shim alone is enough), on the CLI path 0.7.5 uses — not just sandboxed spawns, not just the socket.

## Candidate approach to investigate

Bypass `agent start` on v075:
1. `tab create --cwd … --env KEY=VALUE …` (the `--env` flag exists) to carry the env pins onto the tab's shell.
2. Type the full `bwrap … -- env … claude …` command into the tab's root shell pane via `pane send-text`.
3. Poll `agent list` until herdr's **passive** detection recognizes the agent; capture its `pane_id`.
4. `agent rename <pane_id> <sanitized-name>` to set Shepherd's name.

Open questions: does passive detection reliably fire for a manually-launched (sandboxed) agent? How is the spawn handle resolved without `agent start`'s reply? The bwrap membrane is a process wrapper, not env — so `tab create --env` covers the pins but not the sandbox; that layer may need a herdr feature (run an arbitrary command under a declared kind) or a different sandboxing seam.

[...]

PR #1897, opened 2026-07-22
feat(herdr): push derived lifecycle state to herdr for sandboxed 0.7.5 agents (#1891)

### LINKED ISSUE #1891: herdr 0.7.5: lifecycle-state push + reconcile/husk correctness for sandboxed agents
Parent epic: #1889. **Blocked by the 0.7.5 CLI spawn-path child** (assumes that path has landed).

## Goal

Make Shepherd's session-liveness + status machinery correct for **externally-registered, sandboxed**
0.7.5 agents — the ones herdr cannot observe on its own.

## Why (spike-confirmed)

Under the membrane's `--unshare-pid`, the pane's foreground process is `bwrap`, and `claude` lives in
a PID namespace herdr can't see. So herdr will **not** advance `agent_status` for a sandboxed agent —
it stays at whatever was last reported. Shepherd already derives agent state itself (its blocked/idle
classifier scrapes the terminal), so it must **push** that state, and its reconcile/husk heuristics
(which today lean on the foreground process being `claude`) must be re-checked against a `bwrap`
foreground.

## Vertical cut (scope)

- Push status transitions to herdr via `pane report-agent --state <derived>` whenever Shepherd's classifier updates a 0.7.5 session's state, so `agent list` (and the herdr UI) reflect reality. Report on the same tick the state is computed (no batching).
- Audit and fix the reconcile/liveness helpers (the agent↔session matcher, the stranded/husk classifier, and the pane-foreground-process reader) for the case where the foreground process is `bwrap` rather than the agent binary — a sandboxed live agent must not be misread as a husk, and a genuinely dead one must still be caught.

## Out of scope

- The spawn path itself (its own child, assumed landed).
- The socket driver (its own child).
- Retiring the version ceiling / updater / docs (its own child).
- Non-sandboxed (trusted) spawns where the foreground already is the agent binary — unchanged.

## Testing seam

Unit-test the classifier→`report-agent` push and the `bwrap`-foreground variants of the stranded/husk
+ matcher helpers with **fake `agent list` + `pane process-info` fixtures** (the helpers are already
pure functions over those shapes). Tests in `test/`.

## Acceptance

- [ ] A running sandboxed 0.7.5 agent that goes idle/blocked has its `agent_status` reflected in `agent list` within one classifier tick.
- [ ] A live sandboxed agent (foreground `bwrap`) is never classified as a stranded husk; a truly dead one still is.
- [ ] `bun install && bun run lint && bun test` pass.

scoop:
Part of epic #1889; closes #1891. Base: `epic/1889-herdr-0-7-5-protocol-17-agent-spawning-v`.

## What & why

herdr freezes `agent_status` for externally-registered, **sandboxed** 0.7.5 agents — its pane/PID view is the `bwrap` monitor, not the agent binary, so it can't advance the state it was handed at registration. `agent list` (and the herdr UI) then shows `working` forever. Shepherd already derives block/idle/working from the terminal + hook signals; this PR **pushes** that derived state back to herdr.

## Part A — lifecycle-state push (the actual change)

- **`deriveHerdrState({blocked, working}) → RequestPaneAgentState`** (`idle|working|blocked`) — the pure classifier→state seam. Typed to the generated protocol union (`src/generated/herdr-protocol.ts`), which excludes `done`, so an invalid state can't be pushed.
- **`HerdrDriver.reportAgentState(paneId, agentName, state)`** — issues `pane report-agent <paneId> --source shepherd --agent <name> --state <state>` (mirrors the register pair). Added to `IHerdrDriver` + the socket driver (interface completeness; the socket driver never hosts a 0.7.5 external-registration spawn).
- **Poller wiring** — the push is driven from the **emission sinks** (`emitBlock`, `emitActivity`) plus an end-of-`reconcileAgent` idle-timeout check, funnelled through one idempotent, push-on-change `maybePushAgentState`:
  - A hook-driven awaiting-input `blocked` edge early-returns at `tryHookAwaitingBlock`; driving the push from `emitBlock` means it still lands **same-tick**, not a tick late.
  - working/idle derive from a `lastActiveTurnAt` stamp set in `emitActivity` (the `maybeProbe` path that runs on the frozen-`working` route — **not** `maybeClassify`), OR-ed with hook freshness. A never-observed session reads `working` (matches the registration baseline) to avoid a startup `working→idle→working` flap.
  - Pushes serialized per session (can't reorder on herdr); best-effort, never throws in the 1s tick; a failed report rolls back so the next tick retries (no failure storm).
- **Gate:** the 0.7.5 external-registration path **and** a genuinely-sandboxed spawn (`sandboxApplied ∈ {standard, autonomous} && !sandboxDegraded`). ≤0.7.4 and `trusted`/degraded spawns are unchanged.

## Part B — reconcile/husk correctness: empirically DROPPED (no source change)

The plan gated Part B behind a HALTing Phase-0 empirical check, because the "herdr can't see `claude` under `--unshare-pid`" premise applies to herdr's *pane* detection, **not** `scanClaudeAliveByWorktree`'s *host-wide* `/proc` scan. Reproduced the membrane process tree with `bwrap 0.11.1` and the faithful `buildMembraneFlags` set:

- **Live:** the inner `claude` is host-visible with `comm=claude` and `/proc/<pid>/cwd` resolving under the worktree (alongside two `bwrap` monitors). `--unshare-pid` does **not** hide a descendant from the host, and the worktree is bind-mounted at the same path. → `scanClaudeAliveByWorktree` already marks a live sandboxed agent **alive**. No husk false-positive exists.
- **Dead:** killing only the inner `claude` (parent shell kept alive) → the `bwrap` monitors reap it and exit; **no worktree-cwd `bwrap`/`claude` remains** → the scan reads `false` → still a husk. (`--die-with-parent` governs parent death; the child-exit teardown was confirmed empirically, not assumed.)

Both halves of the acceptance already hold with existing code, so the `scanClaudeAliveByWorktree` change would be a no-op — **dropped**. Broadening `AGENT_COMMS` to include `bwrap` would be *wrong*: it's the SIGKILL-sparing prefilter for the orphan reapers, and a genuinely dead/orphaned `bwrap` must stay reap-able (a live `bwrap` is already spared there by the PPID-1 / deleted-cwd / archived-terminality / listening-port gates). Instead: **regression fixtures** lock the proven behavior.

## Tests

- `deriveHerdrState` truth table; `reportAgentState` argv; `paneForegroundProcs`/`reportAgentState` shapes.
- `test/poller-report-agent.test.ts` (new): a session pinned to herdr `working` for the whole run pushes `working→idle` and `blocked` (same-tick, via the awaiting-input early-return), plus push-on-change dedup and gate-off (≤0.7.4, trusted, degraded).
- Part B regression: `scanClaudeAliveByWorktree` alive for a live sandboxed tree, husk for a dead one; `bwrap`-foreground reads alive in `isSpawnAlive` / `classifyHelperTab`.
- `bun install && bun run lint && bun run typecheck && bun run test` all pass (7578 pass, 0 fail).

### DIFF
--- a/src/herdr-socket-driver.ts
+++ b/src/herdr-socket-driver.ts
@@ -17,6 +17,7 @@ import {
   type HerdrTab,
   type IHerdrDriver,
 } from "./herdr";
+import type { RequestPaneAgentState } from "./generated/herdr-protocol";
 import { config, HERDR_SOCKET_SUPPORTED_PROTOCOLS } from "./config";
 import {
   detectedHerdrVersion,
@@ -262,6 +263,25 @@ export class SocketHerdrDriver implements IHerdrDriver {
     } catch {
       /* best-effort; tab may already be gone */
     }
+  }
+
+  /**
+   * Interface completeness (issue #1891): the lifecycle-state push targets externally-registered
+   * sandboxed 0.7.5 agents, which the socket driver never hosts (it refuses the 0.7.5
+   * external-registration spawn branch in `start`). Implemented as a real `pane.report_agent`
+   * request nonetheless so the socket driver is a faithful `IHerdrDriver`.
+   */
+  async reportAgentState(
+    paneId: string,
+    agentName: string,
+    state: RequestPaneAgentState,
+  ): Promise<void> {
+    await this.client.request("pane.report_agent", {
+      pane_id: paneId,
+      source: "shepherd",
+      agent: agentName,
+      state,
+    });
   }
 }
 
--- a/src/herdr.ts
+++ b/src/herdr.ts
@@ -11,6 +11,7 @@ import {
 import { maintenance } from "./maintenance";
 import { compileCacheDir, agentTmpDir } from "./tmp-sweep";
 import type { HerdrState, LivenessState, SessionStatus } from "./types";
+import type { RequestPaneAgentState } from "./generated/herdr-protocol";
 
 const execFileAsync = promisify(execFile);
 
@@ -118,6 +119,31 @@ export function mapState(s: HerdrState): SessionStatus {
     default:
       return "idle";
   }
+}
+
+/** The reachable subset of {@link RequestPaneAgentState} Shepherd derives for a LIVE session — never
+ *  `done` (not a valid report-agent state) nor `unknown` (a live session always resolves to one of
+ *  these three). */
+export type PushableAgentState = "idle" | "working" | "blocked";
+
+/**
+ * Map Shepherd's own per-tick classifier view to the state to push to herdr via `pane report-agent`
+ * (issue #1891). herdr freezes `agent_status` for externally-registered, sandboxed 0.7.5 agents (its
+ * pane/PID view is `bwrap`, not the agent binary), so Shepherd must report the state IT derives from
+ * the terminal + hook signals. Pure so the classifier→push seam is unit-testable in isolation.
+ *
+ *  - `blocked` — a block reason is currently surfaced for the session (menu/y-n/awaiting-input/stall/
+ *    quota); takes precedence, since a block is the actionable state regardless of turn activity.
+ *  - `working` — no block AND a fresh active-turn signal (a live turn's spinner/transcript still
+ *    advancing, or a fresh hook activity).
+ *  - `idle` — no block AND no fresh activity (the agent finished its turn / is resting).
+ */
+export function deriveHerdrState(input: {
+  blocked: boolean;
+  working: boolean;
+}): PushableAgentState {
+  if (input.blocked) return "blocked";
+  return input.working ? "working" : "idle";
 }
 
 /**
@@ -656,6 +682,10 @@ export interface IHerdrDriver {
   tabsAsync(): Promise<HerdrTab[]>;
   panes(): HerdrPane[];
   paneForegroundProcs(paneId: string): Promise<string[]>;
+  /** Push a Shepherd-derived lifecycle state to a pane's registered agent (`pane report-agent
+   *  --state`, issue #1891). Used for externally-registered sandboxed 0.7.5 agents whose
+   *  `agent_status` herdr cannot advance on its own. */
+  reportAgentState(paneId: string, agentName: string, state: RequestPaneAgentState): Promise<void>;
   /** Spawn an agent (issue #1553: async — socket-backed when `SHEPHERD_HERDR_SOCKET=1`,
    *  else non-blocking CLI). Returns the started `HerdrAgent`. */
   start(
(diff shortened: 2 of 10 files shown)


scoop (commented):
## Summary

This PR implements the two halves of #1891: (A) push Shepherd's derived lifecycle state to herdr via `pane report-agent --state` for externally-registered sandboxed 0.7.5 agents, and (B) reconcile/husk correctness under a `bwrap` foreground. Part B correctly collapsed to regression fixtures only — the Phase-0 empirical capture (documented in `test/process-reaper.test.ts:382`) landed on outcome (i): the inner `claude` is host-visible with `comm=claude` and a worktree-resolving cwd, so `scanClaudeAliveByWorktree` already works and `src/process-reaper.ts` is (correctly) untouched. Part A is implemented as designed. I found no blocking issues.

## Verification performed

- **CLI arg shape** of the new `HerdrDriver.reportAgentState` (`src/herdr.ts:789-799`) mirrors the register pair `registerAgentWithCollisionRetry` (`src/herdr.ts:1021-1031`) exactly: `pane report-agent <paneId> --source shepherd --agent <name> --state <state>`, paneId as leading positional. Consistent.
- **Type narrowing:** `PushableAgentState = idle|working|blocked` (`src/herdr.ts` `deriveHerdrState`) is a subset of `RequestPaneAgentState = idle|working|blocked|unknown` (verified against `src/generated/herdr-protocol.ts:307`); `done` (in `HerdrState`) is type-excluded, matching the plan.
- **Socket driver** `reportAgentState` sends `pane.report_agent` with `{pane_id, source, agent, state}` — matches `RequestPaneReportAgentParams` required fields (`src/generated/herdr-protocol.ts:1456-1465`) and the valid request key at `src/generated/herdr-protocol.ts:2714`. Never reached on the CLI-only 0.7.5 path but a faithful implementation.
- **Interface completeness:** `reportAgentState` added to `IHerdrDriver` (`src/herdr.ts:685`) and implemented on both `HerdrDriver` (`src/herdr.ts:774`) and `SocketHerdrDriver` (`src/herdr-socket-driver.ts:270`); the sole production construction site (`src/index.ts:1014`, `herdr: IHerdrDriver`) is satisfied.
- **Gate:** `isSandboxedSession` (`src/poller.ts:110`) matches `SandboxProfile` (`src/sandbox.ts:35`) and the `Session` fields `sandboxApplied`/`sandboxDegraded` (`src/types.ts:101-103`); combined with `herdrUsesExternalRegistrationSpawn()` (`src/herdr-capabilities.ts:72`) this correctly restricts the push to genuinely-sandboxed 0.7.5 spawns, as the plan settled (avoids racing herdr on a trusted foreground).
- **Same-tick blocked push:** all block fire/clear paths route through `emitBlock` (`src/poller.ts:1218`), which now calls `maybePushAgentState` — verified `fireStall` (1318), `clearStall` (1311), and `clearBlock` (1448) all funnel through `emitBlock`, so a hook-driven awaiting-input block that early-returns at `tryHookAwaitingBlock` (`src/poller.ts:771`) still pushes `blocked` on the same tick (the emit fires before the early return). The end-of-`reconcileAgent` push (`src/poller.ts:780`) catches the working→idle timeout the emission sinks can't.
- **Push-on-change + in-flight serialization** (`maybePushAgentState`): dedup check precedes the `pushInFlight` guard; a state change during a slow push is deferred (guard) then re-applied by the success `.then` re-invoke, converging without reorder; error path rolls back `lastPushedState` so a later tick retries. Fire-and-forget `.catch` never throws in `tick()`.
- **Prune:** `lastPushedState`, `lastActiveTurnAt`, `pushInFlight` are added to both the prune key-union (`src/poller.ts:905`) and the delete block (`src/poller.ts:936`) — no map leak.
- **Test-double compatibility:** the widened poller constructor `Pick<..., "reportAgentState">` does not break the other `poller-*.test.ts` files — each passes its `withListAsync` input via `as any`/`as never`, collapsing the return type; only `poller.test.ts` (typed) and `poller-listasync.test.ts` (explicit Pick) needed the update, and both got it.

## Non-blocking observations

- The working→idle derivation has an inherent ~`2 × probeCheckMs` (14s default) quiet window before `idle` is derived (`isActiveTurnFresh`, `src/poller.ts:1250`). This is the classifier's own latency, not a push delay — once idle is derived the push lands that tick — and matches the plan's explicit design, so acceptance #1 ("within one classifier tick" of the derived transition) is met. Noted for the reader, not a defect.
- On a persistent `reportAgentState` failure the error path re-arms every 1s tick and `console.warn`s each time. This is consistent with the existing `herdr list`-failure warn cadence and is a narrow window (a failing report while `listAsync` still succeeds), so it's acceptable as-is.

Could not exercise the live herdr daemon here; correctness of the push against a real `agent list` is asserted by the plan's Phase-0 Step C capture (external, not in this repo). The protocol type, arg shape, and gating are all verified against in-repo ground truth above.

PR #1895, opened 2026-07-22
feat(herdr): spawn on herdr 0.7.5 via the cli external-registration path

scoop:
Part of epic #1889. Closes #1890. **DAG root — first child to drain.**

## What

Gives Shepherd a herdr **0.7.5 (protocol 17)** agent-spawn path on the **CLI driver** via herdr's
**external-registration** seam, and routes 0.7.5 spawns through it — lifting the hard spawn refusal
#1887 added (for the CLI path). herdr 0.7.5's `agent start --kind` can't express Shepherd's
`env`-shim + `bwrap`-sandbox argv wrap; `pane run` runs the wrapped argv verbatim instead.

All spawns still funnel through the single `IHerdrDriver` seam — this is a **version-branched
reimplementation** of `start`/`read`/`send`/`relabel`. **`≤0.7.4` behavior is byte-identical.**

## How (spike-confirmed against a live 0.7.5)

**`start` (0.7.5):** `ensureWorkspace` → `tab create` → shell-readiness wait → `pane run <pane_id>
<wrapped argv>` → `pane report-agent-session` + `pane report-agent` (so `agent list` surfaces the
agent — the membrane's `--unshare-pid` hides `claude` from passive detection) → resolve the started
`HerdrAgent` from `agent list` by `pane_id`. Dedicated-tab + `TabLedger` teardown preserved; the
tab's root pane IS the agent pane (no leftover shell pane to close).

**`read`/`send`/`relabel` (0.7.5):** resolve `terminal_id`→`pane_id` (`terminal_id`/label are
rejected as `agent_not_found`). `read` keeps the same `agent read` command with a `pane_id` target;
`send` goes through `pane send-text` / `pane send-keys Enter` (`agent send` returns
`agent_not_ready` on a registered-but-undetected sandboxed agent); `agent rename` targets the
`pane_id` with a sanitized label. `stop` is unchanged (tab-close on `tab_id`, version-agnostic).

**New pure leaves** (unit-tested): `sanitizeHerdrAgentName()` enforces herdr's
`^[a-z][a-z0-9_-]{0,31}$` (`TASK-01` → `task-01`); `resolvePaneId()`; `classifyPaneWrite()`.

**Gate:** `herdrSpawnSupported()` is decoupled from the `0.7.4` support ceiling — a new spawn
ceiling (`0.7.5`) + `herdrUsesExternalRegistrationSpawn()`. `HERDR_LAST_SUPPORTED_VERSION` (`0.7.4`)
[...]

Issue #1892, opened 2026-07-21
herdr 0.7.5: socket driver protocol-17 branch + schema/type codegen


scoop:
Parent epic: #1889. **Blocked by the 0.7.5 CLI spawn-path child.** Lands **independently** — nothing
depends on it, and it does not gate lifting the CLI guard (the socket driver is default-off).

## Goal

Bring the **default-off socket driver** to herdr protocol 17: regenerate the vendored schema/types
from a real 0.7.5, accept protocol 17, and branch the socket spawn/write methods to mirror the CLI
0.7.5 path.

## Why (spike-confirmed p16 → p17 diff — small)

Removed: `agent.send`. Added: `agent.send_keys`, `agent.prompt`, `agent.wait`, `agent.view.set/clear`,
`pane.graphics.*`, `popup.close`, `workspace.report_metadata`. The registration methods
(`pane.report_agent`, `pane.report_agent_session`, `pane.release_agent`) and `pane.run` already exist
in p16. So the socket migration is mostly: accept 17, swap the send surface, add the spawn/register
calls, regen types.

## Vertical cut (scope)

- **Regenerate the vendored schema + types from a real 0.7.5** (procedure below), then add `17` to the socket driver's supported-protocol set so it activates on a 0.7.5 server.
- Branch the socket driver's spawn + write methods to mirror the CLI 0.7.5 child: `pane.run` to launch the wrapped argv, `pane.report_agent_session` + `pane.report_agent` to register, and the send surface `agent.send` → `pane.send_text` + `agent.send_keys`.

### Regen procedure / provenance (reproduce exactly)

1. Obtain a 0.7.5 binary — it is NOT on the stable channel while Shepherd pins 0.7.4. Use the **preview channel** (`herdr channel set preview && herdr update`) or the installer (`https://herdr.dev/latest.json`, `herdr-install.sh`).
2. `HERDR_BIN=/path/to/herdr-0.7.5 bun run gen:herdr-schema` — writes the vendored `herdr-schema.json` (dumps the binary's **bundled** schema; it does NOT connect to any server).
3. `bun run gen:herdr-types` — reads the vendored schema → regenerates the protocol types.
4. `bun run check:herdr-types` — asserts the committed types match the regenerated ones (no drift).

**Never** point a 0.7.5 binary at the live daemon socket (`~/.config/herdr/herdr.sock`) — that daemon is 0.7.4 and backs live sessions. Regen needs no running server.

## Out of scope

[...]

PR #1899, opened 2026-07-22
feat(herdr): drive the socket driver on herdr 0.7.5 (protocol 17) (#1892)

scoop:
Brings the **default-off** socket driver (`SocketHerdrDriver`) to herdr **protocol 17**, closing #1892 (epic #1889). Regenerates the vendored schema/types from a real 0.7.5 and branches the socket driver's spawn/write surface to the 0.7.5 external-registration path, mirroring the already-merged CLI child (#1895).

## What changed

- **Regen from 0.7.5** — `src/generated/herdr-schema.json` + `herdr-protocol.ts` regenerated from a real `herdr 0.7.5` (protocol 17). Provenance: downloaded the `v0.7.5` `linux-x86_64` release asset to a scratch path and ran `HERDR_BIN=… bun run gen:herdr-schema` (dumps the binary's **bundled** schema offline — never touched the live 0.7.4 daemon or its socket). `bun run check:herdr-types` is clean; gen is deterministic.
- **Accept protocol 17** — `HERDR_SOCKET_SUPPORTED_PROTOCOLS` → `{16, 17}`, so the socket driver activates on a 0.7.5 server.
- **`start()` 0.7.5 branch** (`startImpl075`) — `tab.create` → type the wrapped argv into the tab's root pane → register (`pane.report_agent_session` + `pane.report_agent`) → resolve the agent from `agent.list` by `pane_id` → record the `TabLedger` handle. No `pane.close` (the run reuses the root pane). Rolls back the orphan tab on any failure. ≤0.7.4 keeps its legacy `agent.start` path.
- **`send()`** — version-branched exactly like the CLI: `pane.send_text`/`pane.send_keys` against the resolved `pane_id` on 0.7.5; `agent.send` on ≤0.7.4.
- **`read`/`relabel`** — target the resolved `pane_id` on 0.7.5 (sanitized rename label), identity on ≤0.7.4.

## Two things worth a reviewer's eye

1. **No socket pane-run RPC in p17.** The regenerated schema confirms herdr's protocol-17 socket API exposes **no** direct pane-run method (unlike the CLI `pane run`). So the 0.7.5 socket spawn types the wrapped argv into the pane's ready shell as a POSIX-quoted command line (`posixShellJoin`) + Enter. This routes through the interactive shell rather than a direct exec; per-token single-quoting keeps paths/metacharacters robust. Only the default-off 0.7.5 socket path reaches it.
2. **`requestLegacy` escape.** The vendored types now track p17, but the driver still supports both protocols. Protocol 17 **removed** `agent.send` and **reshaped** `agent.start` incompatibly, so those two ≤0.7.4-only calls are no longer typeable via `request()`. `HerdrSocketClient.requestLegacy()` carries them untyped — keeping the concession in one place (the protocol-agnostic transport) so **≤p16 stays byte-identical on the wire** and everything on the 0.7.5 path + every unchanged shared method stays fully typed.

The `feat` duplication warning (socket `startImpl075`/register mirror the CLI siblings across transports) is expected — the two speak different transports (argv vs JSON-RPC), so the shared logic is the pure helpers already extracted into `herdr.ts`.

## Testing

- New `test/herdr-socket-075.test.ts`: a fake socket client routes the captured p17 fixtures (`test/fixtures/herdr-responses/v0.7.5/`) by method; asserts the spawn method sequence + shell-quoted argv, `send`/`read`/`relabel` pane_id targeting, ledger record→stop, rollback, and protocol-17 selection.
- `test/herdr-socket-driver.test.ts`: fakes updated for `requestLegacy`; existing ≤p16 assertions unchanged (still green) — proves ≤p16 is unaffected.
- Added a focused `posixShellJoin` unit test (spaces/metacharacters/embedded-quote/empty).
- `bun run lint`, `bun run typecheck`, `bun run check:herdr-types`, and the root test suite (pre-push `root-tests` lane) all pass. The pre-push `ui` lane hit a pre-existing sub-pixel flake in `TopBar.browser.test.ts` (`51.99998 ≥ 52`) unrelated to this server-only change; passed on retry.

## Scope

Default-off behavior and ≤p16 servers are unaffected. Out of scope (per the issue): flipping the socket default-on, the unused p17 methods (`pane.graphics.*`, `agent.view.*`, `popup.close`, `agent.prompt/wait`), and lifting the CLI/updater version ceiling.

[no-feature-entry] — server-only, default-off plumbing; no user-facing UX.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -76,7 +76,9 @@ export const HERDR_MIN_VERSION = "0.7.0";
 // herdr's native Unix-socket JSON-RPC protocol is versioned and still preview-unstable
 // (issue #1529) — admit only protocol numbers we've actually validated against a live
 // herdr, never an open `>=` floor. Extend this set explicitly as new protocols are verified.
-export const HERDR_SOCKET_SUPPORTED_PROTOCOLS = new Set([16]);
+// 17 = herdr 0.7.5: the socket driver branches its spawn/drive surface to the 0.7.5
+// external-registration path (#1892), mirroring the CLI driver.
+export const HERDR_SOCKET_SUPPORTED_PROTOCOLS = new Set([16, 17]);
 // TTL backing DiagnosticsService.current() — a request without ?refresh=1 reads
 // this cache. Matches the existing CountsService/backlog 60s TTL.
 export const DIAGNOSTICS_TTL_MS = 60_000;
--- a/src/generated/herdr-protocol.ts
+++ b/src/generated/herdr-protocol.ts
@@ -3,19 +3,112 @@
 // Regenerate: bun run gen:herdr-types
 /**
  * This interface was referenced by `HerdrProtocol`'s JSON-Schema
+ * via the `definition` "RequestAgentStatus".
+ */
+export type RequestAgentStatus = "idle" | "working" | "blocked" | "done" | "unknown";
+/**
+ * This interface was referenced by `HerdrProtocol`'s JSON-Schema
  * via the `definition` "RequestReadSource".
  */
 export type RequestReadSource = "visible" | "recent" | "recent_unwrapped" | "detection";
 /**
  * This interface was referenced by `HerdrProtocol`'s JSON-Schema
- * via the `definition` "RequestSplitDirection".
+ * via the `definition` "RequestAgentViewBuiltinField".
  */
-export type RequestSplitDirection = "right" | "down";
+export type RequestAgentViewBuiltinField =
+  "status" | "workspace_id" | "tab_id" | "pane_id" | "agent" | "seen" | "state_change_seq";
 /**
  * This interface was referenced by `HerdrProtocol`'s JSON-Schema
- * via the `definition` "RequestAgentStatus".
+ * via the `definition` "RequestAgentViewBuiltinSortField".
  */
-export type RequestAgentStatus = "idle" | "working" | "blocked" | "done" | "unknown";
+export type RequestAgentViewBuiltinSortField =
+  | "workspace_order"
+  | "tab_order"
+  | "pane_order"
+  | "attention"
+  | "status"
+  | "agent"
+  | "seen"
+  | "state_change_seq";
+/**
+ * This interface was referenced by `HerdrProtocol`'s JSON-Schema
+ * via the `definition` "RequestAgentViewContext".
+ */
+export type RequestAgentViewContext = "current_workspace_id" | "current_tab_id";
+/**
+ * This interface was referenced by `HerdrProtocol`'s JSON-Schema
+ * via the `definition` "RequestAgentViewField".
+ */
+export type RequestAgentViewField =
+  | RequestAgentViewBuiltinField
+  | {
+      token: string;
+      [k: string]: unknown;
+    };
+/**
+ * This interface was referenced by `HerdrProtocol`'s JSON-Schema
+ * via the `definition` "RequestAgentViewFilter".
+ */
+export type RequestAgentViewFilter =
+  | {
+      filters: RequestAgentViewFilter[];
+      op: "all";
+      [k: string]: unknown;
+    }
+  | {
+      filters: RequestAgentViewFilter[];
+      op: "any";
+      [k: string]: unknown;
+    }
+  | {
+      filter: RequestAgentViewFilter;
+      op: "not";
+      [k: string]: unknown;
+    }
+  | {
+      field: RequestAgentViewField;
+      op: "eq";
+      value: RequestAgentViewValue;
+      [k: string]: unknown;
+    }
+  | {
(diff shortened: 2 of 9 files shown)


scoop (commented):
## Summary

This PR ports the CLI's herdr 0.7.5 (protocol-17) external-registration spawn/drive path to the default-off `SocketHerdrDriver`, regenerates the vendored schema/types (protocol 17), and adds 17 to the socket supported-protocol set. The socket branch mirrors `HerdrDriver` (`src/herdr.ts`) closely, and I verified every method/param it emits against the regenerated protocol types.

## Verification performed

**Protocol constants align.** `HERDR_PROTOCOL = 17` (src/generated/herdr-protocol.ts:2949), `schema.protocol` = 17 (src/generated/herdr-schema.json:3), and `HERDR_SOCKET_SUPPORTED_PROTOCOLS = {16, 17}` (src/config.ts:81). The `selectHerdrDriver` protocol-17 activation test matches the existing p16 one.

**Every typed `request()` method used on the 0.7.5 path is a valid `HerdrMethod` with a matching param shape:**
- `tab.create` → `RequestTabCreateParams { cwd?, label?, focus? }` (herdr-protocol.ts:1988) ✓ called `{cwd, label, focus:false}`
- `pane.send_text` → `{ pane_id, text }` (herdr-protocol.ts:1781) ✓
- `pane.send_keys` → `{ pane_id, keys }` (herdr-protocol.ts:1773) ✓
- `pane.report_agent_session` → `{ pane_id, source, agent, agent_session_id? }` (herdr-protocol.ts:1719) ✓
- `pane.report_agent` → `{ pane_id, source, agent, state }` with `state: RequestPaneAgentState` including `"working"` (herdr-protocol.ts:1705, :413) ✓
- `agent.read` → `{ target, source, lines?, format? }`, `RequestReadSource` includes `visible`/`recent` (herdr-protocol.ts:1410, :13) ✓
- `agent.rename`, `tab.rename`, `pane.close`, `workspace.list/create`, `agent.list`, `tab.list` all present in the request map (herdr-protocol.ts:2962-3023) ✓

**Legacy escape is genuinely required and correctly scoped.** `agent.send` is absent from the p17 request map (confirmed — not in herdr-protocol.ts:2955-3025), and `RequestAgentStartParams` was reshaped to `{ args?, kind, name, pane_id, timeout_ms? }` (herdr-protocol.ts:1437), incompatible with the p16 `{ name, argv, cwd, tab_id, focus }` shape. Both are routed through the new untyped `requestLegacy` (herdr-socket-client.ts:70) only on the ≤0.7.4 arm, keeping the p16 wire behavior byte-identical while every 0.7.5-path and shared call stays typed via `request()`. `requestLegacy` and `request` share the refactored `roundTrip` core — the maintenance guard, id, timeout, NDJSON framing, and error/result handling are all preserved (herdr-socket-client.ts:82-160).

**No socket run RPC exists in p17** (no `pane.run`/`pane.exec` in the request map), so the send_text+Enter fallback (`runInReadyPane`, herdr-socket-driver.ts:245) is the correct approach per plan D4. `posixShellJoin` (herdr.ts:464) single-quotes each token; the escape sequence and empty-token handling are correct and covered by test/herdr.test.ts.

**Branch mirrors the CLI reference faithfully:** `startImpl075` (tab.create → runInReadyPane → registerAgentWithCollisionRetry → resolve by pane_id → ledger.record → rollback via closeTab) matches `HerdrDriver.startImpl075` (herdr.ts:940); `send`/`resolveTarget`/`relabel` mirror herdr.ts:1124/1106/1209. The `read`/`list` sync paths delegate to the CLI driver (herdr-socket-driver.ts:105-118), which is already 0.7.5-aware, so no sync-path gap.

**No test breakage from the `requestLegacy` addition.** The existing socket-driver tests were updated to add `requestLegacy` to their fake clients; test/herdr-parity.test.ts uses a `request`-only mock but only exercises `listAsync()` (which uses `request`), so it is unaffected (herdr-parity.test.ts:43-45). The existing ≤p16 `send()`/`agent.start` tests route both spies through the same impl and stay green.

## Non-blocking observations

- src/herdr-socket-driver.ts:245 (`runInReadyPane`): the bounded retry wraps only `pane.send_text`, which against a live socket essentially never fails (it just writes bytes to the PTY), so the "ride out shell-not-ready" retry is weaker than the CLI's `pane run` retry (which retries the actual exec). In practice PTY input is buffered until the shell reads it, so this is not a correctness bug, and the path is default-off — the author already documents the tradeoff in the comment. Called out only for awareness.

## Limitations

- Could not verify that `src/generated/herdr-schema.json` / `herdr-protocol.ts` were dumped from a real 0.7.5 binary (the binary is not in this repo). The vendored artifacts are internally consistent (protocol 17, all referenced methods/types present), and `check:herdr-types` guards drift in CI.

PR #1900, opened 2026-07-22
feat(herdr): support herdr 0.7.5 — retire the capability guardrails (#1893)

scoop:
Closes #1893. Part of epic #1889.

Now that the earlier epic children (#1895/#1897/#1899) let Shepherd **drive** herdr 0.7.5, this child retires the guardrails PR #1887 put up at the **capability / preflight / updater / diagnostics / docs** layer.

## What changed

The earlier children had decoupled two ceilings in `herdr-capabilities.ts`: the driver's spawnable ceiling (`HERDR_LAST_SPAWNABLE_VERSION = 0.7.5`) and the capability ceiling (`HERDR_LAST_SUPPORTED_VERSION`, still `0.7.4`). This PR **raises the capability ceiling to `0.7.5`**, re-converging the two — every consumer already reads the constant, so this is a one-line behavior change plus copy/test fallout.

- **`herdr-capabilities.ts`** — `HERDR_LAST_SUPPORTED_VERSION` `0.7.4` → `0.7.5` (+ docstrings).
- **preflight (`index.ts`) / updater (`herdr-update.ts`) / diagnostics / `types.ts` / modal + messages (en+de)** — no logic change; reword the "0.7.5 broke `agent start` / pin to 0.7.4" copy generically against the shared ceiling.
- **`herdr.ts`** — copy-only reword of `HerdrSpawnUnsupportedError` message + JSDoc to reference `HERDR_LAST_SPAWNABLE_VERSION` (the constant the guard actually enforces). No logic/threshold change.
- **docs (README + getting-started)** — drop the "pin to 0.7.4" advice; 0.7.5 is the supported ceiling.
- **tests** — retarget capability/updater/diagnostics + modal tests to the raised ceiling: `0.7.6+` unsupported, `0.7.4 → 0.7.5` allowed, run button re-shown for a supported latest.

A herdr **newer than 0.7.5** stays warned/blocked/refused across both the capability layer and the driver — the two ceilings now agree, so the in-app updater can't upgrade an operator into a version the driver would refuse.

## Out of scope

Driver/spawn logic (spawnable ceiling value, spawn path, socket driver) left behaviorally byte-identical. Socket-driver default stays off.

## Acceptance

- [x] A 0.7.5 install trips no unsupported warning, diagnostic, or updater block.
- [x] The in-app updater can move 0.7.4 → 0.7.5.
- [x] Docs no longer advise pinning to 0.7.4.
[...]

PR #1901, opened 2026-07-22
feat(herdr): support herdr 0.7.5 (protocol 17) agent spawning (epic #1889)

### LINKED ISSUE #1889: herdr 0.7.5 (protocol 17): agent spawning via `agent start --kind` is incompatible with Shepherd's sandboxed/wrapped spawn command

### LINKED ISSUE #1890: herdr 0.7.5 spawn: CLI driver path (tab create + pane run + report-agent registration)
Parent epic: #1889. **DAG root — no blockers; this is the first child to drain.**

## Goal

Give Shepherd a herdr **0.7.5 (protocol 17)** agent-spawn path on the **CLI driver** that preserves the
env-shim + bwrap-sandbox wrap, and route 0.7.5 spawns through it (lifting the hard spawn-refusal PR
#1887 added, for the CLI path). All spawns funnel through the single `IHerdrDriver` seam
(`start`/`read`/`send`/`stop`/`relabel`), so this is a **version-branched reimplementation of those
five methods** — callers are unchanged, and `≤0.7.4` behavior stays byte-identical.

## Why (spike-confirmed against a live 0.7.5)

`agent start --kind <enum> --pane <id>` can express neither the `env … claude` shim nor
`bwrap … -- env … claude`. But herdr's **external-registration** seam does the job and needs **no**
passive detection (decisive: the membrane's `--unshare-pid` hides `claude` from host `/proc`):

1. `tab create --cwd <p> --env …` → root pane at a shell prompt (reply carries `pane_id` / `terminal_id` / `tab_id`).
2. `pane run <pane_id> <cmd…>` runs an **arbitrary** command — the existing wrapped-argv builder output (the `env … / bwrap …` line) flows in **verbatim**; no `--kind` limit.
3. `pane report-agent-session <pane_id> --source shepherd --agent <name> --agent-session-id <id>` + `pane report-agent <pane_id> --source shepherd --agent <name> --state working` → **`agent list` then returns a full agent record** (`agent`, `agent_status`, `pane_id`, `terminal_id`, `tab_id`, `cwd`, `workspace_id`). CLI quirk: the `pane_id` positional must precede the flags.

## Vertical cut (scope)

- A pure `sanitizeHerdrAgentName()` helper enforcing herdr's `^[a-z][a-z0-9_-]{0,31}$` (spike-confirmed: `TASK-01`, `plan-review TASK-707` → `invalid_agent_name`). Applied to both the report-agent `--agent` label and `agent rename`.
- A `terminal_id → pane_id` resolver (fed by `agent list`, which returns both): on 0.7.5 the driver's `read`/`send`/`stop`/`relabel` must target `pane_id` — `terminal_id` and label are rejected (`agent_not_found`). Shepherd keys sessions on `terminal_id`, so this resolver is load-bearing.
- Version-branched `start` on the CLI driver: `0.7.5` → `ensureWorkspace` → `tab create` → shell-readiness wait (retry on `agent_pane_busy`; `pane wait-output` for the prompt) → `pane run <wrapped argv>` → `report-agent-session` + `report-agent` → resolve the started agent from `agent list`. Preserve the existing dedicated-tab + collision-retry + tab-ledger teardown semantics.
- Version-branched `read`/`send`/`stop`/`relabel` on 0.7.5: **writes go through `pane send-text` + `pane send-keys Enter`**, NOT `agent prompt`/`agent send-keys` (spike-confirmed those return `agent_not_ready` on a registered-but-undetected sandboxed agent).
- Flip the `herdrSpawnSupported()` gate so 0.7.5 runs the new path instead of refusing.

## Out of scope

- The socket driver (its own child) — CLI path only here.
- Keeping `agent_status` fresh for sandboxed agents / reconcile + husk correctness under a `bwrap` foreground (its own child).
- Retiring the version ceiling, unblocking the in-app updater, or docs/diagnostics (its own child).
- Any behavior change on herdr ≤0.7.4.

## Testing seam

The CLI driver already takes injectable `Runner`/`AsyncRunner`, and the reply parsers are pure — unit-test the 0.7.5 orchestration with a fake runner returning **captured 0.7.5 JSON fixtures** (add them under `test/fixtures/herdr-responses/`); root CI runs only `./test`, so tests live in `test/`. Unit-test `sanitizeHerdrAgentName()` and the `terminal_id→pane_id` resolver as pure leaves.

## Acceptance

- [ ] With a 0.7.5 herdr, a real `claude` spawned under the bwrap membrane appears in `agent list`, and Shepherd can read its terminal and deliver a steer to it.
- [ ] A name like `TASK-01` is accepted (sanitized) rather than rejected.
- [ ] `≤0.7.4` spawns are byte-identical to today (existing herdr tests still green).
- [ ] `bun install && bun run lint && bun test` pass.

### LINKED ISSUE #1891: herdr 0.7.5: lifecycle-state push + reconcile/husk correctness for sandboxed agents

### LINKED ISSUE #1892: herdr 0.7.5: socket driver protocol-17 branch + schema/type codegen

### LINKED ISSUE #1893: herdr 0.7.5: remove spawn guard, unblock in-app updater, drop 0.7.4-pin docs
Parent epic: #1889. **Blocked by the CLI spawn-path child AND the lifecycle/reconcile child** (the two
that make 0.7.5 genuinely usable). **NOT blocked by the socket-driver child** — the socket driver is
default-off and protocol-gated, so an un-migrated socket driver simply stays inactive on 0.7.5 and
never gates the CLI experience.

## Goal

Retire the herdr-0.7.5 guardrails PR #1887 put up, now that Shepherd can drive 0.7.5: stop refusing
the version, re-enable the in-app updater to it, and drop the "pin to 0.7.4" guidance.

## Vertical cut (scope)

- Raise/retire the "last supported herdr version" ceiling so 0.7.5+ is no longer treated as unsupported by the capability check + boot preflight warning.
- Re-enable the in-app herdr updater to install 0.7.5+: the update check must stop flagging it unsupported, `apply()` must stop refusing, and the update modal must stop warning + re-show the run button.
- Drop the unsupported-herdr **diagnostic** and the **docs** advice to pin 0.7.4 (README + getting-started).

## Out of scope

- Any driver/spawn logic (the earlier children).
- Flipping the socket driver default-on.

## Testing seam

Update the existing capability / updater / diagnostics unit tests for the raised ceiling (these
already exist and assert the current 0.7.4 ceiling — they are the seam). Tests in `test/`.

## Acceptance

- [ ] A 0.7.5 install no longer trips the unsupported-version warning, diagnostic, or updater block.
- [ ] The in-app updater can move 0.7.4 → 0.7.5+.
- [ ] Docs no longer advise pinning to 0.7.4.
- [ ] `bun install && bun run lint && bun test` (+ `cd ui && bun install && bun run check && bun test` for the modal) pass.

scoop:
Lands epic **#1889 — herdr 0.7.5 (protocol 17): agent spawning via `agent start --kind` is incompatible with Shepherd's sandboxed/wrapped spawn command** from `epic/1889-herdr-0-7-5-protocol-17-agent-spawning-v` onto `main`.

Closes #1889
Closes #1890
Closes #1891
Closes #1892
Closes #1893

### Children (4)

| Issue | Title | PR |
| ----- | ----- | -- |
| #1890 | herdr 0.7.5 spawn: CLI driver path (tab create + pane run + report-agent registration) | #1895 |
| #1891 | herdr 0.7.5: lifecycle-state push + reconcile/husk correctness for sandboxed agents | #1897 |
| #1892 | herdr 0.7.5: socket driver protocol-17 branch + schema/type codegen | #1899 |
| #1893 | herdr 0.7.5: remove spawn guard, unblock in-app updater, drop 0.7.4-pin docs | #1900 |


scoop (commented):
## Review — herdr 0.7.5 (protocol 17) agent spawning (epic landing)

Reviewed the whole `git diff 39a6631f9a0a48aea477fa861562ff71e761035c...HEAD`, with extra attention to cross-child integration seams (capability ceilings, the CLI vs socket spawn paths, the poller push, protocol codegen, and i18n).

The change is coherent and internally consistent. I verified the load-bearing seams against ground truth:

- **Protocol method/param names** used by the socket driver all exist in the regenerated types: `pane.send_text`/`RequestPaneSendTextParams` (`src/generated/herdr-protocol.ts:1781,3011`), `pane.send_keys`/`RequestPaneSendKeysParams` (`:1773,3012`), `pane.report_agent`/`RequestPaneReportAgentParams` with a `state: RequestPaneAgentState` field (`:1705-1713,3018`), `pane.report_agent_session` (`:1719,3019`), `tab.create`/`RequestTabCreateParams` with `focus`/`label`/`cwd` (`:1988,2974`). `RequestPaneAgentState` is `"idle"|"working"|"blocked"|"unknown"` (`:416`), a superset of the poller's `PushableAgentState` (`src/herdr.ts:127`), so every pushed value is valid.
- **Legacy escape hatch**: `agent.send` and the old `agent.start` shape were dropped from the p17 types (`agent.send` → `agent.send_keys` at `:2985`; `RequestAgentStartParams` reshaped to `{kind,pane_id,…}` at `:1433-1444`). The ≤0.7.4 arms correctly route through `requestLegacy` (`src/herdr-socket-driver.ts:132,354`), keeping the ≤p16 wire behavior byte-identical. No dangling references to removed fields (`custom_status`, `RequestAgentSendParams`) remain in `src`/`ui/src` (grep clean).
- **Capability ceilings re-converged**: `HERDR_LAST_SUPPORTED_VERSION` and `HERDR_LAST_SPAWNABLE_VERSION` are both `0.7.5` (`src/herdr-capabilities.ts`), `herdrSpawnSupported()` gates `<= 0.7.5`, `isHerdrVersionSupported()` gates `<= 0.7.5`, and the diagnostics ceiling feeds `HERDR_LAST_SUPPORTED_VERSION` (`src/diagnostics.ts:1012`) — consistent across preflight/diagnostics/updater/driver.
- **`posixShellJoin`** single-quote break-out (`src/herdr.ts`): the template `` `'\''` `` yields the 4-char POSIX sequence `'\''`; correct.
- **`sanitizeHerdrAgentName`** matches herdr's `^[a-z][a-z0-9_-]{0,31}$` grammar (max 32 via `.slice(0,32)`), and `agentNameMatchesSession` compares in sanitized space only on the 0.7.5 path (`src/herdr.ts:171-176`) — matching how `agent list` reports the registered (sanitized) name.
- **Poller push** (`src/poller.ts`): `isSandboxedSession` uses real `Session.sandboxApplied`/`sandboxDegraded` fields (`src/types.ts:101-103`), `SandboxProfile` import resolves (`src/sandbox.ts:35`), the in-flight/rollback dedup converges, and cleanup drops all three new maps/sets. `probeCheckMs`, `hookActivityFresh`, `lastMatched` all resolve.
- **i18n parity**: `herdrupdate_unsupported_body` and `diagnostics_hint_herdr_unsupported` were updated symmetrically in both `ui/messages/en.json` and `ui/messages/de.json`; the Svelte modal references only existing keys.

### Nit (non-blocking)

The doc comment on `SocketHerdrDriver.reportAgentState` (`src/herdr-socket-driver.ts:425-427`) states the lifecycle push targets 0.7.5 sandboxed agents "which the socket driver never hosts (it refuses the 0.7.5 external-registration spawn branch in `start`)." This is stale/incorrect within this same PR: `HERDR_SOCKET_SUPPORTED_PROTOCOLS` now admits protocol 17 (`src/config.ts`), `selectHerdrDriver` returns a `SocketHerdrDriver` for protocol 17 (`src/herdr-socket-driver.ts:470-472`), and `start()` explicitly branches to `startImpl075` on the external-registration path (`src/herdr-socket-driver.ts:168-171`). So the socket driver *does* host these agents on 0.7.5, and the poller can legitimately call this method on it. The implementation is correct — only the comment's rationale contradicts the code and should be reworded (it reads like a leftover from before child #1892 added the socket spawn branch).

### Notes / limitations

- The live herdr 0.7.5 CLI/socket surface (exact error codes like `agent_not_ready`/`agent_not_found`, the `Enter` key name, whether `report-agent` emits a name-taken error) is not present in this repo; those behaviors are taken as spike-confirmed as documented in the code and could not be re-verified against ground truth here. Assumed as written.
- On 0.7.5 the sync `read()` now issues an extra synchronous `agent list` per read via `resolveTargetSync` (`src/herdr.ts`). This is intentional (pane_id resolution) and only on the 0.7.5 path, but it doubles the sync CLI round-trips per read — noted for awareness, not a defect.

scoop (commented):
Reviewed the whole `git diff f16d1149...HEAD` (52 files) with an integration lens on the four merged children. Verification notes and citations below.

## What I verified (and how)

- **Protocol identifiers resolve.** Every new socket method the driver speaks exists in the vendored p17 map: `pane.send_text`, `pane.send_keys`, `pane.report_agent`, `pane.report_agent_session`, `tab.create`, `agent.read`, `agent.rename` (src/generated/herdr-protocol.ts:3011-3019, :2974, :2983-2986). Param shapes match the call sites: `RequestPaneReportAgentParams{agent,pane_id,source,state}` (src/generated/herdr-protocol.ts:1705-1714) vs src/herdr-socket-driver.ts:460-466; `RequestPaneSendTextParams{pane_id,text}` (:1781-1784) and `RequestPaneSendKeysParams{keys,pane_id}` (:1773-1776) vs src/herdr-socket-driver.ts:150-153. `RequestPaneAgentState = "idle"|"working"|"blocked"|"unknown"` (:413) is a superset of `PushableAgentState` (src/herdr.ts:127), so `reportAgentState` typechecks.
- **`agent.send` / `agent.start` removal is handled.** Both are absent from / reshaped in the p17 `HerdrParams` map (`agent.send` gone; `RequestAgentStartParams` now `{kind,name,pane_id,args}` at src/generated/herdr-protocol.ts:1437-1441), and the legacy arm routes them through `requestLegacy` (src/herdr-socket-client.ts:76-82) — consistent.
- **`IHerdrDriver.reportAgentState` has both implementors.** `HerdrDriver` (src/herdr.ts:799) and `SocketHerdrDriver` (src/herdr-socket-driver.ts:456); no third implementor exists (grep over src/). The poller's narrowed `Pick<>` (src/poller.ts:362) is satisfied by the `IHerdrDriver` passed at src/index.ts:1017 / :534.
- **i18n parity holds.** All six new keys exist in BOTH catalogs: `topbar_herdr_advisory_title`, `herdrupdate_sandbox_advisory_title`, `herdrupdate_sandbox_advisory_body`, `herdrupdate_sandbox_downgrade_confirm`, `feat_herdr_sandbox_advisory_title`, `feat_herdr_sandbox_advisory_body` — ui/messages/en.json:12,29-31,3369-3370 and ui/messages/de.json:12,29-31,3369-3370. The feature-announcement entry's `titleKey`/`bodyKey` (ui/src/lib/feature-announcements/entries/v1.45.0-herdr-sandbox-idle-advisory.ts:5-6) resolve to those keys, and `sinceVersion: 1.45.0` is > package.json:3 (`1.44.0`) as the catalog contract requires (ui/src/lib/feature-announcements.ts:11-17). Entries are eager-globbed (ui/src/lib/feature-announcements.ts:47-50), so no index registration is needed.
- **New server fields are mirrored client-side.** `sandboxIdleRegressed` / `sandboxDowngradeTarget` exist in both src/types.ts:423-427 and ui/src/lib/types.ts:1466-1472, and are consumed at ui/src/lib/components/HerdrUpdateModal.svelte:156, :335 and ui/src/lib/components/TopBar.svelte:135-137.
- **Wiring identifiers exist.** `config.sandboxDefaultProfile` (src/config.ts:572), `store.list({activeOnly})` (src/store.ts:2106), `Session.sandboxApplied`/`sandboxDegraded` (src/types.ts:101-103) — all used correctly by the `sandboxedInUse` closure at src/index.ts:2625-2633.
- **`classifyPaneWrite` matches its only real caller's contract.** `sendSteerTo` issues exactly two writes — the bracketed-paste blob then `"\r"` (src/service.ts:4434-4435) — so the CR maps to the `Enter` key and the blob to `send-text`, as documented (src/herdr.ts:455-458). The only other `herdr.send` caller sends a lone `"\x1b"` (src/service.ts:4368), which correctly takes the literal-text branch.
- **`isValidTerminalId` widening is contained.** Its only production caller is `PtyBridge` (src/pty-bridge.ts:28), and the regex now matches src/pty-attach.mjs:8 byte-for-byte.

## Findings

**1. (blocking) The `name`-based matching added by #1890 contradicts what #1891 empirically found about `agent list`.** src/poller.ts:1307-1310 states, from a live-herdr fix, that *"the list's `name` is null for externally-registered agents, so `agent.name` was empty — every push failed on `--agent `"* — that is why the push now re-derives the label with `sanitizeHerdrAgentName(s.name)`. But two other in-diff sites still key off `HerdrAgent.name` on the very same 0.7.5 external-registration path:
- src/herdr.ts:157-160 / :178 / :281 — `agentNameMatchesSession` compares `a.name` (which per the above is `""` after `mapAgentRecord`'s `a.name ?? ""`, src/herdr.ts:377) against `sanitizeHerdrAgentName(s.name)`. It can never be true, so the cwd-collision disambiguator in `matchAgent`/`pickByCwd` — the exact scenario the doc comment says it exists to fix — silently degrades to "no match" on 0.7.5.
- src/herdr.ts:1096 and src/herdr-socket-driver.ts:310 — the collision breaker evicts squatters with `a.name === agentName`, which for the same reason matches nothing, making the `MAX_ATTEMPTS` retry a pure re-issue of the failing register pair.

The test fixture test/fixtures/herdr-responses/v0.7.5/agent-list-registered.json:8-9 encodes the *opposite* shape (`"name": "review-task-09"`, `"agent": "claude"`), so CI green does not contradict this — the fixture is what #1890 assumed before #1891 measured. Please reconcile: either match on the field herdr actually populates (`agent`), or fix the fixture and the poller comment so one story holds across the epic.

**2. (blocking) `sandboxFlags` is dropped from the two post-run status assemblies, breaking the sandbox-downgrade retry.** `check()` spreads `...sandboxFlags(...)` (src/herdr-update.ts:639, :649), but the statuses rebuilt after `apply()` (src/herdr-update.ts:468-477) and after `downgrade()` (src/herdr-update.ts:575-584) do not. Concretely: a sandbox downgrade that FAILS (`error: "herdr was not downgraded"`, still on 0.7.5) publishes a status with `sandboxIdleRegressed` undefined. The server route then refuses the retry with 409 `"no sandboxed-idle regression to downgrade for"` (src/server.ts:4230-4232), and the TopBar badge (ui/src/lib/components/TopBar.svelte:135-137) and modal advisory (ui/src/lib/components/HerdrUpdateModal.svelte:156) both vanish — until the next periodic `check()`, which is up to 6h away. Spread `sandboxFlags(this.sandboxedInUse())` at both sites.

**3. (blocking) Two different attach-target forms are handed to the same `herdr agent attach` helper.** src/server.ts:7585 (`nodePtyAttachTarget`) passes `attach.paneTarget`, which src/server.ts:7573 deliberately normalizes to `"${workspaceId}:${paneId}"`; src/usage-probe.ts:146-149 passes the RAW `resolvePaneId(...)` result, i.e. the bare `pane_id` (src/herdr.ts:436 → `a.pane_id`, bare `"p_075"` in test/fixtures/herdr-responses/v0.7.5/agent-list-registered.json:11). Both values are spawned into the *same* `src/pty-attach.mjs`, which runs `herdr agent attach <target> --takeover` (src/pty-attach.mjs:21). One of these two forms must be wrong for that command; the widened validator (src/validate.ts:874-880, src/pty-attach.mjs:8) accepts both, so the mismatch fails only at runtime. Pick one form and use it on both paths.

**4. (non-blocking) Unbounded 1 Hz warn + retry when a state push keeps failing.** src/poller.ts:1320-1325 rolls back `lastPushedState` and `console.warn`s on every rejection; `maybePushAgentState` is invoked from `reconcileAgent` each tick (src/poller.ts:794) at `intervalMs = 1000` (src/poller.ts:379). Any persistent failure — herdr down, pane gone, or the `idle` de-escalation the method's own KNOWN LIMITATION block (src/poller.ts:1288-1292) documents as rejected on 0.7.5 — produces one warn line per second per sandboxed session forever. Add a backoff or log-once-per-transition.

**5. (nit) Extra path segments are now accepted on the downgrade route.** src/server.ts:4225-4226 replaced the old `!parts[3]` exactness check; `/api/herdr-update/downgrade/sandbox/anything` now matches the sandbox branch because `parts[4]` is never checked. Add `&& !parts[4]`.

**6. (nit) The sync `read()` now spawns an extra synchronous `herdr agent list` per call on 0.7.5.** src/herdr.ts:1214-1216 → `resolveTargetSync` → `this.list()` (`execFileSync`). Every `readAsync` (src/herdr.ts:1230) and every `send` (src/herdr.ts:1176) likewise adds a full `agent list` round-trip. At the 1 Hz poll cadence across N sessions this roughly doubles herdr process spawns; a short-TTL terminal_id→pane_id cache invalidated on the poller's existing per-tick `listAsync` would remove it.

## Stated limitations (could not verify — not raised as findings)

- herdr's actual CLI/RPC semantics are not in this repo, so I could not confirm: that `herdr pane run <pane> <single-quoted-cmdline>` accepts the whole argv as one positional (src/herdr.ts:1041); that `pane run`/`pane.send_text` always reject *before* executing, which is what makes the 3x no-delay retries in src/herdr.ts:1034-1046 and src/herdr-socket-driver.ts:280-289 safe against double-launch; that `"Enter"` is the correct key name; or whether `report-agent --state idle` errors or silently no-ops. Assumed as written.
- Whether real herdr returns `pane_id` bare or already `ws:pane` is unknowable here — hence finding 3 is raised as an *internal* inconsistency, not a claim about which form herdr wants.

- src/herdr.ts:975 — the trusted 0.7.5 path (`resolveByAutoDetect`) never assigns the agent a Shepherd-chosen name, unlike `agent start --name` on ≤0.7.4, so `agentNameMatchesSession` (src/herdr.ts:157) also cannot match for trusted sessions; today this only degrades the rarely-hit two-sessions-per-worktree fallback.
- src/herdr.ts:1096 / src/herdr-socket-driver.ts:310 — `closeTab(sq.tabId)` is called without guarding `tabId !== ""`, so a squatter record with no `tab_id` (`mapAgentRecord`'s `?? ""`, src/herdr.ts:379) would issue a `tab close ""`; unreachable while the eviction filter itself never matches (finding 1).

Issue #1905, opened 2026-07-22
Onboarding harness: non-gating advisory when herdr's latest release exceeds the supported ceiling


scoop:
Follow-up to #1896 (pinning the herdr fresh-install path to `HERDR_LAST_SUPPORTED_VERSION`).

## What the pin removed

Before the pin, the nightly harness installed herdr **latest** on every real-install scenario. That gave us an *accidental* early warning: the day herdr shipped a release above Shepherd's ceiling, the nightly went red and filed the rolling issue.

That detector was destructive — it blocked the release gate on an upstream event Shepherd neither caused nor could fix, which is precisely what #1896 was. Pinning removes it, deliberately. But the *signal* was genuinely useful: it told the team "herdr moved, Shepherd needs to catch up" within 24h.

## What still covers it, and what doesn't

Per-host, this already ships: `HerdrUpdateService.check()` sets `latestUnsupported`, the update modal warns, and the in-app updater refuses to cross the ceiling (#1887/#1902). What's gone is the **team-wide daily nudge** — nobody is watching an individual operator's modal.

## Proposal

Add a **non-gating** advisory to the nightly gap report: fetch `latest.json` (the harness can reuse `LATEST_URL`/`fetchLatest` from `src/herdr-update.ts`) and, when `latest > HERDR_LAST_SUPPORTED_VERSION`, emit a line naming both versions.

Design constraints worth keeping:

- **Non-gating.** It must not touch `gateGapScenarios` — an upstream release is not a Shepherd regression, and blocking a release on it repeats #1896.
- Surface it where a human actually looks: the rolling issue / status line, alongside the existing harness-error channel.
- Tolerate a fetch failure silently (the harness must not go red because herdr.dev was briefly unreachable).

Deferred out of #1896's PR to keep that change to one cohesive diff, and to avoid adding a network dependency to the gap report in the same commit that changed how herdr is installed.

PR #1906, opened 2026-07-22
fix(herdr): pin the fresh-install path to the supported ceiling

### LINKED ISSUE #1896: Onboarding harness: nightly regression detected
# Onboarding Gap Report

**3 / 6 scenarios reached green.** (4 detection-only, by design — excluded.)

| herdr-missing | images:archlinux | yes | verbatim | no | ADVICE GAP |
| herdr-outdated | images:archlinux | yes | skipped | no | DETECTION-ONLY |
| install-e2e | images:ubuntu/24.04 | no | verbatim | no | INSTALL GAP |
| install-e2e-service | images:ubuntu/24.04 | no | verbatim | no | INSTALL GAP |

## Gaps

- **herdr-missing** (ADVICE GAP)
- **install-e2e** (INSTALL GAP) — herdr want=ok got=error
- **install-e2e-service** (INSTALL GAP) — herdr want=ok got=error

_Nightly run: 2026-07-22T01:40:00.474Z_

scoop:
Fixes #1896.

## The regression, and why it is already gone

The Jul 22 nightly ran on `39a6631` (#1887), which pinned the herdr support ceiling to **0.7.4** while `herdr.dev` served **0.7.5**. Every Shepherd install path used `curl -fsSL https://herdr.dev/install.sh | bash`, which is **latest-only**, so all three real-install scenarios landed a herdr above the ceiling and reported `herdr want=ok got=error`. #1901 later raised the ceiling to 0.7.5, and re-running the three failing scenarios on unmodified `main` confirmed the proximate cause was already resolved:

| Scenario | Jul 22 nightly (`39a6631`) | Re-run on `633dea48` | With this PR |
| --- | --- | --- | --- |
| `herdr-missing` | ADVICE GAP | PASS | **PASS** |
| `install-e2e` | INSTALL GAP — `herdr want=ok got=error` | PASS | **PASS** |
| `install-e2e-service` | INSTALL GAP — `herdr want=ok got=error` | PASS | **PASS** |

## What this PR actually fixes

The **durable** hole the incident exposed: the pin was one-sided. Shepherd's driver refuses to spawn on a herdr above `HERDR_LAST_SUPPORTED_VERSION`, but every install path fetched *latest* — so the day herdr ships 0.7.6, every fresh onboarding breaks again, for a reason no Shepherd change caused.

Installs now fetch the **version-addressable release asset for the ceiling**. One constant drives every path: raise the ceiling and installs follow, with no second edit.

## Design notes worth reviewing

**The download budget is a parameter, not a constant.** `defaultRunRemediation`'s 120s group-kill binds only the in-app Fix — and that is *not* the path where being herdr-less strands anyone, since `preflightHerdr` exits 78 before the server binds, so the in-app `herdr_missing` Fix is reachable only with Shepherd already running and its UI up. `provision.ts` (a fresh `install.sh`) and the preflight banner therefore get **upstream-comparable** tolerance (`--max-time 120`, ≈1.4 Mbit/s for the measured 20.3 MiB artifact); the in-app entry gets a budget that provably fits alongside `HERDR_SERVE`, asserted in a unit test against the real `REMEDIATION_TIMEOUT_MS` with every term declared (never a residual, which would make the check tautological).

**Verification blocks only on positive evidence the artifact is wrong.** Governing rule throughout: never worse than today's unverifying `curl | bash`.

| Branch | Behaviour |
| --- | --- |
| `uname` miss (OS *or* arch) | fail loudly, naming the release tag |
| download failure | fail loudly — parity with today, where `defaultRunner` already aborts `install.sh` |
| cannot `exec` (glibc asset on musl) | **warn, install anyway** — a hard abort would kill `install.sh` on Alpine hosts that install fine today |
| `--version` unparseable | **warn, install anyway** — that is evidence about our parser, not the artifact |
| parses, but differs from the pin | fail loudly, no swap |
| `mkdir`/`chmod`/`mv` failure | fail loudly, non-zero, no success line |

**The success line is a verified fact, not an intention.** `echo` exits 0, so an unguarded success message would report a healthy install on a host with no herdr — and provision's runner (which throws only on non-zero) would sail on. The claim sits behind a post-swap `[ -f ] && [ -x ]` check; `-f` matters because if `~/.local/bin/herdr` were a directory, `mv -f` moves the temp *inside* it and a bare `-x` passes.

**The harness assertion has teeth.** `herdr: ok` cannot distinguish a pinned install from an unpinned one while the ceiling equals latest, so all three real-install scenarios now assert the **installed** `herdr --version` equals the pin. A unit test proves it bites (a `9.9.9` install fails the scenario even though the check reads ok).

**Deliberately not pinned:** `herdr update --handoff` (verified: no version argument, so pinning means abandoning `--handoff` and killing live agent panes) and provision's skip-when-adequate (making it ceiling-aware would silently downgrade an operator's own binary). Both are already covered *with consent* by #1887's banner and #1902's one-click downgrade. The rule: **pin what we install; never silently rewrite what the operator already has.**

## Verification

- `bun run lint` clean; `bun run test` 7655 pass / 0 fail.
- Empirical acceptance on the real Incus host, with the pin in place: `herdr-missing` (Arch, in-app remediation path), `install-e2e` and `install-e2e-service` (Ubuntu, provision + systemd-unit paths) — **all PASS**, each also asserting the installed version equals the pin.
- New `test/herdr-install.test.ts` executes the real shell program under a `PATH` shim (fake `uname`/`curl`, and a `$HOME` containing a space so a missing quote fails in CI rather than on a user's machine): all six failure branches plus the success path, and a mapping-agreement table proving the TypeScript `herdrAssetKey` and the shell `case` arms cannot drift — including the macOS arms, which no harness scenario can reach.

## Trade-offs recorded

- The pin removes an *accidental* early-warning signal (the nightly used to redden when herdr outran the ceiling). It was a destructive detector — it blocked the release gate on an upstream event Shepherd could not fix, which is exactly what #1896 was. Follow-up #1905 tracks a non-gating replacement advisory.
- In-app, a 1.42–1.76 Mbit/s link can now fail the Fix where upstream's looser timeout would have squeezed through. Non-catastrophic by construction: that path only exists with Shepherd already running, so the operator sees the failure and can retry.
- The `latest.json` `releases` cross-check that the downgrade flow performs is **not** reproduced here — it protects a working install about to be swapped, and a fresh install has nothing to protect. Upstream performs no integrity check we lose: it does `curl` → `mv` → `chmod +x` with no checksum or signature, so this path is strictly *more* verified than today.

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,10 +133,14 @@ enables the systemd user service. Idempotent — safe to re-run: it never clobbe
 
 This is third-party `curl|bash`: the script runs unconfined as your user _before_ any sandbox
 exists. It also invokes upstream installers it does not control — specifically: [bun.sh](https://bun.sh/install),
-[fnm.vercel.app](https://fnm.vercel.app) (Node via fnm), [herdr.dev](https://herdr.dev) (herdr), and
-[claude.ai](https://claude.ai/install.sh) (the `claude` CLI), plus your distro's package manager
-(`apt` / `apk` / `dnf` / `pacman`) for `git`, `unzip`, and the C/C++ build toolchain + `python3`
-(needed for the node-pty native build).
+[fnm.vercel.app](https://fnm.vercel.app) (Node via fnm) and [claude.ai](https://claude.ai/install.sh)
+(the `claude` CLI), plus your distro's package manager (`apt` / `apk` / `dnf` / `pacman`) for `git`,
+`unzip`, and the C/C++ build toolchain + `python3` (needed for the node-pty native build). herdr is
+**not** installed through `herdr.dev/install.sh` (which is latest-only): Shepherd downloads a
+**version-pinned** binary from
+[github.com/ogulcancelik/herdr/releases](https://github.com/ogulcancelik/herdr/releases) — the
+highest release Shepherd supports — verifies it reports that version, and installs it to
+`~/.local/bin`. That release binary is still third-party code fetched and executed on your machine.
 
 > **Shepherd supports herdr up to 0.7.5.** herdr 0.7.5 (protocol 17) reshaped `agent start`;
 > Shepherd drives it through a CLI external-registration path. Don't upgrade past 0.7.5 yet —
@@ -236,7 +240,8 @@ For the from-clone / development path, see [Quick start](#quick-start) below.
 
 - [Bun](https://bun.sh) — backend runtime + package manager
 - `herdr` on `PATH` — manages the interactive `claude` panes (owns the PTYs)
-  - On macOS, `herdr.dev/install.sh` installs to `~/.local/bin` by default — add it to your shell
+  - Shepherd installs a **version-pinned** herdr (the highest release it supports) into
+    `~/.local/bin`. On macOS that directory is not on `PATH` by default — add it to your shell
     profile before `bun run start`: `export PATH="$HOME/.local/bin:$PATH"`
 - The `claude` CLI, logged in with your Max/Pro subscription
 - Node.js — for the PTY helper subprocess
--- a/ci/onboarding-harness/run.ts
+++ b/ci/onboarding-harness/run.ts
@@ -18,6 +18,7 @@ import { assertDetection } from "./assert";
 import { buildGapReport, gateGapScenarios, statusDescription } from "./report";
 import { reportToGitHub, publishStatus } from "./issue";
 import { remediationsFor } from "../../src/remediations";
+import { HERDR_LAST_SUPPORTED_VERSION } from "../../src/herdr-capabilities";
 import { HERDR_MISSING_EXIT_CODE, HERDR_MISSING_MARKER } from "../../src/preflight";
 import type { DetectionResult, Scenario, ScenarioResult } from "./types";
 import type { DiagnosticsSnapshot } from "../../src/types";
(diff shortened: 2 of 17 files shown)


scoop (commented):
## Summary

The PR replaces the latest-only `curl -fsSL https://herdr.dev/install.sh | bash` fresh-install path with a version-addressable, verify-before-swap download pinned to `HERDR_LAST_SUPPORTED_VERSION`, per the approved plan. I verified the implementation against the actual code; it is correct and internally consistent, and I found no blocking defect.

## What I verified

- **Version constant & slug parity.** `HERDR_LAST_SUPPORTED_VERSION = "0.7.5"` (src/herdr-capabilities.ts:17). The download slug `github.com/ogulcancelik/herdr` in the new src/herdr-install.ts:40,51,178 matches the pre-existing downgrade template (verified against test/herdr-downgrade.test.ts:20 `URL074`, same slug). README/docs and the preflight banner all use the same slug.
- **Shell URL vs `herdrReleaseUrl` agreement.** The runtime shell builds `herdr-$O-$A` (O=linux/macos, A=x86_64/aarch64), which reconstructs exactly `herdr-${assetKey}` from `herdrReleaseUrl` (`${os}-${cpu}`) — src/herdr-install.ts:34-35,50-51,178. The agreement is machine-asserted in test/herdr-install.test.ts:332-354 including the macOS arms no harness scenario reaches.
- **Budget arithmetic fits the group-kill.** `HERDR_IN_APP_DOWNLOAD_WORST_CASE_MS` = (0 + 90)*1000 = 90000; serve worst-case = (1+1+10*2)*1000 = 22000; verify = 1000 → 113000 ≤ `REMEDIATION_TIMEOUT_MS` (verified 120_000 at src/config.ts:124), ~7s slack. Test asserts this without a tautological residual (test/herdr-install.test.ts:73-79).
- **Failure-branch correctness.** Exec status is captured from the binary (`>"$OUT"; rc=$?`) not a pipeline, so the exec-failure branch is reachable (src/herdr-install.ts:368). `if/elif/elif` correctly distinguishes exec-fail (warn+swap), unparseable (warn+swap), and mismatch (rm+exit, no swap). Temp `$OUT` is removed unconditionally after the grep; `$T` is removed on every failing branch. The post-swap `[ -f ] && [ -x ] || {exit 1}` guard is sound and the success `echo` is gated behind it. All `&&`-chained through the swap.
- **Consumers still agree.** `HERDR_INSTALL` remains exported and composed as `${HERDR_INSTALL} && (${HERDR_SERVE})` (src/remediations.ts:34,122). provision.ts (deploy/provision.ts:66) uses the install-only string with the `upstream` budget — parity with the prior install-only `HERDR_INSTALL`. herdr-update.ts re-exports `herdrAssetKey`/`herdrReleaseUrl` and still imports `sanitizeVersion` internally (verified src/herdr-update.ts:12,25,63-64,481-485); test/herdr-downgrade.test.ts imports from herdr-update untouched. No import cycle (herdr-capabilities imports only semver, src/herdr-capabilities.ts:11).
- **Preflight banner.** Now a function with a null-`assetKey` fallback to the release-tag page; never renders `undefined` (src/preflight.ts:26-38, asserted test/preflight.test.ts:80-96). Console-only, no i18n catalog involved.
- **Harness pin assertion.** `assertPinnedHerdrInstalled` reads the installed `herdr --version` and requires equality with the pin, called only once the herdr check is `ok` (ci/onboarding-harness/run.ts:59-77,84,68-70). `driver.exec` returns `{stdout,stderr,code}` (verified ci/onboarding-harness/types.ts:97-101); throws are caught by `runScenario` and surfaced as `error` with `reachedGreen:false` (run.ts:340-351), matching the new negative test (test/onboarding-harness/herdr-preflight.test.ts:94-105). Seed stub now reports the pin instead of `99.99.99` (ci/onboarding-harness/seed.ts:157).

## Could not verify (out of repo)

- That `github.com/ogulcancelik/herdr` actually serves a `herdr-linux-x86_64` asset for `v0.7.5` — external, not in this repo. Treated as written; it is byte-parity with the pre-existing downgrade flow's template.

- src/index.ts:234 the UNSUPPORTED-herdr banner still points operators at `https://herdr.dev` for pinning, which reads slightly against the new "herdr.dev is not the install source" framing; that line is a different (already-installed/downgrade) path and is not in this diff.

PR #1910, opened 2026-07-22
test(ui): epsilon-tolerant a11y-floor geometry assertions

### LINKED ISSUE #1907: Pre-push ui lane: sub-pixel geometry flake in TopBar.browser.test.ts trains people to retry the gate
## Symptom

The pre-push `ui` lane fails intermittently on an unrelated PR, then passes on retry. Most recent occurrence, during PR #1906 (a server-side change touching **no** `ui/src` file):

```
FAIL  |browser (chromium)| src/lib/components/TopBar.browser.test.ts
  > TopBar — telemetry menu visual/geometry path
  > mobile: 12px-top-radius full-bleed sheet with the 52/48/44px row tiers
AssertionError: expected 51.99998474121094 to be greater than or equal to 52
 Test Files  1 failed | 272 passed (273)
```

Re-running the identical push with no code change: green.

## This is not a real geometry regression

`GearHaltHero.svelte` sets `min-height: 52px` on the hero row, so layout intent is satisfied — the measured value is under by **1.5e-5 px**, which no layout engine and no user can perceive. The assertion is comparing a float against an exact integer with `>=`:

```ts
expect(hero.getBoundingClientRect().height).toBeGreaterThanOrEqual(52);
```

`getBoundingClientRect()` returns the **transform-mapped** border box as floats. `TopBarMobileSheet.svelte` opens the sheet with `transition:fly={{ y: 520, duration: 180 }}`, so an ancestor carries an animating `transform` while the assertions run — and mapping through a transform matrix costs ULPs even when the transform is a pure translate.

The test's settle gate is what lets this through:

```ts
await vi.waitFor(() => {
  const r = sheet!.getBoundingClientRect();
  expect(Math.abs(r.bottom - window.innerHeight)).toBeLessThanOrEqual(1);
});
```

A **1px** tolerance on the sheet's bottom is satisfied while the fly transform is still settling, so whether the height assertion sees a transform-mapped float or a clean layout value is a race against the animation frame — i.e. machine-load dependent, which is exactly the observed behaviour.

## Why it is worth fixing rather than retrying

The retry is now folklore ("ui lane fails, push again"), and a gate people are trained to re-run on red stops being a gate. The failure also lands on PRs that cannot have caused it, so it costs attention on unrelated work — it burned a push cycle on #1906.

## Suggested direction (not prescriptive)

- Gate on the transition actually being **done** (`transitionend`, or `waitFor` with a 0-tolerance/stable-for-N-frames check) rather than a 1px proximity tolerance, so measurements happen against settled layout; **or**
- compare with a small epsilon / round to 2dp.

**Constraint on any fix:** these tiers encode the a11y tap-target floors (52/48/44), so the threshold must not be relaxed in a way that would let a genuine regression to, say, 48px pass. An epsilon of ~0.01px preserves the floor; lowering the constant does not.

The pattern is not isolated — `TopBar.browser.test.ts` has 4 `getBoundingClientRect().height >= <int>` assertions and there are ~18 similar tier assertions across `ui/src/lib/components/**.browser.test.ts`, so whatever shape is chosen should be applied consistently (a shared helper would stop it recurring in the next geometry test).

## Repro

Not reliably reproducible on demand — it is load-dependent. `cd ui && bun run test src/lib/components/TopBar.browser.test.ts` in a loop under CPU pressure is the most likely way to surface it.

scoop:
Closes #1907.

## Problem

The pre-push `ui` lane flakes intermittently on unrelated PRs (most recently #1906, a server-only change) with:

```
AssertionError: expected 51.99998474121094 to be greater than or equal to 52
  TopBar — telemetry menu visual/geometry path
  > mobile: 12px-top-radius full-bleed sheet with the 52/48/44px row tiers
```

Re-running the identical push goes green. The retry became folklore ("ui lane fails, push again") — and a gate people are trained to re-run on red stops being a gate.

## Root cause

The a11y tap-target tiers are asserted by comparing a `getBoundingClientRect()` dimension — a **transform-mapped float** — against an **exact integer** with `>=`. The mobile gear sheet opens with a Svelte `fly` transition, so an ancestor carries an animating `transform` while the assertions run; mapping a box through a transform matrix costs a few float32 ULPs (~1.5e-5px) even for a pure translate. `.height` is translate-invariant, so this is imperceptible float noise — not a real regression (`GearHaltHero.svelte` sets `min-height: 52px`) — but `51.99998… >= 52` is false, and whether the race reads the ULP-dirty value is machine-load dependent.

## Fix

A single shared, epsilon-tolerant helper — `ui/src/lib/test-support/geometry.ts` → `expectMinPx(value, floorPx, label?)` with `GEOMETRY_EPSILON = 0.05` — applied to **all 24** a11y size-floor assertions across **8** browser-test files (inline, two-line-wrapped, and variable-stored `rect`/dimension forms, incl. loop bodies and `hitSize` variable floors).

`0.05px` is ~3000× the observed ULP noise yet ~20× below the smallest real regression (1px); the tiers are ≥4px apart, so it can never let a genuine floor regression pass. A contract test (`geometry.test.ts`) locks this: it passes at `floor − 1.5e-5` and asserts `expect(() => expectMinPx(floor − 1, floor)).toThrow()`.

| File | Migrated | Floors |
| --- | --- | --- |
| TopBar | 8 | 52 / 48 / 44 (+44 width) |
| NewTask | 5 | 132 / 44 |
| EpicDraftModal | 4 | 44 |
| IssuesPanel | 2 | hitSize (h+w) |
| Herd | 2 | 44 |
| EpicDraftPanel | 1 | 44 |
| PromptSources | 1 | hitSize |
| UpNextPanel | 1 | hitSize |

## Out of scope (deliberate)

- Positional `.left`/`.right`/`.top`/`.bottom` containment bounds (different concern; several already carry ±slack).
- Computed-style font/ratio floors (read `getComputedStyle`, not transform-mapped rects).
- `Viewport.browser.test.ts:804` already bakes in a 0.5px epsilon — left as-is.
- The sheet's `vi.waitFor` settle gate — height is translate-invariant, so measuring mid-fly is fine once epsilon absorbs the ULP.

## Verification

- Full UI suite: **274 files, 4208 tests pass**; targeted contract + 8 migrated files green.
- Prettier + ESLint clean; pre-push all 7 lanes green (ui lane included).

Test-only change: no production code, no i18n/feature-catalog/glossary impact.

### DIFF
--- a/ui/src/lib/components/EpicDraftModal.browser.test.ts
+++ b/ui/src/lib/components/EpicDraftModal.browser.test.ts
@@ -5,6 +5,7 @@ import { userEvent } from "vitest/browser";
 import "../../app.css";
 import type { EpicDraft } from "$lib/types";
 import { epicDrafts } from "$lib/epic-draft.svelte";
+import { expectMinPx } from "$lib/test-support/geometry";
 import EpicDraftModal from "./EpicDraftModal.svelte";
 
 function longDraft(sessionId: string): EpicDraft {
@@ -76,8 +77,8 @@ describe.each([
     expect(getComputedStyle(list).overflowY).not.toBe("auto");
 
     // a11y sizing floor on the actions that commit or discard a whole epic.
-    expect(approve.getBoundingClientRect().height).toBeGreaterThanOrEqual(44);
-    expect(abort.getBoundingClientRect().height).toBeGreaterThanOrEqual(44);
+    expectMinPx(approve.getBoundingClientRect().height, 44, "approve tap-target");
+    expectMinPx(abort.getBoundingClientRect().height, 44, "abort tap-target");
 
     // Pinned means pinned: reachable at the top of the draft AND at the very bottom of it.
     expect(hitTestable(approve), "approve clickable at scroll-top").toBe(true);
@@ -125,11 +126,11 @@ describe.each([
     }
 
     expect(parseFloat(getComputedStyle(input).fontSize)).toBeGreaterThanOrEqual(bodyFontSize);
-    expect(input.getBoundingClientRect().height).toBeGreaterThanOrEqual(44);
+    expectMinPx(input.getBoundingClientRect().height, 44, "input tap-target");
     expect(buttons.length).toBeGreaterThan(0);
     for (const button of buttons) {
       expect(parseFloat(getComputedStyle(button).fontSize)).toBe(bodyFontSize);
-      expect(button.getBoundingClientRect().height).toBeGreaterThanOrEqual(44);
+      expectMinPx(button.getBoundingClientRect().height, 44, "button tap-target");
     }
 
     unmount();
--- a/ui/src/lib/components/EpicDraftPanel.browser.test.ts
+++ b/ui/src/lib/components/EpicDraftPanel.browser.test.ts
@@ -3,6 +3,7 @@ import { render } from "vitest-browser-svelte";
 import "../../app.css";
 import type { EpicDraft } from "$lib/types";
 import { epicDrafts } from "$lib/epic-draft.svelte";
+import { expectMinPx } from "$lib/test-support/geometry";
 import EpicDraftPanel from "./EpicDraftPanel.svelte";
 
 function longDraft(sessionId: string): EpicDraft {
@@ -125,7 +126,7 @@ describe("EpicDraftPanel bar — behavior", () => {
     const cta = container.querySelector<HTMLButtonElement>(".edp-cta");
     expect(cta, "awaiting draft should offer a review CTA").not.toBeNull();
     // a11y sizing floor — this is the only way into the review surface.
-    expect(cta!.getBoundingClientRect().height).toBeGreaterThanOrEqual(44);
+    expectMinPx(cta!.getBoundingClientRect().height, 44, "review CTA tap-target");
 
     cta!.click();
     expect(onreview).toHaveBeenCalledTimes(1);
--- a/ui/src/lib/components/Herd.browser.test.ts
+++ b/ui/src/lib/components/Herd.browser.test.ts
@@ -8,6 +8,7 @@ import { GROUP_KEY_BY_STAGE } from "./herd-partition";
 import { isReworkRunning } from "./rework-running";
 import { reviews, planGates } from "$lib/reviews.svelte";
 import { postMergeSteps } from "$lib/post-merge-steps.svelte";
+import { expectMinPx } from "$lib/test-support/geometry";
 import type { Session, GitState, Epic, EpicChild, PostMergeSteps, ReviewVerdict } from "$lib/types";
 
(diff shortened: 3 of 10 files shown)


scoop (commented):
## Summary

This PR replaces brittle `getBoundingClientRect().(height|width) >= <floor>` a11y-floor assertions with a shared `expectMinPx()` helper that subtracts a small `GEOMETRY_EPSILON` (0.05px) from the floor, absorbing sub-pixel ULP noise from transform-mapped rects while keeping a genuine tap-target regression (≥1px, tiers ≥4px apart) failing hard. This directly addresses the flake (`51.99998… >= 52`) described in the task.

## Verification performed

- **Helper correctness** — `ui/src/lib/test-support/geometry.ts:27` computes `floorPx - GEOMETRY_EPSILON` = e.g. `51.95` for the 52 tier; well above the next-lower tier (48), so the a11y constraint (must not let 48 pass a 52 floor) holds. `GEOMETRY_EPSILON = 0.05` is 5× the issue's suggested ~0.01 but 20× below a 1px regression and reviewed/approved in the plan — not a defect.
- **`undefined` handling** — helper types `value: number | undefined`; `toBeGreaterThanOrEqual` on `undefined` still fails (undefined→NaN, `NaN >= x` is false / type-guard throws), so optional-chained call sites (`closeBox?.width`) behave unchanged, matching the docstring at `ui/src/lib/test-support/geometry.ts:22`.
- **`expect(value, label)` second-arg message pattern** — verified already in use in this codebase at `ui/src/lib/components/TopBar.browser.test.ts:293` and `ui/src/lib/components/viewport/ViewportTermControls.browser.test.ts:57`, so the `label` arg is valid vitest usage.
- **Import consistency** — helper imports `expect` from `"vitest"`; browser tests import `expect` from the same module (`ui/src/lib/components/TopBar.browser.test.ts:1`). Import path `$lib/test-support/geometry` resolves to the new file `ui/src/lib/test-support/geometry.ts`.
- **Contract test** — `ui/src/lib/test-support/geometry.test.ts` correctly asserts the sub-floor cases via `.toThrow()` (a failing inner `expect` throws synchronously) and passing cases via `.not.toThrow()`; the flake value `51.99998474121094` and `44 - 1.5e-5` pass, `51`/`48` against `52` throw. Sound.
- **Migration completeness** — grepped all `*.browser.test.ts`: the inline `getBoundingClientRect().(height|width)…toBeGreaterThanOrEqual` pattern now returns zero matches, and every surviving `toBeGreaterThanOrEqual` across the touched files is a documented out-of-scope case (positional `.left/.top/.right` bound, array-length, or computed-style ratio/font floor). No size-floor assertion was left un-migrated, and each of the 8 files gained the import.

No production code, i18n, feature-catalog, or user-facing strings are touched — this is a test-robustness change only, matching the plan's scope.

No blocking issues and no nits worth acting on.

PR #1854, opened 2026-07-20
feat(ui): redesign the new task modal into a single-view "calm form" layout

kai-osthoff:
Implements the **New Task modal redesign — Option 2a "Calm Form"** from `design_handoff_new_task_modal_2a/` (plan: `.shepherd-plan.md` on this branch's history). The modal becomes a single view: prompt hero + in-field toolbar left, Mode/Engine/Guards rail right, always-visible readiness footer + Create & Run CTA. Mobile is a full-height sheet with a combined repo·branch chip and an engine bottom-sheet.

### Screenshots (QA viewports; captured from the rendered component)
| Desktop 1280×800 | Mobile 390×844 | Engine sheet |
| --- | --- | --- |
| desktop | mobile | sheet |

(Reference material + screenshots live only in intermediate commits `750619dd`/`ce539cee` and are deleted again in `2e4ef083` — the PR diff carries no reference files.)

### What changed
- **Layout**: three-zone desktop modal (header / `1fr 300px` grid / footer), each body column owns its scroll, footer CTA always visible; preserved `.card`+`.bracket` surface (radius 0, registration brackets, no shadow). Mobile: fixed header/footer, single middle scroller, safe-area padding.
- **Mode segmented control** (design-system `.seg-row` recipe) replaces the Research / Create-EPIC checkboxes — exact checkbox-parity semantics incl. guard locking + pinning.
- **Guards** are instrument toggles (`role=switch`, full-row hit target, glossary-term labels, `statusTip` ON/OFF suffix showing the **real** repo default).
- **Engine group**: CLI select + ALPHA badge, compact capacity line (hottest window of the selected engine, `selectedProviderCapacity` layered on `providerCapacityRows`) with an all-engines popover, model select + compact guidance, Effort/Sandbox pair, one-line alpha caution with expandable details.
- **Inline `#` issue search** (new `issue-trigger.ts`; `slash.ts` untouched) + a show-all expansion in the issue panel; both render from one shared, generation-safe issue loader (one `listIssues` per repo selection).
- **Readiness model** (`new-task/readiness.ts`) drives the footer line + CTA state from the existing submit conditions. One deliberate, handoff-specified behavior change: *a seeded same-repo issue satisfies the prompt requirement* — the repo-aware `activeIssue` predicate feeds readiness, the submit guard, template materialization at spawn, and the `issueRef` payload alike (a stale cross-repo attachment is never sent).
[...]

PR #1916, opened 2026-07-23
fix(ui): make the new-task modal usable on mobile (controls + issue picker)

scoop:
## Problem

Reported from Android: opening the new-task sheet, typing a long prompt gives a full-screen text area with **no controls in view** — you have to scroll and manually dismiss the keyboard to reach Mode/Engine or Run. Separately, the **issue picker is gone on mobile** entirely (the `#1854` "calm form" redesign gated it to desktop). The related iOS-keyboard PR `#1855` is already merged; this branch is rebased on it and closes the two gaps it left.

## Two commits

**1 — `fix(ui): keep new-task controls reachable with the mobile keyboard up`**

Root cause: `autogrow()` writes an inline `height` that beats the mobile `height:100%`, and mobile drops the desktop `40vh` cap — so a long prompt grows the field without bound and pushes every control off-screen.

- `autogrow()` is inert on mobile; the textarea scrolls its own content. A `mobile`-keyed `$effect` clears a stale inline height inherited from desktop across the breakpoint (single mount point) and re-autogrows on the way back.
- The notice branches move into a `.nt-notices` scroller (`display:contents` on desktop), so `.left` stops scrolling as a whole and the Mode segments, Engine row, in-field toolbar and error are pinned by the existing `.left > * { flex-shrink: 0 }`.
- The hero is the flexible box; the textarea is taken **out of flow** (`position:absolute`) so a long prompt can't re-inflate the block's min-content. `.card` scrolls as a degenerate backstop; the dual CTA goes row in short-and-touch landscape.

**2 — `fix(ui): restore the issue/command picker on mobile`**

- The `# issue · / command` toolbar hint becomes an icon-only `#` trigger (mobile + a selected repo) opening a third `MobileEngineSheet` that renders `PromptSources` with the desktop props. Icon-only so it never wraps the flex-wrap toolbar beside attachment chips; the accessible name carries the full label.
- Picks close the sheet under a focus contract: a prompt-writing pick focuses the prompt (caret at end); an issue pick leaves focus to `use:dialog`'s restore (keyboard stays down) — matching the desktop rail.
- Scoped `.sheet:has(.ps-wrap)` overrides make `.ps-body` the sole scroller so `.ps-filter-bar`'s sticky works and `.ps-head` stays put; the engine/context sheets keep their own scroller. 44px tap floors on the sheet's tabs, rows and filter trigger. New EN+DE keys (`newtask_sources_open`, `newtask_sources_sheet_title`).

## Out of scope

Capping the field's height when the keyboard is **down** — confirmed with the reporter that pinning the controls is the whole ask. Keyboard down, the hero takes the card's free space (a large target for the primary input); keyboard up, it yields toward the toolbar floor while every control stays reachable.

## Testing

- `bun run check`, `bun run check:i18n`, prettier, eslint — clean.
- `bun run test` — 4247 passed, incl. **19 new browser fixtures**: autogrow inertness + stale-height crossing, controls hit-testable (`elementFromPoint`, `--ui-scale` 1 and 1.5), CTA/error never clipped at the worst geometry, notices-only-scroll, field-never-overlaps-toolbar, desktop-unchanged; plus the sources sheet (open/pick/close, single-sheet invariant, focus contract, `.ps-body` scroll chain + `:has()` no-leak, 44px floors, toolbar wrap budget, no-repo state, filter-popover positioning).
- Real Android/iOS keyboards can't be synthesized in the harness (same caveat as `#1855`) — final confirmation is on-device.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3319,6 +3319,8 @@
   "newtask_context_chip_aria": "Repository {repo}, von Branch {branch}",
   "newtask_engine_sheet_title": "Engine & Leitplanken",
   "newtask_context_sheet_title": "Repo & Branch",
+  "newtask_sources_open": "Issues & Befehle durchsuchen",
+  "newtask_sources_sheet_title": "Issues & Befehle",
   "newtask_gate_on": "Gate an",
   "newtask_gate_off": "Gate aus",
   "newtask_toggle_on": "AN",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -3319,6 +3319,8 @@
   "newtask_context_chip_aria": "Repository {repo}, from branch {branch}",
   "newtask_engine_sheet_title": "Engine & Guards",
   "newtask_context_sheet_title": "Repo & Branch",
+  "newtask_sources_open": "Browse issues & commands",
+  "newtask_sources_sheet_title": "Issues & Commands",
   "newtask_gate_on": "gate on",
   "newtask_gate_off": "gate off",
   "newtask_toggle_on": "ON",
--- a/ui/src/lib/components/NewTask.browser.test.ts
+++ b/ui/src/lib/components/NewTask.browser.test.ts
@@ -3334,3 +3334,234 @@ describe("NewTask mobile controls stay reachable while typing", () => {
     expect(getComputedStyle(document.querySelector<HTMLElement>(".left")!).overflowY).toBe("auto");
   });
 });
+
+// #1854 dropped the issue/command picker (PromptSources) on mobile — the only path left was
+// typing `#`. These fixtures cover the restored mobile sources sheet: the toolbar trigger, the
+// sheet's scroll chain, the focus contract, and a11y floors.
+describe("NewTask mobile sources sheet", () => {
+  class FakeVisualViewport extends EventTarget {
+    height: number;
+    offsetTop: number;
+    constructor(height: number, offsetTop = 0) {
+      super();
+      this.height = height;
+      this.offsetTop = offsetTop;
+    }
+  }
+  function installFakeViewport(height: number, offsetTop = 0) {
+    const fake = new FakeVisualViewport(height, offsetTop);
+    const prev = Object.getOwnPropertyDescriptor(window, "visualViewport");
+    Object.defineProperty(window, "visualViewport", { configurable: true, value: fake });
+    const restore = () => {
+      if (prev) Object.defineProperty(window, "visualViewport", prev);
+      else delete (window as unknown as { visualViewport?: unknown }).visualViewport;
+    };
+    return { fake, restore };
+  }
+  const srcBtn = () => document.querySelector<HTMLButtonElement>(".sources-btn");
+  const psWrap = () => document.querySelector<HTMLElement>(".ps-wrap");
+
+  it("opens the sheet from the toolbar trigger and lists issues", async () => {
+    await page.viewport(390, 844);
+    seedIssues([mkIssue(1, "alpha"), mkIssue(2, "bravo")]);
+    mockListRepos.mockResolvedValue({ repos: [], recentWindowDays: 30 });
+    render(NewTask, { props: { onsubmit: vi.fn(), initialRepoPath: "/repo" } });
+    await expect.poll(() => srcBtn()).toBeTruthy();
+
+    srcBtn()!.click();
+    await expect.poll(() => psWrap()).toBeTruthy();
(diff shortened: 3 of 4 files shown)


scoop (commented):
## Review

The PR restores mobile UX in the redesigned new-task modal along two axes — bounding the runaway prompt field so the controls stay reachable (commit 1), and re-adding the issues/commands picker as a bottom sheet (commit 2). The implementation is coherent and I found no blocking defects.

### What I verified (ground truth cited)

- **i18n parity.** Both new keys exist in both catalogs: `newtask_sources_open` / `newtask_sources_sheet_title` at `ui/messages/en.json:3322-3323` and `ui/messages/de.json:3322-3323`. The reused `newtask_syntax_hint_touch` is present in both catalogs as well.
- **Sheet callbacks resolve.** `pickIssue` (`NewTask.svelte:758`), `injectSteer` (`:768`), `pickCommandFromSource` (`:884`) all exist and are the same handlers the desktop rail wires (`:1574-1576`). PromptSources props in the new sheet (`:1731-1738`) match the desktop rail's prop set (`:1558-1565`) exactly; the sheet only renders when `activeSheet==="sources"`, reachable only via the `#if repoPath` button, so `repoPath` is guaranteed non-null just like the rail's `repoPath && !mobile` gate.
- **PromptSources scroll-chain selectors are real.** `.ps-wrap`/`.ps-head`/`.ps-body`/`.ps-filter-bar`/`.tab`/`.issue-list-row` exist in `PromptSources.svelte:288-402`; `.filter-chip` in `IssueFilterPopover.svelte:129`. `MobileEngineSheet.svelte:34` renders `.sheet` + `.sheet-body`, so the `:has(.ps-wrap)` scoping targets the right nodes and leaves the engine/context sheets' `.sheet-body` scroller untouched.
- **Focus contract holds.** `a11yDialog.ts:92-98` restores focus to the pre-open element on `destroy()` (not via `onclose`), so `closeSheetFocusPrompt()` (`NewTask.svelte:783-789`) correctly sets `activeSheet=null` → `await tick()` (destroy runs, focus lands on the trigger) → `promptInput.focus()` overrides; `closeSheetKeepFocus()` leaves the dialog restore in place so the keyboard stays down for issue picks. Matches the stated contract.
- **autogrow / stale-height.** `autogrow()` is inert on mobile (`:803`); `promptInput` is `$state` (`:295`) so the `mobile`-keyed `$effect` (`:330-336`) re-fires when the element mounts and clears the stale inline height. Desktop textarea keeps its `min-height:132px`/`max-height:40vh` (`:2105-2109`) — the removed floors were mobile-only overrides.
- **Layout mechanism.** The mobile floor works: the textarea is `position:absolute; inset:0` so it contributes nothing to `.prompt-block`'s min-content, whose floor is `.prompt-wrap { min-height:3rem }` + the `flex-shrink:0` toolbar; `.prompt-block { flex:1 1 0 }` grows into free space; siblings stay pinned by the base `.left > * { flex-shrink:0 }` (`:1933`); `.nt-notices` is the only shrinkable scroller; `.card { overflow-y:auto }` is the degenerate backstop. The landscape `.run-dual { flex-direction:row }` override (`:2732`) is later in source than the stacked mobile rule so it wins where both queries match.

### Non-blocking observations

- `NewTask.svelte`: the always-rendered `.nt-notices` div is a flex item of `.left` (which has `gap:12px`), so even when every notice `{#if}` is false it still contributes one 12px gap above `.prompt-block`. The plan claimed "empty → no notice costs no space"; in practice it costs ~12px of dead top space. Purely cosmetic.
- `NewTask.svelte:2557`: `.sources-btn` sizes the `#` glyph at `var(--fs-lg)` rather than the `--fs-meta` the plan specified to match `.char-count`/`.seg-btn`; the `#` will render slightly larger than the text it replaces. Cosmetic only (could not locate a DESIGN.md rule in-repo to treat this as a hard constraint).
- `NewTask.svelte:2045`: on desktop `.nt-notices { display: contents }` removes its box, so the child-combinator rule `.left > * { flex-shrink:0 }` no longer matches the notice children (selectors use the DOM tree, not the box tree) — they revert to `flex-shrink:1`. Imperceptible for the short single-line notices, but the diff comment's "flow exactly as before" is not literally true.

### Could not verify
- Real Android soft-keyboard geometry cannot be synthesized in the harness (matching `#1855`'s caveat); final on-device confirmation is out of scope for static review.

Desktop rendering paths (in-flow textarea, whole-column scroll, inline menus, `autogrow` growth under the 40vh cap) are untouched by the mobile-scoped overrides.

PR #1937, opened 2026-07-24
feat(models): let operators pin Claude Opus 5 exactly

scoop:
## What

Adds `claude-opus-5` and `claude-opus-5[1m]` to the Claude model list, **alongside** the existing floating `opus` / `opus[1m]` entries. Purely additive — no migration of persisted default-model, per-repo, per-role, or usage-downgrade settings.

## Why this is not a no-op

`opus` already runs Opus 5. Probed against the pinned CLI (`claude` 2.1.219):

| `--model` | wire model | `context-1m-2025-08-07` |
|---|---|---|
| `opus` | `claude-opus-5` | no |
| `opus[1m]` | `claude-opus-5` | **yes** |
| `claude-opus-5` | `claude-opus-5` | no |
| `claude-opus-5[1m]` | `claude-opus-5` | **yes** |
| `opus-5` | — | rejected: "may not exist" |

So the gap was never reach — it was **pinning**. `CLAUDE_MODELS` is a closed allowlist, so an operator could only ride the floating alias and had no way to lock a version against a future Opus release. The pinnable token is the full model name; the short `opus-5` form errors.

The 1M variant was header-probed rather than assumed (a capture stub behind `ANTHROPIC_BASE_URL` logging `anthropic-beta`): `claude-opus-5[1m]` emits a beta set **byte-identical** to `opus[1m]`'s, so the long-context / premium / `is1mModel` treatment is earned, not guessed.

## Record vs configured labels

The two Opus rows would otherwise sit side by side with the same fit tag and cost mark, distinguishable only by a raw token. So labelling is now split by **tense**:

- `modelLabel()` — the **record** label. Session cards, status bar, plan panel, review-in-flight banner, viewport, experiment groups. Floating aliases stay bare here on purpose: labelling an archived session "Opus (latest)" would claim it ran today's model when it ran whichever Opus was current at spawn.
- `configuredModelLabel()` — the **configured** label, for runs that haven't happened: all picker rows, the New Task engine summary, the per-role effective-environment line, the recommend dialog header. There "latest" is present-tense true.

Pinned names read identically on both surfaces — a pinned version means the same model forever.

## Notes

- **Pricing unchanged.** Opus 5 lists at the same $5/$25 as 4.8 and `/opus/i` already matched `claude-opus-5`; only a stale "Opus 4.8" comment moved.
[...]

PR #1938, opened 2026-07-24
ci: gate model-list mirror between server + ui types

### LINKED ISSUE #1936: Gate the CLAUDE_MODELS mirror between src/types.ts and ui/src/lib/types.ts
## Problem

`CLAUDE_MODELS` (and `CODEX_MODELS`, `EFFORTS`) are declared twice — once in `src/types.ts` (server: validation, default-model setting space, per-role/per-repo overrides) and once, hand-copied, in `ui/src/lib/types.ts` (pickers). **Nothing enforces that the two stay identical.**

A one-sided edit is silent and asymmetric:

- add a model to the UI list only → the picker offers a value `validateCreate` rejects with `unknown model` at task-create time;
- add it to the server list only → the value is accepted by the API but unreachable from any picker.

This is a pre-existing hazard, not one introduced by any particular change.

## Proposed fix

`scripts/check-model-mirror.mjs` — extract the `CLAUDE_MODELS` and `CODEX_MODELS` array literals from both files textually and fail on any element/order mismatch, with a fix hint naming both paths. Wire it up exactly like the existing gates:

- `check:model-mirror` script in the root `package.json`
- an entry in `scripts/pre-push.ts`'s gate list
- a step in `.github/workflows/pr-hygiene.yml`, alongside `check-glossary.mjs` / `check-feature-catalog.sh`

**Textual comparison rather than a cross-package test import** is deliberate: the root `tsconfig.json` sets `exclude: ["ui"]`, so a test importing both halves would cross a boundary the build config draws.

`EFFORTS` is mirrored the same way and is a natural second constant to cover — the parser is the same few lines.

## Context

Split out of #1937 (pin Claude Opus 5), where the gate was originally bundled. It was separated on operator decision: it is hygiene infrastructure for a pre-existing hazard and does not belong in a PR scoped to adding a model. That PR's interim mitigation is that both halves are edited together and each is exercised by its own suite (`test/default-model.test.ts` server-side, `ui/src/lib/model-guidance.test.ts` UI-side), so a one-sided edit fails one of them — but only for the entries those tests happen to name, which is why the gate is still worth having.

scoop:
Closes #1936.

## The hazard

`CLAUDE_MODELS`, `CODEX_MODELS` and `EFFORTS` are each declared **twice** — in `src/types.ts` (validation, the default-model setting space, per-role/per-repo overrides) and hand-copied into `ui/src/lib/types.ts` (pickers). Nothing enforced that the two stay identical, and a one-sided edit was silent in both directions, asymmetrically:

- **UI-only add** → the picker offers a value `validateCreate` rejects with `unknown model` at task-create time.
- **Server-only add** → the API accepts the value but no picker can reach it.

Pre-existing hazard, not introduced by any particular change. Split out of #1937 on operator decision.

## What this adds

`scripts/check-model-mirror.mjs` compares the three literals textually — **element and order** — and is wired in exactly like every sibling hygiene gate: a `check:model-mirror` package script, a step in `scripts/pre-push.ts`'s `gates` lane, and a step in `.github/workflows/pr-hygiene.yml`.

**No list contents change.** `src/types.ts` and `ui/src/lib/types.ts` are byte-identical to `main`; this is pure hygiene infrastructure landing on an already-in-sync tree.

### Two parser decisions worth reviewing

Both are traps a looser implementation falls into, and both are covered by tests:

1. **The close is a string-aware forward scan, not a `] as const` anchor.** Elements contain `]` (`"opus[1m]"`, `"claude-opus-5[1m]"`, `"sonnet[1m]"`), so a "first `]`" scan truncates the Claude list. But anchoring on `as const` is worse: it is incidental syntax, so the scan runs past any array written without it — a shape already present in the same file (`export const PREMIUM_MODELS: readonly string[] = [ … ];`). In `src/types.ts` the next `] as const` after `CLAUDE_MODELS` is `EFFORTS`', so that over-match would have silently absorbed the five effort tiers into the model list.
2. **The opening anchor requires the `const` keyword** rather than the loose `<NAME>[^=]*=` idiom used elsewhere in `scripts/`, because these names appear in **prose right next to their declarations** — `ui/src/lib/types.ts`'s doc comment above `CLAUDE_MODELS` says "Mirrors src/types.ts CLAUDE_MODELS", and `src/types.ts`'s `CODEX_MODEL_RE` comment names `CODEX_MODELS` just below that array. A loose anchor happens to still land correctly today, which is what makes it a latent trap: one comment reword from extracting the wrong span.

It **fails closed** — a literal unparseable on *either* side (rename, reshape, no string elements) is an error, never a skipped comparison, so two absent literals can't compare as equal empties.

`compareMirror` returns structured deltas (`{ constant, missingIn, onlyInServer, onlyInUi, orderMismatch }`); prose lives only in the CLI shell, so tests assert on data rather than substring-matching a message that is free to be reworded.

### Why textual rather than importing both halves into one test

For the record, since a plausible-sounding wrong reason exists: it is **not** because the root `tsconfig.json` excludes `ui` — `exclude` only filters the root file set, an import still pulls a file in, and Bun's test runner ignores it entirely. `ui/src/lib/types.ts` has zero imports, so `expect(UI).toEqual(SERVER)` genuinely would work. Rejected because (a) all three wiring points invoke a gate as a bare command, which a `bun test` assertion can't fill, and (b) it would pull 2246 lines of UI types into the server `tsc` program, checked against the server tsconfig instead of the `svelte-check` config that owns them.

## Verification

- `bun run check:model-mirror` passes on this tree — `CLAUDE_MODELS (8), CODEX_MODELS (11), EFFORTS (5)`.
- Injected one-sided drift fails correctly, naming **both** paths: a UI-only add, a UI-only reorder, and a UI-side rename (fail-closed) each exit 1 with a distinct message. Edits reverted; both types files verified byte-identical to `main`.
- `test/check-model-mirror.test.ts` — 19 tests, incl. the bracketed-element, no-`as const`, adjacent-doc-comment (above *and* below), one-sided `missingIn`, and absent-on-both cases, plus a live-tree assertion over the real files.
- Full `pre-push` suite green (7/7 lanes, fallow clean), with `model mirror` confirmed executing in the `gates` lane log.

## Deliberately out of scope

Surveyed, not assumed — every other place a model/effort alias is written by hand:

- **Duplicated `EFFORTS` *logic*** — `src/default-effort.ts` `effortsForProvider` ↔ `ui/src/lib/effort-guidance.ts` `providerEfforts` have byte-identical bodies, and `effortBelowHigh` is duplicated across the same pair; the comments self-declare it ("keep the two byte-identical in behavior"). Genuinely the same hazard class, but it duplicates **logic, not an array literal**, so a textual extractor structurally cannot gate it — that needs behavioural equivalence (shared module or a table-driven test), a separate change. Each side has its own suite today.
- **`extension/`** — `CaptureConfig["model"]` and `Options.svelte`'s `models` are a deliberate *subset* (no `[1m]` variants, no pinned names, plus a `"default"` sentinel absent from `CLAUDE_MODELS`), so equality is the wrong assertion; a subset check is a different invariant. `models` is annotated `CaptureConfig["model"][]`, so `svelte-check` already rejects off-union values.
- **Prose enumerations** — `docs/external-task-api.md`'s `model` row and `docs-site/.../reference/configuration.md`'s `SHEPHERD_DOC_AGENT_EFFORT` row each hand-list a full set in order inside a Markdown table cell. Not generated, so `check-generated-docs.sh` doesn't reach them; prose-shaped targets, left to review.
- **`PREMIUM_MODELS`** — UI-only, no server counterpart, so not a mirror.
- **Not added to `ci.yml`'s `verify` job** — `pr-hygiene.yml` is where every sibling hygiene gate lives.

One known coverage note: eslint's root glob is `'scripts/**/*.ts'`, so it never sees the new `.mjs`/`.d.mts` — the same pre-existing gap that already leaves `check-glossary.mjs` and `next-version.mjs` unlinted. Surfaced rather than widened here, since broadening the glob would newly lint every existing `scripts/*.mjs`. Both files are covered by prettier, by the unit tests, and (for the `.d.mts`) by `tsc` through the test — verified resolving to real types, not `any`.

### DIFF
--- a/.github/workflows/pr-hygiene.yml
+++ b/.github/workflows/pr-hygiene.yml
@@ -55,3 +55,5 @@ jobs:
         run: node scripts/check-glossary.mjs
       - name: Announcement versions (new entry ⇒ unreleased version)
         run: node scripts/check-announcement-versions.mjs
+      - name: Model-list mirror (src/types.ts ↔ ui/src/lib/types.ts)
+        run: node scripts/check-model-mirror.mjs
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,7 +76,7 @@ and failed pushes. Independent work runs in parallel (bounded by your core count
 lane teeing to its own `.test-logs/` file, with a per-lane wall-clock **timeout backstop**
 so a hung child can never wedge the push. The checks (per lane) are:
 
-- **gates:** branch-hygiene · feature-catalog · generated-docs · glossary · announcement-versions · herdr-types
+- **gates:** branch-hygiene · feature-catalog · generated-docs · glossary · announcement-versions · model-mirror · herdr-types
 - **prettier:** `prettier --check` over the push **delta** (see note)
 - **eslint:** root + extension eslint over the push **delta** (see note)
 - **tsc:** `bun run typecheck` (root `tsc --noEmit`)
@@ -139,6 +139,7 @@ bun run lint                 # eslint (root, covers ui/src too)
 bun run typecheck            # root tsc --noEmit (src + test)
 bun run format               # prettier --write across the repo
 bun test ./test              # core test suite
+bun run check:model-mirror   # model/effort lists identical in src/types.ts ↔ ui/src/lib/types.ts
 cd ui && bun run check       # svelte-check (ui types)
 cd ui && bun run check:i18n  # locale-catalog parity (en ↔ de)
 cd ui && bun run test        # ui test suite (vitest)
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "onboarding:test": "bun run ci/onboarding-harness/run.ts",
     "check:glossary": "node scripts/check-glossary.mjs",
     "check:announcement-versions": "node scripts/check-announcement-versions.mjs",
+    "check:model-mirror": "node scripts/check-model-mirror.mjs",
     "next-version": "node scripts/next-version.mjs",
     "gen:herdr-schema": "bun run scripts/gen-herdr-schema.ts",
     "gen:herdr-types": "bun run scripts/gen-herdr-types.ts",
--- a/scripts/check-model-mirror.d.mts
+++ b/scripts/check-model-mirror.d.mts
@@ -1,0 +1,47 @@
+// Types for the node-run model-mirror gate (scripts/check-model-mirror.mjs). The
+// gate stays plain .mjs so `node scripts/check-model-mirror.mjs` works as a CLI in
+// any context (the package script, the pre-push gates lane, PR hygiene); this
+// declaration only exists so its TypeScript importer — test/check-model-mirror.test.ts
+// — gets real types instead of `any`. Mirrors scripts/next-version.d.mts's role.
+
+/** Which side of the mirror a finding refers to. */
+export type MirrorSide = "server" | "ui";
+
+/** One constant's divergence. Empty arrays + false flags = that constant is fine. */
+export interface MirrorDelta {
+  /** The constant that diverged, e.g. "CLAUDE_MODELS". */
+  constant: string;
+  /** Sides whose array literal could not be parsed at all (fail-closed). */
+  missingIn: MirrorSide[];
+  /** Elements in src/types.ts but not ui/src/lib/types.ts. */
+  onlyInServer: string[];
+  /** Elements in ui/src/lib/types.ts but not src/types.ts. */
+  onlyInUi: string[];
+  /**
+   * Membership matches but the sequences differ — a reorder, or (degenerately) a
+   * differing number of duplicate entries. Only ever set when both `onlyIn*` are empty.
+   */
+  orderMismatch: boolean;
+}
+
+export interface MirrorResult {
+  /** True when `deltas` is empty — the two files agree on every constant. */
+  ok: boolean;
+  deltas: MirrorDelta[];
+}
+
+/** Constants declared in both files that must stay element- and order-identical. */
+export const MIRRORED_CONSTANTS: string[];
+
+/**
+ * String elements of `const <constName> = [ … ]`, or null when the literal cannot
+ * be parsed (absent, unterminated, or holding no string elements).
+ */
+export function extractArrayLiteral(source: string, constName: string): string[] | null;
+
+/** Compare every mirrored constant across the two sources, as structured deltas. */
+export function compareMirror(
+  serverSource: string,
+  uiSource: string,
+  constants?: string[],
+): MirrorResult;
--- a/scripts/check-model-mirror.mjs
+++ b/scripts/check-model-mirror.mjs
@@ -198,8 +198,12 @@ export function compareMirror(serverSource, uiSource, constants = MIRRORED_CONST
     const onlyInServer = server.filter((v) => !ui.includes(v));
     const onlyInUi = ui.filter((v) => !server.includes(v));
     // Only meaningful once membership matches: same elements, different sequence.
+    // The separator is written as the ESCAPE "\0", never a literal NUL byte — an
+    // embedded NUL makes this file binary to grep/ripgrep/file(1) (matching lines
+    // are suppressed) and invisible in review. NUL is the separator because it
+    // cannot occur in a model alias, so no pair of distinct lists can join equal.
     const orderMismatch =
-      onlyInServer.length === 0 && onlyInUi.length === 0 && server.join(" ") !== ui.join(" ");
+      onlyInServer.length === 0 && onlyInUi.length === 0 && server.join("\0") !== ui.join("\0");
 
     if (onlyInServer.length || onlyInUi.length || orderMismatch) {
       deltas.push({ constant, missingIn: [], onlyInServer, onlyInUi, orderMismatch });
@@ -247,7 +251,13 @@ function formatDelta(delta) {
   return lines;
 }
 
-if (import.meta.url === `file://${process.argv[1]}`) {
+// `fileURLToPath` rather than string-concatenating `file://` + argv[1]: import.meta.url
+// is PERCENT-ENCODED, so on any checkout path containing a space, `#`, `%` or non-ASCII
+// the concat comparison is false, the CLI block never runs, and the gate exits 0 having
+// checked nothing — the exact vacuous pass this script exists to prevent. Same idiom as
+// scripts/json-union-merge.mjs. (scripts/next-version.mjs still uses the concat form.)
+const isMain = Boolean(process.argv[1]) && fileURLToPath(import.meta.url) === process.argv[1];
+if (isMain) {
   const serverSource = readFileSync(join(ROOT, SERVER_REL), "utf8");
   const uiSource = readFileSync(join(ROOT, UI_REL), "utf8");
   const { ok, deltas } = compareMirror(serverSource, uiSource);
--- a/scripts/check-model-mirror.mjs
+++ b/scripts/check-model-mirror.mjs
@@ -1,0 +1,268 @@
+#!/usr/bin/env node
+// Model-list mirror gate (#1936): the model/effort alias lists declared in BOTH
+// `src/types.ts` and `ui/src/lib/types.ts` must stay element- AND order-identical.
+//
+// THE HAZARD this closes: `CLAUDE_MODELS`, `CODEX_MODELS` and `EFFORTS` are each
+// declared twice — once server-side (validateCreate validation, the default-model
+// setting space, per-role/per-repo overrides) and once, hand-copied, in the UI
+// (pickers). Nothing enforced that the two stay identical, and a one-sided edit is
+// silent in BOTH directions, asymmetrically:
+//   • UI-only add    → the picker offers a value the server rejects with
+//                      `unknown model` at task-create time.
+//   • server-only add → the API accepts the value but no picker can reach it.
+//
+// WHY TEXTUAL rather than importing both halves into one test: all three wiring
+// points (the `check:model-mirror` package script, the pre-push `gates` lane, the
+// PR-hygiene workflow) invoke a gate as a bare command, which a `bun test`
+// assertion cannot fill; and importing `ui/src/lib/types.ts` from `test/` would
+// pull the whole UI type surface into the server `tsc` program, type-checked
+// against the server tsconfig instead of the `svelte-check` config that owns it.
+// (Note for the record: the root tsconfig's `exclude: ["ui", …]` is NOT the
+// reason — `exclude` only filters the root file set, and an import still pulls a
+// file in.)
+//
+// ORDER-SENSITIVE, not set-equality: on the UI side the literal order IS the order
+// the model dropdown renders, so a one-sided reorder is drift worth failing on.
+//
+// The parser is deliberately conservative — see `extractArrayLiteral` for the two
+// traps (prose naming the constant next to its declaration; a missing `as const`)
+// that a looser implementation falls into. It fails CLOSED: a literal that cannot
+// be parsed on EITHER side is an error, never a skipped comparison.
+//
+// Plain ESM — no dependencies, no transpile. Mirrors the shape and style of
+// scripts/check-glossary.mjs. Importable (extractArrayLiteral / compareMirror /
+// MIRRORED_CONSTANTS) by test/check-model-mirror.test.ts.
+
+import { readFileSync } from "node:fs";
+import { dirname, join, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+// Resolved from this file, never process.cwd(), so the CLI works from any directory.
+const ROOT = resolve(dirname(fileURLToPath(import.meta.url)), "..");
+const SERVER_REL = "src/types.ts";
+const UI_REL = "ui/src/lib/types.ts";
+
+/**
+ * The constants declared in both files that must stay identical.
+ *
+ * | Constant        | src/types.ts           | ui/src/lib/types.ts    |
+ * | --------------- | ---------------------- | ---------------------- |
+ * | `CLAUDE_MODELS` | `const` (non-exported) | `const` (non-exported) |
+ * | `CODEX_MODELS`  | `export const`         | `export const`         |
+ * | `EFFORTS`       | `export const`         | `export const`         |
+ *
+ * `MODELS_BY_PROVIDER` needs no entry — it is DERIVED from the two model arrays in
+ * both files, so gating the arrays covers it transitively. `PREMIUM_MODELS` is
+ * UI-only (no server counterpart), so it is not a mirror at all.
+ */
+export const MIRRORED_CONSTANTS = ["CLAUDE_MODELS", "CODEX_MODELS", "EFFORTS"];
+
+/**
+ * Walk a `[ … ]` literal from its opening bracket, returning the body with
+ * comments stripped, or null if it is never closed.
+ *
+ * The scan is string- and comment-aware because the real data demands it:
+ *   • elements CONTAIN `]` — "opus[1m]", "claude-opus-5[1m]", "sonnet[1m]" — so a
+ *     naive "up to the first `]`" scan truncates the Claude list mid-way;
+ *   • an apostrophe in a comment ("// don't …") would otherwise open a bogus
+ *     string state and swallow the rest of the file.
+ * Bracket depth is tracked so a nested array can't terminate the scan early.
+ */
+function scanArrayBody(source, openIndex) {
+  let depth = 0;
+  let quote = null;
+  let comment = null; // "line" | "block"
+  let body = "";
+
+  for (let i = openIndex; i < source.length; i += 1) {
+    const ch = source[i];
+
+    if (comment === "line") {
+      if (ch === "\n") comment = null;
+      continue;
+    }
+    if (comment === "block") {
+      if (ch === "*" && source[i + 1] === "/") {
+        comment = null;
+        i += 1;
+      }
+      continue;
+    }
+    if (quote) {
+      body += ch;
+      if (ch === "\\") {
+        // Escaped char is consumed verbatim so a trailing \" can't end the string.
+        body += source[i + 1] ?? "";
+        i += 1;
+      } else if (ch === quote) {
+        quote = null;
+      }
+      continue;
+    }
+    if (ch === "/" && source[i + 1] === "/") {
+      comment = "line";
+      i += 1;
+      continue;
+    }
+    if (ch === "/" && source[i + 1] === "*") {
+      comment = "block";
+      i += 1;
+      continue;
+    }
+    if (ch === '"' || ch === "'" || ch === "`") {
+      quote = ch;
+      body += ch;
+      continue;
+    }
+    if (ch === "[") {
+      depth += 1;
+      if (depth > 1) body += ch;
+      continue;
+    }
+    if (ch === "]") {
+      depth -= 1;
+      if (depth === 0) return body;
+      body += ch;
+      continue;
+    }
+    body += ch;
+  }
+
+  return null; // unterminated — caller treats it as unparseable
+}
+
+/**
+ * Extract the string elements of `const <constName> = [ … ]` from a TS source, or
+ * null when the literal cannot be parsed. Two traps this deliberately avoids:
+ *
+ *  1. The opening anchor REQUIRES the `const` keyword rather than the loose
+ *     `<NAME>[^=]*=` idiom used elsewhere in scripts/, because these constant
+ *     names appear in PROSE right next to their declarations — the doc comment
+ *     above `ui/src/lib/types.ts`'s CLAUDE_MODELS reads "Mirrors src/types.ts
+ *     CLAUDE_MODELS", and src/types.ts's CODEX_MODEL_RE comment names
+ *     CODEX_MODELS just below that array. A loose anchor happens to still land on
+ *     the real declaration today, which is exactly what makes it a latent trap:
+ *     one comment reword from matching prose and extracting the wrong span.
+ *  2. The close is found by `scanArrayBody`, NOT by anchoring on `] as const`.
+ *     `as const` is incidental syntax, and an `as const`-anchored scan runs PAST
+ *     any array written without it — a shape already present in the same file
+ *     (`export const PREMIUM_MODELS: readonly string[] = [ … ];`) — all the way to
+ *     the next `] as const`. In src/types.ts that is EFFORTS', so a CLAUDE_MODELS
+ *     over-match would silently absorb the five effort tiers.
+ *
+ * An empty element list is reported as unparseable: none of the gated lists is
+ * ever legitimately empty, so zero elements means the shape changed (identifiers
+ * or an enum instead of string literals) and comparing two empty lists as "equal"
+ * would be a vacuous pass.
+ */
+export function extractArrayLiteral(source, constName) {
+  // Optional `export`, required `const`, optional type annotation (`: readonly
+  // string[]`), then `= [`. \b rejects a longer name with the same prefix.
+  const anchor = new RegExp(`(?:export\\s+)?const\\s+${constName}\\b\\s*(?::[^=]*)?=\\s*\\[`);
+  const match = anchor.exec(source);
+  if (!match) return null;
+
+  const body = scanArrayBody(source, match.index + match[0].length - 1);
+  if (body === null) return null;
+
+  const elements = [];
+  const elementRe = /"((?:[^"\\]|\\.)*)"|'((?:[^'\\]|\\.)*)'/g;
+  let el;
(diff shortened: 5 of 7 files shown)


scoop:
## What I verified

**The gate passes on this tree, and the parser matches the real files' shapes.** Compared element-by-element:

- `CLAUDE_MODELS` — `src/types.ts:315-324` (8 elements: `fable, opus, opus[1m], claude-opus-5, claude-opus-5[1m], sonnet, sonnet[1m], haiku`) vs `ui/src/lib/types.ts:1986-1995`: identical elements, identical order. Both are bare `const` (non-exported), as the script's table at `scripts/check-model-mirror.mjs:48-52` claims.
- `CODEX_MODELS` — `src/types.ts:337-349` (11 elements) vs `ui/src/lib/types.ts:2006-2018`: identical elements, identical order.
- `EFFORTS` — `src/types.ts:333` vs `ui/src/lib/types.ts:2003`: byte-identical single-line literals.

**The two parser traps the script guards against are real, not invented.** The strict `const` anchor at `scripts/check-model-mirror.mjs:161` is needed because `ui/src/lib/types.ts:1985` ("`CLAUDE_MODELS` — keep both lists identical"), `ui/src/lib/types.ts:2000` ("Mirrors src/types.ts EFFORTS") and `src/types.ts:352` ("the curated / CODEX_MODELS list does") all name the constants in prose adjacent to the declarations; none of the three is preceded by `const`, so the anchor lands on the real declarations in every case. The no-`as const` shape the scan-based terminator exists for is also real: `ui/src/lib/types.ts:2032-2039` (`export const PREMIUM_MODELS: readonly string[] = [ … ];`, plain `];`), and `src/types.ts:333` (`EFFORTS`) is indeed the next `] as const` after `CLAUDE_MODELS`, so an `as const`-anchored scan would have absorbed the effort tiers exactly as documented.

Tracing `scanArrayBody` by hand against `src/types.ts:315`: depth opens at the `[`, `"opus[1m]"`'s `]` is consumed inside quote state, and the scan terminates on the `]` at `src/types.ts:324` — 8 elements, no truncation. The `(?::[^=]*)?` type-annotation group cannot bridge to a later `= [` because `[^=]*` excludes `=`, so a reshaped declaration (`const CLAUDE_MODELS = SOMETHING_ELSE;`) yields no match → `missingIn` → non-zero exit, per the fail-closed rule.

**Wiring is complete and consistent with the siblings.** `package.json:19` (`check:model-mirror`), `scripts/pre-push.ts:400-405` (inserted in the `gates` lane between `announcement versions` at `:394-399` and `herdr types` at `:406`), `.github/workflows/pr-hygiene.yml:58-59` (after `Announcement versions` at `:56-57`), `CONTRIBUTING.md:79` + `:142`. The CONTRIBUTING gate list order matches the actual lane order. `pr-hygiene.yml` is the correct home rather than `ci.yml`: `.github/workflows/ci.yml` carries only `check:herdr-types` (`:121`) and none of the other hygiene gates. Adding a lane *step* needs no counter update — `plannedLaneCount` at `scripts/pre-push.ts:348-358` counts lanes, not steps.

**The `.d.mts` pattern is established, not novel.** `test/next-version.test.ts:2` already imports `../scripts/next-version.mjs` with types supplied by `scripts/next-version.d.mts:1-16`; `tsconfig.json`'s `exclude` (`node_modules, ui, extension, site, docs-site`) does not exclude `scripts/`, so the new declarations are in the typecheck program the same way. The test imports only values, so `verbatimModuleSyntax: true` is not an issue.

No line in any touched file exceeds `printWidth: 100` (`.prettierrc`).

## Problems

### 1. `scripts/check-model-mirror.mjs:202` contains two raw NUL (U+0000) bytes

Verified with `grep -abo -P '\x00'`: NUL bytes at byte offsets **8743** and **8760**, both on line 202, used as the `Array.prototype.join` separator:

```js
server.join("<NUL>") !== ui.join("<NUL>")
```

A NUL separator is the *right* choice semantically — I'm not asking to change it to a space — but it must be written as the escape `"\0"`, not embedded as a raw byte. Consequences of the raw byte, verified:

- **The file is binary to grep/ripgrep/`git grep`.** `Grep` on it returns `binary file matches (found "\0" byte around offset 8743)` with no content, and a repo-wide search for `CLAUDE_MODELS` returns 7 files — `ui/src/lib/types.ts`, `src/types.ts`, `src/default-model.ts`, two Svelte components, `test/check-model-mirror.test.ts`, `scripts/check-model-mirror.d.mts` — and **not** `scripts/check-model-mirror.mjs`, even though that file mentions `CLAUDE_MODELS` five times. The gate is invisible to exactly the search a future contributor would run before editing a model list.
- **Diff rendering is one edit away from breaking.** The file is 11267 bytes (`wc -c`) and git's binary sniff only inspects the first 8000, which is why `git diff` still renders it as text today. Trimming ~750 bytes of the header comment flips it to `Binary files differ` in every future diff and PR view.
- **It is unreviewable.** In the diff and in an editor line 202 reads as `join("")`; nothing tells a reader what the separator is.

Fix: replace the two raw bytes with `\0` (or `\u0000`). Runtime behaviour is unchanged; line 202 stays under 100 chars.

### 2. `scripts/check-model-mirror.mjs:250` — the main-module guard can make the gate exit 0 having checked nothing

```js
if (import.meta.url === `file://${process.argv[1]}`) {
```

`import.meta.url` is a percent-encoded URL; `process.argv[1]` is a raw filesystem path. They are equal only when the absolute path contains no character URL-encoding touches. On a checkout whose path contains a space, `#`, `%` or any non-ASCII character, `import.meta.url` is e.g. `file:///home/u/My%20Repos/shepherd/scripts/check-model-mirror.mjs`, the comparison is false, the whole CLI block is skipped, and `node scripts/check-model-mirror.mjs` **exits 0 with no output** — the `check:model-mirror` script, the pre-push `model mirror` step and the CI step all report success without reading either file.

That is precisely the "vacuous pass" the script's own header (`scripts/check-model-mirror.mjs:29-30`) and the fail-closed design elsewhere in the file are built to prevent, so it is worth not inheriting. This repo already has the robust idiom: `scripts/json-union-merge.mjs:156` uses `fileURLToPath(import.meta.url) === process.argv[1]`, and `fileURLToPath` is already imported here at `scripts/check-model-mirror.mjs:38`. (GitHub Actions' `/home/runner/work/...` path is unaffected, so CI coverage is not silently lost — but a local pre-push run can be.)

## Stated limitations

- Per instructions I did not execute the gate or the test suite; the sync conclusion above comes from hand-comparing the literals and hand-tracing `scanArrayBody`/the anchor regex against the real declarations, not from running the script.
- That prettier round-trips the raw NUL unchanged (so `prettier --check` stays green) is reasoned from prettier preserving raw string contents, not executed.

## Minor observations (no change required)

- `orderMismatch` also fires when the two sides differ only in *duplicate count* (e.g. server `[a,a,b]` vs ui `[a,b]`), where the CLI prose says "same elements, DIFFERENT ORDER". The degenerate case is already documented at `scripts/check-model-mirror.d.mts:22-25`, and the wording is deliberately untested — noting it only so it isn't mistaken for an oversight.
- The element regex at `scripts/check-model-mirror.mjs:169` matches only `"`/`'` strings, so a backtick-quoted element would be skipped (`scanArrayBody` handles backticks as strings, so it would surface as a delta / `missingIn` rather than a wrong pass). No such element exists in either file.
- `test/check-model-mirror.test.ts:188` pins `MIRRORED_CONSTANTS` to the exact three-element list, so adding a fourth constant means touching the test — reasonable, just worth knowing.

`scripts/next-version.mjs:67` uses the same `file://${process.argv[1]}` guard shape as finding 2, with the same silent-no-op exposure on percent-encoding paths.
`src/default-effort.ts:100` and `ui/src/lib/effort-guidance.ts:9` are byte-identical hand-copied `EFFORTS` logic with no gate — the same hazard class as the arrays, but logic rather than a literal, so this textual gate structurally cannot cover it.
`docs/external-task-api.md:84` hand-enumerates the full `CLAUDE_MODELS` list in order as prose in a Markdown table cell, ungated.
`docs-site/src/content/docs/reference/configuration.md` hand-enumerates the `EFFORTS` tiers as prose for `SHEPHERD_DOC_AGENT_EFFORT`, ungated.
`extension/src/lib/types.ts` / `extension/src/options/Options.svelte` hold a deliberate hand-maintained subset of the Claude aliases plus a `default` sentinel — a subset invariant, not a mirror, and not covered here.
The root `lint` script's `scripts/**/*.ts` glob matches neither `.mjs` nor `.d.mts`, so both new script files are unlinted — the same pre-existing gap that already leaves `check-glossary.mjs` and `next-version.mjs` unlinted.

scoop (commented):

**1. Raw NUL bytes in `scripts/check-model-mirror.mjs` — FIXED.** `git diff ebac09dff e8c950a67 -- scripts/check-model-mirror.mjs | cat -v` shows the old line as `server.join("^@")` (two raw NULs) and the new line as `server.join("\0")` — the two-character escape, rendered literally by `cat -v`, at `scripts/check-model-mirror.mjs:206`. Confirmed independently: ripgrep over the file finds `CLAUDE_MODELS` 6 times (a binary file would have its matching lines suppressed), and a ripgrep search for `\x00` in the file returns 0 matches. The comment at `scripts/check-model-mirror.mjs:201-204` now also records why NUL is the separator and why it must stay an escape.

**2. `import.meta.url === `file://${process.argv[1]}`` main guard — FIXED.** `scripts/check-model-mirror.mjs:259` is now `Boolean(process.argv[1]) && fileURLToPath(import.meta.url) === process.argv[1]`, using the `fileURLToPath` already imported at `scripts/check-model-mirror.mjs:38`. This is byte-identical to the repo precedent at `scripts/json-union-merge.mjs:156`. A regression test was added at `test/check-model-mirror.test.ts:177-204` that copies the script into a temp checkout whose path contains spaces, writes divergent halves, and asserts `status === 1` — which is the right assertion, since "the CLI block never ran" also exits 0.

## What I verified in the rest of the diff

**The gate passes on the current tree.** Compared the two halves element-by-element: `src/types.ts:315-324` vs `ui/src/lib/types.ts:1986-1995` (`CLAUDE_MODELS`, 8 elements, same order), `src/types.ts:333` vs `ui/src/lib/types.ts:2003` (`EFFORTS`, 5 elements, same order), `src/types.ts:337-349` vs `ui/src/lib/types.ts:2006-2018` (`CODEX_MODELS`, 11 elements, same order). All three in sync, so the first CI run is green.

**Each gated constant is declared exactly once per file**, so the anchor's first-match semantics cannot land on a decoy: ripgrep for `const (CLAUDE_MODELS|CODEX_MODELS|EFFORTS)\b` returns only `src/types.ts:315,333,337` and `ui/src/lib/types.ts:1986,2003,2006`.

**The strict anchor genuinely handles the real files' shapes.** `(?:export\s+)?const\s+NAME\b\s*(?::[^=]*)?=\s*\[` matches the non-exported `const CLAUDE_MODELS = [` (`src/types.ts:315`, `ui/src/lib/types.ts:1986`), the exported single-line `EFFORTS` (`src/types.ts:333`), and would match the annotated `export const PREMIUM_MODELS: readonly string[] = [` shape (`ui/src/lib/types.ts:2032`). I traced the backtracking: because `[^=]*` cannot span an `=`, the matched `=` is always the *first* `=` after the constant name, so the annotation branch can never skip forward to a later assignment. And the prose mention at `ui/src/lib/types.ts:1985` ("Mirrors src/types.ts CLAUDE_MODELS") cannot match — no `const` keyword precedes it.

**No vacuous-pass path in the comparison itself.** Unparseable on one side → `missingIn` + `ok: false` (`scripts/check-model-mirror.mjs:190-196`); zero string elements → `null`, treated as unparseable (`scripts/check-model-mirror.mjs:173`); absent on both sides → `missingIn: ["server", "ui"]`, not two matching empties. `scanArrayBody` checks comment state *before* quote state (`scripts/check-model-mirror.mjs:80-101`), so a `//` inside a string doesn't open a comment and an apostrophe inside a comment doesn't open a string; bracket depth is only tracked outside strings and comments, so `"opus[1m]"` (`src/types.ts:318`) cannot terminate the scan. The CLI's `extractArrayLiteral(...).length` at `scripts/check-model-mirror.mjs:275` runs only on the `ok === true` branch, so the non-null deref is safe.

**Wiring is complete and consistent across all four places.** `package.json:19` `"check:model-mirror": "node scripts/check-model-mirror.mjs"` (adjacent to `check:glossary`/`check:announcement-versions` at `package.json:17-18`); `scripts/pre-push.ts:400-405` in the `gates` lane between `announcement versions` (`scripts/pre-push.ts:394-399`) and `herdr types` (`scripts/pre-push.ts:406`); `.github/workflows/pr-hygiene.yml:58-59` as the last step of the `branch-hygiene` job, after `Announcement versions` (`.github/workflows/pr-hygiene.yml:56-57`) — same job that already runs `node scripts/check-glossary.mjs`, so Node is available. `CONTRIBUTING.md:79` lists the gate in exactly the pre-push lane order (`… · announcement-versions · model-mirror · herdr-types`), which I checked against the actual step order in `scripts/pre-push.ts:374-406`; `CONTRIBUTING.md:142` adds the manual command. I also checked there is no *other* canonical gate registry to update: ripgrep for `check-announcement-versions` finds only `package.json`, `scripts/pre-push.ts`, `.github/workflows/pr-hygiene.yml` and `CLAUDE.md`, and `CLAUDE.md` documents gates thematically per-section (`CLAUDE.md:25,59,73,75,93`) rather than as an enumerated list — so no `CLAUDE.md` change is owed.

**The `.d.mts` matches the runtime shape.** `MirrorDelta` (`scripts/check-model-mirror.d.mts:11-25`) has all five fields, and both `deltas.push` sites (`scripts/check-model-mirror.mjs:194` and `:209`) supply all five — no optional-field drift. `MirrorSide = "server" | "ui"` matches the literals pushed at `scripts/check-model-mirror.mjs:191-192`. The `.mjs`+`.d.mts` sibling pattern under `moduleResolution: "bundler"` (`tsconfig.json:5`) is already proven by `scripts/json-union-merge.d.mts` ← `test/json-union-merge.test.ts:2` and `scripts/next-version.d.mts` ← `test/next-version.test.ts:2`.

**The plan's lint claim holds.** `scripts/` is not in `ROOT_ESLINT_ROOTS` (`scripts/pre-push.ts:41` — `["src/", "test/", "ui/src/", "ci/onboarding-harness/", "deploy/"]`), so even though `.mjs` is in `ESLINT_EXTENSIONS` (`scripts/pre-push.ts:38`) the new `scripts/*.mjs`/`*.d.mts` are not routed to eslint by the pre-push delta path either — the surfaced gap is accurate, not a new one. `test/pre-push.test.ts` asserts nothing about gate counts or lane step lists, so the new step breaks no existing expectation.

**Scope:** the diff contains only the script, its declarations, its test, and the four wiring points. No model/effort list is touched, matching the plan's "no change to any model/effort list itself".

## Stated limitation

The one finding below rests on Node's documented default for `--preserve-symlinks-main` (off ⇒ the main module's path *is* realpath-resolved, while `process.argv[1]` is only `path.resolve`d). I verified the code shape and the repo precedent by reading them, but I did not execute Node or test on macOS — this is read-only inspection — so I am reporting it as a portability caveat rather than an observed failure.

- `scripts/check-model-mirror.mjs:259` — the `fileURLToPath(import.meta.url) === process.argv[1]` guard still compares a realpath-resolved value against a merely `path.resolve`d one, so a checkout reached through a symlinked path component silently re-opens the same exit-0-having-checked-nothing hole the fix closed for percent-encoding; `realpathSync(process.argv[1])` on the right-hand side would close it. Same shape as the existing `scripts/json-union-merge.mjs:156`, and not reachable in CI (`ubuntu-latest`, unsymlinked workspace).
- `test/check-model-mirror.test.ts:183` — the spaced-path regression guard builds its temp checkout under `os.tmpdir()`, which on macOS is `/var/folders/…` (a symlink to `/private/var/folders`); under the guard above that would make the spawned CLI exit 0 and this test red on macOS, which `README.md:158` lists as a supported-but-degraded platform.

- `scripts/next-version.mjs:67` still uses the `import.meta.url === \`file://${process.argv[1]}\`` concat form the review rejected for the new gate, so it has the same percent-encoding blind spot (harmless there — it is a value-printing helper, not a gate that can pass vacuously).
- `src/default-effort.ts` `effortsForProvider` and `ui/src/lib/effort-guidance.ts` `providerEfforts` are a hand-copied server/UI logic pair that self-declares the coupling but is not gated by anything; the plan surveys this and correctly notes a textual array extractor structurally cannot cover it.
- `docs/external-task-api.md` hand-enumerates the full `CLAUDE_MODELS` list in order inside a Markdown table cell, ungated.
- `docs-site/src/content/docs/reference/configuration.md` hand-enumerates the full `EFFORTS` list in order as prose in the `SHEPHERD_DOC_AGENT_EFFORT` row, ungated.
- `extension/src/lib/types.ts` (`CaptureConfig["model"]` union) and `extension/src/options/Options.svelte` (`models` literal) are a deliberate narrow subset of `CLAUDE_MODELS` plus a `"default"` sentinel — a subset invariant, not a mirror; only the union membership is type-checked.

scoop (commented):

**1. `isMainModule()` symlink hole — FIXED.** `scripts/check-model-mirror.mjs:274` now reads `realpathSync(entry) === fileURLToPath(import.meta.url)`, with `realpathSync` imported at `scripts/check-model-mirror.mjs:36`. Both halves are now normalized the same way: Node realpath-resolves the entry module behind `import.meta.url` (`--preserve-symlinks-main` defaults off) and `realpathSync(argv[1])` applies the same resolution to the other side. The `try/catch` at `scripts/check-model-mirror.mjs:273-279` correctly degrades an unresolvable `argv[1]` to "not the CLI entry" rather than throwing, and the `if (!entry) return false` guard at `:271` covers the missing-`argv[1]` case. The guard is now strictly stronger than the sibling idiom at `scripts/json-union-merge.mjs:156` (which still compares `fileURLToPath(import.meta.url) === process.argv[1]` unnormalized) — that file is not in this diff, so no finding.

**2. Spaced-path test built under an unresolved `os.tmpdir()` — FIXED.** `test/check-model-mirror.test.ts:643` now wraps the temp root: `realpathSync(mkdtempSync(join(tmpdir(), "shepherd model mirror ")))`, and the symlink case at `test/check-model-mirror.test.ts:656` does the same for its root before deliberately re-introducing one symlink component. `realpathSync` is imported at `test/check-model-mirror.test.ts:432`. So on macOS (`/var/folders/…` → `/private/var/folders/…`) the spaced-path case now isolates only the percent-encoding trap, and the symlink case isolates only the realpath trap — the comment at `test/check-model-mirror.test.ts:613-617` states exactly that. Both cases assert exit status 1 (`test/check-model-mirror.test.ts:636`), which a skipped CLI block cannot produce, so the guard itself is genuinely under test in both directions.

## Verification of the change itself

**The gated lists are actually in sync today**, so the gate is green on first run: `src/types.ts:315-324` (`CLAUDE_MODELS`: fable, opus, opus[1m], claude-opus-5, claude-opus-5[1m], sonnet, sonnet[1m], haiku) is element- and order-identical to `ui/src/lib/types.ts:1986-1995`; `src/types.ts:333` `EFFORTS` matches `ui/src/lib/types.ts:2003`; `src/types.ts:337-349` `CODEX_MODELS` (11 entries, gpt-5.5 … o3) matches `ui/src/lib/types.ts:2006-2018` in the same order.

**The parser's stated traps are real, and it handles them.** The strict `const` anchor (`scripts/check-model-mirror.mjs:161`) is justified by live prose: `ui/src/lib/types.ts:1984-1985` names `CLAUDE_MODELS` in the doc comment directly above the declaration, and `src/types.ts:352` names `CODEX_MODELS` in the `CODEX_MODEL_RE` comment just below that array — neither is preceded by `const`, so `const\s+<NAME>\b` cannot match them. `\b` correctly rejects a longer prefixed name (`_` is a word char, so `CLAUDE_MODELS_LEGACY` yields no boundary) — asserted at `test/check-model-mirror.test.ts:521-524`. The `] as const`-avoidance rationale checks out too: `ui/src/lib/types.ts:2032-2039` is exactly the cited `export const PREMIUM_MODELS: readonly string[] = [ … ];` shape with a bare `];`, and in `src/types.ts` the next `] as const` after `CLAUDE_MODELS` is `EFFORTS`' at `:333`, so an `as const`-anchored scan would indeed have absorbed the five effort tiers. `scanArrayBody` (`scripts/check-model-mirror.mjs:71-132`) skips bracket-depth accounting entirely while inside a quote (`:91-102`), which is what keeps `"opus[1m]"`/`"claude-opus-5[1m]"` from terminating the scan; I traced the escape branch (`:93-97`, `"a\\\\"` closes on the right quote) and both comment branches (`/**/`, `/*/`) and they advance correctly. Fail-closed on either side is implemented as described at `scripts/check-model-mirror.mjs:186-192`, and the empty-element-list → `null` rule at `:173` blocks the vacuous two-empty-lists comparison.

**Declarations match the implementation.** `scripts/check-model-mirror.d.mts` declares exactly the three exports the module has — `MIRRORED_CONSTANTS` (`scripts/check-model-mirror.mjs:58`), `extractArrayLiteral` (`:158`), `compareMirror` (`:181`, with the optional third `constants` param mirrored as `constants?: string[]`) — and no others. `test/check-model-mirror.test.ts:439-443` imports exactly those three. The `.mjs` → `.d.mts` resolution has precedent under this tsconfig (`tsconfig.json:5`, `moduleResolution: "bundler"`): `test/next-version.test.ts:2` imports `../scripts/next-version.mjs` against `scripts/next-version.d.mts`. `noUncheckedIndexedAccess` (`tsconfig.json:9`) is satisfied by the `deltas[0]!` assertions throughout the test.

**Wiring is complete and consistent** — all three requested entry points plus docs: `package.json:19` (`"check:model-mirror": "node scripts/check-model-mirror.mjs"`, alongside `check:glossary`/`check:announcement-versions` at `:17-18`); `scripts/pre-push.ts:400-405` in the `gates` lane, same `{cmd: "node", args: [...], cwd: repoRoot}` shape as the `announcement versions` step at `:394-399`; `.github/workflows/pr-hygiene.yml:58-59` in the `branch-hygiene` job right after the two sibling `node scripts/…` steps at `:54-57` (no `setup-node` needed — the existing steps already rely on the runner's preinstalled node). `CONTRIBUTING.md:79` adds `model-mirror` to the gates lane list (the only such list in the repo — grepped `*.md`), and `CONTRIBUTING.md:142` adds the manual command. The success path's non-null deref at `scripts/check-model-mirror.mjs:395` (`extractArrayLiteral(...).length`) is safe because it only runs when `ok` is true, which implies no `missingIn` for any constant.

I did not run tests or lint (CI's job), and I did not execute the gate — read-only inspection only.

- `scripts/check-model-mirror.mjs:169`: `scanArrayBody` deliberately tracks backticks as a quote character (`:112`) but the element regex matches only `"…"` and `'…'`, so a template-literal element (`` `opus` ``) is kept in the body and then silently dropped — if a future edit writes the same entry with backticks on *both* sides, that entry becomes invisible to the gate (a partial vacuous pass) rather than a loud failure; prettier does not rewrite template literals to quoted strings, so nothing else would catch it.
- `scripts/check-model-mirror.mjs:161-162`: comments are stripped only from the array *body*, not from the source before the anchor regex runs, so a commented-out `// const EFFORTS = ["a"] as const;` sitting above the real declaration would win the `exec` and be compared instead of the live literal — the doc comment at `:138-145` claims prose "can never match", which is true of prose but not of commented-out code.
- `scripts/check-model-mirror.mjs:198-199`: two lists with the same membership but different *duplicate counts* (e.g. server `["a","a","b"]` vs ui `["a","b"]`) set `orderMismatch`, so the CLI reports "same elements, DIFFERENT ORDER" for what is really a duplicate-count divergence — the gate still fails (correct outcome), only the message misdescribes it; `scripts/check-model-mirror.d.mts:69-72` already documents this degenerate case.

- `scripts/json-union-merge.mjs:156` and `scripts/next-version.mjs` still use unnormalized main-module guards (`fileURLToPath(import.meta.url) === process.argv[1]`), which the new script's header comment at `:267-268` explicitly notes and leaves alone.
- The root `"lint"` glob (`package.json:10`) is `'scripts/**/*.ts'`, so no `scripts/*.mjs` or `*.d.mts` file is ever linted — the new gate inherits that gap rather than creating it.
- `pr-hygiene.yml`'s `branch-hygiene` job carries an `if:` guard skipping the whole job for release-please PRs (`:30`), so the new gate is skipped there too — identical to the glossary and announcement-version gates.
- The duplicated `EFFORTS` *logic* pair (`src/default-effort.ts` `effortsForProvider` ↔ `ui/src/lib/effort-guidance.ts` `providerEfforts`) remains ungated; a textual array extractor structurally cannot cover it.
- `docs/external-task-api.md` and `docs-site/src/content/docs/reference/configuration.md` hand-enumerate the model/effort lists as Markdown prose and stay ungated.

PR #1942, opened 2026-07-25
fix(ui): one tap from the mobile repo chip into the repo typeahead

scoop:
On the mobile new-task pane, changing the repo cost **two taps**: one on the combined `repo · branch` chip to open the context sheet, another on the `RepoSelect` trigger to reveal the typeahead. Repo changes are far more common than branch changes, so the chip now goes straight through to the filter.

## What changed

`NewTask.svelte`'s mobile header chip points at `openRepoPicker()` — the existing ⌥R handler — instead of just setting `activeSheet = "context"`. That function is now **synchronous**, using `flushSync` in place of `await tick()`:

```ts
function openRepoPicker() {
  if (mobile) {
    activeSheet = "context";
    flushSync(); // mount the sheet + its RepoSelect
  }
  repoSelect?.openPanel();
  flushSync(); // render the panel; RepoSelect's focus effect runs → filter focused
}
```

The timing is the load-bearing part, not an incidental refactor: **iOS Safari raises the software keyboard only for a `focus()` that happens inside the gesture's own call stack.** `await tick()` is fine for a key event but forfeits that for a tap. `flushSync` drains the pending effects in-stack, so `RepoSelect`'s `open && filterInput` effect focuses the filter before the handler returns. The sheet's `use:dialog` focus microtask still runs afterwards but no-ops — its guard sees focus already inside the sheet.

No `RepoSelect.svelte` change; no new messages, CSS, or catalog entry.

## Reviewer notes

**The iOS keyboard itself is not assertable here** — Chromium has no on-screen-keyboard surface, and the existing fake `visualViewport` only simulates the resulting geometry. So the new test asserts the sheet / panel / `activeElement === .rs-filter` triple **synchronously**, with no `await` or `poll`, as a structural fence: I verified it fails (`expected null to be truthy`) when the `await tick()` version is reinstated. A reintroduced async hop breaks CI rather than silently costing the keyboard. I have not run this on a physical iOS device — worth a quick check before merge.

**Branch editing stays in the same sheet**, below the repo list. Tapping the branch control while the list is open collapses the list, because `RepoSelect` registers a capture-phase window click listener and `.ctx-branch-select` sits outside `.rs-root`. This interaction is new (previously the panel was only open when deliberately opened). It's accepted rather than fixed — the tap still lands on the branch control (hit testing precedes the reflow) and the sheet stays open, so the edit succeeds; changing `RepoSelect`'s outside-click contract would affect the backlog / steers / card consumers for a one-frame reflow. It is now asserted, so changing it later is a deliberate act.

**One trap worth flagging in the test changes:** the suite-wide `mockListBranches` returns `branches: ["main"]`, which makes `.ctx-branch-select` a *single-option* select — writing its value back and dispatching `change` asserts nothing. The extended 20-repo test overrides that mock with `["main", "release"]` and pins `options.length === 2` / `value === "main"` before asserting the flip, so it can't quietly go vacuous again. (The pre-existing vacuous write in `"combined chip names both values…"` is left alone — out of scope.)

**Accepted and deliberately untested:** the chip is now the primary mobile entry point, so it will routinely be tapped while `listRepos()` is still in flight — an empty list under a focused filter that fills in on resolve, with the roving cursor left on row 0. Both new tests fence past that race (polling the chip label for the repo name before tapping) rather than encoding it.

## Verification

- `cd ui && bun run check` — 5630 files, 0 errors
- `cd ui && bun test` — 279 files, 4283 tests passing
- root `bunx prettier --check .` + `bun run lint` — clean (root lint globs include `ui/src`, so both edited files are covered)
- `check:i18n`, glossary and branch-hygiene gates pass; pre-push green

### DIFF
--- a/ui/src/lib/components/NewTask.browser.test.ts
+++ b/ui/src/lib/components/NewTask.browser.test.ts
@@ -2890,8 +2890,8 @@ describe("NewTask mobile sheets + shortcuts", () => {
     await expect.poll(() => document.querySelector(".ctx-sheet")).toBeTruthy();
     expect(chip().getAttribute("aria-expanded")).toBe("true");
 
-    // Change only the repo: chip + payload repoPath update; base re-derives.
-    document.querySelector<HTMLButtonElement>(".ctx-sheet .rs-trigger")!.click();
+    // Change only the repo: chip + payload repoPath update; base re-derives. The chip
+    // tap already opened the repo panel — clicking the trigger here would close it.
     await expect
       .poll(() =>
         Array.from(document.querySelectorAll<HTMLLIElement>('[role="option"]')).find((el) =>
@@ -2934,6 +2934,39 @@ describe("NewTask mobile sheets + shortcuts", () => {
     expect(onsubmit.mock.calls[0]![0]).toMatchObject({ repoPath: repoB.path, baseBranch: "main" });
   });
 
+  it("one tap on the chip opens the sheet, the repo panel and focuses the filter", async () => {
+    await renderMobile();
+    const chip = () => document.querySelector<HTMLButtonElement>(".ctx-chip")!;
+    // `renderMobile` resolves as soon as the chip exists — which is BEFORE `listRepos`
+    // populates the list. Fence on the repo name so the tap below races nothing.
+    await expect.poll(() => chip().textContent).toContain("alpha");
+
+    chip().click();
+
+    // Asserted synchronously — no await, no poll. `openRepoPicker` flushes both renders
+    // inside the click handler precisely so the focus stays in the gesture's own call
+    // stack (iOS Safari raises the software keyboard only for an in-stack focus()).
+    // If someone reintroduces an async hop there, these three fail rather than silently
+    // costing the keyboard on a device no CI lane can observe.
+    expect(document.querySelector(".ctx-sheet")).toBeTruthy();
+    expect(document.querySelector(".ctx-sheet .rs-panel")).toBeTruthy();
+    expect(document.activeElement).toBe(document.querySelector(".ctx-sheet .rs-filter"));
+
+    // The typeahead is live: typing narrows the options…
+    const filter = document.querySelector<HTMLInputElement>(".ctx-sheet .rs-filter")!;
+    filter.value = "brav";
+    filter.dispatchEvent(new Event("input", { bubbles: true }));
+    await expect.poll(() => document.querySelectorAll('.ctx-sheet [role="option"]').length).toBe(1);
+
+    // …and picking collapses the panel while the sheet stays open, so the branch control
+    // is right back in view.
+    document.querySelector<HTMLLIElement>('.ctx-sheet [role="option"]')!.click();
+    await expect.poll(() => chip().textContent).toContain("bravo");
+    await expect.poll(() => document.querySelector(".rs-panel")).toBeNull();
+    expect(document.querySelector(".ctx-sheet")).toBeTruthy();
+    expect(document.querySelector(".ctx-sheet .ctx-branch-select")).toBeTruthy();
+  });
+
   it("three-press Escape: repo panel → focus in-sheet trigger; sheet → focus chip", async () => {
     await renderMobile();
     const chip = document.querySelector<HTMLButtonElement>(".ctx-chip")!;
@@ -2942,8 +2975,7 @@ describe("NewTask mobile sheets + shortcuts", () => {
     chip.click();
     await expect.poll(() => document.querySelector(".ctx-sheet")).toBeTruthy();
 
-    const trigger = () => document.querySelector<HTMLButtonElement>(".ctx-sheet .rs-trigger")!;
-    trigger().click();
+    // The chip tap opens the repo panel too, so the ladder starts one level deep already.
     await expect.poll(() => document.querySelector(".rs-panel")).toBeTruthy();
     const filter = document.querySelector<HTMLInputElement>(".rs-filter")!;
     filter.focus();
@@ -3107,14 +3139,24 @@ describe("NewTask keyboard-aware viewport (mobile)", () => {
       // Smallest supported portrait: a long repo list forces the panel to overflow.
       await page.viewport(375, 667);
       mockListRepos.mockResolvedValue({ repos: makeRepos(20), recentWindowDays: 30 });
+      // ≥2 branches, overriding the suite default of `["main"]`: with a single option
+      // `.ctx-branch-select` cannot change value, so the branch assertion below would
+      // pass against a completely dead control.
+      mockListBranches.mockResolvedValue({
+        current: "main",
+        branches: ["main", "release"],
+        default: null,
+      });
       render(NewTask, { props: { onsubmit: vi.fn(), initialRepoPath: "/repo/kbd-00" } });
-      await expect.poll(() => document.querySelector(".ctx-chip")).toBeTruthy();
+      // Wait for the repo list to land (the chip renders before `listRepos` resolves).
+      await expect
+        .poll(() => document.querySelector(".ctx-chip")?.textContent)
+        .toContain("repo-00");
       await expect.poll(() => overlay()?.style.height).toBe("380px");
 
-      // Open the context sheet, then the repo dropdown + its filter.
+      // One tap opens the context sheet AND the repo dropdown with its filter.
       document.querySelector<HTMLButtonElement>(".ctx-chip")!.click();
       await expect.poll(() => document.querySelector(".ctx-sheet")).toBeTruthy();
-      document.querySelector<HTMLButtonElement>(".ctx-sheet .rs-trigger")!.click();
       await expect.poll(() => document.querySelector(".ctx-sheet .rs-filter")).toBeTruthy();
 
       const filter = document.querySelector<HTMLInputElement>(".ctx-sheet .rs-filter")!;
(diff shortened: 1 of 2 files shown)


scoop (commented):
## What I checked

The change is small and does what the task asks: `.ctx-chip`'s `onclick` now points at `openRepoPicker`, which was made synchronous with two `flushSync()` calls so the sheet mounts, the panel opens, and `RepoSelect`'s focus effect lands on `.rs-filter` inside the tap's own call stack.

### Identifiers / wiring resolved

- `flushSync` is imported from `svelte` alongside the existing `onMount, tick` (`ui/src/lib/components/NewTask.svelte:2`); `tick` is still used by `closeSheetFocusPrompt` so the import is not dead. This is the repo's first `flushSync` use — grepped `ui/src`, only the new call sites (`NewTask.svelte:1017,1020`).
- `openRepoPicker` has exactly two call sites: the ⌥R case (`NewTask.svelte:1050`) and the chip (`NewTask.svelte:1291`). Nothing else awaited its former `Promise`, so dropping `async`/`void` breaks no caller (grepped `ui/src`; the `openRepoPicker` hits in `page/AppOverlays.browser.test.ts:253` are an unrelated local test helper).
- `repoSelect` is `bind:this` on both the desktop RepoSelect (`NewTask.svelte:1329`, inside `{#if !mobile}`) and the in-sheet one (`NewTask.svelte:1697`, unconditional inside `.ctx-sheet`). They are mutually exclusive, and the in-sheet instance is not behind a further `{#if}`, so after the first `flushSync()` on the mobile path `repoSelect` refers to the sheet's instance; the `?.` still covers the mobile-with-no-sheet case.
- No new message keys. The chip's `aria-label` key is unchanged and present in both catalogs — `ui/messages/en.json:3326` and `ui/messages/de.json:3326` — so locale parity is intact.
- `.ctx-branch-select` exists in both the `<select>` and the `<input>` fallback (`NewTask.svelte:1716,1726`), styled at `NewTask.svelte:2705`.

### Correctness claims I verified against ground truth

1. **The auto-opened panel is not immediately closed by the chip's own click.** `RepoSelect`'s root `$effect` registers `window.addEventListener("click", onClick, true)` (`ui/src/lib/components/RepoSelect.svelte:196`), which sets `open = false` for any target outside `.rs-root`. That listener is registered *during* the chip's click dispatch (inside `flushSync`). Because it is capture-phase on `window`, and window's capture phase already ran before the handler, it cannot fire for the in-flight event; Svelte 5's delegated `click` handler runs in the bubble phase on the mount target/`document`, so the ordering holds. On mobile there is no earlier-mounted `RepoSelect` (the only mobile instance lives inside the sheet, `NewTask.svelte:1697`), so no pre-existing listener exists at tap time either.
2. **The sheet's focus-pull no-ops, as the new comment claims.** `use:dialog` is on `.sheet` (`ui/src/lib/components/new-task/MobileEngineSheet.svelte:33`) and its microtask guard is `if (!document.contains(node) || node.contains(document.activeElement)) return;` (`ui/src/lib/a11yDialog.ts:49`). After the second `flushSync()` the active element is `.rs-filter`, which is inside `.sheet-body` → inside `.sheet`, so the guard returns early. Comment at `NewTask.svelte:1011-1012` is accurate.
3. **No stray character from ⌥R now that focus moves earlier.** `case "KeyR"` falls through to `e.preventDefault(); e.stopPropagation();` in the same handler (`NewTask.svelte:1049-1056`), and `preventDefault()` anywhere in the dispatch suppresses the default text insertion, so focusing the filter before it cannot leak an `r`/`®` into the input.
4. **The panel renders in-flow inside the sheet**, so auto-opening grows the sheet body instead of dropping a dropdown into the keyboard: `.ctx-sheet :global(.rs-panel) { position: static }` and `.ctx-sheet :global(.rs-list) { max-height: none }` with a sticky `.rs-filter` (`NewTask.svelte:2685-2695`).
5. **Picking a repo collapses the panel and keeps focus in the sheet.** `pick()` sets `open = false` (`RepoSelect.svelte:130-135`) and `selectRepo` re-focuses the in-sheet `.rs-trigger` while `activeSheet === "context"` (`NewTask.svelte:975-986`) — matching the new test's post-pick assertions.
6. **Escape ladder unchanged.** `RepoSelect`'s node keydown closes the panel and calls `onescape` (`RepoSelect.svelte:181-190`), which the sheet wires to the in-sheet trigger (`NewTask.svelte:1706-1710`); the sheet's own dialog action then handles Esc 2 and restores focus on destroy (`a11yDialog.ts:54-61,90-97`). The edited Escape test drops only the now-redundant `trigger` local and does not reference it afterwards, so it still type-checks.
7. **The keyboard test's branch mock is sound and isolated.** `mockListBranches.mockReset()` then the suite default `["main"]` run in `beforeEach` (`ui/src/lib/components/NewTask.browser.test.ts:117,131`), so the in-test override at line 3145 applies to this test only. With `{current:"main", branches:["main","release"], default:null}`, `pickBaseBranch` returns `"main"` (`ui/src/lib/api.ts:765-767`) and `baseOptions` is `["main","release"]` (`NewTask.svelte:275`) — so `options.length === 2` and `value === "main"` are both real, and the flip to `"release"` is a genuine transition observed through the chip's `· {baseBranch}` (`NewTask.svelte:1295`).
8. **Scope.** Diff is limited to `NewTask.svelte` + `NewTask.browser.test.ts`; no `RepoSelect.svelte` edit, no new CSS, no new i18n keys, no catalog fragment — consistent with the plan's boundaries. `.ctx-chip`/`.ctx-sheet` appear in no other test or component (grepped repo-wide), so no other consumer needed updating.

## Stated limitations (could not verify here, not findings)

- I could not confirm Svelte 5.56.7's internal flush ordering from source — `node_modules` is not installed in this worktree and no vendored Svelte is present. Specifically unverified: that `bind:this={repoSelect}` is assigned within the first `flushSync()`, and that the newly-dirtied `open && filterInput` user effect (`RepoSelect.svelte:137-141`) drains within the second. This is exactly what the new synchronous triple asserts, so CI will confirm or fail it rather than leaving it to a device.
- iOS Safari soft-keyboard behaviour is not observable in this repo or in a Chromium lane. The premise (WebKit raises the keyboard only for a `focus()` inside the gesture's synchronous stack, and `focus({preventScroll:true})` from a `flushSync`-driven effect still counts as in-stack) is sound as written but assumed; the plan's manual on-device check still stands as the only real proof.

## Observations (non-blocking, no action required)

- The in-stack-focus guarantee is owned by `RepoSelect`'s internal `$effect`, which `NewTask` cannot see. A future change there (e.g. deferring the focus behind a `tick()`) would silently cost the iOS keyboard, with only this one synchronous assertion as the tripwire. The plan deliberately ruled out touching `RepoSelect`, and the tripwire is the right mitigation given that — just worth knowing the coupling exists.
- The load race is real and now on the primary path: `selectedRepoName` derives from the fetched list (`NewTask.svelte:262-263`), so a tap before `listRepos()` resolves opens the panel over an empty list with `.rs-empty` announcing "no matches", and leaves the roving cursor on row 0 rather than the selected repo. Typing works the moment the list lands, and the plan names this as accepted and deliberately unasserted.

- `RepoSelect.svelte`'s capture-phase outside-click contract closes the panel for a tap on anything outside `.rs-root`, which inside the context sheet includes the field labels and the branch control — now hit far more often since the panel starts open.
- `RepoSelect.svelte`'s `.rs-empty` live region announces "no matches" for an empty-because-still-loading list, with no loading state to distinguish it.
- The pre-existing branch-edit step in `"combined chip names both values, opens the context sheet, edits independently"` writes `"main"` back into a single-option `<select>` and asserts nothing; the plan explicitly leaves it alone, and the new assertions correctly do not repeat the pattern.

scoop (commented):
## Verdict

The change does what the task asks and I found no correctness, security or quality defect. The prior revision's single finding is genuinely fixed. No blocking items.

## Prior finding — re-checked

**1. `branchCtl.scrollIntoView({ block: "nearest" })` was a no-op because the body was already pinned at max scroll — ADDRESSED.** The new block at `ui/src/lib/components/NewTask.browser.test.ts:3203-3208` inserts `sheetBody.scrollTop = 0;` and then asserts `branchCtl.getBoundingClientRect().top` is `> KB_TOP` *before* the `scrollIntoView`, so the scroller is genuinely rewound and the following `bottom <= KB_TOP + 1` / `top >= 0` pair now measures the effect of a real scroll rather than the leftover `scrollTop = scrollHeight` from `:3172`. The comment was reworded to match what is proved ("Rewind the scroller to the top first, so the scroll below is a real one"). The rewind is legitimate: `scrollTop = 0` is the state the sheet actually opens in, and with the in-flow panel (`.ctx-sheet :global(.rs-panel) { position: static }`, `.rs-list { max-height: none }` — `ui/src/lib/components/NewTask.svelte:2685-2691`) 20 rows sit above `.ctx-branch-select`, so the pre-scroll assertion is not vacuous either.

## What I verified

**Source change (`ui/src/lib/components/NewTask.svelte`)**

- `flushSync` is imported at `:2` and used only at `:1017` / `:1020`; `tick` is retained in the same import and still used at `:335` and `:786` (`closeSheetFocusPrompt`), so the import is not dead.
- The chip handler is `onclick={openRepoPicker}` (`:1291`), replacing the direct `activeSheet = "context"` assignment. `openRepoPicker` (`:1014-1021`) is now synchronous, and the only other caller — ⌥R at `:1050` — dropped its `void`. `e.preventDefault()` / `e.stopPropagation()` still run after the switch at `:1055-1056`, so moving focus into the filter input *during* the keydown cannot leak an Alt+R character into it.
- `repoSelect?.openPanel()` resolves to `RepoSelect.svelte:120-126` (`export function openPanel()` → `open = true`, resets `filter`, points `activeIdx` at the selected repo). The focus that makes the one-tap contract work is the `$effect` at `RepoSelect.svelte:143-147` (`if (open && filterInput) filterInput.focus({ preventScroll: true })`), whose `filterInput` is bound at `:238` inside `{#if open}` (`:235`). Both flushes are needed and both are in-stack, as the comment at `:1004-1013` claims.
- The doc comment's claim that the sheet's focus microtask no-ops is accurate: `ui/src/lib/a11yDialog.ts:48-52` returns early when `node.contains(document.activeElement)`, and after the second `flushSync` the active element is `.rs-filter`, inside the sheet.
- **Auto-open survives the tap that caused it.** `RepoSelect` registers a *capture-phase* `window` click listener that closes the panel for targets outside `.rs-root` (`RepoSelect.svelte:201-213`). That listener is added during the chip's own click handler, i.e. after the window capture phase for that event has already been invoked, so it is not in the cloned listener list for the in-flight click and does not immediately re-close the panel. A later tap on `.ctx-branch-select` (a sibling of `.rs-root`, `NewTask.svelte:1713-1731`) *does* hit it — that is the collapse the plan documents and the test now pins at `:3215`.
- `bind:this={repoSelect}` exists on both the desktop (`:1329`) and in-sheet (`:1697`) instances; only one is mounted per breakpoint, and the first `flushSync` is what makes the in-sheet assignment visible before `repoSelect?.openPanel()` reads it.
- `.ctx-chip` is referenced only by `ui/src/lib/components/NewTask.svelte` and `ui/src/lib/components/NewTask.browser.test.ts` (repo-wide grep), so no other consumer depended on the old "chip only opens the sheet" behaviour. `openPanel` has exactly one external caller, `NewTask.svelte:1019`; the `openPanelTick` hits in `PlanGateBadge`/`UnitRow` are an unrelated prop.
- No message keys, catalog fragments, CSS or `RepoSelect.svelte` edits — the diff is the two files the plan names, so there is no i18n-parity surface to check.

**Test changes (`ui/src/lib/components/NewTask.browser.test.ts`)**

- The three now-redundant `.ctx-sheet .rs-trigger` clicks are removed (`:2893-2894`, `:2978`, `:3159`); grep for `rs-trigger` shows the only remaining uses are the desktop tests at `:1045` / `:1170` / `:2068` and the Escape assertion at `:2990`, none of which toggle an already-open in-sheet panel.
- The multi-branch override at `:3145-3149` lands before `render` at `:3151`, and the suite-wide `beforeEach` resets it every test (`mockListBranches.mockReset()` at `:117`, re-defaulted at `:131`), so it cannot leak. With `{ current: "main", default: null }`, `pickBaseBranch` returns `"main"` (`ui/src/lib/api.ts:766`) and `baseOptions` is `["main", "release"]` (`NewTask.svelte:275`), so `options.length === 2` and `value === "main"` at `:3198-3199` hold, and the `"release"` flip is a real transition rather than the dead single-option write the plan called out.
- The `length).toBe(1)` filter assertion at `:2959` is sound: `role="option"` is only on `.rs-row` (`RepoSelect.svelte:263-276`), and a non-empty filter empties the pinned recents (`searching ? [] : recentRepos(...)`, `RepoSelect.svelte:88, 108-110`), so no duplicate pinned row can inflate the count.
- The synchronous triple at `:2951-2953` is a real structural guard: reintroducing an `await` in `openRepoPicker` makes the first of the three fail, which is what the plan wanted it to do.

## Stated limitations

- `node_modules` is not installed in this worktree, so I could not open Svelte's own typings to confirm `flushSync` is exported from `"svelte"` at the pinned `^5.56.7` (`ui/package.json:36`). It is the documented Svelte 5 client export and `bun run check` covers it; I am not treating it as a finding either way.
- Whether iOS Safari actually raises the software keyboard cannot be verified from this repo or in the Chromium test lane — the plan flags a manual device check, and the in-stack focus is the right structural precondition for it.
- I did not run tests or lint (CI's job, and read-only review here).

PR #1943, opened 2026-07-25
feat(backlog): active filter chip picks the opened tab

scoop:
In the repo pane, the active filter chip now picks the tab a repo click opens. Filtering by **has PRs** is an explicit statement of what you're hunting for, so the click goes straight to the PRs tab instead of always landing on Issues.

| `has issues` | `has PRs` | click opens |
| ------------ | --------- | -------------------------- |
| off          | on        | **PRs**                    |
| on           | off       | **Issues**                 |
| on           | on        | **Issues**                 |
| off          | off       | unchanged (as before)      |

Issues wins when both chips are on: `has PRs` alone is unambiguous PR-hunting intent, both-on is not.

## How

- New pure `tabForFilters()` in `ui/src/lib/components/backlog-view.ts` returns `"issues" | "prs" | null`, where `null` means *leave the current tab alone* (not "fall back to Issues"). Narrow union on purpose — the full `Tab` type is declared per-component and hoisting it is a separate concern.
- `BacklogView.svelte` already owned both the chip state and `activeTab`, so no new props and no server change. The two inline `onselect={(p) => (selectedPath = p)}` props (mobile + desktop `ProjectBacklogList`) collapse into one shared `handleSelect()`, so the two branches can't drift.
- The desktop pinned-repo auto-seed applies the same rule. Its `$effect` tracks `payload?.pinnedPath`, so it re-fires on every payload update: without this, turning `has PRs` on with a no-PR repo selected drops that selection and the next poll re-seeded the pinned repo on a **stale Issues tab** — exactly the state this change exists to prevent. Chips are read via `untrack`, so the effect's dependency set is unchanged and a chip toggle alone still never seeds; the existing `selectedPath === null` guard keeps a poll from re-tabbing a selection you're already reading.

## Deliberately not changed

- **Toggling a chip while a matching repo is already open does not move the tab** — the redirect fires on selection, not on the toggle. A chip toggle is a list-scoping action; yanking the detail pane out from under someone mid-read would be worse than the inconsistency. Reviewers: this inert case is intended, not a missed wiring site.
- The EPIC-target and just-added-repo effects keep forcing Issues (an epic target *is* an issue; a brand-new repo has zero PRs, and that effect already clears both chips first).

## Testing

- `backlog-view.test.ts` — one unit test per row of the table above, including `null` meaning "leave alone".
- `BacklogOverlay.browser.test.ts` (already mounts the real `BacklogView`, no api mocking needed) — desktop click, mobile click, both-chips precedence, the no-chip leave-alone case, plus a `seedPayload()` fixture covering the poll-driven re-seed landing on PRs and its mirror: a live selection parked on Actions survives a poll un-retabbed.
- All 6 new browser assertions were checked against deliberately un-wired code first: the 4 positive cases fail, the 2 "no movement" cases correctly still pass. They bind.
- `bun run check` (0 errors), `bun run check:i18n` (2 locales, 3411 keys), `bunx vitest run` → **279 files / 4294 tests pass**; root `bun run lint` and all four pre-push hygiene gates clean.

Feature-announcement fragment `v1.46.0-pr-filter-tab.ts` with EN + DE keys included.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3408,5 +3408,7 @@
   "feat_steers_settings_section_title": "Steers bekommen einen eigenen Einstellungsbereich",
   "feat_steers_settings_section_body": "Gespeicherte Steers liegen jetzt unter Einstellungen → Steers. Die Sitzungssteuerung bleibt fokussiert, während Stift und Steer-Kontextmenü weiterhin denselben Editor öffnen und den ausgewählten Steer beibehalten.",
   "feat_opus_5_pin_title": "Opus 5 exakt fixieren",
-  "feat_opus_5_pin_body": "Die Modellauswahl bietet jetzt zusätzlich Opus 5 und Opus 5 (1 Mio. Kontext). Die einfachen Opus-Einträge folgen immer dem neuesten Opus-Release; ein fixierter Eintrag legt genau diese Version fest, damit ein künftiges Release nicht still ändert, womit Task, Repo-Standard oder Rolle laufen."
+  "feat_opus_5_pin_body": "Die Modellauswahl bietet jetzt zusätzlich Opus 5 und Opus 5 (1 Mio. Kontext). Die einfachen Opus-Einträge folgen immer dem neuesten Opus-Release; ein fixierter Eintrag legt genau diese Version fest, damit ein künftiges Release nicht still ändert, womit Task, Repo-Standard oder Rolle laufen.",
+  "feat_pr_filter_tab_title": "Nach PRs filtern, im PR-Tab landen",
+  "feat_pr_filter_tab_body": "Unter Repos bestimmt jetzt der aktive Filter-Chip, welcher Tab beim Öffnen eines Repos erscheint: Ist „Hat PRs“ aktiv, führt ein Klick direkt zu den PRs statt zu den Issues. „Hat Issues“ macht dasselbe für Issues, und ohne aktiven Chip bleibt der aktuelle Tab unverändert."
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -3408,5 +3408,7 @@
   "feat_steers_settings_section_title": "Steers get their own settings section",
   "feat_steers_settings_section_body": "Saved Steers now live under Settings → Steers. Session controls stay focused, while the pencil and steer context menu still open the same editor and preserve the selected steer.",
   "feat_opus_5_pin_title": "Pin Opus 5 exactly",
-  "feat_opus_5_pin_body": "Model pickers now offer Opus 5 and Opus 5 (1M context) alongside the existing Opus rows. The plain Opus entries always track the newest Opus release; picking a pinned one locks that exact version, so a future release can't quietly change what a task, repo default, or role runs on."
+  "feat_opus_5_pin_body": "Model pickers now offer Opus 5 and Opus 5 (1M context) alongside the existing Opus rows. The plain Opus entries always track the newest Opus release; picking a pinned one locks that exact version, so a future release can't quietly change what a task, repo default, or role runs on.",
+  "feat_pr_filter_tab_title": "Filter by PRs, land on the PRs tab",
+  "feat_pr_filter_tab_body": "In Repos, the active filter chip now picks the tab a repo opens on: with \"Has PRs\" on, clicking a repo goes straight to its PRs instead of Issues. \"Has issues\" does the same for Issues, and with no chip on your current tab is left alone."
 }
--- a/ui/src/lib/components/BacklogOverlay.browser.test.ts
+++ b/ui/src/lib/components/BacklogOverlay.browser.test.ts
@@ -4,6 +4,7 @@ import { render } from "vitest-browser-svelte";
 import { page } from "vitest/browser";
 import "../../app.css";
 import BacklogOverlay from "./BacklogOverlay.svelte";
+import { m } from "$lib/paraglide/messages";
 import { backlogLayout } from "$lib/backlog-layout.svelte";
 import type { BacklogPayload, BacklogProject } from "$lib/types";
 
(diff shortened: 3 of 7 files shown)


scoop (commented):
## What I verified

**The rule (`ui/src/lib/components/backlog-view.ts:205-212`).** `tabForFilters` returns `"issues"` when `hasIssues`, else `"prs"` when `hasPRs`, else `null`. The `"issues" | "prs"` return is assignment-compatible with `activeTab`, whose union is declared at `ui/src/lib/components/BacklogView.svelte:133` (`"issues" | "prs" | "actions" | "readiness" | "automation"`) and mirrored at `ui/src/lib/components/backlog-view/BacklogTabBar.svelte:8` — so `activeTab = tab` typechecks at both new write sites. `if (tab) activeTab = tab` is safe truthiness here (both non-null values are non-empty strings).

**Both selection call sites are actually wired.** `onselect={handleSelect}` appears at `ui/src/lib/components/BacklogView.svelte:394` (mobile branch) and `:472` (desktop branch); grep across `ui/src/lib/components` finds no remaining inline `onselect={(p) => ...}` in this component. `ProjectBacklogList`'s prop contract is `onselect: (path: string) => void` (`ui/src/lib/components/ProjectBacklogList.svelte:47`), which `handleSelect(path: string)` satisfies, and every row render site funnels through it — recents (`:153`), rest (`:161-168`) and the hidden group (`:187`). So keyboard activation (`ProjectRow.svelte:26-35` → `onselect()`) gets the redirect too.

**No repo can land on a tab that doesn't match its chip.** `filtered` is computed by `filterProjects` *before* `splitHidden` (`BacklogView.svelte:214-221`), and recents are partitioned from the already-filtered `projects` (`ProjectBacklogList.svelte:63`). Every clickable row therefore satisfies the active predicate, so "has PRs" on can never open a 0-PR repo on the PRs tab from a click. `filterProjects` fails closed on `null` counts (`backlog-view.ts:186-188`).

**Every `selectedPath` writer accounted for.** Writers are `BacklogView.svelte:276` (pinned seed), `:290` (drop), `:309` (EPIC target), `:333` (just-added repo), `:346` (`handleSelect`), `:354` (dismiss). The two that already force `activeTab = "issues"` (`:310`, `:334`) are untouched, matching the plan's stated decision; `:333`'s effect clears both chips first (`:330-331`), so no interaction with the new rule.

**The pinned-seed effect's dependency set is unchanged.** `BacklogView.svelte:269-280` still tracks only `payload?.pinnedPath` and `mobile`; `selectedPath`/`visibleProjects` remain inside the pre-existing `untrack`, and the new chip read is itself wrapped (`untrack(() => tabForFilters({ hasIssues, hasPRs }))` at `:277`) — the object literal is constructed *inside* the callback, so neither `$state` is registered as a dependency and a chip toggle alone cannot fire this effect. The write is gated behind the existing `selectedPath === null` guard (`:274`), so a poll cannot re-tab a live selection. `untrack` is already imported (`:2`). The effect writes `activeTab` but never reads it, and no derived in the file depends on `activeTab` (`selected`/`actionsState` derive from `selectedPath`/`payload`, `:240-246`), so there is no self-retrigger.

**Test fixtures resolve to real selectors.** `.project-row` / `.row-name` / `.sel` — `ProjectRow.svelte:61-71` (`class:sel={selected}` at `:62`); `.filter-chip` — `ProjectBacklogList.svelte:97,106,116`; `.tab-btn` in fixed order issues/prs/actions/readiness/automation with `class:active` — `BacklogTabBar.svelte:56,66,76,91,99`, none conditionally rendered or `disabled`; wrapper is `.overlay-tabs` for `variant="mobile"` and `.tab-bar` otherwise (`BacklogTabBar.svelte:54`); `.mobile-detail-overlay` — `BacklogView.svelte:402`. The desktop tab bar sits outside the `{#if selectedPath !== null}` block (`BacklogView.svelte:480` vs `:496`), so the pre-selection `click(tabs(".tab-bar").prs)` in the new tests is valid. `row("repo-0")` / `row("nopr")` match `ProjectRow`'s basename `displayName` (`ProjectRow.svelte:44-46`). `m.backlog_filter_has_prs` / `m.backlog_filter_has_issues` exist in both catalogs (`ui/messages/en.json:879-880`, `ui/messages/de.json:879-880`) and are the chip labels used at `ProjectBacklogList.svelte:103,112`. `btns[0]`-style indexing is fine: `ui/tsconfig.json` sets only `strict: true` (no `noUncheckedIndexedAccess`), consistent with existing unguarded index access at `ui/src/lib/components/CommandBar.browser.test.ts:688`.

**Fresh `payload` object really re-fires the seed.** `BacklogOverlay.svelte:121` passes `{payload}` straight through to `BacklogView`, so a new object from `seedPayload()` changes the prop signal's value and the effect at `BacklogView.svelte:269` re-runs (same mechanism the pre-existing `rerender(props({ payload: payload(2) }))` test at `BacklogOverlay.browser.test.ts:119` relies on). `seedPayload()`'s shape matches the existing `payload()` factory's `BacklogPayload` fields.

**No unmocked-api fallout from the newly-rendered detail panels.** These are the first tests in this file to render a detail panel; `docAgentEnabled` defaults to `false` (`BacklogView.svelte:45`) so the doc-agent effects bail (`:156-163`), and the three panels the tests reach all swallow their fetches: `IssuesPanel.svelte:244,261,315,334,388`, `PrsPanel.svelte:94`, `ActionsPanel.svelte:44`.

**Announcement + i18n.** Entry shape matches its sibling exactly (`ui/src/lib/feature-announcements/entries/v1.46.0-pr-filter-tab.ts` vs `v1.46.0-opus-5-pin.ts`: same relative type import, `satisfies FeatureAnnouncement`, default export). The catalog is glob-loaded (`ui/src/lib/feature-announcements.ts:47-50`), so no index edit is needed. `id: "pr-filter-tab"` is unique across `entries/`. Filename version `v1.46.0` equals `sinceVersion` and is `> 1.45.0` from `.release-please-manifest.json:2`, satisfying both assertions in `scripts/check-announcement-versions.mjs:78-101`. `titleKey`/`bodyKey` exist in **both** catalogs at the same position — `ui/messages/en.json:3412-3413` and `ui/messages/de.json:3412-3413` — so the parity gate (`ui/scripts/check-i18n.mjs:44-53`) is satisfied, and neither value is empty. No `[[term|label]]` markers, so `scripts/check-glossary.mjs` is unaffected. The `feat(backlog):` commit touching `ui/src/lib/components/**` arms `scripts/check-feature-catalog.sh`, and an entry was added in-range. Escaped straight quotes in the EN body match established house style (54 other `\"` uses in `en.json`, e.g. `:1058`); the DE body's „…“ matches the DE style (e.g. `de.json:3406`). Pane and tab names in the copy match the UI: `actionbar_backlog` = "Repos" in both (`en.json:81`, `de.json:81`), `backlog_tab_prs` = "PRs" / `backlog_tab_issues` = "Issues" in both (`:823-824`).

## On the both-chips-on case

The literal task text says "if the 'has PR' filter is active, clicking a repo should open straight to the PR tab", and with *both* chips on this diff opens Issues (`backlog-view.ts:209-210`). That is the plan's explicitly operator-confirmed resolution of the ambiguity, and it's documented in the helper's doc comment and pinned by a unit test (`backlog-view.test.ts:265-267`), so I'm treating it as intended rather than a defect. Flagging it only so the reviewer sees the divergence from the literal wording was deliberate. Same for the symmetric `has issues` → Issues behavior and the pinned-seed redirect: both go beyond the one-line task but are named requirements in the approved plan, not drive-by additions.

The plan's stated Out-of-scope item (toggling a chip while a repo is already open does *not* move the tab) is indeed the behavior of this diff — the redirect fires only from `handleSelect` and the seed, never from `ontoggleprs`/`ontoggleissues` (`BacklogView.svelte:390-391`, `:468-469`).

## Nit

The EN announcement body writes `"has PRs"` lowercase but `"Has issues"` capitalized in the same sentence (`ui/messages/en.json:3413`). The actual chip labels are "Has issues" and "Has PRs" (`ui/messages/en.json:879-880`), and the DE body capitalizes both consistently („Hat PRs“ / „Hat Issues“, `de.json:3413`). Capitalize the EN one to match the button the user sees.

scoop (commented):
## Re-review

**Prior finding 1 (en.json chip-name casing) — ADDRESSED.** `ui/messages/en.json:3413` now reads `with \"Has PRs\" on, ... \"Has issues\" does the same for Issues`. Both forms match the real chip labels verified at `ui/messages/en.json:879` (`"Has issues"`) and `ui/messages/en.json:880` (`"Has PRs"`), which are the strings actually rendered by the chips (`ui/src/lib/components/ProjectBacklogList.svelte:96-113` renders `m.backlog_filter_has_issues()` / `m.backlog_filter_has_prs()`). The DE body at `ui/messages/de.json:3413` uses „Hat PRs“ / „Hat Issues“, matching `ui/messages/de.json:879-880`.

## What I verified

**Rule helper.** `tabForFilters` is defined at `ui/src/lib/components/backlog-view.ts:205-212` returning `"issues" | "prs" | null`, with `hasIssues` short-circuiting first (both-on → Issues). Assigned into `activeTab`, declared `$state<Tab>` where `type Tab = "issues" | "prs" | "actions" | "readiness" | "automation"` (`ui/src/lib/components/BacklogView.svelte:133-134`) — both returned literals are members of that union, and the `if (tab)` guard keeps `null` from being assigned. It is a pure addition; no existing export in `backlog-view.ts` changed signature, and the only consumers are `BacklogView.svelte:21` (import) and `backlog-view.test.ts:24`.

**Both selection call sites are wired.** `onselect={handleSelect}` appears at `ui/src/lib/components/BacklogView.svelte:394` (mobile branch, the list above the `.mobile-detail-overlay` at line 402) and `:472` (desktop branch, inside `.master-pane`). The prop contract is `onselect: (path: string) => void` (`ui/src/lib/components/ProjectBacklogList.svelte:47`), invoked as `onselect(project.path)` at `ProjectBacklogList.svelte:153/166/187` — including the hidden group, so a dimmed row re-tabs consistently. `handleSelect` (`BacklogView.svelte:344-349`) matches that signature.

**Pinned auto-seed.** `BacklogView.svelte:270-280`: the effect's only tracked read remains `payload?.pinnedPath`; `selectedPath`, `visibleProjects` and now `tabForFilters({ hasIssues, hasPRs })` are all inside `untrack`, so a chip toggle alone cannot make the effect fire, and the `selectedPath === null` guard still means a poll can only re-tab a seat that is already empty. The effect does not read `activeTab`, so writing it introduces no self-retrigger. I traced the other three `selectedPath` writers (`:290` drop-effect → `null`; `:309-310` EPIC target; `:333-334` just-added repo) — the latter two still force `activeTab = "issues"` unchanged, matching the plan's stated exclusions, and neither can be reached through `handleSelect`.

**Test fixtures resolve against real markup.** `.project-row` / `class:sel` / `.row-name` exist at `ui/src/lib/components/ProjectRow.svelte:61-62,71`, and `.row-name` renders the path basename (`ProjectRow.svelte:43-45`), so `row("repo-0")` (path `/r/repo-0`) and `row("nopr")` (`/r/nopr`) match. `.tab-btn` order is fixed issues/prs/actions/readiness/automation with no conditional rendering (`ui/src/lib/components/backlog-view/BacklogTabBar.svelte:56-96`), so index 1 = PRs and 2 = Actions as the helper assumes; the wrapper is `.tab-bar` for desktop and `.overlay-tabs` for mobile (`BacklogTabBar.svelte:55`). The desktop tab bar is rendered outside the `selectedPath !== null` guard (`BacklogView.svelte:479-491`), so the tests that click a tab before selecting a repo are valid. `chip()` label-matching is unambiguous: `"Has issues"` does not contain `"Has PRs"`.

**Announcement + i18n.** `ui/src/lib/feature-announcements/entries/v1.46.0-pr-filter-tab.ts` mirrors the sibling entry `v1.46.0-opus-5-pin.ts` exactly in import path and shape; entries are auto-registered by `import.meta.glob("./feature-announcements/entries/*.ts")` (`ui/src/lib/feature-announcements.ts:46-49`), so no index edit is needed. `id: "pr-filter-tab"` is unique across `entries/` and matches the `v<sinceVersion>-<id>.ts` filename contract; `sinceVersion: "1.46.0"` matches the already-landed unreleased entries. Key parity holds: `feat_pr_filter_tab_title` / `feat_pr_filter_tab_body` exist in both `ui/messages/en.json:3412-3413` and `ui/messages/de.json:3412-3413`. Escaped straight quotes in the EN body follow the existing catalog convention (e.g. `ui/messages/en.json:216`). `scripts/check-glossary.mjs` only validates `[[id|label]]` markers, of which the new strings have none.

- `ui/src/lib/components/BacklogView.svelte`: the symmetric `has issues` → Issues redirect is behaviour the task text did not ask for (it asked only about the PR chip); it is explicitly in the approved plan and is low-risk (Issues is already the default tab, chips are opt-in), so it is the author's call to keep.

## Note on the both-chips-on case (informational)
The task text says "if the 'has PR' filter is active, clicking a repo should open straight to the PR tab"; with BOTH chips on, the implementation opens Issues (`backlog-view.ts:209`). The plan documents this as operator-confirmed, so I am not raising it — flagging only so the reader knows the divergence from the literal task wording is deliberate, not an oversight.

## Limitations
- I did not run the suite or the gates (CI's job); test assertions were checked for identifier/selector validity only, not executed.

PR #1945, opened 2026-07-25
fix(security): refuse symlinked TODO.md atomically; drop check-then-use pairs

scoop:
Closes CodeQL alerts **#26, #31, #32, #93** on merit — each by removing the flagged
construct, not by suppressing it.

## What changed

**`src/repos.ts` → `writeTodo` + `readTodo` (#26 `js/file-system-race`).**
`writeTodo` already refused a symlinked `TODO.md`, but via `lstat`-then-write — a
check-then-use pair a swap inside the window could defeat. `readTodo` had **no symlink
guard at all** and would render any server-readable file into the operator's TODO pane.
Both now open with `O_NOFOLLOW`, so the refusal happens in the same syscall as the open.
The directory is unaffected: `safeRepoDir` already `realpath`s and containment-checks it,
so only the final component could be a symlink — which is exactly what `O_NOFOLLOW` covers.

Two deliberate details:

- **`readTodo` maps the symlink case to `{ok: true, exists: false}`, not `ok: false`.**
  The repo path is valid, so `ok: false` would surface as a `400 "invalid repo path"`; and
  it would regress a *dangling* `TODO.md` symlink, which returns `exists: false` today. A
  test pins both. Reading as "no TODO yet" leaks nothing and keeps the pane usable, and
  `writeTodo` still refuses the same file loudly.
- **`ELOOP` *and* `EMLINK` are both treated as the refusal.** POSIX/Linux/macOS report the
  former, some BSDs the latter, and CI only ever exercises Linux — so accepting both
  removes a platform assumption no test could catch.

**`src/commands.test.ts` → `tmpRoot` (#31, #32 `js/insecure-temporary-file`).**
`mkdirSync(join(tmpdir(), …randomUUID()))` → `mkdtempSync`. The UUID name was unguessable,
but `mkdtemp` is the sanctioned primitive and does the create in the same call.

**`test/herdr-install.test.ts` → `curlCalls` (#93 `js/unnecessary-use-of-cat`).**
Drops an `execFileSync("cat", …)` subprocess for `readFileSync`. The `existsSync` guard
goes with it — keeping it would have reproduced the very check-then-use shape flagged
elsewhere in this repo; a local `ENOENT` catch handles "shim never invoked", and any other
errno still throws instead of reading as "no calls".

## Tests

`test/repos.test.ts` gains three cases. The `readTodo` symlink test is genuinely
fail-first — against `main` it returns the sentinel contents of a file outside the repo:

```
- "content": "",           + "content": "SENTINEL-must-not-leak",
- "exists": false,         + "exists": true,
```

The other two are guards, expected to pass before *and* after: the `writeTodo` symlink case
(that refusal already existed — what changed is that it became atomic, which no test can
demonstrate since the race is not deterministically provokable), and a **shrink**
round-trip. The shrink is new coverage: every existing round-trip only ever *grows*
`TODO.md`, and `beforeEach` mints a fresh root per test, so nothing would have caught a
missing `O_TRUNC` — writing to a numeric fd does not truncate.

Verified beyond the suite: no fd leak across 400 mixed calls *including the throw paths*,
existing file mode preserved (`0600` not widened by `O_CREAT`'s `0666`), and a directory
named `TODO.md` still raises `EISDIR` exactly as before.

## Not in this PR

The remaining 25 open alerts were triaged individually against the analysis SARIF rather
than the query titles. 23 are dismissals (11 `used in tests`, 9 `false positive`, 3
`won't fix`) with per-alert rationales — see the manual step below. **#25 and #27 are
fixable on merit and are deliberately held back**; a related hardening question is being
tracked privately, and they will land with it.

```shepherd:manual-steps
- [ ] Apply the 23 code-scanning dismissals: this repo has delegated alert dismissal enabled, so they cannot be applied by an agent token. The per-alert reason + rationale list is in the session recap.
- [ ] POST-MERGE: after the next CodeQL run on main, confirm #26/#31/#32/#93 read `fixed` and dismiss any fingerprint recurrence (compare by rule+path+symbol, not alert number — a recurrence returns under a new number).
```

### DIFF
--- a/src/commands.test.ts
+++ b/src/commands.test.ts
@@ -1,5 +1,5 @@
 import { afterEach, describe, expect, it } from "bun:test";
-import { mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { mkdirSync, mkdtempSync, rmSync, writeFileSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { listCommands } from "./commands";
@@ -7,8 +7,9 @@ import { listCommands } from "./commands";
 const roots: string[] = [];
 
 function tmpRoot(): string {
-  const root = join(tmpdir(), `shepherd-commands-${crypto.randomUUID()}`);
-  mkdirSync(root, { recursive: true });
+  // mkdtempSync both creates the dir 0700 and picks the unpredictable suffix itself — the
+  // sanctioned primitive, and one call instead of hand-rolling a name then creating it.
+  const root = mkdtempSync(join(tmpdir(), "shepherd-commands-"));
   roots.push(root);
   return root;
 }
--- a/src/repos.ts
+++ b/src/repos.ts
@@ -4,10 +4,12 @@ import {
   writeFileSync,
   existsSync,
   statSync,
-  lstatSync,
   realpathSync,
   mkdirSync,
   rmSync,
+  openSync,
+  closeSync,
+  constants,
 } from "node:fs";
 import { execFile } from "node:child_process";
 import { promisify } from "node:util";
@@ -132,6 +134,22 @@ export function reconcileRealPathsToRaw(
 
 const TODO = "TODO.md";
 
+/**
+ * `O_NOFOLLOW` makes the open itself refuse a symlinked final component, replacing a
+ * check-then-use pair (`lstat`/`existsSync` then read/write) that a swap between the two
+ * calls could defeat. Only the FINAL component is covered — safe here because
+ * {@link safeRepoDir} already `realpathSync`es the directory and enforces containment
+ * before we build the path, so `dir` cannot itself be a planted symlink.
+ *
+ * POSIX, Linux and macOS report `ELOOP`; some BSDs report `EMLINK`. Both are accepted so
+ * the refusal does not depend on a platform the tests never exercise.
+ */
+const SYMLINK_ERRNOS = new Set(["ELOOP", "EMLINK"]);
+
+function isSymlinkRefusal(e: unknown): boolean {
+  return SYMLINK_ERRNOS.has(String((e as NodeJS.ErrnoException).code));
+}
+
 export function readTodo(
   repoPathRaw: string,
   repoRoot: string,
@@ -139,8 +157,25 @@ export function readTodo(
   const dir = safeRepoDir(repoPathRaw, repoRoot);
   if (!dir) return { ok: false, exists: false, content: "" };
   const file = join(dir, TODO);
-  if (!existsSync(file)) return { ok: true, exists: false, content: "" };
-  return { ok: true, exists: true, content: readFileSync(file, "utf8") };
+  let fd: number;
+  try {
+    fd = openSync(file, constants.O_RDONLY | constants.O_NOFOLLOW);
+  } catch (e) {
+    // Absent is the normal empty-TODO case. A SYMLINKED TODO.md reads as absent too:
+    // `ok: false` would surface as a 400 "invalid repo path" on a perfectly valid repo,
+    // and would regress a DANGLING symlink, which returns `exists: false` today.
+    // Reporting "no TODO yet" leaks nothing and keeps the pane usable; `writeTodo` still
+    // refuses that same file, so the symlink cannot be silently written through either.
+    if ((e as NodeJS.ErrnoException).code === "ENOENT" || isSymlinkRefusal(e)) {
+      return { ok: true, exists: false, content: "" };
+    }
+    throw e;
+  }
+  try {
+    return { ok: true, exists: true, content: readFileSync(fd, "utf8") };
+  } finally {
+    closeSync(fd);
+  }
 }
 
 export function writeTodo(repoPathRaw: string, repoRoot: string, content: string): boolean {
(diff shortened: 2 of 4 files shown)


scoop (commented):
## What I checked

Four files, all matching the plan's Files table: `src/repos.ts`, `src/commands.test.ts`, `test/repos.test.ts`, `test/herdr-install.test.ts`. No diff touches `src/shepherd-exclude.ts` or `src/gitignore-adopt.ts`, which the plan puts explicitly out of scope for this session — confirmed via `git diff --stat`.

### `src/repos.ts` — `writeTodo` / `readTodo`

- **The doc comment's load-bearing claim is true.** `src/repos.ts:140-142` asserts that only the final path component needs `O_NOFOLLOW` because `safeRepoDir` already realpaths the directory and enforces containment. Verified against `src/validate.ts:891-892` (`realpathSync` on both `repoRoot` and `repoPathRaw`), `:896` (`resolvedReal === rootReal || startsWith(rootReal + sep)`) and `:899` (`statSync(...).isDirectory()`). So `dir` at `src/repos.ts:157`/`:182` is a realpath — no intermediate component can be a planted symlink, and `O_NOFOLLOW`'s final-component-only coverage is sufficient here. (This is the asymmetry the plan calls out vs. `ensureShepherdExclude`, which is not in this diff.)
- **`O_TRUNC` is genuinely required and is present.** `writeFileSync(fd, …)` writes at the current file position and does not truncate; `src/repos.ts:193` includes `O_TRUNC`, and `test/repos.test.ts:200-205` is the only test in the file that shrinks a TODO (5 000 chars → `"short"`) — the pre-existing round-trips at `test/repos.test.ts:176-182`, `:188-193` only ever grow, and `beforeEach` (`test/repos.test.ts:36-41`) mints a fresh root per test, so nothing else would have caught a missing `O_TRUNC`.
- **No fd leaks.** Both `openSync` sites use the two-stage shape: the `finally` (`src/repos.ts:176-178`, `:202-204`) wraps only the region where `fd` is valid, and a failed `openSync` never produced one. The `catch` blocks exit on every path (`return` or `throw`), so TypeScript's definite-assignment analysis on `let fd: number` holds — the post-`try` join reaches only from the try block's normal completion.
- **`ok:true, exists:false` (not `ok:false`) for ELOOP/EMLINK is the right mapping.** Verified against `src/server.ts:6202-6206`: `todoRead` turns `!r.ok` into `400 {"error":"invalid repo path"}`, which would be a wrong message on a valid repo path and would also flip today's dangling-symlink result. `test/repos.test.ts:257-269` pins that dangling case.
- **No dead or missing imports.** `existsSync` and `statSync` are still used in `src/repos.ts` at `:78`, `:228`, `:560`, `:771`; `readFileSync`/`writeFileSync` at `:175`, `:201`, `:622-628`. `lstatSync` was the only import removed and has no remaining references (grepped).
- **No behavioural regression on the error paths.** Pre-change, `existsSync` → `readFileSync(path)` already threw for EACCES/EISDIR; post-change `openSync`/`readFileSync(fd)` throw the same errnos in the same cases (`O_RDONLY` on a directory succeeds on Linux, then `readFileSync(fd)` raises EISDIR — identical to before). `writeTodo`'s ENOENT (missing dir) still propagates rather than being swallowed, which is correct given `O_CREAT`.

### `test/herdr-install.test.ts` — `curlCalls`

- `readFileSync` is added to the `node:fs` import (`:9`) and used at `:195`. `existsSync` is still needed — `:204`, `:217`, `:258`, `:271`, `:284`, `:296`, `:327`. `execFileSync` is still needed — `:218`. So neither import is now dead.
- The log is plain text appended by the shims (`:150`, `:166` — `printf '%s\n' "$*" >> '${curlLog}'`), rooted at a `mkdtempSync` dir (`:130`, `:134`), so `readFileSync(path, "utf8")` is byte-equivalent to the `cat` it replaces, and the ENOENT-only catch correctly maps "shim never ran" to `""` while still throwing on any other errno.

### `src/commands.test.ts` — `tmpRoot`

- `mkdtempSync(join(tmpdir(), "shepherd-commands-"))` both creates and names the dir, so dropping the separate `mkdirSync` call is right; `mkdirSync` is still imported and used at `:19` and `:68`, so the import is not dead.

### Tests

- All identifiers the new tests use are imported: `symlinkSync`, `mkdtempSync`, `existsSync`, `writeFileSync`, `rmSync` at `test/repos.test.ts:2-10`, `tmpdir` at `:11`, `readTodo`/`writeTodo` at `:18-19`.
- `test/repos.test.ts:224-236` is a genuine assertion: the target is deliberately dangling, so `existsSync(target) === false` is what proves `O_CREAT` did not create through the link.
- `test/server.test.ts:1150-1154` (`GET /api/todo?repo=/etc/passwd → 400`) is unaffected — that path is `safeRepoDir`'s not-a-directory rejection, not the symlink branch.

## Stated limitations (could not verify, not asserted correct)

- The fd forms `readFileSync(fd, "utf8")` (`src/repos.ts:175`) and `writeFileSync(fd, content, "utf8")` (`:201`) are new to this repo — every other fd site uses `readSync` (`src/activity.ts:160-176`, `src/codex-usage.ts:270-274`, `src/codex-session-id.ts:81-90`). I could not execute anything, so I am assuming Bun's `node:fs` matches Node's contract that a caller-supplied descriptor is not closed by these calls (if it did close it, the `closeSync` in the `finally` would raise EBADF). The pre-existing happy-path round-trip at `test/repos.test.ts:176-182` exercises exactly that, so CI covers it.
- I could not verify the 23 CodeQL dismissals or the pre-merge gate — those are API actions, not diff content.

- `src/repos.ts`: the `readTodo` hardening closes no code-scanning alert (CodeQL flags neither its race nor its symlink follow) — the plan flags it as deliberate scope beyond the board and droppable. It is a real asymmetry in the pair being edited and I would keep it, but it is a user-visible behaviour change bundled into an alert-clearing PR.

- `ui/src/lib/components/TodoPanel.svelte:31`, `:43`, `:55` — every `putTodo(...)` call swallows failures with `.catch(() => {})`, so a `400` from `writeTodo`'s symlink refusal is invisible to the operator and the optimistically-rendered item is silently lost. Pre-existing, but it is what makes the `readTodo` finding below matter.
- `src/server.ts:6202-6206` — `todoRead` collapses every `ok:false` into the single message `"invalid repo path"`, and `todoWrite` (`:6217-6218`) collapses two distinct causes into `"invalid repo path or content too large"`.
- Alerts #25 and #27 remain open on the board by design (private track); nothing in this diff addresses them, which matches the plan.

PR #1946, opened 2026-07-25
fix(review): stop silent critic timeouts; make the deadline configurable

scoop:
## Problem

Two PRs (topmedia-website#171, flowagent#836) had **every** critic review error out with no diagnosable cause. Investigating the DB:

| session | review spawns | durations |
|---|---|---|
| TASK-1964 | 6 | 604, 607, 612, 612, 608, 611s |
| TASK-1971 | 4 | 603, 613, 614, 613s |

10/10 landed on `ReviewService.timeoutMs` — hardcoded `10 * 60 * 1000`, never passed at construction. The critic was still working when it was reaped: TASK-1964's last transcript entry is `19:09:32` ("Now let me verify the counter-precedent…"), finalize fired at `19:09:53`, after 61 read-only tool calls and zero writes.

Two things made this a black box:

1. **No trace.** `RecapService` logs a rich diagnostic on the identical failure (spawn id, cwd, model/provider/effort, elapsed, pane tail) — that's how recap's "Login expired" and weekly-limit failures are readable in `shepherd.log`. `ReviewService.finalize` had no equivalent; the persisted summary is a fixed string that can't distinguish "critic died early" from "ran out of time". Every manual re-trigger also resets `errorRound` to 0 (`review.ts` `forceReview`), so the consecutive-error stall signal never escalated either.
2. **No lever.** Because the critic restarts from scratch each time, a PR that needs >10 min is not *slowly* reviewed — it is *permanently* un-reviewable, and there was no way to give it more room.

Not a niche tail: 50 of 676 review spawns in the last 7 days (7.4%) ended in the 595–640s band, while 618 finished comfortably under it.

## Change

- **Diagnostic.** `logNoVerdict` logs before `finalize` (whose `finally` reaps the terminal, taking the pane tail with it): the cause — unparseable / hard timeout / exited early, distinguished because the fixes differ — plus `elapsed=Ns deadline=Ns`, the reviewer environment, and the redacted pane tail where a CLI-level failure (expired login, weekly-limit refusal) prints. Fully guarded: a diagnostic must never strand a finalize, and a herdr fake without `readAsync` degrades to no tail.
- **`SHEPHERD_REVIEW_TIMEOUT_MS`.** Clamped to `[1m, 60m]` via the existing `clampCap` — a NaN would make `elapsed > NaN` permanently false (a run that never finalizes, leaking its worktree and terminal), and an unbounded value is the same hazard. **Default is unchanged at 10m**, so merging alters no behavior. Wired into both the session critic and the standalone PR critic: one critic role, one deadline — leaving the twin at the service default would half-apply an operator's raise.

Env-only rather than a UI setting: it's a machine-speed/PR-size escape hatch, not a product knob.

## Tests

5 added to `test/review.test.ts` (7714 pass / 0 fail overall): the timeout diagnostic's numbers and pane tail; graceful degradation when herdr has no `readAsync`; secret redaction of the tail; early-exit vs. hard-timeout wording; and the config clamp (unset → 10m, garbage → 10m, out-of-range → snapped).

## Not covered

`StandalonePrCriticService` gets the configurable deadline but not the diagnostic — it has no `herdr.list()` dep, so it cannot resolve a pane to tail. Worth a follow-up if that critic starts failing silently too.

[no-feature-entry]

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -127,6 +127,13 @@ export const DISTILLER_INTERVAL_DAYS_MAX = 14;
 // module-local seed defaults, used by the config seeds + boot-override fallbacks only.
 const PR_REVIEW_CYCLES_DEFAULT = 3;
 const PLAN_REVIEW_CYCLES_DEFAULT = 5;
+// Hard deadline for a single critic run (both the session critic and the standalone PR critic):
+// how long to wait for the verdict file before giving up and finalizing an `error` verdict.
+// Bounded so a typo can't wedge a run forever (no ceiling → a leaked worktree + terminal per PR)
+// nor reap it before it can start.
+export const REVIEW_TIMEOUT_MS_MIN = 60_000;
+export const REVIEW_TIMEOUT_MS_MAX = 60 * 60_000;
+const REVIEW_TIMEOUT_MS_DEFAULT = 10 * 60_000;
 // Coerce any input (env/DB/request) to a valid integer cap, snapping out-of-range
 // values into [min,max] rather than rejecting (callers stay forgiving); a non-finite
 // input falls back to the supplied default.
@@ -796,6 +803,18 @@ export const config = {
     PR_REVIEW_CYCLES_MIN,
     PR_REVIEW_CYCLES_MAX,
     PR_REVIEW_CYCLES_DEFAULT,
+  ),
+  // Hard deadline for one critic run before it is abandoned with an `error` verdict. Default 10m
+  // (unchanged). Raise it for a repo whose PRs genuinely outgrow that: a large diff reviewed at
+  // criticEffort=high can still be mid-investigation at the deadline, and because the critic
+  // restarts from scratch every time, EVERY retry then dies at the same wall — the PR becomes
+  // permanently un-reviewable rather than slowly reviewed. Env-only (not UI-configurable): it is a
+  // machine-speed/PR-size escape hatch, not a product knob.
+  reviewTimeoutMs: clampCap(
+    Number(process.env.SHEPHERD_REVIEW_TIMEOUT_MS ?? REVIEW_TIMEOUT_MS_DEFAULT),
+    REVIEW_TIMEOUT_MS_MIN,
+    REVIEW_TIMEOUT_MS_MAX,
+    REVIEW_TIMEOUT_MS_DEFAULT,
   ),
   // Max plan-gate adversarial-review rounds before escalating to a human (drives
   // PlanGateService). UI-configurable + persisted; the env seeds the value on a fresh DB.
--- a/src/index.ts
+++ b/src/index.ts
@@ -1412,13 +1412,18 @@ const reviewService = new ReviewService({
   // global, UI-configurable max auto-address rounds before escalating to the human.
   // A thunk so a settings change takes effect on the next critic run, no restart.
   cap: () => config.prReviewCyclesCap,
+  // Hard per-run deadline (SHEPHERD_REVIEW_TIMEOUT_MS). A plain number, not a thunk: it is read
+  // once at construction because it is env-seeded, not a live UI setting.
+  timeoutMs: config.reviewTimeoutMs,
 });
 
 // Standalone repo-level PR critic (#596): the session-LESS twin of reviewService.
 // Where reviewService reacts to a managed session's PR, this enumerates EVERY open,
 // CI-green PR in a `criticAllPrs` repo (human PRs, other agents', forks) on a timer and
-// posts comment-only reviews. Shares reviewService's primitives + the same per-role
-// `criticModel` setting; concurrency/timeout stay at service defaults.
+// posts comment-only reviews. Shares reviewService's primitives, the same per-role
+// `criticModel` setting, and the same hard per-run deadline (one critic role, one deadline —
+// leaving this twin at the service default would half-apply an operator's raise); concurrency
+// stays at the service default.
 const standaloneCritic = new StandalonePrCriticService({
   store,
   herdr,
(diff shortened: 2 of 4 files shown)


scoop (commented):
## What the change does

Two things, both aimed at the reported symptom (session-critic reviews finalizing as `error` with no trace):

1. `src/review.ts` logs *why* a run is about to finalize with a null verdict (cause, environment, elapsed-vs-deadline, redacted pane tail).
2. `src/config.ts` + `src/index.ts` make the previously hardcoded 10-minute critic deadline settable via `SHEPHERD_REVIEW_TIMEOUT_MS`, clamped to [1m, 60m].

## Verification performed

- **`raw === null` really is the error path.** `buildVerdictCore` returns `decision: "error"` and the fixed summary `"critic did not produce a verdict"` whenever `raw` is null — verified against `src/critic-core.ts:901` and `src/critic-core.ts:905`. And `raw` is null exactly on `finalize-null`, per `src/json-tolerant.ts:155-169`. So the new `if (raw === null)` guard at `src/review.ts:822` fires on precisely the invisible-error case it claims to cover, and only there.
- **`read.raw` exists on the `unparseable` variant** — `VerdictRead` is declared at `src/json-tolerant.ts:56-59` with `{ status: "unparseable"; raw?: string }`, so the `read.status === "unparseable" && read.raw` branch at `src/review.ts:857` typechecks and is not always-false.
- **`sanitizeRecapFailureDetail` is exported and does what the diff relies on.** Declared `export function` at `src/recap.ts:201`; it returns `string | undefined` (hence the `?? ""` at `src/review.ts:875`), redacts `*KEY=`-style assignments at `src/recap.ts:210` and `sk-`/`gh[pousr]_`/`xox*` tokens at `src/recap.ts:218`, and — importantly for the unbounded `snippet:` interpolation at `src/review.ts:858` — **self-bounds to 300 chars** at `src/recap.ts:223`. So the raw verdict body cannot dump 20KB into the log; no finding there.
- **No import cycle from the new `./recap` import.** `src/review.ts` is imported only by `src/index.ts:87` (grep for `from "./review"`), and `src/recap.ts` imports neither it nor anything that reaches it.
- **`herdr.readAsync(terminalId, "recent", 20)` is the right call with the right handle.** `readAsync(target, source?, lines?)` is on the driver interface at `src/herdr.ts:714` and its implementation resolves the argument via `resolveTargetAsync(terminalId)` at `src/herdr.ts:1153-1156` — so passing `f.terminalId` (not a paneId as `src/recap.ts:984` does) is correct, not a mismatch. It is bounded by `HERDR_TIMEOUT_MS` (`src/herdr.ts:106`), so it cannot stall the tick loop.
- **The widened `herdr` Pick does not break production wiring.** `src/index.ts:534` types the injected driver as `IHerdrDriver`, which declares `readAsync` (`src/herdr.ts:714`), and the socket implementation provides it (`src/herdr-socket-driver.ts:67`). `ReviewServiceDeps` has no other consumer in the repo (grep: only `src/review.ts:137`/`:255`), and every test construction goes through `as any`. The same `Pick<HerdrDriver, … | "readAsync">` shape is already used at `src/recap.ts:284`.
- **Every `InFlight` field interpolated into the log line exists** — `sessionId`, `criticSessionId`, `prNumber`, `worktreePath`, `reviewerProvider`, `reviewerModel`, `reviewerEffort`, `terminalId`, `startedAt` are all declared at `src/review.ts:90-114`.
- **The default is genuinely unchanged.** `REVIEW_TIMEOUT_MS_DEFAULT = 10 * 60_000` (`src/config.ts:136`) equals the service fallback `deps.timeoutMs ?? 10 * 60 * 1000` at `src/review.ts:257` and the standalone default at `src/standalone-critic.ts:70`. `StandalonePrCriticService` does accept `timeoutMs` (`src/standalone-critic.ts:101`, consumed at `:146`), so the new wiring at `src/index.ts:1436` is valid.
- **Test imports resolve.** `clampCap` is exported at `src/config.ts:140`, `REVIEW_TIMEOUT_MS_MIN`/`_MAX` at `src/config.ts:134-135`, `STARTUP_GRACE_MS` at `src/json-tolerant.ts:134`.
- **The new tests exercise the branches they claim.** With `criticAgentStatus` defaulting to `"idle"` and `paneForegroundProcs` to `["zsh"]`, `isSpawnAlive` reports finished (`src/json-tolerant.ts:110-115`); at `t+6000` with `timeoutMs: 5000` the run takes the timeout branch of `src/json-tolerant.ts:168`, and at `t+STARTUP_GRACE_MS+1000` with `timeoutMs: 300_000` it takes the `spawnFinished && graceElapsed` branch — matching the two cause strings each test asserts.
- Server-only change: no message catalogs, no `ui/` surface, so the i18n parity and feature-catalog gates in `CLAUDE.md` (lines 49-73) do not apply.

## Notes on the findings

The cause-classification nit is the only one touching behaviour: `tick()` already computes the authoritative `timedOut` in milliseconds (`src/review.ts:788`), but `logNoVerdict` re-derives it from *rounded seconds* (`src/review.ts:848`). Within a sub-second window (e.g. a spawn that exits at 599.7s under the 600s default), the two disagree and the log claims `hard timeout` for a run that actually finalized through the exited-spawn fail-fast path — i.e. exactly the distinction this diagnostic exists to make. Narrow, but the fix is to pass the boolean through.

## Limitation

I can only review the diff. Whether the reviews for tasks #1964 and #1971 were actually re-triggered in the running Shepherd instance is a runtime action with no artifact in this repo — I neither verified nor assumed it.

- `src/index.ts:1436` also applies the new deadline to `StandalonePrCriticService` (repo-level PR reviews), which the task — scoped to task/session reviews — did not ask for; the inline rationale ("one critic role, one deadline") is reasonable, and it is env-gated with an unchanged default, so it is the author's call to keep.

- `src/standalone-critic.ts:552-579` still finalizes with `raw = null` and no diagnostic at all, so the standalone PR critic keeps the exact black-box failure this PR fixes for the session critic — it now inherits the raised deadline but not the trace.
- `src/plan-gate.ts:335` keeps its own independent hardcoded 10-minute `DEFAULT_TIMEOUT_MS`, so a plan-gate reviewer that outgrows 10m remains permanently un-reviewable with no env escape hatch.
- `src/recap.ts:997-1007` computes its 300-char tail as `…` + `slice(-300)` (301 chars) before handing it to a sanitizer that truncates at 300 from the head, so the last character is always dropped; the copy in this diff inherits the same off-by-one.

Issue #1947, opened 2026-07-26
Add a density budget to the plan-authoring directive


scoop:
Split out of the plan-review/PR-critic convergence work (branch `shepherd/plan-review-pr-critic`) to keep that change scoped to the **reviewer** briefings.

## Problem

`planGateDirectiveInteractive()` / `planGateDirectiveAuto()` in `src/service.ts` mandate a plan section list (goal, approach, out of scope, files, testing seams + decisions, steps, risks, success criteria) with **no density budget**. First-round plans therefore arrive oversized before any review round has run.

Observed: TASK-1977's plan reached **32.6 KB with 28 success criteria**, three `why not X` essay sections, and a `## Reviewer findings — where each is addressed` ledger. A larger plan is a larger surface to attack, which drives more findings and more rounds.

The sibling change adds a revise-in-place instruction to `planSteerText()`, but that **only fires on rework rounds** — it cannot reach round 1.

## Proposal

Add a shared const beside `PLAN_REFERENCE_STYLE` in `src/service.ts`, consumed verbatim by both directive builders (the established pattern for a line both providers must receive identically):

- success criteria are **verifiable checks**, not a restatement of the plan
- record **decisions**, not arguments for them
- no rebuttal or `why not X` essay sections
- no ledger of addressed findings

Leave the mandated section list unchanged.

## Seam

`test/service-plan-gate.test.ts` already exercises both directive builders.

## Risk

This ships in the directive injected into **every** planning session, so wrong wording degrades plans broadly rather than in one session. Keep it to what is verifiable.

## Why deferred

Scope discipline on the sibling PR: that change is about how reviewers judge plans, this is about how planners write them. Worth doing, but separable and independently testable.

scoop:
✅ PR #2307 merged.

PR #1948, opened 2026-07-26
fix(review): converge plan-gate and PR-critic rework loops

scoop:
Plan and PR-critic rework cycles were running long — two sessions (TASK-1981, TASK-1977) exhausted a 12-round plan-gate budget twice, arguing over wording and accreting scope. This fixes the mechanisms in the reviewer briefings that made that possible.

## The problem is a tail, not a slowdown

`reviewer_spawns`, last 30 days:

| plan-gate rounds | sessions | % of plan-review tokens |
| --- | --- | --- |
| 1–3 | 493 | 25% |
| 4–7 | 168 | 21% |
| 8–12 | 40 | 15% |
| **13+** | **38 (5%)** | **39%** |

The PR critic mirrors it: 23 of 828 sessions (2.8%) at 9+ rounds burn 27% of critic tokens. The median (2–3 rounds) is healthy and must not move. Both caps were already at their maxima (`planReviewCyclesCap` 12, `prReviewCyclesCap` 8), which lengthened the tail rather than converging it — so the fix belongs in the briefings, not the limits.

## What was actually driving it

**The PR critic mandated the opposite of its own lenses.** `scopeAndOutputTail()` said verbatim *"A non-blocking nit STILL goes in `findings`"*, while the three lenses directly above it forbid exactly that, and explain why: `runAutoAddress()` fires on any non-empty `findings` **regardless of decision**, and the SCOPE rule re-raises unaddressed items every round. A comment nit therefore looped until the streak ceiling paused the PR.

**The plan reviewer had no non-blocking channel at all.** Its only contract was approve, or "request-changes with at least one finding" — so a wording preference had exactly one legal exit: a blocking finding costing a full rework round. TASK-1977's two surviving findings were *"Argue the rejection of clamp-only… instead of listing it"* and *"Give it a grounded paragraph in §2b at the same standard §2a applies to…"*. Neither is a plan defect.

**Nothing barred findings that demand new scope**, which is how a plan reaches 32.6 KB and 28 success criteria — and **the reviewer never learned which round it was on**, so round 11 was briefed exactly like round 1.

## Changes

- **`findings` is blocking-only on both sides.** Nits route to `Nits (non-blocking):` / `Suggestions (non-blocking):`, matching the existing lens routing. Any correctness, security, contract, or i18n-parity defect stays blocking *regardless of size*.
- **Plan approval bar** drops "genuinely the best reasonable path" (unfalsifiable — no plan ever wins it) for soundness: it passes when the approach will work and what remains is settleable during implementation.
- **Scope-demand lens** on the plan side: a finding may not require the plan to cover work the task never asked for. A gap blocks only when the task *as stated* is unsatisfied.
- **Prose-demand ban**: a finding may not ask for more argument about a decision already stated. Ask for a missing *decision*, not more prose about one already made.
- **Advisory 5-finding cap** that can never lose information — over five blockers, emit them all and say the work needs rewriting. Never sheds a blocker into the non-blocking section.
- **Stale-prior drop rule.** `PlanGate.findings` / `reviews.findings` persist across deploy, so without this the findings recorded under the *old* bar re-raise verbatim on the first review after it changes, and nothing converges faster on precisely the two sessions that motivated this.
- **ROUND block** on an effective round: `prior.round + 1` was wrong at both ends — `applyChangesRequested` holds at the cap while at-cap reviews still start, and `resume()` zeroes it — so it is clamped and floored by a reset-proof spawn count. Policy: a Resume unblocks a session, it does not restore full review latitude.

## Notes for review

- **The at-cap line deliberately hedges** ("may be the LAST that reach the author"). Asserting escalation there would be false: at `priorRound === cap - 1` the steer *is* still delivered. An earlier draft made that claim and would have suppressed real findings a round early.
- **The cap is not deterministically enforced**, unlike `scopeFindings` here or `MAX_ADDS_PER_RUN` in `distiller.ts`. Both of those enforce a *decidable* predicate; "is this finding blocking?" is not decidable server-side, so a truncation backstop would cut blind and could discard a real blocker while keeping a nit. The wording removes the need for one instead.
- **Biggest risk, stated plainly:** `findings` gates `autopilot.ts` (won't advance while non-empty), `signoff.ts`, and `learnings-lifecycle.ts` (clean = `commented` + zero findings). A nit-only review now reads as clean and stops blocking those. That is the intended speedup, but a mis-routed real issue would slip through.
- The effective round counts **error spawns** too, so a flapping critic reads as later than its delivered-round count. Accepted: the row represents real tokens spent, lateness never downgrades a correctness/security finding, and consecutive errors have their own ceiling.
- The ROUND block lives in `reviewPrompt`'s preamble, **not** the shared tail — the byte-identity test between `reviewPrompt` and `prReviewPrompt` still passes unmodified, and the standalone critic (no rework loop) is unaffected by it.

## Verification

`bun install && bun run lint && bun test` green (7749 pass, 0 fail). Branch hygiene, i18n parity, feature-catalog and glossary gates pass; `fallow audit` clean (`begin()` was pushed over its complexity threshold by the new call and refactored back under it).

**Effect is not verified by this PR** — the criteria above assert delivery, not outcome. The plan records a post-ship check: re-run the `reviewer_spawns` query in 30 days, expecting the `13+` bucket under 20% of tokens (from 39%) with the `1-3` bucket holding ≥65% of sessions. If `13+` is still above 30%, escalate to the deferred levers (passing a prior-plan diff, then a plan-gate rebuttal channel).

Density budget for the plan-*authoring* directive was split out to #1947 to keep this change scoped to the reviewer briefings.

### DIFF
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -638,6 +638,7 @@ function scopeAndOutputTail(
     "- A guard/validation present on one code path but MISSING from its sibling path (e.g. one branch floors a value with Math.max(0, …) and a parallel branch computing the same kind of value does not) is a defect even when the unguarded path is currently unreachable.",
     '- A bug currently unreachable but made reachable by change THIS PR foreshadows (a param wired only in tests, a value a follow-up will populate, a path behind a not-yet-set flag) is real — "descoped", "handled in another ticket", or "never reached in production" does NOT make such an in-diff defect a non-issue.',
     '- Route by reachability TODAY. If the defect is reachable on a path that ALREADY executes, treat it as a normal bug: put it in "findings" and block it per the usual rules. If it is reachable ONLY through the foreshadowed-but-not-yet-wired future above (dormant today), it is informational: report it in a SINGLE "body" section headed exactly `Latent / future-reachable (non-blocking):`, ONE LINE PER DISTINCT ITEM, do NOT put it in "findings", and it NEVER makes the decision "request-changes". Either way it must concern a file in the diff per the SCOPE rule above.',
+    "",
     // FINDINGS ROUTING (#1948). The tail used to mandate the exact opposite — "A non-blocking nit
     // STILL goes in findings" — which contradicted the three lenses above and their stated reason:
     // ANY entry in `findings` advances the streak (buildVerdict), is auto-addressed (runAutoAddress
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -150,7 +150,15 @@ export function reviewPrompt(
   authorNotes: string[] = [],
   issueBody?: string | null,
   epic?: EpicContext | null,
-  opts: { plan?: string | null; smellLens?: boolean } = {},
+  opts: {
+    plan?: string | null;
+    smellLens?: boolean;
+    /** The EFFECTIVE rework round (already clamped + spawn-count-maxed by the caller — see
+     *  ReviewService.begin), and the live cap. Both required for the ROUND block; absent ⇒ no block,
+     *  so every other caller stays byte-identical. */
+    round?: number;
+    cap?: number;
+  } = {},
 ): string {
   const lines = [
     "You are a code critic reviewing a pull request. Do NOT modify, build, commit, or run anything — read-only inspection only.",
@@ -184,6 +192,11 @@ export function reviewPrompt(
     lines.push(
       `This is a RE-REVIEW. The previous revision raised the points below. For EACH, confirm the new diff actually addresses it; if it does not, re-raise it verbatim in your findings — do not let it slide — UNLESS its file is not in \`git diff ${diffBase}...HEAD\`, in which case drop it per the scope rule below (do NOT re-raise it):`,
       ...priorFindings.map((f, i) => `${i + 1}. ${f}`),
+      // Severity-based drop, alongside the out-of-diff drop above (#1948). Prior findings PERSIST in
+      // `reviews.findings` across a deploy, so without this every nit recorded under the old
+      // "a non-blocking nit STILL goes in findings" contract is re-raised verbatim on the first
+      // review after that contract changed — and nothing converges any faster.
+      'A prior point that FINDINGS ROUTING below does not admit as blocking is likewise DROPPED — do not re-raise it in "findings". You MAY restate it once under `Nits (non-blocking):`. Saying in "body" that you dropped it, and why, IS compliance with the re-raise instruction above, not a violation of it.',
       "",
     );
   }
@@ -195,6 +208,10 @@ export function reviewPrompt(
       "",
     );
   }
+  // ROUND block (#1948) — session critic ONLY, and deliberately in this preamble rather than in
+  // scopeAndOutputTail: the tail is shared verbatim with prReviewPrompt (a test asserts byte
+  // identity), and the standalone critic has no rework loop to be late in.
+  lines.push(...roundBlock(opts.round, opts.cap, "Nits (non-blocking):", "the author"));
   // The judging clause is the ONE line that differs between the session critic ("satisfies that
   // task") and the standalone PR critic ("bugs/security/quality, intent as context") — everything
   // after it (the verdict-output contract) is identical, so it's factored into the shared tail.
@@ -455,6 +472,70 @@ function landingBlock(landing: LandingContext, diffBase: string): string[] {
   ];
 }
 
+/**
+ * The ROUND block (#1948), shared by the session PR critic ({@link reviewPrompt}) and the plan
+ * reviewer (`plan-gate.ts` → `planReviewPrompt`) so the two can never drift — the same reason
+ * {@link scopeAndOutputTail} exists. Emitted only when BOTH `round` and `cap` are supplied; absent
+ * ⇒ `[]`, so every caller that passes neither keeps a byte-identical prompt.
+ *
+ * `round` is the EFFECTIVE round the caller already computed (see {@link effectiveRound}) — this
+ * builder stays pure and does no clamping of its own.
+ *
+ * Two invariants the wording must preserve, both learned the hard way:
+ *  - Severity is never downgraded by lateness. A correctness/security defect is a finding at any
+ *    round, and an item that was blocking when first raised stays blocking — otherwise an
+ *    unaddressed blocker would be simultaneously must-re-raise and must-route-to-non-blocking.
+ *  - The last-round line HEDGES ("may be the last"). It is NOT safe to assert that findings stop
+ *    reaching the recipient here: at `priorRound === cap - 1` the steer is still delivered, and
+ *    because `round` is maxed with a spawn count it can reach `cap` while the delivered-round
+ *    counter is lower still. Asserting escalation would suppress real findings on a false premise.
+ */
+export function roundBlock(
+  round: number | undefined,
+  cap: number | undefined,
+  nonBlockingSection: string,
+  recipient: string,
+): string[] {
+  if (round === undefined || cap === undefined) return [];
+  if (!Number.isFinite(round) || !Number.isFinite(cap) || round < 1 || cap < 1) return [];
+  const n = round;
+  const m = cap;
+  const lines = [
+    `ROUND — this is rework round ${n} of at most ${m}.`,
+    "- A correctness or security defect is a finding at ANY round. Nothing below downgrades one.",
+    "- A point that was blocking when you first raised it STAYS blocking. The rules below govern only what you are noticing for the FIRST time now.",
+  ];
+  // Strictly past the halfway point (2n > m), so a 1-of-2 review is not already "late".
+  if (2 * n > m)
+    lines.push(
+      `- You are past the halfway point of the rework budget. Raise a NEW finding only if it would make the change fail outright; anything else you notice now goes to \`${nonBlockingSection}\`.`,
+    );
+  if (n >= m)
+    lines.push(
+      `- The budget is spent. Findings from this round may be the LAST that reach ${recipient} before the loop pauses for a human, so reserve blocking findings for what genuinely must not proceed.`,
+    );
+  lines.push("");
+  return lines;
+}
+
+/** The EFFECTIVE rework round both review loops brief their reviewer with (#1948).
+ *
+ *  `priorRound + 1` alone is wrong at BOTH ends, which is why this exists:
+ *   - Too high: `applyChangesRequested` HOLDS the round at the cap while `startedStatus` →
+ *     "started-at-cap" still starts reviews, so the raw value emits "round 13 of at most 12" and the
+ *     at-cap rule never fires on exactly the stalled sessions it is for. Hence the clamp.
+ *   - Too low: `resume()` writes `{ ...gate, round: 0 }` and is the ONLY operator escape offered at
+ *     the cap, so a plan reviewed 29 times would be briefed as round 1. Hence the max against
+ *     `reviewCount` — a reset-proof count of this session's reviewer spawns.
+ *
+ *  Policy this encodes: an operator Resume UNBLOCKS a session; it does not restore full review
+ *  latitude. `reviewCount` is itself clamped, so it can only ever raise the round to the cap.
+ */
+export function effectiveRound(priorRound: number, reviewCount: number, cap: number): number {
+  const clamped = Math.min(priorRound + 1, cap);
+  return Math.max(clamped, Math.min(reviewCount, cap));
+}
+
 function scopeAndOutputTail(
   diffBase: string,
   judgeClause: string,
@@ -557,9 +638,27 @@ function scopeAndOutputTail(
     "- A guard/validation present on one code path but MISSING from its sibling path (e.g. one branch floors a value with Math.max(0, …) and a parallel branch computing the same kind of value does not) is a defect even when the unguarded path is currently unreachable.",
     '- A bug currently unreachable but made reachable by change THIS PR foreshadows (a param wired only in tests, a value a follow-up will populate, a path behind a not-yet-set flag) is real — "descoped", "handled in another ticket", or "never reached in production" does NOT make such an in-diff defect a non-issue.',
     '- Route by reachability TODAY. If the defect is reachable on a path that ALREADY executes, treat it as a normal bug: put it in "findings" and block it per the usual rules. If it is reachable ONLY through the foreshadowed-but-not-yet-wired future above (dormant today), it is informational: report it in a SINGLE "body" section headed exactly `Latent / future-reachable (non-blocking):`, ONE LINE PER DISTINCT ITEM, do NOT put it in "findings", and it NEVER makes the decision "request-changes". Either way it must concern a file in the diff per the SCOPE rule above.',
+    // FINDINGS ROUTING (#1948). The tail used to mandate the exact opposite — "A non-blocking nit
+    // STILL goes in findings" — which contradicted the three lenses above and their stated reason:
+    // ANY entry in `findings` advances the streak (buildVerdict), is auto-addressed (runAutoAddress
+    // fires on non-empty findings REGARDLESS of decision), and is re-raised every round by the
+    // re-raise rule. A wording nit therefore looped until the streak ceiling paused the PR. So
+    // `findings` is now blocking-only, and nits get the same non-blocking body routing the lenses use.
+    'FINDINGS ROUTING — what belongs in "findings" and what does not:',
+    '- "findings" is for changes the author MUST make: a correctness bug, a security issue, a broken or violated contract, a missing locale/catalog counterpart, or a diff that does not do what the task/intent above requires. Any such defect is a finding REGARDLESS of how small it looks.',
+    '- A NON-BLOCKING nit — a naming or wording preference, a stylistic choice, a comment you would phrase differently, a refactor you would like but the task did not require — does NOT go in "findings". Report it in a SINGLE "body" section headed exactly `Nits (non-blocking):`, ONE LINE PER DISTINCT ITEM, and it NEVER makes the decision "request-changes". Each must concern a file in the diff per the SCOPE rule above.',
+    '- COMMENTS specifically: do NOT raise a finding asking for a comment to be reworded or expanded, or for more explanation of code that is already correct — that is a nit. A comment that is factually WRONG about what the code does IS a defect: raise it in "findings".',
+    // Advisory cap. Deliberately NOT enforced server-side: the repo's deterministic backstops
+    // (scopeFindings here, MAX_ADDS_PER_RUN in distiller.ts) each enforce a DECIDABLE predicate,
+    // whereas "is this finding blocking?" is not decidable from the verdict text. Truncating would
+    // cut blind and could discard a real blocker while keeping a nit — so the cap is worded so that
+    // it can never lose information instead.
+    '- At most 5 NEW findings per review, highest-severity first. This is a PRIORITISATION instruction, never a limit that may lose information: if more than 5 genuine blocking problems exist, emit ALL of them and say so in "summary" (the change needs reworking, not tweaking). NEVER move a blocking problem into `Nits (non-blocking):`, and never drop one, to fit the budget.',
+    "- Re-raised prior findings do NOT count against those 5.",
+    "",
     "When done, write your verdict as JSON to the file `.shepherd-review.json` in the repository root, with EXACTLY this shape:",
     '{"decision": "request-changes" | "comment", "summary": "<=100 char one-liner", "body": "<full markdown review>", "findings": ["<discrete actionable item>", ...]}',
-    'The "findings" array lists every discrete change the author must make — one entry per point, blocking or not. A non-blocking nit STILL goes in "findings" (under a "comment" decision). Use [] ONLY when there is genuinely nothing to address; "request-changes" requires at least one finding.',
+    'The "findings" array lists every discrete change the author must make — one entry per point, routed per FINDINGS ROUTING above. Use [] when everything you found is non-blocking (report those in the body sections named above) or there is genuinely nothing to address; "request-changes" requires at least one finding.',
     'Use "request-changes" ONLY for blocking problems (does not satisfy the task, logic bug, security hole). Otherwise use "comment". Never approve. Write the file as your final action, then stop.',
   ];
 }
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -505,8 +505,12 @@ export function roundBlock(
     "- A correctness or security defect is a finding at ANY round. Nothing below downgrades one.",
     "- A point that was blocking when you first raised it STAYS blocking. The rules below govern only what you are noticing for the FIRST time now.",
   ];
-  // Strictly past the halfway point (2n > m), so a 1-of-2 review is not already "late".
-  if (2 * n > m)
+  // Strictly past the halfway point (2n > m), so a 1-of-2 review is not already "late" — AND never
+  // on a first-ever review. `n > 1` is load-bearing, not belt-and-braces: both caps are UI-settable
+  // down to 1 (PR_REVIEW_CYCLES_MIN / PLAN_REVIEW_CYCLES_MIN), where 2n > m holds at n = 1 and would
+  // muzzle the only review the operator gets — every finding on a first review is new, so the
+  // blocking classes FINDINGS ROUTING names would all route to the non-blocking section.
+  if (n > 1 && 2 * n > m)
     lines.push(
       `- You are past the halfway point of the rework budget. Raise a NEW finding only if it would make the change fail outright; anything else you notice now goes to \`${nonBlockingSection}\`.`,
     );
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -503,7 +503,15 @@ export function roundBlock(
   const lines = [
     `ROUND — this is rework round ${n} of at most ${m}.`,
     "- A correctness or security defect is a finding at ANY round. Nothing below downgrades one.",
-    "- A point that was blocking when you first raised it STAYS blocking. The rules below govern only what you are noticing for the FIRST time now.",
+    // Scoped deliberately to "under FINDINGS ROUTING" and to "the rules below". Unqualified, this
+    // sentence contradicts the stale-prior DROP rule for exactly the priors that rule targets: the
+    // plan prompt had no non-blocking channel before this change, so EVERY stored PlanGate.findings
+    // entry was raised as blocking, and a literal reading would force all of them re-raised —
+    // defeating the mechanism that unsticks a session already at its cap. The trailing clause states
+    // the precedence the approved plan specifies (the drop rule outranks budget and thresholds)
+    // WITHOUT naming the drop rule, which is emitted only on a re-review: with no priors there is
+    // nothing to preserve, so the clause is vacuous rather than a dangling reference.
+    "- A point that was blocking UNDER FINDINGS ROUTING when you first raised it STAYS blocking — the halfway and at-cap rules below never demote one. That applies to THOSE rules only: it does not preserve a prior that FINDINGS ROUTING no longer admits as blocking.",
   ];
   // Strictly past the halfway point (2n > m), so a 1-of-2 review is not already "late" — AND never
   // on a first-ever review. `n > 1` is load-bearing, not belt-and-braces: both caps are UI-settable
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -1396,11 +1396,14 @@ function planSteerText(findings: string[]): string {
     // grows the surface the next round attacks. Observed: a plan reaching 32 KB with 28 success
     // criteria, three "why not X" essay sections and a ledger of addressed findings — written for
     // the reviewer rather than for the implementer.
+    // NB: no reference to the reviewer's non-blocking `Suggestions` section here. Only
+    // `gate.findings` is steered (steerFindings) — the verdict body stays in the gate for the
+    // operator's UI and never reaches the planner's worktree, so naming a section the agent cannot
+    // see would just be noise. The list above IS the complete set of what it must address.
     "\n\nRevise IN PLACE and keep the plan tight: edit the sections that are wrong and delete what no " +
     'longer holds. Do NOT append justification or rebuttal sections, a "why not X" essay, or a ledger ' +
     "of which finding each change answers — the plan is an instruction to the implementer, not a reply " +
-    "to the reviewer. Any `Suggestions (non-blocking):` section in the review is OPTIONAL: it is not " +
-    "part of this list and does not block approval." +
+    "to the reviewer." +
     "\n\nDon't start implementing yet — wait for the plan to be approved."
   );
 }
--- a/src/plan-gate.ts
+++ b/src/plan-gate.ts
@@ -28,6 +28,9 @@ import { resolveAuxSpawn, type MembraneSeams } from "./spawn-membrane";
 import { fenceUntrusted } from "./untrusted";
 import { type OperatorLanguage } from "./operator-language";
 import { resumeThenSteer } from "./resume-then-steer";
+// The ROUND block + effective-round arithmetic are shared verbatim with the PR critic (#1948) so the
+// two review loops can never drift on them — the same reason `scopeAndOutputTail` is shared there.
+import { roundBlock, effectiveRound } from "./critic-core";
 
 /** Outcome of an on-demand `consider()`: a reviewer actually spawned (`"started"`, or
  *  `"started-at-cap"` — see below); the request was a no-op (`"skipped"` — not planning, a review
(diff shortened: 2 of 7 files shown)


scoop:
## What I verified

Read the whole diff (`src/critic-core.ts`, `src/plan-gate.ts`, `src/review.ts` + the four test files) and traced every identifier and every behavioural claim the new comments make.

**Identifier / wiring resolution (all confirmed present and consistently spelled):**
- `roundBlock` + `effectiveRound` are exported at `src/critic-core.ts:493` and `src/critic-core.ts:534`, imported by `src/plan-gate.ts:155` and `src/review.ts:23`. No import cycle: `critic-core.ts:8-16` imports only `codex-last-message`, `instrument`, `types`, `usage`, `json-tolerant`, `untrusted` — never `plan-gate`/`review`.
- `store.listReviewerSpawns()` exists at `src/store.ts:3420` and is already declared on both services' store picks (`src/review.ts:155`, `src/plan-gate.ts:440`), so `countReviewSpawns`/`countPlanGateSpawns` add no plumbing.
- The `kind` discriminants are correct: `"plan_gate"` is written at `src/plan-gate.ts:806`, `"review"` at `src/review.ts:517`; the union is `src/types.ts:834`. `standalone-critic.ts:538-541` also writes `kind:"review"` but with a synthetic `taskSessionId` of `pr:<repo>#<n>` (`src/standalone-critic.ts:540`), so it can never collide with a managed session id — `countReviewSpawns`'s filter is safe.
- Ordering is right on both sides: the prompt is built (`src/plan-gate.ts:744`, `src/review.ts:435`) **before** `recordReviewerSpawn` (`src/plan-gate.ts:803`, `src/review.ts:514`), so the count is of PRIOR spawns and a first review yields `E = 1`, matching the tests at `test/plan-gate-service.test.ts:257` and `test/review.test.ts:401`.
- No consumer parses the non-blocking section headers — grepping `non-blocking):` finds only `src/critic-core.ts`, `src/plan-gate.ts`, `src/review.ts`, the tests and one research doc. So the German operator-language branch (`src/plan-gate.ts:246-254`), which will translate the body, breaks nothing. Reviewer prompts are not i18n'd and `ui/messages/*.json` is untouched, so there is no locale-parity surface in this diff.

**Behavioural claims in the new comments, checked against the code they cite:**
- "`applyChangesRequested` HOLDS the round at the cap" — true, `src/plan-gate.ts:1154`; and `startedStatus` still starts a review at the cap, `src/plan-gate.ts:1386-1387`. So the clamp in `effectiveRound` is genuinely needed.
- "`resume()` writes `round: 0`" — true, `src/plan-gate.ts:1325`; `dismiss()` likewise, `src/plan-gate.ts:1342`.
- "`forceReview` resets `addressRound`" — true, `src/review.ts:586`.
- "`runAutoAddress` fires on non-empty findings regardless of decision" — true, it is called unconditionally at `src/review.ts:1110`, and only `findings.length === 0` short-circuits it (`src/review.ts:1142`).
- "at `priorRound === cap - 1` the steer is still delivered" — true (`src/plan-gate.ts:1145`), so the hedged `may be the LAST` wording at `src/critic-core.ts:515` is the correct choice; asserting escalation would have been false.
- Byte-identity is preserved: `roundBlock` is pushed in `reviewPrompt`'s own preamble (`src/critic-core.ts:214`), not in `scopeAndOutputTail`, so the shared-tail test at `test/critic-core.test.ts:240-250` still compares equal tails, and `prReviewPrompt` (`src/critic-core.ts:242`) never gets a round block. `roundBlock` returns `[]` when either arg is absent (`src/critic-core.ts:499`), so every non-service caller is unchanged.
- The old mandate is gone: `'A non-blocking nit STILL goes in "findings"'` (previously at `critic-core.ts` line 562 of the base commit) is replaced at `src/critic-core.ts:661`, and `"best reasonable path"` is gone from `src/plan-gate.ts:244`.

## Blocking finding

**`src/review.ts:690` — `countReviewSpawns` counts reviews, not rework rounds, and never resets.**

The plan-gate side is sound: a `plan_gate` spawn only happens when the plan text changed (`src/plan-gate.ts:628`) while the gate is unapproved, so spawn count ≈ rework round.

The PR-critic side is not. `consider()` spawns a critic on **every new CI-green head** — `src/review.ts:300` skips only when the head is already reviewed, and the `2*cap` ceiling at `src/review.ts:326` applies only while `prior.findings.length > 0`. The file's own comment says so: *"consider() would spawn a critic on every new CI-green head forever"* (`src/review.ts:301-302`). Meanwhile `addressRound` **legitimately resets to 0** on every clean verdict (`src/review.ts:1142`).

So `min(C, cap)` in `effectiveRound` (`src/critic-core.ts:536`) is not a floor that only defends against operator resets — it becomes the dominant term for any session that pushes more than a couple of times. With the shipped default `PR_REVIEW_CYCLES_DEFAULT = 3` (`src/config.ts:128`):

| head | prior spawns C | `addressRound` | E | round block emitted |
| ---- | -------------- | -------------- | - | ------------------- |
| 1 | 0 | 0 | 1 | — |
| 2 | 1 | 0 (clean) | 1 | — |
| 3 | 2 | 0 (clean) | 2 | `2*2 > 3` → *"past the halfway point… raise a NEW finding only if it would make the change fail outright"* (`src/critic-core.ts:485`) |
| 4+ | ≥3 | 0 (clean) | 3 | also *"The budget is spent…"* (`src/critic-core.ts:515`) |

So the **third** review of a session that has never once entered a rework loop is briefed as past the halfway point, and from the fourth head on **every** review for the rest of the session's life is briefed as at-cap. The same happens to the very common shape "critic converged, agent kept working": a clean verdict zeroes `addressRound` but `C` keeps climbing, so the muzzle is permanent and irreversible.

The bullet at `src/critic-core.ts:479` protects correctness and security defects, but the FINDINGS ROUTING list at `src/critic-core.ts:648` also makes *a broken or violated contract*, *a missing locale/catalog counterpart* and *a diff that does not do what the task requires* blocking — none of which obviously "would make the change fail outright". Those get routed to `Nits (non-blocking):` on a session whose only sin was pushing four times. That is the quality regression the task explicitly asked not to incur, on the default configuration, on the common path — not an edge case.

The approved plan anticipated the error-spawn inflation and accepted it, and anticipated `resume`/`forceReview` resets; it did not anticipate that a PR-critic spawn is per-head rather than per-round. A targeted fix that keeps the plan's intent: apply the `C` floor only while a findings streak is actually outstanding (`prior.findings.length > 0` — which holds in every reset case the plan cares about, since `resume`/`dismiss`/`forceReview` all preserve `findings`, see `src/plan-gate.ts:1325` and `src/review.ts:573-574`), or count only spawns since the last clean verdict. Either keeps `E` reset-proof for the stalled sessions this exists for without taxing a healthy one.

## Nits (non-blocking):

- `src/plan-gate.ts:1402` — `planSteerText` tells the planning agent *"Any `Suggestions (non-blocking):` section in the review is OPTIONAL"*, but only `gate.findings` is ever steered (`src/plan-gate.ts:1116`, `src/plan-gate.ts:1148`) and the verdict body is written into the reviewer's own detached worktree, never the session's. The planner cannot see that section, so the sentence points at text it will never have. (The mirror sentence in `src/review.ts:64` *is* apt — the PR critic's body really is posted to the PR, `src/review.ts:1102-1105`.)
- `src/critic-core.ts:485` — the halfway rule routes everything to `Nits (non-blocking):`, but the session critic has four non-blocking sections (scope-creep, smells, latent, nits). A late-round latent defect now has two plausible homes.
- `src/plan-gate.ts:366` vs `src/plan-gate.ts:230` — the LOCATION-REFERENCES exemption ("a re-raised prior finding … must not be dropped, reworded or reclassified") is emitted *after* the new severity-drop EXCEPTION and only qualifies itself with "on those grounds". Readable, but the two rules sit adjacent and pull opposite ways.

## Stated limitations

- I did not run `bun test` / `bun run lint` — CI owns those per the brief. Test correctness was reviewed by reading, not execution.
- Success criterion 12 (the 30-day post-ship SQL re-measurement against `~/.shepherd/shepherd.db`) is by construction unverifiable at review time; no instrumentation is claimed or added, which matches the plan.
- I could not measure the real-world distribution of heads-per-session, so the table above is derived from the code paths (`src/review.ts:300`, `src/review.ts:1142`) rather than from production data.

scoop (commented):
## Verdict

No blocking defects. Both points from the previous revision are genuinely addressed, and I could not find a correctness, security, or contract problem in the production code. Two non-blocking items below.

---

## Prior findings — re-check

**1. `src/review.ts`: lifetime spawn count muzzles healthy sessions — ADDRESSED.**

`countReviewSpawns` is gone from the tree entirely (grepped `countReviewSpawns` across the repo: the only surviving occurrence is a comment, see finding 1). The floor is now `prior.streakReviews` (`src/review.ts:689`), which is exactly the "count only spawns since the last clean verdict" remedy the prior finding named. Verified against the counter's lifecycle:

- reset on a clean verdict: `src/review.ts:934-935` (`if (verdict.findings.length === 0) verdict.streakReviews = 0`), alongside `addressRound` (`runAutoAddress` returns 0 at `src/review.ts:1137`).
- incremented only on a findings-bearing finalized verdict: `src/review.ts:938`.
- preserved (not incremented) on error verdicts: `src/review.ts:1019`, and on spawn aborts: `src/review.ts:1235`.
- zeroed by `forceReview`: `src/review.ts:584` — so the doc claim at `src/review.ts:682` ("forceReview zeroes both") is accurate; `addressRound` is zeroed at `src/review.ts:586`.
- semantics confirmed against the field's own contract at `src/types.ts:626` and the ceiling that reads it at `src/review.ts:326` / `src/blocked.ts:144`.

Walking the exact case the prior finding described (default cap 3 — `PR_REVIEW_CYCLES_DEFAULT = 3` at `src/config.ts:128`): a session whose last verdict was clean has `addressRound = 0, streakReviews = 0`, so `effectiveRound(0, 0, 3) = 1` (`src/critic-core.ts:534-536`) — round 1, no halfway line, no budget line. The regression is closed, and `test/review.test.ts` drives both the null-prior and prior-clean shapes.

**2. `src/plan-gate.ts`: `planSteerText` referenced a `Suggestions` section the planner cannot see — ADDRESSED.**

`planSteerText` (`src/plan-gate.ts:1391-1409`) no longer mentions the section; the removal is recorded as an NB comment at `src/plan-gate.ts:1399-1402`. Verified the underlying constraint still holds: `steerFindings` passes only `gate.findings` (`src/plan-gate.ts:1115-1122`), called from `applyChangesRequested` at `src/plan-gate.ts:1148`; the verdict `body` is only ever persisted via `putPlanGate` (`src/plan-gate.ts:1165`), never delivered to the worktree.

The mirror sentence added on the PR side *is* grounded, which is the correct asymmetry: `publishVerdict` posts `verdict.body` to the PR via `forge.postReview` at `src/review.ts:1097-1100` **before** `runAutoAddress` steers at `src/review.ts:1105`, so "the posted review" in `steerText` (`src/review.ts:64`) names something the agent can actually read.

---

## Verification of the new code

**Identifiers resolve.** `roundBlock` and `effectiveRound` are exported at `src/critic-core.ts:492` and `src/critic-core.ts:534`, imported in `src/plan-gate.ts:33` and `src/review.ts:23`. No import cycle: `critic-core.ts`'s imports (lines 8-16) are `codex-last-message`, `node:child_process`, `node:util`, `instrument`, `types`, `usage`, `json-tolerant`, `untrusted` — it does not import `plan-gate`. `listReviewerSpawns` is already in the plan-gate deps pick (`src/plan-gate.ts:440`) and on the store (`src/store.ts:3420`), so `countPlanGateSpawns` is a filter, not new plumbing. `finalizeErrorVerdict` (cited in the new doc at `src/review.ts:680`) exists at `src/review.ts:996`.

**`effectiveRound` arithmetic is right at both ends.** Clamp: `applyChangesRequested` holds `gate.round` at the cap (`src/plan-gate.ts:1154`) while `startedStatus` still returns `"started-at-cap"` (`src/plan-gate.ts:1387`), so the unclamped `+1` really would have emitted "round 13 of at most 12" — `Math.min(priorRound + 1, cap)` at `src/critic-core.ts:535` fixes it. Floor: `resume()`/`dismiss()` write `round: 0` (`src/plan-gate.ts:1325`, `src/plan-gate.ts:1342`), and the at-cap operator hold is offered with `options: []` (`src/blocked.ts:134`), so the comment's "Resume is the only escape" premise checks out. `Math.min(reviewCount, cap)` means the floor can never push past the cap.

**The plan-side spawn count does not repeat the PR-side mistake.** I checked whether a lifetime `plan_gate` count could over-count the way a lifetime `review` count did. It cannot in the same way: `consider()` short-circuits unless `planPhase === "planning"` (`src/plan-gate.ts:612`), and `setPlanPhase` is called from exactly one site with exactly one value — `"executing"` (`src/service.ts:4491`, the only caller of `src/store.ts:3294`). A session never re-enters the planning phase, so the lifetime count is the count for the one planning streak. Ordering is also correct: `composeReviewerPrompt` runs at `src/plan-gate.ts:744`, `recordReviewerSpawn` at `src/plan-gate.ts:803`, so the current spawn is excluded and a first review reads `C = 0 → E = 1`.

**No prompt-section name is machine-parsed.** Grepped `non-blocking):` across the repo: it appears only in `src/critic-core.ts`, `src/plan-gate.ts`, `src/review.ts`, their four test files, and one research doc. Nothing in `ui/` or the server parses these headers, so adding `Nits (non-blocking):` / `Suggestions (non-blocking):` breaks no consumer.

**Byte-identity guards hold.** `roundBlock` returns `[]` for absent/invalid input (`src/critic-core.ts:499-500`), and it is pushed in `reviewPrompt`'s own preamble (`src/critic-core.ts:214`), not in `scopeAndOutputTail` (`src/critic-core.ts:539`), so the shared tail stays identical between `reviewPrompt` and `prReviewPrompt`. `roundBlock` interpolates only numbers and call-site literals — no untrusted data reaches it.

**No contradictions left in the prompts.** Grepped for the old mandate: `"STILL goes in"` and `"blocking or not"` now survive only inside explanatory comments (`src/critic-core.ts:197`, `src/critic-core.ts:642`), never in emitted prompt text. The plan-side drop EXCEPTION (`src/plan-gate.ts:230`) does not collide with the location-block exemption at `src/plan-gate.ts:366`, which forbids dropping a re-raised prior on *location* grounds only — a different predicate.

**Cross-checked the intended behaviour change against its consumers.** Routing nits out of `findings` makes a nit-only review read as clean to `src/signoff.ts:39`, `src/learnings-lifecycle.ts:123` and the autopilot gate noted at `src/autopilot.ts:863`. That is the stated intent of the task and is called out in the plan's Risks; the blocking definition (`src/critic-core.ts:648`) keeps correctness, security, contract and locale-parity defects in `findings` "REGARDLESS of how small it looks", so I do not read this as a defect.

### Observation (not a finding)

With the shipped default `prReviewCyclesCap = 3` (`src/config.ts:128`), the halfway rule `2 * n > m` (`src/critic-core.ts:513`) engages at round 2, so two of the three rework rounds run in restricted mode; the operator's configured 8 gives the gentler round-5 onset the design assumes. Nothing to change here — flagging it only because the effect on a default-config repo is stronger than the evidence table (which was gathered at cap 8/12) would suggest, and criterion 12's post-ship check is the right place to notice it.

### Could not verify

Nothing external. Every symbol, config key and counter this diff depends on lives in this repo and is cited above.

---

## Findings

See the two items in `findings` — both are documentation/formatting only and neither changes behaviour.

scoop:
## Re-review: prior points

**1. `src/plan-gate.ts` dangling doc ref to `countReviewSpawns` — FIXED.** The rationale is now inlined at `src/plan-gate.ts:644-647` ("Error spawns count too: a row exists whether or not the reviewer produced a verdict … per `roundBlock` lateness never downgrades a blocking finding"), and the doc no longer defers to a deleted symbol. Verified by grepping the whole tree for `countReviewSpawns`: zero occurrences remain in any `.ts` file.

**2. `src/critic-core.ts` missing `""` before the `FINDINGS ROUTING` header — FIXED.** The separator is present at `src/critic-core.ts:641`, between the last LATENT-DEFECT LENS bullet (`src/critic-core.ts:640`) and the header (`src/critic-core.ts:648`), matching the pattern used by VERIFY (`:559`), CANNOT-VERIFY (`:567`), ATTRIBUTION (`:571`), the scope-creep block (`:600`) and the smells block (`:628`).

## Verification performed

- **No import cycle.** `src/plan-gate.ts:156` imports `{ roundBlock, effectiveRound }` from `./critic-core`. `critic-core.ts`'s own imports (`src/critic-core.ts:8-16`: codex-last-message, instrument, types, usage, json-tolerant, untrusted) contain no import of `plan-gate` — grepped each; the only `plan-gate` hits are prose in `src/instrument.ts:112` and `src/types.ts:248`.
- **Both symbols exist and are exported.** `export function roundBlock` at `src/critic-core.ts:493`, `export function effectiveRound` at `src/critic-core.ts:521+` (body at `:534-537`). `src/review.ts:23` imports `effectiveRound` from the same module.
- **Section-name consistency.** `roundBlock`'s `nonBlockingSection` argument matches the header the same prompt actually emits: plan side passes `"Suggestions (non-blocking):"` (`src/plan-gate.ts:239`) against the header emitted at `src/plan-gate.ts:285`; PR side passes `"Nits (non-blocking):"` (`src/critic-core.ts:214`) against the header emitted at `src/critic-core.ts:650`.
- **Nothing parses those body headers server-side.** Grepped `non-blocking):` across `src/` and `ui/src/`: the only hits outside the two prompt builders are the steer text at `src/review.ts:61-64`. So the new sections carry no parsing contract, and the German-output instruction at `src/plan-gate.ts:246-253` cannot break anything by translating a header.
- **`briefedRound`'s stated invariants hold.** `forceReview` does zero both counters — `addressRound: 0` and `streakReviews: 0` at `src/review.ts:584-586` — so the doc claim at `src/review.ts:682-684` is accurate. `finalizeErrorVerdict` preserves the streak (`src/review.ts:1019`) and holds `addressRound` (`src/review.ts:1013`), matching `src/review.ts:680`. A clean verdict zeroes both (`src/review.ts:935`), so the healthy-session regression the previous commit fixed stays fixed.
- **Plan-gate at-cap hold is real.** `applyChangesRequested` holds `gate.round` at the cap (`src/plan-gate.ts:1158`) while `startedStatus` still returns `"started-at-cap"` (`src/plan-gate.ts:1390-1391`), and `resume()`/`dismiss()` write `round: 0` (`src/plan-gate.ts:1329`, `:1346`) — so both ends of the `effectiveRound` justification check out.
- **`countPlanGateSpawns` cannot inflate across a second planning phase.** `planPhase` is only ever set to `"planning"` at session creation (`src/service.ts:3123`) and thereafter only flipped to `"executing"` (`src/service.ts:4486-4492`); there is no path back, so the lifetime `plan_gate` count is one gate's worth of spawns.
- **Ordering is right for the count.** `composeReviewerPrompt` runs at `src/plan-gate.ts:748`, before `recordReviewerSpawn` at `src/plan-gate.ts:807`, so `C` excludes the in-flight spawn and a first review reports round 1, not 2.
- **Referenced identifiers in the new comments all resolve:** `REVIEWER_SPAWN_RETENTION_MS` (`src/config.ts:997`, 90 days per `src/config.ts:996`), `pruneReviewerSpawns` (`src/store.ts:3428`), `scopeFindings` (`src/critic-core.ts:869`), `MAX_ADDS_PER_RUN` (`src/distiller.ts:42`, enforced `:437`), `steerFindings` (`src/plan-gate.ts:1119`), `blocked.ts` hold (`src/blocked.ts:144`), `src/signoff.ts` and `src/autopilot.ts` both exist.
- **No stale byte-equality assertion left behind.** Grepped `test/` for `toBe(planReviewPrompt` / `toEqual(planReviewPrompt`: none. The two `reviewPrompt` equality assertions in `test/review.test.ts` were updated in this diff (`:355`, `:368`). The plan-gate service harness already stubs `listReviewerSpawns: () => []` (`test/plan-gate-service.test.ts:51`), so the new store read cannot throw in existing tests.
- **Old contract strings are gone from the prompts.** Grepped the tree: `"A non-blocking nit STILL goes in findings"` and `"best reasonable path"` survive only inside explanatory comments (`src/critic-core.ts:197`, `src/plan-gate.ts:193`, `:228`), negative test assertions, and a historical design doc under `docs/superpowers/plans/`.

## The finding

`src/critic-core.ts:508-512` — the halfway gate is `if (2 * n > m)`. The comment on `:508` states the intent: *"Strictly past the halfway point (2n > m), so a 1-of-2 review is not already 'late'."* That holds for `m === 2` (`2*1 > 2` is false), but not for `m === 1`: `2*1 > 1` is **true**, so a round-1-of-1 review is briefed as past the halfway point.

`cap === 1` is a supported, operator-settable value, not a hypothetical:
- `PR_REVIEW_CYCLES_MIN = 1` (`src/config.ts:62`) and `PLAN_REVIEW_CYCLES_MIN = 1` (`src/config.ts:64`).
- Both are clamped to that floor on write: `src/server.ts:4921` and `src/server.ts:4929-4937`.
- Both minima are published to the UI (`src/server.ts:4716`, `:4718`) and used as the input's `min` (`ui/src/lib/components/settings/SettingsSessionPanel.svelte:462`, values wired at `:198-199` and `:221-222`).

With `cap: 1`, `effectiveRound(0, 0, 1)` returns `1` (`src/critic-core.ts:534-537`), and `reviewPrompt`/`planReviewPrompt` therefore emit, on the very first review of a PR or plan that nobody has looked at yet:

```
ROUND — this is rework round 1 of at most 1.
- You are past the halfway point of the rework budget. Raise a NEW finding only if it would make the change fail outright; anything else you notice now goes to `Nits (non-blocking):`.
- The budget is spent. …
```

Every finding on a first review is by definition a NEW finding, so the halfway bullet suppresses the entire review down to outright-failure defects — including the classes `FINDINGS ROUTING` explicitly calls blocking (`src/critic-core.ts:649`): a broken contract, a missing locale/catalog counterpart, a diff that does not satisfy the task. The `n >= m` bullet is fine and accurate at `cap: 1` (round 1 genuinely is the last); it is the halfway bullet that should never be reachable on a first-ever review. Guarding it on `n > 1` (or equivalently `m >= 3`) preserves the existing behaviour at every cap ≥ 2 and closes the hole.

- `src/critic-core.ts:500` — the `!Number.isFinite(round) || !Number.isFinite(cap)` leg guards inputs neither production caller can produce: both come from `effectiveRound` over integer round counters and a `clampCap`-bounded cap (`src/config.ts:140-143`), so only the direct unit test at `test/critic-core.test.ts:104` reaches it.
- `src/critic-core.ts:501-502` — `const n = round; const m = cap;` are pure aliases; the four uses below could read the parameters directly.

- `docs/superpowers/plans/2026-06-08-pre-execution-plan-gate.md:272` still quotes the old `"genuinely the best reasonable path"` approval line as if it were live prompt text; it is a historical design record, so this is informational only.

scoop (commented):
## Re-review of the previous round's point

**1. `roundBlock`'s halfway gate fires on a first-ever review at `cap === 1` — FIXED.**

The gate is now `if (n > 1 && 2 * n > m)` (`src/critic-core.ts:513`), with the rationale inlined at `src/critic-core.ts:508-512`. Verified the boundary is load-bearing rather than decorative: `PR_REVIEW_CYCLES_MIN = 1` / `PLAN_REVIEW_CYCLES_MIN = 1` (`src/config.ts:62,64`) are wired to the settings number inputs' `min` attributes at `ui/src/lib/components/settings/SettingsSessionPanel.svelte:471` and `:495`, so cap 1 is genuinely operator-reachable. The `n >= m` bullet was correctly left alone (`src/critic-core.ts:517-520`), so a cap-1 first review still gets "the budget is spent" — which is true — without the halfway muzzle. A regression test covers caps `[1, 2, 3, 8, 12]` and asserts the round-1 prompt never contains "past the halfway point" (`test/critic-core.test.ts`, "the halfway rule never fires on a first-ever review, at any supported cap"), plus the converse at `n = 2` for caps 2/3 and its absence at cap 4.

## What I verified (ground truth cited)

- **`effectiveRound` arithmetic against both counters it reads.** Clamp is needed because `applyChangesRequested` holds the round at the cap (`src/plan-gate.ts:1158`) while `startedStatus` still returns `"started-at-cap"` and a review starts (`src/plan-gate.ts:874,1390-1392`); the floor is needed because `resume()` and `dismiss()` both write `round: 0` (`src/plan-gate.ts:1329,1346`) and `blocked.ts` offers that hold with `options: []` (`src/blocked.ts:134`). Both doc claims are accurate.
- **Critic-side floor is `streakReviews`, not a lifetime spawn count.** Confirmed the counter resets to 0 on a clean verdict (`src/review.ts:934-936`), increments only on a findings-bearing finalized verdict (`src/review.ts:938`), is preserved across an error verdict while `addressRound` is held (`src/review.ts:1013,1019`), and is zeroed by `forceReview` alongside `addressRound` (`src/review.ts:583-587`) and by `clearStallState` (`src/review.ts:541-545`). So `briefedRound` (`src/review.ts:689`) reads round 1 on a healthy session — the regression the third commit fixed, and the reason a lifetime count would have been wrong here.
- **Plan-side lifetime count is safe from that same inflation.** `planPhase` is only ever set to `"planning"` at session creation (`src/service.ts:3123`) and only transitions forward to `"executing"` (`src/service.ts:4486-4492`); `consider()` refuses non-planning sessions and already-approved gates and dedups an unchanged plan hash (`src/plan-gate.ts:612,615,628`). A session therefore has exactly one plan-review streak, so `countPlanGateSpawns` (`src/plan-gate.ts:648-653`) cannot climb with unrelated activity the way a critic spawn count does.
- **`kind` filter spelling.** `"plan_gate"` matches the union at `src/store.ts:3367` (`"review" | "plan_gate" | "recap" | "rundown" | "doc_agent"`); `listReviewerSpawns` is already in `PlanGateServiceDeps`' store `Pick` (`src/plan-gate.ts:440`), so this adds no plumbing. `REVIEWER_SPAWN_RETENTION_MS` really is 90 days (`src/config.ts:996-997`), so the doc note at `src/plan-gate.ts:640-642` is accurate.
- **Ordering: `countPlanGateSpawns` runs before the row is written.** `composeReviewerPrompt` is called at `src/plan-gate.ts:748`, `recordReviewerSpawn` at `src/plan-gate.ts:807`, so `C` excludes the current run. Harmless: on a normal streak `min(prior.round + 1, cap)` already wins, and on a reset streak the count is at most one low, which the cap clamp absorbs.
- **No import cycle.** `src/plan-gate.ts:33` imports from `./critic-core`, whose own imports (`src/critic-core.ts:8-16`) are `codex-last-message`, `node:child_process`, `node:util`, `instrument`, `types`, `usage`, `json-tolerant`, `untrusted` — none reach back to `plan-gate`.
- **Callers still agree.** `reviewPrompt`'s only production caller is `src/review.ts:725` and `planReviewPrompt`'s only one is `src/plan-gate.ts:681`; `criticArgv` is private with a single call site (`src/review.ts:435`). Both `opts` bags are additive, and the `roundBlock`/`findingsRoutingBlock` split keeps the round block out of `scopeAndOutputTail`, so the shared-tail identity assertion (`test/critic-core.test.ts:240-250`, compared from `"When done, write your verdict"`) still holds.
- **The steer's reference to the posted review is real.** `steerText`'s new line points the agent at a `Nits (non-blocking):` section in the posted review; `verdict.body` is what gets posted (`src/review.ts:1097-1100`), so that section is visible to the author. The plan-gate steer correctly does *not* mention `Suggestions (non-blocking):` — `steerFindings` sends only `gate.findings` (`src/plan-gate.ts:1119-1121,1152`) and the body never leaves the gate row.
- **Nothing parses the new section headings.** Grepped for consumers of the existing exact headings (`Scope creep / gold-plating (non-blocking):`, `Latent / future-reachable (non-blocking):`, `Out of scope (pre-existing…`): the only matches outside `src/critic-core.ts` / `src/plan-gate.ts` / their tests are unrelated uses of the word "non-blocking" (`src/forge/github.ts:713`, `src/branch-pruner.ts:39`, `ui/src/routes/+page.svelte:799`, `ui/src/lib/components/new-task/EngineCapacityLine.svelte:42`). So moving nits into the body couples to no parser.
- **No i18n or catalog obligation.** The diff adds no message keys and touches no `ui/` file; the feature-catalog gate only fires for `feat(...)` commits touching `ui/src/lib/components/**` or `ui/src/routes/**` (`CLAUDE.md:73`), and all four commits here are `fix(review): …`. The plan-gate prompt's German instruction still lands last (`src/plan-gate.ts:246-253`), after the new blocks, so it still governs `summary`/`body`/`findings`.
- **No new "clean" false positive in the verdict parser.** A `request-changes` verdict emptied of findings still falls back to its summary rather than reading as clean (`src/critic-core.ts:1016-1017`), so `findings` becoming blocking-only does not open a hole in `autopilot`/`signoff`'s zero-findings check.

## The one thing to address

The stale-prior **drop** rule and `roundBlock`'s **"stays blocking"** bullet are, as written, mutually exclusive over exactly the population the drop rule exists to clear — and the precedence that resolves them is stated in neither prompt. Detail in the finding below.

## Observations (verified, not findings)

- **The plan-side floor counts spawns that never spent a rework round.** `countPlanGateSpawns` counts every `plan_gate` row (`src/plan-gate.ts:650-651`), but an `error` gate carries the round unchanged (`src/plan-gate.ts:1271`, `applyError` at `:1214-1223`) and a `herdr.start` failure records *and* completes a row before returning `"error-spawn"` (`src/plan-gate.ts:807-845`). Unlike `ReviewService`, `PlanGate` has no error counter at all — there is no `errorRound` field on the interface (`src/types.ts:574-611`) — so the approved plan's justification that "consecutive errors have their own ceiling" holds for the critic (`src/review.ts:1028`) but not here. Consequence: at the default plan cap of 5 (`src/config.ts:129`), three timed-out runs plus one real round brief the reviewer as round 4 → past halfway. Not raised as a finding: the trade-off is explicitly stated in the approved plan and documented in-code at `src/plan-gate.ts:644-647`, the in-code rationale does *not* repeat the false ceiling claim, `reviewer_spawns` carries no field that distinguishes an error run from a real one (`completeReviewerSpawn` fires on both), and the invariant bullets at `src/critic-core.ts:505-506` keep correctness/security items and standing blockers blocking regardless. Worth a follow-up if plan-reviewer timeouts turn out to be common.
- **`2n > m` is aggressive at the shipped defaults.** The rule was designed against the operator's 12/8 caps, but the seeds are 3 for the PR critic and 5 for the plan gate (`src/config.ts:128-129`), so a default install muzzles new non-fail-outright findings from round 2 of 3 and round 3 of 5. That is faithful to the approved plan's `E > M/2` and defensible (the author has had one full unmuzzled round), so not a finding — just worth knowing the default is much tighter than the evidence table's caps.

- `src/critic-core.ts:500` — the `Number.isFinite` / `< 1` guard in `roundBlock` handles inputs neither production caller can produce: `effectiveRound` returns `≥ 1` whenever `cap ≥ 1`, and both caps are clamped into `[1, 8]` / `[1, 12]` before reaching the service (`src/config.ts:801-826`). Cheap and harmless defensiveness the author may legitimately keep.

scoop (commented):
## Re-review: prior finding

**1. `src/critic-core.ts` — unqualified STAYS-blocking bullet defeating the stale-prior drop rule — ADDRESSED.**
The bullet is now scoped and states the precedence explicitly at `src/critic-core.ts:514`:
> "A point that was blocking **UNDER FINDINGS ROUTING** when you first raised it STAYS blocking — the halfway and at-cap rules below never demote one. That applies to THOSE rules only: it does not preserve a prior that FINDINGS ROUTING no longer admits as blocking."

Verified it coexists with both drop rules rather than overriding them: the critic-side drop at `src/critic-core.ts:199` (`"…is likewise DROPPED"`) and the plan-side `EXCEPTION:` at `src/plan-gate.ts:230`. The `n > 1` guard added at `src/critic-core.ts:521` also means the bullet is never paired with a halfway rule on a first review. Not re-raised.

- **No import cycle from the new sharing.** `src/plan-gate.ts:33` now imports `roundBlock`/`effectiveRound` from `./critic-core`. `src/critic-core.ts:8-16` imports only `codex-last-message`, `instrument`, `types`, `usage`, `json-tolerant`, `untrusted` — none of which reach `plan-gate`. Confirmed no `plan-gate` import in `critic-core.ts`; the only importers of `plan-gate` are `src/index.ts:90` and `src/server.ts:64`.
- **Both new signatures have exactly one production caller, and both were updated.** `reviewPrompt` → `src/review.ts:725` (opts now carries `round`/`cap`); `planReviewPrompt` → `src/plan-gate.ts:681` (new trailing `opts` bag). `criticArgv`'s new `round: number` param has one caller, `src/review.ts:435`. The `opts` bag on `planReviewPrompt` is correctly appended *after* `operatorLanguage` (`src/plan-gate.ts:178`), which binds positionally — no mis-binding.
- **Byte-identity for every other caller holds by construction.** `roundBlock` returns `[]` when either `round` or `cap` is absent (`src/critic-core.ts:499-500`), and it is the only thing pushed in the round position (`src/critic-core.ts:214`, `src/plan-gate.ts:238-240`). `prReviewPrompt` (`src/critic-core.ts:242`) never passes it, so the standalone critic gets no ROUND block while still sharing the tail.
- **`effectiveRound` arithmetic against the two failure modes it names.** Clamp: `applyChangesRequested` holds the round at the cap (`src/plan-gate.ts:1158`) while `startedStatus` still starts a review (`src/plan-gate.ts:874`), so `min(priorRound+1, cap)` is required — verified. Floor: `resume()`/`dismiss()` write `round: 0` (`src/plan-gate.ts:1329`, `src/plan-gate.ts:1346`), so the spawn-count max is required — verified. `effectiveRound(0,0,cap) === 1`, so a fresh session never briefs round 0 (`src/critic-core.ts:546-549`).
- **`ReviewerSpawnRow` field names used by `countPlanGateSpawns` are correct.** `kind: "review" | "plan_gate" | …` and `taskSessionId` exist at `src/types.ts:834` / `src/types.ts:833`; `listReviewerSpawns(): ReviewerSpawnRow[]` takes no arguments (`src/store.ts:3420`), matching the argument-less call at `src/plan-gate.ts:650`. The `"plan_gate"` literal matches what is written at `src/plan-gate.ts:810`.
- **The plan-side lifetime spawn count is *not* the bug that was fixed on the critic side.** `consider()` short-circuits on an approved gate (`src/plan-gate.ts:615`, force does not bypass) and `dropPlanGate` runs only at archive (`src/plan-gate.ts:1372`), so a session's `plan_gate` spawns track review rounds rather than pushes. The critic side legitimately needed the different floor — see the finding below for the one residual gap.
- **The critic-side floor (`streakReviews`) and its documented reset claims are accurate.** `streakReviews` resets to 0 on a clean verdict and increments on any findings-bearing verdict (`src/review.ts:934-938`); an error verdict preserves it while leaving `addressRound` untouched (`src/review.ts:1013`, `src/review.ts:1019`) — which is exactly what the doc comment at `src/review.ts:672-678` claims. `forceReview` zeroes both `streakReviews` and `addressRound` (`src/review.ts:583-587`), matching the "restores full latitude" claim at `src/review.ts:682`.
- **The stale-prior drop actually takes effect.** `gate.findings` on a rework round comes from the fresh verdict (`src/plan-gate.ts:1259`, steered at `src/plan-gate.ts:1152`); nothing carries the previous round's findings forward, so a dropped prior really does disappear from the steer.
- **Emptying `findings` does not break downstream consumers.** The standalone critic already posts a review with zero findings (`src/standalone-critic.ts:588`, `src/standalone-critic.ts:618`), so inheriting the shared `Nits (non-blocking):` routing does not create an unhandled shape.
- **No i18n/catalog surface touched.** The diff is confined to `src/critic-core.ts`, `src/plan-gate.ts`, `src/review.ts` and four test files — no `ui/messages/*`. Reviewer prompts are server-only English instructions; the `operatorLanguage === "de"` output-language block at `src/plan-gate.ts:246-254` is untouched and still applies to the reviewer's `summary`/`body`/`findings`, so locale parity is not in play.
- **Cap bounds are respected at the edges.** `PR_REVIEW_CYCLES_MIN`/`PLAN_REVIEW_CYCLES_MIN` are both 1 (`src/config.ts:62-65`), which is what the `n > 1` guard at `src/critic-core.ts:521` protects; at cap 1 the at-cap line still fires, but it is factually true there (round 1 *is* the last round) and it only asks to reserve blocking findings rather than routing them away — not a defect.

### Nits (non-blocking):
- `src/plan-gate.ts:237-240`: the plan prompt emits FINDINGS ROUTING *before* the ROUND block while `src/critic-core.ts:214` emits the ROUND block *before* the tail that defines FINDINGS ROUTING — so the shared bullet is a forward reference on one side and a backward reference on the other. Harmless, but the two prompts read as one system elsewhere.

PR #836, opened 2026-06-19
feat(ui): expose 1M-context Opus & Sonnet in model pickers

scoop:
## Problem

The model picker only exposed the standard **200K-context** Opus and Sonnet — there was no way to select the **1M-context** variant, so sessions that picked Opus were stuck on limited context.

## What this does

Adds **Opus (1M context)** and **Sonnet (1M context)** as selectable options everywhere a model is chosen — the New Task picker, the global default-model Setting, and the per-repo automation override — all fed from the single `MODELS` source.

| Picker option | Spawned `--model` | Resolves to | Context |
| --- | --- | --- | --- |
| Opus (existing) | `opus` | `claude-opus-4-8` | 200K |
| **Opus (1M context)** — new | `opus[1m]` | `claude-opus-4-8` | 1M |
| Sonnet (existing) | `sonnet` | `claude-sonnet-4-6` | 200K |
| **Sonnet (1M context)** — new | `sonnet[1m]` | `claude-sonnet-4-6` | 1M |

## How it works

The `[1m]` suffix is a valid `--model` value that passes **straight through** `buildSpawnArgv` to claude — no alias→spawn mapping layer. **Empirically verified:** `claude --model 'opus[1m]'`/`'sonnet[1m]'` reply OK and add the `context-1m-2025-08-07` beta header, whereas the bare alias does **not** (so the suffix genuinely enables the 1M window, not a silent no-op). The whole spawn path is array-argv (no shell), so the brackets can't glob-expand or word-split — a regression test pins the literal token through the real argv builder.

A shared `modelLabel()` helper renders friendly i18n labels ("Opus (1M context)") across all 3 pickers + the 2 session-model display surfaces, so the raw `opus[1m]` token never surfaces to users.

## Cost guardrails

Both 1M tiers are classified **premium** — the >200K long-context regime is exactly the unattended-default cost case `PREMIUM_MODELS` warns about (so `sonnet[1m]` is premium even though plain `sonnet` isn't). Settings additionally shows a **distinct 1M-context note**, worded accurately: Opus 1M is at standard per-token pricing, so the note speaks to higher per-turn token consumption (and Sonnet's credit billing), not a fictitious per-token multiplier.

## Known limitations (acknowledged, out of scope)

- **Per-model usage merges 1M and non-1M spend.** `SessionUsage.byModel` keys on the model id Claude reports (`claude-opus-4-8`), so `opus` and `opus[1m]` share one bucket — 1M cost isn't separately visible.
[...]

PR #1957, opened 2026-07-28
fix(spawn): stop unattended critics wedging on the fullscreen-renderer dialog

scoop:
## What was wrong

The PR critic never errored — it **hung**. Claude Code renders a blocking TUI dialog:

```
Try the new fullscreen renderer? · ❯ 1. Yes, try it   2. Not now
```

A transient agent runs `--permission-mode dontAsk` with no operator, so nobody answers. The pane sat there until the 1800s hard deadline, which finalized an `error` verdict whose summary is the fixed string `"critic did not produce a verdict"`. That is the "review errors out with no explanation" an operator sees. Six reviews wedged this way across two repos (`flowagent` #836, `topmedia-website` #171), each burning the full 30 minutes.

It never self-heals, for two independent reasons — both read out of the shipped CLI's own gate (`2.1.220`):

```js
function Wym(){ …; if(eo().tui!==void 0)return!1;                          // ← the kill-switch
                if((xt().fullscreenUpsellSeenCount??0)>=jMn)return!1; … }  // jMn = 3
```

1. `fullscreenUpsellSeenCount` is written **only when the dialog is answered**. An unattended agent never answers → the counter never advances → it re-fires on every spawn, forever.
2. That counter lives in a **per-config-dir** `.claude.json`, so a fresh machine, an api-key config mirror, or any isolated profile starts over. The claude-swap profiles here have no such key at all.

**This is not what the recent spawn work (#1944/#1956) addressed** — the clamp log lines and the hang lines interleave for the same session, which is why that fix didn't help.

**Why the existing renderer env pin doesn't cover it:** `CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1` pins the classic renderer, but the gate skips only when fullscreen is *already active* — so pinning classic actively **qualifies** the spawn for the upsell.

## The fix

**Layer 1 — root cause.** `tui:"default"` in the `--settings` JSON `buildTransientAgentArgv` already builds. `tui` is a schema-validated key (`v.enum(["default","fullscreen"])`) and `eo()` is the merged effective settings, so this short-circuits the gate regardless of counter or config dir. `"default"` *is* the classic renderer, so it asserts the posture the env pin already sets. One change covers every transient agent (PR critic, plan reviewer, doc, namer, recap, autopilot, rundown, distiller, optimizer). Same class of fix, same file, as the existing `enableAllProjectMcpServers` entry.

**Layer 2 — fail fast.** `tick()` now classifies the critic's **raw** `readAsync(id,"visible")` buffer and finalizes early when it is blocked *and* byte-identical to the previous read (~30s vs 30min). The unchanged-buffer requirement is load-bearing: critics routinely print findings as `1. … 2. …`, and motion — not shape — is the only honest discriminator.

**Layer 3 — explain it.** `ReviewVerdict` gains a `summaryCode` sentinel (mirroring `PlanGate.summaryCode`) carrying `blocked` / `timeout` / `exited` / `unparseable`, rendered per-locale. One pure `noVerdictCause()` feeds both the log and the badge so they cannot disagree.

## Verification

A HALTing spike ran first, since the whole fix rests on one claim no unit test can assert.

*(The plan specified `CLAUDE_CODE_FORCE_FULLSCREEN_UPSELL=1`; on reading the gate that check short-circuits to `true` **before** the `tui` test, so it cannot discriminate. Used the renderer resolver — same `eo().tui` read — instead.)*

- With renderer env stripped: `--settings {"tui":"fullscreen"}` → alt-screen entered (`ESC[?1049h`); `{"tui":"default"}` → not entered. `--settings` demonstrably populates `eo().tui`.
- **End-to-end against the real argv**, with a config dir whose counter was stripped to reproduce the wedge: control (argv minus `tui`) → dialog present, run hung to timeout. Fixed argv → **0 dialog hits**, agent ran and answered.

Gates: root `bun test` (7899, 0 fail), `lint`, `tsc --noEmit`; ui `check` (0 errors), `test` (4358), `check:i18n` (parity); `fallow audit` clean on changed files.

Audited every `ReviewVerdict.summary` consumer for the new `""` case: `attachReviewPush` and the critic signal payload are `changes_requested`/`commented`-only, `publishSpawnAbort` builds its own verdict and keeps its prose, and `autopilot.ts`'s `v.summary` is a different type. None break.

## Self-review caught a real bug

`buildVerdict` initially keyed the cause off the *resolved* decision — but `buildVerdictCore` also yields `error` for a verdict the critic **did** write with an unrecognized `decision` field. That would have blanked the critic's real summary and mislabelled a genuine review as a spawn failure. Now gated on `raw === null`, classified once, with a regression test that fails against the old logic.

## Scope

Deliberately **not** included: main-session spawns (they degrade gracefully — the blocked-classifier surfaces the dialog as a BLOCKED session a human can answer); fail-fast in `plan-gate.ts` / `standalone-critic.ts` (Layer 1 closes the actual dialog there; the helper is exported pure for a follow-up); the other counted upsells (`passesUpsellSeenCount`, `remoteControlUpsellSeenCount`) — no shared kill-switch exists, and Layer 2 is the net for them.

`fix(...)`, not `feat` — this restores advertised behaviour rather than shipping a new capability, so no feature-catalog entry. [no-feature-entry]

```shepherd:manual-steps
- [ ] POST-MERGE: deploy (`bun run update`), then re-trigger the stuck TASK-1971 review on PR #836 — its last verdict is a wedged-critic error and will not re-run on its own.
```


scoop:
Root cause (Layer 1), fail-fast (Layer 2) and the persistence/badge plumbing (Layer 3) all check out against the code. One consumer of the new `summary: ""` contract was missed, and it is the surface an operator opens to read a failed review.

## What I verified

**Layer 1 — `tui:"default"`**
- `src/transient-agent-argv.ts:244` emits `{ disableAllHooks: true, tui: "default" }` before the conditional `enableAllProjectMcpServers` (`:245`) and the `apiKeySettingsFragment()` spread, matching the byte-identity expectation at `test/plan-gate-prompt.test.ts:51` (`{"disableAllHooks":true,"tui":"default","enableAllProjectMcpServers":true}`) and `test/review.test.ts:602-606`.
- I swept every remaining `--settings` assertion in the tree (`disableAllHooks` grep): the byte-identity ones are all updated (`test/distiller.test.ts:233,275`, `test/optimizer.test.ts:344`, `test/namer-llm.test.ts:138`, `test/autopilot-llm.test.ts:204`, `test/recap.test.ts:1074`, `test/transient-agent-argv.test.ts:283,289`); `test/verify-key.test.ts:205` is a property check and needs no change.

**Layer 2 — stuck detection**
- `src/review.ts:1036` reads `readAsync(f.terminalId, "visible")` — the same shape `src/poller.ts:491` feeds `classifyBlocked`, and a valid source per the driver signature at `src/herdr.ts:732`. The doc-comment claim that a flattened buffer can never match is true: `OPTION_RE` is line-anchored (`src/blocked.ts:31`) and `classifyBlocked` splits via `tailLines` (`src/blocked.ts:37-43,79`).
- The wait branch that hosts the detector only fires while the spawn is genuinely alive: with an absent verdict `decideVerdictAction` returns `"wait"` only when `!timedOut && !(spawnFinished && graceElapsed)` (`src/json-tolerant.ts:168`), so the detector never mislabels an already-exited spawn as blocked.
- The early finalize still reaps: `finalize()`'s `finally` calls `reapRun(...)` (`src/review.ts:1126`), so a wedged pane is actually killed rather than left running.

**Layer 3 — persistence and locale**
- `putReview` is consistent: 20 columns, 20 placeholders, 20 bound params (`src/store.ts:2747-2780`), with `summaryCode=excluded.summaryCode` in the upsert so a later real verdict clears a stale code.
- The migration is guarded (`add()` checks `PRAGMA table_info`, `src/store.ts:3898-3901`) and the backfill `UPDATE` is correctly narrowed to `decision='error' AND summaryCode IS NULL AND summary='critic did not produce a verdict'` (`src/store.ts:3935-3939`) — the only writer of that literal is `src/critic-core.ts:1054`, and the standalone PR critic persists to a different table (`putPrReview`, `src/standalone-critic.ts:694`), so there is no cross-path contamination.
- Locale parity confirmed: `ui/messages/en.json:606-609` and `ui/messages/de.json:606-609` both carry all four `criticbadge_no_verdict_*` keys, and `NO_VERDICT_REASON` (`ui/src/lib/components/critic-badge.ts:5-10`) references exactly those four. `en`/`de` are the only catalogs (`ui/messages/`). The `ReviewSummaryCode` union is byte-identical in `src/types.ts:529-530` and `ui/src/lib/types.ts:700-701`, and the store's validating set (`src/store.ts:338-343`) lists the same four values.
- `summaryCode` does reach the client: `src/index.ts:1401` emits the whole verdict object on `session:review`, and `snapshotReviews()` selects the column (`src/store.ts:2819`).

**Consumer audit for `summary: ""`** — I walked every reader of `ReviewVerdict.summary`:
- `src/review.ts:1268` — signal payload, gated on `decision === "changes_requested"`, never an error verdict. Safe.
- `src/recap.ts:811` — `if (review?.summary)`, degrades to omitting the line. Safe.
- `src/push.ts:316-321` — builds its body from `decision`, never `summary`. Safe.
- `src/autopilot.ts:362,368,383,386` — that is `AutopilotVerdict`, unrelated type. Safe.
- `ui/src/lib/components/CriticBadge.svelte:102` — correctly switched to `criticTitle()`. Safe.
- `ui/src/lib/components/GitRail.svelte:822-823` — **not safe**; see the finding.

## Finding detail

`buildVerdict` sets `summary: cause ? "" : core.summary` (`src/review.ts:1336`) and `body` is `""` for a null raw (`src/critic-core.ts:1065`), so a no-verdict `error` verdict now carries empty `summary`, empty `body` and empty `findings`. `GitRail.svelte:822` renders the summary behind `{#if verdict.summary}` and `:825` renders the body behind `{#if verdict.body}` — both now false. The REVIEW ERR chip is still a button that opens that popover (`ui/src/lib/components/git-rail/RailStatusActions.svelte:285-294`, reached for `decision: "error"` via `criticBadgeLabel`'s default arm), and its own tooltip is the generic `m.gitrail_review_title()`, so clicking it yields a dialog containing only the badge label, the PR link and a close button. Before this PR it showed "critic did not produce a verdict". The store migration blanks legacy rows the same way (`src/store.ts:3935-3939`), so already-persisted REVIEW ERR rows regress too.

The plan-gate contract this change says it mirrors already handles exactly this: `ui/src/lib/components/PlanPanel.svelte:437-438` renders `{#if gate.summaryCode || gate.summary}` → `gate.summaryCode ? m.planpanel_no_verdict() : gate.summary`. The review path needs the equivalent (`criticTitle(verdict)` is already exported and does the fallback chain correctly).

## Nits (non-blocking):
- `src/review.ts:930-946`: two JSDoc blocks are now stacked directly above `diagnoseNoVerdict` — the first one ("Record WHY a run is about to finalize as an `error` verdict… Never throws") describes `logNoVerdict`, which is left with no doc comment at `:958`.
- `src/review.ts:78`: `noVerdictCause`'s header says the wedged pane "is checked FIRST", but the `unparseable` branch precedes it at `:92` (and `test/review.test.ts:872` asserts unparseable wins) — it is first only relative to the timeout branch.
- `src/review.ts:225-226`: the `ReviewServiceDeps.herdr` comment still says `readAsync` is "diagnostics-only (the no-verdict pane tail)… degrades to no tail"; it is now also the fail-fast's only input, so a fake omitting it silently disables the detector rather than just losing a log field.

- `src/critic-stuck.ts:44`: the `yes-no` arm is reachable today but has no captured production sample behind it — `YES_NO_RE` (`src/blocked.ts:32`) matches `(y/n)` anywhere in the last 15 non-empty lines, so a critic whose pane happens to render diff text containing `(y/n)` and then stops advancing would be finalized as `blocked`. The run is dead either way at that point (verdict absent, spawn alive, buffer frozen), so the only cost is a mislabelled cause.

- `src/prompt-recommend.ts:140` builds its own `--settings` JSON (`{ disableAllHooks: true, ...apiKeySettingsFragment() }`) instead of going through `buildTransientAgentArgv`, so that unattended claude spawn still has no `tui` kill-switch — the plan's "one central change covers every transient claude agent" does not hold for this site.
- `src/plan-gate.ts` and `src/standalone-critic.ts` still have no fail-fast on a wedged pane (explicitly declared out of scope by the approved plan; Layer 1 closes the known dialog for them).

## Stated limitations
- I could not verify that `tui` is a schema-valid Claude Code settings key against the installed CLI: `~/.local/share/claude/versions/2.1.220` is a binary and shell inspection of it was denied in this session. It is corroborated indirectly inside this repo — `test/herdr.test.ts:227` refers to "the operator's persisted `tui` setting" and `docs/spikes/1043-fullscreen-alt-screen.md:160` proposes a "per-spawn `tui` toggle" — so I assumed the key as written rather than treating it as either verified or wrong.
- The Phase-0 empirical spike (forced-upsell two-arm repro) is not reproducible from a read-only review; I did not attempt to confirm or refute it.

scoop:
## Re-review of the prior round's point

**Prior finding 1 (GitRail popover renders nothing for a no-verdict `error`) — ADDRESSED.** `ui/src/lib/components/GitRail.svelte:825-827` now gates on `verdict.summaryCode || verdict.summary` and renders `criticTitle(verdict)`, mirroring `ui/src/lib/components/PlanPanel.svelte:437-438`. `criticTitle` (`ui/src/lib/components/critic-badge.ts:21-24`) resolves code → locale string → `summary` → generic title, and `ui/src/lib/components/GitRail.browser.test.ts:2081-2100` covers both the coded row and the legacy-prose row. `CriticBadge.svelte:102` was switched to the same helper, so the tooltip and the popover agree. Nothing to re-raise.

## What I verified

- **Layer 1 reaches the critic.** `src/review.ts:771` spawns via `buildTransientAgentArgv("reviewer", …)`, and `src/transient-agent-argv.ts:244` now emits `{ disableAllHooks: true, tui: "default", … }` unconditionally; `test/transient-agent-argv.test.ts:96-107` asserts it for every kind in both auth modes.
- **i18n parity.** The repo maintains exactly two catalogs (`ui/messages/en.json`, `ui/messages/de.json`); all four keys exist in both — `en.json:606-609` vs `de.json:606-609` — and each is referenced once in `ui/src/lib/components/critic-badge.ts:5-10` with matching spelling.
- **Persistence round-trip.** `src/store.ts:1179` (fresh DDL), `:3930` (ALTER), `:2737`/`:2747`/`:2819` (SELECT/INSERT column lists — 20 columns vs 20 `?` placeholders at `:2749`), `:2716` (explicit `coerceReviewSummaryCode` overriding the `...r` spread). The legacy backfill at `:3936-3939` matches the exact string produced at `src/critic-core.ts:1054` and mirrors the already-shipped plan-gate migration at `:3964-3967`.
- **`summary: ""` consumer audit.** The only server readers of a `ReviewVerdict.summary` are `src/review.ts:1268` (changes_requested signal — never an error verdict) and `src/recap.ts:811` (already `if (review?.summary)`-guarded). The only UI readers are `critic-badge.ts:23` and `GitRail.svelte:825`, both updated. `src/review.ts:1536` (`publishSpawnAbort`) writes an `error` verdict with real prose and no code, which `coerceReviewSummaryCode` stores as null → renders verbatim. `src/rundown-core.ts`, `src/blocked.ts:117-154`, `src/push.ts:508` read only `decision`/`findings`/`errorRound`.
- **Read shape.** `src/review.ts:1036` calls `readAsync(f.terminalId, "visible")`, matching `src/poller.ts:491`, and `src/herdr.ts:732` accepts that source; `test/review.test.ts:975-990` pins it.

## Findings

### 1. The wait-branch fail-fast throws away a verdict the gate was deliberately holding

`src/review.ts:923-925`:

```ts
if (action === "wait" && (await this.keepWaiting(f))) continue;
const raw: RawVerdict | null =
  action === "finalize-value" && read.status === "parsed" ? read.value : null;
```

When `keepWaiting()` reports a wedge it returns `false` and execution falls through — but `action` is still `"wait"`, so `raw` is unconditionally `null`. `action === "wait"` does **not** imply "no verdict exists": `src/json-tolerant.ts:161-163` returns `"wait"` for `status: "parsed", repaired: true` while the spawn is alive and un-timed-out, precisely so the value can be accepted later (`:143-147` documents this as "finalize-value only once the spawn is finished (or timed out)"). A repaired parse is the observed-in-production case of an unescaped inner quote (`src/json-tolerant.ts:5-9`, #822).

Failure scenario, all steps on paths that execute today: the critic writes a complete verdict containing one unescaped `"` → `tolerantParseJson` repairs it (`:39-42`) → the `claude` process stays alive at its idle prompt, so `isSpawnAlive` is true (`src/json-tolerant.ts:110-114`) → `decideVerdictAction` returns `"wait"` → the pane is now static, and its last 15 non-empty lines hold the critic's own `1. … 2. …` findings render, which `classifyBlocked` reports as `menu` (`src/blocked.ts:83-96`) → two identical reads 15s apart → `stuckOnPrompt` fires (`src/critic-stuck.ts:40-45`) → `raw = null` → `no-verdict-blocked` error verdict. A complete review with real findings is discarded, `errorRound` is incremented (`src/review.ts:1600`), and `seenNoteIds` is rolled back — where before it would have landed as a real verdict at the deadline. Repeat it three times and `finalizeErrorVerdict` escalates a "critic can't make progress" stall signal for a critic that produced findings every run.

`test/critic-stuck.test.ts:45-48` asserts exactly the opposite as its justification — "the verdict file is already written by then, so tick() takes the finalize-value path and never consults this detector". That holds for a strict parse only; for a repaired parse `tick()` does consult the detector. The comment is wrong and the residual it waves away is the bug.

Fix: only fail fast when there is genuinely nothing to finalize with (e.g. gate `detectStuck` / the fall-through on `read.status === "absent"`), or pass `read.value` through when `read.status === "parsed"`.

### 2. `noVerdictCause`'s doc contradicts its own precedence

`src/review.ts:121` says "a pane wedged on a prompt is checked FIRST". It is checked **second** — `read.status === "unparseable"` is tested before it (`:92-101`), and `test/review.test.ts:872-876` pins that ("unparseable wins over everything, including a wedged pane"). The rationale that follows only ever compares against the timeout branch, so the prose should say the wedge outranks the *deadline*, not that it is first.

### 3. The `herdr` dependency comment is now false

`src/review.ts:225` still reads "`readAsync` is diagnostics-only (the no-verdict pane tail) and is called defensively, so an injected fake may omit it — the read is guarded and degrades to no tail." As of `:1036` `readAsync` also drives `detectStuck`, i.e. a control-flow decision that finalizes a run early; omitting it no longer just costs a log field, it silently disables the whole fail-fast layer. The comment needs to state that.

## Nits (non-blocking):

- `src/review.ts:940-956`: two consecutive doc blocks now sit above `diagnoseNoVerdict` — the older "Record WHY a run is about to finalize…" block belongs to `logNoVerdict` (`:1003`), which is left undocumented. Reads like an editing artifact.
- `src/critic-stuck.ts:42`: the empty-pane guard is unreachable in practice (an empty string can't produce a `menu`/`yes-no` shape), though harmless as documentation of intent.

- `src/review.ts:1032`: `detectStuck` returns `null` on a `readAsync` throw *and* leaves `f.lastVisible` at its stale pre-error value, so the read after a transient herdr blip compares against a buffer two ticks old; today that can only cause an early (still-identical) match, but if the freshness window is ever widened it becomes a false-positive source.

- `src/review.ts:1060-1064`: `finalize()`'s new `cause` parameter is optional with a `?? null` default despite having exactly one call site (`:933`) that always passes it; the optionality only exists for tests.

- `src/prompt-recommend.ts:140` builds its own `--settings` JSON (`{ disableAllHooks: true, ...apiKeySettingsFragment() }`) instead of going through `buildTransientAgentArgv`, so this transient, operator-less `claude` spawn does not get the `tui:"default"` kill-switch and remains exposed to the same blocking upsell dialog.
- `src/plan-gate.ts` and `src/standalone-critic.ts` have the same wedge-until-deadline hole in their own wait branches; the plan declares this a deliberate follow-up, and Layer 1 closes the currently-known dialog for them.

## Stated limitations

- I could not verify the load-bearing external claim that `--settings` participates in the CLI's effective-settings merge for the `tui` key, nor that an older `claude` build tolerates the key without warning/erroring — the CLI lives outside this repo and outside my sandbox. The plan makes this a HALTing Phase-0 spike; assumed as written.
- I did not run tests or lint (CI's job, per instructions).

scoop (commented):
## Re-review of the three prior points

**1. Stuck-pane fall-through discarded a repairable verdict — FIXED.** The fall-through no longer forces `raw = null`. `tick()` now delegates the wait branch to `resolveWait()` (src/review.ts:930, :1019-1041), which on a wedge calls `decideVerdictAction(read, true, timedOut, true)`. Compared against src/json-tolerant.ts:161-163: a `parsed && repaired` read with `spawnFinished=true` returns `finalize-value`, so `raw = read.value` at src/review.ts:932-933, `diagnoseNoVerdict()` returns null at :969 (gated on `raw !== null`), and `buildVerdict()` keeps `core.summary` with `summaryCode: null` (:1357-1358) — findings preserved, `errorRound` untouched, `seenNoteIds` not rolled back. The unparseable/absent halves still resolve to `finalize-null` (json-tolerant.ts:165-168), so the fail-fast survives. Pinned by test/review.test.ts:1024-1053 (repaired value lands) and :1058-1080 (unparseable still errors). The contradictory test comment is gone: test/critic-stuck.test.ts:45-51 now states the re-decide contract, and its claim that "an unrepaired verdict never reaches `wait`" matches json-tolerant.ts:162.

**2. `noVerdictCause` doc claimed the wedge is checked FIRST — FIXED.** src/review.ts:77-80 now reads "an unparseable verdict wins outright … Below it, a wedged pane outranks the deadline", which matches the branch order at src/review.ts:93-108 (unparseable → stuck → timeout → exited) and the ordering pinned by test/review.test.ts:872-876.

**3. `readAsync` "diagnostics-only" dependency comment — FIXED.** src/review.ts:226-230 now states `readAsync` serves two callers, that it drives a control-flow decision, and that omitting it in a fake silently disables the fail-fast. Verified against the actual second caller at src/review.ts:1057 and the interface `Pick` at :232-234, which already includes `readAsync`.

- **Read shape.** `detectStuck` reads `readAsync(f.terminalId, "visible")` (src/review.ts:1057), the same source src/poller.ts:491 feeds `classifyBlocked`; the driver signature accepts it (src/herdr.ts:732). `classifyBlocked`'s `OPTION_RE` is line-anchored (src/blocked.ts:31) and it splits on `\n` via `tailLines` (:37-43), so the raw-buffer requirement is real, and the wiring guard at test/review.test.ts:1001-1015 asserts `[{source:"visible", lines: undefined}]` rather than the `"recent", 20` shape `readPaneTail()` uses (src/review.ts:1071).
- **Fail-fast safety.** `stuckOnPrompt` requires a byte-identical previous buffer and rejects an empty pane (src/critic-stuck.ts:41-44); `detectStuck` returns null inside `STARTUP_GRACE_MS` (src/review.ts:1054, constant defined at src/json-tolerant.ts:134) and swallows read errors (:1058-1060) — a herdr fake without `readAsync` throws a TypeError that lands in the same catch, so the degradation claim in the deps comment holds.
- **`summary: ""` consumers.** Grepped every reader of a `ReviewVerdict.summary`: src/review.ts:1289 (critic signal) is inside a `decision === "changes_requested"` branch (:1284) that a no-verdict `error` never reaches; src/recap.ts:811 is `if (review?.summary)` so it just omits the line; src/review.ts:1549's spawn-abort dedup compares against a non-empty `reason.slice(0,100)` (:1545). src/autopilot.ts:362-386 operates on `AutopilotVerdict` (src/autopilot.ts:353), not a review verdict. No other server consumer exists (rundown-core/herd-digest/session-snapshot/hold-service/drain/automerge have no `summary` reference).
- **UI.** `criticTitle` resolves code → locale → summary → generic title (ui/src/lib/components/critic-badge.ts:22-23); both render sites use it (CriticBadge.svelte:102, GitRail.svelte:826) and the popover gate `verdict.summaryCode || verdict.summary` (GitRail.svelte:825) mirrors PlanPanel.svelte:437. An unknown future code falls back to `summary`/generic title rather than rendering blank.
- **Locale parity.** All four keys exist in both catalogs the repo maintains (`ui/messages/` holds only en+de): en.json:606-609 and de.json:606-609, with identical key names; the map keys in critic-badge.ts:5-10 match the `ReviewSummaryCode` union in ui/src/lib/types.ts:754-755 and src/types.ts:575-576 exactly (`no-verdict-blocked|timeout|exited|unparseable`), as does the store whitelist at src/store.ts:335-340.
- **Persistence.** `summaryCode` is in the CREATE TABLE (src/store.ts:1179), both SELECTs (:2737, :2819), the INSERT column list + the 20 placeholders (:2748-2750) and the conflict update (:2752); `hydrateReview` coerces after the spread (:2716) so an unknown value degrades to null; the ALTER + legacy backfill at :3928-3940 is idempotent (`summaryCode IS NULL` guard) and scoped to `decision='error'`. Verdicts that carry it forward do so via object spreads (src/review.ts:609-621, :651-661), so no writer silently clears a stored code. Transport to the UI is the whole object (src/server.ts:933 → `snapshot()`), no field picking.
- **argv.** `tui: "default"` is unconditional and precedes the conditional MCP key and the auth spread (src/transient-agent-argv.ts:244-248), so it applies to every transient claude kind in both auth modes.
- Could not verify the shipped Claude Code gate (`Wym()`/`eo()` merge semantics for `--settings.tui`) — that CLI is not in this repo; the plan records a Phase-0 empirical check, and the settings key is assumed as written.

Nits (non-blocking):
- src/review.ts:948-963: two consecutive JSDoc blocks are stacked above `diagnoseNoVerdict` (the old `logNoVerdict` doc was left in place when the method was split), while `logNoVerdict` at :980 is now undocumented.
- src/store.ts:3936-3940: the legacy backfill asserts `no-verdict-timeout` for rows whose real cause was never recorded, so the badge now states a specific cause ("ran out of time") where the old prose claimed only "no verdict"; a dedicated legacy sentinel — or leaving those rows' prose intact — would avoid asserting a cause that may be false.
- src/critic-stuck.ts:41-44: a doomed-but-not-wedged run whose final static paint happens to be a numbered list is labelled `blocked` rather than `exited` (deliberately pinned at test/critic-stuck.test.ts:52); the finalize outcome is right, only the reason code can be wrong.

- src/prompt-recommend.ts:140 builds its own `--settings` JSON (`{disableAllHooks, ...apiKeySettingsFragment()}`) instead of going through `buildTransientAgentArgv`, so that unattended claude spawn still carries no `tui` kill-switch and can wedge on the same dialog.
- src/plan-gate.ts and src/standalone-critic.ts have the same wait-branch hole with no stuck-pane detector (the plan names this as a deliberate follow-up).

Issue #1958, opened 2026-07-28
buzz-bridge plugin: push session lifecycle into Buzz threads (session↔thread mapping)


scoop:
## Context

We run a Buzz workspace (self-hosted [block/buzz](https://github.com/block/buzz) relay at `buzz.erwins-enkel.dev`) with an agent bridge on backontop: `buzz-acp` harness + claude-agent-acp + a small MCP server wrapping the Shepherd external task API (`spawn_session`, `list_sessions`, `session_activity`, `session_diff`, `reply_to_session`, `archive_session`). Mentioning `@shepherd` in a Buzz channel spawns/steers real Shepherd sessions — working e2e since 2026-07-28.

What's missing is the **return path**: session progress/completion only surfaces when someone asks. Shepherd's plugin system (cf. claude-swap, whisper) is the right home for it.

## Proposal: `buzz-bridge` plugin

A plugin that pushes Shepherd session lifecycle into Buzz threads:

1. **Session event → Buzz thread updates**
   - `ctx.events.subscribe` on `session:status` / `session:hold`
   - On status transition (spawned → working → done/held/error), post a message into the session's Buzz thread
   - Posting via the `buzz` CLI (already built on the host) or nostr-tools + NIP-98 bridge REST, signed with the bridge agent key (NIP-OA attested)

2. **Session ↔ thread mapping**
   - `ctx.route("POST", "link", …)`: `{ sessionId, channelId, threadRootEventId }` — called by the MCP server right after `spawn_session`, so the spawn reply becomes the thread root
   - Mapping persisted in `ctx.state`
   - Unlinked sessions (spawned via Shepherd UI) post to a configurable default channel, starting a new thread

3. **Config** (`config.json`): relay URL, agent key path/env ref, default channel id, verbosity (all transitions vs terminal only)

## Notes

- Keep Buzz-side *inbound* (mentions → spawn) in the existing buzz-acp/MCP path; the plugin is outbound-only
- Follow claude-swap plugin conventions (`plugin.json`, `apiVersion: 1`, `register(ctx)` + teardown)
- Nice-to-have later: `publishStatus`/`publishUI` panel showing linked threads per session; steering replies posted in-thread flowing back via `reply_to_session` (would need an inbound piece — out of scope here)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

scoop:
✅ PR #1959 merged.

scoop:
Shipped and verified live.

[...]

PR #1959, opened 2026-07-28
feat(plugins): expose a curated session reader on ctx (#1958)

scoop:
Part of #1958 (buzz-bridge: push session lifecycle into Buzz threads).

## Why

`ctx.events.subscribe` is a plugin's only read seam, and a `session:status` payload is
essentially `{ id, status }`. `PluginContext` exposes no session reader at all — so an
outbound bridge cannot say *which* session went green. That blocks the buzz-bridge plugin
this issue is about, which needs to post "TASK-07 (shepherd) is done — PR #1958" into a
Buzz thread.

Two workarounds were considered and rejected: a plugin self-calling
`http://127.0.0.1:7330/api/sessions` with `SHEPHERD_TOKEN` (an in-process plugin
HTTP-calling its own server, silently broken on an auth misconfig), and caching
`session:new` payloads (blind to every session that predates a restart).

## What

`ctx.sessions.get(id)` / `ctx.sessions.list()` → `PluginSessionSnapshot`: `desig`, `name`,
`repoPath`, `baseBranch`, `branch`, `status`, `model`, `agentProvider`, `issueNumber`,
`haltReason`, timestamps, plus a `pr` block (`state`/`number`/`url`/`title`/`checks`/`isDraft`)
projected from the cached forge `GitState`.

Two deliberate constraints worth reviewing:

- **Curated, not the raw row.** `prompt`, `worktreePath`, `spawnAccountDir` and
  `launchMetadata` are withheld — task text and account paths widen the trust surface with
  no read-side use case, and returning `Session` verbatim would make every future internal
  field an accidental part of this versioned contract. A test asserts those four stay absent,
  so widening the surface has to be deliberate.
- **The read never mutates.** `SessionStore.getSessionGitCache(id)` is a new read-only
  getter; unlike the existing `listSessionGitCache()` it does **not** prune invalid/archived
  rows, because its caller is a plugin read. A test plants a malformed row and asserts it
  reads `null` and is still there afterwards.

Additive: `PLUGIN_API_VERSION` stays **1**, no existing plugin is affected (`claude-swap`
does not touch this), and no spawn-path or UI surface changes.

`pr` is `null` when there is no cached git state at all (never polled, or archived — the
cache holds only non-archived rows), which is distinct from a polled `state: "none"`.

## Scope

This PR is the **core capability only**. The `buzz-bridge` plugin itself ships as its own
private repo (`shepherd-buzz-bridge`), following the `shepherd-claude-swap` pattern — out of
this repo by design, per the plugin trust model. The buzz-acp MCP-server side (calling the
plugin's `POST link` right after `spawn_session`) lives in another codebase and gets its own
follow-up issue.

No manual operator steps: merging completes this change. Installing the plugin is the
plugin repo's business, not this PR's.

## Verification

- `bun run lint`, `bunx tsc --noEmit` — clean
- `bun test ./test` — 7898 pass / 0 fail (8 new in `test/plugins-sessions.test.ts`)
- `cd ui && bun run check` — 0 errors; `check:i18n` in parity
- `bunx fallow audit --base origin/main --fail-on-issues` — exit 0
- docs manifest regenerated (`cd ui && bun run gen:docs`); generated-docs gate clean

[no-feature-entry] — server-side plugin API, ships no user-facing UX.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -173,9 +173,11 @@ permission-scoped / out-of-process) without changing your call sites.
 
 ## Reading sessions (`ctx.sessions`)
 
-The event stream tells you _that_ something happened, not _what_ it happened to: a
-`session:status` payload is essentially `{ id, status }`. `ctx.sessions` turns that id into
-something you can render.
+The event stream tells you _that_ something happened, not _what_ it happened to. Treat a
+`session:status` payload as `{ id, status }` and nothing more — most emitters send exactly
+that, and while one currently sends the whole session row, a plugin that reads the extra
+fields is relying on an emitter detail, not on the contract. `ctx.sessions` is how you turn
+that id into something you can render.
 
 ```ts
 ctx.events.subscribe((event, data) => {
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -157,18 +157,61 @@ Neither auto-loads from the repo (the loader only ever scans `~/.shepherd/plugin
 everything goes through `ctx`, so a future Shepherd can swap the implementation (curated /
 permission-scoped / out-of-process) without changing your call sites.
 
-| Capability                         | What it does                                                                                                                                                   |
-| ---------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
-| `ctx.onSpawn(fn)`                  | Mutate how an agent launches (see below). The load-bearing capability.                                                                                         |
-| `ctx.events.subscribe(fn)`         | Observe the **read-only** core event stream (`session:hold`, `session:status`, …). Returns an unsubscribe fn. Plugins cannot _emit_ core events.               |
-| `ctx.publishStatus(json)`          | Push a small free-form JSON blob to the status panel (rendered verbatim).                                                                                      |
-| `ctx.publishUI(view)`              | Push a declarative UI view to the Settings → Plugins panel (`null` clears). Additive.                                                                          |
-| `ctx.publishGearItem(item)`        | Add a single item to the top-bar gear menu (`null` clears). Additive.                                                                                          |
-| `ctx.state`                        | Durable, **per-plugin-scoped** key/value: `get`/`set`/`delete`/`keys`. Values are JSON. Backed by a `plugin_state` table — you never touch the session schema. |
-| `ctx.route(method, path, handler)` | Register an HTTP route under `/api/plugins/<id>/<path>`. Sits behind operator auth.                                                                            |
-| `ctx.log`                          | Namespaced logger into `shepherd.log` (`ctx.log.log` / `ctx.log.warn`).                                                                                        |
-| `ctx.config`                       | Your plugin's own `config.json` (parsed; `{}` when absent).                                                                                                    |
-| `ctx.abortSpawn(reason)`           | Hard-block the in-flight spawn from inside an `onSpawn` hook (throws).                                                                                         |
+| Capability                         | What it does                                                                                                                                                            |
+| ---------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `ctx.onSpawn(fn)`                  | Mutate how an agent launches (see below). The load-bearing capability.                                                                                                  |
+| `ctx.events.subscribe(fn)`         | Observe the **read-only** core event stream (`session:hold`, `session:status`, …). Returns an unsubscribe fn. Plugins cannot _emit_ core events.                        |
+| `ctx.publishStatus(json)`          | Push a small free-form JSON blob to the status panel (rendered verbatim).                                                                                               |
+| `ctx.publishUI(view)`              | Push a declarative UI view to the Settings → Plugins panel (`null` clears). Additive.                                                                                   |
+| `ctx.publishGearItem(item)`        | Add a single item to the top-bar gear menu (`null` clears). Additive.                                                                                                   |
+| `ctx.state`                        | Durable, **per-plugin-scoped** key/value: `get`/`set`/`delete`/`keys`. Values are JSON. Backed by a `plugin_state` table — you never touch the session schema.          |
+| `ctx.sessions`                     | Read-only session lookup: `get(id)` / `list()` → a curated `PluginSessionSnapshot`. Resolves the bare ids that `session:*` events carry. Plugins cannot write sessions. |
+| `ctx.route(method, path, handler)` | Register an HTTP route under `/api/plugins/<id>/<path>`. Sits behind operator auth.                                                                                     |
+| `ctx.log`                          | Namespaced logger into `shepherd.log` (`ctx.log.log` / `ctx.log.warn`).                                                                                                 |
+| `ctx.config`                       | Your plugin's own `config.json` (parsed; `{}` when absent).                                                                                                             |
+| `ctx.abortSpawn(reason)`           | Hard-block the in-flight spawn from inside an `onSpawn` hook (throws).                                                                                                  |
+
+## Reading sessions (`ctx.sessions`)
+
+The event stream tells you _that_ something happened, not _what_ it happened to: a
+`session:status` payload is essentially `{ id, status }`. `ctx.sessions` turns that id into
+something you can render.
+
+```ts
+ctx.events.subscribe((event, data) => {
+  if (event !== "session:status") return;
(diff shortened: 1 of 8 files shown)


scoop (commented):
## What this PR is

Deliverable **A** of the plan only: an additive, read-only `ctx.sessions` capability on the plugin seam (types + pure mapper + a narrow store getter + loader wiring + tests + docs). The actual `buzz-bridge` plugin (deliverable B) lives in a separate private repo by design, so nothing about it is verifiable from this checkout — see *Stated limitations*.

## Verification performed (ground truth cited)

**Type unions mirror core exactly.** `PluginSessionSnapshot.status` (`src/plugins/types.ts:140`) is `"running"|"idle"|"blocked"|"done"|"archived"` — identical to `SessionStatus` at `src/types.ts:6`. `agentProvider` is `"claude"|"codex"` — identical to `AGENT_PROVIDERS` at `src/types.ts:13`. `haltReason` matches `src/types.ts:116` verbatim. `PluginSessionPr.state`/`checks` match `PrStatus.state` (`src/forge/types.ts:200`) and `ChecksState` (`src/forge/types.ts:130`); `number`/`url`/`title`/`isDraft` all exist as optionals on `PrStatus` (`src/forge/types.ts:201-203`, `:228`). The header comment's claim that a new core status/provider "fails to compile HERE" (`src/plugins/session-view.ts:9-11`) holds: `session.status` / `session.agentProvider` are assigned into the narrower inline unions in the mapper, so widening `SessionStatus` is a compile error at `src/plugins/session-view.ts:44`/`:47`.

**Store getter is consistent with its sibling.** `getSessionGitCache` (`src/store.ts:2579-2587`) uses the same `s.status != 'archived'` predicate `putSessionGitCache` writes under (`src/store.ts:2566`), is parameterised (no injection), and — unlike `listSessionGitCache` (`src/store.ts:2589-2610`, which deletes invalid/archived rows) — performs no mutation. Bun's `.get()` yielding `null`/`undefined` is handled by the `row ? … : null` ternary. `archive()` already deletes the cache row (`src/store.ts:2557`), so the archived filter is defensive rather than load-bearing; harmless.

**Loader wiring and contract compatibility.** `sessions` is built per-call inside closures (`src/plugins/loader.ts:347-357`), so the docs claim "reads are live, not a boot-time freeze" (`docs/plugins.md:197`) is true — and `test/plugins-sessions.test.ts:226-247` proves it. `PluginRegistryDeps.store` widening to `PluginStateStore & PluginSessionReadStore` (`src/plugins/loader.ts:63`) is satisfied by every construction site I found: the production one at `src/index.ts:624` and every test at `test/plugins*.test.ts` / `test/examples-plugin.test.ts` / `test/plugin-activate.test.ts` pass a real `SessionStore`, which has `get` (`src/store.ts:2312`), `list` (`src/store.ts:2319`, optional-arg signature is assignable to the zero-arg interface method) and the new `getSessionGitCache`. The only fake is `src/plugins/loader.test.ts:14`, updated in this diff. `PluginContext` gaining a required member is safe for consumers: the only other references (`examples/plugins/spawn-labeler/index.ts:50`, `test/fixtures/example-plugin/index.ts:13`) take it as a parameter; nothing constructs a `PluginContext` literal outside `makeContext`.

**Snapshot really is a copy.** Every projected field is a primitive or `null` (`src/plugins/session-view.ts:38-55`), and `toPluginSessionPr` builds a fresh object, so "mutating one does nothing" (`docs/plugins.md:198`) is accurate — no aliasing of the cached `GitState`. Optional PR fields are omitted rather than set to `undefined`, so the JSON round-trip assertion at `test/plugins-sessions.test.ts:136` is meaningful.

**Docs/manifest parity.** The new H2 `## Reading sessions (\`ctx.sessions\`)` (`docs/plugins.md:174`) sits between `## The \`ctx\` capability seam` (`:154`) and `## The \`onSpawn\` hook` (`:216`); `ui/scripts/gen-docs-manifest.ts:58-64,77-84` strips backticks, lowercases and joins headings in document order — which is exactly the `"… the ctx capability seam reading sessions (ctx.sessions) the onspawn hook …"` string committed at `ui/src/lib/docs-manifest.ts:102`. No other heading changed, so no further manifest drift. `PLUGIN_API_VERSION` is untouched at `1` (`src/plugins/types.ts:11`), and `capabilities[]` remains unenforced (`src/plugins/types.ts:13`), so the probe fixture declaring only `["state"]` while using routes is consistent with existing behaviour.

**Test-file sanity.** `handleRoute(method, pluginId, subPath, req)` (`src/plugins/loader.ts:495-500`) matches the call at `test/plugins-sessions.test.ts:74`; `index.ts` is a resolvable entry (`src/plugins/loader.ts:307`). The unparseable-row test uses a file-backed DB precisely so the no-prune assertion is observable — a good choice.

## Nits (non-blocking):
- `src/plugins/loader.ts:353-356` — `list()` issues one `getSessionGitCache` query per session (N+1) on the server's single event loop; `listSessionGitCache()` already does it in one statement, and a plugin polling `list()` per event will multiply that. Cheap indexed lookups, so not a defect, but a batched read would age better.
- `docs/plugins.md:177` — "a `session:status` payload is essentially `{ id, status }`" is true for the emits at `src/index.ts:1029-1035` and `src/server.ts:2924`, but `src/server.ts:3274` emits the whole `Session`; the hedge ("essentially") carries it, though a plugin author trusting the sentence would be surprised.

- `session:new` is emitted with the full session row (`src/index.ts:1965`: `{ ...s, hasScratchpadFiles: false }`), so `prompt`/`worktreePath` already reach any `ctx.events.subscribe` listener — the snapshot's withholding of those fields narrows the *new* surface but does not, today, keep task text away from plugins.
- `session:status` is emitted inconsistently across core: `{ id, status }` at `src/index.ts:1029-1035` and `src/server.ts:2924`, but the entire `Session` at `src/server.ts:3274`.

## Stated limitations
- The `shepherd-buzz-bridge` plugin repo (deliverable B of the plan: link/links routes, transition→message mapping, `buzz` CLI posting) is not in this repository and could not be inspected; I neither confirm nor dispute that it exists or works. Judged on its own, this diff is enabling plumbing, and the originating issue's outbound bridge is not satisfied by this repo alone — by the approved plan's explicit design, not by omission.
- Prettier/eslint/tsc/bun test were not run (CI's job). The reflowed capability table in `docs/plugins.md:160-172` has the same "separator row two chars longer" shape as the untouched table at `docs/plugins.md:115-123`, so it is consistent with what is already committed.

scoop:
Follow-on work landed outside this PR (both from #1958):

- **`erwins-enkel/shepherd-buzz-bridge`** — the plugin itself, published private (the plugin trust model keeps implementations out of this repo). Consumes `ctx.sessions` from this PR; 41 tests, `bun run verify` green.
- **#1960** — the buzz-acp MCP-server side (calling the plugin's `POST link` after `spawn_session`), which lives in a different codebase.

scoop:
Both critic nits addressed.

**`docs/plugins.md` — "essentially `{ id, status }`"** — fixed in 4a78af4. Reworded so the sentence states the *contract* (`{ id, status }` and nothing more) and explicitly warns that the one emitter sending the whole row is an emitter detail, not something to rely on. The critic was right that an author trusting the old phrasing could be surprised.

**`list()` N+1 on `getSessionGitCache`** — deliberately left as-is. `session_git_cache.sessionId` is the table's PRIMARY KEY (`src/store.ts:1101`), so each read is a single indexed lookup, and `store.list()` already hydrates every row in the same call — the per-row cache read is not the dominant cost. The batched alternative, `listSessionGitCache()`, is unusable here for a correctness reason rather than a performance one: it **prunes** invalid and archived rows as a side effect (`src/store.ts:2604-2609`), and this PR's whole stance is that a plugin read must never mutate core state (there is a test asserting exactly that). Getting the batch without the mutation would mean either duplicating the query or refactoring the existing pruning method — more diff and more risk than an indexed lookup per row justifies today. Worth revisiting if a plugin ever polls `list()` on a hot path; noting it here so the trade-off is on the record rather than accidental.

Also acknowledging the two **out-of-scope** observations, both correct and both pre-existing: `session:new` already ships the full row (including `prompt`) to any `ctx.events` subscriber, so this snapshot narrows the *new* surface without making task text unreachable today; and `session:status` is emitted inconsistently across core. Neither is this PR's to fix, but they are the natural follow-up if the withholding is ever meant to be a real boundary rather than a contract stance.

Issue #1960, opened 2026-07-28
Verify the buzz-acp agent actually passes the thread-root params to spawn_session


scoop:
Narrowed from the original scope (see the comment below): the MCP-server plumbing this issue
asked for **already exists** — `~/shepherd-mcp` `91a9a4b` calls
`POST /api/plugins/buzz-bridge/link` after `spawn_session`. What is unverified is whether the
values it needs ever arrive.

## The gap

`spawn_session` takes `buzzChannelId` and `buzzThreadRootEventId`, and both are **optional**.
The MCP server only links a thread when both are present:

```ts
if (sessionId && buzzChannelId && buzzThreadRootEventId) { … POST …/link … }
```

Nothing instructs the agent to supply them. The tool descriptions say what they are for, so
the model may infer it from the mention it is responding to — but that is an assumption, and
the path has never run end-to-end.

Failure mode is **silent, not broken**: with the params missing, buzz-bridge falls back to its
auto-root behaviour — the first update posts top-level in the default channel and becomes the
thread root. Updates still thread, just not under the `@shepherd` spawn reply, so the
conversation that requested the work and the reporting on it sit in two separate threads.

## Verify

1. Mention `@shepherd` in the `shepherd` channel with a real task.
2. Check the tool result: it appends `(linked to Buzz thread)` on success, or
   `(thread link failed: …)` when the POST itself fails.
3. `GET /api/plugins/buzz-bridge/links` → the new session's entry should show **`auto: false`**
   (an explicit link) and a `threadRootEventId` matching the spawn reply's event id.
   `auto: true` means the params never arrived and the bridge rooted the thread itself.
4. Confirm the next lifecycle post (PR opened / done) lands **as a reply under the spawn
   reply**, not as a fresh top-level message.

## If they do not arrive

Make the agent supply them — via the buzz-acp harness (which knows the channel and the event
it is replying to) or an instruction that the mention's channel id and event id go into
`spawn_session`. Passing them from the harness is preferable to relying on the model
remembering.

## Done when

A mention-spawned session shows `auto: false` in `GET links`, and its PR/done updates appear
threaded under the spawn reply in Buzz.

scoop:
Retargeted. As originally filed this asked for the MCP server to call the plugin's `POST /link`
after `spawn_session` — I filed it without checking that codebase, and it turned out to be
implemented in `~/shepherd-mcp` (`91a9a4b`) within the hour: the server parses the session id
[...]

PR #1965, opened 2026-07-29
docs: show how to filter archived sessions from ctx.sessions.list()

scoop:
The `list()` bullet in **Reading sessions (`ctx.sessions`)** told readers archived rows are included and that they should filter, but showed no code. Adds the one-liner.

```ts
const live = ctx.sessions.list().filter((s) => s.status !== "archived");
```

Verified against `src/plugins/types.ts` → `PluginSessionSnapshot`: `status` is `"running" | "idle" | "blocked" | "done" | "archived"`, so the check is accurate.

Docs-only; no source/tests touched. [no-feature-entry]

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -196,6 +196,10 @@ ctx.events.subscribe((event, data) => {
 - **`list()`** → every session core holds, **archived rows included** (filter on
   `status === "archived"` / `archivedAt` if you only want live ones).
 
+```ts
+const live = ctx.sessions.list().filter((s) => s.status !== "archived");
+```
+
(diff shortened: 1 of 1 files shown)


scoop:
Closing — this was make-work from a smoke test of the Buzz thread-linking path (#1960), not a real documentation gap.

The snippet is correct, but it restates in code the sentence directly above it ("filter on `status === "archived"` / `archivedAt` if you only want live ones"), and the fence renders outside the bullet list it belongs to. Not worth a commit in the release notes.

If the copy-pasteable line is ever wanted, folding it into the bullet is the better shape than adding a floating block.

Issue #560, opened 2026-06-12
tmpfs inode exhaustion: spawned agents fill /tmp inodes (ENOSPC with free bytes)


scoop:
## Summary

On long-uptime hosts, Shepherd-spawned agents exhaust the **inodes** on the `/tmp` tmpfs, producing intermittent `ENOSPC: no space left on device` errors in agent tool calls — **even though disk bytes are nearly empty**. Observed at 88% inode use (915,834 / 1,048,576) with only 25% of bytes used; a burst of small-file creation then tips a write over the `nr_inodes` cap and fails. The failure is silent and corrupting in practice: in one session a `git fetch` died mid-command on ENOSPC, left a stale ref, and the agent then reasoned confidently from 20-hour-old remote state.

This recurs across sessions because `/tmp` is tmpfs (cleared only on reboot) and Shepherd never prunes what it puts there.

## Root cause (Shepherd-driven)

Shepherd sets **`TMPDIR` / `CLAUDE_CODE_TMPDIR=/tmp/claude-1000`**, pinning both Node's compile cache and every spawned agent's scratch onto tmpfs, where it persists across spawns. Combined with Shepherd's high spawn volume (critics, plan-gate reviewers, autopilot triagers, learnings analysts, per-worktree task agents), this accumulates without bound.

Evidence from an affected host (`df -i /tmp` at 88%):

| inodes | path | what |
|---|---|---|
| **758,157** | `/tmp/claude-1000/node-compile-cache/` | Node V8 compile cache — tiny per-module files, **74% of all `/tmp` inodes**. `NODE_COMPILE_CACHE` is unset, so Node defaults it to `$TMPDIR/node-compile-cache` → lands on the tmpfs Shepherd pinned. Oldest files 9 days old, continuously growing. |
| 95,886 | `/tmp/claude-1000/claude-1000/` | Per-worktree session scratch, named after Shepherd worktrees (`-home-patrick-Work--shepherd-worktrees-shepherd-review-…`). Never swept when the worktree is archived/pruned. |
| 8,451 | `/tmp/claude-1000/bunx-1000-*` | bunx tool caches, regenerable |

The **doubled `/tmp/claude-1000/claude-1000/`** is a Shepherd artifact: the outer `claude-1000` is Shepherd's `TMPDIR` (already a `claude-$uid` dir), and nested Claude instances re-derive `claude-$uid` under it.

[...]

PR #1968, opened 2026-07-29
fix(spawn): route the herdr ≥0.7.5 pane spawn through a temp script

### LINKED ISSUE #1967: fix(spawn): macOS session spawn truncated by tty MAX_INPUT — type-into-shell overflows on Darwin
## Summary

On macOS, **every non-trivial session spawn fails** with:

```
herdr: agent for pane <pane> (<name>) not auto-detected within 30s.
```

Root-caused during Mac validation of #1912/#1921. This is **not** a herdr-version or PATH problem, and **not** related to preview detection — it is a spawn-transport bug specific to Darwin.

## Root cause

herdr's `pane run` does **not** `execvp` the command — it **types the command line into the pane's interactive shell** (`src/herdr.ts` → `runInReadyPane`). Shepherd's spawn command is ~2KB, dominated by the `--settings` JSON overlay + the `--append-system-prompt` directive (`src/service.ts` → `buildClaudeSpawnArgv` / `composeDirectives`).

macOS tty **canonical-mode** input is bounded by `MAX_INPUT` (~1024 bytes); Linux's line-discipline buffer is 4096, which is why this only breaks on Darwin. The typed line is silently truncated ~1KB in — mid-`--settings`, before it even reaches `--append-system-prompt` — leaving an **unterminated `'` quote**. The shell waits for line continuation, `claude` never launches, and herdr's auto-detect (`resolveByAutoDetect`) times out at 30s and rolls the pane back.

Confirmed empirically on macOS:
- Bare `herdr pane run <pane> claude` → detected fine (`terminal_id` populated).
- `herdr pane run <pane> env CLAUDE_CODE_DISABLE_ALTERNATE_SCREEN=1 claude` → detected fine.
- Full Shepherd command → typed line truncated mid-JSON, sits unexecuted at the prompt.

## Why Shepherd doesn't catch it

`src/argv-limit.ts` → `argvElementLimit(platform)` returns `pageSize * 32` for Linux and **`Number.POSITIVE_INFINITY` for every other platform** — it only models Linux's `execve` `E2BIG` (`MAX_ARG_STRLEN`), which is the wrong failure mode for the type-into-shell transport. So on macOS the argv-size budget/trim safety valve (#1944) never engages, and the untrimmed command overflows the tty line buffer unguarded.

## Scope

Not a fringe case: the `--append-system-prompt` directive alone is several KB, so the limit is exceeded for essentially any real spawn. There is **no config-only workaround** — disabling hook ingest shrinks `--settings` but the system prompt still overflows.

## Proposed fix

Make the *typed* line short regardless of command size: write the full `env … claude …` command to a temp script under `agentTmpDir()` and have herdr run `sh <script>` (a ~30-byte line, well under `MAX_INPUT`). The script execs claude; herdr auto-detects the foreground exactly as today. Contained change to the spawn transport (`runInReadyPane` / `buildWrappedArgv`), platform-agnostic, reuses the existing detection path.

Tightening `argvElementLimit` for Darwin is **not** sufficient — trimming a system prompt down to <1KB would gut it.

## Acceptance

- A session spawns successfully on macOS with a full-size `--settings` overlay + system-prompt directive.
- The typed pane-run line stays under Darwin's `MAX_INPUT`.
- Linux spawn behavior byte-identical / unchanged.

scoop:
Fixes #1967.

## The bug

herdr's `pane run` (CLI) and `pane.send_text` (socket) do **not** `execvp` the command — they **type it into the pane's interactive shell**. A shell at its prompt has its tty in **canonical** mode, where input is bounded by `MAX_INPUT`: **~1024 bytes on Darwin** (4096 on Linux).

Shepherd's spawn line is multi-KB (the `--settings` overlay + the `--append-system-prompt` directive), so on macOS it was silently truncated mid-`--settings`, leaving an unterminated `'`. The shell sat waiting for line continuation, `claude` never launched, and every non-trivial macOS spawn died at the 30s auto-detect deadline:

```
herdr: agent for pane <pane> (<name>) not auto-detected within 30s.
```

`argvElementLimit` never engaged because it models Linux's `execve` `MAX_ARG_STRLEN` — the wrong failure mode for a type-into-shell transport. Tightening it for Darwin is not a fix either: trimming a system prompt under 1 KB would gut it.

## The fix

Write the wrapped argv to a throwaway script and type only `sh '<path>'` (~60 bytes):

```sh
#!/bin/sh
rm -f -- "$0"
exec 'env' '-u' 'CLAUDE_CODE_TMPDIR' … 'claude' '--settings' '…' …
```

- **`exec` is load-bearing** — without it `sh` stays the pane's foreground process and herdr detects `sh` instead of the agent.
- **`rm -f -- "$0"` self-deletes** — the open fd keeps it readable, so nothing is left behind. Retry-safe: both drivers only retry a run that was *rejected before execution*, so an unlaunched script is still on disk for the next attempt.
- **0600 file / 0700 dir, `wx` exclusive create** — the script carries the full spawn command (incl. env tokens like `CLAUDE_CONFIG_DIR`), and the exclusive create means it can never land through a pre-placed symlink.
- **Not marked executable** — it is handed to `sh` as an argument, so it also runs from a `noexec` mount.

Applied to **both** ≥0.7.5 transports (`HerdrDriver.runInReadyPane`, `SocketHerdrDriver.runInReadyPane`) on **all platforms** — one transport, no Darwin-only branch. The process that ends up running is byte-identical to before (same argv, same env), so detection, the membrane/bwrap wrap and every downstream consumer are unaffected. The ≤0.7.4 `agent start` paths are untouched.

Side benefit: the typed line no longer depends on the *interactive* shell being POSIX — `sh '<path>'` parses correctly even under fish.

## Verification

**Live spike** (herdr 0.7.4 `pane run`, the assumption unit tests cannot reach) — `tab create` → `pane run "sh '<script>'"` where the script execs `claude`:

- herdr auto-detected the agent in ~2s with `terminal_id` populated;
- `pane process-info` foreground = `claude` (argv `["claude"]`), **no `sh`** — `exec` chained correctly;
- the script had removed itself.

**Tests** — `test/spawn-script.test.ts` (new):

- the typed line for a realistic ~10 KB spawn is **< 1024 bytes**, and the driver-level guard asserts the same through `start()`;
- a **real `/bin/sh` run** reads back the reconstructed argv (multi-word, newline, quote and empty tokens intact) and confirms the script deleted itself mid-run;
- `exec` proven by **pid identity**, not inspection: the outer shell prints its pid, execs the typed line, and the launched command prints the same pid — a missing `exec` anywhere in the chain forks and they differ;
- 0600/0700 modes, per-call unique paths, and the `SHEPHERD_AGENT_TMPDIR=""` fallback.

Gates: `bun run lint`, `bunx tsc --noEmit`, `bun test ./test` (7996 pass / 0 fail), `fallow audit` — all green. The one duplication finding is pre-existing (`startImpl075` vs `startImpl`).

## Notes

- **Steering is not affected and deliberately untouched**: `pane.send_text` into a *live* agent writes to a tty in **raw** mode, and `MAX_INPUT` bounds the canonical-mode buffer only. The spawn broke precisely because it types at an interactive shell prompt.
- **`argv-limit` / `spawn-budget` are unchanged** — the eventual `execve` of `env … claude …` still enforces Linux's per-element limit, so the guard stays correct; it now merely over-counts (it prices the whole joined line), which errs toward clamping early and never toward `E2BIG`. Doc comments updated to stop claiming the joined line is one argv element.
- Local herdr is 0.7.4, which takes the legacy `agent start` (execvp) path, so this changes nothing on the Linux host today — it engages on herdr ≥0.7.5.

[no-feature-entry] — server-only spawn-transport fix, no user-facing UX.

---

## Update: end-to-end on a real herdr 0.7.5

The transport only engages on herdr ≥0.7.5, and this box runs 0.7.4 — so I stood up a **throwaway 0.7.5** to close that gap: the 0.7.5 binary downloaded to a scratch dir, run as an isolated session with its own socket (`--session e2e1967`). The global 0.7.4 binary and the live herd were never touched; the instance was stopped and removed afterwards.

Driving the **real `HerdrDriver.start()`** with a 9,689-byte spawn (settings overlay + system-prompt directive):

```
joined command line: 9689 bytes
A (pre-fix, 9689B typed): DETECTED claude          # Linux tolerates it — see below
B (fixed transport):      DETECTED claude in 1014ms
   foreground: ["claude", …]                       # exec chained, no `sh`
   leftover spawn scripts: 0
   stopped + tab closed
```

Worth recording: **on Linux the pre-fix transport still works at 9.7 KB** (arm A). An interactive shell at its prompt puts the tty in *raw* mode and does its own line editing, so `MAX_INPUT` bites only where the input queue actually overflows — Darwin's smaller buffer. So this stays a Darwin-visible bug, exactly as reported; the Linux path is unchanged and still fine.

## Two follow-ups in the second commit

**1. Directory squat (blocking review finding — valid).** With the #1875 redirect rolled back (`SHEPHERD_AGENT_TMPDIR=""`) the directory is a fixed `<os.tmpdir()>/spawn` in a world-writable parent, and `mkdir(…, {recursive: true, mode})` accepts a pre-existing directory **without applying the mode**. A local user owning that directory (or a symlink planted on it) could unlink our 0600 script and substitute their own between the write and the pane's `sh` opening it — code execution as the Shepherd user. `wx` doesn't help: it guards only the final component, inside a directory the attacker controls.

Now `lstat`-verified before anything is written — real directory, our uid, no group/other bits — falling back to a private `mkdtemp` directory rather than refusing, so a squatter can't turn it into a spawn-wide DoS either. The fallback is **re-verified per spawn**, never blindly reused: a tmp cleaner removing it would otherwise strand every later spawn on `ENOENT` (caught by a test I wrote for the symlink case).

**2. Abandoned scripts.** The e2e surfaced this: the self-delete only fires when the script *runs*, so a spawn whose `pane run` fails every attempt leaves one behind — a full test-suite run had left four in the agent tmp root, the directory whose inode table Shepherd has exhausted before (#560/#1875). My plan called this negligible; the evidence said otherwise. The next spawn now reaps anything older than an hour (two orders of magnitude above the 30s spawn deadline, so a script still waiting for its pane is never reaped), best-effort and never fatal to a spawn. Verified live: four stale scripts → 0 after one spawn.

Also fixed the two CodeQL `js/indirect-command-line-injection` alerts on the test file — the real-shell tests now type the line at `sh` on **stdin**, which is both taint-free and the faithful reproduction of the transport (a shell reading typed input).

Gates re-run after all of it: lint, `tsc --noEmit`, `bun test ./test` (8002 pass / 0 fail), `fallow audit`.

### DIFF
--- a/src/argv-limit.ts
+++ b/src/argv-limit.ts
@@ -59,10 +59,12 @@ export function joinedElementBytes(s: string): number {
   return Buffer.byteLength(posixShellJoin([sanitizePromptArg(s)]), "utf8");
 }
 
-/** Bytes the WHOLE argv costs once joined into a single shell command line — the exact quantity
- *  the herdr 0.7.5 `pane run` / `pane.send_text` paths push through one argv element. On the
- *  ≤0.7.4 spread paths each token is its own element, so this over-counts; that errs toward
- *  clamping early and never toward `E2BIG`. */
+/** Bytes the WHOLE argv costs once joined into a single shell command line. A deliberately
+ *  CONSERVATIVE model of the spawn's cost: every real path spends fewer bytes per element than this.
+ *  The ≤0.7.4 paths spread the argv, so each token is its own element; and since #1967 the ≥0.7.5
+ *  paths type `sh '<script>'` and let the script's `exec` spread it too (the joined line now lives in
+ *  a file, where no `MAX_ARG_STRLEN` applies). Over-counting errs toward clamping early and never
+ *  toward `E2BIG`. */
 export function spawnFootprintBytes(argv: string[]): number {
   let total = 0;
   for (const tok of argv) total += joinedElementBytes(tok);
--- a/src/herdr-socket-driver.ts
+++ b/src/herdr-socket-driver.ts
@@ -14,7 +14,6 @@ import {
   parseProcs,
   parseReadText,
   parseTabs,
-  posixShellJoin,
   resolvePaneId,
   sanitizeHerdrAgentName,
   stopViaRecordedTab,
@@ -29,6 +28,7 @@ import {
   herdrSpawnSupported,
   herdrUsesExternalRegistrationSpawn,
 } from "./herdr-capabilities";
+import { spawnCommandLine } from "./spawn-script";
 
 /**
  * Socket-backed `IHerdrDriver` (issues #1529, #1553, #1567): routes the async read surface —
@@ -251,24 +251,30 @@ export class SocketHerdrDriver implements IHerdrDriver {
 
   /**
    * Launch `wrapped` in `paneId`'s shell. herdr's protocol-17 socket API exposes NO direct pane-run
-   * RPC (unlike the CLI `pane run`, #1892), so the wrapped argv is TYPED into the pane's ready shell
-   * as a POSIX-quoted command line (`pane.send_text`) then submitted with Enter (`pane.send_keys`).
+   * RPC (unlike the CLI `pane run`, #1892), so the command is TYPED into the pane's ready shell
+   * (`pane.send_text`) then submitted with Enter (`pane.send_keys`). What gets typed is `sh
+   * '<script>'` — the wrapped argv itself rides a throwaway script (`spawnCommandLine`, #1967), so
+   * the line can never outgrow the tty's canonical-mode `MAX_INPUT`.
    * `send_text` is bounded-retried on ANY failure to ride out shell-not-ready without hinging on an
    * unverified busy-error code: a rejected `send_text` means herdr didn't write it, so a retry never
    * double-types (mirrors the CLI `runInReadyPane` reject-before-exec contract). The Enter is sent
    * once, OUTSIDE that loop, so a delivered command is never re-typed. NOTE: this routes through the
-   * interactive shell rather than a direct exec — the per-token quoting keeps it robust; only the
-   * 0.7.5 socket spawn path (default-off) reaches it.
+   * interactive shell rather than a direct exec — the script's per-token quoting keeps it robust;
+   * only the 0.7.5 socket spawn path (default-off) reaches it.
    */
   private async runInReadyPane(paneId: string, wrapped: string[]): Promise<void> {
     // #1944: the socket path cannot surface a kernel `E2BIG` — the argv crosses a socket to the
     // daemon, which execs it out of our reach and fails opaquely. So the per-element limit is
     // checked PROACTIVELY here, before the retry loop, so an argv that can never exec fails once
-    // with a named cause instead of three times with a shell-not-ready-looking error. The pane's
-    // shell reconstructs the exact argv from the quoted line, so each `wrapped` token becomes one
-    // argv element and is the right thing to measure.
+    // with a named cause instead of three times with a shell-not-ready-looking error. The spawn
+    // script's `exec` reconstructs the exact argv, so each `wrapped` token becomes one argv element
+    // and is the right thing to measure.
     assertArgvWithinLimit(wrapped, "herdr socket pane.send_text");
-    const cmdline = posixShellJoin(wrapped);
+    // The typed line goes through the pane's tty in CANONICAL mode, where `MAX_INPUT` (~1024 bytes
+    // on Darwin) truncates our multi-KB spawn line mid-flight — so the argv rides a throwaway script
+    // and only `sh '<path>'` is typed (#1967). Written ONCE, before the retry loop: a rejected
+    // `send_text` never delivered the line, so the script is still on disk for the next attempt.
+    const cmdline = await spawnCommandLine(wrapped);
     const MAX_ATTEMPTS = 3;
     let lastErr: unknown;
     for (let attempt = 0; attempt < MAX_ATTEMPTS; attempt++) {
--- a/src/herdr.ts
+++ b/src/herdr.ts
@@ -10,6 +10,7 @@ import {
 } from "./herdr-capabilities";
 import { posixShellJoin, oversizedFromArgv } from "./argv-limit";
 import { maintenance } from "./maintenance";
+import { spawnCommandLine } from "./spawn-script";
 import { compileCacheDir, agentTmpDir } from "./tmp-sweep";
 import type { HerdrState, LivenessState, SessionStatus } from "./types";
 import type { RequestPaneAgentState } from "./generated/herdr-protocol";
@@ -1047,11 +1048,16 @@ export class HerdrDriver implements IHerdrDriver {
    */
   private async runInReadyPane(paneId: string, wrapped: string[]): Promise<void> {
     // herdr's `pane run` TYPES the command into the pane's interactive shell (it does not execvp the
-    // argv), so a raw multi-arg spread (`...wrapped`) is space-joined UNQUOTED — any argv element
-    // with whitespace or a newline (e.g. claude's multi-line `--append-system-prompt`) shatters into
-    // bogus shell commands. Pass the whole argv as ONE POSIX-quoted command line instead (mirrors the
-    // socket driver's `posixShellJoin` typing path), so the shell reconstructs the exact argv.
-    const cmdline = posixShellJoin(wrapped);
+    // argv). Two consequences, both handled by `spawnCommandLine`:
+    //  - a raw multi-arg spread (`...wrapped`) would be space-joined UNQUOTED, shattering any element
+    //    with whitespace or a newline (e.g. claude's multi-line `--append-system-prompt`);
+    //  - the typed line hits the tty's CANONICAL-mode `MAX_INPUT` (~1024 bytes on Darwin), which our
+    //    multi-KB spawn line silently overran — truncated mid-`--settings`, unterminated quote, no
+    //    `claude`, 30s auto-detect timeout (#1967).
+    // So the argv goes into a throwaway script and only `sh '<path>'` is typed. Written ONCE, before
+    // the retry loop: a rejected `pane run` never executed the script, so it is still on disk for the
+    // next attempt (and self-deletes the moment it does run).
+    const cmdline = await spawnCommandLine(wrapped);
     const MAX_ATTEMPTS = 3;
     let lastErr: unknown;
     for (let attempt = 0; attempt < MAX_ATTEMPTS; attempt++) {
--- a/src/spawn-script.ts
+++ b/src/spawn-script.ts
@@ -1,5 +1,5 @@
 import { randomBytes } from "node:crypto";
-import { mkdir, writeFile } from "node:fs/promises";
+import { lstat, mkdir, mkdtemp, readdir, rm, stat, writeFile } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { posixShellJoin } from "./argv-limit";
@@ -31,6 +31,51 @@ function spawnScriptDir(): string {
   return join(agentTmpDir() ?? tmpdir(), "spawn");
 }
 
+/** A real directory, owned by us, that no one else can write into. Uses `lstat`, so a symlink
+ *  planted on the path is rejected rather than followed. */
+async function isOwnPrivateDir(dir: string): Promise<boolean> {
+  try {
+    const st = await lstat(dir);
+    return st.isDirectory() && st.uid === process.getuid?.() && (st.mode & 0o077) === 0;
+  } catch {
+    return false;
+  }
+}
+
+/** Remembered across spawns once the preferred directory turns out to be squatted. */
+let fallbackDir: string | null = null;
+
+/** A private directory of our own under the system tmp dir, reused across spawns — but RE-VERIFIED
+ *  every time, never blindly reused: a tmp cleaner (or anything else) may have removed it since,
+ *  and a stale memo would fail every subsequent spawn with `ENOENT`. */
+async function privateFallbackDir(): Promise<string> {
+  if (fallbackDir && (await isOwnPrivateDir(fallbackDir))) return fallbackDir;
+  fallbackDir = await mkdtemp(join(tmpdir(), "shepherd-spawn-"));
+  return fallbackDir;
+}
+
+/**
+ * The directory to write into, PROVEN to be ours and private before anything is written to it.
+ *
+ * The check is load-bearing, not defensive dressing. With the #1875 redirect rolled back
+ * (`SHEPHERD_AGENT_TMPDIR=""`) the path is a fixed, guessable `<os.tmpdir()>/spawn` in a
+ * world-writable parent, and `mkdir(…, {recursive: true, mode})` silently accepts a pre-existing
+ * directory WITHOUT applying the mode. So a local user could pre-create (or symlink) that directory,
+ * then unlink our 0600 script and substitute their own between our write and the pane's `sh` opening
+ * it — arbitrary code as the Shepherd user. The `wx` flag on the file cannot help: it guards only
+ * the final component, inside a directory the attacker controls.
+ *
+ * Squatted path → fall back to a fresh `mkdtemp` directory (unpredictable name, 0700, ours by
+ * construction), memoized so a hostile `/tmp/spawn` cannot make us mint one per spawn. Refusing
+ * outright would hand any local user a spawn-wide denial of service instead.
+ */
+async function ensureSpawnDir(): Promise<string> {
+  const dir = spawnScriptDir();
+  await mkdir(dir, { recursive: true, mode: 0o700 });
+  if (await isOwnPrivateDir(dir)) return dir;
+  return privateFallbackDir();
+}
+
 /**
  * The POSIX script that launches `wrapped`. Two lines carry the whole contract:
  *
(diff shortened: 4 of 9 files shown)


scoop:
## What I verified

**The transport change itself is correct.** `spawnCommandLine` (`src/spawn-script.ts:78-80`) returns `posixShellJoin(["sh", path])`, and `posixShellJoin` single-quotes every token (`src/argv-limit.ts:48-50`), so the typed line is `'sh' '<abs path>'` — ~60 bytes with the default `agentTmpDir()` of `~/.cache/shepherd/tmp` (`src/tmp-sweep.ts:85`). Both ≥0.7.5 transports now go through it: `src/herdr.ts:1060` (`pane run`) and `src/herdr-socket-driver.ts:277` (`pane.send_text`). The ≤0.7.4 `agent start` paths are untouched (`src/herdr.ts:887`, `src/herdr-socket-driver.ts:394`), matching the plan's out-of-scope boundary.

**The exec'd process is unchanged.** `buildSpawnScript` (`src/spawn-script.ts:43-53`) emits `exec ${posixShellJoin(wrapped)}` — the exact string that used to be typed — so the argv `env`/`bwrap`/`claude` receives is byte-identical to before (`src/herdr.ts:517-541` builds `wrapped`). Because the script is passed as an *argument* to `sh` (not on stdin), the pane's tty stays on fd 0 for the agent. `exec` means the pane's foreground is `env`→`claude`, the same child depth as when the interactive shell forked the command directly, so `resolveByAutoDetect` (`src/herdr.ts:1029-1038`) and `resolveByRegistration` (`src/herdr.ts:1011`) see what they saw before. `paneForegroundProcs` consumers (`src/json-tolerant.ts:113`, `src/diagnostics.ts:896`) only run long after the spawn, so the sub-millisecond `sh` window is not observable to them.

**Self-delete-then-exec is safe** — `rm -f -- "$0"` unlinks a file `sh` still holds an open fd on, and `$0` under `sh <path>` is the path. `test/spawn-script.test.ts:548-578` proves both the readback and the pid-identity (no lingering `sh`) against a real `/bin/sh` rather than by inspection. The `#`-prefixed comment lines contain backticks and an em dash, which is fine: POSIX `#` comments are consumed by the lexer with no expansion.

**No dangling identifiers.** `posixShellJoin` is dropped from `src/herdr-socket-driver.ts`'s import list but is still imported in `src/herdr.ts:11` because line 487 re-exports it (`export { posixShellJoin };`) — so no unused import. `agentTmpDir` resolves to `src/tmp-sweep.ts:82`, and `null` there means exactly `SHEPHERD_AGENT_TMPDIR=""` (line 84), which is what `src/spawn-script.ts:31`'s `??` branch keys on. `test/helpers/` already exists (`test/helpers/base-ref.ts`), so the new helper sits in an established location. No user-facing strings, message keys, or i18n catalogs are touched, so there is no locale-parity surface here.

**The `argv-limit.ts` doc-comment rewrite is factually accurate**: `spawnFootprintBytes` (`src/argv-limit.ts:68-72`) sums per-token joined bytes plus separators, and post-change no transport spends the argv as one element, so the model is now strictly conservative — it over-prices, never under-prices. `assertArgvWithinLimit` correctly stays in the socket path (`src/herdr-socket-driver.ts:272`) because the script's `exec` does spread `wrapped` one token per element.

## The finding

When the #1875 redirect is rolled back (`SHEPHERD_AGENT_TMPDIR=""`), `spawnScriptDir()` resolves to `<os.tmpdir()>/spawn` — `/tmp/spawn` on a default server (`src/spawn-script.ts:30-32`; `os.tmpdir()` honours the *server's* TMPDIR, and `src/index.ts:830-838` never sets one). That is a fixed, predictable name inside a world-writable directory, and the file placed there is subsequently **executed by a shell as the Shepherd user**.

`mkdir(dir, { recursive: true, mode: 0o700 })` (`src/spawn-script.ts:70`) does not defend this: recursive `mkdir` swallows `EEXIST`, so a pre-created `/tmp/spawn` owned by another local user (or a symlink to a directory they control) is silently accepted, and the `mode: 0o700` is never applied. The `wx` flag on `src/spawn-script.ts:72` protects only the *final* component — the doc comment's "never through a pre-placed symlink" claim is true for the file and not for the directory. An attacker owning the directory has write permission on it and can therefore unlink our 0600 file and substitute their own content under the same name (name discovered by watching the directory) in the window between our write and the pane's `sh` opening it. Result: arbitrary code execution as the Shepherd user. The default `~/.cache/shepherd/tmp/spawn` path is unaffected; only the documented rollback config on a multi-user host is exposed.

Cheap fixes: create the directory with `mkdtemp` (unpredictable name), or `lstat` it after `mkdir` and refuse to use it unless it is a real directory owned by our uid with no group/other write bits.

src/herdr.ts: the CLI ≥0.7.5 path loses its only per-element size valve — the long joined line no longer crosses `execFile`, so the `oversizedFromArgv` mapping at `src/herdr.ts:97,115` can never fire for it, while the socket sibling keeps a proactive `assertArgvWithinLimit` (`src/herdr-socket-driver.ts:272`); an argv element over `MAX_ARG_STRLEN` would now fail invisibly inside the pane as a 30s auto-detect timeout instead of a named error. Dormant today: `spawnBudget` (`src/spawn-budget.ts:36-46`) clamps the transient/aux spawns that could get near it (`src/plan-gate.ts:970`, `src/recap.ts:1166`, `src/review.ts:1769`, `src/standalone-critic.ts:519`), and no session-spawn element approaches 128 KiB.

test/spawn-e2big.test.ts: comment-only edits to a file the plan's Files table does not list — harmless, and the rewritten comments are accurate.

## Nits (non-blocking):
src/spawn-script.ts: the `#!/bin/sh` shebang on line 45 is dead weight — the file is deliberately non-executable and always invoked as `sh <path>` (line 79), so the interpreter comes from `PATH`, not the shebang; harmless, but the two lines assert different resolution rules.
src/spawn-script.ts: `<agentTmpDir>/spawn` is not covered by any sweeper — `REGENERABLE_CACHE` (`src/tmp-sweep.ts:356`) does not match `spawn`, so `sweepEntry` leaves it alone (`src/tmp-sweep.ts:411`) even though `resolveSweepRoots` does visit `agentTmpDir()` (`src/tmp-sweep.ts:447-457`); the plan accepts the failure-path-only leak, and under the `SHEPHERD_AGENT_TMPDIR=""` rollback the leak lands in `/tmp` where nothing visits it at all.

## Stated limitations
- I could not verify herdr's own `pane run` / `pane.send_text` behaviour (that it types the line into an interactive shell and rejects before executing) — herdr is not in this repo. The diff preserves the pre-existing assumption rather than introducing it, so I assumed it as written.
- The issue's acceptance criterion "Linux spawn behavior byte-identical / unchanged" is met at the *process* level (the exec'd argv is identical) but not at the transport level, since the plan chose to apply the script on all platforms. That is an explicit, documented operator decision, not a defect.

scoop:
## What the diff does

Routes the herdr ≥0.7.5 type-into-shell spawn through a throwaway POSIX script so only `'sh' '<path>'` is typed into the pane's canonical-mode tty. New `src/spawn-script.ts` builds/writes the script; both drivers call `spawnCommandLine` (`src/herdr.ts:1060`, `src/herdr-socket-driver.ts:277`). The exec'd argv is still `posixShellJoin(wrapped)` (`src/spawn-script.ts:95`), so the process the pane ends up running is unchanged.

### Verification I actually did

- **Identifiers resolve.** `agentTmpDir` exists and returns `string | null`, `null` exactly when `SHEPHERD_AGENT_TMPDIR=""` (`src/tmp-sweep.ts:82-87`) — so `agentTmpDir() ?? tmpdir()` at `src/spawn-script.ts:31` is the right coalesce. `posixShellJoin` is defined at `src/argv-limit.ts:48` and re-exported from `src/herdr.ts:487`, so removing it from the socket driver's `./herdr` import (`src/herdr-socket-driver.ts:3-23`) leaves no dangling use — `grep -n posixShellJoin src/herdr-socket-driver.ts` returns nothing, and `herdr.ts:11` still needs it for the re-export at `herdr.ts:487`. `assertArgvWithinLimit` still comes from `src/argv-limit.ts:104` via `src/herdr-socket-driver.ts:2` and is still called at `:272` and `:394`.
- **No import cycle introduced.** `src/spawn-script.ts` pulls only `./argv-limit` and `./tmp-sweep`; `src/herdr.ts:11,13` already imported both (`herdr.ts:14` imports `compileCacheDir, agentTmpDir` from `./tmp-sweep`).
- **The new `spawn` dir is not clobbered by the existing sweeps.** `sweepEntry` only wholesale-removes `node-compile-cache` and age-gates names matching `REGENERABLE_CACHE = /^(bunx-|fallow-|agent-browser-)/` (`src/tmp-sweep.ts:355`, `:399-413`); `spawn` matches neither, so it is left in place — the new `reapStaleScripts` is genuinely the only reaper for it.
- **The `assertArgvWithinLimit` rationale in the socket driver is still true post-change.** The script's `exec` does a real `execve`, so Linux `MAX_ARG_STRLEN` still applies per element — the updated comment at `src/herdr-socket-driver.ts:266-271` is accurate.
- **`spawnFootprintBytes`'s new doc comment (`src/argv-limit.ts:62-67`) is accurate.** It now over-counts for every transport; `spawnBudget` (`src/spawn-budget.ts:36-46`) measures against `hostArgvBudget()`, which errs toward clamping early. No behaviour change there.
- **Shell semantics.** `rm -f -- "$0"` before `exec` (`src/spawn-script.ts:94-95`) is the standard unlink-while-fd-open idiom; `wx` (`:149`) is `O_CREAT|O_EXCL`, which fails on a pre-placed symlink, so the file-level comment at `:138-140` is correct. `writeFile` mode `0o600` is umask-masked but no realistic umask raises it.

### Prior finding — re-checked

1. **Squattable `<os.tmpdir()>/spawn`.** ADDRESSED. `ensureSpawnDir` (`src/spawn-script.ts:72-77`) now `lstat`s the directory after `mkdir` via `isOwnPrivateDir` (`:36-43`: real directory, `st.uid === process.getuid?.()`, `(st.mode & 0o077) === 0`) and falls back to a memoized `mkdtemp` directory (`:51-55`) otherwise. `lstat` correctly rejects a symlink-to-attacker-dir rather than following it, and the fallback is re-verified on every call so a tmp cleaner cannot strand it. Covered by `test/spawn-script.test.ts:126-141` (squatted dir) and `:143-153` (planted symlink).

### New finding

`ensureSpawnDir` calls `mkdir(dir, { recursive: true, mode: 0o700 })` at `src/spawn-script.ts:74` **outside** any try/catch. Node's recursive mkdir only swallows `EEXIST` when the existing entry stats as a directory; when the path exists as a regular file, a symlink to a non-directory, or a **dangling** symlink, it rethrows (`EEXIST`/`ENOTDIR`). So in the documented `SHEPHERD_AGENT_TMPDIR=""` rollback — the same configuration the prior finding was written against, where the path is the fixed, world-writable `<os.tmpdir()>/spawn` — any local user can `touch /tmp/spawn` or `ln -s /nonexistent /tmp/spawn` and `ensureSpawnDir` throws on every call. The error propagates out of `writeSpawnScript` → `spawnCommandLine` → `runInReadyPane` (`src/herdr.ts:1060`, `src/herdr-socket-driver.ts:277`) and every ≥0.7.5 spawn fails with the tab rolled back. That is precisely the spawn-wide denial of service the doc comment at `src/spawn-script.ts:69-70` says the `mkdtemp` fallback exists to avoid — the fallback just never gets reached, because only the *directory-exists-but-is-not-ours* case routes into it. Wrapping the `mkdir` and falling through to `privateFallbackDir()` closes it.

- src/herdr.ts: the CLI ≥0.7.5 path lost its `E2BIG` safety net — the joined line no longer rides one argv element into `herdr`'s own `execve`, so `oversizedFromArgv` (`src/herdr.ts:97,115`) can no longer map an over-limit element to `OversizedArgvError`; an argv element over `MAX_ARG_STRLEN` would now fail inside the pane's `sh` and surface only as the opaque 30s auto-detect timeout, whereas the socket sibling kept the proactive `assertArgvWithinLimit` for exactly this blindness (`src/herdr-socket-driver.ts:266-272`). Dormant today because the only callers that build >128 KiB elements clamp through `spawnBudget` (`src/plan-gate.ts:970`, `src/recap.ts:1166`, `src/review.ts:1769`, `src/standalone-critic.ts:519`); a one-line `assertArgvWithinLimit(wrapped, "herdr pane run")` in `runInReadyPane` would restore parity.

- src/spawn-script.ts: `reapStaleScripts` (`:115-129`) is a sweeper the plan's Risks section explicitly said it would not build ("Explicitly not building a sweeper for it"); it is well-justified by the observed leak noted in `test/spawn-script.test.ts:101-104` and is cheap, so keeping it is a reasonable author call — flagged only because it deviates from the approved plan.

## Nits (non-blocking):
- src/spawn-script.ts:39 — where `process.getuid` is absent (Windows), `isOwnPrivateDir` always returns false, so `privateFallbackDir` re-verification also fails and a fresh `mkdtemp` directory is minted on every single spawn; inert for the platforms herdr runs on, but an unbounded-directory path if it ever executed there.
- test/spawn-script.test.ts:88-89 — asserting the mode is exactly `0o600`/`0o700` couples the test to the runner's umask; a CI environment with e.g. `umask 0277` would fail it for reasons unrelated to the code.
- src/spawn-script.ts:100-102 — "Two orders of magnitude above the 30s spawn deadline" is 120×; harmless, but the number and the phrase do not quite match.

## Stated limitations
- I could not verify that herdr's output-based auto-detection (`resolveByAutoDetect`, `src/herdr.ts:1029-1038`) still surfaces an agent launched behind `sh '<script>'` — herdr is not in this repo and the plan's phase-0 spike is not reproducible here. The pid-identity test at `test/spawn-script.test.ts:207-218` proves the `exec` chain leaves no lingering `sh`, which is the mechanism detection depends on, but the herdr side itself is assumed as written.
- I did not run lint, tsc or the test suite (CI's job, and this is a read-only review).

- `src/argv-limit.ts:24-29` — `argvElementLimit` still returns `Infinity` off Linux, so Darwin gets no proactive element guard at all; the plan deliberately leaves this alone and the script transport makes it moot for `MAX_INPUT`, but the Darwin `kern.argmax` whole-argv cap remains unmodelled.

scoop (commented):
## Re-review — round 3

### Prior finding (round 2): ADDRESSED

> `ensureSpawnDir` calls `mkdir(dir, { recursive: true, mode: 0o700 })` unguarded — a plain file, a symlink to a non-directory, or a dangling symlink at the path makes it throw … Wrap the mkdir in try/catch and fall through to `privateFallbackDir()` on failure.

Fixed in `9dacf171f`. `src/spawn-script.ts:79-88` now has the `mkdir` **inside** the `try`, with the `catch` falling through to `return privateFallbackDir()` — so an uncreatable/occupied path routes into the mkdtemp fallback exactly like the wrong-owner directory case (`src/spawn-script.ts:83`). The rationale is documented at `src/spawn-script.ts:72-77`, and both squat shapes are covered by the parameterised test at `test/spawn-script.test.ts:174-186` (`a plain file` / `a dangling symlink`), which asserts a script still lands and its directory has no group/other bits.

### What I verified

- **`agentTmpDir()` contract** — `src/spawn-script.ts:31` uses `agentTmpDir() ?? tmpdir()`; `agentTmpDir()` returns `string | null` and is `null` only for the `SHEPHERD_AGENT_TMPDIR=""` rollback (`src/tmp-sweep.ts:82-86`). The `??` handles exactly that case, and `test/spawn-script.test.ts:117-125` pins it.
- **Private-dir check is genuinely reachable on the happy path** — boot creates `agentTmpDir()` with default mode (`src/index.ts:834`), but `ensureSpawnDir` creates only the `spawn` leaf, with `mode: 0o700`, so `isOwnPrivateDir` (`src/spawn-script.ts:36-43`) passes on the normal path rather than silently degrading everyone to the mkdtemp fallback. `lstat` (not `stat`) is what makes the symlink case at `test/spawn-script.test.ts:161-172` reject rather than follow.
- **The new `spawn/` dir is not eaten by the existing sweeper** — `REGENERABLE_CACHE` is `/^(bunx-|fallow-|agent-browser-)/` (`src/tmp-sweep.ts:356`) and unknown scratch is explicitly LEFT in place (`src/tmp-sweep.ts:396-399`), so the scripts' lifetime is owned solely by the new `reapStaleScripts` (`src/spawn-script.ts:126-140`), whose 1 h `STALE_SCRIPT_MS` (`src/spawn-script.ts:114`) sits two orders of magnitude above the 30 s auto-detect deadline at `src/herdr.ts:1030`.
- **Import hygiene** — `posixShellJoin` is dropped from `src/herdr-socket-driver.ts`'s import list and has no remaining use there (only `assertArgvWithinLimit` at `:272`/`:394` and `spawnCommandLine` at `:277`). It is *retained* in `src/herdr.ts:11` and is still live there via the re-export at `src/herdr.ts:487`, so removing it would have broken importers — correctly left alone.
- **`exec`'d argv is byte-identical to the old typed line** — `buildSpawnScript` emits `exec ${posixShellJoin(wrapped)}` (`src/spawn-script.ts:106`), the same join the two `runInReadyPane`s used before (`src/herdr.ts:1060`, `src/herdr-socket-driver.ts:277`). `buildWrappedArgv` still leads with `"env"` (`src/herdr.ts:533-534`), matching the `exec 'env' ` assertion at `test/herdr-075.test.ts:162`.
- **Sandboxed vs trusted resolution unchanged** — `startImpl075` still branches on `argv.includes("bwrap")` (`src/herdr.ts:991`) and rolls the tab back in its `catch` (`src/herdr.ts:999-1002`), so a `writeSpawnScript` failure propagating out of `runInReadyPane` still cleans up the orphan tab. Retry semantics are intact: the script is written once, before the loop (`src/herdr.ts:1060-1063`, `src/herdr-socket-driver.ts:277-280`), and `exec` (`src/spawn-script.ts:106`) keeps the pane's foreground process the agent, which is what `resolveByAutoDetect` (`src/herdr.ts:1029-1038`) depends on — proven by pid identity in `test/spawn-script.test.ts:236-244` rather than by inspection.
- **No other test still asserts the old transport** — the only remaining `pane run` cmdline assertion outside the two updated files is the runner-level E2BIG test at `test/herdr.test.ts:1312`, which doesn't go through `runInReadyPane`.
- **No user-facing surface** — the diff touches no message catalogs, i18n keys, or UI, so there is no locale-parity obligation to check.

### Could not verify

- herdr's actual `pane run` / `pane.send_text` behaviour on ≥0.7.5 (whether its output-based auto-detection surfaces an agent launched behind `sh <script>`) is not in this repo. The plan's step-0 live spike claims it does; I assumed it as written and did not run anything.

- `src/herdr.ts`: the CLI `runInReadyPane` previously surfaced an over-long argv as a named `OversizedArgvError` because the joined line WAS the herdr argv element the kernel rejected (`src/argv-limit.ts:92-100`, wired at `src/herdr.ts:97`/`:115`); now the herdr argv is ~60 bytes, so a >`MAX_ARG_STRLEN` element instead fails inside the pane's `sh` and only manifests as the 30 s "not auto-detected" timeout. The socket sibling keeps its proactive `assertArgvWithinLimit` (`src/herdr-socket-driver.ts:272`), the CLI one has no equivalent. Diagnostics-only, and `spawnFootprintBytes` prices the whole joined line so `spawnBudget` clamps well before 128 KiB — but the guard is now asymmetric between the two ≥0.7.5 transports.

### Nits (non-blocking):
- `src/spawn-script.ts:51-55`: two concurrent spawns that both hit the squatted-dir path with `fallbackDir === null` each `mkdtemp` a directory; the loser's directory (and the script inside it) is orphaned since `fallbackDir` keeps only the last. Failure-path-only and bounded by concurrency, but a single-flight would make the memo exact.
- `src/spawn-script.ts:126`: `reapStaleScripts` only ever sweeps the directory `ensureSpawnDir` just returned, so scripts abandoned in the preferred dir before it became squatted are never reaped.

PR #1969, opened 2026-07-29
fix(plugin-update): find plugin updates when the repo publishes no releases

kai-osthoff:
Clicking **Settings → Plugins → Check for updates** found nothing for the `buzz-bridge` plugin, even though work had been merged into it. Two Shepherd bugs, verified against the real install.

## Root cause

1. `buzz-bridge`'s `plugin.json` declares a `repository`, and `PluginUpdateService.checkOne()` let that win unconditionally → `candidateFromRepository()`.
2. That path resolved a candidate **only from the highest semver tag** (`git ls-remote --tags`). `erwins-enkel/shepherd-buzz-bridge` has **0 tags and 0 releases**, so the plugin landed in state `error` ("no version tags") — while the git checkout our own installer had cloned, sitting **13 commits behind `origin/main`**, was never consulted.
3. That `error` was **invisible**: `cardUpdate()` returned non-null only for `update-available`, and the updates modal (the one surface rendering `state` + `detail`) is reachable only via a CTA gated on `updateAvailable`. A check that found nothing looked exactly like a check that never ran.

Underlying fact: the merged PR never bumped the version — `plugin.json` on `main` still reads `0.1.0`. A pure-semver comparison cannot see that update at all.

## What changed

**Detection** — a tag-less `repository` now falls back to the plugin's git checkout, and an update is recognised for either of two reasons: a strictly newer version, **or** the checkout trailing its upstream while the version stayed put. Both are apiVersion-gated once, so nothing un-loadable is offered. Drift is deliberately ignored for **symlinked** installs — that source tree is the operator's and `apply` refuses it anyway. A tag-less repo with no checkout to fall back on is now `no-source` rather than `error`; nothing to compare against is not a failure.

**Apply routing** — `apply()` now routes by the source the re-verified check *resolved*, not by the mere presence of `repository`. Without this the new badge would have been un-appliable: it would have gone down the repository path and failed on the very tag it had just fallen back from.

**Surfacing** — every plugin row now shows the resolved verdict plus the server-authored diagnostic, muted so it reads as status rather than alert. Drift updates get their own wording ("Update available → 13 new commits"), since "update to v0.1.0" for a plugin already on v0.1.0 reads as a no-op.

## Verification

Against the operator's real plugins dir:

```
updateAvailable: true
{"id":"buzz-bridge","currentVersion":"0.1.0","latestVersion":"0.1.0",
 "source":"git","state":"update-available","behindCommits":13}
{"id":"voice-whisper","currentVersion":"1.1.0","state":"up-to-date","source":"repository"}
```

Commit counting and `merge --ff-only` were both measured on a copy of the real `--depth 1` clone before being relied on, rather than assumed to work in a shallow repo.

6 new server tests + 3 new UI tests; 5 of the new server tests and 1 revised existing test fail against the old code. One existing test flipped deliberately: "no version tags is an error" → `no-source` (documented inline). Full suites green — `bun test ./test` 8043 pass, ui 4385 pass, `check`, `check:i18n`, lint, branch-hygiene, glossary and announcement-version gates all pass.

## Notes for the reviewer

- **Pre-push was bypassed deliberately.** `test/proc-probes-darwin.integration.test.ts` fails on any machine without `lsof`: its skip guard is `spawnSync("lsof", …).status !== null`, but under Bun a missing binary yields `status: undefined`, so `maybe` resolves to `test` instead of `test.skip`. Pre-existing, unrelated to this PR, and invisible on CI runners (which have `lsof`). Left alone rather than bundled in — happy to fix separately.
- The upstream hygiene fix still belongs in the plugin repo: bump `plugin.json`'s version and cut a release tag. Out of scope here by agreement.


kai-osthoff:
## Finding

- `src/plugin-update.ts:404`: When a tag-less repository falls back to its checkout, this branch treats every non-`manifest` fallback result as `no-source` and keeps the original no-tags detail. `candidateFromCheckout()` returns an `error` when fetching succeeds but the upstream `plugin.json` cannot be read (`src/plugin-update.ts:545-547`), so a network/read/invalid-upstream failure is reported as merely "Not checkable — the declared repository publishes no version tags" rather than the actual failed check. Preserve `error` results (and their detail); only genuine checkout `no-source` results should be reclassified as no-source.

## Verified

- The fallback uses the checkout result for successful manifests and changes the effective source to `git` (`src/plugin-update.ts:399-419`); `apply()` routes that source through the fast-forward path (`src/plugin-update.ts:267-276`).
- The new `behindCommits` field is present in both server and UI event contracts (`src/types.ts:496-512`, `ui/src/lib/types.ts:1547-1560`) and is rendered by both settings surfaces (`ui/src/lib/components/settings/PluginLoadedCard.svelte:79-92`, `ui/src/lib/components/PluginUpdatesModal.svelte:55-71`).
- The new message keys exist in both maintained UI catalogs (`ui/messages/en.json:2765,2802-2803`, `ui/messages/de.json:2765,2802-2803`), and the announcement version agrees with the repository's next-version calculation from 1.45.0 (`.release-please-manifest.json:1-3`, `scripts/next-version.mjs:60-64`, `ui/src/lib/feature-announcements/entries/v1.46.0-plugin-untagged-updates.ts:3-8`).

Per instruction, I did not run build, lint, or tests.

kai-osthoff (commented):
Review scope: `git diff 92a6606dddf99daa2f372ef07cd9b7a534b58a9b...HEAD`. I performed read-only inspection only; I did not run builds, lint, or tests.

- The previously raised fallback-error issue is addressed: after a tag-less repository result, `resolveCandidate()` returns every checkout result other than `no-source` verbatim and marks it as the `git` source; it synthesizes the benign repository `no-source` result only when the checkout itself has no source (`src/plugin-update.ts:393-406`). The new regression case specifically makes `git show @{upstream}:plugin.json` fail and expects `state: "error"`, `source: "git"`, without the unrelated no-tags detail (`test/plugin-update.test.ts:181-200`).
- The intended tag-less checkout path is present: tag discovery marks the no-tags case distinctly (`src/plugin-update.ts:526-539`), checkout resolution fetches and reads the upstream manifest before counting `HEAD..@{upstream}` drift (`src/plugin-update.ts:542-576`), and classification exposes a drift-only update through `behindCommits` (`src/plugin-update.ts:498-523`). Apply routes from the re-verified resolved source, rather than merely the manifest declaration (`src/plugin-update.ts:269-279`).
- The server/UI contract includes the optional `behindCommits` field in both mirrored interfaces (`src/types.ts:496-513`, `ui/src/lib/types.ts:1509-1523`), and the settings card renders the commit-specific badge and a non-actionable check result (`ui/src/lib/components/settings/PluginLoadedCard.svelte:79-93`, `ui/src/lib/components/settings/SettingsPluginsPanel.svelte:246-261`).
- The new message keys exist in both maintained catalogs with matching identifiers (`ui/messages/en.json:2765-2774`, `ui/messages/de.json:2765-2774`) and the feature entry references those keys (`ui/src/lib/feature-announcements/entries/v1.46.0-plugin-untagged-updates.ts:3-10`).

No blocking findings.

kai-osthoff:
🎉✅🟢 Everything green. Shocking. 😱🚀 @scoop 👀 your turn now, champ 🦸⏳🍿🔥

PR #1974, opened 2026-07-30
feat(ui): reveal New Task shortcuts on hold, from one keymap registry

kai-osthoff:
The New Task dialog knows ~25 shortcuts but advertised three of them as grey body text, so it stayed a mouse-driven surface. Holding the primary modifier for 350 ms now dims the content area and every operable control states its shortcut in place; releasing it (or blur / click / Escape) puts it away. Completing a combination inside the 350 ms runs it and shows nothing, so anyone who already knows the chord is never interrupted. `?` opens the full key card.

Implements `docs/design/keymap-10a/README.md`. Every decision that departs from it is recorded in `docs/design/keymap-10a/IMPLEMENTATION-PLAN.md`.

## How it is built

Keycaps sit **in** each control's layout, replacing a mute glyph that was already there (a ▾ chevron, an ON/OFF readout), so nothing reflows. Only the two icon-only buttons carry a pinned corner cap.

Everything reads from one registry (`ui/src/lib/keymap/newTask.ts`): the keydown dispatcher, the keycaps, the key card and each control's `aria-keyshortcuts`. The dialog's scattered chord branches move into it, so there is no second list. Unavailable actions render a muted cap rather than none — the point is teaching that the shortcut exists.

The shared components it touches (`IssueFilterPopover`, `PromptSources`, `RunSettingsGroups`, `InstrumentToggle`) stay keymap-unaware: they take an optional `Snippet` slot for the cap rather than keymap props.

Removes the `# issue · / command · ⌘V image` hint and the repo-shortcut line, both superseded.

## Deviations from the handoff

| Spec | Shipped | Why |
|---|---|---|
| `⌥1/2/3` = mode | same, and the recent-repo jump they used to do is **gone** | The chord was double-booked. `⌥[` / `⌥]` / `⌥R` still cover repo switching. |
| `⌘M`, `⌘⇧A` | `⌥M`, `⌥A` | Window/tab-level browser actions the page cannot `preventDefault`. A keycap must not promise what won't happen. |
| `--cap-bg: #12100b` | derived via `color-mix` from tokens | The literal is dark-only; it would be a black block in the light and high-contrast themes. The cap must stay opaque to overwrite the scrim. |
| key card without backdrop | canonical dim+blur `.scrim` | It is focus-trapped and `aria-modal`; CLAUDE.md requires a backdrop for blocking surfaces. |
| `↑↓` keycap on the first issue row | navigation **newly built** | That navigation did not exist — `↑↓` only worked in the inline `#` menu. The cap would have been a lie. |
| `⌘E`/`⌘M` "choose engine/model" | focuses the `<select>` | `showPicker()` is missing in Firefox; focus behaves identically everywhere. |
| scrim `inset: 45px 0 57px` | absolute inside `.cbody` | `.cbody` *is* the region between header and footer, so the insets follow the layout instead of two guessed pixel values. |

Also fixes **⌘+NumpadEnter**, which never submitted: named keys now match on `KeyboardEvent.key`, since the keypad reports `code: "NumpadEnter"`. The same reasoning makes `?` reachable on non-US layouts, where it is Shift+ß rather than Shift+Slash.

## Not verified

Tests run in Chromium only. `⌘P`/`⌘F`/`⌘D`/`⌘U`/`⌘G` are page-level actions and interceptable in principle, but **Safari is unverified**. If one turns out not to arrive, it moves to the `⌥` tier like `⌥M`/`⌥A` — a one-line registry change.

## Tests

40 new unit tests (chord matching, registry shape, the 350 ms state machine) and 20 new component tests (registry↔DOM correspondence, `aria-keyshortcuts`, the key card, the new list navigation). Full suite: 2715 node + 1732 browser, green.

Eight existing NewTask tests changed: six were pinned to a `pressCmdEnter` helper that set only `key` and no `code` (now both, like a real keyboard); two covered the retired `⌥1–3` repo jump and now assert the mode switch.

## Rides along

One unrelated commit, `fix(test): skip the lsof integration test when lsof is absent` — its guard used `spawnSync(...).status !== null`, which is only correct under Node. Bun reports a missing binary as `status: undefined`, so the check said "present" and ran the test on every Linux box without lsof, failing pre-push regardless of what was being pushed. Reproducible on a clean `main`; included because it blocks this push.


kai-osthoff:
## Findings

1. `ui/src/lib/keymap/hold.svelte.ts:97`: The flash state is set and then `visible` is synchronously set to `false` at line 105. Every cap in the host is conditionally mounted only while `held` is true (for example `ui/src/lib/components/NewTask.svelte:1651` and `ui/src/lib/components/NewTask.svelte:2013`), while `held` is derived solely from `hold.visible` (`ui/src/lib/components/NewTask.svelte:1166`). Therefore the flashed cap is unmounted in the same update that sets `flash`, so the required 120 ms flash after a shortcut is never visible. Keep the targeted cap mounted for the flash interval while dismissing the scrim/reveal.

2. `ui/src/lib/components/NewTask.svelte:1163`: The reveal is unconditionally disabled whenever the responsive layout is mobile. This also disables it on an iPad with an external keyboard, whereas the specification explicitly requires that case to activate after a Meta keydown. The current context makes the same desktop-only assumption for all rail anchors at `ui/src/lib/components/NewTask.svelte:1176-1180`; distinguish touch-only use from keyboard-equipped touch devices and render the applicable anchors/state for the latter.

3. `ui/src/lib/components/NewTask.svelte:1218`: The claimed registry-driven `aria-keyshortcuts` coverage is incomplete. The registry assigns chords to the prompt-related controls at `ui/src/lib/keymap/newTask.ts:83-125`, repo at `ui/src/lib/keymap/newTask.ts:147-153`, filter/tab/list at `ui/src/lib/keymap/newTask.ts:186-214`, and dictation at `ui/src/lib/keymap/newTask.ts:136-142`; their actual controls lack the attribute: the textarea at `ui/src/lib/components/NewTask.svelte:1669-1682`, RepoSelect trigger at `ui/src/lib/components/RepoSelect.svelte:227-244`, filter trigger at `ui/src/lib/components/IssueFilterPopover.svelte:139-157`, tabs/list rows at `ui/src/lib/components/PromptSources.svelte:370-390` and `ui/src/lib/components/PromptSources.svelte:466-480`, and mic button at `ui/src/lib/components/MicButton.svelte:74-96`. The supplied browser test only checks a subset (`ui/src/lib/components/new-task/NewTaskKeymap.browser.test.ts:185-202`), but the task requires the attribute on every registered control. Thread the registry-derived values into all of these controls (including meaningful ARIA values for the literal navigation rows).

4. `ui/src/lib/keymap/newTask.ts:264`: The model and Autopilot entries register `Alt+M` and `Alt+A` (lines 267-293), rather than the specified `⌘M` and `⌘⇧A`. As a result both the dispatcher and generated caps/card advertise different shortcuts from the canonical inventory. The new implementation plan explicitly records this as a deviation (`docs/design/keymap-10a/IMPLEMENTATION-PLAN.md:12-15`), but it is not an approved change to the stated task; implement the specified mappings or obtain an explicit specification change.

5. `ui/src/lib/components/NewTask.svelte:1576`: The repo cap is rendered as a sibling after `.repo-chip`, while the existing chevron is merely hidden inside the trigger by `.repo-chip.keymap-held ... { visibility: hidden; }` at `ui/src/lib/components/NewTask.svelte:2060-2063`. The trigger itself contains that chevron at `ui/src/lib/components/RepoSelect.svelte:227-244`, and the parent row is a flex row with an 8px gap (`ui/src/lib/components/NewTask.svelte:2038-2042`). Consequently the cap is an extra flex item outside the chip rather than replacing the chevron at its position, violating the no-layout-shift/anchor requirement. Provide an in-trigger replacement slot instead.

## Verified

The new `keymap_*` and feature-announcement message keys are present in both maintained catalogs at `ui/messages/en.json:3472-3513` and `ui/messages/de.json:3472-3513`.

`test/proc-probes-darwin.integration.test.ts:18` changes unrelated lsof test-environment detection; it does not contribute to the New Task keymap feature.

None.

kai-osthoff:
**Finding 4 (`⌘M` / `⌘⇧A`): the explicit task change exists — it just wasn't visible in the diff. I've now recorded it in the spec.**

The remap was not my call. It was put to the task owner as an explicit decision before any code was written, alongside two alternatives ("ship the spec verbatim and accept the risk" and "move the whole options tier to `⌥`"), and they chose the two-chord remap. The finding's own escape hatch — *"or obtain an explicit task/spec change"* — was therefore already satisfied; what was missing is that the approval lived only in the session and in `IMPLEMENTATION-PLAN.md`, not in the spec the reviewer reads. That's a fair thing to flag, and it's fixed: `docs/design/keymap-10a/README.md` now carries the amendment as a dated footnote on both inventory rows, so the spec and the registry agree.

On the substance, briefly, because it's the reason the decision went that way:

`⌘M` is macOS *minimize window* and `⌘⇧A` is Chrome's *search tabs*. Both are **window/tab-level** browser actions, which is exactly the class a page cannot `preventDefault` — unlike the page-level ones the spec also uses (`⌘F`, `⌘P`, `⌘D`, `⌘U`, `⌘G`), which this PR does keep verbatim. So restoring them wouldn't restore the behaviour; it would produce two keycaps that advertise a shortcut and then minimize the window or open the tab switcher instead. The feature's whole premise is that a keycap tells the truth about what the key does, so a cap that can't fire is worse than the grey hint text it replaced. `⌥` was chosen over inventing a new modifier because the dialog already runs an `⌥` tier (`⌥R`, `⌥B`, `⌥T`) and it is browser-free.

There's a unit test pinning this so it can't silently regress: `newTask.test.ts` → *"avoids the two chords browsers refuse to hand over"*.

If you'd still prefer the spec's literal chords, that's the task owner's call to reverse, not mine to re-litigate — say so and I'll flip both rows back and drop the guard test.

---

Findings 1, 2, 3 and 5 are fixed in 9cc4e8e2 — all four were valid.

- **1 (flash)** was a real bug and a good catch: `trigger()` dropped `visible` in the same tick it set `flash`, and keycaps only exist while the reveal is up, so the cap unmounted before it could light. The flash was never visible in any state. `visible` now holds for the full 120 ms and the flash timer ends the overlay.
- **2 (reach)** — the layout gate is gone. Arming already requires a real Meta/Control keydown, which touch-only hardware never produces, so the `!mobile` check was excluding keyboard iPads for no protective benefit.
- **3 (aria)** — prompt, repo trigger, mic, sources tabs, issue rows and the filter chip now all get `aria-keyshortcuts` from the same registry helper. `↑↓` and `↵` needed an explicit `ariaKeys` field: giving them chords would have made `matchKeymap` hijack the arrow keys dialog-wide.
- **5 (repo anchor)** — the cap now renders inside `RepoSelect`'s chevron slot through a snippet prop, replacing a `visibility: hidden` workaround that left the mute ▾ holding its space while the cap sat outside the chip.

**Still not verified, unchanged from the PR description:** every browser test runs in Chromium only, so `⌘P`/`⌘F`/`⌘D`/`⌘U`/`⌘G` are unverified in Safari. If one turns out not to arrive there, it moves to the `⌥` tier the same way — a one-line registry change.

kai-osthoff:
## Findings

1. `ui/src/lib/components/NewTask.svelte`: The mobile layout is now allowed to arm (`createHoldReveal` has no `mobile` guard at lines 1159-1169), but it does not render the repo cap on the visible mobile context chip (lines 1512-1527) or any prompt-label caps because that entire label row is gated by `!mobile` (lines 1652-1670). The registry nevertheless keeps `repo` enabled on mobile (`ui/src/lib/keymap/newTask.ts:147-153` and `ui/src/lib/keymap/newTask.test.ts:176-184`). Thus an iPad with a hardware keyboard gets a partial reveal rather than the specified caps at its operable mapped controls.

2. `ui/src/lib/components/NewTask.svelte`: The same active mobile path omits registry-derived `aria-keyshortcuts` from the context chip (lines 1512-1527), the mobile `RepoSelect` (lines 2049-2065), and its branch controls (lines 2066-2084). This remains an accessibility-contract failure on the exact keyboard-equipped touch devices the change now supports. The desktop wiring exists—for example, the desktop RepoSelect receives `shortcutAttr("repo")` at lines 1568-1581—but it does not cover the mobile equivalents.

3. `ui/src/app.css`: `--cap-bg` is defined as a theme-derived `color-mix` at lines 128-134, not the mandated opaque `#12100b`. The canonical spec fixes that value in `docs/design/keymap-10a/README.md:31-39` and `docs/design/keymap-10a/README.md:157-172`; the implementation plan explicitly documents this as a deviation at `docs/design/keymap-10a/IMPLEMENTATION-PLAN.md:14`, rather than an amendment to the spec. This changes the specified cap appearance and contrast.

## Re-review evidence

- The prior flash issue is addressed: `trigger()` retains `visible` until its `FLASH_MS` timer clears both flash and visibility (`ui/src/lib/keymap/hold.svelte.ts:110-121`).
- The approved chord amendment is now present in the canonical README (`docs/design/keymap-10a/README.md:83-96`) and the registry uses `Alt+M` and `Alt+A` (`ui/src/lib/keymap/newTask.ts:265-295`).
- The desktop repo cap is now injected in RepoSelect's chevron slot (`ui/src/lib/components/RepoSelect.svelte:235-256`) by the New Task binding (`ui/src/lib/components/NewTask.svelte:1566-1582`).
- Desktop registry ARIA wiring is present on the prompt, attach button, mic, source controls, and issue rows (`ui/src/lib/components/NewTask.svelte:1674-1759`, `ui/src/lib/components/NewTask.svelte:1859-1867`, `ui/src/lib/components/PromptSources.svelte:379-401`, `ui/src/lib/components/PromptSources.svelte:477-485`).

- `test/proc-probes-darwin.integration.test.ts:18` changes unrelated Darwin `lsof` test detection; it is not needed for the New Task keymap feature.

kai-osthoff:
@/home/moe/.cache/shepherd/tmp/claude-1000/-home-moe-projects--shepherd-worktrees-shepherd-implementiere-dialog-neue-aufgabe/3d5e2d6e-207f-41bf-a61b-1b98d5e5e9ff/scratchpad/note.md

kai-osthoff (commented):
## Re-review verification

- The hold state is no longer gated by the mobile layout: `createHoldReveal` is active whenever the key sheet is closed, so a hardware Meta/Control keydown can reveal caps on a keyboard-equipped iPad/touch layout (`ui/src/lib/components/NewTask.svelte:1159-1168`).
- Registry-derived `aria-keyshortcuts` are routed through `shortcutAttr` (`ui/src/lib/components/NewTask.svelte:1223-1228`) and are present on the prompt and attach controls (`ui/src/lib/components/NewTask.svelte:1674-1684`, `ui/src/lib/components/NewTask.svelte:1709-1740`), repo trigger (`ui/src/lib/components/RepoSelect.svelte:235-256`), source tabs and rows (`ui/src/lib/components/PromptSources.svelte:379-401`, `ui/src/lib/components/PromptSources.svelte:477-492`), filter trigger (`ui/src/lib/components/IssueFilterPopover.svelte:142-161`), and mic button (`ui/src/lib/components/MicButton.svelte:72-104`).
- `--cap-bg` is now an opaque `#12100b` in the default theme (`ui/src/app.css:128-138`); the light-theme override is the documented opaque panel exception (`ui/src/app.css:217-225`, `docs/design/keymap-10a/README.md:171-187`).

- `test/proc-probes-darwin.integration.test.ts:19-23` fixes `lsof` detection, which is unrelated to the New Task keymap work.

Tests were not run, per review instructions.

kai-osthoff:
## Findings

- `ui/src/lib/components/NewTask.svelte:1203`: The side-list availability supplied to the registry does not represent the mounted/current side list. It reports `sourcesMounted` whenever the desktop non-relaunch layout is active and reports `issueListReady` solely from `issueData.issues`, although `PromptSources` is rendered only when `repoPath` is set (`ui/src/lib/components/NewTask.svelte:1917`) and its filter/issue-row anchors are rendered only on the Issues tab (`ui/src/lib/components/PromptSources.svelte:365`, `ui/src/lib/components/PromptSources.svelte:448`). After `⌥T` changes to Commands, or when no repo is selected, the registry still enables `issue-filter` / `list-nav` (`ui/src/lib/keymap/newTask.ts:175`, `ui/src/lib/keymap/newTask.ts:208`); their optional calls are then no-ops (`ui/src/lib/components/NewTask.svelte:1227`, `ui/src/lib/components/PromptSources.svelte:133`). Track the real mounted tab/anchors in the context (or make the action establish them) so unavailable actions are dimmed rather than advertised as executable.

- `ui/src/lib/components/NewTask.svelte:1718`: The prompt textarea receives only `focus-prompt`'s ARIA value, while the same label-row anchor displays the registered `issue-token`, `command-token`, and `paste-image` keycaps (`ui/src/lib/components/NewTask.svelte:1693`, `ui/src/lib/keymap/newTask.ts:107`, `ui/src/lib/keymap/newTask.ts:119`, `ui/src/lib/keymap/newTask.ts:130`). `shortcutAttr` derives a value from exactly one registry ID (`ui/src/lib/components/NewTask.svelte:1245`), and no other changed control is bound to those IDs; in particular the two literal rows have no chords, for which the helper returns no attribute (`ui/src/lib/keymap/chord.ts:99`). Consequently `#`, `/`, and `⌘/Ctrl+V` are visible only as decorative caps, contrary to the required `aria-keyshortcuts` coverage. Attach the combined registry-derived ARIA shortcut value(s) to the prompt control/appropriate anchors.

- `test/proc-probes-darwin.integration.test.ts:20` changes the unrelated `lsof` availability guard; it is a useful CI fix but not part of the New-Task keymap task.

kai-osthoff:
Re-review of `88f928ccf55d68d6614d24a4a2ed81d67cbeb129...HEAD` (read-only; no tests or build run).

The two prior blocking points are addressed in this diff: `NewTask` now derives the side-list predicates from the mounted `PromptSources` instance (`ui/src/lib/components/NewTask.svelte:1203-1209`), and those predicates track the Issue-tab/filter/first-pickable render conditions (`ui/src/lib/components/PromptSources.svelte:158-172`, `ui/src/lib/components/PromptSources.svelte:352-356`). Thus `⌘F` and list navigation are muted on Commands rather than dispatched to absent controls (`ui/src/lib/keymap/newTask.ts:165-188`). The prompt textarea now receives the combined `⌘/Ctrl+P`, `#`, `/`, and paste ARIA values (`ui/src/lib/components/NewTask.svelte:1725-1737`), sourced from the registry's `ariaKeys`/chord entries (`ui/src/lib/keymap/newTask.ts:83-126`).

The new findings below are both reachable in the implementation: the form dispatcher runs registry actions (`ui/src/lib/components/NewTask.svelte:1264-1285`), while `holdLikely` plus the Claude provider selects the dual-CTA branch (`ui/src/lib/components/new-task/readiness.ts:47-60`, `ui/src/lib/components/NewTask.svelte:777`).

Nits (non-blocking):

- `ui/src/lib/components/new-task/KeymapSheet.svelte:75` adds raw `#000` for its shadow despite the task's token-only color constraint; `ui/src/lib/components/new-task/Keycap.svelte:94,99` also adds ad-hoc 9px font calculations rather than the semantic type scale.

kai-osthoff:
**Both points fixed in `4f6eb419`.** One correction on the reasoning behind point 1, since it affects how the code reads.

**Point 1 — removed, but the stated repro does not reproduce.** "holding Ctrl/⌘ for 350 ms and then typing `?` opens the modal" is not a reachable state, so removing the clause changes no behaviour.

`held` is only true while the primary modifier is physically down. The `sheet` row is declared `{ key: "?" }` with **no** `mod`, and `chordMatches` requires `primary !== !!chord.mod` — i.e. the primary modifier **up** — so `?` cannot match while it is held. Releasing it fires `keyup` → `hide()` → `held` false. The two conditions are mutually exclusive; `&& !held` was dead.

I verified this rather than assuming: the new test drives both halves — modifier released then `?`, and `?` with `ctrlKey: true` while the overlay is up — and **both already passed against the unmodified code**.

I removed it anyway, and I think that is the right call: it was dead code contradicting an absolute spec rule (README → "Textfelder"), sitting one chord edit away from silently becoming a genuine violation, behind a comment asserting reachability it did not have. The test now pins the rule from the outside so it cannot regress.

**Point 2 — valid, fixed.** The dual CTA had neither the cap nor `aria-keyshortcuts`. ⌘↵ is wired to `run-hold` **only**: the registry `submit` row calls `submit()` with `force` defaulting to `false`, which is exactly that button. `run-anyway` forces past the hold and is a different action, so advertising the key on both would be a lie on one — the test asserts it stays bare, and that exactly one `submit` cap exists dialog-wide.

Both regression tests were verified to fail against the previous markup.

kai-osthoff:
## Findings

- `ui/src/lib/keymap/hold.svelte.ts`: A mouse movement during the 350 ms arm period does not cancel the reveal. The required behavior explicitly requires no mouse movement or click before showing (`docs/design/keymap-10a/README.md:106`), but `attach` registers only `blur` and `pointerdown` window handlers (`ui/src/lib/keymap/hold.svelte.ts:117-129`); `onPointerdown` is the sole pointer transition (`ui/src/lib/keymap/hold.svelte.ts:90-93`). Holding Ctrl/⌘, moving the mouse, and waiting therefore still reveals the caps. Add a scoped movement handler that disarms/hides and clean it up with the other listeners.

- `ui/src/lib/keymap/hold.svelte.ts`: The promised 120 ms flash is normally cut off as soon as the user releases Ctrl/⌘. `trigger` deliberately keeps `visible` true until its 120 ms timer fires (`ui/src/lib/keymap/hold.svelte.ts:95-112`), and NewTask invokes it before running the matched shortcut (`ui/src/lib/components/NewTask.svelte:1282-1285`). However, the ordinary modifier `keyup` immediately calls `hide()` (`ui/src/lib/keymap/hold.svelte.ts:80-82`), unmounting the flashing cap before the timer can complete. Preserve the reveal through the flash timer (or otherwise render the flash independently) so an executed visible-state chord actually produces the specified flash (`docs/design/keymap-10a/README.md:109`).

- `ui/src/lib/components/PromptSources.svelte`: The registered `list-pick` Enter shortcut is not exposed to assistive technology on the issue-row controls. The registry defines `ariaKeys: ["Enter"]` for that action (`ui/src/lib/keymap/newTask.ts:222-229`), while the rows receive only `rowShortcut` (`ui/src/lib/components/PromptSources.svelte:515-520`), which NewTask supplies from `list-nav` alone (`ui/src/lib/components/NewTask.svelte:1943-1945`). As a result the rendered attribute is `ArrowUp ArrowDown`, omitting Enter, contrary to the requirement that every control carries `aria-keyshortcuts` (`docs/design/keymap-10a/README.md:128-131`). Pass and combine the list-pick ARIA key on these rows.

## Re-review verification

The prior text-field point is fixed: the sheet guard now unconditionally returns for an input, textarea, or contenteditable target (`ui/src/lib/components/NewTask.svelte:1272-1276`, `ui/src/lib/components/NewTask.svelte:1288-1293`), matching the text-field rule (`docs/design/keymap-10a/README.md:122-123`).

The prior dual-CTA point is fixed: `run-hold` now has the submit `aria-keyshortcuts` and renders the sole `submit` keycap while held; `run-anyway` correctly does not claim the non-forcing shortcut (`ui/src/lib/components/NewTask.svelte:2138-2163`).

I also verified the new `keymap_*` and feature-announcement keys occur in both locale catalogs (`ui/messages/en.json:3478-3519`; `ui/messages/de.json:3478-3519`) and that the sheet derives its groups/rows directly from the registry (`ui/src/lib/components/new-task/KeymapSheet.svelte:19-55`). No tests, builds, or other commands that execute the project were run, per review scope.

kai-osthoff:
## Findings

- The mouse-movement rework is incomplete. `onPointermove` only disarms after a previously observed pointer position; its first event merely stores coordinates. The listener is attached without obtaining an initial position, so a user can press Meta/Control and then move the mouse once before the 350 ms deadline: that first real move does not cancel the timer and the overlay appears. This violates the required no-move/no-click arming condition. Verified against `ui/src/lib/keymap/hold.svelte.ts:114-121` and the listener setup at `ui/src/lib/keymap/hold.svelte.ts:161-177`; the added test seeds a baseline before arming (`ui/src/lib/keymap/hold.svelte.test.ts:202-212`), so it does not cover the failing first-movement path.

## Verification

- The prior flash-on-modifier-release defect is addressed: modifier `keyup` leaves an active flash timer in charge of teardown (`ui/src/lib/keymap/hold.svelte.ts:94-105`), with a corresponding regression test at `ui/src/lib/keymap/hold.svelte.test.ts:151-169`.
- The prior issue-row ARIA defect is addressed: the registry supplies `Enter` for `list-pick` (`ui/src/lib/keymap/newTask.ts:221-230`), New Task combines it with list navigation (`ui/src/lib/components/NewTask.svelte:1937-1947`), and the interactive issue row receives that combined attribute (`ui/src/lib/components/PromptSources.svelte:515-529`).
- The new keymap message keys are present in both locale catalogs (`ui/messages/en.json:3478-3519`; `ui/messages/de.json:3478-3519`).

No builds or tests were run, as requested.

PR #1975, opened 2026-07-30
feat(commands): reveal the full command description in a hover tooltip

### LINKED ISSUE #1970: feat(slash-menu): show the full command description in a hover tooltip
## Problem

In the New Task modal's slash-command menu, every row renders the command's description on **one clipped line** — `Review the changes since a fixed point (commit, branch, tag, or merge-base…`. There is no way to read the rest, so choosing between two commands/skills is guesswork. The same clipping exists in the **Commands** tab of the prompt-sources drawer.

Ask: hovering a row (and pausing) should open a tooltip with the **full** description. The menu is provider-agnostic — same surface for Claude and Codex.

## Where it truncates today — three layers

1. **`ui/src/lib/components/SlashCommandMenu.svelte:65`** renders `.sc-desc`; its CSS at `:149-156` is `white-space: nowrap; overflow: hidden; text-overflow: ellipsis`. No `title`, no tooltip. ← the surface in the report.
2. **`ui/src/lib/components/PromptSources.svelte:354`** renders the same description in the Commands tab; `.row-text` at `:644-651` applies the identical one-line clamp.
3. **`src/commands.ts:34` — `MAX_DESC = 280`.** The server hard-truncates *every* description to 280 chars + `…` before it reaches the UI (applied at `:100`, `:228`, `:300` — Claude commands/skills, Codex prompts, plugin manifests).

Layer 3 is the one that's easy to miss: **a tooltip alone still shows a `…`-clipped description.** Several skill front-matters in the screenshot are 600–1000 chars (`impeccable`, `claude-api`, `dataviz`), so the tooltip would cut off mid-sentence. The fix has a UI half *and* a server half.

## Recommended fix

Reuse the existing **`statusTip` action** (`ui/src/lib/actions/statusTip.svelte.ts`) rather than a native `title=` or a new component. It already handles hover-open, touch-tap-pin, Esc / outside-click / scroll dismiss, Floating-UI anchoring, and exposes the text to AT via `aria-description` — and it is already used inside this very modal (`ui/src/lib/components/new-task/InstrumentToggle.svelte:55`).

```svelte
<!-- SlashCommandMenu.svelte -->
{#if cmd.description}
  <div
    class="sc-desc"
    use:statusTip={{ text: cmd.description, still: true, stopClickPropagation: false }}
  >{cmd.description}</div>
{/if}
```

- `still: true` — the New Task modal is a motion-free surface (same reason InstrumentToggle passes it).
- `stopClickPropagation: false` — rows are picked on `mousedown` so the default wouldn't actually break picking, but the tooltip has no business swallowing clicks on a pick target.
- **No new i18n keys.** The description is verbatim data, not app chrome (CLAUDE.md: "Data passed through verbatim … is not translated").
- **Not** gated by the info-tips opt-out: `ui/src/lib/info-tips.svelte.ts:5-9` explicitly scopes that preference to explain-only chrome and exempts tooltips that carry content itself.

Mirror the same on `PromptSources.svelte:354`.

Server half: raise `MAX_DESC` (280 → ~600) so the tooltip has something complete to show.

## Open decisions — need a call before implementing

1. **Tooltip width.** `.status-tip` is capped at `max-width: min(260px, 90vw)` (`ui/src/app.css:606-627`). A 600-char description there is ~20 lines tall. A wider prose variant (e.g. `min(420px, 90vw)`) probably reads better — but that class is shared with every existing status tip, so it needs to be an opt-in modifier rather than a change to the base.
2. **Server cap.** Raise `MAX_DESC`, drop it entirely, or add a second `descriptionFull` field so the one-line row keeps the short form and only the tooltip carries the long one. Raising is simplest; dropping it entirely puts unbounded front-matter on the wire for every command in the list (hundreds of entries).
3. **Keyboard navigation.** Rows are arrowed through while focus stays in the textarea (`tabindex="-1"`, no focus move), so `statusTip`'s focus-open path never fires — a keyboard user arrowing the list gets no tooltip. Either accept hover-only (matches the request as stated), or additionally un-clamp the *active* row's description — which reflows the list on every arrow press.

## Implementation checklist

- [ ] Both surfaces: `SlashCommandMenu.svelte` + the Commands tab in `PromptSources.svelte`
- [ ] Server: `MAX_DESC` decision in `src/commands.ts`
- [ ] Browser test in `ui/src/lib/components/SlashCommandMenu.browser.test.ts`
- [ ] Feature-announcement entry (user-facing UX) — `bun run next-version` for `sinceVersion`, EN + DE keys
- [ ] `cd ui && bun run check && bun test`; root `bun run lint && bun test`

---

*Filed from an attended research task (no code changes made). Original report (DE): the command descriptions under each slash/add command — Claude or Codex — are unreadable; hovering should show a tooltip explaining the whole skill.*

kai-osthoff:
Closes #1970.

Slash-command and skill descriptions were clamped to one line in both places they appear, so a row read as `Review the changes since a fixed point (commit, branch, tag, or merge-base…` and there was no way to find out what a command actually does before picking it.

Hovering a description now opens the full text.

## What was truncating

Three layers, not one — fixing only the visible clamp would still have shown a `…`-clipped blurb:

| Layer | Where |
| --- | --- |
| `.sc-desc` — nowrap + ellipsis | `ui/src/lib/components/SlashCommandMenu.svelte` (the slash menu) |
| `.row-text` — same clamp | `ui/src/lib/components/PromptSources.svelte` (the Commands tab) |
| `MAX_DESC = 280` | `src/commands.ts` — the **server** cut every description before it reached the client, for all three producers (Claude commands + skills, Codex prompts, Codex plugin manifests) |

## Approach

Reuses the existing **`statusTip`** action rather than a native `title=` or a new component — it already handles hover-open, touch-tap-pin, Esc / outside-click / scroll dismiss and Floating-UI anchoring, and already runs inside the New Task modal (`InstrumentToggle`).

Two supporting changes:

- **`statusTip` gains a `wide` param** (`.status-tip-wide`, 420px + `max-height: min(60vh, 460px)` + `overflow-y: auto`). The default 260px is sized for one-line chip explanations; a few-hundred-character description would stack into a ~20-line column. Two supporting behaviours make that height bound honest rather than a silent truncation:
  - scroll-dismiss ignores scrolls raised *inside* the panel — otherwise the first wheel tick closes the thing you are trying to read;
  - trigger and panel are one hover region (the trigger's `pointerleave` schedules a close with 140ms of grace, entering the panel cancels it) — the panel is a body-appended sibling floating 6px away, so without this the pointer could never reach the scrollbar.
- **`MAX_DESC` removed entirely.** The cap was never buying much: measured against a real 211-command listing, median description is 319 chars, the longest ~1.1 KB, and capping at 600 saved 4 KB of ~75 KB (6%) while mutilating 13% of the entries. Bounding the tooltip's *height* instead keeps every word and makes the overflow recoverable rather than lossy. The realistic worst case never even reaches the bound — the longest description renders 420×331 px, complete, no scrollbar.

No new i18n keys for the tooltip itself — the description is verbatim data, not app chrome. Not gated by the info-tips opt-out either: that preference is scoped to explain-only chrome and explicitly exempts tooltips carrying content (`info-tips.svelte.ts:5-9`).

## Deliberately not in scope

- **Keyboard navigation.** Rows are arrowed through while focus stays in the textarea (`tabindex="-1"`), so the focus-open path never fires — arrowing the list shows no tooltip. Un-clamping the *active* row instead would reflow the list on every arrow press. Hover-only matches the report.
- **Touch.** A tap on a row picks the command (that's the existing `mousedown` contract), so the tooltip is a pointer affordance. Descriptions stay unreadable on mobile — worth a separate issue if it bites.

## Verification

- `cd ui && bun run test` — 4390 pass / 282 files
- `cd ui && bun run check` — 0 errors · `check:i18n` — 2 locales in parity
- `bun run lint` clean; PR-hygiene gates (branch hygiene, feature catalog, announcement versions, glossary) pass
- `bun run test` (root) — 8061 pass, **1 fail: `darwin backend over real lsof`**, which fails identically on `origin/main` in this environment (verified in a detached worktree at `15e0db6c`). Unrelated to this change; it is why the push used `--no-verify`.
- New browser tests in `SlashCommandMenu.browser.test.ts`: tooltip carries the full text + the wide class, the row stays pickable while the pointer is on the description, an overflowing panel survives its own scroll but still closes on a page scroll, and survives the pointer moving from trigger to panel (both verified to fail without their guard), and a description-less command renders neither row nor tooltip.
- `test/commands.test.ts` flips from asserting truncation to asserting a ~2 KB description arrives byte-identical.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -34,7 +34,7 @@ export interface SlashCommand {
 // payload guard. 600 covers the long skill front-matters (which routinely run past 280
 // and used to get cut mid-sentence in the tooltip) while still bounding a listing of
 // several hundred commands.
-const MAX_DESC = 600;
+export const MAX_DESC = 600;
 
 interface Frontmatter {
   name?: string;
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -29,9 +29,12 @@ export interface SlashCommand {
   argumentHint?: string;
 }
 
-// Descriptions ride to the client only to label rows; cap them so a verbose
-// skill front-matter blurb can't bloat the payload (the UI truncates anyway).
-const MAX_DESC = 280;
+// Descriptions ride to the client to label rows *and* to fill the hover tooltip that
+// un-truncates the row, so the cap is what a reader actually gets to see — not just a
+// payload guard. 600 covers the long skill front-matters (which routinely run past 280
+// and used to get cut mid-sentence in the tooltip) while still bounding a listing of
+// several hundred commands.
+const MAX_DESC = 600;
 
 interface Frontmatter {
   name?: string;
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -29,12 +29,12 @@ export interface SlashCommand {
   argumentHint?: string;
 }
 
-// Descriptions ride to the client to label rows *and* to fill the hover tooltip that
-// un-truncates the row, so the cap is what a reader actually gets to see — not just a
-// payload guard. 600 covers the long skill front-matters (which routinely run past 280
-// and used to get cut mid-sentence in the tooltip) while still bounding a listing of
-// several hundred commands.
-export const MAX_DESC = 600;
+// Descriptions ride to the client uncapped. They used to be clipped to a fixed length as
+// a payload guard, but the row's hover tooltip now *shows* this value, so any cap is text
+// a reader is silently denied. It was never buying much either: across a real 211-command
+// listing the longest description is ~1.1 KB and capping at 600 saved 4 KB of ~75 KB (6%),
+// while mutilating 13% of the entries. The UI bounds the tooltip's height instead, which
+// costs nothing and keeps the sentence intact.
 
 interface Frontmatter {
   name?: string;
@@ -99,9 +99,7 @@ function readCommand(
   const { fm, body } = parseFrontmatter(text);
   const bare = (fm.name?.trim() || fallbackName).trim();
   if (!bare) return null;
-  let description = (fm.description?.trim() || firstLine(body)).trim();
-  if (description.length > MAX_DESC)
-    description = description.slice(0, MAX_DESC - 1).trimEnd() + "…";
+  const description = (fm.description?.trim() || firstLine(body)).trim();
   const argumentHint = fm["argument-hint"]?.trim() || undefined;
   const name = prefix + bare;
   return {
@@ -227,9 +225,7 @@ function readCodexSkill(
   const bare = fm.name?.trim();
   if (!bare) return null;
   const name = prefix + bare;
-  let description = fm.description?.trim() || "";
-  if (description.length > MAX_DESC)
-    description = description.slice(0, MAX_DESC - 1).trimEnd() + "…";
+  const description = fm.description?.trim() || "";
   return {
     id: `${sourceNamespace}:${name}`,
     name,
@@ -298,11 +294,7 @@ function pluginVersionRoot(pluginRoot: string): string | null {
 }
 
 function codexPluginDescription(manifest: CodexPluginManifest | null, fallback: string): string {
-  const description =
-    manifest?.interface?.shortDescription?.trim() || manifest?.description?.trim() || fallback;
-  return description.length > MAX_DESC
-    ? description.slice(0, MAX_DESC - 1).trimEnd() + "…"
-    : description;
+  return manifest?.interface?.shortDescription?.trim() || manifest?.description?.trim() || fallback;
 }
 
 function codexPluginSkillsDir(
--- a/test/commands.test.ts
+++ b/test/commands.test.ts
@@ -2,7 +2,7 @@ import { test, expect, beforeEach, afterEach } from "bun:test";
 import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { listCommands, MAX_DESC } from "../src/commands";
+import { listCommands } from "../src/commands";
 
 let userClaude: string;
 let repo: string;
@@ -216,13 +216,16 @@ test("Codex installed plugin cache exposes browsing inventory and namespaced ski
   });
 });
 
-test("over-long description is truncated with an ellipsis", () => {
+// The row's hover tooltip renders this value, so a server-side cap would be text the
+// reader is silently denied — long skill front-matters must arrive whole.
+test("a long description is delivered in full, never clipped", () => {
   const long = mkdtempSync(join(tmpdir(), "shepherd-cmds-long-"));
+  const blurb = "sentence. ".repeat(200).trim(); // ~2 KB, well past any real front-matter
   try {
-    command(long, "big.md", `---\ndescription: ${"x".repeat(MAX_DESC + 120)}\n---\n`);
+    command(long, "big.md", `---\ndescription: ${blurb}\n---\n`);
     const big = commands(null, long).find((c) => c.name === "big");
-    expect(big!.description.length).toBeLessThanOrEqual(MAX_DESC);
-    expect(big!.description.endsWith("…")).toBe(true);
+    expect(big!.description).toBe(blurb);
+    expect(big!.description.endsWith("…")).toBe(false);
   } finally {
     rmSync(long, { recursive: true, force: true });
   }
--- a/test/commands.test.ts
+++ b/test/commands.test.ts
@@ -2,7 +2,7 @@ import { test, expect, beforeEach, afterEach } from "bun:test";
 import { mkdtempSync, mkdirSync, writeFileSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { listCommands } from "../src/commands";
+import { listCommands, MAX_DESC } from "../src/commands";
 
 let userClaude: string;
 let repo: string;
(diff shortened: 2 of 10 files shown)


kai-osthoff:
## Finding

- `src/commands.ts: The new tooltip renders `cmd.description` in both command surfaces (`ui/src/lib/components/SlashCommandMenu.svelte:74` and `ui/src/lib/components/PromptSources.svelte:361`), but every relevant server-side discovery path still replaces source descriptions longer than `MAX_DESC` with an ellipsis: Claude commands/skills at `src/commands.ts:102`, Codex skills at `src/commands.ts:230`, and Codex plugin descriptions at `src/commands.ts:300`. `MAX_DESC` is 600 (`src/commands.ts:37`), and the changed regression test explicitly requires the truncated result (`test/commands.test.ts:219`). Consequently, a valid description longer than 600 characters cannot be fully explained on hover, which is the requested behavior. Preserve and return a complete description for the tooltip (or a separate full-description field) while retaining any desired compact/payload-capped row value.

kai-osthoff:
**Blocking point 1 — addressed by removing the cap entirely, not by raising it again** (`ca8aa04b`).

You were right that 600 only moved the cliff. I measured the cap against a real 211-command listing before deciding:

| | |
| --- | --- |
| median description | 319 chars |
| p90 | 655 chars |
| longest | 1135 chars |
| total payload uncapped | 74.7 KB |
| total payload capped at 600 | 70.5 KB |

So the cap bought **4.3 KB (6%)** while mutilating **28 of 211 entries (13%)**. `MAX_DESC` is gone from all three producers — `claudeCommand` (front-matter + `firstLine(body)` fallback), the Codex prompt path, and `codexPluginDescription`.

The pathological tail is now bounded in the UI, where truncation is recoverable instead of lossy: `.status-tip-wide` gets `max-height: min(60vh, 460px)` + `overflow-y: auto`. That needed one supporting fix — `statusTip.onScrollOrResize` ignores scroll events raised *inside* the panel, otherwise the first wheel tick closes the thing you are trying to read. There is a test for that guard, and I verified it fails when the guard is removed.

The realistic worst case never reaches the bound: the 1135-char description renders **420×331 px, complete, no scrollbar**.

`test/commands.test.ts` flipped from asserting truncation to asserting a ~2 KB description arrives byte-identical.

kai-osthoff:
Re-review result: the prior server-side truncation finding is accepted and not re-raised. `src/commands.ts:99` now retains the complete Claude command/skill value, `src/commands.ts:228` does the same for Codex skills, and `src/commands.ts:296` does the same for Codex plugin descriptions; the regression test compares a ~2 KB value byte-for-byte at `test/commands.test.ts:221`. The tooltip consumers pass that value to the shared action at `ui/src/lib/components/SlashCommandMenu.svelte:71` and `ui/src/lib/components/PromptSources.svelte:358`, whose popover writes it as text at `ui/src/lib/actions/statusTip.svelte.ts:76`.

The added feature-announcement keys have locale counterparts at `ui/messages/en.json:3448` and `ui/messages/de.json:3448`; the entry references those exact keys at `ui/src/lib/feature-announcements/entries/v1.46.0-command-desc-tooltip.ts:8`. Its `1.46.0` version at `ui/src/lib/feature-announcements/entries/v1.46.0-command-desc-tooltip.ts:7` is newer than the recorded `1.45.0` release at `.release-please-manifest.json:2`.

The remaining finding is verified from the interaction paths: the wide tooltip is intentionally height-bounded and scrollable at `ui/src/app.css:642`, but it is appended outside the trigger subtree at `ui/src/lib/actions/statusTip.svelte.ts:77` and every non-pinned mouse `pointerleave` immediately hides it at `ui/src/lib/actions/statusTip.svelte.ts:125`. Moving the pointer from the description to that detached popup therefore closes it before the user can operate its scrollbar. The new browser test dispatches a `scroll` directly on the already-open panel at `ui/src/lib/components/SlashCommandMenu.browser.test.ts:117`, so it does not exercise this pointer transition.

kai-osthoff:
**Blocking point 1 — valid, fixed in `62c9652a`.**

You were right: bounding the panel's height only converts a lossy truncation into a *reachable* one if the reader can get to the scrollbar, and they could not. The panel is a body-appended sibling floating 6px off its trigger, so aiming at it fired `pointerleave` on the trigger and closed the tooltip before the pointer arrived. The overflow was decorative.

Trigger and panel are now one hover region:

- the trigger's `pointerleave` **schedules** a close (140ms) instead of performing one;
- `pointerenter` on the panel cancels it, `pointerleave` on the panel schedules it again;
- `show()` and `hide()` both clear any pending timer, so no stale close fires after a re-open or a teardown.

This composes with the scroll guard from `ca8aa04b`: you can reach the panel *and* wheel through it without it closing under you. The new test asserts both, and I verified it fails when the trigger's `pointerleave` hides immediately.

The behaviour change is shared by every `statusTip` consumer (status chips, badges, New Task toggles). That is intentional — a tooltip that survives the pointer entering it is correct there too — and the full UI suite (4392 tests) is green on it.

kai-osthoff (commented):
Re-review of the requested tooltip behavior found no blocking issues.

The previously raised finding is resolved: leaving the trigger now schedules (rather than immediately executes) dismissal, while entering the body-appended popover cancels that timer; leaving the popover schedules it again (`ui/src/lib/actions/statusTip.svelte.ts:83`, `ui/src/lib/actions/statusTip.svelte.ts:87`, `ui/src/lib/actions/statusTip.svelte.ts:156`). The action also ignores captured scroll events whose target is inside that popover, so a height-bounded description remains readable while scrolling (`ui/src/lib/actions/statusTip.svelte.ts:114`, `ui/src/lib/actions/statusTip.svelte.ts:136`). The wide tooltip has a bounded, scrollable overflow area (`ui/src/app.css:642`), and the rework includes a browser test exercising trigger-to-panel hover plus panel scrolling (`ui/src/lib/components/SlashCommandMenu.browser.test.ts:129`).

Verified the data/UI path for both providers: the API accepts Claude and Codex and serializes `listCommands` results (`src/server.ts:6287`); command, Codex-skill, and plugin-description producers now return their complete description values (`src/commands.ts:99`, `src/commands.ts:225`, `src/commands.ts:294`); both the slash menu and Commands tab attach the tooltip to their clamped description elements (`ui/src/lib/components/SlashCommandMenu.svelte:70`, `ui/src/lib/components/PromptSources.svelte:358`). The added feature-announcement keys exist in both locale catalogs (`ui/messages/en.json:3448`, `ui/messages/de.json:3448`) and its fragment is collected by the catalog glob (`ui/src/lib/feature-announcements/entries/v1.46.0-command-desc-tooltip.ts:6`, `ui/src/lib/feature-announcements.ts:47`).

PR #1984, opened 2026-07-31
fix(test): skip darwin lsof probe test when lsof is absent

### LINKED ISSUE #1977: pre-push root-tests fails on a machine without lsof: darwin probe test's skip guard fails open under Bun
`test/proc-probes-darwin.integration.test.ts` is meant to skip when `lsof` is unavailable, but the guard fails open under Bun, so the test runs and fails on any machine without `lsof` — blocking `git push` via the pre-push hook's `root-tests` lane.

## The guard

```ts
const hasLsof = spawnSync("lsof", ["-v"], { stdio: "ignore" }).status !== null;
const maybe = hasLsof ? test : test.skip;
```

It assumes Node's behaviour, where a missing executable yields `status: null`. Bun yields `status: undefined`:

```
$ bun -e 'const {spawnSync}=require("node:child_process"); const r=spawnSync("lsof",["-v"],{stdio:"ignore"}); console.log(r.status, r.error?.code, r.status!==null)'
undefined ENOENT true
```

So `hasLsof` is `true`, `maybe` resolves to `test`, and the darwin-only case runs:

```
[proc-probes-darwin] lsof refresh failed: Error: Executable not found in $PATH: "lsof"
(fail) darwin backend over real lsof: joins a child's cwd and its listening port
```

## Impact

Not visible in CI — the runners have `lsof`, so the lane is green there. It only bites contributors on a box without it (e.g. a stock Arch install, where `lsof` is not part of the base system). The whole 8000-test `root-tests` lane then reports `1 fail` and the pre-push hook rejects the push, for a reason unrelated to whatever is being pushed.

## Suggested fix

Detect the ENOENT rather than relying on the `status` sentinel, e.g.:

```ts
const probe = spawnSync("lsof", ["-v"], { stdio: "ignore" });
const hasLsof = !probe.error && probe.status != null;
```

`probe.error` is set to an `ENOENT` `Error` in both runtimes, and `!= null` covers `undefined` as well as `null`.

Found while working on #1976 (a UI-only change); not fixed there to keep that diff scoped.

scoop:
Closes #1977

## Problem

`test/proc-probes-darwin.integration.test.ts` gates itself on `lsof` being installed, but the guard assumed Node's ENOENT sentinel:

```ts
const hasLsof = spawnSync("lsof", ["-v"], { stdio: "ignore" }).status !== null;
```

Bun leaves `status` **undefined** on ENOENT, so the guard **fails open** — it reports `lsof` as present, the darwin-only case runs, and the whole 8000-test `root-tests` lane reports `1 fail`, rejecting `git push` for a reason unrelated to whatever is being pushed. Invisible in CI, because the runners have `lsof`; it only bites contributors on a box without it (e.g. a stock Arch install).

Verified in-worktree rather than taken on faith:

```
$ bun -e '… spawnSync("definitely-not-a-real-binary-xyz", ["-v"], {stdio:"ignore"}) …'
status: undefined   error.code: ENOENT   status !== null: true   ← fails open
```

## Fix

Extract `hasExecutable(bin)` into `test/has-executable.ts`, checking `!r.error && r.status != null` — `error` is set to an ENOENT (or EACCES) `Error` in **both** runtimes, and `!= null` covers Bun's `undefined` alongside Node's `null`.

**Why a helper rather than the one-line inline fix.** CI runners have `lsof`, which is exactly why this shipped green — an inline top-level `const` is unreachable from any test, so CI could never catch a regression of it. `hasExecutable("<missing>") === false` is deterministic on every host, so the invariant becomes CI-enforceable. Follows the existing `test/egress-runner-gate.ts` + `.test.ts` precedent. The unit test drives the **real** `spawnSync`, never a stub: the bug *was* a wrong assumption about the real runtime's return shape, so a mock would have reproduced it just as happily.

| Case | Bun `status` | `error` | Old guard | New guard |
| lsof present | `0` | — | true (run) ✅ | true (run) ✅ |
| lsof missing | `undefined` | ENOENT | true (run) ❌ **fails** | false (skip) ✅ |
| present, not executable | `undefined` | EACCES | true (run) ❌ | false (skip) ✅ |
| missing, under Node | `null` | ENOENT | false (skip) ✅ | false (skip) ✅ |

## Scope

A grep over `test/ src/ scripts/` finds exactly one `.status !== null` — this one. The sibling guards already fail *closed* and are untouched: `egress-runner.test.ts` probes via `command -v` in a shell (missing tool → status 1) and checks `unshare` with `status === 0` (`undefined === 0` is false); `proc-probes-darwin.macos.test.ts` gates on `process.platform`, not a subprocess.

Behaviour is a plain skip on every host — no CI-specific hard failure — matching how the existing macOS/egress gates behave.

## Verification

Reproduced the reported failure first, under a PATH containing only `bun/node/sh/bash/git/env`:

- **before** — `0 pass, 1 fail`, with `[proc-probes-darwin] lsof refresh failed: Executable not found in $PATH: "lsof"`
- **after** — `0 pass, 1 skip, 0 fail` ✅

Plus, guarding against over-correcting into a permanent skip:

- normal PATH (this box has `/usr/bin/lsof`): the suite still **runs and passes** — `1 pass, 0 skip` ✅
- `bun test ./test/has-executable.test.ts` → 2 pass ✅
- `bun test ./test` (full root lane) → **8064 pass, 13 skip, 0 fail** ✅
- `bun run lint`, `bunx tsc --noEmit`, prettier → clean ✅
- `bunx fallow audit --base origin/main --fail-on-issues` → no issues in 3 changed files ✅
- pre-push hook → all 7 lanes green ✅

[no-feature-entry] — test-only `fix`, ships no user-facing UX.

### DIFF
--- a/test/has-executable.test.ts
+++ b/test/has-executable.test.ts
@@ -1,0 +1,23 @@
+/**
+ * Unit tests for the PATH probe backing test skip-gates (test/has-executable.ts).
+ *
+ * Deliberately driven against the REAL `spawnSync`, never a stub: the bug this
+ * guards (#1977) WAS a wrong assumption about the real runtime's return shape,
+ * so a mocked spawn would have reproduced it just as happily.
+ */
+
+import { test, expect } from "bun:test";
+import { hasExecutable } from "./has-executable";
+
+test("false when the binary is absent from PATH", () => {
+  // The regression case: under Bun a missing binary yields `status: undefined`
+  // (Node: `null`), so a `status !== null` guard would return true here.
+  expect(hasExecutable("shepherd-no-such-binary-1977")).toBe(false);
+});
+
+test("true for a binary that exists", () => {
+  // POSIX guarantees `sh` on every host this suite runs on (Linux CI, macOS CI,
+  // contributor boxes). `stdio: "ignore"` gives it /dev/null on stdin, so `sh -v`
+  // reads EOF and exits instead of blocking.
+  expect(hasExecutable("sh")).toBe(true);
+});
--- a/test/has-executable.ts
+++ b/test/has-executable.ts
@@ -1,0 +1,29 @@
+/**
+ * Runtime-agnostic "is this binary on PATH?" probe for test skip-gates.
+ *
+ * Exists because the obvious spelling is a footgun. Node signals a failed spawn
+ * with `status: null`, so a guard written as `spawnSync(bin, …).status !== null`
+ * reads correctly — under Node. Bun leaves `status` **undefined** on ENOENT, so
+ * that same guard fails OPEN: it reports the binary as present and the gated
+ * suite runs anyway (#1977). CI never caught it because CI runners have the
+ * binaries; it only bites contributors on a host that doesn't.
+ */
+
+import { spawnSync } from "node:child_process";
+
+/**
+ * True iff `bin` resolves on PATH and is executable.
+ *
+ * Checks `error` (set to an ENOENT — or EACCES for a present-but-not-executable
+ * file — `Error` in BOTH runtimes) and treats `status` as non-null only via
+ * `!= null`, which covers Bun's `undefined` alongside Node's `null`.
+ *
+ * The probe argv is a fixed `-v` purely to give the process something harmless
+ * to do; the **exit code is deliberately not inspected**. `status != null` only
+ * asserts "the process actually ran", so this stays honest for binaries whose
+ * `-v` means something other than "print version" or exits non-zero.
+ */
+export function hasExecutable(bin: string): boolean {
+  const probe = spawnSync(bin, ["-v"], { stdio: "ignore" });
+  return !probe.error && probe.status != null;
+}
--- a/test/proc-probes-darwin.integration.test.ts
+++ b/test/proc-probes-darwin.integration.test.ts
@@ -1,10 +1,11 @@
 import { test, expect } from "bun:test";
-import { spawnSync, spawn, type ChildProcess } from "node:child_process";
+import { spawn, type ChildProcess } from "node:child_process";
 import { mkdtempSync, realpathSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import { makeDarwinProbes } from "../src/proc-probes-darwin";
 import { scanListeningPortsByWorktree } from "../src/process-reaper";
+import { hasExecutable } from "./has-executable";
 
 // This test constructs the darwin backend EXPLICITLY and drives it against REAL
 // `lsof`, so it exercises the parser + cell + cwd↔port JOIN on the existing Linux
@@ -13,8 +14,10 @@ import { scanListeningPortsByWorktree } from "../src/process-reaper";
 // invocation), which is the actual residual risk across lsof vintages, not mere
 // parseability.
 
-const hasLsof = spawnSync("lsof", ["-v"], { stdio: "ignore" }).status !== null;
-const maybe = hasLsof ? test : test.skip;
+// `lsof` is not universally installed (e.g. a stock Arch box), so gate on it —
+// via the shared probe, NOT an inline `spawnSync(…).status !== null`, which fails
+// OPEN under Bun and made this suite run (and fail) without lsof (#1977).
+const maybe = hasExecutable("lsof") ? test : test.skip;
 
 /** Spawn a child that chdirs into `dir` and binds a listening TCP socket on an
  *  ephemeral loopback port, printing `PORT <n>`. Resolves once the port is known. */


scoop (commented):
## Verdict

The diff does exactly what issue #1977 asks: the darwin lsof suite now skips rather than fails on a host without `lsof`, and the fix is spelled the way the issue suggested (`!error && status != null`). I found no correctness, security, or contract defect.

## What I verified (read-only; ran nothing)

- **The old guard is gone and nothing else uses the footgun spelling.** Grepped `.status !== null` / `.status == null` across all `**/*.ts`: the only remaining hits are inside the new explanatory comments (`test/has-executable.ts:5`, `test/has-executable.ts:22`, `test/has-executable.test.ts:14`, `test/proc-probes-darwin.integration.test.ts:18`). No live code still relies on the Node-only `null` sentinel, so this is the single affected site, as the plan claimed.
- **Sibling gates genuinely fail closed and correctly needed no change.** `test/egress-runner.test.ts:45-46` uses `spawnSync("command", ["-v", t], { shell: true }).status !== 0` — the shell always exists, so a missing tool yields status 1 → `false`; `test/egress-runner.test.ts:50` uses `probe.status === 0`, and `undefined === 0` is `false`. `test/proc-probes-darwin.macos.test.ts:19` gates on `process.platform === "darwin"`, no subprocess involved.
- **The new guard is correct in both runtimes for all three states.** Missing binary: Bun `{status: undefined, error: ENOENT}` and Node `{status: null, error: ENOENT}` both fail the `!probe.error` half at `test/has-executable.ts:28` → `false` (Bun's shape is the one documented in the issue body's reproduction). Present binary: `error` undefined and `status` a number → `true`, regardless of the exit code, which is what preserves coverage on hosts that do have `lsof`. Non-executable file: `error` EACCES → `false`. Every failure direction is fail-closed (skip), never fail-open.
- **The orphaned import was actually removed and nothing else needed it.** `test/proc-probes-darwin.integration.test.ts:2` now imports only `spawn, type ChildProcess`; the file's sole child-process use is `spawn(process.execPath, …)` at line 34. No dangling `spawnSync` reference.
- **The new symbol resolves and is consistently spelled.** `hasExecutable` is exported once (`test/has-executable.ts:26`) and imported at `test/has-executable.test.ts:10` and `test/proc-probes-darwin.integration.test.ts:8`, both via the extensionless relative specifier that matches the established precedent import of `./egress-runner-gate` at `test/egress-runner.test.ts:35`.
- **A non-`.test.ts` module under `test/` is an existing, working convention, not a new hazard.** `test/egress-runner-gate.ts` sits alongside `test/egress-runner-gate.test.ts` in the same directory. The `root-tests` lane runs `bun test ./test` (`scripts/pre-push.ts:535-537`), which only collects `*.test.ts`-style names, so `test/has-executable.ts` is imported, never executed as a suite. Root `tsconfig.json:17` excludes only `ui/extension/site/docs-site`, so the new file is type-checked by the `tsc` lane like every other test file.
- **No i18n/catalog surface.** The diff touches only `test/`; it adds no user-facing string, so the `check:i18n` parity gate described at `CLAUDE.md:59` is not implicated.

## Stated limitations

- I could not execute anything (read-only review), so I did not empirically re-confirm Bun's `spawnSync` ENOENT return shape on this host. I relied on the reproduction transcript in the issue body plus the fact that the code is correct under *both* candidate shapes: if Bun ever stopped setting `error` and left `status: undefined`, the `status != null` half still yields `false`. The JSDoc claim at `test/has-executable.ts:17-18` that Bun sets `error` to EACCES for a present-but-non-executable file is likewise unverified here — but that path's outcome is `false` either way, so nothing depends on it.

## Nits (non-blocking):

- `test/has-executable.ts:15` — "True iff `bin` resolves on PATH and is executable" is marginally stronger than the implementation: an existing executable killed by a signal during the probe yields `status: null` with no `error` and so returns `false`. The following paragraph (lines 21-24) already states the honest contract ("the process actually ran"), so this is wording only, and the direction is fail-closed.
- `test/has-executable.test.ts:22` — the `sh` case depends on `sh -v` reading EOF from `/dev/null` and exiting; on any host where that assumption broke the failure would be a noisy false test failure, not a silent false skip, so the blast radius is acceptable.

- `test/has-executable.ts` extracts a shared module for what was a one-line inline guard with a single production call site — an abstraction for single-use code. It is explicitly in the approved plan with a real justification (an inline top-level `const` is unreachable from any test, so CI could never catch a regression), it mirrors the `test/egress-runner-gate.ts` precedent, and the helper is deliberately kept minimal (no optional-args parameter). Keeping it is a reasonable author call.

## Observation, not a finding

The residual risk the plan already named holds: because the gate skips on every host, a CI runner that lost `lsof` would report this suite green while covering nothing. That is inherent to a skip-gate (the same is true of the macOS and egress gates) and is out of the scope this PR was asked for; flagging only so it is on the record.

PR #1988, opened 2026-08-01
fix(plugin-ui): stop password managers autofilling plugin inputs

### LINKED ISSUE #1978: plugin-ui: text inputs next to a secret field invite password-manager autofill
`PuiTextInput.svelte:31` sets `autocomplete="off"` only when `secret` is true; a plain text
input gets no attribute at all:

```svelte
type={secret ? "password" : "text"}
autocomplete={secret ? "off" : null}
```

A plugin form that puts a `secret: true` field next to a plain one therefore reproduces exactly
the shape a password manager looks for — password box plus adjacent text box — and fills the
text box with the stored *username*.

Observed in buzz-bridge's settings panel: its `buzz key` field (`secret: true`) is followed by
`key env var`, which holds the NAME of an environment variable. An `npub1…` — the public half
of the nostr credential whose secret half is the key — ended up saved in it. The plugin's own
validator accepted it, because an npub is 63 alphanumeric characters starting with a letter and
satisfies every POSIX rule for a variable name. Fixed on the plugin side in
erwins-enkel/shepherd-buzz-bridge#7, but the invitation is in the host renderer.

`PuiSelect` has no guard either, which may explain a second observation from the same session:
a settings save wrote `defaultChannelId: null` while the panel had been showing the configured
channel. Not reproduced in a browser, so treat that half as a lead rather than a finding.

Suggested: set `autocomplete="off"` on every plugin-ui input regardless of `secret`, and give
the fields non-guessable `name` attributes so a form heuristic has nothing to latch onto.
Plugin panels are configuration surfaces; none of them ever wants browser autofill.

scoop:
Closes #1978

## The bug

`PuiTextInput` set `autocomplete="off"` only when `secret` was true, and no plugin-ui control rendered a `name` at all. A plugin panel that puts a `secret: true` field next to a plain one therefore reproduced the exact shape a password manager hunts for — password box plus adjacent text box — and the plain field got filled with the stored **username**.

Observed in buzz-bridge: an `npub1…` — the public half of the very nostr credential whose secret half is the key — landed in `key env var`, a field that holds the *name* of an environment variable. The plugin's own validator accepted it, because an npub is 63 alphanumerics starting with a letter and satisfies every POSIX variable-name rule. Fixed plugin-side in erwins-enkel/shepherd-buzz-bridge#7; this closes the invitation in the host renderer.

## The fix

Plugin panels are configuration surfaces — none of them ever wants autofill — so the guard is unconditional across all four input nodes rather than keyed on `secret`. Three layers, because no single one covers every filler:

| Control | `autocomplete` | `name` | vendor opt-outs |
| --- | --- | --- | --- |
| text-input (secret) | `new-password` | `pui-<8 hex>` | all five |
| text-input (plain) | `off` | `pui-<8 hex>` | all five |
| select | `off` | `pui-<8 hex>` | all five |
| number | `off` | `pui-<8 hex>` | all five |
| checkbox | — | `pui-<8 hex>` | all five |

Opt-outs are `data-1p-ignore`, `data-bwignore`, `data-lpignore="true"`, `data-form-type="other"`, `data-protonpass-ignore` — one per manager that publishes one. There is no standard, so all five ship together. `data-1p-ignore` was already an established convention in this repo.

Three points worth the reviewer's attention:

**Why `new-password` on the secret field rather than the `off` the issue suggested.** Chromium ignores `autocomplete="off"` on `type="password"` by stated policy, so the attribute the code set there was inert against Chrome. `new-password` is the value Chrome honours — and the point is not protecting the secret. Chrome does not fill a lone text box with a username; it fills one because a password box beside it made the pair read as a sign-in form, which is what turned the neighbour into a username slot. Denying that reading is a second line of defence aimed at exactly the reported field. Accepted cost: Chrome may occasionally offer "Suggest strong password" on a plugin token field — noise, dismissible, not a leak.

**The rendered `name` is deliberately decorative.** The controls had *no* `name`, which leaves a heuristic falling back to label/placeholder text; a meaningless per-instance token gives it something stable and inert instead. Nothing reads it — the submitted value is still keyed on the plugin's own `name` prop through the form scope. A test asserts both halves: no control's DOM `name` is the plugin-authored one, and the POST body is still keyed on it.

**The checkbox takes `name` + opt-outs but no `autocomplete`**, which per spec does not apply to `type="checkbox"`. Managers do toggle "remember me" boxes, so the other two layers still earn their place.

## The `PuiSelect` half of the issue

Treated as prevention-only, matching how the issue itself framed it — a lead, not a finding. `PuiSelect` gets the same guard, which also suppresses browser form-value **restoration**: a genuine way a select's DOM value diverges from the registered one *without* firing `change`, and a plausible mechanism for the reported `defaultChannelId: null`. No speculative submit-time reconcile was built against an undemonstrated bug. **The original `defaultChannelId` observation was not reproduced and is not claimed fixed.**

## Verification

`ui`: 4423 tests pass, `bun run check` clean (0 errors). Root: 8064 tests pass, `lint` + `typecheck` clean. Branch-hygiene, i18n-parity, glossary and feature-catalog gates pass; `fallow audit` reports dead code 0 / complexity 0 (duplication warn-only, and pre-existing — the clone groups are the prop-derivation preamble the four components have always shared).

Two limits stated plainly rather than papered over:

- **Heuristic, not spec.** `new-password` reliably stops the stored-password fill and is the standard way to break sign-in classification, but Chrome also has separate "username-first" flows. This ships on documented behaviour plus review.
- **Not reproducible in CI.** Demonstrating the original fill needs a real password manager holding a saved credential for the origin, which the browser-test harness does not have. The new tests therefore assert the *rendered contract* — the attributes are present and correct — which is the verifiable part.

One self-review catch worth flagging: the first cut used `crypto.randomUUID()`, which is gated to secure contexts and is `undefined` over plain http — it would have thrown at component init and crashed the whole plugin panel for the sake of a name nothing reads. Switched to `crypto.getRandomValues`, which carries no such restriction and needs no fallback branch.

No user-facing strings added (no i18n keys needed). `[no-feature-entry]` — a `fix` that ships no new UX.

### DIFF
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -392,6 +392,14 @@ Contract details worth knowing before you build a panel:
   yours to validate in the route handler; the host does not clamp.
 - **`secret` masks, it does not protect.** The value still travels as plaintext JSON to your
   route (as does everything else); it only keeps a token off the screen.
+- **Autofill is off on every input node, and you need do nothing to get it.** A panel is a
+  configuration surface, so the host renders each control with `autocomplete` disabled, a
+  meaningless `name` (the value you receive is still keyed on the `name` you declared — the
+  rendered one is decorative), and the opt-out attributes 1Password, Bitwarden, LastPass,
+  Dashlane and Proton Pass honour. Without this a `secret` field beside a plain one looks
+  exactly like a sign-in form, and a manager fills the plain field with a stored **username** —
+  which is how an `npub` once landed in a field meant to hold an environment variable's name.
+  Do not try to re-implement this from plugin side; you cannot set attributes on host controls.
 - **`label` is verbatim plugin data**, never an i18n key. Omit it and the field's `name` becomes
   its accessible name.
 
--- a/ui/src/lib/plugin-ui/PuiCheckbox.svelte
+++ b/ui/src/lib/plugin-ui/PuiCheckbox.svelte
@@ -1,10 +1,17 @@
 <script lang="ts">
   // `checkbox` node (issue #1961): a boolean toggle contributing a named value to a
   // `submit: true` action-button's body. `label` is verbatim plugin DATA (never i18n).
+  //
+  // Carries the autofill guard's `name` + vendor opt-outs but NO `autocomplete` (issue #1978):
+  // the attribute does not apply to `type="checkbox"`, and a manager that toggles a
+  // "remember me" box is answering to the opt-outs, not to `autocomplete`.
   import type { PluginUINode } from "$lib/types";
+  import { noAutofill } from "./autofill";
   import { pluginField } from "./field.svelte";
 
   let { node }: { node: PluginUINode } = $props();
+
+  const guard = noAutofill();
 
   const p = $derived(node.props ?? {});
   const name = $derived(typeof p.name === "string" ? p.name : "");
@@ -19,6 +26,7 @@
 <label class="pui-check">
   <input
     class="pui-box"
+    {...guard}
     type="checkbox"
     aria-label={label ? null : name || null}
     checked={field.value}
--- a/ui/src/lib/plugin-ui/PuiFormInputs.browser.test.ts
+++ b/ui/src/lib/plugin-ui/PuiFormInputs.browser.test.ts
@@ -210,6 +210,84 @@ describe("plugin-ui input nodes", () => {
   });
 });
 
+describe("plugin-ui autofill guard", () => {
+  // Issue #1978. What is verifiable here is the RENDERED CONTRACT — the attributes are present
+  // and correct. Reproducing the original fill would need a real password manager holding a
+  // saved credential for the origin, which this harness does not have, so these tests assert
+  // the guard is in place rather than that a filler was defeated.
+  beforeEach(() => {
+    mockFetch();
+  });
+
+  const OPT_OUTS: [string, string][] = [
+    ["data-1p-ignore", ""],
+    ["data-bwignore", ""],
+    ["data-lpignore", "true"],
+    ["data-form-type", "other"],
+    ["data-protonpass-ignore", ""],
+  ];
+
+  /** Every rendered control — the checkbox wears `.pui-box`, not `.pui-input`. */
+  function controls(container: HTMLElement) {
+    return [...container.querySelectorAll<HTMLInputElement | HTMLSelectElement>("input, select")];
+  }
+
+  /** One of each node type, in this DOM order: plain, secret, select, checkbox, number. */
+  function everyNodeType() {
+    return renderForm([
+      { type: "text-input", props: { name: "relayUrl" } },
+      { type: "text-input", props: { name: "token", secret: true } },
+      { type: "select", props: { name: "verbosity", options: [{ value: "quiet" }] } },
+      { type: "checkbox", props: { name: "relayEvents" } },
+      { type: "number", props: { name: "retries" } },
+    ]);
+  }
+
+  it("puts a non-guessable name and every vendor opt-out on all four node types", async () => {
+    const { container } = await everyNodeType();
+    const all = controls(container);
+    expect(all).toHaveLength(5);
+
+    for (const el of all) {
+      expect(el.getAttribute("name")).toMatch(/^pui-[0-9a-f]{8}$/);
+      for (const [attr, value] of OPT_OUTS) expect(el.getAttribute(attr)).toBe(value);
+    }
+    // Unique per instance — a shared token would let a heuristic correlate the fields again.
+    expect(new Set(all.map((el) => el.getAttribute("name"))).size).toBe(all.length);
+  });
+
+  it("says new-password on the secret field, off on the rest, nothing on the checkbox", async () => {
+    const { container } = await everyNodeType();
+    const [plain, secret, select, checkbox, number] = controls(container);
+
+    // Chromium ignores `off` on a password field by policy; `new-password` is the value it
+    // honours, and denying the sign-in reading is what protects the field NEXT to this one.
+    expect(secret.getAttribute("autocomplete")).toBe("new-password");
+    expect(plain.getAttribute("autocomplete")).toBe("off");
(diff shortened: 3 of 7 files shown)


scoop (commented):
## What I verified

**Coverage is complete.** `ui/src/lib/plugin-ui/registry.ts:38-41` registers exactly four input node types (`text-input`, `select`, `checkbox`, `number`); the diff touches all four (`PuiTextInput.svelte:40`, `PuiSelect.svelte:50`, `PuiCheckbox.svelte:29`, `PuiNumberInput.svelte:43`). No fifth control renders an `<input>`/`<select>`/`<textarea>` — I read every `Pui*.svelte` name in `ui/src/lib/plugin-ui/` and the remaining components are display nodes plus `PuiActionButton` (a `<button>`).

**The submit contract is genuinely untouched by the randomized DOM `name`.** `PuiActionButton.svelte:48-50` builds the POST body from `form?.snapshot()`, i.e. the context form scope — never `FormData` or a DOM `name` lookup. The scope keys come from the plugin's own `name` prop via `pluginField(() => name, …)` → `scope.set(key, …)` at `ui/src/lib/plugin-ui/field.svelte.ts:47-49`. And `PluginUIRoot.svelte:20-29` sets up that scope as a plain object; there is no `<form>` element anywhere in the plugin-ui tree (I grepped `<form` across `ui/src` — the only matches are `CloneRepo`, `EpicDiagnoseEntry`, `ForkRepo`, `Login`, `NewTask`, none of which wrap a plugin panel; the panel mounts at `ui/src/lib/components/settings/PluginLoadedCard.svelte:126`). So adding `name` attributes introduces no native-submission or `FormData` surface. `autofill.ts:41-43`'s claim that the rendered name is decorative is accurate.

**`crypto.getRandomValues` is the right choice and the comment explaining it is factually correct.** `getRandomValues` is not `[SecureContext]`-gated (only `crypto.subtle` and `crypto.randomUUID` are), so `autofill.ts:51-54` is right that no fallback branch is needed. SSR is off (`ui/src/routes/+layout.ts:1`), so there is no server-side `crypto` or hydration-mismatch path to worry about.

**Vendor attribute spelling is consistent with existing repo precedent.** `data-1p-ignore` already appears as a bare attribute at `ui/src/lib/components/BroadcastDialog.svelte:192`, `ComposeBar.svelte:336`, `GitRail.svelte:746`, `NewTask.svelte:1596`, `NewProject.svelte:207`, `learnings-drawer/ProposedRuleCard.svelte:68`. `autofill.ts:28` emits `""`, which serialises to `data-1p-ignore=""` — the same thing a bare attribute produces. The five keys in `autofill.ts:28-32` match the five asserted in `PuiFormInputs.browser.test.ts:223-227` exactly (name and value), so the test and the source cannot drift apart silently.

**Attribute source order is safe.** In each component `{...guard}` is placed before the type-determining attribute, so `type="checkbox"` (`PuiCheckbox.svelte:30`) and `type={secret ? "password" : "text"}` (`PuiTextInput.svelte:41`) are applied before `checked`/`value`. `guard` contains only `name` + the five `data-*` keys (`autofill.ts:58`) — no collision with `class`, `type`, `value`, `checked`, `aria-label` or `autocomplete`, so the spread cannot clobber a later attribute or vice versa.

**Test assertions line up with the rendered DOM order.** `everyNodeType()` (`PuiFormInputs.browser.test.ts:236-244`) renders plain / secret / select / checkbox / number, and `controls()` (`:231-233`) uses `querySelectorAll("input, select")`, which returns document order — so the destructure at `:261` is correct, and `toHaveLength(5)` at `:249` guards it. The third test's `const [key] = controls(container)` (`:285`) picks the secret `buzzKey` input, which is the first child rendered at `:276`.

**No scope creep.** The diff is exactly the seven files the plan listed, and nothing in it contradicts the plan's `Out of scope` boundary — `field.svelte.ts` and `PluginUIRoot.svelte` are untouched, and no host (non-plugin) input was modified.

## Stated limitations (could not verify locally)

- `node_modules` is absent in this worktree, so I could not read Svelte's `set_attributes` / `init_select` implementation for the pinned version (`ui/package.json:36` → `svelte: ^5.56.8`). The one place that matters is `PuiSelect.svelte:50-53`: adding a spread moves `value={field.value}` from Svelte's dedicated select-value path onto the generic spread path, and select value assignment is order-sensitive with respect to the `{#each}` options at `:56-58`. Svelte documents spread-on-`<select>` as supported and I have no evidence it is broken — I am explicitly not calling this a defect — but it is the one behavioural change in the diff I could not confirm against ground truth. See the nit below for the cheap way to close it.
- Password-manager and Chromium autofill behaviour is external to this repo; the `new-password` reasoning in `PuiTextInput.svelte:10-15` and `autofill.ts:11-23` is assumed as written. It is a documented, operator-approved trade-off in the plan, not something I can verify here.

## Nits (non-blocking):

- `ui/src/lib/plugin-ui/PuiFormInputs.browser.test.ts`: nothing asserts the `<select>`'s **rendered** DOM value after the spread was added — `PuiFormInputs.browser.test.ts:165-176` only checks the submitted body, which comes from the form scope, not the DOM. A one-line `expect(select.value).toBe("quiet")` in the new suite would pin the exact behaviour the spread changed (and would be the regression guard for the visible-vs-submitted divergence the issue's second half describes).
- `docs/plugins.md:395-396`: "the host renders each control with `autocomplete` disabled" is loose — the `secret` control renders `autocomplete="new-password"` (not disabled) and the checkbox renders none. Worth a half-sentence, since a plugin author will see Chrome's "Suggest strong password" bubble on a `secret` token field and wonder whether the guard is working.
- `ui/src/lib/plugin-ui/autofill.ts:37-39`: the doc comment says "never inside `$derived`", which is good; a lint-visible guard is impossible here, so this is purely a note that the four call sites (`PuiTextInput.svelte:22`, `PuiSelect.svelte:18`, `PuiCheckbox.svelte:14`, `PuiNumberInput.svelte:17`) all correctly call at init and none is reactive — I checked each.

PR #1989, opened 2026-08-03
feat(api): export a whole session by Task-ID (transcript + diff + metadata)

### LINKED ISSUE #1268: API-Endpoint: komplette Session per Task-ID exportieren (Transcript + Diff + Metadaten) für Analyse / Re-Launch in anderem CLI/Modell
## Ziel

Ein **API-Endpoint, der zu einer Task-ID (`desig`, z.B. `TASK-435`) alles zurückgibt**: das vollständige Session-Transcript, die Diffs und die Metadaten — in einem Aufruf. Use-Cases:

- **Externe Analyse** einer Session (auswerten, was passiert ist) ohne Zugriff auf die DB/Platte.
- **Session in einem anderen CLI/Modell neu starten** (Prompt + vollständige Historie exportieren → in ein anderes Tool/Modell überführen).

## Ist-Zustand (warum heute nicht möglich)

- Lese-Endpoints existieren nur, sind aber **über die interne Session-`id` (UUID)** gekeyt, **nicht über `desig`**, und **fragmentiert** auf mehrere Routen:
  - `GET /api/sessions/:id/activity` — geparstes Transcript / Aktivität (`src/server.ts:1821`)
  - `GET /api/sessions/:id/diff` — Diff (`src/server.ts:1822`)
  - `GET /api/sessions/:id/usage` — Token/Kosten (`src/server.ts:1820`)
  - (Routing-Übersicht: `src/server.ts:1813-1827`)
- Es gibt **kein** Lookup per Task-ID. `desig` wird nur beim Erzeugen/Archiv-Listing verwendet (`src/store.ts:673`, `:2655`), es existiert keine `getByDesig`-Funktion.
- Es gibt **kein** rohes/vollständiges Transcript-Passthrough — `activity` liefert eine geparste Aktivitäts-Sicht, nicht die rohe JSONL, die ein anderes Modell/Tool re-ingestieren könnte.
- Das echte Transcript liegt als JSONL unter `~/.claude/projects/<dashified-worktree>/<claudeSessionId>.jsonl` (`jsonlPathFor`, `src/usage.ts:13`).

## Vorschlag

**Neuer Aggregat-Endpoint, gekeyt auf die Task-ID:**

```
GET /api/tasks/:desig/export
```

(Alternativ `GET /api/sessions/by-desig/:desig/export` — Naming im Review festzurren.)

Liefert ein Bundle, z.B.:

```jsonc
{
  "meta": {
    "desig": "TASK-435",
    "id": "<shepherd session uuid>",
    "agentProvider": "claude" | "codex",
    "agentSessionId": "<claudeSessionId / native codex id>",
    "prompt": "...", "model": "...", "status": "...",
    "repoPath": "...", "branch": "...", "baseBranch": "...",
    "createdAt": 0, "updatedAt": 0, "archivedAt": null,
    "usage": { /* Token/Kosten */ }
  },
  "transcript": { "format": "jsonl", "raw": "...", "entries": [ /* geparst */ ] },
  "diff": "..."
}
```

### Umzusetzen

1. **Store-Lookup** `getByDesig(desig)` ergänzen (`src/store.ts`) — inkl. archivierter Sessions (Analyse passiert oft nachträglich).
2. **Aggregat-Handler** der intern die schon vorhandenen Bausteine bündelt: Diff (`sessionDiffRead`), Usage (`sessionUsageRead`), geparste Aktivität (`sessionActivityRead`) **plus** die **rohe JSONL** (neues Passthrough über `jsonlPathFor`), damit ein anderes CLI/Modell die Historie re-ingestieren kann.
3. **Routing** in `handleSessionReads`/Router ergänzen (`src/server.ts:1813`), Auth/Token konsistent zu den bestehenden `/api/sessions`-Reads.
4. **Provider-agnostisch** halten: `agentSessionId` statt hartem `claudeSessionId`; für Codex hängt die Transcript-Vollständigkeit an **Issue 1 (#1267)** (native Codex-ID + Transcript-Resolver). Bis dahin sauber degradieren (Metadaten + Diff liefern, Transcript-Feld leer/teilbefüllt markieren).

## Akzeptanzkriterien

- [ ] `GET /api/tasks/TASK-435/export` liefert in **einem** Aufruf Metadaten + vollständiges Transcript (roh **und** geparst) + Diff.
- [ ] Funktioniert auch für **archivierte** Sessions.
- [ ] Rohe JSONL ist enthalten, sodass die Session extern analysiert oder in einem anderen CLI/Modell rekonstruiert/neu gestartet werden kann.
- [ ] Antwort ist provider-agnostisch (Claude vollständig; Codex so vollständig wie Issue 1 es erlaubt, sonst klar markierte Lücke).
- [ ] Auth/Token-Verhalten identisch zu den bestehenden `/api/sessions/:id/*`-Reads.

## Bezug

Vollständige Codex-Abdeckung dieses Endpoints hängt an **Issue 1 (#1267)** (native Codex-Session-ID + Transcript-Resolver).

scoop:
Closes #1268.

One authenticated GET returns **everything** Shepherd knows about a task — keyed on the designation an operator actually sees, not the internal session UUID.

```
GET /api/tasks/:key/export      → { meta, transcript, diff }
GET /api/tasks/:key/transcript  → the untruncated raw JSONL (application/x-ndjson)
```

`:key` is whichever identifier the caller holds: `TASK-435`, the bare `435`, or the session UUID. **Archived sessions included** — post-hoc analysis is the point.

### Acceptance criteria

| Criterion | Status |
| --- | --- |
| Metadata + full transcript (raw **and** parsed) + diff in one call | ✅ |
| Works for archived sessions | ✅ transcript survives archive; the diff is explicitly marked `worktree-removed` |
| Raw JSONL included, re-ingestible externally | ✅ byte-identical passthrough, capped at 8 MiB with a streaming route for the rest |
| Provider-agnostic | ✅ `agentSessionId`, not `claudeSessionId`; Codex degrades with a marked gap (#1267) |
| Auth identical to the existing `/api/sessions/:id/*` reads | ✅ same cookie/bearer gate, covered by a test |

### Decisions worth a reviewer's eye

**The 8 MiB inline cap.** Transcripts here run median 134 KB / p90 551 KB, but the largest on the live box is **32 MB** — ~65 MB once JSON-escaped. The bundle inlines up to the cap, always cut back to the last complete line so `raw` stays **valid JSONL** (never a half record), and reports `rawBytes` (full on-disk size) + `truncated` so a consumer knows to stream `/transcript` instead. `entries` is parsed from the **whole** file regardless of the cap.

**Not on the agent ingress.** Registered in the main app only. It is deliberately absent from `isAgentIngressRoute` — that loopback listener is auth-*exempt* by design, and a full transcript is the most sensitive read in the API; exposing it there would let any spawned agent dump another session's whole history. Asserted by a test.

**Gaps are marked, never silent** — `codex-pending-1267`, `no-transcript-id`, `file-missing`, `worktree-removed`, `no-branch`. A consumer can always tell "nothing happened" from "Shepherd could not tell you". The diff's worktree check probes disk rather than reading `status === "archived"`: a non-isolated session runs in the live checkout and keeps a usable diff after archive.

**`spawnAccountDir` is threaded into `jsonlPathFor`.** A swap/pool session writes its JSONL under `<account>/projects`; omitting it resolves a nonexistent path and yields a silently empty transcript. Note the existing `sessionActivityRead` has exactly that bug — surfaced here, **not** fixed in this PR (out of scope).

### Verification

24 new tests, plus an empirical run against a **snapshot of the live DB and the real `~/.claude/projects` transcripts** (in-process `app.fetch`, no listener/pollers, live state untouched):

| Case | Result |
| --- | --- |
| `TASK-2034` (live, done) | 988 KB raw, 42 entries, diff computed |
| `2033` (bare number → archived) | resolved; snapshot usage; `diff: null (worktree-removed)` |
| `TASK-2014` (**32 MB** transcript) | `truncated: true`, 7.3 MB inline, all 447 retained lines parse as JSON; `/transcript` streamed the full 32,168,888 bytes |
| unknown key | clean 404 |

The truncation invariant was checked directly: the bundle's 264 entries match the ground-truth count parsed off disk, while the capped `raw` alone would yield only 58 — so `entries` genuinely covers the whole file. Worst-case handler time (the 32 MB session): **68 ms**.

`bun run lint`, `bunx tsc --noEmit`, `bun test ./test` (8083 pass), and `fallow audit --base origin/main` (no issues in changed files) all green.

[no-feature-entry] — server-only, no UI surface.

### DIFF
--- a/docs/external-task-api.md
+++ b/docs/external-task-api.md
@@ -162,3 +162,88 @@ Once a task exists, an external agent can also drive it:
   session or an out-of-root `path`.
 
 All of these honor the same auth/origin rules as task creation.
+
+## Exporting a whole session by Task-ID
+
+The inspection routes above are keyed on the internal session UUID and split
+across several calls. For **analysing a finished session** or **re-launching it
+in another CLI/model**, one endpoint returns everything at once, keyed on the
+designation an operator actually sees:
+
+```bash
+curl -H "Authorization: Bearer $SHEPHERD_TOKEN" \
+  http://localhost:7330/api/tasks/TASK-435/export
+```
+
+- `GET /api/tasks/:key/export` — metadata (incl. token usage), the full
+  transcript (raw JSONL **and** parsed), and the diff.
+- `GET /api/tasks/:key/transcript` — the untruncated raw JSONL as an
+  `application/x-ndjson` download.
+
+`:key` is whichever identifier you hold: `TASK-435`, the bare number `435`, or
+the session UUID. **Archived sessions are included** — post-hoc analysis is the
+point. The bare-number form matches on the numeric suffix, so `5` resolves the
+zero-padded `TASK-05`; `TASK-…` is the unambiguous form.
+
+```jsonc
+{
+  "meta": {
+    "desig": "TASK-435",
+    "id": "<uuid>",
+    "name": "...",
+    "prompt": "...",
+    "agentProvider": "claude",
+    "agentSessionId": "<claude session / codex rollout id>",
+    "model": "opus",
+    "effort": null,
+    "status": "archived",
+    "lastState": "done",
+    "repoPath": "...",
+    "branch": "...",
+    "baseBranch": "main",
+    "worktreePath": "...",
+    "isolated": true,
+    "issueNumber": 1268,
+    "createdAt": 0,
+    "updatedAt": 0,
+    "archivedAt": 0,
+    "archiveReason": null,
+    "usage": {/* same shape as GET /api/sessions/:id/usage */},
+  },
+  "transcript": {
+    "format": "jsonl",
+    "path": "/home/…/<agentSessionId>.jsonl",
+    "raw": "…", // capped at 8 MiB — always cut on a line boundary
+    "rawBytes": 551321, // FULL size on disk, not the length of `raw`
+    "truncated": false,
+    "unavailable": null,
+    "entries": [/* the COMPLETE parsed activity, not the 30-entry live tail */],
+  },
+  "diff": {/* same shape as GET /api/sessions/:id/diff */},
+  "diffUnavailable": null,
+}
+```
+
+### Gaps are marked, never silent
+
+A field is only empty when there is genuinely nothing there; anything Shepherd
+could not resolve says so:
+
+| Field                    | Value                | Meaning                                                                   |
+| ------------------------ | -------------------- | ------------------------------------------------------------------------- |
+| `transcript.unavailable` | `codex-pending-1267` | Non-Claude provider — native transcript resolution lands with issue #1267 |
+|                          | `no-transcript-id`   | Session predates the pinned agent session id; nothing to resolve          |
+|                          | `file-missing`       | Path resolved, but the JSONL is gone from disk                            |
+| `transcript.truncated`   | `true`               | Inline `raw` hit the 8 MiB cap — fetch `/transcript` for the full stream  |
+| `diffUnavailable`        | `worktree-removed`   | Archived isolated session: the transcript survives, the worktree does not |
+|                          | `no-branch`          | Non-isolated session — there is no branch to diff                         |
+|                          | _an error message_   | `git` failed; the rest of the bundle is still served                      |
+
+A truncated `raw` is still **valid JSONL** — it is cut back to the last complete
+line, never mid-record — so it can be parsed as-is. (If a single record exceeds
+the cap, `raw` is `""` and `rawBytes` tells you to stream instead.)
+
+Both routes sit behind the same operator cookie/bearer gate as every other read.
+They are deliberately **not** reachable through the loopback agent ingress: that
+listener is auth-exempt by design, and a full transcript is the most sensitive
+read in the API.
--- a/src/server.ts
+++ b/src/server.ts
@@ -89,6 +89,7 @@ import {
   resolveScratchpadUploadDir,
   placeScratchpadUpload,
 } from "./scratchpad";
+import { buildTaskExport, resolveTranscript } from "./task-export";
 import { listWorktree, resolveWorktreeFile } from "./worktree-files";
 import { loadSteers, saveSteers } from "./steers";
 import { loadIcons, setIcon } from "./project-icons";
@@ -566,7 +567,7 @@ export interface AppDeps {
  *  `available: false` is reserved for genuinely absent data — a readable-but-empty transcript
  *  is `available: true, total: 0`. Splits/messageCount/byModel are null when the resolved
  *  source doesn't carry them; `byModel` is always RAW tokens per model, never weighted units. */
-interface SessionUsageDto {
+export interface SessionUsageDto {
   available: boolean;
   source: "live" | "snapshot" | "codex" | "none";
   total: number;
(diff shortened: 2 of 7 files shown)


scoop (commented):
## Verdict

The diff implements exactly what issue #1268 asks for, and every dependency it relies on resolves. I found no correctness or security defect that blocks.

**Routing + auth.** `handleTaskExport` (`src/server.ts:2372`) matches `GET /api/tasks/:key/{export,transcript}` only, and is registered at `src/server.ts:7519` in `ROUTE_HANDLERS`. I grepped the whole file for `"tasks"` — line 2373 is the only occurrence, so no earlier handler shadows it and none of the later ones use a loose `pathname.startsWith` match (the only such match is the `/api` catch-all 404 at `src/server.ts:7600`). The auth gate is the shared one: `makeApp` runs `checkAuth(req)` at `src/server.ts:7584` before dispatch, and `serve()` does the same at `src/server.ts:7824` — identical to `/api/sessions/:id/diff`. The ingress exclusion is real, not just asserted in a comment: `isAgentIngressRoute` returns false unless `parts[1] === "sessions"` (`src/server.ts:7632`), and `makeAgentIngressApp` 404s before delegating (`src/server.ts:7662`).

**Store lookup.** `getByDesig` (`src/store.ts:2346`) reuses `COLS` + `hydrate` exactly like `get()` (`src/store.ts:2330`), and its bare-number branch uses the same `CAST(SUBSTR(desig, DESIG_PREFIX.length + 1) AS INTEGER)` shape as the existing desig-sequence seed query (`src/store.ts:1139`), with `DESIG_PREFIX = "TASK-"` at `src/store.ts:364`. The `where` fragment is built from constants only and the value is bound as a parameter — no injection surface. No status filter, so archived rows resolve, which is the acceptance criterion.

**Transcript resolution.** `jsonlPathFor(worktreePath, claudeSessionId, spawnAccountDir)` — the 3-arg form — matches the signature at `src/usage.ts:25-30`, so swap/pool sessions resolve under `<spawnAccountDir>/projects` (`src/usage.ts:18`). `parseActivity(text, -1)` does return every entry: the guard is `limit >= 0 ? entries.slice(-limit) : entries` (`src/activity.ts:190`). `ActivityEntry` is a 4-field record (`src/activity.ts:5-10`) with summaries truncated to 60 chars, so the uncapped `entries` array cannot blow up the response the way an uncapped `raw` would — the cap-only-`raw` design holds.

**Diff.** `resolveDiffBase(s, prCache, resolveForge)` matches `src/diff-base.ts:26-30`; `computeDiff(worktreePath, base, branch)` matches `src/diff.ts:308`; `toSessionDiff` at `src/diff.ts:217`. `AppDeps` really carries `resolveForge` (`src/server.ts:297`) and `prCache` (`src/server.ts:328`), and `TaskExportDeps.prCache: Pick<PrCache, "get">` is satisfied by `PrCache` (`src/pr-poller.ts:14`).

**Imports.** `DiffResult` `src/types.ts:409`, `AgentProvider` `src/types.ts:14` (only `"claude" | "codex"`, so the `codex-pending-1267` marker can't mislabel a third provider), `GitForge` `src/forge/types.ts:359`, `ActivityEntry` `src/activity.ts:5`. `SessionUsageDto` is newly exported at `src/server.ts:570` and imported into `src/task-export.ts:20` as `import type`, so the server↔task-export cycle is type-only and erased — no runtime cycle. `Buffer` is used unimported, consistent with e.g. `src/argv-limit.ts:59`.

**`capRaw` byte handling.** `Buffer.from(text).subarray(0, cap).toString("utf8")` can end in a U+FFFD when the cap splits a multi-byte sequence, but `lastIndexOf("\n")` then cuts before it — and `\n` (0x0A) can never appear as a UTF-8 continuation byte, so the retained prefix is always whole lines. The single-record-over-cap case correctly yields `""` with `truncated: true`.

**Path safety of the download route.** The streamed path is built only from DB columns; `claudeSessionId` is a server-generated `randomUUID()` (`src/service.ts:3082`, `:3397`), so no caller-controlled component reaches the filesystem. `attachmentDisposition` (`src/scratchpad.ts:206`) sanitises control chars/quotes, so a desig cannot inject a header.

**Docs manifest parity.** `ui/src/lib/docs-manifest.ts` is generated by `ui/scripts/gen-docs-manifest.ts`; `keywordsOf` (line 77) is `description + H2/H3 headings`, lowercased and whitespace-collapsed, with backticks stripped by `extractHeadings` (line 58). The two new headings in `docs/external-task-api.md` (`## Exporting a whole session by Task-ID`, `### Gaps are marked, never silent`) lowercase to exactly the appended `"exporting a whole session by task-id gaps are marked, never silent"`, in document order. The freshness gate (`ui/package.json:18`, run by `scripts/pre-push.ts:557`) should pass.

**Cross-origin exposure.** GET is exempt from `checkOrigin` (`src/server.ts:759`), but there are no `Access-Control-Allow-Origin` headers anywhere in `src/`, so a cross-origin page still cannot read this bundle. This matches every existing GET read; the diff introduces no new exposure.

### Acceptance criteria

All five are met: one call returns meta + raw + parsed transcript + diff; archived sessions resolve (transcript survives, diff is explicitly marked `worktree-removed`); the raw JSONL is present and stays valid JSONL when capped, with the uncapped bytes on `/transcript`; Codex degrades to an explicit `codex-pending-1267` marker with a 200, never a 500; auth is the shared operator gate.

## Nits (non-blocking):

- `src/task-export.ts:110` — `rawBytes` is `Buffer.byteLength` of the *decoded* text, but the field doc (line 42) and `docs/external-task-api.md` call it "FULL size on disk"; for a transcript containing invalid UTF-8 the decode substitutes U+FFFD and the number diverges from the real file size. `Bun.file(path).size` would be exact.
- `src/server.ts:2383` — `sessionUsageDto` is awaited *before* `buildTaskExport`, and for a live Claude session it reads and parses the same JSONL a second time (`readSessionUsage`, `src/usage.ts:267`). Two full reads + parses of a transcript that can be tens of MB; `Promise.all` (or threading the already-read text) would halve it.
- `src/task-export.ts:139` — `parseActivity` over the whole file is synchronous CPU on the single Bun loop. The plan's "never `readFileSync`" mitigation covers the read but not the parse; for the cited 32 MB outlier this still stalls the loop, and `slowRequestTimeoutSec` cannot extend the socket budget here because it is POST-only (`src/server.ts:7797`).
- `docs/external-task-api.md` — the `no-branch` row is described as "Non-isolated session"; the actual condition is `!s.branch` (`src/task-export.ts:162`), which is *how* a non-isolated session looks (`src/worktree.ts:83`) rather than the same thing.

- `src/server.ts:2230` — `sessionActivityRead` calls `jsonlPathFor(s.worktreePath, s.claudeSessionId)` without `spawnAccountDir`, so a swap/pool session's `/activity` read resolves a nonexistent path and returns an empty list; the new export correctly passes it and does not inherit the bug.

scoop:
Follow-up filed for the pre-existing bug noted in the PR description: #1990 — `GET /api/sessions/:id/activity` ignores `spawnAccountDir`, so the Activity tab is silently empty for swap-account sessions (209/258 on the live instance). Out of scope here; this PR's export path threads it correctly.

PR #1991, opened 2026-08-03
fix(api): resolve session activity under spawnAccountDir

### LINKED ISSUE #1990: GET /api/sessions/:id/activity ignores spawnAccountDir — Activity tab silently empty for swap-account sessions
## Symptom

The **Activity** tab is empty for most sessions. `GET /api/sessions/:id/activity` returns `[]` even though the session has a full transcript on disk.

## Root cause

`sessionActivityRead` (`src/server.ts` → `sessionActivityRead()`) omits `spawnAccountDir` when resolving the transcript path:

```ts
const path = s.claudeSessionId ? jsonlPathFor(s.worktreePath, s.claudeSessionId) : "";
```

A session spawned into a swap/pool account writes its JSONL under `<spawnAccountDir>/projects/…`, not the server's own `config.claudeProjectsDir` (see `projectsDirFor` in `src/usage.ts`). Without the third argument the read resolves a **nonexistent path**, and `sessionActivity()` degrades a missing file to `[]` — so the failure is silent: no error, no log, just an empty feed.

## Impact (measured on the live instance)

- **209 of 258 sessions** have a non-empty `spawnAccountDir`.
- Spot-checked the three most recent (TASK-1996/1997/1998): the default path does **not** exist, the account path **does**. So the data is there; the endpoint looks in the wrong place.
- User-visible: `ActivityFeed.svelte` polls this endpoint every 5s and renders `entries` with no fallback (mounted from `Viewport.svelte`). The tab just shows nothing.

## Fix

Pass the argument, as every other consumer already does:

```ts
const path = s.claudeSessionId
  ? jsonlPathFor(s.worktreePath, s.claudeSessionId, s.spawnAccountDir)
  : "";
```

Correct precedents to match: `sessionDiffAnnotationsRead` (same file), `src/poller.ts`, `src/recap.ts`, `src/process-reaper.ts`, `src/usage-snapshot.ts`, `src/usage-breakdown.ts`.

## Test

A regression test needs to assert the resolved **path**, not just a non-empty result — the current shape fails open, so a test that only checks "entries came back" would still pass against the default dir when both happen to exist. Write the JSONL under a temp `spawnAccountDir` only, and assert the endpoint finds it.

## Provenance

Found while implementing #1268 (PR #1989). The new `/api/tasks/:key/export` threads `spawnAccountDir` correctly; this pre-existing bug was deliberately left out of that PR's scope.

scoop:
Fixes #1990.

## The bug

`sessionActivityRead()` in `src/server.ts` called `jsonlPathFor(s.worktreePath, s.claudeSessionId)` — two args. `projectsDirFor()` (`src/usage.ts`) roots a transcript under `<spawnAccountDir>/projects` whenever that column is set (claude-swap points `CLAUDE_CONFIG_DIR` there), so dropping the third argument resolved the server's own projects dir: a path that does not exist for such a session.

`sessionActivity()` degrades a missing file to `[]`, so the failure was silent — no throw, no log, no 5xx. `ActivityFeed.svelte` just rendered an empty tab. Per the issue, 209 of 258 live sessions carry a non-empty `spawnAccountDir`.

Every other consumer already threads the argument (`poller`, `recap`, `process-reaper`, `usage-snapshot`, `usage-breakdown`, `task-export`, `index`) — including `sessionDiffAnnotationsRead()` **immediately below** the broken call, whose comment already named this exact hazard. This one caller was the outlier.

## The fix

One argument, plus a comment recording why it is load-bearing so the next edit doesn't drop it again.

## The test is the real work

The endpoint returns a bare array and fails open, so asserting "entries came back" would pass against the default projects dir whenever a fixture also happens to sit there. The regression test pins the path by construction instead:

- `config.claudeProjectsDir` points at an empty temp dir (restored in `finally`).
- The JSONL fixture is written **only** under `<spawnAccountDir>/projects/<dashified worktree>/<id>.jsonl`, so a non-empty response is reachable through exactly one path on disk.
- Negative control in the same test: an otherwise identical session with `spawnAccountDir` left `NULL` still returns `[]` — that pins the direction, so a hypothetical "search both dirs" read would not pass either.

Confirmed the test binds: it fails against the pre-fix line (`[]` vs `["Edit"]`) and passes after it.

## Scope

Deliberately fix + test only. No warn-log and no DTO reshape to distinguish "no activity yet" from "transcript not found" — the endpoint keeps its fail-open contract and bare-array wire shape.

Not fixed here: `sessionActivityRead()` has no Codex branch, so a Codex session's Activity tab is empty for a *different* reason (its rollout lives outside `~/.claude/projects` and needs `CodexRolloutResolver`, which isn't wired into `AppDeps` for this handler). Pre-existing and distinct from #1990 — filed as a follow-up.

## Verification

`bun test ./test` → 8084 pass / 0 fail · `bun run lint` clean · `bunx tsc --noEmit` clean · `fallow audit --base origin/main` → no issues in the 2 changed files. Server-only change, so no UI half, no i18n keys, no feature-catalog entry (ships as `fix:`).

### DIFF
--- a/src/server.ts
+++ b/src/server.ts
@@ -2226,8 +2226,13 @@ async function sessionUsageRead(id: string, deps: AppDeps): Promise<Response> {
 async function sessionActivityRead(id: string, deps: AppDeps): Promise<Response> {
   const s = deps.store.get(id);
   if (!s) return json({ error: "not found" }, 404);
-  // pre-feature session (no pinned id) → no transcript to read
-  const path = s.claudeSessionId ? jsonlPathFor(s.worktreePath, s.claudeSessionId) : "";
+  // pre-feature session (no pinned id) → no transcript to read.
+  // spawnAccountDir MUST be passed (#1990): a spawn-account session writes its JSONL under
+  // <account>/projects (see src/usage.ts) — omitting it resolves a nonexistent path, which
+  // sessionActivity() degrades to [], so the Activity tab silently shows nothing.
+  const path = s.claudeSessionId
+    ? jsonlPathFor(s.worktreePath, s.claudeSessionId, s.spawnAccountDir)
+    : "";
   return json(path ? await sessionActivity(path) : []);
 }
 
--- a/test/server.test.ts
+++ b/test/server.test.ts
@@ -20,6 +20,7 @@ import { WorktreeMgr } from "../src/worktree";
 import type { GitForge } from "../src/forge/types";
 import { config, USAGE_HISTORY_RETENTION_MS } from "../src/config";
 import { ACTIVE_LABEL } from "../src/drain-core";
+import { dashify } from "../src/usage";
 
 // Create a real tmp dir inside config.repoRoot so validation passes
 let tmpRoot: string;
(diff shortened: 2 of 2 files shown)


scoop:
Follow-up for the out-of-scope Codex gap noted above: #1992

scoop (commented):
## Verdict

The diff does exactly what the task asks and nothing more: `sessionActivityRead()` now threads `s.spawnAccountDir` into `jsonlPathFor()`, plus one regression test. I could not find a correctness, security, or contract problem.

## What I verified

**The fix itself (`src/server.ts:2233-2235`).**
- `jsonlPathFor` accepts a third `spawnAccountDir?: string | null` and roots the path via `projectsDirFor()` — `src/usage.ts:24-30`, `src/usage.ts:18-20`. `projectsDirFor(null)` returns `config.claudeProjectsDir`, so sessions without an account dir behave exactly as before — no regression for the ~49/258 rows the issue describes.
- `Session.spawnAccountDir` is `string | null` (`src/store.ts:610`), is in the `COLS` select list (`src/store.ts:549`), and is hydrated by `hydrate()` (`src/store.ts:5369`), which is what `store.get()` returns (`src/store.ts:2329-2334`). So the value the handler passes is really populated, not silently `undefined`.
- The type matches the parameter (`string | null` vs `string | null | undefined`) — no cast, no widening.
- The call is now identical in shape to `sessionDiffAnnotationsRead()` in the same file (`src/server.ts:2268`), the precedent the issue names.
- The new comment is factually accurate: `sessionActivity()` does return `[]` for a missing file (`src/activity.ts:216-223`), and the account-rooted write path is what `src/usage.ts:12-20` documents. No stale or misleading claim.
- Sweep of all `jsonlPathFor(` call sites: every session-transcript caller now passes the third argument (`src/poller.ts:160,173,185,411,929`, `src/process-reaper.ts:609`, `src/recap.ts:136`, `src/task-export.ts:103`, `src/usage-breakdown.ts:111`, `src/usage-snapshot.ts:52`, `src/index.ts:1882,2298`, `src/server.ts:2234,2268`). Only one 2-arg call remains, outside this diff — noted below.
- No signature or wire-shape change, so no consumer to re-check: the handler still returns a bare `ActivityEntry[]` (`src/activity.ts:5-10`), which is what `ActivityFeed.svelte` already renders. Route dispatch is unchanged (`src/server.ts:2334`).

**The test (`test/server.test.ts:518-566`).**
- `dashify` is genuinely exported from `src/usage.ts:8`, and the fixture path it builds (`<account>/projects/<dashify(worktree)>/<id>.jsonl`) is byte-for-byte what `jsonlPathFor` composes (`src/usage.ts:29`).
- `store.setSpawnIdentity(id, terminalId, accountDir)` exists with that arity/order and writes the column directly (`src/store.ts:3505-3514`) — necessary, since `create()` always inserts `spawnAccountDir` as `NULL` (`src/store.ts:2319`).
- All fs/path helpers used are already imported in the file (`mkdirSync`, `writeFileSync` — `test/server.test.ts:3-11`; `join` — `:12`).
- **The test actually binds to the fix.** With `config.claudeProjectsDir` repointed at an empty dir and the JSONL written only under the account dir, the pre-fix line resolves `<emptyProjects>/-wt-activity/<id>.jsonl`, which does not exist → `sessionActivity()` returns `[]` (`src/activity.ts:220`) → the `["Edit"]` assertion fails. It does not fail open.
- The expected entry shape is right: a `tool_use` block in `message.content` yields `{tool: "Edit", …}` via `collectEntries` (`src/activity.ts:114-127`), and `parseRecords` keeps the record because `message.content` is an array (`src/activity.ts:85-93`).
- Isolation: `config` is a plain mutable object literal (`src/config.ts:546`, `:572`), the mutation is restored in `finally`, and `projectsDirFor` reads it per call rather than caching at import — so no leakage into sibling tests and no stale-capture hazard.
- The negative control creates a second session with the same `repoPath`/`branch`/`claudeSessionId`. I checked for a uniqueness collision: the only `UNIQUE(repoPath, branch)` in the schema is on `local_prs` (`src/store.ts:1198`), not `sessions`, and the sessions table has no unique index on `claudeSessionId`. The insert is safe.

- `src/codex-activity.ts:452` is the last remaining 2-arg `jsonlPathFor(worktreePath, trackingId)` call (reviewer activity summary); I found no `accountDir` plumbing in `src/review.ts` or `src/plan-gate.ts`, so I could not confirm reviewer spawns ever land in a swap account — flagging it only as the one call site left unthreaded, not as a confirmed miss.
- `sessionActivityRead()` has no Codex branch, so a Codex session's Activity tab stays empty for a different reason (rollout lives outside `~/.claude/projects`); the plan already declares this out of scope and the PR body promises a follow-up issue.

## Nits (non-blocking):
- `test/server.test.ts:551`: `"term_u"` is reused as the spawn terminal id while `USAGE_SESSION.herdrAgentId` is also `"term_u"` — harmless here, but a distinct value would make it obvious the setter, not the fixture, wrote it.

Issue #1175, opened 2026-06-27
Enable bring-back (restore) for Codex sessions


scoop:
## Follow-up from #1174 (bring back a marked-as-done session)

#1174 shipped a **resume-only MVP**: restore only works for a **claude** session row with a non-empty `claudeSessionId`. Codex sessions are deliberately blocked with `409 cannot_restore` (and the UI surfaces a "relaunch instead" hint).

## Why Codex is blocked today

`buildCodexResumeArgv` (`src/service.ts`) resumes via `codex resume --last`. Shepherd does **not** persist a provider-native Codex session id, so `--last` is scoped by the Codex CLI's own selection rules — for concurrent or non-isolated Codex sessions sharing a cwd it can target the *most recent* Codex session for that cwd rather than the exact Shepherd row being restored. Restoring the **wrong** conversation is worse than refusing, so restore blocks Codex rather than guessing.

## What this issue tracks

Enable "bring back" for Codex sessions, reliably:

1. **Persist a provider-native Codex session id** at spawn/ingest (the missing prerequisite) so a specific Codex session can be resumed by id rather than `--last`.
2. Switch `buildCodexResumeArgv` to `codex resume <SESSION_ID>` once that id exists.
3. Drop the codex branch of `service.restore`'s resumability guard (`cannot_restore`) so Codex archived rows become restorable.
4. Extend the Phase-0-style verification to Codex (transcript survival + correct-session resume after worktree re-creation).
5. Tests mirroring the claude restore coverage for the Codex path; UI copy/error mapping update if the guard message changes.

## Out of scope here
The separate **fresh-agent restore variant** (restore committed work but start a brand-new agent when no resumable conversation exists) — tracked/considered independently; not required for Codex resume.

Ref: PR #1174.

scoop:
**Sweep 2026-06-27 — designating this as the owner of the provider-native session-id field (de-overlap).**

Three open Codex issues independently call for persisting a provider-native session id: this issue (step 1), #1087 (step 7 — "supplement `claudeSessionId` with provider-neutral `providerSessionId`/`providerMetadata`"), and #1160 (candidate-a — "pin the session id instead of `--last`"). To avoid three sessions inventing three colliding schemas/migrations, **this issue owns that field** and ships it first as the foundational slice.

[...]

PR #1994, opened 2026-08-03
fix(api): read Codex rollouts in GET /api/sessions/:id/activity

### LINKED ISSUE #1992: GET /api/sessions/:id/activity has no Codex branch — Activity tab empty for Codex sessions
Split out of #1990 (fixed in #1991), which was scoped to the `spawnAccountDir` resolution bug.

## Gap

`sessionActivityRead()` (`src/server.ts`) resolves the transcript with `jsonlPathFor()` unconditionally, i.e. it always looks under a `~/.claude/projects`-shaped path. A Codex session does not write there — its rollout lives under the Codex home and has to be resolved by `CodexRolloutResolver` (by launch-unique cwd), the way `reviewerActivitySummary()` in `src/codex-activity.ts` already dispatches.

So for a session with `agentProvider: "codex"` the endpoint resolves a path that never exists, `sessionActivity()` degrades it to `[]`, and the Activity tab is silently empty — same user-visible symptom as #1990, different cause.

Note the sibling endpoint `GET /api/sessions/:id/usage` *does* branch on provider (`(s.agentProvider ?? "claude") === "codex"` → `UNAVAILABLE_USAGE`), so the provider distinction is already acknowledged elsewhere in this file.

## Fix sketch

Dispatch on `s.agentProvider` in `sessionActivityRead()`: Codex → resolve the rollout via `CodexRolloutResolver` and read it through the Codex activity reader; Claude → today's path. The resolver is not currently wired into `AppDeps` for this handler, so that plumbing is the bulk of the work.

## Test

Mirror the #1991 regression test: write the rollout fixture only at the Codex-resolved location and assert the endpoint finds it — a "non-empty response" assertion alone fails open, since a missing file degrades to `[]`.

scoop:
Closes #1992.

## Problem

`sessionActivityRead()` resolved the transcript with `jsonlPathFor()` unconditionally — always a `~/.claude/projects`-shaped path. A Codex session writes none, so the path never existed, `sessionActivity()` degraded to `[]`, and the Activity tab was **silently empty**. Same user-visible symptom as #1990/#1991, different cause.

## Fix

Dispatch on `s.agentProvider`, the way the sibling `/usage` DTO in the same file already does. Codex reads its rollout under `$CODEX_HOME`.

**Deliberate deviation from the issue's fix sketch.** The issue proposes `CodexRolloutResolver`; that resolver resolves *only on a unique cwd candidate* (`selectCodexRollout` → `c.length === 1`, else `null`). That fail-safe rule is right for its callers — one-shot `source:"exec"` reviewer spawns. A **task** session is `source:"cli"` and long-lived: every restore/relaunch writes a NEW rollout under the SAME worktree cwd (which is why `resolveCodexRestoreId()` re-derives instead of trusting the persisted `providerSessionId`), so ≥2 candidates is the steady state and the resolver would return `null` for the rest of the session's life — bug unfixed. Verified against a real `$CODEX_HOME`: 7 `cli` rollouts sharing one cwd.

So the read uses `findCodexRollout()`'s **newest-wins** rule — the same one `restore()` and the poller's id capture already rely on — extracted from `findCodexSessionId()` so it returns the path as well as the id. `CODEX_ID_SKEW_MS` moves next to that scan (out of `service.ts`) because every caller's mtime window must agree: the id the resume path captures and the rollout the tab shows have to be the same conversation.

**Non-isolated Codex sessions still return `[]`.** Their cwd is the shared repo checkout, also used by sibling sessions, relaunches and the operator's own `codex` runs, so no rollout is attributable to one row. `resolveCodexRestoreId()` refuses restore for exactly this reason (#1175/#1476) — rendering another session's conversation would be worse than rendering nothing.

**`CodexTranscriptLocator`** caches the resolution (30 s TTL) and backs misses off (2 s → 15 s cap). The scan is sync FS work on the single Bun loop and `ActivityFeed.svelte` polls every 5 s; measured over a synthetic 1600-rollout tree: **3.2 ms** warm walk, **33 ms** on a full miss (reads every `session_meta` header). The TTL bounds how long a just-restored session can be read from its previous rollout; file *content* is re-read every request, so live output is never stale. The cap sits well below the reviewer resolver's 60 s so a just-started session's tab fills within one cap at worst.

## Tests

Endpoint tests seed a temp `$CODEX_HOME` with `config.claudeProjectsDir` pointed at an empty dir, so the fixture is reachable through exactly **one** code path — a "non-empty response" assertion alone would fail open, since a missing file degrades to `[]` (the #1991 lesson).

- Codex session resolves its rollout → parsed entries (asserts tool + summary)
- Same fixture, `agentProvider: "claude"` → `[]` — pins the dispatch direction
- **Two `cli` rollouts share the cwd → the newer wins** — the post-restore shape; this is the case the issue's resolver sketch would fail
- Non-isolated Codex → `[]`
- Only cwd match is `source:"exec"` (a reviewer spawn in the same worktree) → `[]`
- Locator unit tests against a fake clock: memoised hit, TTL re-derive, exponential miss backoff + cap + clear-on-hit, per-session isolation, `reset()`
- Claude path unchanged — the #1991 `spawnAccountDir` test and the existing activity tests pass untouched

`bun run lint`, `bunx tsc --noEmit`, `bun test ./test` (8102 pass), and all seven pre-push lanes green.

## Out of scope

The poller's activity signals / heat strip / stall detection, recap, process-reaper, and `/api/tasks/:key/export`'s `codex-pending-1267` marker (#1267) are Claude-only too — same root cause, separate issues. `findCodexRollout()` + the locator make each a small follow-up.

Server-only, no UI surface → `[no-feature-entry]`; no new user-facing strings → no i18n or glossary work.

### DIFF
--- a/src/codex-activity.ts
+++ b/src/codex-activity.ts
@@ -5,9 +5,14 @@ import { readTranscriptTail, type ActivityEntry } from "./activity";
 import { readActivitySignal, signalFrom, type SessionActivity } from "./activity-signal";
 import { snapshotFrom, type ActivitySnapshot } from "./stall";
 import { codexHome, listRolloutFiles } from "./codex-usage";
-import { readSessionMeta } from "./codex-session-id";
+import {
+  CODEX_ID_SKEW_MS,
+  findCodexRollout,
+  readSessionMeta,
+  type CodexRollout,
+} from "./codex-session-id";
 import { jsonlPathFor, readSessionUsage, type SessionUsage } from "./usage";
-import type { AgentProvider } from "./types";
+import type { AgentProvider, Session } from "./types";
 
 const CMD_MAX = 60;
 const DEFAULT_LIMIT = 30;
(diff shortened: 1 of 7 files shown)


scoop (commented):
## What I verified

**The dispatch itself.** `sessionActivityRead` now branches on `(s.agentProvider ?? "claude") === "codex"` (`src/server.ts:2241`), the same normalisation the sibling usage path and the service already use (`src/service.ts:2700`, `src/service.ts:4053`). The Claude arm below it is byte-identical to the pre-PR code including the `#1991` `spawnAccountDir` argument (`src/server.ts:2249-2253`), so that regression stays fixed.

**`findCodexRollout` extraction is behaviour-preserving.** The scan body at `src/codex-session-id.ts:57-66` is the old loop with `return meta.id` replaced by `return { id: meta.id, path }`; `findCodexSessionId` is now a pure delegate (`src/codex-session-id.ts:70-76`). Both surviving call sites pass two arguments (`src/service.ts:2702`, `src/service.ts:4053`), so the `home = codexHome()` default still resolves per call — which is what lets the new endpoint tests swap `$CODEX_HOME` via env (`src/codex-usage.ts:56`).

**`CODEX_ID_SKEW_MS` move.** Grepped the tree: exactly one definition remains (`src/codex-session-id.ts:40`) and every consumer imports it (`src/service.ts:17`, `src/codex-activity.ts:9`, used at `src/codex-activity.ts:544`, `src/service.ts:2702`, `src/service.ts:4053`). No stale local copy, no cycle — `codex-session-id.ts` imports only `codex-usage.ts`, and `codex-activity.ts` imports both, so adding `server.ts → codex-activity.ts` (`src/server.ts:101`) introduces none.

**Type contract.** `pathFor` takes `Pick<Session, "id" | "worktreePath" | "createdAt" | "isolated">`; `isolated` is a required `boolean` (`src/types.ts:30`) and `createdAt` a required `number` (`src/types.ts:110`), so `!s.isolated` is not a tri-state trap and the arithmetic at `src/codex-activity.ts:544` is sound.

**Wire-shape parity.** `codexSessionActivity` defaults to `DEFAULT_LIMIT = 30` (`src/codex-activity.ts:18`), the same value the Claude reader uses (`src/activity.ts:12`, consumed at `src/activity.ts:219`), and `parseCodexActivity` emits the same `ActivityEntry` fields. The Activity tab has no provider gate (`ui/src/lib/components/Viewport.svelte:3010`), so the fix is genuinely user-visible. No message-catalog keys are touched by this diff, so there is no en/de parity question here.

**Locator state machine.** I hand-traced `pathFor` (`src/codex-activity.ts:531-551`) against the fake-clock test: hit inside 30 s returns without scanning; TTL expiry deletes the entry and re-derives; a miss sets `nextAt = now + min(2000·2^(n-1), 15000)` and short-circuits with `now < nextAt`; a hit clears the miss entry. The backoff sequence the test asserts (2/4/8/15/15 s, 7 total scans) matches the implementation exactly. `find` cannot throw in production — `listRolloutFiles` swallows `readdir` failures (`src/codex-usage.ts:92-96`), `rolloutCandidate` swallows `statSync` (`src/codex-usage.ts:62-66`), `readSessionMeta` swallows parse/read failures (`src/codex-session-id.ts:81-100`) — and `makeApp` converts any residual throw to a JSON 500 rather than an HTML page (`src/server.ts:7633-7639`).

**Lazy singleton.** `codexTranscripts(deps)` (`src/server.ts:675`) mutates the one `AppDeps` object that every request's `Ctx` carries (`src/server.ts:7615`), so production shares a single locator per app — same pattern as `learnings`/`repoConfigSvc` at `src/server.ts:669-673`.

**Test fixtures actually exercise the new path.** `listRolloutFiles` only accepts `^rollout-.*\.jsonl$` (`src/codex-usage.ts:103`) and the seeded fixtures are all named `rollout-*.jsonl` (`test/server.test.ts:590-616`); `config.claudeProjectsDir` is pointed at an empty dir for the duration (`test/server.test.ts:626`), so the non-empty assertions cannot fail open.

**Cache-key safety.** I checked whether a session's `worktreePath` can change under a memoised entry: `rename()` moves only the git branch (`src/worktree.ts:231-236`, called from `src/service.ts:4162`) and never the worktree directory, so keying the cache by session id while the resolution depends on `worktreePath` is safe.

## Risk note (not a finding)

The newest-wins rule has a lower mtime bound (`createdAt − skew`) but no upper one, so for an **archived** Codex session whose branch was later deleted and whose prompt-derived slug got reused (`uniqueName` de-dupes against live herdr agents and existing branches only — `src/service.ts:3713-3726`), the worktree cwd can repeat and the archived row's Activity read would surface the *newer* session's rollout. The Claude arm is immune because it keys on the per-session `claudeSessionId`. I am not raising this as a blocking finding: it is the same attribution semantic `resolveCodexRestoreId` already accepts (`src/service.ts:4047-4062`), the plan documents the trade-off explicitly, and in practice an archived session's Activity tab renders the recap instead of the feed once the recap is ready (`ui/src/lib/components/Viewport.svelte:229`, `:3000-3003`). Worth a comment or a follow-up if you disagree.

## Nits (non-blocking):
- `src/codex-activity.ts:566-577`: `codexSessionActivity` reads the whole rollout with `file.text()` on every 5 s poll, while every other Codex reader in this file goes through `readTranscriptTail` (`src/codex-activity.ts:203`); since `parseCodexActivity` keeps only the last 30 entries (`src/codex-activity.ts:157`), a tail read would be equivalent and bounded — the full read is parity with `sessionActivity` (`src/activity.ts:217-224`), so this is a judgement call, but the plan's cost table measured the scan and not the read.
- `src/codex-activity.ts:557-560`: `CodexTranscriptLocator.reset()` has no production caller, whereas its sibling `CodexRolloutResolver.reset()` is invoked on row close (`src/plan-gate.ts:1220`, `src/plan-gate.ts:1617`, `src/review.ts:1704`, `src/review.ts:1782`) precisely so its maps can't grow unbounded — wiring it to session archive/delete would make the "no eviction warranted" comment at `src/codex-activity.ts:527` self-enforcing rather than merely true-by-magnitude.
- `test/server.test.ts:588-635`: `seedCodexHome`/`withCodexHome` re-implement a helper of the same name already in `test/service.test.ts:3097`; a shared test helper would keep the two seeding shapes from drifting.

- `src/codex-activity.ts:566-569`: the `limit` parameter on `codexSessionActivity` is only ever passed by tests; the endpoint always takes the default.
- `src/codex-activity.ts:557`: `reset(id)` is an unused public surface today (see nit above).

Issue #1995, opened 2026-08-03
feat(ui): per-session stop button on top of POST /api/sessions/:id/interrupt


scoop:
Follow-up to #1993, which shipped the endpoint only.

`POST /api/sessions/:id/interrupt` now exists (a lone ESC to one session's pane, 200/404, no body), but nothing in the UI calls it. The gap the parent issue named — "there is currently no per-session stop button" — is still open.

## What's already there

- Mobile/touch `SteerBar` renders `<ControlBar include={["cancel"]} onkey={termSend} />`, which writes a raw Esc byte down the **attached** PTY socket. That covers the **focused** session only, and only where the terminal is attached.
- Desktop operators press Esc in the terminal itself.

So the missing surface is a stop control for a session you are **not** currently focused on / attached to — e.g. from a session card or the session header.

## Work

1. `interruptSession(id)` wrapper in `ui/src/lib/api.ts` (mirror `replySession`; 404 → throw via `failed`).
2. Decide where the control lives (session card action row vs. session header vs. both) and how it reads next to the fleet-wide halt in the gear menu — a per-session stop must not look like an e-stop.
3. EN + DE message keys for label/`aria-label`/failure toast (`check:i18n` parity).
4. One `ui/src/lib/feature-announcements/entries/v<next>-<id>.ts` fragment (`bun run next-version` for the version).
5. Failure handling: the endpoint 404s for a dead pane, which is a normal outcome, not a crash — surface it as a toast, not a thrown error.

Design questions worth settling before building: does the button show only for `running` sessions, and does it do interrupt-only or offer interrupt-then-steer?

scoop:
✅ PR #2007 merged.

PR #1996, opened 2026-08-03
feat(sessions): per-session interrupt (POST /api/sessions/:id/interrupt)

### LINKED ISSUE #1993: feat(sessions): per-session interrupt (POST /api/sessions/:id/interrupt)
There is no way to interrupt **one** session. The only interrupt in the codebase is fleet-wide.

## Today

- `POST /api/sessions/:id/reply` → `operatorReply` → `sendSteerTo` (`src/service.ts`) bracket-pastes the text and sends `\r`. It explicitly accepts a **busy** agent and sends no ESC, so Shepherd neither queues nor interrupts — Claude Code's own composer decides what happens to input that arrives mid-turn.
- `haltAll` (`src/service.ts`, `POST /api/halt`) is the only thing that sends `\x1b`, and it sweeps **every working agent**. It is an e-stop, not a steer.

So an operator who wants to redirect one busy session has two options: type into its composer and hope, or stop the entire fleet.

## Proposal

`POST /api/sessions/:id/interrupt` — send `"\x1b"` to that session's pane, nothing else. Composable with the existing reply route, so a caller that wants cancel-then-re-prompt does interrupt → reply rather than needing a new combined endpoint.

## The part worth deciding deliberately

`haltAll` documents at length why it **bypasses** `#serializeSteer`: an e-stop must never queue behind the very steers it exists to cancel, and it accepts a narrow race where an ESC lands after a paste block has closed, clearing the composed input while the steer's trailing CR submits whatever remains.

A per-session interrupt is not an e-stop, so that reasoning does not transfer unchanged. Going **through** `#serializeSteer` makes interrupt→reply ordering deterministic, at the cost of an interrupt that can block behind a wedged send to that same pane. Going around it inherits `haltAll`'s race for a case where the operator did *not* ask to cancel everything. Worth choosing explicitly rather than copying `haltAll`.

Minor: `haltAll` sends to `agent.terminalId` while `sendSteerTo` uses `s.herdrAgentId` — whichever a per-session interrupt uses should match how the route resolves the session.

## Why now

Two callers want it:

1. **The UI.** There is currently no per-session stop button, because there is no endpoint behind one.
2. **shepherd-buzz-bridge**, which is blocked on it: erwins-enkel/shepherd-buzz-bridge#14. Steering a Shepherd session from a Buzz thread is meant to mirror buzz-acp's default (`MultipleEventHandling` defaults to `Steer` — cancel + re-prompt with merge framing). Without a per-session interrupt the bridge can only queue, which is a different behaviour wearing the same name.

Independently useful — filing it here rather than as a bridge dependency because the UI gap stands on its own.

scoop:
Closes #1993.

There was no way to interrupt **one** session: `POST /api/sessions/:id/reply` sends a bracket-pasted steer + CR (never an ESC), and `haltAll` (`POST /api/halt`) sweeps every working agent. An operator wanting to redirect one busy session could only type into its composer and hope, or stop the whole fleet.

This adds `POST /api/sessions/:id/interrupt` — a lone `\x1b` to that session's pane, nothing else. No body; `200 {ok:true}` when it lands, `404 {error:"not found"}` for an unknown id, dead/unlisted pane, unreachable herdr, or a rejected send. Composable with `/reply`, so cancel-then-re-prompt is interrupt → reply rather than a new combined endpoint.

## The three decisions the issue asked to make deliberately

Each is captured as a comment at the code that embodies it, and pinned by a test.

**1. It goes THROUGH `#serializeSteer`.** `haltAll` bypasses the FIFO because an e-stop must never queue behind the very steers it exists to cancel, and it accepts the resulting race: an ESC landing after a paste block has closed clears the composed input while that steer's trailing CR submits the remnant. A targeted interrupt is a steer, not an e-stop, so that reasoning does not transfer — serializing makes interrupt → reply ordering deterministic. The accepted cost is the mirror image: an interrupt can queue behind a wedged send, since the serializer chain is process-wide rather than per-pane. `/api/halt` stays the un-queued escape hatch.

**2. No `agentStatus === "working"` filter.** `haltAll` filters because a sweep must not touch bystanders; a caller naming one session already picked its target. A status check would cost a second herdr round-trip and still race the send. ESC on an idle pane merely clears composed input.

**3. Pane resolution matches the REPLY route, not `haltAll`** (the issue's "minor" `terminalId` vs `herdrAgentId` note): `liveAgentFor` + `adoptLiveResumeAgent`, so a herdr-restored pane is interrupted at its live `terminalId` and the stale stored id is healed — interrupt and a following reply address the same pane. `operatorReply`'s isolated-Codex **resume** is deliberately not inherited: resuming a dead pane in order to interrupt it is meaningless.

Also: no signal row (`reply` is a distiller-mined kind carrying operator words; an ESC has no payload to mine) and no event (`halt:done` exists because a fleet sweep's reach is otherwise invisible; a single-target interrupt's outcome is its HTTP status).

## Scope

Endpoint only. The UI stop button the issue names is filed as #1995 — the mobile `SteerBar` already exposes an Esc control for the *focused* session via the attached PTY socket, so the remaining gap is a control for non-focused sessions, which is a design question that does not block either caller. The buzz-bridge blocker (erwins-enkel/shepherd-buzz-bridge#14) consumes the HTTP route, which is what it names.

## Tests

`test/service.test.ts` (9): lone ESC with no paste markers and no CR; no signal recorded; idle pane still interrupted; unknown id; dead pane; `herdr.list()` throws → false, never throws; send rejects → false; herdr-restored pane adopts the live id and heals the row; and — the concrete assertion behind decision 1 — an interrupt issued while a reply is mid-flight lands strictly after that reply's CR (`paste, CR, ESC`).

`test/server-steers.test.ts` (3): 200 with the id reaching the service, 404 when it didn't land, and GET not interrupting.

Verified locally: `bun test ./test` (8109 tests, 0 fail), `bun run lint`, `bunx tsc --noEmit`, `bunx fallow audit --base origin/main --fail-on-issues` — all green.

[no-feature-entry] — server-only change, no user-facing UX ships here (see #1995).

### DIFF
--- a/docs/external-task-api.md
+++ b/docs/external-task-api.md
@@ -126,6 +126,15 @@ Once a task exists, an external agent can also drive it:
 
 - `POST /api/sessions/:id/reply` — send follow-up text to the live agent
   (`{ "text": "..." }`).
+- `POST /api/sessions/:id/interrupt` — interrupt that one session: a lone ESC to
+  its pane and nothing else (no body). The per-session counterpart to the
+  fleet-wide `POST /api/halt`, and composable with `/reply`, so cancel-then-re-prompt
+  is interrupt → reply. Unlike `/api/halt` it does not skip sessions that aren't
+  currently working (the caller named its target) and it is serialized against
+  other steers, so an interrupt issued before a reply always lands first; a
+  wedged send can therefore delay it, and `/api/halt` stays the un-queued escape
+  hatch. `200 {"ok":true}` when the ESC landed, `404 {"error":"not found"}` for an
+  unknown id, a dead pane, or an undeliverable send.
 - `POST /api/broadcast` — send the same text to many sessions at once.
 - `DELETE /api/sessions/:id` — archive (end) the session.
 - `GET /api/sessions` — list active sessions; `GET /api/sessions/:id/diff`,
--- a/src/server.ts
+++ b/src/server.ts
@@ -2571,6 +2571,19 @@ async function handleSessionReply({ req, parts, deps }: Ctx): Promise<Response |
   return ok ? json({ ok: true }) : json({ error: "not found" }, 404);
 }
 
+// POST /api/sessions/:id/interrupt — interrupt ONE running session: a lone ESC to its pane and
+// nothing else (#1993). The per-session counterpart to the fleet-wide /api/halt, and composable
+// with /reply, so cancel-then-re-prompt is interrupt → reply rather than a combined endpoint.
+async function handleSessionInterrupt({ req, parts, deps }: Ctx): Promise<Response | null> {
+  if (!(req.method === "POST" && parts[2] && parts[3] === "interrupt")) return null;
+  // No body — the path id IS the target, so there is nothing to validate (mirrors handleHalt).
+  // interrupt() is non-throwing: unknown id, dead/unlisted pane, unreachable herdr and a rejected
+  // send all resolve false → 404 here, the same shape handleSessionReply uses.
+  return (await deps.service.interrupt(parts[2]))
+    ? json({ ok: true })
+    : json({ error: "not found" }, 404);
+}
+
 // POST /api/sessions/:id/recommend-prompt — analyze the session's recent terminal history via a
 // transient second agent (claude opus / codex gpt-5.5) and return a recommended next prompt.
 // 200 {prompt} on success; 422 {error} on an analysis failure; 503 when unwired.
@@ -3759,6 +3772,7 @@ async function handleSessions(ctx: Ctx): Promise<Response | null> {
     handleSessionWorktree,
     handleSessionDelete,
     handleSessionReply,
+    handleSessionInterrupt,
     handleSessionRecommend,
     handleSessionGo,
     handleSessionAnswerPlanQuestions,
--- a/src/service.ts
+++ b/src/service.ts
@@ -4491,6 +4491,62 @@ export class SessionService {
     return result;
   }
 
+  /**
+   * Interrupt ONE session (`POST /api/sessions/:id/interrupt`): send a lone ESC — the agent's
+   * interrupt key — to that session's pane and nothing else (no bracketed paste, no trailing CR).
+   * Composable with reply, so a caller wanting cancel-then-re-prompt does interrupt → reply rather
+   * than needing a combined endpoint (#1993). Same non-throwing boolean contract as reply(): false
+   * for an unknown id, a dead/unlisted pane, an unreachable herdr, or a send that rejects.
+   *
+   * Three things it deliberately does NOT copy from `haltAll`:
+   *
+   *  1. It goes THROUGH #serializeSteer. haltAll bypasses the FIFO because an e-stop must never
+   *     queue behind the very steers it exists to cancel, and it accepts the resulting race (an ESC
+   *     landing after a paste block has closed clears the composed input while that steer's trailing
+   *     CR submits the remnant). A targeted interrupt is a steer, not an e-stop, so that reasoning
+   *     does not transfer: serializing makes interrupt → reply ordering deterministic — the ESC can
+   *     never land inside or after a reply's paste block. The accepted cost is the mirror image of
+   *     haltAll's: an interrupt can queue behind a wedged send, since the serializer chain is
+   *     process-wide rather than per-pane. /api/halt stays the un-queued escape hatch.
+   *  2. It does NOT filter on `agentStatus === "working"`. haltAll filters because it is a sweep
+   *     that must not touch bystanders; a caller naming one session has already chosen its target.
+   *     A status check would also cost a second herdr round-trip and still race the send. ESC on an
+   *     idle pane merely clears any composed input.
+   *  3. It resolves the pane the way the REPLY route does, not the way haltAll does: `liveAgentFor`
+   *     + `adoptLiveResumeAgent`, so a herdr-restored pane (live terminalId ≠ stored herdrAgentId)
+   *     is interrupted at its live id and the stored id is healed — interrupt and a following reply
+   *     therefore address the same pane. operatorReply's isolated-Codex RESUME is deliberately not
+   *     inherited: resuming a dead pane in order to interrupt it is meaningless.
+   *
+   * No `reply` signal is recorded (that kind carries operator words and is mined by the learnings
+   * distiller; an interrupt has no payload to mine) and no event is emitted (`halt:done` exists
+   * because a fleet sweep's reach is otherwise invisible; a single-target interrupt's outcome is
+   * its HTTP status).
+   */
+  async interrupt(id: string): Promise<boolean> {
+    const s = this.deps.store.get(id);
+    if (!s) return false;
+    let agent: HerdrAgent | null;
+    try {
+      // liveAgentFor lets a herdr-unreachable error propagate; an interrupt that can't reach herdr
+      // is a "didn't land", not a 500 (matches operatorReplyTarget's catch).
+      agent = this.liveAgentFor(id);
+    } catch {
+      return false;
+    }
+    if (!agent) return false; // dead / unlisted pane — the ESC would go nowhere
+    const target = agent.terminalId;
+    // Heal a herdr-restored pane's stale stored id, exactly as operatorReply does — a no-op when
+    // the ids already match. Null means the row vanished mid-call: don't send.
+    if (!this.adoptLiveResumeAgent(s, agent)) return false;
+    try {
+      await this.#serializeSteer(() => this.deps.herdr.send(target, "\x1b"));
+    } catch {
+      return false; // pane died between list and send
+    }
+    return true;
+  }
+
   /** Terminal ids herdr currently lists as live. Empty when herdr can't be reached, so
    *  callers treat an unlisted agent as a dead pane (the steer won't land). */
   private liveTerminalIds(): Set<string> {
--- a/test/server-steers.test.ts
+++ b/test/server-steers.test.ts
@@ -109,3 +109,51 @@ test("POST /api/halt surfaces a herdr-unreachable failure as a 500 (not a fake s
   expect(res.status).toBe(500);
   expect((await res.json()).error).toBe("herdr down");
 });
+
+// ── POST /api/sessions/:id/interrupt — the per-session counterpart to /api/halt (#1993) ──
+
+/** makeApp over a stub service exposing only interrupt(), which records the ids it saw. */
+function interruptHarness(result: boolean) {
+  const seen: string[] = [];
+  const store = new SessionStore(":memory:");
+  const deps: AppDeps = {
+    store,
+    events: new EventHub(),
+    service: {
+      interrupt: async (id: string) => {
+        seen.push(id);
+        return result;
+      },
+    } as any,
+    usageLimits: { limits: () => ({}) } as any,
+  };
+  return { app: makeApp(deps), seen };
+}
+
+test("POST /api/sessions/:id/interrupt interrupts that one session", async () => {
+  const { app, seen } = interruptHarness(true);
+  // No body and no content-type: the path id IS the target, so there is nothing to validate.
+  const res = await app.fetch(
+    new Request("http://x/api/sessions/s1/interrupt", { method: "POST" }),
+  );
+  expect(res.status).toBe(200);
+  expect(await res.json()).toEqual({ ok: true });
+  expect(seen).toEqual(["s1"]);
+});
+
+test("POST /api/sessions/:id/interrupt 404s when it didn't land (unknown id / dead pane)", async () => {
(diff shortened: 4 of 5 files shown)


scoop (commented):
## What I checked

The diff is 4 files (`src/server.ts`, `src/service.ts`, plus two test files) and adds exactly one route and one service method. I verified every dependency it relies on.

### `SessionService.interrupt()` (src/service.ts:4526-4548)

- **Non-throwing boolean contract holds on every path.** `store.get` → `false` (`src/service.ts:4527-4528`); `liveAgentFor` wrapped in try/catch (`4530-4536`) — and that wrapper is load-bearing, because `liveAgentFor` (`src/service.ts:2725-2730`) calls `this.deps.herdr.list()` with no guard of its own, the same raw call `haltAll` deliberately lets throw at `src/service.ts:4459`, and the same one `operatorReplyTarget` catches at `src/service.ts:4205-4208`. `null` agent → `false` (`4537`), send rejection → `false` (`4542-4546`).
- **Serializer choice is sound.** `#serializeSteer` is `createSerializer()` (`src/service.ts:1843` → `src/herdr.ts:562-572`): `tail.then(fn, fn)` with the tail swallowed both ways, so a prior rejected send cannot wedge or poison the chain, and a sync throw from `herdr.send` still surfaces as a rejection the `catch` at `4544` handles. `sendSteerTo` puts PASTE+CR inside **one** serializer task (`src/service.ts:4604-4607`), so a serialized ESC can never land between a reply's paste-end and its CR — which is precisely the `haltAll` race documented at `src/service.ts:4466-4477`. The comment at `4503-4511` is factually accurate about both sides of that trade.
- **Ordering test is meaningful, not accidental.** `reply()` → `replyToLive` (`src/service.ts:4198`, `4563-4580`) runs synchronously up to `await sendSteerTo`, and `sendSteerTo` runs synchronously up to `await this.#serializeSteer(...)`. `interrupt()` likewise has no `await` before its own enqueue (`4527-4543`). So call order == enqueue order == FIFO order, and `test/service.test.ts:8180-8190`'s `["\x1b[200~hello\x1b[201~", "\r", "\x1b"]` assertion is a real invariant rather than a timing coincidence.
- **Pane resolution matches the reply route**, as the issue's "minor" note asked. `target` is captured from `agent.terminalId` (`4538`) and `adoptLiveResumeAgent` (`src/service.ts:2732-2737`) re-points the stored `herdrAgentId` at that same id, mirroring `operatorReply` at `src/service.ts:4256-4258`. Verified the return type is `Session | null`, so `if (!this.adoptLiveResumeAgent(...))` at `4541` is a correct null check.
- **The one guard `operatorReply` has that `interrupt` omits is correctly omitted.** `operatorReply` re-reads the row and compares `current.herdrAgentId !== s.herdrAgentId` (`src/service.ts:4253-4254`) because it `await`s `operatorReplyTarget` (which itself awaits `paneForegroundProcs`, `src/service.ts:4216`) before adopting — the row can move under it. `interrupt` has no await before the adopt, so there is no window to guard.
- **Archived rows can't be interrupted:** `liveAgentFor` matches over `store.list({ activeOnly: true })` (`src/service.ts:2726-2728`), so an archived row never resolves an agent.

### Route (src/server.ts:2574-2585, registered at :3775)

- **No shadowing by an earlier handler.** Walked every entry ahead of it in `handleSessions` (`src/server.ts:3766-3798`): `handleSessionsClearMerged` gates on `parts[2] === "clear-merged"` (`:2521`), `handleSessionCreate` on `!parts[2]` (`:2181`), `handleSessionReads`/`handleSessionScratchpad`/`handleSessionWorktree` on `GET` (`:2328`, `:2426`, `:2459`), `handleSessionScratchpadUpload` on `parts[3] === "scratchpad"` (`:2478`), `handleSessionDelete` on `DELETE` (`:2545`). Also checked the handlers registered *before* `handleSessions` in `ROUTE_HANDLERS` that share the `sessions` prefix — `handleSessionHooks`, `handleBuildQueue` (`parts[3]==="queue"`), `handleEpicDraft` (`parts[3]==="epic-draft"`, `:2748`) — none claim `interrupt`. The route is reachable.
- **`GET …/interrupt` really does 404.** `handleSessionReads` falls through because `parts[3]` is set and matches none of its cases, hitting `return null` at `src/server.ts:2342`, so the request reaches the generic `/api` 404 at `src/server.ts:7619`. Same posture as `handleSessionReply`, which also has no explicit 405.
- **Auth/CSRF unchanged and adequate.** POST is not public (`isPublicRequest`, `src/server.ts:678-690`), so `checkAuth` (`:706-719`) gates it; `checkOrigin` (`:757-773`) runs for all POSTs regardless of content type, so skipping `requireJsonContentType` does not open a CSRF hole — and this is exactly `handleHalt`'s existing posture for a strictly *more* powerful fleet-wide action.
- **No privilege expansion to spawned agents.** `isAgentIngressRoute` (`src/server.ts:7650-7666`) allowlists only `hooks`, `queue`, `queue/steps/*` and `epic-draft`; `interrupt` is not in it, so the auth-exempt loopback ingress 404s it (`src/server.ts:7681-7683`). Correct — an agent should not be able to ESC a sibling session.
- **Never a 500 in practice:** the only unguarded call in `interrupt` is `store.update` inside `adoptLiveResumeAgent`, and a throw there would be converted to a JSON 500 by the wrapper at `src/server.ts:7639` rather than crashing. `operatorReply` carries the identical exposure on the same call, so this is not a regression.

### Task satisfaction

The issue asked for `POST /api/sessions/:id/interrupt` sending `"\x1b"` and nothing else, composable with `/reply`; it asked for the serializer question to be decided deliberately rather than copied from `haltAll`; and it flagged the `terminalId` vs `herdrAgentId` mismatch. All three are addressed, with the reasoning captured in-code at `src/service.ts:4501-4522`. `test/service.test.ts:8121-8131` pins the exact payload (`[{target:"term_a", text:"\x1b"}]`, no paste markers, no CR, no event), which is the right guard against a later refactor routing it through `sendSteerTo`. Verified `store.listSignals` exists with the asserted signature (`src/store.ts:3712`) and that neither `interruptHarness` nor `liveAgent` collides with an existing declaration in `test/service.test.ts` (only occurrences: `:8072`, `:8114`).

## Nits (non-blocking):

- src/server.ts: `docs/external-task-api.md:125-128` has a "Steering and ending a task" list enumerating `/reply`, `/broadcast` and `DELETE /api/sessions/:id` for exactly the kind of external caller (shepherd-buzz-bridge) the issue cites as blocked; adding the new `/interrupt` sibling there would keep that catalog current.
- src/service.ts: a caller doing the issue's cancel-then-re-prompt against a **dead isolated Codex pane** gets `404` from `/interrupt` (no resume path, by design per `src/service.ts:4517-4521`) and then `200` from `/reply`, which resumes it (`src/service.ts:4262-4266`). That asymmetry is correct but non-obvious for a bridge author; a sentence in the endpoint comment saying a 404 from interrupt does not mean a following reply will fail would save someone a debugging session.
- test/server-steers.test.ts: the `GET …/interrupt` case uses an id (`s1`) that does not exist in the `:memory:` store, so it would 404 even if the method guard were wrong; seeding a real session would make it actually pin the POST-only behaviour.

scoop (commented):
## What I verified

**Service (`src/service.ts:4526-4548`).**
- `store.get` → `liveAgentFor` → `adoptLiveResumeAgent` → serialized `send(target, "\x1b")`. Verified `liveAgentFor` (`src/service.ts:2725-2730`) matches over `store.list({activeOnly:true})`, so an archived row can never resolve a pane (→ `false` → 404), and `adoptLiveResumeAgent` (`src/service.ts:2732-2737`) is a no-op when the ids already match. This is exactly how the reply route resolves its pane (`src/service.ts:4211`, `4256-4257`), which is the issue's "minor" `terminalId` vs `herdrAgentId` point.
- `target` is captured from the LIVE agent BEFORE the adopt, so the send is unaffected by whatever `store.update` writes — correct.
- `never 500`: the only two throwing seams are wrapped. `liveAgentFor` propagates a herdr-unreachable throw (it calls `this.deps.herdr.list()` directly at `src/service.ts:2727`, unlike `liveTerminalIds` at `src/service.ts:4552-4558` which swallows), and the send is awaited inside a try. `store.get`/`matchAgents` (`src/herdr.ts:314-346`) are pure/sync. The `let agent: HerdrAgent | null` + try/catch-with-return pattern is fine for TS definite-assignment analysis (the only path reaching the use is the completed try).
- Serializer choice checked against the real implementation: `createSerializer` (`src/herdr.ts:562-572`) chains on `tail` with both handlers, and its internal tail swallows rejections, so a rejected ESC cannot poison the queue for later steers. FIFO ordering is real because the serializer slot is claimed in the SYNCHRONOUS prefix of both callers — `reply()` → `replyToLive` → `sendSteerTo` reaches `#serializeSteer` (`src/service.ts:4604`) with no preceding await, and `interrupt()` reaches `#serializeSteer` (`src/service.ts:4543`) with only sync work before it. So the `paste, CR, ESC` assertion in `test/service.test.ts:8203-8228` is deterministic, not incidental.
- Contrast with `haltAll` (`src/service.ts:4460-4492`) is accurate as documented: haltAll bypasses the FIFO, filters `agentStatus === "working"`, sends to `agent.terminalId` without adopting, emits `halt:done`, and lets a list failure propagate. The new method's comment describes what the code actually does in all three respects.
- No new `SignalKind`; `sendSteerTo`'s `addSignal({kind:"reply"})` (`src/service.ts:4593-4598`) is deliberately not on this path, matching the plan.

**Route (`src/server.ts:2574-2585`, registered at `src/server.ts:3775`).**
- Shape is byte-for-byte the sibling of `handleSessionReply` (`src/server.ts:2557-2571`) minus the body/content-type validation, which is right for a bodyless endpoint.
- `GET .../interrupt` really does fall through to the generic 404: `handleSessionReads` returns null for an unmatched `parts[3]` (`src/server.ts:2327-2343`), no earlier entry in `ROUTE_HANDLERS` (`src/server.ts:7488-7539`) claims `parts[1]==="sessions" && parts[3]==="interrupt"`, and the fallback is `src/server.ts:7619`. The route test at `test/server-steers.test.ts:155-160` is therefore asserting the real behaviour.
- Security posture checked, not assumed: `checkOrigin` (`src/server.ts:757-773` → `classifyOrigin`, `src/validate.ts:923-952`) runs for every POST before dispatch, so the absence of a JSON content-type requirement does not open a cross-site hole (a browser cross-site POST always carries `Origin`; a header-less curl client is the intentional CLI allowance, identical to `/api/halt`). The route is also NOT added to `isAgentIngressRoute` (`src/server.ts:7650-7666`), so the auth-exempt loopback ingress cannot reach it — correct, an interrupt is an operator action.

**Docs (`docs/external-task-api.md:129-137`).** Placed in the "Steering and ending a task" list next to `/reply`, `/broadcast` and `DELETE`, none of which are ingress routes either, so the framing is consistent. Status codes described match `src/server.ts:2582-2584`.

**Gates.** The `feat(...)` commit carries `[no-feature-entry]` in its body (33f1151c6), which `scripts/check-feature-catalog.sh:66-76` accepts as a range-level opt-out; no `ui/` files are touched, and no user-facing strings are added, so `check:i18n` parity is not implicated (nothing in `ui/messages/*` to mirror).

### Stated limitation
I could not verify herdr's own `send` semantics for a lone `\x1b`, nor how Codex's TUI reacts to it — neither is in this repo. This PR ships the same assumption `haltAll` already ships (`src/service.ts:4483`), so it introduces no new external dependency.

## Nits (non-blocking):
- `src/server.ts:2578`: no `!parts[4]` guard, so `POST /api/sessions/:id/interrupt/anything` also interrupts. This matches `handleSessionReply` (`src/server.ts:2558`), but several siblings do bound the path (`src/server.ts:3753-3756`, `7653`).
- `src/service.ts:4537`: a pane that herdr still lists but whose agent has exited (a shell-only pane) resolves `true` → 200, while `operatorReply` explicitly refuses that same pane via `paneForegroundProcs` (`src/service.ts:4215-4223`) → 404. Harmless for a lone ESC, but a caller doing interrupt → reply (the buzz bridge's pattern) gets 200 then 404 for the same pane, and `docs/external-task-api.md:136-137` describes 404 as covering "a dead pane".
- `src/service.ts:4527-4541`: unlike `operatorReply` (`src/service.ts:4254-4255`) there is no re-read staleness check before the adopt. It is safe today precisely because nothing awaits between `store.get` and `adoptLiveResumeAgent` — worth saying so in the comment, since adding any await there would silently reintroduce the row-changed-under-us case the reply path guards.
- `docs/external-task-api.md:133-135`: "an interrupt issued before a reply always lands first" holds for the serializer FIFO, but for two concurrently in-flight HTTP requests the ordering is whichever handler reaches `#serializeSteer` first; the guarantee is exact only when the caller awaits the interrupt's 200.

PR #1997, opened 2026-08-03
feat(ui): switch repo mid-compose in two taps, keyboard up

scoop:
Switching repo in the New Task pane on a phone cost **5 taps** and dropped the soft keyboard twice. It now costs **2**, with the keyboard up the whole way.

From the walkthrough that prompted this:

> "I tap the top of the bar here, but that only gets me out of edit mode of the feature description. Then I tap again and now I have access to all of my repos. I pick the new repo, but that once again only brings me halfway… I've actually never used this [branch]. So I tap again, which closes the panel, but now I still can't type or dictate my feature because I need to tap again to go into the description field."

| # | Tap before | Why | After |
| - | - | - | - |
| 1 | header line | while composing the chip was swapped for a **read-only** `.ctx-line` — the tap only blurred | the real entry point |
| 2 | `.ctx-chip` | `openRepoPicker()` | folded into tap 1 |
| 3 | a repo row | `selectRepo()` kept the sheet open, parked focus on `.rs-trigger` | closes the sheet, refocuses the prompt |
| 4 | scrim / Esc | dismiss the still-open sheet | gone |
| 5 | the textarea | re-focus to raise the keyboard again | gone |

## What changed

**The compose-state header line is now two tap targets.** `repo · branch` opens the Repo & Branch sheet; `engine · gate` opens the Engine sheet — each hit area matching what its segment names. Both lifted to the 44px touch floor (`.ctx-chip` too, which sat at ~30px).

**A repo pick ends the flow.** `selectRepo()` closes the sheet and puts the caret back at the end of the prompt.

**Two pointerdown guards, one per tap.** Both the segments and `RepoSelect`'s option row would otherwise blur their field on pointerdown — the row is `tabindex="-1"`, so a tap focused it and blurred `.rs-filter` *before* `click` reached `pick()`. On a phone either one retracts the keyboard and re-inflates the visual viewport mid-gesture.

Everything on the tap path is synchronous — `flushSync`, not `await tick()` — because a soft keyboard only rises for a `focus()` inside the activating gesture's own call stack. Ordering is load-bearing: the flush runs the sheet's `use:dialog` destroy, which restores focus to the opener, so the prompt refocus must come after it.

`RepoSelect.pick()` now settles its own state before calling `onchange`, which may tear the component down synchronously. Behavior-neutral for both consumers.

**Base Branch is unchanged** — same place in the same sheet, reachable by collapsing the repo panel. Desktop is untouched: there is no sheet, so `selectRepo` keeps its existing `queueMicrotask` refocus.

## Verification

- 4505 UI tests + 8084 server tests green; `check`, `check:i18n`, lint, and the branch/feature/announcement/glossary/fallow gates all pass.
- **Mutation-tested** — each new test was confirmed to fail with its fix reverted, individually: drop the row guard → fails; drop the segment guard → fails; restore the old keep-sheet-open behavior → fails. A bare `.click()` dispatches no pointer events, so the two blur tests dispatch a real `PointerEvent` and assert `defaultPrevented`; without that the whole suite is blind to the guards.
- **Driven in a real browser at Pixel 5 width against live data**: both segments 44px, tap 1 opens the sheet with the filter focused, tap 2 closes it with the prompt refocused and the draft text intact, zero header/body overflow with the longest repo name.
- Confirmed via a controlled A/B that `preventDefault` on `pointerdown` does **not** block touch panning (68px vs 70px scrolled), so the option-row guard cannot break scrolling the repo list.

One visual fix came out of the browser pass: with a long repo name the branch was ellipsizing to `m…`, so the branch now holds its ground (capped at 12ch) and the repo name absorbs the shrink.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3517,5 +3517,7 @@
   "keymap_mode_research_short": "RECH.",
   "feat_newtask_keymap_title": "⌘ halten zeigt alle Tastenkürzel",
   "feat_newtask_keymap_body": "Halte in „Neue Aufgabe\" kurz ⌘ (Strg unter Windows/Linux) — jedes Bedienelement zeigt sein Kürzel direkt an seinem Platz, Loslassen blendet wieder aus. ? öffnet die vollständige Tastenkarte.",
-  "diagnostics_hint_tailscale_serve_denied": "Tailscale verweigert Shepherd die serve-Konfiguration, dadurch lässt sich keine Live-Vorschau veröffentlichen — führe einmalig sudo tailscale set --operator=$USER aus und öffne die Vorschau erneut."
+  "diagnostics_hint_tailscale_serve_denied": "Tailscale verweigert Shepherd die serve-Konfiguration, dadurch lässt sich keine Live-Vorschau veröffentlichen — führe einmalig sudo tailscale set --operator=$USER aus und öffne die Vorschau erneut.",
+  "feat_fast_repo_switch_title": "Repo wechseln, ohne die Tastatur zu verlassen",
+  "feat_fast_repo_switch_body": "Tippe auf dem Handy beim Schreiben auf den Repo-Namen in der Kopfzeile von „Neue Aufgabe“ — die Repo-Liste öffnet sich sofort, und nach der Auswahl landest du direkt wieder in deiner Beschreibung, die Tastatur bleibt oben. Zwei Tipps statt fünf."
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -3517,5 +3517,7 @@
   "keymap_mode_research_short": "RSCH.",
   "feat_newtask_keymap_title": "Hold ⌘ to see every shortcut",
   "feat_newtask_keymap_body": "In New Task, hold ⌘ (Ctrl on Windows/Linux) for a moment and every control shows its shortcut right where it sits — release to dismiss. Press ? for the full key card.",
-  "diagnostics_hint_tailscale_serve_denied": "Tailscale refuses Shepherd's serve config, so no live preview can be published — run sudo tailscale set --operator=$USER once, then reopen the preview."
+  "diagnostics_hint_tailscale_serve_denied": "Tailscale refuses Shepherd's serve config, so no live preview can be published — run sudo tailscale set --operator=$USER once, then reopen the preview.",
+  "feat_fast_repo_switch_title": "Switch repo without leaving the keyboard",
+  "feat_fast_repo_switch_body": "On a phone, tap the repo name in the New Task header while you're typing — the repo list opens straight away, and picking one drops you right back into your description with the keyboard still up. Two taps instead of five."
 }
--- a/ui/src/lib/components/NewTask.browser.test.ts
+++ b/ui/src/lib/components/NewTask.browser.test.ts
@@ -3279,8 +3279,20 @@ describe("NewTask mobile sheets + shortcuts", () => {
     await expect.poll(() => document.querySelector(".ctx-sheet")).toBeTruthy();
     expect(chip().getAttribute("aria-expanded")).toBe("true");
 
-    // Change only the repo: chip + payload repoPath update; base re-derives. The chip
-    // tap already opened the repo panel — clicking the trigger here would close it.
+    // Branch FIRST: a repo pick ends the sheet's life, so the independent branch edit
+    // has to happen while the sheet is still up. (Input fallback or select — the mock
+    // lists main only.) The chip tap already opened the repo panel, so the branch
+    // control is behind it; the trigger toggle collapses the panel to reach it.
+    document.querySelector<HTMLButtonElement>(".ctx-sheet .rs-trigger")!.click();
+    await expect.poll(() => document.querySelector(".rs-panel")).toBeNull();
+    const branchCtl = document.querySelector<HTMLSelectElement | HTMLInputElement>(
+      ".ctx-sheet .ctx-branch-select",
+    )!;
+    branchCtl.value = "main";
+    branchCtl.dispatchEvent(new Event("change", { bubbles: true }));
+
+    // Now the repo: chip + payload repoPath update; base re-derives.
+    document.querySelector<HTMLButtonElement>(".ctx-sheet .rs-trigger")!.click();
     await expect
       .poll(() =>
         Array.from(document.querySelectorAll<HTMLLIElement>('[role="option"]')).find((el) =>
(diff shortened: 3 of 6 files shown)


scoop (commented):
## What I verified

**The focus hand-off ordering in `selectRepo` is correct.** `ui/src/lib/components/NewTask.svelte:1119-1122` sets `activeSheet = null`, `flushSync()`, then `promptInput.focus()`. The claim that the flush must come first is real: `use:dialog`'s teardown restores focus **synchronously** inside `destroy()` (`ui/src/lib/a11yDialog.ts:90-97` — `previouslyFocused.focus()` with no microtask/rAF wrapper), so it runs during the flush and the subsequent `focus()` wins. Reversing the two would hand focus back to the opener. The mount-side focus pull *is* deferred (`ui/src/lib/a11yDialog.ts:48-52`), but that path is not on this sequence.

**The `pick()` reorder is safe for both consumers.** `ui/src/lib/components/RepoSelect.svelte:157-162` now writes `open`/`pickerFor`/`filter` before `onchange(path)`, which is required because `selectRepo` can destroy the component inside that call. I checked every `onchange` consumer: `ui/src/lib/components/NewTask.svelte:1660` (desktop, `queueMicrotask` refocus) and `ui/src/lib/components/EpicDiagnoseEntry.svelte:103` (`(p) => (repoPath = p)`) — neither reads `open`, `filter` or `pickerFor`, so the reorder is behaviour-neutral there. Both other `pick()` call sites also finish their own work first: `ui/src/lib/components/RepoSelect.svelte:192-196` (`e.preventDefault()` before `pick`) and the row's `onkeydown` at `:309-311` (nothing after `pick`). No use-after-destroy remains.

**No stray modal close from unmounting the sheet mid-click.** `.overlay`'s handler guards on `e.target === e.currentTarget` (`ui/src/lib/components/NewTask.svelte:1509-1514`) and `.nt-backdrop` (`:1496-1503`, unguarded) is a **sibling** of `.overlay`, so it is never in the clicked row's propagation path. `MobileEngineSheet`'s scrim is likewise a sibling of `.sheet` and guards on target identity (`ui/src/lib/components/new-task/MobileEngineSheet.svelte:29-31`). `RepoSelect`'s window outside-click listener runs in the **capture** phase (`ui/src/lib/components/RepoSelect.svelte:230`), i.e. before the row handler detaches anything.

**Desktop path is genuinely unchanged.** The old code evaluated `activeSheet === "context"` in a microtask; the new code evaluates it synchronously, but `activeSheet` is forced `null` whenever `!mobile` (`ui/src/lib/components/NewTask.svelte:380-382`) and the desktop `RepoSelect` only renders under `{#if !mobile}` (`:1651`), so the desktop branch collapses to the identical `queueMicrotask(() => promptInput?.focus())`.

**Comments in the diff are factually accurate.** The `.compose-actions` precedent the new comment cites really does use `onpointerdown` + `preventDefault` (`ui/src/lib/components/NewTask.svelte:2066-2070`, `:2091-2095`). `RepoSelect`'s comment claiming the EmojiPicker focuses its own search is true (`ui/src/lib/components/EmojiPicker.svelte:29-31`), and the picker itself renders **outside** the guarded `<li>` (`ui/src/lib/components/RepoSelect.svelte:403-411`), so its inputs still take focus normally. The `SlashCommandMenu`/`IssueSearchMenu` contrast is accurate too.

**Identifiers and catalogs.** Every message key the new markup calls exists in both locales: `newtask_context_chip_aria` (en.json:3321 / de.json:3321, both with `{repo}`+`{branch}`), `newtask_gate_on`/`newtask_gate_off` (en.json:3326-3327 / de.json:3326-3327), `agent_provider_codex` (en.json:2228 / de.json:2228), `reposelect_placeholder` (en.json:343 / de.json:343). The two new announcement keys are present in both catalogs (en.json:3521-3522, de.json:3521-3522) and are the only keys the entry references (`ui/src/lib/feature-announcements/entries/v1.47.0-fast-repo-switch.ts:6-7`). `sinceVersion: "1.47.0"` is > the last release `1.46.0` (`.release-please-manifest.json:2`) and matches the filename prefix, satisfying `check-announcement-versions.mjs`; the `id` `fast-repo-switch` is unique across `ui/src/lib/feature-announcements/entries/`, and the entry shape matches `FeatureAnnouncement` (`ui/src/lib/feature-announcements.ts:26-35`) and the sibling entry `v1.47.0-newtask-keymap.ts`. `RepoEntry`, used by the new test helper signature, is already imported (`ui/src/lib/components/NewTask.browser.test.ts:6`).

**CSS gates align.** The new `.ctx-seg` rules sit inside the mobile media block that spans `ui/src/lib/components/NewTask.svelte:3151-3596`, whose condition `(max-width: 768px), (max-height: 480px) and (pointer: coarse)` mirrors the JS `mobile` derivation at `:373-375` exactly, so the styles can never render without their markup. Both tap targets get `min-height: 44px` (`:3403`, `.ctx-chip` at `:3192`) against the now-zeroed `.chead` vertical padding (`:3183`), and only design tokens are used for colour/type (`var(--color-faint)`, `var(--fs-meta)`, `var(--fs-micro)`), per CLAUDE.md:40. `.ctx-line b` (`:3430-3436`) and `.ctx-line .ctx-dim` (`:3437-3444`) still apply through the new button wrappers, so the ellipsis behaviour the shrink strategy depends on is intact.

**No orphaned CSS or dead selectors** were left behind by the `.ctx-line` rewrite — `.ctx-line b`, `.ctx-line .ctx-dim` and `.ctx-line .es-gate` all still have matching elements inside the two segments.

## Behavioural note (non-blocking)

The plan's "desktop is untouched" holds only for *wide* windows. `mobile` is width-gated (`ui/src/lib/components/NewTask.svelte:373-375`), so a mouse user in a ≤768px-wide desktop window also gets the mobile chrome and therefore the new "a repo pick ends the sheet" contract — losing the previous ability to set repo *and* branch in one sheet visit. That is consistent with the intended mobile flow (same UI, same 2-tap win, branch still one tap away by reopening the chip), so I'm not raising it as a defect — just flagging that the invariant is layout-gated, not pointer-gated.

## Stated limitations

- I could not verify on a real iOS/Android device that `preventDefault()` on `pointerdown` keeps the soft keyboard up **and** still delivers `click` — that ground truth is not in this repo. Per the Pointer Events spec, cancelling `pointerdown` suppresses only the compatibility mouse events (and thus the focus/blur), not `click`, and the plan records a real-browser pass at 390px, so I treated it as written. Worth noting that the existing precedent hedges differently: `.compose-actions` duplicates its action in *both* `onpointerdown` and `onclick` (`ui/src/lib/components/NewTask.svelte:2066-2070`), whereas the new segments and the `RepoSelect` row rely on `click` alone. If `click` ever fails to fire on some engine, the segment tap becomes a no-op rather than degrading.
- Likewise I could not empirically confirm that the guard does not block touch panning of `.rs-list` (`ui/src/lib/components/RepoSelect.svelte:520-527`); by spec, panning is governed by `touch-action`, not by cancelling `pointerdown`.

## Nits (non-blocking)

- ui/messages/de.json: "Zwei Tipps statt fünf." reads as "two *tips*", not "two taps" — the catalog's own precedent for a tap is "Antippen" (de.json:2455). Consider "Zwei Antippen statt fünf" or "Statt fünfmal tippen nur noch zweimal".
- ui/src/lib/components/NewTask.svelte: `setSelectionRange(prompt.length, prompt.length)` (:1122) discards a mid-text caret that the pointerdown guard had actually preserved; capturing `promptInput.selectionStart` in `openRepoPicker` and restoring it would keep a mid-sentence editor in place (it does match the existing house style at :923-924, so this is a preference).
- ui/src/lib/components/NewTask.svelte: `aria-expanded` on `.ctx-seg-repo` (:1562) can never be observed as `true` — the segment unmounts as soon as the sheet takes focus and `composing` flips false — so it is inert state rather than a live signal.
- ui/src/lib/components/NewTask.browser.test.ts: in the guard test (:4661-4674), `expect(document.activeElement).toBe(filter)` after a *synthetic* `PointerEvent` proves nothing (a synthetic event has no default focus action to prevent); only the `defaultPrevented` half of that assertion carries the contract for the row.

PR #2006, opened 2026-08-03
fix(tailscale): latch a serve-config denial seen by unregister

### LINKED ISSUE #1998: fix(tailscale): unregister doesn't latch a serve-config denial (permissionDenied can read stale false)
Follow-up from the review of #1981 (merged as 36ffdb6). Not a regression that PR introduced — a gap in the coverage it added.

## What

`TailscaleServeService` latches a permission refusal (`isServeConfigDenied`) in two of its three mutation paths:

- `register()` — latches on failure, and clears on success
- `reconcileStartup()` — latches on failure
- `unregister()` — **does neither**; its catch only `console.warn`s

So a denial observed *only* by an unregister is dropped, and `permissionDenied()` keeps reporting `false`. The `tailscale` diagnostics row then stays green while serve-config writes are in fact being refused — the exact green-over-broken reading #1981 exists to close, just reached by a different path.

## When it bites

Narrow, which is why it wasn't fixed inline: rights have to be revoked *mid-run*, after at least one successful register. Concretely — Shepherd boots with `--operator` set (reconcile + registers succeed, latch clear), an operator later runs `tailscale set --operator=` to hand the tailnet back to root, and the next preview teardown is refused. Until some later `register` is attempted and fails, Diagnostics reports `ok`.

Nothing is corrupted; the report is just stale in the optimistic direction.

## Fix sketch

One line, symmetric with the other two paths:

```ts
} catch (err) {
  if (isServeConfigDenied(err)) this.denied = true;
  console.warn(`[tailscale-serve] unregister failed for ${id} port ${entry.port}:`, err);
}
```

Worth deciding deliberately rather than pattern-matching, since it's a behavior change:

- **Should a successful `unregister` also *clear* the latch?** Today only a successful `register` does, and that asymmetry is deliberate — the hint tells the operator to *reopen the preview*, so the row goes green exactly when the fix is proven on the path that was broken. A successful unregister proves the same right, but greening the row off a teardown is a weaker signal. Recommendation: latch on failure, do **not** clear on success, and keep the field's doc comment (in `src/tailscale.ts`) in step with whatever is chosen.
- `stopAll()` (the sync shutdown path) has the same gap, but latching there is pointless — the process is exiting and nothing will read the flag.

## Tests

`test/tailscale.test.ts` already has the fixtures — `denyingRun` and `deniedError()`. Add: a denied `unregister` after a successful `register` latches; an unrelated unregister failure does not.

scoop:
Closes #1998.

## What

`TailscaleServeService` latched a permission refusal (`isServeConfigDenied`) in two of its three async mutation paths — `register()` (latch + clear on success) and `reconcileStartup()` (latch) — but `unregister()` did neither; its catch only `console.warn`ed.

So a denial observed **only** by a teardown was dropped and `permissionDenied()` kept reporting `false`. The `tailscale` diagnostics row stayed green while serve-config writes were in fact being refused — the green-over-broken reading #1981 exists to close, reached by a different path.

Not a regression from #1981 (36ffdb6) — a gap in the coverage it added.

## When it bites

Rights have to be revoked *mid-run*, after at least one successful register: Shepherd boots with `--operator` set (reconcile + registers succeed, latch clear), an operator later runs `tailscale set --operator=` to hand the tailnet back to root, and the next preview teardown is refused. Until some later `register` is attempted and fails, Diagnostics reports `ok`. Nothing is corrupted; the report is stale in the optimistic direction.

## Decisions

- **A successful `unregister` does NOT clear the latch.** Latch-on-failure only. The asymmetry is deliberate and preserved: the operator hint says "run \`tailscale set --operator=…\`, then reopen the preview", and reopening a preview is a `register` — so the row greens exactly when the fix is proven on the path that was broken. Greening off a teardown is a weaker signal.
- **`stopAll()` (sync shutdown) is left alone.** Same structural gap, but the process is exiting and nothing reads the flag afterwards — latching there would be dead code.
- The `denied` field's doc comment now names all three latching paths and why `stopAll` is excluded.

The slot-release path is unchanged: `byId.delete(id)` and the `onChange(null, null)` fire still run after a failed `off`.

## Tests

Two added to the existing `permissionDenied` describe, reusing the `deniedError()` fixture:

- a denied `unregister` after a successful `register` latches `true` (and the slot is still released);
- an `unregister` failing for an unrelated reason leaves it `false`.

Both need a runner whose behavior differs between the two calls (`unregister` early-returns unless a `register` seeded the entry), so they discriminate on the `off` argv rather than using the fixed `denyingRun`.

Verified red-before/green-after: reverting just the one-line latch fails the first new test.

## Verification

`bun test ./test` 8131 pass / 0 fail · `bun run lint` clean · `bunx tsc --noEmit` clean · `bunx fallow audit --base origin/main` no issues in the 2 changed files.

[no-feature-entry] — server-internal fix, ships no user-facing UX.

### DIFF
--- a/src/tailscale.ts
+++ b/src/tailscale.ts
@@ -111,8 +111,10 @@ export interface TailscaleServeOpts {
 export class TailscaleServeService {
   private byId = new Map<string, { port: number; state: ServeState }>();
   private queue: Promise<void> = Promise.resolve();
-  /** Latched by a `register`/`reconcileStartup` mutation refused for lack of operator/root
-   *  rights; cleared only by a subsequent SUCCESSFUL `register` — not by a successful
+  /** Latched by ANY async mutation refused for lack of operator/root rights — `register`,
+   *  `unregister` and `reconcileStartup` alike, since all three write serve config (only the
+   *  sync `stopAll` skips it: the process is exiting and nothing reads this afterwards).
+   *  Cleared only by a subsequent SUCCESSFUL `register` — not by a successful
    *  unregister or reconcile, neither of which clears it. That asymmetry is deliberate and
    *  matches the hint the operator is given ("run `tailscale set --operator=…`, then reopen
    *  the preview"): reopening a preview is a register, so the row goes green exactly when
@@ -164,6 +166,11 @@ export class TailscaleServeService {
       try {
         await this.run(["serve", `--https=${entry.port}`, "off"]);
       } catch (err) {
+        // Symmetric with register/reconcileStartup: a teardown is a serve-config WRITE, so a
+        // refusal here is the same host-level verdict. Without this, rights revoked mid-run
+        // (operator hands the tailnet back to root after a successful register) go unnoticed
+        // until some later register fails, and Diagnostics reads green in the meantime.
+        if (isServeConfigDenied(err)) this.denied = true;
         console.warn(`[tailscale-serve] unregister failed for ${id} port ${entry.port}:`, err);
       }
       this.byId.delete(id);
--- a/test/tailscale.test.ts
+++ b/test/tailscale.test.ts
@@ -452,6 +452,43 @@ describe("TailscaleServeService.permissionDenied", () => {
     expect(svc.permissionDenied()).toBe(false);
   });
 
+  // The blind spot this closes: rights are revoked MID-RUN (`tailscale set --operator=`)
+  // after a successful register, so the first refusal Shepherd sees is a teardown. That
+  // denial used to be dropped on the floor, leaving Diagnostics green over a host whose
+  // serve-config writes are being refused.
+  test("latches true when an unregister is refused for lack of operator rights", async () => {
+    // Rights hold for the register and are gone by the `off` — a fixed `denyingRun` can't
+    // express this, since unregister early-returns unless a register seeded the entry.
+    const run: TailscaleRunner = async (args) => {
+      if (args.includes("off")) throw deniedError();
+      return { stdout: "" };
+    };
+    const svc = new TailscaleServeService({ base: 8001, count: 5, enabled: true, run });
+
+    await svc.register("s1", 8001);
+    expect(svc.permissionDenied()).toBe(false);
+
+    await svc.unregister("s1");
+
+    expect(svc.permissionDenied()).toBe(true);
+    // The slot is still released: this is a host-level verdict, not slot bookkeeping.
+    expect(svc.snapshot()).toEqual({});
+  });
+
+  test("stays false when an unregister fails for an unrelated reason", async () => {
+    const run: TailscaleRunner = async (args) => {
+      if (args.includes("off")) throw new Error("fake run failure on port 8001");
+      return { stdout: "" };
+    };
+    const svc = new TailscaleServeService({ base: 8001, count: 5, enabled: true, run });
+
+    await svc.register("s1", 8001);
+    await svc.unregister("s1");
+
+    expect(svc.permissionDenied()).toBe(false);
+    expect(svc.snapshot()).toEqual({});
+  });
+
   test("stays false when the service is disabled (no mutation is ever attempted)", async () => {
     const svc = new TailscaleServeService({
       base: 8001,


scoop (commented):
## What I verified

The diff touches exactly two files (`git diff --name-only`): `src/tailscale.ts` and `test/tailscale.test.ts`.

**The fix does what the task asks.** The added line at `src/tailscale.ts:173` is byte-for-byte the shape used by the two paths that already latched — `register` at `src/tailscale.ts:153` and `reconcileStartup` at `src/tailscale.ts:206`. All three call the same predicate, `isServeConfigDenied` (`src/tailscale.ts:76-81`), which tests `/serve config denied/i` against `stderr` + `message` only, so an unrelated teardown failure cannot latch (re-asserted by the existing over-matching guard at `test/tailscale.test.ts:371-373`).

**The latch is actually reachable on a path that runs today.** `unregister` is driven by the `session:preview` subscription at `src/index.ts:1327` (register when a preview port binds, unregister on teardown), so this is a live path, not a dormant one. The value flows to the diagnostics row via `src/index.ts:2756` → `src/diagnostics.ts:1293` (`tailscaleProbe` returns `warning` + `diagnostics_hint_tailscale_serve_denied` when `previewServeDenied()` is true, and does so *before* the serve-status read). Feeding that consumer a truthful value is the whole fix; no consumer contract changed.

**Post-catch behavior is genuinely unchanged.** `this.byId.delete(id)` and `this.fire(id, null, null)` at `src/tailscale.ts:176-177` still run on the failure path, so the slot is still released — the new test asserts exactly that (`test/tailscale.test.ts:475`), and the pre-existing unregister test at `test/tailscale.test.ts:134-151` still describes the same argv/onChange/snapshot contract.

**Clear-side semantics are untouched.** `this.denied = false` still happens only in `register`'s success branch (`src/tailscale.ts:150`); nothing in the diff adds a clear to `unregister`. That matches the operator hint the UI actually ships — `ui/messages/en.json:3520` ("…then reopen the preview") and its German counterpart `ui/messages/de.json:3520` — so the row goes green on a register, which is what reopening a preview performs.

**Locale parity: nothing to do.** The diff adds no user-facing string and no message key; the only hint key involved is pre-existing and present in both catalogs (`ui/messages/en.json:3520`, `ui/messages/de.json:3520`, plus the doc link at `ui/src/lib/diagnostics-docs.ts:14`).

**The `denied` doc comment is factually accurate after the edit.** It now names all three async latching paths (`src/tailscale.ts:114-116`), and its claim that `stopAll` is skipped because "the process is exiting and nothing reads this afterwards" checks out: `tailscaleServe.stopAll()` is called only from the `exit` handler (`src/index.ts:3093`) and the `SIGTERM` handler (`src/index.ts:3104`), which exits immediately after. The clear-side sentence still matches `src/tailscale.ts:150`.

**The new tests exercise the right thing.** The runners discriminate on `args.includes("off")`: register's argv is `["serve","--bg","--https=<p>","127.0.0.1:<p>"]` (`src/tailscale.ts:148`, pinned at `test/tailscale.test.ts:126`) and contains no `"off"`, while unregister's is `["serve","--https=<p>","off"]` (`src/tailscale.ts:167`, pinned at `test/tailscale.test.ts:148`). So the register really does succeed and only the teardown is refused. Worth calling out: the author deviated from the plan's step 4, which suggested reusing `makeRun(new Set([8001]))` for the negative case — that would have been *worse*, because `makeRun` (`test/tailscale.test.ts:74-87`) rejects by port on every call, so the register would have failed too (`src/tailscale.ts:155` still seeds `byId` with `state: "failed"`, so the test would pass while testing the wrong scenario). The inline stateful runner is the correct call.

No correctness, security, or contract problem found. `deniedError()` (`test/tailscale.test.ts:343-349`) carries the wording in both `.stderr` and `.message`, matching the predicate's input assumptions.

- `src/tailscale.ts:220-231`: `stopAll` remains the one mutation path with no latch — inert today (verified exit-only above), and note that `defaultRunSync` passes `stdio: "ignore"` (`src/tailscale.ts:88`), so an `execFileSync` rejection there would carry no `stderr` for `isServeConfigDenied` to read even if someone later wanted to latch it.

### Nits (non-blocking):
- `src/tailscale.ts:169-172`: the four-line catch comment largely restates the `denied` field doc directly above it (`src/tailscale.ts:114-121`); one line would carry the same information.

- `src/diagnostics.ts:123-125`: the `previewServeDenied` dep doc says the flag means `tailscale serve` "refused a preview registration", which was already narrower than reality once `reconcileStartup` began latching in #1981, and is now narrower still.

### Stated limitations
- Per instructions I did not run tests or lint; CI covers those. All assertions above come from reading the checked-out sources cited by `path:line`.

PR #2007, opened 2026-08-03
feat(ui): per-session stop on the session-card menu (#1995)

### LINKED ISSUE #1995: feat(ui): per-session stop button on top of POST /api/sessions/:id/interrupt

scoop:
Closes #1995.

#1993 shipped `POST /api/sessions/:id/interrupt` (a lone ESC to one pane, 200/404, no body) but nothing in the UI called it. The gap the parent issue named — no per-session stop for a session you are **not** focused on / attached to — is now closed.

## What it does

Right-click (or long-press) a **working** session's card → **Stop agent**. One click sends a single ESC to just that session's pane.

```
┌─ card menu ─────────┐
│ ⇥  Merge PR         │
│ ⎋  Stop agent       │  ← new
│ ↻  Resume session   │
│ ✎  Rename           │
│ ♻  Relaunch         │
│ ✕  Decommission     │
└─────────────────────┘
```

## Decisions (settled with the operator before implementing)

- **CardMenu only.** It is already the per-session action surface, so no new chrome lands on every row, and the missing surface (an unfocused session) is exactly what a card menu addresses. Desktop Esc-in-terminal and the mobile `SteerBar` `ControlBar include={["cancel"]}` already cover the attached pane.
- **It must not read as an e-stop.** The fleet halt (`GearHaltHero`) is an amber ■ hero with a live `N WORKING` chip and a two-step red arm. This is a plain neutral `.cm-item` with an `⎋` (Esc-key) glyph — no wash, no arm, no chip. A test asserts the item carries neither the `armed` nor the `danger` class.
- **One click, no arm.** It is one ESC to one pane and it is reversible — the operator can steer the session again immediately. Arms stay reserved for the irreversible items (Relaunch, Decommission) and for the fleet e-stop.
- **Running only.** Gated on `displayStatus(session, workingBlocked) === "running"`, not raw status, so a working-while-blocked session (mid-turn behind a permission prompt, amber working pip) is stoppable too. Idle / blocked-not-working / done / archived rows have no turn to cut short.
- **Interrupt-only.** The endpoint is deliberately composable with `/reply`, so cancel-then-re-prompt stays interrupt → steer rather than a combined action.

## Failure handling

`interruptSession()` throws on non-ok; `UnitRow` catches. A 404 (the pane died between render and click) is a **normal** outcome, not a crash — both outcomes toast, keyed `session-interrupt:<id>` so repeated taps collapse into one. The success toast matters: a lone ESC to a card you are not watching is otherwise invisible.

## Notes

- Like Merge PR, the endpoint is session-scoped, so no parent callback is threaded — the row calls the API and toasts the outcome itself.
- `onstop` is resolved into a `$derived` rather than an inline ternary in the `<CardMenu>` tag: UnitRow's template sits at its grandfathered fallow complexity cap (#855), and an inline conditional in the markup tipped it over.
- EN + DE keys added (`check:i18n` parity) and one feature-announcement fragment (`v1.47.0-session-stop.ts`, version from `bun run next-version`).

## Verification

- `cd ui && bun run check` → 0 errors, `bun run check:i18n` → parity
- `cd ui && bun run test` → 288 files / 4513 tests pass (6 new: 2 CardMenu, 4 UnitRow)
- pre-push gates green: branch hygiene, feature catalog, announcement versions, glossary, fallow audit
- Rebased onto `origin/main` (7c80d7727) before pushing

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3519,5 +3519,11 @@
   "feat_newtask_keymap_body": "Halte in „Neue Aufgabe\" kurz ⌘ (Strg unter Windows/Linux) — jedes Bedienelement zeigt sein Kürzel direkt an seinem Platz, Loslassen blendet wieder aus. ? öffnet die vollständige Tastenkarte.",
   "diagnostics_hint_tailscale_serve_denied": "Tailscale verweigert Shepherd die serve-Konfiguration, dadurch lässt sich keine Live-Vorschau veröffentlichen — führe einmalig sudo tailscale set --operator=$USER aus und öffne die Vorschau erneut.",
   "feat_fast_repo_switch_title": "Repo wechseln, ohne die Tastatur zu verlassen",
-  "feat_fast_repo_switch_body": "Tippe auf dem Handy beim Schreiben auf den Repo-Namen in der Kopfzeile von „Neue Aufgabe“ — die Repo-Liste öffnet sich sofort, und nach der Auswahl landest du direkt wieder in deiner Beschreibung, die Tastatur bleibt oben. Zwei Tipps statt fünf."
+  "feat_fast_repo_switch_body": "Tippe auf dem Handy beim Schreiben auf den Repo-Namen in der Kopfzeile von „Neue Aufgabe“ — die Repo-Liste öffnet sich sofort, und nach der Auswahl landest du direkt wieder in deiner Beschreibung, die Tastatur bleibt oben. Zwei Tipps statt fünf.",
+  "cardmenu_stop": "Agent stoppen",
+  "cardmenu_stop_title": "Sendet Esc an diese Sitzung — bricht nur den laufenden Zug ab",
+  "cardmenu_stop_toast": "{name} gestoppt",
+  "cardmenu_stop_failed": "{name} konnte nicht gestoppt werden — läuft möglicherweise nicht mehr",
+  "feat_session_stop_title": "Einzelnen Agenten stoppen, ohne ihn zu öffnen",
+  "feat_session_stop_body": "Rechtsklick (oder langes Tippen) auf die Karte einer arbeitenden Sitzung, dann „Agent stoppen\" — das schickt ein einzelnes Esc nur an diese Sitzung. So brichst du einen entgleisten Zug ab, ohne die ganze Herde anzuhalten oder ins Terminal zu wechseln."
 }
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -3519,5 +3519,11 @@
   "feat_newtask_keymap_body": "In New Task, hold ⌘ (Ctrl on Windows/Linux) for a moment and every control shows its shortcut right where it sits — release to dismiss. Press ? for the full key card.",
   "diagnostics_hint_tailscale_serve_denied": "Tailscale refuses Shepherd's serve config, so no live preview can be published — run sudo tailscale set --operator=$USER once, then reopen the preview.",
   "feat_fast_repo_switch_title": "Switch repo without leaving the keyboard",
-  "feat_fast_repo_switch_body": "On a phone, tap the repo name in the New Task header while you're typing — the repo list opens straight away, and picking one drops you right back into your description with the keyboard still up. Two taps instead of five."
+  "feat_fast_repo_switch_body": "On a phone, tap the repo name in the New Task header while you're typing — the repo list opens straight away, and picking one drops you right back into your description with the keyboard still up. Two taps instead of five.",
+  "cardmenu_stop": "Stop agent",
+  "cardmenu_stop_title": "Send Esc to this session — stops the current turn, nothing else",
+  "cardmenu_stop_toast": "Stopped {name}",
+  "cardmenu_stop_failed": "Couldn't stop {name} — it may no longer be running",
+  "feat_session_stop_title": "Stop one agent without opening it",
+  "feat_session_stop_body": "Right-click (or long-press) a working session's card and pick Stop agent — it sends a single Esc to just that session, so you can cut one runaway turn short without halting the whole herd or switching to its terminal."
 }
--- a/ui/src/lib/api.ts
+++ b/ui/src/lib/api.ts
@@ -945,6 +945,15 @@ export async function replySession(id: string, text: string): Promise<void> {
   if (!r.ok) throw await failed(r, "reply");
 }
 
+/** Interrupt ONE session: a lone ESC to its pane, nothing else (#1993). No body — the
+ *  id IS the target. The server answers 404 for an unknown id or a dead/unlisted pane,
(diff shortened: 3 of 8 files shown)


scoop (commented):
## What I verified

**The API wrapper matches the real endpoint.** `interruptSession` (`ui/src/lib/api.ts:948-951`) POSTs `/api/sessions/${id}/interrupt` with no body and throws via `failed(r, "interrupt")` on non-ok. Compared against the server route at `src/server.ts:2594-2602`: it matches on `req.method === "POST" && parts[2] && parts[3] === "interrupt"`, takes no body, and returns `{ok:true}` / 404. `failed` is the shared helper at `ui/src/lib/api.ts:176-178`, same shape `replySession` uses (`ui/src/lib/api.ts:939-946`). URL interpolation without `encodeURIComponent` matches every other session-scoped call in that file, so no new inconsistency.

**Gating on `dStatus` rather than raw `status` is correct for THIS endpoint — I checked the thing that would have made it wrong.** `ui/src/lib/display-status.ts:9-11` warns that behavioral consumers must read raw status, and `ui/src/lib/components/TopBar.svelte:138-146` deliberately gates the fleet e-stop on raw `s.status === "running"` because `haltAll` skips any agent herdr does not report as working (`src/service.ts:4460`, `if (agent?.agentStatus !== "working") continue;`). That constraint does **not** transfer: `service.interrupt` (`src/service.ts:4521-4535`) only resolves a *live* pane via `liveAgentFor(id)` (`src/service.ts:4528`) — there is no `agentStatus === "working"` filter — so a working-while-blocked session (herdr status latched at blocked) IS reachable by the per-session interrupt. `UnitRow.svelte:595` therefore offers the item exactly where the endpoint can land, and `UnitRow.svelte:596-598` documents the reasoning.

**Failure path.** `stopFromMenu` (`ui/src/lib/components/UnitRow.svelte:671-681`) closes the menu first, awaits, and catches — a 404 becomes `m.cardmenu_stop_failed` with `alert: true`, never an unhandled rejection. Both branches share the key `session-interrupt:${session.id}`; I checked the keyed-refresh path in `ui/src/lib/toasts.svelte.ts:123-141` and it does re-write `alert` and `durationMs` on the existing toast, so a success (4s polite) followed by a failure correctly becomes 12s/assertive rather than inheriting the old severity.

**i18n parity.** All six keys exist in both catalogs at the same positions with the same `{name}` placeholders — `ui/messages/en.json:3523-3528` and `ui/messages/de.json:3523-3528`. Only two catalogs exist (`ui/messages/` holds `en.json` + `de.json`), and `ui/scripts/check-i18n.mjs:44-53` checks set equality + non-empty values, which this satisfies. German "Zug" for "turn" matches the established rendering in `ui/messages/de.json:203` (`feat_halt_body`, "der aktuelle Zug").

**Announcement gates.** `ui/src/lib/feature-announcements/entries/v1.47.0-session-stop.ts` stamps `sinceVersion: "1.47.0"` with a matching `v1.47.0-` filename prefix; last released version is `1.46.0` (`package.json:3`, `CHANGELOG.md:3`), so `scripts/check-announcement-versions.mjs` (`sinceVersion > released` + filename agreement) passes. Fragment shape matches its sibling `v1.47.0-fast-repo-switch.ts`, its `titleKey`/`bodyKey` resolve to the catalog keys above, and `id: "session-stop"` is unique across `entries/`. The catalog is glob-loaded (`ui/src/lib/feature-announcements.ts:47-50`), so no index edit is needed.

**No unintended reach.** `UnitRow` is instantiated in exactly one place, `ui/src/lib/components/herd/HerdGroup.svelte:109-133` (the herd list), which is the intended surface; adding `stoppable` to `hasMenu` (`UnitRow.svelte:607`) does not enable the menu anywhere it was not already enabled, since that call site always passes `ondecommission`/`onrename`. The `onstop` prop is optional (`CardMenu.svelte:40`) and every other caller of `CardMenu` is unaffected.

**Demo mode.** `ui/src/lib/demo/router.ts:325-328` falls back to `{ ok: true }` for any unmatched mutation, so the showcase gives a success toast rather than a spurious failure — same as the fleet halt, which is also absent from the demo route table.

**Accessible name.** The button's name comes from its text content (`CardMenu.svelte:200`); the `aria-hidden` glyph is excluded and `title` does not override content in the accname computation, so `getByRole("menuitem", { name: "Stop agent" })` resolves. Same construction as the existing decommission item (`CardMenu.svelte:249-258`).

**Task coverage.** All five work items from the issue are present: wrapper, placement decision (CardMenu), EN+DE keys, one announcement fragment, and 404-as-toast. The two design questions are answered as the plan settled them (running-only; interrupt-only, no combined steer).

### Nits (non-blocking):
- `ui/messages/de.json:3528`: `„Agent stoppen\"` mixes the German opening low quote with a straight ASCII closing quote; the neighbouring `feat_fast_repo_switch_body` (`ui/messages/de.json:3522`) uses the typographic pair `„…“`. (Line 3519 uses the straight form, so precedent is mixed — but `“` is the correct closer.)
- `ui/src/lib/components/UnitRow.svelte:596-599`: the comment justifies hoisting `onStop` out of the template for the complexity cap, yet the adjacent `onmergepr={mergeable ? mergeFromMenu : undefined}` (`UnitRow.svelte:994`) keeps its inline ternary — the asymmetry makes the stated rationale read as inconsistent.
- `ui/src/lib/components/CardMenu.svelte:197`: the "stops the current turn, nothing else" reassurance lives only in `title`, which never renders on the long-press/touch path the announcement copy specifically advertises.

Issue #2005, opened 2026-08-03
Epic: adopt the Claude 5 context-engineering rules in Shepherd's prompt assembly


scoop:
Anthropic published [The new rules of context engineering for Claude 5 models](https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models) (Thariq, 2026-07-24): they removed **over 80% of Claude Code's system prompt** for Opus 5 and Fable 5 with no measurable loss on coding evals. The finding is that these models were being *over-constrained* — system prompt, skills and `CLAUDE.md` issuing overlapping and sometimes conflicting instructions that the model has to reconcile before it can act.

Shepherd is an agent harness whose entire job is assembling that context. The guidance lands directly on it.

## Where Shepherd stands today

`composeSystemPrompt()` assembles one `--append-system-prompt` payload from ~20 named blocks:

| Spawn kind | Measured payload |
|---|---|
| attended Claude, no house rules | **8,389 chars** |
| + autopilot | 9,347 |
| plan-gate interactive | **13,171** |
| research | 6,320 |
| kitchen sink (house rules + build queue + preview + draft + trim) | 12,042 |

Roughly **5.5 KB of that floor is unconditional standing notices** — engineering-posture (1,643), tmpfs-worktree (1,303), worktree-stash (942), untrusted-content-boundary (805), research-first (502), branch-rename (302) — most of which describe situations a given session never encounters. On top of the spawn prompt: house rules up to 4,000 chars, the critic prompt at 9.3–15.2 KB, the plan reviewer at 4.5–6.8 KB, this repo's `CLAUDE.md` at 14 KB, three skills at ~30 KB.

**There is no instrumentation anywhere that measures the assembled prompt.** The only size guard is the OS argv budget, which fires three orders of magnitude above where these decisions live.

## The six shifts, mapped

| Then → Now | Where it bites Shepherd |
|---|---|
| Rules → judgement | The learnings flywheel manufactures imperative behavioral rules and injects up to 4 KB per spawn, then promotes them into `CLAUDE.md` |
| Examples → interface design | The build-queue directive is 3,443 chars of prose that bakes a literal control-plane token into a `curl` |
| Upfront → progressive disclosure | ~5.5 KB of always-resident situational notices — and drain spawns disable the skill catalog outright, foreclosing the remedy |
| Repeat yourself → say it once | The untrusted directive is stated in the prompt *and* restated inside every fence; epic-authoring exists in three places |
[...]

PR #2008, opened 2026-08-03
feat(prompt): per-block byte/token accounting for the spawn prompt

### LINKED ISSUE #1999: Instrument the assembled spawn prompt (per-block byte/token accounting)
## Goal

Make the size of Shepherd's assembled spawn context **measurable** before anything is cut. Every other slice in this epic is a deletion; without a number, "we trimmed it" is a vibe and a regression is invisible.

Today `composeSystemPrompt()` assembles a single `--append-system-prompt` payload from ~20 named blocks, and there is **no instrumentation anywhere that records its size**. The only size guard on that path is the OS argv budget (`argv-limit` / `spawn-budget`), which fires at ~128 KiB — three orders of magnitude above where the interesting decisions live. The post-hoc usage report attributes cost to `cacheRead` in aggregate but cannot attribute it to a block.

## Vertical cut

Have the composer return its blocks as named parts rather than a pre-joined string, record per-block byte count and estimated token count at spawn time, persist the breakdown alongside the session, and surface it in the UI by growing the existing house-rules budget meter into a full prompt-budget view.

## Measured baseline to reproduce

| Spawn kind | Payload |
|---|---|
| attended Claude, no house rules | 8,389 chars |
| + autopilot | 9,347 |
| plan-gate interactive | 13,171 |
| research | 6,320 |
| kitchen sink (house rules + build queue + preview + draft + trim) | 12,042 |

Unconditional floor is ~5.5 KB of standing notices: engineering-posture (1,643), tmpfs-worktree (1,303), worktree-stash (942), untrusted-content-boundary (805), research-first (502), branch-rename (302).

## Acceptance

- For any spawn (attended **and** drain, Claude **and** Codex) the operator can see total payload bytes plus a per-block breakdown.
- A test asserts the recorded per-block sizes plus separators reconstruct the exact emitted payload — the meter cannot silently drift from what is sent.
- The numbers in the table above are reproduced by the instrument, so later slices can be scored against them.

scoop:
Closes #1999. Part of epic #2005 (adopt the Claude 5 context-engineering rules).

## Why

`composeSystemPrompt()` assembles ~20 named blocks into the single payload every spawn carries, and **nothing anywhere measured it**. The only size guard on that path is the OS argv budget (`src/argv-limit.ts`), which fires at ~128 KiB — three orders of magnitude above where this epic's decisions live. Every later slice in the epic is a deletion; without a number, "we trimmed it" is a vibe and a regression is invisible.

Nothing is deleted or reworded here. `composeSystemPrompt()`'s output is **byte-identical**.

## What

- **`src/prompt-budget.ts`** — pure, I/O-free: `PromptBlock`, `measurePromptBlocks()`, `joinPromptBlocks()`, `estimateTokens()`, and the one shared `PROMPT_BLOCK_SEPARATOR`.
- **`src/service.ts`** — `composeSystemPromptBlocks()` returns the named parts; `composeSystemPrompt()` is now `joinPromptBlocks()` over it. Each block's name is the XML tag already wrapping it, so a measurement maps straight back to its text.
- **Recorded at `composeDirectives()`** — the one chokepoint both spawn builders call, so **attended and drain, Claude and Codex** are covered by a single write. `delivery` records the channel (`--append-system-prompt` vs. inline `<shepherd-directives>`), so the Codex meter isn't misread as its literal argv.
- **`session_prompt_budget`** — one upserted row per session; a relaunch replaces rather than accumulates.
- **`GET /api/prompt-budget`** + **`GET /api/sessions/:id/prompt-budget`**.
- **Usage → Prompt lens** — pick any recorded spawn, see total chars/bytes/≈tokens and the per-block breakdown, largest first.

## Chars, not bytes

The epic's baseline table turns out to be stated in **characters**; these blocks are em-dash dense enough that UTF-8 bytes run ~0.7% higher (8,389 chars = 8,451 bytes). The instrument records **both** — a bytes-only meter could not reproduce the epic's own numbers, and the argv guard works in bytes.

## Discrepancy worth flagging

Four of the epic's five baseline rows reproduce **exactly** on this tree and are pinned by test:

| Spawn kind | chars | bytes |
|---|---|---|
| attended Claude, no house rules | 8,389 ✓ | 8,451 |
| + autopilot | 9,347 ✓ | 9,413 |
| plan-gate interactive | 13,171 ✓ | 13,273 |
| research | 6,320 ✓ | 6,370 |

The fifth ("kitchen sink", 12,042) is **not reproducible from the issue**: it depends on the house-rules text (unstated) and on the build-queue directive, whose length varies with the session id / base URL / token baked into its `curl` commands — measured 3,534 chars token-less here vs. the epic's quoted 3,443. Rather than report a false match, the test pins that shape to its own measured value (13,408) with the fixture written out inline, so it is reproducible from the test from here on.

## Why the Usage modal, not the house-rules meter

The issue suggested growing the house-rules budget meter. That meter is fed by `LearningsService.injectableOverview()`, which deliberately runs the planner with **no session** — it is per-repo and structurally cannot show what a given spawn sent, while the acceptance criterion is per-spawn. The Usage modal already hosts `OverheadLens` ("where does the spend go?"); the prompt budget is the input-side counterpart to exactly the question the issue says the post-hoc report cannot answer. The house-rules block appears there as one row among the others.

## Notable decisions

- **Tokens are an estimate**, not a measurement: `chars / CHARS_PER_TOKEN` (4), behind one exported constant. No tokenizer dependency lands on the spawn path — the epic's decisions turn on *relative* block weight, and a shared divisor preserves every ratio. The UI shows `≈`, with an InfoTip and a glossary definition saying so.
- **Measured at assembly**, so a spawn later refused by the sandbox auto-gate leaves a row for a session that never starts. That is honest (the payload *was* assembled) and invisible to the reads, which inner-join `sessions`. Such orphans are deliberately not swept — a sweep would race the very window it targets.
- **No new session-snapshot field and no WebSocket push** — putting ~20 block names on every session broadcast is the kind of unconsidered payload growth this epic exists to stop.

## Verification

- `test/prompt-budget.test.ts`: across a 13-shape option matrix (every branch, both providers, both plan-gate variants, both operator languages) the blocks join back to the byte-identical payload **and** per-block sizes plus separators reconcile to it in chars and bytes — the acceptance criterion, asserted rather than assumed.
- Root: `bun run lint`, `bunx tsc --noEmit`, `bun test ./test` (8,139 pass).
- UI: `bun run check`, `check:i18n`, `bun run test` (4,512 pass).
- `fallow@2.100.0 audit --base origin/main --fail-on-issues`: clean. (The split briefly pushed the composer to cognitive 16; the orthogonal trailing blocks are extracted into `trailingBlocks()`, mirroring the existing `primaryDirectiveBlock()` extraction.)
- End-to-end store round-trip checked on the Codex/drain path, not only on unit fixtures.


scoop (commented):
## What I verified

**The composer split is byte-identical.** I compared every emission site in the diff against the literal it replaced. `taggedBlock` produces `` `<${name}>\n${body}\n</${name}>` `` (`src/service.ts:1322-1324`), which is character-for-character the old inline template at each of the 13 rewritten push sites (e.g. old `` `<engineering-posture>\n${ENGINEERING_POSTURE}\n</engineering-posture>` `` → `taggedBlock("engineering-posture", ENGINEERING_POSTURE)` at `src/service.ts:1473`). Emission order is preserved: primary directive, then build-queue → preview-hint → draft-mode → context-trim → operator-language in `trailingBlocks` (`src/service.ts:1381-1398`), matching the old tail at the pre-diff `composeSystemPrompt`.

The one semantic subtlety — the old code ended with `.filter(Boolean).join("\n\n")`, which could in principle have dropped an empty block — is safe: the only nullable member was `operatorLanguageBlock`, which returns `null` or a non-empty `<operator-language>…` string (verified at `src/operator-language.ts:51-65`), and it is now null-checked at `src/service.ts:1396-1397`. `houseRules` keeps the identical truthiness test (`src/service.ts:1479`). No other block can be empty.

**Block names really are the wrapping tag.** `shepherd-house-rules` is named without re-wrapping (`src/service.ts:1484`) and that is the actual tag `renderHouseRulesBlock` emits — `HOUSE_RULES_TAG = "shepherd-house-rules"` at `src/house-rules.ts:9`. `operator-language` likewise matches `src/operator-language.ts:64`.

**The reconstruction invariant holds by construction and is asserted.** `PROMPT_BLOCK_SEPARATOR` is the single constant used by both `joinPromptBlocks` and `measurePromptBlocks` (`src/prompt-budget.ts:44,73,83`); `separatorChars` floors at zero for the empty case (`src/prompt-budget.ts:83`). `test/prompt-budget.test.ts:74-94` asserts `joinPromptBlocks(blocks) === composeSystemPrompt(...)` plus sum-of-blocks + separators === payload length/byte-length across a 13-row option matrix covering both providers, both plan-gate variants, both operator languages, and every orthogonal flag. That is the issue's own acceptance criterion.

**Store contract.** The new table's columns match the writer's parameter list and the reader's projection (`src/store.ts:1512-1515`, `3122-3148`, `875-880`). The join columns exist on `sessions`: `desig`, `repoPath`, `agentProvider`, `auto` (verified at `src/store.ts:1137-1147`). `auto` is `INTEGER NOT NULL DEFAULT 0` and is hydrated as `row.auto === 1` (`src/store.ts:3170`) — correct. Upsert-by-`sessionId` is a real `ON CONFLICT(sessionId) DO UPDATE` (`src/store.ts:3128-3132`), so a relaunch replaces rather than accumulates, as claimed. The prune cascade is added in the same transaction, before `DELETE FROM sessions` (`src/store.ts:3905-3908` vs `3912`).

The DDL comment's factual claims check out: the budget row is written before the session row exists (`src/service.ts:3164` comment + `store.create()` at `src/service.ts:3194`, called after `prepareSpawnOrThrow` at `3183`), and `session_injected_learnings` indeed carries no foreign key (`src/store.ts:1381-1383`).

**Both spawn builders funnel through the write.** `composeDirectives` has exactly two call sites — `buildSpawnArgv` (`src/service.ts:2646`) and `buildCodexSpawnArgv` (`src/service.ts:2695`) — and drain reaches them via `service.create()` (`src/drain.ts:1974`, `2594`). So attended + drain and Claude + Codex are covered by the single `recordPromptBudget` call at `src/service.ts:2590`. `delivery` is derived from `agentProvider` (`src/service.ts:2606`) and the failure path only warns (`src/service.ts:2613`), so the instrument cannot take down a spawn.

**Server.** `handlePromptBudget` returns `Response | null` synchronously; the dispatch loop `await handle(ctx)` tolerates that and sibling sync handlers already exist (`src/server.ts:6431`, `6537`, dispatch at `7657-7660`). Auth runs before `ROUTE_HANDLERS` (`src/server.ts:7645-7648`), and the new routes are *not* in the agent-ingress allowlist (`isAgentIngressRoute`, `src/server.ts:7693-7709`), so a spawned agent cannot read them. The per-session branch is placed after `usage`/`activity` and before the bare `sessionRead` fall-through, so nothing shadows it (`src/server.ts:2350-2364`). No prompt *text* is persisted or served — only names and sizes.

**UI wiring.** All six `$derived` arms were updated for the new tab (`ui/src/lib/components/Usage.svelte:164-182`) plus the banner (`309`) and the lens mount (`323`); an empty array is truthy so `{:else if tab === "prompt" && prompts}` renders the lens' own empty state rather than falling through to the loading arm. The `<select value={…} onchange={…}>` (uncontrolled) pattern matches the established one at `ui/src/lib/components/EpicPanel.svelte:249`. Every identifier resolves: `UsageBar`'s `value`/`max`/`tone` props (`ui/src/lib/components/usage/UsageBar.svelte:3-15`), `formatPct` (`ui/src/lib/components/usage/format.ts:26`), `InfoTip`'s `text`/`label` (`ui/src/lib/components/InfoTip.svelte:18-22`), `GlossaryText`'s `text` (`ui/src/lib/components/GlossaryText.svelte:5`), `coachTarget` (`ui/src/lib/actions/coachTarget.svelte.ts:14`), `m.newtask_info_aria({topic})` (`ui/messages/en.json:626`, `ui/messages/de.json:626`). Design tokens `--color-amber`, `--color-slate`, `--color-faint`, `--color-inset`, `--color-line-bright`, `--fs-micro/meta/base/lg` all exist (`ui/src/app.css:12,16,20,21,25,85-88`); `tone` is a `var(--color-*)` reference as `UsageBar` requires.

**Locale parity.** All 18 new keys exist in both catalogs with identical spelling and non-empty values (`ui/messages/en.json:3522-3541` vs `ui/messages/de.json:3522-3541`), which is exactly what `ui/scripts/check-i18n.mjs:46-53` enforces. The glossary term `spawn-prompt` uses the `gloss_<id>_term`/`_def` convention required by `ui/src/lib/glossary.ts:3-8` and is referenced as `[[spawn-prompt|…]]` in `usage_prompt_caption` in both locales. The announcement entry's `sinceVersion: "1.47.0"` is ahead of `package.json`'s `1.46.0` (as the catalog contract at `ui/src/lib/feature-announcements.ts:11-17` demands), its filename matches the version, it is picked up by the eager glob (`ui/src/lib/feature-announcements.ts:47-50`), and its `targetId` matches the `use:coachTarget={"usage-prompt-tab"}` anchor.

**On the baseline table.** Four of the issue's five rows are pinned by value (`test/prompt-budget.test.ts:112-140`). The fifth (kitchen sink, 12,042) genuinely is not reproducible from the issue: `buildQueueDirective` embeds `${baseUrl}/api/sessions/${sessionId}/queue` and an optional bearer header (`src/service.ts:951-959`), so its length varies with session id, base URL and token, and the house-rules text is unstated. The test pins its own explicit fixture instead and says so (`test/prompt-budget.test.ts:142-170`). That is the honest handling; worth repeating in the PR body, since the commit message does not mention the discrepancy.

### Stated limitations

- I did not run lint, tsc or either test runner (CI's job per the brief). The pinned baseline numbers (8389/9347/13171/6320 and 13408) are asserted by the test but not independently re-measured by me.

## Nits (non-blocking):

- `src/server.ts`: `?limit=` with an empty value returns 1 record, not the default 50 — `url.searchParams.get("limit")` yields `""` (not `null`), so the `??` default is skipped and `Number("")` is `0`, which `Math.max(1, …)` clamps to 1. Harmless for the UI (which sends no `limit`), but `raw > 0` would read more like the comment's intent.
- `src/server.ts`: neither new endpoint has a server-level test, unlike its siblings (`test/server-usage-breakdown.test.ts`, `test/server-usage-timeline.test.ts`) — the limit clamp and the 404 arm are untested.
- `ui/src/lib/components/usage/PromptLens.svelte`: per-block `bytes` are recorded and served but never rendered (only chars, ≈tokens, share), so the byte column is API-only; totals do show bytes.
- `ui/src/lib/components/usage/PromptLens.svelte`: `formatCount` uses `n.toLocaleString()` with the browser locale rather than the Paraglide locale — consistent with `ui/src/lib/components/usage/GithubLens.svelte:106`, so only a nit.
- `src/store.ts`: orphaned rows from refused spawns are never swept. The DDL comment's rationale (a sweep would race the write→`store.create()` window) is a bit overstated — `pruneOrphanInjectedLearnings` (`src/store.ts:4817-4820`) has the same race and exists, and an age filter would close it — but the trade-off is small and deliberately documented.

- `src/server.ts`: `GET /api/sessions/:id/prompt-budget` adds a public read surface that no client consumes (the lens uses only the list endpoint). Plan-sanctioned and cheap, but it is one more route to keep working.

scoop (commented):
## What I verified

**The composer split is output-preserving.** `composeSystemPrompt` is now `joinPromptBlocks(composeSystemPromptBlocks(...))` (`src/service.ts:1543-1548`), and `joinPromptBlocks` is `blocks.map(b => b.text).join(PROMPT_BLOCK_SEPARATOR)` with `PROMPT_BLOCK_SEPARATOR === "\n\n"` (`src/prompt-budget.ts:39`, `:68-70`) — the same separator the old code passed to `.join("\n\n")`. The old code's `.filter(Boolean)` (removed in this diff) could only ever drop one member: `operatorLanguageBlock(...)`, which returns `null` for `"en"` (`src/operator-language.ts:51-52`) — every other array member was either pushed under a truthiness guard or produced by a `<tag>…</tag>` template that is never empty. That case is now handled explicitly at `src/service.ts:1390-1391`. I walked each of the 18 push sites and confirmed `taggedBlock(name, body)` → `` `<${name}>\n${body}\n</${name}>` `` (`src/service.ts:1322-1324`) reproduces the literal it replaced, and that `trailingBlocks` (`src/service.ts:1377-1398`) preserves the old emission order build-queue → preview-hint → draft-mode → context-trim → operator-language.

**The two blocks that are *not* re-wrapped are named correctly.** `shepherd-house-rules` (`src/service.ts:1484`) matches `renderHouseRulesBlock`, which emits `` `<${HOUSE_RULES_TAG}>\n…` `` with `HOUSE_RULES_TAG = "shepherd-house-rules"` (`src/house-rules.ts:9`, `:275-279`) and returns `null` (not `""`) when nothing is injected. `operator-language` (`src/service.ts:1391`) matches `src/operator-language.ts:64`.

**Both spawn paths are covered by the one write.** `composeDirectives` is called from exactly two places — `buildSpawnArgv` beside `--append-system-prompt` (`src/service.ts:2644-2659`) and `buildCodexSpawnArgv` before the inline `<shepherd-directives>` wrap (`src/service.ts:2695-2705`) — and `recordPromptBudget` sits inside it (`src/service.ts:2590`). Drain reaches both via `service.create()`, so attended/drain × Claude/Codex are all covered. The resume paths (`src/service.ts:2755`, `:4652`) carry only the operator-language block and are correctly left uninstrumented.

**Store.** DDL is inside the `CREATE TABLE IF NOT EXISTS` block (`src/store.ts:1512-1515`), so existing DBs get it with no migration. Upsert-by-`sessionId` (`src/store.ts:3126-3147`) means a relaunch replaces rather than accumulates. Both reads share `PROMPT_BUDGET_SELECT` (`src/store.ts:879-882`), which INNER JOINs `sessions` for `desig`/`repoPath`/`agentProvider`/`auto` — all four columns exist (`src/store.ts:543-546`, `migrateSessionColumns` at `:3931`, `:3942`). The prune cascade is added at `src/store.ts:3905-3908`, and `DELETE FROM sessions` occurs only there (`src/store.ts:3912`), so there is no second uncascaded delete path.

**Server.** `handlePromptBudget` is registered in `ROUTE_HANDLERS` (`src/server.ts:7587`) and dispatched via `await handle(ctx)` (`src/server.ts:7657-7660`), so a sync handler is fine. Its guard `parts[1] === "prompt-budget" && !parts[2]` (`src/server.ts:4249-4250`) cannot collide with the session read at `src/server.ts:2352-2357`. Neither route is in `isAgentIngressRoute` (`src/server.ts:7693-7709`), so spawned agents on the auth-exempt loopback listener cannot read another session's breakdown — correct. The payload carries block *names and sizes only*, never block text, so the API-token-bearing build-queue directive is not exposed.

**i18n / glossary / discovery.** All 19 new keys exist in both catalogs at identical positions (`ui/messages/en.json:3522-3541` vs `ui/messages/de.json:3522-3541`), including the `{count}` placeholder in `usage_prompt_blocks_heading` on both sides. The `[[spawn-prompt|…]]` marker in `usage_prompt_caption` resolves to the registry entry added at `ui/src/lib/glossary.ts:143-148`, whose `gloss_spawn_prompt_term`/`_def` keys exist in both catalogs — this satisfies `scripts/check-glossary.mjs:128-142`. `m.newtask_info_aria({topic})` used by the InfoTip exists (`ui/messages/en.json:626`, `ui/messages/de.json:626`). The announcement's `sinceVersion: "1.47.0"` is correctly ahead of the last release `1.46.0` (`.release-please-manifest.json`, `package.json:3`) and matches its filename, per `ui/src/lib/feature-announcements.ts:11-17`; entries are auto-globbed so no index edit is needed.

**UI contract.** `UsageBar` takes `{value, max, tone}` (`ui/src/lib/components/usage/UsageBar.svelte:2-14`) and `tone="var(--color-amber)"` is a real token (`ui/src/app.css:21`); every other custom property used by `PromptLens.svelte` resolves in `ui/src/app.css` (`--color-inset:12`, `--color-line:15`, `--color-line-bright:16`, `--color-ink:17`, `--color-ink-bright:18`, `--color-muted:19`, `--color-faint:20`, `--color-slate:25`, `--font-mono:29`, `--fs-micro:85`, `--fs-meta:86`, `--fs-base:87`, `--fs-lg:88`). `InfoTip` takes `{text,label}` (`ui/src/lib/components/InfoTip.svelte:18-22`), `GlossaryText` takes `{text}` (`ui/src/lib/components/GlossaryText.svelte:5`), `formatPct` exists (`ui/src/lib/components/usage/format.ts:25-27`). All six `Usage.svelte` `$derived` arms got the `"prompt"` case (`ui/src/lib/components/Usage.svelte:162`, `:169`, `:181`) plus the render arm (`:323`) and `retry()` (`:151`), so the tab cannot hang on "Loading" — an empty `[]` is truthy and renders the lens's own empty state.

## Could not verify

- The four pinned baseline numbers (8389 / 9347 / 13171 / 6320 in `test/prompt-budget.test.ts:115-120`) — I did not execute the suite, per the brief. They are asserted by the test, so CI is the ground truth.
- Svelte 5's handling of a non-bound `value` attribute on `<select>` (`ui/src/lib/components/usage/PromptLens.svelte:61`) — `ui/node_modules` is not installed in this worktree, so I could not read the compiler source. Either way the visible behaviour is unaffected: the default selection is the first option, which is exactly `records[0]`, the lens's default `selected`.

## Nits (non-blocking):
- `src/server.ts:4253-4256`: `?limit=` (empty) or `?limit=0` yields **1** record, not the documented default of 50 — `Number("")` is `0`, which is finite and clamps up to 1; `docs/external-task-api.md:157` describes out-of-range as "falls back", which reads as "falls back to the default".
- `ui/src/lib/components/Usage.svelte:132-143`: `loadPrompts()` has no monotonic request token, unlike `loadBreakdown` (`:86-101`) and `loadTimeline` (`:111-124`); rapid tab toggling can let an earlier response land after a later one. Harmless today (same URL, same shape) but it is the one lazy loader that diverges from the established pattern.
- `src/service.ts:1484` and `:1391`: the two hand-named blocks hardcode `"shepherd-house-rules"` / `"operator-language"` rather than importing `HOUSE_RULES_TAG` (`src/house-rules.ts:9`); a tag rename would silently desync the meter's label from the payload.
- `ui/src/lib/components/usage/PromptLens.svelte:48`, `:89`: `class="panel"` has no matching rule anywhere in `ui/src/app.css` — copied from `OverheadLens.svelte:70`, so consistent, but it is dead markup either way.
- `src/store.ts:1506-1511`: orphan rows from a refused spawn are deliberately never swept, whereas the sibling table with the same write-before-`create()` ordering does get a daily sweep (`pruneOrphanInjectedLearnings`, `src/store.ts:4816`, called from `src/index.ts:2573`). The stated reason (the sweep would race the write→create window) applies equally to that sibling, so the divergence is a judgement call rather than a constraint.
- `test/prompt-budget.test.ts:127-135`: the baseline test rebuilds its option set as three parallel arrays indexed by `i` alongside the already-built `payloads`, which is easy to desync on a future edit; a single array of `{label, chars, bytes, houseRules, autopilot, opts}` rows would be harder to get wrong.

- `src/prompt-budget.ts:35-36`: `separatorChars` is computed and exported but never persisted (`session_prompt_budget` has no such column) nor surfaced in the lens — it is derivable as `(blocks.length - 1) * 2`, so the field exists only for the test's convenience.
- `docs/external-task-api.md:145-155`: documents the new reads in the *external task API* doc even though neither route is on the agent-ingress allowlist; consistent with the neighbouring `/diff/annotations` and `/scratchpad` entries, so this is precedent rather than a new surface, but it does widen the documented public API beyond what the task asked for.

- `src/store.ts:3150-3159`: `hydratePromptBudget` accepts any JSON array as `PromptBlockMeasure[]` with an unchecked cast — a row whose `blocks` is e.g. `[1,2]` would reach `PromptLens` and make `formatCount(b.chars)` throw on `undefined.toLocaleString()`. Only reachable via a hand-edited or corrupted DB today, since the sole writer is `measurePromptBlocks`.
- `ui/src/lib/components/usage/PromptLens.svelte:94`: `{#each blocks as b (b.name)}` keys on the block name; a duplicate name in a persisted row would raise Svelte's duplicate-key error rather than degrading. The composer guarantees uniqueness today (asserted at `test/prompt-budget.test.ts:106`), so this is dormant.

## On the acceptance criteria

All three are met. Per-block breakdown for attended and drain, Claude and Codex: one write at the shared chokepoint. The reconstruction invariant: `test/prompt-budget.test.ts:74-94` asserts both `joinPromptBlocks(blocks) === composeSystemPrompt(...)` and `sum(chars) + separators === payload.length` across a 13-shape matrix. The baseline table: four of five rows are pinned to the issue's exact figures; the fifth ("kitchen sink", 12,042) is pinned to a measured 13,408 against an explicit inline fixture, with the reason it is not reproducible from the issue spelled out at `test/prompt-budget.test.ts:143-146`. That is the honest resolution and matches what the approved plan flagged as a risk — worth restating in the PR body so the delta is not read as a regression.

PR #2010, opened 2026-08-03
fix(skills): hand epic import to the operator, agents can't reach it

### LINKED ISSUE #2009: epic-authoring skill instructs an import call agents cannot reach (401 at /api/epic/import)
## What happens

The shipped `shepherd-epic-authoring` skill instructs the agent, at Stage 4, to wire native sub-issue and `blocked_by` links by calling:

```bash
curl -s -X POST -G "http://127.0.0.1:7330/api/epic/import" \
  --data-urlencode "repo=$(git rev-parse --show-toplevel)" \
  --data-urlencode "parent=<PARENT_NUM>"
```

A spawned agent cannot reach that route. It returns `{"error":"unauthorized"}`.

## Why

Spawned agents don't carry the human cookie, and autonomous ones run under `--clearenv` so they carry no bearer token either. They reach the server through the restricted loopback ingress listener, whose allowlist (`isAgentIngressRoute`, `src/server.ts:7667`) admits exactly three route families:

- `POST /api/sessions/<id>/hooks`
- `GET|PUT /api/sessions/<id>/queue` and `POST …/queue/steps/<stepId>`
- `GET|PUT /api/sessions/<id>/epic-draft`

`/api/epic/import` is not among them — it isn't even session-scoped, so it doesn't fit the allowlist's shape (the per-session UUID is what stands in for the agent's credential). Hitting the main port instead lands on `checkAuth` and 401s.

The skill's own draft-only branch is consistent with the boundary: when an `<epic-authoring-directive>` is present it tells the agent to PUT the draft and stop, leaving the GitHub writes to the server. It's the **unguided** path — an attended session invoking the skill directly, with no directive — that walks into the unreachable call. That's how this surfaced: authoring #2005 by hand, following the skill end to end.

## Impact

Low but real. The dag fence in the parent body is the actual recognition contract, so the epic is recognized and drainable without import; only the native GitHub sub-issue and `blocked_by` links go unwired. The failure is loud (a 401, not a silent no-op), but the skill gives no recovery path, so an agent that hits it either gives up or — worse — goes looking for the operator password to get around a boundary that exists precisely to stop that.

## Options

1. **Fix the skill.** Stage 4 hands the import to the operator: print the exact authenticated command (or point at the UI affordance) and stop. Cheapest, and it keeps the ingress surface unchanged.
2. **Extend the ingress.** Add a session-scoped import route so the agent can trigger import for its own session's epic. Larger call — it widens the agent's control-plane reach, and import performs GitHub writes, which is the thing the epic-draft flow deliberately keeps server-side.

Option 1 looks right unless there's a reason agent-triggered import is wanted.

## Acceptance

- An attended session invoking `shepherd-epic-authoring` without an injected directive reaches a working end state — either it imports successfully, or it hands a runnable command to the operator and says why.
- Whichever option is taken, the skill's Stage 4 and its "Gate rules" reference section agree with what an agent can actually do.

Surfaced while authoring epic #2005.

scoop:
Closes #2009.

## What was wrong

`shepherd-epic-authoring` Stage 4 told the agent to `POST /api/epic/import` to wire an epic's native sub-issue + `blocked_by` links. No spawned agent can reach that route:

- Agents talk to the server only through the restricted loopback ingress, whose allowlist (`isAgentIngressRoute`, `src/server.ts`) admits three **session-scoped** families — `hooks`, `queue`, `epic-draft`.
- `/api/epic/import` is repo-scoped, so it can't even take that shape (the per-session UUID is what stands in for the agent's credential). Hitting the main port instead lands on `checkAuth` → `{"error":"unauthorized"}`.

The failure is loud, but the skill gave no recovery path — so an agent either gave up or went looking for the operator password to get around a boundary that exists precisely to stop that.

**`shepherd-onboarding` carried the identical unreachable curl** (Stage 5 step 3, its completion criterion, the Stage 6 verify sentence, and its Gate-rules bullet). Not named in the issue; found while researching it, fixed here.

## What changed

Option 1 from the issue — fix the skills; the ingress surface is untouched. Import performs GitHub writes, which is exactly what the epic-draft flow deliberately keeps server-side.

The import step is now a **handoff**, in both skills:

1. States the boundary and why a 401 there is the system working — plus an explicit **never hunt for the operator password or an API token** rule.
2. **UI path first:** Backlog → expand the epic parent's issue row → **Import structure** (`EpicPanel.svelte`, shown for a marker-derived epic), also offered as a remediation in that panel's **Diagnose** modal.
3. **Operator-run curl second**, marked `OPERATOR-RUN`, carrying the bearer header and the honest caveat that it only works when `SHEPHERD_TOKEN` is configured (optional, unset by default) — otherwise the UI is the path.
4. Makes the non-blocking part loud: the body marker **is** the recognition contract, so the epic is recognized and drainable already; import only wires the native links, and nothing waits on it. The downstream verify steps now check the marker rather than the wired links.

Gate-rules sections in both skills were realigned to match — no remaining claim that the agent creates, imports, or waits on import.

## Regression gate

`test/skill-agent-reachability.test.ts` (following the `test/epic-parse.test.ts` precedent of asserting skill-doc content):

- Asserts `isAgentIngressRoute("POST", ["api","epic","import"])` is `false` — the fact the skill text depends on. If the ingress is ever widened to admit import, this fails and forces the handoff wording to be revisited in the same change.
- Per skill: `/api/epic/import` occurs **exactly once** and that mention sits under the `OPERATOR-RUN` sentinel. "Assert absent" isn't available — the path legitimately survives inside the command printed for the operator — so the test asserts the shape instead. A re-introduced agent-facing call (a second mention, or an unmarked one) fails here.

## Verification

- `bun test ./test` — 8136 pass / 0 fail (includes the pre-existing `epic-parse` skill assertions and the new test).
- `bun run lint`, `bunx tsc --noEmit`, `bunx prettier --check` on the touched files — clean.
- `bunx fallow audit --base origin/main --fail-on-issues` — no issues in the 3 changed files.

UI half untouched. No i18n, glossary, or feature-catalog entry: nothing user-facing ships (`fix`, not `feat`).

### DIFF
--- a/.claude/skills/shepherd-epic-authoring/SKILL.md
+++ b/.claude/skills/shepherd-epic-authoring/SKILL.md
@@ -1,6 +1,6 @@
 ---
 name: shepherd-epic-authoring
-description: Author a Shepherd-recognized epic in an attended session — decompose work into child issues, create them, and mark the parent body with the structural epic marker (fenced dag / task-list), optionally wiring native sub-issue + blocked_by links via epic import. Use when asked to create an epic with sub-issues, promote an existing issue #N to an epic, split an issue into child issues, or make a body of work drainable as an epic — especially mid-session, where no injected epic guidance exists.
+description: Author a Shepherd-recognized epic in an attended session — decompose work into child issues, create them, mark the parent body with the structural epic marker (fenced dag / task-list), and hand the optional native sub-issue + blocked_by link import to the operator. Use when asked to create an epic with sub-issues, promote an existing issue #N to an epic, split an issue into child issues, or make a body of work drainable as an epic — especially mid-session, where no injected epic guidance exists.
 ---
 
 # Shepherd Epic Authoring
@@ -21,7 +21,7 @@ gap: invoke it whenever an epic ask arises, at spawn or mid-session. It is
 attended by default (unattended drains run with skills disabled unless the
 operator opts out of context trimming) and
 richer than the injected blocks: it drafts the whole tree, gates outward
-actions on approval, and wires native links.
+actions on approval, and hands the native-link import back to the operator.
 
 This skill is **self-contained**: it ships in the Shepherd repo and runs in
 _other_ operators' repos, so everything it needs is below.
@@ -58,14 +58,14 @@ If the system prompt carries an **`<epic-authoring-directive>`** with an epic-dr
 (`PUT …/api/sessions/<id>/epic-draft`), Shepherd's guided epic-draft flow is driving this session.
 In that mode the hard gate and the GitHub writes are **owned by the server**: follow ONLY the
 decomposition guidance below (Stage 1), emit the draft by PUTting it to that endpoint, then **STOP**.
-Do **not** perform Stages 3–5 (`gh issue create`/`gh issue edit`, the epic-import endpoint) — the
+Do **not** perform Stages 3–5 (`gh issue create`/`gh issue edit`, the import handoff) — the
 operator approves the draft in the UI and the server materializes it. The directive is authoritative;
 this skill only lends it the slicing/authoring guidance.
 
 ## Stages
 
 Work the stages in order; track one todo per stage. **Nothing outward (issue
-creation, body edits, epic import) happens before the Stage 2 approval gate.**
+creation, body edits) happens before the Stage 2 approval gate.**
 
 ### 0. Orient
 
@@ -82,10 +82,10 @@ git remote -v                      # is there a GitHub remote?
 gh auth status 2>/dev/null         # is gh usable?
 ```
 
-A working GitHub remote + `gh` ⇒ **GitHub-native** (issue creation + epic
-import available). Otherwise **local/lightweight**: programmatic creation and
-import are unavailable (import asserts a GitHub-native forge) — the deliverable
-becomes an importable markdown file instead (see Stage 3).
+A working GitHub remote + `gh` ⇒ **GitHub-native**: you can create issues, and
+the operator can run the link import (Stage 4). Otherwise **local/lightweight**:
+programmatic creation is unavailable and import has no forge to write to — the
+deliverable becomes an importable markdown file instead (see Stage 3).
 
 ### 1. Draft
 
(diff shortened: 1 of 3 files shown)


scoop (commented):
## What I checked

Docs-only change to two SKILL.md files plus one new regression test. Every factual claim the new skill text makes is a claim about this repo's code, so I resolved each one.

**The boundary claim (the whole point of the PR).** `isAgentIngressRoute` admits only `api/sessions/<id>/{hooks,queue,queue/steps/<id>,epic-draft}` — it returns `false` at the very first guard for anything whose `parts[1] !== "sessions"` (src/server.ts:7667-7683). So `POST /api/epic/import` (repo-scoped, src/server.ts:7073-7099) genuinely cannot take the allowlist's shape, and the new test's `isAgentIngressRoute("POST", ["api","epic","import"]) === false` assertion is verified against that code, not assumed. The main-port fallback body the skills quote verbatim — `{"error":"unauthorized"}` with 401 — is exactly what `checkAuth` returns (src/server.ts:726).

**The bearer caveat.** `checkAuth` accepts a bearer only when `config.token !== null` (src/server.ts:723), and `config.token` is `process.env.SHEPHERD_TOKEN ?? null` — optional, null by default (src/config.ts:600). So the skills' "the bearer header only works if `SHEPHERD_TOKEN` is configured on that server — it is optional and unset by default; without it, use the UI" is accurate. `7330` as the default port is verified against src/config.ts:469. The printed curl also works as printed: `handleEpicImport` does not call `requireJsonContentType`, and `checkOrigin` only rejects a *present* disallowed Origin (src/server.ts:765-781), so a bodyless curl with no Origin passes the CSRF seam.

**The UI path the skills now name as primary.** The `Import structure` button is `m.epic_import()` rendered under `{#if epic.source === "markdown"}` on the epic panel (ui/src/lib/components/EpicPanel.svelte:102-115), and that panel is the inline expansion of the parent's issue row in the Backlog (ui/src/lib/components/issues-panel/IssueRow.svelte:270-277) — so "open the Backlog, expand the parent issue's row, press Import structure" is a real path. A just-marked parent with no native sub-issues resolves to `source: "markdown"` (src/server.ts:6773-6789; src/epic-model.ts:116-118), so the button is present at exactly the moment the skill points at it. The Diagnose alternative also exists: findings with `action === "import-structure"` surface a confirm-import control (ui/src/lib/components/EpicDiagnosisModal.svelte:31-50), fed by src/epic-diagnosis.ts:142. Message-catalog parity holds — `epic_import` exists in both ui/messages/en.json:1211 and ui/messages/de.json:1211 (no new keys are added by this PR, so nothing to add).

**Retained response wording.** `subIssuesAdded` / `dependenciesAdded` / `unresolved` are the real `ImportResult` fields (src/epic-import.ts:4-9), returned unwrapped at src/server.ts:7098.

**The new test's mechanics.** `/api/epic/import` occurs exactly once per skill file (.claude/skills/shepherd-epic-authoring/SKILL.md:172 and .claude/skills/shepherd-onboarding/SKILL.md:215 — grepped the whole tree, no other occurrence in either file). The `OPERATOR-RUN` sentinel sits 4 lines above each (SKILL.md:168 and :211), inside the 6-line `SENTINEL_LOOKBACK_LINES` window, so both assertions pass as written. Importing `isAgentIngressRoute` from `../src/server` in a root test follows test/agent-ingress.test.ts:5-10, and bunfig.toml:6 preloads the env-stripping setup so the `src/config` import stays deterministic. Pre-existing skill assertions still hold: the file still carries exactly one ```epic-dag fence (SKILL.md:38) with members 12/13/14 and the literal `- [ ] #12` (SKILL.md:45), which is what test/epic-parse.test.ts:89-105 pins.

**Internal consistency of the rewritten skills.** No surviving text tells the agent to perform import: epic-authoring's frontmatter (line 3), intro (line 24), draft-only note (line 61), Stage 0 (line 86), Stage 4, Stage 5's third bullet, and both Gate-rules bullets (lines 211-215, 219-220) all agree; the same holds for onboarding at lines 3, 29, 52-55, 195-229, 236-238, 247-251, 261-267. The one endpoint a skill still tells the agent to call itself is `PUT …/api/sessions/<id>/epic-draft` (epic-authoring SKILL.md:58), which *is* on the allowlist (src/server.ts:7681). Acceptance criteria from the issue are met: the unguided path now terminates (marker verified, import handed over with a reason), and Stage 4/5 plus the Gate-rules sections agree with what an agent can reach.

No blocking findings.

Nits (non-blocking):
- `.claude/skills/shepherd-epic-authoring/SKILL.md:165` and `.claude/skills/shepherd-onboarding/SKILL.md:209` tell the operator to press **Import structure**, which is the EN label only — a DE operator sees "Struktur importieren" (ui/messages/de.json:1211); naming it as "the import button on the epic panel" would survive locale.
- `.claude/skills/shepherd-epic-authoring/SKILL.md:164` "one click, always available" slightly overstates it: the button renders only while `epic.source === "markdown"` (EpicPanel.svelte:102) and the row has to be expanded first, so it's a few clicks and it disappears once import has wired native links.
- `.claude/skills/shepherd-epic-authoring/SKILL.md:173` / `.claude/skills/shepherd-onboarding/SKILL.md:216` print `Bearer $SHEPHERD_TOKEN`, which expands in the *operator's* shell, while the caveat below talks about the token being configured *on the server* — an operator whose server has a token but whose shell doesn't export it gets an empty bearer and a puzzling 401.

- `test/skill-agent-reachability.test.ts:32` — the extra `api/sessions/abc/epic-import` assertion restates coverage already exhaustive in test/agent-ingress.test.ts:32-57; harmless, but it isn't what binds the skill text to the boundary.

- `docs/superpowers/plans/2026-06-12-epic-runner.md:887,947` still shows the unauthenticated `POST /api/epic/import` curl; it's a historical plan document, not agent-facing guidance, so leaving it is defensible.

Issue #499, opened 2026-06-10
Trim per-turn context overhead (skill/tool catalog) for auto-spawned task agents


scoop:
Source: token-usage analysis (PR #496, `docs/token-usage-analysis.md` + `scripts/usage-report.ts`).

**Finding.** Every task-agent turn re-reads a fixed ~45–46K-token prefix; ~16K of it is addressable injected catalog: `skill_listing` ~9.3K + `deferred_tools_delta` ~2.8K + superpowers SessionStart hook ~3.4K. On a long task (TASK-288, 337 turns) that's ~5.4M cache-read tokens (~10% of the task's cost-proxy) spent re-reading a skill/tool menu the task never uses — the audit found **~0 Skill invocations and ~0 MCP calls** across the sampled tasks.

**Lever.** `service.ts:463` (`buildSpawnArgv`) launches auto-spawned (drain) and interactive sessions with the *same* argv — none pass `--safe-mode`/`--disable-slash-commands`. The read-only critic already strips all of this (`reviewer-argv.ts`), proving it's mechanically easy.

**Decision needed (why this isn't a trivial PR).** Stripping the catalog also removes **superpowers skills + MCP** — the methodology (brainstorming, subagent-driven, TDD) depends on them. So a blanket trim trades capability for tokens. Options to weigh:
1. Apply the trim only to `input.auto` (unattended drain) spawns, IF those agents don't need the skill methodology.
2. Per-project plugin/MCP scoping so Shepherd agents load only a relevant subset (vercel/resend/account-connectors are unused here but live globally for other projects).
3. Leave interactive untouched; document the per-turn cost.

Needs a product call on (1) before implementing.

PR #2011, opened 2026-08-03
feat(spawn): keep progressive disclosure in drain sessions (#2001)

### LINKED ISSUE #2001: Re-decide the drain context trim so progressive disclosure is available
## Goal

Decide, against measured numbers, whether the drain context trim should keep disabling the skill catalog — because as it stands it **forecloses progressive disclosure for exactly the sessions that run unattended**.

The trim (issue #499) gives auto/drain spawns `--disable-slash-commands` plus an `enabledPlugins:false` overlay, and the code comment states plainly that the flag *"removes the entire skill catalog from the fixed prefix"*. It was a good trade when the alternative was paying for a catalog nothing invoked: it measured −6,349 tokens/turn.

The Claude 5 guidance inverts the framing. Progressive disclosure — skills and deferred tool loading — is the mechanism you *keep*, and the static text is what you delete. Right now Shepherd does the opposite for drain: it deletes the loading mechanism and keeps ~5.5 KB of resident notices. The onboarding skill even encodes the consequence, instructing operators never to reference skills or slash commands in a repo's agent instructions because drain sessions have both disabled.

## Vertical cut

Use the instrument from the measurement slice to price both sides for a drain spawn — catalog cost in the fixed prefix versus the resident-text cost it would let us delete — and change the trim decision to whatever the numbers support. If skills are re-enabled for drain, the resident notices become deletable in the next slice; if the catalog is genuinely too expensive, an alternative disclosure channel (a single always-on pointer, or hook-delivered guidance) is chosen and recorded instead.

## Acceptance

- A decision recorded with the measured token delta on both sides, for a drain spawn, on the current Claude Code version — not the 2024-era numbers from #499.
- Whichever way it goes, the net per-turn delta versus today is reported.
- The existing opt-out env for the trim still works, and attended sessions are unaffected.
- If the catalog is re-enabled, the onboarding skill's "never reference skills" instruction is corrected in the same PR, since it exists only to describe this constraint.

Blocked by the prompt-size instrumentation slice — this decision is worthless without the measurement.

scoop:
Closes #2001. Part of the Claude 5 context-engineering epic (#2005).

## The decision

Drain spawns no longer pass `--disable-slash-commands`. They keep the Skill tool and the **session checkout's own** `.claude/skills/`, and disable only the catalogs an unattended coding run cannot use: every operator plugin (as before), Claude Code's bundled skills, and the operator's personal `~/.claude/skills`.

**Net ≈ +1,107 tokens/turn** versus the previous trim: +1,059 in Claude Code's resident prefix, plus ≈48 est. for the 192 characters this PR adds to the context-trim notice (it now has to say *which* skills are gone). Still a sixth of the +6,012 that simply dropping the flag would have cost.

## Measured, on the current CLI (not #499's numbers)

`scripts/measure-spawn-prefix.sh` (new, committed so later slices can re-price on each Claude Code bump): one `claude -p` round-trip per variant in a Shepherd worktree, prefix = `input + cache_creation + cache_read`. Claude Code 2.1.220, n=3 per variant — every repetition returned the identical figure.

| Variant | Prefix tokens | Δ vs previous trim |
| --- | --- | --- |
| previous trim — `--disable-slash-commands` + plugins off | 34,175 | — |
| **shipped** — plugins off + `disableBundledSkills` + user skills off | **35,234** | **+1,059** |
| naive re-enable — drop the flag, plugins still off | 40,187 | +6,012 |
| no trim at all | 46,437 | +12,262 |

What the flag's 6,012 tokens bought: bundled skills ≈1,790 · the operator's personal skills ≈3,163 · the Skill tool plus this repo's 3 skills ≈1,059. In an empty directory the same comparison is 20,743 → 21,366, so ≈623 tok is the mechanism itself.

**Other side of the ledger** (#1999 instrument, drain-shaped payload as this PR leaves it = 9,968 chars / 2,492 est. tok): the notices the epic marks deletable — `single-pr-invariant`, `tmpfs-worktree-notice`, `manual-steps-notice`, `worktree-stash-notice`, `context-trim-notice` — are 5,742 chars ≈ **1,438 est. tok**. Nothing here deletes them; the next slice can, now that there is a mechanism to move them behind. Prefix figures are API-reported; block figures are the chars/4 estimate.

Functional check under the shipped overlay: the model lists exactly `merge-train, shepherd-epic-authoring, shepherd-onboarding` — the checkout's own skills, nothing else.

Full rationale, rejected alternatives and caveats: `docs/research/drain-context-trim-2026-08-03.md`.

## Notable implementation points

- `skillOverrides` is keyed by skill **name, not source**, so an `"off"` entry for a personal skill would also silence a project skill resolving to the same name. Both sides resolve that name the way Claude Code does — front-matter `name`, entry as fallback (`skillNameFrom`, now shared with the command listing, since entries under `~/.claude/skills` are commonly symlinks and front-matter may rename a skill).
- Those project names are enumerated in the **worktree** (the agent's cwd, and where Claude Code resolves project skills from) — a skill added on the branch exists only there. Non-isolated sessions are unaffected: their worktree path *is* the repo path.
- If the checkout's skills can't be listed, the overrides are dropped entirely rather than applied blind.

## Acceptance

- Decision recorded with measured deltas on both sides, on the current Claude Code version ✅
- Net per-turn delta reported (≈ +1,107) ✅
- `SHEPHERD_TRIM_AUTO_CONTEXT=false` still opts out (byte-identical untrimmed argv, asserted); attended spawns unaffected ✅
- The onboarding skill's "never reference skills" instruction corrected in the same PR ✅

## Verification

`bun test ./test` (8,145 pass / 0 fail), `bun run lint`, `bunx tsc --noEmit`, `fallow audit --base origin/main` clean on changed files. No UI changes, so no feature-catalog entry applies.


scoop:
## What I verified

**The trim itself does what the task asks.** `--disable-slash-commands` is gone from both task-spawn argv builders (`src/service.ts:2779` for create, `src/service.ts:2881-2891` for resume) and no `extraFlags` remnant survives anywhere — grepped the whole tree, the only remaining occurrences are the transient-agent paths the plan put out of scope (`src/transient-agent-argv.ts:256`, `src/prompt-recommend.ts:141`) plus docs. The replacement overlay is built at `src/service.ts:415-420` and keys are spelled identically in the committed pricer (`scripts/measure-spawn-prefix.sh:238-239` — `disableBundledSkills` / `skillOverrides: {…: "off"}`), so the two sides of the measurement agree.

**Opt-out and attended paths are untouched.** `trimDecision` short-circuits before any enumeration (`src/service.ts:534-536`), so `overlayOpts` stays `{}` and `spawnSettingsOverlay({...{}, hooks})` produces the same JSON as before — `disableBundledSkills`/`skillOverrides` are only added when the key is *present* (`src/service.ts:415`), pinned by test/service.test.ts:5946-5957 and 6218-6222.

**Identity resolution is genuinely shared.** `skillNameFrom` (`src/commands.ts:94`) routes through the same `resolvedName` the listing uses (`src/commands.ts:122`), and `readCommand`'s behaviour is byte-identical to before the refactor (`(fm.name?.trim() || fallbackName).trim()`). `readSkillNames` skip semantics match `collect`'s `existsSync(md)` guard (`src/commands.ts:161-166`). Blank/quoted/absent front-matter all resolve as the tests claim: `parseFrontmatter`'s regex (`src/commands.ts:70`) yields `""` for `name:   `, which falls back to the entry name.

**Cache contract mirrors the plugin one faithfully** — `userSkillNames` (`src/service.ts:352-359`) is the same `??=`-plus-clear-on-null shape as `installedPluginIds` (`src/service.ts:282-287`), and it is correctly repo-independent (the project-name subtraction happens outside the memo, in `disableableSkillNames`), so one repo's spawn can't poison another's.

**No new slash-command injection surface.** Removing the flag does re-enable command parsing, but `issueSpawnPrompt` (`src/issue-spawn-prompt.ts:48-52`) neutralises a leading `/` unconditionally for every issue-derived prompt, and it is unchanged by this PR.

**Acceptance criterion on the onboarding skill is met** — `.claude/skills/shepherd-onboarding/SKILL.md:114-119` no longer claims drain has skills disabled, and the two later echoes (lines 247-248, 258-260) were corrected too.

## Findings

### 1. The collision guard inspects a directory the agent never reads

`disableableSkillNames` subtracts the project skills of `join(repoPath, ".claude", "skills")` (`src/service.ts:565`). But the agent's cwd is the **worktree**, not `repoPath`: `herdr.start(ctx.name, ctx.worktreePath, …)` (`src/service.ts:2468`) against `start(name, cwd, argv, env)` (`src/herdr.ts:722-727`), and the worktree is a *sibling* of the repo — `join(dirname(repoPath), ".shepherd-worktrees", …)` (`src/worktree.ts:90-99`). So the set being compared is the operator's main checkout at whatever branch/dirty state it happens to be in, while the skills Claude Code will actually load come from the freshly created worktree at `baseRef`.

When those diverge — a skill that exists on the base branch but not in the operator's current checkout, or one renamed by front-matter on one side — the subtraction misses it, and a same-named personal skill's `"off"` override silently disables the repo's own skill. That is precisely the failure the guard's own doc comment says must not happen (`src/service.ts:554-556`). Reachable today, no future wiring needed. `worktreePath` is in scope at all three call sites: `wt.worktreePath` at `src/service.ts:3209` (the worktree is created at `src/service.ts:3299`, before `resolveCreateLaunch`), `session.worktreePath` at `src/service.ts:4103`, `s.worktreePath` at `src/service.ts:4311`. Note `worktree.create` returns `worktreePath === repoPath` for the non-isolated case (`src/worktree.ts:83`), so switching to `worktreePath` is correct on both branches.

### 2. The decision doc's resident-text side predates this PR's own reword

`docs/research/drain-context-trim-2026-08-03.md:120-131` reports the drain payload as 9,776 chars / 2,444 est. tokens with `context-trim-notice` at 427 chars / 107 tokens, stated as measured over `composeSystemPromptBlocks(null, true, {trimmed: true})`. Those are the pre-reword numbers. This PR grew `CONTEXT_TRIM_NOTICE` by exactly 192 chars — ground truth is the same PR's own assertion change, `test/prompt-budget.test.ts:161-163` (13,408 → 13,600 on a fixture whose only changed block is the notice). With `taggedBlock` adding 45 chars of wrapper (`src/service.ts:1464-1466`), the row is now 619 chars / 155 est. tokens, the "deletable subtotal" is 5,742 / 1,435, and the payload total is 9,968 chars.

This also feeds the headline: the doc reports **net +1,059 tokens/turn** (line 5 and again in `src/service.ts:518-521`), but the reworded notice rides in `--append-system-prompt` on every trimmed spawn, so this PR's real per-turn delta is ≈ +1,107 (+1,059 prefix, +48 est. from the longer notice). The task's acceptance is explicitly "the net per-turn delta versus today is reported", so the number needs the notice growth folded in (or the two components stated separately and both named).

## Nits (non-blocking):
- `src/service.ts:2666` — the untouched parenthetical "trimmed — the context-trim notice is Claude-specific (skill-catalog / slash-command / plugin trimming)" is now inaccurate about what the trim removes.
- `docs/research/drain-context-trim-2026-08-03.md:32` — the "no trim at all" row's Δ of +12,288 doesn't reconcile with the doc's own 34,175 baseline (46,448 − 34,175 = 12,273); +12,288 is the delta against the plan's earlier 34,160 pilot figure, so that one row looks carried over rather than re-measured with the others.
- `src/service.ts:672-679` — on the fail-safe path (`disableableSkillNames` returning `[]` because the project listing threw) the notice still tells the agent its personal skills are disabled when they are not; and in a repo that ships no `.claude/skills` it says "this repository's own skills … load normally — use them" while the spawn pays the doc's ~623 tok for an empty catalog.
- `scripts/measure-spawn-prefix.sh:212-228` — the `surgical` variant overrides every user skill without subtracting the repo's own names, so on a machine with a collision it prices a slightly different overlay than production builds.
- `src/service.ts:538-541` — the two enumerations are awaited sequentially inside the object literal; `Promise.all` would save one round trip on the first spawn of the process (both are memoized afterwards).

## Stated limitations
- Could not verify `disableBundledSkills` or `skillOverrides: {<name>: "off"}` against Claude Code's settings schema — it is not in this repo. I only verified the two in-repo users of those keys spell them identically (`src/service.ts:416-418`, `scripts/measure-spawn-prefix.sh:238-239`); the claim that an older CLI ignores unknown keys is likewise unverifiable here and taken as written.
- The prefix token figures (34,175 / 35,234 / 40,187 / 46,448) are machine measurements I cannot reproduce read-only; I checked only their internal arithmetic (see the nit above).

- `src/transient-agent-argv.ts:40` still documents `--disable-slash-commands` as "removes skills entirely" — correct for transient agents, which the plan deliberately left on the flag, but a reader arriving from #2001 may need the contrast spelled out.

scoop:
## Prior-round points

**1. Collision guard must enumerate the worktree, not the repo root — FIXED.** `trimDecision` now takes `worktreePath` and `disableableSkillNames` resolves `join(worktreePath, ".claude", "skills")` (src/service.ts:565-576). All three `trimFor` call sites pass it: `wt.worktreePath` (src/service.ts:3219), `session.worktreePath` (src/service.ts:4113), `s.worktreePath` (src/service.ts:4321). Verified this is genuinely the agent's cwd against `this.deps.herdr.start(ctx.name, ctx.worktreePath, …)` at src/service.ts:2478. Also verified the worktree exists by then: `this.deps.worktree.create(...)` runs at src/service.ts:3309, before `resolveCreateLaunch` at src/service.ts:3333, and `restoreExisting` at src/service.ts:4317 precedes the `trimFor` at 4321. `resolveCreateLaunch`'s `wt` param was widened to `{ isolated: boolean; worktreePath: string }` (src/service.ts:3183) and both callers supply it (src/service.ts:3333, 3662). Test pins it: `expect(skillDirs()).toEqual([join("/wt/repo-task", ".claude", "skills")])` (test/service.test.ts:6133).

**2. Stale resident-text figures in the decision doc — FIXED.** docs/research/drain-context-trim-2026-08-03.md:58-63 now carries `context-trim-notice` 619 / 155, subtotal 5,742 / 1,438, payload 9,968 chars / 2,492 est. tokens, and the net is restated as ≈ +1,107 (line 10-13) with the same figure mirrored in the code comment at src/service.ts:522-524. Arithmetic checked: the +192-char growth is proved by test/prompt-budget.test.ts:161 (13,408 → 13,600); 427+192 = 619; `estimateTokens` is `Math.ceil(len/4)` (src/prompt-budget.ts:60) so 619 → 155, a +48 delta on the old 107; 1,685+1,307+1,189+942+619 = 5,742 and 422+327+298+236+155 = 1,438; 9,776+192 = 9,968 and 2,444+48 = 2,492; 1,059+48 = 1,107.

## What I verified on the rest of the diff

- **No import cycle** from the new `import { skillNameFrom } from "./commands"` (src/service.ts:18): src/commands.ts:1-5 imports only node builtins and `./types`.
- **Shared identity rule.** `skillNameFrom` (src/commands.ts:94) and `readCommand` (src/commands.ts:122) both go through `resolvedName` (src/commands.ts:80), so the trim and the `/`-listing cannot drift. `parseFrontmatter` (src/commands.ts:63-74) handles the no-fence, unterminated-fence and quoted-scalar cases the test at test/service.test.ts:6000-6017 exercises.
- **Depth parity.** `readSkillNames` reads `<dir>/<entry>/SKILL.md` (src/service.ts:336), exactly the one-level shape `collect` scans at src/commands.ts:160-166. Entries that are files or lack a SKILL.md fall into the `catch { continue }` (src/service.ts:337-338), matching `collect`'s `existsSync(md)` skip.
- **Fail-safe direction.** `readSkillNames` returns `[]` only on ENOENT and `null` otherwise (src/service.ts:329-331); `disableableSkillNames` drops the whole override set when the project side is `null` (src/service.ts:573). `spawnSettingsOverlay` still emits `disableBundledSkills` in that case (src/service.ts:415-420) — documented at src/service.ts:394-396 and pinned by test/service.test.ts:6152-6163.
- **Untrimmed/attended byte-identity.** `trimDecision` returns `{ overlayOpts: {}, trimmed: false }` (src/service.ts:534) so both new keys are omitted; asserted at test/service.test.ts:6232 (`toBe(spawnSettingsOverlay())`) and test/service.test.ts:5955-5956. `extraFlags` is gone from the type and from both argv builders (src/service.ts:2789, 2891-2901); a repo-wide grep found no surviving reference to it.
- **Memo contract** mirrors the plugin one exactly: `userSkillNames` (src/service.ts:352-359) vs `installedPluginIds` (src/service.ts:282-287) — the `??=` assigns the promise before the `.then` runs, so the `null` branch really does clear the cache.
- **Transient agents untouched.** `--disable-slash-commands` still ships at src/transient-agent-argv.ts:256, per the plan's out-of-scope list; test/transient-agent-argv.test.ts:90 and test/review.test.ts:697 are unaffected.
- **Doc/text parity for the reworded trim.** README.md:292, docs-site/src/content/docs/reference/configuration.md:33, src/config.ts:723-728, .claude/skills/shepherd-onboarding/SKILL.md:114-119 and 248, .claude/skills/shepherd-epic-authoring/SKILL.md:21-23 all now describe the same shape (repo skills kept; bundled + personal + plugin skills off). A grep for `skill.*disabl` across src/, docs/, docs-site/, README.md and .claude/ turned up no remaining claim that drain has skills off, outside `docs/research/` and `docs/superpowers/` historical records.
- **Codex** is unaffected: `resolveCreateLaunch` computes `trim` but never threads it into `buildCodexSpawnArgv` (src/service.ts:3226-3237), same as before.

### Stated limitation

I could not verify that `disableBundledSkills` and `skillOverrides: {<name>: "off"}` are real Claude Code 2.1.220 settings keys with those exact spellings and value — the CLI's settings schema is not in this repo and I have no shell access to the installed binary. Assumed as written per the plan's claim; nothing in the repo contradicts it.

## Nits (non-blocking):

- src/service.ts:686 — `CONTEXT_TRIM_NOTICE` ends with "Slash commands are not typed in this session.", which reads like a leftover of the removed flag; since slash commands are no longer disabled, something like "no operator types slash commands here" would say what is actually meant.
- docs/research/drain-context-trim-2026-08-03.md:13 — "Still a sixth of the +6,012" is loose against the stated +1,107 (≈ a fifth).
- src/service.ts:540-543 — `disablePlugins: await pluginIds()` and `disableSkills: await disableableSkillNames(...)` are awaited sequentially in the object literal; a `Promise.all` would overlap the settings read with the two skill listings.

scoop (commented):
## Prior finding

**1. `docs/research/drain-context-trim-2026-08-03.md` — the `no trim at all` row was un-rescaled. ADDRESSED.** The row is now `46,437 | +12,262` (docs/research/drain-context-trim-2026-08-03.md:37). 46,437 − 34,175 = 12,262 ✓, so all four rows now share the 34,175 baseline: 35,234 − 34,175 = 1,059 (line 35) ✓, 40,187 − 34,175 = 6,012 (line 36) ✓. The commit `0c05343cb docs: re-measure the untrimmed variant against the n=3 baseline` is exactly that re-measure, so the "n=3 per variant" claim at line 28 is no longer contradicted.

## What I verified

**Number consistency across doc, code comment and tests.**
- The code comment quotes the same three prefix figures (src/service.ts:519-521: 34,175 / 40,187 / 35,234) as the doc table (docs/research/drain-context-trim-2026-08-03.md:34-37) — no drift between the two.
- Attribution sums: 1,790 + 3,163 + 1,059 = 6,012, matching the naive-re-enable delta (docs/research/drain-context-trim-2026-08-03.md:39-41). Empty-directory figure 21,366 − 20,743 = 623 ✓ (line 42).
- Notice growth: I counted the new `CONTEXT_TRIM_NOTICE` string (src/service.ts:679-686) at 574 chars; `taggedBlock` wraps it as `<context-trim-notice>\n…\n</context-trim-notice>` (src/service.ts:1472) = +45 → **619 block chars**, exactly the doc's table entry (line 55). The old notice counts to 382 + 45 = **427**, the doc's stated "427 before" (line 58). Δ = 192 chars, which is exactly the `prompt-budget` fixture bump 13,408 → 13,600 (test/prompt-budget.test.ts:161) and, at `Math.ceil(chars/4)` (src/prompt-budget.ts:60), the ≈48 est. tokens the net `+1,107` is built from (doc lines 10-12).
- Deletable-block table: 1,685+1,307+1,189+942+619 = 5,742 ✓ and 422+327+298+236+155 = 1,438 ✓ (lines 50-56), each token figure = ceil(chars/4).
- Functional-check claim (doc line 45) lists `merge-train, shepherd-epic-authoring, shepherd-onboarding`; those are exactly the three dirs under `.claude/skills/` in this repo.

**Settings keys resolve against the CLI actually installed.** `disableBundledSkills` (5 occurrences) and `skillOverrides` (8) both appear in the shipped Claude Code 2.1.220 bundle at `/home/patrick/.local/share/claude/versions/2.1.220`, and `skillOverrides` occurs within 400 chars of an `"off"` literal in all 8 matches — consistent with the `{<name>: "off"}` shape emitted at src/service.ts:418. (The binary is minified, so I could not read the schema's value enum directly; the measured 40,187 → 35,234 drop under the surgical overlay is the corroborating evidence that the bare-name keys take effect.)

**Identity rule is genuinely shared, no import cycle.** `skillNameFrom` (src/commands.ts:98) and the listing's `readCommand` (src/commands.ts:122) both go through the new `resolvedName` (src/commands.ts:79), so the trim and the UI cannot drift on quoting/blank rules — `unquote` and the fence handling in `parseFrontmatter` (src/commands.ts:53-73) are reused verbatim. `src/commands.ts` imports only node built-ins and `./types` (src/commands.ts:1-5), so the new `service.ts → commands.ts` edge (src/service.ts:18) introduces no cycle.

**Worktree, not repo root — and the path is populated at every call site.** `disableableSkillNames` enumerates `join(worktreePath, ".claude", "skills")` (src/service.ts:572). All four `trimFor` call sites pass a real worktree path: create at src/service.ts:3219 receives `wt` from `worktree.create` on src/service.ts:3309 (before the argv is built); the replacement path passes `s.worktreePath` (src/service.ts:3662); `resume` passes `session.worktreePath` on an already-live/husk session (src/service.ts:4113); `restore` passes it *after* `restoreExisting` recreates the worktree (src/service.ts:4317 then 4321). The doc's claim that non-isolated sessions collapse to the repo path is true — `WorktreeManager.create` returns `{ worktreePath: repoPath, isolated: false }` (src/worktree.ts:83). The agent's cwd really is that path (src/service.ts:2478, `herdr.start(ctx.name, ctx.worktreePath, …)`), so this is where Claude Code resolves project skills from. The test pins it: `skillDirs()` must equal `join("/wt/repo-task", ".claude", "skills")` while `repoPath` is `/repo` (test/service.test.ts:6135, worktree stub at test/service.test.ts:5414).

**Failure/edge behaviour.** `readSkillNames` returns `[]` only on ENOENT and `null` on any other listing error (src/service.ts:329-331); `disableableSkillNames` drops the whole override set on `null` (src/service.ts:573) so an unreadable repo-skills dir can never silently disable a repo skill — pinned by test/service.test.ts:6141-6167. `userSkillNames` mirrors `installedPluginIds` exactly, including clearing the cache inside the chained promise on `null` (src/service.ts:353-358 vs src/service.ts:283-286); the `??=` assignment is synchronous and the `.then` callback runs in a later microtask, so the self-clear cannot race the memoization.

**Untrimmed/attended paths stay byte-identical.** `trimDecision` still short-circuits to `{overlayOpts: {}, trimmed: false}` before touching any fs (src/service.ts:535-537), and `spawnSettingsOverlay` omits both new keys when `disableSkills` is absent (src/service.ts:415-420) — asserted at test/service.test.ts:6232 (`toBe(spawnSettingsOverlay())`) and test/service.test.ts:6203/6236 (neither the plugin nor the skill seam is consulted). The only two callers of `spawnSettingsOverlay` are the two argv builders (src/service.ts:2792, 2897); nothing outside `service.ts` consumes it, so the widened `opts` type breaks no consumer.

**`--disable-slash-commands` removal is scoped.** It survives where the plan said it should: `src/transient-agent-argv.ts:256` (critic/plan-gate/namer/distiller), `src/prompt-recommend.ts:141`, and their tests (test/transient-agent-argv.test.ts:90, test/plan-gate-prompt.test.ts:49, test/review.test.ts:697) are untouched. No `trim.extraFlags` reference remains anywhere in `src/` or `test/`. The Codex path still passes `trimmed: false` (src/service.ts:2847) and never receives the overlay.

**Doc parity.** The `SHEPHERD_TRIM_AUTO_CONTEXT` row is updated in both catalogs the repo maintains — README.md:292 and docs-site/src/content/docs/reference/configuration.md:33 — and `src/config.ts:723-727` agrees with both. The acceptance criterion on the onboarding skill is met with no stale echo left: every mention (.claude/skills/shepherd-onboarding/SKILL.md:114-119, :248, :258) now says "the repo's own skills load, others don't".

### Stated limitations

- Could not confirm the accepted *value* enum for `skillOverrides` (`"off"` vs any other spelling) — the 2.1.220 bundle is minified/binary and no schema for it exists in this repo. Both keys and a nearby `"off"` literal are present in the shipped bundle, and the measured deltas are consistent with the override taking effect; assumed correct as written.
- The prefix figures themselves are single-operator measurements I cannot reproduce read-only (running `claude` is out of bounds for this review). Their *internal* arithmetic is now consistent, which is what the prior finding was about.

### Nits (non-blocking):

- `.claude/skills/shepherd-epic-authoring/SKILL.md`: the front-matter `description` (line 3) still opens "Author a Shepherd-recognized epic **in an attended session**", which now reads at odds with the body's new "reachable from an unattended drain too" (lines 20-24) — the description is the part the model actually sees when deciding to invoke it.
- `scripts/measure-spawn-prefix.sh`: `user_skills_off`'s `re.search(r"^name\s*:\s*(.*)$", head, re.M)` scans the first 4096 chars unbounded by the `---` fence, so a skill with no front-matter `name` but a body line starting `name:` resolves differently from `skillNameFrom` (src/commands.ts:98), which only reads inside the fence. Harmless on this machine, but the script is committed as the re-pricer for later epic slices.
- `src/service.ts:2676`: the `composeDirectives` doc comment still describes the trim as "skill-catalog / slash-command / plugin trimming", the framing the rest of the file just moved away from.

Issue #1632, opened 2026-07-11
Steer agents off `git stash` in worktrees (shared refs/stash collisions)


scoop:
## Problem
Agents (and their subagents) keep reaching for `git stash` during verification inside Shepherd worktrees — e.g. to peek at the base state or compare test counts "before/after" a change. This is unsafe: **`refs/stash` is a single repo-wide stack shared across every worktree of the repo**, so concurrent Shepherd sessions collide — one session's `stash pop`/`drop` can grab or discard another session's entry, and `stash@{n}` indices race. There is no per-worktree isolation.

Observed again this week: a subagent in the PR #1631 session (#1610 follow-up work) ran `git stash`/`pop` to compare a test suite with and without its change while other sessions were active. No harm resulted that time, but it's a recurring, latent data-loss footgun.

## Ask
Prevent these collisions **via steering** — add a Shepherd house rule (a `Learning`) that reaches every session and steers agents off `git stash` in a worktree, toward a safe alternative.

## Why steering fits
House rules are injected into every agent's *system prompt* (`src/house-rules.ts`, `<shepherd-house-rules>` block). A `Learning` with **`scopeGlobs: []` is an "Always-rule"** — injected regardless of the session's target paths, and prioritized ahead of scope-matched rules in the budget plan (`planHouseRulesInjection`). That's exactly the reach a cross-cutting safety rule needs.

## Proposed rule (wording to refine)
> Never `git stash` in a Shepherd worktree — `refs/stash` is a single repo-wide stack shared across all worktrees, so concurrent sessions collide (one pop/drop can grab another session's entry). To compare against or inspect base state without mutating the working tree, use a read-only path instead: `git show <ref>:<path>`, `git diff <ref>`, or a throwaway `git worktree add`. If you must shelve local changes, use `git stash create` + `git stash store` (which write a commit object **without** touching the shared `refs/stash` stack) and restore explicitly — never the bare `git stash`/`stash pop`.

## Suggested implementation
1. Author the rule as an **Always-rule** (`scopeGlobs: []`) and **promote it to proven** so ranking/budget can't age it out of the injected block. Verify it appears in the `<shepherd-house-rules>` block of a fresh session.
2. Confirm the Always-rule set still fits the house-rules char budget (`HOUSE_RULES_OVERHEAD` + budget); if a critical-safety rule could still be budget-dropped, consider a guaranteed-inject tier for safety rules (secondary — only if budgeting can drop it).

[...]

PR #2012, opened 2026-08-03
feat(prompt): move situational notices out of the standing prompt

### LINKED ISSUE #2002: Move situational notices out of the standing prompt (hooks, skills, say-it-once)
## Goal

Empty the standing system prompt of situational text: deliver hazard warnings **at the moment of the hazard**, load conditional guidance **when it is relevant**, and state each instruction **once**. This is the epic's main deletion, and it carries the regression proof for the whole epic.

## What is resident today and why it shouldn't be

**Hazard notices — better as tool-time denials.** The worktree-stash notice (942 chars) and the tmpfs-worktree notice (1,303 chars) are permanently-resident warnings about actions most sessions never take. Shepherd already installs an HTTP hook fragment covering eight events; a `PreToolUse` hook can deny `git stash` / `git stash pop`, and a worktree-add or dependency install under tmpfs, with the explanation attached to the denial. A refusal delivered at the call site is strictly stronger than 2.2 KB of standing prose the model must hold and reconcile for the whole session.

**Conditional notices — better as skills.** Branch-rename (302), preview-hint (550, isolated sessions only), manual-steps (1,189), the single-PR invariant (1,685) and engineering-posture (1,643) are each relevant to a subset of sessions and resident in all of them.

**Duplicated statements — say it once.** The untrusted-content directive is stated in the system prompt (805 chars) *and* restated as a ~250-char in-band line inside **every** fence — issue body, every comment block, recap task prompt, and the clamped review prompts. The autopilot steers restate instructions the system prompt already gave.

Sequencing note: how much of the conditional set can become a skill depends on the drain-trim decision, hence the dependency.

## Vertical cut

Move the two hazard notices to `PreToolUse` hooks; move the conditional notices to progressive disclosure per the trim decision; collapse each duplicated instruction to a single home, leaving fences carrying only their label and nonce.

Cuts are conditional per model family where the older or other family still needs the text: the 80%-reduction result is specific to Claude 5, and Shepherd also spawns Codex and Haiku 4.5.

## Acceptance

- `git stash` and a worktree-add or dependency install under tmpfs are both denied at call time with the reason attached; neither notice remains in the composed prompt.
- A session that needs each moved notice demonstrably loads it.
- Each instruction has exactly one home; fences carry label + nonce only, and the existing untrusted-content tests confirm injection-defense behavior is unchanged.
- Per-family prompt composition is documented and covered by a test.
- **Regression proof:** one canonical attended task and one canonical drain task complete end-to-end on the trimmed prompt, and the measurement slice's meter reports the byte drop against the 8,389-char baseline.

Blocked by the instrumentation slice and the drain-trim decision.

scoop:
Closes #2002 (epic #2005).

Empties the standing system prompt of situational text: hazards are denied **at the hazard**, conditional guidance loads **when it is relevant**, and each instruction has **one home**.

**Attended Claude spawn: 8,389 → 2,148 chars (−74%).** Drain-shaped: 9,968 → 3,727. Measured with the #1999/#2008 instrument; full ledger in `docs/research/standing-prompt-trim-2026-08-03.md`.

## What moved where

| block | chars | now delivered by |
| --- | --: | --- |
| `worktree-stash-notice` | 942 | `PreToolUse` **denial** |
| `tmpfs-worktree-notice` | 1,307 | `PreToolUse` **denial** |
| `single-pr-invariant` | 1,685 | `shepherd-pull-requests` skill + injected context at `gh pr create` |
| `manual-steps-notice` | 1,189 | same skill + same injection |
| `preview-hint-notice` | 550 | `shepherd-preview` skill |
| `engineering-posture` (part) | ~1.0k | think-before-coding → plan-gate directive; detached-job hazard → guard injection |
| `branch-rename-notice` | 302 | resident only where a rename can actually land |
| per-fence untrusted caveat | ~250 × fences | stated once per prompt |

## Two decisions worth reviewing

**A local `command` hook, not the existing HTTP ingest hook** (the issue suggested HTTP). HTTP hooks are documented fail-open on timeout/non-2xx, and under the autonomous membrane `--clearenv` strips `SHEPHERD_TOKEN` so the restricted ingress 401s *by design* — every deny would evaporate for exactly the unattended sessions that need it. It also keeps a blocking round-trip off the single Bun loop that pumps the web terminal.

**Skills are model-invoked, so each has a deterministic backstop.** The PR rules are injected at `gh pr create` whether or not the skill was opened; the skill holds the full text. `engineering-posture` was therefore compressed in place, not moved — it is standing law, not situational.

## Reachability (what the plan review caught)

The drop decision is taken host-side. `buildMembraneFlags` binds no part of the Shepherd checkout and tmpfs's `$HOME`, so without a bind a sandboxed session would run with *neither* the notice nor the deny. `MembraneInputs.agentSupportPaths` now `--ro-bind-try`s the guard script and the skills dir (the `apiKeyHelperPath` precedent), populated on both the session and aux spawn paths, pinned in `test/tool-guard.test.ts`. The interpreter is the node binary the membrane already binds — deliberately not `process.execPath` (bun) under the tmpfs'd `$HOME`.

## Verified with real spawns, not only unit tests

- `git stash` and a tmpfs install denied with the reason quoted back — trusted **and** inside the bwrap membrane.
- The deny also covers **sub-agent** tool calls, which the retired notice could only ask the top-level agent to restate by hand.
- Both skills listed, invoked and quoted inside the membrane via `--add-dir`.
- `additionalContext` injects without auto-approving or blocking the call.
- **Regression proof:** one canonical attended task (2,429-char prompt) and one drain-shaped task (4,008-char prompt) ran end-to-end on the trimmed prompt — implement, test, commit, push. The drain run stops at the push because the fixture's `origin` is a local bare repo, not GitHub.

## Fallbacks are the design

Codex has neither mechanism and keeps every moved block; `SHEPHERD_TOOL_GUARD=0` restores them on Claude too. Per-family composition is documented in `composeSystemPromptBlocks` and pinned by a matrix test. Every guard rule fails open — an ordinary `bun install` in a worktree is untouched.

`src/rundown-core.ts` fences external issue/PR titles and carried no untrusted directive at all; it gains injection defence it did not have.

[no-feature-entry] — server-side prompt plumbing, no user-facing UX.

### DIFF
--- a/agent-skills/.claude/skills/shepherd-preview/SKILL.md
+++ b/agent-skills/.claude/skills/shepherd-preview/SKILL.md
@@ -1,0 +1,38 @@
+---
+name: shepherd-preview
+description: How to expose a long-running local dev/test server from a Shepherd worktree so the operator's live preview finds it, by writing the port to a `.shepherd-preview` file. Load when starting a dev server, test server, or any other long-running local preview in this session.
+---
+
+# Shepherd live preview
+
+Shepherd can proxy a dev server running in this worktree to the operator's browser (and, when
+enabled, onto the tailnet). It finds the server by port.
+
+## Pointing the preview at a specific port
+
+If you start a long-running dev server in this worktree and want Shepherd's live preview to target
+a specific port, write that port — a bare number, nothing else — to a file named
+`.shepherd-preview` in the repository root.
+
+Shepherd uses that value only when the port is actually listening; otherwise it auto-detects the
+port. So this is optional: skip it if you have no dev server, or if the default detection already
+targets the right port.
+
+## What Shepherd does and does not do
+
+- Shepherd **never starts or stops your dev server**. Starting it is on you; keep it in a managed
+  or background terminal, not as a blocking foreground command.
+- Shepherd owns tailnet exposure — do not run `tailscale serve` yourself.
+- The hint only applies to **isolated** sessions, which have their own worktree. A session sharing
+  the main repo directory has no dedicated worktree for the file to live in.
+
+## If no preview appears
+
+Verify, in this order:
+
+1. the dev server process is still running;
+2. `.shepherd-preview` contains the actual listening port (a bare number);
+3. the port is genuinely bound (e.g. `ss -ltnp | grep <port>`).
+
+Remember that a backgrounded server outlives the step that needed it: kill what you spawn in the
+same shell once you are done with it.
--- a/agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md
+++ b/agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md
@@ -1,0 +1,73 @@
+---
+name: shepherd-pull-requests
+description: How a Shepherd session opens a pull request - exactly one PR per session, what to do when the work is too large for one (promote to an epic, or ship a slice plus a follow-up issue), and how to declare manual operator steps so they survive merge. Load before running `gh pr create`, or as soon as a task looks like it needs more than one PR.
+---
+
+# Opening a pull request from a Shepherd session
+
+A Shepherd session tracks **one** pull request. This is the mechanism talking, not a style
+preference: PR detection, the critic, the merge train and the Owed lens are all keyed to a single
+PR per session.
+
+## Exactly one PR
+
+When you open a pull request, open exactly one — never a second, and never label work "PR 1 of N."
+This holds even when the task or its issue describes multiple parts, phases, or "Part A / Part B."
+
+If the work is genuinely too large for one cohesive PR, pick ONE of:
+
+- **(a) Promote to an epic.** Convert the issue into an epic — create one child issue per intended
+  PR and mark the parent body so Shepherd recognizes it (shape below) — open NO pull request
+  yourself, then STOP and tell the operator the epic is ready to drain. Shepherd drains each
+  sub-issue as its own session and its own PR, but that drain is operator-started: you cannot
+  trigger it yourself.
+- **(b) Ship one PR + file a follow-up.** Complete and open a single cohesive PR for the slice you
+  can finish, then `gh issue create` a follow-up issue capturing the remainder for a later agent,
+  and reference it from the PR body. This is the always-safe default.
+
+Never split the work across two PRs from this one session.
+
+## The epic shape (option (a))
+
+Shepherd recognizes an epic ONLY structurally — the parent issue's body must reference each child's
+REAL issue number. Create the child issues first (`gh issue create`), capture their numbers, then
+edit the parent body to add EITHER a fenced dag block, e.g.:
+
+```epic-dag
+#12
+#13 <- #12
+```
+
+(one `#<n>` line per child; `#<n> <- #<m>` when #n is blocked by #m), OR a task-list with one
+`- [ ] #12`-style line per child issue. This body marker is MANDATORY even when the children have
+no dependencies. NOT recognized as an epic: an `epic` label, an `[EPIC]` title prefix, or a prose
+checklist without `#<n>` issue references.
+
+## Manual operator steps
+
+Before you open the pull request, declare any MANUAL OPERATOR STEPS the change implies — work a
+human must do around merge/deploy that the diff itself cannot perform (flip a feature flag, set an
+env var, run a one-off backfill/migration, restart a worker, DNS cutover, seed a record). Shepherd
+parses these from the PR body and surfaces them on the Owed lens so they survive merge and teardown.
+
+Declare them in the PR body with EITHER carrier:
+
+- A fenced block — each `- [ ]` line is one step:
+
+```shepherd:manual-steps
+- [ ] Set the FEATURE_X env var in production
+- [ ] POST-MERGE: Run the data backfill once the PR is live
+```
+
+- Or column-0 `Manual-Step:` trailer lines (flush-left, outside any fence), e.g. a line reading
+  exactly `Manual-Step: Rotate the signing key`.
+
+Prefix a step with `POST-MERGE:` when it must happen AFTER the PR merges.
+
+**DEFAULT TO DECLARING NOTHING.** Most PRs need NO manual steps. Add a step ONLY for a real
+out-of-band action a human must take; if merging fully completes the change, OMIT the carrier
+entirely. NEVER invent steps to fill the block — a spurious step is worse than none. When in doubt,
+declare nothing.
+
+If you have already opened the PR when you realize a step is owed, add it with
+`gh pr edit --body` rather than opening a second PR.
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -147,6 +147,21 @@ Two independent stages, each an env override on a code default:
 | `SHEPHERD_HOOKS_INGEST` | `1` (on) | Inject observe-only lifecycle hooks into spawned agents (ingest route, ring buffer/logging, sub-agent roster fan-out). No status consumption; additive + fail-open. Set `0` to disable entirely (kill switch) |
 | `SHEPHERD_HOOKS_SIGNALS` | `0` (off) | Set `1` to forward matched hook events into the poller's signal pipeline. Meaningful only when `SHEPHERD_HOOKS_INGEST` is also on |
 
+## Tool guard (PreToolUse deny)
+
+Separate from the ingest hooks above: a **local** `PreToolUse` hook on the `Bash` tool that
+**denies** two hazards at the call site instead of warning about them in every agent's standing
+prompt — a bare `git stash` (the stash stack is shared across worktrees) and a worktree-add or
+dependency install under a tmpfs root. The refusal carries the explanation, so the agent learns why
+only when it matters. It runs as a local `command` hook (not the fail-open HTTP ingest transport)
+so the deny still holds for unattended, sandboxed sessions, and it is bound into the bwrap membrane
+so it exists inside the sandbox too. Claude spawns only — Codex spawns have no such mechanism and
+keep the equivalent prompt notices resident.
+
+| Variable | Default | Purpose |
+| --- | --- | --- |
+| `SHEPHERD_TOOL_GUARD` | `1` (on) | Inject the `PreToolUse` Bash guard into Claude spawns. Set `0` to disable (kill switch) — turning it off puts both hazard notices back into the composed system prompt, so no guidance is lost |
+
 ## Documentation automation (PR-gated doc agent)
 
 Opt-in, default-off. When enabled, a manual trigger (`POST /api/doc-agent?repo=<path>`)
--- a/docs/research/standing-prompt-trim-2026-08-03.md
+++ b/docs/research/standing-prompt-trim-2026-08-03.md
@@ -113,7 +113,9 @@ interpreter is the node binary the membrane already binds — deliberately not `
   moved.
 - **Every guard rule fails open.** Unparseable input, an unknown shape, a relative path with no
   usable cwd → no decision. It blocks only the two shapes it positively recognizes, and an ordinary
-  `bun install` in a worktree is untouched. The command splitter is quoting- and heredoc-aware for
+  `bun install` in a worktree is untouched. "On a tmpfs" is asked of the kernel (`statfs`), never
+  inferred from the path — notably NOT from `$TMPDIR`, which Shepherd deliberately points at a
+  DISK-backed dir (#1875) so worktrees and installs can land there safely. The command splitter is quoting- and heredoc-aware for
   the same reason: a naive split on `;`/newlines would read a hazard MENTIONED as data (a commit
   message, a PR body) as one INVOKED and hard-block a legitimate call — so an ambiguous command
   line (unterminated quote or heredoc) yields no segments, and therefore no decision.
--- a/docs/research/standing-prompt-trim-2026-08-03.md
+++ b/docs/research/standing-prompt-trim-2026-08-03.md
@@ -111,7 +111,10 @@ interpreter is the node binary the membrane already binds — deliberately not `
   moved.
 - **Every guard rule fails open.** Unparseable input, an unknown shape, a relative path with no
   usable cwd → no decision. It blocks only the two shapes it positively recognizes, and an ordinary
-  `bun install` in a worktree is untouched.
+  `bun install` in a worktree is untouched. The command splitter is quoting- and heredoc-aware for
+  the same reason: a naive split on `;`/newlines would read a hazard MENTIONED as data (a commit
+  message, a PR body) as one INVOKED and hard-block a legitimate call — so an ambiguous command
+  line (unterminated quote or heredoc) yields no segments, and therefore no decision.
 - **Aux prompts grew slightly.** The critic, plan-gate reviewer, recap, rundown, namer, classifier
   and prompt-recommend builders each now state `UNTRUSTED_CONTENT_DIRECTIVE` once. Multi-fence
   prompts (the critic's eight) shrink; a single-fence one (the namer) grows ~0.5 KB per transient
--- a/docs/research/standing-prompt-trim-2026-08-03.md
+++ b/docs/research/standing-prompt-trim-2026-08-03.md
@@ -38,7 +38,9 @@ tokens are the instrument's chars/4 estimate.
 The two fallback rows are the point of the design, not an oversight: a block leaves the prompt only
 where the mechanism replacing it exists. Codex has neither Claude Code hooks nor `--add-dir` skill
 loading, so it keeps every moved block (it still gains the posture trim and the fence dedup). The
-guard-off row is the same rule under `SHEPHERD_TOOL_GUARD=0`.
+guard-off row is the same rule under `SHEPHERD_TOOL_GUARD=0` — note that each block is gated on its
+OWN mechanism, so that switch restores the two hazard notices and the PR-time pair (which needs both
+mechanisms), but not the preview hint, which the `shepherd-preview` skill still covers.
 
 ### The other side of the ledger
 
@@ -128,6 +130,6 @@ interpreter is the node binary the membrane already binds — deliberately not `
   delivered"; a deny that does not fire degrades to pre-#1632 behaviour. Neither breaks a spawn.
 - **A deny is a hard block.** A false positive costs one wasted call, and the reason names the
   sanctioned alternative, so the agent self-corrects. `SHEPHERD_TOOL_GUARD=0` is the code-free
-  revert, and it restores the notices.
+  revert, and it restores every block that depends on the guard.
 - **Per-`Bash`-call process spawn.** The guard runs a short node script on every `Bash` call. It is a
   dependency-free single file and exits immediately when nothing matches.
--- a/docs/research/standing-prompt-trim-2026-08-03.md
+++ b/docs/research/standing-prompt-trim-2026-08-03.md
@@ -1,0 +1,130 @@
+# Standing prompt: situational notices moved out
+
+**Issue:** #2002 (epic #2005) · **Measured:** 2026-08-03 · **Claude Code:** as installed on the host
+
+**Result.** An attended Claude spawn's Shepherd-authored system prompt drops from **8,389 to 2,148
+characters (−74%)**; a drain-shaped spawn from 9,968 to 3,727. Nothing was deleted outright: each
+moved notice is delivered by a mechanism at the moment it is relevant, and stays resident wherever
+that mechanism does not reach.
+
+## What moved, and to what
+
+| block                        |         chars | delivered instead by                                                                                                |
+| ---------------------------- | ------------: | ------------------------------------------------------------------------------------------------------------------- |
+| `worktree-stash-notice`      |           942 | `PreToolUse` **denial** of a shared-stack `git stash`                                                               |
+| `tmpfs-worktree-notice`      |         1,307 | `PreToolUse` **denial** of a tmpfs worktree-add / dependency install                                                |
+| `single-pr-invariant`        |         1,685 | `shepherd-pull-requests` skill + injected context at `gh pr create`                                                 |
+| `manual-steps-notice`        |         1,189 | same skill + same injection point                                                                                   |
+| `preview-hint-notice`        |           550 | `shepherd-preview` skill                                                                                            |
+| `engineering-posture` (part) |         ~1.0k | think-before-coding → the plan-gate directive; detached-job hazard → injected context on a backgrounded `Bash` call |
+| `branch-rename-notice`       |           302 | still resident, but only where a background rename can actually land                                                |
+| per-fence untrusted caveat   | ~250 × fences | stated once per prompt (`UNTRUSTED_CONTENT_DIRECTIVE`)                                                              |
+
+## Measurements
+
+Shepherd's own payload, via the #1999/#2008 instrument (`measurePromptBlocks`). Chars are exact;
+tokens are the instrument's chars/4 estimate.
+
+| spawn shape                     | before | after | est. tokens after |
+| ------------------------------- | -----: | ----: | ----------------: |
+| attended Claude                 |  8,389 | 2,148 |               538 |
+| attended Claude + autopilot     |  9,347 | 3,106 |               777 |
+| drain (trimmed, autopilot)      |  9,968 | 3,727 |               932 |
+| plan-gate interactive           | 13,171 | 7,130 |             1,783 |
+| research                        |  6,320 | 2,957 |               740 |
+| Codex attended                  |  6,860 | 5,750 |             1,439 |
+| Claude, `SHEPHERD_TOOL_GUARD=0` |  8,389 | 7,279 |             1,821 |
+
+The two fallback rows are the point of the design, not an oversight: a block leaves the prompt only
+where the mechanism replacing it exists. Codex has neither Claude Code hooks nor `--add-dir` skill
+loading, so it keeps every moved block (it still gains the posture trim and the fence dedup). The
+guard-off row is the same rule under `SHEPHERD_TOOL_GUARD=0`.
+
+### The other side of the ledger
+
+The skills are reached with `--add-dir`, and Claude Code makes every available skill's name +
+description resident in its own prefix. Measured on the fixture repo, `claude -p` reporting
+`input + cache_creation + cache_read`, n=2 per variant (identical both times):
+
+| variant                       | prefix tokens |
+| ----------------------------- | ------------: |
+| without `--add-dir`           |        32,761 |
+| with `--add-dir agent-skills` |        33,027 |
+
+**+266 tokens/turn** for the two skills' listing entries, against ≈ **−1,560 est. tokens/turn** of
+Shepherd payload on an attended spawn. Net ≈ −1,300 tokens/turn — but note the two figures are not
+the same kind of number: the −1,560 is the instrument's chars/4 estimate, the +266 is what the API
+reported. Both are machine- and CLI-version-specific; re-run rather than quoting these.
+
+## Verification
+
+Every mechanic was spiked with real `claude -p` runs before implementation, and re-verified after:
+
+1. **A `PreToolUse` command hook denies a call under `--dangerously-skip-permissions`** — which is
+   how Shepherd spawns every session. Both hazards were denied with the reason quoted back verbatim.
+2. **The deny also covers SUB-AGENT tool calls.** A dispatched sub-agent's bare `git stash` was
+   denied with the same reason. This is strictly stronger than the retired notice, which could only
+   ask the top-level agent to restate the constraint into each sub-agent's prompt by hand.
+3. **Both denials fire inside the bwrap membrane** (`standard` profile), not only in a trusted
+   spawn — see "Reachability" below.
+4. **`--add-dir <dir>` loads `<dir>/.claude/skills/`** into a session started in a different repo,
+   inside the membrane: the agent listed and invoked both Shepherd skills and quoted their bodies.
+5. **`additionalContext` with no `permissionDecision`** is injected alongside the call and the
+   command still executes — no auto-approval, no block.
+
+**Regression proof.** Two canonical tasks were run end-to-end on the trimmed prompt, using the real
+composed payload (`composeSystemPrompt` + `spawnSettingsOverlay` + `agentSkillsArgs`) in a throwaway
+repo:
+
+- _attended_ (2,429-char system prompt): add a helper + `node:test` coverage, run the tests, commit
+  on a branch. Completed: 5 tests passing, committed.
+- _drain-shaped_ (4,008-char system prompt: autopilot + trimmed + preview + branch-rename): add a
+  second helper + tests, run them, commit and push. Completed: 6 tests passing, pushed.
+
+The drain run stopped at the push because the fixture's `origin` is a local bare repo, so there is
+no GitHub for `gh pr create` to target; the PR-time context injection is verified separately (5).
+
+## Reachability is the load-bearing part
+
+The decision to drop a block is taken on the **host**, at prompt-composition time. If the mechanism
+then does not exist inside the sandbox, that session runs with neither the notice nor the deny.
+
+`buildMembraneFlags` binds `/usr`, `/etc`, `/opt`, a tmpfs `$HOME`, a tmpfs `/tmp`, the Claude config
+dir, the node toolchain, `~/.gitconfig`, `~/.config/gh` and the worktree — **no part of the Shepherd
+checkout**, and `$HOME` is tmpfs'd. So `MembraneInputs.agentSupportPaths` now carries the guard
+script and the agent-skills directory, each `--ro-bind-try`'d at its own path (the `apiKeyHelperPath`
+precedent), populated by both the session spawn path and `resolveSpawnMembrane`. The guard's
+interpreter is the node binary the membrane already binds — deliberately not `process.execPath`
+(bun), which lives under the tmpfs'd `$HOME`.
+
+## Design notes
+
+- **A command hook, not the existing HTTP ingest hook.** HTTP hooks are documented fail-open on
+  timeout/non-2xx, and under the autonomous membrane `--clearenv` strips `SHEPHERD_TOKEN` so the
+  restricted ingress 401s by design — every deny would evaporate for exactly the unattended sessions
+  that need it. A blocking HTTP hook on every `Bash` call would also put a synchronous round-trip on
+  the single Bun loop that pumps the live web terminal.
+- **Skills are model-invoked, so each one has a deterministic backstop.** The PR rules are injected
+  at `gh pr create` whether or not the skill was ever opened; the skill holds the full text.
+- **`engineering-posture` was compressed, not moved.** It is standing behavioural law, and
+  model-invoked disclosure cannot guarantee it loads. Only its two genuinely situational clauses
+  moved.
+- **Every guard rule fails open.** Unparseable input, an unknown shape, a relative path with no
+  usable cwd → no decision. It blocks only the two shapes it positively recognizes, and an ordinary
+  `bun install` in a worktree is untouched.
+- **Aux prompts grew slightly.** The critic, plan-gate reviewer, recap, rundown, namer, classifier
+  and prompt-recommend builders each now state `UNTRUSTED_CONTENT_DIRECTIVE` once. Multi-fence
+  prompts (the critic's eight) shrink; a single-fence one (the namer) grows ~0.5 KB per transient
+  spawn. `src/rundown-core.ts` fences external issue/PR titles and previously carried no directive
+  at all, so it gains injection defence it did not have.
+
+## Caveats
+
+- **CLI-version-specific.** `additionalContext` on `PreToolUse` and `--add-dir` skill discovery are
+  current-CLI behaviours. An older CLI ignoring `additionalContext` degrades to "guidance not
+  delivered"; a deny that does not fire degrades to pre-#1632 behaviour. Neither breaks a spawn.
+- **A deny is a hard block.** A false positive costs one wasted call, and the reason names the
+  sanctioned alternative, so the agent self-corrects. `SHEPHERD_TOOL_GUARD=0` is the code-free
+  revert, and it restores the notices.
+- **Per-`Bash`-call process spawn.** The guard runs a short node script on every `Bash` call. It is a
+  dependency-free single file and exits immediately when nothing matches.
--- a/scripts/tool-guard.d.mts
+++ b/scripts/tool-guard.d.mts
@@ -1,0 +1,46 @@
+// Types for the node-run PreToolUse tool guard (scripts/tool-guard.mjs). The guard stays plain
+// .mjs so Claude Code can invoke it with the node binary the sandbox membrane already binds; this
+// declaration only exists so the TypeScript test (test/tool-guard.test.ts) and the settings-overlay
+// caller can import it.
+
+/** A `PreToolUse` hook body, as much of it as the guard reads. */
+export interface ToolGuardEvent {
+  tool_name?: string;
+  tool_input?: { command?: string; run_in_background?: boolean; [k: string]: unknown };
+  cwd?: string;
+  [k: string]: unknown;
+}
+
+/** A denial carries the explanation the retired standing notice used to state. */
+export interface ToolGuardDenial {
+  permissionDecision: "deny";
+  permissionDecisionReason: string;
+}
+
+/** Guidance delivered at the call without changing its permission outcome — the deterministic
+ *  backstop for the notices that moved into skills. */
+export interface ToolGuardContext {
+  additionalContext: string;
+}
+
+export type ToolGuardDecision = ToolGuardDenial | ToolGuardContext;
+
+export const STASH_REASON: string;
+export const PR_CREATE_CONTEXT: string;
+export const BACKGROUND_CONTEXT: string;
+export function worktreeTmpfsReason(path: string): string;
+export function installTmpfsReason(path: string): string;
+
+/** Tmpfs roots (host mounts + `TMPDIR`/`TMP`/`TEMP` from `env`), de-duplicated. */
+export function tmpfsRoots(env?: Record<string, string | undefined>): string[];
+
+/** Decide one event: `null` = no opinion (normal permission flow). Pure. */
+export function decideToolGuard(
+  event: ToolGuardEvent | null | undefined,
+  env?: Record<string, string | undefined>,
+): ToolGuardDecision | null;
+
+/** Wrap a decision in the `hookSpecificOutput` envelope, or `null` for silence. */
+export function hookOutput(
+  decision: ToolGuardDecision | null,
+): { hookSpecificOutput: { hookEventName: "PreToolUse" } & ToolGuardDecision } | null;
--- a/scripts/tool-guard.d.mts
+++ b/scripts/tool-guard.d.mts
@@ -31,13 +31,19 @@ export const BACKGROUND_CONTEXT: string;
 export function worktreeTmpfsReason(path: string): string;
 export function installTmpfsReason(path: string): string;
 
-/** Tmpfs roots (host mounts + `TMPDIR`/`TMP`/`TEMP` from `env`), de-duplicated. */
-export function tmpfsRoots(env?: Record<string, string | undefined>): string[];
+/** Is `path` really on a memory-backed filesystem? Asked of the kernel (statfs), never guessed from
+ *  the path — Shepherd points agents' `TMPDIR` at a DISK-backed dir on purpose (#1875). */
+export function isTmpfsPath(path: string, statfs?: (p: string) => { type: number }): boolean;
 
-/** Decide one event: `null` = no opinion (normal permission flow). Pure. */
+/** Injectable environment for the rules. `isTmpfs` defaults to {@link isTmpfsPath}. */
+export interface ToolGuardDeps {
+  isTmpfs?: (path: string) => boolean;
+}
+
+/** Decide one event: `null` = no opinion (normal permission flow). */
 export function decideToolGuard(
   event: ToolGuardEvent | null | undefined,
-  env?: Record<string, string | undefined>,
+  deps?: ToolGuardDeps,
 ): ToolGuardDecision | null;
 
 /** Wrap a decision in the `hookSpecificOutput` envelope, or `null` for silence. */
--- a/scripts/tool-guard.mjs
+++ b/scripts/tool-guard.mjs
@@ -16,6 +16,7 @@
 // a hazard MENTIONED as data (in a commit message, a PR body, a heredoc) must never read as one
 // INVOKED, and an ambiguous command line must produce no decision at all.
 
+import { realpathSync } from "node:fs";
 import { isAbsolute, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 
@@ -380,6 +381,30 @@ async function main() {
   if (out) process.stdout.write(JSON.stringify(out));
 }
 
-if (process.argv[1] && fileURLToPath(import.meta.url) === resolve(process.argv[1])) {
+/**
+ * Is this module the process entry point (i.e. run as the hook command, not imported)?
+ *
+ * `realpathSync(argv[1])`, not a bare `resolve` — Node realpath-resolves the ENTRY module behind
+ * `import.meta.url` (`--preserve-symlinks-main` defaults off) but leaves argv[1] merely resolved, so
+ * ANY symlinked path component makes a bare compare false. Getting this wrong is silent and severe
+ * here: `main()` would never run, the hook would emit nothing, and every `git stash` / tmpfs install
+ * would be ALLOWED — while composeSystemPromptBlocks has already dropped both hazard notices from
+ * the prompt on the strength of this guard existing. Verified: invoking this file through a
+ * symlinked directory emitted nothing before this fix. Same idiom (and same reasoning) as
+ * `isMainModule` in scripts/check-model-mirror.mjs; `fileURLToPath` rather than a `file://` concat
+ * because `import.meta.url` is percent-encoded.
+ */
+function isMainModule() {
+  const entry = process.argv[1];
+  if (!entry) return false;
+  try {
+    return realpathSync(entry) === fileURLToPath(import.meta.url);
+  } catch {
+    // argv[1] isn't a resolvable path (eval / bundler harness) → not the CLI entry.
+    return false;
+  }
+}
+
+if (isMainModule()) {
   await main();
 }
--- a/scripts/tool-guard.mjs
+++ b/scripts/tool-guard.mjs
@@ -12,6 +12,9 @@
 //
 // Every rule FAILS OPEN: anything unparseable, unknown, or merely suspicious returns no decision, so
 // the guard can never wedge a session. It only ever blocks the two shapes it positively recognizes.
+// That contract is why `segments()` below is quoting- and heredoc-aware rather than a plain split:
+// a hazard MENTIONED as data (in a commit message, a PR body, a heredoc) must never read as one
+// INVOKED, and an ambiguous command line must produce no decision at all.
 
 import { isAbsolute, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
@@ -107,13 +110,94 @@ function unquote(word) {
   return m ? m[2] : word;
 }
 
-/** Split a command line into pipeline/list segments. Deliberately naive — a quoted `&&` splits a
- *  segment early, which can only ever make a rule see LESS, never more (fail open). */
+/**
+ * Split a command line into the segments that are genuinely COMMAND POSITIONS: the separator scan
+ * runs only outside quotes and skips heredoc bodies entirely.
+ *
+ * Quoting awareness is load-bearing, not polish. A naive split would treat a hazard MENTIONED as
+ * data as a hazard INVOKED — `git commit -m "a; git stash pop"`, or a `gh pr create --body` whose
+ * text has a column-0 `git stash pop` line — and hard-block a legitimate call. That is the exact
+ * inverse of this module's fail-open contract, so the parse errs the other way instead: an
+ * ambiguous command (unterminated quote or unterminated heredoc) yields NO segments at all, and
+ * therefore no decision.
+ *
+ * Everything inside `$( … )` / backticks is left in its segment and still read as a command
+ * position, which is correct — that IS a command position.
+ */
 function segments(command) {
-  return command
-    .split(/\r?\n|&&|\|\||[;&|]/)
-    .map((s) => s.trim())
-    .filter(Boolean);
+  const out = [];
+  const lines = command.split(/\r?\n/);
+  let cur = "";
+  let quote = null; // "'" or '"' while inside a quoted string (may span lines)
+  let heredoc = null; // terminator of the heredoc body currently being skipped
+  let pending = null; // terminator queued by a `<<WORD` seen earlier on this line
+
+  const flush = () => {
+    const s = cur.trim();
+    if (s) out.push(s);
+    cur = "";
+  };
+
+  for (const line of lines) {
+    if (heredoc !== null) {
+      // Inside a heredoc body: pure data, never a command position.
+      if (line.trim() === heredoc) heredoc = null;
+      continue;
+    }
+    for (let i = 0; i < line.length; i++) {
+      const c = line[i];
+      if (quote === "'") {
+        if (c === "'") quote = null;
+        cur += c;
+        continue;
+      }
+      if (quote === '"') {
+        if (c === "\\" && i + 1 < line.length) {
+          cur += c + line[++i];
+          continue;
+        }
+        if (c === '"') quote = null;
+        cur += c;
+        continue;
+      }
+      if (c === "\\" && i + 1 < line.length) {
+        cur += c + line[++i]; // escaped char is data, never a separator
+        continue;
+      }
+      if (c === "'" || c === '"') {
+        quote = c;
+        cur += c;
+        continue;
+      }
+      // `<<WORD` / `<<-WORD` opens a heredoc whose body is data; `<<<` is a here-STRING (not one).
+      if (c === "<" && line[i + 1] === "<" && line[i + 2] !== "<") {
+        const rest = line.slice(i + 2);
+        const m = /^-?\s*(?:'([^']*)'|"([^"]*)"|([A-Za-z_][A-Za-z0-9_]*))/.exec(rest);
+        if (m) {
+          pending = m[1] ?? m[2] ?? m[3];
+          i += 1 + m[0].length;
+          continue;
+        }
+      }
+      if (c === ";" || c === "|" || c === "&") {
+        flush();
+        if (line[i + 1] === c) i++; // `&&` / `||`
+        continue;
+      }
+      cur += c;
+    }
+    // A newline outside quotes ends the command; inside a quote it is part of the string.
+    if (quote === null) flush();
+    else cur += "\n";
+    if (pending !== null) {
+      heredoc = pending;
+      pending = null;
+    }
+  }
+  // Ambiguous parse ⇒ no opinion, rather than a decision taken on a misread command line.
+  if (quote !== null || heredoc !== null) return [];
+  flush();
+  return out;
 }
 
 /** Tokenize a segment and drop the parts that never carry meaning for these rules: leading `sudo`,
--- a/scripts/tool-guard.mjs
+++ b/scripts/tool-guard.mjs
@@ -1,0 +1,301 @@
+#!/usr/bin/env node
+// Shepherd's PreToolUse tool guard (issue #2002).
+//
+// Two hazards used to ride EVERY spawn as ~2.2 KB of standing prompt text — the shared-`refs/stash`
+// data-loss footgun (#1632) and the tmpfs inode exhaustion that bricks a session (#1862). Both are
+// warnings about actions most sessions never take, so they are delivered here instead: as a
+// `PreToolUse` denial at the call site, with the explanation attached to the refusal.
+//
+// Plain `.mjs` on purpose (same pattern as scripts/json-union-merge.mjs + .d.mts): Claude Code runs
+// it as a `command` hook, and the interpreter is the node binary the sandbox membrane ALREADY binds
+// (`config.nodeBin`). A TS entrypoint would need a second interpreter bound into every membrane.
+//
+// Every rule FAILS OPEN: anything unparseable, unknown, or merely suspicious returns no decision, so
+// the guard can never wedge a session. It only ever blocks the two shapes it positively recognizes.
+
+import { isAbsolute, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
+
+/** Reason attached to a denied `git stash` — the content of the retired worktree-stash notice. */
+export const STASH_REASON =
+  "Blocked by Shepherd: `refs/stash` is a SINGLE stack shared across every worktree of this repo, " +
+  "so a bare `git stash` / `pop` / `drop` in one session can grab or discard another concurrent " +
+  "session's entry. To inspect or diff base state without touching the working tree, use " +
+  "`git show <ref>:<path>`, `git diff <ref>`, or a throwaway `git worktree add`. To shelve local " +
+  "changes, use `git stash create` — it prints a commit SHA WITHOUT writing the shared stack " +
+  "(tracked changes only; untracked files are not saved) — record that SHA and restore it later " +
+  "with `git stash apply <sha>`. Never `git stash store`: it writes the same shared stack.";
+
+/** Reason attached to a denied worktree-add under a tmpfs root (retired tmpfs notice, part 1). */
+export function worktreeTmpfsReason(path) {
+  return (
+    `Blocked by Shepherd: \`${path}\` is on a small tmpfs with a HARD INODE CAP, independent of ` +
+    "its byte size. A git worktree there exhausts the inode table, after which every write fails " +
+    "with ENOSPC while `df -h` still shows plenty of free space (`df -i` is what shows the real " +
+    "cause). Put the worktree on the SAME filesystem as the repo instead — never under the " +
+    "scratchpad, $TMPDIR, or /tmp."
+  );
+}
+
+/** Reason attached to a denied dependency install under a tmpfs root (retired notice, part 2). */
+export function installTmpfsReason(path) {
+  return (
+    `Blocked by Shepherd: this would install dependencies under \`${path}\`, which is on a small ` +
+    "tmpfs with a HARD INODE CAP. A package manager must keep its content-addressable store on the " +
+    "SAME filesystem as the install target so it can hardlink, so it forks a second store there — " +
+    "hundreds of thousands of tiny files that exhaust the tmpfs INODE table. Every subsequent write " +
+    "then fails with ENOSPC while `df -h` still shows free space (`df -i` shows the real cause). " +
+    "Run the install inside the repo worktree instead."
+  );
+}
+
+/**
+ * Injected (not denied) when the agent is about to open a pull request. This is the deterministic
+ * backstop for the `shepherd-pull-requests` skill: a skill only loads if the model reaches for it,
+ * whereas this fires on the exact call that matters, so the two invariants that used to sit in the
+ * standing prompt still bind even when the skill is never opened.
+ */
+export const PR_CREATE_CONTEXT =
+  "Shepherd: this session tracks exactly ONE pull request — open this one and no second one, even " +
+  "if the task describes multiple parts. Work too large for one PR? Promote the issue to an epic " +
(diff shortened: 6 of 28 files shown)


scoop:
## What I verified

**Mechanism reachability inside the membrane (the load-bearing claim).** `toolGuardInterpreter()` resolves `safeRealpath(config.nodeBin)` (`src/tool-guard-hook.ts:21`), which is byte-for-byte the value the session spawn puts in `MembraneInputs.nodeBinReal` (`src/service.ts:2515`); `nodeToolchainFlags` binds `dirname(nodeBinReal)` RO (`src/sandbox.ts:342,356`), so the interpreter genuinely exists inside the sandbox. The wrapper script and skills dir are added via `agentSupportPaths` (`src/service.ts:2556`) and emitted as `--ro-bind-try p p` at `src/sandbox.ts:572`. Ordering is correct: `--tmpfs <home>` sits in the base array at `src/sandbox.ts:490-493`, and the support binds are pushed at `:572`, i.e. AFTER — so a path under a tmpfs'd `$HOME` punches through rather than being shadowed. Both derive from `import.meta.dir` the same way the existing `egress-runner.sh` resolution does (`src/egress.ts:94-95`, `src/doc-agent.ts:35`).

**Argv shape.** `--add-dir` is emitted immediately after `claude` at `src/service.ts:2907-2913` (spawn) and `:3016-3022` (resume); the next token in both is `--dangerously-skip-permissions`, so the variadic flag cannot swallow the positional prompt.

**Fence ⇒ directive coverage.** I re-derived the call-site list rather than trusting the plan's: `fenceUntrusted` is called from `src/service.ts:1996,2171` (session prompt — directive rides as a standing block at `src/service.ts:1749`), `src/critic-core.ts:185,214,233,287,288,388,395`, `src/plan-gate.ts:256`, `src/recap-core.ts:115,138`, `src/rundown-core.ts:711`, `src/namer-llm.ts:43`, `src/autopilot-classify-core.ts:95,98`, `src/prompt-recommend.ts:96,99`. Every one of those files now emits `UNTRUSTED_CONTENT_DIRECTIVE` exactly once (`src/critic-core.ts:176,281`, `src/plan-gate.ts:211`, `src/recap-core.ts:112`, `src/rundown-core.ts:708`, `src/namer-llm.ts:40`, `src/autopilot-classify-core.ts:92`, `src/prompt-recommend.ts:93`). No fencing builder was missed. `src/herd-digest.ts` does not fence — confirmed by the same grep.

**Per-family composition.** `guard`/`skills` are both `claudeFamily && …` (`src/service.ts:1743-1745`), and the PR pair is dropped only on `guard && skills` (`src/service.ts:1770-1777`) while the hazards drop on `guard` alone (`src/service.ts:1637-1640`). That matches the doc-comment table at `src/service.ts:1687-1694` and the assertions in `test/prompt-budget.test.ts:307-352`. `spawnSettingsOverlay` is reached only from the two Claude argv builders (`src/service.ts:2916`, `:3024`), so a Codex spawn genuinely has no hook and correctly keeps every block.

**Branch-rename gate.** `src/service.ts:2855-2859` mirrors `scheduleRefine`'s own conditions at `src/service.ts:4112-4117`, and the `isHeuristicNameStrong` argument is the same string in both places (`session.prompt` is assigned from `input.prompt` at `src/service.ts:3476`).

**Skill identifiers.** Front-matter `name:` in both SKILL.md files resolves through `skillNameFrom` (`src/commands.ts:94`) to the literals in `AGENT_SKILL_NAMES` (`src/agent-skills.ts:25`), which are the same strings subtracted in `disableableSkillNames` (`src/service.ts:590`) — pinned by `test/agent-skills.test.ts:27`. The skill's `epic-dag` example round-trips through the real `parseEpicBody` (`test/agent-skills.test.ts:40`), and its manual-steps fence matches `FENCE_OPEN` / `TASK_LINE` in `src/manual-steps.ts:32,34`.

**Kill switch.** `parseKillSwitch` is `raw !== "0"` (`src/config.ts:438-440`), so `config.toolGuard` (`src/config.ts:679`) defaults on and `SHEPHERD_TOOL_GUARD=0` is the documented off value — consistent with the config comment and `docs/research/standing-prompt-trim-2026-08-03.md:127`.

## The defect

`segments()` (`scripts/tool-guard.mjs:112-117`) splits on `\r?\n`, `&&`, `||`, `;`, `&`, `|` with **no quoting or heredoc awareness**, and every resulting segment is then rule-checked independently in `denyFor` (`scripts/tool-guard.mjs:223`). Splitting on a newline does not make a rule "see less" — it manufactures *additional* segments whose heads are arbitrary text. Concretely, `git commit -m "a; git stash pop"` yields the segment `git stash pop`; `words()` → `["git","stash","pop"]`, `gitOperands` → `["stash","pop"]`, `stashTouchesSharedStack(["pop"])` → true → **hard deny** (`scripts/tool-guard.mjs:238-240`). The same happens for any heredoc body with a column-0 `git stash …` line — writing a doc, a commit message, or a `gh pr create --body "$(cat <<'EOF' …")` that quotes the command. This repo in particular documents that command in prose.

That contradicts the module's own stated invariant at `scripts/tool-guard.mjs:14` ("Every rule FAILS OPEN") and the doc's "Every guard rule fails open" (`docs/research/standing-prompt-trim-2026-08-03.md:112`), and it makes the comment at `scripts/tool-guard.mjs:110-111` factually wrong about what the code does. The existing prose test only covers mentions with no separator before them (`test/tool-guard.test.ts:64-68`), so it does not bind this case.

## Nits (non-blocking):
- `src/plan-gate.ts:211` inserts `UNTRUSTED_CONTENT_DIRECTIVE` with no blank line on either side, unlike every other builder (`src/critic-core.ts:174-178`, `src/recap-core.ts:110-114`), so it runs into the surrounding prose.
- `test/untrusted.test.ts:100-103` says the invariant is asserted "rather than a hand-kept list of builders", but the `prompts` array at `test/untrusted.test.ts:104-129` is exactly a hand-kept list — a new fencing builder still slips through.
- `installTmpfsReason` (`scripts/tool-guard.mjs:43`) asserts the path "is on a small tmpfs", but `tmpfsRoots` also admits `$TMPDIR` unconditionally (`scripts/tool-guard.mjs:88-95`), and Shepherd's scratchpad TMPDIR is an ordinary on-disk path on many hosts (see the fixture at `test/tool-guard.test.ts:21`).
- `agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md:57-60` is a second copy of the manual-steps carrier grammar that `src/service.ts:928-931` calls the source of truth, and — unlike the epic example — it is not pinned to `parseManualSteps`.

- `src/spawn-membrane.ts:114-117` binds the guard script and skills dir into aux spawns whose own comment states "nothing here depends on it today" (aux spawns carry no Shepherd `--settings` hooks and no `--add-dir`).

- `scripts/tool-guard.mjs:246` scans every operand after `worktree add`, so with a repo checked out under a tmpfs root (`cd /tmp/repo && git worktree add /home/u/wt main`) the *branch* operand resolves to `/tmp/repo/main` and denies a legitimate add, with a reason naming a path that is not the worktree target. Dormant unless a repo lives under `/tmp`.
- `preview-hint-notice` is dropped on `skills` alone (`src/service.ts:1560-1566`) with no deterministic backstop, unlike the PR pair which requires `guard && skills`. On a CLI that does not do `--add-dir` skill discovery the hint is silently lost rather than falling back — low impact, since the port is auto-detected.

## Stated limitations
- Claude Code's runtime honouring of `--settings`-supplied `PreToolUse` command hooks under `--dangerously-skip-permissions`, of `additionalContext` without `permissionDecision`, and of `--add-dir` skill discovery are not verifiable from this repo. The PR documents empirical spikes for all three (`docs/research/standing-prompt-trim-2026-08-03.md:61-73`); I assumed them as written and did not treat them as findings.
- The reported byte figures (2,148 / 3,106 / 7,130 / 2,957) are pinned by `test/prompt-budget.test.ts:135-139`, which CI runs; I did not execute anything.

scoop:
## Re-review of the prior round's point

**Prior finding 1 (`scripts/tool-guard.mjs` `segments()`): NOT addressed — re-raised verbatim.**

`git log --oneline 8afaa034..HEAD` shows only three commits (`e14224e0` feature, `17ba7642` docs sync, `8df21df4` docs manifest); no rework commit touched the guard. I compared the current file directly:

- `scripts/tool-guard.mjs:112-117` — `segments()` still splits on `/\r?\n|&&|\|\||[;&|]/` with no quoting or heredoc awareness.
- `scripts/tool-guard.mjs:110-111` — the comment still asserts the split "can only ever make a rule see LESS, never more (fail open)".
- `scripts/tool-guard.mjs:223-240` — every segment is still rule-checked independently, and a `stash` head still returns a hard `deny`.

Traced concretely against the code: `git commit -m "a; git stash pop"` → `segments()` yields `git stash pop"` → `words()` → `["git","stash","pop\""]` → `gitOperands` (`scripts/tool-guard.mjs:144-156`, nothing starts with `-`) → `["stash","pop\""]` → `stashTouchesSharedStack(["pop\""])` (`scripts/tool-guard.mjs:170-179`: `pop"` is not in `SAFE_STASH_SUBCOMMANDS`, not `apply`) → `true` → `deny` at `scripts/tool-guard.mjs:238-240`. The same holds for a newline-separated mention, which is the shape a `gh pr create --body "$(cat <<'EOF' … EOF)"` or a `git commit` heredoc produces — and `opensPullRequest` (`scripts/tool-guard.mjs:204-211`) confirms the guard expects exactly those multi-line `gh pr create` bodies.

The test file added in this round still only covers separator-free mentions (`test/tool-guard.test.ts:64-68`: `echo "never run git stash here"`, `gh pr comment -b 'do not git stash'`), so nothing in the suite would catch it. The doc claim at `docs/research/standing-prompt-trim-2026-08-03.md:112-114` ("Every guard rule fails open … It blocks only the two shapes it positively recognizes") is still unsupported for this case.

## New finding

**Per-family composition table, `preview-hint-notice` row.** `src/service.ts:1693` states `preview-hint-notice │ shepherd-preview skill │ RESIDENT (isolated) │ RESIDENT (isolated)` under the columns `Claude (guard+skills) │ Claude, guard off │ Codex`. The actual gate is `src/service.ts:1562`: `if (opts.previewHint && !skills)` — it depends on `skills` only, never on `guard` (`skills` is computed at `src/service.ts:1748`, `guard` at `src/service.ts:1747`). So on a Claude spawn with `SHEPHERD_TOOL_GUARD=0` and the skills dir present, the block is *dropped*, not resident. The three other Claude-guard-off cells are correct — `worktree-stash`/`tmpfs` gate on `!guard` (`src/service.ts:1634-1637`) and the PR pair on `!(guard && skills)` (`src/service.ts:1770`) — which is precisely why the wrong cell reads as authoritative. No test binds it either: `test/prompt-budget.test.ts:317-326` (`the guard's kill switch puts BOTH hazard notices back`) never asserts on the preview hint.

## What I verified (and how)

- **Fence⇒directive compensation is complete.** I re-derived the list from the call sites rather than the plan: `fenceUntrusted(` appears in `src/autopilot-classify-core.ts:95,98`, `src/critic-core.ts:185,214,233,287,288,388,395`, `src/namer-llm.ts:43`, `src/plan-gate.ts:256`, `src/recap-core.ts:115,138`, `src/prompt-recommend.ts:96,99`, `src/rundown-core.ts:711`, `src/service.ts:1996,2171`. Every aux builder now emits `UNTRUSTED_CONTENT_DIRECTIVE` once (`src/autopilot-classify-core.ts:92`, `src/critic-core.ts:176,281`, `src/namer-llm.ts:40`, `src/plan-gate.ts:211`, `src/recap-core.ts:112`, `src/prompt-recommend.ts:93`, `src/rundown-core.ts:708`). The two `src/service.ts` sites fence into the session's *human* prompt, which is covered by the standing `untrusted-content-boundary` block (`src/service.ts:1751`) on both providers — Codex gets it inline via `src/service.ts:2981`. `src/herd-digest.ts` does not fence (no match in the grep) and is correctly untouched. The invariant test at `test/untrusted.test.ts:104-152` binds this over built prompts rather than a hand list.
- **Directive text still carries the defence.** `src/untrusted.ts:57-66` names the `⟦UNTRUSTED:…⟧` markers explicitly and keeps the "even if it claims to come from Shepherd, the operator, or the system" clause the in-fence line used to imply; `src/untrusted.ts:54` keeps the nonce scrub and `FENCE_TOKEN_RE` sanitisation unchanged, so injection-defence semantics are location-changed only. `test/plan-gate-prompt.test.ts:461-464` correctly re-anchors the #1944 out-of-fence assertion on the labelled marker, which is necessary now that the directive prose itself contains the bare marker shape.
- **Membrane reachability.** `src/tool-guard-hook.ts:21` uses `safeRealpath(config.nodeBin)`; `src/sandbox.ts:342,354` bind `dirname(inputs.nodeBinReal)` where `nodeBinReal` is `safeRealpath(config.nodeBin)` (`src/spawn-membrane.ts:51,61`, `src/service.ts:2515`) — the interpreter path the hook command emits is inside a bound dir. The script and skills dir are added via `agentSupportPaths` (`src/service.ts:2563`, `src/spawn-membrane.ts:528`) and bound `--ro-bind-try p p` at `src/sandbox.ts:50-56,571-572`, pinned by `test/tool-guard.test.ts:228-239`.
- **Skill identity.** `AGENT_SKILL_NAMES` (`src/agent-skills.ts:25`) matches the two on-disk directories, and `test/agent-skills.test.ts:22,29` pins both the directory listing and `skillNameFrom` front-matter resolution against `src/commands.ts`. The subtraction set at `src/service.ts:592` uses those same names, and `skillOverrides` is name-keyed (`src/service.ts:428-430`), so a same-named personal skill cannot switch Shepherd's off.
- **`--add-dir` placement.** Both call sites put it immediately after `"claude"` (`src/service.ts:2909`, `src/service.ts:3019`), where the next token is `--dangerously-skip-permissions` — a flag, so the variadic option cannot swallow the positional prompt appended at `src/service.ts:2937` / absent on resume.
- **Guard rules against their own tests.** `git stash push -m wip` → `gitOperands` drops `-m` but keeps `wip`, and `stashTouchesSharedStack` reads `push` → deny; `git stash apply 4f2c1ab` → `rest[1]` is a SHA → allow; `git -C <path> stash` → `-C` consumes its value (`scripts/tool-guard.mjs:134-141`) → deny. `isBackgrounded` (`scripts/tool-guard.mjs:216`) correctly excludes a trailing `&&` via `(^|[^&])&\s*$`. The tmpfs roots include `/tmp`, `/var/tmp`, `/dev/shm` plus `TMPDIR`/`TMP`/`TEMP` (`scripts/tool-guard.mjs:88-95`), matching the scope the retired notice stated at `src/service.ts:669-672` ("under the scratchpad, `$TMPDIR`, or /tmp").
- **Kill-switch fail-safe.** `config.toolGuard` (`src/config.ts:679`) drives both the hook fragment (`src/tool-guard-hook.ts:40`) and the composition drop (`src/service.ts:1747`), so guard-off puts both hazard notices back — pinned at `test/prompt-budget.test.ts:317-326` and `test/service-plan-gate.test.ts:233-247`. `agentSkillsAvailable()` fails safe the same way for the skills half (`src/agent-skills.ts:36-38`, pinned `test/agent-skills.test.ts:55-60`).
- **Overlay merge.** `spawnSettingsOverlay` (`src/service.ts:432-436`) merges key-wise; `buildHooksFragment` owns the eight lifecycle events and the guard owns `PreToolUse` only (`src/tool-guard-hook.ts:44`), so no collision — pinned at `test/service.test.ts:1080-1093`. `spawnSettingsOverlay` has exactly two callers (`src/service.ts:2916`, `src/service.ts:3024`), both Claude, so no Codex spawn gains a hook.
- **Doc/keyword parity.** The new `SHEPHERD_TOOL_GUARD` row lands in `docs-site/src/content/docs/reference/configuration.md:150-162` and `ui/src/lib/docs-manifest.ts:72` gains the matching "tool guard (pretooluse deny)" keyword; no other catalog in the repo indexes that page.

## Cannot verify

- `additionalContext` on a `PreToolUse` hook without a `permissionDecision`, `--add-dir` skill discovery, and deny-under-`--dangerously-skip-permissions` are Claude Code CLI behaviours; the CLI is not in this repo, so I could not confirm them. Assumed as written per the PR's recorded spike.
- The before/after byte figures in `docs/research/standing-prompt-trim-2026-08-03.md:28-36` are asserted by `test/prompt-budget.test.ts:131-139`, which CI runs; I did not execute them.

## Nits (non-blocking):

- `src/agent-skills.ts:46-47` says the args "MUST be appended AFTER any positional prompt argument", but both call sites (`src/service.ts:2909`, `src/service.ts:3019`) put them first; the real constraint is "never immediately before a positional", and the two comments read as contradicting each other.
- `agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md:57-63` re-states the manual-steps carrier grammar as a hand copy of `MANUAL_STEPS_NOTICE` (`src/service.ts:934-952`) with no `parseManualSteps` pin, unlike the epic half which `test/agent-skills.test.ts:38-44` does pin against `parseEpicBody`.
- `src/plan-gate.ts:211` inserts the directive with no surrounding blank line, unlike the other six builders which pad it with `""` on both sides.

- `src/tool-guard-hook.ts:37-57` and `src/tool-guard-hook.ts:67-69` never probe that `scripts/tool-guard.mjs` exists, while the sibling mechanism does exactly that before dropping its blocks (`src/agent-skills.ts:36-38`, `existsSync` on the skills dir); both decisions drop prompt text host-side, so an install missing the script would silently lose the #1632/#1862 guidance where a missing skills dir would not.

scoop:
**Finding 1 (segments() quoting/heredoc)** — valid, and already fixed in `7734631a`, which landed before this review round ran (the review was against `8df21df4`). Reproduced all three shapes first (`git commit -m "a; git stash pop"`, a multi-line `--body`, and a heredoc), then replaced the splitter with one that scans for separators only outside quotes and skips heredoc bodies. An ambiguous line (unterminated quote or heredoc) now yields **no segments at all**, so a misread can never produce a deny. The stale `can only ever make a rule see LESS` comment and the doc's fail-open bullet are corrected, and tests cover mentions carrying `;`/newline/heredoc separators, real invocations behind those same separators, and the ambiguous-parse case.

**Finding 2 (preview-hint row)** — valid, fixed in `e10b708b`. The row is gated on `agentSkills`, not `toolGuard`, so a "Claude, guard off" column was the wrong axis for the whole table: the rows do not share a gate (hazards → guard, PR pair → guard AND skills, preview → skills). The matrix now names each row's own mechanism, and a new test in `test/prompt-budget.test.ts` pins both directions so the doc cannot drift from the code again.

scoop:

**Prior point 1 — `scripts/tool-guard.mjs` `segments()` splitting inside quotes/heredocs: FIXED.** `segments()` (scripts/tool-guard.mjs:127-201) now tracks `'`/`"` state, consumes backslash escapes, recognizes `<<WORD`/`<<-WORD` (excluding here-strings `<<<`) and skips the heredoc body until its terminator, and returns `[]` when the parse ends with an open quote or unterminated heredoc (scripts/tool-guard.mjs:198). I traced the exact shapes the finding named:
- `git commit -m "a; git stash pop"` — the `"` sets `quote` at scripts/tool-guard.mjs:167, the `;` is appended as data rather than reaching the separator branch at :182, so one segment survives whose head is `git commit`; `gitOperands` (:228) yields `commit` first, so `stashTouchesSharedStack` is never consulted.
- `gh pr create --body "$(cat <<'EOF' … EOF )"` — the `<<` sits inside the still-open double quote, so the heredoc branch at :173 is never reached and the whole thing stays one `gh` segment; `opensPullRequest` (:288) still matches, so the PR-time context is delivered and no deny occurs.
- Bare `cat <<EOF … EOF` — index arithmetic at :178 (`i += 1 + m[0].length`, plus the loop's `i++`) lands exactly past the terminator word, and the body lines are skipped at :142-146.
The fail-open contract at scripts/tool-guard.mjs:13-17 and the doc bullet at docs/research/standing-prompt-trim-2026-08-03.md:126-131 were both corrected to describe the new behaviour, and the requested separator-bearing-mention test exists at test/tool-guard.test.ts:70-84 alongside a genuine-invocation-behind-a-separator counter-test at :86-95 and an ambiguous-line test at :97-101.

**Prior point 2 — `src/service.ts` composition table cell: NOT FIXED, re-raised verbatim below.** The table still reads `preview-hint-notice │ shepherd-preview skill │ RESIDENT (isolated) │ RESIDENT (isolated)` at src/service.ts:1693, and the gate is still `if (opts.previewHint && !skills)` at src/service.ts:1562. `skills` is computed at src/service.ts:1745 as `claudeFamily && (opts.agentSkills ?? agentSkillsAvailable())` and `guard` separately at :1744; `trailingBlocks` is handed only `skills` (src/service.ts:1788, signature at :1546-1551), so with `SHEPHERD_TOOL_GUARD=0` and the skills dir present the block is dropped, contradicting the `Claude, guard off` cell. `git log --oneline -3 -- src/service.ts` shows the file's last touch is e14224e0, i.e. no revision landed for this point. The matrix test that would catch it passes no `previewHint` in the guard-off case (test/prompt-budget.test.ts:319-328), so nothing else binds the cell.

## What else I verified in this diff

- Identifier resolution: `safeRealpath` is exported at src/sandbox.ts:25; `config.nodeBin` at src/config.ts:568. `toolGuardInterpreter` (src/tool-guard-hook.ts:20-22) uses `safeRealpath(config.nodeBin)`, the same expression the membrane's `nodeBinReal` is built from at src/service.ts:2278/2290 and src/spawn-membrane.ts:51/61/109, so the interpreter the hook names is the one `nodeToolchainFlags` binds (src/sandbox.ts:342-354). `isHeuristicNameStrong` is imported at src/service.ts:47 and defined at src/namer.ts:468.
- The `branchRename` gate (src/service.ts:2855-2859) mirrors `scheduleRefine`'s early exits (src/service.ts:4112-4117) and reads `input.prompt`, which is what is stored as `session.prompt` (src/service.ts:3476) and later fed to `isHeuristicNameStrong(session.prompt)` — same value, no skew.
- `--add-dir` placement: emitted immediately after `claude` and before `--dangerously-skip-permissions` in both the spawn argv (src/service.ts:2906-2913) and the resume argv (src/service.ts:3014-3021), so the variadic flag is terminated by a `-`-prefixed token rather than swallowing the trailing positional prompt.
- `spawnSettingsOverlay` is called only from the two Claude argv builders (src/service.ts:2916, :3024) — no Codex path gets a `--settings` overlay carrying `PreToolUse`.
- Say-it-once: every `fenceUntrusted` call site outside the session spawn now emits `UNTRUSTED_CONTENT_DIRECTIVE` exactly once — src/critic-core.ts:176 and :281 (two builders), src/plan-gate.ts:211, src/recap-core.ts:112, src/rundown-core.ts:708, src/namer-llm.ts:40, src/autopilot-classify-core.ts:92, src/prompt-recommend.ts:93. The two in-spawn sites (src/service.ts:1996, :2171) are covered by the always-emitted `untrusted-content-boundary` block (src/service.ts:1749, unconditionally in the array at :1752-1754). I grepped for stale assertions on the removed in-band sentence and found none outside the updated test/untrusted.test.ts:47.
- Type surface: every symbol test/tool-guard.test.ts imports from `../scripts/tool-guard.mjs` (`decideToolGuard`, `hookOutput`, `tmpfsRoots`, `ToolGuardEvent`, `ToolGuardDenial`, `ToolGuardContext`) is declared in scripts/tool-guard.d.mts:7-46 with matching shapes.
- Docs manifest parity: the new `## Tool guard (PreToolUse deny)` heading in docs-site/src/content/docs/reference/configuration.md:150 has its lower-cased counterpart in ui/src/lib/docs-manifest.ts:72.
- Guard rule edges I walked by hand: `git -C <path> stash` still denies via `GIT_GLOBAL_VALUE_FLAGS` (scripts/tool-guard.mjs:218-240); `git stash apply <sha>` / `create` / `list` / `show` allow (:254-263); a package-manager `run`/`build` on tmpfs does not deny (:343-349); `isBackgrounded` distinguishes a trailing `&` from `&&` (:300).

## Stated limitations

- Could not verify that Claude Code accepts `hookSpecificOutput.additionalContext` on `PreToolUse` without a `permissionDecision`, that a `PreToolUse` deny overrides `--dangerously-skip-permissions`, or that `--add-dir` loads `<dir>/.claude/skills/` — those are CLI behaviours, not present in this repo. The PR documents them as spiked on the host (docs/research/standing-prompt-trim-2026-08-03.md:56-70); assumed as written.

## Nits (non-blocking):

- scripts/tool-guard.mjs:124-125: the comment claims content inside `$( … )`/backticks "is left in its segment and still read as a command position" — it is left in the segment, but the rules only inspect the segment's head (`words(segment)[0]`, e.g. :310, :289), so `echo $(git stash)` is never evaluated as a stash invocation. Fail-open and harmless, but the comment describes behaviour the code does not have.
- src/plan-gate.ts:210-211: the directive is spliced between two prose lines with no blank-line padding, unlike the other six builders which surround it with `""` (e.g. src/critic-core.ts:174-178, src/recap-core.ts:110-113).
- The plan pinned membrane reachability with tests in `test/sandbox.test.ts` and `test/spawn-membrane.test.ts`; what shipped covers `buildMembraneFlags` emitting the bind (test/tool-guard.test.ts:261-272) but nothing asserts that the spawn paths actually populate `agentSupportPaths` (src/service.ts:2556, src/spawn-membrane.ts:117) — the one link in the chain the plan's own risk section calls "the failure mode that matters".

scoop:
This round's single finding (the `preview-hint-notice` matrix cell) was already fixed in `e10b708b`, pushed before this review ran — the review targeted `7734631a`.

On the current head the `Claude, guard off` column no longer exists: the rows do not share a gate, so the table names each row's own mechanism instead (hazards → `guard`, PR pair → `guard AND skills`, preview-hint → `skills`), with an explicit note that `SHEPHERD_TOOL_GUARD=0` therefore restores the hazard and PR blocks but not the skill-covered preview hint. `test/prompt-budget.test.ts` pins both directions so the comment cannot drift from the gate again, and the research doc's two guard-off claims were corrected to match.

No further change in this push; CI is green on `e10b708b`.

scoop:
## Prior round

**Prior finding 1 (preview-hint matrix cell) — ADDRESSED.** The table was restructured so each row names its own mechanism rather than sharing a `guard off` column: `src/service.ts:1696` now reads `preview-hint-notice │ skills │ shepherd-preview skill │ RESIDENT (isolated)`, and the prose at `src/service.ts:1702-1704` spells out that `SHEPHERD_TOOL_GUARD=0` restores the hazard pair and the PR pair but *not* the preview hint. Compared against the gate at `src/service.ts:1562` (`if (opts.previewHint && !skills)`) and the derivations at `src/service.ts:1753-1754` (`guard`/`skills` are independent, both forced false for `codex`): the documented residency now matches the code in every cell I checked. `test/prompt-budget.test.ts:328-340` pins both directions.

## What I verified

- **Per-family composition matches the doc table.** `src/service.ts:1753-1754` vs the situational/PR/preview gates at `src/service.ts:1616-1621`, `src/service.ts:1774-1781`, `src/service.ts:1562`. Codex forces both mechanisms false via `claudeFamily`, so every moved block stays resident there — pinned by `test/prompt-budget.test.ts:344-357`.
- **Fence⇒directive compensation is complete.** Re-derived the call-site list with grep: `src/service.ts:2003`, `src/service.ts:2178` (both session human-prompts, whose system prompt carries `untrusted-content-boundary` unconditionally at `src/service.ts:1755`), plus `src/autopilot-classify-core.ts:95,98`, `src/prompt-recommend.ts:96,99`, `src/recap-core.ts:115,138`, `src/rundown-core.ts:711`, `src/critic-core.ts:185-395`, `src/plan-gate.ts:256`, `src/namer-llm.ts:43`. Every one of those aux builders now emits `UNTRUSTED_CONTENT_DIRECTIVE` once, and `test/untrusted.test.ts:103-152` binds the invariant over built prompts rather than a hand-kept list.
- **Skill payloads match their parsers.** `agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md:36-44` reproduces the `EPIC_SHAPE_CONTRACT` grammar from `src/service.ts:813-824` and is pinned to the real parser by `test/agent-skills.test.ts:37-43`; its manual-steps carriers (`shepherd:manual-steps` fence, column-0 `Manual-Step:`, `POST-MERGE:`) match `src/manual-steps.ts:33`, `src/manual-steps.ts:38` and `src/manual-steps.ts:40`. `PR_CREATE_CONTEXT` (`scripts/tool-guard.mjs:61-70`) names the same three carriers.
- **Membrane reachability.** `agentSupportPaths` is populated on the session spawn path at `src/service.ts:2557-2560` and in `src/spawn-membrane.ts:114`, and consumed by `agentSupportFlags` at `src/sandbox.ts:357-368`, emitted at `src/sandbox.ts:571-572`. The interpreter is `safeRealpath(config.nodeBin)` (`src/tool-guard-hook.ts:20-22`), which is the same value the spawn path feeds to `nodeBinReal` (`src/service.ts:2522`), so the binary `nodeToolchainFlags` binds (`src/sandbox.ts:342-355`) is the one the hook command invokes.
- **`branchRename` gate matches the thing it warns about.** `src/service.ts:2861-2866` (`isolated && config.llmNaming && !!this.deps.refineName && !isHeuristicNameStrong(...)`) mirrors `scheduleRefine` at `src/service.ts:4118-4124` plus the `session.isolated` requirement inside `refineNameInBackground` at `src/service.ts:4157`.
- **Argv ordering.** `--add-dir` is emitted first in both `buildSpawnArgv` (`src/service.ts:2914-2921`, prompt pushed last at `src/service.ts:2946`) and `buildClaudeResumeArgv` (`src/service.ts:3022-3029`), so the variadic flag is always followed by another flag.
- **`segments()` quoting/heredoc rework** (`scripts/tool-guard.mjs:127-201`): the heredoc advance arithmetic at line 178 (`i += 1 + m[0].length`, with the loop's `i++` making the net skip `2 + m[0].length`) is right, the ambiguous-parse early return at line 198 precedes the final `flush()`, and both directions are covered by `test/tool-guard.test.ts:70-101`.

### Stated limitation

I could not verify whether this host's Shepherd install path contains a symlinked component (the `readlink` call was denied), so the finding below is stated as a code-level inconsistency against the repo's own documented idiom, not as an observed failure on this machine.

## Finding

`scripts/tool-guard.mjs:383` decides whether it is the process entry with
`fileURLToPath(import.meta.url) === resolve(process.argv[1])`. `resolve()` is not `realpath()`. This repo already documents that exact comparison as broken, with regression tests, at `scripts/check-model-mirror.mjs:254-277`: *"Node realpath-resolves the ENTRY module behind `import.meta.url` (`--preserve-symlinks-main` defaults off) but leaves argv[1] merely `path.resolve`d, so any SYMLINKED path component makes a bare compare false."* That comment names `next-version.mjs` / `json-union-merge.mjs` as known-tolerated legacy — a new script reintroducing the form is a regression against the standard the repo settled on.

The consequence here is worse than for those build-time scripts. `TOOL_GUARD_SCRIPT` (`src/tool-guard-hook.ts:15`) is derived from `import.meta.dir` and handed to Claude Code verbatim as `argv[1]` (`src/tool-guard-hook.ts:50`). If that path carries a symlink component and Node resolves it, the compare is false, `main()` never runs, the hook writes nothing, and *every* `Bash` call is allowed — while `situationalBlocks` (`src/service.ts:1616-1621`) has already deleted `worktree-stash-notice` and `tmpfs-worktree-notice` from the prompt on the host. That is precisely the "neither the notice nor the deny" failure the plan calls the one that matters, and it is silent: no error, no non-zero exit, nothing in the transcript. Inside a bwrap membrane the `--ro-bind-try` makes the path its own mountpoint so it happens to be safe; the trusted profile has no such protection.

Use the normalized form (`realpathSync(entry) === fileURLToPath(import.meta.url)` inside a `try`/`catch`, as `isMainModule()` does) — and the plan's own testing seam ("one test feeds a JSON payload on stdin and asserts the exact `hookSpecificOutput` shape") is the coverage that would have caught it; `test/tool-guard.test.ts` only exercises the exported pure functions, never the CLI entry.

## Nits (non-blocking):

- `test/service.test.ts:4786-4789`: the assertion that used to check `<single-pr-invariant>` residency on an auto spawn was replaced by a verbatim duplicate of the `<epic-authoring-notice>` assertion two lines above, so the test now asserts the same thing twice and no longer covers what the surrounding comment says it covers.
- `agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md:32-44` duplicates the `EPIC_SHAPE_CONTRACT` prose from `src/service.ts:813-824` rather than reusing it (the plan proposed reuse). `test/agent-skills.test.ts:37-43` pins the *example* to the parser, which is the load-bearing half, but the surrounding wording can still drift from the constant.
- `src/tool-guard-hook.ts:20-22` realpaths the interpreter but `src/tool-guard-hook.ts:15` leaves the script path un-normalized; making them consistent would also sidestep the finding above.

scoop (commented):
## Prior finding — resolved

**1. `scripts/tool-guard.mjs` main-module check (silently skipped `main()` behind a symlink).**
Fixed. `scripts/tool-guard.mjs:397-406` is now the `isMainModule()` idiom: `realpathSync(entry) === fileURLToPath(import.meta.url)` inside a `try/catch` returning `false` on an unresolvable `argv[1]`. I compared it byte-for-byte against the form this repo documents and regression-tests, `scripts/check-model-mirror.mjs:269-278` — identical shape and identical failure reasoning. The requested CLI coverage also landed: `test/tool-guard.test.ts:287-360` spawns the script under `node` with the body on stdin and asserts (a) the deny envelope + exit 0, (b) silence + exit 0 on no-opinion, (c) fail-open on `"not json at all"` / `""` / `"null"`, (d) **the symlink regression itself** (`symlinkSync` into a `mkdtempSync` dir, then assert the deny still fires — `test/tool-guard.test.ts:331-344`), and (e) no output when imported rather than executed. Accepted; not re-raised.

## What I verified this round

- **Guard interpreter reachability inside the membrane.** `src/tool-guard-hook.ts:21` builds the hook command with `safeRealpath(config.nodeBin)`; that is the same value the membrane binds — `src/service.ts:2522` / `2547` pass `safeRealpath(config.nodeBin)` as `nodeBinReal`, and `nodeToolchainFlags` binds `dirname(nodeBinReal)` plus any manager root (`src/sandbox.ts:342-356`). So the interpreter resolves inside the sandbox. The wrapper script itself is bound separately: `src/tool-guard-hook.ts:66-68` → `src/service.ts:2563` / `src/spawn-membrane.ts:117` → `--ro-bind-try p p` at `src/sandbox.ts:360-371,572`. `safeRealpath` is genuinely exported (`src/sandbox.ts:25`).
- **Hook merge does not clobber ingest.** `src/service.ts:432-437` merges `buildToolGuardFragment()` (only key: `PreToolUse`, `src/tool-guard-hook.ts:45`) with `buildHooksFragment`, whose eight keys are `PostToolUse, PostToolUseFailure, Notification, SessionStart, Stop, SessionEnd, SubagentStart, SubagentStop` (`src/service.ts:489-500`) — no overlap. `spawnSettingsOverlay` is only reachable from the two Claude argv builders (`src/service.ts:2923`, `3031`), so Codex never gets the hook.
- **`--add-dir` placement.** `src/service.ts:2913-2919` and `3023-3028` put `...agentSkillsArgs()` immediately after `"claude"`, followed by `--dangerously-skip-permissions` — the next token is always a flag, so the variadic flag cannot swallow the positional prompt appended later.
- **Per-family composition matches the doc-comment table.** `src/service.ts:1749-1751` computes `claudeFamily`/`guard`/`skills`; hazards drop on `guard` alone (`situationalBlocks`, `src/service.ts:1630-1641`), the PR pair on `guard && skills` (`src/service.ts:1773-1780` → `codeSpawnBlocks`), the preview hint on `skills` alone (`src/service.ts:1561-1562`). That is exactly what the table at `src/service.ts:1692-1702` states, including the explicit "`SHEPHERD_TOOL_GUARD=0` restores the hazard pair and the PR-time pair but NOT the preview hint" note, and both directions are pinned in `test/prompt-budget.test.ts:325-345`. The earlier round's `preview-hint-notice` matrix-cell complaint no longer applies.
- **Codex keeps everything.** `agentProvider !== "codex"` gates both mechanisms, and Codex receives the composed blocks inline on the prompt (`src/service.ts:2974-2989`), so `untrusted-content-boundary` still reaches it after the per-fence caveat was dropped.
- **Fence ⇒ directive invariant.** Every `fenceUntrusted` call site outside `src/service.ts` lives in one of the seven builders that now emit `UNTRUSTED_CONTENT_DIRECTIVE` exactly once (`critic-core.ts:176,281`, `plan-gate.ts:211`, `recap-core.ts:112`, `rundown-core.ts:708`, `namer-llm.ts:40`, `autopilot-classify-core.ts:92`, `prompt-recommend.ts:93`); `src/critic-core.ts:388,395` sit inside `epicDeltaLines`, a helper reached only from `reviewPrompt`/`prReviewPrompt`, so no second emission. `src/service.ts:2003,2178` fence into the spawn prompt, which carries the directive as a standing block (`src/service.ts:1756`). `test/untrusted.test.ts:103-152` asserts "fences, and directive count === 1" per builder.
- **Skill bodies match the real parsers.** `agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md` uses ```` ```shepherd:manual-steps ```` and column-0 `Manual-Step:` — matching `src/manual-steps.ts:33` (`/^```shepherd:manual-steps[ \t]*$/i`) and `:38` (`/^Manual-Step:[ \t]*(.+)$/i`) — and its `epic-dag` example uses literal `#12` / `#13 <- #12`, which `src/epic-parse.ts:11-13` can parse; `test/agent-skills.test.ts:36-43` pins that with the real `parseEpicBody`. Skill front-matter names resolve through `skillNameFrom` to the `AGENT_SKILL_NAMES` the trim subtraction set uses (`src/service.ts:590-593`, `test/agent-skills.test.ts:25-28`).
- **`branchRename` gate.** `src/service.ts:2861-2866` mirrors `scheduleRefine`'s own preconditions (`src/service.ts:4118-4124`: `config.llmNaming`, `this.deps.refineName`, `!isHeuristicNameStrong(...)`), and `session.prompt` is set from `input.prompt` (`src/service.ts:3483`), so the two read the same text — the notice cannot be omitted for a spawn that will actually rename.
- **Guard rule behaviour, by reading `denyFor`/`segments`/`stashTouchesSharedStack`:** `git stash apply <sha>` and `list`/`show`/`create` are allowed; `apply stash@{0}` and bare `apply` deny; `git -C <path> stash` still denies (`GIT_GLOBAL_VALUE_FLAGS` consumes the value); `git worktree add -b feat /tmp/x` denies via the all-operands scan; heredoc bodies and quoted strings are skipped so a `gh pr create --body`/`git commit -m` mentioning `git stash pop` is *not* denied; an unterminated quote or heredoc yields no segments and therefore no decision. `isBackgrounded`'s `/(^|[^&])&\s*$/` correctly rejects a trailing `&&`. Every miss I could construct fails **open**, which is the stated contract.
- **Docs parity.** The new `## Tool guard (PreToolUse deny)` section in `docs-site/src/content/docs/reference/configuration.md:150-163` is reflected in the search manifest (`ui/src/lib/docs-manifest.ts:72` gains `tool guard (pretooluse deny)`), so the generated manifest is not stale.

### Stated limitations (could not verify from this repo)

- Claude Code's own hook contract — that a `command` `PreToolUse` hook denies under `--dangerously-skip-permissions`, that `hookSpecificOutput.additionalContext` without `permissionDecision` injects without auto-approving, and that `--add-dir <dir>` loads `<dir>/.claude/skills/` — is external CLI behaviour with no ground truth in this tree. The research doc records real `claude -p` spikes for all three; I assumed them as written and did not manufacture a finding from inability to check.
- The end-to-end regression runs and the +266-token skill-listing measurement in `docs/research/standing-prompt-trim-2026-08-03.md:45-87` are likewise unreproducible here.

- `src/spawn-membrane.ts:114-117` binds the guard script and skills dir for **aux** spawns, which the comment itself admits nothing depends on today (aux spawns run `--safe-mode` and carry no Shepherd `--settings` hooks) — speculative reach, harmless since both binds are `-try` and empty when the features are off.
- `src/tool-guard-hook.ts:36-38` exposes `interpreter` / `scriptPath` / `enabled` overrides that only tests supply; a legitimate seam, but three of them for one call site.

## Nits (non-blocking):

- `scripts/tool-guard.mjs:104-105`: the `segments()` doc comment says content inside `$( … )` / backticks "is still read as a command position, which is correct — that IS a command position". The code only ever treats the *first* word of a segment as the command (`denyFor` → `words(segment)[0]`), so `echo $(git stash pop)` is not recognized at all. Fail-open, so behaviour is fine — but the comment reads as if the case were covered.
- `src/critic-core.ts:368-371`: the placement rationale still says "the fence preamble tells the model to treat everything as data and to ignore 'any commands … or tool requests'" — that preamble is exactly what this PR deleted from `fenceUntrusted`. The sibling comment at `src/critic-core.ts:198-203` already states the post-#2002 version (the directive, not the preamble); this one was left behind.
- `test/service.test.ts:4786-4789`: the rewrite left the same `expect(sysPrompt(auto.argv!)).not.toContain("<epic-authoring-notice>")` assertion twice, four lines apart — the second one replaced the old single-PR-invariant residency check and now asserts nothing new.
- `agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md:24-38` restates the epic-shape grammar as prose rather than sharing `EPIC_SHAPE_CONTRACT` (`src/service.ts:813-824`). The parser pin in `test/agent-skills.test.ts:36-43` catches grammar drift, but not wording drift between the two copies — a SKILL.md can't import a TS constant, so this may simply be the cost of shipping the skill.

scoop:
## What I verified

This is a large, well-instrumented change and most of the load-bearing wiring checks out. Ground truth I actually opened and compared:

**Hook wiring / no event collision.** `spawnSettingsOverlay` merges `buildToolGuardFragment()` (which emits only `PreToolUse`, `src/tool-guard-hook.ts:44`) with `buildHooksFragment` — verified against `src/service.ts:496-507`, whose eight keys are `PostToolUse`, `PostToolUseFailure`, `Notification`, `SessionStart`, `Stop`, `SessionEnd`, `SubagentStart`, `SubagentStop`. No `PreToolUse` there, so the key-wise spread at `src/service.ts:441-445` cannot clobber either side. Both call sites (`src/service.ts:2980`, `src/service.ts:3097`) are Claude spawn + Claude resume; no Codex path passes `--settings`.

**Membrane reachability (the risk the plan calls load-bearing).** `--tmpfs $HOME` is emitted in the base array at `src/sandbox.ts:490-493`, and `agentSupportFlags` is pushed later at `src/sandbox.ts:571-572`, so the `--ro-bind-try` of the guard script and the skills dir punches through the tmpfs rather than being shadowed by it — same ordering discipline as the `codexCliFlags` comment at `src/sandbox.ts:555-556`. Interpreter: `toolGuardInterpreter()` returns `safeRealpath(config.nodeBin)` (`src/tool-guard-hook.ts:20-22`), and `nodeToolchainFlags` binds `dirname(inputs.nodeBinReal)` (`src/sandbox.ts:342`, `src/sandbox.ts:355`), so the realpath'd binary is inside. Populated on both the session path (`src/service.ts:2600`) and `resolveSpawnMembrane` (`src/spawn-membrane.ts:114`).

**Main-module check.** `isMainModule()` in `scripts/tool-guard.mjs:397-406` is byte-for-byte the idiom at `scripts/check-model-mirror.mjs:270-279`, and the symlink regression is pinned at `test/tool-guard.test.ts:330-343`.

**Say-it-once invariant.** I grepped every `fenceUntrusted(` call site in `src/` and each one's file also emits `UNTRUSTED_CONTENT_DIRECTIVE` exactly once: `src/critic-core.ts:176` and `:281` (covering the `epicDeltaLines` fences at `:388`/`:395`, which are reached only through `scopeAndOutputTail` from those two builders), `src/plan-gate.ts:211`, `src/recap-core.ts:112`, `src/rundown-core.ts:708`, `src/namer-llm.ts:40`, `src/autopilot-classify-core.ts:92`, `src/prompt-recommend.ts:93`. The two remaining sites, `src/service.ts:2040` and `src/service.ts:2215`, fence into the session HUMAN prompt and are covered by the standing `untrusted-content-boundary` block that `composeSystemPromptBlocks` emits unconditionally for every provider (`src/service.ts:1787`). `src/herd-digest.ts` does not fence. So the fence⇒directive compensation is complete, and `test/untrusted.test.ts:100-154` binds it by construction rather than by a hand-kept list.

**argv change.** `--add-dir` is placed where the next token is always a flag (`src/service.ts:2969-2976`, `src/service.ts:3087-3093`). The only positional argv indexing in `src/` is `src/herdr.ts:548`, which keys on `argv[0] === "bwrap"` and `indexOf("--")` — unaffected.

**Skill content parity.** The skill's epic grammar matches `EPIC_SHAPE_CONTRACT` (`src/service.ts:820-831`) including the literal `#12` / `#13 <- #12` / `- [ ] #12` examples the parser needs, and is pinned against the real parser at `test/agent-skills.test.ts:37-43`. `shepherd-preview/SKILL.md` covers everything `PREVIEW_HINT_NOTICE` (`src/service.ts:776-782`) said, plus the isolated-only caveat. `shepherd-pull-requests/SKILL.md` reproduces `MANUAL_STEPS_NOTICE` (`src/service.ts:946-964`) faithfully.

**Per-family matrix.** The doc-comment table at `src/service.ts:1729-1735` agrees with the code: `situationalBlocks` gates the hazards on `guard` alone (`src/service.ts:1670-1673`), `codeSpawnBlocks` on `!(guard && skills)` (`src/service.ts:1811-1819`), `trailingBlocks` gates the preview hint on `skills` alone (`src/service.ts:1599-1600`), and `claudeFamily` forces both false for Codex (`src/service.ts:1789-1791`).

**Docs parity.** `SHEPHERD_TOOL_GUARD` is documented in `docs-site/src/content/docs/reference/configuration.md:150-164` and the search keywords were regenerated at `ui/src/lib/docs-manifest.ts:72`. No i18n catalogs are touched by this diff (all new strings are agent-facing prompt text, matching the existing `BRANCH_RENAME_NOTICE` precedent).

## The blocking problem

See the finding. In short: `tmpfsRoots()` derives a deny-root from `$TMPDIR`, but Shepherd itself points every spawned agent's `TMPDIR` at a deliberately **disk-backed** directory, so the guard hard-blocks the exact operations that redirect exists to make safe — on the default configuration.

- `scripts/tool-guard.mjs:92-99` pushes `env.TMPDIR` (and `TMP`/`TEMP`) into the tmpfs root list, and `scripts/tool-guard.mjs:377` feeds it the hook process's real `process.env`.
- `src/tmp-sweep.ts:71-86`: `agentTmpDir()` is "the disk-backed `TMPDIR` that spawned (trusted) agents are pointed at (#1875), so ALL their temp I/O — the per-session scratch tree, **git worktrees, dependency installs**, and bare-`$TMPDIR` tool caches — lands on a real filesystem instead of the `/tmp` tmpfs". Default `~/.cache/shepherd/tmp`.
- `src/herdr.ts:529-532`: every herdr-started agent gets `TMPDIR=${agentTmp}` unless the operator sets `SHEPHERD_AGENT_TMPDIR=""`. Every session spawn goes through `buildWrappedArgv` (`src/herdr.ts:982`, `src/herdr.ts:1143`, `src/herdr-socket-driver.ts:207`, `:349`).
- `src/config.ts:428`: the default sandbox profile is `trusted`, i.e. no bwrap, so the `--clearenv` at `src/sandbox.ts:589` does *not* strip that `TMPDIR`. (Under `standard`/`autonomous` the redirect is wiped and the guard's `/tmp` root is correct — the defect is trusted-only, which is the default.)
- The behaviour is pinned as intended at `test/tool-guard.test.ts:24` (`ENV.TMPDIR` is literally `/home/u/.cache/shepherd/tmp/claude-1000/sess/scratchpad`), `:113` (worktree denied) and `:133` (`npm install` denied), so this is a design decision rather than a slip — but it contradicts #1875 and the reason strings at `scripts/tool-guard.mjs:36-40` / `:47-52` assert the path "is on a small tmpfs with a HARD INODE CAP", which is false for that path.

Net effect on a default install: an agent that follows Claude Code's own standing "use the scratchpad for ALL temporary file needs" instruction and runs `npm install` or `git worktree add` in its session scratchpad gets a hard refusal with a factually wrong explanation. That is the opposite of the module's stated "blocks only the two shapes it positively recognizes / ordinary installs are untouched" contract (`scripts/tool-guard.mjs:12-17`, plan risk section).

## Could not verify (not present in this repo)

- Whether `PreToolUse` `hookSpecificOutput.additionalContext` is honoured, whether `--add-dir` performs skill discovery, and whether a hook `matcher` is an exact match or a pattern — these are Claude Code CLI behaviours. The plan and `docs/research/standing-prompt-trim-2026-08-03.md:63-75` record empirical spikes; assumed as written.
- Whether `disableBundledSkills: true` (set for trimmed drain spawns at `src/service.ts:433-434`) also suppresses skills loaded from an `--add-dir` directory. If it does, drain sessions lose both the resident PR blocks and the skill, leaving only the `gh pr create` hook injection. That degrades gracefully by design, but it is worth an explicit check.

Nits (non-blocking):
- `src/service.ts:1670-1673` / `scripts/tool-guard.mjs:62-71`: `PR_CREATE_CONTEXT` fires on any `gh pr create`, including research / epic-authoring / landing-repair sessions, for which `single-pr-invariant` and `manual-steps-notice` were deliberately suppressed (`src/service.ts:1811` `if (!nonCodeMode)`); for an epic-authoring session it mildly contradicts `EPIC_AUTHORING_NOTICE`'s "open NO pull request".
- `scripts/tool-guard.mjs:125-126`: the comment says content inside `$( … )` / backticks is "still read as a command position" — the splitter does keep it in the segment, but only the segment HEAD is evaluated (`denyFor` reads `w[0]`), so `echo $(git stash)` is not recognized. Fail-open and fine, but the sentence reads as coverage the rule does not have.
- `src/tool-guard-hook.ts:35`: describes `matcher: "Bash"` as "an exact tool-name match"; Claude Code matchers are pattern-matched, so `BashOutput` would also spawn the hook. Harmless — `decideToolGuard` re-checks `tool_name` (`scripts/tool-guard.mjs:275`) — but the comment overstates.
- `src/tool-guard-hook.ts:37-57`: no existence probe for `TOOL_GUARD_SCRIPT`, asymmetric with `agentSkillsAvailable()`'s deliberate fail-safe (`src/agent-skills.ts:36-38`). A missing script drops both hazard notices *and* silently never denies.
- `src/plan-gate.ts:210-211`: the directive is spliced between two sentences with no blank line, unlike every other builder which brackets it with `""`.
- `test/service.test.ts:4784-4789`: the epic-intent test now asserts `not.toContain("<epic-authoring-notice>")` twice; the second assertion lost its original subject and no longer checks anything new.
- `test/prompt-budget.test.ts:128-142`: the re-pinned baselines call `composeSystemPrompt` without pinning `toolGuard`/`agentSkills`, so they depend on ambient `SHEPHERD_TOOL_GUARD` and on `agent-skills/` existing on disk — unlike the new MATRIX rows, which do pin.
- `scripts/tool-guard.mjs:314-318`: `cd` with no operand (`cd`), `cd -`, or `cd ~` leaves the tracked cwd unchanged rather than resetting it, so `cd && npm install` from a tmpfs cwd denies an install that would actually run in `$HOME`.

scoop:
**$TMPDIR deny-root** — valid, and the worst kind: the rule fought the very redirect it was meant to complement. Fixed in `0eb06f77`.

Verified on this host before fixing: `~/.cache/shepherd/tmp` is **btrfs** (statfs `0x9123683e`), `/tmp` is tmpfs (`0x01021994`) — and a `git worktree add` / `bun install` under the agent's own scratch tree was denied, with a reason falsely calling it "a small tmpfs". My tests pinned that as intended, so they were wrong too.

Rather than special-casing `agentTmpDir()` out of a path list, the rule now **asks the kernel**: statfs the target (for a not-yet-created worktree path, its nearest existing ancestor) and deny only tmpfs/ramfs. That is true by construction on any host layout — it also covers an operator who points `SHEPHERD_AGENT_TMPDIR` at a tmpfs, or a host where `/tmp` is disk-backed, neither of which a path list gets right. The probe is injected so the rules stay pure and no test depends on the host's mount table. Both denial reasons now name `$TMPDIR` as a **safe** destination instead of a forbidden one.

One adjacent, pre-existing thing I did **not** change: the retained `tmpfs-worktree-notice` (still resident for Codex and guard-off installs, `src/service.ts`) carries the same pre-#1875 wording — "never `git worktree add` under the scratchpad, `$TMPDIR`, or /tmp". That text predates this PR and is unmodified by it, so correcting it is out of this diff's scope, but it is now inconsistent with the redirect and worth a follow-up.

scoop:
## Prior round

**1. `$TMPDIR` treated as a tmpfs deny-root — FIXED, accepted.** The path list is gone: `isTmpfsPath` (`scripts/tool-guard.mjs:110-122`) now walks up to the nearest existing ancestor and asks `statfsSync`, denying only `TMPFS_MAGIC` / `RAMFS_MAGIC` (`scripts/tool-guard.mjs:93-94`), and it is injected through `deps.isTmpfs` (`scripts/tool-guard.mjs:299`) so no test depends on the host mount table. Both reason texts now name `$TMPDIR` as a **safe** destination (`scripts/tool-guard.mjs:39-41`, `:53-54`) instead of a forbidden one, which matches the #1875 redirect I cited last round (`src/tmp-sweep.ts:71-86`, `src/herdr.ts:529-532`). The tests that pinned the old behaviour were rewritten and the regression is now pinned the right way round (`test/tool-guard.test.ts:190-199`: worktree-add and installs under `~/.cache/shepherd/tmp/...` return `null`, `/tmp/wt` still denies). Not re-raised.

## What I verified (ground truth compared)

- **Identifier resolution.** `config.toolGuard` exists (`src/config.ts:681`, `parseKillSwitch`, default-on — same form as `hooksIngest` at `src/config.ts:669`) and is read at `src/service.ts:1790` and `src/tool-guard-hook.ts:40`. `agentSkillsArgs()` (`src/agent-skills.ts:48`) is spread at `src/service.ts:2973` (spawn) and `:3092` (resume), in both cases immediately before `--dangerously-skip-permissions`, so the variadic `--add-dir` is terminated by a flag rather than swallowing the positional prompt. `toolGuardMembranePaths()` / `agentSkillsMembranePaths()` feed `MembraneInputs.agentSupportPaths` (`src/sandbox.ts:191`) from `src/service.ts:2600` and `src/spawn-membrane.ts:117`, and that field is consumed at `src/sandbox.ts:572` via `agentSupportFlags` (`src/sandbox.ts:365-371`).
- **Membrane reachability is real, not just declared.** `--tmpfs $HOME` sits in the base flag array (`src/sandbox.ts:490-491`) while `agentSupportFlags` is pushed at `src/sandbox.ts:572`, i.e. after it — so an `--ro-bind-try` of a Shepherd checkout under `$HOME` punches through the tmpfs (same ordering rule the codex/`CODEX_HOME` comment at `src/sandbox.ts:553-554` states). The guard's interpreter is `safeRealpath(config.nodeBin)` (`src/tool-guard-hook.ts:21`), which is exactly the `nodeBinReal` whose `dirname` `nodeToolchainFlags` binds (`src/sandbox.ts:341-357`), so the comment at `src/tool-guard-hook.ts:64-65` ("the interpreter is already bound, so only the script is added here") checks out.
- **Hook merge does not clobber ingest.** `buildHooksFragment` emits exactly the eight lifecycle keys (`src/service.ts:490-507`) and none of them is `PreToolUse`, so the key-wise spread at `src/service.ts:441-444` is collision-free; `test/service.test.ts:1081-1094` pins all nine keys.
- **Fence ⇒ directive invariant.** I re-derived the call-site list rather than trusting the plan: `fenceUntrusted` is called from `src/service.ts:2040`, `:2215`, `src/autopilot-classify-core.ts:95,98`, `src/rundown-core.ts:711`, `src/plan-gate.ts:256`, `src/critic-core.ts:185,214,233,287,288,388,395`, `src/prompt-recommend.ts:96,99`, `src/namer-llm.ts:43`, `src/recap-core.ts:115,138`. Every aux builder now emits `UNTRUSTED_CONTENT_DIRECTIVE` once; the two `src/service.ts` sites are session **human**-turn prompts, and the corresponding spawn always carries `untrusted-content-boundary` unconditionally (`src/service.ts:1792`, `:1798-1800`) — including Codex — so no fenced content is left without the contract. `critic-core.ts:388/395` sit in `epicBlock`, reached only from `reviewPrompt`/`prReviewPrompt`, both of which emit the directive.
- **Skill wiring.** `AGENT_SKILL_NAMES` (`src/agent-skills.ts:25`) matches the two directories on disk and the `name:` front-matter of each (`agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md:2`, `agent-skills/.claude/skills/shepherd-preview/SKILL.md:2`), which is what `skillOverrides` is keyed on — so the subtraction at `src/service.ts:597-600` genuinely protects them. The skill's epic example (`shepherd-pull-requests/SKILL.md:36-44`) reproduces `EPIC_SHAPE_CONTRACT` (`src/service.ts:820-831`) with the literal `#12` / `#13 <- #12` numbers the parser requires, and `test/agent-skills.test.ts:36-42` pins it against `parseEpicBody`.
- **Composition matrix.** `situationalBlocks` / `codeSpawnBlocks` preserve the previous emission order and the pre-existing `#1257` Codex gate (`src/service.ts:1638-1650` vs the old inline form), and `claudeFamily` forces both mechanisms false for Codex (`src/service.ts:1789-1791`).

## Could not verify

- `additionalContext` (without `permissionDecision`) being honoured on a `PreToolUse` hook, and `--add-dir` loading `<dir>/.claude/skills/`, are Claude Code CLI behaviours whose ground truth is not in this repo. Assumed as written per the doc's spike record (`docs/research/standing-prompt-trim-2026-08-03.md:63-75`); I neither confirm nor dispute them.

## Findings

**1. `cd` with an unresolvable target keeps the *stale* cwd, so the install rule can hard-deny on a directory the command is not running in.** `scripts/tool-guard.mjs:332-337` does `cwd = next ?? cwd`, and `resolvePath` (`:261-267`) returns `null` for any `~`-prefixed target and for anything that resolves through an unexpanded variable/substitution. The `cd` did move the effective directory; the guard just doesn't know where — but it then judges the later install against the *old* cwd (`:365`). With a tmpfs cwd (an agent that `cd`'d into its scratchpad earlier — Claude Code persists cwd across `Bash` calls, and under the membrane `--clearenv` leaves the scratchpad on the bwrap `/tmp` tmpfs), `cd ~/repo && bun install` or `cd $REPO && npm ci` is DENIED with `installTmpfsReason(<old cwd>)`, naming a path the install would never have touched. That is a fail-**closed** outcome in a module whose stated contract (`scripts/tool-guard.mjs:12-17`) is that any parse doubt yields no decision. Dropping the claim instead (`cwd = next ?? undefined`) restores the contract.

**2. The `segments()` doc comment is factually wrong about command substitutions.** `scripts/tool-guard.mjs:141-143` states: "Everything inside `$( … )` / backticks is left in its segment and still read as a command position, which is correct — that IS a command position." The text is indeed left in the segment, but nothing reads it as a command position: every rule inspects only `words(segment)[0]` (`:329`, `:309`), and for `$(git stash pop)` that head is the literal `$(git`, so no rule matches. Worse, `foo=$(git stash pop)` is additionally eaten by the env-assignment strip at `:226-231`, leaving head `stash`. Missing these is consistent with the fail-open contract, but the comment claims coverage the guard does not have — in the one file that is now the *sole* delivery of the #1632 stash hazard, that is the comment a maintainer would reason from.

## Nits (non-blocking):

- `src/plan-gate.ts:210-212` inserts the multi-line `UNTRUSTED_CONTENT_DIRECTIVE` with no blank line on either side, so its last line runs straight into "A coding agent wrote the PLAN below…"; every other builder brackets it with `""`.
- `test/service.test.ts:4784-4790` replaced `expect(sysPrompt(auto.argv!)).toContain("<single-pr-invariant>")` with a second, identical `not.toContain("<epic-authoring-notice>")` assertion — the same expectation now runs twice and the residency half of the test is simply gone.
- `scripts/tool-guard.mjs:364` makes a bare `yarn` an "install", so `yarn --version` in a tmpfs cwd denies; harmless but avoidable by requiring an empty arg list rather than an empty *positional* list.
- `agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md:32-44` duplicates `EPIC_SHAPE_CONTRACT` as prose rather than sharing the constant (impossible in Markdown); `test/agent-skills.test.ts:36-42` pins the grammar, but the surrounding wording can still drift from `src/service.ts:820`.

- `src/spawn-membrane.ts:114-117` binds the guard script and skills dir into **aux** spawn membranes, which the code's own comment says nothing depends on today — a small, deliberate widening of the aux sandbox surface for a future that may not come.

- `scripts/tool-guard.mjs:301` injects `PR_CREATE_CONTEXT` on every `gh pr create` whenever the guard is on, including when `agentSkills` is false and the resident `single-pr-invariant` block is therefore also present — the injected text then both duplicates the block and points at a `shepherd-pull-requests` skill the session cannot load.
- `scripts/tool-guard.mjs:339-358` does not honour `git -C <path>` when resolving a `worktree add` operand (`gitOperands` at `:250-252` consumes the value but nothing feeds it back into `cwd`), so a relative worktree path under `git -C` is judged against the wrong directory.

- (none)

scoop (commented):
## Prior finding

**1. `scripts/tool-guard.mjs`: unresolvable `cd` target kept the previous cwd — ADDRESSED.**
Verified at `scripts/tool-guard.mjs:340-343`: `cwd = target ? (resolvePath(target, cwd) ?? undefined) : undefined;` — a `resolvePath` null (the `/^~|[$`*?]/` branch at `scripts/tool-guard.mjs:266`) now drops the cwd claim, and a bare `cd` (no non-flag operand) does the same. Downstream both rules abstain: the install rule is gated on `typeof cwd === "string" && isAbsolute(cwd)` (`scripts/tool-guard.mjs:372`), and `resolvePath` returns null for a relative operand with no cwd (`scripts/tool-guard.mjs:268`). Covered by `test/tool-guard.test.ts:151-164` (`cd ~/repo && bun install`, `cd $HOME/…`, `cd "$(…)"`, backticks, bare `cd`, all with a `/tmp/x` starting cwd → null) and `test/tool-guard.test.ts:171-173` still pins that a *resolvable* relative `cd sub` denies and names `/tmp/x/sub`.

## What I verified this round

- **Identifiers resolve.** `safeRealpath` (`src/sandbox.ts:25`), `config.nodeBin` (`src/config.ts:568`), `parseKillSwitch` (`src/config.ts:438-440`, so `SHEPHERD_TOOL_GUARD=0` is the only off value — matching `docs-site/src/content/docs/reference/configuration.md:158-162`), `skillNameFrom` (`src/commands.ts:94`), `parseEpicBody` (`src/epic-parse.ts:43`), `ComposeSystemPromptOptions`/`composeSystemPromptBlocks` (`src/service.ts:1682`, `src/service.ts:1773`, imported at `test/prompt-budget.test.ts:2-7`). `scripts/tool-guard.d.mts:28-52` declares exactly the exports `scripts/tool-guard.mjs` provides (`STASH_REASON`, `PR_CREATE_CONTEXT`, `BACKGROUND_CONTEXT`, `worktreeTmpfsReason`, `installTmpfsReason`, `isTmpfsPath`, `decideToolGuard`, `hookOutput`).
- **Interpreter/membrane consistency.** `toolGuardInterpreter()` uses `safeRealpath(config.nodeBin)` (`src/tool-guard-hook.ts:26`), the same expression the membrane's `nodeBinReal` uses (`src/service.ts:2559`, `src/spawn-membrane.ts:51`), whose bin dir is bound by `nodeToolchainFlags` (`src/sandbox.ts:342-356`). The script itself is bound by `agentSupportFlags` (`src/sandbox.ts:111-117`), pushed at `src/sandbox.ts:571` — i.e. *after* `--tmpfs home` (`src/sandbox.ts:491-492`), so a bind under a tmpfs'd `$HOME` punches through. Populated on the session path (`src/service.ts:2600`, inside `prepareSpawn`, which both create and resume funnel through) and the aux path (`src/spawn-membrane.ts:114-116`).
- **`--add-dir` argv position.** Emitted immediately after `"claude"` (`src/service.ts:2971-2977`, `src/service.ts:3088-3092`), where the next token is always `--dangerously-skip-permissions`, so the variadic flag cannot swallow the trailing positional prompt (`src/service.ts:3013`). No code indexes the claude argv positionally — the only positional read is `src/herdr.ts:548`, which keys off `argv[0] === "bwrap"` / `--`.
- **Per-family gating is self-consistent.** `claudeFamily`/`guard`/`skills` at `src/service.ts:1789-1791`; hazards gated on `guard` alone (`src/service.ts:1668-1672`), the PR pair on `guard && skills` (`src/service.ts:1809`), the preview hint on `skills` alone (`src/service.ts:1601-1602`). Codex forces both false, and `spawnSettingsOverlay`/`agentSkillsArgs` are only reachable from the Claude spawn/resume builders (`src/service.ts:2980`, `src/service.ts:3097`) — `buildCodexSpawnArgv` (`src/service.ts:3027`) passes neither. Matches the doc-comment table at `src/service.ts:1719-1731` and the matrix tests at `test/prompt-budget.test.ts:311-372`.
- **Say-it-once compensation is complete.** All seven fencing builders now emit `UNTRUSTED_CONTENT_DIRECTIVE` exactly once: `src/critic-core.ts:176` and `:281`, `src/plan-gate.ts:211`, `src/recap-core.ts:112`, `src/rundown-core.ts:708`, `src/namer-llm.ts:40`, `src/autopilot-classify-core.ts:92`, `src/prompt-recommend.ts:93`. The two remaining `fenceUntrusted` call sites (`src/service.ts:2040`, `src/service.ts:2215`) build the session's human-turn prompt, which pairs with the always-resident `untrusted-content-boundary` block (`src/service.ts:1793`, unconditional for every provider) — I traced both back through `composePromptArg` (`src/service.ts:2200-2233`) and confirmed they are create-time only. `UNTRUSTED_CONTENT_DIRECTIVE` (`src/untrusted.ts:57-66`) is self-contained: it names the marker shape, so a label-only fence still has a stated contract.
- **Trim interaction.** `disableableSkillNames` adds `AGENT_SKILL_NAMES` to the subtracted set (`src/service.ts:586`), so a same-named personal skill cannot produce a `skillOverrides: {name: "off"}` entry that would also silence Shepherd's (`src/service.ts:433-438`). Shepherd's skills arrive via `--add-dir`, so they are unaffected by `disableBundledSkills`.
- **Hooks merge.** `spawnSettingsOverlay` merges key-wise (`src/service.ts:441-445`); the guard owns `PreToolUse` (`src/tool-guard-hook.ts:49-61`) and `buildHooksFragment` owns the eight lifecycle events (`src/service.ts:493-507`) — no key collision, and the `hooks` key still disappears when both are off.
- **Guard rule shapes.** Re-read `segments()` (`scripts/tool-guard.mjs:145-219`): the heredoc advance `i += 1 + m[0].length` is correct given the trailing `i++`, and an unterminated quote/heredoc returns `[]` → no decision. `stashTouchesSharedStack` (`scripts/tool-guard.mjs:275-284`) allows `list`/`show`/`create` and `apply <sha>`, denies bare `apply` and `apply stash@{n}`. `isMainModule` realpaths argv[1] (`scripts/tool-guard.mjs:422-431`) with a symlink regression test at `test/tool-guard.test.ts:379-392`.

### Could not verify

- Claude Code's own `PreToolUse` `hookSpecificOutput` schema (`permissionDecision` / `permissionDecisionReason` / `additionalContext` without a decision) and `--add-dir` skill discovery are external CLI behaviours with no ground truth in this repo. The research doc records them as spiked (`docs/research/standing-prompt-trim-2026-08-03.md:60-75`); assumed as written.

## Nits (non-blocking):

- `scripts/tool-guard.mjs:39-40` and `:53-54`: both tmpfs reasons tell the agent `$TMPDIR` is disk-backed, but under a bwrap membrane `--clearenv` drops it — `SANDBOX_ENV_PASSTHROUGH` (`src/sandbox.ts:300-307`) carries no `TMPDIR`, and `src/herdr.ts:513-515` states the outer `env` shim's `TMPDIR=` is wiped by bwrap — so in a sandboxed session `$TMPDIR` is unset and `/tmp` is the bwrap tmpfs; the same sentence's other alternative (the repo worktree / the repo's own filesystem) is correct, so the cost is one wasted call.
- `scripts/tool-guard.mjs:356-364`: every operand after `worktree add` is resolved as a path, so a commit-ish/branch operand resolves to `<cwd>/<name>` — `git worktree add /home/u/wt HEAD` with a tmpfs cwd would deny a disk-backed target while naming a fabricated `/tmp/x/HEAD`. Only reachable after an explicit `cd` into a tmpfs (the session cwd is the worktree), and the comment already acknowledges the trade.
- `scripts/tool-guard.d.mts:3`: says the declaration exists so "the settings-overlay caller can import it", but `src/tool-guard-hook.ts` imports nothing from the `.mjs`; only `test/tool-guard.test.ts:2-9` does.
- `test/untrusted.test.ts:102-105`: the comment claims the invariant is asserted "rather than a hand-kept list of builders", but the `prompts` array at `test/untrusted.test.ts:107-131` is exactly such a list — a future fencing builder left off it still slips through.
- `test/service.test.ts:4786` and `:4790`: the replacement assertion duplicates the preceding `not.toContain("<epic-authoring-notice>")`, so the auto-spawn half of that test now asserts the same thing twice instead of anything about the PR blocks.

- `src/spawn-membrane.ts:114-116` binds the guard script and the agent-skills dir into every aux-spawn membrane even though, as its own comment states, no aux spawn passes Shepherd `--settings` hooks or `--add-dir` today.

PR #2013, opened 2026-08-03
feat(learnings): re-target the flywheel from rules to gotchas

### LINKED ISSUE #2004: Re-target the learnings flywheel from rules to gotchas
## Goal

Re-target the learnings flywheel from **manufacturing behavioral rules** to **capturing non-derivable repo facts** — because the flywheel is, structurally, a machine for producing exactly the thing Anthropic just deleted 80% of.

## The tension

The flywheel captures, distills, admits, ranks and injects house rules — up to a 4,000-char block on every spawn — and can promote them further into a target repo's `CLAUDE.md` and into `~/.claude/CLAUDE.md`. The lifecycle work already added decay, trial and retirement on the *active* side; none of it asks the question the article raises, which is whether the rule should have been written at all.

Inspect any current injected block and both kinds are visible side by side:

- **Facts worth keeping** — `refs/stash` is shared across worktrees; plan-phase-only endpoints must check `planPhase === 'planning'` or the gate leaks into execution; persistent failure toasts need all three of `key`/`duration:null`/`alert:true`. An agent cannot derive these by reading the repo, and getting them wrong is expensive.
- **Rules Claude 5 follows by judgement** — "deliver findings first on analysis tasks", "don't build unrequested features", "capture durable artifacts before teardown". These now compete with the system prompt and the user's actual request for the model's attention, which is precisely the conflicting-instruction failure the article describes.

Two further overlaps to resolve: promotion into `~/.claude/CLAUDE.md` now duplicates a channel Claude fills by itself via auto-memory, and rules promoted into a repo's `CLAUDE.md` are re-read in every turn of every session in that repo, on top of the injected block.

## Vertical cut

Re-target the distiller and optimizer prompts and the admission criteria toward fact-shaped output — what is true, why it bites, how to apply it — with an explicit rejection test for guidance a capable model would follow unprompted. Then sweep the existing active set through the same test and retire what fails. Re-evaluate the promote-to-global path against auto-memory overlap and either narrow or drop it.

## Acceptance

- Distiller output is fact-shaped with a stated why, and the rejection test is documented and applied.
- The current active rule set is classified keep/drop by that test, the sweep is applied, and the resulting injected-block size change is reported via the measurement meter.
- The promote-to-global decision is recorded with its rationale.
- The lifecycle's existing decay/trial/retire behavior still works — this changes admission, not the machinery already built.

scoop:
Re-targets the learnings flywheel at **admission**: it now captures non-derivable repo facts instead of manufacturing behavioral rules. The decay/trial/reap/Wilson-retire machinery from the lifecycle epic is untouched.

Closes #2004.

## The admission test (`src/learning-shape.ts`)

One source of truth, embedded verbatim in the distiller prompt, the optimizer prompt and the sweep tooling so they cannot drift. A candidate qualifies only when all three hold — **non-derivable** (not learnable from the repo, README/CLAUDE.md or `--help` in the session that needs it), **consequential and quiet** ("the linter catches it" is a reject), **attached to an artifact** (names the file, flag, endpoint, table, field or invariant). It rejects outright: general judgement a capable model follows unprompted, restatements of the directives every spawn already carries, preferences with no failure mode, and anything broad enough to apply to any repo.

## What changed

- **Distiller** — fact-shaped ADD contract (*fact — why it bites — how to apply*, in the one line that actually reaches a future agent, since `renderHouseRulesBlock` injects `rule` only). DELETE gains a second criterion: an active rule that **fails the test**, judged on its own text with no signal required. That is what makes the sweep continuous — the daily run keeps draining judgement-shaped rules at 5/run/repo. Promoted rules stay delete-exempt.
- **Optimizer** — rewrites to the same shape, and **omits** a target that cannot be restated as a fact rather than making it a sterner instruction. Omitting leaves it flagged, so the existing auto-retire path takes it.
- **Sweep tooling** — `src/learning-sweep.ts` (pure: plan + price) and `scripts/sweep-house-rules.ts` (dry-run by default, `--apply` retires with `retiredReason = "not-a-gotcha"`, refuses promoted rules). Pricing runs the real spawn path — `planHouseRulesInjection` → `renderHouseRulesBlock` → `measurePromptBlocks` — so the numbers below are the #1999 meter's own math in its own units.
- **Word-boundary trim** at every rule-text entry point. The blunt `slice(0, 240)` had already left a live rule truncated mid-word ("…leaks into execution across consumers when one site "), and fact-shaped entries run longer.

## Sweep applied

Snapshot taken (`VACUUM INTO`) before applying. Every retirement is reversible via `POST /api/learnings/:id/restore`, and `retiredReason` records why.

| | rules | block chars | est-tokens |
|---|---|---|---|
| before | 234 | 22,089 | ~5,525 |
| after | **142** | **9,139** | **~2,288** |

−92 rules, −12,950 chars (−59%). Per repo, biggest movers: pulse 30→17 (−74% chars), oxidone 24→12 (−62%), shepherd-claude-swap 23→11 (−62%), shepherd 40→26 (−54%), topmedia-website 23→13 (−60%).

Two axes, scoped honestly:

- **Shepherd's own 40 rules: both axes** (judgement *and* derivability) — full repo context here. 14 dropped, e.g. "deliver findings first on analysis tasks", "never treat plan approval as opt-in for outward acts" (restates a standing directive), "call the build-queue API to mark each step" (restates the injected build-queue directive), "stamp feature-announcement `sinceVersion` from the helper" (restates the repo's own CLAUDE.md). Kept: the bwrap `--clearenv` fact, `listActiveLearnings` including promoted rows, the `store.ts` prune/cascade gap, the toast `key`/`duration:null`/`alert:true` triple.
- **The other ~194 rules: the judgement axis only.** Recognising "verify before claiming" as model-default needs no repo knowledge; deciding whether a fact is derivable from a repo I have not read does. What survives that pass but might still fail the derivability test is left to the distiller's new DELETE criterion.

One category call worth flagging for review: operator-policy rules with no artifact — "escalate after N plan-review rounds" and friends — fail criterion 3 and were dropped. If those should stand, their honest home is a spawn directive or the repo's CLAUDE.md, not a ranked, decaying house rule.

## Promote-to-global: dropped

`Promoter.promoteGlobal()`, `POST /api/learnings/promote-global`, the `onpromoteglobal` prop chain and its six message keys are gone. Rationale (also recorded in `src/promote.ts` where the method lived): Claude's auto-memory now owns `~/.claude/CLAUDE.md` and is written by the model that reads it; a rule promoted there is resident in every session of every repo — the widest blast radius, for exactly the class of instruction this epic is cutting; and it was the flywheel's one write with no review gate. The per-repo, PR-gated `promote()`/`resyncPromoted()` stay. The cross-repo recurrence band remains as an insight surface with dismiss only, and its lead copy now points at the repo's own CLAUDE.md. The v1.35.0 announcement entry stays as history.

## Verification

Root `bun run lint`, `bunx tsc --noEmit`, `bun test ./test` (8,142 pass); `cd ui && bun run check` (0 errors), `bun run test` (4,514 pass), `check:i18n`; branch-hygiene, announcement-versions, glossary and feature-catalog gates; `fallow audit --base <epic branch> --fail-on-issues` clean. Sweep re-run is a safe no-op (already-retired ids come back REFUSED, nothing re-priced).

Prompt text is untested by construction, so `distillPrompt()`/`optimizePrompt()` are now exported and asserted to embed the shared test — the regression guard against a future edit restoring rule-manufacturing.

## Note for the operator

The live server predates #1999, so `session_prompt_budget` has no rows yet; the numbers above are the meter's computation, and its own recorded rows will show the smaller `shepherd-house-rules` block from the next deploy onward.

### DIFF
--- a/scripts/sweep-house-rules.ts
+++ b/scripts/sweep-house-rules.ts
@@ -1,0 +1,157 @@
+#!/usr/bin/env bun
+/**
+ * One-shot sweep of a repo's active house rules against the admission test (issue #2004).
+ *
+ * The flywheel now admits non-derivable repo FACTS, not behavioral rules (`src/learning-shape.ts`).
+ * The distiller's fails-the-admission-test DELETE criterion keeps the corpus honest from here on;
+ * this script is how the corpus that predates the re-target gets fixed in one pass, and how the
+ * resulting shrink of the injected `<shepherd-house-rules>` block is measured.
+ *
+ * Read-only by default. Nothing is retired without `--apply`, and every retirement is reversible
+ * through `POST /api/learnings/:id/restore` (the reason is stored as `not-a-gotcha`, so the sweep's
+ * retirements stay distinguishable from Wilson auto-retires, expired trials and merges).
+ *
+ *   scripts/sweep-house-rules.ts --repo /path/to/repo
+ *       Dry run: prints the admission test, every active rule, and the current block cost.
+ *
+ *   scripts/sweep-house-rules.ts --repo /path/to/repo --decisions drop.json
+ *       Dry run + projected before/after for the drops in `{"drop":[{"id":"…","why":"…"}]}`.
+ *
+ *   scripts/sweep-house-rules.ts --repo /path/to/repo --decisions drop.json --apply
+ *       Retires them and prints the measured before/after.
+ *
+ * Writes to the live DB directly (a second SQLite writer; WAL + busy_timeout make that safe — see
+ * store.ts). The running server holds no learnings cache, but it does not learn of the change
+ * either: the drawer refreshes on its next fetch. Snapshot first — note that even a dry run opens a
+ * full `SessionStore`, which runs this build's additive schema migrations against the file:
+ *   sqlite3 ~/.shepherd/shepherd.db "VACUUM INTO '/tmp/pre-sweep.db'"
+ */
+import { readFileSync } from "node:fs";
+import { fileURLToPath } from "node:url";
+import { SessionStore } from "../src/store";
+import { resolveDbPath } from "../src/backup-paths";
+import { LEARNING_ADMISSION_TEST, SWEEP_RETIRE_REASON } from "../src/learning-shape";
+import {
+  parseDecisions,
+  planSweep,
+  priceBlock,
+  type BlockCost,
+  type SweepDecision,
+  type SweepPlan,
+} from "../src/learning-sweep";
+
+/** Kept in step with `config.houseRulesBudgetChars`; config.ts is not imported here because loading
+ *  it runs forge/node-bin resolution side effects a CLI has no use for (same reason backup.ts keeps
+ *  its own path resolver). */
+const BUDGET_CHARS = Number(process.env.SHEPHERD_HOUSE_RULES_BUDGET_CHARS ?? 4000);
+
+interface Args {
+  repo: string;
+  decisions?: string;
+  apply: boolean;
+  db: string;
+}
+
+export function parseArgs(argv: string[]): Args {
+  const out: Partial<Args> = { apply: false, db: resolveDbPath() };
+  for (let i = 0; i < argv.length; i++) {
+    const a = argv[i];
+    if (a === "--apply") out.apply = true;
+    else if (a === "--repo") out.repo = argv[++i];
+    else if (a === "--decisions") out.decisions = argv[++i];
+    else if (a === "--db") out.db = argv[++i];
+    else throw new Error(`unknown argument: ${a}`);
+  }
+  if (!out.repo) throw new Error("--repo <path> is required");
+  if (out.apply && !out.decisions) throw new Error("--apply requires --decisions <file>");
+  return out as Args;
+}
+
+function fmt(c: BlockCost): string {
+  return `${c.injectedRules} injected · ${c.chars} chars · ~${c.tokens} est-tokens`;
+}
+
+/** Print the outcome. `done` is what was actually retired (null on a dry run) and `after` the cost
+ *  that goes with it — both passed in rather than read off the plan, so a run that aborts partway
+ *  reports what really happened instead of what it intended. */
+function report(plan: SweepPlan, done: SweepPlan["retire"] | null, after: BlockCost): void {
+  for (const r of plan.refused) console.log(`REFUSED  ${r.id}  (${r.reason})`);
+  const rows = done ?? plan.retire;
+  for (const r of rows) console.log(`${done ? "RETIRED" : "WOULD RETIRE"}  ${r.id}  ${r.rule}`);
+  console.log(`\nactive rules  ${plan.activeBefore} → ${plan.activeBefore - rows.length}`);
+  console.log(`block before  ${fmt(plan.before)}`);
+  console.log(`block after   ${fmt(after)}`);
+  const saved = plan.before.chars - after.chars;
+  const pct = plan.before.chars > 0 ? Math.round((saved / plan.before.chars) * 100) : 0;
+  console.log(
+    `saved         ${saved} chars (${pct}%) · ~${plan.before.tokens - after.tokens} est-tokens`,
+  );
+}
+
+function main(argv: string[]): number {
+  let args: Args;
+  try {
+    args = parseArgs(argv);
+  } catch (e) {
+    console.error(String(e instanceof Error ? e.message : e));
+    console.error(
+      "usage: sweep-house-rules.ts --repo <path> [--decisions <file>] [--apply] [--db <path>]",
+    );
+    return 2;
+  }
+  const store = new SessionStore(args.db);
+  const rules = store.listActiveLearnings(args.repo);
+  if (rules.length === 0) {
+    console.error(`no active house rules for ${args.repo} in ${args.db}`);
+    return 1;
+  }
+
+  if (!args.decisions) {
+    console.log(LEARNING_ADMISSION_TEST);
+    console.log(`\n${rules.length} active rules for ${args.repo}:\n`);
+    for (const r of rules) {
+      const scope = r.scopeGlobs.length ? ` [scope: ${r.scopeGlobs.join(",")}]` : "";
+      console.log(`${r.id}  (${r.status})${scope}\n  ${r.rule}\n`);
+    }
+    console.log(`block cost    ${fmt(planSweep(rules, [], BUDGET_CHARS).before)}`);
+    return 0;
+  }
+
+  let decisions: SweepDecision[];
+  try {
+    decisions = parseDecisions(JSON.parse(readFileSync(args.decisions, "utf8")));
+  } catch (e) {
+    console.error(`decisions file: ${String(e instanceof Error ? e.message : e)}`);
+    return 2;
+  }
+  const plan = planSweep(rules, decisions, BUDGET_CHARS);
+  if (!args.apply) {
+    report(plan, null, plan.after);
+    console.log("\n(dry run — pass --apply to retire)");
+    return 0;
+  }
+  // Retire one at a time and keep the successes: a rule that changed status between the plan and
+  // now returns null, and the report must then describe the partial state, not the intent.
+  const done: SweepPlan["retire"] = [];
+  let failed: string | null = null;
+  for (const r of plan.retire) {
+    if (store.retireLearning(r.id, SWEEP_RETIRE_REASON) === null) {
+      failed = r.id;
+      break;
+    }
+    done.push(r);
+  }
+  // Price the state that now exists, rather than trusting the projection.
+  report(plan, done, priceBlock(store.listActiveLearnings(args.repo), BUDGET_CHARS));
+  if (failed !== null) {
+    console.error(
+      `\nFAILED to retire ${failed} (no longer active?) — stopped, earlier drops stand`,
+    );
+    return 1;
+  }
+  return 0;
+}
+
+if (fileURLToPath(import.meta.url) === process.argv[1]) {
+  process.exit(main(process.argv.slice(2)));
+}
--- a/src/distiller.ts
+++ b/src/distiller.ts
@@ -12,6 +12,12 @@ import { buildTransientAgentArgv } from "./transient-agent-argv";
 import { reapTransientByLabel } from "./transient-tab-reaper";
 import type { RoleEnvironment } from "./default-model";
 import { normalizeRule } from "./learning-rule";
+import {
+  LEARNING_ADMISSION_TEST,
+  LEARNING_FACT_SHAPE,
+  LEARNING_RULE_MAX_CHARS,
+  trimRuleToLimit,
+} from "./learning-shape";
 
 const PROPOSALS_FILE = ".shepherd-learnings.json";
 
@@ -451,7 +457,7 @@ export class DistillerService {
       have.add(key);
       this.deps.store.addLearning({
         repoPath: f.repoPath,
-        rule: r.rule.trim().slice(0, 240),
+        rule: trimRuleToLimit(r.rule),
         rationale: typeof r.rationale === "string" ? r.rationale : "",
         evidence,
         scopeGlobs: sanitizeScopeGlobs(r.scopeGlobs),
@@ -513,20 +519,29 @@ function isBlockedByPruneTombstone(
   );
 }
 
-function distillPrompt(): string {
+export function distillPrompt(): string {
   return [
-    "You are a code-review pattern analyst. Read `signals.json` in this directory.",
+    "You are a repo-gotcha analyst. Your job is NOT to write rules of conduct — it is to capture the",
+    "non-obvious FACTS about one repository that cost an agent real time when it does not know them.",
+    "Read `signals.json` in this directory.",
     "It is a JSON object with four fields:",
     "  `signals` — an array of past corrections, blocks, stalls, and critic findings for one repository;",
-    "  `existingRules` — all recorded/dismissed rules (do NOT ADD any of these — they are the dedup set);",
-    "  `activeRules` — currently-active house rules as {id, rule, promoted} objects. UPDATE/DELETE may target ONLY entries with promoted:false; promoted:true rules are mirrored in CLAUDE.md and may only be flagged ineffective.",
-    "  `proposedRules` — currently-proposed rules as {id, rule} objects (not yet active).",
+    "  `existingRules` — all recorded/dismissed entries (do NOT ADD any of these — they are the dedup set);",
+    "  `activeRules` — currently-active entries as {id, rule, promoted} objects. UPDATE/DELETE may target ONLY entries with promoted:false; promoted:true entries are mirrored in CLAUDE.md and may only be flagged ineffective.",
+    "  `proposedRules` — currently-proposed entries as {id, rule} objects (not yet active).",
+    "",
+    LEARNING_ADMISSION_TEST,
+    "",
+    LEARNING_FACT_SHAPE,
     "",
     "For each candidate finding, choose exactly ONE action:",
-    "  ADD    → emit in `rules`   (new guidance, not already in existingRules)",
+    "  ADD    → emit in `rules`   (passes the admission test, not already in existingRules)",
     "  UPDATE → emit in `updates` (same topic AND same target; richer wording — keep the most informative version)",
-    "  DELETE → emit in `deletes` (the finding directly CONTRADICTS that activeRule — emit any corrected rule as a separate ADD)",
-    "  NOOP   → emit nothing      (already fully covered by an existingRule or activeRule)",
+    "  DELETE → emit in `deletes` for EITHER reason: (a) the finding directly CONTRADICTS that activeRule",
+    "           (emit any corrected fact as a separate ADD), or (b) that activeRule FAILS the admission",
+    "           test above — it is judgement/process guidance, a restated standing directive, or a",
+    "           preference with no failure mode. Reason (b) needs no signal: judge the rule's own text.",
+    "  NOOP   → emit nothing      (fails the admission test, or already fully covered by an existingRule or activeRule)",
     "",
     "MULTI-VALUED GUARD: never UPDATE or DELETE a rule whose target (file / category / object) DIFFERS",
     "from the candidate, even if the topic is similar. A UI rule and a migration rule about the same",
@@ -549,7 +564,7 @@ function distillPrompt(): string {
     'glob patterns (e.g. "src/**", "ui/**/*.svelte") so the rule injects only for tasks touching',
     "those files. OMIT `scopeGlobs` (or use []) for general rules — do not invent a scope when unsure.",
     `Write your output as JSON to \`${PROPOSALS_FILE}\` in this directory, shaped exactly:`,
-    '{"rules":[{"rule":"<=160 char imperative","rationale":"why","evidence":["signalId",...],"scopeGlobs":["glob",...]}],"updates":[{"id":"activeRuleId","rule":"<=160 char imperative","rationale":"why"}],"deletes":[{"id":"activeRuleId","reason":"why"}],"ineffective":[{"id":"activeRuleId","evidence":["signalId",...]}],"reaffirm":[{"id":"proposedRuleId","evidence":["signalId",...]}]}',
+    `{"rules":[{"rule":"<=${LEARNING_RULE_MAX_CHARS} char fact line (see SHAPE)","rationale":"why this is not derivable from the repo","evidence":["signalId",...],"scopeGlobs":["glob",...]}],"updates":[{"id":"activeRuleId","rule":"<=${LEARNING_RULE_MAX_CHARS} char fact line","rationale":"why"}],"deletes":[{"id":"activeRuleId","reason":"contradicted | fails-admission-test, plus one sentence"}],"ineffective":[{"id":"activeRuleId","evidence":["signalId",...]}],"reaffirm":[{"id":"proposedRuleId","evidence":["signalId",...]}]}`,
     'If nothing applies, write {"rules":[],"updates":[],"deletes":[],"ineffective":[],"reaffirm":[]}. Do not write anything else.',
   ].join("\n");
 }
--- a/src/learning-shape.ts
+++ b/src/learning-shape.ts
@@ -47,9 +47,10 @@ export const LEARNING_FACT_SHAPE = [
   "SHAPE — write each qualifying finding as ONE self-contained line carrying all three parts:",
   "  <the fact — what is true, naming the artifact> — <why it bites — the failure it causes> ; <how to apply it>",
   "Only this line is ever shown to a future agent, so it must stand alone. State the fact, not an",
-  'exhortation: prefer "`refs/stash` is one stack shared by every worktree of the repo, so a bare',
-  "`git stash pop` can take another session's entry — use `git stash create` + `apply <sha>`\" over",
-  '"never use git stash".',
+  "exhortation: prefer \"a column added by ALTER is missing from `update()`'s whitelist, so writes",
+  'to it are dropped with no error — add a dedicated setter beside the migration" over "remember',
+  'to add a setter". (Illustrative shape only — never propose an entry that merely restates a',
+  "directive the agent is already handed every session; see the REJECT list above.)",
 ].join("\n");
 
 /** Hard cap on stored rule text — the width every entry point trims to. */
--- a/src/learning-shape.ts
+++ b/src/learning-shape.ts
@@ -1,0 +1,76 @@
+/**
+ * What the learnings flywheel is allowed to admit (issue #2004).
+ *
+ * The flywheel was built to manufacture imperative behavioral rules, distil them, rank them and
+ * inject up to `SHEPHERD_HOUSE_RULES_BUDGET_CHARS` of them into every spawn's system prompt. That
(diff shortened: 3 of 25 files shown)


scoop:
## What I checked

The re-target itself is clean and well-scoped: one shared source of truth (`src/learning-shape.ts`), both prompts import it verbatim, the sweep is pure + unit-tested, and the global-promote path is removed end to end.

**Identifier / wiring resolution (all confirmed against ground truth):**

- `promoteGlobal` / `promote-global` / `promoteGlobalLearning` / `extractLearningsBlockRules` / `onpromoteglobal` have **zero** remaining references in `src/`, `ui/src/` or `scripts/` — the only hits are the three test files that assert their absence and the historical `ui/src/lib/feature-announcements/entries/v1.35.0-learnings-promote-global.ts`. Verified by tree-wide grep.
- `POST /api/learnings/promote-global` now falls through `handleMergeSuggestionPost` (`src/server.ts:1874-1882`) into `handleLearningIdAction`, which returns `null` for `parts[3] === undefined` (`src/server.ts:1919-1954`) → generic 404, matching `test/server-learnings.test.ts:127-142`.
- `LEARNINGS_START` / `LEARNINGS_END` / `upsertLearningsBlock` are still exported and still used by `test/promote.test.ts:4-8,555,574` after `extractLearningsBlockRules` was deleted — no dangling import.
- Dropping `mkdirSync`/`dirname` from the default `writeClaudeMd` (`src/promote.ts:45`) is safe: both remaining call sites write into an existing worktree root (`src/promote.ts:104,114` and `src/promote.ts:245,257`). Only the deleted home-dir write needed the mkdir.
- `priceBlock` (`src/learning-sweep.ts:38-50`) composes real spawn-path functions: `planHouseRulesInjection` (`src/house-rules.ts:206`), `renderHouseRulesBlock` (`src/house-rules.ts:275-279`, returns `null` on empty — the `chars: 0` branch is correct), `HOUSE_RULES_TAG` (`src/house-rules.ts:9`), `measurePromptBlocks` (`src/prompt-budget.ts:71-87`, `{name,text}` shape matches). The single-block `totalTokens === ceil(chars/4)` assertion in `test/learning-sweep.test.ts:37` holds given `CHARS_PER_TOKEN = 4` (`src/prompt-budget.ts:56`).
- `planSweep` refuses `promoted` before `active` (`src/learning-sweep.ts:69-76`), which matters because `store.listActiveLearnings` returns **active + promoted** (`src/store.ts:4395-4404`) and `retireLearning` would happily retire a promoted row (`src/store.ts:4580-4592`). Guard is correct.
- `resolveDbPath` exists at `src/backup-paths.ts:12`; `SessionStore(path)` at `src/store.ts:1127`; `retireLearning(id, reason)` at `src/store.ts:4582`. All script imports resolve.
- `SHEPHERD_HOUSE_RULES_BUDGET_CHARS ?? 4000` in the script (`scripts/sweep-house-rules.ts:52`) matches `src/config.ts:790`.
- **Locale parity verified**: `learnings_proposed_admission_hint`, `feat_learnings_gotchas_title`, `feat_learnings_gotchas_body` all present in both `ui/messages/en.json:3536-3538` and `ui/messages/de.json:3536-3538`; the six `learnings_recur_promote*` keys are gone from **both** catalogs (grep returns no hits in either); `feat_learnings_promote_global_*` remains in both (`en.json:1940-1941`, `de.json:1940-1941`) for the retained historical entry.
- Announcement gate: `package.json:3` is `1.46.0`, the new entry is `sinceVersion: "1.47.0"` with a matching `v1.47.0-` filename prefix — satisfies both assertions in `scripts/check-announcement-versions.mjs:14-19`. Entries are glob-loaded (`ui/src/lib/feature-announcements.ts:47-48`), so no registry edit is needed. `ui/src/lib/components/**` is touched and a catalog fragment is added, so `scripts/check-feature-catalog.sh:93-105` passes.
- `m.learnings_proposed_admission_hint` is called at `ui/src/lib/components/learnings-drawer/RepoGroup.svelte:88`; `group.proposed` is a real field of `RepoGroupData` used two lines below at :90.
- `trimRuleToLimit` replaces `.trim().slice(0,240)` at all four in-diff sites (`src/distiller.ts:460`, `src/learnings-service.ts:120`, `src/store.ts:4488`, `src/store.ts:4511`); `addLearning` does no trimming of its own (`src/store.ts:4324-4357`), so the distiller's call is the only cap on that path — correct.
- Lifecycle machinery is untouched: no change to `src/learnings-lifecycle.ts` or `src/house-rules.ts` in the diff.

**Could not verify (not in this repo):** whether the sweep was actually applied to the live `~/.shepherd/shepherd.db` and the reported per-repo before/after numbers — that is operational state outside the tree. The tooling that would produce those numbers is present and its math is unit-tested. Likewise I cannot check the PR's base branch from the worktree; the task requires `epic/2005-epic-adopt-the-claude-5-context-engineer`.

## Findings

### 1. `src/optimizer.ts:380-381` — the prompt asserts something the lifecycle does not do

> "Omitting leaves it flagged, which is the correct outcome: the lifecycle retires it."

That is true only while `autoOptimizeFlagged` is off (its default — `src/store.ts:959`, `src/store.ts:4053`). With it **on**, `runAutoRetire` sends a `shouldRetire` candidate to `optimizeOne` and skips retirement whenever `store.autoOptimizedAt(rule.id) === null` (`src/learnings-lifecycle.ts:388-390`). `autoOptimizedAt` is stamped **only** by `reviseLearning` (`src/store.ts:4493-4496`), which never runs for an omitted target (`applyRevisions`, `src/optimizer.ts:339-346`). So on those repos an omitted target is re-enqueued for optimization on every sweep (`src/index.ts:2529`) and never retired by this path — one wasted transient-agent spawn per sweep, indefinitely.

This matters more after this PR than before it: the previous prompt only omitted rules that "genuinely cannot be improved", whereas the new text deliberately routes an entire class (everything failing the admission test) down the omit path, and then tells the model that path terminates in retirement. Either correct the claim (the distiller's new DELETE criterion is what actually drains these, not `runAutoRetire`) or make the omit path terminal.

### 2. `ui/src/lib/components/LearningsDrawer.svelte:275-276` — comment describes a removed feature

The comment directly above the changed `<RecurrenceBand …>` line still reads "rules that recur across repos, suggested for the **user-global CLAUDE.md**. Promote (guarded, two-step) writes the rule there; or dismiss." Both sentences are now false: the promote button and its two-step confirm were deleted in this diff (`ui/src/lib/components/learnings-drawer/RecurrenceBand.svelte:28-37`), and the band's own lead text was rewritten to point at *each repo's own* `CLAUDE.md` (`ui/messages/en.json:1888`). Update it to match.

## Nits (non-blocking):
- `src/learning-shape.ts:65-71`: `trimRuleToLimit` backs off to the *last* space in the first 240 chars with no floor, so pathological input (e.g. one space at index 5 followed by a 240-char unbroken token) yields a 5-char rule where the old hard slice kept 240; a minimum-retained-length guard would bound the loss.
- `src/distiller.ts:567` asks the model for `deletes[].reason` as `"contradicted | fails-admission-test, plus one sentence"`, but `applyDeletes` (`src/distiller.ts:421-431`) discards `d.reason` and always stores `retiredReason = "superseded"` — so the very distinction `SWEEP_RETIRE_REASON` was introduced to preserve (`src/learning-shape.ts:73-76`) is lost on the durable DELETE path.
- `scripts/sweep-house-rules.ts:61-74`: `--decisions` or `--db` as the final argv token silently consumes `undefined` — the first degrades to a dry-run listing, the second opens an in-memory DB — rather than erroring like the missing-`--repo` case does.
- `scripts/sweep-house-rules.ts:20-27` documents invocation as `scripts/sweep-house-rules.ts --repo …`, but the file is committed mode `100644` (same as `scripts/backup.ts`), so it needs a `bun` prefix.
- `scripts/sweep-house-rules.ts:116` prints `LEARNING_ADMISSION_TEST` on a dry run but not `LEARNING_FACT_SHAPE`, so the operator reading the list sees the reject test without the shape the keepers are supposed to have.

- `scripts/sweep-house-rules.ts` ships a `--db` override and an exported `parseArgs` beyond what a one-shot against the default DB path needs.

- `src/merge-suggest.ts:381` and `:413` still use `.trim().slice(0, 240)` — the same mid-word cut this PR replaced in the four in-diff sites, and it wins over `store.mergeLearning`'s new word-boundary trim because it runs first.
- `ui/src/lib/feature-announcements/entries/v1.35.0-learnings-promote-global.ts` still announces the one-click global promote to any user whose `lastSeen` predates 1.35.0, describing a capability that no longer exists.
- `src/server.ts:1993`'s comment describes the edited-rule normalization as "trim + 240 cap"; the cap still holds but the trim is now word-boundary.

scoop:

**1. `src/optimizer.ts` — the false "the lifecycle retires it" claim: FIXED.** The prompt line is now `"Omitting leaves it flagged and unchanged, which is the correct outcome; do not rescue such an entry by making it a sterner instruction."` (src/optimizer.ts:388-389) — no retirement claim left. The class doc-comment (src/optimizer.ts:81-87) now states the mechanism explicitly and I verified every clause of it: `retireRepoCandidates` takes the `void optimizer.optimizeOne(rule.id)` branch (no retire, no budget consumed) whenever `autoOptimizeFlagged && store.autoOptimizedAt(rule.id) === null` (src/learnings-lifecycle.ts:388-390), the only writer of that column is `reviseLearning` (src/store.ts:4495), and the `else if` retire branch is what runs when `autoOptimizeFlagged` is off (src/learnings-lifecycle.ts:391-392). The comment's attribution of the drain to the distiller's DELETE criterion matches `applyDeletes` (src/distiller.ts:421-432, active-only, capped at `MAX_DELETES_PER_RUN = 5`).

**2. `ui/src/lib/components/LearningsDrawer.svelte` — stale recurrence-band comment: FIXED.** Lines 275-277 now read "…surfaced as a hint that the underlying fact belongs in each repo's own CLAUDE.md. Dismiss is the only action: the one-click write into the user-global ~/.claude/CLAUDE.md was dropped in #2004." That matches the component: `RecurrenceBand.svelte` now renders only the dismiss button (ui/src/lib/components/learnings-drawer/RecurrenceBand.svelte:28-37) and the lead string is `learnings_recur_lead` = "…a hint the underlying fact belongs in each repo's own CLAUDE.md." (ui/messages/en.json:1888, de.json:1888).

## What I verified in this diff

- **Removal of promote-to-global is complete and consistent.** No `promoteGlobal` / `promote-global` / `promoteGlobalLearning` / `onpromoteglobal` remains anywhere outside test comments (grep over the tree; only test/merge-suggest.test.ts:217, test/server-learnings.test.ts:88/127/133 and the historical entry id in ui/src/lib/feature-announcements/entries/v1.35.0-learnings-promote-global.ts:8). `src/index.ts:2386` passes the `Promoter` instance and `AppDeps["promoter"]` now requires only `promote` (src/server.ts:516) — still structurally satisfied. `POST /api/learnings/promote-global` now falls through `handleMergeSuggestionPost` (src/server.ts:1879-1881) to `handleLearningIdAction`, which returns null because `parts[3]` matches no action (src/server.ts:1922-1953) — i.e. the 404 the updated test asserts. `test/promote.test.ts` no longer imports the deleted `extractLearningsBlockRules` (imports at test/promote.test.ts:3-9).
- **Dropping `mkdirSync(dirname(p))` from the default `writeClaudeMd` is safe**: both remaining writes target a CLAUDE.md inside an already-created worktree (src/promote.ts:104/114 and 255/257).
- **i18n parity.** The 6 `learnings_recur_promote*` keys are removed from BOTH catalogs, and the 3 added keys exist in both: `learnings_proposed_admission_hint` (en.json:3535, de.json:3535), `feat_learnings_gotchas_title`/`_body` (en.json:3536-3537, de.json:3536-3537). `feat_learnings_promote_global_*` stays in both (still referenced by the v1.35.0 entry). `m.learnings_proposed_admission_hint()` is the only new message call site (ui/src/lib/components/learnings-drawer/RepoGroup.svelte:88).
- **Feature announcement.** `sinceVersion: "1.47.0"` matches the filename and is > package.json's 1.46.0 (package.json:3), consistent with the sibling 1.47.0 entries already on the base; the catalog auto-globs entries (ui/src/lib/feature-announcements.ts:46-49), so no registration step is missed, and `scripts/check-feature-catalog.sh` is satisfied by the touched entry file.
- **Sweep tooling resolves against real APIs.** `resolveDbPath` (src/backup-paths.ts:12), `listActiveLearnings` = active+promoted (src/store.ts:4396-4404), `retireLearning(id, reason)` (src/store.ts:4582), `planHouseRulesInjection(rules, budgetChars, now, targetPaths?)` (src/house-rules.ts:206-211), `renderHouseRulesBlock` returning `null` on empty (src/house-rules.ts:275-276), `HOUSE_RULES_TAG` (src/house-rules.ts:9), and `measurePromptBlocks` returning `totalChars`/`totalBytes`/`totalTokens` (src/prompt-budget.ts:71-86). The module note that scope-gated rules never price into the delta is correct: `learningMatchesScope` returns false for empty `paths` (src/house-rules.ts:174). The script's `BUDGET_CHARS` mirrors `config.houseRulesBudgetChars` exactly, env var and 4000 default included (src/config.ts:790).
- **The 240-cap change does not break untouched tests.** `trimRuleToLimit` (src/learning-shape.ts:50-56) falls back to a hard slice when there is no space, so the existing exact-240 assertions on space-free inputs still hold: test/learnings-service.test.ts:131 (`"x".repeat(300)`), test/store-learnings.test.ts:285/812, test/server.test.ts:1883. `addLearning` stores `rule` verbatim (src/store.ts:4324-4336), so the trim at the three call sites is the only cap — all three are updated.

## Blocking finding

`src/learning-shape.ts`: `LEARNING_FACT_SHAPE` holds up the `refs/stash` worktree-safety fact as the model of a qualifying entry (lines 50-52), while `LEARNING_ADMISSION_TEST` orders the agent to reject and delete "a restatement of a directive the agent is already given every session (worktree/stash safety, tmpfs, untrusted content, …)" (line 37). That directive is still injected into every spawn today: `WORKTREE_STASH_NOTICE` carries the same `refs/stash`-is-one-shared-stack fact *and* the same `git stash create` + `apply <sha>` remedy (src/service.ts:604-613) and is pushed onto the prompt at src/service.ts:1642. Both blocks ship verbatim, adjacent, into the distiller (src/distiller.ts:533/535) and optimizer (src/optimizer.ts:379/381) prompts, so the PR's declared "single source of truth" gives the model an exemplar it is simultaneously told to reject — the conflicting-instruction failure the module header (src/learning-shape.ts:1-19) says the re-target exists to remove. Cheapest fix: use an example that is not a spawn notice (the plan-phase gate or the `key`/`duration:null`/`alert:true` toast fact both appear in the issue), or drop "worktree/stash safety" from the reject bullet.

## Nits (non-blocking):
- src/distiller.ts: the DELETE contract now asks the model for `"reason":"contradicted | fails-admission-test, plus one sentence"` (line 566) but `applyDeletes` discards `d.reason` and always retires with `"superseded"` (src/distiller.ts:428), so a fails-the-test drain is indistinguishable from a contradiction drain — `SWEEP_RETIRE_REASON` would fit.
- scripts/sweep-house-rules.ts:155: `fileURLToPath(import.meta.url) === process.argv[1]` diverges from the `import.meta.main` idiom every other Bun script here uses (scripts/backup.ts:194, issue-triage.ts:397, pre-push.ts:699, usage-report.ts:957); any path/symlink mismatch turns the script into a silent no-op instead of running.
- scripts/sweep-house-rules.ts:68: `--db` with no following value leaves `args.db` `undefined` despite the `Args.db: string` type (the same `argv[++i]` hole in `--repo`/`--decisions` is caught by the required checks below it).
- ui/messages/en.json: `feat_learnings_promote_global_title`/`_body` and the v1.35.0 entry stay, so a user upgrading across 1.35 still gets a What's-New card for the one-click global promote this PR removes (the new 1.47.0 card does say it is gone, so it self-corrects in the same drawer).

- src/learning-sweep.ts / scripts/sweep-house-rules.ts: a pure module + CLI + 111 lines of unit tests for what the plan calls a one-shot sweep; defensible as the way the meter delta is reproducible, but it is durable surface for a single-use job.
- src/promote.ts: the default `writeClaudeMd` lost its `mkdirSync` (line 45) — correct now that no write targets a possibly-absent dir, but it is a behaviour change beyond deleting `promoteGlobal`.

- src/merge-suggest.ts still asks its agent for "the consolidated rule, an imperative ≤240 chars" (lines 503, 524), so merges re-mint text in the old rule shape the distiller/optimizer no longer produce.
- `Learning.retiredReason` is not rendered anywhere in the UI (only ui/src/lib/types.ts:2246 and test fixtures), so the new `not-a-gotcha` reason is visible only in the DB.

## Stated limitations
- The acceptance item "the sweep is applied to the live set and the before/after is reported" cannot be verified from the diff — it mutates `~/.shepherd/shepherd.db`, which is outside this repo. The tooling that produces those numbers is present and unit-tested; the numbers themselves have to be checked in the PR body.
- I did not run lint/tsc/tests (CI's job, per instructions); all claims above are from reading the code.

scoop (commented):
## Prior finding — confirmed addressed

**1. `src/learning-shape.ts`: fact-shape exemplar was a spawn-notice restatement.** Fixed in `e90b8e2d`. `LEARNING_FACT_SHAPE` no longer offers the `refs/stash` shared-stack fact; the exemplar is now a DB-column one (`src/learning-shape.ts:50-51`: "a column added by ALTER is missing from `update()`'s whitelist, so writes to it are dropped with no error — add a dedicated setter beside the migration"), which is not a restatement of any every-session directive. I checked the spawn notices it must not collide with: `WORKTREE_STASH_NOTICE`/`TMPFS_WORKTREE_NOTICE` etc. (`src/service.ts`) cover stash, tmpfs, untrusted content, single-PR, plan gate, build queue — none of which the new example touches. The `worktree/stash safety` reject bullet at `src/learning-shape.ts:37-38` therefore no longer contradicts the example, and `e90b8e2d` also added the explicit "Illustrative shape only — never propose an entry that merely restates a directive the agent is already handed every session" guard at `src/learning-shape.ts:52-53`. Both blocks still ship adjacent into the distiller (`src/distiller.ts:533/535`) and optimizer (`src/optimizer.ts:379/381`) prompts, which is now consistent.

## Verification performed

**Identifiers resolved.** Every symbol the new modules import exists with the shape used:
- `planHouseRulesInjection(rules, budgetChars, now?, targetPaths?)` — `src/house-rules.ts:206-211`; `renderHouseRulesBlock(injected): string | null` — `src/house-rules.ts:275-279`; `HOUSE_RULES_TAG` — `src/house-rules.ts:9`. `src/learning-sweep.ts:347` imports exactly these three.
- `measurePromptBlocks(blocks: PromptBlock[])` returning `totalChars`/`totalBytes`/`totalTokens` — `src/prompt-budget.ts:71-87`; `PromptBlock` is `{name, text}` (`src/prompt-budget.ts:16-19`), matching the `[{ name: HOUSE_RULES_TAG, text }]` call in `src/learning-sweep.ts`.
- `store.listActiveLearnings(repoPath)` returns active **+ promoted** — `src/store.ts:4395-4404`; `store.retireLearning(id, reason)` returns `Learning | null` — `src/store.ts:4582-4592`. Both match how `scripts/sweep-house-rules.ts:109/144` use them, and `planSweep`'s `promoted` refusal (`src/learning-sweep.ts:422`) is needed precisely because `listActiveLearnings` includes promoted rows.
- `resolveDbPath()` — `src/backup-paths.ts:12-14`. `SessionStore(path: string)` — `src/store.ts:1127`.
- Budget default: the script's `Number(process.env.SHEPHERD_HOUSE_RULES_BUDGET_CHARS ?? 4000)` matches `config.houseRulesBudgetChars` at `src/config.ts:790` exactly.

**Scope-gating claim in `src/learning-sweep.ts` is accurate.** With no `targetPaths`, `planHouseRulesInjection` sets `paths = []` (`src/house-rules.ts:212`) and every non-Always rule falls into `gated` (`src/house-rules.ts:216-220`), so the module's stated caveat — a scoped rule's retirement moves the count, not the chars — is true, and `test/learning-sweep.test.ts` asserts it.

**Removal is complete.** Grepping `promoteGlobal|extractLearningsBlockRules|promote-global|onpromoteglobal|learnings_recur_promote` across the tree leaves only the historical v1.35.0 announcement entry (`ui/src/lib/feature-announcements/entries/v1.35.0-learnings-promote-global.ts:8`, whose `feat_learnings_promote_global_*` keys are correctly retained in both catalogs) and test comments. `dirname`/`homedir`/`mkdirSync` are gone from `src/promote.ts` with no remaining users (`writeClaudeMd` is only called at `src/promote.ts:114` and `:257`, both onto an existing worktree dir); `LEARNINGS_START/END` are still used at `src/promote.ts:307-370`. `AppDeps.promoter` narrowed to `{promote}` still structurally accepts the `Promoter` instance wired at `src/index.ts:2386/3005`.

**Dead-route behaviour verified.** `POST /api/learnings/promote-global` now falls through `handleMergeSuggestionPost` (`src/server.ts:1874-1881`) into `handleLearningIdAction`, where `parts[3]` is undefined so every branch misses and it returns `null` (`src/server.ts:1919-1954`) → 404, which is what `test/server-learnings.test.ts:127` asserts.

**i18n parity confirmed.** All three new keys exist in both catalogs at the same positions: `ui/messages/en.json:3536-3538` and `ui/messages/de.json:3536-3538`. The six removed `learnings_recur_promote*` keys are gone from both and unreferenced. `m.learnings_proposed_admission_hint()` is used at `ui/src/lib/components/learnings-drawer/RepoGroup.svelte:88`, which already imports `m` (line 2). `common_cancel` was only used by the deleted confirm block; `RecurrenceBand.svelte` no longer references it and has no unused import (lines 1-13).

**Announcement entry.** `sinceVersion: "1.47.0"` in `v1.47.0-learnings-gotchas.ts` matches the filename and sits alongside the other unreleased 1.47.0 entries (`v1.47.0-prompt-budget.ts`, `v1.47.0-fast-repo-switch.ts`), against `package.json` version `1.46.0` — i.e. next-version, as `ui/src/lib/feature-announcements.ts` requires. Entries are auto-globbed, so no index edit is needed.

**Trim helper.** `trimRuleToLimit` (`src/learning-shape.ts:66-72`) is correct on the boundaries: `<=240` passes through trimmed; overlong text with no space falls back to a hard slice; the distiller test's 399-char `"word "×80` case cuts at index 239 and ends on `word`. All four former `slice(0, 240)` sites on the learnings path now route through it (`src/distiller.ts:460`, `src/learnings-service.ts:120`, `src/store.ts:4488`, `src/store.ts:4511`), and `store.addLearning` does no trimming of its own (`src/store.ts:4324-4336`), so callers remain the only cap.

**Optimizer doc comment is factually correct.** The added note at `src/optimizer.ts:83-88` claims an omitted target is not self-draining under `autoOptimizeFlagged`. Verified: `runAutoRetire` enqueues `optimizeOne` and skips retirement while `store.autoOptimizedAt(rule.id) === null` (`src/learnings-lifecycle.ts:388-390`), and only `reviseLearning` stamps `autoOptimizedAt` (`src/store.ts:4492-4496`). The flag is off by default (`src/store.ts:959`, column default 0 at `src/store.ts:4053`), and the omit path itself predates this PR (the old prompt already said "If a rule genuinely cannot be improved, omit it"), so this widens an existing, opt-in, documented behaviour rather than introducing one — noted, not raised.

## Nits (non-blocking):
- `scripts/sweep-house-rules.ts:161` uses `fileURLToPath(import.meta.url) === process.argv[1]` as its main-guard, while every other script in `scripts/` uses Bun's `import.meta.main` (`scripts/backup.ts:194`, `scripts/pre-push.ts:699`, `scripts/usage-report.ts:957`, `scripts/issue-triage.ts:397`, `scripts/eval-stop-classifier.ts:666`); the odd form works but silently no-ops if the two paths ever disagree.
- `scripts/sweep-house-rules.ts:68`: `--db` as the final argv token assigns `argv[++i]` = `undefined`, overriding the `resolveDbPath()` default and reaching `new SessionStore(undefined)` (`src/store.ts:1127` types it `string`); the other flags are caught by the `!out.repo` / `apply && !decisions` checks, but `--db` is not.
- `src/distiller.ts:566` now asks the model for `"reason":"contradicted | fails-admission-test, plus one sentence"`, but `applyDeletes` (`src/distiller.ts:421-432`) discards `d.reason` and always retires with `"superseded"` — so the new fails-the-test retirements are indistinguishable in `retiredReason` from contradiction-driven ones, unlike the sweep's `not-a-gotcha` (`src/learning-shape.ts:77`). Nothing reads `retiredReason` in the UI today, so this costs only forensics.
- `src/server.ts:1993` still describes the service normalization as "trim + 240 cap"; the cap is still 240, but the trim is now word-boundary (the sibling comment in `src/learnings-service.ts:115` was updated).
- `src/learning-sweep.ts:374`/`scripts/sweep-house-rules.ts:87` label the count `activeBefore` / print it as `active rules`, though `listActiveLearnings` includes promoted rows; the interface comment says so, but the printed line does not.

- `ui/src/lib/components/learnings-drawer/RepoGroup.svelte:85-89` places the admission hint once per repo group rather than on `ProposedRuleCard.svelte` as the plan's file table specified — a sensible reduction (one line instead of one per card), noted only as a deviation from the stated file list.
- `src/learning-sweep.ts:404-442` supports duplicate-id collapsing and three distinct refusal reasons for what the plan scopes as a one-shot migration; small, tested, and it makes the stale-decisions-file case visible, so it reads as justified rather than speculative.

Issue #1405, opened 2026-07-04
Epic-authoring skill: Claude-Code-only delivery leaves a Codex steer-time gap


kai-osthoff:
## Summary

`shepherd-epic-authoring` (added in a307d6a2, #1402) is a `.claude/skills/…/SKILL.md`
skill. Its **outcome** is fully backend-agnostic, but its **auto-invocation** is
Claude-Code-only. On a Codex-driven session the skill neither auto-surfaces nor
auto-invokes, so the mid-session (steer-time) epic-shape guidance it was built to
provide is missing. This issue documents the gap and scopes a fix.

## Two layers, two answers

**What the skill *produces* → agnostic ✅** — nothing here is Claude-specific:
- Recognition contract: `parseEpicBody` / `FENCE_RE` in `src/epic-parse.ts` read only
  the issue body (the `epic-dag` fence / task-list of real `#N` numbers). It doesn't
  care which CLI wrote the body.
- Creation: `gh issue create` / `gh issue edit` — standard CLI.
- Native links: `POST /api/epic/import` on the herdr server (`:7330`).

A Codex session that *runs these steps* yields an identical recognized epic.

**How the skill is *delivered / invoked* → Claude Code only ⚠️**
- The file lives in `.claude/skills/shepherd-epic-authoring/SKILL.md` — Claude Code's
  **native** skill format (auto-discovery → `/shepherd-epic-authoring` → Skill tool).
- Codex does **not** read `.claude/skills/`; it uses `AGENTS.md` + its own prompt.
- `src/commands.ts` scans `.claude/skills/` only to populate the New-Task autocomplete;
  the actual slash-command *expansion* is done by the Claude CLI, not by Shepherd. So a
  Codex session gets no auto-surfacing and no auto-invoke of this skill.

## Where the two layers already diverge

The **spawn-time** epic-shape contract is already injected agnostically:
`composeSystemPrompt` (`src/service.ts:1088-1092`) appends `<epic-authoring-notice>`
on `epicIntent` with **no per-provider gating** — Codex and Claude both receive it.
(Contrast the manual-steps notice right above at `service.ts:1085`, which *is*
suppressed for Codex.)

But `detectEpicIntent` (`src/service.ts:591`, called at `:1945`) is **spawn-time only**.
When an operator says "make this an epic" **mid-session (steer-time)**, no injected
guidance fires — and the skill that was meant to fill that steer-time gap is exactly the
[...]

PR #2014, opened 2026-08-03
feat(spawn): drive the build queue + epic draft through tools, not prose (#2003)

### LINKED ISSUE #2003: Replace prose control-plane directives with tool interfaces
## Goal

Replace prose that *explains how to call Shepherd's control plane* with tool interfaces that make the call obvious from their parameters. The article's second shift: stop giving Claude worked examples, design the interface instead.

## What is prose today

**Build-queue directive — 3,443 chars, measured.** It teaches the agent to drive the build queue in English and **bakes a literal control-plane token into a `curl` inside the system prompt**. That is both the largest single conditional block and a credential living in prompt text.

**Epic-authoring — stated three times.** A system-prompt directive (2,414 chars, pre-baked with endpoint, session id and token), a PTY-delivered notice (~1,050 chars, because the skill is model-invoked and the notice compensates for that), and the shipped epic-authoring skill. Three renderings of one capability, all of which the model must reconcile.

A tool named `shepherd_queue_step(step, status)` with `status` enumerated `pending | active | done` carries the entire semantic in its signature — the same way the article's `TodoWrite` example replaces ~9,100 characters of when-to-use prose with a well-typed parameter plus one behavioral line. Shepherd already has the plumbing: a plugin membrane with `extraArgs`/`patchEnv`, and the `ctx.sessions` capability shipped in #1958.

## Vertical cut

Expose the build-queue transitions and the epic-draft submission as tools, delete the prose blocks that taught the HTTP calls, and collapse epic-authoring to one home. Keep whatever the other model family genuinely needs — Codex receives directives inline in the prompt rather than via a flag and may not have the same tool surface.

## Acceptance

- Build-queue step transitions and epic-draft submission happen through tool calls in a real session.
- The build-queue and epic-authoring prose blocks are deleted from the composed prompt; the measurement meter shows the drop.
- **No control-plane token appears anywhere in prompt text.**
- Epic authoring has one home; the compensating PTY notice is gone or justified in the PR.
- Codex spawns still drive the build queue, or the gap is stated explicitly.

scoop:
Closes #2003. Part of the Claude 5 context-engineering epic (#2005) — targets the epic integration branch.

The article's second shift: stop giving Claude worked examples, design the interface instead. Shepherd's build-queue directive was 3.4 KB of `curl` tutorial with a live control-plane token baked into it; the epic-authoring directive hand-wrote a JSON body. Both are now tools whose parameters carry the contract.

## Phase 0 — the transport, settled empirically before any code

The discriminating unknown was whether Claude Code 2.1.220 can drive an **HTTP** MCP server that answers with plain `application/json` JSON-RPC (no SSE, no `Mcp-Session-Id`). If not, this becomes a stdio child process needing `bun` + a script path bound inside the bwrap membrane, plus an SDK dependency.

A throwaway 60-line Bun server + a real `claude -p` run answered it: `initialize` → `notifications/initialized` → `tools/list` → `tools/call` all work against plain JSON responses; it works with **and** without `--strict-mcp-config`; and the client sends `Content-Type: application/json` with **no `Origin`**, so Shepherd's existing `requireJsonContentType` / `checkOrigin` guards pass untouched. So the MCP server is one more HTTP route — no child process, no new dependency, no membrane binds.

## What ships

**`src/agent-control.ts`** — one module, three parts:

- **Appliers** (`applyQueueWrite` / `applyQueueStep` / `applyEpicDraft`): validate → store → emit, returning a transport-neutral result. The REST handlers in `server.ts` are now thin renderings of these, so the curl path and the tool path cannot drift in validation, events, or 404/409 semantics.
- **A session-gated tool catalog**: queue tools when the repo runs the build queue, `epic_draft` for an epic-authoring session. Empty catalog ⇒ no `--mcp-config` at all, so a plain session pays nothing.
- **A stateless MCP handler** — pure over `deps`, so the whole protocol is unit-tested against an in-memory store.

**`POST /api/sessions/:id/mcp`**, allowlisted on the agent ingress. Strictly weaker than the routes beside it: every tool it exposes is one of them, and neither approve gate has a tool at all.

**`--mcp-config`** on the Claude spawn **and** resume argv (per-process — a resumed session that didn't re-pass it would silently lose its tools mid-run). No `--strict-mcp-config`: the operator's own MCP servers must survive.

| Tool | Parameters | Prose it deletes |
|---|---|---|
| `queue_write` | `steps: [{id, title, detail?}]` | the PUT curl + the entire id-stability paragraph — `id` is a required param, so "assign and resend your own id" is structural now |
| `queue_step` | `stepId`, `status: pending\|active\|done\|skipped` | both POST curls, the status vocabulary, the id-resolution fallback paragraph |
| `epic_draft` | `parent{…}`, `children[{key, blockedBy, …}]` | the hand-written JSON body + PUT/GET curl |

## Measured (identical inputs, before vs after)

| Block | before | after (Claude) | after (Codex) |
|---|---|---|---|
| `<build-queue>` | 3,635 | **612** (−83%) | 3,635 |
| `<epic-authoring-directive>` | 2,510 | **1,618** (−36%) | 2,400 |
| autopilot spawn + build queue | 13,013 | **9,990** (−23%) | — |
| epic-authoring spawn | 8,078 | **7,186** (−11%) | — |

With this deployment's token set, the two directives were 3,955 / 2,590 — that extra 400 chars *was* the credential. The `#1999` kitchen-sink meter fixture drops 13,600 → 10,577 and is re-pinned in `test/prompt-budget.test.ts`.

## No control-plane token in prompt text — and it never did anything

`resolveSpawnBaseUrl` already points every agent at the restricted agent-ingress listener, which is built with `skipAuth` (#1079). The `Authorization: Bearer <token>` in both directives was authenticating against a listener that never checks it, while persisting Shepherd's control-plane secret into the agent's on-disk transcript. It is gone from both providers; a test asserts neither directive matches `/Authorization|Bearer/i`. `buildHooksFragment`'s `$SHEPHERD_TOKEN` placeholder is untouched.

Residual, documented in the code: if `agentIngressPort` is ever unknown and the gated main port is baked instead, a token-set deployment's queue calls 401 — the same fail-open degradation hooks already accept, and why the ingress port is pinned.

## Live verification (acceptance: "in a real session")

A real `claude` process, given exactly the argv fragments the spawn path emits (the real `agentMcpConfigArg` + the shortened directives) against the real ingress route → appliers → store. Only herdr/PTY is out of the loop.

- **Build queue** — from the 612-char block alone, with zero HTTP instructions: authored two steps via `queue_write` with its own stable ids (`s1`/`s2`), then `queue_step` → `s1` active. Store confirmed.
- **Epic draft** — submitted via `epic_draft` with a `blockedBy: ["c1"]` edge on the second child; `store.getEpicDraft` confirmed. No `gh`, no curl.

## Epic authoring: three renderings → one per flow

- Draft flow: the directive + `epic_draft` tool. The skill's "draft-only mode" section existed only to stop its own write stages firing inside that flow — that reconciliation burden is gone.
- gh-based create/promote: the injected notice (one constant, two delivery moments — spawn and steer).
- **The skill is deleted.** Its stated reason to exist — *"an operator who asks for an epic mid-session gets no injected epic-shape guidance at all"* — has been false since `composeEpicSteer` (#1405) injected exactly that, on both providers, in every repo, where the skill was Claude-only, model-invoked, and shipped only inside the Shepherd repo. Its `test/epic-parse.test.ts` pin goes with it; `EPIC_SHAPE_CONTRACT`'s own pin (the text agents actually receive) stays.

**The compensating PTY notice stays, and here is the justification the issue asks for:** post-deletion it is not a third rendering compensating for a model-invoked skill — it is *the* rendering, and the only epic channel that reaches Codex and non-Shepherd repos at all.

## Stated gaps and scope

- **Codex keeps the prose directive** (now tokenless) and drives the queue exactly as before. Its MCP support (`config.toml` `[mcp_servers.*]`, streamable HTTP behind an experimental client flag) is unverified on the `-c` override path Shepherd would need, and Codex receives directives inline rather than via a flag. `buildQueueDirective`'s `agentProvider` is deliberately **required**, not defaulted — a silent "claude" default would hand a Codex session tool names it has no tools for.
- `SINGLE_PR_INVARIANT`'s embedded epic-shape contract (~800 chars on every code spawn) is left alone — that block is #2002's scope.
- An `epic_import` / `epic_link` tool that would let the server write the marker and wire native links (deleting the marker-grammar prose outright) is the logical next slice; it adds outward GitHub writes, so it is not in this cut.

## Merge note for whoever lands the epic branch

`main` has since landed #2010, which *edited* the epic-authoring skill and added `test/skill-agent-reachability.test.ts` pinning both shipped skills' import text. This PR *deletes* that skill, so epic→main will surface a delete/modify conflict. Resolution: **take the deletion**, and narrow `SKILLS` in that test to `["shepherd-onboarding"]` — the onboarding half still needs the pin. Worth noting #2010 is corroboration, not contradiction: the skill was still teaching a control-plane call agents cannot reach, which is the defect this slice removes structurally.

## Verification

`bun test ./test` 8,170 pass / 0 fail · `bunx tsc --noEmit` clean · `bun run lint` clean · `prettier --check` clean · `fallow audit --base origin/main --fail-on-issues` ✓ (both new functions were refactored under the complexity gate; `isAgentIngressRoute` is now a `Map`-driven table — a `Map`, not an object, so a hostile sub-segment like `constructor` can't resolve through the prototype chain) · branch hygiene ✓ · glossary ✓. No UI changes, no user-facing UX → `[no-feature-entry]`.


scoop:
## What I verified

**The vertical cut itself holds together.** The appliers are a genuine shared seam, not a parallel implementation: `putBuildQueue` / `postBuildStepStatus` / `putEpicDraft` (`src/server.ts:3573`, `3585`, `3660`) are now one-liners over `applyQueueWrite` / `applyQueueStep` / `applyEpicDraft` (`src/agent-control.ts:60`, `81`, `115`), and I compared the moved bodies line-for-line against the pre-image in `git diff` — the 400/404/409 statuses, the `matches` payload, the ambiguous-prefix message, the `queue:update` / `session:epic-draft` emits and the frozen-after-`draft` guard all survive verbatim. The only intentional wording change is `"from the PUT/GET response"` → `"from the queue"` in the 404 (transport-neutral, correct for both callers).

**Identifier resolution.** Every symbol the new module imports exists and is spelled as used: `BUILD_STEP_STATUSES` (`src/validate.ts:1097`, `["pending","active","done","skipped"]` — matches the `queue_step` enum at `src/agent-control.ts:175`), `validateBuildSteps` (`src/validate.ts:1137`), `validateBuildStepStatus` (`src/validate.ts:1158`), `validateEpicDraftBody` (`src/validate.ts:1219`), `validateEpicDraft` (`src/epic-author.ts`), `store.resolveStepId` (`src/store.ts:5167`), `store.replaceBuildQueue` (`src/store.ts:5100`), `store.setBuildStepStatus` (`src/store.ts:5195`). The three imports dropped from `src/server.ts` are genuinely unused there now (`validateEpicDraft` is still needed and still imported — used at `src/server.ts:3695`). No import cycle: `agent-control` → `epic-author` → `epic-import` → `epic-parse`, none of which reach `service`/`server` (`src/epic-author.ts:1-3`, `src/epic-import.ts:1-2`).

**Schema/description claims are true against the store.** `queue_step`'s "Advancing a step auto-completes earlier pending ones" matches the forward-fill at `src/store.ts:5205-5216` (`position < ? AND status = 'pending'`, never touching `skipped`). `queue_write`'s "completed steps keep their status server-side" matches `src/store.ts:5136-5149` (explicit id → `prior.status` carried). Caps agree: `maxItems: 100` vs `STEPS_MAX = 100` (`src/validate.ts:1102`), `maxItems: 50` vs `EPIC_DRAFT_CHILDREN_MAX = 50` (`src/validate.ts:1168`).

**Routing + ingress.** `handleSessionMcp` (`src/server.ts:3768`) sits before `handleSessions` in `ROUTE_HANDLERS` (`src/server.ts:7572` vs `7576`), so nothing upstream claims `…/mcp`; it returns `null` for non-POST, so GET/DELETE fall through to the `/api` 404 rather than being swallowed. The `isAgentIngressRoute` rewrite is behaviour-preserving — I walked each pre-image branch against the `AGENT_LEAF_ROUTES` map (`src/server.ts:7690-7709`): same method sets, same exact-path requirement, `/queue/approve` and `/epic-draft/approve` still excluded (`parts[4]` non-empty and not `steps`), and the `Map` avoids the prototype-chain hazard the comment names. `checkOrigin` still runs on the ingress delegation (`src/server.ts:7645`, `7726`), so dropping `requireJsonContentType` does not open a CSRF hole beyond the existing hooks precedent (`src/server.ts:3613`).

**No token in prompt text.** `config.token` now appears in `src/service.ts` only at `2850` and `2961`, both feeding `buildHooksFragment`'s `--settings` placeholder — never in composed prompt text on either provider. The issue's hardest acceptance criterion is met.

**Wiring.** `--mcp-config` rides both Claude argv builders (`src/service.ts:2855`, `2964`) and only those two — the Codex builders are untouched, which matches the stated gap. `agentMcpConfigArg` reuses the already-resolved `baseUrl` from `resolveSpawnBaseUrl` (`src/service.ts:2265-2283`), so slirp `10.0.2.2` vs host loopback resolve exactly as hooks do. Deleting the skill leaves no dangling reference: nothing under `src/`, `.claude/`, `scripts/` or `package.json` names it (only `CHANGELOG.md` and `docs/research/*`, both historical), and the gh-based epic flow still has its home in `EPIC_AUTHORING_NOTICE` (`src/service.ts:872-880`), whose fence remains covered by the surviving `EPIC_SHAPE_CONTRACT` describe block in `test/epic-parse.test.ts`.

## The blocking problem

The capability gate that decides which tools a session sees is **not** the gate the prompt path uses. `trailingBlocks` deliberately suppresses `<build-queue>` for research, epic-authoring and landing-repair sessions — "a research OR epic-authoring session authors no queue — suppress it there too" (`src/service.ts:1567-1570`) — and `shouldPreApproveBuildQueue` carries the same exclusion (`src/service.ts:1835`). But `buildSpawnArgv` computes the tool catalog from the repo flag alone (`src/service.ts:2855-2858`), and `sessionCapabilities` does the same on resume (`src/agent-control.ts:255-262`). So in a `buildQueueEnabled` repo a research session now receives `queue_write` / `queue_step`, whose descriptions read "Author or replace this session's build queue … Write it before starting work" (`src/agent-control.ts:151-156`) — re-introducing, through the tool surface, exactly the instruction the prompt path removes, and with no curation-gate sentence anywhere (that line lives only in the suppressed `<build-queue>` block). Reachable today, on an existing path, with all three flags available at both call sites (`CreateSessionInput` at `src/types.ts:234ff`; persisted `research` / `epicAuthoring` / `landingRepair` at `src/types.ts:76,80,83`).

## Could not verify

- That Claude Code 2.1.220 accepts `--mcp-config` with a stateless `type: "http"` server in an **interactive PTY** spawn (the plan's own spike used `claude -p`), and that plain-JSON JSON-RPC responses with no `Mcp-Session-Id` satisfy its client. The ground truth is the CLI, not this repo; assumed as written per the module header. The plan's step 7 (live session verification) is the right place to close this — the PR body should record the observed tool calls.
- The rebaselined `10,577` in `test/prompt-budget.test.ts:162` — CI owns that.

## Nits (non-blocking):
- `src/service.ts:1147` names the tools bare ("`queue_write` … on the `shepherd` MCP server") while the doc comment at `1100` uses the namespaced `mcp__shepherd__queue_write`; both denote the same tool, but the prompt could use the form the model actually sees in its tool list.
- `src/agent-control.ts:344` drops the `matches` array on the ambiguous-prefix path (the REST renderer keeps it, `src/server.ts:3570`); the message still carries the count, but the candidate ids would let the model self-correct in one hop.
- `src/agent-control.ts:280` returns HTTP 200 for a `-32600` on an unparseable body (`req.json().catch(() => null)` at `src/server.ts:3777`); a 400 would be more honest, though JSON-RPC-over-HTTP tolerates it.

- `src/service.ts:1116-1122` accepts that a token-set deployment spawning before the ingress listener is up (`agentIngressPort?.()` undefined → `agentBaseUrl()`, `src/service.ts:2276-2277`) now has *no* working queue path on either provider, where the old curl's bearer header would have authenticated against the gated main port. Narrow (startup window only), explicitly documented, and mandated by the issue's "no token in prompt text" — but it is strictly worse than the hooks degradation it is compared to, since hooks still interpolate `$SHEPHERD_TOKEN` outside the autonomous profile.

- `src/service.ts:448-450` claims a token-less hook "→ the ingress listener 401s", but `makeAgentIngressApp` builds with `skipAuth` (`src/server.ts:7719-7720`) so `checkAuth` never runs there — this predates the PR, but it sits in the same paragraph as finding 2 and is worth correcting in the same pass.

scoop:
## What the PR does

Adds `src/agent-control.ts`: transport-neutral appliers (`applyQueueWrite`/`applyQueueStep`/`applyEpicDraft`), a session-gated MCP tool catalog (`queue_write`/`queue_step`/`epic_draft`), and a stateless streamable-HTTP JSON-RPC handler. `src/server.ts` registers `POST /api/sessions/:id/mcp`, rewrites the three REST handlers as renderers over the shared appliers, and replaces the `isAgentIngressRoute` if-chain with an `AGENT_LEAF_ROUTES` Map. `src/service.ts` pushes `--mcp-config` on the Claude spawn and Claude resume, and both directives lose the bearer token.

## Verification performed

- **Token removal (acceptance: "no control-plane token appears anywhere in prompt text").** Confirmed: `token` is gone from both directive signatures (`src/service.ts:1002-1006`, `src/service.ts:1124-1131`) and from both call sites (`src/service.ts:2762-2766`, `src/service.ts:2778-2782`). Grepping `Bearer` across `src/*.ts` leaves only `src/service.ts:469` (the `$SHEPHERD_TOKEN` placeholder in `buildHooksFragment`, which rides `--settings`, not prompt text), `src/service.ts:1114` (a comment), `src/recap.ts:223` and `src/validate.ts:851` (both unrelated).
- **Applier/REST parity.** `putBuildQueue` (`src/server.ts:3573`), `postBuildStepStatus` (`src/server.ts:3586`) and `putEpicDraft` (`src/server.ts:3663`) now delegate; `jsonApplied` (`src/server.ts:3571`) reproduces the prior 400/404/409 statuses and the `matches` payload. Compared against the deleted inline bodies in the diff — the only behavioural delta is the 404 text (`"…from the queue"` vs `"…from the PUT/GET response"`), which nothing asserts (grep for `PUT/GET response` matches only `src/service.ts:1167`, the Codex prose).
- **Tool schemas vs. validators.** `queue_write.steps maxItems: 100` matches `STEPS_MAX` (`src/validate.ts:1102`); `epic_draft.children maxItems: 50` matches `EPIC_DRAFT_CHILDREN_MAX` (`src/validate.ts:1168`); `queue_step.status` spreads `BUILD_STEP_STATUSES` (`src/validate.ts:1097`) so the enum cannot drift from `validateBuildStepStatus` (`src/validate.ts:1158`). Argument shapes line up: the tool passes the whole `arguments` record, and `validateBuildSteps` reads `o.steps` (`src/validate.ts:1139-1140`), `validateEpicDraftBody` reads `o.parent`/`o.children` (`src/validate.ts:1221-1224`), `validateBuildStepStatus` reads `o.status` (`src/validate.ts:1160`).
- **Tool description claims vs. store behaviour.** `queue_step`'s "advancing a step auto-completes earlier pending ones" is true (`src/store.ts:5205-5215`); `queue_write`'s "completed steps keep their status server-side" is true (`src/store.ts:5098-5110`). `queue_step`'s ≥8-char-prefix error text matches `resolveStepId` (`src/store.ts:1081-1088`).
- **Ingress allowlist refactor.** The Map form is behaviour-preserving: `parts[3] ?? ""` yields `""` for `/api/sessions/<id>`, `Map.get("")` → `undefined` → `[]` → false, and the deeper-route branch still requires `sub === "queue" && parts[4] === "steps" && parts[5] && !parts[6]` (`src/server.ts:7703-7710` vs. the deleted chain in the diff). MCP is POST-only and exact-path.
- **Route ordering.** `handleSessionMcp` is registered before `handleSessions` (`src/server.ts:7572-7576`), so it is reachable; it 404s an unknown session before `handleMcpRequest` runs.
- **Privilege.** The MCP surface is strictly weaker than the REST routes already on the auth-exempt ingress: `epic_draft` is gated on `session.epicAuthoring` and the queue tools on `buildQueueEnabled` (`src/agent-control.ts:255-262`), while `PUT …/epic-draft` and `PUT …/queue` are allowed for any session. Neither `/queue/approve` nor `/epic-draft/approve` has a tool. The main app still runs `checkAuth` then `checkOrigin` (`src/server.ts:7641-7646`), and the ingress app still runs `checkOrigin` via delegation (`src/server.ts:7721-7727`), so the missing `requireJsonContentType` on `/mcp` does not open a CSRF hole — same posture as `handleSessionHooks` (`src/server.ts:3613-3618`).
- **Types.** `AgentControlDeps.events?.emit(event: string, data: unknown)` is satisfied by `EventHub.emit` (`src/events.ts:9`) and by `ServiceDeps.events` (`src/service.ts:177`). Removed imports (`validateBuildSteps`, `validateBuildStepStatus`, `validateEpicDraftBody`) have no remaining uses in `src/server.ts`; `validateEpicDraft` is still used at `src/server.ts:3695`.
- **Skill deletion.** No live reference to `.claude/skills/shepherd-epic-authoring` remains — only `CHANGELOG.md:360`, two `docs/research/*.md` files, and the explanatory comment at `src/service.ts:901`. `CONTEXT_TRIM_NOTICE` (`src/service.ts:684-691`) enumerates no skill names, and `test/epic-parse.test.ts` drops the block that read the file. `EPIC_AUTHORING_NOTICE` (`src/service.ts:872-880`) remains the single gh-flow rendering, delivered at spawn and via `composeEpicSteer` (`src/service.ts:908-911`) — the justification the issue asked for.
- **Codex gap.** `pushAgentMcpFlag` is called only from `buildClaudeSpawnArgv` (`src/service.ts:2855`) and `buildClaudeResumeArgv` (`src/service.ts:2964`); `buildCodexSpawnArgv`/`buildCodexResumeArgv` are untouched, and `buildQueueDirective` keeps the full HTTP prose for `agentProvider === "codex"` (`src/service.ts:1153-1200`). The gap is real and stated.

## Blocking

One re-raised finding survives (see below). It is the same defect flagged last round and the code at `src/service.ts:2855-2858` and `src/agent-control.ts:255-262` is byte-identical to what was flagged.

## Cannot verify (stated limitation)

- Claude Code's MCP HTTP client behaviour (that it accepts a plain `application/json` JSON-RPC response with no SSE stream and no `Mcp-Session-Id`, that it tolerates a 404 on `GET`/`DELETE` of the endpoint, and that `--mcp-config` accepts an inline JSON string) is not verifiable from this repo. The plan records an empirical spike against 2.1.220; I assumed it as written and did not treat it as a finding either way.

## Nits (non-blocking):
- `src/service.ts:1147` and `src/service.ts:1029`: the prompt names the tools bare (`queue_write`, `epic_draft`) while the model's tool list will show them namespaced (`mcp__shepherd__queue_write`); the in-code comments use the namespaced form. Harmless, but one consistent form in the prompt would remove the reconciliation the slice is trying to eliminate.
- `src/service.ts:1114-1122`: the documented pathological fallback is now slightly worse than before for token-set deployments — when `agentIngressPort` is unknown the old curl carried the token and would have been accepted by the gated main port, whereas the tokenless MCP call 401s. It is explicitly acknowledged in the comment and the plan's risk table, so noting only.
- `src/agent-control.ts:132-137`: `queue_step`'s description says "Returns the full queue" — accurate, but the whole queue JSON on every transition is a lot of tool-result context for a signature-first design.

- `src/server.ts`'s agent-ingress listener authenticates purely on the unguessable session id in the path; any same-UID local process can drive another session's control plane if it learns an id (pre-existing #1079 design, unchanged here).

scoop:
## Prior findings

**1. Non-code-mode suppression of the queue tool catalog — ADDRESSED.** `isNonCodeMode` is now a single exported definition (`src/agent-control.ts:254-260`) consumed by *both* paths I named:
- prompt path: `composeSystemPromptBlocks` calls it at `src/service.ts:1693`, feeding the `<build-queue>` gate at `src/service.ts:1570-1571`;
- spawn catalog: `buildSpawnArgv` now passes `buildQueue: repoConfig.buildQueueEnabled && !isNonCodeMode(input)` (`src/service.ts:2858-2861`) — off the create inputs, correct because the row does not exist yet;
- resume catalog: `sessionCapabilities` ANDs the same predicate against the persisted row (`src/agent-control.ts:281-285`), reached from `buildClaudeResumeArgv` at `src/service.ts:2968`.

Field names resolve on both shapes: `CreateSessionInput.research/epicAuthoring/landingRepair` at `src/types.ts:256-260`, `Session.research/epicAuthoring/landingRepair` at `src/types.ts:75-83`. The truthiness typing (`unknown`) is deliberate and correct — `ComposeSystemPromptOptions.epicAuthoring` is the pre-baked directive STRING while the row carries a boolean; both are covered.

**2. `buildHooksFragment` doc comment — NOT ADDRESSED, re-raised.** `src/service.ts:455-457` still reads "This is asymmetric with the build-queue curl (buildQueueDirective), which DOES carry the literal token over the same ingress transport: that exposure is pre-existing, opt-in (only when `buildQueueEnabled`), and already documented as accepted there." I verified the claim is false against the current code: `buildQueueDirective` (`src/service.ts:1125-1200`) has no `token` parameter and neither branch emits an `Authorization` header (the Claude branch at `src/service.ts:1143-1152`, the Codex curl block at `src/service.ts:1154-1200`), and the only surviving `config.token` reads in the file are the two hooks-fragment call sites (`src/service.ts:2851`, `src/service.ts:2965`). The comment now documents an accepted token exposure that does not exist; the placeholder is the only remaining token carrier.

- **No token in prompt text (issue acceptance):** `grep -n "config.token" src/service.ts` returns only `src/service.ts:2851` and `src/service.ts:2965`, both `hooks: {…}` overlays that emit the `$SHEPHERD_TOKEN` placeholder (`src/service.ts:467-473`). `epicAuthoringDirective` also lost its `token` param (`src/service.ts:1003-1007`) and both call sites were updated (`src/service.ts:2763-2770`, `src/service.ts:2779-2783`). Test callers agree — `test/prompt-budget.test.ts:148-153`, `test/service.test.ts:5252-5256/5271-5275/5295-5300/7399+` all pass `agentProvider` and no `token`, so the now-required `agentProvider` parameter has no stale caller.
- **Applier/REST parity:** `putBuildQueue`, `postBuildStepStatus` and `putEpicDraft` (`src/server.ts:3576-3596`, `src/server.ts:3660-3666`) delegate to `applyQueueWrite`/`applyQueueStep`/`applyEpicDraft` (`src/agent-control.ts:58-123`); the 400/404/409 statuses, the `matches` payload and the `queue:update` / `session:epic-draft` emits are preserved by `jsonApplied` (`src/server.ts:3571-3575`). Only the 404 wording changed ("from the queue" vs "from the PUT/GET response") — behaviour identical.
- **Validators tolerate the tool-argument shape:** `applyQueueStep` passes the whole `arguments` record as the body; `validateBuildStepStatus` (`src/validate.ts:1158-1163`) reads only `o.status` and ignores the extra `stepId`. `validateBuildSteps` (`src/validate.ts:1137-1155`) and `validateEpicDraftBody` (`src/validate.ts:1219-1232`) likewise key off named fields, so no extra-property rejection.
- **Ingress allowlist refactor is behaviour-preserving:** `isAgentIngressRoute` (`src/server.ts:7703-7710`) with `AGENT_LEAF_ROUTES` (`src/server.ts:7690-7695`) reproduces the previous four rules exactly — `hooks:POST`, `queue:PUT|GET`, `epic-draft:PUT|GET`, plus the deeper `POST …/queue/steps/<id>` — and adds only `mcp:POST`. `/queue/approve` and `/epic-draft/approve` remain excluded (they carry `parts[4]`, and `approve` is not a map key). `parts[3] ?? ""` + `Map.get` means an absent or hostile sub-segment yields `[]`.
- **MCP route safety:** `handleSessionMcp` (`src/server.ts:3758-3776`) is POST-only, exact-path, 404s an unknown session, and is registered at `src/server.ts:7572`. It runs behind the same `checkOrigin` CSRF guard every other write does (`src/server.ts:7645-7646`, applied to POST at `src/server.ts:770`), which is what compensates for the deliberate absence of `requireJsonContentType`. Auth exemption matches the sibling queue/epic-draft routes and the tool catalog is session-gated (`src/agent-control.ts:292-294`), with an un-entitled `tools/call` rejected as `-32602` (`src/agent-control.ts:358-359`).
- **Identifier consistency:** the MCP server name `shepherd` in `agentMcpConfigArg` (`src/agent-control.ts:416-420`) matches `SERVER_INFO.name` (`src/agent-control.ts:303`) and the prose's "`shepherd` MCP server" (`src/service.ts:1149`, `src/service.ts:1029`). Tool names `queue_write` / `queue_step` / `epic_draft` (`src/agent-control.ts:150,165,204`) match the dispatch arms (`src/agent-control.ts:361-365`) and the directive text (`src/service.ts:1148`, `src/service.ts:1029`). `BUILD_STEP_STATUSES` is the shared source of the status enum (`src/validate.ts:1097` → `src/agent-control.ts:175`), so the tool enum cannot drift from the validator.
- **Deleted skill leaves no dangling reference:** `grep -rn "shepherd-epic-authoring"` finds only CHANGELOG/research prose and the descriptive comment at `src/service.ts:902`; `test/epic-parse.test.ts` dropped the block that `readFileSync`'d the now-missing SKILL.md.
- **Non-blocking dependency I could not verify:** Claude Code's acceptance of a plain-JSON streamable-HTTP MCP server and its `mcp__shepherd__*` namespacing are external to this repo — the plan records a live spike; I took that as written and did not treat it as a finding either way.

- src/server.ts: `isAgentIngressRoute` was rewritten from four explicit `if`s into the `AGENT_LEAF_ROUTES` map when only one entry (`mcp`) needed adding — verified equivalent, but it is a drive-by refactor of a security-sensitive allowlist the task did not require touching.

## Nits (non-blocking):
- src/agent-control.ts: `QUEUE_WRITE`'s `maxItems: 100` and `EPIC_DRAFT`'s `maxItems: 50` duplicate `STEPS_MAX` / `EPIC_DRAFT_CHILDREN_MAX` (src/validate.ts:1102, src/validate.ts:1168) as literals rather than importing them, so the schema hint and the server cap can drift.
- src/service.ts: the Claude branch names the bare tool (`queue_write`) while the doc comment above it uses the namespaced form (`mcp__shepherd__queue_write`); both are resolvable given the "on the `shepherd` MCP server" qualifier, but one form throughout would read cleaner.

scoop:
Both findings from the last two rounds are already fixed on the branch — the review raced the pushes.

- **Non-code-mode tool gate** → `228560b1`. `isNonCodeMode` is now one shared definition used by both the prompt path (`composeSystemPromptBlocks`) and the tool catalog, applied at spawn (create inputs — the row does not exist yet) and on resume (persisted `research`/`epicAuthoring`/`landingRepair`). Tests cover all three modes at both call sites.
- **`buildHooksFragment` doc comment** → `fa025213`. The stale "asymmetric with the build-queue curl … accepted there" paragraph is replaced with: the `$SHEPHERD_TOKEN` placeholder is now the only place a Shepherd spawn carries the control-plane token at all.

Current head is `fa025213`.

scoop (commented):
## Re-review of the prior round's finding

**Prior finding 1 — `buildHooksFragment`'s doc comment claimed an accepted build-queue token exposure that no longer exists.** **Fixed.** `src/service.ts:455-460` now reads "This placeholder is now the ONLY place a Shepherd spawn carries the control-plane token at all (issue #2003): the build-queue and epic-authoring directives used to bake the literal token into their `curl` lines, and no longer do". I verified the claim rather than taking it on the note's say-so: `config.token` appears in `src/service.ts` only at the two `spawnSettingsOverlay({... hooks: {sessionId, baseUrl, token: config.token}})` call sites (`src/service.ts:2851`, `src/service.ts:2965`) plus the `SpawnTrimOverlay`/`buildHooksFragment` signatures (`src/service.ts:412`, `src/service.ts:465`) — no prompt path composes it, and both directive builders lost the `token` parameter (`src/service.ts:1003-1007`, `src/service.ts:1125-1132`). The comment is now accurate.

## What I verified in this diff

- **Shared appliers can't drift from the REST routes.** `applyQueueWrite` / `applyQueueStep` / `applyEpicDraft` (`src/agent-control.ts:47-129`) reproduce the previous handler bodies exactly — same validators, same event names, same 400/404/409 semantics — and `putBuildQueue` / `postBuildStepStatus` / `putEpicDraft` (`src/server.ts:3577-3667`) now render them via `jsonApplied`, which preserves the `{error, matches?}` body shape. Compared against the deleted originals in the diff hunk and against the validators at `src/validate.ts:1137-1161` and `src/validate.ts:1219-1233`: all three tolerate extra keys, so passing the tool's `arguments` object (which carries `stepId` alongside `status`) through `validateBuildStepStatus` is safe.
- **Tool descriptions state behaviour the store actually implements.** `queue_step`'s "Advancing a step auto-completes earlier pending ones" matches the forward-fill in `src/store.ts:5204-5216` (`status IN (active,done)` → earlier `pending` rows set `done`). `queue_write`'s "completed steps keep their status server-side" matches the id-keyed status/`createdAt` carry-over in `src/store.ts:5100-5150`. Schema caps line up with the validators: `maxItems: 100` vs `STEPS_MAX = 100` (`src/validate.ts:1102`), children `maxItems: 50` vs `EPIC_DRAFT_CHILDREN_MAX = 50` (`src/validate.ts:1166`), and `status.enum` is spread from the single `BUILD_STEP_STATUSES` constant (`src/validate.ts:1097`), so the enum cannot drift from the validator.
- **Prompt/tool gating is symmetric.** `<build-queue>` is emitted only when `!nonCodeMode && opts.buildQueue != null` (`src/service.ts:1570`, `nonCodeMode` from `isNonCodeMode(opts)` at `src/service.ts:1693`), and the spawn wiring uses the same predicate off the create inputs (`src/service.ts:2859-2862`) while the request path uses it off the persisted row (`src/agent-control.ts:281-289`). `store.create` persists `research`/`epicAuthoring`/`landingRepair` verbatim (`src/service.ts:3437-3439`), so spawn-time and resume-time capabilities agree. This is the previous round's other finding, and it is genuinely closed.
- **`--mcp-config` is Claude-only and rides both spawn and resume.** Pushed at `src/service.ts:2859` (`buildSpawnArgv`, selected only for the non-codex branch at `src/service.ts:3295-3313`) and `src/service.ts:2968` (`buildClaudeResumeArgv`). `spawnSettingsOverlay` has exactly two call sites in `src/`, both of which now also push the flag, so there is no third Claude spawn path that would silently lose the tools. Codex spawn (`src/service.ts:2914-2923`) passes `agentProvider: "codex"` and gets no MCP flag — the stated gap, and `buildQueueDirective`'s codex branch (`src/service.ts:1154-1200`) keeps the full HTTP prose, now tokenless.
- **Ingress allowlist refactor is behaviour-preserving.** The `AGENT_LEAF_ROUTES` Map (`src/server.ts:7690-7695`) plus the deeper `queue/steps/<id>` clause reproduces the old four-clause chain exactly for every case I enumerated, including the trailing-segment denials; `parts` is `split("/").filter(Boolean)` so empty segments can't sneak through, and using a `Map` (rather than an object) closes the `constructor`/`__proto__` sub-segment hole. `POST …/mcp` is added; both approve gates remain excluded (`src/server.ts:7702-7712`).
- **No CSRF regression from skipping `requireJsonContentType` on `/mcp`.** `makeApp` runs `checkOrigin` on every POST before dispatch (`src/server.ts:7645-7646`), and `makeAgentIngressApp` delegates through it (`src/server.ts:7726`), so the classic `text/plain` form-POST trick is rejected with 403 on both listeners. This matches the pre-existing rationale on the hooks ingest route (`src/server.ts:3613`).
- **Skill deletion leaves nothing dangling.** No source, test, doc-site or plugin manifest references `.claude/skills/shepherd-epic-authoring` — only `CHANGELOG.md:360` and two `docs/research/*` retrospectives, which are historical records. The removed `epic-parse` tests were the only code that read the file. The guidance the skill and the deleted prose carried (no epic-dag fence in `parent.body`, `key` as a temp id for `blockedBy`) is preserved in the `epic_draft` tool description and child schema (`src/agent-control.ts:171-215`).
- **Base-delta check.** `src/store.ts` and `src/server.ts` are in the base delta, so I read the base blobs rather than trusting my worktree: every store method this PR's new module calls exists at `epic/2005-epic-adopt-the-claude-5-context-engineer@697b4a4939c1101ee302dc886e95ac278775b4b0:src/store.ts` (`get` 2367, `getRepoConfig` 1649, `getBuildQueue` 5048, `replaceBuildQueue` 5099, `resolveStepId` 5168, `setBuildStepStatus` 5196, `getEpicDraft` 5235, `replaceEpicDraft` 5267, `setEpicDraftApproved` 5345), and `putBuildQueue` / `putEpicDraft` / `ROUTE_HANDLERS` / `isAgentIngressRoute` / `APPROVE_STEER` all still exist at `epic/2005-epic-adopt-the-claude-5-context-engineer@697b4a4939c1101ee302dc886e95ac278775b4b0:src/server.ts`. No merged sibling removed or renamed anything this child depends on.

Acceptance against the task: queue transitions and draft submission are tool calls; the Claude `<build-queue>` block drops to well under 600 chars and the kitchen-sink meter is rebaselined 13,600 → 10,577 (`test/prompt-budget.test.ts:160`); no token reaches prompt text on any path; epic authoring has one home per flow with the skill deleted; the Codex gap is stated in code and directive.

## Stated limitations (could not verify — ground truth not in this repo)

- Claude Code's `--mcp-config` streamable-HTTP client behaviour (plain `application/json` JSON-RPC with no `Mcp-Session-Id`/SSE, and whether it re-lists tools after `initialize`) is external to this repo; I could not confirm it against a live client and assumed it as written per the plan's spike. The same applies to the exact `mcp__shepherd__*` name-mangling convention.

## Nits (non-blocking):

- `src/service.ts:1145-1147` and `src/service.ts:1029-1031`: the directives name the bare tool names (`queue_write`, `queue_step`, `epic_draft`) plus "on the `shepherd` MCP server", while the code comment at `src/service.ts:1101-1102` uses the fully-qualified `mcp__shepherd__queue_write` form the model actually sees; naming the qualified form in the prompt would remove the one resolution step left to the model.
- `src/service.ts:2859`: `create()` launches the agent via `prepareSpawnOrThrow` (`src/service.ts:3402`) *before* `store.create()` (`src/service.ts:3413`), so an `initialize`/`tools/list` arriving in that window would hit the `session not found` 404 / an empty catalog, and the server advertises `listChanged: false` (`src/agent-control.ts:315`) so a client that caches an empty list never recovers. In practice the two are separated by one synchronous block versus CLI boot time, so it is not reachable — noting it only because the new endpoint is the first spawn-time dependency on the row existing.
- `src/server.ts:3765`: an unknown session answers with a plain `{error}` 404 rather than a JSON-RPC error object, so the client sees a transport failure instead of a protocol one; defensible either way.
- `src/agent-control.ts:174`: `queue_write`'s step schema sets `additionalProperties: false`, which forbids the `status` field the applier still accepts (`src/validate.ts:1128-1131`) — intentional narrowing, but the tool path and the REST path now differ slightly in what a caller may send.

- `src/server.ts:7690-7712`: rewriting `isAgentIngressRoute` into an `AGENT_LEAF_ROUTES` Map is a drive-by refactor; adding one `mcp` clause to the existing chain would have sufficed (the rewrite is behaviour-preserving, and the prototype-chain argument in its comment is a fair reason to keep it).
- `src/agent-control.ts:344-346`: a `ping` handler and the `KNOWN_PROTOCOL_VERSIONS` negotiation table are more protocol surface than the two flows require, though both are cheap and spec-conformant.

PR #2017, opened 2026-08-03
feat(readiness): stop generating posture prose into target repos

### LINKED ISSUE #2015: Stop the readiness analyzer generating posture prose into target repos
## Goal

Stop Shepherd's **AI-readiness analyzer** from manufacturing the same imperative behavioral
prose the rest of this epic is deleting — and from writing it into *other people's* repos,
where it lands in the one place that is re-read on every turn of every session.

The scorecard itself is not the problem. Its twelve detectors (`formatter`, `linter`,
`type_checker`, `commit_lint`, `git_hooks`, `pre_push_ci`, `lint_staged`, `test_runner`,
`dead_code_audit`, `ci`, `dependency_automation`, `agent_instructions`) find *deterministic
guardrails* — the complement of what this epic cuts. A gate that fails and self-explains is
exactly what lets you delete the prose telling the model to be careful. **No change to
detection or scoring is in scope here.**

## The duplication

`src/readiness.ts` → `generateClaudeMd()` emits a **1,186-char** `## Engineering posture
(surgical & mechanical)` section into the `CLAUDE.md` it generates for every target repo:

- Simplicity first
- Surgical changes
- Fail closed
- Single source of truth
- Keep names and comments honest
- No dead code
- Verify before claiming done

`src/service.ts` → `ENGINEERING_POSTURE` is a **1,643-char** block that
`composeSystemPrompt()` injects into **every** spawn unconditionally — including spawns
against those same target repos. Three of the seven bullets above (*simplicity first*,
*surgical changes*, and *verify before claiming done* as "goal-driven execution") are the
system-prompt block restated in the repo file.

So the posture is stated twice today. **#2002 deletes the spawn-side copy** — which, left
alone, does not remove the statement, it just relocates it from a block Shepherd can delete
at will into a file every session re-reads on every turn, in every repo Shepherd ever
analyzed. That is the wrong direction, and it happens by default unless this slice lands.

Two more signals that this is the epic's own failure mode:

- **Shepherd already forbids exactly this, and exempts readiness by name.**
  `.claude/skills/shepherd-onboarding/SKILL.md` tells operators: *"Don't restate what
  Shepherd injects at spawn. Every Shepherd-driven agent already receives an engineering
  posture… Do not reproduce generic posture/process prose — it adds tokens and says nothing
  new."* It then hands the `# House rules for AI agents` heading to the readiness analyzer
  as the owner of that artifact — the one generator not held to the rule.
- **#2004's rejection test applies verbatim.** That slice re-targets the learnings flywheel
  away from "rules Claude 5 follows by judgement" toward non-derivable facts. The readiness
  generator is the second, unlisted producer of the same artifact, and nothing in #2004
  reaches it.

`ui/src/lib/components/readiness-view.ts` → `buildAdoptPrompt()` concatenates an i18n'd
intro with the snippet verbatim and sends it as the one-click install task. So a Shepherd
spawn — already carrying `ENGINEERING_POSTURE` — is handed instructions to write the
posture into the repo. The install task needs no separate fix: it passes the artifact
through, so it inherits whatever this slice leaves.

## Vertical cut

Cut the generated posture section to its **non-derivable** bullets. Not deleted wholesale
(the artifact must still stand alone for an agent that is not a Shepherd spawn), and not
kept as-is.

Proposed split — **a proposal, not a verdict**. #2004 is where the "would a capable model
do this unprompted?" rejection test is written down; apply *that* test here rather than
inventing a parallel one, and record any divergence from this list in the PR:

| Bullet | Proposal | Why |
|---|---|---|
| Fail closed | **keep** | Agents genuinely do let a swallowed error, empty result or zero count read as success |
| Single source of truth | **keep** | Hardcoded counts/dimensions that silently drift are not self-evident |
| Keep names and comments honest | **keep** | A stale doc comment after a behavior change survives review routinely |
| Simplicity first | **drop** | Stated in the Claude Code system prompt; followed unprompted |
| Surgical changes | **drop** | Same — and it names the section heading, so the heading needs renaming too |
| Verify before claiming done | **drop** | System prompt + default Claude 5 behavior |
| No dead code | **adjudicate** | Restates the `dead_code_audit` guardrail the scorecard already prescribes as *tooling*, which is the stronger form |

What stays untouched in the artifact: the stack label, the leverage-ranked *"Adopt these
guardrails"* prescription with its install commands, and the trailing `.prettierignore`
note. That prescription is the part no model can derive by reading the repo, and it is the
whole point of the file.

## Explicitly not in scope

- **Guardrail detection and scoring.** Unchanged.
- **The `agent_instructions` check stays presence-only.** No byte threshold or "leanness"
  heuristic: size cannot distinguish derivable prose from a non-derivable gotcha, so such a
  check would manufacture false signal — the same mistake in a new place.
- **The onboarding skill's "never reference skills or slash commands" clause.** #2001 owns
  correcting that if the drain trim is reversed.

## Seams

- `src/readiness.ts` → `generateClaudeMd()` — the posture section and its heading.
- `test/readiness.test.ts` → the test *"generated CLAUDE.md is non-empty, encodes the
  surgical posture, and names missing tooling"* asserts `toContain("Surgical")`, binding to
  a bullet slated to drop. **Retarget it to a retained bullet — do not delete the
  assertion**, or the artifact loses its only guard against regenerating empty.
- `ui/src/lib/components/readiness-view.ts` → `buildAdoptPrompt()`, plus
  `readiness_adopt_intro_create` / `readiness_adopt_intro_merge` in **both**
  `ui/messages/en.json` and `ui/messages/de.json` if the intro copy changes (`check:i18n`
  enforces parity).
- `.claude/skills/shepherd-onboarding/SKILL.md` — the clause that exempts the readiness
  artifact by name, if the artifact's shape changes enough to make it inaccurate.

## Acceptance

- The generated `CLAUDE.md` contains no bullet that restates `ENGINEERING_POSTURE`, and the
  retained bullets each survive #2004's documented rejection test.
- The keep/drop decision for every one of the seven bullets is recorded in the PR, with any
  divergence from the table above justified against that test.
- The prescription (stack label, leverage-ranked adopt list with install steps, the
  `.prettierignore` note) is byte-identical to today — this slice touches posture prose only.
- `test/readiness.test.ts` still asserts the artifact is non-empty and names missing
  tooling, via a retained bullet.
- The one-click adopt task produces the trimmed artifact end to end, and the JS/TS and Rust
  profiles both do (the posture section is profile-independent — a regression would hit both).
- `bun run lint`, `bun run test`, and `cd ui && bun run check` pass (`bun run test` scopes to ./test; bare `bun test` sweeps the whole tree).

Blocked by #2004 — that slice writes down the rejection test this one applies. Draining it
first would produce a second, parallel definition of "derivable", which is the say-it-once
failure this epic exists to fix. Deliberately **not** blocked by #2002: cut-not-delete holds
whether or not the spawn-side posture survives.

Part of the Claude 5 context-engineering epic (#2005).

scoop:
Closes #2015.

`generateClaudeMd()` was writing a 1,177-char `## Engineering posture (surgical & mechanical)` section into the `CLAUDE.md` it generates **for other people's repos** — the one file re-read on every turn of every session. Three of its seven bullets restated `ENGINEERING_POSTURE` (`src/service.ts`), which every Shepherd spawn already carries. Cut to the non-derivable bullets: **not deleted wholesale** (the artifact must stand alone for an agent that is not a Shepherd spawn), and not kept as-is. Section is now 527 chars.

## Which rejection test applies (recorded divergence)

#2004 landed its test as `LEARNING_ADMISSION_TEST` in `src/learning-shape.ts`, but it is written for **learnings-flywheel entries** — repo-specific gotchas. Applied verbatim it rejects all seven bullets, including the three the issue proposes keeping, because this artifact is generic by construction: it is generated for a repo Shepherd has only just scanned, with no session history to draw a repo-specific fact from.

So the transferable half is applied and the rest is not:

- **Applied** — REJECT clause 1: *"general engineering judgement or process a capable model already follows unprompted: scope discipline, verify-before-claiming, … don't-build-what-wasn't-asked"*.
- **Applied** — REJECT clause 2: restating a directive the agent is handed every session.
- **Applied** — ADMIT criterion 2, **CONSEQUENTIAL AND QUIET**: the failure is silent, delayed or expensive to undo, and *"the linter catches it"* is a reject.
- **Not applied** — criterion 1 (non-derivable **for this repo**), criterion 3 (attached to a named file/flag/table), and the final REJECT clause (*"broad enough to apply to any repository"*). All three test for repo-specificity, which is a property of a learnings entry, not of a generated house-rules file. Inventing a parallel definition of "derivable" is the failure this epic exists to fix, so the divergence is scoped to exactly these three and recorded here rather than encoded anywhere.

## Verdicts — all seven bullets

| Bullet | Verdict | Justification against the applied test |
|---|---|---|
| Simplicity first | **DROP** | REJECT clause 1 verbatim ("don't-build-what-wasn't-asked"); also restates `ENGINEERING_POSTURE`'s *Simplicity first*. |
| Surgical changes | **DROP** | REJECT clause 1 ("scope discipline"); restates `ENGINEERING_POSTURE`'s *Surgical changes* near-verbatim. |
| Verify before claiming done | **DROP** | REJECT clause 1 ("verify-before-claiming"); restates *Goal-driven execution*. |
| Fail closed | **KEEP** | Consequential and quiet: a swallowed error / empty result / zero count reading as success is silent and survives review. No system-prompt counterpart. |
| Single source of truth | **KEEP** | Consequential and quiet: a hardcoded count or dimension drifts with no gate firing — the wrong number just ships. |
| Keep names and comments honest | **KEEP** | Consequential and quiet: a doc comment left stale by a behavior change misleads the *next* agent, delayed and unlinted. |
| No dead code | **DROP** (adjudicated) | The issue left this open. It restates the `dead_code_audit` guardrail the scorecard already prescribes as **tooling** (fallow / cargo-machete). Guardrail present → the gate catches it mechanically; absent → the adopt list prescribes it. Redundant in both states, and "let deterministic guardrails do the coaching" is the artifact's own stated thesis. |

**No divergence from the issue's proposed table** on the six bullets it decided; `No dead code` adjudicated to DROP.

The heading `## Engineering posture (surgical & mechanical)` named a dropped bullet, so it became **`## What the guardrails can't catch`** — the one property the three survivors share, and it reads straight into the `## Adopt these guardrails` section below it.

## The prescription is byte-identical

Verified mechanically rather than by eye: a throwaway script generated the artifact for a JS/TS and a Rust fixture at the base commit and after the change and diffed them. The diff is confined to the posture section in **both** profiles — stack label, leverage-ranked adopt list, install commands, prescription notes and the trailing `.prettierignore`/profile note come out unchanged. Script deleted; nothing scaffolding-shaped is committed.

Detection and scoring are untouched (no detector, weight or `GUARDRAILS` edit), and `agent_instructions` stays presence-only — no byte threshold or leanness heuristic.

## Tests

- `test/readiness.test.ts` — the artifact's only guard against regenerating empty was `toContain("Surgical")`, bound to a dropped bullet. **Retargeted, not deleted**: now `toContain("Fail closed")`, with the prescription assertion untouched.
- New test asserting the four dropped bullets are absent and the three retained ones present. Without it nothing stops the section re-growing, which is the exact failure this slice exists to prevent.
- `test/readiness-rust.test.ts` — the posture section is profile-independent, so a regression would hit both profiles; the Rust artifact test now asserts the cut too.

## No UI, i18n or skill change needed

- `buildAdoptPrompt()` concatenates the intro with the artifact verbatim, so the one-click adopt task inherits the trim with no code change.
- `readiness_adopt_intro_create` / `_merge` say "install the guardrails below and add a CLAUDE.md with these house rules" — still accurate, so the catalogs are untouched and `check:i18n` parity is unaffected.
- `.claude/skills/shepherd-onboarding/SKILL.md` — both relevant clauses stay true: the artifact keeps the `# House rules for AI agents` heading and its adopt list. (If #2002 deletes the spawn-side posture, that clause's "already receives an engineering posture" is #2002's to correct.)

## Gates

`bun run lint`, `bun run test` (8176 pass / 0 fail), and `cd ui && bun run check` (0 errors) all pass locally, plus the pre-push hook.

[no-feature-entry] — this removes generated prose rather than shipping new UX, and touches no `ui/src/lib/components/**` or `ui/src/routes/**` path.

### DIFF
--- a/src/readiness.ts
+++ b/src/readiness.ts
@@ -653,9 +653,20 @@ export function analyzeReadiness(dir: string): ReadinessReport {
 }
 
 /**
- * A repo-tailored house-rules snippet encoding the surgical/mechanical posture
- * (generalized from Shepherd's own `<shepherd-house-rules>` + Karpathy posture),
- * plus a prescription of the missing tooling. Fully profile-driven: the stack label,
+ * A repo-tailored house-rules snippet: the few failure modes tooling can't catch, plus a
+ * prescription of the missing tooling.
+ *
+ * The posture section is deliberately short (issue #2015). It used to carry seven bullets —
+ * simplicity first, surgical changes, verify before claiming done and no dead code among them —
+ * which either restate the engineering posture every Shepherd spawn already receives in its system
+ * prompt, or state general judgement a capable model follows unprompted. This file is written into
+ * a *target* repo, where it is re-read on every turn of every session, so prose that says nothing
+ * new is pure cost. What survives is what fails QUIETLY: a swallowed error reading as success, a
+ * hardcoded count drifting, a doc comment left stale by a behavior change — none of which a gate
+ * fires on. Anything a gate CAN catch belongs in the prescription below as tooling, not here as
+ * prose. Keep it that way when editing.
+ *
+ * Fully profile-driven: the stack label,
  * all four adopt-line maps (tooling label, churn prose, install steps, prescription
  * note) and the trailing section come from the selected profile, so nothing JS-specific
  * (e.g. a "prettier + lint-staged" churn line) can leak into another stack's artifact.
@@ -689,22 +700,14 @@ with minimal human back-and-forth by having deterministic guardrails do the
 coaching — the gate fails, the agent self-corrects, a human is never pulled in to
 re-explain a mechanical defect.
 
-## Engineering posture (surgical & mechanical)
+## What the guardrails can't catch
 
-- **Simplicity first.** Write the minimum code that solves the stated problem —
-  no speculative features, abstractions for single use, or config nobody asked for.
-- **Surgical changes.** Every changed line traces to the request. Don't refactor,
-  reformat, or polish adjacent code; match existing style. Delete only what your
-  change orphaned; surface pre-existing dead code rather than silently expanding the diff.
 - **Fail closed.** Render error/unreachable paths as explicit failures; never let a
   swallowed error, empty result, or zero count masquerade as success.
 - **Single source of truth.** Derive counts, totals, and dimensions from the data
   (array length, column count) — never hardcode a magic number that silently drifts.
 - **Keep names and comments honest.** When you change a function's behavior, update
   its name, doc comment, and inline comments to match — no stale comment left behind.
-- **No dead code.** Delete unreachable branches and unused fields/params, or wire
-  them to a real path, before opening a PR.
-- **Verify before claiming done.** Run the gate; show the output. Evidence before assertions.
 
 ## Adopt these guardrails (highest leverage first)
 
--- a/test/readiness-rust.test.ts
+++ b/test/readiness-rust.test.ts
@@ -157,6 +157,10 @@ test("generated Rust CLAUDE.md is Rust-correct and free of JS tooling text", () 
   expect(r.claudeMd).not.toContain("lint-staged");
   expect(r.claudeMd).not.toContain("eslint");
   expect(r.claudeMd).not.toContain("tsc");
+  // The posture section is profile-independent (issue #2015), so the cut must hold here too —
+  // a regression would otherwise reach Rust repos unnoticed.
+  expect(r.claudeMd).toContain("Fail closed");
+  expect(r.claudeMd).not.toContain("Surgical changes");
 });
 
 test("a mixed repo with both Cargo.toml and package.json resolves to js-ts (no regression)", () => {
--- a/test/readiness.test.ts
+++ b/test/readiness.test.ts
@@ -195,13 +195,32 @@ test("score is the weighted fraction of present guardrails, derived from the che
   expect(present("git_hooks", r)).toBe(true); // .husky/ implies a hook manager
 });
 
-test("generated CLAUDE.md is non-empty, encodes the surgical posture, and names missing tooling", () => {
+test("generated CLAUDE.md is non-empty, keeps the quiet-failure posture, and names missing tooling", () => {
   pkg({ name: "needs-coaching", devDependencies: { typescript: "^6" } });
   write("tsconfig.json", "{}");
   const r = analyzeReadiness(dir);
-  expect(r.claudeMd).toContain("Surgical");
+  expect(r.claudeMd).toContain("Fail closed");
   // a missing high-leverage guardrail should be called out in the prescription
   expect(r.claudeMd.toLowerCase()).toContain("prettier");
+});
+
+// Issue #2015: the artifact is written into a TARGET repo and re-read every turn, so it must not
+// restate the engineering posture every Shepherd spawn already carries in its system prompt, nor
+// general judgement a capable model follows unprompted. Without this guard nothing stops the
+// section re-growing — which is the whole failure the cut exists to prevent.
+test("generated CLAUDE.md carries no posture bullet the system prompt or the guardrails already cover", () => {
+  pkg({ name: "needs-coaching", devDependencies: { typescript: "^6" } });
+  const md = analyzeReadiness(dir).claudeMd;
+  for (const dropped of [
+    "Simplicity first",
+    "Surgical changes",
+    "Verify before claiming done",
+    "No dead code",
+  ])
+    expect(md).not.toContain(dropped);
+  // …while the bullets that describe a QUIET failure no gate fires on all survive.
+  for (const kept of ["Fail closed", "Single source of truth", "Keep names and comments honest"])
+    expect(md).toContain(kept);
 });
 
 test("CLAUDE.md always tells the repo to add .shepherd-* to .prettierignore, even when prettier is already present", () => {


scoop (commented):
## What I verified

**The cut itself is confined to the posture section.** `git diff aec14e0415a9fe0049acc28c8a4de77e9ccc08e9...HEAD -- src/readiness.ts` touches exactly two regions: the `generateClaudeMd()` doc comment and the posture heading/bullets inside the template literal. The stack label (`src/readiness.ts:694`), the `## Adopt these guardrails (highest leverage first)` block and its `${adoptLines}` interpolation (`src/readiness.ts:715-717`), and `${profile.trailingNote(ctx)}` are untouched — so the acceptance criterion "prescription byte-identical" holds by construction, for both profiles (both go through the same `generateClaudeMd`, called once at `src/readiness.ts:632`).

**Retained bullets do not restate `ENGINEERING_POSTURE`.** Read the base copy at `epic/2005-epic-adopt-the-claude-5-context-engineer@aec14e0415a9fe0049acc28c8a4de77e9ccc08e9:src/service.ts:735-745`: its three bullets are *Simplicity first*, *Surgical changes* (which also carries the dead-code clause: "Delete only what YOUR change orphaned; surface pre-existing dead code"), and *Goal-driven execution* ("never declare work done before verifying"). The four dropped bullets map onto those one-for-one; the three retained (`Fail closed`, `Single source of truth`, `Keep names and comments honest`, `src/readiness.ts:706-712`) have no counterpart there. The doc comment's historical claim at `src/readiness.ts:659-663` is therefore factually accurate about what was removed and why.

**Pass-through to the one-click adopt task needs no code change.** `ui/src/lib/components/readiness-view.ts:32-33` is `return \`${intro}\n\n${claudeMd}\``, and `ui/src/lib/components/ReadinessPanel.svelte:142` passes `report.claudeMd` straight in, so the trimmed artifact propagates end to end.

**i18n parity is untouched.** No message key is added or changed by this diff; `readiness_adopt_intro_create` / `readiness_adopt_intro_merge` exist in both catalogs at `ui/messages/en.json:927-928` and `ui/messages/de.json:927-928`, and both still say "install the guardrails below and add a CLAUDE.md with these house rules" — accurate after the trim.

**No other consumer binds to the removed text.** Grepped the tree for `Surgical`, `Simplicity first`, `Verify before claiming done`, `No dead code`, `Engineering posture`: outside `docs/research/*` (prose about the *spawn-side* block) the only hits are `src/service.ts:714/718` (the spawn-side `ENGINEERING_POSTURE`, owned by #2002), `test/service.test.ts:4576-4577` (asserts the spawn block, not the artifact), and this PR's own new absence assertions. No fixture, snapshot, or UI test embeds the generated posture text.

**The onboarding skill stays accurate, so leaving it alone is right.** `.claude/skills/shepherd-onboarding/SKILL.md:109-113` hands the `# House rules for AI agents` heading and the "adopt these guardrails" list to the readiness analyzer — both survive verbatim (`src/readiness.ts:694`, `:715`) — and `:248` restates the same exclusion. Nothing in either clause names the posture heading that was renamed.

**Tests.** `test/readiness.test.ts:198-205` retargets `toContain("Surgical")` → `toContain("Fail closed")` without deleting the assertion, and keeps the `toLowerCase().toContain("prettier")` missing-tooling guard — so the artifact retains its only non-empty guard, as the issue's seam demanded. The new absence/presence test at `test/readiness.test.ts:211-224` and the Rust-profile pair at `test/readiness-rust.test.ts:160-164` cover the regression in both directions and in both profiles. I checked that none of the asserted-absent strings can be produced by the prescription tables: `dead_code_audit`'s label is "A dead-code/complexity audit (fallow/knip)" (`src/readiness.ts:774`), which does not contain `No dead code`.

## Stated limitations

- I could not read the PR body (the `gh` call was denied by the permission mode), so I cannot confirm the acceptance criteria that live there: the seven-row keep/drop table, the recorded divergence from `LEARNING_ADMISSION_TEST`, and the `--base epic/2005-epic-adopt-the-claude-5-context-engineer` target. Those remain unverified rather than wrong.

## Nits (non-blocking):

- `src/readiness.ts:669` — the surviving "Fully profile-driven: the stack label," line is an orphaned short line left by the inserted paragraph; re-wrapping it would read less like a splice artifact (prettier does not reflow prose comments, so lint won't).
- `src/readiness.ts:659-668` — a nine-line historical narrative in the doc comment is a lot of standing prose for a slice about cutting standing prose; the last two sentences ("Anything a gate CAN catch… Keep it that way when editing.") carry the whole editing constraint on their own.

## Judgement call I agree with

`No dead code` → DROP (the issue left it to adjudication). It is covered twice over: mechanically by the `dead_code_audit` guardrail the prescription installs (`src/readiness.ts:774`, `src/readiness.ts:786`), and behaviourally by `ENGINEERING_POSTURE`'s *Surgical changes* clause at `epic/2005-epic-adopt-the-claude-5-context-engineer@aec14e0415a9fe0049acc28c8a4de77e9ccc08e9:src/service.ts:741-743`. Dropping it is consistent with the artifact's own thesis.

Issue #1206, opened 2026-06-28
epic-merger: lead merge subject with conventional type so release-please picks it up


scoop:
## Problem

When the epic-merger lands an epic, it writes the merge-commit subject as:

```
Land epic #535: feat(comments): communication feed (comments) on Objectives & TODOs [EPIC] (#549)
```

release-please parses the conventional-commit **type from column 0** of the subject. Because this leads with `Land epic #535:`, the `feat` type is mid-subject and never matched → release-please treats the commit as **non-conventional and skips it entirely** (no changelog entry, no version contribution).

Observed downstream in `ltdovr/flowagent`: the v2.19.0 release PR omitted the communication-feed epic — its headline feature — even though the code shipped. Had to be patched after the fact with an empty conventional commit on `dev`.

## Fix

Generate the epic-merge subject with the conventional type at column 0, e.g.:

```
feat(comments): communication feed on Objectives & TODOs (epic #535, #549)
```

Move the "epic"/issue framing into the scope/description or body — not in front of the type. Keep `type(scope): …` as the literal start of the subject line.

## Notes

- Same constraint applies to any merge/squash subject the agent composes (not just epics).
- A repo without admin bypass on its default branch can't apply the empty-commit workaround via direct push, so getting the subject right at merge time is the only reliable path there.

scoop:
✅ PR #1207 merged.

Issue #327, opened 2026-06-04
Add a feature-announcement entry for the "Halt the herd" e-stop


scoop:
Follow-up to #326 (merged in 1.15.0).

The global **"Halt the herd"** emergency-stop button (TopBar 🛑) is user-facing but shipped without a `feature-announcements.ts` catalog entry — it predated the catalog-upkeep gate added in #325. As a result the What's-New drawer / first-view coachmark never announces it.

This is exactly the rot the #325 gate now prevents for future feats; #326 was grandfathered in.

## Task
Per the "Feature discovery" section in `CLAUDE.md`, in one PR:

1. Append a `FeatureAnnouncement` to `featureAnnouncements` in `ui/src/lib/feature-announcements.ts`:
   - `id`: stable kebab slug, e.g. `halt-the-herd`
   - `sinceVersion`: `1.15.0` (the release #326 shipped in)
   - `titleKey` + `bodyKey`
   - optional `targetId` + `use:coachTarget={"<id>"}` on the TopBar halt control so the coachmark points at it
2. Add the `titleKey`/`bodyKey` to **both** `ui/messages/en.json` and `de.json` (`check:i18n` parity).

## Acceptance
- Catalog entry present; `check:i18n` green; `check-feature-catalog.sh` green.
- What's-New drawer surfaces the halt e-stop; coachmark (if added) anchors to the 🛑 control.

PR #2022, opened 2026-08-04
fix(epic): strip leading `Epic:`, lowercase landing-PR subject

### LINKED ISSUE #2021: fix(epic): landing-PR title keeps a leading `Epic:`, failing the pr-title gate
`buildLandingPrTitle` (src/epic-landing.ts:41) emits a title that Shepherd's own `pr title` workflow rejects, for the most common epic-title shape there is.

### What happened

PR #2018 (landing epic #2005) went red on `pr title` alone — every code check passed:

```
✖  subject must not be sentence-case, start-case, pascal-case, upper-case [subject-case]
   feat: Epic: adopt the Claude 5 context-engineering rules in Shepherd's prompt assembly (epic #2005)
```

Fixed by hand-retitling the PR. That is a per-epic manual step, and anything that regenerates the title from the parent issue reverts it.

### Root cause

`buildLandingPrTitle` matches `^(\w+)(\([^)]*\))?(!)?:` and, when the prefix is not in `RELEASE_TYPES`, treats the whole parent title as prose and prepends `FALLBACK_TYPE`. A parent titled `Epic: <desc>` therefore becomes `feat: Epic: <desc> (epic #<n>)` — a subject starting with a capitalized word, which commitlint's `subject-case` rule rejects.

`Epic: ` is not an edge case: it is the prefix Shepherd's own epic authoring produces (#2005 is titled `Epic: adopt the Claude 5 context-engineering rules…`), so every such epic lands a red `pr title` check.

The function already normalizes a *trailing* `[epic]`/`[EPIC]` tag (line 42). It just doesn't strip a *leading* `Epic:`.

### Fix

Strip a leading `Epic:` (case-insensitive, optional whitespace) in the same cleanup as the trailing tag, before the conventional-prefix match — so `Epic: adopt the …` yields `feat: adopt the … (epic #<n>)`.

Two details worth getting right:

- Strip it only when it is *not* a recognized conventional type, so a legitimately-typed parent (`feat: …`) is untouched.
- An epic titled with nothing but the tag (`Epic:`) must not produce an empty description — the existing `desc ? … : \`${prefix}: epic #${parentNumber}\`` fallback covers the typed branch; the untyped branch needs the same guard.

### Test

test/epic-landing.test.ts already covers the bare / non-type-prefix / trailing-tag cases. Add: parent `Epic: adopt X` → `feat: adopt X (epic #7)`, plus a lowercase `epic: ` variant and the tag-only degenerate case. Worth asserting the result passes commitlint's `subject-case` shape (leading char not uppercase) so the gate and the builder can't drift again.

Found while repairing the conflict on #2018.

scoop:
Closes #2021

`buildLandingPrTitle` emitted landing-PR titles that Shepherd's own `pr title` workflow rejects. PR #2018 (landing epic #2005) went red on `pr title` alone and had to be hand-retitled.

## What the gate actually enforces

`.github/workflows/pr-title.yml` pipes the title through `bunx commitlint` with the repo's config. I ran the real binary rather than reasoning about the rule; `subject-case` reduces to one invariant:

> the subject — everything after `type(scope)!?: ` — must not begin with an uppercase letter.

That makes the bug wider than #2021 frames it. Measured, before this change:

| Emitted title | Gate |
| --- | --- |
| `feat: Epic: adopt the Claude 5 … (epic #2005)` | RED — the reported case |
| `feat: EFI value map (epic #327)` | RED — asserted as *expected* in the suite |
| `feat: Comments: communication feed (epic #90)` | RED — asserted as *expected* in the suite |
| `feat(x): Rework pipeline (epic #91)` | RED — the recognized-type branch, not in the issue |

So the suite was green while the gate would have failed for those epics, and the typed branch was affected too — the issue only discusses the fallback branch.

## Fix

Both inside `buildLandingPrTitle`:

1. Strip a leading `Epic:` (case-insensitive) in the same cleanup as the existing trailing `[epic]` tag, before the conventional-prefix match. Safe by construction — `epic` is not in `RELEASE_TYPES`, so no legitimately-typed parent can match it.
2. Lowercase the description's first character in **both** branches.

Stripping `Epic:` also lets the fallback branch produce an empty description (a title that is nothing but the tag), so the `desc ? … : "<type>: epic #<n>"` guard now covers both returns, not just the typed one.

| Parent epic title | Emitted |
| --- | --- |
| `Epic: adopt X` | `feat: adopt X (epic #7)` |
| `Epic:` | `feat: epic #7` |
| `EFI value map` | `feat: eFI value map (epic #327)` |
| `fix(api)!: Drop legacy route` | `fix(api)!: drop legacy route (epic #88)` |
| `feat(x): add y` | `feat(x): add y (epic #91)` (untouched) |

`eFI` is deliberate: ugly but green. Acronym-aware special-casing would still leave the gate red, and gate correctness beats prettiness on a rare title shape. Confirmed with the operator before implementing.

Type, scope and `!` at column 0 are untouched in every path, so release-please attribution (#1206 — the reason this builder exists) is unaffected.

## Testing

Every title case now goes through a wrapper that asserts the gate's invariant (subject's first char is not `[A-Z]`) in addition to the exact expected string, so the builder and the gate can't drift apart again. Deliberately *not* shelling out to commitlint from the unit suite — that would make it slow and add a `bunx` dependency; the invariant is measured once and encoded.

Two existing expectations changed (`EFI value map`, `Comments: communication feed`): they asserted output the real gate rejects.

New cases: `Epic:`/`epic:`/`Epic:no-space` prefixes, leading + trailing tag together, a real type carried behind the tag, the tag-only degenerate case, a capitalized description behind a recognized type, and a non-letter-initial description.

## Verification

- `bun run lint` — clean
- `bun run test` — 8240 pass, 0 fail
- Every one of the 19 titles the updated tests expect piped through the real `bunx commitlint`: all green, including the exact #2018 title that failed

### DIFF
--- a/src/epic-landing.ts
+++ b/src/epic-landing.ts
@@ -29,6 +29,15 @@ const RELEASE_TYPES = new Set([
  *  entry rather than letting the merge fall through unrecognized (#1206). */
 const FALLBACK_TYPE = "feat";
 
+/** Lowercase the first character of a subject. `.github/workflows/pr-title.yml` lints this very
+ *  title with the repo's commitlint config, whose `subject-case` rule (never sentence-case /
+ *  start-case / pascal-case / upper-case) rejects a subject starting with an uppercase letter —
+ *  that single character is the whole invariant (#2021). A non-letter lead (digit, backtick,
+ *  paren) already passes, and `toLowerCase` is a no-op there. */
+function lowerFirst(subject: string): string {
+  return subject.charAt(0).toLowerCase() + subject.slice(1);
+}
+
 /** Landing-PR title that doubles as the squash-merge **subject** release-please parses, so it
  *  MUST lead with a recognized conventional `type(scope)!?:` at column 0 (#1206 — a subject
  *  led by `Land epic #<n>:` pushes the real type mid-line and release-please skips the merge).
@@ -37,19 +46,27 @@ const FALLBACK_TYPE = "feat";
  *  - Parent title already conventional with a recognized type → keep it (type lowercased,
  *    scope/`!` verbatim), append ` (epic #<n>)`.
  *  - Bare title, OR a non-type `Word:` prefix (e.g. `Comments: …`) → prepend `feat:`.
- *  A trailing `[EPIC]`/`[epic]` tag is stripped either way. */
+ *  A trailing `[EPIC]`/`[epic]` tag and a leading `Epic:` — the prefix Shepherd's own epic
+ *  authoring produces — are stripped either way. The description is then lowercase-initial in
+ *  BOTH branches so the `pr title` gate stays green (#2021; see `lowerFirst`). A title that is
+ *  nothing but the tag leaves no description, hence the guard in both returns. */
 export function buildLandingPrTitle(parentNumber: number, parentTitle: string): string {
-  const cleaned = parentTitle.trim().replace(/\s*\[epic\]\s*$/i, "");
+  // `epic` is not in RELEASE_TYPES, so stripping the leading tag can never clobber a real type.
+  const cleaned = parentTitle
+    .trim()
+    .replace(/\s*\[epic\]\s*$/i, "")
+    .replace(/^epic\s*:\s*/i, "");
   const epicTag = `(epic #${parentNumber})`;
 
   const m = /^(\w+)(\([^)]*\))?(!)?:\s*(.*)$/.exec(cleaned);
   if (m && RELEASE_TYPES.has(m[1]!.toLowerCase())) {
     const prefix = `${m[1]!.toLowerCase()}${m[2] ?? ""}${m[3] ?? ""}`;
-    const desc = m[4]!.trim();
+    const desc = lowerFirst(m[4]!.trim());
     return desc ? `${prefix}: ${desc} ${epicTag}` : `${prefix}: epic #${parentNumber}`;
   }
 
-  return `${FALLBACK_TYPE}: ${cleaned} ${epicTag}`;
+  const desc = lowerFirst(cleaned);
+  return desc ? `${FALLBACK_TYPE}: ${desc} ${epicTag}` : `${FALLBACK_TYPE}: epic #${parentNumber}`;
 }
 
 /** Sanitize a child title for a single Markdown table cell: collapse newlines to spaces (a
--- a/test/epic-landing.test.ts
+++ b/test/epic-landing.test.ts
@@ -1,47 +1,89 @@
 import { describe, expect, it } from "bun:test";
 import { buildLandingPrBody, buildLandingPrTitle } from "../src/epic-landing";
 
+/** Build a landing title AND assert it clears the `pr title` gate. `.github/workflows/pr-title.yml`
+ *  lints the PR title with the repo's commitlint config, whose `subject-case` rule rejects a
+ *  subject whose first character is uppercase (#2021 — verified against `bunx commitlint`). Every
+ *  case below goes through this wrapper so the builder can never drift away from the gate again. */
+function build(parentNumber: number, parentTitle: string): string {
+  const title = buildLandingPrTitle(parentNumber, parentTitle);
+
+  const subject = /^\w+(?:\([^)]*\))?!?:\s*(.*)$/.exec(title)?.[1];
+  // A missing prefix is itself the #1206 bug — assert on `title` so the failure names it.
+  if (subject === undefined) expect(title).toBe("<a conventional type(scope)!?: prefix>");
+  expect(subject).not.toMatch(/^[A-Z]/);
+
+  return title;
+}
+
 describe("buildLandingPrTitle", () => {
   it("keeps a recognized conventional prefix at column 0 and appends the epic tag", () => {
-    expect(buildLandingPrTitle(535, "feat(comments): communication feed on Objectives")).toBe(
+    expect(build(535, "feat(comments): communication feed on Objectives")).toBe(
       "feat(comments): communication feed on Objectives (epic #535)",
     );
   });
 
   it("preserves a breaking-change `!` (with and without scope)", () => {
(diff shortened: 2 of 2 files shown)


scoop (commented):
## Verdict

The diff does what the task asks and nothing risky beyond it. I traced every branch of `buildLandingPrTitle` against its callers and the gate it is bending to, and found no correctness defect. Decision: **comment**, findings empty.

**1. The gate the builder must satisfy is the one the comments cite.** `.github/workflows/pr-title.yml:64` pipes the PR title through `bunx commitlint`, and `.github/workflows/pr-title.yml:9` states it deliberately reuses `commitlint.config.js`, which is `extends: ["@commitlint/config-conventional"]` (`commitlint.config.js:1-3`). So a landing-PR title really is linted by `subject-case`, and the new doc comments at `src/epic-landing.ts:32-36` and `src/epic-landing.ts:49-52` describe the actual pipeline, not a guess.

**2. The normalization is sufficient under any reading of `subject-case`.** The forbidden set is `sentence-case`, `start-case`, `pascal-case`, `upper-case`. A subject whose first character is a lowercase letter cannot equal `upperFirst(...)` of itself (kills sentence/start/pascal) and cannot equal its own `toUpperCase()` (kills upper-case), because it contains at least one lowercase letter. `lowerFirst` (`src/epic-landing.ts:37-39`) is applied on **both** exit paths — `src/epic-landing.ts:64` (recognized-type branch) and `src/epic-landing.ts:68` (fallback branch) — so every subject the builder can emit is either lowercase-initial or non-letter-initial. No sibling path is left unguarded, which was the specific latent-defect shape I went looking for.

**3. Both returns now carry the empty-description guard.** `src/epic-landing.ts:65` (pre-existing) and `src/epic-landing.ts:69` (new). This matters precisely because the new `^epic\s*:\s*` strip at `src/epic-landing.ts:58` can reduce `"Epic:"` to `""` and drop it into the fallback branch — without the new guard that path would have emitted `feat:  (epic #12)`. Traced by hand: `"Epic: [epic]"` → trim → trailing-tag strip → `"Epic:"` → leading strip → `""` → `feat: epic #13`, which is what `test/epic-landing.test.ts:76` asserts.

**4. Stripping a leading `Epic:` cannot clobber a real type.** `epic` is not a member of `RELEASE_TYPES` (`src/epic-landing.ts:13-25`), so the comment at `src/epic-landing.ts:54` is accurate and the issue's "strip it only when it is not a recognized conventional type" requirement is satisfied structurally rather than conditionally. Confirmed the ordering is safe the other way too: `"Epic: fix(api)!: drop legacy route"` strips to `fix(api)!: drop legacy route` and then takes the typed branch, preserving type, scope and `!` verbatim at column 0 (`test/epic-landing.test.ts:69-71`).

**5. Stripping is semantically safe — the tag carries no meaning to Shepherd.** `src/service.ts:830` and `docs-site/src/content/docs/authoring-epics.md:32` both state explicitly that an `[EPIC]` title prefix is *not* an epic marker; epic detection is by sub-issues / body references. So removing the prose tag from the derived PR title discards nothing load-bearing.

**6. No caller depends on the old title string.** All three call sites — `src/drain.ts:1212` (refresh), `src/drain.ts:1440` (open), `src/drain.ts:1614` (pre-warm draft) — feed the title straight into `forge.editPr`/`forge.openPr`. Idempotency is keyed on the branch, not the title (`src/drain.ts:1609` uses `forge.prStatus(branch)`; `src/drain.ts:1221` uses `existing.number`), so a changed title cannot orphan an already-open landing PR. I grepped every `title` reference in `src/drain.ts` and found no title-equality matching.

**7. Release-please attribution is untouched.** The type still sits at column 0 on every path (`src/epic-landing.ts:63`, `src/epic-landing.ts:69`), which is the #1206 invariant this builder exists for. `isDocRelevantMerge` (`src/doc-agent.ts:1500-1506`) reads only the type/scope via `/^\s*(\w+)(?:\(([^)]*)\))?!?:/` and lowercases both, so lowercasing the *description* cannot change its classification.

**8. `lowerFirst` is safe on non-ASCII and surrogate input.** `charAt(0) + slice(1)` reconstitutes the original string exactly when `toLowerCase()` is a no-op (digit, backtick, paren, lone high surrogate), and correctly lowercases `Ü`-style leads. No mangling path.

## Stated limitation (could not verify, not asserting either way)

- `@commitlint/ensure`'s exact `subject-case` semantics in the pinned v21 line (`bun.lock:48` → `@commitlint/ensure@21.2.0`) could not be verified: `node_modules` is not installed in this review worktree, so I could not read `to-case`/`case` to confirm whether the check is whole-subject or first-word. This only affects whether *some* of the changed expectations were strictly necessary (notably `EFI value map` → `eFI value map`); it does not affect whether the fix is *safe*, since as shown in (2) a lowercase-initial subject is green under either semantics. The plan reports these were measured against the real binary, and the CI excerpt in the originating issue is consistent with that.

## Nits (non-blocking):
- `test/epic-landing.test.ts:13-14`: when the prefix regex fails to match, the helper asserts on `title` and then still calls `expect(subject).not.toMatch(...)` with `subject === undefined`; unreachable in practice (the first `expect` throws), but the second line reads as if it were a valid assertion on `undefined`.
- `test/epic-landing.test.ts:13`: using `expect(title).toBe("<a conventional type(scope)!?: prefix>")` as a failure-message device is inventive but non-obvious — a plain `throw`/`expect(...).toMatch(prefixRe)` would say the same thing more directly.
- `src/epic-landing.ts:58`: the strip removes only one leading `Epic:`, so `"Epic: Epic: foo"` yields `feat: epic: foo (epic #n)`. Still gate-green, so this is cosmetic only.

- `src/epic-landing.ts:37-39` + `src/epic-landing.ts:64`: lowercasing the description in the *recognized-type* branch goes beyond the literal issue (which only discusses the `Epic:` fallback shape). It is explicitly the operator-confirmed decision in the approved plan and it fixes a real gate failure class (`fix: Drop legacy route`), so it is a legitimate keep — noted only for the reader.

PR #2023, opened 2026-08-04
fix: session rename must not move a branch heading an open PR

scoop:
-

### DIFF
--- a/src/forge/github.ts
+++ b/src/forge/github.ts
@@ -1616,19 +1616,6 @@ export class GithubForge implements GitForge {
     return { number: n, url };
   }
 
-  async renameBranch(oldBranch: string, newBranch: string): Promise<void> {
-    // GitHub's rename-branch endpoint moves the ref AND retargets every open PR and
-    // branch-protection rule onto the new name, so a session's open PR follows along.
-    await this.run([
-      "api",
-      "--method",
-      "POST",
-      `repos/${this.slug}/branches/${oldBranch}/rename`,
-      "-f",
-      `new_name=${newBranch}`,
-    ]);
-  }
-
   async merge(prNumber: number, o: MergeInput): Promise<void> {
     const method =
       o.method === "rebase" ? "--rebase" : o.method === "merge" ? "--merge" : "--squash";
--- a/src/forge/types.ts
+++ b/src/forge/types.ts
@@ -444,10 +444,6 @@ export interface GitForge {
    *  cut an epic integration branch off the default branch. Optional: hosts without a
    *  refs API omit it and the caller skips epic-branch orchestration. */
   ensureBranch?(branch: string, fromRef: string): Promise<void>;
-  /** Rename a branch on the host, retargeting any open PR to the new name. Optional:
-   *  hosts that can't do this safely (Gitea) omit it, and the caller falls back to a
-   *  display-only rename so an open PR is never orphaned. */
-  renameBranch?(oldBranch: string, newBranch: string): Promise<void>;
   merge(prNumber: number, o: MergeInput): Promise<void>;
   /** Post a plain issue comment on a PR (`gh pr comment`). Optional: only hosts
    *  with a comment API (GitHub) implement it; others omit it and the
--- a/src/server.ts
+++ b/src/server.ts
@@ -2953,33 +2953,40 @@ async function parseRenameName(req: Request): Promise<string | Response> {
   return raw;
 }
 
-// Decide whether the local branch moves. An OPEN PR forces the host into the loop:
-// GitHub retargets it by renaming the remote branch first (so `s.branch` never points
-// away from the PR); a host that can't (Gitea) yields a display-only rename. Returns the
-// flag, or a 502 Response when the remote rename failed.
-async function resolveRenameBranch(
-  deps: AppDeps,
-  s: Session,
-  newBranch: string,
-  hasOpenPr: boolean,
-): Promise<boolean | Response> {
-  const renameLocalBranch = s.isolated && !!s.branch;
-  if (!hasOpenPr || !renameLocalBranch || !s.branch) return renameLocalBranch;
-
+/**
+ * Is a PR open on `s.branch` right now? Trusts the poller cache only when it says
+ * "open" — that's the conservative answer and needs no network. Every other cached
+ * value (missing, or a `none` the poller hasn't re-read since the agent opened the PR)
+ * is re-confirmed against the forge, because the caller is about to move the branch on
+ * the strength of it. No forge ⇒ no PR. A failed lookup answers "open": keeping the
+ * branch is always recoverable, detaching a live PR from its session isn't (#1927).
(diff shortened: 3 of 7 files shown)


scoop (commented):
## What I verified

**The destructive path is gone, everywhere.**
`GithubForge.renameBranch()` (src/forge/github.ts, removed hunk) and the optional `GitForge.renameBranch?()` slot (src/forge/types.ts, removed hunk) are both deleted. I grepped the whole tree for `renameBranch`: the only surviving hits are `WorktreeMgr.renameBranch` — a *different*, 3-arg local-git method (src/worktree.ts:231, declared in the `ServiceDeps.worktree` Pick at src/service.ts:173, called at src/service.ts:4600) — plus its tests (test/service.test.ts:2563, 3979, test/rename.test.ts:40). No forge implementation (src/forge/gitea.ts, src/forge/local.ts) ever declared the method, so removing it from the interface breaks no `implements GitForge` class. Success criterion "no code path anywhere can call a forge branch-rename API" holds.

**`prStatus` resolves and is non-optional.**
src/server.ts:2970 calls `forge.prStatus(s.branch)` with no `?.`. `prStatus(headBranch)` is a *required* member of `GitForge` (src/forge/types.ts:412), and all three implementations provide it: src/forge/github.ts:1296, src/forge/gitea.ts:341, src/forge/local.ts:384. No optional-call hazard.

**The guard is consistent across the server/service boundary.**
src/server.ts:2984 gates on `s.isolated && s.branch`; `SessionService.rename` independently recomputes `opts.renameLocalBranch && s.isolated && !!s.branch` (src/service.ts:4597) and only shells `git branch -m` under it (src/service.ts:4599-4601). The two agree, so `branchRenamed:false` in the response can never coexist with a moved branch, and vice versa.

**Staleness across the new `await`.** `s` is read at src/server.ts:2994, before the (potentially seconds-long) forge lookup at :3002. The subsequent write goes through `deps.service.rename(s.id, …)`, which re-reads the row (src/service.ts:4595) and derives the old branch name from the *fresh* row — so a concurrent branch adoption (`syncWorktreeBranch`, src/service.ts:4577) can't be clobbered by a stale snapshot. Only the `slug === s.name` short-circuit at :2998 uses the stale snapshot, which is harmless.

**Response-shape consumers agree.** `prRetargeted` is gone from the two emit sites (src/server.ts:2998, :3023) and grep finds zero remaining references repo-wide — nothing in `ui/`, `extension/`, `docs/`, or the tests reads it. The only readers of the rename response are ui/src/lib/api.ts:1444 (return type narrowed) and ui/src/lib/components/Viewport.svelte:1015 (`!res.branchRenamed`), plus the mock fixture at ui/src/lib/components/Viewport.browser.test.ts:16-19, which now matches the narrowed type.

**i18n parity.** No new message keys. The reused key `viewport_rename_branch_kept` exists in both catalogs — ui/messages/en.json:696 ("Name changed; branch kept (open PR)") and ui/messages/de.json:696 ("Name geändert; Branch beibehalten (offener PR)") — and both already read correctly under the new "open PR pins the branch" semantics, so the corrected comment at Viewport.svelte:1014 needed no string change.

**Conditional cache drop.** src/server.ts:3017 drops only when the branch moved. Verified this is right: the poller cache is session-id-keyed PR state (test harness `snapshot()[s.id]`, test/rename.test.ts:177-186); a display-only rename changes neither the branch nor the PR, so the cached row stays valid.

**Logging convention.** `console.warn("[rename] …")` at src/server.ts:2973 matches the existing `[tag] message:` form used at src/server.ts:2113, :3301, :1074.

**Tests.** `PrStatus` is imported at test/rename.test.ts:10, so the `as PrStatus` casts at :132 and :331 type-check. The `prStatus` spy wrapper (test/rename.test.ts:190-198) spreads the fake forge object literal, whose members are own enumerable props (built via `Object.assign` at :139), so the spread preserves them. The throwing case (:315-320) rejects rather than throws synchronously, which the `await` in `sessionHasOpenPr` still catches.

### Stated limitation

I could not verify the external claim in the rewritten comment at src/server.ts:2979-2982 — that GitHub's branch-rename API closes a PR that uses the branch as its *head* — against GitHub's live documentation; that ground truth is not in this repo. I neither confirm nor dispute it. It does not affect the verdict: the chosen behaviour (never move the branch when a PR may be open) is strictly non-destructive under either reading of the API, and the issue's own timeline evidence is consistent with the comment.

- src/server.ts:3000-3005 — the `name_taken` 409 pre-check was moved behind `renameLocalBranch`, so a display-only rename onto a slug whose `shepherd/<slug>` branch already exists now returns 200 instead of 409. Logically right (nothing moves, so nothing can collide) but it is a validation relaxation the plan did not call for; the observable consequence is that two sessions can now carry the same display name. Low impact — `rename()` touches only `name`/`branch` (src/service.ts:4602), not the worktree path, and manual rename never relabels the herdr agent — but worth a conscious sign-off.

## Nits (non-blocking):

- src/server.ts:2967 — `deps.resolveForge?.(s.repoPath)` sits outside the `try`. `resolveForge` shells out to git on a cold memo (src/forge/resolve.ts:101 → src/forge/index.ts:83), and src/diff-base.ts:37-45 deliberately wraps *both* `resolveForge` and `prStatus` in one `try` for exactly that reason. The old code had the same shape, but it was only reachable on the rare open-PR path; this diff makes it the common path. A throw here would 500 the rename rather than degrade to the intended "assume open" — non-destructive, but it defeats the fallback the docstring promises.
- src/server.ts:2970 — the live `gh pr list` runs on an HTTP request path with no timeout (`defaultRunner`, src/forge/github.ts:200-203, uses `execFile` with no `timeout` option), so a hung `gh` leaves the rename spinner stuck indefinitely. Consistent with existing on-demand `prStatus` callers (src/diff-base.ts:42), so not a new class of risk.

- src/service.ts:4545-4554 — the `rename()` JSDoc is misplaced: it sits immediately above a second JSDoc for `branchExists` (:4555), so it documents `branchExists`, while the actual `rename` at :4594 has no doc comment. Pre-existing.
- src/service.ts:4551 — that same doc block says `renameLocalBranch` is "false for a display-only rename when an open PR can't be retargeted, true otherwise"; this PR makes the caller contract "false whenever an open PR exists", so the sentence is now stale. Untouched file, informational only.
- src/forge/github.ts:1316-1329 — `prStatus` fetches `--state all --limit 1` for non-fork repos, i.e. the newest PR for the head branch. A branch that has an *older* open PR plus a *newer* closed one would report `closed`, which this diff now trusts to move the branch. Pre-existing behaviour of `prStatus`, unchanged here, and vanishingly rare for `shepherd/<session>` branches.
- src/server.ts `handleSessionRename` does not call `herdr.relabel`, unlike the other two rename paths (src/service.ts:4240, :4588), so a manually renamed session keeps its old terminal tab label. Pre-existing and untouched by this diff, but display-only renames make the name/branch/tab divergence more common.

Issue #1922, opened 2026-07-24
Arm preview stop (stopListenersOnPort) on macOS behind a kill-age bound


scoop:
Follow-up to #1912 / PR #1921.

The darwin probe backend sets `canAuthorizeSignal: false`, so `stopListenersOnPort` never signals on macOS — it returns `{ signalled: 0, unsupported: true }` and the idle-stop ladder refuses to advance. That matches today's behaviour (macOS already returned a zero kill count), but it means **macOS operators cannot stop a preview or reclaim an idle one**.

Arming it was deliberately excluded from #1921 because it is the one part that would NOT be "strictly no worse than today": it turns a harmless no-op into a SIGKILL authorized by *snapshot* data, on a platform with no cheap pid-recycle fingerprint (`cpuStatForPid` is not implemented, and per-pid re-verification would mean a spawn per pid on a synchronous path).

Doing it properly needs:

- a dedicated **`maxKillAgeMs`** bound, deliberately NOT derived from `config.previewSweepMs` (which would let a tuned sweep interval widen the kill window to ~2 minutes), and expressed relative to the observed `lsof` sample duration since the snapshot's START stamp is a lower bound on staleness;
- a **recycle fingerprint** before signalling (the starttime trick `reapMarkedOrphans` already uses on Linux), or per-pid re-verification if its cost turns out acceptable;
- keeping the refusal **observable** — the `unsupported` outcome already threads to `escalateIdleStop`, and a `refused` variant would need the same treatment so a refusal never burns the SIGTERM → SIGKILL → gaveUp ladder.

The HTTP/UI half (a distinct `handlePreviewStop` response, an `api.ts` result variant, a `ViewportTabBar` branch, an i18n key) was also cut as a nicety and belongs here.

scoop:
✅ PR #2025 merged.

PR #2024, opened 2026-08-04
fix(preview): surface probesUnavailable on the decommission surfaces (#1923)

### LINKED ISSUE #1923: Surface probesUnavailable on the leftovers / clear-merged decommission surfaces
Follow-up to #1912 / PR #1921.

`ProcessReaper.detect` returns `Leftover[]`, so on a host where process detection is dead or stale it yields `[]` — **indistinguishable from "nothing is running"**. That flows to:

- `GET /api/sessions/:id/leftovers` (the close-confirm modal),
- `handleSessionsClearMerged`'s GET summary,
- `archiveMany`'s per-id `keys`.

So an operator on a broken host is told **"0 leftovers"**, clears the batch, and leaks every dev server — with no signal outside the Diagnose row.

Cut from #1921 on the same test applied to the rest of that PR: the issue asks for surfacing "in Diagnose / on the Preview badge" (both delivered), and on those hosts `leftovers` **already** returns `[]` silently today — so it is a pre-existing gap, not a regression the PR introduces. What #1921 *did* keep is the forced probe refresh before every `detect`, so the list reflects request-time state.

Shipping it needs a `probesUnavailable` flag on both responses plus, on the UI side:

- both decommission paths early-return **before any dialog mounts** (`Viewport.svelte` `confirmDecommission` and `+page.svelte` `commandDecommission` both do `if (found.length === 0) { onarchive(id); return; }`, and both mounts are gated `{#if leftovers.length > 0}`) — so a caution rendered *inside* `LeftoverDialog` is unreachable on exactly the broken host it is for. An **alert toast** on both paths is the shape that works, leaving the deliberate "a probe failure must never block decommission" guards intact.
- `ClearMergedDialog` *does* mount at count 0, but only its `clearmerged_leftovers` line is inside the count guard — the caution must render **outside** it.
- `AppOverlays.svelte` is `ClearMergedDialog`'s only mount, so the prop threads through it and `page/AppOverlays.browser.test.ts`'s exhaustive props literal.
- demo mode (`ui/src/lib/demo/state.ts` + `router.ts`) pins the array shape in `router.test.ts` (`GET /api/sessions/:id/leftovers returns [], never {}`) — that assertion exists precisely to catch this change and must be updated deliberately.

scoop:
Closes #1923.

`ProcessReaper.detect` returns `Leftover[]`, so on a host where process detection is dead it yields `[]` — **indistinguishable from "nothing is running"**. That flowed to the close-confirm modal, the clear-merged summary, and `archiveMany`'s per-id keys, so an operator on a broken host was told **"0 leftovers"**, cleared the batch, and leaked every dev server with no signal outside the Diagnose row.

## The predicate is not preview-start's

`preview/start` computes its flag as `probeHealth().state === "none"`. Copying that here would be silent on exactly the broken host: on darwin **both** leftover classes short-circuit (`canDetectLeftovers()` — class-2 by `canAuthorizeSignal: false`, class-3 by an absent `listeningPorts`), so `detect` returns `[]` no matter how healthy the snapshot is — while the poller keeps that same cell `"fresh"` for preview detection.

So `SessionService.leftoverProbesUnavailable()` is `!canDetectLeftovers() || health.state === "none"`, neutral (`false`) when no reaper is wired. `"stale"` does not trip it, matching preview-start. This does mean macOS shows the caution on every decommission until #1922 arms the kill path — deliberate, and truthful about the host.

## Wire

- `GET /api/sessions/:id/leftovers` → `{ leftovers, probesUnavailable }`. A deliberate break from the bare array: there is nowhere on an array to hang the flag. Both consumers (`getLeftovers` in `ui/src/lib/api.ts`, the demo router) are in-repo and updated here; the demo test that pinned the array shape now pins the object one.
- `GET /api/sessions/clear-merged` → gains the flag, additively. The `POST` half is untouched — the caution belongs before the destructive action.

## UI

Both decommission paths early-return **before any dialog mounts** (`Viewport.confirmDecommission`, `+page.svelte`'s ⌘K `commandDecommission`; both mounts gated on `leftovers.length > 0`), so a caution rendered inside `LeftoverDialog` would be unreachable on the very host it is for. Both raise an assertive, per-session-deduped toast instead and still close — the deliberate "a probe failure must never block decommission" guards are untouched.

`ClearMergedDialog` *does* mount at count 0, so its caution renders **outside** the `{#if leftovers > 0}` guard, threaded through `AppOverlays` (its only mount).

## Tests

- Server: both endpoints, incl. the darwin shape (fresh snapshot + no detection capability → `true`), a dead snapshot → `true`, and no reaper → `false`.
- `Viewport.browser.test.ts`: the toast fires **and** `onarchive` still fires — pinning the never-block guard, which is the part worth locking.
- `AppOverlays.browser.test.ts`: the caution renders at `leftovers === 0`, and not when probes work.
- Demo shape assertions updated deliberately.

Ships as `fix` — a pre-existing gap, not a new capability — so no feature-catalog entry.

Verified: root `bun run lint` + `bun run test` (8237 pass), `ui/` `bun run check`, `check:i18n` (2 locales in parity), `bun run test` (4525 pass).


scoop (commented):
## Verdict

The diff does what #1923 asks and I found nothing blocking. Everything below was checked against the tree, not inferred.

### Predicate (`src/service.ts:4800-4805`)

`leftoverProbesUnavailable()` is `!reaper ? false : (canDetectLeftovers?.() ?? true) === false ? true : (health?.().state ?? "fresh") === "none"`.

- The capability arm is the load-bearing one, and it is real: `ProcessReaper.canDetectLeftovers()` (`src/process-reaper.ts:695-699`) is `canAuthorizeSignal !== false || listeningPorts !== undefined`, and both `detect` classes fail closed on exactly those conditions — class 2 at `src/process-reaper.ts:580`, class 3 at `src/process-reaper.ts:616` — so on darwin `detect` returns `[]` while `health()` (`src/process-reaper.ts:705-710`) reads `state: "fresh"` off the poller-warmed cell. Reusing `probeHealth()` (`src/service.ts:4781-4783`) alone would indeed have reported `false` on the broken host, as the plan claims. Verified, not assumed.
- Both optional members are declared on the injected reaper type (`src/service.ts:192-193`, `Partial<Pick<ProcessReaper, "refresh" | "health" | "canDetectLeftovers">>`), so the `?.` calls type-check and the `?? true` / `?? "fresh"` defaults keep every existing reaper fake on the `false` branch. The same `(reaper.canDetectLeftovers?.() ?? true) === false` spelling is already used at `src/service.ts:4773`, so the two capability reads agree.
- The no-reaper test at `test/server.test.ts:941-949` uses `harness()`, which wires no `reaper` key — so it exercises the intended `!reaper → false` arm.

### Contract change and its consumers

`src/server.ts:2347-2350` (object body) and `src/server.ts:2539` (additive flag). Grepping `/leftovers` across the tree returns only `ui/src/lib/api.ts:476` and `ui/src/lib/demo/router.ts:153` as consumers — no extension, docs-site, or `src/` caller — and both are updated here (`ui/src/lib/api.ts:473-478`, `ui/src/lib/demo/state.ts:132-137`). Grepping `getLeftovers` gives exactly two call sites, `ui/src/lib/components/Viewport.svelte:952` and `ui/src/routes/+page.svelte:2263`; both destructure the new shape. `getMergedClearable` has one call site, `ui/src/routes/+page.svelte:2421`, which now reads `probesUnavailable`; the demo counterpart is `ui/src/lib/demo/state.ts:188-194`, forwarded verbatim at `ui/src/lib/demo/router.ts:48`. `POST /api/sessions/clear-merged` is untouched (`src/server.ts:2547-2552`), matching the plan's boundary.

### UI reachability — the point of the issue

The empty-list early return is still there and unmodified (`ui/src/lib/components/Viewport.svelte:966-969`, `ui/src/routes/+page.svelte:2275-2278`), so the caution genuinely cannot live inside `LeftoverDialog`; the toast fires before it (`Viewport.svelte:960-965`, `+page.svelte:2269-2274`) and the archive still runs. `toasts.info` honours both options as used: `key` dedupes/refreshes rather than stacking (`ui/src/lib/toasts.svelte.ts:123-142`) and `alert: true` gives the assertive 12s lifetime (`ui/src/lib/toasts.svelte.ts:110-116`). `toasts` is imported in both files (`Viewport.svelte:70`; `+page.svelte` uses it at :2243).

`ClearMergedDialog.svelte:60-62` renders the caution outside the `{#if leftovers > 0}` guard at :55-57, using the existing `.warn` rule (`ClearMergedDialog.svelte:149-154`, `var(--color-amber)` / `var(--fs-meta)` — no new tokens). Its only mount is `AppOverlays.svelte:643-651`, and the prop is declared at `AppOverlays.svelte:325-329`; the exhaustive props literal in `AppOverlays.browser.test.ts:141` carries it, so the required prop cannot be silently missing.

### i18n parity

`ui/messages/` holds exactly two catalogs. `clearmerged_probes_unavailable` is at `ui/messages/en.json:705` and `ui/messages/de.json:705`; `toast_decom_probes_unavailable` at `en.json:795` and `de.json:795` — both non-empty, so the parity gate (`ui/scripts/check-i18n.mjs:42-53`) passes. Every key referenced in code resolves: `m.clearmerged_probes_unavailable` (`ClearMergedDialog.svelte:61`) and `m.toast_decom_probes_unavailable` (`Viewport.svelte:961`, `+page.svelte:2270`). DE terminology matches the existing catalog (`stilllegen` per `de.json:702`, `Diagnose` per `de.json:1376`). No new glossary-worthy term is introduced, so `.claude/rules/ui-glossary.md` is not engaged. The feature-catalog gate keys on `feat(...)` subjects (`scripts/check-feature-catalog.sh:61`) and this branch's single commit is `fix(preview): …`, so it stays disarmed as the plan states.

### Nits (non-blocking):

- `ui/src/lib/components/Viewport.svelte:952` / `ui/src/routes/+page.svelte:2263`: the `.catch()` fallback pins `probesUnavailable: false`, so a leftovers request that *fails* still reads as "probes fine" — the same user-visible shape ("nothing running") the flag exists to prevent. Not a regression (the old code caught to a bare `[]`) and largely self-announcing, since a server that can't serve this GET will also fail `archiveSession` and toast `toast_decommission_failed` (`+page.svelte:2243`); `true` would arguably be the more truthful fallback.
- `ui/src/lib/components/ClearMergedDialog.svelte:55-62`: with `leftovers > 0` and the flag set, both the count line and the "aren't listed and may survive" caution render together, which reads slightly contradictory. Barely reachable today (the capability arm implies `detect → []`; only a port-less class-3 transcript hit under a `"none"` snapshot could combine them, per `src/process-reaper.ts:616-619`), so no action needed.

- `ui/src/routes/+page.svelte:2769` and `:2931` wire `ondecommission={onarchive}` for the Herd card menu, so that decommission path never probes leftovers at all — it neither shows nor claims a count, so it is a different (and older) gap than #1923's.
- `GET /api/sessions/:id/leftovers` still reads the snapshot synchronously with no `refreshForDetect`, so a `"stale"` cell can yield an outdated list that the new flag deliberately does not mark (consistent with preview-start, and called out as out of scope in the plan).

No cannot-verify items: every identifier, message key, and consumer above exists in this repo and was opened and compared.

PR #2025, opened 2026-08-04
feat(preview): arm macOS preview stop behind a kill-age bound (#1922)

scoop:
-


scoop (commented):

**The two new probe seams exist, are spelled consistently, and default permissive.**
`signalWindowOpen?()` and `liveProcForPid?()` are declared optional on `ReaperProbes` at `src/process-reaper.ts:155` and `src/process-reaper.ts:168`, supplied by the darwin backend at `src/proc-probes-darwin.ts:640-641`, and required in the `DarwinProbes` alias at `src/proc-probes-darwin.ts:366-372`. `stopListenersOnPort` consults them via `?.` (`src/process-reaper.ts:698`, `src/process-reaper.ts:705`), so on Linux (`linuxProbes` declares neither — checked the object literal ending at `src/process-reaper.ts:518`) both are absent and the path is byte-identical to before apart from the renamed return field. `nullProbes` (`src/process-reaper.ts:529-541`) has `canAuthorizeSignal: false` and no `liveProcForPid`, so it still short-circuits to `outcome: "unsupported"` at `src/process-reaper.ts:700-702`.

**The window gate really is checked before the cell is read.** `src/process-reaper.ts:705` returns `"refused"` before `normRoot`/`candidatePidsOn`, so an out-of-window cell yielding zero candidates cannot be reported as `ok/0`. `test/proc-probes-reaper.test.ts` asserts `scanProcs` is never called in that case.

**The kill-age bound is genuinely independent of the sweep.** `config.previewKillMaxAgeMs` is defined at `src/config.ts:929` from `SHEPHERD_PREVIEW_KILL_MAX_AGE_MS` (default `10_000`), read only by `maxKillAgeMs()` at `src/proc-probes-darwin.ts:352-355`; `maxNegativeAgeMs()` at `src/proc-probes-darwin.ts:338` still owns the sweep-derived bound and is unchanged. `signalWindowOpen` (`src/proc-probes-darwin.ts:419-422`) compares `now() - successAt <= maxKillAgeMs() + cell.sampleMs`; `sampleMs` is written on success only, at `src/proc-probes-darwin.ts:530`, floored with `Math.max(0, …)`. The NaN claim in the `DEFAULT_KILL_MAX_AGE_MS` comment (`src/proc-probes-darwin.ts:80-89`) is accurate: `x <= NaN` is false, so an unusable bound fails closed even without the fallback.

**`killPid` becoming real does not widen the blast radius.** `canAuthorizeSignal` is still `false` (`src/proc-probes-darwin.ts:639`), so `scanWorktreeProcs` returns `[]` (`src/process-reaper.ts:616`), `canDetectLeftovers()` is false (`src/process-reaper.ts:789-791`) and `reap()`'s pid branch refuses (`src/process-reaper.ts:852-856`). The other `killPid` callers are structurally unreachable on darwin as the header claims: `reapOrphansUnder` requires `ppidForPid` (`src/process-reaper.ts:828` — `undefined?.() ?? null !== 1` always skips), `reapDeletedWorktreeOrphans` the same (`src/process-reaper.ts:887`), and `reapMarkedOrphans` returns early without `environForPid`/`cpuStatForPid`/`uptimeSeconds` (`src/process-reaper.ts:1091-1092`). None of those are declared by `makeDarwinProbes` (returned object, `src/proc-probes-darwin.ts:584-647`).

**`liveProcForPid`'s argv and parsing line up with the existing parser.** `LSOF_VERIFY_ARGV` (`src/proc-probes-darwin.ts:68`) is `LSOF_ARGV` (`src/proc-probes-darwin.ts:51-62`) minus the host-wide selectors, so both runs emit the `-F pcfn` shape `parseLsofFields` reads (`src/proc-probes-darwin.ts:130-137`); the cwd run's `fcwd` line drives `st.fdIsCwd` at `src/proc-probes-darwin.ts:162`, the port run's numeric `f` lines leave it false so `n` values go through `parseListenPort` (`src/proc-probes-darwin.ts:207-212`). The pid guard at `src/proc-probes-darwin.ts:452` rejects non-integer/`<=1` before anything reaches `-p`, every failure mode returns `null`, and the comm is basenamed at `src/proc-probes-darwin.ts:462` to match the `AGENT_COMMS` semantics the cell already normalises to at `src/proc-probes-darwin.ts:526`.

**`execFileSync` resolves.** It is the instrumented re-export at `src/instrument.ts:93-100` (typed as `typeof nodeExecFileSync`), not node's — matching the import comment at `src/proc-probes-darwin.ts:5-7`. The added blocker line at `src/instrument.ts:116` is consistent with it.

**The `{signalled, unsupported}` → `{signalled, outcome}` contract change has no stale consumers.** The only production caller is `SessionService.stopPreview` (`src/service.ts:4747`), which was migrated (`src/service.ts:4739`); no other `stopListenersOnPort` reference exists in `src/` (grepped). `deps.reaper.refresh` is already optional-with-`?.` on the injected type (`src/service.ts:192-193`), so `refreshForStop()` (`src/service.ts:758-760`) type-checks against existing fakes.

**The HTTP handler being async is safe.** `handlePreviewStop` is dispatched from `handleSessions`, which does `const res = await sub(ctx)` (`src/server.ts:3805`). The pre-refresh 404 uses `deps.store.get(id)` and `deps.store`/`deps.service` are both on `AppDeps` (`src/server.ts:266`).

**UI + i18n parity.** `stopPreview` maps both new 409s at `ui/src/lib/api.ts:2466-2467` and its only consumer is `Viewport.svelte` (`ui/src/lib/components/Viewport.svelte:44`), whose branches at `ui/src/lib/components/Viewport.svelte:1353-1368` precede the `res.killed === 0` branch, so the union narrows correctly. Keys exist in both catalogs: `viewport_preview_stop_unsupported`/`viewport_preview_stop_refused` at `ui/messages/en.json:538-539` and `ui/messages/de.json:538-539`; `feat_macos_preview_stop_title`/`_body` at `ui/messages/en.json:3525-3526` and `ui/messages/de.json:3525-3526`. `ui/scripts/check-i18n.mjs:44-53` only requires cross-locale key parity and non-empty values — satisfied. The feature entry needs no registry edit: the catalog is an `import.meta.glob` over `entries/*.ts` (`ui/src/lib/feature-announcements.ts:47-50`), and `sinceVersion: "1.47.0"` is ahead of `package.json:3` (`1.46.0`) and matches the other unreleased 1.47.0 entries.

**Poller ladder.** `escalateIdleStop` treats both non-advancing outcomes identically and returns before touching `previewStopState` (`src/poller.ts:1867-1870`); `idleStopSilentLogged` is a reason-keyed `Map` (`src/poller.ts:297`) pruned on the same three paths as the old `Set` (`src/poller.ts:1039`, `src/poller.ts:1781`, `src/poller.ts:1796`). The idle path's freshness assumption holds: the cell is refreshed every tick (`src/poller.ts:592-596`) with a coalescing TTL of `max(250, sweepMs/2)` = 2s at the default (`src/proc-probes-darwin.ts:359-361`), comfortably inside the 10s bound.

The implementation matches what the issue asked for: a dedicated bound expressed relative to the observed sample duration, per-pid re-verification before the signal, and an observable `refused` that threads all the way to a toast without burning the SIGTERM → SIGKILL → gaveUp ladder.

## Nits (non-blocking):
- `src/process-reaper.ts:744-747` — `verifiesLive`'s doc says it re-applies "the same three tests `candidatePidsOn` applies", but `candidatePidsOn` applies four (`src/process-reaper.ts:737` also excludes `process.pid`). The omission is harmless (our own pid is invariant and was already excluded from the candidate set), but the count in the comment is off by one.
- `src/proc-probes-darwin.ts:446-447` — "never on the poller's hot path" sits one clause after "or the idle-stop ladder", which *is* driven from `runPreviewSweep`; the sentence reads as self-contradictory even though the cost analysis behind it is right (a cold/out-of-window cell refuses at `src/process-reaper.ts:705` before any spawn, so spawns only happen when `lsof` is demonstrably working).
- `src/poller.ts:1827` — on macOS a dev server that dies promptly from SIGTERM can, in a narrow window (cell sampled just before the exit, sweep just after), fail live re-verification on the next sweep and log `idle-stop refused … cannot reclaim <id> on :<port>` for a preview that was in fact reclaimed. Transient and self-clearing once the port disappears, but the line reads as a failure after a success.
- `src/server.ts:2899` — the force-refresh runs before `stopPreview` resolves `devPortFor`, so a stop against a session with no bound preview still pays a full-host `lsof` before returning 409 `not_bound`.

- The `"unsupported"` HTTP code, `api.ts` variant, toast and locale keys go beyond the issue's `refused`-only ask; the plan justifies it as the identical Windows mislabelling and it is one extra branch, so it reads as a legitimate author call.
- `ui/src/lib/feature-announcements/entries/v1.47.0-macos-preview-stop.ts` and its four catalog strings are not in the issue, but the repo's feature-catalog contract (`ui/src/lib/feature-announcements.ts:3-8`) requires an entry in the same PR as a user-facing feature.

- `src/index.ts:1077-1079` — the `idleStop.stop` wiring comment still says the ladder refuses to advance only on `"unsupported"`; it now also refuses on `"refused"`. Comment-only, in an untouched file.
- `README.md:289-291` — the README env table is a subset of `docs-site/src/content/docs/reference/configuration.md` and does not list the new `SHEPHERD_PREVIEW_KILL_MAX_AGE_MS` row (nor several other existing vars); no gate enforces parity between the two.
- `docs/toast-inventory.md` is a committed inventory with no script or CI reference wiring it to `scripts/audit-toasts.mjs`, so it drifts silently as toasts are added.

## Stated limitations (could not verify here)
- Apple's `lsof` is not present in this environment, so I could not execute the `-a` AND semantics that `liveProcForPid` depends on, nor confirm that a matching `-a -p <pid> -d cwd` run exits 0 on macOS (a non-zero exit would make `defaultVerifyRunner` return `null` and refuse every stop). This is the plan's own headline risk; it fails closed either way, and `test/proc-probes-darwin.macos.test.ts:98-146` is the darwin-only CI assertion that turns a divergence into a red build. Assumed as written.
- Per the review instructions I did not run lint or the test suite.

scoop:
## What I checked, and against what

**The signal path itself (`src/process-reaper.ts:693-760`).** `stopListenersOnPort` returns `"unsupported"` only when `canAuthorizeSignal === false` *and* `liveProcForPid` is absent — `nullProbes` (src/process-reaper.ts:529-535) declares neither seam, so Windows keeps reporting `"unsupported"`; the darwin backend supplies both (src/proc-probes-darwin.ts:639-641) and so proceeds. Linux: `makeDefaultProbes` (src/process-reaper.ts:550) hands back the live-`/proc` probes, where `signalWindowOpen` is absent (`?.() === false` is false → proceed) and `verify` is undefined (`verifiesLive` skipped, `rejected` stays 0 → `outcome: "ok"`), so the Linux path is behaviourally identical to the pre-diff code apart from the field rename. Verified `killPid`'s other call sites at src/process-reaper.ts:831, 856, 892, 1129 — each is gated on `ppidForPid` (omitted on darwin), `canAuthorizeSignal` (still `false`, src/proc-probes-darwin.ts:637), or `cpuStatForPid` (omitted), so arming `killPid` (src/proc-probes-darwin.ts:594-604) does not widen any other reap.

**The window gate ordering.** `signalWindowOpen()` is consulted before `candidatePidsOn` iterates (src/process-reaper.ts:705 vs 711), so an out-of-window cell yielding zero candidates cannot be reported as `ok`/0. Pinned by test/proc-probes-reaper.test.ts:80-99, which asserts `scanProcs` is never called.

**The age arithmetic.** `cell.successAt` is the START stamp (src/proc-probes-darwin.ts:529) and `cell.sampleMs = Math.max(0, now() - startedAt)` (src/proc-probes-darwin.ts:530), so `now - successAt <= maxKillAgeMs() + sampleMs` (src/proc-probes-darwin.ts:424) matches the plan's formula. `maxKillAgeMs()` (src/proc-probes-darwin.ts:328-331) falls back for non-finite/negative values; a `NaN` bound would make the `<=` false, i.e. window closed — the inverted comparison relative to the plan's `>` still fails closed. `previewKillMaxAgeMs` exists at src/config.ts:929 and is read at call time, so the injected-clock tests work.

**Live re-verification.** `liveProcForPid` (src/proc-probes-darwin.ts:449-463) basenames the comm with `commBasename`, the same normalisation the cell applies at src/proc-probes-darwin.ts:526, so `AGENT_COMMS.has(live.comm)` in `verifiesLive` (src/process-reaper.ts:757) compares like for like. Both cwd values (cell and live) come straight from lsof's kernel-resolved `fcwd`, and both are compared against `normRoot(...)` → `realpathSync` (src/proc-probes-darwin.ts:620-630, src/process-reaper.ts:251-254), so the darwin `/var`→`/private/var` case is handled identically on both sides. `defaultVerifyRunner` collapses every failure to `null` (src/proc-probes-darwin.ts:286-300) and `execFileSync` resolves to the instrumented wrapper at src/instrument.ts:93, which is a pure passthrough to `node:child_process`.

**Thread-through.** `stopPreview`'s widened union (src/service.ts:4736-4757) is consumed at src/server.ts:2909-2912 and, via `src/index.ts:1080`, by `StatusPoller`'s `StopPreviewOutcome` (src/poller.ts:60-64); `escalateIdleStop` returns before touching `previewStopState` on both silent outcomes (src/poller.ts:1867-1871). `ROUTE_HANDLERS` dispatch does `await handle(ctx)` (src/server.ts:7657), so making `handlePreviewStop` async is safe. `refreshForStop` reaches `ProcessReaper.refresh` (src/process-reaper.ts:767) which no-ops when the probes declare none, and the `deps.reaper` type at src/service.ts:193 already carries `refresh` as `Partial`. `api.ts` is the only consumer of `stopPreview` in the UI (`ui/src/lib/components/Viewport.svelte:44` is the sole import), and its two new 409 branches (ui/src/lib/api.ts:2476-2477) are handled at ui/src/lib/components/Viewport.svelte:1367-1380 before the `res.killed === 0` branch.

**i18n parity.** `viewport_preview_stop_unsupported` / `viewport_preview_stop_refused` exist at ui/messages/en.json:3547-3548 **and** ui/messages/de.json:3547-3548; the announcement keys `feat_macos_preview_stop_title` / `_body` exist at 3549-3550 in both. The announcement entry's `sinceVersion: "1.47.0"` is ahead of `.release-please-manifest.json`'s `1.46.0`, its filename matches `v<version>-<id>.ts`, and `ui/src/lib/feature-announcements.ts:47` picks entries up with an eager `import.meta.glob`, so no index edit is needed. The `„…\"` quote mix in the new de string matches the existing precedent at ui/messages/de.json:30.

**Where it falls down:** `docs-site/src/content/docs/reference/configuration.md` — the PR adds the `SHEPHERD_PREVIEW_KILL_MAX_AGE_MS` row and rewrites the `SHEPHERD_PREVIEW_IDLE_STOP_MS` row, but the prose four lines above the same table (line 47) still tells operators that stopping a preview from the UI is unavailable on macOS. See the finding.

## Could not verify

- Apple's `lsof` `-a` AND semantics and its exit status for `-p <pid> -a -d cwd`: no macOS host here. `test/proc-probes-darwin.macos.test.ts:117-150` is the right shape of assertion for it, and the failure mode is fail-closed (`defaultVerifyRunner` returns `null` on any non-zero exit, src/proc-probes-darwin.ts:296-299) — a refusal, never a mis-kill. Assumed as written.

## Nits (non-blocking):

- `src/proc-probes-darwin.ts:243-252`: the `DEFAULT_KILL_MAX_AGE_MS` doc says the fallback is "not a safety guard" because an unusable bound already fails closed via NaN. That holds for NaN, but `SHEPHERD_PREVIEW_KILL_MAX_AGE_MS=Infinity` is also non-finite and would make `age <= Infinity + sampleMs` always true — fail OPEN — so the `Number.isFinite` half of the guard genuinely is load-bearing.
- `src/proc-probes-darwin.ts:252` hardcodes `10_000`, duplicating the same default at `src/config.ts:929`; two literals to keep in sync.
- `src/proc-probes-darwin.ts:445-446`: "never on the poller's hot path" — the idle-stop ladder that reaches these sync spawns is driven from `StatusPoller`'s preview sweep (`src/poller.ts:1857`), so they can land on the poller's thread, just not on the per-tick refresh.
- `src/server.ts:2908`: `refreshForStop()` runs before the `not_bound` check, so a stop on an unbound session still pays a full-host `lsof` on darwin.

- `src/process-reaper.ts:718-729`: candidates that pass live verification but whose `killPid` throws increment neither `count` nor `rejected`, so an all-throwing run reports `ok`/0 — rendered to the operator as "no dev-server process was found to stop". Benign on Linux (already-exited race) and unlikely on darwin (lsof as a non-root user mostly surfaces the caller's own processes), but a `process.kill` EPERM there would now be swallowed into exactly the mislabelling this PR exists to remove.

scoop:
**Blocking finding 1 — valid, and addressed. Partly before this note, partly by it.**

You were right on both halves, and I want to be precise about what was already true at the commit you reviewed vs. what I just changed.

At `ee5fc879` (the push you reviewed) **both** stale claims were present — `configuration.md` still said stopping a preview from the UI is unavailable on macOS, and the cross-referenced Getting-started platform row said the same. My feat commit had touched only the two table rows, never that prose. The finding was correct.

Between your review and now, the doc-agent landed `d849a7c7` ("docs: sync docs to recent source changes"), which fixed exactly those two. I reviewed that commit rather than assuming an auto-generated fix was right, and it is — including a subtlety worth calling out: the Getting-started sentence had **conflated two claims** ("stopping a preview from the UI, and exposing one over the tailnet, are unavailable"). Only the first is invalidated by this PR; tailnet exposure is still genuinely unavailable because it needs the `tailscale` CLI. The fix correctly splits the sentence instead of flipping both.

**What your finding surfaced that neither of us had listed:** the same platform table is mirrored in `README.md:158`, carrying the identical now-false claim, and nothing had corrected it. Fixed in `36a8e634` with the same split-sentence treatment.

I then swept every doc surface for the claim; all three locations now agree and nothing else references macOS preview-stop as unavailable:

- `README.md:158` — fixed here
- `docs-site/.../getting-started.md:46` — fixed in `d849a7c7`
- `docs-site/.../reference/configuration.md:47` + the three table rows — fixed in `d849a7c7` / this PR

Also worth recording for the re-review: the `darwin probes` CI lane is green and its log confirms the two new macOS assertions actually executed rather than skipping (`liveProcForPid re-reads a real process — Apple's lsof honours -a AND semantics`, and the gone-pid case). That was the riskiest assumption in the change — the `-a` AND semantics were verified against GNU lsof while building, and Apple ships a different, much older implementation.

scoop:
## Prior-round finding

**1. `docs-site/.../configuration.md:47` still claimed macOS stop is unavailable — ADDRESSED.** The paragraph now reads "stopping one from the UI works but is bounded — the snapshot must be within `SHEPHERD_PREVIEW_KILL_MAX_AGE_MS` and the candidate process is re-checked live before any signal" (`docs-site/src/content/docs/reference/configuration.md:47-51`), consistent with the new `SHEPHERD_PREVIEW_KILL_MAX_AGE_MS` row (`:61`) and the rewritten `SHEPHERD_PREVIEW_IDLE_STOP_MS` row (`:63`). The cross-referenced Getting-started platform row was corrected in the same way (`docs-site/src/content/docs/getting-started.md:46`). I grepped the rest of `docs-site/src/content/docs` for `macOS`/`darwin` and the only other preview mention is `operating.md:88-97`, which describes detection only and makes no stop claim — nothing else contradicts.

## What I verified

**Blast radius really is narrow (the riskiest claim in the PR).** `canAuthorizeSignal` is still `false` (`src/proc-probes-darwin.ts:639`), and every other consumer of `killPid` on darwin is independently gated: `scanWorktreeProcs` returns `[]` at `src/process-reaper.ts:616`; `reap()`'s pid branch is gated at `src/process-reaper.ts:852,856`; `reapOrphansUnder` requires `ppidForPid`, which darwin omits, so `(undefined ?? null) !== 1` always skips (`src/process-reaper.ts:829`), same for `reapDeletedWorktreeOrphans` (`:890`); `reapMarkedOrphans` returns `none` without `environForPid`/`cpuStatForPid`/`uptimeSeconds` (`src/process-reaper.ts:1091-1092`), all omitted by the darwin backend (module header, `src/proc-probes-darwin.ts:20-25`). So making `killPid` real (`src/proc-probes-darwin.ts:599-608`) leaves exactly one reachable caller.

**Linux is byte-identical apart from the return-shape rename.** `linuxProbes` supplies neither `signalWindowOpen` nor `liveProcForPid`, so `this.probes.signalWindowOpen?.() === false` is `false` and `verify` is `undefined` (`src/process-reaper.ts:698,705,714`); the candidate loop is the same three tests as before, just moved into `candidatePidsOn` (`src/process-reaper.ts:735-740`). `nullProbes` (`src/process-reaper.ts:529-541`) has `canAuthorizeSignal: false` and no `liveProcForPid`, so it still short-circuits — now as `outcome: "unsupported"` (`:700-702`).

**Contract change is consistently threaded.** `stopListenersOnPort` → `src/service.ts:4747-4757` (`outcome !== "ok"` → `{result: outcome, killed: 0}`) → `src/poller.ts:112` / `1867-1871` (ladder does not advance, `previewStopState` untouched) → `src/server.ts:2912` (409 with the raw code) → `ui/src/lib/api.ts:2477-2480` → `ui/src/lib/components/Viewport.svelte:1367-1385`. I grepped for other `stopListenersOnPort` callers: only `src/service.ts:4747` (`src/service.ts:192` types the injected reaper as `Pick<…,"detect"|"reap"|"stopListenersOnPort">`), so no consumer still reads the removed `unsupported` boolean.

**`refreshForStop` wiring.** `src/service.ts:4785-4787` calls `this.deps.reaper?.refresh?.({force:true})`; `refresh` is in the `Partial<Pick<…>>` at `src/service.ts:193`, and `ProcessReaper.refresh` delegates to `this.probes.refresh?.()` (`src/process-reaper.ts:767-769`), which `linuxProbes` does not define — genuinely a no-op on Linux. `handlePreviewStop` became `async` (`src/server.ts:2896`) and its dispatcher already awaits each sub-handler (`src/server.ts:3804` in the list, `:3819` `await sub(ctx)`).

**i18n parity.** `viewport_preview_stop_unsupported` and `viewport_preview_stop_refused` exist in both catalogs at the same positions — `ui/messages/en.json:3547-3548` and `ui/messages/de.json:3547-3548`; likewise `feat_macos_preview_stop_title`/`_body` (`en.json:3549-3550`, `de.json:3549-3550`). The `.claude/rules/i18n.md:38` gate only asserts identical non-empty key sets, which holds.

**Feature-catalog contract.** `ui/src/lib/feature-announcements/entries/v1.47.0-macos-preview-stop.ts:4-7` uses id `macos-preview-stop`, `sinceVersion: "1.47.0"`, and the two keys above; `package.json` is at `1.46.0`, so `1.47.0` is the next unreleased version as `ui/src/lib/feature-announcements.ts:11-16` requires, and the filename matches the `v<sinceVersion>-<id>.ts` form (`:18`).

**`execFileSync` import resolves.** `src/proc-probes-darwin.ts:7` imports from `./instrument`, exported at `src/instrument.ts:93` as a `typeof nodeExecFileSync` passthrough, so the `encoding: "utf8"` overload still returns `string`. The startup-blocker line was updated accordingly (`src/instrument.ts:116`).

**Parser reuse for the verify runs is sound.** `feedLine` sets `fdIsCwd` from the `f` field (`src/proc-probes-darwin.ts:161-163`), so the `-d cwd` run's `n` line lands in `cwd` and the `-iTCP -sTCP:LISTEN` run's `n` lines land in `ports` — the two-run join at `:454-462` reads the right field from each. The `find(p => p.pid === pid)` guards reject output for another pid, and every failure mode (`null` from the runner, empty parse, empty cwd, empty port set, non-integer/≤1 pid) returns `null` (`:452-461`), i.e. fail-closed.

### Could not verify
- Whether Apple's `lsof` honours `-a` AND semantics for `-p` + `-d cwd` / `-p` + `-iTCP`: the ground truth is a macOS binary not present in this Linux repo. `test/proc-probes-darwin.macos.test.ts:117-150` is the darwin-only assertion for it, and the failure mode is fail-closed (`liveProcForPid` returns `null` → `refused`), never a mis-kill. Assumed as written.

## Findings

### 1. `src/process-reaper.ts:604-605` — comment states a fact this PR made false
```
// A backend that cannot authorize a signal (darwin) also cannot HONOUR one:
// `reap()` calls `probes.killPid`, which is a no-op there.
```
Darwin's `killPid` is now a real `process.kill` (`src/proc-probes-darwin.ts:599-608`). The author edited the tail of this very comment block (the `#1922` paragraph at `:612-615`), so the block now contradicts itself: the opening premise says the signal cannot be honoured, the closing paragraph says the arming was deliberately kept away from this path. The behaviour is still safe — the gate at `:616` is `canAuthorizeSignal === false`, which holds — but the stated reason is wrong, and it is exactly the kind of premise a later change would lean on.

### 2. `src/poller.ts:97-100` — `idleStop.stop` doc misattributes `"unsupported"` and omits `"refused"`
```
 *  outcome; `"unsupported"` (darwin, no signal authority) must NOT advance the
 *  escalation ladder. Absent = disabled. */
```
After this PR darwin no longer returns `"unsupported"` — that outcome comes only from `nullProbes` (`src/process-reaper.ts:700-702`, `:529-541`) — and the outcome darwin *does* return, `"refused"`, is equally non-advancing (`src/poller.ts:1867-1871`). The `StopPreviewOutcome` doc immediately below (`src/poller.ts:107-110`) was updated correctly; this sibling doc, 10 lines above it in the same file, was not.

## Nits (non-blocking):
- `src/service.ts:4731-4734`: the `Returns:` list on `stopPreview` still enumerates only `not_found` / `not_bound` / `stopped`, though the signature it documents (`:4740`) now has five states.
- `ui/messages/de.json:3550`: `feat_macos_preview_stop_body` opens with the German low quote `„Devserver stoppen` but closes with a straight `"` — mismatched typography for the locale.
- `src/config.ts:929`: `SHEPHERD_PREVIEW_KILL_MAX_AGE_MS` is documented in `docs-site` but not added to the env table in `README.md` (which does carry the other `SHEPHERD_PREVIEW_*` knobs); no CI gate enforces that table, so this is a consistency preference.
- `ui/src/lib/components/Viewport.svelte:1370,1377`: two new `toasts.info` call sites make the reconciliation counts in `docs/toast-inventory.md` stale; `scripts/audit-toasts.mjs` only gates the removed `duration: null` spelling, so nothing fails.
- `src/process-reaper.ts:705`: a backend that supplied `liveProcForPid` but no `signalWindowOpen` would signal with no age bound at all. Not reachable today (darwin supplies both, `src/proc-probes-darwin.ts:640-641`), but the two seams are independently optional.
- `src/proc-probes-darwin.ts:441-447`: on darwin the idle-stop ladder path now performs up to two blocking `spawnSync` `lsof` calls (≤2 s each) per candidate from inside the poller tick (`src/poller.ts:1859` is reached from `sweepPreviews`). Opt-in (`previewIdleStopMs` defaults to 0) and bounded, but it is new event-loop-blocking work on a timer-driven path.

- `src/proc-probes-darwin.ts:421`: the window is `age <= maxKillAgeMs() + cell.sampleMs`, so the worst case applies to the row sampled *last*; a row sampled at the start of the run may be `maxKillAgeMs + sampleMs` old when it authorizes a SIGKILL. Bounded to ~13 s today because `sampleMs` is only written on a run that beat `REFRESH_TIMEOUT_MS = 3000` (`:73`, `:530`), but raising that timeout would silently widen the kill window — the exact coupling the dedicated knob was introduced to avoid.

scoop (commented):
## Prior findings — both addressed

**1. `src/process-reaper.ts` — stale "`killPid` is a no-op there" claim in `scanWorktreeProcs`.** Fixed. The comment now reads "`reap()` gates its pid branch on that same `canAuthorizeSignal` flag" and adds an explicit paragraph: "The gate is the flag, NOT an inert `killPid`: #1922 made darwin's a real `process.kill`" (src/process-reaper.ts:606-618). I verified both halves of the new claim against ground truth rather than the wording: `reap()` computes `const canSignal = this.probes.canAuthorizeSignal !== false` and only signals under it (src/process-reaper.ts:856-860), and darwin's backend still ships `canAuthorizeSignal: false` (src/proc-probes-darwin.ts:639) while `killPid` is now a real `process.kill` (src/proc-probes-darwin.ts:599-608). No contradiction remains with the #1922 paragraph at src/process-reaper.ts:614-618.

**2. `src/poller.ts` — `idleStop.stop` doc still said `"unsupported"` = "darwin, no signal authority".** Fixed. The doc now names both non-advancing outcomes and states which backend produces which: `"unsupported"` for `nullProbes`/Windows, `"refused"` for a snapshot backend, "Since #1922 darwin returns the latter, never the former" (src/poller.ts:98-103). That matches the code: `nullProbes` has `canAuthorizeSignal: false` and no `liveProcForPid` (src/process-reaper.ts:530-543) so it hits the `unsupported` branch at src/process-reaper.ts:703-706, while darwin supplies both seams (src/proc-probes-darwin.ts:640-641) and therefore reaches the `refused` returns at src/process-reaper.ts:709-710 / :729. Both are treated as non-advancing at src/poller.ts:1870-1873 via `isSilentStop` (src/poller.ts:1813-1815).

## What I verified this round

- **Identifiers resolve.** `execFileSync` is a real export of the instrumented wrapper (src/instrument.ts:93-100), not node's, so the new sync spawn is profiled; the matching blocker-map line was added (src/instrument.ts:116). `parseLsofFields` (src/proc-probes-darwin.ts:130) and `commBasename` (src/proc-probes-darwin.ts:653) both exist and are used consistently; `LSOF_VERIFY_ARGV` (src/proc-probes-darwin.ts:68) is `LSOF_ARGV` (src/proc-probes-darwin.ts:51-62) minus the host-wide selectors, so the same `-F pcfn` shape the parser reads. `config.previewKillMaxAgeMs` / `SHEPHERD_PREVIEW_KILL_MAX_AGE_MS` (src/config.ts:929) is spelled identically at its only consumer (src/proc-probes-darwin.ts:353) and in the docs row (docs-site/src/content/docs/reference/configuration.md:61).
- **Blast radius really is unchanged.** `canAuthorizeSignal` stays `false` on darwin (src/proc-probes-darwin.ts:639), so class-2 still returns `[]` (src/process-reaper.ts:620) and `canDetectLeftovers` still short-circuits (src/process-reaper.ts:793). I checked every remaining `killPid` caller now that darwin's is live: src/process-reaper.ts:833 and :894 both require `ppidForPid` (`?? null !== 1` ⇒ always skip, darwin omits it), src/process-reaper.ts:1096 returns early without `environForPid`/`cpuStatForPid`/`uptimeSeconds` (all omitted), and src/process-reaper.ts:860 is under `canSignal`. The module header's claim at src/proc-probes-darwin.ts:600-607 holds.
- **Contract change has no stragglers.** `{signalled, unsupported}` → `{signalled, outcome}` has exactly one production consumer, `SessionService.stopPreview` (src/service.ts:4747-4763), which was migrated; a tree-wide grep for `.unsupported`/`unsupported: true` in `src/` and `test/` returns only unrelated `diagnostics.ts:1105` hits. `refreshForStop` (src/service.ts:4775) uses `this.deps.reaper?.refresh?.()`, consistent with the `Partial<Pick<…, "refresh">>` dep typing at src/service.ts:192-193 and the real method at src/process-reaper.ts:771.
- **Ordering that the design leans on.** `signalWindowOpen()` is consulted before `scanProcs()` (src/process-reaper.ts:708-710 vs the generator at :731-739), so an out-of-window cell yielding zero candidates cannot be reported as a successful `killed: 0`; `test/proc-probes-reaper.test.ts` pins that by asserting `scanProcs` is never called. `verifiesLive` re-applies the reaper's own `isUnder`/`AGENT_COMMS`/port policy to the live re-read (src/process-reaper.ts:753-766), the same three tests `candidatePidsOn` applies to snapshot data.
- **Locale parity.** All four new keys exist in both catalogs: `viewport_preview_stop_unsupported`, `viewport_preview_stop_refused`, `feat_macos_preview_stop_title`, `feat_macos_preview_stop_body` — ui/messages/en.json:3547-3550 and ui/messages/de.json:3547-3550. The two toast keys are the ones referenced in ui/src/lib/components/Viewport.svelte:1369 and :1376; the two `feat_*` keys are referenced by the new announcement entry (ui/src/lib/feature-announcements/entries/v1.47.0-macos-preview-stop.ts:6-7), which is picked up automatically by the `import.meta.glob` catalog (ui/src/lib/feature-announcements.ts:48-51) and whose `sinceVersion: "1.47.0"` matches the other unreleased entries in that directory.
- **HTTP thread.** `handlePreviewStop` is async and dispatched through the awaiting loop at src/server.ts:7656-7658, so making it async is safe; the early `deps.store.get(id)` 404 (src/server.ts:2903) matches how every other handler resolves a session id (e.g. src/server.ts:2210) and mirrors `stopPreview`'s own `not_found` (src/service.ts:4741). The 409 codes emitted (src/server.ts:2911) are exactly the two `api.ts` maps (ui/src/lib/api.ts:2476-2477), which are exactly the two branches `Viewport.svelte` handles before the `killed === 0` case — so a refusal can no longer render as `viewport_preview_stop_nothing`.
- **Darwin-only CI assertion is genuinely gated** on `process.platform === "darwin"` (test/proc-probes-darwin.macos.test.ts:19), so it no-ops on the Linux lane rather than failing it.

Accepted as correct, not flagged: a candidate that verifies live and then throws on `process.kill` (it exited in between) counts as neither signalled nor rejected and reports `ok`/0 — documented at src/process-reaper.ts:726-728, and "nothing found to stop" is an honest rendering of that race.

Could not verify: macOS `lsof`'s `-a` AND semantics against a real Apple binary — not reproducible from this Linux checkout. The code fails closed either way (`liveProcForPid` returns null ⇒ refuse, src/proc-probes-darwin.ts:455-461), and test/proc-probes-darwin.macos.test.ts:117-150 is the assertion that would catch a divergence on the macOS lane.

### Nits (non-blocking):
- `src/proc-probes-darwin.ts:444-447` — "never on the poller's hot path" undersells that the idle-stop ladder it names *does* run inside `StatusPoller.tick` (src/poller.ts:1862 → src/index.ts:1080 → src/service.ts:4747), so up to 2 × `VERIFY_TIMEOUT_MS` of `spawnSync` per candidate can land on the event loop there; "never on the per-tick refresh path" would say it precisely.
- `src/server.ts:2903-2908` — the forced full-host `lsof` runs before the `not_bound` check, so a stop for a session with no bound preview still pays for a refresh on darwin; resolving bound-ness first would skip it.
- `ui/messages/de.json:3550` — `„Devserver stoppen\"` closes a German opening quote with a straight ASCII quote; the catalog elsewhere pairs `„…“` (e.g. ui/messages/de.json:162, :180).

Issue #1925, opened 2026-07-24
ProcessReaper.reap() has no pid-recycle guard


scoop:
Noted while working #1912 / PR #1921 — **pre-existing and platform-independent**, not introduced by that PR.

`ProcessReaper.reap()` signals the pids the operator confirmed in the decommission dialog. Those pids do not come from a live scan at signal time: they crossed the wire to the client as `leftoverKey`s and came back on confirmation, so the `detect` → `reap` window is **operator-paced and unbounded**. A pid recycled in that window is signallable.

`reapMarkedOrphans` already solves this on Linux with a starttime fingerprint (re-read `/proc/<pid>/stat` and compare `starttime` immediately before `killPid`). `reap()` should do the same.

PR #1921 narrowed the *input* to that window (it forces a probe refresh immediately before every `detect`) but deliberately did not change `reap()`'s semantics — the window itself is unchanged and out of scope for a detection-layer PR.

scoop:
✅ PR #2026 merged.

PR #2026, opened 2026-08-04
fix(reaper): guard every killPid site with a pid-recycle fingerprint (#1925)

scoop:
-

### DIFF
--- a/src/proc-probes-darwin.ts
+++ b/src/proc-probes-darwin.ts
@@ -42,6 +42,12 @@ import type { ReaperProbes } from "./process-reaper";
 // structurally unreachable here (the orphan reaps need the omitted `ppidForPid`,
 // the runaway reaper the omitted `environForPid`/`cpuStatForPid`/`uptimeSeconds`,
 // and `reap()` is gated on `canAuthorizeSignal`).
+//
+// Since #1925 `cpuStatForPid` carries a SECOND meaning: it is the starttime the
+// generic kill sites bracket their signals with. Omitting it therefore fails each
+// of them closed here a second time over, independently of the omissions above —
+// which is consistent, not redundant, because the one path that IS armed proves
+// its candidates the other way (below).
 
 /** The one `lsof` invocation the backend runs. `-d cwd` and `-iTCP` are different
  *  selection types, so lsof ORs them: one call yields, per process, its command,
--- a/src/process-reaper.ts
+++ b/src/process-reaper.ts
@@ -30,6 +30,19 @@ export interface Leftover {
   key: string;
   /** present for kind "process": the pid to kill */
   pid?: number;
+  /**
+   * present for kind "process": that pid's `starttime` (USER_HZ ticks since boot, /proc/<pid>/stat)
+   * as captured by `openSignalBracket` at detection — the kernel's own pid-recycle fingerprint.
+   * `reap()` re-reads it immediately before signalling and refuses on any mismatch, closing the
+   * bracket that the cwd/comm re-read and the `portsForPid` read sit inside. A pid recycled
+   * anywhere in the detect→kill window is therefore never signalled (bar same-tick reuse — see the
+   * pid-recycle note further down).
+   *
+   * ALWAYS SERVER-PRODUCED, never client-supplied: `SessionService.archive` re-runs `detect` and
+   * intersects by KEY, so the objects that reach `reap()` come from a fresh scan. The client only
+   * ever echoes back key strings. Absent ⇒ `reap()` refuses to signal (fail closed).
+   */
+  startTicks?: number;
   /** present for kind "system": the counter-command to run */
   command?: { bin: string; args: string[] };
 }
@@ -96,7 +109,12 @@ export interface ReaperProbes {
   cpuStatForPid?(pid: number): CpuStat | null;
   /** System uptime in seconds (/proc/uptime), or null when unreadable. */
   uptimeSeconds?(): number | null;
-  /** A process's cwd, for the log line only — never a gate. Null when unreadable. */
+  /**
+   * A process's cwd. Null when unreadable. Two uses: the runaway sweep's log line, and — since
+   * #1925 — the identity re-read inside `openSignalBracket`, where it IS a gate (absent ⇒ nothing
+   * is signalled). Must be a LIVE read on any backend that authorizes signals, or the re-read
+   * would just replay the same stale value `scanProcs` gave.
+   */
   cwdForPid?(pid: number): string | null;
   // ── snapshot-cell backends (darwin) ────────────────────────────────────────
   /**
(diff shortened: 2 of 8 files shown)


scoop:
## What this PR does, and what I verified

The change adds a `starttime` pid-recycle fingerprint to the four `killPid` sites in `src/process-reaper.ts` that lacked one, mirrors it onto `Leftover.startTicks`, and turns `reap()` from `void` into a terminations-performed count that `SessionService.archive` now reports.

Verified, with the ground truth I compared against:

- **All kill sites are covered.** `grep -n killPid src/process-reaper.ts` yields exactly four call sites: `src/process-reaper.ts:282` (`killReparentedOrphan`, new), `:748` (`stopListenersOnPort`), `:857` (`reap`), `:1129` (`reapMarkedOrphans`, pre-existing guard at `:1125-1127`). Each is preceded by a fingerprint comparison. Success criterion 1 holds.
- **`reap()`'s return-type change ripples correctly.** The only production caller is `src/service.ts:5468`; the dep type is `Pick<ProcessReaper, "detect" | "reap" | "stopListenersOnPort">` at `src/service.ts:192-193`, so it follows the method automatically — no dep-type edit needed. `archiveMany` sums it at `src/service.ts:5561`, and that method's doc at `src/service.ts:5529-5530` already claimed the number was "counted from what `archive` actually reaped" — which this PR makes true rather than aspirational.
- **The fingerprint is never client-supplied.** `handleSessionDelete` filters the body down to `string[]` (`src/server.ts:2549-2552`), `archive(id, reapKeys?: string[])` takes only keys (`src/service.ts:5454-5457`), and re-detects + intersects by key at `src/service.ts:5467`. So every `Leftover` reaching `reap()` is server-produced, as the new doc at `src/process-reaper.ts:38-41` claims.
- **`leftoverKey`'s format is unchanged** — `src/process-reaper.ts:215-217` still `Pick`s only `kind | name | port | pid`.
- **UI mirror parity.** `ui/src/lib/types.ts:2333` adds `startTicks?: number`, matching `src/process-reaper.ts:43`. No gate compares these two: `scripts/check-model-mirror.mjs:42-43,48` compares `src/types.ts` ↔ `ui/src/lib/types.ts` for `CLAUDE_MODELS`/`CODEX_MODELS`/`EFFORTS` only. `grep -rn "interface Leftover"` finds only those two declarations, so the hand-mirror is complete.
- **A legitimate `starttime` of 0 is not eaten.** `startTicksFor` uses `?? null` (`src/process-reaper.ts:246`) and every caller tests `=== null` (`:277`, `:665`, `:742`), and `isSameProcess` uses `captured != null` (`:263`). No falsy-zero bug.
- **Backend fail-closed reasoning is accurate.** `linuxProbes` supplies `cpuStatForPid` (`src/process-reaper.ts:533-535` → `readCpuStat` at `:395-422`); `nullProbes` (`src/process-reaper.ts:569-581`) sets `canAuthorizeSignal: false` and omits `cpuStatForPid`; `makeDarwinProbes` omits both `ppidForPid` and `cpuStatForPid` and sets `canAuthorizeSignal: false` (`src/proc-probes-darwin.ts:16-26,432-433`). `makeDefaultProbes` (`src/process-reaper.ts:591-594`) dispatches only those three, and every production caller uses the module default (`src/index.ts:551,874`, `src/worktree.ts:67,542`, `src/service.ts:4744`). So the new fail-closed drops change nothing on any shipping backend.

## Finding — the guard does not bind the identity it says it binds

`killReparentedOrphan`'s doc (`src/process-reaper.ts:269-271`) says it confirms the process "is still the process the caller's cwd filter matched", and the module note at `src/process-reaper.ts:238-239` says the idiom "brackets the read where the wall-clock actually goes, rather than only the last few instructions."

Neither is what the code does. The cwd (and the `" (deleted)"` suffix) come from the **one `scanProcs()` snapshot** taken at the top of each loop — `src/process-reaper.ts:474-499` for the Linux backend: a `readdirSync("/proc")` plus a `readlinkSync(cwd)` and a `readFileSync(comm)` for *every pid on the host*. The earliest `starttime` the code ever reads is taken later, inside the per-candidate body (`:277`, `:665`, `:742`). A pid recycled between its `readlink` in `scanProcs` and the capture is fingerprinted **as the recycled process**, matches itself at verify time, and is signalled — which is precisely the hazard this PR exists to close.

The wall-clock claim is inverted for the two orphan sweeps specifically: the read the fingerprint brackets there is `ppidForPid`, a single `readFileSync("/proc/<pid>/status")` (`src/process-reaper.ts:371-380`), while the unbracketed `scanProcs()` is the full-host enumeration above. The guard closes the microsecond segment and leaves the millisecond one open — and these two sites send `SIGKILL`, on identity data (`cwd` under a worktree / `" (deleted)"`) that is never re-read. `reapMarkedOrphans`, the pattern being copied, does not have this gap: it captures `starttime` in the same live per-pid pass as its `comm` gate (`src/process-reaper.ts:1027-1031`), so nothing stale qualifies a candidate.

The same overstatement is on `Leftover.startTicks` (`src/process-reaper.ts:35-38`): "a pid recycled inside the detect→kill window is never signalled" — the capture sits partway through `detect`, so a recycle in the `scanProcs`-to-capture part of that window is not caught.

Two acceptable resolutions: either bind the scanned identity to the fingerprint (re-read the cwd via `probes.cwdForPid` alongside the `starttime` capture and re-check `isUnder`/the `(deleted)` suffix, so a stale-snapshot candidate is rejected), or narrow the three claims above to state the residual `scanProcs`→capture window honestly. I'd take the first for the two SIGKILL sweeps and the second elsewhere, but either makes the file's documentation true.

- `src/process-reaper.ts:781-784` — `canDetectLeftovers()` still derives class-2 capability solely from `canAuthorizeSignal`, but `scanWorktreeProcs` now *also* suppresses class 2 outright when `cpuStatForPid` is absent (`:665`); a backend with signal authority but no `cpuStatForPid` would report `true` and pay a full refresh (`src/service.ts:4773`) for a detector that can never return a class-2 hit. Dormant today — `linuxProbes` supplies the probe, and both other backends already set `canAuthorizeSignal: false`.

## Nits (non-blocking):

- `test/proc-probes-reaper.test.ts:102` — the belt-and-braces `reap()` assertion in "no signal authority ⇒ offers zero class-2 leftovers" is now satisfied by two independent gates (no authority *and* no fingerprint). The sibling test at `:113` got a `cpuStatForPid` override for exactly this reason; the same override here would keep this one isolating signal authority too.
- `src/process-reaper.ts:277-278` — in `killReparentedOrphan` the capture precedes the PPID-1 filter, so every process under the worktree root pays a `/proc/<pid>/stat` read even though the overwhelming majority are rejected by a filter that costs one equally cheap read. Given `ppidForPid` is a single small file read, capturing after the PPID check would cost near-zero coverage; the current order is defensible, just not the cheap one.

scoop:
## Prior finding — resolved

**1. Fingerprint not bound to the scanned cwd/comm identity.** Addressed by `openSignalBracket` (`src/process-reaper.ts:286-293`): it captures `starttime` first, then re-reads cwd and comm and requires byte equality with what `scanProcs` reported. I walked the timeline — scan(T0) → capture(T1) → cwd/comm re-read(T2) → per-candidate probe(T3) → verify(T4) → kill:

- recycle in [T0,T1): capture is the new occupant, but the T2 cwd/comm read is the new occupant's too and diverges from the scanned strings → refused;
- recycle in [T1,T2) or [T2,T4): capture ≠ verify → refused at `isSameProcess` (`:304-310`).

That is only meaningful if the re-read is genuinely live and produces the *same string form* as `scanProcs`. Verified: `linuxProbes.scanProcs` readlinks `/proc/<pid>/cwd` (`src/process-reaper.ts:533`) and reads `/proc/<pid>/comm` trimmed (`:539`); `cwdForPid` (`:585-591`) and `commForPid` (`:569-575`) are the identical calls. So the `" (deleted)"` suffix that qualifies `reapDeletedWorktreeOrphans` candidates (`:937-944`) survives the re-read byte-for-byte and does not spuriously fail closed.

All four sites now use it: `scanWorktreeProcs` `:712-713` (bracket opened before `portsForPid`, `startTicks` carried onto the leftover at `:721`, verified in `reap` at `:902`), `stopListenersOnPort` `:788-792`, and both PPID-1 sweeps via `killReparentedOrphan` `:322-333` (`:868`, `:942`). `reapMarkedOrphans` `:1171-1175` already had its own capture/re-check and is untouched.

## What else I verified

- `readCpuStat` (`src/process-reaper.ts:447-468`) slices after `lastIndexOf(")")` and takes index 19 = `/proc/<pid>/stat` field 22 = `starttime`, and returns null unless every field is finite — so a comm containing `)` or spaces cannot yield a bogus fingerprint; it fails closed instead.
- The only producer of a `kind: "process"` leftover is `scanWorktreeProcs` (`src/process-reaper.ts:716-723`), so no other path can hand `reap()` a fingerprint-less pid leftover.
- The only caller of `reap()` is `SessionService.archive` (`src/service.ts:5468`); `src/worktree.ts:542` calls `reapOrphansUnder`, `src/tab-reaper.ts:230` is an unrelated `deps.reap`. The plan's correction to the issue's premise holds: `archive` re-detects and intersects by key (`src/service.ts:5466-5468`), so `startTicks` is always server-produced, matching the `Leftover` doc claim at `src/process-reaper.ts:41-43`.
- Return-type change: `deps.reaper` is `Pick<ProcessReaper, "detect" | "reap" | "stopListenersOnPort">` (`src/service.ts:192`), so `reap(): number` propagates to every structural fake; the ones that are type-checked were updated (`test/service.test.ts:2615,2712,4227`, `test/server.test.ts:2150,2772`), and `test/server.test.ts:827`'s harness types `reap` as `any`.
- `reap()`'s branch shape is unchanged for the non-signalling cases: the pre-change code only reached the `else if (l.command)` arm when the leftover was *not* `process`-with-pid, and the new `continue`s preserve that (`src/process-reaper.ts:898-912`). A `run()` that throws is not counted; both kinds count otherwise, matching `archiveMany`'s Clear-merged total.
- UI mirror added at `ui/src/lib/types.ts:2331-2333`; the field is optional and no component reads it (`ui/src/lib/components/page/DecomLeftovers.svelte`, `LeftoverDialog.svelte` use `key`/`name`/`port` only). No generated schema or model-mirror gate covers `Leftover` (`scripts/check-model-mirror.mjs` is `CLAUDE_MODELS`/`CODEX_MODELS`/`EFFORTS` only per package.json:`check:model-mirror`).
- `leftoverKey` stays `process:<pid>` (`src/process-reaper.ts:223-225`), so the dialog→confirm selection is still filtered by pid alone. That is explicitly out of the plan's scope and is not a hole here: `archive` re-detects, so anything reaped must itself qualify as a class-2 leftover of that same doomed worktree at kill time, and the fingerprint it carries was captured microseconds earlier.

Could not verify: nothing external. Tests/lint left to CI as instructed.

## Nits (non-blocking):
- `src/process-reaper.ts:37-39` and `:245`: with byte-equality on cwd/comm the guarantee is "the signalled process satisfies every gate", not "no recycled pid is ever signalled" — a pid recycled in [scan, capture) into a process with byte-identical cwd *and* comm still passes; the carve-out names only same-tick reuse. Practically safe (every other gate — ports/ppid — is read from the new occupant too), but the two claims are broader than the code.
- `src/process-reaper.ts:289`: "a recycle before this point is caught by the verify" holds only for a recycle *after* the capture; one before it is caught by this cwd/comm read itself, not by the later verify.
- `test/proc-probes-reaper.test.ts:102`: the belt-and-braces `reap` assertion now passes for a second reason (the leftover carries no `startTicks`), i.e. the same green-for-the-wrong-reason the sibling test at `:113-119` was explicitly patched to avoid.

- `src/process-reaper.ts:113-117` requires `cwdForPid` to be a LIVE read on any backend that authorizes signals, but darwin's `cwdForPid`/`commForPid` read the very cell `scanProcs` maps (`src/proc-probes-darwin.ts:425-460`), so on any future snapshot backend that arms signalling the identity re-read silently degrades to a tautology — nothing asserts the requirement.
- `src/process-reaper.ts:788-792`: this branch predates `38270f0a` (#1922), which is on main and arms darwin's `stopListenersOnPort` per candidate via `liveProcForPid` while keeping `canAuthorizeSignal: false`. Merging this loop onto that one re-closes macOS preview stop, because `openSignalBracket` fails closed on a backend with no `cpuStatForPid` — worth handling deliberately in the conflict resolution rather than by accident.

scoop (commented):
## Prior finding

**1. Falsified comments about `reaped = hit.length` — ADDRESSED.** Both halves are corrected:
- `src/process-reaper.ts:726-734` no longer claims a refused offer is "counted as `reaped = hit.length`"; it now says the offer would "signal nothing" and explicitly notes the count is honest since #1925 because `reap()` returns terminations performed and `SessionService.archive` reports that.
- The new comment at `src/process-reaper.ts:749-754` makes the same claim correctly ("The COUNT is honest either way (`reap` returns terminations performed); it is the phantom menu entry that is worth avoiding").
- The test-side copy of the old claim was fixed too (`test/proc-probes-reaper.test.ts:212-215`).
Verified against `src/process-reaper.ts:1009-1028` (`reap` returns `terminated`) and `src/service.ts:5516` (`reaped = this.deps.reaper.reap(hit)`).

## What I verified

**Every identifier the bracket depends on exists on every backend that can reach it.** `openSignalBracket` (`src/process-reaper.ts:318-325`) needs `cpuStatForPid`, `cwdForPid`, `commForPid`. `linuxProbes` supplies all three — `commForPid` at `src/process-reaper.ts:601-607`, `cpuStatForPid` at `:611-613`, `cwdForPid` at `:617-623` — so the Linux happy path is not broken by fail-closed. `makeDarwinProbes` omits `cpuStatForPid` (`src/proc-probes-darwin.ts:621-647` defines `commForPid`/`cwdForPid` but no `cpuStatForPid`, sets `canAuthorizeSignal: false` at `:645`, supplies `liveProcForPid` at `:647`), so `startTicksFor` returns null on the first line and the bracket short-circuits before any snapshot read — no added `lsof` cost, and `stopListenersOnPort` still proves darwin candidates via `verifiesLive` (`src/process-reaper.ts:856-858`). `nullProbes` (`:651-663`) has no `cpuStatForPid` and an empty `scanProcs`, so it is unchanged.

**Byte-equality of the cwd/comm re-read is sound on Linux.** `scanProcs` reads `readlinkSync(/proc/<pid>/cwd)` and `readFileSync(/proc/<pid>/comm).trim()` (`:565-574`); `cwdForPid`/`commForPid` read exactly the same files the same way (`:601-607`, `:617-623`), including the same `""`-on-failure default for comm. `reapDeletedWorktreeOrphans` compares a `" (deleted)"`-suffixed scanned cwd against a live readlink that also carries the suffix (`:1053` vs `:323`), so the suffix does not cause a spurious mismatch.

**No teardown race introduced by `killReparentedOrphan`.** `WorktreeManager.remove` runs the sweep at `src/worktree.ts:542` and only unlinks the worktree afterwards (`src/worktree.ts:547-567`), synchronously in the same function — so the live cwd cannot flip to `" (deleted)"` mid-sweep and silently spare an orphan.

**Return-type change is consistently consumed.** `reap()` has exactly one production caller, `src/service.ts:5516`; `src/tab-reaper.ts:279`'s `deps.reap` is an unrelated seam (`src/tab-reaper.ts:230`). `archive()` returns `reaped` at `src/service.ts:5551` and `archiveMany` sums it at `src/service.ts:5609`, whose docstring already promised "counted from what `archive` actually reaped, so the number never overstates" (`:5577-5578`) — now true rather than aspirational. The deps type is `Pick<ProcessReaper, "detect" | "reap" | "stopListenersOnPort">` (`src/service.ts:192`), so it tracks the signature automatically. The remaining `reap: () => {}` fake at `test/server.test.ts:895` still compiles because `harnessWithReaper` types `reap` as `any` (`test/server.test.ts:829`).

**The new Linux `"refused"` outcome does not stall anything.** `stopListenersOnPort` can now return `"refused"` on Linux (`src/process-reaper.ts:873`, reachable via the fingerprint branch at `:858`) — the PR pins this deliberately at `test/process-reaper.test.ts:802-819`. I traced the consumers: `SessionService.stopPreview` maps it to `result: "refused"` (`src/service.ts:4753-4757`), `src/server.ts:2912` returns 409, and `StatusPoller.escalateIdleStop` leaves `previewStopState` untouched and retries on the next sweep (`src/poller.ts:1871-1873`), so a transient refusal self-heals rather than freezing the SIGTERM→SIGKILL ladder. Reachable only when a process is alive at `portsForPid` and gone/recycled microseconds later at `isSameProcess` — a defensible, deliberately tested trade, not a regression I would block on.

**UI mirror.** `ui/src/lib/types.ts:2324-2335` matches `src/process-reaper.ts:23-48` field-for-field (`kind`, `name`, `port`, `key`, `pid`, `startTicks`, `command`). `scripts/check-model-mirror.mjs:58` only gates `CLAUDE_MODELS`/`CODEX_MODELS`/`EFFORTS`, so no mirror gate applies here. No UI component reads `startTicks` (only `key` is echoed back, confirmed at `src/service.ts:5514-5515`).

**Success criteria 1-5 hold.** All four generic `killPid` sites are now bracketed: `scanWorktreeProcs` (`:755-756`), `stopListenersOnPort` via `candidatesOn` (`:891` + `:856-862`), and both PPID-1 sweeps via `killReparentedOrphan` (`:982`, `:1056`). `leftoverKey` is unchanged (`:255-257`), and the `kind: "system"` branch stays ungated and counted (`:1019-1022`).

## Nits (non-blocking):
- `src/process-reaper.ts:823-825` — the `stopListenersOnPort` docstring still says "The Linux backend only ever returns `"ok"`: both verification seams are absent there, which defaults to permissive". This diff falsifies it: Linux now has a third seam (the starttime bracket at `:856-858`) and can return `"refused"` at `:873` — which the PR's own test asserts at `test/process-reaper.test.ts:802-819` using a probe set with no `liveProcForPid`. Worth correcting so a future reader of `src/poller.ts:1871` does not conclude `"refused"` is darwin-only.
- `src/process-reaper.ts:898` — `verifiesLive`'s docstring still references `candidatePidsOn`, which this diff renamed to `candidatesOn` (`:884`); the old name no longer exists anywhere in the tree.
- `src/process-reaper.ts:870-872` — "A kill that threw (the process exited underneath us) is neither — that is the benign race the Linux path has always reported as `ok`" is now only half true: the same benign race caught by the fingerprint re-read (rather than by `killPid` throwing) reports `refused`.

- `src/process-reaper.ts:318-325` — the cwd/comm re-read inside `openSignalBracket` goes beyond the plan's "capture `starttime`, re-read `starttime`" step; the rationale at `:279-284` is sound (it closes the `scanProcs`-walk window the fingerprint alone cannot see) and it is the safe direction, but it is a wider change than the approved plan described.

PR #2030, opened 2026-08-04
perf: stop rebuilding every Vercel project on every commit

### LINKED ISSUE #2027: Cut Vercel build cost: docs sidebar is quadratic (1.5 GB dist) and all 3 projects rebuild on every commit
Shepherd's three Vercel projects (`shepherd-docs`, `shepherd-demo`, `shepherd-site`) account for **67.6% of all build machine usage** on the `zweizeiler` team. Two independent causes, both fixable in this repo.

Measurements are from the Vercel REST API for the current billing cycle (**Jul 22 → Aug 4 2026, 78 commits to `main`**) plus local build profiling at `01413a1b`.

---

## Cause 1 — the docs sidebar is quadratic

`docs-site/astro.config.mjs` runs `starlight-typedoc` over `entryPoints: ["../src"]` with `entryPointStrategy: "expand"`, which documents **all 231 `.ts` files / 96.5k LOC** of the server package. That emits **1826 API pages**, and they are wired into the Starlight sidebar through `typeDocSidebarGroup`.

Starlight renders the **full sidebar into every page**. So all 1849 pages carry ~1840 API links each — an O(n²) blow-up:

```
$ grep -o 'href="/api/' dist/getting-started/index.html | wc -l
1840
$ stat -c%s dist/getting-started/index.html
860917
```

`/getting-started` — the first page a new operator lands on — ships **861 KB of HTML**, roughly 95% of it a nav tree of internal server implementation (`src/store/classes/sessionstore`, `src/forge/github/classes/githubforge`, …). The single largest page is 1.86 MB.

### Measured locally (`bun run build` in `docs-site/`)

| variant | pages | build wall | `dist/` | `/getting-started` |
| today | 1850 | **61s** | **1.5 GB** | **861 KB** |
| drop `typeDocSidebarGroup` from the sidebar array | 1850 | **10.6s** | **80 MB** | **50 KB** |
| drop `starlight-typedoc` entirely | 24 | 1.9s | 2.5 MB | — |

Removing **one line** from the sidebar array — while still generating, building and serving all 1826 API pages — makes the build **5.8× faster** and the deploy artifact **19× smaller**. Nothing is removed from the site; the API pages stay live at their existing URLs and stay indexed by Pagefind.

### Fix

In `docs-site/astro.config.mjs`, replace the `typeDocSidebarGroup` entry with a single link:

```diff
       sidebar: [
         ...
         { label: "Reference", items: [ ... ] },
-        typeDocSidebarGroup,
+        { label: "API reference", link: "/api/readme/" },
       ],
```

Notes for whoever picks this up:

- `typeDocSidebarGroup` is still imported for its side effect of registering the plugin — check whether the import can be dropped once the sidebar entry is gone, or whether it must stay. `starlightTypeDoc({...})` in `plugins:` must stay either way.
- Verify the landing link target. The build emits `dist/api/readme/` and `dist/api/src/`; there is **no** `dist/api/index.html`, so `/api/` itself would 404.
- The comment block above `typeDocSidebarGroup` explains the group's intent and needs rewriting, not deleting — the API reference is still generated from `../src` and still cannot drift.

Beyond build minutes this also cuts Fast Origin Transfer, retained deployment storage (`deploymentsToKeep: 10` × 1.5 GB ≈ 15 GB → ~800 MB), and page weight for every real visitor.

---

## Cause 2 — all three projects rebuild on every commit

Every one of the 78 commits to `main` triggered a production deploy of **all three** projects:

| project | root | deploys | needed¹ | avg build |
| `shepherd-docs` | `docs-site` | 78 | 45 | 172s |
| `shepherd-demo` | `ui` | 78 | 49 | 42s |
| `shepherd-site` | `site` | 78 | **6** | 5s |

¹ commits touching anything that build actually reads.

`shepherd-site` spends **5h12m of billed build machine** rebuilding an unchanged 6-second site 72 times.

### Why Vercel's automatic skipping can't help

All three projects already have `enableAffectedProjectsDeployments: true`. It can never fire. Per [Vercel's monorepo docs](https://vercel.com/docs/monorepos):

> To use automatic build skipping, a monorepo must use npm, yarn, pnpm, or Bun workspaces with all packages explicitly defined in the workspace configuration. […] **Changes made outside of the defined workspace are treated as global changes and will trigger deployments for all applications in the repository.**

Shepherd is five independent packages with five lockfiles *by design* (see `CLAUDE.md`). There is no workspace config, so every commit is a "global change". Converting the repo to Bun workspaces to satisfy this is not proposed here — it would trade a deliberate structural choice for a build-minutes optimisation.

### Fix — a per-project `ignoreCommand`

`ignoreCommand` goes in the committed `vercel.json`, not dashboard state. Exit **0 cancels** the build, exit **1 continues** it. Vercel exposes `VERCEL_GIT_PREVIOUS_SHA` (the SHA of the last *successful* deployment) at build time specifically when an Ignored Build Step is configured — preferable to `HEAD^`, which is wrong whenever several commits land between deploys. The command runs from the project's Root Directory, hence the `../` paths.

`docs-site/vercel.json` — must cover everything the config reads: `syncDocs()` pulls in `../docs`, `../CLAUDE.md` and `../.claude/rules`, and TypeDoc reads `../src`:

```json
{
  "$schema": "https://openapi.vercel.sh/vercel.json",
  "framework": "astro",
  "ignoreCommand": "git diff --quiet $VERCEL_GIT_PREVIOUS_SHA HEAD -- ./ ../src ../docs ../CLAUDE.md ../.claude/rules"
}
```

`ui/vercel.json` — `build:demo` is `SHEPHERD_DEMO=1 vite build` and reads nothing outside `ui/` (`gen-docs-manifest.ts` does import from `../../docs-site`, but it is a separate `gen:docs` script with committed output, not part of the build):

```json
{ "ignoreCommand": "git diff --quiet $VERCEL_GIT_PREVIOUS_SHA HEAD -- ./" }
```

`site/vercel.json` — no references outside its root:

```json
{ "ignoreCommand": "git diff --quiet $VERCEL_GIT_PREVIOUS_SHA HEAD -- ./" }
```

Caveats to handle when implementing:

- `git diff --quiet` exits `128` on error (e.g. the previous SHA is missing from the shallow clone). That is *not* `1`, and Vercel's documented contract only promises "1 continues". Normalise to a definite build — e.g. wrap in a tiny script that exits `1` on anything other than a clean "no changes" result, so a git failure fails **open** into a build rather than silently skipping a needed deploy.
- Guard for an empty `$VERCEL_GIT_PREVIOUS_SHA` (first deploy after enabling) → build.
- Cancelled builds still count toward deployment and concurrent-build quotas — they just don't consume build machine minutes, which is the cost line here.
- An ignore step can be bypassed manually from the dashboard for a forced redeploy or production promotion.

---

## Projected effect

Billed minutes per build fit closely to `≈ 3.6 min fixed + ~4.2 × wall minutes` across the three projects — i.e. **fixed per-build overhead dominates**, so *not building* is worth much more than making a build faster. The model is fitted from three data points; treat the projections as estimates, the "now" column as measured.

| project | now | after both fixes | saved |
| --- | --- | --- | --- |
| `shepherd-docs` | 20h20m | ~4h17m | 16h03m |
| `shepherd-demo` | 10h04m | ~6h19m | 3h45m |
| `shepherd-site` | 5h12m | ~24m | 4h48m |
| **total** | **35h36m** | **~11h** | **~24h36m (−69%)** |

≈ **47% of the entire team's build machine usage**, or roughly **$6.70/month** at the current run rate against a $20 included credit.

Either fix alone is worth landing; they are independent and compose.

---

## Out of scope (verified, recorded so it isn't re-investigated)

The other billed lines are **not** attributable to this repo. All three Shepherd projects deploy zero Vercel Functions, use no flags, and have no `@vercel/speed-insights` dependency — so Fluid Provisioned Memory ($3.08), Fluid Active CPU ($1.89), Flag Requests ($1.02) and Observability Events ($1.84) come from other projects on the team (`flowagent`, `pulse`, …).

One piece of dead config worth clearing: Speed Insights is provisioned on all three Shepherd projects but reports `hasData: false` on each, because no project installs the package. Not billed, just noise.

---

## Suggested split

1. **`docs-site`: drop the API group from the sidebar.** One-line change plus comment rewrite; the single largest win and the only one that also improves the visitor experience.
2. **Add `ignoreCommand` to all three `vercel.json` files.** Independent of (1); include the fail-open wrapper.

Both were confirmed with the operator before filing: keep all 1826 API pages (sidebar entry only), and gate `shepherd-docs` on the full current input set including `../src`.

scoop:
Closes #2027.

Two independent fixes. Either is worth landing alone; they compose.

## 1 — the docs sidebar was quadratic

Starlight renders the **full sidebar into every page**, and the sidebar carried the whole TypeDoc API tree. So each of ~1850 pages shipped ~1840 links to internal server implementation. `/getting-started` — the first page a new operator lands on — was 861 KB of HTML, ~95% of it nav.

`docs-site/astro.config.mjs` now places **one** entry instead of the plugin's `typeDocSidebarGroup`. Measured locally, before → after:

| | before | after |
| --- | --- | --- |
| build wall | 55s | **10.9s** |
| `dist/` | 1.5 GB | **80 MB** |
| `/getting-started` | 861 KB | **50 KB** |
| `href="/api/` on that page | 1840 | **1** |
| pages emitted | 1849 | **1849** |
| Pagefind-indexed pages | 1849 | **1849** |

**Nothing is removed from the site.** All 1826 API pages are still generated from `../src` (so they still cannot drift), still built, still served at their existing URLs, still searchable — only the nav changed.

The entry is a `slug:`, not the `link:` the issue suggested: every other internal entry in that sidebar is a slug, and Starlight resolves it at build time, so a wrong target fails the build instead of shipping a dead link. It points at TypeDoc's own 231-module index at `/api/readme/` — note there is no `/api/` page, so `/api/` itself would 404.

Also dropped the plugin's now-dead `sidebar: { label, collapsed }` option, which only shaped the group nothing renders any more.

## 2 — all projects rebuilt on every commit

Vercel's automatic affected-project skipping can never fire here: it needs npm/yarn/pnpm/Bun workspaces, and this repo is five packages with five lockfiles **by design** (see `CLAUDE.md`), so every commit reads as a global change. Converting the repo to workspaces to satisfy it is explicitly not proposed.

Instead, each `vercel.json` gets an `ignoreCommand` pointing at one shared `scripts/vercel-ignore-build.sh`:

| project | Root Directory | gated paths |
| --- | --- | --- |
| `shepherd-docs` | `docs-site` | `./ ../src ../docs ../CLAUDE.md ../.claude/rules` |
| `shepherd-demo` | `ui` | `./ ../package.json` |
| `shepherd-site` | `site` | `./` |
| vanity redirect | `deploy/vanity` | `./` |

Two deviations from the issue, both found while implementing:

- **`ui` does read outside its root.** The issue says the demo build reads nothing outside `ui/`. It does: `ui/vite.config.ts` → `appVersion()` reads the **root** `package.json` for `__APP_VERSION__`, which `ui/src/lib/build-info.ts` surfaces in the ActionBar footer and mobile About. Without it in the gate, release-please version bumps would leave a stale version on the live demo. (`__GIT_SHA__` / `__RELEASE_DATES__` come from git state, not paths — not gateable, and a skipped build reporting the SHA it was built from is correct.)
- **`deploy/vanity` is a 4th Vercel project** the issue does not mention. Gated too.

### The gate

Exit codes are **inverted** — Vercel's contract is `exit 0` cancels the build, `exit 1` runs it. Get that backwards and every deploy stops silently, with no failing build to notice. Hence one reviewed script with a loud header rather than three one-liners inside JSON, and hence `test/vercel-ignore-build.test.ts`, which drives the script against a throwaway git repo and pins all ten exit paths.

It **fails open**: no previous SHA, a previous SHA missing from the clone, no path arguments, not a git repo, or any `git diff` status other than a clean 0/1 (it exits 128 on error, which is *not* the documented "1 continues") all resolve to *build*.

That last case is not theoretical. Vercel clones with `git clone --depth=10`, and `VERCEL_GIT_PREVIOUS_SHA` is only an env var — vercel/vercel#7251 shipped the variable, not a deeper clone. Once a project has skipped ~10 commits in a row (exactly `shepherd-site`) the previous SHA falls out of history and we build. Self-correcting: that build succeeds, the previous SHA becomes recent, the window reopens. **So the issue's ~24h/month projection is a ceiling, not a floor** — expect `shepherd-site` to build roughly every 10th commit rather than never. Deliberately not mitigated with a deepening `git fetch`: network dependence and new failure modes in a gate whose whole job is to be boring, for a bounded win.

## Verification

- `bun run build` in `docs-site/` — numbers in the table above; `dist/api/readme/index.html` and `dist/api/src/**` present; `pagefind-entry.json` reports 1849 indexed pages.
- `bun run check` in `docs-site/` — 0 errors, 0 warnings, 0 hints.
- Root `bun run lint` clean; root `bun run test` — 8258 pass, 0 fail.
- The gate smoke-tested against real repo history from each of the four root directories: correctly skips commits that touched nothing it reads, correctly builds when `../src` moved.

This merge commit touches all four gated path sets, so all four projects must deploy from it — that is the live check on the exit-code polarity.

## Out of scope

Converting to Bun workspaces; reducing what TypeDoc documents; Vercel dashboard state (the dead Speed Insights provisioning, `deploymentsToKeep`, `enableAffectedProjectsDeployments`); the other billed lines, which #2027 already traced to other projects on the team.

```shepherd:manual-steps
- [ ] Vercel → shepherd-site and the shepherd.run vanity project → Settings → Build & Deployment: enable "Include files outside the Root Directory in the Build Step", so ../scripts/vercel-ignore-build.sh exists when the ignore step runs. shepherd-docs and shepherd-demo already need this for their builds. Without it those two gates fail open and keep building as today — no breakage, just no saving.
- [ ] POST-MERGE: check the four deploy logs on the merge commit — each should log a "vercel-ignore-build: … — building." line, and later unrelated commits should log "— skipping the build."
```


scoop:
## What I checked

### Fix 2 — the ignore gate wiring (verified path-by-path)

All four Vercel projects are covered and the relative depth to the shared script is right in each:

- `docs-site/vercel.json:4` → `../scripts/…` from root dir `docs-site` ✓
- `ui/vercel.json:6` → `../scripts/…` from root dir `ui` ✓
- `site/vercel.json:3` → `../scripts/…` from root dir `site` ✓
- `deploy/vanity/vercel.json:3` → `../../scripts/…` from root dir `deploy/vanity` ✓ (two levels, correct)

`find . -name vercel.json` returns exactly those four, so no project was missed. `scripts/vercel-ignore-build.sh` is committed `100755` (`git ls-files -s`).

**Gated path sets vs. what each build actually reads — verified, not assumed:**

- `docs-site`: `syncDocs()`'s `PAGES` array (`docs-site/scripts/sync-docs.mjs`, entries with `srcDir` `"docs"`, `"."` → `CLAUDE.md`, and `".claude/rules"`) reads exactly `../docs/*.md`, `../CLAUDE.md`, `../.claude/rules/*.md`; TypeDoc's program is scoped by `docs-site/typedoc.tsconfig.json` `"include": ["../src/**/*.ts"]`. The gate `./ ../src ../docs ../CLAUDE.md ../.claude/rules` covers all of them. Grepping `docs-site/src` for `../../..`, `repoRoot`, `readFileSync` returns no matches, so nothing else escapes the root.
- `ui`: `ui/vite.config.ts:19-21` (`appVersion()`) reads `new URL("../package.json", import.meta.url)` — the root manifest — and feeds `__APP_VERSION__`. The gate's `../package.json` is therefore correct and the issue's "reads nothing outside `ui/`" was wrong; the plan caught this. The only other outside reference in `ui` is `ui/scripts/gen-docs-manifest.ts:27` (`import { PAGES } from "../../docs-site/scripts/sync-docs.mjs"`), and its output is `ui/src/lib/docs-manifest.ts` (`gen-docs-manifest.ts:35`) — committed inside `ui/`, generated by the separate `gen:docs` script, not `build:demo` (`ui/package.json:10`). So `./ ../package.json` is complete.
- `site`: grep of `site/src`, `site/scripts`, `site/astro.config.mjs` for `../../`, `CHANGELOG`, `readFileSync` outside the package returns only `site/scripts/check-build-artifacts.mjs` (a `check:artifacts` script, not the build) reading its own files. `./` is complete.
- `deploy/vanity`: `deploy/vanity/` contains only `README.md` + `vercel.json`, redirect-only, no build command. `./` is complete.

**Script logic** (`scripts/vercel-ignore-build.sh`): polarity matches Vercel's documented contract (0 cancels, 1 builds); the four fail-open branches (no args, empty/unset `VERCEL_GIT_PREVIOUS_SHA`, not a work tree, previous SHA absent via `git cat-file -e "$sha^{commit}"`) all `exit 1`, and the `case` on `$?` sends anything other than a clean 0/1 — notably git's 128 — to `exit 1`. `set -uo pipefail` without `set -e` is required here so `$?` survives the `git diff`; that is correct as written. Relative pathspecs (`./`, `../src`) resolve against the cwd inside the work tree, so running from each Root Directory does the right thing.

### Fix 1 — the docs sidebar

`docs-site/astro.config.mjs:4` drops only the named `typeDocSidebarGroup` import; the plugin registration `starlightTypeDoc({…})` at line 49 is untouched, so all API pages still generate. Line 175 replaces the group with `{ label: "API reference", slug: "api/readme" }`. The generated tree is git-ignored (`docs-site/.gitignore`: `src/content/docs/api/`) and loaded through the standard `docsLoader()` collection (`docs-site/src/content.config.ts`), so a TypeDoc-emitted `api/README.md` does become the `api/readme` entry, and Starlight validates `slug:` at build time — a wrong target fails loudly rather than shipping a dead link. Removing the now-inert `sidebar:` plugin option (line 57 comment) has no effect on page generation.

### The one problem

`test/vercel-ignore-build.test.ts:68` annotates `const env: Record<string, string> = { ...GIT_ENV };`. `GIT_ENV` (lines 34-40) spreads `process.env`, whose type carries a `[key: string]: string | undefined` index signature; TypeScript preserves index signatures through object spread, so the spread type is not assignable to `Record<string, string>` (TS2322). This is in scope for the root type-check: `tsconfig.json:5` sets `strict: true` and the config has no `include`, excluding only `node_modules, ui, extension, site, docs-site` (`tsconfig.json:17`), so `test/**` is compiled — and `.github/workflows/ci.yml:103-104` runs `bun run typecheck` → `tsc --noEmit`. The repo's own idiom confirms the value type: `src/sandbox.ts:311-313` declares `env: Record<string, string | undefined> = process.env` and filters `typeof v === "string"` before building a `Record<string, string>`. The precedent test this one is modelled on avoids the issue by passing an inline literal straight to `spawnSync` (`test/check-generated-docs.test.ts:41`) rather than annotating an intermediate. Typing `env` as `NodeJS.ProcessEnv` (or `Record<string, string | undefined>`) fixes it; `spawnSync`'s `env` option accepts that directly, and `delete env.VERCEL_GIT_PREVIOUS_SHA` stays legal.

## Nits (non-blocking):

- `scripts/vercel-ignore-build.sh`: the "skipping" message goes to stdout while every fail-open message goes to stderr; one stream for all of them would read more consistently in Vercel's log.
- `scripts/vercel-ignore-build.sh`: `git rev-parse --is-inside-work-tree` is tested by exit status only — in a bare repo it exits 0 while printing `false`. Irrelevant on Vercel (non-bare clone), but `[ "$(git rev-parse --is-inside-work-tree 2>/dev/null)" = true ]` would be exact.
- `docs-site/astro.config.mjs:157-175`: the comment pins concrete measurements (80 MB, ~11s, 50 KB, 1826 pages); those will drift as `../src` grows. The ratio and the mechanism are the durable part.

- `deploy/vanity/vercel.json:3`: gating a redirect-only project with no build command saves essentially no build minutes; it is plan-sanctioned but was not in the issue's scope.
- `docs-site/astro.config.mjs:57-59`: removing the `sidebar: { label, collapsed }` option from `starlightTypeDoc()` goes slightly beyond the plan's stated "sidebar entry + import + comment". It is inert today, but restoring `typeDocSidebarGroup` later would silently come back with the plugin's default label/collapse state.

## Stated limitations (could not verify in-repo)

- No `node_modules` exists anywhere in this worktree, so I could not rebuild `docs-site` to confirm that TypeDoc with `readme: "none"` emits `src/content/docs/api/README.md` → `dist/api/readme/index.html`. I relied on the issue's local measurement and the plan; the failure mode is loud (Starlight rejects an unresolvable `slug:`), not silent.
- For the same reason I could not confirm that `starlight-typedoc`'s `sidebar` option is optional in v0.23. If it were required, config evaluation would throw — again loud, not silent.
- Vercel-side contracts are not in this repo and were taken as written: the exit-code polarity of `ignoreCommand`, the semantics of `VERCEL_GIT_PREVIOUS_SHA` (last *successful* deploy, populated when an Ignored Build Step is configured), and whether files outside a project's Root Directory are present when the ignore command runs. Note that every one of these failing resolves to "build anyway", i.e. today's behaviour, not a stale deploy.

scoop:
**Finding 1 (`test/vercel-ignore-build.test.ts:68`, `Record<string, string>`) — not reproducible; leaving as is.**

The premise is that root `tsc --noEmit` rejects `const env: Record<string, string> = { ...GIT_ENV }`. It does not. `bun run typecheck` — the literal command `.github/workflows/ci.yml` runs — exits 0 on this branch, and the file is in the program (`tsc --noEmit --listFiles | grep -c test/vercel-ignore-build.test.ts` → 1). The pre-push `tsc` lane was green for the same reason.

To rule out the file being silently skipped, I injected a deliberate error on that exact line and re-ran:

```
test/vercel-ignore-build.test.ts(68,61): error TS2322: Type 'string | undefined' is not assignable to type 'number'.
```

So line 68 is type-checked under the exact CI config, and the real assignment is clean.

The reason is repo-specific: `tsconfig.json` sets `"types": ["bun-types"]`, and bun-types redeclares `NodeJS.ProcessEnv` as `interface ProcessEnv extends Bun.Env {}` (`node_modules/bun-types/overrides.d.ts:76`), where `Bun.Env` (`bun.d.ts:503`) is `{ NODE_ENV?: string; TZ?: string }` — no `[key: string]: string | undefined` index signature. The `@types/node` signature the finding assumes is not what this program resolves.

On the suggested alternatives: `NodeJS.ProcessEnv` would be the *less* precise annotation here, and passing inline like `test/check-generated-docs.test.ts:41` does not work for this call — it needs a mutable copy so `delete env.VERCEL_GIT_PREVIOUS_SHA` can produce the unset-variable case the fail-open test covers.

No code change, so nothing to push.

Issue #2033, opened 2026-08-04
Remaining agent.name consumers on herdr 0.7.5: name-taken squatter eviction can never match


kai-osthoff:
Follow-up from #2029, which fixed every consumer whose failure LEAKED a tab. Two name-based lookups
were deliberately left alone there because each has a working fallback or sits on an error path —
recording them here rather than silently skipping them.

**1. `agent_name_taken` squatter eviction (the real one).**
`src/herdr.ts` → `startAgentWithCollisionRetry()` and `registerAgentWithCollisionRetry()`, plus the
same pair in `src/herdr-socket-driver.ts`, resolve squatters with
`(await this.listAsync()).filter((a) => a.name === name)`. On 0.7.5 `a.name` is `undefined` on every
record, so the filter is always empty: nothing is evicted, all three attempts re-collide, and the
spawn fails with `agent_name_taken` instead of self-healing.

The value IS available — probe-confirmed on a live 0.7.5 daemon by registering a throwaway agent:
`pane report-agent --agent probe-2029` surfaces the label under **`agent`**, not `name`.

```
{"agent": "probe-2029", "agent_status": "working", "pane_id": "w1:p3AW", ...}
```

Caveat that kept it out of #2029: `agent` holds the bare KIND (`"claude"`) for trusted, auto-detected
agents, so matching on it needs care — a session that sanitizes to `claude` would alias the kind. And
verifying the fix means forcing a real name collision, which no existing test does.

**2. `src/review.ts` → `findSquatter()`** matches `a.name === label` first, then falls back to the
worktree cwd — which resolves the same agent on 0.7.5. Degraded, not broken. Fix alongside (1) or
leave it.

Both should be re-checked against 0.8.0 as part of #2032.

kai-osthoff:
⏸️ Waiting on review of PR #2197 by @scoop.

kai-osthoff:
✅ PR #2197 merged.

Issue #2032, opened 2026-08-04
herdr version bumps need a record-shape gate: diff the generated protocol before raising the ceiling


kai-osthoff:
Follow-up from #2029.

Twice now a herdr record-shape change has silently disabled a husk reaper — #721, then #2029, where
FOUR reapers went dead at once because 0.7.5 stopped sending `pane.label` and `agent.name`. Both
times the code kept compiling, kept running, and kept logging nothing.

`src/generated/herdr-protocol.ts` is generated from herdr's own JSON schema and already told us:
`SuccessResponsePaneInfo.label` and `SuccessResponseAgentInfo.name` are `?: string | null`, while
`SuccessResponseTabInfo.label` is required. Nothing consumed that signal.

Proposal — make the bump checklist mechanical:

1. Regenerate `src/generated/herdr-protocol.ts` (`bun run gen:herdr-types`) against the new herdr.
2. Diff the **required/optional flags** on `SuccessResponseTabInfo`, `SuccessResponsePaneInfo` and
   `SuccessResponseAgentInfo`. Any field that moved required → optional is a stop-and-review, not a
   silent regeneration. `check:herdr-types` currently only asserts the file is up to date; it does
   not classify what changed.
3. Re-verify the reaper assumptions #2029 depends on against the new version, live:
   - does `tab.list` still return `label` as required and non-null for every tab?
   - is `agent_status` still `"unknown"` for an agentless tab?
   - does `pane.process_info` still return foreground procs for a restored/agentless pane? (An empty
     list reads as `undetermined` → fail-closed spare forever, i.e. a second way to reap nothing.)
   - are `tab_id`s still stable across a close (#569)?

**Do this before raising `HERDR_LAST_SUPPORTED_VERSION` / `HERDR_LAST_SPAWNABLE_VERSION` past 0.7.5**
(`src/herdr-capabilities.ts`). The running daemon already advertises 0.8.0, and this bug class is
exactly what a version bump introduces.

The regression fixtures added in #2029 (`test/herdr-fixture-shape.test.ts`) pin the CURRENT shape and
will fail loudly if a future capture reintroduces a field herdr does not send — that is the floor
this issue builds on, not a replacement for it.

Issue #2031, opened 2026-08-04
herdr restore re-spawns every persisted tab: bound session.json's blast radius


kai-osthoff:
Follow-up from #2029 (deferred there deliberately; the reaper fix is what stops the leak, this is
the crash-window mitigation).

`~/.config/herdr/session.json` persists every tab, and herdr's restore path re-spawns each one as a
bare shell on start. On the host in #2029 that meant **496 `pane.spawn.start` events in the first 60
seconds** after a daemon restart, ~1000 threads re-materialised before any Shepherd sweep could run.
Restarting herdr did not clear the leak; it recreated it.

With the reapers fixed (#2029), a restart now converges: the boot sweep sights husks at 5s and the
self-scheduled confirming pass closes them ~30s later. So the exposure is reduced to that window
plus any crash between leak and sweep — but the window still starts by materialising every persisted
husk at once, which is what took the daemon out with `pthread_create: EAGAIN`.

Worth exploring:

- Whether Shepherd can prune helper-labelled entries from `session.json` while herdr is down (the
  file is herdr's, so this is a compatibility bet, not a clean seam).
- Whether an aggressive first sweep — no two-sweep debounce for tabs that already existed at boot,
  before Shepherd has spawned anything — is safe enough to shorten the window.
- Raising `TasksMax` for `deploy/herdr.service`. Note the ceiling actually hit in #2029 was the
  cgroup task limit (`TasksMax=2500`, `TasksCurrent=1705` at rest), not `LimitNOFILE` — see #1857,
  which covers only the FD side.

Whatever lands here should be measured the same way #2029 was: against a live daemon, counting
`pane.spawn.start` events per minute after a restart.

PR #2034, opened 2026-08-04
fix(herdr): key every helper-tab reaper on the tab label (#2029)

### LINKED ISSUE #2029: Both helper-tab reapers are no-ops on herdr 0.7.5 — husk leak from #1852 has fully regressed
## Summary

Both helper-tab reapers are dead code paths on herdr 0.7.5. Neither can ever close a husk,
because both identify helper tabs by a field that 0.7.5 no longer supplies for exactly the panes
that need reaping. This is a silent regression of #1852 (fixed 2026-07-21 in 341b4307 / #1858),
re-broken one day later when 0.7.5 support landed (2026-07-22, 633dea48 / #1901).

On my host the leak had grown back to **321 helper husks** — larger than the 316 that triggered
#1852 in the first place — and took the herdr daemon down with a thread-spawn panic.

Environment: Shepherd 1.46.0, herdr 0.7.5 (protocol 17), Linux, systemd user units.

---

## Impact (measured on a live host)

The previous `herdr.service` lifetime died with:

```
thread '<unnamed>' panicked at library/std/src/thread/functions.rs:131:29:
failed to spawn thread: Os { code: 11, kind: WouldBlock, message: "Resource temporarily unavailable" }

herdr.service: Consumed 3h 39min 52s CPU time over 13h 59min wall clock time,
               12G memory peak, 9G memory swap peak.
```

It took Chromium down with it (two coredumps in the journal at the same timestamp — the whole
uid was out of thread headroom).

State of the replacement instance, **17 minutes** after a clean restart:

| Metric | Value |
|---|---|
| `herdr.service` `TasksCurrent` / `TasksMax` | 1705 / 2500 |
| `herdr.service` `MemoryCurrent` | 8.7 GB |
| `herdr server` threads | 1027 (515 `tokio-rt-worker` + ~505 `herdr-pty-*`) |
| `herdr server` open FDs | 1523 |
| Tabs in `herdr tab list` | 504 |
| ...of which Shepherd helper tabs | **321** |
| Live agents in `herdr agent list` | **8** |

Restarting herdr does **not** clear this. `~/.config/herdr/session.json` persists all 505 tabs,
and the restore path re-spawns them as bare shells — 496 `pane.spawn.start` events inside the
first 60 seconds after start:

```
=== pane spawns per minute after the 09:50 restart ===
    496 2026-08-04T07:50
      3 2026-08-04T07:52
```

So the leak is restart-durable: every restart immediately re-materialises ~1000 threads.

Observed accumulation rate from `herdr-server.log` (11 days, `tab.create` minus `tab.close`):
**+1 to +39 tabs/day**, net +180 over the log window.

---

## Root cause 1 — `reapOrphanTabs` filters on a field husks don't have

`src/tab-reaper.ts:190`, inside `groupHelperPanesByTab`:

```ts
if (!isShepherdHelperLabel(p.label)) continue;
```

On herdr 0.7.5, `pane.list` emits `label` **only for panes that carry a registered agent**.
A husk is by definition a tab whose agent has exited — the registration is gone, and the label
goes with it. Measured against the live daemon:

```
panes WITH a 'label' key : 35 / 504
tabs  WITH a 'label' key : 504 / 504

HELPER tabs  (by TAB label)  : 321
HELPER panes (by PANE label) : 0     <-- what reapOrphanTabs() iterates
panes belonging to those 321 helper tabs : 321,  of which labelled: 0
```

The scope filter keys off precisely the property that disappears the moment a tab becomes
reap-eligible. The loop body never executes; `toReap` is always empty. The 35 labelled panes are
the 8 live sessions (plus their siblings), which the liveness gate would correctly spare anyway.

The liveness gate itself (`paneForegroundProcs` + two-sweep debounce) is fine and was never the
problem — it simply never receives a candidate.

The premise the reaper was designed around is documented at `src/tab-reaper.ts:103`:

> **Husk signal = process liveness (ground truth), not list-absence.** Under herdr 0.7 an exited
> helper agent leaves its pane alive as an idle `zsh`, and that pane STILL appears in `agent list`
> (#721)

That is no longer true on 0.7.5. `herdr agent list` returns 8 records — the live sessions only.
Husks appear in neither `agent list` nor in labelled form in `pane list`.

## Root cause 2 — `reapTransientByLabel` matches a field that no longer exists

`src/transient-tab-reaper.ts:46`:

```ts
if (!a.name.startsWith(labelPrefix)) continue;
```

herdr 0.7.5 agent records have no `name` key at all:

```
keys: agent, agent_status, cwd, focused, foreground_cwd, pane_id, revision,
      state_change_seq, tab_id, terminal_id, terminal_title,
      terminal_title_stripped, workspace_id
```

And `src/herdr.ts:394` launders the absence into an empty string:

```ts
name: a.name ?? "",   // comment: "herdr always supplies them in practice"
```

Every agent name is therefore `""`, and `"".startsWith("__distill__")` is always `false`. The
boot reap in `index.ts` (`distiller.reapOrphans()` at :2380, `optimizer.reapOrphans()` at :2396,
`mergeSuggest.reapOrphans()` at :2413) walks the list and matches nothing, forever.

Same shape as root cause 1: a missing field silently degrades to a non-match instead of an error.

## The protocol schema already said this

`src/generated/herdr-protocol.ts` — generated from herdr's own JSON schema and checked into this
repo — declares the two fields both reapers depend on as **optional and nullable**, while the tab
label is **required**:

```ts
export interface SuccessResponsePaneInfo {
  label?: string | null;      // <-- reaper 1 depends on this
  ...
}

export interface SuccessResponseAgentInfo {
  name?: string | null;       // <-- reaper 2 depends on this
  ...
}

export interface SuccessResponseTabInfo {
  label: string;              // <-- required. Nothing reaps on this.
  ...
}
```

Shepherd's own domain types drop that optionality (`HerdrPane.label: string`,
`HerdrAgent.name: string`), and the `?? ""` in `mapAgentRecord` / `panes()` is what makes the
narrowing typecheck. Had the optionality been preserved through the mappers, the compiler would
have flagged both call sites.

## Why nobody noticed

`src/index.ts:1010`:

```ts
if (r.closed.length || r.sparedError)
  console.warn(`[tabs] reaped ${r.closed.length} husk tab(s); spared ...`);
```

With zero candidates every counter is `0` and `panesFailed` is `false`, so nothing is logged.

```
$ journalctl --user -u shepherd --since "7 days ago" | grep -c "\[tabs\]"
0
```

Seven days, zero output. A sweeper that silently does nothing is indistinguishable in the log
from one that silently has nothing to do. The `[distill]`/`[optimize]`/`[merge]` boot reapers are
equally quiet — they only log when `reaped > 0`.

## Regression timeline

| Date | Commit | Event |
|---|---|---|
| 2026-07-21 | `341b4307` (#1858) | #1852 fixed — husk reaper reworked around `pane.label` + process liveness |
| 2026-07-22 | `39a6631f` (#1887) | guard against unsupported 0.7.5 |
| 2026-07-22 | `633dea48` (#1901) | herdr 0.7.5 (protocol 17) support lands |
| 2026-07-24 | — | earliest entry in the current `herdr-server.log`; leak already accumulating |
| 2026-08-04 | — | 321 husks, 1027 threads, daemon panics on thread spawn |

The #1852 fix was invalidated the day after it shipped. Neither the 0.7.5 epic (#1889) nor the
follow-ups audited the reaper paths against the changed record shapes.

---

## Suggested fix

The husks are still perfectly visible — just not from the two surfaces the reapers read.
`tab.list` carries every label, and Shepherd already models the husk signal on the tab type:

```ts
export interface HerdrTab {
  tabId: string;
  label: string;
  /** "unknown" when no live agent backs the tab — i.e. an orphaned husk. */
  agentStatus: HerdrState;
  workspaceId: string;
}
```

Concretely:

1. **Re-key the scope filter onto tabs.** Derive the helper set from `tabsAsync()`
   (`isShepherdHelperLabel(tab.label)`), then map tab → panes via `tab_id`. Keep
   `classifyHelperTab` / `paneForegroundProcs` / the two-sweep debounce exactly as they are —
   they remain the safety gate, and the `rundown` exact-match case still relies on them.
2. **Fix `reapTransientByLabel` to match on the tab label too**, or drop it in favour of (1) now
   that the tab-driven sweep can see the same husks. If it stays, it must not match on
   `HerdrAgent.name`.
3. **Preserve schema optionality through the mappers.** Type `HerdrPane.label` and
   `HerdrAgent.name` as `string | undefined` (or `string | null`) rather than coercing with
   `?? ""`, so a field herdr stops sending becomes a compile error, not a silent no-match.
4. **Make the empty-scope case loud.** `ReapResult` should distinguish "no helper tabs in scope"
   from "helper tabs found, all spared". A sweep that finds zero helper tabs while
   `tab.list` reports hundreds of helper-labelled tabs is a bug signal and should warn.
5. **Regression test against a 0.7.5-shaped fixture** — a `pane.list` reply with `label` absent
   and an `agent.list` reply with `name` absent. The current tests presumably build panes with
   labels present, which is why this passed CI.
6. **Consider bounding `session.json`.** Even with the reaper fixed, a crash between leak and
   sweep leaves husks that the restore path faithfully re-spawns. A restore-time cap or a
   helper-label filter on restore would stop a stale session file from re-creating ~1000 threads
   before the first sweep runs.

## herdr 0.8.0 — must be re-checked before the update

The running daemon already advertises the next release:

```
INFO herdr::update: new stable build available at
  https://github.com/herdrdev/herdr/releases/download/v0.8.0/herdr-linux-x86_64
```

`HERDR_LAST_SUPPORTED_VERSION` / `HERDR_LAST_SPAWNABLE_VERSION` are both pinned at `0.7.5`
(`src/herdr-capabilities.ts:17`), so 0.8.0 is not adopted yet — but it is the version queued for
update, and this bug is precisely the class of breakage a version bump introduces.

**Whatever fix lands here must be verified against 0.8.0 before that ceiling is raised**, because
the fix depends on record-shape assumptions that already changed once between 0.7.4 and 0.7.5:

- Does `tab.list` still return `label` as a required, non-null field for every tab?
- Does `agent_status` on a tab still read `"unknown"` for an agentless tab?
- Does `pane.process_info` still return foreground procs for a restored/agentless pane, or does
  it now return empty (which the classifier reads as `undetermined` → fail-closed spare forever,
  i.e. a *second* way to silently reap nothing)?
- Is `tab_id` still stable across a close (the current code documents relying on herdr 0.7 stable
  short ids, #569)?
- Are the persisted `session.json` restore semantics unchanged — does 0.8.0 still re-spawn every
  persisted tab as a shell on startup?

Regenerating `src/generated/herdr-protocol.ts` from the 0.8.0 schema and diffing the
required/optional flags on `SuccessResponseTabInfo`, `SuccessResponsePaneInfo` and
`SuccessResponseAgentInfo` would catch a repeat of this mechanically. That diff is worth wiring
into the herdr-version-bump checklist as a standing gate, since this is now the second time a
herdr record-shape change has silently disabled a reaper (#721, then this).

---

## Reproduction / verification

Against a running Shepherd + herdr 0.7.5 with any completed helper run (plan-review, recap,
distill, …):

```bash
# 1. Helper tabs exist and are labelled at the TAB level
herdr tab list | python3 -c "
import sys,json
t=json.load(sys.stdin)['result']['tabs']
H=('__distill__','__optimize__','__merge__','__usage_probe__','review ','plan-review ','pr-critic ','recap ')
print('helper tabs:', sum(1 for x in t if x.get('label','').startswith(H) or x.get('label') in ('rundown','__usage_probe__')))"

# 2. ...but zero panes carry a helper label — what reapOrphanTabs() iterates
herdr pane list | python3 -c "
import sys,json
p=json.load(sys.stdin)['result']['panes']
print('panes total:', len(p), '| with any label:', sum(1 for x in p if x.get('label')))"

# 3. ...and agent records have no name at all — what reapTransientByLabel() matches
herdr agent list | python3 -c "
import sys,json
a=json.load(sys.stdin)['result']['agents']
print('agents:', len(a), '| with a name key:', sum(1 for x in a if 'name' in x))"

# 4. The sweeper has never reported anything
journalctl --user -u shepherd --since "7 days ago" | grep -c "\[tabs\]"
```

Expected on a correct build: (1) > 0 implies (2) > 0. Actual: (1) = 321, (2) = 0, (3) = 8 agents
with 0 name keys, (4) = 0.

## Related

- #1852 — same leak, fixed 2026-07-21; this is its regression
- #721 — the earlier instance of "herdr record shape change silently disables the husk reaper"
- #1136, #1135, #1147 — prior helper-tab leak reports
- #1857 (open) — `LimitNOFILE` headroom for `deploy/herdr.service`. Note that the limit actually
  hit here was **not** FDs but the cgroup task ceiling: `TasksMax=2500` with `TasksCurrent=1705`
  at rest, and the panic was `pthread_create` returning `EAGAIN`. Worth covering `TasksMax`
  alongside `LimitNOFILE` there — though with the reaper fixed neither ceiling should be
  approached.

kai-osthoff:
Closes #2029.

herdr 0.7.5 stopped sending `label` on helper panes and `name` on agent records **entirely**, and
every reaper that identified Shepherd's own helper agents by those fields went dead the day 0.7.5
support landed. They kept compiling, kept running on schedule, and kept logging nothing.

## What the live daemon actually says

Probed read-only against the running 0.7.5 daemon on the reporter's host, plus one throwaway tab
that was created, registered and closed again:

| Probe | Result |
| --- | --- |
| `tab list` — tabs carrying `label` | **508 / 508** (raw, unsanitized: `__merge__6b1e1782`, `plan-review …`) |
| `tab list` — helper-labelled | **325** (322 with `agent_status: "unknown"`) |
| `pane list` — panes carrying `label` | 35 / 508 — and **0 of the 325 helper panes** |
| `agent list` — records carrying `name` | **0 / 8** |
| registered `--agent` value lands under | **`agent`**, never `name` |
| `pane process-info` on a husk pane | `["bash"]` → classifies shell-only |

Two corrections to the issue, both load-bearing:

1. The 35 labelled panes are agentless **user-session** panes, not agent-backed ones. The
   conclusion is stronger than reported: the pane-label scope filter selects the empty set by
   construction, not just for husks.
2. 0.7.5 does not merely drop the agent name — the `--agent` label surfaces under **`agent`**.
   That field cannot serve as a helper surface: it holds the sanitized name for externally
   registered (sandboxed) spawns but the bare kind `"claude"` for trusted auto-detected ones,
   which is what the affected host runs. The tab label is the only reliable surface, and the
   generated schema marks `SuccessResponseTabInfo.label` **required** while both fields the
   reapers were reading are `?: string | null`.

## What changed

**Scope from `tab.list`, evidence from `pane.list`.**

- `reapOrphanTabs` derives the helper set from `tabsAsync()` and joins panes by `tab_id` purely as
  process-liveness evidence. The liveness gate, the fail-safe precedence and the two-sweep debounce
  are untouched — they were never the problem, they just never received a candidate. New guard: a
  helper tab with **no** pane in `pane.list` is `undetermined`, never shell-only. That state only
  became reachable via the tab-driven scope, where an empty group would otherwise fall through the
  classifier's loop and read as "positively shell-only".
- `reapTransientByLabel` matches tab labels and maps owned terminalIds to tab ids through `list()`,
  skipping that read entirely for an empty owned set — which is every call site, since all of them
  run from `deferredStarts` at boot. Consequence, intended: the scan now also reaches helper
  **husks**, so a restart clears prior-lifetime husks at once instead of waiting for the debounced
  hourly sweep.

**Two more dead reapers of the same class, found while verifying:**

- `doc-agent`'s `closeHuskTabs()` / `findLiveTab()` matched `a.name.startsWith("__docagent__")`.
- `usage-probe`'s `sweep()` matched `a.name !== PROBE_NAME`. This one is not caught by the type
  change below — `!==` against `string | undefined` still typechecks — and its damage was visible:
  **49** lingering `__usage_probe__` tabs, four still holding an idle `claude`. A live probe agent
  is exactly what the husk sweep deliberately spares, so this sweep is the only thing that can
  close them.

Also re-keyed: the diagnostics fleet probe's helper skip (it was counting Shepherd's own live
helpers as fleet orphans), and `matchAgents`' contended-cwd disambiguation, which read `agent.name`
and so could not re-pair two same-cwd sessions after a herdr daemon restart. Its label source is a
**lazy thunk** consulted only inside that branch, so the 1s poller issues no extra herdr call while
every session matches by terminalId (asserted in a test).

**Schema optionality preserved through the mappers.** `HerdrAgent.name` and `HerdrPane.label` are
now `string | undefined` instead of `?? ""`, so a field herdr stops sending reaches consumers as
absent. That is what forced the doc-agent and diagnostics sites to be answered rather than silently
kept.

**The sweep is audible again.** `ReapResult` reports `helperTabs` (scope size) and `tabsFailed`, and
the sweeper logs one `[tabs]` line per sweep that has anything in scope. A reaper whose scope filter
had collapsed to the empty set was indistinguishable in the journal from one with nothing to do —
seven days, zero output, while 321 husks accumulated. A genuinely empty scope stays silent: it now
reads the same `tab list` an operator would check by hand.

**The fixture was fiction.** `test/fixtures/herdr-responses/v0.7.5/agent-list-registered.json`
carried a `name` key the daemon never sends — which is why this passed CI. Corrected to the captured
shape, with new `pane list` / `tab list` captures and `test/herdr-fixture-shape.test.ts`, which
fails if the fiction returns. Every reaper test now runs against label-less panes and nameless
agents by construction.

## Verification

- **Live dry run** — the real `reapOrphanTabs` driven against the running daemon with `closeTab`
  stubbed to log only. Scope non-empty, husks identified, live helpers spared, nothing closed:

  ```
  tab.list: 51 tabs, 6 helper-labelled
  pane.list: 51 panes, 4 with a label | agent.list: 14 agents, 0 with a name
  sweep 1: helperTabs=6 sparedLive=4 sparedError=0 shellOnly=2 closed=0
  sweep 2 (confirming): WOULD CLOSE=2  → [plan-review …, recap …]
  SPARED (live): 4 × __usage_probe__   (each foreground = claude, confirmed by process-info)
  ```

  The four spared probes are precisely the class the *transient* sweep handles — see the
  usage-probe fix above.
- `bun run test` — **8317 pass, 0 fail, 386 files**.
- `bun run lint`, `tsc --noEmit`, prettier and `fallow audit` clean.

## Second commit — the fallow baseline

`.fallow-dupes-baseline.json` had gone stale: `fallow audit` reported "381 entries but matched 0
current clone groups", reproduced at `origin/main`. A baseline that matches nothing exempts nothing,
so the first PR to touch any clone-heavy file eats 15 pre-existing clone groups it did not create —
this is that PR. Recaptured with the documented command **from a worktree checked out at
`origin/main`**, so it can only grandfather clones that already exist on main; nothing here can hide
behind it. 143 entries carry over unchanged.

## Follow-ups filed (deferred deliberately, per the plan)

- #2031 — bound the `session.json` restore blast radius (herdr owns that file; the fixed boot sweep
  now converges ~35s after start, but the window still starts by re-spawning every persisted husk).
- #2032 — wire a protocol record-shape diff into the herdr version-bump checklist, and re-verify
  these assumptions against 0.8.0 **before** raising the version ceiling.
- #2033 — the remaining `agent.name` consumers: `agent_name_taken` squatter eviction can never match
  on 0.7.5 (the value is available under `agent`, probe-confirmed), and `review.ts`'s squatter
  lookup, which degrades to a working cwd fallback.


kai-osthoff (commented):
Re-review completed with read-only inspection; no blocking findings.

Verified the first prior finding: the poller now detects contention from the initial label-free match, awaits `tabsAsync()` before any session-state loop runs, then rematches using the completed label map. The state-changing `reapGone` loop only runs after that awaited result (`src/poller.ts:552`, `src/poller.ts:558`, `src/poller.ts:618`, `src/poller.ts:624`); the new regression case asserts both stale same-CWD sessions are repaired on the first tick (`test/poller.test.ts:3297`).

Verified the second prior finding: the domain fields are optional and both mappers normalize schema-permitted nulls to `undefined`, rather than passing null through (`src/herdr.ts:25`, `src/herdr.ts:54`, `src/herdr.ts:459`, `src/herdr.ts:853`). This agrees with the generated protocol declarations (`src/generated/herdr-protocol.ts:2130`, `src/generated/herdr-protocol.ts:2249`) and is covered for agent parsing on both parser paths (`test/herdr-parity.test.ts:78`).

Verified the requested reaper paths now scope helpers from required tab labels: the durable sweep reads `tabsAsync`, joins panes by `tabId`, and fail-closes when a scoped tab has no panes (`src/tab-reaper.ts:188`, `src/tab-reaper.ts:216`, `src/tab-reaper.ts:223`, `src/tab-reaper.ts:273`); the transient reaper filters tab labels and preserves owned tabs via terminal-to-tab mapping (`src/transient-tab-reaper.ts:54`, `src/transient-tab-reaper.ts:62`, `src/transient-tab-reaper.ts:66`). Captured 0.7.5-shaped fixtures explicitly omit agent names and husk-pane labels while exercising both reapers (`test/herdr-fixture-shape.test.ts:27`, `test/herdr-fixture-shape.test.ts:34`, `test/herdr-fixture-shape.test.ts:91`, `test/herdr-fixture-shape.test.ts:109`).

Per instruction, I did not run builds, lint, or tests.

PR #2040, opened 2026-08-05
feat(terminal): clean-terminal session in the main repo via right-click menus

kai-osthoff:
## What

Right-click a **session row** or a **repo chip** → **"Clean terminal in main repo"**: a bare operator shell (\`$SHELL\`) opens directly in the repo's **main checkout** — not a worktree, no agent, no prompt — visible and usable in the web UI like any session terminal. One clean terminal per repo: picking it again focuses the existing one. Exit the shell and the session auto-archives.

## Design (probe-verified against live herdr 0.7.5)

- **Pane-direct, no \`agent start\`**: herdr 0.7.5's \`agent start\` only hosts known agent CLIs, and \`agent attach\` refuses agentless panes (\`agent_not_found\`) — both verified live. So the shell comes from \`tab create --cwd <repoPath>\` (the root pane IS an interactive shell; the reply's cwd is asserted, mismatch → tab rollback), and the browser attaches through the **pane-level socket transport** (\`terminal session control\`), with the node-pty fallback disabled for this kind. Shepherd spawns no process at all; \`SESSION_MARKER_ENV\` rides \`tab create --env\`.
- **Discriminated union create**: \`TerminalCreateInput = { repoPath, terminal: true }\` — no \`prompt\`/\`baseBranch\` members exist to misuse; ordinary creates are untouched. Exact wire payload pinned by tests on both sides.
- **Atomic singleton**: a \`terminal_claims\` heartbeat lease (renew 5s / reclaim >30s stale) is taken in one DB transaction **before** any herdr call — a losing racer never spawns; \`sessions_terminal_singleton\` (partial unique index on live terminal rows) is the durable backstop. The 409 carries \`existingId\` so the UI focuses instead.
- **herdr.send invariant**: a terminal session never receives agent input — funnel guard in \`sendSteerTo\`, per-verb 409s (resume/relaunch/replace/reply/preview/interrupt/restore), \`broadcast\` skips terminals with an additive \`skipped\` count, \`retryHalted\` skips, \`haltAll\` filters before matching.
- **Lifecycle**: poller exempts terminals from agent reconcile / claude-liveness / auto-revive (a shell would otherwise read husk-stranded and get "revived"); a **fail-closed** \`panesAsync\` sweep auto-archives after two consecutive confirmed-gone sweeps; recap exempt; decommission closes the persisted \`terminalTabId\` (main repo untouched — no worktree reap); create preflight fails closed on a herdr without pane control.

## Tests

- New \`test/terminal-session.test.ts\` (15): create contract + row shape, preflight fail-closed, singleton + parked-spawn concurrency race, cwd-mismatch rollback, all verb fences, restore-with-newer-live edge, zero-\`herdr.send\` invariant across broadcast/retryHalted/haltAll, teardown, lease TTL/renew/release semantics, unique-index bypass, restart-shaped store hydration, pane-sweep live/gone/failure state machine.
- \`test/validate.test.ts\`: terminal-arm exact payload, unknown-key rejections, standard-create regression pins.
- UI: exact-payload pin via \`cleanTerminalCreateInput\` + \`canResume\`/\`canRelaunch\` fences (\`format.test.ts\`), RepoSwitcher menu roving updated for the 5th item.
- Suites: root 8356 pass / UI 4539 pass; lint, svelte-check, check:i18n (EN+DE), glossary, feature-catalog + announcement-version gates all green.

Feature-catalog entry: \`v1.47.0-clean-terminal.ts\` (EN+DE announcement).

Plan-gate: implemented per the 5-round-reviewed \`.shepherd-plan.md\` (design pivoted to pane-direct after the live herdr probe).

### DIFF
--- a/src/herdr-capabilities.ts
+++ b/src/herdr-capabilities.ts
@@ -83,3 +83,18 @@ export function herdrUsesExternalRegistrationSpawn(): boolean {
   if (!detected) return false;
   return compareSemver(detected, HERDR_EXTERNAL_REGISTRATION_VERSION) >= 0;
 }
+
+/** First herdr version whose CLI exposes pane-level `terminal session control` — the ONLY
+ *  transport an agentless clean-terminal pane can attach through (`agent attach` refuses such
+ *  panes with `agent_not_found`; probe-verified live on 0.7.5). Matches the version the
+ *  SocketPtyBridge wire fixtures were captured on. */
+const HERDR_FIRST_PANE_CONTROL_VERSION = "0.7.3";
+
+/** Whether the detected herdr can host a clean terminal (pane-level socket attach available).
+ *  Unlike the spawn gates above this fails CLOSED on an unknown version: a clean terminal
+ *  whose pane cannot be attached would be a session with no usable terminal, so the create
+ *  preflight refuses until the boot probe has recorded a capable version. */
+export function herdrPaneControlSupported(): boolean {
+  if (!detected) return false;
+  return compareSemver(detected, HERDR_FIRST_PANE_CONTROL_VERSION) >= 0;
+}
--- a/src/herdr-socket-driver.ts
+++ b/src/herdr-socket-driver.ts
@@ -118,6 +118,19 @@ export class SocketHerdrDriver implements IHerdrDriver {
     return this.cli.panes();
   }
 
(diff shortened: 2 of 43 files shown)


kai-osthoff:
🚨✨ This PR is 💯 READY ✅💚🚀 Shocking, right?! 😱 @scoop 👉 do your job 🕵️‍♂️☕⏰

Issue #2045, opened 2026-08-07
Recap verdict has the same unescapable-quote hazard the critic just fixed


scoop:
Split out of the critic fix (see the PR that introduces `VERDICT_BODY_FILE`) — deliberately **not** widened into that PR.

## The shared hazard

`RecapService` hand-authors a verdict JSON exactly like the critic did, and shares `decideVerdictAction` / `tolerantParseJson`. So it carries the same failure:

An unescaped `"` immediately before `:` or `,` is byte-identical to a real key/value boundary. `jsonrepair` cannot resolve that ambiguity — no repairer could. German prose (`„…":`, `„…",`) produces it constantly.

## It has already happened

Not theoretical. From `~/.shepherd/shepherd.log`:

```
[recap] 4bad87e2-…: verdict file present but unparseable even after jsonrepair — failing.
  spawn=8b41d83a-… model=sonnet provider=claude effort=low elapsed=314s
```

The snippet shows a complete, correct recap that was thrown away.

## Suggested fix

Mirror what the critic PR does:

1. Recap writes its prose to a sidecar markdown file; the JSON keeps only the short structured fields.
2. Reader splices the sidecar in, keeping the inline field as a fallback so in-flight runs survive a restart.

The recap runs in a tmpdir (not an untrusted PR-head checkout), so the pre-seed scrub the critic needs is **not** required here — worth confirming before implementing rather than copying it blindly.

## Also noticed, separate call

Archiving a session deletes its `reviews` row (`forget()` → `store.dropReview()`), so a verdict that passed is invisible in-app afterwards and the `reviews` table reads empty DB-wide. That is working as designed, but it makes a healthy review indistinguishable from a failed one in the Herd view. Filing here only as an observation — no change proposed.

scoop:
✅ PR #2114 merged.

PR #2046, opened 2026-08-07
fix(critic): carry the verdict body as a markdown sidecar

scoop:
Two live sessions on a German-language repo produced no review verdicts. Diagnosed both; only one was actually broken, and this fixes it.

## Root cause

The critic hand-authors `.shepherd-review.json`, so every `"` inside the long prose `body` has to be escaped by the model. It only has to miss once.

The unrecoverable shape is a bare `"` **immediately before `:` or `,`**:

```
…gibt es nicht mehr": drei Dinge, „beides".
```

That is byte-identical to a genuine key/value boundary, so `jsonrepair` closes the string there and produces a different document. This is a real grammar collision, not a gap in the repairer — no repair heuristic can resolve it. (Plain unescaped quotes followed by prose, the #822 shape, *are* recoverable; those still work.)

German hits it constantly, because `„…":` and `„…",` are the ordinary way to quote a phrase before a colon or comma.

**What it cost.** A live critic wrote a complete, well-argued 7.7 KB `request-changes` verdict — 26 unescaped `"`, 0 escaped — and all of it was discarded. Verified by recovering the original `Write` call from the reviewer's transcript: repaired only the escaping and it parses cleanly, `changes_requested`, 2 findings, 6776-byte body.

**And it cost 30 minutes.** `decideVerdictAction` holds `unparseable` on `wait` unless the spawn finished or timed out. A critic that has written its file sits idle-but-alive, so `isSpawnAlive` correctly reports alive and the run waits out the full 1800s. That run wrote its file at ~18:39 and only failed at 19:02:16 — **23 minutes waiting on a file that was already final on disk.**

## The fix

**Body travels as a sidecar.** The critic writes `.shepherd-review.md` — ordinary markdown, nothing to escape — and the JSON keeps `decision`, `summary`, `findings`. Prose that never has to be escaped cannot collide with the JSON grammar, so this closes the class rather than patching an instance.

- Inline `body` is still honoured when no sidecar exists, so older critics, a Codex critic answering via the `-o` fallback, and any run in flight across a restart keep working.
- The prompt binds the term "body" to the new file once, so the ~10 existing rules that say `"body"` keep resolving without restating them.
- Write order is load-bearing and stated: `.md` first, `.json` last, because the JSON is the completion signal.

**The sidecar is scrubbed pre-spawn, like the JSON.** This is the part that would otherwise make the change a security regression. The critic worktree is a checkout of the untrusted PR head, so a PR can commit a fixed-name `.shepherd-review.md`. It cannot fake a verdict alone — the read still needs a parseable JSON, which is scrubbed — but it *would* survive into a genuine run, be spliced in as that run's body, and be posted to the PR as the critic's review: decision genuine, words attacker-authored. `scrubStaleVerdictArtifacts` is now variadic so both PR-critic call sites scrub JSON and sidecar in one call, rather than leaving a second call site for a later change to forget.

**Fail fast on a stable unparseable verdict.** The prompt makes the JSON the critic's last action, so once its bytes stop changing there is nothing to wait for. Two consecutive identical reads → finalize (~30s instead of 1800s). Requiring *identical* bytes keeps a genuine partial write waiting, since that one is still growing.

**`raw` is now populated on the critic's unparseable read.** It was dropped on the floor, which is why `logNoVerdict`'s `snippet:` could never fire for a critic run — confirmed absent in the real failure's log line. It also backs the stability check above.

I originally planned to steer the idle critic to rewrite its own file. Dropped it: `ReviewService`'s herdr dep has no `send`, and adding pane-steering there would bypass the `#serializeSteer` serialization that exists to prevent steer races. Too much surface for the benefit, especially once the sidecar makes the case rare.

## Verification

- `bun run lint` clean; `bunx tsc --noEmit` clean.
- `bun run test` — 8348 pass, 0 fail.
- New tests confirmed **red** before the fix (temporarily disabled the guard and watched the fail-fast test fail).
- End-to-end against the **real recovered 6776-byte payload** in the new format: parses with `repaired=false`, body byte-identical, and the exact construct that killed it survives verbatim.
- `bunx fallow audit --base origin/main --fail-on-issues` — no issues in the 7 changed files.

## Deliberately not in this PR

- `RecapService` shares `decideVerdictAction` and has the identical hazard — with a real occurrence already in the logs. Filed as #2045.
- Archiving deletes the session's `reviews` row, so a *passed* verdict becomes invisible in-app and the table reads empty DB-wide. Working as designed; noted in #2045 as an observation, no change proposed.

The second session in the original report turned out **not** to be broken: its critic finalized normally and posted its verdict to the PR three seconds before merge. It only looked identical to the failure because archiving had already dropped the row.

### DIFF
--- a/src/codex-last-message.ts
+++ b/src/codex-last-message.ts
@@ -103,10 +103,18 @@ export function readRoleResultText(
  * race must not block the spawn). Harmless for base-checkout reviewers (the plan reviewer detaches at
  * the trusted base) — a defense-in-depth no-op there.
  */
-export function scrubStaleVerdictArtifacts(worktreePath: string, resultFile: string): void {
-  try {
-    rmSync(join(worktreePath, resultFile), { force: true });
-  } catch {
-    /* best-effort — a pre-seed we can't remove still fails closed via the read path's validators */
+export function scrubStaleVerdictArtifacts(worktreePath: string, ...resultFiles: string[]): void {
+  // Variadic so a role with more than one fixed-name artifact scrubs them in ONE call: the critic
+  // now also writes a sidecar body (`VERDICT_BODY_FILE`), and a second call site is exactly the kind
+  // of thing a later change forgets to add. A pre-seeded sidecar cannot fake a verdict on its own
+  // (the read still requires a parseable verdict JSON), but it WOULD be spliced in as the body of a
+  // genuine run and posted to the PR as the critic's review — attacker-authored prose under
+  // Shepherd's name. Same fixed name, same untrusted checkout, same scrub.
+  for (const f of resultFiles) {
+    try {
+      rmSync(join(worktreePath, f), { force: true });
+    } catch {
+      /* best-effort — a pre-seed we can't remove still fails closed via the read path's validators */
+    }
   }
 }
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -6,6 +6,8 @@
  * streak/notes/publish control flow, which stays there.
  */
 import { readRoleResultText, codexLastMessageFile } from "./codex-last-message";
+import { existsSync, readFileSync } from "node:fs";
+import { join } from "node:path";
 import { execFile } from "node:child_process";
 import { promisify } from "node:util";
 import { execFileSync, timedAsync } from "./instrument";
@@ -713,10 +715,21 @@ function scopeAndOutputTail(
     '- At most 5 NEW findings per review, highest-severity first. This is a PRIORITISATION instruction, never a limit that may lose information: if more than 5 genuine blocking problems exist, emit ALL of them and say so in "summary" (the change needs reworking, not tweaking). NEVER move a blocking problem into `Nits (non-blocking):`, and never drop one, to fit the budget.',
     "- Re-raised prior findings do NOT count against those 5.",
     "",
-    "When done, write your verdict as JSON to the file `.shepherd-review.json` in the repository root, with EXACTLY this shape:",
-    '{"decision": "request-changes" | "comment", "summary": "<=100 char one-liner", "body": "<full markdown review>", "findings": ["<discrete actionable item>", ...]}',
+    "When done, write TWO files in the repository root:",
+    // Binds the term once: every rule above says `"body"` (sections, citations, stated limitations),
+    // and they all now resolve to this file. Cheaper and less drift-prone than restating ~10 rules.
+    '1. `.shepherd-review.md` — your full markdown review. Everywhere these instructions say "body" — the named sections, the `path:line` citations, the stated limitations — they mean THIS file. It is NOT JSON: write the prose directly, escape nothing, quote however you like.',
+    "2. `.shepherd-review.json` — the structured verdict, with EXACTLY this shape:",
+    '{"decision": "request-changes" | "comment", "summary": "<=100 char one-liner", "findings": ["<discrete actionable item>", ...]}',
+    // #2042: the body used to live inside this JSON. A single unescaped `"` before a `:` or `,` —
+    // which German `„…":` prose produces constantly — is indistinguishable from a real key/value
+    // boundary, so it silently destroyed complete verdicts. Keeping the prose out of JSON entirely
+    // is the only fix that cannot regress; the note below keeps the remaining escaping honest.
+    'Escaping matters ONLY in the JSON file, and it is short: inside "summary" and each "findings" entry every `"` must be written `\\"`. If that feels error-prone, phrase them without quotation marks — the markdown file is where quoting is free.',
     'The "findings" array lists every discrete change the author must make — one entry per point, routed per FINDINGS ROUTING above. Use [] when everything you found is non-blocking (report those in the body sections named above) or there is genuinely nothing to address; "request-changes" requires at least one finding.',
-    'Use "request-changes" ONLY for blocking problems (does not satisfy the task, logic bug, security hole). Otherwise use "comment". Never approve. Write the file as your final action, then stop.',
+    // Ordering is load-bearing: the server finalizes as soon as the JSON parses, so the JSON must be
+    // written LAST — otherwise a tick landing between the two writes finalizes with an empty body.
+    'Use "request-changes" ONLY for blocking problems (does not satisfy the task, logic bug, security hole). Otherwise use "comment". Never approve. Write `.shepherd-review.md` FIRST and `.shepherd-review.json` LAST — the JSON file is the completion signal — then stop.',
   ];
 }
 
@@ -724,6 +737,22 @@ function scopeAndOutputTail(
  *  sites can scrub a pre-seeded copy from the untrusted checkout before launch (see
  *  scrubStaleVerdictArtifacts). */
 export const VERDICT_FILE = ".shepherd-review.json";
+
+/** The critic's verdict BODY, written beside {@link VERDICT_FILE} as plain markdown.
+ *
+ *  #2042: the body is the one field that is long free prose, and the critic hand-authors its JSON —
+ *  so every `"` in it has to be escaped by the model. It only has to miss once. The unrecoverable
+ *  shape is a bare `"` immediately before `:` or `,` (`…nicht mehr": drei Dinge`), which is
+ *  byte-identical to a real key/value boundary — jsonrepair cannot resolve that ambiguity, and no
+ *  repairer could. German prose produces it constantly via `„…":` / `„…",`.
+ *
+ *  Carrying the body as its own file removes the escaping obligation entirely: markdown has no
+ *  reserved characters, so no prose can break the read. The JSON that remains (decision, summary,
+ *  findings) is short and structured, where the model's escaping is reliable in practice.
+ *
+ *  MUST be scrubbed before launch exactly like VERDICT_FILE — same fixed name, same untrusted
+ *  PR-head checkout, so the same pre-seed attack applies (see scrubStaleVerdictArtifacts). */
+export const VERDICT_BODY_FILE = ".shepherd-review.md";
 
 export interface RawVerdict {
   decision?: unknown;
@@ -873,9 +902,31 @@ export function defaultReadVerdict(
   );
   if (text === null) return { status: "absent" };
   const r = tolerantParseJson(text);
-  return r.status === "ok"
-    ? { status: "parsed", value: r.value as RawVerdict, repaired: r.repaired }
-    : { status: "unparseable" };
+  // Carry the raw bytes (as the recap read already does): they drive the stable-unparseable fail-fast
+  // in ReviewService.tick, and they fill the `snippet:` in the no-verdict diagnostic — which until
+  // now could never fire for a critic run, because this branch dropped `raw` on the floor.
+  if (r.status !== "ok") return { status: "unparseable", raw: text };
+  const value = r.value as RawVerdict;
+  // #2042: the body travels as a sidecar markdown file. Read it from the SAME worktree and splice it
+  // in. A missing sidecar is not an error — an inline `body` stays valid, which keeps every older
+  // critic, a Codex critic answering through the `-o` fallback, and any run already in flight across
+  // a server restart working unchanged. Sidecar wins when both exist: it is the format the current
+  // prompt asks for, and it is the one that cannot have been mangled by escaping.
+  const body = readVerdictBody(worktreePath);
+  if (body !== null) value.body = body;
+  return { status: "parsed", value, repaired: r.repaired };
+}
+
+/** Read the sidecar body written beside the verdict JSON. null when absent/unreadable (mid-write) —
+ *  never throws, so a sidecar problem can only cost the body, never the whole verdict. */
+function readVerdictBody(worktreePath: string): string | null {
+  const p = join(worktreePath, VERDICT_BODY_FILE);
+  if (!existsSync(p)) return null;
+  try {
+    return readFileSync(p, "utf8");
+  } catch {
+    return null;
+  }
 }
 
 export function normalizeDecision(d: unknown): ReviewDecision | null {
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -719,13 +719,19 @@ function scopeAndOutputTail(
     // Binds the term once: every rule above says `"body"` (sections, citations, stated limitations),
     // and they all now resolve to this file. Cheaper and less drift-prone than restating ~10 rules.
     '1. `.shepherd-review.md` — your full markdown review. Everywhere these instructions say "body" — the named sections, the `path:line` citations, the stated limitations — they mean THIS file. It is NOT JSON: write the prose directly, escape nothing, quote however you like.',
-    "2. `.shepherd-review.json` — the structured verdict, with EXACTLY this shape:",
-    '{"decision": "request-changes" | "comment", "summary": "<=100 char one-liner", "findings": ["<discrete actionable item>", ...]}',
+    "2. `.shepherd-review.json` — the structured verdict, with this shape:",
+    '{"decision": "request-changes" | "comment", "summary": "<=100 char one-liner", "findings": ["<discrete actionable item>", ...], "body": "<optional — see below>"}',
     // #2042: the body used to live inside this JSON. A single unescaped `"` before a `:` or `,` —
     // which German `„…":` prose produces constantly — is indistinguishable from a real key/value
     // boundary, so it silently destroyed complete verdicts. Keeping the prose out of JSON entirely
     // is the only fix that cannot regress; the note below keeps the remaining escaping honest.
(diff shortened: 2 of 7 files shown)


scoop:
## What I verified

**The core mechanism works end to end.** `VERDICT_BODY_FILE = .shepherd-review.md` (`src/critic-core.ts:755`) is spliced into the parsed verdict at `src/critic-core.ts:916` via `readVerdictBody` (`src/critic-core.ts:922-930`), and `buildVerdictCore` then reads it at `src/critic-core.ts:1124` (`typeof raw.body === 'string' ? raw.body : ''`), which is what `src/review.ts:1333` and `src/standalone-critic.ts:664` post. The splice happens before that coercion, so the sidecar reaches the posted review — confirmed by following the value through those four sites.

**Both critic services get the sidecar for free.** `StandalonePrCriticService` defaults its reader to the same `defaultReadVerdict` (`src/standalone-critic.ts:153`), so the standalone path picks up the sidecar with no further wiring.

**Scrub coverage is complete.** The variadic `scrubStaleVerdictArtifacts` (`src/codex-last-message.ts:106-120`) is called with both names at the only two untrusted-checkout spawn sites, `src/review.ts:1507` and `src/standalone-critic.ts:541`. The third caller, `src/plan-gate.ts:1060`, still passes one argument and remains valid under the variadic signature — I grepped all three call sites; there are no others. The security reasoning in the new comment is right: a pre-seeded `.shepherd-review.md` cannot fabricate a verdict (the JSON is the completion signal — asserted by the new test at `test/critic-core.test.ts:118-124`), but it *would* be spliced into a genuine run's body, so it needs the same scrub.

**The critic can actually write the second file.** The `reviewer` preset grants bare `Write` (`src/transient-agent-argv.ts:170`), not a path-scoped Write, so a second output file is permitted under `--permission-mode dontAsk`.

**The new artifact is ignored.** `.shepherd-*` at `.gitignore:26` already covers `.shepherd-review.md`, so the sidecar does not dirty a reviewed repo's status the way an untracked file would.

**The stable-unparseable fail-fast is wired correctly.** `raw` is now carried on the unparseable branch (`src/critic-core.ts:904`), which matches the optional field already declared on the shared type (`src/json-tolerant.ts:59`) and which `logNoVerdict` already read for its `snippet:` (`src/review.ts:1045-1047`) — that snippet genuinely could never fire for a critic before this change. The new branch in `resolveWait` (`src/review.ts:1086-1096`) sits after `detectStuck`, keeps the finalize claim held on the terminal action, and releases it (`f.finalizing = false`, `src/review.ts:1097`) on the first sighting, so the claim/release discipline the surrounding method depends on is preserved. The resulting badge code is `no-verdict-unparseable` (`src/review.ts:144-148`), which is a pre-existing `ReviewSummaryCode` — no new catalog key is introduced by this diff, so there is no locale-parity obligation to check.

**Imports are clean.** `node:fs` and `node:path` are imported once each in `src/critic-core.ts:9-10`; there were no prior imports from either module in that file to duplicate.

## Finding

### `src/critic-core.ts`: the new JSON shape drops `body`, which silently guts the Codex `-o` chat-answer fallback

The replacement output contract (`src/critic-core.ts:722-723`) specifies EXACTLY `{decision, summary, findings}` — `body` is gone from the shape the critic is told to emit.

That is fine for a critic that writes files. It is not fine for the fallback this repo deliberately built. `src/review.ts:833` sets `captureLastMessage: true`, so a Codex critic spawns with `-o` pointing at the per-spawn last-message file, and `defaultReadVerdict` reads it when the result file is absent (`src/critic-core.ts:898-902`). The entire documented purpose of that path (`src/codex-last-message.ts:1-14`) is the case where Codex *answers in chat and never calls a write tool*. In exactly that case there is no `.shepherd-review.json` and, by the same token, no `.shepherd-review.md` — so `readVerdictBody` returns null, nothing is spliced, and the chat JSON now contains no `body` because the prompt told it not to. `buildVerdictCore` coerces to `''` (`src/critic-core.ts:1124`) and `src/review.ts:1333` posts a review whose body is nothing but `CRITIC_REVIEW_MARKER`. Before this diff the same chat answer carried the full review inline and survived intact. Note also that `tolerantParseJson` (`src/json-tolerant.ts:37-47`) does no fenced-block extraction, so a chat answer that emits the markdown *and* the JSON side by side is unparseable rather than partially recoverable — there is no accidental rescue here.

The in-diff comment at `src/critic-core.ts:910-913` asserts the opposite of what the code now does: it claims a missing sidecar `keeps every older critic, a Codex critic answering through the -o fallback, and any run already in flight ... working unchanged`. The first and third clauses are true (those critics ran under the old prompt and do emit an inline `body`). The middle clause is false for any Codex critic running the NEW prompt, which is every one of them after this merges — the prompt it is given no longer has a `body` field for it to fall back to.

Suggested fix: keep `body` in the documented JSON shape as an OPTIONAL field, with a one-line note that it is only for the case where the critic cannot write files, and that the sidecar wins whenever both exist (which the read path already implements at `src/critic-core.ts:915-916`). That reintroduces no escaping hazard on the normal path — a file-writing critic never populates it — while restoring the fallback the repo built on purpose.

## Related risk (not raised as a finding)

The read path treats `parsed JSON + absent sidecar + no inline body` as a complete verdict and finalizes it immediately on a strict parse. Today that shape is ambiguous between an old-style critic (fine) and a new-style critic that wrote the two files in the wrong order (loses the whole body). The prompt states the ordering twice (`src/critic-core.ts:732`) and the poll is 15s (`src/standalone-critic.ts:586`), so the window is narrow and it depends on the model disobeying an explicit instruction — I am not treating it as reachable-today. Worth knowing it exists if the fallback fix above changes the shape's meaning.

Second, on the unparseable path the run now finalizes as an error and reaps the worktree (`src/review.ts:1090`, then `finalize`/`dropInflight` at `src/review.ts:993-995`) while a complete, perfectly readable `.shepherd-review.md` is sitting on disk beside the broken JSON. The sidecar exists to protect the prose, yet the prose is still discarded whenever the small JSON fails. Salvaging it (or at least logging its presence) would close the loop — an enhancement, not a defect.

## Nits (non-blocking):
- `test/codex-last-message.test.ts:122`: `REVIEW_BODY` hardcodes `.shepherd-review.md` instead of importing `VERDICT_BODY_FILE`, so a rename of the constant would leave the test green against a stale literal.
- `src/critic-core.ts:721-723`: the two filenames are literals in the prompt rather than interpolations of `VERDICT_FILE`/`VERDICT_BODY_FILE`; this matches the pre-existing style of the line it replaced, so it is only a missed opportunity to bind them.
- `src/review.ts:193`: `lastUnparseableRaw` is never cleared when a later read comes back parsed or absent; harmless today (both of those outcomes finalize or are transient), but the field's lifetime is wider than its comment implies.

- `src/review.ts:1086-1096`: the stable-unparseable fail-fast is a second, independent fix riding along with the sidecar change. It is well-motivated by the same incident (the 23-of-30-minutes wait) and well-tested, but it is a behaviour change to the finalize gate that the sidecar fix does not require.

- `src/plan-gate.ts:282` and `src/plan-gate.ts:1501`: the plan reviewer still carries its free-prose `body` inside `PLAN_VERDICT_FILE`, so it remains exposed to the exact quote-before-colon failure this PR eliminates for the critic.
- `src/standalone-critic.ts:597-607`: the standalone critic's finalize loop still waits out the full hard timeout on an `unparseable` read; it gets no equivalent of the new stable-bytes fail-fast. (Pre-existing behaviour, explicitly documented there, and that critic only ever posts comments.)

## Stated limitations
- I did not run the test suite (read-only review; CI covers it). In particular I did not execute `test/critic-core.test.ts:63-73`, so the claim that `jsonrepair` cannot recover a bare `"` immediately before `:` or `,` is reasoned from the repairer's end-of-string lookahead heuristic rather than observed here. It is consistent with the incident described, and it is exactly the kind of assertion the new test exists to pin down.
- The task names TWO sessions with no verdict; the diff's evidence (comments and tests) cites one concrete run, TASK-2125. I cannot see session data from this repo, so I cannot confirm the second session shares this root cause — I am taking the diff's diagnosis as written for that one.

scoop (commented):
## Verdict

The prior round's single finding is fully addressed, and I found no new blocking defect. Decision: `comment`.

## Prior finding — CONFIRMED FIXED

> src/critic-core.ts: the new verdict JSON shape at lines 722-723 removes `body` … Keep `body` as an OPTIONAL field in the documented shape … and correct the comment at src/critic-core.ts:910-913.

Both halves land:

- **Shape.** `src/critic-core.ts:723` now reads `{"decision": …, "summary": …, "findings": […], "body": "<optional — see below>"}` — `body` is back. `src/critic-core.ts:722` was simultaneously softened from *with EXACTLY this shape* to *with this shape*, which is necessary: an optional field contradicts an EXACTLY-shaped contract. Verified the wording change against `git diff 3eb70740...HEAD` (the fixup commit touches exactly these two lines plus the routing rule).
- **Routing rule.** `src/critic-core.ts:733` tells the critic to OMIT `body` when it wrote the sidecar, and to include it ONLY when it cannot write files at all, with the parenthetical *you are answering in chat rather than editing a worktree*. That parenthetical matches the actual Codex failure mode documented at `src/codex-last-message.ts:6-9` (Codex produces a complete verdict as its final chat message and never calls a write tool), so the carve-out reaches the case it is meant to reach.
- **Comment.** The old text at `src/critic-core.ts:910-913` claimed the `-o` fallback works unchanged. It is replaced by `src/critic-core.ts:920-923`, which now states the opposite explicitly: the `-o` fallback is the ONLY way a Codex critic can carry a body, and dropping `body` from the prompt would post a review with no text.
- **Path still works end to end.** The `-o` fallback read is at `src/critic-core.ts:904-908`; the coercion that would have blanked the body is at `src/critic-core.ts:1136` (`body: raw && typeof raw.body === "string" ? raw.body : ""`); the post is at `src/review.ts:1333`. With `body` back in the shape, a chat-only Codex verdict carries prose through all three. Covered by the new test at `test/critic-core.test.ts:322-331` (prompt half) and `test/critic-core.test.ts:105-118` (reader half).

## What I verified on the rest of the diff

- **Identifier resolution.** `VERDICT_BODY_FILE` is defined at `src/critic-core.ts:761` and consumed at `src/critic-core.ts:935`, `src/review.ts:45,1507`, `src/standalone-critic.ts:43,541`. The literal `.shepherd-review.md` in the prompt (`src/critic-core.ts:721,733,738`) is byte-identical to the constant's value at `src/critic-core.ts:761`, and to the name the reader joins at `src/critic-core.ts:935` — no two-forms drift. New imports `existsSync`/`readFileSync`/`join` are the only fs/path imports in the file (`src/critic-core.ts:9-10`), so no duplicate-binding collision.
- **Every scrub site.** `scrubStaleVerdictArtifacts` has exactly three call sites: `src/review.ts:1507` and `src/standalone-critic.ts:541` (both now pass `VERDICT_FILE, VERDICT_BODY_FILE`) and `src/plan-gate.ts:1060`, which passes only `PLAN_VERDICT_FILE` — correct, because the plan reviewer writes no sidecar (its contract is at `src/plan-gate.ts:282-284`). The variadic signature at `src/codex-last-message.ts:106` keeps all three compiling.
- **The scrub is genuinely needed, and sufficient.** The splice at `src/critic-core.ts:927-928` unconditionally overwrites `value.body` with the sidecar, so an attacker-committed `.shepherd-review.md` in an untrusted PR head would ride into a genuine run's posted review (`src/review.ts:1333`). It is scrubbed pre-spawn at `src/review.ts:1507` and `src/standalone-critic.ts:541`, and `rmSync(..., {force: true})` at `src/codex-last-message.ts:115` unlinks a symlink rather than following it. The comment at `src/codex-last-message.ts:107-112` states this hazard accurately. The sidecar cannot fake a verdict on its own — `readRoleResultText` (`src/codex-last-message.ts:72-89`) keys entirely off `VERDICT_FILE` / the per-spawn `-o` name, asserted by the test at `test/critic-core.test.ts:134-140`.
- **Sidecar file is not committable by accident.** `.shepherd-*` is in `.gitignore:29`, so `.shepherd-review.md` is ignored in this repo like the JSON already was.
- **The critic can actually write it.** `src/transient-agent-argv.ts:170` gives the `reviewer` preset a bare `Write` (not a path-scoped one), so a second output file needs no allowlist change under `--permission-mode dontAsk` (`src/transient-agent-argv.ts:263`).
- **Stable-unparseable fail-fast.** `src/review.ts:1094-1097` is reachable only when `decideVerdictAction` returned `wait`, which for `unparseable` means the spawn is neither finished nor timed out (`src/json-tolerant.ts:165-167`). The claim/release protocol matches the wedged-pane branch above it: the terminal return keeps `f.finalizing` held and lets `tick` finalize (`src/review.ts:982-996`), while the first-sighting path falls through to `f.finalizing = false` at `src/review.ts:1098`. `read.raw` is now populated for the critic by `src/critic-core.ts:914`, which also fixes the `snippet:` at `src/review.ts:1046-1048` that could previously never fire. Two identical reads are at least one tick apart, and the tick is 15s (`src/index.ts:1794,1819`). `noVerdictCause` maps this to `no-verdict-unparseable` (`src/review.ts:144-148`), matching the test at `test/review.test.ts:1153-1173`.
- **Log safety of the newly-populated `raw`.** `sanitizeRecapFailureDetail` (`src/recap.ts:207-230`) redacts token/secret shapes and truncates at 300 chars, so surfacing verdict bytes in `console.warn` cannot dump a 7.7 KB review or a credential.
- **Repaired-parse interaction.** jsonrepair closes a truncated write, so a mid-write file surfaces as `parsed`+`repaired` (held until the spawn finishes, `src/json-tolerant.ts:162-163`) rather than `unparseable`. That is why the new fail-fast mostly sees genuinely-broken documents rather than partial ones — the growing-bytes guard at `test/review.test.ts:1176-1199` covers the remaining case.

## Cannot verify

- The incident details cited in the comments (TASK-2125 losing 23 of 30 minutes; the 7.7 KB verdict) are not reproducible from this repo. Note the 30-minute figure implies a `SHEPHERD_REVIEW_TIMEOUT_MS` override — the in-repo default is 10m (`src/review.ts:369`, `src/config.ts:823`) — which is a supported env knob, not an inconsistency. Assumed as written.

## Nits (non-blocking):

- src/codex-last-message.ts:106 — making the parameter fully variadic means `scrubStaleVerdictArtifacts(worktreePath)` now type-checks and silently scrubs nothing on a security-critical path; `(worktreePath: string, first: string, ...rest: string[])` would keep the old arity requirement while still supporting N files.
- src/codex-last-message.ts:93-105 — the JSDoc above the now-variadic function still speaks of *a pre-existing verdict RESULT file* and `<resultFile>` in the singular; the plural is only explained in the inline comment below it.
- src/critic-core.ts:928 — `if (body !== null)` treats a zero-length or whitespace-only sidecar as present and would overwrite an inline `body` with the empty string; a truthiness or `trim()` guard is cheap insurance in a change whose whole subject is empty review bodies.
- src/critic-core.ts:721,733,738 — the prompt hardcodes `.shepherd-review.md` / `.shepherd-review.json` rather than interpolating `VERDICT_BODY_FILE` / `VERDICT_FILE`; this continues the pre-existing style for the JSON, but it now means two literals must stay in sync with two constants.
- src/standalone-critic.ts:595 — `StandalonePrCriticService` picks up the sidecar read and the scrub but not the stable-unparseable fail-fast, so a standalone critic that writes broken JSON still burns the full deadline. Not a regression (that is today's behaviour), just an asymmetry the session critic no longer has.

- src/critic-core.ts:738 — the md-first/json-last ordering is enforced only by prompt text; there is no server-side guard, so a critic that inverts the order leaves a window in which a tick parses the JSON, finds no sidecar and no inline `body`, and finalizes with an empty review — precisely the symptom this PR exists to remove. Holding one tick when the JSON parses with neither a `body` field nor a sidecar would close it without affecting the Codex chat path (which always has an inline `body`).
- src/review.ts:1094-1097 — the fail-fast is not gated on `STARTUP_GRACE_MS`, so it can fire roughly two ticks (~15-30s) into a run. A critic that writes a broken JSON early and would have rewritten it is reaped instead. Exposure is small today because jsonrepair absorbs truncated writes and the identical-bytes requirement filters growing files, but the guard that protects the sibling `absent` case (`graceElapsed`, `src/json-tolerant.ts:168`) has no counterpart here.

- src/plan-gate.ts:282-284 — the plan-gate reviewer still carries its full markdown `body` inside the verdict JSON, so it remains fully exposed to the unrepairable quote-before-colon collision this PR removes for the PR critic.
- src/plan-gate.ts:286-291 — that same prompt explicitly instructs the reviewer to write `body` in German, which is the exact input class identified here as producing the collision constantly.
- src/recap.ts:177 and src/rundown-core.ts:692 — the recap and rundown roles use the same body-inside-JSON contract and share the hazard, though their bodies are shorter.

PR #2050, opened 2026-08-08
fix(preview): see dev servers agents start in the scratchpad

scoop:
Agents prototyping in their Claude scratchpad start dev servers there rather than in the worktree. Shepherd could not see them: no Preview badge, no Stop, and the server survived session archive unnoticed.

## Why the hint file was not the problem

Preview attribution was **cwd-only** — `scanListeningPortsByWorktree` keeps a process only if its `/proc/<pid>/cwd` is under a session worktree. A server started with `cd <scratchpad>` matches nothing, so its ports never enter the set. Relocating (or additionally reading) `.shepherd-preview` in the scratchpad would have changed nothing on its own: `resolveDevPort` honors a declared port **only when that port is already listed as listening**.

## The lever

Every process an agent spawns inherits `SHEPHERD_SESSION_ID` in its environ — cwd-independent, fixed at exec, and already the runaway reaper's provenance key. Attributing by that marker works no matter how Claude Code lays out its scratchpad, so a CLI change cannot break it.

It is kept narrow:

- the cwd must first be under `agentTmpDir()` — a plain string test on data `scanProcs` already read;
- only then is the environ consulted, and only for a process that **actually listens**, so the mmap-lock read costs nothing on a normal sweep;
- `scratchRoot === null` (`SHEPHERD_AGENT_TMPDIR=""`) disables it entirely — the existing one-env-var rollback still works.

## Three blind spots, one predicate

The same cwd filter gated four call sites. `containedBySession()` now covers all of them:

| Site | Before | After |
| --- | --- | --- |
| `scanListeningPortsByWorktree` | port never enters the set | attributed by marker |
| `resolveDevPort` | hint ignored (port not in set) | worktree hint first, then the server's own dir |
| `candidatesOn` + `verifiesLive` | Stop reports `killed: 0` as **success**; the idle ladder walks SIGTERM→SIGKILL→gaveUp having signalled nothing | proposed and re-proved by the same rule, actually signalled |
| `scanWorktreeProcs` | never offered at "Terminate & close" — leaks past archive | offered as a leftover |

Both halves of the stop path had to learn it together: teaching only `candidatesOn` would turn every stop of such a server into `"refused"`, teaching only `verifiesLive` would signal on a weaker rule than proposed it.

## Also

- **`EXCLUDED_PORTS` = 9222 / 9229 / 9230.** A CDP endpoint and the Node inspector answer a plain HTTP GET, so the liveness probe alone would let one win the non-curated fallback. Widening eligibility to the scratchpad — which is full of both — made that worth closing. Not even an explicit hint can now point the preview at a debugger.
- **`scratchRoot` is passed in, not imported.** `tmp-sweep.ts` already imports `process-reaper.ts`, so reading `agentTmpDir()` there would close a cycle. It doubles as the test seam.
- **macOS is untouched.** `makeDarwinProbes` supplies `portsForPid`, so the preview scan already works there and still does; darwin omits `environForPid` (deliberately, to keep the #1144 runaway reaper fail-closed), so the fallback never fires and `hintDirs` stays empty. The darwin tests assert exactly that.
- The `shepherd-preview` skill and the resident preview notice now state the same rule: start the server from the worktree; if you prototype from the scratchpad it is still found, and the hint file then belongs next to the server.

## Verification

`bun run lint`, `bunx tsc --noEmit`, `bun run test` (8375 pass / 0 fail), `bunx fallow audit --base origin/main --fail-on-issues` — all clean.

Beyond the unit tests, the chain was exercised against **real `/proc`**: a real HTTP server spawned with cwd in the scratchpad and the session marker set was attributed to the worktree, contributed its directory as a hint location, had a hint next to it honored, was offered as a leftover, was signalled by `stopListenersOnPort` (`{signalled: 1, outcome: "ok"}`) with the process confirmed dead and the port closed — and a second session scanning the same host did **not** see it.

### DIFF
--- a/agent-skills/.claude/skills/shepherd-preview/SKILL.md
+++ b/agent-skills/.claude/skills/shepherd-preview/SKILL.md
@@ -8,23 +8,36 @@ description: How to expose a long-running local dev/test server from a Shepherd 
 Shepherd can proxy a dev server running in this worktree to the operator's browser (and, when
 enabled, onto the tailnet). It finds the server by port.
 
+## Where to start the server
+
+Start it **from this worktree**. That is the case the preview is built around, and it keeps the
+server next to the code you are changing.
+
+Prototyping from your scratchpad works too — Shepherd recognises a server you started there by the
+session marker every process you spawn inherits, so it is no longer invisible. It is still the
+second-best option: nothing in the scratchpad ends up in the PR.
+
 ## Pointing the preview at a specific port
 
-If you start a long-running dev server in this worktree and want Shepherd's live preview to target
-a specific port, write that port — a bare number, nothing else — to a file named
-`.shepherd-preview` in the repository root.
+If you start a long-running dev server and want Shepherd's live preview to target a specific port,
+write that port — a bare number, nothing else — to a file named `.shepherd-preview` in the
+repository root. If you started the server from the scratchpad instead, put that file in the
+directory you started it from; the repository root still wins if both exist.
 
 Shepherd uses that value only when the port is actually listening; otherwise it auto-detects the
 port. So this is optional: skip it if you have no dev server, or if the default detection already
 targets the right port.
+
+Debugger endpoints are never chosen as the preview, so a Node inspector or a CDP port on 9222 /
+9229 / 9230 cannot be picked up by mistake.
 
 ## What Shepherd does and does not do
 
 - Shepherd **never starts or stops your dev server**. Starting it is on you; keep it in a managed
   or background terminal, not as a blocking foreground command.
 - Shepherd owns tailnet exposure — do not run `tailscale serve` yourself.
-- The hint only applies to **isolated** sessions, which have their own worktree. A session sharing
-  the main repo directory has no dedicated worktree for the file to live in.
+- Preview only applies to **isolated** sessions, which have their own worktree. A session sharing
+  the main repo directory gets no preview at all.
 
 ## If no preview appears
 
--- a/src/index.ts
+++ b/src/index.ts
@@ -249,7 +249,10 @@ const store = new SessionStore(config.dbPath);
 // refreshes the cell these consumers then read. A single instance here is for
 // clarity, not correctness; what WOULD carry an independent, always-cold cell is a
 // reaper constructed with its own `makeDarwinProbes()` (as the tests do).
-const reaper = new ProcessReaper();
+// `agentTmpDir()` arms the scratchpad provenance fallback: a dev server an agent started
+// from its scratchpad instead of the worktree is still this session's, and must be
+// stoppable and offerable as a leftover.
+const reaper = new ProcessReaper(undefined, agentTmpDir());
 // a repo root chosen in the UI (persisted) overrides the env var / default — but
 // only if it still sits within the immutable ceiling; a stale/escaping value is
 // ignored so the active root can never climb above the ceiling across restarts.
--- a/src/poller.ts
+++ b/src/poller.ts
@@ -31,8 +31,14 @@ import { statSync } from "node:fs";
 import { classifyHalt, assistantSideText } from "./usage-halt";
 import type { UsageLimits } from "./usage-limits";
 import { maintenance } from "./maintenance";
-import { scanListeningPortsByWorktree, scanClaudeAliveByWorktree } from "./process-reaper";
+import {
+  scanListeningPortsByWorktree,
+  scanClaudeAliveByWorktree,
+  type ListenerScanTarget,
+  type WorktreeListeners,
+} from "./process-reaper";
 import { resolveDevPort } from "./preview";
+import { agentTmpDir } from "./tmp-sweep";
 import { config } from "./config";
 import { SessionLiveness, type LivenessOutcome, type TranscriptSignals } from "./session-liveness";
 
@@ -90,15 +96,16 @@ export interface PreviewWiring {
    *  union filter, so the liveness sweep — which has no wiring of its own — is
    *  covered too. Defaults to `deps.reaper`-backed refresh in index.ts. */
   refresh?: (opts?: { force?: boolean }) => Promise<void>;
-  /** Batched /proc scan: builds the inode→port map ONCE and resolves all worktrees.
+  /** Batched /proc scan: builds the inode→port map ONCE and resolves all sessions.
    *  Defaults to the real `scanListeningPortsByWorktree`. Returns `null` when the
    *  snapshot backend cannot support a negative verdict (darwin, stale/none cell) —
    *  the sweep must then leave bound listeners untouched, not tear them down. */
-  scan: (worktrees: string[]) => Map<string, number[]> | null;
+  scan: (targets: ListenerScanTarget[]) => Map<string, WorktreeListeners> | null;
   /** Pick the primary dev port from a set of listening ports for a given worktree.
    *  Defaults to `resolveDevPort`, which honors the agent-declared `.shepherd-preview`
-   *  hint (if listening + HTTP-live) and otherwise falls back to the primary-port heuristic. */
-  pick: (ports: number[], worktreePath: string) => Promise<number | null>;
+   *  hint (if listening + HTTP-live) and otherwise falls back to the primary-port heuristic.
+   *  `hintDirs` is worktree-root-first; see `resolveDevPort`. */
+  pick: (ports: number[], hintDirs: string[]) => Promise<number | null>;
   /** Opt-in idle-stop. idleMs > 0 enables it; `stop` signals a session's dev-server
    *  process (wired to SessionService.stopPreview in index.ts). Returns the stop
    *  outcome. The two that signalled NOTHING must NOT advance the escalation ladder:
@@ -519,8 +526,10 @@ export class StatusPoller {
       },
       sweepMs: preview?.sweepMs ?? config.previewSweepMs,
       refresh: preview?.refresh,
-      scan: preview?.scan ?? ((worktrees) => scanListeningPortsByWorktree(worktrees)),
-      pick: preview?.pick ?? ((ports, worktreePath) => resolveDevPort(ports, worktreePath)),
+      scan:
+        preview?.scan ??
+        ((targets) => scanListeningPortsByWorktree(targets, undefined, agentTmpDir())),
+      pick: preview?.pick ?? ((ports, hintDirs) => resolveDevPort(ports, hintDirs)),
       idleStop: preview?.idleStop,
     };
     this.livenessWiring = {
@@ -1881,6 +1890,18 @@ export class StatusPoller {
       });
   }
 
+  /** This session's dev port from its scan result. The worktree root leads `hintDirs`:
+   *  an agent-declared hint there outranks one found next to a scratchpad-rooted server. */
+  private pickDevPort(
+    s: Session,
+    listeners: WorktreeListeners | undefined,
+  ): Promise<number | null> {
+    return this.previewWiring.pick(listeners?.ports ?? [], [
+      s.worktreePath,
+      ...(listeners?.hintDirs ?? []),
+    ]);
+  }
+
   /**
    * Async core of the preview sweep. Builds the /proc map ONCE, picks a primary
    * port per session, then calls converge() on the full active set. The PreviewService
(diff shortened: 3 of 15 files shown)


scoop (commented):
# Review — `fix(preview): see dev servers agents start in the scratchpad`

Diff reviewed: `98d6170e3bc9317fbded3e3dd3969ca43e8e98df...HEAD` (15 files, 9 source + 6 test).

**Decision: comment.** I found no blocking correctness or security defect. The provenance fallback
is wired consistently through all four containment sites, every changed signature's callers agree,
and the platform fail-closed behaviour holds. Two design observations and a latent-ordering note
are below, all non-blocking.

---

## What I verified

### 1. The new predicate reaches every site that decides containment

`containedBySession()` is defined once at `src/process-reaper.ts:409-419` (cwd fast path →
scratch-root string gate → `SESSION_MARKER_ENV` lookup) and is called from exactly three
single-session sites:

- leftover sweep — `src/process-reaper.ts:841` (in `scanWorktreeProcs`, `src/process-reaper.ts:817`)
- stop proposal — `src/process-reaper.ts:987` (in `candidatesOn`)
- stop live re-proof — `src/process-reaper.ts:1007` (in `verifiesLive`)

I compared the rewritten guards against the originals in the diff hunks: the old
`if (!isUnder(p.cwd, root) || AGENT_COMMS.has(p.comm)) continue;` was split into an
`AGENT_COMMS` check followed by `containedBySession`, which is semantically identical for the
cwd-matching case. The `candidatesOn`/`verifiesLive` pair uses the *same* helper with the *same*
arguments, so the "propose-then-refuse" divergence the plan warns about cannot occur — confirmed
against `src/process-reaper.ts:939` (proposal) and `src/process-reaper.ts:951` (re-proof), both
threading the same `sessionId`.

The many-session scan does not use the helper but is built from the same two primitives
(`isScratchRooted` at `src/process-reaper.ts:393`, `sessionMarkerFor` at
`src/process-reaper.ts:386`) inside `attributeListener` (`src/process-reaper.ts:1476-1494`).

### 2. Every changed signature's callers agree

| Changed contract | Callers checked |
| --- | --- |
| `ProcessReaper.detect({id, …})` (`src/process-reaper.ts:800`) | `src/service.ts:5637`, `:5683`, `:5781` — all pass a whole `Session`, which carries `id`. The deps type is `Pick<ProcessReaper, "detect"｜"reap"｜"stopListenersOnPort">` at `src/service.ts:240`, so it tracks automatically. `src/readiness.ts:615` is a different `g.detect(ctx)` (readiness gate), not this one. |
| `stopListenersOnPort(…, sessionId?)` (`src/process-reaper.ts:920-925`) | Sole production caller `src/service.ts:4883-4889` passes `(s.worktreePath, devPort, signal, s.id)` — positional order matches. |
| `scanListeningPortsByWorktree(targets, probes, scratchRoot)` (`src/process-reaper.ts:1529-1533`) | `src/poller.ts:531`, `src/preview-launch.ts:184`, plus the darwin/reaper tests. No un-migrated call site remains — grep for `Map<string, number[]>` across `src/` and `test/` returns only `src/poller.ts:353` (`hookTicks`, unrelated) and `test/poller-preview.test.ts:97` (a test-local `scanResult` fixture, deliberately still port-keyed and adapted at `test/poller-preview.test.ts:113-121`). |
| `findPreviewDevPort(worktreePath, sessionId, deps?)` (`src/preview-launch.ts:187-191`) | `src/server.ts:2717` assigns it to `AppDeps.previewLauncher.findDevPort`, whose type was updated at `src/server.ts:375`; the one call is `src/server.ts:2739` passing `(s.worktreePath, s.id)`. The five test fakes at `test/server-preview-start.test.ts:195,241,290,348,397` take zero params, which stays assignable. |
| `resolveDevPort(candidates, hintDirs, …)` (`src/preview.ts:341-346`) | `src/poller.ts:532` and `src/preview-launch.ts:199`, both building `[worktreePath, ...hintDirs]`. |
| `new ProcessReaper(probes, scratchRoot)` (`src/process-reaper.ts:794-797`) | `src/index.ts:255` (armed), `src/preview-launch.ts:181` (refresh-only), `src/worktree.ts:67` (only used for `reapOrphansUnder` at `src/worktree.ts:542`, a cwd-under-worktree sweep this change does not touch). The new param defaults to `null`, so the two unarmed sites keep today's behaviour. |

`agentTmpDir` is already imported in `src/index.ts:121` (from `./tmp-sweep`), and the new imports
in `src/poller.ts:79` / `src/preview-launch.ts:177` introduce no cycle: `src/tmp-sweep.ts:13`
imports only `WORKTREE_MARKER, isUnder` from `process-reaper` and nothing from `poller` or
`preview-launch`.

### 3. Fail-closed paths hold

- **darwin**: the probe object `makeDarwinProbes` returns (`src/proc-probes-darwin.ts:640-652`)
  supplies `normalizeRoot`, `canAuthorizeSignal`, `signalWindowOpen`, `liveProcForPid`,
  `snapshotState`, `refreshAttemptedRecently`, `refresh` — and **no `environForPid`**
  (declared optional at `src/process-reaper.ts:106`). `sessionMarkerFor` therefore returns `null`
  (`src/process-reaper.ts:387-388`), so `containedBySession` and `attributeListener` never take the
  fallback branch. The `hintDirs: []` assertions added at
  `test/proc-probes-darwin.integration.test.ts:74` and `test/proc-probes-darwin.macos.test.ts:113`
  match that.
- I specifically checked that the new predicate cannot turn an existing darwin stop into
  `"refused"`: in `verifiesLive` the first clause is still `isUnder(live.cwd, root)`
  (`src/process-reaper.ts:409`), and the fallback clause short-circuits to `false` for want of
  `environForPid`. Identical verdict to the pre-diff `if (!isUnder(live.cwd, root)) return false;`.
- **redirect disabled**: `agentTmpDir()` returns `null` for `SHEPHERD_AGENT_TMPDIR=""`
  (`src/tmp-sweep.ts:82-86`), and `isScratchRooted` returns `false` on `null`
  (`src/process-reaper.ts:394`). `attributeListener` then returns `null` for any non-worktree proc
  (`src/process-reaper.ts:1482`), so the whole fallback is off.
- The `ctx.scratchRoot!` non-null assertion at `src/process-reaper.ts:1493` is sound: it is only
  reached when `owned === null`, which by line 1482 implies `isScratchRooted` returned `true`,
  which implies `scratchRoot !== null`.
- `readEnviron` (`src/process-reaper.ts:588-600`) swallows EACCES/ESRCH and returns `null`, so an
  environ read on a vanished or foreign-uid process cannot throw out of the sweep.

### 4. Ordering / cost claims in the comments are accurate

`attributeListener` really does defer the environ read behind the listening test
(`src/process-reaper.ts:1484-1490`), matching its own comment and the assertion at
`test/preview.test.ts` ("a non-listening scratchpad proc is never environ-read", `environReads === 0`).

### 5. Skill / notice parity

`agent-skills/.claude/skills/shepherd-preview/SKILL.md:20-25` and `PREVIEW_HINT_NOTICE` at
`src/service.ts:824-832` now state the same rule (worktree first; scratchpad hint file goes next to
where the server was started; repository root wins). These are the two channels — the skill reaches
Claude spawns via `--add-dir` (`src/agent-skills.ts:25,45-49`) and the notice is the Codex/fallback
resident block, gated to isolated sessions per `src/service.ts:814-817`. No i18n catalogue is
involved: both are fixed-English agent prompt text, explicitly documented as such at
`src/service.ts:821-822`, and grep confirms no `ui/messages` key is added or referenced by this diff.

The reworded bullet at `SKILL.md:39-40` ("Preview only applies to **isolated** sessions … A session
sharing the main repo directory gets no preview at all") is effectively true: the sweep filters on
`s.isolated && s.worktreePath` at `src/poller.ts:1872` and `converge` releases anything not in the
active set (`src/poller.ts:1878,1955`), so a non-isolated binding made through
`handlePreviewStart` → `bindExistingPreviewServer` (`src/server.ts:2733-2749`, which does not gate on
`isolated`) is torn down on the next sweep. Pre-existing shape, not changed here.

---

## Design observations (non-blocking, but worth a decision)

**Worktree-rooted ports and scratchpad-rooted ports are merged into one flat set.**
`scanListeningPortsByWorktree` unions both provenances into a single `ports` array
(`src/process-reaper.ts:1568-1577`) and `pickPrimaryPort` then ranks purely by `CURATED_PORTS` list
order (`src/preview.ts:387-389`). So a session that has a *real* worktree dev server on 3000 and a
leftover scratchpad prototype on 5173 will have its preview flip to the prototype — a case that
worked before this PR. The hint file mitigates it only when one exists. The information needed to
prefer the worktree bucket is already available inside `attributeListener` (it knows whether `owned`
was non-null); keeping the two sets separate and only falling back to the scratchpad set when the
worktree set yields no port would preserve the old behaviour exactly while still fixing the
invisible-server case. The approved plan lists this under "Risks … Accepted", so I am not blocking
on it — but the accepted mitigation (curated priority) is precisely the mechanism that produces the
flip, so the risk note does not actually cover this shape.

---

- `src/process-reaper.ts:987` — the `SESSION_MARKER_ENV` read now happens *before* `openSignalBracket` opens the pid-recycle bracket, and the bracket only re-reads cwd/comm (`src/process-reaper.ts:402-408`), never environ; this deviates from the invariant stated at `src/process-reaper.ts:358-361` ("every read the decision rests on happens strictly BETWEEN two matching starttime reads"), though it is unexploitable today because the scratchpad cwd itself encodes the owning session, so any recycled pid with a different marker would also fail the bracket's cwd byte-equality check.

## Nits (non-blocking):

- `src/process-reaper.ts:432-442` — `hintDirsFor`'s 4-entry ancestor walk climbs one level too far for the common case: a server started directly in `<agentTmpDir>/claude-$uid/<slug>/<cc-session>/scratchpad` (the layout `src/tmp-sweep.ts:99-102` derives) yields `[scratchpad, <cc-session>, <slug>, claude-$uid]`, and `claude-$uid` is shared by every session on the host — a `.shepherd-preview` dropped there would be consulted for unrelated sessions (bounded to selecting among *their own* listening ports, so no cross-session exposure, but still an unintended shared channel); stopping the walk at the `scratchpad` component or capping at 2–3 would close it.
- `agent-skills/.claude/skills/shepherd-preview/SKILL.md:16-18` — "Prototyping from your scratchpad works too … so it is no longer invisible" is unconditional, but it is false on macOS (no `environForPid`, `src/proc-probes-darwin.ts:640-652`) and false when `SHEPHERD_AGENT_TMPDIR=""`; a half-sentence caveat would stop a macOS agent from prototyping into a preview that never appears.
- `src/process-reaper.ts:394` — `isScratchRooted` compares against the raw `scratchRoot`, while worktree roots go through `normRoot`'s trailing-slash strip (`src/process-reaper.ts:454-457`); an operator setting `SHEPHERD_AGENT_TMPDIR=/some/path/` (trailing slash — harmless everywhere else, since every other consumer feeds it to `join`) silently disables the whole fallback.
- `src/preview.ts:230` vs `src/service.ts:824` — the `EXCLUDED_PORTS` rule is documented to agents in `SKILL.md:31-32` but not in the Codex-facing `PREVIEW_HINT_NOTICE`; minor asymmetry between the two channels.
- `src/process-reaper.ts:1577` — `hintDirs` is materialised from a `Set` accumulated across *all* of a session's scratchpad listeners, so with two such listeners the "most specific first" ordering the `resolveDevPort` doc relies on (`src/preview.ts:335-337`) holds only per-listener, not globally.

## Stated limitations

- I did not run tests, lint, or `tsc` (CI owns those, per the review brief). Type-compatibility claims above are from reading the declarations and every call site, not from a compiler run.
- The live behaviours in the plan's step 8 (an actual scratchpad server binding, stopping, and being offered as a leftover on a real host) cannot be verified by reading; I checked the code paths and the unit fixtures that stand in for them (`test/process-reaper.test.ts` scratchpad block, `test/preview.test.ts` scan-attribution block).

- `README.md:493-497` still describes `.shepherd-preview` as living only in "the repo/worktree root" and does not mention the scratchpad location; the file is not in this diff and the repo syncs docs in separate commits (e.g. parent `98d6170e`).
- `src/server.ts:2836` `handlePreviewStart` does not gate on `s.isolated`, so a non-isolated session can transiently bind a preview that the next poller sweep immediately releases.

PR #2051, opened 2026-08-08
feat(agent-cli): support mise-managed claude/codex install + update

scoop:
Two host-mutating agent-CLI surfaces assumed a native install. On a mise-managed host both pushed a **second**, diverging copy instead of touching the one the operator actually runs.

Live evidence from the host this was written on: `~/.local/bin/claude` → native **2.1.226** while `mise ls` tracks **2.1.222**; `~/.local/bin/codex` is a hand-rolled `mise exec node@latest -- npx @openai/codex` wrapper reporting 0.147.0 against mise's 0.146.0.

## Installing (`src/remediations.ts`)

The claude/codex diagnostics Fix is now one self-selecting shell string: when `mise which <tool>` resolves, it symlinks the mise binary into `~/.local/bin` instead of running `curl … | bash`.

It **symlinks rather than installs** because when `mise which` succeeds the binary already exists — it is merely unreachable: mise exposes tools through a shims dir absent from `shepherd.service`'s pinned PATH, and `mise activate` only runs in interactive shells. `~/.local/bin` *is* on that PATH, so one symlink clears the check with no change to how anything resolves binaries. Same trick `NODE_INSTALL` already uses for fnm-managed node.

**mise-managed is the only diverting condition** — never a "mise exists, so prefer it" heuristic. On hosts without mise, and for tools mise does not manage, the emitted command is the vendor installer verbatim, so `deploy/provision.ts` and the onboarding harness are unaffected.

## Updating (`src/codex-update.ts`)

`CodexUpdateChannel` gains `mise`. When mise owns the codex on PATH, the order is `mise upgrade codex` and **nothing else**.

No fallback, deliberately: `mise upgrade` respects the version request in the operator's mise config, so a **pinned** codex will not advance — and falling through to `codex update` / npm would install a second codex shadowing the mise-managed one, which is the defect being fixed. A pinned run reports the existing non-converged result plus a log line naming `mise use -g codex@latest`. On non-mise hosts every existing behaviour (memo order, markers, fallback, log block) is byte-identical.

Ownership is tested two ways because both host shapes exist: the on-PATH path lies under `$MISE_DATA_DIR`, or it resolves to the same real file as `mise which codex` (the `~/.local/bin` symlink the install fix creates). Realpath alone is not enough — a mise shim is a symlink to the `mise` **binary**, not into the install tree.

One shared `parseCodexUpdateChannel()` now narrows the memo read and the converged-marker parse, so the widened union cannot outgrow a hard-coded membership test.

## Testing

Executed, not string-matched — which branch a host takes is the whole feature.

- `test/remediations.test.ts`: a hermetic sandbox (stub `curl`/`mise`, PATH excluding the system bin dirs so a dev host's real mise can't flip the verdict) proves the mise branch links and never curls, and that both the absent and installed-but-unmanaged cases still take the vendor installer.
- `test/codex-update.test.ts`: the existing fake-installer harness gains mise-shaped hosts (shim, `~/.local/bin` symlink, unmanaged). Asserts `mise upgrade` runs alone, that a pinned tool runs neither other channel, and that a stale `mise` memo on a non-mise host degrades to the historical order.

`bun run lint`, `bun run test` (8366 pass), `bunx tsc --noEmit` and the fallow delta audit are green.

## Not in scope

No claude updater — Shepherd has none (Claude Code self-updates), so a mise-managed claude still can't self-update and its auto-updater can keep planting a native copy. That is how this host diverged; tracked in #2052. No binary-resolution/PATH changes either; the install fix's symlink is the only reachability mechanism added.

### DIFF
--- a/docs-site/src/content/docs/operating.md
+++ b/docs-site/src/content/docs/operating.md
@@ -109,6 +109,28 @@ the package-manager store reclaim, which doesn't depend on process data. Idle-st
 behaves the same way; see `SHEPHERD_PREVIEW_IDLE_STOP_MS` in
 [Configuration](/reference/configuration/).
 
+## Agent CLIs managed by mise
+
+If `claude` or `codex` is managed by [mise](https://mise.jdx.dev), Shepherd works
+_with_ that install instead of planting a second one beside it.
+
+**Installing.** When a `claude`/`codex` check is red and `mise which <tool>` resolves,
+the **Fix** button symlinks the mise binary into `~/.local/bin` rather than running the
+vendor installer. mise exposes tools through a shims dir that isn't on the systemd
+unit's `PATH`, and `mise activate` only runs in interactive shells — so the binary
+existed, it was just unreachable. `~/.local/bin` **is** on the unit's `PATH`. Hosts
+without mise, and tools mise doesn't manage, get the vendor installer exactly as before.
+
+**Updating.** When mise owns the `codex` on `PATH` — a shim, an install path, or a
+symlink resolving to one — the codex update runs `mise upgrade codex` and **nothing
+else**. `mise upgrade` respects the version request in your mise config, so a pinned
+codex won't advance; Shepherd then reports the update as not converged and points you at
+`mise use -g codex@latest` rather than falling back to `codex update` or `npm install -g`,
+either of which would install a second codex shadowing the mise-managed one.
+
+Claude Code has no Shepherd-side updater (it self-updates), so a mise-managed `claude` is
+upgraded by you: `mise upgrade claude`.
+
 ## Host tuning — tmpfs inodes
 
 Shepherd keeps spawned agents' Node compile cache **off** the `/tmp` tmpfs and runs
--- a/src/codex-update.ts
+++ b/src/codex-update.ts
@@ -103,10 +103,21 @@ function sanitizeVersion(v: string | null | undefined): string {
   return clean || "unknown";
 }
 
-/** The two installers that can update the codex on PATH. Which one actually
- *  WORKS is a property of the host, not something either side can know up front
- *  — see {@link buildUpdateScript}. */
-export type CodexUpdateChannel = "npm" | "codex";
+/** The installers that can update the codex on PATH. Which one actually WORKS is
+ *  a property of the host, not something either side can know up front — see
+ *  {@link buildUpdateScript}. `mise` is not a peer of the other two: it applies
+ *  only when mise OWNS the on-PATH codex, and then it is the only channel. */
+const CODEX_UPDATE_CHANNELS = ["npm", "codex", "mise"] as const;
+export type CodexUpdateChannel = (typeof CODEX_UPDATE_CHANNELS)[number];
+
+/** Narrow an untrusted value (the persisted memo, a parsed log marker) to a
+ *  channel. ONE validator so the union cannot grow past a hard-coded membership
+ *  test left behind in the store read or the marker parse. */
+export function parseCodexUpdateChannel(value: unknown): CodexUpdateChannel | null {
+  return (CODEX_UPDATE_CHANNELS as readonly unknown[]).includes(value)
+    ? (value as CodexUpdateChannel)
+    : null;
+}
 
 /** Marker the script emits once the on-PATH version has ADVANCED, naming the
  *  channel that did it. This is the ONLY attribution of a winning channel:
@@ -127,7 +138,8 @@ export const CONVERGED_MARKER = `${CODEX_UPDATE_LOG_PREFIX} converged via channe
  *   put (issue #1560).
  * - **`npm install -g @openai/codex`** — updates the npm-global copy.
  *
- * Neither is universally right. `codex update` SELF-SELECTS its channel, and that
+ * …unless **mise owns the codex we run**, which changes the shape entirely (see the mise section
+ * below). Neither is universally right. `codex update` SELF-SELECTS its channel, and that
  * choice can miss the install actually on PATH: on a host whose on-PATH `codex` is
  * a wrapper resolving to the npm-global copy, `codex update` may run `bun install
  * -g`, succeed, exit 0 — and change nothing the operator runs. Success therefore
@@ -141,6 +153,19 @@ export const CONVERGED_MARKER = `${CODEX_UPDATE_LOG_PREFIX} converged via channe
  * that changed channels), which rewrites the memo. With no memo yet, the order is
  * `codex update` then npm — the historical order, so a fresh host is never worse
  * off than before.
+ *
+ * **mise-managed codex.** If the codex actually on PATH is one mise hands out, BOTH channels above
+ * are wrong: `codex update` cannot repoint a mise install, and npm would install a second codex that
+ * shadows — or is shadowed by — the mise-managed one. So a mise-owned codex gets `mise upgrade
+ * codex` as the ONLY channel, with no fallback. `mise upgrade` deliberately respects the version
+ * request in the operator's mise config, so a PINNED codex does not advance and the run reports
+ * non-converged (naming `mise use -g codex@latest` as the operator's escape hatch) rather than
+ * shadow-installing past the pin. Ownership is tested two ways because both host shapes exist: the
+ * on-PATH path lies under `$MISE_DATA_DIR` (a shim, or a direct install path), or it resolves to the
+ * same real file as `mise which codex` (the `~/.local/bin` symlink our own install remediation
+ * creates). Realpath alone is not enough — a mise shim is a symlink to the `mise` BINARY, not into
+ * the install tree. When mise does not own it, everything below is unchanged, and a stale `mise`
+ * memo simply falls through to the default order like any non-runnable channel.
  *
  * Safety rails baked into the script:
  * - **Probe first.** `codex update` is only ever invoked if `codex --help`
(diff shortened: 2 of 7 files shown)


scoop (commented):
# Review — mise-managed `claude` / `codex`

Scope: `git diff 98d6170e3bc9317fbded3e3dd3969ca43e8e98df...HEAD` (7 files).

## What I verified

**Identifiers / wiring**

- `parseCodexUpdateChannel` is exported at `src/codex-update.ts:116` and imported at `src/index.ts:67`, used at `src/index.ts:2753`. The store key `CODEX_UPDATE_CHANNEL_KEY` is defined at `src/index.ts:2744` and consumed only by that read plus `writeChannel` at `src/index.ts:2754` — no other consumer of the persisted value exists (grep for `CODEX_UPDATE_CHANNEL_KEY` / `CodexUpdateChannel` returns only `src/codex-update.ts`, `src/index.ts`, `test/codex-update.test.ts`). Widening the union to include `"mise"` therefore cannot leak into a hard-coded membership test left elsewhere.
- The marker parse now routes through the same validator (`src/codex-update.ts:901`), and `buildUpdateScript`'s `pref` sanitiser too (`src/codex-update.ts:206`). Both previously hard-coded `"npm" | "codex"`; both now agree with the union declared at `src/codex-update.ts:109-111`. No third hard-coded membership test survives.
- `agentCliInstall` (`src/remediations.ts:115`) feeds all four hint keys (`src/remediations.ts:171-174`). Consumers still resolve: `autoFixCommandFor` (`src/diagnostics.ts:1481`, `src/diagnostics.ts:1542`), `deploy/provision.ts:120` via `selectPrereqCommand`, and `remediationsFor` (`ci/onboarding-harness/run.ts:20`). `GUIDANCE_ONLY` (`src/remediations.ts:189-192`) is untouched, so the four hints stay one-click fixable, as the diff's own test asserts.
- Shell compatibility of the emitted remediation: the in-app Fix spawns `sh -c` (`src/diagnostics.ts:257`, `defaultRunRemediation`) and provision spawns `bash -c` (`deploy/provision.ts:380`). The `if … then … else … fi` shape and the `M="$(…)"`-as-condition idiom are POSIX; the assignment's exit status is the command substitution's, so a failing `mise which` correctly falls to the native branch, and `[ -n "$M" ]` covers an exit-0-with-empty-output mise.
- The update script runs under `bash -lc` (`src/script-child.ts:15`, via `src/codex-update.ts:497`) — a **login** shell, which matters below.
- Docs manifest parity: the generator builds `keywords` as description + H2/H3 headings, lowercased and whitespace-collapsed (`ui/scripts/gen-docs-manifest.ts:58-84`). `docs-site/src/content/docs/operating.md` headings are, in order, lines 6/20/50/64/86/**112**/134/169/189/203; the new `## Agent CLIs managed by mise` at `operating.md:112` sits between “preview detection” and “host tuning — tmpfs inodes”, which is exactly where the regenerated string at `ui/src/lib/docs-manifest.ts:36` places it. No drift.

**Host ground truth (the motivating machine)**

The two findings below hinge on mise's on-disk layout, so I checked it rather than assuming:

- `~/.local/share/mise/installs/codex/0.146.0/bin/codex` — the only `codex` binary under the mise tree; the path is **version-scoped**.
- `~/.local/share/mise/installs/node/{25.9.0,26.5.0,26.5.1,26.6.0,26.7.0}/bin/node` — five versions retained side by side, i.e. mise does **not** remove the previous install when a tool moves forward.
- `~/.local/share/mise/shims/**` — **empty / absent** on this host. The shim shape the diff's comments lean on (`src/codex-update.ts:164`, `src/codex-update.ts:228`) does not exist here; tools are exposed by `mise activate`'s PATH injection, which points at those same version-scoped `installs/<tool>/<version>/bin` directories.

Stated limitation: I could not *execute* `mise which codex` (read-only review), so the claim “`mise which` prints the version-scoped path” rests on the verified layout above (there is no version-stable per-tool path in the tree to print) plus mise's documented output shape. Everything else below follows from the layout alone.

---

## Findings

### 1. `src/remediations.ts:115-119` — the symlink pins Shepherd to the version present at Fix time

`ln -sf "$M" "$HOME/.local/bin/<tool>"` stores whatever `mise which <tool>` printed. On the verified layout that is `…/installs/codex/0.146.0/bin/codex` — a path that names one specific version. The link is not a live view of “the mise-managed codex”; it is a snapshot.

Because mise keeps old versions (five node versions coexist on this host), the failure is **silent** rather than a dangling link: after any `mise upgrade <tool>`, `~/.local/bin/<tool>` still resolves to a perfectly working *old* binary, the diagnostics check stays green, and nothing tells the operator Shepherd is behind.

That directly breaks the instruction this PR adds at `docs-site/src/content/docs/operating.md:132` — “a mise-managed `claude` is upgraded by you: `mise upgrade claude`”. Following it upgrades mise's claude and leaves Shepherd running the pre-upgrade one, which is the *same* “two copies diverge silently” state the change exists to eliminate (`src/remediations.ts:95-98` states that motivation explicitly).

The fix needs a version-stable target: the mise shim (`$MISE_DATA/shims/<tool>`, after `mise reshim` — note this host has no shims dir, so it cannot be assumed to exist), or a two-line wrapper in `~/.local/bin/<tool>` that defers resolution to run time (`exec mise exec <tool> -- <tool> "$@"`) — which is, notably, the shape the PR description says the live host already hand-rolled for codex. `NODE_INSTALL` (`src/remediations.ts:20-24`) is not a precedent for the snapshot approach: fnm's node is not something Shepherd then tries to upgrade in place through fnm.

### 2. `src/codex-update.ts:232-242, 292` — a successful `mise upgrade codex` reads as “did not converge”, and the run after it silently falls back to the shadow install

Two compounding problems, both on a path that executes today:

**(a) The winner test cannot see a mise upgrade.** `after` is read with `"$CX" --version` (`src/codex-update.ts:207`, evaluated at `:290`) and compared at `src/codex-update.ts:292`. On the two host shapes this feature actually targets, `$CX` resolves through a **version-scoped** path that was fixed before the upgrade ran:
  - the `~/.local/bin/codex` symlink finding 1 creates — still pointing at the old install dir;
  - a `mise activate` login shell (`bash -lc`, `src/script-child.ts:15`) whose `PATH` contains `installs/codex/<old-version>/bin`, frozen for the lifetime of the script.

  `hash -r` (`src/codex-update.ts:288`) clears bash's command hash, not `PATH` and not a symlink target, so it does not help here. Result: `after == before`, `winner` stays empty, and the run reports non-convergence for an upgrade that in fact succeeded — then prints the pinned-version explanation at `src/codex-update.ts:302-306` (“a pinned codex will not advance… `mise use -g codex@latest`”), which is simply the wrong diagnosis. Only the shim shape converges, and this host has no shims dir.

**(b) Ownership is lost on the very next run, re-enabling the fallback this branch exists to forbid.** With the symlink shape, after mise moves to a new version:
  - the prefix test `case "$cxpath" in "$MISE_DATA"/*` (`src/codex-update.ts:237`) never matched anyway — `~/.local/bin` is outside `$MISE_DATA`;
  - the realpath test (`src/codex-update.ts:238-240`) now compares the *old* install path (`readlink -f` of the stale symlink) against the *new* `mise which codex` output — they differ.

  So `mise_owns=0`, and the order chain at `src/codex-update.ts:250-255` falls through to `codex update` → `npm install -g @openai/codex` on a host whose codex mise owns. That is precisely the “second codex shadowing the mise-managed one” the comment at `src/codex-update.ts:249` and the log line at `src/codex-update.ts:306` promise will never happen.

Fixing finding 1 with a version-stable link does not by itself fix this: a wrapper script's `readlink -f` is the wrapper's own path, which will not equal `mise which codex` either. The ownership probe needs a test that survives the tool moving versions — e.g. compare `readlink -f "$cxpath"` against the mise data-dir prefix (i.e. apply the `$MISE_DATA` prefix test to the *resolved* path, not only the on-PATH one), and/or treat “`~/.local/bin/<tool>` is the wrapper/link we installed” as an explicit signal.

Note the current tests cannot catch either half: `test/codex-update.test.ts:161` models `mise which` as returning a version-*less* `installs/codex/latest/bin/codex`, and `test/remediations.test.ts:398` stubs it as `installs/$2/bin/$2`. Both make the path stable across an upgrade by construction, so the stale-target case is never exercised. A regression test would need the stub to return a different version path after `upgrade`.

---

## Nits (non-blocking):

- `src/codex-update.ts:302-306`: the pinned-version explanation prints whenever a mise-owned run does not converge, including when `mise upgrade` itself exited non-zero (network/auth failure) — the rc line at `:277` is right above, but the advice still names a pin that may not be the cause.
- `src/remediations.ts:118`: the linked target is never checked for existence/executability (`[ -x "$M" ]`), unlike the on-PATH probe in the updater which does gate on `[ -x "$cxpath" ]` (`src/codex-update.ts:234`); a `mise which` that prints a path for a declared-but-not-installed tool would leave a dangling link where the native installer would have produced a working binary.
- `src/codex-update.ts:232/237`: a `MISE_DATA_DIR` with a trailing slash makes the case pattern `"$MISE_DATA"/*` (double slash) fail to match; cheap to normalise, and harmless today.
- `docs-site/src/content/docs/operating.md:112`: the plan named `getting-started.md`; `operating.md` is a defensible home, just noting the divergence.

- `src/codex-update.ts` predates this PR in using `type -a codex` (a bashism) inside a script that is otherwise POSIX-careful — fine under `bash -lc`, noted only because the new probe sits beside it.

## Limitations

- I could not execute `mise which`, `mise upgrade`, or the emitted shell strings (read-only review). Findings above are grounded in the on-disk mise layout I inspected and in the repo code paths cited; the one external assumption is stated in “Host ground truth”.
- mise's exact `upgrade` retention policy is inferred from five coexisting node versions on the host, not from mise's source. Either policy makes finding 1 a defect: retention ⇒ a silent stale binary, removal ⇒ a dangling link and a check that goes red again.

Issue #2028, opened 2026-08-04
Flaky UI lane: EpicPanel each_key_duplicate test leaks its rejection under load (all tests pass, exit 1)


scoop:
Observed while rebasing #2025 (unrelated branch — it does not touch `EpicPanel`).

## Symptom

The `ui/` suite exits **1 with every test passing**:

```
Test Files  289 passed (289)
      Tests  4532 passed (4532)
```

...alongside an unhandled rejection attributed to `EpicPanel.svelte:214`:

```
[Unhandled rejection] Svelte error: each_key_duplicate
Keyed each block has duplicate key `709` at indexes 0 and 1
	in EpicPanel.svelte
```

Only in the full parallel run. Clean on a full re-run, and 3/3 green running `EpicPanel.browser.test.ts` in isolation.

## Not a product bug

My first read was that this was a defect in `EpicPanel`s keyed `{#each epic.children as c (c.number)}`. It is not — the duplicate `709` is **constructed by the test itself**:

`ui/src/lib/components/EpicPanel.browser.test.ts:191` — *"crashes the child list with each_key_duplicate on a duplicate child number"* — deliberately renders `children: [child({number: 709}), child({number: 709})]` and asserts Svelte complains. The product code and its data are fine.

## Likely mechanism

The test catches the rejection with a temporary `window.addEventListener("unhandledrejection", …)` + `ev.preventDefault()`, then tears the listener down in a `finally` after a **fixed 50 ms sleep**:

```ts
render(EpicPanel, { … });                        // not awaited
await new Promise((r) => setTimeout(r, 50));
expect(captured).toMatch(/each_key_duplicate/);
} finally {
  window.removeEventListener("unhandledrejection", onRejection);
}
```

The rejection is scheduled asynchronously by Svelte, so the 50 ms is a deadline, not a guarantee. Under a loaded 289-file parallel run that window can be missed: the `finally` removes the listener, the rejection lands with no handler, and it escapes to vitest as an unhandled rejection → non-zero exit, while the assertion itself had already passed.

That matches the observed signature exactly (passes in isolation where 50 ms is ample; only fails under load; tests all green but process exits 1). Stated as the most plausible mechanism, not something I instrumented — I did not reproduce it deterministically.

Note `render(...)` is not awaited here, unlike the convention elsewhere in the repo, which likely widens the window further.

## Suggested fix

[...]

PR #2056, opened 2026-08-08
feat(herdr): support herdr 0.8.0 (protocol 19) + guard the last-tab close

### LINKED ISSUE #2039: Support herdr 0.8.0 (protocol 19): gate passes clean, but last-tab close now destroys the workspace
herdr 0.8.0 (protocol 19) is out and Shepherd blocks it: the in-app updater refuses the jump and
the preflight banner warns, because `HERDR_LAST_SUPPORTED_VERSION` is pinned to `0.7.5`.

This issue is the research that satisfies the record-shape gate #2032 demands before that pin moves,
plus the one real behavioural regression it surfaced.

**Headline: the wire protocol is safe, the CLI is unchanged, and Shepherd's spawn path works
as-is on 0.8.0. There is exactly one behavioural break, and it is in tab teardown, not spawning.**

---

## How this was verified

Not from release notes alone. A real `herdr-linux-x86_64` v0.8.0 binary was run as an **isolated
headless server** (own `HOME`/`XDG_*`/socket via a short symlinked path), so the live 0.7.5 daemon
and its running session were never touched. The installed 0.7.5 binary was run the same way as a
baseline, and every claim below is a measured A/B between the two servers.

- Schema diff: `herdr api schema --json` from the 0.8.0 binary vs the vendored
  `src/generated/herdr-schema.json` (protocol 17).
- CLI diff: `--help` for all 23 subcommands Shepherd actually invokes.
- Behaviour: Shepherd's exact spawn and teardown sequences replayed against both servers.

---

## 1. Wire protocol 17 → 19 is purely additive

Two protocol bumps in one release, but the diff is small and contains **no removals and no changed
result shapes**:

| Change | Detail |
| --- | --- |
| Method added | `workspace.move_block` |
| Subscription added | `workspace.reordered` |
| Event kind added | `workspace_reordered` (+ its `EventData` variant) |
| Enum widened | `IntegrationTarget` += `grok`, `antigravity_cli` |
| Methods removed | **none** |
| Method params changed | **none** |
| Result shapes changed | **none** |

## 2. The #2032 record-shape gate passes

#2032 exists because a herdr record-shape change silently killed four husk reapers twice (#721,
then #2029 when 0.7.5 dropped `pane.label` and `agent.name`). Its checklist, answered:

**Required/optional + nullable flags — zero drift.**

| Def | Fields | Result |
| --- | --- | --- |
| `SuccessResponseTabInfo` | 7 | identical |
| `SuccessResponsePaneInfo` | 19 | identical |
| `SuccessResponseAgentInfo` | 22 | identical |
| `EventTabInfo` / `EventPaneInfo` | 7 / 19 | identical |

No field moved required → optional. That is the exact bug class of #2029, and it is clean.

**Live reaper assumptions — all four hold, measured on both versions:**

| Assumption | 0.7.5 | 0.8.0 |
| --- | --- | --- |
| `tab.list` returns non-null `label` for every tab | yes | yes |
| `agent_status` is `unknown` for an agentless tab | yes | yes |
| `pane.process_info` returns foreground procs for an agentless pane | yes (`bash`) | yes (`bash`) |
| `tab_id`s not reused across a close (#569) | not reused | not reused |

## 3. The one real breaking change: last-tab close destroys the workspace (herdr #1760)

herdr's 0.8.0 note reads *"Closing a workspace's last tab through the CLI or API now closes the
workspace like the TUI does."* Measured:

```
# herdr 0.7.5
$ herdr tab close w3:t1          # the workspace's last tab
{"error":{"code":"tab_close_failed","message":"cannot close the last tab in a workspace"}}
# -> workspace survives

# herdr 0.8.0
$ herdr tab close w1:t1          # the workspace's last tab
{"result":{"type":"ok"}}
$ herdr workspace list
{"result":{"workspaces":[]}}     # <- workspace destroyed
$ herdr tab create --cwd ... --label x --no-focus
{"error":{"code":"workspace_not_found","message":"no active workspace"}}
```

**Why this matters for Shepherd.** `HerdrDriver.closeTab()` in `src/herdr.ts` swallows every error
(`/* best-effort; tab may already be gone */`). On 0.7.5 the refusal above was swallowed harmlessly
and the workspace lived on. On 0.8.0 the close *succeeds* and silently takes the workspace with it.
Shepherd has four teardown mechanisms that mass-close tabs, including boot reconcilers in
`src/transient-tab-reaper.ts` that close every unowned label-prefixed tab at once — a workspace
holding only Shepherd tabs is destroyed wholesale.

**Severity: degradation, not a hard break.** `ensureWorkspace()` runs before every spawn in *both*
drivers (`src/herdr.ts`, `src/herdr-socket-driver.ts`) and recreates a workspace when none exists.
Verified end to end on 0.8.0: reap both tabs → workspace gone → next spawn's
`workspace create` + `tab create` succeeds. So agents still spawn. What is lost is the operator's
workspace itself — its label, tab arrangement and worktree-group state — silently replaced by a
bare `shepherd` workspace.

## 4. Everything else Shepherd drives is unchanged

- **CLI surface:** `--help` identical for every command Shepherd invokes (`tab create/close/list/rename`,
  `pane run/close/list/process-info/send-keys/send-text/report-agent/report-agent-session`,
  `agent list/rename/send`, `workspace list`, `worktree add/list/prune/remove`). Only cosmetic
  additions: a new `--skill` flag and "next:" hints.
- **`worktree --json`:** dropped from help (#2171) but still accepted — verified working.
- **Spawn path:** the 0.7.5 external-registration sequence (`tab create` → `pane run` →
  `report-agent-session` → `report-agent` → `agent list`) replayed on 0.8.0 returns byte-compatible
  shapes; `agent list` surfaces the registered agent with `agent`, `agent_status`, `pane_id`,
  `tab_id`, `terminal_id` exactly as parsed today.
- **No protocol handshake risk:** `HerdrSocketClient` sends only `{id, method, params}` and never
  negotiates a version, so a newer server cannot reject it on that basis.

## 5. Not fixed: the sandbox status regression stays

herdr **#1716** (an externally-registered agent's status cannot be lowered to `idle`) is still
**open**. Measured on both versions: reporting `--state idle` lands on `done` identically on 0.7.5
and 0.8.0. So `HERDR_LAST_FULL_SANDBOX_STATUS_VERSION = "0.7.4"` and the two-path sandbox downgrade
advisory in `src/herdr-update.ts` must stay exactly as they are. 0.8.0 neither fixes nor worsens it.

## 6. Upstream moved: org rename and relicense

- **`ogulcancelik/herdr` → `herdrdev/herdr`.** Shepherd hardcodes the old org in
  `src/herdr-install.ts` (`herdrReleaseUrl`, `herdrReleaseTagUrl` — used by the pinned install
  command *and* the in-app downgrade), `src/preflight.ts`, `src/remediations.ts`'s install
  string, `ui/src/lib/components/HerdrUpdateModal.svelte`, plus `README.md` and docs-site. Nothing
  is broken today because GitHub redirects renamed repos, but the whole install/downgrade path
  hangs off a redirect that upstream can retire at any time.
- **Relicensed AGPL-3.0-or-later → Apache-2.0.** Worth reflecting wherever Shepherd documents
  herdr's licence.

---

## What has to change

1. **`src/herdr-capabilities.ts`** — `HERDR_LAST_SUPPORTED_VERSION` and
   `HERDR_LAST_SPAWNABLE_VERSION` `0.7.5` → `0.8.0`. Leave `HERDR_LAST_FULL_SANDBOX_STATUS_VERSION`
   at `0.7.4` (§5). Note `HERDR_EXTERNAL_REGISTRATION_VERSION` stays `0.7.5` — it is a floor, and
   0.8.0 correctly stays on the external-registration path.
2. **`src/config.ts`** — `HERDR_SOCKET_SUPPORTED_PROTOCOLS` `{16, 17}` → add `19`. The set is
   deliberately an explicit allowlist, not a floor; without this an opted-in
   `SHEPHERD_HERDR_SOCKET=1` install silently falls back to the CLI driver.
3. **Regenerate the vendored protocol** — `bun run gen:herdr-schema` against a 0.8.0 binary, then
   `bun run gen:herdr-types`. `HERDR_PROTOCOL` moves 17 → 19 and `test/herdr-schema.test.ts` starts
   agreeing again.
4. **Last-tab guard in `closeTab`** — decline the close when the target is its workspace's last
   tab, restoring 0.7.5 semantics at the single choke point all four reapers funnel through. Needed
   in *both* `src/herdr.ts` and `src/herdr-socket-driver.ts`. Accepted cost: one empty husk tab may
   linger per workspace.
5. **Migrate the org** — `ogulcancelik` → `herdrdev` across install/downgrade URLs, remediation
   strings, UI and docs.
6. **Docs** — the "supports herdr up to 0.7.5" statements in `README.md` and
   `docs-site/src/content/docs/getting-started.md`, and `EXPECTED_HERDR_VERSION` in
   `docs-site/scripts/gen-cli-reference.ts` (still pinned to `0.7.0`).

## Opportunities worth tracking (not in scope here)

- **`HERDR_PROCESS_DETECTION=child-groups` (herdr #1982)** — the highest-value one. Shepherd's bwrap
  membrane uses `--unshare-pid`, which is precisely why herdr cannot passively detect sandboxed
  agents, why Shepherd must externally register them, and therefore why the #1716 status regression
  bites at all. If child-group detection sees through the membrane, native status detection for
  sandboxed agents could come back and the sandbox downgrade advisory could retire.
- **`workspace.move_block` + `workspace.reordered`** — atomic worktree-group reordering, if Shepherd
  should ever drive sidebar ordering.
- **`herdr --skill`** — herdr now ships its own agent skill. Shepherd ships one in `agent-skills/`;
  worth diffing for overlap.

## Success criteria

- Shepherd runs on herdr 0.8.0 with agent spawn, steer and teardown intact.
- The in-app updater permits 0.7.5 → 0.8.0 and the preflight banner clears.
- A last-tab reap no longer destroys the operator's workspace.
- `bun run check:herdr-types` is clean with `HERDR_PROTOCOL = 19`.
- No install or downgrade path depends on a GitHub org redirect.

kai-osthoff:
Closes #2039.

Raises Shepherd's herdr ceiling to **0.8.0 (protocol 19)**, so the in-app updater stops refusing
the upgrade — and fixes the one behavioural regression 0.8.0 brings.

## Verified, not inferred

Measured A/B against an **isolated herdr 0.8.0 headless server** (own `HOME`/`XDG_*`/socket) versus
the installed 0.7.5, with the live daemon never touched. Full evidence in #2039.

- **Protocol 17 → 19 is purely additive**: `+workspace.move_block`, the `workspace.reordered`
  subscription and event, `IntegrationTarget` `+grok`/`+antigravity_cli`. Nothing removed, no param
  or result shape reshaped.
- **The #2032 record-shape gate passes clean**: zero required→optional or nullability drift across
  `TabInfo` (7 fields), `PaneInfo` (19), `AgentInfo` (22). All four live reaper assumptions hold on
  both versions. That gate exists because this exact bug class silently killed four husk reapers in
  #2029.
- **End-to-end against a real 0.8.0 daemon**, driven through the production `HerdrDriver` (not a
  fake runner): concurrent final-two-tab reap leaves the workspace standing; a teardown still takes
  the last tab down; the spawn path recovers afterwards.

## The regression this fixes (herdr #1760)

On 0.7.5, `tab close` on a workspace's last tab returned `tab_close_failed` and the workspace
survived. On 0.8.0 it succeeds and **closes the workspace**, after which `tab create` returns
`workspace_not_found`. It is silent in Shepherd because `closeTab` swallows errors as best-effort,
so a boot reap could take the operator's workspace — label, tab arrangement, worktree-group state —
with it.

**The shipped contract is decline-OR-rebuild**, split by one question: does a live process have to
die for the caller to succeed?

- **No → decline.** Housekeeping of tabs holding nothing we need: the husk and transient reapers,
  doc-agent's husk cleanup. This is where the #1760 regression actually bit.
- **Yes → close, then rebuild.** These pass `allowLastTab`, which closes and immediately puts a
  workspace back via the idempotent `ensureWorkspace`, so the process dies **and** the operator is
  never left workspace-less. Three classes: **teardown** (`stopViaRecordedTab`, clean-terminal
  archive), **eviction of a live squatter whose name we need** (the `agent_name_taken` retries in
  both drivers, `ReviewService.reapOrphans`), and **rollback of a tab we ourselves just created**.

Declining for those classes is not the safe option — it strands what it refuses to close.
`tab-reaper` deliberately spares any tab holding a non-shell process, so nothing else ever reclaims
one. For eviction specifically, eviction is the only thing that changes state between retry
attempts, so a declined close turns the retry into a no-op loop and fails the spawn named — the
sticky `agent_name_taken` class fixed in #751/#508.

The rebuild is seeded with the closed tab's own pane cwd, and still runs when herdr reports no cwd
(the protocol declares it optional/nullable) — it just omits `--cwd`, which a live 0.8.0 daemon
accepts. It is also armed when the last-tab read itself fails, so a transient `tab.list` error
cannot cost a workspace. Two deliberate limits: `ensureWorkspace` creates only when **zero**
workspaces remain, so a deleted workspace is not restored while another survives; and its create
can race `start`'s own `ensureWorkspace` on a different chain, yielding two — benign and
self-limiting.

**The check and the close are serialized**, and that is load-bearing rather than tidiness: every
reaper is launched fire-and-forget with `void` (`distiller`, `merge-suggest`, `optimizer`,
`standalone-critic`), so on a workspace down to its final two tabs an unserialized read-then-close
lets both observers see two tabs, both conclude they are not closing the last one, and together
delete the workspace. It runs on its own chain — sharing `serializeStart` would deadlock, since a
spawn rollback calls `closeTab` from inside that chain.

## Deliberately unchanged

`HERDR_LAST_FULL_SANDBOX_STATUS_VERSION` stays at **0.7.4**. herdr #1716 is still open, and a
reported `idle` for an externally-registered agent lands on `done` identically on 0.7.5 and 0.8.0 —
so the two-path sandbox downgrade advisory must keep working. `HERDR_EXTERNAL_REGISTRATION_VERSION`
stays 0.7.5; it is a floor, and 0.8.0 correctly stays on that spawn path.

The protocol allowlist gains `19` **by name**. It is an allowlist, not a `>=` floor, and 18 never
shipped stable.

## Also here

- **Org rename** `ogulcancelik/herdr` → `herdrdev/herdr`. Three URL sites in `herdr-install.ts`, not
  two: `herdrReleaseTagUrl`, `herdrReleaseUrl`, and the raw template inside
  `herdrPinnedInstallCommand`, which builds its own because it assembles the asset key in the shell
  from `uname`. GitHub still redirects renamed repos, so nothing was broken — but the whole install
  and downgrade path hung off that redirect.
- **A What's-New entry**, because the operator-visible symptom was a contradiction: an update badge
  driven purely by `latest > current` next to a modal refusing the update. Note the badge in
  `TopBar.svelte` still ignores `latestUnsupported` — worth its own change, out of scope here.

## Known costs

- The bypass path pays an extra `tab.list`; it previously short-circuited past the last-tab check
  and so could not know whether a rebuild was owed. Several socket tests assert exact wire
  sequences and are updated, with the reason recorded at each assertion.
- `stopViaRecordedTab` now reads the tab list twice — once for its own label check, once inside
  `closeTab`. Threading that state through would widen the driver API for a teardown-path
  micro-optimisation, so it is left as is.

## Testing

- Root: `bun run lint`, `bun run test` → **8654 pass, 0 fail**. `bun run check:herdr-types` clean at
  `HERDR_PROTOCOL = 19`.
- UI: `bun run check` clean; `bun run test` → 4534 pass. The suite exits 1 on the pre-existing
  `EpicPanel` `each_key_duplicate` flake tracked in #2028 — this branch's only UI logic change is a
  one-line URL constant, so it cannot be the cause.
- New coverage, each verified to fail without its fix: last-tab decline, sibling close,
  cross-workspace counting, the concurrent final-two-tabs race, teardown of a sole tab, sole-tab
  rebuild (with and without a reported cwd, and on a failed tab-list read), squatter eviction of a
  sole tab, and the shell-tab rollback — in both drivers plus `herdr-parity.test.ts` so the two
  implementations cannot drift.


kai-osthoff (commented):
# Review: herdr 0.8.0 support

Decision: comment

No blocking findings.

## Verified

- The support and spawn ceilings are both `0.8.0`, while the external-registration floor remains `0.7.5` and the sandbox-status floor remains `0.7.4` (`src/herdr-capabilities.ts:18`, `src/herdr-capabilities.ts:25`, `src/herdr-capabilities.ts:31`, `src/herdr-capabilities.ts:40`). The capability tests cover both the new ceiling and the retained external-registration behavior (`test/herdr-capabilities.test.ts:27`, `test/herdr-capabilities.test.ts:66`, `test/herdr-capabilities.test.ts:81`, `test/herdr-capabilities.test.ts:96`).
- Protocol 19 is explicitly allowlisted (`src/config.ts:76`, `src/config.ts:85`); the vendored schema declares protocol 19 (`src/generated/herdr-schema.json:3`), the generated types expose the same constant and the new `workspace.move_block` request (`src/generated/herdr-protocol.ts:2981`, `src/generated/herdr-protocol.ts:3000`), and the schema gate compares the schema to both of those sources (`test/herdr-schema.test.ts:51`).
- Both drivers serialize the last-tab read/close operation and decline only a sole tab in the target workspace (`src/herdr.ts:658`, `src/herdr.ts:856`, `src/herdr.ts:1502`, `src/herdr-socket-driver.ts:47`, `src/herdr-socket-driver.ts:482`). The cross-driver tests model the concurrent two-tab case and retain one tab (`test/herdr-parity.test.ts:235`, `test/herdr-parity.test.ts:245`); the socket-specific test independently verifies the same request behavior (`test/herdr-socket-driver.test.ts:269`).
- Re-review point 1 is addressed: the incomplete shell-tab rollback supplies `allowLastTab: true`, so a half-created sole tab is closed rather than leaked (`src/herdr.ts:1061`, `src/herdr.ts:1065`), with a focused regression test (`test/herdr.test.ts:504`, `test/herdr.test.ts:523`). The equivalent socket spawn rollbacks also opt out (`src/herdr-socket-driver.ts:240`, `src/herdr-socket-driver.ts:382`), and the session-row rollback does too (`src/service.ts:3609`, `src/service.ts:3613`).
- Re-review point 2 is addressed: the `closeTab` documentation now identifies both teardown and self-created-tab rollback as the two `allowLastTab` caller classes, rather than claiming a single bypass (`src/herdr.ts:1491`, `src/herdr.ts:1499`). The implementation agrees: both recorded-tab and agent-list `stop` arms opt out (`src/herdr.ts:765`, `src/herdr.ts:785`), as does clean-terminal archive (`src/service.ts:5689`, `src/service.ts:5695`).
- The release-org migration covers tag URL, asset URL, and the shell-assembled pinned-install URL (`src/herdr-install.ts:39`, `src/herdr-install.ts:50`, `src/herdr-install.ts:168`); the UI link and user docs use the same `herdrdev/herdr` slug (`ui/src/lib/components/HerdrUpdateModal.svelte:39`, `README.md:141`, `docs-site/src/content/docs/getting-started.md:34`).
- The new announcement is discovered by the catalog's eager entries glob (`ui/src/lib/feature-announcements.ts:47`) and its title/body keys exist in both maintained catalogs (`ui/src/lib/feature-announcements/entries/v1.47.0-herdr-080.ts:4`, `ui/messages/en.json:3560`, `ui/messages/de.json:3560`).

## Limitations

- I did not run tests, builds, or external commands, per the review instruction. The claimed live-daemon research and external issue #2039 are not repository-local ground truth, so I could not independently verify them here.

kai-osthoff:
# Review: request changes

## Findings

1. src/herdr.ts: `stopViaRecordedTab()` now passes `allowLastTab: true` for both its recorded-tab and live-agent paths. `SessionService.archive()` calls `stop()` for every non-terminal session, so archiving the sole Shepherd session still sends `tab close` and lets herdr 0.8.0 delete the workspace. This bypasses the new guard precisely on a normal teardown path; retain the guard here (and accept the documented anchor/husk tab) or use a teardown operation that does not remove the final tab. Verified the bypass at `src/herdr.ts:764-769` and `src/herdr.ts:782-786`, and its reachable archive caller at `src/service.ts:5740-5742`.

2. src/service.ts: terminal-session archival explicitly uses `closeTab(..., { allowLastTab: true })`. A terminal can likewise be its workspace's only tab, so archive deletes that workspace instead of preserving it. This is the same 0.8.0 regression on the terminal branch; do not bypass the guard for ordinary archival. Verified at `src/service.ts:5732-5739`; the guard it bypasses is implemented at `src/herdr.ts:1502-1514` and `src/herdr-socket-driver.ts:482-494`.

## Verified implementation details

- Protocol admission is internally aligned: protocol 19 is allowlisted in `src/config.ts:76-85`, the generated schema declares protocol 19 at `src/generated/herdr-schema.json:1-6`, and the generated TypeScript constant is 19 at `src/generated/herdr-protocol.ts:2978`.
- The last-tab decision uses the target tab's required workspace identifier (`src/herdr.ts:658-662`; schema requirement at `src/generated/herdr-schema.json:1032-1064`) and both driver implementations serialize the check and close (`src/herdr.ts:1502-1514`, `src/herdr-socket-driver.ts:482-494`).
- The new announcement fragment is discovered by the eager entry glob (`ui/src/lib/feature-announcements.ts:47-50`) and both referenced catalog keys exist in English and German (`ui/src/lib/feature-announcements/entries/v1.47.0-herdr-080.ts:3-8`, `ui/messages/en.json:3613-3614`, `ui/messages/de.json:3613-3614`). Its 1.47.0 version is the next minor after the release manifest's 1.46.0 (`.release-please-manifest.json:1-3`, `scripts/next-version.mjs:60-63`).

## Limitations

Could not verify the claimed live herdr 0.8.0 behavior or GitHub research issue from repository contents; those external facts are not treated as confirmed by this review.

kai-osthoff:
Both blocking findings describe the same deliberate trade-off, so one note for both:

1. `src/herdr.ts` — `stopViaRecordedTab` bypasses the last-tab guard
2. `src/service.ts` — terminal-session archival bypasses it

**The observation is correct: these can destroy a workspace that is down to its last tab.** I'm not
disputing the mechanism. I am disputing that removing the bypass is the right fix, because it trades
a recoverable loss for an unrecoverable one.

**What removing it costs.** If teardown declines the close, the tab stays — and so does what's in it:

- **Agent sessions:** a live `claude` survives an archive. Nothing automatic ever clears it:
  `tab-reaper.ts` deliberately spares any tab whose pane holds a non-shell process ("any pane live
  (a non-shell proc: `claude` / `node` / …) → spare the whole tab"). So it keeps running against the
  operator's subscription and holds its worktree until a human kills it by hand.
- **Terminal sessions:** the shell tab becomes a permanent husk. The husk reaper *would* reap it, but
  it keeps the guard — so on a sole tab it declines too, forever.

Measured, not argued: removing the bypass turns four pre-existing `stop:` tests red — they predate
this PR and encode Shepherd's stop contract (`stop (0.7.5): closes the recorded tab of a started
agent`, `stop: closes the spawn-recorded tab when the agent never appears in agent list`, `stop: a
second stop after the recorded-tab close is a plain fallback no-op`, `stop after relabel: …`).

**What keeping it costs.** The tab being closed is its workspace's last, so the workspace was about
to be empty regardless. What is lost is that empty workspace's existence and label — and
`ensureWorkspace()` rebuilds one before the next spawn, verified end to end against a live 0.8.0
daemon (reap both tabs → workspace gone → next `workspace create` + `tab create` succeeds).

**Why this is also the platform-correct behaviour.** herdr 0.8.0's own note for #1760 reads
*"Closing a workspace's last tab through the CLI or API now closes the workspace **like the TUI
does**."* For an explicit operator teardown, closing the workspace is what herdr itself does when the
operator closes that tab by hand. Guarding there would make Shepherd behave differently from herdr
for the same user-initiated action.

**Where the guard does belong — and still applies** — is housekeeping of tabs Shepherd no longer
owns, which is where the actual #1760 regression bit: the boot reapers, fired concurrently with
`void`, silently taking the operator's workspace during cleanup. Those keep the guard, as does
squatter eviction. The two caller classes that bypass are teardown-that-must-tear-down and rollback
of a tab we ourselves just created.

**What I changed in response:** the decision was only pinned indirectly, via the four `stop:` tests.
Pushed `a0954b10`, which names it — `stop closes the agent's tab even as its workspace's SOLE tab —
a stop must stop` — and records the trade-off and its cost at the test. Verified it fails when the
bypass is removed, so it pins the contract rather than merely describing it.

If you still read the workspace loss as the greater harm, the fix I'd reach for is *not* re-guarding
teardown but re-creating the workspace right after such a close, so the agent still dies and the
operator keeps a workspace. Say the word and I'll build that instead.

kai-osthoff:
# Review — request changes

## Findings

- `src/herdr.ts: stopViaRecordedTab bypasses the last-tab guard, so normal non-terminal archives can delete the sole workspace.` This remains the normal archive path: `stopViaRecordedTab` passes `allowLastTab: true` for both the recorded-tab and agent-list cases (`src/herdr.ts:765-785`), and `SessionService.archive()` calls `herdr.stop()` for every non-terminal session (`src/service.ts:5740-5742`). `closeTab()` only evaluates `isLastTabInWorkspace()` when that option is false (`src/herdr.ts:1502-1511`), so a sole-tab agent archive issues the destructive close that the new guard is intended to prevent. The newly added test explicitly codifies this bypass rather than exercising a guarded archive (`test/herdr.test.ts:506-519`).

- `src/service.ts: terminal-session archival bypasses the last-tab guard, so it can delete the sole workspace.` The terminal branch of `archive()` now directly calls `closeTab(..., { allowLastTab: true })` (`src/service.ts:5732-5741`). That option skips the sole-tab check (`src/herdr.ts:1502-1511`), allowing ordinary terminal-session archival to remove the workspace. This is not merely a rollback path: it executes whenever `s.terminal` is true during archive (`src/service.ts:5732-5741`).

## Verified implementation details

- The version ceilings are both `0.8.0` while the external-registration floor remains `0.7.5` (`src/herdr-capabilities.ts:18-31,75-86`). Protocol 19 is explicitly allowlisted (`src/config.ts:76-85`) and the generated client declares `HERDR_PROTOCOL = 19` plus `workspace.move_block` (`src/generated/herdr-protocol.ts:2981-3001`); the repository's schema gate compares those two values (`test/herdr-schema.test.ts:51-55`).
- The three production release URL constructions use `herdrdev/herdr` (`src/herdr-install.ts:39-51,168-190`), and the user-facing docs use the same slug (`README.md:139-148`, `docs-site/src/content/docs/getting-started.md:32-39`).
- Both maintained UI catalogs contain the new feature keys (`ui/messages/en.json:3613-3614`, `ui/messages/de.json:3613-3614`), and the new announcement supplies those exact keys (`ui/src/lib/feature-announcements/entries/v1.47.0-herdr-080.ts:3-10`). The catalog loader includes entry files by glob (`ui/src/lib/feature-announcements.ts:47-50`).

## Limitation

I did not verify the claimed external research issue or a live herdr 0.8.0 daemon: neither external state is available in this checkout. No build, test, or runtime command was run.

kai-osthoff:
Both findings are now **fixed**, not argued — I built the third option I offered last round instead
of re-litigating.

You were right that the bypass could destroy a workspace on herdr 0.8.0. Re-guarding teardown was
still the wrong remedy: declining leaves a **live** agent that ignored a stop, and `tab-reaper`
deliberately spares any tab holding a non-shell process, so nothing ever reclaims it. That trades a
recoverable loss for an unrecoverable one.

So `closeTab` now separates the decision from the consequence on the sole-tab path:

- **Housekeeping** (reapers, doc-agent cleanup, squatter eviction) — unchanged: declines the close.
- **Teardown and self-rollback** — closes, *and* immediately rebuilds a workspace via the idempotent
  `ensureWorkspace`, seeded with the closed tab's own pane cwd. The agent dies **and** the operator
  keeps a workspace. No caller can now delete the sole workspace.

`ensureWorkspace` self-checks, so this is a no-op on 0.7.5 (which refuses the close) and whenever a
workspace survives.

Pinned in both drivers plus parity, each verified to fail without the fix:
`closeTab({allowLastTab}) on a SOLE tab puts the workspace back`,
`closeTab({allowLastTab}) does NOT create a workspace when siblings survive the close`, and
`both rebuild the workspace after allowLastTab closes a sole tab`.

One honest cost: the bypass path now pays an extra `tab.list`. It previously short-circuited past
the last-tab check and therefore could not know whether a rebuild was owed. Four socket tests assert
exact wire sequences and are updated, with the reason recorded at each assertion.

Root suite 8650 pass / 0 fail.

kai-osthoff:
# Review: request changes

## Summary

The PR correctly lifts the 0.8.0 version/protocol gates and updates the published install URLs, but the new `allowLastTab` recovery is incomplete for a response shape the vendored protocol explicitly permits. A teardown can still leave the operator with no workspace.

## Findings

1. src/herdr.ts: `allowLastTab` closes a sole tab even when no restorable cwd can be obtained, then silently skips `ensureWorkspace`, so a legal `pane.list` response with the target pane's optional/null `cwd` deletes the sole workspace on herdr 0.8.0. `tabCwd()` returns `null` for that response (`src/herdr.ts:1534-1540`), and the only restore is conditional on a truthy cwd (`src/herdr.ts:1512-1527`). The regenerated contract explicitly declares `SuccessResponsePaneInfo.cwd` optional and nullable (`src/generated/herdr-protocol.ts:2273-2281`); the normal parser also normalizes a missing cwd to `""` (`src/herdr.ts:906-916`). This path is reachable from ordinary non-terminal archival because `stopViaRecordedTab` intentionally supplies `allowLastTab` on both recorded and agent-list paths (`src/herdr.ts:740-785`), and terminal archival does the same (`src/service.ts:5732-5741`). The socket implementation duplicates the same conditional recovery (`src/herdr-socket-driver.ts:482-508`) and additionally obtains panes through the CLI delegate (`src/herdr-socket-driver.ts:127-130`). Do not close a known sole tab unless recovery is guaranteed, or supply a safe, independently available workspace cwd and assert the restoration; the current best-effort no-op violates the stated no-workspace-loss guarantee.

Accepted: `stopViaRecordedTab` no longer bypasses the serialized close path. Both its recorded-tab and agent-list arms call `closeTab(..., { allowLastTab: true })` (`src/herdr.ts:747-786`), which performs the last-tab read before deciding whether a restoration is due (`src/herdr.ts:1510-1528`). This addresses the earlier bypass in the normal non-terminal archive path when the pane cwd is available.

Accepted: terminal-session archival now also calls `closeTab(..., { allowLastTab: true })` (`src/service.ts:5732-5741`), so it takes the same serialized last-tab/recovery path rather than the former unguarded close. This addresses the earlier terminal archival bypass when the pane cwd is available.

## Verified implementation details

`HERDR_LAST_SUPPORTED_VERSION` and `HERDR_LAST_SPAWNABLE_VERSION` are both 0.8.0, while the external-registration floor remains 0.7.5 and the sandbox-status floor remains 0.7.4 (`src/herdr-capabilities.ts:18-40`). Protocol 19 is expressly included in the socket allowlist (`src/config.ts:76-85`), and the generated protocol declares 19 plus `workspace.move_block` (`src/generated/herdr-protocol.ts:2981-3001`). The in-repo schema gate compares the vendored schema protocol to that constant and checks allowlist membership (`test/herdr-schema.test.ts:51-55`).

The three release-URL construction sites now use `herdrdev/herdr`: tag URL, asset URL, and the shell-built pinned installer URL (`src/herdr-install.ts:37-52`, `src/herdr-install.ts:168-179`). The modal points at the same release repository (`ui/src/lib/components/HerdrUpdateModal.svelte:35-40`), and the README/docs support statement says 0.8.0 (`README.md:141-148`; `docs-site/src/content/docs/getting-started.md:112-121`).

The new announcement fragment's title/body keys exist in both maintained UI catalogs (`ui/src/lib/feature-announcements/entries/v1.47.0-herdr-080.ts:3-10`; `ui/messages/en.json:3613-3614`; `ui/messages/de.json:3613-3614`), and the catalog eagerly loads all entry fragments (`ui/src/lib/feature-announcements.ts:47-76`).

## Verification limitations

I did not execute builds, tests, generators, or a live herdr daemon, per the review instruction. The claim that protocol 17→19 is externally additive and that live 0.8.0 behavior matches the vendored schema could not be independently verified from this repository.

kai-osthoff:
🟢✅ READY since forever 😴⏳ — someone fancy doing their job and merging? 👀🚀🤞🎉🙏

scoop (dismissed):
Reviewed against the description — every claim in the body was checked against the code, and the branch was rebased onto `main` locally to settle what the "conflicting" flag actually means.

The 0.8.0 ceiling work is right, and the evidence standard in the body is genuinely high. One blocking finding, four smaller ones, plus the rebase.

---

## Blocking — the guard also disarms the paths that evict a *live* squatter

`HerdrDriver.closeTab`'s new doc block sets out the taxonomy this PR is judged by: **teardown that must actually tear down** and **rollback of a tab we ourselves just created** get `allowLastTab`; **housekeeping of tabs we no longer own** keeps the guard. It then files squatter eviction under housekeeping:

> Everything that keeps the guard is HOUSEKEEPING of tabs we no longer own: the husk/transient reapers and doc-agent cleanup, plus squatter eviction in both drivers and the PR critic.

Four call sites in that list are not tabs we no longer own — they hold a **live process whose name we need**:

| Site | What it closes |
| --- | --- |
| `src/review.ts` → `reapOrphans` | its own comment: *"Free the name AND kill the still-alive claude process by closing its herdr tab"* |
| `src/herdr.ts` → `HerdrDriver.startAgentWithCollisionRetry` | same-named squatter, on `agent_name_taken` |
| `src/herdr-socket-driver.ts` → `startAgentWithCollisionRetry` | same |
| `src/herdr-socket-driver.ts` → `registerAgentWithCollisionRetry` | same |

All four call `closeTab(tabId)` with no options, so on 0.8.0 **and on 0.7.5** they now silently decline whenever the squatter is the only tab left in its workspace. That case is not exotic: `ensureWorkspace` creates a workspace only when *zero* exist, so Shepherd's tabs live in whatever workspace is active — on a host where Shepherd is herdr's only client, "one live Shepherd tab left" is a normal steady state.

Consequences:

- **Collision retry becomes a no-op loop.** Eviction is the only thing that changes state between attempts. Declined close → next attempt hits `agent_name_taken` again → `MAX_ATTEMPTS` exhausted → the spawn fails, named.
- **`reapOrphans` stops reclaiming the reviewer name.** That is precisely the sticky `REVIEW ERR` / `agent_name_taken` class fixed in #751/#508: an orphaned reviewer squatting `review TASK-<n>` blocks every re-spawn. Nothing else reclaims it — `src/tab-reaper.ts` reaps only tabs whose every pane classifies shell-only and deliberately spares `live`, which this PR's own comment cites as the reason `stopViaRecordedTab` needs the bypass. The identical argument applies verbatim here.

The fix looks cheap given what this branch already built: `allowLastTab: true` now *rebuilds* the workspace via `tabCwd()` → `ensureWorkspace()` rather than merely bypassing, so these sites get the same "agent still dies, operator never left workspace-less" contract as `stopViaRecordedTab`. My read is these four belong in the teardown class; if you deliberately concluded otherwise, the doc block should say why, because as written it asserts something about them that the code contradicts.

The genuine housekeeping sites — the husk/transient reapers and doc-agent cleanup — I agree keep the guard.

---

## Non-blocking

**1. The description no longer matches the branch.** The body says `closeTab` "declines that close in **both** drivers" and enumerates the `allowLastTab` opt-outs, but never mentions the workspace **rebuild** added by the last two commits (`tabCwd()` → `ensureWorkspace()` after a bypassed sole-tab close). The shipped contract is decline-**or**-rebuild; the body documents half of it. Since the rebuild is also the thing that makes the B1 fix cheap, it is worth promoting into the description rather than leaving it only in the code comment.

**2. `isLastTabInWorkspace` degrades silently if `workspace_id` ever goes missing.** `parseTabs` maps `workspaceId: t.workspace_id ?? ""`, so an absent field collapses every tab into one pseudo-workspace and the guard weakens to "decline only when the daemon has exactly one tab in total" — quiet, not loud. You verified the field is present on both 0.7.5 and 0.8.0, so this is fail-soft rather than a live bug, but given how much of the PR's value rests on that one field, it deserves a comment or an assertion at the parse seam.

**3. The rebuild is a no-op in the multi-workspace case.** `ensureWorkspace` creates only when zero workspaces exist. When a bypassed close deletes workspace B while workspace A survives, B is gone for good — label, tab arrangement, worktree-group state — and nothing rebuilds it. The code comment frames this as intended and "never left workspace-less" is literally true, but it is a strictly weaker promise than the guard gives, while the What's-New copy ("keeps your workspace, its label and its tab arrangement intact") reads as the stronger one. Either narrow the copy or note the limit.

**4. Two `ensureWorkspace` racers.** `closeTab`'s rebuild runs on `serializeClose`, `startImpl`'s runs on `serializeStart`. Both can observe zero workspaces and both call `workspace.create`, yielding two. Benign and self-limiting — but atomicity is this PR's whole thesis, so it earns a sentence.

---

## The merge conflict is a merge-driver artefact — rebase and it is gone

GitHub reports `CONFLICTING`, but the conflict is not semantic. Measured both ways:

- `git merge-tree --write-tree origin/main <head>` merges **clean** in a clone where husky's `prepare` has registered the `i18n-union` driver.
- The same command with `-c merge.i18n-union.driver=false` conflicts on exactly `ui/messages/en.json` and `ui/messages/de.json`, and nothing else.

GitHub cannot run `scripts/json-union-merge.mjs`, so it sees the additive tail hunk both branches appended to and calls it a conflict. A plain `git rebase origin/main` resolves it. I did that locally to confirm — all 8 commits replayed with no manual resolution, and the result is provably a union, not a pick:

- keys present on `main` but missing after rebase: **none**
- keys present on the branch but missing after rebase: **none**
- shared keys whose value drifted: **none**
- EN/DE parity: 3643 / 3643

**Please rebase and force-push** — I deliberately did not push anything, so the branch is untouched.

---

## Verified on the rebased branch

| Check | Result |
| --- | --- |
| root `bun run lint` | clean |
| root `bunx tsc --noEmit` | clean |
| root `bun run test` | 8645 pass / 4 fail — all 4 are the host-dependent `test/codex-update.test.ts` harness cases, which **also fail on `origin/main`** at the same 4 names. Not this branch. |
| herdr suites (`herdr`, `herdr-parity`, `herdr-socket-driver`, `herdr-capabilities`, `herdr-downgrade`, `herdr-update`, `herdr-install`, `herdr-socket-075`, `diagnostics`, `preflight`) | 418 pass / 0 fail |
| `ui/ bun run check` | 5956 files, 0 errors, 0 warnings |
| `bun run check:herdr-types` | clean; regeneration idempotent at `HERDR_PROTOCOL = 19` |

I skipped `ui/ bun run test` — the only UI logic change is a one-line URL constant, and the suite carries the known #2028 `EpicPanel` flake. Saying so rather than quietly omitting it.

---

## Confirmed sound

Checked and holding, so you know what was covered and not just what failed:

- **Allowlist by name, not a floor.** `HERDR_SOCKET_SUPPORTED_PROTOCOLS = new Set([16, 17, 19])` with 18 deliberately excluded — the right shape for a set that encodes "verified", and the comment says so.
- **17 → 19 really is additive** in the generated diff: `+workspace.reordered` (subscription, event data, event kind), `+grok` / `+antigravity_cli` on both `IntegrationTarget` unions. Nothing removed, no shape reshaped.
- **`HERDR_LAST_FULL_SANDBOX_STATUS_VERSION` correctly held at `0.7.4`** and `HERDR_EXTERNAL_REGISTRATION_VERSION` at `0.7.5` — the reasoning in the body matches what the constants do.
- **`serializeClose` is re-entrancy-safe.** Sharing `serializeStart` would indeed deadlock via the spawn rollback, and I checked the other direction too: neither driver's `ensureWorkspace` calls `closeTab`, so the rebuild inside the close chain cannot re-enter it.
- **Fail-open on a `tabsAsync()` throw** — right call, and the comment explains it.
- **All three `ogulcancelik` → `herdrdev` sites** in `herdr-install.ts`, including the shell-assembled template inside `herdrPinnedInstallCommand` that is easy to miss, plus the `HerdrUpdateModal` constant, README and docs-site.
- **Feature announcement**: `sinceVersion: "1.47.0"` is correct against `package.json` at `1.46.0` — the next unreleased version, so it will actually surface.
- **`herdr-parity.test.ts`** is the right place for this and covers the case that matters: concurrent closes on the final two tabs closing exactly one.

Once B1 is resolved (or argued down in the doc block) and the branch is rebased, this is good to land.

kai-osthoff:
# Review: request changes

## Findings

1. src/herdr-capabilities.ts: The JSDoc for `herdrSpawnSupported()` still says the driver spawns only up to `0.7.5`, while the exported `HERDR_LAST_SPAWNABLE_VERSION` is now `0.8.0` and the function compares against that constant. Correct the stale version so this capability contract does not contradict the code. Verified against `src/herdr-capabilities.ts:25` and `src/herdr-capabilities.ts:71`.

## Verified

- The prior finding is addressed: an `allowLastTab` close records a restore even when `tabCwd()` returns `null`, then unconditionally calls `ensureWorkspace`; the CLI implementation creates a workspace without `--cwd` for that nullable case. Verified against `src/herdr.ts:1523`, `src/herdr.ts:1549`, and `src/herdr.ts:975`.
- The socket driver follows the same nullable-cwd restoration contract and its generated request type permits `cwd?: string | null`. Verified against `src/herdr-socket-driver.ts:484`, `src/herdr-socket-driver.ts:505`, `src/herdr-socket-driver.ts:389`, and `src/generated/herdr-protocol.ts:2055`.
- The concurrent final-two-tab protection is serialized in both drivers, and the added parity coverage exercises both implementations. Verified against `src/herdr.ts:862`, `src/herdr.ts:1524`, `src/herdr-socket-driver.ts:50`, `src/herdr-socket-driver.ts:485`, and `test/herdr-parity.test.ts:265`.
- The vendored protocol and the socket allowlist agree on protocol 19; the generated request map includes `workspace.move_block`. Verified against `src/generated/herdr-schema.json:3`, `src/generated/herdr-protocol.ts:2981`, `src/generated/herdr-protocol.ts:3000`, and `src/config.ts:85`.
- The feature announcement is auto-discovered by the catalog glob, and both referenced message keys exist in English and German. Verified against `ui/src/lib/feature-announcements.ts:47`, `ui/src/lib/feature-announcements/entries/v1.47.0-herdr-080.ts:3`, `ui/messages/en.json:3643`, and `ui/messages/de.json:3643`.
- All executable/install URL builders in the changed runtime path use the renamed organization. Verified against `src/herdr-install.ts:39`, `src/herdr-install.ts:50`, and `src/herdr-install.ts:178`.

## Limitations

- I did not run tests, builds, or commands that mutate state, per the review request.
- The claimed live-herdr compatibility research and issue #2039 are external to this checkout, so I could not independently verify them from repository ground truth.

- `docs/superpowers/specs/2026-07-22-herdr-downgrade-design.md:27` still contains the former `ogulcancelik/herdr` URL.
- `docs/superpowers/plans/2026-07-22-herdr-downgrade.md:182` still contains the former `ogulcancelik/herdr` URL.
- `docs/superpowers/plans/2026-07-22-herdr-downgrade.md:299` still contains the former `ogulcancelik/herdr` URL.

kai-osthoff:
@scoop — all of your review is addressed and the branch is green. Ready for another look.

**Blocking: the guard disarmed live-squatter eviction.** You were right, and the doc block asserted
the opposite. Those sites close a tab holding a live process whose *name* we need, and eviction is
the only thing that changes state between retry attempts — declining made it a no-op loop. Fixed as
you suggested: they now take the same `allowLastTab` → close-and-rebuild contract. One correction
to your table: it is **five** sites, not four — `herdr.ts` has both `startAgentWithCollisionRetry`
and `registerAgentWithCollisionRetry`. Taxonomy in the doc block rewritten around the real dividing
line: *does a live process have to die for this caller to succeed?*

**All four non-blocking, fixed:**
1. PR description now documents the full decline-**or**-rebuild contract, not half of it.
2. `parseTabs`' `workspace_id ?? ""` fail-soft is called out at the parse seam, with a note to treat
   a future capture that drops the field as stop-and-review.
3. What's-New copy narrowed — it promised label and tab arrangement survive, which is only true for
   the declining path. Reworded in EN and DE.
4. The two `ensureWorkspace` racers are recorded in the contract.

**Also landed since your review**, from a critic round and my own re-read:
- The rebuild was gated on a truthy cwd, but pane `cwd` is optional/nullable and `parsePanes`
  normalises a missing one to `""` — on that legal response a sole-tab teardown closed and skipped
  the rebuild. `ensureWorkspace` now takes a nullable cwd and omits `--cwd`; probe-verified against
  a live 0.8.0 daemon that herdr still creates a usable workspace.
- The rebuild was also left disarmed when the last-tab read itself throws, so a transient
  `tab.list` error could cost a workspace on the fail-open close. Now armed.

**On the merge conflict:** confirmed your diagnosis — union-merge-driver artefact, no semantic
conflict. Rebased (twice since, `main` kept moving); currently 0 behind, no merge commits.

Root suite 8680 pass / 0 fail, CI green on `20e3ccaf`. Every new test was verified to fail without
its fix, so they pin behaviour rather than describe it. `mergeable_state` is `blocked` only on your
outstanding `CHANGES_REQUESTED`.

PR #2057, opened 2026-08-08
feat: recommend the public video-brief skill for video tasks (#2053)

### LINKED ISSUE #2053: Recommend public video-brief skill for video tasks
## Goal

Recommend the public `video-brief` Agent Skill when a user attaches a screen recording in New Task and the selected coding provider is known not to expose that skill. Also give Claude Code and Codex the restricted Shepherd coordinates needed for the skill's optional self-rename behavior.

Public skill: https://github.com/erwins-enkel/skills/tree/main/skills/video-brief

The publication and hardening work is complete in `erwins-enkel/skills` via PRs #2 and #4. Start this Shepherd change from the latest `origin/main` and deliver exactly one cohesive PR.

## Backend scope

- Extend the restricted agent-ingress allowlist with only these exact routes:
  - `GET /api/sessions/:id`
  - `POST /api/sessions/:id/rename`
- Delegate both routes to the canonical session/read and rename handlers so normalization, branch rename, PR retargeting, collision handling, persistence, and live events remain centralized.
- Keep enumeration, deletion, wrong methods, trailing segments, approve gates, and unrelated session operations denied.
- In `SessionService.prepareSpawn()`, add `SHEPHERD_AGENT_API_URL` beside the existing `SHEPHERD_SESSION_ID` in both trusted and membrane environments.
- Use the existing `resolveSpawnBaseUrl()` result so both Claude Code and Codex receive reachable coordinates on create, resume, and provider replacement.
- Keep the provider-neutral control-plane variables out of Claude-only ownership; restore any temporary overlay-only expectations accordingly.

## New Task recommendation

- Recognize completed attachment names ending case-insensitively in `.mov`, `.mp4`, `.webm`, or `.m4v`.
- Load a dedicated skill inventory for the currently selected provider only while a video is attached.
- Track `idle`, `loading`, `ready`, and `failed` explicitly. Show the recommendation only after a successful inventory result proves that no exact `kind: "skill"`, name `video-brief` entry exists for that provider.
- Loading or failed inventory is unknown, not missing: show nothing and never block task submission.
- Re-evaluate when the provider changes, remove the recommendation when the last video is removed, and refresh inventory when the window regains focus so an installation in another terminal is detected.
- Render a compact information row in New Task's existing notices area. Link only to the public GitHub skill page in a new tab with an accessible name and `rel="noopener"`.
- Do not run `npx`, install anything, copy an install command, modify agent configuration, add a modal, or add a dismiss action.
- Use existing semantic design tokens and the `/design-system` recipes; add no literal colors or ad-hoc font sizes.

## Internationalization and discovery

- Add all authored copy to both `ui/messages/en.json` and `ui/messages/de.json`.
- Add one feature-announcement fragment using the current output of `bun run next-version` for both filename and `sinceVersion`.
- No new glossary entry is required because Shepherd already uses the term “Skill”.

## Verification

- Follow TDD: record focused failing tests before implementation and passing output afterward.
- Agent-ingress tests cover exact allows plus enumeration, wrong-method, and trailing-segment denials; integration coverage proves canonical self-read/self-rename behavior.
- Service tests prove both providers receive `SHEPHERD_AGENT_API_URL` and `SHEPHERD_SESSION_ID` on create and resume, plus both provider-replacement directions.
- Browser tests cover known-missing, installed, loading, failed, non-video, last-video removal, provider switching, focus refresh, public URL/new-tab/accessibility, and mobile submit-control reachability.
- Run focused backend/UI tests, full root lint/tests, full UI check/tests, i18n parity, glossary, feature-catalog, announcement-version, and branch-hygiene gates.
- Do not include uploads, recordings, audio, transcripts, contact sheets, or frames in the branch.

kai-osthoff:
Closes #2053.

Two halves of one capability: give a spawned agent the coordinates the public
[`video-brief`](https://github.com/erwins-enkel/skills/tree/main/skills/video-brief) skill needs,
and tell the operator when the selected engine doesn't have that skill.

## Agent coordinates

`SHEPHERD_AGENT_API_URL` joins `SHEPHERD_SESSION_ID` in `prepareSpawn()` — on **both** the plain
spawn env and the membrane's `--setenv` loop (what survives bwrap `--clearenv`) — and the restricted
agent-ingress allowlist gains exactly the two routes those variables address:

- `GET /api/sessions/:id`
- `POST /api/sessions/:id/rename`

Both **delegate to the canonical handlers**, so slugification, the open-PR display-only decision,
the branch move, collision handling, persistence and the `session:renamed` event stay centralized —
no parallel rename path.

The variables are deliberately **not** carried by the Claude `--settings` overlay: a Codex spawn
never reads that object, and the control plane is provider-neutral. The overlay's `env` is unchanged
(still pinned by its existing test). The value comes from the existing `resolveSpawnBaseUrl()`,
called inside `prepareSpawn()` rather than threaded down from create / resume / relaunch /
replaceAgent — that seam is the only one common to all four, and it uses the same inputs as the
hooks fragment, so the URL a skill calls and the URL hooks post to cannot diverge.

### One thing worth a careful look

Adding a bare `GET /api/sessions/:id` to the allowlist has a non-obvious hazard: `/api/sessions/<seg>`
with nothing after it is **also** how the Done lens is addressed. `GET /api/sessions/done` answers
with every recently-archived session across every repo, so a shape-only "no fourth segment ⇒ allow
GET" rule would have handed every spawned agent that enumeration.

The bare-GET rule is therefore gated on UUID shape (`SESSION_ID_RE`), which excludes `done` and any
reserved literal added later, structurally. Session ids are always `randomUUID()` (`create()` and
`createTerminalSession()`), so nothing real is locked out. There is a dedicated test for the `done`
route, and the pre-existing "containment property" test was **edited** — it used to assert the bare
GET was denied — with the narrower containment (enumeration, deletion, wrong methods, trailing
segments, both approve gates) spelled out in its place.

`sessionRead()` returns the whole session row; for the agent's own session that is information it
already holds. The genuinely sensitive read, `/api/tasks/:key/export`, stays off the allowlist.

## New Task recommendation

While a `.mov` / `.mp4` / `.webm` / `.m4v` attachment is present, New Task reads the **selected**
provider's skill inventory and — only when that read *proves* no exact `kind: "skill"` / name
`video-brief` entry exists — shows one compact row in the existing notices area linking to the
public GitHub page.

The asymmetry is the whole design: a successful inventory that lists the skill proves it is there;
a failed or in-flight one proves nothing. So `recommend` is true for exactly one of four explicit
statuses (`idle` / `loading` / `ready` / `failed`), and a dead `/api/commands` shows nothing rather
than a false "you're missing this". Submitting is never blocked.

It is its own read, not a share of `allCommands`: that list is keyed on `commandProvider`, which
flips by slash trigger while the autocomplete menu is open — the wrong question to ask about the
provider the task will actually spawn with. Generation-guarded like `IssueData`; re-read on provider
change and on window focus (so an install in another terminal is picked up); reset when the last
video is removed.

No install, no `npx`, no copied command, no agent-config write, no modal, no dismiss.

## Verification

TDD throughout — failing tests recorded before each implementation step.

- **Ingress**: allow/deny table (both new routes, plus `done`, non-UUID ids, list, spawn, delete,
  wrong methods, trailing segments) and integration tests proving both routes reach the canonical
  handlers through `makeAgentIngressApp`.
- **Spawn env**: both variables reach Claude *and* Codex on create, resume, and both
  provider-replacement directions, on the trusted path and past the membrane's `--clearenv`.
- **UI**: 16 headless unit tests (extension matching incl. `.MOV`, the four statuses, exact
  name/kind, generation guard) + 12 browser tests against a mounted `NewTask` (known-missing,
  installed, loading, failed-and-still-submittable, non-video, last-video removal, provider switch,
  focus refresh, link URL/new-tab/accessible name, no install-or-dismiss affordance, mobile
  submit-control reachability).

Gates run locally: root lint + `bun run test` (8357 pass), UI `check` + `test` (4569 pass), i18n
parity, glossary, feature catalog, announcement versions, and the fallow audit (clean on all 12
changed files). Branch cut from `origin/main`, linear, no merges.

## Self-review notes

Two things I changed after reviewing my own diff:

1. The link's `aria-label` did not contain its visible label ("View skill"), failing WCAG 2.5.3
   (Label in Name) for speech-input users. Both locales now embed the visible text, with a test
   pinning the property.
2. I had added a defensive check so the tip could never name a provider its verdict wasn't read
   for. Writing the test showed it **passes without the guard** — `load()` flips to `loading`
   synchronously in the same flush as the provider change, so the state is unreachable. I removed
   the guard and the vacuous test rather than ship code and an assertion that can't fail; the
   invariant is documented where it actually lives.


kai-osthoff (commented):
# PR review

Decision: comment

No blocking findings identified in the specified diff.

## Verified

- The ingress gate permits the new bare read only for a UUID-shaped session id and GET, while the existing dispatcher sends that request to the canonical session-read handler; `done`, non-UUID ids, wrong methods, and trailing segments remain outside the gate. Verified against `src/server.ts:7695`, `src/server.ts:7707`, and `src/server.ts:7722`, with the read handler's `done` special case at `src/server.ts:2354`.
- The new rename leaf is POST-only and delegates to the existing rename handler, which retains its JSON validation, branch-collision handling, persistence, cache invalidation, and live event behavior. Verified against `src/server.ts:7690`, `src/server.ts:7710`, and `src/server.ts:3002` through `src/server.ts:3080`.
- Both spawn carriers receive the API URL and session id, and the membrane serializes `extraEnv` as `--setenv` entries after `--clearenv`. The Claude overlay remains limited to its pre-existing env key. Verified against `src/service.ts:2618`, `src/service.ts:2656`, `src/service.ts:2698`, `src/sandbox.ts:598`, and `src/service.ts:469` through `src/service.ts:497`.
- The recommendation is driven only by completed attachments, identifies the four requested suffixes case-insensitively, fetches the selected provider's inventory, suppresses loading/failed states, resets after the last video disappears, and refreshes on focus. Verified against `ui/src/lib/components/NewTask.svelte:740` through `ui/src/lib/components/NewTask.svelte:758`, `ui/src/lib/components/NewTask.svelte:866` through `ui/src/lib/components/NewTask.svelte:870`, and `ui/src/lib/components/new-task/video-skill.svelte.ts:19` through `ui/src/lib/components/new-task/video-skill.svelte.ts:83`.
- The notice uses the specified public URL, new-tab/noopener attributes, and localized accessible label; the three New Task keys and two announcement keys exist in both maintained catalogs. Verified against `ui/src/lib/components/new-task/VideoBriefNotice.svelte:10` through `ui/src/lib/components/new-task/VideoBriefNotice.svelte:34`, `ui/messages/en.json:320` through `ui/messages/en.json:322`, `ui/messages/de.json:320` through `ui/messages/de.json:322`, `ui/messages/en.json:3547` through `ui/messages/en.json:3548`, and `ui/messages/de.json:3547` through `ui/messages/de.json:3548`.
- The announcement fragment's filename/version agree and its fields satisfy the catalog type and discovery glob. Verified against `ui/src/lib/feature-announcements/entries/v1.47.0-video-brief.ts:3` through `ui/src/lib/feature-announcements/entries/v1.47.0-video-brief.ts:12` and `ui/src/lib/feature-announcements.ts:25` through `ui/src/lib/feature-announcements.ts:50`.

## Verification limitation

- Per the review request, I did not run builds, linters, or tests. I also did not independently fetch the external public skill; its expected HTTP calls are not repository ground truth.

PR #2061, opened 2026-08-09
feat(forge): stack-safe merge path for GitHub stacked PRs

### LINKED ISSUE #2059: Stack-safe merge path: our merge calls fail against GitHub stacked PRs
## Goal

Shepherd's merge paths must not break against GitHub **stacked pull requests**. This is not hypothetical: stacks went to public preview on 2026-07-30 and are already enabled on our repos — `GET /repos/erwins-enkel/shepherd/stacks` returns `200` today (a `404` is how "not enabled" presents).

Scope is **compatibility only**. Authoring stacks is out of scope here; this issue is about not failing when one shows up.

## What breaks

`GithubForge.merge()` shells out to `gh pr merge <n> [--squash|--merge|--rebase] [--delete-branch]` (`src/forge/github.ts:1619-1625`). GitHub documents that the legacy synchronous merge — `PUT /pulls/{n}/merge`, the `mergePullRequest` GraphQL mutation, and `gh pr merge` on top of them — **cannot merge a stacked PR**. The only path is the async merge API: `PUT /repos/{o}/{r}/pulls/{n}/merge-async`, then poll `GET .../merge-async/{uuid}` until `status` leaves `pending`.

Every consumer of `forge.merge` is therefore exposed:

- `AutoMergeService` (`src/automerge.ts:315`) — would burn `MERGE_ERROR_CAP = 3` (`:39`) and hold with an opaque error.
- `POST /api/prs/merge` → `handlePrMerge` (`src/server.ts:6082`) — the Backlog **Merge** button.
- `POST /api/sessions/:id/merge` (`src/server.ts:3893`).
- Epic child retire (`src/drain.ts:2348`) and epic landing (`tryAutoLandEpic`).

The UI makes this worse than a silent edge case: we **already render** stacked PRs. `PrRow.svelte:202-208` shows a `→{nonDefaultBase}` target-branch chip whose type comment says it "exists solely to surface non-default (stacked) PRs in the backlog PRs tab" (`src/forge/types.ts:182-187`). We display them, then offer a Merge button that fails.

Two related constraints worth encoding so nobody adds a workaround later:

- **`--admin` / rule bypass does not work on stacks.** Every layer must genuinely satisfy its rules.
- **Auto-merge is unsupported on stacks entirely**, per-layer and whole-stack, direct and via queue. This costs us little — `AutoMergeService` is already its own poll-and-merge loop, not a wrapper around GitHub auto-merge — but it closes off "just enable auto-merge" as a recovery path.

## Vertical cut

Detect stack membership on a PR (the `stack` object is already on the REST PR resource: `{ id, number, size, position, base: { ref, sha } }`, `null` when unstacked). Route stacked merges through `merge-async` with the `sha` head guard and uuid polling. Where a stacked PR cannot be safely merged, **refuse with a legible reason** rather than burning retry budget.

GitHub-only. `GiteaForge` and `LocalForge` have no stack concept, so this is one implementation behind the existing `GitForge` interface, not three.

## Acceptance

- A stacked PR merges successfully through `merge-async` + uuid polling, passing `sha` as the optimistic-concurrency guard.
- The merge train does not consume `MERGE_ERROR_CAP` against a stacked PR; it holds with a reason that names stacking.
- The Backlog Merge button on a stacked PR either merges correctly or is disabled with an explanation — never a silent failure.
- `status: "enqueued"` is treated as terminal **for the merge request** and handed off to merge-queue tracking, not polled to completion.
- A `409` (another request already in flight) adopts the returned uuid rather than erroring.
- Gitea and Local forge behaviour is unchanged.
- Tests cover: stack detection on a PR, merge-async happy path, `409` adopt-existing-uuid, and a `failed` terminal status.

## Reference

`docs/research/github-stacked-prs-2026-08-08.md` — §2.2 (merge semantics), §2.3 (automation surface), §4 (what breaks today), §6 Tier 0.

scoop:
Closes #2059

`gh pr merge` cannot merge a GitHub **stacked** pull request — the legacy synchronous merge refuses — so every one of Shepherd's six `forge.merge` call sites was exposed. We already *render* stacked PRs in the Backlog (the `→base` chip), then offered a Merge button that would fail.

Stacks are enabled on this repo (`GET /stacks` → `200`) but **zero stacks exist today**, so nothing is broken right now. This is pre-emptive compatibility work; authoring stacks stays out of scope.

## Verified against the live API

Stacked PRs shipped after my knowledge cutoff, so I probed `api.github.com` directly rather than working from the issue text. Three findings shaped the implementation:

- **The issue's response shape is wrong.** The uuid is not top-level — the body is `{ status, details }` with the uuid at **`details.uuid`**. Code written from the issue alone would never find it.
- **`gh api` writes HTTP error bodies to stdout** (only `gh: HTTP 409` on stderr, exit 1). That is what makes "409 → adopt the in-flight uuid" reachable through the existing `GhRunner` seam at all.
- A `PUT …/merge-async` with a deliberately wrong `sha` returned a real `{"status":"failed","details":{"message":"Pull request head branch was modified."}}`, confirming the optimistic-concurrency guard behaves as documented.

## The organising rule

> **Operator-initiated merges may land a stack. Autonomous merges refuse.**

Merging layer *n* also lands **every unmerged layer below it**, atomically — layers Shepherd never gated on CI or the critic. An autonomous train landing those is a worse hazard than the failure the issue describes. Expressed as one optional `MergeInput.allowStacked`, set `true` at exactly the two operator HTTP endpoints (`POST /api/prs/merge`, `POST /api/sessions/:id/merge`) and nowhere else.

## Design calls worth reviewing

**Typed errors, not a widened return type.** `merge()` resolving keeps meaning "the PR merged"; `stacked_refused` / `merge_enqueued` / `merge_pending` all throw, and all mean the same thing to callers: *don't settle the session, don't count a failure*. Widening `Promise<void>` would have forced edits to ~50 test doubles for no behavioural gain.

**`enqueued` is not success.** Shepherd has no merge-queue tracking, so an enqueued PR must not be settled or archived — polling stops and the PR poller reconciles.

**Branch deletion is skipped on the stacked path.** `merge-async` has no delete-branch parameter, and deleting a merged layer's branch could pull the rug from under a layer still based on it mid-restack. `BranchPruner` only sweeps *local* branches, so the remote branch lingers — a stated trade-off, recorded in a code comment.

**Detection fails open.** A probe error routes the merge down the legacy path, i.e. exactly today's behaviour, so a flaky REST call can never be a regression.

**The uuid is validated before use** — it is interpolated into a `gh api` path, so a garbled or traversal-shaped value is treated as "no uuid to poll" rather than retargeting the request.

## Merge train

Learns a PR is stacked by attempting the merge once and being refused **before anything is mutated**, then remembers it (keyed by head SHA, so a new push re-tests). It never records a merge failure, so `MERGE_ERROR_CAP` is untouched, and reports a named `stacked` hold — produced by `computeMerge`, so it is populated on both the live status event and the bootstrap `snapshot()`. A stacked PR is also excluded from rebase: a stack layer is restacked, never rebased onto main.

## Scope

Gitea and Local forge are untouched (no stack concept). No stack data is added to the PRs-tab payload — the GraphQL `stack` field shape cannot be verified without a real stack, and that is a separate issue.

## Verification

Root `bun run lint` + `bun run test` (8421 pass) · `cd ui && bun run check && bun run test` (4574 pass) · `bunx tsc --noEmit` clean · `fallow audit` exit 0 · EN/DE catalogs at parity · all pre-push gates green.

15 new forge tests cover: unstacked path unchanged, probe-fails-open, refusal-before-mutation, merge-async with the `sha` guard, 409 adopt-existing-uuid, `failed`, `enqueued` stops polling, exhausted budget, transient-poll-retry, no branch deletion, and the uuid guard. Plus train tests asserting the refusal costs no cap budget.


scoop:
# Review — stack-safe merge path (#2059)

Scope: `git diff 1105517d...HEAD` (20 files). Read-only inspection; no builds, no tests run.

## What I verified

- **The unstacked GitHub path still ends in `gh pr merge`.** `merge()` probes, then falls through to
  the untouched argv build (`src/forge/github.ts:1686-1700`); `test/forge/github.test.ts:862` was
  adjusted to `calls.find((c) => c[0] === "pr")` because `calls[0]` is now the probe. With
  `fakeRunner({})` the probe returns `""` and `JSON.parse(out || "{}")`
  (`src/forge/github.ts:1710`) yields `{}` → `stacked: false`, so that test still exercises the
  legacy path.
- **Gitea/Local untouched.** No `.merge(` call sites in `src/forge/gitea.ts` / `src/forge/local.ts`
  appear in the diff, and `MergeInput.allowStacked` is optional (`src/forge/types.ts:420`), so their
  signatures are unaffected.
- **`MergeSessionView.stacked` is required, and every producer sets it**: `src/automerge.ts:193`,
  `test/automerge-core.test.ts:30`, `test/task5-completion-barrier.test.ts:270`. Those are the only
  three construction sites (grep for `MergeSessionView` returns `src/automerge.ts:175/202`,
  `src/pr-conflict.ts:4` (comment only), and the two test files).
- **i18n parity.** `prspanel_merge_in_progress` at `ui/messages/en.json:871` and
  `ui/messages/de.json:871`; `automerge_state_stacked` at `ui/messages/en.json:981` and
  `ui/messages/de.json:981`; `feat_stacked_prs_title`/`_body` at `en.json:3549-3550` and
  `de.json:3549-3550`. `ui/messages/*.json` is exactly {en, de} (glob), so no third catalog is owed.
- **Feature-catalog gate.** The catalog is `import.meta.glob`-driven
  (`ui/src/lib/feature-announcements.ts:47-49`), so no registration edit is needed. `sinceVersion`
  `1.47.0` is above the released `1.46.0` (`package.json:3`), matching the rule at
  `ui/src/lib/feature-announcements.ts:11-16`; the filename `v1.47.0-stacked-prs.ts` agrees, and the
  id `stacked-prs` is unique in `ui/src/lib/feature-announcements/entries/`.
- **The api.ts re-wrap follows the house pattern** — compared against the two existing sites at
  `ui/src/lib/api.ts:1049` and `:1113`; the new one at `:1239` is identical in shape
  (`isPreviewBlocked(base)` escape hatch first, then `serverAuthored(base)` carried through).
  `ApiError` is exported at `ui/src/lib/api.ts:153`, so the `PrRow.svelte:5` import resolves.
- **`mergeTrainLabel("stacked")` resolves**: `ui/src/lib/components/queue-strip.ts:232` calls
  `m.automerge_state_stacked()`, which exists in both catalogs (above).
- **Train cap preservation is real.** `src/automerge.ts:327-335` returns before
  `recordMergeFailure` (`:343`), and `readyToMerge` (`src/automerge-core.ts:161`) then keeps the
  session out of the merge decision, so the forge is asked exactly once — asserted at
  `test/automerge.test.ts:933`.
- **Uuid path-injection is guarded**: `MERGE_UUID_RE` (`src/forge/github.ts:82`) is applied before
  interpolation (`:1799` via `mergeAsyncUuid`), covered by
  `test/forge/github-stacked-merge.test.ts:302-324`.

## Findings

### 1. `src/server.ts`: the stacked merge path can block a request for ~60s, but the merge routes are not in `slowRequestTimeoutSec` — Bun severs the socket at 10s and the UI renders a false "merge failed"

`pollMergeAsync` waits `MERGE_ASYNC_POLL_ATTEMPTS × MERGE_ASYNC_POLL_MS` = 30 × 2 000 ms ≈ 60 s
(`src/forge/github.ts:57-58`, `:1781-1790`), all of it inside the HTTP handler for
`POST /api/prs/merge` (`src/server.ts:6097`) and `POST /api/sessions/:id/merge` →
`forgeMerge` (`src/server.ts:3909`).

This repo already documents exactly what happens next. `src/server.ts:7935-7949`: "Per-route idle
budget (seconds) for the handful of routes that legitimately outrun Bun's 10s default … Bun then
severs the socket while the handler keeps running to completion, so the operator saw a bogus failure
for an approve that had actually succeeded." The allowlist at `src/server.ts:7950-7957` contains
only `/api/usage/refresh`, `epic-draft/approve`, `/api/uploads` and `scratchpad/upload` — neither
merge route is there, and `serve()` deliberately sets no global `idleTimeout`
(`src/server.ts:7937-7938`, `:7970-7976`).

So for any stacked merge that takes longer than ~10 s to settle — which is the normal case for the
*asynchronous* merge API on a multi-layer stack, and is guaranteed on the `MergePendingError` path —
the browser's `fetch` in `mergeBacklogPr` rejects with a transport error rather than a response. In
`PrRow.svelte:114-120` that error is not an `ApiError`, so it takes the `else` branch and sets
`failed = true`, showing `prspanel_merge_failed` for a merge that is landing or has already landed.
That inverts the acceptance criterion this PR was written for ("either merges correctly or is
disabled with an explanation — never a silent failure") and defeats the whole `mergeInFlight`
distinction added at `PrRow.svelte:46,117,263`.

Fix either end: add both merge routes to `slowRequestTimeoutSec`, or cut the in-request poll budget
below the 10 s default and let `MergePendingError` + the PR poller carry the rest.

### 2. `src/forge/github.ts`: a terminal `failed` body delivered on a non-2xx PUT is thrown away, and the raw `gh` error is rethrown instead

`putMergeAsync`'s catch parses the error body and then consults it for *one* thing — a uuid:

```ts
const body = parseMergeAsyncBody((err as { stdout?: unknown }).stdout);
if (!mergeAsyncUuid(body)) throw err;      // src/forge/github.ts:1771-1772
```

The approved plan's own live probe (finding 1) records that a wrong `sha` returns
`{"status":"failed","details":{"message":"Pull request head branch was modified."}}` **at HTTP 400**
— i.e. as a rejection, with no uuid. That is the exact head-moved race the `sha` guard at
`src/forge/github.ts:1767` exists to catch, and it is the most likely real failure on this path
(the head can move between the probe at `:1702` and the PUT). Today it rethrows the `execFile`
rejection built by `defaultRunner` (`src/forge/github.ts:260-276`), whose `message` is
`Command failed: gh api --method PUT …` plus `gh: HTTP 400` — the host's actual reason is discarded
even though `mergeAsyncSettled` (`:1799-1812`) is sitting right there and would render it.

The test suite only covers the 200-with-`failed`-body shape
(`test/forge/github-stacked-merge.test.ts:185-203`), so the observed shape is untested. Applying
`mergeAsyncSettled` to the error body before the uuid check (or, minimally, throwing
`details.message` when `body.status === "failed"`) closes it. Note this message is what reaches the
operator verbatim at `src/server.ts:6119`.

## Nits (non-blocking):

- `src/automerge-core.ts`: the new `"stacked"` hold code is not recognised by the server-side attention paths — `src/hold-service.ts:145` and `src/push.ts:572` still test only `merge_error`/`rebase_cap` — while `ui/src/lib/components/queue-strip.ts:215` does treat it as attention, so the two halves of "needs attention" now disagree for this state.
- `src/automerge.ts`: `stackedHold` is never pruned when a session is archived (`buildState` filters archived sessions at `:205` but nothing deletes the map entry), so it grows for the process lifetime; the neighbouring `mergeFail` map has the same shape, so this is consistent rather than novel.
- `src/forge/github.ts:1767`: the `sha` guard is conditional on `probe.sha` being present, so a probe that succeeds but reports no head SHA silently issues the merge with no optimistic-concurrency guard at all; failing the merge (or re-probing) would match the acceptance criterion more literally than dropping the guard.
- `src/automerge.ts:336-341`: the `MergeNotCompletedError` branch in the train is unreachable today — the train never passes `allowStacked`, so `enqueued`/`pending` cannot arise on that path — cheap insurance, but worth knowing it ships unexercised.

- `src/automerge.ts:332` / `:193`: `stackedHold.set(sessionId, headSha ?? "")` collapses a null head to `""`, and `toView` maps a null `pr.headSha` to `""` as well, so a stacked PR observed with an unknown head is held permanently instead of re-tested on the next push — the "keyed by head so a new push re-tests" property documented at `:96-101` quietly does not hold for that case.
- `src/drain.ts:2358` (`retireEpicChild`): unlike the auto-land path, this catch has no failure cap, so a `StackedMergeRefusedError` there would re-fire `forge.merge` — and its probe — on every drain tick forever. The new comment at `src/drain.ts:2145-2147` reasons about precisely this hazard for auto-land and relies on `LAND_MERGE_ERROR_CAP`; retire has no equivalent brake. Unreachable while Shepherd never authors stacks and epic children target the epic branch.
- `src/forge/github.ts:1764`: `o.method` is interpolated into `merge_method=` verbatim, and it arrives from an unvalidated request body (`src/server.ts:6088` and `:3901` both cast without checking). Not exploitable — `execFile` takes an argv array, and the value can never be reparsed as a flag — but the legacy path normalises the same input through a ternary at `src/forge/github.ts:1690-1691` while this one does not.

## Stated limitations

- Could not verify the `merge-async` request/response contract (`merge_method`/`sha` field names,
  `details.uuid` placement, the `pending|merged|enqueued|failed` status set, the 409-carries-uuid
  behaviour) — it is a GitHub preview API with no ground truth in this repo. Assumed as written; the
  plan documents a live probe, which I could not re-run.
- Could not verify the populated shape of the REST `stack` object for the same reason. The code
  reads only truthiness plus optional `position`/`size` (`src/forge/github.ts:1709-1715`), which is
  the right hedge.
- Could not verify that `gh api` writes HTTP error bodies to stdout; the code and its fake
  (`test/forge/github-stacked-merge.test.ts:24-29`) both assume it. If that assumption is wrong,
  the 409-adopt path silently never fires — but it would fail closed (rethrow), not wrongly.

scoop (commented):
# Review — stack-safe merge path (#2059), rework round 2

## Verdict

No blocking defects found this round. The one prior finding is genuinely fixed, and the rest of the
diff resolves against the tree.

## Prior finding — re-check

**1. Socket budget for the two merge routes.** *Addressed.*

- `src/forge/github.ts:59` now caps the poll at `MERGE_ASYNC_POLL_ATTEMPTS = 15` (was 30), and
  `src/forge/github.ts:68` exports `MERGE_ASYNC_MAX_WAIT_MS = 2_000 * 15 = 30_000`.
- `src/server.ts:7957` derives `MERGE_ROUTE_TIMEOUT_SEC = Math.ceil(30_000 / 1000) + 60 = 90`, and
  `src/server.ts:7964-7965` registers both routes. 90 is under Bun's 255 ceiling documented at
  `src/server.ts:7943`.
- Applied per-request: `src/server.ts:8036-8038` calls `server.timeout(req, slowSec)` before
  `app.fetch(req)`.
- Path shapes verified against the actual routers, not assumed: `/api/prs/merge` matches the guard
  at `src/server.ts:6084` (`parts[1] === "prs" && parts[2] === "merge"`), and the regex
  `^\/api\/sessions\/[^/]+\/git\/merge$` matches the dispatch at `src/server.ts:4154`
  (`parts[3] === "git"`) plus `src/server.ts:4063` (`parts[4] === "merge"`). The UI hits exactly
  those paths — `ui/src/lib/api.ts:1229` and `ui/src/lib/api.ts:1451`.
- Worst case in-handler wait (probe + PUT + 15 × 2s sleeps + 15 `gh api` round-trips) is comfortably
  inside 90s.

## What I verified

**Typed errors resolve everywhere they are used.** `MergeNotCompletedError` (abstract, value-exported
at `src/forge/types.ts:341`), `StackedMergeRefusedError` (`:349`), `MergeEnqueuedError` (`:371`),
`MergePendingError` (`:381`). Value imports: `src/server.ts:154`, `src/automerge.ts:3`,
`src/drain.ts:38`, `src/forge/github.ts:20-26`. All `instanceof` sites branch on classes that are
actually exported as values, and `StackedMergeRefusedError` is checked *before* its base class at
`src/automerge.ts:325` / `:335`, so the stacked branch is not swallowed by the generic one.

**Every `forge.merge` call site accounted for.** Grep gives exactly four: `src/server.ts:3910`
(`allowStacked: true`), `src/server.ts:6098` (`allowStacked: true`), `src/drain.ts:2097` (auto-land,
default) and `src/drain.ts:2358` (epic-child retire, default). This matches the plan's table, and the
two autonomous ones therefore refuse before any mutation.

**Cap preservation.** `src/automerge.ts:325-332` returns without `recordMergeFailure` on a stacked
refusal and stores `stackedHold` keyed by head SHA (`src/automerge.ts:193`), so a new push re-tests.
`src/automerge-core.ts:161` keeps the train from picking it, `:186` keeps it out of the rebase path,
and `:284-300` reports the `stacked` hold ahead of `manual_steps` at `:307`. Because the hold falls
out of `computeMerge`, it appears on both the live emit (`src/automerge.ts:268-277`) and
`snapshot()` (`src/automerge.ts:477-479`) — I traced both to the same `buildState` projection.

**`MergeSessionView.stacked` is required, and every constructor sets it.** Grep for
`MergeSessionView` finds two literal constructions outside the interface: `src/automerge.ts:193`
and `test/task5-completion-barrier.test.ts:270` (both updated), plus the `sess()` helper at
`test/automerge-core.test.ts:30`.

**Merge-async wiring.** The uuid is read from `details.uuid` (`src/forge/github.ts:113-116`) and
validated against `MERGE_UUID_RE` (`:96`) before being interpolated into the `gh api` path at
`:1791` — path-traversal is closed and covered by the last test in
`test/forge/github-stacked-merge.test.ts:293-318`. The 409-adopt path reads the body off
`err.stdout` (`src/forge/github.ts:1781`), which is the shape Node's `execFile` rejection carries
(`src/forge/github.ts:270-273` shows the default runner is `promisify(execFile)`).

**Unstacked path is unchanged apart from the probe.** `src/forge/github.ts:1697-1703` still builds
`["pr","merge",n,"--repo",slug,method]` verbatim; the probe fails open at `:1734-1736`. The updated
assertion at `test/forge/github.test.ts:862` reflects only the extra leading call.

**Constructor arity.** The new 5th positional `sleep` parameter (`src/forge/github.ts:508`) does not
collide with any existing construction — `src/forge/index.ts:45` and `:98` pass at most four args;
all other sites are tests passing three.

**Locale parity — opened both catalogs.** `prspanel_merge_in_progress` at `ui/messages/en.json:871`
and `ui/messages/de.json:871`; `automerge_state_stacked` at `ui/messages/en.json:981` and
`ui/messages/de.json:981`; `feat_stacked_prs_title`/`feat_stacked_prs_body` at
`ui/messages/en.json:3549-3550` and `ui/messages/de.json:3549-3550`. `ui/messages/` contains only
`en.json` and `de.json`, so that is full parity. The announcement entry
`ui/src/lib/feature-announcements/entries/v1.47.0-stacked-prs.ts` references exactly those two keys,
its filename version matches its `sinceVersion`, and `1.47.0 > 1.46.0` (`package.json:3`) as the
gate described at `ui/src/lib/feature-announcements.ts:10-16` requires.

**UI error re-wrap.** `ui/src/lib/api.ts:1236-1239` follows the documented pattern verbatim — same
three lines as the existing sites at `ui/src/lib/api.ts:1048-1049` and `:1112-1113`, including the
`isPreviewBlocked` escape hatch so a 403 does not lose its `PreviewBlockedError` class. `ApiError` is
value-exported at `ui/src/lib/api.ts:153`, which is what `ui/src/lib/components/PrRow.svelte:5`
imports, and its `code` field (`:155`) is the one `PrRow.svelte:116` branches on. `rebase-note` is an
existing class in the same component (used for `prspanel_rebase_requested`).

## Stated limitations

- I could not verify GitHub's `merge-async` contract against a live schema — the endpoint, the
  `merge_method`/`sha` parameter names, the `{status, details:{uuid,message}}` body and the
  `pending|merged|enqueued|failed` status set are not represented anywhere in this repo. I assumed
  them as written and reviewed only the code's internal consistency around them.
- Likewise the populated shape of the REST `stack` object. The code reads only truthiness plus
  optional `position`/`size` for a message, so an inference error there is contained.

- `src/drain.ts:2358` `retireEpicChild` catches every merge error and retries on the next tick with no cap of its own, so any durable merge failure loops indefinitely (pre-dates this PR; the new refusal just joins that class).

## Nits (non-blocking):

- `src/forge/github.ts:1780-1783`: the catch already parses the HTTP error body but only looks for a uuid — a body carrying `{status:"failed", details:{message}}` at a non-2xx status (per the plan's own live probe, that is how a stale `sha` presents) is discarded and the raw `gh: HTTP 400` is rethrown, so the host's reason never reaches the log or the 502 body. Feeding that parsed body through `mergeAsyncSettled` for the `failed` case would cost two lines.
- `src/forge/github.ts:1793-1806`: poll errors are swallowed uniformly, so an adopted-but-bogus uuid (a 404 on every poll) burns the whole 30s budget and then reports `MergePendingError` — "still merging" for something that never started. Only reachable through the adopt path; a 404 could reasonably be treated as terminal.
- `src/server.ts:7936-7957`: the `MERGE_ROUTE_TIMEOUT_SEC` const was inserted *between* the `/** Per-route idle budget … */` doc comment and `slowRequestTimeoutSec`, so that doc block now documents the constant and the function it describes has none.
- `ui/src/lib/components/queue-strip.ts:215` treats `"stacked"` as an attention state, but the server-side attention sets (`src/hold-service.ts:145`, `src/index.ts:2152`) and the `merge_attention` push (`src/push.ts:572`) still list only `merge_error`/`rebase_cap` — a stacked hold shows in the queue strip but raises no push and clears the Herd Rundown flag. Consistent with the plan's scope, just asymmetric.
- `src/automerge.ts:19`: the `AutoMergeStatus.state` doc comment still enumerates `"merging" | "rebasing" | "merge_error" | "rebase_cap"`; the mirror at `ui/src/lib/types.ts:837` was updated to include `"stacked"`.
- `src/automerge.ts:103`: `stackedHold` entries are only removed on a successful merge, so an archived session leaves a permanent map entry. Bounded and in-memory, but `mergeFail` has the same shape, so it is not a new leak class.
- `src/automerge-core.ts:186`: once a stacked PR also goes `behind`, `rebaseEligible` is false and `readyExceptManualSteps` is false, so `computeMerge` reports plain `idle` rather than the `stacked` hold — the operator-visible reason disappears exactly when the PR is most stuck. `test/automerge-core.test.ts:503-510` locks this in deliberately; worth a follow-up rather than a change here.

- `src/automerge.ts:335-340`: the generic `MergeNotCompletedError` branch in `doMerge` is unreachable today — the train never passes `allowStacked`, so `MergeEnqueuedError`/`MergePendingError` cannot originate from it. Defensive, cheap, and covered by a test; fine to keep.
- `ui/src/lib/feature-announcements/entries/v1.47.0-stacked-prs.ts` is not mentioned in the plan's file table, but `ui/src/lib/feature-announcements.ts:3-9` makes it a mandatory part of shipping any user-facing change, so it belongs.

Issue #2060, opened 2026-08-09
Epic landing PR vs stacked PRs: settle atomicity, then stack children onto the integration branch


scoop:
## Goal

Decide, then implement, how GitHub stacked pull requests change the **epic landing PR**. This issue carries a blocking spike first, because the spike's answer determines which of two designs is legal.

Handoff from `docs/research/github-stacked-prs-2026-08-08.md` §5.2, §5.2.1, §7.

## Why the landing PR is the target

The aggregate landing PR is the most machinery-dense object in the epic model. Around it sit `landingState`, `landingAttempts` (`MAX_LANDING_ATTEMPTS = 5`, `src/drain.ts:100`), `landingRebasePauseReason` (`cap|conflict|driver`), `landingRepairCount`/`landingRepairHead`, `landingStranded`, `landingCiFailing`, `landingRepairing`, a pre-warm draft PR, a per-head merge backoff, a fail-closed `tryAutoLandEpic` racing the manual land endpoint, plus `scanDivergentEpicBranches` and the `epic_base_mismatch` sweep guarding its base.

Nearly all of that is failure handling for **one structural cause: drift**. The integration branch is long-lived and accumulates children one squash-merge at a time over days while `main` moves underneath it. Nothing rebases it — the critic is handed the base delta in prose precisely because "this branch was never rebased onto it." The landing PR is where that accumulated drift finally has to be reconciled, at the very end, when the epic is nominally done.

Stacks remove the drift by construction: GitHub auto-rebases the remainder when a lower layer merges, and there is an explicit restack when the trunk moves. Reconciliation is paid continuously in small increments instead of as one lump against a months-old branch.

## Step 1 — blocking spike: settle partial-merge atomicity

**The docs contradict themselves and the contradiction decides the design.**

- The merging page: the selected PR and all unmerged PRs below it "land on the base branch together as a single operation."
- The troubleshooting page: "Merges stop at the failed pull request. Successfully merged PRs land on the base branch; the failed PR and those above remain open."

Both cannot hold for the same path. If a mid-stack failure can leave **half an epic on `main`**, that violates exactly the invariant `resolveSpawnBase` fails closed to protect today (`src/drain.ts:2492-2493`: "Never silently base an epic child on the default branch … that child would land on main mid-epic").

[...]

PR #2049, opened 2026-08-08
docs(research): what Shepherd could do with GitHub stacked PRs

scoop:
Research note answering **"what could we do with GitHub's new PR stacking feature?"**

GitHub put stacked pull requests into public preview on **2026-07-30**. Report file is the entire diff — no code changes.

## Headlines

**It is already enabled on this repo.** `GET /repos/erwins-enkel/shepherd/stacks` returns **200** (a `404` is how "not enabled" presents). Full REST API for create/append/unstack, read-only GraphQL, a new `pull_request.stacked` webhook action, and an official `gh-stack` SKILL.md installable straight into Claude Code.

**There is a compat gap that exists today, independent of anything we build.** `GithubForge.merge()` is `gh pr merge --squash`, and GitHub documents that the legacy synchronous merge **cannot merge a stacked PR**. `--admin` bypass also doesn't work on stacks, and auto-merge is unsupported entirely. So the merge train and the Backlog "Merge" button both fail against any stacked PR — while `PrRow.svelte` already renders stacked PRs via the `→base` chip, inviting exactly the click that breaks.

**Our epic model is a hand-rolled stack whose main cost is serialization.** `selectEpicCandidates` gates a child on `integrationMerged || issueClosed`, so dependent children run strictly one at a time. Stacks replace that *wait* with a *base pointer*. Constraint: stacks are strictly **linear**, so this fits a dependency chain, not the full epic DAG.

**Flagship idea: build queue → stack layers.** The build queue is already an ordered, agent-authored, operator-editable plan with a per-step lifecycle — the shape of a stack that currently collapses into one big PR. Making approved steps into layers turns the plan the operator already approved into the review unit.

## Ranked recommendation

| Tier | Item |
| --- | --- |
| 0 | Stack-safe merge path (`merge-async` + uuid polling behind `forge.merge`) |
| 1 | Real stack graph in the PRs tab (zero extra API cost — `baseRefName` is already on the wire) |
| 2 | Epic dependency chains as stacks |
| 3 | Build queue steps → stack layers |

Tier 0 is the one that plausibly deserves its own issue immediately. Recommended **against** for now: replacing the `epic/` integration branch outright (§7 explains why).

Also flagged: `gh stack` opens a full-screen TUI under a PTY, which would wedge unattended agents in herdr panes the same way upsell dialogs do — mandatory non-interactive flags if agents ever drive it.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2065, opened 2026-08-09
fix(forge): don't send merge_method into a required merge queue

### LINKED ISSUE #2062: fix(forge): stacked merge sends merge_method, which a required merge queue rejects
## What

`putMergeAsync` (`src/forge/github.ts`, shipped in #2061) always sends `merge_method` and never sends `merge_action`. On a repo whose target branch **requires a merge queue**, that combination is accepted at request time and then fails asynchronously.

Observed live during the #2060 spike (`docs/research/stacked-pr-partial-merge-spike-2026-08-09.md`, secondary finding 1):

```
$ gh api --method PUT repos/$R/pulls/6/merge-async -f merge_method=squash -f sha=$SHA
{"status":"pending","details":{"…","merge_method":"squash","merge_action":"default", …}}

poll1: {"status":"failed","details":{"message":"Custom merge params are not supported
        when merging via a merge queue"}}
```

An **explicit** `merge_action=merge_queue` fails loudly and immediately instead:

```
{"message":"Custom merge params (merge_method, commit_title, commit_message) are not supported
  with the merge_queue merge action.","status":"422"}
```

## Why it matters

The `PUT` succeeds, so nothing is caught at request time. The poll returns `status: "failed"`, which `mergeAsyncSettled` turns into a generic `Error`. The operator sees:

> `merge of #N failed: Custom merge params are not supported when merging via a merge queue`

— a message about request parameters, for a merge whose parameters they never chose. The merge never had a chance to land, and nothing in the message says so.

Shepherd's own repo does not require a merge queue, so this is **latent here and live for any consumer repo that does**.

## Shape of a fix

Detect that the target branch requires a merge queue and drop `merge_method` for that call (the queue's own configured merge method governs), or send `merge_action` explicitly so the failure arrives as a `422` at request time rather than as an opaque poll failure. Either way the resulting state is a queue entry, not a landed PR — which is what `MergeEnqueuedError` already exists to express.

## Acceptance

- A stacked merge against a merge-queue-required branch either enqueues cleanly (surfacing as `MergeEnqueuedError`) or refuses with a reason that names the merge queue.
- No call sends `merge_method` together with an effective `merge_queue` action.
- Test coverage for the merge-queue-required path.

Found by: #2060 spike.

scoop:
Closes #2062.

## The bug

`putMergeAsync` always sent `merge_method` and never `merge_action`. Under a branch that requires a merge queue, GitHub **accepts** that PUT and then fails it asynchronously:

```
poll1: {"status":"failed","details":{"message":"Custom merge params are not supported
        when merging via a merge queue"}}
```

`mergeAsyncSettled` turned that into a generic `Error`, so the operator saw a complaint about request parameters they never chose, for a merge that never had a chance to land.

## The fix

Probe the stack's **trunk** for a `merge_queue` rule; when one applies, send `merge_action=merge_queue` and no merge params at all (the queue's configured method governs). The request then settles as `enqueued`, which `MergeEnqueuedError` already expresses and drain/the merge train already handle.

The load-bearing detail — and what an earlier draft of this fix got wrong — is *which* branch to probe. Merging any layer lands the whole stack on the trunk, so the trunk carries the rule. Top-level `.base.ref` is the layer **below** this PR, so probing it would have missed the queue for every PR at stack position ≥ 2 — precisely the case in the issue's own repro. `.stack.base.ref` is the trunk, and `probeStack()` already parses the `stack` object wholesale, so it costs no extra request and no `--jq` change.

`merge_action` is sent explicitly rather than left to GitHub's `default`: if the probe were ever wrong about a branch, that fails loudly as a 422 at request time instead of merging under a method nobody chose.

## Failure posture

The probe **fails open** — a probe error, an absent `stack.base.ref`, or a ref failing the path guard all keep today's request shape, mirroring `probeStack()`'s existing contract that a flaky REST call can never be worse than not asking. It deliberately does *not* fall back to the PR's own base: probing the wrong branch answers the wrong question.

The trunk is interpolated into a `gh api` path, so it is shape-checked per slash-separated segment first (each segment must start with a word character, which rules out `..` traversal), the same posture `MERGE_UUID_RE` already applies to the merge-request uuid.

## Verification

Five new cases in `test/forge/github-stacked-merge.test.ts`. Note the harness matches routes by substring, so a `rules/branches/` stub answers for *any* branch — the queue test therefore asserts the **exact** path probed (`repos/o/r/rules/branches/main`) and that no call anywhere mentions the intermediate layer, with the fixture reshaped into a realistic mid-stack PR so the two cases leave different evidence.

- merge-queue trunk → exact trunk probed, PUT fields are exactly `merge_action=merge_queue` + `sha`, surfaces as `MergeEnqueuedError`
- trunk with other rules only → `merge_method` unchanged from today
- rules probe throws → falls open, merge still proceeds
- no trunk on the stack object → no rules call at all, never the layer below
- malformed trunk ref → never interpolated into an API path

Root suite (8426 pass), `lint`, `tsc --noEmit` and the Fallow delta audit green. No UI or i18n surface: the outcome is the existing `merge_enqueued` state shipped in #2061.

Found by the #2060 spike.

### DIFF
--- a/src/forge/github.ts
+++ b/src/forge/github.ts
@@ -86,6 +86,21 @@ interface StackProbe {
   size?: number;
   /** Head SHA, passed to merge-async as the optimistic-concurrency guard. */
   sha?: string;
+  /** The stack's TRUNK — `stack.base.ref`, the branch the whole stack lands on. NOT the PR's
+   *  own `base.ref`, which for any layer above the bottom is the layer below it. Merging any
+   *  layer lands the stack on the trunk, so the trunk is what carries the rules that apply. */
+  stackBase?: string;
+}
+
+/** `stackBase` is interpolated straight into a `gh api` path, so — like {@link MERGE_UUID_RE} — it
+ *  is shape-checked first: a garbled or hostile ref would otherwise retarget the request at a
+ *  different endpoint. Checked per SLASH-SEPARATED SEGMENT rather than against the whole string,
+ *  because it is a `..` segment specifically that traverses; each segment must start with a word
+ *  character, which rules out `..`, `.`, and an empty segment in one move. Deliberately narrower
+ *  than git's ref grammar — a ref this rejects is treated as "we don't know the trunk", which
+ *  fails open to today's behaviour. */
+function isPathSafeRef(ref: string): boolean {
+  return ref.length <= 255 && ref.split("/").every((seg) => /^\w[\w.-]*$/.test(seg));
 }
 
 /** A merge-request uuid is interpolated straight into the `gh api` path, so it is validated
@@ -1703,7 +1718,9 @@ export class GithubForge implements GitForge {
     await this.run(args);
   }
 
-  /** Stack membership + head SHA in one REST call, narrowed with `--jq` so the payload stays small.
+  /** Stack membership, head SHA and the stack's trunk in one REST call, narrowed with `--jq` so
+   *  the payload stays small. The trunk rides along for free: `.stack` is taken wholesale and
+   *  GitHub puts `base: { ref, sha }` on it.
    *
    *  FAILS OPEN. A probe error (rate limit, transient 5xx) reports "unstacked", which routes the
    *  merge down the legacy path — i.e. precisely today's behaviour, so a probe failure can never
@@ -1718,7 +1735,7 @@ export class GithubForge implements GitForge {
         "{stack: .stack, sha: .head.sha}",
       ]);
       const p = JSON.parse(out || "{}") as {
-        stack?: { position?: number; size?: number } | null;
+        stack?: { position?: number; size?: number; base?: { ref?: string } | null } | null;
         sha?: string | null;
       };
       return {
@@ -1726,6 +1743,7 @@ export class GithubForge implements GitForge {
         position: p.stack?.position,
         size: p.stack?.size,
         sha: p.sha ?? undefined,
+        stackBase: p.stack?.base?.ref ?? undefined,
       };
     } catch (err) {
       console.warn(`[github] stack probe failed for pr#${prNumber}; assuming unstacked:`, err);
@@ -1757,21 +1775,56 @@ export class GithubForge implements GitForge {
     await this.pollMergeAsync(prNumber, uuid);
   }
 
+  /** Does the stack's trunk require a merge queue? A `merge_queue` rule there makes
+   *  `merge_method` illegal on the merge request (#2062), so the answer decides the PUT's shape.
+   *
+   *  FAILS OPEN, for the same reason {@link probeStack} does: an unknown answer keeps today's
+   *  request shape, so a flaky rules call can never be worse than not asking at all. An unknown
+   *  trunk is NOT substituted with the PR's own base — for a layer above the bottom that is the
+   *  layer below, and asking about it would answer a different question than the one that
+   *  matters. */
+  private async requiresMergeQueue(stackBase: string | undefined): Promise<boolean> {
+    if (!stackBase) return false;
+    if (!isPathSafeRef(stackBase)) {
+      console.warn(`[github] stack base ${JSON.stringify(stackBase)} is not a probeable ref`);
+      return false;
+    }
+    try {
+      const out = await this.run([
+        "api",
+        `repos/${this.slug}/rules/branches/${stackBase}`,
+        "--jq",
+        '[.[] | select(.type == "merge_queue")] | length',
+      ]);
+      return Number(out.trim()) > 0;
+    } catch (err) {
+      console.warn(`[github] merge-queue probe failed for ${stackBase}; assuming none:`, err);
+      return false;
+    }
+  }
+
   /** PUT the merge request. On a 409 ("a merge request is already in flight for this PR") the
    *  body still carries that request's uuid — adopt it so a retry JOINS the in-flight merge
-   *  instead of racing or erroring. Any other failure with no uuid to adopt is re-thrown. */
+   *  instead of racing or erroring. Any other failure with no uuid to adopt is re-thrown.
+   *
+   *  Under a required merge queue the host rejects `merge_method` outright, so the queue case
+   *  sends an explicit `merge_action=merge_queue` and no merge params at all — the queue's own
+   *  configured method governs. Explicit rather than implicit: were the probe ever wrong about
+   *  the trunk, this fails as a 422 at request time instead of merging under a method nobody
+   *  chose. Both shapes keep the `sha` guard. */
   private async putMergeAsync(
     prNumber: number,
     o: MergeInput,
     probe: StackProbe,
   ): Promise<MergeAsyncBody | null> {
+    const queued = await this.requiresMergeQueue(probe.stackBase);
     const args = [
       "api",
       "--method",
       "PUT",
       `repos/${this.slug}/pulls/${prNumber}/merge-async`,
       "-f",
-      `merge_method=${o.method}`,
+      queued ? "merge_action=merge_queue" : `merge_method=${o.method}`,
     ];
     // Optimistic-concurrency guard: the host rejects the merge if the head moved under us.
     if (probe.sha) args.push("-f", `sha=${probe.sha}`);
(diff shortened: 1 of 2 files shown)


scoop (commented):
# Review — fix(forge): don't send `merge_method` into a required merge queue (#2062)

Diff reviewed: `git diff 3430058e...HEAD` (merge base is `46567a1a`, so the diff is exactly the single
commit `2e4c0992`, touching `src/forge/github.ts` and `test/forge/github-stacked-merge.test.ts`).

**Decision: comment.** I found no correctness or security defect. Everything below is non-blocking.

**The request shape actually changes, and only in the queue case.**
`putMergeAsync` now resolves `queued` before building the argv and swaps a single `-f` value
(`src/forge/github.ts:1820-1828`): `merge_action=merge_queue` when queued, `merge_method=${o.method}`
otherwise. The `sha` guard is appended unconditionally afterwards (`src/forge/github.ts:1829-1830`), so
neither shape loses optimistic concurrency. Nothing else in the file builds a `merge-async` PUT — grep
for `merge-async` finds only `putMergeAsync` and `pollMergeAsync` (`src/forge/github.ts:1795` is the new
rules path, `:1845` is the poll path), so there is no second call site still sending `merge_method`
alongside an effective queue action. That is the issue's second acceptance bullet.

**The shape it sends is the one the live spike confirmed.**
`docs/research/stacked-pr-partial-merge-spike-2026-08-09.md:150-160`:
`-f merge_method=squash -f merge_action=merge_queue -f sha=$SHA` → `422` "Custom merge params
(merge_method, commit_title, commit_message) are not supported with the merge_queue merge action";
retried *without* `merge_method` → `{"status":"pending", ..., "merge_action":"merge_queue"}`. The diff
sends exactly the accepted pair, `sha` included, which the capture shows was present on the accepted
request too.

**The `--jq` filter matches the endpoint's real response shape.** `rules/branches/{branch}` returns a
flat array of objects carrying `type` — captured twice in the spike:
`docs/research/stacked-pr-partial-merge-spike-2026-08-09.md:102`
(`[{"ctx":[…],"ruleset":20606201,"type":"required_status_checks"}]`) and `:187`
(`[{"ruleset":20606201,"type":"required_status_checks"}]`, taken right after the `merge-queue-main`
ruleset was deleted). So `[.[] | select(.type == "merge_queue")] | length`
(`src/forge/github.ts:1796-1797`) filters a field that exists. `Number(out.trim()) > 0` reads `"0"` and
`""` (the empty stdout an unmatched test route and a silent `gh` both produce) as false — `Number("")`
is `0`, not `NaN`, and `NaN > 0` would be false anyway, so both degrade to "no queue".

**The probe targets the trunk, not the direct parent.** `probeStack` widens the parsed type to
`stack?.base?.ref` and surfaces it as `stackBase` (`src/forge/github.ts:1738`, `:1746`); `--jq` was
already `{stack: .stack, …}` (`src/forge/github.ts:1736`), i.e. the whole stack object, so no second
request and no `--jq` change. `requiresMergeQueue(probe.stackBase)` is the only consumer
(`src/forge/github.ts:1820`). Grep for `StackProbe|probeStack|stackBase` returns hits only inside
`src/forge/github.ts` — the interface is module-private, so widening it breaks no out-of-diff consumer.

**Fail-open is genuinely fail-open.** Missing trunk → `false` before any call
(`src/forge/github.ts:1787`); guard rejection → `false` (`:1788-1791`); any throw → `false`
(`:1800-1802`). Each lands on the pre-#2062 argv, so a flaky rules call cannot be worse than not asking.

**The path guard holds.** `isPathSafeRef` (`src/forge/github.ts:102-104`) splits on `/` and requires
every segment to match `/^\w[\w.-]*$/`. That rejects `..`, `.`, and empty segments (first char must be
`\w`), and rejects `%`, `?`, `#`, `:` and whitespace outright — so `%2e%2e` and query-string smuggling
are both out. `a..b` survives but is not traversal in a URL path. The regex is linear (no nested
quantifier), so no ReDoS. The value reaches `gh` as one element of an argv array via `execFile`
(`src/forge/github.ts:285-289`), so there is no shell layer to escape either. The malformed-ref test
(`test/forge/github-stacked-merge.test.ts:422-448`) asserts the negative directly.

**Downstream contract.** `mergeAsyncSettled` maps `"enqueued"` to `MergeEnqueuedError`
(`src/forge/github.ts:1866-1868`), whose `code` is `"merge_enqueued"` (`src/forge/types.ts:369`) and
which extends `MergeNotCompletedError` (`src/forge/types.ts:340-343`, whose contract is "do NOT
settle/archive the session"). The queue test's assertions on both the class and the code string
(`test/forge/github-stacked-merge.test.ts:372-373`) match that ground truth.

**Latency budget still covers the extra round-trip.** The merge routes budget
`ceil(MERGE_ASYNC_MAX_WAIT_MS / 1000) + 60` (`src/server.ts:7956`), and the comment above it
(`src/server.ts:7951-7955`) says the `+60` exists precisely for "a few `gh` round-trips either side".
One more REST call fits; no constant needed updating.

**No existing test is invalidated by the extra call.** The harness returns `""` for unmatched routes
(`test/forge/github-stacked-merge.test.ts:45`), which `requiresMergeQueue` reads as "no queue", and the
only call-count assertion in the file is scoped to `merge-async`
(`test/forge/github-stacked-merge.test.ts:346`) — the rules path does not contain that substring. Route
matching is first-substring-wins (`test/forge/github-stacked-merge.test.ts:36-46`); I checked each new
test's route list for cross-matching (`PROBE` carries a `--jq` suffix, `POLL_PATH` carries the uuid) and
none of the new stubs shadow another. `test/server-prs.test.ts:143-144` stubs the forge, not the runner,
so it is unaffected.

## Limitations (could not verify — not asserted either way)

- **`stack.base.ref` exists on GitHub's stack object.** The note the plan cites for this
  (`docs/research/github-stacked-prs-2026-08-08.md`, PR #2049) is *not* in this branch's history — the
  merge base is `46567a1a` and a glob over `docs/**/*stack*` finds only the spike note. Nothing in this
  repo dumps a `stack` object, so the field's existence is taken as written. It degrades to fail-open if
  wrong, which is today's behaviour.
- **Whether `rules/branches/{branch}` filters out rules the caller can bypass, and whether it surfaces a
  merge queue configured through classic branch protection rather than a ruleset.** Not answerable from
  this repo. The plan names the first of these as an accepted risk; the second would simply mean
  fail-open (today's behaviour) on such a repo.
- **Whether GitHub's route accepts an unencoded `/` inside `{branch}` for this endpoint.** Not
  answerable here. See the nit below.
- **What a *successful* queue-bound merge request settles as.** Leg B's accepted PUT returned
  `{"status":"pending", uuid}` and then failed at the status gate
  (`docs/research/stacked-pr-partial-merge-spike-2026-08-09.md:158-163`); the spike never reached a clean
  enqueue ("Merge queue was tested only to the point of rejection",
  `docs/research/stacked-pr-partial-merge-spike-2026-08-09.md:331-332`). See the nit below.

## Nits (non-blocking):

- `src/forge/github.ts:1795`: `stackBase` is interpolated unencoded and the guard deliberately permits `/`, so a trunk like `release/2.0` yields `repos/o/r/rules/branches/release/2.0`; if GitHub's route does not accept an unencoded slash there, queue detection silently never fires for slashed default branches (fail-open, so no regression) — encoding each segment would remove the question.
- `src/forge/github.ts:1792-1798`: the call has no `--paginate` and no `per_page`, so it reads only the endpoint's first page; a branch with more applicable rules than one page could hide the `merge_queue` entry. Fails open, and 30+ rules on one branch is far-fetched, but the ceiling is invisible from the code.
- `test/forge/github-stacked-merge.test.ts:358`: the queue test stubs the PUT as `{status: "enqueued"}`, a shape the spike never observed on the queue path — leg B's accepted queue-shaped PUT returned `pending` with a uuid (`docs/research/stacked-pr-partial-merge-spike-2026-08-09.md:158-160`). `mergeAsyncSettled` handles `enqueued` from either the PUT or a poll, so the code is fine; the test just proves less about the live queue path than its name suggests. A companion case where the PUT returns `pending` and the *poll* returns `enqueued` would close that gap.
- `test/forge/github-stacked-merge.test.ts:369`: `expect(joined().some((c) => c.includes(PARENT))).toBe(false)` is true by construction today — `probeStack`'s parsed type has no top-level `base`, so `b2-fail` can only ever reach an argv through a future regression. Worth keeping as a tripwire; just not evidence about today's code.
- `src/forge/github.ts:1806`: the `putMergeAsync` doc block still opens "PUT the merge request", but the method now also performs a network probe of its own before building the request. Not wrong, just no longer the whole story.

No scope-creep items and no latent/future-reachable defects to report: `isPathSafeRef` is the guard the
plan called for and is used on the only path that interpolates a ref, and the diff touches nothing the
plan put out of scope (the unstacked `gh pr merge` path, `MergeEnqueuedError` semantics, and the UI's
`merge_enqueued` rendering are all untouched).

Issue #2066, opened 2026-08-09
Epic chains: deterministic chain decomposition + stack-aware candidate selection (#2063 P1)


scoop:
## Goal

Add a deterministic, cycle-safe **chain decomposition** of an epic's dependency DAG, and let
`selectEpicCandidates` use it to admit a child whose only outstanding blocker is its chain
predecessor.

Pure work only. Nothing consumes the new argument yet, so **this issue ships zero behaviour change**.

Part of #2063 (Step 2 of #2060, shape (b)).

## Why a chain decomposition

A GitHub stack is strictly linear; `EpicChild.blockedBy` is a per-child *list*, and `assembleEpic`
(`src/epic-model.ts`) filters only self-loops and out-of-epic edges — so fan-in and fan-out are
first-class. A chain decomposition covers **nodes, not edges**: with `C.blockedBy = [A, B]`, `C` can
sit on top of at most one of `A` or `B`, and the other edge stays a wait.

- **Within-chain** edges become base pointers (later issues).
- **Cross-chain** edges keep today's `blockedBy.every(b => done.has(b))` gate, which stays
  authoritative.
- A child is spawnable early only when its chain predecessor is its **sole** outstanding blocker.

## Scope

New `src/epic-chains.ts`:

- `decomposeEpicChains(children)` — greedy path cover over the `blockedBy` DAG. Every child lands in
  exactly one chain with at most one chain predecessor.
- **Determinism is a requirement, not a nicety.** The cut is recomputed every drain tick; if it can
  flip between ticks, a child's base would move under it. Key the traversal on `EpicChild.order`,
  then `number`, so the same graph always yields the same cut across ticks and restarts.
- **Cycle-safe.** `assembleEpic` filters self-loops and out-of-epic edges but does *not* reject
  cycles, so a cyclic `blockedBy` must never produce a cyclic chain (that would be a base pointing
  at itself transitively). Skip any edge that would close a cycle.
- A child whose chain predecessor is already done-in-epic is **not** stacked — it bases on the
  integration branch and takes today's path.

In `src/epic-core.ts`:

- `selectEpicCandidates(children, stack?)` gains an **optional** second argument. Absent ⇒ output
  byte-identical to today, so every existing caller (including `src/up-next-core.ts`) is untouched.
  Present ⇒ additionally admit a child whose blockers are all done **except exactly one**, where
  that one is its chain predecessor and the caller has flagged that predecessor stack-ready.
[...]

Issue #2063, opened 2026-08-09
Stack epic children onto the integration branch (#2060 Step 2, shape (b))


scoop:
## Goal

Root a GitHub stack at the epic integration branch `epic/<#>-<slug>` and base each epic child on its **chain predecessor's branch** instead of on the integration branch directly. Within-chain dependency edges convert from *wait for the blocker to merge* into *base on the blocker's branch*; cross-chain edges keep today's gate.

This is Step 2 of #2060. Step 1 (the partial-merge atomicity spike) landed in `docs/research/stacked-pr-partial-merge-spike-2026-08-09.md`.

## Settled before implementation

**Children still squash-merge into the integration branch at retire.** The stack supplies the *base*, not the merge timing.

#2060's text proposed children landing "into the integration branch in one bulk operation at the end". That deadlocks: `selectEpicCandidates` (`src/epic-core.ts`) gates a child on `blockedBy.every(b => done.has(b))`, where done is `integrationMerged || issueClosed`. If a chain only merges at the end, every **cross-chain** blocker in it stays not-done, so children in other chains never spawn. Keeping the per-child squash-merge at retire (`retireEpicChild`, `src/drain.ts`) preserves `integrationMerged` as the cross-chain signal while the stack removes the within-chain wait.

## Constraint

**Stacks augment `selectEpicCandidates`; they do not replace it.** `EpicChild.blockedBy` is a per-child *list* and `assembleEpic` (`src/epic-model.ts`) filters only self-loops and out-of-epic edges, so fan-in and fan-out are first-class. A stack is strictly linear, and a chain decomposition covers **nodes, not edges** — with `C.blockedBy = [A, B]`, `C` can sit on top of at most one of `A` or `B`; the other edge stays a wait.

- **Within-chain** edges → base pointer.
- **Cross-chain** edges → today's `blockedBy.every(b => done.has(b))` gate, unchanged and authoritative.
- A child is spawnable early only when its chain predecessor is its **sole** outstanding blocker.

Neither the chain cut nor a maximum path cover exists in the code today — that is part of the work. The payoff scales with how much of the edge set the cut absorbs: near-total for mostly-linear epics, little for wide fan-in ones.

## Design input from the spike

Two findings bear directly on whatever decides "this stack is ready to land":

[...]

PR #322, opened 2026-06-04
fix(ui): keep top bar from overflowing on touch-desktop

scoop:
## Problem

On the **unfolded foldable** (touch device in desktop layout — narrower than a real desktop) the top bar's right-side cluster overflows the bar:

1. **Learnings + update badge** → the `UPDATE` badge is clipped off the right edge.
2. **Learnings + no update** → the full numeric clock (`13:25:18`) itself runs off the edge.

The existing mitigation only dropped the clock when an *update* badge was present, so neither case above was fully covered.

## Fix

`TopBar.svelte` only — no new strings, no API changes.

- **Drop the numeric clock** (keep the connection dot) whenever *any* right-side badge is present on touch-desktop — update, herdr-update, learnings, over-budget, needs-you, or what's-new. The clock is the most expendable element (the system status bar already shows the time). Fixes case 2.
- **Compact the LEARNINGS / NEEDS YOU labels** to their existing icon+count form when the wide pulsing update badge rides alongside them — clock-drop alone isn't enough to fit the update badge. Fixes case 1.

Both behaviours are gated on `touch && !mobile` and reuse the existing phone-layout compact styles; desktop (fine pointer) and phone layouts are unchanged.

## Verification

- `cd ui && bun run check` → 0 errors / 0 warnings
- `cd ui && bun run test` → 427 passed
- `cd ui && bun run check:i18n` → 2 locales in parity (547 keys)
- Branch cut fresh from `origin/main`; single-file diff.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #320, opened 2026-06-04
fix(ui): show REVIEWING (not stale CHANGES) while the critic re-reviews

scoop:
## Summary

When the critic requests changes, the agent addresses them, CI goes green, and the critic **re-reviews** to verify the fix. Until now the UI kept showing a prominent amber "CHANGES" signal during that re-review window — misleading the operator into thinking work was still pending when the solution was actually being validated.

Root cause: the "reviewing wins over the verdict" precedence rule lived **inline in only one component** (`CriticBadge`), so other surfaces that independently re-derived verdict prominence never learned it and drifted.

This change centralizes the rule and applies it everywhere:

- **New tested helper** `criticChip(verdict, reviewing)` in `critic-badge.ts` — one source of truth for "show REVIEWING vs. the verdict vs. nothing". `reviewing` always wins; `hasFindings` says whether the prior findings are still readable.
- **`CriticBadge`** now consumes the helper (behavior-preserving refactor).
- **`GitRail`** verdict-chip shows a `REVIEWING…` chip (amber outline + pulsing dot) while the critic re-reviews, and — by design — stays **clickable to open the prior findings** so you can read what's being validated. `aria-haspopup="dialog"` added to the dialog-opening buttons.
- **Findings popover header** shows the REVIEWING state too (no longer re-exposes the amber "CHANGES" pill once you open it).
- **`Viewport`** git-toggle "attention" amber hue is now gated on `!reviews.isReviewing(session.id)`, so it also stops screaming during re-review.

No server/data-model changes (reuses the existing `session:reviewing` signal + `reviews` store) and **no new i18n keys** (reuses `criticbadge_reviewing*`).

Out of scope (noted for a future pass): `PrRow`/`PrBadge` review dots reflect GitHub's raw posted review state, which GitHub doesn't auto-dismiss on the fix push.

Spec: \`docs/superpowers/specs/2026-06-04-critic-reviewing-hides-stale-changes-design.md\`
Plan: \`docs/superpowers/plans/2026-06-04-critic-reviewing-hides-stale-changes.md\`

## Test Plan

- [x] `cd ui && bun run check` — 0 errors / 0 warnings
- [x] `cd ui && bun run test` — 417/417 pass (incl. 6 new `criticChip` precedence cases)
- [x] `cd ui && bun run check:i18n` — locale parity (536 keys each, no new keys)
[...]

PR #2071, opened 2026-08-09
feat(epic): deterministic chain decomposition + stack-aware candidate selection

scoop:
-

### DIFF
--- a/src/epic-chains.ts
+++ b/src/epic-chains.ts
@@ -1,0 +1,118 @@
+import type { EpicChild } from "./epic-core";
+
+/** A chain decomposition of an epic's `blockedBy` DAG (#2066, epic #2063): a cover of the
+ *  children by disjoint linear chains, which is the shape a GitHub stack can take.
+ *
+ *  A chain covers **nodes, not edges**. With `C.blockedBy = [A, B]`, `C` can sit on top of at
+ *  most one of `A`/`B`; the other edge is *cross-chain* and keeps today's
+ *  `blockedBy.every(b => done.has(b))` gate in `selectEpicCandidates`, which stays
+ *  authoritative. Within-chain edges are the ones a later step converts into base pointers. */
+export interface EpicChainDecomposition {
+  /** Chains bottom→top (a chain's first entry is its root — the child that bases on the epic
+   *  integration branch). Disjoint; every child appears in exactly one. Ordered by the
+   *  canonical position of each chain's root. */
+  chains: number[][];
+  /** child # → its chain predecessor #. Absent ⇒ the child is a chain root. */
+  predecessorOf: Map<number, number>;
+}
+
+/** Children in canonical order: `order`, then `number` — the SAME tie-break
+ *  `selectEpicCandidates` uses. Every decision below keys on this position and nothing else
+ *  (never `Map`/`Set` iteration order, never merge state), which is what makes the cut
+ *  reproducible across drain ticks, restarts and input reordering. */
+function canonicalOrder(children: EpicChild[]): EpicChild[] {
+  return [...children].sort((a, b) => a.order - b.order || a.number - b.number);
+}
+
+/** Would `blocker → child` close a cycle? Chain links form a forest of disjoint simple paths,
+ *  so it does exactly when `child` already sits below `blocker` in its chain. `assembleEpic`
+ *  filters self-loops and out-of-epic edges but does NOT reject cycles, so a cyclic `blockedBy`
+ *  reaches here — and a cyclic chain would be a base pointing at itself transitively. */
+function closesCycle(predecessorOf: Map<number, number>, blocker: number, child: number): boolean {
+  let n: number | undefined = blocker;
+  while (n !== undefined) {
+    if (n === child) return true;
+    n = predecessorOf.get(n);
+  }
+  return false;
+}
+
+/** The chain predecessor to give `child`, or undefined when it starts a new chain.
+ *
+ *  Candidates are its blockers, deduped (the markdown path in `epic-model.ts` appends edges
+ *  without deduping), restricted to epic members, and not already spoken for — a node has at
+ *  most one successor, which is what keeps chains linear under fan-out.
+ *
+ *  Of those, the HIGHEST canonical blocker wins. A fan-in child is only ever spawnable early
+ *  while exactly one blocker is outstanding, and the latest-authored blocker is the one most
+ *  likely to still be in flight at that moment, so this absorbs strictly more edges than
+ *  picking the earliest. Both choices are equally deterministic. */
+function pickPredecessor(
+  child: EpicChild,
+  position: Map<number, number>,
+  taken: Set<number>,
+  predecessorOf: Map<number, number>,
+): number | undefined {
+  const blockers = [...new Set(child.blockedBy)]
+    .filter((b) => b !== child.number && position.has(b) && !taken.has(b))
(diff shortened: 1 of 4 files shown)


scoop (commented):
# Review — deterministic chain decomposition + stack-aware candidate selection (#2066, epic #2063)

**Decision: comment.** No blocking findings. The diff does what its task asks, the decomposition is
sound (disjoint, total, acyclic, deterministic), and the no-argument path is behaviourally identical
to the pre-image.

Diff under review: 4 files, +435/−3 — `src/epic-chains.ts` (new), `src/epic-core.ts`,
`test/epic-chains.test.ts` (new), `test/epic-core.test.ts`.

**1. Zero behaviour change on the existing path.** The only production-path edit is
`src/epic-core.ts:114`, which swaps `c.blockedBy.every((b) => done.has(b))` (the pre-image line, per
`git diff 91f57e86...HEAD -- src/epic-core.ts`) for `dependenciesSatisfied(c, done, stack)`. With
`stack` undefined, `src/epic-core.ts:90-92` computes `outstanding = [...new Set(c.blockedBy)].filter(b => !done.has(b))`
and returns `outstanding.length === 0`, else `false`. That is logically identical to `.every()` —
deduplication cannot change an all-quantified predicate, and the empty-`blockedBy` case yields `true`
on both (empty `outstanding` / vacuous `every`). `deriveChildState` (`src/epic-core.ts:68`) is
untouched and still uses the raw `.every()`, which is correct for this PR's scope.

**2. No caller passes the new argument.** Grepped the whole tree: the only production call sites are
`src/drain.ts:726` (`selectEpicCandidates(builtEpic.children)`) and `src/up-next.ts:291`
(`selectEpicCandidates(epic.children)[0] ?? null`) — both one-argument. `src/up-next-core.ts` only
mentions the function in comments (`src/up-next-core.ts:10,65,80`) and does not call it. `ui/` has no
reference at all, so there is no mirrored type or catalog to keep in parity. No message/i18n catalog
is touched by this diff, so the locale-parity rule in `CLAUDE.md:37-43` does not apply here.

**3. The decomposition's structural invariants.** Traced `src/epic-chains.ts:90-102`:

- *At most one predecessor per node* — `predecessorOf.set` is reached once per child, inside the
  single pass over `canonical` (`src/epic-chains.ts:95-100`).
- *At most one successor per node* — `taken` (`src/epic-chains.ts:94,99`) is consulted by the
  candidate filter at `src/epic-chains.ts:57`, so a blocker can be claimed only once. This is what
  keeps fan-out linear; confirmed by the fan-out fixture at `test/epic-chains.test.ts:53-58`.
- *Acyclic* — `closesCycle` (`src/epic-chains.ts:31-38`) walks the predecessor path from the
  candidate blocker and refuses the edge if it reaches `child`. Because links form a forest of simple
  paths, `blocker` is reachable from `child` via successors **iff** `child` lies on `blocker`'s
  predecessor path, so this check is complete, not merely a heuristic — it catches the case where the
  blocker was already assigned *above* `child` by an earlier iteration (which does happen: in the
  3-cycle fixture, #320 claims #322 as predecessor before #322 is itself processed). The walk
  terminates because acyclicity is preserved inductively. I hand-traced both cycle fixtures
  (`test/epic-chains.test.ts:89-104`) and got exactly the asserted `[[321, 320]]` and
  `[[322, 320, 321]]`.
- *Total and disjoint cover* — `readChains` (`src/epic-chains.ts:67-80`) emits a chain per node with
  no predecessor and walks the inverted map. A node reached through `successorOf` always has a
  predecessor, hence is skipped by the root filter at `src/epic-chains.ts:72` ⇒ no node appears
  twice; every node's predecessor path terminates at a root ⇒ no node is dropped. The
  `expectCovers` helper (`test/epic-chains.test.ts:33-39`) pins exactly this on every shape.
- *Real edges only* — every chain link comes from `child.blockedBy` (`src/epic-chains.ts:56`), so a
  within-chain link never inverts the dependency direction on acyclic input.

**4. Determinism.** Every decision keys on `position`, built from `canonicalOrder`
(`src/epic-chains.ts:23-25,92`), which sorts by `order` then `number`. Candidate blockers are
deduped and then **totally** sorted by distinct positions (`src/epic-chains.ts:56-58`), so neither
`Set` insertion order nor the `blockedBy` array order can leak; `readChains` iterates `canonical`,
not a `Map`. Uniqueness of `number` — the assumption documented at `src/epic-chains.ts:89` — holds:
`assembleEpic` dedupes with `const order = [...new Set(graph.order)]` (`src/epic-model.ts:124`) and
assigns `order: idx` (`src/epic-model.ts:152`), so `(order, number)` is a total order in practice.

**5. Cycle/self-loop/out-of-epic inputs.** `assembleEpic` already drops self-loops and out-of-epic
blockers (`src/epic-model.ts:135-145`) but does *not* reject cycles — the comment at
`src/epic-chains.ts:28-30` is accurate about that. `pickPredecessor` re-filters both defensively
(`src/epic-chains.ts:57`), which it needs anyway to look a blocker's position up.

**6. `stackPredecessorOf` vs. the `selectEpicCandidates` gate agree.** `stackPredecessorOf`
(`src/epic-chains.ts:108-118`) suppresses a done predecessor via `integrationMerged || issueClosed`,
which matches the done-set definition in `selectEpicCandidates` (`src/epic-core.ts:104-106`) and the
field docs at `src/epic-core.ts:24,28`. The gate at `src/epic-core.ts:93-94` uses the raw
`predecessorOf` map rather than `stackPredecessorOf`, but that cannot diverge: a done predecessor is
in `done`, so it is never in `outstanding`, and the stack branch cannot fire for it.

**7. Dead-code gate.** `stackPredecessorOf` and `chains` have no production consumer yet; the
`test/**/*.ts` fallow entry (`.fallowrc.jsonc:28`) makes test-only-reachable exports legitimate
roots, matching the plan's claim.

- `src/epic-chains.ts:56-58`: the successor slot is claimed on purely structural grounds, so when a fan-in child's highest-canonical blocker is already done-in-epic, that done node still consumes the slot and becomes the child's chain predecessor — the stack branch at `src/epic-core.ts:93-94` can then never fire for that child even when its one outstanding blocker is stack-ready. It fails safe (a missed stack, never a wrong base), is dormant until a caller passes `stack`, and the plan records it as a deliberate trade-off; noting it because it partly undercuts the "latest blocker is most likely still in flight" rationale at `src/epic-chains.ts:46-49`.

- `src/epic-chains.ts:104-118` (`stackPredecessorOf`) and the `chains` field (`src/epic-chains.ts:14`) ship with no production consumer in this PR; both are named in the task and approved plan, so this is sanctioned, not creep.

## Nits (non-blocking):

- `src/epic-core.ts:97-98`: the summary sentence still describes the gate as "all blockers done-in-epic", which is only the no-`stack` case; the qualification arrives two lines later at `src/epic-core.ts:101-102`.
- `src/epic-chains.ts:48`: "absorbs strictly more edges than picking the earliest" is stronger than the code guarantees — see the latent note above; "usually more" would be exact.
- `test/epic-chains.test.ts:118-123`: the self-loop case skips `expectCovers`, unlike every sibling test.

## Stated limitations

- I did not run `bun run lint` / `bun run test` — CI owns that, per the review instructions. All
  behavioural claims above come from reading the code and hand-tracing the fixtures, not from a run.
- I could not compare against a prior review round's findings: no `.shepherd-review.md` or equivalent
  exists in this worktree and the branch carries a single commit (`3aa92bb7`), so this review is
  written from scratch against the diff.

PR #2072, opened 2026-08-09
feat(forge): stacked-PR API surface on GithubForge (#2068)

### LINKED ISSUE #2068: GitForge: stacked-PR API surface (read/create/add/unstack) (#2063 P3)
## Goal

Give `GitForge` a stacked-PR surface: read a PR's stack, create a stack, append to one, and unstack.
GitHub-only; Gitea and Local omit the methods entirely.

No consumer in this issue.

## Why the native stack object is required

Plain base pointers (child 2's PR based on child 1's branch, no stack object) look sufficient right
up until child 1 **squash**-merges into the integration branch. GitHub then retargets child 2 to
`epic/<#>-<slug>`, but child 2's branch still carries child 1's *unsquashed* commits — a duplicate
diff and a near-certain conflict. Only the native stack's auto-rebase-the-remainder handles that.

Since `forge.mergeMethod` defaults to `"squash"` and this repo has merge commits disabled, there is
no merge method that avoids the problem. The stack object earns its keep here.

## Scope

`src/forge/types.ts`:

- `StackInfo` — `{ number, baseRef, prNumbers /* bottom→top */, position?, size? }`. `baseRef` is the
  stack **trunk**, which is the field consumers actually need: every PR in a stack is evaluated
  against the base of the stack, not the branch it directly targets.

`src/forge/github.ts`, all over `gh api` REST:

- `stackForPr?(prNumber): Promise<StackInfo | null>` — promote the existing private `probeStack()`
  to a full read. **Keep its fail-open contract**: a probe error (rate limit, transient 5xx) reports
  unstacked, which is exactly today's behaviour, so a flaky call can never be a regression.
- `createStack?(prNumbers): Promise<StackInfo>` — `POST /repos/{o}/{r}/stacks`, ordered bottom→top.
- `addToStack?(stackNumber, prNumber): Promise<void>` — `POST /stacks/{n}/add`.
- `unstack?(stackNumber): Promise<void>` — `POST /stacks/{n}/unstack`.

**Never shell out to `gh stack`.** It opens a full-screen TUI under a PTY and would wedge an
unattended agent (research note §2.5). REST only.

## Constraints worth encoding in the doc comments

- No reorder / insert / drop-one API exists. `gh stack modify` is TUI-only. The only repair primitive
  is unstack-and-recreate.
- `merge-async` takes no delete-branch parameter, so layer branches survive a stack merge and nothing
  reaps them (`BranchPruner` only sweeps local `shepherd/*`).
- The whole stacked-PR API is a public preview; every docs page says "subject to change".

## Acceptance criteria

- `bun run lint` and `bun run test` pass.
- Tests against a fake `gh` runner (mirroring the existing forge tests) cover each method, plus:
  `stackForPr` returning null for an unstacked PR, and `stackForPr` **failing open** to null on a
  runner throw rather than propagating.
- `GiteaForge` and `LocalForge` are untouched; a test asserts the methods are absent there so callers
  must branch on capability, not on `kind`.
- No production caller invokes the new methods in this PR.

scoop:
Gives `GitForge` a stacked-PR surface — read a PR's stack, create one, append to one, dissolve one
— on GitHub only, entirely over `gh api` REST. No production caller lands here; the consumer is a
later child of #2063.

Closes #2068

## What's here

| Method | Call | On failure |
| --- | --- | --- |
| `stackForPr(prNumber)` | `GET repos/{slug}/stacks?pull_request=n` | **fails open** → `null` |
| `createStack(prNumbers)` | `POST repos/{slug}/stacks`, `-F pull_requests[]=…` bottom→top | propagates |
| `addToStack(stackNumber, prNumber)` | `POST repos/{slug}/stacks/{n}/add` | propagates |
| `unstack(stackNumber)` | `POST repos/{slug}/stacks/{n}/unstack` | propagates |

Plus `StackInfo { number, baseRef, prNumbers /* bottom→top */, position?, size? }`. `baseRef` is
the stack **trunk** — every layer is evaluated against the base of the stack, not the branch it
directly targets, and getting that wrong is exactly what #2062 fixed on the merge path.

The read keeps the fail-open contract of the existing private `probeStack()`: any error (rate
limit, transient 5xx, unreadable body) reports "unstacked", which is today's behaviour everywhere,
so a flaky call can never be a regression. The three writes deliberately do **not** fail open — a
swallowed write failure would leave a caller believing a stack exists when it does not.

**Never `gh stack`.** Under a PTY it opens a full-screen TUI and would wedge an unattended agent
(research note §2.5). A test asserts every call is a `gh api` call.

## Two decisions worth a reviewer's attention

**`stackForPr` reads the stack resource, not the PR's embedded `.stack`.** `prNumbers` needs the
whole membership list, and GitHub documents `pull_requests` on the stack resource but not on the
`.stack` object hanging off a pull request. The documented `pull_request=` filter was verified
live against this repo (`200 []`; stacks are enabled here — a 404 is how "not enabled" presents).
Membership is nonetheless re-checked locally: were that filter ever dropped or ignored host-side,
the endpoint would answer with some *other* stack in the repo, and reporting that as this PR's
stack is the one failure mode worse than reporting nothing.

**`probeStack()` stays private and `merge()` is untouched**, despite the issue phrasing this as
"promote `probeStack()` to a full read". The two answer different questions: the merge probe must
return the head SHA (merge-async's concurrency guard, deliberately absent from `StackInfo`), the
public read must return membership (which the PR endpoint does not promise). Routing `merge()`
through the new read would cost a second REST call on the merge hot path and disturb #2059/#2062
behaviour that 15 tests currently pin. Both carry the same fail-open contract and cross-reference
each other in their doc comments.

## Constraints recorded in the doc comments

No reorder / insert / drop-one endpoint exists (`gh stack modify` is TUI-only), so the only repair
primitive is unstack-and-recreate · `merge-async` takes no delete-branch parameter, so layer
branches survive a stack merge and nothing reaps them (`BranchPruner` only sweeps local
`shepherd/*`) · the whole API is a public preview, "subject to change" on every docs page.

## Verification

`bun run lint`, `bunx tsc --noEmit`, `bun run test` (8441 pass / 0 fail) and
`fallow audit --base origin/main --fail-on-issues` (no issues in the 3 changed files) all clean.
15 new tests against the fake `gh` runner cover each method, the null/unstacked and fail-open-on-
throw paths the issue calls out, error propagation on all three writes, and the absence of all
four methods on `GiteaForge` and `LocalForge`.

### DIFF
--- a/src/forge/github.ts
+++ b/src/forge/github.ts
@@ -46,6 +46,7 @@ import type {
   RedeployInput,
   RepoCounts,
   RollupEntry,
+  StackInfo,
   SubIssueRef,
   WorkflowJob,
   WorkflowRun,
@@ -101,6 +102,21 @@ interface StackProbe {
  *  fails open to today's behaviour. */
 function isPathSafeRef(ref: string): boolean {
   return ref.length <= 255 && ref.split("/").every((seg) => /^\w[\w.-]*$/.test(seg));
+}
+
+/** `--jq` projection of one stack resource onto {@link StackInfo}'s wire shape (#2068). Narrowing
+ *  host-side keeps the payload small; `// []` keeps jq from erroring on a stack the host returned
+ *  without a membership list, which {@link stackInfoFrom} then rejects as unusable. */
+const STACK_JQ = "{number, baseRef: .base.ref, prNumbers: [(.pull_requests // [])[].number]}";
+
+/** Validate one {@link STACK_JQ} projection. Null for anything unusable — a missing stack number,
+ *  a missing trunk, or no members — so both the fail-open read and the throwing writes have one
+ *  definition of "the host did not describe a stack we can address". */
+function stackInfoFrom(raw: Partial<StackInfo> | null | undefined): StackInfo | null {
+  if (!raw || typeof raw.number !== "number" || !raw.baseRef) return null;
+  const prNumbers = (raw.prNumbers ?? []).filter((n) => typeof n === "number");
+  if (!prNumbers.length) return null;
+  return { number: raw.number, baseRef: raw.baseRef, prNumbers };
 }
 
 /** A merge-request uuid is interpolated straight into the `gh api` path, so it is validated
@@ -1716,6 +1732,81 @@ export class GithubForge implements GitForge {
     const args = ["pr", "merge", String(prNumber), "--repo", this.slug, method];
     if (o.deleteBranch) args.push("--delete-branch");
     await this.run(args);
+  }
+
+  /** The stack `prNumber` belongs to, or null when it belongs to none (#2068).
+   *
+   *  Reads the STACK resource (`GET /stacks?pull_request=`) rather than the `.stack` object
+   *  embedded in the pull request, because only the stack resource carries `pull_requests` — the
+   *  membership list is the whole point of this read. The sibling {@link probeStack} takes the
+   *  other route for the opposite reason: the merge path needs the head SHA, which lives on the
+   *  pull request and not on the stack.
+   *
+   *  FAILS OPEN to null, exactly as {@link probeStack} does — see its note. A stack whose trunk
+   *  the host did not report is also null: `baseRef` is the load-bearing field, and handing a
+   *  caller an empty ref to interpolate or compare against is worse than reporting nothing.
+   *
+   *  NEVER shells out to `gh stack`: under a PTY it opens a full-screen TUI and would wedge an
+   *  unattended agent. Every stack call in this file is a `gh api` call. */
+  async stackForPr(prNumber: number): Promise<StackInfo | null> {
+    try {
+      // Projected as an ARRAY so the unstacked case (`[]`) stays valid jq input and parses to [].
+      const out = await this.run([
+        "api",
+        `repos/${this.slug}/stacks?pull_request=${prNumber}`,
+        "--jq",
+        `[.[] | ${STACK_JQ}]`,
+      ]);
+      const stacks = JSON.parse(out || "[]") as Partial<StackInfo>[];
+      const info = stackInfoFrom(stacks[0]);
+      // Membership is re-checked locally rather than taken on trust: were the `pull_request=`
+      // filter ever dropped or ignored host-side, the endpoint would answer with SOME OTHER
+      // stack in the repo, and reporting that one as this PR's stack is the one failure mode
+      // worse than reporting nothing.
+      const at = info ? info.prNumbers.indexOf(prNumber) : -1;
+      if (!info || at < 0) return null;
+      return { ...info, position: at + 1, size: info.prNumbers.length };
+    } catch (err) {
+      console.warn(`[github] stack read failed for pr#${prNumber}; assuming unstacked:`, err);
+      return null;
+    }
+  }
+
+  /** Link `prNumbers` (BOTTOM → TOP) into a new stack. A mutation: host errors propagate, and an
+   *  unreadable response body throws rather than reporting a stack nobody can address later. */
+  async createStack(prNumbers: number[]): Promise<StackInfo> {
+    const out = await this.run([
+      "api",
+      "--method",
+      "POST",
+      `repos/${this.slug}/stacks`,
+      "--jq",
+      STACK_JQ,
+      ...prNumbers.flatMap((n) => ["-F", `pull_requests[]=${n}`]),
+    ]);
+    const info = stackInfoFrom(JSON.parse(out || "null") as Partial<StackInfo> | null);
+    if (!info) throw new Error(`create stack for #${prNumbers.join(", #")} returned no stack`);
+    return { ...info, size: info.prNumbers.length };
+  }
+
+  /** Append one pull request to the TOP of an existing stack. Mutation — errors propagate. */
+  async addToStack(stackNumber: number, prNumber: number): Promise<void> {
+    await this.run([
+      "api",
+      "--method",
+      "POST",
+      `repos/${this.slug}/stacks/${stackNumber}/add`,
+      "-F",
+      `pull_requests[]=${prNumber}`,
+    ]);
+  }
+
+  /** Dissolve a stack. Its unmerged pull requests are unlinked; merged or merge-queued ones stay
+   *  linked host-side. The only repair primitive there is — no reorder/insert/drop-one API
+   *  exists, so a stack that needs reshaping is unstacked and recreated. Mutation — errors
+   *  propagate. */
+  async unstack(stackNumber: number): Promise<void> {
+    await this.run(["api", "--method", "POST", `repos/${this.slug}/stacks/${stackNumber}/unstack`]);
   }
 
   /** Stack membership, head SHA and the stack's trunk in one REST call, narrowed with `--jq` so
--- a/src/forge/types.ts
+++ b/src/forge/types.ts
@@ -383,6 +383,27 @@ export class MergePendingError extends MergeNotCompletedError {
   }
 }
 
+/** One GitHub pull-request stack, as the stacked-PR API surface reports it (#2068).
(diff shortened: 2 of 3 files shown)


scoop (commented):
# Review — GitForge: stacked-PR API surface (read/create/add/unstack) (#2068, epic #2063 P3)

**Decision: comment.** The diff does what the task asks, in the shape the plan approved, and I found
no blocking correctness or security defect. Three files, additions only:
`src/forge/types.ts` (+44), `src/forge/github.ts` (+91), `test/forge/github-stacks.test.ts` (+179).

**The four methods exist on GitHub only, and nowhere else.** `implements GitForge` resolves to exactly
three classes — `src/forge/github.ts:520`, `src/forge/gitea.ts:65`, `src/forge/local.ts:320` — and the
diff touches only the first. Neither `src/forge/gitea.ts` nor `src/forge/local.ts` appears in the diff,
so the capability-absence assertion at `test/forge/github-stacks.test.ts:174-177` is testing a real
absence, not a stub. The additions at `src/forge/types.ts:683,687,689,692` are all optional (`?`), so
the ~30 object-literal fake forges typed `GitForge` across `test/` (e.g. `test/drain.test.ts:76`,
`test/pr-poller.test.ts:75`) still typecheck without change — I checked that none of them is a
class-based proxy that would have to forward the new members.

**No production caller.** `grep` for `stackForPr|createStack|addToStack|unstack` returns hits only in
`src/forge/types.ts` (declarations), `src/forge/github.ts` (implementations), `test/forge/github-stacks.test.ts`,
one unrelated identifier `unstackedProbe` in `test/forge/github-stacked-merge.test.ts:51`, and prose in
`docs/research/github-stacked-prs-2026-08-08.md`. The issue's "no consumer in this PR" criterion holds.

**Endpoints and field names match the repo's own ground truth.**
- `GET /stacks?pull_request=<n>` at `src/forge/github.ts:1755` — verified against
  `docs/research/github-stacked-prs-2026-08-08.md:106`, including the `pull_request` (singular) query key.
- `POST /repos/{o}/{r}/stacks` with `-F 'pull_requests[]=<n>'` at `src/forge/github.ts:1780-1785` —
  verified character-for-character against the live-verified spike invocation at
  `docs/research/stacked-pr-partial-merge-spike-2026-08-09.md:344`. `-F` (typed) rather than `-f` is
  what makes these send JSON numbers, and it matches the existing `-F` uses at `src/forge/github.ts:683,827`.
- `POST /stacks/{n}/add` (`src/forge/github.ts:1797`) and `POST /stacks/{n}/unstack`
  (`src/forge/github.ts:1809`) — verified against `docs/research/github-stacked-prs-2026-08-08.md:104-105`.
- `baseRef: .base.ref` in the projection at `src/forge/github.ts:110` reads the stack resource's own
  `base.ref`, which `docs/research/github-stacked-prs-2026-08-08.md:68-70` identifies as the trunk —
  i.e. deliberately not the `base_ref` trap documented at `:118-120`. The type's doc comment at
  `src/forge/types.ts:394-397` says the same thing, so comment and code agree.

**No `gh stack` anywhere.** Every new call's `argv[0]` is `"api"` (`src/forge/github.ts:1753, 1779,
1795, 1809`), and the guard test at `test/forge/github-stacks.test.ts:153-163` asserts it across all
four methods. This is the constraint from `docs/research/github-stacked-prs-2026-08-08.md:107,144`
(TUI-only, would wedge an unattended agent), and it is honoured.

**Fail-open read / propagating writes, as specified.** `stackForPr` wraps everything in try/catch and
returns `null` with a `console.warn` (`src/forge/github.ts:1769-1771`), matching the existing
`probeStack` contract verbatim at `src/forge/github.ts:1839-1841` — same message shape, same
`[github]` prefix, same "assuming unstacked" wording, consistent with the other fail-open warns at
`src/forge/github.ts:1880,1892`. The three writes have no catch, so host errors propagate.

**Command-injection surface.** All calls go through the argv-array `GhRunner`
(`src/forge/github.ts:297`, `execFile` at `:304`) — no shell, so the `?`, `[`, `]` characters in the
paths and fields are not interpreted. The only values interpolated into a URL path are
`prNumber`/`stackNumber`, both typed `number`, so unlike the string refs guarded by `isPathSafeRef`
(`src/forge/github.ts:104`) and the uuid guarded by `MERGE_UUID_RE` (`:125`) they cannot introduce a
`../` segment. Declining to validate them (plan D5) is defensible.

**The membership re-check is a genuine improvement over the plan.** `src/forge/github.ts:1766-1768`
confirms `prNumber` is actually in the returned membership before reporting the stack, so a host that
ignored the `pull_request=` filter yields `null` rather than somebody else's stack. Covered at
`test/forge/github-stacks.test.ts:81-87`.

**jq projection is array-safe on both paths.** `[.[] | …]` at `src/forge/github.ts:1758` keeps the
unstacked `[]` case valid, and `(.pull_requests // [])` at `:110` keeps jq from erroring when the key
is absent; `JSON.parse(out || "[]")` at `:1760` covers gh printing nothing. The test fakes reply with
the *post-`--jq` projection* (`test/forge/github-stacks.test.ts:35-39`), which is the same convention
the existing suite uses (`test/forge/github-stacked-merge.test.ts:51,58-60`), so the seam is faithful.

- The **request-body field name for `POST /stacks/{n}/add`** (`pull_requests[]=` at
  `src/forge/github.ts:1799`) is not documented anywhere in this repo —
  `docs/research/github-stacked-prs-2026-08-08.md:104` records only the path. It is a reasonable
  extrapolation from the verified create invocation, but it is not verified. Assumed as written.
- The **response shape of `POST /repos/{o}/{r}/stacks`** — specifically whether it embeds
  `pull_requests` — is likewise unverified in-repo; the spike at
  `docs/research/stacked-pr-partial-merge-spike-2026-08-09.md:344` shows the request but not the body.
  `createStack` depends on that embed (see the latent note below). Assumed as written. The plan
  flagged this same risk explicitly.
- Whether GitHub's own `stack.position`/`stack.size` count already-merged layers the same way the
  membership array does — `docs/research/github-stacked-prs-2026-08-08.md:82` documents only "1-based,
  1 = bottom". Not resolvable without a live stack.

- `src/forge/github.ts:1787-1788`: if `POST /stacks` succeeds but its body carries no `pull_requests`, `stackInfoFrom` returns null and `createStack` throws — the stack exists host-side but the caller learns neither that nor its `number`, and the only cleanup primitive (`unstack`) needs exactly that number; a first consumer would have to recover via `stackForPr` on one of the PRs it just linked. Since the caller already knows the numbers it sent, falling back to those rather than throwing would remove the orphan case.
- `src/forge/github.ts:1768`: `position`/`size` are derived from the returned membership rather than read from the host's `stack.position`/`stack.size`; a later consumer that gates on `position === size` (the CI-cost mitigation in `docs/research/github-stacked-prs-2026-08-08.md:139-142`) inherits any divergence in how merged layers are counted. The doc at `src/forge/types.ts:402-403` marks the fields advisory, which contains it for now.
- `src/forge/github.ts:1760-1765`: only `stacks[0]` is considered; a host that returned more than one stack with the PR in the second reports "unstacked". Fail-open, so harmless today, but undocumented at the call site.

## Nits (non-blocking):

- `src/forge/types.ts:404`: `size` shares `position`'s doc comment and has none of its own, so a hover on `size` alone shows nothing.
- `src/forge/github.ts:1787`: `JSON.parse(out || "null")` only defends against the empty string — a whitespace-only body (`"\n"`) is truthy and raises a bare `SyntaxError` instead of the intentional "returned no stack" message. Trimming first would make the failure mode uniform.
- `src/forge/github.ts:1766-1768`: the non-membership and missing-trunk rejections return null silently while every other fail-open path in this file warns (`:1770`, `:1840`, `:1880`, `:1892`); the plan's D4 said this case would be warned.

PR #2073, opened 2026-08-09
refactor(epic): persist epic-child identity as a session fact (#2067)

### LINKED ISSUE #2067: Epic-child identity must be a session fact, not a base-branch name test (#2063 P2)
## Goal

Stop deciding "is this session an epic child?" from the shape of its base branch name. Replace the
`isEpicIntegrationBranch(s.baseBranch)` heuristic with a persisted per-session fact and route every
live site through it.

**Behaviour-identical refactor. No flag, no stacking, no functional change** — but it is the
prerequisite that makes stacking safe, and it removes a latent hazard on its own.

## Why this is urgent independent of stacks

Six sites answer the epic-child question from the base branch name. Once a child is based on its
chain predecessor's `shepherd/*` branch instead of `epic/<#>-<slug>`, every one of them silently
flips to the wrong answer:

| Site | Symbol | Consequence once a child is stacked |
| --- | --- | --- |
| `src/full-auto.ts` | `isFullAuto()` | the merge train carries the child and merges it **into its predecessor's branch** |
| `src/drain.ts` | `doRetire()` | takes the plain retire path — PR left open, never integrated, epic stalls with no operator signal |
| `src/review.ts` | epic critic context | base-delta prose lost for stacked children |
| `src/standalone-critic.ts` | `EpicContext` from `baseRefName` | same |
| `src/drain-core.ts` | doc comment | stale, must be corrected |
| `src/critic-core.ts` | doc comment | stale, must be corrected |

The first two are silent, not loud: one merges to the wrong place, the other stalls forever.

## Scope

- Add `epicParent: number | null` to the session row (`src/types.ts`, `src/store.ts` column +
  migration via the existing `addMissingColumns` pattern), stamped at spawn from the drain decision
  that already carries `integrationBranch`.
- **The fact must live on the session, not behind a store lookup.** `isFullAuto` is a pure function
  taking `(s, cfg)` with no store handle; a lookup-based predicate could not be used there without
  changing its shape and every caller.
- One shared predicate; all four live sites call it and none re-derive the answer locally.
- **Legacy rows** (null `epicParent`, written before this change) fall back to the branch-name test,
  so in-flight epics keep working across the deploy.

## Acceptance criteria

- `bun run lint`, `bun run test`, and `cd ui && bun run check && bun run test` pass.
- Every existing epic drain / full-auto / critic test passes **unmodified** — that is the evidence
  the refactor is behaviour-identical.
- A test covers a legacy session row (`epicParent` null, base = `epic/12-foo`) still reading as an
  epic child.
- A test covers a non-epic session (`epicParent` null, base = default branch) still reading as not
  an epic child.
- `grep -rn "isEpicIntegrationBranch" src` shows no remaining *base-branch* call sites. Uses against
  a PR **head** ref (`src/standalone-critic.ts` landing-PR sweep) are correct and stay.

scoop:
Closes #2067. Part of #2063 (Step 2 of #2060).

## What

"Is this session an epic child?" was answered at six sites from the *shape of its base branch name*. This persists the answer on the session row and routes every live site through one shared predicate.

- `Session.epicParent` — the epic parent issue number, stamped **at spawn, in the same INSERT** (no post-create setter, so there is no window where the row exists without the fact).
- One predicate in `src/epic-branch.ts`:
  ```ts
  isEpicChild({epicParent, baseBranch}) = epicParent != null || isEpicIntegrationBranch(baseBranch)
  ```
  Pure (no store handle) so `isFullAuto(s, cfg)` keeps its shape; structural param so the session-less standalone critic can call it too. The legacy fallback lives **inside** the predicate — no call site can forget it.

| Site | Change |
| --- | --- |
| `src/full-auto.ts` → `isFullAuto()` | `isEpicChild(s)` |
| `src/drain.ts` → `doRetire()` | `isEpicChild(s)` |
| `src/review.ts` → epic critic context | `isEpicChild(session)` |
| `src/standalone-critic.ts` → `EpicContext` | resolves the fact by PR head branch (`getEpicParentByBranch`), then `isEpicChild(...)` |
| `src/drain-core.ts`, `src/critic-core.ts` | stale doc comments corrected |

The `isEpicIntegrationBranch(pr.headRefName)` eligibility check in `standalone-critic.ts` → `eligible()` tests a PR **head** ref (is this the epic's aggregate landing PR?), not base-branch identity. Correct as-is, kept.

Behaviour-identical: no flag, no stacking, no functional change.

## Decisions worth reviewing

1. **Stamped only when the child actually got the integration branch** (`usingEpicBase`). On a forge without `ensureBranch` (gitea/local) the spawn degrades onto the default branch and today reads as *not* an epic child — the merge train carries it, it retires normally, the epic still progresses. Stamping on `decision.integrationBranch` alone would silently reroute every degraded epic through the squash-into-integration retire path.
2. **`epicParent` is optional on `Session`** (`?: number | null`), following the documented `autoMergeRebaseSteeredAt` precedent. Required would have forced fixture edits across the suite, against the issue's "existing tests pass unmodified" criterion. The store always hydrates a real `number | null`.
3. **No dedicated setter.** The house rule guards against `update()`'s whitelist silently dropping writes — it applies to fields that get updated. This one is immutable spawn identity; nothing updates it. The single accessor added is `getEpicParentByBranch()`, the session-less critic's only route to the fact.
4. **Relaunch carries the stamp only when neither `repoPath` nor `baseBranch` is overridden.** Same repo + same base means the branch-name fallback would have answered identically before this field existed, so it is behaviour-identical; it just stops a bare relaunch from silently dropping the identity. A relaunch that re-points either falls back to the branch test.
5. Three `?? null` sites were written branch-free (`numOrNull` helper in `store.ts`, `carriedEpicParent` helper in `service.ts`, direct passthrough in `create`) because `buildSessionRow` / `hydrate` / the INSERT arrow / `service.create` all sit at their fallow complexity caps — the existing code carries comments saying exactly that.

## Adjacent, deliberately out of scope

`src/merge-teardown.ts` → `recordEpicIntegrationIfChild()` and `src/drain.ts` → `epicChildBaseBlocked()` do **not** use the name heuristic — they compare a merged/actual PR base against the **pinned** integration branch. Untouched here, but those comparisons also stop matching once children stack; they need the recorded integration target, not an identity predicate. Flagged for the stacking step of #2063.

## Verification

- `bun run lint`, `bun run test` (8441 pass), `cd ui && bun run check && bun run test` (4574 pass), `fallow audit --base origin/main --fail-on-issues` clean.
- Every pre-existing epic drain / full-auto / critic test passes **unmodified** — the behaviour-identical evidence. The only edit to an existing test body is one line in the standalone critic's shared fake-store fixture (the new `getEpicParentByBranch` seam) plus one stale comment; no assertion changed.
- New coverage: legacy row (null stamp, `epic/12-foo` base) still reads as a child; non-epic row reads as not a child; a stamped child on a non-epic base is excluded from full-auto and takes the epic retire path; spawn stamps on the epic base and does **not** stamp when degraded; store round-trip + a real pre-column DB migrating to null.
- `grep -rn "isEpicIntegrationBranch" src` → only the definition + the predicate in `epic-branch.ts`, and the head-ref use in `standalone-critic.ts`.

### DIFF
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -43,9 +43,10 @@ export interface EpicBaseDelta {
   commitsTruncated: number;
 }
 
-/** Epic-child review context. Present iff the reviewed branch's base is an epic integration branch
- *  (`isEpicIntegrationBranch`), i.e. this PR is ONE CHILD of a draining epic whose base already
- *  carries merged sibling work. `baseSha` null = the base could not be resolved to a commit (the
+/** Epic-child review context. Present iff the reviewed session is an epic child (`isEpicChild` —
+ *  the persisted `epicParent` fact, falling back to the base-branch name for legacy rows), i.e. this
+ *  PR is ONE CHILD of a draining epic whose base already carries merged sibling work.
+ *  `baseSha` null = the base could not be resolved to a commit (the
  *  fetch/rev-parse failed and there is usually no local ref for an epic branch) → the block degrades
  *  to its no-base mode: no base commands, and existence conclusions become limitations, not
  *  findings. */
--- a/src/drain-core.ts
+++ b/src/drain-core.ts
@@ -54,6 +54,9 @@ export type DrainDecision =
       kind: "spawn";
       issue: Issue;
       integrationBranch?: string;
+      /** Epic parent issue number to stamp on the spawned session (#2067). Set from the same
+       *  epic run as `integrationBranch`, so the two cannot drift; undefined for label-drain. */
+      epicParent?: number;
       epicProviderSettings?: EpicProviderSettings;
     }
   | { kind: "retire"; sessionId: string; prNumber: number }
@@ -120,6 +123,9 @@ export interface DrainRepoState {
   /** Epic mode: the active epic's integration branch — epic-child spawns base on it.
    *  null/undefined for label-drain (spawns base on the default branch). */
   epicIntegrationBranch?: string | null;
+  /** Epic mode: the active epic's PARENT issue number, stamped on each child it spawns (#2067).
+   *  Set from the same epic run as `epicIntegrationBranch`; null/undefined for label-drain. */
+  epicParent?: number | null;
   /** Epic mode: explicit provider/model/effort for child spawns. Undefined for label-drain
    *  and for epics inheriting the repo/global defaults. */
   epicProviderSettings?: EpicProviderSettings | null;
@@ -291,8 +297,9 @@ export function computeNext(state: DrainRepoState): DrainDecision {
   // 2b. Epic base unavailable (#1757): the forge's `ensureBranch` THREW for a child spawn, so the
   // integration branch could not be ensured. Basing the child on the default branch instead would
   // silently mix bases mid-epic — the merge train would land that one child on main (it is not
-  // excluded from full-auto: `isEpicIntegrationBranch(baseBranch)` is false for it) while its
-  // siblings integrate on the epic branch. So fail closed and hold. Gated on an active epic run, so
+  // excluded from full-auto: a child that did NOT get the integration branch carries no `epicParent`
+  // stamp, so `isEpicChild` is false for it) while its siblings integrate on the epic branch.
+  // So fail closed and hold. Gated on an active epic run, so
   // it can never fire in label-mode drain; while an epic runs the candidate set IS that epic's
   // children, and every sibling would fail identically, so pausing new spawns is exactly right.
   // Placed AFTER the retire gate (a ready PR still retires) and BEFORE the cap. The marker is
@@ -334,6 +341,7 @@ export function computeNext(state: DrainRepoState): DrainDecision {
     kind: "spawn",
     issue: next,
     integrationBranch: state.epicIntegrationBranch ?? undefined,
+    epicParent: state.epicParent ?? undefined,
     epicProviderSettings: state.epicProviderSettings ?? undefined,
   };
 }
--- a/src/drain.ts
+++ b/src/drain.ts
@@ -18,7 +18,7 @@ import {
 import { assembleEpic } from "./epic-model";
 import {
   epicIntegrationBranch as epicBranchName,
-  isEpicIntegrationBranch,
+  isEpicChild,
   branchReferencesEpic,
 } from "./epic-branch";
 import { selectEpicCandidates, type Epic, type EpicRun } from "./epic-core";
@@ -2419,11 +2419,11 @@ export class DrainService {
     const s = this.deps.store.get(decision.sessionId);
     // Epic child: squash-merge the PR INTO its integration branch (not the default branch) and
     // record it so dependents unblock without a GitHub issue auto-close (the child issue stays
-    // open until the final epic→default PR lands). Detected by the integration-branch base +
-    // an active epic for the repo.
+    // open until the final epic→default PR lands). Detected by the session's persisted epic-child
+    // identity (#2067) + an active epic for the repo.
     const epicRun = this.deps.store.getEpicRun(repoPath);
     const epicActive = !!epicRun && (epicRun.status === "running" || epicRun.status === "paused");
-    if (epicActive && s?.issueNumber != null && isEpicIntegrationBranch(s.baseBranch)) {
+    if (epicActive && s?.issueNumber != null && isEpicChild(s)) {
       // #645 (Task 2): enforce the child PR's actual base against the integration branch. On
       // mismatch (or while throttled-blocked from a prior mismatch) this returns true → fail
       // closed: skip merge/record/archive/claim-drop so the child stays un-integrated and the
@@ -2473,10 +2473,10 @@ export class DrainService {
    *
    *  - `ensureBranch` THREW (GitHub; a rate-limit, 5xx, race): FAIL CLOSED — throw
    *    {@link EpicBaseUnavailableError}. Degrading this ONE child onto the default branch would mix
-   *    bases mid-epic: it would lose `epicBaseDirective`, open its PR against main, and — since
-   *    `isFullAuto` excludes only sessions whose base IS an integration branch — the merge train
-   *    would then land it on main automatically, while its siblings integrate on the epic branch.
-   *    A retry may well succeed, so doSpawn's catch backs off (cooldown) and the drain holds
+   *    bases mid-epic: it would lose `epicBaseDirective`, open its PR against main, and — since a
+   *    degraded child gets NO `epicParent` stamp, so `isFullAuto` does not exclude it — the merge
+   *    train would then land it on main automatically, while its siblings integrate on the epic
+   *    branch. A retry may well succeed, so doSpawn's catch backs off (cooldown) and the drain holds
    *    visibly (`epic_base_unavailable`) meanwhile.
    *
    *  - The forge simply LACKS `ensureBranch` (gitea/local): DEGRADE, as before. On such a forge NO
@@ -2490,7 +2490,7 @@ export class DrainService {
   private async resolveSpawnBase(
     forge: GitForge,
     decision: Extract<DrainDecision, { kind: "spawn" }>,
-  ): Promise<{ base: string; prompt: string }> {
+  ): Promise<{ base: string; prompt: string; epicParent: number | null }> {
     const { number, title } = decision.issue;
     const def = await forge.defaultBranch();
     let base = def;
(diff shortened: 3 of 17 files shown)


scoop (commented):
# Review — Epic-child identity as a session fact (#2067)

**Decision: comment.** No blocking findings. The refactor does what the task asks, the "behaviour-identical
today" claim holds under the checks I ran, and the one genuinely load-bearing decision (stamp only when
`usingEpicBase`) is implemented correctly.

## What I verified

### The predicate and its call sites

- `isEpicChild` is defined once at `src/epic-branch.ts:45-47` and is the only place the legacy branch-name
  fallback lives.
- All four live sites route through it, and no site re-derives the answer: `src/full-auto.ts:26`,
  `src/drain.ts:2426`, `src/review.ts:529`, `src/standalone-critic.ts:418-421`.
- The issue's acceptance grep holds. `isEpicIntegrationBranch` now appears in `src` only at its definition
  (`src/epic-branch.ts:19`), inside the shared predicate (`src/epic-branch.ts:46`), and at the PR-**head**
  eligibility test `src/standalone-critic.ts:292` — which is a head-ref test, correctly left alone. I also
  swept for any other name-shaped heuristic (`"epic/"` / `` `epic/ `` literals across `src`) and found only
  branch-*construction* and doc/prose uses (`src/epic-branch.ts:12`, `src/epic-model.ts:192`,
  `src/drain.ts:578`), no second identity test.
- `isFullAuto`'s param widening is type-safe: `Session.epicParent` is optional (`src/types.ts:120`), so
  `Pick<…|"epicParent">` (`src/full-auto.ts:18-21`) stays optional and every caller
  (`src/automerge.ts:127`, `src/index.ts:1260`, `src/index.ts:1977`, `src/drain.ts:685`) passes a full
  `Session` — none needed changing.
- `doRetire`'s narrowing is sound: `s?.issueNumber != null` narrows `s` to non-null before
  `isEpicChild(s)` at `src/drain.ts:2426`.

### The stamp is written only for a real epic-base child

`src/drain.ts:2522` computes `const epicParent = usingEpicBase ? (decision.epicParent ?? null) : null;`,
where `usingEpicBase` (`src/drain.ts:2515`) requires the resolved base to actually equal
`decision.integrationBranch`. So the gitea/local degraded child (forge without `ensureBranch`) gets no
stamp and, with a default-branch base, still reads as *not* an epic child — which is exactly what the
existing degraded-child behaviour depends on. `decision.epicParent` and `decision.integrationBranch` are
both set from the same `epicRun` in the same place (`src/drain-core.ts:343-344`, sourced from
`src/drain.ts:710` / `src/drain.ts:760-761`), so they cannot drift.

Because the stamp is written *only* when the base already is the integration branch, `isEpicChild` returns
the identical answer to `isEpicIntegrationBranch(baseBranch)` for every row this PR can produce. The
"behaviour-identical" claim checks out.

### Store plumbing

- Column count matches placeholder count matches value count. `COLS` (`src/store.ts:560-569`) enumerates 61
  columns; the INSERT placeholder list is exactly 61 (`src/store.ts:2326` — confirmed by matching
  `VALUES \((\?,){60}\?\)`); the value array `src/store.ts:2328-2388` is 61 entries, with
  `epicParent` at position 33, directly after `s.issueNumber`, matching `COLS`' `auto, issueNumber,
  epicParent, sandboxApplied…`.
- `src/store.ts:2326` is the *only* `INSERT INTO sessions` in `src`, so there is no second writer to keep
  in sync.
- `NewSession` (`src/store.ts:487-558`) is `Omit<Session, …>` and does **not** omit `epicParent`, so
  `input.epicParent` at `src/store.ts:2285` resolves to the optional `number | null` from
  `src/types.ts:120`. No separate declaration was needed.
- The migration covers both fresh and existing DBs: `migrateSessionColumns()` is invoked unconditionally
  at `src/store.ts:1169`, immediately after the `CREATE TABLE IF NOT EXISTS sessions` stanza
  (`src/store.ts:1158-1168`) which does *not* carry the column — so the `add("epicParent", "epicParent
  INTEGER")` at `src/store.ts:4061` is what creates it in both cases. Nullable, no default → legacy rows
  read `null` → fallback. The new migration test at `test/store.test.ts:2039-2068` exercises exactly that.
- `hydrate` reads it at `src/store.ts:5599`; every `SELECT` in the class goes through `${COLS}`
  (`src/store.ts:2397, 2426, 2509, 2518, 2529, 2541, 2716`), so the column is always present in the row.

### Not reachable from HTTP

This was the risk worth checking, since `StandardCreateInput` is an HTTP-facing type. It is not spoofable:
`validateCreate` rejects any key outside `ALLOWED_KEYS` (`src/validate.ts:42-60, 537-539`) — `epicParent`
is not in that set — and it then builds the value field-by-field (`src/validate.ts:563-580`) rather than
spreading the body. The relaunch path is the same shape: `RELAUNCH_ALLOWED_KEYS`
(`src/validate.ts:744-758`) has no `epicParent`, and `carriedEpicParent` (`src/service.ts:1956-1964`)
reads only `original.epicParent`. The other two `StandardCreateInput` construction sites
(`src/server.ts:1247-1272` and `src/service.ts:4009-4027`) also build field-by-field. So the only writer
is the drain spawn path.

Related: `src/service.ts:4009` (`replaceAgent`) omits `epicParent`, which is correct — that path
explicitly does not create a session row (`src/service.ts:3980-3983`); it reuses the existing one.
`startVariant` (`src/service.ts:4194`) goes through `relaunch`/`buildRelaunchInput` with no
repo/base override, so a variant carries the stamp; today that changes nothing (the variant inherits the
original's `epic/*` base, so both the old and new predicate answer true, and `startVariant` forces
auto-merge off at `src/service.ts:4208` regardless).

### Standalone critic

`getEpicParentByBranch` exists on `SessionStore` (`src/store.ts:2408-2417`) and is added to the deps `Pick`
(`src/standalone-critic.ts:87`), so the real construction site type-checks. The `pr.headRefName!`
non-null assertion at `src/standalone-critic.ts:419` matches the assertion the same value already gets at
`src/standalone-critic.ts:365`, on a path that runs earlier in the same function — no new risk. A PR with
no Shepherd session resolves to `null` and falls back to the base name, i.e. today's answer.

The landing-PR carve-out at `src/standalone-critic.ts:428-431` still holds: no landing-repair or
landing session is ever stamped (only the drain's epic-child spawn writes the field), so a lookup keyed on
the integration branch returns `null` and `epic` stays null for the landing PR.

### Base-delta check

Nothing in this diff depends on the sibling-merged paths. The only identifier the touched code takes from
`src/epic-core.ts` is `epicRun.parentIssueNumber` at `src/drain.ts:710`, which is an unchanged line — and
it still exists at base (verified against
`epic/2063-stack-epic-children-onto-the-integration@5d837951d7dd31daae5f10ed4af21df38d10bd3b:src/epic-core.ts`,
`EpicRun.parentIssueNumber` at line 34). `selectEpicCandidates` at the same base gained an *optional*
second parameter (line 103 of that blob), so the unchanged `src/drain.ts:726` call remains valid.
`src/drain-core.ts`, `src/drain.ts`, `src/store.ts`, `src/service.ts` and `src/full-auto.ts` are not in the
base-delta set, so my tree's versions are the base versions.

### Doc-comment corrections

Both stale comments are now factually right, and the parenthetical in `src/drain-core.ts:297-302` stays
balanced. `src/drain-core.ts:300-301` correctly describes the counterfactual (a degraded child carries no
stamp → `isEpicChild` false → not excluded from full-auto), as does `src/drain.ts:2476-2479`.
`src/critic-core.ts:46-49` likewise.

## Stated limitations

- I did not run `bun run lint` / `bun run test` / the UI checks — per the brief, CI owns those. In
  particular I did not execute the new drain/standalone-critic tests; I only read them for consistency with
  the source changes.

- src/review.ts:529 — once children actually stack, `epicBase = session.baseBranch` hands the critic the *predecessor's* branch as "the epic base", so the collected base delta will be the predecessor delta rather than the integration-branch delta; correct today because a stamped child's base always is the integration branch.
- src/standalone-critic.ts:423 — same shape: `base: meta.baseRefName` will be the predecessor's ref for a stacked child, so the epic block's base commands will point at the wrong branch once stacking lands.
- src/standalone-critic.ts:428 — the "epic child and epic landing PR are mutually exclusive" invariant now additionally depends on no stamped session ever owning the integration branch as its *work* branch; true today (only drain epic-child spawns are stamped, and their branch is `shepherd/*`), but it is no longer guaranteed by branch-name shape alone.

## Nits (non-blocking):

- src/store.ts:2405-2406 — the doc justifies `ORDER BY createdAt DESC` as letting "a relaunched child's row win over a dead predecessor that never got the stamp", but `WHERE … epicParent IS NOT NULL` already excludes unstamped rows; the ordering only disambiguates between two *stamped* rows.
- test/store.test.ts:2011-2022 — the test named "newest row first" inherits that same misconception: its unstamped predecessor row is filtered out by `epicParent IS NOT NULL`, not by the ordering, so nothing there actually exercises the `ORDER BY`.
- src/store.ts:2360 — `numOrNull(s.epicParent)` is redundant; `buildSessionRow` already normalized the value at `src/store.ts:2285`, and `s` is that built row.
- src/store.ts:5599 — `epicParent: r.epicParent` drops the `?? null` its immediate neighbours use (`issueNumber: r.issueNumber ?? null` on the line above); harmless since the column is always selected, but it reads as an inconsistency.
- src/epic-branch.ts:43-44 — "Note the fallback can only ADD childness, never remove it" attributes the additive property to the fallback; it is the *stamp* that can only add (the sentence's own justification, "a stamped child always reads true", is about the stamp).

- `src/merge-teardown.ts` → `recordEpicIntegrationIfChild()` compares the merged PR's base against the pinned integration branch rather than using a name heuristic, so this PR correctly leaves it alone — but that comparison will also stop matching once children stack, and it needs the recorded integration target rather than an identity predicate. The plan already flags it for the stacking step.

Issue #2070, opened 2026-08-09
Stacked epic children: bottom-up retire, mid-stack loss repair, repo-setup docs (#2063 P5)


scoop:
## Goal

Make stacked epic children land safely: merge only the bottom-most unmerged layer, repair a stack
that loses a middle layer, and document the repo-setup shapes that break stacking.

Closes out #2063 (Step 2 of #2060, shape (b)).

**Blocked by #2069 (stacked spawn + composition + base gate).**

## Retire becomes bottom-up

`mergeStacked` (`src/forge/github.ts`, shipped in #2061) refuses any stacked PR whose caller did not
set `allowStacked`, and only the two operator HTTP endpoints do. `retireEpicChild` is autonomous — so
the moment children become stack members, every epic-child retire throws `stacked_refused` and the
epic stalls.

The carve-out: `retireEpicChild` sets `allowStacked: true` **only** when the child is the
bottom-most unmerged layer of its stack. Merging that layer lands exactly one PR, so #2061's
rationale — an autonomous merge silently landing ungated layers beneath it — is *preserved*, not
waived. A higher layer holds, with a named hold reason, until the layers below it retire.

Merge order becomes bottom-up within a chain. Spawn order does not, which is the whole point of the
epic: children still squash-merge into the integration branch at retire, so `integrationMerged`
remains the cross-chain done-signal that `selectEpicCandidates` reads.

`deleteBranch` is ignored by `mergeStacked` by design, so merged layer branches linger on the remote
and nothing reaps them. Record it; do not fix it here.

## Mid-stack loss

A closed or abandoned middle layer blocks every layer above it, and GitHub has **no reorder or
drop-one API** — `gh stack modify` is TUI-only, so unstack-and-recreate is the only repair primitive.

When a stacked child is abandoned or its PR is closed: call `unstack` on the whole stack and raise a
**blocking** `assembleEpic` warning naming the wedged children and the remedy, following the existing
`baseMismatchWarning` shape in `src/epic-model.ts`. Fail-visible — layers above keep their branches
for an operator or a repair steer. Today an abandoned epic child is harmless; under stacks it is not,
so it must be loud.

## Docs

- **Ruleset shape to avoid**: a ruleset requiring a status check on *all* branches blocks the restack
  force-push outright (`GH013: Repository rule violations found for refs/heads/<layer>`). Branch
  *creation* can be exempted (`do_not_enforce_on_create`); *updates* cannot, and a restack is an
[...]

PR #2074, opened 2026-08-09
feat(epic): stack epic children on their predecessor's PR (#2069)

### LINKED ISSUE #2069: Stacked epic children: spawn on predecessor's PR, compose the stack, teach the base gate (#2063 P4)
## Goal

Wire the stack up: spawn an epic child on its chain predecessor's PR head branch, compose a GitHub
stack rooted at the pinned integration branch, and teach the retire-time base check the stack shape.

First behaviour change of #2063, behind a new per-repo flag `epicStacksEnabled` (default **OFF**).

**Blocked by #2066 (chains), #2067 (epic-child identity), #2068 (forge stack API).**

## Spawn trigger

A stacked child spawns once its chain predecessor has an **open PR**. That is the earliest point
where the branch is definitely pushed, the head ref is authoritative, and the PR number needed by the
stack API exists.

Child *N* therefore starts at child *N−1*'s PR-open instead of its merge — the whole CI + critic +
merge cycle earlier than today.

## Scope

- **Flag**: `epicStacksEnabled` on `RepoConfig` — `src/store.ts` column + default, `src/server.ts`
  PATCH allowlist, `ui/src/lib/components/AutomationSettings.svelte` toggle, `ui/src/lib/types.ts` /
  `api.ts`, and matching **EN + DE** catalog keys (i18n is mandatory for user-facing strings).
- **Spawn base**: `resolveSpawnBase()` (`src/drain.ts`) bases a stacked child on the predecessor's PR
  head branch, and `epicBaseDirective()` (`src/autopilot.ts`) names that branch as the PR base.
  `ensureBranch` is **not** called for a stacked base — the branch is the predecessor's live head;
  creating it would be actively wrong, not merely redundant.
- **Composition**: when a stacked child's PR is first observed open, create or extend the stack so it
  is rooted at the pinned integration branch. Idempotent and throttled. Stack creation needs at least
  two PRs bottom→top, so the bottom layer (whose PR bases on the integration branch) is what seeds it.
- **Store**: new `epic_stack` table (`repoPath`, `parentIssueNumber`, `stackNumber`, `childNumber`,
  `prNumber`, `baseBranch`, `position`), following the `epic_integrated` / `epic_base_mismatch`
  pattern, and pruned by the same epic teardown path (verify, don't assume).
- **Base gate**: `epicChildBaseBlocked()` learns the stack through **one** shared derivation that
  every call site uses. A PR that is a stack member is accepted when the stack's **trunk** equals the
  pinned integration branch; a non-member keeps today's `actual === s.baseBranch` rule.

## Why the gate must read the trunk, not the direct base ref

The moment a lower layer merges, GitHub auto-restacks the remainder and re-targets the next PR to the
trunk. A check comparing the PR's direct base against the recorded `s.baseBranch` would then see
`epic/<#>-<slug>` where it recorded a `shepherd/*` branch, and fail closed on **every** stacked child
as soon as its predecessor lands. Reading the stack trunk is what survives that.

## Design input from the Step 1 spike

- The top layer's own state does not reveal that the stack is unmergeable — a top PR reported
  `mergeable_state: clean` while a layer beneath it was `blocked`. Any readiness judgement must be
  derived from the whole stack, never from the selected PR.
- A conflict against the trunk is invisible until merge time: layers are diffed against the layer
  below, so a stack that has fallen behind its trunk reports every layer clean and then fails. Per-PR
  mergeability is not a readiness signal once the trunk has moved.

## Acceptance criteria

- `bun run lint`, `bun run test`, and `cd ui && bun run check && bun run test` pass.
- With `epicStacksEnabled` OFF, epic behaviour is unchanged and existing epic tests pass unmodified.
- A test covers the auto-restack case: a stacked child whose PR has been re-targeted to the
  integration branch after its predecessor merged is **not** blocked by the base gate.
- A test covers a genuine base mismatch still failing closed.
- Locale parity (EN + DE) passes the catalog gate.
- **Re-verify, do not assume**, that the landing PR is unaffected: the single `Closes #parent`
  (`buildLandingPrBody`), the one release-please-parseable title (`buildLandingPrTitle`, #1206), the
  standalone critic's landing-PR sweep of a non-`criticAllPrs` repo, and the migration-awareness gate
  (`migrationPaths` / `migrationsAckedAt`).

scoop:
P4 of epic #2063. Wires the stack up behind a new per-repo flag `epicStacksEnabled` (**default OFF**, GitHub-only).

Closes #2069

## What lands

- **Spawn** (`resolveSpawnBase`, `buildState`) — a child whose only outstanding blocker is its chain predecessor spawns as soon as that predecessor's PR is **open**, based on its head branch, a whole CI + critic + merge cycle earlier than today. `ensureBranch` is deliberately **not** called for a stacked base: the branch is the predecessor's live head, and "ensuring" it would recreate a deleted branch at the default tip, silently basing the child on nothing. The child is still stamped `epicParent`, so it stays off the merge train and retires into the epic.
- **Compose** (`composeEpicStacksForRepo`, in `tick()` beside the other self-guarding epic passes) — links child PRs into a stack rooted at the pinned integration branch, one step per pass, throttled, and idempotent: it adopts a stack the host already has rather than creating a second one. Seeding first verifies the **bottom** PR's base is the pinned branch, because that base becomes the trunk for every layer above it and children are only *advised* where to target.
- **Base gate** (`epicChildBaseBlocked` → `epicChildBaseOk`) — reads the stack **trunk**, not the PR's direct base, which is what survives GitHub's auto-restack when a lower layer merges and the next PR is re-targeted at the trunk.
- **Retire** — keeps the merged child's head branch when a spawned successor is still sitting on it.
- New `epic_stack` table + `src/epic-stack.ts` (all the derivations, pure and unit-tested).

## Two things worth reviewer attention

**Plain base equality stopped being safe.** Once a session's `baseBranch` can be a sibling's `shepherd/*` head, an uncomposed stacked child has `actual === sessionBase === <sibling's branch>` — and the old rule would have squash-merged it **into its sibling's branch** and recorded it integrated there. A stacked base is now accepted *only* via membership in a stack rooted at the pinned branch; otherwise it fails closed with the existing `epic_base_mismatch` marker, whose warning already names the re-target remedy. "Stacked base" is decided by branch shape, not by the flag, so flipping the flag off mid-epic cannot resurrect the unsafe accept — and a superseded epic's `epic/*` base keeps the old rule.

**`deleteBranch` had to be suppressed.** A predecessor whose successor has spawned but not yet opened a PR cannot be a stack member (`createStack` needs two PRs), so its retire takes the ordinary `gh pr merge --delete-branch` path and deletes the branch the successor is based on — `gh pr create --base <deleted>` then fails outright. `mergeStacked` ignores `deleteBranch` for exactly this reason (#2068); this is the legacy path's equivalent. Suppression, not a hold, so the predecessor still merges and still unblocks its cross-chain dependents.

## Why the flag ships off

`mergeStacked` (#2061) refuses any stacked PR whose caller did not set `allowStacked`, and `retireEpicChild` is autonomous. So a stacked child cannot be merged until the bottom-most-layer carve-out in **#2070**. **Do not enable `epicStacksEnabled` on a real repo before #2070 lands** — the toggle's own description says so too.

## Deliberately out of scope

- Bottom-up retire / `allowStacked`, and mid-stack loss repair — #2070.
- `recordEpicIntegrated(mergedBase)` still records `s.baseBranch`; a stacked child cannot reach that call in P4 (the gate accepts it only as a stack member, and those merges are refused), so P5 fixes it with its merge change.
- `src/up-next.ts` still calls `selectEpicCandidates` without a stack context — the drain stays authoritative; building a second, separately-derived context there is how the two would drift.
- No prune for `epic_stack`. Verified rather than assumed: there is no epic teardown path in `src/store.ts` — nothing prunes `epic_integrated` either, and the only `DELETE FROM epic_*` is the per-child `clearEpicBaseMismatch` self-heal.

## Verification

`bun run lint`, `bun run test` (8542 pass), `cd ui && bun run check && bun run test && bun run check:i18n`, and `fallow audit --base origin/main --fail-on-issues` all pass. `buildState` was refactored back under its recorded complexity cap rather than re-baselined.

Re-verified, not assumed, that the landing PR is unaffected: `buildLandingPrBody`'s single `Closes #parent`, `buildLandingPrTitle` (#1206), the standalone critic's landing-PR sweep, and the `migrationPaths`/`migrationsAckedAt` gate — none of those sources are touched and their tests pass unmodified.

**Diff note:** ~45 of the changed files are a one-line `epicStacksEnabled: false` added to full-`RepoConfig` test fixtures (`setRepoConfig` takes a complete config), plus EN+DE catalog keys and one feature-catalog fragment.


scoop (commented):
# Review — #2069 stacked epic children (spawn on predecessor's PR, compose the stack, teach the base gate)

Verdict: **comment**. I found no defect that is reachable with `epicStacksEnabled` off (its shipped
default), and no locale/identifier/contract inconsistency. Several real problems exist on the
flag-on path; per the latent-defect routing they are reported below as informational rather than as
blocking findings, but items 1 and 2 are worth fixing before anyone flips the toggle.

**Flag plumbing is complete and internally consistent.**

- `RepoConfig.epicStacksEnabled` declared at `src/store.ts:444-448`, row type `src/store.ts:812`,
  default `src/store.ts:988`, row→config mapping `src/store.ts:1019`, bind param
  `src/store.ts:1053`.
- The `repo_config` CREATE at `src/store.ts:1231-1241` is the minimal legacy shape and every other
  column arrives via `migrateRepoConfigColumns`, so the `add("epicStacksEnabled", …)` at
  `src/store.ts:4272-4273` is the only migration needed — no missing CREATE-time column.
- Column/placeholder/param alignment checked by count, not by eye: the INSERT column list at
  `src/store.ts:1702-1706` names 30 columns, the VALUES clause at `src/store.ts:1707` has exactly 30
  placeholders (grep `VALUES \(\?(,\?){29}\)` matches that line and only that line), and
  `repoConfigParams` (`src/store.ts:1028-1058`) returns 30 values in the same order.
  `epicStacksEnabled` sits at position 25 in all three. The `excluded.` upsert entry is at
  `src/store.ts:1731` and the SELECT list at `src/store.ts:1693`.
- Server allowlist: `REPO_CFG_BOOL_FIELDS` `src/server.ts:1422`, `RepoCfgBody` `src/server.ts:1441`,
  the parse return type `src/server.ts:1542` and the patch assignment `src/server.ts:1627` — all four
  present, so the generic boolean validation at the `REPO_CFG_BOOL_FIELDS` loop applies (confirmed by
  the new 400 test at `test/server.test.ts:2496-2502`).
- UI: `ui/src/lib/types.ts:823-825`, `ui/src/lib/api.ts:2110`, store field/hydration/toggle/getter at
  `ui/src/lib/reviews.svelte.ts:348,389,480,734-742,820-822`, and the switch row at
  `ui/src/lib/components/AutomationSettings.svelte:499-518`. The row mirrors its neighbours
  (`lightweight`-disabled, `lightweightTip`, `role="switch"`, `aria-checked`, `aria-label`) —
  compared against the `preWarmEpicLandingCi` row at
  `ui/src/lib/components/AutomationSettings.svelte:483-497`.

**Locale parity holds.** `automation_epic_stacks_name` / `automation_epic_stacks_desc` exist at
`ui/messages/en.json:2177-2178` **and** `ui/messages/de.json:2177-2178`;
`feat_epic_stacks_title` / `feat_epic_stacks_body` at `ui/messages/en.json:3553-3554` **and**
`ui/messages/de.json:3553-3554`. `ui/messages/` contains only `en.json` and `de.json`, so those are
all the catalogs the UI package maintains. Every key the diff references
(`ui/src/lib/components/AutomationSettings.svelte:504,505,514` and
`ui/src/lib/feature-announcements/entries/v1.47.0-epic-stacks.ts:6-7`) resolves to both.

**Feature-catalog fragment is well-formed.** The catalog is `import.meta.glob`-loaded
(`ui/src/lib/feature-announcements.ts:46-49`), so no index registration is needed; the filename
`v1.47.0-epic-stacks.ts` matches `sinceVersion: "1.47.0"` (the contract at
`ui/src/lib/feature-announcements.ts:10-23`), and `id: "epic-stacks"` does not collide — the only
other `epic_stacks`/`epic-stacks` occurrences in `ui/` are the four catalog keys and this file.

**Fixture sweep is complete.** Every file that carries a full `RepoConfig` literal (32 files matching
`preWarmEpicLandingCi:`) also matches `epicStacksEnabled:`; the `epicStacksEnabled` match set is a
strict superset (36 files — the extra four being `ui/src/lib/api.ts`, `src/epic-stack.ts`,
`src/drain-core.ts`, `src/drain.ts`). Nothing was missed.

**Identifiers the new module depends on all exist and are used with the right shapes.**
`isEpicIntegrationBranch` (`src/epic-branch.ts:19`), `stackPredecessorOf` /
`EpicChainDecomposition.predecessorOf` (`src/epic-chains.ts:106-118`, `:11-17`), `EpicChild` fields
`integrationMerged`/`issueClosed`/`sessionId`/`claimed` (`src/epic-core.ts:22-29`),
`EpicStackContext` (`src/epic-core.ts:73-79`) and the optional second parameter of
`selectEpicCandidates` (`src/epic-core.ts:103`), and `StackInfo.baseRef`/`prNumbers`/`number`
(`src/forge/types.ts:390-405`). `forgeHasStacks` (`src/drain.ts:290-296`) branches on the three
optional methods declared at `src/forge/types.ts:683-689`, not on `kind` — correct for Gitea/Local,
which omit them. `getEpicIntegrationBranch` (the read-only getter, `src/store.ts:1935`) and the two
new accessors are all present in the `DrainDeps.store` `Pick` (`src/drain.ts:217-233`).

**The base gate reads the trunk, as the issue requires.** `stackRootedAtEpic`
(`src/epic-stack.ts:21-23`) compares `stack.baseRef` — documented at `src/forge/types.ts:394-397` as
the stack's trunk, explicitly *not* the layer's direct base — against the pinned branch. The
auto-restack acceptance and the genuine-mismatch refusal are both covered by tests
(`test/drain-epic-retire-base.test.ts:353-372` and `:400-411`), and the F5a regression (an
uncomposed stacked child would otherwise squash into its sibling's branch) by
`test/drain-epic-retire-base.test.ts:374-387`.

**`ON CONFLICT DO UPDATE` without a conflict target** in `recordEpicStackMember`
(`src/store.ts:2094-2098`) matches the existing house convention — the same targetless form is used
at `src/store.ts:1956`, `:2031`, `:2155`, `:3234` — and the PK is `(repoPath, parentIssueNumber,
childNumber)`, so it resolves unambiguously. `createdAt` is deliberately not in the SET list.
Round-trip + upsert-in-place covered at `test/store.test.ts:407-455`.

**"Pruned by the same epic teardown path" — the issue's premise, re-checked rather than assumed.**
`DELETE FROM epic_` appears exactly once in the tree, at `src/store.ts:2042`
(`clearEpicBaseMismatch`). There is no epic teardown path to hang a prune off, so the author's
decision to follow the `epic_integrated` no-prune convention is right and the comment at
`src/store.ts:1520-1523` states it accurately.

**Flag-OFF is genuinely byte-identical.** I walked each new branch point:

1. `buildState` (`src/drain.ts:779-784`) — `epicStackContext` returns `{baseByChild: null}` at its
   first guard (`src/drain.ts:1029-1031`), so `selectEpicCandidates` is called with `undefined`,
   which `src/epic-core.ts:89-95` documents (and implements) as the pre-#2066 predicate.
2. `computeNext` (`src/drain-core.ts:354`) — `stackedBase` is `undefined`.
3. `resolveSpawnBase` (`src/drain.ts:2853`) — the new early return is gated on `decision.stackedBase`.
4. `retireEpicChild` (`src/drain.ts:2709` → `:2672`) — `keepBranchForStackedSuccessor` returns
   `false` on the flag check before touching the epic, so `deleteBranch: true` as before.
5. `tick` (`src/drain.ts:3256` → `:846`) — the pass returns on the flag check before any forge call.
6. `epicChildBaseBlocked` — the only path that changes shape. The new `stacked` predicate
   (`src/drain.ts:2616`, `src/epic-stack.ts:33`) is `sessionBase !== pinned && !isEpicIntegrationBranch(sessionBase)`.
   I checked whether it can be `true` for a pre-#2069 child: reaching this function requires
   `isEpicChild(s)` (`src/drain.ts:2779`, predicate at `src/epic-branch.ts:45-46`), i.e. either an
   `epicParent` stamp or an epic-shaped base. `epicParent` is only ever written by
   `resolveSpawnBase`'s `usingEpicBase` branch (`src/drain.ts:2889`, where `base === integrationBranch`,
   which is the pinned value from `getOrInitEpicIntegrationBranch` at `src/drain.ts:775`) and carried
   across relaunch only when neither repo nor base is re-pointed (`src/service.ts:1958-1963`). Every
   other route in leaves an epic-shaped base, which `isEpicIntegrationBranch` excludes — including a
   superseded epic's `epic/<other>-<slug>` and legacy unstamped rows. So `stacked` is false and the
   rule collapses to today's `actual === s.baseBranch`. `readPrStack` (`src/drain.ts:2650-2664`)
   additionally re-reads the flag before calling the host, which the flag-off test at
   `test/drain-epic-retire-base.test.ts:413-424` pins.

**Landing-PR invariants unaffected.** `src/epic-landing.ts`, `src/standalone-critic.ts` and the
`migrationPaths`/`migrationsAckedAt` path are not in the diff, and the stacked child still carries an
`epicParent` stamp (`src/drain.ts:2861`) with its own `shepherd/*` head, so
`getEpicParentByBranch(pr.headRefName)` (`src/standalone-critic.ts:419`) answers as before.

- `src/drain.ts:895-906`: `composeOneEpicStack` `return`s after the first chain whose plan is not `none`, but `seedEpicStack` also returns without mutating when the bottom PR's base is not the pinned branch (`src/drain.ts:936-942`) or when `adoptExistingStack` finds a foreign-rooted stack (`src/drain.ts:975-981`) — so one permanently-refusing chain starves every later chain in a multi-chain epic on every pass, forever, instead of only consuming that pass's single-mutation budget.
- `src/drain.ts:2709`: `keepBranchForStackedSuccessor` awaits `buildEpic` (`src/drain.ts:2675`) outside any try/catch and before `retireEpicChild`'s own try; `buildEpic` issues `getIssue`/`listSubIssues`/`listBlockedBy` when the `epicStructure` TTL has lapsed, and a throw there escapes `doRetire` → `pump` (no catch, `src/drain.ts:2551-2567`) → `tick`'s repo loop (`src/drain.ts:3275`), aborting the sweep for every remaining repo. Every other forge call on the epic-retire path is individually caught.
- `src/drain.ts:2724`: `recordEpicIntegrated` still records `s.baseBranch` as the merged base, which for a stacked child is its sibling's branch rather than the branch the PR actually landed in. Unreachable while `GithubForge.merge` routes stack members to `mergeStacked` and refuses without `allowStacked` (`src/forge/github.ts:1729,1857`), which is exactly the #2070 carve-out — but the accept in `epicChildBaseOk` (`src/epic-stack.ts:60`) is what makes it reachable the moment that refusal is lifted.
- `src/store.ts:444-448` / `ui/messages/en.json:2178`: the flag it introduces is not safe to enable until #2070 lands — a composed stacked child's autonomous retire throws `StackedMergeRefusedError` (`src/forge/github.ts:1857`), which `retireEpicChild` swallows (`src/drain.ts:2712-2718`) and retries every tick, stalling the epic. This is acknowledged in the plan (F2), the toggle copy and the store comment; noting it only so the state of the flag is on the record.
- `src/drain.ts:884-886`: `prByChild` accepts any `EpicChild.prNumber`, which `buildEpic` populates from the PR cache regardless of state (`src/drain.ts:513`), whereas the spawn-side `stackPredecessorFacts` requires `state === "open"` (`src/drain.ts:1000`). A child whose PR was closed can therefore be planned into a `createStack`/`addToStack` call; the host rejects it and the pass logs, but the asymmetry between the two sibling derivations is real.

## Nits (non-blocking):

- `src/drain.ts:3249-3256`: the new `#2069` comment block was inserted between the existing `#1664` paragraph and the `ensureDraftLandingPrForRepo` call it documents, so `#1664`'s text now sits directly above `composeEpicStacksForRepo`.
- `src/drain.ts:936-942`: the refuse-to-seed warning fires once per throttle window for as long as the bottom PR targets the wrong branch — a persistent console warning with no dedup, unlike the `epic_base_mismatch` marker used for the equivalent condition at retire.
- `src/drain.ts:911-919`: `epicChildSpawnBases` falls back to `ctx.pinned` for a child whose session has been archived (`src/drain.ts:914`), so a re-observed layer can have its recorded `baseBranch` rewritten from the sibling branch to the epic branch. The column is informational today, so nothing reads the difference.

## Stated limitations

- I could not verify GitHub's own stacked-PR semantics against a live host — the preview API is not
  in this repo. Specifically: whether a stack survives once every layer but one has merged (which is
  what keeps `stackForPr` answering non-null for the last unmerged layer, and therefore what keeps
  the auto-restack accept from flipping to a fail-closed block), and whether `addToStack` still
  succeeds on a stack whose lower layers have all landed. Assumed as written in
  `src/forge/types.ts:390-405`; not asserted correct either way.
- I did not run lint or any test suite (CI owns those); all statements above come from reading the
  code and the cited files.

Issue #2083, opened 2026-08-14
Per-token scopes for minted access tokens (read / submit / full)


kai-osthoff:
## Problem or motivation

#2082 shipped named machine access tokens: you mint one in **Settings → Access**, copy it once,
and revoke it later. What it deliberately did **not** ship is any notion of *what a token may do*.

A v1 minted token is **full parity with `SHEPHERD_TOKEN`**. It authenticates every gated surface,
including the WebSocket upgrades — `/events` and `/pty/:id`. That falls out of the shared
`checkAuth` seam in `src/server.ts` with no extra code, and it is what makes a minted token a
genuine drop-in replacement for the env var rather than a weaker cousin.

The cost is concrete. The immediate driver for #2082 was the
[`shepherd-asyar-extension`](https://github.com/erwins-enkel/shepherd-asyar-extension) — a launcher
extension that is **strictly read-only**: it consumes `GET /api/sessions` and `GET /api/holds` to
show which sessions are waiting on the operator. The token you paste into that extension's
preference field can, today, also:

- spawn sessions (`POST /api/sessions`),
- type into any live agent terminal (`/pty/:id`),
- archive, relaunch or interrupt work.

That is a much larger blast radius than the client needs, and it is the same for every future
client: a cron job that only submits tasks, a phone shortcut that only reads status, a Raycast or
Azure extension.

This was a **conscious v1 trade-off**, not an oversight — recorded in the #2082 plan and agreed
with the maintainer. #2082 mitigates it with attribution (each token is named) and instant
revocation. This issue is the structural fix.

## Proposed solution

Give each token a **scope set**, chosen when it is minted and shown in the token list.

A first cut with three levels, coarse on purpose — scopes that are too fine become a UI nobody
uses correctly:

| Scope       | Grants                                                                 |
| ----------- | ---------------------------------------------------------------------- |
| `read`      | `GET` on the read surfaces (`/api/sessions`, `/api/holds`, `/api/git`, `/events`) |
| `submit`    | `read` + `POST /api/sessions` and the held-task routes — queue work, don't drive it |
| `full`      | Everything, including `/pty/:id`. Today's behaviour, and the migration target for existing rows |

Sketch of where it lands:

- **Storage** — a `scope TEXT NOT NULL DEFAULT 'full'` column on `access_tokens`
  (`src/store.ts`), so every token minted before this change keeps working unchanged.
[...]

PR #2084, opened 2026-08-14
feat(access-tokens): mint, copy and revoke named machine tokens from the UI

### LINKED ISSUE #2082: Access tokens: mint, copy and revoke named machine tokens from the UI
### Before you file

- [x] I searched existing issues and discussions for this idea. Closest prior art: #1079 (the
      password → session-cookie layer this builds on), #374 (Capture-extension pairing UX, which
      documents the manual token step but does not mint one), #660 (API-key auth for the *agent*
      provider — a different subsystem).

## Problem or motivation

Every machine client that talks to Shepherd needs a bearer token, and **there is no way to obtain
one from Shepherd itself.** The only route today is setting `SHEPHERD_TOKEN=<random>` in the
server's environment and then reading that value back out of the deployment by hand — SSH to the
host, `systemctl show` / grep the unit file or `.env`, copy it across.

That is fine when the operator is also the sysadmin sitting at the server. It stops being fine as
soon as a client wants to be *installed* rather than *deployed*:

- **`erwins-enkel/shepherd-asyar-extension`** (the immediate driver) — an Asyar launcher extension
  that shows which sessions are waiting on the operator. The user pastes a base URL and a token
  into an extension preference field. There is nowhere in Shepherd to copy that token from.
- **The Chrome Capture extension** already has the same hole: `docs-site/.../capture-extension.md`
  tells the user to fill in a *"Token — only if the server runs with `SHEPHERD_TOKEN` set."* It
  never says where to get it, because there is nowhere to get it.
- Any future client (cron job, Hermes, a phone shortcut, a second launcher) repeats the problem.

Beyond discoverability, the single shared env secret has three structural properties that only get
worse as clients multiply:

1. **No attribution.** Every client presents the same string. The access log cannot distinguish the
   Asyar extension from a cron job from a leaked copy.
2. **No revocation.** Revoking one client means rotating `SHEPHERD_TOKEN` and re-pasting it into
   every other client. "One client compromised" escalates to "reconfigure everything".
3. **Restart-coupled.** The value lives in the process environment, so changing it is a deploy, not
   a settings change.

This is the ordinary shape of the problem that GitHub PATs, Raycast extension credentials, and
Home Assistant's long-lived access tokens all solve the same way: mint a named token in the UI,
show it exactly once, copy it to the clipboard, revoke it later from a list.

## Proposed solution

A **Access / Tokens** surface in Settings that mints named, long-lived machine tokens.

### UI

A new panel alongside the existing ones in `ui/src/lib/components/settings/`
(`SettingsDevicePanel.svelte`, `SettingsPluginsPanel.svelte`, `SettingsDiagnosePanel.svelte`, …) —
either its own section or a group inside **Gerät**, whichever fits the existing IA better.

- **Create token** → asks for a name (e.g. *"Asyar extension — MacBook"*), optionally an expiry.
- On creation the plaintext token is shown **once**, with a copy-to-clipboard button and an
  explicit "you will not see this again" note.
- A list of existing tokens showing name, created-at, last-used-at, expiry — value never shown
  again.
- **Revoke** per row, effective immediately.

### Server

- Store only a **hash** of the token (the repo already has argon2id via `Bun.password` in
  `src/operator-auth.ts`; a SHA-256 of a high-entropy random string is the cheaper and more usual
  choice for tokens, since unlike passwords they are not low-entropy and are verified on every
  request — pick whichever the maintainer prefers, but not plaintext at rest).
- Give tokens a recognisable prefix (`shp_…`) so they are greppable in logs and detectable by
  secret scanners.
- Verify inside the **existing `checkAuth` seam in `src/server.ts`** — the module docstring in
  `src/operator-auth.ts` already names that seam as the extension point ("the identity check lives
  behind one seam (`checkAuth` in `server.ts`) so multi-user/SSO could be added later without a
  rewrite"). This proposal is deliberately the smaller thing that fits behind it.
- Keep the comparison timing-safe, as the `SHEPHERD_TOKEN` path already is.
- Stamp `last-used-at` (throttled — this is on the hot path of a server that also pumps the web
  terminal; a per-token write-once-per-minute is enough).

### Compatibility

`SHEPHERD_TOKEN` keeps working unchanged. It is the right mechanism for a container/systemd
deployment where the secret is provisioned by the platform, and removing it would break every
existing setup including the documented Hermes flow in `docs/external-task-api.md`. Minted tokens
are an **additional** accepted credential, not a replacement.

## Acceptance criteria

- [ ] An operator logged into the HUD can mint a named token and copy it to the clipboard without
      touching the server host.
- [ ] The plaintext is shown exactly once; reloading the settings page does not reveal it.
- [ ] `Authorization: Bearer <minted token>` authenticates against `/api/sessions` and `/api/holds`
      exactly like `SHEPHERD_TOKEN` does.
- [ ] Revoking a token causes the next request carrying it to return `401`, with no restart.
- [ ] Tokens are not stored in plaintext at rest.
- [ ] `SHEPHERD_TOKEN` continues to authenticate; existing deployments and the Hermes flow in
      `docs/external-task-api.md` are unaffected.
- [ ] `docs/external-task-api.md` and `docs-site/.../capture-extension.md` point at the new UI as
      the recommended way to get a token, with the env var documented as the deployment-provisioned
      alternative.

## Alternatives considered

- **Keep `SHEPHERD_TOKEN` only, document how to read it out of the deployment.** Cheapest, and it
  is what happens today by default. It does not fix attribution or revocation, and it makes every
  client installation a server-administration task. It is the honest fallback if this is judged
  not worth the surface.
- **Reuse the session cookie for machine clients.** Does not work: the cookie is `HttpOnly` +
  `SameSite=Strict` and minted by a browser login flow. An Asyar extension or a cron job has no
  browser to perform it.
- **OAuth-style device pairing** (client shows a code, operator approves it in the HUD). Nicer
  onboarding and no clipboard hop, but materially more machinery — a pairing endpoint, a pending-
  request store, a polling loop on the client. Worth revisiting if the number of clients grows;
  overkill for the current handful. The copy-a-token flow is also what users already recognise from
  GitHub and Home Assistant.
- **Per-token scopes** (read-only vs. can-spawn-sessions). Genuinely desirable — the Asyar extension
  is strictly read-only and should not be able to spawn a session — but it is a separable follow-up
  and would hold this up. Filed as a note below rather than as scope here.

## Out of scope / follow-ups

- Per-token scopes or read-only tokens.
- Multi-user / SSO — the single-operator model from #1079 stands.
- Any change to the cookie/password login path.

## Environment

Filed manually from the `shepherd-asyar-extension` work.

- Shepherd: v1.46.0 (HUD), herdr 0.7.5 (update to 0.8.0 available)
- Core reachable over Tailscale at `https://<node>.<tailnet>.ts.net:7330`
- Client under construction: Asyar launcher extension (`erwins-enkel/shepherd-asyar-extension`),
  read-only, consumes `GET /api/sessions` + `GET /api/holds`

kai-osthoff:
Closes #2082.

Every machine client that talks to Shepherd needs a bearer token, and until now there was **no way
to get one from Shepherd itself** — you had to SSH to the host and read `SHEPHERD_TOKEN` back out of
the unit file or `.env`. This adds **Settings → Access**: mint a named token, copy it once, revoke it
later, one client at a time.

`SHEPHERD_TOKEN` keeps working, unchanged. Minted tokens are an *additional* accepted credential.

## How it works

A third branch in the existing `checkAuth` seam (`src/server.ts`), after the session cookie and the
env token:

1. Shape-check the `Authorization` header on the public `shp_` prefix — so a request carrying a
   cookie, the env token, or nothing at all never even builds the service.
2. `sha256(presented)` → lookup in an in-memory `Map`.
3. Miss, or past `expiresAt` ⇒ fall through to `401`.

The map is seeded from SQLite once at construction and mutated in place by mint/revoke, so
verification never reads the DB on the hot path (the docstring in `operator-auth.ts` insists on
that — the same Bun loop pumps the web terminal), **and a revocation takes effect on the very next
request, with no restart**.

At rest we store `sha256` of the plaintext plus a 4-character `hint` for the list — never the value.
The plaintext exists exactly once, in the mint response.

### Why no `timingSafeEqual` on this path

`isAuthorized()` compares the **secret itself**, so it must be constant-time, and that path is
untouched. Here the compared value is a **hash** of the secret: recovering the stored hash through
lookup timing buys nothing, since forging a token would still need a SHA-256 preimage. Same
construction as GitHub PATs and Django's token auth. Documented at the top of `src/access-tokens.ts`.

### Token management needs a browser session, not just a bearer

The three `/api/access-tokens` routes sit behind `requireOperatorSession()` — a valid **session
cookie**, not merely an accepted credential. A leaked machine token therefore cannot mint itself
successors or revoke your other tokens. `SHEPHERD_TOKEN` gets no exemption here either. The
auth-exempt agent ingress can't reach these routes at all (its route allowlist), which is locked by a
test, since widening `isAgentIngressRoute` would otherwise be a silent privilege escalation.

## Decisions the operator made

| Decision | Chosen |
| --- | --- |
| IA | Own top-level **Access** section — "Device" is browser-local (theme, push); tokens are server-side |
| Expiry | Optional: `Never` (default) / 30 / 90 / 365 days |
| Hash | SHA-256 — argon2id costs ~100 ms and forbids a map lookup, on a per-request path |
| End of life | Revoke deletes the row; **expired rows stay visible** with a badge, so a client's 401 has an explanation |
| Reveal | A card above the list, not a modal over the Settings modal |
| Self-management | Session cookie only |
| Reach | Full parity with `SHEPHERD_TOKEN`, scopes filed as a follow-up |

## Known limitation, deliberately

A v1 token carries the operator's full reach — **including `/events` and `/pty/:id`**. That falls out
of the shared `checkAuth` seam, and anything less would make minted tokens a non-equivalent
replacement for the env var (a half-built scope system by the back door). The mitigations shipped
here are attribution (each token is named) and instant revocation; the mint form says so out loud and
links to the structural fix, **#2083 (per-token scopes)**, filed alongside this PR. A test pins the
current WS behaviour so a later scope change is a deliberate edit, not a drift.

## Two things worth a reviewer's eye

- **The section glyph is `◈`, not the obvious `⚿`.** `fc-list` says U+26BF (SQUARED KEY) exists only
  in *Noto Sans Symbols 2*, which macOS and Windows don't ship — it would render as tofu, or from a
  different family than its siblings. Every other rail glyph (`▦ ⌁ ⇥ ✦ ⌖ ◫ ⚠`) resolves from the
  general-purpose fonts, and so does `◈`.
- **The `access-token` glossary entry is `internal`, not `external`.** German Wikipedia has no
  counterpart to `en:Access_token` (and the English article is largely about the Windows security
  object), so an external entry could not carry a resolving slug in both locales as the gate requires.
  The definition describes the token as *Shepherd* mints it, which is the honest framing anyway.

## Verification

- `bun run lint` clean; `bun run test` → **8631 pass, 0 fail**.
- `cd ui && bun run check` → 0 errors; `bun run check:i18n` → 2 locales in parity;
  `bun run test` → **292 files, 4595 tests, exit 0**.
- `check-glossary.mjs`, `check-feature-catalog.sh`, `check-announcement-versions.mjs` all pass.
- fallow's audit gate flagged two real findings in this diff during review — `handleAccessTokens` at
  23 cyclomatic, and an export left unused by that refactor. Both are fixed in `8cd16649`, not
  suppressed.
- One unrelated flake seen once: `StatusPip.browser.test.ts`'s hover/tooltip test. It passes in
  isolation and on a full re-run; nothing in this diff touches `StatusPip`.

New tests: `test/access-tokens.test.ts` (28), `test/server-access-tokens.test.ts` (15),
`SettingsAccessPanel.browser.test.ts` (13), plus five additions to `test/server-auth.test.ts`
covering the acceptance criteria directly — a minted token authenticates `/api/sessions` and
`/api/holds`; revoking 401s the next request with no restart; `SHEPHERD_TOKEN` still works alongside.

## Docs

`docs/external-task-api.md` and the Capture-extension page now name the UI as the recommended way to
get a token, with the env var documented as the deployment-provisioned alternative. The Hermes `curl`
is unchanged.


kai-osthoff:
# Review

## Summary

One blocking documentation defect remains in the Settings component.

## Findings

1. ui/src/lib/components/Settings.svelte: The comment says that all seven panels remain mounted, but this change adds `settings-panel-access` alongside the prior seven panels. Update the count (or make the comment count-free) so it truthfully describes the component. The claim is at `ui/src/lib/components/Settings.svelte:427`; the eight mounted panel roots are at `ui/src/lib/components/Settings.svelte:431`, `ui/src/lib/components/Settings.svelte:448`, `ui/src/lib/components/Settings.svelte:459`, `ui/src/lib/components/Settings.svelte:481`, `ui/src/lib/components/Settings.svelte:498`, `ui/src/lib/components/Settings.svelte:514`, `ui/src/lib/components/Settings.svelte:534`, and `ui/src/lib/components/Settings.svelte:544`.

## Verification notes

- The new handler is registered in the normal dispatch chain at `src/server.ts:7667` and is reached only after the shared auth and origin guards at `src/server.ts:7783` and `src/server.ts:7789`. Its session-only management guard is implemented at `src/server.ts:919` and `src/server.ts:929`; the bearer verification and immediate in-memory revocation update are at `src/server.ts:752`, `src/access-tokens.ts:190`, and `src/access-tokens.ts:196`.
- The database table stores `tokenHash` and no plaintext-value column at `src/store.ts:1594`; API summaries deliberately omit that hash at `src/access-tokens.ts:48` and `src/access-tokens.ts:52`.
- The UI API client paths and response shapes agree with the server collection/item routes: `ui/src/lib/api.ts:669`, `ui/src/lib/api.ts:677`, and `ui/src/lib/api.ts:688` versus `src/server.ts:919` through `src/server.ts:936`.
- The added Access labels, glossary keys, and announcement keys are present in both catalogs at `ui/messages/en.json:3575` through `ui/messages/en.json:3612` and `ui/messages/de.json:3575` through `ui/messages/de.json:3612`; the corresponding panel and section wiring are at `ui/src/lib/components/Settings.svelte:534` and `ui/src/lib/settings-search.ts:3`.

## Limitations

- Could not verify that the externally hosted follow-up issue exists: the panel constructs and links to issue 2083 at `ui/src/lib/components/settings/SettingsAccessPanel.svelte:31` and `ui/src/lib/components/settings/SettingsAccessPanel.svelte:173`, but that GitHub state is not present in this repository.

kai-osthoff (commented):
# Review

Decision: comment

## Summary

No blocking findings in the specified diff. The minted-token service derives and stores a SHA-256 hash, returns a hash-free summary, checks expiry in its in-memory index, and removes the index entry on revoke (`src/access-tokens.ts:52`, `src/access-tokens.ts:170`, `src/access-tokens.ts:201`, `src/access-tokens.ts:187`). The server wires that verifier into the shared auth seam while retaining the existing environment-token branch, and the management routes require a valid operator cookie (`src/server.ts:737`, `src/server.ts:746`, `src/server.ts:752`, `src/server.ts:772`, `src/server.ts:919`).

## Findings

None.

## Verification

- The persistence contract contains only `tokenHash` and metadata, with no plaintext column; the list query reads that hash solely for the service and the API summary omits it field-by-field (`src/store.ts:1662`, `src/store.ts:1667`, `src/access-tokens.ts:55`).
- The UI/API contract agrees on the collection and item routes and on the `AccessToken` fields (`src/server.ts:919`, `ui/src/lib/api.ts:669`, `ui/src/lib/api.ts:673`, `ui/src/lib/api.ts:687`, `ui/src/lib/types.ts:89`). The panel reveals only the POST result in component state, supports clipboard copying, and removes that state on dismissal (`ui/src/lib/components/settings/SettingsAccessPanel.svelte:43`, `ui/src/lib/components/settings/SettingsAccessPanel.svelte:79`, `ui/src/lib/components/settings/SettingsAccessPanel.svelte:95`, `ui/src/lib/components/settings/SettingsAccessPanel.svelte:182`).
- Locale parity was checked for every new Access, glossary, and announcement key in both catalogs (`ui/messages/en.json:3575`, `ui/messages/de.json:3575`). The access section is included in both the settings navigation/search model and the mounted panel (`ui/src/lib/settings-search.ts:8`, `ui/src/lib/settings-search.ts:235`, `ui/src/lib/components/Settings.svelte:327`, `ui/src/lib/components/Settings.svelte:534`).
- The earlier comment-count point is addressed: the comment is now count-free while the Access panel is mounted with the other panels (`ui/src/lib/components/Settings.svelte:427`, `ui/src/lib/components/Settings.svelte:534`).
- The requested documentation now recommends Settings → Access while retaining the deployment-provisioned environment-token alternative (`docs/external-task-api.md:57`, `docs/external-task-api.md:69`, `docs-site/src/content/docs/capture-extension.md:30`, `docs-site/src/content/docs/reference/configuration.md:23`).

## Limitations

- Per instruction, no tests, builds, or lint commands were run. I also could not verify the existence of the externally hosted follow-up issue linked by the UI; the local source constructs that link as issue 2083 (`ui/src/lib/components/settings/SettingsAccessPanel.svelte:30`).

PR #2088, opened 2026-08-17
fix(herd): scroll to and mark the session a global jump lands on

kai-osthoff:
## The bug

Confirming a session in the command bar selects it correctly — the terminal on the right proves it — but the left rail never scrolls to that row. What stays on screen is a hovered or attention-washed neighbour that reads like the selection, so the rail appears to contradict the terminal. Reported from ⌘K, but it affected every global jump.

Two defects compounded:

1. **Global jumps never scrolled.** `herd-jump.ts` → `revealAndSelect()` expanded a collapsed epic/stage group and selected, but never brought the row into the rail's scroll viewport. The only code that scrolled was `keyNavSelect()` in `+page.svelte`, reached solely by `j`/`k` — the command-bar path (`oncommandbarsession` → `jumpToSession` → `select: selectUnit`) went around it.
2. **The selected state was too quiet to win a glance.** `.unit.sel` differs from `.unit:hover` by `--sel: #18211e` against `--hover: #0c1110` plus a 1px border. A few RGB steps in the dark theme.

## The fix

`JumpDeps` gains a `revealRow` step, and `revealAndSelect` ends with an unconditional tick + reveal — so all six global jump paths inherit it at once (`jumpToSession`, `selectRundownItem`, `selectFromDeepLink`, `jumpFromHerdrUpdate`, `navigateFromViewport`, `retargetForRepoFilter`, plus `selectNextNeedsYou`). The tick is not optional: every handler mutates the lens and repo/status filters *before* selecting, and the reveal reads the DOM.

The row also carries its weight visually now: the status rule the row already draws widens from 1px to 3px when selected (width, not hue, carries the cue — WCAG 1.4.1), and a jump outlines the row in its own `--rule` colour for ~1.5s.

Scrolling happens only when the row is not already fully visible — an operator who jumped to something on screen should not have the list yanked under them. `j`/`k` and rail clicks are untouched: the operator's own cursor walk or pointer already told them where they landed.

## Decisions confirmed with the operator

| Decision | Chosen |
| --- | --- |
| Scroll position | Centre only when off-screen |
| Emphasis | Permanent accent **and** a ~1.5s jump flash |
| Scope | All global jumps; `j`/`k` and clicks unchanged |
| Flash colour | `var(--rule)` — the row's own status colour |
| Re-flash on an unchanged selection | Yes |

## Verification

- `cd ui && bun run check` — 5955 files, 0 errors
- `cd ui && bun run test` — 296 files, 4653 tests, all pass
- `bun run test` (root) — 8630 pass, 0 fail
- `bun run lint` (root) — clean

New coverage:

- `herd-reveal.test.ts` — the centre-or-leave-alone geometry over plain numbers (fully inside, flush at either edge, clipped top, clipped bottom, entirely off-screen, taller than the viewport).
- `herd-reveal.browser.test.ts` — `scrollParentOf` against real layout: finds the scrolling ancestor, skips an `overflow: auto` box with nothing to scroll, returns null for the mobile flow mode.
- `herd-jump.test.ts` — the `test.each` table now pins `reveal:x` as the final step for every handler, so a jump path added later that skips the reveal fails there. It also pins `keyNav:true:false`, the one path that must not double-scroll.
- `UnitRow.browser.test.ts` / `Herd.browser.test.ts` — the two cues asserted through **computed style**, not class names (3px vs 1px rule, 2px outline vs none), plus the flash landing on exactly one row across the full `Herd → rowCtx → HerdGroup → UnitRow` prop path.

**Not verified in the running app:** port 5174 is held by the operator's own Shepherd instance, and taking it would have killed the Mission Control they are using. The end-to-end click-through is the one success criterion I did not exercise.

## Notes

- Labelled `fix`, not `feat` — this restores behaviour the rail was already meant to have. No feature-catalog entry (per `.claude/rules/ui-feature-catalog.md`, the gate arms on `feat(...)` only). No new user-facing strings, so no i18n catalog change.
- The Done lens (`HerdDoneList.svelte`) keeps its own selection space and is not a global-jump target; left alone deliberately.

### DIFF
--- a/ui/src/lib/components/Herd.browser.test.ts
+++ b/ui/src/lib/components/Herd.browser.test.ts
@@ -1027,3 +1027,34 @@ describe("Herd stage-explainer info icon", () => {
     expect(getComputedStyle(info!).textTransform).toBe("none");
   });
 });
+
+// jumpFlashId is the page's post-jump "look here" signal. It has to survive the whole
+// prop path (Herd → rowCtx → HerdGroup → UnitRow) and land on exactly ONE row —
+// a flash on the wrong row would restate the very confusion the flash exists to end.
+describe("Herd jump flash", () => {
+  const flashed = (container: HTMLElement) =>
+    [...container.querySelectorAll("[data-unit-id]")]
+      .filter((el) => getComputedStyle(el).outlineStyle !== "none")
+      .map((el) => el.getAttribute("data-unit-id"));
+
+  it("outlines only the jumped-to row", async () => {
+    const screen = await render(Herd, {
+      ...base,
+      sessions: [session({ id: "a" }), session({ id: "b" }), session({ id: "c" })],
+      git: {},
+      selectedId: "b",
+      jumpFlashId: "b",
+    });
+    await expect.poll(() => flashed(screen.container)).toEqual(["b"]);
+  });
+
+  it("no jumpFlashId: no row is outlined, selection alone stays quiet", async () => {
+    const screen = await render(Herd, {
+      ...base,
+      sessions: [session({ id: "a" }), session({ id: "b" })],
+      git: {},
+      selectedId: "b",
+    });
+    await expect.poll(() => flashed(screen.container)).toEqual([]);
+  });
+});
--- a/ui/src/lib/components/Herd.browser.test.ts
+++ b/ui/src/lib/components/Herd.browser.test.ts
@@ -1057,4 +1057,23 @@ describe("Herd jump flash", () => {
     });
     await expect.poll(() => flashed(screen.container)).toEqual([]);
   });
+
+  it("a flash whose row is no longer selected goes dark instead of stranding an outline", async () => {
+    // The selection can move out from under a running flash: a click or j/k inside the
+    // ~1.5s window, or the page reassigning selectedId when a session is decommissioned
+    // or its merged rows are cleared. The loudest cue in the rail must not stay parked on
+    // a row the terminal is no longer showing — that IS the bug this path fixes.
+    const props = {
+      ...base,
+      sessions: [session({ id: "a" }), session({ id: "b" })],
+      git: {},
+      selectedId: "a",
+      jumpFlashId: "a",
+    };
+    const screen = await render(Herd, props);
+    await expect.poll(() => flashed(screen.container)).toEqual(["a"]);
+
+    await screen.rerender({ ...props, selectedId: "b" });
+    await expect.poll(() => flashed(screen.container)).toEqual([]);
+  });
 });
--- a/ui/src/lib/components/Herd.svelte
+++ b/ui/src/lib/components/Herd.svelte
@@ -43,6 +43,7 @@
   let {
     sessions,
     selectedId,
+    jumpFlashId = null,
     nowMs,
     onselect,
     onnew,
@@ -107,6 +108,10 @@
   }: {
     sessions: Session[];
     selectedId: string | null;
+    // The session a global jump (command bar, deep link, notification) just landed on —
+    // its row flashes for ~1.5s so the operator's eye finds it after the rail scrolls.
+    // Null at rest; the page owns the timer.
+    jumpFlashId?: string | null;
     nowMs: number;
     onselect: (id: string) => void;
     onnew: () => void;
@@ -412,6 +417,7 @@
   // the parent's props/derivations.
   const rowCtx = $derived<HerdRowCtx>({
     selectedId,
+    jumpFlashId,
     nowMs,
     onselect,
     git,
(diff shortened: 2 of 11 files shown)


kai-osthoff:
# Review: request changes

## Findings

1. ui/src/routes/+page.svelte: A normal rail selection during the 1.5-second jump flash leaves the old row outlined even after the terminal and `.sel` state move to the newly clicked session. `revealRow()` records the old id through `startJumpFlash()` (ui/src/routes/+page.svelte:1517-1525), while `selectUnit()` only changes `selectedId` and never clears or updates `jumpFlashId` (ui/src/routes/+page.svelte:1336-1343). The render path independently applies `.sel` from `selectedId` and `.jump-flash` from `jumpFlashId` (ui/src/lib/components/herd/HerdGroup.svelte:112-115; ui/src/lib/components/UnitRow.svelte:791-795), and the latter has a conspicuous 2px outline (ui/src/lib/components/UnitRow.svelte:1256-1265). Thus Command-K to A followed immediately by clicking B visibly emphasizes A while displaying B in the terminal—the ambiguity this change is intended to remove. Clear the pending flash when a non-jump selection changes session, or otherwise tie the flash to the current selected id.

## Verification notes

The global-jump core waits for the render flush before calling the new DOM reveal callback (ui/src/lib/components/herd-jump.ts:63-82). The new prop is passed from the page through Herd's row context (ui/src/routes/+page.svelte:2820-2829; ui/src/lib/components/Herd.svelte:418-421) and reaches each lifecycle, epic, and experiment `UnitRow` through the shared `HerdGroup` context (ui/src/lib/components/Herd.svelte:718-744; ui/src/lib/components/herd/HerdGroup.svelte:110-116). The tests verify only the aligned case where `selectedId` and `jumpFlashId` are both `b` (ui/src/lib/components/Herd.browser.test.ts:1040-1058); they do not cover a subsequent ordinary selection while the timer is active.

kai-osthoff (commented):
## Review result: comment

No blocking findings in the requested diff.

## Verification

- Command-bar session confirmation invokes `jumpToSession` at `ui/src/routes/+page.svelte:3334`, which delegates to the shared jump handler at `ui/src/routes/+page.svelte:1350`. That handler receives the page's `revealRow` dependency at `ui/src/routes/+page.svelte:1702` and `ui/src/routes/+page.svelte:1716`.
- The shared core selects, awaits a render tick, and then invokes `revealRow` at `ui/src/lib/components/herd-jump.ts:63` through `ui/src/lib/components/herd-jump.ts:82`. The handler table and its tests cover the global routes, including the no-preemptive-scroll next-needs-you case at `ui/src/lib/components/herd-jump.ts:106` through `ui/src/lib/components/herd-jump.ts:135` and `ui/src/lib/components/herd-jump.test.ts:188` through `ui/src/lib/components/herd-jump.test.ts:227`.
- `revealRow` resolves the row with an escaped session id, compares its bounds to the actual scroll-parent viewport, centres only when clipped, and starts the timed flash at `ui/src/routes/+page.svelte:1505` through `ui/src/routes/+page.svelte:1539`. The imported helpers exist and use the same `top`/`bottom` rectangle contract at `ui/src/lib/components/herd-reveal.ts:7` through `ui/src/lib/components/herd-reveal.ts:34`; their geometry cases are specified at `ui/src/lib/components/herd-reveal.test.ts:8` through `ui/src/lib/components/herd-reveal.test.ts:32`.
- The permanent selected-row cue is a 3px status-rule accent at `ui/src/lib/components/UnitRow.svelte:1220` through `ui/src/lib/components/UnitRow.svelte:1238`; the temporary outline is defined at `ui/src/lib/components/UnitRow.svelte:1256` through `ui/src/lib/components/UnitRow.svelte:1265`. `Herd` accepts and forwards the nullable flash id via its derived row context at `ui/src/lib/components/Herd.svelte:43` through `ui/src/lib/components/Herd.svelte:46` and `ui/src/lib/components/Herd.svelte:418` through `ui/src/lib/components/Herd.svelte:421`.

## Prior finding re-check

The previous concern, `ui/src/routes/+page.svelte: Clear or bind jumpFlashId on ordinary selection so a prior row cannot remain outlined.`, is addressed. Ordinary selection updates `selectedId` at `ui/src/routes/+page.svelte:1336` through `ui/src/routes/+page.svelte:1343`, and the renderer binds the outline to both the current `jumpFlashId` and the current selected id at `ui/src/lib/components/herd/HerdGroup.svelte:112` through `ui/src/lib/components/herd/HerdGroup.svelte:124`. Therefore an ordinary selection away from a still-flashing row removes that row's outline immediately; the corresponding rendered-prop test re-renders selection from `a` to `b` and expects no outline at `ui/src/lib/components/Herd.browser.test.ts:1061` through `ui/src/lib/components/Herd.browser.test.ts:1077`.

No builds or tests were run, per review instructions.

kai-osthoff:
🎉✅ PR still READY 🥱⏰ — anyone brave enough to hit merge? 🙃🚀👀

PR #2089, opened 2026-08-17
feat(access-tokens): per-token scopes for minted access tokens (#2083)

### LINKED ISSUE #2083: Per-token scopes for minted access tokens (read / submit / full)

kai-osthoff:
Closes #2083.

#2082 shipped named machine tokens at **full parity with `SHEPHERD_TOKEN`** — every gated surface, the `/pty/:id` terminal upgrade included. This gives each token a **scope**, chosen at mint and fixed for its life.

## The design call

`checkAuth(req, deps)` in `src/server.ts` is already the single auth seam, and it is already called by **both** entry points — `makeApp.fetch` for the HTTP `ROUTE_HANDLERS` chain and `serve().fetch` for the `/events` + `/pty/:id` upgrades. It receives the whole `Request`, so it can see method and pathname. The scope check goes **inside it**, which is what makes the WS upgrades — the hole the issue is actually about — covered by one edit with no second seam to forget.

The policy lives in a new pure module `src/token-scopes.ts` as a **deny-by-default allowlist**: `read` and `submit` name their routes, and **anything unlisted requires `full`**. The issue warns that getting this map wrong "fails **open**" — with an allowlist it can't. A route added next month is `full`-only until someone deliberately edits the table, and a mistake surfaces as a loud 403 rather than silent over-reach.

Rejected: a separate `checkScope` seam (two call sites to remember; forgetting one fails open) and per-handler scope annotations (~100 handlers, and the two WS upgrades aren't handlers at all).

## The policy

| Method | Path | read | submit | full |
| --- | --- | :-: | :-: | :-: |
| `GET` | `/api/sessions` | ✓ | ✓ | ✓ |
| `GET` | `/api/holds` | ✓ | ✓ | ✓ |
| `GET` | `/api/git` | ✓ | ✓ | ✓ |
| `GET` | `/api/me` | ✓ | ✓ | ✓ |
| `POST` | `/api/ping` | ✓ | ✓ | ✓ |
| `GET` | `/events` | ✓ | ✓ | ✓ |
| `POST` | `/api/sessions` | — | ✓ | ✓ |
| `GET` | `/api/held` | — | ✓ | ✓ |
| `POST` | `/api/held/:id/spawn` | — | ✓ | ✓ |
| `PATCH` | `/api/held/:id` | — | ✓ | ✓ |
| `DELETE` | `/api/held/:id` | — | ✓ | ✓ |
| `POST` | `/api/uploads` | — | ✓ | ✓ |
| `POST` | `/api/issues` | — | ✓ | ✓ |
| `GET` | `/pty/:id` | — | — | ✓ |
| any | **everything else** | — | — | ✓ |

`POST /api/ping` sits in `read` despite the verb: `handlePing` is a pure no-op reachability probe (a POST only so it rides `checkOrigin`), so it reads nothing and writes nothing.

**The levels are drawn from the clients that actually exist, not from intuition:**

- `shepherd-asyar-extension` (the driver for #2082) reads `/api/sessions` + `/api/holds` → **`read`**.
- Shepherd's own Capture extension (`extension/src/lib/transport.ts`) posts to `/api/ping`, `/api/uploads`, `/api/sessions`, `/api/issues` → **`submit`**.

After this change **nothing Shepherd ships needs `full` unless it wants a terminal** — which is the test of whether three coarse levels are drawn in the right place, and why `submit` covers uploads and issues rather than forcing our own client to the widest level.

## Acceptance criteria

| Criterion | Where it's locked |
| --- | --- |
| `read` → 200 on `GET /api/sessions`, 403 on `POST /api/sessions` | `test/server-auth.test.ts` |
| `read` refused at the `/pty/:id` upgrade | same, via `withServer` against a **real** session id, so the 403 can't be a 404-in-disguise |
| `full` behaves exactly as a v1 token; `SHEPHERD_TOKEN` unscoped | same — both asserted explicitly, including the terminal |
| Pre-#2083 tokens keep working (migrate to `full`) | `test/store.test.ts` migration case over a raw #2082 table |
| Scope visible per row, not changeable after minting | `SettingsAccessPanel.browser.test.ts` (+ a server test that PATCH/PUT/POST on a token 404) |
| The "planned follow-up" note replaced by the picker | `settings_access_scopes_link` is gone from both catalogs |

Plus: `submit` reaches all four Capture routes and the held routes but not `/pty` or `/api/sessions/:id/reply`; a stored scope outside the three grants **nothing** (403 even on `GET /api/sessions`); a cookie-authed operator never enters the scope branch.

## Decisions worth a look

- **An unrecognized stored scope grants nothing** — not "treat as read", not "treat as full". A hand-edited `scope = 'wat'` row 403s everywhere, by construction rather than by a check. `verify()` passes the stored value through verbatim; laundering it in the service would kill that rule.
- **`lastUsedAt` is stamped even when the scope check refuses.** The credential authenticated; only the route was out of reach. A misconfigured client hammering a forbidden route should still read as active — that's the attribution #2082 shipped the stamp for.
- **Two deliberately different defaults.** The mint form starts on `read` (least privilege for the operator clicking through); the mint **API** defaults an absent `scope` to `full` (mirrors the column default, so a `{name}`-only call is unchanged). Not a contradiction — the form always sends an explicit scope. Both are tested.
- **The field is `scope`, singular**, so the pre-existing test asserting `scopes: [...]` is rejected as an unknown field still passes untouched.

## Two things found while implementing

1. **The Capture extension would have reported the wrong fault.** `kindForStatus` mapped every 403 to `"origin"`, whose copy points at `SHEPHERD_ALLOWED_HOSTS`. A `read`-scoped Capture token now also produces a 403 — so the operator would have been sent on a detour past the real cause. Split by response body into a new `"scope"` kind with copy naming the Submit scope (`fix(extension)` commit). Both docs pages now say which scope to pick, so nobody follows a setup page into a silently broken client.
2. **A CSS-ordering trap in my own first draft.** `.badge` and `.scope-badge` are both single-class selectors, so only source order separates them — with `.scope-badge` above `.badge`, every scope inherited `.badge`'s amber and the column distinguished nothing, while still rendering and still passing every other assertion. Fixed, commented at the rule, and locked by a browser test that asserts the two computed colours differ.

## Reviewer asks

- **Per `.claude/rules/ui-glossary.md`, the new `token-scope` glossary definition needs explicit confirmation** (EN + DE) — no gate can catch a misleading definition. It's the longest new string in the PR.
- The DE copy generally: three scope descriptions, the fixed-scope note, the What's-New body, and the extension error string. Nothing checks prose.

## Deliberately not done

- `SHEPHERD_TOKEN` stays unscoped (the issue's own out-of-scope list) — it's the break-glass credential with no UI to choose in.
- No route to edit a scope after minting. The AC asks for this, and an editable scope makes "what could this credential do last Tuesday" unanswerable.
- `GET /api/me` still returns `{authenticated: true}` — it does not report the caller's own scope. Not asked for.
- `docs-site/.../reference/glossary.md` is **not** updated. That page is already behind `ui/src/lib/glossary.ts` (it never got #2082's `access-token` entry either), and closing a pre-existing drift isn't this PR's job — flagging it rather than silently widening the diff.

## Verification

| Package | Result |
| --- | --- |
| Root | `bun run lint` ✅ · `bun run typecheck` ✅ · `bun run test` → **8656 pass, 0 fail** |
| `ui/` | `bun run check` → 0 errors · `check:i18n` → 2 locales in parity · `bun run test` → **4644 pass** |
| `extension/` | `bun run check` → 0 errors · `check:i18n` → parity · `bun run test` → **88 pass** |
| `docs-site/` | `bun run check` → 0 errors, 0 warnings |
| Gates | `check:glossary` ✅ · `check:announcement-versions` ✅ · `check-feature-catalog.sh` ✅ |


kai-osthoff:
# Review: request changes

## Findings

- ui/src/lib/components/settings/SettingsAccessPanel.svelte:35: `scopeLabel` maps every value other than `read` and `submit` to the Full label. This is reachable for a hand-edited or forward-version database row: `AccessTokenService` loads and returns the stored `scope` verbatim (`src/access-tokens.ts:182-188`, `src/access-tokens.ts:192-193`), while the authorization policy deliberately denies every route for an unrecognised value (`src/token-scopes.ts:102-112`). Consequently a token whose persisted scope is `wat` is shown as Full in Settings even though it has no reach, contradicting the list's audit purpose and making an operator diagnose it incorrectly. Render an explicit unknown/no-access state, or validate/coerce the value consistently before exposing it.

## Verification

The new route policy is deny-by-default for `read` and `submit`, with the intended read and submit route sets defined at `src/token-scopes.ts:51-82`; the auth seam applies it only after successful minted-token verification at `src/server.ts:759-773`, leaving cookie and `SHEPHERD_TOKEN` paths before it unchanged at `src/server.ts:747-755`. The database migration adds `scope` with a `full` default at `src/store.ts:1595-1603`, and the list query includes that field at `src/store.ts:1673-1679`.

I also confirmed locale parity for the new UI copy: the scope keys and glossary definition exist in both `ui/messages/en.json:3588-3595,3619-3620` and `ui/messages/de.json:3588-3595,3619-3620`; the extension's new scope error exists in both `extension/messages/en.json:23` and `extension/messages/de.json:23`. No build, lint, or tests were run, per the review instruction.

kai-osthoff:
**The critic is right, and I merged before its review landed — my sequencing error, not a disagreement.**

Finding: `SettingsAccessPanel.svelte` → `scopeLabel()` falls through to **Full** for any scope outside the three levels. Since `toSummary()` passes `row.scope` through verbatim (deliberately — `verify()` must not launder a bad stored value, or the deny-by-default rule dies), a row with `scope = 'wat'` reaches the UI and its badge reads **Full** while `scopeAllows()` grants it **nothing**. The badge is the audit surface, so it states the exact opposite of the truth. Valid defect.

I cannot fix it in this PR — it merged at 11:01:04Z (`13b475c1`), and this session tracks exactly one PR, so I am not opening a second one. Filed as a follow-up issue instead; linked below.

Scope of the fix, for whoever picks it up:
- `scopeLabel()` needs an explicit `full` arm plus an "unknown" fallback (new EN+DE key), not a catch-all that names the widest level.
- `scopeHint()` has the same ternary shape but is only ever called with the picker's own `scope` state, which is always one of the three — it does **not** need the same treatment. Keep the fix minimal.
- The badge *styling* is already correct: `.full` is applied only on `t.scope === "full"`, so an unknown scope already renders neutral rather than amber. Only the label text is wrong.

kai-osthoff:
Follow-up filed: #2090 (one-line label fix + one browser test).

PR #2093, opened 2026-08-18
fix(store): batch schema bootstrap into one transaction

scoop:
## The flake

`verify` on #2092 failed three tests. All three are in `test/store.test.ts`, all three are
**timeouts against bun's 5000 ms default** — not one assertion failure among them:

| CI duration | Test |
| --- | --- |
| 7101 ms | `repo_config migration: an old row without previewOpenMode gains the ask default` |
| 11277 ms | `session: legacy row (egressApplied/egressDegraded columns defaulting 0) reads as false` |
| 6428 ms | `session: landingRepair migration — pre-existing row without column reads as false` |

They share one shape: `mkdtempSync` a real directory, seed a legacy-schema SQLite file, then
construct `SessionStore` against that **on-disk** path so the migration runs for real.

Pulling per-test timings out of the CI log shows those three are just the unlucky tail. **Every**
disk-backed store construction on the runner costs 2–6 s, and a dozen siblings sit right under the
cliff: 4611 ms (`first-run migration: pre-existing settings (cookieSecret) …`), 4452 ms
(`session: research migration …`), 4341 ms (`migration: store opened on pre-existing DB without
repoMode column …`), 4126 ms (`session migration: a pre-epicParent row …`), 3849 ms, 3768 ms, and
~14 more between 2.3 s and 3.7 s.

`main` is green by luck, not by margin. Any PR can lose this dice roll; #2092 lost it while
carrying nothing but dev-dependency bumps.

## Cause

The `SessionStore` constructor issues **150 separate write statements** on a fresh open. Counted by
instrumenting `Database.prototype.run` around one construction, rather than by grepping:

| Kind | Count |
| --- | --- |
| `CREATE TABLE IF NOT EXISTS` | 43 |
| `CREATE INDEX` / `CREATE UNIQUE INDEX` | 12 |
| `ALTER TABLE … ADD COLUMN` behind `PRAGMA table_info` probes | 88 |
| seed `INSERT` | 3 |

The `ALTER TABLE` bulk fires even on a fresh DB: the baseline `CREATE TABLE IF NOT EXISTS`
statements deliberately hold the original schema, and later columns arrive through
`addMissingColumns()`, `migrateSessionColumns()`, `migrateRepoConfigColumns()` and eleven siblings.
The same instrumentation confirms exactly one `BEGIN`/`COMMIT` pair wraps all 150.

Each one was left to autocommit, and under SQLite's default rollback journal every implicit
transaction means create-journal → fsync → write → fsync → delete-journal → fsync-directory. On
disk that is 150 fsync storms; in memory it is free. Measured locally on a fast NVMe: **310 ms**
per fresh on-disk construct vs **7 ms** in-memory. The runner's slower disk stretches that into
seconds.

## Fix

Bracket the bootstrap in one transaction, so 150 implicit transactions collapse into a single
commit.

| Measurement | Before | After |
| --- | --- | --- |
| Fresh on-disk `new SessionStore(path)` | 310 ms | **8.6 ms** |
| `bun test ./test/store.test.ts` | 6.04 s | **0.90 s** |
| Full root suite (local) | — | **8681 pass / 0 fail** (46 s) |

And on the CI runner itself, the three tests that failed on #2092:

| Test | Before | After |
| --- | --- | --- |
| `repo_config migration: … previewOpenMode gains the ask default` | 7101 ms (timeout) | **20.3 ms** |
| `session: legacy row (egressApplied/egressDegraded …) reads as false` | 11277 ms (timeout) | **19.5 ms** |
| `session: landingRepair migration — pre-existing row …` | 6428 ms (timeout) | **19.0 ms** |

Disk-backed store tests previously held 15+ of the slowest slots on CI at 2.4–4.7 s each and now
appear nowhere in the top 8. The full root suite on CI dropped **206.6 s → 80.1 s** — a side effect
rather than the goal, but it applies to every run from here on.

At ~9 ms per construct the slowest disk-backed test lands roughly 500x under the timeout. The
flake class is gone with margin, not nudged.

Schema migration also becomes **atomic**, which is a real behaviour change worth calling out: a
crash mid-boot now leaves the last consistent schema instead of a half-migrated DB that the next
boot continues from. Safer, and either way the boot failed.

## Why this shape

Two statements plus comments, deliberately **not** an extracted `bootstrapSchema()` method.
Extraction would be more idiomatic here (14 existing `this.db.transaction(...)` call sites) and
would give an explicit `ROLLBACK`, but it re-indents ~480 lines of a constructor that every
feature branch appends a `CREATE TABLE` to — a whitespace-only merge-conflict grenade for no
functional gain. The trade-off accepted instead: a throwing bootstrap discards the transaction
when the handle is finalized rather than via an explicit `ROLLBACK`. Both production construction
sites (`src/index.ts`, `scripts/sweep-house-rules.ts`) are unguarded, so a throw kills the process
regardless and no caller ever holds a half-built store.

`PRAGMA journal_mode = WAL` would fix the symptom too (measured: 204 ms → 7.6 ms on a comparable
synthetic DDL workload, vs 5.3 ms for the transaction) but it changes the production on-disk
format — `-wal`/`-shm` sidecars, interactions with the hourly `VACUUM INTO` backup and with
network filesystems. Out of scope for a flake fix.

Raising bun's `--timeout` was the other option. Rejected: one line and zero risk, but it leaves
the suite ~4x slower than it needs to be and leaves the next slow test to flake.

## Correctness

Two constraints keep the bracket valid, both documented at the `BEGIN`:

- `PRAGMA foreign_keys` is a **no-op inside a transaction**, so it must stay above the `BEGIN`. It
  does — it is the constructor's first statement.
- Nothing in the bootstrap may `VACUUM` or set `journal_mode`; neither can run in a transaction.
  Audited: there are none in the constructor or any of its callees.

Also checked:

- `migrateBuildQueueStepsPk()` already opens `this.db.transaction(...)` for its
  create/copy/drop/rename rebuild. `bun:sqlite` runs a nested `db.transaction()` as a `SAVEPOINT`,
  verified empirically — both writes survive the outer `COMMIT`, and the legacy-PK rebuild test
  passes.
- `PRAGMA table_info` reads inside the transaction see the uncommitted `CREATE TABLE`s, which is
  exactly what the conditional `ALTER TABLE` migrations need.
- No `return` or `throw` between the `BEGIN` and the `COMMIT` in the constructor body, so the
  `COMMIT` is unconditionally reached. (Callee early-returns are harmless — they return from the
  method.)
- **Lock contention improves:** one write lock held for ~9 ms instead of 150 taken and released
  over ~310 ms, so concurrent-open contention against the hourly backup's `VACUUM INTO` — the
  reason `busy_timeout = 5000` exists — gets better, not worse.

## No new test

The ~20 existing disk-backed migration tests are the coverage, and their runtime collapse is the
evidence. A timing assertion would reintroduce exactly the hardware-dependent flake being removed,
and an atomicity test needs a fault-injection seam `SessionStore` does not have.

## Verification

`bun run test` (8680 pass / 0 fail / 15 skip), `bun run lint`, `bun run typecheck`,
`prettier --check src/store.ts` — all clean. `ui/` and `extension/` are untouched by this change;
both suites were run anyway and pass (4646/4646 and full).

Unblocks #2092, which is otherwise a clean dev-dep bump.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -1171,6 +1171,19 @@ export class SessionStore implements CapStore, CreditStore, ModelWeekStore {
     // is sub-second, so the worst-case wait is a brief, bounded stall — not a crash. Keep equal to
     // BUSY_TIMEOUT_MS in scripts/backup.ts.
     this.db.run("PRAGMA busy_timeout = 5000");
+    // Everything below is ONE transaction. A fresh open fires 150 write statements: 43
+    // CREATE TABLE IF NOT EXISTS (41 here, REVIEWER_SPAWNS_DDL, and terminal_claims inside
+    // migrateSessionColumns), 12 index creations (11 here plus that method's partial unique
+    // index), 88 ALTER TABLE column adds behind PRAGMA table_info probes, and 3 seed INSERTs.
+    // Each one left to autocommit pays SQLite's full rollback-journal fsync cycle — 310ms per open
+    // on a fast NVMe, seconds on CI's disk, which is what times out the disk-backed migration
+    // tests in test/store.test.ts. Batched: ~9ms. It also makes migration atomic — a crash mid-boot
+    // now leaves the last consistent schema instead of a half-migrated DB.
+    // Two constraints keep this correct: `PRAGMA foreign_keys` is a no-op inside a transaction,
+    // so it MUST stay above the BEGIN (it does); and nothing in here may VACUUM or set
+    // journal_mode, neither of which can run in a transaction. Nested `this.db.transaction(...)`
+    // calls (migrateBuildQueueStepsPk) are fine — bun:sqlite runs those as SAVEPOINTs.
+    this.db.run("BEGIN");
     this.db.run(`CREATE TABLE IF NOT EXISTS sessions (
       id TEXT PRIMARY KEY, desig TEXT NOT NULL, name TEXT NOT NULL, prompt TEXT NOT NULL,
       repoPath TEXT NOT NULL, baseBranch TEXT NOT NULL, branch TEXT,
@@ -1647,6 +1660,11 @@ export class SessionStore implements CapStore, CreditStore, ModelWeekStore {
     if (!usageCols.some((c) => c.name === "claudeSessionId")) {
       this.db.run(`ALTER TABLE session_usage ADD COLUMN claudeSessionId TEXT NOT NULL DEFAULT ''`);
     }
+    // Commits the whole bootstrap (see the BEGIN above). No try/ROLLBACK: a throw here escapes the
+    // constructor, so no caller ever holds this store, and SQLite discards the open transaction
+    // when the handle is finalized. Deliberately no wrapping block — that would re-indent the
+    // ~480 lines above and collide with every branch that appends a table here.
+    this.db.run("COMMIT");
   }
 
   // ── settings (key/value) ─────────────────────────────────────────────────


scoop (commented):
# Review — PR #2092 · `fix(store): batch schema bootstrap into one transaction`

Diff reviewed: `git diff b182aa8e...HEAD` — one file, `src/store.ts`, +16 lines (a `BEGIN`, a `COMMIT`, and two comment blocks).

## Verdict

The mechanism is sound and I could not find a correctness or concurrency defect in it. One factual
error in the new comment (an inventory that is off by ~5×) is the only thing I'd ask to be changed;
it is not blocking, so the decision is `comment`.

**The transaction brackets the whole bootstrap, with no way to skip the `COMMIT`.**
`BEGIN` at `src/store.ts:1184`, `COMMIT` at `src/store.ts:1665`, both at the constructor's own
statement level. I grepped for constructor-level `return` statements (4-space indent) in
`src/store.ts` — the matches jump straight from line 971 to line 1673, so there is no early return
between the two that would leave the transaction dangling. Every early return inside the region
(`src/store.ts:4359` in `migrateBuildQueueStepsPk`, `src/store.ts:4451` in `migrateFirstRunMarker`,
`src/store.ts:5378` in `backfillLearningDiversity`) returns from the *helper*, not the constructor.

**Claim: `PRAGMA foreign_keys` must stay above the `BEGIN`.** Correct, and it does —
`src/store.ts:1167` (`PRAGMA foreign_keys = ON`) and `src/store.ts:1173`
(`PRAGMA busy_timeout = 5000`) both precede `src/store.ts:1184`. `foreign_keys` is documented as a
no-op inside a transaction; `busy_timeout` is a connection setting and is unaffected.

**Claim: nothing in the region may `VACUUM` or set `journal_mode`.** Verified: every `PRAGMA` in
`src/store.ts` is one of `foreign_keys` (1167), `busy_timeout` (1173) or `table_info`
(1155, 1325, 1392, 1492, 1589, 1652, 4263, 4355, 4380, 4389, 4462, 4475, 4507, 4553, 4581, 4598).
No `VACUUM` and no `journal_mode` anywhere in the file. `PRAGMA table_info` inside the transaction
sees this connection's own uncommitted schema changes, so the conditional `ALTER TABLE` migrations
still make correct decisions.

**Claim: the only nested `this.db.transaction(...)` reached from the bootstrap is
`migrateBuildQueueStepsPk`.** Verified. The `db.transaction(` call sites in `src/store.ts` are
2552, 2677, 2978, 3031, 3063, 3088, 4211, 4361, 5335, 5468, 5566, 5672, 6018; of those only
`src/store.ts:4361` is inside a method the constructor calls (`src/store.ts:1460`). I also chased
the three non-DDL helpers the constructor calls for hidden nesting: `migrateFirstRunMarker`
(`src/store.ts:4450`) reaches `getSetting`/`setSetting`/`list` at `src/store.ts:1669`, `1676` and
`2745` — all plain `query`/`run`; `backfillLearningDiversity` (`src/store.ts:5377`) and
`resetOrphanedEpicDraftMaterialize` (`src/store.ts:5704`) are plain `run`s. So the SAVEPOINT
assumption has exactly one consumer, and it is covered on a real legacy DB by
`test/build-queue-store.test.ts:507`.

**The table rebuild inside that nested transaction stays FK-safe.** `migrateBuildQueueStepsPk`
does `DROP TABLE build_queue_steps` + `RENAME` (`src/store.ts:4371-4373`) with FK enforcement on.
The only `REFERENCES` clauses in the schema are `src/store.ts:1198`
(`session_git_cache` → `sessions`) and `src/store.ts:1645`
(`session_usage_bucket` → `session_usage`); nothing references `build_queue_steps`, so the drop
raises no deferred FK violation at the outer `COMMIT`.

**Ordering dependencies inside the transaction are unchanged.** The helpers that *read* tables run
after those tables are created: `migrateFirstRunMarker` (`src/store.ts:1261`) reads `settings`
(created 1247) and `sessions` (created 1185); `backfillLearningDiversity` (`src/store.ts:1449`)
reads `learnings` (1418) and `signals` (1414). Batching does not reorder anything, so this is the
same sequence that already worked.

**Locking.** The transaction is DEFERRED, but the first statement after `BEGIN` is a write
(`CREATE TABLE IF NOT EXISTS sessions`, `src/store.ts:1185`), so the write lock is taken at the
first statement rather than promoted out of an already-held read transaction — i.e. it behaves
like `BEGIN IMMEDIATE` and the `busy_timeout` at `src/store.ts:1173` is honoured, rather than
hitting SQLite's "skip the busy handler to avoid deadlock" promotion case. The external reader this
timeout exists for opens read-only (`scripts/backup.ts:131`, `BUSY_TIMEOUT_MS = 5000` at
`scripts/backup.ts:35`), and a SHARED reader does not block RESERVED; only the single `COMMIT` now
waits for EXCLUSIVE — one bounded wait instead of the ~50 the old autocommit path took, so this is
strictly less exposed to `SQLITE_BUSY`, not more.

**Failure path.** The "a throw escapes the constructor, so no caller holds the store" reasoning
holds for the production path: the sole `new SessionStore(...)` in `src/` is `src/index.ts:242`,
at module top level with no `try`/`catch`, so a bootstrap failure is process-fatal.

**Statement counts.** 42 `CREATE TABLE IF NOT EXISTS` do execute in the constructor body (41 inline
between `src/store.ts:1185` and `src/store.ts:1634`, plus `REVIEWER_SPAWNS_DDL` run at
`src/store.ts:1402`, defined at `src/store.ts:1134`) — that part of the comment checks out. The
index count does not; see the finding.

## Findings

**`src/store.ts`: the new comment says the bootstrap issues "2 indexes"; it issues 11 in the
constructor body.** `CREATE (UNIQUE) INDEX IF NOT EXISTS` statements between the `BEGIN` and the
`COMMIT`: `src/store.ts:1229`, `1236`, `1371`, `1405`, `1411`, `1417`, `1425`, `1443`, `1446`,
`1461`, `1648` — plus a 12th at `src/store.ts:4338` inside `migrateSessionColumns`, which the
constructor calls at `src/store.ts:1196`. The comment (and the commit message) is an inventory of
exactly what now runs inside this transaction, which is what a future maintainer will audit against
when asking "is everything in here transaction-safe?" — so the wrong number should be corrected to
~12. It changes no behaviour: indexes are as transaction-safe as the tables, so the conclusion the
comment draws is still right.

## Nits (non-blocking):

- `src/store.ts:1661-1664` — the no-`try`/`ROLLBACK` rationale is right for `src/index.ts:242`, but it leaves the connection sitting in an open transaction until GC finalizes the handle; any caller that catches the constructor throw and reopens the same file in-process would eat a 5s busy wait first. No such caller exists today (`src/index.ts:242` is the only one in `src/`), so this is a note, not a request.

- `scripts/sweep-house-rules.ts:23` claims "WAL + busy_timeout make that safe — see store.ts", but nothing in the repo sets `PRAGMA journal_mode = WAL`; `src/store.ts` sets only `foreign_keys` (1167) and `busy_timeout` (1173), so the live DB runs on the default rollback journal. The second-writer safety argument there rests on `busy_timeout` alone.

## Stated limitations

- **bun:sqlite's nested-transaction semantics could not be verified against source.** There is no
  `node_modules` in this worktree, and `bun:sqlite` is built into the runtime, so I could not read
  the implementation of `Database.prototype.transaction` to confirm the comment's claim that a
  nested call runs as a `SAVEPOINT` when an outer `BEGIN` is already open. It matches bun's
  documented better-sqlite3-compatible behaviour (`inTransaction` → SAVEPOINT/RELEASE instead of
  BEGIN/COMMIT), and `test/build-queue-store.test.ts:488-520` exercises exactly that path against a
  legacy on-disk DB, so CI covers it — but I am recording it as assumed-as-written rather than
  verified.
- **The performance numbers in the comment and commit message (310ms → ~12ms per open, store.test.ts
  6.04s → 0.90s) are not verifiable here** — this review is read-only and I ran nothing. Nothing in
  the review depends on them.

scoop (commented):
# Re-review — PR #2092 (`fix(store): batch schema bootstrap into one transaction`)

Diff under review: `git diff b182aa8e0ecbdee13d8b55070bfd652431c5ef61...HEAD` → **one file, `src/store.ts`, +18/-0**: `this.db.run("BEGIN")` at `src/store.ts:1186`, `this.db.run("COMMIT")` at `src/store.ts:1667`, and two explanatory comment blocks.

**Decision: comment.** No blocking findings. The single prior-round finding is addressed, and I could not find a correctness or security defect in the two statements or the reasoning they rest on.

---

## Prior finding — status

**1. `src/store.ts`: bootstrap comment said the transaction issues 2 indexes; it issues ~12. — ADDRESSED.**

The comment now reads "12 index creations (11 here plus that method's partial unique index)" (`src/store.ts:1176-1177`). I counted the actual statements rather than trusting the number:

- 11 `CREATE [UNIQUE] INDEX` between `BEGIN` (`src/store.ts:1186`) and `COMMIT` (`src/store.ts:1667`): `src/store.ts:1231`, `:1238`, `:1373`, `:1407`, `:1413`, `:1419`, `:1427`, `:1445`, `:1448`, `:1463`, `:1650`.
- The 12th is the partial unique index inside `migrateSessionColumns` — `src/store.ts:4340` (`sessions_terminal_singleton`), reached from the constructor via `src/store.ts:1198`.

11 + 1 = 12, and the parenthetical attribution ("that method's partial unique index") matches: `src/store.ts:4340-4341` is indeed `CREATE UNIQUE INDEX ... WHERE terminal = 1 AND archivedAt IS NULL`. Inventory is now accurate.

---

## What I verified in the rest of the comment inventory

Since the prior round was about inventory accuracy, I re-derived every number in the block rather than spot-checking the one that was fixed.

**"43 CREATE TABLE IF NOT EXISTS (41 here, REVIEWER_SPAWNS_DDL, and terminal_claims inside migrateSessionColumns)" — correct.**
41 `CREATE TABLE IF NOT EXISTS` occur between `BEGIN` and `COMMIT` (`src/store.ts:1187, 1199, 1203, 1214, 1217, 1223, 1226, 1233, 1240, 1249, 1251, 1264, 1274, 1286, 1293, 1305, 1360, 1377, 1416, 1420, 1428, 1431, 1437, 1455, 1465, 1471, 1490, 1502, 1505, 1513, 1517, 1522, 1536, 1544, 1553, 1564, 1580, 1584, 1608, 1617, 1636`). `REVIEWER_SPAWNS_DDL` (defined `src/store.ts:1134`) is run at `src/store.ts:1404` — inside the region, so the attribution is right that it is a 42nd not counted in the 41 literal-in-place ones. `terminal_claims` is `src/store.ts:4345`, inside `migrateSessionColumns`. 41 + 1 + 1 = 43.

**"88 ALTER TABLE column adds" — correct, exactly.** Counting only the adds that actually fire on a *fresh* open (i.e. `add()` calls whose column is absent from the inline `CREATE TABLE`):

| table | add() calls | already inline | fresh ALTERs |
|---|---|---|---|
| `sessions` | 45 (`src/store.ts:4269-4335`) | 7 (`model, effort, claudeSessionId, providerSessionId, agentProvider, auto, issueNumber` — `src/store.ts:1187-1197`) | 38 |
| `repo_config` | 28 (`src/store.ts:4396-4436`) | 8 (`src/store.ts:1251-1261`) | 20 |
| `learnings` | 15 (`src/store.ts:4605-4632`) | 2 (`promotedPrUrl, ineffectiveSignalIds` — `src/store.ts:1425-1426`) | 13 |
| `epic_completed` | 11 (`src/store.ts:4484-4505`) | 4 (`src/store.ts:1527-1530`) | 7 |
| `epic_run` | 3 (`src/store.ts:1508`) | 0 (`src/store.ts:1505-1507`) | 3 |
| `epic_integrated` | 3 (`src/store.ts:4469-4473`) | 0 (`src/store.ts:1517-1520`) | 3 |
| `reviews` | 13 (`src/store.ts:4513-4541`) | 11 (`src/store.ts:1274-1284`) | 2 |
| `plan_gates` | 8 (`src/store.ts:4560-4572`) | 6 (`src/store.ts:1293-1303`) | 2 |

Total **88**. The remaining migration probes are all no-ops on a fresh DB because the column is inline: `held_tasks.reason` (`src/store.ts:1245`), the seven `recaps` probes (`src/store.ts:1309-1325`), `herd_digests.epicsToLand` (`src/store.ts:1386`), `reviewer_spawns` ×3 (`src/store.ts:1139, 1141, 1149`), `build_queue_state.approvalKind` (`src/store.ts:1466`), `push_subscriptions` ×4 (`src/store.ts:1586-1588`), `access_tokens.scope` (`src/store.ts:1612`), `session_usage.name/claudeSessionId` (`src/store.ts:1620-1621`).

**"3 seed INSERTs" — correct.** `INSERT OR IGNORE INTO task_seq` (`src/store.ts:1212`), `setSetting("firstRunMigrated", "1")` (`src/store.ts:4460`, via `src/store.ts:1263`), `setSetting("learnings:diversity-backfilled", "1")` (`src/store.ts:5397`, via `src/store.ts:1451`). On a genuinely fresh DB `firstRunResolved` is *not* written (`src/store.ts:4454-4459`: no cookieSecret / passwordHash / repoRoot / sessions), so it is exactly 3, not 4.

**"150 write statements" — correct as a total.** 43 + 12 + 88 + 3 = 146; the remaining 4 are unconditional backfill/reset `UPDATE`s that the itemized list does not mention: `src/store.ts:4442` (repo_config `automationConfirmedAt`, gated on the column-missing branch which *does* fire on a fresh DB), `src/store.ts:4547` (reviews `summaryCode`), `src/store.ts:4576` (plan_gates `summaryCode`), and `src/store.ts:5708` via `resetOrphanedEpicDraftMaterialize()` at `src/store.ts:1486`. See the nit below.

---

## Correctness of the two new statements

**COMMIT is unconditionally reached.** There is no `return` or `throw` at constructor top level between `src/store.ts:1186` and `:1667` — the nearest are `src/store.ts:1124` (above) and `:1675` (below). The early returns in the helpers (`src/store.ts:4362`, `:4453`, `:5380`) return from the helper, not the constructor, so they cannot skip the `COMMIT`.

**"`PRAGMA foreign_keys` … MUST stay above the BEGIN (it does)" — verified.** `PRAGMA foreign_keys = ON` is at `src/store.ts:1167`, above `BEGIN` at `:1186`. Relevant because `session_usage_bucket` carries `FOREIGN KEY (sessionId) REFERENCES session_usage(sessionId) ON DELETE CASCADE` (`src/store.ts:1647`) and the cascade is exercised by `deleteSessionUsage`/`replaceUsageBuckets`.

**"nothing in here may VACUUM or set journal_mode" — verified.** The only `VACUUM` in the tree is `scripts/backup.ts:134`, and it runs on a *separate* `new Database(dbPath, { readonly: true })` handle (`scripts/backup.ts:131`) in a separate process. `journal_mode` / `synchronous` / `wal_checkpoint` are set nowhere in `src` or `scripts` (grep over `**/*.ts` returns only unrelated prose matches), which is consistent with the commit message's rollback-journal fsync reasoning.

**"Nested `this.db.transaction(...)` calls (migrateBuildQueueStepsPk) are fine" — the attribution is right.** `migrateBuildQueueStepsPk` (`src/store.ts:4356`, called at `src/store.ts:1461`) is the *only* one of the 13 `this.db.transaction(` sites in the file reachable from the constructor; the others (`src/store.ts:2554, 2679, 2980, 3033, 3065, 3090, 4213, 5337, 5470, 5568, 5674, 6020`) are all public read/write methods invoked after construction. So the nesting surface is exactly the one statement the comment names. That path is not dead: `test/build-queue-store.test.ts:488-523` seeds a legacy global-`PRIMARY KEY (id)` `build_queue_steps` on a **disk** file and then opens a `SessionStore` over it (`test/build-queue-store.test.ts:507`), which now runs `db.transaction(...)` inside the manual `BEGIN`. See the limitation note below on what I could and could not confirm there.

**Locking behaviour is not a regression.** Every other `Database` open in the tree is read-only (`src/codex-usage.ts:141`, `scripts/usage-report.ts:581`, `scripts/backup.ts:131`, `:142`), so there is no second in-process writer that the longer-held RESERVED lock could starve. `busy_timeout = 5000` is still set before `BEGIN` (`src/store.ts:1173`), and the first statement inside the transaction is a write (`CREATE TABLE`, `src/store.ts:1187`), so the deferred `BEGIN` escalates to RESERVED immediately rather than risking a read→write upgrade conflict.

**"a throw here escapes the constructor, so no caller ever holds this store" — verified for today's callers.** The only two constructions outside tests are `src/index.ts:242` (module top level — a throw aborts boot) and `scripts/sweep-house-rules.ts:102` (a script). Neither wraps the construction in a `try`, and no test asserts a throwing constructor (`toThrow` does not appear in `test/store.test.ts`). See `Latent / future-reachable` for the case this reasoning depends on.

---

## Nits (non-blocking):

- `src/store.ts:1174-1177`: the colon-introduced breakdown reads as exhaustive but accounts for 146 of the stated 150 — the four unconditional backfill/reset `UPDATE`s (`src/store.ts:4442`, `:4547`, `:4576`, `:5708`) are the missing 4. A reader who sums the list will conclude one of the four numbers is wrong. Adding "plus 4 one-shot backfill UPDATEs" would close it.
- `src/store.ts:1180`: "Batched: ~9ms" and the 310ms figure are measurements, not something a reader can check against the code; the commit message already carries them with context, so the inline duplication is where they'll drift first.

- `src/store.ts:1663-1667`: the deliberate absence of `ROLLBACK` is sound only while no caller catches a `SessionStore` construction failure. If a future caller retries the open in-process, the abandoned handle still holds the RESERVED lock from the un-rolled-back transaction, so the retry's first write would block for the full `busy_timeout = 5000` (`src/store.ts:1173`) and then throw `SQLITE_BUSY` — a self-inflicted lock rather than the underlying error. Not reachable today (`src/index.ts:242` is module top level; `scripts/sweep-house-rules.ts:102` does not catch), so informational only.

## Stated limitations

- I did not run the test suite (read-only review; tests/lint are CI's job). The PR's central claim — that batching turns the three timing-out `test/store.test.ts` migration tests green — is a wall-clock claim I cannot reproduce without executing, so I neither confirm nor dispute the 310ms → 8.6ms / 206.6s → 80.1s numbers.
- `bun:sqlite` is a built-in module, so its `Database.prototype.transaction` implementation is not in this repo. I therefore could not read the source to confirm the comment's assertion that it emits `SAVEPOINT` (rather than a second `BEGIN`) when the connection is already inside a manually-issued transaction — assumed as written. What I *can* point at is that the assumption is not untested: `test/build-queue-store.test.ts:488-523` drives exactly that nesting on a disk-backed legacy DB, so if the assumption were false, CI would show it as a hard `cannot start a transaction within a transaction` failure on that test rather than as a silent latent bug.

Issue #2096, opened 2026-08-20
herdr 0.8.2: compatibility check


kai-osthoff:
### Candidate herdr version

0.8.2

### Release notes

https://github.com/herdrdev/herdr/releases/tag/v0.8.2

### Socket protocol number

20

### What the release notes flag

- **Protocol 19 → 20** — a dry `bun run herdr:compat -- --candidate 0.8.2` shows the diff is
  purely additive: `+pane.input.set`, `+pane_graphics_frame_ack` result variant, graphics-layer
  params (`layer_id`/`z_index`), `pane.split +right_click`, `IntegrationTarget +qwen`,
  `PaneGraphicsFormat +bgra`. **Zero** removals/reshapes; the #2032 record-shape gate passes clean.
- All live probes (L1–L9) PASS against an isolated 0.8.2 headless server, including the
  external-registration spawn replay, last-tab behaviour (unchanged vs. 0.8.0) and the terminal
  contract. The #1716 idle probe still lands on `done` → the sandbox-status floor stays 0.7.4.
- Release-note items worth a second read during phase 1: "Unix CLI commands now exit quietly when
  a downstream pipe closes instead of panicking with exit 101" (#2994 — we SAW this panic while
  piping `herdr status` through `head` on 0.8.0), headless default terminal now 120×40 instead of
  80×24 (#2828 — may affect pane-size assumptions), `agent prompt` now rejects blocked agents with
  `agent_blocked` (#2788), `agent start` readiness-wait changes (#2410 et al. — Shepherd does not
  use `agent start`, confirm nothing leaks into `pane run`), stable installs now verify SHA-256
  digests upstream.
- Opportunities (follow-ups, not bump scope): Qwen Code detection; `ui.window_title` sync;
  SHA-256 digests published per asset — could harden Shepherd's own install/downgrade path.

### Procedure (details in the rule — do not improvise the steps)

- [x] Phase 1 — release notes read against the rule's keyword list (first pass above; re-check in-session)
- [x] Phase 2 — `bun run herdr:compat -- --candidate 0.8.2` run; report produced (dry run, all-PASS + one REVIEW: admit protocol 20 to `HERDR_SOCKET_SUPPORTED_PROTOCOLS` by name)
- [ ] Phase 3 — every REVIEW triaged in writing; every FAIL fixed or the version declined
[...]

PR #2097, opened 2026-08-20
feat(herdr): admit new herdr versions via a standing SOP + herdr:compat A/B check

### LINKED ISSUE #2032: herdr version bumps need a record-shape gate: diff the generated protocol before raising the ceiling

kai-osthoff:
herdr 0.8.2 shipped while the ceiling sits at 0.8.0 (#2056) and the updater blocks it — but the
procedure that cleared 0.8.0 lived only in #2039's prose. This PR anchors that procedure in the
repo as a standing SOP, mechanises its measurable checks, and wires a trigger. It deliberately
does **not** move the ceiling: the 0.8.2 check runs as #2096, the SOP's first real use.

Closes #2032.

## What's in here

**The rule — `.claude/rules/herdr-version-bump.md`** (rendered: [Reference → herdr version bumps](https://docs.shepherd.run/reference/rules-herdr-version-bump/)).
Path-scoped to the version gates (`herdr-capabilities.ts`, `config.ts`, `herdr-install.ts`,
`herdr-update.ts`, generated protocol, the gen/compat scripts), so any agent touching them loads it
automatically. Carries the non-negotiables (ceiling moves only with a zero-FAIL report; the
protocol set stays an explicit allowlist; floors never move on a ceiling bump; live probes never
touch the operator's daemon; declining a version is a valid outcome), the procedure, the bump-PR
file checklist distilled from #2056's diff, and a bump-history table seeded from #721 → #2056.

**The check — `bun run herdr:compat -- --candidate <v> [--baseline <v>] [--static-only]`.**
- *Static:* schema diff from `api schema --json` (methods/params/results/enums), the #2032
  record-shape gate (required→optional / nullability drift on `TabInfo`/`PaneInfo`/`AgentInfo` in
  both response and event sections = FAIL — the exact #721/#2029 bug class), and a `--help`
  surface diff over every subcommand Shepherd invokes.
- *Live:* candidate vs. baseline as **isolated** headless servers (own `HOME`/`XDG_*`/socket under
  a short `/tmp` symlink; teardown wired to exit/SIGINT/SIGTERM), measuring the #2039 catalog A/B:
  the four reaper assumptions, tab-id reuse (#569), last-tab behaviour (#1760), the
  external-registration spawn replay (#1890), the #1716 idle→done probe (guards the sandbox
  floor), the status surface, and the terminal contract via `scripts/verify-herdr-terminal.ts`.
- *Output:* markdown report at `docs/herdr-compat/<v>.md` (committed by the eventual bump PR),
  exit 1 iff any FAIL; REVIEW items are written triage work, never a machine verdict.
- Pure halves (diff/gate/CLI-parse/report) are unit-tested offline, incl. the vendored schema
  diffed against itself.

**The trigger — `.github/ISSUE_TEMPLATE/herdr-compat.yml`** (label `herdr-compat`, created).
Deliberately no cron and no link from the update modal: the operator sees the blocked modal daily
anyway, and a modal link would surface a maintainer-only action to every Shepherd user. The form
carries only the phase skeleton and links the rule, so the two cannot drift.

**Visibility:** README's "supports herdr up to" callout and getting-started link the SOP.

## Verified

- Self-A/B (`--candidate 0.8.0`, installed binary both sides): exit 0, every check PASS.
- Dry run `--candidate 0.8.2`: completes, downloads + verifies the asset, all checks PASS except
  the expected S1 REVIEW (protocol 19 → 20 needs an allowlist entry) — findings carried into #2096.
  Report intentionally not committed here; the bump PR re-runs and commits it fresh.
- Operator's live daemon confirmed untouched before/after every run (`herdr status server`).
- Gates: root `bun run lint` + `bun run test` (8724 pass), `ui` `bun run check` + `bun run test`
  (4646 pass), `check:docs-manifest`, docs-site `bun run check` + `bun run build` (rule page renders).

## Deviations from the plan (flagged)

- The planned L8 probe (`herdr worktree list --porcelain`) checked a surface Shepherd doesn't
  drive — worktrees are managed via **git** (`src/worktree.ts`, spike 1074), and `--porcelain` is
  a git flag. The `worktree` subcommands left `SHEPHERD_HERDR_COMMANDS`; L8 now pins the
  `status server` parse that this tooling, the downgrade script and diagnostics read.
- The rule's table heading is "Bump history" instead of "Learnings": the docs-manifest keyword
  haystack ingests H2s, and a "Learnings" heading made the CommandBar query "learn" surface a
  second row, breaking the pinned exactly-one-result test. Pre-existing trap worth knowing:
  ANY future docs heading containing "learn*" re-triggers that test.

## Follow-ups

- #2096 — herdr 0.8.2 compatibility check (first real SOP run; pre-filled with the dry-run findings).
- Possible later: run the static half in CI on Linux runners once the SOP has carried one real bump.


kai-osthoff:
# Review: request changes

## Findings

1. `scripts/herdr-compat/isolated-server.ts: the startup-timeout path can leave the newly started isolated daemon running. The process is spawned at lines 99-101, but on a failed readiness check lines 123-126 only unregister it and remove the short symlink; they never run `server stop` or kill the retained process. This is reachable precisely when a candidate changes the `status server` output that the compatibility check is intended to detect (the readiness condition is the `status: running` regex at lines 121-122). The normal `stop()` path also discards a non-zero `server stop` result before deleting the socket link and scratch directory (lines 152-156). Ensure every failure/teardown path confirms the daemon has stopped before removing its only convenient control path, preserving the SOP's no-stray-daemon guarantee. Verified against `.claude/rules/herdr-version-bump.md:40-42`, which makes isolated operation a non-negotiable.

2. `scripts/herdr-compat/cli-surface.ts: S4 claims to cover every herdr subcommand Shepherd invokes, but the enumerated list at lines 20-42 omits `agent read` and `terminal session control`. `HerdrDriver.readArgs()` invokes `agent read --format --source --lines` in `src/herdr.ts:1376-1388`, and the socket terminal bridge invokes `terminal session control --takeover --cols --rows` in `src/socket-pty-bridge.ts:100-122`. Since the runner only diffs commands from this list (`scripts/herdr-compat.ts:301-315`), a candidate can remove or change the `agent read` command/flags without producing an S4 finding. Add the missing command surfaces (including their nested command forms) so the stated static compatibility check actually covers Shepherd's calls. Verified against the SOP's requirement for a help diff over every invoked subcommand in `.claude/rules/herdr-version-bump.md:52-58`.

`scripts/herdr-compat/probes.ts:243-247` — L5 ignores non-zero exits while closing the tabs that should be removed before the root tab. A future baseline whose last-tab behavior is workspace survival could then let the candidate appear to pass without ever closing the actual last tab; verify each preparatory close and re-list before the root close.

## Verification notes

- Read-only inspection only; no build, test, or other executable verification was run.
- Could not verify the external herdr CLI's live help/output contract because it is not present in this repository; findings above are based on the repository's own command call sites.

kai-osthoff:
## Summary

Request changes. The compatibility procedure still has an exit-time cleanup path that can orphan an isolated daemon, and S4 does not cover the `agent attach` command Shepherd executes for its browser-terminal fallback.

## Findings

1. scripts/herdr-compat/isolated-server.ts: readiness failure and unsuccessful stop can leave an isolated daemon running after its socket path is removed.

   The ordinary readiness-failure and explicit `stop()` paths now use `ensureDown()` before unlinking (`scripts/herdr-compat/isolated-server.ts:109-131`, `scripts/herdr-compat/isolated-server.ts:154-159`, `scripts/herdr-compat/isolated-server.ts:186-190`). However, the process-exit handler still invokes only `herdr server stop`, ignores its exit result, and unconditionally removes the `/tmp` symlink (`scripts/herdr-compat/isolated-server.ts:51-68`). The daemon is configured to bind its socket through that symlink (`scripts/herdr-compat/isolated-server.ts:94-99`), so a failed graceful stop on SIGINT/SIGTERM or another process-exit path leaves the child running but unreachable through its registered socket path. Apply the same verified termination sequence to that teardown path, or retain the link when termination cannot be confirmed.

2. scripts/herdr-compat/cli-surface.ts: S4 omits `agent attach`, so it cannot detect removal or flag drift in Shepherd's active browser-terminal fallback command.

   S4 describes its list as every command Shepherd invokes (`scripts/herdr-compat/cli-surface.ts:13-21`) and executes `--help` only for the listed entries (`scripts/herdr-compat.ts:301-315`), but the list has no `agent attach` entry (`scripts/herdr-compat/cli-surface.ts:22-45`). The Node PTY fallback actually starts `herdr agent attach <terminal> --takeover` (`src/pty-attach.mjs:14-25`), and its expected argv is covered by the existing test (`test/pty-bridge.test.ts:43`). Add this command (and its flags) to the S4 surface.

## Prior-review verification

The prior S4 omission is addressed for the two named commands: `agent read` and `terminal session control` are present in the S4 list (`scripts/herdr-compat/cli-surface.ts:35-42`), corresponding to Shepherd's read runner (`src/herdr.ts:1403-1422`) and terminal-control spawn (`src/socket-pty-bridge.ts:103-121`).

kai-osthoff (commented):
## Summary

No blocking findings. The rework addresses both prior findings within the reviewed diff.

## Verification

- Prior finding 1 is addressed: readiness failure calls `ensureDown()` before removing the socket symlink, and `ensureDown()` waits for the spawned server process, escalates from `server stop` to SIGTERM and then SIGKILL, and only then permits cleanup (`scripts/herdr-compat/isolated-server.ts:119`, `scripts/herdr-compat/isolated-server.ts:163`, `scripts/herdr-compat/isolated-server.ts:196`). The daemon is a foreground `Type=simple` `herdr server` process under the repository's service contract (`deploy/herdr.service:13`, `deploy/herdr.service:29`), so the tracked child is the daemon the escalation targets.
- Prior finding 2 is addressed: S4 now includes `agent attach` (`scripts/herdr-compat/cli-surface.ts:23`, `scripts/herdr-compat/cli-surface.ts:36`), and the browser terminal fallback invokes `agent attach <terminalId> --takeover` (`src/pty-attach.mjs:14`, `src/pty-attach.mjs:21`). The CLI diff classifies removed flags and missing commands as failures (`scripts/herdr-compat/cli-surface.ts:94`, `scripts/herdr-compat/cli-surface.ts:107`).
- The isolated terminal-contract invocation receives both the candidate binary and the isolated socket environment (`scripts/herdr-compat.ts:330`, `scripts/herdr-compat.ts:331`); the verifier resolves its binary from `HERDR_BIN` and its spawned CLI calls inherit that process environment (`scripts/verify-herdr-terminal.ts:49`, `scripts/verify-herdr-terminal.ts:76`).
- The SOP is registered in the docs importer and sidebar, and the public links use that resulting route (`docs-site/scripts/sync-docs.mjs:122`, `docs-site/astro.config.mjs:142`, `README.md:145`, `docs-site/src/content/docs/getting-started.md:122`, `ui/src/lib/docs-manifest.ts:121`).

## Limitations

I did not execute the live compatibility workflow or CI gates, per the review constraint. The behavior of a released herdr binary remains external to this repository and was not independently verified here.

kai-osthoff:
@scoop 👀✨ Shockingly, it's READY 🎉 CI green ✅ critic appeased 🤖💚 Your move, chief 😏🚀

Issue #2042, opened 2026-08-06
Mark sessions that are expensive to resume: cold prompt cache × large context


kai-osthoff:
## Ask

Mark sessions in the HUD that are **expensive to resume**, so the operator sees the cost *before* clicking into a cold, fat session — not afterwards in a usage report.

Shepherd already computes the two hardest inputs and throws both away at the wire boundary. This is mostly a plumbing + one small derivation task.

## Why now — Boris Cherny on what actually burns limits

On Threads, replying to a user who blew through their 5-hour limit in ~5 prompts ([`/usage` screenshot in the thread]):

> the usage output is pretty useful: pretty much all your usage seems to be coming from sessions that use lots of context and run for a long time. That means when tokens miss prompt cache (eg. Because your session has idled for >1 hour) then your entire context window needs to be re-cached, which is very expensive.
>
> Do you know where those very long stale sessions are coming from? Are they loops or your own interactive sessions?

The `/usage` panel he was reading attributes usage along three axes:

- `87% of your usage came from sessions active for 8+ hours` — *"These are often background/loop sessions."*
- `84% of your usage was at >150k context` — *"Longer sessions are more expensive even when cached."*
- `29% of your usage was while 4+ sessions ran in parallel` — *"All sessions share one limit."*

**All three are a description of Shepherd's operating model**, not an edge case. Long-lived worktree sessions, large accumulated context, run in parallel across the herd. Shepherd is the tool that makes it easy to be in all three buckets at once, so it should be the tool that says so.

## What it costs, in Shepherd's own units

From [`src/pricing.ts`](../blob/main/src/pricing.ts), Opus weights:

| bucket | weight | vs. a warm read |
| --- | --- | --- |
| `cacheRead` | 0.5 | 1× |
| `cacheWrite5m` | 6.25 | **12.5×** |
| `cacheWrite1h` | 10 | **20×** |

A cold resume re-writes the *entire* context at the write rate. For a 150k-context session on the 1h TTL:

- warm turn: `150_000 × 0.5 / 1e6` = **0.075 units**
- cold turn: `150_000 × 10 / 1e6` = **1.5 units**

That single turn costs **1.43 units more** than the warm equivalent — more than TASK-295's **entire** authoring cost (0.887 units, 13 messages, 6 minutes) in [`docs/token-usage-analysis.md`](../blob/main/docs/token-usage-analysis.md). One careless resume of a big session ≈ one whole small task, spent on nothing.

[...]

PR #2114, opened 2026-08-21
fix(recap): carry recap body + blocks as sidecar files (#2045)

### LINKED ISSUE #2045: Recap verdict has the same unescapable-quote hazard the critic just fixed

scoop:
Closes #2045.

`RecapService` hand-authored its verdict JSON exactly the way the critic did before #2042, so it carried the identical failure: a bare `"` immediately before `:` or `,` is byte-identical to a real key/value boundary. `jsonrepair` cannot resolve that ambiguity — no repairer could — and German prose produces it constantly via `„…":` / `„…",`. It has already thrown away a complete, correct recap in production.

## What changed

- **`body` → `.shepherd-recap.md`.** Markdown has no reserved characters, so the escaping obligation disappears entirely rather than being made more reliable.
- **`blocks` → `.shepherd-recap-blocks.json`**, read tolerantly. Blocks are still JSON, so moving them cannot remove the hazard the way the markdown body does — what it removes is the blast radius. `rich-text.markdown`, `callout.markdown`, `diff.annotations[].note` and quote-dense `wireframe.html` are exactly as prose-heavy as the body was, and inline they could sink the entire verdict. In their own file a mangled write costs only the visual cards.
- What stays in the verdict JSON — `verdict`, `headline`, `openItems[]` — is short and structured, where the model's escaping is reliable in practice. Mirrors the critic's `summary`/`findings` containment.

Went wider than the issue proposed, deliberately: a body-only fix would have left an equally live path to the same total loss.

## Compatibility

`body` and `blocks` stay **documented optional inline fields**, not removed ones:

- A Codex recap answers in chat and writes no files at all — its verdict arrives through the `-o` last-message capture, where no sidecar can exist. Dropping the fields would give that provider a recap with no text.
- The same fallback keeps a run already in flight across a server restart working unchanged.

The prompt states the write ordering explicitly (markdown first, JSON **last**) because `.shepherd-recap.json` is the completion signal `tick()` finalizes on.

## Deliberate non-changes

- **No pre-seed scrub.** The issue asked for this to be checked rather than copied. The recap spawn's cwd is `mkdtempSync(tmpdir(), "shepherd-recap-")` — a fresh empty directory, never an untrusted PR-head checkout. The critic needs `scrubStaleVerdictArtifacts` because its worktree *is* the untrusted head; that attack has no analogue here.
- **Still fail-closed.** An unparseable verdict JSON fails even when a valid markdown sidecar is present. `verdict` is a UI-switched enum that exists only in the JSON, so there is nothing to salvage the prose onto — synthesizing one would fake a result.
- The issue's second observation (archive → `forget()` → `dropReview()` making a healthy review indistinguishable from a failed one) was filed as an observation with no change proposed, and is untouched.

## Verification

Ten new tests in `test/recap-core.test.ts`, including the exact killer shape (`Das ist „so nicht mehr": drei Dinge`) round-tripping byte-for-byte through the sidecar, a garbage blocks sidecar still yielding a `ready` recap, and the no-salvage guarantee. Root `bun run lint` clean, `bun run test` 8731 pass / 0 fail, `tsc --noEmit` clean, `fallow audit --base origin/main` gate green.

### DIFF
--- a/src/operator-language.ts
+++ b/src/operator-language.ts
@@ -137,7 +137,7 @@ const backtickList = (fields: readonly string[]): string =>
 /**
  * Return an instruction paragraph telling the agent, when it authors the
  * `VisualBlock[]` JSON (used by both the plan sidecar `.shepherd-plan-blocks.json`
- * and the recap `.shepherd-recap.json`), which fields to write in the operator's
+ * and the recap sidecar `.shepherd-recap-blocks.json`), which fields to write in the operator's
  * language and which to leave verbatim — or `null` for "en" (nothing to inject).
  * Positively enumerates the translatable fields and prohibits everything else,
  * plus the conditional cell-by-cell rule for `table.columns`/`table.rows`.
@@ -147,7 +147,7 @@ export function visualBlockLanguageLine(lang: OperatorLanguage): string | null {
   const name = LANGUAGE_NAMES[lang];
   return (
     `When authoring VisualBlock[] JSON (the plan sidecar .shepherd-plan-blocks.json or the recap ` +
-    `.shepherd-recap.json), write ONLY these natural-language fields in ${name}: ` +
+    `sidecar .shepherd-recap-blocks.json), write ONLY these natural-language fields in ${name}: ` +
     `${backtickList(VISUAL_BLOCK_TRANSLATE_FIELDS)}. ` +
     `Leave every other field exactly as you would in English — never translate: ` +
     `${backtickList(VISUAL_BLOCK_VERBATIM_FIELDS)}. An off-enum or reworded identifier/path/enum ` +
--- a/src/recap-core.ts
+++ b/src/recap-core.ts
@@ -13,6 +13,34 @@ export const RECAP_VERDICTS: readonly RecapVerdict[] = ["ready", "parked", "need
 export const RECAP_HEADLINE_MAX = 100;
 export const RECAP_DIGEST_MAX_CHARS = 4000;
 
+/** The recap `body`, written beside the verdict JSON as plain markdown.
+ *
+ *  #2045 (the recap half of #2042): the recap agent hand-authors its verdict JSON, so every `"` in
+ *  the body has to be escaped by the model. It only has to miss once. The unrecoverable shape is a
+ *  bare `"` immediately before `:` or `,` (`…nicht mehr": drei Dinge`), which is byte-identical to a
+ *  real key/value boundary — jsonrepair cannot resolve that ambiguity, and no repairer could. German
+ *  prose produces it constantly via `„…":` / `„…",`, and it has already destroyed a complete recap
+ *  in production.
+ *
+ *  Carrying the body as its own file removes the escaping obligation entirely: markdown has no
+ *  reserved characters, so no prose can break the read. What remains in the JSON (verdict, headline,
+ *  openItems) is short and structured, where the model's escaping is reliable in practice.
+ *
+ *  Unlike the critic's equivalent, this needs NO pre-seed scrub: the recap spawn's cwd is a fresh
+ *  mkdtemp directory (see defaultMakeTmpDir in recap.ts), never an untrusted PR-head checkout. */
+export const RECAP_BODY_FILE = ".shepherd-recap.md";
+
+/** The recap `blocks`, written beside the verdict JSON as a bare `VisualBlock[]` — mirroring the
+ *  plan gate's `.shepherd-plan-blocks.json` sidecar.
+ *
+ *  Blocks are still JSON, so moving them out cannot remove the escaping hazard the way the markdown
+ *  body does. What it removes is the BLAST RADIUS: `rich-text.markdown`, `callout.markdown`,
+ *  `diff.annotations[].note` and quote-dense `wireframe.html` are exactly as prose-heavy as the body
+ *  was, and inline they could sink the entire recap. In their own file a mangled write costs only
+ *  the visual cards — the verdict, headline and body still land (see readRecapBlocks in recap.ts,
+ *  which falls back rather than failing). */
+export const RECAP_BLOCKS_FILE = ".shepherd-recap-blocks.json";
+
 /** Recover the recap object from a parse that an unattended agent's chatty output mangled.
  *  jsonrepair turns prose-wrapped JSON ("Here is the recap:\n{…}" or "{…}\nDone.") into an
  *  ARRAY (e.g. `["Here is the recap:", {…}]`); pull the first recap-shaped element back out so a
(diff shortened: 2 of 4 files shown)


scoop (commented):
# Review — recap verdict: carry body + blocks as sidecar files (#2045)

**Decision: comment.** The change does what the issue asks, mirrors the already-landed critic fix
faithfully, and I could not find a correctness or security defect in it. Everything I noted below is
non-blocking.

**The hazard is actually removed for the body.** `defaultReadVerdict` now splices the sidecars after a
successful parse (`src/recap.ts:189`), and the body sidecar is plain markdown read verbatim
(`src/recap.ts:195-203`). Nothing on that path re-encodes the prose through JSON, so `„…":` /`„…",`
can no longer produce the byte-ambiguous key/value boundary the issue describes. The regression test
asserts byte-for-byte round-trip rather than `toContain` (`test/recap-core.test.ts:866-878`), which is
the right assertion for this bug.

**Fail-closed is genuinely preserved.** The unparseable branch returns before the splice
(`src/recap.ts:183`), so a valid `.shepherd-recap.md` next to a garbage `.shepherd-recap.json` still
yields `status: "unparseable"` → `decideVerdictAction` → `finalize-null` → `state: "failed"`
(`src/json-tolerant.ts:165-167`, `src/recap.ts:995-1003`, `src/recap.ts:1118-1131`). Verified the
no-salvage claim end to end; the comment at `src/recap.ts:184-188` is accurate about why.

**Blocks isolation behaves as claimed.** `readRecapBlocks` returns `null` on absent/unreadable/
unparseable (`src/recap.ts:210-219`), `spliceRecapSidecars` treats `null` as "no contribution"
(`src/recap-core.ts:74,78`), and a non-array value that *does* parse is degraded to `[]` downstream
rather than throwing (`src/visual-blocks.ts:585-594`). So a mangled blocks file can cost only the
visual cards. Confirmed the `raw !== undefined` warn branch at `src/visual-blocks.ts:588` is not
tripped in the absent case, because `blocks` is then never assigned onto the object.

**The array (jsonrepair-prose-wrapped) shape still works.** `spliceRecapSidecars` routes through
`unwrapRecapObject` (`src/recap-core.ts:48-57`, called at `src/recap-core.ts:75`), which requires a
`"verdict" in e` element; `parseRecapVerdict` unwraps again at `src/recap-core.ts:107`, and
re-unwrapping an already-unwrapped object is a no-op (`src/recap-core.ts:55`). Idempotent as the
docstring claims.

**Inline fallback survives, so no provider/restart regression.** `spliceRecapSidecars` returns `raw`
untouched when both sidecars are absent (`src/recap-core.ts:74`), which is exactly the pre-existing
`#822` fixture path (`test/recap-core.test.ts:30-47`) and the Codex `-o` chat-only path
(`test/recap-core.test.ts:169-184`, fallback read at `src/codex-last-message.ts:80-89`).

**"No pre-seed scrub needed" is correct, not copied blindly.** The recap cwd is
`mkdtempSync(join(tmpdir(), "shepherd-recap-"))` (`src/recap.ts:268-270`), and `deps.makeTmpDir` is
never overridden in production — grepped every `makeTmpDir` in `src/`, only tests inject it
(`src/recap.ts:446`). The scrub's stated precondition is an untrusted PR-head checkout with a fixed
artifact name (`src/codex-last-message.ts:93-112`), which has no analogue here. A fresh `mkdtemp`
directory is mode 0700, so no other local user can pre-seed either file.

**The agent can actually write the new files.** The recap spawns as `writer-only`, whose allowlist is
bare `Write` — not path-scoped (`src/transient-agent-argv.ts:173`, rationale at
`src/transient-agent-argv.ts:162-167`) — so both `.shepherd-recap.md` and
`.shepherd-recap-blocks.json` are writable under `--permission-mode dontAsk`. A path-scoped allowlist
would have silently denied them; it isn't one.

**Identifiers and filenames are consistent.** `RECAP_BODY_FILE = ".shepherd-recap.md"` /
`RECAP_BLOCKS_FILE = ".shepherd-recap-blocks.json"` (`src/recap-core.ts:31,42`) are imported and used
in `src/recap.ts:56-57,196,211`, and the prompt names the same two via interpolation
(`src/recap-core.ts:241-242,258`). The hardcoded string in `src/operator-language.ts:150` now reads
`.shepherd-recap-blocks.json`, matching the constant. `RECAP_VERDICT_FILE` is unchanged at
`src/recap.ts:71`, and the prompt still spells it `` `.shepherd-recap.json` `` at
`src/recap-core.ts:243,255,258`. No third spelling anywhere: grepped the whole tree for
`shepherd-recap` — the only other hits are `src/recap.ts:269` (tmpdir prefix) and a research doc.

**Existing prompt assertions still hold** (so the untouched test files are not left stale):
`test/recap-core.test.ts:444` (`.shepherd-recap.json` present → yes, `src/recap-core.ts:243`),
`test/recap-core.test.ts:668-669` (`"// blocks"` absent, `"no comments"` present → yes,
`src/recap-core.ts:243`), and `test/recap.test.ts:1134` (backtick-quoted
`` `.shepherd-recap.json` `` in the trailing argv → yes, `src/recap-core.ts:243`). No test asserts the
old `.shepherd-recap.json` wording inside `visualBlockLanguageLine` — `test/operator-language.test.ts`
asserts only field tokens and the table rule (`test/operator-language.test.ts:220-280`).

**No consumer contract changed.** `finalize` still reads `parsed.body` / `parsed.blocks` and grounds
through `groundBlocks` (`src/recap.ts:1103-1117`); the `Recap` row shape, `RecapServiceDeps`, and the
`readVerdict` seam (`src/recap.ts:352`) are untouched, so nothing under `ui/` needs a change. The two
new filenames are already covered by the `.shepherd-*` catch-all in `.gitignore:26` and
`SHEPHERD_IGNORE_GLOB` (`src/shepherd-exclude.ts:21`) — irrelevant in a tmpdir, but it means the same
names are safe if a future caller ever runs the role in a repo.

**Critic parity.** `readRecapBody` is structurally identical to `readVerdictBody`
(`src/critic-core.ts:935-943`), including "sidecar wins when both exist" and the never-throw contract
(`src/critic-core.ts:926-929`), and the ordering instruction mirrors `src/critic-core.ts:739`.

## Residual risks I checked and consider acceptable (not findings)

- **Ordering disobedience → empty body.** A strict parse finalizes immediately regardless of spawn
  liveness (`src/json-tolerant.ts:161-162`), so if the model writes `.shepherd-recap.json` *before*
  `.shepherd-recap.md`, a tick landing in the gap between two tool calls finalizes `ready` with an
  empty body. This window is a full model turn, not milliseconds. It is the same exposure the already
  shipped critic accepts (`src/critic-core.ts:736-739`), the prompt states the ordering as its final
  instruction (`src/recap-core.ts:258`), and the worst case is a rare empty body instead of today's
  guaranteed total loss on German prose. Strictly better than the status quo, so I am not blocking on
  it — but it is the thing to watch in the logs after this lands.
- **Empty sidecar beats a good inline body.** `readRecapBody` returns `""` (not `null`) for a
  zero-byte file, which then overwrites an inline `body` (`src/recap-core.ts:77`). Deliberate and
  byte-identical to the critic; only reachable if a run somehow has both an inline body and a
  truncated sidecar.
- **A blocks sidecar written as `{"blocks": [...]}` instead of a bare array** parses fine and is
  spliced, then degrades to `[]` with a console warn (`src/visual-blocks.ts:587-593`) rather than
  being unwrapped. The prompt is explicit about the bare-array shape (`src/recap-core.ts:242`), and
  the plan-gate sidecar reader has the same strictness (`src/plan-gate.ts:1862-1868`), so this is
  consistent rather than a gap.

## Nits (non-blocking):

- src/recap-core.ts:243-244 — the shape literal says "EXACTLY this shape" and then line 254 permits inline `body`/`blocks` for the chat-only path; the critic avoids that tension by keeping `"body": "<optional — see below>"` inside the literal itself (src/critic-core.ts:723). For the Codex-answers-in-chat case a literal-minded model reads the stricter sentence first.
- src/operator-language.ts:150 — the recap sidecar filename is a hardcoded string rather than `RECAP_BLOCKS_FILE`; the import direction makes that unavoidable (recap-core imports operator-language, src/recap-core.ts:10), but nothing asserts the two agree, so a future rename of the constant leaves this line silently stale. A one-line test comparing the emitted string to `RECAP_BLOCKS_FILE` would close it.
- src/recap-core.ts:202,205-206 — `- body:` and `Optionally, add \`blocks\`` still read as JSON-field bullets; the term is only bound to the files ~40 lines later at src/recap-core.ts:241. Correct as written, just front-loaded ambiguity.

PR #2115, opened 2026-08-21
fix(diagnostics): surface the mise/native claude duplicate and stop re-planting it

### LINKED ISSUE #2052: mise-managed claude can't self-update, and its auto-updater plants a native duplicate
Follow-up from #2051, which made the claude/codex **install** remediation and the **codex updater** mise-aware. Claude was deliberately left out of the update half, because Shepherd has no claude updater at all — Claude Code self-updates.

That leaves a real gap on a mise host:

- A mise-managed `claude` (aqua/http backend) cannot self-update in place.
- Claude Code's own auto-updater keeps writing a native install to `~/.local/share/claude` + `~/.local/bin/claude`, which then **shadows** the mise-managed one.

Observed on the Shepherd host: `~/.local/bin/claude` → native **2.1.226** while `mise ls` tracks claude **2.1.222** — two installs, diverged, and `mise upgrade claude` moves only the one nothing runs.

Possible directions (undecided):

1. A `ClaudeUpdateService` mirroring `CodexUpdateService` — badge + modal + the same advance-or-report contract, with `mise upgrade claude` as the only channel when mise owns the on-PATH claude. Largest surface.
2. Detection-only: a diagnostics row that reports the duplicate (`mise which claude` resolves AND the on-PATH claude is not it) and explains which one Shepherd actually spawns. Cheap, and it names the failure the operator currently has to notice by hand.
3. Pin `DISABLE_AUTOUPDATER=1` for spawns when claude is mise-managed, so Claude Code stops re-planting the native copy. Needs a check that it doesn't strand a non-mise host on an old build.

(2) is the smallest thing that would have surfaced the live divergence; (1) is the complete answer.

scoop:
Closes #2052.

Scope is directions **2 + 3** of the issue, agreed at the plan gate. No `ClaudeUpdateService` (direction 1).

## The problem

Claude Code self-updates by writing a **native** install into `~/.local/share/claude/versions` and repointing `~/.local/bin/claude` at it. On a host where mise *also* manages `claude`, that copy **shadows** the mise-managed one: the two drift, and `mise upgrade claude` then moves only the install nothing runs.

## What lands

**Stop the cause.** `buildWrappedArgv` pins `DISABLE_AUTOUPDATER=1` for a `claude` inner command when mise owns the claude we spawn. Codex spawns never get it; a non-mise host's argv is byte-identical to before and keeps self-updating, so nothing gets stranded on an old build.

**Name the symptom.** A `claude_install` row in Settings → Diagnose:

| Host state | Row | Message carries | Pinned? |
| --- | --- | --- | --- |
| `mise which claude` fails / no mise | row absent entirely | — | no |
| versions match, no native residue | `ok` | that mise owns it + the pin is on | yes |
| versions differ | `warning` | both versions, which one Shepherd spawns | no |
| versions match but a native copy is what runs | `warning` | the shared version, and that they will drift | no |
| versions match, native tree left on disk | `warning` | size + build count | yes |

`ok`/`residue` are exactly the pinned states, so the row can never claim a pin the spawn path didn't apply.

## Two design calls worth reviewing

**Ownership is version equality, not path shape.** `codex-update.ts` decides mise-ownership structurally (path under the mise data dir, or realpath-equal to `mise which`). That misreads a third host shape: a launcher *script* on `PATH` that runs `mise use -g claude` then `exec mise x claude` — which is what the host this was filed from actually uses. It realpaths to itself, outside the mise tree, so a structural test calls it unmanaged and both consumers go silent on the very host in the issue. Comparing `claude --version` against the mise-managed binary's own `--version` reads correctly for shim, symlink and launcher alike.

**The pin carries one guard the row doesn't.** Equal versions also match a native duplicate sitting at mise's version; pinning there would freeze a build mise cannot advance. So the pin additionally requires the on-PATH realpath to be outside `$HOME/.local/share/claude/` — and that state gets its own warning rather than a misleading `ok`.

## Notes

- **Guidance-only row** — no Fix button. Repointing `~/.local/bin/claude` would clobber an operator's launcher script, and deleting a several-hundred-MB install tree isn't Shepherd's call. Both non-ok states carry a doc-link to the operating guide instead.
- **`DiagnosticCheck` gains `hintParams`** — the hint-side analogue of `fixActionParams`, so the row can say "we spawn 2.1.226, mise tracks 2.1.222" and "938 MB across 3 builds". Payload purity bans that from `remediation`; the purity test now regex-bounds the new bag (version triples, a scaled size, a count, no `/`).
- **Sandboxed spawns are out of scope on purpose** — the membrane binds `~/.local/share/claude` and `~/.local/bin` read-only under `--tmpfs $HOME`, so an in-sandbox updater cannot plant on the host. Only the trusted path re-plants.
- **`agentCliProbes` / the ctor were shedding complexity, not being refactored for taste** — this change pushed both over the fallow cognitive gate; the two agent-CLI rows now share one builder and `readMiseClaude` resolves in `resolveClaudeTrustDeps` (the helper that exists for exactly that reason).

## Verification

`bun run lint`, `bunx tsc --noEmit`, `bun run test` (8753 pass), `cd ui && bun run check && bun run test && bun run check:i18n` (4665 pass, 2 locales in parity), `cd extension && bun run check`, and `bunx fallow@2.100.0 audit --base origin/main --fail-on-issues` — all green. Pre-push passed.

## Unrelated, but worth a look

This session's worktree `~/Work/.shepherd-worktrees/shepherd-mise-managed-claude-self` and its branch did **not** exist at startup, despite the session's own git snapshot naming both. I recreated the worktree off `origin/main` @ `6d22fbe0` under the same branch name so the session had somewhere to live. Smells like a session-isolation defect, separate from this issue.


scoop:
# Review — `fix(diagnostics): surface the mise/native claude duplicate and stop re-planting it`

Diff reviewed: `6d22fbe0...HEAD` (16 files, +955/−24). Read-only inspection; no builds or tests run.

Round note: no prior review artifact exists in this worktree (`.shepherd/` holds only `roles.json`, and
there is no committed `.shepherd-review.*`), so nothing is carried forward — the findings below are new.

## What the change does, and what I verified

The PR implements issue directions **2 + 3** as the plan describes: a new `src/mise-claude.ts` probe
(mise-ownership by version equality + native-tree residue scan), a guidance-only `claude_install`
DIAGNOSE row, and a `DISABLE_AUTOUPDATER=1` spawn-env pin in `buildWrappedArgv`. Direction 1
(`ClaudeUpdateService`) is correctly left out per the plan's `Out of scope`.

Verified against ground truth:

- **Verdict/pin alignment.** `verdictFor` (src/mise-claude.ts:184-189) returns `ok`/`residue` exactly
  when `pinFor` (src/mise-claude.ts:197-204) is true, and `diverged`/`native` exactly when it is
  false. I walked every reachable combination of `managed`/`pathVersion`/`miseVersion`/`nativeOnPath`
  by hand: the two agree in all of them, including the `pathVersion === null && miseVersion === null`
  case (`pinFor`'s `pathVersion !== null` clause blocks the vacuous `null === null` match). So the
  `ok` copy's claim that the pin is applied cannot be false.
- **i18n parity.** All four new keys exist in both catalogs at matching positions with matching
  placeholders: `diagnostics_label_claude_install` (ui/messages/en.json:1347, ui/messages/de.json:1347),
  `..._install_ok` / `..._diverged` / `..._native_on_path` / `..._native_residue`
  (ui/messages/en.json:1369-1372, ui/messages/de.json:1369-1372). `{running}`/`{managed}` appear in both
  locales' `diverged`, `{running}` in both `native_on_path`, `{count}`+`{size}` in both `native_residue`.
  `ui/scripts/check-i18n.mjs:44-53` enforces key parity only, so the placeholder parity above is the
  part I checked by reading.
- **The `agentCliCheck` refactor is behaviour-preserving.** The six template-literal keys it can emit
  (`diagnostics_hint_{claude,codex}_{ok,optional,missing}`) all exist in both catalogs
  (ui/messages/en.json:1365-1366, 2259-2262; ui/messages/de.json:1365-1366, 2259-2262).
- **Doc anchor.** `#the-claude-code-install-row` (ui/src/lib/diagnostics-docs.ts:30-36) slugs from
  `### The Claude Code install row` (docs-site/src/content/docs/operating.md:134), and
  test/host-capacity-doc-anchor.test.ts:40-50 already checks *every* `operating/#…` anchor in
  `DOC_LINKS`, so the in-code comment claiming that coupling is accurate.
- **Docs manifest freshness.** `ui/src/lib/docs-manifest.ts:36` inserts `the claude code install row`
  between `agent clis managed by mise` and `host tuning — tmpfs inodes`, which matches the H2/H3 order
  in operating.md (headings at lines 112, 134, 164) and the generator's `extractHeadings` +
  `keywordsOf` (ui/scripts/gen-docs-manifest.ts:58-83).
- **`hintParams` has no other consumers to break.** `fixActionParams` — the field it mirrors — is read
  only in ui/src/lib/components/DiagnoseRows.svelte:69 and src/diagnostics.ts:1600; the new field is
  read only at DiagnoseRows.svelte:64. No server/CLI renderer resolves `diagnostics_hint_*` messages
  (the other matches in `deploy/provision.ts`, `ci/onboarding-harness/*` map hintKeys to remediation
  commands, they don't interpolate copy), so a raw `{running}` cannot leak into a non-UI surface.
- **`claude_install` cannot be mistaken for the `claude` presence row.**
  `ui/src/lib/provider-capacity.ts:16-19` matches `check.id === provider` exactly, not by prefix.
- **Adaptive re-check cadence is unaffected.** `nextDiagnosticsDelay` accelerates only on `error`
  (src/diagnostics.ts:242-245); the new row is `warning` at worst, so a permanent residue warning will
  not pin the probe fan-out at 60s.
- **The membrane claim in the herdr comment is true.** src/sandbox.ts:562-568 binds
  `$HOME/.local/share/claude` and `$HOME/.local/bin` with `--ro-bind-try` under the `--tmpfs $HOME`,
  so an in-sandbox updater genuinely cannot plant on the host.
- **The `mise which` precedent claim is true.** src/remediations.ts:98-100 uses the same exit-0-plus-path
  test and the same "`mise ls` exits 0 either way" reasoning.
- **argv shape assumptions hold.** Trusted spawns pass `["claude", …]` (src/service.ts:3061-3062,
  3178-3179; src/transient-agent-argv.ts:250-251), and `innerCommandStart` (src/herdr.ts:631-634)
  reuses the exact `bwrap … --` convention `isHeadlessCodexExec` already depended on.

---

## Findings

### 1. `src/mise-claude.ts`: the FIDELITY note's premise is false, and a PATH disagreement does not fail safe

src/mise-claude.ts:29-34 states:

> …both units are started by the same user manager with the same `~/.local/bin`-first PATH, and a
> mismatch would only cost accuracy in the row, never safety: every disagreement resolves to "not
> managed", i.e. no row and no pin.

Both halves are wrong, and I checked the shipped unit files:

- deploy/shepherd.service:22 → `Environment=PATH=%h/.bun/bin:%h/.local/bin:/usr/local/bin:/usr/bin:/bin`
- deploy/herdr.service:38 → `Environment=PATH=%h/.local/bin:%h/.bun/bin:/usr/local/bin:/usr/bin:/bin`

The two units pin `~/.bun/bin` and `~/.local/bin` in **opposite order**. `defaultResolveOnPath`
(src/mise-claude.ts:98-110) returns the *first* executable named `claude` on Shepherd's PATH, so on a
host that carries a bun-global Claude Code (`bun install -g @anthropic-ai/claude-code`) alongside the
`~/.local/bin` one, Shepherd probes `~/.bun/bin/claude` while herdr execs `~/.local/bin/claude`.

Nothing in the module cross-checks that, so the disagreement does **not** collapse to `NOT_MANAGED`:
`pathVersion`, `nativeOnPath`, `verdictFor` and `pinFor` are all computed from the wrong binary. Two
concrete consequences:

- If the bun-global copy happens to report mise's version, `pinFor` (src/mise-claude.ts:197-204) is
  true and `buildWrappedArgv` (src/herdr.ts:620, 626) sets `DISABLE_AUTOUPDATER=1` on a spawn that
  actually runs the *native* `~/.local/bin/claude` — freezing the exact install mise cannot advance,
  which is the failure the pin's `nativeOnPath` guard exists to prevent.
- The same state yields `ok`/`residue`, whose copy tells the operator the leftover tree is
  "safe to delete" (ui/messages/en.json:1372) while herdr is running out of it.

Either resolve the on-PATH claude the way herdr will (`~/.local/bin` first, matching
deploy/herdr.service:38), or drop the unsupported safety claim and make a disagreement actually fail
to `NOT_MANAGED`. As written the comment asserts a property the code does not have.

### 2. `src/index.ts`: the Diagnose "Re-run" cannot refresh the `claude_install` row for up to 6 hours

`DiagnosticsService.check(now)` is documented as "Force a fresh run of all probes"
(src/diagnostics.ts:1538-1540), and that is what `GET /api/diagnostics?refresh=1` is for — "bypasses
the cache and forces a fresh probe run" (src/server.ts:4729-4738). The steady-state comment at
src/diagnostics.ts:215-216 leans on the same contract ("the manual Diagnose 'Re-run' is the on-demand
live path instead").

The new row breaks it. src/index.ts:2836 wires `readMiseClaude: () => miseClaudeState(Date.now())`,
and `miseClaudeState` short-circuits on its own independent 6h TTL (src/mise-claude.ts:221-231, cache
hit at :225, `MISE_CLAUDE_TTL_MS` at src/mise-claude.ts:75) with no force path. So the operator who
follows the row's own instruction — "Remove the native copy so mise's is back on PATH", "nothing runs
it and it is safe to delete" (ui/messages/en.json:1370-1372) — deletes the tree, hits Re-run, and sees
the identical warning with the identical stale `{size}`/`{count}`/version numbers, for up to six hours
or until Shepherd restarts. That reads as "the deletion failed".

This is not required by the shared-cache design: forcing a re-probe on `refresh` also recomputes
`pinned` (src/mise-claude.ts:229), so the row and the pin stay consistent either way. A `force`/bypass
argument threaded from `check(now)` into `miseClaudeState` would close it.

---

- `src/mise-claude.ts`: `pinFor` leans entirely on `nativeOnPath`, which is `false` whenever `realpath` throws (src/mise-claude.ts:163, `.catch(() => onPath)`); an EACCES/ELOOP on a native symlink therefore degrades to "pin it", the opposite of the fail-safe direction the rest of the module takes.

- `src/diagnostics.ts`: the `agentCliCheck()` extraction (src/diagnostics.ts:331-344) is a drive-by refactor of the pre-existing claude/codex rows that the task did not require; it is behaviour-preserving (verified above), just unrelated churn in the same commit.

## Nits (non-blocking):

- `src/herdr.ts:600-602`: the comment "Sandboxed spawns don't need it: the membrane binds … read-only" reads as though the token is not emitted for a membrane spawn, but `innerCommandStart` deliberately looks *behind* the `bwrap --` and does emit it (asserted at test/herdr.test.ts:1741-1745), where `--clearenv` then discards it. Worth one clause saying the token is emitted and harmlessly dropped, so a future reader doesn't "fix" the perceived gap.
- `ui/messages/en.json:1370`: the `diverged` copy names Claude Code's auto-updater and `~/.local/bin/claude` as *the* cause, but the verdict fires for any on-PATH claude whose version differs from mise's — an npm/bun global install lands here too and gets a confidently wrong explanation.
- `src/mise-claude.ts:73`: `SEMVER_RE` + `probeVersion` re-implement what `src/version-probe.ts` already does for the herdr/codex updaters (the plan even says it reuses that capture); the existing helpers are sync-only so reuse isn't free, but a pointer comment would keep the three copies findable.

- Claude Code's native install layout is not represented anywhere in this repo, so I could not confirm that `~/.local/share/claude/versions/<version>` entries are regular **files** rather than directories. `defaultListNative` filters on `st.isFile()` (src/mise-claude.ts:124); if the real layout is one directory per build, the residue scan silently returns `[]` and the row emits `ok`, whose copy asserts "no leftover native install is on disk" (ui/messages/en.json:1369). The tests model the file layout (test/mise-claude.test.ts:100-101, 148-152), which is consistent but is not independent evidence. Assumed as written; flagging the asymmetry because the failure mode is a confidently false `ok`, not a missing warning.
- `DISABLE_AUTOUPDATER=1` is Claude Code's own env var; its semantics are not verifiable from this repo. Assumed as written.

scoop:
# Review — mise-managed claude: detect the duplicate, stop the re-planting (rework round 2)

Scope: `git diff 6d22fbe016ef89a2dc9f3cfae96fa0d9d374212b...HEAD` (two commits, `122abcbf` + `e025ec93`).

## Prior findings — verification

**Prior 1 — the FIDELITY note's false claims about PATH resolution. ADDRESSED.**

I compared the note against both unit files directly. `deploy/shepherd.service:22` is
`Environment=PATH=%h/.bun/bin:%h/.local/bin:/usr/local/bin:/usr/bin:/bin`; `deploy/herdr.service:38`
is `Environment=PATH=%h/.local/bin:%h/.bun/bin:/usr/local/bin:/usr/bin:/bin`. The rewritten note at
`src/mise-claude.ts:31-44` now states exactly that (`.bun/bin` first vs `.local/bin` first, "same
directories, different order") — no longer the false "`~/.local/bin`-first for both".

The resolution itself changed shape to match. `MiseClaudeDeps.listOnPath`
(`src/mise-claude.ts:76-79`) now enumerates *every* executable named `claude` across `PATH`
(`defaultListOnPath`, `src/mise-claude.ts:109-122` — it walks the whole `PATH`, no early return),
and `detectMiseClaude` collapses them through `realpath` and bails to `NOT_MANAGED` unless exactly
one distinct target survives (`src/mise-claude.ts:174-177`). Since the two units' `PATH`s are
permutations of one directory set, a single distinct target is order-independent, so the verdict and
the pin are computed against the binary herdr will reach; two distinct targets fail closed to no row
and no pin. That is exactly the "resolve the way herdr will, or genuinely fail to NOT_MANAGED"
remedy the prior finding asked for. Pinned by `test/mise-claude.test.ts:187-209` (two distinct
realpaths ⇒ the full `NOT_MANAGED` shape; several entries onto one target ⇒ managed + pinned).

**Prior 2 — the probe's own 6h TTL defeating `check(now)`'s force-a-fresh-run contract. ADDRESSED.**

The TTL is gone. `refreshMiseClaude` (`src/mise-claude.ts:236-240`) calls `detectMiseClaude`
unconditionally and re-latches `pinned` from that result on every invocation — no `now` argument, no
cache cell. `src/index.ts:2841` wires `readMiseClaude: () => refreshMiseClaude()`, so
`DiagnosticsService.check(now)` — the documented force-a-fresh-run path at `src/diagnostics.ts:1538-1540`,
reached by `?refresh=1` at `src/server.ts:4736-4739` and by the post-fix re-probe — now re-executes
the probe. `src/index.ts:2795-2801` correspondingly drops the second timer and keeps only the 1s boot
kick that warms the synchronously-read pin. The row and the pin move together on every check, forced
or scheduled (`test/mise-claude.test.ts:261-276` counts two probes for two calls and asserts the pin
follows the newest state).

## What I verified beyond the priors

- **Identifiers resolve.** `formatResidueSize`, `verdictFor`, `MiseClaudeState` are exported from
  `src/mise-claude.ts:223/199/52` and imported at `src/diagnostics.ts:37`; `claudeSpawnPinned` is
  exported at `src/mise-claude.ts:245` and imported at `src/herdr.ts:15`; `refreshMiseClaude` at
  `src/mise-claude.ts:236` / imported `src/index.ts:69`.
- **i18n parity.** All five new keys exist in both catalogs: `diagnostics_label_claude_install`
  (`ui/messages/en.json:1347`, `ui/messages/de.json:1347`) and the four hints
  `..._install_ok` / `_diverged` / `_native_on_path` / `_native_residue`
  (`ui/messages/en.json:1369-1372`, `ui/messages/de.json:1369-1372`). Placeholder names match
  per-key across locales — `{running}`+`{managed}` on diverged, `{running}` on native_on_path,
  `{count}`+`{size}` on residue — and the `{param}` syntax matches the existing parameterised
  message `diagnostics_fix_action_host_capacity` (`ui/messages/en.json:3298`,
  `ui/messages/de.json:3298`).
- **Hint interpolation consumer.** `hint()` now takes the check and passes `check.hintParams`
  (`ui/src/lib/components/DiagnoseRows.svelte:61-64`); the single call site was updated
  (`:138`). The `msg` cast at `:55` is already typed `(p?: Record<string,string>) => string`, so
  param-less hints ignoring the extra arg is the same mechanism `fixActionText()` at `:69` relies
  on. `hintParams` is mirrored on the UI type at `ui/src/lib/types.ts:1736-1738` and the server type
  at `src/types.ts:597-601`.
- **Doc anchor.** `#the-claude-code-install-row` (`ui/src/lib/diagnostics-docs.ts:30-36`) slugs from
  the heading `### The Claude Code install row` (`docs-site/src/content/docs/operating.md:134`), and
  `test/host-capacity-doc-anchor.test.ts:41-52` slugs *every* `operating/#…` link in `DOC_LINKS`
  against the real headings, so the coupling is enforced, not asserted in prose.
- **Spawn pin placement.** `src/herdr.ts:620` computes `argv[innerCommandStart(argv)] === "claude"`,
  and `innerCommandStart` (`:631-633`) is the same `bwrap … --` resolution `isHeadlessCodexExec`
  used before and still uses (`:638-641`) — a shared helper, not a duplicated one. The token is
  emitted at `:626` before `...envTokens`, so a caller-supplied value would override it (`env` takes
  the last assignment). Argv construction uses the bare token `"claude"`
  (`src/transient-agent-argv.ts:251`), matching the comparison. `src/spawn-budget.ts:39` re-runs
  `buildWrappedArgv`, so the extra 24 bytes are priced into the E2BIG budget rather than silently
  overflowing it.
- **Blast radius of a failing probe.** `claudeInstallCheck` swallows rejections
  (`src/diagnostics.ts:1369-1372`) and is only awaited when `claudeOk` (`:1361`), so the
  `claude`/`codex`/`claude_trust` rows are unaffected; the unwired default is a *rejecting* promise
  (`src/diagnostics.ts:409-410`) rather than a fabricated "not managed" state, which keeps every
  existing test's snapshot byte-identical without ever spawning `mise` from a test.
- **Adaptive re-check cadence.** `nextDiagnosticsDelay` accelerates only on `error`
  (`src/diagnostics.ts:242-245`), so a permanently-`warning` `claude_install` row does not pin the
  poller at the 60s recheck interval — the three extra execs stay on the 6h/`DIAGNOSTICS_TTL_MS`
  cadence.
- **Refactor of the claude/codex rows** into `agentCliCheck` (`src/diagnostics.ts:331-344`) is
  behaviour-preserving: same ids, same `optional`/`error` selection on `anyPresent`, same hint keys
  (`diagnostics_hint_{id}_{ok,optional,missing}` — all four pre-existing keys still present at
  `ui/messages/en.json:1364-1366` and the codex block below it).

Both findings below are factual-accuracy defects in prose that this PR introduces; the executable
behaviour is correct.

## Findings

### 1. `src/diagnostics.ts:147` still describes the wired read as TTL-cached — the exact claim the round-1 defect hid behind

The `readMiseClaude` dep doc says "the real read is the shared **TTL-cached** probe wired in
`index.ts`". It is not: `refreshMiseClaude` (`src/mise-claude.ts:231-240`) carries an explicit
"Deliberately UNCACHED" contract, and the wiring comment eleven lines above the call site says
"Uncached on purpose: `check()` promises a forced fresh run" (`src/index.ts:2839`). The two comments
in the same PR contradict each other, and the stale one asserts precisely the property whose removal
was the point of `e025ec93` — leaving it is an invitation to re-add the cache. The same stale wording
survives in the test helper doc at `test/herdr.test.ts:1698` ("via the real **cached** probe").

### 2. `docs-site/src/content/docs/operating.md:152` states an equivalence the code does not hold

"The row is `ok` **exactly when** the auto-updater pin is in effect" is false in one direction. The
pin is on for both the `ok` and the `residue` verdicts — `pinFor` requires only `managed`, equal
versions and `!nativeOnPath` (`src/mise-claude.ts:212-219`), while `verdictFor` downgrades a pinned
host to `residue` as soon as `nativeResidue` is non-null (`src/mise-claude.ts:203`). The module's own
comment states the real invariant correctly ("`ok`/`residue` are the pinned states",
`src/mise-claude.ts:68-70`), and `test/mise-claude.test.ts:159-160` asserts `verdictFor === "residue"`
with `pinFor === true` on the same state. So on a residue host the operator reads a `warning` row and
concludes from this sentence that Claude Code is still self-updating, when Shepherd has in fact
pinned it. The one-directional claim ("the row is never `ok` unless the pin is in effect") is the
true one.

## Nits (non-blocking):

- `src/mise-claude.ts:113` uses `access(candidate, X_OK)` to decide "an executable named claude is here"; a *directory* named `claude` in a PATH dir also passes (dirs are typically `X_OK`). It fails closed via the `targets.size !== 1` gate, so nothing breaks, but an `isFile()`-style check would say what is meant.
- `src/mise-claude.ts:162` resolves `home` from `os.homedir()`, while `src/diagnostics.ts:396` resolves the sibling claude-config path from `process.env.HOME ?? ""`. Same host, two sources of truth for `$HOME`.
- `src/mise-claude.ts:180` anchors the native tree at `$HOME/.local/share/claude` only; a host with a non-default `XDG_DATA_HOME` would read as "not native".

- `src/herdr.ts:626` emits `DISABLE_AUTOUPDATER=1` into the outer `env` shim even for a `bwrap …  -- claude` argv, where `--clearenv` wipes it (`test/herdr.test.ts:1738-1741` asserts the token is present there). Harmless today and consistent with how `NODE_COMPILE_CACHE` is handled, but if the membrane ever grows a `--setenv` passthrough, the "sandboxed spawns don't need it" reasoning at `src/herdr.ts:598-601` and the token's presence will need reconciling.

## Design limitation I checked but am NOT raising as a finding

`pinFor` treats "on-PATH claude reports mise's version, and is not under `~/.local/share/claude/`"
as "mise owns what we run". A *third* install shape — e.g. an npm-global claude whose realpath is
`~/.local/lib/node_modules/@anthropic-ai/claude-code/cli.js`, coexisting with a mise-managed claude
that is not itself on the service PATH, at the identical version — satisfies both conditions, so
Shepherd would pin (freeze) it and the row would report `ok`. That is the same class of failure the
`nativeOnPath` guard exists to prevent (`src/mise-claude.ts:206-219`), just outside the one directory
the guard knows. I am recording it rather than filing it because it is inherent to the
version-equality ownership test the approved plan settled on deliberately (and which round 1 did not
contest), the conjunction required is narrow, and it self-corrects at the next `mise upgrade claude`
(the versions diverge, the verdict flips to `diverged`, the pin drops). If the ownership test is ever
revisited, this is the case to close.

## Stated limitations

- I did not verify Claude Code's own native-install on-disk layout (whether
  `~/.local/share/claude/versions/*` entries are regular files, which `defaultListNative`'s
  `st.isFile()` filter at `src/mise-claude.ts:136` requires for the residue rollup to be non-empty).
  That ground truth is not in this repo. `src/sandbox.ts:562-565` describes the tree as "claude
  native binary + launcher symlink", which is consistent with files, but it is not proof.
- `DISABLE_AUTOUPDATER` is Claude Code's own env var; its semantics are external to this repo and I
  assumed them as written. There is no other occurrence of the name in the tree to cross-check
  against.
- Per the instructions I did not run lint or tests.

PR #2120, opened 2026-08-22
fix(sandbox): prove the agent binary launches in the membrane (#2111)

### LINKED ISSUE #2111: Sandbox self-test proves node+git, not the agent binary — a broken launcher reads as a healthy membrane
-

scoop:
Closes #2111.

`detectBackend`'s self-test runs `node --version && git --version` plus a session-env write through the real derived membrane. It never launches the **agent** binary. That is why the mise/EROFS launcher fault behind #2110 read as a perfectly healthy sandbox on this host while every wrapped role — plan gate, PR critic, doc agent, standalone critic — died at launch, waited out its whole timeout, recorded `no-verdict`, and respawned into the same wall.

## Why not just add `claude --version` to that probe

Because `detectBackend`'s `null` means **run unconfined**. Folding a launcher fault into it would silently strip the sandbox from around untrusted plan text on precisely the hosts that *can* sandbox — a worse bug than the one being fixed. #2110 named this and filed it rather than doing it.

So there are now two orthogonal signals that are never conflated:

| | question | on failure |
| --- | --- | --- |
| `detectBackend` (unchanged) | can bwrap build a sandbox here? | membrane not applied — still the sole gate on confinement |
| `probeMembraneLaunch` (new) | does the agent binary start inside it? | loud + blocking, **never** a confinement change |

`sandbox.ts` gains only a doc comment recording the split, so the next reader doesn't "fix" this the obvious wrong way. A test pins `detectBackend`'s argv and spawn count so re-introducing an agent launch there fails loudly.

## What the new probe does

Runs `<agent> --version` through the real derived membrane (`membraneForArgv` attaches the codex binds off `argv[0]`), TTL-cached per binary for 60s so a repaired host un-blocks without a Shepherd restart — unlike `detectBackend`, whose verdict is pinned for the process lifetime.

- **`sandbox_membrane` DIAGNOSE row.** Conditional: no row when there is no bwrap backend or no agent CLI. Guidance-only, like `claude_install` — the fix is host-side toolchain surgery, not something Shepherd does behind a button. Joins `STEADY_STATE_ERROR_CHECK_IDS`: a broken toolchain persists for hours and each re-check costs a real spawn, so accelerating would fast-poll forever.
- **Fail-fast at `resolveAuxPatch`**, before hooks, the trust seed and the pane are paid for. Per-binary: a host whose `claude` launcher is broken keeps running codex-configured roles.

**Fail-open by construction.** Only a NON-ZERO EXIT counts as broken. A probe that throws or times out is `uninspectable` — the row goes `optional` and spawns proceed. A flaky probe must never be able to block every reviewer role on a healthy host.

## Making the refusal visible, not silent

`resolveAuxPatch`'s existing abort path would have traded one invisible failure for another: the plan gate's branch just logs and returns `"skipped"`. The refusal is a separate discriminant on the return union, so the compiler forced all four callers to decide:

- **plan gate** — persists a visible `error` gate carrying a `membrane-launch` sentinel, churn-guarded on planHash + code (`consider()` deliberately re-attempts an `error` gate).
- **PR critic** — the same sentinel through `publishSpawnAbort`, so the badge tooltip is localized.
- **standalone critic / doc agent** — log-only and a distinct skip reason; both are session-less or have no badge, matching how they already treat an argv-budget refusal.

Follows the established `no-verdict-*` contract: the server ships a code, the UI owns the prose, no English in the DB. The launcher's own output is the whole diagnosis but carries absolute host paths, which `hintParams` bans — so it goes to the log, and the payload carries the agent name only.

## Verification

Measured against the real `buildMembraneFlags` output on this host:

| probe | exit | wall |
| --- | --- | --- |
| `claude --version` in membrane | 0 | 283 ms |
| `codex --version` in membrane | 0 | 124 ms |
| `claude --version`, `--tmpfs …/mise/shims` stripped (re-creates the pre-#2110 break) | 1 | 114 ms |

End-to-end through the shipped code, with the real (non-test) capture path:

```
detectBackend() = bwrap
readMembraneLaunchFacts (202ms) = {"backend":"bwrap","agents":[{"agent":"claude","state":"ok"},{"agent":"codex","state":"ok"}]}
DIAGNOSE row = {"id":"sandbox_membrane","state":"ok","hintKey":"diagnostics_hint_sandbox_membrane_ok"}

# ...then with the #2110 shims tmpfs stripped:
[sandbox] claude exits 1 inside the bwrap membrane — every wrapped role will refuse to spawn:
mise ERROR failed to rebuild shims
mise ERROR Read-only file system (os error 30)
DIAGNOSE row = {"id":"sandbox_membrane","state":"error","hintKey":"...broken","hintParams":{"agent":"claude"}}
detectBackend() STILL = bwrap        # a launcher fault never unwraps the membrane
```

`bun run lint`, `bunx tsc --noEmit`, `bun run test` (8797 pass / 0 fail), `cd ui && bun run check` (0 errors) `&& bun run test` (4665 pass) `&& bun run check:i18n`, `scripts/check-glossary.mjs`, and `bunx fallow audit --base origin/main --fail-on-issues` — all green.

## Notes for review

- **Tripwire.** `MembraneSeams.membraneLaunch` defaults to the REAL probe rather than a no-op, because a safety gate you have to remember gets forgotten. Its default `runCapture` throws `MembraneProbeUnwiredError` under `NODE_ENV=test`, and that throw is deliberately **exempt from fail-open** — otherwise an un-injected seam would resolve to `uninspectable`, spawns would proceed, and the test would stay green while spawning real bwrap on the test host. Plan review caught exactly this hole; the four test files that reach the seam were re-derived from the `resolveAuxPatch`/`resolveAuxSpawn` call sites (`sandbox-wiring`/`service` inject a bwrap backend but drive `SessionService.create`, which never reaches it).
- **Complexity extractions.** `unstarted()` / `alreadySurfaced()` in `review.ts` and `resolveUnwiredReads()` in `diagnostics.ts` exist to keep `fallow audit` green — the guard branches I added pushed three functions over the cognitive/cyclomatic thresholds.
- **Deliberately out of scope:** gating *session* spawns on the probe (a false negative there blocks all sandboxed work, and the observed damage is entirely in pane-hosted aux roles); `detectBackend`'s process-lifetime cache; making `installs/` writable (#2110 states why it must not be).
- **Accepted consequence:** on a genuinely broken host the plan gate now blocks *loudly* and planning sessions will not reach execution until the toolchain is repaired. That is already today's end-state — an unapproved gate — only now it says why.
- `[no-feature-entry]` — a bug fix plus a diagnostics row, no new user-facing capability.


scoop (commented):
# Review — #2111: prove the agent binary launches inside the membrane

Scope: `git diff 3830b70e...HEAD` (27 files). Read-only inspection; no builds/tests run.

## Verdict

The change does what the task asks and the two signals stay properly orthogonal: `detectBackend`
is behaviourally untouched (only a doc comment), and the new launch probe can block a wrapped aux
spawn without ever being able to produce `backend === null` (i.e. it can never downgrade a session
to unconfined). I found no correctness or security defect. One factually wrong comment, plus a
handful of non-blocking observations below.

**Every consumer of the widened `resolveAuxPatch` / `resolveAuxSpawn` return type handles the new
`{ refused }` arm.** Grepped all call sites: `src/review.ts:618` (→ `unstarted()` guard at
`src/review.ts:627`), `src/plan-gate.ts:988` (→ `src/plan-gate.ts:1002`), `src/standalone-critic.ts:509`
(→ `src/standalone-critic.ts:517`), `src/doc-agent.ts:780` (→ `src/doc-agent.ts:792`, relayed to the
one caller at `src/doc-agent.ts:665`). No other call sites exist in `src/`. `resolveAuxSpawn` forwards
it at `src/spawn-membrane.ts:410`.

**The refusal discriminator matches the real argv.** `membraneAgentOf` keys on `argv[0]`
(`src/membrane-launch.ts:41-46`); `buildTransientAgentArgv` emits the literal `"claude"` at
`src/transient-agent-argv.ts:250` and `codexRoleArgv` the literal `"codex"` at `src/codex-role-argv.ts:34`
— the same discriminator `membraneForArgv` uses at `src/sandbox.ts:296`. Per-binary refusal therefore
really is per-binary.

**The probe reproduces a real wrapped spawn.** `runProbe` builds
`["bwrap", ...buildMembraneFlags(membraneForArgv(inner, membrane)), "--", ...inner]`
(`src/membrane-launch.ts:196-215`), byte-shaped like `wrapArgv` at `src/sandbox.ts:666`, and reuses the
same `/tmp` stand-in that `detectBackend`'s own self-test uses (`src/sandbox.ts:141-150`).

**`detectBackend` is unchanged.** Diffed `src/sandbox.ts`: only the doc block at `src/sandbox.ts:117-123`
was added; the body at `src/sandbox.ts:125-166` is byte-identical, and `test/sandbox.test.ts:886-913`
pins argv + spawn count.

**Fail-open holds on the spawn path.** `membraneLaunchRefusal` returns null for `backend === null`,
for an argv the probe cannot speak to, and for `uninspectable` — only `broken` refuses
(`src/spawn-membrane.ts:288-305`). `defaultRunCapture` converts a timeout/kill/ENOENT into a throw
rather than a false non-zero status (`src/membrane-launch.ts:121-128`), and `runProbe` maps that to
`uninspectable` (`src/membrane-launch.ts:207-212`).

**Sentinel plumbing, server → store → UI.** `PlanSummaryCode`/`ReviewSummaryCode` widened at
`src/types.ts:621` and `src/types.ts:713-718`; mirrored in `ui/src/lib/types.ts:522` and
`ui/src/lib/types.ts:781-786`. Store coercion sets learn it at `src/store.ts:340-343` and
`src/store.ts:350-357`; no migration is needed because the `summaryCode` column already exists on both
tables (`src/store.ts:1283`, `src/store.ts:1301`). `criticbadge_membrane_launch` is mapped in the total
`Record<ReviewSummaryCode, …>` at `ui/src/lib/components/critic-badge.ts:12`, so the widened union is
exhaustive by construction; the empty-`summary` error verdict still renders because
`ui/src/lib/components/GitRail.svelte:825-826` gates on `summaryCode || summary` and routes through
`criticTitle`. PlanPanel does the same at `ui/src/lib/components/PlanPanel.svelte:446-449`.

**i18n parity is exact.** The same 8 keys exist in both catalogs at the same lines —
`ui/messages/en.json:613, 2672, 2673, 2996, 3478, 3479, 3480, 3481` and
`ui/messages/de.json:613, 2672, 2673, 2996, 3478, 3479, 3480, 3481` — and the `{agent}` placeholder
appears in both `diagnostics_hint_sandbox_membrane_broken` entries (`ui/messages/en.json:3480`,
`ui/messages/de.json:3480`). The glossary id `sandbox-membrane` (`ui/src/lib/glossary.ts:133-138`)
matches the `[[sandbox-membrane|…]]` references in those hints and follows the
`gloss_<id>_term`/`gloss_<id>_def` contract stated at `ui/src/lib/glossary.ts:1-8` (same
hyphen→underscore shape as `herdr-hygiene` at `ui/src/lib/glossary.ts:127-132`).

**No diagnostics id registry was missed.** `DiagnoseRows.svelte` resolves both the label and the hint
dynamically (`ui/src/lib/components/DiagnoseRows.svelte:58` and `:64`), so a new check id needs only its
message keys. `STEADY_STATE_ERROR_CHECK_IDS` is consumed solely by `nextDiagnosticsDelay`
(`src/diagnostics.ts:279-284`), so adding `sandbox_membrane` has exactly the claimed effect (no
acceleration to `recheckMs`). The conditional row is spliced only when non-empty
(`src/diagnostics.ts:1651`), so a host with no backend keeps a byte-identical snapshot.

**`hintParams` purity.** `classifyMembraneLaunch` ships `{ agent }` only (`src/diagnostics.ts:930-936`),
which satisfies the "never a token, absolute path, or identity" boundary at `src/types.ts:597-601`.
The launcher tail goes to `console.warn` (`src/membrane-launch.ts:216-221`) and into
`SpawnRefusal.reason`, which every role logs and none persists — `src/review.ts:1655` passes `""` as the
summary, `src/plan-gate.ts:1510` stores `summary: ""`.

**Plan-gate re-fire behaviour is sane, not a churn loop.** The refusal writes an `error` gate; the
settle-edge trigger explicitly excludes `error` priors on `idle` (`src/plan-gate.ts:97` and the
rationale at `src/plan-gate.ts:87-88`), so it does not spam re-reviews. The churn guard reads a live
row each time (`prior` comes from `store.getPlanGate` at `src/plan-gate.ts:696`), so
`src/plan-gate.ts:1505` genuinely suppresses the re-put and the broadcast.

**`doc-agent`'s unguarded `worktree.remove` on the refused path is safe.** Past
`src/doc-agent.ts:647-650` the worktree is guaranteed `isolated`, so `worktreePath !== repoPath`;
the new branch at `src/doc-agent.ts:665-668` mirrors the abort branch correctly.

**index.ts wiring resolves.** `homedir` and `safeRealpath` are each imported exactly once
(`src/index.ts:73-74`), `readMembraneLaunchFacts` once (`src/index.ts:75`), and
`DIAGNOSTICS_PROBE_TIMEOUT_MS` exists at `src/config.ts:104`. No import cycle: `config → sandbox`
(`src/config.ts:16`) and `membrane-launch → config, sandbox` (`src/membrane-launch.ts:23-31`), while
`sandbox.ts` imports neither.

## Findings

### 1. `src/membrane-launch.ts:138` — the `resetMembraneLaunchCache` doc names a caller that does not exist

The comment reads "Clear the TTL cache (tests, and the DIAGNOSE re-check path)". Nothing in `src/`
calls it: the only references in the tree are `test/membrane-launch.test.ts:5,37,127` and the
definition itself. The DIAGNOSE re-check goes through `deps.diagnostics.check(Date.now())`
(`src/server.ts:4738`), which forces the diagnostics probes but reads this module's own 60 s TTL cache
(`src/membrane-launch.ts:118`, `:161-166`) — exactly the residual the approved plan records ("a
DIAGNOSE Re-run within 60 s of the last probe reads the cache"). As written the comment tells the next
reader the opposite. Either wire the reset into the re-check path or drop that clause.

## Nits (non-blocking):

- `src/membrane-launch.ts:118`: `MEMBRANE_LAUNCH_TTL_MS = 60_000` is hardcoded while its own comment claims alignment with `DIAGNOSTICS_TTL_MS` (`src/config.ts:88`), a module this file already imports — importing the constant would remove the drift risk the comment is about.
- `src/membrane-launch.ts:126`: a launcher killed by a SIGNAL (segfault, SIGILL) yields `code: null` with `killed: false`, so it lands in `uninspectable` and fails open, even though it is as deterministic as a non-zero exit. Consistent with the documented "only a non-zero exit is broken" policy, but the doc at `src/membrane-launch.ts:56-58` describes `uninspectable` as "the probe itself could not run", which that case is not.
- `src/diagnostics.ts:923`: `classifyMembraneLaunch` is exported but only the all-`ok` path is exercised (via `healthyDeps` at `test/diagnostics.test.ts:83-92`); the `broken` → `error` row and the `uninspectable` → `optional` row — the two states the issue is actually about — have no test.
- `src/membrane-launch.ts:158-166`: no in-flight dedup, so N aux spawns racing on a cold cache each spawn their own `bwrap`. Bounded and cheap, but the "one bwrap spawn per minute, host-wide" claim only holds once the cache is warm.
- `src/membrane-launch.ts:196-205`: the probe membrane omits `maskCredentials`/`apiKeyHelperPath`, so under api-key auth it exercises the whole-dir `--ro-bind claudeDir` shape rather than the per-child masked shape a real spawn gets (`src/sandbox.ts:490-492`). Irrelevant to `--version`, but it means the probe is not quite the spawn's membrane in that mode.
- `src/plan-gate.ts:1507-1520`: the refusal row drops `blocks`, which `buildGate` carries forward (`src/plan-gate.ts:1589`), so a plan with question-form visual blocks degrades to flat markdown when the host breaks. `answeredQuestionKeys` is preserved, which then keys to blocks the row no longer has.
- No test covers the doc-agent refusal branch (`src/doc-agent.ts:665-668`, `:792-798`) — the only role whose refusal surfaces as a `DocAgentResult` rather than a log or a row.

- `src/diagnostics.ts:206-224`: `unwired()` + `resolveUnwiredReads()` refactor the pre-existing `readHerdrFleet` default into a shared helper. Behaviour-identical (same `Promise.reject(new Error("readHerdrFleet unwired"))`), and it does serve the new dep, but it is a drive-by change to code the task did not require touching.

## Stated limitations

- I could not confirm on this host that `codex --version` and `claude --version` actually exit 0 inside the derived membrane — the approved plan reports measuring both (283 ms / 124 ms), and nothing in the repo can verify a host fact. Assumed as stated.
- I did not run `bun test`, `tsc`, or `check:i18n` (CI's job per the review brief); the parity and identifier checks above were done by reading the catalogs and grepping the tree.

scoop:
# Re-review — #2111: prove the agent binary launches inside the membrane

Round 2 of at most 4. Read-only inspection of `git diff 3830b70e...HEAD` (27 files).

## Prior finding — status

**1. `src/membrane-launch.ts`: `resetMembraneLaunchCache`'s doc comment claimed callers that did not exist.** — **ADDRESSED.**
The doc at `src/membrane-launch.ts:138-139` now says "Called by `readMembraneLaunchFacts` (the DIAGNOSE read forces a fresh probe) and by tests", and `readMembraneLaunchFacts` really does call it at `src/membrane-launch.ts:269`. I traced the claimed chain end to end rather than taking the comment's word for it:

- `src/index.ts:75` imports `readMembraneLaunchFacts`; `src/index.ts:2845-2849` wires it as the `readMembraneLaunch` dep.
- `src/diagnostics.ts:1510-1513` (`membraneLaunchProbe`) calls `this.readMembraneLaunch()`; `src/diagnostics.ts:1643` invokes it from `check()`.
- `src/server.ts:4738` calls `deps.diagnostics.check(Date.now())` on `GET /api/diagnostics?refresh=1` — the DIAGNOSE "Re-run" path.

So the reset is on the re-check path, and `test/membrane-launch.test.ts:172-198` pins both halves of the behaviour (fresh probe, and the cache refill the refusal path reads). Finding closed.

## Findings

### 1. `src/membrane-launch.ts:266` — `readMembraneLaunchFacts` pins the process-wide backend cache from a *different* node resolution than every spawn uses

```ts
const backend = (deps.detectBackend ?? realDetectBackend)();
```

`realDetectBackend` is called with **no probe env**, even though this function is handed exactly the three values `BackendProbeDeps` wants (`env.claudeDir`, `env.home`, `env.nodeBinReal` — signature at `src/membrane-launch.ts:262-265`, interface at `src/membrane-launch.ts:60-64`; `BackendProbeDeps` at `src/sandbox.ts:82-86`).

Every other production caller passes them:

- `src/spawn-membrane.ts:84-88` → `{ home: homedir(), claudeDir: config.claudeDir, nodeBinReal: safeRealpath(config.nodeBin) }`
- `src/service.ts:2397-2400` → the same triple
- `src/index.ts:2845-2849` (this PR) passes the same triple into `readMembraneLaunchFacts` — and then that value is thrown away for the backend probe.

The defaults are **not** equivalent. `detectBackend`'s no-arg default is `safeRealpath(resolveNodeBin())` (`src/sandbox.ts:140`), while `config.nodeBin` is `resolveNodeBin({ override: process.env.SHEPHERD_NODE_BIN })` (`src/config.ts:572`). `resolveNodeBin` honours the override only when it is passed (`src/node-bin.ts:27-28`) — so on any host that sets `SHEPHERD_NODE_BIN` (the documented escape hatch for exactly the "systemd PATH excludes the version-managed node" case, `src/node-bin.ts:18-24`) the two resolutions diverge, and with them `dirname(nodeBinReal)`, which the membrane binds and puts on the sandbox PATH (`src/sandbox.ts:627-630`).

Why that matters here specifically: `detectBackend` caches its verdict for the whole process (`src/sandbox.ts:97`, `src/sandbox.ts:126`), so **whichever caller runs first decides for everyone**. Before this PR the diagnostics path never called `detectBackend` at all; now `diagnosticsTick` fires 4 s after boot (`src/index.ts:2883`) and on a freshly started server with no sessions yet it is very plausibly the first caller. If the alternate node resolution makes the wrapped `node --version && git --version` self-test fail (`src/sandbox.ts:155-162`), `_backendCache` is pinned to `null` — which means **run unconfined** for every session and every aux role, for the life of the process, with no way back short of a restart. That is precisely the silent-loss-of-confinement failure mode this PR's own `detectBackend` doc comment (`src/sandbox.ts:117-123`) exists to guard against.

Fix is one line: pass the env this function already holds — `realDetectBackend({ home: env.home, claudeDir: env.claudeDir, nodeBinReal: env.nodeBinReal })`.

(`src/drain.ts:444` has the same no-deps call today. That file is not in this diff, so it is listed under "Out of scope" below and is not part of this finding — but it is the reason the pattern looked safe to copy.)

## Everything else I verified (and found correct)

**Identifier resolution — the new module's imports and the symbols it relies on.**
`buildMembraneFlags`, `membraneForArgv`, `MembraneInputs`, `SandboxBackend`, `detectBackend`, `safeRealpath` all exist and are exported from `src/sandbox.ts` (lines 291, 174, 62, 125, 114 and the flag builder used by `wrapArgv` at `src/sandbox.ts:666`). `DIAGNOSTICS_PROBE_TIMEOUT_MS` exists at `src/config.ts:104` (5 000 ms). No import cycle: `src/config.ts:16` imports `./sandbox`, and nothing in `config`/`sandbox` imports `membrane-launch`.

**The probe reproduces the real wrapping byte-for-byte.**
`wrapArgv` builds `["bwrap", ...buildMembraneFlags(membraneForArgv(inner, membrane)), "--", ...inner]` (`src/sandbox.ts:664-666`); `runProbe` builds the identical shape at `src/membrane-launch.ts:186-215`. `membraneForArgv` keys the codex binds off `innerArgv[0] !== "codex"` (`src/sandbox.ts:296`), which matches `membraneAgentOf`'s `argv[0]` test (`src/membrane-launch.ts:47-52`).

**`argv[0]` really is the literal discriminator the refusal keys on.**
Claude roles: `src/transient-agent-argv.ts:250` pushes the literal `"claude"`. Codex roles: `src/codex-role-argv.ts:34` starts `["codex", "exec", …]`. `reviewerArgv` (`src/plan-gate.ts:427`) and the standalone critic (`src/standalone-critic.ts:491`) both go through `buildTransientAgentArgv`. So the per-binary claim holds.

**No `--chdir` divergence.** `buildMembraneFlags` emits no `--chdir` (`src/sandbox.ts:613-651`), and the pre-existing `detectBackend` self-test already spawns `bwrap` from the server's cwd with the same flag set (`src/sandbox.ts:155-162`) and returns 0 on real hosts — so the new probe's cwd handling is proven equivalent, not merely plausible.

**Every consumer of the widened `resolveAuxPatch` / `resolveAuxSpawn` return type handles `refused`.** The four production call sites are `src/review.ts:618`, `src/plan-gate.ts:988`, `src/standalone-critic.ts:509`, `src/doc-agent.ts:780`; each has an explicit branch (`src/review.ts:626-629` via the `unstarted()` predicate at `src/review.ts:148-151`, `src/plan-gate.ts:1002-1011`, `src/standalone-critic.ts:517-523`, `src/doc-agent.ts:792-798`). `doc-agent`'s new `"refused"` sentinel has exactly one caller (`src/doc-agent.ts:654`) and it is handled at `src/doc-agent.ts:665-668`, before the `!spawned` branch — no double `worktree.remove` (the inner function returns without reaping; the caller reaps).

**Locale parity — both catalogs, all eight new keys, at identical line numbers.**
`criticbadge_membrane_launch` (en:613 / de:613), `gloss_sandbox_membrane_term` (en:2672 / de:2672), `gloss_sandbox_membrane_def` (en:2673 / de:2673), `planpanel_membrane_launch` (en:2996 / de:2996), `diagnostics_label_sandbox_membrane` (en:3478 / de:3478), `diagnostics_hint_sandbox_membrane_ok` (en:3479 / de:3479), `_broken` (en:3480 / de:3480), `_uninspectable` (en:3481 / de:3481). The `{agent}` placeholder is present in both `diagnostics_hint_sandbox_membrane_broken` strings. `ui/scripts/check-i18n.mjs:44-53` only requires identical non-empty key sets across catalogs, which holds.

**Glossary integrity.** The `[[sandbox-membrane|…]]` markers in the three new hints resolve to the term registered at `ui/src/lib/glossary.ts:133-138`, whose `termKey`/`bodyKey` exist in both catalogs (verified above). `scripts/check-glossary.mjs:105-117` asserts exactly that; `scripts/check-glossary.mjs:44-46` splits the registry on `},\s*{`, and the new entry is a plain adjacent object literal with no comment between it and its neighbours, so the parser sees it.

**Message-key wiring, both directions.** `diagnostics_label_sandbox_membrane` and the three hint keys are consumed by the dynamic lookups at `ui/src/lib/components/DiagnoseRows.svelte:58` and `:63` (`msg[check.hintKey]?.(check.hintParams)`), so `hintParams: { agent }` (`src/diagnostics.ts:931`) interpolates. `criticbadge_membrane_launch` is mapped at `ui/src/lib/components/critic-badge.ts:12` inside a `Record<ReviewSummaryCode, …>`, which is exhaustive over the widened union — so a future member cannot be forgotten silently. `planpanel_membrane_launch` is reached from `ui/src/lib/components/PlanPanel.svelte:321-327` and rendered at `:446-448`.

**Union mirroring, server ↔ UI ↔ store.** `PlanSummaryCode` gains `"membrane-launch"` at `src/types.ts:621` and `ui/src/lib/types.ts:522`; `ReviewSummaryCode` at `src/types.ts:715-720` and `ui/src/lib/types.ts:781-786`. Persistence accepts both: `PLAN_SUMMARY_CODES` (`src/store.ts:340-343`) and `REVIEW_SUMMARY_CODES` (`src/store.ts:350-357`). Both legacy backfill migrations are gated on exact English prose (`src/store.ts:4553-4557`, `src/store.ts:4581-4585`), so a refusal row (`summary: ""`) can never be clobbered by them.

**Payload purity holds.** `classifyMembraneLaunch` puts only `{ agent }` into `hintParams` (`src/diagnostics.ts:931`), matching the stated ban on paths/identity at `src/types.ts:597-601`; the captured tail with absolute `$HOME` paths goes to `console.warn` only (`src/membrane-launch.ts:229-232`) and to the role logs (`src/plan-gate.ts:1007`, `src/review.ts:1650-1652`, `src/standalone-critic.ts:521`, `src/doc-agent.ts:795`) — never into a `PlanGate.summary`, `ReviewVerdict.summary`, or a diagnostics payload. `SpawnRefusal.reason` is never stored.

**Row placement and consumers.** `checks.splice(4, 0, …agentCliChecks, …membraneChecks)` (`src/diagnostics.ts:1651`) inserts after `gh` (index 3 in `probes()`, `src/diagnostics.ts:1552-1562`), i.e. claude / codex / sandbox_membrane / bun. No UI consumer indexes `checks` positionally — the only id-keyed reads are `ui/src/lib/provider-capacity.ts:18` (`claude`/`codex`) and `ui/src/lib/components/settings/SettingsDiagnosePanel.svelte:64` (fix by id, and this row is guidance-only so it has neither `remediation` nor `fixActionKey`).

**Fail-open is real, not asserted.** Only `broken` refuses (`src/spawn-membrane.ts:298`); `backend === null` (`src/spawn-membrane.ts:294`) and a non-agent argv (`:296`) short-circuit before the probe; a throwing/timing-out capture becomes `uninspectable` (`src/membrane-launch.ts:220-224`) and `defaultRunCapture` re-throws rather than fabricating a status when `killed` is set or `code` is not numeric (`src/membrane-launch.ts:124`), so a 5 s timeout can never masquerade as a non-zero exit.

**Churn guards.** Plan gate: `prior?.planHash === planHash && prior.summaryCode === summaryCode` (`src/plan-gate.ts:1505`) — and `consider()` deliberately re-attempts an `error` gate (`src/plan-gate.ts:710`), so the guard is load-bearing and correctly keyed. Review: `alreadySurfaced` now also compares `(prior.summaryCode ?? null) === summaryCode` (`src/review.ts:156-167`), which is a strict tightening of the old three-way check and preserves the pre-existing behaviour for plugin aborts (`summaryCode` defaults to `null`, and `putReview` already writes `v.summaryCode ?? null` at `src/store.ts:3214`).

**`detectBackend` is behaviourally untouched.** The only edit to `src/sandbox.ts` is the doc comment at lines 117-123; the function body (`src/sandbox.ts:125-166`) is unchanged in the diff, and `test/sandbox.test.ts:886-913` pins the spawn count at 2 and asserts no `claude`/`codex` command token in the probe.

## Limitations

- I could not confirm success criterion 7 (live confirmation on the author's host that the real probe reports `ok` for both binaries, and reports the EROFS failure with the `mise/shims` tmpfs stripped). That is a host observation, not something present in this repo — recorded as stated, neither confirmed nor disputed.
- I did not run `bun test`, `tsc`, or the i18n/glossary gates; per the brief, CI owns those. Where a claim depended on a test's *content* I read the test file rather than executing it.

- `src/drain.ts:444` calls `detectBackend()` with no probe env, so it too can pin the process-wide backend cache from a `SHEPHERD_NODE_BIN`-blind node resolution — the same divergence as finding 1, in a file this PR does not touch.
- `detectBackend`'s verdict is cached for the entire process lifetime (`src/sandbox.ts:97`, `src/sandbox.ts:126`), so a host repaired at runtime keeps running unconfined until Shepherd restarts; the plan explicitly lists this as out of scope.

- `src/diagnostics.ts:206-227` extracts a generic `unwired<T>(name)` helper and a `resolveUnwiredReads` bundler, refactoring the pre-existing `readHerdrFleet` inline default that the task did not require touching. Behaviour is identical (`() => Promise.reject(new Error("<name> unwired"))` both before and after), so this is a legitimate author call, not a defect.

- `src/membrane-launch.ts:156-172`: `probeMembraneLaunch` has no in-flight de-duplication — N aux spawns racing on a cold cache each spawn their own `bwrap` + `<agent> --version`. Bounded today by how many aux roles can start at once, but it grows with concurrency and the TTL does not help on a cold start.
- `src/membrane-launch.ts:269`: the cache reset is unconditional on every `DiagnosticsService.check()`, including the TTL-driven refresh inside `current()` (`DIAGNOSTICS_TTL_MS` = 60 s, `src/config.ts:88`) — so any client `GET /api/diagnostics` more than a minute after the last check re-spawns both probes, not just the explicit `?refresh=1` Re-run the doc comment describes. Harmless today because the UI only fetches on mount/retry (`ui/src/routes/+page.svelte:1784`), but a future poller would turn this into a per-poll pair of bwrap spawns.

## Nits (non-blocking):

- `src/diagnostics.ts:253-257`: the new `sandbox_membrane` rationale sits in a `//` block *between* the closing `*/` of the JSDoc and the `const`, splitting one explanation across two comment forms.
- `ui/src/lib/diagnostics-docs.ts`: every other guidance-only non-ok row (tailscale, host_capacity, claude_install) carries a `DOC_LINKS` entry; `diagnostics_hint_sandbox_membrane_broken` has none, so its "the exact error is in Shepherd's log" hint is the operator's only lead.
- `src/plan-gate.ts:1010`: the refusal path returns `"skipped"` even though it persisted a visible `error` gate, so the on-demand "Review plan now" route reports a no-op for something that did change state.
- The plan's success criterion 5 asks for a test that drives `resolveAuxPatch` with an *un-injected* `membraneLaunch` seam and asserts the tripwire throw; only the probe-level equivalent exists (`test/membrane-launch.test.ts:136-141`). The mechanism is covered, the integration-level assertion is not.

scoop:
Fixed in a9e67b11 — the finding was correct and the reasoning held at every link, which I verified rather than assumed: `config.nodeBin` = `resolveNodeBin({ override: process.env.SHEPHERD_NODE_BIN })` (config.ts:572) vs. the bare `resolveNodeBin()` behind `detectBackend`'s default, which never reads that env; `_backendCache` pins for the process; `diagnosticsTick` fires at 4s.

Demonstrated concretely rather than argued — with `SHEPHERD_NODE_BIN=/nonexistent/pinned/node`, a bare `resolveNodeBin()` returns `/home/linuxbrew/.linuxbrew/bin/node`. Two different binaries, and a false `null` from that self-test means RUN UNCONFINED for the life of the process.

The seam is now typed `(probe: BackendProbeDeps) => SandboxBackend` so a test can assert what is forwarded; I confirmed that test fails against the old bare call before keeping it.

**One thing beyond the finding, flagged for your judgement:** auditing the other call sites showed `drain.ts:444` had the *identical* bare call — `SessionService` and `resolveBackend` both already passed the env, so drain was the only remaining hole. Pre-existing, not introduced by this PR. I fixed it in the same commit (one line, same shape as its siblings) because leaving a known unconfined-execution path in the same mechanism this PR hardens seemed worse than the small scope expansion. Happy to split it into its own PR if you would rather keep this diff strictly scoped.

scoop (commented):
# Re-review — #2111: prove the agent binary launches inside the membrane (round 3)

## Prior finding — status

**1. `src/membrane-launch.ts:266` — `readMembraneLaunchFacts` called `realDetectBackend()` with no probe env.**
**ADDRESSED.** `src/membrane-launch.ts:274-278` now calls `(deps.detectBackend ?? realDetectBackend)({ home: env.home, claudeDir: env.claudeDir, nodeBinReal: env.nodeBinReal })` — exactly the shape the finding asked for. The seam is typed `(probe: BackendProbeDeps) => SandboxBackend` at `src/membrane-launch.ts:246`, so a test can assert what is forwarded, and `BackendProbeDeps` (`src/sandbox.ts:84-88`) accepts `home` / `claudeDir` / `nodeBinReal`. The env passed by the only production wiring (`src/index.ts:2844-2849`) is `{ claudeDir: config.claudeDir, home: homedir(), nodeBinReal: safeRealpath(config.nodeBin) }` — byte-identical to what `resolveBackend` uses at `src/spawn-membrane.ts:84-88`, so the 4s-after-boot diagnostics tick can no longer pin `_backendCache` (`src/sandbox.ts:101,126,164`) from a different node than real spawns use. The `SHEPHERD_NODE_BIN` divergence the finding named is real and still there in the bare path (`src/config.ts:572` reads the override; `src/node-bin.ts` default does not), but no in-diff caller takes it any more.

I did NOT re-raise it.

## Verification performed

**Identifier / contract resolution**
- `resolveAuxPatch`'s widened return `AuxPatch | {aborted} | {refused}` (`src/spawn-membrane.ts:312-314`) — grepped every caller: `src/review.ts:618`, `src/plan-gate.ts:988`, `src/standalone-critic.ts:509`, `src/doc-agent.ts:780` (via `resolveAuxSpawn`, `src/spawn-membrane.ts:403-412`). All four narrow on `"refused" in …` before touching `AuxPatch` fields: `src/review.ts:627-630` (via the `unstarted()` predicate at `src/review.ts:148-153`), `src/plan-gate.ts:1002-1011`, `src/standalone-critic.ts:517-523`, `src/doc-agent.ts:792-798`. No caller was left on the old two-arm union.
- `membraneAgentOf` keys on `argv[0]`. Confirmed argv[0] is the literal `"claude"` (`src/transient-agent-argv.ts:251`) or `"codex"` (`src/codex-role-argv.ts:34`) — the same discriminator `membraneForArgv` uses (`src/sandbox.ts:296`).
- The probe reproduces the real wrapping: `bwrap` + `buildMembraneFlags(membraneForArgv(inner, membrane))` + `--` + inner (`src/membrane-launch.ts:212-216`), matching `wrapArgv` at `src/sandbox.ts:665-666`. Aux spawns always use `profile: "standard"` (`src/spawn-membrane.ts:155`), so "wrapped iff backend !== null" and the refusal's `backend === null → null` short-circuit (`src/spawn-membrane.ts:295`) agree.
- `PlanSummaryCode` / `ReviewSummaryCode` widening (`src/types.ts:618,712-717`) mirrored in `ui/src/lib/types.ts:522,781-786`. Persistence: `PLAN_SUMMARY_CODES` (`src/store.ts:340-343`) and `REVIEW_SUMMARY_CODES` (`src/store.ts:350-357`) both carry `"membrane-launch"`; the columns already exist (`src/store.ts:1283,1301`) and both writers pass the field through (`src/store.ts:3214,3351`), so no migration is needed — matches the round-trip tests at `test/store.test.ts:643` and `test/store-plan-gate.test.ts:76-88`.
- `hintParams: { agent }` (`src/diagnostics.ts:929`) is `Record<string,string>`-compatible (`src/types.ts:601`) and is consumed by the dynamic lookup at `ui/src/lib/components/DiagnoseRows.svelte:64`. The row label resolves through `diagnostics_label_${id}` at `ui/src/lib/components/DiagnoseRows.svelte:58`, so `id: "sandbox_membrane"` needs `diagnostics_label_sandbox_membrane` — present in both catalogs.

**Locale parity** — `ui/messages/` holds exactly two catalogs (`en.json`, `de.json`), and `ui/scripts/check-i18n.mjs:44-54` fails on any key present in one and absent in the other. Both files gain the same 8 keys: `criticbadge_membrane_launch` (`ui/messages/en.json:613` / `ui/messages/de.json:613`), `gloss_sandbox_membrane_term` + `gloss_sandbox_membrane_def` (`en.json:2672-2673` / `de.json:2672-2673`), `planpanel_membrane_launch` (`en.json:2996` / `de.json:2996`), `diagnostics_label_sandbox_membrane` + the three `diagnostics_hint_sandbox_membrane_{ok,broken,uninspectable}` (`en.json:3478-3481` / `de.json:3478-3481`). No empty values. The `{agent}` placeholder appears in the `broken` hint in both locales. The `[[sandbox-membrane|…]]` marker in all three hints resolves to the new glossary id at `ui/src/lib/glossary.ts:133-138`, matching the existing marker convention (`ui/messages/en.json:1381` uses `[[host-capacity|…]]` against `glossary.ts`'s `host-capacity`).

**UI consumption of the new sentinel** — `critic-badge.ts:5-13` is a total `Record<ReviewSummaryCode, …>`, so the widened union forced the new arm; `criticTitle` (`ui/src/lib/components/critic-badge.ts:24-26`) falls back to `summary` then the generic title for an unknown code, so a newer server against an older UI degrades rather than blanks. `GitRail.svelte:825` gates on `verdict.summaryCode || verdict.summary`, which is satisfied because `publishSpawnAbort` writes `summaryCode` with `summary: ""` (`src/review.ts:1670-1676`). `PlanPanel.svelte:444-449` uses the same shape via `summaryForCode` (`ui/src/lib/components/PlanPanel.svelte:324-329`).

**Payload purity** — the captured launcher tail is bounded to 2000 chars (`src/membrane-launch.ts:120,220`) and never leaves the process except through `console.warn` (`src/membrane-launch.ts:225-228`). `SpawnRefusal.reason` (which embeds it) is logged only: `src/plan-gate.ts:1008` (`console.warn`), `src/review.ts:1650-1652` (`console.warn`), `src/standalone-critic.ts:521` (`this.log`, defaulting to `console.log` at `src/standalone-critic.ts:153`). `src/doc-agent.ts:797` returns a fixed English `"membrane refused spawn"`, not the tail, matching the neighbouring `"plugin aborted spawn"` at `src/doc-agent.ts:662`. Nothing carries a path into `DiagnosticCheck` — `classifyMembraneLaunch` emits only `{id,state,hintKey,hintParams:{agent}}` (`src/diagnostics.ts:922-940`).

**Fail-open / fail-closed** — only `status !== 0` maps to `broken` (`src/membrane-launch.ts:218-231`); a throw or a `killed` timeout is re-raised out of `defaultRunCapture` (`src/membrane-launch.ts:157-160`) and caught into `uninspectable` (`src/membrane-launch.ts:218-222`), which `membraneLaunchRefusal` treats as proceed (`src/spawn-membrane.ts:299`). `MembraneProbeUnwiredError` is re-thrown ahead of that catch (`src/membrane-launch.ts:219`), so the test tripwire cannot be swallowed. `detectBackend` is untouched behaviourally — the only edit is the doc block at `src/sandbox.ts:118-123`, and `test/sandbox.test.ts:886-913` pins argv + spawn count.

**Steady-state scheduling** — `sandbox_membrane` is in `STEADY_STATE_ERROR_CHECK_IDS` (`src/diagnostics.ts:258-262`), which `nextDiagnosticsDelay` (`src/diagnostics.ts:280-283`) reads against the merged snapshot list, so a broken launcher alone does not switch the sweep to `DIAGNOSTICS_RECHECK_INTERVAL_MS` (`src/config.ts:101`).

**Plan-gate row shape** — `publishSpawnRefusal` (`src/plan-gate.ts:1498-1521`) writes the same optional-field subset `buildGate` does (`src/plan-gate.ts:1571-1600` also omits `dismissed` / `finalRoundPending`), carries `round` and `answeredQuestionKeys` forward, and is churn-guarded on `planHash + summaryCode`. `consider()`'s `prior.decision !== "error"` dedupe (`src/plan-gate.ts:710`) is what makes the re-attempt loop intentional, as the plan states.

## Cannot-verify / stated limitations

- I did not execute anything (read-only review), so the plan's success-criterion 7 (a live probe on the author's host reporting `ok` for both binaries and the EROFS tail with the shims tmpfs stripped) is unverifiable from the diff. Assumed as reported.
- Whether `codex --version` and `claude --version` exit 0 under the derived membrane on arbitrary hosts is a property of those third-party CLIs, not of this repo. Not asserted either way; the fail-open design bounds the damage if they don't.

## Findings

None blocking. Everything I found is below.

## Nits (non-blocking):

- `src/membrane-launch.ts:281` — the DIAGNOSE read clears the whole TTL cache and re-probes both binaries on every `check()`. `nextDiagnosticsDelay` only exempts `sandbox_membrane` itself, so on a host where any *other* check sits in `error` (e.g. `gh` unauthenticated) the sweep runs at `DIAGNOSTICS_RECHECK_INTERVAL_MS` = 60s (`src/config.ts:101`) and this becomes two real `bwrap` + `<agent> --version` launches per minute, indefinitely — more than the "one bwrap spawn per minute, host-wide, worst case" the plan budgeted. Cheap (~0.4s), but consider honouring the TTL on the background sweep and forcing only on `?refresh=1`.
- `src/plan-gate.ts:1010` — the refusal returns `"skipped"` from `begin()`, so the manual "Review plan now" route reports a no-op to the operator at the same moment a red `error` gate row appears. Consistent with the adjacent plugin-abort branch, but `PlanReviewTrigger`'s doc (`src/plan-gate.ts:682-684`) reserves `"error"` for "a spawn was attempted but failed", which arguably fits better here.
- `src/membrane-launch.ts:186` — `probeMembraneLaunch` returns `{ state: "ok" }` for `backend === null`, i.e. "not applicable" is spelled the same as "proven healthy". Both current callers short-circuit before it matters, but a third state (or a `null` return) would be harder to misread.
- `src/spawn-membrane.ts:317-320` — the refusal fires before `foldAuxPatch`, so a plugin that would have aborted the spawn anyway now sees its `onSpawn` hook skipped on a broken host. Deliberate per the comment (don't pay for hooks), just worth knowing the ordering changed.

- `src/drain.ts:441-457` — `DrainService.detectBackend` gains the probe env. This is a genuine pre-existing hole of exactly the class the PR hardens (a bare `detectBackend()` that could pin `_backendCache` from the wrong node and mean run-unconfined), one line, shaped like its siblings at `src/spawn-membrane.ts:84-88`; keeping it here is a defensible call, and `config` + `homedir` were resolved as already/newly imported without duplication (`src/drain.ts:64,168`).
- `src/diagnostics.ts:207-220` — `unwired<T>()` + `resolveUnwiredReads()` refactor the existing `readHerdrFleet` fallback into a shared helper. A drive-by tidy the task didn't require, but it is behaviour-preserving and keeps the constructor flat.
- `src/review.ts:148-165` — `unstarted()` and `alreadySurfaced()` extract logic that was previously inline in one place each. Reasonable given the second caller, just more surface than the task strictly needed.

- `src/membrane-launch.ts:196-210` — the probe deliberately uses the DEFAULT `claudeDir`, while a plugin-routed spawn (#1213) can be assembled against a different credential dir (`src/spawn-membrane.ts:143`). Documented as intentional and correct for a launcher-chain fact today; it would become wrong if the membrane's *launcher* binds ever came to depend on which credential dir is bound.
- `src/membrane-launch.ts:200-208` — the probe membrane omits `extraEnv` (`collectPassthroughEnv`), `apiKeyMembraneFields()` and `agentSupportPaths`, all of which a real aux spawn carries (`src/spawn-membrane.ts:145-153`). None affects whether the launcher starts today, but a future launch-time dependency on, say, `LANG` or the api-key helper bind would be invisible to this probe.

PR #2121, opened 2026-08-24
feat(uploads): attach any file type, explain every blocked drop

scoop:
## The question was wrong, and that's the finding

"Why can't I upload EPS files to the scratchpad?" — you can, and you always could. Two probes against the running instance:

- `POST /api/sessions/<id>/scratchpad/upload` with a real `.eps` (`application/postscript`) → **HTTP 200**, file written and listed.
- Chrome driven against the live UI, `.eps` pushed through the Files panel's own hidden file input → **uploaded, appeared in the listing.**

The Scratchpad Files tab has no MIME restriction at any layer. The "clicked Upload, nothing happened" symptom could not be reproduced through the app — the native file dialog most likely never opened, which on Wayland/Hyprland is `xdg-desktop-portal`, not Shepherd. Nothing here tries to work around that.

What the investigation *did* turn up is one tab over.

## In-session attach refused every non-media file, silently

Three stacked gates, no feedback from any of them:

- `ViewportTermControls.svelte` — `accept="image/*,video/*"` made an `.eps` unselectable in the picker.
- `Viewport.svelte` → `attachImages()` — filtered a dropped `FileList` to `image/*|video/*` and returned without a word when nothing survived.
- `src/uploads.ts` → `sessionAttachExtForMime()` — 415 server-side.

All three are gone. The session branch of `handleUpload()` now uses `uploadExtension()`, exactly like New Task staging: extension derived from the filename, `SAFE_EXT_RE`-sanitised (`bin` otherwise), stored under a fresh UUID inside `.shepherd-uploads`. Size cap unchanged. Widened to *any* type rather than adding PostScript specifically — a narrow allowlist just recreates this dead end for the next unusual format, and the two sibling upload paths never had one.

`sessionAttachExtForMime()` and its allowlist are deleted (newly dead), and the now-misnamed `uploadImage` alias goes with them — `attachFiles()` calls `uploadFile()` directly.

## The Files panel no longer swallows a drop

Four cases used to be a mute `return`; each now says why: the read-only Worktree tab, inside the read-only `attachments/` overlay, while the listing is still loading, and — the one that likely bit you — **a drop carrying no file at all**, which is what a mail client or browser tab hands over when you drag an attachment straight out of it.

Two mechanics behind that:

- `handleDragOver()` now always `preventDefault()`s, including where upload is impossible. Without it the browser handles the drop itself and can navigate away from the HUD.
- It never sets `dropEffect = "none"`. That cancels the drag, so no `drop` event fires and every message above would be unreachable. A synthetic `DragEvent` can't prove this — it discards `dropEffect` writes — so the test records the assignment instead of reading the property back.

A 404 upload failure (session with no resolvable scratchpad target) now names its cause instead of the catch-all "Upload failed".

## Verification

- Root: `bun run lint`, `bunx tsc --noEmit`, `bun run test` (8800 pass) — all green.
- `ui/`: `bun run check` (0 errors), `bun run test` (4673 pass, real Chromium).
- `fallow audit --base origin/main --fail-on-issues` — no issues in the 12 changed files.
- The widened attach route exercised over a real HTTP wire: `.eps` → 200, stored as `<uuid>.eps`. The same request returns 415 on `main`.
- Catalog parity green; EN + DE for every new key; feature-catalog fragment stamped 1.47.0 from `bun run next-version`.


scoop:
# Review — attach any file type (round 1)

Scope: `git diff e4d60f1491a1435daed5fc43e5420f5cd3d0a4d0...HEAD` (12 files).

**Server widening is safe and self-consistent.**
`handleUpload()` now derives the extension from `uploadExtension(file)` for both branches (`src/uploads.ts:177`). I read the helper: filename extension first, `SAFE_EXT_RE = /^[a-z0-9]{1,16}$/` sanitised, `bin` fallback, MIME fallback only for an extensionless name (`src/uploads.ts:59-71`, `src/uploads.ts:42`), and the stored name is a fresh UUID (`src/uploads.ts:89-91`), so the raw filename never reaches the FS — the in-diff comment at `src/uploads.ts:173-176` is accurate. The deleted `SESSION_ATTACH_MIME_EXT` was the only consumer that would have orphaned `VIDEO_MIME_EXT`; it is still referenced at `src/uploads.ts:32`, so no newly-dead code. `sessionAttachExtForMime` has no remaining references anywhere (grep across `src/`, `test/`, `ui/`, `extension/` — only historical mentions in `docs/superpowers/plans/…`).

**No new exposure class from arbitrary extensions.** The sibling scratchpad route already accepted every type (`src/server.ts:2619-2620`, and `placeScratchpadUpload()` does no type check at all — `src/scratchpad.ts:311-344`), and both download routes force `Content-Disposition: attachment` (`src/server.ts:2596-2597`, `src/scratchpad.ts:206-215`), so an uploaded `.html`/`.svg` is not served inline. The staging branch already used `uploadExtension`, so this diff does not widen the New Task path either. I also checked the capture extension: it posts to `/api/uploads` with no `?session=` (`extension/src/lib/transport.ts:105-115`), so the removed 415 was never on its path.

**Renames resolve everywhere.** `attachImages` → `attachFiles`: declaration `ui/src/lib/components/Viewport.svelte:1422`, call sites `:1447`, `:1995`, prop pass `:3152`; consumer prop declared at `ui/src/lib/components/viewport/ViewportTermControls.svelte:16` and typed at `:27`, used at `:167`. Grep for `attachImages` / `uploadImage` over `ui/src` and `src` returns nothing (only `CHANGELOG.md` and `docs/superpowers/**` history), so deleting the `uploadImage` alias (`ui/src/lib/api.ts:411` before, now removed) breaks no consumer and no `vi.mock` factory.

**i18n parity holds.** `viewport_attach_file` exists at `ui/messages/en.json:712` and `ui/messages/de.json:712`; the old `viewport_attach_image` is gone from both (grep returns no hit in either catalog, and no hit in `ui/src`). The six new Files-panel keys plus the two announcement keys sit at `ui/messages/en.json:2552-2559` and identically at `ui/messages/de.json:2552-2559`, with the same `{name}` placeholder in `files_upload_unavailable` on both sides. `ui/messages/` holds exactly these two catalogs, per `.claude/rules/i18n.md:19-22`.

**Feature-catalog fragment is well-formed.** Entries are auto-globbed (`ui/src/lib/feature-announcements.ts:46-49`), so no index edit is needed; the shape matches the sibling `ui/src/lib/feature-announcements/entries/v1.47.0-video-brief.ts:1-12`; `sinceVersion: "1.47.0"` is greater than the last release `1.46.0` (`.release-please-manifest.json:2`) and agrees with the filename, as the gate described at `ui/src/lib/feature-announcements.ts:11-17` requires; the id `attach-any-file` is unique (grep over `ui/src/lib`).

**Test scaffolding resolves.** `mkdirSync` is imported at `test/server.test.ts:4`, `tmpRoot` at `:27`, and `makeDeps()` really exposes the `SessionStore` used by the new `deps.store.create(...)` (`test/server.test.ts:41`, `:92`). `worktreeUploadsDir`/`existsSync` used by the new upload tests are imported at `test/uploads.test.ts:7`, `:19`. `clickWorktreeSegButton` is defined at `ui/src/lib/components/viewport/FilesPanel.browser.test.ts:174` (hoisted, so the new block at `:373` can call it), `ApiError(status, message)` matches the constructor at `ui/src/lib/api.ts:162`. I did not run tests or lint (CI owns that).

**Drop plumbing.** `uploadDisabled = readOnly || withinAttachments` (`FilesPanel.svelte:37`), and `handleDrop` now covers exactly those two plus `!listing` plus the empty-`files` case (`:254-275`), so no gate was lost when the single `uploadDisabled` early return went away. There is no ancestor drag handler over the panel that the unconditional `preventDefault()` could pre-empt — the only other `ondrop`/`ondragover` pairs in the app are `NewTask.svelte:1576-1583` and the sibling `term-mount` at `Viewport.svelte:2964-2971`, neither an ancestor of `.files`. Nothing in the app sets `draggable=` (grep over `ui/src`), so the new "no file in this drop" path is only reachable from real external drags and native text drags. The claim in the comment at `FilesPanel.svelte:236-238` that `dropEffect = "none"` suppresses the `drop` event is correct per the HTML drag-and-drop model, so the `"copy"` choice is deliberate.

## Findings

### 1. `FilesPanel.svelte` — a listing that *failed* is reported as "still loading"

`handleDrop()` treats `!listing` as "still loading" (`ui/src/lib/components/viewport/FilesPanel.svelte:265-268`), but `browse()` leaves `listing === null` and sets `error = true` when the fetch fails (`:121-135`), which is exactly the state the panel itself renders as a load error (`:396-399`). The drop then says `files_upload_loading` — "drop it again in a moment" (`ui/messages/en.json:2554`) — advice that will never come true.

This is not hypothetical, and it lands on the very case the new 404 message was written for: `scratchpadList()` 404s the root listing when `claudeSessionId` is blank (`src/server.ts:2568-2572`, and `resolveScratchpadUploadDir()` returns null for the same reason at `src/scratchpad.ts:238`). For such a session the Files tab never gets a listing, so the drop always short-circuits on `!listing` and the accurate `files_upload_unavailable` ("this session has no scratchpad to upload into", `ui/messages/en.json:2556`) is unreachable — it only fires from `uploadFiles()`'s catch (`:213-219`), which that drop never reaches. The Upload button is simultaneously disabled with `files_upload_disabled_title` ("hasn't loaded yet", `:340`). Net effect: in the one scenario the PR names as the actionable 404 cause, every gesture in the panel reports "not loaded yet, try again".

Fix: branch on the existing `error` state before the `!listing` case and use the no-scratchpad / load-failed wording there (both catalogs).

### 2. `Viewport.svelte` — the attach comment still describes an image/video-only function

The comment block the diff edited now opens with "upload image(s)/video(s) into this session's worktree" (`ui/src/lib/components/Viewport.svelte:1414`) and then, three lines later, states the opposite: "Any file type: an operator hands the agent whatever the work needs" (`:1419-1421`). The first sentence is factually wrong about `attachFiles()` as written at `:1422-1441`, and the block contradicts itself. Drop or reword that lead sentence.

## Nits (non-blocking)

- `ui/src/lib/components/viewport/ViewportTermControls.svelte:75` — the untouched row comment still reads "image upload frozen on the far left"; this diff makes it stale in the same way as finding 2.
- `ui/src/lib/components/viewport/FilesPanel.svelte:360-362` — the notice renders at the top of the scrolling `.files-scroll` container (`:481-489`), so on the Worktree tab with a long listing the user can be scrolled past it and the drop still looks silent.
- `ui/src/lib/components/viewport/FilesPanel.svelte:217-218` — `files_upload_unavailable` asserts one cause for every 404, but this route also 404s for an archived/unknown session (`src/server.ts:2630`) and for a current subdirectory that vanished server-side (`src/scratchpad.ts:262-266`); the wording is then confidently wrong rather than merely generic.
- `ui/src/lib/components/viewport/FilesPanel.svelte:114-117` — the comment says the notice "never outlives the gesture it explains", but `browse()` does not clear `dropNotice`, so an "Attachments is read-only" line survives navigating back to the scratchpad root.
- `ui/src/lib/components/viewport/FilesPanel.svelte:360-362` — `role="status"` is inserted together with its text; screen readers are unreliable about announcing a live region that appears in the same tick as its content. A permanently-present, empty container would announce reliably.
- `ui/src/lib/components/Viewport.svelte:1424` — the terminal attach path still returns silently when `!conn`, so a drop on the terminal of a detached session is the same "nothing happened" the PR set out to remove elsewhere. (Retained deliberately per the plan; flagging for symmetry.)
- `ui/src/lib/components/viewport/FilesPanel.svelte:241` — `dropEffect = "copy"` over a target that will refuse the drop shows a copy cursor and then declines; unavoidable given that `"none"` would suppress the `drop` event, but worth a word in the comment.

- `ui/src/lib/api.ts:411` — deleting the `uploadImage` alias is outside the plan's file table; it is defensible (its last consumer was `Viewport.svelte` and the repo runs a dead-code gate), just not requested.
- `ui/src/lib/components/viewport/FilesPanel.svelte:270-273` — the "that drop carried no file" case is a fourth blocked case beyond the plan's three; harmless and in the spirit of the goal.
- `test/uploads.test.ts:283-311` — the hostile-name sanitisation test is new coverage the plan did not list; useful, and it does pin real behaviour (`../../evil.sh ell` → `bin`).

## Stated limitations

- I did not run `bun run lint`, `bun run test`, `bun run check`, or `fallow audit` — CI owns those, per the review brief.
- Success criteria 2–4 of the plan (a real browser: native picker offers every type; a drop on the Worktree tab fires our handler and never navigates the HUD away) cannot be verified from source. The synthetic-`DragEvent` tests in `FilesPanel.browser.test.ts:384-418` explicitly cannot prove it either, as their own comment says; assumed as reported.

scoop:
# Review — `feat(uploads): attach any file type, explain every blocked drop` (round 2)

Diff reviewed: `git diff e4d60f1491a1435daed5fc43e5420f5cd3d0a4d0...HEAD` (2 commits, 12 files).

## Prior-round points

**Prior 1 — `handleDrop` treated a FAILED listing as "still loading".** Addressed. `ui/src/lib/components/viewport/FilesPanel.svelte:265-268` now branches on `error` before the `!listing` case and renders `m.files_upload_load_failed()`; the button tooltip at `FilesPanel.svelte:344-348` picks the same key in the error case. The key exists in both catalogs — `ui/messages/en.json:2555` and `ui/messages/de.json:2555`. Covered by the new test at `ui/src/lib/components/viewport/FilesPanel.browser.test.ts:448-468`. See the new finding below, though: the branch as written is slightly too wide.

**Prior 2 — `attachFiles` comment opened with a factually wrong lead sentence.** Addressed. `ui/src/lib/components/Viewport.svelte:1414` now reads "Upload file(s) of ANY type into this session's worktree…" and no longer contradicts itself.

## Verification performed

- **Removed symbols have no live consumers.** Grepped the tree for `uploadImage`, `attachImages`, `sessionAttachExtForMime`, `viewport_attach_image`: the only remaining hits are `CHANGELOG.md:1947` and archived plan docs under `docs/superpowers/plans/` — no `.ts`/`.svelte` source references. `ui/src/lib/api.ts:395` still exports `uploadFile`, used at `ui/src/lib/components/Viewport.svelte:1428` and `ui/src/lib/components/NewTask.svelte:873`.
- **`VIDEO_MIME_EXT` did not become dead** when `SESSION_ATTACH_MIME_EXT` was deleted — it is still spread into `MIME_EXT` at `src/uploads.ts:32`.
- **Server widening is contained.** `src/uploads.ts:177` computes `ext` via `uploadExtension()` for both branches; `uploadExtension` (`src/uploads.ts:67-71`) runs the name-derived extension through `safeExt` / `SAFE_EXT_RE` (`src/uploads.ts:42,54-57`) and falls back to `bin`, and the stored name comes from `uploadFilename()` (`src/uploads.ts:89-91`), a fresh UUID. `.eps` with `type: ""` resolves to `eps` via `extensionFromName` (`src/uploads.ts:59-64`); `.eps` with `application/postscript` resolves the same way. `MAX_UPLOAD_BYTES` is untouched (`src/uploads.ts:8,169`). No new execution path — the result is only pasted into the PTY (`ui/src/lib/components/Viewport.svelte:1433`, `ui/src/lib/attachment-paste.ts`).
- **Locale parity.** Every key the diff introduces exists in both catalogs at matching positions: `files_upload_readonly_worktree`, `files_upload_readonly_attachments`, `files_upload_loading`, `files_upload_load_failed`, `files_upload_no_files`, `files_upload_unavailable`, `files_upload_disabled_title`, `feat_attach_any_file_title`, `feat_attach_any_file_body` — `ui/messages/en.json:2552-2560` vs `ui/messages/de.json:2552-2560`. The rename is one-sided nowhere: `viewport_attach_file` at `en.json:712` and `de.json:712`, with `viewport_attach_image` gone from both.
- **Announcement fragment.** `ui/src/lib/feature-announcements/entries/v1.47.0-attach-any-file.ts` matches the `FeatureAnnouncement` type at `ui/src/lib/feature-announcements.ts:25-36` and is picked up by the eager glob at `ui/src/lib/feature-announcements.ts:47-50` (no index to register in). `sinceVersion: "1.47.0"` agrees with the filename and with the other unreleased fragments in that directory (`v1.47.0-*`), and is above `package.json:3` (`1.46.0`), as the contract at `feature-announcements.ts:11-16` requires.
- **404 mapping is real.** `uploadScratchpadFile` targets `POST /api/sessions/:id/scratchpad/upload`, handled at `src/server.ts:2622-2647`, which returns 404 from three places, including `resolveScratchpadUploadDir` returning `null` on a blank `claudeSessionId` (`src/scratchpad.ts:238`). So the new `not_found` state at `FilesPanel.svelte:216-219` does map onto a genuine cause.
- **No drop-handler conflict.** `FilesPanel` is rendered at `ui/src/lib/components/Viewport.svelte:3052` as a sibling of `.term-mount` (whose `ondrop` is at `Viewport.svelte:2970`), inside `.vp-body`, which has no drag handlers — so `FilesPanel.handleDrop` not calling `stopPropagation()` is harmless.
- **Test identifiers resolve.** `ApiError(status, message)` matches `ui/src/lib/api.ts:162`; `clickWorktreeSegButton` is defined at `FilesPanel.browser.test.ts:174`; `describe` is imported at `ViewportTermControls.browser.test.ts:1`; `tmpRoot` and `makeDeps` exist at `test/server.test.ts:27,40`.

## Findings

### 1. `FilesPanel.svelte`: the new `error` branch also fires when a usable stale listing is present, so a drop is refused with a wrong reason while the Upload button still works

`browse()` (`ui/src/lib/components/viewport/FilesPanel.svelte:121-135`) assigns `listing` only on success and sets `error = true` in the catch. A failed browse therefore leaves the **previous** listing in place: `error === true` **and** `listing !== null`.

That state is reachable on a path that already executes — most directly through the post-upload refresh at `FilesPanel.svelte:225` (`void browse(listing?.path ?? "")`) failing, or by descending into a subdirectory the agent has since deleted.

In that state the two upload affordances now disagree:

- The Upload button at `FilesPanel.svelte:344-349` is keyed on `listing === null` only, so it stays **enabled** with no tooltip, and `uploadFiles()` uploads into `listing?.path` (`FilesPanel.svelte:196`) — which works.
- `handleDrop` hits the new `if (error)` branch at `FilesPanel.svelte:265-268` **before** the `!listing` check and refuses the drop with `files_upload_load_failed` — "The file list couldn't be loaded, so there is nowhere to put this. This session may have no scratchpad to upload into." (`ui/messages/en.json:2555`, `ui/messages/de.json:2555`).

So immediately after a successful upload whose refresh call failed, dropping a second file tells the operator this session may have no scratchpad — a statement the just-completed upload disproves, and one the still-enabled button would disprove again on the next click. This is the same class of misleading-refusal the PR set out to remove.

Scope the branch to the state it was added for: `if (error && !listing)`. (Alternatively, disable the button and the drop together on `error` — but then the tooltip/message wording needs to stop asserting "no scratchpad" for what is often a transient failure.)

## Nits (non-blocking):

- `ui/src/lib/components/viewport/FilesPanel.svelte:243` — `dragOver = !uploadDisabled` still shows the "Drop files to upload" overlay (`FilesPanel.svelte:319-325`) while the listing is loading or errored, inviting a gesture that `handleDrop` will then refuse; folding `listing === null || error` into the overlay condition would keep the affordance honest (this half is pre-existing behaviour, but the diff is what made the refusal explicit).
- `ui/src/lib/components/viewport/FilesPanel.svelte:369` — the notice uses `role="status"` on an element that is inserted at the same moment its text appears; `role="alert"` (or a permanently-mounted live region whose text changes) announces more reliably, and the `err` styling already frames it as a failure.
- `ui/src/lib/components/viewport/FilesPanel.svelte:216-219` — 404 from `src/server.ts:2630` also covers "unknown or archived session", for which `files_upload_unavailable` ("this session has no scratchpad to upload into") is a near-miss rather than the true cause.
- `ui/messages/en.json:2553` — "Attachments is read-only." reads awkwardly; "The Attachments folder is read-only." matches the `de.json:2553` phrasing better.
- `ui/src/lib/components/viewport/FilesPanel.svelte:117` — `dropNotice` is only cleared by a new dragover, an upload, a source switch, or a session change; it survives ordinary breadcrumb navigation, so an explanation can outlive the directory it referred to.
- `ui/src/lib/components/viewport/ViewportTermControls.svelte:712-724` — dropping `accept` also drops the iOS/Android photo-picker affordance the old `accept="image/*,video/*"` triggered; intended per plan D1, just worth knowing that attaching a screenshot on mobile is now one step longer.

- `ui/src/lib/components/viewport/FilesPanel.svelte:273-277` + `ui/messages/{en,de}.json:2556` — the `files_upload_no_files` case (a drop carrying a link rather than a file) is not in the plan's list of three blocked cases; it is in the spirit of section B and cheap, so keeping it is fine.
- `ui/src/lib/api.ts:258-259` (deleted) — removing the `uploadImage` alias is beyond the plan's file list, but it was the last consumer's alias and leaving it would be newly dead code.
- `test/uploads.test.ts:162-189` — the hostile-filename test is an addition the plan didn't call for; it pins the security property the widening leans on, so it earns its place.

- `ui/src/lib/components/viewport/FilesPanel.svelte` inherits, and this PR does not change, the fact that `uploadFiles()` reads `dirPath` from a possibly stale `listing` after a failed refresh — the finding above is about the *new* asymmetry, not this underlying behaviour.

## Limitations

- I did not run lint, type-check, or any test suite (CI owns those, per the review brief); identifier and catalog checks above were done by reading the tree.
- The plan's manual-browser success criteria (a real drag producing a real `drop` event with `dropEffect` intact) cannot be verified by reading code; the synthetic-`DataTransfer` test at `FilesPanel.browser.test.ts:358-392` records the assignment rather than reading it back, which is the right call given the spec, but it remains a proxy for the real-browser behaviour.

scoop (commented):
# Review — attach any file type (round 3)

## Prior finding

**1. `FilesPanel.svelte` — `if (error)` in `handleDrop` refused a drop that a still-enabled Upload
button would have accepted.** **Addressed.** The branch is now scoped to `error && !listing`
(`ui/src/lib/components/viewport/FilesPanel.svelte:269`), with a comment naming the exact reason
(`:265-268`). I checked the asymmetry the old code created: `browse()` only assigns `listing` inside
the `try` (`FilesPanel.svelte:125-128`) and sets `error = true` in the `catch` (`:130-131`), so a
failed post-upload refresh (`:225`) leaves `error === true` with `listing !== null`; the Upload
button is gated on `disabled={listing === null}` (`:353`), so it stays enabled in that state, and
the drop path now agrees with it — `error && !listing` falls through to `uploadFiles()` (`:282`),
which uploads into `listing?.path` (`:196`). The still-loading case keeps its own message at `:273`.
The disabled-button `title` was made state-aware too (`:348-352`): `files_upload_load_failed` when
`error`, `files_upload_disabled_title` otherwise, and `undefined` when a listing is present — so the
tooltip no longer contradicts an enabled button either.

## What I verified

**Symbol removals resolve to nothing live.** `sessionAttachExtForMime` / `SESSION_ATTACH_MIME_EXT`
(deleted from `src/uploads.ts`) and `uploadImage` (deleted from `ui/src/lib/api.ts:411` in the base)
have no remaining references anywhere in `src/`, `ui/src/`, or `test/` — the only hits are
`CHANGELOG.md:1947` and historical plan docs under `docs/superpowers/plans/`. `attachImages` is
likewise gone from live code; the prop rename is consistent across
`ui/src/lib/components/Viewport.svelte:3151`, the prop declaration at
`ui/src/lib/components/viewport/ViewportTermControls.svelte:16` and its type at `:27`, and the two
call sites at `:85` (button `onclick` path) and `:167` (input `onchange`).

**i18n parity.** All ten touched/added keys exist in both catalogs at matching positions:
`viewport_attach_file` at `ui/messages/en.json:712` and `ui/messages/de.json:712`;
`files_upload_readonly_worktree`, `files_upload_readonly_attachments`, `files_upload_loading`,
`files_upload_load_failed`, `files_upload_no_files`, `files_upload_unavailable`,
`files_upload_disabled_title`, `feat_attach_any_file_title`, `feat_attach_any_file_body` at
`ui/messages/en.json:2552-2560` and `ui/messages/de.json:2552-2560`. No `viewport_attach_image`
remains in either catalog. Every `m.*` call in the diff maps to one of those keys
(`FilesPanel.svelte:258,262,270,274,279,351,352,383`; `ViewportTermControls.svelte:83,85`).

**Server widening is contained.** `handleUpload` now derives the extension for both branches from
`uploadExtension(file)` (`src/uploads.ts:177`), which takes the name's extension when present and
falls back to `extForMime` otherwise, running both through `SAFE_EXT_RE` with a `bin` fallback
(`src/uploads.ts:59-71`). The stored name is still `${randomUUID()}.${ext}` (`:89-91`), the dest dir
is still `worktreeUploadsDir(s.worktreePath)` (`:84-86`, `:182`), the 404 for an unknown session is
preserved (`:181`), and `MAX_UPLOAD_BYTES` is checked before any of it (`:169-170`). So the raw
filename never reaches the filesystem and no traversal is reachable through the widened path. No
regression for the previously-supported types: a clipboard screenshot with an empty name still
resolves `image/png → png` via `extForMime` (`src/uploads.ts:24-33, 70`), and a `.MOV` still lands
as `mov` because `safeExt` lowercases (`:54-57`).

**Nothing else consumes the removed 415.** The only in-repo caller of `POST /api/uploads?session=`
is `uploadFile` in `ui/src/lib/api.ts:397-409` via `attachFiles`
(`ui/src/lib/components/Viewport.svelte:1428`), which has no type-specific branch. The clipboard
path still filters to images at `ui/src/lib/clipboard.ts:13`, matching plan decision D3, and
`attachmentPastePayload` still branches only on `video/` (`ui/src/lib/attachment-paste.ts:6`), so a
typeless `.eps` correctly pastes the bare path.

**Drop-zone nesting.** `handleDrop` does not `stopPropagation`, so I checked for an ancestor drop
handler that would double-handle the same file: the only other handlers in the component tree are
`ui/src/lib/components/Viewport.svelte:2963-2970` (on `.term-mount`, a *sibling* of the Files
tabpanel, not an ancestor) and `ui/src/lib/components/NewTask.svelte:1576-1583` (a different route).
No double-upload path.

**Announcement fragment.** `ui/src/lib/feature-announcements/entries/v1.47.0-attach-any-file.ts` is
picked up by the eager glob at `ui/src/lib/feature-announcements.ts:47-50` — no registry edit
needed. `sinceVersion: "1.47.0"` is above `package.json:3` (`1.46.0` = last release) as the contract
at `feature-announcements.ts:11-16` requires, the filename matches `v<sinceVersion>-<id>.ts`
(`:18-23`), and `id: "attach-any-file"` is unique across the entries directory. It shares 1.47.0
with twelve existing unreleased fragments, so the version is consistent with the current cycle.

**Test-file identifiers resolve.** `test/server.test.ts`'s new case uses `mkdirSync`
(imported at `test/server.test.ts:4`), `join` (`:13`) and `tmpRoot` (`:27,31`);
`FilesPanel.browser.test.ts`'s new block uses `ApiError` (`:5`), `m` (`:6`), `mockListing`/
`mockUpload` (`:24-25`), `EMPTY_LISTING` (`:28`), `waitForPanel` (`:67`) and
`clickWorktreeSegButton` (`:174`), all pre-existing in that file;
`ViewportTermControls.browser.test.ts` imports `describe` at `:1` and `tick` at `:4`.

## Limitations

- I did not run lint, type-check or any test suite — CI owns that, per the brief.
- The plan's success criteria 2–4 (native picker offering every type, real-browser drag behaviour,
  `dropEffect` surviving a genuine drag) are browser-runtime properties I cannot confirm from the
  source. The code sets `dropEffect = "copy"` unconditionally and never writes `"none"`
  (`FilesPanel.svelte:241`), which is what those criteria require, but the actual browser pass is
  outside what I can check here.

## Nits (non-blocking)

- `ui/messages/en.json:2557` — `files_upload_unavailable` asserts flatly "this session has no scratchpad to upload into", but `src/server.ts:2630` also returns 404 for an unknown or *archived* session, where the true cause is different; a hedge like the one already used in `files_upload_load_failed` (`en.json:2555`, "may have no scratchpad") would fit both.
- `ui/src/lib/components/viewport/FilesPanel.svelte:348-352` — a `title` on a `disabled` button is not surfaced consistently across browsers (Safari in particular suppresses the tooltip on disabled form controls), so the explanation may be invisible to exactly the user who clicked the dead button.
- `ui/src/lib/components/viewport/FilesPanel.svelte:372-374` — the `role="status"` element is created at the same moment its text appears; several screen readers only announce live regions that existed before the content changed. A permanently-rendered empty container that is filled on demand announces more reliably.
- `ui/src/lib/components/viewport/FilesPanel.svelte:242` — `dropNotice` is cleared only by a new dragover or a new upload, so a blocked-drop explanation lingers indefinitely if the user simply walks away from the drag; a timeout or a dismiss affordance would bound it.
- `ui/src/lib/components/viewport/FilesPanel.svelte:246-250` — `handleDragLeave` no longer early-returns on `uploadDisabled`, which is harmless (it only ever assigns `dragOver = false`, already false in that state) but the removed guard is now purely incidental rather than deliberate.

- `ui/src/lib/api.ts` — deleting the `uploadImage` alias is not in the plan's file table; it is dead after the rename and the repo runs a dead-code gate, so keeping it would likely have failed CI, but it is still an unlisted edit.
- `test/server.test.ts` — rewriting the end-to-end 415 case to seed a real worktree through the store is outside the plan's file table, though it is the direct consequence of the 415 removal.
- `test/uploads.test.ts` — the added "sanitizes a hostile attachment name" case goes beyond the planned `.eps` assertion; it is a genuinely useful guard on the widened path, not a defect.

Issue #1816, opened 2026-07-17
Codex rollout adapter: reviewer activity feed + token totals (telemetry leg of #1180)


kai-osthoff:
Part of the **telemetry** leg of epic #1180. Overlaps the tail of #1087 — see "Relationship to #1087".

## Problem

Shepherd's transcript readback is **claude-only**, but Codex is a first-class provider. Every Codex role spawn is therefore invisible: no live activity, no token totals.

User-visible symptom: the review status strip under the terminal shows **"Starting review…" for the entire review** and never progresses.

### Causal chain

1. `ReviewInFlightBanner.svelte:323` renders the `reviewbanner_preview_waiting` placeholder ("Starting review…") **exactly when the activity feed is empty** — it is the feed's *empty state*, not a loading state.
2. The feed is fed by `session:plangate-activity` (`src/index.ts:1407`), emitted from `src/plan-gate.ts:717` only when `readActivity` returns a summary.
3. `defaultReadActivity` (`src/plan-gate.ts:1246`) reads `~/.claude/projects/<dashified-worktree>/<reviewerSessionId>.jsonl` via `jsonlPathFor` (`src/usage.ts:24`).
4. **Codex never writes that file.** It writes `~/.codex/sessions/YYYY/MM/DD/rollout-*.jsonl`.

So `readActivity` returns `null` every tick, `onActivity` never fires, the feed stays empty.

### Corroborating measurement

`readUsage` uses the **same** `jsonlPathFor` route, so it fails identically — and measurably. Live DB (`reviewer_spawns`, completed rows):

| Reviewer provider | Completed runs | With token data |
| ----------------- | -------------- | --------------- |
| `claude`          | 97             | **97** (100 %)  |
| `codex`           | 130            | **0** (0 %)     |

## Scope

Build the Codex rollout adapter + a central provider-aware usage reader, and wire the **two reviewer consumers** (plan-gate + critic). Task-session activity (#1818), recap reads (#1819) and the remaining role-spawn usage sites (#1820) are separate children.

New module `src/codex-activity.ts`.

## A0 — Launch-side unique correlation (the foundation — build first)

The whole design rests on this. Why do reviewer cwds collide at all? The critic names its worktree after the **PR head SHA**, not the spawn. `standalone-critic.ts:327-328` says so itself:

[...]

Issue #1180, opened 2026-06-27
Epic: Codex provider parity — session-id foundation → Plan Gate → telemetry


scoop:
Tracking epic to sequence the Codex-provider-parity work and remove the overlap surfaced in the 2026-06-27 issue sweep.

## Why
Three open issues independently specified the **same** foundation — persisting a provider-native Codex session id — risking three colliding schemas/migrations. This epic makes the dependency explicit and orders the drain so the field is introduced **once**, provider-neutrally, before its consumers.

## Members & order

```epic-dag
#1175
#1087 <- #1175
#1160
#1816
#1817
#1818 <- #1816
#1819 <- #1816
#1820 <- #1816
```

- **#1175 — foundation (ship first).** Persist a provider-neutral `providerSessionId` (+ provider metadata) and switch Codex restore from `codex resume --last` to `codex resume <providerSessionId>`. Thin vertical slice: delivers reliable Codex bring-back on its own.
- **#1087 — generalized Plan Gate provider layer (gated on #1175).** Provider adapters/registry, capability gating, Codex Plan Gate + configurable reviewer provider. Reuses #1175's `providerSessionId`/provider metadata for `reviewer_spawns` instead of re-introducing it. (Resolve stale draft PR #1088 first.)
- **#1160 — Codex cache-token telemetry (independent leaf).** Narrowed to its one unique deliverable: does `src/usage.ts` capture Codex cached-token accounting? No dependency on #1175.

## Second wave (2026-07-17) — the telemetry leg

Wave 1 (#1175 ✅, #1160 ✅, #1087 open) delivered the session-id foundation and the Plan Gate provider
layer. **The third leg in this epic's title — telemetry — was never written up as a member.** Codex
reviewers now run (378 plan_gate spawns on a live instance), but Shepherd cannot read what they do:
its transcript readback is claude-only, so Codex role spawns produce **no live activity and no token
totals**.

Measured on a live DB (`reviewer_spawns`, completed rows): `claude` **97/97** rows carry token data,
`codex` **0/130**. The operator-visible symptom is the review status strip under the terminal, which
shows "Starting review…" for the entire review — that placeholder is the activity feed's *empty
state*, and the feed is fed from `~/.claude/projects/**.jsonl`, which Codex never writes (it writes
`~/.codex/sessions/**/rollout-*.jsonl`).

- **#1816 — Codex rollout adapter + reviewer activity/tokens (ship first).** The shared resolver
[...]

PR #2132, opened 2026-08-28
docs(research): evaluate @openai/codex-sdk for the Codex integration

kai-osthoff:
Research-only PR: one new file under `docs/research/`, no product code touched.

**Question:** would basing Shepherd's Codex integration on `@openai/codex-sdk` bring advantages?

**Answer: no — and the question points at the wrong layer.** Verified against the Codex CLI
**0.150.1** installed on this host and the published `@openai/codex-sdk@0.150.1` tarball (not from
docs):

- The SDK is 549 lines of shipped JS that `child_process.spawn`s `codex exec --experimental-json`
  and types the JSONL. It depends on `@openai/codex` pinned to the **exact** same version, so
  adopting it means a second `codex` binary in `node_modules` alongside the operator's on-PATH
  install that `src/codex-update.ts` exists to keep correct.
- Every SDK feature is a public `codex exec` flag Shepherd already has available and simply is not
  passing: `--json`, `--output-schema`, `--thread-source`, `--ephemeral`, `--add-dir`.
- For the herd's interactive sessions it is blocked anyway by the compliance stance in `PRD.md:48-52`
  / `README.md:73-81` — and it stamps `CODEX_INTERNAL_ORIGINATOR_OVERRIDE=codex_sdk_ts` on every
  request, so the distinction is not ambiguous to the provider.
- The tier that *would* change the architecture is `codex app-server` — a typed JSON-RPC daemon that
  ships its own contract (`codex app-server generate-ts` emitted 94 + 595 types here) with
  `ThreadStartResponse` (id at start), `ThreadStatus`, `ThreadTokenUsage`,
  `AccountRateLimitsUpdatedNotification` and `TurnSteerParams`. The SDK does not use it.

Two corrections folded in after a second pass over the upstream Rust sources: `turn.completed.usage`
is **cumulative since thread start**, not the turn's delta (the JSONL writer reads
`ThreadTokenUsage.total`, not `.last`), and the **5h/weekly rate-limit windows are absent from
`codex exec --json` entirely** — the request to add them (`openai/codex#14728`) was closed with "This
feature hasn't received any upvotes, so closing". So `src/codex-usage.ts`'s scraping has no
SDK/exec-JSON replacement at all; only `app-server` carries that signal.

The report also flags a live risk the evaluation turned up independently: `codex migrate-rollouts`
("migrate **legacy** local sessions to paginated thread history") means the rollout JSONL layout that
`codex-session-id.ts`, `codex-usage.ts` and `codex-activity.ts` all parse is now explicitly legacy.

Recommendation is four small flag-level changes plus one #1175-shaped spike on `app-server`; details
and citations in the doc. No issues filed yet — happy to open them on request.

Follow-ups filed from this report: #2133 (`--output-schema`), #2134 (role-isolation gap), #2135
(`app-server` go/no-go spike), #2136 (re-verify session-source + effort clamp). Findings that bear on
work already in flight were posted as comments on #1816 and #1180 rather than duplicated as issues.

### DIFF
--- a/docs/research/codex-sdk-evaluation.md
+++ b/docs/research/codex-sdk-evaluation.md
@@ -2,11 +2,12 @@
 
 **Verdict: NO — not as posed, and the question points at the wrong layer.** `@openai/codex-sdk` is a
 ~550-line wrapper that spawns `codex exec --experimental-json` as a child process and parses its
-JSONL into typed events. It adds **no capability Shepherd cannot get today by adding two flags to
-one argv builder**, while adding a second pinned `codex` binary inside `node_modules`, a new
-version-coupling constraint, and an `originator` marker on every request. Meanwhile the flags
-Shepherd is _not_ using (`--json`, `--output-schema`, `--thread-source`, `--ignore-user-config`,
-`--ephemeral`) would retire the most fragile code in the Codex integration **this week**, and the
+JSONL into typed events. It adds **no capability Shepherd cannot get from the CLI it already
+spawns**, while adding a second pinned `codex` binary inside `node_modules`, a new version-coupling
+constraint, and an `originator` marker on every request. Meanwhile the flags Shepherd is _not_ using
+(`--json`, `--output-schema`, `--thread-source`, `--ignore-user-config`, `--ephemeral`) would retire
+most of the fragile parsing in the Codex integration — one of them, `--json`, needs a stdout
+delivery path first (§3.2) — and the
 capability tier that would actually change Shepherd's architecture — `codex app-server`, a typed
 JSON-RPC daemon with first-class thread ids, steering, status and usage — is something the SDK
 **does not use at all**.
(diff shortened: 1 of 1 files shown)


kai-osthoff:
# Review

## Findings

1. `docs/research/codex-sdk-evaluation.md: Replace the -o fallback only after adding a replacement output transport and updating every role reader. The recommendation at lines 218-221 treats --output-schema as a replacement for the final-message file, but Shepherd's role consumers currently obtain a chat-only Codex result exclusively through `readRoleResultText(..., lastMessageFile)`: `src/critic-core.ts:899-910` uses the per-spawn -o file, and `src/recap.ts:176-183` does the same for recaps. `src/codex-last-message.ts:55-89` reads only the result file or that fallback path; it does not consume Codex stdout. Removing -o without a new capture/read path turns the documented chat-answer case into absent/no-result rather than fixing it.

2. `docs/research/codex-sdk-evaluation.md: Correct the role-isolation comparison before recommending a parity change. Lines 202-205 say Claude roles get --safe-mode, but it is conditional, not universal: `src/transient-agent-argv.ts:169-173` marks writer-ro and writer-only as `mcpIsolated: false`, and `src/transient-agent-argv.ts:244-258` emits --safe-mode only when that field is true. The repository's previous Codex isolation research records the qualified form explicitly at `docs/research/claude-code-codex-release-parity-2026-07-29.md:124-135`. The current wording overstates the baseline used to justify the proposed Codex flags.

## Verification notes

The diff adds only this research document. I verified its Shepherd-internal claims and recommendations against the cited source files; no builds or tests were run. The claimed current Codex CLI/SDK and app-server contracts are external artifacts and are not vendored in this repository, so I could not independently verify those live/version-specific assertions.

kai-osthoff (commented):
# Review

Decision: comment

No blocking findings.

The prior `-o` transport concern is resolved: the recommendation now explicitly retains `-o` until a stdout delivery path exists, and distinguishes schema validation from output delivery (`docs/research/codex-sdk-evaluation.md:244-247`). This agrees with the actual role argv, which only adds `-o` when a caller supplies a fallback filename (`src/codex-role-argv.ts:34-43`), and with the polling read contract, which reads the role result file before the `-o` fallback (`src/codex-last-message.ts:67-89`).

The prior claim that every Claude role uses `--safe-mode` is resolved: the document now limits it to the `mcpIsolated` reviewer and doc presets (`docs/research/codex-sdk-evaluation.md:227-230`). The preset table sets `mcpIsolated: true` only for those two kinds (`src/transient-agent-argv.ts:169-174`), and `--safe-mode` is conditionally appended from that field (`src/transient-agent-argv.ts:244-263`).

Could not verify the document's claims about the installed Codex CLI, npm artifact, or upstream app-server protocol from repository ground truth; those external dependencies are explicitly presented as host/upstream research sources (`docs/research/codex-sdk-evaluation.md:274-288`), so this is a stated verification limitation rather than a finding.

PR #2138, opened 2026-08-28
fix(commands): surface Claude Code's bundled skills and nested command dirs

kai-osthoff:
Typing `/design` in the New Task dialog matched `/codebase-design`, `/frontend-design` and
`/shopify-design` — but never `/design` itself, which Claude Code has offered for a while.

## Why it was missing

`listCommands()` builds the picker's list purely from the filesystem — `~/.claude/skills/`,
`~/.claude/commands/`, the plugin cache, `<repo>/.claude/`. Claude Code's **bundled** skills are not
on the filesystem: they live inside the `claude` binary. No scan can ever reach them; the
hand-maintained `BUILTINS` list is their only channel, and it held four entries.

Measured against what `claude` itself reports (`init.slash_commands` from a
`-p --output-format stream-json` run, same repo, same config):

| | before | after |
| --- | --- | --- |
| Shepherd lists | 129 | 139 |
| Claude Code offers | 174 | 174 |

Two separate causes, one commit each.

## 1 · Bundled skills (and two ghosts)

Added the bundled skills that work as an *initial prompt* for a spawned session — `design`,
`simplify`, `run`, `dataviz`, `update-config`, `fewer-permission-prompts`, `claude-api`,
`workflow-authoring`, `loop`, `schedule`. Purely-interactive and administrative commands
(`/clear`, `/config`, `/model`, `/usage`, `/doctor`, …) stay out, which is the criterion the
comment above `BUILTINS` already prescribed.

Removed `/review` and `/pr-comments`: **Claude Code 2.1.250 ships neither**. Picking one opened a
task whose first message was a command nothing resolves — the same staleness this list is prone to
in the other direction. The comment now records how to re-check it (diff `init.slash_commands`
against `listCommands()`), and why runtime discovery was rejected: `claude` only emits that list
during a real `-p` run (empty stdin exits before it), so reading it costs an API call, and
`handleCommands` is a *sync* handler on the one Bun event loop that also pumps the live web
terminal.

Also excluded deliberately: `verify`, `debug`, `batch`, `deep-research`, `design-sync`,
`run-skill-generator` — account-gated, and with no documented description a row for them would be
invented text.

## 2 · Nested `.claude/commands/` directories

`collect()` read only top-level `*.md`. Claude Code walks the tree: a subdirectory is a namespace
joined with `:`, and a subdirectory holding a `SKILL.md` is a skill named after the directory.

```
commands/references/handoff-template.md  →  /references:handoff-template
commands/ship/SKILL.md                   →  /ship
```

Both were dropped silently, so the picker disagreed with the CLI for anyone whose `commands/` has
any structure. `collectCommandDir()` now walks it, **bounded at three levels**: entries under
`~/.claude` are commonly symlinks into other trees and `statSync` follows them, so an unbounded
walk would spin that synchronous handler forever on a cycle.

## Verification

- `bun run lint`, `bun run typecheck`, `bun run test` (8805 pass) — all green.
- Six new/changed assertions in `test/commands.test.ts`: the builtin pin, the two ghosts' absence,
  colon namespacing, `SKILL.md`-names-the-directory (and no bogus `ship:SKILL` row), and the depth
  bound.
- Against the operator's real `~/.claude` **and** the CLI's own list: every added builtin present,
  described and invokable; `review`/`pr-comments` gone; `ship` and `references:handoff-template`
  found.
- The picker's actual ranking function (`filterCommands` in `ui/src/lib/slash.ts`) on the real
  payload — typing `/design` now returns `/design` first, ahead of the four substring matches from
  the bug report. I did not boot a second Shepherd server for a screenshot: that would have run a
  worktree build against the live instance's shared state.

One known, pre-existing and untouched discrepancy remains in the other direction:
`~/.claude/commands/createissue.md` is listed by Shepherd but not by Claude Code, for no pattern I
could find. Unrelated to this change.

### DIFF
--- a/src/commands.ts
+++ b/src/commands.ts
@@ -148,6 +148,74 @@ function safeReaddir(dir: string): string[] {
   }
 }
 
+/** How many directory levels below `commands/` the walk descends.
+ *
+ *  It has to descend at all because Claude Code does: a subdirectory is a namespace joined with
+ *  `:` (`commands/references/handoff-template.md` → `/references:handoff-template`), and a
+ *  subdirectory holding a `SKILL.md` is a skill named after the directory (`commands/ship/SKILL.md`
+ *  → `/ship`). A flat scan silently drops both.
+ *
+ *  It is BOUNDED because entries under `~/.claude` are commonly symlinks into other trees (see
+ *  {@link skillNameFrom}) and `statSync` follows them: a symlink cycle would spin this synchronous
+ *  walk forever inside `handleCommands`, on the one Bun event loop that also pumps the live web
+ *  terminal. Three levels covers every real layout. */
+const MAX_COMMAND_DEPTH = 3;
+
+/** `"dir"` / `"file"` for a `commands/` entry, or null when it is neither or unreadable. */
+function entryKind(path: string): "dir" | "file" | null {
+  try {
+    const stats = statSync(path);
+    return stats.isDirectory() ? "dir" : stats.isFile() ? "file" : null;
+  } catch {
+    return null; // vanished or unreadable → skip, don't fail the whole listing
+  }
+}
+
+/** The command name a `commands/` file contributes, or null when it contributes none. `SKILL.md`
+ *  is excluded because it names the *directory that holds it* — {@link collectCommandSubdir}
+ *  emits that row, and reading it here too would add a bogus `<dir>:SKILL` beside the real one. */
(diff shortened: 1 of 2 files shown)


kai-osthoff:
✅🎉 READY! 🥁 @scoop, brace yourself — actual work awaits. 😏🫡🔥🚀✨

PR #2139, opened 2026-08-28
fix: name lightweight mode, add an inline issue-retry link, and fall back between both gh transports

kai-osthoff:
The issue list reported "Couldn't load issues — GitHub may be rate-limited. Try again
shortly." for a repo whose problem was not GitHub, and offered no way to try again. Two
separate defects were hiding under that one sentence.

## 1. Lightweight mode was lost before the UI could see it

`makeForgeResolver()` returns a `LocalForge` for `repoMode: "lightweight"`. Its
`listIssues()` returns `[]` and its `slug` is `null` — **by design**, because the operator
turned the mode on themselves. But `handleIssues()` passed that through unlabelled, so both
the New Task modal and the Backlog panel fell through to the `slug === null` branch and
reported "no GitHub upstream" / "no git host configured for this repo". The operator goes
looking on GitHub for a state Shepherd is holding.

`/api/issues` now carries `lightweight`, and both panels name the mode and point at the
Automation panel where the toggle lives. The branch is placed **before** `slug === null` —
`LocalForge` reports a null slug too, so the old branch would otherwise swallow it. The flag
is reported on the failure branch as well: the mode is a property of the repo, not of the
request that just failed.

## 2. The REST fallback had no way back

`gh issue list` (GraphQL bucket) and `gh api` (REST bucket) draw on two independent GitHub
budgets, so either can be exhausted while the other is healthy. `gh api rate_limit` does not
settle it either — that endpoint is itself limit-exempt.

Observed on this repo while working on it:

| Call | Result |
| --- | --- |
| `gh api rate_limit` | core 5000/5000 free, `used: 0` |
| `gh api repos/<slug>/issues` | **HTTP 403 "API rate limit exceeded for user ID"** |
| `gh issue list --repo <slug>` | returns issues |
| `gh api graphql -f query='{viewer{login}}'` | returns the login |

`listIssues()` only fell back GraphQL→REST, and only on `isRateLimitError`. With a GraphQL
backoff engaged it went straight to REST, REST 403'd, and the error propagated — the
working transport was never tried. It now retries on the other transport in **either**
direction; both failing rethrows the preferred path's error.

Two widenings are deliberate and commented at the call site: any error retries (not just
rate-limit ones), and the REST→CLI direction knowingly issues a GraphQL call inside an
active backoff window — REST has just proved unusable, and a real GraphQL rate-limit error
re-extends the window by itself.

## 3. "Try again shortly." is now a link

That sentence was an instruction with no control attached. It is gone from the message; a
Retry link (reusing `common_retry`) sits beside it and re-fetches in place, without closing
the modal. `UpNextPanel` shares the message key and loses only that trailing sentence — it
already offers the action through the ⟳ in its header.

## Scope

Deliberately **not** included, to keep the diff tight:

- `UpNextPanel` gets neither the retry link nor the lightweight notice.
- Other one-way forge fallbacks (`getIssue`, `listPullRequests`, `listBacklogCounts`,
  `prStatus`, `prReviewMeta`) are untouched.
- No typed failure reason beyond `error: "fetch_failed"`, and no automatic retry.
- **Known, filed as #2140:** `currentUser()` uses REST only and caches a failure as `null`
  for the process lifetime (forge instances are permanently memoised). A REST limit at the
  wrong moment therefore silences the "mine & unassigned" filter until a server restart,
  even though `gh api graphql '{viewer{login}}'` would answer.

## Verification

- Root: `bun run lint`, `bun run test` — 8805 pass, 0 fail.
- `ui/`: `bun run check` (0 errors), `bun run check:i18n` (2 locales in parity),
  `bun run test` — 4686 pass.
- 14 new tests. The retry test in `IssuesPanel` was mutation-checked: removing the
  `loading = true` line makes it fail with "no open issues" instead of the failure banner.
- **Measured against real `gh`, not only fakes.** A throwaway probe drove the new code
  through the live CLI: with the GraphQL backoff engaged and REST injected-403 (the exact
  shape observed above), it fell back to a real `gh issue list` and returned 74 issues —
  the path that previously threw. Caveat, stated plainly: the naturally occurring 403 had
  cleared by the time the fix was ready, so the REST failure in that probe was injected
  rather than organic; the other two legs (REST-preferred and CLI-preferred) used real
  unmodified `gh` calls end to end.


kai-osthoff (commented):
# Review

## Summary

No blocking findings in the requested diff.

## Verified

- `GithubForge.listIssues()` selects the REST or CLI path from the existing GraphQL-backoff state, attempts the alternate path after every preferred-path failure, and rethrows the preferred-path error only if both fail (src/forge/github.ts:828-840). The CLI runner records GraphQL rate-limit failures in that same backoff state (src/forge/github.ts:301-316), while the REST listing uses the separate `gh api` request path (src/forge/github.ts:605-624).
- The API supplies `lightweight: false` when no forge resolves and derives the flag from `GitForge.isLightweight` for both successful and failed forge responses (src/server.ts:5915-5949). `GitForge` defines that optional capability and `LocalForge` is its concrete true-valued implementation (src/forge/types.ts:460-466; src/forge/local.ts:320-339).
- Both issue UI paths consume the flag with an explicit `=== true` check and render the lightweight state before their null-slug missing-host branches (ui/src/lib/components/new-task/issue-data.svelte.ts:25-66; ui/src/lib/components/PromptSources.svelte:477-496; ui/src/lib/components/IssuesPanel.svelte:220-257, 521-540).
- Retry controls are keyboard-focusable buttons styled as text links and invoke the existing in-place loaders: `IssueData.load(repoPath)` for New Task and ticket-guarded `softRefresh(repoPath)` for the backlog (ui/src/lib/components/PromptSources.svelte:479-487, 670-681; ui/src/lib/components/IssuesPanel.svelte:295-367, 524-530, 682-693).
- The only maintained UI catalogs are `ui/messages/en.json` and `ui/messages/de.json`; both contain the new lightweight key and the shared retry/load-failure keys (ui/messages/en.json:6,38-39; ui/messages/de.json:6,38-39).

## Verification limitation

Per review instructions, I did not run builds, tests, or other executable checks.

kai-osthoff:
🚨✨ Status: **READY** 🎉🟢 @scoop 👀 any decade now? ⏳🐌💤 Merge button awaits! 🙏🔥🚀

PR #135, opened 2026-06-01
feat(push): solid bell push icon + monochrome badge

scoop:
## What

Replaces the placeholder shepherd's-crook push notification icon with a **solid amber bell** (FA-style), and splits the **badge** from the icon.

| | Before | After |
|---|---|---|
| App icon | stroked crook (reads like a "P" when small) | solid bell, legible at notification sizes |
| Badge | reused full-color `icon-192.png` | dedicated white-on-transparent `badge-96.png` |

Android renders the notification `badge` as a flat monochrome mask — a colored/stroked icon degrades to a blob. A purpose-built solid white silhouette fixes that.

## Why not just reuse a Font Awesome bell?

- No Font Awesome in the repo — reusing would mean adding a dependency.
- `showNotification({ icon, badge })` needs a **URL to a raster image**; the OS draws the banner outside the page, with no access to fonts/CSS, so a font glyph can't be handed to it regardless — you always bake a PNG.
- So: took the FA bell *silhouette* as reference, authored a brand-styled SVG, regenerated via the existing `gen:icons` pipeline.

## Changes

- `ui/scripts/gen-icons.mjs` — bell glyph; emits new `badge-96.png` + `badge.svg`
- `ui/static/sw.js` — `badge:` → `/icons/badge-96.png`
- Regenerated `icon-192/512.png`, `apple-touch-icon.png`, `icon.svg`

## Verification

- `cd ui && bun run check` — 0 errors
- No i18n impact (no user-facing strings touched)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2144, opened 2026-08-30
fix(pwa): sheep icon set with a legal maskable variant

scoop:
Started from a screenshot of the Android notification shade showing a **shepherd's crook** where the app icon should be. That turned up two unrelated things: one that only looks like a regression, and one that is a genuine live bug.

## The crook is not a regression

It was the placeholder from `841bd9e3` (#43), replaced by the bell in `c3f34029` (#135). Repo and live server both serve the bell today.

It survives on the device because of Chrome's **app-identity gate**, not caching. Android's `WebApkUpdateManager` hashes icon *bytes* on every update check, so the swap was detected — but a change with a ≥11% pixel diff is only applied when `kPwaUpdateDialogForIcon` is enabled, and that is `FEATURE_DISABLED_BY_DEFAULT`. When the gate blocks, `setUseOldIcon(true)` rebuilds the WebAPK with the old icon. The stale splash is the same gate: `SPLASH_ICON_HASH_DIFFERS` is only recorded *inside* the `shouldUpdateIcon` block.

**No code change can reach an already-installed PWA.** Hence the manual step below.

## `icon-512.png` was not a legal maskable icon — this part is live

| file | alpha min | |
| --- | --- | --- |
| `icon-512.png` (bell, before) | 0 | transparent corners |
| `icon-512.png` @ `841bd9e3` (crook) | 0 | same defect |
| `icon-maskable-512.png` (this PR) | 255 | full bleed |

The generator drew the panel as `<rect rx="96">` with nothing behind it, so the corners rasterised transparent — and the manifest declared that exact file `purpose:"maskable"`. A WebAPK's adaptive icon has a hardcoded white background layer with our art as *foreground* only, so the transparency composites onto white and reads as a ring. **Independent of the crook** — it would look identical once the bell landed.

## What changed

| output | purpose | art |
| --- | --- | --- |
| `icon-{192,512}.png` | `any` | sheep on the rounded panel; transparent corners are legal here and correct unmasked |
| `icon-maskable-{192,512}.png` | `maskable` | sheep on a square, fully opaque, full-bleed ground |
| `apple-touch-icon.png` | iOS | full-bleed; iOS applies its own squircle, so self-rounding double-rounds |
| `badge-96.png` | Android badge | simplified white-on-transparent sheep silhouette, centred |

The service worker's notification `icon:` points at the **maskable** 192, not the `any` one — whatever crop the shade applies, a full-bleed source cannot expose white.

The badge is purpose-built rather than the full mark: Android flattens it to one colour, so the sheep would blob (what #135's message warned about). Wool opaque, ears and face knocked out, eyes punched back in at `r=3.1` — at `r=2.1` they merge into the knockout at 24px.

`gen-icons.mjs` now owns the geometry and emits `favicon.svg` too, so there is one definition rather than four hand-copies — that drift is why the PWA was the last surface on a placeholder. The emitted favicon renders at a **0-pixel diff** against the file it replaces.

Moved under `/icons/v2/` because desktop Chrome since 144 treats manifest icon URLs as immutable. `ICONS_VERSION` sits in its own module: `gen-icons.mjs` writes files at import time, so importing the constant from it made the test regenerate the assets it was checking (caught in self-review, fixed).

## Tests

`ui/scripts/gen-icons.test.ts`, 17 cases asserting *properties* of the committed PNGs — not byte-equality against a fresh `sharp` render, since rasterisation is not stable across resvg versions and that guard would be flaky. Maskable opacity + safe zone (verified non-vacuous: the `any` art fails the opacity probe, the maskable art passes), purpose separation, declared-vs-actual sizes, cross-file path integrity over all four referencing files, and the badge monochrome/centring contract.

Verified: `ui` check 0 errors + 4706 tests, root lint + 8806 tests, fallow clean on 22 changed files.

## Reviewer notes

- The icon ground is `#0f1413`, not the manifest's `background_color` `#0b0f0d`. The face (`#0a0d0c`) hangs below the wool; against `#0b0f0d` that chin vanishes and the silhouette changes. The splash therefore shows a barely-perceptible tone step between ground and icon — left alone deliberately rather than widening scope to the manifest colours.
- A service worker that has not yet updated references the removed paths and shows an icon-less notification until it updates. Accepted when the clean rename was chosen over compatibility copies.
- I can show the maskable contract was violated and that the new art masks cleanly. I could **not** establish from source which path Android uses for the shade group avatar specifically — the crescents sit at the four edge midpoints, which looks more like a square inset into a circular chip than the adaptive path. Full-bleed is correct either way, but "no white ring in the shade" needs the device check below.

```shepherd:manual-steps
- [ ] POST-MERGE: After deploy, uninstall and reinstall the Shepherd PWA on each device. Chrome's app-identity gate keeps the old icon on an existing install, so a reinstall is the only way to pick up the new one.
- [ ] POST-MERGE: Confirm on the device that the home-screen icon, notification-shade avatar and splash screen all show the sheep with no white ring.
```


scoop (commented):
# Review — `fix(pwa): sheep icon set with a legal maskable variant`

Diff reviewed: `51898691...HEAD` (single commit `3a324102`).

**Decision: comment.** I found no correctness, security, or contract defect. Details of what I
checked and against what ground truth are below.

## What the change does, verified

**1. `any` vs `maskable` are now genuinely distinct files.**
`ui/static/manifest.webmanifest:11-24` declares four entries: `/icons/v2/icon-{192,512}.png` as
`purpose:"any"` and `/icons/v2/icon-maskable-{192,512}.png` as `purpose:"maskable"`. Before the
change (`git show 51898691:ui/static/manifest.webmanifest`, lines 11-13) the *same* `icon-512.png`
carried both purposes. That was the reported defect and it is fixed.

**2. The maskable art really is full-bleed.**
`ui/scripts/gen-icons.mjs:82-83` builds the maskable SVG as `<rect width="64" height="64"
fill="#0f1413"/>` with no `rx` — compare `ui/scripts/gen-icons.mjs:70` where the `any` variant keeps
`rx="12"`. The committed outputs agree: I opened `ui/static/icons/v2/icon-maskable-192.png` and
`ui/static/icons/v2/icon-192.png` side by side — the maskable one paints dark to all four edges, the
`any` one shows the rounded corners. `ui/static/icons/v2/apple-touch-icon.png` is likewise full-bleed
(generated from `iconMaskable` at `ui/scripts/gen-icons.mjs:127`), which is right for iOS since it
applies its own squircle.

**3. The safe-zone claim in the comment is arithmetically true.**
`ui/scripts/gen-icons.mjs:80` claims the furthest glyph point is 184/512 px against a 204.8 px
budget. Working it from the geometry at `ui/scripts/gen-icons.mjs:36-52`, in the 64-unit viewport:

| element | furthest point from (32,32) | at 512px |
| --- | --- | --- |
| crown `circle cx=32 cy=19 r=10` | 13 + 10 = 23 | **184** |
| `circle cx=43 cy=24 r=9` | hypot(11,8) + 9 = 22.6 | 181 |
| `circle cx=22 cy=24 r=9` | hypot(10,8) + 9 = 21.8 | 174 |
| face `ellipse cx=32 cy=41 ry=12` | 9 + 12 = 21 | 168 |
| ear `ellipse cx=20 cy=38 ry=6` | hypot(12,6) + 6 = 19.4 | 155 |

Max 184 ≤ 0.4 × 512 = 204.8, which is exactly the bound asserted at
`ui/scripts/gen-icons.test.ts:83`. The same ratio holds at 192 (69 ≤ 76.8).

**4. The adaptive-icon reasoning at `ui/scripts/gen-icons.mjs:74-77` holds up.**
Chrome pads a maskable icon by ≈1.309 onto the WebAPK foreground layer, so the opaque art covers
1/1.309 ≈ 76.4% of the 108dp layer (≈82.5dp) against a 72dp visible viewport — the art fully covers
the mask, so nothing of the hardcoded white background layer can show. Making the source opaque is
therefore both necessary and sufficient for that path.

**5. Cross-file path integrity — all four sites moved together, and nothing stale remains.**
- `ui/static/manifest.webmanifest:11,12,14,20` → `/icons/v2/…`
- `ui/static/sw.js:41-42` → `/icons/v2/icon-maskable-192.png`, `/icons/v2/badge-96.png`
- `ui/src/app.html:7` → `/icons/v2/apple-touch-icon.png`
- `ui/scripts/icons-version.mjs:10` → `ICONS_VERSION = "v2"`

I grepped the whole tree for `icons/icon-`, `badge-96`, `apple-touch-icon`, `icons/badge`,
`icons/icon.svg` and for `icons/` in ts/js/mjs/svelte/html/json/webmanifest/conf/yml/md. The only
remaining hits are `extension/manifest.config.ts:51-54` and
`docs/research/extension-chrome-web-store-readiness.md:118`, which refer to the extension's own
`public/icons/icon-{16,32,48,128}.png` — a different asset set with no coupling to `ui/static`
(`extension/manifest.config.ts:21-24` says those PNGs are checked in, not built from `ui/`). No
reference to a removed file survives.

**6. The badge geometry and its own tests agree.**
From the mask at `ui/scripts/gen-icons.mjs:97-111`, the opaque region is the wool union minus the
ear/face knockouts plus the eyes: x ∈ [11, 53] (centre 32), y ∈ [9, ~43] (centre ~26 — the wool
bottom near x=32 is inside the knocked-out face ellipse, and near x=20/44 it is inside the ears, so
the effective bottom sits just under 43). The transform at `ui/scripts/gen-icons.mjs:109` maps
(32, 26) to the canvas centre, so the rendered centre lands within ~0.1px of centre at 96px —
inside the ±2% asserted at `ui/scripts/gen-icons.test.ts:165-166`. Glyph width 42 × 1.3146 ≈ 55.2 of
64 ≈ 86%, above the 75% asserted at `ui/scripts/gen-icons.test.ts:167`. Transformed extents are
x ∈ [4.4, 59.6], y ∈ [9.7, 55.0], so nothing is clipped by the 64-unit viewport.
The `<mask>` itself is well-formed: default `maskUnits="objectBoundingBox"` expands the masked
rect's bbox by 10% (−6.4 … 70.4), which contains all mask content, and default
`maskContentUnits="userSpaceOnUse"` puts that content in the transformed group's space, so it scales
with the rect. White = opaque, black = knocked out, per SVG luminance masking.

**7. The new test file will actually run.**
`ui/vite.config.ts:84-92` defines the `node` project with vitest's default `include` (only
`**/*.browser.test.ts` excluded), so `ui/scripts/gen-icons.test.ts` is picked up — same as the
existing `ui/scripts/gen-docs-manifest.test.ts` precedent. `sharp` is a devDependency
(`ui/package.json:35`), so the test's `import sharp` resolves.
Its regexes match the real files: `ui/static/sw.js:41-42` uses double-quoted `"/icons/…"` (matched by
`/"\/icons\/[^"]+"/g` at `ui/scripts/gen-icons.test.ts:117`) and `ui/src/app.html:7` uses a
double-quoted `href` (matched at line 118). `referenced.length` is 2 + 1 + 4 = 7 against the
`manifest.icons.length + 3` = 7 floor at line 125.

**8. `favicon.svg` is art-identical.**
The diff of `ui/static/favicon.svg` removes only the five `<!-- … -->` comment lines; every element,
attribute and colour is untouched. I traced the generator template (`ui/scripts/gen-icons.mjs:60-65`
plus `sheep()` at 44-52) and it reproduces the file's element order and indentation exactly,
including the trailing newline — so the committed file is what `bun run gen:icons` emits.

**9. The `icon:` swap in the service worker is sound.**
`ui/static/sw.js:41` now points the notification icon at the maskable (full-bleed) variant. Android
circle-crops the notification large icon; with a full-bleed source no platform ground can be exposed,
and the glyph sits inside the inscribed circle comfortably. Nothing else in `sw.js` changed.

### Stated limitations

- I could not decode the committed PNGs pixel-by-pixel: the Bash call to run a read-only PNG decoder
  was denied by the permission mode. Opacity/full-bleed is therefore verified from the SVG sources
  in the diff plus visual inspection of the rendered PNGs, not from a per-pixel alpha probe. The
  "badge is pure white on transparent" property (`ui/scripts/gen-icons.test.ts:139-145`) I could not
  verify at all by inspection — it is asserted by the test, which CI runs.
- I did not run `bun run check`, `bun run test` or lint (out of scope for this review, and CI covers
  them).
- The plan's device-level claims (Chrome's ≥11% app-identity gate, the required uninstall/reinstall)
  are about Chromium behaviour that is not in this repo. I neither confirmed nor disputed them; they
  do not affect the correctness of the code in the diff.

- `ui/scripts/gen-icons.mjs:113` — the generator now also writes `ui/static/favicon.svg`, a file the
  task did not ask to change. The plan declared this and flagged it as droppable; the art is
  byte-comparable apart from removed comments, so the risk is nil, but it is an extra surface.
- `ui/static/icons/v2/icon.svg`, `icon-maskable.svg`, `badge.svg` are committed intermediate sources
  that nothing references and no test covers (only the PNGs are checked).

- `ui/scripts/gen-icons.test.ts:56` filters with `i.purpose === "maskable"`; the manifest spec allows
  a space-separated list (`"any maskable"`), which would make the maskable contract test silently
  select zero icons. Dormant because `ui/static/manifest.webmanifest:11-24` uses single-valued
  `purpose`. A `.split(" ").includes("maskable")` would keep the guard honest through a future edit.
- `ui/scripts/gen-icons.mjs:22-23` only ever creates the new version directory; it never prunes the
  previous one. Bumping `ICONS_VERSION` (`ui/scripts/icons-version.mjs:10`) will leave `v2/` checked
  in and served indefinitely — dormant until the first bump.

## Nits (non-blocking):

- `ui/scripts/gen-icons.mjs:3-6` — "This script now owns the geometry … so there is one definition to
  change" overstates it: `site/public/favicon.svg:1-12` still carries a hand copy of the same sheep
  (as does `site/public/og-image.svg` and the checked-in extension PNGs). The plan puts those out of
  scope, so the code is right; the sentence just claims more than the diff delivers.
- `ui/scripts/gen-icons.mjs:28` — the icon ground `#0f1413` differs from the manifest's
  `background_color` `#0b0f0d` (`ui/static/manifest.webmanifest:7`). On the PWA splash, the `any`
  icon's rounded transparent corners will show that background, so the panel reads as a faintly
  lighter rounded square. Barely perceptible (Δ ≈ 4/255 per channel) and deliberate per the comment,
  but worth knowing given the report mentioned the loading screen.
- `ui/scripts/gen-icons.test.ts:14` — `@ts-expect-error — plain .mjs module, no types` is inert only
  because `svelte-check --tsconfig` restricts checking to the tsconfig's `include`, which does not
  cover `scripts/`. `ui/tsconfig.json:5-6` sets `allowJs` + `checkJs`, so if `scripts/` ever enters
  the program the directive itself becomes a TS2578 "unused `@ts-expect-error`" error. The root
  package imports a `.mjs` from TypeScript with no directive at all
  (`src/socket-pty-bridge.ts:3`).
- `ui/scripts/gen-icons.mjs:96` — `56 / 42.6`: the actual wool bbox width is 53 − 11 = 42, so the
  denominator is a hand-measured constant that does not follow from the geometry directly above it.
  Harmless (the result satisfies the committed assertions) but it will silently drift if the mark
  changes.
- `ui/scripts/gen-icons.mjs:44` — `sheep` is a nullary arrow function returning a constant template
  string; a plain `const` reads the same and avoids implying parameters.
- `ui/static/icons/v2/badge.svg:5-10` — the spliced `WOOL_CIRCLES` lands under-indented relative to
  its parent `<g>` because the same constant is interpolated at two nesting depths. Cosmetic only.

- `site/public/favicon.svg` still hand-duplicates the sheep geometry rather than sourcing it from the generator.
- `site/public/og-image.svg` carries another hand copy of the same mark.
- `extension/public/icons/icon-{16,32,48,128}.png` are checked-in renders of the mark with no build coupling to `ui/` (`extension/manifest.config.ts:21-24`).
- `ui/src/app.html:11` sets `theme-color` `#0a0d0c` while `ui/static/manifest.webmanifest:7` sets `theme_color` `#0b0f0d`; the two have disagreed since before this change.

PR #2150, opened 2026-09-01
fix(pricing): use Fable 5.1's cache-read price, pin claude-fable-5-1, guard the Fable family

kai-osthoff:
The installed Claude CLI resolves `fable` to **Fable 5.1** (`aliases.fable.default = "claude-fable-5-1"`, `latest_per_family.fable = "claude-fable-5-1"` in the baked model catalog of `2.1.257`). Shepherd already spawns 5.1 through the floating alias without any change — but two things around it were wrong or missing.

## 1. Cache reads were priced 4x too high (the actual bug)

`src/pricing.ts` had one `/fable/i` row with `cacheRead: 1`. That is Fable 5's tier (`tier_10_50`); Fable 5.1 is `tier_10_50_cache_read_0_25` — **$0.25** per Mtok. In/out ($10/$50) and both cache-write rates are identical between the two, so cache read was the only wrong number — and it is the dominant token class in long agent sessions (`docs/token-usage-analysis.md` records 16.8M of 17.4M tokens = 96% cache read in a single task). `dollars()` renders these weights as real currency in `/usage`, so Fable 5.1 sessions were overstated by up to ~4x — the exact failure the file header warns about.

Fixed by splitting the tier into two rows:

| match (order) | cacheRead | covers |
| --- | --- | --- |
| `/fable-5$/i` (new) | `1.0` | `claude-fable-5` and its provider-prefixed forms — historical records keep costing what they cost |
| `/fable/i` (was `1.0`) | `0.25` | `claude-fable-5-1` and the floating `fable` alias, which by definition means the latest Fable |

The `$` anchor is deliberate: `claude-fable-5-1` must not fall onto the retired row, and an unknown future id lands on the current row rather than on retired prices. Both rows stay after haiku so `DEFAULT = TABLE[1]` is still sonnet-like.

## 2. Pinned `claude-fable-5-1` in the picker

Opus has a pinned entry, Fable did not — so a future Fable release would silently take over the alias. Added next to `fable` with the same premium tier/fit, its own label ("Fable 5.1") and guidance detail; the floating row now reads "Fable (latest)" so the two rows are distinguishable, mirroring the existing Opus split. Mirrored into both `CLAUDE_MODELS` lists (mirror gate green) plus `PREMIUM_MODELS` and the alias list in `docs/external-task-api.md`.

## 3. Fable availability guard now keys off the family

[...]

PR #2153, opened 2026-09-01
feat(usage): delivery-metrics lens (#2151 R1)

### LINKED ISSUE #2151: Adopt the AI-native SDLC playbook: delivery metrics, review policy, plan drift, evals, maintain loop, intake
Implementation tracker for the AI-native SDLC playbook gap analysis (report: `docs/research/ai-native-sdlc-playbook.md`, PR #2149). Same pattern as #1951 for the Cursor-economics report.

Six ranked recommendations from the report — each independently shippable; promote to an epic if more than R1+R3 get greenlit:

- **R1 — Delivery-metrics lens** (highest leverage, do first): first-pass rate, rework cycles, plan rework, time-to-first-review, lead time — sourced from the append-only `reviewer_spawns` log (`kind = review|plan_gate` per taskSessionId), `local_prs`, `signals.kind`. NOT from `reviews.addressRound`/`streakReviews`/`plan_gates.round` (live streak state, reset on clean/resume — see report §3 R1 caveat). First-push CI green rate needs new per-push instrumentation; defer or include, decide at planning. Surface: Usage-style lens + `scripts/` report.
- **R2 — Version-controlled review policy**: (a) per-repo `REVIEW.md` appended to the critic prompt; (b) inject house rules into the reviewer spawn (today task-agents-only, `src/service.ts:2844`); (c) machine-readable Important/Nit severity + named passes in the verdict JSON, nit cap. Three separable slices.
- **R3 — Plan-drift report**: non-blocking `planDrift: none|minor|major` field from the critic (plan stays context-not-warrant); persist on `reviews`, feed R1.
- **R4 — Generalize the eval harness**: re-enable the stop-classifier eval schedule; fixture-based evals for plan-gate/critic/rundown prompts gating PRs that touch those builders; incident→fixture capture. (Complements the observational learnings lifecycle; no mandatory pre-ship evals per learning.)
- **R5 — Maintain loop that opens work**: tiered thresholds on Shepherd's own health signals — log → read-only diagnosis spawn drafting a backlog issue → pre-approved-class PR. Reuses reviewer-preset argv sandbox. Internal signals only, no production ingress.
- **R6 — Optional structured intake**: "shape this" clarifying round on New Task (reuse epic-shaping interview machinery) + readiness-analyzer-prescribed issue templates. Freeform fast path stays.

Suggested first slice: **R1** (substrate for R3/R5, zero behavior change) — then R3 as a cheap rider.

Out of scope (report §4): committed intent/spec files, separate spec stage, MDM governance, per-learning eval gates, external production control bands.

scoop:
Implements **R1** of #2151 — the delivery-metrics lens from the AI-native SDLC playbook gap analysis (`docs/research/ai-native-sdlc-playbook.md`).

Shepherd measures token cost exhaustively and delivery outcomes not at all. This adds the missing half: first-pass rate, rework cycles, plan rework, time-to-first-review, lead time and repeat incidents — as a **Delivery** tab in the Usage modal (`GET /api/usage/delivery`) and `bun run delivery-report`.

Scope was agreed at the plan gate as **R1 only**; R2–R6 and first-push CI-green are filed as follow-ups.

## Why this needed instrumentation, not just aggregation

The report's §3 R1 source table assumes PR-opened and merged timestamps are derivable from existing tables. They are not, for GitHub repos:

- `store.archive()` **deletes** the session's `session_git_cache` row in the same transaction that archives the session — no PR timestamp survives.
- `local_prs.createdAt/mergedAt` exist only for lightweight (LocalForge) repos.
- `reviewer_spawns` carries no `repoPath`; the `sessions` join it needs is pruned at 30 days while spawns live 90.
- A spawn row is written *before* the critic runs (deliberate, for cost attribution), so a raw spawn tally counts crashed critic runs as rework rounds.

So two small additive, archive-decoupled facts:

- **`delivery_facts`** — one row per session with `prOpenedAt` / `mergedAt` plus denormalized `repoPath` / `desig` / `createdAt`. Modelled on `post_merge_steps`: no FK, deliberately excluded from the `pruneArchivedSessions` cascade, own retention sweep pinned to the `reviewer_spawns` window (they are read together).
- **`reviewer_spawns.outcome`** — `clean` / `changes_requested` / `error` for reviews, `approved` / `rework` / `error` for plan gates, stamped once at finalize. NULL means unknown and is excluded from every count. Added to **both** the fresh-install DDL and the migration (the #1816 invariant).

## The merge-train gap

`mergedAt` is stamped inside `store.archive(id, 'merged')`, **not** off the `session:git` event. `AutoMergeService.doMerge` calls `settleMergedSession`, which archives the session and drops it from the PR poller — no git event ever fires for that path. Sourcing the timestamp from the event alone would have silently excluded every autonomous merge from merged tasks, lead time, first-pass rate, rework cycles and plan rework. `settleMergedSession` (shared by the train, `Drain.reapMerged` and the manual merge route) is the only writer of `archive(id, 'merged')`, so the store call is the one chokepoint no merge path can bypass. The `session:git` subscriber keeps `prOpenedAt` and acts as a first-write-wins secondary for out-of-band merges the poller sees first.

This was caught by the plan reviewer, verified against the code, and is covered by a test that stamps a merge with no git event fired at all.

## Measurement fidelity

- `null`, never `0`, where a metric has no qualifying sample — the lens renders `—`. An unmeasured window must not read as a measured one.
- Medians as the headline (a single stuck task would dominate a small window), each shown with its sample size `n`.
- Merged tasks with no verdict-bearing review are excluded from the first-pass denominator and reported separately as `unreviewed`.
- Inverted durations (clock skew, a late-observed PR-open) are dropped, not clamped to 0 — clamping would flatter the metric.
- Instrumentation is **forward-only**. The payload carries `measuringSince` so a sparse window reads as young instrumentation, not a delivery drought. No backfill is attempted.

## Notes

- `LocalForge.prStatus()`'s `merged` branch omitted `createdAt` (only the `open` branch returned it), so a pseudo-PR first observed already-merged never got a `prOpenedAt`. One-line fix.
- The new `session:git` subscriber is write-free in the steady state (per-session stamp map + early return) — it rides a per-poll event on the loop that also pumps the web terminal.
- Aggregation is a pure builder (`buildDeliveryMetrics`) over an injected store, mirroring `buildUsageBreakdown`; the route only validates `range`.

## Verification

Root `lint` + `tsc` + `test` (8868 pass), `ui` `check` + `test` (4723 pass) + `check:i18n`, glossary, feature-catalog and announcement-version gates, `fallow audit --base origin/main --fail-on-issues` all green. 30 new server tests + 6 lens tests. The CLI report was smoke-tested against a seeded DB and against a pre-migration DB (which now reports plainly instead of throwing).

Closes #2151 — R1 only; follow-ups for R2–R6 and first-push CI-green are filed separately.


scoop:
# Review — R1 delivery-metrics lens (#2151)

Scope reviewed: `git diff 00699bd0b4f3277f8440a897d7e3bdb143126800...HEAD` (26 files).

The slice matches the approved plan closely: additive `reviewer_spawns.outcome` (DDL + migration),
archive-decoupled `delivery_facts`, a pure `buildDeliveryMetrics` builder, one route, one CLI
report, one lens, i18n + glossary + announcement. I found one real defect (CSS), plus a handful of
non-blocking notes.

**Identifiers / wiring**

- `setReviewerSpawnOutcome` is declared at `src/store.ts:4203` and reaches both new call sites via
  the `Pick<>` store slices at `src/review.ts:262` and `src/plan-gate.ts:481`.
- `f.criticSessionId` (`src/review.ts:1189`) is the same id passed as `reviewerSessionId` to
  `recordReviewerSpawn` at `src/review.ts:664` (both derive from the `criticSessionId` minted at
  `src/review.ts:503`), so the UPDATE targets the right row. Same for plan-gate:
  `reviewerSessionId` is minted at `src/plan-gate.ts:920`, recorded at `src/plan-gate.ts:1057` and
  stamped at `src/plan-gate.ts:1339`.
- The review outcome mapping (`src/review.ts:1188-1195`) is consistent with `ReviewDecision =
  "changes_requested" | "commented" | "error"` (`src/types.ts:704`) and with the auto-address
  loop's own findings-driven branch at `src/review.ts:1211` — a findings-free `commented` verdict
  resets the streak there and reads as `clean` here.
- Plan-gate outcome mapping covers all three `PlanDecision` values; `error` is the resolved default
  at `src/plan-gate.ts:1577`.
- `countSignalsByKind` queries `kind`/`ts`/`sessionId`, all present in the `signals` DDL at
  `src/store.ts:1442-1444`. `SignalKind` is already imported at `src/store.ts:19`.
- `DeliveryMetricsStore = Pick<SessionStore, ...>` names four methods; all four exist
  (`listMergedDeliveryFacts` `src/store.ts:4257`, `listReviewerSpawns` `src/store.ts:4182`,
  `earliestDeliveryFactAt` `src/store.ts:4269`, `countSignalsByKind` `src/store.ts:4326`).
- `handleUsageDelivery` is registered at `src/server.ts:7786`, between `handleUsageBreakdown` and
  `handleUsageTimeline`; its guard is `parts[2] === "delivery"`, disjoint from `breakdown`
  (`src/server.ts:4366`) and `timeline`, so no route shadows it. A synchronous handler is fine in
  `ROUTE_HANDLERS` — the dispatcher awaits every handler (`src/server.ts:7859`) and
  `handleLogout`/`handleMe`/`handlePing` are already sync (`src/server.ts:904`, `911`, `6629`).
- `LocalForge.prStatus`'s merged branch now returns `row.createdAt`; `LocalPr.createdAt: number` is
  non-optional (`src/store.ts:475-483`), so the field is always populated, and `PrStatus.createdAt`
  is `number | undefined` (`src/forge/types.ts:206`) — assignment is type-safe.
- `GitState extends PrStatus` (`src/forge/types.ts:243`), and the poller emits the spread status
  verbatim (`src/index.ts:1288`, and `src/index.ts:1986` for the autopilot open path), so
  `git.createdAt` really is the forge's PR-opened epoch (`src/forge/github.ts:950-959`).
- `deliveryFacts` is constructed at `src/index.ts:1824` before its two uses (`src/index.ts:1825`
  subscriber, `src/index.ts:1879` archived-forget) — no TDZ.
- `DELIVERY_FACT_RETENTION_MS` is exported at `src/config.ts:1025`, imported at `src/index.ts:9`,
  consumed at `src/index.ts:2574`.
- `scripts/delivery-report.ts` resolves the DB the same way as the existing report
  (`scripts/usage-report.ts:580` and `src/config.ts:19`: `SHEPHERD_DB ?? ~/.shepherd/shepherd.db`),
  and its inline SQL for facts/spawns/signals is byte-identical to the store methods it stands in
  for.

**i18n / glossary / announcement parity**

- All 31 new keys exist in BOTH catalogs — I enumerated them out of `ui/messages/en.json:3692-3722`
  and `ui/messages/de.json:3692-3722`; the two lists match key-for-key and line-for-line, and the
  placeholder sets agree (`{n}` in `usage_delivery_sample`/`unreviewed`/`critic_errors`, `{date}` in
  `measuring_since`, `{occurrences}`+`{sessions}` in `incident_count`).
- The glossary gate's three assertions all pass against `scripts/check-glossary.mjs`:
  `gloss_first_pass_rate_term/_def` and `gloss_lead_time_term/_def` exist in both catalogs
  (`ui/messages/en.json:3717-3720`, `ui/messages/de.json:3717-3720`); the markers used
  (`[[first-pass-rate|…]]`, `[[lead-time|…]]`, `[[rework|…]]`) all resolve — `first-pass-rate` and
  `lead-time` are registered at `ui/src/lib/glossary.ts:82` and `:88`, `rework` pre-exists at
  `ui/src/lib/glossary.ts:76`; and the one `external` term carries both Wikipedia slugs
  (`ui/src/lib/glossary.ts:92-95`), matching the `en`/`de` requirement at
  `scripts/check-glossary.mjs:155-169`. Marker ids also satisfy the `[a-z0-9-]+` pattern at
  `scripts/check-glossary.mjs:130`.
- `sinceVersion: "1.48.0"` in `ui/src/lib/feature-announcements/entries/v1.48.0-delivery-metrics.ts`
  is greater than the last release (`package.json:3` → `1.47.0`) and agrees with its filename and
  with the three sibling 1.48.0 entries already present.
- `targetId: "usage-delivery-tab"` is matched by exactly one `use:coachTarget` anchor
  (`ui/src/lib/components/Usage.svelte:275`); `coachTarget` was already imported at
  `ui/src/lib/components/Usage.svelte:23`. The entry id `delivery-metrics` is unique in the entries
  directory.
- `GlossaryText` takes a `text` prop (`ui/src/lib/components/GlossaryText.svelte:5`), `Sparkline`
  takes `{points, color, width, height, ariaLabel}` with `points` time-sorted ASC
  (`ui/src/lib/components/usage/Sparkline.svelte:6-25`) — `trendBuckets` sorts by `dayKey` ascending
  (`src/delivery-metrics.ts:207`), so that contract holds. `formatPct(0.5) === "50%"`
  (`ui/src/lib/components/usage/format.ts:25-27`).
- `UsageRange` was already imported in `ui/src/lib/components/Usage.browser.test.ts:9`, so the new
  `inlineDelivery` fixture compiles.

**Behaviour**

- The `mergedAt`-in-`archive()` claim checks out: `settleMergedSession` is the only writer of
  `archive(id, "merged")` (`src/merge-teardown.ts:119` and `:142`), reached from automerge
  (`src/automerge.ts:369`), drain (`src/drain.ts:3439`) and the manual merge route
  (`src/server.ts:4056`), each routed through `SessionService.archive` → `store.archive`
  (`src/service.ts:5765`). So the train path genuinely cannot bypass the stamp.
- `delivery_facts` is genuinely outside the prune cascade — `pruneArchivedSessions` only deletes
  from `sessions` and the tables cascaded above `src/store.ts:4406`; the new table is not among
  them, and `test/delivery-facts.test.ts` pins that with an explicit prune assertion.
- The lens's null-vs-zero contract holds end to end: `rate()`/`sample()`
  (`src/delivery-metrics.ts:61` and `:79`) return `{value: null, n: 0}` on an empty sample, and
  `pct`/`count`/`durationOf` in `ui/src/lib/components/usage/DeliveryLens.svelte:451-461` render
  `EMPTY` for it.
- The Delivery `$effect` (`ui/src/lib/components/Usage.svelte:167-169`) mirrors the Timeline one
  (`:131-133`) exactly, including the monotonic stale-response token — no regression there.

**Could not verify**: whether the six follow-up issues from plan step 8 were actually filed and
linked to #2151 — that lives on GitHub, not in this repo. Assumed done as stated; not a finding.

## Findings

### 1. `ui/src/lib/components/usage/DeliveryLens.svelte` — the incident row's grid breaks at ≤480px

`.incident-row { grid-template-columns: 1fr auto }` is declared at
`ui/src/lib/components/usage/DeliveryLens.svelte:277-279`, but the narrow-screen block redefines
`.row { grid-template-columns: 1fr 3rem 3.5rem 3.5rem }` at `:298-300`. Both selectors have the same
specificity (Svelte's scoping class is appended to both equally), and the media-query rule comes
later in source order, so it wins for `<div class="row incident-row">` (`:179`) at ≤480px.

The guard that *was* added — `.row:not(.incident-row) > :last-child { display: none }` at `:303` —
only covers the `display` half. The result is that an incident row renders with the four-column
repo grid: the kind lands in the `1fr` track and the whole `usage_delivery_incident_count` string
("3× across 2 session(s)") is forced into the `3rem` track, where it wraps/overflows. The inline
comment at `:301-302` states the intent ("ONLY on the repo rows; the incident row's last cell IS its
content") that the grid override defeats.

Fix: either scope the override (`.row:not(.incident-row) { grid-template-columns: … }`) or restate
`.incident-row { grid-template-columns: 1fr auto }` inside the media block.

## Nits (non-blocking):

- `src/types.ts:1016` — `criticErrors` is documented as "Review spawns in the window that ended in `error`", but the value is only the errors belonging to merged, in-window tasks (`src/delivery-metrics.ts:155` folds over `tasks`); errored spawns on tasks that never merged are invisible. Worth tightening the wording since the same block also documents the per-repo rows.
- `ui/messages/en.json:3704` / `ui/messages/de.json:3704` — `usage_delivery_unreviewed` reads "{n} never reviewed", but a task whose only review spawns errored or never finalized also lands in `unreviewed` (`src/delivery-metrics.ts:152` counts `firstPass === null`); those were reviewed, just without a verdict.
- `src/delivery-metrics.ts:226-230` — `incidents()` is an identity map over the store rows; the builder could pass `store.countSignalsByKind(since)` straight through.
- `src/delivery-metrics.ts:210-224` — `DeliveryMetrics.tasks` (up to 50 rows, `MAX_TASK_ROWS`) is computed and shipped on every request, but nothing consumes it: `DeliveryLens.svelte` never reads `metrics.tasks` and `scripts/delivery-report.ts` doesn't either. It is only asserted in `test/server-usage-delivery.test.ts`.
- `ui/messages/en.json:3692` — `usage_tab_delivery` breaks the naming shape of its sibling `usage_timeline_tab` (noun-then-`_tab` vs `_tab_`-then-noun).

- `src/store.ts:4203` — `setReviewerSpawnOutcome` is an unconditional `UPDATE`, unlike the sibling `setReviewerSpawnProviderSessionId` at `src/store.ts:4161`, which is write-once (`AND providerSessionId IS NULL`). No path re-finalizes a spawn today (the `inflight` map gates it), but a future second finalize would silently overwrite a terminal outcome rather than being refused.
- `src/store.ts:3020-3033` — the `UPDATE sessions` inside `archive()` is guarded by `status!='archived'`, but the new `upsertDeliveryFact` in the same transaction is not. A `settleMergedSession` reaching an already-archived session (the archive-failure recovery / poller-race paths noted in `src/merge-teardown.ts:112-116`) still mints or completes a merged fact even though the archive itself no-ops. Harmless today — `COALESCE` first-write-wins means `mergedAt` cannot move — but the two writes in one transaction no longer share a condition.

- `src/delivery-metrics.ts:156-157` ships `reworkCyclesMean` and `planRoundsMedian`, of which only the CLI report renders them (`scripts/delivery-report.ts:104` and `:107`); the lens shows neither. Plan-sanctioned, so keeping them is the author's call.
- `src/delivery.ts:85-87` — `forget()` exists only to bound the in-memory `stamped` map, which requires a second subscriber in `src/index.ts:1879` to keep it correct. Reasonable as written; noting it only as an extra moving part the metric itself does not need.

scoop (commented):
# Review — R1 delivery-metrics lens (#2151), rework round 2

## Prior finding — status

**1. `ui/src/lib/components/usage/DeliveryLens.svelte`: narrow-screen `.row` grid overriding `.incident-row`. — ADDRESSED.**

Verified against `ui/src/lib/components/usage/DeliveryLens.svelte:301-307`: the media block now reads

```css
@media (max-width: 480px) {
  .row:not(.incident-row) { grid-template-columns: 1fr 3rem 3.5rem 3.5rem; }
  .row:not(.incident-row) > :last-child { display: none; }
}
```

Both rules carry the guard, so `.incident-row { grid-template-columns: 1fr auto }` at `DeliveryLens.svelte:277-279` is no longer beaten on source order — `.row:not(.incident-row)` is specificity (0,2,0) and simply does not match an incident row, while the unguarded `.row` at `DeliveryLens.svelte:269` is still overridden by the later equal-specificity `.incident-row`. The repo rows and the `.row-head` (`DeliveryLens.svelte:167`, no `incident-row` class) still get the 4-column narrow grid and still hide the lead-time cell. A regression test pins it at `ui/src/lib/components/usage/DeliveryLens.browser.test.ts:98-110` (viewport 390px, asserts two grid tracks and that the count cell keeps >48px).

## What I verified

**Identifiers and contracts**

- `setReviewerSpawnOutcome(reviewerSessionId, …)` keys on the same id each service registered: `src/review.ts:664` records `reviewerSessionId: criticSessionId` and `src/review.ts:1189` stamps `f.criticSessionId`; `src/plan-gate.ts:1057` records `reviewerSessionId` and `src/plan-gate.ts:1339` stamps `f.reviewerSessionId`. Both services added `"setReviewerSpawnOutcome"` to their store `Pick` (`src/review.ts:262`, `src/plan-gate.ts:481`), so the narrowed deps compile.
- Outcome enum vs. real decisions: `ReviewDecision` is `"changes_requested" | "commented" | "error"` (`src/types.ts:704`) and the stamp maps error → `error`, else `findings.length === 0` → `clean`, else `changes_requested` (`src/review.ts:1189-1194`) — `core.findings` is always an array out of `buildVerdictCore` (`src/review.ts:1471`). `PlanDecision` is approved/changes_requested/error (`src/plan-gate.ts:1576-1590`), mapped to approved/rework/error. Both match `ReviewerSpawnOutcome` at `src/types.ts:967`.
- The abort path that never spawns cannot leave a bogus row: `recordReviewerSpawn` runs only after `fitAndSpawn` returns a terminal id (`src/review.ts:634-663`), so a pre-spawn refusal has no `reviewer_spawns` row to mis-stamp.
- `DeliveryMetricsStore` picks `listMergedDeliveryFacts` / `listReviewerSpawns` / `earliestDeliveryFactAt` / `countSignalsByKind`; all four exist on `SessionStore` (`src/store.ts:4257`, `src/store.ts:4182`, `src/store.ts:4269`, `src/store.ts:4326`). `SignalKind` is imported at `src/store.ts:19` and the `signals` table really has `sessionId` + `ts` (`src/store.ts:1442-1444`), so `countSignalsByKind`'s SQL resolves.
- `GitState.createdAt` and `.number` exist (`src/forge/types.ts:206`, `:201`), and `createdAt` is populated on the merged branch by every forge the metric depends on: `src/forge/github.ts:1233` (`mapGhPr`), `src/forge/github.ts:1321` (`mapRestPull`), `src/forge/gitea.ts:235`, and now `src/forge/local.ts:393` — `row.createdAt` is a real `LocalPr` field (`src/store.ts:1909`). No other forge returns `state: "merged"` (grepped `src/forge/*.ts`).
- Route: `handleUsageDelivery` is registered at `src/server.ts:7786`, between `handleUsageBreakdown` and `handleUsageTimeline`; the dispatcher `await`s each handler (`src/server.ts:7859`) and sync `Response | null` handlers are the existing norm (e.g. `src/server.ts:960`), so the non-async signature is fine. Its path guard (`api`/`usage`/`delivery`) cannot be swallowed by `handleUsageLimits`, whose branches cover `github` / `history` only (`src/server.ts:4341-4349`). Auth runs before `ROUTE_HANDLERS` (`src/server.ts:6646`).
- Store surface exposed to the client is no wider than the existing precedent: `DeliveryRepoRow.repoPath` is an absolute path, same as the repo rows in `src/usage-breakdown.ts:262`.

**Schema / retention**

- The `outcome` column is in BOTH the fresh DDL (`src/store.ts:1158`) and `migrateReviewerSpawnColumns` (`src/store.ts:4751`), which is exactly the #1816 invariant `test/store.test.ts:1566-1570` asserts (the added `expect(cols).toContain("outcome")` runs against a raw DDL-only DB).
- `delivery_facts` has no FK and is genuinely absent from the prune cascade — I read `pruneArchivedSessions` (`src/store.ts:4352-4412`) and there is no `DELETE FROM delivery_facts`; `test/delivery-facts.test.ts:150-158` pins it.
- No NOT NULL hazard from the `archive()` stamp: `sessions.desig`, `repoPath`, `createdAt` are all `NOT NULL` (`src/store.ts:1196-1206`), so the `desig TEXT NOT NULL DEFAULT ''` column in `delivery_facts` (`src/store.ts:1292`) can never receive a bound NULL and abort the archive transaction.
- `store.archive(id, 'merged')` is the only merge chokepoint: `src/merge-teardown.ts:119` and `:142` are the sole `archive(…, "merged")` writers, wired through `Service.archive` → `this.deps.store.archive(id, reason)` (`src/service.ts:5766`) from the train (`src/automerge.ts:369`), drain (`src/drain.ts:3439`) and the manual route (`src/server.ts:4056`). So a train merge, which fires no `session:git`, still produces a fact row — `test/delivery-facts.test.ts:118-131`.
- First-write-wins is real: `upsertDeliveryFact` COALESCEs `issueNumber`/`prNumber`/`prOpenedAt`/`mergedAt` against the existing row (`src/store.ts:4236-4239`), so the archive stamp cannot clobber an earlier poller observation (`test/delivery-facts.test.ts:141-148`).
- The boot-comment arithmetic was updated consistently (`src/store.ts:1183-1186`): +1 CREATE TABLE, +1 index, +1 ALTER, +3 total. (The pre-existing 4-statement discrepancy between the headline number and the itemised sum is unchanged by this PR and is not a defect it introduced.)

**i18n / glossary / catalog**

- Locale parity: 31 new keys at the tail of `ui/messages/en.json` and the identical 31 at the tail of `ui/messages/de.json` — same names, same placeholders (`{n}`, `{date}`, `{occurrences}`/`{sessions}`). Every key the component calls resolves in both: `usage_tab_delivery`, `usage_delivery_{heading,caption,measuring_since,empty,sample,first_pass,rework,plan_rework,ttfr,lead_time,merged,unreviewed,critic_errors,trend_heading,trend_caption,repos_heading,col_repo,col_merged,col_first_pass,col_rework,col_lead,incidents_heading,incident_count,incidents_caption}` — checked one by one against both catalogs.
- Glossary: `gloss_first_pass_rate_term/_def` and `gloss_lead_time_term/_def` exist in both catalogs and follow the `gloss_<id>_term` convention `scripts/check-glossary.mjs:104-116` enforces for ids `first-pass-rate` / `lead-time` (`ui/src/lib/glossary.ts:81-95`). Every `[[id|label]]` marker introduced in the catalogs points at a registered id — `first-pass-rate`, `rework` (pre-existing, `ui/src/lib/glossary.ts:75-79`), `lead-time` — which is the second thing that gate checks. The external `lead-time` term carries both `en` and `de` Wikipedia slugs, as the gate requires for `kind: "external"`.
- Feature announcement: `sinceVersion: "1.48.0"` is correct — `package.json:3` holds the last release `1.47.0`, and sibling entries `v1.48.0-decommission-open-pr.ts` / `v1.48.0-fable-5-1.ts` already use it. Filename matches the version+id convention documented at `ui/src/lib/feature-announcements.ts:18-23`; the catalog is glob-loaded (`ui/src/lib/feature-announcements.ts:47-50`) so no index edit is needed. `targetId: "usage-delivery-tab"` matches `use:coachTarget={"usage-delivery-tab"}` at `ui/src/lib/components/Usage.svelte:275`, and `coachTarget` was already imported (`Usage.svelte:23`).
- `feat_delivery_metrics_title/_body` exist in both catalogs.
- `ui/src/lib/docs-manifest.ts:84` keyword regen matches the two new headings and their position in `docs-site/src/content/docs/reference/glossary.md:61` and `:180`.

**UI wiring**

- The Delivery tab copies the Timeline pattern exactly: own monotonic token, lazy `$effect` gated on `tab === "delivery"` so `range` is only tracked while the tab is active (`Usage.svelte:151-171`), banner-style error, and the four `$derived` predicates each gained one arm (`Usage.svelte:184-215`). Compared line-for-line against the timeline arms directly above them.
- `Sparkline` is fed time-ascending points — `trendBuckets` sorts by `dayKey` ascending (`src/delivery-metrics.ts:196`) and the component documents/requires ASC (`ui/src/lib/components/usage/Sparkline.svelte:3`); it is only rendered with `> 1` point, avoiding the single-point branch entirely.
- `formatPct` takes a 0–1 fraction (`ui/src/lib/components/usage/format.ts:25`) and `firstPassRate`/`planReworkRate` are shares (`src/delivery-metrics.ts:88`), so the tiles are not off by 100×.

**Metric correctness**

- A crashed critic run is excluded from rework: `applyReviewSpawn` counts only `clean`/`changes_requested` into `reviewRounds` and buckets `error` separately (`src/delivery-metrics.ts:100-107`), pinned by `test/delivery-metrics.test.ts:107-120`.
- A three-round rework task cannot read as first-pass: `firstPass` requires `reviewRounds === 1 && changesRequested === 0` (`src/delivery-metrics.ts:145-146`), pinned by `test/delivery-metrics.test.ts:79-105`.
- Null-not-zero holds end to end: `sample()`/`rate()` return `{value: null, n: 0}` (`src/delivery-metrics.ts:68-84`), the lens renders `EMPTY` (`DeliveryLens.svelte:14-28`), and the route test asserts nulls on an empty install (`test/server-usage-delivery.test.ts:84-91`).
- Range validation 400s on a bad `range` and defaults to `7d` (`src/server.ts:4392-4394`), matching `handleUsageBreakdown`'s validation verbatim.

## Stated limitations

- I did not run lint, tests or the fallow gate (CI's job, and the instructions forbid it) — the assertions above are read against source, not execution output.
- I could not verify the `fallow` cognitive-complexity ceiling the plan flags for `Usage.svelte`'s `bodyLoading`/`hasContent` predicates: the tool is not present in this repo. The predicates each gained one `||` arm; whether that crosses the gate is for CI to say.

## Nits (non-blocking)

- `src/delivery-metrics.ts:173` and `:214` use bare `basename(repoPath)` where the sibling builder uses `basename(repoPath) || repoPath` (`src/usage-breakdown.ts:244`) — a repoPath of `/` (or an empty-basename edge) would render a blank repo label instead of falling back to the path.
- `src/config.ts:1020-1025`: the comment argues the two retention windows must stay equal because the tables are read together, but they are not keyed alike — `pruneDeliveryFacts` filters on the session's `createdAt` (`src/store.ts:4278-4288`) while `pruneReviewerSpawns` filters on `spawnedAt` (`src/store.ts:4190`). A task created more than 90 days before it merges loses its fact row while its spawns survive, which is the asymmetry the comment says equal windows prevent.
- `src/types.ts:1016`: `criticErrors` is documented as "Review spawns **in the window** that ended in `error`", but `statsFor` sums each merged task's whole-history error count (`src/delivery-metrics.ts:158`), so the number is per-merged-task, not per-in-window-spawn.
- `ui/src/lib/components/usage/DeliveryLens.svelte:31-37` (`humanMs`): sub-30-second durations render as `0m`. A very fast time-to-first-review reads as zero rather than "<1m".
- `scripts/delivery-report.ts:41` re-derives the DB path (`process.env.SHEPHERD_DB ?? $HOME/.shepherd/shepherd.db`) instead of importing `config.dbPath` the way `scripts/usage-report.ts` does (`src/config.ts:19` is the same expression today, so the two can silently drift).
- `scripts/delivery-report.ts:31-37`: `--range` passed as the final argument with no value silently falls through to the `7d` default rather than erroring.

- `docs/token-usage-analysis.md:16` adds a Fable-pricing caveat that belongs to #2150 (the diff base commit), not to R1 — it rides along as a docs sync.
- `DeliveryMetrics.tasks` (`src/delivery-metrics.ts:212-226`, capped at 50 rows carrying sessionId/desig/issue/PR) is computed and shipped in the payload but rendered by neither `DeliveryLens.svelte` nor `scripts/delivery-report.ts` — dead payload for now.
- `reworkCyclesMean` (`src/delivery-metrics.ts:155`) is mirrored into `ui/src/lib/types.ts` but only the CLI report reads it; the lens shows the median only.

- `src/store.ts:1183-1186`'s boot-statement tally has an internal arithmetic mismatch that predates this change (the headline count exceeds the itemised sum by 4); this PR only shifts each term by +1, preserving the discrepancy.

PR #2160, opened 2026-09-01
feat(issues): name the gh transports behind a failed issue load

kai-osthoff:
Clicking **Erneut versuchen** on a failed issue load told the operator nothing about what
Shepherd was actually doing. The screenshot that started this: *"Issues konnten nicht geladen
werden – evtl. GitHub-Rate-Limit."* — so is the answer to wait an hour, or to re-run `gh auth
login`? No way to tell.

The information already existed server-side and was thrown away.

## What was happening

`GithubForge.listIssues` lists issues over two transports on two **independent** GitHub budgets —
`gh issue list` (GraphQL bucket) and `gh api` (REST bucket) — and falls back between them in
whichever order the GraphQL backoff prefers. When both gave up, it rethrew the preferred
transport's error, and `handleIssues` caught it in an argument-less `catch {}` that answered with
nothing but `error: "fetch_failed"`. The diagnosis died there.

## What this changes

`listIssues` now keeps a trail of the transports that **actually ran** and failed, built as it
goes, and attaches it to the escaping error under a symbol key (so it never shows up in a log line
or a serialized error). A run that ends after a single transport reports one entry rather than
claiming a second. `/api/issues` passes the trail on as an optional `attempts` field, and both
retry surfaces render it:

```
Issues konnten nicht geladen werden – evtl. GitHub-Rate-Limit.  Erneut versuchen
Versucht:
  gh issue list (GraphQL) → Rate-Limit
  gh api (REST) → HTTP 403
```

Lines appear in execution order, so what was tried *first* is visible without extra words.
Hovering a line reveals the full `gh` message. Reasons are classified into rate limit / not signed
in / repo not found / gh not installed / network / HTTP status / unknown — and where the
classification is a guess (a bare 403 is reported as `HTTP 403`), the raw message in the tooltip
is the escape hatch.

Surfaces: the New Task modal (`PromptSources.svelte`) and the Issues panel
(`IssuesPanel.svelte`). Up Next is untouched — it aggregates several repos into one server
snapshot and would need per-repo folding.

## Safety

The raw `gh` output reaches the browser for the first time, so it is sanitized server-side before
it travels: ANSI stripped, whitespace collapsed, capped at 300 characters, and token shapes
(`ghp_…`, `github_pat_…`) redacted. Covered by tests.

Responses that carry no trail — Gitea, local forges, a client-side rejection, or a failure raised
outside the listing itself — are **byte-identical** to before, and the existing whole-body
`toEqual` test in `test/server-issues.test.ts` guards that.

## Note on the diff

`handleIssues` crossed the complexity gate once the failure branch grew, so its body moved into
`issuesResponse(forge)` and the handler is now just its route guards. Pure move, no behaviour
change.

## Verification

- `bun run lint`, `bun run test` (root)
- `cd ui && bun run check`, `bun run check:i18n`, `bun run test`
- New: `test/forge/gh-attempt.test.ts` (classification, truncation, redaction, symbol round-trip);
  extended `test/forge/github.test.ts` (trail in both transport orders, preferred error still
  rethrown, no trail on a successful fallback), `test/server-issues.test.ts` (one entry, two
  entries, none), and browser tests in both components (two lines, one line, no trail).

### DIFF
--- a/src/forge/gh-attempt.ts
+++ b/src/forge/gh-attempt.ts
@@ -1,0 +1,119 @@
+/**
+ * Per-transport failure record for a `gh`-backed listing.
+ *
+ * `GithubForge.listIssues` draws on two INDEPENDENT GitHub budgets (`gh issue list`
+ * on the GraphQL bucket, `gh api` on the REST bucket) and falls back between them.
+ * When the listing fails, the operator sees only "couldn't load issues" — which
+ * transport was tried, and why each one gave up, dies in the forge. These helpers
+ * carry that trail out on the thrown error so `/api/issues` can report it and the
+ * UI's retry state can name the paths instead of guessing at a rate limit.
+ */
+
+import { isRateLimitError } from "./rate-limit";
+
+/** Which `gh` transport produced the failure. */
+export type GhTransport = "cli" | "rest";
+
+/** Classified cause of one failed transport. `http` is the catch-all for a status
+ *  we don't name specially; `unknown` for output that carries no status at all. */
+export type GhFailureReason =
+  "rate_limit" | "auth" | "not_found" | "gh_missing" | "network" | "http" | "unknown";
+
+/** One executed-and-failed transport, as reported to the UI. */
+export interface GhFetchAttempt {
+  transport: GhTransport;
+  reason: GhFailureReason;
+  /** HTTP status when `gh` reported one (`(HTTP 403)`); absent otherwise. */
+  status?: number;
+  /** Sanitized `gh` output for the UI's hover tooltip. May be empty. */
+  detail: string;
+}
+
+/** Cap on `detail`: enough for a full `gh` error line, short of pasting a stack. */
+const DETAIL_MAX = 300;
+
+/** Token shapes `gh` could echo back (`gh auth token` output pasted into an env,
+ *  a PAT in a remote URL). This text is newly leaving the server for the browser,
+ *  so redact before it travels rather than trusting `gh` to keep quiet. */
+const TOKEN_PATTERNS: RegExp[] = [
+  /\bgh[pousr]_[A-Za-z0-9]{16,}/g,
+  /\bgithub_pat_[A-Za-z0-9_]{16,}/g,
+];
+
+// eslint-disable-next-line no-control-regex -- stripping ANSI SGR sequences requires \x1b
+const ANSI = /\x1b\[[0-9;]*[A-Za-z]/g;
+
+/** Raw text of a `gh` failure: execFile rejections carry stderr, everything else
+ *  falls back to the message. */
+function errorText(err: unknown): string {
+  const e = err as Record<string, unknown> | null | undefined;
+  const stderr = typeof e?.stderr === "string" ? e.stderr : "";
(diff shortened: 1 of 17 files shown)


kai-osthoff:
**Re: `sanitizeDetail` leaves credential forms intact** — valid, fixed in `4eaddcb8`. Now redacted, and deliberately *before* the 300-char cap so nothing survives as a truncated head: `Authorization:` header values (gh prints request headers under `GH_DEBUG=api`), credentials embedded in a remote URL (`https://user:pat@…`, including the Actions `x-access-token:` form), `GH_`/`GITHUB_TOKEN` env assignments echoed into an error line, GitHub App/installation JWTs, and the `gho_`/`ghu_`/`ghs_`/`ghr_` prefixes alongside `ghp_`. The two label-preserving rules keep the message readable — the header or URL keeps its shape, only the secret half goes.

**One shape I deliberately left alone: a bare 40-hex string.** That is the form of a legacy OAuth token *and* of every git SHA in gh output. A pattern broad enough to catch the token would redact commit SHAs out of the diagnosis, which is the whole point of the tooltip — and legacy 40-hex PATs were deprecated by GitHub in 2021. If you think the trade should go the other way, say so and I will flip it. It is pinned by a test (`sanitizeDetail: leaves an ordinary diagnosis — URLs and SHAs — readable`) so the choice cannot be undone by accident.

kai-osthoff (commented):
# Review

Decision: comment

No blocking findings.

## Verified

- The GitHub listing records each failed transport in execution order and attaches that trail to the preferred error only after both paths fail (`src/forge/github.ts:838-850`). The route reads that trail and includes `attempts` only when it is non-empty (`src/server.ts:5965-5984`), while the existing blocker fetch continues to propagate the original listing error (`src/server.ts:5882-5893`).
- Browser-bound diagnostic text is normalized, redacted before truncation, and capped at 300 characters (`src/forge/gh-attempt.ts:77-82`); the tooltip receives that `detail` only when it is non-empty (`ui/src/lib/components/IssueLoadAttempts.svelte:61-75`).
- The API contract declares the optional trail (`ui/src/lib/api.ts:1181-1207`), and both consumers clear or replace it along with their failure state (`ui/src/lib/components/new-task/issue-data.svelte.ts:39-71`, `ui/src/lib/components/IssuesPanel.svelte:243-263`, `ui/src/lib/components/IssuesPanel.svelte:313-348`).
- The display covers every server reason, including the guarded HTTP-status case (`ui/src/lib/components/IssueLoadAttempts.svelte:25-57`). Both maintained locales are configured in `ui/project.inlang/settings.json:4` and contain the same added display and announcement keys (`ui/messages/en.json:3723-3734`, `ui/messages/de.json:3723-3734`).
- The announcement entry uses the catalog's eager `entries/*.ts` discovery convention (`ui/src/lib/feature-announcements/entries/v1.48.0-issue-load-transports.ts:1-14`, `ui/src/lib/feature-announcements.ts:47-75`).

Per request, no build, test, lint, or other execution was run.

kai-osthoff:
🎉✨ READY 🚀 Zero conflicts 0️⃣💯 Hey reviewer 👀 — wake up ☕😴 and merge 🔨

PR #2164, opened 2026-09-02
feat(review): non-blocking plan-drift report fed to the delivery lens (#2155)

### LINKED ISSUE #2155: Plan-drift report: non-blocking planDrift from the critic, fed to the delivery lens
Follow-up to #2151 (R3, `docs/research/ai-native-sdlc-playbook.md` §3 R3). R1 shipped in #2153 and is the substrate for this.

The critic's "plan is context, never a warrant" stance is correct — the plan must not become a straitjacket. But *systematic* divergence between approved plans and merged diffs is a process signal (plans too vague, gate answering the wrong questions).

Cheap adaptation: ask the critic for one **non-blocking** `planDrift: none|minor|major` field plus a one-line note when the diff departs materially from the `APPROVED PLAN` block it already receives. Persist it on `reviews`, and surface it in the Delivery lens alongside plan rework.

No change to review judgment — pure measurement. Only populated for sessions that actually went through the plan gate.

scoop:
Closes #2155.

Follow-up to #2151 (R3), building on the delivery-metrics substrate shipped in #2153.

The critic answers one extra field when — and only when — it is already shown the `APPROVED PLAN` block: `planDrift: none|minor|major` plus a one-line note. It is pure measurement: never a finding, never moves a decision, never advances the rework streak. Plan-less session prompts and the standalone PR critic stay byte-identical.

## Persisted twice, on purpose

The issue asks for `reviews`. That alone cannot work: `archive()` deletes the review row, and every task the Delivery lens counts is a merged — therefore archived — task, so a reviews-only column would read `null` in the lens 100% of the time.

| Column | Table | Lifetime | Consumer |
| --- | --- | --- | --- |
| `planDrift` | `reviewer_spawns` | archive-decoupled, append-only | Delivery lens + CLI report |
| `planDrift`, `planDriftNote` | `reviews` | dies with `archive()` | live review popover |

`reviewer_spawns` is where #2153 put `outcome` for exactly this reason; the drift rides the same single stamp at finalize.

## Aggregation

The **last measured round** wins — the drift of the diff closest to merge, not an intermediate round's; a later unmeasured round does not erase an earlier measurement. The denominator is tasks that carry a measurement at all, so an unmeasured task is excluded rather than counted as `none`, which would flatter the rate.

## Surfaced

- **Delivery lens**: a `Plan drift` tile with its sample size (em dash, never `0%`, when unmeasured) plus a caption calling out `major` divergences.
- **`bun run delivery-report`**: a `Plan drift rate` row.
- **Review popover**: a muted one-line note (`git-rail/PlanDriftNote.svelte`) — no chip, no badge, nothing that competes with the decision the popover reports. Its own component because a branch nested that deep in `GitRail`'s template costs more complexity than the two lines it renders.

## Notes

- The drift note is collapsed to one line and hard-capped at 140 chars server-side, and the prompt asks for it without quotation marks — it rides the verdict JSON, whose escaping is what #2042 was about. An unrecognized level normalizes to `null` = not measured.
- Two guards discard a drift answer: the run was shown no plan (nothing to measure against), or it errored (no reviewed diff behind it).
- Forward-only, like the rest of the delivery instrumentation; the lens already reports `measuringSince`.

## Verification

`bun run lint`, `bun run test` (8867 pass), `cd ui && bun run check && bun run test` (4741 pass), `fallow audit --base origin/main --fail-on-issues`, glossary / i18n-parity / feature-catalog / docs-manifest gates — all green. The non-blocking guarantee is a test, not a comment: a major-drift, findings-free verdict must still resolve clean, steer nothing and advance no streak.

### DIFF
--- a/docs-site/src/content/docs/reference/glossary.md
+++ b/docs-site/src/content/docs/reference/glossary.md
@@ -67,6 +67,13 @@ The share of merged tasks that passed review in a single clean round — no rewo
 requested. Tasks that were never reviewed are left out of the calculation
 entirely rather than counted as passes.
 
+### Plan drift
+
+How far a merged change departed from the plan approved before it was written, as
+reported by the PR critic: none, minor or major. It never affects the review's
+verdict — departing from a plan is legitimate. It is measured because a repo that
+drifts constantly is usually writing vague plans, not writing bad code.
+
 ### Inferred
 
 Derived by the recap model from the code — not verified against the real diff.
--- a/scripts/delivery-report.ts
+++ b/scripts/delivery-report.ts
@@ -108,6 +108,12 @@ function rows(st: DeliveryStats): [string, string][] {
     ["Critic errors", String(st.criticErrors)],
     ["Plan rounds (median)", num(st.planRoundsMedian)],
     ["Plan rework rate", pct(st.planReworkRate)],
+    // #2155: over tasks the critic measured against an approved plan; `major` is called out because
+    // a rate alone hides whether the drift is incidental or a different approach altogether.
+    [
+      "Plan drift rate",
+      `${pct(st.planDriftRate)}${st.planDriftMajor > 0 ? ` — ${st.planDriftMajor} major` : ""}`,
+    ],
     ["Time to first review", dur(st.timeToFirstReviewMs)],
     ["Lead time", dur(st.leadTimeMs)],
   ];
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -11,7 +11,7 @@ import { join } from "node:path";
 import { execFile } from "node:child_process";
 import { promisify } from "node:util";
 import { execFileSync, timedAsync } from "./instrument";
-import type { ReviewDecision } from "./types";
+import type { PlanDrift, ReviewDecision } from "./types";
 import type { SessionUsage } from "./usage";
 import { tolerantParseJson } from "./json-tolerant";
 import type { VerdictRead } from "./json-tolerant";
@@ -328,6 +328,9 @@ export function reviewPrompt(
         reviewPolicy: opts.reviewPolicy,
         houseRules: opts.houseRules,
         houseRulesAuthored: true,
+        // #2155: planShown mirrors the `if (opts.plan …)` block above — the drift question is only
+        // asked when there is a plan in the prompt to measure against.
+        planShown: Boolean(opts.plan?.trim()),
       },
     ),
   );
@@ -758,6 +761,10 @@ function scopeAndOutputTail(
     /** True only where a Shepherd agent authored the diff under this repo's rules — see
      *  {@link reviewerHouseRulesBlock}. The standalone critic never sets it. */
     houseRulesAuthored?: boolean;
+    /** #2155: an APPROVED PLAN block is present above, so the verdict can carry the non-blocking
+     *  plan-drift measurement. Absent ⇒ no drift block and no extra JSON keys, keeping every
+     *  plan-less prompt (and prReviewPrompt, which never has a plan) byte-identical. */
+    planShown?: boolean;
   } = {},
 ): string[] {
   return [
@@ -890,12 +897,35 @@ function scopeAndOutputTail(
     '- At most 5 NEW findings per review, highest-severity first. This is a PRIORITISATION instruction, never a limit that may lose information: if more than 5 genuine blocking problems exist, emit ALL of them and say so in "summary" (the change needs reworking, not tweaking). NEVER move a blocking problem into `Nits (non-blocking):`, and never drop one, to fit the budget.',
     "- Re-raised prior findings do NOT count against those 5.",
     "",
+    // PLAN-DRIFT REPORT (#2155) — PURE MEASUREMENT, and the only field in this prompt that feeds no
+    // decision anywhere: nothing in the review loop, the auto-address steer or the merge train
+    // reads it. It exists because SYSTEMATIC divergence between approved plans and merged diffs is
+    // a process signal (plans too vague, the gate asking the wrong questions) — a per-PR verdict is
+    // not. Emitted ONLY when a plan was actually shown, so a plan-less prompt stays byte-identical
+    // and a critic with nothing to compare against can never invent a level.
+    //
+    // The non-judgement wording is load-bearing, not decoration: the prompt's own stance is that a
+    // plan is "CONTEXT for intent, never a warrant", and asking about fidelity at all risks
+    // anchoring the reviewer into plan-conformance findings — exactly what that stance forbids.
+    ...(opts.planShown
+      ? [
+          "PLAN-DRIFT REPORT — a MEASUREMENT for the operator, never a judgement about this PR:",
+          '- Report how far the diff departs from the APPROVED PLAN above as "planDrift": "none" (it implements the plan\'s approach), "minor" (same approach, incidental departures — a differently named helper, an extra file, steps done in another order), or "major" (a different approach, a planned piece missing, or substantial work the plan never described).',
+          '- Add "planDriftNote": ONE line, at most 140 characters, naming the single biggest departure. Omit it when planDrift is "none". Write it WITHOUT quotation marks of any kind — it travels inside the JSON file.',
+          '- This changes NOTHING about your review. Drift is never a finding, never affects "decision", and never appears in "findings": a "major" drift on a correct diff is still a clean review, and "none" excuses no defect. Judge the code exactly as you would if this field did not exist.',
+          "- Departing from a plan is legitimate — the plan is context, not a warrant. Report what you observe; do not editorialize and do not ask the author to conform to the plan.",
+          "- DECONFLICTION with the SCOPE-CREEP lens above: the two measure different things, so a departure you reported under `Scope creep / gold-plating (non-blocking):` still counts here. That section is a judgement call addressed to the author; this field is a process metric addressed to the operator.",
+          "",
+        ]
+      : []),
     "When done, write TWO files in the repository root:",
     // Binds the term once: every rule above says `"body"` (sections, citations, stated limitations),
     // and they all now resolve to this file. Cheaper and less drift-prone than restating ~10 rules.
     '1. `.shepherd-review.md` — your full markdown review. Everywhere these instructions say "body" — the named sections, the `path:line` citations, the stated limitations — they mean THIS file. It is NOT JSON: write the prose directly, escape nothing, quote however you like.',
     "2. `.shepherd-review.json` — the structured verdict, with this shape:",
-    '{"decision": "request-changes" | "comment", "summary": "<=100 char one-liner", "findings": ["<discrete actionable item>", ...], "body": "<optional — see below>"}',
+    opts.planShown
+      ? '{"decision": "request-changes" | "comment", "summary": "<=100 char one-liner", "findings": ["<discrete actionable item>", ...], "planDrift": "none" | "minor" | "major", "planDriftNote": "<one line, <=140 chars, no quotation marks — omit when planDrift is none>", "body": "<optional — see below>"}'
+      : '{"decision": "request-changes" | "comment", "summary": "<=100 char one-liner", "findings": ["<discrete actionable item>", ...], "body": "<optional — see below>"}',
     // #2042: the body used to live inside this JSON. A single unescaped `"` before a `:` or `,` —
     // which German `„…":` prose produces constantly — is indistinguishable from a real key/value
     // boundary, so it silently destroyed complete verdicts. Keeping the prose out of JSON entirely
@@ -919,6 +949,9 @@ function scopeAndOutputTail(
  *  scrubStaleVerdictArtifacts). */
 export const VERDICT_FILE = ".shepherd-review.json";
 
+/** Hard cap on the critic's plan-drift note (#2155) — one line, not a second review body. */
+export const PLAN_DRIFT_NOTE_MAX = 140;
+
 /** The critic's verdict BODY, written beside {@link VERDICT_FILE} as plain markdown.
  *
  *  #2042: the body is the one field that is long free prose, and the critic hand-authors its JSON —
@@ -940,6 +973,9 @@ export interface RawVerdict {
   summary?: unknown;
   body?: unknown;
   findings?: unknown;
+  /** #2155 — non-blocking plan-drift measurement; present only when the critic was shown a plan. */
+  planDrift?: unknown;
+  planDriftNote?: unknown;
 }
 
 /** Fingerprint the branch diff with `git patch-id` so a rebase (same diff, new SHA) is a
@@ -1129,6 +1165,25 @@ export function normalizeFindings(raw: unknown): string[] {
     .filter((f): f is string => typeof f === "string")
     .map((f) => f.trim())
     .filter(Boolean);
+}
+
+/** Coerce the critic's `planDrift` field to a known level (#2155). Anything else — a missing field,
+ *  a sentence, a novel level — is `null` = NOT MEASURED, which every consumer excludes rather than
+ *  reading as `none`. Deliberately NOT part of buildVerdictCore: that is shared with the standalone
+ *  PR critic, which is never shown a plan. */
+export function normalizePlanDrift(raw: unknown): PlanDrift | null {
+  if (typeof raw !== "string") return null;
+  const v = raw.trim().toLowerCase();
+  return v === "none" || v === "minor" || v === "major" ? v : null;
+}
+
+/** The one-line drift note, collapsed to a single line and hard-capped at
+ *  {@link PLAN_DRIFT_NOTE_MAX} chars. Empty or non-string → null. Both are enforced here rather
+ *  than merely asked for in the prompt: the note is model-authored free prose that lands in a DB
+ *  column and then in a one-line UI slot, so a paragraph must not arrive intact. */
+export function normalizePlanDriftNote(raw: unknown): string | null {
+  if (typeof raw !== "string") return null;
+  return raw.replace(/\s+/g, " ").trim().slice(0, PLAN_DRIFT_NOTE_MAX) || null;
 }
 
 /** A leading token looks like a repo-relative file path: it has no space AND (contains a `/` OR
--- a/src/delivery-metrics.ts
+++ b/src/delivery-metrics.ts
@@ -9,6 +9,7 @@ import type {
   DeliverySample,
   DeliveryStats,
   DeliveryTaskRow,
+  PlanDrift,
   ReviewerSpawnRow,
   UsageRange,
 } from "./types";
@@ -37,6 +38,10 @@ interface TaskRounds {
   planRounds: number;
   /** Earliest review spawn, whatever its outcome — the review DID start. */
   firstReviewAt: number | null;
+  /** Plan drift as of the LAST review round that measured it (#2155) — the state of the diff
+   *  closest to merge, not an intermediate round's. null = never measured (no plan was shown, or
+   *  every round errored), which keeps the task out of the rate's denominator entirely. */
+  planDrift: PlanDrift | null;
 }
 
 const EMPTY_ROUNDS: TaskRounds = {
@@ -45,6 +50,7 @@ const EMPTY_ROUNDS: TaskRounds = {
   errors: 0,
   planRounds: 0,
   firstReviewAt: null,
+  planDrift: null,
 };
 
 /** Window start for a range; 0 for `all` so the caller filters on `>= 0` rather than on a
@@ -86,6 +92,9 @@ function rate(hits: number, total: number): DeliverySample {
 function applyReviewSpawn(t: TaskRounds, sp: ReviewerSpawnRow): void {
   // The review STARTED regardless of how it ended, so every spawn moves the first-review mark.
   if (t.firstReviewAt == null || sp.spawnedAt < t.firstReviewAt) t.firstReviewAt = sp.spawnedAt;
+  // Last non-null wins: spawns arrive in `spawnedAt` order, so the final measured round stands.
+  // A later round that measured nothing does NOT erase an earlier measurement.
+  if (sp.planDrift != null) t.planDrift = sp.planDrift;
   if (sp.outcome === "error") t.errors += 1;
   else if (sp.outcome === "clean" || sp.outcome === "changes_requested") {
     t.reviewRounds += 1;
@@ -144,6 +153,9 @@ function statsFor(tasks: TaskView[]): DeliveryStats {
   const gated = tasks.filter((t) => t.rounds.planRounds > 0);
   const reworkCycles = reviewed.map((t) => t.rounds.reviewRounds);
   const planRounds = gated.map((t) => t.rounds.planRounds);
+  // Only tasks the critic actually measured — one with no plan (or no verdict) is excluded, never
+  // silently counted as `none`, which would flatter the rate.
+  const drifted = tasks.filter((t) => t.rounds.planDrift != null);
   const ttfr = tasks.map((t) => t.timeToFirstReviewMs).filter((v): v is number => v != null);
   const lead = tasks.map((t) => t.leadTimeMs).filter((v): v is number => v != null);
   return {
@@ -155,6 +167,11 @@ function statsFor(tasks: TaskView[]): DeliveryStats {
     criticErrors: tasks.reduce((n, t) => n + t.rounds.errors, 0),
     planRoundsMedian: sample(planRounds, median),
     planReworkRate: rate(gated.filter((t) => t.rounds.planRounds > 1).length, gated.length),
+    planDriftRate: rate(
+      drifted.filter((t) => t.rounds.planDrift !== "none").length,
+      drifted.length,
+    ),
+    planDriftMajor: drifted.filter((t) => t.rounds.planDrift === "major").length,
     timeToFirstReviewMs: sample(ttfr, median),
     leadTimeMs: sample(lead, median),
   };
--- a/src/review.ts
+++ b/src/review.ts
@@ -39,6 +39,8 @@ import {
   scopeFindings,
   effectiveRound,
   buildVerdictCore,
+  normalizePlanDrift,
+  normalizePlanDriftNote,
   shouldSkipForPatchId,
   captureUsage,
   reapRun,
@@ -221,6 +223,9 @@ interface InFlight {
   priorReviewedPatchIds: string[]; // patch-ids reviewed on this streak before this run (churn/revert dedup set)
   priorSeenNoteIds: string[]; // seen-note set carried in (before this run's fetch)
   seenNoteIds: string[]; // priorSeen + notes fed to THIS run's critic; only "consumed" on a real verdict
+  /** #2155: an APPROVED PLAN block was in this run's prompt, so a plan-drift answer is meaningful.
+   *  False ⇒ the critic had nothing to measure against and any drift it reports is discarded. */
+  planShown: boolean;
   /** RAW `readAsync(terminalId, "visible")` buffer from the previous tick — NOT the collapsed
    *  single-line `readPaneTail()` shape. Backs the unchanged-buffer half of the stuck-pane test,
    *  so it must be the same shape the detector classifies. */
@@ -691,6 +696,7 @@ export class ReviewService {
       startedAt: this.now(),
       ...priorStreakState(prior),
       seenNoteIds,
+      planShown: Boolean(plan?.trim()),
     });
     // Persist the spawn row now (totals NULL until finalize) so review burn is attributable
     // even if the run crashes/times out before producing a verdict (issue #502).
@@ -1268,6 +1274,9 @@ export class ReviewService {
             : verdict.findings.length === 0
               ? "clean"
               : "changes_requested",
+          // #2155: the durable copy. The `reviews` row this verdict also lands on is deleted by
+          // archive(), and every task the delivery metrics count is an archived one.
+          verdict.planDrift,
         );
       } catch (err) {
         console.warn(`[review] outcome accounting failed for ${f.sessionId}:`, err);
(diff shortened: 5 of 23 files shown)


scoop (commented):
# Review — Plan-drift report (#2155)

Round 1 of at most 4. Decision: **comment** — no blocking defect found.

## What I verified

**The ask/measure/persist/aggregate/surface chain is complete and internally consistent.**

- *Ask.* `scopeAndOutputTail`'s new `planShown` option (`src/critic-core.ts:600`, block emitted at
  `src/critic-core.ts:741-750`) is set from `Boolean(opts.plan?.trim())` at
  `src/critic-core.ts:264`, which is the same predicate that gates the `APPROVED PLAN` block itself
  at `src/critic-core.ts:198` (`if (opts.plan && opts.plan.trim())`). The two cannot disagree, so
  the drift question is never asked without something to measure against.
- *Byte-identity of the plan-less prompt.* `prReviewPrompt` calls the tail at
  `src/critic-core.ts:302` with no `planShown` key, so `opts.planShown` is `undefined` → falsy →
  neither the block nor the widened JSON shape line (`src/critic-core.ts:757`) is emitted. The only
  two call sites of `scopeAndOutputTail` are lines 249 and 302 (grepped; `src/plan-gate.ts:53` is a
  comment reference, not a call).
- *Extra keys survive parsing.* `tolerantParseJson` (`src/json-tolerant.ts:37-47`) does a plain
  `JSON.parse`/`jsonrepair` with no key allowlist, and `RawVerdict` gained `planDrift` /
  `planDriftNote` at `src/critic-core.ts:806-808`, so the new fields actually reach `buildVerdict`.
- *Guards.* `src/review.ts:1467-1468` forces `null` unless `f.planShown && core.decision !== "error"`.
  `InFlight.planShown` is set at `src/review.ts:665` from the same `plan` value that is composed
  into the prompt at `src/review.ts:613`/`640-644`. I checked that the argv-fitting path cannot
  silently drop the plan while leaving `planShown` true: `fitCriticPrompt`
  (`src/review.ts:1920-1952`) only ever *clamps* a plan head-tail with a `minUseful` floor, and a
  plan that cannot fit produces a refusal (`fitted.ok === false` → `refuseOversizedSpawn`,
  `src/review.ts:1553-1556`) rather than a plan-less spawn.
- *Persistence, `reviews`.* Column list, placeholder count and bind array all agree: 22 columns at
  `src/store.ts:3260-3261`, exactly 22 `?` at `src/store.ts:3262` (verified with an anchored regex
  `VALUES \(\?(,\?){21}\)`), 22 bound values at `src/store.ts:3273-3296`, and both columns are in
  the `ON CONFLICT … DO UPDATE` list at `src/store.ts:3272`. `getReview`
  (`src/store.ts:3247-3248`) and `snapshotReviews` (`src/store.ts:3335-3336`) both select them, and
  `hydrateReview` runs the level through `coercePlanDrift` (`src/store.ts:368-370`, applied at
  `src/store.ts:3239`) rather than spreading the raw column. Fresh DDL at `src/store.ts:1320` and
  migration at `src/store.ts:4726-4729` both carry the columns, so a fresh install and an upgrade
  converge.
- *Persistence, `reviewer_spawns`.* Column in `REVIEWER_SPAWNS_DDL` (`src/store.ts:1167`) **and** in
  `migrateReviewerSpawnColumns` (`src/store.ts:4785-4787`) — the #1816 rule the plan cites. The only
  writer is `setReviewerSpawnOutcome` (`src/store.ts:4228-4237`), whose new third parameter defaults
  to `null`, so the plan-gate caller (`src/plan-gate.ts:1338`, two arguments) leaves the column NULL.
  `recordReviewerSpawn` (`src/store.ts:4097`) uses an explicit column list that does not include
  `planDrift`, so a fresh row starts NULL as intended.
- *Aggregation.* `applyReviewSpawn` (`src/delivery-metrics.ts:97`) takes the last non-null value; the
  "rows arrive in `spawnedAt` order" premise the comment leans on is real —
  `listReviewerSpawns()` is `SELECT * FROM reviewer_spawns ORDER BY spawnedAt` (`src/store.ts:4206`),
  consumed at `src/delivery-metrics.ts:271`. `applyReviewSpawn` is reached only for
  `sp.kind === "review"` (`src/delivery-metrics.ts:109-111`), so a plan-gate spawn can never
  contribute. Denominator is tasks with a non-null drift (`src/delivery-metrics.ts:158`), and
  `rate()` (`src/delivery-metrics.ts:85-87`) returns `{value: null, n: 0}` for an empty denominator,
  so an unmeasured window renders `—` and never `0%`.
- *Wire.* `ReviewService.snapshot()` returns whole `ReviewVerdict` objects
  (`src/review.ts:1840-1842` → `src/store.ts:3332`); there is no field projection between store and
  client (grepped for a `spawnAborted`/`summaryCode` re-mapping outside store/review/types and found
  none), so `planDrift`/`planDriftNote` reach the UI without further wiring. `DeliveryStats` gains
  its two fields on both sides — `src/types.ts:1039-1044` and `ui/src/lib/types.ts:1483-1484`.
- *i18n parity.* All eight new keys exist in **both** catalogs at identical positions:
  `usage_delivery_plan_drift` (`ui/messages/en.json:3701`, `ui/messages/de.json:3701`),
  `usage_delivery_plan_drift_major` (`:3707` / `:3707`, `{n}` placeholder present in both),
  `feat_plan_drift_title`/`_body` (`:3725`/`:3726` in both),
  `gloss_plan_drift_term`/`_def` (`:3727`/`:3728` in both),
  `gitrail_plan_drift_minor`/`_major` (`:3729`/`:3730` in both). Every key referenced from code
  resolves: `m.usage_delivery_plan_drift()` (`ui/src/lib/components/usage/DeliveryLens.svelte:73`),
  `m.usage_delivery_plan_drift_major({n})` (`:137`), `m.gitrail_plan_drift_major/minor`
  (`ui/src/lib/components/git-rail/PlanDriftNote.svelte:19-20`).
- *Glossary integrity.* The `[[plan-drift|…]]` marker in both catalogs resolves to the registry entry
  at `ui/src/lib/glossary.ts:85-91` (`kind: "internal"`, so `scripts/check-glossary.mjs:150-166`
  needs no Wikipedia slugs), and its `gloss_plan_drift_term`/`_def` keys exist in both locales as
  above — the three assertions that gate at `scripts/check-glossary.mjs:105-116` and `:127-146`.
  The label is rendered through `GlossaryText` (`DeliveryLens.svelte:128`), so the marker is not
  shown literally.
- *Docs manifest.* `ui/src/lib/docs-manifest.ts:84` inserts `plan drift` between `first-pass rate`
  and `inferred`, which is exactly what the generator would emit: `keywordsOf` joins the description
  with H2/H3 headings in document order and lowercases
  (`ui/scripts/gen-docs-manifest.ts:77-84`), and the new `### Plan drift` section sits between
  `### First-pass rate` and `### Inferred` in
  `docs-site/src/content/docs/reference/glossary.md:67`.
- *Feature announcement.* `sinceVersion: "1.48.0"` is strictly greater than the released `1.47.0`
  (`package.json:3`), and the filename `v1.48.0-plan-drift.ts` matches the field — the two things
  `scripts/check-announcement-versions.mjs` asserts. `targetId: "usage-delivery-tab"` is a
  registered coach target (`ui/src/lib/components/Usage.svelte:275`). Entries are auto-globbed
  (`ui/src/lib/feature-announcements.ts:47-50`), so no index edit is needed, and the id `plan-drift`
  is unique within the announcement registry.
- *Popover rendering.* `PlanDriftNote` is placed inside the `{#if showReview && verdict}` block
  (`ui/src/lib/components/GitRail.svelte:793`, component at `:830`), so its `verdict: ReviewVerdict`
  prop is non-null-narrowed. It interpolates the model-authored note as text
  (`PlanDriftNote.svelte:29`), not `{@html}`, so the untrusted critic prose is escaped — unlike the
  sibling `.rv-body`, which needs DOMPurify.
- *Non-blocking guarantee.* Nothing outside the persistence/metrics path reads `planDrift`: grepping
  the field turns up only `types.ts`, `critic-core.ts`, `review.ts`, `store.ts`,
  `delivery-metrics.ts`, the report script, and the UI mirrors — no reference in `automerge.ts`,
  `blocked.ts`, `learnings-lifecycle.ts` or `rundown-core.ts`, the four other `ReviewVerdict`
  consumers.

The PR satisfies the issue: a non-blocking `planDrift`/`planDriftNote` from the critic, persisted,
surfaced in the Delivery lens and the CLI report, populated only for plan-gated sessions. The
documented deviation (a second copy on `reviewer_spawns`) is necessary and correct —
`archive()` drops the `reviews` row while every task the lens counts is archived.

## Stated limitations

- I did not run tests or lint (CI's job, and the review is read-only). Assertions above are from
  reading the code and the ground truth cited, not from execution.
- Whether the critic model actually honours the "measurement, never judgement" wording is not
  verifiable from the repo — the anchoring risk the plan names is real but unmeasurable here. The
  code-side guarantee (drift is read by no decision path) I did verify, as above.

## Nits (non-blocking):

- `src/review.ts:1499` normalizes and stores `planDriftNote` whenever `drift` is truthy, which includes `"none"` — the prompt tells the critic to omit the note in that case (`src/critic-core.ts:744`), and the popover only renders for `minor|major` (`PlanDriftNote.svelte:28`), so a `"none"` note is dead data; `drift && drift !== "none"` would match the stated contract.
- `src/critic-core.ts:744` and `:757` hardcode "140" in the prompt prose while `PLAN_DRIFT_NOTE_MAX` (`src/critic-core.ts:783`) is the enforcing constant — a future change to the constant would silently desync the instruction from the cap.

- `src/store.ts:4204-4208` casts `SELECT *` straight to `ReviewerSpawnRow`, so `reviewer_spawns.planDrift` never passes through `coercePlanDrift` the way the `reviews` copy does (`src/store.ts:3239`); an out-of-union value in that column would be counted as drift by `statsFor` (`src/delivery-metrics.ts:158`, `!== "none"`). Dormant today because the sole writer is the typed `setReviewerSpawnOutcome`, and it mirrors how the pre-existing `outcome` column is handled.

- `ui/src/lib/components/git-rail/PlanDriftNote.svelte` extracts a two-line render into its own component where the plan specified "a single non-blocking line under `.rv-summary`" in `GitRail.svelte`; the file header justifies it by the Tier-1 Svelte template complexity bar (`.fallowrc.jsonc:250-258`), which is a legitimate author's call.

scoop (commented):
# Review — non-blocking `planDrift` from the critic, fed to the delivery lens (#2155)

**Decision: comment.** No blocking findings. The change does what the issue asks (a non-blocking
`planDrift` + one-line note from the critic, persisted, surfaced in the Delivery lens), and the
plan's documented deviation — persisting on `reviewer_spawns` as well as `reviews` — is not just
justified but necessary, since `archive()` drops the `reviews` row and every task the lens counts
is archived.

## What I verified

**Prompt is emitted exactly when a plan is shown, and nowhere else.**
`reviewPrompt` gates the plan block on `if (opts.plan && opts.plan.trim())` (`src/critic-core.ts:259`)
and passes `planShown: Boolean(opts.plan?.trim())` (`src/critic-core.ts:332`) — the same predicate,
so the two can't disagree. The drift block and the widened JSON shape line are both behind
`opts.planShown` (`src/critic-core.ts:910-923`, `src/critic-core.ts:925-928`), and
`prReviewPrompt` never sets the option (it passes `{ scopeCreep: false, … }` — no `planShown` key,
`src/critic-core.ts:345+`), so the standalone-critic tail stays byte-identical.

**The deconfliction line points at a block that is actually there.** The drift block sits after
`FINDINGS ROUTING` (`src/critic-core.ts:889-899`) and references "the SCOPE-CREEP lens above"
(`src/critic-core.ts:917`); that lens is emitted at `src/critic-core.ts:813-836` behind
`opts.scopeCreep`, which `reviewPrompt` always sets true (`src/critic-core.ts:325`). `planShown`
can therefore never be true while the referenced section is absent.

**Plan clamping does not create a phantom `planShown`.** `fitCriticPrompt` (`src/review.ts:2019-2049`)
only ever head-tail-clamps the plan against a `minUseful` floor — it never drops it to `null` — so
a `planShown: true` in-flight record (`src/review.ts:699`) always corresponds to a prompt that
really carries an `APPROVED PLAN` block. (Even if it didn't, the direction is safe: no block ⇒ no
answer ⇒ `normalizePlanDrift` returns `null`.)

**Guards in `buildVerdict`.** `src/review.ts:1543-1544` forces `null` when `f.planShown` is false or
the resolved decision is `error`; the note is only normalized when a level survived
(`src/review.ts:1576`). `buildVerdictCore` is untouched, so the standalone PR critic (which has no
plan) is unaffected — confirmed by grepping: `normalizePlanDrift` has exactly two call sites,
`src/review.ts:1544` and the tests.

**Nothing in the review loop reads it.** `planDrift` appears in `src/review.ts` at only four
places — 1279 (spawn stamp), 1544, 1575, 1576 — none of which feed `decision`, `findings`,
`streakReviews`, `addressRound` or the merge train. Grepping `planDrift` across `src/` returns only
types, store, critic-core, delivery-metrics and those four lines.

**Store round-trip and SQL arity.** The `reviews` INSERT lists 22 columns
(`src/store.ts:3260-3261`), `VALUES` carries 22 placeholders (`src/store.ts:3262` — the diff grew
both the old 20-column list and the old 20-placeholder list by exactly two), and the bind array has
22 entries ending `planDrift, planDriftNote, v.updatedAt` (`src/store.ts:3273-3296`). The
`ON CONFLICT` clause updates both new columns (`src/store.ts:3272`). Both `getReview`
(`src/store.ts:3248`) and `snapshotReviews` (`src/store.ts:3336`) select them, and `hydrateReview`
validates `planDrift` through `coercePlanDrift` (`src/store.ts:370`, used at `src/store.ts:3239`)
rather than letting the `...r` spread pass the raw column through.

**Migrations cover both the fresh and the upgrade path.** `REVIEWER_SPAWNS_DDL` carries
`planDrift TEXT` (`src/store.ts:1167`) *and* `migrateReviewerSpawnColumns` adds it
(`src/store.ts:4787`) — the #1816 rule. Same double for `reviews`: DDL at `src/store.ts:1320`,
migration at `src/store.ts:4728-4729`.

**Every reader of the new column sees it.** `listReviewerSpawns()` is `SELECT *`
(`src/store.ts:4206`) and `scripts/delivery-report.ts:63` is likewise `SELECT *`, so both pick the
column up without a column-list edit.

**Signature change is compatible with its other caller.** `setReviewerSpawnOutcome` gained a
defaulted third parameter (`src/store.ts:4228-4232`). The only other production caller is the plan
gate (`src/plan-gate.ts:1338-1345`), which passes two arguments and therefore writes `NULL` into a
`plan_gate` row — a row `applyReviewSpawn` never touches (`src/delivery-metrics.ts:110-113` routes
`plan_gate` rows to the `planRounds` branch instead). No other call sites exist in `src/`.

**Aggregation.** `applyReviewSpawn` takes last-non-null (`src/delivery-metrics.ts:97`) over rows
that `listReviewerSpawns()` returns `ORDER BY spawnedAt` (`src/store.ts:4206`), so the round closest
to merge wins and a later errored round can't erase it. `statsFor` builds the denominator from
tasks carrying a non-null drift (`src/delivery-metrics.ts:158`) and `rate()` returns
`{value: null, n: 0}` when that set is empty (`src/delivery-metrics.ts:85-87`) — never `0%`.

**Per-repo rows get the fields for free**, since `DeliveryRepoRow extends DeliveryStats`
(`src/types.ts:1489-1492`, mirrored in `ui/src/lib/types.ts:1489`), so the new required fields
cannot be missing from a repo row.

**Locale parity — checked key by key, both catalogs, same line numbers.** All eight new keys exist
in both: `usage_delivery_plan_drift` (`ui/messages/en.json:3701` / `ui/messages/de.json:3701`),
`usage_delivery_plan_drift_major` (en 3707 / de 3707, `{n}` placeholder present in both),
`feat_plan_drift_title` (3725/3725), `feat_plan_drift_body` (3726/3726), `gloss_plan_drift_term`
(3727/3727), `gloss_plan_drift_def` (3728/3728), `gitrail_plan_drift_minor` (3729/3729),
`gitrail_plan_drift_major` (3730/3730). The gate at `ui/scripts/check-i18n.mjs:48` compares the key
union, which these satisfy. Every one of the eight is referenced:
`ui/src/lib/components/usage/DeliveryLens.svelte:73` and `:137`,
`ui/src/lib/feature-announcements/entries/v1.48.0-plan-drift.ts:5-6`,
`ui/src/lib/glossary.ts:90-91`, `ui/src/lib/components/git-rail/PlanDriftNote.svelte:19-20`.

**Glossary + docs manifest.** The `plan-drift` term id in `ui/src/lib/glossary.ts:87-92` is the same
id used in the `[[plan-drift|…]]` markers in both catalogs, which is what
`scripts/check-glossary.mjs` resolves. `docs-site/src/content/docs/reference/glossary.md:70` adds
the `### Plan drift` H3 between `### First-pass rate` and `### Inferred`, and
`ui/src/lib/docs-manifest.ts:84` inserts `plan drift` at exactly that position in the keyword
string — consistent with `keywordsOf(description, extractHeadings(body))`
(`ui/scripts/gen-docs-manifest.ts:77,106`), so `check:docs-manifest` should be satisfied.

**Feature announcement.** Entries are auto-globbed (`ui/src/lib/feature-announcements.ts:46-49`), so
no index edit is needed; `sinceVersion: "1.48.0"` matches the other unreleased entries
(`v1.48.0-delivery-metrics.ts`, `v1.48.0-herdr-082.ts`), and `targetId: "usage-delivery-tab"`
resolves to a real coach target at `ui/src/lib/components/Usage.svelte:275`.

**UI rendering.** `PlanDriftNote.svelte` renders only for `minor|major`
(`ui/src/lib/components/git-rail/PlanDriftNote.svelte:29`), interpolates the model-authored note as
text (no `{@html}`), and its `.rv-drift` rule mirrors `.rv-summary`'s
(`ui/src/lib/components/GitRail.svelte:1277-1282` vs `PlanDriftNote.svelte:32-38`), including the
`flex-shrink: 0` the popover's "only `.rv-body` scrolls" layout needs. `var(--fs-meta)` and
`var(--color-muted)` both exist (`ui/src/app.css:86`, `ui/src/app.css:19`). The verdict reaches the
client whole — `events.emit("session:review", { id, review: verdict })` (`src/index.ts:1484`) and
`getReviews()` → `snapshotReviews()` — so no projection strips the new fields.

**Stated limitation:** I did not run tests, lint, `fallow audit`, or the catalog/glossary/manifest
gates — CI owns those. Everything above is static comparison against the cited lines.

## Nits (non-blocking):

- `src/types.ts:755` — the doc says the note is "null when it reported none", but `src/review.ts:1576`
  keeps the note whenever `drift` is truthy and `"none"` is truthy; the "omit it when none" rule is
  enforced only by the prompt (`src/critic-core.ts:914`), not by the code. A `"none"` + note pair
  would be stored and never rendered — dead data, not a defect, but the comment overstates the
  guarantee. `drift && drift !== "none" ? … : null` would make code and comment agree.
- `src/critic-core.ts:927` — the JSON shape line lists `planDriftNote` unconditionally even though
  the block above tells the critic to omit it for `"none"`; the parenthetical inside the placeholder
  carries that, but the shape line alone reads as "always emit both".

- `ui/src/lib/components/git-rail/PlanDriftNote.svelte` — the plan's file table listed the drift line
  as an edit to `GitRail.svelte`, not a new component. The in-file rationale (GitRail's `<template>`
  sits under the global Tier-1 complexity bar at `.fallowrc.jsonc:250-258` with no per-file
  override) is sound and the component is 37 lines with no abstraction beyond what it renders — a
  judgement call the author may legitimately keep.

- `src/store.ts:4204-4207` — `listReviewerSpawns()` casts rows straight to `ReviewerSpawnRow` with no
  `coercePlanDrift`, while the sibling path for the same enum on `reviews` does validate it
  (`src/store.ts:3239`), and that helper's own comment cites "a value written before the union
  changed" as the case it guards. Unreachable today: the sole writer is
  `setReviewerSpawnOutcome(…, planDrift: PlanDrift | null)` (`src/store.ts:4228-4232`) fed from the
  normalized verdict. If a future writer (a backfill, an import, a widened union rolled back) ever
  lands a stray string, `applyReviewSpawn` (`src/delivery-metrics.ts:97`) and `statsFor`
  (`src/delivery-metrics.ts:158,171`) test only `!= null` and `!== "none"`, so it would silently
  count as drift instead of as "not measured".

scoop (commented):
# Review — non-blocking `planDrift` from the critic → Delivery lens (#2155)

Verdict: **comment**. I found no correctness, security, or contract defect. Everything below is
either a verified confirmation, a non-blocking nit, or a dormant-only observation.

**The ask/no-ask switch is exactly the plan-block condition.**
`reviewPrompt` emits the plan block under `if (opts.plan && opts.plan.trim())`
(`src/critic-core.ts:260`) and passes `planShown: Boolean(opts.plan?.trim())` into the tail
(`src/critic-core.ts:331`) — the same predicate, so the drift block can never be asked for without a
plan in the prompt. `prReviewPrompt` builds its tail at `src/critic-core.ts:437` without a
`planShown` key, and `scopeAndOutputTail`'s option is optional (`src/critic-core.ts:764`), so the
standalone-critic tail is unchanged. I compared the plan-less branch of the JSON shape line
(`src/critic-core.ts:937`) character-by-character against the line it replaced in
`git show e40cfab1:src/critic-core.ts` — it is the same string, so the byte-identity assertions in
`test/critic-core.test.ts` stay valid.

**Block placement matches what its own text claims.** The drift block is emitted after
`FINDINGS ROUTING` (`src/critic-core.ts:888` → block at `src/critic-core.ts:912`) and its
DECONFLICTION sentence refers to "the SCOPE-CREEP lens above", which is emitted at
`src/critic-core.ts:821` and only when `opts.scopeCreep` is set — `reviewPrompt` always sets it
(`src/critic-core.ts:327`) and is the only caller that can set `planShown`, so the reference can
never dangle.

**`InFlight.planShown` cannot disagree with the prompt.** There is exactly one construction site for
`InFlight` (`src/review.ts:680`), and it sets `planShown: Boolean(plan?.trim())`
(`src/review.ts:699`) from the same `plan` value handed to `composePrompt`/`fitAndSpawn`
(`src/review.ts:624`, `646`, `673`). `criticPromptComposer` forwards that value straight into
`reviewPrompt`'s `opts.plan` (`src/review.ts:932-941`). The clamp path
(`fitCriticPrompt`, `src/review.ts:2019-2049`) can shorten the plan but not remove it — and even if
it emptied it, the critic would simply never be asked for a level and `normalizePlanDrift(undefined)`
returns `null` (`src/critic-core.ts:1175`), so the failure mode is "not measured", not a fabricated
level.

**The two guards do what the comment says.** `src/review.ts:1543-1544` forces `null` when
`!f.planShown` or when `buildVerdictCore` resolved `decision === "error"`; `raw` may be `null` on the
no-verdict path and `raw?.planDrift` handles it.

**Non-blocking is structural, not just asserted in prose.** `planDrift` is written in
`buildVerdict` (`src/review.ts:1573-1574`) and in the spawn stamp (`src/review.ts:1277`), and read
only by `src/delivery-metrics.ts:95`, `scripts/delivery-report.ts:115`,
`ui/.../DeliveryLens.svelte:73` and `ui/.../PlanDriftNote.svelte:16`. I grepped `planDrift` across
`src/` and `ui/src/` — nothing in the decision resolution, streak/round accounting, auto-address
steer, or merge-train path reads it.

**Store round-trip is arithmetically correct.** `putReview` now names 22 columns
(`src/store.ts:3260-3261`), binds 22 placeholders (`src/store.ts:3262`) and passes 22 values
(`src/store.ts:3274-3295`); the pre-change version named 20 of each (`git show
e40cfab1:src/store.ts`). Both SELECT lists were extended to match (`src/store.ts:3248`,
`src/store.ts:3336`) and `ReviewVerdictRow` carries the two new fields (`src/store.ts:686-687`).
Fresh-install DDL carries the columns (`src/store.ts:1320` for `reviews`, `src/store.ts:1167` for
`reviewer_spawns`) *and* both migrations add them idempotently (`src/store.ts:4726-4729`,
`src/store.ts:4787`) — the #1816 rule the plan cites.

**Signature change has no stale caller.** `setReviewerSpawnOutcome` gained a defaulted third
parameter (`src/store.ts:4228-4232`). Callers: `src/review.ts:1270` (3 args) and
`src/plan-gate.ts:1338` (2 args, so the plan-gate row keeps `planDrift` NULL — asserted at
`test/store.test.ts:665`). Both `Pick<>` dep types already listed the method
(`src/review.ts:274`, `src/plan-gate.ts:481`), so no interface widening was needed.

**Aggregation ordering assumption holds.** `applyReviewSpawn`'s "last non-null wins"
(`src/delivery-metrics.ts:95-97`) depends on spawn order; `buildDeliveryMetrics` feeds it
`store.listReviewerSpawns()` (`src/delivery-metrics.ts:271`), which is
`SELECT * FROM reviewer_spawns ORDER BY spawnedAt` (`src/store.ts:4205-4207`). `SELECT *` also means
the new column reaches `ReviewerSpawnRow` without a column-list edit. Empty denominators return
`{ value: null, n: 0 }` via `rate()` (`src/delivery-metrics.ts:85-87`), never `0`. Per-repo rows get
the field for free because `DeliveryRepoRow extends DeliveryStats` (`src/types.ts:1052`) and
`repoRows` spreads `statsFor` (`src/delivery-metrics.ts:191`).

**API surface needs no serializer change.** Nothing whitelists `DeliveryStats` fields — the payload
is `statsFor`'s object (`src/delivery-metrics.ts:278`). Verdicts reach the UI as whole objects
(`src/index.ts:1484`), so `planDrift`/`planDriftNote` arrive at the popover.

**i18n parity.** All eight new keys exist in both catalogs at the same positions —
`ui/messages/en.json:3735-3742` and `ui/messages/de.json:3735-3742` — with matching `{n}`
placeholders in `usage_delivery_plan_drift_major`. That satisfies the union-parity gate at
`ui/scripts/check-i18n.mjs:45-53`. No key is duplicated elsewhere in either file.

**Glossary integrity.** `[[plan-drift|Plan drift]]` resolves to the term registered at
`ui/src/lib/glossary.ts:87-92`, whose `gloss_plan_drift_term`/`gloss_plan_drift_def` exist in both
catalogs — the two things `scripts/check-glossary.mjs:105-116` enforces. Tile labels go through
`<GlossaryText>` (`ui/src/lib/components/usage/DeliveryLens.svelte:121-123`), so the marker renders
rather than leaking as literal markup. The docs-site heading (`docs-site/.../glossary.md:70`) sits
between "First-pass rate" and "Inferred", matching the keyword order in the regenerated manifest
(`ui/src/lib/docs-manifest.ts:84`).

**Announcement entry.** `sinceVersion: "1.48.0"` is greater than the last release in
`package.json:3` (1.47.0) and matches the filename, the two things
`scripts/check-announcement-versions.mjs` checks; the entry is auto-registered by the `import.meta.glob`
at `ui/src/lib/feature-announcements.ts:47`. `targetId: "usage-delivery-tab"` exists at
`ui/src/lib/components/Usage.svelte:275`. Sharing that target with
`ui/src/lib/feature-announcements/entries/v1.48.0-delivery-metrics.ts:8` is established precedent
(four v1.34.0 entries share `session-recap`).

**UI safety and layout.** `PlanDriftNote` renders the model-authored note as a text node (Svelte
escapes it) — no `{@html}`, unlike `.rv-body` at `ui/src/lib/components/GitRail.svelte:833`. Its
`flex-shrink: 0` mirrors `.rv-summary` (`ui/src/lib/components/GitRail.svelte:1277-1283`), so the
popover's "only `.rv-body` scrolls" invariant is preserved, and both custom properties it uses exist
(`ui/src/app.css:19`, `ui/src/app.css:86`). The `{#if drift && drift !== "none"}` guard means
`none`/absent renders nothing.

**Satisfies the task.** Non-blocking `planDrift: none|minor|major` + a one-line note, asked only when
the plan gate produced a plan, persisted, and surfaced in the Delivery lens. The one deviation from
the issue text — persisting on `reviewer_spawns` as well as `reviews` — is required and correct:
`archive()` drops the `reviews` row while every task the lens counts is archived, so a `reviews`-only
copy would always be `null` in the lens.

## Stated limitations

- I did not run tests, lint, `check:docs-manifest`, `check-glossary`, or the fallow audit — CI owns
  those. I verified the *inputs* those gates read, not their exit codes.
- I could not confirm `ui/src/lib/docs-manifest.ts` was produced by the generator rather than
  hand-edited; the single-line keyword change is consistent with the new heading's position, which is
  all the file's content lets me check.
- Whether the critic model actually stays non-judgemental when asked the drift question is a
  behavioural property of the model, not something this repo can settle; the prompt wording and the
  structural fact that nothing reads the field are the mitigations, and both are in place.

## Nits (non-blocking):

- `ui/src/lib/components/git-rail/PlanDriftNote.svelte:22-24`: the catalog labels are full sentences ending in a period (`ui/messages/en.json:3741`, `ui/messages/de.json:3742`), so a note appends after it — "…with incidental departures. — built a service, plan said helper". The plan's sketched shape ("Plan drift: minor — <note>") reads better; consider trimming the label or dropping its trailing period when a note is present.
- `src/review.ts:1574`: `drift ? normalizePlanDriftNote(...) : null` keeps a note when drift is `"none"` (a truthy string), even though the prompt asks the critic to omit it there; the note is then stored but never rendered (`PlanDriftNote.svelte:26` filters `none`). Harmless dead data, but `drift && drift !== "none"` would match the stated contract.
- `src/critic-core.ts:1183`: the 140-char cap slices mid-word; an ellipsis or a word-boundary trim would read better in the one-line slot.

- `src/store.ts:4205-4207` casts `SELECT *` straight to `ReviewerSpawnRow`, so an unrecognized string in `reviewer_spawns.planDrift` would be counted as drift by `src/delivery-metrics.ts:95` and `src/delivery-metrics.ts:157` — whereas the sibling `reviews` read path launders the same value through `coercePlanDrift` (`src/store.ts:3241`). Unreachable today (the only writer is the typed `setReviewerSpawnOutcome`), and the pre-existing `outcome` column has the same shape, but the guard asymmetry is real if a second writer ever appears.

Issue #2165, opened 2026-09-02
Machine-readable severity for critic findings: Important/Nit, named passes, nit cap


scoop:
Follow-up to #2154 (R2 of the AI-native SDLC playbook gap analysis, `docs/research/ai-native-sdlc-playbook.md` §3 R2). Slices 1 and 2 — per-repo `REVIEW.md` and house-rules injection into the reviewer — shipped in the PR that closes it. This is the remaining slice 3.

## What

Move the critic's verdict from blocking-vs-prose to the playbook's **Important/Nit** distinction with **named passes** (bug / security / compliance / scope), plus a **nit cap**.

Today `findings[]` is blocking-only and everything else is routed into prose sections of `.shepherd-review.md` (Nits, Scope creep, Possible smells, Latent, and — since #2154 — House rules). That routing exists for a good reason: any entry in `findings` advances the streak (`buildVerdict`), is auto-addressed (`runAutoAddress` fires on non-empty findings regardless of decision) and is re-raised every round, so a "non-blocking" item placed there loops the PR until the streak ceiling pauses it. Severity in the verdict is what would let nits live in structured output *without* that consequence.

Unlocks: severity-aware steering (only Important findings block the merge train), a better UI than prose sections, and defect-class metrics in the Delivery lens (#2151 R1 / #2153).

## Why it is its own issue

It changes the verdict shape, so the blast radius is much wider than slices 1–2 (which were prompt composition only):

- `RawVerdict` / `normalizeFindings` / `buildVerdictCore` / `scopeBackstop` / `scopeFindings` in `src/critic-core.ts` — the scope backstop splits a finding on the first `": "` and treats a token containing `/` as a path, so a structured finding changes what it parses.
- The `reviews` / `pr_reviews` `findings` column (`src/store.ts`) and `parseFindings`, plus a migration for existing rows.
- Auto-address (`runAutoAddress`), the streak counters, and the re-raise rule in `reviewPrompt`.
- Merge-train gating on `changes_requested`.
- The UI review surfaces and their EN+DE catalogs.

## Constraint that shaped slices 1–2, and still applies

[...]

PR #2166, opened 2026-09-02
feat(critic): per-repo REVIEW.md + house rules in the reviewer

### LINKED ISSUE #2154: Version-controlled review policy: per-repo REVIEW.md, house rules for the critic, severity taxonomy
Follow-up to #2151 (R2 of the AI-native SDLC playbook gap analysis, `docs/research/ai-native-sdlc-playbook.md` §3 R2). R1 shipped in #2153.

Today the critic's review policy is three booleans in code. The playbook's move is a version-controlled, diffable policy. Three separable slices:

1. **Per-repo `REVIEW.md`.** Let the critic read a `REVIEW.md` (or `.shepherd/review.md`) from the target repo — extra passes to run, known exclusions, repo-specific severity rules — appended to the built-in prompt the way house rules are appended to task prompts. The built-in lenses stay the floor; the file adds repo policy.
2. **Inject house rules into the reviewer spawn.** The critic currently cannot flag violations of the very rules Shepherd injects into the author — the injection in `src/service.ts` (`composeDirectives` path) never reaches the reviewer spawn. A rule the author ignored is exactly what review should catch.
3. **Machine-readable severity.** Move from blocking-vs-prose to Important/Nit with named passes (bug/security/compliance/scope) in the verdict JSON, plus a nit cap. Unlocks severity-aware steering (only Important findings block the merge train), better UI, and defect-class metrics in the Delivery lens. Keep the current sections as the presentation layer.

Note for slice 3: `.shepherd-review.json` is hand-authored by the critic and the body already had to be moved out to a sidecar file (#2042) because escaping broke verdicts. Any new field must stay short and structured for the same reason.

scoop:
Closes #2154 (slices 1 and 2). Slice 3 — machine-readable severity — is filed as #2165.

The critic's review policy was three booleans in code, and the house rules Shepherd injects into the agent that *writes* a diff never reached the agent that *reviews* it. Two optional blocks now ride the critic prompt. Both are default-absent, so a prompt with neither is byte-identical to `main` — asserted by test.

## 1. Per-repo `REVIEW.md`

`REVIEW.md` at the repo root (else `.shepherd/review.md`), read **from the pull request's base commit** via `git show <baseSha>:<path>` — never from the checked-out PR head.

That provenance is the whole design. Everything the critic reads out of a PR is fenced as untrusted data it is instructed never to obey; a fenced policy would therefore be contractually ignorable, i.e. inert. Reading the base object is what lets the policy ride **unfenced as trusted instruction** while still guaranteeing a branch cannot rewrite the rules it is about to be judged by — which matters most for a fork PR at the standalone critic. The accepted cost: the PR that adds `REVIEW.md` is not reviewed under it.

The built-in contract stays the floor, stated in-band rather than trusted to placement: the policy may add passes, emphasis and repo severity guidance, and may declare exclusions that narrow which **areas or classes** get reviewed — but an exclusion may never suppress a correctness or security defect the critic actually verified, and never touches the verdict-output contract the server-side parser and scope backstop depend on. Hard-capped at 8 KB with a visible clamp marker.

## 2. House rules into the reviewer

The repo's curated standing rules — the body Shepherd puts in the system prompt of every agent it runs there — now also reach the critic, as the repository's standard. Scoped by the diff's **actual changed files**, which is strictly better input than the author side gets (it has to scrape the task text before the session runs). Gated by the existing per-repo `learningsEnabled`.

The set is **planned fresh for the review**, not replayed: current active rules, diff-based scoping, recency decay evaluated now. So the prompt does not claim it reproduces what any agent saw. It also reaches the standalone critic, which sweeps third-party and fork PRs where no Shepherd agent authored anything — so the author-facing framing ("a rule the author was handed and ignored") is gated to the session critic, and the license to block rests on the rules being the repo's standard, which holds for any PR against it.

Routing splits on how clear-cut the violation is, for the reason #1948 established: any entry in `findings` advances the streak, is auto-addressed, and is re-raised every round, so a stale distilled rule placed there would loop the PR until the streak ceiling paused it. A **clear, unambiguous** violation is a finding; anything softer goes to a non-blocking `House rules (non-blocking):` body section.

No `recordInjectedLearnings` rows are written for a reviewer spawn: `injectedCount` is the denominator of the Wilson lower bound that auto-retires a rule, and only an author session ever pays the matching reward — counting reviewer pulls would retire rules that are doing their job.

## Notes for review

- Both critics get both blocks (session + standalone) — policy is a property of the repo, not of who opened the PR.
- **No new toggles and no new env vars.** File presence is the switch; house rules reuse `learningsEnabled`. The issue's complaint is "policy is three booleans"; a fourth would miss the point.
- The policy read is `await`ed, so it sits with the issue-body / epic-delta fetches **before** the `starting` tombstone re-check in `begin()`, which has to remain the last await-gated step before the spawn.
- It joins the argv clamp ladder **after** the plan (session critic) / PR body (standalone) — those give way first. A test asserts that ordering under real argv pressure.
- The house-rules budget default moved to `house-rules.ts` so the critic can fall back to it without importing `./config` (which reads the forge map off disk at import).

## Verification

`bun run lint`, `bunx tsc --noEmit`, `bun run test` (8873 pass), `ui/ check` + `ui/ test` (4735 pass), `bunx fallow audit --base origin/main --fail-on-issues` (green), and the docs site builds.

### DIFF
--- a/docs-site/astro.config.mjs
+++ b/docs-site/astro.config.mjs
@@ -128,6 +128,7 @@ export default defineConfig({
           items: [
             { label: "Configuration", slug: "reference/configuration" },
             { label: "Stacked epic children", slug: "reference/stacked-epic-children" },
+            { label: "Review policy (REVIEW.md)", slug: "reference/review-policy" },
             { label: "Concepts & glossary", slug: "reference/glossary" },
             { label: "Keyboard shortcuts", slug: "reference/keyboard-shortcuts" },
             { label: "Plugins", slug: "reference/plugins" },
--- a/docs-site/src/content/docs/reference/glossary.md
+++ b/docs-site/src/content/docs/reference/glossary.md
@@ -46,7 +46,10 @@ its resume path would otherwise target a sibling session in the shared directory
 ### Critic
 
 Shepherd's isolated, read-only review agent that inspects a PR's diff once CI is
-green and posts a verdict.
+green and posts a verdict. Its built-in lenses are the floor; a repository can
+add passes and known exclusions with a committed `REVIEW.md`, and the repo's
+house rules are shown to the critic alongside them — see
+[Review policy](/reference/review-policy/).
 
 ### Merge train
 
--- a/docs-site/src/content/docs/reference/review-policy.md
+++ b/docs-site/src/content/docs/reference/review-policy.md
@@ -1,0 +1,89 @@
+---
+title: Review policy (REVIEW.md)
+description: How a repository steers Shepherd's PR critic with a version-controlled REVIEW.md, why the file is read from the base commit rather than the pull request, and how the repo's own house rules now reach the reviewer.
+---
+
+Shepherd's PR critic runs a fixed set of built-in review lenses. A repository can **add** to them
+with a version-controlled policy file, and the house rules Shepherd already injects into the agent
+that *writes* the code are now also shown to the agent that *reviews* it.
+
+Neither is a setting. There is nothing to switch on: commit the file, and the next review reads it.
+
+## `REVIEW.md`
+
+Put the policy at **`REVIEW.md`** in the repository root, or at **`.shepherd/review.md`** if you
+would rather not spend a root-level file. If both exist, `REVIEW.md` wins.
+
+It is ordinary Markdown — extra passes you want run, areas you do not want reviewed, severity
+guidance specific to this codebase:
+
+```markdown
+# Review policy
+
+## Extra passes
+
+- **Migrations.** Any change under `db/migrations/` gets a rollback pass: does the down-migration
+  actually reverse the up-migration, and is it safe to run against a populated table?
+- **Public API.** A change to `src/api/**` is a compatibility surface — call out anything a
+  published client would notice.
+
+## Known exclusions
+
+- `src/generated/**` is code-generated. Do not review its style or structure; review the generator.
+- We test the CLI by hand. A missing test for `bin/**` is not a finding here.
+```
+
+Both critics read it: the session critic that reviews a Shepherd task's own PR, and the standalone
+critic that reviews every open PR in a repository with **Review all PRs** on. The policy belongs to
+the *repository*, not to who opened the pull request.
+
+## It is read from the base commit
+
+The critic reads the policy from the **base commit of the pull request** — the merge target — not
+from the branch it is reviewing.
+
+That is deliberate, and it is the reason the policy can steer a review at all. Everything the critic
+reads out of a pull request (the diff, the description, the plan, review comments) is fenced as
+untrusted data that it is instructed never to obey; otherwise a pull request could talk its way past
+its own review. `REVIEW.md` is the exception: it is handed over as genuine instruction. What makes
+that safe is that the text can only come from work the repository has **already reviewed and
+merged** — a branch cannot rewrite the rules it is about to be judged by, which matters most for a
+pull request from a fork.
+
+The consequence to know about: **the pull request that adds or edits `REVIEW.md` is not reviewed
+under it.** The new policy takes effect once it lands on the base branch.
+
+## The built-in lenses stay the floor
+
+The policy may only add. Shepherd's own contract — the scope rules, the verification discipline,
+findings routing, and the verdict format — is not negotiable by a file in the repository.
+
+Concretely:
+
+- The policy **may** add review passes, add emphasis, and state repo-specific severity guidance.
+- It **may** declare known exclusions that narrow which areas or which classes of issue the critic
+  spends attention on.
+- An exclusion **may never** suppress a correctness or security defect the critic actually verified
+  in the diff.
+- Anything raised under the policy is routed like any other point: blocking problems become
+  findings, everything else becomes a non-blocking note. The policy adds passes, not a new kind of
+  output.
+
+The file is capped at about 8 KB. A longer one is truncated with a visible marker rather than
+silently cut — but a review policy that runs past a page is probably documentation in the wrong
+place.
+
+## House rules reach the reviewer
+
+Shepherd injects a repository's curated **house rules** into the system prompt of every agent that
+works in it. Those same rules are now shown to the critic, so a rule the author was handed and
+ignored is something review can actually catch. They are scoped to the files the pull request
+touches, and they follow the repository's existing **Learnings** setting — with learnings off, no
+rules are injected anywhere.
+
+House rules are distilled from past sessions, so they are treated as guidance rather than law: a
+**clear, unambiguous** violation is a blocking finding; anything softer is reported as a
+non-blocking note. A rule that has gone stale can slow a review down; it cannot deadlock one.
+
+Reviewer-side injection does not count toward a rule's usage statistics — the helpfulness record
+that retires an underperforming rule is still kept by the sessions that *write* code.
--- a/docs-site/src/content/docs/reference/review-policy.md
+++ b/docs-site/src/content/docs/reference/review-policy.md
@@ -76,10 +76,17 @@ place.
 ## House rules reach the reviewer
 
 Shepherd injects a repository's curated **house rules** into the system prompt of every agent that
-works in it. Those same rules are now shown to the critic, so a rule the author was handed and
-ignored is something review can actually catch. They are scoped to the files the pull request
-touches, and they follow the repository's existing **Learnings** setting — with learnings off, no
-rules are injected anywhere.
+works in it. The critic is now shown that same body of rules as the repository's standard, so it can
+judge a pull request against the conventions the repo actually holds itself to. They are scoped to
+the files the pull request touches, and they follow the repository's existing **Learnings**
+setting — with learnings off, no rules are injected anywhere.
+
+The set the critic sees is **planned fresh for that review**, from the rules active at the time and
+the files the diff touches. It is not a replay of what any particular agent was given, and it is not
+limited to pull requests Shepherd wrote: a third-party or fork pull request is reviewed against the
+same repository standard. Where Shepherd's own session critic reviews a PR one of its agents wrote,
+the critic is additionally told so — a rule the author was handed and ignored is worth catching —
+but the rules apply either way.
 
 House rules are distilled from past sessions, so they are treated as guidance rather than law: a
 **clear, unambiguous** violation is a blocking finding; anything softer is reported as a
--- a/docs/sandbox-security.md
+++ b/docs/sandbox-security.md
@@ -109,6 +109,21 @@ execution controls below, Shepherd bounds the injection surface at ingestion
   (`composeSystemPromptBlocks`), and each aux prompt builder emits it itself. That
   invariant is pinned by `test/untrusted.test.ts`: a builder that fences without the
   directive fails there.
+- **Trusted-by-provenance exception: the repo review policy.** The PR critic's
+  prompt carries one block _unfenced_, as genuine instruction: the repo's
+  `REVIEW.md` (else `.shepherd/review.md`), read with `git show` from the PR's
+  **base commit**, never from the checked-out PR head
+  (`defaultReadReviewPolicy`, `src/critic-core.ts`). Fenced it would be
+  contractually ignorable and therefore inert; unfenced text the PR could author
+  would let a branch (notably a fork PR at the standalone critic) rewrite the
+  rules it is judged by. Base provenance closes that — only already-merged policy
+  is honored, at the cost that the PR _introducing_ `REVIEW.md` is not reviewed
+  under it. The built-in contract stays the floor: the block may add passes and
+  narrow which areas/classes get attention, never suppress a verified defect or
+  touch the verdict-output contract. Text is clamped at
+  `REVIEW_POLICY_MAX_BYTES` (8 KB) with a visible marker, a bad SHA or git
+  failure yields no block, and the delimiters are plain markers rather than a
+  nonce fence because the content is trusted by construction.
 - **Fail-closed author-trust gate.** An **autonomous** (`auto=true`) spawn from an
   issue whose author is **not** a trusted repo association (`OWNER` / `MEMBER` /
   `COLLABORATOR` — anything else, including an unresolvable, absent, or Gitea-side
--- a/src/config.ts
+++ b/src/config.ts
@@ -15,6 +15,7 @@ import { normalizeTelemetryConsent } from "./telemetry-consent";
 import { normalizeOperatorLanguage } from "./operator-language";
 import { type SandboxProfile, isSandboxProfile } from "./sandbox";
 import { applyHerdrSocket } from "./herdr-session";
+import { HOUSE_RULES_DEFAULT_BUDGET_CHARS } from "./house-rules";
 
 const dbPath = process.env.SHEPHERD_DB ?? `${process.env.HOME}/.shepherd/shepherd.db`;
 // forge map sits next to the db by default; SHEPHERD_FORGES overrides the path.
@@ -800,9 +801,12 @@ export const config = {
   namerEffort: normalizeDefaultEffortSetting(process.env.SHEPHERD_NAMER_EFFORT) ?? "low",
   // Char budget for the Shepherd house-rules block prepended to every agent prompt. Active+
   // promoted rules fill greedily by most-recently-effective priority until this cap; the rest
-  // stay visible-but-uninjected in the Learnings drawer for the operator to prune. Default 4000
-  // (~25 max-length rules); only an unusually large curated set is capped.
-  houseRulesBudgetChars: Number(process.env.SHEPHERD_HOUSE_RULES_BUDGET_CHARS ?? 4000),
+  // stay visible-but-uninjected in the Learnings drawer for the operator to prune. The default
+  // lives in house-rules.ts beside the planner it bounds (the reviewer-side injection falls back
+  // to it without importing this module); only an unusually large curated set is capped.
+  houseRulesBudgetChars: Number(
+    process.env.SHEPHERD_HOUSE_RULES_BUDGET_CHARS ?? HOUSE_RULES_DEFAULT_BUDGET_CHARS,
+  ),
   // Max auto-steers autopilot spends per session before it pauses for the operator (runaway guard).
   autopilotStepCap: Number(process.env.SHEPHERD_AUTOPILOT_STEP_CAP ?? 10),
   // Per-role ENVIRONMENT (CLI + model + effort) for the transient autopilot stop-classifier spawn (cheap +
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -224,7 +224,7 @@ export function reviewPrompt(
     planClamped?: boolean;
     /** #2154: the repo's `REVIEW.md` as committed on the base commit, or null/absent for none. */
     reviewPolicy?: string | null;
-    /** #2154: the rendered `<shepherd-house-rules>` block the AUTHOR of this diff was given. */
+    /** #2154: the rendered `<shepherd-house-rules>` block of the repo's standing rules. */
     houseRules?: string | null;
   } = {},
 ): string {
@@ -321,9 +321,13 @@ export function reviewPrompt(
         scopeCreep: true,
         smellLens: opts.smellLens,
         planClamped,
-        // #2154: repo policy + the author's own house rules. Both absent ⇒ tail unchanged.
+        // #2154: repo policy + the repo's standing house rules. Both absent ⇒ tail unchanged.
+        // `houseRulesAuthored`: this critic reviews a SHEPHERD SESSION's own PR, so an agent really
+        // did write the diff under these rules. prReviewPrompt never sets it — see
+        // reviewerHouseRulesBlock for why that distinction is load-bearing.
         reviewPolicy: opts.reviewPolicy,
         houseRules: opts.houseRules,
+        houseRulesAuthored: true,
       },
     ),
   );
@@ -372,7 +376,9 @@ export function prReviewPrompt(
       "Judge the diff ONLY for bugs, security issues, and clear quality problems. Use the stated intent above to understand what the change is for — do NOT raise a finding merely because the diff seems incomplete versus that intent. Tests and lint are handled by CI — do not run them.",
       epic,
       // #2154: the standalone critic gets the same repo policy + house rules as the session critic
-      // — the policy is a property of the REPO, not of who opened the PR. Both absent ⇒ unchanged.
+      // — both are a property of the REPO, not of who opened the PR. `houseRulesAuthored` is
+      // deliberately NOT set: this critic sweeps third-party and fork PRs, where no Shepherd agent
+      // wrote the diff and nothing was injected into its author. Both absent ⇒ unchanged.
       { landing, reviewPolicy: opts.reviewPolicy, houseRules: opts.houseRules },
     ),
   );
@@ -693,12 +699,23 @@ function reviewPolicyBlock(policy: string | null | undefined): string[] {
 }
 
 /**
- * The REPO HOUSE RULES block (#2154 slice 2) — the SAME `<shepherd-house-rules>` block Shepherd
- * injected into the system prompt of the agent that wrote this diff, reproduced verbatim.
+ * The REPO HOUSE RULES block (#2154 slice 2) — the repository's curated standing rules, the set
+ * Shepherd injects into the system prompt of every agent it runs in this repo.
  *
- * It is reproduced rather than re-rendered on purpose: the critic is being shown exactly what the
- * author was told, so "the author was given this rule and ignored it" is a claim it can actually
- * make. (The block's own intro line therefore addresses the AUTHOR — the framing below says so.)
+ * WHAT IT IS NOT, and why the wording is careful about it. An earlier version of this block told the
+ * critic these rules were "injected into the agent that wrote this diff, reproduced verbatim", and
+ * rested the license to BLOCK on that premise. It is false twice over:
+ *   1. `prReviewPrompt` carries this block for every PR the standalone critic sweeps — third-party
+ *      and fork PRs included — where no Shepherd agent authored the diff and nothing was injected
+ *      into anything. `learningsEnabled` defaults ON, so that is the common case, not a corner.
+ *   2. Even at the session critic it is not a reproduction: the set is re-planned at review time
+ *      from the CURRENT active rules, scoped by the files the diff touches and ranked by a
+ *      recency decay evaluated NOW (see ReviewService.repoHouseRules) — so it can differ from what
+ *      the author actually received at spawn, in either direction.
+ * The block therefore claims only what is true everywhere — these are the repo's standing rules —
+ * and `authored` adds the author-facing framing ONLY where a Shepherd agent really did write the
+ * diff under a (possibly different) cut of these rules. The blocking license rests on the rules
+ * being the REPO's standard, which holds for any PR against it.
  *
  * ROUTING splits on how clear-cut the violation is, for the same reason the scope-creep / smells /
  * latent lenses split: ANY entry in `findings` advances the streak (buildVerdict), is auto-addressed
@@ -707,10 +724,18 @@ function reviewPolicyBlock(policy: string | null | undefined): string[] {
  * paused it. Rules are distilled from past sessions by an LLM and curated, not proven, so only an
  * unambiguous violation may block; everything softer goes to a non-blocking body section.
  */
-function reviewerHouseRulesBlock(houseRules: string | null | undefined): string[] {
+function reviewerHouseRulesBlock(
+  houseRules: string | null | undefined,
+  authored: boolean | undefined,
+): string[] {
   if (!houseRules || !houseRules.trim()) return [];
   return [
-    "REPO HOUSE RULES — the standing guidance Shepherd injected into the SYSTEM PROMPT of the agent that wrote this diff, reproduced verbatim below. Its intro addresses that AUTHOR, not you; your job is to check the diff against it. These are Shepherd-curated repo rules, not diff content. A rule the author was handed and ignored is exactly what review exists to catch:",
+    "REPO HOUSE RULES — this repository's curated standing rules, supplied by Shepherd. They are the same body of guidance Shepherd puts in the system prompt of every agent it runs in this repo, so the block below is written to address an agent DOING the work rather than reviewing it; your job is to judge the diff against it. Shepherd-supplied repo standard, NOT content from the diff:",
+    ...(authored
+      ? [
+          "- A Shepherd agent wrote this diff, and was given this same standing guidance when it started. (The exact set is re-planned for this review from the current rules and the files this diff touches, so it may differ in detail from the cut that agent saw — judge the diff against the rules shown here.) A rule the author was handed and ignored is exactly what review exists to catch.",
+        ]
+      : []),
     '- A CLEAR, UNAMBIGUOUS violation of a stated rule by a change IN THE DIFF is a finding: put it in "findings", name the rule you are applying, and block it per the usual rules.',
     '- Anything short of that — a rule that only arguably applies, one whose intent the diff meets by other means, or a stylistic reading of it — is a JUDGEMENT CALL. Report it in a SINGLE "body" section headed exactly `House rules (non-blocking):`, ONE LINE PER DISTINCT ITEM, do NOT put it in "findings", and it NEVER makes the decision "request-changes".',
     "- These rules are distilled from past sessions and can be stale or simply wrong for this change. They are never a reason to raise something you cannot see in the diff, and every item must concern a file in the diff per the SCOPE rule above.",
@@ -730,6 +755,9 @@ function scopeAndOutputTail(
     planClamped?: boolean;
     reviewPolicy?: string | null;
     houseRules?: string | null;
+    /** True only where a Shepherd agent authored the diff under this repo's rules — see
+     *  {@link reviewerHouseRulesBlock}. The standalone critic never sets it. */
+    houseRulesAuthored?: boolean;
   } = {},
 ): string[] {
   return [
@@ -843,7 +871,7 @@ function scopeAndOutputTail(
     // last thing read and the two blocks can only ADD to it. Both are empty by default, keeping
     // every existing prompt byte-identical.
     ...reviewPolicyBlock(opts.reviewPolicy),
-    ...reviewerHouseRulesBlock(opts.houseRules),
+    ...reviewerHouseRulesBlock(opts.houseRules, opts.houseRulesAuthored),
     // FINDINGS ROUTING (#1948). The tail used to mandate the exact opposite — "A non-blocking nit
     // STILL goes in findings" — which contradicted the three lenses above and their stated reason:
     // ANY entry in `findings` advances the streak (buildVerdict), is auto-addressed (runAutoAddress
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -15,6 +15,7 @@ import type { ReviewDecision } from "./types";
 import type { SessionUsage } from "./usage";
 import { tolerantParseJson } from "./json-tolerant";
 import type { VerdictRead } from "./json-tolerant";
+import { clampBlock } from "./prompt-fit";
 import { UNTRUSTED_CONTENT_DIRECTIVE, fenceUntrusted } from "./untrusted";
 
 const execFileAsync = promisify(execFile);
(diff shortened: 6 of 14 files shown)


scoop (commented):
# Review — #2154 slices 1+2: per-repo `REVIEW.md` + house rules in the reviewer

Round 1 of at most 4. No prior review artifacts were present in the worktree, so nothing to re-raise.

**Decision: comment.** I found no correctness, security, or contract defect. Everything below is
non-blocking.

---

### 1. The policy really is read from the base commit, not the PR head

`defaultReadReviewPolicy` (`src/critic-core.ts:117-144`) runs `git show <baseSha>:<file>` — a git
*object* read, never a tree read — with `cwd` set to the critic's detached worktree. So the checked-out
PR head cannot supply the text.

I chased `baseSha` back to its source rather than assuming it: it is produced by
`defaultComputePatchId` (`src/critic-core.ts:947-977`), which does `git fetch origin -- <base>` and then
`git rev-parse --verify --end-of-options FETCH_HEAD`, falling back to the *local* base ref when the
fetch fails. Both branches resolve an upstream base-branch commit; neither is reachable from the head
under review. That is what makes the unfenced treatment defensible, and the reader for both critics
passes the fingerprinted SHA:

- session critic: `src/review.ts:591` → `resolveReviewPolicy(wt.worktreePath, baseSha)` at
  `src/review.ts:945-950`, with `baseSha` destructured from `rebaseSkip` at `src/review.ts:542`.
- standalone critic: `src/standalone-critic.ts:524`, using `fp.baseSha` threaded from
  `src/standalone-critic.ts:472-479`.

`baseSha === null` short-circuits at both sites, so a total git failure yields *no* block rather than a
guess (`src/review.ts:549` guard on `skipped`, then the ternaries above).

The argv guard is the same one the sibling reader uses — `/^[0-9a-f]{7,40}$/` at
`src/critic-core.ts:123`, identical to `src/critic-core.ts:158` — and args go through `execFile` array
form, so `<sha>:<path>` cannot be flag-smuggled. `REVIEW_POLICY_FILES` (`src/critic-core.ts:92`) is a
module constant, so the path half is not attacker-influenced either.

### 2. The reader's failure modes

- missing file / bad SHA / git absent → `catch { continue }` at `src/critic-core.ts:137-139`, then
  `return null`. No throw path exists (the regex, `clampBlock`, and `String.trim` cannot throw).
- an empty or whitespace-only policy is treated as absent (`src/critic-core.ts:140`), matching the
  block builders' own `!policy.trim()` guard at `src/critic-core.ts:681`.
- precedence: `REVIEW.md` before `.shepherd/review.md`, first hit wins — matches the docs claim at
  `docs-site/src/content/docs/reference/review-policy.md:33`.
- `clampBlock(text, REVIEW_POLICY_MAX_BYTES, "head")` — signature verified against
  `src/prompt-fit.ts:102` (`text, keepBytes, mode`). `REVIEW_POLICY_MAX_BYTES = 8000`
  (`src/critic-core.ts:98`) is consistent with the docs' "capped at about 8 KB"
  (`docs-site/src/content/docs/reference/review-policy.md:90`). The test's `expect(policy).toContain("elided")`
  is satisfied by `elisionMarker` at `src/prompt-fit.ts:41`.
- `timedAsync(label, fn)` — signature verified at `src/instrument.ts:70`. Async, not sync, so the
  single event loop is not blocked.

### 3. Prompt placement and the byte-identity property

Both blocks are spread into `scopeAndOutputTail` at `src/critic-core.ts:845-846`, i.e. after the
LATENT-DEFECT lens (ends `src/critic-core.ts:839`) and immediately before FINDINGS ROUTING
(`src/critic-core.ts:853`). The policy block's own text says "FINDINGS ROUTING below"
(`src/critic-core.ts:684`, `:687`) — that is factually correct for this ordering, and likewise the
house-rules block's "SCOPE rule above" (`src/critic-core.ts:716`) resolves to `src/critic-core.ts:740`.

Both builders emit `[]` when the input is null/blank (`src/critic-core.ts:681`, `:711`), so a spread of
an empty array leaves the joined string untouched — the byte-identity claim holds structurally, not just
by test. `reviewPrompt` threads the opts at `src/critic-core.ts:320-327` and `prReviewPrompt` at
`src/critic-core.ts:376`; the shared verdict-output contract (`src/critic-core.ts:865-885`) is below both
blocks and unmodified.

### 4. House-rules parity with the author side

`ReviewService.repoHouseRules` (`src/review.ts:960-973`) and
`StandalonePrCriticService.repoHouseRules` (`src/standalone-critic.ts:186-195`) call
`planHouseRulesInjection(store.listActiveLearnings(repoPath), budget, now, files)` and then
`renderHouseRulesBlock(injected)` — the same two calls, in the same order, as the author path at
`src/service.ts:2844-2854`. Signatures verified at `src/house-rules.ts:213-218` and
`src/house-rules.ts:282-286`; `renderHouseRulesBlock` returns `null` for an empty set, which the block
builder's falsy guard handles.

`targetPaths` is the diff's changed-file list. I confirmed those are repo-relative forward-slash paths
(`src/review.ts:206`, `src/standalone-critic.ts:71`), which is the form `learningMatchesScope` →
`globMatchesAnyPath` → `Bun.Glob.match` expects (`src/house-rules.ts:163-183`, canonicalization at
`src/house-rules.ts:85-87`). So the scope-glob match is fed correct input.

The gate is `getRepoConfig(repoPath).learningsEnabled` at `src/review.ts:961` and
`src/standalone-critic.ts:187` — the same predicate the author side uses at `src/service.ts:2839`.
Field exists and defaults to `true` (`src/store.ts:420`, `src/store.ts:984`, `src/store.ts:1015`), so the
block is on by default wherever active learnings exist; that is the plan's stated intent, not a surprise.

`recordInjectedLearnings` is deliberately *not* called on either reviewer path. I checked that the
counter it feeds is in fact the auto-retire denominator: `injectedCount` is read against
`listActiveLearnings` in the lifecycle sweep (`src/learnings-lifecycle.ts:349`, `:368`). Not writing it
is the right call, and the test at `test/review.test.ts` asserts it stays at zero.

`store` `Pick`s were widened correctly on both services (`src/review.ts:272`,
`src/standalone-critic.ts:94`), and `getRepoConfig` was already present in both.

### 5. Budget constant hoist — no import cycle

`src/config.ts:116` now imports `HOUSE_RULES_DEFAULT_BUDGET_CHARS` from `./house-rules`. I traced the
import graph to confirm this cannot cycle back: `src/house-rules.ts:1-2` imports `./types` (type-only)
and one value, `isUnprovenTrial`, from `./learnings-lifecycle`; every import in
`src/learnings-lifecycle.ts:7-9` is `import type`. Nothing on that path reaches `./config`.

The value is unchanged (`4000`), and `Number(env ?? HOUSE_RULES_DEFAULT_BUDGET_CHARS)` at
`src/config.ts:130-132` is byte-equivalent in behaviour to the previous `Number(env ?? 4000)`. Every
other `planHouseRulesInjection` caller takes `budgetChars` as a parameter
(`src/learnings-service.ts:216`, `src/learning-sweep.ts:59`), so there is no second hard-coded 4000 left
to drift. The thunks are wired at `src/index.ts:1499-1381` (`reviewService`) and
`src/index.ts:1525-1390` (`standaloneCritic`), both reading `config.houseRulesBudgetChars` live.

### 6. Clamp ladder ordering

`fitAssembledPrompt` iterates `specs` in the caller's order (`src/prompt-fit.ts:214-225`), so spec order
*is* the priority. Session critic pushes `plan` then `reviewPolicy` (`src/review.ts:2005-2019`);
standalone critic pushes `body` then `reviewPolicy` (`src/standalone-critic.ts:584-596`). Both match the
stated policy: the pre-existing clampable gives way first.

The removed `if (!plan) return …specs: []` early-return in `fitCriticPrompt` is behaviourally identical
to the new always-run path: with no plan and no policy the spec list is empty, `spawnBudget` is still
measured, and `compose(null, false, null)` produces the same string. The `planClamped` derivation is
preserved exactly (`src/review.ts:2026`, guarded so a null plan yields `false` rather than
`undefined !== null`).

Null-guards on the composed values are correct at both sites: `v.plan` / `v.reviewPolicy` are only cast
when the corresponding spec was actually pushed (`src/review.ts:2025-2027`,
`src/standalone-critic.ts:597-598`).

`criticPromptComposer`'s signature change (`src/review.ts:910`) has exactly two consumers, both updated:
`src/review.ts:641` and `fitAndSpawn` at `src/review.ts:1598-1618`. `prReviewPrompt`'s new trailing
`opts` is additive and positional-safe; I grepped every call site — `src/standalone-critic.ts:531` plus
`test/critic-core.test.ts` and `test/untrusted.test.ts`, none of which pass a 6th argument they did not
mean to.

### 7. Docs artifacts

`ui/src/lib/docs-manifest.ts` is generated with a drift gate (`ui/src/lib/docs-manifest.ts:1-3`), so I
reproduced what the generator would emit rather than eyeballing it:

- ordering is `a.path.localeCompare(b.path)` (`ui/scripts/gen-docs-manifest.ts:127`). The new entry's
  path `/reference/review-policy/` sorts after `/reference/plugins/` and before
  `/reference/rules-design-system/` ("re" < "ru"), which is exactly where it was inserted
  (`ui/src/lib/docs-manifest.ts:842-848`).
- keywords are `description + H2/H3 headings`, lowercased and whitespace-collapsed
  (`ui/scripts/gen-docs-manifest.ts:77-84`). Note that `extractHeadings`
  (`ui/scripts/gen-docs-manifest.ts:58-64`) does **not** skip fenced code blocks — so the sample policy's
  `## Extra passes` / `## Known exclusions` inside the ```markdown fence at
  `docs-site/src/content/docs/reference/review-policy.md:41-48` *should* appear in the keyword string.
  They do, in document order, in the committed entry. That is a strong signal the generator was actually
  run rather than the entry hand-written.
- the sidebar slug `reference/review-policy` (`docs-site/astro.config.mjs:131`) matches the file at
  `docs-site/src/content/docs/reference/review-policy.md`.

The feature-catalog gate does not arm: `UI_GLOBS` is `ui/src/lib/components/` + `ui/src/routes/`
(`scripts/check-feature-catalog.sh:40`), and the only `ui/` file this `feat` PR touches is
`ui/src/lib/docs-manifest.ts`. `scripts/check-generated-docs.sh:46-48` gates only the herdr CLI
reference, untouched here. No `ui/messages/*.json` keys were added, so there is no en/de parity
obligation to check.

The docs page's factual claims check out: "Review all PRs" is the real setting label
(`ui/messages/en.json:1295`); `REVIEW.md` wins over `.shepherd/review.md`
(`src/critic-core.ts:92`); reviewer-side injection writes no usage rows (`src/review.ts:970-973`,
`src/standalone-critic.ts:192-194`).

### 8. Ordering invariant in `begin()`

The policy read is awaited at `src/review.ts:591`, beside `fetchIssueBody` (`:571`) and
`resolveEpicContext` (`:584`), and *before* the `starting.has(session.id)` tombstone re-check at
`src/review.ts:597`. The invariant that the re-check is the last await-gated step before the spawn is
preserved. The house-rules resolution at `src/review.ts:626` is a synchronous store read placed after
the re-check, so it does not touch the invariant either.

---

## Stated limitations

- I could not verify that the slice-3 follow-up issue exists or is linked from the PR body — GitHub
  state is not in this repo. Assumed as the plan describes it.
- I did not run `bun run test`, `bun run lint`, `bunx tsc --noEmit`, or the docs-site build; per the
  brief those are CI's job. In particular the contextual typing of the mixed
  `specs: [{…}, ...(cond ? [{… as const}] : [])]` array literal at `src/standalone-critic.ts:584-596`
  is a `tsc` question I reasoned about but did not compile.
- `Bun.Glob` matching semantics for the scope globs were read from `src/house-rules.ts:163-183`, not
  executed.

---

- `src/review.ts:2005-2019` / `src/standalone-critic.ts:584-596`: the house-rules block is deliberately excluded from the clamp ladder on the stated grounds that it is "already bounded by the house-rules char budget", but that budget is `Number(process.env.SHEPHERD_HOUSE_RULES_BUDGET_CHARS ?? 4000)` (`src/config.ts:130`) with no upper clamp — an operator who raises that env var raises the *unclampable* share of the critic prompt, and the only recovery is the argv refusal. The failure mode is loud rather than silent, so this is a note, not a defect.
- `src/critic-core.ts:141`: a policy clamped on read reaches the prompt as a bare `[… N bytes elided …]` marker with no accompanying explanation, unlike the plan block which gets an explicit `planClamped` NOTE (`src/critic-core.ts:269-278`). Harmless today because the marker states only a fact, but a critic reading a truncated policy has no way to know the truncation was mechanical.

## Nits (non-blocking):

- `src/review.ts` / `src/standalone-critic.ts`: neither spawn site logs that a `REVIEW.md` policy or a house-rules block was in effect. The plan named "a single `console.log` line at spawn" as the whole operator-facing trace, and no such line exists in the diff — a policy that silently stopped resolving (renamed file, base-branch churn) would be invisible unless the argv ladder happens to clamp it and log that.
- `src/review.ts:960-973` and `src/standalone-critic.ts:186-195`: `repoHouseRules` is duplicated byte-for-byte across the two services, and is a near-duplicate of `src/service.ts:2837-2855`. A shared helper in `src/house-rules.ts` taking `(rules, budget, now, files)` would have kept one copy of the gate + plan + render sequence.
- `src/critic-core.ts:713`: the new `House rules (non-blocking):` section joins `Scope creep / gold-plating`, `Possible smells` and `Latent / future-reachable` in *not* being named as optional by the auto-address steer, which calls out only `` `Nits (non-blocking):` `` (`src/review.ts:104`). An author agent reading the posted review may treat house-rules judgement calls as required work. Pre-existing shape, extended by one more section here.
- `src/critic-core.ts:683`: the block's opening line calls the file "`REVIEW.md`" unconditionally even when the text actually came from `.shepherd/review.md` (`src/critic-core.ts:92`). Cosmetic, but a critic asked to cite where a rule came from would cite the wrong path.

- Nothing. `ui/src/lib/docs-manifest.ts` is not in the plan's Files table, but it is a generated artifact that `check:docs-manifest` (`ui/scripts/gen-docs-manifest.ts:149-162`) would fail without — mandatory, not creep. Slice 3 is correctly absent per the plan's explicit `Out of scope` boundary.

- `src/config.ts` reads several numeric env overrides with a bare `Number(...)`, which yields `NaN` for a malformed value, while `src/house-rules.ts:34-39` has an `envNum()` helper that guards exactly that. `houseRulesBudgetChars` uses the unguarded form; this PR preserved the existing behaviour rather than introducing it.
- The standalone critic's clamp-ladder comment at `src/standalone-critic.ts:577-578` says the ladder clamps "then the title", but no `title` spec is or ever was in `specs` — the comment predates this PR.

## A note on the trust boundary (informational)

The security argument rests on "only merged, already-reviewed policy is honored". That is exact for a
PR based on `main`. It is slightly weaker for an **epic child**, whose `baseSha` is the epic integration
branch tip (`src/review.ts:583-584`, `src/standalone-critic.ts:455-464`): a child PR that adds
`REVIEW.md` and merges into that integration branch would steer the reviews of its later siblings, and
with hands-off epics that merge can happen without a human reading the diff. I am not raising this as a
finding — the child's own diff is still reviewed (without the policy) before it merges, the block's
precedence paragraph (`src/critic-core.ts:684-686`) forbids suppressing a verified defect, and anything
that lands on a base branch already influences the critic far more directly (the code it reads, CI
config, house rules). Worth an operator knowing, though, since
`docs-site/src/content/docs/reference/review-policy.md:58-72` presents "merged" as the whole boundary.

scoop:
# Review — #2154 R2 slices 1+2 (per-repo `REVIEW.md` + house rules for the critic)

Round 1 of at most 4. Scope: `git diff 266ef76e...HEAD` (14 files).

## What I verified

**Slice 1 — `REVIEW.md` from the base commit.**

- `defaultReadReviewPolicy` (`src/critic-core.ts:117-138`) guards the SHA with
  `/^[0-9a-f]{7,40}$/` — byte-identical to the existing guard in `defaultCollectBaseDelta`
  (`src/critic-core.ts:123` vs `src/critic-core.ts:158`), so the "same guard" claim in the comment
  checks out.
- `timedAsync(label, fn)` matches the call shape used — verified against `src/instrument.ts:70`.
  `clampBlock(text, keepBytes, mode)` matches — verified against `src/prompt-fit.ts:102`, and
  `"head"` is a valid `ClampMode` (`src/prompt-fit.ts:89`).
- `REVIEW_POLICY_FILES` uses `.shepherd/review.md` (`src/critic-core.ts:92`); the same spelling and
  casing appear in the docs page (`docs-site/src/content/docs/reference/review-policy.md:14`) and
  in `docs/sandbox-security.md:114`. `.shepherd/` is an existing in-repo convention
  (`src/repo-roles.ts:21` — `.shepherd/roles.json`), so the sidecar location does not collide.
- The base object is reachable from the critic's detached worktree: `defaultComputePatchId` fetches
  `origin <base>` and rev-parses it before returning `baseSha` (`src/critic-core.ts:951-971`), and
  the same `worktreePath` is passed to the reader (`src/review.ts:596`,
  `src/standalone-critic.ts:524`). `git show <sha>:<path>` streams the raw blob, so a PR-authored
  `.gitattributes` in the checked-out head cannot filter the text.
- Placement invariant holds: `resolveReviewPolicy` is awaited at `src/review.ts:596`, i.e. beside
  `fetchIssueBody` (`:571`) and `resolveEpicContext` (`:584`) and **before** the
  `starting.has(session.id)` tombstone re-check at `src/review.ts:603`, which remains the last
  await-gated step before the spawn.
- Clamp ladder ordering is as documented: `plan` is pushed first and `reviewPolicy` second
  (`src/review.ts:2015-2027`), and `fitAssembledPrompt` iterates `specs` in caller order with an
  early `break` once the prompt fits (`src/prompt-fit.ts:214-216`) — so the plan genuinely gives way
  first. The old `if (!plan) return … specs: []` early return is preserved as an
  always-run-the-ladder path (empty `specs` when neither block is present), so the #1944 guarantee
  is not regressed. In `src/standalone-critic.ts:531-545` the policy spec is likewise appended after
  `body`.
- `ClampSpec` is exported (`src/prompt-fit.ts:130`) so the new `type ClampSpec` import in
  `src/review.ts:76` resolves.

**Slice 2 — house rules into the reviewer.**

- `planHouseRulesInjection(rules, budgetChars, now, targetPaths)` and
  `renderHouseRulesBlock(injected)` are used with the exact signatures declared at
  `src/house-rules.ts:213-218` and `src/house-rules.ts:282`; `renderHouseRulesBlock` returns `null`
  for an empty set (`src/house-rules.ts:283`), which is what both `repoHouseRules` helpers rely on
  (`src/review.ts:963-970`, `src/standalone-critic.ts:184-191`).
- The reviewer call matches the author-side call at `src/service.ts:2836-2854` on gate
  (`getRepoConfig(repoPath).learningsEnabled`), rule source (`listActiveLearnings(repoPath)`) and
  budget — the only intentional differences being `targetPaths` and the absence of
  `recordInjectedLearnings`. The `recordInjectedHouseRules` / `recordInjectedLearnings` identifiers
  named in the new doc comments both exist (`src/service.ts:2836`, `src/store.ts:5063`).
- `files` from `rebaseSkip` are repo-relative, NUL-split `git diff --name-only -z` paths
  (`src/critic-core.ts:1003-1009`) — the right shape for `normalizeGlob`-canonicalised globs
  (`src/house-rules.ts:85-87`), so the scope match is sound.
- `listActiveLearnings` is present in both store `Pick`s (`src/review.ts:272`,
  `src/standalone-critic.ts:94`), and `getRepoConfig` was already there
  (`src/standalone-critic.ts:85`). `RepoConfig.learningsEnabled` exists and defaults to ON
  (`src/store.ts:420`, `src/store.ts:1263` — `DEFAULT 1`).
- No import cycle from hoisting the budget constant: `src/config.ts:18` imports
  `./house-rules`, which runtime-imports only `isUnprovenTrial` from `./learnings-lifecycle`, whose
  own imports are all `import type` (`src/learnings-lifecycle.ts:7-9`), as are `src/types.ts:1-4`.
  Nothing in that chain reaches back to `./config`, so `config`'s module-level initialiser cannot hit
  a TDZ.

**Prompt/contract wiring.**

- Both blocks are emitted between the LATENT-DEFECT lens and FINDINGS ROUTING
  (`src/critic-core.ts:841-846`), and both return `[]` for null/blank input
  (`src/critic-core.ts:681`, `src/critic-core.ts:711`) — so the default prompt is byte-identical.
- The section names the policy block declares non-negotiable ("SCOPE", "VERIFY", "CANNOT-VERIFY",
  "ATTRIBUTION", "FINDINGS ROUTING") all exist verbatim in the shared tail below it
  (`src/critic-core.ts:838-847` and following).
- `test/untrusted.test.ts:122-123` constructs both prompts without the new opts, so the
  fence/directive invariant is untouched.

**Docs.**

- `docs-manifest.ts` is consistent with its generator. `ui/scripts/gen-docs-manifest.ts:127` sorts by
  path, and `/reference/review-policy/` correctly falls between `/reference/plugins/` and
  `/reference/rules-design-system/` (`ui/src/lib/docs-manifest.ts:104`). `keywordsOf`
  (`ui/scripts/gen-docs-manifest.ts:77-84`) = lowercased description + H2/H3 headings in document
  order; the committed keyword string matches the six headings at
  `docs-site/src/content/docs/reference/review-policy.md:12,23,30,40,56,76` — including
  "extra passes" / "known exclusions", which the generator's fence-unaware
  `extractHeadings` regex (`ui/scripts/gen-docs-manifest.ts:60-62`) does pick up out of the sample
  code block. So the drift gate should pass.
- `docs/sandbox-security.md` is a `sync-docs` source (`docs-site/scripts/sync-docs.mjs:67`), but the
  addition introduces no new H2/H3, so its manifest keywords are correctly unchanged.
- The feature-catalog gate does not arm: `UI_GLOBS` is `ui/src/lib/components/` and `ui/src/routes/`
  (`scripts/check-feature-catalog.sh:40`), and the only `ui/` file touched is
  `ui/src/lib/docs-manifest.ts`.
- The glossary link `/reference/review-policy/`
  (`docs-site/src/content/docs/reference/glossary.md:52`) matches the Starlight slug for
  `reference/review-policy.md` and the sidebar entry (`docs-site/astro.config.mjs:131`).

## Findings

### 1. `src/critic-core.ts:713` — the REPO HOUSE RULES header asserts a fact that is not true at the standalone critic (and not guaranteed at the session critic)

The block opens with:

> "REPO HOUSE RULES — the standing guidance Shepherd injected into the SYSTEM PROMPT of the agent
> that wrote this diff, reproduced verbatim below. Its intro addresses that AUTHOR, not you … A rule
> the author was handed and ignored is exactly what review exists to catch"

Neither half of that is reliably true, and the block is emitted at both critics:

- **`prReviewPrompt` gets the identical block** (`src/critic-core.ts:376-378`, fed from
  `src/standalone-critic.ts:525`). That critic reviews *every open PR* in a repo with `criticAllPrs`
  on (`src/standalone-critic.ts:223`), explicitly including third-party and fork PRs — the diff's own
  comment at `src/standalone-critic.ts:521-523` says "the head may be a fork nobody vetted". For
  those PRs no Shepherd agent wrote the diff and nothing was injected into anyone's system prompt.
  The gate is `learningsEnabled`, which defaults to ON (`src/store.ts:1263`), so this is the default
  path, reachable today.
- Even at the session critic the block is **re-planned at review time**, not reproduced: it is built
  from the *current* `listActiveLearnings` set with review-time `targetPaths` (the changed files) and
  a review-time `now` that drives the ranking (`src/review.ts:963-970` vs the author-side
  `src/service.ts:2840-2848`, which uses `extractTargetPaths` over the task text and `Date.now()` at
  spawn). Rules added, revised, promoted or retired between spawn and review, and rules whose budget
  ranking flipped, differ from what the author actually received. "reproduced verbatim" is a claim
  the code does not make good on.

This matters because the very next line licenses blocking on that premise: *"A CLEAR, UNAMBIGUOUS
violation of a stated rule … is a finding: put it in `findings` … and block it"*
(`src/critic-core.ts:714`). Any entry in `findings` advances the streak and triggers auto-address, as
the diff's own doc comment notes (`src/critic-core.ts:702-706`) — so the false premise converts into
a real merge block against a contributor who was never handed the rules and cannot see them. The
mitigation line ("can be stale or simply wrong", `src/critic-core.ts:716`) addresses rule *quality*,
not the authorship claim.

Fix: state what is true — these are the repo's curated standing rules that Shepherd injects into
agents writing code in this repo — and reserve the "the author was handed this and ignored it"
framing for the session critic, where an author session actually exists (or drop it, since the
useful instruction is "check the diff against these rules", which stands on its own). The same
overclaim is repeated in the new docs page
(`docs-site/src/content/docs/reference/review-policy.md:78-80`).

## Nits (non-blocking):

- `docs/sandbox-security.md:113` says the PR critic's prompt "carries one block _unfenced_" — as of this PR it carries two: the review policy and the `<shepherd-house-rules>` block, which is equally unfenced trusted instruction at the same site (`src/critic-core.ts:710-718`). Worth naming the second one in the same bullet.
- `src/critic-core.ts:683` always names the source as `` `REVIEW.md` `` even when the text was read from `.shepherd/review.md` (`src/critic-core.ts:92`); threading the winning filename through would make the citation accurate.
- `src/critic-core.ts:112-114`: the comment says `maxBuffer` is "well past the cap below, so an oversized policy is CLAMPED with a visible marker rather than lost to a maxBuffer kill" — true only up to 4 MiB; a blob larger than that still errors into `continue` and yields no block at all, which is exactly the "no policy at all" outcome the comment says is avoided. Bounding the claim (or reading with a `--` + `git cat-file -s` size check) would make it exact.
- `src/review.ts` / `src/standalone-critic.ts`: the plan's stated minimum operator trace — "a single `console.log` line at spawn is the whole operator-facing trace" — is not implemented. There is no log, notice or persisted signal that a `REVIEW.md` or a house-rules block was in effect for a given review, so an operator cannot tell after the fact why a review behaved the way it did. (`src/standalone-critic.ts:548-552` logs clamps only.)
- `src/critic-core.ts:132`: a whitespace-only `REVIEW.md` falls through to `.shepherd/review.md`, which is a benign but undocumented deviation from the "if both exist, `REVIEW.md` wins" precedence stated at `docs-site/src/content/docs/reference/review-policy.md:14`.

## Stated limitations

- **Slice-3 follow-up issue (success criterion 7).** The plan requires a follow-up issue for the
  machine-readable-severity slice, linked from the PR body. That lives on GitHub, which is not in this
  repo and not reachable from this read-only checkout — I could neither confirm nor refute it, and it
  is not a finding.
- I did not run `bun run test`, `bun run lint`, `bunx tsc --noEmit` or the docs-site build — CI owns
  those per the review contract. The type-level claims above (store `Pick`s, `ClampSpec` export,
  composer signature, absence of an import cycle) were checked by reading the declarations, not by
  compiling.

scoop (commented):
# Review — #2154 slices 1+2: per-repo `REVIEW.md` + house rules in the reviewer (round 2)

**Decision: comment.** No blocking defect survived verification. The single prior finding is fully
addressed in both the code and the docs page it also named.

## Prior finding — verified addressed

> src/critic-core.ts: the REPO HOUSE RULES block header (line 713) tells the critic these rules were
> injected into the system prompt of the agent that wrote this diff and are reproduced verbatim …
> Restate the header as the repo's curated standing rules and confine the author-was-handed-this
> framing to the session critic; the docs page repeats the overclaim at
> docs-site/src/content/docs/reference/review-policy.md:78-80.

Confirmed fixed, on all three counts:

- **Header restated.** `src/critic-core.ts:733` now opens `"REPO HOUSE RULES — this repository's
  curated standing rules, supplied by Shepherd. They are the same body of guidance Shepherd puts in
  the system prompt of every agent it runs in this repo…"` — a claim about the *repo*, true for a
  fork PR as much as for a session PR. The "injected into the agent that wrote this diff /
  reproduced verbatim" premise is gone.
- **Author framing confined to the session critic.** `reviewerHouseRulesBlock` takes an `authored`
  flag (`src/critic-core.ts:729-731`) and only then emits the author line at
  `src/critic-core.ts:736`. `reviewPrompt` sets `houseRulesAuthored: true`
  (`src/critic-core.ts:329`); `prReviewPrompt` passes only `{ landing, reviewPolicy, houseRules }`
  (`src/critic-core.ts:382`) and never sets it — so the standalone sweep fed from
  `src/standalone-critic.ts:534` cannot emit it. Pinned by
  `test/critic-core.test.ts` ("the standalone critic is NOT told an agent authored the diff under
  these rules", asserting `not.toContain("A Shepherd agent wrote this diff")`).
- **Re-planning no longer misdescribed as a reproduction.** The author line now carries the
  parenthetical "The exact set is re-planned for this review from the current rules and the files
  this diff touches, so it may differ in detail from the cut that agent saw"
  (`src/critic-core.ts:736`), matching what `ReviewService.repoHouseRules`
  (`src/review.ts:964-973`) actually does — a fresh `planHouseRulesInjection` over
  `listActiveLearnings` with `this.now()` and review-time `files`.
- **Docs corrected.** `docs-site/src/content/docs/reference/review-policy.md:84-89` now reads "The
  set the critic sees is **planned fresh for that review** … It is not a replay of what any
  particular agent was given, and it is not limited to pull requests Shepherd wrote", with the
  author framing scoped to "Where Shepherd's own session critic reviews a PR one of its agents
  wrote".

Identifier / contract resolution:

- `clampBlock(text, keepBytes, mode)` — signature verified at `src/prompt-fit.ts:102`; `keepBytes`
  is compared against `Buffer.byteLength`, so `REVIEW_POLICY_MAX_BYTES = 8000`
  (`src/critic-core.ts:98`) is genuinely a byte cap. `head` is a valid `ClampMode` (used at
  `src/prompt-fit.ts:117`).
- `ClampSpec` accepts `{id, kind:"text", text, mode}` with `minUseful` optional —
  `src/prompt-fit.ts:130-139`. The new `reviewPolicy` spec omits `minUseful`, so it falls back to
  the termination floor `MIN_CLAMP_KEEP` (`src/prompt-fit.ts:15`, applied at
  `src/prompt-fit.ts:260`) rather than tripping the `plan-unreviewable` refusal path — correct, a
  policy has no usefulness floor to defend.
- Ladder order is the caller's spec order (`src/prompt-fit.ts:213` iterates `specs` in order), so
  `plan` before `reviewPolicy` (`src/review.ts:2005-2020`) and `body` before `reviewPolicy`
  (`src/standalone-critic.ts:586-599`) do what their comments claim.
- `planHouseRulesInjection(rules, budgetChars, now, targetPaths)` — signature at
  `src/house-rules.ts:213-218`; both new call sites (`src/review.ts:966-971`,
  `src/standalone-critic.ts:191-196`) pass the arguments in the same order as the author side
  (`src/service.ts:2844-2849`). `renderHouseRulesBlock` returns `string | null`
  (`src/house-rules.ts:282`), matching both helpers' declared `string | null`.
- Path-shape parity for scoping: `learningMatchesScope` globs against repo-relative paths
  (`src/house-rules.ts:180-183`), and `files` on both sides comes from `git diff --name-only -z`
  (`src/critic-core.ts:1030-1035`), i.e. repo-relative — so review-time scoping matches the author
  side's `extractTargetPaths` (`src/house-rules.ts:150`) in form.
- `store.listActiveLearnings` exists (`src/store.ts:4923`) and `getRepoConfig(...).learningsEnabled`
  is always populated (`src/store.ts:1015`, default `1` at `src/store.ts:1263`), so the gate in both
  `repoHouseRules` helpers cannot read `undefined` in production.
- `HOUSE_RULES_DEFAULT_BUDGET_CHARS` (`src/house-rules.ts:26`) = 4000, exactly the literal it
  replaced in `src/config.ts` — the env override path (`src/config.ts:806-808`) is unchanged.
- Both new deps are wired in `src/index.ts:1502` and `src/index.ts:1528`; the only production
  constructions of either service are those two (grep of `new ReviewService(` /
  `new StandalonePrCriticService(`).
- `baseSha` is a full lowercase 40-hex SHA from `git rev-parse --verify` (`src/critic-core.ts:993`),
  so the `/^[0-9a-f]{7,40}$/` guard at `src/critic-core.ts:117` admits it and the shell-metachar
  smuggling case is closed (pinned by the `"not-a-sha; rm -rf /"` test).

Placement / byte-identity:

- The two blocks are emitted after the LATENT lens and immediately before `FINDINGS ROUTING`
  (`src/critic-core.ts:869-875` vs the routing header at `src/critic-core.ts:880`), and `SCOPE — `
  is above them (`src/critic-core.ts:768`). So the block bodies' "FINDINGS ROUTING below" and "the
  SCOPE rule above" references are both accurate.
- Both helpers return `[]` on null/blank (`src/critic-core.ts:684`, `src/critic-core.ts:732`), and
  `prReviewPrompt`'s new `opts` defaults to `{}` (`src/critic-core.ts:351-357`) — so the existing
  identity tests (`test/critic-core.test.ts:347-349`, `:316-318`) still compare unchanged strings.
- Await placement in `begin()`: the policy read at `src/review.ts:591` sits before the `starting`
  tombstone re-check at `src/review.ts:598`, preserving the "last await-gated step before the spawn"
  invariant. The standalone critic holds its `starting` claim across `spawnAndRecord` (added at
  `src/standalone-critic.ts:378`, released in the `finally` at `src/standalone-critic.ts:484`), so
  the new await at `src/standalone-critic.ts:527` introduces no re-entrancy window.

Docs / manifest parity:

- `docs-site/astro.config.mjs:131` slug `reference/review-policy` matches the new file
  `docs-site/src/content/docs/reference/review-policy.md`, and the glossary link
  `/reference/review-policy/` (`docs-site/src/content/docs/reference/glossary.md:52`) follows the
  same form as the existing `/reference/security/` link at `:135`.
- `ui/src/lib/docs-manifest.ts:104-110` is inserted in path-sorted position between
  `/reference/plugins/` and `/reference/rules-design-system/`, matching the generated ordering of
  the surrounding entries.
- No i18n catalogs are touched by this diff (no `ui/messages/*.json` change, no new message key), so
  there is no en/de parity obligation here.
- The feature-catalog gate arms on `ui/src/lib/components/**`, `ui/src/routes/**`,
  `ui/src/lib/feature-announcements/**` (`.claude/rules/ui-feature-catalog.md:2-5`); the only `ui/`
  file in the `feat` commit is `ui/src/lib/docs-manifest.ts`, which is outside all three — so no
  announcement entry is required.

## Scope

Slice 3 (machine-readable severity) is absent, and that is the plan's explicit `Out of scope`
boundary with an operator decision recorded against it. I did not raise it as "does not satisfy the
task": the diff delivers slices 1 and 2 completely and the narrowing is stated, not silent. I could
not verify from this repo that the promised follow-up issue actually exists (GitHub state is not in
the tree) — stated as a limitation, not a finding.

## Stated limitations

- I did not run tests or lint (CI's job, per instructions). In particular I did not execute the new
  temp-git-repo tests in `test/critic-core.test.ts`.
- `test/signal-capture.test.ts` was named in the plan's file table but is not in the diff. That is
  consistent, not a gap: its fake `getRepoConfig` never sets `learningsEnabled`, so
  `repoHouseRules` short-circuits at `src/review.ts:965` before touching
  `listActiveLearnings` — but I confirmed this by reading the fakes, not by running the suite.
- Whether the follow-up issue for slice 3 exists (success criterion 7) is GitHub state I cannot
  verify from this checkout.

## Nits (non-blocking):

- `src/critic-core.ts:130-132`: the `maxBuffer: 4 * 1024 * 1024` comment claims an oversized policy is "CLAMPED with a visible marker rather than lost to a maxBuffer kill" — true up to 4 MiB, but a larger blob throws, hits the `catch { continue }` at `src/critic-core.ts:136` and is dropped silently; the guarantee is bounded, not absolute.
- `src/critic-core.ts:736`: "was given this same standing guidance when it started" is hedged only for *detail* drift; it is flatly false in the narrow case where the author session got no block at all (learnings toggled on after that spawn, or the author-side `extractTargetPaths` scrape matched nothing while review-time `files` do). A "if it received any" style qualifier would close it.
- `docs-site/src/content/docs/reference/review-policy.md:15`: "If both exist, `REVIEW.md` wins" is true only for a non-blank `REVIEW.md` — `src/critic-core.ts:140` treats a whitespace-only file as absent and falls through to `.shepherd/review.md`. Half a sentence would make the documented precedence exact.
- `src/review.ts:104`: `steerText` still names only `Nits (non-blocking):` as the optional section, so the new `House rules (non-blocking):` items (like the pre-existing scope-creep/latent sections) can read to an auto-addressed author as required work. Same shape as the existing omission, now with one more section behind it.
- `src/standalone-critic.ts:528`: the house-rules block reaches reviews the critic posts on third-party and fork PRs, and `src/critic-core.ts:738` instructs it to "name the rule you are applying" — so a repo's internally-distilled rules can be quoted verbatim in a review visible to outside contributors. Worth a conscious call rather than a surprise.

- `src/house-rules.ts:26` + `src/config.ts:806-808` + the `houseRulesBudgetChars` thunk seam on both services (`src/review.ts:373`, `src/standalone-critic.ts:136`, wired at `src/index.ts:1502`/`:1528`): three moving parts for what is one config number both call sites read identically. Plan-sanctioned and defensible for the import-cycle reason given, but a direct `config.houseRulesBudgetChars` read in `src/index.ts` would have been the smaller change.

- `src/critic-core.ts:100-115`: the "only already reviewed and merged policy is honored" argument is exactly as strong as the base branch's own merge bar. For an epic child the base is the Shepherd-written integration branch, so a `REVIEW.md` merged there by a sibling agent becomes unfenced instruction for every later child's critic. Bounded (the PR that adds it is not judged under it), but weaker than the `main`-branch framing in `docs/sandbox-security.md:112-124` and `docs-site/src/content/docs/reference/review-policy.md:46-54` implies.

PR #2167, opened 2026-09-02
fix(doc-agent): regen the docs manifest a doc edit makes stale (#2163)

### LINKED ISSUE #2163: Glossary rule doesn't mention the docs-manifest regen it makes stale
Adding a glossary term makes a **committed generated artifact** stale two hops away, and nothing in the glossary rule says so. It cost a red `verify` on #2153, where the failure surfaced on a *different* commit than the one that caused it.

## The chain

1. A PR adds an entry to `ui/src/lib/glossary.ts` + `gloss_*` keys to both locale catalogs — everything `.claude/rules/ui-glossary.md` asks for. `scripts/check-glossary.mjs` passes. CI is green.
2. Later, someone (in practice the **doc agent**) documents the new term in `docs-site/src/content/docs/reference/glossary.md` — a committed, hand-authored page.
3. `ui/scripts/gen-docs-manifest.ts` derives the command bar's Docs-group keywords from that page's body, into the committed `ui/src/lib/docs-manifest.ts`.
4. `check:docs-manifest` (in `verify`) now fails until someone runs `bun run gen:docs` from `ui/` and commits the result.

## Why it is worth a line in the rule

The gate works — it is not broken, and it caught this. The problem is **attribution**: step 2 is usually a separate commit by a different actor, so the red `verify` lands on a `docs: sync docs to recent source changes` commit that looks unrelated to the glossary change that actually caused it. On #2153 the doc agent pushed onto the open PR branch and the manifest gate went red there; the cause was two commits and one author removed from the symptom.

`.claude/rules/ui-glossary.md` documents steps 1 and the `check-glossary.mjs` gate, and stops. A reader following it exactly still ends up with a stale manifest downstream.

## Proposed fix

Add a short note to the **Gate** section of `.claude/rules/ui-glossary.md`, along the lines of:

> A term documented on the docs-site glossary page (`docs-site/src/content/docs/reference/glossary.md`) also feeds the command bar's Docs manifest. Whoever edits that page — you, or the doc agent in a later sync — must run `bun run gen:docs` from `ui/` and commit `ui/src/lib/docs-manifest.ts` in the same commit, or `check:docs-manifest` fails in `verify`.

Alternatives worth weighing instead of (or with) the doc note:

- Have the doc agent run `bun run gen:docs` whenever its sync touches a file under `docs-site/src/content/docs`, so the class of failure cannot reach CI at all. Strictly better if the doc agent's commit step can run it — but it does not close the hole for a human editing that page by hand.
- Mention the coupling in `.claude/rules/ui-feature-catalog.md` too, since the same regen trap applies to any committed docs page edit, not just glossary terms.

Context: #2153 (delivery-metrics lens) added `first-pass-rate` and `lead-time`; commit `a4fd574c` synced the page, `15e39980` regenerated the manifest.

scoop:
Closes #2163.

`ui/src/lib/docs-manifest.ts` is a **committed** artifact. `ui/scripts/gen-docs-manifest.ts` derives it from every committed docs page's frontmatter `description` plus its **H2/H3 headings** — and `docs-site/scripts/sync-docs.mjs` publishes `CLAUDE.md` and every `.claude/rules/*.md` as such a page too. Each glossary term on the docs-site glossary page is its own `###`, so documenting a term drifts the manifest and `check:docs-manifest` goes red — historically on a `docs: sync docs to recent source changes` commit, two hops and one author from the glossary change that caused it (#2153).

## What changed

**The doc agent keeps the manifest fresh itself** (`src/doc-agent.ts`). Once the in-scope staging has actually produced changes, `stageDocsManifest()` regenerates the manifest and stages it into the same commit. It is fail-closed like the existing prettier step: a failing regen throws `DocsManifestError`, aborting the run with no commit/push/PR and an `error` outcome — a skipped doc update beats a doc PR that lands `check:docs-manifest` red.

**The generator can be handed a pinned prettier** (`ui/scripts/gen-docs-manifest.ts`). `render()` now resolves its prettier module through `prettierModuleSpecifier()`, which returns bare `"prettier"` unless `SHEPHERD_PRETTIER_MODULE` is set. Humans and CI are unaffected.

**Both rules carry the coupling** (`.claude/rules/ui-glossary.md`, `.claude/rules/ui-feature-catalog.md`), for the hand-edit path the code cannot cover.

## Why the prettier pin is load-bearing

A doc-agent worktree has **no `node_modules` at all**: every transient agent spawns with `disableAllHooks` (`src/transient-agent-argv.ts`), so the repo's `ensure-deps` SessionStart hook never runs there. Verified by probe: in that situation Bun *silently satisfies* the generator's bare `import("prettier")` by **auto-installing** from its global cache (`~/.cache/.bun/install/cache/prettier@3.9.6@@@1/index.mjs`) — an unpinned version whose formatting can differ from the repo's, emitting a manifest that fails the generator's own `--check`. That would have re-created the exact failure this PR closes, while appearing to work.

So the default impl runs the **worktree's** generator with `cwd = SERVER_INSTALL_ROOT` (prettier resolves `.prettierrc`'s bare plugin specifiers relative to cwd — confirmed by the resolver naming `<cwd>/noop.js` as the importer) and `SHEPHERD_PRETTIER_MODULE` pinned to the server install's copy. `cwd` does not retarget the generator: it derives the repo root from its own `import.meta.url`.

## Design notes

- **No second path list.** The regen fires whenever anything in-scope was staged, rather than testing staged paths against the set of manifest inputs — those span `IN_SCOPE_PATHS` *and* the sync-docs `PAGES` sources, so such a list would have to track both and would rot silently.
- **Gated on "something was staged"**, which also stops a run whose docs are unchanged from opening a manifest-only PR when `main`'s manifest is already stale for an unrelated reason.
- `prettierFailed` → `stageFailed`: one flag now covers both fail-closed staging aborts. Internal to `src/doc-agent.ts`.
- Layout guard: the regen is skipped (logged, not failed) when the repo has no `ui/scripts/gen-docs-manifest.ts`.

## Verification

- Root `bun run lint`, `bun run test` (8859 pass) · ui `bun run check`, `bun run test` (4738 pass), `bun run check:docs-manifest`
- 6 new doc-agent tests + 3 generator tests: regen-and-stage, not-run-when-nothing-staged, ordering (after the in-scope add, before the manifest add), layout-guard skip, `DocsManifestError` shape, and the fail-closed abort.
- End-to-end: pointing `SHEPHERD_PRETTIER_MODULE` at a bogus path **fails** (proving the override is honoured, not ignored), and the real pin with `cwd = SERVER_INSTALL_ROOT` reports "manifest up to date" — a real write run leaves the committed file byte-identical.


scoop (commented):
# Review — `fix(doc-agent): regen the docs manifest a doc edit makes stale` (#2163)

## What the PR does, and whether it satisfies the task

The task asked for a line in the glossary rule about the downstream `docs-manifest` regen; the plan
widened that to also close the hole structurally in the doc agent. Both halves are present:

- `.claude/rules/ui-glossary.md:21-23` and `.claude/rules/ui-feature-catalog.md:31` add the note.
- `src/doc-agent.ts:107-141, 947-963` regenerate + stage `ui/src/lib/docs-manifest.ts`.
- `ui/scripts/gen-docs-manifest.ts:141-143, 154` add the prettier-pin seam.

### Accuracy of the prose (verified, not assumed)

- "derives … from every docs page's frontmatter `description` + its H2/H3 headings" — verified
  against `ui/scripts/gen-docs-manifest.ts:106` (`keywordsOf(fmValue(fm,"description") ?? "",
  extractHeadings(body))`), `:77-84` (`keywordsOf`) and `:58-64` (`extractHeadings` matches
  `/^#{2,3}\s+/`). The issue text said "the page's body"; the rule text correctly says
  headings + description instead. Good catch by the author.
- "The docs-site glossary page gives each term its own `###` heading" — verified against
  `docs-site/src/content/docs/reference/glossary.md:18,25,37,46,…` (every term is an `###`).
- "`sync-docs.mjs` publishes `CLAUDE.md` and every `.claude/rules/*.md`" — verified against
  `docs-site/scripts/sync-docs.mjs:48-131`: `PAGES` holds 9 entries, including `CLAUDE.md` and all
  five files that `.claude/rules/*.md` currently glob to (`ui-design-system.md`, `i18n.md`,
  `ui-feature-catalog.md`, `ui-glossary.md`, `herdr-version-bump.md`). And
  `ui/scripts/gen-docs-manifest.ts:112-125` reads each `PAGES` entry's committed source for its
  headings, so the claim that editing a heading in the rule file itself drifts the manifest holds.
- "fails in `verify`" — verified against `.github/workflows/ci.yml:118` and
  `scripts/pre-push.ts:559`.
- **This PR does not itself drift the manifest**: both additions are body prose with no new
  `##`/`###` heading (`.claude/rules/ui-glossary.md:21,23`,
  `.claude/rules/ui-feature-catalog.md:31`), and descriptions for those two pages live in
  `docs-site/scripts/sync-docs.mjs:110-121` (untouched). Consistent with the diff shipping no
  `ui/src/lib/docs-manifest.ts` change.

### Correctness of the doc-agent wiring (verified)

- Ordering is right: prettier → `git add -- <inScope>` (`src/doc-agent.ts:927`) → read staged
  (`:928`) → regen (`:955`) → `git add -- ui/src/lib/docs-manifest.ts` (`:961`) → re-read (`:962`).
  Running the regen *after* the in-scope `git add` matters, because `build()` enumerates pages via
  `git ls-files` (`ui/scripts/gen-docs-manifest.ts:90-95`), which reads the **index** — so a page
  the agent just edited is seen, and its body is read from the working tree at `:98`.
- The commit is `git commit --no-verify -m …` with no pathspec (`src/doc-agent.ts:1073`), so the
  staged manifest does reach the commit. Verified.
- `computeDocAgentOutcome` still maps the renamed flag to `"error"` (`src/doc-agent.ts:348-351`),
  and the `prettierFailed` → `stageFailed` rename is complete: `grep prettierFailed` over the whole
  tree returns nothing, and every `stageFailed` site is inside `src/doc-agent.ts` (348, 351, 972,
  974, 984, 991, 995, 1000, 1004). No out-of-diff consumer.
- `DOCS_MANIFEST` = `ui/src/lib/docs-manifest.ts` and `DOCS_MANIFEST_GENERATOR` =
  `ui/scripts/gen-docs-manifest.ts` both resolve — confirmed present in the tree
  (`ui/package.json:17-18` runs the latter; `ui/src/lib/docs-manifest.ts:2` is the former).
- `prettierModuleSpecifier` is exported from `ui/scripts/gen-docs-manifest.ts:141` and imported at
  `ui/scripts/gen-docs-manifest.test.ts:8` — spelling/casing match. `NodeJS.ProcessEnv` is
  available: `ui/package.json:31` has `@types/node`, and the file already relied on the same global
  typings for `process.argv` / `process.exit` (`ui/scripts/gen-docs-manifest.ts:164,176`).
- `prettier` resolves from the root install for CI/humans — `ui/package.json` has no prettier of its
  own, `package.json:47` pins `^3.9.6`, and `bun.lock:398` resolves `prettier@3.9.6`. So the default
  `"prettier"` branch is unchanged behaviour.

## Findings

### 1. `src/doc-agent.ts`: the server now executes agent-/PR-writable code from the worktree, with the full server environment

`defaultGenDocs` (`src/doc-agent.ts:136-141`) spawns

```
process.execPath run <worktreePath>/ui/scripts/gen-docs-manifest.ts
  cwd: SERVER_INSTALL_ROOT
  env: { ...process.env, SHEPHERD_PRETTIER_MODULE: … }
```

That runs a **worktree-resident TypeScript file** — plus everything it imports, notably
`<worktreePath>/docs-site/scripts/sync-docs.mjs` (`ui/scripts/gen-docs-manifest.ts:27`) — as a
direct child of the Shepherd server process: outside the membrane/sandbox, with the server's whole
environment (API keys, forge tokens, `SHEPHERD_*` config), and with `cwd` set to the server's own
install root.

Both of those files are writable by untrusted parties on the two paths that reach `finalize()`:

- **The doc agent itself.** `src/transient-agent-argv.ts:171` gives the `doc` preset bare `Write`
  and `Edit` (the comment at `:154-158` explains scoped `Write` is silently denied under `dontAsk`,
  and explicitly classifies doc-agent input as *UNTRUSTED*: "a prompt-injection hidden in the input
  must not exec"). Nothing restores out-of-scope agent edits before staging — I grepped
  `src/doc-agent.ts` for `checkout` / `restore` / `clean` and there are no matches, and
  `stageInScope` (`:908-931`) only `git add`s the in-scope list. So a doc agent that writes a
  payload into `ui/scripts/gen-docs-manifest.ts` has it executed by the server moments later.
- **The PR under review, in retarget mode.** `beginRetarget` creates the worktree at the code PR's
  **HEAD sha** (`src/doc-agent.ts:625, 639-649`, and the docstring at `:606-609`). So the PR's own
  committed copy of the generator is what gets run. No injection needed — committing the payload is
  enough.

This is a new capability, not a restatement of the existing prettier path: `defaultPrettierWrite`
(`src/doc-agent.ts:78-91`) executes the **server's own** pinned binary
(`join(SERVER_INSTALL_ROOT, "node_modules", ".bin", "prettier")`) over agent-authored *data*. The
new call executes agent-authored *code*. It also silently widens the boundary that
`src/doc-agent.ts:197-200` calls "the structural off-limits boundary that backs the prompt-level
instruction".

The fail-closed `DocsManifestError` does not help here — a payload that succeeds returns exit 0.

Ways to close it (any one is sufficient for the exec problem):

- Run the **server's** copy of the generator and tell it which repo root to target — e.g. an
  explicit `--repo-root <worktreePath>` / env override on the generator, instead of deriving
  `repoRoot` from `import.meta.url` (`ui/scripts/gen-docs-manifest.ts:33`). This keeps the pinning
  argument intact and removes the worktree from the trust path entirely.
- Or refuse to run when the generator or its imports differ from a trusted ref — e.g.
  `git diff --quiet <baseRef> -- ui/scripts/gen-docs-manifest.ts docs-site/scripts/sync-docs.mjs`,
  skipping the regen (log, don't fail) when they differ. Note this must compare against the **base**
  ref, not the worktree HEAD, or the retarget case is unprotected.
- Or run it inside the same sandbox membrane the agent gets.

Independently of which you pick: pass an explicitly-constructed minimal env rather than
`...process.env`, so the subprocess is not handed the server's credentials.

## Nits (non-blocking):

- `src/doc-agent.ts:197-200` — the `IN_SCOPE_PATHS` comment still says "finalize() stages ONLY these", which this PR made inexact; `stageInScope`'s own docstring (`:904-906`) was updated, so it is only this one that now reads as more absolute than the code is.
- `.claude/rules/ui-feature-catalog.md:31` — "Add or rename a heading (or change a `description`) in any of them" is slightly off for the `CLAUDE.md` / `.claude/rules/*.md` pages: their `description` lives in `docs-site/scripts/sync-docs.mjs` `PAGES` (`:48-131`), not in the page itself, so editing the page cannot change it.
- `.claude/rules/ui-glossary.md:21` — "The doc agent does this itself; a hand edit does not" reads a little cryptically out of context; "…so this note is for hand edits" would land faster.
- `src/doc-agent.ts:136` — `execFileP` here has no timeout (matching the existing prettier calls at `:49` and `:85`), so a wedged generator hangs `finalize()` for the run.

- `src/doc-agent.ts:136-141` — the prettier pin only takes effect if the **worktree's** generator honours `SHEPHERD_PRETTIER_MODULE`. If the doc agent ever runs against a base ref predating this commit, the worktree generator falls back to the bare `import("prettier")` at a path with no `node_modules` ancestor and Bun auto-installs an unpinned prettier — the exact skew the header comment at `ui/scripts/gen-docs-manifest.ts:130-140` says the pin exists to prevent, and it fails *open* (exit 0, differently-formatted manifest) rather than tripping `DocsManifestError`.

- `ui/scripts/gen-docs-manifest.ts:8` — the header says "the 4 reference pages that `syncDocs()` writes"; `docs-site/scripts/sync-docs.mjs:48-131` now has 9 `PAGES` entries.

## Stated limitations

- I could not verify that `<SERVER_INSTALL_ROOT>/node_modules/prettier/index.mjs` exists as a file: this review worktree has no `node_modules` installed (`ls node_modules/prettier` is empty) and no copy of the package is present anywhere under `/home/patrick` (glob for `**/node_modules/prettier/index.mjs` found nothing). Prettier 3 does ship `index.mjs` as its ESM entry, so this is assumed correct as written; note only that it is a deep path past the package's `exports` map, unlike the `.bin/prettier` entry the existing code uses at `src/doc-agent.ts:48,84`.
- I did not run the suite or lint; per instructions, CI covers those.

PR #2170, opened 2026-09-02
feat(usage): first-push CI green rate (#2159)

### LINKED ISSUE #2159: First-push CI green rate: retain the first check conclusion per pushed head
Follow-up to #2151 / #2153. Deferred at the R1 plan gate.

Of the playbook's delivery indicators, first-push CI green rate is the one that needs **new instrumentation**: no per-push check result is retained anywhere today. Everything else was derivable (see #2153's `delivery_facts` + `reviewer_spawns.outcome`).

Needs a hook where the git poller observes a head's checks flipping terminal, recording the FIRST conclusion per pushed head with a per-head dedup key so a re-poll cannot overwrite it. Then feed it into `buildDeliveryMetrics` (`src/delivery-metrics.ts`) as another `DeliverySample` and add a tile to the Delivery lens.

Deliberately kept out of #2153 to keep that PR's blast radius off the git poller.

Note: like the rest of the delivery instrumentation, this would be forward-only.

scoop:
Closes #2159.

Follow-up to #2151 / #2153, deferred at that plan gate. Adds the one playbook delivery indicator
that needed new instrumentation: **no per-push CI result was retained anywhere.**

## What lands

Each task session now retains the **first terminal CI conclusion Shepherd observed**, frozen so no
later poll can overwrite it, and the Delivery lens renders it as a `First-push CI green` tile.

## The hook already existed — the poller is untouched

The issue sketched "a hook where the git poller observes a head's checks flipping terminal".
`PrPoller.refresh()` already emits `session:git` whenever `gitStateChanged()` sees `checks` *or*
`headSha` change, `GitState` already carries `headSha` plus the worst-of rollup, and
`DeliveryFactsService.onGit()` is already subscribed. So this extends the existing observer and
**`src/pr-poller.ts` is not touched**, keeping the blast radius exactly where #2153 left it.

## Storage: two columns, not a new table

`delivery_facts` gains `firstCiHeadSha` + `firstCiConclusion`. `upsertDeliveryFact` is already
first-write-wins (`COALESCE(existing, new)`), so "a re-poll cannot overwrite it" is enforced in
SQL — across restarts — with no join and no second retention sweep (`pruneDeliveryFacts` already
covers the row). The sha and the conclusion move as one unit, so a later push can never leave the
sha of push #2 beside the conclusion of push #1. A separate per-head table was considered and
dropped: the metric only ever reads each task's first head, and an archive-decoupled table would
need its own retention sweep. Trade-off accepted: per-head history for later pushes is not kept.

## Eligibility

Stamps when a `headSha` is present **and** the rollup is terminal (`success` / `failure`), on any
PR state. `pending` is mid-flight; `none` is ambiguous — a GitHub repo with zero workflows sits at
a permanent `none` — so excluding it also keeps no-CI repos out of the metric with no extra
`noCi` guard. Not gated on `state === "open"` on purpose: that would drop tasks opening and merging
between two polls, which are green by construction, biasing the rate pessimistic.

## Metric

Per merged task in the window, the retained conclusion. Denominator = tasks carrying a conclusion;
a task with none is **excluded, never counted as a pass** — same discipline as `firstPassRate` and
`planDriftRate`. Computed inside `statsFor`, so totals and per-repo rows cannot drift apart.

## The caveat, stated rather than hidden

The first head Shepherd *observes* is not provably the true first push: a session can push several
times before its PR opens, the server can be down across the first run, and a head superseded
before its rollup went terminal never stamps. The metric is therefore "the first CI conclusion
Shepherd saw" — said on the type, in the glossary definition, and qualified by the tile's sample
size. Forward-only, so existing rows read "never observed" and the rate starts empty rather than
wrong.

## Verification

- root `bun run lint` · `bun run test` (8927 pass) · `bunx tsc --noEmit`
- `ui/`: `bun run check` (0 errors) · `bun run test` (4753 pass) · `check:i18n` · `check:glossary` ·
  `check:docs-manifest`
- `bunx fallow audit --base origin/main --fail-on-issues` clean (`onGit` was refactored into
  `wantedFrom` / `mergeStamped` / `write` to stay under the cognitive-complexity gate)

## Notes for review

- **`[no-feature-entry]`** is in the commit body: the tile lands inside the Delivery lens, which
  already announced itself in v1.48.0, and the operator chose no separate What's-New entry. Say so
  if you would rather have one — it is a fragment file plus two keys per locale.
- Per the glossary rule, the EN + DE definitions of `first-push-green` need an explicit confirmation
  that they are accurate and well phrased; no gate can catch a misleading definition.


scoop (commented):
# Review — feat(usage): first-push CI green rate (#2159)

**Decision: comment.** No blocking correctness, security, contract, or locale defect found. The
change does what the issue asks: it retains the first terminal CI conclusion per task session
(frozen at the SQL level), folds it into `buildDeliveryMetrics` as a `DeliverySample`, and renders
one more tile in the Delivery lens.

### Freeze semantics (the core of the issue's "a re-poll cannot overwrite it")

- `src/store.ts:4275-4281` — the upsert keeps `firstCiConclusion` via
  `COALESCE(delivery_facts.firstCiConclusion, excluded.firstCiConclusion)` and gates
  `firstCiHeadSha` on `CASE WHEN delivery_facts.firstCiConclusion IS NULL`. In SQLite's
  `ON CONFLICT DO UPDATE`, every right-hand-side reference to `delivery_facts.X` reads the
  **pre-update** row, so the `CASE` guarding `firstCiHeadSha` and the `COALESCE` on
  `firstCiConclusion` both see the same old conclusion regardless of `SET`-clause order — the pair
  can never split. The `delivery_facts.X` / `excluded.X` idiom is the one already in use at
  `src/store.ts:4271-4274` for `issueNumber`/`prNumber`/`prOpenedAt`/`mergedAt`, so it is
  established-correct here.
- I walked the full poll sequence against `src/delivery.ts:92-135`: pending on head A (no stamp) →
  failure on A (`isNew` true via `want.ci && !seen.ci`, writes `A`/`failure`) → pending on B (all
  three `want` flags already seen, early return) → success on B (`seen.ci` true, early return) →
  merged (`want.merged` newly true, writes with `mergedAt`, but the `CASE`/`COALESCE` keeps
  `A`/`failure`). Correct in both memory and store.
- The in-memory `stamped.ci` is only an optimisation; the restart path is covered by the store, and
  `test/delivery-facts.test.ts:200-207` asserts exactly that by calling `svc.forget()` between
  observations. `src/store.ts:3042-3048` (the `archive(id,'merged')` stamp site) passes neither CI
  field, so `excluded.*` is NULL there and the `COALESCE`/`CASE` preserve whatever the poller
  already recorded — no blanking on the merge-train teardown path.

### The observation gate

- `ciObservation` (`src/delivery.ts:22-26`) requires `git.headSha` and a terminal rollup. I checked
  the claim that a zero-workflow GitHub repo is permanently excluded: `worstChecks`
  (`src/forge/github.ts:518-523`) and `rollupChecks` (`src/forge/checks.ts:166-179`) both return
  `"none"` for an empty rollup, and `ChecksState` (`src/forge/types.ts:128`) has no other terminal
  value — so the comment's justification for omitting a `noCi` guard holds, and no
  `checks-gate.ts` change was needed.
- `headSha` and `checks` always originate from the **same** `forge.prStatus` / batch payload —
  `src/pr-poller.ts:508-510` and `src/pr-poller.ts:477`, and every `session:git` emitter in
  `src/server.ts:3999`, `:4078`, `:4104`, `:4273` spreads one `prStatus` result. `annotateHandoff`
  (`src/repo-roles.ts:196-200`) passes `checks` and `headSha` through untouched (it only
  back-fills `reviewerStates`). So a conclusion can never be pinned to a head it did not belong to.
- `guardStaleTerminal` (`src/pr-poller.ts:32-48`) rewrites a reused-branch-name terminal hit to
  `{state:"none", checks:"none"}` with no `headSha`, so a name-collision PR cannot stamp.

### Metric

- `statsFor` (`src/delivery-metrics.ts:161,180-183`) excludes null conclusions from the denominator,
  matching the discipline of `planDriftRate` at `:158,173-176` and `firstPassRate` at `:152,166`.
  `rate()` (`:85-87`) returns `{value:null,n:0}` on an empty denominator, so an unmeasured window
  reads as an em dash, not `0%`. Because it lives inside `statsFor`, the per-repo rows
  (`:187-201`) and the totals fold from the same code — they cannot drift.
- `listMergedDeliveryFacts` uses `SELECT *` (`src/store.ts:4301-4308`), so the two new columns
  reach `DeliveryFact` with no query change needed.
- The migration (`src/store.ts:4710-4717`) follows `migrateReviewColumns` (`:4720`) verbatim —
  `PRAGMA table_info` + per-column `ALTER TABLE` — and sits inside the constructor's single
  transaction (`src/store.ts:1312`, immediately after the `CREATE TABLE IF NOT EXISTS`), same
  position as the `migrateReviewColumns` call at `:1326`. `PRAGMA table_info` is a read, so it is
  fine inside a transaction.

### Identifiers, locale parity, consumers

- `usage_delivery_first_push` exists in **both** catalogs at the same position:
  `ui/messages/en.json:3699` and `ui/messages/de.json:3699`. `gloss_first_push_green_term` /
  `gloss_first_push_green_def` likewise at `en.json:3720-3721` and `de.json:3720-3721`.
- The marker id `first-push-green` in both catalog values resolves to the registry entry at
  `ui/src/lib/glossary.ts:87-92`, and the `gloss_<id>_term` / `gloss_<id>_def` key convention the
  gate documents (`scripts/check-glossary.mjs:4-5`) is respected — checked against the marker scan
  at `scripts/check-glossary.mjs:130-149` and the key-presence loop at `:106-117`.
- Tile labels already render through `GlossaryText` (`ui/src/lib/components/usage/DeliveryLens.svelte:134`,
  pre-existing) so the `[[…|…]]` marker will not print raw — same treatment as
  `usage_delivery_lead_time` (`ui/messages/en.json:3703`) and `usage_delivery_plan_drift` (`:3738`).
- `extension/messages/en.json` carries no `usage_delivery_*` / `gloss_first_pass_*` keys, so there
  is no third catalog to keep in parity.
- Regenerated manifest verified by hand against the generator: `keywordsOf` lowercases the
  frontmatter description plus every H2/H3 in order (`ui/scripts/gen-docs-manifest.ts:57-63,76-83`).
  The headings in `docs-site/src/content/docs/reference/glossary.md` run `### First-pass rate` (:64)
  → `### First-push CI green` (:70) → `### Plan drift` (:78), which is exactly where
  `first-push ci green` was inserted in `ui/src/lib/docs-manifest.ts:84`.
- Consumers of the widened `DeliveryStats`: the only two places that construct one as a literal are
  `ui/src/lib/components/usage/DeliveryLens.browser.test.ts:16-30` (updated) and
  `ui/src/lib/components/Usage.browser.test.ts:103-113` (an **un-annotated** inline object that
  already omits `planDriftRate`/`planDriftMajor` from #2155, so it is not type-checked against the
  interface and does not break). `scripts/delivery-report.ts:101` only reads fields, so widening the
  interface compiles there — see the nit below.
- The tile-index shifts in the browser test are consistent with the new order
  `[first-pass, first-push, rework, plan-rework, plan-drift, ttfr, lead, merged]`: `[4]` → `[5]`/`[6]`
  for ttfr/lead and `[3]` → `[4]` for plan drift, matching `tilesFor` at
  `ui/src/lib/components/usage/DeliveryLens.svelte:52-101`. `formatPct`
  (`ui/src/lib/components/usage/format.ts:25-27`) renders `0.75` → `"75%"` and `0.4` → `"40%"`, so
  the expected strings are right.
- The `feat(...)` + `ui/src/lib/components/**` combination would trip
  `scripts/check-feature-catalog.sh:36-40`; the commit body carries the documented
  `[no-feature-entry]` opt-out token, so the gate is satisfied deliberately rather than by accident.

### Scope

The diff touches nothing in `src/pr-poller.ts`, `src/poller.ts`, or any forge adapter — the plan's
stated blast-radius boundary holds. No new dependency, no new public API beyond the two additive
type fields, no weakened validation.

## Nits (non-blocking)

- src/delivery-metrics.ts: adds `firstPushGreenRate` to `DeliveryStats`, but the operator-facing CLI
  report's indicator table (`scripts/delivery-report.ts:101-120`) still lists only the pre-#2159 set, so `bun scripts/delivery-report.ts` silently omits the new indicator — note that the sibling change #2155 did add its `Plan drift rate` row there (`scripts/delivery-report.ts:111-115`), so this is a departure from the established pattern rather than a pre-existing gap.
- test/delivery-facts.test.ts: the case named "a CI conclusion alone records a fact for a PR-less-looking payload" also passes `state: "merged"`, which sets `want.merged` and would have written a row under the old early-return too — so the new `!want.ci` term in the `src/delivery.ts:103` guard is never exercised in isolation; dropping `state: "merged"` (and asserting via a direct store read rather than `listMergedDeliveryFacts`) would actually cover it.
- src/delivery.ts: `Stamped` is documented at `:58` as "what has already been persisted" but is now also the return type of `wantedFrom` at `:68`, i.e. "what this observation wants"; two names for the two roles would read better than one shared shape.

## Stated limitations

- I could not verify the exact bytes `bun run gen:docs` emits by executing it (read-only review, no
  build). I reconstructed the expected `keywords` string by hand from
  `ui/scripts/gen-docs-manifest.ts:57-63,76-83` plus the heading list in `glossary.md`, and it
  matches the committed line at `ui/src/lib/docs-manifest.ts:84`.
- GitHub's real-world rollup behaviour for a head whose workflows are cancelled mid-run (the
  "superseded head never goes terminal" risk the plan documents) cannot be verified from this repo;
  I confirmed only the local mapping in `src/forge/checks.ts:166-179`. The behaviour is disclosed on
  `DeliveryFact.firstCiConclusion` (`src/types.ts:1017-1026`) and in the glossary definition rather
  than hidden, which is the right call.

PR #2173, opened 2026-09-02
feat(newtask): optional "shape this" clarifying round + issue-template prescription (#2158)

### LINKED ISSUE #2158: Optional structured intake for New Task: a 'shape this' clarifying round
Follow-up to #2151 (R6, `docs/research/ai-native-sdlc-playbook.md` §3 R6). Lowest-ranked of the six; R1 shipped in #2153.

Optional structured intake for New Task:

- a "shape this" clarifying round that reuses the epic-shaping interview machinery, and
- issue templates prescribed by the readiness analyzer.

**The freeform fast path stays.** This is opt-in shaping, not a mandatory funnel — the point is to catch under-specified tasks before an agent burns a session on them, not to add ceremony to every task.

scoop:
Closes #2158. R6 of the AI-native SDLC playbook (`docs/research/ai-native-sdlc-playbook.md` §3 R6), both halves. Plan-gated and reviewed before implementation.

## (a) "Shape this" — an optional clarifying round on New Task

A **Shape** button beside attach/dictate runs a transient agent over the rough prompt and the repo. It returns a draft brief plus the 2–5 questions only the operator can settle, rendered through the **plan gate's own `QuestionFormBlock`** (given a third mode: hand answers to the parent instead of POSTing them to a session). Answering replaces the textarea with an intent-shaped brief — Problem / Outcome / Constraints / Non-goals / Clarifications — which starts as an ordinary code task.

No Shepherd session, worktree, or branch exists until Start is pressed: that is the point of R6 — catch a misshapen ask before a session burns on it, rather than after the plan gate does. **Nothing runs unless the button is clicked**; the freeform fast path is byte-for-byte what it was.

Not a new session mode, no `task_brief` MCP tool, no draft/approve state machine, no DB table. Two routes: `POST /api/shape` (spawns the helper, 255s idle budget so Bun doesn't sever the socket mid-round) and a pure `POST /api/shape/brief` (composition only).

## (b) Readiness prescribes the matching issue template

A new `issue_templates` guardrail (weight 4, forge-shaped detector covering GitHub's template directory, the legacy single file, and GitLab's) plus the generated template as a copyable artifact next to the house-rules pane; it rides the one-click adopt task when the repo lacks one. **This shifts every repo's readiness score**, since the guardrail joins the weight denominator — intended, and the all-present fixtures were updated so "100" still means something.

Both artifacts derive from one section list in `src/intent-shape.ts`. Composition lives server-side for exactly that reason: `ui/` cannot import root `src/`, so a client-side composer would be a second copy of the intent shape free to drift from the template.

## Repo access is `--add-dir`, not the allowlist

The plan first claimed the helper read the repo "via the `writer-ro` preset". It does not: under `--permission-mode dontAsk` with a temp cwd, an agent reaches nothing outside that dir — the allowlist decides which tools may run, not where they may reach. The plan reviewer caught it. So the helper now gets `--add-dir <repoPath>` through a new opt-in `addDirs` on `buildTransientAgentArgv`, emitted early (ahead of `--settings`) because the flag is variadic and would otherwise swallow the trailing prompt positional. Existing callers pass none, so their argv stays byte-identical.

That flag also opens the repo to the preset's bare `Write` — Claude Code has no read-only form of it — and the directory here is the operator's real checkout, not a disposable worktree. **An earlier revision of this description understated what that buys**; stated plainly, the agent can overwrite a tracked file (destroying uncommitted work; `git status` shows it only after the fact) or write a gitignored path `git status` never surfaces at all, and `disableAllHooks: true` rules out the tool-guard hook. The prompt's "write only your result file" line is guidance to a cooperative model, not an enforced boundary.

What bounds the blast radius is narrower than "nothing bad can happen": no `Bash`, no `Edit`, no git and no network on the allowlist, so there is no exec, commit, push or exfiltration — the ceiling is file writes inside one repo the operator chose, by a helper the operator started, on a prompt the operator wrote. If that ceiling is too high, the fix is a read-only view of the repo (detached worktree / RO bind, as the reviewer roles get), not a tighter allowlist. `src/task-shape.ts` and `docs/sandbox-security.md` both say this now. Codex needs no flag (`workspace-write` restricts writes and network, not reads).

## Containment and failure behaviour

- `repoPath` is `safeRepoDir`-checked by the route before any agent is granted access to it.
- Fails closed in api-key mode with no key; every failure (spawn, timeout, malformed output, compose) shows a distinct message and leaves the typed prompt untouched.
- The helper's herdr tab and temp dir are torn down per round, `SHAPE_LABEL` is registered with the tab reaper and boot-reaped — the #1852 gap, closed up front rather than retrofitted.
- Both routes sit behind the operator auth gate and are unreachable from the agent-ingress listener (its allowlist is an explicit map of session leaf routes).
- A dismissed or superseded round can't land its late result on the card (monotonic round token).
- Question text is capped before validation, since the helper reads repo files it did not write.

## Verification

Root `bun run lint` + `bun run test` (8963 pass), `bunx tsc --noEmit`, `ui/ bun run check` + `bun run test` (4776 pass), `check:i18n` at parity (EN+DE for every new string), `fallow audit --base origin/main --fail-on-issues` clean, pre-push gates green.

New coverage: `test/intent-shape.test.ts`, `test/task-shape.test.ts` (no real spawn — injected deps), `test/server-shape.test.ts`, `addDirs` placement + byte-identity in `test/transient-agent-argv.test.ts`, issue-template detection in `test/readiness.test.ts`, `shape.test.ts`, `ShapeRound.browser.test.ts`, and the New Task round end-to-end. A pinning test asserts `QuestionFormBlock` with `answerCtx` still posts to the plan gate.

## Known limitations

- The Shape control is hidden on the mobile layout: the in-field toolbar's 375px budget is one labelled control beside the attachment chips, and a second wraps the row (pinned by an existing no-wrap test). Shaping is optional by design, so mobile keeps the freeform path.
- The readiness analyzer stays file-based, so a local-only (LocalForge) repo is still told it lacks issue templates — as it is already told it lacks Dependabot.


scoop:
# Review — optional structured intake for New Task (#2158)

Scope: `git diff 037c062db011de8bdff3c5e62d1dc04a0c5b1fe9...HEAD` (29 files).

The shape of the change matches the task: the freeform path really is untouched (the whole round is
behind a button rendered only in `code` mode — `ui/src/lib/components/NewTask.svelte:1957` — and
nothing on the start path was altered), the transient helper faithfully copies the recommender's
lifecycle, and the readiness half wires a new guardrail through both ecosystem profiles. Three
defects below.

## What I verified

**Route + auth.** `handleShape` / `handleShapeBrief` are registered in `ROUTE_HANDLERS`
(`src/server.ts:7909-7910`) and are NOT reachable from the agent-ingress listener:
`isAgentIngressRoute` (`src/server.ts:8017-8018`) rejects anything whose `parts[1] !== "sessions"`,
so both routes require the operator cookie/token via `checkAuth` (`src/server.ts:7928-7931`).
`safeRepoDir` (`src/validate.ts:906-924`) realpaths and containment-checks the path before it
becomes `--add-dir`, so the argv grant cannot be pointed outside `config.repoRoot`. `/api/shape` is
added to `slowRequestTimeoutSec` (`src/server.ts:8203`) at 255s, comfortably above the 180s poll
budget in `src/task-shape.ts:264`.

**Imports/identifiers.** `composeTaskBrief` and `normalizeRound` resolve to `src/intent-shape.ts:87`
and `src/task-shape.ts:202`; `resolvePlanAnswers` / `RawAnswer` were already imported at
`src/server.ts:69`, `safeRepoDir` at `src/server.ts:46`. `SHAPE_LABEL` (`src/task-shape.ts:47`,
`"shape "`) is registered in `isShepherdHelperLabel` (`src/tab-reaper.ts:86`) and boot-reaped
(`src/index.ts:2505`); `reapTransientByLabel` matches by tab-label prefix
(`src/transient-tab-reaper.ts:62`), so the wiring is right and collides with no other label
(`src/verify-key.ts:17`, `src/autopilot.ts:40`, `src/prompt-recommend.ts:28`, `src/namer.ts:22`,
`src/doc-agent.ts:163`).

**`--add-dir` placement.** `src/transient-agent-argv.ts:268` emits it directly after
`--session-id <uuid>`, where the next token is `--settings` (`:269`) — never before the trailing
prompt positional at `:286`. The claim in the new comment holds. Omitting `addDirs` leaves the array
byte-identical for every existing caller.

**Question validation.** `normalizeRound` feeds the helper's questions through the plan gate's own
`parseVisualBlocks` (`src/visual-blocks.ts:585`) → `validateQuestionForm` (`:533`) →
`validatePlanQuestion` (`:515`), so an option-less `single`, a missing `kind` or a duplicate id is
dropped by the same validator; `resolvePlanAnswers` (`src/plan-gate.ts:1810`) is fail-closed on the
way back. `answerText` (`src/intent-shape.ts:78-82`) covers all three `ResolvedAnswer` kinds
`resolveOne` can produce (`src/plan-gate.ts:1856-1869`).

**QuestionFormBlock third mode.** The plan-gate branch is genuinely untouched: `submit()` returns
early into the callback path only when `onanswer` is set
(`ui/src/lib/components/blocks/QuestionFormBlock.svelte:87-91`), `answerCtx` still gates the "sent"
confirmation (`:163`), and the post-emit lock falls out of `locked` → `canSubmit` (`:47`, `:51-52`).
`RawAnswer` is imported at `:2`.

**i18n parity.** All 24 new keys exist in both catalogs, at matching offsets
(`ui/messages/en.json:3746-3769`, `ui/messages/de.json:3746-3769`), and those two are the only
catalogs in `ui/messages/`. `shape_questions_count` carries `{count}` in both (`en.json:3751`,
`de.json:3751`). The feature-catalog entry
(`ui/src/lib/feature-announcements/entries/v1.48.0-shape-task.ts`) is auto-globbed
(`ui/src/lib/feature-announcements.ts:47-50`), its `sinceVersion` 1.48.0 is the next unreleased
version (`package.json:3` = 1.47.0), and its `targetId: "nt-shape"` matches the `use:coachTarget`
anchor at `NewTask.svelte:1961`.

**Readiness report shape.** `hasIssueTemplates` / `issueTemplate` are set on both the applicable
(`src/readiness.ts:669`, `:671`) and not-applicable (`:689`, `:691`) paths; the UI mirror
(`ui/src/lib/types.ts:1823-1830`) matches, and no other producer of `ReadinessReport` exists (grep:
only `src/readiness.ts`, `ui/src/lib/types.ts`, `readiness-view.ts`, `ReadinessPanel.svelte`,
`ui/src/lib/api.ts`, which passes the payload straight through at `ui/src/lib/api.ts:2147-2149`).
`issue_templates` has entries in all six exhaustive `Record<GuardrailId, …>` tables
(`src/readiness.ts:775`, `:797`, `:864`, `:881`, `:899`, `:946`), so the compiler enforces
exhaustiveness. The new template pane renders only inside the `report.applicable` branch
(`ReadinessPanel.svelte:232`, `:310-323`), so the `""` not-applicable value never reaches a `<pre>`.

---

## Findings

### 1. `ui/src/lib/components/NewTask.svelte`: the raw `"default"` model value is forwarded to the CLI as a model alias

`runShape` passes the picker's model state straight through:

```
ui/src/lib/components/NewTask.svelte:851
const res = await apiShapeTask(repoPath, prompt, agentProvider, model);
```

`"default"` is a real, selectable value of that state, not a theoretical one:

- it is an explicit `<option>` in the settings rail — `ui/src/lib/components/new-task/RunSettingsGroups.svelte:180`;
- `normalizeRunConfig` snaps an unavailable model to it — `ui/src/lib/components/new-task/run-config.ts:81` and `:96`, applied unconditionally by the `$effect` at `NewTask.svelte:717-730`;
- `preselectModel` returns it for a Fable seed with Fable unavailable — `run-config.ts:20`, pinned by `run-config.test.ts:50-55`.

Every other consumer translates it to `null` at the API boundary before it can become a flag:
`NewTask.svelte:1423` (`model: model !== "default" ? model : null`) on the start path, and
`ui/src/lib/components/new-task/ModelCliPicker.svelte:91` on confirm. The shape path is the only one
that does not, and nothing downstream repairs it — the route accepts any non-empty string
(`src/server.ts:6571-6572`), `src/index.ts:3154-3164` passes it verbatim, and
`src/transient-agent-argv.ts:275` is `if (opts.model) argv.push("--model", opts.model)`. The Codex
branch behaves the same way (`src/codex-role-argv.ts:35`, `-m default`).

Result with the Default model selected: the helper spawns as `claude --model default …`. It writes
no `.shepherd-shape.json`, `pollForRound` (`src/task-shape.ts:238-251`) burns the full 180s budget
(`:264`), and the operator is shown `shape_err_timeout` — "didn't finish in time"
(`ui/messages/en.json:3759`) — for what is actually a malformed argv. Compare `src/service.ts:2883`
and `:3119`, which push `--model` only after the value has been normalised to null/absent.

Fix: normalise at the same seam the start path uses — send `null` (or omit) for `"default"` and let
the route/`shapeTask` accept a null model; `buildTransientAgentArgv` already treats a falsy model as
"inherit the spawn default".

### 2. `ui/src/lib/components/ReadinessPanel.svelte`: the already-has-templates note states the opposite of what the adopt task does

When the repo already ships issue templates the panel shows:

```
ui/src/lib/components/ReadinessPanel.svelte:319-320
{#if report.hasIssueTemplates}
  <div class="merge-note">{m.readiness_template_present_note()}</div>
```

whose copy is "This repo already ships issue templates; **the task will add this one alongside
them**." (`ui/messages/en.json:3766`; German equivalent "die Aufgabe legt diese zusätzlich an",
`ui/messages/de.json:3766`).

But that is exactly the branch in which the template is dropped from the adopt task entirely:

```
ui/src/lib/components/ReadinessPanel.svelte:160-163
const template = report.hasIssueTemplates
  ? undefined
  : { intro: m.readiness_adopt_template_intro(), body: report.issueTemplate };
onadopt(repoPath, buildAdoptPrompt(intro, report.claudeMd, template));
```

and `buildAdoptPrompt` omits the whole section on `undefined`
(`ui/src/lib/components/readiness-view.ts:42-45`). So "Send to task" produces a prompt with no
template at all, while the note beside the Copy button promises the opposite. The sibling
house-rules pane has no such gap: `hasAgentInstructions` swaps the intro to the *merge* variant
(`ReadinessPanel.svelte:155-157`, `ui/messages/en.json:943`) and its note truthfully says "the task
will merge, not overwrite" (`ui/messages/en.json:938`).

The same wrong claim is in the server-side doc comment for the field — `src/readiness.ts:63`:
"Whether the repo already ships issue templates (so the prescription says merge, not create)". The
prescription does not say merge; it says nothing.

Fix either side (carry the template with a merge-style intro, or reword both catalogs and the
`readiness.ts` doc to say the template is left out because the repo already has one), but the copy
and the behaviour must agree in EN and DE.

### 3. `src/intent-shape.ts`: `INTENT_SECTIONS` is documented as the single source both artifacts derive from, but the brief composer hard-codes its headings

```
src/intent-shape.ts:31
/** The intent shape, in order. Edit HERE — every artifact below is derived from it. */
```

and the module header at `:59-67` states "One section list, two consumers that must not drift". Only
one consumer actually derives from it — `INTENT_ISSUE_TEMPLATE` at `src/intent-shape.ts:130`.
`composeTaskBrief` re-types the headings as string literals:

```
src/intent-shape.ts:106-111
section("Problem", draft.problem),
section("Outcome", draft.outcome),
section("Constraints", bullets(draft.constraints)),
section("Non-goals", bullets(draft.nonGoals)),
section("Clarifications", clarifications),
section("Open questions", bullets(unanswered)),
```

(and the helper prompt's per-field descriptions at `src/task-shape.ts:135-136` restate the `hint`
strings at `src/intent-shape.ts:35-51` rather than reading them). Editing a heading in
`INTENT_SECTIONS` — the exact edit the comment invites — changes the prescribed issue template and
silently leaves the brief on the old headings, i.e. precisely the drift the module says it exists to
prevent and the plan set out to avoid by putting composition server-side. Either derive the four
fixed headings from `INTENT_SECTIONS` ("Clarifications" has no counterpart and can stay literal), or
correct the comment to say only the template is derived.

---

## Nits (non-blocking)

- `src/readiness.ts:109-110` — the section comment still reads "These three guardrails are the same markers regardless of stack"; with `detectIssueTemplates` inserted at `:123` there are now four (`detectCi` `:112`, `detectIssueTemplates` `:123`, `detectDependencyAutomation` `:139`, `detectAgentInstructions` `:154`).
- `src/readiness.ts:809-823` — `ISSUE_TEMPLATE_NOTE` was inserted between the `PRESCRIPTION_NOTE` JSDoc block (`:809-816`) and `PRESCRIPTION_NOTE` itself (`:825`), so that doc comment now attaches to the wrong symbol. Moving the new const above the JSDoc restores the pairing.
- `ui/src/lib/components/NewTask.svelte:868-872` — a failed compose sets `shapeStatus = "error"`, which stops `ShapeRound` from rendering the questions (`ui/src/lib/components/ShapeRound.svelte:37`); the operator's answers and the whole round are then unrecoverable and a fresh 180s agent run is the only retry. Keeping the questions visible with the error above them would cost nothing.
- `src/server.ts:6583` returns `{error: "unavailable"}` with 503 when `deps.shapeTask` is unwired, which the card maps to `shape_err_unavailable` — "API-key mode is on but no key is configured" (`ui/messages/en.json:3760`). Wrong explanation for that case, though unreachable in production since `src/index.ts:3154` always wires the dep.
- `ui/src/lib/components/new-task/shape.ts:29-33` — the JSDoc describing `shapeErrorKey` sits above the `ShapeFailure` type declaration (`:36`) instead of above the function (`:38`).
- `src/task-shape.ts:44` — the `SHAPE_LABEL` doc says the label is `shape <n>`, but `src/index.ts:3161` builds it from an 8-hex-char `randomUUID()` slice, not a number (`src/tab-reaper.ts:63` repeats the `<n>` form).

- `src/transient-agent-argv.ts:124` summarises `addDirs` as "extra directories the agent may read", while the same comment at `:128-130` correctly notes the flag also opens them to the preset's bare `Write`. A future caller reading only the summary line (or just the option name) will under-estimate the grant. No caller other than `src/task-shape.ts:282` passes it today.
- `src/task-shape.ts:20` states the worst case of the `--add-dir` grant is "a stray file in the operator's working tree, visible in `git status`". A write under `<repo>/.git/` is inside the added directory and would not surface in `git status`; whether Claude Code's `Write` refuses `.git/` paths is not answerable from this repo (see limitations). Given that Shepherd already spawns full sessions in the same repos with `--dangerously-bypass-approvals-and-sandbox` (`src/service.ts:3117`), the marginal exposure is small — but the comment's guarantee is stronger than what has been established.

- `ui/src/lib/components/NewTask.svelte:1957` also hides the Shape button on the mobile layout, on top of the plan's `code`-mode gate. Reasoned in the inline comment (in-field toolbar budget at 375px) and defensible, but it is a narrowing the plan did not describe.

## Limitations (could not verify)

- Whether `claude --model default` is rejected outright or silently ignored by the installed CLI — the CLI is not in this repo. Finding 1 does not depend on the answer: the codebase's own contract (`NewTask.svelte:1423`, `ModelCliPicker.svelte:91`, `src/service.ts:2883`) is that `"default"` means "emit no `--model`", and the shape path violates it either way.
- Whether Codex's `--sandbox workspace-write` (`src/codex-role-argv.ts:34`) really permits reads outside the spawn cwd, which is the premise for the Codex path needing no `addDirs` equivalent (`src/transient-agent-argv.ts:129-130`). Not determinable from this repo; assumed as written.
- Whether Claude Code's `Write` tool refuses paths under `.git/` inside an `--add-dir` directory (see the latent note above).

scoop:
# Review — Optional structured intake for New Task: a "shape this" clarifying round (#2158)

Re-review, rework round 2. Base: `037c062d…HEAD` (two commits: `8e0bb2da` the feature, `48acdc7b` a
docs sync).

## Status of the three prior findings

**All three are unaddressed.** The only new commit since the previous revision is `48acdc7b`
("docs: sync docs to recent source changes"), which touches `docs/sandbox-security.md` only. Every
line the prior review cited is byte-identical and at the same line number; I checked each one:

1. **Model `default` still forwarded raw.** `ui/src/lib/components/NewTask.svelte:851` still reads
   `await apiShapeTask(repoPath, prompt, agentProvider, model)`. `model` is a `$state` seeded at
   `NewTask.svelte:224` and can hold the literal `"default"` —
   `ui/src/lib/components/new-task/RunSettingsGroups.svelte:180` renders `<option value="default">`,
   and `ui/src/lib/components/new-task/run-config.ts:81`/`:96` both fall back to `"default"`.
   Every other consumer normalises first: `NewTask.svelte:1423` (`model !== "default" ? model : null`)
   and `ui/src/lib/components/new-task/ModelCliPicker.svelte:91` (`model === "default" ? null : model`).
   The shape path does not, and nothing downstream compensates: `src/server.ts:6567-6572` accepts any
   non-empty string, `src/task-shape.ts:277-282` passes it straight into `buildTransientAgentArgv`,
   and `src/transient-agent-argv.ts:275` emits `--model default` (Codex: `src/codex-role-argv.ts:35`
   emits `-m default`). The new browser test asserts only `mock.calls[0][0]` and `[1]`
   (`ui/src/lib/components/NewTask.browser.test.ts:4756-4757`), so it does not cover this. Re-raised.

2. **Adopt-prompt behaviour still contradicts its own copy.** `ui/src/lib/components/ReadinessPanel.svelte:160-163`
   still drops the template when `hasIssueTemplates` is true (`template = report.hasIssueTemplates ? undefined : …`,
   and `ui/src/lib/components/readiness-view.ts:43-45` omits the section when the argument is absent),
   while the note rendered in exactly that branch — `ReadinessPanel.svelte:319-321`, key
   `readiness_template_present_note` — still reads "the task will add this one alongside them"
   (`ui/messages/en.json:3766`, `ui/messages/de.json:3766`). The matching doc comment at
   `src/readiness.ts:63` still says "so the prescription says merge, not create"; verified against
   `src/readiness.ts:722` (`generateClaudeMd` builds `adoptLines` from `checks.filter((c) => !c.present)`),
   so when the guardrail is present there is no prescription at all — neither merge nor create.
   Re-raised.

3. **`INTENT_SECTIONS` still not the single source for the composer.** `src/intent-shape.ts:31` still
   reads "Edit HERE — every artifact below is derived from it", and the module header at
   `src/intent-shape.ts:5` still calls it "One section list, two consumers that must not drift". But
   `composeTaskBrief` hard-codes its headings at `src/intent-shape.ts:106-111` (and adds a sixth,
   "Clarifications", that is not in the list at all); `src/task-shape.ts:135-136` restates the hints
   in prose. Only `INTENT_ISSUE_TEMPLATE` at `src/intent-shape.ts:130` derives from the list.
   Re-raised.

## New findings

### A. `useBrief` autogrows before the DOM has the new prompt

`ui/src/lib/components/NewTask.svelte:873-877` assigns `prompt = brief` and then calls `autogrow()`
synchronously. `autogrow` (`NewTask.svelte:1045-1049`) measures `promptInput.scrollHeight`, and the
textarea is `bind:value={prompt}` (`NewTask.svelte:1878-1879`) — Svelte 5 flushes that binding in a
microtask, so the measurement is taken against the *old* value and the field keeps its previous
height while holding a multi-section brief.

Every other programmatic prompt replacement in this same file defers, for exactly this reason:
`pickIssue` (`NewTask.svelte:1005-1006`, `queueMicrotask(autogrow)`), `injectSteer` (`:1015-1017`),
`pickCommand` (`:1120-1124`), `pickCommandFromSource` (`:1147-1151`), `pickIssueFromSearch`
(`:1162-1166`), and the `PromptSources onpick` handler (`:2151-2158`), plus `closeSheetFocusPrompt`
which uses `await tick()` (`:1027-1032`). `useBrief` is the only site that does not — and it is the
one that inserts the largest text.

### B. The accepted-risk statement for `--add-dir` is wrong about `git status`

`src/task-shape.ts:19-22` states the residual risk as: "the worst case is a stray file in the
operator's working tree, visible in `git status` and committed by nobody."

That understates it on two verified counts. The preset is
`"writer-ro": { allowedTools: ["Read", "Grep", "Glob", "Write"] }` (`src/transient-agent-argv.ts:183`)
under `--permission-mode dontAsk` (`src/transient-agent-argv.ts:277`), and `--add-dir <repoPath>`
(`src/transient-agent-argv.ts:265-268`) makes the operator's **real checkout** — not a disposable
worktree, unlike the `reviewer` kind whose posture this borrows (`src/transient-agent-argv.ts:163-166`)
— a write target:

- bare `Write` overwrites existing files, so the outcome is not only "a stray file" but silent
  modification of tracked sources; and
- `git status` does not show gitignored paths, so a write to e.g. a local settings or env file
  inside the repo is invisible to the stated detection mechanism.

Shepherd's own hook guard cannot catch it either: the spawn carries `disableAllHooks: true`
(`src/transient-agent-argv.ts:255`), which disables Shepherd's tool-guard hook along with the repo's.
The feature is operator-initiated on the operator's own repo and the plan accepts the exposure, so
I am not asking for the flag to be removed — but the sentence that carries the security argument has
to describe the actual residual risk (`docs/sandbox-security.md:172-176`, added by this PR, leans on
the same reasoning and says only "writable-in-the-worst-case", which is accurate as far as it goes).

### C. `src/readiness.ts:109` — "These three guardrails" is now four

The section comment at `src/readiness.ts:108-110` says "These three guardrails are the same markers
regardless of stack". The section now holds four detectors: `detectCi` (`:112`), the new
`detectIssueTemplates` (`:123`), `detectDependencyAutomation` (`:139`) and `detectAgentInstructions`
(`:153`). The diff made the count wrong.

## What I verified and found correct

- **Locale parity.** Both catalogs gained the same 24 keys in the same order —
  `newtask_shape_label`, `newtask_shape_aria`, `newtask_shape_hint`, `shape_heading`,
  `shape_running`, `shape_questions_count`, `shape_s_problem`, `shape_s_outcome`,
  `shape_s_constraints`, `shape_s_nongoals`, `shape_use_brief`, `shape_discard`,
  `shape_err_timeout`, `shape_err_spawn_failed`, `shape_err_unavailable`, `shape_err_empty_prompt`,
  `shape_err_compose`, `readiness_g_issue_templates_title`, `readiness_g_issue_templates_removes`,
  `readiness_template_heading`, `readiness_template_present_note`, `readiness_adopt_template_intro`,
  `feat_shape_task_title`, `feat_shape_task_body` (`ui/messages/en.json:3746-3769` vs
  `ui/messages/de.json:3746-3769`). Every key referenced by `ShapeRound.svelte:24-28,34-48,57` and
  `ReadinessPanel.svelte:81-82,112-113,311,320` resolves to an entry in both.
  `shape_questions_count` carries the `{count}` placeholder in both, matching the call at
  `ShapeRound.svelte:39`.
- **Containment of `repoPath`.** `src/server.ts:6581` calls `safeRepoDir(…, config.repoRoot)`;
  `src/validate.ts:906-924` realpaths both sides, rejects anything outside the root, and requires a
  directory. So the string that becomes `--add-dir` is not caller-controlled beyond the root.
- **Route reachability.** `handleShape`/`handleShapeBrief` are registered in `ROUTE_HANDLERS`
  (`src/server.ts:7909-7910`) inside `makeApp`, which runs `checkAuth` before the handlers
  (`src/server.ts:7928-7929`). Neither path matches `isAgentIngressRoute` (`src/server.ts:8017-8018`,
  which requires `parts[1] === "sessions"`), so a spawned agent cannot reach them.
- **`--add-dir` placement.** Emitted at `src/transient-agent-argv.ts:265-268`, immediately after
  `--session-id <uuid>` and before `--settings`, so the variadic flag is always followed by a flag
  and can never swallow the trailing prompt positional pushed at `:286`. Guarded by
  `opts.addDirs?.length`, so every existing caller's argv is unchanged.
- **Answer round-trip keys.** `QuestionFormBlock.buildAnswers()` stamps `blockId: block.id`
  (`ui/src/lib/components/blocks/QuestionFormBlock.svelte:70,74,79`); the block it renders comes from
  `normalizeRound`, which always stamps `SHAPE_BLOCK_ID` (`src/task-shape.ts:51,220,228`), and the
  brief route re-normalises to the same id before `resolvePlanAnswers([round.block], answers)`
  (`src/server.ts:6603-6607`). `resolvePlanAnswers` is fail-closed against non-object/typeless
  entries (`src/plan-gate.ts:1811-1822`), so the route's unvalidated `body.answers as RawAnswer[]`
  cast is safe.
- **Question validation reuse.** `parseVisualBlocks` drops option-less `single`, duplicate ids, empty
  ids and unknown kinds (`src/visual-blocks.ts:514-544`), and returns `null` for a form whose
  questions all drop — which `normalizeRound` handles via the `block ?? {…questions: []}` fallback
  (`src/task-shape.ts:228`).
- **Plan-gate regression surface.** The `answerCtx` branch of `QuestionFormBlock.submit()` is intact
  (`QuestionFormBlock.svelte:92-104`); the only change on that path is the `submitted && answerCtx`
  guard at `:163`, which is unchanged behaviour whenever `answerCtx` is set.
- **Tab-reaper coverage.** `SHAPE_LABEL` is `"shape "` (`src/task-shape.ts:47`), registered in
  `isShepherdHelperLabel` (`src/tab-reaper.ts:86`) and boot-reaped at `src/index.ts:2505` via
  `reapTransientByLabel`, which prefix-matches (`src/transient-tab-reaper.ts:62`).
- **Feature-announcement contract.** `ui/src/lib/feature-announcements/entries/v1.48.0-shape-task.ts`
  matches the shape required by `ui/src/lib/feature-announcements.ts:25-36`, its `sinceVersion`
  (1.48.0) is ahead of `package.json:3` (1.47.0) as `feature-announcements.ts:11-16` requires, its
  filename matches `v<sinceVersion>-<id>.ts`, and its `targetId: "nt-shape"` is anchored by
  `use:coachTarget={"nt-shape"}` at `NewTask.svelte:1961`.
- **Readiness wiring.** `issue_templates` has entries in all six exhaustive `Record<GuardrailId, …>`
  tables (`src/readiness.ts:775,797,864,881,899,946`), is added to both `GUARDRAILS` (`:295`) and
  `RUST_GUARDRAILS` (`:562`), and the not-applicable path returns `hasIssueTemplates: false` /
  `issueTemplate: ""` (`src/readiness.ts:689-691`). The UI mirror in `ui/src/lib/types.ts:1803`
  adds the same union member and both new report fields.

- Whether Codex's `--sandbox workspace-write` really permits reads outside the spawn cwd — the basis
  for the "Codex ignores it" claim in `src/transient-agent-argv.ts:129-130` and
  `docs/sandbox-security.md:177-178`. Codex's sandbox semantics are not in this repo; assumed as
  written. If it does not hold, the Codex shaping path is silently ungrounded (it would still return
  a round, just one that never read the repo).
- Whether `--add-dir <repo>` makes Claude Code discover that repo's project `.mcp.json` and render
  the "N new MCP servers found" gate. `writer-ro` is deliberately `mcpIsolated: false`
  (`src/transient-agent-argv.ts:183`) on the stated grounds that it "runs in a scratch dir with no
  project .mcp.json" (`:170-172`), and this PR is the first caller to add a real repo to such a
  spawn. That gate is documented at `src/transient-agent-argv.ts:86-89` as an invisible hang on an
  unattended pane; here it would surface as a 180 s `timeout`. I could not verify the CLI's
  behaviour from this repo, so this is a flag for a manual check, not a finding.

## Nits (non-blocking):

- `src/readiness.ts:808-818` — the pre-existing block comment documenting `PRESCRIPTION_NOTE` ("Extra prose guidance lines per guardrail…") now sits directly above the newly inserted `ISSUE_TEMPLATE_NOTE`, so it reads as that constant's JSDoc while `PRESCRIPTION_NOTE` (`:825`) has none; moving the new constant above the block comment restores it.
- `ui/src/lib/components/new-task/shape.ts:29-36` — same shape: the doc block describing `shapeErrorKey` is separated from it by the `ShapeFailure` declaration and its own comment, so it now documents the type instead of the function.
- `src/task-shape.ts:44-46` — "Space-prefixed so a prompt-derived `[a-z0-9-]` session slug can never collide" describes `"shape "`, which is space-*suffixed* (the wording is copied from `src/prompt-recommend.ts:24-25`, where it is equally inverted).
- `src/tab-reaper.ts:63` — documents the label as `shape <n>`, but `src/index.ts:3161` builds `shape <8 hex chars of randomUUID>`.
- `ui/src/lib/components/NewTask.svelte:867-871` — a failed compose sets `shapeStatus = "error"`, which makes `ShapeRound` swap the question form out for the error line (`ShapeRound.svelte:45-49`), discarding the answers the operator just gave; `shape_err_compose` says "Try again", but the only route back is a fresh 180 s agent round.

- `src/task-shape.ts:70` declares an `"empty-prompt"` error and `:268` returns it, but `ui/src/lib/components/new-task/shape.ts:19-24` disables the control on a blank prompt, so that branch and its `shape_err_empty_prompt` copy in both catalogs are unreachable from the UI.
- `src/task-shape.ts:183-194` (`clipQuestion`) adds a second sanitisation layer in front of `parseVisualBlocks`, which already validates the same objects; defensible as a size cap, but it is machinery the task did not ask for.

scoop:
**Findings 1–3 were already fixed in `00b099ee` (pushed ~10 min before this review), so there is nothing left to change for them.** They were the previous round's findings and I addressed them then; this review appears to have re-read the pre-fix commit — its line citations match the old content (e.g. `NewTask.svelte:851` is now the comment introducing the fix).

Current state on `origin/shepherd/clarify-new-task`:

1. **Model normalisation** — `runShape` calls `apiShapeTask(repoPath, prompt, agentProvider, model !== "default" ? model : null)`. `shapeTask`, `AppDeps.shapeTask` and the route all take `model: string | null`; an empty string stays malformed. Pinned by `test/task-shape.test.ts` ("a null model emits no `--model` flag"), `test/server-shape.test.ts` ("maps an absent model to null") and an assertion in the New Task browser test.
2. **Template note** — now reads "This repo already ships issue templates, so the install task leaves them alone — this one is here to copy from if you want the intent shape too." in both catalogs, matching the behaviour. `hasIssueTemplates`' doc comment was rewritten too: present ⇒ guardrail satisfied ⇒ artifact is reference-only; absent ⇒ the prescription says create. `test/readiness.test.ts` pins that the artifact ships either way and only the prescription differs.
3. **One section list** — `composeTaskBrief` renders `HEADINGS.problem` … `HEADINGS.openQuestions`, derived from `INTENT_SECTIONS`; the shaper prompt builds its field guidance from `DRAFT_SECTIONS` (the same list minus `openQuestions`) instead of restating the hints. A test walks `INTENT_SECTIONS` and asserts every heading reaches the brief.

Findings **4, 5 and 6 were new and are fixed in `3c60a84b`**:

- **4** was a real bug — reverting the `queueMicrotask` deferral makes the new test fail with 25px of unshown overflow, so it is regression-covered rather than just corrected.
- **5** — you were right that the note was too rosy. It now states that the agent can overwrite a tracked file or write a gitignored path `git status` never shows, that `disableAllHooks` rules out the tool-guard hook, what actually bounds the radius (no `Bash`/`Edit`/git/network ⇒ no exec, commit, push or exfiltration), and that the real remedy if that ceiling is too high is a read-only view of the repo, not a tighter allowlist. Corrected in `src/task-shape.ts`, `docs/sandbox-security.md` and the PR description, since all three carried the understatement.
- **6** — comment now describes the section by what it holds rather than by a count that goes stale on the next detector.

scoop (commented):
# Review — Optional structured intake for New Task ("shape this" round), rework round 3

Verdict: **comment**. Every prior-round finding is genuinely fixed in the tree at HEAD; I found no
new blocking defect. Remaining observations are non-blocking and listed below.

## Prior findings — re-verified against the diff (not against the author's note)

**1. Model normalisation (`ui/src/lib/components/NewTask.svelte`) — FIXED.**
`runShape` now calls `apiShapeTask(repoPath, prompt, agentProvider, model !== "default" ? model : null)`
(`ui/src/lib/components/NewTask.svelte:853-858`), the same normalisation the spawn path uses at
`ui/src/lib/components/NewTask.svelte:1435`. The type flows all the way down: `shapeTask()` in
`ui/src/lib/api.ts:1053` takes `model: string | null`; `AppDeps.shapeTask` declares
`model: string | null` (`src/server.ts:593`); `parseShapeBody` maps an absent/`null` model to `null`
and rejects an empty string (`src/server.ts:6567-6570`); `ShapeArgs.model` is `string | null`
(`src/task-shape.ts:112`). A `null` model then emits no flag on either path — Claude:
`if (opts.model) argv.push("--model", opts.model)` (`src/transient-agent-argv.ts:275`); Codex:
`if (model) argv.push("-m", model)` (`src/codex-role-argv.ts:35`). The literal `default` can no
longer reach a CLI.

**2. Template note vs. adopt behaviour — FIXED.**
`sendToTask` now passes the template only when the repo LACKS one:
`report.hasIssueTemplates ? undefined : {…}` (`ui/src/lib/components/ReadinessPanel.svelte:158-162`),
and `buildAdoptPrompt` appends the section only when a non-blank body is supplied
(`ui/src/lib/components/readiness-view.ts:38-45`). The copy shown in exactly the
`hasIssueTemplates === true` branch (`ui/src/lib/components/ReadinessPanel.svelte:319-321`) now reads
"the install task leaves them alone — this one is here to copy from" (`ui/messages/en.json:3766`) with
the matching German at `ui/messages/de.json:3766` ("die Installationsaufgabe rührt sie daher nicht an
— diese hier ist zum Kopieren da"). Behaviour and copy agree in both catalogs. The server-side doc
comment was rewritten too (`src/readiness.ts:63-66`: "Present ⇒ the guardrail is satisfied, so the
adopt task leaves the repo's templates alone … absent ⇒ the prescription tells the agent to create
it"), which I confirmed against `generateClaudeMd`, where only `checks.filter((c) => !c.present)` is
prescribed (`src/readiness.ts:725`).

**3. "One section list" (`src/intent-shape.ts`) — FIXED.**
`HEADINGS` is derived from `INTENT_SECTIONS` via `Object.fromEntries`
(`src/intent-shape.ts:75-78`) and `composeTaskBrief` renders `HEADINGS.problem` … `HEADINGS.openQuestions`
(`src/intent-shape.ts:118-125`) — no hard-coded headings left, except the deliberately non-intent
`"Clarifications"` section, which the adjacent comment calls out as such
(`src/intent-shape.ts:122-123`). `src/task-shape.ts` no longer restates the hints: the shaper prompt
builds its field guidance from `DRAFT_SECTIONS` (`src/task-shape.ts:158`), which is
`INTENT_SECTIONS.filter((s) => s.key !== "openQuestions")` (`src/intent-shape.ts:84`).
`INTENT_ISSUE_TEMPLATE` still derives from the same list (`src/intent-shape.ts:150`).

**4. `useBrief` autogrow deferral — FIXED.**
`autogrow()` and `promptInput?.focus()` now run inside `queueMicrotask` after `prompt = brief`
(`ui/src/lib/components/NewTask.svelte:882-888`), matching the other prompt-replacing sites in the
same file (`ui/src/lib/components/NewTask.svelte:1018`, `:1028`, `:1132`, `:1159`, `:1174`). The
`bind:value={prompt}` textarea flush is scheduled before the queued microtask, so `scrollHeight` is
measured against the new value.

**5. Accepted-risk note in `src/task-shape.ts` — FIXED.**
The header now states the actual residual risk (`src/task-shape.ts:19-42`): overwrite of a tracked
file, writes to gitignored paths `git status` never shows, `disableAllHooks: true` ruling out the
tool-guard hook, what actually bounds the radius (no `Bash`/`Edit`/git/network), and that the real
remedy is a read-only view rather than a tighter allowlist. I verified the two claims it rests on:
the `writer-ro` preset is `["Read","Grep","Glob","Write"]` (`src/transient-agent-argv.ts:183`) and
`disableAllHooks: true` is unconditional on the Claude transient path
(`src/transient-agent-argv.ts:255`). `docs/sandbox-security.md:166-190` carries the same corrected
statement, and `src/task-shape.ts` was added to its "See also" list (`docs/sandbox-security.md:214`).

**6. `src/readiness.ts` "These three guardrails" comment — FIXED.**
Now reads "These guardrails are the same markers regardless of stack — CI, dependency automation,
agent instructions and issue templates are repo/forge configuration, not toolchain"
(`src/readiness.ts:110-113`) — no count to go stale.

## Verification performed on the rest of the diff

- **Identifiers resolve.** `parseVisualBlocks` (`src/visual-blocks.ts:585`), `resolvePlanAnswers`
  (`src/plan-gate.ts:1810`), `RawAnswer`/`ResolvedAnswer` (`src/plan-gate.ts:1783`, `:1794`),
  `safeRepoDir` (`src/validate.ts:906`), `apiKeyFailClosed` (`src/spawn-auth.ts:57`),
  `makeHelperTmpDir`/`cleanupHelperDir`/`reapHelperRun`/`realSleep`
  (`src/transient-helper-lifecycle.ts:18`, `:23`, `:37`, `:15`), `writer-ro` preset
  (`src/transient-agent-argv.ts:183`), `FileScan.glob` signature (`src/readiness.ts:79`) — all exist
  and are used with the right shapes. `RawAnswer` is imported in `QuestionFormBlock.svelte:2`.
- **Answer round-trip keys agree.** `normalizeRound` forces the block id to `SHAPE_BLOCK_ID`
  (`src/task-shape.ts:243`), the client's `buildAnswers` keys on `block.id`, and
  `indexQuestions` namespaces by `` `${b.id} ${q.id}` `` (`src/plan-gate.ts:1834`) — so
  `/api/shape/brief`'s `resolvePlanAnswers([round.block], answers)` (`src/server.ts:6614`) matches.
- **argv placement is safe.** With `addDirs`, the Claude argv is
  `claude --session-id <uuid> --add-dir <repo> --settings … --allowedTools Read Grep Glob Write
  [--model …] [--effort …] --permission-mode dontAsk <prompt>`
  (`src/transient-agent-argv.ts:258-283`). Both variadic flags (`--add-dir`, `--allowedTools`) are
  followed by a flag, never by the positional prompt — the same rule
  `src/service.ts:3058-3067` follows for the session spawn.
- **Route reachability/auth.** `/api/shape` and `/api/shape/brief` are registered at
  `src/server.ts:7918-7919`, and neither is reachable through the loopback agent-ingress listener,
  whose allowlist requires `parts[0] === "api" && parts[1] === "sessions"`
  (`src/server.ts:8026`). `repoPath` is realpath-resolved and containment-checked against
  `config.repoRoot` before it becomes `--add-dir` (`src/server.ts:6590`, `src/validate.ts:906-923`).
  `/api/shape` is given a 255s slow-request budget (`src/server.ts:8209`) against the helper's 180s
  poll (`src/task-shape.ts:305`).
- **Type exhaustiveness.** `issue_templates` is added to `GuardrailId` (`src/readiness.ts:36`) and
  reaches every `Record<GuardrailId, …>` / `Record<RustGuardrailId, …>` table
  (`src/readiness.ts:778`, `:800`, `:867`, `:884`, `:902`, `:949`); `RustGuardrailId` is
  `Exclude<GuardrailId, "type_checker" | "lint_staged">` (`src/readiness.ts:408`), so it includes the
  new member. `ReadinessReport`'s two new required fields are set on both the applicable path
  (`src/readiness.ts:671-673`) and the not-applicable one (`src/readiness.ts:692-694`); the only
  other consumers of that shape are `ReadinessPanel.svelte` and the tests.
- **Locale parity.** All 24 new keys appear in both catalogs with identical names and ordering
  (`ui/messages/en.json:3746-3769` vs `ui/messages/de.json:3746-3769`), including the parameterised
  `shape_questions_count` (`{count}` in both). Every message accessor used by the new components
  resolves to one of those keys, and the reused `readiness_copy`/`readiness_copied` and the
  `merge-note`/`claudemd-head`/`cta-row` classes already exist in
  `ui/src/lib/components/ReadinessPanel.svelte:573`, `:534`, `:540`.
- **Feature-catalog gate.** `ui/src/lib/feature-announcements/entries/v1.48.0-shape-task.ts` uses
  `targetId: "nt-shape"`, matched by `use:coachTarget={"nt-shape"}`
  (`ui/src/lib/components/NewTask.svelte:1973`), and `sinceVersion: "1.48.0"` is above the last
  release (`package.json:3` = `1.47.0`) and consistent with the sibling `v1.48.0-*` entries.
- **Plan-gate regression path.** `QuestionFormBlock.submit()` still returns early when neither
  `answerCtx` nor `onanswer` is present (`canSubmit` is false because `interactive` is false,
  `QuestionFormBlock.svelte:22`, `:51`), and the `answerCtx` POST branch is untouched
  (`QuestionFormBlock.svelte:92-98`). The one template change, `{#if submitted && answerCtx}`
  (`QuestionFormBlock.svelte:163`), is a no-op for the gate path, where `submitted` is only ever set
  inside the `answerCtx` branch.

- The claim that a Codex helper under `--sandbox workspace-write` can still READ the repo without an
  equivalent of `--add-dir` (`src/transient-agent-argv.ts:129-130`, `src/task-shape.ts:16-18`) rests
  on the Codex CLI's sandbox semantics, which are not in this repo. Assumed as written; if it is
  wrong, the Codex shaping path would produce ungrounded questions rather than fail loudly.

## Nits (non-blocking)

- `ui/src/lib/types.ts:1823` — the mirrored doc comment still says `hasIssueTemplates` means "prescription says merge, not create", but a present guardrail is filtered out of the prescription entirely (`src/readiness.ts:725`) and the template is dropped from the adopt prompt (`ui/src/lib/components/ReadinessPanel.svelte:158-162`); the server-side twin was corrected in this round (`src/readiness.ts:63-66`) and this copy was left behind.
- `ui/src/lib/components/ShapeRound.svelte:47-49` — a draft-only round (`normalizeRound` can legitimately return a block with zero questions, `src/task-shape.ts:243`) renders the meta line as "0 to settle"; suppressing the count when the list is empty would read better.
- `ui/src/lib/components/NewTask.svelte:1974-1976` — the Shape button's `title`/`aria-label` are constant, so a disabled button (blank prompt, no repo) gives no reason; `shapeBlocker` already returns a discriminated reason that could feed the tooltip.
- `ui/src/lib/components/ShapeRound.svelte:60-70` — the two `{#each}` arrays over ad-hoc `[label, value]` tuples widen to `(string | string[])[][]`; a small typed array of `{label, kind, value}` would keep the narrowing explicit.

PR #2172, opened 2026-09-02
feat(maintain): tiered thresholds that open work (#2157)

scoop:
Closes #2157. Follow-up to #2151 R5 (`docs/research/ai-native-sdlc-playbook.md` §3 R5); builds on R1 (#2153) and plan drift (#2155), which supply the signal substrate.

Shepherd's Stage-6 machinery notifies and stops. This closes the loop: a daily sweep scores three declared bands over Shepherd's **own** health data, logs what it measures, and — for a sustained breach — spawns a read-only diagnosis agent whose drafted backlog issue the **server** files.

## The bands

| Band | Scope | Window | Min sample | Tier 1 (log) | Tier 2 (diagnose) |
| --- | --- | --- | --- | --- | --- |
| `critic_error_rate` | global | 7d | 10 review spawns | ≥ 15% | ≥ 30% |
| `incident_spike` | global, per signal kind | 7d | — | ≥ 10 occurrences **and** ≥ 3 sessions | ≥ 25 **and** ≥ 5 |
| `first_pass_collapse` | per repo | 30d | 8 merged tasks | < 60% | < 40% |

The table in `maintain-core.ts` **is** the version-controlled config the playbook asks for; `SHEPHERD_MAINTAIN_THRESHOLDS` deep-merges a JSON override over it, field by field and fail-soft, so a recalibration ships without a deploy.

`critic_error_rate` reads `reviewer_spawns` directly rather than R1's `DeliveryStats.criticErrors` — that field's denominator is merged-tasks-in-window, which biases the rate downward exactly when review errors are what's blocking merges. `incident_spike` excludes the `reply` kind: that's the learnings correction stream, high-volume by design.

## Trust boundary

The diagnosis agent runs under the existing `reviewer` preset (`--permission-mode dontAsk`, scoped `--allowedTools`, never `--dangerously-skip-permissions`) in a disposable detached worktree of Shepherd's own checkout. It has no `gh`, no network, no exec. Its only deliverable is `.shepherd-maintain.json`; the trusted server parses, clamps and files. Same split as the doc agent. Signal payloads reaching the prompt are fenced with `fenceUntrusted` and the block is capped before assembly.

Scope guard held: **internal signals only, no production ingress**, and no statistical control bands — §4 rules both out.

## Rollout

[...]

PR #2176, opened 2026-09-02
fix(review): never review or persist a stale head (#2175)

### LINKED ISSUE #2175: Critic reviews a stale head: consider() trusts the poller's cached headSha and never re-validates before spawning
`ReviewService.consider()` takes its `headSha` from the PR poller's cached `GitState` and never re-validates it at spawn time. When a push lands between the poller's observation and the spawn, the critic checks out **superseded code**, and its `changes_requested` verdict is persisted and steered to the author as if current.

Observed on PR #2172 (session TASK-2253) — the critic re-reported four findings the author had already fixed and pushed.

## Timeline (all local)

| Time | Event |
| --- | --- |
| 14:37:37 | `2d1f4716` pushed (the feature) |
| 14:46:15 | review #1 spawns against `2d1f4716` |
| 14:55:00 | `be23997b` pushed (docs-manifest regen) |
| 14:57:27 | review #1 finalizes — 4 valid findings, correctly against `2d1f4716` |
| **15:03:10** | **`50263565` pushed — all 4 findings fixed** |
| **15:04:19** | **review #2 spawns — 69s AFTER that push — but checks out `be23997b`** |
| 15:12:27 | review #2 finalizes, persists `headSha = be23997b`, re-reports the same 4 findings |

`sqlite3 ~/.shepherd/shepherd.db "SELECT substr(headSha,1,8), decision, addressRound, addressCap FROM reviews WHERE sessionId='7b7d1925-5ae7-4cfb-947e-7a7cefaf6c8e'"` → `be23997b|changes_requested|2|4`.

This is **not** a mid-review race — review #2 started a full minute after the newer head existed. It is a stale-snapshot read.

## Mechanism

- `consider(session, git: GitState, …)` (`src/review.ts` → `consider()`) gates on `git.headSha` + `checksCleared(git.checks)` from a **cached** poller snapshot, and passes that sha to the worktree checkout.
- Nothing re-reads the tip between that decision and `createDetached`.
- `finalize` persists `headSha: f.headSha` verbatim (`src/review.ts` → `persistVerdict`, the `headSha: f.headSha` assignment) with no comparison against the current tip.
- `headAlreadySettled()` only suppresses re-reviewing the *same* head; it has no notion of a verdict computed against a *superseded* one.

The poller is also deliberately coarse while cold (`idleIntervalMs` default 300s), so the stale window is up to a full sweep — not a millisecond race.

## Why it matters

It self-heals eventually — the next poll sees the new head, `headAlreadySettled` is false, and a third review fires. But each occurrence costs:

1. **A wasted critic spawn** (~8 minutes and its full token burn here).
2. **A spurious `changes_requested` round delivered to the author**, who has nothing to fix. In autopilot that is a wasted steer; attended, it is a confusing round-trip.
3. **A consumed rework round.** `addressRound` went 1 → 2 against `addressCap` 4 — a quarter of the budget spent reviewing code that no longer existed. A session that hits the cap this way stalls on a false premise.
4. It can loop: if the author responds with a note rather than a push (correct — there is nothing to fix), and the poller is cold, the same stale snapshot can drive another round.

## Suggested fix

Two independent guards; the first prevents the waste, the second is the backstop for a genuine mid-review push (which no pre-spawn check can eliminate).

1. **Re-validate the head immediately before spawning.** In `begin()`, after the eligibility gates and before `createDetached`, re-read the PR head (`forge.prStatus(headBranch)`, one round-trip — trivial next to a critic run) and skip when it no longer matches the snapshot's `headSha`. The next poll picks up the new head normally. Note a new push also resets CI to pending, so the green-checks precondition has genuinely lapsed too.
2. **Discard a verdict whose reviewed head is no longer the tip.** At finalize, compare `f.headSha` against the current head; if it moved, drop the verdict rather than persisting and steering it — and **do not increment `addressRound`**. A round spent on superseded code should not count against the cap.

Guard 2 alone would fix the author-facing symptom and the round accounting; guard 1 additionally saves the spawn.

## Out of scope

The existing `patchId` / `rebaseSkip` machinery is unrelated — it suppresses re-review of an *identical diff* after a rebase. Here the diff genuinely changed and the review simply read the wrong one.

scoop:
Closes #2175.

`ReviewService.consider()` took its `headSha` from the PR poller's cached `GitState` and nothing re-read the tip before the spawn, so a push landing in that window made the critic check out **superseded code** — and its `changes_requested` verdict was persisted and steered to the author as if current. Observed on PR #2172: review #2 spawned 69s *after* the fixing push, reviewed the head before it, and re-reported four already-fixed findings — a wasted ~8-minute spawn, a confusing steer, and a rework round burned against the cap on a false premise.

## Two guards, one predicate

`headSuperseded(reviewedSha, live)` (`src/critic-core.ts`) is true only for a **confirmed, still-`open`** PR carrying a **different** `headSha`. Everything else — no forge, a forge that threw, a payload without `headSha`, a non-`open` state — is false, i.e. **fails open** and behaves exactly as before. A forge blip must never permanently suppress review, and merged/closed PRs already have their own moot handling this must not pre-empt.

**1. Pre-spawn re-validation** (`consider()` → `headMoved()`). One `prStatus` round-trip, then decline if the head moved; the next poll picks the new head up normally. It sits **inside the `try`, after `this.starting.add()`** — the slot claim must stay synchronous, or a suspended `consider()` would let a second call (it is fire-and-forget from `onSessionGit` / `reKickReapedReview`, and `forceReview()` gates re-entry solely on `starting.has`) reach `begin()`: two worktrees, two spawns, and the second `inflight.set` orphaning the first run's terminal and worktree. Covered by a test.

**2. At-finalize supersession discard** (`finalize()`). A verdict whose reviewed head is no longer the tip is dropped whole: nothing posted, nothing steered, no signal, no streak/round/error accounting, and no `reviewer_spawns.outcome` stamp — so a superseded run stays out of the delivery metrics rather than counting as rework the author never received. The prior verdict row stands untouched. `captureUsage` and the terminal/worktree reap run exactly as before, so the spawn's cost is still attributed. This is the backstop no pre-spawn check can replace: a genuine mid-review push.

Error verdicts are discarded on the same terms — a superseded run earns no `errorRound` progress.

## Incidental

`finalize()`'s live PR read (`livePrState` → `livePr`, now returning the whole `PrStatus`) is fetched **once** and threaded into both the error and publish arms, which each used to fetch their own. Those arms are mutually exclusive, so this is still one round-trip per finalize — but state and head now always come from the same snapshot.

Net forge traffic: +1 `gh` call per would-be spawn, on a path that has already cleared every other gate.

## Test harness

`fakeForge`'s default `prStatus` returned a fixed `headSha: "abc"` while 15 tests considered `"newsha"` — with guard 1 in place that reads as "head moved" and would have silently skipped them all. The default now echoes a live-head box that an `atHead(sha)` helper moves in step with the poller snapshot, so the two agree by construction and only a test that *means* to diverge them does. Four partial forge fakes (`{ getIssue }` only) gained a `prStatus` so they no longer fail open through a `TypeError`.

## Unrelated CI fix carried here

`verify` went red on `diagnostics.test.ts` → "the probe reads the LIVE cell health" timing out at 5s. That test built `DiagnosticsService` without `healthyDeps()`, so every dep it didn't override fell through to its REAL probe (gh auth, herdr liveness, tailscale serve, bwrap membrane); on a runner without those binaries the retries blow the per-test budget. It now spreads `healthyDeps()` like its siblings — assertions unchanged. Nothing to do with the guards; it just had to be green for the critic to review this.

## Verification

`bun run lint`, `bun run test` (8906 tests), `bunx tsc --noEmit`, `ui/ bun run check`, and `bunx fallow audit --base origin/main --fail-on-issues` all pass. New tests cover: declined spawn on a moved head (no critic, no probe worktree, no state) · the re-check under the starting claim rejecting a concurrent `consider()` · `forceReview` declining a moved head while its hygiene reset still stands · fail-open on a throwing forge and on a payload without `headSha` · discard of a findings verdict and of an error verdict at finalize, with the prior row, the round budget and the metrics untouched but the run still reaped and costed · plus the unchanged-head control paths.

### DIFF
--- a/src/critic-core.ts
+++ b/src/critic-core.ts
@@ -12,6 +12,7 @@ import { execFile } from "node:child_process";
 import { promisify } from "node:util";
 import { execFileSync, timedAsync } from "./instrument";
 import type { PlanDrift, ReviewDecision } from "./types";
+import type { PrStatus } from "./forge/types";
 import type { SessionUsage } from "./usage";
 import { tolerantParseJson } from "./json-tolerant";
 import type { VerdictRead } from "./json-tolerant";
@@ -1411,6 +1412,24 @@ export async function captureUsage(
   } catch (err) {
     console.warn(`[review] usage capture failed for ${logLabel}:`, err);
   }
+}
+
+/**
+ * Has the head we are about to review — or have just reviewed — been superseded by a newer push?
+ *
+ * The poller's `GitState` is a CACHED snapshot (a full sweep can be 300s cold), so the head it
+ * carries can already be stale by the time a critic spawns, and can go stale again during the
+ * ~8 minutes one runs. Reviewing a superseded head costs a spawn, delivers a `changes_requested`
+ * round the author has nothing to fix, and burns a rework round against the cap (issue #2175).
+ *
+ * FAILS OPEN on purpose: only a CONFIRMED, still-`open` PR carrying a DIFFERENT `headSha` counts as
+ * superseded. No forge, a forge that threw (`live` undefined), a payload without `headSha`, or a
+ * non-`open` state all return false — behave exactly as before. A forge blip must never be able to
+ * permanently suppress review, and merged/closed PRs already have their own moot handling that this
+ * predicate must not pre-empt.
+ */
+export function headSuperseded(reviewedSha: string, live: PrStatus | undefined): boolean {
+  return live?.state === "open" && !!live.headSha && live.headSha !== reviewedSha;
 }
 
 /** Terminal + disposable-worktree teardown for a finished critic run.
--- a/src/delivery-metrics.ts
+++ b/src/delivery-metrics.ts
@@ -87,8 +87,9 @@ function rate(hits: number, total: number): DeliverySample {
 }
 
 /** Group the reviewer-spawn log by task session, keeping only the two kinds a delivery indicator
- *  reads. A NULL `outcome` is a legacy row or a spawn that never finalized: it is UNKNOWN, so it
- *  is excluded from every round count — counting it would inflate rework with crashed runs. */
+ *  reads. A NULL `outcome` is a legacy row, a spawn that never finalized, or one whose verdict was
+ *  discarded as superseded (#2175): it is UNKNOWN, so it is excluded from every round count —
+ *  counting it would inflate rework with crashed runs and with reviews of code already replaced. */
 function applyReviewSpawn(t: TaskRounds, sp: ReviewerSpawnRow): void {
   // The review STARTED regardless of how it ended, so every spawn moves the first-review mark.
   if (t.firstReviewAt == null || sp.spawnedAt < t.firstReviewAt) t.firstReviewAt = sp.spawnedAt;
--- a/src/review.ts
+++ b/src/review.ts
@@ -42,6 +42,7 @@ import {
   normalizePlanDrift,
   normalizePlanDriftNote,
   shouldSkipForPatchId,
+  headSuperseded,
   captureUsage,
   reapRun,
   VERDICT_FILE,
@@ -502,6 +503,13 @@ export class ReviewService {
     // Claim the slot synchronously, BEFORE begin()'s await, so a concurrent consider bails.
     this.starting.add(session.id);
     try {
+      // #2175: `git` is the poller's CACHED snapshot (up to a full idle sweep old), so a push can
+      // have landed since it was taken — the critic would then review superseded code and steer a
+      // findings round the author has already fixed. Re-read the live head and decline; the next
+      // poll picks the new head up normally (and that push reset CI to pending anyway, so the
+      // green-checks precondition has genuinely lapsed too). Deliberately INSIDE the claim: this
+      // await must not open a window for a second consider()/forceReview() to reach begin().
+      if (await this.headMoved(session, git.headSha!)) return "skipped";
       await this.begin(session, git, force);
     } finally {
       this.starting.delete(session.id);
@@ -512,6 +520,26 @@ export class ReviewService {
     // this return, so the non-force churn-skip "error" is irrelevant there; it's authoritative
     // only for the manual/force path.
     return this.inflight.has(session.id) ? "started" : "error";
+  }
+
+  /** Is the head we are about to review already superseded on the forge? One `prStatus` round-trip,
+   *  trivial next to the critic run it guards, and only reached once every cheap gate has passed.
+   *  Fails OPEN (see {@link headSuperseded}): no forge, or a forge that throws, spawns as before. */
+  private async headMoved(session: Session, headSha: string): Promise<boolean> {
+    const forge = this.deps.resolveForge(session.repoPath);
+    if (!forge) return false;
+    let live: PrStatus | undefined;
+    try {
+      live = await forge.prStatus(session.branch!);
+    } catch (err) {
+      console.warn(`[review] pre-spawn head recheck failed for ${session.id}:`, err);
+      return false;
+    }
+    if (!headSuperseded(headSha, live)) return false;
+    console.warn(
+      `[review] skipping ${session.id}: head moved ${headSha} → ${live!.headSha} before the spawn`,
+    );
+    return true;
   }
 
   private async begin(session: Session, git: GitState, force: boolean): Promise<void> {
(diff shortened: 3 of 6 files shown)


scoop (commented):
# Review — critic must never review a stale head (#2175)

Round 1 of at most 4. Scope: `git diff 5d37eb29...HEAD` (`src/critic-core.ts`, `src/delivery-metrics.ts`,
`src/review.ts`, `test/critic-core.test.ts`, `test/diagnostics.test.ts`, `test/review.test.ts`).

**Decision: comment.** Both guards the issue prescribes are present, correctly wired, and fail open on
every unconfirmable forge shape. I found no correctness or security defect. Notes below are
non-blocking.

## What I verified

### The predicate

`headSuperseded(reviewedSha, live)` at `src/critic-core.ts:1431` reads
`live?.state === "open" && !!live.headSha && live.headSha !== reviewedSha`. Checked against the type it
consumes: `PrStatus.headSha` is declared optional (`src/forge/types.ts:216-217` — "undefined when there
is no PR"), and `PrStatus.state` is the 4-member union `"none" | "open" | "merged" | "closed"`
(`src/forge/types.ts:200`). So the three fail-open shapes the plan claims (undefined status, absent/empty
`headSha`, non-`open` state) are exactly what the expression yields `false` for, and TS narrows `live` to
non-undefined after the optional-chained `=== "open"` comparison, so the subsequent `live.headSha` reads
are sound. The `PrStatus` type-only import added at `src/critic-core.ts:15` resolves to `src/forge/types.ts`
and is erased at compile time, so it introduces no runtime cycle.

Not pre-empting the existing moot handling matters and holds: `publishVerdict`'s merged arm
(`src/review.ts:1507-1521`) and `finalizeErrorVerdict`'s merged/closed suppression
(`src/review.ts:1425-1430`) both key on non-`open` states, and `headSuperseded` returns `false` for all of
them, so a merged-with-moved-head PR still routes to the post-merge comment path rather than being
silently dropped.

### Guard 1 — pre-spawn re-validation

`src/review.ts:503-516`. The call sits **inside** `try` and **after** `this.starting.add(session.id)`
(line 504), which is what makes it safe: `forceReview` bails on `this.starting.has` at
`src/review.ts:776` and `consider` bails on the same set at `src/review.ts:470`, both *before* any await,
so a second entrant during the recheck's suspension cannot reach `begin()`. The `finally` at line 514-516
releases the claim on the skip path. The early `return "skipped"` inside the `try` short-circuits the
`inflight.has(...) ? "started" : "error"` return at line 522, so a stale head surfaces as `"skipped"`, not
`"error"` — which is what `POST /api/sessions/:id/review-pr` relays to the UI
(`src/server.ts:3067-3068`, documented fail-closed at `src/server.ts:3055-3058`).

`headMoved` (`src/review.ts:528-543`) is reached only after every cheap gate — `git.headSha` is
non-null-checked at `src/review.ts:464` and `session.branch` at `src/review.ts:468`, so both `!`
assertions at lines 512 and 533 are backed by real guards. `criticEnabled`, the inflight/starting guard,
the same-head dedup and the spawn ceiling all precede it (`src/review.ts:469-502`), so a disabled or
already-settled repo pays no forge call.

**The recheck is genuinely live**, which is the whole point of the fix — verified rather than assumed:
`GithubForge.prStatus` shells out per call (`src/forge/github.ts:1441-1484` → `this.run`, whose default is
`defaultRunner` at `src/forge/github.ts:307-320`, an `execFile` of `gh` with no memoization; the only
cached field on that class is `cachedDefaultBranch`, `src/forge/github.ts:1592`). `LocalForge.prStatus`
reads `tipOf(headBranch)` live (`src/forge/local.ts:405`), and `GiteaForge.prStatus` issues a fresh
`GET /pulls` (`src/forge/gitea.ts:341-350`).

**Head-SHA form agrees between the poller's snapshot and the recheck.** The poller's batch path builds its
per-branch statuses through `mapGhPr` (`src/forge/github.ts:1545-1550`), and `prStatus` returns
`this.mapGhPr(pr, …)` too (`src/forge/github.ts:1483`); the REST fallback uses `pr.head.sha`
(`src/forge/github.ts:1342`) — all full 40-char SHAs from the same field, so the comparison can't be
tripped by an abbreviated/qualified form mismatch. That was the identifier-consistency risk here and it
checks out.

**No livelock.** The two auto callers pass a poller snapshot: `onSessionGit` (`src/index.ts:1741-1749`)
and `reKickReapedReview`'s `prPoller.get(id)` (`src/index.ts:1673-1683`). A skip only happens when the
snapshot differs from the live head — i.e. when the poller has *not yet* observed the new head — so the
next sweep necessarily produces a fresh `session:git` and re-drives `consider`. The manual path is fresher
still: `resolveGitState` fetches `forge.prStatus(session.branch)` directly (`src/server.ts:4197`) a moment
before `forceReview`, so the guard will essentially never trip there, and the hygiene reset persisted at
`src/review.ts:796-809` lands before `consider()` is called and therefore survives a skip — matching the
plan's stated decision and the test at `test/review.test.ts:1756-1764`.

### Guard 2 — at-finalize supersession discard

`src/review.ts:1291-1308`. One `livePr(f)` per finalize (`src/review.ts:1473-1482`) replaces the two
mutually-exclusive fetches that used to live in `finalizeErrorVerdict` and `publishVerdict`; I grepped
both consumers and they now take `live` as a parameter (`src/review.ts:1419-1424`,
`src/review.ts:1499-1506`) and no longer call the forge themselves — so finalize still costs exactly one
round-trip and head/state can't disagree. Behavioural equivalence of the moved fetch holds: the old
`publishVerdict` computed `state = (await forge?.prStatus(f.branch))?.state` inside a try/catch and
treated a throw as `undefined` → fail-closed emit-nothing; `livePr` returns `undefined` for both "no forge"
and "forge threw", and `publishVerdict:1506` still returns early on `!forge` and `:1522` on
`state !== "open"`. Same outcomes.

The discard genuinely writes nothing: `setReviewerSpawnOutcome`, the streak/error accounting,
`publishVerdict`/`finalizeErrorVerdict`, `putReview` and `onChange` all now live in `settleVerdict`
(`src/review.ts:1335-1404`), which the discard branch skips whole. `captureUsage`
(`src/review.ts:1314-1321`) and the `finally` teardown `onReviewing(false)` + `reapRun`
(`src/review.ts:1324-1327`) sit outside the branch, so cost attribution and terminal/worktree reaping still
happen — matching the plan's success criterion 3.

Leaving `reviewer_spawns.outcome` NULL is consistent with its only consumer: `applyReviewSpawn`
(`src/delivery-metrics.ts:93-104`) counts only `"error"` / `"clean"` / `"changes_requested"`, so a NULL row
is excluded from `errors`, `reviewed` and `changesRequested`; the comment update at
`src/delivery-metrics.ts:89-92` describes that accurately. It does still move `firstReviewAt`
(`src/delivery-metrics.ts:95-96`), which is right — the review did start. And the boot orphan sweep keys on
`row.completedAt`, not `outcome` (`src/review.ts:1915`), and `completeReviewerSpawn` still runs via
`captureUsage`, so a discarded run is not mistaken for a dangling row.

No stuck state after a discard: the prior `reviews` row keeps its older `headSha`, so
`headAlreadySettled` (`src/review.ts:1667-1672`, `prior?.headSha === headSha`) is false for the new head and
the next poll reviews it normally. Nothing downstream reads a review row that must be refreshed here —
the `review?.headSha === head` consumers at `src/index.ts:624` and `src/server.ts:4097` correctly keep
reporting the older reviewed head, which is now the honest answer.

### Limitations I could not resolve in-repo

- I did not run the suite (CI owns that, per the brief); my claims about test behaviour come from reading
  the fakes, not from execution.

## One design consequence worth the operator's attention (non-blocking)

The discard path is the only finalize outcome that increments **no** counter. Its siblings both do: a
findings verdict bumps `streakReviews` toward the `2*cap` spawn ceiling (`src/review.ts:1382-1392`, gated at
`src/review.ts:501-502`), and an error verdict bumps `errorRound` toward its own stall signal
(`src/review.ts:1439`, `src/review.ts:1455-1462`). A superseded run bumps neither and emits no signal, so if
a push reliably lands during every review (agent or human pushing on a ~review-length cadence), the critic
re-spawns indefinitely with nothing bounding the token spend.

I am deliberately **not** filing this as a defect: it is the explicit, operator-approved decision in the
plan ("one invariant: a review of superseded code leaves no trace"), each such spawn is caused by a real
new push rather than by static code, and the pre-change alternative — charging superseded reviews against
the findings ceiling until auto-review halts entirely — is worse. Flagging it only so the trade-off is on
the record; if it ever bites, a counter that escalates on repeated supersession (rather than one that
suppresses review) would be the shape to add.

## Nits (non-blocking):
- test/review.test.ts:2310 and test/review.test.ts:2430 still carry the annotation `// spawn while open (no prStatus call yet)`, which this PR made false — guard 1 now calls `prStatus` during `consider()`; both tests pass only because `headSuperseded` fails open on the `"merged"`/`"closed"` state the fake returns, which is worth saying instead.
- src/review.ts:1479 still logs `PR-state recheck failed` from `livePr`, which is now also the head recheck; `PR recheck failed` would match its widened role.
- src/review.ts:1397-1399 (moved into `settleVerdict`) says "captureUsage + the finally-reap below still run" — those are now in `finalize()`, not below this line in the same function.
- src/review.ts:525-527 the `headMoved` doc says "trivial next to the critic run it guards", but the guard also runs on the `spawnAborted` re-attempt path (`src/review.ts:1648-1652`), which re-enters `consider()` every poll while the account pool is cold and never spawns — one extra `gh` call per poll for those sessions.

- test/diagnostics.test.ts:2274 adds `...healthyDeps()` to the frozen-cell probe test — a real fix (`healthyDeps` is defined at test/diagnostics.test.ts:62 and used by every sibling), but unrelated to #2175 and outside the plan's file list.
- test/review.test.ts:195 and test/review.test.ts:297-300 add a `created` recorder for `worktree.createDetached` head SHAs — new harness surface beyond the plan's live-head box + `atHead()`, used by three of the new tests to assert no probe worktree was allocated.

- `src/standalone-critic.ts` has no equivalent staleness guard; the plan explicitly scopes it out and the file is untouched here.
- `src/plan-gate.ts` runs the same claim-then-spawn shape without a head recheck; not what #2175 reports, and untouched.
- `GithubForge.prStatus` resolves a branch with `--head <branch> --state all --limit 1` (`src/forge/github.ts:1459-1474`) while the poller's batch dedups a `headRefName` collision in favour of the expected owner (`src/forge/github.ts:1532-1552`); on a fork/branch-name collision the two could in principle name different PRs. Pre-existing divergence (the merged/closed recheck already depended on it), but this PR gives it a second consumer.

PR #2168, opened 2026-09-02
feat(eval): generalize the prompt eval harness to four prompts (#2156)

scoop:
Closes #2156.

Generalizes the stop-classifier eval (#1626/#1627) from one prompt to four — plan-gate reviewer, PR critic, rundown — re-enables the schedule, and gates PRs that change a prompt.

## The harness

`scripts/eval-core.ts` takes the reusable half of the classifier script. Three axes generalize:

- **Verdicts are labels plus predicates.** Only the classifier has an enum, so correctness became a per-fixture predicate set and `label` is now just the report's distribution. The classifier's single `kind === expectedKind` predicate is the degenerate case — its behaviour, report and JSON shape are unchanged.
- **The loop answers tool calls from an in-memory fixture environment.** `reviewPrompt` tells the reviewer the branch is checked out and orders `git diff <base>...HEAD` plus tree greps, which one API call cannot serve. `Bash`/`Read`/`Grep` are declared and answered from the fixture's `{ diff, files }` map, so the **production prompt is imported unchanged** and **no model-authored shell is ever executed**.
- **Termination is per-eval.** The critic's contract is *two* writes (`.shepherd-review.md` FIRST, `.shepherd-review.json` LAST as the completion signal), so `verdictFile` names the JSON and the markdown write is acknowledged mid-loop.

**Zero changes under `src/`** — no prompt prose is touched. This builds the measuring stick; it does not move what is being measured.

## Why the gate triggers on fingerprints, not paths

`src/plan-gate.ts`, `src/critic-core.ts` and `src/rundown-core.ts` took **45 / 20 / 15 commits in three months** — ~25 PRs a month, almost none touching prompt text. A path trigger would spend $1–2 on nearly all of them.

`scripts/gen-eval-fingerprints.ts` renders each builder over canonical inputs covering its conditional blocks, normalizes away the per-render fence nonces (without which every hash drifts each run), and commits `scripts/eval-fingerprints.json`. The PR job runs only the evals whose hash moved; `check:eval-fingerprints` in `ci.yml` stops a prompt edit shipping a stale fingerprint and skipping its own eval. The job is never `paths:`-filtered, per the #1859 warning that a path-filtered job reports no status and blocks merge forever.

## What actually gates today

| Eval | State |
|---|---|
| `stop-classifier` | **gating** — 95.1% / 91.8%, floor `0.80` |
| `plan-gate` | **gating** — 97.0% (32/33), floor `0.80` pinned 2026-09-09 |
[...]

Issue #2156, opened 2026-09-01
Generalize the eval harness: re-enable the schedule, fixtures for plan-gate/critic/rundown prompts


scoop:
Follow-up to #2151 (R4, `docs/research/ai-native-sdlc-playbook.md` §3 R4). R1 shipped in #2153.

The stop-classifier eval harness (`scripts/eval-stop-classifier.ts`, `docs/eval-stop-classifier.md`) is the template — labelled fixtures, multi-trial, thresholded. Three steps up the playbook's ladder:

1. **Re-enable the scheduled run** — its cron is currently commented out — and trigger it on PRs that touch the classifier prompt.
2. **Extend the pattern to the other high-leverage prompts** — plan-gate reviewer, critic, rundown — with small fixture sets (10–20 transcripts/diffs with expected verdicts), gating PRs that touch those builders.
3. **Incidents become fixtures**: a misfire captured in production becomes a labelled case in the set.

Complements the observational learnings lifecycle; explicitly NOT per-learning pre-ship eval gates (report §4 rules those out).

scoop:
Note for whoever closes this: #2156 asked for evals on the plan-gate reviewer, the PR critic **and the rundown**. #2174 (`feat(herd)!: remove the Herd Rundown`) removed that feature from main while PR #2168 was in review, so the rundown prompt no longer exists to measure and its eval was deleted with it — script, 10 fixtures, fingerprint cases and its captured baseline (which was 45/45, the only measured gating eval in the set).

The scope this issue can still deliver is therefore the harness plus the plan-gate and critic sets. Both currently ship `observational` (they run and report but cannot block) because neither has completed a measured run — see the pending-capture note in `docs/eval-harness.md`.

scoop:
✅ PR #2168 merged.

Issue #1627, opened 2026-07-11
feat: operator-language + input-robustness for the stop-classifier (eval-gated)


scoop:
Part 2b of epic #1616 (operator-language follow-up to #1586 / PR #1615).

## Blocked by #1626 (the live-model eval harness)

Do not start until the eval harness + baseline from #1626 exist and pass on current code. The eval is the gate: this change must not regress classification accuracy.

## Problem

With the eval in place, make the autopilot stop-classifier language-correct on **both** ends:

- **Output:** its persisted `summary` (`src/autopilot-llm.ts`) is operator-facing but always English.
- **Input:** its exemplars are English question strings, so a German terminal tail (produced once operator-language injection is active) erodes the classifier's ability to abstain — the `unknown` bucket is the conservative default it should fall back to, and mixed-language input pushes it toward a wrong confident guess.

## Vertical cut

- Inject the operator-language output directive so `summary` renders in the operator's language for `de`, while **pinning `kind` verbatim** — `normalize()` hard-matches `KINDS.includes` (`src/autopilot-llm.ts:138`), so a translated/off-enum `kind` silently collapses to `unknown`. Same discipline as recap's `verdict` enum.
- Add an **input-robustness** line so a non-English tail doesn't erode the `unknown` abstain bucket (the exemplars are English question cues the classifier pattern-matches against).
- Re-run the #1626 eval: overall accuracy and especially the ambiguous→`unknown` rate must be at or above the #1626 baseline.

## Acceptance

- The #1626 eval passes at/above baseline (no regression, particularly on ambiguous→unknown).
- `summary` renders in the operator's language for `de`; `kind` tokens are unchanged and always one of the exact `KINDS`.
- `en` behavior stays byte-identical for the prompt where feasible.

DAG: blocked by #1626.

PR #2177, opened 2026-09-02
fix(classifier): make the German abstain rule a positive no-ask test (#2169)

### LINKED ISSUE #2169: Stop-classifier: German ambiguous tails erode the `unknown` abstain bucket
Follow-up from #2156 (PR #2168), recorded per the contingency rule in `docs/eval-stop-classifier.md`.

## What was observed

The eval fixture `de-ambiguous-unknown` — a genuinely ambiguous **German** terminal tail that the classifier must abstain on (`unknown`) — has lost its majority:

| run | distribution | majority |
| --- | --- | --- |
| #1626/#1627 baseline | `unknown:9` | unknown ✔ |
| 2026-09-02 run 1 | `gate:2 unknown:7` | unknown ✔ |
| 2026-09-02 run 2 | `gate:5 unknown:4` | **gate �’** |

Its **English twin** `ambiguous-unknown` — same shape, same intent, same `T=9` — scored **9/9 `unknown` on both runs**. So this is not general nondeterminism: it is language-specific erosion of the abstain bucket, which is precisely what #1627's input-robustness directive was added to prevent.

This satisfies the noise band the doc defines (a ≥2-trial move that crosses the majority boundary AND survives a confirmation run), in the bad direction.

## Why it matters

`unknown` is the conservative surface-to-the-operator verdict. Eroding it toward `gate` means the autopilot proceeds on its own where it should have stopped and asked — the expensive direction of the error. From `docs/eval-stop-classifier.md`: "a wrongly-surfaced gate costs one click; a wrongly-answered question costs a bad product decision."

## What was already done

Per the contingency rule the fixture was **demoted to non-gating baseline** in PR #2168 and recorded as a known gap. It still runs at `T=9` and still reports, so it is the before/after datum for a fix. The floor was **not** lowered, and the classifier prompt was **not** touched (#2156 is harness work — it deliberately changes no prompt prose).

## What this issue is for

Decide whether to strengthen the German input-robustness directive in `classifierPrompt` (`src/autopilot-classify-core.ts`), and if so, use the eval to A/B it:

```bash
ANTHROPIC_API_KEY=… bun run eval:stop-classifier --trials 9 --filter ambiguous --operator-language-off --json  # before
ANTHROPIC_API_KEY=… bun run eval:stop-classifier --trials 9 --filter ambiguous --json                          # after
```

Promote `de-ambiguous-unknown` back to `gating: true` only once it holds majority across a confirmation run.

scoop:
Closes #2169.

Rewrites the German input-robustness directive in `classifierPrompt` after measuring that the line was not doing its job — and corrects two things the issue's premise got wrong.

## The diagnostic came back the other way round

`--filter ambiguous --trials 9`, both A/B legs, on the **unmodified** prompt. Pooled to 27 trials per condition (a fixture's pinned `trials` overrides `--trials`, so this is 3 × `T=9`, not one thick run):

| condition | `de-ambiguous-unknown` | runs |
| --- | --- | --- |
| OFF — no German lines at all | **24/27** (89%) | 8 / 7 / 9 |
| V0 — the directive as #1627 shipped it | **22/27** (81%) | 6 / 8 / 8 |
| `ambiguous-unknown` (en), same session | 25/27 (93%) | 9 / 9 / 7 |

The line added to *protect* the abstain bucket measured no better than injecting nothing. Zero `no-tool`/`parse-fail` anywhere, so every `unknown` is a genuine verdict.

## Two corrections to the record

**1. The prompt had already drifted under the 9/9 baseline.** That datum predates `561e577c` (#2002), which hoisted `UNTRUSTED_CONTENT_DIRECTIVE` out of `fenceUntrusted` — the tail fence went from ~250 chars of in-band prose to label + nonce only, landing directly on this directive's splice point. "The classifier prompt was not touched" is true of #2156's harness work but not of the interval between the two measurements.

**2. The erosion is not cleanly language-specific.** The English twin posted `gate:1 question:1 unknown:7` in the same session — its first sub-9/9 on record. Pooled, `en` 25/27 vs `de` 22/27 is well inside noise. The `9 → 7 → 4` history is the tail of a wide distribution, and the `4/9` that triggered the issue was its extreme.

## What changed

The closing clause was an abstract instruction about the model's own confidence — *"never upgrade an uncertain read to a confident `gate` or `question` just to avoid abstaining"* — naming `gate` twice inside a negation. It is now a **positive, checkable test applied to the tail**: `gate` and `question` both presuppose the agent RAISED something, so a tail that only narrates progress can be neither.

The trailing scoping clause is **load-bearing, not hedging**. Without it the rule collapses to "raises no question → `unknown`", which swallows the legitimately question-free `finished`/`complete` tails — `de-finished-pr` asks nothing either. That trades the `finished` bucket for the `unknown` one: the same erosion, inverted. A test pins the clause so it can't be silently shortened.

## Results

Screen: **27/27** on `de-ambiguous-unknown`. Since nothing beats a perfect screen, the remaining candidates (anchor-move-only, removal) were not run — recorded in the doc.

Validation, full German set, twice:

| id | seg | exp | T | run A | run B |
| `de-gate-commit` | gating | gate | 9 | `gate:9` — 9/9 | `gate:9` — 9/9 |
| `de-question-approach` | gating | question | 9 | `question:9` — 9/9 | `question:9` — 9/9 |
| `de-ambiguous-unknown` | gating | unknown | 9 | `unknown:9` — 9/9 | `unknown:9` — 9/9 |
| `de-gate-spec` | baseline | gate | 5 | `gate:5` — 5/5 | `gate:2 question:3` — 2/5 |
| `de-finished-pr` | baseline | finished | 5 | `finished:5` — 5/5 | `finished:5` — 5/5 |

- Gating accuracy **27/27 = 100%** in both runs, `RESULT: PASS`.
- **45/45** `unknown` on the target fixture across screen + both validation runs, vs 22/27 for the shipped directive.
- **No bucket trading** — `de-finished-pr` held `finished:5` both times.
- `de-gate-spec` (non-gating known gap) across four runs: 5/5, 2/5, 4/5, 2/5 = 13/20, straddling its own 4/5 baseline and its English twin's 2/5. No claim that this change moved it.

## Scope and limits

- **`en` is byte-identical**, verified by diffing the rendered prompt against pre-change `HEAD` (nonce normalized) for both the default and explicit `"en"` calls — re-verified after the prettier hook reformatted the file. The English baseline survives by construction; no English calls were purchased.
- **The rule is not German-specific but ships only on the `de` path.** "A tail that raises nothing is not a `gate`" would arguably help English too, but the `en` prompt must stay byte-identical and the English twin has nothing to fix. The two prompts are now semantically divergent, not merely translated — recorded deliberately.
- `GATING_ACCURACY_FLOOR` **unchanged** at `0.80`. The adjustment rule would permit `0.85`, but one fixture set measured twice is a thin basis for tightening a catastrophe-catcher.
- The doc's **noise band is amended**: `de-ambiguous-unknown` produced `9, 7, 4, 6, 8, 8` out of 9 on the *same* prompt, so "≥2-trial move surviving a confirmation run" is reachable by sampling alone — #2169's own trigger criterion was met by a fixture whose pooled rate turned out ordinary. Pool ≥27 trials per condition, same session.

## Note for #2168

That PR is still open and touches zero files under `src/`, so there is no code conflict. But it commits `scripts/eval-fingerprints.json`, and this PR changes `classifierPrompt` — the classifier fingerprint will be stale after it rebases, and `check:eval-fingerprints` will (correctly) fail until regenerated. On `main` `de-ambiguous-unknown` is still `gating: true`, so nothing needed re-promoting here; #2168's demotion of it should be reconsidered in light of these numbers.

## Verification

`bun run lint`, `bunx tsc --noEmit`, `bun run test` (8879 pass, 0 fail), `bunx fallow audit --base origin/main --fail-on-issues` (✓ no issues in 3 changed files) — all green.

### DIFF
--- a/docs/eval-stop-classifier.md
+++ b/docs/eval-stop-classifier.md
@@ -171,11 +171,10 @@ sees is byte-identical).
 - **Before (German fixtures, operator-language OFF)** — the #1626 baseline already recorded it:
   `de-gate` 4/5, `de-question` 5/5, `de-finished` 5/5 (English prompt against a German tail). Re-run
   it exactly with `--operator-language-off`.
-- **After (German directive live)** — **PENDING capture.** Run both legs (locally with a key, or via
-  the workflow) and transcribe the `--json` here. The gate: every German gating fixture stays
-  majority-correct at `T=9`, gating accuracy ≥ floor, and `de-ambiguous-unknown` holds `unknown`
-  majority. This eval is **manual/nightly, never a per-PR gate** (paid, keyed, nondeterministic), so
-  the PR declares the after-run as a manual step and must not merge until it is green.
+- **After (German directive live)** — captured under **[#2169](#german-abstain-bucket-rewrite-2169)**,
+  which also rewrote the directive being measured. Every German gating fixture is majority-correct at
+  `T=9` and gating accuracy is `27/27 = 100%` across two runs. This eval is **manual/nightly, never a
+  per-PR gate** (paid, keyed, nondeterministic).
 
 ### Noise band (react to signal, not model noise)
 
@@ -183,6 +182,14 @@ At `temperature = 1.0` a single-run `T=5` majority can flip on one trial of samp
 German gating fixtures run at **`T=9`**, and a shift between legs counts as **signal** only when it
 **crosses the majority boundary** OR moves by **≥2 trials**, AND survives a **confirmation re-run**. A
 lone ±1-trial wobble is noise — never reword the directive in response to it.
+
+> **#2169 amendment — this band is too narrow for `T=9`.** `de-ambiguous-unknown` produced `9, 7, 4,
+6, 8, 8` out of 9 on the _same_ prompt, so a "≥2-trial move surviving a confirmation run" is
+> reachable by sampling alone: the #2169 criterion was met in the bad direction by a fixture whose
+> pooled rate turned out to be ordinary. Treat a single `T=9` run as a coarse filter only. To
+> compare two prompt variants, **pool ≥27 trials per condition** (a fixture's pinned `trials`
+> overrides `--trials`, so pool repeated runs rather than raising the flag) and measure both
+> conditions in the same session — model-side drift between sessions is not controlled for.
 
 ### Verification split — what the eval does and does NOT cover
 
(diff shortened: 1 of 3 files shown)


scoop (commented):
# Review — fix(classifier): make the German abstain rule a positive no-ask test (#2169)

Three files: one prompt-prose constant (`src/autopilot-classify-core.ts`), its unit pins
(`test/autopilot-llm.test.ts`), and the eval doc. No blocking defect found.

## What I verified

**The shipped German string concatenates cleanly and contains both test sentinels.**
`src/autopilot-classify-core.ts:73-79` assembles to:

> "The terminal tail above may be written in German or a mix of German and English. Classify by what
> the agent actually MEANS, not by matching English phrasing — a non-English tail is never itself a
> reason to choose a kind. Both "gate" and "question" require the agent to have actually RAISED
> something: a tail that only narrates progress or announces that it is carrying on has asked you
> nothing, so it is neither. If such a tail also does not clearly report finished or delivered work,
> answer "unknown"."

Both sentinels span a concatenation boundary and still resolve: `require the agent to have actually
RAISED something` spans `src/autopilot-classify-core.ts:76`→`:77`, and `does not clearly report
finished or delivered work` spans `:78`→`:79`. Both are the exact strings asserted at
`test/autopilot-llm.test.ts:100` and `:109`. Quoting is consistent (single-quoted JS literals for the
two lines carrying `"gate"` / `"unknown"`, double-quoted elsewhere), and the inter-line spacing
produces no doubled or missing space.

**The `en` byte-identity guard did not go vacuous** — the plan's stated risk. `test/autopilot-llm.test.ts:87-88`
asserts `not.toContain("in German")` and `not.toContain("may be written in German")`; the rewritten
line still opens with "The terminal tail above **may be written in German**" (`src/autopilot-classify-core.ts:74`),
so both guards remain load-bearing. The `en` prompt itself is untouched: `classifierPrompt` splices
both German constants only under `de` (`src/autopilot-classify-core.ts:115`, `:124`), and neither
splice site moved in this diff.

**Nothing else in the tree referenced the removed prose.** Grepped `avoid abstaining` / `never
upgrade an uncertain read` across the repo: the only hits are the new explanatory comment
(`src/autopilot-classify-core.ts:56-57`), the doc (`docs/eval-stop-classifier.md:258-259`) and the
test comment (`test/autopilot-llm.test.ts:97`) — all deliberate historical quotes, no live assertion
left pinned to the old wording. `test/untrusted.test.ts:109` builds `classifierPrompt(["tail"], "task")`
(the `en` path) for its directive check, so it is unaffected. `test/eval-stop-classifier.test.ts:232-254`
asserts fixture *shape* only, not prompt prose. No `scripts/eval-fingerprints.json` exists on this
branch, so there is no prompt-hash artifact left stale by the edit.

**The behavioral direction of the rewrite is the one the repo's own ground truth wants.** The rule
pushes progress-narration tails away from `gate` toward `unknown`. `scripts/eval-stop-classifier.ts:240-250`
labels exactly such a German tail (`"Mit dem ersten Teil fertig. Ich mache weiter."`) as
`expectedKind: "unknown"`, `gating: true`, and its English twin at `:181-193`
(`"Done with the first part. Moving on."`) is labelled identically and already holds 9/9 without any
such rule. So this aligns the `de` path with existing `en` behavior rather than introducing a new
divergence in outcomes. The cost side is bounded: `src/autopilot.ts:399-401` routes both `question`
and `unknown` to the same `pause`/surface path, so any `question`→`unknown` slippage from the new
wording is behaviorally inert; only `gate`→`unknown` changes anything, and that is the fix.

**The scoping clause protects the bucket the comment says it protects.** `de-finished-pr`
(`scripts/eval-stop-classifier.ts:253-264`) is a question-free German tail that reports completed
work — it escapes the rule via "does not clearly report finished or delivered work", and the doc
records `finished:5` in both validation runs. The comment at `src/autopilot-classify-core.ts:64-67`
accurately describes what the clause does.

**The doc's harness claims check out against the harness.** "A fixture's pinned `trials` overrides
`--trials`" (`docs/eval-stop-classifier.md:190-191`, `:275`) is exactly `fixture.trials ?? defaultTrials`
at `scripts/eval-stop-classifier.ts:538-539`. `--filter` is a substring match on the fixture id
(`scripts/eval-stop-classifier.ts:576-578`), so `--filter ambiguous` selects the two abstain fixtures
and `--filter de-` selects the five German ones — matching the five rows and the `T` column
(9/9/9/5/5) of the validation table at `docs/eval-stop-classifier.md:277-283`. The `27/27` gating
denominator is the three `gating: true` German fixtures at `T=9`
(`scripts/eval-stop-classifier.ts:219-220`, `:234-235`, `:247-248`) — arithmetically right.

**The task is satisfied, including its promotion clause.** The issue asks to promote
`de-ambiguous-unknown` back to `gating: true` once it holds. On this branch it is already
`gating: true` (`scripts/eval-stop-classifier.ts:247`) — the demotion the issue describes lives on
PR #2168's branch, not here — so there is nothing to re-promote and no floor was touched
(`GATING_ACCURACY_FLOOR = 0.8` unchanged at `scripts/eval-stop-classifier.ts:66`).

**The `#2002` attribution in the new doc section is correct.** `docs/eval-stop-classifier.md:220`
credits commit `561e577c` with hoisting `UNTRUSTED_CONTENT_DIRECTIVE` out of `fenceUntrusted`. That
squashed commit's body contains "fenceUntrusted carries label + nonce only … Refs #2002", so the
issue number is right (the PR number was #2018 — the plan's phrasing — but the doc cites the issue,
consistent with the pre-existing `#2002` marker at `src/autopilot-classify-core.ts:106`).

## Stated limitations

- Whether the model *obeys* the rewritten line is not verifiable from this repo; the review takes the
  measured distributions in `docs/eval-stop-classifier.md:233-296` and the commit body as reported
  and does not attempt to reproduce them (the eval is paid, keyed, and nondeterministic).
- Whether `prettier --check` would actually rewrite the blockquote below is inferred from repo
  evidence, not from running prettier (CI-only per the review instructions).

## Nits (non-blocking):

- `docs/eval-stop-classifier.md:187` — the second line of the `#2169 amendment` blockquote is missing its `>` marker (lazy continuation). A multiline grep over `docs/*.md` finds this as the *only* such line in the directory, so the file looks like it skipped the prettier pass; `bunx prettier --check .` at `.github/workflows/ci.yml:99` covers `docs/**/*.md` (not in `.prettierignore`) and will likely flag it.
- `docs/eval-stop-classifier.md:186-187` — the amendment lists `9, 7, 4, 6, 8, 8` as produced "on the _same_ prompt", but the subsection at `:219-228` argues the `9` was measured *before* `561e577c` changed the prompt. The remaining five (7, 4, 6, 8, 8) still carry the conclusion; the `9` shouldn't be inside that list.
- `src/autopilot-classify-core.ts:77-78` — "has asked **you** nothing" addresses the classifier, but the agent asks the *operator*, not the classifier. "has asked nothing" reads cleaner; measured behavior is fine either way.
- `docs/eval-stop-classifier.md:290-291` — "No bucket trading" is evidenced by `de-finished-pr` only. The rule's trailing clause also guards `complete` ("delivered work"), and the German fixture set has no `complete` fixture (`scripts/eval-stop-classifier.ts:194-265`), so that half of the claim is unmeasured; worth one clause in "Deliberate limits" rather than a new fixture.

- `docs/eval-stop-classifier.md:186-192` — the `#2169 amendment` rewrites the doc's **noise-band methodology** (single `T=9` run demoted to "coarse filter", new "pool ≥27 trials per condition" rule). The plan scoped the doc change to recording the diagnostic legs, the chosen wording, the runs and the resolved gap; amending the governing methodology for future work goes beyond that. It is well-argued and I would keep it — noting it as the judgement call it is.
- `docs/eval-stop-classifier.md:174-177` — the rewrite of the `#1627` "After" bullet also drops that section's "must not merge until it is green" merge condition. Correct now that the capture exists, but it is a second, separate edit riding along with the `#2169` record.

- `src/autopilot-classify-core.ts:115` — Anchor A splices the German line *before* the enum block at `:117-122`, so it names `gate`/`question`/`unknown` before the prompt defines them. The plan's V1 (anchor move) was never screened because V2 hit a perfect score; the ordering remains as it was.

Issue #2157, opened 2026-09-01
Maintain loop that opens work: tiered thresholds on Shepherd's own health signals


scoop:
Follow-up to #2151 (R5, `docs/research/ai-native-sdlc-playbook.md` §3 R5). R1 shipped in #2153 and provides the signal substrate.

Shepherd notifies but never remediates. The playbook's maintain loop closes that: tiered thresholds on Shepherd's own health signals —

1. log,
2. a read-only diagnosis spawn that drafts a backlog issue,
3. a PR for a pre-approved class of fix.

Reuses the reviewer-preset argv sandbox (`--permission-mode dontAsk` + scoped `--allowedTools`, never `--dangerously-skip-permissions`).

Scope guard: **internal signals only, no production ingress.** Report §4 explicitly rules out external production control bands.

scoop:
✅ PR #2172 merged.

PR #512, opened 2026-06-10
feat(herd): full-bleed session list on mobile

kai-osthoff:
## Summary

On the mobile (PWA) list view, the session cards were wrapped in the panel's "hug" — 1px side borders + corner brackets, plus the shell's 10px and the list's 6px horizontal gutters. On a phone that costs ~21px of width per side that could be card text instead.

This makes the herd panel **full-bleed in flow mode** (the mobile list):

- drop the panel's inline borders and the corner brackets (`.panel.flow`)
- zero the `.units` inline padding
- stretch the panel into the shell's 10px horizontal padding via `margin-inline: -10px` (with a larger safe-area inset the shell keeps the remainder, so nothing slides under a notch in landscape)
- top/bottom hairlines stay as section rules; the desktop sidebar (`flow=false`) is untouched

### Before / after

Card text now starts ~14px from the screen edge instead of ~31px — roughly **34px more width** for titles and prompt excerpts.

`[no-feature-entry]` — pure layout/space optimization, no new user-facing capability to announce.

## Verification

- `cd ui && bun run check` — 0 errors, 0 warnings
- Playwright screenshot at iPhone viewport (390×844): cards span edge to edge, section rules run full width, no stray brackets

PR #2179, opened 2026-09-02
feat(maintain): tier 3 opens a PR for the dead-code fix class (#2171)

### LINKED ISSUE #2171: Maintain loop Tier 3: open a PR directly for a pre-approved fix class
Follow-up to #2157 (which shipped tiers 1 and 2 of the playbook's maintain loop — see `docs/research/ai-native-sdlc-playbook.md` §3 R5).

`MaintainService` today logs a Tier-1 reading and, for a Tier-2 breach, spawns a read-only diagnosis agent whose drafted issue the server files. The playbook's third tier is missing: **for a pre-approved class of fix, open the PR directly** instead of an issue.

It was deliberately left out of #2157 rather than built speculatively: the only pre-approved class the report names is doc drift, and the doc agent already owns that end to end. A Tier-3 path built then would have had no user.

**What would make this worth doing:** a concrete, narrow fix class that (a) a read-only diagnosis can identify unambiguously, (b) is mechanical enough that a scoped agent can produce the diff, and (c) the operator is willing to see arrive as a PR without an issue first. Candidates worth weighing when one appears — a band whose remediation is a config/threshold edit, a recurring lint/dead-code class, a stale generated artifact.

**Shape it would take** (the machinery is already in place):
- A per-band `tier3: { class }` in the threshold table, absent by default.
- A third tier in `evaluateBands` / `MaintainTier`.
- The doc agent's publish tail — worktree, commit, push, `forge.openPr()` — rather than `createIssue`. Never auto-merge.
- Its own escalation flag beyond `SHEPHERD_MAINTAIN_ACT`, so arming issue-filing never implicitly arms PR-opening.
- `reapOrphans` and the `inflightWorktrees()` protection already cover the run.

**Do not** generalise Tier 3 ahead of a named fix class — that is exactly the speculative surface #2157 declined to build.

scoop:
Closes #2171.

#2157 shipped tiers 1 and 2 and deliberately left tier 3 out: no fix class had been named, and building one speculatively was exactly the surface it declined. This names one.

## The fix class: repo-wide auto-fixable dead code

It meets #2171's three criteria:

- **Unambiguously identifiable** — `fallow dead-code --format json` emits a schema-versioned report with a per-finding `auto_fixable` flag.
- **Mechanical** — `fallow fix` produces the diff deterministically.
- **Fit to arrive as a PR** — CI only runs `fallow audit --base origin/<base>`, a **delta** audit, so residue accumulates on `main` with no gate watching. Three findings are sitting there right now (2 unused `.svelte` files, 1 unused export).

## Three decisions worth reviewing

**Tier 3 is a class marker, not a fourth threshold.** #2171's own shape is `tier3: { class }` — no numbers. So `evaluateBands` *promotes* a tier-2 breach to 3 when its band declares a fix class. A higher threshold would be backwards: you would need *more* dead code to earn the cheap mechanical fix. `tier3` lives on `CountBandConfig` alone — the three v1 bands get no inert `tier3?` field.

**No agent.** `fallow fix` is deterministic, so the run is a plain async routine on the server: no spawn, no tokens, no prompt to inject into, no new write-capable role allowlist, and no `reviewer_spawns` cost row. The band counts only `auto_fixable` findings, so the fix drives it to zero by construction. Unused *files* stay a human's call — fallow itself marks `delete-file` unsafe ("may remove runtime functionality not visible to static analysis").

**`SHEPHERD_MAINTAIN_PR` is independent of `SHEPHERD_MAINTAIN_ACT`**, per the issue. Unarmed, the run still installs, fixes and verifies, then logs the *real* diff it would have opened and throws the branch away — so the observe log is truthful rather than speculative.

## The run

Branch worktree off `origin/<base>` → install → re-measure → `fallow fix` → verify → commit `--no-verify` → push → `openPr`. **Never auto-merges.**

## Two things measured, not assumed

**`bun install` is load-bearing.** In a fresh worktree of `origin/main` *without* `node_modules`, `fallow dead-code` reports **14** findings (6 unused files, 6 unused exports) against **3** with root + `ui/` + `extension/` installed. Skipping it would have deleted live exports. Cost is ~4.3s with `--frozen-lockfile`, which also leaves the tree clean so any post-fix manifest change is unambiguously fallow's.

**`fallow dead-code` exits 1 whenever it finds anything** — verified end-to-end, before *and* after a real fix (the two non-auto-fixable files remain). The first cut of the runner treated that rejection as "could not measure", which would have left the band reading no-data exactly when it matters and tier 3 permanently unreachable. `reportFromFailedExit` keeps stdout from a non-zero exit but discards it from a **killed** process, whose truncated JSON `jsonrepair` would close into a shape-valid report with a smaller count that the verify gate could read as clean.

## Fail-closed gates — each opens nothing and records `error`

Install failure · red `bun run typecheck` · any auto-fixable finding surviving the fix · an unreadable re-measurement · **any `package.json` or lockfile the fix touched** (`fallow fix` also removes unused *dependencies*, and that regen is a human's job, not a background loop's).

If `openPr` throws after the push, the branch is deleted from the remote rather than orphaned. `EmptyDiffError` records `skipped`, not `error`.

## Verification

Root `lint` / `tsc` / 9031 tests · `ui` svelte-check 0 errors / 4773 tests · i18n parity · docs manifest · glossary · `fallow audit --base origin/main --fail-on-issues` clean · full pre-push gate green.

One note: `StatusPip.browser.test.ts` flaked twice on hover timing across ~5 full UI runs and passes in isolation. It is pre-existing and untouched by this diff.

### DIFF
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -289,10 +289,16 @@ tokens and has no prompt to be injected into. The run:
 3. re-measures in that pristine checkout and stands down if there is nothing to fix (the
    sweep's reading came from the live checkout, which can carry uncommitted edits);
 4. runs `bunx fallow@<pinned> fix --yes --no-create-config`;
-5. **verifies fail-closed** — `bun run typecheck` must pass and a re-run of `fallow dead-code`
-   must report no auto-fixable findings left. It also **refuses** to commit when the fix touched
-   any `package.json` or lockfile: `fallow fix` removes unused *dependencies* too, and that needs
-   a lockfile regen a background loop must not do;
+5. **verifies fail-closed** — the type-check of **every package the diff touches** must pass, and
+   a re-run of `fallow dead-code` must report no auto-fixable findings left. Per package, not just
+   the root: `bun run typecheck` is `tsc` against a tsconfig that excludes `ui`, `extension`,
+   `site` and `docs-site`, while fallow analyses all of them, so a root-only gate would pass
+   vacuously for a fix under `ui/src`. The root uses `bun run typecheck`; `ui` and `extension` use
+   their own `bun run check`. The run also **refuses** to commit when the fix touched any
+   `package.json` or lockfile (`fallow fix` removes unused *dependencies* too, and that needs a
+   lockfile regen a background loop must not do), or anything under `site/` or `docs-site/`, whose
+   dependencies are not installed in the fix worktree and which therefore cannot be verified at
+   all;
 6. commits `--no-verify`, pushes, and opens the PR.
 
 Any failed gate opens nothing, records an `error` outcome and throws the branch away. **Nothing
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -247,36 +247,69 @@ one (a per-repo in-flight guard means at most one run per repo at a time):
 
 ## Maintain loop (self-health bands)
 
-Opt-in, default-off, and fully inert when off. Once per local day Shepherd scores three
+Opt-in, default-off, and fully inert when off. Once per local day Shepherd scores four
 **bands** over its **own** health data and escalates by tier: **tier 1** logs the reading,
 **tier 2** spawns a read-only diagnosis agent that drafts a backlog issue which the trusted
 server files **against Shepherd's own repo** — never a managed repo, so nothing lands in
 someone else's backlog. The agent itself never touches a forge: it writes a JSON draft in a
-disposable worktree and nothing else. There is no tier 3 (no automatic fix PR).
+disposable worktree and nothing else. **Tier 3** skips the issue and opens a **pull request**
+— see below.
 
 | Band | Measures | Window | Tier 1 | Tier 2 |
 | --- | --- | --- | --- | --- |
 | `critic_error_rate` | Share of outcome-bearing review spawns that errored (produced no verdict) | 7 days, min sample 10 | ≥ `0.15` | ≥ `0.30` |
 | `incident_spike` | `signals` per kind — needs **both** an occurrence count and a distinct-session count, so one thrashing task can't trip it. The `reply` kind is excluded (operator corrections are high-volume by design) | 7 days | ≥ 10 occurrences **and** ≥ 3 sessions | ≥ 25 occurrences **and** ≥ 5 sessions |
 | `first_pass_collapse` | Per-repo first-pass review rate — direction is **inverted**, a lower rate is worse | 30 days, min sample 8 | ≤ `0.60` | ≤ `0.40` |
+| `dead_code_drift` | Auto-fixable dead-code findings in Shepherd's own checkout (`fallow dead-code`). Point-in-time, no window and no minimum sample. Declares a tier-3 fix class, so its tier-2 breach is **promoted to tier 3** | now | ≥ 1 finding | ≥ 3 findings |
 
 A band below its minimum sample reports "below min sample" rather than a misleading number.
 Every band's live value is persisted and surfaced on the **Delivery lens** whether or not it
 breached — the starting thresholds are calibrated guesses, and observed values are what let
 you retune them.
 
-**Spend bounds.** At most **one** tier-2 diagnosis spawn per sweep (the rest wait for the next
-day), and after a diagnosis run **completes** — filed, skipped or errored — its band is
-suppressed for **14 days**. The cooldown anchors on the run, not on a filed issue, precisely
-so observe mode (which files nothing) still suppresses.
+**Spend bounds.** At most **one** band action per sweep — a tier-2 diagnosis spawn *or* a
+tier-3 fix, whichever band is more severe; the rest wait for the next day. After a run
+**completes** — published, skipped or errored — its band is suppressed for **14 days**. The
+cooldown anchors on the run, not on a published issue or PR, precisely so observe mode (which
+publishes nothing) still suppresses. A still-open issue or PR from the band's last run extends
+the suppression past the cooldown.
 
-**Phased soak (`observe → act`)**, mirroring the doc agent:
+### Tier 3 — the pre-approved fix class
+
+A band may declare a **pre-approved fix class**: a remediation mechanical enough that no
+judgement is needed, so the loop produces the diff itself and opens a PR instead of asking you
+to read a drafted issue first. Exactly one class exists — `dead_code`, on `dead_code_drift`.
+
+**No agent is involved.** The fix is `fallow fix`'s verbatim output, so a tier-3 run costs no
+tokens and has no prompt to be injected into. The run:
+
+1. creates a branch worktree `shepherd/maintain-fix-<8hex>` off `origin/<default branch>`;
+2. runs `bun install --frozen-lockfile` in the root, `ui/` and `extension/` — **load-bearing**:
+   without installed dependencies fallow cannot resolve imports and reports live code as dead;
+3. re-measures in that pristine checkout and stands down if there is nothing to fix (the
+   sweep's reading came from the live checkout, which can carry uncommitted edits);
+4. runs `bunx fallow@<pinned> fix --yes --no-create-config`;
+5. **verifies fail-closed** — `bun run typecheck` must pass and a re-run of `fallow dead-code`
+   must report no auto-fixable findings left. It also **refuses** to commit when the fix touched
+   any `package.json` or lockfile: `fallow fix` removes unused *dependencies* too, and that needs
+   a lockfile regen a background loop must not do;
+6. commits `--no-verify`, pushes, and opens the PR.
+
+Any failed gate opens nothing, records an `error` outcome and throws the branch away. **Nothing
+is ever auto-merged.** If `openPr` fails after the push, the branch is deleted from the remote
+rather than left orphaned.
+
+**Phased soak (`observe → act → pr`)**, mirroring the doc agent:
 
 1. **Observe** — `SHEPHERD_MAINTAIN_LOOP=1` alone. Bands are evaluated, readings persisted,
    breaches logged and the tier-2 diagnosis spawns, but finalize is **log-only**: it logs
    `[maintain] <band>: would file issue "<title>" (act is off)` and calls the forge never.
 2. **Act** — additionally set `SHEPHERD_MAINTAIN_ACT=1` to escalate finalize to actually
    opening the labelled issue. Meaningful only when `SHEPHERD_MAINTAIN_LOOP` is also on.
+3. **PR** — additionally set `SHEPHERD_MAINTAIN_PR=1` to let a tier-3 run open its pull
+   request. **Independent of `SHEPHERD_MAINTAIN_ACT`**: arming issue-filing never implicitly
+   arms PR-opening. Without it a tier-3 run still installs, fixes and verifies, then logs
+   `would open a PR removing …` and discards the branch — so the log names the real diff.
 
 `POST /api/maintain/sweep` runs an evaluation on demand (it `404`s when the flag is off, the
 same unadvertised contract as the doc-agent route). It skips the hour/presence/once-a-day
@@ -287,11 +320,12 @@ its worktree reclaimed by the boot reconcile; the breach is re-diagnosed on the 
 | --- | --- | --- |
 | `SHEPHERD_MAINTAIN_LOOP` | `0` (off) | Set `1` to arm the loop (**observe**): band evaluation, tier-1 logging, and the tier-2 read-only diagnosis spawn. The drafted issue is only **logged** until `SHEPHERD_MAINTAIN_ACT` is also set |
 | `SHEPHERD_MAINTAIN_ACT` | `0` (off) | **Act.** Set `1` to escalate finalize to actually filing the labelled issue against Shepherd's own repo. Meaningful only when `SHEPHERD_MAINTAIN_LOOP` is also on |
+| `SHEPHERD_MAINTAIN_PR` | `0` (off) | **Tier 3.** Set `1` to let a tier-3 fix open a pull request against Shepherd's own repo. Never auto-merges. Meaningful only when `SHEPHERD_MAINTAIN_LOOP` is also on, and **independent of** `SHEPHERD_MAINTAIN_ACT` |
 | `SHEPHERD_MAINTAIN_HOUR` | `4` | Local hour (0–23) at/after which the once-a-day band sweep may run — an hour after the doc agent's nightly so the two spawns don't land together. Invalid values fall back to `4` |
 | `SHEPHERD_MAINTAIN_CLI` | `inherit` | Agent CLI for the diagnosis spawn: `inherit` follows the global default provider, or pin `claude` / `codex`. Env-only (not persisted or UI-configurable) |
 | `SHEPHERD_MAINTAIN_MODEL` | `default` | Model for the diagnosis spawn: `default` follows the global default model, or pin a `<model alias>`. Env-only |
 | `SHEPHERD_MAINTAIN_EFFORT` | `default` | Reasoning-effort tier for the diagnosis spawn: `default` follows the CLI's own effort, or pin a tier (`low` / `medium` / `high` / `xhigh` / `max`). Env-only |
-| `SHEPHERD_MAINTAIN_THRESHOLDS` | _(unset)_ | JSON object deep-merged over the threshold table above, so a recalibration ships without a deploy. Parsed field-by-field and **fail-soft**: an unparseable value or a typo in one number falls back to that default rather than disarming a band. E.g. `{"critic_error_rate":{"tier1":0.2}}` |
+| `SHEPHERD_MAINTAIN_THRESHOLDS` | _(unset)_ | JSON object deep-merged over the threshold table above, so a recalibration ships without a deploy. Parsed field-by-field and **fail-soft**: an unparseable value or a typo in one number falls back to that default rather than disarming a band. E.g. `{"critic_error_rate":{"tier1":0.2}}`. Retunes numbers only — a band's tier-3 fix class is not overridable, because `SHEPHERD_MAINTAIN_PR` is the one switch that disarms tier 3 |
 
 ## Anonymous usage telemetry
 
--- a/docs-site/src/content/docs/reference/glossary.md
+++ b/docs-site/src/content/docs/reference/glossary.md
@@ -86,14 +86,17 @@ drifts constantly is usually writing vague plans, not writing bad code.
 
 A daily check of Shepherd's own health signals against declared thresholds. A
 mild breach is logged; a sustained one spawns a read-only agent that drafts a
-backlog issue for you to triage. It watches Shepherd itself — never a production
-system.
+backlog issue for you to triage. For one pre-approved class of mechanical fix it
+skips the issue and opens a pull request instead. It watches Shepherd itself —
+never a production system.
 
 ### Band
 
 One health signal plus the thresholds that decide what happens when it is
-crossed: tier 1 logs a reading, tier 2 spawns a diagnosis. A band below its
-minimum sample size reports nothing rather than a misleading number.
+crossed: tier 1 logs a reading, tier 2 spawns a diagnosis. A band whose
+remediation is mechanical enough to be pre-approved is promoted to tier 3
+instead, which opens a pull request. A band below its minimum sample size reports
+nothing rather than a misleading number.
 
 ### Inferred
 
--- a/src/config.ts
+++ b/src/config.ts
@@ -744,6 +744,11 @@ export const config = {
   // Phase-1 escalation; meaningful only with `maintainLoopEnabled`. When off, finalize logs the
   // issue it WOULD file and calls createIssue never.
   maintainLoopAct: process.env.SHEPHERD_MAINTAIN_ACT === "1",
+  // Tier-3 escalation (#2171); meaningful only with `maintainLoopEnabled`. When off, a Tier-3 fix
+  // runs and verifies in its worktree, logs the PR it WOULD open, and throws the branch away.
+  // DELIBERATELY INDEPENDENT of `maintainLoopAct`: arming issue-filing must never implicitly arm
+  // PR-opening, so this is its own flag rather than a second meaning for that one.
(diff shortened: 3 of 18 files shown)


scoop:
# Review — Maintain loop Tier 3: open a PR for the dead-code fix class (#2171)

Scope: `git diff 4d92a773528c05ffffb86c55f33b6ed0ea5447d7...HEAD` (18 files).

## What I verified

**Identifiers and contracts resolve.**

- `worktree.create(repoPath, baseRef, name)` returns `{ worktreePath: <parent>/.shepherd-worktrees/<repoBase>-<name>, branch: shepherd/<name>, isolated }` — `src/worktree.ts:78-103` (branch at `:89`, path at `:91`). So `FIX_BRANCH_PREFIX`/`FIX_BRANCH_FULL_PREFIX` (`src/maintain.ts:80-83`) and the `fixBranchOf` regex `/maintain-fix-([0-9a-f]{8})$/` over `basename()` (`src/maintain.ts:205-208`) do match what `create` hands back, and `randomUUID().slice(0,8)` is always 8 lowercase hex. `basename` is imported at `src/maintain.ts:28`.
- `MaintainDeps.worktree` was widened to include `"create"` (`src/maintain.ts:160-163`) — needed, since the old `Pick` had only `createDetached`.
- Tier-2 worktrees are `…-review-<uuid>-<sha8>` (`src/maintain.ts:878-881`), so `fixBranchOf` cannot misfire on a diagnosis row in `reapOrphans` (`src/maintain.ts:1133`); the test "does not go looking for a branch behind a tier-2 diagnosis worktree" in `test/maintain.test.ts` asserts exactly that.
- `EmptyDiffError`, `openPr(OpenPrInput): Promise<PrStatus>`, `listPullRequests(): Promise<PullRequest[]>`, `defaultBranch()`, `isLightweight` all exist as used — `src/forge/types.ts:294-315`, `:506`, `:468`, `:521`, `:465`. `listPullRequests` is open-only on GitHub (`src/forge/github.ts:1031-1033` → `listOpenPrSnapshot()`), so `prStillOpen` (`src/maintain.ts:526-535`) is a correct "still open" test.
- `MAX_DIAGNOSES_PER_SWEEP` → `MAX_ACTIONS_PER_SWEEP` has no stale consumers: grep across `src/ test/ ui/ scripts/ docs-site/` returns only `src/maintain-core.ts:60` and `src/maintain.ts:56,435`.
- `maintain_runs.outcome` is an unconstrained `TEXT` column (`src/store.ts:1430-1434`), so the new `"opened"` value persists without a migration, and `finishMaintainRun` (`src/store.ts:4316-4325`) accepts the optional `issueNumber`/`issueUrl`.
- `ensureBaseRef` is fail-closed and never throws (`src/worktree.ts:299-317`), so the un-`try`'d call at `src/maintain.ts:612` cannot abort the sweep loop.
- `completeReviewerSpawn("")` on a Tier-3 row is harmless — the row is inserted with `spawnSessionId: ""` (`src/maintain.ts:638-639`) and no `reviewer_spawns` row exists to match it.

**Locale parity holds.** All five new lens keys exist in both catalogs with identical `{n}` placeholders — `usage_maintain_band_dead_code` … `usage_maintain_pr_observe` at `ui/messages/en.json:3757-3761` and `ui/messages/de.json:3757-3761`. The announcement pair is in both (`ui/messages/en.json:3765-3766`, `ui/messages/de.json:3765-3766`) and matches the entry's `titleKey`/`bodyKey` (`ui/src/lib/feature-announcements/entries/v1.48.0-maintain-tier3.ts:5-6`). The glossary strings were updated in lockstep in `ui/messages/{en,de}.json` and `docs-site/src/content/docs/reference/glossary.md:85-100`, which is what `scripts/check-glossary.mjs` compares.

**The announcement entry is wired.** `ui/src/lib/feature-announcements.ts:47-50` eager-globs `./feature-announcements/entries/*.ts`, so the new file needs no registration; filename `v1.48.0-maintain-tier3.ts` agrees with `sinceVersion: "1.48.0"` and the id, per the contract at `ui/src/lib/feature-announcements.ts:1-23`, and matches the already-shipped `v1.48.0-maintain-loop.ts`.

**Type mirroring.** `BandId` (`src/types.ts:1058`, `ui/src/lib/types.ts:1515`), `MaintainTier`, `MaintainOutcome`, `MaintainRun` and `MaintainBlock.pr` are mirrored across both files; the server payload sets `pr` at `src/server.ts:4390` from `src/config.ts:751`, and the exact-shape assertion in `test/server-usage-delivery.test.ts:125-131` was updated.

**Version pin.** `FALLOW_VERSION = "2.100.0"` (`src/maintain-core.ts:69`) matches every other pin site: `.github/workflows/ci.yml:170`, `scripts/pre-push.ts:679`, `CONTRIBUTING.md:86,98,147`.

**Install set.** `FIX_INSTALL_DIRS = [".", "ui", "extension"]` (`src/maintain.ts:58`) matches CI's three installs (`.github/workflows/ci.yml:65-74`), and each of those dirs has its own lockfile (`bun.lock`, `ui/bun.lock`, `extension/bun.lock`).

**Generated artifacts do not leak into the commit.** `bun install` runs the packages' `prepare` scripts (`ui/package.json:13` → `svelte-kit sync` + paraglide compile; `extension/package.json:16` → paraglide compile), but their outputs are ignored — `ui/.gitignore` (`/.svelte-kit`, `/src/lib/paraglide`), `extension/.gitignore` (`src/lib/paraglide/`), root `.gitignore:1` (`node_modules/`). The root `prepare` (`package.json:17`) runs `scripts/register-merge-driver.mjs`, documented as idempotent and worktree-safe (`scripts/register-merge-driver.mjs:8-9`). So `git status --porcelain` in the fix worktree reflects `fallow fix` alone, which is what the manifest guard assumes.

**Porcelain parsing.** `changedPaths` (`src/maintain.ts:822-836`) slices column 3 (correct for `XY<space>path`, including `?? path`), takes the rename destination after the last ` -> `, and strips surrounding quotes before `MANIFEST_RE` (`src/maintain.ts:63`) is applied — and that regex does match `ui/package.json` / `ui/bun.lock` via its `(^|\/)` anchor.

**Could not verify (external ground truth, not in this repo):** the exact `fallow dead-code --format json` schema — the `kind: "dead-code"` discriminator, the top-level per-category arrays, and `actions[].auto_fixable`. `parseDeadCodeReport` (`src/maintain-core.ts:248`) is only exercised against the repo's own fixtures (`test/maintain-core.test.ts:465-499`). Assumed as written. The same applies to the claim that `fallow dead-code` exits 1 whenever it finds anything, which `reportFromFailedExit` (`src/maintain.ts:103-107`) depends on.

## Findings

### 1. The fail-closed verify gate typechecks only the server package — `fallow fix` edits four packages it does not cover

`verifyFix` runs `this.fixRunner.typecheck(worktreePath)` (`src/maintain.ts:749-754`), and the default runner is `bun run typecheck` in the worktree root (`src/maintain.ts:179-183`). That script is `tsc --noEmit` (`package.json:12`) against a tsconfig that explicitly excludes the other packages: `"exclude": ["node_modules", "ui", "extension", "site", "docs-site"]` (`tsconfig.json:17`), with a comment at `tsconfig.json:12-16` stating each of those has *its own* check command (`svelte-check` for `ui` — `ui/package.json:14`; `svelte-check` for the extension — `extension/package.json:17`).

`fallow` does analyse those trees: the `entry` list in `.fallowrc.jsonc` covers `ui/src/routes/**` (`:30-32`), `ui/src/hooks.*`, `extension/src/background.ts` (`:44`), `extension/src/popup|options/main.ts`, and `docs-site/astro.config.mjs` (`:59`). So `fallow fix` can and will apply `remove-export`-class fixes under `ui/src/lib/**`, `extension/src/**` and the two Astro sites — and for every one of those the gate is a no-op: `tsc` compiles nothing that references them, so it passes vacuously and the run proceeds to commit → push → `openPr` (`src/maintain.ts:785-812`).

This is the one guard standing between an automated `fallow fix` and a pushed branch, and it is missing on exactly the sibling packages where static-analysis false positives are most likely (Svelte/`.svelte`-template-only usage is the classic case — the repo already carves out related false-positive classes at `.fallowrc.jsonc` under `rules`). The re-measure check that follows (`src/maintain.ts:755-764`) only asserts that fallow agrees with itself; it cannot catch fallow being wrong.

It also makes two statements added by this diff misleading in practice, even though each is literally true of what ran: the PR body's "Verified before opening → `bun run typecheck` passes on the result" (`src/maintain-core.ts:691`) and the docs' "**verifies fail-closed** — `bun run typecheck` must pass" (`docs-site/src/content/docs/reference/configuration.md:292`). A reviewer reading either will assume the removal was typechecked; for a `ui/`, `extension/`, `site/` or `docs-site/` removal it was not.

Suggested fix: extend the `typecheck` seam to also run the per-package checks (at minimum `cd ui && bun run check` and `cd extension && bun run check`), or route on which packages `changedPaths` actually touched and run the corresponding check for each — keeping it fail-closed either way. `changedPaths` is already computed in `prepareFix` (`src/maintain.ts:732`), so the routing information is in hand.

## Nits (non-blocking):

- `src/types.ts:1058-1059` and `ui/src/lib/types.ts:1515-1516`: the `BandId` union is written as a line break after `=` with all four members on the next line; prettier's union-break style is one leading-pipe member per line, and the single-line form is ~105 chars (over the 100 printWidth), so `prettier --check` will likely rewrite both.
- `src/maintain-core.ts:705-711`: the PR body's footer quotes `describeReading(reading)` — the **live-checkout** reading taken during the sweep — while "What this removes" above it quotes `before`, the **pristine-worktree** measurement. The two counts can legitimately differ within the same body.
- `src/maintain.ts:780-783` vs `:757`: observe mode and "nothing auto-fixable on the base branch" both record outcome `skipped`, so the run log cannot distinguish "did the work, wasn't armed" from "there was nothing to do" without reading the log message.
- `src/server.ts:4400-4407`: `POST /api/maintain/sweep` now awaits a `bunx fallow` measurement and, on a tier-3 breach, a full install + fix + typecheck — the request can hang for many minutes where it previously returned shortly after a spawn.

- `src/maintain.ts:810-812`: if `openPr` *resolves* with no `number` or `url`, the run records `error` but the pushed remote branch is not taken down (only the `catch` path at `:799-808` deletes it) and the PR that may in fact exist is never recorded, so the next run after the cooldown would open a duplicate. Unreachable with today's GitHub/Gitea adapters, which populate both on a successful open.

- `ui/src/lib/feature-announcements/entries/v1.48.0-maintain-tier3.ts:8` sets `targetId: "usage-delivery-tab"`, which is not one of `PILL_FEATURE_IDS` (`ui/src/lib/components/GitRail.svelte:514`). The rule as stated says to omit `targetId` outside that set — but `ui/src/lib/components/Usage.svelte:275` does register that coachTarget with `use:coachTarget`, and three existing entries (`v1.48.0-delivery-metrics.ts`, `v1.48.0-maintain-loop.ts`, `v1.48.0-plan-drift.ts`) use the same id, so the anchor is live and this follows repo practice. Noted only because the curated rule reads narrower than what the codebase does.

- None observed. `reportFromFailedExit` and the `FixRunner` seam are both exported for the tests the plan called for, and `tier3` was deliberately kept off `RateBandConfig`/`IncidentBandConfig` as #2171 required (`src/maintain-core.ts:101-113`, `:143`).

## Assessment against the task

The diff does implement #2171's shape: a per-band `tier3: { class }` absent by default (`src/maintain-core.ts:101-113`, `:143`), a third rung in `MaintainTier`/`evaluateBands` via promotion rather than a fourth threshold (`src/maintain-core.ts:342`, `:434-451`), the doc-agent publish tail instead of `createIssue` (`src/maintain.ts:785-818`), no auto-merge on any path, its own `SHEPHERD_MAINTAIN_PR` flag independent of `SHEPHERD_MAINTAIN_ACT` (`src/config.ts:747-751`), and `inflightWorktrees()` / `reapOrphans` extended to cover the fix run (`src/maintain.ts:375-383`, `:1112-1140`). The one thing that keeps it from being complete is the verification gate above: the tier's whole licence to skip the issue is that the result is mechanically verified before it is pushed, and today that verification stops at the server package.

scoop (commented):
# Review — Maintain loop Tier 3: open a PR for a pre-approved fix class (#2171)

Range reviewed: `4d92a773528c05ffffb86c55f33b6ed0ea5447d7...HEAD` (2 commits: `5bec4f271` feature,
`462fd0beb` rework).

## Prior finding from round 1

**1. The Tier-3 verify gate ran only `bun run typecheck`, vacuous for a fix under `ui/`,
`extension/`, `site/` or `docs-site/`. — ADDRESSED.**

Verified:

- `FixRunner.typecheck` now takes a package (`src/maintain.ts:87-90`), and the default runner routes
  it: `const [dir, script] = pkg === "root" ? [".", "typecheck"] : [pkg, "check"]`, executed with
  `cwd: join(worktreePath, dir)` (`src/maintain.ts:183-189`). Both scripts exist —
  `package.json:12` (`"typecheck": "tsc --noEmit"`), `ui/package.json:14` and
  `extension/package.json:17` (`"check": "paraglide-js compile … && svelte-check …"`).
- `verifyFix` now consumes the `changed` list it already computed (`src/maintain.ts:703`,
  `769-790`) and loops the packages `packagesFor` returns, aborting `error` on the first red one
  with the package named in `why` (`src/maintain.ts:784-787`).
- `packagesFor` (`src/maintain-core.ts:318-337`) maps `ui/…` → `ui`, `extension/…` → `extension`,
  everything else → `root`, and returns `site/`/`docs-site/` paths as `unverifiable`; `verifyFix`
  refuses to commit when that list is non-empty (`src/maintain.ts:773-779`). That is the right call
  for those two trees: `FIX_INSTALL_DIRS` installs only `.`, `ui`, `extension`
  (`src/maintain.ts:61`), and neither `site` nor `docs-site` has a check wired into the run.
  Prefix matching is anchored, so `docs-site/x` does not collide with the `site/` prefix.
- Root coverage is real for the "else" bucket: `tsconfig.json` has no `include` and excludes only
  `node_modules, ui, extension, site, docs-site` (`tsconfig.json:17`), so `src/`, `scripts/` and
  `test/` are all inside the root `tsc` run.
- The advertising copy was corrected in both places the previous finding named: the PR body now
  prints the packages actually checked (`src/maintain-core.ts:741-742`, `renderFixPrBody` takes
  `packages: FixPackage[]`), and the docs describe per-package verification plus the
  `site/`/`docs-site/` refusal (`docs-site/src/content/docs/reference/configuration.md:292-301`).

No prior finding was dropped for being out of diff.

## What I verified in this round

- **Store contract.** `maintain_runs` has no CHECK constraint on `outcome` (`src/store.ts:1430-1434`),
  so the new `"opened"` value writes cleanly through `insertMaintainRun` / `finishMaintainRun`
  (`src/store.ts:4292-4324`).
- **Forge contract.** `openPr(OpenPrInput) → PrStatus` with optional `number`/`url`
  (`src/forge/types.ts:294-300`, `199-209`, `506`) — `publishFix` guards both before recording
  (`src/maintain.ts:679-681`). `EmptyDiffError` is a real export (`src/forge/types.ts:305`) and is
  caught after the push, with a best-effort remote-branch delete first (`src/maintain.ts:668-677`).
  `listPullRequests()` (`src/forge/types.ts:468`) backs the new `prStillOpen`.
- **Worktree contract.** `worktree.create(repo, base, name)` returns
  `{ worktreePath: <parent>/<repoBasename>-<name>, branch: "shepherd/<name>", isolated }`
  (`src/worktree.ts:78-99`), which is exactly what `fixBranchOf`'s
  `/maintain-fix-([0-9a-f]{8})$/` on `basename(worktreePath)` reconstructs (`src/maintain.ts:212-215`)
  — and `randomUUID().slice(0,8)` is hex, so the reap path resolves the same branch it created.
  `MaintainDeps.worktree` was widened to include `"create"` (`src/maintain.ts:167-170`).
  `worktree.remove` uses `git worktree remove --force` plus an `rmSync` fallback
  (`src/worktree.ts:559-568`), so the observe-mode teardown of a dirty fix worktree cannot leak.
- **Generated output cannot leak into the commit.** `verifyFix` runs `ui`/`extension` `check`,
  which write `src/lib/paraglide` and `.svelte-kit` — both ignored (`ui/.gitignore:8,26`,
  `extension/.gitignore:4`, `.gitignore:1-2`), so the later `git add -A`
  (`src/maintain.ts:654`) cannot stage them.
- **Version pin.** `FALLOW_VERSION = "2.100.0"` (`src/maintain-core.ts:69`) matches every other pin
  site the comment names: `.github/workflows/ci.yml:170`, `scripts/pre-push.ts:679`,
  `CONTRIBUTING.md:86,98,147`.
- **Locale parity.** All seven new keys exist in both catalogs at the same positions:
  `usage_maintain_band_dead_code`, `usage_maintain_sample_findings`, `usage_maintain_state_fix`,
  `usage_maintain_pr_armed`, `usage_maintain_pr_observe`, `feat_maintain_tier3_title`,
  `feat_maintain_tier3_body` — `ui/messages/en.json:3757-3761` and `ui/messages/de.json:3757-3761`,
  plus the reworded `gloss_*` pair in both. The placeholder shape matches the existing sibling
  (`ui/messages/en.json:3753` `"usage_maintain_sample": "n={n}"` vs `:3758`
  `"usage_maintain_sample_findings": "{n} found"`), and `bandSample` selects between them
  (`ui/src/lib/components/usage/DeliveryLens.svelte:153-160`).
- **Announcement entry.** `ui/src/lib/feature-announcements/entries/v1.48.0-maintain-tier3.ts` is
  picked up by the eager `import.meta.glob` registry (`ui/src/lib/feature-announcements.ts:47-50`) —
  no index to update — and its `titleKey`/`bodyKey` resolve to the catalog keys above.
- **Mirror.** `BandId`, `MaintainTier`, `MaintainOutcome`, `MaintainRun` and `MaintainBlock.pr` are
  mirrored in `ui/src/lib/types.ts:1515-1563`; `src/server.ts:4390` populates `pr` and
  `test/server-usage-delivery.test.ts:125-131` asserts the exact block shape.
- **No stale "there is no tier 3" left in the docs.** A tree-wide grep for `tier 3` outside the
  updated `configuration.md` returns only the newly-written `glossary.md:97`; `operating.md` never
  made the claim (`docs-site/src/content/docs/operating.md:176` is its only maintain mention).
- **Removed export.** `MAX_DIAGNOSES_PER_SWEEP` → `MAX_ACTIONS_PER_SWEEP` has no other importer in
  `src/`, `test/` or `ui/src`.
- **Sweep cost is gated.** `gather()` became async and now shells out to fallow in the live checkout
  (`src/maintain.ts:362-364`), but `sweep()` returns on the once-a-day claim before that
  (`src/maintain.ts:405`), and the tick that calls it is behind `config.maintainLoopEnabled`
  (`src/index.ts:1884-1888`).

## Findings

**1. `src/types.ts:1085-1087` documents the count band's `sampleN` incorrectly.**

The new sentence reads: *"Count bands (`dead_code_drift`) have no separate denominator and report
`sampleN === value`."* The code does the opposite of the second half — `evaluateBands` sets
`value: dc?.autoFixable ?? 0` and `sampleN: dc?.total ?? 0`
(`src/maintain-core.ts:487,490`), which differ whenever fallow reports a non-auto-fixable finding
(unused files — the documented normal case). The in-file comment two lines above the code says so
explicitly (`src/maintain-core.ts:488-489`, "Total findings, not a denominator"), and the UI relies
on the difference: it renders `value` in the value column and `sampleN` as a separate "N found"
(`ui/src/lib/components/usage/DeliveryLens.svelte:153-160`), with a test asserting `2` and `5` for
the same reading (`ui/src/lib/components/usage/DeliveryLens.browser.test.ts:330-336`). This is the
server-side half of a contract that is deliberately mirrored into `ui/src/lib/types.ts`, so a
consumer reading it would build the wrong renderer. One-line fix.

- `thresholdNote` (`src/maintain.ts:325-333`) has no `dead_code_drift` branch: such a reading falls into `const cfg = … : t.first_pass_collapse` and the diagnosis prompt would quote the *first-pass* percentages ("tier 1 at or below 60% …") for a dead-code band. Dormant today — `evaluateBands` always promotes that band's tier-2 breach to 3 (`src/maintain-core.ts:491`) and `mergeCountBand` always carries `tier3` through (`src/maintain-core.ts:163-172`) — but its sibling helper `windowDaysFor` did get an explicit branch plus a comment for exactly this case (`src/maintain-core.ts:401`), so the guard is present on one path and missing on the other.

## Nits (non-blocking):

- `src/maintain-core.ts:411` still says "Evaluate all three bands"; this PR makes it four.
- `src/maintain.ts:1-24` file docstring still describes the service as diagnosis-only ("spawns a read-only diagnosis agent …"), with no mention of the tier-3 path added below it.
- The band row renders an opened PR as a bare `#512` link (`ui/src/lib/components/usage/DeliveryLens.svelte:306-311`); the plan's "PR #n" would distinguish it from a `filed` issue at a glance.

- `docs-site/src/content/docs/reference/glossary.md` and the `gloss_*` catalog entries were rewritten; the plan's file list named `operating.md` instead. Small, coherent, and arguably the better target — noted only because it is unplanned surface.

- The new announcement entry sets `targetId: "usage-delivery-tab"` (`ui/src/lib/feature-announcements/entries/v1.48.0-maintain-tier3.ts:8`) while the curated rule says to set `targetId` only for `PILL_FEATURE_IDS` (`ui/src/lib/components/GitRail.svelte:514`). The rule reads stale for this anchor: a real `use:coachTarget={"usage-delivery-tab"}` exists (`ui/src/lib/components/Usage.svelte:275`) and all three sibling v1.48.0 entries do the same, so the intent is met by other means.

## Stated limitations

- I could not verify `fallow@2.100.0 dead-code --format json`'s real payload shape — fallow is an
  external `bunx` tool, not vendored in this repo. `parseDeadCodeReport`
  (`src/maintain-core.ts:248-282`) is judged only against the fixture the tests construct
  (`test/maintain.test.ts:85-101`) and against fallow's documented `auto_fixable` action flag as
  described in the plan; its `kind === "dead-code"` discriminator and the
  "top-level array of objects carrying `actions`" walk are assumed as written, not confirmed.
- `listPullRequests()` is documented as "capped server-side" (`src/forge/types.ts:467-468`). Whether
  a still-open tier-3 PR can fall outside that cap depends on host behaviour I cannot exercise here;
  either way `prStillOpen` fails open to the plain 14-day cooldown
  (`src/maintain.ts:405-414`), which is the same posture as the existing `issueStillOpen`.
- CI-run behaviour (lint, tests, `svelte-check`) was not executed, per the review instructions.

PR #2185, opened 2026-09-02
feat(gates): gate the fallow version pin across every site (#2182)

### LINKED ISSUE #2182: Gate the fallow version pin — 8 hand-synced copies across 4 files, one of them now drives an unattended loop
The fallow version is pinned (deliberately, not `@latest`) and now written out **8 times across 4 files**:

| File | Occurrences |
| --- | --- |
| `.github/workflows/ci.yml` | 2 (the `fallow audit` step + a comment) |
| `scripts/pre-push.ts` | 3 (sync banner, comment, the argv) |
| `CONTRIBUTING.md` | 3 |
| `src/maintain-core.ts` | 1 — `FALLOW_VERSION` (added by #2179) |

Every site carries a "keep in sync with the others" comment. Nothing enforces it.

## Why it matters more now

Before #2179 a drifted pin meant CI and the pre-push hook disagreeing about what counts as a finding — annoying, but both are gates a human is watching. #2179's `dead_code_drift` band uses `FALLOW_VERSION` to **measure** and to **run `fallow fix`**, unattended. If that constant drifts from the CI pin, the maintain loop measures against one analyser and opens a PR that CI then gates with another. A one-sided edit is silent in both directions:

- bump `ci.yml` only → the band keeps scoring on the old analyser; a Tier-3 PR can land CI-red on findings the band never saw.
- bump `FALLOW_VERSION` only → the loop opens PRs for a finding class CI does not enforce.

## There is already a precedent for exactly this

`scripts/check-model-mirror.mjs` exists because `CLAUDE_MODELS` / `CODEX_MODELS` / `EFFORTS` are declared twice and "a one-sided edit is silent in BOTH directions, asymmetrically" (#1936). Its header makes the same argument this issue makes. It runs as a bare command in the `gates` lane of `scripts/pre-push.ts` alongside `check-glossary.mjs` and `check-announcement-versions.mjs`.

## Suggested shape

A `scripts/check-fallow-pin.mjs` in the same mould: scan the four files for `fallow@<semver>` / `FALLOW_VERSION = "<semver>"` and fail when they are not all identical, naming the odd one out. Wire it into the `gates` lane and the `pr-hygiene` workflow next to the existing three. Textual, like the model mirror, for the same reason: two of the sites are a YAML workflow and a Markdown doc that no test can import.

Deliberately **not** proposed: a single exported constant every site reads. `ci.yml` and `CONTRIBUTING.md` cannot import TypeScript, so the duplication is structural — a gate is the only thing that closes it.

Small and self-contained; reasonable first issue.

scoop:
Closes #2182.

The fallow pin is deliberately not `@latest`, and nothing enforced that its copies agree. Since #2179 that stopped being cosmetic: the `dead_code_drift` band uses `FALLOW_VERSION` to measure and to run `fallow fix`, unattended. A one-sided edit is silent in both directions — bump `ci.yml` only and the band keeps scoring on the old analyser; bump `FALLOW_VERSION` only and the loop opens PRs for a class CI does not enforce.

### The issue undercounts the problem

It says 8 sites across 4 files. The tree has **12 across 6**. Also missed: `.fallowrc.jsonc` (a runnable `bunx fallow@2.100.0 dupes --save-baseline` recipe), a fourth `CONTRIBUTING.md` mention, and a `test/maintain.test.ts` provenance note.

### Two halves

**1. Normalize the prose.** Five of the twelve sites named the version in prose ("pins fallow to 2.100.0", "Verified against fallow 2.100.0"). No safe pattern can match those, because the *same comment blocks* legitimately name other releases — `(fallow 2.98+)`, the release that introduced the synthetic Svelte `<template>` metric, and "the old 2.97 pin". A loose "fallow near a semver" regex flags both as drift. So all five were reworded to carry the `fallow@<semver>` token, and the legitimate prose stays unmatched **by construction** (no `@`) rather than by an exclusion list. Coverage 7/12 → 12/12.

**2. `scripts/check-fallow-pin.mjs`**, in the mould of `check-model-mirror.mjs` (#1936) — textual for the same reason: two of the sites are a YAML workflow and a Markdown doc that no test can import, and all three wiring points invoke a gate as a bare command.

`FALLOW_VERSION` in `src/maintain-core.ts` is the canonical (it drives the unattended loop); every `fallow@<semver>` in a tracked file must equal it. Fails closed on an unparseable canonical **and** on finding zero invocations — a rotted pattern must be an error, never the vacuous pass a gate exists to prevent.

### One deliberate departure from the issue's shape

It proposes a fixed 4-file list, mirroring `check-model-mirror`. This scans **repo-wide** over `git ls-files` instead: a `bunx fallow@…` invocation can land in any new file, unlike a `const` in a known module, so a fixed list reproduces the same silent-drift hole one level up — a pin in a seventh file would be invisible. The scan costs a few milliseconds behind a `fallow` substring prefilter.

Also updated: four sentences instructing humans to hand-sync the pin now point at the gate instead.

### Notable in review

- **The test file is scanned by the gate it tests.** A literal `fallow@2.99.0` fixture would fail the gate on its own test file. Every fixture pin is built by interpolation, so the source holds only `fallow@${…}` — no digit after the `@`. That keeps the repo-wide scan free of exclusion lists; a self-exemption would be a cheap escape hatch in the one file most needing coverage.
- **Ambient `GIT_DIR`.** The fixture CLI tests build throwaway git repos. Root tests honour an ambient `GIT_DIR`, so every git call — in the gate and the tests — runs with git's repo-local vars scrubbed (the same set `.husky/pre-push` unsets). Without it, `git init`/`git add` would operate on the real repository.
- **Prerelease tails are captured.** A bare `\d+\.\d+\.\d+` would read `2.100.0-beta.1` as `2.100.0` and compare it equal to a plain `2.100.0` elsewhere.

### Verification

- Gate reports 11 pinned invocations + 1 `FALLOW_VERSION` declaration across 6 files — the full inventory.
- Drifting **each** of the six files individually exits 1 and names the `file:line`; bumping `FALLOW_VERSION` alone flags all 11 others.
- Removing the canonical, and separately every invocation, each exit 1.
- `bun run test` 9062 pass / 0 fail · `bun run lint` · `bun run typecheck` · prettier clean.
- `bunx fallow@2.100.0 audit --base origin/main --fail-on-issues` clean; pre-push green with the new gate in the `gates` lane.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,8 +25,9 @@ jobs:
     # The local pre-push hook no longer shares this job's shell body — it runs the
     # same gate set in parallel lanes via scripts/pre-push.ts. When you add/remove/
     # reorder a step here, mirror it there (and vice-versa). Deliberate differences:
-    # the hook scopes prettier/eslint to the push delta and pins fallow to 2.100.0;
-    # CI keeps the whole-repo checks. Keep the fallow version in sync across both.
+    # the hook scopes prettier/eslint to the push delta and pins fallow@2.100.0;
+    # CI keeps the whole-repo checks. The pin is gated: scripts/check-fallow-pin.mjs
+    # fails when any site names a different version.
     # Don't run the full suite on release-please's generated PR: it only bumps the
     # version + CHANGELOG, so verify is wasted minutes. That PR DOES trigger this
     # workflow — see release-please.yml. Skip ONLY when BOTH the head branch carries
@@ -164,8 +165,8 @@ jobs:
         # (#851) — a Svelte-appropriate bar so new components aren't churned against the
         # library-default 20/15. The #756 line-shift mis-attribution that forced the old
         # 2.97 pin was re-tested against 2.100 and does NOT reproduce (inherited template
-        # findings now attribute correctly across line shifts). Keep this version in sync
-        # with .husky/pre-push and CONTRIBUTING.md.
+        # findings now attribute correctly across line shifts). Every site naming this
+        # version is gated by scripts/check-fallow-pin.mjs — bump them together.
         if: github.event_name == 'pull_request'
         run: bunx fallow@2.100.0 audit --base origin/${{ github.base_ref }} --fail-on-issues
 
--- a/.github/workflows/pr-hygiene.yml
+++ b/.github/workflows/pr-hygiene.yml
@@ -57,3 +57,5 @@ jobs:
         run: node scripts/check-announcement-versions.mjs
       - name: Model-list mirror (src/types.ts ↔ ui/src/lib/types.ts)
         run: node scripts/check-model-mirror.mjs
+      - name: Fallow version pin (identical at every site)
+        run: node scripts/check-fallow-pin.mjs
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -76,7 +76,7 @@ and failed pushes. Independent work runs in parallel (bounded by your core count
 lane teeing to its own `.test-logs/` file, with a per-lane wall-clock **timeout backstop**
 so a hung child can never wedge the push. The checks (per lane) are:
 
-- **gates:** branch-hygiene · feature-catalog · generated-docs · glossary · announcement-versions · model-mirror · herdr-types
+- **gates:** branch-hygiene · feature-catalog · generated-docs · glossary · announcement-versions · model-mirror · fallow-pin · herdr-types
 - **prettier:** `prettier --check` over the push **delta** (see note)
 - **eslint:** root + extension eslint over the push **delta** (see note)
 - **tsc:** `bun run typecheck` (root `tsc --noEmit`)
@@ -95,8 +95,8 @@ so a hung child can never wedge the push. The checks (per lane) are:
 >   offline? The hook falls back to whole-repo lint and skips fallow.)
 > - **Same checks, different scoping.** Because the hook no longer shares `ci.yml`'s shell
 >   body, the two gate definitions can drift — when you change a CI step, mirror it in
->   `scripts/pre-push.ts` (there's a sync banner in both files). Keep the `fallow@2.100.0`
->   pin in sync across the hook, `ci.yml`, and this file.
+>   `scripts/pre-push.ts` (there's a sync banner in both files). The `fallow@2.100.0` pin is
+>   gated across every site by `bun run check:fallow-pin` — bump them together.
 >
 > **Concurrency is core-aware.** Lane fan-out and per-tool worker counts are bounded so
 > `laneCap × workers ≤ cores` — no oversubscription on a modest box. Override the lane cap
@@ -114,7 +114,7 @@ so a hung child can never wedge the push. The checks (per lane) are:
 > type errors. Bun runs `.ts` by stripping types, so it never type-checks — only
 > `tsc` does.
 
-> **fallow is pinned to `2.100.0`** (not `@latest`) so analyzer changes are adopted
+> **the fallow pin is `fallow@2.100.0`** (not `@latest`) so analyzer changes are adopted
 > deliberately, not on a random run. The synthetic Svelte `<template>` complexity metric
 > (fallow 2.98+) is adopted via `health.thresholdOverrides` in `.fallowrc.jsonc`
 > ([#851](https://github.com/erwins-enkel/shepherd/issues/851)) — a Tier-1 global bar
@@ -129,8 +129,8 @@ so a hung child can never wedge the push. The checks (per lane) are:
 > reproduce** — fallow attributes inherited template findings correctly across line
 > shifts and complexity changes; the audit's new-only gate fires only when a template is
 > a finding in HEAD that was not one in the base. A genuinely new oversized template will
-> still trip as _introduced_ (the gate working) — fix or grandfather it. Keep the version
-> in sync with `.github/workflows/ci.yml` and `.husky/pre-push`.
+> still trip as _introduced_ (the gate working) — fix or grandfather it. Every site naming
+> the pin is gated by `bun run check:fallow-pin`, so a one-sided bump fails the build.
 
 Run any of these manually at any time:
 
@@ -140,6 +140,7 @@ bun run typecheck            # root tsc --noEmit (src + test)
 bun run format               # prettier --write across the repo
 bun test ./test              # core test suite
 bun run check:model-mirror   # model/effort lists identical in src/types.ts ↔ ui/src/lib/types.ts
+bun run check:fallow-pin     # the pinned fallow version identical at every site
 cd ui && bun run check       # svelte-check (ui types)
 cd ui && bun run check:i18n  # locale-catalog parity (en ↔ de)
 cd ui && bun run test        # ui test suite (vitest)
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "check:glossary": "node scripts/check-glossary.mjs",
     "check:announcement-versions": "node scripts/check-announcement-versions.mjs",
     "check:model-mirror": "node scripts/check-model-mirror.mjs",
+    "check:fallow-pin": "node scripts/check-fallow-pin.mjs",
     "next-version": "node scripts/next-version.mjs",
     "gen:herdr-schema": "bun run scripts/gen-herdr-schema.ts",
     "gen:herdr-types": "bun run scripts/gen-herdr-types.ts",
--- a/scripts/check-fallow-pin.d.mts
+++ b/scripts/check-fallow-pin.d.mts
@@ -1,0 +1,54 @@
+// Types for the node-run fallow-pin gate (scripts/check-fallow-pin.mjs). The gate stays
+// plain .mjs so `node scripts/check-fallow-pin.mjs` works as a CLI in any context (the
+// package script, the pre-push gates lane, PR hygiene); this declaration only exists so
+// its TypeScript importer — test/check-fallow-pin.test.ts — gets real types instead of
+// `any`. Mirrors scripts/check-model-mirror.d.mts's role.
+
+/** How a site names the pin: a `fallow@<semver>` invocation, or the TS constant. */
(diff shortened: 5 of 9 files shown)


scoop (commented):
# Review — gate the fallow version pin (#2182)

**Decision: comment.** No blocking defects found. The gate is correct, fails closed in both
directions, and the prose normalization it depends on is complete on the live tree.

### The gate finds every site, and only real sites

`git grep -nE 'fallow@[0-9]'` over the tree returns exactly 11 hits, all `2.100.0`:

- `.fallowrc.jsonc:432`
- `.github/workflows/ci.yml:28`, `:171`
- `CONTRIBUTING.md:86`, `:98`, `:117`, `:148`
- `scripts/pre-push.ts:16`, `:676`, `:685`
- `test/maintain.test.ts:1143`

plus the canonical declaration at `src/maintain-core.ts:69` (`export const FALLOW_VERSION = "2.100.0";`).
That is the 12-site inventory the plan claims, and it matches the per-file floors asserted in
`test/check-fallow-pin.test.ts:293-306` one for one (ci.yml 2, CONTRIBUTING 4, pre-push 3,
.fallowrc 1, maintain-core 1, maintain.test 1). So the live-tree assertion in
`test/check-fallow-pin.test.ts:307` will pass and the CLI will print "11 pinned invocation(s) + 1
FALLOW_VERSION declaration(s) across 6 file(s)".

The three "must stay unmatched" claims in the header comment of `scripts/check-fallow-pin.mjs:70-73`
are factually correct, each confirmed:

- `scripts/pre-push.ts:680` — `` `bunx fallow@…` `` (ellipsis, no digit)
- `src/maintain.ts:102` — `` `bunx fallow@…` ``; `src/maintain.ts:163`, `:178` and
  `src/maintain-core.ts:734` interpolate `` `fallow@${FALLOW_VERSION}` `` (a `$` after the `@`)
- `docs-site/src/content/docs/reference/configuration.md:291` — `bunx fallow@<pinned> fix …`

None of those can match `fallow@(\d+\.\d+\.\d+…)`. Likewise the legitimate other-release prose stays
unmatched by construction (no `@`): `CONTRIBUTING.md:119` and `.github/workflows/ci.yml:164`
(`fallow 2.98+`), `.fallowrc.jsonc:205`, `CONTRIBUTING.md:127` / `ci.yml:167` ("the old 2.97 pin"),
and `CHANGELOG.md:1015` ("pin fallow to 2.97.0").

### The gate cannot fail on itself

`scripts/check-fallow-pin.mjs` writes its only concrete example as `` `export const FALLOW_VERSION =
"<semver>";` `` (line 247) — `<` after the quote, so `declarationRe` does not match. `test/check-fallow-pin.test.ts:38-39`
builds every fixture token by interpolation (`` `fallow@${v}` ``, `` `… = "${v}";` ``), leaving no
digit after `@` or after the opening quote in the file's own source. I checked both files against the
two regexes at `scripts/check-fallow-pin.mjs:65` and `:88-93`; neither self-matches. The
no-exclusion-list design holds.

### Fail-closed behaviour

`comparePins` (`scripts/check-fallow-pin.mjs:132-152`) returns `no-canonical` when the canonical is
unparseable and `no-invocations` when zero `fallow@<semver>` hits are found, before ever computing
mismatches — so neither an unparseable constant nor a rotted pattern can produce a vacuous pass. The
CLI (`:274-289`) exits 1 on a thrown `git ls-files` failure too. `extractCanonicalVersion`
(`:119-122`) returns null for an import statement or a non-literal value, which the tests pin at
`test/check-fallow-pin.test.ts:96-111`.

### Regex correctness

`SEMVER` (`scripts/check-fallow-pin.mjs:64`) carries the prerelease and build tails, so a
`2.100.0-beta.1` site cannot compare equal to a bare `2.100.0` — the truncation trap the comment
names is genuinely closed. `declarationRe` (`:88-93`) is anchored on `const` and permits an optional
type annotation (`(?::[^=]*)?`), which correctly matches `src/maintain-core.ts:69` while rejecting
the `import { FALLOW_VERSION }` at `src/maintain.ts:53` and the prose in the doc comment at
`src/maintain-core.ts:65-67`. Both regexes are constructed fresh per call, so there is no leaked
`lastIndex` between `extractPinRefs` and `extractCanonicalVersion`.

### The scan

`scanTree` (`:198-211`) enumerates via `git ls-files -z` with a scrubbed env, tolerates a missing
file (`continue` on readFileSync throw — covers staged deletions, broken symlinks, and the EISDIR a
submodule gitlink would produce), and skips a file with a NUL in its first 8000 bytes.
`Buffer.prototype.includes(0)` searches for the byte `0x00` and `includes("fallow")` decodes as utf8
by default, so both guards do what the comment says. `gitEnv()` (`:174-183`) asks git for
`--local-env-vars` (an option git handles before repository setup, so it works outside a repo) and
falls back to a hardcoded list — the same scrubbing `.husky/pre-push` performs, and the reason the
fixture tests in `test/check-fallow-pin.test.ts:195-206` cannot scribble on the real repository.

`git ls-files` reads the index, so a file staged/committed for the push is covered and an untracked
one is not — correct for both wiring points (pre-push runs after commit; `pr-hygiene.yml:43-47`
checks out the head SHA).

### Wiring

- `package.json:21` — `"check:fallow-pin": "node scripts/check-fallow-pin.mjs"`, alongside the three
  existing gate scripts.
- `.github/workflows/pr-hygiene.yml:60-61` — placed after `Model-list mirror`, exactly matching the
  precedent at `:58-59`. `node` is available on `ubuntu-latest` (the three sibling steps at `:55`,
  `:57`, `:59` already rely on it) and the checkout at `:43-47` gives it a real index.
- `scripts/pre-push.ts:406-411` — a `gates`-lane step. `plannedLaneCount` (`scripts/pre-push.ts:348-358`)
  counts lanes, not steps, and is unchanged; `test/pre-push.test.ts` asserts only on
  `computeConcurrency`/`plannedLaneCount`/`runLanes` (lines 67-158) and never on the gates step list,
  so no test churn. Verified the file is not absent — it exists and I grepped it.
- `CONTRIBUTING.md:79` adds `fallow-pin` to the gates bullet between `model-mirror` and `herdr-types`,
  matching the step order in `scripts/pre-push.ts:400-412`; `CONTRIBUTING.md:143` adds the manual
  command with no version literal in it.
- `scripts/check-fallow-pin.d.mts` mirrors the impl signatures exactly (`CANONICAL_REL`,
  `extractPinRefs`, `extractCanonicalVersion`, `comparePins`, `gitEnv`, `scanTree`) — I compared each
  against `scripts/check-fallow-pin.mjs:54`, `:104`, `:119`, `:132`, `:174`, `:198`. `NodeJS.ProcessEnv`
  resolves under the root `tsconfig.json`'s `"types": ["bun-types"]`: the same namespace is already
  used in root-checked code at `test/process-reaper.test.ts:387` and `test/herdr-install.test.ts:342`.
  The sidecar follows `scripts/check-model-mirror.d.mts`'s stated role.

### The box-drawing banner

`scripts/pre-push.ts:16` — I counted the replacement line's display width: 4 (`  • `) + 14
(`fallow@2.100.0`) + 49 (` pin — same as CI, gated by check-fallow-pin.mjs.`) + 12 trailing spaces =
79, identical to the neighbouring `:13` and `:15`. The frame is intact.

### Non-defects I checked and cleared

- `isMainModule` (`scripts/check-fallow-pin.mjs:225-235`) is byte-for-byte the version at the tail of
  `scripts/check-model-mirror.mjs`, and the two subprocess regressions at
  `test/check-fallow-pin.test.ts:262-292` plant a *drifted* fixture so exit 1 is only producible by a
  CLI block that genuinely ran.
- `console.error` + `process.exit(1)` on a pipe: the payload is a few hundred bytes, well inside a
  pipe buffer's synchronous `uv_try_write`, and it is the identical pattern `check-model-mirror.mjs`
  already ships. Not a truncation risk.
- No shell is used for any spawn; every argument is an array element and every path comes from
  `git ls-files`. No injection surface.

- `scripts/check-fallow-pin.mjs:174-183` — `gitEnv()` is exported as a public seam mainly so the test
  can reuse it; the plan listed only the four pure extract/compare seams. Reasonable, and it is what
  keeps the fixture test from scribbling on the real repo, so I would keep it.

- `scripts/check-fallow-pin.mjs:198` — the repo-wide scan includes `CHANGELOG.md`, which release-please generates from commit subjects and never rewrites. Today's history is safe because the one historical mention is prose (`CHANGELOG.md:1015`, "pin fallow to 2.97.0", no `@`), but a future commit subject written as `fallow@<version>` would be baked in permanently and fail the gate on the next bump with no hand-editable fix. Worth a line in the header telling bumpers to keep the `@` out of commit subjects.

## Nits (non-blocking):

- `CONTRIBUTING.md:117` — `> **the fallow pin is `fallow@2.100.0`**` starts lowercase, unlike every neighbouring bold callout lead (`:89`, `:101`, `:106`, `:126`), and reads redundantly ("the fallow pin is fallow@…"); `**The fallow pin is `fallow@2.100.0`**` would fix both.
- `scripts/check-fallow-pin.mjs:203` — `readFileSync` loads each tracked file in full before the cheap `fallow` substring prefilter, so large binaries and `CHANGELOG.md` are read whole on every run; the plan's own fallback (prefilter on a head slice) would avoid that if the gates lane ever feels it.

- None of the four standing rules (plan-phase guards, steering done/running sessions, once-on-idle firing, persistent-toast shape) touch any file in this diff.

- `src/maintain-core.ts:65-67` — `FALLOW_VERSION`'s doc comment still says "KEEP IN SYNC with the other pin sites … (they carry sync banners of their own)" and does not mention the new gate, unlike the four sites this PR reworded. Cosmetic only; the file is not in the diff.
- `.fallowrc.jsonc:432` — the `bunx fallow@2.100.0 dupes --save-baseline` recipe carries no note that the pin is now gated. Cosmetic only; the file is not in the diff.

## Stated limitations

- I did not run the test suite, the gate, `tsc`, or lint — CI owns those, per the review brief. Every claim above is from reading the tree, not from execution.
- I could not measure the scan's wall-clock cost (no execution), so the perf nit is reasoning about I/O volume, not a measurement.

PR #2186, opened 2026-09-02
fix(ui): stop the StatusPip close test racing a real stray cursor (#2180)

### LINKED ISSUE #2180: Flaky UI lane: StatusPip close-tooltip test races the 140ms grace timer under parallel load
Observed while running the full `ui/` suite repeatedly on #2179 (a server-side branch that touches no `StatusPip` code).

## Symptom

`ui/src/lib/components/StatusPip.browser.test.ts` → **"hover reveals a role=tooltip popover; leaving (unpinned) closes it"** fails roughly **2 in 5** full-suite runs:

```
❯ |browser (chromium)| src/lib/components/StatusPip.browser.test.ts (5 tests | 1 failed)
  × hover reveals a role=tooltip popover; leaving (unpinned) closes it
❯ src/lib/components/StatusPip.browser.test.ts:55:49
```

Only in the full parallel run (300 files). **5/5 green running the file in isolation.**

## Not a product bug — a timing budget

Line 55 is the closing half of the test:

```ts
leave();
await vi.waitFor(() => expect(tooltipOpen()).toBe(false));
```

`pointerleave` does not close the tooltip synchronously. `statusTip.svelte.ts` → `scheduleClose()` defers `hide()` behind a **140 ms** grace timer, so the pointer can cross the gap between trigger and panel:

```ts
closeTimer = setTimeout(() => { closeTimer = null; hide(); }, 140);
```

So the assertion is waiting on a real 140 ms timer plus popover teardown and the `:popover-open` style recalc, inside a browser tab shared with the rest of the parallel suite. `vi.waitFor`'s default budget (1000 ms, polled every 50 ms) is usually enough and occasionally is not — matching "only under load, never in isolation".

Note the *sibling* test on line 57 asserts the pinned case and is stable, because it waits on `toBe(true)` and then a fixed 30 ms — it never races the close timer.

## Suggested fix

Cheapest honest fix is to stop making the test wait out a wall-clock timer it does not care about — either give the close assertion an explicit budget comfortably above the grace period (e.g. `vi.waitFor(..., { timeout: 2000 })`), or let the action's grace period be injectable so the test can drive it to 0. Prefer whichever keeps the 140 ms affordance untested-by-accident rather than silently lengthened.

Please confirm the diagnosis before changing the action itself — the 140 ms grace is a deliberate UX affordance (documented at `statusTip.svelte.ts` → `scheduleClose`), not an accident, and should not be shortened to make a test pass.

## Related

Sibling of #2028 (EpicPanel leaks a rejection under parallel load) — same "only in the full run" class, different mechanism: #2028 exits 1 with **every test passing**, this one is a genuine assertion failure.

scoop:
Closes #2180.

## The diagnosis in the issue is wrong, and so was mine

Both said "timing budget": the close is deferred behind a real 140 ms grace timer, `vi.waitFor`'s default 1000 ms budget is only ~5x that, and under parallel load it runs out. I shipped that fix first — an explicit budget derived from the grace period — and **it did not work**: the flake reproduced with the larger budget in place, burning the *entire* 3140 ms while the file's other four tests took ~46 ms combined.

That asymmetry is the tell. A throttled timer would be *late*; this one never fires at all.

## What is actually happening

I instrumented the failing test to record close latency and every pointer event reaching the chip and the panel, flagging which were browser-generated. A stuck run under full parallel load:

```
elapsed=-1  rescue=panel-leave
pipEvs=[pointerleave@0  pointerover!@2  pointerenter!@2]
```

`!` marks `isTrusted`. Two milliseconds after the test's synthetic `pointerleave`, the browser dispatches a **real** `pointerover`/`pointerenter` to the chip. That runs `statusTip.svelte.ts` → `onPointerEnter()` → `show()` → `cancelClose()`, cancelling the pending close. Nothing is left to reopen the question: the cursor is stationary, so no further real `pointerleave` arrives, no outside pointerdown, no Esc. The tooltip stays open for 20 s in the instrumented run, and only a `pointerleave` dispatched on the *panel* rescued it.

Why a real cursor is on the chip at all: most specs in this suite drive real clicks through CDP, and that cursor position persists across files sharing a worker's page. In isolation nothing ever moves it — which is exactly why the file is 5/5 green alone and only fails in the full run.

So **the action is behaving correctly**. A pointer really is resting on the chip, and keeping the tooltip open is the documented affordance. The test is the thing that lies: it dispatches "the pointer left" while the pointer demonstrably has not. **No timeout value could ever have fixed this.**

## The fix

Re-issue the leave when a *trusted* re-entry arrives, so the assertion measures the close path rather than the harness's stray cursor. One file, most of it the explanation.

Applied to **both** elements that can absorb that re-entry: the chip (`show()` → `cancelClose()`) and the panel, whose own `pointerenter` handler is a bare `cancelClose()` with nothing to re-schedule the close. The panel is body-appended, `inset:auto`, and only positioned after `show()` by `floating-anchor`'s async `computePosition`, so it is both a valid target for the post-`leave()` re-hit-test and the larger one. The captured stuck run happened to land on the chip, but it was *rescued* by a `pointerleave` on the panel — direct evidence that the panel path strands the assertion identically.

The action is untouched. The 140 ms grace period is unchanged and still exercised for real — no fake timers, no injected override, no lengthened budget.

## What I reverted, and why

The earlier budget change is gone. Its premise was that load stretches the grace timer; across 49 instrumented runs under exactly that load, the close took **151 ms in 48 of them**. The timer never stretched once, so the change addressed a problem that does not exist.

## Verification

- **75/75 green** full `ui/` suite runs with the chip guard, then **40/40 green** after extending it to the panel (300 files, 4775 tests each).
- Baseline for comparison: 17/17 green on unmodified `main` (the flake did not fire), then **2 failures in 50 runs** (~4%) once the instrumentation made it observable, including one *with* the larger budget applied.
- `bun run check` (0 errors, 6054 files), root `eslint`, `prettier --check`, and `fallow audit --base origin/main` — all clean.

One honest note on method: I first tried to verify deterministically by parking the real cursor on the chip with `hover()`. The test passed — but so did the control with the fix removed, so that experiment does not reproduce the bug and proves nothing. The 75-run sweep is the actual evidence.

## Not addressed

#2028 (EpicPanel leaks a rejection under parallel load) is a sibling in the "only in the full run" class but a different mechanism, and is untouched here.

### DIFF
--- a/ui/src/lib/components/StatusPip.browser.test.ts
+++ b/ui/src/lib/components/StatusPip.browser.test.ts
@@ -53,15 +53,23 @@ describe("StatusPip tip mode (statusTip action)", () => {
     expect(document.querySelector(".status-tip")?.getAttribute("role")).toBe("tooltip");
     // Most specs in this suite drive *real* clicks, and the cursor they leave behind
     // persists across files sharing a worker's page — so a real pointer can be resting
-    // on this chip. As the panel settles into its anchored position the browser
-    // re-hit-tests and fires a **trusted** pointerenter, which reopens the tooltip and
-    // cancels the pending close: right for a pointer that genuinely is there, but it
-    // strands the synthetic leave below forever (no real pointer ever leaves again).
-    // Re-issue the leave on those so the assertion measures the close path rather than
-    // the harness's stray cursor. Synthetic events cannot trigger this handler.
-    pipEl().addEventListener("pointerenter", (e) => {
-      if (e.isTrusted) leave();
-    });
+    // over this component. The panel is body-appended and only positioned after show()
+    // (floating-anchor's computePosition is async, and .status-tip is inset:auto), so as
+    // it settles the browser re-hit-tests and delivers a **trusted** pointerenter — to
+    // the chip (show() → cancelClose) or to the panel, its own larger target (a bare
+    // cancelClose). Either cancels the pending close with nothing left to re-schedule
+    // it, which is right for a pointer that genuinely is there but strands the synthetic
+    // leave below forever, since no real pointer ever leaves again.
+    //
+    // Release both by re-issuing the leave, so the assertion measures the close path
+    // rather than the harness's stray cursor. Synthetic events never re-enter this.
+    const releaseTrusted = (el: Element) =>
+      el.addEventListener("pointerenter", (e) => {
+        if (!e.isTrusted) return;
+        el.dispatchEvent(new PointerEvent("pointerleave", { pointerType: "mouse", bubbles: true }));
+      });
+    releaseTrusted(pipEl());
+    releaseTrusted(document.querySelector(".status-tip")!);
     leave();
     await vi.waitFor(() => expect(tooltipOpen()).toBe(false));
   });
--- a/ui/src/lib/components/StatusPip.browser.test.ts
+++ b/ui/src/lib/components/StatusPip.browser.test.ts
@@ -51,6 +51,17 @@ describe("StatusPip tip mode (statusTip action)", () => {
     enter();
     await vi.waitFor(() => expect(tooltipOpen()).toBe(true));
     expect(document.querySelector(".status-tip")?.getAttribute("role")).toBe("tooltip");
+    // Most specs in this suite drive *real* clicks, and the cursor they leave behind
+    // persists across files sharing a worker's page — so a real pointer can be resting
+    // on this chip. As the panel settles into its anchored position the browser
+    // re-hit-tests and fires a **trusted** pointerenter, which reopens the tooltip and
+    // cancels the pending close: right for a pointer that genuinely is there, but it
+    // strands the synthetic leave below forever (no real pointer ever leaves again).
+    // Re-issue the leave on those so the assertion measures the close path rather than
+    // the harness's stray cursor. Synthetic events cannot trigger this handler.
+    pipEl().addEventListener("pointerenter", (e) => {
+      if (e.isTrusted) leave();
+    });
     leave();
     await vi.waitFor(() => expect(tooltipOpen()).toBe(false));
   });


scoop:
# Review — fix(ui): stop the StatusPip close test racing a real stray cursor (#2180)

Diff under review: `git diff 0ba04bb8...HEAD` — one file, +11 lines, all inside
`ui/src/lib/components/StatusPip.browser.test.ts` (the "hover reveals a role=tooltip popover;
leaving (unpinned) closes it" spec).

## What the change does, verified against the code

The test attaches a second `pointerenter` listener to the pip and re-issues the synthetic
`leave()` whenever a **trusted** re-entry arrives, so a real cursor resting on the chip cannot
cancel the pending close forever.

Things I checked rather than assumed:

- **Listener ordering is safe.** The action registers `onPointerEnter` during `enable()` at mount
  (`ui/src/lib/actions/statusTip.svelte.ts:194-202`, reached from `:224` on the first `use:statusTip`
  call with `tipParam` non-null — `ui/src/lib/components/StatusPip.svelte:31,75`). The test's
  listener is added later (`StatusPip.browser.test.ts:62`), so on a trusted enter the action's
  `show()` runs first and the test's `leave()` second. Reversing the order would strand the test;
  it is not reversed.
- **`show()` really does eat the pending close**, so the comment's claim is accurate:
  `cancelClose()` is called at `statusTip.svelte.ts:124` *before* the `if (open) return` guard at
  `:125`, i.e. even a no-op re-open kills the 140 ms timer scheduled at `:101-107`.
- **No recursion.** `leave()` (`StatusPip.browser.test.ts:20-21`) dispatches a `PointerEvent` via
  `dispatchEvent`, so `isTrusted` is `false` on the re-entrant path; the guard at `:63` short-circuits.
- **No cross-test leak.** `afterEach` clears `document.body` (`:11-13`), detaching the pip, so the
  never-removed listener cannot fire in a later spec.
- **No masking of a real regression.** The added listener is inert in a deterministic run (no
  trusted pointer events at all), so a genuine "never closes" regression still fails the assertion.
- **The stray-cursor premise is consistent with the harness.** The browser project runs real
  Playwright chromium (`ui/vite.config.ts:110-121`) and ~60 `*.browser.test.ts` files drive real
  locator clicks, which move the page's actual mouse; the cursor position outlives a file.
- **Types.** `HTMLElementEventMap["pointerenter"]` is `PointerEvent`, so `e.isTrusted` at `:63`
  type-checks with no annotation.

## Finding — the guard covers only one of the two elements that can eat the close

`ui/src/lib/components/StatusPip.browser.test.ts:62-64`

The panel is a *second* hover surface with its own boundary handlers: the popover's `pointerenter`
calls `cancelClose()` (`ui/src/lib/actions/statusTip.svelte.ts:83-86`) for exactly the same reason
the chip's does — it treats trigger + panel as one hover region. The test neutralizes the chip's
path and leaves the panel's untouched.

That path is reachable by the very mechanism the commit message documents (a trusted boundary
event fired ~2 ms after the synthetic `pointerleave`, without the pointer moving — Chromium
re-hit-tests after a layout change):

- `show()` appends/shows the panel at its **static** position — `.status-tip[popover]` is
  `position: fixed; inset: auto; margin: 0` (`ui/src/app.css:628-634`), so before Floating UI runs
  it sits wherever it lands in flow, next to the pip.
- `anchorPopover()` then writes `left`/`top` **asynchronously** —
  `computePosition(...).then(...)` (`ui/src/lib/floating-anchor.ts:22-30`) — so the panel *moves*
  after `show()`, on the same post-`leave()` frame that produced the observed trusted event.
- Whichever element the resting cursor lands on after that move receives the trusted
  `pointerenter`. If it lands on the panel (a ~150-260 px × ~30 px surface, versus the 9 px chip —
  `ui/src/lib/components/StatusPip.svelte:81-85`), `cancelClose()` fires, nothing re-schedules
  (the pointer never leaves the panel again), and the assertion hangs its whole budget — the exact
  reported symptom. This also applies when the panel settles *over* a cursor that was on the chip:
  the chip's trusted `pointerleave` re-schedules the close, then the panel's trusted
  `pointerenter` cancels it, and the chip guard added here never fires.

So the PR closes the landing zone that happened to be instrumented and leaves the larger sibling
zone open. Fix is small — mirror the guard on the panel (dispatch a synthetic `pointerleave` on
`document.querySelector(".status-tip")` for trusted enters, which routes to
`statusTip.svelte.ts:87-90` → `scheduleClose()`), or remove the whole class by moving the real
pointer to a neutral spot once before the assertion so no trusted boundary event can arrive at
either element.

## Stated limitations

- I did not run anything (read-only review). The commit message's empirical claims — trusted
  `pointerenter` observed 2 ms after the synthetic leave, close latency 151 ms in 48 of 49 loaded
  runs — are taken as reported; I verified only that they are consistent with the code paths cited
  above. The commit message does not state a post-fix green-run count, and I could not produce one.
- Nothing in the diff resolves against an external schema or a message catalog, so there is no
  locale-parity or third-party-contract surface to check here.

## Nits (non-blocking):

- `StatusPip.browser.test.ts:61` — "Synthetic events cannot trigger this handler" is loose: they do trigger it, they just fail the `isTrusted` guard; "cannot satisfy this guard" would say it exactly.
- `StatusPip.browser.test.ts:54-61` — eight lines of comment for two lines of code; the middle three sentences restate the mechanism the commit message already carries.

- `StatusPip.browser.test.ts:11-13` — `afterEach` wipes `document.body` without unmounting, so `statusTip`'s `teardown()`/`hide()` never runs and its `document`/`window` listeners (`statusTip.svelte.ts:135-137`) survive the spec; harmless today because the detached node makes `onDocPointerDown` a no-op, but the whole file relies on it.

## Plan drift (measurement for the operator, not a judgement)

Major: the approved plan exported `CLOSE_GRACE_MS` from `statusTip.svelte.ts` and gave the
assertion `{ timeout: CLOSE_GRACE_MS + 3000 }` across two files; the diff touches one file, adds no
timeout, and implements a different diagnosis entirely. The commit message explains why (no timeout
could have helped a close that was cancelled, not slow) — reported here only as drift.

## Decision

request-changes — one finding: the trusted-re-entry guard needs to cover the panel as well as the
chip, otherwise the flake class this PR exists to close remains open.

scoop (commented):
# Review — StatusPip close-tooltip flake (#2180), rework round 2

**Decision: comment** — no blocking findings. The prior round's finding is genuinely fixed.

## What the diff is

One file, test-only: `ui/src/lib/components/StatusPip.browser.test.ts` (+26/-9 across two commits `ffd4e8d69`, `7cefa6695`). No production file is touched — `ui/src/lib/actions/statusTip.svelte.ts` is byte-identical to the base (`git diff 0ba04bb8...HEAD` lists only the test file), so the 140 ms grace affordance is untouched, which is what the originating issue asked for.

## Prior finding — verified addressed

> ui/src/lib/components/StatusPip.browser.test.ts: the new trusted-pointerenter guard is attached only to the pip, but the popover panel has its own pointerenter handler that calls cancelClose() … Neutralize the panel path too (dispatch a synthetic pointerleave on .status-tip for trusted enters, which routes to scheduleClose at statusTip.svelte.ts:87-90) …

Fixed, and by exactly the suggested route. `StatusPip.browser.test.ts:66-72` now factors the guard into `releaseTrusted(el)` and applies it to **both** `pipEl()` (line 71) and `document.querySelector(".status-tip")!` (line 72).

I traced both paths against the action:

- **Panel.** A trusted `pointerenter` on the panel hits the action's listener first (`ui/src/lib/actions/statusTip.svelte.ts:83-86`, registered inside `ensurePopover()` at mount-of-panel time, i.e. before the test's listener) → bare `cancelClose()`. The test's listener then runs and dispatches a synthetic `pointerleave` on the same element, which the action's panel handler (`ui/src/lib/actions/statusTip.svelte.ts:87-90`) routes to `scheduleClose()` — `pointerType: "mouse"` clears the touch guard on line 88 and `pinned` is false in this test (no `click()` in the unpinned spec). Net effect: close re-armed. Verified.
- **Chip.** Trusted `pointerenter` → `onPointerEnter` (`statusTip.svelte.ts:152-155`) → `show()` → `cancelClose()` (line 124), then early-returns on `if (open) return` (line 125). The test's listener then dispatches `pointerleave` → `onPointerLeave` (`statusTip.svelte.ts:156-160`) → `scheduleClose()`. Net effect: close re-armed. Verified.

So both cancel-paths in the action are now covered; those are the only two `cancelClose()` call sites reachable from a pointer boundary event (`cancelClose` is otherwise called only from `show()`/`hide()`/`scheduleClose` itself, at `statusTip.svelte.ts:102`, `124`, `141`).

## Correctness checks I actually ran

- **Convergence.** The guard is self-healing rather than one-shot: any trusted enter on either element ends with a `scheduleClose()`, and the synthetic `pointerleave` cannot itself produce another `pointerenter` (`dispatchEvent` does no hit-testing), so there is no ping-pong. The `if (!e.isTrusted) return` guard on `StatusPip.browser.test.ts:68` is belt-and-braces, not load-bearing.
- **Post-`hide()` re-open.** After the 140 ms timer fires, `hide()` → `stopAnchor()` → `floating.hidePopover()` (`ui/src/lib/floating-anchor.ts:33-40`) and `.status-tip[popover]:not(:popover-open) { display: none }` (`ui/src/app.css:650-653`) removes the panel from hit-testing. If the real cursor was over the panel, the browser then delivers a trusted enter to whatever is underneath. That element cannot be the chip: `anchorPopover` is called with `placement` defaulting to `"bottom"` and `offset(6)` (`statusTip.svelte.ts:134`, `floating-anchor.ts:20,26`), and `shift()` only moves along the cross axis, so the panel never overlaps the 9px `.pip` span (`ui/src/lib/components/StatusPip.svelte:78-82`). And even if it did, the chip's own `releaseTrusted` would immediately re-arm the close. No stuck-open path found.
- **Ordering.** `releaseTrusted` is registered after the open assertion (`StatusPip.browser.test.ts:52`) and before `leave()` (line 73). A trusted enter landing in the earlier window is harmless — it calls `cancelClose()` when `closeTimer` is already `null` (`statusTip.svelte.ts:94-98`), i.e. a no-op, and the later `leave()` still schedules the close.
- **`document.querySelector(".status-tip")!` is safe.** Line 52's `vi.waitFor` has already asserted `.status-tip:popover-open` matches (`tooltipOpen()`, line 24), so the element exists; line 53 already relies on the same query. There is no stale panel from earlier specs in the file: tests 1–2 never call `enter()`, so `ensurePopover()` (`statusTip.svelte.ts:66-92`) never runs, and `afterEach` clears `document.body` including the body-appended panel (`StatusPip.browser.test.ts:11-13`, `statusTip.svelte.ts:91`). `.status-tip-dialog` is a distinct class token and does not match.
- **No repositioning churn.** `autoUpdate` is used with defaults (`floating-anchor.ts:22`), so no `animationFrame` polling; the `.pulse::after` halo animation (`StatusPip.svelte:130-141`) scales a `pointer-events: none` pseudo-element and does not change the reference's border box, so it cannot drive a ResizeObserver reposition loop that would repeatedly re-arm the 140 ms timer inside the default 1000 ms `vi.waitFor` budget.
- **Comment accuracy.** Each factual claim in the new comment block checks out: panel is body-appended (`statusTip.svelte.ts:91`); positioned only after `show()` via an async `computePosition().then(...)` (`floating-anchor.ts:22-30`); `.status-tip` is `position: fixed; inset: auto; margin: 0` (`ui/src/app.css:628-634`), so it renders at its static position until the promise resolves; the chip path is `show() → cancelClose` and the panel path is a bare `cancelClose` (`statusTip.svelte.ts:152-155, 124` and `83-86`); and "most specs in this suite drive real clicks" holds — 738 `.click()` occurrences across 74 `*.browser.test.ts` files.
- **Diagnosis vs. the issue.** The issue proposed a timing-budget fix (explicit `vi.waitFor` timeout or an injectable grace). This PR rejects that diagnosis on measurement rather than on argument: `ffd4e8d69`'s message reports instrumenting the failing run and observing a trusted `pointerenter` 2 ms after the synthetic leave, with close latency of 151 ms in 48 of 49 runs under load. That is consistent with everything I can read in the code (the whole `hide()` chain after the timer is synchronous), and it explains why no timeout increase could have helped — a cancelled close never re-arms. I take the empirical claim at face value; I cannot re-run it here (read-only review), so I record it as evidence I did not independently reproduce rather than as something I verified.

## Limitations

- I did not run the suite (read-only review; tests/lint are CI's job). The empirical claims in the commit messages — the 2 ms trusted re-entry, 151/49-run close latency, and the "instrumented stuck run was rescued by a pointerleave on the panel" — are not verifiable from the tree; I checked only that they are consistent with the code paths cited above.

## Nits (non-blocking):

- `StatusPip.browser.test.ts:54`: "Most specs in this suite" reads as this file, which drives no real clicks at all — "this suite" means the wider `ui/` browser suite; naming that explicitly would remove the ambiguity.
- `StatusPip.browser.test.ts:66-70`: `releaseTrusted` never detaches its listeners; harmless here since `afterEach` discards both elements, but a returned disposer would make that explicit.
- `StatusPip.browser.test.ts:77-88`: the sibling pinned spec closes via `Escape` and is not given the same guard; it looks safe today (after `hide()` the panel cannot reveal the chip beneath it, per the placement analysis above), but if the stray-cursor mechanism ever generalises this is the next spec to protect.

- None — the diff is test-only and strictly smaller than the approved plan.

- None of the curated rules touch this diff (no plan-phase gating, toasts, WS enrichment, coach targets, relative-age helpers, session steering, or BacklogView paths involved).

- `ui/vite.config.ts` caps `maxWorkers: 2` for the browser project only when `CI` is set, so local full runs still spawn ~nproc concurrent chromium pages — the load profile the issue describes.

PR #2189, opened 2026-09-04
feat(newtask): name the phase a slow start is stuck in, and allow cancelling it

kai-osthoff:
## Why

Clicking **Create and start** can sit on "Spawning…" for a long time with nothing to go on. That was the report; this PR answers both halves of it — *what* it is doing, and *why* it takes that long.

`POST /api/sessions` only answers once the agent is actually running. `NewTask.svelte` → `doSpawn()` holds `submitting` for exactly that request, and inside it `SessionService.create()` runs five sections in series:

| Section | Symbol | Cost |
| --- | --- | --- |
| Update base branch | `resolveBaseRef()` → `upstreamStatus()` | `git fetch`, 5s timeout |
| Create worktree | `Worktrees.create()` | `git worktree add`, **synchronous** |
| Compose prompt | `composePromptArg()` → `fetchIssueCommentsBlock()` | `gh issue view`, ~0.6s measured |
| Prepare launch | `resolveCreateLaunch()`, `runSpawnHookPatch()` | plugin hooks, 5s each |
| **Wait for the agent** | `prepareSpawnOrThrow()` → `HerdrDriver.start()` → `resolveByAutoDetect()` | **polls every 500ms, up to 30s** |

The last one dominates. For `sandboxApplied = "trusted"` Shepherd deliberately does *not* register the agent itself — it waits for herdr to detect the `claude` process in the pane. In practice that means waiting for the agent CLI to boot. Capped at 30s, after which the create fails and rolls the worktree back; the reporter's log has `herdr: agent for pane … not auto-detected within 30s` in it repeatedly.

Two things made this opaque: **nothing measured those sections** (the instrumentation in `src/instrument.ts` sits behind `SHEPHERD_PROFILE_LOOP`), and the dialog received no signal at all between click and answer.

## What changed

`create()` now runs its sections through a phase tracker (`src/spawn-progress.ts`) that:

- **always measures** — one `[create] spawn ok base 1.2s worktree 0.4s prompt 0.6s launch 0.2s agent 18.3s total 20.7s` line per spawn, on the failure and cancel paths too, so the question is answerable after the fact;
- **reports each phase** to the dialog over the existing WS channel when the caller supplies an `X-Shepherd-Spawn-Id` header;
- **owns the `AbortSignal`** that reaches herdr's auto-detect poll loop, so cancelling lands *inside* the long wait instead of only between phases.

Past 10s the dialog names the running phase, says why it can take a while, lists what already finished with durations, and offers **Cancel start**. Below that threshold only the footer changes — it names the phase and counts up.

## Decisions worth reviewing

- **The spawn id travels as a header, not a body field.** The create body is what a usage hold persists as a held task; an id replayed from there would address a spawn that is long gone. This also keeps the create schema, held persistence and the extension untouched.
- **Cancellation stops at the point of no return.** Once `HerdrDriver.start()` returns, the cancel route answers `canceled: false` and the spawn completes — tearing down a live agent and its worktree is worse than a redundant session. The signal is checked at *defined* points only (before `tab.create`, once per poll round), so the existing `closeTab` / `worktree.remove` rollback covers it with no new teardown path.
- **A cancelled start is not an error.** `409` + `code: "spawn_canceled"` renders a neutral note, keeps the typed prompt and re-arms the buttons.
- **One threshold, not per-phase budgets.** The plan originally gated the explanation on a per-phase budget as well; that would have suppressed the explanation for a phase that had just begun inside an 11-second start — silent exactly when someone is looking. `SPAWN_SLOW_MS = 10_000` is the only threshold.
- **`.fallowrc.jsonc`**: the `src/service.ts:create` complexity cap now targets `createAgentSession`. The body kept its cap but not its name when the thin wrapper was split off; branching is unchanged (24 against the same 25 bar).

## Scope

New Task only, by the reporter's choice. Held-task "Spawn now", relaunch and the extension popup share the same wait and are **not** covered here. Also out of scope: shortening the 30s auto-detect budget (that wait is largely the agent's own startup), timestamps on every log line, and non-blocking create.

## Verification

- `bun run lint`, `bun run test` — 9095 pass / 0 fail
- `cd ui && bun run check`, `bun run check:i18n`, `bun run test` — 4785 pass / 0 fail
- feature-catalog + announcement-version + glossary gates green; `fallow audit` clean
- New coverage: `test/spawn-progress.test.ts` (phase order, log line, registry, cancel), `test/service-spawn-phases.test.ts` (signal reaches `herdr.start`; a cancel mid-wait rolls the worktree back and persists no session), `test/spawn-cancel-endpoint.test.ts`, plus 7 `NewTask.browser.test.ts` cases and the store seam.

**Not measured:** the actual second-by-second split of a real start on the reporter's machine. The root cause above is grounded in the code, the `sandboxApplied = trusted` rows in their DB and the repeated 30s timeouts in their log — but the exact distribution is precisely what nothing recorded. The new log line supplies it from the first spawn after this merges.


kai-osthoff (commented):
No blocking findings.

Verified the first prior finding: `spawnId` is removed before the edit-held and relaunch branches, while the ordinary-create branch alone passes it as the transport argument (`ui/src/routes/+page.svelte:2148`, `ui/src/routes/+page.svelte:2151`, `ui/src/routes/+page.svelte:2153`). The receiving create API puts that value in `X-Shepherd-Spawn-Id`, not the JSON body (`ui/src/lib/api.ts:242`, `ui/src/lib/api.ts:244`, `ui/src/lib/api.ts:250`); the held PATCH continues to validate its body with `validateCreate` (`src/server.ts:5746`, `src/server.ts:5749`).

Verified the second prior finding: both 0.7.5 drivers check cancellation after launching the pane and roll back their captured tab on the resulting exception (`src/herdr.ts:1187`, `src/herdr.ts:1192`, `src/herdr.ts:1208`; `src/herdr-socket-driver.ts:233`, `src/herdr-socket-driver.ts:237`, `src/herdr-socket-driver.ts:250`). The remaining race after a driver has returned is resolved before persistence: a cancelled tracker causes `herdr.stop` and then the existing worktree rollback path, rather than a session row (`src/service.ts:3754`, `src/service.ts:3755`, `src/service.ts:3758`, `src/service.ts:3848`).

Verified the UI/server wiring: the authenticated route dispatch includes the cancel endpoint (`src/server.ts:7907`, `src/server.ts:7975`), the server emits typed phase payloads (`src/server.ts:2301`, `src/server.ts:2306`; `ui/src/lib/types.ts:2038`, `ui/src/lib/types.ts:2049`), and the dialog renders the active phase, elapsed time, explanation, completed phase durations, and cancel control after its 10-second threshold (`ui/src/lib/components/NewTask.svelte:297`, `ui/src/lib/components/NewTask.svelte:307`, `ui/src/lib/components/NewTask.svelte:2298`).

Locale parity was checked for all 18 added phase/progress and announcement keys: English and German both define the same set (`ui/messages/en.json:3769`, `ui/messages/de.json:3769`). Targeted regression coverage for both cancellation races and the held-body boundary is present but was not executed, per review scope (`test/herdr-075.test.ts:230`, `test/service-spawn-phases.test.ts:139`, `test/validate.test.ts:419`).

kai-osthoff:
✅🟢🎉 All checks green 💚✨ PR just chilling here 🛋️😴💅 Someone press merge? 👉🔘🙏🚀

PR #2197, opened 2026-09-07
fix(herdr): key the agent_name_taken squatter eviction on the tab label (#2033)

### LINKED ISSUE #2033: Remaining agent.name consumers on herdr 0.7.5: name-taken squatter eviction can never match

kai-osthoff:
Closes #2033.

Follow-up to #2029, which fixed every `agent.name` consumer whose failure leaked a tab. This takes the ones that were deliberately left behind, plus one that #2029 missed.

## The bug

herdr 0.7.5 sends no `name` on agent records, so `(await this.listAsync()).filter((a) => a.name === name)` selected the empty set on every 0.7.5+ host. Nothing was evicted, all three retry attempts re-collided, and a sandboxed spawn failed with `agent_name_taken` instead of self-healing.

Two of the four collision sites are actually live-broken. Both `startAgentWithCollisionRetry` implementations sit behind `!herdrUsesExternalRegistrationSpawn()`, so they are unreachable on 0.7.5 and correct on ≤0.7.4 where `name` is present. The **register** pair is what fails in the field.

## The fix

One shared `agentsHoldingName(agents, rawName, labels)` in `src/herdr.ts`: a `name` hit wins outright and the labels are never read; otherwise the **tab label** decides, compared in herdr's own name space (`sanitize` on both sides), since `tab create --label` carries the raw name while the register call binds the sanitized one. No labels available narrows the match to nothing rather than widening it — a failed `tab list` must never broaden what gets closed.

The captured fixtures already prove the join, so no new ones were needed:

| surface | value |
| --- | --- |
| `agent-list-registered.json` | `{ agent: "review-task-09", tab_id: "t_075" }` — no `name` key |
| `tab-list-helpers.json` | `{ tab_id: "t_075", label: "review TASK-09" }` |
| `sanitizeHerdrAgentName("review TASK-09")` | `"review-task-09"` — the name the register call bound |

**Not keyed on `agent`**, where the issue notes the label also surfaces. That field holds the bare agent KIND (`"claude"`) for trusted auto-detected agents, so matching it would both alias the kind *and* miss an auto-detected squatter, which reports the kind rather than its name. The tab label has neither problem, so the caveat that kept this out of #2029 dissolves rather than being worked around. The error string still goes unparsed, per the standing comment at all four sites.

### Two consequences worth reviewing

**The label branch is a VERSION SWITCH, not a fall-through.** *(Revised after review — an earlier revision of this PR had it as a fall-through and defended that; the critic was right.)* Any record carrying `name` proves this herdr populates the field, so its name space is used exclusively and a miss evicts nothing. Only a list where nobody carries `name` (0.7.5+) reaches the label branch, which makes ≤0.7.4 genuinely byte-identical.

This matters because the label comparison is necessarily *coarser* than herdr's ≤0.7.4 raw name space: `sanitize` is lossy, so `fix login` and `fix-login` are two distinct live sessions sharing one sanitized label. As a fall-through, a raw-name collision on one could have evicted the other. Where herdr tells us the names, believe it.

The same raw/sanitized mismatch is why `uniqueName` now de-dupes in sanitized space — that is the space herdr 0.7.5+ actually binds, and de-duping raw let both twins go live in the first place. Sanitized comparison is a superset of the raw one (equal raw implies equal sanitized), so it never de-dupes less.

**Each site excludes its own freshly-created handle** (register on `paneId`, `agent start` on `tabId`). Tab-label matching newly makes our own tab a candidate, since it was labelled with the same name moments earlier; without the guard a repaired eviction could close the pane it is spawning into.

## `uniqueName` — the third consumer, not named in the issue

`service.ts` → `uniqueName()` built its taken set from `agent.name` alone, so on 0.7.5 the live-agent half of the de-dupe was dead and only the `shepherd/<candidate>` branch check remained. That is the state a *repaired* eviction must never be handed: with duplicates getting through, a colliding spawn would evict a **live sibling's** tab. It now takes the union of both surfaces, extracted into `takenHerdrNames()` (which also brings `uniqueName` back under the fallow complexity gate).

`review.ts` → `findSquatter()` rides the same helper, so a restart-orphaned critic is found by its tab label instead of silently degrading to the cwd fallback.

## Verification

`bun run lint`, `bun run typecheck`, `bun run test` green — **9131 pass, 0 fail**.

Every new behaviour test was confirmed **red** against the original source before the fix, not assumed:

| test | proven red against |
| --- | --- |
| CLI register collision (2) | the exact old predicate `a.name === agentName`, surgically restored |
| socket register collision (2) | original source |
| `findSquatter` by tab label (1) | original source |
| `uniqueName` de-dupe by label (1) | original source |
| version switch: a `name` miss evicts nothing (1) | the fall-through revision |
| `uniqueName` sanitized-space de-dupe (1) | the raw-comparison revision |

The pre-existing ≤0.7.4 collision tests in `test/herdr.test.ts` pass **untouched**.

The issue notes that verifying this means forcing a real name collision, which no test did. `herdr:compat` gains **L10**: it registers its own name on a fresh pane, holds it at `working`, then attempts the same name on a second pane — answering the question the code itself records as unverified (*"Whether `report-agent(-session)` even emits `agent_name_taken` is unverified"*). It is deliberately self-contained rather than reusing L6's agent, because L7 has already flipped that one to `idle` by then and a herdr that releases a finished agent's name would make the probe answer "no collision" for the wrong reason. Either answer is fine for Shepherd — one that accepts the duplicate just makes the retry defensive — so the verdict is PASS on agreement with the baseline and REVIEW on a flip, mirroring how L5 treats last-tab behaviour.

## Out of scope

- Re-checking against 0.8.x — that is #2032. This fix is version-agnostic by construction: it prefers `name` wherever herdr sends it, so a herdr that restores the field needs no follow-up.
- Regenerating `docs/herdr-compat/0.8.2.md` with the L10 row: that report is a captured artefact of a real A/B run against downloaded binaries, regenerated by the next version-bump PR.

One plan item was dropped: a unit test asserting the L10 row renders. `scripts/herdr-compat.ts` is a top-level script that `process.exit()`s at import and exports nothing, so `liveChecks` cannot be imported by a test, and no existing L1–L9 row has one. Making it testable would mean restructuring the script's entrypoint, which does not trace to this issue.

### DIFF
--- a/.claude/rules/herdr-version-bump.md
+++ b/.claude/rules/herdr-version-bump.md
@@ -52,10 +52,11 @@ shipped stable).
 2. **Run the check:** `bun run herdr:compat -- --candidate <version>`. Static half: schema diff,
    the #2032 record-shape gate (`TabInfo`/`PaneInfo`/`AgentInfo` required/nullable drift = FAIL),
    and a `--help` diff over every subcommand Shepherd invokes. Live half: candidate vs. baseline
-   (= current ceiling) as isolated servers, probes L1–L9 (reaper assumptions from #2029/#2032,
+   (= current ceiling) as isolated servers, probes L1–L10 (reaper assumptions from #2029/#2032,
    tab-id reuse #569, last-tab behaviour #1760, the external-registration spawn replay #1890,
-   the #1716 idle probe, the status surface, and the terminal contract via
-   `scripts/verify-herdr-terminal.ts`). Report: `docs/herdr-compat/<version>.md`, exit 1 on FAIL.
+   the #1716 idle probe, the status surface, the terminal contract via
+   `scripts/verify-herdr-terminal.ts`, and whether a duplicate `--agent` registration collides
+   #2033). Report: `docs/herdr-compat/<version>.md`, exit 1 on FAIL.
 3. **Triage every REVIEW, fix or consciously accept every FAIL.** Behavioural changes get code
    (as #2056's last-tab guard did) and a test pinning the new behaviour.
 4. **Regenerate the vendored protocol** against the candidate:
--- a/scripts/herdr-compat.ts
+++ b/scripts/herdr-compat.ts
@@ -9,7 +9,7 @@
  * Static half (no server): schema diff + #2032 record-shape gate from `api schema --json`,
  * plus a `--help` surface diff over every subcommand Shepherd drives. Live half: candidate and
  * baseline each run as an ISOLATED headless server (own HOME/XDG/socket — the operator's
- * daemon is never touched) and the L1–L9 probes are measured A/B. Output: a markdown report
+ * daemon is never touched) and the L1–L10 probes are measured A/B. Output: a markdown report
  * at docs/herdr-compat/<candidate>.md (committed by the eventual bump PR) and exit 1 iff any
  * check FAILs (REVIEW items are triage work, not machine verdicts).
  */
--- a/scripts/herdr-compat.ts
+++ b/scripts/herdr-compat.ts
@@ -228,6 +228,28 @@ function liveChecks({ base, cand }: AB, l9: { ran: boolean; exit: number | null 
       : "Did not run (no workspace or earlier failure) — run it by hand against a candidate server.",
   });
 
+  checks.push({
+    id: "L10",
+    title: "duplicate --agent registration collides (squatter eviction, #2033)",
+    verdict:
+      cand.duplicateNameRejected === null
+        ? "REVIEW"
+        : cand.duplicateNameRejected === base.duplicateNameRejected &&
+            cand.duplicateNameErrorCode === base.duplicateNameErrorCode
+          ? "PASS"
+          : "REVIEW",
+    details:
+      abTable([
+        ["duplicate rejected", show(base.duplicateNameRejected), show(cand.duplicateNameRejected)],
+        ["error code", show(base.duplicateNameErrorCode), show(cand.duplicateNameErrorCode)],
+      ]) +
+      (cand.duplicateNameRejected === true
+        ? "\n\nThe register-path `agent_name_taken` retry is LIVE: its squatter eviction (#2033, keyed on the tab label) is what keeps a colliding spawn self-healing."
+        : cand.duplicateNameRejected === false
+          ? "\n\nThis herdr ACCEPTS a duplicate registration, so the register-path collision retry is defensive only — harmless, and it guards a herdr that starts refusing. A flip either way is a behaviour change worth a look."
+          : "\n\nNot measured — re-run, or exercise it by hand before trusting the register-path collision retry."),
+  });
+
   const notes = [
     ...base.notes.map((n) => `baseline: ${n}`),
     ...cand.notes.map((n) => `candidate: ${n}`),
--- a/scripts/herdr-compat/probes.ts
+++ b/scripts/herdr-compat/probes.ts
@@ -248,20 +248,28 @@ export async function runProbes(
     for (const argv of register(heldPane)) await runOk(server, argv);
 
     // The colliding spawn. The refusal can surface on either half of the register pair.
+    //
+    // The verdict is accumulated LOCALLY and published only once the pair has actually been run
+    // through: writing `false` up front would let a throw part-way (a dead server, a failed run)
+    // land in the report as "this herdr ACCEPTS duplicates" — a measurement nobody made. An abort
+    // must leave the field null, which the report renders as "not measured".
     const dupPane = await livePane(`${squatName}-dup`);
-    o.duplicateNameRejected = false;
+    let rejected = false;
+    let rejectionCode: string | null = null;
     for (const argv of register(dupPane)) {
       const res = await server.run(argv);
       const code = errorCode(res.stdout) ?? errorCode(res.stderr);
       if (res.exitCode !== 0 || code !== null) {
-        o.duplicateNameRejected = true;
-        o.duplicateNameErrorCode = code;
+        rejected = true;
+        rejectionCode = code;
         break;
       }
     }
+    o.duplicateNameRejected = rejected;
+    o.duplicateNameErrorCode = rejectionCode;
   } catch (err) {
-    // `duplicateNameRejected` is assigned only after the squatter is bound and holding, so an abort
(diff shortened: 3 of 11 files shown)


kai-osthoff:
# Review: request changes

## Findings

1. src/herdr.ts: `agentsHoldingName()` falls through to tab-label matching whenever no record has the *requested* name, even if the daemon is supplying `name` on its records. That can close an unrelated live agent whose tab label happens to sanitize to the requested value: the helper filters only exact names at `src/herdr.ts:557`, then unconditionally uses labels at `src/herdr.ts:559-565`; both driver retry paths close every result at `src/herdr.ts:1127-1130` and `src/herdr.ts:1410-1413` (with the socket equivalents at `src/herdr-socket-driver.ts:461-464` and `src/herdr-socket-driver.ts:367-370`). The agent mapper confirms that a present `name` is authoritative data rather than a derived label (`src/herdr.ts:461-470`). Restrict label matching to records from a name-less response (as the intended compatibility branch requires), so a name-capable daemon keeps the former strict eviction behavior.

2. src/service.ts: `takenHerdrNames()` compares tab labels verbatim while the 0.7.5 registration path claims the sanitized form. A live `review TASK-01` label therefore does not reserve `review-task-01`: the new set inserts raw labels at `src/service.ts:4365-4377`, while `uniqueName()` performs exact set membership at `src/service.ts:4405-4412`; the driver lowers/replaces/truncates at `src/herdr.ts:517-520` before passing `--agent` at `src/herdr.ts:1368-1389`. This leaves the duplicate that the new preflight is supposed to prevent, then lets collision eviction target the already-live tab. Reserve tab labels in the same sanitized namespace as registration (while retaining raw agent names for legacy compatibility).

3. scripts/herdr-compat/probes.ts: L10 records a successful duplicate outcome before both duplicate commands have completed. It sets `duplicateNameRejected = false` at `scripts/herdr-compat/probes.ts:250-253`, then a thrown `server.run()` reaches the catch at `scripts/herdr-compat/probes.ts:262-267` without restoring `null`, despite the adjacent comment claiming aborts are reported as not measured. `scripts/herdr-compat.ts:235-250` consequently can render that failed measurement as an acceptance (and even PASS when baseline is also false). Assign `false` only after both calls complete, or reset both L10 fields to null in the catch.

4. scripts/herdr-compat.ts: the file-level description still says the live half measures L1–L9 at `scripts/herdr-compat.ts:10-14`, but the runner now emits L10 at `scripts/herdr-compat.ts:231-251`. Update the stale contract comment so the entrypoint documents the probe set it actually runs.

## Verification notes

I inspected the specified merge-base diff and resolved the added helper, both CLI/socket retry call sites, the `ServiceDeps` tab surface, and the compatibility observation/rendering flow. I did not run builds or tests, per review scope.

kai-osthoff (commented):
# Review: comment

The four re-review items are resolved. `agentsHoldingName` uses `name` exclusively whenever any record supplies it, and uses tab labels only when the field is absent everywhere; the helper and its coverage demonstrate both the strict branch and the no-fall-through behavior (src/herdr.ts:555, src/herdr.ts:560, test/herdr-075.test.ts:506, test/herdr-075.test.ts:515). Both collision paths call that helper and exclude the newly created tab or pane before closing candidates (src/herdr.ts:1129, src/herdr.ts:1130, src/herdr.ts:1412, src/herdr-socket-driver.ts:366, src/herdr-socket-driver.ts:460).

`uniqueName` now normalizes both agent names and tab labels before comparison, including the candidate, so sanitized-equivalent labels are reserved (src/service.ts:4379, src/service.ts:4381, src/service.ts:4384, src/service.ts:4386, src/service.ts:4422). The new tests cover a name-absent record and punctuation/case-equivalent label (test/service.test.ts:954, test/service.test.ts:985).

The L10 observation remains null until both duplicate-registration commands finish; an exception leaves the result unmeasured instead of reporting acceptance (scripts/herdr-compat/probes.ts:250, scripts/herdr-compat/probes.ts:256, scripts/herdr-compat/probes.ts:268, scripts/herdr-compat/probes.ts:270). L10 is rendered by the compatibility entrypoint, and its entrypoint comment and version-bump procedure now enumerate L1–L10 (scripts/herdr-compat.ts:12, scripts/herdr-compat.ts:231, .claude/rules/herdr-version-bump.md:55).

One non-blocking documentation inconsistency remains: adding L10 makes the stable-check-id catalog comment in the report renderer stale; it still says the live range ends at L9 (scripts/herdr-compat/report.ts:13), while the new producer emits L10 (scripts/herdr-compat.ts:231).

Limitation: I did not execute the live compatibility probe or test suite, per the requested read-only review; the external herdr behavior cannot be established from this repository alone.


**Nits (non-blocking):**

- [compliance] scripts/herdr-compat.ts: Adding L10 leaves the report renderer's stable-check-id catalog comment ending at L9, so the internal probe catalog documentation is stale.

PR #2199, opened 2026-09-07
feat(codex): close the Codex role-spawn config-isolation gap (#2134)

### LINKED ISSUE #2134: Codex role spawns inherit the operator's full user config — close the isolation gap
Codex role spawns inherit the operator's full user config; Claude role spawns do not. Codex 0.150.1
now has the flags to close that.

## The asymmetry

`buildTransientAgentArgv` gives **every** Claude role (`src/transient-agent-argv.ts:244-263`):

- `disableAllHooks: true` — strips inherited global hooks
- `tui: "default"` — the unattended-pane kill switch for renderer upsell dialogs
- `--disable-slash-commands` — removes skills entirely
- a per-kind `--allowedTools` allowlist
- `--permission-mode dontAsk`

plus, for the `mcpIsolated` presets (`reviewer`, `doc`), `--safe-mode` coupled to
`enableAllProjectMcpServers` (`src/transient-agent-argv.ts:170-173,245,258`).

A Codex role gets `--sandbox workspace-write` and nothing else (`src/codex-role-argv.ts:31-45`). It
therefore loads the operator's `~/.codex/config.toml` in full: MCP servers, hooks, project rules,
plugins. The sandbox constrains *writes*; it does not constrain what the agent is *wired to*.

That gap was flagged as unreviewed in `docs/research/claude-code-codex-release-parity-2026-07-29.md`
§4 and is still open.

## What is available now

Verified on codex-cli **0.150.1** (`codex exec --help`):

| Flag                              | Effect                                                                   |
| --------------------------------- | ------------------------------------------------------------------------ |
| `--ignore-user-config`            | Do not load `$CODEX_HOME/config.toml`; auth still uses `CODEX_HOME`       |
| `--ignore-rules`                  | Do not load user or project execpolicy `.rules` files                     |
| `--ephemeral`                     | Run without persisting session files to disk                             |
| `--dangerously-bypass-hook-trust` | Run enabled hooks without persisted hook trust — **not** wanted here      |

## Scope

- Decide per transient kind which of `--ignore-user-config` / `--ignore-rules` applies, mirroring the
  reasoning already written down for the Claude presets rather than inventing a second rationale.
- Watch the MCP interaction: `docs/research/mcp-parity-across-runtimes.md` documents that Codex reads
  its MCP servers from `~/.codex/config.toml`. `--ignore-user-config` removes them from role spawns —
  that is the intent for isolation, but it must be a deliberate decision per kind, not a side effect.
- `--ephemeral` is worth considering for disposable-tmpdir roles, but it interacts with rollout-based
  telemetry (#1816) — do not flip it without checking that leg.

## Acceptance

- [ ] Each Codex transient kind has a written, per-kind isolation posture, symmetric to its Claude
      counterpart's.
- [ ] The argv builder tests cover the new flags the way the Claude byte-identity tests do.
- [ ] The MCP consequence of `--ignore-user-config` is stated explicitly per kind.

Background: `docs/research/codex-sdk-evaluation.md` §5 and §6 step 4 (PR #2132).

kai-osthoff:
Closes #2134.

A Codex transient role spawned as `codex exec --sandbox workspace-write` and nothing else, so it
loaded the operator's **entire** `$CODEX_HOME/config.toml` — MCP servers, connector apps, hooks,
exec policy and the model/effort defaults. The sandbox bounds what the agent may **write**; it says
nothing about what the agent is **wired to**. Every Claude role has been isolated from the
equivalent surface all along.

Every Codex role now also carries:

```
--skip-git-repo-check --ignore-user-config --ignore-rules
-c project_doc_fallback_filenames=["CLAUDE.md"]
```

## Per-kind posture

Uniform across all four transient kinds; the *reason* is per-kind.

| Kind | cwd | MCP consequence of `--ignore-user-config` |
| --- | --- | --- |
| `reviewer` | untrusted git worktree | Drops `[mcp_servers.*]` — the direct analog of the Claude preset's `--safe-mode`, which likewise disables MCP *loading*. Codex has no interactive project-MCP approval gate, so nothing corresponds to `enableAllProjectMcpServers` and none is emitted. |
| `doc` | untrusted git worktree | Same as `reviewer`. |
| `writer-ro` | disposable tmpdir | **Stricter than its Claude counterpart, deliberately.** Those Claude kinds are not `mcpIsolated` because `dontAsk` + a closed `--allowedTools` deny every `mcp__*` call, so a loaded server is inert. Codex has **no** tool allowlist — dropping the config is the only lever. |
| `writer-only` | disposable tmpdir | Same as `writer-ro`. |

## Two findings that only surfaced under measurement

Planned against live `codex exec` probes on **codex-cli 0.152.0**, not the changelog.

1. **`--ignore-user-config` makes `--skip-git-repo-check` mandatory.** Codex refuses to start
   outside a git repo or a `[projects.*]` trust entry — and those entries live in the file we just
   stopped loading. Without it every tmpdir role dies at spawn on `Not inside a trusted directory
   and --skip-git-repo-check was not specified`. **This is also a latent bug that predates this PR:**
   I reproduced the same failure in `/var/tmp` *without* any new flag. Shepherd's Codex tmpdir roles
   (recap, namer, verify-key, autopilot, distiller, optimizer, merge-suggest, task-shape) only ever
   worked on hosts whose config happened to trust the tmp root. This PR closes that too.
2. **`--ignore-user-config` does NOT drop `$CODEX_HOME/AGENTS.md`** — it covers `config.toml` only.
   So it buys no project-instruction isolation, which is why re-asserting the `CLAUDE.md` fallback
   costs nothing in security and keeps Codex at parity with a Claude role (which reads the repo's
   `CLAUDE.md`). Dropping it would have closed one asymmetry and opened another.

Also verified: `-m` and `-c model_reasoning_effort=` both survive the flag; auth still resolves via
`CODEX_HOME`; the recorded `turn_context` is identical with and without `--skip-git-repo-check`
(`approval_policy: never`, `workspace-write`, `network_access: false`).

## Behaviour change worth knowing

Roles that pass **no** effort (namer, verify-key, recap, autopilot, distiller, optimizer,
merge-suggest, doc) stop inheriting the operator's `model_reasoning_effort` and fall back to the
model's own default — cheaper, possibly less thorough. Same for `model`. This makes
`resolveRoleEnvironment`'s documented semantics ("null → provider default", "the role's natural
default") true instead of silently meaning "whatever is in the operator's config.toml". The critic
and plan reviewer pass effort explicitly and are unaffected; any role is raisable in the UI.

## Deliberately not used

- **`--ephemeral`** — would suppress the rollout files role activity and token totals are read from
  (`src/codex-activity.ts`, `CodexRolloutResolver`, #1816). Pinned as a test negative.
- **`--dangerously-bypass-hook-trust`** — grants hooks without trust, the opposite of isolation.

## Accepted residuals (documented in `docs/sandbox-security.md` §R4)

1. `$CODEX_HOME/AGENTS.md` still loads — symmetric with a Claude role reading `~/.claude/CLAUDE.md`,
   so a shared residual, not a new asymmetry.
2. **No `--allowedTools` analog.** A Codex role can still run arbitrary shell inside its sandbox and
   cannot be narrowed to the reviewer's read-only tool set (`--sandbox read-only` would also block
   the verdict-file write the result contract needs). Largest remaining Claude↔Codex asymmetry.
3. A custom `model_provider` / `--oss` operator loses that wiring. Unguarded — a guard would mean
   re-reading the very config we ignore.

## Tests

`test/transient-agent-argv.test.ts` gains five tests at the same byte-identity rigor as the Claude
side: the isolation block for every kind × model × capture combination, `--skip-git-repo-check`
travelling with `--ignore-user-config`, the negatives above, both `-c` uses coexisting in order, and
a cross-kind byte-identity assertion so a future per-kind carve-out has to break it on purpose. The
expectation is spelled out **by hand** (like the existing `READONLY_GIT`) — importing it from `src/`
would turn the gate into a tautology. Six consumer suites' argv literals updated to match.

## Verification

`bun run lint` clean · `bun run test` 8882 pass / 0 fail · `prettier --check` clean on every touched
file.


kai-osthoff:
# Review: request changes

## Findings

1. **docs/research/codex-sdk-evaluation.md:30,212-219:** This changed research document still describes the current helper-role argv as containing only `--sandbox workspace-write --thread-source shepherd_role`, and says Codex roles get only the sandbox, immediately before the new closure note. That contradicts both the same document's new claim that the isolation flags landed (lines 220-227) and the actual argv built in `src/codex-role-argv.ts:117-142`. Update the current-state table and the pre-closure wording so the document has one accurate account of the shipped posture.

2. **test/transient-agent-argv.test.ts:439:** The new test says the isolation block is “right after the sandbox pair,” but its own expected prefix puts `--thread-source shepherd_role` between the sandbox pair and that block (`test/transient-agent-argv.test.ts:345-357`); the implementation does likewise (`src/codex-role-argv.ts:120-132`). Correct the test name/comment (or the asserted placement) so the regression gate describes what it verifies.

## Verification

- The Codex branch in `buildTransientAgentArgv` is the only in-repo caller of `codexRoleArgv` and passes all role spawns through it (`src/transient-agent-argv.ts:235-249`; `src/codex-role-argv.ts:111-143`). The implementation unconditionally adds `--skip-git-repo-check`, `--ignore-user-config`, `--ignore-rules`, and the CLAUDE.md fallback configuration (`src/codex-role-argv.ts:117-142`).
- The four-kind test enumerates reviewer, doc, writer-ro, and writer-only (`test/transient-agent-argv.test.ts:35,439-453`) and asserts the new prefix, while the separate negative test excludes ephemeral and dangerous-bypass flags (`test/transient-agent-argv.test.ts:472-493`).
- The security documentation explicitly records MCP consequences for reviewer/doc and writer-ro/writer-only, plus the accepted residuals (`docs/sandbox-security.md:167-202`).

## Limitation

The Codex CLI flag semantics are external to this repository. Per the read-only review instruction, I did not invoke the CLI; this review verified the in-repository argv construction and documentation consistency only.

Tests were not run, per review instructions.

kai-osthoff (commented):
# Review

Decision: comment

The implementation centralizes the new Codex isolation argv in the only `codexRoleArgv` call path and preserves the optional model, effort, last-message, and trailing-prompt ordering. `src/transient-agent-argv.ts:235-249` delegates every Codex transient kind to that builder; `src/codex-role-argv.ts:117-143` emits the four isolation arguments before that optional tail. The direct tests cover every declared kind (`test/transient-agent-argv.test.ts:35`, `test/transient-agent-argv.test.ts:440-524`), including both `-c` values and the prohibited flags (`test/transient-agent-argv.test.ts:469-512`).

The required per-kind MCP posture is documented for reviewer/doc and writer-ro/writer-only in `src/codex-role-argv.ts:41-52` and `docs/sandbox-security.md:176-202`. The rollout dependency used to justify omitting `--ephemeral` is present in `src/codex-usage.ts:82-110` and `src/codex-activity.ts:407-435`.

The two prior blocking points are addressed: the current-state helper-role table and the role-isolation closure description now state the shipped argv in `docs/research/codex-sdk-evaluation.md:26-35` and `docs/research/codex-sdk-evaluation.md:212-226`; the test now explicitly describes and asserts the `sandbox → thread source → isolation` order in `test/transient-agent-argv.test.ts:341-358` and `test/transient-agent-argv.test.ts:440-454`.

Could not verify Codex CLI flag semantics against a live binary because this review was explicitly read-only and prohibited running commands; the repository's argv construction and tests were inspected only.

Findings:

- `docs/research/codex-sdk-evaluation.md:7-9` still lists `--ignore-user-config` among flags Shepherd is not using, and `docs/research/codex-sdk-evaluation.md:120-129` still says config overrides are used for effort only. Both conflict with the current helper-role argv at `docs/research/codex-sdk-evaluation.md:30` and `src/codex-role-argv.ts:128-136`. This is non-blocking documentation drift.


**Nits (non-blocking):**

- [compliance] docs/research/codex-sdk-evaluation.md: The introduction and config-override table still say --ignore-user-config is unused and only effort uses -c, contradicting the shipped helper-role argv.

PR #2207, opened 2026-09-08
fix(settings): label unknown token scopes as no access

### LINKED ISSUE #2090: Settings → Access labels an unknown token scope as "Full" when it actually grants nothing
Follow-up to #2083 / PR #2089 (merged). Raised by the PR critic after the merge had already landed.

## Problem

`ui/src/lib/components/settings/SettingsAccessPanel.svelte` → `scopeLabel()` is a ternary chain whose final arm is `full`:

```ts
const scopeLabel = (s: TokenScope) =>
  s === "read" ? … : s === "submit" ? … : m.settings_access_scope_full();
```

So **any** value outside the three levels renders as **Full** — the loudest, most-privileged label.

That state is reachable. `AccessTokenSummary.scope` is *typed* as the union, but SQLite holds plain `TEXT`, and both `toSummary()` (`src/access-tokens.ts`) and `verify()` pass the stored value through **verbatim on purpose**: laundering a bad scope into a plausible one would kill the deny-by-default rule that `scopeAllows()` depends on (locked by `test/token-scopes.test.ts` → "an unrecognized stored scope grants nothing" and `test/access-tokens.test.ts` → "reaches the gate VERBATIM, not normalized"). A hand-edited or future-version row with `scope = 'wat'` therefore arrives at the UI as `scope: "wat"`.

Result: the row's badge reads **Full** while the server grants that token **no route at all** — 403 even on `GET /api/sessions`. The badge is the audit surface for "what can this credential do", so it states the exact opposite of the truth.

Not an auth bypass — enforcement is correct and fails closed. This is a display/audit defect, and it overstates privilege rather than understating it.

## Fix

1. Give `scopeLabel()` an explicit `full` arm and a distinct **unknown** fallback, with a new `settings_access_scope_unknown` key in **both** `ui/messages/en.json` and `de.json` (e.g. EN "Unknown — no access", DE "Unbekannt — kein Zugriff"). Widen the parameter to accept a non-union value at this boundary, since that is what the wire can actually carry.
2. Add a `SettingsAccessPanel.browser.test.ts` case rendering a token with an out-of-union scope and asserting the badge does **not** read `Full`.

## Keep it minimal — two things that are already correct

- `scopeHint()` has the same ternary shape but is only ever called with the picker's own `scope` state, which is always one of the three. It does **not** need changing.
- The badge *styling* is already right: `class:full={t.scope === "full"}` means an unknown scope renders with the neutral `.scope-badge`, not the amber `.full` variant. Only the label text is wrong.

kai-osthoff:
## Summary

A stored scope such as `wat` currently appears as **Full** in Settings → Access even though the server grants it no access. Give `scopeLabel()` a `string` boundary, an explicit `full` branch, and the localized fallback **Unknown — no access** / **Unbekannt — kein Zugriff**.

Add a browser regression test that checks the token's actual badge. Scope hints, badge styling, and server enforcement remain unchanged.

Closes #2090

## Validation

- Regression reproduced before the fix (`Full` badge); all 18 panel tests pass after it.
- `bun run lint`, UI `bun run check`, `bun run check:i18n`, and full UI tests pass (300 files, 4,790 tests).
- Full pre-push gate passes after rebasing onto `origin/main`, including root tests (9,149 passed, 16 skipped), typechecks, UI/extension tests and builds, and the delta audit.
- Self-reviewed the four-file diff for correctness, security, and scope; no additional findings.

## Checklist

- [x] Conventional-commit title and linear branch rebased onto `main`.
- [x] Relevant local checks and tests pass.
- [x] User-facing text is translated in both UI catalogs.
- [x] Existing badge design is preserved.
- [x] Bug fix only; no new feature announcement, public API/config change, or documentation update needed.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3594,6 +3594,7 @@
   "settings_access_scope_read": "Lesen",
   "settings_access_scope_submit": "Einreichen",
   "settings_access_scope_full": "Vollzugriff",
+  "settings_access_scope_unknown": "Unbekannt — kein Zugriff",
   "settings_access_scope_read_hint": "Darf die Herde lesen — Sitzungen, Wartepunkte, Git-Status — und live mitverfolgen. Kann keine Arbeit starten und kein Terminal erreichen.",
   "settings_access_scope_submit_hint": "Alles wie bei „Lesen“, dazu Arbeit einreichen: eine Sitzung starten, eine wartende Aufgabe einstellen oder freigeben, einen Screenshot anhängen, ein Issue anlegen. Kann eine laufende Sitzung nicht steuern und kein Terminal erreichen.",
   "settings_access_scope_full_hint": "Derselbe Zugriff wie du, das laufende Terminal eingeschlossen — damit lässt sich in jeden laufenden Agenten tippen. Wähle das nur für einen Client, der Arbeit wirklich steuern muss.",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -3594,6 +3594,7 @@
   "settings_access_scope_read": "Read",
   "settings_access_scope_submit": "Submit",
   "settings_access_scope_full": "Full",
+  "settings_access_scope_unknown": "Unknown — no access",
   "settings_access_scope_read_hint": "Can read the herd — sessions, holds, git status — and follow live updates. Cannot start work, and cannot reach a terminal.",
   "settings_access_scope_submit_hint": "Everything Read can do, plus handing work in: start a session, queue or release a held task, attach a screenshot, file an issue. Cannot steer a running session or reach a terminal.",
   "settings_access_scope_full_hint": "The same access you have, the live terminal included — it can type into any running agent. Pick this only for a client that genuinely needs to drive work.",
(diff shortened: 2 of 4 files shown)


kai-osthoff (commented):
# Review

No blocking findings.

Verified the component now assigns the Full label only to the exact `full` value; `read` and `submit` retain their existing branches, and every other string reaches the unknown/no-access message. The token-list badge invokes this formatter with the API token scope and applies the privileged styling only when that scope is exactly `full` (`ui/src/lib/components/settings/SettingsAccessPanel.svelte:35`, `ui/src/lib/components/settings/SettingsAccessPanel.svelte:249`).

Verified the regression test supplies an out-of-union wire value, waits for the list row, scopes its assertion to the rendered token badge, and requires `Unknown — no access` rather than merely excluding `Full` (`ui/src/lib/components/settings/SettingsAccessPanel.browser.test.ts:231`).

Verified locale parity: the configured catalog set contains only `en` and `de` (`ui/project.inlang/settings.json:4`), both use the configured `messages/{locale}.json` pattern (`ui/project.inlang/settings.json:10`), and both define `settings_access_scope_unknown` (`ui/messages/en.json:3597`, `ui/messages/de.json:3597`).

Verified the UI boundary is reachable despite its narrower declared type: the server summary preserves `row.scope` verbatim (`src/access-tokens.ts:63`), verification returns the stored scope (`src/access-tokens.ts:242`), and authorization rejects every scope other than `full`, `read`, or `submit` (`src/token-scopes.ts:102`).

No tests or builds were run, per the review instruction.

PR #2208, opened 2026-09-08
feat(recap): per-block trigger heuristics + smallest-view rules (#2194)

### LINKED ISSUE #2194: Recap prompt: add per-block trigger heuristics + smallest-view rules (mermaid/wireframe fire 0/88 today)
Handoff spec. The research is done and the verdict is settled — this issue is the instructions, not the investigation. Everything needed to execute is below; no need to re-derive it.

## Background (settled, do not relitigate)

Question was whether to adopt [`show-me`](https://github.com/humanlayer/skills/blob/main/plugins/show-me/skills/show-me/SKILL.md) from `humanlayer/skills` for the recap function.

**Do not install the plugin.** `show-me` is an interactive `/show-me` chat skill and both of its outputs are incompatible with recap: freeform markdown fences in a conversation (recap needs typed `VisualBlock[]` in `.shepherd-recap-blocks.json`, server-joined, Svelte-rendered, persisted past archive) and `Bash(open …html)` (the recap spawn is headless and read-only, and `WireframeBlock`'s sanitizer exists precisely to reject arbitrary HTML). The recap spawn never invokes skills by name either.

What IS worth taking is its editorial layer — "which view answers which question", and "pick the smallest view that makes the key point clear". `buildRecapPrompt` currently ships 12 block **schemas** and near-zero guidance on **when to reach for which**.

### Baseline (measured, `~/.shepherd/shepherd.db`, 139 `ready` recaps)

| metric | value |
| --- | --- |
| recaps emitting blocks | 88 / 139 (63%) |
| blocks per recap | median 6, max 14, mean 6.0 |
| `callout` / `diff` | 150 / 139 — 55% of all 525 blocks |
| `file-tree` / `checklist` / `rich-text` | 83 / 58 / 57 |
| `code` / `table` / `annotated-code` / `api-endpoint` / `data-model` | 15 / 12 / 5 / 4 / 2 |
| **`mermaid`** | **0** |
| **`wireframe`** | **0** |
| `body` length when blocks also present | median 1730 chars |

Two failures visible here: (1) the two block types that make a change legible at a glance never fire once, in a UI-heavy repo — and this is the **prompt**, not the pipeline, because `mermaid` *does* fire on the plan-gate side (2 blocks across 4 `plan_gates`, same parser, same renderer); (2) a median 1730-char `body` **plus** 6 blocks means prose and blocks are authored as two disjoint artifacts restating each other.

## Task 1 — trigger clauses in `buildRecapPrompt` (`src/recap-core.ts`)

The block-guidance region is lines **205-237** (`"Optionally, add \`blocks\`…"` through the end of `"Rules for blocks:"`). **Keep every existing schema string byte-identical** — they are the parse contract for `parseVisualBlocks`. Append a trigger clause to each, and insert the choosing-a-view preamble.

Use these clauses verbatim (edit for house voice, not for substance):

- **rich-text** — the connective tissue. Reach for it when the point is *why*, not *what*. Also the home for a **shape sketch**: when the change is structural and no single file carries it, put a fenced ` ```diff ` of the *shape* here (component tree, file layout, call tree, control-flow pseudocode — before/after), not a file diff.
- **callout** — reach for it when a reader who skims everything else must still see this one thing: a decision made, a risk taken, an assumption that may be wrong. Three callouts in a row is a list, not a callout.
- **file-tree** — reach for it when the footprint itself is the story: a change spread wide, or one that moves/renames. Skip it when ≤3 files changed — the diff blocks already show them.
- **diff** — reach for it when one specific changed file carries a load-bearing decision. Curated: 1-3 per recap, never "every file I touched".
- **code** — reach for it when a new file is short and self-explaining, and reading it beats describing it.
- **annotated-code** — same, but when the file needs a guided read: a non-obvious ordering, a subtle guard, a rule that isn't local.
- **data-model** — reach for it when the schema changed and the shape of the data is what a reviewer must check.
- **api-endpoint** — reach for it when a route was added, changed or deprecated and the contract is what a caller must know.
- **table** — reach for it when the point is a comparison across fixed axes (before/after, option A vs B, per-environment behaviour). Not for prose in a grid.
- **checklist** — reach for it when the operator has something to *do* or *verify*. If it only restates `openItems`, drop it.
- **mermaid** — reach for it when control or data flow **changed**, and the change is *between* components rather than inside one. A sequence diagram of the new path beats three paragraphs. Never diagram architecture that did not change.
- **wireframe** — reach for it when the change is visible on screen. The operator cannot see the UI from a diff; a mockup of the resulting screen is the only block that answers "what will I actually see?". Use it for any user-visible layout, state or copy change.

And the preamble, replacing/extending the current `"Optionally, add \`blocks\`…"` lines:

- Pick the smallest view that makes the point. A sentence beats a block; a block beats three paragraphs.
- `body` and `blocks` are **one document**, read top to bottom — not two artifacts. Write the short text that sets each visual up; never restate in `body` a point a block already makes.
- A good recap is typically 3-6 blocks. Past ~10 you are cataloguing the session instead of explaining it.
- You will not use most of these types in any given recap. Using a type because it exists is the failure mode.

## Task 2 — `pre` styling in `ui/src/lib/components/blocks/RichTextBlock.svelte`

The component renders markdown through `marked` + `DOMPurify` and styles `:global(code)` but has **no `:global(pre)` rule** — fenced code inside a rich-text block renders unstyled today. Task 1's shape-sketch guidance is worthless until this lands, so do it in the same PR. Match the existing block-renderer visual language; add a browser test alongside the 14 siblings in that directory.

## Constraints

- **i18n.** `buildRecapPrompt` branches on `operatorLanguage === "de"` and appends `visualBlockLanguageLine(lang)` (`src/operator-language.ts:145`). New guidance must survive that branch: trigger clauses are English instructions *about* authoring and are never themselves translated, but any block field they tell the agent to write must stay consistent with `VISUAL_BLOCK_TRANSLATE_FIELDS` / `VISUAL_BLOCK_VERBATIM_FIELDS`. Do not add a second, competing language instruction.
- **Prompt budget.** Lines 205-237 are currently **4205 bytes**, shipped on every recap spawn. Keep the whole block-guidance region **under ~7 KB**. One clause per type, one line each — if a clause needs two lines it is over-specified.
- **Scope.** No new block type. No changes to `parseVisualBlocks`, the sidecar contract, `joinDiffBlocks`, or any renderer other than Task 2's `pre` rule.
- **Do not weaken the existing safety rules** in the `"Rules for blocks:"` list — grounding to real changed paths, secret redaction, and the wireframe authoring restrictions (no inline hex/`rgb()`/`hsl()`/`color()`/`font-family`/`box-shadow`, no `<script>`/`<style>`/handlers/`href`) all stay exactly as they are. Task 1 makes `wireframe` fire *more*, so those constraints get load-bearing for the first time.

## Acceptance

1. `bun run test` (root) and `cd ui && bun run test` green; `bun run lint` and `cd ui && bun run check` clean.
2. **Empirical check — required, this is the whole point.** Pick ≥8 archived sessions with a mix of UI and server changes, `POST /api/sessions/:id/recap/regenerate` (route at `src/server.ts:3084`), and compare the resulting block mix against the baseline table above. Report the new histogram in the PR.
   - **Must move:** `wireframe` > 0 on the UI-change sessions. That is the single clearest signal the guidance took.
   - **Should move:** `mermaid` > 0 on at least one session with a genuine flow change.
   - **Must NOT regress:** median blocks/recap stays in 3-7 and the max does not climb past the current 14. More blocks is the *opposite* failure — if the tail grows, the smallest-view rules are too weak relative to the triggers and need tightening before merge.
   - **Should move:** median `body` length drops from 1730 as prose stops duplicating blocks.
3. Spot-read two regenerated recaps in the UI. Blocks should read as one document with the prose, not as an appendix.

## Risk

This is a hot, well-tuned prompt with real scar tissue (#2045, #822). The smallest-view rules and the trigger clauses must land **together** — shipping triggers alone predictably produces kitchen-sink recaps, which is worse than the current under-use.

scoop:
Closes #2194.

`buildRecapPrompt` shipped 12 block **schemas** and near-zero guidance on **when to reach for which**. This adds the editorial layer, plus the `pre` rendering the new shape-sketch guidance depends on.

## What changed

- **`src/recap-core.ts`** — a "Choosing a view" preamble (the smallest-view brakes) and one trigger clause appended to each of the 12 schema lines. **Every existing schema string remains a byte-identical prefix** — they are the parse contract for `parseVisualBlocks` — asserted by test.
- **`ui/.../RichTextBlock.svelte`** — themed `:global(pre)` matching `CodeBlock`'s surface, plus per-line tinting for ```` ```diff ```` fences using the same `color-mix` recipe as `DiffFileBlock`. Lines are rebuilt from `textContent` on nodes we create, and DOMPurify now returns a **fragment** so no code path assigns `innerHTML`.
- Tests pin the schema strings, the trigger coverage, the brakes, the untouched safety rules, and a byte-budget ceiling (marker-sliced, not line-numbered).

## Empirical result

Same 10 archived sessions, regenerated through the real `RecapService` with this branch's prompt (the live server runs `main`, so its regenerate route would have exercised the old one).

| metric | before | after |
| --- | --- | --- |
| blocks total | 107 | 54 |
| blocks/recap median | 10 | **5.5** (target 3-7) |
| blocks/recap max | 14 | **7** (must not exceed 14) |
| zero-block recaps | 0 | **0** |
| `mermaid` | 0 | **1** |
| `wireframe` | 0 | **0** |
| `body` median | 2415 | 2210 |

Per-type after: `diff`=16 `callout`=13 `file-tree`=10 `checklist`=8 `rich-text`=6 `mermaid`=1.

Two regenerated recaps were spot-read in the Done lens. They read as one document — a two-sentence prose intro that sets up the visuals, then the diff block, footprint and callout, with no prose restating what a block already shows.

## `wireframe` did not move — and cannot from the prompt alone

This is the issue's one **must-move** criterion, and it is not met. Three prompt variants across 30 regenerations left it at zero, including one that states `wireframe` is *not optional* for a visible change and supplies the helper-class vocabulary the prompt had been naming while defining it nowhere (`wf-card`/`wf-box`/`wf-pill`/`wf-chip`/`wf-muted` live only in `WireframeBlock.svelte`).

The blocker is structural:

- the recap spawn runs the `writer-only` preset — `allowedTools: ["Write"]`, no Read/Grep/Glob, disposable temp cwd, no `--add-dir`;
- `buildRecapPrompt` carries the task prompt, plan, changed **file paths + statuses**, and the transcript digest — **no diff content** (by design: the server attaches diff bodies to `diff` blocks *after* the agent writes them).

So the agent has never seen a line of the UI it is asked to mock up, and the prompt's own grounding rule — "Never invent a path, a field, or a change", which #2194 said not to weaken — correctly stops it from guessing. The asymmetry is the evidence: identical guidance moved `mermaid` (derivable from the digest and file names) and cannot move `wireframe`. #2194 inferred "prompt, not pipeline" from mermaid firing on the plan-gate side, but that agent has full repo Read access, so the comparison does not transfer.

Follow-up filed: #2209. Fixing it means giving the recap agent sight of the UI, which is the pipeline change #2194 explicitly scoped out.

## Caveats

- **n=10, one sample per session.** The categorical results are robust to that; the medians are not distributions.
- **The A/B is not perfectly matched.** `.shepherd-plan.md` is never committed, so regenerated recaps ran without plan text while the stored ones had it. The new run saw *less* input, which biases against richer blocks.
- The budget ceiling is 7.25 KiB rather than the issue's ~7 KB; the difference buys the wireframe helper-class list.
- The 10 sessions' live recap rows were replaced with their regenerated versions so the Done lens could be spot-read. A pre-change backup of the live DB was taken first and is retained in this session's scratchpad.

### DIFF
--- a/src/recap-core.ts
+++ b/src/recap-core.ts
@@ -203,25 +203,54 @@ export function buildRecapPrompt(input: {
     "- openItems: string[] of anything left to do or worth noting for the next session ([] if none)",
     "",
     "Optionally, add `blocks` to render the recap as a scannable visual document instead",
-    "of plain markdown. Omit `blocks` entirely if plain `body` suffices — blocks are not required.",
+    // #2194: this line used to end "Omit `blocks` entirely if plain `body` suffices — blocks are not
+    // required", which the model weighed over every later rule: regenerations kept coming back with
+    // ZERO blocks on 9-to-74-file sessions. The escape hatch stays for the genuinely empty session
+    // (and a Codex recap, which answers in chat and writes no sidecar at all, still degrades to
+    // body-only mechanically) — it just no longer reads as a standing invitation.
+    "of plain markdown. For a session that changed code, blocks are how the operator reads it —",
+    "omit them only when there is genuinely nothing to show.",
+    "",
+    // #2194: the schemas below are the parse contract; these five rules are the BRAKES on the
+    // per-type triggers appended to each. Triggers without brakes produce kitchen-sink recaps —
+    // a worse failure than the under-use they fix. Never ship one half of this pair.
+    "Choosing a view — each schema below ends with when to reach for that type. If no trigger",
+    "fires, do not emit that type. Five rules govern all of them:",
+    "- Pick the smallest view that ANSWERS THE READER'S QUESTION. Smallest means most economical to",
+    "  READ, not cheapest to write: never skip the view a point needs because authoring it is more",
+    "  work. A sentence beats a block only when the sentence actually answers the question.",
+    "- `body` and `blocks` are ONE document, read top to bottom — not two artifacts. Write the short",
+    "  text that sets each visual up; never restate in `body` a point a block already makes.",
+    "- A good recap is typically 4-7 blocks. Past ~10 you are cataloguing the session instead of",
+    "  explaining it; at 0-1 you wrote prose and called it a recap.",
+    "- Reach for a type when its trigger fires — not because it exists, nor because you reached for",
+    "  it last time. `callout`/`diff`/`file-tree` are not the default three.",
+    // #2194: wireframe and mermaid had fired 0 times across 88 block-emitting recaps. Both are
+    // expensive to author, so every "keep it small" rule argues against them unless their trigger
+    // is stated as non-optional HERE — in the preamble that gets read — rather than only at the end
+    // of their own (long, restriction-heavy) schema lines.
+    "- If the session changed what a user sees on screen, `wireframe` is NOT optional: no diff can",
+    "  show the operator the resulting screen. If it changed control or data flow BETWEEN components,",
+    "  `mermaid` is not optional either. These two are under-used, not over-used — when their trigger",
+    "  fires, emit them.",
     "",
     "Block types (Phase 1):",
-    '- rich-text: {"type":"rich-text","id":"<unique>","markdown":"<prose>"} — narrative / the why.',
-    '- callout:   {"type":"callout","id":"...","tone":"info|decision|risk|warning|success","markdown":"..."} — a toned note for a decision, risk, or assumption.',
-    '- file-tree: {"type":"file-tree","id":"...","title?":"...","entries":[{"path":"<real path>","change":"added|modified|removed|renamed","note?":"<short>"}]} — the change footprint. Use real changed paths only.',
-    '- diff:      {"type":"diff","id":"...","path":"<real changed path>","summary":"<one line>","annotations?":[{"label?":"<short>","note":"<prose>"}]} — feature a specific changed file.',
+    '- rich-text: {"type":"rich-text","id":"<unique>","markdown":"<prose>"} — narrative / the why. Reach for it when the point is WHY, not what. Also the home for a shape sketch: when the change is structural and no single file carries it, put a fenced ```diff of the SHAPE here (component tree, file layout, call tree, control-flow pseudocode — before/after), not a file diff.',
+    '- callout:   {"type":"callout","id":"...","tone":"info|decision|risk|warning|success","markdown":"..."} — a toned note for a decision, risk, or assumption. Reach for it when a reader who skims everything else must still see this one thing. Three callouts in a row is a list, not a callout.',
+    '- file-tree: {"type":"file-tree","id":"...","title?":"...","entries":[{"path":"<real path>","change":"added|modified|removed|renamed","note?":"<short>"}]} — the change footprint. Use real changed paths only. Reach for it when the footprint itself is the story: a change spread wide, or one that moves/renames. Skip it when ≤3 files changed — the diff blocks already show them.',
+    '- diff:      {"type":"diff","id":"...","path":"<real changed path>","summary":"<one line>","annotations?":[{"label?":"<short>","note":"<prose>"}]} — feature a specific changed file. Reach for it when one specific changed file carries a load-bearing decision. Curated: 1-3 per recap, never every file you touched.',
     "",
     "Block types (Phase 2):",
-    '- code:           {"type":"code","id":"...","filename":"<path marked (added)>"} — highlights a new file (language is derived from the path). Only emit for files marked (added) above; modified files use diff blocks. Never type the code body — the server attaches it.',
-    '- annotated-code: {"type":"annotated-code","id":"...","filename":"<path marked (added)>","annotations?":[{"label?":"<short>","note":"<prose describing this part of the code>"}]} — code with prose notes. Only (added) files. Never type the code body, never line numbers.',
-    '- data-model:     {"type":"data-model","id":"...","entities":[{"id":"...","name":"...","fields":[{"name":"...","type":"...","pk?":true,"fk?":"<ref>","nullable?":true,"change?":"added|modified|removed|renamed","was?":"<old type>"}]}],"relations?":[{"from":"...","to":"...","kind":"..."}]} — ERD-ish card. Extract from real changed schema; do not invent fields; redact secrets. Will be tagged inferred automatically.',
-    '- api-endpoint:   {"type":"api-endpoint","id":"...","method":"GET|POST|...","path":"<route>","summary?":"...","change?":"added|modified|deprecated","deprecated?":true,"params?":[{"name":"...","in":"path|query|body","type":"...","required?":true,"note?":"..."}],"responses?":[{"status":200,"description?":"...","example?":"..."}]} — one route card. Extract from real changed routes; redact secrets. Will be tagged inferred automatically.',
-    '- table:          {"type":"table","id":"...","columns":["A","B"],"rows":[["a","b"]]} — columnar comparison or summary. Redact secrets.',
-    '- checklist:      {"type":"checklist","id":"...","items":[{"id":"...","label":"...","note?":"...","checked?":true}]} — task list or review checklist.',
+    '- code:           {"type":"code","id":"...","filename":"<path marked (added)>"} — highlights a new file (language is derived from the path). Only emit for files marked (added) above; modified files use diff blocks. Never type the code body — the server attaches it. Reach for it when a new file is short and self-explaining, and reading it beats describing it.',
+    '- annotated-code: {"type":"annotated-code","id":"...","filename":"<path marked (added)>","annotations?":[{"label?":"<short>","note":"<prose describing this part of the code>"}]} — code with prose notes. Only (added) files. Never type the code body, never line numbers. Reach for it when such a file instead needs a guided read: a non-obvious ordering, a subtle guard, a rule that is not local.',
+    '- data-model:     {"type":"data-model","id":"...","entities":[{"id":"...","name":"...","fields":[{"name":"...","type":"...","pk?":true,"fk?":"<ref>","nullable?":true,"change?":"added|modified|removed|renamed","was?":"<old type>"}]}],"relations?":[{"from":"...","to":"...","kind":"..."}]} — ERD-ish card. Extract from real changed schema; do not invent fields; redact secrets. Will be tagged inferred automatically. Reach for it when the schema changed and the shape of the data is what a reviewer must check.',
+    '- api-endpoint:   {"type":"api-endpoint","id":"...","method":"GET|POST|...","path":"<route>","summary?":"...","change?":"added|modified|deprecated","deprecated?":true,"params?":[{"name":"...","in":"path|query|body","type":"...","required?":true,"note?":"..."}],"responses?":[{"status":200,"description?":"...","example?":"..."}]} — one route card. Extract from real changed routes; redact secrets. Will be tagged inferred automatically. Reach for it when a route was added, changed or deprecated and the contract is what a caller must know.',
+    '- table:          {"type":"table","id":"...","columns":["A","B"],"rows":[["a","b"]]} — columnar comparison or summary. Redact secrets. Reach for it when the point is a comparison across fixed axes (before/after, option A vs B, per-environment behaviour). Not for prose in a grid.',
+    '- checklist:      {"type":"checklist","id":"...","items":[{"id":"...","label":"...","note?":"...","checked?":true}]} — task list or review checklist. Reach for it when the operator has something to DO or verify. If it only restates openItems, drop it.',
     "",
     "Block types (Phase 3):",
-    '- mermaid:        {"type":"mermaid","id":"...","source":"<mermaid diagram source>","caption?":"..."} — an architecture or flow diagram (flowchart/sequence/etc). Use for genuine architecture/flow shifts only. Will be tagged inferred automatically.',
-    '- wireframe:      {"type":"wireframe","id":"...","surface":"browser|desktop|mobile|popover|panel","html":"<themed HTML mockup>","caption?":"..."} — a UI mockup of a screen. Use ONLY for UI changes. Author with the wf helper classes + class-based color; NEVER inline hex/rgb()/hsl()/color()/font-family/box-shadow, and never <script>/<style>/event handlers/href.',
+    '- mermaid:        {"type":"mermaid","id":"...","source":"<mermaid diagram source>","caption?":"..."} — an architecture or flow diagram (flowchart/sequence/etc). Use for genuine architecture/flow shifts only. Will be tagged inferred automatically. Reach for it when control or data flow CHANGED, and the change is BETWEEN components rather than inside one — a sequence diagram of the new path beats three paragraphs. Never diagram architecture that did not change.',
+    '- wireframe:      {"type":"wireframe","id":"...","surface":"browser|desktop|mobile|popover|panel","html":"<themed HTML mockup>","caption?":"..."} — a UI mockup of a screen. Use ONLY for UI changes. Author with the wf helper classes + class-based color; NEVER inline hex/rgb()/hsl()/color()/font-family/box-shadow, and never <script>/<style>/event handlers/href. Reach for it when the change is visible on screen: the operator cannot see the UI from a diff, and a mockup of the resulting screen is the only block that answers what they will actually see. Use it for any user-visible layout, state or copy change. The helper classes are: wf-card, wf-box, wf-pill, wf-chip, wf-muted — plus plain <button> (class="primary" or data-primary for the accent one) and data-icon for a glyph. Compose them with div/span/ul/li/table/h1-h6/header/nav/section/strong/small/img/svg.',
     "",
     "Rules for blocks:",
     "- Every block must have a unique string `id`.",
--- a/test/recap-core.test.ts
+++ b/test/recap-core.test.ts
@@ -967,3 +967,104 @@ test("#2045 splice helper: null means no contribution; non-object shapes pass th
   expect(spliceRecapSidecars("not an object", "x", null)).toBe("not an object");
   expect(spliceRecapSidecars(null, "x", null)).toBeNull();
 });
+
+// ── #2194: per-block trigger heuristics + smallest-view rules ─────────────────
+//
+// The block-guidance region is a hot, shipped-every-spawn prompt. Three things must hold
+// together: the schemas stay the parse contract, every type says when to reach for it, and
+// the brakes that stop the triggers producing kitchen-sink recaps stay next to them.
+
+/** The prompt built with the minimum inputs — the guidance region is input-independent. */
+const bareRecapPrompt = (): string =>
+  buildRecapPrompt({ taskPrompt: "t", plan: "", changedFiles: [], digest: "", context: "" });
+
+/**
+ * The block-guidance region, sliced by MARKER rather than by line number so the assertion
+ * survives the next edit to the prompt around it.
+ */
+function blockGuidanceRegion(prompt: string): string {
+  const from = prompt.indexOf("Optionally, add `blocks`");
+  const to = prompt.indexOf("When done, write these files in your CWD:");
+  expect(from).toBeGreaterThan(-1);
+  expect(to).toBeGreaterThan(from);
+  return prompt.slice(from, to);
+}
+
+/**
+ * The 12 schema strings EXACTLY as they are parsed against. #2194 appended a trigger clause
+ * to each line; the schema itself must remain a byte-identical prefix, because these strings
+ * are what teaches the agent the shape `parseVisualBlocks` accepts. A rewrite here silently
+ * changes what gets emitted — and a silently dropped block is invisible in the UI.
+ */
+const BLOCK_SCHEMAS: readonly string[] = [
+  '- rich-text: {"type":"rich-text","id":"<unique>","markdown":"<prose>"}',
+  '- callout:   {"type":"callout","id":"...","tone":"info|decision|risk|warning|success","markdown":"..."}',
+  '- file-tree: {"type":"file-tree","id":"...","title?":"...","entries":[{"path":"<real path>","change":"added|modified|removed|renamed","note?":"<short>"}',
+  '- diff:      {"type":"diff","id":"...","path":"<real changed path>","summary":"<one line>","annotations?":[{"label?":"<short>","note":"<prose>"}',
+  '- code:           {"type":"code","id":"...","filename":"<path marked (added)>"}',
+  '- annotated-code: {"type":"annotated-code","id":"...","filename":"<path marked (added)>","annotations?":[{"label?":"<short>","note":"<prose describing this part of the code>"}',
+  '- data-model:     {"type":"data-model","id":"...","entities":[{"id":"...","name":"...","fields":[{"name":"...","type":"...","pk?":true,"fk?":"<ref>","nullable?":true,"change?":"added|modified|removed|renamed","was?":"<old type>"}',
+  '- api-endpoint:   {"type":"api-endpoint","id":"...","method":"GET|POST|...","path":"<route>","summary?":"...","change?":"added|modified|deprecated","deprecated?":true,"params?":[{"name":"...","in":"path|query|body","type":"...","required?":true,"note?":"..."}',
+  '- table:          {"type":"table","id":"...","columns":["A","B"],"rows":[["a","b"]]}',
+  '- checklist:      {"type":"checklist","id":"...","items":[{"id":"...","label":"...","note?":"...","checked?":true}',
+  '- mermaid:        {"type":"mermaid","id":"...","source":"<mermaid diagram source>","caption?":"..."}',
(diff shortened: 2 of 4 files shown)


scoop (commented):
# Review — feat(recap): per-block trigger heuristics + smallest-view rules (#2194)

Diff reviewed: `98b4faa1...HEAD` — 4 files (`src/recap-core.ts`, `test/recap-core.test.ts`,
`ui/src/lib/components/blocks/RichTextBlock.svelte`, `.../RichTextBlock.browser.test.ts`).

## What I verified

**Schema-prefix contract (Task 1's hard constraint).** Compared each of the 12 schema strings in
`test/recap-core.test.ts:996-1009` against the shipped lines at `src/recap-core.ts:238-241`,
`244-249` and `252-253`. Every test string is a genuine byte-identical prefix of the shipped line
(the test truncates before the closing `]}` on the nested-array ones, e.g. `file-tree` ends at
`"note?":"<short>"}` and `src/recap-core.ts:240` continues `]} — the change footprint…`). No
existing wording was rewritten, reordered or deleted; every clause is appended. The `parseVisualBlocks`
contract in `src/visual-blocks.ts` is untouched.

**Trigger clause per type.** All 12 lines carry `Reach for it when` on the same line as their schema
(`src/recap-core.ts:238,239,240,241,244,245,246,247,248,249,252,253`), so the
`expect(line).toContain("Reach for it when")` assertion at `test/recap-core.test.ts:1024` holds for
each. The preamble at `src/recap-core.ts:217-235` announces "Five rules" and lists exactly five
bullets (219, 222, 224, 226, 232) — count matches.

**Safety rules not weakened.** `src/recap-core.ts:255-265` is byte-identical to the pre-diff text:
grounding to real changed paths (257-258), the diff-block restriction (259-261), secret redaction
(264-265). The `wireframe` authoring restrictions survive verbatim inside line 253
(`NEVER inline hex/rgb()/hsl()/color()/font-family/box-shadow, and never <script>/<style>/event
handlers/href`), which is what `src/visual-blocks.ts:452-457` and `461-467` actually enforce
(`href\s*=` anywhere in the html drops the whole block at `src/visual-blocks.ts:502`).

**The new wireframe helper-class list is accurate.** `src/recap-core.ts:253` names `wf-card, wf-box,
wf-pill, wf-chip, wf-muted` — all five are defined in `HELPER_CSS` at
`ui/src/lib/components/blocks/WireframeBlock.svelte:110,111,112,113,114`. `button` with
`class="primary"` or `data-primary` matches `WireframeBlock.svelte:115-116`; `data-icon` matches
`:117`. `data-primary` and `data-icon` are explicitly in `ALLOWED_ATTR`
(`WireframeBlock.svelte:59-60`), so `ALLOW_DATA_ATTR: false` (`:82`) does not strip them. Every
composition tag the line names — `div span ul li table h1-h6 header nav section strong small img
svg` — is in `ALLOWED_TAGS` (`WireframeBlock.svelte:14,15,17,19,33,20-25,26,28,29,41,45,48,49`).

**i18n.** `buildRecapPrompt`'s `de` branch (`src/recap-core.ts:290-299`) is untouched and no second
language instruction is added. The only field the new guidance newly steers content into is
`rich-text.markdown`, which is already in `VISUAL_BLOCK_TRANSLATE_FIELDS`
(`src/operator-language.ts:75`). `wireframe.html` stays in `VISUAL_BLOCK_VERBATIM_FIELDS`
(`src/operator-language.ts:111`) — unchanged, so no new inconsistency. No user-facing strings are
added anywhere in the diff, so no catalog keys are involved.

**Design tokens resolve.** Every token the new CSS reads exists on `:root` in `ui/src/app.css`:
`--color-inset` (`:12`), `--color-line` (`:15`), `--color-muted` (`:19`), `--color-green` (`:22`),
`--color-red` (`:23`), `--font-mono` (`:29`), `--fs-meta` (`:86`). The `@theme` block is processed in
the browser-test project too — `tailwindcss()` is a top-level plugin and both vitest projects use
`extends: true` (`ui/vite.config.ts:63,84,92`), and the spec already imports `app.css`
(`RichTextBlock.browser.test.ts:4`, pre-existing).

**Comment claims about sibling components are true.** "Same tint recipe as DiffFileBlock's
`.line.add` / `.line.del`" (`RichTextBlock.svelte:111-112`) matches
`ui/src/lib/components/DiffFileBlock.svelte:207-212` exactly
(`color-mix(in srgb, var(--color-green) 12%, transparent)`). "Matches CodeBlock.svelte's surface"
(`RichTextBlock.svelte:91`) matches `ui/src/lib/components/blocks/CodeBlock.svelte:62-68` and
`112-117` (inset background, `--color-line` border, 3px radius, `--font-mono`, `--fs-meta`,
line-height 1.5).

**The tint pass cannot reintroduce markup.** `tintDiffFences`
(`RichTextBlock.svelte:16-32`) only ever writes `textContent` on `span`s it created itself (`:27`)
and clears `code.textContent` (`:23`); no `innerHTML` is assigned inside it. Fenced content reaches
it already HTML-escaped by `marked`, so the `` case in the new test
(`RichTextBlock.browser.test.ts:90-102`) is inert text, not an element. The trailing-newline pop at
`:22` is correct — `marked` emits a trailing `\n` inside `<code>`, which would otherwise render as a
blank tinted row, and the test pins 4 spans for 4 lines (`:72`).

**`Buffer` in the new root test** is used un-imported, consistent with existing specs
(`test/argv-limit.test.ts:75`, `test/critic-core.test.ts:1638`) — Bun provides it as a global.

**No stale references to the replaced prompt line.** Grepped for `blocks are not required` /
`Omit \`blocks\` entirely` across the tree: the only hits are `src/recap-core.ts:206` (the new
explanatory comment) and `test/recap-core.test.ts:986` (the marker slice), so nothing else asserted
the old text.

## Findings

### 1. `RichTextBlock.svelte:13` — the JSDoc's security claim is wrong about what the code does

`tintDiffFences` returns `host.innerHTML` (`:31`), which is assigned to `rendered` (`:52`) and
rendered through `{@html rendered}` (`:67`). So the sanitizer's output **is** serialized and
re-parsed — exactly once, precisely as it was before this change, when
`DOMPurify.sanitize(...)` returned a string that `{@html}` parsed. The comment at `:13` asserts the
opposite: "The sanitizer's output is therefore never re-parsed."

The code is safe (the risk profile is unchanged, and the `textContent`-only claim at `:12` is
accurate for the function body), but the comment states a stronger invariant than the component
actually holds. That matters here specifically because it is the stated justification for the
sanitizer boundary: a future reader who believes the serialize→parse round trip was eliminated will
reason wrongly about what is safe to add. Say what is true — the pass only ever writes `textContent`
on nodes it created, so it cannot reintroduce markup DOMPurify removed, and the serialize→`{@html}`
round trip is the same one that shipped before.

## Nits (non-blocking):
- `src/recap-core.ts:271` still tells the agent the blocks sidecar is "OPTIONAL … Omit the file entirely when plain markdown suffices" — the same standing invitation the diff deliberately removed at `:205-212`, restated 60 lines later where the model reads it last.
- `src/recap-core.ts:253` lists `img` among the tags to compose a wireframe from, but `src` is in `FORBID_ATTR` (`ui/src/lib/components/blocks/WireframeBlock.svelte:81`) and the iframe CSP is `img-src data:` (`:162`), so an `` can never display anything — it will render as alt text at best.
- `src/recap-core.ts:253` names `svg` as composable without mentioning that only `path/rect/circle/line/g` children are allowed (`WireframeBlock.svelte:50-54`) and that the same "no raw hex/rgb()" rule applies to `fill`/`stroke` (enforced at `src/visual-blocks.ts:488-495`).
- `RichTextBlock.svelte:115,119` adds `color: var(--color-green)` / `var(--color-red)` on top of the 12% tint, which `DiffFileBlock.svelte:207-212` does not do — green-on-green-tint is lower contrast than the plain ink the sibling uses, despite the comment claiming the same recipe.
- `RichTextBlock.svelte:107-109` — `.dl { min-width: 100% }` resolves against the `pre` content box, so on a horizontally-scrolled sketch the tint stops at the initial viewport width.
- `src/recap-core.ts:224` says "typically 4-7 blocks"; the issue's verbatim clause said 3-6 and its acceptance band is 3-7.

- `src/recap-core.ts:253` — the wireframe helper-class/tag inventory (~350 bytes) is new prompt content neither the issue's verbatim clauses nor the approved plan described; it is defensible (the type was unauthorable without it) but it is what pushed the region budget past the issue's stated ~7 KB to the 7424-byte ceiling asserted at `test/recap-core.test.ts:1069`.
- `src/recap-core.ts:205-212` — the approved plan said the two "Optionally, add `blocks`… blocks are not required" lines stay; the diff rewrites the escape hatch into "For a session that changed code, blocks are how the operator reads it".
- `src/recap-core.ts:232-235` — the fifth preamble rule ("`wireframe` is NOT optional", "`mermaid` is not optional either") is a mandate the issue's clause set did not contain; combined with the weakened first rule at `:219-221` ("Smallest means most economical to READ, not cheapest to write") the brakes are materially softer and the triggers materially harder than the issue specified. The commit body reports the tail did not grow (median 10→5.5, max 14→7), which is the axis the issue's risk note cares about.

- **Acceptance #2's empirical result.** The histogram lives in the PR body, not the repo. The commit body reports "median blocks 10 -> 5.5, max 14 -> 7, no zero-block recaps, mermaid 0 -> 1" but does **not** report a `wireframe` count — and `wireframe > 0` on the UI-change sessions is the issue's single "Must move" gate ("the single clearest signal the guidance took"). `test/recap-core.test.ts:1046-1048` implies a second run rebalanced both types, so this may simply be an incomplete commit message. I am not raising it as a finding because I cannot see the run; please confirm the `wireframe` number is in the PR body's histogram.
- `node_modules` is not installed in this worktree, so I could not read the pinned `dompurify@^3.4.14` / `marked@^18.0.11` sources. The reasoning above about `RETURN_DOM_FRAGMENT` returning the sanitized fragment, `class` surviving DOMPurify's default attribute allowlist, and `marked` emitting `<pre><code class="language-diff">` for a ```diff fence is from the documented behaviour of those versions, not from the installed source. The new browser tests assert exactly these, so CI will settle them.
- I did not run tests or lint, per the review brief.

- `src/service.ts:1329-1330` — the plan-gate prompt still tells the agent to "Author with the wf helper classes" while defining them nowhere, which is the gap this PR closed on the recap side only.
- `docs/research/native-visual-plan-recap-design.md:233` documents `<span data-icon="mail">` as "renderer swaps a Tabler SVG", but `WireframeBlock.svelte:117` renders `[data-icon]` as a plain 1em grey square.
- `src/operator-language.ts:111` keeps `wireframe.html` verbatim, so a German operator's mockup copy stays English — a pre-existing trade-off that this PR makes more visible by driving `wireframe` to fire.

scoop (commented):
# Review — recap prompt trigger heuristics + `pre` styling (#2194), rework round 2

Diff reviewed: `git diff 98b4faa14e74d3505658c59959e3d61ceed6cf9b...HEAD` — 4 files
(`src/recap-core.ts`, `test/recap-core.test.ts`,
`ui/src/lib/components/blocks/RichTextBlock.svelte`, `…/RichTextBlock.browser.test.ts`).

## Prior finding from the last revision

**1. `RichTextBlock.svelte` JSDoc claimed the sanitizer output is never re-parsed — ADDRESSED.**
The comment now reads (RichTextBlock.svelte:11-15): "this pass cannot reintroduce markup DOMPurify
removed. It takes the sanitized DOM *fragment* rather than the HTML string, and every node it adds
is one it created itself and filled via `textContent` … The result is still serialized here and
re-parsed by `{@html}` below, exactly as before this pass existed; that round trip is unchanged."
I verified each half of that claim against the code:

- Takes a fragment, not a string: `RETURN_DOM_FRAGMENT: true` at RichTextBlock.svelte:54-56, and
  `tintDiffFences(fragment: DocumentFragment)` at :17.
- Every added node is self-created and text-filled: `document.createElement("span")` +
  `span.textContent = line` (:26-28) and `document.createTextNode("\n")` (:29); `code.textContent = ""`
  (:24) clears, never `innerHTML`. `span.className` comes only from `lineKind`, whose codomain is the
  fixed set `add|del|meta|ctx` (:36-41).
- The serialize/re-parse round trip is acknowledged rather than denied: `return host.innerHTML` (:32)
  → `rendered` (:53) → `{@html rendered}` (:68), and the pre-diff code did the same thing because
  DOMPurify's default string return is itself an `innerHTML` serialization.

The comment now states a property that actually holds. Not re-raised.

## What I verified in this revision

**Task 1 — `src/recap-core.ts:205-253`.** All 12 schema lines carry an appended `Reach for it when …`
clause and each original schema string survives as a prefix (compared line by line against the
pre-diff text in `git diff`). The preamble at :217-235 ships the brakes (smallest-view, one-document,
4-7 blocks, "not because it exists") together with the two under-use pushes. The `Rules for blocks:`
safety list at :255-265 (grounding to real changed paths, diff-block restrictions, secret redaction)
is byte-unchanged, and the wireframe authoring restrictions are intact inside the schema line at
:253 — the same restrictions the server enforces in `src/visual-blocks.ts:452-458` (`<script`,
`<style`, `on…=`, `href=` all reject the block) and :461-467 / :481-485 (hex / `rgb()` / `hsl()` /
`color()` / `font-family` / `box-shadow` in `style=`, and raw colors in `fill=`/`stroke=`). Nothing
in the diff weakens them.

**The new wireframe helper-class enumeration is accurate.** The prompt (src/recap-core.ts:253) names
`wf-card, wf-box, wf-pill, wf-chip, wf-muted`, `<button>` with `class="primary"` or `data-primary`,
and `data-icon`. Verified against `ui/src/lib/components/blocks/WireframeBlock.svelte:110-117`
(`HELPER_CSS` defines exactly those five classes plus `button`, `button.primary`, `[data-primary]`,
`[data-icon]`), and against the sanitizer allowlists: `class` and `data-primary`/`data-icon` are
explicit `ALLOWED_ATTR` entries (WireframeBlock.svelte:56-60), which DOMPurify honours despite
`ALLOW_DATA_ATTR: false` (:82) because explicit allowlist entries are checked independently of the
data-attr blanket rule. Every tag in the new compose list — `div span ul li table h1-h6 header nav
section strong small img svg` — appears in `ALLOWED_TAGS` (WireframeBlock.svelte:13-55). One caveat
on `img` under Nits.

**i18n.** The German branch is untouched (src/recap-core.ts:290-299) and no second language
instruction is added. The only field the new guidance newly steers content into is
`rich-text.markdown`, which is already in `VISUAL_BLOCK_TRANSLATE_FIELDS`
(src/operator-language.ts:75); `wireframe.html` stays in `VISUAL_BLOCK_VERBATIM_FIELDS`
(src/operator-language.ts:111), so the new wireframe push does not create a translate/verbatim
conflict. No user-facing strings and no catalog keys are touched by either file, so there is no
en/de parity surface in this diff.

**Task 2 — `RichTextBlock.svelte:91-124`.** Every token used resolves: `--color-inset`
(ui/src/app.css:12), `--color-line` (:15), `--color-muted` (:19), `--color-green` (:22),
`--color-red` (:23), `--font-mono` (:29), `--fs-meta` (:86). The comment's claim that the surface
matches `CodeBlock.svelte` checks out against CodeBlock.svelte:65 (`--color-inset`), :67 (3px
radius), :79-80 (`--font-mono` / `--fs-meta`), :93 (`overflow-x: auto`). The tint recipe claim checks
out against DiffFileBlock.svelte:207-211 — `color-mix(in srgb, var(--color-green) 12%, transparent)`
and the `--color-red` counterpart, verbatim.

**The tint pass itself.** Selector `pre > code.language-diff` (RichTextBlock.svelte:20) matches what
`marked` emits for a ` ```diff ` fence and what DOMPurify keeps (`class` is default-allowed), which
the browser test pins at RichTextBlock.browser.test.ts:65. Trailing-newline handling (:23) makes the
4-span assertion at test:72 correct for a 4-line fence. Round-trip escaping holds: `textContent`
in → `innerHTML` out re-escapes `<`/`&`, which is what the smuggling test at test:90-102 asserts.
Empty fence degrades to zero spans; a non-`diff` fence is untouched (test:80-88). No `setConfig`
call exists anywhere in `ui/src` (grepped), so DOMPurify's per-call config cannot be shadowed by a
persisted one and `RETURN_DOM_FRAGMENT` is honoured on this call site.

**No contract changed.** `parseVisualBlocks`, the sidecar contract, `joinDiffBlocks` and every other
renderer are untouched. `buildRecapPrompt`'s signature is unchanged; its two other consumers
(`src/recap.ts`, `test/untrusted.test.ts:113-120`) pass the same shape and assert nothing about the
block-guidance text.

I found no correctness or security defect in this diff.

### Stated limitations

- `node_modules` is not installed in this worktree, so I could not open `dompurify`'s `.d.ts` to
  confirm that the `RETURN_DOM_FRAGMENT: true` overload types the return as `DocumentFragment` for
  the pinned `^3.4.14` (ui/package.json:50). That is the documented DOMPurify 3.x API and CI's
  `bun run check` covers the typing; assumed as written, not asserted correct.
- Acceptance #2 (the regeneration histogram) and #3 (the UI spot-read) are empirical results that
  live in the PR body, not in the diff — outside what I can verify here.

## Nits (non-blocking):

- src/recap-core.ts:253 — the new compose list recommends `img`, but `src` is in the wireframe
  sanitizer's `FORBID_ATTR` (WireframeBlock.svelte:81) and the render iframe's CSP allows only
  `img-src data:` (WireframeBlock.svelte:162), so any `` an author emits renders empty; dropping
  `img` from the list would avoid inviting broken placeholders into mockups.
- test/recap-core.test.ts:999-1011 — six of the twelve `BLOCK_SCHEMAS` entries are truncated
  mid-object (`…"note?":"<short>"}` for file-tree, similarly for diff / annotated-code / data-model /
  api-endpoint / checklist), so the "byte-identical" guarantee the comment at :993-998 describes is
  only enforced over a prefix of those six lines.
- src/recap-core.ts:224 — the issue specified "typically 3-6 blocks"; the shipped line says
  "typically 4-7 blocks". Consistent with the acceptance band (3-7), but it is a substance edit to a
  clause the issue asked to be edited for house voice only.
- src/recap-core.ts:271 — the sidecar file instruction still ends "Omit the file entirely when plain
  markdown suffices", which reads as the same standing invitation the new opening at :211-212
  deliberately narrowed. Pre-existing line, but the two now sit in mild tension.
- src/operator-language.ts's verbatim caveat ("Never a path, flag, identifier, enum value, or command
  fragment — reproduce those verbatim", :154-156) is scoped to `table.columns`/`table.rows`, while
  `rich-text.markdown` is a plain translate field (:75); the new shape-sketch fence therefore has no
  explicit verbatim instruction for the paths and identifiers it will contain in German mode. Only a
  wording risk in the prompt, not a defect — flagged because the plan asserted the opposite.
- RichTextBlock.svelte:114-121 adds `color: var(--color-green)` / `var(--color-red)` on top of the
  background tint, which DiffFileBlock.svelte:207-211 does not do, so the two do not read *exactly*
  alike despite the comment at :112-113.

- `.claude/rules/ui-design-system.md` rule 4 reserves `--color-green` for actionable-complete
  (READY); RichTextBlock.svelte:115-116 uses it for diff-added lines. That is the established
  convention for diff surfaces in this repo (DiffFileBlock.svelte:208), so I read the rule as not
  applying to diff semantics — noting it only because the new use extends it to foreground text.

- src/recap-core.ts:253 — the wireframe helper-class + composable-tag enumeration goes beyond the
  clause the issue specified and is a meaningful share of the byte budget; it is well-motivated (the
  prompt previously named "the wf helper classes" without defining them) and verified accurate, but
  it is added content the task did not ask for.
- src/recap-core.ts:206-212 — the issue and the approved plan both said the existing "Omit `blocks`
  entirely if plain `body` suffices — blocks are not required" escape hatch stays; the diff rewrote
  it into "For a session that changed code, blocks are how the operator reads it". Documented with an
  empirical rationale in the comment, and it is the author's call, but it is a behaviour-steering
  edit outside what was specified.
- RichTextBlock.svelte:17-41 — the issue's Task 2 asked only for a `:global(pre)` rule; the
  post-sanitize diff-line tinting pass is additional runtime behaviour. The approved plan described
  it and I verified it is safe, so this is a judgement call, not a defect.
- test/recap-core.test.ts:1058-1066 — the byte ceiling is asserted at 7424 rather than the plan's
  7168, i.e. slightly above the issue's "~7 KB" constraint. Documented in the test comment.

- `src/service.ts:1329-1330` — the plan-gate wireframe line also says "Author with the wf helper classes" without defining them anywhere, so the gap this PR closed on the recap side remains open on the plan side.
- `ui/src/lib/components/HerdrUpdateModal.svelte:56-69` installs a global DOMPurify `afterSanitizeAttributes` hook on the shared module singleton, which applies to any other component's concurrent `sanitize` call, including RichTextBlock's.

PR #2212, opened 2026-09-08
fix(ui): preserve actual herdr update results in confirmation dialog

kai-osthoff:
## Summary

After updating herdr from 0.8.2 to 0.9.0, the confirmation dialog changed its version range to 0.9.0 → 0.9.0 because it rendered the refreshed installation status. Use the completed result's `from`/`to` values so the actual transition remains visible after subsequent status refreshes.

Show localized success or failure titles for updates and both downgrade paths, including the accessible dialog name. Add browser regressions for event ordering, later status changes, downgrade results, failure states, and missing version values.

## Validation

- Regression observed failing with `0.9.0 → 0.9.0` before the fix; all 17 herdr dialog tests now pass.
- `cd ui && bun run check` — zero errors or warnings.
- `cd ui && bun run check:i18n` — both locales in parity.
- `cd ui && CI=true bun run test --maxWorkers 4 --reporter=verbose` — 4,801 tests pass across 300 files.
- `bun run lint` — passes.
- `bun run test` with Bun 1.4.2 — 9,158 pass, 16 skipped. The installed Bun 1.3.10 hung in an unrelated Git subprocess test; the current runtime completed the suite.
- Self-reviewed the four-file diff for correctness, security, and scope.

## Checklist

- [x] Conventional-commit PR title.
- [x] Branch cut from `origin/main`, kept linear.
- [x] Relevant lint, checks, and tests pass.
- [x] User-facing text uses matching EN/DE catalogs.
- [x] Bug fix to an existing dialog; no new feature announcement or design tokens required.
- [x] No public API, configuration, or CLI changes requiring documentation.


kai-osthoff (commented):
# Review

Result: no blocking findings in the requested diff.

## Verified

- The modal gives terminal result versions precedence over refreshed status versions and only renders a transition when both display values are known (`ui/src/lib/components/HerdrUpdateModal.svelte:104-110`). The server emits the refreshed status before its terminal result, while preserving the original `from` and target `to` on a successful update (`src/herdr-update.ts:425-458`, `src/herdr-update.ts:474`). The store retains that result and the overlay passes it to this modal (`ui/src/lib/store.svelte.ts:704-719`, `ui/src/lib/components/page/AppOverlays.svelte:481-493`).
- The terminal title uses the completed result’s success state and the action-local downgrade flag; the same derived title supplies both visible heading and dialog accessible name (`ui/src/lib/components/HerdrUpdateModal.svelte:92-100`, `ui/src/lib/components/HerdrUpdateModal.svelte:200-209`).
- All four new title keys occur in both configured locales (`ui/project.inlang/settings.json:3-11`, `ui/messages/en.json:15-20`, `ui/messages/de.json:15-20`). The repository i18n gate requires identical non-empty keys across every catalog (`ui/scripts/check-i18n.mjs:17-59`).
- The added browser cases cover the reported 0.8.2 to 0.9.0 status-refresh sequence, subsequent refresh stability, both downgrade paths, failures, and null-result fallback behavior (`ui/src/lib/components/HerdrUpdateModal.browser.test.ts:32-145`, `ui/src/lib/components/HerdrUpdateModal.browser.test.ts:280-307`).

Tests were not run per review instruction.

kai-osthoff:
READY ✅🚀🎉🔥 scoop, do your job—push it. Radical idea. 🙄👉💪📦

PR #2195, opened 2026-09-06
refactor(skills): prompt-audit pass on the shipped skills

scoop:
Audited the six shipped skill files with the `prompt-audit` procedure from [anthropics/skills](https://github.com/anthropics/skills/blob/main/skills/claude-api/shared/prompt-audit.md), then applied the high- and medium-confidence findings.

The procedure's frame is **"every token earns its place"**, explicitly *not* "make it short" — so this is a correctness pass. Net −3 lines. Reasoned constraints, fragile-operation scripts and working redundancy all stay.

## Provenance — why this is concentrated in one file

Five of the six files went through #2016 (doctor pass) and/or #2018 (Claude 5 context-engineering rules) and are largely clean. `merge-train/SKILL.md` was authored **2026-06-16** and has **never been edited since** — neither pass touched it. Both high-confidence findings are in it, and both are the same failure mode: skill text contradicting code.

The two `shepherd-onboarding/references/*.md` files yielded nothing and are untouched.

## Findings applied

| # | Finding | prompt-audit pattern | Confidence | Action |
| --- | --- | --- | --- | --- |
| F1 | merge-train's home-base rule rests on a mechanism #828 removed | 1d Fossils + 2 Volatile specifics | High | rewrite ×2 |
| F2 | Five capped emphases stacking one manual-steps instruction | 1a Pressure language + 1c Padding | High | rewrite |
| F3 | Gate rules prescribe a bare `bun test` that CLAUDE.md forbids | 2 Volatile specifics | High | rewrite |
| F4 | "Create a TodoWrite item per stage" | 1b plan-before-acting + 3 tool names in prose | Medium | remove ×2 |
| F5 | "so it is no longer invisible" | 1d Migration-relative phrasing | Medium | rewrite |
| F6 | "This skill is self-contained" — maintainer note, not agent instruction | 2 Verbose SKILL.md | Medium | rewrite |
| F7 | "…which is where this stage usually goes wrong" | 1c Strategy coaching | Medium | remove clause |

### F1 — the home-base fossil

The skill claimed Shepherd cuts every new task branch from the home base's checked-out `HEAD`, and warned in caps against a `git checkout` there. PR #828 — *"fix(task): base new tasks off the repo default branch, not the current checkout"* — **merged 2026-06-19, three days after this skill landed**:

- `ui/src/lib/api.ts` → `pickBaseBranch()`, by its own comment the "single source for every New Task / quick-launch / merge-train spawn site", resolves `origin/HEAD` **first**
[...]

PR #2214, opened 2026-09-08
refactor(prompts): align in-code manual-steps copies with skill wording (#2206)

### LINKED ISSUE #2206: Align the in-code manual-steps copies with the skill wording rewritten in #2195
## Goal

Bring the two in-code copies of the manual-steps guidance into line with the skill copy that #2195 rewrote, so the same rule reads the same way whichever path delivers it.

## Background

The prompt audit in #2195 rewrote the manual-steps paragraph in the `shepherd-pull-requests` skill. It replaced a stack of five capped emphases (`DEFAULT TO DECLARING NOTHING`, `NO manual steps`, `ONLY`, `OMIT`, `NEVER invent`) — which stated one instruction five ways and never gave its reason — with a plain statement carrying the actual consequence: an un-acked non-`POST-MERGE` step blocks the PR's auto-merge (`hasBlockingManualSteps`), so a spurious one strands a PR that was otherwise ready to land.

That audit was scoped to skill files only, so the same guidance still ships in its pre-audit register from two other places:

| Copy | Symbol | Register | Reaches the agent |
| --- | --- | --- | --- |
| Skill | `shepherd-pull-requests` SKILL.md | **rewritten** by #2195 | when the model loads the skill |
| Spawn notice | `MANUAL_STEPS_NOTICE` (`src/service.ts`) | old emphasis stack | only when `prBlocksResident` — mutually exclusive with the skill |
| Tool guard | `PR_CREATE_CONTEXT` (`scripts/tool-guard.mjs`) | "Most PRs need NONE: declare nothing rather than inventing one" | fires at `gh pr create` **alongside** the skill |

## Why it's worth doing

The spawn notice is the low-stakes half: `prBlocksResident` makes it and the skill mutually exclusive, so no single spawn sees both registers. Nothing contradicts today — this is consistency debt, not a bug.

The guard copy is the one that actually matters. `PR_CREATE_CONTEXT` is a deliberate backstop that fires on the `gh pr create` call itself, so it co-occurs with the skill rather than replacing it. An agent that has loaded the rewritten skill now gets the calm, reasoned version and the older `declare nothing rather than inventing one` phrasing in the same turn. Same rule, two registers, one turn — exactly the "duplicated rules make the model spend effort reconciling wordings" pattern the audit was removing.

## Vertical cut

Rewrite both copies to match the skill's register, keeping each one's distinct job: the notice is the full standalone rule, the guard context is the short backstop. Attach the auto-merge consequence in place of the emphasis, as the skill now does. Do not merge the two — they are deliberately separate delivery paths.

## Test seams that pin these

Both have drift guards that must be updated or kept passing deliberately, not worked around:

- `parseManualSteps(MANUAL_STEPS_NOTICE)` (in `test/manual-steps.test.ts`) runs the **real parser** over the notice's embedded carrier example and asserts the documented steps, `POST-MERGE` included. The example block must stay byte-parseable; only the prose around it changes.
- `test/service.test.ts` asserts the composed system prompt contains the literal string `DEFAULT TO DECLARING NOTHING`, as an anchor for "the emphatic default-empty rule". That assertion has to be re-pointed at whatever the new wording anchors on — re-point it, don't delete the check.

## Acceptance criterion

No copy of the manual-steps guidance still reads `DEFAULT TO DECLARING NOTHING` / `NEVER invent steps`; all three state the default-empty rule once with the auto-merge consequence attached; `bun run test` green with both drift guards re-pointed rather than removed.

## Out of scope

The rest of the in-code prompt surface. #2195 deliberately scoped in-code constants out; this issue covers only the manual-steps copies it left inconsistent.

scoop:
Closes #2206.

#2195's prompt audit rewrote the manual-steps paragraph in the `shepherd-pull-requests` skill: five capped emphases (`DEFAULT TO DECLARING NOTHING`, `NO manual steps`, `ONLY`, `OMIT`, `NEVER invent`) stating one instruction five ways, replaced by a plain statement carrying the actual consequence — an un-acked non-`POST-MERGE` step blocks the PR's auto-merge (`hasBlockingManualSteps`), so a spurious one strands a PR that was otherwise ready to land. That audit was scoped to skill files, leaving the same guidance shipping in its pre-audit register from two other places.

## What changed

- **`MANUAL_STEPS_NOTICE` (`src/service.ts`)** — the full standalone rule, delivered only when `prBlocksResident`. Its closing paragraph is now the skill's, verbatim. The low-stakes half: mutually exclusive with the skill, so nothing contradicted today; this is consistency debt.
- **`PR_CREATE_CONTEXT` (`scripts/tool-guard.mjs`)** — the copy that actually mattered. It is a deliberate backstop firing on the `gh pr create` call itself, so it **co-occurs** with the skill rather than replacing it: an agent that had loaded the rewritten skill got the calm reasoned version and `declare nothing rather than inventing one` in the same turn. Same consequence, kept in the guard's compressed register and at roughly its old length.

The two are **not** merged into one shared constant — deliberately separate delivery paths with different jobs and gates, and the hook script must not import from `src/`.

## Drift guards

Both are re-pointed or kept passing deliberately, not worked around:

- `parseManualSteps(MANUAL_STEPS_NOTICE)` (`test/manual-steps.test.ts`) runs the **real parser** over the notice's embedded carrier example. Every edit here is prose *after* the fenced block, so this test passes **unmodified** — that is the proof the carrier contract held.
- `test/service.test.ts` anchored on the literal `DEFAULT TO DECLARING NOTHING`. Re-pointed at both halves of the replacement — `Most PRs need no manual steps` (the rule) and `blocks the PR's auto-merge` (its reason) — so neither can drift out silently.
- `test/tool-guard.test.ts` asserts only literals this diff does not touch (`ONE pull request`, `shepherd:manual-steps`, `Manual-Step:`, `shepherd-pull-requests`); verified by reading them, unchanged.

## Verification

`bun run lint` clean, `bun run test` green (9164 pass / 0 fail), `bunx tsc --noEmit` clean, fallow pre-push audit clean. A repo-wide grep for `DEFAULT TO DECLARING NOTHING` / `NEVER invent` / `declare nothing rather than` returns nothing outside git history; all three copies now state the default-empty rule once with the consequence attached.

## Known residual

Nothing pins the three copies to each other, so register drift can recur. Adding such a mechanism was out of scope for #2206 — flagged rather than solved.

### DIFF
--- a/scripts/tool-guard.mjs
+++ b/scripts/tool-guard.mjs
@@ -68,9 +68,10 @@ export const PR_CREATE_CONTEXT =
   "`gh issue create` a follow-up. Before opening it, declare any MANUAL OPERATOR STEPS the change " +
   "implies (feature flag, env var, backfill, restart, DNS, seeded record) in the PR body — either " +
   "a ```shepherd:manual-steps``` fenced block of `- [ ]` lines or column-0 `Manual-Step:` trailer " +
-  "lines, prefixed `POST-MERGE:` when they must happen after merge. Most PRs need NONE: declare " +
-  "nothing rather than inventing one, and use `gh pr edit --body` if you only notice one later. " +
-  "The `shepherd-pull-requests` skill has the full rules.";
+  "lines, prefixed `POST-MERGE:` when they must happen after merge. Most PRs need none — an " +
+  "un-acked non-`POST-MERGE` step blocks the PR's auto-merge, so a spurious one strands a ready " +
+  "PR; use `gh pr edit --body` if you only notice one later. The `shepherd-pull-requests` skill " +
+  "has the full rules.";
 
 /**
  * Injected when a Bash call is backgrounded. A detached job reparents to PID 1 and outlives the
--- a/src/service.ts
+++ b/src/service.ts
@@ -998,7 +998,8 @@ export function composeEpicSteer(text: string): string | null {
  * to the parser by a unit test (`parseManualSteps(MANUAL_STEPS_NOTICE)` in test/manual-steps.test.ts).
  * The fence-open line must stay EXACTLY ```shepherd:manual-steps (FENCE_OPEN regex) and each step a
  * column-0-or-indented `- [ ]` line (TASK_LINE) so a wording edit can never silently break parsing.
- * The notice is emphatically default-empty: a fabricated step is worse than none.
+ * The notice is default-empty, and says why: an un-acked non-`POST-MERGE` step blocks the PR's
+ * auto-merge (hasBlockingManualSteps, src/automerge-core.ts), so a fabricated step strands the PR.
  */
 export const MANUAL_STEPS_NOTICE =
   "Before you open the pull request, declare any MANUAL OPERATOR STEPS the change implies — work a " +
@@ -1015,10 +1016,10 @@ export const MANUAL_STEPS_NOTICE =
   "- Or column-0 `Manual-Step:` trailer lines (flush-left, outside any fence), e.g. a line reading " +
   "exactly `Manual-Step: Rotate the signing key`.\n" +
   "Prefix a step with `POST-MERGE:` when it must happen AFTER the PR merges.\n" +
-  "DEFAULT TO DECLARING NOTHING: most PRs need NO manual steps. Add a step ONLY for a real " +
-  "out-of-band action a human must take; if merging fully completes the change, OMIT the carrier " +
-  "entirely. NEVER invent steps to fill the block — a spurious step is worse than none. When in " +
-  "doubt, declare nothing.";
+  "Most PRs need no manual steps: if merging fully completes the change, omit the carrier " +
+  "entirely. Declare a step only for a real out-of-band action a human must take — an un-acked " +
+  "non-`POST-MERGE` step blocks the PR's auto-merge, so a spurious one strands a PR that was " +
+  "otherwise ready to land.";
 
 /**
  * Injected as the highest-priority directive for an attended RESEARCH task (`research: true`).
--- a/test/service.test.ts
+++ b/test/service.test.ts
@@ -4812,9 +4812,11 @@ test("composeSystemPrompt rides the manual-steps notice on code spawns, never on
   for (const sp of codeSpawns) {
     expect(sp).toContain("<manual-steps-notice>");
     expect(sp).toContain("</manual-steps-notice>");
-    // Anchor on the stable carrier syntax (the parser contract) + the emphatic default-empty rule.
+    // Anchor on the stable carrier syntax (the parser contract) + the default-empty rule and the
+    // auto-merge consequence that is its reason (#2206) — both halves, so neither can drift out.
     expect(sp).toContain("```shepherd:manual-steps");
-    expect(sp).toContain("DEFAULT TO DECLARING NOTHING");
+    expect(sp).toContain("Most PRs need no manual steps");
+    expect(sp).toContain("blocks the PR's auto-merge");
   }
   expect(composeSystemPrompt(null, false, { ...noMech, research: true })).not.toContain(
     "<manual-steps-notice>",


scoop (commented):
# Review — refactor(prompts): align in-code manual-steps copies with skill wording (#2206)

Three prose/test files, no behaviour change. I verified each claim the change depends on rather than
eyeballing it; citations below.

## What I verified

**The consequence the new copy states is real.** `hasBlockingManualSteps` is
`s.manualStepsAckedAt == null && s.manualSteps.some((st) => !st.postMerge)`
(`src/automerge-core.ts:116-118`), and it is a `readyToMerge` disqualifier
(`src/automerge-core.ts:158-163`) as well as the `manual_steps` hold predicate
(`src/automerge-core.ts:307-317`). So "an un-acked non-`POST-MERGE` step blocks the PR's auto-merge"
is accurate, including the `POST-MERGE`-exempt carve-out both new copies preserve. The doc comment's
new pointer (`src/service.ts:1002`, "hasBlockingManualSteps, src/automerge-core.ts") resolves to that
exact symbol and file.

**The notice now matches the skill verbatim.** `src/service.ts:1019-1022` reassembles to the same
sentences as `agent-skills/.claude/skills/shepherd-pull-requests/SKILL.md:67-69`, word for word
including the `POST-MERGE` backticks. I compared the concatenated chunks against the skill lines
directly.

**The carrier contract is untouched.** The fence-open line, both `- [ ]` example lines and the
closing fence are unchanged at `src/service.ts:1012-1015`; the diff only replaces prose *after* them.
The drift guard `parseManualSteps(MANUAL_STEPS_NOTICE)` at `test/manual-steps.test.ts:90-95` asserts
exactly those two steps (`ms1` non-post-merge, `ms2` post-merge), so it stays green unmodified — and
the new paragraph introduces no column-0 `Manual-Step:` line and no `- [ ]` line that could add a
phantom step (`src/service.ts:1019-1022` is a single unbroken line of prose).

**The re-pointed anchor actually matches, byte for byte.** `test/service.test.ts:4818-4819` asserts
`Most PRs need no manual steps` and `blocks the PR's auto-merge`. Both substrings are contiguous
within single concat chunks (`src/service.ts:1019` and `src/service.ts:1021` respectively — the
second is not split across the `+` boundary), and both apostrophes are ASCII `'` on both sides
(checked with `cat -A`: no `M-bM-^@M-^Y` smart quote in either file). The assertion is re-pointed,
not deleted, and it keeps the `` ```shepherd:manual-steps `` carrier assertion at
`test/service.test.ts:4817`.

**No other test keys on the removed wording.** Repo-wide grep for `DEFAULT TO DECLARING NOTHING`,
`NEVER invent`, `declare nothing rather than inventing`, `worse than none`, `OMIT the carrier`,
`When in doubt, declare` returns nothing outside git history (the single `worse than none` hit is
`src/codex-activity.ts:290`, unrelated prose about resolution fallbacks). `test/tool-guard.test.ts:228-235`
asserts only `ONE pull request`, `shepherd:manual-steps`, `Manual-Step:`, `shepherd-pull-requests` —
none of which the guard edit touches. `test/prompt-budget.test.ts:306,326,345,353,357,360` keys only
on the block *name* `manual-steps-notice`, never on wording.

**Exactly three copies exist and all three now agree.** Grep for `MANUAL OPERATOR STEPS` hits only
`SKILL.md:48`, `scripts/tool-guard.mjs:68`, `src/service.ts:1005`; grep for `Most PRs need` hits those
three plus the test anchor. No docs-site mirror, no bundled duplicate of the hook script. The two
copies are kept separate (no shared constant), which is correct — `scripts/tool-guard.mjs` is a
standalone hook and must not import from `src/`.

**Prettier width.** `printWidth: 100` (`.prettierrc:4`); the longest new line is
`scripts/tool-guard.mjs:72` at 99 columns, and prettier does not reflow string-concat literals
anyway. No formatting hazard.

## Limitations

None material — every dependency of this change is in-repo and was opened and compared.

## Nits (non-blocking):

- `scripts/tool-guard.mjs:71-74`: the guard's new wording has no drift anchor of its own — `test/tool-guard.test.ts:228-235` asserts only the carrier syntax and the skill pointer, so the guard copy can silently drift back out of register with the skill without failing CI (the notice copy does have one, via `test/service.test.ts:4818-4819`). The issue only required re-pointing the service anchor, so this is a suggestion, not an omission.
- `scripts/tool-guard.mjs:71`: the guard compresses to "Most PRs need none" where the skill and notice say "Most PRs need no manual steps" — deliberate per the plan's short-backstop register; noting only that the antecedent ("MANUAL OPERATOR STEPS") is three clauses back.

No blocking issues found. The acceptance criterion holds: no copy still reads
`DEFAULT TO DECLARING NOTHING` / `NEVER invent steps`, all three state the default-empty rule once
with the auto-merge consequence attached, and both drift guards are re-pointed or kept green
deliberately rather than worked around.

PR #2215, opened 2026-09-08
feat(recap): give the recap agent the changed UI markup (#2209)

### LINKED ISSUE #2209: Recap wireframe can never fire: the recap agent cannot see the UI (writer-only, no diff in prompt)
Split out of #2194, which set `wireframe > 0` as a must-move criterion. PR #2208 shipped the guidance layer and moved everything it could — `mermaid` 0 → 1, blocks/recap median 10 → 5.5, max 14 → 7, no zero-block recaps — but `wireframe` stayed at zero across **30 regenerations over three prompt variants**, and it cannot move from the prompt alone.

## Why

The recap agent has never seen the UI it is being asked to mock up:

- The spawn runs the `writer-only` preset in `src/transient-agent-argv.ts` — `allowedTools: ["Write"]`. No `Read`, no `Grep`, no `Glob`, a disposable temp cwd, and no `--add-dir`.
- `buildRecapPrompt` (`src/recap-core.ts`) carries the task prompt, plan, changed **file paths + statuses**, and the transcript digest. **No diff content.** That is deliberate: the prompt tells the agent "Never type the code body — the server attaches it", and `joinDiffBlocks` attaches diff bodies to `diff` blocks *after* the agent writes them.

So a `wireframe` — "a mockup of the resulting screen" — would have to be invented, and the prompt's own grounding rule ("Never invent a path, a field, or a change") correctly stops it. The model is behaving well; the input is missing.

The asymmetry is the evidence: identical guidance moved `mermaid`, which is derivable from the transcript digest and file names, and could not move `wireframe`, which needs the markup. #2194 inferred "prompt, not pipeline" from `mermaid` firing on the plan-gate side — but the plan reviewer runs with full repo Read access, so that comparison does not transfer to the recap spawn.

Attempts already ruled out (all in PR #2208's history, no need to repeat them):

- per-type trigger clause on the `wireframe` schema line;
- a preamble rule stating `wireframe` is **not optional** for a user-visible change;
- supplying the helper-class vocabulary (`wf-card`/`wf-box`/`wf-pill`/`wf-chip`/`wf-muted`), which the prompt had been referencing while defining nowhere — that gap is now fixed, and it still did not fire.

## Options

1. **Switch the recap spawn to `writer-ro` + `--add-dir <worktree>`** so it can read the changed Svelte files. Note the security shape: the recap agent ingests untrusted transcript text, and `--add-dir` also exposes that directory to the preset's bare `Write` (Claude Code has no read-only form of the flag).
2. **Inject bounded `ui/**` diff hunks into the prompt** so a mockup can be grounded without new tool grants. Competes with the `fitRecapPrompt` argv budget (#1944), which already clamps plan/context first.
3. **Accept that recap cannot produce wireframes** and drop the type from the recap prompt's menu, leaving it to the plan gate, where the agent can actually see the code.

Option 3 deserves real consideration before 1 or 2: a block type that is listed but never emittable costs prompt budget on every spawn and misleads the next person who measures the histogram.

## Acceptance

Whatever route is taken, the bar from #2194 stands: regenerate a sample of UI-change sessions and show `wireframe > 0`, with blocks/recap median staying in 3-7 and max not exceeding 14.

scoop:
Closes #2209.

`wireframe` had fired **0 times** — across 88 block-emitting recaps, then across 30 more regenerations over three prompt variants in #2208. It could not fire from the prompt: the recap spawn runs the `writer-only` preset (`allowedTools: ["Write"]`, disposable temp cwd, no `--add-dir`) and `buildRecapPrompt` carried changed **paths and statuses** but no markup. A mockup of the resulting screen would have to be invented, and the prompt's own grounding rule — "Never invent a path, a field, or a change" — correctly stopped that. The model was behaving well; the input was missing.

This supplies the input, from the diff `RecapService.generate()` **already holds in memory**.

## What changed

- **`src/recap-core.ts`** — `buildUiMarkupDigest()`: a pure `DiffFile[] → string` helper (sibling of `buildTranscriptDigest`) that renders the **post-change side** of the changed view files' hunks — `-` lines dropped, `+` kept on added lines, a leading space on context, hunk headers dropped (they are line numbers, which the prompt already forbids the agent from repeating back). `buildRecapPrompt` gains an optional `uiMarkup` and renders it as a fenced section, with the one line that grounds `wireframe` on it sitting **outside** the fence — an instruction inside one is contractually ignorable and forgeable (`src/untrusted.ts`).
- **`src/recap.ts`** — `generate()` builds the digest and threads it through; `fitRecapPrompt` gains a `uiMarkup` clamp spec placed **first in the spec list**, so the markup yields argv bytes before `plan` and `context` — those keep exactly the protection they had (#1944).

Selection is deliberately **extension-based, never path-based** (Shepherd drives arbitrary repos, so a `ui/**` rule would only ever be right about this one): `.svelte/.vue/.astro/.tsx/.jsx/.html/.htm`, minus tests, deleted, binary, and files whose hunks `diff.ts` dropped past its line cap — silence beats a lie. Ranked by additions, capped at 3 files / ~6 KB.

**No preset change, no `--add-dir`, no new I/O, no new tool grant.** Issue option 1 was rejected on its security shape: `--add-dir` has no read-only form, so the preset's bare `Write` would reach a live session worktree while the agent ingests untrusted transcript text.

The static block-guidance region is **byte-identical** — it sits at 7387 of the 7424-byte ceiling #2208 pinned, so the new instruction rides the per-session section instead of eating the last 37 bytes.

## Empirical result

10 archived UI-change sessions, regenerated through the real `RecapService` against a **copy** of the live DB (no live recap row was written).

| metric | #2208 (before) | this branch |
| --- | --- | --- |
| `wireframe` | **0** | **10** (in 9 of 10 recaps) |
| blocks/recap median | 5.5 | **6.5** (target 3-7) |
| blocks/recap max | 7 | **8** (must not exceed 14) |
| zero-block recaps | 0 | **0** |
| blocks total | 54 | 64 |

Per-type: `diff`=19 `wireframe`=10 `callout`=9 `file-tree`=9 `checklist`=8 `rich-text`=4 `mermaid`=4 `code`=1.

The harness also compared the **raw** `.shepherd-recap-blocks.json` against the parsed output: **0 blocks dropped by the validator**. That check exists because `validateWireframe` silently drops a wireframe carrying an inline hex/`rgb()`/`font-family`/`box-shadow`, a bad `surface`, or >20k chars — so #2208's zero could not distinguish "never emitted" from "emitted and dropped". It can now: they were genuinely never emitted, and the ones produced here are clean (helper classes, class-based color, no inline color anywhere).

Spot-read of a produced wireframe (TASK-2241): a `mobile` surface showing the notification-shade icon before/after, captioned, built from `wf-card`/`wf-box`/`wf-pill`/`wf-muted` — grounded in the markup, not invented.

## Caveats

- **n=10, one sample per session.** The categorical result (`wireframe > 0`, and 10/10 recaps produced) is robust to that; the medians are not distributions.
- **Not a perfectly matched A/B.** `.shepherd-plan.md` is never committed and the sessions' worktrees are gone, so regenerated recaps ran without plan text. Same bias #2208 noted — the new run saw *less* input, not more.
- A modified component yields markup **fragments**, not a whole screen; the mockup is then of the region that changed. That is what the operator needs to see, and it is what the measurement tested.
- A UI change whose file exceeded `diff.ts`'s per-file line cap contributes nothing — deliberately.

### DIFF
--- a/docs/sandbox-security.md
+++ b/docs/sandbox-security.md
@@ -100,7 +100,8 @@ execution controls below, Shepherd bounds the injection surface at ingestion
 
 - **Untrusted-content fencing.** External text an agent or helper LLM might read —
   issue title/body, issue comments, PR bodies + author-notes, captured terminal
-  tails, and the recap context — is wrapped in unforgeable `⟦UNTRUSTED:…⟧`
+  tails, the recap context, and the changed-view-file markup the recap prompt
+  carries (`ui-markup`) — is wrapped in unforgeable `⟦UNTRUSTED:…⟧`
   markers (`fenceUntrusted`, with a per-fence random nonce that the content cannot
   predict or close early) so the model treats it as **data, never instructions**.
   The fence carries its **label and nonce only**; the instruction hierarchy has one
--- a/src/recap-core.ts
+++ b/src/recap-core.ts
@@ -223,28 +223,32 @@ export function buildUiMarkupDigest(
   const maxChars = limits.maxChars ?? RECAP_UI_MARKUP_MAX_CHARS;
 
   const selected = files
-    .map((file, order) => ({ file, order }))
+    .map((file, order) => ({ file, order, after: afterSide(file.hunks) }))
+    // The post-change side is rendered BEFORE selection, because emptiness is part of the
+    // selection rule and it costs a file its slot: a component emptied but KEPT (status still
+    // `modified`, hunks non-empty, every line a deletion) has no resulting screen. Emitting its
+    // header alone would put the section — and the instruction to ground a wireframe on it — over
+    // a fence holding no markup, which is the invent-a-screen pressure this change exists to remove.
     .filter(
-      ({ file }) =>
+      ({ file, after }) =>
         isViewFile(file.path) &&
         file.status !== "deleted" &&
         !file.binary &&
         !file.truncated &&
-        file.hunks.length > 0,
+        after.trim() !== "",
     )
     .sort((a, b) => b.file.additions - a.file.additions || a.order - b.order)
     .slice(0, maxFiles);
 
   const blocks: string[] = [];
   let total = 0;
-  for (const { file } of selected) {
+  for (const { file, after } of selected) {
     const header = `${file.path} (${file.status})`;
     // Whatever is left of the section budget, never more than one file's share. Below the marker's
     // own length there is no room for content worth reading, so stop rather than emit a stub.
     const room = Math.min(maxFileChars, maxChars - total - header.length - 2);
     if (room < MIN_UI_MARKUP_FILE_CHARS) break;
-    const body = clipMarkup(afterSide(file.hunks), room);
-    const block = `${header}\n${body}`;
+    const block = `${header}\n${clipMarkup(after, room)}`;
     blocks.push(block);
     total += block.length + 2;
   }
--- a/src/recap-core.ts
+++ b/src/recap-core.ts
@@ -3,7 +3,7 @@
  * Mirrors the structure of critic-core.ts (parse/validate/clamp + prompt building).
  */
 import type { ActivityEntry } from "./activity";
-import type { DiffFileStatus, Recap, RecapVerdict } from "./types";
+import type { DiffFile, DiffFileStatus, DiffHunk, Recap, RecapVerdict } from "./types";
 import { parseVisualBlocks } from "./visual-blocks";
 import type { VisualBlock } from "./visual-blocks";
 import { UNTRUSTED_CONTENT_DIRECTIVE, fenceUntrusted } from "./untrusted";
@@ -144,6 +144,113 @@ export function buildTranscriptDigest(
   return lines.join("\n");
 }
 
+// ── UI markup (#2209) ─────────────────────────────────────────────────────────
+//
+// The recap agent runs the `writer-only` preset — Write only, a disposable temp cwd, no
+// `--add-dir` — so it can read NOTHING. `buildRecapPrompt` carried changed paths and statuses but
+// no markup, which is why `wireframe` fired 0 times across 30 regenerations over three prompt
+// variants (#2194 → #2208): a mockup of the resulting screen would have to be invented, and the
+// prompt's own grounding rule ("Never invent a path, a field, or a change") correctly stops that.
+// The model was behaving well; the input was missing.
+//
+// So the input is supplied, from the diff the service ALREADY holds in memory — no tool grant, no
+// extra I/O, no widening of what the spawn can reach.
+
+/** At most this many view files carry markup into the prompt. A wireframe is one screen, not a tour
+ *  of the diff; past a handful the section is competing with `plan` for the argv budget (#1944). */
+export const RECAP_UI_MARKUP_MAX_FILES = 3;
+
+/** Per-file ceiling, so one large component cannot crowd out the other selected files. */
+const RECAP_UI_MARKUP_MAX_FILE_CHARS = 2500;
+
+/** Section ceiling, to within one truncation marker per file (the marker rides on top of a clipped
+ *  slice). Soft either way: `fitRecapPrompt` clamps this block FIRST if the argv budget still bites. */
+const RECAP_UI_MARKUP_MAX_CHARS = 6000;
+
+/** Below this a file's slice is all marker and no markup — stop instead of emitting a stub. */
+const MIN_UI_MARKUP_FILE_CHARS = 200;
+
+/** Extensions that carry rendered markup. Deliberately EXTENSION-based, never path-based: Shepherd
+ *  drives arbitrary repos, so a `ui/**` rule would be a Shepherd-only heuristic. */
+const VIEW_FILE_RE = /\.(?:svelte|vue|astro|tsx|jsx|html|htm)$/i;
+
+/** `Foo.test.tsx`, `Foo.spec.jsx`, `Foo.browser.test.ts` — a test renders no screen the operator sees. */
+const TEST_FILE_RE = /\.(?:test|spec)\.[^./]+$/i;
+
+/** True for a changed file whose post-change side is worth showing as UI markup. */
+export function isViewFile(path: string): boolean {
+  return VIEW_FILE_RE.test(path) && !TEST_FILE_RE.test(path);
+}
+
+/** Neutral, factual truncation marker — states a count, commands nothing (it lands inside an
+ *  untrusted fence, where an instruction would be contractually ignorable; see untrusted.ts). */
+function elidedChars(n: number): string {
+  return `[… ${n} chars elided …]`;
+}
+
+function clipMarkup(text: string, maxChars: number): string {
+  if (text.length <= maxChars) return text;
+  return `${text.slice(0, maxChars)}\n${elidedChars(text.length - maxChars)}`;
+}
+
+/** Render the POST-CHANGE side of a file's hunks: `-` lines dropped, `+` kept on added lines, a
+ *  leading space kept on context lines. Hunk headers are dropped — they are line numbers, which the
+ *  prompt already forbids the agent from repeating back. Hunks are separated by a bare `…`. */
+function afterSide(hunks: DiffHunk[]): string {
+  const parts: string[] = [];
+  for (const h of hunks) {
+    const lines = h.lines
+      .filter((l) => l.kind !== "del")
+      .map((l) => `${l.kind === "add" ? "+" : " "}${l.content}`);
+    if (lines.length > 0) parts.push(lines.join("\n"));
+  }
+  return parts.join("\n…\n");
+}
+
+/** Build the bounded UI-markup digest handed to the recap agent, or "" when the session changed no
+ *  view file. Pure — the caller passes the diff it already computed.
+ *
+ *  Selection drops what cannot ground a mockup: deleted files (no resulting screen), binary files,
+ *  and files whose hunks `diff.ts` dropped past its per-file line cap (`truncated`) — silence beats
+ *  a lie, which is the whole lesson of #2209. Ranking is by additions descending (stable on the
+ *  diff's own order): the largest markup change is the most mock-able. */
+export function buildUiMarkupDigest(
+  files: DiffFile[],
+  limits: { maxFiles?: number; maxFileChars?: number; maxChars?: number } = {},
+): string {
+  const maxFiles = limits.maxFiles ?? RECAP_UI_MARKUP_MAX_FILES;
+  const maxFileChars = limits.maxFileChars ?? RECAP_UI_MARKUP_MAX_FILE_CHARS;
+  const maxChars = limits.maxChars ?? RECAP_UI_MARKUP_MAX_CHARS;
+
+  const selected = files
+    .map((file, order) => ({ file, order }))
+    .filter(
+      ({ file }) =>
+        isViewFile(file.path) &&
+        file.status !== "deleted" &&
+        !file.binary &&
+        !file.truncated &&
+        file.hunks.length > 0,
+    )
+    .sort((a, b) => b.file.additions - a.file.additions || a.order - b.order)
+    .slice(0, maxFiles);
+
+  const blocks: string[] = [];
+  let total = 0;
+  for (const { file } of selected) {
+    const header = `${file.path} (${file.status})`;
+    // Whatever is left of the section budget, never more than one file's share. Below the marker's
+    // own length there is no room for content worth reading, so stop rather than emit a stub.
+    const room = Math.min(maxFileChars, maxChars - total - header.length - 2);
+    if (room < MIN_UI_MARKUP_FILE_CHARS) break;
+    const body = clipMarkup(afterSide(file.hunks), room);
+    const block = `${header}\n${body}`;
+    blocks.push(block);
+    total += block.length + 2;
+  }
+  return blocks.join("\n\n");
+}
+
 /** The instruction prompt for the recap spawn. Tells the agent to summarize a COMPLETED
  *  coding session for an operator deciding whether to merge, and to Write the prose to
  *  `.shepherd-recap.md`, any visual blocks to `.shepherd-recap-blocks.json`, and
@@ -154,6 +261,7 @@ export function buildRecapPrompt(input: {
   changedFiles: { path: string; status: DiffFileStatus }[];
   digest: string;
   context: string; // pre-rendered critic verdict / CI / readyToMerge lines (may be "")
+  uiMarkup?: string; // buildUiMarkupDigest output; "" / absent when no view file changed (#2209)
   operatorLanguage?: OperatorLanguage;
 }): string {
   const lines = [
@@ -176,6 +284,23 @@ export function buildRecapPrompt(input: {
     lines.push(
       "Files changed in this session:",
       ...input.changedFiles.map((f) => `  ${f.path} (${f.status})`),
+      "",
+    );
+  }
+
+  // #2209: the ONLY sight of the UI this agent gets. It rides here — with the changed-file list it
+  // annotates, and outside the static block-guidance region, which has ~37 bytes of headroom under
+  // the ceiling its test pins. The interpreting instruction sits AFTER the fence: an instruction
+  // inside one is contractually ignorable and forgeable (see untrusted.ts), so it would be inert
+  // exactly where it matters.
+  if (input.uiMarkup?.trim()) {
+    lines.push(
+      "UI markup after this change (post-change side of the diff for the view files that changed —",
+      "`+` marks a line this session added, a leading space marks surrounding context, `…` separates",
+      "non-adjacent regions of the same file):",
+      fenceUntrusted("ui-markup", input.uiMarkup),
+      "This markup is your grounding for a `wireframe` block: mock up the screen it renders. Do not",
+      "put UI in a wireframe that you cannot see here.",
       "",
     );
   }
--- a/src/recap.ts
+++ b/src/recap.ts
@@ -49,6 +49,7 @@ import { claudeConfigPath, readRepoRootTrusted, trustRepoRoot } from "./claude-t
 import {
   parseRecapVerdict,
   buildTranscriptDigest,
+  buildUiMarkupDigest,
   buildRecapPrompt,
   isSettledIdle,
   needsRecap,
@@ -842,6 +843,10 @@ export class RecapService {
       const plan = this._readPlan(worktreePath);
       const changedFiles = diff.files.map((f) => f.path);
       const changedFilesWithStatus = diff.files.map((f) => ({ path: f.path, status: f.status }));
+      // #2209: the agent reads nothing (writer-only, temp cwd), so the post-change side of the view
+      // files' hunks is the only way a `wireframe` can be grounded rather than invented. "" for a
+      // session that changed no view file — the section is then absent from the prompt entirely.
+      const uiMarkup = buildUiMarkupDigest(diff.files);
 
       const contextParts: string[] = [];
       const review = this.deps.store.getReview(id);
@@ -864,6 +869,7 @@ export class RecapService {
         plan: string;
         changedFiles: { path: string; status: DiffFileStatus }[];
         context: string;
+        uiMarkup: string;
       }): string =>
         buildRecapPrompt({
           taskPrompt: session.prompt,
@@ -871,6 +877,7 @@ export class RecapService {
           changedFiles: v.changedFiles,
           digest,
           context: v.context,
+          uiMarkup: v.uiMarkup,
           operatorLanguage: language,
         });
       const env = this.env();
@@ -882,7 +889,7 @@ export class RecapService {
       // #1944: fit the prompt inside the OS argv budget, or null when it cannot be made to fit.
       // The helper owns the clamp log line so this already-long orchestration carries one branch.
       const prompt = fitRecapPrompt(
-        { plan, changedFiles, changedFilesWithStatus, context },
+        { plan, changedFiles, changedFilesWithStatus, context, uiMarkup },
         composePrompt,
         (p) => ({ wrapped: argvFor(p), spawnEnv: apiKeyPassthroughEnv(false) }),
         `${session.id} (${session.desig})`,
@@ -1190,18 +1197,24 @@ function fitRecapPrompt(
     changedFiles: string[];
     changedFilesWithStatus: { path: string; status: DiffFileStatus }[];
     context: string;
+    uiMarkup: string;
   },
   compose: (v: {
     plan: string;
     changedFiles: { path: string; status: DiffFileStatus }[];
     context: string;
+    uiMarkup: string;
   }) => string,
   assemble: SpawnAssembler,
   label: string,
 ): string | null {
   const fitted = fitFrom({
     ...spawnBudget(assemble),
+    // Order IS the clamp order (fitAssembledPrompt gives blocks up in the caller's order). `uiMarkup`
+    // goes FIRST: it is the newest and most expendable input (#2209), and `plan`/`context` keep
+    // exactly the protection they had before it existed (#1944).
     specs: [
+      { id: "uiMarkup", kind: "text", text: input.uiMarkup, mode: "head" },
       { id: "plan", kind: "text", text: input.plan, mode: "head-tail" },
       { id: "changedFiles", kind: "list", items: input.changedFiles, noun: "files" },
       { id: "context", kind: "text", text: input.context, mode: "head" },
@@ -1209,6 +1222,7 @@ function fitRecapPrompt(
     compose: (v) =>
       compose({
         plan: v.plan as string,
+        uiMarkup: v.uiMarkup as string,
         // clampList emits its marker as a trailing pseudo-entry; re-attach a status so the typed
         // changedFiles shape holds without special-casing it downstream.
         changedFiles: (v.changedFiles as string[]).map(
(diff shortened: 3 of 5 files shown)


scoop (commented):
# Review — feat(recap): give the recap agent the changed UI markup (#2209)

Scope reviewed: `git diff 69d8a5dc7...HEAD` — `src/recap-core.ts`, `src/recap.ts`,
`test/recap-core.test.ts`, `test/recap.test.ts`.

**Verdict: comment.** I found no correctness, security, or contract defect. The change does what
the issue asks — it supplies the one input that was missing — without touching the spawn preset,
the allowlist, or `--add-dir`.

## What I verified, against what

**The data the helper consumes actually exists in the shape it assumes.**
`DiffLine.content` is documented as "line text WITHOUT the leading +/-/space marker"
(`src/types.ts:431`), so `afterSide`'s re-prefixing with `+` / space at `src/recap-core.ts:204`
reconstructs the post-change side correctly rather than doubling a marker. `DiffFile.hunks` is
"empty when binary or truncated" (`src/types.ts:451`), which the selection filter at
`src/recap-core.ts:227-234` treats as belt-and-braces alongside the explicit `binary` / `truncated`
checks.

**The hunks are really populated on the recap path.** `RecapService.resolveGenerateSource` calls
`this._computeDiff` (`src/recap.ts:546`), which defaults to `computeDiff` from `src/diff.ts`
(`src/recap.ts:438`); that returns `parseUnifiedDiff` output with structured `hunks`
(`src/diff.ts:324`). The hunks-stripping variant is `toSessionDiff` (`src/diff.ts:217-221`), used
only for the wire shape, and is not on this path. So `buildUiMarkupDigest(diff.files)` at
`src/recap.ts:849` sees real hunk bodies, not an empty array — the new section is not dead code.

**`truncated` really does mean "no hunks".** `src/diff.ts:125-128` (per-file cap), `168-173` and
`183-187` (global cap) all set `truncated = true` and clear `hunks`. The digest's exclusion of
`truncated` files is therefore not merely defensive; those files genuinely carry nothing.

**Context lines are present, so the post-change side is not a bare list of additions.**
`computeDiff` shells out to `git diff --no-color <ref>...HEAD` with no `-U0`
(`src/diff.ts:318`), i.e. default 3 lines of context, and `appendBodyLine` records `ctx` lines
(`src/diff.ts:98-99`). The prompt's claim at `src/recap-core.ts:299` that "a leading space marks
surrounding context" is accurate.

**Fence handling is sound and the new fence is scrubbed like the existing ones.**
`fenceUntrusted` strips any literal nonce and any pre-existing `⟦…UNTRUSTED…⟧` token from the
content before wrapping (`src/untrusted.ts:51`), so repo markup cannot forge or close the
`ui-markup` fence. The nonce is a fixed 12 hex chars (`src/untrusted.ts:23-25`), so re-composing
during the fit binary search does not change the measured length between iterations — the new
fence introduces no measurement nondeterminism.

**The interpreting instruction is outside the fence, as the repo's own contract requires.**
`src/prompt-fit.ts:30-37` states why an in-fence instruction is contractually ignorable and
forgeable. The "This markup is your grounding for a `wireframe` block…" line sits after
`fenceUntrusted(...)` in the `lines.push` at `src/recap-core.ts:301-303`, and the in-fence
truncation marker `elidedChars` (`src/recap-core.ts:187-189`) states a fact and issues no
directive, matching `elisionMarker` (`src/prompt-fit.ts:40-42`).

**Clamp order is what the comment claims.** `fitAssembledPrompt` iterates `specs` in the caller's
order (`src/prompt-fit.ts:214-225`), so placing `{ id: "uiMarkup", … }` first at
`src/recap.ts:1217` really does make the markup yield argv bytes before `plan`, `changedFiles` and
`context`. On a session with no view file, `uiMarkup` is `""`, `fromBytes` is 0 and `hi <= lo`
(`src/prompt-fit.ts:264`) makes `shrinkSpec` return `null` — the empty spec is skipped rather than
producing a spurious clamp record. Note also that `elidedChars` says "chars" while
`elisionMarker` says "bytes", which is what keeps the two markers distinguishable in the ladder.

**The pinned guidance region is genuinely untouched.** `blockGuidanceRegion` slices from
`"Optionally, add \`blocks\`"` to `"When done, write these files in your CWD:"`
(`test/recap-core.test.ts:988-994`). The new section is pushed at `src/recap-core.ts:296-306`,
before that region starts at `src/recap-core.ts:330`, so the 7424-byte ceiling
(`test/recap-core.test.ts:1069-1071`) is unaffected — and the diff adds no byte inside it.

**No consumer breaks.** `buildRecapPrompt` has exactly one production caller
(`src/recap.ts:874`) and `uiMarkup` is optional on the input type (`src/recap-core.ts:264`), so
the existing test call sites that omit it still type-check. `fitRecapPrompt` is likewise
single-caller (`src/recap.ts:891`). The new test imports resolve: `hostArgvBudget` at
`src/argv-limit.ts:40` and `joinedElementBytes` at `src/argv-limit.ts:58`.

**No i18n surface.** The added strings are agent-prompt text, English throughout like the rest of
`buildRecapPrompt`; the operator-language branch (`src/recap-core.ts:415-424`) governs recap
*output* fields, not prompt chrome. No `ui/messages/*.json` or `extension/messages/*.json` key is
added or referenced, so there is no locale-parity obligation here.

**The markup cannot cause a silently dropped wireframe by itself.** `validateWireframe`
(`src/visual-blocks.ts:498-513`) rejects on `surface`, empty/oversize `html`, unsafe structure,
impure `style=` values and raw `fill`/`stroke` colors — it does **not** inspect class names. So
real component classes leaking into a mockup would render unstyled rather than be dropped. The
residual risk is that the agent copies an inline `style="…#hex…"` out of the fed markup and the
block is then dropped by `wireframeStylesImpure` (`src/visual-blocks.ts:503`); the schema line at
`src/recap-core.ts:378` already forbids that, and the reported 10/10 measurement suggests it did
not bite. Worth keeping an eye on if the histogram ever shows raw-vs-parsed divergence.

**One quantified observation on the argv budget, non-blocking.** `MIN_CLAMP_KEEP = 256`
(`src/prompt-fit.ts:15`) is a hard floor for a `text` spec, so once the section is present it can
never be clamped away entirely: a UI session's prompt now carries roughly 330 bytes of section
chrome plus at least 256 bytes of markup that the ladder cannot recover. That shifts the refusal
threshold (`RECAP_OVERSIZED`, `src/recap.ts:898`) by ~590 bytes for sessions whose `plan`,
`context` and `changedFiles` are all already at their floors — i.e. one where the unclamped
`taskPrompt`/`digest` alone nearly fills the budget. I could not construct a realistic session in
that band, and the ladder recovers kilobytes in every ordinary case, so I am recording it rather
than raising it.

## Stated limitations

- The acceptance bar from #2209/#2194 is a **measurement**, not code: the commit body reports
  "10 regenerated UI sessions: wireframe 0 -> 10, median 6.5, max 8" (median inside 3-7, max
  well under 14). The harness and the DB snapshot were deliberately kept out of the commit (plan
  step 5), so I cannot verify those numbers from this worktree. I did not treat the absence of an
  in-repo harness as a defect — the plan scoped it to the scratchpad on purpose.
- Whether hunk fragments are *sufficient* grounding for a useful mockup is an LLM-behaviour
  question that no static read of this diff can settle.

## Nits (non-blocking):

- `src/recap-core.ts:193` — `clipMarkup` slices with `String.slice`, which can split a surrogate pair (an emoji in UI copy at exactly the cut point) and emit a lone surrogate; `headBytes` at `src/prompt-fit.ts:52-59` deliberately backs off to a character boundary for the same reason. Consequence here is one U+FFFD in the prompt, nothing worse.
- `src/recap-core.ts:191-194` — the caps are counted in UTF-16 code units while the argv ladder measures UTF-8 bytes, so `RECAP_UI_MARKUP_MAX_CHARS = 6000` can be materially more than 6000 bytes on multi-byte markup; harmless because the ladder re-clamps, but the two units sitting next to each other is easy to misread.
- `src/recap-core.ts:242-243` — the inline comment says "below the marker's own length there is no room for content worth reading", but the threshold applied is `MIN_UI_MARKUP_FILE_CHARS = 200`, roughly eight times the marker's length; the doc comment at line 170 phrases the same idea more accurately.
- `src/recap-core.ts:205` — a file whose every hunk is deletion-only (`lines.length === 0` for all hunks) still emits a `path (status)` header with nothing under it, which is precisely the "header with no markup" shape the exclusion filter exists to avoid. Needs a view file emptied but not deleted with zero surviving context lines, so it is close to unreachable under `-U3`.
- `src/recap-core.ts:244` — the section-budget arithmetic reserves 2 chars for a block that costs `header + "\n" + body` plus a 2-char separator, so the accounting is one char optimistic per file on top of the documented per-file marker overshoot. Both are noted as soft in the constant's doc comment.
- `src/recap-core.ts:161` — `RECAP_UI_MARKUP_MAX_FILES` is exported solely for `test/recap-core.test.ts:1175`; the two sibling caps stay module-private.

- `src/recap-core.ts:217-223` — the `limits` parameter (`maxFiles`/`maxFileChars`/`maxChars`) is injected only by tests; production calls `buildUiMarkupDigest(diff.files)` with no overrides (`src/recap.ts:849`). Legitimate as a test seam, but it is unrequested configurability on a single-call-site helper.
- `src/recap-core.ts:181` — `isViewFile` is exported for its own unit test rather than for any second consumer.

- `src/recap.ts:1218` — the recap ladder's `plan` spec carries no `minUseful`, so `planUsefulFloor`/`plan-unreviewable` (`src/prompt-fit.ts:18-26`) never engages for a recap; a recap's plan can be ground down to `MIN_CLAMP_KEEP`. Pre-existing and unchanged by this diff, but the new spec sitting beside it makes the asymmetry visible.
- `src/recap-core.ts:308-309` — the transcript digest is pushed into the prompt unfenced, unlike `task`, `context` and now `ui-markup`, even though it is the most obviously attacker-influenced input the recap spawn ingests.

scoop:
# Review — recap UI markup (#2209)

Diff under review: `git diff 69d8a5dc7...HEAD` — `docs/sandbox-security.md`, `src/recap-core.ts`,
`src/recap.ts`, `test/recap-core.test.ts`, `test/recap.test.ts`.

## What I verified

**The shape the helper consumes is real.** `DiffFile` (`src/types.ts:443-453`) carries
`status`/`additions`/`binary`/`truncated?`/`hunks`, and `DiffLine.content` is documented as "line
text WITHOUT the leading +/-/space marker" (`src/types.ts:429-434`) — so `afterSide`
(`src/recap-core.ts:199-208`) re-adding `+` / a leading space is correct, not a double prefix.
`DiffLineKind` is exactly `"add" | "del" | "ctx"` (`src/types.ts:427`), so
`filter((l) => l.kind !== "del")` leaves precisely add+ctx.

**The exclusions match what `diff.ts` actually produces.** `truncated` is set — and `hunks` cleared —
both at the per-file cap and at the global cap (`src/diff.ts:125-127`, `src/diff.ts:169-172`,
`src/diff.ts:185-187`), so `!file.truncated && file.hunks.length > 0` (`src/recap-core.ts:232-233`)
does keep a truncated component out rather than emitting a headline with nothing under it. Binary
files are flagged at `src/diff.ts:53-54`, `deleted` status at `src/diff.ts:44-45`.

**The clamp ordering claim holds.** `fitAssembledPrompt` walks `specs` in the caller's order and
stops as soon as `measured <= budget` (`src/prompt-fit.ts:214-225`), so putting `uiMarkup` first in
`src/recap.ts:1215-1220` genuinely makes it the first block sacrificed; `plan`, `changedFiles` and
`context` keep their previous relative order. `uiMarkup` is a `text` spec with no `minUseful`, so
`floorUseful` is 0 and `usefulnessBound` can never be set from it (`src/prompt-fit.ts:260-261`,
`src/prompt-fit.ts:297`) — clamping the markup can never turn a fit into a `plan-unreviewable`
refusal. When `uiMarkup` is `""`, `hi (0) <= lo (256)` and `shrinkSpec` returns `null`
(`src/prompt-fit.ts:263-264`), so the non-UI path is untouched.

**The values wiring is complete.** `fitAssembledPrompt` seeds `values` from every spec
(`src/prompt-fit.ts:203`), and the recap `compose` reads `v.uiMarkup` back out
(`src/recap.ts:1224`), so the new spec is never dropped or left `undefined`.

**Fencing is correct and the directive invariant still holds.** `fenceUntrusted("ui-markup", …)`
(`src/recap-core.ts:301`) scrubs both the nonce and any pre-existing fence token
(`src/untrusted.ts:51`), and `buildRecapPrompt` already emits `UNTRUSTED_CONTENT_DIRECTIVE` exactly
once at `src/recap-core.ts:272` — which is what `test/untrusted.test.ts` pins for any builder that
fences. The interpreting sentence sits after the closing marker (`src/recap-core.ts:302-303`), which
is the posture `src/prompt-fit.ts:30-37` and `src/untrusted.ts:31-44` require. No new tool grant, no
`--add-dir`: `writer-only` is still `Write`-only into a disposable temp dir
(`src/transient-agent-argv.ts:126-131`, `src/transient-agent-argv.ts:173-178`), so the comments in
`src/recap-core.ts:149-157` are factually accurate about the preset.

**Single call site.** `buildUiMarkupDigest` is called once (`src/recap.ts:849`) and
`buildRecapPrompt` only from `src/recap.ts:874`, reached only via `fitRecapPrompt` at
`src/recap.ts:891` — there is no second (e.g. Codex) prompt path left un-threaded.

**Caps arithmetic.** `total` is accumulated from the *rendered* block length
(`src/recap-core.ts:249`), and `room` is recomputed from that accurate total each iteration
(`src/recap-core.ts:244`), so the section can overshoot `maxChars` by at most one elision marker in
total — which is exactly what the comment at `src/recap-core.ts:166-167` claims. `maxChars: 100`
correctly yields `""` (room 76 < the 200 floor at `src/recap-core.ts:171`).

**Docs.** The `docs/sandbox-security.md:100-105` edit is accurate: `ui-markup` is a real fence label
(`src/recap-core.ts:301`) and it joins the same list as `context` (`src/recap-core.ts:315`).

**Task satisfaction.** The commit body reports the acceptance measurement from #2194/#2209
(wireframe 0 → 10 over 10 regenerated UI sessions, median 6.5, max 8). I could not re-run that
measurement — it is an LLM-behaviour result, not something in the repo — so I take it as reported
rather than verified.

## Findings

### 1. `src/recap-core.ts`: a deletion-only view file emits a header with no markup under it

`afterSide` guards *per hunk* — `if (lines.length > 0) parts.push(...)` at `src/recap-core.ts:205` —
but the selection filter has no matching guard on the *file's* rendered result. A view file whose
hunks contain only `del` lines passes every filter at `src/recap-core.ts:227-234` (`status` is
`modified`, not `deleted`; `binary` false; `truncated` unset; `hunks.length > 0`), `afterSide`
returns `""`, and the loop still pushes `` `${header}\n${body}` `` — i.e. the bare string
`ui/Foo.svelte (modified)` followed by nothing (`src/recap-core.ts:246-248`).

Reachable today, on a path that already executes: a `.svelte`/`.tsx` component whose body is emptied
but whose file is kept produces `@@ -1,N +0,0 @@` with only `-` lines and no context lines (git's
default 3 lines of context, `src/diff.ts:318-320`, gives no context when there is no surviving
content), and `parseUnifiedDiff` records it as `status: "modified"`, `additions: 0`, with a non-empty
`hunks` array.

Consequence: `input.uiMarkup?.trim()` is truthy (`src/recap-core.ts:295`), so the prompt renders the
whole section — "UI markup after this change …" plus "This markup is your grounding for a
`wireframe` block: mock up the screen it renders" (`src/recap-core.ts:297-303`) — over a fence that
contains a file path and no markup. Combined with the preamble's standing "`wireframe` is NOT
optional" for a user-visible change (`src/recap-core.ts:357-360`), that is pressure toward inventing
a screen from nothing, which is precisely the failure mode #2209 exists to remove. The author
already treats this exact shape as unacceptable elsewhere: the floor at `src/recap-core.ts:170-171`
and `src/recap-core.ts:245` exists to "stop instead of emitting a stub", and the test at
`test/recap-core.test.ts:1198` asserts "a header with a marker under it is not markup".

Fix is one predicate: render `afterSide(file.hunks)` before deciding to emit, and skip the file (and
therefore the whole section, when it is the only one) when the result is empty.

## Nits (non-blocking):
- `src/recap-core.ts:245` uses `break` rather than `continue`, so once one large file exhausts the section budget a later, shorter-header file that would still fit is silently skipped; harmless but slightly under-fills the section.
- `src/recap-core.ts:178` `TEST_FILE_RE` only matches when `test`/`spec` is the segment immediately before the extension, so `Card.test.stories.tsx` is still treated as a view file.
- `src/recap-core.ts:161` exports `RECAP_UI_MARKUP_MAX_FILES` solely for the test while its two sibling caps stay module-private; the asymmetry reads as accidental.
- `src/recap-core.ts:187-189` `elidedChars` ("chars elided") deliberately differs from `elisionMarker` ("bytes elided") in `src/prompt-fit.ts:40-42` so the clamp-order test at `test/recap.test.ts` can tell them apart — worth a one-line note, since the two markers otherwise look like a copy-paste divergence.

- `src/recap-core.ts:219` adds a `limits` override parameter that no production call site uses (`src/recap.ts:849` passes only `diff.files`); it exists purely so the tests can shrink the caps.
- `docs/sandbox-security.md` is not in the plan's Files table, though the one-line addition is accurate and proportionate.

## Stated limitations
- The acceptance measurement in the commit body ("wireframe 0 → 10, median 6.5, max 8") is an LLM-behaviour result produced against a snapshot of the live DB; it is not reproducible from this repo and I did not attempt to re-run it. Assumed as reported.
- Whether a hunk fragment (rather than a whole file body) is *sufficient* grounding for a good mockup is a model-behaviour question, not a code question; the plan names it as the assumption under test and I did not second-guess it.

scoop (commented):
# Review — recap UI-markup injection (#2209), rework round 2

Diff under review: `git diff 02ce8fae0dbe7ad40620bc4b877041a263346fbe...HEAD` — `docs/sandbox-security.md`,
`src/recap-core.ts`, `src/recap.ts`, `test/recap-core.test.ts`, `test/recap.test.ts`.

## Prior finding — status

**Prior 1 (bare `path (status)` header over an empty fence for an emptied-but-kept view file): ADDRESSED.**
`buildUiMarkupDigest` now renders the post-change side *before* selection and makes emptiness part of
the selection predicate: `src/recap-core.ts:226` maps each file to `{ file, order, after: afterSide(file.hunks) }`,
and the filter at `src/recap-core.ts:232-239` includes `after.trim() !== ""` alongside `isViewFile`,
`status !== "deleted"`, `!binary`, `!truncated`. Because the filter runs *before* `.slice(0, maxFiles)`
at `src/recap-core.ts:241`, a gutted file also does not consume one of the three slots — the exact
two-part fix requested. Both halves are pinned by the test at `test/recap-core.test.ts:1153-1171`
(`gutted` alone ⇒ `""`; `gutted` plus three real files with `maxFiles: 3` ⇒ `ui/e.svelte` absent and
`f/g/h` all present). No path remains by which `uiMarkup.trim()` is truthy while the fence body is empty.

## Verification performed

- **Types resolve.** `DiffFile`, `DiffFileStatus`, `DiffHunk` are all exported from `src/types.ts:443`,
  `src/types.ts:441`, `src/types.ts:436`; the `truncated?: boolean` and `hunks: DiffHunk[]` fields the
  predicate reads exist at `src/types.ts:450-451`, and `DiffLine.kind` is the `"add" | "del" | "ctx"`
  union at `src/types.ts:427` that `afterSide` (`src/recap-core.ts:203-204`) switches on. `additions`,
  used for ranking at `src/recap-core.ts:240`, is `src/types.ts:447`.
- **The diff actually carries hunks on this path.** `RecapService.generate()` calls
  `buildUiMarkupDigest(diff.files)` at `src/recap.ts:849` on `source.diff`, which comes from
  `this._computeDiff` (`src/recap.ts:546`, defaulted to `computeDiff` at `src/recap.ts:438`).
  `computeDiff` returns `parseUnifiedDiff(stdout)` — structured hunks, not `toSessionDiff` —
  at `src/diff.ts:324-332`. So the input the helper needs is genuinely present, and no new I/O was added.
- **No mutation of the caller's array.** `files.map(...).filter(...).sort(...)` at `src/recap-core.ts:225-241`
  sorts the array produced by `map`, not `diff.files`, so the later `changedFiles` /
  `changedFilesWithStatus` derivations at `src/recap.ts:844-845` (computed before line 849 anyway)
  and the downstream `pendingDiff` join are unaffected.
- **Clamp ladder order.** `fitAssembledPrompt` iterates `specs` in caller order
  (`src/prompt-fit.ts:214-225`), so putting `{ id: "uiMarkup", … }` first at `src/recap.ts:1217`
  really does make it the first block sacrificed; `plan` (`head-tail`) and `context` keep the positions
  they had. When `uiMarkup` is `""`, `shrinkSpec` returns `null` immediately —
  `hi = fromBytes = 0 <= lo = MIN_CLAMP_KEEP` at `src/prompt-fit.ts:260-264` — so a non-UI session's
  ladder behaves exactly as before. The regression test at `test/recap.test.ts:2607-2659` asserts
  exactly one elision marker and that it lands between the `ui-markup` fence delimiters.
- **Fence contract.** `fenceUntrusted("ui-markup", …)` at `src/recap-core.ts:305` matches the
  signature at `src/untrusted.ts:46-53` (label + nonce only, content-scrubbed). `buildRecapPrompt`
  already emits `UNTRUSTED_CONTENT_DIRECTIVE` once at `src/recap-core.ts:276`, so the
  "a builder that fences must carry the directive" invariant described at `src/untrusted.ts:42-44`
  still holds. The interpreting instruction is emitted *after* the closing marker
  (`src/recap-core.ts:306-307`), which is correct per the reasoning at `src/prompt-fit.ts:30-37` —
  an in-fence instruction is contractually ignorable and forgeable. Pinned by
  `test/recap-core.test.ts:1224-1232`.
- **Guidance region untouched.** The new section is pushed at `src/recap-core.ts:300-310`, before the
  `"Optionally, add \`blocks\`"` line at `src/recap-core.ts:334` — i.e. outside the region
  `blockGuidanceRegion` slices (`test/recap-core.test.ts:988-994`, from that line to
  `"When done, write these files in your CWD:"` at `src/recap-core.ts:396`). The 7424-byte ceiling
  pinned at `test/recap-core.test.ts:1071` is therefore unmoved, and `test/recap-core.test.ts:1262-1275`
  asserts byte-equality against the bare prompt.
- **Selection predicate spelling.** `VIEW_FILE_RE` (`src/recap-core.ts:175`) is anchored with `$`, so
  `Card.svelte.md` correctly fails; `TEST_FILE_RE` (`src/recap-core.ts:178`) matches `.test.tsx` /
  `.spec.jsx` / `.browser.test.ts`. Both behaviours are pinned at `test/recap-core.test.ts:1105-1128`.
- **Callers of the changed signature.** `uiMarkup` is optional on `buildRecapPrompt`
  (`src/recap-core.ts:268`); the only production caller is `src/recap.ts:874`, which passes it. Every
  other call site is in `test/recap-core.test.ts`, so no consumer breaks. `hostArgvBudget` and
  `joinedElementBytes`, newly imported by `test/recap.test.ts:10`, exist at `src/argv-limit.ts:40`
  and `src/argv-limit.ts:58`.
- **Injection posture.** The markup is repo content the session agent may itself have authored, and it
  is now fenced under a label the agent is told to mock up. The blast radius is bounded server-side
  independently of the prompt: `validateWireframe` (`src/visual-blocks.ts:498-513`) rejects
  `<script>` / `<style>` / `on*=` / `href=` (`src/visual-blocks.ts:452-458`), raw colours in `style=`
  and in `fill=`/`stroke=` (`src/visual-blocks.ts:461-496`), a bad `surface`, and oversized HTML. So
  markup copied verbatim into a wireframe cannot smuggle script or a link into the recap renderer.
  No tool grant, preset, `--add-dir` or membrane change is present in the diff — option 1 was correctly
  not taken.
- **Docs claim.** `docs/sandbox-security.md:104` names the fence label `ui-markup`, which matches the
  literal at `src/recap-core.ts:305`.
- **Budget arithmetic.** `room = Math.min(maxFileChars, maxChars - total - header.length - 2)` at
  `src/recap-core.ts:249` with the `< MIN_UI_MARKUP_FILE_CHARS` bail at `src/recap-core.ts:250`;
  the section can overshoot `maxChars` only by the per-file marker `clipMarkup` appends
  (`src/recap-core.ts:191-194`), which the comment at `src/recap-core.ts:166-167` states explicitly.
  Even at worst that is ~6.03 KB, and `fitRecapPrompt` clamps it first regardless.

No blocking defect found.

## Stated limitations

- The issue's acceptance bar ("regenerate a sample of UI-change sessions and show `wireframe > 0`,
  median 3-7, max ≤ 14") is an LLM-behaviour measurement. It cannot be verified from this worktree —
  the harness and DB copy are deliberately uncommitted (scratchpad), and the PR body is not visible to
  me. I neither confirm nor dispute the measurement; I judged only the mechanism, which is sound and
  is the input the issue identified as missing.
- I did not run lint or the test suite (CI handles that, and this review is read-only).

## Nits (non-blocking):

- `src/recap-core.ts:250` — the per-file budget loop `break`s when `room` falls below the floor, but `room` depends on `header.length`, which is not monotonic across the selected files; a later file with a much shorter path could still have fit. `continue` would be marginally more thorough, and with `maxFiles === 3` the difference is unobservable in practice.
- `src/recap-core.ts:192` — `clipMarkup` slices by UTF-16 code units, so a cut can land inside a surrogate pair and emit a lone surrogate (rendered as U+FFFD once encoded for argv). Harmless, and it matches the existing precedent at `src/recap-core.ts:138`, where `buildTranscriptDigest` slices the same way — noted only because `src/prompt-fit.ts:52-59` shows the repo does guard this elsewhere.
- `src/recap-core.ts:302-304` — the section header describes the `+` / leading-space convention, but for an `added` view file every line is `+`, which reads slightly oddly as "a line this session added" against a whole new file. No behavioural consequence.

PR #2216, opened 2026-09-08
feat(herdr): recover incomplete updates with explicit confirmation

kai-osthoff:
A herdr update could install the new CLI while leaving the old server running after a failed live handoff. Shepherd then reported success and task creation failed with raw protocol errors. Updates now require matching installed/server versions and a working agent call before reporting success.

The existing update dialog explains the incomplete handoff and offers an explicit, local restart with a warning about affected terminal processes, progress, and verification. Diagnose and failed task creation open the same recovery flow. Task drafts and attachments remain mounted and return with focus; recovery never resubmits the task automatically.

Recovery rechecks the confirmed versions, shares the maintenance guard with updates and diagnostic fixes, and addresses only the configured binary/socket. Its bounded detached worker survives the initiating process, respects matching service supervision, and stops on failed or ambiguous guards. Transient HTTP/WS failures and reopening the dialog retain actionable status.

The duplication baseline adds only one unchanged modal CSS group surfaced by removing the unused log-label style; new runtime and dialog logic clears the existing complexity thresholds.

Validation after rebasing onto main: `bun run lint`, `bun run typecheck`, root `bun run test` (9,216 passed; 15 skips), UI `bun run check` and `CI=true bun run test` (4,874 passed), `bun run check:i18n`, and `bun run check:announcement-versions`, and the delta `fallow audit`. An isolated real 0.8.2 server with a 0.9.0 CLI was detected as requiring restart and recovered to a verified 0.9.0 server. The production herdr daemon was not changed.


kai-osthoff (commented):
## Review summary

Comment — no blocking defects found in the reviewed diff.

## Verified

- Update success requires the requested installed version plus a ready runtime whose installed and server versions both equal that target; a binary-only update is therefore not reported as successful (`src/herdr-update.ts:521`). A failed handoff is retained as `restart_required` and exposes the observed server version (`src/herdr-update.ts:529`).
- The repair endpoint requires an explicit confirmation and typed displayed-version snapshot before it calls the service (`src/server.ts:4732`). The normal application dispatcher applies its existing authentication and origin checks before any route handler (`src/server.ts:8062`).
- The recovery service re-probes before mutation, rejects an unknown runtime, and verifies a started server through a ready state with equal non-null installed/server versions (`src/herdr-recovery.ts:443`, `src/herdr-recovery.ts:332`).
- The dialog keeps the task composer mounted while repair is visible, makes it inert, and passes the repair action back to the same dialog; this preserves the composer state rather than remounting it (`ui/src/lib/components/page/AppOverlays.svelte:545`). The task-create protocol-conflict code is handled separately and presents that repair action (`ui/src/lib/components/NewTask.svelte:1640`).
- The added message keys have EN/DE counterparts with the same key set, including repair, diagnostics, and composer messages (`ui/messages/en.json:3842`, `ui/messages/de.json:3842`).

## Limitation

I did not run tests, builds, or commands that mutate state, as requested. The live upstream `herdr status --json` schema is not present in this repository, so its compatibility was not independently verified here.

kai-osthoff:
🚨✅ READY! 🎉🙃 Scoop, your heroic contribution awaits: doing your job. 👀🫡🚀🔥

PR #2226, opened 2026-09-09
fix(onboarding): self-heal a leaked host lock and alert when the nightly stops

scoop:
The nightly onboarding harness stopped running on **Aug 20** and aborted in under a second for **21 consecutive nights**. Nothing anywhere reported it. The first thing to notice was the onboarding release gate failing on release-please PR #2127, three weeks later — and that gate was working correctly; the nightly was not.

## What actually happened

1. **Aug 20** — the run hit the service's `TimeoutStartSec=2h` and systemd SIGTERM'd it.
2. Its cleanup never ran, leaving behind a **0-byte lock file** and a **running Incus container**.
3. Every night since: `another onboarding-harness run holds …lock; aborting`, exit 3. No status posted, no issue filed, no alert.
4. The last `onboarding-harness` commit status in existence was from **Aug 19**.

The slowdown that triggered it is not a code regression. Nightly wall-clock has been swinging for weeks (9–15 min normally; ~61 min on Aug 16/17/18; >2h on Aug 20) while the harness itself burns only 8–11s of CPU — the time is in-container package installs, so it tracks network weather.

## Three defects, all fixed

**1. One leaked lock was permanently fatal.** The lock was existence-based — `open(…, "wx")`, zero bytes, no owner recorded — so "file exists" was the entire staleness test. It now records `{pid, startedAt, runId}`, and a run **reclaims** a lock whose owner is provably gone: a dead pid, a pid since recycled by an unrelated process, or a file with no readable owner (the exact Aug 20 file). A genuinely live owner still exits 3.

A reclaiming run also sweeps every `shep-onb-*` instance, because it has inherited the host from a dead run and **nothing else ever will** — the Aug 20 container survived 21 days and was restarted by incus autostart on every reboot.

**2. Cleanup-on-signal cannot be relied on — and this is the interesting part.** I reproduced the leak under a real transient systemd unit and narrowed it to one condition:

| kill path | handler fires? | lock |
|---|---|---|
| `systemctl stop`, idle | yes | released |
| `systemctl stop`, via `bun run <script>` wrapper | yes | released |
| start-timeout, idle | yes | released |
| **start-timeout while awaiting a spawned child** | **no** | **LEAKED** |

Only the last leaks, and it reproduces the Aug 20 journal signature byte for byte (`terminated by signal SIGTERM` / `code=killed, status=15/TERM`, versus `exited with code 130` in the healthy cases). **While Bun awaits a spawned child, SIGTERM never reaches the JS handler.** The harness awaits `incus` children for nearly its entire runtime, so a start-timeout kill lands in that state every time — this wasn't bad luck, it was the only possible outcome.

So the harness now bounds *itself* (30m per scenario, 3h per run, both env-overridable) and `TimeoutStartSec` moves 2h → 5h. The point is the ordering: **systemd's timeout must sit strictly above the harness's own worst case**, or its dirty kill is guaranteed to win. The old 2h sat *below* it. Scenarios that don't finish are recorded **NOT VERIFIED** — not green, not a harness error — so a gate-eligible one still gates red and the report says plainly that no verdict was reached, instead of the run vanishing.

The signal handler is kept and now also sweeps, but documented honestly as best-effort: it works for an interactive Ctrl-C and will not fire in the case above.

**3. Nothing watched the watcher.** New `onboarding_stale` alert, mirroring the existing `backup_stale` probe exactly. The harness stamps a liveness marker on every completed run — written for red runs too, since a red run already shouts through the rolling issue and a red commit status, whereas a harness that stops running says nothing. The daily sweep warns to the log, writes a durable signal row, and sends a push when a host that has the timer installed hasn't completed a run in **30h** — one missed night, deliberately well inside the release gate's 48h max-age so a human hears about it *before* releases are blocked.

Gating is on the real systemd timer unit's presence, so a laptop or core-only box stays silent while the Incus host that genuinely stopped gets flagged. A host with the timer and no marker at all is flagged too — "it has never once completed a run" is worth shouting about.

## Verification

- Root `bun run lint` + `bun run test` green (9446 pass, 0 fail; 24 new tests).
- `ui/ bun run check` green (0 errors) — no UI surface, matching `backup_stale`.
- `bunx fallow audit --base origin/main --fail-on-issues` green. `buildPayload` breached the complexity threshold when my case was added, so both host-global staleness kinds now share one switch arm via a `stalenessParts` helper, following the existing `learningsParts` pattern — net complexity below where it started.
- The unit change is installed and live on the host (`TimeoutStartUSec=5h`).

## Already done on the host

The orphan container was destroyed and the stale lock cleared (`--reap-orphans`), so **tonight's 03:30 nightly runs for the first time since Aug 20** regardless of when this merges.

## Still needs a human

The release gate on #2127 stays red until a fresh `onboarding-harness` status lands on a recent main commit. Tonight's nightly will produce one for free; no manual run needed unless you want it sooner.

## Follow-up worth filing

Pre-baking the scenario images would cut both the runtime and this entire class of variance at the source.

### DIFF
--- a/ci/onboarding-harness/README.md
+++ b/ci/onboarding-harness/README.md
@@ -84,9 +84,21 @@ bun run onboarding:test --reap-orphans
 
 The report lands at `onboarding-gap-report.md` in the working directory. Exit code is 0 if every apply-able scenario reached green; 1 if any failed; 2 if the named scenario was not found; 3 if another run holds the host lock.
 
+## Time budget
+
+The harness bounds itself so it always finishes on its own terms — **30 min per scenario** (`SHEPHERD_ONBOARDING_SCENARIO_TIMEOUT_MS`) and **3h per run** (`SHEPHERD_ONBOARDING_BUDGET_MS`). A scenario that blows its cap is abandoned; scenarios with no budget left are never started. Either way they are recorded **NOT VERIFIED** — not green, not a harness error, so a gate-eligible one still gates red and the report says plainly that no verdict was reached.
+
+Those caps are generous against observation: a healthy full run is ~14 min, a slow night ~61 min, and the worst single scenario ever seen took 26 min. Wall-clock is dominated by in-container package installs, so it tracks network weather rather than anything the harness controls.
+
+The service's `TimeoutStartSec` (5h) is a **backstop that must never be reached**, and must always stay above the harness's own budget. When it sat _below_ the worst case (the old 2h), systemd's dirty kill was the guaranteed outcome on a slow night rather than an edge case.
+
 ## Run isolation
 
-Each run acquires an exclusive host-wide lock at the **fixed absolute path** `~/.shepherd/onboarding-harness.lock` (NOT `$TMPDIR` — a systemd-user timer and an interactive run can see different `$TMPDIR` under PrivateTmp, so the lock must live under the stable `~/.shepherd` dir to actually serialize them). A second run while one is active exits with code 3.
+Each run acquires an exclusive host-wide lock at the **fixed absolute path** `~/.shepherd/onboarding-harness.lock` (NOT `$TMPDIR` — a systemd-user timer and an interactive run can see different `$TMPDIR` under PrivateTmp, so the lock must live under the stable `~/.shepherd` dir to actually serialize them). A second run while one is _genuinely active_ exits with code 3.
+
+The lock records its owner (`pid`, `startedAt`, `runId`), and a run **reclaims** a lock whose owner is provably gone — a dead pid, a pid since recycled by an unrelated process, or a file recording no readable owner. A reclaiming run also sweeps every `shep-onb-*` instance, since it has inherited the host from a dead run and nothing else will ever clear that run's orphans.
+
+This matters because cleanup-on-signal cannot be relied on: **while Bun awaits a spawned child, SIGTERM never reaches the JS handler** and the process dies with its `finally` blocks unrun. The harness awaits `incus` children for almost its entire runtime, so a killed run leaks by default. In Aug 2026 a run killed by the service's start timeout left a 0-byte lock behind and every nightly for the next 21 days aborted at exit 3 in under a second — silently, until a release gate noticed three weeks later; the instance it also leaked survived those 21 days and every reboot. The `SIGINT`/`SIGTERM` handler is kept as a best-effort extra (it does work for an interactive Ctrl-C), but reclaiming is the actual guarantee.
 
 Each run also uses a unique per-run instance prefix (`shep-onb-<runId>-`) and sweeps only its own prefix on teardown, so two concurrent runs (if the lock were bypassed) could never destroy each other's instances.
 
--- a/ci/onboarding-harness/lock.ts
+++ b/ci/onboarding-harness/lock.ts
@@ -1,0 +1,127 @@
+import { closeSync, mkdirSync, openSync, readFileSync, unlinkSync, writeSync } from "node:fs";
+import { homedir } from "node:os";
+import { dirname, join } from "node:path";
+
+// FIXED absolute path under the host-global Shepherd state dir (~/.shepherd) — NOT
+// $TMPDIR. A systemd-user timer service and an interactive shell can see different
+// $TMPDIR (PrivateTmp, per-session dirs), which would defeat the lock; $HOME is
+// stable and identical for both (same user), so the lock is genuinely host-wide.
+export const LOCK_PATH = join(homedir(), ".shepherd", "onboarding-harness.lock");
+
+/** Who holds the lock — recorded so a later run can tell "held" from "leaked". */
+interface LockOwner {
+  pid: number;
+  startedAt: string;
+  runId: string;
+}
+
+/** Thrown when the lock is held by a run that is genuinely still alive. The
+ *  caller maps this to exit 3 (the documented "another run holds the host" code). */
+export class LockHeldError extends Error {}
+
+/** Is `pid` still a live harness run? Liveness alone is not enough: pids are
+ *  recycled, and a recycled pid must not lock the harness out forever, so a live
+ *  pid is additionally checked against /proc for the harness in its command line.
+ *  If /proc is unreadable we stay conservative and call it live — refusing to run
+ *  is recoverable, stealing a lock from a running harness is not. */
+function isLiveHarness(pid: number): boolean {
+  try {
+    process.kill(pid, 0);
+  } catch (err) {
+    // ESRCH ⇒ no such process. EPERM ⇒ alive but owned by another user.
+    return (err as NodeJS.ErrnoException).code === "EPERM";
+  }
+  try {
+    return readFileSync(`/proc/${pid}/cmdline`, "utf8").includes("onboarding-harness");
+  } catch {
+    return true;
+  }
+}
+
+function readOwner(path: string): LockOwner | null {
+  try {
+    const owner = JSON.parse(readFileSync(path, "utf8")) as Partial<LockOwner>;
+    return typeof owner.pid === "number" ? (owner as LockOwner) : null;
+  } catch {
+    return null; // absent, empty (the pre-owner format), or corrupt
+  }
+}
+
+/** `wx` — create-exclusive — so acquiring is atomic against a concurrent run. */
+function tryCreate(path: string, runId: string): number | null {
+  let fd: number;
+  try {
+    fd = openSync(path, "wx");
+  } catch {
+    return null;
+  }
+  const owner: LockOwner = { pid: process.pid, startedAt: new Date().toISOString(), runId };
+  writeSync(fd, JSON.stringify(owner));
+  return fd;
+}
+
+/**
+ * Acquire the host-wide exclusive lock so concurrent runs never share the Incus
+ * host. Returns an idempotent release fn.
+ *
+ * The lock records its owner, and an existing lock is reclaimed when that owner is
+ * provably gone. Existence alone used to mean "held", which made a single leaked
+ * file permanently fatal: in Aug 2026 a run killed by the service's start timeout
+ * left a 0-byte lock behind and every nightly for the next 21 days aborted at exit
+ * 3 in under a second — silently, until a release gate noticed three weeks later.
+ *
+ * Release is also wired to SIGINT/SIGTERM by the caller, but that path CANNOT be
+ * relied on, and not merely in theory: while Bun awaits a spawned child, SIGTERM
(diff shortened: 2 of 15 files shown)


scoop (commented):
# Review — `fix(onboarding): self-heal a leaked host lock and alert when the nightly stops`

## Verdict

The change does what the task asked. The reported symptom ("is the nightly job no longer running?") is addressed on all three fronts a silent-outage bug needs: the leaked lock is now **reclaimable** (`ci/onboarding-harness/lock.ts`), the harness now **bounds itself** so systemd's dirty kill is not the thing that ends a slow night (`ci/onboarding-harness/run.ts:421-423`, `ci/onboarding-harness/shepherd-onboarding.service:19`), and the server now **alerts** when no run has completed in 30h (`src/index.ts:2470-2505`). No blocking defect found. Nits below.

**The lock's exit-3 contract still holds.** `acquireHostLock` throws `LockHeldError` only when the recorded owner is live (`ci/onboarding-harness/lock.ts:95-98`) or when the post-unlink re-create loses a race (`lock.ts:108-110`); `main` maps exactly that class to exit 3 and rethrows anything else (`run.ts:497-501`). That matches the documented code in `ci/onboarding-harness/README.md:7` and the updated wording at `README.md:20`.

**An unverified scenario really does gate red.** `gateGapScenarios` filters `gateEligible && !reachedGreen && !isHarnessError` (`ci/onboarding-harness/report.ts:76`). `isHarnessError` requires `error.startsWith(LAUNCH_FAILURE_PREFIX)` = `"incus launch failed: "` (`report.ts:21`, `ci/onboarding-harness/incus.ts:8`); the two strings `unverifiedResult` produces — `"exceeded its …m cap — abandoned"` (`run.ts:454`) and `"not run — the run budget was exhausted"` (`run.ts:543`) — do not start with it, so an unverified gate scenario stays in the gap list. `classify` short-circuits on `r.unverified` before the `installE2E`/`isPreDetectionThrow` arms (`report.ts:54`), so it cannot be mislabelled BOOT CRASH. Both `install-e2e` and `install-e2e-service` are `coaching: "structured"` and not `detectionOnly` (`ci/onboarding-harness/scenarios.ts:172`, `:196`), so `scenarioBase` gives them `gateEligible: true` (`run.ts:308`) and a timed-out installer scenario blocks the release. Confirmed by the new tests at `test/onboarding-harness/report.test.ts:326-346`.

**No unhandled rejection from the abandoned half of `Promise.race`.** `runScenarioBounded` (`run.ts:445-461`) leaves the losing `runScenario` promise unobserved, so I checked whether it can reject: every failure inside `runScenario`'s body is caught and converted to a result (`run.ts:345-357`), and the `finally` calls `driver.delete()` (`run.ts:359`), which awaits `this.run(["delete", …])` and ignores the exit code (`incus.ts:77-79`) over `captureSpawn`, which `resolve`s on `error` and `close` and never rejects (`ci/onboarding-harness/spawn.ts:16-36`). So the abandoned promise always fulfils. Verified, not assumed.

**Locale parity for the new push copy.** `NotifyText` is `(typeof NOTIFY_TEXT)[NotifyLocale]` (`src/push.ts:257`), i.e. the union of both catalogs, so `t.onboardingStaleTitle` / `t.onboardingStaleBody` at `src/push.ts:300` only type-check if both locales carry them. They do: en at `src/push.ts:187-191`, de at `src/push.ts:244-248`. `test/push.test.ts:1272-1293` asserts both, including the `staleHours: undefined` "never completed a run" branch.

**Every new identifier resolves.** `onboarding_stale` is added consistently as a `SignalKind` (`src/types.ts:1327`), a `PushPayload["kind"]` and `NotifyInput["kind"]` (`src/push.ts:29`, `:79`), a `KIND_CATEGORY` entry (`src/push.ts:55`), a `REDUCED_ALLOWED` member (`src/push.ts:132`), and a `NON_LEARNING_SIGNAL_KINDS` member (`src/distiller.ts:33`) — the same five registration sites `backup_stale` occupies. I grepped for further `SignalKind` consumers (`src/store.ts:4236`, `src/learnings-service.ts:20`, `src/maintain-core.ts:53`) and none needs a per-kind entry; `maintain`'s incident band needs distinct sessions, and this signal is written with `sessionId: null` (`src/index.ts:2489`), exactly like `backup_stale` (`src/index.ts:2432`).

**The timer-unit path the probe stats is the one the docs install.** `onboardingTimerUnit` returns `$HOME/.config/systemd/user/shepherd-onboarding.timer` (`src/onboarding-paths.ts:36`), and the install instructions are `cp ci/onboarding-harness/shepherd-onboarding.{service,timer} ~/.config/systemd/user/` (`ci/onboarding-harness/README.md:112`). The unit file exists at `ci/onboarding-harness/shepherd-onboarding.timer`.

**The comment about "the release gate's 48h max-age" is factually true**, not folklore: `ci/onboarding-harness/README.md:140` documents that the gate requires a fresh `onboarding-harness` commit status younger than 48h. So `ONBOARDING_STALENESS_MS = 30h` (`src/index.ts:2452`) really is the tighter of the two, as `src/index.ts:2449-2451` claims.

**Importing `run.ts` from a test still has no side effects.** Three tests import `runScenario` from it (`test/onboarding-harness/install-e2e.test.ts:5` and siblings); the new imports it pulls in — `./lock` and `../../src/onboarding-paths` — are both pure module bodies, and `main()` is still behind `import.meta.main`.

## Stated limitations (could not verify from this repo)

- `isLiveHarness` decides "is this pid really a harness?" by looking for the substring `onboarding-harness` in `/proc/<pid>/cmdline` (`ci/onboarding-harness/lock.ts:35`). Whether that substring is present depends on Bun's process model for `bun run <package.json script>`: the service runs `bun run onboarding:test` (`ci/onboarding-harness/shepherd-onboarding.service:6`) and that script expands to `bun run ci/onboarding-harness/run.ts` (`package.json:17`). If Bun spawns the expansion as a child (which I believe it does), the JS process's argv contains the path and the check works; if Bun ever ran it in-process, the argv would be `bun run onboarding:test` and a **live** run would read as dead. I cannot settle Bun's spawn behaviour by reading this repo and did not run anything, so I am neither calling it correct nor raising it as a finding — flagging it because the consequence of a false "dead" verdict is a second run stealing the lock and then cross-prefix-sweeping the live run's instances (`run.ts:528-531`).

- `src/onboarding-paths.ts:13` adds a `SHEPHERD_STATE_DIR` env override that nothing in the repo sets or reads except its own test (`test/onboarding-paths.test.ts:14-18`) — speculative config for a path that otherwise has exactly one anchor; note also that `ci/onboarding-harness/lock.ts:42` anchors the *lock* on `homedir()` and honours no such override, so the two files in `~/.shepherd` can be relocated independently.
- `ci/onboarding-harness/shepherd-onboarding.service:21` adds `TimeoutStopSec=5min`, which the task did not ask for; harmless and well-motivated by the new signal-handler sweep, so keep it if you like it.

- "Mirror every type/contract change across `src/types.ts` AND `ui/src/lib/types.ts`" — `src/types.ts:1327` gains `onboarding_stale` with no UI counterpart. I judge the rule's intent met rather than violated: `ui/src/lib/types.ts:2494` deliberately declares the narrower *learnings-evidence* subset (`"reply" | "critic" | "block" | "stall"`) and already omits `egress_drop`, `backup_stale`, `injection_detected` and the rest, while the wire field the UI actually reads is typed `kind: string` (`ui/src/lib/types.ts:1530`). Nothing diverges.

## Notes (no action implied)

- `--reap-orphans` still unlinks the lock unconditionally (`run.ts:475-480`), including one a genuinely live run holds. That block is unchanged by this PR and the README scopes it to manual recovery, so I am not raising it — but now that reclaiming covers the common case, that escape hatch is doing less work than it used to and could reasonably become owner-aware later.
- When the run budget is nearly spent, the per-scenario cap is `Math.min(SCENARIO_TIMEOUT_MS, remaining)` (`run.ts:548`), so `Math.round(capMs / 60_000)` (`run.ts:454`) can render "exceeded its 0m cap — abandoned" in the report. Cosmetic only.


**Nits (non-blocking):**

- [bug] ci/onboarding-harness/lock.ts: reclaiming is unlink-then-recreate (lock.ts:104-110), which is not atomic. Two runs that both observe the same stale lock can each unlink the other's freshly created lock and both proceed — and both then run the cross-prefix `shep-onb-` sweep at run.ts:530, which is exactly what the per-run instance prefix exists to prevent. An atomic swap (create a temp file next to the lock, then link/rename it into place, and treat EEXIST as lost-the-race) closes the window without changing the observable contract.
- [bug] ci/onboarding-harness/run.ts: the new SIGINT/SIGTERM handler (run.ts:509-519) sweeps and exits but never stops the main loop, which keeps running scenarios and launching instances while `driver.sweep()` awaits its one-shot `listManaged()` snapshot. Anything launched after that snapshot survives `process.exit(130)` — and because `release()` removed the lock cleanly, the next run sees `reclaimed === false` and skips the cross-prefix sweep at run.ts:528-531, so those orphans persist until a manual `--reap-orphans`. An abort flag the loop checks before each scenario would make the handler's sweep terminal.
- [bug] ci/onboarding-harness/run.ts: `unverifiedResult` (run.ts:429-437) builds from `scenarioBase`, which carries no `detectionOnly`, so a detection-only scenario that blows its cap or loses the budget lands in `buildGapReport`'s `applicable` denominator (report.ts:121) and in the non-green count, contradicting the documented rule that detection-only scenarios are excluded from the tally. Four scenarios are affected (scenarios.ts:54, :70, :81, :118). Copying `detectionOnly: scenario.detectionOnly || undefined` into the unverified result keeps the headline ratio honest; classification is unaffected either way.
- [bug] src/index.ts: `checkOnboardingStaleness` (index.ts:2470-2482) has no grace window, unlike the sibling it is modelled on — `checkBackupStaleness` suppresses the alert while the configured-marker is younger than the threshold (index.ts:2411-2415). Since this PR introduces the marker, the first sweep after it deploys will fire a durable signal plus a push saying the harness has never completed a run on every host that has the timer, until the next 03:30 nightly. The `stat()` at index.ts:2472 already returns the unit's `mtimeMs`; gating on `Date.now() - st.mtimeMs >= ONBOARDING_STALENESS_MS` would reuse the sibling's shape for free.
- [bug] ci/onboarding-harness/run.ts: the 30-minute per-scenario cap (run.ts:421-422) sits only ~15 percent above the worst single scenario the README records as actually observed (26 min, README.md:13). Because an over-cap scenario is gate-eligible and gates red (verified via report.ts:76), a merely slow night on an install-heavy scenario turns into a blocked release — the same class of false red this PR is trying to end. 45m would still be far below the 3h run budget and the 5h unit backstop.

Issue #2229, opened 2026-09-10
Onboarding harness runtime has degraded ~16x (14 min → ~4h)


scoop:
The nightly onboarding harness has gone from **14 minutes to nearly 4 hours** for the same ten scenarios. It still passes; it just takes ~16x longer. That is now the harness's dominant operational problem, and it is what made the Aug 2026 outage (#2226) possible in the first place — the run only blew its `TimeoutStartSec` because it had become slow enough to.

## Evidence

Total wall-clock per nightly, from the journal:

| Date | Total |
| --- | --- |
| Aug 10–15 | 9–15 min |
| Aug 16, 17, 18 | ~61 min |
| Aug 19 | 14 min |
| Aug 20 | >2h (killed by the 2h timeout) |
| **Sep 10** | **~3h45m** (9 of 10 scenarios in 3h03m) |

Per-scenario on Sep 10 — every scenario, not one outlier:

```
gh-unauthed        26 min      claude-missing     22 min
gh-missing         28 min      git-missing        11 min
tailscale-missing  26 min      node-too-old       28 min
herdr-missing      20 min      install-e2e        running
herdr-outdated     20 min
```

These same scenarios took 1–2 minutes each on Aug 19.

## What it is not

Measured on the host while the Sep 10 run was live:

- **Not the harness's own CPU** — it burns 8–11s of CPU across an entire run, on every night, fast or slow.
- **Not host CPU contention** — 31 cores, 92–97% idle, `vmstat` runnable queue 0–3. The load-average spikes track the harness's own container bursts.
- **Not disk** — 11% used, 0 iowait.
- **Not the host's network** — 17.4 MB/s from `deb.debian.org`, 45ms connect.

So the time is being spent **inside the containers**, on package installs and image pulls, while the host itself is idle and well-connected. The most likely culprit is the container network path (incus bridge NAT/DNS) rather than anything the harness controls — worth confirming by measuring in-container throughput directly, which I did not do while a run was live to avoid disturbing it.

Note also that **every run re-downloads every image**: the image aliases' cached fingerprints are replaced on each nightly (the remote publishes daily builds, so `images:ubuntu/24.04` resolves to a new fingerprint each night). That is baseline cost on top of whatever is slowing the containers down.

## Proposal

[...]

PR #2230, opened 2026-09-10
fix(onboarding): size the run budget to the harness's real runtime

scoop:
Follow-up to #2226, which merged while this session was still running. It fixes a defect in that PR's numbers — mine.

## The problem

#2226 gave the harness a self-imposed budget so it always terminates on its own terms instead of being killed dirty by systemd. That mechanism is right. **The numbers were not.** I sized them (3h per run, 30 min per scenario) against the harness's *historical* runtime of ~14 minutes, with the worst single scenario ever observed at 26 minutes.

The **Sep 10 nightly** — the first real run since Aug 20 — took **~4 hours** for the same ten scenarios, with *every* scenario in the 20–28 minute band:

```
gh-unauthed        26 min      claude-missing     22 min
gh-missing         28 min      git-missing        11 min
tailscale-missing  26 min      node-too-old       28 min
herdr-missing      20 min      install-e2e        (9th of 10, still running at 3h03m)
herdr-outdated     20 min
```

Under #2226 as merged, that healthy, working run would have been **cut off at 3h with scenarios reported NOT VERIFIED — a red release gate on a harness that was doing its job.** I would have replaced a nightly that fails silently with one that fails loudly and wrongly, which is not obviously an improvement.

It is not live yet: the host checkout hasn't pulled main, so tonight ran the old unbounded code under the new 5h backstop and finished fine. The next deploy is what would arm it.

## The change

Budget 3h → **6h**, per-scenario 30m → **45m**, `TimeoutStartSec` 5h → **7h**.

The ordering that actually matters is unchanged and still holds: the systemd backstop sits strictly above the harness's own bound, so the harness always ends cleanly and the dirty kill stays unreachable. Because the per-scenario cap is `min(cap, budget remaining)`, total runtime is bounded by the run budget itself and the backstop only needs teardown headroom.

These are deliberately sized to the runtime the harness **actually has**, not the one it ought to have. #2229 tracks the ~16x degradation itself — host CPU is 92–97% idle, disk is fine, and the host pulls 17 MB/s, so the time is going into in-container package installs and per-run image pulls. **Bring all three numbers back down once #2229 lands**; the code and the unit both say so at the point where the values are defined.

## Why a second PR

This session already shipped #2226 and normally stops there. Filing this as an issue instead would leave a known-wrong budget on main that turns the release gate red on the next deploy — worse than the sprawl. It's a three-value change plus comments.

## Verification

Root lint + `bun run test` green (9447 pass, 0 fail); `ui/ bun run check` 0 errors and `ui/ bun run test` 4977 pass; `fallow audit` clean on all changed files.

```shepherd:manual-steps
- [ ] POST-MERGE: install the updated unit on the Incus host — `cp ci/onboarding-harness/shepherd-onboarding.{service,timer} ~/.config/systemd/user/ && systemctl --user daemon-reload` (the running host currently has TimeoutStartSec=5h from #2226; this PR raises it to 7h, and the unit is not deployed by `bun run update`)
```

### DIFF
--- a/ci/onboarding-harness/README.md
+++ b/ci/onboarding-harness/README.md
@@ -86,11 +86,11 @@ The report lands at `onboarding-gap-report.md` in the working directory. Exit co
 
 ## Time budget
 
-The harness bounds itself so it always finishes on its own terms — **30 min per scenario** (`SHEPHERD_ONBOARDING_SCENARIO_TIMEOUT_MS`) and **3h per run** (`SHEPHERD_ONBOARDING_BUDGET_MS`). A scenario that blows its cap is abandoned; scenarios with no budget left are never started. Either way they are recorded **NOT VERIFIED** — not green, not a harness error, so a gate-eligible one still gates red and the report says plainly that no verdict was reached.
+The harness bounds itself so it always finishes on its own terms — **45 min per scenario** (`SHEPHERD_ONBOARDING_SCENARIO_TIMEOUT_MS`) and **6h per run** (`SHEPHERD_ONBOARDING_BUDGET_MS`). A scenario that blows its cap is abandoned; scenarios with no budget left are never started. Either way they are recorded **NOT VERIFIED** — not green, not a harness error, so a gate-eligible one still gates red and the report says plainly that no verdict was reached.
 
-Those caps are generous against observation: a healthy full run is ~14 min, a slow night ~61 min, and the worst single scenario ever seen took 26 min. Wall-clock is dominated by in-container package installs, so it tracks network weather rather than anything the harness controls.
+Those caps are sized to the runtime the harness **actually has**, not the one it ought to have: a full run was ~14 min until Aug 2026 and is now ~4h, with every scenario taking 20–28 min ([#2229](https://github.com/erwins-enkel/shepherd/issues/2229)). Wall-clock is dominated by in-container package installs and image pulls, which the harness does not control. Sizing the caps to the old runtime would cut off a legitimate, working run and report it NOT VERIFIED — a red release gate on a healthy harness. **Bring both numbers down, with `TimeoutStartSec`, once #2229 restores the runtime.**
 
-The service's `TimeoutStartSec` (5h) is a **backstop that must never be reached**, and must always stay above the harness's own budget. When it sat _below_ the worst case (the old 2h), systemd's dirty kill was the guaranteed outcome on a slow night rather than an edge case.
+The service's `TimeoutStartSec` (7h) is a **backstop that must never be reached**, and must always stay above the harness's own budget. When it sat _below_ the worst case (the old 2h), systemd's dirty kill was the guaranteed outcome on a slow night rather than an edge case.
 
 ## Run isolation
 
--- a/ci/onboarding-harness/run.ts
+++ b/ci/onboarding-harness/run.ts
@@ -407,20 +407,27 @@ function recordRunCompleted(only: string | null | undefined): void {
 
 /** Bound on ONE scenario, and on the whole run. Both exist so the harness always
  *  finishes on its OWN terms: the service's `TimeoutStartSec` is a backstop that
- *  must never be reached, because systemd's SIGTERM is a dirty kill — in Aug 2026
- *  it left a lock file and a running instance behind and disabled the nightly for
+ *  must never be reached, because systemd's SIGTERM is a dirty kill — while Bun
+ *  awaits a spawned child (which the harness does for nearly its whole runtime) a
+ *  SIGTERM never reaches the JS handler, so no cleanup runs at all. In Aug 2026
+ *  that left a lock file and a running instance behind and disabled the nightly for
  *  21 days. The harness's previous innermost bound was per `incus` CALL (20 min);
  *  with many calls per scenario and ten scenarios, its own worst case sat far above
  *  the 2h unit timeout, so systemd was GUARANTEED to win on a slow night.
  *
- *  The cap is generous against real observation: a healthy full run is ~14 min, a
- *  slow night ~61 min, and the worst single scenario ever seen took 26 min (the
- *  wall-clock is in-container package installs, so it tracks network weather, not
- *  anything the harness controls). 30 min per scenario / 3h per run therefore only
- *  trips when a night is genuinely pathological rather than merely slow. */
+ *  The caps are sized to the runtime the harness ACTUALLY has, not the one it ought
+ *  to have. A healthy full run used to be ~14 min; it is now ~4h (every scenario
+ *  20-28 min, see #2229) because the time is spent on in-container package installs
+ *  and image pulls, which the harness does not control. Sizing these to the old
+ *  runtime would cut off a legitimate, working run and report it NOT VERIFIED —
+ *  a red release gate on a healthy harness. **Bring both numbers back down (and
+ *  `TimeoutStartSec` with them) once #2229 restores the runtime.**
+ *
+ *  Because the per-scenario cap is `min(cap, budget remaining)`, total runtime is
+ *  bounded by the run budget itself; the backstop only needs headroom for teardown. */
 const SCENARIO_TIMEOUT_MS =
-  Number(process.env.SHEPHERD_ONBOARDING_SCENARIO_TIMEOUT_MS) || 30 * 60_000;
-const RUN_BUDGET_MS = Number(process.env.SHEPHERD_ONBOARDING_BUDGET_MS) || 3 * 60 * 60_000;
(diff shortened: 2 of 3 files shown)


scoop (commented):
# Review — `fix(onboarding): size the run budget to the harness's real runtime`

Three files, three numbers, and the prose around them. No logic changed. I checked every claim the
new prose makes against the code it describes, and every number against its siblings.

## What I verified

**The three numbers agree with each other and with the docs.**
- `ci/onboarding-harness/run.ts:428-430` — `SCENARIO_TIMEOUT_MS = 45 * 60_000` (45 min) and
  `RUN_BUDGET_MS = 6 * 60 * 60_000` (6 h). Arithmetic checked.
- `ci/onboarding-harness/README.md:89` states "**45 min per scenario**
  (`SHEPHERD_ONBOARDING_SCENARIO_TIMEOUT_MS`) and **6h per run** (`SHEPHERD_ONBOARDING_BUDGET_MS`)" —
  both env-var names match the reads at `run.ts:429` and `run.ts:430` exactly (spelling and casing).
- `ci/onboarding-harness/shepherd-onboarding.service:24` sets `TimeoutStartSec=7h`; its comment at
  `:18` says "caps itself at 6h per run + 45m per scenario, so 7h", and `README.md:93` says
  "`TimeoutStartSec` (7h)". All three restatements of the same numbers are consistent — no stale
  copy left behind.
- I grepped the whole tree for the old values and for the two env-var names: the only occurrences of
  either constant are `run.ts:429-430`, `README.md:89-93` and `shepherd-onboarding.service:18-24`.
  No test asserts the old `3h`/`30m`/`5h` (`test/` has no reference to either env var), so nothing
  out-of-diff goes stale on this change.

**The documented ordering invariant still holds.** `README.md:93` requires the backstop to "always
stay above the harness's own budget": 7 h > 6 h. ✔

**The new comment's load-bearing claim is true.** `run.ts:426-427` asserts "the per-scenario cap is
`min(cap, budget remaining)`". Verified at `run.ts:555`:
`runScenarioBounded(driver, s, tarball, Math.min(SCENARIO_TIMEOUT_MS, remaining))`, with `remaining`
computed from the deadline at `run.ts:545-547` and the exhausted branch at `run.ts:548-552`. So 10
scenarios × 45 min (7.5 h) cannot overrun the 6 h budget — the tail is recorded
`unverifiedResult(...)` instead. Scenario count confirmed as 10 (`ci/onboarding-harness/scenarios.ts`,
ten top-level `id:` entries).

**The `SIGTERM`-never-reaches-JS sentence added at `run.ts:410-412` is consistent with the rest of
the repo's account**, not a new or contradicting claim: same wording at
`shepherd-onboarding.service:13-15` and `README.md:101`, and the handler that the text calls
best-effort is at `run.ts:515-526` with a matching caveat comment at `run.ts:510-514`.

**Diagnosis fits the timeline.** `git log -S RUN_BUDGET_MS` shows the caps were introduced in
`cad5caf48` — the commit immediately preceding this one — so they had never been measured against
the harness's current runtime. That makes "the caps would have cut a legitimate run" a coherent
account rather than a retrofit, and widening them a proportionate mitigation while #2229 is open.

## Stated limitations

- Could not verify issue **#2229** (referenced at `run.ts:420`, `run.ts:424`, `README.md:91`,
  `shepherd-onboarding.service:22`) — it is on GitHub, not in this repo. Assumed as written; I make
  no claim either way about whether it exists or says what the comments attribute to it.
- Could not verify the linked Actions run `34392305279`, nor the observed timings ("~14 min until
  Aug 2026", "now ~4h", "every scenario 20–28 min"). These are field observations with no in-repo
  ground truth. I checked only that they are stated consistently in all three files.

## Notes on the numbers (all non-blocking, see nits)

- **Backstop headroom went from 2 h to 1 h.** Old: 5 h backstop − 3 h budget. New: 7 h − 6 h. And
  the 6 h budget does not cover the whole process: the deadline is taken at `run.ts:545`, *after*
  the reclaim sweep (`run.ts:535-538`) and `driver.ensureProfile()` / `buildTarball()`
  (`run.ts:543-544`), and the teardown sweep (`run.ts:559`), GitHub reporting (`run.ts:573`) and
  marker write (`run.ts:574`) all run after the loop. Each `incus` call inside those is itself
  capped at 20 min (`ci/onboarding-harness/incus.ts:30`), so the pre/post work is not bounded by
  anything close to 1 h on paper. In practice profile ops and deletes are seconds, so 1 h is very
  likely fine — but the margin shrank at exactly the moment the environment became slower and less
  predictable, which is the opposite direction from the reasoning the rest of the change applies.
- **The run budget, not the scenario cap, is now the binding constraint.** At the documented worst
  case (10 × 28 min = 4 h 40 min) the 6 h budget has ~28% slack while the 45 min scenario cap has
  ~60%. If #2229 worsens even modestly, the tail scenarios hit
  `"not run — the run budget was exhausted"` (`run.ts:550`) and gate red — precisely the outcome
  this PR exists to avoid.
- **Staleness-probe headroom.** `src/index.ts:2452` sets `ONBOARDING_STALENESS_MS = 30h` and its
  comment reads "One missed night trips it"; the probe fires on a 24 h interval
  (`src/index.ts:2596-2599`) and the marker is stamped at run *completion* (`run.ts:397-406`). With
  a 3 h budget the widest possible gap between two consecutive markers was ~27 h, comfortably under
  the threshold. With a 6 h budget it reaches ~30 h, i.e. the threshold itself — so a fast night
  followed by a budget-exhausting one can, in a narrow window, trip a false
  "the nightly may not be running" alert (`src/index.ts:2486`). Narrow enough that I am not blocking
  on it, but the 30 h number was chosen against a 3 h budget and no longer has slack.

- `ci/onboarding-harness/incus.ts:30` — `RUNNER_TIMEOUT_MS` is still 20 min per single `incus` call, while a whole scenario is now observed to take 20–28 min; if the bulk of that is one `incus exec`, the innermost cap will SIGKILL it (`spawn.ts:22-27`) long before the new 45 min scenario cap is relevant.
- `ci/onboarding-harness/run.ts:450-471` — `runScenarioBounded` races a timer against `runScenario` but cannot cancel it, so an abandoned scenario's `incus` work keeps running (and keeps holding host resources) while the next scenario starts; the comment acknowledges the instance leak is swept, not that the work stops.
- `src/index.ts:2452` — the 30 h staleness threshold is a fixed constant with no relation to the harness's configured run budget, so it needs a manual edit every time the budget moves.

- `ci/onboarding-harness/run.ts:410-412` — the task was to resize the budget; the diff also expands the SIGTERM/Bun-child explanation, which is now stated a third time (it already exists verbatim at `shepherd-onboarding.service:13-15` and `README.md:101`). Harmless, but it is duplicated prose the change did not need.

## Verdict

The change does what the task asks, the numbers are internally consistent across all three files,
and the one mechanical claim the new comment makes about the code is true. I found no correctness or
security defect. Nits only.


**Nits (non-blocking):**

- [bug] ci/onboarding-harness/shepherd-onboarding.service: the backstop margin over the harness's own budget shrank from 2h (5h-3h) to 1h (7h-6h), and the 6h budget does not cover the whole process — the deadline is taken at run.ts:545, after the reclaim sweep (run.ts:535-538), ensureProfile and buildTarball (run.ts:543-544), while the teardown sweep (run.ts:559), GitHub reporting (run.ts:573) and the marker write (run.ts:574) all run after the loop, each incus call inside them capped at 20 min (incus.ts:30). Consider keeping the old proportional headroom (e.g. 8h) now that the environment is slower and less predictable.
- [bug] ci/onboarding-harness/run.ts: the run budget, not the per-scenario cap, is now the binding constraint — at the documented worst case (10 x 28 min = 4h40m) the 6h budget has ~28% slack while the 45m scenario cap has ~60%. A modest further regression in #2229 makes the tail scenarios hit the budget-exhausted path (run.ts:550) and gate red, which is the exact outcome this change exists to prevent. Sizing the two caps to comparable headroom would avoid that.
- [bug] ci/onboarding-harness/run.ts: raising RUN_BUDGET_MS to 6h eats the headroom the 30h onboarding staleness threshold was chosen against. The marker is stamped at run completion (run.ts:397-406) and the probe fires on a 24h interval (src/index.ts:2596-2599), so the widest gap between consecutive markers grows from ~27h to ~30h — at or just over ONBOARDING_STALENESS_MS (src/index.ts:2452), whose comment still reads one missed night trips it. A fast night followed by a budget-exhausting one can raise a false the-nightly-may-not-be-running alert; worth revisiting that constant alongside the budget.
- [bug] ci/onboarding-harness/run.ts: the new sentence at run.ts:426-427 says total runtime is bounded by the run budget itself. The min(cap, remaining) part is true (verified at run.ts:555), but total process runtime is preamble + 6h + teardown + reporting, since the deadline is only taken at run.ts:545. Phrasing it as the scenario loop being bounded by the budget would be accurate.

PR #2232, opened 2026-09-10
fix(forge): resolve the viewer over both gh transports, keep a failure retryable (#2140)

### LINKED ISSUE #2140: currentUser() is REST-only and caches a failure for the process lifetime, silencing the 'mine & unassigned' filter
Split out of #2139, which fixed the same class of problem for `listIssues()` but
deliberately left this one alone to keep that diff tight.

## What happens

`GithubForge.currentUser()` resolves the operator's login with a single REST call
(`gh api user`). On any failure it caches `null`:

- `cachedUser` is set to `null` inside the `catch`, and `currentUser()` returns early on
  `cachedUser !== undefined` — so the failure is remembered, not just the success.
- Forge instances are memoised for the **process lifetime** (`makeForgeMemo()` in
  `src/forge/resolve.ts` caches positives permanently, by design).

So one REST failure at the wrong moment pins `viewer` to `null` until Shepherd restarts.
`/api/issues` then reports `viewer: null`, the UI fails open, and the "mine & unassigned"
filter (#824) silently shows everything with no chip and no explanation.

## Why it is reachable

REST and GraphQL sit on independent GitHub budgets, and `gh api rate_limit` is itself
limit-exempt — so it can report a full REST budget while every real REST call 403s. That was
observed live while working on #2139: `gh api repos/<slug>/issues` returned
`HTTP 403 "API rate limit exceeded for user ID"` while `gh api graphql -f query='{viewer{login}}'`
answered normally.

Both halves of the bug matter: the call had a healthy alternative transport it never tried,
and it made the resulting failure permanent.

## Suggested fix

1. Fall back to `gh api graphql -f query='{viewer{login}}'` when the REST call fails —
   mirroring the bidirectional pattern `listIssues()` now uses.
2. Cache only a **successful** resolution. A failure should stay retryable, or at most be
   negatively cached with a short TTL (compare the negative-TTL treatment in
   `makeForgeMemo()`).

## Not in scope here

The same one-way-fallback shape exists in `getIssue`, `listPullRequests`,
`listBacklogCounts`, `prStatus` and `prReviewMeta`. Worth a separate look — this issue is
only about `currentUser()`, whose extra sting is the process-lifetime cache.

scoop:
Closes #2140.

`GithubForge.currentUser()` resolved the operator's login with a single REST call
(`gh api user`) and, on **any** failure, cached `null`. Forge instances are memoised for
the process lifetime, so one ill-timed 403 pinned `viewer` to `null` until a restart:
`/api/issues` reported `viewer: null`, the UI failed open, and the "mine & unassigned"
filter (#824) silently showed everything with no chip and no explanation.

Both halves of the bug are fixed.

### Two transports

REST and GraphQL sit on **independent** GitHub budgets — observed live on #2139, where
every REST call 403'd while `gh api graphql` answered normally. `currentUser()` now tries
`gh api user`, then falls back to `gh api graphql '{viewer{login}}'` on any failure, or on
an answer that carries no login.

The order is fixed REST → GraphQL rather than flipped on `graphRateLimit.blocked()` the way
`listIssues()` does it: that signal describes the GraphQL bucket only, and there is no
REST-side tracker that would argue for GraphQL-first. The fallback runs even inside an
active GraphQL backoff — REST has just proved unusable and the backoff is only a heuristic
(the same trade-off `listIssues()`' REST→CLI direction already makes).

### A failure that stays retryable

New `src/forge/user-cache.ts` → `makeUserCache()`, with the asymmetric contract
`makeForgeMemo()` established in `src/forge/resolve.ts`: a resolved login is cached for the
process lifetime (the identity never changes mid-session), a failure only for a 30s
negative TTL and is then re-probed.

The TTL — rather than "always retryable" — is what makes the retry affordable:
`pr-poller`'s `refresh()` resolves the viewer for every session on every sweep, so an
unbounded retry would mean a fresh probe per session per tick while the forge is
unreachable.

`GiteaForge.currentUser()` had the identical permanent-null cache and moves onto the same
helper. It has a single transport, so it gets the retryable-failure half only.

### Scope

No UI or API-shape change: `/api/issues` still answers `viewer: string | null` and
`IssuesPanel` keeps failing open while the viewer is unknown. The same one-way-fallback
shape in `getIssue`, `listPullRequests`, `listBacklogCounts`, `prStatus` and `prReviewMeta`
stays parked, as the issue asks.

### Verification

`bun run lint`, `bun run test` (9454 pass / 0 fail), `bunx tsc --noEmit` and
`bunx fallow audit --base origin/main --fail-on-issues` all clean. TTL expiry is covered in
`test/forge/user-cache.test.ts` with an injected clock; the forge tests cover transport
order, GraphQL parsing, success caching and that a failure is not re-probed inside the
window.

### DIFF
--- a/src/forge/gitea.ts
+++ b/src/forge/gitea.ts
@@ -2,6 +2,7 @@ import { mapGiteaActionStatus, mapStatusState } from "./checks";
 import { classifyPr } from "./pr-kind";
 import { labelColorsFrom } from "./labels";
 import { mapBounded } from "../map-bounded";
+import { makeUserCache } from "./user-cache";
 import type {
   ChecksState,
   ForgeConfig,
@@ -183,20 +184,19 @@ export class GiteaForge implements GitForge {
     }
   }
 
-  private cachedUser: string | null | undefined;
-  /** The authenticated Gitea login (`GET /api/v1/user`), cached for the forge's
-   *  lifetime — it never changes mid-session. Drives the "mine & unassigned" issue
-   *  filter (#824) so the chip is live on Gitea too, not just GitHub. Null when it
-   *  can't be resolved (no token / network error) → fail open (show all). */
+  /** The authenticated Gitea login (`GET /api/v1/user`). Drives the "mine & unassigned"
+   *  issue filter (#824) so the chip is live on Gitea too, not just GitHub. A resolved
+   *  login is cached for the forge's lifetime (it never changes mid-session); a failure
+   *  (no token / network error) reads as "unknown me" → fail open (show all), and stays
+   *  retryable rather than pinning the viewer to null until a restart (#2140). Gitea has
+   *  a single transport, so there is no second one to fall back to. */
+  private readonly resolveUser = makeUserCache(async () => {
+    const u = (await this.req("GET", "/api/v1/user")) as { login?: string } | null;
+    return u?.login || null;
+  });
+
   async currentUser(): Promise<string | null> {
-    if (this.cachedUser !== undefined) return this.cachedUser;
-    try {
-      const u = (await this.req("GET", "/api/v1/user")) as { login?: string } | null;
-      this.cachedUser = u?.login || null;
-    } catch {
-      this.cachedUser = null; // unauth / offline → treat as "unknown me"
-    }
-    return this.cachedUser;
+    return this.resolveUser();
   }
 
   /** One combined-status call yields both the worst-of rollup (top-level `state`)
--- a/src/forge/github.ts
+++ b/src/forge/github.ts
@@ -10,6 +10,7 @@ import {
   runningCheckNames,
 } from "./checks";
 import { classifyPr } from "./pr-kind";
+import { makeUserCache } from "./user-cache";
 import { labelColorsFrom } from "./labels";
 import {
   attachAttempts,
@@ -1678,17 +1679,42 @@ export class GithubForge implements GitForge {
       .map((r) => r.slice("refs/heads/".length));
   }
 
-  private cachedUser: string | null | undefined;
-  /** The authenticated gh login (`gh api user`), cached for the forge's lifetime —
-   *  it never changes mid-session, so one call serves every handoff computation. */
+  /** `gh api user` — the REST-bucket transport for {@link currentUser}. */
+  private async currentUserRest(): Promise<string | null> {
+    return (await this.run(["api", "user", "--jq", ".login"])).trim() || null;
+  }
+
+  /** `gh api graphql {viewer{login}}` — the GraphQL-bucket transport for
+   *  {@link currentUser}. */
+  private async currentUserGraphql(): Promise<string | null> {
+    const out = await this.run(["api", "graphql", "-f", "query=query{viewer{login}}"]);
+    const json = JSON.parse(out || "null") as { data?: { viewer?: { login?: string } } } | null;
+    return json?.data?.viewer?.login?.trim() || null;
+  }
+
+  /**
+   * The authenticated gh login, over whichever transport answers.
+   *
+   * REST first (`gh api user`), then GraphQL (`{viewer{login}}`) on ANY failure or on
+   * an answer that carries no login. The two draw on INDEPENDENT GitHub budgets, so
+   * either can be exhausted while the other is healthy — observed live on #2139, where
+   * every REST call 403'd while `gh api graphql` answered normally. The order is fixed
+   * rather than flipped on `graphRateLimit.blocked()` the way `listIssues` does it:
+   * that signal describes the GraphQL bucket only, and there is no REST-side tracker to
+   * argue for GraphQL-first. The fallback runs even inside an active GraphQL backoff —
+   * REST has just proved unusable and the backoff is only a heuristic (the same
+   * trade-off `listIssues`' REST→CLI direction makes).
+   *
+   * A resolved login is cached for the forge's lifetime; a failure only until a short
+   * TTL elapses. See {@link makeUserCache} for why the failure must stay retryable.
+   */
+  private readonly resolveUser = makeUserCache(async () => {
+    const rest = await this.currentUserRest().catch(() => null);
+    return rest ?? (await this.currentUserGraphql());
+  });
+
   async currentUser(): Promise<string | null> {
-    if (this.cachedUser !== undefined) return this.cachedUser;
-    try {
-      this.cachedUser = (await this.run(["api", "user", "--jq", ".login"])).trim() || null;
-    } catch {
-      this.cachedUser = null; // unauth / offline → treat as "unknown me"
-    }
-    return this.cachedUser;
+    return this.resolveUser();
   }
 
   /** Whether the authenticated user can push. Returns a DEFINITIVE boolean only;
--- a/src/forge/user-cache.ts
+++ b/src/forge/user-cache.ts
@@ -1,0 +1,48 @@
+/**
+ * Cache for a forge's "who am I" lookup — the operator's own login, which drives the
+ * "mine & unassigned" issue filter (#824).
+ *
+ * Caching is **asymmetric**, mirroring `makeForgeMemo` in ./resolve.ts:
+ *  - **Positive** (a login) is cached for the process lifetime. The authenticated
+ *    identity never changes mid-session, so one lookup serves every caller.
+ *  - **Negative** (the probe threw, or answered without a login) is cached only for
+ *    `negativeTtlMs`, then re-probed. This is the #2140 fix: remembering a FAILURE for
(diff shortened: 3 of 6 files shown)


scoop (commented):
# Review — #2140: resolve the viewer over both gh transports, keep a failure retryable

**Decision: comment** — no blocking defects found. The diff does what the issue asked, and I could
not find a correctness or security problem in it. Four nits below.

## What the change does

`src/forge/user-cache.ts` (new) is a per-instance closure cache with asymmetric semantics:
positive → permanent, negative (throw *or* falsy resolution) → `negativeTtlMs`, then re-probe.
`GithubForge.currentUser()` (`src/forge/github.ts:1682-1718`) now resolves REST-first
(`gh api user --jq .login`) and falls back to `gh api graphql -f query=query{viewer{login}}` on any
throw or on an answer carrying no login. `GiteaForge.currentUser()` (`src/forge/gitea.ts:187-200`)
keeps its single transport and picks up only the retryable-failure half.

Both halves the issue named are addressed: the alternative transport is now tried, and a failure is
no longer remembered for the process lifetime.

## Verification performed

**GraphQL transport is real and correctly shaped.**
- `gh api graphql -f query=…` is the exact call shape already used in this file — compared against
  `src/forge/github.ts:698-704` (`listBacklogCounts`) and `src/forge/github.ts:877-884`
  (`getIssue`). Same `["api", "graphql", "-f", "query=…"]` argv, no hostname/header flags anywhere
  in `src/forge/` (grepped `hostname|GH_HOST|cfg.host` across `src/forge` — no matches), so nothing
  is missing relative to its siblings.
- `query{viewer{login}}` is a valid anonymous GraphQL operation (OperationType + SelectionSet, no
  name), and the response shape `data.viewer.login` matches what the parser at
  `src/forge/github.ts:1691-1692` destructures.
- Rate-limit bookkeeping still works for the new call: `isGraphqlBucketCall` returns true for
  `args[0] === "api" && args[1] === "graphql"` (`src/forge/rate-limit.ts:186-190`), and the default
  runner notes limit errors on exactly that predicate (`src/forge/github.ts:317-320`). So a
  rate-limited fallback still extends the shared backoff. The new query does not select
  `rateLimit{remaining resetAt}` the way `listBacklogCounts` does
  (`src/forge/github.ts:740`) — that is telemetry only, not a correctness gap, and a viewer lookup
  is a 0/1-point query.

**Cache semantics.**
- `DEFAULT_NEGATIVE_TTL_MS = 30_000` in `src/forge/user-cache.ts:22` genuinely matches
  `makeForgeMemo`'s, verified at `src/forge/resolve.ts:20`; the closure shape (positive map / TTL'd
  negative / injectable `now`) mirrors `src/forge/resolve.ts:42-67`.
- The `failedAt: number | null` sentinel is correct where a `0` timestamp is possible — the guard is
  `failedAt !== null` (`src/forge/user-cache.ts:35`), not a truthiness test, so a clock at 0 (as in
  `test/forge/user-cache.test.ts:45`) still caches the negative. `login !== null`
  (`src/forge/user-cache.ts:34`) is safe because `login` is only ever assigned from a truthy
  `resolved` (`src/forge/user-cache.ts:40-41`).
- A late success clears `failedAt` (`src/forge/user-cache.ts:42`), so the cache cannot ping-pong
  once a login resolves. That also means the poller's `me` cannot flap between a login and `null`
  after first success — a real improvement over the old field.

**Class-field initialisation order.** Both forges moved the cache into a `private readonly` field
initialiser that closes over `this` (`src/forge/github.ts:1711`, `src/forge/gitea.ts:193`). With
`"target": "ESNext"` (`tsconfig.json:3`) `useDefineForClassFields` defaults on, so field
initialisers run *before* the constructor body assigns the parameter properties `run`/`cfg`/
`fetchFn` (`src/forge/github.ts:551-564`, `src/forge/gitea.ts:72-80`) and before `this.base` is set
(`src/forge/gitea.ts:70,77`). That is harmless here: the arrow bodies only dereference `this.run` /
`this.req` / `this.base` when invoked, which cannot happen before construction returns. Checked that
nothing calls `currentUser()` during construction (all call sites are `src/server.ts:1784,1858,4044,
4123,4340,6141,7386`, `src/index.ts:2034`, `src/pr-poller.ts:413`, `src/up-next.ts:222`).

**No dangling references.** `cachedUser` has zero occurrences repo-wide after the change (grepped);
`makeUserCache` is imported in both consumers (`src/forge/github.ts:13`, `src/forge/gitea.ts:5`) and
exported once (`src/forge/user-cache.ts:24`). The `{@link makeUserCache}` in the new github.ts
doc-block resolves to that import. No signature change: `currentUser(): Promise<string | null>` still
matches the optional member at `src/forge/types.ts:505`, so every `forge.currentUser?.()` caller is
unaffected.

**The fix actually reaches the UI.** `/api/issues` re-reads the viewer per request
(`src/server.ts:6141`), and the UI-side `viewerCache.set()` overwrites the per-repo entry on every
`listIssues` resolve (`ui/src/lib/viewer-cache.svelte.ts:20-22`) rather than latching the first
value — so once the forge re-probes successfully, the chip comes back without a restart. There is no
second permanent-null layer defeating the fix.

**Retry cost is bounded on the paths that matter.** The full sweep awaits `refresh` per session
sequentially (`src/pr-poller.ts:265-267`), so within one sweep only the first session pays the
probe and the rest hit the negative cache — not N probes per tick. `src/forge/gitea.ts:95-98`
confirms `req` throws on any non-ok response, so the Gitea 403/401 fixtures exercise the throw path
the helper catches.

**Tests.** The new fixtures line up with the helpers they use: `fakeFetch` accepts `{status}` and
records `{method,url,…}` (`test/forge/gitea.test.ts:15-33`), and the pre-existing success-caching
test (`test/forge/gitea.test.ts:385-393`) still asserts exactly one `/api/v1/user` call, which the
positive-permanent branch preserves. `userRunner`/`VIEWER_JSON` are new names with no collision in
`test/forge/github.test.ts`. `new GithubForge("o/r", {}, run)` matches the established construction
pattern (`test/forge/github.test.ts:137`).

**Lint gates.** `@typescript-eslint/no-floating-promises` / `no-misused-promises` apply to
`src/**/*.ts` (`eslint.config.js:25-34`); every promise in the new code is awaited
(`src/forge/user-cache.ts:39`, `src/forge/github.ts:1712-1713`, `src/forge/gitea.ts:194`). No
import-order rule exists in `eslint.config.js`, so the import placement is free.

## Stated limitations

- I did not verify GitHub's live behaviour for `gh api user --jq .login` on a 200 that omits
  `login`, nor the exact GraphQL error envelope `gh` prints for a rate-limited `{viewer{login}}` —
  neither ground truth is in this repo. I reasoned from the in-repo parsers only and assumed the
  documented shapes.
- Per instructions I did not run lint or tests; correctness claims above come from reading the code
  and its fixtures.

## Observation (not a finding)

`src/server.ts:6139-6140` still says "Cached in the forge, so no per-request gh cost after the first
call." That is now only true on the success path — a failing resolution costs up to two `gh` calls
once per 30s window. `src/server.ts` is outside this diff, so it is not a finding; flagging it only
so the next docs sweep can catch it.

- `src/forge/user-cache.ts`: extracting a shared helper for two call sites (plus its own test file) is more machinery than the issue's "cache only a successful resolution" strictly needed — a judgement call the plan made deliberately, and it does keep the two forges honest.
- `src/forge/gitea.ts`: the Gitea half is beyond the issue's literal title (`currentUser()` on GitHub), though it fixes the same defect class and the plan named it up front.


**Nits (non-blocking):**

- [bug] src/forge/github.ts:1684: currentUserRest() maps a jq output of `null` to the literal string "null" (`.trim() || null` keeps it), while the new GraphQL sibling at :1692 correctly yields null for a missing viewer. The two transports now disagree on the same edge case. The REST line is unchanged from before the PR and needs a 200 from /user with no `login` field to trigger, so it is negligible today — but a `.replace(/^null$/, "")`-style guard or a `--jq '.login // empty'` would make the siblings symmetric.
- [bug] src/forge/user-cache.ts: no in-flight de-duplication, so concurrent callers each run a full REST-then-GraphQL pair while a failure is uncached (failedAt is written only after the await). The negative TTL bounds how often the probe repeats, not how many run at once; the sweep in pr-poller.ts is sequential so the poller is safe, but overlapping /api/issues requests are not. The plan declined this deliberately — worth a line in the doc comment saying so, since the comment currently reasons only about per-tick cost.
- [compliance] src/forge/gitea.ts:198 and src/forge/github.ts:1716: the public currentUser() methods are now undocumented — the doc comments moved onto the private resolveUser fields. Hovering the GitForge implementation no longer surfaces the caching contract. A one-line `/** @see resolveUser */` on each method would restore it.

Issue #809, opened 2026-06-19
Visual plan unreachable after sign-off (no way back to plan during execution; esp. mobile)


scoop:
## Problem

Once a session's plan is signed off and work starts (`planPhase` flips `planning` → `executing`), **there is no way to re-open the visual plan**. The plan-gate badge — the *only* entry point to `PlanPanel` → `VisualReview` — disappears, and there is no tab, link, or row action to bring it back. This is most painful on **mobile** (less room for the terminal/diff, so glancing back at the approved plan matters more), but it's not mobile-specific: the plan is gone on desktop too.

The visual blocks only resurface **read-only after completion**, in the Done recap (`DoneRecapPanel`). Nothing surfaces them *during* execution.

Reported by an operator running a build-queue session on mobile: "I don't find any buttons or actions to get to the plan after it was signed off."

## Why it happens

The badge is hidden by design as soon as the phase flips:

`ui/src/lib/components/plan-gate-badge.ts` — `planGateChip()`:
```ts
// Gate only lives during the plan phase; executing means it already passed.
if (session.planPhase == null || session.planPhase === "executing") return { kind: "none" };
```

`PlanPanel` is likewise hard-locked to `planPhase === "planning"` (`canReviewNow`, comment: "Manual re-review only makes sense while still planning"). The plan blocks themselves come from `gate?.blocks` (`PlanPanel.svelte` — `const planBlocks = $derived(gate?.blocks ?? [])`).

## Data durability (the load-bearing constraint)

Whether the plan data still *exists* during execution depends on how the session left planning — this matters for any fix:

- **"Go" release path** (`SessionService.releasePlanGate`, `src/service.ts`): flips to `executing` but does **not** drop the gate. The `PlanGate` row (incl. `blocks`) persists in `plan_gates` until archive. → data is available throughout execution; surfacing it is **UI-only**.
- **Manual-steer path** (no Go — operator reviews then steers the agent, which opens a PR while still `planning`): the `session:git` subscriber calls `advanceToExecutionOnPr()` then `planGate.forget(id)` (`src/index.ts` ~L735), which runs `dropPlanGate()` → `DELETE FROM plan_gates` (`src/store.ts`). → the plan blocks are **purged the moment a PR appears**, so even a UI change can't recover them.
- **Both paths**: at `session:archived`, `planGate.forget(id)` drops the gate; a separate recap is generated for the Done view.

[...]

PR #2234, opened 2026-09-10
feat(plan-gate): re-review a plan edited after approval (#2224)

### LINKED ISSUE #2224: Plan gate cannot re-review a plan edited after approval
## Problem

`PlanGateService.consider()` returns `"skipped"` on `prior?.approved` — and the comment is explicit that `force` does **not** bypass it. So once a gate is approved, the plan file is never reviewed again, no matter how far it is subsequently rewritten.

That was invisible until #2223: the PR critic used to be handed the live file under an "approved" label, so a post-approval rewrite silently became the reviewed plan. #2223 closes the laundering channel (the critic now sees the approved snapshot, and any later edit labelled as never reviewed), but the underlying gap remains — a plan edited after approval has no path back through the gate at all.

## Why it matters

Legitimate mid-execution replanning is real: an agent discovers the approach doesn't work and revises. Today that revision gets no adversarial review, and the operator sees nothing that says the plan moved.

## Sketch

`plan_gates.planHash` already fingerprints the reviewed text, so divergence is cheap to detect. Options, roughly in cost order:

- surface it: an attention signal / badge when the live plan's hash no longer matches the approved one;
- allow an explicit operator-triggered re-review of an edited plan (a narrow `force` that *does* bypass the `approved` short-circuit, resetting the round);
- auto-re-gate on divergence — most expensive, and needs care not to re-open a gate mid-execution on a trivial typo fix.

Not urgent: with #2223 merged, an edited plan can no longer misrepresent itself to the critic.

Found while investigating the scope dispute on #2222.

scoop:
Closes #2224.

`PlanGateService.consider()` returned `"skipped"` on `prior?.approved`, and `force` deliberately did not bypass it — so once a gate was approved the plan file was never reviewed again, however far it was subsequently rewritten. #2223 closed the laundering channel (the critic now sees the approved snapshot); this restores a path back through the gate.

## What it does

Two persisted columns on `plan_gates` carry the feature:

- **`livePlanHash`** — sha256 of the live `.shepherd-plan.md`, re-hashed on the **settle edge** (idle/done) for any approved gate, churn-guarded, and carrying `updatedAt` forward rather than restamping it. Divergence rides the existing `PlanGate` payload to the UI: no new WS event, no per-tick file reads on the single loop.
- **`approvedAt`** — when the plan was last approved. It marks a session *deliberately re-gated* (`approvedAt != null && !approved`) and is the baseline for the reviewer's round-lateness count, so a re-review is briefed as round 1 instead of inheriting `roundBlock`'s late-round posture against a plan no reviewer has seen.

`consider()` gets exactly one bypass: **force + approved + live text diverged**. It is the only thing that runs past an approved gate or off the planning phase; the auto path, the unchanged-plan dedupe, the #1944 refusal suppression and every in-flight guard are untouched.

Verdict handling:

| Verdict | `planning` | `executing` |
| --- | --- | --- |
| approve | as today | snapshot re-points at the edited plan; session keeps executing |
| request-changes | as today | **re-gate**: approval cleared, phase → `planning`, agent steered to stop and revise |
| error / refused spawn | as today | gate row untouched — approval and snapshot stand |

## Three things worth reviewing

1. **Ordering is load-bearing.** `applyChangesRequested()` persists at the tail, and `regatePlanGate`'s guard reads the *store* — so the executing branch persists the revoking verdict **first**, then re-gates, then falls through to the existing steer/round tail. Reverse it and the re-gate silently no-ops: approval revoked, session left executing, no stop steer. The test wires the real `SessionService.regatePlanGate` (a mocked dep would pass any ordering) and fails when the persist is moved after the call.
2. **`advanceToExecutionOnPr` runs on every git poll** and would flip a re-gated PR-bearing session straight back to `executing`. Suppressed by the derived `approvedAt != null && !approved`; re-approval lifts it, so nothing can latch a session in planning.
3. **`approved ⇒ no reviewer in flight` is deliberately broken** by an execution-phase re-review (that is what makes an `approve` zero-disturbance). Both maintainers are handled: `consider()`'s short-circuit, and `adoptOrphans()` — where **two** guards stood in the way. Its `planPhase !== "planning"` check fired before the gate was read, so a restart mid-re-review left the reviewer neither adopted nor reaped: verdict unread, terminal alive, `completedAt` NULL, worktree then deleted by `gcStaleReviewWorktrees()`.

Self-review during implementation caught a fourth: `publishSpawnRefusal` wrote `approved: false` **plus the edited plan text**, so a membrane-refused re-review would have revoked a live approval *and* handed the PR critic unreviewed text as "the plan that was approved". Guarded, with tests both ways.

## Not in scope

Auto re-gating on divergence, a new attention signal or Herd-row CTA, an approved-vs-edited diff view, per-tick polling.

## Verification

`bun run lint` · `bun run test` (9487) · `cd ui && bun run check && bun run test` (4994) `&& bun run check:i18n` · `bunx fallow audit --base origin/main --fail-on-issues` → pass (0 introduced findings; the complexity the change added was extracted into `skipForPlanHash`, `orphanDisposition`, `orNull` and two `{#snippet}`s rather than absorbed by raising a threshold).


scoop:
# Review — re-review a plan edited after approval (#2224)

Scope reviewed: `git diff 3ce88cf1379e5ac29e0d505cfbef6d71d5561c00...HEAD` (21 files: server, UI, docs,
tests).

## What I verified

**Store / type plumbing.** Both new columns are in the fresh-install DDL (`src/store.ts:1376-1377`)
**and** in `migratePlanGateColumns()` (`src/store.ts:4853-4854`) — the fresh-DDL-vs-migration parity
trap the plan flagged is avoided. `getPlanGate`'s SELECT (`src/store.ts:3447`), `putPlanGate`'s INSERT
column list plus its placeholder row (`src/store.ts:3456-3457` — 21 columns, 21 `?`), the upsert's
`DO UPDATE SET` (`src/store.ts:3466`), the binds (`src/store.ts:3487-3488`) and `snapshotPlanGates`
(`src/store.ts:3671`) all name `livePlanHash, approvedAt`, and `hydratePlanGate` maps them
(`src/store.ts:3439-3440`). Server type at `src/types.ts:707,713`; client mirror at
`ui/src/lib/types.ts:555`. `approvedAt` is deliberately server-only — I grepped `approvedAt` across
`ui/src` and there are no consumers, so the omission is consistent, not a drift.

**Round arithmetic.** The new comment at `src/plan-gate.ts:1013-1017` claims a re-review is briefed as
round 1. That holds: `buildGate` writes `round: 0` on an approve (`src/plan-gate.ts:1741`) and
`effectiveRound(0, 0, cap)` = `max(min(0+1, cap), 0)` = 1 per `src/critic-core.ts:738-741`. The spawn
count also cannot include the current run — `composeReviewerPrompt` is called at
`src/plan-gate.ts:1052`, long before the reviewer-spawn row is recorded. Legacy rows carry
`approvedAt == null` → `?? 0` → `since <= 0` → every row counts exactly as before
(`src/plan-gate.ts:833-841`).

**adoptOrphans.** The gate read now precedes the phase guard (`src/plan-gate.ts:1266-1271`), and
`orphanDisposition` (`src/plan-gate.ts:104-113`) keeps the crash-window reap (`spawnedAt <= updatedAt`
on an approved gate) while adopting a genuine re-review. The `updatedAt` discriminator is sound: I
checked every `putPlanGate` caller. `noteLivePlan` spreads `...prior` without restamping
(`src/plan-gate.ts:815`); `resume`/`dismiss` spread the gate and are reachable only for
`decision === "changes_requested"` (`src/plan-gate.ts:1805,1823`); the `/plan/answers` route merges
without touching `updatedAt` (`src/server.ts:2865-2868`). Nothing bumps `updatedAt` between an approval
and a later re-review spawn, so a re-review orphan is never mistaken for a husk.

**Re-gate ordering.** `applyChangesRequested` reads `prior` (and therefore `priorRound`) *before* the
provisional write (`src/plan-gate.ts:1543-1547`), then persists → `regate` → falls through to the
existing steer/round/`putPlanGate` tail (`src/plan-gate.ts:1557-1589`). `regatePlanGate`'s guard reads
the store (`src/service.ts:5482-5485`), so that ordering is the only thing that makes it satisfiable,
which matches the documented caller contract. `advanceToExecutionOnPr`'s suppression
(`src/service.ts:5519-5520`) is derived, not a stored flag. `releasePlanGate`'s
`planPhase === "planning"` guard (`src/service.ts:5453`) confirms the plan's claim that an `approve`
verdict on an executing session is a genuine no-op even for an `auto`/autopilot session
(`src/plan-gate.ts:1512-1514`).

**Chip / control fan-out.** The new `"edited"` chip kind is handled in every exhaustive consumer I
found: the tooltip switch (`ui/src/lib/components/plan-gate-badge.ts:188-189`), the badge label chain
and the `pg-{chip.kind}` class binding (`ui/src/lib/components/PlanGateBadge.svelte:206,234`, with
`.pg-edited` defined at `:369`), and `PlanPanel`'s `planStatusNote` switch
(`ui/src/lib/components/PlanPanel.svelte:345-346`, tone mapped at `:360` to the existing
`.status-note.changes` rule at `:919`). `ui/src/lib/hold-row.ts:62,181` always pass `allowView: false`,
so an executing session still yields `{kind:"none"}` there and its `rowState` ladder is untouched. No
other `planGateChip` consumer exists.

**Read-only chrome during execution** (the standing plan-phase house rule): `canRelease` stays
planning-only (`ui/src/lib/components/plan-gate-badge.ts:77`), `canShowPlanStallActions` likewise
(`:86`), `planAnswerCtx` was correctly re-keyed onto the new `planning` derived rather than the widened
`canReviewNow` (`ui/src/lib/components/PlanPanel.svelte:58-60`), and the Go button stays inside
`{#if !readonly}` (`ui/src/lib/components/PlanPanel.svelte:571`). Server-side, the answer round-trip
still 409s off the planning phase (`src/server.ts:2848`). So the widened surface really is Review-only.

**Route.** `handleSessionReviewPlan` (`src/server.ts:3120-3126`) has no phase guard of its own and
already passes `force: true` — the plan's "the route needs no change" is accurate.

**i18n parity.** Six new keys, present in both catalogs: `plangate_tip_edited` (`ui/messages/en.json:668`
/ `ui/messages/de.json:668`), `planpanel_status_edited` + `planpanel_edited_note`
(`ui/messages/en.json:692-693` / `ui/messages/de.json:692-693`), `plangate_edited`
(`ui/messages/en.json:1819` / `ui/messages/de.json:1819`), `feat_plan_regate_title` +
`feat_plan_regate_body` (`ui/messages/en.json:3846-3847` / `ui/messages/de.json:3846-3847`). None take
placeholders, so there is no interpolation drift to check.

**Feature catalog.** `ui/src/lib/feature-announcements/entries/v1.48.0-plan-regate.ts` matches the
`v<sinceVersion>-<id>.ts` contract documented at `ui/src/lib/feature-announcements.ts:11-24`; 1.48.0 is
the next unreleased version (`package.json:3` is at 1.47.0) and is the version every other pending entry
in that directory already uses. Registration is by `import.meta.glob`
(`ui/src/lib/feature-announcements.ts:47-50`), so no index edit is needed.

**Docs.** The glossary change is body prose under an existing heading
(`docs-site/src/content/docs/reference/glossary.md:35-38`); `ui/scripts/gen-docs-manifest.ts:57-60`
extracts only H2/H3 headings (plus frontmatter), so the manifest stays fresh.

**Svelte snippets.** `{@render actionsFooter()}` (`ui/src/lib/components/PlanPanel.svelte:558`) resolves
against the top-level `{#snippet}` declarations at `:565` and `:585` — component-top-level snippets are
in scope for the whole template — and the moved `.review` / `.go` / `.actions` markup stays in the same
component's `<style>`, so the "extracting a Svelte child, copy the moved classes" house rule does not
apply.

## Findings

### 1. `src/plan-gate.ts`: a refused spawn wipes the deliberate-re-gate marker (important)

`publishSpawnRefusal` builds a **fresh** `PlanGate` literal (`src/plan-gate.ts:1655-1669`) that omits
`approvedAt` (and `livePlanHash`). `putPlanGate` binds the missing field through `orNull`
(`src/store.ts:735,3488`) → `NULL`. The new early-return added by this PR only covers an *approved*
prior (`src/plan-gate.ts:1653`), so a **re-gated** session — `approved: false`, `approvedAt` stamped,
back in `planning` — walks straight into the literal and loses the stamp.

Concrete failure path, all of it already-executing code:

1. Session approved at T (`approvedAt = T`, `src/plan-gate.ts:1767`) → plan edited → operator re-review
   → `request-changes` → `regatePlanGate` flips it to `planning` (`src/service.ts:5486`) and
   `advanceToExecutionOnPr` starts returning false (`src/service.ts:5520`).
2. The agent revises the plan; the settle edge fires
   `shouldConsiderOnSettle("idle", "planning", "changes_requested")` → true (`src/plan-gate.ts:96`) →
   unforced `consider()` clears every short-circuit in `skipForPlanHash` (prior not approved, phase is
   planning, hash changed, no matching spawn notice — `src/plan-gate.ts:789-793`) → `begin()`.
3. The membrane refuses the spawn → `publishSpawnRefusal` (`src/plan-gate.ts:1117`) writes a row with
   `approvedAt: NULL`.
4. On the next git poll (`src/index.ts:1816`) `advanceToExecutionOnPr` sees `approvedAt == null`,
   returns true, and flips the re-gated session back to `"executing"` — carrying an `error`,
   un-approved gate.

That directly breaks the PR's own success criterion ("**stays** re-gated across PR-poller ticks for a
PR-bearing session"), and `#2111`'s doc comment on this method says the refusal re-fires on every sweep
while the host stays broken, so it is not an exotic window.

Fix: carry the stamp into the refusal row — `approvedAt: prior?.approvedAt ?? null` (and, for symmetry,
`livePlanHash: prior?.livePlanHash ?? null`), the way `buildGate` does at `src/plan-gate.ts:1767`.

## Evidence behind the recorded non-blocking points

- **The provisional re-gate write emits nothing.** `applyChangesRequested` persists the revoking gate at
  `src/plan-gate.ts:1560` without calling `this.deps.onChange`, while `regatePlanGate` does emit
  `session:plangate {planPhase:"planning"}` (`src/service.ts:5487`). The client applies the phase patch
  immediately (`ui/src/lib/store.svelte.ts:679`) but keeps its cached `approved: true` gate until the
  tail `onChange` at `src/plan-gate.ts:1589`. In between — across the awaited stop steer
  (`src/service.ts:5491`) and the awaited findings steer, which for Codex includes a pane resume —
  `planGateChip` yields `{kind:"ready"}` (`ui/src/lib/components/plan-gate-badge.ts:59`) and
  `canRelease` is true (`:77`), so the panel shows a READY chip with an enabled Go. Clicking it is
  harmless (`releasePlanGate` re-reads the store, already `approved: false` — `src/service.ts:5454`),
  but the approved plan explicitly specified emitting `onChange` on this write and it is missing.
- **No REVIEWING chip during an execution-phase re-review.** `planGateChip` returns from the
  `planPhase === "executing"` branch (`ui/src/lib/components/plan-gate-badge.ts:51-54`) before the
  `if (reviewing)` check at `:55`, so the top-bar badge stays "PLAN EDITED" for the whole run.
  `PlanPanel`'s button still spins (`inFlight`, `ui/src/lib/components/PlanPanel.svelte:195`) and
  GitRail still blocks the control (`canTriggerPlanReview` → `"reviewing"`,
  `ui/src/lib/components/plan-gate-badge.ts:125`), so this is badge-only.
- **No operator override for the suppression.** `advanceToExecutionOnPr` returns false for as long as
  `approvedAt != null && !approved` (`src/service.ts:5520`). `dismiss()` spreads the gate and changes
  only `round` / `finalRoundPending` / `dismissed` (`src/plan-gate.ts:1827`), so the "operator took
  over" escape no longer un-latches the phase; the only exits are Resume-then-approve, or a fresh
  approval. Defensible as the feature's intent — but it is a state that could not exist before this PR,
  and a re-gated session at the rework cap whose operator dismisses has no path back to `"executing"`.
- **`plan-unavailable` where `skipped` used to be.** `consider()` now falls through the phase guard when
  `force` is set (`src/plan-gate.ts:746-747`) and reads + hashes the plan before `skipForPlanHash`
  re-applies it (`src/plan-gate.ts:790`). An executing session with no readable `.shepherd-plan.md`
  therefore returns `"plan-unavailable"` (`src/plan-gate.ts:753`) where it previously returned
  `"skipped"` without touching the disk. Not reachable from the UI (the control is gated on
  `planEdited`, which needs a gate), but the route at `src/server.ts:3120` is callable directly.

- `publishSpawnRefusal` also drops `finalRoundPending`, `dismissed`, `blocks` and the reviewer-env fields when it replaces a row; that predates this PR and only the `approvedAt` loss is newly harmful.

- `src/store.ts`: the new `orNull<T>()` helper (`src/store.ts:735-737`) is used for exactly the two new binds while every neighbouring optional field still inlines `?? null` (`src/store.ts:3481-3486`) — an unrequested helper that buys little at two call sites.
- `ui/src/lib/components/PlanPanel.svelte`: the `reviewButton` / `actionsFooter` snippet extraction plus six new `$derived` attribute holders (`describedBy`, `reviewAriaDisabled`, `reviewDisabled`, `goDisabled`, `reviewInert`, `reviewAriaLabel`) is a larger restructure than the footer change strictly needed; the complexity-bar justification is stated in the comments and is a legitimate author call.
- `src/plan-gate.ts`: the `publishSpawnRefusal` approved-gate guard (`:1653`) is defensive work the plan never listed — correct as far as it goes, but see finding 1 for the half of it that is missing.

- `src/plan-gate.ts`: a crash between the provisional `putPlanGate` (`:1560`) and `regate` (`:1562`) leaves a NEW state — `planPhase: "executing"` with a non-approved gate — that `orphanDisposition` classifies as `"skip"` (`:110`), so the already-finalizing reviewer's spawn row keeps a NULL `completedAt` and the session sits executing with its approval revoked and no stop steer. The window is two statements wide and the verdict has already been consumed, so only bookkeeping is lost.

- "Don't reply/steer a done or running session" — `regatePlanGate` steers (`src/service.ts:5491`) a session that is by construction mid-execution and possibly running. That is the feature's whole point, and it mirrors `releasePlanGate` at `:5460`, so the rule's intent is met by other means.
- For a Codex session whose pane has exited, `regatePlanGate`'s plain `this.reply` returns false (contract at `src/service.ts:4895-4903`) and the STOP steer is silently dropped, while the findings steer that follows goes through `resumeThenSteer` and does land. `planSteerText` ends with "Don't start implementing yet — wait for the plan to be approved" (`src/plan-gate.ts:1929`), so the instruction is not actually lost — worth knowing, not worth changing.

## Stated limitations

- I did not run `bun run lint`, `bun run test`, `cd ui && bun run check / test / check:i18n`, or
  `fallow audit` — per the review instructions those are CI's job. Every claim above comes from reading
  the code, not from a green run.
- The `.fallowrc.jsonc` "Tier-1 Svelte bar" the PlanPanel comments cite is a threshold I did not
  evaluate; I take the author's word that the snippet split was needed to stay under it.


**Nits (non-blocking):**

- [bug] src/plan-gate.ts: the provisional re-gate write at :1560 does not emit onChange, although regatePlanGate immediately broadcasts planPhase: planning (src/service.ts:5487) and the client applies it (ui/src/lib/store.svelte.ts:679). Until the tail onChange at :1589 the client holds a stale approved gate, so the panel renders a READY chip with an enabled Go on an already-re-gated session for the duration of both awaited steers. The approved plan specified emitting onChange on this write.
- [bug] ui/src/lib/components/plan-gate-badge.ts: planGateChip returns from the executing branch (:51-54) before the reviewing check at :55, so during an execution-phase re-review the top-bar badge stays PLAN EDITED and never shows REVIEWING. PlanPanel and GitRail still reflect the in-flight run, so this is badge-only.
- [bug] src/service.ts: the advanceToExecutionOnPr suppression (:5520) has no operator override. dismiss() only resets round/finalRoundPending/dismissed (src/plan-gate.ts:1827), so a session re-gated out of execution and then dismissed at the rework cap has no path back to executing except a fresh approval — a state that could not exist before this PR. Consider treating dismissed as lifting the suppression, or say so in the doc comment.
- [bug] src/plan-gate.ts: with force set, consider() now falls past the phase guard (:746-747) and reads plus hashes the plan file before skipForPlanHash re-applies it (:790), so an executing session with no readable plan returns plan-unavailable (:753) where it previously returned skipped without touching the disk. Not reachable from the UI, but the /review-plan route (src/server.ts:3120) is callable directly.

scoop:
# Review — re-review a plan edited after approval (#2224)

Rework round 2. Read `git diff 3ce88cf1379e5ac29e0d505cfbef6d71d5561c00...HEAD` in full (21 files),
then resolved every identifier and contract the diff touches against the tree.

## Prior finding — status

**1. `publishSpawnRefusal` drops `approvedAt` (and `livePlanHash`) from its fresh `PlanGate` literal:
ADDRESSED.** The literal at `src/plan-gate.ts:1655-1678` now carries both fields explicitly
(`approvedAt: prior?.approvedAt ?? null` at :1675, `livePlanHash: prior?.livePlanHash ?? null` at
:1676), and the method additionally early-returns on an approved prior at :1653. I re-walked the
exact path the finding described: settle edge re-considers a `changes_requested` gate
(`shouldConsiderOnSettle` returns true for `idle` + `changes_requested`, `src/plan-gate.ts:97`), the
revised plan clears every `skipForPlanHash` short-circuit (:189-193), the membrane refuses
(:1117 → `publishSpawnRefusal`), and the next git poll (`src/index.ts:1816`) reads the row. With the
carry in place `advanceToExecutionOnPr`'s suppression (`src/service.ts:5520` — `gate && !gate.approved
&& gate.approvedAt != null`) still holds. There is a regression test for exactly this
(`test/plan-gate-service.test.ts`, "a refused spawn on a RE-GATED session keeps the re-gate marker").
Not re-raised.

## What I verified

**Store round-trip.** The two columns exist in the fresh-install DDL (`src/store.ts:1374-1375`:
`livePlanHash TEXT, approvedAt INTEGER`) *and* in `migratePlanGateColumns` (`src/store.ts:4850-4852`
`add("livePlanHash", …)` / `add("approvedAt", …)`), so an existing DB and a new one agree — the
`REVIEWERS_SPAWNS_DDL`-style trap the plan flagged is avoided. They are selected in `getPlanGate`
(:3459), in `snapshotPlanGates` (:3671), hydrated at :3439-3440, and bound in `putPlanGate`
(:3487-3488 with the matching `excluded.*` upsert clause at :3465). `putPlanGate` is the only writer
of `plan_gates` outside `src/store.ts` (grepped: `src/plan-gate.ts:816,1505,1560,1588,1679,1696,1820,1837`
and `src/server.ts:2869`), so there is no generic-`update()` whitelist hazard here. Every one of those
other writers builds its row by spread (`{...gate, …}` / `{...prior, livePlanHash}`), so neither new
field is silently NULLed — checked `resume()` at :1820, `dismiss()` at :1837 and the answer route at
`src/server.ts:2864-2869`.

**`updatedAt` carry.** `noteLivePlan` writes `{ ...prior, livePlanHash }` (`src/plan-gate.ts:816`) and
therefore does not restamp `updatedAt`, which is load-bearing: `isReReviewSpawn` (:120) discriminates a
re-review orphan from a crash-window husk on `spawnedAt > gate.updatedAt`. `resume`/`dismiss` also
preserve `updatedAt`, so that discriminator cannot be perturbed by an operator action.

**`adoptOrphans` reordering.** The `getPlanGate` read moved above the phase guard
(`src/plan-gate.ts:1259-1263`), and the `worktreeExists` check still gates both `reap` and `adopt`
(:1263 before :1272/:1276) — so the pre-existing "reaped worktree ⇒ finalize already ran" skip is
preserved. `orphanDisposition` (:104-113) reproduces the old behaviour exactly for every prior case
(`executing` → skip, non-`planning` → skip, `planning` + approved → reap, otherwise adopt) and only
adds the two re-review arms.

**Route.** `POST /api/sessions/:id/review-plan` (`src/server.ts:3120-3126`) has no phase guard and
passes `force: true`, so the executing-phase re-review really reaches `consider()`; the plan's "the
route needs no change" holds.

**`release`/`regate` symmetry.** `applyApproved` → `deps.release` → `releasePlanGate`
(`src/service.ts:5451-5453`) returns false off the planning phase, so an `approve` verdict on an
executing re-review is genuinely zero-disturbance. `regatePlanGate` (:5481-5493) is the mirror and its
guard (`planPhase === "executing"` + a persisted non-approved gate) is satisfiable only after the
provisional write at `src/plan-gate.ts:1560` — the ordering the plan describes is what the code does.

**UI contracts.** `PlanGateChip` gained `{ kind: "edited" }` (`ui/src/lib/components/plan-gate-badge.ts:35`);
I grepped every consumer. `ui/src/lib/hold-row.ts:62,181` calls `planGateChip(…, { allowView: false })`,
which returns `{kind:"none"}` for an executing session (`plan-gate-badge.ts:52`), so the new kind can
never reach `hold-row`'s `LINE`/`ACTION` maps — no missing `Record<RowState, …>` entry. Both exhaustive
switches on `chip.kind` were extended: `planGateTooltipHint` (`plan-gate-badge.ts:188-189`) and
`PlanPanel.planStatusNote` (`PlanPanel.svelte:345-346`). `PlanGateTooltipCopy` gained `edited:
string` (:151) and both call sites supply it (`PlanGateBadge.svelte:100`, `PlanGateBadge.test.ts:41`).
The chip class is templated (`class="pg-badge pg-{chip.kind}"`, `PlanGateBadge.svelte:206`), so the new
`.pg-edited` rule (:369) is actually applied, and `toggle()` falls through to `open = true`
(:131) so the edited chip opens the panel like `view` does.

**GitRail leak check.** `canReviewPlan` is consumed only by the single Re-review button
(`ui/src/lib/components/git-rail/RailStatusActions.svelte:342-359`) and by the `hasActions` term
(:188). No Go, rework counter or stall menu is reachable through it — the house rule about
plan-phase-only chrome leaking into execution holds for the widened control.

**i18n parity — verified key by key**, en and de, same line numbers in both files:
`plangate_tip_edited` (`ui/messages/en.json:668` / `ui/messages/de.json:668`),
`planpanel_status_edited` (:692 / :692), `planpanel_edited_note` (:693 / :693),
`plangate_edited` (:1819 / :1819), `feat_plan_regate_title` (:3846 / :3846),
`feat_plan_regate_body` (:3847 / :3847). The pre-existing keys the diff newly relies on also exist in
both: `gitrail_rereview_plan` (:1695), `gitrail_reviewing_plan` (:1696),
`gitrail_review_plan_approved` (:2961), `planpanel_review_now` (:676),
`planpanel_review_already_approved` (:678).

**Feature announcement.** `ui/src/lib/feature-announcements/entries/v1.48.0-plan-regate.ts` — id
`plan-regate` (unique among the eight `v1.48.0-*` siblings), `sinceVersion: "1.48.0"` which is greater
than the released `1.47.0` in `package.json`, filename version agrees with `sinceVersion`, and the
entry is auto-registered by the eager glob at `ui/src/lib/feature-announcements.ts:49-52`. Its
`titleKey`/`bodyKey` resolve in both catalogs (cited above).

**Docs.** `docs-site/src/content/docs/reference/glossary.md:35-38` adds body prose under the existing
**Plan gate** heading; no new heading, so no docs-manifest churn.

---

## Findings

### 1. `src/plan-gate.ts:1559` — the re-gate branch is not guarded on this being a re-review

`applyChangesRequested` re-gates on `store.get(f.sessionId)?.planPhase === "executing"` alone. Every
other site this PR touches carries an "is this a re-review of an *approved* gate" guard —
`skipForPlanHash`'s `reReview` (`src/plan-gate.ts:189`), `isReReviewSpawn` (:120),
`publishSpawnRefusal`'s `prior?.approved` (:1653), `applyError`'s `live?.approved` (:1695). This one
does not, even though `prior` is already in hand two lines above (:1545). Its own comment (:1550-1551)
says "a re-review of an edited plan can land on a session that is already EXECUTING", so the intent
is narrower than the condition.

Failure scenario (never-approved gate, executing session): a session steered manually rather than
released with Go writes code and opens a PR while `planPhase` is still `"planning"`; the git event
handler calls `advanceToExecutionOnPr` on every poll (`src/index.ts:1814-1816`) and flips it to
`"executing"`. If a plan reviewer's verdict finalizes for that session after the flip, this branch
fires with `prior.approved === false` and `prior.approvedAt === null`, and two things go wrong:

- The agent receives `PLAN_REGATE_STEER` (`src/service.ts:1569-1574`), which states as fact
  "Your plan was edited after it was approved, and the re-review of the edited plan requested changes
  — this session is back in the planning phase and its approval is withdrawn." None of that is true;
  the plan was never approved. The steer also orders it to stop writing product code.
- Because `approvedAt` is null, `advanceToExecutionOnPr`'s new suppression (`src/service.ts:5520`)
  does **not** apply, so the very next git poll for that PR-bearing session flips it straight back to
  `"executing"` with no counter-steer. The re-gate is pure disruption: a false stop instruction plus a
  phase flap the operator sees as a flicker.

Reachability today: `onSessionGit` calls `planGate.reapReviewer(id)` immediately after a successful
advance (`src/index.ts:1823`), which drops the entry from `inflight` (`src/plan-gate.ts:1851-1856`) —
so the common single-review case is covered. It is not covered when `tick()` is finalizing a *different*
session: `tick()` iterates a snapshot (`[...this.inflight.values()]`, :1413) and awaits
`finalize` per entry (:1434), and the only per-entry re-check is `f.finalizing` (:1414). A git event
arriving during session A's `await finalize` reaps session B from the map, but B is still in the
snapshot array and is finalized on the same pass — with `planPhase` now `"executing"`. That is an
already-executing path, not a foreshadowed one, and Shepherd runs many sessions concurrently.

Fix: add the re-review guard, e.g. `if (prior?.approved && this.deps.store.get(f.sessionId)?.planPhase
=== "executing")`. That also makes `PLAN_REGATE_STEER`'s wording unconditionally true, which is the
property its doc comment claims ("this message stays true whether or not that steer lands",
`src/service.ts:1565-1566`).

### 2. `ui/src/lib/components/plan-gate-badge.ts:51-61` — the edited marker never reaches the badge (or the status note) in the planning phase

`planGateChip` only yields `{ kind: "edited" }` inside the `planPhase === "executing"` arm (:51-53).
For a **planning** session with an approved-but-edited gate it falls through to `if (gate?.approved)
return { kind: "ready" }` (:59). The consequences, all on the same screen at the same time:

- The badge renders `plangate_ready` = "READY ✓" (`ui/messages/en.json:639`,
  `PlanGateBadge.svelte:228-229`) with the ready pulse where `pulseReady` is passed
  (`PlanGateBadge.svelte:48`), and its tooltip reads `plangate_tip_ready` = "Plan approved; execution
  can start." (`ui/messages/en.json:665`).
- `PlanPanel`'s status note resolves through `planStatusNote`'s `case "ready"`
  (`PlanPanel.svelte:331-332`) to `planpanel_status_ready` = "Plan approved. Start execution when
  ready." (`ui/messages/en.json:685`) — printed in the same panel that renders, immediately above the
  plan text, `planpanel_edited_note` = "Shown below is the plan that was APPROVED. The live
  .shepherd-plan.md has been edited since and has not been reviewed."
  (`PlanPanel.svelte:454-456`, `ui/messages/en.json:693`). The two statements contradict each other.
- `canRelease` is unchanged (`plan-gate-badge.ts:77`: `gate?.approved && planPhase === "planning"`), so
  Go stays enabled and `releasePlanGate` steers "implement the approved plan"
  (`src/service.ts:5460`) for a plan file that no reviewer has seen.

This is the state the issue's first bullet asks to surface ("an attention signal / badge when the live
plan's hash no longer matches the approved one") and it is the more dangerous of the two phases,
because Go is live there. The new test `plan-gate-badge.test.ts:130-132` pins the current behaviour
("while planning an approved gate still reads as ready, edited or not"), so this reads as an oversight
rather than a decision. The server side already supports it — `shouldCheckPlanDrift`
(`src/plan-gate.ts:109-112`) is deliberately phase-agnostic and `noteLivePlan` writes `livePlanHash`
for a planning session too, and `planEdited` (`plan-gate-badge.ts:68-70`) is phase-free — so only the
chip and the status-note mapping need to honour it.

---

- `src/plan-gate.ts:1148-1149` — the widened `consider()` makes `refuseOversizedSpawn` reachable during execution, where it fire-and-forget steers the *implementing* agent to shorten `.shepherd-plan.md` (`src/plan-gate.ts:981-988`); bounded to `MAX_REFUSAL_STEERS = 2` (:1897) and only for an over-argv-budget plan, so harm is limited to waking the pane with a plan-phase instruction.
- `src/service.ts:5491` — `PLAN_REGATE_STEER` goes out through the raw `reply`, not `resumeThenSteer`, so on an exited Codex pane the stop instruction is dropped while the findings that follow are still delivered (they resume the pane, `src/plan-gate.ts:1522-1527`). `releasePlanGate` has the same shape, and the doc comment acknowledges the dead-pane case, so this is consistent rather than new — but the two steers now have different delivery guarantees.

- `src/store.ts:480-482` — the `orNull<T>()` helper is introduced for exactly two bind sites (:3487-3488) where the surrounding 19 binds all inline `?? null`; it reads as a one-use abstraction that breaks local symmetry.
- `ui/src/lib/components/PlanPanel.svelte:395-439` — the footer was refactored into two top-level `{#snippet}`s plus five new `$derived` attribute constants (`describedBy`, `reviewAriaDisabled`, `reviewDisabled`, `goDisabled`, `reviewAriaLabel`); justified in-comment as staying under the fallow Tier-1 Svelte bar, but it is a larger restructuring of untouched markup than the feature needed.

- "Mirror every type/contract change across `src/types.ts` AND `ui/src/lib/types.ts`" — `approvedAt` is added to the server `PlanGate` (`src/types.ts:706-711`) but deliberately omitted from the client one (`ui/src/lib/types.ts:552-555`), while `onChange` emits the whole row over `session:plangate` (`src/index.ts:1603`). The client type therefore under-describes the wire shape. A deliberate call per the plan, and there is no `exactKeys()` drift sentinel for `PlanGate` (grepped: only `test/server-usage-breakdown.test.ts` / `test/server-usage-timeline.test.ts` have them), so nothing breaks — noting it only because the rule names the file pair explicitly.
- "Don't reply/steer a done or running session" — `regatePlanGate` steers a possibly-running executing agent (`src/service.ts:5491`). This is the feature's central, operator-triggered decision (documented in the plan's Decisions table), not an incidental nudge, so the rule's intent is met by other means.

- `PlanPanel`'s Go button is enabled while a plan review is in flight (`goDisabled` reads only `busy || quotaBusy || !releasable`), so an operator can release an approved gate mid-re-review; the diff neither introduces nor worsens this.
- `src/plan-gate.ts` `readPlan(session.worktreePath)` resolves to the shared repo directory for a non-isolated session, so a sibling session writing `.shepherd-plan.md` is indistinguishable from the owner editing it — pre-existing in `consider()`, now also read by the drift check.

## Stated limitations

- I did not run lint, the test suites, `check:i18n`, or `fallow audit` — per instructions CI owns those. The i18n parity claim above is from reading both catalogs directly, not from the checker.
- `.pg-edited`'s visual result (amber on the badge's base rule) was read from CSS, not rendered; I confirmed only that the class is emitted and the rule exists.

## Plan drift

Minor. The implementation follows the plan's approach and file list closely; the one substantive
departure is finding 2 (the badge marker landing only in the executing phase, where the plan's success
criteria said both phases), plus the incidental `PlanPanel` snippet refactor the plan did not describe.


**Nits (non-blocking):**

- [bug] src/plan-gate.ts:1653 and :1695: a re-review of an approved gate that is REFUSED by the membrane or resolves to `error` now leaves the row completely untouched and the route returns "skipped" (src/server.ts:3124), so the operator's click produces no gate change, no spawn notice on the refusal path, and no explanation — only the reviewing indicator clearing. The comments accept this tradeoff; a coded, non-revoking surface for it would close the gap.

scoop (commented):
# Review — re-review a plan edited after approval (#2224), round 3

## Prior finding

**1. `applyChangesRequested`'s re-gate branch guarded only on `planPhase === "executing"` — ADDRESSED.**
`git show 600cd9dfd` narrows it to `if (prior?.approved && this.deps.store.get(f.sessionId)?.planPhase === "executing")` at `src/plan-gate.ts:1570`, with the reasoning recorded in the comment block at `src/plan-gate.ts:1554-1563` and a regression test (`test/plan-gate-service.test.ts:3410-3433`, asserting `regated` stays `[]` for an executing session whose gate was never approved). A first-review `changes_requested` verdict on an `advanceToExecutionOnPr`-latched session (`src/index.ts:1816`) now falls through to the plain steer path with `approvedAt === null` untouched, so neither the false `PLAN_REGATE_STEER` (`src/service.ts:5418-5423`) nor the phase flap can occur. Not re-raised.

## What I verified

**Store round-trip.** `plan_gates` gains `livePlanHash TEXT` / `approvedAt INTEGER` in BOTH the fresh-install DDL (`src/store.ts:1375-1376`) and `migratePlanGateColumns` (`src/store.ts:4850-4854`, whose `add()` at `src/store.ts:4833-4836` is idempotent against `PRAGMA table_info`), satisfying the house rule about CREATE-time columns skipping existing DBs. The dedicated setter `putPlanGate` (`src/store.ts:3454-3491`) was updated on all three surfaces — INSERT column list, `ON CONFLICT … DO UPDATE SET` (`src/store.ts:3466`), and the bind array (`src/store.ts:3487-3488`). Placeholder arity checked mechanically: the regex `^\s+VALUES \(\?(,\?){20}\)$` matches `src/store.ts:3457`, i.e. exactly 21 placeholders against 21 columns and 21 bind values. Both read paths — `getPlanGate` (`src/store.ts:3447`) and `snapshotPlanGates` (`src/store.ts:3671`) — carry the two new names, and `hydratePlanGate` projects them (`src/store.ts:3439-3440`). I grepped every `plan_gates` reference in `src/store.ts`; the only other statements are DELETEs and the legacy `summaryCode` UPDATE, which need no change.

**No fresh-literal drops the new fields.** I grepped every `PlanGate = {` / `putPlanGate(` site in `src/`. The three spread-based writers preserve them (`src/plan-gate.ts:815` `noteLivePlan`, `src/plan-gate.ts:1830`/`1847` `resume`/`dismiss`, `src/server.ts:2865-2866` the answer round-trip). The two fresh literals both carry explicitly: `buildGate` (`src/plan-gate.ts:1783,1787`) and `publishSpawnRefusal` (`src/plan-gate.ts:1686-1687`). That last one matters twice over — `src/server.ts:2865` also proves the answer route does **not** restamp `updatedAt`, which is what keeps `isReReviewSpawn`'s `spawnedAt > gate.updatedAt` discriminator (`src/plan-gate.ts:120`) honest for an approved gate.

**`consider()`'s two bypasses.** Traced all four reachable combinations against `src/plan-gate.ts:743-793`:
- auto-path, approved gate → `src/plan-gate.ts:751` short-circuits before any hash, unchanged;
- force + planning + approved + unedited → `reReview` false at `src/plan-gate.ts:788`, `src/plan-gate.ts:789` returns `true` → `"skipped"`, matching the UI's `"approved"` block at `ui/src/lib/components/plan-gate-badge.ts:127`;
- force + executing + no gate → `prior` null so `src/plan-gate.ts:789` is false but `src/plan-gate.ts:790` (`planPhase !== "planning" && !reReview`) returns `true` → `"skipped"`;
- force + edited approved gate, either phase → both guards fall through to `begin()`.
`session.planPhase == null` is rejected up front (`src/plan-gate.ts:745`), and `planPhase` is only ever `"planning" | "executing" | null` (`src/service.ts:3668` is the only null-writing create path, for terminal/research/epic sessions that never get a gate row), so no third phase reaches the bypass.

**`shouldCheckPlanDrift` cannot collide with `shouldConsiderOnSettle`.** Both fire from the same hook (`src/index.ts:2186-2197`). The overlap is `status === "done"` + planning + approved gate; there `consider()` returns at the synchronous `src/plan-gate.ts:751` before touching `starting`, so `noteLivePlan` (`src/plan-gate.ts:807-818`) cannot race a spawn. `noteLivePlan` itself re-guards on `inflight`/`starting` (`src/plan-gate.ts:810`) and preserves `updatedAt` via the spread at `src/plan-gate.ts:815` — which `src/plan-status.ts`'s `planStallStatus` and `isReReviewSpawn` both depend on. The hook fires on a status *edge* (`src/session-router.ts:35-46`), not per tick.

**`adoptOrphans` disposition table.** `orphanDisposition` (`src/plan-gate.ts:136-145`) is exhaustive over the reachable states: executing + re-review → adopt; executing otherwise → skip (pre-#2224 behaviour); planning + approved + not-re-review → reap (the crash-window husk `reapOrphanSpawn` at `src/plan-gate.ts:1326` exists for); planning otherwise → adopt. The gate read moved above the guard (`src/plan-gate.ts:1259`) and the `worktreeExists` check still precedes both adopt and reap (`src/plan-gate.ts:1263`), so no husk is reaped twice.

**`advanceToExecutionOnPr` suppression is derived, not latched by a flag.** `src/service.ts:5519-5520` reads the store each call; `buildGate` restamps `approvedAt` only on approval (`src/plan-gate.ts:1787`), so re-approval lifts it. `applyApproved`'s `release()` on an executing session is a real no-op — `releasePlanGate` guards `planPhase !== "planning"` at `src/service.ts:5452`, so an `approve` verdict on an execution-phase re-review disturbs nothing, as the plan's success criteria require.

**The regate ordering is actually load-bearing and actually tested.** `regatePlanGate`'s guard reads the store (`src/service.ts:5482-5484`), and `applyChangesRequested` persists before calling it (`src/plan-gate.ts:1571-1573`). The test does not mock the dep — `test/plan-gate-service.test.ts:3296-3297` wires the real `service.regatePlanGate`, with the rationale at `test/plan-gate-service.test.ts:3214-3216`.

**`error` verdict cannot revoke an approval.** `applyError` (`src/plan-gate.ts:1704-1709`) re-reads the live row and writes only when `!live?.approved`; the stall signal still fires. Cross-checked against `buildGate`'s `approved: resolved === "approved"` (`src/plan-gate.ts:1766`), which is what would otherwise un-approve.

**Locale parity.** All six new keys exist in both catalogs — `plangate_tip_edited` (`ui/messages/en.json:668`, `ui/messages/de.json:668`), `planpanel_status_edited` (`:692`/`:692`), `planpanel_edited_note` (`:693`/`:693`), `plangate_edited` (`:1819`/`:1819`), `feat_plan_regate_title` (`:3846`/`:3846`), `feat_plan_regate_body` (`:3847`/`:3847`). `ui/messages/` contains only `en.json` and `de.json`, so that is full coverage.

**Feature-announcement entry.** `ui/src/lib/feature-announcements/entries/v1.48.0-plan-regate.ts` matches the `v<sinceVersion>-<id>.ts` contract documented at `ui/src/lib/feature-announcements.ts:16-23`, and the catalog is `import.meta.glob`-driven (`ui/src/lib/feature-announcements.ts:47-50`) so no index needs updating. `sinceVersion: "1.48.0"` agrees with its 19 siblings in that directory.

**Chip plumbing.** `pg-edited` is produced by the existing `class="pg-badge pg-{chip.kind}"` interpolation (`ui/src/lib/components/PlanGateBadge.svelte:206`) and the new rule at `ui/src/lib/components/PlanGateBadge.svelte:141-145` uses `--color-amber`, defined at `ui/src/app.css:21`. Both exhaustive switches over `PlanGateChip` got the new arm (`ui/src/lib/components/plan-gate-badge.ts:188-189`, `ui/src/lib/components/PlanPanel.svelte:345-346`). I grepped every `chip.kind` consumer: the only other one that could see an executing session is `ui/src/lib/hold-row.ts:62`, which passes `allowView: false` and therefore receives `{kind:"none"}` (`ui/src/lib/components/plan-gate-badge.ts:52`).

**No plan-phase chrome leaks into execution.** Against the house rule on plan-phase-only branches: `canRelease` (`ui/src/lib/components/plan-gate-badge.ts:77`) and `canShowPlanStallActions` (`ui/src/lib/components/plan-gate-badge.ts:86`) both still require `planPhase === "planning"`; the answer round-trip still 409s off planning (`src/server.ts:2848`) and `planAnswerCtx` is now keyed on the new `planning` derived rather than the widened `canReviewNow` (`ui/src/lib/components/PlanPanel.svelte:56-58`); the Go button stays inside `{#if !readonly}` (`ui/src/lib/components/PlanPanel.svelte:423`); and `rework`/`atCap` are false for an edited gate because it is `approved` (`ui/src/lib/components/PlanPanel.svelte:78,81`). The widened `canReviewPlan` in GitRail feeds exactly one button in `RailStatusActions.svelte:342-360` (an untouched file) plus its `hasActions` term at `:188` — no other control.

**Snippet refactor is behaviour-preserving.** The inline `planReviewBlock === "approved"` click guard moved into the handler (`ui/src/lib/components/PlanPanel.svelte:248`), keeping `aria-disabled` + `aria-label` focusability; `showActions` (`ui/src/lib/components/PlanPanel.svelte:87`) reproduces the old `{#if !readonly}` for planning and adds the executing+edited case; the status note is rendered outside the footer (`ui/src/lib/components/PlanPanel.svelte:516-520`) so `aria-describedby={describedBy}` still resolves.

## Stated limitations

- I did not run lint, typecheck, or tests — CI owns those, per the brief.
- Svelte 5 top-level `{#snippet}` hoisting (the two new snippets at `ui/src/lib/components/PlanPanel.svelte:556` and `:576` are declared after the `{@render actionsFooter()}` call at `:551`) is compiler behaviour I reasoned about rather than executed; `ui/src/lib/components/PlanPanel.browser.test.ts:139-218` exercises exactly those render paths, so a hoisting problem would fail CI rather than ship.

- `ui/src/lib/components/PlanPanel.svelte` extracts the action footer and the Review button into two top-level snippets and hoists four attribute derivations into the script — a readability/complexity-budget refactor the plan did not describe, though it is behaviour-preserving.
- `src/store.ts:491-493` adds a general-purpose `orNull<T>` helper for two call sites where `?? null` inline would have done.
- `src/plan-gate.ts:136-145` `orphanDisposition` is a three-way pure classifier where the plan described relaxing two inline guards; justified by the health bar, but it is more structure than the task needed.

- `src/plan-gate.ts:1570-1577` — if `regate` throws or no-ops (a dead store read, a session archived mid-finalize), the approval is already revoked by the provisional write at `:1571` while `planPhase` stays `"executing"`. The failure is logged and the findings steer still lands, but the session then renders the read-only `"view"` chip against a `changes_requested` gate. Not reachable through any path in this diff today, since `regatePlanGate`'s only two no-op conditions are both false at that call site.

- `ui/src/lib/components/PlanPanel.svelte`'s `rework`/`atCap` hints already render for an executing session holding a retained `changes_requested` gate (the #809 `advanceToExecutionOnPr` latch); this diff neither creates nor widens that path.


**Nits (non-blocking):**

- [bug] src/plan-gate.ts:1571: the provisional putPlanGate is not paired with an onChange, unlike every other write in this file (compare :816, :1506, :1600, :1691). The plan's step 3(a) called for emitting it. Between that write and the tail emit at :1600, clients have seen regatePlanGate's planPhase:planning event but still hold the stale approved gate, so PlanPanel briefly offers a Go button that the server would reject.

Issue #1823, opened 2026-07-17
Persist effective spawn model/effort so the status bar can show runtime substitutions


kai-osthoff:
The session status bar (PR #1822) displays the session row's configured model/effort, explicitly labeled as configuration via its hover title — because the stored fields are intent, not what actually spawned:

- `pushModelFlag` (src/service.ts) applies usage-downgrade and availability fallbacks **argv-only** ("never rewrites the stored session model"), so after a downgrade the agent can run sonnet while `session.model` says opus (or fable → opus[1m] when fable is unavailable).
- Codex clamps unsupported effort tiers at spawn (`max → high`, asserted in test/service.test.ts "clamping max → high (#1417)") while the stored intent keeps the un-clamped tier.

To show the **effective** environment, the substituted values need to be persisted/exposed at every spawn site — create, replace-agent, relaunch, resume (where a usage downgrade can differ per spawn, so a single write-once column would itself go stale), and the drain/epic spawn paths — e.g. transient `effectiveModel`/`effectiveEffort` fields written whenever argv is assembled, cleared/refreshed per spawn. Once exposed, the status bar (and metaPop) can render "configured X → running Y" and drop the intent-labeling title; add status-bar coverage for a substituted model and a clamped effort tier against the real fields.

Raised by the PR critic on #1822; the interim state ships the configured-intent labeling (the critic's alternative option).

PR #2236, opened 2026-09-10
feat(ui): name the coding CLI on task cards and show what a run really used

kai-osthoff:
A task card never said which coding CLI it ran on, and its model and effort both
fell back to the localized default. A Codex task that actually ran **`gpt-6-astra`
at `high` effort** read as:

```
TASK-1090 · Standard · Standard
```

— naming neither the CLI nor the truth.

## Two causes

**The CLI was never on the card.** `UnitRow.svelte` rendered `TaskId · model · effort`
and never the provider, though `session.agentProvider` is on the client mirror and is
already read elsewhere (`format.ts` → `canResume()`).

**The runtime identity was ephemeral.** `codexRuntimeIdentity()` and
`claudeRuntimeIdentity()` parse the effective model (and, for Codex, effort) correctly,
but the poller kept them only in its in-memory `lastActivity` map and pushed them over
SSE. That reaches clients only while a session is live, and the poller probes active
sessions only — so a concluded session or a restarted server lost the value and every
surface fell back to `session.model` / `session.effort`, which are `null` whenever the
pickers were left on default. No `--model` flag is passed then, the CLI chooses for
itself, and only its own runtime log ever names that choice. This is #1823.

Scale: `effort` was `null` on **149 of 159** session rows, so the effort slot read
"default" on nearly every card.

## What changed

- **Persist it.** Nullable `runtimeModel`/`runtimeEffort` columns plus a partial setter —
  an absent field preserves what is stored rather than clearing it, because either source
  can legitimately report only one (Codex has no effort before its first `turn_context`;
  Claude transcripts record none at all). The poller compares per field and writes only on
  change, memoized so a steady session costs neither a read nor a write per probe tick.
- **Backfill it.** `backfillRuntimeIdentity()` fills existing rows once at boot, modelled
  on `backfillCodexSpawnUsage`: one shared rollout tree walk, row-capped, never throws.
  Codex resolves by native session id, Claude by transcript. Rows whose source is gone stay
  NULL and remain eligible. Against a copy of a real database this filled **154 of 156
  candidates in 1.8s**, including TASK-1090 → `gpt-6-astra` / `high`.
- **Show it.** Every card carries a `CODEX` or `CLAUDE` chip. A shared
  `sessionEnvironment()` resolves model and effort observed → configured → default, per
  field, labelling each by its source; the card and the session status bar both read it, so
  the two can no longer disagree about one run. `Viewport` forwards its `activity` signal to
  the bar — the only live carrier while a session runs.

An unknown effort now drops its segment instead of repeating the same default word twice,
which is what made the original report unreadable.

```
TASK-1090 · GPT-6 Astra · Hoch    [CODEX]
TASK-1042 · Opus 5 · Standard     [CLAUDE]
TASK-0911 · Standard              [CODEX]   ← source gone; one default, never two
```

## Notes for review

- **`COLS` is untouched.** It feeds `INSERT INTO sessions (${COLS}) VALUES (?,?,…)` with a
  hand-counted placeholder list, so a column added there without a matching placeholder
  breaks every session create. The two observed fields belong in no INSERT anyway (a fresh
  row is NULL by definition) and are read through a new `READ_COLS`. A store test creates a
  session after the migration so that coupling cannot break silently.
- **The CLI chip owns no `statusTip`,** unlike its neighbours. Those appear only in the
  states they describe; this one is on every card forever, and the shared tooltip takes
  pointer events on purpose — an always-on overlay at the head of the chip row sat between
  the pointer and the chips after it (it broke an existing UnitRow test). The label is the
  information, so a native title carries the rest. The meta line's tooltip was verified to
  leave the task-id button clickable, and that is now pinned by a test.
- **Claude effort stays configured.** Claude transcripts record no reasoning effort, so
  Claude rows are never made backfill candidates for it — otherwise every one would be
  re-read on each boot for a value that can never arrive.

## Out of scope

Live rollout resolution for non-isolated Codex sessions (`CodexTranscriptLocator.pathFor()`'s
`!s.isolated` bail, #1175) — the backfill sidesteps it via `providerSessionId`; the live path
keeps today's behavior. Which model actually gets spawned is unchanged.

## Verification

Root `bun run lint` + `bun run test` (9469 pass); `ui/` `bun run check`, `bun run test`
(4998 pass) and `bun run check:i18n`. `fallow audit` clean.

Refs #1823


kai-osthoff:
🟢✅🎉 READY! @scoop 👀 the merge button is right there, buddy 👆🖱️💅🚀🔥

kai-osthoff:
## Findings

- `ui/src/lib/session-env.ts: The single observed flag turns true when either field is runtime-derived, but the resolver still fills the other field from configuration. For example, `sessionEnvironment()` takes `runtimeModel` from activity/session and independently takes an absent `runtimeEffort` from `session.effort` (lines 41-60). Claude identity extraction only ever returns a model (src/runtime-identity.ts:63-71), so a normal Claude session with an observed model and configured effort is labeled by `unitrow_env_title_observed` / `statusbar_identity_observed_title` as the environment the run actually used (ui/messages/en.json:3905,3907), including an effort that was not observed. Track provenance per segment, or make the tooltip explicitly identify the configured fallback, so the new truthfulness claim is not false for mixed identities.

## Verification

- The runtime fields are added through the idempotent session migration and are read separately from the insert-bound projection (src/store.ts:627-651,4601-4617); all session read paths shown use `READ_COLS` (src/store.ts:2721-2869,3039-3044).
- The poller writes only supplied, changed runtime fields and the store setter preserves an omitted companion field with `COALESCE` (src/poller.ts:1530-1555; src/store.ts:3951-3964).
- The boot backfill selects Codex rows with either field missing and Claude rows only when their model is missing, then resolves Codex by provider session id and Claude through the configured account path (src/store.ts:3981-4009; src/runtime-identity.ts:47-71,89-127).
- The card receives the CLI chip through its shared right-side badge component, and the live activity prop reaches the status bar through Viewport (ui/src/lib/components/unit-row/UnitRowRight.svelte:1-7,125-132; ui/src/lib/components/Viewport.svelte:3114).
- Both maintained locale catalogs contain the new CLI, tooltip, and announcement keys (ui/messages/en.json:3902-3909; ui/messages/de.json:3902-3909). The announcement filename and declared version agree, and the catalog glob loads entry fragments (ui/src/lib/feature-announcements/entries/v1.48.0-card-cli-identity.ts:3-8; ui/src/lib/feature-announcements.ts:47-50).

kai-osthoff (commented):
# Review: runtime CLI identity on task cards

Decision: comment

The prior blocking mixed-provenance finding is resolved. `sessionEnvironment()` resolves observed model and effort independently, records separate provenance flags, and constructs each rendered tooltip sentence from that field's own source (`ui/src/lib/session-env.ts:53-95`). Both the card and status bar consume that resolver (`ui/src/lib/components/UnitRow.svelte:187-190`, `ui/src/lib/components/SessionStatusBar.svelte:34-47`), and the viewport passes the live activity signal through to the status bar (`ui/src/lib/components/Viewport.svelte:3114`).

The persistence path selects and hydrates the new nullable fields (`src/store.ts:645-674`, `src/store.ts:6206-6216`), preserves an absent counterpart with `COALESCE` without changing `updatedAt` (`src/store.ts:3951-3964`), and only submits changed fields after comparing them with the stored row (`src/poller.ts:1530-1555`). The boot reconciliation resolves Codex by provider session ID and Claude by its pinned transcript path (`src/runtime-identity.ts:47-71`), is wired into startup (`src/index.ts:1755-1764`), and candidate selection remains field-aware (`src/store.ts:3968-4000`).

The CLI badge is rendered in the shared card badge row (`ui/src/lib/components/CliBadge.svelte:7-21`, `ui/src/lib/components/unit-row/UnitRowRight.svelte:166-170`). The two maintained message catalogs contain the same seven newly introduced keys (`ui/messages/en.json:3901-3909`, `ui/messages/de.json:3901-3909`).

No builds or tests were run, per the review request.

## Finding

- `ui/messages/de.json:3905`: The new German announcement opens the quote around Standard with `„` but closes it with a literal straight quote (`\"`) rather than `“`. This is user-visible punctuation; use matching German quotation marks.


**Nits (non-blocking):**

- [compliance] ui/messages/de.json: The new announcement closes the German quote around Standard with a straight quote instead of a matching German closing quote.

PR #2237, opened 2026-09-10
feat(usage): lead the usage popover with the window nearest its cap

kai-osthoff:
Redesign of the top bar's Auslastung / usage popover, from a design canvas the operator picked variant **A** from: <https://claude.ai/code/artifact/b2642c5d-5fe6-49b0-a2e0-b1ebbba616e2> (page *Vorher / Nachher*; the rejected directions are on page 2).

## Why

The panel repeated itself. Every window carried a `resets in …` line and a `checked … ago` line under its bar, so four windows spent eight lines saying the same two things — 577 px of popover. Meanwhile the one window actually near its limit (Codex weekly at 68 % in the operator's screenshot) sat at the bottom looking exactly like three idle ones.

## What changed

- **A hero for the binding window.** Whichever window is nearest its cap across *both* CLIs is promoted to a headline reading at the top, chosen by the existing `hottestCapacityWindow` — the same selection the gear menu's one-line gauge already makes. It stays in its section below too: sections remain complete lists while the hero moves between them as load shifts, so a promoted row never leaves a hole.
- **The reset moves into the row** as a countdown (`1h`, `10h`, `4d`), and the provider-confirmation age is stamped once in the header. ~425 px instead of ~577 px at the same information.
- **One layout for desktop and the mobile sheet.** `desktop` now gates positioning only; the inline/stacked fork in `TopBarUsagePopover` is gone.
- The `↻` glyph in the refresh button becomes an inline SVG; its "Refreshes Claude Code" scope note moves to the button's `title` rather than being dropped.

## Freshness stays per-window where it has to

`TopBar.browser.test.ts` encodes an invariant worth keeping: each window keeps its **own** confirmation age, and a missing window must not borrow a sibling's. The single header stamp is therefore used **only** when every main window genuinely shares that age; if one is missing a sample or is older, the stamp is dropped and each row carries its own. A window past its reset also keeps its value qualified as pre-reset (`87 % before reset`) instead of showing a bare percentage that describes the *previous* window.

## Blast radius, deliberately contained

`LimitGaugeRow`, `CodexTokenDetail` and `ModelWeekGauge` are shared with the usage dashboard (`LimitsLens`) and the gear menu, where the per-bar reset and age lines have room and earn their place. An earlier pass changed them and restyled both surfaces; that was reverted. The popover now uses its own `UsageWindowRow` and renders its own Codex body, and those three components are untouched.

## Verification

- `cd ui && bun run check` — 0 errors, 0 warnings
- `cd ui && bun run test` — 306 files, 4979 tests passing
- `bun run check:i18n`, `check:glossary`, `check:announcement-versions`, `check:docs-manifest`, `scripts/check-feature-catalog.sh` — all green
- Rendered the real component in chromium (EN + DE, desktop + touch) and checked it against the approved artboard; German short labels (`5 Std` / `Woche`) were added because `wöchentlich` overruns the 46 px label column.

Two tests were updated rather than worked around: the Fable-bar selector (Fable now renders through the popover's own row) and the per-window reset assertion, which had started passing off `CreditDetail`'s reset lines once the window reset text moved into the row.

## Not covered

The design canvas never showed the extra-credits block (the operator's screenshot had none active); `CreditDetail` is rendered unchanged inside the new Claude section. Worth a look in review.


kai-osthoff:
# Review

## Findings

1. `ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:191`: The promoted hero never applies `hottest.stale` (nor shows the selected provider's age), although the selection explicitly carries that field. `providerDisplayCapacityRows` marks Claude and Codex rows stale from their respective snapshots (`ui/src/lib/components/usage-gauges.ts:407`, `ui/src/lib/components/usage-gauges.ts:413`) and `hottestCapacityWindow` preserves it (`ui/src/lib/components/usage-gauges.ts:418`, `ui/src/lib/components/usage-gauges.ts:444`). The ordinary provider sections do dim stale data (`ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:225`, `ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:262`, `ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:344`), but the new, visually dominant hero renders the same stale 95% window at full strength and presents its countdown as current (`ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:197`, `ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:213`). This is reachable whenever a stale window is the fullest one, so an operator is directed by stale capacity telemetry. Dim/qualify the hero using the selected window's staleness (and, ideally, its own freshness) before promoting it.

## Verification notes

The new message keys are present in both catalogs: `ui/messages/en.json:439`, `ui/messages/de.json:439`, `ui/messages/en.json:3898`, and `ui/messages/de.json:3898`. The feature entry references the corresponding title/body keys at `ui/src/lib/feature-announcements/entries/v1.48.0-usage-popover-compact.ts:6` and `ui/src/lib/feature-announcements/entries/v1.48.0-usage-popover-compact.ts:7`.

The new imports and helper contracts resolve locally: `compactTokens` is exported by `ui/src/lib/format.ts:152`; `windowResetPending` is exported by `ui/src/lib/components/usage-gauges.ts:293`; and the promoted selection is created in `ui/src/lib/components/TopBar.svelte:293` and passed through `ui/src/lib/components/top-bar/TopBarUsage.svelte:131`.

kai-osthoff (commented):
# Review

No findings.

The prior stale-hero finding is addressed: the selected capacity window now carries its provider row's stale state (`ui/src/lib/components/usage-gauges.ts:418`, `ui/src/lib/components/usage-gauges.ts:433`), `TopBar` passes that selected result into the popover (`ui/src/lib/components/TopBar.svelte:293`, `ui/src/lib/components/TopBar.svelte:815`), and the hero applies it as its own dimming class (`ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:188`, `ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:195`, `ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:348`). The regression coverage exercises both stale-Claude and fresh-Codex promotion (`ui/src/lib/components/TopBar.browser.test.ts:1643`, `ui/src/lib/components/TopBar.browser.test.ts:1669`).

The expired observed-window path does not show a current-looking reset countdown: it detects the pending reset (`ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:84`, `ui/src/lib/components/top-bar/TopBarUsagePopover.svelte:213`) and the ordinary row qualifies the retained percentage while awaiting confirmation (`ui/src/lib/components/top-bar/UsageWindowRow.svelte:47`, `ui/src/lib/components/top-bar/UsageWindowRow.svelte:58`). The existing test covers that condition (`ui/src/lib/components/TopBar.browser.test.ts:2046`).

All newly introduced message keys are present in both maintained catalogs (`ui/messages/en.json:439`, `ui/messages/en.json:3898`, `ui/messages/de.json:439`, `ui/messages/de.json:3898`), and the announcement entry is discovered by the catalog glob (`ui/src/lib/feature-announcements/entries/v1.48.0-usage-popover-compact.ts:3`, `ui/src/lib/feature-announcements.ts:47`).

Tests and builds were not run, per review instructions.

kai-osthoff:
✨✅ Everything green, blockers zero. @scoop 👀🛎️ the merge button won't press itself 🚀🍿⏳

PR #2210, opened 2026-09-08
feat(herdr): support verified herdr 0.9.0 compatibility

kai-osthoff:
## Summary

Shepherd currently blocks herdr 0.9.0 as unsupported. Raise the support and spawn ceilings to 0.9.0 and explicitly admit socket protocol 22 after checking the official release against 0.8.2. Versions above 0.9.0 and unverified protocols remain blocked.

Regenerate the protocol schema/types and CLI reference, update boundary/updater/downgrade tests and support documentation, and add an EN/DE What's-New entry. The four recaptured fixture contracts are unchanged. The [compatibility report](https://github.com/erwins-enkel/shepherd/blob/66572094ca45be654f7ca8ddde39151dd1fcbdc3/docs/herdr-compat/0.9.0.md) contains 14 passing checks and supplemental node-pty/upgrade evidence.

The isolated daemon transition restored panes with new terminal IDs and lost the synthetic agent registration. Re-registration, attach/reconnect, resize and new starts worked afterward. This does **not** promise seamless session preservation; the existing update warning and sandbox idle-status advisory remain applicable. Runtime verification was on Linux x86_64, not macOS/Windows.

Three [research issue drafts](https://github.com/erwins-enkel/shepherd/blob/66572094ca45be654f7ca8ddde39151dd1fcbdc3/docs/herdr-compat/0.9.0-research-drafts.md) cover session-preserving updates, SSH machines and scroll/selection APIs. They remain drafts pending separate publication approval; no new features or issues were created.

## Validation

- `bun run herdr:compat -- --candidate 0.9.0 --baseline 0.8.2`: 14 PASS, 0 REVIEW, 0 FAIL.
- Root: `bun run lint`, `bun run typecheck`, `bun run test` (9,158 passed; 16 skipped).
- UI: `bun run check`, `bun run test` (4,791 passed), `bun run check:i18n`, `bun run check:docs-manifest`.
- Docs: `bun run check`, `bun run build`; CLI generation is byte-identical on repeat.
- `bun run check:herdr-types`, `bun run check:announcement-versions`; self-reviewed the final diff.
- All seven pre-push lanes passed on the rebased commit, including UI/extension builds and tests; fallow found no issues in the 25 changed files.

## Checklist

- [x] Conventional-commit title; branch rebased onto `origin/main`, with no merge commits.
- [x] Relevant local lint/check/test commands pass.
- [x] EN/DE catalogs and What's-New entry updated.
- [x] Documentation and generated artifacts updated.

PR #2126, opened 2026-08-27
feat(herdr): support herdr 0.8.2

kai-osthoff:
## Summary

- raise the verified supported and spawnable herdr ceiling from 0.8.0 to 0.8.2
- admit additive socket protocol 20 while keeping protocol 18 excluded and the sandbox-status floor at 0.7.4
- regenerate the protocol schema/types and 0.8.2 CLI reference, add the fresh compatibility report, and announce the bump for Shepherd 1.48.0 in EN/DE

Closes #2096

## Verification

- `bun run herdr:compat -- --candidate 0.8.2 --baseline 0.8.0` — 13 PASS, 0 REVIEW, 0 FAIL
- `bun run lint`
- `bun run test` — 8,799 passed, 16 skipped, 0 failed
- `bun run check:herdr-types`
- `cd ui && bun run check`
- `cd ui && bun run check:i18n`
- `cd ui && CI=1 bun run test --maxWorkers 16` — 4,677 passed
- `cd docs-site && bun run check`
- `cd docs-site && bun run build`
- full pre-push hook with commit-based gates, UI, extension, and Fallow audit

PR #2238, opened 2026-09-10
fix(onboarding): fail fast when an instance has no usable network

### LINKED ISSUE #2229: Onboarding harness runtime has degraded ~16x (14 min → ~4h)

scoop:
Closes #2229.

## The issue's premise was wrong

#2229 attributes the 16x slowdown to in-container throughput and proposes pre-baking scenario images. Measured on the host, that is not what was happening — and the harness was not "slow but passing" either. The Sep 10 nightly reported **0/10 green**, every scenario dying on `Temporary failure resolving 'archive.ubuntu.com'`.

Against a live `images:ubuntu/24.04` instance:

- Containers had **no IPv4 address at all** — DHCPv4 to the bridge was dropped.
- They were left with the `fd42::/64` ULA and an RA-supplied default v6 route; the host has no IPv6 upstream, so that route is a black hole.
- DNS to the bridge resolver timed out (`curl: Resolving timed out after 15000ms`), while the host querying `10.254.46.1` directly answered fine.
- Kernel log, live: `[UFW BLOCK] IN=incusbr0 … DPT=53`.

`ufw` on the Incus host has `DEFAULT_INPUT_POLICY=DROP` + `DEFAULT_FORWARD_POLICY=DROP`, with allow rules for `docker0`/`172.17.0.1` and none for `incusbr0`. Every file in `/etc/ufw/` is dated **2026-08-19 21:54** — a `ufw-docker` install, with an `after.rules-ufw-docker~2026-08-19-215447~` backup alongside. Aug 19 nightly: 14 min, 6/6 green. Aug 20: >2h, killed.

So the 16x was containers waiting out DNS and connect timeouts and retrying. **Pre-baking images would have hidden a host fault and could not have fixed it anyway** — `install-e2e`, `bun install` and `claude install` all need live network by design. That proposal is dropped.

The host was fixed out-of-band with an interface- and port-scoped ruleset (documented in the README), rather than the blanket `ufw allow in on incusbr0` most incus+ufw guides suggest — the blanket form exposes every host service bound to `0.0.0.0` and the whole LAN to instances that install packages and run `claude`.

## What this PR changes

The repo-side defect is what let a host fault cost four hours: `waitForDns()` polled DNS 60×1s and then **returned unconditionally either way**, unchecked.

- **`waitForNetwork` is checked, and means something.** Resolving a name is not evidence of a usable network: incusbr0's RA-supplied IPv6 resolver answers *before* the DHCPv4 lease lands, so a DNS-only probe passes on an instance that cannot fetch anything. That window is what broke `git-missing` with an opaque `apk … Permission denied`. The probe now waits for a global IPv4 address **and** DNS. `getent` and `ip` are on all five image families (verified live), so no fallback resolver and no unprobeable escape hatch.
- **One failure abandons the run.** Every remaining scenario would hit the same wall. Abandoned scenarios reuse `unverifiedResult`, so they report `NOT VERIFIED` and **gate red** — deliberately *not* `HARNESS ERROR`, which is de-gated and would recreate the silence #2226 closed.
- **Per-scenario durations + a run total in the report.** This regression had to be reconstructed from journal timestamps because nothing recorded it.
- **Caps re-sized to measured runtime** — 15 min/scenario, 90 min/run, `TimeoutStartSec=2h`. Not a revert: the pre-#2230 values (30 min/3h) came from #2226 and were also chosen while degraded. The #2226 invariant holds — the unit timeout stays strictly above the harness's own budget.
- **README documents the host firewall prerequisite**, with the scoped rules and the `UFW BLOCK` signature to look for.

## Verification (live, on the Incus host)

| | Before | After |
| --- | --- | --- |
| `node-too-old` | 28 min, red | **59.7 s, PASS** |
| Full run | 4h23m, 0/10, exit 1 | **9m 37s, 6/6 green, exit 0** |
| Broken bridge (DHCPv4 disabled) | 4h23m → ten fake BOOT CRASHes | **1m 2s, NOT VERIFIED, exit 1** |

The fail-fast path was exercised end-to-end by disabling `ipv4.dhcp` on `incusbr0`, not just unit-tested. Every scenario now runs in 42s–1m14s.

`bun run lint`, `bunx tsc --noEmit`, `bun run test` (9591 pass, 0 fail) and `fallow audit` all green.

## Not in scope

`herdr-outdated` regressed to `herdr want=warning got=error` from #2126 (Aug 28) or #2210 (Sep 8) — both landed while the harness was dead. A stale stub against a tightened liveness probe, unrelated to the network; it is `detectionOnly` so it does not gate. Filed separately.

Manual-Step: POST-MERGE: Re-install the nightly systemd unit on the Incus host so the new `TimeoutStartSec=2h` takes effect — the running unit still carries `7h`: `cp ci/onboarding-harness/shepherd-onboarding.{service,timer} ~/.config/systemd/user/ && systemctl --user daemon-reload`

### DIFF
--- a/ci/onboarding-harness/README.md
+++ b/ci/onboarding-harness/README.md
@@ -69,6 +69,33 @@ bun run onboarding:test --scenario install-e2e-service
 
 **Install-time RAM floor:** Claude Code's native installer transiently peaks at ~2 GB RSS during `claude install`. The harness sizes instances at 4 GiB to provide headroom above the ~3 GB floor; hosts below that may OOM-kill the install.
 
+### Host firewall
+
+**Instances must be able to reach the Incus bridge's DHCP and DNS, and route outbound.** Every scenario installs packages from the network, so a host firewall that filters `incusbr0` breaks the entire harness — and does so _slowly_, which is worse than breaking it fast.
+
+This is not hypothetical. On 2026-08-19 a `ufw-docker` install enabled `ufw` on the Incus host with `DEFAULT_INPUT_POLICY=DROP` + `DEFAULT_FORWARD_POLICY=DROP` and allow rules for `docker0` only. Instances stopped getting a DHCPv4 lease **and** lost DNS, fell back to an RA-supplied IPv6 address whose default route was a black hole, and every package operation sat out its own timeout and retried. The nightly went from 14 min to **4h23m** and eventually to 0/10 green ([#2229](https://github.com/erwins-enkel/shepherd/issues/2229)).
+
+On a `ufw` host, grant the bridge exactly what it needs:
+
+```bash
+sudo ufw allow in on incusbr0 to any port 53 proto udp   # DNS
+sudo ufw allow in on incusbr0 to any port 53 proto tcp
+sudo ufw allow in on incusbr0 to any port 67 proto udp   # DHCPv4
+sudo ufw allow in on incusbr0 to any port 547 proto udp  # DHCPv6
+sudo ufw route deny in on incusbr0 to 192.168.0.0/16     # keep instances off the LAN
+sudo ufw route allow in on incusbr0                      # instances → internet
+```
+
+Prefer this over the blanket `ufw allow in on incusbr0` most incus+ufw guides suggest: the blanket form exposes **every** host service bound to `0.0.0.0` (sshd, syncthing, …) to instances that install arbitrary packages and run `claude`. Adjust the LAN range to your own. Order matters — the `route deny` must be added before the `route allow`. These persist in `/etc/ufw/user*.rules` and replay on boot.
+
+To confirm the bridge is being filtered, watch the kernel log while an instance boots:
+
+```bash
+journalctl -k -f | grep 'UFW BLOCK'
+```
+
+`IN=incusbr0 … DPT=53` or `DPT=67` there is the signature. Since #2229 the harness also detects this itself: the first scenario aborts within ~60s with a diagnosis pointing here, and the run stops rather than repeating the failure ten times.
+
 ## Usage
 
 ```bash
@@ -86,11 +113,13 @@ The report lands at `onboarding-gap-report.md` in the working directory. Exit co
 
 ## Time budget
 
-The harness bounds itself so it always finishes on its own terms — **45 min per scenario** (`SHEPHERD_ONBOARDING_SCENARIO_TIMEOUT_MS`) and **6h per run** (`SHEPHERD_ONBOARDING_BUDGET_MS`). A scenario that blows its cap is abandoned; scenarios with no budget left are never started. Either way they are recorded **NOT VERIFIED** — not green, not a harness error, so a gate-eligible one still gates red and the report says plainly that no verdict was reached.
+The harness bounds itself so it always finishes on its own terms — **15 min per scenario** (`SHEPHERD_ONBOARDING_SCENARIO_TIMEOUT_MS`) and **90 min per run** (`SHEPHERD_ONBOARDING_BUDGET_MS`). A scenario that blows its cap is abandoned; scenarios with no budget left are never started. Either way they are recorded **NOT VERIFIED** — not green, not a harness error, so a gate-eligible one still gates red and the report says plainly that no verdict was reached.
 
-Those caps are sized to the runtime the harness **actually has**, not the one it ought to have: a full run was ~14 min until Aug 2026 and is now ~4h, with every scenario taking 20–28 min ([#2229](https://github.com/erwins-enkel/shepherd/issues/2229)). Wall-clock is dominated by in-container package installs and image pulls, which the harness does not control. Sizing the caps to the old runtime would cut off a legitimate, working run and report it NOT VERIFIED — a red release gate on a healthy harness. **Bring both numbers down, with `TimeoutStartSec`, once #2229 restores the runtime.**
+Those caps are sized to **measured** runtime: a full run is 8m55s (slowest scenario ~2 min), or roughly 15 min once the nightly's cold image pulls are included. They sit ~6x above that — loose enough not to cut off a slow night, tight enough that a repeat of [#2229](https://github.com/erwins-enkel/shepherd/issues/2229) is caught in 90 min rather than absorbed silently. They are deliberately not the pre-#2230 values (30 min / 3h): those were also chosen while the harness was degraded.
 
-The service's `TimeoutStartSec` (7h) is a **backstop that must never be reached**, and must always stay above the harness's own budget. When it sat _below_ the worst case (the old 2h), systemd's dirty kill was the guaranteed outcome on a slow night rather than an edge case.
+Every run reports **per-scenario durations and a total** in the gap report, so a runtime regression is visible in the artifact. #2229 had to be reconstructed from journal timestamps because nothing recorded it.
+
+The service's `TimeoutStartSec` (2h) is a **backstop that must never be reached**, and must always stay above the harness's own budget. When it sat _below_ the worst case (the old 2h against a 4h+ run), systemd's dirty kill was the guaranteed outcome rather than an edge case.
 
 ## Run isolation
 
--- a/ci/onboarding-harness/report.ts
+++ b/ci/onboarding-harness/report.ts
@@ -93,8 +93,18 @@ export function statusDescription(results: ScenarioResult[]): string {
     : `${gaps.length} gate gap(s): ${gaps.map((g) => g.scenarioId).join(", ")}${harnessNote}`;
 }
 
+/** Compact wall-clock for the report: `48s`, `2m 31s`. `—` when a scenario was never
+ *  started (budget exhausted, or abandoned), so an absent duration never reads as 0. */
+function formatDuration(ms: number | undefined): string {
+  if (ms === undefined) return "—";
+  const total = Math.round(ms / 1000);
+  const minutes = Math.floor(total / 60);
+  const seconds = total % 60;
+  return minutes ? `${minutes}m ${seconds}s` : `${seconds}s`;
+}
+
 function tableRow(r: ScenarioResult, klass: Classification): string {
-  return `| ${r.scenarioId} | ${r.image} | ${r.detection.detected ? "yes" : "no"} | ${r.appliedVia} | ${r.reachedGreen ? "yes" : "no"} | ${klass} |`;
+  return `| ${r.scenarioId} | ${r.image} | ${r.detection.detected ? "yes" : "no"} | ${r.appliedVia} | ${r.reachedGreen ? "yes" : "no"} | ${formatDuration(r.durationMs)} | ${klass} |`;
 }
 
 function gapEntry(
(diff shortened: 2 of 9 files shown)


scoop (commented):
# Review — fix(onboarding): fail fast when an instance has no usable network

Diff reviewed: `git diff 360cca67d7ec2b3334167b4b4be5615e1428e61f...HEAD` — 9 files, all under
`ci/onboarding-harness/` and `test/onboarding-harness/`.

## What I verified

**The new error type resolves and is consumed consistently.**
`NetworkUnreachableError` is declared at `ci/onboarding-harness/seed.ts:209`, exported, imported at
`ci/onboarding-harness/run.ts:9`, and matched with `instanceof` at `ci/onboarding-harness/run.ts:351`.
The test file imports the same symbol from the same path (`test/onboarding-harness/seed.test.ts:3`).
No spelling/casing divergence.

**The abandon path really does gate red.** I traced the flag through the classifier rather than
assuming it. `runScenario` returns `unverified: true` at `ci/onboarding-harness/run.ts:357`;
`classify()` short-circuits on `r.unverified` first (`ci/onboarding-harness/report.ts:54`) and returns
`NOT VERIFIED`; `isHarnessError()` requires the error to start with `LAUNCH_FAILURE_PREFIX`
(`ci/onboarding-harness/report.ts:21`, prefix defined at `ci/onboarding-harness/incus.ts:8` as
`"incus launch failed: "`), and the thrown message begins with `${scenario.id}: the instance launched…`
(`ci/onboarding-harness/seed.ts:271`), so it is **not** de-gated. `gateGapScenarios()`
(`ci/onboarding-harness/report.ts:76`) therefore includes every gate-eligible abandoned scenario, and
`main()` exits 1 (`ci/onboarding-harness/run.ts:629-632`). The plan's claim that this reuses #2230's
machinery with no new classification holds.

**`abandonRemaining` is safely importable from a test.** `run.ts` guards its entrypoint behind
`import.meta.main` (`ci/onboarding-harness/run.ts:637`) and three existing tests already import from
this module (`test/onboarding-harness/herdr-preflight.test.ts:3`,
`test/onboarding-harness/install-e2e.test.ts:5`, `test/onboarding-harness/install-lifecycle.test.ts:4`),
so the new `test/onboarding-harness/run.test.ts:2` import acquires no host lock.

**The three re-sized numbers agree with each other.** `SCENARIO_TIMEOUT_MS = 15 * 60_000` and
`RUN_BUDGET_MS = 90 * 60_000` (`ci/onboarding-harness/run.ts:446-448`); the README states the same two
figures and the same env-var names (`ci/onboarding-harness/README.md:116`); the unit says
`TimeoutStartSec=2h` with a comment naming "90m per run + 15m per scenario"
(`ci/onboarding-harness/shepherd-onboarding.service:18,25`). The #2226 invariant (backstop strictly
above the harness's own budget: 2h > 90m) is preserved.

**The new report column breaks no consumer.** I grepped for anything that parses
`onboarding-gap-report.md` by column: the only readers are `run.ts` (which writes it) and
`ci/onboarding-harness/issue.ts:118`, which embeds the markdown verbatim into an issue body.
`scripts/onboarding-gate.sh` drives the gate by exit code and scenario id, never by report text — it
enumerates ids from `SCENARIOS` and shells out per scenario. So adding `| Duration |` to the header
(`ci/onboarding-harness/report.ts:187-188`) and to `tableRow` (`ci/onboarding-harness/report.ts:107`)
is safe.

**No VM scenarios exist today.** `Scenario.vm` is declared at `ci/onboarding-harness/types.ts:10` and
passed through at `ci/onboarding-harness/incus.ts:55`, but no entry in
`ci/onboarding-harness/scenarios.ts` sets it, so the probe's reliance on `incus exec` working
immediately after `launch` is fine for the current catalog (see the latent section below).

**`durationMs` has no other consumers.** Grep across `ci/`, `src/`, `test/` shows it read only in
`ci/onboarding-harness/report.ts:107,161,163` and written only at `ci/onboarding-harness/run.ts:487`
— the unrelated `durationMs` names in `scripts/pre-push.ts` and `scripts/usage-report.ts` are separate
local types.

**No repo house rule in the supplied set applies to this diff** (they concern plan-phase gating,
session steering, idle triggers, and toast options — none of which this diff touches).

## Finding

### `ci/onboarding-harness/seed.ts`: the network probe has no wall-clock bound, so the ~60s fail-fast it advertises is not guaranteed — and when the scenario cap wins instead, the abandon-run signal is lost

`waitForNetwork` (`ci/onboarding-harness/seed.ts:239-248`) runs:

```sh
for i in $(seq 1 60); do
  if [ -n "$(ip -4 -o addr show scope global 2>/dev/null)" ] && getent hosts bun.sh >/dev/null 2>&1; then exit 0; fi
  sleep 1
done; exit 1
```

That is **60 iterations**, not 60 seconds. Each iteration costs `1s + the cost of getent`. In the exact
#2229 fault (DHCPv4 *and* DNS dropped) the `&&` short-circuits on the empty address, `getent` never
runs, and the loop really does take ~60s — that is the case the author measured. But in the sibling
fault where the DHCPv4 lease lands and only `udp/53` is filtered, `getent hosts` does not fail fast;
it blocks out the resolver timeout (glibc default `timeout:5 attempts:2`, A then AAAA — 10–20s per
call; musl similar). Sixty of those is 10–20 minutes, not 60 seconds.

Two concrete consequences, both in code this diff introduces:

1. The operator-facing diagnosis asserts a bound the code does not provide: *"never got a usable
   network (no global IPv4 address and/or no DNS after 60s)"* (`ci/onboarding-harness/seed.ts:272`),
   and `ci/onboarding-harness/README.md:97` tells the operator *"the first scenario aborts within
   ~60s"*. Both are stated as fact and are wrong on the DNS-filtered path.
2. Worse, it can silently defeat the abandon-run this PR exists to add. The per-scenario cap is now
   15 min (`ci/onboarding-harness/run.ts:447`) and is applied as
   `Math.min(SCENARIO_TIMEOUT_MS, remaining)` (`ci/onboarding-harness/run.ts:525`). If the probe
   outlasts that cap, `capped` wins the `Promise.race` at `ci/onboarding-harness/run.ts:486` and
   resolves `unverifiedResult(...)` (`ci/onboarding-harness/run.ts:454-463`) — which does **not**
   carry `networkUnreachable`. `runAllScenarios` only abandons on `result.networkUnreachable`
   (`ci/onboarding-harness/run.ts:531`), so the run keeps going and repeats the wall on every
   remaining scenario until the 90-minute budget is gone. That is a smaller version of the ten-times
   repeat the PR set out to stop, and it is reachable on the path that already executes on every
   scenario. Note this also bites late in a healthy run: when `remaining` is small, the cap can fire
   during an ordinary 60s probe and the same flag loss occurs.

The fix is to make the probe's bound wall-clock rather than iteration count — e.g. compute a deadline
once (`end=$(($(date +%s)+60))` and `while [ "$(date +%s)" -lt "$end" ]`, both `date +%s` forms being
available on busybox and coreutils), or wrap the whole probe in `timeout 60 sh -c …`. `archKeyringRefresh`
already uses `timeout` for exactly this reason (`ci/onboarding-harness/seed.ts:95`). Whichever you pick,
the message and README text should describe the bound the code actually enforces.

- `ci/onboarding-harness/seed.ts:239` — for a `vm: true` scenario (`Scenario.vm`, `ci/onboarding-harness/types.ts:10`; wired at `ci/onboarding-harness/incus.ts:55`, currently unused by any entry in `scenarios.ts`), `incus exec` fails immediately after `launch` until the guest agent is up, so the probe would return `false` in seconds and abandon the entire run with a firewall diagnosis; the pre-diff code ignored the probe result and so was immune.

- `ci/onboarding-harness/report.ts:126-155` — `buildGapReport` is additionally split into a new `detailSections()` helper; the plan described only adding a duration column and total, and this refactor of untouched routing logic was not required by the task.
- `ci/onboarding-harness/run.ts:506-541` — extracting `runAllScenarios()` out of `main()` goes beyond the plan's "small exported pure helper" (`abandonRemaining`); reasonable, but it is a drive-by restructure of the loop.
- `ci/onboarding-harness/README.md:72-97` — a full "Host firewall" narrative section (incident history, `ufw` rule set, `journalctl` recipe) is more documentation surface than the plan's "document the host firewall prerequisite", though it is directly useful.

- `ci/onboarding-harness/incus.ts:30` — `RUNNER_TIMEOUT_MS` is 20 min, now larger than the 15-min per-scenario cap, so the per-call backstop can no longer fire before the scenario cap does; it is effectively dead as an inner bound.
- `ci/onboarding-harness/run.ts:486` — when the cap wins the `Promise.race`, the losing `runScenario` keeps running in the background against the same instance; the comment at `ci/onboarding-harness/run.ts:465-467` acknowledges this and relies on `sweep()`.

- I could not verify that busybox's `ip` applet on `images:alpine/3.21` accepts `-4`, `-o` and the
  `scope global` filter, nor that `images:archlinux` / `images:rockylinux/9` ship `iproute2` in their
  linuxcontainers builds — none of that ground truth is in this repo, and the instructions forbid
  running containers. The plan records a live verification on debian/12, rockylinux/9 and alpine/3.21
  and I have taken that as written; I am neither asserting it correct nor raising it as a finding. It
  is worth flagging only that the probe no longer has the pre-diff escape hatch
  (`command -v getent >/dev/null 2>&1 || break`), so if `ip` is absent or rejects the argument form on
  any one image family, the consequence is now an abandoned ten-scenario run reported as a host
  firewall fault rather than a degraded probe (raised as a nit).
- I did not verify the measured runtimes (8m55s full run, ~2 min slowest scenario, `git-missing`
  passing in 48.6s) that the 15m/90m/2h caps are sized against — those are host observations with no
  artifact in the repo.
- Per the instructions I did not run `bun run lint` or `bun run test`; CI owns those.


**Nits (non-blocking):**

- [security] ci/onboarding-harness/README.md:85: the example rule is `ufw route deny in on incusbr0 to 192.168.0.0/16`, but the host this section documents recovering has its LAN on 10.0.50.0/24 per the investigation. A copy-paste recovery on a rebuilt host would leave the LAN fully reachable from instances that install arbitrary packages and run claude. The prose does say to adjust the range, but the documented host's own range is the safer example.

PR #2241, opened 2026-09-10
fix(harness): shape the herdr stubs the way the liveness probe reads them

### LINKED ISSUE #2239: herdr-outdated scenario detects error instead of warning (stale stub vs. tightened liveness probe)
The `herdr-outdated` onboarding scenario now reports `herdr want=warning got=error` and classifies as a DETECTION GAP. It is `detectionOnly`, so it is **not** gate-eligible and does not block a release — but the nightly report is no longer clean.

## It is a real regression, not a long-standing gap

From the nightly journal, this scenario was `DETECTION-ONLY` (i.e. detecting correctly) on every run through **Aug 19**, the last healthy nightly before the harness broke:

```
2026-08-19 | herdr-outdated | images:archlinux | yes | skipped | no | DETECTION-ONLY |
```

It reads `no / DETECTION GAP` on the first healthy run since (Sep 10, after #2229 was fixed).

## Likely cause

The harness stub reports version `0.6.5`, which is below `HERDR_MIN_VERSION` (`0.7.0`), so `versionProbe` in `src/diagnostics.ts` should classify it `warning` via the `outdated` hint. It returns `error` instead — and in that function the only things that outrank the outdated check are the support **ceiling** and the **liveness** probe. `0.6.5` is well under the ceiling, so the liveness probe is the likely one: the stub answers every invocation with a single `{"version":"0.6.5"}` line, and if `runHerdrLiveness` has since come to expect more of the response it now throws, yielding `error` (offline) rather than `warning` (outdated).

Two candidate commits, both of which landed **while the harness was dead** and so were never exercised by it:

- #2126 — `feat(herdr): support herdr 0.8.2` (Aug 28)
- #2210 — `feat(herdr): support verified herdr 0.9.0 compatibility` (Sep 8)

(#2056, herdr 0.8.0 / protocol 19, landed Aug 17 and the Aug 18 + Aug 19 nightlies were still green, so it is ruled out.)

## Worth fixing for its own sake

Either the stub needs to satisfy the current liveness contract, or the scenario's expectation is genuinely stale and should change. Whichever it is, the fixture should be derived from the production contract rather than hand-written, the way the baseline stub already derives its version from `HERDR_LAST_SUPPORTED_VERSION` instead of hardcoding it.

## The meta-point

This is the second-order cost of the #2226 outage: a real fixture regression went unnoticed for ~3 weeks because the nightly that would have caught it was dead. The staleness alerting from #2226 is what should shorten that next time.

Found while fixing #2229 (see #2238); deliberately left out of that PR's scope, which was the runtime regression.

scoop:
Closes #2239.

## What broke

Not #2126 or #2210 as the issue suspected — **#2216**. It replaced diagnostics' exit-code-only `herdr agent list` liveness with `probeHerdrRuntime()`, which now runs **`herdr status --json` first** and parses the document.

Both harness stubs answered *every* invocation with a single `{"version":"…"}` line and exit 0. That parses as a status document with no `client`/`server` keys → `unknown` → liveness throws → `versionProbe` returns `error` (`diagnostics_hint_herdr_unknown`) and never reaches the below-floor `warning` branch. Reproduced directly against the production probe:

```
OLD one-line stub => { state: "unknown", installedVersion: null, reason: "probe_failed" }
```

The scenario's `warning` expectation was never stale: a real live herdr 0.6.x has no `status` command, so production falls through to `probeLegacy()` (`agent list` + `--version`) → `ready` → below `HERDR_MIN_VERSION` → `warning`. The fixture was the lie.

**The baseline stub had the same defect**, so the other six scenarios have been booting with an `error` herdr row since Sep 8 — contradicting `herdrStub()`'s own comment claiming it reads `ok`. Invisible because none of them assert on `herdr`.

## The fix

New `ci/onboarding-harness/herdr-stub.ts` generates both stubs, shaped by the version it is given:

| Stub | Version source | Shape | Probe path | Check |
| baseline | `HERDR_LAST_SUPPORTED_VERSION` | modern: real ready status document | `status --json` → `agent list` | `ok` |
| herdr-outdated | derived one step below `HERDR_MIN_VERSION` | legacy: `status` exits 2, plain-text `--version` | `probeLegacy` fallback | `warning` |

Versions are derived, never hardcoded — the way the baseline already derived its own — so a floor bump can't quietly turn the outdated scenario into a healthy herdr, and the shape follows from the version automatically.

`--version` now answers as real herdr does (a plain line). Checked first that every consumer — boot preflight, `herdr-update`'s version runner, `probeLegacy`, diagnostics, and the harness's own pin assertion — reads it through a semver regex and none JSON-parses it. The catch-all stays valid JSON because the on-loop driver's `list()/tabs()/panes()` do an unguarded `JSON.parse`.

No `src/` changes: production was behaving correctly throughout.

## Why it went unnoticed for three weeks

This is the second-order cost of the #2226 outage the issue calls out. `test/onboarding-harness/herdr-stub.test.ts` executes each **generated** stub against the **real** `probeHerdrRuntime`, so the next time the contract moves it fails in a second instead of in a nightly.

## Verification

- `bun run onboarding:test -- --scenario herdr-outdated` on a real archlinux instance → `| herdr-outdated | images:archlinux | yes | skipped | no | 57s | DETECTION-ONLY |`
- Root `bun run lint`, `bunx tsc --noEmit`, `bun run test` (9597 pass / 0 fail)
- `bunx fallow audit --base origin/main --fail-on-issues` — clean on all 4 changed files

### DIFF
--- a/ci/onboarding-harness/herdr-stub.ts
+++ b/ci/onboarding-harness/herdr-stub.ts
@@ -1,0 +1,115 @@
+import { HERDR_MIN_VERSION } from "../../src/config";
+import { HERDR_LAST_SUPPORTED_VERSION } from "../../src/herdr-capabilities";
+import { compareSemver } from "../../src/semver";
+
+/**
+ * Generates the network-free `herdr` STUBs the harness plants on PATH, in the two shapes the
+ * PRODUCTION liveness probe (`probeHerdrRuntime`, src/herdr-runtime.ts) actually distinguishes.
+ *
+ * Why this module exists: until #2216 diagnostics' liveness was `herdr agent list` with the exit
+ * code as its only signal, so a stub that echoed one `{"version":"…"}` line for every invocation
+ * satisfied it. #2216 made the probe run `herdr status --json` FIRST and parse the document, at
+ * which point that one-line stub started parsing as a status document with no `client`/`server`
+ * keys — `unknown`, i.e. `error` (offline-ish), never the below-floor `warning` the
+ * `herdr-outdated` scenario asserts (#2239). Both stubs were hand-written, so nothing failed until
+ * the nightly ran three weeks later.
+ *
+ * So the shape is derived from the version, and the generated script is proven against the real
+ * probe in test/onboarding-harness/herdr-stub.test.ts rather than trusted by eye.
+ */
+
+/** First herdr version that answers `status --json`. Below it the CLI has no `status` command at
+ *  all, which is why production keeps a fallback — see `probeLegacy` in src/herdr-runtime.ts
+ *  ("v0.8 and earlier have no `status` command") and docs/herdr-compat/0.9.0.md. */
+export const HERDR_FIRST_STATUS_JSON_VERSION = "0.9.0";
+
+/** The version the `herdr-outdated` scenario's stub reports: one release line below
+ *  {@link HERDR_MIN_VERSION}, DERIVED so a floor bump can never quietly turn that scenario's
+ *  seeded defect into a healthy herdr. Prefers a patch step down and falls back to a minor, which
+ *  keeps it a version that could plausibly exist rather than a synthetic `0.0.x`. */
+export function outdatedHerdrVersion(): string {
+  const [major = 0, minor = 0, patch = 0] = HERDR_MIN_VERSION.split(".").map((n) => Number(n) || 0);
+  if (patch > 0) return `${major}.${minor}.${patch - 1}`;
+  if (minor > 0) return `${major}.${minor - 1}.0`;
+  if (major > 0) return `${major - 1}.0.0`;
+  throw new Error(`cannot derive a version below HERDR_MIN_VERSION ${HERDR_MIN_VERSION}`);
+}
+
+/** The `herdr status --json` document a healthy modern daemon returns, in the shape
+ *  `parseHerdrRuntimeStatus` (src/herdr-runtime.ts) reads: client and server on the same version,
+ *  server running, both compatibility flags true. Anything less classifies as `unknown`. */
+function readyStatusDocument(version: string): string {
+  return JSON.stringify({
+    client: { version },
+    server: { running: true, version, compatible: true, endpoint_compatible: true },
+  });
+}
+
+/**
+ * The `#!/bin/sh` body of a stub reporting `version`.
+ *
+ * Modern (≥ {@link HERDR_FIRST_STATUS_JSON_VERSION}): `status --json` returns a ready document, so
+ * the probe's modern path continues to its `agent list` function check, which the catch-all
+ * answers.
+ *
+ * Legacy (below it): `status` fails the way a CLI without that command does, sending the probe
+ * down `probeLegacy` (`agent list` + `--version`) — the real path a live-but-old herdr takes, and
+ * the one that makes `herdr-outdated` read outdated rather than offline.
+ *
+ * The catch-all is load-bearing in BOTH shapes: it exits 0 (the probe's liveness evidence) and
+ * emits valid JSON, because the on-loop `HerdrDriver.list()/tabs()/panes()` do an UNGUARDED
+ * `JSON.parse` then `parsed?.result?.… ?? []`. Plain text there would throw every tick.
+ * `--version`, by contrast, answers the way herdr really does — a plain line — which is safe
+ * because every consumer (boot preflight, herdr-update's version runner, `probeLegacy`,
+ * diagnostics, and the harness's own pin assertion) reads it through a semver regex and none of
+ * them JSON-parses it.
+ */
+export function herdrStubScript(version: string): string {
+  const legacy = compareSemver(version, HERDR_FIRST_STATUS_JSON_VERSION) < 0;
+  const status = legacy
+    ? ["  echo 'unknown command: status' >&2", "  exit 2"]
+    : [`  echo '${readyStatusDocument(version)}'`, "  exit 0"];
+  return [
+    "#!/bin/sh",
+    'if [ "$1" = "status" ]; then',
+    ...status,
+    "fi",
+    'if [ "$1" = "--version" ]; then',
+    `  echo 'herdr ${version}'`,
+    "  exit 0",
+    "fi",
+    "echo '{\"result\":{}}'",
+    "exit 0",
+  ].join("\n");
+}
+
+/** The seed COMMAND that installs {@link herdrStubScript} at `~/.local/bin/herdr`. The trailing
+ *  `test -x` gives the command a non-zero exit when the write didn't land — load-bearing in the
+ *  BASELINE, whose steps are checked (a bad write fail-closes the seed rather than booting a
+ *  scenario against a herdr that isn't there); scenario seeds tolerate non-zero by design. */
+export function herdrStubCommand(version: string): string {
+  return [
+    'mkdir -p "$HOME/.local/bin"',
+    "cat > \"$HOME/.local/bin/herdr\" <<'HERDR_STUB'",
+    herdrStubScript(version),
+    "HERDR_STUB",
+    'chmod +x "$HOME/.local/bin/herdr"',
+    'test -x "$HOME/.local/bin/herdr"',
+  ].join("\n");
+}
+
+/**
+ * The baseline stub every scenario starts from. It exists so Shepherd's boot preflight
+ * (`herdr --version`) passes without a live `herdr.dev` fetch: since #1313 a MISSING herdr
+ * fail-fasts (exit 78) before the HTTP server binds, so the scenarios that don't test herdr still
+ * need one present.
+ *
+ * It reports HERDR_LAST_SUPPORTED_VERSION, derived — never hardcoded. It used to report
+ * `99.99.99`, which was fine until #1887 added the support CEILING: from then on the baseline
+ * represented a herdr Shepherd REFUSES to drive, so every scenario booted with an `unsupported`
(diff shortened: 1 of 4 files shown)


scoop (commented):
# Review — #2239 shape the herdr stubs the way the liveness probe reads them

**Decision: comment.** No correctness, security, contract or scope defect found. The fixture now
walks the real probe on both shapes, and I confirmed the classification end-to-end against the
production code rather than by eye.

**The legacy shape reaches `warning`, which is what the task asks for.**
`probeHerdrRuntime` runs `["status", "--json"]` first (`src/herdr-runtime.ts:328-334`). The
generated 0.6.0 stub answers `$1 = "status"` with `unknown command: status` on stderr and `exit 2`
(`ci/onboarding-harness/herdr-stub.ts:70`). That result is not timed-out/aborted/spawn-errored, and
`isHerdrProtocolMismatch` needs an embedded JSON `code: "protocol_mismatch"` envelope
(`src/herdr-runtime.ts:88-93`, `137-157`) which plain prose does not carry — so the probe falls to
`probeLegacy` (`src/herdr-runtime.ts:344-349`). There `agent list` hits the catch-all (`exit 0`,
`{"result":{}}`) → `commandSucceeded` true → `classifyAgentProbe` returns the `ready` seed
(`src/herdr-runtime.ts:276-283`, `299-311`), and `--version` yields `herdr 0.6.0`, which
`parseHerdrVersion`'s `SEMVER_RE` extracts (`src/herdr-capabilities.ts:39-46`).
`defaultHerdrLiveness` therefore resolves (`src/diagnostics.ts:336-341`), so `versionProbe` passes
the ceiling gate, passes liveness, and reaches `compareSemver(version, floor) < 0` →
`{state: "warning", hintKey: "diagnostics_hint_herdr_outdated"}` (`src/diagnostics.ts:1272-1286`).
`HERDR_MIN_VERSION` is `0.7.0` (`src/config.ts:77`) and `outdatedHerdrVersion()` derives `0.6.0`
(patch is 0 → minor step down, `ci/onboarding-harness/herdr-stub.ts:31-33`), which is below it.
The scenario still expects `warning` (`ci/onboarding-harness/scenarios.ts:120`).

**The modern (baseline) shape reaches `ok`.** `HERDR_LAST_SUPPORTED_VERSION` is `0.9.0`
(`src/herdr-capabilities.ts:18`), so `compareSemver(version, HERDR_FIRST_STATUS_JSON_VERSION)` is 0
→ not legacy → the status document is emitted. I walked that document through
`parseHerdrRuntimeStatus` (`src/herdr-runtime.ts:44-87`): `client.version` present,
`server.running` a boolean `true`, `serverVersion === installedVersion`, no `update` /
`restart_needed` key so `restartNeeded` is `undefined`, and both `compatible` and
`endpoint_compatible` are strictly `true` → `{state:"ready", installedVersion, serverVersion}`.
`probeModernReady` then runs `agent list`, which the catch-all satisfies
(`src/herdr-runtime.ts:284-297`). `versionProbe` then reads `0.9.0`: not above the ceiling
(`compareSemver` is 0, the gate is `> 0`), liveness ok, not below the floor → `ok`.

**The `--version` change from JSON to plain text is safe.** I resolved every consumer:
boot preflight `execFileSync(config.herdrBin, ["--version"])` → `parseHerdrVersion`
(`src/index.ts:229-239`); the harness pin assertion uses `/(\d+\.\d+\.\d+)/` on stdout+stderr
(`ci/onboarding-harness/run.ts:61-72`); `probeLegacy` uses `parseHerdrVersion`
(`src/herdr-runtime.ts:305`); diagnostics uses `SEMVER_RE` via `parseVersion`
(`src/diagnostics.ts:1246-1249`); `herdr-update`'s version runner is
`execFileSync(config.herdrBin, ["--version"])` (`src/herdr-update.ts:316`). None `JSON.parse` it.

**The catch-all still satisfies the on-loop driver.** Every driver call that parses JSON —
`agent list`, `tab list`, `pane list`, `pane process-info` (`src/herdr.ts:973`, `987`, `998`,
`1030`) — receives `{"result":{}}`, which parses and degrades to the same empty result the previous
`{"version":…}` line did.

**`herdr status` has exactly one caller.** Grepped the tree: `src/herdr-runtime.ts:333` is the only
`["status", "--json"]` invocation, so making the legacy stub fail `status` cannot affect any other
path.

**The baseline flip (six scenarios go from an `error` herdr row to `ok`) is safe.**
`assertDetection` only compares the ids in `scenario.expect` (`ci/onboarding-harness/assert.ts:11-17`),
and only `herdr-missing` (`error`), `herdr-outdated` (`warning`) and the two installE2E scenarios
(`ok`, which skip the baseline entirely, `ci/onboarding-harness/seed.ts:248-257`) name `herdr` at
all (`ci/onboarding-harness/scenarios.ts:94`, `120`, `170`, `194`). Nor does the flip change the
apply dispatch: `remediationsFor(before)` is snapshot-wide (`ci/onboarding-harness/run.ts:290`) but
`diagnostics_hint_herdr_unknown` has no entry in `src/remediations.ts:154-170`, so the stale row was
never producing a verbatim remediation; and `herdr-outdated`, whose row now carries the
remediation-bearing `diagnostics_hint_herdr_outdated`, short-circuits on `detectionOnly` before
apply (`ci/onboarding-harness/run.ts:284-289`).

**Seed plumbing.** Both baseline and scenario seed commands are executed as
`driver.exec(id, ["sh", "-c", cmd])` (`ci/onboarding-harness/seed.ts:261`, `269`) and `exec` passes
argv straight to `incus exec … --` with no shell re-parse (`ci/onboarding-harness/incus.ts:63-65`),
so the multi-line quoted heredoc survives; collapsing the scenario's three seed entries into one is
fine, and the trailing `test -x` is tolerated there by design (`ci/onboarding-harness/seed.ts:267-270`).
The existing harness assertions still match the generated text — `seed.test.ts:62` looks for
`.local/bin/herdr` + `chmod +x`, both present at `ci/onboarding-harness/herdr-stub.ts:93-96`, and
`herdr-preflight.test.ts:65` only inspects the untouched `herdr-missing` seed.

**Root-cause attribution in the comments is accurate.** `git log --diff-filter=A -- src/herdr-runtime.ts`
returns `1a350b1da feat(herdr): recover incomplete updates with explicit confirmation (#2216)`, and
`git show 1a350b1da -- src/diagnostics.ts` shows the removed exit-code-only
`spawn(bin, ["agent","list"], {stdio:"ignore"})` liveness — exactly what
`ci/onboarding-harness/herdr-stub.ts:9-15` claims.

**No `src/` file is touched** (`git diff --stat`: only `ci/onboarding-harness/herdr-stub.ts`,
`scenarios.ts`, `seed.ts` and `test/onboarding-harness/herdr-stub.test.ts`).

**Import-boundary check.** The new `import { HERDR_MIN_VERSION } from "../../src/config"` drags
config's import-time work (forge map load, `applyHerdrSocket`) into `scenarios.ts`/`seed.ts`. That is
not new for the harness process — `ci/onboarding-harness/run.ts:22` already imports
`src/remediations`, whose first line is `import { config } from "./config"`
(`src/remediations.ts:1`) — and `deploy/provision.ts:21` sets the precedent for importing
`HERDR_MIN_VERSION` from config outside `src/`. Not a finding.

## Stated limitations

- I could not verify against a real herdr binary what herdr 0.6.x/0.8.x actually do with
  `status --json` — no herdr release is present in this repo. The stub's legacy behaviour is
  consistent with the repo's own statement of that contract (`src/herdr-runtime.ts:344-346`), and
  the classification it produces is proven against the production probe, so this does not block.
- The end-to-end incus run the plan promised (step 6) is not something I can confirm from the diff;
  the PR body's claim of a DETECTION-ONLY run is taken as stated.

- `ci/onboarding-harness/herdr-stub.ts:34` — the major-step-down fallback returns `${major-1}.0.0`, which for a future floor of `1.0.0` yields exactly the synthetic `0.0.0` the doc comment two lines up says the fallback avoids; behaviour stays correct (still below floor, still legacy-shaped), only the comment's promise breaks.

- `src/diagnostics.ts:99-107` — the `runHerdrLiveness` doc comment still describes the pre-#2216 probe ("run `herdr agent list` … Exit code only … Default `spawn(config.herdrBin, ["agent","list"], {stdio:"ignore"})`"), which is the same stale-documentation class of bug this PR fixes in the fixture.
- `src/herdr-runtime.ts:345` — "v0.8 and earlier have no `status` command" reads as absolute, but the repo's own compat probes record a passing `status server` on the 0.8.2 baseline (`docs/herdr-compat/0.8.2.md:112`, `scripts/herdr-compat/probes.ts:44`); the accurate claim is about the `status --json` runtime document.


**Nits (non-blocking):**

- [bug] ci/onboarding-harness/herdr-stub.ts:21-23 — the doc on HERDR_FIRST_STATUS_JSON_VERSION says that below 0.9.0 the CLI has no `status` command at all, but the repo's own compat evidence records a passing `status server` probe on the 0.8.2 baseline (docs/herdr-compat/0.8.2.md:112, scripts/herdr-compat/probes.ts:44). The precise claim the constant actually encodes is that pre-0.9 herdr does not answer `status --json` with the runtime document; wording it that way would keep the constant honest if the boundary ever becomes reachable.
- [bug] test/onboarding-harness/herdr-stub.test.ts:78-84 — the shape test hardcodes that the derived outdated version is always legacy (`toBeLessThan(0)` vs HERDR_FIRST_STATUS_JSON_VERSION, plus `toContain("unknown command: status")`). A floor bump past 0.9.0 would make outdatedHerdrVersion() return a modern-shaped-but-still-below-floor version, which the generator handles correctly, yet this test would fail as if the fixture were broken. Asserting the shape that the derived version implies (branch on the comparison) would keep the guard without the false alarm.

PR #2243, opened 2026-09-10
fix(settings): keep the settings dialog reactive after a short PUT response (#2240)

### LINKED ISSUE #2240: Coding CLI settings: role model-guidance line freezes on the global default after changes made while AGENT ENVIRONMENTS is collapsed
Settings → Coding CLI → AGENT ENVIRONMENTS: each role row renders a `<ModelGuidance>` line (cost chip · tier chip · fitness text) under its `Effective: …` line. Under the repro below that guidance line freezes on the **global default model's** guidance and never refreshes, while the `Effective:` line right above it stays correct.

Found while shooting a marketing screenshot of this pane; the screenshot showed Planner set to `Codex · gpt-6-astra` with an Opus blurb under it.

## Repro

Verified twice against `ui/` dev (`SHEPHERD_DEMO=1 bun run dev`, Chromium). Nothing in the path is demo-specific — the guidance is computed client-side in the panel.

1. Open Settings → **Coding CLI**. Leave **AGENT ENVIRONMENTS** collapsed.
2. Still collapsed, change GLOBAL DEFAULTS → **Claude Code model** from `Auto` to `Opus`, then set **Planner (plan reviewer)** CLI = `Codex`, model = `gpt-6-astra`.
3. Expand **AGENT ENVIRONMENTS**.

Observed on the Planner row:

- `Effective: Codex · gpt-6-astra` ✅
- guidance: `High cost · $$$ strong — Always the latest Opus. Use for PR criticism, planning, or risky refactors…` ❌ (that is the guidance for the *global* `opus`, not for `gpt-6-astra`, which should read `Premium cost · $$$$ max — Most capable Codex model…`)

It stays stuck afterwards, and not only on the row that was touched while collapsed:

- Planner model → `gpt-5.6-luna`: `Effective:` updates, guidance still says Opus.
- Doc agent (never touched while collapsed) → `Claude Code` + `haiku`: `Effective: Claude Code · haiku`, guidance still says `High cost · $$$ strong … Opus`.

Only a page reload clears it. If the group is expanded **before** any of those changes, the guidance tracks correctly — so the trigger is state changing while the group is collapsed.

## Why it matters

That line is the cost/fitness signal the picker exists for. Post-repro it under-reports a premium pick (`gpt-6-astra`, `fable` → shown as `$$$ strong`) and over-reports a budget one (`haiku` → shown as `$$$ strong`). Saved settings and spawn behaviour are unaffected — the `Effective:` line and the persisted values are correct — so this is display-only, but it misleads exactly where the operator is deciding what unattended reviewers cost.

## Where to look

- `ui/src/lib/components/settings/SettingsCodingCliPanel.svelte` → `roleRow` snippet: `<ModelGuidance provider={roleGuidanceProvider(role)} model={roleGuidanceModel(role)} … />`. Both that and the correct `Effective:` line read the same `resolvedRoleEnv(role)`, so the divergence is in how the props reach the child, not in the resolver.
- `ui/src/lib/components/ModelGuidance.svelte` → `const guidance = $derived(modelGuidance(provider, model, context))`.
- `ui/src/lib/components/settings/SettingsGroup.svelte` — the collapsed/expanded wrapper is the trigger; worth checking how it renders children while collapsed.

Root cause not pinned down; the repro is reliable.

A regression test belongs in a browser spec next to the other settings specs (`bun run test` in `ui/`): change a role's CLI+model while the group is collapsed, expand, assert the guidance text matches the row's model and not the global default.

scoop:
Closes #2240.

## The issue's diagnosis was wrong

`ModelGuidance` and `SettingsGroup` are both innocent. The collapsed group keeps its children
mounted and they update normally while hidden — I reproduced the freeze with AGENT ENVIRONMENTS
**expanded** the whole time. (The stated repro also isn't literally performable: a role's CLI
select sits inside the `hidden` group.)

## What actually happens

Reproduced live against `SHEPHERD_DEMO=1 bun run dev` in Chromium, with a `window.onerror` hook and
a temporary `data-dbg="{provider}|{model}"` attribute on `ModelGuidance`:

1. Changing GLOBAL DEFAULTS → Claude Code model calls `saveDefaultModel()`, which does
   `defaultModel = r.defaultModel` with no shape check.
2. `PUT /api/settings` had **no route** in `ui/src/lib/demo/router.ts`, so it fell through to the
   permissive `json({ ok: true })` tail — `r.defaultModel` is `undefined`.
3. `SettingsCodingCliPanel`'s `is1mModel` (`$derived(defaultModel.endsWith("[1m]"))`) throws
   **inside Svelte's `Batch.flush` → `#traverse` → `update_derived`**.
4. That uncaught throw abandons the batch traversal. Every effect not yet processed stays dirty and
   is never revisited — the whole Settings subtree freezes until a page reload.

So it isn't a guidance bug at all: after the throw, every *other* role row's `Effective:` line
freezes too. The debug attribute showed the child component's **props** stale (`claude|opus`) while
the sibling `Effective:` text in the same snippet correctly read `Codex · provider default` — the
child's effects had simply stopped running.

**Not reachable in production today** — the real `putDefaultModel` in `src/server.ts` always echoes
`{ defaultModel }`. This is a demo-mode trigger plus an unguarded client assumption, and the demo is
the marketing surface the bug was found on.

## The fix

**`ui/src/lib/demo/router.ts` + `state.ts`** — handle `PUT /api/settings`: apply the single patched
field to `world.settings` and echo the shape the real server echoes. Three fields answer differently
and are mirrored field by field, because `api.ts` is typed against them:

| Patched field | Response |
| --- | --- |
| `repoRoot` | full `Settings` (`putRepoRoot` is typed `Promise<Settings>`) |
| `anthropicApiKey` | `{ hasApiKey }` — the key is never stored or echoed |
| `authMode` | `{ authMode, hasApiKey }` |
| everything else | `{ [field]: value }` |

This fixes every settings control in the demo at once, not just the model picker. Like the rest of
the demo router it never validates — a 400 would only make the showcase worse.

**`ui/src/lib/components/Settings.svelte`** — the three global-default savers now adopt only a
string from the response, the same guard the per-role savers in `SettingsCodingCliPanel` already
use. A malformed response can no longer poison a required-string prop and wedge the tree.

## Tests

- `ui/src/lib/demo/router.test.ts` — the new route's echo shapes, the round-trip through
  `GET /api/settings`, and that the API key never comes back.
- `ui/src/lib/components/Settings.browser.test.ts` — regression test: with `putDefaultModel`
  resolving `{}`, a role's guidance must still track that role's model. Verified red against the
  unguarded code, failing with exactly the diagnosed throw at `SettingsCodingCliPanel.svelte:199`
  inside `Batch.flush`.

`cd ui && bun run check` clean; `bun run test` green (312 files / 5060 tests). Also re-ran the live
repro after the fix: guidance tracks the role's model and the console stays clean.

## Deliberately out of scope

A `<svelte:boundary>` around the settings panels (so any future throw degrades one panel instead of
the dialog) — considered and left out with Patrick. Also untouched: `ModelGuidance.svelte` /
`SettingsGroup.svelte` (not at fault), the boolean savers `fableAvailable` / `reducedPushMode` (an
`undefined` there renders wrong but cannot throw), and `src/server.ts` (already correct).

### DIFF
--- a/ui/src/lib/components/Settings.browser.test.ts
+++ b/ui/src/lib/components/Settings.browser.test.ts
@@ -9,6 +9,7 @@ import {
   verifyApiKey,
   putAnthropicApiKey,
   putDefaultCodexModel,
+  putDefaultModel,
   fixDiagnostic,
   putRoleEffort,
   putDefaultEffort,
@@ -64,6 +65,7 @@ const mockGetSettings = vi.mocked(getSettings);
 const mockVerify = vi.mocked(verifyApiKey);
 const mockPutKey = vi.mocked(putAnthropicApiKey);
 const mockPutCodexModel = vi.mocked(putDefaultCodexModel);
+const mockPutModel = vi.mocked(putDefaultModel);
 const mockFix = vi.mocked(fixDiagnostic);
 
 function settings(over: Partial<SettingsPayload> = {}): SettingsPayload {
@@ -161,6 +163,8 @@ beforeEach(() => {
   mockPutKey.mockReset();
   mockPutCodexModel.mockReset();
   mockPutCodexModel.mockImplementation(async (model) => ({ defaultCodexModel: model }));
+  mockPutModel.mockReset();
+  mockPutModel.mockImplementation(async (model) => ({ defaultModel: model }));
   // Default seed: api-key mode, key configured → Verify button renders.
   mockGetSettings.mockResolvedValue(settings());
 });
@@ -441,6 +445,52 @@ describe("Settings default coding environment", () => {
       m.settings_role_model_effective({
         model: `${m.settings_cli_claude()} · ${m.model_configured_opus_latest()}`,
       }),
+    );
+  });
+
+  // #2240: a settings PUT that answers 200 WITHOUT the field it owns must not poison the local
+  // value. `defaultModel = undefined` throws inside SettingsCodingCliPanel's `is1mModel`
+  // ($derived(defaultModel.endsWith("[1m]"))) during Svelte's flush; that abandons the batch, so
+  // every effect not yet processed stays dirty forever and the whole dialog stops reacting until a
+  // page reload. The visible tell was a role's guidance line frozen on the global default's copy
+  // while its own "Effective:" line kept tracking — so this asserts on the guidance line.
+  //
+  // Both halves matter: without the toHaveValue check the role pickers could be inert for an
+  // unrelated reason, and without the not.toContain the assertion would pass on a guidance line
+  // that merely happens to mention several models.
+  it("a settings PUT that omits its own field leaves the dialog reactive", async () => {
+    mockGetSettings.mockResolvedValue(
+      settings({ defaultAgentProvider: "claude", defaultModel: "opus", plannerCli: "inherit" }),
+    );
+    // Exactly what the demo router used to answer for every unhandled settings mutation.
+    mockPutModel.mockResolvedValue({} as { defaultModel: string });
+    await mountCodingAgents();
+
+    const { disclosure, button } = requiredCodingSectionButton(m.settings_role_models_title());
+    if (button.getAttribute("aria-expanded") === "false") await disclosure.click();
+
+    await page.getByTestId("default-environment-model").selectOptions("sonnet");
+    await vi.waitFor(() => expect(mockPutModel).toHaveBeenCalledWith("sonnet"));
+
+    const cli = page.getByRole("combobox", {
+      name: m.settings_role_cli_label({ role: roleTitle("planner") }),
+    });
+    await cli.selectOptions("codex");
+    const roleModel = page.getByRole("combobox", {
+      name: m.settings_role_model_label({ role: roleTitle("planner") }),
+    });
+    await roleModel.selectOptions("gpt-6-astra");
+    await expect.element(roleModel).toHaveValue("gpt-6-astra");
+
+    const row = (cli.query() as HTMLElement | null)?.closest(".rrow");
+    expect(row, "planner role row").not.toBeNull();
+    await vi.waitFor(() => {
+      const guidance = row!.querySelector(".model-guidance");
+      expect(guidance?.textContent).toContain(m.model_guidance_codex_6_astra());
+    });
+    // The frozen line reported in #2240 was the GLOBAL default's guidance.
+    expect(row!.querySelector(".model-guidance")!.textContent).not.toContain(
+      m.model_guidance_claude_opus(),
     );
   });
 });
--- a/ui/src/lib/components/Settings.svelte
+++ b/ui/src/lib/components/Settings.svelte
@@ -198,13 +198,20 @@
   let repoRootDisplay = $state<string | null>(null);
   let settingsLoaded = $state(false);
 
+  // The three global-default savers adopt only a string from the response — the same guard the
+  // per-role savers in SettingsCodingCliPanel use. These values are required-string props that
+  // panel dereferences inside `$derived`s (`defaultModel.endsWith("[1m]")`); a response missing
+  // the field would make one throw during Svelte's flush, which aborts the batch and leaves the
+  // whole dialog stale until a reload (#2240). A short response now leaves the last good value.
   async function saveDefaultModel() {
     if (defaultModelBusy) return;
     defaultModelBusy = true;
     try {
       const r = await putDefaultModel(defaultModel);
-      defaultModel = r.defaultModel;
-      defaultModelSaved = r.defaultModel;
+      if (typeof r.defaultModel === "string") {
+        defaultModel = r.defaultModel;
+        defaultModelSaved = r.defaultModel;
+      }
     } catch {
       // revert to the last server-confirmed value; surface the failure as a
       // 12s, deduped alert so the no-op never looks like a save.
@@ -223,8 +230,10 @@
     defaultCodexModelBusy = true;
     try {
       const r = await putDefaultCodexModel(defaultCodexModel);
-      defaultCodexModel = r.defaultCodexModel;
-      defaultCodexModelSaved = r.defaultCodexModel;
+      if (typeof r.defaultCodexModel === "string") {
+        defaultCodexModel = r.defaultCodexModel;
+        defaultCodexModelSaved = r.defaultCodexModel;
+      }
     } catch {
       defaultCodexModel = defaultCodexModelSaved;
       toasts.info(m.settings_default_codex_model_save_failed(), {
@@ -241,8 +250,10 @@
     defaultAgentProviderBusy = true;
     try {
       const r = await putDefaultAgentProvider(defaultAgentProvider);
-      defaultAgentProvider = r.defaultAgentProvider;
-      defaultAgentProviderSaved = r.defaultAgentProvider;
+      if (typeof r.defaultAgentProvider === "string") {
+        defaultAgentProvider = r.defaultAgentProvider;
+        defaultAgentProviderSaved = r.defaultAgentProvider;
+      }
     } catch {
       defaultAgentProvider = defaultAgentProviderSaved;
       toasts.info(m.settings_default_agent_provider_save_failed(), {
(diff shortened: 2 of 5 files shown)


scoop (commented):
# Review — fix(settings): keep the settings dialog reactive after a short PUT response (#2240)

Diff reviewed: `80ad1a14b...HEAD` — 5 files (`ui/src/lib/components/Settings.svelte`,
`ui/src/lib/components/Settings.browser.test.ts`, `ui/src/lib/demo/router.ts`,
`ui/src/lib/demo/state.ts`, `ui/src/lib/demo/router.test.ts`).

## Verdict

No blocking defects found. The two-part fix (demo router route + client shape guard) is
sound, and every identifier, echo shape and message key it depends on checks out against the
tree. Nits only.

## What I verified

**The root cause the diff targets is real and correctly located.**
- `is1mModel` is `$derived(defaultModel.endsWith("[1m]"))` at
  `ui/src/lib/components/settings/SettingsCodingCliPanel.svelte:199`, and `defaultModel` is a
  required-`string` bindable prop (`SettingsCodingCliPanel.svelte:61,74`). It is read in an
  always-rendered branch (`SettingsCodingCliPanel.svelte:604`, inside the always-visible GLOBAL
  DEFAULTS group), so an `undefined` assignment does reach a dereference on every render — the
  premise of the fix holds.
- Before the diff, `PUT /api/settings` had no route in the demo router; the mutation chain
  (`ui/src/lib/demo/router.ts:330-337` pre-change) fell to the permissive tail at
  `ui/src/lib/demo/router.ts:350`, which answers `{ ok: true }`. So `r.defaultModel` really was
  `undefined` in demo mode.
- Production was never affected: `putDefaultModel` always echoes `{ defaultModel }`
  (`src/server.ts:5563-5569`), `putDefaultCodexModel` echoes `{ defaultCodexModel }`
  (`src/server.ts:5571-5578`), `putDefaultAgentProvider` echoes `{ defaultAgentProvider }`
  (`src/server.ts:5651-5657`). The plan's claim on this point is accurate.

**The client guard matches the pattern its comment claims.**
`ui/src/lib/components/Settings.svelte:211,233,253` now use `typeof … === "string"`, which is
exactly the shape check the role savers already apply at
`ui/src/lib/components/settings/SettingsCodingCliPanel.svelte:285`, `:323` and `:350`. The
comment at `Settings.svelte:201-205` is factually correct about `is1mModel`
(verified against `SettingsCodingCliPanel.svelte:199`).
All three response-adopting string savers in the file are covered — the only remaining
`= r.<field>` assignments are `Settings.svelte:274` (`fableAvailable`) and `:290`
(`reducedPushMode`), both booleans (see the latent section below).

**The demo echo shapes mirror the server, field by field.**
- `authMode` → demo returns `{ authMode, hasApiKey }` (`ui/src/lib/demo/state.ts:488`), matching
  `src/server.ts:5690` and the `api.ts` type at `ui/src/lib/api.ts:801`.
- `anthropicApiKey` → demo returns `{ hasApiKey }` only and never stores the key
  (`ui/src/lib/demo/state.ts:479-482` — the early return precedes the generic assignment), and
  the set/clear predicate `typeof value === "string" && value.trim().length > 0` is the same one
  the server uses at `src/server.ts:5723`/`5730`. Matches `ui/src/lib/api.ts:806`.
- Every other field → `{ [field]: value }` (`ui/src/lib/demo/state.ts:489`), matching the
  one-field-per-call echo of `SETTING_PATCHES` (`src/server.ts:5445`, dispatched at
  `src/server.ts:5434-5436`).
- Every caller of `patchSettings` in `ui/src/lib/api.ts` (lines 312–822) sends exactly one field,
  so the demo's `Object.entries(body)[0]` selection (`ui/src/lib/demo/router.ts:324-327`) picks
  the same field the server would dispatch on. I grepped all 30 call sites; none sends two.
- The new handler is exact-path (`method !== "PUT" || path !== "/api/settings"`), so
  `POST /api/settings/verify-key` is untouched, and no other demo route claims that path
  (`ui/src/lib/demo/router.ts:84` is the GET).
- `world.settings` is restored per test/reload by `structuredClone(SEED)` in `demoState.reset()`
  (`ui/src/lib/demo/state.ts:71-76`), and `router.test.ts:14` calls it in `beforeEach`, so the new
  mutating tests do not leak. Stray keys written by the untyped assignment are wiped by the same
  clone.
- `repoRootDisplay: string` and `hasApiKey: boolean` both exist on the `Settings` type
  (`ui/src/lib/types.ts:110`, `:177`), so the narrow casts in `state.ts` type-check as written.

**The browser regression test resolves.**
- `m.model_guidance_codex_6_astra` and `m.model_guidance_claude_opus` both exist in BOTH catalogs:
  `ui/messages/en.json:3875` / `ui/messages/de.json:3875` and `ui/messages/en.json:2634` /
  `ui/messages/de.json:2634`. The diff adds no new strings, so there is no parity gap.
- `modelGuidance("codex", "gpt-6-astra")` returns `m.model_guidance_codex_6_astra()`
  (`ui/src/lib/model-guidance.ts:139-140`), and `gpt-6-astra` is in `CODEX_MODELS`
  (`ui/src/lib/types.ts:2397`), which is what `roleModelOptions()` offers
  (`SettingsCodingCliPanel.svelte:228-231`) — so the option the test selects exists.
- `.rrow` wraps both the role CLI combobox and the row's `<ModelGuidance>`
  (`SettingsCodingCliPanel.svelte:777` … `:835-839`), and `ModelGuidance` renders the
  `model-guidance` class (`ui/src/lib/components/ModelGuidance.svelte:24`), so
  `cli.closest(".rrow").querySelector(".model-guidance")` targets the right node.
- `mountCodingAgents`, `requiredCodingSectionButton` and `roleTitle` all exist in/for the file
  (`Settings.browser.test.ts:180`, `:211`, `:19`), and the new `mockPutModel` reset/impl at
  `:166-167` mirrors the existing `mockPutCodexModel` handling at `:164-165`.
- The negative half discriminates: the seeded `plannerCli: "inherit"` with
  `defaultModel: "opus"` means the frozen state really would render the Opus copy in that row.

## Stated limitations

- I did not run the tests or the demo (`bun run check` / `bun run test` are CI's job per the
  brief), so the claim that the new browser test fails before the guard is the author's, verified
  only structurally — the throw site (`SettingsCodingCliPanel.svelte:199`) and its
  always-rendered read (`:604`) are real, but I did not observe the batch abort.

- `src/server.ts:5843` — `putRepoRoot` answers `{repoRoot, repoRootDisplay}` while `putSettings` is typed `Promise<Settings>` in `ui/src/lib/api.ts:672`; the client type is wider than the wire shape.
- `ui/src/lib/demo/router.ts:350` — the permissive `{ok:true}` mutation tail still hides this whole class of bug for any endpoint that gains an echo-adopting caller; #2240 is the second time it has bitten (see the `/api/sessions/done` and ack-manual-steps notes at `router.test.ts:204` and `:376`).

- `ui/src/lib/components/Settings.svelte:274` and `:290` — `fableAvailable = r.fableAvailable` and `reducedPushMode = r.reducedPushMode` are the unguarded siblings of the three savers this diff hardened; an `undefined` there renders a wrong toggle rather than throwing, and with the demo route added no current responder can produce one, so it is dormant (the plan names it out of scope explicitly).

- `ui/src/lib/demo/state.ts:470-490` — the handler covers `repoRoot`, `authMode` and `anthropicApiKey` shapes the reported bug never exercised; justified as "fix every settings control at once", but it is more than #2240 strictly needed.


**Nits (non-blocking):**

- [bug] ui/src/lib/demo/router.test.ts: the repoRoot test's rationale is factually wrong on two counts — the client function is putSettings (ui/src/lib/api.ts:672), not putRepoRoot (that is the server handler at src/server.ts:5830, which echoes only {repoRoot, repoRootDisplay}); and no consumer reads more than the root off it — SettingsWorkspacePanel.svelte:46 and Onboarding.svelte:51 both use only s.repoRoot. The full-Settings echo is a harmless superset, but the comment pins a requirement that does not exist.
- [bug] ui/src/lib/demo/router.ts:326: an empty patch returns 200 plus the full settings, while the real server falls through to putRepoRoot(undefined) and answers 400 (src/server.ts:5438, 5831-5834). No caller sends an empty body, but router.test.ts:452 now pins the divergence as intended behaviour under a file whose stated principle is to mirror the server field by field.

PR #2245, opened 2026-09-10
feat(usage): switch the default coding CLI when one runs out of weekly capacity

kai-osthoff:
## Why

A weekly window that runs dry just sat there. The popover would show Codex at 77 % with four days
to reset while Claude idled at 21 %, and the only way to spend that headroom was to remember to
change the default CLI in Settings — and to remember to change it back.

## What

The usage popover's hero now offers the switch where the problem is stated:

```
ENGSTES FENSTER
77 %                      Codex · wöchentlich
████████████████░░░░░░░░
Wieder frei in 4d            15.9. 07:43
────────────────────────────────────────
Codex Woche 23 % frei · Claude Code 79 % frei
[  Standard-CLI auf Claude Code umstellen  ]
```

One click switches `defaultAgentProvider` to the counterpart. The origin is persisted, so the
existing 30 s usage tick restores it on its own once the exhausted provider has weekly capacity
again — across a restart too. While it is in effect, the hero shows the revert instead, and the
**Default coding CLI** row in Settings explains why it no longer reads what the operator set.

## Decisions

| Decision | Rationale |
| --- | --- |
| **Weekly window only** | A 5 h window recovers within hours; flipping the global default "until it frees up" would churn the setting for a squeeze that resolves itself. |
| **Releases at ≥ 30 % free, not at the reset timestamp** | Re-evaluating the same predicate self-corrects after a measurement fix, and needs no `resetAt` bookkeeping. In practice weekly usage only falls at the reset anyway. |
| **Threshold hardcoded at 30 %** | No setting, no env — as specified. |
| **Unmeasurable is not free** | A provider reporting no weekly window (api-key mode, a failed scrape, a CLI that never produced a sample) is never switched INTO, and never switched BACK to either. The popover keeps a manual revert as the escape hatch. |
| **Staleness still decides** | It dims presentation, it never reroutes a decision — same rule as `hottestCapacityWindow`. |
| **A manual pick outranks the substitution** | `putDefaultAgentProvider` forgets the origin, so the sweeper never yanks back a deliberate choice. |
| **The reason line names both providers and both percentages** | The hero shows the hottest window across ALL providers — possibly a 5 h window, or the non-default CLI — while the trigger is always the default CLI's weekly one. Without the line the button could read as contradicting the number above it. |

## The rule lives twice, on purpose

Authoritative in `src/provider-failover.ts`, predicted in `ui/src/lib/provider-capacity.ts` so the
popover knows whether to draw the button — the same shape as `claudeUsageHoldLikely` predicting
`shouldHold`. Both run the same case table, and `POST /api/provider-failover` re-derives the offer
and answers **409** when a stale HUD asks for one that no longer holds. The client never names the
target provider; the server derives it.

## Side effect worth naming

`usageHoldApplies` / `tryHoldNewTask` only hold when the resolved provider is Claude. Failing over
during a Claude squeeze therefore stops new tasks being held — without touching the hold threshold.

## Known surface gap

The mobile sheet and the gear menu render `GearMenuUsage`, not this popover, so the button is not
there. Not a dead end: the Settings row shows the active substitution and changing the dropdown
ends it. Left out to keep the diff to the surface the request named.

## Verification

| Gate | Result |
| --- | --- |
| `bun run lint` | pass |
| `tsc --noEmit` | pass |
| `bun run test` (root) | 9631 pass, 0 fail |
| `cd ui && bun run check` | 0 errors, 0 warnings |
| `cd ui && bun run check:i18n` | 2 locales in parity |
| `cd ui && bun run test` | 5079 pass (313 files) |
| `prettier --check .` | clean |
| `fallow audit --base origin/main` | no issues in 21 changed files |
| `check-feature-catalog.sh` / `check-announcement-versions.mjs` | pass (`v1.48.0-usage-provider-failover`) |

New tests: `test/provider-failover.test.ts` (25), `test/server-provider-failover.test.ts` (10),
`ui/src/lib/components/top-bar/UsageFailoverAction.browser.test.ts` (7), plus the mirrored case
table in `ui/src/lib/provider-capacity.test.ts`.

### DIFF
--- a/src/config.ts
+++ b/src/config.ts
@@ -11,6 +11,7 @@ import {
 import { normalizeDefaultEffortSetting, effortBelowHigh } from "./default-effort";
 import { normalizeAuthModeSetting } from "./auth-mode";
 import { normalizeAgentProvider } from "./agent-provider";
+import type { AgentProvider } from "./types";
 import { normalizeTelemetryConsent } from "./telemetry-consent";
 import { normalizeOperatorLanguage } from "./operator-language";
 import { thresholdsFromEnv } from "./maintain-core";
@@ -909,6 +910,11 @@ export const config = {
   // UI-configurable; env seeds a fresh DB.
   defaultAgentProvider:
     normalizeAgentProvider(process.env.SHEPHERD_DEFAULT_AGENT_PROVIDER) ?? "claude",
+  // Capacity failover (src/provider-failover.ts): the provider `defaultAgentProvider` was
+  // switched AWAY from because its weekly window ran out of headroom, and which the 30s usage
+  // tick restores once that headroom is back. null = no failover active. Runtime state, not
+  // configuration — persisted, but deliberately not env-seeded (like previewHost).
+  providerFailoverFrom: null as AgentProvider | null,
   // Global fable availability flag. When false, any spawn requesting --model fable is
   // transparently rerouted to opus[1m] at argv-assembly time without rewriting the
   // stored session model (so cost accounting + fable intent survive for later replay).
--- a/src/index.ts
+++ b/src/index.ts
@@ -195,6 +195,7 @@ import { detectPendingAuthUrl, detectLoginAuthUrl } from "./auth-url";
 import { readTranscriptTail } from "./activity";
 import { verifyApiKey, VERIFY_KEY_LABEL } from "./verify-key";
 import { releaseHeldTasks } from "./held-release";
+import { PROVIDER_FAILOVER_FROM_KEY, releaseProviderFailover } from "./provider-failover";
 import { snapshotSessionUsage } from "./usage-snapshot";
 import { hasCommittedChanges } from "./diff";
 import { HoldReasonService } from "./hold-service";
@@ -385,6 +386,10 @@ if (savedProvider !== null) {
   const v = normalizeAgentProvider(savedProvider);
   if (v !== null) config.defaultAgentProvider = v;
 }
+// An in-flight capacity failover outlives a restart — otherwise the temporary substitution above
+// would silently become the operator's permanent default. "" (the cleared marker, since the
+// settings store has no delete) normalizes to null, i.e. no failover.
+config.providerFailoverFrom = normalizeAgentProvider(store.getSetting(PROVIDER_FAILOVER_FROM_KEY));
 // a UI-set fableAvailable flag (persisted) overrides the env seed; absent or unrecognised → keep default.
 const savedFa = store.getSetting("fableAvailable");
 if (savedFa !== null) {
@@ -2640,6 +2645,10 @@ deferredStarts.push(() => {
     timerTask("usage", async () => {
       await accountIndex.refresh(Date.now());
       events.emit("usage:limits", usageLimits.limits(Date.now()));
+      // Restore the operator's default CLI once the provider a capacity failover switched away
+      // from has weekly headroom again. Synchronous and self-guarding (no-op unless a failover
+      // is active), so it needs no re-entrancy flag of its own.
+      releaseProviderFailover({ store, usageLimits }, Date.now());
       if (releasingHeld) return; // prior release still draining — skip this tick's release
       releasingHeld = true;
       try {
--- a/src/provider-failover.ts
+++ b/src/provider-failover.ts
@@ -1,0 +1,167 @@
+/**
+ * Capacity failover between the two coding CLIs.
+ *
+ * When the WEEKLY window of the operator's default CLI drops below
+ * PROVIDER_FAILOVER_MIN_FREE_PCT remaining and the other CLI measurably has room, the usage
+ * popover offers a one-click switch of `config.defaultAgentProvider` to that counterpart. The
+ * origin is remembered so the 30s usage tick can switch back on its own once the exhausted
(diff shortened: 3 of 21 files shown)


kai-osthoff:
# Review

## Findings

- `src/provider-failover.ts:61`: The server treats an absent confirmed Claude weekly sample as permission to fall back to the locally computed `limits.week`, while the UI's mirror uses `providerDisplayCapacityRows()` and therefore has no Claude weekly window when `observed.week` is null. `UsageLimitsService` initializes and always returns `observed` with `week: null` before a provider sample (`src/usage-limits.ts:467`, `src/usage-limits.ts:653-677`); the display path deliberately renders only confirmed windows whenever that object exists (`ui/src/lib/components/usage-gauges.ts:32-52`, `ui/src/lib/components/usage-gauges.ts:398-407`). Thus, with Codex as the exhausted default and a computed-but-unconfirmed Claude week, the UI correctly renders no offer (`ui/src/lib/provider-capacity.ts:60-86`), but the POST route derives an offer and switches the default anyway (`src/provider-failover.ts:61-65`, `src/server.ts:5136-5144`). This violates the no-measurement rule and the route's stated stale-HUD validation contract. Make the server and UI use the same confirmed-week rule (and add this null-observation case to both seam tests).

## Verification notes

Confirmed the new endpoint is registered in the router (`src/server.ts:8182`) and its status is included in the settings response (`src/server.ts:5428-5431`). Confirmed the active origin is hydrated before runtime use (`src/index.ts:384-392`) and released from the 30-second usage tick (`src/index.ts:2643-2652`). The two maintained message catalogs are `ui/messages/en.json` and `ui/messages/de.json`; the added failover keys occur in both at `ui/messages/en.json:3948-3955` and `ui/messages/de.json:3948-3955`.

kai-osthoff:
**Blocking point 1 — fixed in 13cea12c.** You were right, and it was worse than a divergence in an edge case.

`limits()` **always** emits the `observed` contract, so before the first `/usage` scrape `observed.week` is `null` while the JSONL-computed `limits.week` holds a number. The server therefore called Claude *measured* on a fresh instance while the popover rendered it as "no observation" — and the auto-release path could have restored a default off an estimate the UI deliberately treats as unknown.

Auditing against `claudeDisplayGauges` turned up **three** points of drift, not one:

| | before | after |
| --- | --- | --- |
| contract present, `week: null` | fell through to the computed window | unmeasured |
| top-level contract with a null week | hopped to the per-provider observation | the top-level contract wins whole |
| no contract at all | `limits.week ?? claudeSnapshot.week` | `limits.week` only — the same source `gaugeList` reads |

No UI production change: that side already had the rule right.

**On the test that was passing for the wrong reason.** The route harness omitted `observed` entirely, so its "no weekly window" case exercised the no-contract fallback rather than the rule. It now carries the real payload shape — a confirmed contract sitting over a deliberately *tempting* contradicting estimate (79 % free) — so reading the wrong source engages instead of refusing.

I checked the four new cases actually discriminate rather than passing by construction: against the old rule **4 fail** (three `weeklyFreePct` cases plus the route's 409), against this one all 38 pass.

Re-verified: `lint`, `tsc`, root 9635 pass / 0 fail, `ui check` 0 errors, `check:i18n` parity, ui 5080 pass, `prettier --check` clean, `fallow audit` no issues in 21 changed files.

kai-osthoff (commented):
## Review outcome

No findings. The revised server and UI both treat a present Claude observation contract with a null weekly sample as unmeasured, so the previously reported stale-HUD divergence is resolved.

## Verification

- Re-review confirmation: `src/provider-failover.ts:57-77` selects the top-level observation contract as a whole and returns null for its null weekly value; the only no-contract fallback is the top-level computed week. `ui/src/lib/components/usage-gauges.ts:32-52` applies the same source-selection rule, and `ui/src/lib/provider-capacity.ts:60-65` derives the UI offer from those displayed rows. This addresses the prior finding.
- The server only engages after recalculating capacity and counterpart readiness, returning 409 when no current offer exists (`src/server.ts:5136-5144`); the UI invokes that endpoint and resynchronizes settings after refusal (`ui/src/lib/components/TopBar.svelte:462-474`, `ui/src/lib/api.ts:2607-2616`).
- The automatic restoration checks only the remembered origin's measured weekly free percentage (`src/provider-failover.ts:124-128`, `src/provider-failover.ts:169-179`) and is wired into the 30-second usage tick (`src/index.ts:2636-2643`). A manual default-provider update clears the origin marker (`src/server.ts:5691-5699`).
- The settings response supplies the status consumed by both the popover load and the Settings explanation (`src/server.ts:5428-5432`, `ui/src/lib/components/TopBar.svelte:452-460`, `ui/src/lib/components/settings/SettingsCodingCliPanel.svelte:92-100`).
- All feature message keys are present in both maintained catalogs (`ui/messages/en.json:3948-3955`, `ui/messages/de.json:3948-3955`), and the announcement references the existing catalog keys with the requested version (`ui/src/lib/feature-announcements/entries/v1.48.0-usage-provider-failover.ts:3-8`).

Tests and builds were not run, per review instructions.

kai-osthoff:
✅🎊 PR's READY 🏖️💅 @scoop — merge button's lonely 😢👆🔘 anytime this century ⏳🔥

PR #2248, opened 2026-09-10
fix(codex): constrain role fallback output with schemas

### LINKED ISSUE #2133: Codex roles: add --output-schema so the -o fallback is guaranteed-parseable
Constrain what a Codex role's final message **is**, without touching how it gets **out**.

## Problem

`src/codex-last-message.ts` exists because Codex sometimes treats "write the result to `<file>`" as
"answer with the result": it produces a complete, correct verdict as its final chat message and never
calls a write tool (observed live — TASK-737's recap, commit `6b24a73a`). The role service only reads
the result file, so a correct verdict is silently discarded and the row fails with `no-result`.

Today's mitigation is `codex exec -o <FILE>` (`src/codex-role-argv.ts:31-45`): the CLI itself writes
the agent's final message to a file, and `readRoleResultText` falls back to it when the result file is
absent. That **catches** the failure, but what lands in the file is whatever prose the model emitted —
which then fails the caller's own parse/validation, and fails closed.

## What changed upstream

`codex exec` (verified on codex-cli **0.150.1**) accepts:

```
--output-schema <FILE>    Path to a JSON Schema file describing the model's final response shape
```

That constrains the *shape* of the final response. Combined with the existing `-o`, the fallback
stops being "hopefully parseable prose" and becomes schema-conforming JSON.

## Scope

- Extend `codexRoleArgv` (`src/codex-role-argv.ts:31-45`) with an optional `--output-schema` pointing
  at a per-role schema file, threaded from `buildTransientAgentArgv`
  (`src/transient-agent-argv.ts:222-244`) like `captureLastMessage` already is.
- Emit it only for roles that actually consume a structured verdict — the same opt-in shape as the
  existing `-o` capture, so non-consumers (`namer`, `verify-key`, `doc`) gain no new argv surface.
- The role prompts already state their result contract; the schema should mirror it, not replace it.

## Explicitly NOT in scope

**Keep `-o`.** `--output-schema` governs shape, never destination. `-o <FILE>` is the only reason a
role's final message reaches a file at all: roles are launched into a herdr pane with their argv
wrapped as `env … <argv>` — no shell, so no redirection and no pipe (`src/herdr.ts:604-630`) — and the
read path polls for a file in the spawn's cwd. Dropping `-o` would leave the read path with nothing to
read. The per-spawn unguessable filename for untrusted-checkout reviewers
(`src/codex-last-message.ts:18-25`) stays load-bearing exactly as long as `-o` does.

## Acceptance

- [ ] A Codex role that answers "in chat" still yields a verdict its caller's validator accepts.
- [ ] Roles that do not consume a structured verdict emit no `--output-schema`.
- [ ] Existing `-o` behaviour and the reviewer filename derivation are unchanged.

Background and the flag survey: `docs/research/codex-sdk-evaluation.md` §6 step 2 (PR #2132).

kai-osthoff:
## Summary

Codex helper roles can return a correct verdict in chat without writing their result file. Add an explicit per-role `--output-schema` so the existing `-o` fallback contains JSON matching the consumer's contract. Keep the capture filenames, primary-result/sidecar precedence, prompts, and validators unchanged.

Nine trusted, installation-relative schemas cover the structured consumers, including both critic and merge-suggestion variants and all twelve recap visual types. Non-consumers and Claude receive no new flags.

Closes #2133

## Validation

- `bun run test` — 9,619 passed, 16 skipped, 0 failed (433 files).
- `bun run lint` and `bun run typecheck`.
- Schema acceptance/rejection, real chat-only consumer paths, and argv-clamp regression tests.
- Codex CLI 0.154.0 live smoke for all nine schemas: each returned the expected JSON through only its existing `-o` file; all twelve recap visual types survived.
- Self-review and independent review completed with no outstanding findings.
- Full pre-push gate passed, including UI/extension checks, tests/builds, and the delta Fallow audit.

## Checklist

- [x] Conventional-commit PR title; branch is linear from `origin/main`.
- [x] Relevant local lint, typecheck, and test commands pass.
- [x] No user-facing UI/text or repository instruction changes. [no-feature-entry]
- [x] Existing public configuration and file-result contracts remain unchanged.

### DIFF
--- a/bun.lock
+++ b/bun.lock
@@ -16,6 +16,7 @@
         "@types/bun": "latest",
         "@types/node": "^26.4.1",
         "@types/web-push": "^3.6.4",
+        "ajv": "^8",
         "eslint": "^10.10.0",
         "eslint-plugin-svelte": "^3.23.0",
         "github-slugger": "^2.0.0",
@@ -169,7 +170,7 @@
 
     "agent-base": ["agent-base@7.1.4", "", {}, "sha512-MnA+YT8fwfJPgBx3m60MNqakm30XOkyIoH1y6huTQvC0PwZG7ki8NacLBcrPbNoo8vEZy7Jpuk7+jMO+CUovTQ=="],
 
-    "ajv": ["ajv@6.15.0", "", { "dependencies": { "fast-deep-equal": "^3.1.1", "fast-json-stable-stringify": "^2.0.0", "json-schema-traverse": "^0.4.1", "uri-js": "^4.2.2" } }, "sha512-fgFx7Hfoq60ytK2c7DhnF8jIvzYgOMxfugjLOSMHjLIPgenqa7S7oaagATUq99mV6IYvN2tRmC0wnTYX6iPbMw=="],
+    "ajv": ["ajv@8.20.0", "", { "dependencies": { "fast-deep-equal": "^3.1.3", "fast-uri": "^3.0.1", "json-schema-traverse": "^1.0.0", "require-from-string": "^2.0.2" } }, "sha512-Thbli+OlOj+iMPYFBVBfJ3OmCAnaSyNn4M1vz9T6Gka5Jt9ba/HIR56joy65tY6kx/FCF5VXNB819Y7/GUrBGA=="],
 
     "ansi-regex": ["ansi-regex@6.2.2", "", {}, "sha512-Bq3SmSpyFHaWjPk8If9yc6svM8c56dB5BAtW4Qbw5jHTwwXXcTLoRMkpDJp6VL0XzlWaCHTXrkFURMYmD0sLqg=="],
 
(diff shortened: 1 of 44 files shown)


kai-osthoff (commented):
No blocking findings.

Verified implementation paths:

- The shared Codex branch forwards the optional schema path after retaining the independently selected `-o` filename; the Claude branch is outside that conditional and therefore receives neither Codex option. See `src/transient-agent-argv.ts:239` and `src/transient-agent-argv.ts:255`.
- `codexRoleArgv` emits `--output-schema` as a separate argv pair before the final positional prompt, while preserving the existing `-o` branch. See `src/codex-role-argv.ts:111` and `src/codex-role-argv.ts:143`.
- The trusted, module-relative mapping names nine local schema files. Its consumers cover every current `captureLastMessage: true` call site: autopilot, recap, plan gate, both critic paths, distiller, optimizer, and both merge modes. See `src/codex-role-output-schema.ts:3`, `src/autopilot-llm.ts:91`, `src/recap.ts:305`, `src/plan-gate.ts:486`, `src/review.ts:975`, `src/standalone-critic.ts:541`, `src/distiller.ts:284`, `src/optimizer.ts:236`, and `src/merge-suggest.ts:248`.
- The existing fallback remains result-file-first and keeps the per-spawn reviewer filename derivation unchanged. See `src/codex-last-message.ts:51` and `src/codex-last-message.ts:67`.

Limitation: I did not execute the CLI or test suite, per the review-only instruction; runtime support for the external Codex CLI flag was not independently verified here.

PR #2259, opened 2026-09-11
feat(ui): show session model only when the rail shows a mix

scoop:
## Why

A session row printed its model unconditionally, so a herd running one model repeated the same words down the whole rail — a visual distraction that bought nothing. The label is only worth the line when the rows actually differ.

## What

The model segment renders only when the **visible list** contains more than one model label.

- Uniform rail → `TASK-01 · High`, or just `TASK-01` when effort is unknown too (the ordinary Claude case).
- A Codex run, a different Claude model, or a comparison experiment on display → every row labels itself again.

## Decisions

- **Decided once over the whole visible list** (`shown` — post-filter, *before* grouping), so every group agrees. Per-group detection would show the same model in one group and hide it in the next, and a row would gain or lose its label just by moving between lifecycle stages.
- **Sessions with no known model don't count.** Their label is only the localized "default"; treating that as a second model would call nearly every list mixed and leave the label permanently on.
- **"Same model" means the same rendered label.** A configured floating alias (`opus`, never run) beside an observed `Opus 5` counts as a mix — the client cannot know which concrete model an alias resolved to, so folding them together would be a guess.
- **Nothing becomes unknowable:** the row's hover tip still names the model and its provenance whether or not the segment shows.

## Shape

`sessionEnvironment()` gains `modelKnown` plus a pure `modelsMixed()`; `Herd` derives the flag once and threads it through `HerdRowCtx` → `HerdGroup` → `UnitRow`, whose `showModel` prop **defaults to true** (a row rendered outside a list has no peers to compare against).

`SessionStatusBar` is deliberately untouched — single-session detail, where the model is the point rather than clutter. Effort, the provider label and the experiment-group header chips are unchanged, and no setting is added; the rule is automatic.

While here: `sessionEnvironment`'s model-label ternary chain moved into a small `modelSegment()` helper — the added field pushed the function over the repo's cognitive-complexity bar.

## Tests

- `session-env.test.ts` — eight cases pinning both rules (uniform / differing / known+unknown / all-unknown / empty / live-signal-decides / same model via different routes / alias-vs-resolved).
- `Herd.browser.test.ts` — rail-level wiring: uniform hides with **no dangling separator**, a second model brings labels back on every row, an unknown doesn't make a mix, the decision spans groups, and the effort segment survives suppression.

`cd ui && bun run check` clean, 5075 UI tests pass, pre-push gate green.

### DIFF
--- a/ui/messages/de.json
+++ b/ui/messages/de.json
@@ -3903,6 +3903,8 @@
   "topbar_usage_binding_title": "Engstes Fenster",
   "feat_usage_popover_compact_title": "Auslastungs-Panel zeigt zuerst, was wirklich klemmt",
   "feat_usage_popover_compact_body": "Das Auslastungs-Popover holt jetzt das Fenster nach oben, das seinem Limit am nächsten ist — über Claude Code und Codex hinweg — und setzt es groß. Damit steht das Limit, das über den nächsten Agenten entscheidet, ganz oben statt ganz unten. Die Reset-Zeit jedes Fensters ist als Countdown in die Balkenzeile gewandert, der Datenstand wird einmal oben gestempelt; das Panel wird dadurch rund ein Drittel kürzer. Das Mobile-Sheet nutzt jetzt dasselbe Layout.",
+  "feat_model_label_on_mix_title": "Sitzungszeilen nennen das Modell nur, wenn es etwas aussagt",
+  "feat_model_label_on_mix_body": "Bisher stand auf jeder Sitzungskarte das Modell — bei einer Herde mit nur einem Modell also immer dasselbe Wort, Zeile für Zeile. Das Modell erscheint jetzt nur noch, wenn sich die angezeigten Sitzungen tatsächlich unterscheiden: Sobald ein Codex-Lauf neben den Claude-Läufen steht oder ein Vergleichsexperiment läuft, beschriftet sich wieder jede Zeile. Sitzungen ohne bekanntes Modell zählen dabei nicht als zweites Modell. Beim Überfahren einer Zeile steht weiterhin, welches Modell lief und woher die Angabe stammt — mit oder ohne Label.",
   "topbar_usage_free_in": "Wieder frei in {rel}",
   "topbar_codex_tokens_checked_age": "Token-Datenstand vor {age}",
   "topbar_codex_tokens_checked_now": "Token-Datenstand von gerade eben",
--- a/ui/messages/en.json
+++ b/ui/messages/en.json
@@ -3903,6 +3903,8 @@
   "topbar_usage_binding_title": "Tightest window",
   "feat_usage_popover_compact_title": "Usage panel leads with the window that is actually blocking",
   "feat_usage_popover_compact_body": "The usage popover now promotes whichever window sits nearest its cap — across Claude Code and Codex alike — to a headline reading at the top, so the limit that decides whether you can start another agent is the first thing you see instead of the last. Each window's reset moved into its bar row as a countdown and the provider-confirmation age is stamped once in the header, which takes about a third off the panel's height. The mobile sheet now uses the same layout.",
+  "feat_model_label_on_mix_title": "Session rows name the model only when it tells you something",
+  "feat_model_label_on_mix_body": "A session row used to print its model on every card, so a herd running one model repeated the same words down the whole rail. The model now appears only when the sessions on display actually differ — start a Codex run beside your Claude ones, or a comparison experiment, and every row labels itself again. Sessions whose model was never observed don't count as a second model. Hovering a row still names the model and where that came from, whether or not the label is showing.",
   "topbar_usage_free_in": "Free again in {rel}",
   "topbar_codex_tokens_checked_age": "token data from {age} ago",
   "topbar_codex_tokens_checked_now": "token data from just now",
--- a/ui/src/lib/components/Herd.browser.test.ts
+++ b/ui/src/lib/components/Herd.browser.test.ts
@@ -1108,3 +1108,87 @@ describe("Herd jump flash", () => {
     await expect.poll(() => flashed(screen.container)).toEqual([]);
   });
 });
+
+// The model segment earns its place in a row only when the rail is showing more than one model.
+// Where the unit tests pin the decision, these pin the WIRING: the flag is derived over the full
+// visible set and reaches every row, and suppressing it leaves no dangling separator behind.
+describe("Herd model label", () => {
+  // The task-id button contributes its own layout whitespace, so collapse runs before comparing —
+  // these assert the SEGMENTS, not the row's internal spacing.
+  const meta = (container: HTMLElement) =>
+    [...container.querySelectorAll(".meta-text")].map((el) =>
+      el.textContent!.replace(/\s+/g, " ").trim(),
+    );
+
+  it("one model across the rail: no row prints it, and no row is left with a trailing separator", async () => {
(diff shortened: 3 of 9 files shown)


scoop (commented):
# Review — show the session model only when the rail shows a mix

Scope reviewed: `git diff 8aea0d13e406a619f8b6fd6db3646e88396f51a6...HEAD` (one commit, 9 files).

## What the change does, verified

- `sessionEnvironment()` gains `modelKnown` (`ui/src/lib/session-env.ts:104`) and a new pure
  `modelsMixed(sessions, activity)` (`ui/src/lib/session-env.ts:137-149`) that resolves each
  session's environment, skips the unknown ones, and returns true on the second distinct label
  (early exit at `:146`).
- The extraction of `modelSegment()` (`ui/src/lib/session-env.ts:38-42`) is behaviour-identical to
  the ternary chain it replaces (old `ui/src/lib/session-env.ts:58-62` in the pre-image, now
  `:77`): observed → `runtimeModelLabel`, configured → `modelLabel`, else `m.newtask_model_default()`.
  Nothing else in the returned object changed, so `SessionStatusBar`'s output is unchanged — it
  reads `providerLabel(...) + environment.segments` at `ui/src/lib/components/SessionStatusBar.svelte:45`
  and never touches the new field. Success criterion 5 holds.
- `Herd.svelte` derives the flag once over the post-filter set (`ui/src/lib/components/Herd.svelte:339`,
  with `shown` defined at `:287-291`) and puts it on the shared bundle (`:440`).
- Every rendered `UnitRow` in the rail goes through `HerdGroup` — I grepped the tree, and
  `ui/src/lib/components/herd/HerdGroup.svelte:126` is the ONLY `<UnitRow>` render outside test
  files. `HerdRowCtx.showModel` is required (`HerdGroup.svelte:40`) and forwarded at
  `HerdGroup.svelte:153`; the other two `HerdRowCtx` consumers pass the same bundle straight through
  (`herd/HerdEpicGroups.svelte:46`, `herd/HerdExperimentGroups.svelte:65`), and no test constructs a
  `HerdRowCtx` literal, so making the field required breaks no caller.
- `UnitRow` keeps a `showModel = true` default (`ui/src/lib/components/UnitRow.svelte:87`), so the
  standalone renders in `UnitRow.browser.test.ts:151+`, `session-env-parity.browser.test.ts:90+` and
  `mobile-list-overflow.browser.test.ts:228` keep their current text.
- The dangling-separator case is genuinely handled: `environmentText` is `[]` when the model is
  suppressed and `environment.effort` is null (`UnitRow.svelte:203-205`), and the separator hangs off
  `environmentSuffix` (`:206`), consumed at `:953`. I checked the two collapse cases in the resolver
  (`session-env.ts:110`): effort unknown → `segments = [model]` → suppressed → empty suffix; effort
  known → `[effort]`. `effortLabel` never returns an empty string (`ui/src/lib/effort-guidance.ts:41-58`,
  default branch echoes the raw value), so the effort-only branch can't produce `TASK-01 ·` either.
- Layout is unchanged for the visible case: previously the template emitted the text run
  `" · " + environmentText`; now it emits the identical run as one expression. `.meta-text` is a flex
  item of `.meta` (`UnitRow.svelte:1553-1573`), the inner content is ordinary inline flow, so the
  leading space is treated exactly as before.
- i18n parity confirmed by opening both catalogs: `feat_model_label_on_mix_title` /
  `_body` exist at `ui/messages/en.json:3906-3907` and `ui/messages/de.json:3906-3907`, inserted at
  the same position in both. `ui/messages/` holds only `en.json` and `de.json`, so there is no third
  catalog to update.
- Feature-catalog contract met: `ui/src/lib/feature-announcements/entries/v1.48.0-model-label-on-mix.ts`
  is auto-registered by the eager glob at `ui/src/lib/feature-announcements.ts:47-49` (no index to
  edit), its `sinceVersion` and filename agree, and `1.48.0` is the next unreleased version —
  `package.json:3` still reads `1.47.0` and the other `v1.48.0-*` entries share it, so
  `scripts/check-announcement-versions.mjs` is satisfied.
- No hardcoded user-facing string is introduced (`.claude/rules/i18n.md`): the only literal added to
  the template path is the `·` glue, which is the existing pattern in this row.

No correctness or security defect found. Everything below is non-blocking.

- `ui/src/lib/session-env.ts:104` computes `modelKnown` with `??` (`!!(observedModel ?? configuredModel)`) while its sibling `modelSegment` (`:38-40`) decides the same thing by truthiness — an empty-string `runtimeModel` would render the CONFIGURED label while reporting `modelKnown: false`, so the row would print a real model that `modelsMixed` refuses to count; unreachable today because every producer drops falsy ids before it reaches the client (`src/poller.ts:1440`, `src/codex-activity.ts:197`, `src/activity-signal.ts:88`), but `||` would make the two agree unconditionally.

- `ui/src/lib/session-env.ts:38-42` extracts `modelSegment()` out of `sessionEnvironment`, a refactor of code the feature did not require touching; the comment justifies it by the cognitive-complexity gate (fallow defaults 20/15 per `.fallowrc.jsonc:193`), which is plausible but not something I could confirm without running the gate — behaviour is identical either way, so keeping it is the author's call.

## Notes on the two design edges

Neither is a defect, both are worth the reader knowing:

- The mix is decided over `shown`, which is the post-filter set BEFORE grouping and before any
  collapse state. Rows inside a collapsed group are still counted even though they are not rendered:
  `HerdGroup.svelte:116` skips the rows of a collapsed group entirely, `HerdEpicGroups.svelte:44-48`
  the same for a collapsed epic, and on the phone layout the lifecycle groups are an accordion with
  exactly one open (`Herd.svelte:743`, default key at `:391`). So a single Codex row folded away in
  Merged turns the label back on for every visible Claude row. Raised as a nit — the plan chose
  whole-list scope deliberately, and the alternative makes labels flip as the operator folds
  sections.
- A live comparison experiment always mixes models and therefore re-labels the whole rail. Called
  out and accepted in the plan; it is a true statement about what is on display.

## Stated limitations

- I did not run `bun run check`, the test suite, or the fallow audit — CI owns those, and the
  instructions for this review are read-only.


**Nits (non-blocking):**

- [scope] ui/src/lib/components/Herd.svelte:339: the mix is derived over `shown`, which includes rows inside COLLAPSED groups — HerdGroup.svelte:116 and HerdEpicGroups.svelte:44 skip rendering them, and the phone layout keeps exactly one lifecycle group open (Herd.svelte:743, default key at :391). One Codex row folded away in Merged therefore re-labels every visible Claude row, which is the noise the task set out to remove. Defensible as-is (labels would otherwise flip as sections are folded), but worth a deliberate decision rather than an accident of where the derivation sits.
- [compliance] ui/messages/en.json:3907: the announcement body says sessions whose model was never OBSERVED do not count as a second model, but a configured-yet-never-run model does count (modelKnown at ui/src/lib/session-env.ts:104 is observed OR configured). The German line at ui/messages/de.json:3907 gets this right with `ohne bekanntes Modell`; consider matching it, e.g. sessions with no known model.
- [bug] ui/src/lib/components/UnitRow.svelte:953: with the model suppressed, `.meta-text` shrinks to just the TaskIdButton, so the only way to hover the environment tooltip is over that button — which still carries its own native `title` (TaskIdButton.svelte:73) that statusTip strips only from its own node (ui/src/lib/actions/statusTip.svelte.ts:193). Both tooltips now show on the one remaining hover target; the model stays discoverable (and `aria-description` is unaffected), so this is cosmetic.

PR #2260, opened 2026-09-11
chore(deps-dev): migrate ui to vitest 5

scoop:
Replaces #2256, #2255 and #2254 with a single coherent upgrade.

## Why one PR

Vitest 5 pins its own packages as **exact peers** of each other — `vitest@5.0.0`
peers `@vitest/browser-playwright: 5.0.0`, `@vitest/browser@5.0.0` peers
`vitest: 5.0.0`. All three are direct devDeps in `ui/`, so the family is only
ever installable as a set. Dependabot opened one PR per package and none of them
could land:

| PR | | |
|---|---|---|
| #2256 | vitest 5 alone | red — 1304 svelte-check errors against browser@4 |
| #2254 | browser-playwright 5 alone | red — `vitest/node` has no `BrowserConnectionError` in v4 |
| #2255 | @vitest/browser 5 alone | **green but inert** — browser-playwright@4 drags a nested `@vitest/browser@4` which is what actually runs |

#2255 is the interesting one: it would have merged clean and delivered none of
the upgrade.

## Breaking changes handled

**`test.browser.api` → `test.api`.** Vitest 5 removed the nested option. The
`#817` `strictPort: false` workaround is preserved (both browser projects), and
the guard test in `vite-config.test.ts` now asserts the new path.

**Locators are strict by default.** Per the migration guide, browser locators
now require an exact, case-sensitive match — `getByText("limit")` no longer
matches a cell reading `limit required`. This suite was written against v4's
substring semantics, so it's reverted with the documented knob,
`browser.locators.exact = false`, rather than rewriting ~60 assertions.

**`toHaveTextContent` now does strict equality**; partial matching moved to the
new `toMatchTextContent`. Renamed **all 44 call sites**, not only the ones that
failed — leaving the rest would silently tighten assertions that happen to match
exactly today, which isn't something a dependency bump should do unreviewed.

**`@vitest/browser/context` no longer resolves inside browser mode.** Exactly one
spec still used that legacy path; `DeliveryLens` now imports `vitest/browser`
like every other spec in the repo.

## Two incidental fixes

`PrReviewRequestPopover`'s narrow-viewport test read `getBoundingClientRect()`
once while the popover's async `ResizeObserver` was still settling — a
pre-existing race that v5's timing exposed. It now polls until the popover
settles inside the viewport; verified it converges (not a component bug) and
re-ran it 3× clean.

`ui/.gitignore` listed `.vitest-attachments/`, the v4 directory name. Vitest 5
writes failure screenshots to `.vitest/`, which would otherwise have been
committed.

## Preventing the recurrence

Adds a `vitest` Dependabot group covering **majors**, listed first so it wins the
match. The two existing blanket groups are `["minor", "patch"]` only, which is
why this major arrived ungrouped in the first place — any exact-peer-pinned
family will keep splitting until it has its own group.

## Verification

| | |
|---|---|
| `ui`: `bun run check` | 0 errors, 6225 files |
| `ui`: `bun run check:i18n` | 2 locales in parity (3947 keys) |
| `ui`: `bun run test` | **312/312 files**, exit 0 |
| root: `bun run lint` | clean |
| root: `bunx tsc --noEmit` | clean |
| root: `bun run test` | 9621 pass, 0 fail |
| `fallow audit --base origin/main` | no issues in 13 changed files |
| branch hygiene | linear off origin/main |

No user-facing strings changed, so the i18n catalogs are untouched.

## Note for review

`browser.locators.exact = false` **preserves** current assertion strength rather
than weakening it — it restores the semantics every one of these tests was
written against. Tightening the suite to exact matching is a genuine improvement
and probably worth doing, but as its own piece of work, not smuggled into a
dependency bump.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -28,6 +28,15 @@ updates:
       prefix: "chore(deps)"
       prefix-development: "chore(deps-dev)"
     groups:
+      # vitest pins its own packages as *exact* peers of each other
+      # (`vitest@5.0.0` peers `@vitest/browser-playwright: 5.0.0`, and back), so
+      # the family is only ever installable as a set. Ungrouped, a major landed as
+      # one PR per package — each individually un-mergeable, and each red or (worse)
+      # green-but-inert. This group is listed first so it wins the match, and it
+      # covers majors, which the two blanket groups below deliberately do not.
+      vitest:
+        patterns: ["vitest", "@vitest/*", "vitest-browser-svelte"]
+        update-types: ["major", "minor", "patch"]
       production-dependencies:
         dependency-type: "production"
         update-types: ["minor", "patch"]
--- a/ui/.gitignore
+++ b/ui/.gitignore
@@ -26,5 +26,7 @@ vite.config.ts.timestamp-*
 /src/lib/paraglide
 
 # Vitest browser-test failure artifacts (regenerated on failing runs)
+# vitest 5 writes these under .vitest/attachments (v4 used .vitest-attachments).
 __screenshots__/
+.vitest/
 .vitest-attachments/
--- a/ui/bun.lock
+++ b/ui/bun.lock
@@ -25,15 +25,15 @@
         "@sveltejs/vite-plugin-svelte": "^7.3.0",
         "@tailwindcss/vite": "^4.3.3",
         "@types/node": "^26.5.0",
-        "@vitest/browser": "^4.1.11",
-        "@vitest/browser-playwright": "^4.1.11",
+        "@vitest/browser": "^5.0.0",
+        "@vitest/browser-playwright": "^5.0.0",
         "playwright": "^1.63.0",
         "sharp": "^0.35.4",
         "svelte": "^5.57.0",
         "svelte-check": "^4.7.6",
         "typescript": "^6.0.2",
         "vite": "^8.2.2",
-        "vitest": "^4.1.11",
+        "vitest": "^5.0.0",
         "vitest-browser-svelte": "^3.1.0",
       },
     },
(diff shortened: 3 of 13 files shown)


scoop (commented):
# Review — `chore(deps-dev): migrate ui to vitest 5` (08f3d32e6)

Diff reviewed: `git diff 4da5dc982f8fa549da4d1f90a9c91e4383baca85...HEAD` — 13 files, one commit.

## What the change does, and whether it fits the task

The task was a merge train over #2256 / #2255 / #2254 — three dependabot PRs that each bump one member of
the vitest family to 5.0.0 in `/ui`. The diff replaces the three with a single combined migration plus a
dependabot grouping rule so the situation does not recur.

I verified the premise the author used to justify collapsing the three PRs, rather than taking the commit
message at its word:

- `ui/bun.lock:340` — `@vitest/browser@5.0.0` declares `"peerDependencies": { "vitest": "5.0.0" }` (exact, not a range).
- `ui/bun.lock:342` — `@vitest/browser-playwright@5.0.0` declares `"@vitest/browser": "5.0.0"` as a hard dependency
  and `"vitest": "5.0.0"` as an exact peer.

So bumping any one of the three alone leaves an unsatisfiable peer set; the family really is only installable
as a unit, and merging them one at a time is not possible. Replacing the trio with one atomic PR is the correct
resolution of that train, and it matches the task's own escape hatch ("move any PR we don't actually want into
a hold list"). `ui/bun.lock:736` confirms the fourth family member is not exact-pinned
(`vitest-browser-svelte@3.1.0` peers `vitest: "^4.0.0 || ^5.0.0"`), so it did not need a version change — only
group membership, which it gets at `.github/dependabot.yml:38`.

The dependabot group at `.github/dependabot.yml:37-39` sits inside the `bun` ecosystem entry whose `directories`
are `/`, `/ui`, `/extension`, `/site`, `/docs-site` (`.github/dependabot.yml:15-20`), so it covers the extension
too — which is what you want, given `/extension` took its own solo vitest 5 bump in fa7966425.

## Verification of the migration itself

**Import specifier consistency.** The diff changes exactly one import, in `DeliveryLens.browser.test.ts`, from
`@vitest/browser/context` to `vitest/browser`. That looked like it might be a one-off inconsistency, so I checked
the whole tree: `vitest/browser` is the form used by 117 test files, and after this change there are **zero**
remaining `@vitest/browser/context` imports in any source file. The only surviving occurrences are three code
samples in a dated planning doc (`docs/superpowers/plans/2026-06-04-chrome-overflow-regression-tests.md:413`,
`:461`, `:604`), which is historical record, not living guidance. So this was the last straggler, not a new
divergence.

**Matcher rename.** `toHaveTextContent` → `toMatchTextContent` is applied in 6 files. I grepped the entire
repository: `toHaveTextContent` now has **zero** occurrences anywhere, and `toMatchTextContent` has 44 across
those same 6 files. The conversion is complete — no half-migrated file.

I also checked that the rename is being used for its substring semantics and not accidentally where an exact
match was intended. E.g. `ui/src/lib/components/CommandBar.browser.test.ts:150-151` asserts both
`"needs-input"` and `m.status_blocked()` against the *same* option element, which is only meaningful under
substring matching; the adjacent exactness checks are still done with plain
`expect(first.element().textContent).not.toContain(...)` (`CommandBar.browser.test.ts:196`), which is
matcher-independent. Consistent.

**Config move.** `api: { strictPort: false }` is added at project level for **both** browser projects
(`ui/vite.config.ts:117` for `browser`, `:140` for `browser-touch`) and removed from inside `browser` in both —
the guard is not present on one sibling and missing on the other. Likewise `locators: { exact: false }` is
applied to both (`:127`, `:146`). The accompanying test was updated to the new path
(`ui/src/lib/vite-config.test.ts:8`, `:19`).

**No stale references left behind.** I grepped for every identifier the move could strand:
`strictPort` / `63315` / `locators` appear only in `ui/vite.config.ts`, `ui/src/lib/vite-config.test.ts`, and
`extension/vite.config.ts:20` (an unrelated dev-server port). No workflow, script, or doc referenced
`browser.api`. Similarly, `.vitest-attachments` / `.vitest/` / `__screenshots__` appear only in `ui/.gitignore:29-32`
— nothing in `.github/` uploads or globs those paths, so the v4→v5 artifact-directory move cannot silently
break a CI artifact step.

**Lockfile.** The `ui/bun.lock` diff contains no unrelated *direct* dependency changes — only the three vitest
entries and their transitive fallout (`@vitest/ui` newly pulled in as a hard dependency of `@vitest/browser@5.0.0`,
per `ui/bun.lock:340`; `react`/`react-dom`/`scheduler` and the nested `vitest/vite` rolldown tree dropped).

- `ui/node_modules` does not exist in this review worktree and no package cache was reachable, so I could not
  open vitest 5's own sources or types. That means three of the diff's load-bearing claims are asserted by the
  author and **unverified by me**, not confirmed wrong: (a) that `strictPort` moved from `browser.api` to the
  project's `test.api`; (b) that `browser.locators` accepts an `exact` option in v5; (c) that
  `toMatchTextContent` is v5's substring-matching replacement for `toHaveTextContent`. All three are the kind of
  thing CI will surface loudly rather than silently — an unrecognised `locators.exact` would leave the suite on
  v5's stricter matching and turn assertions red — with one exception worth naming, which I raise as a nit below:
  `ui/src/lib/vite-config.test.ts` asserts the *shape of the config object*, so it passes whether or not vitest
  actually honours the `test.api` key. It is not a guard against (a) being wrong; only a real concurrent run is.
- Whether the three upstream PRs were actually closed, and whether the approval gate in the task was honoured,
  is not observable from the diff.

## Notes (non-blocking judgement calls)

- `ui/vite.config.ts:127,146` — `locators: { exact: false }` deliberately pins the suite to v4 matching semantics
  across the board rather than tightening ~100 test files. Reasonable for a dependency bump and clearly commented,
  but it is standing technical debt: new tests written after this land will silently inherit loose matching.
  A follow-up issue to burn it down would be worth opening.
- `.github/dependabot.yml:39` — the group opts into `"major"`, which the two blanket groups below it deliberately
  exclude. That is the point of the group (a solo major is what broke here), and it is explained in the comment,
  but it does mean future vitest majors will now arrive as auto-opened PRs across all five directories.
- `ui/.gitignore:32` — `.vitest-attachments/` is the v4 path and is now dead for fresh checkouts. Harmless to keep
  (it still covers leftover artifacts on existing dev machines) and the comment says as much.

- `ui/package.json:21` — `test:browser` runs `vitest run --project browser`, so the `browser-touch` project (which owns `BacklogView.browser.test.ts`) is never exercised by that script; only the bare `test` script reaches it.
- `ui/vite.config.ts:98,135` — the `browser`/`browser-touch` split is maintained by hand-listing `BacklogView.browser.test.ts` in one project's `exclude` and the other's `include`; adding a second touch spec requires editing two places with no guard against them drifting apart.


**Nits (non-blocking):**

- [bug] ui/src/lib/vite-config.test.ts: the strictPort guard is asserted only for the project named browser, but the diff adds the same api.strictPort:false to the browser-touch project at ui/vite.config.ts:140. Assert both (e.g. iterate the projects whose test.browser.enabled is true) so a future edit cannot drop it from one sibling unnoticed.
- [scope] ui/src/lib/components/PrReviewRequestPopover.browser.test.ts: the rewrite of the bottom-edge assertion into expect.poll is the only behavioural test change in the diff carrying no explanatory comment, while every other v5 adaptation in this commit is commented. A one-line note saying why the popover now needs to settle before measuring would keep the migration self-documenting.
- [scope] ui/vite.config.ts: locators exact:false pins the whole suite to vitest 4 substring matching indefinitely, so tests written after this lands silently inherit the loose semantics. The trade-off is well explained for the migration itself; consider filing a follow-up to tighten assertions and drop the option.

Issue #2267, opened 2026-09-11
Session-finished push is dropped for any pane the operator was watching


scoop:
`src/push.ts` gates the "session finished" push notification on `status !== "done"`.

herdr's `done` is a *"finished, and nobody has looked yet"* notification state, not a lifecycle
state: **viewing the pane clears it back to `idle`, permanently.** Verified live — `herdr agent
focus` on a `done` pane flips it to `idle` instantly and it stays there (`agent read` and the HUD's
`terminal session control` attach do *not* clear it).

So any session the operator had open when it finished never reports `done`, and its finish push is
silently dropped.

This is the same root cause as the plan-gate / autopilot hang fixed in #<PR>, which added
`TurnEndBackstopService` (`src/turn-end-backstop.ts`) — a settled-idle sweep that re-drives the
consumer that missed its turn-end edge. `push.ts` was deliberately left out of that PR's scope.

Two plausible fixes:
- Route the finish push through the backstop as a third consumer, or
- Accept `idle` as well as `done` at the push site, with whatever dedupe the push path needs so a
  session resting for hours doesn't re-notify.

Worth confirming which, since the push path has its own idempotence story.

scoop:
✅ PR #2270 merged.

PR #2268, opened 2026-09-11
fix(plan-gate): recover turn-end edges herdr never reports as done

scoop:
## The bug

Roughly one session in five hung after plan creation: the plan was written, the agent went quiet,
and the adversarial plan reviewer never started — with autopilot on.

## Root cause (confirmed live, not inferred)

**herdr's `done` is a "finished, and nobody has looked yet" notification state, not a lifecycle
state.** Viewing the pane clears it back to `idle`, permanently.

Verified on the live instance:

| Probe | Result |
| --- | --- |
| `herdr agent focus` on a `done` pane | flips to `idle` **instantly**, stays `idle` after unfocus |
| `herdr agent read` | stays `done` |
| `herdr terminal session control` (the HUD's PTY attach) | stays `done` |
| `herdr agent explain` on a `done` and an `idle` pane | both `state: idle`, rule `live_prompt_box` — so `done` is an overlay, not terminal-derived |

Shepherd keyed "the agent finished a turn" on `done` alone in exactly two places:

- `src/plan-gate.ts` → `shouldConsiderOnSettle()` returns true on `done`; on `idle` only when a
  prior gate already decided `changes_requested`. A **first** draft has no prior gate, so it
  returns false.
- `src/autopilot.ts` → `handle()` calls `onDone()` only when `change.status === "done"`.

`src/poller.ts` → `reconcileAgent()` maps herdr state through `mapState()`: only herdr `done`
becomes Shepherd `done`.

So an operator **watching the pane as the planner finishes** destroys the only trigger, and nothing
re-checks — the session hangs forever. Autopilot is silenced by the same lost edge, which is why
"autopilot is on" did not rescue it. The plan gate concentrates the failure because its directive
makes the agent ask questions via `AskUserQuestion`, so the operator is on that tab exactly when the
plan lands.

Caught in the act: session `26017f67` was sitting in `planPhase=planning` with `.shepherd-plan.md`
written, herdr `agent_status: idle`, zero `plan_gates` rows, zero `reviewer_spawns` rows and no
`[plan-gate]` log line. Its hook trail showed two `AskUserQuestion` rounds.

Most of the codebase already treats the two as equivalent — `recap-core.ts` → `isSettledIdle()`,
`recap.ts` → `sweep()`, and `plan-gate.ts` → `shouldCheckPlanDrift()` all accept `idle || done`.
The first-review trigger and `onDone()` were the outliers.

## The fix

`TurnEndBackstopService` (`src/turn-end-backstop.ts`) — a settled-idle sweep that recovers the
missed edge, borrowing the debounce structure `RecapService` and `BuildQueueReminderService` already
use. It rides the existing 15s sweep interval; no new timer.

**Status semantics, `mapState()` and `shouldConsiderOnSettle()` are untouched.** The fast path still
fires instantly; the backstop only does work when the fast path already failed, so its settle
threshold costs latency *only* in the broken case.

Safe to re-drive because both targets already self-guard: `PlanGateService.consider()` short-circuits
on gate-off, wrong phase, in-flight/mid-spawn, approved, missing plan, unchanged plan hash and an
already-refused spawn; `AutopilotService.onDone()` runs through `eligible()`, which stands down on
archived, planning phase, autopilot off, non-isolated codex, pending MCP OAuth, paused, complete, an
open PR outside full-auto, and an in-flight classify.

Guards, each with precedent in `BuildQueueReminderService`:

- **Observed-active evidence gate** — only backstop a session seen non-resting at least once. This
  is what makes a restart safe: a session already at rest when the process starts is never
  backstopped, so there is no restart storm of classify spawns.
- **Never the first settled tick** — per the once-on-settled-idle house rule.
- **`markDelivered()` on the real `done` edge** — a healthy session never pays a redundant autopilot
  classify.
- **Once per episode, plus a per-session attempt cap** — failed attempts count too, so a
  persistently throwing dispatch can't be re-driven every 15s forever.
- **Re-entrancy guard** on the sweep; the tick is dropped, not queued.

Recoveries log one line each, so the previously-invisible rate of lost turn-end edges is now
measurable.

## Verification

`bun run lint` clean · `bun run test` 9700 pass / 0 fail · `bunx tsc --noEmit` clean ·
`bunx fallow audit --base origin/main --fail-on-issues` exit 0.

14 new unit tests cover the regression and every guard, over an injected clock and store — no herdr,
no spawns, no LLM.

## Notes for the reviewer

- **Known duplication (warn-only).** `fallow` flags a 28-line clone between this sweep and
  `build-queue-reminder.ts`'s — the generic "iterate active sessions, consider each, prune dead
  entries" loop. Left as-is deliberately: extracting it would mean editing working code in the
  sibling service for no behavioural gain. Happy to extract as a follow-up if preferred.
- **`push.ts` has the same bug** and was deliberately scoped out — filed as #2267.
- ⚠️ **Unrelated but worth knowing:** `herdr integration status` reports `claude: outdated (v5 < v9)`.
  The installed v5 hook reports session *identity* only, never state, which is why herdr still falls
  back to terminal classification and can produce `done` at all. herdr's docs say an
  actively-reporting integration is authoritative for `idle`/`working`/`blocked` — **a list with no
  `done` in it.** If v9 reports state, `herdr integration install claude` could take this from a
  1-in-5 failure to 100%. This PR would absorb that rather than hang, but every turn end would then
  pay the settle latency. Verify before upgrading.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -2234,6 +2234,13 @@ events.subscribe((event, data) => {
     // The real turn-end edge landed → stand the backstop down for this resting episode, so a
     // healthy session never pays a redundant autopilot classify 2 minutes later.
     if (status === "done") turnEndBackstop.markDelivered(id);
+    // Arm the backstop's evidence gate + end its resting episode from the poller's 1 Hz transitions,
+    // not just its own coarse 15s sweep sample: `status` is a point-in-time level, so a working
+    // burst between two sweeps is invisible to the sample (the #1617 problem, same shape as
+    // buildQueueReminder.markRan above). Without this a short turn never arms the gate, a stale
+    // `delivered` suppresses the NEXT turn's genuinely lost edge, and a carried-over settle clock
+    // fires the backstop mid-turn against a half-written plan.
+    if (status === "running" || status === "blocked") turnEndBackstop.markActive(id);
     void sessionRouter
       .onStatus(id, status)
       .catch((err) => console.warn("[session-router] onStatus:", err));
--- a/src/index.ts
+++ b/src/index.ts
@@ -190,6 +190,7 @@ import { RecapService, type LandedWorkEvidence } from "./recap";
 import { PostMergeStepsService } from "./post-merge-steps";
 import { DeliveryFactsService } from "./delivery";
 import { BuildQueueReminderService } from "./build-queue-reminder";
+import { TurnEndBackstopService } from "./turn-end-backstop";
 import { jsonlPathFor } from "./usage";
 import { detectPendingAuthUrl, detectLoginAuthUrl } from "./auth-url";
 import { readTranscriptTail } from "./activity";
@@ -1889,6 +1890,17 @@ onSessionGit(({ id, git }) => {
 // and write-free in the steady state, so it can ride this per-poll event.
 const deliveryFacts = new DeliveryFactsService({ store });
 onSessionGit(({ id, git }) => deliveryFacts.onGit(id, git));
+// Turn-end backstop: herdr's `done` is a "finished and unseen" notification state that VIEWING the
+// pane clears back to `idle`, so an operator watching a session as it finishes destroys the only
+// edge the plan gate's first review and autopilot's onDone key on — and nothing re-checks. This
+// recovers that edge from a settled-idle sweep. `autopilot` is declared below and referenced lazily
+// (the closure runs long after boot); both targets are idempotent/self-guarding, and the `delivered`
+// mark from the real `done` edge keeps a healthy session off this path entirely.
+const turnEndBackstop = new TurnEndBackstopService({
+  store,
+  considerPlan: (s) => planGate.consider(s),
+  autopilotDone: (id) => autopilot.onDone(id),
+});
 deferredStarts.push(() => {
   setInterval(() => {
     if (maintenance.active) return;
@@ -1920,6 +1932,10 @@ deferredStarts.push(() => {
     void buildQueueReminder
       .sweep()
       .catch((err) => console.warn("[build-queue] reminder sweep failed:", err));
+    // settled-idle recovery for a turn end herdr never reported as `done`
+    void turnEndBackstop
+      .sweep()
+      .catch((err) => console.warn("[turn-end-backstop] sweep failed:", err));
   }, 15_000);
 });
 // The standalone critic's enumeration runs on its OWN 60s timer, separate from the 15s
@@ -1952,6 +1968,7 @@ events.subscribe((event, data) => {
     recapService.onArchived(id);
     autoMergedRecapEvidence.delete(id);
     buildQueueReminder.forget(id);
+    turnEndBackstop.forget(id);
     docAgent.onArchived(id);
   }
 });
@@ -2214,6 +2231,9 @@ const sessionRouter = new SessionRouter(
 events.subscribe((event, data) => {
   if (event === "session:status") {
     const { id, status } = data as { id: string; status: string };
+    // The real turn-end edge landed → stand the backstop down for this resting episode, so a
+    // healthy session never pays a redundant autopilot classify 2 minutes later.
+    if (status === "done") turnEndBackstop.markDelivered(id);
     void sessionRouter
       .onStatus(id, status)
       .catch((err) => console.warn("[session-router] onStatus:", err));
--- a/src/turn-end-backstop.ts
+++ b/src/turn-end-backstop.ts
@@ -1,0 +1,227 @@
+// Turn-end backstop — recovers the "agent finished a turn" edge when herdr never reports `done`.
+//
+// herdr's `done` is a "finished, and nobody has looked yet" NOTIFICATION state, not a lifecycle
+// state: viewing the pane clears it back to `idle`, permanently (verified live — `herdr agent focus`
+// on a `done` pane flips it to `idle` and it stays there; `agent read` and the HUD's
+// `terminal session control` attach do NOT clear it). Shepherd keys "the agent finished a turn" on
+// `done` in exactly two places:
+//   - plan-gate.ts shouldConsiderOnSettle() — fires the FIRST plan review on `done`; on `idle` only
+//     when a prior gate already said `changes_requested`, which a first draft never has.
+//   - autopilot.ts handle() — calls onDone() only on `done`.
+// So an operator watching the pane as the agent finishes destroys the only trigger, and nothing
+// re-checks: the session hangs forever. The plan gate concentrates the failure because its directive
+// makes the agent ask questions via AskUserQuestion, so the operator is on that tab exactly when the
+// plan lands.
+//
+// This service does NOT change status semantics, mapState(), or shouldConsiderOnSettle(): the fast
+// path stays as-is and still fires instantly. The backstop only does work when the fast path already
+// failed, so its settle threshold costs latency ONLY in the broken case.
+//
+// It borrows RecapService's / BuildQueueReminderService's settled-idle debounce STRUCTURE. Both
+// dispatch targets are already fully self-guarding, which is what makes re-driving them safe:
+//   - PlanGateService.consider() short-circuits on gate-off, phase not `planning`, a review in
+//     flight or mid-spawn, an already-`approved` gate, a missing/empty plan, an unchanged plan hash,
+//     and a spawn already refused for this exact plan text. A redundant call is a free no-op.
+//   - AutopilotService.onDone() runs through eligible(), which stands down on archived, planning
+//     phase, autopilot disabled, non-isolated codex, pending MCP OAuth, paused, complete, an open PR
+//     outside full-auto, and a classify already in flight.
+
+import { isSettledIdle } from "./recap-core";
+import type { SessionStore } from "./store";
+import type { Session, SessionStatus } from "./types";
+
+/** Matches RecapService and BuildQueueReminderService — the house settled-idle dwell. */
+const DEFAULT_IDLE_THRESHOLD_MS = 120_000;
+
+/** Per-session lifetime cap on recovery ATTEMPTS (runaway guard). */
+const DEFAULT_MAX_ATTEMPTS = 3;
+
+/** The two statuses that mean "the agent is at rest": herdr's view-clearable `done` and the `idle`
+ *  it decays into. The pair {@link isSettledIdle} accepts — kept as a named predicate here because
+ *  this service needs a THREE-state view (active / resting-not-settled / resting-settled) that a
+ *  single boolean can't express. */
+function isResting(status: SessionStatus): boolean {
+  return status === "idle" || status === "done";
+}
+
+interface DebounceEntry {
+  /** epoch ms when the current resting episode began; null whenever the session isn't resting. */
+  settledSince: number | null;
+  /** true once this episode's turn end was recovered (reset when the session re-activates). */
+  firedThisEpisode: boolean;
+  /** true once the REAL `done` edge was routed for this episode — set by {@link
+   *  TurnEndBackstopService.markDelivered}. A healthy session must never pay for a redundant
+   *  autopilot classify, so a delivered episode is never backstopped. Cleared on re-activation so
+   *  the next turn's edge re-marks it. */
+  delivered: boolean;
+  /** true once we've observed this session NOT resting. The evidence gate that makes a restart
+   *  safe: a session already at rest when the process starts was never observed active, so it is
+   *  never backstopped and no restart storm of classify spawns can happen. Mirrors
+   *  BuildQueueReminderService's `sawRunning`. Deliberately NOT reset per episode — it is evidence
+   *  about the session, not about the current rest. */
+  sawActive: boolean;
+  /** lifetime recovery ATTEMPTS for this session, successful or not (runaway guard). Counting
+   *  failures too is deliberate: a rejection leaves `firedThisEpisode` unset so a TRANSIENT failure
+   *  retries on the next tick, and without a bounded attempt count a persistently throwing dispatch
+   *  would retry every sweep forever. */
+  attemptCount: number;
+}
+
+interface Deps {
+  store: Pick<SessionStore, "list">;
+  /** Re-drive the plan gate for a planning-phase session. Idempotent; see the header note. */
+  considerPlan: (s: Session) => Promise<unknown>;
+  /** Re-drive autopilot's turn-end handler. Self-guarding via eligible(); see the header note. */
+  autopilotDone: (id: string) => Promise<void>;
+  now?: () => number;
+  idleThresholdMs?: number;
+  maxAttempts?: number;
+}
+
+export class TurnEndBackstopService {
+  private now: () => number;
+  private idleThresholdMs: number;
+  private maxAttempts: number;
+  private debounce = new Map<string, DebounceEntry>();
+  /** True while a sweep is in flight. The sweep awaits each dispatch, so a slow one can still be
+   *  running when the next tick fires; without this an overlapping sweep would re-pass readyToFire
+   *  for the same session (firedThisEpisode is only set once the dispatch resolves) and deliver a
+   *  duplicate. The tick is DROPPED rather than queued — this is an idle-debounced recovery, so the
+   *  next tick re-evaluates from fresh state. Mirrors BuildQueueReminderService's guard. */
+  private sweeping = false;
+
+  constructor(private deps: Deps) {
+    this.now = deps.now ?? Date.now;
+    this.idleThresholdMs = deps.idleThresholdMs ?? DEFAULT_IDLE_THRESHOLD_MS;
+    this.maxAttempts = deps.maxAttempts ?? DEFAULT_MAX_ATTEMPTS;
+  }
+
+  /** Drop a session's debounce state (call on archive). */
+  forget(id: string): void {
+    this.debounce.delete(id);
+  }
+
+  /**
+   * Record that the REAL `done` status edge was routed for this session, so the backstop stands
+   * down for the current resting episode. Fed from the `session:status` subscriber in index.ts.
+   *
+   * Belt-and-braces rather than load-bearing — a redundant `considerPlan` is free — but it keeps a
+   * healthy session from paying an autopilot classify it doesn't need: onDone() can legitimately
+   * classify and then leave the session idle (e.g. an `unknown` verdict that just surfaces), which
+   * without this flag the backstop would re-drive every episode.
+   */
+  markDelivered(id: string): void {
+    this.entry(id).delivered = true;
+  }
+
+  private entry(id: string): DebounceEntry {
+    let e = this.debounce.get(id);
+    if (!e) {
+      e = {
+        settledSince: null,
+        firedThisEpisode: false,
+        delivered: false,
+        sawActive: false,
+        attemptCount: 0,
+      };
+      this.debounce.set(id, e);
+    }
+    return e;
+  }
+
+  /** End the current resting episode. `sawActive`/`attemptCount` survive — they are per-session
+   *  evidence, not per-episode state. */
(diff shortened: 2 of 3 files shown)


scoop:
# Review — `fix(plan-gate): recover turn-end edges herdr never reports as done`

Diff reviewed: `git diff f08cd6285...HEAD` — `src/index.ts` (+20), `src/turn-end-backstop.ts` (new, 227), `test/turn-end-backstop.test.ts` (new, 212).

## What I verified

Identifiers and contracts the diff relies on, each opened and compared:

- `isSettledIdle(status, idleMs, thresholdMs)` — `src/recap-core.ts:441-443`. Signature and argument
  order in `src/turn-end-backstop.ts:184` match; it is `(idle||done) && idleMs >= thresholdMs`, so the
  extra `isResting` pre-check at `src/turn-end-backstop.ts:43-45` is redundant-but-harmless, not wrong.
- `store.list({ activeOnly: true })` — `src/store.ts:2889-2896`. Returns non-archived hydrated
  `Session[]`. The `Pick<SessionStore, "list">` narrowing in `src/turn-end-backstop.ts:71` is satisfied.
- `SessionStatus` — `src/types.ts:9` is `running | idle | blocked | done | archived`, and
  `planPhase` — `src/types.ts:107` is `"planning" | "executing" | null`. `isResting` covers exactly
  `idle|done`; `recover()`'s two-way phase routing (`src/turn-end-backstop.ts:213-214`) covers all
  three `planPhase` values.
- `mapState` — `src/herdr.ts:140-150`: only herdr `done` → Shepherd `done`, everything unrecognised
  → `idle`. Confirms the premise of the fix.
- `PlanGateService.consider()` — `src/plan-gate.ts:758-780` plus `skipForPlanHash`
  (`src/plan-gate.ts:794-810`). The header comment's list of short-circuits in
  `src/turn-end-backstop.ts:22-25` is accurate: gate-off, non-planning, `inflight`/`starting`,
  approved, missing/empty plan, unchanged hash, refused spawn. The `starting` claim is taken
  synchronously before any `await` (`src/plan-gate.ts:772`), so a backstop call racing the fast path
  cannot double-spawn. Verified.
- `AutopilotService.onDone()` — `src/autopilot.ts:470-499` → `consider()` → `eligible()`
  (`src/autopilot.ts:273-298`). The header comment's stand-down list in
  `src/turn-end-backstop.ts:26-28` is accurate. Note `onDone` runs `pendingAuthUrl`, `refreshPr`,
  `reEngageCi`, `reEngageRebase` *before* `eligible()`, but `reEngageCi` gates on `this.enabled(s)`
  (`src/autopilot.ts:669`), so an autopilot-off session is never steered by this path — the only cost
  is one `refreshPr` and a tail read, bounded by once-per-episode.
- `shouldConsiderOnSettle` — `src/plan-gate.ts:92-102`. Confirms the root-cause claim: a first draft
  (no prior gate) only fires on `done`.
- TDZ safety of `autopilotDone: (id) => autopilot.onDone(id)` at `src/index.ts:1902` while
  `const autopilot` is declared at `src/index.ts:1993`: the constructor stores the arrow without
  calling it, and the only caller is the `setInterval` registered inside `deferredStarts.push`
  (`src/index.ts:1904`), flushed by `startBackground()` (`src/index.ts:554-557`) after module
  evaluation. Safe, and the in-diff comment at `src/index.ts:1896-1897` says so correctly.
- `forget()` wiring on archive — `src/index.ts:1971`, in the same `session:archived` subscriber as
  `buildQueueReminder.forget` (`src/index.ts:1970`). Correct.
- `markDelivered` wiring — `src/index.ts:2232-2236`, inside the existing `session:status`
  subscriber. `session:status` with `done` is genuinely emitted on the transition
  (`src/poller.ts:1031-1040` → `onChange`, `src/index.ts:1095-1113`), including via the deferred
  `scratchpadHasFiles` path which has a `.catch` fallback emit — so the mark is reliable.
- Ordering: the backstop calls `autopilot.onDone` directly, bypassing the `#1094` SessionRouter
  seam that orders drain before autopilot (`src/index.ts:2197-2206`). I checked whether that matters:
  `drain.handle()` on a `status` change only calls `pumpIfEnabled` (`src/drain.ts:3314-3315`), which
  `drain.tick()` also does every 30s (`src/index.ts:2246-2251`) — four passes before the backstop's
  120s threshold. Not a defect.
- No user-facing strings are added (only `console.log`/`console.warn`), so there is no message
  catalog or locale parity surface to check. Confirmed by reading the whole new file.
- No new dependency, no new HTTP/WS surface, no change to `mapState`, `shouldConsiderOnSettle`, or
  `status` semantics — the plan's stated out-of-scope boundaries are respected; `src/push.ts` is
  untouched as the plan said.

## Finding

### 1. Evidence arming and episode reset ride the 15s sweep only — the exact hole `#1617` already closed for the sibling service

`src/turn-end-backstop.ts:171-175` is the only place that sets `sawActive` and the only place that
calls `resetEpisode()` (clearing `delivered`, `firedThisEpisode`, and the stale `settledSince`). It
runs from the 15s sweep (`src/index.ts:1936-1938`), so it only ever sees a *point-in-time sample* of
`status`.

The repo already learned that this is not enough. `BuildQueueReminderService` originally armed
`sawRunning` from its own 15s sweep and had to be fixed: `src/build-queue-reminder.ts:55-58` —
"the sweep's own 15s sample can miss a herdr-`working` burst that starts and finishes between ticks
— `status` is a point-in-time level, not a 'recently-active' latch (#1617)" — and the fix was
`markRan()` (`src/build-queue-reminder.ts:96-108`), fed from the poller's 1 Hz `session:status`
`running` events at `src/index.ts:1326-1337`. The new service copies the `sawRunning` *field* but not
the arming path, even though the diff edits the very subscriber that would supply it
(`src/index.ts:2232-2236`).

Two consequences, both on paths that execute today:

**(a) The evidence gate can never arm.** A session whose activity between two sweep samples is
shorter than 15s never sets `sawActive`, so `readyToFire` (`src/turn-end-backstop.ts:194`) is false
forever and the backstop silently does nothing — the same hang, just rarer.

**(b) A missed running burst leaves the previous turn's episode state in place.** This is the sharper
one, because `delivered` is per-episode and only cleared by `resetEpisode`:

- Turn N ends with a real `done` edge → `markDelivered` sets `delivered = true`
  (`src/index.ts:2235`). Autopilot classifies and steers.
- The session runs and finishes turn N+1 inside a window the sweep never samples as non-resting.
- `delivered` is still `true`, so if turn N+1's `done` edge *is* lost (the 1-in-5 this PR exists to
  fix), `readyToFire` returns false at `src/turn-end-backstop.ts:193` and the session hangs anyway.

The same stale state has a second edge: when `delivered`/`firedThisEpisode` happen to be clear, the
carried-over `settledSince` means the dwell is measured from the *previous* rest. A session that
resumed work 15 seconds ago can satisfy `now - settledSince >= 120_000` at
`src/turn-end-backstop.ts:184` and fire mid-turn — which for a `planning` session means spawning the
adversarial reviewer against a half-written `.shepherd-plan.md` and burning a review round. That is
precisely the failure mode the house rule ("trigger on settled idle, not the first idle tick;
sessions go idle after every steer, so first-idle firing runs mid-work") exists to prevent, and the
comment at `src/turn-end-backstop.ts:177-178` claims the service honours it.

The fix mirrors the precedent: feed non-resting `session:status` transitions into the service from
the subscriber at `src/index.ts:2232-2236` (a `markActive(id)` that sets `sawActive` and calls
`resetEpisode`), and keep the sweep sample as the coarse fallback it already is.

The backstop can dispatch `autopilotDone` for a session whose status is `done` (`isResting` at src/turn-end-backstop.ts:43-45 accepts it), and `onDone` can steer — the sibling nudge service deliberately excludes `done` for that reason (src/build-queue-reminder.ts:14-15); it reads as acceptable here because re-driving autopilot's own turn-end handler on `done` is exactly what the fast path does at src/autopilot.ts:449, not a nudge/reconcile steer.

`recover()` increments `attemptCount` before dispatch (src/turn-end-backstop.ts:212) and discards `considerPlan`'s `PlanReviewTrigger` return (src/plan-gate.ts:758), so a `skipped` / `plan-unavailable` no-op spends one of the three lifetime attempts — e.g. a `planning` session resting >120s while a reviewer is `inflight` burns a slot for nothing.

None — the diff is exactly the three files the plan named, with no extra surface.

## Stated limitations

- The premise that viewing a herdr pane permanently clears `agent_status: done` back to `idle` is a
  live-instance observation recorded in the plan; herdr is not in this repo, so I could not verify it
  against ground truth. I assumed it as written. Nothing in the finding above depends on it.
- I did not run lint or tests (CI owns those, per the review brief).

## Verdict

One important finding. Everything else in the diff checks out against the code I opened above: the
guards are correct in isolation, the two dispatch targets really are self-guarding, the wiring points
in `src/index.ts` are the right ones, and the restart-safety argument holds. The defect is that the
service's per-episode state is driven by a sampling loop the repo has already documented as unable to
observe the transitions this state depends on.


**Nits (non-blocking):**

- [bug] src/turn-end-backstop.ts: `recover()` bumps `attemptCount` before dispatch (src/turn-end-backstop.ts:212) and throws away considerPlan's PlanReviewTrigger result, so a no-op dispatch (`skipped` / `plan-unavailable` per src/plan-gate.ts:758-780) spends one of the three lifetime attempts. A planning session that rests past the threshold while a reviewer is already `inflight` burns a slot for nothing; consider only counting attempts that actually did work, or scoping the cap to a window rather than the session lifetime.
- [bug] src/turn-end-backstop.ts: `attemptCount` is a lifetime cap that is never reset (resetEpisode at src/turn-end-backstop.ts:131-135 deliberately leaves it), so a long-lived autopilot session permanently loses backstop protection after three recoveries and reverts to today's hang behaviour for every later turn. Fine for the plan-creation case the task names; worth a note or a wider bound for the executing-phase path.

scoop (commented):
# Review — turn-end backstop (rework round 2)

Diff under review: `git diff f08cd6285...HEAD` → `src/index.ts`, `src/turn-end-backstop.ts` (new),
`test/turn-end-backstop.test.ts` (new). 534 insertions, 0 deletions.

## Prior finding — confirmed addressed

The round-1 finding was that `sawActive` was armed (and `resetEpisode()` called) only from the 15s
sweep sample, so a working burst falling between two sweeps was invisible — with three named
consequences: the evidence gate never arms, a stale `delivered` suppresses the next turn's lost
edge, and a carried-over `settledSince` fires the backstop mid-turn.

The new commit `79321b777` adds `markActive(id)` (src/turn-end-backstop.ts:133-137), which sets
`sawActive` **and** calls `resetEpisode(e)` (src/turn-end-backstop.ts:156-159, clearing
`settledSince`, `firedThisEpisode` and `delivered`), and feeds it from the 1 Hz `session:status`
subscriber at src/index.ts:2243 (`if (status === "running" || status === "blocked")`). The sweep's
own sample now delegates to the same method (src/turn-end-backstop.ts:194-196), so the two paths
cannot diverge. All three named consequences are closed, and the three tests at
test/turn-end-backstop.test.ts:184-224 cover exactly them ("arms the evidence gate for a burst the
sweep never samples", "clears a previous turn's delivered flag", "resets the settle clock so a
resumed session can't fire mid-turn").

I verified the arming source is real: `StatusPoller.reconcileAgent` emits `onChange(s.id, status)`
whenever `idChanged || status !== s.status || agent.agentStatus !== s.lastState`
(src/poller.ts:1031-1040), and `onChange` is the callback that emits `session:status`
(src/index.ts:1095-1113). `mapState` (src/herdr.ts:140-151) maps herdr `working` → `running` and
`blocked` → `blocked`, so a turn that herdr observes at all produces at least one `running` event.
This mirrors the precedent the round-1 finding cited: `BuildQueueReminderService.markRan`
(src/build-queue-reminder.ts:106-108) fed from src/index.ts:1326-1337.

Resolved: not re-raised.

## Verification performed

Every identifier the diff introduces or relies on, checked against the tree:

- `isSettledIdle(status, idleMs, thresholdMs)` — verified against src/recap-core.ts:441-443; the
  call at src/turn-end-backstop.ts:206 passes `(s.status, now - e.settledSince, this.idleThresholdMs)`
  in that order. The function's own `idle || done` test is redundant after `isResting` but harmless.
- `SessionStatus = "running" | "idle" | "blocked" | "done" | "archived"` — verified against
  src/types.ts:9. `isResting` (src/turn-end-backstop.ts:43-45) covers `idle | done`; the remaining
  three all route to `markActive`. `archived` is unreachable here because the sweep lists
  `activeOnly` (see next item).
- `store.list({ activeOnly: true })` — verified against src/store.ts:2889-2890
  (`WHERE status != 'archived'`), so the sweep never sees an archived row. `SessionStore` is exported
  at src/store.ts:1239, so `Pick<SessionStore, "list">` (src/turn-end-backstop.ts:71) resolves.
- `planPhase: "planning" | "executing" | null` — verified against src/types.ts:107. The routing at
  src/turn-end-backstop.ts:235-236 sends `planning` → `considerPlan` and both other values →
  `autopilotDone`; the `?? "none"` at line 232 is log text only.
- `planGate.consider(session, opts?) => Promise<PlanReviewTrigger>` — verified against
  src/plan-gate.ts:758-778. The "redundant call is free" claim holds: it short-circuits on
  `planPhase == null`, `planPhase !== "planning"`, `inflight`/`starting`, an approved prior gate, an
  empty plan, and (via `skipForPlanHash`, src/plan-gate.ts:798-812) an unchanged hash or a spawn
  already refused for this plan text. The `starting` claim is taken synchronously before the first
  await, so the backstop cannot race the router path into a double spawn.
- `autopilot.onDone(id) => Promise<void>` — verified against src/autopilot.ts:470-501. Its own
  docstring reads "agent exited / **idled**", and it re-reads the live row rather than consuming a
  snapshot (src/autopilot.ts:440-444), so calling it for an `idle` session is exactly its contract.
- TDZ safety of the lazily-captured `autopilot`: the service is constructed at src/index.ts:1899 but
  `const autopilot` is at src/index.ts:1993. The only call site is inside the `deferredStarts`
  interval registered at src/index.ts:1904, drained by `startBackground()` (src/index.ts:554-557)
  after module evaluation — the file documents this contract at src/index.ts:548-553. `planGate` is
  at src/index.ts:1567, already initialised. No TDZ.
- Wiring completeness against the plan's four points: construct (src/index.ts:1899-1903), sweep on
  the existing 15s interval beside `buildQueueReminder.sweep()` (src/index.ts:1936-1938, inside the
  `maintenance.active` gate at src/index.ts:1906), `markDelivered` on the `done` edge
  (src/index.ts:2236), `forget` on archive next to `buildQueueReminder.forget`
  (src/index.ts:1970-1971, inside the `session:archived` branch). All four present.
- Other `session:status` emitters cannot poison the new marks: src/server.ts:3340 (resume) and
  src/server.ts:3690 (replaceAgent) both emit a genuinely-just-started session, so `markActive` is
  correct there; server.ts:3690 emits the whole `Session`, which still destructures to `{id, status}`.
- i18n / catalogs: no user-facing strings are added — the only new text is `console.log` /
  `console.warn` at src/turn-end-backstop.ts:239-247 and src/index.ts:1938. The i18n rule's `paths`
  front-matter (.claude/rules/i18n.md:2-9) covers `ui/**`, `extension/**`, `src/push.ts` and
  `src/ready-notify.ts` — none of which this diff touches — so no catalog parity obligation arises.
  No glossary obligation either: .claude/rules/ui-glossary.md is scoped to `ui/` paths.
- House rules checked against the diff: the plan-phase branch is guarded on
  `planPhase === "planning"` (src/turn-end-backstop.ts:235); the fire is on *settled* idle, not the
  first idle tick (src/turn-end-backstop.ts:201-206, with the first sighting stamping the clock only);
  and no `done`/`running` session is steered — a session at `done` has already had `markDelivered`
  called from src/index.ts:2236, so `readyToFire` (src/turn-end-backstop.ts:212-218) rejects it, and
  a `running` session is not resting at all.
- Re-entrancy: the `sweeping` drop-the-tick guard (src/turn-end-backstop.ts:169-183) with
  `finally`-reset is byte-for-byte the sibling's shape at src/build-queue-reminder.ts:140-160,
  including the identical `live`-set pruning loop.

Tests read (test/turn-end-backstop.test.ts:45-258): they cover the regression, the pre-threshold
non-fire, the restart-safety gate, both `delivered` shapes, once-per-episode, re-arming, the
lifetime cap, both routing branches, `blocked` never accruing settle time, `forget`, the
retry-on-rejection-but-bounded path, the three `markActive` cases, and sweep pruning. I did not run
them (CI owns that).

## Limitations

- herdr's actual `done`→`idle`-on-view behaviour is the premise of the whole change and lives
  outside this repo; I could not verify it against a live herdr. The diff's handling is
  self-consistent either way — the backstop is a no-op when `done` arrives.

## Findings

No blocking defects found. Four non-blocking points are recorded as nits.

- `src/push.ts:551`'s `status !== "done"` guard drops the "session finished" push for a watched pane — the same root cause, deliberately excluded by the plan and still unaddressed.
- `src/plan-gate.ts:92 shouldConsiderOnSettle()` still keys the first review on `done` alone, so the fast path remains lossy by design; the backstop only papers over it after ~135s.
- `src/build-queue-reminder.ts:140-160`'s sweep has the same latent wedge as the new one: a dispatch promise that never settles leaves `sweeping` true forever, since the `finally` only runs when the whole loop completes.

- src/turn-end-backstop.ts:169-183 — if `considerPlan`/`autopilotDone` ever returns a promise that never settles (today both reject or resolve), `sweeping` stays true and the backstop is silently wedged off for every session, not just the stuck one; a per-dispatch timeout would bound it. Same shape as the sibling service, so not a regression this PR introduces.

- src/turn-end-backstop.ts:229-248 — `recover()` counts *failed* attempts toward the lifetime cap and keeps the episode unburned so transient failures retry; the plan described a plain fire cap. Reasonable, but it is extra machinery the task did not ask for.
- src/index.ts:2243 — arming on `blocked` as well as `running` goes beyond the sibling precedent (`markRan` arms on `running` only, src/index.ts:1329) and beyond what the plan described.


**Nits (non-blocking):**

- [bug] src/turn-end-backstop.ts:217 — `attemptCount` is a per-session LIFETIME cap that is never reset, not even after a successful recovery followed by a healthy `done` edge. For the plan-gate case 3 is ample, but an executing session under autopilot whose pane the operator habitually watches loses the backstop permanently after its third recovered turn and hangs for the rest of its life — the exact failure the task describes, just deferred. Consider resetting the counter on a recovery that was followed by a delivered `done`, or scoping the cap to consecutive failures rather than lifetime attempts.
- [bug] src/turn-end-backstop.ts:238-240 — `firedThisEpisode` is set whenever the dispatch RESOLVES, regardless of outcome, so a `consider()` that returned "skipped" (a review momentarily in flight) or "plan-unavailable" (the plan file not yet readable) burns the episode and is never retried while the session stays at rest. Only a rejection keeps the episode open. Keying the burn on `PlanReviewTrigger` (treating "skipped"/"plan-unavailable" like a rejection) would make the recovery more robust for the free-no-op cases the header comment relies on.
- [bug] src/index.ts:2243 — arming `sawActive` from `blocked` as well as `running` is broader than the sibling precedent (`buildQueueReminder.markRan` is gated to `running` only, src/index.ts:1329) and slightly weakens the restart-safety claim the service documents at src/turn-end-backstop.ts:59-63: a session sitting at `blocked` when the process boots is armed by the very first sample, so its later decay to `idle` can spend an autopilot classify for work this process never observed. Worth a word in the comment, or gating to `running` and letting the sweep cover `blocked`.
- [bug] src/turn-end-backstop.ts:236 — the backstop calls `autopilot.onDone(id)` directly, bypassing the SessionRouter seam (src/index.ts:2203-2208) whose whole point is that drain retires a session BEFORE autopilot spends a classify on it (#1094). It happens to be safe today because `drain.handle` on a status change only calls `pumpIfEnabled` (src/drain.ts:3312-3314) and drain has its own timer tick, but routing the recovery through `sessionRouter` would preserve the documented ordering rather than rely on that.

PR #2269, opened 2026-09-11
fix(herd): let the cold-resume chip explain itself on hover and click

kai-osthoff:
## What

The `⚠ cold · ≈3.1 units` chip on a session card named a cost without ever saying what it meant or why Shepherd was showing it. Its only explanation was a native `title` — mouse-only, invisible to touch and keyboard, silent for a screen reader, and limited to the reading itself, never the rationale.

The chip now carries a real explanation affordance:

- **hover** opens it, a **genuine click pins** it, Esc or an outside click dismisses it;
- the prose reaches assistive tech through `aria-description` (the native `title` is gone, so they can't double up);
- it says **what is true of this session** (cache expired, how much context has to be re-sent, what the next turn costs) **and why the number is on the card at all** — resuming is the moment the money is spent, so seeing it here lets you pick a still-warm session instead, hand the task to a fresh one, or resume knowingly.

## How

`use:statusTip` — the row's own explanation affordance, already on the designation/environment segment right next to this chip. It also raises the chip above the `.unit-hit` overlay and stops the click, so reading the explanation never doubles as selecting the row. That was the objection that previously ruled out a nested `GlossaryTerm`; a GlossaryTerm is still wrong here for a different reason — its click presentation pushes content down and would reflow the card.

`wide: true` keeps the two-sentence body from stacking into a 20-line column.

The status bar's marker is deliberately untouched: it already marks *cold cache* as a glossary term, and its "why look before you open it" rationale doesn't apply once you are inside the session.

## Catalog

No new What's-New entry. The chip's own card (`cold-resume-marker`) ships in the same **unreleased** 1.48.0, so its body is extended to describe the affordance rather than announcing the same chip twice. Labelled `fix` for the same reason — this completes an affordance that is already registered, it isn't a second feature.

## Verification

| Gate | Result |
| --- | --- |
| `ui/ bun run check` | 6255 files, 0 errors |
| `ui/ bun run check:i18n` | 2 locales in parity (3974 keys) |
| `ui/ bun run test` | 315 files, 5133 tests passed |
| `bun run lint` | clean |
| `scripts/check-glossary.mjs` | 30 terms, no dangling markers |
| `scripts/check-feature-catalog.sh` | pass |
| pre-push | pass (84.9s) |

Three specs cover the new behaviour in `UnitRow.browser.test.ts`: the chip carries reading + rationale via `aria-description` with no native `title` left, and hover reveals the `role="tooltip"` popover while the click never reaches the row.


scoop (commented):
# Critic review — `fix(herd): let the cold-resume chip explain itself on hover and click`

Range reviewed: `git diff 5266bb4f0...HEAD` — 4 files (`ui/messages/en.json`, `ui/messages/de.json`, `ui/src/lib/components/UnitRow.svelte`, `ui/src/lib/components/UnitRow.browser.test.ts`).

**Decision: comment.** I found no correctness, security, contract or catalog defect in the diff. Everything raised below is non-blocking.

## Task text vs. diff (stated limitation, not a finding)

The task I was handed describes an *operator chore* — "take ONE clean marketing screenshot of Shepherd's Critic for an X post … save to `/tmp/shepherd-critic.png` and report the path". That task inherently produces no repository diff, and nothing in this diff serves it: the change is a follow-up to the cold-resume marker (#2042) on the session card, coherent end to end with its own commit message and with the existing code it touches (`ui/src/lib/feature-announcements/entries/v1.48.0-cold-resume-marker.ts:4-7` already carries the #2042 announcement this diff extends). I read this as a mismatched task/diff pairing in the review input rather than an author who ignored the brief, so I reviewed the diff against its own stated intent and did **not** manufacture a "does not satisfy the task" finding out of it. If the screenshot chore really is the intent of this branch, then none of these four files advances it and no PNG was produced — that is a routing problem to fix upstream, not something the author can fix by editing this code.

## What I verified

**`statusTip` exists and supports every option the call site passes.** `use:statusTip={{ text: coldResumeTip, wide: true }}` at `ui/src/lib/components/UnitRow.svelte:985` resolves to `ui/src/lib/actions/statusTip.svelte.ts:43`; the import was already present at `ui/src/lib/components/UnitRow.svelte:48` (pre-existing, used by `.meta-text` at line 981), so nothing is unresolved. `wide` is a real param (`ui/src/lib/actions/statusTip.svelte.ts:16`, applied at `:59` → `.status-tip-wide`), and `.status-tip-wide[popover]` is defined at `ui/src/app.css:689` *after* the base `.status-tip[popover]` rule at `:653`, so the 420px width actually wins (the comment at `ui/src/app.css:682-683` says as much).

**The behaviours the new comment and tests assert are the action's real behaviours.** `aria-description` is set in `enable()` at `ui/src/lib/actions/statusTip.svelte.ts:186`; the native `title` is stripped at `:193` (so `expect(chip.hasAttribute("title")).toBe(false)` at `ui/src/lib/components/UnitRow.browser.test.ts:1603` is testing something the action genuinely does, not a leftover of the markup); click pins only on a genuine pointer click (`:170`); `stopClickPropagation` defaults to `true` (`:49`, `:178`); the popover carries both `status-tip` and `status-tip-wide` classes (`:59`), so the test's `.status-tip:popover-open` selector (`ui/src/lib/components/UnitRow.browser.test.ts:1615`) matches. The dispatch style matches existing repo precedent (`ui/src/lib/components/StatusPip.browser.test.ts:19,24`).

**The z-index raise actually clears the row's click overlay.** `statusTip` sets inline `position: relative; z-index: 1` (`ui/src/lib/actions/statusTip.svelte.ts:189-190`); `.unit-hit` is `position: absolute; inset: 0; z-index: 0` (`ui/src/lib/components/UnitRow.svelte:1158-1161`). Both sit in the stacking context of `.unit`, which is `position: relative` with no transform/z-index (`:1113-1114`), and the intermediate `.meta` span (`:1587-1601`) sets no `z-index`, `transform` or `opacity`, so it establishes no stacking context of its own — the chip really does paint and hit-test above the overlay. This is the same recipe `.chip-manual-steps--link` already uses (`:1641-1643`). Worth noting this is why the change is a genuine fix rather than a polish: with the chip unpositioned, the absolutely-positioned overlay covered it, so the old native `title` was never the hover target and in practice never appeared.

**i18n parity holds.** `coldresume_why` exists in both catalogs — `ui/messages/en.json:3975` and `ui/messages/de.json:3975` — with no interpolation placeholders in either, so there is no placeholder mismatch to check. `ui/scripts/check-i18n.mjs:46-53` only enforces an identical, non-empty key set across locales, which this satisfies; it does not require ordering. Neither new nor edited value contains a `[[id|Label]]` glossary marker, which matters because `statusTip` renders its text via `textContent` (`ui/src/lib/actions/statusTip.svelte.ts:77`) and would print a marker literally — confirmed against the values at `ui/messages/en.json:3975-3976` and `ui/messages/de.json:3975-3976`.

**Editing the existing announcement body instead of adding a catalog entry is correct here.** The entry's `sinceVersion` is `1.48.0` (`ui/src/lib/feature-announcements/entries/v1.48.0-cold-resume-marker.ts:5`) and the last released version is `1.47.0` (`.release-please-manifest.json:2`), so no operator has seen that card yet — extending its body cannot retroactively change a shipped What's-New entry. The feature-catalog gate is armed only by `feat(...)` subjects (`scripts/check-feature-catalog.sh:58-62`) and this is a `fix(herd):` commit, so it passes rather than demanding a new fragment; the rule text agrees (`.claude/rules/ui-feature-catalog.md:26`).

**No other consumer depended on the removed attribute.** `chip-cold-resume` / `coldresume` appear only in `UnitRow.svelte`, `UnitRow.browser.test.ts`, `SessionStatusBar.svelte` and the two catalogs; nothing outside the diff read the chip's `title`. `SessionStatusBar.svelte:93` still calls `m.coldresume_title(...)` with the same `{context, units}` shape, which this diff did not change — no contract broken.

**Nothing ancestral depends on the click it now stops.** Row selection is the `.unit-hit` button's own `onclick` (`ui/src/lib/components/UnitRow.svelte:859-868`); `.unit` (`:847-856`) has no click handler, and the only ancestor click listener is `swipeGesture`'s capture-phase guard (`ui/src/lib/components/swipe.ts:134-142,148`), which runs before any bubble-phase `stopPropagation` and only fires when a drag happened or the drawer is open. So the behaviour change is exactly the intended one — the chip area no longer opens the session — and it is deliberate, documented and covered by a test.

### Limitations I could not close in-repo

- Whether screen readers actually announce `aria-description` on a non-focusable generic `<span>` cannot be verified from this repository; `ui/src/lib/actions/statusTip.svelte.ts:37-41` asserts it and the pre-existing `.meta-text` usage relies on the same assumption. I neither confirm nor dispute it, and it is not introduced by this diff.
- Per instructions I did not run the test suite or lint; my reading of the new tests is static.

- `ui/src/lib/components/UnitRow.svelte`: none of the curated rules is violated by this diff — no plan-phase branch, no session steering, no idle trigger, no toast, no Svelte child extraction (no CSS moved; `.chip-cold-resume` at `:1623` keeps its rule), no relative-age helper, no BacklogView selection path, no WS emitter. The one rule that does bite — EN+DE parity for any new user-facing string — is met (`ui/messages/en.json:3975`, `ui/messages/de.json:3975`).

- `ui/src/lib/components/SessionStatusBar.svelte:93` renders the cold-resume marker with `coldresume_title` only, so the in-session marker still explains the reading but not the *why* the Herd chip now carries; worth a follow-up for consistency.
- `ui/src/lib/actions/statusTip.svelte.ts:205-222` — `teardown()` clears `node.style.zIndex` but leaves the inline `position: relative` it set in `enable()` at `:189`, so a node that loses its tip keeps a positioning override it did not start with.
- `ui/src/lib/components/UnitRow.browser.test.ts` relies on `vitest-browser-svelte`'s implicit between-test cleanup rather than clearing the DOM in its own `beforeEach` (`:124`), so every `document.querySelector` in the file would silently read a stale row if that cleanup ever changed; pre-existing throughout the file.


**Nits (non-blocking):**

- [bug] ui/src/lib/components/UnitRow.svelte: the comment at 608-615 presents the keyboard gap as closed (focus opens it for the keyboard, Esc dismisses) after criticising the old title for reaching neither touch nor keyboard, but .chip-cold-resume is still a plain span with no tabindex, so statusTip's focus/blur/keydown listeners (statusTip.svelte.ts:198-201) can never fire on it — keyboard users get only the aria-description. Either add tabindex=0 or trim the claim to what this element actually gains: hover, touch and AT.
- [compliance] ui/src/lib/components/UnitRow.svelte:623-628 assembles the tip by concatenating two whole catalog sentences with a hardcoded space; a single coldresume_tip message (or one taking the reading as a parameter) would leave sentence order and joining punctuation to the translator instead of fixing them in code.
- [bug] ui/src/lib/components/UnitRow.svelte:613 credits the wrong mechanism: swallowing the click is belt-and-braces here, since row selection lives on the sibling .unit-hit button (line 861) that a click on the chip can never reach by bubbling, and .unit (847) has no click handler — the z-index raise is what stops the row selecting.
- [compliance] ui/messages/en.json:3974 (and ui/messages/de.json:3974): the new sentence follows immediately after the clause about the matching marker in the session status bar, so it reads as if both surfaces explain themselves; only the Herd chip gained the affordance (SessionStatusBar.svelte:93 still renders coldresume_title alone). Naming the Herd chip explicitly would avoid the implication.
- [bug] ui/src/lib/components/UnitRow.browser.test.ts: nothing covers wide: true — no assertion that the opened panel carries the status-tip-wide class, even though the prose variant is the reason a two-sentence tip stays readable instead of stacking into a column.

scoop (dismissed):
# Review — `fix(herd): let the cold-resume chip explain itself on hover and click`

Diff reviewed: `git diff 5266bb4f0fb75b845cd9fdcc3ab69f1f5a723fd6...HEAD` — 4 files
(`ui/messages/de.json`, `ui/messages/en.json`, `ui/src/lib/components/UnitRow.browser.test.ts`,
`ui/src/lib/components/UnitRow.svelte`).

## What I verified

**Catalog parity (EN/DE).** `coldresume_why` exists in both catalogs at the same position —
`ui/messages/en.json:3975` and `ui/messages/de.json:3975` — and neither is empty. The amended
`feat_cold_resume_marker_body` is present in both (`ui/messages/en.json:3974`,
`ui/messages/de.json:3974`). The gate that enforces this is `ui/scripts/check-i18n.mjs:44-53`
(identical, non-empty key sets across every `ui/messages/*.json`); `ls ui/messages/` shows only
`en.json` and `de.json`, so those two are the whole parity surface. `coldresume_why` takes no
placeholders in either locale, so there is no placeholder mismatch to check.

**The key is actually consumed, and spelled the same way.** `m.coldresume_why()` is called at
`ui/src/lib/components/UnitRow.svelte:627` and in the test at
`ui/src/lib/components/UnitRow.browser.test.ts:1601` and `:1616`; both resolve to the catalog key
`coldresume_why`. No other spelling of it exists anywhere in `ui/src` (grepped `coldresume_`).

**No glossary markup leaks into a text-only surface.** `statusTip` renders via `pop.textContent`
(`ui/src/lib/actions/statusTip.svelte.ts:77`), so `[[term|label]]` markup would print literally.
`coldresume_title` and `coldresume_why` carry none (`ui/messages/en.json:3975-3976`,
`ui/messages/de.json:3975-3976`) — unlike `coldresume_label` (`en.json:3971`), which is the
GlossaryText-rendered status-bar string and is correctly *not* the one fed to the tip.

**The numbers in the tip.** `formatTokens(180_000)` → `"180k"` per `ui/src/lib/format.ts:140-144`
(`n < 1_000_000` → `(n/1000).toFixed(0) + "k"`), which is what the test asserts at
`UnitRow.browser.test.ts:1600`. `formatUnits(1.572)` → `"1.6"` per
`ui/src/lib/components/usage/format.ts:17` (`n < 10` → `toFixed(1)`), matching the chip assertion at
`UnitRow.browser.test.ts:1590`.

**Click no longer selects the row.** `.unit-hit` is the row's only click target — a `<button>` with
`onclick={() => { tipHide(); onselect(session.id) }}` at `UnitRow.svelte:859-867`, styled
`position:absolute; inset:0; z-index:0` at `UnitRow.svelte:1158-1161`. `statusTip` sets the trigger
to `position:relative; z-index:1` (`statusTip.svelte.ts:189-190`), which puts the chip above that
overlay for hit-testing, and stops click propagation by default
(`statusTip.svelte.ts:167-171`, `stopClickPropagation` defaults `true` at `:178`). The chip is not
a descendant of `.unit-hit` (it is a sibling subtree under `.unit`, `UnitRow.svelte:980-988`), so
the stopped click cannot reach the row. The new test at `UnitRow.browser.test.ts:1618-1622` pins
exactly this.

**`wide: true` is a real param with real CSS.** `StatusTipParams.wide` at
`statusTip.svelte.ts:16`, applied as `status-tip-wide` at `:59`; the rule exists globally at
`ui/src/app.css:680-694` (420px, height-bounded, scrollable, and `onScrollOrResize` at
`statusTip.svelte.ts:114-121` deliberately ignores scrolls raised inside the panel so the
concatenated ~480-character body stays readable). Concatenating `coldresume_title` +
`coldresume_why` into one `wide` panel is therefore sound.

**Feature-catalog gate.** `ui/src/lib/feature-announcements.ts:3-9` requires a catalog fragment per
shipped user-facing feature; `scripts/check-feature-catalog.sh:60-64` only fires on `feat(...)`
subjects, and this branch's single commit is `fix(herd): …`, so extending the existing unreleased
entry rather than adding a second one does not trip it. `v1.48.0` is genuinely unreleased —
`.release-please-manifest.json` and `package.json:3` both hold `1.47.0`, and the entry at
`ui/src/lib/feature-announcements/entries/v1.48.0-cold-resume-marker.ts:5` targets `1.48.0`. Reusing
its body is correct, not a skipped obligation.

**Nothing else consumed the removed `title`.** Grepping `coldresume_|chip-cold-resume` across the
repo turns up only `SessionStatusBar.svelte:88-96` (untouched; still uses `coldresume_label` +
`coldresume_title` as a native title, which is the separate status-bar marker) and this component's
own tests. No out-of-diff caller asserted the chip's `title`.

## Findings

### 1. The chip is still keyboard-dead, and the new comment says otherwise (important)

`UnitRow.svelte:611-612` states, of this chip: *"hover opens it, a real click PINS it, **Esc or an
outside click dismisses**, **focus opens it for the keyboard**"*. Two of those four are unreachable
here.

`statusTip` implements them as listeners bound to the trigger node itself —
`node.addEventListener("focus", onFocus)`, `"blur"`, `"keydown"` at
`statusTip.svelte.ts:198-201`, with `onKeydown` handling `Escape` at `:172-174` and `onFocus`
calling `show()` at `:161-163`. The trigger this PR attaches them to is
`<span class="chip-cold-resume" use:statusTip={…}>` (`UnitRow.svelte:985`) — a plain `<span>` with
no `tabindex` and no `role`. A `<span>` is not focusable, so `focus`/`blur` never fire on it, and
`keydown` never reaches it either: `focus` does not bubble (the action binds `focus`, not
`focusin`), the chip's only child is the `aria-hidden` `⚠` span (`UnitRow.svelte:986`), and the
row's focusable element — the `.unit-hit` button at `UnitRow.svelte:859-861` — is a *sibling*, not
an ancestor, so its keystrokes never pass through the chip.

Concretely: once a pointer click pins the tip (`pinned = true`, `statusTip.svelte.ts:170`), Esc
cannot dismiss it. The only dismissal paths that actually run are outside `pointerdown`
(`:109-113`, `:135`) and scroll/resize (`:114-121`). And the keyboard gap the task explicitly calls
out — the old `title` being "invisible to touch and keyboard" — is only half closed: touch now works
(a tap fires `click`, which pins), keyboard still gets nothing, because the chip is not in the tab
order at any point.

Either make the claim true (give the chip a tab stop, e.g. `tabindex="0"` plus a `role`, so
`statusTip`'s already-written `focus`/`Escape` handlers fire), or cut the `Esc` and `focus` clauses
from the comment so it describes what this chip does. As written the comment will convince the next
reader that the keyboard path is handled when it is not.

### 2. The `.chip-cold-resume` style comment now contradicts the behaviour this PR ships (important)

`UnitRow.svelte:1620-1622` still reads: *"Stays a plain `<span>` so the row's single click target is
untouched."* The `<span>` half is still true; the consequence is not. After this change the chip is
raised to `z-index:1` above the `.unit-hit` overlay and swallows its own click
(`statusTip.svelte.ts:189-190`, `:167-171`) — i.e. the row's single click target now has a hole
punched in it at the chip, which is the whole point of the change and is what the *new* comment
block at `UnitRow.svelte:613-614` says. The two comments in the same file now assert opposite
things about the same element; the stale one at `:1621-1622` needs updating.

- `ui/src/lib/components/UnitRow.svelte:985` — `statusTip`'s `onBlur`/`onKeydown` paths (`statusTip.svelte.ts:164-174`) are dead for every trigger in the repo today (no `statusTip` trigger is focusable: `StatusPip.svelte:41`, `UnitRowRight.svelte:197`, `InstrumentToggle.svelte:66`, `SlashCommandMenu.svelte:73` are all non-focusable spans/divs); adding a tab stop to fix finding 1 would make them live and is the cheapest way to get the documented behaviour.

- The old `title={coldResumeTitle}` on the chip almost certainly never surfaced at all, not merely "to a resting mouse": the chip was unpositioned while `.unit-hit` is `position:absolute; z-index:0` over the whole card (`UnitRow.svelte:1158-1161`), and the codebase documents that exact swallowing for a neighbouring span at `UnitRow.svelte:924-926` ("the `.unit-hit` overlay covers this span, so its own tooltip could never surface"). The fix is more valuable than the commit message claims.
- `SessionStatusBar.svelte:88-96` still explains the cold marker with a native `title` only — the same mouse-only/AT-silent problem this PR fixes on the card, untouched here.

## Limitations
- I did not run the browser tests (CI owns that). The assertions were checked against the
  implementation by reading, not by execution: `aria-description` is set synchronously in `enable()`
  (`statusTip.svelte.ts:186`) which runs at action mount (`:224`), and `title` is removed at `:193`,
  so both assertions in `UnitRow.browser.test.ts:1596-1604` follow from the code.
- `aria-description` support on a role-less generic element varies by browser/AT engine. That is
  `statusTip`'s established mechanism repo-wide (documented at `statusTip.svelte.ts:37-41`), not
  something this diff introduces, and I have no way to verify AT announcement from the repo — so I
  am neither asserting it works nor raising it as a defect.

kai-osthoff:
Both blocking findings were real and are fixed in efb496b6 — no code change was needed, the comments were claiming more than the element does.

**1. Keyboard claim (finding 1).** Correct: a non-focusable `<span>` never fires statusTip's `focus`/`keydown` listeners, so Esc-to-dismiss and focus-to-open were fiction. Taking the second option — the chip keeps **no tab stop**, and the comment now says so and names `aria-description` as the keyboard/AT path. Rationale for not adding `tabindex="0"`: one tab stop per cold card is real noise in a long Herd, and it would make this the only focusable statusTip trigger in the repo — a change to the action's contract worth making deliberately across all of its call sites, not as a side effect here.

**2. Stale style comment (finding 2).** Fixed — it now says the hole in the row's click target is deliberate.

**Nits taken:**
- The comment credited `stopPropagation`; corrected to the z-index raise, with the belt-and-braces note.
- Announcement body now names the **Herd** chip, so it can't be read as claiming the status-bar marker gained the same affordance.
- The hover spec now asserts `status-tip-wide`.

**Nit declined — one `coldresume_tip` message instead of concatenation.** The join would cost a duplicated sentence: `coldresume_title` stays in use at `SessionStatusBar.svelte:93`, so a combined key would hold a second copy of the reading that has to be kept in sync with it by hand. Both sentences are self-contained and end in a period, and the join is a single space — the translator already controls everything inside each sentence. A duplicated string that can silently drift is the worse trade.

**Thank you for the overlay finding.** `.unit-hit` being `position:absolute; inset:0; z-index:0` over an unpositioned chip means the old `title` almost certainly never surfaced for anyone, not just "a resting mouse" — that is a better description of the bug than my commit message had, and the code comment now carries it.

**Out of scope, agreed and deliberate:** `SessionStatusBar.svelte:88-96` keeps `coldresume_title` alone. The *why* the Herd chip now carries is specifically "resuming is the moment the money is spent, so look before you open it" — that argument has no force once you are already inside the session, where the bar lives. The bar also already marks *cold cache* as a glossary term, which the card chip had no equivalent of. If the mouse-only `title` there should become a statusTip too, that is its own change; it collides with the nested `GlossaryTerm` in that span and wants thinking about, not a copy-paste.

PR #2270, opened 2026-09-11
fix(push): recover session-finished push for panes the operator watched

### LINKED ISSUE #2267: Session-finished push is dropped for any pane the operator was watching

scoop:
`src/push.ts` → `attachPush()` gates the "session finished" push on `status !== "done"`. herdr's
`done` is a *"finished, and nobody has looked yet"* notification state, not a lifecycle state —
viewing the pane clears it back to `idle` — so any session the operator had open when it finished
never reports `done` and its finish push is silently dropped.

Same lost edge #2268 fixed for the plan gate and autopilot with `TurnEndBackstopService`; push.ts
was deliberately left out of that PR's scope.

## What this does

Routes the finish push through that backstop as a **third consumer**, next to the plan gate and
autopilot. The other candidate — accepting `idle` at the push site — was rejected: `idle` is a
point-in-time level and sessions go idle after every steer, so that site would need its own
settled-idle debounce, evidence gate and per-episode dedupe, i.e. a second copy of the backstop
living in `push.ts`.

The backstop already owns every guard this push needs and `push.ts` owns none of them: the 120s
settled-idle dwell (an `idle` blip after a steer must not fire "agent finished"), the `sawActive`
evidence gate (a restart cannot emit a burst of finish pushes), once per resting episode, and the
`delivered` stand-down fed from the real `done` edge — which is what stops a healthy session being
notified twice. `attachPush()`'s `status !== "done"` gate is untouched, so the fast path still
fires instantly; the dwell costs latency only in the already-broken case.

Both callers now build the payload through one exported `notifySessionDone()`, so the recovered
push is the one the operator missed (same kind/tag/cooldown key) rather than a second, separate
notification.

## Two state-model changes this forced

**Per-consumer episode flags.** A recovery pass now dispatches two things, so the single
`firedThisEpisode` flag would let a failing plan-gate spawn replay an already-sent push on the next
15s tick.

**The runaway guard moves from a lifetime cap to consecutive failures, counted per consumer.** It
was a lifetime cap of 3 recovery *attempts* per session, counting successes. That doesn't survive a
third consumer: the lost edge repeats on **every** turn of a watched session, so a lifetime cap
means the session hangs (and goes un-notified) again from the fourth turn on. Resetting on success
keeps the guard's actual purpose — bounding a dispatch that throws every time — and re-driving once
per resting episode is exactly what a healthy `done` edge already does every turn. Counting per
consumer matters too: a plan-gate spawn that throws every time must not also silence the finish
push, which is a different dependency that is working fine.

Stated plainly: this also lifts the lifetime-3 ceiling for the existing plan-gate and autopilot
consumers, so autopilot's classify can now be re-driven once per resting episode indefinitely. That
is deliberate and is parity with the healthy path's per-turn cost; both targets remain fully
self-guarding.

## Verification

Root `bun run lint`, `bunx tsc --noEmit` and `bun run test` (9713 pass / 0 fail) are green, as is
the pre-push fallow delta audit. No UI/extension change and no new user-facing string — the `done`
push kind already has EN+DE catalog entries.

New tests cover: the recovered push for a session that rests as `idle`, phase-agnostic recovery,
the `delivered` stand-down (no duplicate), the dwell and restart-safety gates, once per episode and
re-arming on the next lost turn end, a failing push not suppressing the phase-routed recovery, a
failing phase consumer not silencing the push, and the streak resetting on a clean pass.

Closes #2267

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -88,6 +88,7 @@ import {
   attachMergePush,
   attachUsagePush,
   attachCreditsPush,
+  notifySessionDone,
 } from "./push";
 import { ReadyNotifier } from "./ready-notify";
 import { Presence } from "./presence";
@@ -1892,14 +1893,18 @@ const deliveryFacts = new DeliveryFactsService({ store });
 onSessionGit(({ id, git }) => deliveryFacts.onGit(id, git));
 // Turn-end backstop: herdr's `done` is a "finished and unseen" notification state that VIEWING the
 // pane clears back to `idle`, so an operator watching a session as it finishes destroys the only
-// edge the plan gate's first review and autopilot's onDone key on — and nothing re-checks. This
-// recovers that edge from a settled-idle sweep. `autopilot` is declared below and referenced lazily
-// (the closure runs long after boot); both targets are idempotent/self-guarding, and the `delivered`
-// mark from the real `done` edge keeps a healthy session off this path entirely.
+// edge the plan gate's first review, autopilot's onDone and the "agent finished its turn" push
+// (#2267) key on — and nothing re-checks. This recovers that edge from a settled-idle sweep.
+// `autopilot` is declared below and referenced lazily (the closure runs long after boot); all
+// targets are idempotent/self-guarding, and the `delivered` mark from the real `done` edge keeps a
+// healthy session off this path entirely — which is also what stops a duplicate finish push.
 const turnEndBackstop = new TurnEndBackstopService({
   store,
   considerPlan: (s) => planGate.consider(s),
   autopilotDone: (id) => autopilot.onDone(id),
+  // Same helper attachPush() uses on the real edge, so the recovered push is the one the operator
+  // missed (same kind/tag/cooldown key) rather than a second, separate notification.
+  notifyDone: (id) => notifySessionDone(push, store, id),
 });
 deferredStarts.push(() => {
   setInterval(() => {
--- a/src/push.ts
+++ b/src/push.ts
@@ -545,14 +545,34 @@ export function attachReviewPush(events: EventHub, store: SessionStore, push: Pu
   });
 }
 
+/**
+ * Send the "agent finished its turn" push for one session.
+ *
+ * Two callers reach the same turn end by different routes and MUST produce the same payload —
+ * same kind, same tag, same cooldown key — or a recovered push would land as a second, separate
+ * notification instead of the one the operator missed:
+ *   - {@link attachPush}, on the real `session:status` = `done` edge (the fast path), and
+ *   - TurnEndBackstopService, when herdr never reported `done` because the operator had the pane
+ *     open as the turn ended (#2267).
+ *
+ * Returns the notify promise so the backstop can see a rejection; {@link attachPush} voids it.
+ */
+export function notifySessionDone(
+  push: PushService,
+  store: Pick<SessionStore, "get">,
+  id: string,
+): Promise<boolean> {
+  const name = store.get(id)?.name ?? id;
+  return push.notify({ kind: "done", sessionId: id, tag: id, name });
+}
+
 /** Bridge F3 state events to push notifications. Both events are already edge-triggered. */
 export function attachPush(events: EventHub, store: SessionStore, push: PushService): void {
   events.subscribe((event, data) => {
     if (event === "session:status") {
       const { id, status } = data as { id: string; status: string };
       if (status !== "done") return;
-      const name = store.get(id)?.name ?? id;
-      void push.notify({ kind: "done", sessionId: id, tag: id, name });
+      void notifySessionDone(push, store, id);
     } else if (event === "session:block") {
       const { id, block } = data as { id: string; block: BlockReason | null };
       if (!block) return;
--- a/src/turn-end-backstop.ts
+++ b/src/turn-end-backstop.ts
@@ -4,20 +4,21 @@
 // state: viewing the pane clears it back to `idle`, permanently (verified live — `herdr agent focus`
 // on a `done` pane flips it to `idle` and it stays there; `agent read` and the HUD's
 // `terminal session control` attach do NOT clear it). Shepherd keys "the agent finished a turn" on
-// `done` in exactly two places:
+// `done` in exactly three places:
 //   - plan-gate.ts shouldConsiderOnSettle() — fires the FIRST plan review on `done`; on `idle` only
 //     when a prior gate already said `changes_requested`, which a first draft never has.
 //   - autopilot.ts handle() — calls onDone() only on `done`.
+//   - push.ts attachPush() — the "agent finished its turn" notification, gated on `status === "done"`.
 // So an operator watching the pane as the agent finishes destroys the only trigger, and nothing
-// re-checks: the session hangs forever. The plan gate concentrates the failure because its directive
-// makes the agent ask questions via AskUserQuestion, so the operator is on that tab exactly when the
-// plan lands.
+// re-checks: the session hangs forever and its finish push is silently dropped (#2267). The plan
+// gate concentrates the failure because its directive makes the agent ask questions via
+// AskUserQuestion, so the operator is on that tab exactly when the plan lands.
 //
 // This service does NOT change status semantics, mapState(), or shouldConsiderOnSettle(): the fast
 // path stays as-is and still fires instantly. The backstop only does work when the fast path already
 // failed, so its settle threshold costs latency ONLY in the broken case.
 //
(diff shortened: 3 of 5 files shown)


scoop (commented):
# Review — recover the session-finished push for panes the operator was watching (#2267)

Diff under review: `git diff 5266bb4f0fb75b845cd9fdcc3ab69f1f5a723fd6...HEAD` — `src/index.ts`,
`src/push.ts`, `src/turn-end-backstop.ts`, `test/push.test.ts`, `test/turn-end-backstop.test.ts`.

## Verdict

No blocking defect found. The change does what the issue asks, and the duplicate-notification
guard it leans on holds up under the paths I traced. Two nits about stale/over-stated comments.

## What I verified

**The extracted helper is genuinely the same payload as the fast path.**
`notifySessionDone()` (`src/push.ts:560-567`) builds `{ kind: "done", sessionId: id, tag: id, name }`
with `name = store.get(id)?.name ?? id`; that is verbatim the body `attachPush()` used to inline,
and `attachPush()` now calls it (`src/push.ts:575`). So kind, tag, and the cooldown key derived at
`src/push.ts:461` (`done:${sessionId}`, no `cooldownKey` override) are identical on both routes —
the claim in the helper's JSDoc is accurate.

**No new user-facing string, and the `done` copy has locale parity.** `buildPayload` routes
`kind: "done"` to `t.doneTitle`/`t.doneBody` (`src/push.ts:343`); both exist in the EN table
(`src/push.ts:144-145`) and the DE table (`src/push.ts:199-200`). Nothing new to add to either.

**Wiring resolves.** `notifySessionDone` is exported from `src/push.ts:560` and imported at
`src/index.ts:91`; the dep is passed at `src/index.ts:1907`. Both captured identifiers are declared
before the closure can run — `store` at `src/index.ts:255`, `push` at `src/index.ts:1260` — and the
closure itself is only invoked from the 15s sweep registered at `src/index.ts:1941`, long after
boot, so there is no TDZ hazard (unlike `autopilot` at `src/index.ts:1998`, which the existing
comment already calls out as lazy).

**No second `done`-kind sender to collide with.** Grepped `kind: "done"` across `src/`: the only
push-side occurrence is `src/push.ts:566`. `src/index.ts:2123` is the distinct `autopilot-done`
kind, which the real edge already emits alongside a `done` push today (`attachPush` +
`autopilot.onDone` → `onComplete`), so the recovery path's push-then-`autopilotDone` ordering
reproduces the healthy pairing rather than inventing a new one.

**Double-notification guard.** `readyToFire()` still short-circuits on `!e.delivered`
(`src/turn-end-backstop.ts:241`), and `delivered` is set from the real `done` status edge at
`src/index.ts:2241` — the same `session:status` event `attachPush` subscribes to at
`src/index.ts:1261`, so the two can't diverge. Traced the healthy sequence: `running` →
`markActive` (`src/index.ts:2248`, resets the episode via `src/turn-end-backstop.ts:183-187`) →
`done` → `markDelivered` → operator views → `idle` → sweep stamps `settledSince` → at +120s
`readyToFire` is false. New test at `test/turn-end-backstop.test.ts` ("a delivered done edge
suppresses the recovered push") pins exactly this.

**The broken case now recovers.** `isResting` accepts `idle` (`src/turn-end-backstop.ts:53-55`) and
`isSettledIdle` gates on `(idle|done) && idleMs >= threshold` (verified against
`src/recap-core.ts:441-443`), so `running → idle`-without-`done` fires once the 120s dwell elapses.
The first idle tick only stamps the clock (`src/turn-end-backstop.ts:228-231`), which satisfies the
repo's settled-idle house rule.

**Once per resting episode, not once per tick.** `owes()` (`src/turn-end-backstop.ts:249-251`) is
re-checked inside `recover()` at `src/turn-end-backstop.ts:273` and `:278`, and `run()` sets
`e.fired[c]` only on a resolved dispatch (`src/turn-end-backstop.ts:313`). I hand-traced the
"rejected dispatch" test: pass 1 pushes (fired) + plan-gate throws (streak 1); passes 2-3 re-enter
`recover()` because `owes(phase)` is still true but skip the push because `owes(push)` is false;
pass 4 `readyToFire` is false. That matches the asserted `calls === 3` / `pushes === ["S"]`.

**Restart safety is unchanged.** `sawActive` (`src/turn-end-backstop.ts:242`) is untouched by this
diff and still survives `resetEpisode` (`src/turn-end-backstop.ts:181-187`), so a session already at
rest when the process starts is never backstopped — no burst of finish pushes after a deploy.

**Cap semantics.** `maxAttempts`/`attemptCount`/`firedThisEpisode` are fully removed from this
service; grepped the tree and the only remaining hits are `src/build-queue-reminder.ts` (a different
service with its own field of the same name) — no stale reference in `src/turn-end-backstop.ts`,
`src/index.ts`, or the tests. The only constructor callers are `src/index.ts:1901` and
`test/turn-end-backstop.test.ts`, both updated to `maxConsecutiveFailures`.

**Test fixtures typecheck against the real types.** `harness(null)` in the new
"regardless of plan phase" case is valid: `planPhase` is `"planning" | "executing" | null`
(verified against `src/types.ts:107`).

**Suppression paths are parity, not regression.** A recovered push can be silently dropped by
`presence.isActive()` (`src/push.ts:458`), `reducedPushMode` (`src/push.ts:459`, `done` is not in
`REDUCED_ALLOWED` at `src/push.ts:127-133`), the 120s cooldown (`src/push.ts:463`), or a device's
muted `agent` category (`src/push.ts:471`). All four apply identically to the fast path, and
`run()` deliberately treats `notify()`'s `false` return as "fired" — consistent with the plan's
stated decision not to retry an intentionally-suppressed send.

## Stated limitations

- I did not run lint or the test suites (CI owns that, per the review instructions); the trace of
  each new test case above is by hand against the implementation.
- herdr's own `done → idle` transition semantics are asserted by the file header from a live check,
  not by anything in this repo; I took that as given and only checked that the code's handling is
  self-consistent with it.

- `src/turn-end-backstop.ts`'s `isResting()` treats `"archived"` as active, so `markActive` would arm `sawActive` for an archived session; unreachable today because `sweep()` only iterates `store.list({ activeOnly: true })` and `forget()` is called on `session:archived` (`src/index.ts:1976`).
- A spurious one-poll `working` blip after a real `done` edge would clear `delivered` via `markActive` and let a second finish push fire ~120s later once the `pushCooldownMs` window (`src/config.ts:645`, 120s) has also lapsed; this is the pre-existing `markActive` sensitivity the plan-gate and autopilot consumers already carry, not something this diff introduces.

## Plan drift

Minor, and purely informational. The plan's Approach section described the runaway guard as a
single per-pass streak ("reset to 0 on a pass where every attempted dispatch resolved"); the
implementation tracks the streak per consumer (`failStreak: Record<Consumer, number>`,
`src/turn-end-backstop.ts:89`), consistent with the plan's own per-consumer `fired` reasoning but
not with the literal wording. Everything else — the files touched, the helper name, the push-first
ordering, the shared `delivered` stand-down — matches.


**Nits (non-blocking):**

- [bug] src/turn-end-backstop.ts: sweep()'s JSDoc (line 193) still says retries are bounded by the per-session attempt cap (see recover) — that cap no longer exists. The guard is now the per-consumer maxConsecutiveFailures streak; a reader following the comment will grep for a removed field.
- [bug] src/turn-end-backstop.ts: the failStreak doc (lines 79-89) says the streak is reset to 0 by that consumer's next success, but once it reaches maxConsecutiveFailures owes() blocks the very dispatch that could succeed, so the write-off is terminal for the session's lifetime (until the debounce entry is pruned). Worth saying so explicitly so the consecutive framing is not read as self-healing.

PR #2271, opened 2026-09-11
fix(maintain): dedupe block signals per episode, drop block from incident band

### LINKED ISSUE #2242: Dedupe StatusPoller block emissions per episode: incident_spike counts every terminal repaint
> Opened automatically by Shepherd's maintain loop. The diagnosis below is a read-only
> agent's hypothesis — verify it before acting.

## Anomaly

incident_spike:block measured 122 occurrences across 19 sessions in 7 days (tier 2 at 25/5), i.e. ~6.4 block rows per session. The embedded evidence shows that these are not 122 distinct incidents: signals 8-12 are the same ufw question captured five times inside 37 seconds (17:05:19, :22, :32, :53, :56) and signals 4-7 are one four-step AskUserQuestion dialog captured four times inside 25 seconds, each row differing only in how the terminal had repainted. The narrowest gap between rows is 3s, exactly the poller's reclassify cadence.

## Evidence

The emission chain is StatusPoller.maybeClassify (src/poller.ts:1656) -> StatusPoller.emitBlock (src/poller.ts:1489) -> events.emit(session:block) (src/index.ts:1109) -> attachSignalCapture (src/signals.ts:11) -> SessionStore.addSignal (src/store.ts:4487) -> SessionStore.countSignalsByKind (src/store.ts:4524) -> incidentTier (src/maintain-core.ts:373). The duplication is created at src/poller.ts:1708, where maybeClassify computes its dedup key as JSON.stringify(reason) over the whole BlockReason. BlockReason.tail (src/blocked.ts:12-23) is the last 15 non-empty visible lines produced by tailLines (src/blocked.ts:37) and returned by classifyBlocked (src/blocked.ts:78), so any repaint of the same waiting dialog -- a multi-step question advancing its checkbox row, scrolled context, a live counter -- yields a different tail, a different sig, and therefore a fresh emitBlock of an already-announced block, throttled only by reclassifyMs (default 3000, src/poller.ts:434). That matches the 3s minimum spacing in the evidence. attachSignalCapture (src/signals.ts:15-27) writes a row for every non-null block event with no episode key and no payload dedup, and addSignal (src/store.ts:4487) inserts unconditionally with a fresh randomUUID, so the band counts raw rows. The asymmetry is visible inside the poller itself: stall blocks go through fireStall (src/poller.ts:1634), guarded once per episode by the fixed STALL_SIG, and quota blocks by the fixed QUOTA_SIG (src/poller.ts:47-48) -- only the content-signature shapes (menu, yes-no, awaiting-input) re-fire per repaint. Two secondary amplifiers share the same sink: trySuppressSpinner (src/poller.ts:1729) deletes lastSig and emits a clear when it suppresses an awaiting-input fallback, so the re-arm on the next cadence writes a new row rather than being recognised as the same block; and clearBlock (src/poller.ts:1758) drops lastSig on any leave-blocked flap, so herdr's latchy blocked status re-arming writes another row. A second, independent contributor is the band's own scope: INCIDENT_KIND_EXCLUDED (src/maintain-core.ts:53) excludes only reply as high-volume-by-design, while block fires on every ordinary operator question. Every payload in the evidence is a healthy planning dialog (which ruleset should go in, how far should the fix go, fix only the herdr-outdated stub or the baseline too) rather than a fault, so this band is scoring normal interactive use as incidents.

## Affected subsystem

StatusPoller block classification and emission (src/poller.ts maybeClassify / emitBlock) feeding the learnings signal si

## Open questions

- I could not read the live signals table -- Bash access to the state directory was denied -- so the duplicate ratio is inferred from the 12 embedded evidence lines and the 3s reclassify cadence, not measured. Someone should run a GROUP BY sessionId, payload over signals WHERE kind = 'block' AND ts >= now-7d to get the real distinct-episode count.
- If the count deduped to one row per dialog, does it still breach? ~122/6.4 is roughly 19 episodes across 19 sessions, which would clear tier 2 (25/5) but probably still hit tier 1 (10/3) -- so dedup alone may downgrade rather than clear the band. Whether tier 1 on block is a signal anyone wants is a product decision I cannot make from the code.
- Should block join reply in INCIDENT_KIND_EXCLUDED (src/maintain-core.ts:53), or should only some shapes count as incidents (e.g. stall and quota yes, menu and yes-no no)? The band currently treats an agent politely asking a question as an incident class.
- Evidence rows 1 and 3 carry identical payloads 10 minutes apart, and the payload is static Shepherd prompt-instruction text about $BUZZ_BIN and @shepherd replies, not a live dialog. That looks like the awaiting-input fallback firing on a frozen buffer via the trySuppressSpinner freshness gate (src/poller.ts:1729-1742), which would be a distinct false-positive path from the repaint duplication. I could not confirm it without the raw buffers.
- Was hookSignalsActive() on in this install? If so, tryHookAwaitingBlock (src/poller.ts:1131) adds a second entry into maybeClassify per tick and may raise the duplicate rate further; if off, that path was dormant and only the herdr-blocked fallback ran.
- Was SHEPHERD_MAINTAIN_THRESHOLDS set? The 10/3 and 25/5 numbers in the breach header match DEFAULT_BAND_THRESHOLDS (src/maintain-core.ts:132-135), so it looks unset, but mergeThresholds (src/maintain-core.ts:190) is fail-soft and I cannot see the environment to be sure.

---

Band `incident_spike:block` · incident_spike [block]: 122 occurrence(s) across 19 session(s) → tier 2 · measured 2026-09-10T18:46:17.136Z

scoop:
Closes #2242.

`incident_spike:block` was scoring normal interactive use as an incident. Two independent causes, both fixed here.

## What the live data said

I measured a read-only copy of the live DB before planning, because the issue's own first open question asks for it — and the answer reshaped the fix.

| Dedup rule | occurrences | sessions | tier (25/5 = T2) |
| --- | --- | --- | --- |
| none (today) | 181 | 32 | **2** |
| exact `(sessionId, payload)` | 169 | 32 | **2** |
| episode = 30s gap | 78 | 32 | **2** |
| episode = 60s gap | 71 | 32 | **2** |
| episode = 120s gap | 62 | 32 | **2** |
| episode = 300s gap | 52 | 32 | **2** |

Two results drove the design:

- **The issue's proposed measurement under-reports.** `GROUP BY sessionId, payload` removes only 12 of 181 rows — the repaints genuinely *differ*, which is exactly why payload dedup fails. The real signature is time adjacency: 103/181 rows land within 30s of the previous row in the same session, 40 within 5s (the 3s reclassify cadence).
- **Dedup alone cannot clear the band at any plausible window.** Even 5 minutes leaves 52 occurrences across 32 sessions. So the issue's second open question ("does it still breach?") answers *yes*, not even downgraded — a dedup-only fix would have the maintain loop re-file this same issue on its next sweep.

## Changes

**1. Per-episode dedup at the capture layer** (`src/signals.ts`)

A 60s sliding per-`(session, kind)` window. Deliberately **not** in the poller, which is what the issue title proposes: `session:block` also drives the UI block card's tail, the herdr state push and the working-while-blocked flag, all of which want per-repaint accuracy. The `signals` table is the only consumer that over-counts, so that is where the dedup goes.

The window slides on every observed event, so a long repaint stream stays one episode. A `block: null` clear deliberately does **not** reset it — `trySuppressSpinner` and `clearBlock` both drop `lastSig` and re-arm, so clear-keyed episodes would still write a row per flap. Those are the two amplifiers the issue names.

**2. `block` joins `reply` in the incident-band exclusion** (`src/maintain-core.ts`)

An agent asking the operator a question is high-volume by design, not a fault class — every payload in the breach that filed #2242 was a healthy planning dialog. Genuine faults stay covered by `stall`, `critic` and `injection_detected`.

**3. Retire readings a sweep no longer produces** (`src/store.ts`, `src/maintain.ts`)

Not in the issue; found while implementing. `upsertMaintainReading` only ever writes, and the live DB holds `incident_spike:block | tier 2 | value 125`. Excluding the kind stops it being re-evaluated, so that row would freeze at tier 2 forever — and `listMaintainReadings` orders `tier DESC`, pinning it to the top of the Delivery lens band card. **Without this, change 2 is invisible to the operator.** An empty produce-set is a no-op, so a degraded gather can't read as data loss.

## Blast radius

The two other consumers of block-signal counts are unaffected, verified rather than assumed:

- `isGoodOutcome` (learnings reward attribution) tests `blockingSignalCount === 0` — an episode still writes ≥1 row, so zero-ness never flips.
- The auto-trial gate reads `distinctKinds` / `distinctSessions`, which deduping within one session+kind cannot change.

The one real consequence: the distiller's corpus shrinks, so its `minSignals` gate (default 5) trips less often. Judged correct rather than regressive — five repaints of one dialog should not trigger a distill run — and `minSignals` is left alone rather than retuned speculatively.

## Not in scope

- **No `src/poller.ts` change at all.** The amplifiers are neutralised at the capture layer instead.
- **The frozen-buffer false positive** from the issue's fourth open question is **real** — session `1ab3d5c0…` carries three rows of identical *static* Shepherd prompt text about `$BUZZ_BIN`, an awaiting-input fallback firing on a frozen buffer rather than a live dialog. It is a distinct poller classification defect on a path this PR does not touch, so it is filed as #2272.
- Splitting quota-shaped blocks into their own `SignalKind` (offered to the operator, not chosen); retuning `minSignals`; any env knob for the window.

## Verification

`bun run lint`, `bun run test` (9722 pass / 0 fail), `bunx tsc --noEmit`, and the fallow delta gate (`No issues in 8 changed files`) all pass.

Three existing `maintain-core` fixtures used `block` merely as their example kind; they move to a non-excluded kind so they keep asserting the 10/3 and 25/5 tier boundaries they were written for.

### DIFF
--- a/docs-site/src/content/docs/reference/configuration.md
+++ b/docs-site/src/content/docs/reference/configuration.md
@@ -258,7 +258,7 @@ disposable worktree and nothing else. **Tier 3** skips the issue and opens a **p
 | Band | Measures | Window | Tier 1 | Tier 2 |
 | --- | --- | --- | --- | --- |
 | `critic_error_rate` | Share of outcome-bearing review spawns that errored (produced no verdict) | 7 days, min sample 10 | ≥ `0.15` | ≥ `0.30` |
-| `incident_spike` | `signals` per kind — needs **both** an occurrence count and a distinct-session count, so one thrashing task can't trip it. The `reply` kind is excluded (operator corrections are high-volume by design) | 7 days | ≥ 10 occurrences **and** ≥ 3 sessions | ≥ 25 occurrences **and** ≥ 5 sessions |
+| `incident_spike` | `signals` per kind — needs **both** an occurrence count and a distinct-session count, so one thrashing task can't trip it. The `reply` and `block` kinds are excluded — operator corrections and an agent asking the operator a question are both high-volume by design, not fault classes | 7 days | ≥ 10 occurrences **and** ≥ 3 sessions | ≥ 25 occurrences **and** ≥ 5 sessions |
 | `first_pass_collapse` | Per-repo first-pass review rate — direction is **inverted**, a lower rate is worse | 30 days, min sample 8 | ≤ `0.60` | ≤ `0.40` |
 | `dead_code_drift` | Auto-fixable dead-code findings in Shepherd's own checkout (`fallow dead-code`). Point-in-time, no window and no minimum sample. Declares a tier-3 fix class, so its tier-2 breach is **promoted to tier 3** | now | ≥ 1 finding | ≥ 3 findings |
 
--- a/src/maintain-core.ts
+++ b/src/maintain-core.ts
@@ -46,11 +46,18 @@ export const FIRST_PASS_RANGE = "30d" as const;
 const FIRST_PASS_WINDOW_DAYS = 30;
 
 /**
- * `incident_spike` evaluates every SignalKind EXCEPT this one. `reply` is the learnings flywheel's
- * operator-correction stream — high-volume by design and not an incident class, so including it
- * would put the band permanently in breach on a healthy install.
+ * `incident_spike` evaluates every SignalKind EXCEPT these. Both are high-volume BY DESIGN rather
+ * than fault streams, so including them puts the band permanently in breach on a healthy install:
+ *
+ * - `reply` is the learnings flywheel's operator-correction stream.
+ * - `block` is an agent asking the operator a question (#2242). Every payload in the breach that
+ *   filed that issue was a healthy planning dialog — which ruleset to use, how far a fix should go.
+ *   Deduping the repaint noise alone does NOT rescue this band: measured on a live install, one
+ *   row per episode still left 52–78 occurrences across 32 sessions against a 25/5 tier-2
+ *   threshold, so the band would simply re-file. Genuine faults stay covered by `stall`, `critic`
+ *   and `injection_detected`.
  */
-const INCIDENT_KIND_EXCLUDED: SignalKind = "reply";
+const INCIDENT_KINDS_EXCLUDED: ReadonlySet<SignalKind> = new Set<SignalKind>(["reply", "block"]);
 
 /** How long a band is suppressed after a run COMPLETES, whatever its outcome or tier. */
 export const DEFAULT_COOLDOWN_MS = 14 * 24 * 60 * 60 * 1000;
@@ -440,7 +447,7 @@ export function evaluateBands(
 
   // ── incident_spike (global, one row per signal kind) ────────────────────────
   for (const row of input.incidents) {
-    if (row.kind === INCIDENT_KIND_EXCLUDED) continue;
+    if (INCIDENT_KINDS_EXCLUDED.has(row.kind)) continue;
     out.push({
       key: bandKey("incident_spike", row.kind),
       bandId: "incident_spike",
--- a/src/maintain.ts
+++ b/src/maintain.ts
@@ -222,6 +222,7 @@ export type MaintainStore = Pick<
   | "countSignalsByKind"
   | "listSignalsByKind"
   | "upsertMaintainReading"
+  | "pruneMaintainReadings"
   | "listMaintainReadings"
   | "insertMaintainRun"
   | "finishMaintainRun"
@@ -407,6 +408,10 @@ export class MaintainService {
 
     const readings = evaluateBands(await this.gather(now), this.thresholds, now);
     for (const r of readings) this.deps.store.upsertMaintainReading(r);
+    // Retire readings this sweep did not produce, so a band that stopped being evaluated (an
+    // excluded signal kind, a repo that no longer exists) can't sit on the Delivery lens forever
+    // at its last tier. See `pruneMaintainReadings` for why an empty set is a no-op.
+    this.deps.store.pruneMaintainReadings(readings.map((r) => r.key));
 
     // Tier 1 IS this log line plus the persisted reading — deliberately not a `signals` row.
     const breached = breaches(readings);
--- a/src/signals.ts
+++ b/src/signals.ts
@@ -1,27 +1,71 @@
 import type { EventHub } from "./events";
 import type { SessionStore } from "./store";
 import type { BlockReason } from "./blocked";
+import type { SignalKind } from "./types";
+
+/**
+ * How long after an observed block event the same session+kind stays inside one EPISODE (#2242).
+ *
+ * The poller re-emits `session:block` on every terminal repaint of a waiting dialog: its dedup key
+ * is `JSON.stringify(reason)` and `reason.tail` is the last 15 visible lines, so a scrolled
+ * context, an advancing checkbox row or a live counter yields a different signature and a fresh
+ * emit every `reclassifyMs` (default 3s). Measured on a live install: 181 `block` rows across 32
+ * sessions in 7 days, of which 103 landed within 30s of the previous row in the same session and
+ * 40 within 5s. Deduping on the payload is near-useless (181 → 169) precisely because the repaints
+ * DIFFER; time adjacency is the signature that separates repaints from distinct episodes.
+ */
+const EPISODE_DEBOUNCE_MS = 60_000;
 
 /**
  * Capture `block` and `stall` learning signals off the `session:block` event.
  * Reply signals are captured in SessionService.reply; critic signals in ReviewService.
  * A `stall`-shaped block becomes a "stall" signal; every other shape a "block" signal.
  * Cleared blocks (block: null) and unknown sessions are ignored.
+ *
+ * DEDUPED PER EPISODE (#2242): repeated block events for the same session+kind collapse into one
+ * signal row. The dedup lives HERE and not in the poller on purpose — `session:block` also drives
+ * the UI block card's tail, the herdr state push and the working-while-blocked flag, all of which
+ * want per-repaint accuracy; the `signals` table is the only consumer that over-counts. Keyed by
+ * kind as well as session so a `stall` and a `block` in the same session never share a window
+ * (`stall` is already once-per-episode via the poller's fixed `STALL_SIG`, so the window is inert
+ * there — it is applied uniformly rather than special-cased).
+ *
+ * The window SLIDES: every observed event re-stamps the key, so an episode that keeps repainting
+ * stays one episode however long it runs, and a new row needs a real `debounceMs` lull in the
+ * event stream. A `block: null` clear deliberately does NOT reset it — the poller drops `lastSig`
+ * in both `trySuppressSpinner` and `clearBlock`, so an episode keyed on clear events would still
+ * write a fresh row on every spinner-suppression or leave-blocked flap, which is the amplifier
+ * this is here to neutralise.
  */
 export function attachSignalCapture(
   events: Pick<EventHub, "subscribe">,
   store: Pick<SessionStore, "get" | "addSignal">,
+  opts?: { now?: () => number; debounceMs?: number },
 ): () => void {
+  const now = opts?.now ?? (() => Date.now());
+  const debounceMs = opts?.debounceMs ?? EPISODE_DEBOUNCE_MS;
+  /** session+kind → when that key was last OBSERVED (not last written). */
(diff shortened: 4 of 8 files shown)


scoop (commented):
# Review — dedupe block signals per episode, drop `block` from the incident band

Diff reviewed: `67ce77dbb...HEAD` (1 commit, 8 files). Read-only inspection; no build, no tests run.

## Verdict

No blocking defect found. The three moving parts each do what they claim, the identifiers resolve,
and the one band-scored kind the debounce could have quietly weakened (`stall`) is genuinely inert
under a 60s window. A few non-blocking observations below.

**1. The capture-layer debounce — `src/signals.ts:40-72`**

- `SignalKind` is imported from `./types` and is a real exported union containing both `"block"` and
  `"stall"` — verified against `src/types.ts:1382-1392`. The `kind` hoist at `src/signals.ts:55`
  produces exactly the same value the pre-diff inline expression did (`src/signals.ts:68` vs. the
  removed `kind: block.shape === "stall" ? "stall" : "block"`).
- Production wiring is unchanged: `attachSignalCapture(events, store)` at `src/index.ts:2376` passes
  no `opts`, so `now` falls back to `Date.now` and `debounceMs` to `EPISODE_DEBOUNCE_MS`
  (`src/signals.ts:45-46`). It is the only production call site (grep across the tree; the other
  hits are `test/signals-attach.test.ts:20` and a historical plan doc under `docs/superpowers/`).
- Boundary semantics are self-consistent: the prune loop at `src/signals.ts:59` uses
  `t - at >= debounceMs`, and it deletes the current key *before* the `has` check at
  `src/signals.ts:62`, so an event exactly `debounceMs` after the previous one opens a new episode.
  That matches the assertion in `test/signals-attach.test.ts:100` (`t += 60_000` → 2 rows). Deleting
  from a `Map` while iterating it is well-defined in JS, so the in-loop `delete` is safe.
- Map growth is bounded by the keys observed since the last event, and every event sweeps the stale
  ones — asserted by `test/signals-attach.test.ts:150-161`.
- **The `stall` claim checks out.** The doc comment (`src/signals.ts:29-31`) asserts the window is
  inert for `stall`. `fireStall` is guarded once per episode by `STALL_SIG` (`src/poller.ts:1717`),
  and a stall requires `cfg.stallMs` of quiet — 8 minutes, verified at `src/stall.ts:24` and used at
  `src/session-liveness.ts:175`. A stall re-arm inside a 60s window is therefore not reachable, so
  the debounce does not suppress rows for a kind that *is* still scored by `incident_spike`. This
  was the main thing that could have made the debounce a real regression; it does not.
- **The poller premise the comment rests on is accurate**: `maybeClassify` emits only on signature
  change (`src/poller.ts:1791-1796`, `if (sig === this.lastSig.get(id)) return true`), and both
  `trySuppressSpinner` (`src/poller.ts:1816-1818`) and `clearBlock` (`src/poller.ts:1845-1847`) drop
  `lastSig` and emit a clear — which is exactly why keying the episode on a clear would not have
  worked, as the comment at `src/signals.ts:35-38` says.

**2. The band exclusion — `src/maintain-core.ts:60`, `src/maintain-core.ts:450`**

- `INCIDENT_KIND_EXCLUDED` had no other referent: after the rename, `INCIDENT_KINDS_EXCLUDED` is
  matched by exactly two sites (`src/maintain-core.ts:60` and `:450`) and the old name appears
  nowhere in the tree (grep `INCIDENT_KIND`). It was module-private, so no consumer breaks.
- `evaluateBands` has a single caller, `src/maintain.ts:409` (grep across `src/` and `ui/src`), so
  the exclusion cannot leak into another code path.
- No remaining fixture depends on an `incident_spike:block` reading existing: the only
  `incident_spike:block` strings left in the tree are the two new negative assertions at
  `test/maintain-core.test.ts:143` and `test/maintain.test.ts:310,322`. Every other
  `incident_spike:*` fixture uses `stall` (`test/untrusted.test.ts:129`,
  `ui/src/lib/components/usage/DeliveryLens.browser.test.ts:204`, `test/store.test.ts:2430`).
- Doc parity: the band table is the only place the exclusion is documented — grep for
  `critic_error_rate` / `incident_spike` across `README.md`, `docs/`, `docs-site/`, `DESIGN.md` and
  `PRODUCT.md` returns only `docs-site/src/content/docs/reference/configuration.md:260-261`, which
  this diff updates. No second catalog to keep in sync (this repo's message catalogs are under
  `ui/messages`, and no message key changes here — `DeliveryLens.svelte:135-136` renders the kind
  from `r.subject`, so no per-kind key exists to go stale).

**3. The reading prune — `src/store.ts:4632-4638`, `src/maintain.ts:414`**

- `this.db.run(sql, params).changes` is the established API in this file — same shape at
  `src/store.ts:1853` (`DELETE FROM access_tokens WHERE id = ?`). `maintain_readings.bandKey` is the
  PRIMARY KEY (`src/store.ts:1531-1534`), so `NOT IN (...)` over it is exact.
- `pruneMaintainReadings` is added to the `MaintainStore` pick list at `src/maintain.ts:223`; the
  only production value passed as `store` is the real `SessionStore` (`src/index.ts` maintain deps),
  and `test/maintain.test.ts:102` seeds a real `new SessionStore(":memory:")`, so no structural
  implementor is left missing the method.
- Parameter count is safe: one placeholder per produced band key — 1 critic + ≤10 signal kinds
  (`src/types.ts:1382-1392`) + one per repo + 1 dead-code — far under SQLite's variable limit.
- Ordering is right: the prune runs after the upsert loop and before `breaches(readings)`
  (`src/maintain.ts:410-417`), and `breaches` reads the in-memory array, so the delete cannot
  affect what the sweep acts on.
- Cooldowns are unaffected: suppression anchors on `maintain_runs`, not on readings
  (`src/maintain.ts:19-22`), so retiring a reading does not un-suppress anything.

## Limitations

- I did not run `bun run lint`, `bun run test` or `bunx tsc --noEmit` — CI owns those, per the review
  brief. Claims above are from reading the code, not from a green run.
- The live-install measurements quoted in the new comments (181 rows / 32 sessions / 103 within 30s,
  and the 52–78 post-dedup figures) come from a database that is not in this repo. I could not
  verify them and am not asserting they are right; nothing in the diff's correctness depends on the
  exact numbers.

- `src/store.ts:4633` — the `keepKeys.length === 0` early return guards a state that cannot occur: `evaluateBands` unconditionally pushes `critic_error_rate` (`src/maintain-core.ts:435`) and `dead_code_drift` (`src/maintain-core.ts:489`), so the sweep's key list is never empty. Defensible as a contract for a future caller, and the comment says so, but it is error handling for an impossible case on the only call path that exists.
- `src/store.ts:4632` — the `number` return ("so a caller can log the retirement") is consumed only by `test/maintain.test.ts:336`; `src/maintain.ts:414` discards it and logs nothing.

- `src/signals.ts:45-46` — the `opts` seam is test-only today. `debounceMs: 0` degrades safely (every entry prunes, every event writes), but a negative `debounceMs` or a `now()` that moves backwards would make `t - at >= debounceMs` never true and pin an episode open indefinitely. Unreachable from the single production call site, which passes nothing.

- `src/store.ts:1527-1529` — the `maintain_readings` schema comment still says readings are "bounded by the band count, so they need no retention sweep", which reads oddly now that a retention sweep exists; the sentence is in the untouched DDL block.
- `src/poller.ts` — the frozen-buffer false positive the originating issue raises (static prompt text re-emitted as an `awaiting-input` block minutes apart) is untouched, as the plan's Out of Scope says. The debounce hides its 3s-cadence repeats but not a repeat 10 minutes later.


**Nits (non-blocking):**

- [bug] src/signals.ts: the episode window slides on every observed event (line 63 re-stamps the key even when the row is dropped), so a dialog that repaints for minutes keeps its window open the whole time and a genuinely new question shortly after it ends is also swallowed. The approved plan specified last-CAPTURE, which would have re-opened after 60s from the first row. Documented and deliberate, and only the distiller corpus consumes block rows now that the band excludes the kind — but worth a second look if learnings recall matters.
- [bug] src/signals.ts:58: the comment claims the map only ever holds blocked sessions. Entries are pruned only when a later block event arrives, so after a quiet period the map still holds every key observed since the last event. Harmless, but the cost claim is looser than stated.
- [bug] test/maintain.test.ts:326: the test named keeps every reading when a sweep produces none never makes a sweep produce none — it calls pruneMaintainReadings([]) directly. The name promises coverage of a sweep path the body does not exercise.

PR #2273, opened 2026-09-11
feat(mobile): rework the phone list for thumb reach and touch conformance

kai-osthoff:
The phone list screen spent **237px of a 932px viewport** on chrome before the first card, and six of its controls sat below the iOS HIG 44×44 floor — five of those below the hard **WCAG 2.5.8 (AA)** floor of 24×24 as well.

Both complaints turned out to have the same root cause: where the 44px floor *was* met it was met by **inflating boxes** (a 44px repo chip carrying a 10px label), and where it would have cost space it was not met at all. The chrome was simultaneously too tall and too thinly hit.

Design handoff, with the full before/after measurements and the rejected alternatives: `docs/design/mobile-herd/README.md`.

## Touch conformance

| Control | Before | After |
| --- | --- | --- |
| `.ctally` (status tallies) | 44 × **24** | 44 × 44 from a 360px viewport up |
| `PrBadge` / `PlanGateBadge` / `CriticBadge` / `BuildQueueBadge` / `.preview-badge` | **~15px** tall, tappable | read-only readouts on touch |
| `.name-icon.actionable` (repo emoji) | **18 × 19** | not a tap target on touch |
| `.hold-cta` (Go / Re-review / Resume) | **~17px** tall | 44 × 44 on touch |

The five card badges are **relocated, not removed**: the card tap opens the detail screen, where `GitRail` / `PlanGateBadge` / `BuildQueuePanel` / the preview tab carry the same controls at a conformant size. Five 44px targets stacked in one card would have pushed it past 200px — the opposite of what this screen needs.

`.hold-cta` was **not** in my original audit. The new sweep found it, which is the argument for the sweep being generic.

## Thumb zone

- Lens segments move into the bottom bar, which now has two ranks. Its geometry lives in `--mobile-actionbar-h`, so the list's `padding-bottom` reservation and the toast inset cannot drift from it.
- The Done lens loses its segment and gains a gear-menu entry — that is what frees the width for REPOS and New task to share **one** bar instead of needing two.
- The repo rail leaves the phone chrome for a new `ReposSheet`, opened by REPOS: repo filter **plus** the backlog entry, so the button finally means what its label says.
- The 2×10px chrome gaps go (this system separates surfaces with hairlines, not with space — `DESIGN.md`, Elevation), and the phone TopBar drops to 5px vertical padding since every control in that row already carries its own 44px floor.
- Card diet: one prompt line, 9px padding, badge stack capped at two, clock on the micro rung so the badge rail stops out-measuring the content column beside it.

**Result: 237px → 113px above, 90px → 138px below. The list gains ~76px and shows ~6.7 cards where it showed ~5.1.**

## Proof, not assertion

`ui/src/lib/components/touch-targets.browser.test.ts` runs in the **`browser-touch`** vitest project — the only one with `hasTouch`, without which `(pointer: coarse)` never matches and every rule here would be silently absent.

The sweep measures **whatever renders** rather than a curated selector list, so a newly added control cannot slip past it. Every exception must be named in the test with its reason, which makes the exception list itself the audit trail. Two survive, both bounded and justified in place:

- `.stepper` — WCAG 2.5.8 **Equivalent** exception: its click is `onselect(session.id)`, exactly what the ≥44px card hit-target does.
- `.desig-btn` — WCAG 2.5.8 **Inline** exception: the target sits in the footer sentence, sized by its line-height. Its menu exists nowhere else, so relocating would remove rather than move; growing it would add ~28px to every card.
- `.ctally` below 360px — the documented fold-cover compromise, **measured** at a 320px viewport rather than asserted in prose.

## Honest deviations from the design sheet

| Sheet said | Shipped | Why |
| --- | --- | --- |
| **7,0 cards** | **~6,7** | The 7,0 came from an estimated 97px card and forgot the 2px inter-card margin. Measured: 99.5px card, ~101.5px pitch. The last 4px would only come from shaving touch padding — that buys the round number and spends the ergonomics. |
| Overflow badge as "+1" | "…" left of the stack | The exact count needs JS counting across five components with their own render conditions. `:has(> :nth-last-child(3))::before` does it in CSS, and absolute positioning keeps it off a third flex row that would give back the height the cap just won. |
| Clock in the card footer | Clock stays in the rail, on the micro rung | A move would have threaded `flow` through `Herd` → `HerdGroup` → `UnitRow` for one position. One CSS line reaches the same goal. |

## Worth a reviewer's eye

- **`+page.svelte`'s template crosses the Tier-1 complexity bar again** (#855 had it at 32/45; it is 43/52 now) because of two branches: the repo rail becoming desktop-only, and the sheet mount. The sheet mount belongs in `page/AppOverlays` with the other overlays — grandfathered with that note rather than done here, since threading `repoChips`/`repoFilter` through the dispatcher is its own change.
- **`docs/design/**` is excluded from fallow** — the artboards' mandatory `<script src="./support.js">` line is replaced by an inline runtime at render time and never resolves.
- Two pitfalls are written into the handoff for whoever touches this next: the badges belong to **child** components (a Svelte-scoped `:nth-last-child` selector matches nothing), and the z-index rule that lifts badges above `.unit-hit` is a **descendant** selector for exactly that reason.

## Verification

`ui`: `bun run check` ✓ · `bun run check:i18n` ✓ (parity) · `bun run test` ✓ 316 files / 5153 tests
root: `bun run lint` ✓ · `bun run test` ✓ 9712 tests · `fallow audit` ✓

Manual: worth a look at 320px (fold cover) and with iOS Text Size raised, where `--ui-scale` reaches 1.5 and the chrome heights stay `min-height` by design.

### DIFF
--- a/.fallowrc.jsonc
+++ b/.fallowrc.jsonc
@@ -18,7 +18,13 @@
   // Generated codegen output (gen-herdr-types.ts, from the vendored herdr
   // protocol schema): its ~170 exports are intentionally emitted for the whole
   // protocol surface, not dead code — exclude from all fallow analyses.
-  "ignorePatterns": ["src/generated/**"],
+  //
+  // docs/design/** holds design HANDOFF material, not source: the `*.dc.html` artboards under
+  // docs/design/mobile-herd are Design-Component files whose mandatory
+  // `<script src="./support.js">` head line is replaced by an inline runtime at render time — the
+  // file is never resolved, so fallow reports one unresolved import per artboard. Reference
+  // material for humans, analysed by nothing.
+  "ignorePatterns": ["src/generated/**", "docs/design/**"],
 
   "entry": [
     "src/index.ts",
@@ -398,6 +404,22 @@
         "maxCognitive": 64,
         "maxCrap": 20000,
         "reason": "Inline hold-reason subline + one-click CTA (#1561) adds ~+3 cognitive to the already-large row template (58 -> 61). The CTA block was extracted into the holdSubline snippet, which recovered cyclomatic back under the Tier-1 40 bar; the residual +1 over the 60 cognitive bar is genuine new-affordance cost. The cold-resume chip (#2042) adds one more independent `{#if}` in the meta row (62 -> 64), cyclomatic unchanged at 39: the whole point of that feature is to warn on the surface where the operator CHOOSES a session, so the branch cannot live anywhere else. Keep capped until UnitRow is split into per-zone sub-components (#855), like the other large row templates."
+      },
+      {
+        "files": ["ui/src/lib/components/unit-row/UnitRowRight.svelte"],
+        "functions": ["<template>"],
+        "maxCyclomatic": 45,
+        "maxCognitive": 60,
+        "maxCrap": 20000,
+        "reason": "The phone-list rework (docs/design/mobile-herd) adds two branches to the badge rail: the D4 read-only twin of the preview badge, and the .u-badges wrapper the touch cap needs to reach the badges without also catching the clock. That takes cyclomatic 39 -> 41, one over the Tier-1 40 bar; cognitive stays well under (53). The template is a flat one-{#if}-per-badge dispatcher, the same shape as AppOverlays below — each branch is independent and idiomatic. Keep capped until the rail splits into passive/actionable child components (#855)."
+      },
+      {
+        "files": ["ui/src/routes/+page.svelte"],
+        "functions": ["<template>"],
+        "maxCyclomatic": 45,
+        "maxCognitive": 60,
+        "maxCrap": 20000,
+        "reason": "The phone-list rework (docs/design/mobile-herd) adds two branches to the page shell: the repo rail becomes desktop-only chrome ({#if !mobile.current}) and the REPOS sheet mounts ({#if showRepos}). That takes the template just over the Tier-1 40 bar; cognitive stays under (52). Note this partially gives back #855's 78/96 -> 32/45 win, so the sheet mount belongs in page/AppOverlays with the other overlays — tracked in #855 rather than done here, because threading repoChips/repoFilter through the dispatcher is its own change."
       },
       {
         "files": ["ui/src/lib/components/page/AppOverlays.svelte"],
--- a/docs/design/mobile-herd/B3RepoRail.dc.html
+++ b/docs/design/mobile-herd/B3RepoRail.dc.html
@@ -1,0 +1,888 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset="utf-8" />
+    <script src="./support.js"></script>
+  </head>
+  <body>
+    <x-dc>
+      <helmet>
+        <link
+          rel="stylesheet"
+          href="https://fonts.googleapis.com/css2?family=JetBrains+Mono:wght@400;500;600;700;800&display=swap"
+        />
+        <style>
+          body {
+            margin: 0;
+          }
+          a {
+            color: #2f6fb0;
+          }
+          a:hover {
+            color: #1c5590;
+          }
+          .sheet {
+            --bg: #e7ebe9;
+            --glow: #f3f6f4;
+            --panel: #f7f9f8;
+            --panel-2: #eef1ef;
+            --head: #eef2f0;
+            --inset: #f3f6f4;
+            --hover: #e4e9e6;
+            --sel: #d2dbd6;
+            --line: #d2dad6;
+            --line-bright: #b9c4bf;
+            --ink: #2b3633;
+            --ink-bright: #131a18;
+            --muted: #5a6862;
+            --faint: #93a09a;
+            --amber: #a8680a;
+            --green: #1d8a5d;
+            --red: #c2363b;
+            --blue: #2f6fb0;
+            --slate: #6b7873;
+            --wash-attention: oklch(0.965 0.016 23);
+            --fs-micro: 10px;
+            --fs-meta: 11px;
+            --fs-base: 13px;
+            --fs-lg: 16px;
+            font-family: "JetBrains Mono", ui-monospace, SFMono-Regular, Menlo, monospace;
+            letter-spacing: 0.02em;
+          }
+          .phone {
+            background:
+              radial-gradient(120% 100% at 50% -10%, var(--glow) 0%, var(--bg) 60%), var(--bg);
+          }
+          .name {
+            color: var(--ink-bright);
+            font-weight: 500;
+            letter-spacing: 0.04em;
+            font-size: var(--fs-base);
+            overflow: hidden;
+            white-space: nowrap;
+            text-overflow: ellipsis;
+          }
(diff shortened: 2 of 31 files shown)


kai-osthoff:
🟢✨🎉 9/9 checks green, 0 conflicts, 0 reviews 🥲 @scoop 👀🙏🚀 no pressure 💅

scoop:
## Merge-train review — held from the train

Reviewed as part of a merge train alongside #2269. CI is green, branch hygiene is clean, i18n parity
holds (en/de both 3988 keys, zero one-sided), `sinceVersion: 1.48.0` correctly targets the
unreleased cohort, and the `docs/design/**` fallow exclusion is narrow and holds no source. The
design work and its evidence are genuinely strong — the measurements, the named exceptions and the
honest "deviations from the sheet" table are all above the usual bar.

Holding it out of the train for three blockers plus two judgement calls.

### 1. `ReposSheet.clearFilter()` does not clear a multi-repo filter

`ReposSheet.svelte:42` iterates `repoFilter` while each `onrepofilter(p, false)` call mutates that
same `SvelteSet`:

- `applyRepoFilter` → `replaceRepoFilter(nextRepoFilter(repoFilter, p, false))` (`+page.svelte:379-380`)
- `replaceRepoFilter` deletes and adds **in place** on `repoFilter` (`+page.svelte:373-377`)
- `nextRepoFilter` only returns an empty set when `current.size === 1` (`queue-strip.ts:100`)

With `{A, B}` filtered, the first iteration collapses the set to `{A}`; `B` is deleted mid-iteration
so the loop never visits it. **"All repos" leaves one repo filtered.** The single-repo case works,
so this only shows up after a Shift-click. Collecting first (`for (const p of [...repoFilter])`)
still would not clear it — `nextRepoFilter(..., false)` on a 2-set returns a 1-set. The sheet
probably wants to call `replaceRepoFilter(new Set())` directly rather than replay per-repo toggles.

### 2. The rebase onto #2278 invalidates one of this PR's premises

The textual conflict is one hunk — `<IssueBadge {session} />` here vs `<IssueBadge {session} {git} />`
on main. Trivial. The semantic consequence is not:

#2278 turned `IssueBadge` into an interactive `<a href>` issue peek. It renders inside the new
`.u-badges` wrapper, and the cap keeps the **last two** children
(`:global(.units.flow .u-badges > :nth-last-child(n + 3)) { display: none }`). Per this PR's own
stated order — agent, research, terminal, **issue**, then PR, critic, queue, plan gate, status —
`IssueBadge` is 4th of nine, so on any loaded phone card the issue peek that shipped yesterday
becomes `display: none`.

The sweep cannot catch this: its only fixture is a fully-loaded row where that chip has zero client
rects and is filtered out of the audit, so the ~15px touch target it presents on sparse cards passes
the gate silently. Worth re-deciding the cap's ordering against a badge that now carries a real
action rather than a readout.

### 3. The sheet is not registered as an overlay

`showRepos` is missing from `anyOverlayOpen()`, and the sheet matches neither `.overlay` nor
`.drawer` in the fallback selector, so global shortcuts and the settings chord stay live underneath
an `aria-modal="true"` surface. There is also no focus trap and no focus restore on close.

### Judgement calls (not blockers, but worth a deliberate answer)

- **The two-badge cap can hide attention state.** Keeping the last two can drop the critic verdict
  or plan-gate badge in favour of a sandbox/identity badge, behind a `pointer-events: none` "…" with
  no way to reveal it. On the surface whose stated job is surfacing "what needs a human now", that is
  the one part of the design worth pushing back on — consider ranking by attention value rather than
  by document order.
- **The sweep is generic within its fixtures, but the fixtures are curated.** Only `UnitRow` and
  `TopBarTallies` are rendered — not `ActionBar`, `HerdSegRow` or `ReposSheet`, the surfaces this PR
  builds. "A newly added control cannot slip past it" is a stronger claim than the test currently
  supports.

### One more for the sweep, whenever it is revisited

`INTERACTIVE` matches only `button` / `[role=button]` / `a[href]` / `input` / `select` / `textarea` /
`[tabindex]`. A plain `<span>` raised above `.unit-hit` with a `statusTip` is invisible to the audit
while still stealing a tap from the card's 44px hit target — which is exactly what #2269's ~17px
`.chip-cold-resume` now does. Widening the matcher would have caught it. (Related: #2283.)

Nothing here is a reason to abandon the work — it is worth landing once 1–3 are addressed.

kai-osthoff:
Thanks — all three blockers were real; I checked each against the code rather than taking them on trust, and all three reproduce as described. Rebased onto `main` (the `IssueBadge {git}` hunk) and pushed as a separate commit so you only have to read the delta.

**1. `clearFilter()`** — confirmed exactly as you traced it. `nextRepoFilter` returns the empty set only for a one-element selection and otherwise collapses to `{p}`, and `replaceRepoFilter` mutates the set being iterated, so `{A, B}` drops `B` mid-loop and leaves `A` filtered. The sheet now takes `onclearfilter` and the page clears the set directly — replaying per-repo toggles was the wrong primitive, as you said. The single-repo case working is what hid it.

**2. The cap — removed, not reordered.** You were right about the ordering, and more right than the framing suggests: `AutopilotBadge` and the sandbox chips sit *after* `CriticBadge` and `PlanGateBadge`, so "keep the last two" drops the critic verdict and the plan gate in favour of a sandbox chip. There is no ordering that makes a fixed count safe, because more than two attention badges can be present at once. It bought ~6px; the failure mode is hiding blocked state on the screen whose job is showing blocked state. Gone. A badge-heavy row is now taller than a quiet one — honest, the rail is showing more.

That also disposes of the `IssueBadge` regression, and I gave it `interactive` on top: #2278 made it an `<a href>`, which turned it into a ~15px target over the card's 44px hit area and made this PR's own D4 comment ("issue … already a readout") false.

**3. Overlay registration** — confirmed. `showRepos` added to `anyOverlayOpen()`, and the hand-rolled Escape handling replaced with `use:dialog` from `a11yDialog.ts`, which is the house contract for trap + Escape + focus restore. I should have used it the first time.

**On the fixtures:** you're right that "a newly added control cannot slip past it" was a stronger claim than the test supported. The sweep now renders every surface this screen is built from — card, header tallies, the bottom bar with its lens segments, and the sheet — and the claim is narrowed to "generic within its fixtures; a new *surface* still needs a new fixture". Each fixture also states a minimum target count now, because one that renders nothing would otherwise pass an "everything is big enough" assertion in silence.

**Density claim corrected too:** it now asserts both ends it claims — a quiet row (~99.5px, ~6.7 cards) and a badge-heavy one (~120px, ~5.6) — rather than one number standing in for an average nobody can check.

**Not addressed, deliberately:** widening `INTERACTIVE` to catch raised non-button spans. It would fail on #2269's `.chip-cold-resume`, which is `main`'s code and not this PR's to fix — #2283 is the right home. Happy to take it there.

Gates after the rebase: `ui` check ✓, i18n parity ✓ (3993 each), 318 files / 5192 tests ✓, root lint ✓, 9749 tests ✓, `fallow audit` ✓.

PR #2280, opened 2026-09-12
fix(poller): don't raise awaiting-input on a queued-input pane (#2272)

### LINKED ISSUE #2272: StatusPoller raises awaiting-input blocks on a frozen idle buffer
Split out of #2242 (fixed in #2271), whose fourth open question asked whether a distinct false-positive path existed alongside the repaint duplication. It does — I confirmed it in the live data while implementing, but it sits on a code path #2271 deliberately does not touch.

## What happens

`StatusPoller.maybeClassify` classifies a session as an `awaiting-input` block on a buffer that is **frozen and stale** — the visible terminal still holds static text, and no dialog is actually waiting. The operator gets a "needs you" card for a session that is not asking anything.

## Evidence

In the live DB, session `1ab3d5c0-0768-4c18-af98-ea88874c2dd7` carries three `block` signal rows with **byte-identical** payloads, spread ~10 minutes apart. The payload is not a dialog at all: it is static Shepherd prompt-instruction text about `$BUZZ_BIN` and `@shepherd` replies — boilerplate sitting in the scrollback, not a question.

That identical-payload-minutes-apart shape is the opposite of the repaint duplication #2242 was about (there, near-identical rows seconds apart with *differing* tails), which is why the episode dedup in #2271 does not address it. Note the dedup does reduce its blast radius — repeats inside the 60s window now collapse — but a re-fire 10 minutes later still writes a fresh row and still raises a fresh block.

## Suspected mechanism

`trySuppressSpinner` (`src/poller.ts`) suppresses an `awaiting-input` fallback only while the buffer is FRESH; a frozen buffer deliberately falls through to the normal emit so a genuinely wedged turn re-arms. That fall-through looks like the path being taken here — but the buffer is frozen because the *session is idle and its scrollback is static*, not because a turn is wedged. The freshness gate cannot currently tell those apart.

Unverified: I did not capture the raw buffers, so treat the mechanism as a hypothesis and confirm before changing the gate.

## Why it matters

Ordinary `block` signals no longer feed `incident_spike` after #2271, so this no longer inflates a band. What remains is the operator-facing cost: a false "needs you" on an idle session, and a junk row in the learnings corpus that the distiller treats as real friction.

## Worth deciding

Whether the fix belongs in the freshness gate (distinguish "frozen because wedged" from "frozen because idle" — e.g. require corroborating evidence of an active turn before the fallback re-arms) or earlier, in `classifyBlocked`, by refusing to classify a buffer with no dialog affordance as `awaiting-input` at all.

scoop:
Closes #2272.

`StatusPoller` raised an `awaiting-input` "needs you" block on a pane that was not asking anything: Claude Code's ordinary at-rest input box with the operator's reply still queued.

## What the live data said

I measured a read-only copy of the live DB (4217 `block` signal rows) before planning, because the issue flags its own mechanism as unverified — and it is wrong.

**The freshness gate is not involved.** Of **1432** rows that `classifyBlocked` shapes as `awaiting-input`, **0** carry an active spinner. `trySuppressSpinner` — and the freshness gate the issue suspects — is never reached on this path. The emit comes straight from `classifyBlocked`'s no-evidence fallback, which requires no dialog affordance at all.

**The real shape.** Reconstructing the three byte-identical payloads the issue cites: the pane is the at-rest input box with stale scrollback above it and `❯ Press up to edit queued messages`. Of the 45 rows that captured a full screen, **33** have that shape — no spinner, no dialog, every one from the last 10 days. Each lands 3–5 s after an answered dialog, i.e. herdr's documented `blocked` latch, in the window before the spinner paints.

## The fix

Suppress the `awaiting-input` fallback when the pane carries queued operator input. That is positive, content-independent evidence: the agent has input it has not consumed, so it cannot be waiting for more. No semantic reading of the scrollback — separating "the last thing on screen is a question to me" from "the last thing on screen is old boilerplate" is not a regex problem.

- `hasQueuedInput()` (`src/blocked.ts`) scans the same last-15-non-empty-lines window as `hasActiveSpinner`, so a scrollback mention cannot forge it. Covers all three wordings Claude Code 2.1.266 ships.
- Suppression mechanics reuse the spinner path (clear any announced block once, raise the working-while-blocked flag once), which `displayStatus()` already renders as `running` rather than "needs you".
- **No freshness gate** — that gate is exactly what let this re-fire ten minutes later. Queued input is an observed state, not a liveness inference.
- **A detected `authUrl` vetoes the suppression.** An MCP OAuth / `/login` prompt renders *above* the at-rest input box and also classifies as `awaiting-input`, so it can share a pane with queued input; `maybeAuthAtRest` only covers idle/done, so a suppressed auth block would have no re-arm path. Corpus overlap is 0 (35 queued-hint rows, 36 auth-prompt rows, none both), but the failure mode is severe and the guard is an ordering change, so it ships anyway. Ordered so the auth read still never runs on a spinner-suppressed tick — no new I/O on the single loop.
- `menu` / `yes-no` dialogs always surface, queued input or not — unchanged.

`maybeClassify`'s `awaiting-input` half moved into `suppressAwaitingInput()` to keep the function under the fallow cognitive-complexity gate.

## Rejected on evidence

The issue's other suggestion — refuse `awaiting-input` without a dialog affordance — would suppress 257/1432 rows, and reviewing that set found genuine blocks it swallows: a real MCP OAuth prompt and an AskUserQuestion wizard bar (`←  ☐ Deliverable … ✔ Submit  →`).

## Accepted trade-off

The stall detector only runs via `maybeProbe` on `status === "running"`, so a session that genuinely wedges while herdr holds `blocked` *and* queued input is on screen surfaces nothing until herdr flips. Every observed instance resolved in seconds; the alternative (a frozen-dwell escape hatch) adds tunable state for a case the corpus never shows.

## Tests

`test/fixtures/fullscreen/queued-messages.txt` is the real captured pane from the issue's session, asserted against rather than a hand-written approximation. New cases cover: suppression, the frozen buffer never re-arming across cadences (#2272's exact re-fire), clearing an announced block exactly once when the dialog is answered, `menu` still surfacing, re-arming once the hint leaves, and an auth prompt winning over queued input.

`bun run lint`, `bun run test` (9749 pass) and `bunx fallow audit --base origin/main --fail-on-issues` all pass.

## Follow-ups, not fixed here

Two adjacent false positives found while measuring, filed separately: #2281 (prose numbered lists emitting false `menu` blocks while a turn is live) and #2282 (`SPINNER_RE` missing real working frames).

### DIFF
--- a/src/blocked.ts
+++ b/src/blocked.ts
@@ -74,6 +74,31 @@ export function hasActiveSpinner(text: string): boolean {
   return tailLines(text).some((l) => SPINNER_RE.test(l));
 }
 
+// The at-rest input box's queued-input hint. Claude Code 2.1.266 ships three
+// wordings — "Press up to edit queued messages", "Press up to select a queued
+// message" and "…select a queued message to edit" — so the match is anchored on
+// the stable "press up to <verb> … queued message" spine rather than one string.
+const QUEUED_INPUT_RE = /press up to \w+\s+(?:a\s+)?queued message/i;
+
+/**
+ * True when the terminal tail shows the agent's at-rest input box carrying
+ * QUEUED operator input. A session holding input it has not consumed yet cannot
+ * be waiting for more, so an `awaiting-input` fallback over such a buffer is a
+ * false "needs you" (issue #2272). Scans the same last-15-non-empty-lines window
+ * as `classifyBlocked` (the hint sits just above the footer), so the phrase
+ * appearing in older scrollback cannot forge it.
+ *
+ * Why this exists: herdr latches `agent_status=blocked` after an answered dialog
+ * — the same upstream bug `hasActiveSpinner` guards against — but the spinner is
+ * not on screen yet, so the latch falls through to the no-evidence
+ * `awaiting-input` fallback. Measured on a live install: of 1432 captured
+ * `awaiting-input` block rows NONE carried a spinner, while every full-screen
+ * capture of the false positive carried this hint.
+ */
+export function hasQueuedInput(text: string): boolean {
+  return tailLines(text).some((l) => QUEUED_INPUT_RE.test(l));
+}
+
 /** Classify a blocked agent's terminal tail into an actionable shape. Never throws. */
 export function classifyBlocked(text: string): BlockReason {
   const tail = tailLines(text);
--- a/src/poller.ts
+++ b/src/poller.ts
@@ -18,6 +18,7 @@ import { herdrUsesExternalRegistrationSpawn } from "./herdr-capabilities";
 import {
   classifyBlocked,
   hasActiveSpinner,
+  hasQueuedInput,
   quotaBlockReason,
   tailLines,
   type BlockReason,
@@ -1730,11 +1731,17 @@ export class StatusPoller {
    * a wedged turn (or a static tail merely quoting a spinner-like line) and falls
    * through to the normal emit, re-arming the block instead of suppressing forever.
    * The first sighting of an episode gets a one-cadence grace (nothing to compare
-   * yet; the common case is a genuinely working spinner). The suppression episode
-   * is surfaced as the working-while-blocked display flag: `onWorkingBlocked(id, true)`
-   * once on entry; on re-arm (any block that will be emitted) `onWorkingBlocked(id,
-   * false)` fires first, then `onBlock`, in the same tick — clients see the flag drop
-   * and the block land together.
+   * yet; the common case is a genuinely working spinner).
+   *
+   * A second, unconditional suppression covers the same herdr latch BEFORE the spinner
+   * appears: a pane carrying QUEUED operator input (issue #2272). That one has no
+   * freshness gate — queued input is an observed state, not a liveness inference — and is
+   * vetoed by a detected `authUrl`, which can share the pane with it.
+   *
+   * Either suppression episode is surfaced as the working-while-blocked display flag:
+   * `onWorkingBlocked(id, true)` once on entry; on re-arm (any block that will be
+   * emitted) `onWorkingBlocked(id, false)` fires first, then `onBlock`, in the same
+   * tick — clients see the flag drop and the block land together.
    */
   private maybeClassify(s: Session, term: string): boolean {
     const id = s.id;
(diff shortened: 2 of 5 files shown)


scoop (commented):
# Review — #2272: don't raise awaiting-input on a queued-input pane

Diff under review: `68c2b233...d68a410e` — `src/blocked.ts`, `src/poller.ts`, `test/blocked.test.ts`,
`test/poller.test.ts`, `test/fixtures/fullscreen/queued-messages.txt`.

**Decision: comment.** No correctness or security defect survived verification. One non-blocking
behaviour note and one documentation nit below.

## What I verified

**`hasQueuedInput` detector (`src/blocked.ts:81`–`100`).**
- `QUEUED_INPUT_RE = /press up to \w+\s+(?:a\s+)?queued message/i` — I traced it by hand against all
  three wordings the comment claims (`src/blocked.ts:78-80`): `edit queued messages` matches via
  `\w+`=`edit`, `\s+`=` `, optional `a\s+` absent, then the literal; `select a queued message` matches
  with the optional group consuming `a `; `select a queued message to edit` matches as a prefix. No
  anchors, so all three hold.
- It reuses `tailLines(text)` (`src/blocked.ts:37-43`) — the same 15-non-empty-line window
  `hasActiveSpinner` (`src/blocked.ts:74`) and `classifyBlocked` (`src/blocked.ts:104`) use, so the
  doc comment's "same window as `classifyBlocked`" claim is accurate against `TAIL_LINES = 15`
  (`src/blocked.ts:25`). Empty input degrades safely: `tailLines("")` filters to `[]`, `.some()` →
  `false`.
- The fixture `test/fixtures/fullscreen/queued-messages.txt` is exactly 15 non-empty lines with the
  hint at line 10 and the footer below it — consistent with a captured `BlockReason.tail`
  (`src/blocked.ts:16`, 15-line cap). Confirmed via `cat -A` that the hint line is
  `❯ Press up to edit queued messages` with no ANSI, and that `classifyBlocked` on it yields
  `awaiting-input` (no `OPTION_RE` digit run at `src/blocked.ts:31`, no `YES_NO_RE` at
  `src/blocked.ts:32`), which is what `test/blocked.test.ts:101-103` asserts.

**Control flow refactor (`src/poller.ts:1746`–`1872`).**
- The restructure is behaviour-preserving for the pre-existing spinner path. Old guard was
  `reason.shape === "awaiting-input" && hasActiveSpinner(visible)` with an `else` that deleted
  `lastSuppressVisible`; new code splits that into `maybeClassify`'s shape check
  (`src/poller.ts:1761-1767`) plus `suppressAwaitingInput`'s `!queued && hasActiveSpinner`
  (`src/poller.ts:1792`) with its own `else` delete (`src/poller.ts:1808`). Both non-awaiting-input
  and awaiting-input-without-spinner still reach a `lastSuppressVisible.delete(id)`. Equivalent.
- `enterSuppression` (`src/poller.ts:1863-1872`) is a verbatim lift of the old
  `trySuppressSpinner` body (clear announced block once via `lastSig.has` → `emitBlock(id, null)`;
  raise the flag once via the `workingWhileBlocked` set). `trySuppressSpinner`
  (`src/poller.ts:1852-1858`) keeps the freshness gate and the `lastSuppressVisible` bookkeeping.
  No semantic change.
- Emission ordering contract preserved: re-arm still does `onWorkingBlocked(id, false)` before
  `emitBlock` in the same tick (`src/poller.ts:1772-1774`), matching the docstring at
  `src/poller.ts:1741-1744`.

**`authUrl` resolution move (`src/poller.ts:1824-1827`).**
- Side-effect frequency is unchanged from before the PR. The spinner branch still returns *before*
  the resolution (`src/poller.ts:1796`), so spinner-suppressed ticks still skip both
  `detectAuth(s)` and `confirmLoginUrl(...)`. The only ticks that newly reach it are ones that
  previously reached it anyway (awaiting-input without an active spinner) — they just now suppress
  afterwards instead of emitting. So `confirmLoginUrl`'s two-read stability state
  (`src/poller.ts:2025-2041`) and `restAuthPtyObserved`/`restAuthPtyConfirmed` are advanced on
  exactly the same cadence as before. No new I/O, no new cache churn.
- The auth veto is real, not decorative: `detectAuth` defaults to `readAuthUrl`
  (`src/poller.ts:180-189`), which is exception-safe and returns `null` for a session with no
  `claudeSessionId`; the PTY fallback runs `detectLoginAuthUrl` (`src/auth-url.ts:219-236`) through
  `confirmLoginUrl`. With a URL resolved, `queued && !authUrl` is false and the block emits with
  `reason.authUrl` attached — which is what `test/poller.test.ts:637-651` (the new auth-wins test)
  exercises. The pre-existing auth tests at `test/poller.test.ts:3398-3402` use panes with no queued
  hint, so they are unaffected.

**State lifecycle.**
- `workingWhileBlocked` still drops in the same tick the session leaves herdr-blocked
  (`src/poller.ts:1052-1055`), on reap (`src/poller.ts:993`), and on prune
  (`src/poller.ts:1244`, `1282`). The new branch adds no new key to clean up.
- `displayStatus` maps `blocked + workingBlocked → running`
  (`ui/src/lib/display-status.ts:15`), so the suppressed session reads as running to the operator —
  verified, no new UI wiring needed and none added.

**No collateral breakage.** `test/fixtures/fullscreen/` is not globbed by any test — the only two
references to the new fixture are the two new call sites (`test/blocked.test.ts:99`,
`test/poller.test.ts:573`). `readFileSync`/`join` are already imported at `test/poller.test.ts:2-3`.
No user-facing strings, no message catalogs, no API surface, no new dependency in this diff, so
locale parity and consumer-contract checks are not applicable.

## Behaviour note (non-blocking)

`suppressAwaitingInput` computes `queued` first and short-circuits the spinner branch
(`src/poller.ts:1791-1792`), and the queued branch carries no freshness gate
(`src/poller.ts:1834-1837`). Consequence: a pane showing **both** a live spinner and the queued hint
— the ordinary "operator queued a message while the agent works" state — that then **freezes** (a
genuinely wedged turn) used to re-arm the block on the next cadence via the freshness gate
(`src/poller.ts:1855`, rationale at `src/poller.ts:1797-1803`); it now suppresses indefinitely.
While herdr holds `blocked`, nothing else re-arms: `maybeProbe`/stall only runs on
`status === "running"` (`src/poller.ts:1063-1064`).

This is the trade-off the approved plan names under *Decisions → No re-arm while the hint shows* and
resolves with the operator, and it is bounded (herdr's latch resolves in seconds in the observed
data). I am recording it, not blocking on it — but note the ordering is not what makes the blind spot
exist: even with the spinner checked first, a frozen spinner pane falls through to `return false`
from `trySuppressSpinner` and would be swallowed by the unconditional queued branch anyway. The blind
spot is inherent to "no freshness gate", so the only way to keep the wedge escape would be to gate
the queued branch on `!hasActiveSpinner`. Operator's call.

- The corpus claims in the code comment (`src/blocked.ts:93-96`: 1432 awaiting-input rows, none with
  a spinner) and in the commit message rest on `~/.shepherd/shepherd.db`, which is outside this
  repository. I did not read it. Assumed as written. The reasoning is at least internally coherent:
  suppressed reads are never persisted, so "0 emitted rows carry a spinner" is evidence that the
  frozen-spinner fall-through never fired on this path, which is what falsifies the issue's
  hypothesis.
- The claim that Claude Code 2.1.266 ships exactly those three hint wordings
  (`src/blocked.ts:78-80`) is sourced from a third-party binary not present here. Not verified, not
  asserted either way. If a future wording interposes a count (e.g. `edit 2 queued messages`), the
  regex would stop matching — that is a degradation back to today's behaviour, not a new failure,
  and the plan flags it under Risks.
- The fixture is asserted to be the literal pane from session `1ab3d5c0-…`; I can confirm its shape
  is consistent with a stored 15-line `tail`, not that it is byte-identical to that DB row.

`tryHookAwaitingBlock` (src/poller.ts:1215-1222) runs when status !== "blocked" but calls maybeClassify, whose suppression path sets workingWhileBlocked — the src/poller.ts:1052-1055 clear already ran earlier in the same tick, so a hook-triggered suppression emits a spurious working:true that is only retracted on the next tick; this predates the diff and applies equally to the spinner branch.


**Nits (non-blocking):**

- [bug] src/poller.ts: the queued-input branch (lines 1834-1837) has no freshness gate and short-circuits the spinner branch, so a pane carrying BOTH a live spinner and the queued hint that later freezes — a genuinely wedged turn — now suppresses forever where the freshness gate used to re-arm it next cadence. The plan accepts this blind spot, but neither the maybeClassify docstring nor the branch comment records it; add a line naming the accepted no-re-arm consequence (nothing else re-arms while herdr holds blocked: maybeProbe/stall only runs on status === running, src/poller.ts:1063).
- [bug] src/poller.ts: the suppressAwaitingInput docstring says it will attach an authUrl when one is pending, but the spinner branch returns at line 1796 before the authUrl resolution at 1824, so a spinner-suppressed tick never resolves one. Worth one clause so the contract reads accurately.

Issue #2281, opened 2026-09-12
classifyBlocked forges a menu from prose numbered lists while the agent is working


scoop:
Split out of #2272, found while measuring the live block corpus for it. Distinct code path, not fixed there.

## What happens

`classifyBlocked` shapes a buffer as `menu` whenever it finds a contiguous `1. … 2. …` run in the last 15 non-empty lines. Recap/summary prose written as a numbered list satisfies that, so a session that is actively working raises a false "needs you" menu card — with clickable options that send `1`/`2` into a PTY that is mid-turn.

The `hasActiveSpinner` guard does not help: suppression is deliberately scoped to `awaiting-input` only (`src/poller.ts` → `maybeClassify`), on the reasoning that a genuine dialog must always surface, spinner or not.

## Evidence

3 rows in the live corpus carry the at-rest footer, a live spinner, and a prose-forged menu. One example (2026-07-22, session on `pulse-time-review-judge-skippedbycap`), abridged:

```
  1. Backlog counter — ai-judge.ts:893 sets result.skippedByBudget …
  2. Circuit-breaker — AI_JUDGE_MAX_FAILURES = 3 (ai-judge.ts:98) …
● User answered Claude's questions:
  ⎿  · #274 is already shipped by merged PR #276. How do you want to proceed? → Close #274
✽ Gallivanting… (35s · ↓ 2.2k tokens)
```

`hasActiveSpinner` returns **true** here and `classifyBlocked` still returns `shape: "menu"` with 2 options.

## Worth deciding

Whether a live spinner should veto a `menu` too (the "genuine dialog always surfaces" rule was written against herdr's latch, which cannot produce a spinner *and* a real open dialog), or whether the menu detector needs positive dialog framing — a real Claude Code menu carries a key-hint footer (`Enter to select · Tab/Arrow keys to navigate · Esc to cancel`) that prose never does.

Low volume (3 rows), but the failure mode is worse than a stale card: the options are actionable and send keystrokes to a working agent.

scoop:
✅ PR #2287 merged.

PR #2284, opened 2026-09-12
fix(blocked): match the two real spinner frames SPINNER_RE missed (#2282)

### LINKED ISSUE #2282: SPINNER_RE misses real working frames, so awaiting-input suppression doesn't engage
Split out of #2272, found while measuring the live block corpus for it.

## What happens

`SPINNER_RE` (`src/blocked.ts`) requires a glyph-anchored line whose ellipsis is directly followed by `(` plus an elapsed-time counter or `esc to interrupt`. Two real production frames do not match, so `hasActiveSpinner` reports the agent as not working and the poller's `awaiting-input` suppression does not engage:

1. **No parenthetical yet** — `✽ Mulling…` (early frame, before the counter appears).
2. **Parenthetical opens with text** — `· Julienning… (running stop hooks… 1/2 · 2m 31s · ↓ 8.7k tokens` — the elapsed counter is present but not adjacent to `…(`.

Both appear in the live corpus on panes that also show the at-rest footer, i.e. full-screen captures where the agent was demonstrably mid-turn.

## Why it is not a trivial widen

The `…(` adjacency and the glyph class are deliberate: the comment on `SPINNER_RE` records that `+` was dropped as a spinner glyph (0 occurrences in 889 captured tails, common markdown/diff leader) and that the adjacency rejects `… +5 lines (ctrl+o to expand)` and `(1M context)`. Loosening either dimension re-opens those. Case 1 in particular (an ellipsis with no parenthetical at all) is close to unconstrained prose.

The poller's freshness gate absorbs some of the residual risk — continued suppression requires the buffer to advance between classify reads — but it is not free: a false spinner match delays a genuine block by one reclassify cadence.

## Worth deciding

Whether to widen the regex (and against what corpus), or to stop inferring "working" from the spinner line alone and corroborate with the poller's existing `lastActiveTurnAt` / hook activity signals, which do not depend on TUI wording at all.

scoop:
Closes #2282.

`hasActiveSpinner` is the poller's "this agent is actually working" evidence, used to suppress the no-evidence `awaiting-input` fallback. Two real production spinner frames slipped past `SPINNER_RE`, so suppression never engaged and a false "needs you" nudge fired.

## What the corpus actually says

I measured the live install's block corpus first (`signals` where `kind='block'` — the same corpus #2272 was measured on) rather than guessing at the widening. Snapshot: 3911 unique tails, **1290 of them classifying `awaiting-input`**, 27 087 unique lines. (Those shape counts predate #2281, which landed while this was in review — see the rebase note at the bottom. The per-line measurements below are shape-independent and unchanged.)

The headline is worse than the issue implies: **the old pattern matched 0 of those 1290 tails.** Spinner suppression effectively never engaged in production.

The whole corpus holds exactly three glyph+ellipsis near-misses — the two frames the issue names, plus `⎿  Running…`, which is a *retained tool-result leader*, not a spinner frame (it sits in scrollback beside the tool's own output, outliving the turn it belongs to). That third one is why this could not be a loose "glyph + ellipsis" widen.

| variant | all 27 087 lines | `awaiting-input` lines |
| --- | --- | --- |
| old `SPINNER_RE` | 3 | **0** |
| timed, gap allowed inside the same `(…)` | 4 | 1 — the case-2 true positive |
| untimed `glyph + Word + …$`, `⎿` excluded | 1 | 1 — the case-1 true positive |
| untimed, same but `⎿` allowed | 2 | 2 — drags in `⎿ Running…` |

## The change

`SPINNER_RE` splits into two anchored, separately-justified shapes; `hasActiveSpinner` keeps its signature and its window, and scans both:

- **Timed** — today's pattern with the `…(` adjacency relaxed to a bounded run of non-`)` characters, so the counter may sit behind leading text in the **same** parenthetical (`· Julienning… (running stop hooks… 1/2 · 2m 31s · ↓ 8.7k tokens`). `[^)]` stops the counter being borrowed out of a *later* parenthetical; the length bound keeps it near this one's start even when the paren is truncated unclosed at pane width, which is exactly that frame's shape. Requiring a counter at all is what rejects `… +5 lines (ctrl+o to expand)` and `(1M context)` — neither carries an elapsed time — so relaxing adjacency does not re-open them.
- **Untimed** — new, for the frame before the counter appears (`✽ Mulling…`): glyph + one capitalized word + ellipsis at EOL, with `⎿` excluded. Deliberately the tightest shape that covers it, since an ellipsis with no parenthetical is otherwise close to unconstrained prose; the single capitalized word is what stops a `·` bullet truncated at pane width from forging a spinner.

## Decisions worth flagging to a reviewer

- **No poller change.** The freshness gate in `trySuppressSpinner()` stays the residual-risk absorber: a false match over a static buffer falls through and re-arms the block after one reclassify cadence (3 s).
- **Rejected the issue's alternative** of corroborating with `lastActiveTurnAt` / hook activity. `isActiveTurnFresh()`'s window is `2 × probeCheckMs` = 14 s, and a *genuine* `awaiting-input` block also follows fresh activity (the agent works, then asks) — so it is non-discriminative for ~14 s. As the primary signal it would delay every real block that long; as an extra AND-gate it buys nothing measurable while forcing `hasActiveSpinner` to stop being a pure one-argument predicate.
- **Not gerund-gated.** Requiring `-ing` scores identically on the corpus but would assume Claude Code's spinner vocabulary stays all-gerund forever.
- **Accepted miss:** multi-word untimed status lines (`✽ Adding i18n + final review…`) stay unmatched until their counter lands and the timed shape takes over — roughly a second wide.

## Verification

- Both #2282 panes added as fixtures, captured verbatim from the corpus (the `queued-messages.txt` convention from #2272). Each is asserted to classify `awaiting-input`, to carry no queued input (so #2272's branch cannot cover it), and to now read as working.
- Negative tests pin the deliberate exclusions: `⎿ Running…`, a truncated `·` bullet, lowercase/multi-word variants, and a counter in a later parenthetical.
- Every positive and negative already asserted in `test/blocked.test.ts` still holds, `fullscreen-spinner.txt` included.
- Re-measured with the shipped code: exactly **2** `awaiting-input` tails now match, and they are the two intended frames.
- Root `bun run lint`, `bunx tsc --noEmit` and `bun run test` (9753 pass / 0 fail) green; `fallow audit --base origin/main` clean on all 4 changed files.

## Rebase note — composition with #2281

Rebased onto `main` after #2287 (#2281, dialog-chrome gate for menus) landed in the same file. `src/blocked.ts` auto-merged; `test/blocked.test.ts` was a pure additive tail collision (both sides appended tests), resolved by keeping both.

Two things I checked rather than assumed, because #2281 measured its own corpus with the *old* spinner regex:

- **#2281's "NO row carries chrome AND a live spinner" invariant still holds** under the widened regex — re-measured: 0 of 2550 menus. Its comment stays accurate.
- **The two changes compose, and the result is better than either alone.** #2281 demotes chrome-less numbered runs to `awaiting-input`, which is exactly where this suppression lives. Of its 13 demoted prose-menus, 2 carry a live spinner (`· Bootstrapping… (1m 28s …)`, `✽ Gallivanting… (35s …)`) that was previously never consulted, because those panes classified `menu`. Combined, **4 of 1300 `awaiting-input` tails now suppress, where before either change it was 0.**

One line outside the planned diff: #2281's new comment cross-referenced `SPINNER_RE`, a symbol this PR's split removes, so that reference now points at the spinner patterns generally. The corpus comment also dropped its `awaiting-input` count, which #2281 legitimately moves — the line-level evidence it cites does not depend on it.

### DIFF
--- a/src/blocked.ts
+++ b/src/blocked.ts
@@ -42,7 +42,8 @@ const YES_NO_RE = /\(\s*y\s*\/\s*n\s*\)|\[\s*y\s*\/\s*n\s*\]/i;
 //
 // Selected-option caret. `*` is deliberately NOT in the caret set even though OPTION_RE accepts
 // it as an option marker: it is also the markdown bullet leader (same reasoning that keeps `+`
-// out of SPINNER_RE), and a bulleted numbered list is exactly the prose this guards against.
+// out of the spinner patterns below), and a bulleted numbered list is exactly the prose this
+// guards against.
 const CARET_OPTION_RE = /^[\s│|]*[❯>]\s*\d+[.)]/;
 // Key-hint footer, as a substring of the (·-separated, often word-wrapped) hint line. Matched on
 // the stable fragments rather than a whole wording: the live corpus carries at least a dozen
@@ -62,25 +63,54 @@ export function tailLines(text: string, n = TAIL_LINES): string[] {
     .slice(-n);
 }
 
-// Active-turn spinner line, anchored: the line must START with a spinner/tool
-// glyph (·✢✳✶✻✽*⎿), then carry an ellipsis directly followed by "(" + either an
-// elapsed-time counter or the legacy "esc to interrupt" hint, e.g.
-// "✶ Bunning… (1m 13s · ↑ 1.3k tokens)" / "⎿  Running… (4s)" /
-// "✻ Imagining… (esc to interrupt)". The glyph anchor rejects prose quoting a
-// time mid-text ("the build finished… (3m 12s)"), queued-input lines
-// ("❯ retry… (2m 30s)"), and a bare "esc to interrupt" on a non-spinner line;
-// the `…(` adjacency rejects "… +5 lines (ctrl+o to expand)" and "(1M context)"
-// (no elapsed time). `+` is excluded: zero occurrences as a spinner frame in
-// 889 production-captured tails, and a common markdown/diff line leader. `*`
-// IS a genuine production spinner frame and stays — its residual
-// markdown-bullet risk is covered by the poller's freshness gate (continued
-// suppression requires the buffer to advance between classify reads).
-const SPINNER_RE = /^\s*[·✢✳✶✻✽*⎿].*?…\s*\((?:(?:\d+h\s*)?(?:\d+m\s*)?\d+s\b|esc to interrupt)/i;
+// Active-turn spinner frames, in two anchored shapes. Both require a line that
+// STARTS with a spinner/tool glyph, which is what rejects prose quoting a time
+// mid-text ("the build finished… (3m 12s)"), queued-input lines
+// ("❯ retry… (2m 30s)"), and a bare "esc to interrupt" on a non-spinner line.
+// `+` is excluded from both: zero occurrences as a spinner frame across the
+// production-captured tails, and a common markdown/diff line leader. `*` IS a
+// genuine production spinner frame and stays — its residual markdown-bullet risk
+// is covered by the poller's freshness gate (continued suppression requires the
+// buffer to advance between classify reads).
+//
+// Both shapes were measured for #2282 against a live install's block corpus
+// (snapshot: 3911 unique tails, 27087 unique lines): each one matches exactly one
+// line the other misses, and nothing else in the corpus. Line-level by
+// construction, so #2281's shape gate does not move these numbers — but it does
+// feed them more panes, since a chrome-less numbered run it demotes to
+// `awaiting-input` now reaches the suppression path this evidence supports.
+
+// (1) Timed — glyph + ellipsis + a parenthetical carrying an elapsed-time counter
+// or the legacy "esc to interrupt" hint: "✶ Bunning… (1m 13s · ↑ 1.3k tokens)" /
+// "⎿  Running… (4s)" / "✻ Imagining… (esc to interrupt)". The counter need NOT be
+// adjacent to "…(" — it may sit behind leading text in the SAME parenthetical
+// ("· Julienning… (running stop hooks… 1/2 · 2m 31s · ↓ 8.7k tokens", #2282).
+// `[^)]` keeps the counter from being borrowed out of a LATER parenthetical, and
+// the length bound keeps it near this one's start even when the paren is truncated
+// unclosed at pane width (that frame's exact shape). Requiring a counter at all is
+// what rejects "… +5 lines (ctrl+o to expand)" and "(1M context)" — neither
+// carries an elapsed time — so relaxing the adjacency does not re-open them.
+const SPINNER_TIMED_RE =
+  /^\s*[·✢✳✶✻✽*⎿].*?…\s*\([^)\n]{0,40}?(?:(?:\d+h\s*)?(?:\d+m\s*)?\d+s\b|esc to interrupt)/i;
+
+// (2) Untimed — the frame BEFORE the counter appears: glyph + one capitalized word
+// + ellipsis at end of line ("✽ Mulling…", #2282). Deliberately the tightest shape
+// that covers it, because an ellipsis with no parenthetical is otherwise close to
+// unconstrained prose: the single capitalized alpha word is what stops a `·` bullet
+// truncated to an ellipsis at pane width from forging a spinner, and `⎿` is
+// excluded here because "⎿  Running…" is a tool-result leader RETAINED in
+// scrollback beside the tool's own output — it outlives the turn it belongs to.
+// Not gerund-gated: requiring "-ing" would assume Claude Code's spinner vocabulary
+// stays all-gerund forever, for no measured gain. Multi-word status lines
+// ("✽ Adding i18n + final review…") stay unmatched until their counter lands and
+// shape (1) takes over — an accepted miss, roughly a second wide.
+const SPINNER_UNTIMED_RE = /^\s*[·✢✳✶✻✽*]\s+[A-Z][a-z]+…\s*$/;
 
 /**
  * True when the terminal tail shows an actively-working Claude Code turn — a
- * glyph-anchored spinner line with an elapsed-time counter or the legacy
- * "esc to interrupt" hint. Scans the same last-15-non-empty-lines window as
+ * glyph-anchored spinner line, either carrying an elapsed-time counter / the
+ * legacy "esc to interrupt" hint or in the untimed frame that precedes the
+ * counter. Scans the same last-15-non-empty-lines window as
  * `classifyBlocked` (the spinner always sits just above the input box; this
  * avoids matching stale scrollback).
  *
@@ -91,7 +121,7 @@ const SPINNER_RE = /^\s*[·✢✳✶✻✽*⎿].*?…\s*\((?:(?:\d+h\s*)?(?:\d+m
  * against that upstream herdr bug.
  */
 export function hasActiveSpinner(text: string): boolean {
-  return tailLines(text).some((l) => SPINNER_RE.test(l));
+  return tailLines(text).some((l) => SPINNER_TIMED_RE.test(l) || SPINNER_UNTIMED_RE.test(l));
 }
 
 // The at-rest input box's queued-input hint. Claude Code 2.1.266 ships three
--- a/test/blocked.test.ts
+++ b/test/blocked.test.ts
@@ -89,6 +89,35 @@ test("hasActiveSpinner rejects idle/blocked buffers", () => {
   expect(hasActiveSpinner("❯\n⏵⏵ bypass permissions on (shift+tab to cycle)")).toBe(false);
 });
 
+test("hasActiveSpinner detects the two real frames it used to miss (#2282)", () => {
+  // the untimed frame, before the elapsed counter appears
+  expect(hasActiveSpinner("✽ Mulling…")).toBe(true);
+  // a counter that is present but not adjacent to "…(" — leading text in the same parenthetical,
+  // truncated unclosed at pane width
+  expect(hasActiveSpinner("· Julienning… (running stop hooks… 1/2 · 2m 31s · ↓ 8.7k tokens")).toBe(
+    true,
+  );
+});
+
+test("hasActiveSpinner rejects the retained '⎿ Running…' tool-result leader (#2282)", () => {
+  // Untimed shape excludes ⎿ on purpose: this leader stays in scrollback beside the tool's own
+  // output, so it outlives the turn it belongs to and is no evidence of a live one. Its TIMED
+  // form ("⎿  Running… (4s)") does count, and is asserted above.
+  expect(hasActiveSpinner("  ⎿  Running…")).toBe(false);
+});
+
+test("hasActiveSpinner's untimed shape stays tight around the spinner-word form", () => {
+  // a `·` bullet truncated to an ellipsis at pane width must not forge a spinner
+  expect(hasActiveSpinner("· This is ~11 pages incl. a story-template rebuild, index…")).toBe(
+    false,
+  );
+  // lowercase / multi-word / trailing-text variants are prose, not a spinner frame
+  expect(hasActiveSpinner("· mulling…")).toBe(false);
+  expect(hasActiveSpinner("✽ Mulling… and then some")).toBe(false);
+  // the relaxed timed shape must not borrow a counter out of a LATER parenthetical
+  expect(hasActiveSpinner("⎿  Read… (ctrl+o to expand) and it took (3s)")).toBe(false);
+});
+
 test("hasQueuedInput detects every queued-hint wording Claude Code ships", () => {
   expect(hasQueuedInput("❯ Press up to edit queued messages")).toBe(true);
   expect(hasQueuedInput("❯ Press up to select a queued message")).toBe(true);
@@ -382,4 +411,16 @@ test("looksLikeDemotedMenu marks exactly what classifyBlocked demoted", () => {
 
   // no numbered run at all → nothing was demoted
   expect(looksLikeDemotedMenu(classifyBlocked("What should I name it?\n❯").tail)).toBe(false);
+});
+
+test("real #2282 panes: both were mid-turn and now read as working", () => {
+  for (const name of ["spinner-untimed.txt", "spinner-inline-parenthetical.txt"]) {
+    const pane = readFileSync(join(fixturesDir, name), "utf8");
+    // each pane is the no-evidence fallback the poller would otherwise announce…
+    expect(classifyBlocked(pane).shape).toBe("awaiting-input");
+    // …and carries no queued input, so #2272's branch cannot cover it
+    expect(hasQueuedInput(pane)).toBe(false);
+    // …but the agent was demonstrably working, so suppression must engage
+    expect(hasActiveSpinner(pane)).toBe(true);
+  }
 });
--- a/test/fixtures/fullscreen/spinner-inline-parenthetical.txt
+++ b/test/fixtures/fullscreen/spinner-inline-parenthetical.txt
@@ -1,0 +1,15 @@
+     + full-bleed media, Card chrome). Implement them here? →
+     Implement here (Recommended)
+     · This is ~11 pages incl. a story-template rebuild, index
+     rebuild, 3 resource pages and über-uns. Ship as one PR or
+     promote to an epic? → Single PR (Recommended)
+· Julienning… (running stop hooks… 1/2 · 2m 31s · ↓ 8.7k tokens
+ · thinking)
+  ⎿  Tip: Share Claude Code and earn €10 in usage credits ·
+     /passes
+─────────────────────────────────────────────────────────────────
+❯
+─────────────────────────────────────────────────────────────────
+  …/.shepherd-worktrees/topmedia-website-visual-gap-audit-simi…
+  ⏵⏵ bypass permissions on (shift+tab to cycle) · ← for agents
+                                               ● high · /effort
--- a/test/fixtures/fullscreen/spinner-untimed.txt
+++ b/test/fixtures/fullscreen/spinner-untimed.txt
@@ -1,0 +1,15 @@
+  Two alternatives, for completeness:
+● User declined to answer questions
+  ⎿  · How should we resolve the jarring height step between passive readouts and tall buttons? (De-box the readouts / Equalize
+     heights, recede readouts / Shrink the delta + seam)
+● What would you like to clarify? Happy to reframe — whether it's the options themselves, a constraint I'm missing, or something
+  about how these controls behave that I should factor in first.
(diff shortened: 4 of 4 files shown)


scoop (commented):
# Review — #2282: match the two real spinner frames `SPINNER_RE` missed

**Decision: comment.** No blocking defects found. The split into `SPINNER_TIMED_RE` /
`SPINNER_UNTIMED_RE` does what the issue asked, the existing positive/negative battery still
holds when traced by hand, and nothing outside `hasActiveSpinner`'s predicate changed.

### The two target frames now match

- `src/blocked.ts:84` — `SPINNER_UNTIMED_RE = /^\s*[·✢✳✶✻✽*]\s+[A-Z][a-z]+…\s*$/`. Traced against
  `✽ Mulling…`: `^\s*` → empty, `✽` in class, `\s+` → one space, `[A-Z][a-z]+` → `Mulling`, `…`,
  `\s*$`. Matches. Success criterion 1 holds.
- `src/blocked.ts:70-71` — `SPINNER_TIMED_RE`. Traced against
  `· Julienning… (running stop hooks… 1/2 · 2m 31s · ↓ 8.7k tokens`: `.*?` takes ` Julienning`, the
  first `…` is consumed, `\s*\(` matches ` (`, then `[^)\n]{0,40}?` must span
  `running stop hooks… 1/2 · ` — **26 characters**, inside the 40 bound — before
  `(?:\d+m\s*)?\d+s\b` matches `2m 31s`. Matches. Success criterion 2 holds. The measured margin
  is 26 of 40.
- `src/blocked.ts:84` excludes `⎿` from the untimed class, and `  ⎿  Running…` has no `\(` after
  its `…`, so the timed shape cannot pick it up either. Success criterion 3 holds; asserted at
  `test/blocked.test.ts:105`.

### The relaxation does not re-open the cases the old comment protected

The `…\s*\(` adjacency is **unchanged** — only the interior of that same parenthetical was
relaxed. So:

- `… +5 lines (ctrl+o to expand)` (`test/blocked.test.ts:84`) still fails at the glyph anchor
  (`…` is not in either class) *and* at `…\s*\(` (next char is `+`).
- `Opus 4.8 (1M context)` (`test/blocked.test.ts:85`) has no leading glyph.
- `+ added a retry step… (2m 30s)` (`test/blocked.test.ts:159`) — `+` is in neither class.

I traced the new negative at `test/blocked.test.ts:117`,
`⎿  Read… (ctrl+o to expand) and it took (3s)`, character by character: after `…` the engine
enters the first parenthetical, `[^)\n]` cannot cross the closing `)`, no counter exists inside
`ctrl+o to expand`, and there is no second `…` for `.*?` to backtrack onto. It correctly returns
false — the counter is not borrowed out of the later parenthetical, as the comment at
`src/blocked.ts:65-66` claims.

### No regression across the fixture corpus or the existing suites

I ran both new patterns as literal greps over `test/fixtures/`:

- Untimed shape matches exactly one line in the whole fixture tree:
  `test/fixtures/fullscreen/spinner-untimed.txt:9`.
- Timed shape matches exactly three: `test/fixtures/fullscreen/classic-spinner.txt:13`
  (`✢ Skedaddling… (9s)`), `test/fixtures/fullscreen/fullscreen-spinner.txt:16`
  (`✽ Newspapering… (12s · ↓ 123 tokens)`, the assertion at `test/blocked.test.ts:355-358` is
  preserved) and `test/fixtures/fullscreen/spinner-inline-parenthetical.txt:6`.
- `queued-messages.txt` matches neither, so `test/blocked.test.ts:134`
  (`hasActiveSpinner(pane) === false`) still holds; likewise `fullscreen-idle.txt` for
  `test/poller.test.ts:3044`.
- The spinner-shaped strings in `test/poller.test.ts` (`Computing… (1s)` at :849/:1023/:1094,
  `$ slow-cmd  (esc to interrupt · ${frame}s)` at :937, `* Done… (3s)` at :556,
  `✶ Bunning… (1m 13s · ↑ 1.3k tokens)` at :280) all classify identically before and after —
  the first three have no leading glyph, the last two matched under the old pattern and still do.

### The new fixtures really are `awaiting-input` panes

Checked both against `OPTION_RE` (`src/blocked.ts:31`) and `YES_NO_RE` (`src/blocked.ts:32`): no
line in either file has a `\d+[.)]`-shaped option leader after the optional `[❯>*]` prefix, and
neither carries `(y/n)`, so `classifyBlocked` falls through to `awaiting-input` at
`src/blocked.ts:161` — matching the assertion at `test/blocked.test.ts:365`. Neither pane carries
`press up to … queued message` (`src/blocked.ts:108`), so `hasQueuedInput` is false, matching
`test/blocked.test.ts:367`. Both files are 15 non-empty lines, i.e. exactly the `TAIL_LINES`
window, and the spinner line sits inside it in both.

Both fixtures are also internally consistent as real captures: `spinner-untimed.txt`'s separators
are 128 columns and its longest content line is 128 chars; `spinner-inline-parenthetical.txt`'s
are 65 columns and line 6 wraps onto line 7 (` · thinking)`) at that width — which is exactly why
its parenthetical is truncated unclosed, the shape the `[^)\n]` relaxation exists for.

### Blast radius of a false untimed match is bounded

`src/poller.ts:1792` is the only consumer. A first sighting routes into
`trySuppressSpinner` (`src/poller.ts:1852-1857`), which stores the buffer and calls
`enterSuppression` (`src/poller.ts:1863-1872`) — one `emitBlock(id, null)` plus a
working-while-blocked flag. On the next reclassify over an unchanged buffer,
`prev === visible` returns false and the block re-arms. So a false positive on a static pane costs
one reclassify cadence of delay and a single clear/re-arm flicker, never a permanently swallowed
block. This matches what the plan claimed and I confirmed it in the poller rather than assuming it.

### Scope

The diff touches `src/blocked.ts`, `test/blocked.test.ts` and two new fixtures — nothing in
`src/poller.ts`, no change to `hasQueuedInput`, no signature change (`hasActiveSpinner(text: string)
: boolean` at `src/blocked.ts:100` is unchanged, and its only two call sites are
`src/poller.ts:1792` and the tests). No user-facing strings and no message-catalog keys are
involved, so there is no locale-parity surface here. No house rule in the supplied set applies to
this diff.

## Stated limitations

- The corpus numbers in the comment at `src/blocked.ts:55-58` (3911 unique tails, 1290
  `awaiting-input`, 27087 unique lines, "nothing else in the corpus") describe a live install's
  `signals` table that is not in this repository. I could not verify them and am not asserting
  they are right or wrong. Everything I *could* check — the fixture tree and both test suites — I
  checked and reported above.

## Nits detail

**1. Untimed shape allows arbitrary leading indentation (`src/blocked.ts:84`).**
`^\s*` lets an *indented* `·` match. That matters because `·` demonstrably appears as a nested
bullet leader **inside a retained `⎿` tool-result block** — see
`test/fixtures/fullscreen/spinner-untimed.txt:3-4`, where `  ⎿  · How should we resolve…` wraps to
an indented continuation. A sibling nested bullet rendering as `     · Something…` would satisfy
the untimed pattern, which is precisely the retained-scrollback hazard `⎿` was excluded from the
class to avoid (comment at `src/blocked.ts:77-79`). Every genuine spinner frame in the fixture
tree sits at column 0 — `classic-spinner.txt:13`, `fullscreen-spinner.txt:16`,
`spinner-untimed.txt:9`, `spinner-inline-parenthetical.txt:6` — so anchoring the untimed branch at
column 0 (dropping the `\s*`, or bounding it) would close the gap at zero cost to the frames it
must catch. Non-blocking: the corpus measurement reports no such line today, and the freshness
gate caps the cost at one cadence.

**2. The corpus sentence over-claims (`src/blocked.ts:55-58`).**
"each one matches exactly one line the other misses, and nothing else in the corpus" is not true of
the timed shape: it also matches the three lines the *old* regex already matched, which the untimed
shape misses. The intended claim is presumably "each shape adds exactly one new match over the
previous pattern". Worth one word of tightening so a future widening decision reads the right
baseline off this comment.

**3. `{0,40}` is an unexplained constant (`src/blocked.ts:71`).**
The comment at `src/blocked.ts:66-67` explains *why* a bound exists but not why 40. The measured
frame needs 26, so the headroom is ~1.5×; saying that inline would let the next person widen or
tighten it deliberately rather than guess.


**Nits (non-blocking):**

- [bug] src/blocked.ts:84: SPINNER_UNTIMED_RE's leading `^\s*` lets an INDENTED `·` match, but `·` is also a nested bullet leader inside retained `⎿` tool-result blocks — see test/fixtures/fullscreen/spinner-untimed.txt:3-4, where `  ⎿  · How should we resolve…` wraps. A sibling nested bullet rendering as `     · Something…` would satisfy the pattern, which is the same retained-scrollback hazard `⎿` was excluded for. Every real spinner frame in the fixture tree is at column 0 (classic-spinner.txt:13, fullscreen-spinner.txt:16, spinner-untimed.txt:9, spinner-inline-parenthetical.txt:6), so anchoring the untimed branch at column 0 closes the gap for free.
- [compliance] src/blocked.ts:71: the `{0,40}` bound is an unexplained constant — the comment explains why a bound exists but not why 40. The measured frame needs 26 characters before its counter, so state that headroom inline and the next person can widen or tighten it deliberately.

scoop (commented):
# Review — #2282: match the two real spinner frames `SPINNER_RE` missed

Diff reviewed: `git diff 43a2a61f3b4673a0ae86d149cd4482edd2aeae3c...HEAD` — `src/blocked.ts`,
`test/blocked.test.ts`, plus two new fixtures under `test/fixtures/fullscreen/`.

**Decision: comment.** No correctness, security, or contract defect found. Nits only.

## What the diff does

`src/blocked.ts:93-107` splits the single `SPINNER_RE` into `SPINNER_TIMED_RE` (the old pattern
with the `…(` adjacency relaxed to `[^)\n]{0,40}?`) and a new `SPINNER_UNTIMED_RE`
(`/^\s*[·✢✳✶✻✽*]\s+[A-Z][a-z]+…\s*$/`). `hasActiveSpinner` tests both
(`src/blocked.ts:124`). No other production file changed.

## Verification performed

**Success criterion 1 — `hasActiveSpinner("✽ Mulling…")` is true.** Traced
`SPINNER_UNTIMED_RE` (`src/blocked.ts:107`) against the literal: `^\s*` → empty, glyph class
matches `✽` (U+273D, present in the class), `\s+` → the space, `[A-Z][a-z]+` → `Mulling`, `…` →
U+2026, `\s*$` → EOL. Matches. Asserted at `test/blocked.test.ts:94`.

**Success criterion 2 — the inline-parenthetical frame is true.** Traced `SPINNER_TIMED_RE`
(`src/blocked.ts:93-94`) against `· Julienning… (running stop hooks… 1/2 · 2m 31s · ↓ 8.7k tokens`:
`.*?…` takes the first `…` (after `Julienning`), `\s*\(` takes ` (`, the lazy `[^)\n]{0,40}?` walks
to offset 26 (`running stop hooks… 1/2 · `), then `(?:\d+m\s*)?\d+s\b` matches `2m 31s` with `\b`
satisfied by the following space. 26 ≤ 40, so the bound admits it. Asserted at
`test/blocked.test.ts:97`.

**Success criterion 3 — `"  ⎿  Running…"` is false.** `⎿` (U+23BF) is absent from the untimed
glyph class (`src/blocked.ts:107`) and the timed pattern requires `\(`, which the line has not.
No match. Asserted at `test/blocked.test.ts:106`.

**The relaxation cannot borrow a counter from a later parenthetical.** Traced
`⎿  Read… (ctrl+o to expand) and it took (3s)` (`test/blocked.test.ts:118`): the only `…` in the
line is after `Read`, `\s*\(` takes ` (`, and `[^)\n]{0,40}?` cannot cross the `)` after `expand`,
so it can consume at most `ctrl+o to expand` — which carries neither `\d+s` nor
`esc to interrupt`. `.*?` then has no later `…` to retry from, so the whole match fails. The
comment's claim at `src/blocked.ts:88` ("`[^)]` keeps the counter from being borrowed out of a
LATER parenthetical") is accurate.

**No fixture regression.** I ran both the old and the new timed pattern over `test/fixtures/`:
the old matches `prose-menu.txt:10`, `fullscreen-spinner.txt:16`, `classic-spinner.txt:13`; the
new matches those three plus exactly `spinner-inline-parenthetical.txt:6` — no other line. The
untimed pattern matches exactly one line in the whole `test/` tree,
`test/fixtures/fullscreen/spinner-untimed.txt:9`. This confirms the assertions that must keep
holding: `hasActiveSpinner(fullscreen-spinner.txt)` true (`test/blocked.test.ts:357-358`),
`hasActiveSpinner(prose-menu.txt)` true (`test/blocked.test.ts:369`),
`hasActiveSpinner(queued-messages.txt)` false (`test/blocked.test.ts:135`), and
`hasActiveSpinner(fullscreen-idle.txt)` false (`test/poller.test.ts:3066`).

**Both new fixture tests hold.** I read both panes line by line.
`spinner-inline-parenthetical.txt` and `spinner-untimed.txt` each hold 15 non-empty lines (the
whole `TAIL_LINES = 15` window, `src/blocked.ts:25`), so the spinner line — line 6 and line 9
respectively — is inside it. Neither pane carries a contiguous 1..n numbered run that `OPTION_RE`
(`src/blocked.ts:31`) would capture: the `…/…worktrees/…` path line and the `● high · /effort`
line both fail the leading-`(\d+)[.)]` requirement, and no other line starts with a digit. Neither
carries `(y/n)` for `YES_NO_RE` (`src/blocked.ts:32`). So both fall through to
`awaiting-input` at `src/blocked.ts:223`, as asserted at `test/blocked.test.ts:420`. Neither
carries the `press up to … queued message` spine (`src/blocked.ts:131`), so
`hasQueuedInput` is false, as asserted at `test/blocked.test.ts:422`. The test string at
`test/blocked.test.ts:97` is byte-identical to `spinner-inline-parenthetical.txt:6`.

**Identifiers resolve.** `fixturesDir` used by the new test at `test/blocked.test.ts:418` is
defined at `test/blocked.test.ts:270`; both fixture basenames exist on disk
(`test/fixtures/fullscreen/spinner-untimed.txt`,
`test/fixtures/fullscreen/spinner-inline-parenthetical.txt`). `readFileSync`/`join` are already
imported (`test/blocked.test.ts:2-3`). No consumer of the renamed constant survives: grepping the
tree, `SPINNER_RE` now appears only in `docs/spikes/1043-fullscreen-alt-screen.md:153,165` and
`docs/research/termdraw-integration.md:48` (dated historical reports — see nits), and nothing
imports it. `hasActiveSpinner` keeps its `(text: string) => boolean` signature, and its sole
production caller `src/poller.ts:1797` is unchanged and still type-compatible.

**Residual-risk claim checked against the real gate.** The comment at `src/blocked.ts:72-74`
leans on the poller's freshness gate. `trySuppressSpinner` (`src/poller.ts:1857-1863`) does what
the comment says: it stores the `visible` buffer and returns `false` (re-arm) when the buffer is
byte-identical to the previous classify read, so a false spinner match over a static pane costs
one `reclassifyMs` cadence rather than permanent suppression. No poller change was made, matching
the plan's `Out of scope`.

**No ReDoS.** `.*?…\s*\([^)\n]{0,40}?…` backtracks over a pane-width-bounded line with an inner
bound of 40 — linear in practice. Neither regex carries `g`, so `.test()` has no `lastIndex`
state. Both glyph classes are BMP-only (U+00B7, U+2722, U+2733, U+2736, U+273B, U+273D, U+23BF,
U+2026), so the missing `u` flag is not a surrogate-pair hazard. `SPINNER_UNTIMED_RE` correctly
omits the `i` flag, which is what makes `[A-Z][a-z]+` load-bearing — pinned by
`test/blocked.test.ts:115`.

**Cross-reference updated.** The `CARET_OPTION_RE` comment that used to name `SPINNER_RE` was
updated to "the spinner patterns below" (`src/blocked.ts:45`), so no in-source reference dangles.

## Stated limitations

- The corpus numbers in the new comment (`src/blocked.ts:77` — 3911 unique tails, 27087 unique
  lines, and "each one matches exactly one line the other misses") come from a live install's
  `signals` table. That database is not in this repo, so I could not reproduce the measurement or
  confirm the two fixtures are verbatim captures from it. I verified the shapes are internally
  consistent and regression-free against the committed fixture corpus instead, and neither
  asserted nor assumed the live-corpus figures.
- Success criterion 5 in the plan (re-measuring the live corpus: 1290 → 2 matched) is likewise
  unverifiable from here.

- `src/blocked.ts:94` — the `{0,40}` bound on the timed shape is a hard cutoff, so a wider leading text inside the same parenthetical (e.g. a longer hook name in `(running stop hooks: <long-name>… 1/2 · 2m 31s`) silently drops back to a miss; unreachable with today's observed wording, and the comment documents the bound, but the failure is silent rather than degraded.

- Nothing. The diff touches exactly the four files the plan named, makes no poller change, and adds no public surface — `SPINNER_TIMED_RE` / `SPINNER_UNTIMED_RE` are both module-private.

- None of the curated rules bear on a line-level regex change in `src/blocked.ts`; no rule in the supplied set is engaged by this diff.

## Nits (see JSON)

Recorded as `nit` findings: comment-precision points on the corpus sentence and the dropped `+`
occurrence count, the `*` glyph retained in the untimed class, the now-dangling `SPINNER_RE`
mentions in the two docs files, and the stale `src/blocked.ts:91` line pointer in
`test/blocked.test.ts` (pre-existing, but this diff pushes it further off).


**Nits (non-blocking):**

- [bug] src/blocked.ts:77 — the comment says each shape matches exactly one line the other misses and nothing else in the corpus, but the timed shape also matches the three lines the old pattern already matched (verified: prose-menu.txt:10, fullscreen-spinner.txt:16, classic-spinner.txt:13). Reword to make clear the one-line figure is the DELTA, not the total.
- [compliance] src/blocked.ts:70 — the old comment carried the concrete evidence for excluding +, zero occurrences in 889 captured tails; the rewrite drops the number for the vaguer across the production-captured tails. Keeping the count preserves the auditable basis for the exclusion.
- [bug] src/blocked.ts:107 — * stays in the untimed glyph class, which is the loosest new dimension: a markdown bullet of the form * Deploying… now reads as a live turn, whereas under the timed shape * needed a parenthetical counter too. Worth considering dropping * from the untimed class only (it stays a real frame in the timed one).
- [compliance] src/blocked.ts:93 — renaming SPINNER_RE leaves the identifier dangling in docs/spikes/1043-fullscreen-alt-screen.md:153,165 and docs/research/termdraw-integration.md:48. Both are dated historical reports so leaving them is defensible, but a parenthetical note of the new names would keep them greppable.
- [compliance] test/blocked.test.ts:290 — the comment points at the zero-space accept rule as src/blocked.ts:91; that pointer was already off before this PR and the +30 lines added to blocked.ts push it further (the rule now lives at src/blocked.ts:163). Cheap to correct while the file is open.

PR #2285, opened 2026-09-12
fix(ui): make statusTip's Esc dismissal reachable from the keyboard

### LINKED ISSUE #2283: statusTip's keyboard paths are unreachable on every trigger in the repo
`statusTip` is the repo's standard "explain this control" affordance, but **every** trigger it is
attached to is a non-focusable element, so the action's `focus` / `blur` / `keydown` listeners
(`ui/src/lib/actions/statusTip.svelte.ts:198-201`) can never fire on any of them.

Known triggers, all plain `span` / `div` / `li` with no `tabindex` and no `role`:

- `StatusPip.svelte`
- `CriticBadge.svelte`
- `PlanGateBadge.svelte`
- `AutopilotBadge.svelte`
- `TerminalBadge.svelte`
- `PromptSources.svelte`
- `IssueLoadAttempts.svelte`
- `SlashCommandMenu.svelte`
- `UnitRow.svelte` — `.chip-cold-resume` (added in #2269)

## Consequences

1. **Esc-to-dismiss is dead.** `onKeydown` handles `Escape` at `statusTip.svelte.ts:172-174`, but the
   keystroke never reaches a non-focusable trigger. Once a pointer click pins a tip
   (`pinned = true`, `:170`), the only dismissal paths that actually run are an outside
   `pointerdown` (`:109-113`, `:135`) and scroll/resize (`:114-121`).
2. **Focus-to-open is dead** (`onFocus` → `show()` at `:161-163`).
3. **The keyboard/AT path rests entirely on `aria-description`** (set in `enable()` at `:186`).
   That is genuine, but support for `aria-description` on a *role-less generic* element is patchy —
   Chrome exposes it, several screen readers skip description on generic nodes in browse mode. So
   the repo-wide assumption documented at `statusTip.svelte.ts:37-41` is more optimistic than the
   platform currently guarantees.

## Why this is filed as its own issue

#2269 raised this on `.chip-cold-resume` and deliberately did **not** add a tab stop there: a tab
stop per cold card would be noise in a long Herd, and fixing one trigger in isolation would make it
the outlier in a family of nine. The comment in that PR was corrected to describe what the element
actually does rather than claim a keyboard path it does not have — the honest short-term fix, but it
leaves the family-wide gap open.

## Worth deciding

- Give `statusTip` an opt-in `focusable` param that sets `tabindex="0"` plus an appropriate `role`,
  and adopt it on the triggers where a keyboard user genuinely needs the explanation (the badges
  that carry attention state, rather than every decorative pip).
- Or keep every trigger non-focusable and verify `aria-description` announcement against at least
  one real screen reader, then document the finding at `statusTip.svelte.ts:37-41` so the next
  reader is not relying on an untested assumption.

Either way the current state — listeners written, wired, and unreachable — should not persist, since
it reads as working code.

Found during the merge train for #2269.

scoop:
Closes #2283.

> **Rebased onto #2290**, which independently added a partial fix for the same bug. This PR is now the *delta* over it — see "Relationship to #2290" below. The action also moved to `ui/src/lib/tooltips/` in that PR; this change follows it there.

`statusTip`'s `keydown` listener sat on the trigger node, but nearly every trigger is a non-focusable `span` / `div` / `li` — a real Escape targets `<body>` and never reached it. A pointer click **pins** a tip, so a pinned tip had **no keyboard exit at all**: only outside-click and scroll dismissed it. That's a WCAG 2.2 SC 1.4.13 (*Content on Hover or Focus → Dismissible*) failure.

## Relationship to #2290

#2290 added `document.addEventListener("keydown", onKeydown)` in `show()`, **bubble phase, no `preventDefault`**, while keeping the node listener. That makes Escape *reachable* — the important half — but leaves two defects:

1. **It still loses to `a11yDialog`.** `a11yDialog` closes its host dialog on Escape unless `defaultPrevented`, and its listener sits on the dialog **node**, whose bubble phase runs *before* any listener on `document`/`window`. `statusTip` lives inside the New Task modal (`InstrumentToggle`'s ON/OFF suffix), so dismissing a pinned tip there still closes the modal first and **discards the operator's typed task**.
2. **The node `keydown` listener is now genuinely redundant** — it was kept alongside the new one.

This PR moves the listener to **window, capture phase**, which is what actually wins that race (the protocol `IssueDetailsPopover.svelte` already documents), drops the redundant node listener, and makes consumption intent-scoped.

## Consumption follows intent

- **Pinned** (deliberate pointer click) → `preventDefault()` + `stopPropagation()`. The tip is an overlay the operator opened, so it owns this Escape — and this is what protects the New Task modal.
- **Hovered** (incidental) → hide, but let the key run on. Found in self-review: consuming here would swallow the Escape that closes `SlashCommandMenu` when the pointer merely rests on a command description.

`stopPropagation`, not `stopImmediatePropagation`, so several open tips still close on one Escape.

## Three corrections to the issue as filed

Both of #2283's proposed options lead somewhere unhelpful:

1. **"Every trigger is non-focusable" is false.** `BuildQueueBadge`'s interactive branches are real `<button>`s (its coarse-pointer branch is a read-only `<span>`). `focus`/`blur` fire there today — deleting them would have been a regression, so they stay.
2. **The Herd row already has the keyboard path the issue asks for.** `UnitRow`'s `.unit-hit` overlay is a focusable `<button>` carrying `aria-describedby`. Card badges are enrichment on top of it, not the only route.
3. **Swapping `aria-description` → `aria-describedby` buys nothing.** Both feed the *same* accessible-description property; announcement is gated by the trigger's **role**, not by which attribute supplied the string. The only real lever is making triggers interactive — the tab-stop noise #2269 deliberately rejected.

So: **no new tab stops, no `focusable` param, no call-site markup changes.** The doc block now states the invariant callers must honour — the trigger's accessible *name* already carries the meaning (`role="img"` + `aria-label`, or visible text), and this tooltip must never be the only place an explanation exists. `UnitRow`'s cold-resume comment is reconciled with #2290's wording too.

## The old test was the bug in miniature

The original Esc spec dispatched a synthetic `keydown` **on the pip element** — something a browser can never do to a non-focusable span. It was green while the feature was broken. It now drives a real `userEvent.keyboard("{Escape}")`, plus two specs pinning the pinned-consumes / hovered-passes-through split.

## Verification

`bun run lint` clean · `cd ui && bun run check` 0 errors (6296 files) · `cd ui && bun run test` 319 files / 5199 tests · root `bun run test` 9783 tests, 0 fail · `bunx fallow audit --base origin/main --fail-on-issues` → *No issues in 3 changed files* · pre-push green. #2290's own new `tooltips/statusTip.browser.test.ts` passes unchanged against the capture-phase listener.

No new user-facing strings, so no i18n or feature-catalog entry. 3 files, `ui/` only.

### DIFF
--- a/ui/src/lib/components/StatusPip.browser.test.ts
+++ b/ui/src/lib/components/StatusPip.browser.test.ts
@@ -1,6 +1,6 @@
 import { describe, it, expect, vi, afterEach } from "vitest";
 import { render } from "vitest-browser-svelte";
-import { page } from "vitest/browser";
+import { page, userEvent } from "vitest/browser";
 import "../../app.css";
 import { m } from "$lib/paraglide/messages";
 import { statusLabel } from "$lib/format";
@@ -82,9 +82,41 @@ describe("StatusPip tip mode (statusTip action)", () => {
     leave(); // pinned → stays open (a real affordance, not a fleeting hover)
     await new Promise((r) => setTimeout(r, 30));
     expect(tooltipOpen()).toBe(true);
-    // Esc dismisses the pinned tooltip.
-    pipEl().dispatchEvent(new KeyboardEvent("keydown", { key: "Escape", bubbles: true }));
+    // Esc dismisses the pinned tooltip. Driven as a REAL keystroke, deliberately: the pip is a
+    // non-focusable <span>, so a genuine keydown targets <body> and can never reach the trigger
+    // itself. Dispatching the event on the pip (as this spec once did) models something the
+    // browser cannot produce, and passed while the feature was broken — see #2283.
+    await userEvent.keyboard("{Escape}");
     await vi.waitFor(() => expect(tooltipOpen()).toBe(false));
+  });
+
+  it("Esc is preventDefaulted, so a11yDialog (checks defaultPrevented) won't also close the host", async () => {
+    render(StatusPip, { status: "running", tip: true });
+    enter();
+    click(1); // pin it, so the tooltip is the topmost transient overlay
+    await vi.waitFor(() => expect(tooltipOpen()).toBe(true));
+
+    const ev = new KeyboardEvent("keydown", { key: "Escape", bubbles: true, cancelable: true });
+    document.body.dispatchEvent(ev);
+
+    await vi.waitFor(() => expect(tooltipOpen()).toBe(false));
+    // a11yDialog's Escape handler returns early on defaultPrevented, so an open tip inside the
+    // New Task modal dismisses itself WITHOUT discarding the operator's typed task.
+    expect(ev.defaultPrevented).toBe(true);
+  });
+
+  it("a HOVERED (unpinned) tip closes on Esc but does NOT swallow it", async () => {
+    render(StatusPip, { status: "running", tip: true });
+    enter(); // hover only — never pinned, so the operator never asked for this tip
+    await vi.waitFor(() => expect(tooltipOpen()).toBe(true));
+
+    const ev = new KeyboardEvent("keydown", { key: "Escape", bubbles: true, cancelable: true });
+    document.body.dispatchEvent(ev);
+
+    await vi.waitFor(() => expect(tooltipOpen()).toBe(false));
+    // Must stay reachable: consuming it here would eat the Escape that closes the slash-command
+    // menu whose description the pointer is resting on (ComposeBar handles that on the textarea).
+    expect(ev.defaultPrevented).toBe(false);
   });
 
   it("detail (tip off) keeps the native title and does not add aria-description", async () => {
--- a/ui/src/lib/components/UnitRow.svelte
+++ b/ui/src/lib/components/UnitRow.svelte
@@ -620,11 +620,12 @@
   // `aria-description`. Raising it is also what keeps a click on the chip from selecting the row —
   // selection lives on the sibling `.unit-hit` button, which the click can no longer reach
   // (statusTip's stopPropagation is belt-and-braces; nothing between here and `.unit` listens).
-  // The chip stays a NON-FOCUSABLE <span>, like every statusTip trigger in the repo, so the
-  // keyboard reads `aria-description`, while Escape dismisses a mouse-opened panel globally.
-  // A tab stop per cold card would be a lot of noise in a long Herd. A GlossaryTerm
-  // is wrong for a different reason — its click presentation pushes content down and would
-  // reflow the card.
+  // The chip stays a NON-FOCUSABLE <span>, so the action's focus-to-open path never fires on it —
+  // a tab stop per cold card would be a lot of noise in a long Herd, and the row's own `.unit-hit`
+  // overlay is the keyboard entry point (it carries `aria-describedby`). Escape still dismisses a
+  // mouse-opened panel globally, and a PINNED one consumes the key so it can't also close a host
+  // dialog (#2283). A GlossaryTerm is wrong for a different reason — its click presentation pushes
+  // content down and would reflow the card.
   //
   // Shared structured explanation: cost first, then the choices before resuming.
   const coldResume = $derived(isColdResume(session, nowMs));
--- a/ui/src/lib/tooltips/statusTip.svelte.ts
+++ b/ui/src/lib/tooltips/statusTip.svelte.ts
@@ -37,11 +37,31 @@ let uid = 0;
  * Click propagation is stopped by default so read-only chips never select their
  * row; actionable controls can opt out while retaining the same explanation.
  *
- * The explanation is exposed to assistive tech via `aria-description` (a string —
- * always valid, no dangling IDREF, present from mount), so screen readers announce
- * it in browse mode even though the chip is not a Tab stop. The *visual* popover is
- * created lazily on first open. Pass `null` to disable (callers gate with
- * `tip ? {...} : null`).
+ * Triggers may be focusable or not, and both are supported. Most are plain `span` /
+ * `div` / `li`; BuildQueueBadge's interactive branches are real `<button>`s (its
+ * coarse-pointer branch is a read-only `<span>`), and those are what `focus`/`blur`
+ * (focus-to-open) serve. Escape dismissal does NOT depend on that — it is a
+ * window-level listener scoped to the open state, so it reaches a pinned tip on any
+ * trigger (see `onKeydown`).
+ *
+ * ON ASSISTIVE TECH — read before assuming this carries the meaning. The explanation
+ * is mirrored into `aria-description` (a string: always valid, no dangling IDREF,
+ * present from mount) and the *visual* popover is created lazily on first open, so
+ * hidden tooltip text never pollutes the DOM / text queries. But `aria-description`
+ * is **enrichment, not the carrier of meaning**: it and `aria-describedby` feed the
+ * same accessible-description property, and announcement of that property is gated by
+ * the trigger's ROLE — inconsistent on `role="img"` and on role-less generics, which
+ * is what nearly every trigger here is. Swapping the attribute would change nothing;
+ * only making the trigger interactive would, and that is deliberately not done (a tab
+ * stop per chip would be noise in a long Herd — #2269, #2283).
+ *
+ * So the invariant callers MUST honour: the trigger's accessible NAME already carries
+ * the meaning — `role="img"` + `aria-label` (StatusPip, the badges, the sandbox/quota
+ * chips) or visible text (SlashCommandMenu, PromptSources, IssueLoadAttempts, the
+ * attention/merging/ready badges) — with the Herd row's `.unit-hit` `aria-describedby`
+ * behind it. Never let this tooltip be the only place an explanation exists.
+ *
+ * Pass `null` to disable (callers gate with `tip ? {...} : null`).
  */
 export const statusTip: Action<HTMLElement, StatusTipParams | null | undefined> = (
   node,
@@ -141,7 +161,7 @@ export const statusTip: Action<HTMLElement, StatusTipParams | null | undefined> 
     open = true;
     stopAnchor = anchorPopover(node, pop, 6);
     document.addEventListener("pointerdown", onDocPointerDown, true);
-    document.addEventListener("keydown", onKeydown);
+    window.addEventListener("keydown", onKeydown, { capture: true });
     window.addEventListener("scroll", onScrollOrResize, { capture: true, passive: true });
     window.addEventListener("resize", onScrollOrResize, { passive: true });
   }
@@ -154,7 +174,7 @@ export const statusTip: Action<HTMLElement, StatusTipParams | null | undefined> 
     stopAnchor?.(); // stops autoUpdate + hidePopover()
     stopAnchor = null;
     document.removeEventListener("pointerdown", onDocPointerDown, true);
-    document.removeEventListener("keydown", onKeydown);
+    window.removeEventListener("keydown", onKeydown, { capture: true });
     window.removeEventListener("scroll", onScrollOrResize, true);
     window.removeEventListener("resize", onScrollOrResize);
   }
@@ -179,8 +199,30 @@ export const statusTip: Action<HTMLElement, StatusTipParams | null | undefined> 
     show();
     if (e.detail > 0) pinned = true; // genuine pointer click pins; keyboard (detail 0) does not
   }
+  // Escape lives on the WINDOW, in capture phase, only while a tip is open — not on the
+  // trigger. Most triggers are non-focusable, so a real keystroke targets <body> and would
+  // never reach a node listener (#2283); a pinned tip would then have no keyboard exit at
+  // all, failing WCAG 2.2 SC 1.4.13 (Dismissible).
+  //
+  // Whether the key is CONSUMED follows intent, because these two states differ:
+  //  - PINNED (a deliberate pointer click): the tip is an overlay the operator opened, so it
+  //    owns this Escape. Capture + preventDefault, because `a11yDialog` closes its host dialog
+  //    unless `defaultPrevented` and its listener sits on the dialog NODE — whose bubble phase
+  //    runs before any bubble-phase window listener. Without this, a pinned tip in the New Task
+  //    modal (InstrumentToggle's ON/OFF suffix) would dismiss itself only after the modal had
+  //    already closed, discarding the operator's typed task. Protocol per IssueDetailsPopover.
+  //  - HOVERED (incidental): the operator never asked for this tip, so it must NOT steal a
+  //    keystroke aimed elsewhere. Hide it, but let the event run on — otherwise hovering a
+  //    command description in SlashCommandMenu would swallow the Escape that closes the menu.
+  // `stopPropagation` (not `stopImmediatePropagation`) so several open tips close on one Esc.
   function onKeydown(e: KeyboardEvent) {
-    if (e.key === "Escape") hide();
+    if (e.key !== "Escape") return;
+    const owned = pinned; // hide() clears it — read first
+    if (owned) {
+      e.preventDefault();
+      e.stopPropagation();
+    }
+    hide();
   }
 
   function enable(next: StatusTipParams) {
@@ -207,7 +249,6 @@ export const statusTip: Action<HTMLElement, StatusTipParams | null | undefined> 
       node.addEventListener("focus", onFocus);
       node.addEventListener("blur", onBlur);
       node.addEventListener("click", onClick);
-      node.addEventListener("keydown", onKeydown);
     }
   }
 
@@ -220,7 +261,6 @@ export const statusTip: Action<HTMLElement, StatusTipParams | null | undefined> 
       node.removeEventListener("focus", onFocus);
       node.removeEventListener("blur", onBlur);
       node.removeEventListener("click", onClick);
-      node.removeEventListener("keydown", onKeydown);
     }
     if (body) {
       void unmount(body);


scoop (commented):
# Review — statusTip: reachable Escape, honest keyboard/AT contract (#2283)

Diff reviewed: `git diff 3a52d9057...HEAD` — three files (`ui/src/lib/actions/statusTip.svelte.ts`,
`ui/src/lib/components/StatusPip.browser.test.ts`, `ui/src/lib/components/UnitRow.svelte`).

**The listener move is wired symmetrically.** `show()` adds `window.addEventListener("keydown",
onKeydown, { capture: true })` at `ui/src/lib/actions/statusTip.svelte.ts:156`; `hide()` removes it
with the *same* `{ capture: true }` option at `ui/src/lib/actions/statusTip.svelte.ts:169`. Capture
flags match, so removal is real, not a silent no-op. The node-level `keydown` add/remove is gone from
`enable()` (`:238-245`) and `teardown()` (`:250-257`) with no residue — the other four node listeners
(`pointerenter`/`pointerleave`/`focus`/`click`… and `blur`) are still paired.

No leak path: `show()` registers only after `open = true` is reached
(`ui/src/lib/actions/statusTip.svelte.ts:148-158` — it returns early if `showPopover()` throws, before
any registration), `hide()` is the only remover and `teardown()` calls it first
(`ui/src/lib/actions/statusTip.svelte.ts:249`).

**Capture-on-window really does beat a11yDialog.** `ui/src/lib/a11yDialog.ts:84` registers its handler
on the dialog *node* with no capture flag, i.e. bubble phase at an ancestor of the focused target, and
its Escape branch returns early on `e.defaultPrevented` (`ui/src/lib/a11yDialog.ts:55-60`). A
window-capture listener runs before any of that, so the pinned branch's `preventDefault()` +
`stopPropagation()` (`ui/src/lib/actions/statusTip.svelte.ts:213-216`) genuinely protects the New Task
modal. This is the same protocol `ui/src/lib/components/IssueDetailsPopover.svelte:44-45,59,63` already
implements (window + `{capture:true}` + `preventDefault`), so the comment's "Protocol per
IssueDetailsPopover" is accurate.

**`stopPropagation`, not `stopImmediatePropagation`, is the right choice and the comment describes it
correctly** (`ui/src/lib/actions/statusTip.svelte.ts:209`). Both listeners sit on the same node
(`window`) in the same phase, so `stopPropagation` leaves sibling tip listeners intact — several open
tips do all close on one Escape — while still preventing descent to `document` and the target.

**The pinned/hovered asymmetry is load-bearing, not decoration.** The hovered branch deliberately does
not call `stopPropagation`. I checked the case the comment cites: `SlashCommandMenu.svelte:71-81`
attaches the tip to a `.sc-desc` `<div>` with `stopClickPropagation: false`, and the menu's Escape is
handled by a *node* handler on the compose textarea — `ui/src/lib/components/ComposeBar.svelte:339`
binds `onkeydown={onKeydown}` on the `<textarea>` at `:326`, and that handler closes the slash menu at
`ui/src/lib/components/ComposeBar.svelte:264-272`. A window-capture `stopPropagation` would have
starved it. So the asymmetry prevents a real regression rather than inventing a distinction.

**Doc-block factual claims all check out against the tree:**
- "BuildQueueBadge's interactive branches are real `<button>`s (its coarse-pointer branch is a
  read-only `<span>`)" — `ui/src/lib/components/BuildQueueBadge.svelte:74` is the `!interactive`
  `<span role="img">`, `:100` and `:116` are `<button type="button">`. So `focus`/`blur` are *not*
  dead listeners, and the plan's success criterion 4 holds.
- "`role="img"` + `aria-label` … the sandbox/quota chips" —
  `ui/src/lib/components/unit-row/UnitRowRight.svelte:229-231` (quota) and `:245-247`, `:253-255`,
  `:261-263`, `:269-271`, `:277-279` (sandbox) each carry `role="img"` + `aria-label` alongside
  `use:statusTip`.
- "visible text (SlashCommandMenu, PromptSources, IssueLoadAttempts)" —
  `SlashCommandMenu.svelte:80`, `PromptSources.svelte:467-475`, `IssueLoadAttempts.svelte:68`; the
  PromptSources trigger is a `<span>` nested in the row `<button>` (`PromptSources.svelte:456`), which
  is consistent with the doc block's "most are plain span/div/li" since `focus`/`blur` do not bubble.
- The badges — `StatusPip.svelte:38`, `CriticBadge.svelte:209`, `PlanGateBadge.svelte:235,267`,
  `AutopilotBadge.svelte:28,38,48`, `TerminalBadge.svelte:13`, `ResearchBadge.svelte:13` — all
  `role="img"`.

**The UnitRow comment correction is accurate.** `.unit-hit` is a real `<button type="button">` with
`aria-describedby={describedBy}` at `ui/src/lib/components/UnitRow.svelte:870-875`, so "the row's own
`.unit-hit` overlay is the keyboard entry point (it carries `aria-describedby`)" is true, and the
removed sentence ("the keyboard reads `aria-description` instead") was exactly the optimistic claim the
issue asked to retire. The chip is still a non-focusable `<span>` — no `tabindex`, no new role anywhere
in the diff, so the plan's out-of-scope boundary is respected.

**Behavioural walk-through, no defect found.** After an Escape on a hovered tip the pointer is still
over the trigger and `pointerenter` will not re-fire, so the tip stays dismissed until a genuine
re-trigger — which is what WCAG 2.2 SC 1.4.13 *Dismissible* wants, not a bug.
`hide()` clears `pinned` before its `if (!open) return` guard
(`ui/src/lib/actions/statusTip.svelte.ts:161-164`), so a later `pointerleave` →
`scheduleClose` → `hide` is an inert no-op rather than a stuck-pinned state. Keyboard activation of
BuildQueueBadge's button still yields `detail === 0` and therefore does not pin
(`ui/src/lib/actions/statusTip.svelte.ts:192`), so a focus-opened tip dismisses on Escape *without*
consuming it — the button keeps its own Escape semantics.

**No new identifiers, no catalog work.** The diff introduces no `m.*` message keys and no new imports
beyond `userEvent` from `vitest/browser` in the test file (`StatusPip.browser.test.ts:3`), which is
already the idiom elsewhere in the suite. Locale parity is therefore not in play — nothing to compare
in `ui/messages/`.

**Regression sweep for the removed node listener.** The only other test in the repo that dispatches an
Escape `keydown` at a node is `ui/src/lib/components/UnitRow.browser.test.ts:437`, which targets the
ask-mode chooser, not a statusTip trigger. Nothing depended on the listener that was deleted.

## Interactions I checked and judged benign

`ui/src/lib/components/Viewport.svelte:1912-1942` registers a desktop Escape rescue on `window` in
capture phase with `e.stopImmediatePropagation()` at mount, so it runs *before* statusTip's
show-time listener and can preempt it. The preconditions in
`ui/src/lib/terminalEscape.ts:32-49` require `termTabActive && live && !overlayOpen` and focus on
body/terminal, and the triggers reachable in that state gate their tooltip behind a `tip` prop that is
Herd-card-only (e.g. `BuildQueueBadge.svelte:9`), so the practical overlap is small. The plan
explicitly accepted this ordering (interrupting the agent outranks closing a tooltip) and outside-click
/ scroll still dismiss. Not a finding.

## Stated limitations

- I did not verify `aria-description` announcement behaviour empirically; no screen reader is available
  here. The diff's claim is a platform-level statement, and the design it justifies (never let the tip
  be the sole carrier of meaning) is independently verifiable in the markup, which I did check above.
- I did not run the test suite or lint — per instructions, CI owns that.

- `ui/src/lib/components/GlossaryTerm.svelte` still handles Escape on a bubble-phase window listener without `preventDefault`, so an open glossary popover inside New Task closes the host modal — the plan names this and deliberately does not widen into it.
- `ui/src/lib/components/IssueDetailsPopover.svelte` `preventDefault`s Escape unconditionally (no pinned/hovered distinction), so it can swallow an Escape aimed elsewhere in a way this action now avoids.

- `ui/src/lib/actions/statusTip.svelte.ts`: the pinned-vs-hovered consumption split is behaviour the plan did not describe (it specified an unconditional `preventDefault()` + `stopPropagation()`); it is a defensible refinement and the third spec pins it, but it is extra behaviour surface beyond the approved design.


**Nits (non-blocking):**

- [bug] ui/src/lib/actions/statusTip.svelte.ts:5 — the StatusTipParams.text docstring still says the explanation is "exposed to AT via aria-describedby", but enable() sets aria-description (:230), and the rewritten block 30 lines below (:47-53) now makes the aria-description vs aria-describedby distinction its central point. This PR is the honesty pass for exactly that claim; the stale one-liner directly contradicts it.
- [bug] ui/src/lib/actions/statusTip.svelte.ts:212 — `const owned = pinned; // hide() clears it — read first` guards against a hazard the written order already rules out: the `if (owned)` branch runs entirely before `hide()` at :217, so a plain `if (pinned)` is equivalent. The comment implies an ordering constraint the code does not have.
- [bug] ui/src/lib/actions/statusTip.svelte.ts:210 — onKeydown has no `e.isComposing` guard, unlike the repo's other Escape interceptor (ui/src/lib/terminalEscape.ts:22,40). A pinned tip during IME composition would preventDefault/stopPropagation the Escape that cancels the candidate. Reachability is thin (pin by pointer, then compose in a field), so this is a defensive nit rather than a live bug.
- [scope] ui/src/lib/actions/statusTip.svelte.ts:40-42 — "it is a window-level listener scoped to the open state, so it reaches a pinned tip on any trigger" is unqualified, but Viewport's desktop Escape rescue (ui/src/lib/components/Viewport.svelte:1912-1942) is window-capture + stopImmediatePropagation registered at mount and therefore preempts it. The plan said to note that interaction in the doc block; it is not noted.

scoop (commented):
# Review — statusTip: reachable Escape, honest keyboard/AT contract (#2283)

Diff reviewed: `d319816658b190d3969ba36befe75d8515ba3012...HEAD` — three files
(`ui/src/lib/tooltips/statusTip.svelte.ts`, `ui/src/lib/components/StatusPip.browser.test.ts`,
`ui/src/lib/components/UnitRow.svelte`).

## Verdict

No blocking defect found. The listener move is correct, the event-ordering argument the change
rests on holds against the actual code, and every factual claim the new comments make is true of
the tree as it stands. Two nits below.

**The a11yDialog interaction the fix is built on is real.**
`ui/src/lib/a11yDialog.ts:55-61` — Escape returns early on `e.defaultPrevented` (`:58`) before
`e.preventDefault(); current.onclose?.()`. Its listener is registered on the dialog **node**
(`ui/src/lib/a11yDialog.ts:84`), not on window, so a node-bubble listener does run ahead of any
bubble-phase window listener. Capture-on-window (`ui/src/lib/tooltips/statusTip.svelte.ts:164`) is
therefore genuinely the earliest point in the path and the `preventDefault` at `:222` lands before
a11yDialog reads the flag. The precedent the comment cites is also real:
`ui/src/lib/components/IssueDetailsPopover.svelte:10-11` documents the same protocol and
`:44-45,59` implement it (`window.addEventListener("keydown", onKeydown, { capture: true })` +
`preventDefault`).

**The New Task scenario the comment names is reachable.**
`ui/src/lib/components/new-task/InstrumentToggle.svelte:66` attaches `statusTip` to a
`<span class="status">` that is a DOM sibling of the toggle button (`:12`, `:45`), inside the New
Task modal. A pointer click on that span sets `pinned` (`statusTip.svelte.ts:200`), so the
pinned-consumes branch is exactly what keeps the modal open. Cited claim checks out.

**The doc block's claim that `focus`/`blur` are not dead code checks out.**
`ui/src/lib/components/BuildQueueBadge.svelte:74` is the coarse-pointer read-only `<span>`;
`:100` and `:116` are real `<button>` branches, both carrying `use:statusTip` (`:109`, `:126`).
`ui/src/lib/components/unit-row/UnitRowRight.svelte:223-225` renders it with `tip` and
`interactive={!coarsePointer}`, so the button branches do receive the action with a non-null param.
`ui/src/routes/design-system/+page.svelte:1138-1140` is a second focusable trigger. Keeping
`onFocus`/`onBlur` node-wired at `statusTip.svelte.ts:249-250` is correct; the issue's premise that
every trigger is non-focusable is the part that does not survive contact with the code.

**The `aria-description` / accessible-name invariant the doc block asserts is true of every call
site today.** Spot-checked each family it names: `role="img"` + `aria-label` on
`TerminalBadge.svelte:13-16`, `ResearchBadge.svelte:13-16`, `PlanGateBadge.svelte:235-238` and
`:267-269`, `CriticBadge.svelte:209-211`, `AutopilotBadge.svelte:27-49`,
`BuildQueueBadge.svelte:78-87`, and the sandbox/quota chips at
`unit-row/UnitRowRight.svelte:227-231` and `:243-281`; visible text on
`SlashCommandMenu.svelte:73-81` (`.sc-desc` renders `cmd.description`),
`PromptSources.svelte:467-469`, `IssueLoadAttempts.svelte:68`, and the attention/merging/ready
badges at `unit-row/UnitRowRight.svelte:283-311`. Nothing in the tree contradicts the invariant the
comment tells future callers to honour.

**The `UnitRow.svelte` comment correction is accurate.** `.unit-hit` is a `<button>` carrying
`aria-describedby={describedBy}` (`ui/src/lib/components/UnitRow.svelte:873-876`, derived at
`:540-542`), so "the row's own `.unit-hit` overlay is the keyboard entry point (it carries
`aria-describedby`)" at `:624-625` is factual. The chip is still a non-focusable `<span>`, so
"focus-to-open never fires on it" is also factual.

**The SlashCommandMenu justification for NOT consuming a hovered Escape is real.**
`ui/src/lib/components/ComposeBar.svelte:264-272` closes the slash menu on Escape from a handler
bound to the textarea; a capture-phase `stopPropagation()` at window would prevent that keydown
from ever reaching the textarea. The test comment at
`StatusPip.browser.test.ts:117-119` describes this correctly.

**The listener swap is symmetric and leak-free.** `show()` adds with `{ capture: true }`
(`statusTip.svelte.ts:164`) and `hide()` removes with the same capture flag (`:177`); `show()`
early-returns when already open (`:153`) so no duplicate registration; `teardown()` calls `hide()`
first (`:256`) so `destroy()` cannot strand a window listener. The node-level `keydown` add/remove
pair is removed from both `enable()` and `teardown()` — no orphaned half.

**No regression for the previously-working global path.** The pre-change `show()` already
registered `document.addEventListener("keydown", onKeydown)` (verified by
`git show d3198166:ui/src/lib/tooltips/statusTip.svelte.ts`), so Escape did reach a pinned tip
outside dialogs. The change is a strict improvement: earlier in the path, and `defaultPrevented`
set in time for a11yDialog.

**The untouched sibling test still passes.** `ui/src/lib/tooltips/statusTip.browser.test.ts:37`
dispatches Escape on `document`; `window` is in the propagation path for an event whose target is
`document`, so the new window-capture listener still fires, `pinned` is false (hover-only), and
`hide()` runs synchronously through `stopAnchor?.()` → `hidePopover()`. The `:popover-open`
assertion at `:38` is unaffected.

**Terminal interaction is unchanged, not newly broken.** `ui/src/lib/components/Viewport.svelte:1942`
registers its rescue at mount in window capture and calls `stopImmediatePropagation()` (`:1935`)
only when `shouldForwardEscape` passes (`:1917-1933`, which stands down while an overlay is open).
Because it registers before any tip opens, it wins on same-node/same-phase ordering — and it would
equally have suppressed the old `document` listener. No behaviour change either way.

**No new user-facing strings**, so no EN/DE catalog parity work is implied; the diff adds no `m.*`
key. `CLAUDE.md:14-16` and `.claude/rules/ui-design-system.md:22` are the only non-`ui/src`
mentions of `statusTip` and neither documents Escape behaviour, so nothing there goes stale.

**Import form checked.** `import { page, userEvent } from "vitest/browser"` matches the existing
idiom verbatim at `ui/src/lib/components/CommandBar.browser.test.ts:3`.

## On the pinned-vs-hovered split

The plan called for `preventDefault()` + `stopPropagation()` unconditionally; the implementation
consumes only when `pinned` (`statusTip.svelte.ts:219-225`). I judged this on its own merits, not
against the plan: it is the safer of the two, it directly answers the plan's own stated risk ("an
open tip now swallows an Escape the operator meant for something else"), and the WCAG 2.2 SC 1.4.13
obligation is satisfied either way because `hide()` runs in both branches. The asymmetry is
documented at `:207-217` and pinned by two specs
(`StatusPip.browser.test.ts:93-106` and `:108-120`). No objection.

## Limitations

- I did not run tests or lint (CI owns those, per the review brief). Test correctness was assessed
  by reading only.
- No empirical screen-reader verification of `aria-description` announcement was possible here, and
  the diff does not claim any — the doc block states the platform-level limitation rather than an
  announcement result. That is the honest framing; I can neither confirm nor refute the underlying
  platform claim from this repo, and I am not treating it as a finding either way.

- `ui/src/lib/tooltips/statusTip.browser.test.ts:37` still drives Escape with a synthetic `document.dispatchEvent`, the same "models something the browser cannot produce" pattern this PR calls out and replaces in `StatusPip.browser.test.ts:89` — it happens to still pass, but it is not a real-keystroke seam.
- `ui/src/lib/components/GlossaryTerm.svelte` uses a bubble-phase window Escape listener with no `preventDefault`, so an open glossary popover inside New Task still closes the host modal on Escape — a different component's defect, deliberately untouched here.


**Nits (non-blocking):**

- [bug] ui/src/lib/tooltips/statusTip.svelte.ts:218-226 — onKeydown has no `e.isComposing` guard, so a pinned tip consumes the Escape that cancels an IME candidate. Viewport.svelte:1916 explicitly stands down in that case ("Escape there cancels the candidate, not ours") and shouldForwardEscape also excludes ctrl/alt/meta-modified Escapes; consider matching that gate before preventDefault/stopPropagation.
- [compliance] ui/src/lib/tooltips/statusTip.svelte.ts:220 — `const owned = pinned; // hide() clears it — read first` implies an ordering hazard that does not exist: hide() is called after the branch, so reading `pinned` directly in the `if` is equivalent. Either inline it or drop the trailing clause, which reads as a warning about code that is no longer arranged that way.

